PR #7220, opened 2023-03-06
test: demonstrate bug depending on package that installs directories

anmonteiro:
The test is filed under directory-targets, but this is present in any package that installs with `(install (dirs ...))`


emillon:
@anmonteiro FYI there's a tentative fix for this in #8918

PR #4422, opened 2021-03-30
swallow stdout on success

ghost:
This PR adds an option so that Dune may swallow the output of build actions when they succeed. We do that in Jenga, so we need a similar option in Dune.

In practice, as soon as your build is large enough you want a successful build to print nothing, otherwise it's all just noise.

This option is allowed (by decreasing precedence):

- on the command line
- in the workspace file
- in the config file

PR #2706, opened 2019-10-07
Re-implement File_tree on top of memoization

rgrinberg:
Instead of generating a lazy file system tree, functions which read a
directory are memoized.

This PR is by no means done, and I realize that we need `Memo.Cell.t` before it's really usable but I'd like to push through anyway not to lose the progress I've made a couple of weeks ago.

There are three things left to do:

~Implement `Memo.Cell.t` will be done in a separate PR at first~ <- This one will be done separately.

- [x] Fix cycle detection.

- [x] Add progress reporting (of directories scanned).

The last two points I need some advice on. IIRC, there was a suggestion to have a memoized function `val dirs_visited : Path.t -> File.Set.t`, but I'm not entirely sure how that's going to work.

PR #7224, opened 2023-03-06
fix: add a dependency on current build run in `Source_tree.root`

### LINKED ISSUE #5549: dune build -w restricting job number for no reason
## Problem description

There seems to be an issue with `dune build -w`. It can be observably slow due to a restriction of jobs. This can be seen as slow, since stopping watch mode and doing a fresh `dune build` will be much faster. The issue seems to be during the rule finding phase.

Here is a demonstration GIF of this behaviour. Details of what is happening are below:

dune_build_watch_slowdown_rule_finding

1. `dune build -w` is run, and it works as expected. Most of the time is spent rule finding, but due to cache and everything already being built, it finishes quickly.
2. Next whilst watch mode is still running, an .ml file early in the dependency tree is changed, triggering a rebuild. Immediately you can observe the number of jobs drop to a low count. What is dune doing here?
3. To demonstrate that this is an issue with watch mode, a fresh `dune build` is triggered, and as expected everything builds normally and quickly.

There are some key details with this setup. This is the main Coq repo being changed, and importantly there is ml code being edited which are later being depended on by dune Coq stanzas. Retrying the same test, but this time with the `@check` target (building only ml code) results in no observable slow down, leading me to believe that this is an issue with the interaction of rule finding and the Coq rules.

## Reproduction

1. Clone the https://github.com/coq/coq repo.
2. Do `dune build -w`. This will take a while but once it starts building .v files (see from `--display=short` or `--verbose`) you can do the next step.
3. Edit an ml file, for example `kernel/declareops.ml`, you can even drop `let _hello = "world"` inside.
4. Observe that the rebuild will now restrict itself to a low job count.
5. Stopping watch mode and rebuilding will speed up the build again.

## Specifications

- Version of `dune` (output of `dune --version`): 3.0.3
- Version of `ocaml` (output of `ocamlc --version`) 4.12.1
- Operating system (distribution and version): Ubuntu

cc @rgrinberg

jchavarri:
Fixes #5549.

Adds a dependency on current build run in `Source_tree.root`. The fix was found by bisecting until I found the place where the parallelism in `dune watch` regressed (more details on original issue conversation). Hopefully the fix is correct, and someone with more knowledge can explain what the issue was, and why this fixes it 😅 Debugging and understanding memoization issues is not straightforward to me.

- The dependency on current build run was originally added in https://github.com/ocaml/dune/pull/2706.
- Then removed in https://github.com/ocaml/dune/pull/4422, where dune watch parallelism regressed.

### DIFF
--- a/src/dune_engine/source_tree.ml
+++ b/src/dune_engine/source_tree.ml
@@ -586,6 +586,7 @@ end = struct
     | Some kind -> Dir0.This { Vcs.kind; root = Path.(append_source root) path }
 
   let root () =
+    let* (_ : Memo.Run.t) = Memo.current_run () in
     let path = Path.Source.root in
     let dir_status : Sub_dirs.Status.t = Normal in
     let error_unable_to_load ~path unix_error =


jchavarri:
Hm, one realization while testing this branch is that incremental builds in watch mode that would be almost instant before are way slower in the branch because dune has to re-create rules, etc. So the parallelization works again, but at the cost of having to recalculate a lot of things that were already calculated.

PR #7219, opened 2023-03-05
chore: update flakes, melange

anmonteiro:
- run `nix flake update` to get the newest melange hash
  - tests dune with the most recent melange, using dune to compile its own stdlib / runtime

PR #7211, opened 2023-03-03
test: nested melange.emit stanzas

anmonteiro:
as noted in https://github.com/ocaml/dune/pull/7208#issuecomment-1452421445, we should forbid nesting `melange.emit` stanzas.

Adding a test that shows it's currently allowed.

PR #7229, opened 2023-03-06
test(melange): demonstrate issue with 2 melange.emit stanzas in the same dir

anmonteiro:
I believe this is a regression introduced within the past 2-3 days

### DIFF
--- a/src/dune_rules/gen_rules.ml
+++ b/src/dune_rules/gen_rules.ml
@@ -386,35 +386,24 @@ type for_melange =
     stanza : Melange_stanzas.Emit.t
   }
 
-let rec nearest_parent_melange_emit ~dir =
+(* Detect if [dir] is under the target directory of a melange.emit stanza. *)
+let rec under_melange_emit_target ~dir =
   match Path.Build.parent dir with
   | None -> Memo.return None
   | Some parent -> (
     let* stanzas = Only_packages.stanzas_in_dir parent in
     match stanzas with
-    | None -> nearest_parent_melange_emit ~dir:parent
+    | None -> under_melange_emit_target ~dir:parent
     | Some stanzas -> (
       match
         List.find_map stanzas.stanzas ~f:(function
-          | Melange_stanzas.Emit.T mel -> Some mel
+          | Melange_stanzas.Emit.T mel ->
+            let target_dir = Melange_rules.emit_target_dir ~dir:parent mel in
+            Option.some_if (Path.Build.equal target_dir dir) mel
           | _ -> None)
       with
-      | None -> nearest_parent_melange_emit ~dir:parent
+      | None -> under_melange_emit_target ~dir:parent
       | Some stanza -> Memo.return @@ Some { stanza_dir = parent; stanza }))
-
-(* Detect if [dir] is under the target directory of a melange.emit stanza. *)
-let rec under_melange_emit_target ~dir =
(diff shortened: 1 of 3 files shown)


anmonteiro:
This was broken in https://github.com/ocaml/dune/pull/7211

test/blackbox-tests/test-cases/melange/two-emit-stanzas.t
```diff
@@ -0,0 +1,27 @@
+
+  $ cat > dune-project <<EOF
```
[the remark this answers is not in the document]
anmonteiro: done

anmonteiro:
fixed the issue in [d647296](https://github.com/ocaml/dune/pull/7229/commits/d647296a9cf21c6046a0f24bd2af6e4ea46d9ebf). The root cause is that we were just returning the first melange.emit stanza in the target directory rather than all.

anmonteiro:
The only test failure is related to #7219 which I'm fixing separately.

src/dune_rules/gen_rules.ml
```diff
@@ -386,35 +386,41 @@ type for_melange =
...
+      List.find_opt for_melange_stanzas ~f:(fun for_melange ->
+          let target_dir =
+            Melange_rules.emit_target_dir ~dir:for_melange.stanza_dir
+              for_melange.stanza
+          in
+          Path.Build.is_descendant dir ~of_:target_dir)
+    with
+    | Some for_melange -> Memo.return (Some for_melange)
+    | None ->
+      let for_melange_in_this_dir = List.hd for_melange_stanzas in
```
jchavarri, line 422: The "for_melange" prefixes used across values in this function are a bit confusing to me. What do they mean? Would it make sense to replace them with "emit" or just removed?

jchavarri (approved):

src/dune_rules/gen_rules.ml
anmonteiro: `for_melange` refers to the `type for_melange` just above the diff (you need to expand it once).

src/dune_rules/gen_rules.ml
```diff
@@ -386,35 +386,41 @@ type for_melange =
     stanza : Melange_stanzas.Emit.t
   }
 
-let rec nearest_parent_melange_emit ~dir =
+let rec nearest_parent_melange_emits ~dir =
```
rgrinberg, line 389: Can you revert to the original approach with `under_melange_emit_target`? I think it's confusing to collect this list of emit stanzas when only one is going to be relevant.

anmonteiro: I can do that, but I think that'll imply we can't reuse this function.

rgrinberg: I don't see why not, in both uses of this function, you end up looking at only one element in the list.

anmonteiro: yeah, my bad, we can reuse the logic.

anmonteiro: I think I had just originally overcomplicated the logic here. `a/ (target output)` and `a/b/c/ (target output)` are allowed to co-exist, and I was being overly restrictive. However `a/ (target output)` and `a/output/b` can't, and the original function can detect this case too.

anmonteiro: This is fixed in the most recent push.

PR #7231, opened 2023-03-07
test(melange): fix merlin test under linux

anmonteiro:


anmonteiro:
This is still `***** UNREACHABLE *****` on linux.. looking for other fixes

Alizter:
Why not do `which melc`?

anmonteiro:
`which melc` prints the path to the melange binary

`melc --where` prints the melange standard library paths

anmonteiro:
I'm also unable to repro the failure on a linux box by just running `dune build @test/blackbox-tests/test-cases/melange/merlin`

test/blackbox-tests/test-cases/melange/merlin.t
```diff
@@ -2,11 +2,18 @@
...
-  $ export BUILD_PATH_PREFIX_MAP="/MELC_STDLIB=$melc_stdlib3:$BUILD_PATH_PREFIX_MAP"
+  $ melc --where > melc_where.txt
+  $ cat > ./script.js <<EOF
+  > #!/usr/bin/env node
+  > var fs = require("fs");
+  > var contents = fs.readFileSync("melc_where.txt")
+  >    .toString("utf8")
+  >    .split(/:/)
+  >    .map(x => x.trim());
+  > contents.forEach(x => process.stdout.write(\`export BUILD_PATH_PREFIX_MAP="/MELC_STDLIB=\${x}:\$BUILD_PATH_PREFIX_MAP"; \`))
```
rgrinberg, line 13: If we don't want to both with sh let's just stick to OCaml. `String.split_on_char`, `String.trim`, it's really not that bad :)

I'd like to see:

1. Write the script in a separate file
2. Add it as a dependency to the test
3. Run it as `export BUILD_PATH_PREFIX_MAP="$(script.exe file)":$BUILD_PATH_PREFIX_MAP`

See `test/blackbox-tests/utils` for some examples.

anmonteiro: attempted something like this in [d3d3f8d](https://github.com/ocaml/dune/pull/7231/commits/d3d3f8d23f78520e275c8c2163fe51a4268e9f36)

rgrinberg:
Thanks for the fix

PR #7233, opened 2023-03-07
feature: promote files atomically

rgrinberg:
Instead of over writing a file in the source tree, we first write in a
staging area and then atomically move it to the source


otherlibs/stdune/src/io.ml
```diff
@@ -336,6 +336,20 @@ let portable_symlink ~src ~dst =
         Unix.symlink src dst)
     | exception Unix.Unix_error _ -> Unix.symlink src dst
 
+let copy_atomically f ?chmod ~src ~dst () =
```
anmonteiro, line 339: why does this function need to take `f`? It looks like it's only ever called with `copy_file`

anmonteiro (approved):

otherlibs/stdune/src/io.ml
rgrinberg: Fixed. Good catch.

otherlibs/stdune/src/io.ml
```diff
@@ -252,23 +252,24 @@ struct
...
     match Copyfile.available () with
-    | false -> copy_file
+    | false -> gen_copy_file
     | true -> (
-      fun ?chmod ~src ~dst () ->
+      fun ?chmod ~src ~dst ~dst_maybe_present () ->
         let src = Path.to_string src in
         let dst = Path.to_string dst in
-        (try Copyfile.copyfile src dst with
+        (try Copyfile.copyfile src dst ~dst_maybe_present with
```
anmonteiro, line 272: what's the goal of this labeled arg? It's always true for `Copyfile.copyfile` it seems?

rgrinberg: Yes, but it's not always true when `~atomically:true`. The goal is to shave off an `unlink` that we know is always going to be useless on macos.

anmonteiro: ooops, missed that. Thanks!

anmonteiro (approved):
LGTM

otherlibs/stdune/src/io.ml
```diff
@@ -336,6 +339,19 @@ let portable_symlink ~src ~dst =
         Unix.symlink src dst)
     | exception Unix.Unix_error _ -> Unix.symlink src dst
 
+let copy_atomically ?chmod ~src ~dst () =
+  let dst_basename = Path.basename dst in
+  let dir = Temp.create Dir ~prefix:(Path.basename src) ~suffix:"" in
```
snowleopard, line 344: Is that going to create a temp directory for every file copy? Seems pretty bad.

snowleopard:
Have you measured how much slower such atomic copies are, given that the current implementation creates a temp directory every time? It seems worth doing that and mentioning the cost in the API.

rgrinberg:
> Have you measured how much slower such atomic copies are, given that the current implementation creates a temp directory every time? It seems worth doing that and mentioning the cost in the API.

Never measured it, but it would make sense to use a single directory for all of this. Note that its unlikely to matter for performance either way. We only do the copies atomically for promotions, and its quite unlikely there's a significant number of those.

snowleopard:
> We only do the copies atomically for promotions

This PR changes some code in sandboxing, so surely that's not true?

rgrinberg:
> This PR changes some code in sandboxing, so surely that's not true?

Oh yeah, but that's only for `Patch_back_source_tree` which we're going to remove. We can make it non-atomic, I don't particularly care.

rgrinberg:
Actually, if we only use a single directory for all promotions, we can get rid of the entire performance penalty. Currently, we make sure to `unlink` the destination we're promoting, so we'd really be doing the exact same thing but replacing `unlink` with `rename`. Seems like a win-win to me.

snowleopard:
> Oh yeah, but that's only for `Patch_back_source_tree` which we're going to remove. 

I don't think we can remove it -- we use it internally for some things.

> Actually, if we only use a single directory for all promotions, we can get rid of the entire performance penalty. 

Sounds good to me!

PR #7199, opened 2023-02-28
feat(melange): add `runtime_deps` to copy assets to the target dir

anmonteiro:
- as the test indicates, this isn't yet ready for consumption because dune currently lacks the ability to generate aliases for generated directories

PR #7234, opened 2023-03-07
feat(melange): add `melange.runtime_deps` field in library

anmonteiro:


test/blackbox-tests/test-cases/melange/library-with-runtime-deps.t
```diff
@@ -0,0 +1,63 @@
...
+  $ ls _build/default/output/lib
+  foo.js
+  index.txt
+  $ node _build/default/output/main.js
+  hello from file
+  
+  Some text
+  
+
+
```
rgrinberg, line 61: Is this blank space needed?

test/blackbox-tests/test-cases/melange/public-library-with-runtime-deps.t
```diff
@@ -0,0 +1,94 @@
...
+  >  (melange.runtime_deps index.txt nested/hello.txt))
+  > EOF
+  $ cat > lib/foo.ml <<EOF
+  > let dirname = [%bs.raw "__dirname"]
+  > let () = Js.log2 "dirname:" dirname
+  > let file_path = "./index.txt"
+  > let read_asset () = Node.Fs.readFileSync (dirname ^ "/" ^ file_path) \`utf8
+  > EOF
+
+  $ dune build --display=short
```
rgrinberg, line 34: As a general rule, we should avoid the `--display=short` whenever possible in tests. It usually just adds noise.

src/dune_rules/dune_package.ml
```diff
@@ -62,6 +62,19 @@ module Lib = struct
...
     in
+    let melange_runtime_deps =
+      match Lib_info.melange_runtime_deps info with
+      | Local _ -> assert false
+      | External paths ->
+        let lib_dir = Obj_dir.dir obj_dir in
+        List.map
+          ~f:(fun p ->
+            let p =
+              Path.as_in_build_dir_exn p |> Path.Build.drop_build_context_exn
```
rgrinberg, line 73: Can external paths make it here?

src/dune_rules/melange/melange_rules.mli
```diff
@@ -18,3 +18,7 @@ val setup_emit_js_rules :
   -> sctx:Super_context.t
   -> Melange_stanzas.Emit.t
   -> unit Memo.t
+
+module Runtime_deps : sig
+  val eval : expander:Expander.t -> Dep_conf.t list -> Path.Set.t Memo.t
```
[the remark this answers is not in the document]
rgrinberg: Also, if this only returns build paths, then it should be `Path.Build.Set.t`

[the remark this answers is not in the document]
anmonteiro: I believe it can return external paths. There was a test for that case in #7199

src/dune_rules/dune_package.ml
anmonteiro: yeah I guess there's nothing stopping that. What do you think is the best course of action here? Forbid external paths by raising a user_error here?

rgrinberg: Here would probably be late. You would be able to build the library while its local and only fail when running the install rules. I would think you need to verify this earlier then.

Come to think of it, we shouldn't allow external paths as runtime deps for public libraries at all.

anmonteiro: > Come to think of it, we shouldn't allow external paths as runtime deps for public libraries at all.

This makes sense to me. Is `install_rules.ml` the best place for that check?

rgrinberg: I guess so. I would also add it to the corresponding `@check` for the library

test/blackbox-tests/test-cases/melange/installed-library-with-runtime-deps.t
```diff
@@ -0,0 +1,80 @@
+Test simple interactions between melange.emit and copy_files
```
jchavarri, line 1: I think this should be

```suggestion
Test simple interactions between melange.emit and runtime_deps
```

? Same for other 2 tests.

src/dune_rules/install_rules.ml
```diff
@@ -130,42 +139,57 @@ end = struct
...
               make_entry ?sub_dir Lib source ?dst))
     in
+    let* melange_runtime_deps =
+      let* expander = Super_context.expander sctx ~dir:lib_src_dir in
+      Melange_rules.eval_runtime_deps ~expander (snd lib.melange_runtime_deps)
+    in
+    let melange_runtime_entries =
+      Path.Set.to_list_map
+        ~f:(fun path ->
+          let path = Path.as_in_build_dir_exn path in
```
jchavarri, line 180: If the path does not exist, it will fail here right? Do you think this case should be handled? Or add a test to see how current code behaves?

src/dune_rules/install_rules.ml
```diff
@@ -130,42 +139,57 @@ end = struct
...
+          let sub_dir =
+            let src_dir = Path.Build.parent_exn path in
```
jchavarri, line 182: Same questions: in which cases could it fail here? and are they reproducible in tests?

src/dune_rules/install_rules.ml
```diff
@@ -560,11 +589,22 @@ end = struct
             and* modules =
               Dir_contents.ocaml dir_contents
               >>| Ml_sources.modules ~for_:(Library name)
+            and* melange_runtime_deps =
+              match Lib_info.melange_runtime_deps info with
+              | External _paths -> assert false
```
jchavarri, line 594: I wonder if we could read from `Lib.t` instead to remove this assertion.

jchavarri (approved):
I just have a few questions. Amazing work!

src/dune_rules/install_rules.ml
anmonteiro: I'm gonna address this as part of https://github.com/ocaml/dune/pull/7234#discussion_r1130098303

src/dune_rules/install_rules.ml
anmonteiro: I believe this is related to how we represent stuff in `Lib_info`. there's other code that shares the same behavior. I don't think we can put this in `Lib.t` because it comes from the dune file, which creates a Lib_info.t anyway.

src/dune_rules/dune_package.ml
anmonteiro: Added a check for this in 885b58d8ebb772561f852ee88cef4daaba02b58a

src/dune_rules/install_rules.ml
anmonteiro: 885b58d8ebb772561f852ee88cef4daaba02b58a addresses this.

test/blackbox-tests/test-cases/melange/installed-library-with-runtime-deps.t
anmonteiro: Done in ef9926dbc

src/dune_rules/melange/melange_rules.ml
```diff
@@ -248,41 +248,85 @@ let setup_emit_cmj_rules ~sctx ~dir ~scope ~expander ~dir_contents
...
+    let raise_external_dep_error src =
+      let lib_info =
+        match for_ with
+        | `Library lib_info -> lib_info
+        | `Emit -> assert false
+      in
+      let loc =
+        match Lib_info.melange_runtime_deps lib_info with
+        | Local (loc, _) -> loc
+        | External _ -> Loc.none
```
rgrinberg, line 275: Shouldn't this be `assert  false` as well? If the runtime deps are external, the library must have been installed. Therefore, all of the paths in it cannot be external deps.

anmonteiro: Indeed. Fixed

PR #7156, opened 2023-02-22
bc-for-jsoo: use -noautolink, do not depend on stublibs

nojb:
Typically the ocaml bytecode compiler checks that all native shared objects in library dependencies are present when linking a bytecode executable `.bc`. However, in the case of `.bc-for-jsoo` this is not necessary since we will not actually execute the file with the bytecode interpreter.

This PR implements a small optimization in this case:

- passes `-noautolink` when building `.bc-for-jsoo` to disable the compiler check,
- modifies the JSOO rules so as to no longer depend on the foreign shared objects of the dependent libraries.

PR #7236, opened 2023-03-07
test: fix macOS test suite

anmonteiro:
- macOS tests are currently printing an `Abort trap: 6` message ([example run](https://github.com/ocaml/dune/actions/runs/4356406454/jobs/7614372124))
- this seems to have been introduced in https://github.com/ocaml/dune/pull/7156
- this PR keeps the same behavior but suppresses the abort output by running the relevant command inside a shell `if`


Alizter:
Why not just disable on macos? Do we expect different behaviour because of the platform?

anmonteiro:
The behavior is the same, but macOS additionally prints an "abort trap: 6" message.

Alizter (approved):

PR #7238, opened 2023-03-08
fix(memo): optimize node invalidation

rgrinberg:
Invalidating memoized dependency nodes has the following property:

1. We scan all the dependnecy nodes in order
2. Whenever we need to check for a node that has cutoff, we compute its
   value

This can make invalidation needlessly slow if recopmuting a cutoff node
is slow.

This commit modifies the algorithm to check for all non cutoff nodes
first, and only then start evaluating the nodes that require cutoff.

### DIFF
--- a/src/memo/memo.ml
+++ b/src/memo/memo.ml
@@ -465,8 +465,9 @@ module M = struct
 
     val changed_or_not :
          t
-      -> f:(Dep_node.packed -> Changed_or_not.t Fiber.t)
-      -> Changed_or_not.t Fiber.t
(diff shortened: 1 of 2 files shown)


Alizter:
Have you got some bench results?

snowleopard:
Yeah, I think this kind of extra complexity should come with some convincing benchmarks. As a start, I would add a test that demonstrates the problem qualitatively, merge it, and then show how this PR improves things.

rgrinberg:
I mentioned on slack that this fix isn't right. Forgot to close.

PR #7239, opened 2023-03-08
feat(melange): support `(select ...)` in `melange.emit`

### LINKED ISSUE #6417: melange: support `select` field for dependencies in `melange.emit`
In the PR that introduced `melange.emit` stanza, support for `select` fields on dependencies was not added: https://github.com/ocaml/dune/pull/6349#discussion_r1008712460. But it might be useful in some cases.

anmonteiro:
fixes https://github.com/ocaml/dune/issues/6417


src/dune_rules/melange/melange_rules.ml
```diff
@@ -245,6 +245,7 @@ let setup_emit_cmj_rules ~sctx ~dir ~scope ~expander ~dir_contents
         ~dialects:(Dune_project.dialects (Scope.project scope))
         ~modes:`Melange_emit )
   in
+  let* () = Buildable_rules.gen_select_rules sctx compile_info ~dir in
```
anmonteiro, line 248: @rgrinberg I thought this would be enough, but while the rules for the selected modules are added, the targets never end up being produced in the test that I wrote, and the compilation fails with:

```
+  File "main.ml", line 2, characters 16-27:
+  2 | let () = Js.log Foo.message
+                      ^^^^^^^^^^^
+  Error: Unbound module Foo
+  [1]
```

src/dune_rules/melange/melange_stanzas.ml
```diff
@@ -29,25 +29,48 @@ module Emit = struct
     let add kind name acc =
       match Lib_name.Map.find acc name with
       | None -> Lib_name.Map.set acc name kind
-      | Some _present ->
-        User_error.raise ~loc
-          [ Pp.textf "library %S is present twice" (Lib_name.to_string name) ]
+      | Some kind' -> (
+        match (kind, kind') with
```
rgrinberg, line 33: Seems like this is the same check as in dune_file.ml for normal libraries. Can we share the check?

anmonteiro: did that in [917bb1a](https://github.com/ocaml/dune/pull/7239/commits/917bb1ae8c5d88d61d9a3bdede7f27e48e512558), ended up using the whole decoder.

src/dune_rules/lib_deps.mli
```diff
@@ -0,0 +1,7 @@
+type t = Lib_dep.t list
+
+val decode :
+     allow_re_export:bool
+  -> (Lib_dep.t list, Dune_lang.Decoder.values) Dune_lang.Decoder.parser
+
+val of_pps : Lib_name.t list -> Lib_dep.t list
```
rgrinberg, line 7: Could you move this under `Lib_dep.L` or does it cause cycles?

anmonteiro: that's nicer indeed. done in [d4cb726](https://github.com/ocaml/dune/pull/7239/commits/d4cb726879888395b1e9a0b9c77d625333f12ec5)

PR #7240, opened 2023-03-08
Add commutative monoids

snowleopard:
@rgrinberg As promised.

Also cleaned up some unnecessary qualifications and type signatures.


snowleopard:
Maybe we should make the "proof" thing more principled but not sure it's worth it :)

otherlibs/stdune/src/monoid.mli
```diff
@@ -6,7 +7,6 @@ module type S = Monoid_intf.S
     operator synonym [( @ ) = combine], as well as derived functions [reduce]
     and [map_reduce]. *)
 module Make (M : Basic) : S with type t := M.t
-[@@inlined always]
```
snowleopard, line 9: I think this isn't doing anything. Please correct me if I'm wrong.

Alizter:
> Maybe we should make the "proof" thing more principled but not sure it's worth it :)

It is probably not worth it, but you can add a module of assertion tests that can be instantiated for any created monoid. The test-suite would just require applying this functor to the given monoid.

Also you could extract this interface from Coq but that is definitely not worth it IMO.

snowleopard:
@Alizter Yeah, agreed. Having `combine_is_commutative` seems sufficient for now.

snowleopard:
Though I'd love to generate some Dune code from Coq some day!

rgrinberg:
I'm not really convinced this helps with writing safer code. We don't currently ask the user to provide `combine_is_associative` for monoids even though we rely on it. Why do we special cause commutativity?

snowleopard:
> Why do we special cause commutativity?

Because we don't have [magmas](https://en.wikipedia.org/wiki/Magma_(algebra)) :)

All I want to do is to prevent `Fiber.map_reduce_seq(List)` in https://github.com/ocaml-dune/fiber/pull/14 from type-checking. And it will if we drop `combine_is_commutative` from the interface.

The caller can easily add `combine_is_commutative` to `List` if they want to, but it would be much more obvious during review that they are pointing this footgun at their feet.

I'm not super attached to `combine_is_commutative`, so I can drop it if you think it's too ugly/useless.

rgrinberg:
> All I want to do is to prevent Fiber.map_reduce_seq(List) in https://github.com/ocaml-dune/fiber/pull/14 from type-checking.

Note that this is actually something quite useful to prevent. It's often enough that you don't particularly care about the order of the list but you would like that sweet cons in `O(1)`. It would be weird for the code to falsely claim the operation is commutative to get this.

> I'm not super attached to combine_is_commutative, so I can drop it if you think it's too ugly/useless.

I feel like it just adds a little more overhead than OCaml programmers are used to. For example, something like `Non_empty_list` seems fine to me it's a simple function call to eliminate it. While with the module language, well it can be a lot more awkward, and to this day I still sometimes need to read the manual to refresh my knowledge of the module syntax.  Normally, I wouldn't don't care what we do internally, but my goal is to increase the adoption of Fiber outside of dune. So I'm not sure such a heavy construct is worth the trade-off. Tbh, if I could do everything my way, I would honestly just drop all the modules and stick to records with good old polymorphism. But at this point, I don't want the library to diverge from dune too much.

What do you do internally at janestreet for this? I see libraries like:

https://github.com/janestreet/abstract_algebra/blob/master/src/abstract_algebra.ml

And it seems like you they don't carry dummy placeholders for theorems.

snowleopard:
> It's often enough that you don't particularly care about the order of the list but you would like that sweet cons in O(1). 

We can introduce `Unordered_list : Monoid.Commutative.S` for this sort of thing when needed. Doesn't seem to hard but makes the code explicit about the intent.

> And it seems like you they don't carry dummy placeholders for theorems.

That one doesn't have morally incompatible modules that happen to have the same signature, so there is no problem to solve.

snowleopard:
> I feel like it just adds a little more overhead than OCaml programmers are used to. For example, something like `Non_empty_list` seems fine to me it's a simple function call to eliminate it.

Well, we can expose `Make_commutative` so that "it's a simple ~function~ functor call to eliminate it" :) 

I feel like we are both arguing about hypothetical problems. I'm worried about hypothetical subtle bugs due to incorrect use of `Fiber.map_reduce_seq`, and you are worried about hypothetical sad OCaml programmers that are forced to be more explicit when calling `Fiber.map_reduce_seq` (I mean, as if writing `List.equal Int.equal` wasn't bringing enough sadness already!)

Alizter (approved):

otherlibs/stdune/src/monoid.mli
rgrinberg: Perhaps it does something in your internal fork of OCaml? I'd imagine flambda pays more attention to these annotations.

rgrinberg:
> I mean, as if writing things like List.equal Int.equal every day wasn't bringing them enough sadness already

It's bringing us plenty of sadness. Now imagine if `List.equal` had the following signature:
```ocaml
module type Eq = sig
  type t
  val equal : t -> t -> bool
  type equal_is_commutative = unit
  type equal_is_transitive = unit
  type equal_is_reflexive = unit
end
val equal : (module Eq with type = 'a) -> 'a list -> bool
```
We'd be inflicting a lot more sadness on OCaml programmers.

To reiterate, my position is that libraries meant for wider adoption shouldn't be so opinionated without a very good reason. I'd argue that this is a major reason why base and other libraries from janestreet have such low adoption despite being high quality. Before we adopt any new convention in a public library, we should look around and see if there's any successful precedent. If there is, it would make the case for this pattern a lot stronger. Otherwise, we can err on the conservative side and if there's more demand for safety around this, we can always reconsider.

snowleopard:
> To reiterate, my position is that libraries meant for wider adoption shouldn't be so opinionated without a very good reason.

Wait, what? Since when is Stdune meant for wider adoption? It says:

>  "Standard library of Dune.\nThis library offers no backwards compatibility guarantees. Use at your own risk.")

in the `dune` file. Surely, a library with such a warning isn't meant for wider (if any) adoption.

Stdune is an implementation detail and we (Dune devs) can be as opinionated about it as we like.

Now to the "without a very good reason" bit: I argue that correctness *should* be our goal, and if it is our goal, we should be ready to tradeoff some convenience for having APIs that are harder to misuse by accident.

I've no idea why you keep bringing Base or any other Jane Street libraries as examples (or counterexamples?) here. Surely we can decide the fate of this PR directly, based on its merits and concrete examples of code that you plan to write. You still haven't shown me a single realistic example for which the API I proposed forces you to write the code that makes you sad. You mentioned `List` and for that I proposed `Unordered_list` as a solution, which I'm happy to add. Is that all that worries you?

rgrinberg (approved):
I would prefer if the publicly released Fiber had a self contained interface and was agnostic to the choice of standard library. Because I believe that would help the adoption of the library. So yes, I would like to avoid having to export Stdune's various sub modules in fiber.mli.

That's all a little contrary to what we'd like to do internally, which is to use stdune as much as possible.

I suppose this PR is just a change to stdune, so I agree there's no problem with it.

otherlibs/stdune/src/monoid.mli
snowleopard: Actually, I removed it because internally I'm getting an error:

> the "inlined" attribute cannot appear in this context

which presumably means we enable some of the warnings that aren't enabled by default in the Dune repo. Maybe we should enable them to clean up such needless annotations.

snowleopard:
(Quickly merging until you changed your mind! :sweat_smile:)

> I would prefer if the publicly released Fiber had a self contained interface and was agnostic to the choice of standard library.

Hmm, I see. My impression was that Fiber needed quite a bit of scaffolding from Stdune but perhaps that's not the case. 

In any case, I don't mind if you decide that the public version of `Fiber.map_reduce_seq` should just take a `Monoid`, or if you dispose of monoids altogether. 

But also, maybe we'll find a better API. In some sense, the fact that the APIs of monoids and commutative monoids currently coincide feels like an accident that may some day change, rendering the `combine_is_commutative` type unnecessary.

PR #7139, opened 2023-02-21
test: demonstrate crash in modules.ml when `(stdlib .. )` used with `(wrapped false)`

anmonteiro:

PR #7241, opened 2023-03-08
fix(stdlib): pass flags when building stdlib.ml

emillon:
Reported by @gretay-js.

This ensures that when building `stdlib.ml` (the main module of a library with `(stdlib)`), flags set in the corresponding stanza `(library)` are correctly passed.

### DIFF
--- a/src/dune_rules/compilation_context.ml
+++ b/src/dune_rules/compilation_context.ml
@@ -209,11 +209,16 @@ let create ~super_context ~scope ~expander ~obj_dir ~modules ~flags
   }
 
 let for_alias_module t alias_module =
+  let keep_flags = Modules.is_stdlib_alias (modules t) alias_module in
   let flags =
-    let project = Scope.project t.scope in
-    let dune_version = Dune_project.dune_version project in
-    let profile = (Super_context.context t.super_context).profile in
-    Ocaml_flags.default ~dune_version ~profile
+    if keep_flags then
+      (* in the case of stdlib, these flags can be written by the user *)
+      t.flags
+    else
+      let project = Scope.project t.scope in
+      let dune_version = Dune_project.dune_version project in
+      let profile = (Super_context.context t.super_context).profile in
+      Ocaml_flags.default ~dune_version ~profile
   in
   let sandbox =
     let ctx = Super_context.context t.super_context in
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@ Unreleased
 - Bootstrap: correctly detect the number of processors by allowing `nproc` to be
   looked up in `$PATH` (#7272, @Alizter)
 
+- Pass correct flags when compiling `stdlib.ml`. (#7241, @emillon)
+
 - Speed up file copying on macos by using `clonefile` when available
   (@rgrinberg, #7210)
 
--- a/test/blackbox-tests/test-cases/stdlib/stdlib-flags.t
+++ b/test/blackbox-tests/test-cases/stdlib/stdlib-flags.t
@@ -1,0 +1,18 @@
+  $ cat > dune-project << EOF
+  > (lang dune 3.7)
+  > (using experimental_building_ocaml_compiler_with_dune 0.1)
+  > EOF
+
+  $ cat > dune << EOF
+  > (library
+  >  (name mystdlib)
+  >  (stdlib)
+  >  (flags :standard -w -8))
+  > EOF
+
+  $ cat > mystdlib.ml << EOF
+  > (* This triggers warning 8 *)
+  > let None = None
+  > EOF
+
+  $ dune build


emillon:
I'm not sure how to add that to the test suite. To do that locally, I added `(ocamlopt_flags :standard -alert +sample-alert)` to the test in `stdlib-compilation.t` and checked that it's passed in the log, but that's not great.
Also @rgrinberg I'm not too sure how related this fix is to the comment just above so maybe there's a better fix. Thanks!

anmonteiro:
@emillon I recently added some stdlib tests in https://github.com/ocaml/dune/pull/7139, perhaps that could be helpful.

src/dune_rules/compilation_context.ml
```diff
@@ -208,12 +208,15 @@ let create ~super_context ~scope ~expander ~obj_dir ~modules ~flags
...
-let for_alias_module t alias_module =
+let for_alias_module t ?flags alias_module =
   let flags =
-    let project = Scope.project t.scope in
-    let dune_version = Dune_project.dune_version project in
-    let profile = (Super_context.context t.super_context).profile in
-    Ocaml_flags.default ~dune_version ~profile
+    match flags with
+    | Some f -> f
+    | None ->
```
anmonteiro, line 215:
```suggestion
    match with_flags with
    | true -> t.flags
    | false ->
```

src/dune_rules/compilation_context.ml
```diff
@@ -208,12 +208,15 @@ let create ~super_context ~scope ~expander ~obj_dir ~modules ~flags
   ; loc
   }
 
-let for_alias_module t alias_module =
+let for_alias_module t ?flags alias_module =
```
anmonteiro, line 211:
```suggestion
let for_alias_module t ?(with_flags=false) alias_module =
```

anmonteiro: I think it'd be better if this didn't take flags from the outside, since the only use case is to use the flags from the compilation context itself.

src/dune_rules/module_compilation.ml
```diff
@@ -486,7 +486,8 @@ let build_all cctx =
             if Modules.is_stdlib_alias modules m then
               (* XXX it would probably be simpler if the flags were just for this
                  module in the definition of the stanza *)
-              Compilation_context.for_alias_module cctx m
+              let flags = Compilation_context.flags cctx in
+              Compilation_context.for_alias_module cctx m ~flags
```
anmonteiro, line 490:
```suggestion
              Compilation_context.for_alias_module cctx m ~with_flags:true
```

rgrinberg, line 490: Seems weird to pass a flag to Compilation_context that's coming from the Compilation_context

rgrinberg (changes requested):

emillon:
Thanks. I added a test and implemented your suggestions. I didn't realize it was coming from the same value.

src/dune_rules/compilation_context.ml
```diff
@@ -208,12 +208,14 @@ let create ~super_context ~scope ~expander ~obj_dir ~modules ~flags
   ; loc
   }
 
-let for_alias_module t alias_module =
+let for_alias_module t ?(keep_flags = false) alias_module =
```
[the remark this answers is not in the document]
emillon: done

rgrinberg (approved):

src/dune_rules/compilation_context.ml
```diff
@@ -209,11 +209,14 @@ let create ~super_context ~scope ~expander ~obj_dir ~modules ~flags
   }
 
 let for_alias_module t alias_module =
+  let keep_flags = Modules.is_stdlib_alias (modules t) alias_module in
   let flags =
-    let project = Scope.project t.scope in
-    let dune_version = Dune_project.dune_version project in
-    let profile = (Super_context.context t.super_context).profile in
-    Ocaml_flags.default ~dune_version ~profile
+    if keep_flags then t.flags
```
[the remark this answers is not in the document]
emillon: what about the following comment - is this related?

https://github.com/ocaml/dune/blob/35e3958215e0fed860aecb24e977a46448950103/src/dune_rules/module_compilation.ml#L486-L490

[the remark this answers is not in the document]
rgrinberg: Yes, it's related but it's not quite the same. I think the main distinction is that:

1. For normal libraries, the alias module is generated. Therefore, there's no need to let the user control the flags on it. We know exactly how it should be compiled.
2. For the stdlib, the alias module is written by the user. Therefore, we need to give the user some control to modify the flags if necessary.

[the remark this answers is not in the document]
emillon: Thanks.

PR #7247, opened 2023-03-08
test(coq): duplicate theory in a dune file

Alizter:
Appears to be bugged since the directory targets are being reported first.


test/blackbox-tests/test-cases/coq/duplicate-theory.t/run.t
```diff
@@ -0,0 +1,17 @@
...
+  > 
+  > (coq.theory
+  >  (name foo))
+  > EOF
+
+BUG The directory target is found before the theory
+
+  $ dune build
+  Error: the following both define the directory target: _build/default/foo.tex
+  - dune:1
```
ejgallego, line 15: That's because laziness in rule creation and coqdoc not forcing the Coq lib db to be built.

Alizter: I'm not sure. coqdoc rules are generated when we generate the rules for the theory. The issue here is that we have to pre-declare directory targets which due to lazyness means we have identical dir targets.

ejgallego: I guess we can reorder and get a better error message? Or just force the Coq.Lib.db in the coqdoc rules.

Alizter: Again it's not the coqdoc rules, its the declaration of the directory target. The coqdoc rules already force the database.

ejgallego: Ahh, I see what you mean. Not sure what can be done for it.

PR #7249, opened 2023-03-08
test(coq): duplicate theory in a workspace

Alizter:
Appears to be bugged since Dune does not complain at all.

### DIFF
--- a/test/blackbox-tests/test-cases/coq/duplicate-theory-workspace.t/run.t
+++ b/test/blackbox-tests/test-cases/coq/duplicate-theory-workspace.t/run.t
@@ -1,0 +1,38 @@
+Duplicate theories in the same project should be caught by Dune:
+
+  $ cat > A/dune << EOF
+  > (coq.theory
+  >  (name foo))
+  > EOF
+
+  $ cat > B/dune << EOF
+  > (coq.theory
+  >  (name foo))
+  > EOF
+
+BUG this is not caught
+
+  $ dune build
+
+  $ ls _build/default/A _build/default/B
+  _build/default/A:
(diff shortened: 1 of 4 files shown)


test/blackbox-tests/test-cases/coq/duplicate-theory-workspace.t/run.t
```diff
@@ -0,0 +1,15 @@
...
+  > (coq.theory
+  >  (name foo))
+  > EOF
+
+  $ cat > B/dune << EOF
+  > (coq.theory
+  >  (name foo))
+  > EOF
+
+BUG this is not caught
```
ejgallego, line 13: I guess we are lazier now, as the theories are not public, `dune build @install` won't complain.

Alizter: Oh yes let me check again.

Alizter: As you can see now, they are both being built.

ejgallego: I guess that's OK as they are private theories, right?

Alizter: Oh yes, I guess the package layer restricts this from being an issue.

ejgallego: Yes, they will never make it to the workspace theory db as they are not public.

PR #7250, opened 2023-03-08
coq: improve directory target message

Alizter:
Since this is in gen_rules @rgrinberg could you take a look.

### DIFF
--- a/src/dune_rules/gen_rules.ml
+++ b/src/dune_rules/gen_rules.ml
@@ -303,12 +303,11 @@ let collect_directory_targets ~init ~dir =
         | Coq_stanza.Theory.T m ->
           Coq_rules.coqdoc_directory_targets ~dir m
           |> Path.Build.Map.union acc ~f:(fun path loc1 loc2 ->
-                 User_error.raise
+                 User_error.raise ~loc:loc1
                    [ Pp.textf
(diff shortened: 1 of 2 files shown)


src/dune_rules/gen_rules.ml
```diff
@@ -305,10 +305,10 @@ let collect_directory_targets ~init ~dir =
           |> Path.Build.Map.union acc ~f:(fun path loc1 loc2 ->
                  User_error.raise
```
rgrinberg, line 306: You could add `~loc:loc1` or `~loc:loc2` here

rgrinberg (approved):

Alizter:
Hmm actually I'm not sure including the location there made it easier. I will revert.

rgrinberg:
You mean the `~loc` arg? Without it, the editor won't be able to jump to the error

Alizter:
Oh yes, you are correct. I've update the test that was affected and cleared up the wording (since anybody can create this dir target and not just a theory stanza).

Alizter:
Do you think it matters to choose `loc1` or `loc2`. I'm thinking the latter might be more correct since it was found "later"?

rgrinberg:
I don't think it matters much. In RPC, you can provide both locations anyway

PR #7251, opened 2023-03-08
test(memo): demonstrate loss of concurrency

rgrinberg:
Reproduces the loss of concurrency observed in #5549 in a unit test


snowleopard (approved):
Awesome. Thank you!

rgrinberg:
Note that I wasn't able to reproduce the issue without cutoff for some reason. I'm still investigating why is that the case. Any hints would be welcome if you know off the top of your head.

snowleopard:
That's because without a cutoff, we just invalidate the caller immediately. There is no need to attempt to recompute the value: we know that everything down the line need to be recomputed, and then we do that when we reach the top-level node (or the first node with a cutoff).

Issue #7177, opened 2023-02-24
dune formats in new lines when flags in pps, breaking later


davesnx:
**Describe the bug**

dune formats `(pps ppx1 -- --flag true)` in new lines when the line is too long and in the next dune build it breaks since interprets `--flag` as a ppx.

## Expected Behavior

dune should format into something that allows dune to build/execute.

Worth mentioned a closed issue slightly related to where the proposed syntax might solve elegantly the issue https://github.com/ocaml/dune/issues/2573.

## Actual Behavior

```
(* dune *)

(library
 (name content_explorer)
 (libraries
  bs_css
  bs_css_emotion
  reason_react
  webapi)
 (preprocess
  (pps
   bs-react-intl-ppx.lib
   bs-emotion-ppx.lib
   styled-ppx.lib -- --compat-with-bs-emotion-ppx true)))
```

```
$ dune fmt
File "dune", line 1, characters 0-0:
diff --git a/_build/default/dune b/_build/default/.formatted/dune
index 133c0ed..ad9a112 100644
--- a/_build/default/dune
+++ b/_build/default/.formatted/dune
@@ -1,12 +1,11 @@
 (library
  (name content_explorer)
- (libraries
-  bs_css
-  bs_css_emotion
-  reason_react
-  webapi)
+ (libraries bs_css bs_css_emotion reason_react webapi)
  (preprocess
   (pps
    bs-react-intl-ppx.lib
    bs-emotion-ppx.lib
-   styled-ppx.lib -- --compat-with-bs-emotion-ppx true)))
\ No newline at end of file
+   styled-ppx.lib
+   --
+   --compat-with-bs-emotion-ppx
+   true)))
Promoting _build/default/.formatted/dune to dune.
``` 

If I run `dune build` it doesn't apply arguments into the ppx and fails to build since treats `--compat-with-bs-emotion-ppx` as a ppx. 

## Specifications

- Version of `dune` (output of `dune --version`): 3.7.0
- Version of `ocaml` (output of `ocamlc --version`): 4.14

nojb:
Sorry, but this feels more like a problem with the way we interpret whitespace (is there a reason why we don't ignore it altogether?), rather than with the formatter; do you agree, or am I missing something?

emillon:
I agree that it's a problem with the parser, not the formatter. I'll have a look at this.

emillon:
I had a look and wrote something to check that the ASTs are similar, but it couldn't find a difference.
[...]

PR #7253, opened 2023-03-09
Add reproduction for formatting bug when pp is long

davesnx:
https://github.com/ocaml/dune/issues/7177

### DIFF
--- a/test/blackbox-tests/test-cases/github7177.t
+++ b/test/blackbox-tests/test-cases/github7177.t
@@ -1,0 +1,40 @@
+dune formats (pps ppx1 -- --flag true) in new lines when the line is too long
+and it breaks on build
+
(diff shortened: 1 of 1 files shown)


emillon:
Thanks for the PR, but how can I reproduce the issue? As far as I can see, the output represents the same AST as the input (I have tried with a check that the roundtrip is correct), so it should be interpreted identically.
Newlines are not significant in dune files (except inside string litterals etc), so this formatting is the expected outcome.
Can you tell me more about your setup? Any chance `esy` or another tool interpreting the dune file or something like that?

davesnx:
I can't repo either. I swear I got the error, maybe in dune 2.9 this wasn't the case?

Alizter:
@davesnx What OS are you using?

PR #7255, opened 2023-03-09
benchmarks: add synthetic watch

jchavarri:
Related to https://github.com/ocaml/dune/issues/5549.

Adds a synthetic generator that produces a project where the parallelism degradation discussed in that issue is reproducible.

It also adds a benchmark that builds upon the generator above, so that we can measure rebuild times in dune watch over time.

The synthetic project consists on a couple of libraries with a dependency between them, with a few hundred modules each, where there are cross dependencies between the modules. One thing to note is that the parallelism degradation can only be reproduced when there are `mli` files. Without them rebuilds are almost instant.


bench/gen_synthetic_dune_watch.ml
```diff
@@ -0,0 +1,105 @@
...
+let write_directory base_dir dir_row dir_col =
+  let dir_cols = 10 in
+  let mod_rows = 10 in
+  let mod_cols = 10 in
+
+  let dirname = sprintf "%s/dir_%d_%d" base_dir dir_row dir_col in
+  Unix.mkdir dirname 0o777;
+
+  for row = 1 to mod_rows do
+    for col = 1 to mod_cols do
```
rgrinberg, line 18: Do we need so many directories?

jchavarri: I thought 2k modules represents mid-large projects well enough. Do you have a number of folder / modules that would be better?

rgrinberg: I feel like the best benchmarks contain the minimum required to clearly exhibit the performance issue. So the smaller the project, the better.

jchavarri: I tried with 200 modules (instead of 2000) and the build finishes in a few seconds, so the benchmark becomes useless as it is impossible to see anything. I can try with 1000 modules, but I don't see how that changes anything meaningfully.

I agree it's desirable to reduce the complexity of the benchmark to the minimum. But the issue we are discussing is more visible in mid-large projects, when small performance issues like the concurrency one here compound over time. I would rather err on the "too many modules" side in this case.

jchavarri: Reduced the number of modules / dirs, from 2k / 20 to 800 / 8: https://github.com/ocaml/dune/pull/7255/commits/7cfbefa45ee474f3d99dfaba16e95b3e2f16561b.

The timings are now 32.95 ([with `main`](https://github.com/jchavarri/dune/actions/runs/4378936899/jobs/7664300071#step:8:44)) vs 21.98s (with [the patch](https://github.com/jchavarri/dune/actions/runs/4378964590/jobs/7664358253#step:8:44))

rgrinberg: Sure, I don't mean to say we only need a few modules. If 800 is enough to demonstrate the issue, then let's use that.

Are separate directories essential for this issue? I imagine that probably not. If so, we should simplify the directory structure.

jchavarri: Sounds good. Removed dirs in https://github.com/ocaml/dune/pull/7255/commits/0a81a6f553b759cacf3de88aafaaa7a177f8ff4a.

snowleopard:
This is great, thank you! I've always wanted Dune to have some synthetic benchmarks.

PR #7256, opened 2023-03-09
chore: fix melange in the main default devShell

anmonteiro:


flake.nix
```diff
@@ -49,7 +49,18 @@
       };
       pkgs = nixpkgs.legacyPackages.${system}.appendOverlays [
         (self: super: {
-          ocamlPackages = self.ocaml-ng.ocamlPackages_4_14;
+          ocamlPackages = self.ocaml-ng.ocamlPackages_4_14.overrideScope' (oself: osuper: {
```
rgrinberg, line 52: Shouldn't this overlay come from the melange flake somehow? It seems weird that dune has to know about this

anmonteiro: Ah, no, this is because we're using the melange overlay (to avoid compiling an extra ocaml 4.14 from my overlays).

rgrinberg: It's not possible for the melange overlay to include this?

rgrinberg: Sorry, I don't know much about nix.

anmonteiro: I think the alternative is:

- switch back to `melange.outputs.packages.${system}.default`
  - 👍  gives us the correct dune version, tracked in the melange flake
  - 👎 requires compiling OCaml+flambda from the ocaml overlays

anmonteiro: your alternative should also work: override dune_3 in the melange overlay. I can switch to that approach if you prefer.

Alizter (approved):
Works for me

PR #7257, opened 2023-03-09
feature: optimize file copying on linux

rgrinberg:
Use sendfile to copy files on Linux.

### DIFF
--- a/otherlibs/stdune/src/copyfile_stubs.c
+++ b/otherlibs/stdune/src/copyfile_stubs.c
@@ -50,7 +50,54 @@ CAMLprim value stdune_copyfile(value v_from, value v_to) {
   CAMLreturn(Val_unit);
 }
 
+CAMLprim value stdune_sendfile(value v_in, value v_out, value v_size) {
+  caml_failwith("sendfile: linux");
+}
+
+#elif __linux__
+
+#include <caml/threads.h>
+#include <caml/unixsupport.h>
+
+#include <sys/sendfile.h>
+#include <unistd.h>
+
+#define FD_val(value) Int_val(value)
+
+CAMLprim value stdune_copyfile(value v_from, value v_to) {
+  caml_failwith("copyfile: only on macos");
+}
+
+static int dune_sendfile(int in, int out, int length) {
+  int ret;
+  while (length > 0) {
+    ret = sendfile(out, in, NULL, length);
+    if (ret < 0) {
+      return ret;
+    }
+    length = length - ret;
+  }
+  return length;
+}
+
+CAMLprim value stdune_sendfile(value v_in, value v_out, value v_size) {
+  CAMLparam3(v_in, v_out, v_size);
+  caml_release_runtime_system();
+  /* TODO Use copy_file_range once we have a good mechanism to test for its
+   * existence */
+  int ret = dune_sendfile(FD_val(v_in), FD_val(v_out), Int_val(v_size));
+  caml_acquire_runtime_system();
+  if (ret < 0) {
+    uerror("sendfile", Nothing);
+  }
+  CAMLreturn(Val_unit);
+}
+
 #else
+
+CAMLprim value stdune_sendfile(value v_in, value v_out, value v_size) {
+  caml_failwith("sendfile: linux");
+}
 
 CAMLprim value stdune_copyfile(value v_from, value v_to) {
   caml_failwith("copyfile: only on macos");
--- a/otherlibs/stdune/src/io.ml
+++ b/otherlibs/stdune/src/io.ml
@@ -72,6 +72,126 @@ let copy_channels =
   in
   loop
 
+let setup_copy ?(chmod = Fun.id) ~src ~dst () =
+  let ic = open_in src in
+  let oc =
+    try
+      let perm = (Unix.fstat (Unix.descr_of_in_channel ic)).st_perm |> chmod in
+      Stdlib.open_out_gen
+        [ Open_wronly; Open_creat; Open_trunc; Open_binary ]
+        perm dst
+    with exn ->
+      close_in ic;
+      Exn.reraise exn
+  in
+  (ic, oc)
+
+module Copyfile = struct
+  (* Bindings to mac's fast copy function. It's similar to a hardlink, except
+     it does COW when edited. It will also default back to regular copying if
+     it fails for w/e reason *)
+  external copyfile : string -> string -> unit = "stdune_copyfile"
+
+  external sendfile : src:Unix.file_descr -> dst:Unix.file_descr -> int -> unit
+    = "stdune_sendfile"
+
+  let available =
+    match Platform.OS.value with
+    | Darwin -> `Copyfile
+    | Linux -> `Sendfile
+    | Windows | Other -> `Nothing
+
+  let sendfile =
+    let setup_copy ?(chmod = Fun.id) ~src ~dst () =
+      match Unix.openfile src [ O_RDONLY ] 0 with
+      | exception Unix.Unix_error (Unix.ENOENT, _, _) -> Error `Src_missing
+      | fd_src -> (
+        match Unix.fstat fd_src with
+        | exception exn ->
+          Unix.close fd_src;
+          Error (`Exn (Exn_with_backtrace.capture exn))
+        | src_stat -> (
+          match src_stat.st_kind with
+          | S_DIR -> Error `Src_is_a_dir
+          | _ ->
+            let open Result.O in
+            let+ fd_dst, src_size =
+              match
+                let dst_perm = chmod src_stat.st_perm in
+                Unix.openfile dst [ O_WRONLY; O_CREAT; O_TRUNC ] dst_perm
+              with
+              | fd_dst -> Ok (fd_dst, src_stat.st_size)
+              | exception exn -> (
+                Unix.close fd_src;
+                match exn with
+                | Unix.Unix_error (Unix.EISDIR, _, _) -> Error `Dst_is_a_dir
+                | _ -> Error (`Exn (Exn_with_backtrace.capture exn)))
+            in
+            (fd_src, fd_dst, src_size)))
+    in
+    fun ?chmod ~src ~dst () ->
+      (* All of this exception translation is done for now to maintain the
+         same error messages as the other file copying functions.
+
+         Eventually, we should stop using exceptions for signalling these
+         errors. But that's a bit of a large change since there's a lot of
+         exception catching to audit. *)
+      match setup_copy ?chmod ~src ~dst () with
+      | Error (`Exn exn) -> Exn_with_backtrace.reraise exn
+      | Error `Src_is_a_dir -> raise (Sys_error "Is a directory")
+      | Error `Dst_is_a_dir ->
+        let message = Printf.sprintf "%s: Is a directory" dst in
+        raise (Sys_error message)
+      | Error `Src_missing ->
+        let message = Printf.sprintf "%s: No such file or directory" src in
+        raise (Sys_error message)
+      | Ok (fd_src, fd_dst, src_size) ->
+        Exn.protectx (fd_src, fd_dst, src_size)
+          ~f:(fun (src, dst, src_size) -> sendfile ~src ~dst src_size)
+          ~finally:(fun (src, dst, _) ->
+            Unix.close src;
+            Unix.close dst)
+
+  let copyfile ?chmod ~src ~dst () =
+    let src_stats =
+      match Unix.stat src with
+      | exception Unix.Unix_error (Unix.ENOENT, _, _) ->
+        let message = Printf.sprintf "%s: No such file or directory" src in
+        raise (Sys_error message)
+      | { st_kind = S_DIR; _ } -> raise (Sys_error "Is a directory")
+      | stats -> stats
+    in
+    (try copyfile src dst with
+    | Unix.Unix_error (Unix.EPERM, "unlink", _) ->
+      let message = Printf.sprintf "%s: Is a directory" dst in
+      raise (Sys_error message)
+    | Unix.Unix_error (Unix.ENOENT, "realpath", _) ->
+      let message = Printf.sprintf "%s: No such file or directory" src in
+      raise (Sys_error message));
+    match chmod with
+    | None -> ()
+    | Some chmod -> src_stats.st_perm |> chmod |> Unix.chmod dst
+
+  let copy_file_portable ?chmod ~src ~dst () =
+    Exn.protectx (setup_copy ?chmod ~src ~dst ()) ~finally:close_both
+      ~f:(fun (ic, oc) -> copy_channels ic oc)
+
+  let copy_file_best =
+    match available with
+    | `Sendfile -> sendfile
+    | `Copyfile -> copyfile
+    | `Nothing -> copy_file_portable
+
+  let copy_file_impl = ref `Best
+
+  let copy_file ?chmod ~src ~dst () =
+    match !copy_file_impl with
+    | `Portable -> copy_file_portable ?chmod ~src ~dst ()
+    | `Best -> copy_file_best ?chmod ~src ~dst ()
+end
+
+let set_copy_impl m = Copyfile.copy_file_impl := m
+
 module Make (Path : sig
   type t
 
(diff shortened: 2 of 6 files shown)


rgrinberg:
Here's some benchmarks. 1K source files, around 50K bytes (normal source file size). I ran the benchmarks 3 times and picked the best result.

Before the PR:

```
time _build/default/bench/micro/copyfile.exe ../

real    0m1.410s
user    0m0.176s
sys     0m1.210s
```

After the PR:

```
time _build/default/bench/micro/copyfile.exe ../

real    0m1.106s
user    0m0.053s
sys     0m1.031s
```

Around a 20% increase.

otherlibs/stdune/src/copyfile_stubs.c
```diff
@@ -50,13 +50,66 @@ CAMLprim value stdune_copyfile(value v_from, value v_to) {
...
+
+#include <sys/sendfile.h>
+
+#define FD_val(value) Int_val(value)
+
+CAMLprim value stdune_copyfile(value v_from, value v_to) {
+  caml_failwith("copyfile: only on macos");
+}
+
+CAMLprim value stdune_sendfile(value v_out, value v_in, value v_size) {
```
anmonteiro, line 80: in the interest of readability, I'd change the argument order here to (src, dst) rather than (dst, src).

otherlibs/stdune/src/io.ml
```diff
@@ -254,32 +254,91 @@ struct
...
+          raise (Sys_error message)
+        | Error `Src_missing ->
+          let message =
+            Printf.sprintf "error: %s: No such file or directory"
+              (Path.to_string src)
+          in
+          raise (Sys_error message)
+        | Ok (ic, oc, size) ->
+          Exn.protectx (ic, oc, size)
+            ~f:(fun (i, o, size) -> sendfile o i size)
```
anmonteiro, line 311: re: readability (https://github.com/ocaml/dune/pull/7257/files?w=1#r1131466187) it felt weird to get `i,o,size` and send `o i size`.

anmonteiro (approved):

snowleopard:
Nice! I'm looking forward to trying this internally.

Do you think it's worth having a way to opt-in/opt-out of these faster copying implementations? I'm thinking that `#elif __linux__` firing might not necessarily mean `sendfile` is available/not-buggy on a particular system. And on Mac OS, copy-on-write might give observably different results (which I suppose might trip up some weird rules).

rgrinberg:
> Do you think it's worth having a way to opt-in/opt-out of these faster copying implementations? I'm thinking that #elif __linux__ firing might not necessarily mean sendfile is available/not-buggy on a particular system.

`sendfile` is quite old (around 20 years now?), so I would hope the Linux folks ironed out the bugs by now. What about a runtime toggle instead? Since I'm pretty confident that the compilation isn't going to fail given how ancient this call is.

I didn't introduce `copy_file_range` even though it's even more promising, precisely because it's a bit of a pain to test things for availability at compile time (it would need to be done at bootstrap as well).

> And on Mac OS, copy-on-write might give observably different results (which I suppose might trip up some weird rules).

On macos, the runtime toggle would only be useful for benchmarking IMO. The underlying function is what's used by all regular mac application to do file copies (which is why copying a file in "finder" is faster than using `cp` as I was surprised to discover!). Moreover, the call will also fallback to regular copying when it detects a COW isn't possible.

I was going to propose this in a separate issue, but I feel like with this implementation of copy, the copy mode for sandboxing is the best default on macos.

snowleopard:
> What about a runtime toggle instead?

Ah, I think that's what I meant as well: just a regular Dune configuration setting. That could accept a ʼcopy_file_range` variant too.

> I was going to propose this in a separate issue, but I feel like with this implementation of copy, the copy mode for sandboxing is the best default on macos.

Indeed, I'd love to use that for sandboxing too.

Alizter (approved):

snowleopard:
> Do you want the setting in this PR or separately?

I'm weakly in favour of not merging this change until there is a way to turn it off, but as you say, the risk seems small, so adding configurability in a separate PR seems OK too.

otherlibs/stdune/src/io.ml
```diff
@@ -254,32 +254,92 @@ struct
...
+    external linux : unit -> bool = "stdune_is_linux"
+
+    let available =
+      if darwin () then `Copyfile else if linux () then `Sendfile else `Nothing
+
+    let sendfile =
+      let setup_copy ?(chmod = Fun.id) ~src ~dst () =
+        match Unix.openfile (Path.to_string src) [ O_RDONLY ] 0 with
+        | exception Unix.Unix_error (Unix.ENOENT, _, _) -> Error `Src_missing
+        | ic ->
```
snowleopard, line 272: Why do you call this `ic`? Isn't this a file descriptor? `src_fd` seems like a clearer name. Same for `oc`.

otherlibs/stdune/src/io.ml
```diff
@@ -254,32 +254,92 @@ struct
...
+          let open Result.O in
+          let+ size, oc =
```
snowleopard, line 274: Why do we care about the `size` of `dst`, and isn't it supposed to be 0 since we create it?

rgrinberg: It's the size of `src` actually. We need it because `sendfile` needs to know how many bytes to write.

otherlibs/stdune/src/io.ml
rgrinberg: Will rename

otherlibs/stdune/src/io.ml
snowleopard: Oh, sorry, the naming got me confused. Of course, we need to know the `src`'s size.

snowleopard: (Maybe call it `src_size`.)

rgrinberg:
@snowleopard made things easier to read and added a runtime variable to disable using sendfile (or clonefile)

otherlibs/stdune/src/io.ml
```diff
@@ -254,42 +254,114 @@ struct
...
-          raise (Sys_error message)
-        | Unix.Unix_error (Unix.ENOENT, "realpath", _) ->
-          let message =
-            Printf.sprintf "error: %s: No such file or directory" src
-          in
-          raise (Sys_error message));
-        match chmod with
-        | None -> ()
-        | Some chmod -> src_stats.st_perm |> chmod |> Unix.chmod dst)
+    match Stdlib.Sys.getenv_opt "DUNE_COPY_FILE_IMPL" with
```
snowleopard, line 356: Why not use a standard Dune configuration setting, which can be set via `dune-workspace` if needed? Sometimes it's convenient to use a variable, but sometimes not.

snowleopard:
> @snowleopard made things easier to read 

Thanks, much better! (Though the code might still benefit from splitting into smaller functions.)

> and added a runtime variable to disable using sendfile (or clonefile)

As I mentioned in a comment, why not use `Dune_config`? I think we should use the same configuration mechanism for everything.

rgrinberg:
> As I mentioned in a comment, why not use Dune_config? I think we should use the same configuration mechanism for everything.

A couple of reasons why I chose to use an env var:

1. It would be quite annoying to plumb through this configuration option through the rules, the engine, etc.
2. Ultimately, I don't think this configuration option is useful. The only use I envision is benchmarking and perhaps ironing out some early bugs. I'd rather not bloat the config file and manual with implementation details that aren't all that relevant

snowleopard:
> 1. It would be quite annoying to plumb through this configuration option through the rules, the engine, etc.

I think plumbing annoyance is a wrong reason for not doing right things. If we want to make certain aspects of Dune's behaviour configurable, we should aim to do that in a uniform way. Cutting corners can surely make some things easier in the short term but good practices matter in the long term. And I argue that uniform configurability is good practice.

> 2. Ultimately, I don't think this configuration option is useful. The only use I envision is benchmarking and perhaps ironing out some early bugs.

OK, so let's assume we do hit an early bug. How do I fix builds for all internal users of Dune? I can't change their `.bashrc` to set this variable. The current mechanism just doesn't work for "ironing out some early bugs".

> I'd rather not bloat the config file and manual with implementation details that aren't all that relevant

We should make it easy for us to add configuration settings and deprecate them when they are no longer needed, so we don't need to make this sort of compromises.

Do you agree in principle that uniform configurability is good practice in the long term?

rgrinberg:
I guess I thought of this to more of a debugging option for developers of dune rather than a full blown configuration option. Sure, I agree that configuring things should be done in a single way, I just don't like too much configuration in general. How much is too much is subjective, but in this case I don't see a trade-off for the user to choose between. Looking at our other configuration options in dune-config, this one would stand out as one that would probably make users scratch their heads.

> OK, so let's assume we do hit an early bug. How do I fix builds for all internal users of Dune? I can't change their .bashrc to set this variable. The current mechanism just doesn't work for "ironing out some early bugs".

I would do the same thing that is done for all dune bugs. Revert the upgrade and wait until a fix. I suppose that this workflow could be disturbing for you internally. But then I still don't see how disabling this via dune-config would be easier than just toggling a flag in the code. Both options would have to go through VCS to propagate to all the devs. I'm making assumptions here, so it would help to know why is dune-config the preferred way to make such a change.

snowleopard:
> I would do the same thing that is done for all dune bugs. Revert the upgrade and wait until a fix.

The problem is that there is no "revert" for us. In a monorepo, we have a file at the root of the monorepo that essentially links to a particular Dune binary that can be used to build this revision of the monorepo. We never remove old versions of the build system because that would mean that old revisions become unbuildable. So, once a binary is deployed, it's deployed ~forever.

What do we do if we hit a bug? We deploy a new version of Dune binary, with that bug fixed, and update the monorepo to point to the new version. Then people can rebase to the new revision to fix their builds but this is quite a disruptive process, i.e. everybody needs to rebase to pick up the fix. Even though it's disruptive and slow, it's still feasible. On the other hand, setting an environment variable across all machines, users and automated jobs, is just infeasible in practice.

I guess we could also have a file in our monorepo that lists various environment variable settings that Dune should be given when run. That might work, I suppose, but that would be a relatively big change.

> But then I still don't see how disabling this via dune-config would be easier than just toggling a flag in the code.

My thinking was: while the new file copying mechanism is experimental, I can just disable it in our `dune-workspace` file for everybody but enable it for my own builds and various benchmarking builds (because I control the corresponding `.bashrc` files). Then, after we've used it for some time with no issues, I'll just toggle the configuration option.

An alternative would be to change the default value for the environment variable in the `janestreet` branch: set it to `portable` to preserve the old behaviour. I suppose we could think about doing something like this in the long term if we'd like to avoid plumbling all settings through `Dune_config`: say, we could use `ENABLE_EXPERIMENTAL_DUNE_FEATURES=true` to pick up some new experimental defaults for some of the settings.

(I still think using a uniform configuration mechanism is better but maybe it's just too difficult.)

rgrinberg:
I see. Let me just clarify that my main objection with this setting in dune-config is not that it's a bit difficult to thread through, but that it would have to stay in dune forever (or until we decide to break things). If for example we decide to yank this feature, we would not be able to remove the configuration option at the same time.

I guess I could add it to dune-config and mark it as unstable so that users are aware.

snowleopard:
Right, so maybe we can adopt the following approach for such cases:

* Let's assume we are adding some experimental `feature`. We plan to eventually make it enabled by default for everybody.
* We add a new `experimental_feature` option to `Dune_config`.
* Initially, users get the `(experimental_feature disabled)` behaviour by default. At this point it's easy to opt in.
* After some time, we can decide to toggle the default to `(experimental_feature enabled)`. At this point, it's easy to opt out.
* After some more time, we are happy to garbage-collect this setting and just enable the feature for everybody. We remove the `experimental_feature` option from `Dune_config`. At this point, whoever opted in or out, will need to fix their configuration. Hopefully, the number of such users is small.

Does this sound reasonable? If yes, I can move this text somewhere to our coding conventions.

rgrinberg:
Yeah, that sounds fine. A couple of other issues come to mind:

1. We need to split `dune_config`. It currently contains a lot of dependencies, so it can be hard to use it in some places. Instead, it could be two separate parts. The part that lists the various features and configuration options, and the part that populates them using our frontend.

2. How do we change what the defaults are between `main` and the `janestreet` branch? Just have an additional commit that reverts whenever the public defaults change? Or you can just maintain the defaults file separately and ignore ours?

snowleopard:
Yeah, splitting `dune_config` seems fine.

> How do we change what the defaults are between main and the janestreet branch? Just have an additional commit that reverts whenever the public defaults change? Or you can just maintain the defaults file separately and ignore ours?

I think if there is a way to opt in/out via `dune-workspace` then we can just do that, i.e. there is no need to have different defaults between the branches.

rgrinberg:
Okay, we can now turn off the optimization through our universal configuration.

@jchavarri you might want to test this PR as well. To make sure there are no surprises.

snowleopard:
It would be lovely to benchmark this PR on @gridbugs's monorepo benchmark.

jchavarri:
> @jchavarri you might want to test this PR as well. To make sure there are no surprises.

I tested this branch against `main` with a couple of large builds and could not see any meaningful difference, not sure if the new copying mechanism is kicking in the device I am using.

Alizter:
@jchavarri do you notice it if you set cache to copy mode and enable it?

rgrinberg:
> I tested this branch against main with a couple of large builds and could not see any meaningful difference, not sure if the new copying mechanism is kicking in the device I am using.

That's fine. As long as there's no issues or regressions. I would never expect this to move the needle for a large build.

snowleopard:
> I would never expect this to move the needle for a large build.

This sounds a little sad. If we basically expect no performance benefit from this in practice, why maintain the extra complexity?

snowleopard:
This all sounds reasonable, thanks for adding a bit more colour!

rgrinberg:
I should mention that I'm quite hopeful for [copy_file_range](https://lwn.net/Articles/659523/) to be usable eventually. Which will give us even better performance and will be the best mechanism for sandboxing/caching when available.

snowleopard:
> I should mention that I'm quite hopeful for [copy_file_range](https://lwn.net/Articles/659523/) to be usable eventually. Which will give us even better performance and will be the best mechanism for sandboxing/caching when available.

Just for clarity: could you write down the concrete system call (with arguments) you're thinking of using eventually? I presume you're not interested in the "range" functionality of `copy_file_range`, but rather in that it calls out to `reflink` if possible -- is that right? If yes, why not use `reflink` directly?

rgrinberg:
> I presume you're not interested in the "range" functionality of copy_file_range, but rather in that it calls out to reflink if possible -- is that right? If yes, why not use reflink directly?

I would hope that even if reflinks aren't supported, the file system might be able to speed up the copying somehow. With `copy_file_range`, we don't need to worry about reflink support and we get automatic fallback for free.

> If yes, why not use reflink directly?

Do you mean using [ioctl](https://man7.org/linux/man-pages/man2/ioctl_ficlonerange.2.html)? 

It would be an option, but as I mentioned, we would need to write the fallback ourselves

snowleopard:
I see, that all sounds good!

otherlibs/stdune/src/io.ml
```diff
@@ -72,6 +72,131 @@ let copy_channels =
...
+
+  let copy_file_portable ?chmod ~src ~dst () =
+    Exn.protectx (setup_copy ?chmod ~src ~dst ()) ~finally:close_both
+      ~f:(fun (ic, oc) -> copy_channels ic oc)
+
+  let copy_file ?chmod ~src ~dst () = !copy_file_impl ?chmod ~src ~dst ()
+end
+
+let set_copy_impl m =
+  Copyfile.copy_file_impl :=
```
snowleopard, line 195: Not that it matters much but it seems faster to `match` on every call instead of following the pointer in the `ref`: the former benefits from branch prediction, so it's essentially free, while the latter always requires an indirection. Am I missing something?

src/dune_config/config.mli
```diff
@@ -27,6 +27,9 @@ val global_lock : Toggle.t t
     reduces concurrency *)
 val cutoffs_that_reduce_concurrency_in_watch_mode : Toggle.t t
 
+(** whether dune should optimize file copying on Linux/MacOS *)
+val copy_file : [ `Portable | `Fast ] t
```
snowleopard, line 31: I wonder if "fast" is a good choice in the long term. I can imagine that in future we can come up with several faster than portable implementations, neither of which could be clearly preferable in all circumstances. In that world, it would be awkward to choose between "fast" and something else. Choosing between "sendfile" and something else seems cleaner.

Not a very strong opinion but it's going to be hard to change the name in future, so we might as well discuss it now.

snowleopard (approved):
Thanks for adding configurability! I left a couple of non-blocking comments.

src/dune_config/config.mli
rgrinberg: `Sendfile` is a bit of a misnomer because this option controls the optimization on macos as well (which doesn't use sendfile). I suppose we could call the option `Sendfile_or_clonefile`, but that would be a bit ugly as well. I guess I can choose `Best` for now.

If we need more granular options, we can always add them. It's only removing existing options that would be painful.

snowleopard: > `Sendfile` is a bit of a misnomer because this option controls the optimization on macos as well (which doesn't use sendfile).

You could allow both `Sendfile` and `Clonefile` I suppose. This could even be a list of preferences rather than a single value, so the user could say something like "try Clonefile first, and then fallback to portable if Clonefile isn't supported".

snowleopard: Qualitative descriptions like "fast" or "best" seem nice but they often hide important details that make it hard to diagnose problems.

PR #7258, opened 2023-03-09
chore(nix): remove the opam2nix dependency on the default package

anmonteiro:
I followed the derivation here: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/ocaml/dune/3.nix

Inspired by the changes in https://github.com/ocaml/ocaml-lsp/pull/1040

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -88,7 +88,24 @@
           });
     in
     {
-      packages.default = scope.dune;
+      packages = {
+        dune = scope.dune;
+        default = with pkgs; stdenv.mkDerivation rec {
+          pname = package;
+          version = "n/a";
+          src = ./.;
+          nativeBuildInputs = with ocamlPackages; [ ocaml findlib ];
+          buildInputs = lib.optionals stdenv.isDarwin [
+            darwin.apple_sdk.frameworks.CoreServices
+          ];
+          strictDeps = true;
+          buildFlags = [ "release" ];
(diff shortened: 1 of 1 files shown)


anmonteiro:
To try this out: 

```
$ nix run github:anmonteiro/dune/anmonteiro/nix-package-default# -- --version
```

flake.nix
```diff
@@ -88,7 +88,24 @@
...
     in
     {
-      packages.default = scope.dune;
+      packages = {
+        dune = scope.dune;
+        default = with pkgs; stdenv.mkDerivation rec {
+          pname = package;
+          version = "n/a";
+          src = ./.;
+          nativeBuildInputs = with ocamlPackages; [ ocaml findlib ];
```
rgrinberg, line 97: Why do we need findlib? We don't use any OCaml libraries. Or is it for the installation step?

anmonteiro: Yup, findlib has a `setupHook` ([source](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/ocaml/findlib/default.nix#L30)) that sets the `OCAMLFIND_DESTDIR` variable.

Alizter: Btw if you ever need to use opam inside this flake, be aware that `OCAMLFIND_DESTDIR` will break package builds in opam so make sure to unset it.

Issue #7017, opened 2023-02-07
melange: introduce `melange.root` stanza, to avoid duplicated .js artifacts


jchavarri:
At the moment, every `melange.emit` stanza generates its own version of the js artifacts produced by the libs it consumes. It there are N `melange.emit` stanzas using the same library `foo`, the `js` artifacts for `foo` will be found N times under the `_build` folder.

Besides unnecessary disk space usage, this is problematic for build tools like webpack, that become slower as they have to process the same files multiple times. It is also punitive for the resulting JavaScript bundle size, as it is not possible out of the box  to reuse the same library js bundle in multiple places, unless one goes going through complex webpack configurations.

To avoid this, @rgrinberg suggested the idea of introducing a new stanza `melange.root`. This new stanza would define a target, and the `javascript_extension` and `module_system` used by that target, e.g.:

```
(melange.root
 (target tools_es6)
 (module_system es6)
 (javascript_extension js))
```

Then, afaiu, `melange.emit` stanzas that are spread across the project folders would be used to define "entry points", referencing the target already defined in `melange.root`, e.g.
```
(melange.emit
 (target tools_es6)
 (alias melange)
 (libraries lib))
```

The fields `module_system` and `javascript_extension` would be removed from `melange.emit`.

Some open questions:
- Should `melange.root` be in `dune` or `dune-project` file?
- Is it possible to solve the duplication problem without adding a new stanza?
- If `melange.root` was ultimately added, would it make sense to rethink stanza naming? e.g. `melange.emit` could be rather `melange.entry_point`  or `melange.executable`, as it would not be really emitting anything anymore. And `melange.root` could be `melange.emit` instead.

jchavarri:
A couple more open questions:
- Would `target` still be strictly needed in `melange.emit`? It is nice to reduce the amount of js targets to what the user strictly needs (e.g. tests use commonjs, source code uses es6), but the alternative (produce js targets for all the `melange.root` stanzas found) is also possible.
- I am not sure I get how rules production would work with `melange.root`. If user calls `dune build @melange` in the example shared in description, dune would find the alias in `melange.emit`, then start looking from the `melange.emit` folder up the tree, until it finds a `melange.root` stanza, then produce the rules in that folder, right? I am not sure about the implications with `copy_files` and `include_subdirs` 🤔

rgrinberg:
> Should melange.root be in dune or dune-project file?

[...]

PR #6489, opened 2022-11-16
feature: Add coqffi stanza

Alizter:
This continues from #6248 but is left as draft as more work needs to be done to support the wish in https://github.com/ocaml/dune/pull/6248#issuecomment-1315828441.

Close #3995

PR #7261, opened 2023-03-10
melange: add test for inter-module deps

jchavarri:
It seems that when generating rules for `.js` targets, we don't generate rules for the modules that these `.js` files depend upon.

This PR adds a test that shows the failure.


jchavarri:
@rgrinberg I tried fixing this but to get the inter-module dependencies one has to call `Compilation_context.dep_graphs`, which requires the compilation context, but `setup_emit_js_rules` doesn't receive it. I am trying to figure out if there's a way to pass it through in `gen_rules`, but I'm not sure if I'm looking at the problem from the wrong side.

Alizter:
@jchavarri I had a similar issue trying to inspect the module deps in https://github.com/ocaml/dune/pull/6489. I'm not sure Compilation_context.dep_graphs really needs the entire context but I didn't investigate further.

anmonteiro:
Your test has empty entries so they don’t get compiled.

jchavarri:
> Your test has empty entries so they don’t get compiled.

@anmonteiro I don't think that matters. The `melange.emit` can have no modules and still depend in the library `a` (which would have both modules as `modules`).

But it was possible to simplify the test, did so in https://github.com/ocaml/dune/pull/7261/commits/0f0d46c6cc88e4f7b6c41837813f43345cd59501.

jchavarri:
> I still think this is on purpose, they should be produced if you add the alias

@rgrinberg suggested the idea of targeting a specific `.js` file to be able to build subsets of some Melange project without having to add multiple `melange.emit` stanzas or introduce new ones like discussed in #7017. So I was testing this and realized it won't work as the dep rules are defined right now.

rgrinberg:
This PR demonstrates expected behavior. When you do `dune build x`, dune will only build `x` and its dependencies. Other things such as runtime dependencies are not expected to build. This is why we introduced the alias to the melange.emit stanza

jchavarri:
@rgrinberg Right. Maybe I misinterpreted you when you said (on DM):

> I’m also not sure that multiple emit stanzas is the solution. I wonder if it would be enough to be able to specify the subset of targets that are related to a library inside a melange.emit

Weren't you referring to targeting an individual `js` file?

jchavarri:
> Other things such as runtime dependencies are not expected to build.

I am not sure I get the semantics of what runtime deps means. To me, dependencies between resulting js files generated by melange are just regular deps. In this case, the behavior might be correct but it does not look consistent: if dune builds the dependencies of a cmj when targeting it, why doesn't it build the dependencies of a js target?

rgrinberg:
> Weren't you referring to targeting an individual js file?

My idea was to have multiple aliases. A separate alias for every app.

> I am not sure I get the semantics of what runtime deps means. To me, dependencies between resulting js files generated by melange are just regular deps. In this case, the behavior might be correct but it does not look consistent: if dune builds the dependencies of a cmj when targeting it, why doesn't it build the dependencies of a js target?

You're right, runtime deps isn't a well defined term. Dune has the nice property of only building the minimum to satisfy a build request. So for dune, dependencies are things that are needed to generate the action, or to execute it. Here, the other .js files you've pointed out fall into neither category. So they will not be built

jchavarri:
> My idea was to have multiple aliases. A separate alias for every app.

But this means multiple `melange.emit` stanzas right? Or is there a way to tie a specific subset of libraries under an alias using a single `melange.emit` that I'm missing?

rgrinberg:
> My idea was to have multiple aliases. A separate alias for every app.

But this means multiple melange.emit stanzas right? Or is there a way to tie a specific subset of libraries under an alias using a single melange.emit that I'm missing?

We can make it work with a single emit stanza. There's currently no mechanism to do this, but we could add one.

jchavarri:
> We can make it work with a single emit stanza. There's currently no mechanism to do this, but we could add one.

Something like this?

```
(melange.emit
 (alias all_apps)
 (target es6)
 (libraries
  (library
   (name app1)
   (alias app1))
  (library
   (name app2)
   (alias app2))
  (library
   (name app3)
   (alias app3)))
 (module_systems es6))
```

Then, users can build everything with `dune build @all_apps` or specific subsets with `dune build @appX`.

rgrinberg:
Looks fine to me. Although I would lean towards a new field instead:

```
(alias
 (name app1)
 (libraries app1)
(alias
 (name app2
 (libraries app2)
...
```

jchavarri:
@rgrinberg I added a ticket to define the changes required: https://github.com/ocaml/dune/issues/7339

I think this PR is still valuable as it reflects current behavior when targeting `.js` files. Please lmk if anything is missing to get it merged, thanks!

Issue #6831, opened 2023-01-05
dune doc


EmileTrotignon:
This would be a new command `dune doc` that would do the following :

- Call `dune build @doc`
- Open the doc index in the user's browser.

Do you think this is a good idea ? 
If maintainers like the feature, I will start working on it soon.

Alizter:
Can you not open the index in the browser and just refresh everytime you do a `dune build @doc`?

esope:
Maybe some issue that is related is: once I have built de doc using `dune build @doc`, where is the generated documentation?
Providing some easy, quick access to the documentation (through a link to the right target, or by printing out a message that shows the path of the generated documentation) could already be helpful?

Alizter:
Have a look in `_build/default/_doc`. This is where `odoc` will but all the doc artifacts. The reference manual even mentions the location of the documentation: https://dune.readthedocs.io/en/stable/documentation.html

Alizter:
`dune build fmt` is really the same as `dune build @fmt --auto-promote`.

Alizter:
We are currently in the process of designing a new terminal UI for Dune, to allow for a sleeker user experience. One thing that might be useful to add is the ability to view newly build targets. So a user could do the following:

1. Have Dune running the TUI in watch mode
2. Ask Dune to build the `@doc` alias.
3. Inspect (the locations of) all the newly built targets.

EmileTrotignon:
> Have a look in `_build/default/_doc`. This is where `odoc` will but all the doc artifacts. The reference manual even mentions the location of the documentation: https://dune.readthedocs.io/en/stable/documentation.html

I am sure it is explained in the documentation, but the documentation is hard to navigate, and even if it wasn't, beginners will not have read it back to back (and if they did, will not remember this detail). The existing workflow for someone who doesn't know this by heart is :
- Call `dune build @doc`
- Open your browser, then the dune doc, then find the right page.
- Go back to your terminal to open the index, 

I propose to have the following workflow : 
- Call `dune doc`.

@esope 's proposal is the following :
- Call `dune build @doc`
- Click / copy the link from your terminal.

To me, this seems a bit worst, but also a more minor change, so it may be the best choice.

> > Have a look in `_build/default/_doc`. This is where `odoc` will but all the doc artifacts. The reference manual even mentions the location of the documentation: https://dune.readthedocs.io/en/stable/documentation.html

[...]

PR #7262, opened 2023-03-10
feature: introduce [$ dune ocaml doc]

EmileTrotignon:
This is a feature from issue #6831

It adds a new command "dune ocaml doc" that builds the doc, and then open its index in a browser. If no browser can be found, it print the location of the index.

There are no tests for this feature, as I do not know of a way to automatically test it.

I also only tested this on my machine, and this does very different stuff on different systems, so I would like if someone with a windows or an osx machine would be able to test it.


Alizter:
Can you setup a cram test showing the feature. It should be possible to mock a browser for xdg-open in the cram test.

EmileTrotignon:
@Alizter I have not managed to mock a browser for xdg-open, so I justed mock xdg-open itself.

boot/libs.ml
Alizter, line 1: You probably want to revert this change. This happens automatically I think on macos.

test/blackbox-tests/test-cases/doc-browser.t/run.t
```diff
@@ -0,0 +1,12 @@
+This tests shows how to use the `dune ocaml doc` command to open the
+documentation index to a browser.
+  $ ls
+  dune
+  dune-project
+  foo.ml
+  spoof.desktop
+  xdg-open
```
Alizter, line 8: You can remove this as this is just noise.

test/blackbox-tests/test-cases/doc-browser.t/run.t
```diff
@@ -0,0 +1,12 @@
...
+  $ PATH=.:$PATH 
+  $ export PATH
```
Alizter, line 10:
```suggestion
  $ export PATH=.:$PATH 
```

test/blackbox-tests/test-cases/doc-browser.t/run.t
```diff
@@ -0,0 +1,12 @@
...
+  $ dune ocaml doc | grep -c "file:///.*/_build/.sandbox/.*/default/test/blackbox-tests/test-cases/doc-browser.t/_build/default/_doc/_html/index.html"
```
Alizter, line 11:
```suggestion
  $ dune ocaml doc | sed -e 's/.*file:\/\/\([^ ]*\).*/\1/'
  $TESTCASE_ROOT/_build/default/_doc/_html/index.html
```
Here is a better way to scrub the output. cram tests usually detect and scrub the current directory but because we start with `file:///` it gets confused. Therefore we can manually remove it with sed.

src/dune_rules/odoc.mli
```diff
@@ -2,6 +2,10 @@
 
 open Import
 
+module Paths : sig
+  val toplevel_index : Context.t -> Path.Build.t
```
Alizter, line 6: Is this supposed to be the location of the index file? If so could you add a comment detailing it as such?

test/blackbox-tests/test-cases/doc-browser.t/dune-project
```diff
@@ -0,0 +1,4 @@
+(lang dune 3.6)
```
Alizter, line 1: We can use lang 3.8 for this.

bin/build_cmd.ml
Alizter, line 1: It's a shame that GitHub messes up the syntax highlighting for OCaml here due to the `{| |}` comments. I have reported the issue upstream, but it is unlikely to be fixed anytime soon.

bin/build_cmd.ml
```diff
@@ -215,3 +215,63 @@ let fmt =
     run_build_command ~common ~config ~request
   in
   Cmd.v (Cmd.info "fmt" ~doc ~man ~envs:Common.envs) term
+
+let doc =
+  let doc = "Build the documentation of an OCaml project" in
```
Alizter, line 220:
```suggestion
  let doc = "Build and view the documentation of an OCaml project" in
```

bin/build_cmd.ml
```diff
@@ -215,3 +215,63 @@ let fmt =
...
+  let man =
+    [ `S "DESCRIPTION"
+    ; `P
+        {|$(b,dune ocaml doc) build and open the documention of an ocaml project.|}
```
Alizter, line 224:
```suggestion
        {|$(b,dune ocaml doc) builds and then opens the documention of an OCaml project in the users default browser.|}
```
Or something similar

bin/build_cmd.ml
```diff
@@ -215,3 +215,63 @@ let fmt =
...
+        , (* First element of argv is the name of the command. *)
+          (cmd_name :: args) @ [ url ] )
+      in
+      match cmd with
+      | Some (cmd, args) ->
+        Proc.restore_cwd_and_execve
+          (Path.to_absolute_filename cmd)
+          args ~env:Env.initial
+      | None ->
+        Printf.printf "Docs built. Index can be found here: %s\n"
```
Alizter, line 271: Rather than `Printf.printf` use `Dune_console.print`. It might also be worth printing this message in both cases, whether or not the command was found. In fact, it might be clearer to raise a user error / warning here and say that a web browser command or whatever could not be found / used so that it doesn't mysteriously fail.

Alizter (commented):
I know this is still a draft, but some initial comments and suggestions.

bin/build_cmd.ml
```diff
@@ -215,3 +215,70 @@ let fmt =
...
+        [ Pp.textf "Docs built. Index can be found here: %s\n"
+            absolute_toplevel_index_path
+        ];
+
+      let url = Printf.sprintf {|file://%s|} absolute_toplevel_index_path in
+      let cmd =
+        let open Option.O in
+        let path = Env_path.path Env.initial in
+        let* cmd_name, args =
+          match Ocaml_config.system doc_ctx.ocaml_config with
```
rgrinberg, line 259: We now have a `Platform` module that you can use

bin/build_cmd.ml
```diff
@@ -215,3 +215,70 @@ let fmt =
...
+      let toplevel_index_path = Dune_rules.Odoc.Paths.toplevel_index doc_ctx in
+      let absolute_toplevel_index_path =
+        Path.(toplevel_index_path |> build |> to_absolute_filename)
+      in
+      Dune_console.print
+        [ Pp.textf "Docs built. Index can be found here: %s\n"
+            absolute_toplevel_index_path
+        ];
+
+      let url = Printf.sprintf {|file://%s|} absolute_toplevel_index_path in
```
rgrinberg, line 254: Any reason why you don't use normal strings?

EmileTrotignon: I guess I was worried of having to escape the `//`, but that's not needed, I'll change that.

bin/build_cmd.ml
EmileTrotignon: Oh that's very nice, thanks !

EmileTrotignon:
I have finally tested this on a mac, and it works ! I still have to test it on windows.
I think mac and linux support is enough to get this out of draft though.

bin/build_cmd.mli
```diff
@@ -12,3 +12,5 @@ val runtest_term : unit Term.t
 val build : unit Cmd.t
 
 val fmt : unit Cmd.t
+
+val doc : unit Cmd.t
```
Alizter, line 16: I don't think this is the correct place for this. Rather than putting it here, could you put it in a file called doc.ml and place that file in bin/ocaml next to all the other ocaml subcommands?

bin/build_cmd.ml
```diff
@@ -213,3 +213,70 @@ let fmt =
...
+      in
+      match cmd with
+      | Some (cmd, args) ->
+        Proc.restore_cwd_and_execve
+          (Path.to_absolute_filename cmd)
+          args ~env:Env.initial
+      | None ->
+        Dune_console.print
+          [ Pp.text
+              "No browser could be found, you will have to open the \
```
Alizter, line 276: Please use `Console.print` instead.

bin/build_cmd.ml
```diff
@@ -213,3 +213,70 @@ let fmt =
...
+      let toplevel_index_path = Dune_rules.Odoc.Paths.toplevel_index doc_ctx in
+      let absolute_toplevel_index_path =
+        Path.(toplevel_index_path |> build |> to_absolute_filename)
+      in
+      Dune_console.print
+        [ Pp.textf "Docs built. Index can be found here: %s\n"
+            absolute_toplevel_index_path
+        ];
+
+      let url = Printf.sprintf "file://%s" absolute_toplevel_index_path in
```
Alizter, line 252: String concatenation is perhaps simpler.

EmileTrotignon:
@Alizter The last commit takes your comment into account.

test/blackbox-tests/test-cases/doc-browser.t/run.t
```diff
@@ -0,0 +1,8 @@
+This tests shows how to use the `dune ocaml doc` command to open the
+documentation index to a browser.
+  $ export PATH=.:$PATH 
+  $ dune ocaml doc | sed -e 's/.*file:\/\/\([^ ]*\).*/\1/'
```
Alizter, line 4: You can use a different deliminator rather than `/` for sed so you don't have to escape so many characters. Something like # should work.

test/blackbox-tests/test-cases/doc-browser.t/open
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+echo $@
```
Alizter, line 2: Adding something like
```
echo "Web browser received arg:"
```
might make the test easier to read. Same for the other one.

Better yet have a single copy and create symlinks so we don't have to duplicate.

test/blackbox-tests/test-cases/doc-browser.t/dune-project
```diff
@@ -0,0 +1,4 @@
+(lang dune 3.7)
```
Alizter, line 1: Let's use 3.8 here.

bin/ocaml/doc.ml
```diff
@@ -0,0 +1,70 @@
...
+    match cmd with
+    | Some (cmd, args) ->
+      Proc.restore_cwd_and_execve
+        (Path.to_absolute_filename cmd)
+        args ~env:Env.initial
+    | None ->
+      Console.print
+        [ Pp.text
+            "No browser could be found, you will have to open the \
+             documentation yourself.\n"
```
Alizter, line 65:
```suggestion
             documentation yourself."
```

bin/ocaml/doc.ml
```diff
@@ -0,0 +1,70 @@
...
+      , (* First element of argv is the name of the command. *)
+        (cmd_name :: args) @ [ url ] )
+    in
+    match cmd with
+    | Some (cmd, args) ->
+      Proc.restore_cwd_and_execve
+        (Path.to_absolute_filename cmd)
+        args ~env:Env.initial
+    | None ->
+      Console.print
```
Alizter, line 62: If this is meant to be a warning, then you can use User_warning.emit instead.

bin/ocaml/doc.ml
```diff
@@ -0,0 +1,70 @@
...
+    let url = "file://" ^ absolute_toplevel_index_path in
+    let cmd =
+      let open Option.O in
+      let path = Env_path.path Env.initial in
+      let* cmd_name, args =
+        match Platform.OS.value with
+        | Platform.OS.Darwin -> Some ("open", [ "-u" ])
+        | Platform.OS.Linux -> Some ("xdg-open", [])
+        | Platform.OS.Windows -> Some ("start", [])
+        | Platform.OS.Other -> None
```
Alizter, line 49:
```suggestion
        match (Platform.OS.value : Platform.OS.t) with
        | Darwin -> Some ("open", [ "-u" ])
        | Linux -> Some ("xdg-open", [])
        | Windows -> Some ("start", [])
        | Other -> None
```

Alizter (commented):
Just a few more comments.

EmileTrotignon:
Hey @Alizter 
I have pushed a commit that takes your comment into account

test/blackbox-tests/test-cases/doc-browser.t/dune-project
Alizter: This is now 3.10

bin/ocaml/doc.ml
```diff
@@ -0,0 +1,70 @@
...
+
+let term =
+  let+ common = Common.term in
+  let config = Common.init common in
+  let request (setup : Import.Main.build_system) =
+    let dir = Path.(relative root) (Common.prefix_target common ".") in
+    let open Action_builder.O in
+    let+ () =
+      Alias.in_dir
+        ~name:(Dune_engine.Alias.Name.of_string "doc")
```
emillon, line 23: Can you add that as `Dune_engine.Alias.doc`? (and change the of_string "doc" in that file to use it)

test/blackbox-tests/test-cases/doc-browser.t/dune-project
emillon: It doesn't matter. If the test works with 3.7 and is not about a 3.7-versioned feature it's fine as is.

bin/ocaml/doc.ml
```diff
@@ -0,0 +1,70 @@
...
+
+    let url = "file://" ^ absolute_toplevel_index_path in
+    let cmd =
+      let open Option.O in
+      let path = Env_path.path Env.initial in
+      let* cmd_name, args =
+        match (Platform.OS.value : Platform.OS.t) with
+        | Darwin -> Some ("open", [ "-u" ])
+        | Linux -> Some ("xdg-open", [])
+        | Windows -> None
```
emillon, line 48: what does `webbrowser` do here?

bin/ocaml/doc.ml
```diff
@@ -0,0 +1,70 @@
...
+    let toplevel_index_path = Dune_rules.Odoc.Paths.toplevel_index doc_ctx in
+    let absolute_toplevel_index_path =
+      Path.(toplevel_index_path |> build |> to_absolute_filename)
+    in
+    Dune_console.print
+      [ Pp.textf "Docs built. Index can be found here: %s\n"
+          absolute_toplevel_index_path
+      ];
+
+    let url = "file://" ^ absolute_toplevel_index_path in
```
emillon, line 40: does `absolute_toplevel_index_path` need to be urlencoded here?

bin/ocaml/doc.ml
```diff
@@ -0,0 +1,70 @@
+open Stdune
+open Import
+
+let doc = "Build and view the documentation of an OCaml project"
+
+let man =
+  [ `S "DESCRIPTION"
+  ; `P
+      {|$(b,dune ocaml doc) builds and then opens the documention of an OCaml project in the users default browser.|}
```
emillon, line 9:
```suggestion
      {|$(b,dune ocaml doc) builds and then opens the documentation of an OCaml project in the user's default browser.|}
```

bin/ocaml/doc.ml
EmileTrotignon: Something similar, although it uses something very weird to get reload instead of a new tab each time.
The relevant code is here : https://github.com/dbuenzli/webbrowser/blob/master/src/webbrowser.ml

bin/ocaml/doc.ml
```diff
@@ -0,0 +1,70 @@
...
+    in
+    match cmd with
+    | Some (cmd, args) ->
+      Proc.restore_cwd_and_execve
+        (Path.to_absolute_filename cmd)
+        args ~env:Env.initial
+    | None ->
+      User_warning.emit
+        [ Pp.text
+            "No browser could be found, you will have to open the \
```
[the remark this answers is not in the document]
EmileTrotignon: I It is printed in both cases

bin/ocaml/doc.ml
EmileTrotignon: I have tested a path with a space on my machine (linux with firefox), and it works fine as is.

bin/ocaml/doc.ml
EmileTrotignon: There is already an Alias.doc variable, but it has the wrong type.

emillon: This has been moved since then, you can find it in the `Alias0` module.

emillon: You don't need to change the `Alias` module. This identifier already exists, it is in `Alias0` module https://github.com/ocaml/dune/blob/main/src/dune_rules/alias0.ml

bin/ocaml/doc.ml
```diff
@@ -44,7 +44,7 @@ let term =
         | Darwin -> Some ("open", [ "-u" ])
         | Linux -> Some ("xdg-open", [])
         | Windows -> None
-        | Other -> None
+        | Other | FreeBSD | NetBSD | OpenBSD -> None
```
Alizter, line 47: FTR BSD platforms can have open or xdg-open but it's probably not worth adding support for it in this PR.

PR #7263, opened 2023-03-10
fix(coq): fix duplicate dir targets being detected before theories

Alizter:


ejgallego (approved):
Nice !

rgrinberg:
This PR needs to be reverted because it's loading the coq scope for projects where coq isn't enabled. This is a performance penalty for all dune projects which is a no-no.

ejgallego:
This is strange, how is the code called if there is not `coq.theory` stanza in your development?

Code after #9149 seems to do the same, right?

ejgallego:
Never mind, got to see the other PRs

PR #7264, opened 2023-03-10
chore: pin but not install melange

rgrinberg:
So that we install melange along with the rest of the deps at once

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,8 +74,8 @@ dev-depext:
 
 .PHONY: melange
 melange:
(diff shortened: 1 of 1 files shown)


Alizter (approved):

anmonteiro (approved):

jchavarri (approved):

Alizter:
The Coq error is unrelated, I pushed a test then a fix in quick succession however it seems the output of the test wasn't quite correct at the time.

PR #7267, opened 2023-03-10
fix(melange): copy, don't symlink, runtime_deps

anmonteiro:
Bundlers will follow symlinks, and if a runtime_dep is a `.js` file that points to another file in source, that reference can now be broken. 

We want the relative requires to be relative to the melange target directory rather than anything in source.

~Incidentally, this ends up fixing the directory targets test.~ this was an instance of https://github.com/ocaml/dune/pull/7270

### DIFF
--- a/src/dune_rules/melange/melange_rules.ml
+++ b/src/dune_rules/melange/melange_rules.ml
@@ -287,7 +287,7 @@ let setup_runtime_assets_rules sctx ~dir ~target_dir ~mode
     let loc = mel.loc in
     Memo.parallel_iter copy ~f:(fun (src, dst) ->
         Super_context.add_rule ~loc ~dir ~mode sctx
-          (Action_builder.symlink ~src ~dst))
+          (Action_builder.copy ~src ~dst))
   and+ () =
     match mel.alias with
     | None -> Memo.return ()
--- a/test/blackbox-tests/test-cases/melange/emit-with-runtime-deps-dir-target.t
+++ b/test/blackbox-tests/test-cases/melange/emit-with-runtime-deps-dir-target.t
@@ -33,11 +33,7 @@ Rules created for the assets in the output directory
   $ dune build @mel --display=short
           melc .output.mobjs/melange/melange__Main.{cmi,cmj,cmt}
           melc output/main.js
-  File "dune", line 5, characters 0-71:
-  5 | (melange.emit
-  6 |  (alias mel)
-  7 |  (target output)
-  8 |  (runtime_deps ./some_dir))
-  Error: Error trying to read targets after a rule was run:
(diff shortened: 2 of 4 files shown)


src/dune_rules/melange/melange_rules.ml
```diff
@@ -287,7 +287,7 @@ let setup_runtime_assets_rules sctx ~dir ~target_dir ~mode
     let loc = mel.loc in
     Memo.parallel_iter copy ~f:(fun (src, dst) ->
         Super_context.add_rule ~loc ~dir ~mode sctx
-          (Action_builder.symlink ~src ~dst))
+          (Action_builder.copy ~src ~dst))
```
jchavarri, line 290: This solves the problem with JS bundlers. But I am thinking now there are cases where symlinking could be more desirable (e.g. imagine having some melange library with its own `node_modules` folder that includes a few js-only 3rd party libraries that the melange library binds against).

Could we expose the way that runtime_deps copies are created? e.g. `runtime_deps (link ./foo.js)`.

Or maybe what the feature i'm bringing up should be a new feature altogether? 🤔

jchavarri (approved):
Thanks!

src/dune_rules/melange/melange_rules.ml
anmonteiro: Why is symlinking more desirable in that case? If it's just for performance, I think we should wait for concrete reports of slow builds.

rgrinberg:
Tbh, this seems like a problem with the bundlers. But if this is what's needed, then so be it.

src/dune_rules/melange/melange_rules.ml
jchavarri: Yeah, I was referring to performance. But I agree about testing it first in real world.

anmonteiro:
The macOS test failure will be addressed by #7270.

jchavarri:
> Tbh, this seems like a problem with the bundlers. But if this is what's needed, then so be it.

After reading Rudi's comment, I wondered if there could be an option to configure this in Webpack. Turns out there is an option [resolve.symlinks](https://webpack.js.org/configuration/resolve/#resolvesymlinks), and it's enabled by default. Please let me some time to explore if setting it to `false` works, before moving forward with this approach.

anmonteiro:
Even in that scenario you might want it turned on (e.g. for yarn links). Melange should still copy in that case

jchavarri:
Don't you want the same behavior for `yarn link`? (set webpack `resolve.symlink` to `false`). Otherwise, Webpack would be resolving the dependencies of the linked package on its original location, and not in your project.

In any case, disabling `resolve.symlink` seems to break HMR and package managers that rely on symlinks like pnpm, so not sure it's the best way forward.

PR #7210, opened 2023-03-02
feature: use clonefile on macos

rgrinberg:
Use mac's [clonefile] instead of manually copying. [clonefile] is like
hardlink but it will copy-on-write when edited.

The attached benchmark before the optimization:

```
real    0m3.593s
user    0m0.249s
sys     0m2.849s
```

And after:

```
real    0m2.044s
user    0m0.083s
sys     0m1.954s
```

So about a 45% reduction. Not bad. The benchmark also includes how long it took to delete the temp directory at the end (Around 10k additional unlinks), so 45% is likely pessimistic.

PR #7270, opened 2023-03-10
test: demonstrate `Io.copy_file` difference between macOS / other OS

anmonteiro:
It looks like `clonefile` (#7210) will succeed if we ask it to create a directory.

This is best looked at without whitespace: https://github.com/ocaml/dune/pull/7270/files?w=1

### DIFF
--- a/otherlibs/stdune/src/io.ml
+++ b/otherlibs/stdune/src/io.ml
@@ -268,6 +268,18 @@ struct
       fun ?chmod ~src ~dst () ->
         let src = Path.to_string src in
         let dst = Path.to_string dst in
+        let src_stats =
+          match Unix.stat src with
+          | exception Unix.Unix_error (Unix.ENOENT, _, _) ->
+            let message =
+              Printf.sprintf "error: %s: No such file or directory" src
+            in
+            raise (Sys_error message)
+          | stats -> stats
+        in
+        (match src_stats.st_kind with
+        | S_DIR -> raise (Sys_error "Is a directory")
+        | _ -> ());
         (try Copyfile.copyfile src dst with
         | Unix.Unix_error (Unix.EPERM, "unlink", _) ->
           let message = Printf.sprintf "%s: Is a directory" dst in
@@ -279,7 +291,7 @@ struct
           raise (Sys_error message));
         match chmod with
         | None -> ()
-        | Some chmod -> (Unix.stat src).st_perm |> chmod |> Unix.chmod dst)
+        | Some chmod -> src_stats.st_perm |> chmod |> Unix.chmod dst)
 
   let file_line path n =
     with_file_in ~binary:false path ~f:(fun ic ->
--- a/otherlibs/stdune/test/io_tests.ml
+++ b/otherlibs/stdune/test/io_tests.ml
@@ -32,3 +32,14 @@ let%expect_test "copy file chmod" =
   [%expect {|
     foobarbaz
     permissions: 428 |}]
+
+let%expect_test "copy file on a directory" =
+  let dir = temp_dir () in
+  let src = Path.relative dir "initial" in
+  let dst = Path.relative dir "final" in
+  Unix.mkdir (Path.to_string src) 0o755;
+  Io.copy_file ~src ~dst ();
+  let dst = Path.to_string dst in
+  Format.printf "exists: %B; is directory: %B @." (Sys.file_exists dst)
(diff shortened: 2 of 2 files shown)


rgrinberg:
Thanks. We are doing `stat` on src, so it should be easy to fail if it returns a directory. Do you mind doing that?

anmonteiro:
[cae8bbf](https://github.com/ocaml/dune/pull/7270/commits/cae8bbfb915b8ba1ddec06771cd1662198510942) checks whether `src` is a directory and fails in that case.

otherlibs/stdune/src/io.ml
```diff
@@ -268,18 +268,24 @@ struct
       fun ?chmod ~src ~dst () ->
         let src = Path.to_string src in
         let dst = Path.to_string dst in
-        (try Copyfile.copyfile src dst with
-        | Unix.Unix_error (Unix.EPERM, "unlink", _) ->
-          let message = Printf.sprintf "%s: Is a directory" dst in
+        let src_stats = Unix.stat src in
+        match src_stats.st_kind with
+        | S_DIR ->
+          let message = Printf.sprintf "Is a directory" in
```
rgrinberg, line 274: You don't need the `sprintf` here or the `message` variable

otherlibs/stdune/src/io.ml
```diff
@@ -268,18 +268,22 @@ struct
...
-        | Unix.Unix_error (Unix.ENOENT, "realpath", _) ->
-          let message =
-            Printf.sprintf "error: %s: No such file or directory" src
-          in
-          raise (Sys_error message));
-        match chmod with
-        | None -> ()
-        | Some chmod -> (Unix.stat src).st_perm |> chmod |> Unix.chmod dst)
+        let src_stats = Unix.stat src in
+        match src_stats.st_kind with
```
rgrinberg, line 272: The extra nesting here can be eliminated:

```
let () =
  match src_stats.st_kind with
  | S_DIR -> ...
  |  _ -> ()
in
...
```

otherlibs/stdune/src/io.ml
```diff
@@ -268,6 +268,10 @@ struct
       fun ?chmod ~src ~dst () ->
         let src = Path.to_string src in
         let dst = Path.to_string dst in
+        let src_stats = Unix.stat src in
+        (match src_stats.st_kind with
+        | S_DIR -> raise (Sys_error "Is a directory")
```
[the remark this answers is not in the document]
anmonteiro: Tried something in [42fdb7a](https://github.com/ocaml/dune/pull/7270/commits/42fdb7a2fa98f5541ad3102bd92c01551e7c2731), let me know if that's what you had in min.d

otherlibs/stdune/src/io.ml
```diff
@@ -268,22 +268,28 @@ struct
...
-        let src_stats = Unix.stat src in
-        (match src_stats.st_kind with
-        | S_DIR -> raise (Sys_error "Is a directory")
-        | _ -> ());
-        (try Copyfile.copyfile src dst with
-        | Unix.Unix_error (Unix.EPERM, "unlink", _) ->
-          let message = Printf.sprintf "%s: Is a directory" dst in
-          raise (Sys_error message)
-        | Unix.Unix_error (Unix.ENOENT, "realpath", _) ->
+        match Unix.stat src with
```
rgrinberg, line 271: Looks fine, but I prefer to avoid nesting whenever possible:

```
let stats =
  match Unix.stat src with
  | s -> s
  | exception .. -> .. raise
in
```

anmonteiro: Ok, I can do that

anmonteiro: Removed nesting in latest push.

otherlibs/stdune/src/io.ml
```diff
@@ -268,6 +268,18 @@ struct
       fun ?chmod ~src ~dst () ->
         let src = Path.to_string src in
         let dst = Path.to_string dst in
+        let src_stats =
+          match Unix.stat src with
+          | exception Unix.Unix_error (Unix.ENOENT, "realpath", _) ->
```
rgrinberg, line 273: The string isn't going to be `"realpath"` in this pattern. I think it would be better not to match on it at all

anmonteiro: good catch, this is an oversight.

PR #7271, opened 2023-03-10
fix(console): clean up console on exit

rgrinberg:


Alizter (approved):
Can confirm it fixes the issue. Needs a change log. (It might be worth including in a 3.7.1 however cc @emillon).

PR #7272, opened 2023-03-11
fix(boot): fix env vars when finding out concurrency

Alizter:
We were passing an empty enviornment to Unix.open_process_full_cmd before. We now pass the full available Unix.environment since we are only querying the concurrency by calling nproc or getconf.

On NixOS this was causing the bootstrap to default to a single processor since nproc wasn't available in the empty environment.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 Unreleased
 ----------
 
+- Bootstrap: correctly detect the number of processors by allowing `nproc` to be
+  looked up in `$PATH` (#7272, @Alizter)
+
 - Speed up file copying on macos by using `clonefile` when available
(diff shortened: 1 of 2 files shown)


rgrinberg:
CHANGES?

CHANGES.md
```diff
@@ -1,6 +1,9 @@
 Unreleased
 ----------
 
+- Fix the detection of the number of processors in a clean enviornment when
```
rgrinberg, line 4: What does it mean in a "clean environment"?

I would simplify to something like:

```
- Bootstrap: correctly detect the number of processors by allowing `nproc` to be looked up in `$PATH`.
```

Alizter: Thanks that's much better.

PR #7274, opened 2023-03-11
feature(boot): remove reliance on shell

rgrinberg:
Previously, we'd run the shell to discover the number of processes by
using `Unix.open_process_full`. The shell isn't being used for anything,
so we switch to executing the process directly.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 Unreleased
 ----------
 
+- Bootstrap: remove reliance on shell. Previously, we'd use the shell to get
+  the number of processors. (#7274, @rgrinberg)
+
 - Bootstrap: correctly detect the number of processors by allowing `nproc` to be
   looked up in `$PATH` (#7272, @Alizter)
 
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -189,21 +189,41 @@ let path =
   | exception Not_found -> []
   | s -> split_path s
 
+let find_prog ~f =
+  let rec search = function
+    | [] -> None
+    | dir :: rest -> (
+      match f dir with
+      | None -> search rest
+      | Some fn -> Some (dir, fn))
+  in
+  search path
+
 let exe = if Sys.win32 then ".exe" else ""
 
 (** {2 Concurrency level} *)
 
 let concurrency =
-  let try_run_and_capture_line cmd =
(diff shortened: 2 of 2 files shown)


boot/duneboot.ml
```diff
@@ -189,21 +189,41 @@ let path =
...
+    match
+      find_prog ~f:(fun dir ->
+          if Sys.file_exists (dir ^/ prog) then Some prog else None)
+    with
+    | None -> None
+    | Some (dir, prog) -> (
+      let path = dir ^/ prog in
+      let args = Array.of_list @@ (path :: args) in
+      let ic, oc, ec =
+        Unix.open_process_args_full path args (Unix.environment ())
```
anmonteiro, line 217: should this be the empty environment again, if we know the path of the program we're calling?

boot/duneboot.ml
```diff
@@ -218,7 +238,10 @@ let concurrency =
         | n -> n)
     else
       let commands =
-        [ "nproc"; "getconf _NPROCESSORS_ONLN"; "getconf NPROCESSORS_ONLN" ]
+        [ ("nproc", [])
+        ; ("getconf", [ "_NPROCESSORS_ONLN" ])
+        ; ("getconf", [ " NPROCESSORS_ONLN" ])
```
Alizter, line 243:
```suggestion
        ; ("getconf", [ "NPROCESSORS_ONLN" ])
```

CHANGES.md
```diff
@@ -1,6 +1,9 @@
 Unreleased
 ----------
 
+- Remove reliance on shell for bootstrapping. Previously, we'd use the shell to
+  get the number of processors. (#7274, @rgrinberg)
+
```
Alizter, line 6: Maybe just merge both our entries.

Alizter (approved):
I will test after Antonio's suggestion is considered.

boot/duneboot.ml
rgrinberg: I think clearing the path is generally unintuitive and I'd rather not do it unless there's a good reason. I have no idea how any of these programs I'm running work and if they could be affected by environment variables. It would be safer to just let them pass through.

CHANGES.md
rgrinberg: Why? One is a general improvement and the other is a bug fix. If they are merged, it would only obfuscate it for readers.

Alizter:
Works for me

PR #7277, opened 2023-03-11
nix: upgrade to patch that track `doc` section

anmonteiro:
This PR showcases how https://github.com/NixOS/nixpkgs/pull/220592 allows dune's cinaps tests to be run.

example of setup:

```shell
$ nix develop .#slim -c $SHELL

$ ./dune.exe build @test/blackbox-tests/test-cases/cinaps/runtest
```

PR #7279, opened 2023-03-11
chore(nix): remove OCaml nix-overlays

anmonteiro:


Alizter (commented):
I would like to have #7277 so I am not sure it makes sense to remove overlays atm.

Alizter (approved):
We can add them back in if they are needed.

Issue #6907, opened 2023-01-20
`dune build --watch` crashes when it detects filesystem change on macOS 13.1 on some Apple Silicon machines


dmzimmerman:
## Expected Behavior
I expected it to watch the filesystem and rebuild as necessary, without crashes.

## Actual Behavior
It crashes every time the filesystem changes.

## Reproduction

Start with any Dune project (I happened to use https://github.com/OCamlPro/ocaml-solidity). Open two terminals and change working directory in both to the root directory of the project. Run `dune build`, then `dune build --watch` in one (the reason for the initial `build` run is that if the project needs building, `dune build --watch` crashes immediately with the same error as you'll see on a filesystem change). In the other, change the filesystem - either by editing one of the files, or even something like `touch foo` to create a new file that Dune wouldn't even need to build. It should crash immediately with the message `Internal error, please report upstream including the contents of _build/log` and a stack trace (which you can see in the verbose gist below) that seems to indicate an `as_outside_build_dir_exn` being thrown. 

## Specifications

- Version of `dune` (output of `dune --version`): 3.6.1
- Version of `ocaml` (output of `ocamlc --version`): 4.14.1
- Operating system (distribution and version): macOS 13.1 / Apple Silicon

## Additional information 

I replicated this on completely clean installs of OCaml 4.14.1 and Dune via OPAM (i.e., `rm -rf ~/.opam; opam init; opam switch create 4.14.1` followed by `opam` installation of Dune and the required libraries for the project I was building, on both a MacBook Pro/M1 Pro/32GB RAM and a Mac Studio/M1 Ultra/128GB RAM. Colleagues of mine who have tried the same on 16GB M1 or M1 Pro machines do not seem to be experiencing this issue, but I don't know if that's just a coincidence or something meaningful.

- Link to gist with verbose output (run `dune` with the `--verbose` flag): https://gist.github.com/dmzimmerman/89f37f639c1fe4ce7e5ee00dc6e734dd
- Link to gist with `_build/log`: https://gist.github.com/dmzimmerman/0c533e90be627946637746b321874672

mro:
same here, any news?

mro:
dune.3.7.0 shows the same crash.

nojb:
Just an update: I will try to take a look (probably won't happen before next week -- I need to procure an M1 mac).

nojb:
[...]

PR #7282, opened 2023-03-11
ignore relative paths in PATH

Alizter:
Reading relative paths present in PATH is fragile behaviour and leads to issues such as #6907. We explicitly filter any relative paths when reading PATH.

### DIFF
--- a/otherlibs/stdune/src/bin.ml
+++ b/otherlibs/stdune/src/bin.ml
@@ -4,7 +4,9 @@ let parse_path ?(sep = path_sep) s =
   String.split s ~on:sep
   |> List.filter_map ~f:(function
        | "" -> None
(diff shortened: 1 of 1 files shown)


nojb (commented):
This change is more general than what I had in mind (filtering out relative paths only for the file watching mechanism), but indeed changing this across the board may make sense as it would make Dune more predictable.

There is a small chance this make break some workflows (eg someone setting `PATH=.:$PATH` in a CRAM test), but I am overall in favour.

Alizter:
Yeah, I suppose this could be too restrictive and heavy handed. I guess we could change it so that only watch mode filters, but I don't like the difference in behaviour between the regular build.

Maybe at the end of the day, the current behaviour is the most expected by a user and we should just fix watch mode's fsevents.

Alizter:
I don't want to do this in the end.

PR #7286, opened 2023-03-11
fix(nix): conflicting ocamlfind dependencies

anmonteiro:
In #7258, we removed opam2nix from the `default` package, but the default shell needs to depend on the correct one (now dune rather than default)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -166,7 +166,6 @@
 
       devShells.default =
         pkgs.mkShell {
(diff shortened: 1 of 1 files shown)


anmonteiro:
This makes `dune build @check` work again in the main shell

Alizter (approved):
`nix develop` and then `./dune.exe build @check` works for me.

PR #7291, opened 2023-03-11
ci: Add Nix Dev jobs

Alizter:
This will fail due to the doc location failures. When that is fixed we can add this job.

I've also added #7277 to see if it fixes the test-suite. It does not because it only affects slim :facepalm:.


Alizter:
No point keeping this open for now.

PR #7292, opened 2023-03-11
ci(coq): add nix coq job

Alizter:
Just checking for now.


Alizter:
FTR our tests need coq-native so we need to pull Coq with native enabled. Not sure how to do this atm and it probably won't be cached anyway.

PR #7221, opened 2023-03-06
fix: ignore sigpipe

rgrinberg:
Dune would disconnect on writes to clients that suddenly disconnected
because of a sigpipe. We ignore sigpipe and manually shutdown the
scheduler

fixes #6870

Signed-off-by: Rudi Grinberg <me@rgrinberg.com>

PR #7299, opened 2023-03-12
fix(rpc): ignore sigpipe on macos clients

rgrinberg:
RPC clients that suddenly disconnect will no longer raise sigpipe and
make dune terminate.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 Unreleased
 ----------
 
+- RPC: Ignore SIGPIPE when clients suddenly disconnect on OSX (#7299, partially
+  fixes #6879, @rgrinberg)
+
 - Always clean up the UI on exit. (#7271, fixes #7142 @rgrinberg)
 
 - Bootstrap: remove reliance on shell. Previously, we'd use the shell to get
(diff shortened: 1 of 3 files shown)


nojb (approved):
LGTM

An alternative is to ignore `SIGPIPE` globally (using `Sys.set_signal`), which would not need writing C code.

rgrinberg:
> An alternative is to ignore SIGPIPE globally (using Sys.set_signal), which would not need writing C code.

Yes, but then we need to handle terminating dune ourselves whenever stderr/stdout is closed. I think that ends up being messier, especially when we forget to do it for the various non-build sub commands (dune merlin, dune describe, etc.) I did try it here if you prefer to evaluate that approach: https://github.com/ocaml/dune/pull/7221

nojb:
> > An alternative is to ignore SIGPIPE globally (using Sys.set_signal), which would not need writing C code.
> 
> Yes, but then we need to handle terminating dune ourselves whenever stderr/stdout is closed. I think that ends up being messier, especially when we forget to do it for the various non-build sub commands (dune merlin, dune describe, etc.) I did try it here if you prefer to evaluate that approach: #7221

Got it, thanks! But don't we have the same problem on Linux?

rgrinberg:
> But don't we have the same problem on Linux?

Yes, I have a PR coming for linux as well.

PR #7302, opened 2023-03-12
fix(melange): rename `entries` to `modules` in `melange.emit`

anmonteiro:
- these are OCaml modules
- explaining the distinction between `modules` and `entries` is extra cognitive overhead

### DIFF
--- a/src/dune_rules/dune_file.ml
+++ b/src/dune_rules/dune_file.ml
@@ -96,8 +96,7 @@ module Buildable = struct
       located
         (only_in_library
            (field_o "cxx_names" (use_foreign >>> Ordered_set_lang.decode)))
-    and+ modules =
-      Stanza_common.Modules_settings.decode ~modules_field_name:"modules"
+    and+ modules = Stanza_common.Modules_settings.decode
     and+ self_build_stubs_archive_loc, self_build_stubs_archive =
       located
         (only_in_library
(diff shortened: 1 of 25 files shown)


jchavarri (approved):
At this point, I wonder if it'd make more sense to remove `modules` field altogether and have  `melange.emit` just.... emit. Downside: Melange projects should always define at least 1 library. Upsides: a lot of duplicated logic, fields etc between `emit` and `library` would go away.

anmonteiro:
I wouldn't like to have a library that performs side-effects such as rendering to the `DOM`.

jchavarri:
All libraries that would use reason-react would render to the DOM. Or they could make network requests, or log to the console. I'm not sure I understand how side effects are related to the suggestion?

anmonteiro:
I don't want to write a library with this code:

```ocaml
let () = launch_missiles
```

anmonteiro:
My point in a few more words: what you're describing already works -- if you don't have any source files in a directory with `melange.emit` or `entries` (now `modules`) is the empty set, then the target dir is the output of your `libraries`.

Having `modules` allows to have more side-effectful code like my contrived snippet above which I'd like not to write in library code.

jchavarri:
I still don't get the focus on side effects. I expect Melange libraries to include all sorts of side effects like network requests or rendering to the DOM, otherwise a very large part of functionality of frontend apps would have to be removed.

> what you're describing already works

True. I guess what I'm saying is that if the most common case of `melange.emit` will be to not have any modules inside, and load all of them through `libraries` field, we might as well simplify things on dune side to streamline maintenance and implementation. But it's probably too early to tell :)

anmonteiro:
> I still don't get the focus on side effects. I expect Melange libraries to include all sorts of side effects like network requests or rendering to the DOM, otherwise a very large part of functionality of frontend apps would have to be removed.

There's a distinction between side effects and side-effectful functions.

PR #7304, opened 2023-03-12
ocaml_flags: filter all warns,alerts from vendored_dirs flags

### LINKED ISSUE #7034: Dune treats warnings as errors according to the workspace `lang dune` version when building vendored packages with a more permissive `lang dune` version
## Expected Behavior

When compiling vendored packages, dune should act like the version specified in `dune lang` in the dune-project file when determining which warnings should be treated as errors.

## Actual Behavior

Dune appears to determine which warnings should be treated as errors based on the `dune lang` version in the top-level dune-project when compiling vendored libraries.

## Reproduction

[Here's a simple repro of this issue](https://github.com/gridbugs/repro-dune-errors-should-be-warnings)

It looks like between dune 3.2 and 3.3 dune was changed to treat warning 69 (declaring a mutable record field which is never mutated) as an error whereas it was previously just a warning. The top-level dune-project uses `dune lang 3.6` but the vendored package (resource-pooling) uses `dune lang 1.11`, and running `dune build` gives the error:
```
File "vendored/resource-pooling/resource_pool.ml", line 21, characters 2-27:
21 |   mutable node_prev : 'a t;
       ^^^^^^^^^^^^^^^^^^^^^^^^^
Error (warning 69 [unused-field]): mutable record field node_prev is never mutated.
File "vendored/resource-pooling/resource_pool.ml", line 22, characters 2-27:
22 |   mutable node_next : 'a t;
       ^^^^^^^^^^^^^^^^^^^^^^^^^
Error (warning 69 [unused-field]): mutable record field node_next is never mutated.
```

But if we take the source of resource-pooling by itself, it builds with a warning instead.

The reverse situation also causes unexpected behaviour. If we modify resource-pooling to use `dune lang 3.6` it no longer builds as warning 69 is now an error, but if we vendor the modified resource-pooling in a project using `dune lang 3.2` or below then it builds with a warning.

## Specifications

- Version of `ocaml` (output of `ocamlc --version`): 5.0.0
- Operating system (distribution and version): nixos

Alizter:
- [x] Fix tests
- [x] Changes
- [x] Documentation - need to check if we mention the disabling of warnings in the doc
- [x] Should fix #7034

### DIFF
--- a/src/dune_rules/ocaml_flags.ml
+++ b/src/dune_rules/ocaml_flags.ml
@@ -167,9 +167,31 @@ let map_common t ~f =
 
 let append_common t flags = map_common t ~f:(fun l -> l @ flags)
 
-let with_vendored_warnings t = append_common t vendored_warnings
+let filter_warnings l =
+  let rec loop acc = function
+    | [] -> List.rev acc
+    | "-w" :: _ :: rest -> loop acc rest
+    | "-w" :: [] -> List.rev acc
+    | "-warn-error" :: _ :: rest -> loop acc rest
+    | "-warn-error" :: [] -> List.rev acc
+    | a :: rest -> loop (a :: acc) rest
+  in
+  loop [] l
 
-let with_vendored_alerts t = append_common t vendored_alerts
+let with_vendored_warnings t =
+  append_common (map_common ~f:filter_warnings t) vendored_warnings
+
+let filter_alerts l =
+  let rec loop acc = function
+    | [] -> List.rev acc
+    | "-alert" :: _ :: rest -> loop acc rest
+    | "-alert" :: [] -> List.rev acc
+    | a :: rest -> loop (a :: acc) rest
+  in
+  loop [] l
+
+let with_vendored_alerts t =
+  append_common (map_common ~f:filter_alerts t) vendored_alerts
 
 let dump t =
   let+ common = t.common
--- a/test/blackbox-tests/test-cases/github7034.t/run.t
+++ b/test/blackbox-tests/test-cases/github7034.t/run.t
@@ -98,12 +98,12 @@ But when lang dune is 3.3 or higher the warning becomes an error:
   Leaving directory 'outer'
   $ dune build --root=outer
   Entering directory 'outer'
-  File "vendored/inner/inner.ml", line 6, characters 11-18:
-  6 | type t = { x : int }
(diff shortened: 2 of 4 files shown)


gridbugs (approved):

rgrinberg:
@emillon could you review this one?

test/blackbox-tests/test-cases/github7034.t/run.t
```diff
@@ -98,12 +98,7 @@ But when lang dune is 3.3 or higher the warning becomes an error:
...
   $ dune build --root=outer
   Entering directory 'outer'
-  File "vendored/inner/inner.ml", line 6, characters 11-18:
-  6 | type t = { x : int }
-                 ^^^^^^^
-  Error (warning 69 [unused-field]): unused record field x.
   Leaving directory 'outer'
-  [1]
 
 This is unexpected as vendored projects should be built according to their
```
emillon, line 103: The comments here should be updated since they don't reflect what the test is doing now

src/dune_rules/ocaml_flags.ml
```diff
@@ -167,9 +167,31 @@ let map_common t ~f =
 
 let append_common t flags = map_common t ~f:(fun l -> l @ flags)
 
-let with_vendored_warnings t = append_common t vendored_warnings
```
emillon, line 170: I'm a bit surprised that it's necessary to remove the warning options from the CLI. Isn't `-w <whatever> -w -a` equivalent to `-w -a`?

Alizter: I've forgotten my exact thoughts here, but one reason to do this is so that vendored libraries do not get affected by warning changes in the workspace.

test/blackbox-tests/test-cases/github7034.t/run.t
Alizter: I changed the comment to have a header saying "notes from before".

Alizter:
@emillon perhaps this could be considered form 3.8.

src/dune_rules/ocaml_flags.ml
rgrinberg: We should address Etienne's question here before we consider this for 3.8

Alizter:
Can be reopened if we deem it important.

PR #7309, opened 2023-03-13
melange: make alias non-optional in `melange.emit`

jchavarri:
As shown in https://github.com/ocaml/dune/pull/7261, building Melange projects without using aliases will lead users to confusing situations, as the dependencies between modules are not tracked between `js` targets. One just needs to:
- have a `melange.emit` stanza with 2 modules where one depends on the other
- try to build using the `.js` target of the first

To run into this situation.

See https://github.com/ocaml/dune/pull/7261#issuecomment-1466393229:

> This PR demonstrates expected behavior. When you do `dune build x`, dune will only build x and its dependencies. Other things such as runtime dependencies are not expected to build. This is why we introduced the alias to the melange.emit stanza

### DIFF
--- a/src/dune_rules/melange/melange_rules.ml
+++ b/src/dune_rules/melange/melange_rules.ml
@@ -216,27 +216,24 @@ let setup_emit_cmj_rules ~sctx ~dir ~scope ~expander ~dir_contents
     let stdlib_dir = ctx.stdlib_dir in
     let+ () =
       let target_dir = Path.Build.relative dir mel.target in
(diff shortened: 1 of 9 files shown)


anmonteiro:
> Perhaps there should be a default alias like "melange"?

I like this a lot better, and it's something we had talked about. 

This allows us to have the 2 options:

1. build _all_ JS
2. opt-out of building all JS

jchavarri:
Hm I see, I will open PR with default alias.

Issue #6151, opened 2022-09-15
fsevents segfaulting under M1 mac


emillon:
Hi,

I'm trying to make the test suite work on M1 macs.

At the moment, the inline tests for `fsevents` are triggering a segfault.
I reduced that to the following piece of code:

```ocaml
open Stdune
module Event = Fsevents.Event

let main () =
  let dir = Temp.create Dir ~prefix:"fsevents_dune" ~suffix:"" in
  Sys.chdir (Path.to_string dir);
  let cwd = Sys.getcwd () in
  let fsevent_r = ref None in
  let fsevent =
    Fsevents.create ~paths:[ cwd ] ~latency:0.0 ~f:(fun _ ->
        Fsevents.stop (Option.value_exn !fsevent_r);
        raise Exit)
  in
  fsevent_r := Some fsevent;
  let runloop = Fsevents.RunLoop.in_current_thread () in
  Fsevents.start fsevent runloop;
  match Fsevents.RunLoop.run_current_thread runloop with
  | Ok () -> ()
  | Error e -> raise e
```

Here's the crash location:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xbeadde8d2dc8)
    frame #0: 0x00000001000f7d90 t.exe`dune_fsevents_callback(streamRef=<unavailable>, t=0x0000600000202dc0, numEvents=<unavailable>, eventPaths=1 element, eventFlags=0x000000010058c000, eventIds=0x0000000100588000) at fsevents_stubs.c:120:23 [opt]
   117    }
   118    v_res = caml_callback_exn(t->v_callback, v_events_xs);
   119    if (Is_exception_result(v_res)) {
-> 120      t->runloop->v_exn = Extract_exception(v_res);
   121      CFRunLoopStop(t->runloop->runloop);
   122    }
   123    CAMLdrop;
Target 0: (t.exe) stopped.
warning: t.exe was compiled with optimization - stepping may behave oddly; variables may not be available.
(lldb) p/x t
(dune_fsevents_t *) $0 = 0x0000600000202dc0
(lldb) p/x t->runloop
(dune_runloop *) $1 = 0x0000beadde8d2dc0
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xbeadde8d2dc8)
[...]

PR #6215, opened 2022-10-11
fsevents: move free to finalizer

emillon:

PR #7312, opened 2023-03-14
fsevents: move free to finalizer

gridbugs:
Fixes https://github.com/ocaml/dune/issues/6151

This just a copy of https://github.com/ocaml/dune/pull/6215 rebased onto main. I've tested this on both an M1 and x86 mac and it fixes the problem.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -64,6 +64,9 @@ Unreleased
 
 - Allow parallel execution of inline tests partitions (#7012, @hhugo)
 
+- Fix segfault on MacOS when dune was being shutdown while in watch mode.
+  (#7312, fixes #6151, @gridbugs, @emillon)
+
 3.7.0 (2023-02-17)
 ------------------
 
--- a/src/fsevents/fsevents_stubs.c
+++ b/src/fsevents/fsevents_stubs.c
@@ -39,6 +39,20 @@ static struct custom_operations dune_fsevents_runloop_ops = {
     custom_serialize_default,   custom_deserialize_default,
     custom_compare_ext_default, custom_fixed_length_default};
 
+#define Fsevents_val(v) (*(dune_fsevents_t **)Data_custom_val(v))
+
+static void dune_fsevents_t_finalize(value v_t) {
+  dune_fsevents_t *t = Fsevents_val(v_t);
+  caml_remove_global_root(&t->v_callback);
+  caml_stat_free(t);
+}
+
+static struct custom_operations dune_fsevents_t_ops = {
+    "dune.fsevents.fsevents_t", dune_fsevents_t_finalize,
+    custom_compare_default,     custom_hash_default,
+    custom_serialize_default,   custom_deserialize_default,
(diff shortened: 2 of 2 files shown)


gridbugs:
(copied from https://github.com/ocaml/dune/issues/6151)

I did some digging today into why the crash happens when the fsevents callback calls Fsevents.stop.

The c function dune_fsevents_callback is called on each event, and invokes the user-provided callback (the ~f argument to Fsevents.create) before dereferencing a field of the dune_fsevents_t struct . The c function dune_fsevents_stop deallocates the dune_events_t struct and lucky for us the deallocator seems to corrupt its pointer fields. dune_fsevents_stop is called by the ocaml function Fsevents.stop. So when the fsevents callback calls Fsevents.stop dune_fsevents_callback tries to dereference a field of the dune_events_t after its pointer fields have been corrupted and a segfault happens.

This implies that we'd also see this crash if we ever started, stopped, and then re-started a Fsevents.t as stop causes the value to be deallocated but it's only allocated inside create, though is moot as the Fsevents.state state machine doesn't allow stopping and re-starting.

@emillon's fix (https://github.com/ocaml/dune/pull/6215) solves the problem by removing the call to caml_stat_free(t); from dune_fsevents_stop and registering a custom finalizer for the dune_fsevents_t object instead. I've copied to https://github.com/ocaml/dune/pull/7312 and rebased onto main.

gridbugs:
> Sounds good of course but I think you could have pushed the original PR too.

I couldn't figure out how to push to your branch :)

> My comments from there still apply - I'm not confident about the correctness of the memory management here but I think it's an improvement compared to the current situation (and it fixes the issue!).

Should we hold off merging this until we better understand the memory management?

> Please add a changelog entry.

Done!

emillon (approved):

CHANGES.md
```diff
@@ -60,6 +60,9 @@ Unreleased
 - Allow `(stdlib ...)` to be used with `(wrapped false)` in library stanzas
   (#7139, @anmonteiro).
 
+- Fix segfault on macos when `Fsevents.stop` is called inside the callback to
+  `Fsevents.create`. (#7312, fixes #6151, @gridbugs, @emillon)
```
rgrinberg, line 64: I'd remove such details from the CHANGE list. It's supposed to make sense to end users that know nothing about dune's code. Perhaps something like:

```
- Fix segfault on MacOS when dune was being shutdown while in watch mode.
```

nojb (commented):
LGTM

Another approach: using a custom block as in this PR, you can set the "data pointer" to NULL after deallocating the `dune_fsevents_t` data structure. The callback would check whether the field is NULL before trying to dereference it and not do anything otherwise. This avoids the use of the finalisation function.

rgrinberg:
@gridbugs what do you think about my suggesting for the CHANGES entry?

gridbugs:
> @gridbugs what do you think about my suggesting for the CHANGES entry?

Looks good. I've updated CHANGES with your suggestion.

PR #7319, opened 2023-03-14
fix(rpc): ignore sigpipe on linux

rgrinberg:
We ignore sigpipe by sending messages with `sendmsg` with the appropriate flag.

All of this requires us to get rid of the `Out_channel`, but this desirable anyway, since we'll need it gone for evented IO.

### DIFF
--- a/src/csexp_rpc/csexp_rpc.ml
+++ b/src/csexp_rpc/csexp_rpc.ml
@@ -128,7 +128,13 @@ module Session = struct
   type state =
     | Closed
     | Open of
-        { out_channel : out_channel
+        { out_buf : Io_buffer.t
+        ; fd : Unix.file_descr
+        ; (* A mutex for modifying [out_buf].
+
+             Needed as long as we use threads for async IO. Once we switch to
+             event based IO, we won't need this mutex anymore *)
+          write_mutex : Mutex.t
         ; in_channel : in_channel
         ; writer : Worker.t
         ; reader : Worker.t
@@ -139,13 +145,22 @@ module Session = struct
     ; mutable state : state
     }
 
-  let create in_channel out_channel =
+  let create fd in_channel =
     let id = Id.gen () in
     if debug then
       Log.info [ Pp.textf "RPC created new session %d" (Id.to_int id) ];
     let* reader = Worker.create () in
     let+ writer = Worker.create () in
-    let state = Open { in_channel; out_channel; reader; writer } in
+    let state =
+      Open
+        { fd
+        ; in_channel
+        ; out_buf = Io_buffer.create ~size:8192
+        ; write_mutex = Mutex.create ()
+        ; writer
+        ; reader
+        }
+    in
     { id; state }
 
   let string_of_packet = function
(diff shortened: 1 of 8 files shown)


src/csexp_rpc/io_buffer.ml
```diff
@@ -0,0 +1,79 @@
...
+    let new_size = min buf_len needed in
+    let bytes = Bytes.create new_size in
+    Bytes.blit ~src:t.bytes ~src_pos:t.pos_r ~dst:bytes ~dst_pos:0
+      ~len:(t.pos_w - t.pos_r);
+    t.bytes <- bytes;
+    t.pos_w <- t.pos_r;
+    t.pos_r <- 0)
+
+let write_char_exn t c =
+  Bytes.set t.bytes t.pos_w c;
```
nojb, line 37: I would add an assert checking the precondition.
```suggestion
  assert (t.pos_w < Bytes.length t.bytes);
  Bytes.set t.bytes t.pos_w c;
```

src/csexp_rpc/io_buffer.ml
```diff
@@ -0,0 +1,79 @@
...
+  t.pos_w <- t.pos_w + 1
+
+let write_string_exn t src =
+  let len = String.length src in
```
nojb, line 41:
```suggestion
  assert (t.pos_w + len <= Bytes.length);
  let len = String.length src in
```

src/csexp_rpc/io_buffer.ml
```diff
@@ -0,0 +1,79 @@
...
+   sigpipes. Eventually we'll move to event based IO and ditch the threads,
+   so we'll need this anyway *)
+
+let create () =
+  { bytes = Bytes.create 8192; pos_r = 0; pos_w = 0; total_written = 0 }
+
+let maybe_resize_to_fit t write_size =
+  let buf_len = Bytes.length t.bytes in
+  let capacity = buf_len - t.pos_w in
+  if capacity < write_size then (
```
nojb, line 26: Often it is better to grow the buffer exponentially (eg by multiplying the length by a power of 2) to keep performing a linear number of resizings.

nojb (commented):
LGTM (I left a few comments)

One point that I would be careful with is that the buffers seem to be long-lived, but there is no facility for "shrinking" them, which means that memory usage will only increase for them. This may become a problem in certain scenarios, right?

src/csexp_rpc/csexp_rpc.ml
```diff
@@ -203,6 +219,30 @@ module Session = struct
...
+        let bytes = Io_buffer.bytes out_buf in
+        let pos = Io_buffer.pos out_buf in
+        let len = Io_buffer.length out_buf in
+        try write fd bytes pos len
+        with exn ->
+          Mutex.unlock write_mutex;
+          reraise exn
+      in
+      Io_buffer.read out_buf written;
+      Mutex.unlock write_mutex;
```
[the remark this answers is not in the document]
rgrinberg: I think it's better because it's allowing callers to refill the write buffer on every send. Less write calls and more mutex locks sounds like a better trade off. Anyway, the mutex will be gone soon so I suggest we don't dwell on it.

rgrinberg:
> One point that I would be careful with is that the buffers seem to be long-lived, but there is no facility for "shrinking" them, which means that memory usage will only increase for them. This may become a problem in certain scenarios, right?

There's natural push back on writing. I tweaked it to stop doubling past a certain size. That should restore the size after sending particularly large.

PR #6625, opened 2022-12-02
melange: demonstrate issue with ocamldep + melange-ppx

jchavarri:
Showcases some issue that leads to race conditions due to missing dependencies on Melange large codebases.

I am not sure what the fix should be, but I assume `ocamldep` can't be updated to run Melange ppxs, so maybe Dune should start using `meldep` for Melange libraries?

Issue #171, opened 2017-06-27
Preprocess using both cppo and PPXs


aantron:
From https://github.com/ocsigen/lwt/issues/411#issuecomment-311214821:

> it is possible to both apply a PPX and preprocess with cppo, but it requires the [separate] cppo `rule` stanza, and so presumably a source file with a name like `foo.cppo.ml`. It would be nice if one could both apply cppo and PPXs to source named `foo.ml`.

> This is not a high priority, but it would make files like `lwt_unix.cppo.ml`, where we both want to use cppo, and apply Bisect_ppx, ever so slightly more maintainable.

I'm basically thinking of being able to have either multiple `preprocess` stanzas, or list both `pps` and [`action`](http://jbuilder.readthedocs.io/en/latest/quick-start.html#using-cppo) in one `preprocess` stanza. There may, of course, be other, better ways.

aantron:
Looking at ocsigen/lwt#411 closer, the issue with paths appears to be an inconsistency between how Jbuilder runs PPXs and how it [suggests](https://github.com/janestreet/jbuilder/blob/3b1182c23a0c1993e4077a2cb5e3a611697806e5/doc/quick-start.rst#using-the-cppoml-style-like-the-ocamlbuild-plugin) to run cppo. I believe this is what @andrewray [found](https://github.com/ocsigen/lwt/issues/411#issuecomment-311935687).

```scheme
(rule
 ((targets (lwt_unix.ml))
  (deps    (lwt_unix.cppo.ml))
  (action
   (run ${bin:cppo} -V OCAML:${ocaml_version} ${<} -o ${@}))))
```

Presumably, the above causes Jbuilder to pass the filename `lwt_unix.cppo.ml` to cppo, so of course cppo has to use that. This filename is relative to the `src/unix/` subdirectory.

When the PPX is later run, the working directory is (a copy of) the root directory of the project, so the path to access this file is `src/unix/lwt_unix.cppo.ml` instead.

It would be nice to eliminate this inconsistency (as well as allow running cppo and PPXs from the `preprocess` stanzas).

Lwt is working around this like so:

```scheme
(rule
 ((targets (lwt_unix.ml))
  (deps    (lwt_unix.cppo.ml))
  (action
   (chdir ${ROOT} (run ${bin:cppo} -V OCAML:${ocaml_version} ${<} -o ${@})))))
```

as suggested by @andrewray.

diml:
[...]

PR #7321, opened 2023-03-15
test(melange): show work around for ocamldep issue + melange ppx

anmonteiro:
This is an alternative version of https://github.com/ocaml/dune/pull/6625/ with a workaround that's not manual.

Right now it's possible to overcome the limitation of `ocamldep` running before the melange PPX by:
- adding a `(preprocess (action melc --as-ppx %{input-file}))` field

This, however, has a pretty significant limitation: it isn't possible to run both actions + ppxes in a `preprocess` field (see https://github.com/ocaml/dune/issues/171).

So, it sounds to me that we need to solve this in melange too:
- expose a melange ppx library that works with the `pps` preprocessing stack
  - https://github.com/melange-re/melange/pull/480 was already a step in this direction, by decoupling a few things in the melange source that make it possible to publish a separate PPX.

### DIFF
--- a/test/blackbox-tests/test-cases/melange/ocamldep-melange-ppx-workaround.t
+++ b/test/blackbox-tests/test-cases/melange/ocamldep-melange-ppx-workaround.t
@@ -1,0 +1,58 @@
+Showcases issues with ocamldep when using melange ppx
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.7)
+  > (using melange 0.1)
+  > EOF
+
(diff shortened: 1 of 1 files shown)


jchavarri:
@anmonteiro I believe this can be closed now?

jchavarri:
ref: https://github.com/ocaml/dune/pull/7323#issuecomment-1519406280

anmonteiro:
Indeed this can be closed. We added tests in melange for this case.

PR #7207, opened 2023-03-02
fix(rules): merge allowed subdirs correctly for Redirect_to_parent rules

anmonteiro:
- After https://github.com/ocaml/dune/pull/6637, the engine may redirect rule loading to the parent if it finds that it needs to find rules in the parent directory
- Part of generating rules in a directory is also finding out which sub-directories Dune is allowed to traverse; we store that in the record fields `build_dir_only_sub_dirs`
- However, while `Build_config.Rules.t` can now contain rules for multiple directories, `build_dir_only_sub_dirs` is still a flat set of sub-directories, leading Dune to (wrongly) assume that both directories which had their rules combined can traverse the same set of sub-directories, when that's not the case.
- This diff fixes that inconsistency by tracking the allowed set of sub-directories at each level
  - Specifically, this change unlocks traversing all directories when finding rules for recursive aliases, not just the directories in the source tree.

PR #7322, opened 2023-03-15
test(melange): show dependency cycle depending on runtime_deps in target dir

anmonteiro:


test/blackbox-tests/test-cases/melange/emit-with-runtime-deps-edge-cases.t
```diff
@@ -53,11 +53,9 @@ The runtime_dep index.txt was copied to the build folder
   hello from file
   
 
-
-
 Test depending on non-existing paths
 
-  $ mkdir another
+  $ mkdir -p another/another-output/another
```
anmonteiro, line 58: I had only ran the action builder eagerly because of this test, but it looks like creating its target dir in source makes the error pop up again, so it's rather an instance of #7207 and the recursive alias work that's gonna come after it.

test/blackbox-tests/test-cases/melange/emit-with-runtime-deps-dir-target.t
```diff
@@ -31,7 +31,9 @@ Rules created for the assets in the output directory
   [1]
 
   $ dune build @mel --display=short
+          bash some_dir
```
rgrinberg, line 34: In a separate PR, it would be good to switch this test to `sh` unless you need some special bash features.

PR #7323, opened 2023-03-15
melange: add test to reproduce "inconsistent assumptions" issue when using melange ppx

jchavarri:
Related to #6625.

Found recently another error that is related to the Melange ppx, in this case, considering a sufficiently complex dep tree between modules, changing a leaf in the dep tree leads to inconsistencies on how inner modules see the cmis.

I can merge this PR into #6625 if it makes things easier.

### DIFF
--- a/test/blackbox-tests/test-cases/melange/library-cmi-consistency.t
+++ b/test/blackbox-tests/test-cases/melange/library-cmi-consistency.t
@@ -67,7 +67,7 @@ and B (transitive)
          make inconsistent assumptions over interface My_lib__D
   [1]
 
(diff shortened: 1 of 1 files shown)


jchavarri:
Added showcase of a fix using @anmonteiro workaround from #7321: https://github.com/ocaml/dune/pull/7323/commits/ba30e9b33712814da3ddd190ddaaafca8cf2ebf0

jchavarri:
Is there any value on keeping this test around?

anmonteiro:
no, we added it in melange.

PR #7326, opened 2023-03-15
refactor: introduce source tree library

rgrinberg:
Here's a `Dune_source_tree` library that strips `Source_tree` to the bare essentials. There's a bit of functors to make it independent from the engine. I thought that would be a bit more cleaner this way, but it's debatable.

@snowleopard this is the library that you would use internally to reimplement source tree walking. There's one more design decision that I haven't yet made:

Should we add a field for users to attach arbitrary data to `Source_tree.t` nodes making the type `'a Source_tree.t`? That might make it simpler to port existing code. Since it wouldn't require moving data to separate memo tables.

### DIFF
--- a/src/dune_source_tree/dune
+++ b/src/dune_source_tree/dune
@@ -1,0 +1,3 @@
+(library
+ (name dune_source_tree)
+ (libraries stdune unix ordering dyn memo))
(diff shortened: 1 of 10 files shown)


rgrinberg:
Although this PR works well, the problem with the trie based approach is that it's inherently memoization unfriendly. Closing until we find something that will be worth sharing.

PR #7327, opened 2023-03-15
melange: add implicit alias

jchavarri:
As discussed in https://github.com/ocaml/dune/pull/7309#issuecomment-1466661192:

- Add new implicit alias `melange` that includes as deps every target generated from melange emit stanzas
- Handle case where the implicit alias is used explicitly


src/dune_rules/melange/melange_stanzas.ml
```diff
@@ -92,7 +94,7 @@ module Emit = struct
                  ])
          in
          field "target" (plain_string (fun ~loc s -> of_string ~loc s))
-       and+ alias = field_o "alias" Alias.Name.decode
+       and+ alias_loc, alias = located (field_o "alias" Alias.Name.decode)
```
anmonteiro, line 97:
```suggestion
       and+ alias_loc, alias = field_o "alias" (located Alias.Name.decode)
```

src/dune_rules/melange/melange_rules.ml
```diff
@@ -333,14 +338,19 @@ let setup_runtime_assets_rules sctx ~dir ~target_dir ~mode
...
+      Action_builder.paths
+        (non_copy @ List.rev_map copy ~f:(fun (_, target) -> Path.build target))
+    in
+    let* () =
+      let alias =
+        Alias.make Melange_stanzas.Emit.implicit_alias ~dir:target_dir
+      in
+      Rules.Produce.Alias.add_deps alias deps
+    in
     match mel.alias with
```
anmonteiro, line 351: You can run these in parallel
```suggestion
    let+ () =
      let alias =
        Alias.make Melange_stanzas.Emit.implicit_alias ~dir:target_dir
      in
      Rules.Produce.Alias.add_deps alias deps
    and+ () =
    match mel.alias with
```

src/dune_rules/melange/melange_rules.ml
```diff
@@ -216,27 +216,32 @@ let setup_emit_cmj_rules ~sctx ~dir ~scope ~expander ~dir_contents
...
           Compilation_context.requires_link cctx
           >>= js_targets_of_libs sctx ~module_systems ~target_dir
         in
-        Action_builder.paths deps)
-        |> Rules.Produce.Alias.add_deps alias
+        Action_builder.paths deps
+      in
+      let alias = Alias.make Melange_stanzas.Emit.implicit_alias ~dir in
+      let* () = Rules.Produce.Alias.add_deps alias emit_deps in
+      let* () = Rules.Produce.Alias.add_deps alias lib_deps in
```
anmonteiro, line 238:
```suggestion
      let* () = Rules.Produce.Alias.add_deps alias (Action_builder.both emit_deps lib_deps) in
```

src/dune_rules/melange/melange_rules.ml
```diff
@@ -216,27 +216,32 @@ let setup_emit_cmj_rules ~sctx ~dir ~scope ~expander ~dir_contents
...
+      match mel.alias with
+      | None -> Memo.return ()
+      | Some alias_name ->
+        let alias = Alias.make alias_name ~dir in
+        let* () = Rules.Produce.Alias.add_deps alias emit_deps in
+        Rules.Produce.Alias.add_deps alias lib_deps
```
anmonteiro, line 244:
```suggestion
        Rules.Produce.Alias.add_deps alias (Action_builder.both emit_deps lib_deps)
```

anmonteiro (commented):
Approach looks good, and code's almost there too IMO.

I'm torn on whether we should a) error when using the implicit alias or b) coalesce them both

src/dune_rules/melange/melange_rules.ml
jchavarri: `add_deps` expects `unit Action_builder.t` so this change doesn't work.

```
This expression has type (unit * unit) Action_builder.t
but an expression was expected of type unit Action_builder.t
```

anmonteiro: you can toss an `Action_builder.ignore` or use the let syntax.

src/dune_rules/melange/melange_rules.ml
jchavarri: Done in https://github.com/ocaml/dune/pull/7327/commits/950d3f91a7f6231d4e49de706e46b339dede5389.

src/dune_rules/melange/melange_rules.ml
jchavarri: You mean st like this?

```
let* () =
        Rules.Produce.Alias.add_deps alias
          (Action_builder.ignore (Action_builder.both emit_deps lib_deps))
```

Sorry, I think I still don't understand what you mean, and never feel comfortable using `ignore`.

jchavarri: Please let me know if https://github.com/ocaml/dune/pull/7327/commits/f9c061e6ea1fe2084d06ba97c4767a7d7a73ac0c is what you had in mind.

src/dune_rules/melange/melange_stanzas.ml
anmonteiro, line 97:
```suggestion
       and+ alias_loc, alias = field_o "alias" (located Alias.Name.decode)
```

src/dune_rules/melange/melange_stanzas.ml
```diff
@@ -110,6 +112,22 @@ module Emit = struct
        and+ allow_overlapping_dependencies =
          field_b "allow_overlapping_dependencies"
        and+ modules = Stanza_common.Modules_settings.decode in
+       let () =
+         match alias with
+         | None -> ()
+         | Some a -> (
```
anmonteiro, line 118:
```suggestion
         | Some (loc, a) -> (
```

src/dune_rules/melange/melange_stanzas.ml
jchavarri: If I do this change, then I have to `alias = Option.map ~f:snd alias` or similar when creating the record that will be returned from the function:

https://github.com/ocaml/dune/blob/4245029f2914bcfdca915e9916a5155a0afa95fb/src/dune_rules/melange/melange_stanzas.ml#L125

I'd rather keep it as is, but i don't have a strong opinion.

anmonteiro: no strong opinion

anmonteiro (approved):

src/dune_rules/melange/melange_stanzas.ml
rgrinberg: @jchavarri I think this comment needs to be addressed?

anmonteiro: It’s actually OK, we discussed it in https://github.com/ocaml/dune/pull/7327#discussion_r1141161355

jchavarri: Yes, I understood this comment as part of the same change as the other one. @rgrinberg I decided to leave it as is as it was not a blocker for Antonio, but I can change it if you think it is.

rgrinberg:
Does it make sense to have the user defined alias and the implicit alias? At first glance, it would make sense if it was only one or the other. E.g if I write

```
(melange.emit
 (alias foo))
```

I wouldn't expect `@melange` would build this emit.

Alizter:
> Does it make sense to have the user defined alias and the implicit alias? At first glance, it would make sense if it was only one or the other. E.g if I write
> 
> ```
> (melange.emit
>  (alias foo))
> ```
> 
> I wouldn't expect `@melange` would build this emit.

I think it would be better for `alias` to override the default rather than append to it as that is confusing behaviour. We have this confusing behaviour for the cram stanza for example. If this is expanded to an `aliases` field then users can also add `melange` if they so wish without being forced to choose between the default or their alias.

jchavarri:
Changed logic so that `melange` is not used for the targets of the stanza if it already has one alias in https://github.com/ocaml/dune/pull/7327/commits/c30cbf085c0002faebc8fa3ac286088791bc463d.

src/dune_rules/melange/melange_stanzas.ml
```diff
@@ -112,6 +114,22 @@ module Emit = struct
...
+           match Alias.Name.equal a implicit_alias with
+           | false -> ()
+           | true ->
+             User_error.raise ~loc:alias_loc
+               [ Pp.textf
+                   "The alias %S is assigned implicitly by Dune to include all \
+                    dependencies generated by all melange.emit stanzas found \
+                    in the project. To use an alias for this particular \
+                    stanza, use a different name."
+                   (Alias.Name.to_string implicit_alias)
```
rgrinberg, line 130: Do we really need this check anymore? I see nothing wrong with writing `(alias melange)` explicitly even if it's useless.

jchavarri: Fixed in https://github.com/ocaml/dune/pull/7327/commits/685a9b31d9f92241b6e884cf964cc67f0abe4576.

Issue #6767, opened 2022-12-23
Update build_path_prefix_map


rgrinberg:
To fix these warnings:

```sh
$ nix shell nixpkgs#ocaml-ng.ocamlPackages_5_0.ocaml -c make -B _boot/dune.exe 
ocamlc -output-complete-exe -w -24 -g -o .duneboot.exe -I boot -I +unix unix.cma boot/libs.ml boot/duneboot.ml
./.duneboot.exe
Done: 55/974 (jobs: 4)cd _boot && /nix/store/4j1a8c3nkl97v7a01fifgdf0xjzkk5iy-ocaml-5.0.0/bin/ocamlopt.opt -c -g -no-alias-deps -w -49-6 -alert -unstable -I +unix -I +threads build_path_prefix_map.ml
File "vendor/build_path_prefix_map/src/build_path_prefix_map.ml", line 5, characters 17-31:
Alert deprecated: Stdlib.Printf.kprintf
Use Printf.ksprintf instead.
Done: 213/974 (jobs: 4)cd _boot && /nix/store/4j1a8c3nkl97v7a01fifgdf0xjzkk5iy-ocaml-5.0.0/bin/ocamlopt.opt -c -g -no-alias-deps -w -49-6 -alert -unstable -I +unix -I +threads opamLexer.ml
File "vendor/opam-file-format/src/opamLexer.mll", line 21, characters 2-16:
Alert deprecated: Stdlib.Printf.kprintf
Use Printf.ksprintf instead.

```

but doesn't seem fatal.

_Originally posted by @Alizter in https://github.com/ocaml/dune/issues/6753#issuecomment-1363453542_

Alizter:
The opam lexer is way ahead of our forked version, so updating it wasn't a good use of time. We could fix the kprintf's directly. WDYT @rgrinberg?

rgrinberg:
Let's see what @Leonidas-from-XIV says. Perhaps he can do this for us since he's worked with the opam libs

Leonidas-from-XIV:
It looks like I need to update `opam-file-format` since the opam libs depends on a newer version anyway.

PR #7328, opened 2023-03-15
Update opam-file-format to 2.1.5

Leonidas-from-XIV:
The opam libraries need a newer version of `opam-file-format` so in this PR I'm updating our vendored copy to the latest release, 2.1.5 and changing it to use the non-deprecated `FullPos` API (which explains why there are so many changes to the dune code)


src/dune_engine/package.ml
```diff
@@ -99,13 +99,17 @@ module Dependency = struct
...
-    let to_relop : t -> OpamParserTypes.relop = function
-      | Eq -> `Eq
-      | Gte -> `Geq
-      | Lte -> `Leq
-      | Gt -> `Gt
-      | Lt -> `Lt
-      | Neq -> `Neq
+    let to_relop : t -> OpamParserTypes.FullPos.relop =
+      let nopos pelem =
+        OpamParserTypes.FullPos.{ pelem; pos = Opam_file.nopos }
```
rgrinberg, line 104: No need to open the module. This should be enough:
```
{ OpamParserTypes.FullPos. pelem; pos = Opam_file.nopos }
```

src/dune_engine/package.ml
```diff
@@ -756,21 +761,21 @@ let load_opam_file file name =
...
+    match value.pelem with
+    | String s -> Some [ s ]
+    | List l ->
       let+ l =
-        List.fold_left l ~init:(Some []) ~f:(fun acc v ->
+        List.fold_left l.pelem ~init:(Some []) ~f:(fun acc v ->
             let* acc = acc in
-            match v with
-            | OpamParserTypes.String (_, s) -> Some (s :: acc)
+            match v.OpamParserTypes.FullPos.pelem with
```
rgrinberg, line 777: We don't qualify module names in fields. So this should be `v.pelem`. If necessary, you can add a type annotation to `v`.

src/dune_engine/package.ml
```diff
@@ -756,21 +761,21 @@ let load_opam_file file name =
...
+            | OpamParserTypes.FullPos.String s -> Some (s :: acc)
```
rgrinberg, line 778: Same here. Should be `String s`

src/dune_rules/install.ml
```diff
@@ -429,44 +429,37 @@ let gen_install_file entries =
...
 let load_install_file path =
-  let open OpamParserTypes in
+  let open OpamParserTypes.FullPos in
   let file = Io.Untracked.with_lexbuf_from_file path ~f:Opam_file.parse in
-  let fail (fname, line, col) msg =
-    let pos : Lexing.position =
-      { pos_fname = fname; pos_lnum = line; pos_bol = 0; pos_cnum = col }
+  let fail { filename = pos_fname; start; stop } msg =
+    let position_of_loc (pos_lnum, pos_cnum) =
+      Lexing.{ pos_fname; pos_lnum; pos_bol = 0; pos_cnum }
```
rgrinberg, line 437: Should be `{ Lexing.pos_fname ; .. }`

src/dune_engine/package.ml
```diff
@@ -227,34 +231,35 @@ module Dependency = struct
     <|> let+ name = Name.decode in
         { name; constraint_ = None }
 
-  let rec opam_constraint : Constraint.t -> OpamParserTypes.value =
-    let nopos = Opam_file.nopos in
+  let nopos pelem = OpamParserTypes.FullPos.{ pelem; pos = Opam_file.nopos }
```
rgrinberg, line 234: `{ OpamParserTypes.FullPos.pelem; pos = Opam_file.nopos }`

Leonidas-from-XIV: Fixed and moved the whole function one up into the module, since I had it defined multiple times.

Leonidas-from-XIV:
This unfortunately doesn't solve #6767 because the latest release, 2.1.5, is older than https://github.com/ocaml/opam-file-format/pull/54. Maybe we can ask the OPAM team to release 2.1.6, as I would prefer to stick to vendoring releases.

PR #7330, opened 2023-03-16
test: guard melange tests behind an environment variable

anmonteiro:
as requested by @emillon

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -106,6 +106,8 @@ jobs:
         if: ${{ matrix.os == 'windows-latest' && matrix.skip_test == false }}
 
       - name: Run test suite on Unix
+        env:
+          DUNE_MELANGE_TEST: enable
         run: opam exec -- make test
         if: ${{ matrix.os != 'windows-latest' && matrix.skip_test == false }}
 
--- a/Makefile
+++ b/Makefile
@@ -111,7 +111,7 @@ test-coq: $(BIN)
 	DUNE_COQ_TEST=enable $(BIN) build @runtest-coq
 
 test-melange: $(BIN)
-	$(BIN) build @runtest-melange
+	DUNE_MELANGE_TEST=enable $(BIN) build @runtest-melange
 
 test-all: $(BIN)
 	$(BIN) build @runtest @runtest-js @runtest-coq @runtest-melange
--- a/flake.nix
+++ b/flake.nix
@@ -115,8 +115,8 @@
             })
             melange.overlays.default
           ];
-          mkSlim = { extraBuildInputs ? [ ] }:
-            pkgs.mkShell {
+          mkSlim = { extraBuildInputs ? [ ], ... }@attrs:
+            pkgs.mkShell ({
               nativeBuildInputs = testNativeBuildInputs;
               inputsFrom = [ pkgs.ocamlPackages.dune_3 ];
               buildInputs = testBuildInputs ++ (with pkgs.ocamlPackages; [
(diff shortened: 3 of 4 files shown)


rgrinberg:
How does this help? I think it just makes it easier to submit PR's and leave parts untested. Currently, the test suite works fine even if you don't have melange running.

anmonteiro:
I feel like this helps because:

- dune devs who don't want to install melange don't have to
- CI still runs the melange tests which will catch regressions for those unrelated PRs

rgrinberg:
It's easy enough not to run the melange tests already. If you don't have the melange tests installed, dune will just fail on them and move on.

anmonteiro:
I also made that argument when discussing this with Etienne, so I'll leave it to him to argue his case.

emillon:
The reasoning was the following. The "easy" workflow, `git clone; make dev-switch; make test` should have the test suite succeed for the "core" part of dune. This is not possible right now because melange needs a special setup for now. To fix this, we isolate this dependency, so that developers not working on melange do not have to get out of the easy workflow.

Let me expand a bit:
- it's important to have a quick path to "success", especially for new contributors. Stable dev instructions will help with this. We've seen several people ask for support related to these kind of errors recently.
- this kind of stable instructions also have benefits when bisecting `dune`: if intermediate versions are hard to test, `bisect` becomes impossible to use.
- We've done a similar thing with coq recently: building coq, especially the native version, is very slow, and its test suite takes a nontrivial amount of time. In addition, this support is fairly independent: it's unlikely that something not specific to coq will break the coq test suite or require a coq-specific fix. Even if that's the case, the CI will notice this.

So yes I think that isolating that (temporary) unstability with melange is important. I should add that if melange is installable through `opam` without pinning, this becomes a non-issue. `js_of_ocaml` is a similar test-dependency but it is not a problem because it is easy to install.

> It's easy enough not to run the melange tests already. If you don't have the melange tests installed, dune will just fail on them and move on.

It's difficult to determine that all tests pass except the melange ones. I think it's better to have a way to run all tests except the melange ones, which is what this PR is solving.

In terms of implementation, I think it would be preferable to run the tests automatically if melange is installed (in the `%{lib-available:melange}` sense), so that it's indeed harder to forget running tests, and in particular that it's not necessary to set extra variables in CI. But we should discuss this until once we have reached agreement about whether to bring this modularity or not to melange.

rgrinberg:
> this kind of stable instructions also have benefits when bisecting dune: if intermediate versions are hard to test, bisect becomes impossible to use.

When I use bisect, I manually pick the test I want to run. It would be very efficient to do something like `make test` for every commit.

> It's difficult to determine that all tests pass except the melange ones. I think it's better to have a way to run all tests except the melange ones, which is what this PR is solving.

I agree, it's just that selecting by disabling with an env var isn't a good way to do it. Whenever you try to run a test and it's disabled this way, it's hard to tell whether it's because the test passed or it didn't run.

How about we disable the tests in a different way. It would be enough to just not add the melange tests to the `runtest` alias. We already have the `alias` field in the `cram` stanza, but it's additive. We could add a different field such as `no_default_alias` and make it so that `runtest` isn't added by default

anmonteiro:
I suspect this won't be needed anymore for a few reasons:

- melange is going to be released imminently after dune 3.8 is merged on opam-repository
- melange works across 4.13 - 5.1 now that we use some ppxlib code to migrate AST versions

emillon:
@anmonteiro  yes I agree. let's wait until everything is released and the "normal" workflow (clone, make dev-switch, make test) works and I agree we can close this.

emillon:
Now that melange is released, this is not necessary anymore.

PR #7331, opened 2023-03-16
refactor(melange): propagate runtime_deps `loc` in Melange_rules

anmonteiro:
- some cleanup after both #7234   #7199 
- we still had a TODO to thread a `loc` for the `(runtime_deps (sandbox ...` error.

### DIFF
--- a/src/dune_rules/melange/melange_rules.ml
+++ b/src/dune_rules/melange/melange_rules.ml
@@ -248,21 +248,44 @@ let setup_emit_cmj_rules ~sctx ~dir ~scope ~expander ~dir_contents
   let* () = Buildable_rules.gen_select_rules sctx compile_info ~dir in
   Buildable_rules.with_lib_deps ctx compile_info ~dir ~f
 
-let raise_external_runtime_dep_error ~loc lib_name path =
-  User_error.raise ~loc
-    [ Pp.textf
-        "Public library %s depends on external path `%s'. This is not allowed."
-        (Lib_name.to_string lib_name)
-        (Path.to_string path)
-    ]
-    ~hints:
-      [ Pp.textf
-          "Move the external dependency to the workspace and use a relative \
-           path."
-      ]
-
 module Runtime_deps = struct
-  let targets ~output ~for_ deps =
+  type path_specification =
+    | Allow_all
+    | Disallow_external of Lib_name.t
+
+  let raise_disallowed_external_path ~loc lib_name path =
+    User_error.raise ~loc
+      [ Pp.textf
+          "Public library %s depends on external path `%s'. This is not \
+           allowed."
+          (Lib_name.to_string lib_name)
+          (Path.to_string path)
+      ]
+      ~hints:
+        [ Pp.textf
+            "Move the external dependency to the workspace and use a relative \
+             path."
+        ]
+
+  let eval ~loc ~expander ~paths:path_spec (deps : Dep_conf.t list) =
+    let runtime_deps, sandbox =
+      Dep_conf_eval.unnamed_get_paths ~expander deps
+    in
+    Option.iter sandbox ~f:(fun _ ->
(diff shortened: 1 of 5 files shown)


src/dune_rules/melange/melange_rules.mli
```diff
@@ -19,7 +19,9 @@ val setup_emit_js_rules :
...
   -> unit Memo.t
 
-val eval_runtime_deps :
-  expander:Expander.t -> Dep_conf.t list -> Path.Set.t Memo.t
+module Runtime_deps : sig
+  val eval :
+    expander:Expander.t -> loc:Loc.t -> Dep_conf.t list -> Path.Set.t Memo.t
 
-val raise_external_runtime_dep_error : loc:Loc.t -> Lib_name.t -> Path.t -> 'a
+  val raise_disallowed_external_path : loc:Loc.t -> Lib_name.t -> Path.t -> 'a
```
rgrinberg, line 26: We use `Nothing.t` instead of a polymorphic value for such a type.

src/dune_rules/melange/melange_stanzas.ml
```diff
@@ -11,7 +11,7 @@ module Emit = struct
     ; libraries : Lib_dep.t list
     ; package : Package.t option
     ; preprocess : Preprocess.With_instrumentation.t Preprocess.Per_module.t
-    ; runtime_deps : Dep_conf.t list
+    ; runtime_deps : Loc.t * Dep_conf.t list
```
rgrinberg, line 14: The location should be come from the atom that introduces the sandbox mode. I guess this is still an improvement for now.

rgrinberg (approved):

src/dune_rules/melange/melange_stanzas.ml
anmonteiro: I can look into that separately, but having experimented with it in the past, I found that carrying a loc for all elements in the list was a bit heavy for no significant return.

rgrinberg:
Looking at the code, I find it suspect that `Runtime_deps.eval` returns a value that later needs to be validated separately. I think it should just take a parameter such as `~paths:[ Allow_all | No_external ]` and make it impossible for users outside the module to even access a non validated set.

anmonteiro:
I like that suggestion. Did something in [c731484](https://github.com/ocaml/dune/pull/7331/commits/c7314848e91830750273b023057eef1870d547bb)

src/dune_rules/melange/melange_rules.ml
```diff
@@ -248,21 +248,45 @@ let setup_emit_cmj_rules ~sctx ~dir ~scope ~expander ~dir_contents
...
+    let open Memo.O in
+    let+ paths, _ = Action_builder.run runtime_deps Eager in
+    (match path_spec with
+    | Allow_all -> ()
+    | Disallow_external lib_name ->
+      Path.Set.iter paths ~f:(fun path ->
+          match Path.as_external path with
+          | None -> ()
+          | Some _ -> raise_disallowed_external_path ~loc lib_name path));
+
```
rgrinberg, line 285: extra space

src/dune_rules/melange/melange_rules.mli
```diff
@@ -19,7 +19,15 @@ val setup_emit_js_rules :
   -> Melange_stanzas.Emit.t
   -> unit Memo.t
 
-val eval_runtime_deps :
-  expander:Expander.t -> Dep_conf.t list -> Path.Set.t Memo.t
+module Runtime_deps : sig
+  type path_specification =
+    | Allow_all
+    | Disallow_external of Lib_name.t
```
rgrinberg, line 25: If you want to be fancy, you could make this type:

```ocaml
type _ path_specification =
  | Allow_all : Path.Set.t path_specification
  | Disallow_external : Path.Build.Set.t path_specification
```

And then use it to type the return value

```ocaml
val eval : 'a path_specification -> 'a Memo.t
```

PR #7333, opened 2023-03-16
refactor: rename dune_config

rgrinberg:
Rename dune_config to dune_config_file. We can still access the old
names by opening Dune_config.

In preparation for introducing a dune_config library that will introduce
a [Dune_config.Config] module.


rgrinberg:
First step to introduce a general dune config library. The library name `dune_config` needs to be reserved for general configuration and not just the handling of the `dune-config` file.

snowleopard (approved):
Looks good, thanks!

Alizter (approved):

PR #7208, opened 2023-03-02
fix(rules): don't descend into automatic subdirs infinitely

anmonteiro:

PR #7334, opened 2023-03-16
feat(alias_rec): look up aliases in all directories, not just source dirs 

anmonteiro:
- Previously, `Action_build.dep_on_alias_rec` would traverse all the directories in _source_ and check whether an alias was defined for those directories
- Rules might, however, introduce aliases in directories that will be created by the build process. This patch traverses all the directories dune is aware of, including automatic sub-directories, and build-dir-only subdirectories, when looking for aliases

This implementation is only possible after #7207 and #7208 (and was, in fact, what prompted the discovery of both bugs), since dune would return the same set of automatic sub directories at any level.


rgrinberg:
Looks OK. We need to preserve the old behavior for 3.7 and below.

Can you move the implementation to the rules?

anmonteiro:
Added in [80c7f72](https://github.com/ocaml/dune/pull/7334/commits/80c7f72b0ce41af3431a4d2b54bb799024c17587):
- Moved the alias functions to the rules
- guarded the feature behind syntax version 3.8

src/dune_rules/alias_rec.mli
```diff
@@ -0,0 +1,15 @@
...
+
+module Alias_status : sig
+  type t =
+    | Defined
+    | Not_defined
+
+  include Monoid.S with type t := t
+end
+
+val dep_on_alias_rec :
```
jchavarri, line 11: Why not bringing over the previously existing comments / docs?

src/dune_rules/alias_rec.ml
```diff
@@ -1,23 +1,122 @@
...
+        { Lookup_alias.alias_exists = Not_defined
+        ; allowed_subdirs = Filename.Set.empty
+        }
+
+  let dep_on_alias_rec name dir =
+    let rec map_reduce dir ~f =
+      let open Action_builder.O in
+      let* should_traverse =
+        match Path.Build.drop_build_context dir with
+        | None -> Action_builder.return true
```
jchavarri, line 97: Shouldn't dune error out in this case? Unlike the `| None -> true` branch in line 104, in this case it has not been possible to remove the `_build` prefix. I fail to see in which cases this would be expected.

src/dune_rules/alias_rec.ml
```diff
@@ -1,23 +1,122 @@
...
+  include T
+  include Monoid.Make (T)
+end
+
+module Source_tree_only = struct
+  let dep_on_alias_if_exists alias =
+    let open Action_builder.O in
+    let* definition = Action_builder.of_memo (Load_rules.alias_exists alias) in
+    match definition with
+    | false -> Action_builder.return Alias_status.Not_defined
```
jchavarri, line 27: ~If I understand correctly, this allows to depend on aliases that don't exist? I wonder if this should be an error.~ Edit: never mind, i see we accumulate aliases statuses and error out at the end if nothing has been found.

jchavarri (approved):
Looks great, thanks. I just have a couple of questions.

src/dune_rules/alias_rec.ml
anmonteiro: This means it’s not in source, and we should traverse it

jchavarri:
Maybe `runtime_deps` and `melange.runtime_deps` fields should be modified so they are only allowed for 3.8+ versions of the dune language. Otherwise dune should emit a user error.

The current behavior at the moment in 3.7 is all rules generated by these fields are ignored silently, unless the target paths are created in source as well. Which is very confusing.

jchavarri:
I tested this branch more thoroughly, and observed the build times have increased quite a bit (~31% slower).

Build using Dune `main` branch:

```
$ time dune build @fe.all

real	0m58.579s
user	10m0.066s
sys	6m31.861s
```

With this branch:

```
$ time dune build @fe.all

real	1m16.244s
user	10m36.046s
sys	6m16.956s
```

I then checked the folders dune is loading using the `--debug-load-dir` flag. I could see that a lot of folders are now being checked as opposed to before, as expected, but some of them seem kind of "private", like:
```
_build/default/backend/admin/.cluster.objs/jsoo/use-js-string+effects/.formatted
_build/default/backend/fmt_read/.utop/.utop.eobjs/native/.bin
_build/default/backend/admin/.merlin-conf
```

I wonder if there are ways to filter these private folders somehow?

I also noticed that folders that dune should ignore as I'm using `data-only-dirs` are now evaluated, which also contributes to larger rule evaluation time.

I am not sure we want to penalize all dune users with longer build times to support a feature for Melange 😬  Which makes me wonder if there might be some alternate solution to Melange `runtime_deps` rules, where instead of attaching the rules to each individual dir, we attach them to the dir where the melange stanza exists? Would this be possible? This approach would remove the need of the changes added by this PR, and also allow using non-recursive aliases to build Melange projects, with the subsequent build speed gains as Dune doesn't have to walk whole tree to look for the alias.

If the decision is to move forward with the PR, it'd be worth checking the [benchmarks](https://ocaml.github.io/dune/dev/bench/) once it's merged.

rgrinberg:
Okay how about we just keep it simple for now and add this custom alias functionality only to `melange.emit` directories. It should be possible to walk all the directories with .js files starting from the `output` directory of `melange.emit` and force the aliases there.

anmonteiro:
In my latest commit:

- we preserve the same behavior as before for recursive aliases
- in addition: if we find that a directory defines a `melange.emit` rule, we fully traverse it and its sub-directories to find any nested `runtime_deps` rules

test/blackbox-tests/test-cases/melange/public-library-with-runtime-deps.t
```diff
@@ -58,7 +57,6 @@ Test `melange.runtime_deps` in a public library in the workspace
   > let () = Js.log (Foo.read_asset ())
   > EOF
 
-  $ mkdir -p output
   $ dune build @mel
```
jchavarri, line 60: Could include a test to show what happens with non-recursive aliases please?

jchavarri (changes requested):

jchavarri (approved):
Talked about this offline with @anmonteiro. It seems most of the additional time the build takes is being spent recursively copying the content from node_modules folders that were symlinked before. We tried to use either `glob_files_rec` and `source_tree` to define these deps but it didn't seem to affect performance much. We also discussed about the possibility to include something like `melange.runtime_deps (symlink node_modules)` in the future to alleviate some of the work dune has to do.

Re: the inclusion of data_only_dirs, there's no way around visiting the build folders, because otherwise this PR wouldn't make sense (runtime_deps would not be copied).

As this PR unblocks runtime_deps usage at Ahrefs and we can start defining builds of subsets of the codebase, I'd suggest to merge it and try to tackle the performance impact in future PRs.

jchavarri:
Still trying to fully understand this 😅 

Before having `runtime_deps`, the approach we were using to copy assets was something like the following:

Say there is a lib `foo` that needs some runtime assets placed `foo/assets`.

`foo/dune`:
```
(data_only_dirs assets)

(library
 (name foo)
 (modes melange))
```

`dune`:
```
(melange.emit
 (alias mel)
 (target output)
 (libraries foo)
 (module_systems es6))

(subdir
 output/foo/assets
 (alias
  (name mel)
  (deps some_dep.txt)))
 (copy_files
  (files
   %{project_root}/foo/assets/some_dep.txt))
```

I am not sure I understand why the above worked without dune having to look up aliases in build dirs? Is it because Dune would interpret the `subdir` rules as rules in source dirs, and not build dirs, because they are written by the user and not generated after the fact?

anmonteiro:
> Is it because Dune would interpret the subdir rules as rules in source dirs,.. ?

yes

jchavarri:
@rgrinberg gentle ping about this one. It makes `runtime_deps` usable, so it's a significant improvement for us and other Melange users (thanks @anmonteiro !). 

The downside, as we discussed, is the performance hit. Even after narrowing down to `melange.emit` directories, in our melange builds dune has to walk a _lot_ more folders, compare trying to run a recursive, non-existent alias in this branch:

```
$ time dune build @doesntexist
Error: Alias "doesntexist" specified on the command line is empty.
It is not defined in . or any of its descendants.

real	0m14.426s
user	0m13.659s
sys	0m0.760s
```

vs `main`:

```
$ time dune build @doesntexist
Error: Alias "doesntexist" specified on the command line is empty.
It is not defined in . or any of its descendants.

real	0m3.565s
user	0m3.207s
sys	0m0.376s
```

In any case, my vote fwiw is that we merge this PR and work on performance in a future one, with features like symlinking mentioned in #7379 to reduce the number of folders that end up in `_build` in our project.

jchavarri:
re: performance, reminder that we can see how melange benchmark changes in https://ocaml.github.io/dune/dev/bench/ —after merging to `main`—, or add this branch name to gh workflow yml to get metrics while the PR is opened:

https://github.com/ocaml/dune/blob/fe3c98fae85de5542534f5e7e96a9d12f0838da4/.github/workflows/bench.yml#L6-L8

src/dune_rules/melange/melange_stanzas.mli
```diff
@@ -22,6 +22,8 @@ module Emit : sig
   type Stanza.t += T of t
 
   val decode : t Dune_lang.Decoder.t
+
+  val target_dir : t -> dir:Path.Build.t -> Path.Build.t
```
rgrinberg, line 26: Is this change needed for something?

src/dune_engine/load_rules.ml
```diff
@@ -707,7 +707,7 @@ end = struct
                don't check directory targets as these are already checked
                earlier. *)
             (match
-               String.Set.choose
+               Filename.Set.choose
```
rgrinberg, line 710: Could you move unrelated changes to separate commits/prs?

rgrinberg:
> Moved the alias functions to the rules

Sorry, my earlier comment wasn't clear at all. What i meant is to move to only the new implementation of recursive aliases to the melange rules. The old implementation can stay where it was before.

src/dune_rules/melange/melange_stanzas.mli
anmonteiro: This is used in `alias_rec.ml` and it'd create a conflict if it remained in `melange_rules`

anmonteiro: Would you like me to move this change to a separate PR?

src/dune_rules/dep_conf_eval.ml
```diff
@@ -27,30 +27,19 @@ let to_action_builder = function
   | Other x -> x
 
 let dep_on_alias_rec alias ~loc =
-  let src_dir = Path.Build.drop_build_context_exn (Alias.dir alias) in
-  Action_builder.of_memo (Source_tree.find_dir src_dir) >>= function
```
rgrinberg, line 31: Again, I think this combines general refactoring and adding a new feature which makes it hard to review. I would appreciate it if you did the refactoring separately (before or after the new feature is added).

anmonteiro: This is not a refactoring. We can't check whether the source tree exists here because it could be a melange-target-dir entrypoint that we're checking.

src/dune_rules/alias_rec.ml
```diff
@@ -1,23 +1,129 @@
...
+      let* { Lookup_alias.alias_exists; allowed_subdirs } = f dir in
+      Action_builder.List.fold_left (String.Set.to_list allowed_subdirs)
+        ~init:alias_exists ~f:(fun alias_exists s ->
+          let+ alias_exists' = map_reduce (Path.Build.relative dir s) ~f in
+          Alias_status.combine alias_exists alias_exists')
     in
+    map_reduce dir ~f:(fun dir -> dep_on_alias_if_exists (Alias.make ~dir name))
+end
+
+let rec map_reduce ~build_dir ~name t ~f =
```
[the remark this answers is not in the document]
anmonteiro: Because we need to visit directories that `Source_tree` doesn't find.

src/dune_rules/melange/melange_stanzas.mli
rgrinberg: If the feature cannot be implemented without it, then we can let it stay

src/dune_rules/alias_rec.ml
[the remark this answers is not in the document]
rgrinberg: Because we need to visit the build path, right?

```
    let build_path =
      Path.Build.append_source build_dir (Source_tree.Dir.path t)
```

Is it possible to use the old traversal and just "traverse" the extra build directory in `f`?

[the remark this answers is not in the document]
anmonteiro: let me try to make that work.

src/dune_rules/dep_conf_eval.ml
```diff
@@ -27,30 +27,19 @@ let to_action_builder = function
...
-  | None ->
-    Action_builder.fail
-      { fail =
-          (fun () ->
-            User_error.raise ~loc
-              [ Pp.textf "Don't know about directory %s!"
```
rgrinberg, line 37: Is this error impossible now?

anmonteiro:
Thanks for the review. In my latest push:

- introduced a `Lookup_alias` module in `Action_builder`
  - returns the allowed build-only subdirs in addition to the alias status
- moved the `Alias_rec` functor back to `Action_builder`
- moved the melange target folders handling to the traverse callback
- reverted the changes in `Dep_conv_eval`, which are not necessary for this PR, and can be discussed separately

rgrinberg:
> introduced a Lookup_alias module in Action_builder
returns the allowed build-only subdirs in addition to the alias status

This change has to go in a separate PR because it changes the engine. For now, you cant revert it here and re-implement it in the rules with `Action_builder.of_thunk` (Along with Alias_rec).

src/dune_rules/alias_rec.ml
```diff
@@ -1,4 +1,22 @@
 open Import
+module Alias_status = Action_builder.Alias_status
+module Lookup_alias = Action_builder.Lookup_alias
+
+module In_melange_target_dir = struct
+  let dep_on_alias_rec =
+    let rec map_reduce dir ~f =
```
rgrinberg, line 7: The name `map_reduce` is a bit weird here. It just looks like `reduce` to me without any mapping.

rgrinberg (approved):
> I copied the Alias_rec functor faithfully from the engine, though it's not really needed in the rules, it could just be inlined. @rgrinberg what do you prefer here?

It's okay to leave it as is. Just leave a comment as to why this implementation differs from `Action_builder`.

I've left some minor comments. Feel free to merge once they're addressed.

src/dune_rules/alias_rec.ml
```diff
@@ -1,21 +1,116 @@
...
+end
+
+module In_melange_target_dir = struct
+  let dep_on_alias_rec =
+    let rec map_reduce dir ~f =
+      let open Action_builder.O in
+      let* { Lookup_alias.alias_status; allowed_build_only_subdirs } = f dir in
+      (* TODO there should be traversals that don't require this conversion *)
+      Filename.Set.to_list allowed_build_only_subdirs
+      |> Action_builder.List.fold_left ~init:alias_status
```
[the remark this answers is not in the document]
anmonteiro: Would you like me to address that in a follow-up PR?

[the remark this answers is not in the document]
rgrinberg: Sure. There's no rush though, the issue has been here forever.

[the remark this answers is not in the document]
anmonteiro: Sorry, now I'm more confused. Are you talking about `Action_builder.List.fold_left` or this melange-specific traversal?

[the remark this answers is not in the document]
rgrinberg: I was talking about `Action_builder.List.fold_left`.

rgrinberg:
Please squash the commits before merging btw.

Issue #7335, opened 2023-03-16
Running "dune build @install" in a workspace sub-project puts files in an unexpected directory


rlepigre:
## Expected Behavior

When I work in a dune workspace containing several dune projects, I expect all produced files (barring promoted ones) to end up in `<workspace_root>/_build`.

## Actual Behavior

This does not seem to be the case when using `dune build @install` from a project subdirectory: files end up in a `_build` placed under project's root (instead of the workspace root).

## Reproduction

See ~~https://github.com/ocaml/dune/pull/7336~~ [install-workspace.tar.gz](https://github.com/ocaml/dune/files/11360984/install-workspace.tar.gz).

emillon:
Thanks. The title mentions `dune install` but the body `dune build @install`. This is about the latter, right?

rlepigre:
> Thanks. The title mentions `dune install` but the body `dune build @install`. This is about the latter, right?

Indeed, sorry, I fixed the title.

Alizter:
@rlepigre as explained in #7336 this is the intended behaviour since `-p` will set the root to the current directory. As I explained in the PR, the original cram test was incorrect due to the setting of INSIDE_DUNE. I also mention why unsetting that in fact gives the desired behaviour. All I can add to this is "don't use -p unless you are in the project root or your name is opam".

@emillon maybe it is worth adding some documentation about this difference (the fact that -p is setting the root). However I can only find documentation about -p in the help page rather than in the manual.

rlepigre:
OK, I see. This behaviour sounds a bit weird to me, but I understand this cannot be changed at this point. Thanks!

Alizter:
@rlepigre Do you think we should close this issue now that some documentation has been added. Is there anything else to document?

rlepigre:
> @rlepigre Do you think we should close this issue now that some documentation has been added. Is there anything else to document?

Yes, sure, feel free to close.

PR #7336, opened 2023-03-16
Add repro for #7335.

rlepigre:


rgrinberg:
I think you're forgetting that root discovery is disabled inside the cram tests.

rlepigre:
> I think you're forgetting that root discovery is disabled inside the cram tests.

I did not know about that when I set up this test, but the issue is real as this is a minimized example of something I came upon in a real project.

rlepigre:
I don't really see the poing of putting that in a separate repo, but I can easily put that in a tarball that you guys can use to test. It has basically the same contents as the test. This is still broken in `main`.

Here it is: [install-workspace.tar.gz](https://github.com/ocaml/dune/files/11360984/install-workspace.tar.gz).

It contains a `README.md` which explains the steps to reproduce the various problems.

Alizter:
@rlepigre OK I can reproduce the issue, it definitely looks like something fishy is going on when using -p. I think I should be able to modify your cram test to reproduce.

Alizter:
The bug here is that when -p is used, the workspace detection behaviour seems to change. The behaviours of dune build and dune clean are however intended. You can `unset INISDE_DUNE` for this cram test since there is a dune-workspace file that will be chosen as the root. (by build and clean).

Alizter:
I'm going to overtake this PR.

Alizter:
So for the record, here is what the test looks like from before when unsetting INSIDE_DUNE:
```
The setup for this test is two nested dune projects:
- a project at the root of the workspace (defining package "test1",
- a project in the "dep" folder defining package "test2".

When we prepare installation for the "test2" package from the workspace root,
the installation files are produced under "_build/install".

  $ unset INSIDE_DUNE

  $ dune build --display=quiet -p test2 @install
  $ find . -name _build
  ./_build
  $ dune clean

If we do the same from the "dep" directory, we expect the same result, but the
files are rather produced in "dep/_build/install".

  $ (cd dep && dune build --display=quiet -p test2 @install)
  $ find . -name _build
  ./dep/_build

Calling "dune clean" at the workspace root does not clean the nested "_build"
directory that was produced by the previous call to "dune build".

  $ dune clean
  $ find . -name _build
  ./dep/_build

  $ (cd dep && dune clean)
  Entering directory '$TESTCASE_ROOT'
  Leaving directory '$TESTCASE_ROOT'
  $ find . -name _build
  ./dep/_build

Similarly, calling "dune build @install" in the "dep" folder only prepares the
installation of the packages under "dep", not all packages of the workspace.
The resulting files end up again under "dep/_build".

  $ (cd dep && dune build @install)
  Entering directory '$TESTCASE_ROOT'
  Leaving directory '$TESTCASE_ROOT'
  $ find . -name _build
  ./dep/_build
  ./_build
```

Alizter:
The reason -p is getting the wrong root by the way, is that it is a command line alias for:
```
--release --only-packages
```
and then release is super fun:
```
          [ "--root"
          ; "."
          ; "--ignore-promoted-rules"
          ; "--no-config"
          ; "--profile"
          ; "release"
          ; "--always-show-command-line"
          ; "--promote-install-files"
          ; "--require-dune-project-file"
          ; "--default-target"
          ; "@install"
          ]
```
I am not very happy with --release's behaviour. At the very least it should be better documented. The root setting is kind of stupid IMO. @rgrinberg WDYT about the fact that --release (and by extension -p) is setting the root to the current directory?

rgrinberg:
It can't be changed at this point. Too many released packages rely on it. So it sounds like this bug was from improper use of `-p`.

PR #7343, opened 2023-03-17
Framework for running multiple monorepo benchmark scenarios

gridbugs:
Based on some feedback on the monorepo benchmarks from @snowleopard we're going to add multiple scenarios for benchmarking the monorepo. This changes the benchmark runner script to make it possible to run multiple benchmarks in sequence and demonstrates running an initial build from scratch and a second build after making no changes to the monorepo, printing a current-bench json object with the results.

Putting this up for review in its current state to get early feedback on my approach. The plan is to add additional scenarios that apply patches to change some of the packages in the monorepo and measure the time it takes to rebuild which I'll do in a subsequent PR.

### DIFF
--- a/bench/monorepo/.ocamlformat
+++ b/bench/monorepo/.ocamlformat
@@ -1,0 +1,12 @@
+version=0.24.1
+profile=conventional
+ocaml-version=4.08.0
+break-separators=before
+dock-collection-brackets=false
+doc-comments=before
+let-and=sparse
+type-decl=sparse
+cases-exp-indent=2
+break-cases=fit-or-vertical
+parse-docstrings=true
+module-item-spacing=sparse
--- a/bench/monorepo/bench.Dockerfile
+++ b/bench/monorepo/bench.Dockerfile
@@ -137,10 +137,10 @@ RUN opam install -y opam-monorepo ppx_sexp_conv ocamlfind ctypes ctypes-foreign 
 RUN mkdir -p $HOME/monorepo-bench
 
 # Download the monorepo benchmark and copy files into the benchmark project
-ENV MONOREPO_BENCHMARK_TAG=2023-03-17.1
+ENV MONOREPO_BENCHMARK_TAG=2023-03-21.0
 RUN wget https://github.com/ocaml-dune/ocaml-monorepo-benchmark/archive/refs/tags/$MONOREPO_BENCHMARK_TAG.tar.gz -O ocaml-monorepo-benchmark.tar.gz && tar xf ocaml-monorepo-benchmark.tar.gz && mv ocaml-monorepo-benchmark-$MONOREPO_BENCHMARK_TAG ocaml-monorepo-benchmark
 WORKDIR $HOME/monorepo-bench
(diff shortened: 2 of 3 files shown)


snowleopard:
Just an organisational comment: I would expect any non-trivial code related to benchmarking to go to the benchmarking repo. With all the features that we want, it will likely grow quite a bit (for example, what we have internally contains a bunch of modules with some internal APIs, some command-line interface, etc.).

gridbugs:
@snowleopard I have added some patches to https://github.com/ocaml-dune/ocaml-monorepo-benchmark/tree/main/benchmark/bench-patches which exercise various scenarios we'd like to benchmark. Within the dune repo I'm going to have a small program which applies each patch in some order and benchmarks running `dune build` in between each change, producing output compatible with current-bench.

snowleopard:
@gridbugs One gotcha we noticed with applying actual patches is that commands like `patch` would generate a bunch of file-system events in the same directory, which will likely look quite different to Dune compared to how a user would edit those files: i.e. instead of seeing a `source.ml` modified, Dune might see some temporary files created in the same directory first.

I think it may be less confusing to avoid using `patch` and modify files using some ad-hoc method.

gridbugs:
> i.e. instead of seeing a `source.ml` modified, Dune might see some temporary files created in the same directory first.

Are you saying that when a patch is applied dune might see some temporary files created by the `patch` program? Or are you saying that when editing a file with a text editor, the editor can create temporary files which are visible to dune?

Is this only relevant when dune is being run in watch mode? The benchmarks currently use multiple invocations of `dune build` and don't run dune in watch mode.

snowleopard:
> Are you saying that when a patch is applied dune might see some temporary files created by the `patch` program?

Yes. You can `strace` it to see what's actually going on.

> Or are you saying that when editing a file with a text editor, the editor can create temporary files which are visible to dune?

I suppose some editors may be creating temporary files in the same directory, though at least with VSCode, I think, Dune is receiving only one file-system event when I save a file.

> Is this only relevant when dune is being run in watch mode?

Indeed. Though I suppose it makes sense to use the same mechanism when benchmarking both watching and non-watching scenarios, for the sake of simplicity.

gridbugs:
Can you clarify your requirements for benchmarking dune in watch mode. I wasn't aware that that's something we wanted to do for this project.

snowleopard:
All this is to say: we want to have some control/understanding over how exactly the file system is modified. Otherwise, we might end up benchmarking and optimising code paths that aren't relevant for users.

gridbugs:
Rather than using `patch` to update the files in the tree directly, would it be reasonable to use `cp` to overwrite files in the tree? Is that what you had in mind when you said:
> modify files using some ad-hoc method.

And can you please clarify, does the Plat178 project involve benchmarking dune in watch mode? I'm happy to make this change regardless so that if we ever add benchmarks for watch mode we're not accidentally triggering it with a bunch of temporary files.

snowleopard:
> Can you clarify your requirements for benchmarking dune in watch mode. I wasn't aware that that's something we wanted to do for this project.

Well, our goal is to make Dune fast for users, so we want to benchmark use-cases that matter. Build from scratch and null builds surely matter. But so do file-watching builds: for many users, running `dune build -w` directly or via an editor is a pretty common (if not most common) usage scenario.

See https://github.com/ocaml/dune/issues/5549 for an example of a performance issue that having a large realistic benchmark could help with.

snowleopard:
> Rather than using `patch` to update the files in the tree directly, would it be reasonable to use `cp` to overwrite files in the tree? Is that what you had in mind when you said:
> 
> > modify files using some ad-hoc method.

As I mentioned before, the benchmarking logic/scenarios are likely to get quite complex with time, so it makes sense to implement them using OCaml. You could, of course, call `cp` from OCaml but I wouldn't want to prescribe a specific implementation approach upfront. It's fine to start with whatever is easiest, to get things rolling, but then I would recommend taking some time to think through and evaluate possible options, including `patch`, `cp` and friends.

Issue #7344, opened 2023-03-17
The "dune coq top" is unreliable when not called from the workspace root


rlepigre:
## Expected Behavior

I would expect that whenever `dune coq top` is invoked from wherever in a dune workspace, with a relative path to a file that is part of a stanza in the workspace, then it simply works.

## Actual Behavior

Currently `dune coq top` seems to only really work well when invoked from the workspace root. That is probably my fault, since I wrote the initial implementation.

Note that similarly to https://github.com/ocaml/dune/issues/7335, it may happen that invoking `dune coq top` in a sub-project of a workspace create a `_build` folder in a subdirectory (instead of at the dune workspace root).

## Reproduction

See MR (will update).

rlepigre:
CC @Alizter @ejgallego

Alizter:
In fact this is an issue for a regular OCaml project too. Calling the build command within a sub directory means it is unable to detect the root. The workaround for `dune coq top` and in general is to explicitly pass the `--root` option.

Alizter:
I am reminded that `INSIDE_DUNE` is present in cram tests which is why the root is not detected by dune coq top. I suspect that passing `--root` explicitly will fix this issue as I don't think we are doing anything vastly different to dune build when it comes to root detectino.

Alizter:
@rlepigre Can you confirm that this is still an issue outside of the cram tests?

rlepigre:
The issue is definitely real, I experienced the problem in an actual Coq project. I wrote the cram test to more easily experiment with the `dune coq top` implementation to try and solve the problem myself. I'll run the test outside of a cram test and report on the actual results (I planned good coverage in the cram test, but did not try all combinations in a project).

rlepigre:
Here you go: [7344.tar.gz](https://github.com/ocaml/dune/files/11015752/7344.tar.gz). This contains the same setup as in the cram test, but as a shell script.
Here is the output you get when running the script (assuming you setup the path to `dune.exe` according to your setup):
```
Running toplevel on [file1.v] from [.]
Running toplevel on [coq_dir1/file2.v] from [.]
Running toplevel on [coq_dir1/dir1/file3.v] from [.]
Running toplevel on [project/file4.v] from [.]
Running toplevel on [project/coq_dir2/file5.v] from [.]
Running toplevel on [project/coq_dir2/dir2/file6.v] from [.]
Running toplevel on [../file1.v] from [coq_dir1]
[...]

PR #7348, opened 2023-03-17
test: show running `dune build` in subdir fails to find project root

Alizter:
Running `dune build` in a subdirectory fails to find the project root. This is the underlying cause of #7344.


rgrinberg:
root finding is disabled in tests because of INSIDE_DUNE

Alizter:
Of course it is. I keep forgetting this.

PR #7349, opened 2023-03-17
feature: introduce dune_config

rgrinberg:
[dune_config] is a library to help developers configure dune in a
consistent way. It does not aim to replace dune's existing configuration
capabilities which are meant to be stable and user facing. The type of
configuration it plans to support is experimental, undocumented, and may
be removed without notice.

For ease of use, it's possible to set the configuration values using
environment vairables and the dune-config file. the former will take
precedence over the latter.

As a demonstration, I've added a single configuration value for
disabling dune's global lock. This isn't a safe thing to do, but I know
that @snowleopard has his own locking implementation at JST and
@jchavarri needs it for a particular workflow.


rgrinberg:
I haven't yet added a way to read these values from `dune-config`. Will do so momentarily

Alizter:
Should we name the internal config values `DUNE_INTERNAL_CONFIG` so that users don't think they have come across an undocumented config value?

rgrinberg:
It is an undocumented config value. So that's exactly what they've come across. We will actually document that this prefix is reserved for experimental options. So I don't see a need to add `INTERNAL` to the name. Although I'm not opposed to it either

rgrinberg:
There's an existing `Dune_util.Config` module that contains a weird collection of things:

```
val dev_null : Path.t

(** [dev_null] opened in read mode *)
val dev_null_in : Unix.file_descr Lazy.t

(** [dev_null] opened in write mode *)
val dev_null_out : Unix.file_descr Lazy.t

(** Are we running inside an emacs shell? *)
val inside_emacs : bool

(** Are we running inside Dune? *)
val inside_dune : bool

(** Are we running in CI?. This checks the CI environment variable which is
    supported by travis, gitlab.*)
val inside_ci : bool
```

The things in it have nothing to do with user configuration, and fall into two categories:

1. General portability concerns. Those should be moved into stdune
2. Various environment variable that dune uses to determine the type of environment it's being executed in. They're not meant to be set by the user. I'm thinking about moving those into an `Exec_env` module in `Dune_util`.

I'll address this in a subsequent PR.

src/dune_config/config.ml
```diff
@@ -0,0 +1,51 @@
...
+
+let initialized = ref false
+
+type 'a t =
+  { name : string
+  ; of_string : string -> ('a, string) result
+  ; mutable value : 'a
+  }
+
+let env_name t = sprintf "DUNE_CONFIG__%S" (String.uppercase_ascii t.name)
```
jchavarri, line 11:
```suggestion
let env_name t = sprintf "DUNE_CONFIG__%s" (String.uppercase_ascii t.name)
```

src/dune_config/config.mli
```diff
@@ -0,0 +1,15 @@
...
+type 'a t
+
+(** [get t] return the value of the configuration for [t] *)
+val get : 'a t -> 'a
+
+(** should dune acquire the global lock before building *)
+val global_lock : [ `Enabled | `Disabled ] t
+
+(** Before any configuration value is accessed, this function must be called *)
+val init : unit -> unit
```
jchavarri, line 15: Would it make sense to have this type be:

```
val init : unit -> t
```

? 

It would require passing config through functions, but it'd be more future proof and explicit than having the side effect hidden.

jchavarri (approved):
Thanks! Works great, there is just this small issue with env vars quotes.

src/dune_config/config.mli
rgrinberg: For some of the places we want to use this it would be a bit too intrusive. Plus, many of these experimental options are not going to be permanent, so it would be a shame to mangle the code just to accommodate them.

rgrinberg:
@jchavarri should be ready now

src/dune_config/config.ml
```diff
@@ -0,0 +1,69 @@
...
+  initialized := true
+
+let global_lock =
+  let t =
+    { name = "global_lock"
+    ; of_string =
+        (function
+        | "enabled" -> Ok `Enabled
+        | "disabled" -> Ok `Disabled
+        | _ -> Error (sprintf "only %S and %S are allowed" "enabled" "diabled"))
```
jchavarri, line 64:
```suggestion
        | _ -> Error (sprintf "only %S and %S are allowed" "enabled" "disabled"))
```

src/dune_config/dune
```diff
@@ -0,0 +1,5 @@
+(library
+ (name dune_config)
+ ;; careful with adding dependencies here
+ ;; anything that depends on this library cannot itself be configurable
```
jchavarri, line 4: I don't understand this comment, considering the one in the line above. Shouldn't it say "anything that this library depends on"?

jchavarri (approved):
Looks great, tested and works fine. Thanks!

jchavarri:
Some test need to be updated it seems.

rgrinberg:
@snowleopard Gonna merge to unblock myself and others. I'm not particularly concerned with how we do the configuration as long as it exists. We can change the configuration module however you need once you're back.

PR #7350, opened 2023-03-17
add: Path.as_outside_build_dir

Alizter:
We already had Path.as_outside_build_dir_exn so we add the option variant so we can decide what to do with it.

This is used in Coq's installed composition, specifically due to a use of Fs_memo.dir_contents and adding a nice user message if _build/ was somehow passed.

### DIFF
--- a/otherlibs/stdune/src/path.ml
+++ b/otherlibs/stdune/src/path.ml
@@ -950,6 +950,11 @@ let as_in_source_tree_exn t =
       "[as_in_source_tree_exn] called on something not in source tree"
       [ ("t", to_dyn t) ]
 
+let as_outside_build_dir : t -> Outside_build_dir.t option = function
+  | In_source_tree s -> Some (In_source_dir s)
+  | External s -> Some (External s)
+  | In_build_dir _ -> None
+
 let as_outside_build_dir_exn : t -> Outside_build_dir.t = function
   | In_source_tree s -> In_source_dir s
   | External s -> External s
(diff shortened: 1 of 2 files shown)


otherlibs/stdune/src/path.ml
```diff
@@ -950,6 +950,11 @@ let as_in_source_tree_exn t =
       "[as_in_source_tree_exn] called on something not in source tree"
       [ ("t", to_dyn t) ]
 
+let as_outside_build_dir : t -> Outside_build_dir.t option = function
+  | In_source_tree s -> Some (In_source_dir s)
+  | External s -> Some (External s)
+  | In_build_dir _ -> None
+
 let as_outside_build_dir_exn : t -> Outside_build_dir.t = function
```
anmonteiro, line 958: this can be implemented in terms of `as_outside_build_dir` now.

Alizter: I followed the implementation of the other `as_`s which also don't implement in terms of eachother.

anmonteiro: I only looked at `as_in_source_tree_exn` which is right above your function, and it's implemented in terms of the other one.

Alizter: `as_in_source_tree` loses the content of the `In_build_dir` constructor which we include in the code error so we cannot use it for `as_in_source_tree`.

rgrinberg:
The function is fine but can you add it along with where you're going to use it? The `Path` module isn't really general purpose so there's no need to extend its api beyond what we're going to use ourselves. Also, if you're only going to use it once, a pattern match would suffice as well.

Alizter:
OK

PR #7351, opened 2023-03-18
rpc: error out any requests to build if server is running in eager mode

### LINKED ISSUE #7314: `--watch` breaks RPC
## Expected Behavior

When running `dune build --watch`, an RPC build request either fails or succeeds.

## Actual Behavior

RPC build request hangs.

## Reproduction

1. When you start dune in watch mode,
```
dune build --watch
```
2. RPC ping thinks the server is running:
```
~/PROJECT $ dune rpc ping
Server appears to be responding normally
```
3. but an RPC request for build hangs:
```
~/PROJECT $ dune rpc build path/file.v 
<hangs>
```

4. when running just dune build, both ping and build RPC request correctly return
```
~/PROJECT $ dune rpc ping
Error: rpc server not running
```
5. and when in passive watch mode, build request succeeds:
```
~/PROJECT $ dune rpc build path/file.v 
Success
```

## Specifications

- Version of `dune` (output of `dune --version`):

```
3.6.1
```

```
4.14.0
```

- Operating system (distribution and version):

```
NAME=Gentoo
ID=gentoo
PRETTY_NAME="Gentoo Linux"
VERSION_ID="2.9"
```

jchavarri:
Fixes #7314.

It seems `dune rpc build` is only compatible with watch "passive" mode, but not "eager" mode:

https://github.com/ocaml/dune/blob/4245029f2914bcfdca915e9916a5155a0afa95fb/doc/rpc.rst#L47-L49

This PR modifies the rpc server so that if a request for `build` comes from a client when the server is running in eager mode, it returns an error.

To test:

```
$ dune build -w # on one terminal
$ dune rpc build .
Error: { payload = None
; message =
    "the rpc server is running with eager watch mode using --watch. to run builds through an rpc client, start the server using --passive-watch-mode"
; kind = Invalid_request
}
```


bin/common.ml
```diff
@@ -515,7 +515,7 @@ module Builder = struct
     ; no_print_directory : bool
     ; store_orig_src_dir : bool
     ; default_target : Arg.Dep.t (* For build & runtest only *)
-    ; watch : Watch_mode_config.t
+    ; watch : Dune_rpc_impl.Watch_mode_config.t
```
jchavarri, line 518: I looked around for a place to put `Watch_mode_config` that both `bin` executable and the `dune_rpc_impl` could share, but could not find any suitable. The best would be `dune_config_file` but `dune_rpc_impl` doesn't depend on it.

rgrinberg:
LGTM. Can we get a test in?

jchavarri:
> Can we get a test in?

Done in https://github.com/ocaml/dune/pull/7351/commits/b1a3beca2c10a60f6e81e4d197c6e79510c43cc1.

bin/common.ml
rgrinberg: That's fine. Moving stuff out of the binary is usually fine.

PR #6476, opened 2022-11-15
fix(melange): run melc ppx with merlin

jchavarri:
Fixes #6456.

A couple of questions:
- I am not sure what to do with `to_dot_merlin`. It seems all `FLG` directives are commented? I don't understand what's the purpose
- Unsure if `-bs-jsx 3` should be conditionally added, based on `flags` field value on `library` or `melange.emit` stanzas.

PR #7354, opened 2023-03-20
melange: don't pass -bs-jsx flag to merlin

jchavarri:
Melange was updated recently to remove the need to pass `bs-jsx` flag, and instead model that transformation as a proper ppx:
- https://github.com/melange-re/melange/pull/517
- https://github.com/melange-re/melange/pull/525

This PR updates dune so the flag is not passed to merlin.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,8 +74,8 @@ dev-depext:
 
 .PHONY: melange
 melange:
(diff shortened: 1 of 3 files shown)


anmonteiro (approved):
Nice! this is the last step of https://github.com/ocaml/dune/pull/6476#discussion_r1024322957

Issue #5808, opened 2022-06-01
[minor] XDG default values do not seem right under Windows


nojb:
In `Xdg`:

- `XDG_CACHE_HOME` --> `%HOME%/Local Settings/Cache`
- `XDG_CONFIG_HOME` --> `%HOME%/Local Settings`
- `XDG_DATA_HOME` --> `%APPDATA%`

This does not seem very good as the `HOME` environment variable is not typically defined on Windows (it is under Cygwin, but that is an "accident").

Instead, suggest the following correspondence:

- `XDG_CACHE_HOME` --> `%TEMP%`
- `XDG_CONFIG_HOME` --> `%LOCALAPPDATA%`
- `XDG_DATA_HOME` --> `%LOCALAPPDATA%`

This coincides eg with what is done in Racket https://docs.racket-lang.org/basedir/index.html

jonahbeckford:
I'll mention that Opam also equates `XDG_CONFIG_HOME` with `%LOCALAPPDATA%`. See https://github.com/ocaml/opam/pull/4815#issuecomment-910137754 . That is great!

Only thing I would add is that `XDG_CACHE_HOME` is probably better as [FOLDERID_InternetCache](https://docs.microsoft.com/en-us/windows/win32/shell/knownfolderid) like is used in [GLib on Windows](https://docs.gtk.org/glib/func.get_user_cache_dir.html). The problem with `%TEMP%` is that Windows does not delete the files in it; you (the developer) are expected to clean up after yourself. The Internet Cache folder is a much more managed cache and is the first thing that Windows suggests to delete when the system drive is running out of space; it can also be deleted from the Control Panel.

The only annoying thing with the cache folder is that its value should be looked up with a Windows API (there is a C code sample in the FOLDERID_InternetCache link above) or with `cygpath --folder 32 --absolute --windows`.

PR #7356, opened 2023-03-20
Doc: fix default location of cache on Windows

nojb:
Small Doc fix.

### DIFF
--- a/doc/caching.rst
+++ b/doc/caching.rst
@@ -22,8 +22,9 @@ Choose the one that is more convenient for you:
 * Run Dune with the ``--cache=enabled`` flag.
 
 By default, Dune stores the cache in your ``XDG_CACHE_HOME`` directory on \*nix
-systems and ``"HOME\\Local Settings\\Cache"`` on Windows. You can change the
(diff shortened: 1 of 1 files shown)


Alizter:
Why is it there?

Alizter:
I see, so it is for Windows users to easily manager their cache when it grows too large? In that case, shouldn't they be using `dune cache trim` rather than Windows nuking it?

Alizter (approved):
My questions are not really related to this PR. Looks good.

nojb:
> I see, so it is for Windows users to easily manager their cache when it grows too large?

I wouldn't say there was a clear reason for this specific choice; the way I see it, in the absence of an specific reason, we chose to follow the choice done by GLib on Windows for `XDG_CACHE_HOME` as mentioned in https://github.com/ocaml/dune/issues/5808#issuecomment-1171561776. But if this location has downsides we can always revisit this choice.

In any case, the location was not chosen to simplify the manual management of the cache; as you mention one should use `dune cache trim` for that purpose.

PR #7357, opened 2023-03-20
Fix for 7344.

rlepigre:
This disables some code that was meant at supporting absolute paths, which I think we want to ditch. But @Alizter mentioned this could be done using existing infrastructure.


Alizter:
What was the reason you supported absolute paths again? I tried playing around with it today and it is not so straightforward to get Dune to recognize the source directory in an absolute path.

rlepigre:
> What was the reason you supported absolute paths again? I tried playing around with it today and it is not so straightforward to get Dune to recognize the source directory in an absolute path.

No good reason. Initially, I think it was because PG and Emacs passed absolute paths. But they need to be patched anyway to work with `dune coq top`, so we might as well require that they detect the workspace and using relative paths from the workspace root.

Alizter:
In that case, can I push a patch to this PR forbidding absolute paths? It will also cleanup some of the code.

rlepigre:
@Alizter I don't understand what is happening in CI. We get:
```
   $ unset INSIDE_DUNE
   $ . ./util.sh
+  ***** UNREACHABLE *****
 
 Calling "dune coq top" from the workspace root.
 
   $ coqtop_test file1.v
+  ***** UNREACHABLE *****
```
and it seems that code after sourcing the `util.sh` file is considered unreachable. I also had that locally at some point, but this went away (I think by changing `. util.sh` into `. ./util.sh`, but not sure since both should do the same I think). Running `make dev && DUNE_COQ_TEST=enable ./dune.exe build @coqtop-workspace` succeeds when you checkout the branch locally.

Any idea how to fix this?

Alizter:
There is a helpers.sh script in test-cases/watching which doesn't have the bash header and also writes functions like this. I wonder if that will fix the script.

rlepigre:
> There is a helpers.sh script in test-cases/watching which doesn't have the bash header and also writes functions like this. I wonder if that will fix the script. My suspicion is it was a bash / sh thing.

Oh, that might well be, good idea.

rlepigre:
Seems that you were right. The failure is now different, and related to the conditionals in the script. An easy fix is to remove the `check_build` function and its uses. This is not really related to `dune coq top` anyway. I'll push something.

Alizter:
Ah whoops we pushed at the same time. You can remove my commit that comes later.

rlepigre:
I'll stop for today, so go ahead. If you're happy with my removal of the `check_build` function then probably your last commit is not useful. But I'll let you decide what you prefer.

Alizter (approved):

rlepigre:
Should we add an entry to the CHANGELOG since we removed some absolute paths stuff @Alizter?

Alizter:
@rlepigre I can add one later.

PR #7360, opened 2023-03-21
feature: async sandboxing

rgrinberg:
We construct and remove sandboxes in background threads to allow the build engine to proceed while sandboxes are still being prepared.

I still need to do some benchmarking to show this is useful.

### DIFF
--- a/src/dune_config/config.mli
+++ b/src/dune_config/config.mli
@@ -40,4 +40,7 @@ val copy_file : [ `Portable | `Best ] t
 (** Compute digests of files in a background thread *)
 val background_digests : Toggle.t t
 
+(** Build and destroy sandboxes in background threads *)
+val background_sandboxes : Toggle.t t
+
 val init : (Loc.t * string) String.Map.t -> unit
--- a/src/dune_engine/sandbox.ml
+++ b/src/dune_engine/sandbox.ml
@@ -2,6 +2,14 @@ open Import
 
 let sandbox_dir = Path.Build.relative Path.Build.root ".sandbox"
 
+let maybe_async f =
+  (* It would be nice to do this check only once and return a function, but the
+     type of this function would need to be polymorphic which is forbidden by the
+     relaxed value restriction. *)
+  match Config.(get background_sandboxes) with
+  | `Disabled -> Fiber.return (f ())
+  | `Enabled -> Scheduler.async_exn f
+
 let init =
   let init =
     lazy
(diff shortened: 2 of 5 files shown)


rgrinberg:
@jchavarri do you rely on sandboxing? might be good for you to test this out if yes.

snowleopard:
Nice! I'm very curious to see hot it affects build times in the monorepo benchmark.

jchavarri:
> do you rely on sandboxing? might be good for you to test this out if yes.

I don't think we do, so I'm not sure I would see any changes cc @Khady

Khady:
we don't use any of the options mentioned in https://dune.readthedocs.io/en/stable/concepts.html#global-sandboxing-configuration so I guess we don't rely on sandboxing?

rgrinberg:
@snowleopard this is now ready for review.

> Nice! I'm very curious to see how this PR affects build times in the monorepo benchmark.

Note that we disable sandboxing for the compiler so we need a benchmark that uses sandboxing extensively.

src/dune_engine/sandbox.ml
```diff
@@ -2,6 +2,11 @@ open Import
 
 let sandbox_dir = Path.Build.relative Path.Build.root ".sandbox"
 
+let async_sandboxing f =
+  match Config.(get background_sandboxes) with
```
snowleopard, line 6: You could probably move the `match` to the top level.

src/dune_engine/sandbox.ml
```diff
@@ -2,6 +2,11 @@ open Import
 
 let sandbox_dir = Path.Build.relative Path.Build.root ".sandbox"
 
+let async_sandboxing f =
```
snowleopard, line 5: Should be called `maybe_async` instead? This function is not always async (and is doesn't by itself do any sandboxing).

src/dune_engine/sandbox.ml
rgrinberg: I wish. But then it would mean that the feature wouldn't be configurable through dune-config. If we could live with configurations being set through the environment, we could do it at the toplevel though.

src/dune_config/config.mli
```diff
@@ -40,4 +40,7 @@ val copy_file : [ `Portable | `Best ] t
 (** Compute digests of files in a background thread *)
 val background_digests : Toggle.t t
 
+(** Build and remove sandboxes in background threads *)
```
snowleopard, line 43:
```suggestion
(** Build and destroy sandboxes in background threads *)
```

snowleopard (approved):
Looks good! We can benchmark this internally.

PR #7371, opened 2023-03-21
feature: execute actions in background

rgrinberg:
This PR makes all the actions execute in background threads. The advantage is that we're able to proceed with builds without waiting for various synchronous IO operations.

### DIFF
--- a/src/dune_config/config.mli
+++ b/src/dune_config/config.mli
@@ -37,6 +37,10 @@ val copy_file : [ `Portable | `Best ] t
     Note that environment variables take precedence over the values passed here
     for easy overriding. *)
 
+(** Execute some actions in background threads. See [Action_exec] for the
+    concrete list of actions *)
+val background_actions : Toggle.t t
+
 (** Compute digests of files in a background thread *)
 val background_digests : Toggle.t t
 
--- a/src/dune_engine/action_exec.ml
+++ b/src/dune_engine/action_exec.ml
@@ -2,6 +2,15 @@ open Import
 open Fiber.O
 module DAP = Dune_action_plugin.Private.Protocol
 
+let maybe_async =
+  let maybe_async =
+    lazy
+      (match Config.(get background_actions) with
+      | `Enabled -> Scheduler.async_exn
+      | `Disabled -> fun f -> Fiber.return (f ()))
+  in
+  fun f -> (Lazy.force maybe_async) f
+
 (** A version of [Dune_action_plugin.Private.Protocol.Dependency] where all
     relative paths are replaced by [Path.t]. (except the protocol doesn't
     support Globs yet) *)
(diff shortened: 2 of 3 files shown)


rgrinberg:
@jchavarri similar to async sandboxing

Alizter:
The Coq tests seem OK. If you like I can push them updated if you don't want to whirl up Coq.

rgrinberg:
The same problem exists with all tests. I'll make the tests reproducible if we decide to merge this

jchavarri:
> @jchavarri similar to async sandboxing

Tested with 2 large builds, no meaningful changes in build time.

snowleopard (commented):
The change looks good to me but here are some thoughts:
* Firstly, did you actually check that the code you're running in separate threads is thread-safe? I can imagine that some of the functions you're calling can be using some global state (buffers?) that can't be shared between threads. 
* How about first putting this behind an experimental configuration flag? That would make it easier for me to test/benchmark on my end. 
* Running work in a separate thread is nice but there are also some overheads, in that the total work is surely increasing. Do you have any estimate of how small/big these overheads are?

rgrinberg:
> Firstly, did you actually check that the code you're running in separate threads is thread-safe? I can imagine that some of the functions you're calling can be using some global state (buffers?) that can't be shared between threads.

Yes, it's all safe. Although it's a bit fragile because it relies on us using absolute paths in the underlying actions. Because if we use a relative path, we are racing against `chdir`. It should be possible to address this by using the collection of `*at` calls on Unix.

> How about first putting this behind an experimental configuration flag? That would make it easier for me to test/benchmark on my end.

I think that's a good idea. Will do.

> Running work in a separate thread is nice but there are also some overheads, in that the total work is surely increasing. Do you have any estimate of how small/big these overheads are?

Indeed I do. I added a set of benchmarks (`thread_pool_bench`) to give me an idea. To give an idea, the cheapest syscall (`Unix.sleep 0`) with all the benchmarking overhead is around `66ms` without the thread pooling. While, with the thread pool it's around `150 ms`. So about a 2.3x slow down in the worst case. All of our actions are usually, but not always, considerably more expensive than that.

rgrinberg:
@snowleopard this is now ready

snowleopard:
> To give an idea, the cheapest syscall (`Unix.sleep 0`) with all the benchmarking overhead is around `66ms` without the thread pooling. While, with the thread pool it's around `150 ms`. So about a 2.3x slow down in the worst case. All of our actions are usually, but not always, considerably more expensive than that.

Wait, how is the cheapest syscall 66 milliseconds? Did you mean microseconds?

src/dune_config/config.mli
```diff
@@ -37,6 +37,9 @@ val copy_file : [ `Portable | `Best ] t
     Note that environment variables take precedence over the values passed here
     for easy overriding. *)
 
+(** Execute some actions in background threads *)
```
snowleopard, line 40: It would be nice to specify which actions are affected by this ("some" is a little too vague).

src/dune_engine/action_exec.ml
```diff
@@ -289,20 +301,27 @@ let rec exec t ~display ~ectx ~eenv =
...
+          List.iter xs ~f:(fun fn ->
+              Io.with_file_in fn ~f:(fun ic ->
+                  Io.copy_channels ic (Process.Io.out_channel eenv.stdout_to))))
+    in
+    Done
   | Copy (src, dst) ->
     let dst = Path.build dst in
-    Io.copy_file ~src ~dst ();
-    Fiber.return Done
+    let+ () = Scheduler.async_exn (fun () -> Io.copy_file ~src ~dst ()) in
```
snowleopard, line 313: Should this be `maybe_async`?

snowleopard (commented):
Looks good but see a couple of comments.

src/dune_engine/action_exec.ml
rgrinberg: Indeed it should be. Fixed

src/dune_config/config.mli
rgrinberg: Done

snowleopard (approved):

PR #7373, opened 2023-03-22
feat: allow preprocessing action chains before `pps`

### LINKED ISSUE #171: Preprocess using both cppo and PPXs

anmonteiro:
fixes #171


rgrinberg (commented):
Now I'm wondering if it would be simpler if we made a `ppx` action and then compose it with the other actions using our redirections?

```
(progn
 (with-stdout-to %{input-file}.cppo (run cppo %{input-file}))
 (ppx (lib1 lib2) %{input-file}.cppo))
```

It's a little bit more manual, but perhaps cleaner to understand.

We could also allow users to insert `ppx` in any position if we add a flag to make it output pretty printed AST. E.g. `(ppx_pp_ast (lib1 lib2) %{input-file})`.

anmonteiro:
> Now I'm wondering if it would be simpler if we made a `ppx` action and then compose it with the other actions using our redirections?
> 
> ```
> (progn
>  (with-stdout-to %{input-file}.cppo (run cppo %{input-file}))
>  (ppx (lib1 lib2) %{input-file}.cppo))
> ```
> 
> It's a little bit more manual, but perhaps cleaner to understand.
> 
> We could also allow users to insert `ppx` in any position if we add a flag to make it output pretty printed AST. E.g. `(ppx_pp_ast (lib1 lib2) %{input-file})`.

I'm not such a big fan of this:

- right now there's a user need for running preprocessing actions before running the preprocessor in a single pass
- this is still a bit restrictive and steers users towards using the feature correctly, in a way that won't build too many intermediate artifacts (and potentially impact performance).
- adding a `(ppx ..)` action increases the functionality surface area quite a bit in that regard

What do you think about keeping the current restriction (I think we could even allow a single action (though you could always use `progn`) and wait until there's an actual user need for what you described?

rgrinberg:
Removing the restriction wasn't the goal with my proposal. I just prefer to avoid adding an extra type (a list of preprocessors) of a value for the prepreocessor field. If the limitation stays, I don't particular care.

anmonteiro:
- Added `ml_kind` as an argument to `Action_unexpanded.expand{,_no_targets}`
- Planning on adding docs in a follow-up PR

src/dune_lang/action.ml
```diff
@@ -270,6 +312,10 @@ let decode =
           , Syntax.since Stanza.syntax (2, 7)
             >>> let+ script = sw in
                 Cram script )
+        ; ( "ppx"
+          , let+ pps_and_flags = enter Pps_and_flags.decode
```
rgrinberg, line 316: There should be a version check here

src/dune_rules/preprocessing.ml
```diff
@@ -1,101 +1,6 @@
 open Import
 open Memo.O
-
-(* Encoded representation of a set of library names + scope *)
-module Key : sig
```
rgrinberg, line 5: How come you needed to split this into a separate module?

anmonteiro: `Action_unexpanded` would need `Preprocessing` now. And `Preprocessing` already depends on `Action_unexpanded`.

src/dune_rules/merlin/merlin.ml
```diff
@@ -390,8 +390,10 @@ module Unprocessed = struct
       | Some args ->
         let action =
           let action = Preprocessing.chdir (Run (exe, args)) in
-          Action_unexpanded.expand_no_targets ~loc ~expander ~deps:[]
+          Action_unexpanded.expand_no_targets ~dir ~loc ~expander ~deps:[]
             ~what:"preprocessing actions" action
+            ~lib_name (* TODO: Need ml_kind for merlin support  *)
```
rgrinberg, line 395: could you try adding merlin support?

test/blackbox-tests/test-cases/github1811.t/run.t
```diff
@@ -4,5 +4,5 @@ Check that %{targets} is forbidden in preprocessing actions
   File "dune", line 3, characters 37-47:
   3 |  (preprocess (action (with-stdout-to %{targets} (run cat %{input-file})))))
                                            ^^^^^^^^^^
-  Error: You cannot use %{targets} in preprocessing actions.
+  Error: You cannot use %{targets} with inferred rules.
```
rgrinberg, line 7: We should keep the old error message somehow.

rgrinberg:
A merlin test would be quite helpful as well.

anmonteiro:
- renamed `Preprocessing` -> `Pp_spec_rules`
- renamed `Pps` -> `Ppx_driver`

src/dune_rules/action_unexpanded.ml
```diff
@@ -415,6 +475,18 @@ let rec expand (t : Dune_lang.Action.t) ~context : Action.t Action_expander.t =
     let+ t = expand t in
     O.Ignore (outputs, t)
   | Progn l ->
+    let rec check_ppx_is_last as_ =
+      match (as_ : Dune_lang.Action.t list) with
+      | Ppx ({ loc; _ }, _) :: _ :: _ ->
+        User_error.raise ~loc
+          [ Pp.text
+              "The (ppx ..) action is currently only supported in the last \
```
jchavarri, line 483: Is "currently" needed in the error message? Are there any plans to remove that limitation soon?

src/dune_rules/action_unexpanded.ml
```diff
@@ -479,8 +551,22 @@ let rec expand (t : Dune_lang.Action.t) ~context : Action.t Action_expander.t =
   | Cram script ->
     let+ script = E.dep script in
     Cram_exec.action script
+  | Ppx ({ pps; flags; loc }, input_file) -> (
+    match ml_kind with
+    | None ->
+      Code_error.raise ~loc
```
jchavarri, line 557: I wonder if this is is "code error" or "user error". For example, this stanza triggers this case:

```
(rule
 (action
  (progn
   (run cat %{input-file})
   (ppx (some_ppx) %{input-file}.pp.1))))
```

It would be more informative to tell users that this case is not supported, rather than show stack trace and "Uncertainty is the mind-killer" error message.

Alternatively, can't dune just forbid the usage of `ppx` in simple user rules? Is there a use case for that?

test/blackbox-tests/test-cases/preprocess-with-action-and-pps-errors.t
```diff
@@ -0,0 +1,42 @@
+Show an error when `(ppx ...)` appears in any position other than the last
```
jchavarri, line 1: Might be useful to add a test for the case where `ppx` is used inside simple `rule`, e.g.

```
(rule
 (action
  (progn
   (run cat %{input-file})
   (ppx (some_ppx) %{input-file}.pp.1))))
```

src/dune_rules/pp_spec_rules.ml
```diff
@@ -0,0 +1,449 @@
+open Import
```
jchavarri, line 1: Did anything meaningful change in this file and `ppx_driver.ml`? Hard to know with this diff.

jchavarri (approved):
Just have this minor concern about the usage with `(rule)`.

src/dune_rules/action_unexpanded.ml
rgrinberg: >  Is there a use case for that?

It can be useful to write expect tests for ppx

src/dune_rules/action_unexpanded.ml
anmonteiro: Yes. I'm planning on improving the usability of this feature in future PRs.

src/dune_rules/action_unexpanded.ml
anmonteiro: I'll add a cram test for this error that will be fixed in a future PR.

src/dune_rules/pp_spec_rules.ml
anmonteiro: Ppx_driver.ml is a new file where I moved the ppx driver helpers. I needed to move them because Action_unexpanded depends on pp_spec_rules. Then, I moved `preprocessing.ml` to `pp_spec_rules.ml`, which contains a) deletions of those functions moved to ppx_driver.ml 2) a few lines of code related to calling action_unexpanded for the new ppx action.

src/dune_rules/action_unexpanded.ml
anmonteiro: added a cram test in [test/blackbox-tests/test-cases/preprocess-with-action-and-pps-errors.t](https://github.com/ocaml/dune/pull/7373/files#diff-068ca8d8bc3a68def25a8b915f10c72a371d850596714f3fe535e43ce8d9e4dd)

jchavarri:
@rgrinberg do you have any remaining concerns about this PR? It unblocks the extraction of the Melange ppx into its own preprocessor that Antonio implemented in https://github.com/melange-re/melange/pull/527, which itself unblocks some errors with the deps calculation using ocamldep that we were tracking in https://github.com/ocaml/dune/pull/6625, https://github.com/ocaml/dune/pull/7321 and https://github.com/ocaml/dune/pull/7323 (I will revisit those PRs to make sure those tests are added to the melange test suite).

src/dune_rules/action_unexpanded.ml
```diff
@@ -415,6 +475,18 @@ let rec expand (t : Dune_lang.Action.t) ~context : Action.t Action_expander.t =
     let+ t = expand t in
     O.Ignore (outputs, t)
   | Progn l ->
+    let rec check_ppx_is_last as_ =
```
rgrinberg, line 478: Could you move this function the to the toplevel? We handle a lot of `Progn` constructors, so this allocation should be avoided. It also makes the code clearer.

src/dune_rules/action_unexpanded.mli
```diff
@@ -10,6 +10,10 @@ end
 
 val remove_locs : t -> t
 
+type what =
+  | Default
+  | Override of Expander.Expanding_what.t
```
rgrinberg, line 15: Is it not possible to use `Expander.set_expanding_what`?

src/dune_rules/merlin/merlin.ml
rgrinberg: Or actually, does it? Doesn't merlin decide to pass `--intf` or `--impl`?

@jchavarri did you check if this PR works with merlin?

anmonteiro: It won't work with merlin yet because we're not generating the `ppx` directive. e.g. `progn` falls here: https://github.com/ocaml/dune/blob/a5fd306a4e24c7bd67330b411b3e5662b7e03039/src/dune_rules/merlin/merlin.ml#L409

> I think this still needs to be addressed, right?

I thought we had decided to add it in a follow-up PR, hence the `TODO`.

src/dune_rules/action_unexpanded.ml
```diff
@@ -367,15 +380,62 @@ end = struct
       register_dep b env acc ~f:(function
         | Ok p, _ -> Some p
         | Error _, _ -> None)
+
+    let ppx ~context ~dir ~expander ~scope ~loc ~lib_name ~ml_kind ~src pps
```
rgrinberg, line 384: Can't this `dir` be `Expander.dir`?

src/dune_rules/action_unexpanded.ml
```diff
@@ -367,15 +379,76 @@ end = struct
...
+             (exe, ppx_flags, flags))
+        and+ src =
+          src >>= function
+          | External p -> Action_builder.return (Path.External.to_string p)
+          | In_source_tree _ -> assert false
+          | In_build_dir b ->
+            (* PPX runs at the workspace root *)
+            let+ dir =
+              String_with_vars.virt_pform __POS__ (Var Workspace_root)
+              |> Expander.expand_path env
```
rgrinberg, line 420: I think this is just `context.build_dir`.

rgrinberg:
There's still some regressions in the error messages

src/dune_rules/action_unexpanded.ml
rgrinberg: Actually, is this the right place to do this? It's the `preprocess` field that is responsible for setting `chdir` for normal preprocessors. I don't see why this is any different.

anmonteiro: I just need to find the path relative to where the PPX is running, to find the source file. What's the best way to do that? I can use` context.build_dir`

rgrinberg: I guess there's no way to do that here because there could be a `(chdir ..)` wrapping the action later. Perhaps we should just hard code the chdir action here and expand relative to context.build_dir?

In the future, we really need to stop flattening paths inside actions that aren't wrapped with a `(chdir ..)`

rgrinberg: This is a general problem we have in many other cases, so there's no need for a principled fix here.

src/dune_rules/pp_spec_rules.ml
```diff
@@ -0,0 +1,442 @@
...
+    let bindings =
+      Pform.Map.of_list_exn
+        [ (Var Input_file, [ Value.Path (Path.build src) ]) ]
+    in
+    Expander.add_bindings expander ~bindings
+  in
+  let+ action =
+    let+ action =
+      let ml_kind = Some ml_kind in
+      Action_unexpanded.expand action ~chdir ~loc ~expander ~deps:[] ~lib_name
```
rgrinberg, line 149: Shouldn't this be `Action_unexpanded.expand_no_targets`?

anmonteiro: I don't know. On line 166 below we do:

```ocaml
      Action.Full.map build ~f:(Action.with_stdout_to target))
```

rgrinberg: Yes, but that's different. We're adding a target ourselves, but we shouldn't allow users to add actions that introduce targets. For example, I think you're now allowing actions such as `(progn (with-stdout-to unrelated (echo foo)) ...)` which produce this `unrelated ` target.

rgrinberg: Weird. For normal pp, we redirect the action's stdout to the target. With ppx only, we use the `-o` argument. I would stick to using stdout.

anmonteiro: I think we're speaking past each other. I think your example is now valid because of the intermediate files produced by intermediate actions.

rgrinberg: I don't quite remember, but I think to make the driver write to stdout, you need the `-o -` flag.

rgrinberg: Yes, I know it's valid, I'm just not sure it should be allowed. Ideally, all of these temporary files would disappear after the action runs.

anmonteiro: > Weird. For normal pp, we redirect the action's stdout to the target. With ppx only, we use the `-o` argument. I would stick to using stdout.

did you mean `cppo` here then?

rgrinberg: No I meant the ppx driver exe. It has a -o option.

By the way, I could rephrase the issue in a much simpler way:

The user action shouldn't introduce its temporary files as additional targets. Ideally, the temporary files would all be deleted after the preprocessing is done.

Probably the simplest thing to do is to introduce a construct for temporary directories where actions can write temp files into. This is still quite a bit more work than I expected though. Let me think about this a bit more.

anmonteiro: In parallel, I could explore another way to compile the melange repo in a way that doesn't need this feature.

emillon:
> Planning on adding docs in a follow-up PR

This doesn't have to be in this PR but I do think that this should be documented for 3.8. I recently moved this stuff around so here's my suggestion:
- in reference/preprocessing-spec, update the grammar on top
- add a section about this combination
- not sure if explanation/preprocessing needs updating, maybe a note to explain where this action-based preprocessing fits in the fast and classic pipelines (can you combine cppo with `staged_pps`?)

emillon:
Ah, after reading the implementation I realize it does not quite do what I think it does.

Exposing `(ppx)` as an action seems low level to me but that works. I was hoping for something a bit higher level like `(chain cppo (pps ppx_expect))` but this is something we can build on top of this.

Is there a difference between `(pps ...)` and `(action (ppx ...))` ?

anmonteiro:
This is not pressing for 3.8 anymore. I still plan to work on it since it's a useful feature to have, but at least that gives us time to get it right.

emillon:
ok! I'm clearing the milestone. don't forget to change the lang version when retargeting. Thanks!

rgrinberg:
Btw, I quite like the refactoring changes (new modules, new names) in this PR. If those could be extracted that would be appreciated.

anmonteiro:
Good idea. I'll work on that.

maiste:
@rgrinberg @anmonteiro rebase conflict apart, was this PR ready to be merged?

anmonteiro:
I think this implemented chaining of actions but Rudi's idea was to have a more specific `(ppx ..)` action for chaining PPXes with other actions. I don't think it was ready to be merged.

maiste:
Thanks for the quick reply! Do you think we should consider closing this one and start from a fresh start?

craff:
Any progress ? I really miss that features for the chaml preprocessor (a php like syntax) for which I need a rule + preprocessing because I can not chain pps and action.

rgrinberg:
Feel free to re-open if you intend to work on this again

Issue #7374, opened 2023-03-22
melange, merlin: go to definition in stdlib or belt modules doesn't work


jchavarri:
There are a couple of issues with merlin that surfaced after the stdlib installation changes done in https://github.com/melange-re/melange/pull/493. This cause problems when jumping to definition for either Stdlib or Belt modules.

In particular:
- Melange object files live in `<install-root>/lib/<lib-name>/melange` but the source files are one level higher. Right now we only tell merlin about the object files
- Melange has the `Stdlib` module in one folder, and the submodules in a subfolder `stdlib_modules`

jchavarri:
Second part fixed in https://github.com/melange-re/melange/pull/526.

jchavarri:
First part fixed in https://github.com/ocaml/dune/pull/7375.

PR #7375, opened 2023-03-22
fix(melange): Tell merlin about stdlib sources

anmonteiro:


jchavarri (changes requested):
Looks good, ~but there seems to be some mixed results in the tests with another PR.~

Also, please add "Fixes #7374" to the description (I don't have permissions to edit it).

test/blackbox-tests/test-cases/melange/merlin.t
```diff
@@ -52,16 +52,20 @@ All 3 modules (Foo, Foo__ and Bar) contain a ppx directive
 The melange.emit entry contains a ppx directive
 
   $ dune ocaml merlin dump-config $PWD | grep -i "ppx"
-   (FLG (-ppx "/MELC_COMPILER -as-ppx -bs-jsx 3"))
+    (-ppx
+     -as-ppx
```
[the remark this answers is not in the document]
jchavarri: Oh, n/m is just an artifact from `grep` 🤦

jchavarri (approved):

rgrinberg:
> but we're not doing that so e.g. libraries can be shared with native.

That only follows because we have a single effective `libraries` list. If we were to keep separate lists for melange and byte + native, this would no longer be a problem.

anmonteiro:
That makes sense. It's not yet clear to me whether having different library lists is going to be necessary beyond the stdlib / runtime, though.

What do you think about gathering feedback from users before we design the feature you're suggesting?

rgrinberg:
Okay, but we can still hard code the look up the melange stdlib? E.g. if the mode is melange, we look up the melange stdlib using our library database and fetch the information from there.

jchavarri:
Could we do something similar than https://github.com/ocaml/dune/pull/7362, but in addition to `melange.emit` do so for `library` stanzas? i.e we could in `Melange_rules` add `melange` library to the requires_link closure.

That might avoid introducing mechanisms that define two separate lists of libraries (imo we should promote vlibs for that), and solve merlin problem.

One downside is that it would not be possible for users to define a library named `melange` 🤔

anmonteiro:
That makes sense. Are you suggesting 

- a) doing that lookup in this merlin code specifically?
- or b) adding `melange` to `(libraries .. )` e.g. in `dune_file.ml` when we see `modes melange`?

rgrinberg:
I meant the former. Although I see nothing wrong with the latter if you could somehow stub out `melange` to do nothing for byte/native.

anmonteiro:
Thanks for the suggestion -- did that in my latest commit. This looks a lot better now, arguably.

src/dune_rules/merlin/merlin.ml
```diff
@@ -466,18 +466,35 @@ module Unprocessed = struct
...
+          Action_builder.of_memo
+            (let open Memo.O in
+            let libs =
+              let scope = Expander.scope expander in
+              Scope.libs scope
+            in
+            Lib.DB.find libs (Lib_name.of_string "melange") >>= function
+            | Some lib ->
+              let+ libs =
+                Lib.closure [ lib ] ~linking:true |> Resolve.Memo.read_memo
```
rgrinberg, line 494: I think `~linking` should respect `implicit_transitive_deps` like it does elsewhere.

Also, instead of `read_memo`, should we just fail silently?

anmonteiro: I don't understand what you mean by failing silently here? We still need the `libs` to `Lib.Set.union` it to the other `requires`.

anmonteiro: do you mean I should just `Resolve.peek` and ignore if it's an `Error`?

anmonteiro: Pushed a version of the patch that uses `Resolve.Memo.peek` and takes `implicit_transitive_deps` into account

rgrinberg: Yeah, that's what I meant. I'm not even sure if it's the right thing to do, but it does seem a bit better to offer some support if some libraries fails to resolve.

PR #7179, opened 2023-02-24
Source fetching

Leonidas-from-XIV:
This is a PR to implement source fetching as suggested in #7158. It uses the opam-libraries to construct a command to download the source, which has the advantage of using the same fetching commands as OPAM does, but it wraps the execution using the process spawning/concurrency logic of Dune.

Requires https://github.com/ocaml/opam/pull/5452 to access the contents of the `command` type, to execute the command.

Currently work in progress as it still requires some testing and some parts are implemented in a very rudimentary way.

PR #7378, opened 2023-03-22
Document style conventions for qualifying modules

Leonidas-from-XIV:
This documents some comments I got from @rgrinberg when working on #7179, that weren't covered in the document yet.

### DIFF
--- a/doc/hacking.rst
+++ b/doc/hacking.rst
@@ -440,8 +440,49 @@ Good:
   ``foo``.
 
 - Avoid optional arguments. They increase brevity at the expense of readability
-  and are annoying to grep. Further more, they encourage callers not to think
+  and are annoying to grep. Furthermore, they encourage callers not to think
   at all about these optional arguments even if they often should.
+
+- Avoid qualifying modules when accessing fields of records or constructors.
+  Avoid it altogether if possible, or add a type annotation if
+  necessary.
+
+Bad:
+
+.. code:: ocaml
+
+    let result = A.b () in
+    match result.A.field with
+    | B.Constructor -> ...
+
+Good:
+
+.. code:: ocaml
+
+    let result : A.t = A.b () in
+    match (result.field : B.t) with
+    | Constructor -> ...
+
+- When constructing records, use the qualified names in in the record. Do not
+  open the record. The local open syntax pulls in all kinds of names from the
+  opened module and might shadow the values that you're trying to put into the
+  record, leading to difficult debugging.
+
+Bad; if ``A.value`` exists, it will pick that over ``value``:
+
+.. code:: ocaml
+
+    let value = 42 in
+    let record = A.{ field = value; other } in
+    ...
+
+Good:
+
+.. code:: ocaml
+
+    let value = 42 in
+    let record = { A.field = value; other } in
+    ...
 
 - Stage functions explicitly with the ``Staged`` module.
 


doc/hacking.rst
```diff
@@ -443,6 +443,48 @@ Good:
...
+  Prefer to avoid it altogether if possible or add a type annotation if
+  necessary.
+
+Bad:
+
+.. code:: ocaml
+
+    let result = A.b () in
+    match result.A.field with
+    | B.Constructor -> ...
```
Alizter, line 456: There is an OCaml warning for this no? Maybe we should use it. Is it 40?

doc/hacking.rst
```diff
@@ -443,6 +443,48 @@ Good:
...
+
+Good:
+
+.. code:: ocaml
+
+    let result : A.t = A.b () in
+    let field : B.t = result.field in
+    match field with
+    | Constructor -> ...
```
Alizter, line 465:
```suggestion
    let result : A.t = A.b () in
    match (result.field : B.t) with
    | Constructor -> ...
```
Seems more common to me.

doc/hacking.rst
```diff
@@ -443,6 +443,48 @@ Good:
   and are annoying to grep. Further more, they encourage callers not to think
```
christinerose, line 443:
```suggestion
  and are annoying to grep. Furthermore, they encourage callers not to think
```

doc/hacking.rst
```diff
@@ -443,6 +443,48 @@ Good:
...
   at all about these optional arguments even if they often should.
 
+- Avoid qualifying modules when accessing fields of records or constructors.
+  Prefer to avoid it altogether if possible or add a type annotation if
```
christinerose, line 447:
```suggestion
  Avoid it altogether if possible, or add a type annotation if
```

doc/hacking.rst
```diff
@@ -443,6 +443,48 @@ Good:
...
+Good:
+
+.. code:: ocaml
+
+    let result : A.t = A.b () in
+    let field : B.t = result.field in
+    match field with
+    | Constructor -> ...
+
+- When constructing records, use the qualified names in in the record, do not
```
christinerose, line 467:
```suggestion
- When constructing records, use the qualified names in in the record. Do not
```

doc/hacking.rst
```diff
@@ -443,6 +443,48 @@ Good:
...
+  open the record. The local open syntax pulls in all kinds of names from the
+  opened module and might shadow the values that you're trying to put into the
+  record, leading to hard to debug issues.
```
christinerose, line 470:
```suggestion
  record, leading to difficult debugging.
```

christinerose (changes requested):
minor formatting/syntax/grammar...

doc/hacking.rst
Leonidas-from-XIV: It seems like the warning is enabled by default but dune disables it. Unfortunately, upon trying to figure out why it seems it has been always that way [since jbuilder days](https://github.com/ocaml/dune/commit/cdcd7e907f949fae66a0e57e569c7e0d46f40c3b).

rgrinberg: I don't see the point in the warning. If the code stop compiling due to a missing type annotation, we just add it. A warning to pre-empt that is just noise.

Alizter: Ah ok I had it backwards nevermind.

PR #7380, opened 2023-03-22
dune coq top: add --no-build option to avoid building dependencies

### LINKED ISSUE #7355: Option to make "dune coq top" more permissive
## Desired Behavior

Currently, running `dune coq top` on a Coq file starts by ensuring that all dependencies of that file are built and up-to-date, which in turn require that the file itself is part of a stanza (so, saved to disk) and that it can be parsed by `coqdep`.

This disturbs some people's workflow when writing Coq, as they often want to edit files more quickly without constantly recompiling deps even if they are not completely consistent, work on dirty files only a prefix of which is syntactically correct, or work on new files that have not been saved yet.

I feel like we could have an opt-in option to `dune coq top`  to support such use-cases. In particular, it would do no building of dependencies and rely on the user having called, e.g., `dune build` to get the dependencies in a reasonable state. Of course, to achieve this, one needs to figure out what options to pass to `coqtop`. One possibility (that was suggested by a colleague of mine) would be to compute a "most general configuration" from the workspace, that would put all theories and all plugins in scope.

CC @Alizter @ejgallego

Alizter:
We add a `--no-build` option to `dune coq top` that allows avoiding the building of dependencies.

- [x] changelog
- [x] doc

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,9 @@ Unreleased
   longer supported (due to being buggy) (#7357, fixes #7344, @rlepigre and
   @Alizter)
 
+- Added a `--no-build` option to `dune coq top` for avoiding rebuilds (#7380,
+  fixes #7355, @Alizter)
+
 - RPC: Ignore SIGPIPE when clients suddenly disconnect (#7299, #7319, fixes
   #6879, @rgrinberg)
 
--- a/bin/coq/coqtop.ml
+++ b/bin/coq/coqtop.ml
@@ -29,6 +29,10 @@ let term =
     Arg.(required & pos 0 (some string) None (Arg.info [] ~docv:"COQFILE"))
   and+ extra_args =
     Arg.(value & pos_right 0 string [] (Arg.info [] ~docv:"ARGS"))
+  and+ no_rebuild =
+    Arg.(
+      value & flag
+      & info [ "no-build" ] ~doc:"Don't rebuild dependencies before executing.")
   in
   let config = Common.init common in
   let coq_file_arg =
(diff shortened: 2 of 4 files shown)


rlepigre:
I'm getting the following error after trying this out:
```
Description:
  ("Command.Expert.expand: Dyn encountred", {})
Raised at Stdune__code_error.raise in file
  "otherlibs/stdune/src/code_error.ml", line 11, characters 30-62
Called from Stdlib__List.rev_map.rmap_f in file "list.ml" (inlined), line
  103, characters 22-25
Called from Stdlib__List.rev_map in file "list.ml", line 105, characters 2-13
Called from Stdune__list.map in file "otherlibs/stdune/src/list.ml"
  (inlined), line 5, characters 19-33
Called from Dune_rules__command.Expert.expand in file
  "src/dune_rules/command.ml", line 155, characters 26-56
Called from Main__coqtop.term.(fun) in file "bin/coq/coqtop.ml", line 122,
  characters 18-64
Called from Fiber__scheduler.exec in file "otherlibs/fiber/src/scheduler.ml",
  line 73, characters 8-11
```
My test was run in a complicated project involving plugins.

Alizter:
@rlepigre I added a way for dynamic arguments to be expanded. See if this now works.

rlepigre:
Thanks @Alizter!

Testing in the same project as before, the internal error is gone, but even with `--no-build` all `.vo` files seem to be rebuilt (and there are additionally some errors due to `.cmxs` files not being found for plugins).

Alizter:
@rlepigre OK, looks like I am running too much now. I will try to tweak it so that it is I between. Not easy to do but I will have a go soon.

Alizter:
OK I think I have a much simpler way to do it. Can you give it a go @rlepigre? This should avoid any rebuilding (coqdep included).

ejgallego:
Current version looks good to me, I think it should work. Happy to merge once test a doc are added.

rlepigre:
Seems to work perfectly! Thanks @Alizter.

CHANGES.md
```diff
@@ -15,6 +15,9 @@ Unreleased
   longer supported (due to being buggy) (#7357, fixes #7344, @rlepigre and
   @Alizter)
 
+- Added `-no-build` option to `dune coq top` for avoiding rebuilds (#7380, fixes
```
rlepigre, line 18:
```suggestion
- Added `--no-build` option to `dune coq top` for avoiding rebuilds (#7380, fixes
```

Alizter:
@rlepigre yes it is, I'll try to get @ejgallego to give it a look and merge.

ejgallego:
Shouldn't we document this in the rst files too? Otherwise looks good to me.

Alizter:
@ejgallego I didn't initially want to advertise this specialist feature in the documentation so I didn't include it. Now I've changed my mind and included it.

ejgallego:
Thanks @Alizter !

Indeed it seems to me that this is something often discussed by users so it makes sense to advertise it I think.

PR #6392, opened 2022-11-06
Add doc for config file

nojb:
Fixes #5493

Moreover, the "man" page for `dune-config(5)` is terribly out-of-date and does not seem very pratical to have the information in two places (the manual and the man page). Since we don't document any other Dune file format in man page format, shall we get rid of the `dune-config(5)` man page?

PR #7381, opened 2023-03-22
doc: fix jobs field in config

Alizter:
In https://github.com/ocaml/dune/pull/6392 we introduce the config options into the main doc. There was a typo where the `jobs` stanza was written as `concurrency`. Here we fix that.

@emillon could you get this fix into the stable version of the docs?

I only noticed this after some users were confused about it.

### DIFF
--- a/doc/dune-files.rst
+++ b/doc/dune-files.rst
@@ -764,16 +764,16 @@ where ``<setting>`` is one of:
 - ``copy`` copies entries to the cache. This is less efficient than using hard
   links.
 
-.. _concurrency:
+.. _jobs:
 
-concurrency
------------
+jobs
+----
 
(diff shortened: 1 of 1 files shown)


nojb (commented):
I would suggest changing the variable `concurrency` to `jobs` in
https://github.com/ocaml/dune/blob/7cd8b42c62ae489363e9ceb34df664d669478b88/src/dune_config_file/dune_config_file.ml#L296
(and to propagate that renaming everywhere else)

christinerose (approved):

Alizter:
If think that can be an independent change, if we want to do it, though I am not sure it is a good idea.

emillon:
> @emillon could you get this fix into the stable version of the docs?

This can be backported to 3.7.x like any other change, let's discuss if there are other candidates.

PR #7382, opened 2023-03-22
Add safe Path.With_check.mkdir_p

Alizter:
We add a `Path.With_check.mkdir_p` which does an extra syscall when the file exists to make sure it is actually a directory.

### DIFF
--- a/otherlibs/stdune/src/temp.ml
+++ b/otherlibs/stdune/src/temp.ml
@@ -44,6 +44,7 @@ let create_temp_dir ?perms path =
   match Fpath.mkdir ?perms dir with
   | Created -> Ok ()
   | Already_exists -> Error `Retry
+  | Not_a_directory -> Error `Retry
   | Missing_parent_directory ->
     Code_error.raise "[Temp.create_temp_dir] called in a non-existing directory"
       []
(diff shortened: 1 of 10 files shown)


rgrinberg:
Could you separate the PR outlining the issues with mkdir_p with unit tests in a separate PR?

Alizter:
@rgrinberg OK, I will include the spec in a later PR.

otherlibs/stdune/src/temp.ml
```diff
@@ -44,6 +44,7 @@ let create_temp_dir ?perms path =
   match Fpath.mkdir ?perms dir with
   | Created -> Ok ()
   | Already_exists -> Error `Retry
+  | Not_a_directory -> Error `Retry
```
Alizter, line 47: I am unsure about this but it seemed like the sensible behaviour to me.

Alizter:
@rgrinberg You mentioned there was a cram test for this exact case already however I could not find it so I kept this cram test.

rgrinberg:
This is actually a useful bug fix. It just needs to be finished

Alizter:
Not gonna finish it soon however so drafting.

PR #7047, opened 2023-02-11
feature(coq): composition of installed theories

Alizter:
We add support for composition with installed theories to the Coq rules.

This allows people to use Dune even if their project depends on an installed project built using make.

This works by first checking if coq lang is enabled, then scanning user-contrib and directories in COQPATH for potential theories. Then a database is constructed allowing users to write `(thories foo.bar)` and Dune will look for the Coq_lib_name `foo.bar`. The database would have an entry for `user-contrib/foo/bar` say and would in turn be able to describe a Coq theory.

Closes #6982

Any testing would be appreciated.

#### Tasks:

- [x] pass `-boot` to every coqc. We probably need to make a `:standard` for plugins if we do this.
- [x] coq lang version bump to 0.8  
- [x] investigate performance in presence of a larger user-contrib (my scanning is probably terrible atm)
  - Emilio made loading of the coq database lazy in #7133 
  - Since Coq DB creation is lazy and guarded behind coq lang, it doesn't affect OCaml projects.
- [x] composition with plugins
  - [x] Stop linking plugin tutorial 
  - [x] Sort flags when printing to make tests repro
- [x] fix extraction
- [x] fix dune coq top
- [x] make sure compatibility with < 0.8
  - [x] Nice warning when trying to compose but dune lang < 0.8 
- [x] Coq Universe testing - When using 0.8, projects are able to compose with non-dune projects and install and build.
- [ ] Documentation: https://github.com/ocaml/dune/pull/7384
- [x] Changelog

PR #7384, opened 2023-03-22
doc(coq): Add documentation for composition with installed theories

Alizter:
Doc for #7047

### DIFF
--- a/doc/coq.rst
+++ b/doc/coq.rst
@@ -16,6 +16,9 @@ and ``.mlg`` file preprocessing.
 A *Coq theory* is a collection of ``.v`` files that define Coq modules whose
 names share a common prefix. The module names reflect the directory hierarchy.
 
+Coq theories may be defined using :ref:`coq.theory<coq-theory>` stanzas, or be
+auto-detected by Dune by inspecting Coq's install directories.
+
 A *Coq plugin* is an OCaml :ref:`library` that Coq can load dynamically at
 runtime. Plugins are typically linked with the Coq OCaml API.
 
@@ -33,7 +36,7 @@ version<coq-lang>` in the :ref:`dune-project` file. For example, adding
 
 .. code:: dune
 
-    (using coq 0.7)
+    (using coq 0.8)
 
 to a :ref:`dune-project` file enables using the ``coq.theory`` stanza and other
 ``coq.*`` stanzas. See the :ref:`Dune Coq language<coq-lang>` section for more
@@ -95,9 +98,9 @@ The semantics of the fields are:
   ``lib/coq/user-contrib/<module_prefix>``, as is customary in the Make-based
   Coq package ecosystem.
 
-  For compatibility, Dune also installs, under the ``user-contrib``
-  prefix, the ``.cmxs`` files that appear in ``<ocaml_plugins>``. This
-  will be dropped in future versions.
+  For compatibility, Dune also installs, under the ``user-contrib`` prefix, the
+  ``.cmxs`` files that appear in ``<ocaml_plugins>``. This will be dropped in
+  future versions.
 
 - ``<coq_flags>`` are passed to ``coqc`` as command-line options. ``:standard``
   is taken from the value set in the ``(coq (flags <flags>))`` field in ``env``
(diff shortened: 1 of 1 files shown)


doc/coq.rst
```diff
@@ -494,6 +500,23 @@ All three of the theories we defined before were *private theories*. In order to
...
+
+We can also compose with theories that are installed. If we wanted to have a
+theory that depends on the Coq theory ``mathcomp.ssreflect`` we can add the
+following to our stanza:
+
+.. code:: dune
+
+  (coq.theory
+   (name my_mathcomp_theory)
+   (theories mathcomp.ssreflect))
```
Blaisorblade, line 514: What happens on overlapping paths? Can I use both `theories mathcomp` and `theories mathcomp.ssreflect` in the same project?

ejgallego: Yup, you can even today, as long as you declared these theories of course.

Dune will just create the corresponding -Q flags and pass them to coqc.

For installed libraries, if the library is in user-contrib then dune has to infer the name, so it uses an approximation, it is on the user to keep sanity on the naming so the "install substitution lemma" does work, that's the price to pay to have 2-way compatibility as we aim (for now)

Alizter: I've added a note about `theories mathcomp` while in principle it will work, if mathcomp starts using a coq.theory stanza and you have a local copy it obviously won't find the correct name. Therefore I've mentioned that it is advisable to use `mathcomp.ssreflect`.

doc/coq.rst
```diff
@@ -134,13 +137,15 @@ The semantics of the fields are:
   Interproject composition has been available since :ref:`Coq lang
   0.4<coq-lang>`.
 
-  As of today, Dune cannot depend on installed Coq theories. This restriction
-  will be lifted in the future. Note that composition with the Coq standard
-  library is supported, but in this case the ``Coq`` prefix has been made
-  available in a qualified way, since :ref:`Coq lang 0.2<coq-lang>`.
+  Composition with installed theories has been available since :ref:`Coq lang
+  0.8<coq-lang>`. Instaled theories do not have to be installed using Dune. Dune
```
christinerose, line 141:
```suggestion
  0.8<coq-lang>`. Installed theories do not have to be installed using Dune. Dune
```

doc/coq.rst
```diff
@@ -266,6 +271,8 @@ The supported Coq language versions (not the version of Coq) are:
 - ``0.7``: ``(mode )`` is automatically detected from the configuration of Coq
   and ``(mode native)`` is deprecated. The ``dev`` profile also no longer
   disables native compilation.
+- ``0.8``: Support for composition of installed theories; support for vos
```
christinerose, line 274:
```suggestion
- ``0.8``: Support for composition of installed theories; support for VoS
```

doc/coq.rst
```diff
@@ -500,6 +507,26 @@ All three of the theories we defined before were *private theories*. In order to
 depend on them, we needed to make them *public theories*. See the section on
 :ref:`public-private-theory`.
 
+Composing With Installed Theories
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We can also compose with theories that are installed. If we wanted to have a
+theory that depends on the Coq theory ``mathcomp.ssreflect`` we can add the
```
christinerose, line 514:
```suggestion
theory that depends on the Coq theory ``mathcomp.ssreflect``, we can add the
```

doc/coq.rst
```diff
@@ -500,6 +507,26 @@ All three of the theories we defined before were *private theories*. In order to
...
+following to our stanza:
+
+.. code:: dune
+
+  (coq.theory
+   (name my_mathcomp_theory)
+   (theories mathcomp.ssreflect))
+
+Note that ``mathcomp`` on its own would also work, since there would be a
+``matchcomp`` directory in ``user-contrib``, however it would not compose
```
christinerose, line 524:
```suggestion
``matchcomp`` directory in ``user-contrib``; however, it would not compose
```

christinerose (approved):
Minor formatting/grammar suggestions. One typo caught.

doc/coq.rst
Alizter: I'll change this to `.vos` as monospace.

doc/coq.rst
```diff
@@ -28,12 +28,14 @@ collection of Coq theories and plugins.
 The ``.v`` files of a theory need not be present as source files. They may also
 be Dune targets of other rules.
 
+Coq theories may come from the same workspace or from an install location.
```
ejgallego, line 31: Maybe "local" workspace? I would expand here a bit more as to illustrate what we mean, for example

"Dune will locate Coq theories first by looking at the set of stanzas defined in the current workspace plus the set of theories globally available for a given Coq install, similarly to what is done for OCaml.

Theories in the same scope will be used first, then public theories in the same workspace will be referenced, and if the desired theory is not found there, Dune will scan the `COQPATH` and `user-contrib` Coq global location to try to find the required theory"

doc/coq.rst
```diff
@@ -114,9 +116,10 @@ The semantics of the fields are:
   plugins. Starting with ``(lang coq 0.6)``, ``<ocaml_plugins>`` must contain
   public library names.
 
-- Your Coq theory can depend on other theories by specifying them in the
-  ``<coq_theories>`` field. Dune then passes to Coq the corresponding flags for
-  everything to compile correctly (this corresponds to the ``-Q`` flag for Coq).
+- Your Coq theory can depend on other theories, installed or in your workspace,
```
ejgallego, line 119: Just a question about terminology here, not sure we want to say "globally installed" to make things clearer.

doc/coq.rst
ejgallego, line 141: I would certainly move the second sentence to its own section, IMHO the way we lookup installed theories deserves its own entry.

ejgallego: Typo *instaled

doc/coq.rst
```diff
@@ -134,13 +137,15 @@ The semantics of the fields are:
...
+  0.8<coq-lang>`. Instaled theories do not have to be installed using Dune. Dune
+  will scan for theories installed in the ``user-contrib`` directory of the Coq
+  installation or those given in the ``COQPATH`` environment variable.
 
-  You may still use installed libraries in your Coq project, but there is
-  currently no way for Dune to know about it.
+  Composition with the Coq standard library is supported, but in this case the
+  ``Coq`` prefix has been made available in a qualified way, since :ref:`Coq
+  lang 0.2<coq-lang>`. Composition with installed theories is disabled if Coq is
+  found in the workspace, as that always takes precedence.
```
ejgallego, line 148: Maybe worth moving to its own section? We could be more clear here and explain that "when Coq is found in the workspace, it makes no sense to use globally installed theories as they'd need to be compiled with the local Coq".

doc/coq.rst
```diff
@@ -134,13 +137,15 @@ The semantics of the fields are:
...
-  available in a qualified way, since :ref:`Coq lang 0.2<coq-lang>`.
+  Composition with installed theories has been available since :ref:`Coq lang
+  0.8<coq-lang>`. Instaled theories do not have to be installed using Dune. Dune
+  will scan for theories installed in the ``user-contrib`` directory of the Coq
+  installation or those given in the ``COQPATH`` environment variable.
 
-  You may still use installed libraries in your Coq project, but there is
-  currently no way for Dune to know about it.
+  Composition with the Coq standard library is supported, but in this case the
+  ``Coq`` prefix has been made available in a qualified way, since :ref:`Coq
```
ejgallego, line 146: That seems a bit confusing to me, we should be more clear and say:
- if `(stdlib yes)`, the `Coq` is a added using -R + Require Import Prelude is assumed
- if `(stdlib no)`, then `Coq` can be added to the regular theory field, and it will be handled as a regular library.

ejgallego (approved):
Modulo a typo it looks good to me.

I would do some further rewriting as noted in the comments but I'm happy we do that in a subsequent PR.

Alizter:
@ejgallego Thanks for the explanation of the databases. I have gone ahead and cleaned up the wording a bit and will merge later.

PR #7386, opened 2023-03-23
fix: mkdir with trailing slash

rgrinberg:
to get us the ENOTDIR error code


rgrinberg:
This only works on MacOS

PR #7388, opened 2023-03-23
doc: move DSL sections from concepts to Reference

emillon:
As noted in #7325, these documents only contain reference info, so move them to the corresponding directory.


doc/dune-files.rst
```diff
@@ -562,6 +562,7 @@ use_standard_c_and_cxx_flags
    Control how flags coming grom ``ocamlc -config`` are passed to the C
    compiler command line.
 
+<<<<<<< HEAD
```
Alizter, line 565: conflict leftover

doc/reference/actions.rst
```diff
@@ -62,13 +62,14 @@ The following constructions are available:
 - ``(bash <cmd>)`` to execute a command using ``/bin/bash``. This is obviously
   not very portable.
 - ``(diff <file1> <file2>)`` is similar to ``(run diff <file1> <file2>)`` but
-  is better and allows promotion. See :doc:`promotion` for more details.
+  is better and allows promotion. See :doc:`../concepts/promotion` for more
+  details.
 - ``(diff? <file1> <file2>)`` is similar to ``(diff <file1>
   <file2>)`` except that ``<file2>`` should be produced by a part of the
   same action rather than be a dependency, is optional and will
```
christinerose, line 69:
```suggestion
  same action rather than be a dependency. It's optional and will
```
I'm not sure this should be a separate sentence like this or if this is what was meant, but the sentence as-is doesn't make sense. The part up until "than be a dependency" does, but what is optional and be consumed by `diff`?

doc/stanzas/library.rst
```diff
@@ -52,12 +52,12 @@ order to declare a multi-directory library, you need to use the
...
+  :doc:`reference/ordered-set-language`, where elements are module names and
   don't need to start with an uppercase letter. For instance, to exclude module
   ``Foo``, use ``(modules (:standard \ foo))``
 
 - ``(libraries <library-dependencies>)`` specifies the library's dependencies.
-  See :doc:`concepts/library-dependencies` for more details.
+  See :doc:`reference/library-dependencies` for more details.
 
 - ``(wrapped <boolean>)`` specifies whether the library modules should be
   available only through the top-level library module, or if they should all be
```
christinerose, line 63: Is this toplevel (like the OCaml toplevel) or the top level or something? If the latter, it's fine as it is. If the former, write it as one word: toplevel. It appears to be the latter...just double checking! 😄

christinerose (approved):

doc/stanzas/library.rst
Alizter: When you have a library you can specify if it is "wrapped" or not. All this means is that a certain module called the "top-level module" is used as a wrapper around the rest of the library. Say you had 3 modules in your library:
```
secret.ml
my_logic.ml
my_lib.ml
```
Then you could imagine my_lib.ml being the top-level module but it only gives users access to my_logic say. Because you don't want people knowing what secret does.

When a library is unwrapped all the modules are available as "top-level modules".

christinerose: Great! Thanks for explaining.

doc/reference/actions.rst
emillon: file2 is optional and is consumed by `diff?`. I agree that the sentence is not great at the moment. I'll merge the move and will try to improve that.

PR #7391, opened 2023-03-23
refactor: thread pooling improvements

rgrinberg:
* Add some more documentation
* Micro optimize joining dead threads. Shave off a mutation when we
  don't need to do any work.

### DIFF
--- a/src/thread_pool/thread_pool.ml
+++ b/src/thread_pool/thread_pool.ml
@@ -1,5 +1,16 @@
 open Stdune
 
+(** The thread pool implementation allows callers to offload (non raising) work
+    to background threads.
+
+    The pool maintains a [min_workers, max_workers] number of threads active and
+    dispatches the work to them.
+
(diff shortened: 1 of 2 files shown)


rgrinberg:
@snowleopard this PR addresses your comments from an earlier PR

src/thread_pool/thread_pool.ml
```diff
@@ -1,5 +1,16 @@
...
 
+(** The thread pool implementation allows callers to offload (non raising) work
+    to background threads.
+
+    The pool maintains a [min_workers, max_workers] number of threads active and
+    dispatches the work to them.
+
+    There's no attempt to dispatch the work fairly.
+
+    There's no ways to for callers to pick an underlying thread to use for some
```
snowleopard, line 11:
```suggestion
    There's no way for the caller to pick an underlying thread to use for some
```

snowleopard (approved):
Thanks!

PR #7372, opened 2023-03-21
feature: async digest handling

rgrinberg:
Move the computation of digests to a background thread. For now, this is guarded behind a configuration option that requires this to be enabled.

Signed-off-by: Rudi Grinberg <me@rgrinberg.com>

Issue #1282, opened 2018-09-17
Making hashing of values faster


jeremiedimino:
Looking at the compiler C code, it seems that it should be easy to provide a C function that does the same as `fun x -> Digest.to_string (Marshal.to_string x [])` but with using constant memory. We should look at adding such a function in the stdlib, that should make Dune a bit faster.

rgrinberg:
Note that this many not be enough for things like digesting maps properly. Since we'd like to digest things in the key order. Ideally, a new primitive would allow us to express that.

nojb:
I could spare a bit of time to work on this suggestion, if it helps. I am very very interested in anything that can make `dune` faster.

diml:
@nojb that would be great! To be clear, I don't know exactly how much we'd win, but since dune computes quite a lot of digests, I believe the win wouldn't be negligible. At least it would reduce GC pressure.

nojb:
Yes, I suspect the effect won't be dramatic, but it can't hurt. Also, it just seems a useful function to have.

I will give it a try when I can and report back here.

jchavarri:
fwiw I landed on this issue today, while I was profiling dune builds in a largish synthetic repo (50k modules) and the 2nd most time consuming function is `caml_MD5transform`:

```
$ perf report --no-children
+   14.38%  dune          dune                [.] mark_slice
+   13.88%  dune          dune                [.] caml_MD5Transform
+    7.58%  dune          dune                [.] caml_page_table_lookup
+    5.25%  dune          dune                [.] sweep_slice
+    4.00%  ocamldep.opt  [kernel.vmlinux]    [k] filemap_map_pages
+    3.93%  ocamldep.opt  [kernel.vmlinux]    [k] clear_page_rep
+    3.76%  ocamldep.opt  [kernel.vmlinux]    [k] copy_page
+    3.42%  ocamldep.opt  [kernel.vmlinux]    [k] unmap_page_range
+    2.31%  ocamldep.opt  ocamldep.opt        [.] fill_hashtable
+    1.77%  dune          [kernel.vmlinux]    [k] copy_user_generic_string
+    1.27%  ocamldep.opt  ld-2.28.so          [.] _dl_relocate_object
+    1.22%  ocamldep.opt  ocamldep.opt        [.] init_frame_descriptors
+    1.21%  ocamldep.opt  [kernel.vmlinux]    [k] asm_exc_page_fault
[...]

PR #7392, opened 2023-03-23
feature: switch to xxHash for digests

rgrinberg:
xxHash is a far faster algorithm than md5.

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -710,6 +710,7 @@ end
 module Library = struct
   module File_kind = struct
     type t =
+      | Header
       | C
       | Ml
       | Mli
@@ -721,6 +722,7 @@ module Library = struct
       let fn = Filename.basename fn in
       let i = try String.index fn '.' with Not_found -> String.length fn in
       match String.sub fn ~pos:i ~len:(String.length fn - i) with
+      | ".h" -> Some Header
       | ".c" -> Some C
       | ".ml" -> Some Ml
       | ".mli" -> Some Mli
@@ -756,6 +758,7 @@ module Library = struct
       in
       match kind with
       | C -> base ^ ".c"
+      | Header -> base ^ ".h"
       | _ -> (
         let ext =
           match kind with
@@ -812,7 +815,7 @@ module Library = struct
     let modules =
       List.fold_left files ~init:StringSet.empty ~f:(fun acc (fn, kind) ->
           match (kind : File_kind.t) with
-          | C -> acc
+          | Header | C -> acc
           | Ml | Mli | Mll | Mly ->
             let module_name =
               let fn = Filename.basename fn in
@@ -834,7 +837,7 @@ module Library = struct
             let mangled = Wrapper.mangle_filename wrapper fn kind in
             let dst = build_dir ^/ mangled in
             match kind with
-            | C ->
+            | Header | C ->
               copy fn dst;
               Fiber.return [ mangled ]
             | Ml | Mli ->
--- a/boot/libs.ml
+++ b/boot/libs.ml
@@ -17,6 +17,7 @@ let local_libraries =
   ; ("src/dune_console", Some "Dune_console", false, None)
   ; ("src/memo", Some "Memo", false, None)
   ; ("src/dune_metrics", Some "Dune_metrics", false, None)
+  ; ("vendor/xxh", Some "Dune_xxh", false, None)
   ; ("src/dune_digest", Some "Dune_digest", false, None)
   ; ("src/dune_sexp", Some "Dune_sexp", false, None)
   ; ("src/ocaml-config", Some "Ocaml_config", false, None)
--- a/src/dune_digest/dune
+++ b/src/dune_digest/dune
@@ -1,5 +1,5 @@
 (library
  (name dune_digest)
- (libraries dune_metrics stdune)
+ (libraries dune_metrics dune_xxh stdune)
  (instrumentation
   (backend bisect_ppx)))
--- a/src/dune_digest/dune_digest.ml
+++ b/src/dune_digest/dune_digest.ml
@@ -1,8 +1,8 @@
 open Stdune
+open Dune_xxh
 
 type t = string
 
-module D = Stdlib.Digest
 module Set = String.Set
 module Map = String.Map
 module Metrics = Dune_metrics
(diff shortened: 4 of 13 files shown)


jchavarri:
ref: https://github.com/ocaml/dune/issues/1282.

Alizter:
Some preliminary bench results:

| | Main | PR | %diff |
| --- | --- | --- | --- |
| Clean | 108.8421618938446 | 102.72804498672485 | -5.61742% |
| Null | 0.9267170429229736 | 0.8870739936828613 | |
| | 0.910944938659668 | 0.8958990573883057 | |
| | 0.9389081001281738 | 0.9134318828582764 | |
| | 0.9436089992523193 | 0.901669979095459 | |
| | 0.9183239936828613 | 0.9315218925476074 | |
| Avg. Null | 0.9277006149292 | 0.9059193611145 | -2.34788% |

rgrinberg:
This should play nicely with https://github.com/ocaml/dune/pull/7372

(although some tweaks still remain)

rgrinberg:
@jchavarri let's see if this has an effect on your internal builds

rgrinberg:
@Alizter thanks for the benchmark. With this PR, it should be enough to catch up to make in the HoTT build I hope.

snowleopard (changes requested):
We can't use weak hashing algorithms, so this should obviously not be used by default, but also I'm really not keen on adding a 7KLOC dependency for something that isn't used by default.

snowleopard:
> Some preliminary bench results:

What is this building?

Alizter:
@snowleopard that is the bench you get from `make bench` not sure what it is building but I think it is just the dune executable.

Why is a weak hashing algorithm not usable?

rgrinberg:
> We can't use weak hashing algorithms, so this should obviously not be used by default, but also I'm really not keen on adding a 7KLOC dependency for something that isn't used by default.

What's md5 if not a weak hashing algorithm? :)

Suggestions for other hashing algorithms are welcome of course. But it seems like we're making the wrong trade-off if we're defaulting to cryptographic hashing algorithm. I don't use any build artifacts that I don't produce myself, and I highly doubt any of our users do. Things might change if we have a distributed build cache, but it remains to be seen how adoption would look like. Or are there other situations where a strong hash would be useful?

rgrinberg:
> 7KLOC dependency

I could remove the variants of the hashing functions that we don't use, but honestly it sounds seem like it's worth the trouble. I doubt our 11 meg binary would shrink much.

snowleopard:
> @snowleopard that is the bench you get from `make bench` not sure what it is building but I think it is just the dune executable.

I see. 5-6% difference sounds close to the noise you'd get from any change, e.g. due to the executable layout changing or just plain build non-determinism, so I'm not convinced by this figure. Did you run the benchmark just once? On your personal machine that was running a bunch of other applications? I'd like to see the results of the monorepo benchmark.

> Why is a weak hashing algorithm not usable?

When collisions are computationally easy, one will hit them whether on purpose or not, and hitting a hash collision in Dune means getting incorrect build results. All levels of Dune caching (workspace-local, shared and distributed caches) assume cryptographically secure hashes.

> What's md5 if not a weak hashing algorithm? :)

Finding MD5 collisions is computationally pretty hard. Not impossible these days, though, so I'd welcome stronger hashes.

> Suggestions for other hashing algorithms are welcome of course. 

I suggest trying BLAKE3, which is available in Cryptokit.

> But it seems like we're making the wrong trade-off if we're defaulting to cryptographic hashing algorithm

I'm not quite sure what you are suggesting. To tradeoff build correctness for 5% of performance gain? :)

snowleopard:
Also, just as a general point: we should keep MD5 available while we are transitioning from Jenga to Dune internally. We can't just switch to a different hashing scheme: that would mean that we can't reuse the same build artifact cache between Jenga and Dune, thus doubling disk usage, which would be a severe problem for us. So, whatever new hashing algorithm we introduce, we should keep MD5 as a configuration option.

emillon:
To add to @snowleopard's point: I agree that xxhash is not suited to what we're doing with dune.
I think that the threat model that dune uses is not too different from git's use of a hashing function: the fact that there are collisions is a good sign that something should be fixed, but is not an immediate problem. Concretely, you can set up some files and rules that will confuse dune, but there's no way at the moment to use that weakness to inject a different file. You need a preimage attack for that, and even md5 is still relatively immune to that (the cost is > 2**100).
So, yes, md5 is broken and slow, which are two reasons to look for alternatives, but the replacement function needs some properties that xxhash do not have.
(in contrast, it could be a good replacement for `Hashtbl.hash`, but I'm not sure how important that function is in our profiles)

Alizter:
So I am under the impression that there are two critical points to Dune hashing being slow:
1. Marshalling
2. Hashing Algorithm

Since Rudi chose an algorithm which is extremely fast, this confirms my suspicion that both together are causing the observable slowdown seen in builds such as HoTT.

In fact I ran some detailed benchmarks comparing this with main and make: https://github.com/HoTT/Coq-HoTT/pull/1687#issuecomment-1482957815

I'll resummarise here:


| Builder | `make` | `dune` | xxH `dune` |
| --- | --- | --- | --- |
| Mean (s) | 65.519 | 68.799 | 66.285 |
| Standard deviation σ | 0.325 | 0.292 | 0.222 |
| Min | 65.068 | 68.450 | 66.041 |
| Max | 66.031 | 69.175 | 66.728 |

ejgallego:
@Alizter what's the size of the .vo build for HoTT?

You can indeed just put all the .vo files in a folder and measure with a simple program how long it does takes to hash them.

Alizter:
The total size of all the vo files is `98M` according to `du`.

Here is a hyperfine bench of the vo files with md5sum vs hxxsum:
```
[ali@allosaurus:~/HoTT]$ hyperfine -w 10 'find all_vo/|xargs md5sum |tee all_vo.md5'
Benchmark 1: find all_vo/|xargs md5sum |tee all_vo.md5
  Time (mean ± σ):     130.0 ms ±   1.7 ms    [User: 115.1 ms, System: 19.3 ms]
  Range (min … max):   126.6 ms … 133.0 ms    23 runs


[ali@allosaurus:~/HoTT]$ hyperfine -w 10 'find all_vo/|xargs xxhsum |tee all_vo.md5'
Benchmark 1: find all_vo/|xargs xxhsum |tee all_vo.md5
  Time (mean ± σ):       3.7 ms ±   0.2 ms    [User: 0.9 ms, System: 4.4 ms]
  Range (min … max):     3.1 ms …   4.4 ms    510 runs
```

ejgallego:
Thanks for getting the numbers, 100ms is within expected (tho we should use the same exact hash Dune uses to be sure)

If so, that would hardly explain the HoTT slowdown.

jchavarri:
I can see significant gains on our internal builds when using this branch. In particular, the Melange build gets 1.16x faster. An OCaml executable build I tested (for the api server) is 1.1x faster. 🎉 

I'll keep an eye on https://ocaml.github.io/dune/dev/bench/ after it's merged.

Alizter:
@ejgallego that's why I said it's not just the hashing algorithm contributing to the slowdown but also the marshaling.

ejgallego:
> @ejgallego that's why I said it's not just the hashing algorithm contributing to the slowdown but also the marshaling.

What data are we marshalling ?

Alizter:
@ejgallego 
https://github.com/ocaml/dune/blob/bb61fb7f503f993f1b46d7ba03d5f919ddb93583/src/dune_digest/dune_digest.ml#L68-L76

ejgallego:
> @ejgallego
> 
> https://github.com/ocaml/dune/blob/bb61fb7f503f993f1b46d7ba03d5f919ddb93583/src/dune_digest/dune_digest.ml#L68-L76

Ummm, that's the generic hash, sure, but that part should be rarely used in Coq rules, certainly not for .vo files I think.

Are we generically hashing some large in-memory structs? If so, which ones?

PR #7395, opened 2023-03-23
fix: disable cutoff where it hurts concurrency

rgrinberg:
Make it possible to enable it with the "speculative_cutoff" internal
config option

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 Unreleased
 ----------
 
+- Dune in watch mode no longer builds concurrent rules in serial (#7395
+  @rgrinberg, @jchavarri)
+
 - `dune coq top` now correctly respects the project root when called from a
   subdirectory. However, absolute filenames passed to `dune coq top` are no
   longer supported (due to being buggy) (#7357, fixes #7344, @rlepigre and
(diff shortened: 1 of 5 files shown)


snowleopard (approved):
The changes look good but I'm not sure about the name: "speculative" doesn't sound like the right word here because there isn't much speculation going on. 

I'd use something longer but hopefully clearer: `cutoffs_that_reduce_concurrency_in_watch_mode`.

snowleopard (approved):
Thanks!

CHANGES.md
```diff
@@ -1,6 +1,9 @@
 Unreleased
 ----------
 
+- Dune in watch mode no longer builds concurrent rules in serial (#7395
```
Alizter, line 4: fixes https://github.com/ocaml/dune/issues/5549 ?

rgrinberg: We'll claim that we fixed it once we bring back cutoff without the penalty.

rgrinberg:
Waiting for @jchavarri to confirm the fix

jchavarri (approved):
Thanks @rgrinberg !

I have spent some time testing this branch in watch mode, adding changes in multiple files on both in Ahrefs internal repo and the [public repro](https://github.com/jchavarri/dune_repro_issue_5549) and concurrency stayed optimal at all times during both tests. 🎉

snowleopard:
@jchavarri Awesome, thank you! I'll give it a try internally too once I'm back to work.

PR #7398, opened 2023-03-23
melange: account for `melange.compile_flags` when generating merlin config

anmonteiro:


test/blackbox-tests/test-cases/melange/merlin-compile-flags.t
```diff
@@ -0,0 +1,37 @@
+Show that the merlin config knows about melange.compile_flags
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.7)
```
jchavarri, line 4: Can use 3.8
```suggestion
  > (lang dune 3.8)
```

test/blackbox-tests/test-cases/melange/merlin.t
```diff
@@ -64,11 +64,12 @@ Dump-dot-merlin includes the melange flags
   B $TESTCASE_ROOT/_build/default/.output.mobjs/melange
   S $TESTCASE_ROOT
   # FLG -ppx '/MELC_COMPILER -as-ppx'
-  # FLG -w @1..3@5..28@30..39@43@46..47@49..57@61..62@67@69-40 -strict-sequence -strict-formats -short-paths -keep-locs
+  # FLG -w @1..3@5..28@30..39@43@46..47@49..57@61..62@67@69-40 -strict-sequence -strict-formats -short-paths -keep-locs -g
```
jchavarri, line 67: Interesting that `-g` is added by default by dune. Do you think it has any perf implications in Melange, or in other words, is it handled any differently than in OCaml?

src/dune_rules/merlin/merlin.ml
```diff
@@ -351,7 +351,11 @@ module Unprocessed = struct
       Path.Set.singleton
       @@ obj_dir_of_lib `Private mode (Obj_dir.of_local obj_dir)
     in
-    let flags = Ocaml_flags.common flags in
+    let flags =
+      match mode with
+      | `Ocaml -> Ocaml_flags.common flags
+      | `Melange -> Ocaml_flags.get flags Melange
+    in
```
jchavarri, line 358: Do you know why `Ocaml_flags.get` isn't used also for the OCaml mode? This is actually the only usage of `Ocaml_flags.common` that I could find in the whole repo. When building modules we use `get`:

https://github.com/ocaml/dune/blob/ece416166f6696fa565dc2f36594ec2e0ecd0f85/src/dune_rules/module_compilation.ml#L205

Maybe this can be replaced with

```suggestion
    let flags = Ocaml_flags.get flags mode
```

The only change is that merlin would take into account specific flags for OCaml compilation, but I only see that as an improvement? 🤔 

If you decide to do that, we could remove `common` from `Ocaml_flags.mli`.

jchavarri (commented):
Looks good, but I wonder if it can be simplified further.

src/dune_rules/merlin/merlin.ml
anmonteiro: Maybe. One thing I noticed is that `common` doesn't have `-g`, which you can actually see in the test output.

rgrinberg: This use of `common` does seem weird. We can leave it alone for now and save it for another PR

test/blackbox-tests/test-cases/melange/merlin.t
anmonteiro: Melange sets `-g` by default even if not passed, so this is fine.

src/dune_rules/merlin/merlin.ml
anmonteiro: I don't know if there's anything actionable here. I'm assuming "leave it alone for now" means "keep common" since that's how it was in the first place.

rgrinberg: The PR looks fine for me. I'm waiting for @jchavarri to approve

jchavarri (approved):
Looks good, thanks for the fix @anmonteiro !

PR #7401, opened 2023-03-24
Revert "Update opam-file-format to 2.1.5 (#7328)"

Leonidas-from-XIV:
This reverts commit 649031728346482a83441c26c0bbe06ecd89cce2 as @rgrinberg has found that the format that is generated has changed slightly.


Alizter (approved):

rgrinberg:
Just a note that we aren't against using the new format, it's just that it has to be gated behind a version check  as we need to support the old behavior as well.

PR #7404, opened 2023-03-24
doc: improve synopsis for all commands

emillon:
There are several improvements in there:

- build: do not mention "installable targets"
- ensure all command groups are documented
- rules: dune rules dumps all rules, not just internal ones
- upgrade: is not specific to jbuilder
- use a fixed style of command docs (Capital, dot)

### DIFF
--- a/bin/cache.ml
+++ b/bin/cache.ml
@@ -5,7 +5,7 @@ open Import
 
 let trim =
   let info =
-    let doc = "Trim the Dune cache" in
+    let doc = "Trim the Dune cache." in
     let man =
       [ `P "Trim the Dune cache to a specified size or by a specified amount."
       ; `S "EXAMPLES"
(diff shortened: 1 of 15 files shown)


Alizter:
Could you add some document station about writing doc somewhere? For now hacking.rst would be enough. We should mention guidelines like not ending sentences with a full stop in man pages.

Alizter:
Honestly, I think using full stops in command synopses looks way better / more polished.

emillon:
ok, I added the dots back.

bin/init.ml
```diff
@@ -207,7 +207,7 @@ let project =
      print_completion "project" common.name
 
 let group =
-  let doc = "Command group for initializing dune components" in
+  let doc = "Command group for initializing dune components." in
```
Alizter, line 210: Dune

bin/cache.ml
```diff
@@ -80,7 +80,7 @@ let size =
 
 let command =
   let info =
-    let doc = "Manage the shared cache of build artifacts" in
+    let doc = "Manage the shared cache of build artifacts." in
```
Alizter, line 83: Manage Dune's shared cache...

Alizter (approved):

PR #7405, opened 2023-03-24
doc: mention correct exit codes in the manpage

### LINKED ISSUE #7296: Exit codes don't match help text
## Expected Behavior
`dune --help` output says:
```
EXIT STATUS
       dune exits with the following status:

       0   on success.

       123 on indiscriminate errors reported on standard error.

       124 on command line parsing errors.

       125 on unexpected internal errors (bugs).
```
This is the default from Cmdliner.

## Actual Behavior
Possible exit codes are 0, 1 and 130:
https://github.com/ocaml/dune/blob/449cbd322ca2a2c50d437d2a2b69d81d5bdf0439/bin/main.ml#L90-L106

## Reproduction

1. Run dune with invalid arguments: `dune --foo`.
2. Check exit code: `echo $?` – currently outputs 1, not 124.

## Specifications

- Version of `dune` (output of `dune --version`): 3.7.0
- Operating system (distribution and version): Xubuntu 22.04.2

emillon:
Fixes #7296

This passes the correct `~exits` docs to cmdliner.
In addition this adds a tiny abstraction layer to exit codes so that we keep this in sync.

### DIFF
--- a/bin/exit_code.ml
+++ b/bin/exit_code.ml
@@ -1,0 +1,18 @@
+type t =
+  | Success
+  | Error
+  | Signal
+
+let all = [ Success; Error; Signal ]
+
+let code = function
+  | Success -> 0
+  | Error -> 1
+  | Signal -> 130
+
+let doc = function
+  | Success -> "on success."
+  | Error -> "if an error happened."
+  | Signal -> "if it was interrupted by a signal."
+
+let info e = Cmdliner.Cmd.Exit.info (code e) ~doc:(doc e)
--- a/bin/exit_code.mli
+++ b/bin/exit_code.mli
@@ -1,0 +1,10 @@
+type t =
+  | Success
+  | Error
+  | Signal
+
+val all : t list
+
+val info : t -> Cmdliner.Cmd.Exit.info
+
+val code : t -> int
--- a/bin/import.ml
+++ b/bin/import.ml
@@ -38,7 +38,15 @@ include struct
   open Cmdliner
   module Term = Term
   module Manpage = Manpage
-  module Cmd = Cmd
+
+  module Cmd = struct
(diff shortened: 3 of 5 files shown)


emillon:
Ah, this does not fix `dune --help`, `dune build --help` etc.

emillon:
This should be good now.

Alizter (approved):

bin/exit_code.ml
```diff
@@ -0,0 +1,18 @@
...
+
+let code = function
+  | Success -> 0
+  | Error -> 1
+  | Signal -> 130
+
+let doc = function
+  | Success -> "on success"
+  | Error -> "if an error happened"
+  | Signal -> "if it was interrupted by a signal"
```
Alizter, line 16: Actually I would prefer
```suggestion
  | Success -> "On success."
  | Error -> "If an error happened."
  | Signal -> "If it was interrupted by a signal."
```
And we should document what happens with code_error.
At the moment it looks like:
```
EXIT STATUS
       build exits with the following status:

       0   on success

       1   if an error happened

       130 if it was interrupted by a signal
```
Which I think looks kind of poor.

Alizter (changes requested):

bin/exit_code.ml
emillon: the defaults for cmdliner use lowercase, which I read as a sentence:

- 0 on success
- 1 if an error happened
- 130 it was interrupted by a signal

emillon: I do agree about the dots though (once again my north star with that is to be consistent with the cmdliner docs)

Alizter (approved):
Sorry I did not mean to block my confirmation.

PR #4050, opened 2020-12-20
[coq] Add .vos support

rgrinberg:
We now build and install .vos files. This requires us to parse coqdep
with -vos and setup a .vos rule for every file.

PR #7406, opened 2023-03-24
Support for Coq "vos" builds

rlepigre:
This is an alternative to https://github.com/ocaml/dune/pull/4050, implementing @ejgallego's idea from https://github.com/ocaml/dune/pull/4050#issuecomment-1300916807.

The "vos" mode is enabled using `(mode vos)` in `coq.theory` stanzas.

CC @Alizter.

### DIFF
--- a/bin/coq/coqtop.ml
+++ b/bin/coq/coqtop.ml
@@ -81,7 +81,7 @@ let term =
           let stanza =
             Dune_rules.Coq_sources.lookup_module coq_src coq_module
           in
-          let args, use_stdlib, coq_lang_version, wrapper_name =
+          let args, use_stdlib, coq_lang_version, wrapper_name, mode =
             match stanza with
             | None ->
               User_error.raise
@@ -93,17 +93,24 @@ let term =
                   ~dir_contents:dc theory coq_module
               , theory.buildable.use_stdlib
               , theory.buildable.coq_lang_version
-              , Dune_rules.Coq_lib_name.wrapper (snd theory.name) )
+              , Dune_rules.Coq_lib_name.wrapper (snd theory.name)
+              , theory.buildable.mode )
             | Some (`Extraction extr) ->
               ( Dune_rules.Coq_rules.coqtop_args_extraction ~sctx ~dir extr
                   coq_module
               , extr.buildable.use_stdlib
               , extr.buildable.coq_lang_version
-              , "DuneExtraction" )
+              , "DuneExtraction"
+              , extr.buildable.mode )
           in
           let* (_ : unit * Dep.Fact.t Dep.Map.t) =
             let deps_of =
-              Dune_rules.Coq_rules.deps_of ~dir ~use_stdlib ~wrapper_name
+              let mode =
+                match mode with
+                | None -> Dune_rules.Coq_mode.VoOnly
+                | Some mode -> mode
+              in
+              Dune_rules.Coq_rules.deps_of ~dir ~use_stdlib ~wrapper_name ~mode
                 ~coq_lang_version coq_module
             in
             Action_builder.(run deps_of) Eager
--- a/doc/coq.rst
+++ b/doc/coq.rst
@@ -152,6 +152,12 @@ The semantics of the fields are:
   Previous versions of Dune before 3.7 would disable the native rules depending
   on whether or not the ``dev`` profile was selected.
 
+- From version :ref:`Coq lang 0.8<coq-lang>` onwards, ``(mode vos)`` makes it
+  so that only Coq compiled interface files are produced for the theory. This
+  is mainly useful in conjunction with ``dune coq top``, since this makes the
+  compilation of dependencies much faster (thought the proofs they contain are
+  not checked).
+
 Coq Dependencies
 ~~~~~~~~~~~~~~~~
 
--- a/src/dune_rules/coq/coq_rules.ml
+++ b/src/dune_rules/coq/coq_rules.ml
@@ -132,6 +132,16 @@ let coqc_native_flags ~sctx ~dir ~theories_deps ~theory_dirs
       ; "-native-compiler"
       ; "ondemand"
       ]
+  | VosOnly ->
+    Command.Args.As
+      [ "-vos"
+      ; "-w"
+      ; "-deprecated-native-compiler-option"
+      ; "-w"
+      ; "-native-compiler-disabled"
+      ; "-native-compiler"
+      ; "ondemand"
+      ]
   | Native ->
     let args =
       let open Action_builder.O in
@@ -294,7 +304,7 @@ let setup_coqdep_for_theory_rule ~sctx ~dir ~loc ~theories_deps ~wrapper_name
   let file_flags =
     [ Command.Args.S
         (coqc_file_flags ~dir ~theories_deps ~wrapper_name ~boot_type ~ml_flags)
-    ; As [ "-dyndep"; "opt" ]
+    ; As [ "-dyndep"; "opt"; "-vos" ]
     ; Deps sources
     ]
   in
@@ -350,10 +360,15 @@ let get_dep_map ~dir ~wrapper_name : Path.t list Dep_map.t Action_builder.t =
       ; ("entry 2", Dyn.list Path.to_dyn r2)
       ]
 
-let deps_of ~dir ~boot_type ~wrapper_name coq_module =
+let deps_of ~dir ~boot_type ~wrapper_name ~mode coq_module =
   let open Action_builder.O in
   let vo_target =
-    Path.Build.set_extension ~ext:".vo" (Coq_module.source coq_module)
+    let ext =
+      match mode with
+      | Coq_mode.VosOnly -> ".vos"
+      | _ -> ".vo"
+    in
+    Path.Build.set_extension ~ext (Coq_module.source coq_module)
   in
   get_dep_map ~dir ~wrapper_name >>= fun dep_map ->
   match Dep_map.find dep_map (Path.build vo_target) with
@@ -397,6 +412,12 @@ let generic_coq_args ~sctx ~dir ~wrapper_name ~boot_type ~mode ~coq_prog
     Command.Args.dyn coq_flags (* stanza flags *)
   in
   let coq_native_flags =
+    let mode =
+      (* Tweak the modes for coqtop since it has no "-vos" option *)
+      match (mode, coq_prog) with
+      | Coq_mode.VosOnly, `Coqtop -> Coq_mode.VoOnly
+      | _ -> mode
+    in
     coqc_native_flags ~sctx ~dir ~theories_deps ~theory_dirs ~mode
   in
   let file_flags =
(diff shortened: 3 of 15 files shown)


src/dune_rules/coq/coq_rules.ml
```diff
@@ -132,6 +132,16 @@ let coqc_native_flags ~sctx ~dir ~theories_deps ~theory_dirs
       ; "-native-compiler"
       ; "ondemand"
       ]
+  | VosOnly ->
+    Command.Args.As
+      [ "-vos"
```
Alizter, line 137: Since you add vos in file_flags do you need this here? Can be the same as VoOnly I think.

Alizter:
What happens when a user builds once in `VoOnly` stops the build realising they should have `VosOnly` and then continues? Doesn't this make the build inconsistent? IIRC Coq loads vo files if they are available and then vos files in vos mode.

rlepigre:
Since you configure `vo` or `vos` per theory, you can't have both `vo` and `vos` at the same time. Will that not save us?

Alizter:
The issue is that Coq goes and finds the files itself so if you have a dirty build directory you might corrupt the build.

Alizter:
What I am suggesting is the opposite situation. If you had 3 files:
```
A.v -> B.v -> C.v
```
Now if I had `(mode vo)` and built `B.vo` I would have `A.vo` and `B.vo` in my build directory. Now if I switch to `(mode vos)` I will and build `C.vos`, Dune will give us: `A.vos`, `B.vos` and `C.vos`. However when compiling `C.vos`, `coqc` would have used `B.vo` rather than `B.vos`.

Alizter:
Actually nevermind. Reading the Coq refman I see that `vos` is indeed preferred over `vo`. https://coq.inria.fr/refman/practical-tools/coq-commands.html#compiled-interfaces-produced-using-vos

src/dune_rules/coq/coq_rules.ml
rlepigre: Yeah, the other one (variable `file_flags`) is for `coqdep`.

rlepigre: I opted to always call `coqdep` with the `-vos` option, so that we can share the dependency analysis between modes.

bin/coq/coqtop.ml
```diff
@@ -93,17 +93,24 @@ let term =
...
+              , Dune_rules.Coq_lib_name.wrapper (snd theory.name)
+              , theory.buildable.mode )
             | Some (`Extraction extr) ->
               ( Dune_rules.Coq_rules.coqtop_args_extraction ~sctx ~dir extr
                   coq_module
               , extr.buildable.use_stdlib
               , extr.buildable.coq_lang_version
-              , "DuneExtraction" )
+              , "DuneExtraction"
+              , extr.buildable.mode )
```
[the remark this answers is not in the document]
rlepigre: No idea, I never actually used extraction. But I don't see why it shouldn't, and it felt like it would make sense to use the information since we have it.

ejgallego (changes requested):
It is a pity we can't be more dynamic yet but the patch is a start thanks!

Needs documentation / changes entry, rest looks good to me!

Alizter:
One more design consideration is if we want users to have to edit their stanzas at all. Perhaps it would be OK to have both rules at the same time as they are independent. We could even add `vos` to the `check` alias for example.

ejgallego:
> What happens when a user builds once in `VoOnly` stops the build realising they should have `VosOnly` and then continues? Doesn't this make the build inconsistent? IIRC Coq loads vo files if they are available and then vos files in vos mode.

You don't have that problem as all the `.vo` targets in the first run become stale (as rules won't generate them anymore) so Dune will clean them up.

ejgallego:
> @ejgallego It won't become nondeterministic see my retort above.

If I understood your comment above, you actually are giving an example of a non-determinitic build for C.vos, as it now depends on what was built before.

Alizter:
No, according to the Coq refman, vos is always preferred in -vos mode.

ejgallego:
Tho indeed maybe the race is for the vo objects, let me check coq's source

rlepigre:
One question I have is: can we somehow make it so that we can configure the mode in the profile? Is it already possible with what I did?

Alizter:
@rlepigre We can detect the `-vos` flag in `(coq (flags ...))` if a user has that and prefer the mode that way. I think it is better to do that in a later PR though.

If I am correct about being able to have both the vo and vos rules at the same time, then that would mean we wouldn't need it anyway.

Blaisorblade:
@Alizter coqc _can_ mix vo and vos files, so you do need separate directories; also the `vo` build overrides `vos`/`vok` files but dune might be able to handle _that_. And since we need separate profiles, can we use that to enable `vos` mode?

Let me [quote the manual](https://coq.inria.fr/refman/practical-tools/coq-commands.html#compiled-interfaces-produced-using-vos) (since there's no anchor to link)

> **Interaction with standard compilation**
>
> When compiling a file foo.v using coqc in the standard way (i.e., without -vos nor -vok), an empty file foo.vos and an empty file foo.vok are created in addition to the regular output file foo.vo. If coqc is subsequently invoked on some other file bar.v using option -vos or -vok, and that bar.v requires foo.v, if Coq finds an empty file foo.vos, then it will load foo.vo instead of foo.vos.
>
> The purpose of this feature is to allow users to benefit from the -vos option even if they depend on libraries that were compiled in the traditional manner (i.e., never compiled using the -vos option).

ejgallego:
Indeed I'm not sure we have the right mechanism to perform this kind of selection at runtime, it seems to me that using flags is not better than using `mode`, as both require updating the stanza which we def don't want.

I can't check the exact dependency rules now, so indeed the next step would be to see if we can enable both modes without getting unsoundness as Ali proposes.

In the immediate term, I propose we merge this PR once the docs are added, and continue the discussion in the issue / subsequent PR

rlepigre:
I guess I need to change the Coq language version @ejgallego?

Alizter:
@rlepigre You don't need to bump the coq lang version since for Dune 3.8 it will be 0.8 and we haven't released Dune yet.

ejgallego:
> @rlepigre You don't need to bump the coq lang version since for Dune 3.8 it will be 0.8 and we haven't released Dune yet.

Indeed, thanks @Alizter ; the changelog of language versions should make this clear by maybe adding the Dune version next to the coq lang version https://github.com/ocaml/dune/blob/main/doc/coq.rst#coq-language-version 

Note that indeed we bumped the lang in the implementationhttps://github.com/ocaml/dune/blob/main/src/dune_rules/coq/coq_stanza.ml#L4
but not in the docs.

Alizter:
It will be bumped in the docs, that is what https://github.com/ocaml/dune/pull/7384 is about.

rlepigre:
Sorry, I'm just only seeing your last messages now: I pushed some documentation, but maybe not exactly what you had in mind.

bin/coq/coqtop.ml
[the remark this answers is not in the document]
rlepigre: Should we do anything here @Alizter?

doc/coq.rst
```diff
@@ -152,6 +152,12 @@ The semantics of the fields are:
   Previous versions of Dune before 3.7 would disable the native rules depending
   on whether or not the ``dev`` profile was selected.
 
+- Form version :ref:`Coq lang 0.8<coq-lang>` onwards, ``(mode vos)`` makes it
```
Alizter, line 155:
```suggestion
- From version :ref:`Coq lang 0.8<coq-lang>` onwards, ``(mode vos)`` makes it
```

Alizter (approved):
LGTM I'll let @ejgallego decide when to merge.

bin/coq/coqtop.ml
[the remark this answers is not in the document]
Alizter: No nothing actionable, just a question.

rlepigre:
Actually, we have a CI failure on the test I added: apparently, the output is not consistent with what I get locally. Any clue why this happens @Alizter? I'm running Coq 8.16.0, and CI seems to be using 8.16.1: could that be the reason?
```diff
diff --git a/_build/.sandbox/d3bcb07409ea29304d4499ea0e5fe788/default/test/blackbox-tests/test-cases/coq/vos-build.t/run.t b/_build/.sandbox/d3bcb07409ea29304d4499ea0e5fe788/default/test/blackbox-tests/test-cases/coq/vos-build.t/run.t.corrected
index 1b5e703..65ef517 100644
--- a/_build/.sandbox/d3bcb07409ea29304d4499ea0e5fe788/default/test/blackbox-tests/test-cases/coq/vos-build.t/run.t
+++ b/_build/.sandbox/d3bcb07409ea29304d4499ea0e5fe788/default/test/blackbox-tests/test-cases/coq/vos-build.t/run.t.corrected
@@ -35,8 +35,8 @@ Checking that we can go back to vo mode (without cleaning).
 
   $ mv dune-vo dune
   $ dune build --display short --debug-dependency-path @all
-          coqc foo.{glob,vo}
-          coqc bar.{glob,vo}
+          coqc Nbasic_foo.{cmi,cmxs},foo.{glob,vo}
+          coqc Nbasic_bar.{cmi,cmxs},bar.{glob,vo}
   $ dune build --debug-dependency-path @default
   lib: [
     "_build/install/default/lib/base/META"
@@ -44,6 +44,10 @@ Checking that we can go back to vo mode (without cleaning).
     "_build/install/default/lib/base/opam"
   ]
   lib_root: [
+    "_build/install/default/lib/coq/user-contrib/basic/.coq-native/Nbasic_bar.cmi" {"coq/user-contrib/basic/.coq-native/Nbasic_bar.cmi"}
+    "_build/install/default/lib/coq/user-contrib/basic/.coq-native/Nbasic_bar.cmxs" {"coq/user-contrib/basic/.coq-native/Nbasic_bar.cmxs"}
+    "_build/install/default/lib/coq/user-contrib/basic/.coq-native/Nbasic_foo.cmi" {"coq/user-contrib/basic/.coq-native/Nbasic_foo.cmi"}
+    "_build/install/default/lib/coq/user-contrib/basic/.coq-native/Nbasic_foo.cmxs" {"coq/user-contrib/basic/.coq-native/Nbasic_foo.cmxs"}
     "_build/install/default/lib/coq/user-contrib/basic/bar.v" {"coq/user-contrib/basic/bar.v"}
     "_build/install/default/lib/coq/user-contrib/basic/bar.vo" {"coq/user-contrib/basic/bar.vo"}
     "_build/install/default/lib/coq/user-contrib/basic/foo.v" {"coq/user-contrib/basic/foo.v"}
```

ejgallego:
@rlepigre the difference you see is because dune's CI test with the coq-native opam package installed

test/blackbox-tests/test-cases/coq/vos-build.t/dune-vo
```diff
@@ -0,0 +1,9 @@
+(coq.theory
+ (name basic)
```
Alizter, line 2: Add (mode vo) here

Alizter: Dune will autodetect coq-native which is enabled in the test suite. You can force the mode by giving the field

rlepigre: OK thanks. Done.

ejgallego (approved):

ejgallego:
We can still improve the docs of the `(mode ...)` field but IMHO that can be done in a different PR.

PR #7407, opened 2023-03-24
adding function to emacs editor integration to get project root.

yilinwei:
Add a helper function for user scripts to get the root directory of the current dune project.

### DIFF
--- a/editor-integration/emacs/dune.el
+++ b/editor-integration/emacs/dune.el
@@ -424,6 +424,20 @@ For customization purposes, use `dune-mode-hook'."
   (compile (format "%s build @@runtest" dune-command))
   (dune-promote))
 
+(defun dune-root (&optional directory)
+  "Return the root directory of the dune project of DIRECTORY.
+
+DIRECTORY defaults to `default-directory' if not provided."
+  (let
+      ((dir (or directory default-directory)))
+    (while
+	(and dir
+	     (not
+	      (file-exists-p
+	       (file-name-concat dir "dune-project"))))
+      (setq dir (file-name-parent-directory dir)))
+    dir))
+
 (provide 'dune)
 
 ;;; dune.el ends here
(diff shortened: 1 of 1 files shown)


Alizter:
This won't give the correct project root since you can have a dune-project file in a subdirectory with respect to another. The outermost dune project file would end up being the root of the workspace. And if there is a dune-workspace outside of that, then that will be the root.

Of course, this is how I understood the code.

yilinwei:
Thanks @Alizter; I didn't realise that the `dune-project`/`dune-workspace` could be recursive. I have corrected it according to the [docs](https://dune.readthedocs.io/en/stable/usage.html#finding-the-root) and done a quick test on my system with a test project.

editor-integration/emacs/dune.el
```diff
@@ -424,6 +424,28 @@ For customization purposes, use `dune-mode-hook'."
...
+DIRECTORY defaults to `default-directory' if not provided."
+  (let
+      (root
+       workspace
+       (dir (or directory default-directory)))
+    (while dir
+      (cond
+       ((file-exists-p
+	 (file-name-concat dir "dune-workspace"))
+	(setq workspace t
```
anmonteiro, line 439: what's `t`?

anmonteiro (approved):

rgrinberg:
@Gopiandcode could you take a look at this PR?

Alizter (approved):
@ejgallego might be able to comment too.

ejgallego:
Looks OK to me, not an expert on Emacs tho, only a user :hammer_and_pick:

Gopiandcode:
Yes, looks good to me as well.

yilinwei:
@Gopiandcode 

Thanks! Should have expected that Emacs had a utility function for file traversals.

PR #7408, opened 2023-03-24
doc: allow for downloads of pdf and epub of documentation

Alizter:


emillon:
Where can the result be downloaded?

Alizter:
In the select version menu in the bottom left corner.

Alizter:
Well at least according to https://docs.readthedocs.io/en/latest/tutorial/index.html#enabling-pdf-and-epub-builds but it didn't seem to work.

emillon:
that might only works on "releases" (including `/latest`). do you have a local copy? if it does not look too awful we can enable that.

Alizter:
It doesn't show locally since that menu is javascript. I think readthedocs just generates the artifacts as you said for each release. It appears they even have download statistics.

emillon:
I mean, if you run sphinx-build in epub or pdf mode, how does it look?

emillon (approved):

emillon:
Looks good. These backends are not a priority for me but this gives a low-effort way of downloading the docs.

PR #7409, opened 2023-03-24
doc: make doc warnings fatal

Alizter:
This can be useful for catching doc builds earlier rather than in CI. Most of the time we never build locally anyway.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -6,6 +6,7 @@ build:
     python: "3.10"
 
 sphinx:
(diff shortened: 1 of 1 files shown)


emillon:
> Most of the time we never build locally anyway.

I build docs locally all the time and I like the fact that `make livedoc` can build with warnings. `make doc` will stop on errors so it's a good test if you want to checks before CI.

You changed `.readthedocs.yml` which I think is only used when deploying.

PR #7412, opened 2023-03-25
bench: fix melange benchmark (#7412)kkkkf

jchavarri:
It seems the melange benchmark regressed since https://github.com/ocaml/dune/commit/9b53ef8dabfd657eca35a48cdd3c2fa655d3257b: https://github.com/ocaml/dune/actions/runs/4514007710/jobs/7949562525

This PR brings the benchmark back.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -5,7 +5,7 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


rgrinberg:
Thanks. Should we create a ticket to create a benchmark that doesn't depend on rescript? I think rescript occludes the results here.

PR #7415, opened 2023-03-26
fix: find host ppx dependencies in the host context

### LINKED ISSUE #6191: How to install custom sites when cross-compiling
Hi,

It's not clear to me how to install packages using custom `dune-site` when cross-compiling. 

Here's a gist to reproduce: https://gist.github.com/toots/1859a37c264f3c1f75b75e21011ed17e

Normal build:
```shell
$ dune build -p repro-main --promote-install-files=false @install
$ dune install -p repro-main --create-install-files repro-main
Copying to _destdir/home/opam/.opam/4.12.0/lib/repro-main/META
Copying to _destdir/home/opam/.opam/4.12.0/lib/repro-main/Sites.ml
Copying to _destdir/home/opam/.opam/4.12.0/lib/repro-main/dune-package
Copying to _destdir/home/opam/.opam/4.12.0/lib/repro-main/repro_main.a
Copying to _destdir/home/opam/.opam/4.12.0/lib/repro-main/repro_main.cma
Copying to _destdir/home/opam/.opam/4.12.0/lib/repro-main/repro_main__Sites.cmt
Copying to _destdir/home/opam/.opam/4.12.0/lib/repro-main/repro_main.cmxs
```

Cross-compiled build:
```shell
$ dune build -p repro-main --promote-install-files=false @install -x windows
$ dune install -p repro-main --create-install-files repro-main -x windows
Error: The following <package>.install are missing:
- _build/default.windows/repro-main.install
Hint: try running: dune build [-p <pkg>] @install
```

anmonteiro:
While researching the cross-compilation story to fix #6191, I found about this failing test case that never got fixed

cc @TheLortex


Alizter:
Would this mean you could have two opam switches and use the ppx from one when compiling packages depending on the other?

anmonteiro:
I'm not too familiar with how contexts and opam switches are related, but I think there's a relation. If they're 1:1 I suspect that should be possible.

anmonteiro:
> I'm not too familiar with how contexts and opam switches are related, but I think there's a relation. If they're 1:1 I suspect that should be possible.

Reading the docs and the code a little more, it seems like this is in fact the case if you're setting up your cross compilation environment in OPAM. I think it might be even a bit more restrictive: you can _only_ depend on PPXes in the other switch.

Alizter:
That still seems OK to me. I wouldn't consider ppx a "runtime dependency" i.e. something that needs to be around when the code is executing but rather a "build time dependency". Modifying source code should be independent of platform, so whether this is performed in the target context or the host context doesn't really matter.

anmonteiro:
Note that this doesn't exclude PPXes from being cross-compiled. This change just interprets the library dependencies in `(pps ... )` in the host context.

Alizter:
@anmonteiro Could you add a note about that in the cross-compilation section? i.e. that deps from pps are interpreted in the host?

anmonteiro:
I believe this is already covered, it's just that it's broken. From https://dune.readthedocs.io/en/stable/cross-compilation.html#how-does-it-work :

> One consequence of this is that all preprocessing (PPX or otherwise) will be done using binaries built in the default context.****

src/dune_rules/lib.ml
```diff
@@ -372,6 +372,7 @@ end
 
 type db =
   { parent : db option
+  ; db_host : db Lazy.t
```
rgrinberg, line 375: Let's avoid being cute here, `db option` is enough.

Also, the field should be named `host` for consistency.

test/blackbox-tests/test-cases/ppx-cross-context-issue.t/run.t
```diff
@@ -1,12 +1,29 @@
...
-                        ^^^^^^
-  Error: Library "fooppx" in _build/cross-environment/ppx is hidden
-  (unsatisfied 'enabled_if').
-  -> required by _build/cross-environment/lib/lib.pp.ml
-  -> required by alias lib/all (context cross-environment)
-  -> required by alias default (context cross-environment)
-  [1]
+Dune uses the host context to look up dependencies and build PPXes
+
+  $ dune build --display=short
```
rgrinberg, line 3: In general, we should avoid `--display=short` in the tests. Its output changes with many arbitrary changes. I think the fact that the artifacts are built should be enough here.

anmonteiro: That's fair. I find it very useful to check the shortened build log. I'll try to remember to omit it in future changes.

src/dune_rules/lib.ml
anmonteiro: Did that in the latest push

src/dune_rules/scope.ml
```diff
@@ -273,10 +273,24 @@ module DB = struct
...
+               in
+               (project, db)
+             | Some host_context ->
+               let* scopes, public_libs = create_from_stanzas host_context in
+               let host =
+                 let scope = find_by_project scopes project in
+                 scope.db
+               in
+               let+ db =
+                 create_db_from_stanzas stanzas ~parent:public_libs
```
rgrinberg, line 290: why we are using the public_libs from the host context?

anmonteiro: looks like a bug. thanks for catching

anmonteiro: cleaned this up a bit in the most recent push

src/dune_rules/scope.ml
```diff
@@ -273,8 +273,21 @@ module DB = struct
...
+                   let+ scopes, _public_libs_host =
+                     create_from_stanzas host_context
+                   in
+                   find_by_project scopes project
+                 in
+                 Some scope.db
+             in
              let+ db =
-               create_db_from_stanzas stanzas ~parent:public_libs ~lib_config
+               create_db_from_stanzas stanzas ~parent:public_libs ~host
```
rgrinberg, line 289: Thanks for restructuring. I think we can also take advantage of laziness here and make `host` lazily evaluated. So you would wrap `host` here with `Memo.lazy` and change the type in `Lib.DB.t` to `host : Memo.Lazy.t t option`.

anmonteiro: yup, makes sense. Made it a `Memo.Lazy.t`

anmonteiro:
I tested this PR in my cross-compilation setup and everything seems to work as expected 👍

src/dune_rules/scope.ml
```diff
@@ -298,27 +310,47 @@ module DB = struct
...
+      let+ public_libs =
+        match host_context with
+        | None ->
+          let+ installed_libs = Lib.DB.installed ~host:None context in
+          public_libs t ~lib_config ~installed_libs ~host:None stanzas
+        | Some host_context ->
+          let host =
+            let host =
+              Memo.Lazy.create @@ fun () ->
+              let+ installed_libs = Lib.DB.installed ~host:None host_context in
```
rgrinberg, line 332: One day we should allow host contexts to have their host contexts as well.

rgrinberg (approved):
LGTM. Don't forget a CHANGES entry

anmonteiro:
Rebased on `main` and added a CHANGES entry

Alizter:
Just to be clear about the terminology, since this always confuses me, are we saying:
- Host platform (where we are targetting)
- Build platform (where we are building)

anmonteiro:
In my understanding, it's actually a triplet: build, host and target. This is how I think about it:

| Autotools | Dune |
---- | --- |
| Build  | Host |
| Host | --- |
| Target | Target |

TheLortex:
That's great, thank you for the fix @anmonteiro !

PR #7417, opened 2023-03-26
fix: properly report uncaught errors in threads

rgrinberg:

### DIFF
--- a/src/dune_util/report_error.ml
+++ b/src/dune_util/report_error.ml
@@ -135,7 +135,7 @@ let format_memo_stack pps =
                    (Pp.seq (Pp.verbatim "-> ")
                       (Pp.seq (Pp.text "required by ") pp))))))
 
-let report { Exn_with_backtrace.exn; backtrace } =
+let gen_report exn backtrace =
   let exn, memo_stack =
     match exn with
     | Memo.Error.E err -> (Memo.Error.get err, Memo.Error.stack err)
@@ -156,10 +156,13 @@ let report { Exn_with_backtrace.exn; backtrace } =
     let msg =
       if responsible = User && not !report_backtraces_flag then msg
       else
-        append msg
-          (List.map
-             (Printexc.raw_backtrace_to_string backtrace |> String.split_lines)
-             ~f:(fun line -> Pp.box ~indent:2 (Pp.text line)))
+        match backtrace with
+        | None -> msg
+        | Some backtrace ->
+          append msg
+            (List.map
+               (Printexc.raw_backtrace_to_string backtrace |> String.split_lines)
+               ~f:(fun line -> Pp.box ~indent:2 (Pp.text line)))
     in
     let memo_stack =
       if !print_memo_stacks || needs_stack_trace then memo_stack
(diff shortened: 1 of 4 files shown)


src/dune_util/report_error.ml
```diff
@@ -195,3 +198,8 @@ let report { Exn_with_backtrace.exn; backtrace } =
       | Developer -> append msg (i_must_not_crash ())
     in
     Console.print_user_message msg
+
+let report { Exn_with_backtrace.exn; backtrace } =
+  gen_report exn (Some backtrace)
+
+let report_exception exn = gen_report exn None
```
rgrinberg, line 205: This got me thinking that it might be better to introduce an exception wit reraising like base does:

```ocaml
exception reraise of (exn * Printexc.raw_backtrce)
```

To make exceptions and exceptions with a backtrace interchangeable

src/dune_util/report_error.ml
```diff
@@ -135,7 +135,7 @@ let format_memo_stack pps =
                    (Pp.seq (Pp.verbatim "-> ")
                       (Pp.seq (Pp.text "required by ") pp))))))
 
-let report { Exn_with_backtrace.exn; backtrace } =
+let gen_report exn backtrace =
```
snowleopard, line 138: Looking at the call sites, perhaps `?backtrace exn` would be slightly more idiomatic.

snowleopard (approved):
The change looks good but I wonder if we could add a test-case. How did the lack of proper reporting manifest?

rgrinberg:
> How did the lack of proper reporting manifest?

It would manifest by displaying the default OCaml stack trace along with the thread id. Instead of our litany against exceptions (or friendly error message against "expected" exceptions)

> The change looks good but I wonder if we could add a test-case. 

I think it's not possible without exploiting a bug in dune somewhere.

src/dune_util/report_error.ml
rgrinberg: Did you see my other suggestion? I wonder if making `Exn_with_backtrace.t` an exception itself would be good. There's a few other situations where we provide a junk backtrace just because the type required is `Exn_with_backtrace.t` rather than `Exn.t`

snowleopard: Sorry, I didn't fully understand that other suggestion. I think I'll need to see the code first. Happy to leave the type as is, if you think we're going to refactor this all anyway.

rgrinberg: It's quite similar to your suggestion. Basically, if the backtrace can be optional in many situations, we might as well change `Exn_with_backtrace.t` to:

```ocaml
type t =
  { exn : Exn.t
   ; backtrace : Printexc.backtrace option
  }
```

But then this type becomes equivalent to extending the exception type with a non optional back trace:

```ocaml
exception of Exn_with_backtrace of Exn_with_backtrace.t (* here the backtrace isn't optional *)
```

With this change, we can just `exn` everywhere and not require callers to provide fake backtraces.

Alizter:
> > The change looks good but I wonder if we could add a test-case.
> 
> I think it's not possible without exploiting a bug in dune somewhere.

Can you not make an expect test that spins up a mock Dune binary and then assert false?

src/dune_util/report_error.ml
snowleopard: I see, makes sense.

PR #6996, opened 2023-02-04
feature: add terminal ui backend based on NoTTY

Alizter:
We add a new display mode `tui` which is based on [NoTTY](https://github.com/pqwy/notty). This allows us to programmatically describe images before they are rasterized to the terminal.

It also makes reasoning about what is drawn to the terminal easier in the future, if we ever want to communicate displays over RPC.

I plan to reimplement the TUI in the future with a more sophisticated UI. For now, this subsumes the behavior of old.

PR #7418, opened 2023-03-26
fix(rpc): remove threads

### LINKED ISSUE #4244: Dune RPC - Switch from multithreading to select.
Switch csexp_rpc to use select instead of threads. Currently, csexp_rpc is creating two threads per client. This is deemed to be inefficient, and we've decided to use select instead. We already use select for the self pipe trick, so it should be natural.

rgrinberg:
replace them with evented io based on Unix.select

### DIFF
--- a/src/csexp_rpc/csexp_rpc.ml
+++ b/src/csexp_rpc/csexp_rpc.ml
@@ -1,22 +1,7 @@
 open Stdune
 open Fiber.O
+open Dune_async_io
 module Log = Dune_util.Log
-
-module type Scheduler = sig
-  val async : (unit -> 'a) -> ('a, Exn_with_backtrace.t) result Fiber.t
-end
-
-let scheduler = Fdecl.create Dyn.opaque
-
-let async f =
-  let module Scheduler = (val Fdecl.get scheduler : Scheduler) in
-  Scheduler.async f
-
-let async_exn f =
-  let+ res = async f in
-  match res with
-  | Ok s -> s
-  | Error e -> Exn_with_backtrace.reraise e
 
 module Session_id = Id.Make ()
 
@@ -97,19 +82,21 @@ end
 let debug = Option.is_some (Env.get Env.initial "DUNE_RPC_DEBUG")
 
 module Session = struct
+  let fail = function
+    | `Cancelled -> raise Dune_util.Report_error.Already_reported
+    | `Exn exn -> raise exn
+
   module Id = Session_id
 
   type state =
     | Closed
     | Open of
         { out_buf : Io_buffer.t
+        ; in_buf : Io_buffer.t
         ; fd : Unix.file_descr
-        ; (* A mutex for modifying [out_buf].
-
-             Needed as long as we use threads for async IO. Once we switch to
-             event based IO, we won't need this mutex anymore *)
-          write_mutex : Mutex.t
-        ; in_channel : in_channel
+        ; mutable read_eof : bool
+        ; write_mutex : Fiber.Mutex.t
+        ; read_mutex : Fiber.Mutex.t
         }
 
   type t =
@@ -117,39 +104,47 @@ module Session = struct
     ; mutable state : state
     }
 
-  let create fd in_channel =
+  let create fd =
+    Unix.set_nonblock fd;
     let id = Id.gen () in
     if debug then
       Log.info [ Pp.textf "RPC created new session %d" (Id.to_int id) ];
     let state =
+      let size = 8192 in
       Open
         { fd
-        ; in_channel
-        ; out_buf = Io_buffer.create ~size:8192
-        ; write_mutex = Mutex.create ()
+        ; in_buf = Io_buffer.create ~size
+        ; out_buf = Io_buffer.create ~size
+        ; read_eof = false
+        ; write_mutex = Fiber.Mutex.create ()
+        ; read_mutex = Fiber.Mutex.create ()
         }
     in
     { id; state }
 
   let string_of_packet = function
     | None -> "EOF"
-    | Some csexp -> Sexp.to_string csexp
+    | Some csexp -> Dyn.to_string (Sexp.to_dyn csexp)
 
   let string_of_packets = function
     | None -> "EOF"
     | Some sexps -> String.concat ~sep:" " (List.map ~f:Sexp.to_string sexps)
 
   let close t =
+    let* () = Fiber.return () in
     match t.state with
-    | Closed -> ()
-    | Open { write_mutex = _; fd = _; in_channel; out_buf = _ } ->
-      (* with a socket, there's only one fd. We make sure to close it only once.
-         with dune rpc init, we have two separate fd's (stdin/stdout) so we must
-         close both. *)
-      close_in_noerr in_channel;
+    | Closed -> Fiber.return ()
+    | Open { fd; _ } ->
+      let+ () = Async_io.close fd in
       t.state <- Closed
 
+  module Lexer = Csexp.Parser.Lexer
+  module Stack = Csexp.Parser.Stack
+
+  let min_read = 8192
+
   let read t =
+    let* () = Fiber.return () in
     let debug res =
       if debug then
         Log.info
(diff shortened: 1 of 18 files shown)


rgrinberg:
ping @nojb

nojb (commented):
I will try to review this, but I have some general questions:

- What is the motivation for this change? In other words, what is the main expected benefit of the changes in this PR with respect to the existing code?
- The doc mentions that pipes are used on Unix. They are not used on Windows? Is something else used instead?

rgrinberg:
There's a couple of benefits:

1. We stop leaking memory on every connected client because we no longer spawn threads we don't cleanup.
2. We no longer spawn 3 threads for every connected clients. Threads are heavy and we don't want to have too many running at once. The background threads are better saved for offloading work that can be sped up or that does background file IO.

> The doc mentions that pipes are used on Unix. They are not used on Windows? Is something else used instead?

Where does it mention that? I think we switched to sockets everywhere to simplify the implementation.

nojb:
> Where does it mention that? I think we switched to sockets everywhere to simplify the implementation.

Here:

https://github.com/ocaml/dune/pull/7418/files#diff-e846a205c1d2ea5350a77593df11a40c52bea09ea62b8cbf33200c4e4e23917aR1

and here:

https://github.com/ocaml/dune/pull/7418/files#diff-45f34c5c2485bce0c8e8b59794034f972dad8b9b34c0d3dc64333b51c2e672ecR190-R193

rgrinberg:
Oh yeah that's a single pipe I use to be able to interrupt the select. I'm using non-blocking pipes only on Unix - that's what I meant.

rgrinberg:
Some additional notes:

1. the async reading of csexp is mostly ripped off from my previous implementation in ocamllsp. it should be mostly bug free.
2. We should be careful to make sure we never close an fd that is being selected on.
3. We should make sure there no fd reuse bugs. E.g. two threads should never race to close a single fd as the the thread to lose this race might close some unrelated fd.

src/async_io/async_io.mli
```diff
@@ -0,0 +1,65 @@
...
+      locking.
+
+    - Uses the rather slow select primitive. There's much better options on
+      every operation system.
+
+    - Relies on the "pipe trick" to be interruptible. This is the best we can do
+      with select.
+
+    TODO one day switch to lev and integrate all of this directly into the
+    scheduler. to solve all the problems above *)
```
Alizter, line 16:
```suggestion
    TODO one day switch to lev and integrate all of this directly into the
    scheduler. To solve all the problems above. *)
```

src/async_io/async_io.mli
```diff
@@ -0,0 +1,65 @@
...
+  (** Cancel a running task *)
+  val cancel : _ t -> unit Fiber.t
+
+  (** Wait for a task to complete *)
+  val await : 'a t -> ('a, [ `Cancelled | `Exn of exn ]) result Fiber.t
+end
+
+(** [ready fd what ~f] wait until [what] can be done on [fd] in a non-blocking
+    way and then call [f]. Note that [f] will be called in a different thread,
+    so it should only be used for atomic or synchronized operations *)
```
Alizter, line 51:
```suggestion
(** [ready fd what ~f] wait until [what] can be done on [fd] in a non-blocking
    way and then call [f]. Note that [f] will be called in a different thread,
    so it should only be used for atomic or synchronized operations. *)
```

src/async_io/async_io.mli
```diff
@@ -0,0 +1,65 @@
...
+end
+
+(** [with_io scheduler f] runs [f] with [scheduler]. All operations in this
+    module must be executed inside [f] *)
+val with_io : (module Scheduler) -> (unit -> 'a Fiber.t) -> 'a Fiber.t
+
+(** [close fd] must be used to call any file descriptor which has been watched
+    at some point. This is needed to get make sure we never close a file
+    descriptor that is being selected. Any associated operations with [fd] will
+    be cancelled *)
```
Alizter, line 35:
```suggestion
    be cancelled. *)
```

src/async_io/async_io.mli
```diff
@@ -0,0 +1,65 @@
...
+
+  val register_job_started : unit -> unit
+
+  val cancel_job_started : unit -> unit
+
+  val spawn_thread : (unit -> unit) -> unit
+end
+
+(** [with_io scheduler f] runs [f] with [scheduler]. All operations in this
+    module must be executed inside [f] *)
```
Alizter, line 29:
```suggestion
(** [with_io scheduler f] runs [f] with [scheduler]. All operations in this
    module must be executed inside [f]. *)
```

src/async_io/async_io.ml
```diff
@@ -0,0 +1,302 @@
...
+    let module Scheduler = (val scheduler : Scheduler) in
+    Queue.to_list tasks
+    |> Fiber.parallel_iter ~f:(fun (Task t) ->
+           Scheduler.cancel_job_started ();
+           Fiber.Ivar.fill t.ivar (Error `Cancelled))
+
+let close fd =
+  let* t = Fiber.Var.get_exn t_var in
+  Mutex.lock t.mutex;
+  (* everything below is guaranteed not to raise so the mutex will be unlocked *)
```
Alizter, line 241: I don't get this comment, can you explain a little more?

src/csexp_rpc/io_buffer.mli
Alizter, line 1: Could you add some comments here?

src/csexp_rpc/csexp_rpc.ml
```diff
@@ -162,26 +148,73 @@ module Session = struct
...
-      let+ res = async read in
-      let res =
+      let+ res =
+        let* res = read Stack.Empty in
         match res with
-        | Error _ ->
-          close t;
+        | Error _exn ->
+          let+ () = close t in
           None
```
Alizter, line 213: Can you not do `>>>` for these, make them a little more compact.

src/csexp_rpc/io_buffer.mli
rgrinberg: What you like me to explain in this code?

src/csexp_rpc/io_buffer.mli
```diff
@@ -31,3 +31,15 @@ val pos : t -> int
...
+
+val commit_write : t -> len:int -> unit
+
+val max_write_len : t -> int
+
+val read_char_exn : t -> char
+
+val read_into_buffer : t -> Buffer.t -> max_len:int -> int
+
+val maybe_resize_to_fit : t -> int -> unit
```
Alizter, line 45: Just adding small comments explaining these functions is what I had in mind. So it is more uniform with the rest of the file. For example what is "commit_write"?

src/async_io/async_io.mli
```diff
@@ -0,0 +1,65 @@
+(** Poor man's asynchronous IO on sockets (and pipes on Unix)
+
+    Problematic in three ways:
+
+    - Needs to run in a separate thread because our scheduler loop does not
+      allow polling for fd's and custom events. This requires unnecessary
+      locking.
+
+    - Uses the rather slow select primitive. There's much better options on
+      every operation system.
```
gridbugs, line 10:
```suggestion
      every operating system.
```

src/async_io/async_io.mli
```diff
@@ -0,0 +1,65 @@
...
+
+  val spawn_thread : (unit -> unit) -> unit
+end
+
+(** [with_io scheduler f] runs [f] with [scheduler]. All operations in this
+    module must be executed inside [f]. *)
+val with_io : (module Scheduler) -> (unit -> 'a Fiber.t) -> 'a Fiber.t
+
+(** [close fd] must be used to call any file descriptor which has been watched
+    at some point. This is needed to get make sure we never close a file
```
gridbugs, line 33:
```suggestion
(** [close fd] must be used to close any file descriptor which has been watched
    at some point. This is needed to make sure we never close a file
```

src/async_io/async_io.ml
```diff
@@ -0,0 +1,303 @@
...
+    Unix.close t.pipe_read
+  | true ->
+    let readers, writers, ex =
+      let read = t.pipe_read :: Table.keys t.readers in
+      let write = Table.keys t.writers in
+      Mutex.unlock t.mutex;
+      (* At this point, if any [ready] acquires the lock, they need to check if
+         [read] or [write] contain their fd. If it doesn't, the write
+         [t.pipe_write] will interrupt this select *)
+      Unix.select read write [] (-1.0)
```
kit-ty-kate, line 165: According to its manual: https://www.man7.org/linux/man-pages/man2/select.2.html
```
WARNING: select() can monitor only file descriptors numbers that
       are less than FD_SETSIZE (1024)—an unreasonably low limit for
       many modern applications—and this limitation will not change.
       All modern applications should instead use [poll(2)](https://www.man7.org/linux/man-pages/man2/poll.2.html) or [epoll(7)](https://www.man7.org/linux/man-pages/man7/epoll.7.html),
       which do not suffer this limitation.
```
Is that a concern here?

rgrinberg: Not too much of a concern. The thread based implementation had similar scalability limitations.

Although eventually we'll want to graduate away from select as well.

src/async_io/async_io.ml
```diff
@@ -0,0 +1,303 @@
...
+            acc
+        in
+        (pipe, acc))
+
+let rec drain_pipe pipe buf =
+  (* a single read should be enough, but just in case *)
+  match Unix.read pipe buf 0 (Bytes.length buf) with
+  | 0 -> ()
+  | _ -> drain_pipe pipe buf
+  | exception Unix.Unix_error (Unix.EAGAIN, _, _) -> ()
```
nojb, line 122:
```suggestion
let rec drain_pipe pipe buf =
  (* a single read should be enough, but just in case *)
  match Unix.read pipe buf 0 (Bytes.length buf) with
  | 0 -> ()
  | _ -> drain_pipe pipe buf
  | exception Unix.Unix_error (Unix.EAGAIN, _, _) -> ()
```
This function will block forever if the pipe is blocking, right? You can use https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-peeknamedpipe to check if a pipe has bytes available.

rgrinberg: Good call. I removed the recursive read. It's not really needed.

rgrinberg:
@nojb could you try this again? I tried reproducing this the issue without success.

nojb:
> @nojb could you try this again? I tried reproducing this the issue without success.

Still not working:

WindowsTerminal_lvTSRzzVXQ

rgrinberg:
Thanks for your patience @nojb that video is helpful. I wasn't able to reproduce the issue because I was trying to interrupt the build before it completed. The only way I'm able to get it stuck is by waiting until the build finishes in watch mode and then hitting C-c.

By the way, I do eventually get the following error after a delay:

```
Error: exception Invalid_argument("Fswatch_win.wait: already shut down")

```

rgrinberg:
A few remarks after glancing at the code. Probably not useful, but just in case:

* I'm not sure why `wait` should fail if the queue has been shutdown. Seems like it should just return an empty list of events?

* `fsenv->thread` is being initialized in the main thread, while `wait` is being called in another thread. Why does `shutdown` wait for the main thread to the terminate? Shouldn't it be waiting for the `wait` thread?

* I would personally prefer if `Fswatch_win.shutdown` was idempotent. This is consistent with how we define such functions elsewhere in dune (when possible).

nojb:
> A few remarks after glancing at the code.

Thanks, all good points. Not sure about the second point (I have forgotten some of the details); I'll take a closer look and report back. And I'll prepare a PR with the fixes.

rgrinberg:
You can try bisecting. I can't think of anything that is Windows specific.

By the way, I discovered some issues with this PR and I've updated it to fix them. Nothing is windows specific, but perhaps windows fails in some special way. So perhaps it's worth trying this PR again.

nojb:
> You can try bisecting. I can't think of anything that is Windows specific.

The culprit seems to be #6996 cc @Alizter

rgrinberg:
Hmm, that's quite weird. I would assume tui does nothing if it's not enabled.

Alizter:
There were a few changes for tui (in a different PR) to improve how signals were handled. We'll have to see what went wrong. I have a windows vm so I can also take a closer look.

nojb:
> There were a few changes for tui (in a different PR) to improve how signals were handled. We'll have to see what went wrong. I have a windows vm so I can also take a closer look.

It seems to be related to the replacement of
```
Dune_threaded_console.progress ()
```
by
```
Dune_console.Backend.progress
```
see
https://github.com/ocaml/dune/pull/6996/files#diff-3b5679e0e3859b1dc8498f8d8b27a51fafc2265b72bf2fa4593d944ee950b634L37

rgrinberg:
> It seems to be related to the replacement of

For a quick fix, how about we go back to the unthreaded console for windows. Should be quite easy.

nojb:
> From the diff you've pointed out, switching from threaded to unthreaded introduced an issue. I have no idea why, but it's worth it to toggle it back to threaded to see if you can get the bug to go away.

I tried, but then the progress line completely disappeared, and the system did not appear very responsive... Not sure if it is an issue on Windows only or if it affects Linux as well; I'll try to investigate more later.

rgrinberg:
In any case, I think we're convinced this PR doesn't introduce this problem. Since it also fixes a serious issue with RPC, I'm merging.

PR #7419, opened 2023-03-26
test(x-compilation): `dune install -p package -x foo` build only -p packages

anmonteiro:
`dune install -p pkg-name -x cross-env` should only install `pkg-name`, not all libraries in the project


rgrinberg:
I think this count as a bug fix so let's mention it in the CHANGES

Alizter:
I think it is not a good user experience to have a `-p` flag for `dune install` that differs from `dune build`. Can we introduce a long form and deprecate this short form?

anmonteiro:
I believe that this PR addresses that concern exactly. As explained to me by @rgrinberg, `-p` is just for stanza filtering, it does nothing when installing. So this change just disables the flag for `dune install`.  Did you have something else in mind?

Alizter:
From dune build:
```
       -p PACKAGES, --for-release-of-packages=PACKAGES (required)
           Shorthand for --release --only-packages PACKAGE. You must use this
           option in your <package>.opam files, in order to build only what's
           necessary when your project contains multiple packages as well as
           getting reproducible builds.
```
from dune install:
```
       -p PACKAGES, --for-release-of-packages=PACKAGES (required)
           Shorthand for --release --only-packages PACKAGE. You must use this
           option in your <package>.opam files, in order to build only what's
           necessary when your project contains multiple packages as well as
           getting reproducible builds.
```

test/blackbox-tests/test-cases/install/disable--p.t
```diff
@@ -0,0 +1,29 @@
...
+  > (library
+  >  (name repro_other)
+  >  (public_name repro-other)
+  >  (modules bar))
+  > EOF
+
+  $ cat > foo.ml <<EOF
+  > let x = 1
+  > EOF
+  $ cp foo.ml bar.ml
```
rgrinberg, line 23: Using `cp` is problematic sometimes on macos. Creating the file twice should be fine.

rgrinberg:
@anmonteiro could you update the existing tests to remove `-p`?

CHANGES.md
```diff
@@ -1,6 +1,10 @@
 Unreleased
 ----------
 
+- Disable `-p PKG` for `dune install`. I was previously a no-op, and it now
```
Alizter, line 4:
```suggestion
- Disable `-p PKG` for `dune install`. It was previously a no-op, and it now
```

rgrinberg:
Do you think we can go for a cleaner fix? Instead of checking the command line options after parsing, we can exclude these options from `Common.term`?

emillon:
With 3.8.0 around the corner, I'd be in favour of merging as is and cleaning this up later.

bin/install_uninstall.ml
```diff
@@ -598,16 +598,29 @@ let install_uninstall ~what =
                     (Dune_engine.Context_name.to_string name)
                 ])
         in
+
```
Alizter, line 601:
```suggestion
```

emillon:
As discussed in the last meeting, we agree it needs to be fixed but a partial fix is not needed. The recommended fix is to remove `Common.term` and add the flags that are actually used.

PR #7420, opened 2023-03-27
test: spec of Path.mkdir_p w.r.t existing file / dir

Alizter:
We demonstrate that `Path.mkdir_p` silently does nothing when trying to create a directory which exists but is a file. This is an issue that stems from the underlying `Fpath` implementation).

### DIFF
--- a/otherlibs/stdune/test/io_tests.ml
+++ b/otherlibs/stdune/test/io_tests.ml
@@ -72,3 +72,14 @@ let%expect_test "copy file - dst is a directory" =
     in
     print_endline s;
     [%expect {| $DIR: Is a directory |}]
+
+let%expect_test "making a directory for an existing file" =
+  let dir = temp_dir () in
+  let fn = Path.relative dir "foo" in
+  Io.write_file fn "";
+  (* This does not error, but it will if it ends with a "/" on MacOS *)
+  ignore (Fpath.mkdir (Path.to_string fn));
+  [%expect {| |}];
+  Path.mkdir_p fn;
+  (* This in turn does not error *)
+  [%expect {| |}]


otherlibs/stdune/test/path_tests.ml
```diff
@@ -413,3 +413,13 @@ let%expect_test _ =
   [%expect {|
 None
 |}]
+
+let%expect_test "making a directory for an existing file" =
+  let fn = Path.of_string "foo" in
+  Io.write_file fn "";
+  (* This does not error, but ti will if it ends with a "/" on MacOS *)
```
Alizter, line 420:
```suggestion
  (* This does not error, but it will if it ends with a "/" on MacOS *)
```

otherlibs/stdune/test/path_tests.ml
```diff
@@ -413,3 +413,13 @@ let%expect_test _ =
   [%expect {|
 None
 |}]
+
+let%expect_test "making a directory for an existing file" =
+  let fn = Path.of_string "foo" in
```
rgrinberg, line 418: Can we do this in a temporary directory? Otherwise there' a chance it will interfere with other tests.

Alizter: Yes I will fix that.

Alizter: I've moved it to the more appropriate IO tests.

otherlibs/stdune/test/path_tests.ml
rgrinberg, line 1: What's this change about?

Alizter: Not sure how that happened.

PR #7421, opened 2023-03-27
melange: add aliases field to melange.emit

Alizter:
As discussed in #7327


jchavarri (approved):
Looks good, maybe add some test to exercise the new field?

jchavarri:
I don't plan to use it.

Alizter:
OK then I will close this PR since I demonstrated the idea. If needed it can be copied.

PR #7423, opened 2023-03-27
Add (link_flags ...) to (cinaps) stanza

### LINKED ISSUE #7416: cinaps stanza should support link_flags
## Desired Behavior
The `cinaps` stanza should support `link_flags`, just like `executable`. (Maybe cinaps should be some sort of wrapper around `executable`?)

## Example
Specifically, I want to specify `(link_flags -linkall)`, since I'm using a c library from cinaps that cannot load all its data without this flag.

The concrete example is here:
https://github.com/smuenzel/smtcaml/blob/2ca431db88310b70c6b5840228da6e77103723d8/bindings/boolector.ml#L34
This library uses cinaps to automatically generate binding code, but it needs to access data from a foreign library in order to do it, and it won't work without `-linkall` or hacky workarounds.

nojb:
Do necessary plumbing to support `(link_flags ...)` in the `(cinaps ...)` stanza. cc @smuenzel

Fixes #7416


anmonteiro (approved):

src/dune_rules/cinaps.ml
```diff
@@ -38,6 +39,7 @@ let decode =
          (Dune_lang.Syntax.since syntax (1, 1) >>> repeat Dep_conf.decode)
      and+ cinaps_version = Dune_lang.Syntax.get_exn syntax
      and+ alias = field_o "alias" Alias.Name.decode
+     and+ link_flags = Link_flags.Spec.decode ~since:(Some (3, 8))
```
rgrinberg, line 42: Note that the cinaps has its own syntax used for versioning. In this case, you should introduce `(1, 3)` to  fix the issue.

PR #7424, opened 2023-03-27
test(x-compilation): demonstrate Dune doesn't search default toolchain paths

anmonteiro:
- When cross-compiling, the default context searches `$OCAMLPATH` but it doesn't read the `findlib.conf` file if `OCAMLFIND_CONF` is present
- this works as expected for the cross-toolchain, but the default toolchain might also have configuration

### DIFF
--- a/test/blackbox-tests/test-cases/cross-compilation-ocamlfind.t
+++ b/test/blackbox-tests/test-cases/cross-compilation-ocamlfind.t
@@ -1,0 +1,79 @@
+In a cross-compilation scenario, Dune should be able to find libraries defined
+in the default findlib.conf
+
+  $ mkdir -p etc/findlib.conf.d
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/cross-compilation-ocamlfind.t
```diff
@@ -0,0 +1,77 @@
...
+  > (package (name libdep))
+  > EOF
+  $ cat > lib/dune <<EOF
+  > (library (public_name libdep))
+  > EOF
+  $ cat > lib/libdep.ml <<EOF
+  > let x = 1
+  > EOF
+
+  $ dune build --root lib
```
rgrinberg, line 23: You could build `@install` rather than `@all` here

rgrinberg:
Btw, I looked at our test suite and we have a `custom-cross-compilation` directory. We should move all the cross compilation tests over there.

PR #7425, opened 2023-03-27
`opam-file-format`: backwards-compatible update

### LINKED ISSUE #6767: Update build_path_prefix_map

Leonidas-from-XIV:
Revert of the revert in #7401. We print the same way the older `opam-file-format` used to print. It is a bit tricky to make this configurable (to e.g. print it the new way in dune 3.8) without adding some API to control this behavior to `o-f-f` and we need to support the older way of printing anyway, so this does not come with an opt-in switch to the new way of printing.

This should also closes #6767 since the vendored version of `opam-file-format` is `2.1.6` (plus our changes) which has the fix for OCaml 5 deprecations.


vendor/opam-file-format/src/dune
```diff
@@ -1,6 +1,20 @@
 (library
- (name opam_file_format)
- (wrapped false))
+  (name opam_file_format)
```
rgrinberg, line 2: Should be single space

vendor/update-opam-file-format.sh
```diff
@@ -1,20 +1,26 @@
...
+)
 
-SRC=$TMP/opam-file-format.$version
+SRC=$TMP/$name
 
-cp -v $SRC/src/*.{ml,mli,mll,mly} opam-file-format/src
+cp -v $SRC/src/*.{ml,mli,mll,mly} $name/src
+cp -v $SRC/src/dune $name/src
 
-git checkout opam-file-format/src/dune
```
rgrinberg, line 19: how come this line was removed?

vendor/opam-file-format/src/dune
```diff
@@ -1,6 +1,20 @@
...
+  (synopsis "Parser and printer for the opam file syntax")
+  (wrapped false)
+  (modules :standard \ flags)
+  (flags :standard (:include flags.sexp)))
+
+(rule
+  (with-stdout-to flags.ml
+    (echo "print_string (if String.sub Sys.ocaml_version 0 5 = \"4.02.\" then \"(-w -50)\" else \"()\")")))
```
rgrinberg, line 10: We don't support 4.02 in dune so this isn't needed

vendor/opam-file-format/src/dune
Leonidas-from-XIV: I didn't want to reformat the dune file in our [upstream fork](https://github.com/ocaml-dune/opam-file-format) for ease of rebasing onto newer releases, shall I reformat it anyway?

vendor/update-opam-file-format.sh
Leonidas-from-XIV: Now the `dune` file changes are [vendored](https://github.com/ocaml-dune/opam-file-format/commit/35d311bd38d1d83414b7095b702ae24d99885a99), I thought that is a better place to track the patch stack than in dune itself, as we have to keep a fork anyway.

rgrinberg: The changes to the dune files themselves is something we keep in the dune repo actually. In fact, we don't even track the upstream dune files and write them all ourselves. So there's no need to patch them.

Note that in this PR, we have a slightly different build in bootstrap (which doesn't use the dune files at all) and the normal build we get from `./dune build`. So it would be good to just write a minimal dune file to remove the differences.

vendor/opam-file-format/src/dune
Leonidas-from-XIV: Ok, removed that bit and the whole part about flags.

vendor/update-opam-file-format.sh
Leonidas-from-XIV: Ok, I've minimized the changes vs the dune file that we have. Wasn't very much in the end.

PR #1707, opened 2018-12-21
Prefer findlib over PATH

rgrinberg:
Attempt to fix #1701 

By default, dune basically uses OCAMLPATH and whatever is available in PATH. This seems a bit inconsistent and surprising as I think that most would expect dune to use whatever configuration findlib is using.

On the other hand, this is technically a breaking change. A possibility to avoid it is to turn this option off by default, and make it possible to toggle it in the workspace file. But this seems like it's just going to add entropy to an already complicated part of the code.

The first commit can be merged on its own as it's just fixing dune's ability to understand toolchains defined in the same config file.

Would like to hear some feedback about how accurately dune should emulate findlib (@nojb, @bobot, @emillon)

PR #7428, opened 2023-03-28
x-compilation: take host findlib.conf into account when cross-compiling

### LINKED ISSUE #1701: Dune doesn't use ocamlfind config in default context
If there's no `$OCAMLPATH` envvar set, dune build will fail because [it doesn't rely](https://github.com/ocaml/dune/blob/847bc67b2f02ab51c6ea524b34dbbfa8cc221a32/src/context.ml#L233) on `findlib.conf` in default context.
`$OCAMLPATH` is only supposed to be _an additional_ search path for package directories.

I stumbled upon this bug [while working](https://github.com/esy/esy/pull/732) on making `esy`'s build plan use findlib configs instead of environment variables. The main purpose of it, besides cleaner env, is laying out the ground work for cross-target compilation.

Unless you have other suggestions on which direction to follow, I think `dune` should respect `OCAMLFIND_CONF` env var.

Cc: @rgrinberg

anmonteiro:
- The first 3 commits are a port of https://github.com/ocaml/dune/pull/1707 to the new dune codebase. I couldn't cherry pick the commits because the branch seems gone and git can't find the SHAs anywhere
- the 4th commit fixes the bug after the setup in the first 3, and promotes the newly passing test

fixes https://github.com/ocaml/dune/issues/1701

### DIFF
--- a/src/dune_rules/context.ml
+++ b/src/dune_rules/context.ml
@@ -266,11 +266,6 @@ end = struct
     fun ~env ~root ~switch -> Memo.exec memo (env, root, switch)
 end
 
-let ocamlpath_sep =
-  if Sys.cygwin then (* because that's what ocamlfind expects *)
-    ';'
-  else Bin.path_sep
-
 module Build_environment_kind = struct
   (* Heuristics to detect the current environment *)
 
@@ -302,20 +297,36 @@ module Build_environment_kind = struct
           match opam_prefix with
           | Some s -> Opam2_environment s
           | None -> Unknown)))
-end
 
-let ocamlfind_printconf_path ~env ~ocamlfind ~toolchain =
-  let args =
-    let args = [ "printconf"; "path" ] in
-    match toolchain with
-    | None -> args
-    | Some s -> "-toolchain" :: Context_name.to_string s :: args
-  in
-  let+ l =
-    Memo.of_reproducible_fiber
-      (Process.run_capture_lines ~display:Quiet ~env Strict ocamlfind args)
-  in
-  List.map l ~f:Path.of_filename_relative_to_initial_cwd
+  let findlib_paths t ~findlib ~ocamlc_dir =
+    match (findlib, t) with
+    | ( Some findlib
+      , ( Cross_compilation_using_findlib_toolchain _
+        | Opam2_environment _
+        | Unknown ) ) -> Findlib.Config.path findlib
+    | None, Cross_compilation_using_findlib_toolchain toolchain ->
+      User_error.raise
+        [ Pp.textf
+            "Could not find `ocamlfind' in PATH or an environment variable \
+             `OCAMLFIND_CONF' while cross-compiling with toolchain `%s'"
+            (Context_name.to_string toolchain)
+        ]
+        ~hints:
+          [ Pp.enumerate
+              [ "`opam install ocamlfind' and/or:"
+              ; "Point `OCAMLFIND_CONF' to the findlib configuration that \
+                 defines this toolchain"
+              ]
+              ~f:Pp.text
+          ]
+    | _, Hardcoded_path l ->
+      List.map l ~f:Path.of_filename_relative_to_initial_cwd
+    | None, Opam2_environment opam_prefix ->
+      let p = Path.of_filename_relative_to_initial_cwd opam_prefix in
+      let p = Path.relative p "lib" in
+      [ p ]
+    | None, Unknown -> [ Path.relative (Path.parent_exn ocamlc_dir) "lib" ]
+end
 
 let check_fdo_support has_native ocfg ~name =
   let version = Ocaml.Version.of_ocaml_config ocfg in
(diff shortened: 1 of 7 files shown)


src/dune_rules/context.ml
```diff
@@ -302,20 +297,26 @@ module Build_environment_kind = struct
...
-    Memo.of_reproducible_fiber
-      (Process.run_capture_lines ~display:Quiet ~env Strict ocamlfind args)
-  in
-  List.map l ~f:Path.of_filename_relative_to_initial_cwd
+  let findlib_paths ocamlfind ~kind ~context ~findlib_toolchain ~env ~dir =
+    match (ocamlfind, query ~kind ~findlib_toolchain ~env) with
+    | ( Some ocamlfind
+      , ( Cross_compilation_using_findlib_toolchain _
+        | Opam2_environment _
+        | Unknown ) ) -> Ocamlfind.conf_path ocamlfind
```
anmonteiro, line 306: callout: we're also overriding the opam environment findlib.conf lookup if `OCAMLFIND_CONF` is set. This is consistent with the other behavior, and I've verified with @toots that `opam-cross` doesn't rely on `OCAMLFIND_CONF`.

test/blackbox-tests/test-cases/trace-file.t/run.t
```diff
@@ -3,6 +3,7 @@
 This captures the commands that are being run:
 
   $ <trace.json grep '"X"' | cut -c 2- | sed -E 's/:[0-9]+/:.../g'
+  {"args":{"process_args":["printconf","conf"],"pid":...},"ph":"X","dur":...,"name":"ocamlfind","cat":"process","ts":...,"pid":...,"tid":...}
```
rgrinberg, line 6: Are we always running `ocamlfind printconf` now?

anmonteiro: I think we're always running `ocamlfind printconf conf` now if `OCAMLFIND_CONF` is not set. This is akin to the approach in https://github.com/ocaml/dune/pull/1707

For OPAM environments, I think we know that we can find a config, if it exists, in `$OPAM_SWITCH_PREFIX/lib/findlib.conf`. If we're to prefer `findlib.conf` over `OCAMLPATH` then I believe we need to call this command.

As per the findlib reference manual, `OCAMLPATH` is meant to be an additional, prepended list of paths:

> OCAMLPATH
This variable may contain an additional search path for package directories. It is treated as if the directories were prepended to the configuration variable path.

rgrinberg: > I think we're always running ocamlfind printconf conf now if OCAMLFIND_CONF is not set. This is akin to the approach in https://github.com/ocaml/dune/pull/1707

Always? Do you mean only for cross compilation? I'm pretty sure we don't call `ocamlfind printconf` in normal builds. 1707 indeed introduces the same issue as this PR

anmonteiro: Sorry, I mean "now" as in this PR. 

I thought we wanted to read the default findlib.conf file in all cases, not just when cross-compiling.

rgrinberg: It would be good to know where it is, but running ocamlfind by default to get it doesn't seem like the right way to go. Running ocamlfind by default is something that was explicitly removed in 2.0, so I'd prefer not to bring it back.

See the comment here https://github.com/ocaml/dune/pull/1707#issuecomment-463147994 and the discussion leading to it

anmonteiro: I see. That comment is super helpful. I'll make these changes:

> [...] go for full findlib support whenever OCAMLFIND_CONF or OCAMLFIND_TOOLCHAIN are set.

anmonteiro: Alright, in my most recent push I reverted to only calling `ocamlfind printconf conf` when looking for the configuration for the target context.

anmonteiro:
I've verified that this works on my cross-compilation setup as well.

src/dune_rules/findlib/ocamlfind.ml
```diff
@@ -0,0 +1,92 @@
...
+    ; toolchain = Some toolchain
+    }
+  | Some old_toolchain ->
+    Code_error.raise "Ocamlfind.set_toolchain: cannot set toolchain twice"
+      [ ("old_toolchain", Dyn.string old_toolchain)
+      ; ("toolchain", Dyn.string toolchain)
+      ]
+
+let conf_path t =
+  match Findlib.Config.get t.config "path" with
```
rgrinberg, line 36: is this "path" config something that we've added to dune, or is it from the findlib standard?

I don't remember anymore.

src/dune_rules/context.ml
```diff
@@ -302,20 +297,26 @@ module Build_environment_kind = struct
...
+    | None, Cross_compilation_using_findlib_toolchain _toolchain ->
```
rgrinberg, line 307: Why is this case impossible? Isn't it possible to not have `ocamlfind` in `PATH`, but still set `OCAMLFIND_CONF`?

src/dune_rules/findlib/ocamlfind.ml
anmonteiro: this is `path = " ... "` or `path(TOOLCHAIN) = "..."` in findlib.conf. It's the ocaml library path (which `OCAMLPATH` appends to).

anmonteiro:
> Are the last 2 commits independent fixes?

Nope, they're fixing bugs I had introduced in my implementation. 

I'm not sure how to organize the commits here. if you want to preserve the commits you added in #1707, I'm happy to preserve them. Otherwise happy to squash everything together too.

src/dune_rules/context.ml
anmonteiro: Oh I see what you’re saying. I can make this a user error and show it in a test

anmonteiro:
This should be ready for another look.

src/dune_rules/findlib/findlib.ml
```diff
@@ -97,44 +104,164 @@ module Vars = struct
...
 
-  let get { vars; preds } var = Vars.get vars var preds
+  let path_var = Bin.parse_path ~sep:ocamlpath_sep
 
-  let env t =
-    let preds = Ps.add t.preds (P.make "env") in
-    String.Map.filter_map ~f:(Rules.interpret ~preds) t.vars
+  let ocamlpath env =
+    match Env.get env "OCAMLPATH" with
+    | None -> []
```
rgrinberg, line 188: Is it ok that we treat the empty OCAMLPATH variable the same as not setting it at all?

anmonteiro: It might be a good idea to revert this to returning an option type. I'll try that.

src/dune_rules/context.ml
```diff
@@ -355,67 +366,43 @@ type instance =
...
-        | Some s -> Memo.return s
-        | None ->
-          Memo.of_reproducible_fiber
-            (Process.run_capture_line ~display:Quiet ~env Strict fn
-               [ "printconf"; "conf" ]))
-        >>| Path.External.of_filename_relative_to_initial_cwd)
+  let env_ocamlpath = Findlib.Config.ocamlpath env in
+  let ocamlpath =
+    let initial_ocamlpath = Findlib.Config.ocamlpath Env.initial in
+    match (env_ocamlpath, initial_ocamlpath) with
```
rgrinberg, line 373: Treating the empty and unset OCAMLPATH as unset for this match looks suspicious to me tbh.

Could you add some comments regarding the selection of these rules (and perhaps preserve some of the old ones). This code is confusing enough to deserve some explnations.

anmonteiro: This is a bit cargo-culted from #1707 though I did check the logic. I'll revert it back to option types which shouldn't change the previous code, and will make it clearer.

anmonteiro: I made this a lot clearer in [302b45c](https://github.com/ocaml/dune/pull/7428/commits/302b45c68596c7d83bca09df4db27b5ca8fff8c7). Added comments explaining the different branches too.

src/dune_rules/findlib/findlib.mli
```diff
@@ -51,10 +51,25 @@ module Config : sig
...
+  (** Finds program [prog] for this configuration, if it exists *)
+  val tool : t -> prog:string -> Path.t option Memo.t
 
-  val env : t -> Env.t
+  val ocamlpath_sep : char
+
+  (** Read and parse the [OCAMLPATH] environment variable *)
+  val ocamlpath : Env.t -> Path.t list option
+
+  val extra_env : t -> Env.t
```
rgrinberg, line 66: This function could use a doc string

src/dune_rules/context.ml
```diff
@@ -302,20 +297,36 @@ module Build_environment_kind = struct
...
-    match toolchain with
-    | None -> args
-    | Some s -> "-toolchain" :: Context_name.to_string s :: args
-  in
-  let+ l =
-    Memo.of_reproducible_fiber
-      (Process.run_capture_lines ~display:Quiet ~env Strict ocamlfind args)
-  in
-  List.map l ~f:Path.of_filename_relative_to_initial_cwd
+  let findlib_paths findlib ~kind ~findlib_toolchain ~env ~dir =
```
rgrinberg, line 301: The name `~dir` is a bit opaque. Can we pick something better? Perhaps `ocamlc_dir`?

src/dune_rules/context.ml
```diff
@@ -302,20 +297,36 @@ module Build_environment_kind = struct
...
+    match (findlib, query ~kind ~findlib_toolchain ~env) with
```
rgrinberg, line 302: This is a bit of a matter of taste, but I'm not a function that takes a bunch of arguments only to apply them to another function. Why not pass the result of the function directly then? E.g.

```
let findlib_paths t findlib ~dir =
  match (findlib, t) with ...
```

src/dune_rules/context.ml
```diff
@@ -355,67 +366,53 @@ type instance =
...
-        let* path = Memo.Lazy.force findlib_config_path in
-        let toolchain = Context_name.to_string toolchain in
-        let context = Context_name.to_string name in
-        let+ config = Findlib.Config.load (External path) ~toolchain ~context in
-        Some config
+    let ocamlpath =
+      match (kind, findlib_toolchain) with
+      | Default, None -> Option.value env_ocamlpath ~default:[]
+      | _, _ -> (
+        let initial_ocamlpath = Findlib.Config.ocamlpath Env.initial in
```
rgrinberg, line 377: This value can be computed only once.

src/dune_rules/findlib/findlib.mli
```diff
@@ -51,10 +51,25 @@ module Config : sig
...
+
+  (** Load the findlib configuration for this [findlib_toolchain], if any. *)
+  val discover_from_env :
+       env:Env.t
+    -> which:(string -> Path.t option Memo.t)
+    -> ocamlpath:Path.t list
+    -> findlib_toolchain:Dune_engine.Context_name.t option
```
rgrinberg, line 73: is this really a requirement that we currently have? that a tool chain name should be a valid context name?

src/dune_rules/findlib/findlib.mli
```diff
@@ -51,10 +51,25 @@ module Config : sig
...
+
+  (** Read and parse the [OCAMLPATH] environment variable *)
+  val ocamlpath : Env.t -> Path.t list option
+
+  val extra_env : t -> Env.t
+
+  (** Load the findlib configuration for this [findlib_toolchain], if any. *)
+  val discover_from_env :
+       env:Env.t
+    -> which:(string -> Path.t option Memo.t)
```
rgrinberg, line 71: I suppose this should be `Filename.t -> Path.t option Memo.t`

src/dune_rules/findlib/findlib.ml
```diff
@@ -97,44 +104,161 @@ module Vars = struct
...
+        match findlib_toolchain with
+        | None -> Memo.return None
+        | Some _ -> (
+          which "ocamlfind" >>= function
+          | None -> Memo.return None
+          | Some fn ->
+            Memo.of_reproducible_fiber
+              (Process.run_capture_line ~display:Quiet ~env Strict fn
+                 [ "printconf"; "conf" ])
+            |> Memo.map ~f:Option.some))
```
rgrinberg, line 238: Personally, I find it weird to see code like `g (f x) |> h`. It might as well be `f x |> g |> h`

src/dune_rules/findlib/findlib.ml
```diff
@@ -97,44 +104,161 @@ module Vars = struct
...
+
+  let ocamlfind_config_path ~env ~which ~findlib_toolchain =
+    let open Memo.O in
+    let+ path =
+      match Env.get env "OCAMLFIND_CONF" with
+      | Some s -> Memo.return (Some s)
+      | None -> (
+        match findlib_toolchain with
+        | None -> Memo.return None
+        | Some _ -> (
```
rgrinberg, line 231: Could you add a comment here? Some might find the behavior we chose to be confusing.

src/dune_rules/findlib/findlib.ml
```diff
@@ -97,44 +104,161 @@ module Vars = struct
...
+            let+ all_vars =
+              Memo.parallel_map (Fs_cache.Dir_contents.to_list dir_contents)
+                ~f:(fun (p, _kind) ->
+                  let p = Path.Outside_build_dir.relative config_dir p in
+                  load p)
+            in
+            List.fold_left all_vars ~init:vars ~f:(fun acc vars ->
+                Vars.union acc vars ~f:(fun _ x y -> Some (Rules.union x y)))
+          | Error _ -> Memo.return vars)
+        | Ok false | Error _ -> Memo.return vars
```
rgrinberg, line 154: Our style is to keep the shorter clauses first.

src/dune_rules/findlib/findlib.mli
anmonteiro: I'm not sure what our limitations are. I changed it to a string, which should allow more names.

rgrinberg (approved):
The tests need to be updated, but otherwise this is good to go.

PR #7429, opened 2023-03-28
chore(nix): add bare opam shell

anmonteiro:
This was useful for me to debug an opam failure in https://github.com/ocaml/dune/pull/7428.

It's, however, fairly easy to write it again if I need it, so I'm not sure whether it belongs in flake.nix or not.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -159,6 +159,16 @@
               pkgs.ocamlPackages.mel
             ];
           };
+          slim-opam = with pkgs; mkShell {
+            nativeBuildInputs = lib.remove pkgs.ocamlformat testNativeBuildInputs;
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -159,6 +159,12 @@
               pkgs.ocamlPackages.mel
             ];
           };
+          slim-opam = with pkgs; mkShell {
```
rgrinberg, line 162: Could you leave a comment with a couple of sentences what this is for. We should probably do this for all the shells we define, but we can start here.

anmonteiro: Good idea. Added a description. Will send a separate PR for the others.

rgrinberg (approved):

Alizter (approved):

PR #7431, opened 2023-03-28
[Ctypes] fix compilation in byte-only

bobot:
When the native compilation is not present, the compilation of the generators for ctypes fails since it only use native compilation. For example https://ocaml.ci.dev/github/bobot/ocaml-flint/commit/24ee67cb3939b936777d5f9899d558de3773816d/variant/debian-11-5.0_s390x_opam-2.1#L313-313

This fix is better than the current code, but perhaps we should use one of the compilation modes selected for the library.

CC @kit-ty-kate


rgrinberg:
Could you add a test somehow? I think we we have an example of hiding ocamlopt to simulate a bytecode only switch.

PR #7442, opened 2023-03-29
refactor: move /dev/null related stuff

rgrinberg:
Now it lives in [Stdune.Dev_null] rather than [Dune_util.Config]


snowleopard (approved):
Looks good!

PR #7443, opened 2023-03-29
Add command to wait until nth watch-mode build completes

gridbugs:
This is an experiment to make it easier to benchmark dune in watch mode. This adds a command `dune internal watch-wait N` which when run alongside `dune build --watch`, blocks until the Nth (re)build has completed by dune in watch mode (since dune was started - not since `dune internal watch-wait N` was run).

My plan for benchmarking dune in watch mode is to make a benchmark runner program that runs `dune build --watch` in the background, and repeatedly changes files to trigger rebuilds. Dune will write out benchmarking results through the existing `--trace-file` mechanism. The benchmark runner needs a way of waiting until a build is complete before making additional changes to files, otherwise the in-progress build could be interrupted. The `dune internal watch-wait` command is intended to solve this problem.

### DIFF
--- a/bin/internal_watch_wait.ml
+++ b/bin/internal_watch_wait.ml
@@ -1,0 +1,50 @@
+open Import
+module Client = Dune_rpc_client.Client
+
+let doc =
+  "Run alongside `build --watch`, this command will exit after the nth rebuild \
+   has completed"
+
+let man =
+  [ `S "DESCRIPTION"
+  ; `P
+      {|Run this while dune is running in watch mode. This command takes an integer argument n and blocks until the nth rebuild has completed since dune was started. |}
+  ; `Blocks Common.help_secs
+  ]
+
+let info = Cmd.info "watch-wait" ~doc ~man
+
+let witness = Dune_rpc_private.Decl.Request.witness
+
+let request_exn client witness n =
+  let open Fiber.O in
+  let* decl = Client.Versioned.prepare_request client witness in
+  match decl with
+  | Error e -> raise (Dune_rpc_private.Version_error.E e)
+  | Ok decl -> Client.request client decl n
+
+let run ~seqno =
+  let open Fiber.O in
+  let where = Rpc.active_server () in
+  let* conn = Client.Connection.connect_exn where in
+  Client.client conn
+    ~private_menu:[ Request Dune_rpc_impl.Decl.watch_mode_event_long_poll ]
+    (Dune_rpc.Initialize.Request.create
+       ~id:(Dune_rpc.Id.make (Csexp.Atom "status")))
+    ~f:(fun client ->
+      let* _response =
+        request_exn client
+          (witness Dune_rpc_impl.Decl.watch_mode_event_long_poll)
+          seqno
+        >>| Result.get_ok
+      in
+      Fiber.return ())
+
+let term =
+  let+ (common : Common.t) = Common.term
+  and+ seqno =
+    Arg.(required & pos 0 (some int) None & Arg.info [] ~docv:"INT")
+  in
+  Rpc.client_term common @@ fun _common -> run ~seqno
+
+let command = Cmd.v info term
--- a/src/dune_util/event_queue.ml
+++ b/src/dune_util/event_queue.ml
@@ -1,0 +1,108 @@
+open Stdune
+
+module Map_queue = struct
+  (** A queue storing a maximum number of elements which are automatically
+      dropped as new elements are added. It provides random access via
+      monotonically incrementing integer indices. *)
+  type 'a t =
+    { map : 'a Int.Map.t
+    ; max_size : int
+    }
+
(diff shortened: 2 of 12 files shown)


src/dune_util/event_queue.mli
```diff
@@ -0,0 +1,28 @@
+(** Stores an association between monotonically increasing integer sequence
+    numbers and events. There is a limit to the number of events that can be
+    stored, and once this limit is reached, adding a new event will cause the
+    oldest event to be removed. *)
```
rgrinberg, line 4: Sounds like a ring buffer?

src/dune_util/event_queue.mli
```diff
@@ -0,0 +1,28 @@
...
+type 'a t
+
+(** Create a new event queue *)
```
rgrinberg, line 7: I'd move this module `dune_rpc_impl`. It seems like an implementation detail of rpc, so I don't see a need to make it available to anyone else

bin/internal_watch_wait.ml
```diff
@@ -0,0 +1,50 @@
...
+let run ~seqno =
+  let open Fiber.O in
+  let where = Rpc.active_server () in
+  let* conn = Client.Connection.connect_exn where in
+  Client.client conn
+    ~private_menu:[ Request Dune_rpc_impl.Decl.watch_mode_event_long_poll ]
+    (Dune_rpc.Initialize.Request.create
+       ~id:(Dune_rpc.Id.make (Csexp.Atom "status")))
+    ~f:(fun client ->
+      let* _response =
```
rgrinberg, line 35: What's the time we're ignoring here?

src/dune_util/event_queue.mli
```diff
@@ -0,0 +1,28 @@
...
+val create : max_stored_events:int -> 'a t
+
+(** Returns the sequence number that will be assigned to the next event added to
+    the queue *)
+val next_seqno : 'a t -> int
```
rgrinberg, line 12: Seems like this function can be private

src/dune_util/event_queue.ml
```diff
@@ -0,0 +1,108 @@
...
+  let next_seqno t = Map_queue.next_index t.events
+
+  let add_event t ~event =
+    let open Fiber.O in
+    let events, seqno = Map_queue.insert t.events event in
+    let+ () =
+      match Int.Map.find t.callbacks seqno with
+      | None -> Fiber.return ()
+      | Some callbacks ->
+        List.map callbacks ~f:(fun f -> f event) |> Fiber.all_concurrently_unit
```
rgrinberg, line 62: Can be `Fiber.parallel_iter` here

src/dune_util/event_queue.ml
```diff
@@ -0,0 +1,108 @@
...
+    in
+    let callbacks = Int.Map.remove t.callbacks seqno in
```
rgrinberg, line 64: Is it possible that `seqno` might be get invalidated after we wait for all the callbacks?

rgrinberg:
> Why not use Dune RPC directly to figure out when a build has finished?

As far as I can tell, all of the functionality in this PR is implemented via RPC. The commands are just wrappers around that.

gridbugs:
> Why not use Dune RPC directly to figure out when a build has finished?

I assume you're asking why I introduced a new subcommand to dune which runs an RPC client as opposed to just allowing the benchmark runner to be the RPC client. I think eventually it will make sense to expose this RPC interface publicly but I wanted to keep it private for now until I'm sure about its semantics. (Dune distinguishes between public and private RPC commands. "build" is an example of another private RPC command (@rgrinberg please correct me if I'm wrong about this!)). The subcommand is a minimal way for me to test the RPC interface and I figured it could be useful on its own when writing tests/benchmarks in a language other than ocaml (e.g. cram tests).

gridbugs:
> Waiting for Nth build to finish seems like an odd piece of functionality to me.

This seemed conceptually simpler to me (as a user) than waiting for the current build to finish, but I think you're right in that the only way this will be used in practice will be to wait for the current build to finish. I'll change the RPC to just wait for the current build to finish. That will also simplify the implementation a bunch.

gridbugs:
> Also, it seems odd to put this functionality in Dune: this is benchmarking specific, and I'd expect to see the corresponding code in the benchmarking suite rather than adding more bloat to Dune itself.

Which functionality? Are you just referring to the new subcommand? I'm not convinced that this is benchmarknig-specific. I've written some pretty messy cram tests of watch-mode features in the past that would have been much simpler with a feature like this.

gridbugs:
Closing this and opening a new PR which changes the behaviour of the RPC/command to just wait until the end of the next build in watch-mode. This removes the need for the `Event_queue` ring buffer module which simplifies the code. New PR is here: https://github.com/ocaml/dune/pull/7473

PR #7445, opened 2023-03-29
fix: Allow `package` in any location in `rule`

Leonidas-from-XIV:
While fixing the package associations in https://github.com/ocaml/opam/pull/5496 I realized that the `package` stanza is only accepted when it is at the end of the `rule` stanza.

Thanks for @emillon for pointing out the fix, his prediction that this is a one-line change was spot on.


test/blackbox-tests/test-cases/package-rule.t/run.t
```diff
@@ -0,0 +1,5 @@
+When --only-packages is passed, it runs
+
+  $ dune build --only-packages a @runtest
+  A
+  A
```
Alizter, line 5: Can you make sure that this test actually succeeds in the commit where you are adding it. Otherwise it is difficult to see how the change changes things.

Leonidas-from-XIV: Ah, I see your point. Made the test succeed with the failing state first before fixing it in the subsequent commit.

Alizter (approved):

emillon:
Thanks

PR #7446, opened 2023-03-29
Fix cinaps (link_flags) support version check

nojb:
A mistake in #7423 meant that the new feature was versioned in terms of the version of the Dune language, instead of the version of the `cinaps` extension. This PR fixes that.


rgrinberg (approved):

anmonteiro (approved):
Nice

PR #7450, opened 2023-03-30
test(x-compilation): demonstrate overlap check failure with ppx_runtime_libraries

anmonteiro:
- There's a lingering issue after #7415 where a library's runtime dependencies get resolved at the time of its instantiation
- In a cross-compilation scenario, we only know the context when we're resolving the library that depends on the ppx with runtime_deps, so they can't be resolved eagerly when the ppx library is being instantiated.
- This patch proposes deferring their evaluation until they're needed by the rules post-preprocessing.

### DIFF
--- a/src/dune_rules/lib.ml
+++ b/src/dune_rules/lib.ml
@@ -299,7 +299,7 @@ module T = struct
     ; (* [requires] is contains all required libraries, including the ones
          mentioned in [re_exports]. *)
       requires : t list Resolve.t
-    ; ppx_runtime_deps : t list Resolve.t
+    ; ppx_runtime_deps_host : t list Resolve.t Memo.Lazy.t
     ; pps : t list Resolve.t
     ; resolved_selects : Resolved_select.t list Resolve.t
     ; implements : t Resolve.t option
@@ -353,6 +353,30 @@ module Hidden = struct
     { lib = info; reason = "unsatisfied 'exist_if'"; path }
 end
 
+module Private_deps = struct
+  type t =
+    | From_same_project of [ `Public | `Private_package ]
+    | Allow_all
+
+  let equal a b =
+    match (a, b) with
+    | Allow_all, Allow_all
+    | From_same_project `Public, From_same_project `Public
+    | From_same_project `Private_package, From_same_project `Private_package ->
+      true
+    | _ -> false
+
+  let check t ~loc ~lib =
+    match t with
+    | Allow_all -> Ok lib
+    | From_same_project kind -> (
+      match Lib_info.status lib.info with
+      | Private (_, Some _) -> Ok lib
+      | Private (_, None) ->
+        Error (Error.private_deps_not_allowed ~kind ~loc lib.info)
+      | _ -> Ok lib)
+end
+
 module Status = struct
   type t =
     | Found of lib
@@ -374,6 +398,8 @@ type db =
   { parent : db option
   ; host : db Memo.Lazy.t option
   ; resolve : Lib_name.t -> resolve_result Memo.t
+  ; resolve_ppx_runtime_deps :
+      (Path.t Lib_info.t * Private_deps.t, t list Resolve.t) Memo.Table.t
   ; all : Lib_name.t list Memo.Lazy.t
   ; lib_config : Lib_config.t
   }
@@ -399,7 +425,7 @@ let implements t = Option.map ~f:Memo.return t.implements
 
 let requires t = Memo.return t.requires
 
-let ppx_runtime_deps t = Memo.return t.ppx_runtime_deps
+let ppx_runtime_deps t = Memo.Lazy.force t.ppx_runtime_deps_host
 
 let pps t = Memo.return t.pps
 
@@ -603,20 +629,6 @@ end = struct
         { stack = x :: t.stack; seen = Id.Set.add t.seen x; implements_via }
 end
 
-type private_deps =
-  | From_same_project of [ `Public | `Private_package ]
-  | Allow_all
-
-let check_private_deps lib ~loc ~(private_deps : private_deps) =
-  match private_deps with
-  | Allow_all -> Ok lib
-  | From_same_project kind -> (
-    match Lib_info.status lib.info with
-    | Private (_, Some _) -> Ok lib
-    | Private (_, None) ->
-      Error (Error.private_deps_not_allowed ~kind ~loc lib.info)
-    | _ -> Ok lib)
-
 module Vlib : sig
   (** Make sure that for every virtual library in the list there is at most one
       corresponding implementation.
@@ -776,7 +788,10 @@ module rec Resolve_names : sig
   val find_internal : db -> Lib_name.t -> Status.t Memo.t
 
   val resolve_dep :
-    db -> Loc.t * Lib_name.t -> private_deps:private_deps -> lib Resolve.Memo.t
+       db
+    -> Loc.t * Lib_name.t
+    -> private_deps:Private_deps.t
+    -> lib Resolve.Memo.t
 
   val resolve_name : db -> Lib_name.t -> Status.t Memo.t
 
@@ -785,7 +800,7 @@ module rec Resolve_names : sig
   val resolve_simple_deps :
        db
     -> (Loc.t * Lib_name.t) list
-    -> private_deps:private_deps
+    -> private_deps:Private_deps.t
     -> t list Resolve.Memo.t
 
   type resolved =
@@ -798,7 +813,7 @@ module rec Resolve_names : sig
   val resolve_deps_and_add_runtime_deps :
        db
     -> Lib_dep.t list
-    -> private_deps:private_deps
+    -> private_deps:Private_deps.t
     -> pps:(Loc.t * Lib_name.t) list
     -> dune_version:Dune_lang.Syntax.Version.t option
     -> resolved Memo.t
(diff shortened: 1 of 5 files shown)


src/dune_rules/lib.ml
```diff
@@ -1902,7 +1927,7 @@ let to_dune_lib ({ info; _ } as lib) ~modules ~foreign_objects
       Lib_name.mangled (Package.name pkg) (Lib_name.to_local_exn lib.name)
     | _ -> lib.name
   in
-  let add_loc = List.map ~f:(fun x -> (loc, mangled_name x)) in
+  (* let add_loc = List.map ~f:(fun x -> (loc, mangled_name x)) in *)
```
anmonteiro, line 1930: Still need to add this in the new, correct place.

anmonteiro:
This ends up fixing https://github.com/ocaml/dune/pull/2794 too.

src/dune_rules/lib.mli
```diff
@@ -57,6 +55,8 @@ type sub_system = ..
 
 type db
 
+val ppx_runtime_deps : t -> db -> t list Resolve.Memo.t
```
rgrinberg, line 58: The new API seems wrong to me. Once a library is resolved, it belongs to a particular database and all of its dependencies have been resolved. You should be able to delay the evaluate of ppx_runtime_deps without changing the API.

src/dune_rules/lib.ml
```diff
@@ -1081,6 +1093,16 @@ end = struct
   let resolve_simple_deps db names ~private_deps =
     Resolve.Memo.List.map names ~f:(resolve_dep db ~private_deps)
 
+  let resolve_ppx_runtime_deps db names ~private_deps ~for_lib =
```
rgrinberg, line 1096: Instead of passing `for_lib` everywhere, how about just adding the stack frame before the call that resolves the runtime deps?

src/dune_rules/lib.ml
```diff
@@ -1305,18 +1312,37 @@ end = struct
...
+            let* ppx_runtime_deps =
+              let open Memo.O in
+              match db.host with
+              | None -> ppx_runtime_deps pp
+              | Some _host ->
+                (* (ppx_runtime_libraries ...) run in the target context, so
+                   these dependencies need to be resolved here rather than at
+                   instantiation of the ppx library (in the host context). *)
+                let+ resolved =
+                  resolve_simple_deps db
```
anmonteiro, line 1325: this can still probably be memoized?

anmonteiro:
In my latest push:

- preserved the same behavior for the non-cross-compilation code path
- added comments to the relevant sections for this patch and #7415, explaining why we're doing something different if `db.host` is `Some _`

src/dune_rules/lib.ml
anmonteiro: did this too.

src/dune_rules/lib.ml
```diff
@@ -950,16 +970,14 @@ end = struct
...
-    in
     let src_dir = Lib_info.src_dir info in
-    let map_error x =
-      Resolve.push_stack_frame x ~human_readable_description:(fun () ->
-          Dep_path.Entry.Lib.pp { name; path = src_dir })
+    let ppx_runtime_deps_host =
+      Memo.lazy_ (fun () ->
+          Lib_info.ppx_runtime_deps info
+          |> resolve_simple_deps db ~private_deps
+          >>| add_stack_frame name ~path:src_dir)
```
rgrinberg, line 978: Can't you use the memoized function you've introduced to get the ppx runtime deps for the host?

rgrinberg: It's also a little weird that we first resolve the deps and then add the stack frame. Shouldn't it be the other way?

src/dune_rules/lib.ml
```diff
@@ -1305,18 +1351,34 @@ end = struct
...
+        Resolve.Memo.List.concat_map pps ~f:(fun pp ->
+            let* ppx_runtime_deps =
+              let open Memo.O in
+              match db.host with
+              | None -> ppx_runtime_deps pp
+              | Some _host ->
+                (* (ppx_runtime_libraries ...) run in the target context, so
+                   these dependencies need to be resolved here rather than at
+                   instantiation of the ppx library (in the host context). *)
+                let+ resolved = resolve_ppx_runtime_deps db pp ~private_deps in
```
rgrinberg, line 1363: I would expect the stack frame to be added before the call to `resolve_ppx_runtime_deps`

src/dune_rules/lib.ml
```diff
@@ -834,17 +852,18 @@ end = struct
...
          have to check it again. It has been checked when compiling the
          libraries before their installation *)
-      | Installed_private | Private (_, None) | Installed -> Allow_all
+      | Installed_private | Private (_, None) | Installed ->
+        Private_deps.Allow_all
       | Private (_, Some _) -> From_same_project `Private_package
       | Public (_, _) -> From_same_project `Public
     in
     let resolve name = resolve_dep db name ~private_deps in
+    let for_lib = Some info in
```
rgrinberg, line 866: I find it a bit weird that we need this parameter at all. I would expect us to push the stackframe that corresponds to it and then resume the work as normal

rgrinberg:
Please add a CHANGES entry

src/dune_rules/lib.ml
anmonteiro: OK. I might be misunderstanding how it all works, but they seem to be added after resolution here: https://github.com/ocaml/dune/blob/bfbc453f5d47fcf2b50ec3260e0252900651cfba/src/dune_rules/lib.ml#L957-L962

I tried to preserve the same behavior.

src/dune_rules/lib.ml
anmonteiro: > Can't you use the memoized function you've introduced to get the ppx runtime deps for the host?

OH I can probably do that, indeed. Let me tyr

src/dune_rules/lib.ml
```diff
@@ -1259,6 +1277,30 @@ end = struct
     ; runtime_deps : t list Resolve.Memo.t
     }
 
+  let resolve_ppx_runtime_deps =
+    let resolve_ppx_runtime_deps_impl (db, info, private_deps) =
+      resolve_simple_deps db (Lib_info.ppx_runtime_deps info) ~private_deps
+    in
+    let module Input = struct
+      type t = db * Path.t Lib_info.t * Private_deps.t
```
rgrinberg, line 1285: Any change to the database is going to invalidate all the inputs with this database because of the physical equality, so how about we just move move this memo to the db itself:

```
; resolve_ppx_runtime_deps : ((Path.t Lib_info.t * Private_deps.t), _) Memo.t
```

This way the entire memo table can be gc'd if the db is gc'd.

rgrinberg (approved):
LGTM. Some minor comment regarding the lifetimes of memoized values. You can try implementing it, but it's not required to merge this PR.

src/dune_rules/lib.ml
anmonteiro: what's the `_` in your example?

anmonteiro: I'm a bit unsure how to store the memo for the ppx runtime libraries in the DB  (which DB?!). My understanding is that we discover these `ppx_runtime_deps` when instantiating the host library (because it's a PPX). so I'm a bit unsure how to resolve them during instantiation since we don't have the DB to resolve them.

I think I'm missing something in your suggestion.

rgrinberg: > what's the _ in your example?

It's the output of the memoized function. In this case, the list of resolved libraries.

>  My understanding is that we discover these ppx_runtime_deps when instantiating the host library (because it's a PPX). so I'm a bit unsure how to resolve them during instantiation since we don't have the DB to resolve them.

It's mostly a syntactic transformation. Everywhere where we have:

```
Memo.exec (db, info, private_deps)
```

We turn it into:

```
db.resolve_ppx_runtime_deps (info, private_deps
```

I imagine this should be possible

rgrinberg: PS: I realize I made a typo in my original comment that made it confusing. It should be `Memo.Table.t` rather than `Memo.t`

anmonteiro: aha, ok, I'll play with this.

anmonteiro: This is nicer, indeed. Moved the `resolve_ppx_runtime_deps` function to the `db` type.

anmonteiro: @rgrinberg I'm waiting for a final look after your suggestion before I merge this, to be sure.

src/dune_rules/lib.ml
```diff
@@ -1733,8 +1765,42 @@ module DB = struct
...
+          ~private_deps
+      in
+      Memo.create "lib-resolve-ppx-runtime-libraries"
+        ~input:(module Input)
+        resolve_ppx_runtime_deps_impl
+        ~human_readable_description:(fun (info, _private_deps) ->
+          Dep_path.Entry.Lib.pp
+            { name = Lib_info.name info; path = Lib_info.src_dir info })
+    in
+    fun ~parent ~host ~resolve ~all ~lib_config () ->
```
rgrinberg, line 1792: unrelated: this `unit` argument isn't needed. Maybe leave a TODO?

rgrinberg (approved):
LGTM

Issue #7454, opened 2023-03-30
`dune fmt` unexpectedly triggers menhir rules


voodoos:
## Expected Behavior
I would expect `dune fmt` or `dune build @fmt` to never build anything.
(This is an assumption made by OCaml-CI for the `lint-fmt` job. cc @maiste)

## Actual Behavior
Menhir parser that have a corresponding `.mli` file are generated when calling `dune build @fmt`

## Reproduction

It was a bit subtle to reproduce and minimize the issue we encountered in Merlin.
I made a PR with what I think is a fairly minimal reproduction: https://github.com/ocaml/dune/pull/7455

## Specifications

- Version of `dune`: reproducing PR is based on main
- Version of `ocaml`: unrelevant
- Operating system: unrelevant

voodoos:
I investigated a bit and I think the issue lies not in the formatting rules but in the menhir rules.

The formatting rules do only work for files in the source tree:
> <img width="530" alt="image" src="https://user-images.githubusercontent.com/5031221/230095918-353ae24f-f493-4588-b8aa-3ba6bef054a5.png">
https://github.com/ocaml/dune/blob/main/src/dune_rules/format_rules.ml#L107-L108

But in the test, if I show the rules for the mli file that is in the source-tree:
<img width="657" alt="image" src="https://user-images.githubusercontent.com/5031221/230096314-71fc65d3-d3f7-4bad-a3ff-d47b90f8e1fa.png">
It says that the file is generated by Menhir, not that is is simply in the source tree. I think this is why dune runs menhir when trying to format the mli files: it is in the source tree so it is formatted, but a rule says that is generated thus it is generated along with the ml file.

The target configuration in menhir:
> <img width="398" alt="image" src="https://user-images.githubusercontent.com/5031221/230097266-ed3becf9-12de-4534-b329-3e17c9940d25.png">
https://github.com/ocaml/dune/blob/main/src/dune_rules/menhir/menhir_rules.ml#L77-L80

If I remove the mli from the targets the test pass and the rules for the mli files are correct:
> <img width="544" alt="image" src="https://user-images.githubusercontent.com/5031221/230097464-410f7c08-74fb-4df7-a807-45ab21fa6efd.png">

Of course this is not an acceptable patch, we should conditionally have that mli file as target if it is not already in the source tree. Any ideas how I could do that ?

cc @rgrinberg @anmonteiro @emillon who intervened in the reproduction PR

rgrinberg:
[...]

PR #6585, opened 2022-11-26
fix: dune files from source tree

rgrinberg:
The dune file path returned by Source_tree.Dune_file is always present.
This is wrong because some dune files are generated using subdir. This
commit tweaks the API to reflect this fact and updates the rest of the
code to handle such virtual dune files.

Signed-off-by: Rudi Grinberg <me@rgrinberg.com>

ps-id: b2dfc4df-ea48-491c-987e-b9b3d193e597

PR #7455, opened 2023-03-30
Test for issue #7454: fmt triggers menhir

voodoos:
See issue #7454.

## Expected Behavior
I would expect `dune fmt` or `dune build @fmt` to never build anything.
(This is an assumption made by OCaml-CI for the `lint-fmt` job. cc @maiste)

## Actual Behavior
Missing menhir parsers that have a corresponding `.mli` file are generated when calling `dune build @fmt`


test/blackbox-tests/test-cases/formatting/no-gen.t
```diff
@@ -0,0 +1,78 @@
+See issue 7454
```
rgrinberg, line 1: Could you paste the relevant details from the issue here?

rgrinberg (approved):
I guess the issue is that we're setting up formatting rules for all sources used for building. We should only be formatting sources that are present in the sources.

anmonteiro:
I wonder if https://github.com/ocaml/dune/pull/6585 is related to this.

emillon (approved):
Sounds good to me too. Please add DCO and I'll rebase+merge.

voodoos:
> Sounds good to me too. Please add DCO and I'll rebase+merge.

Done !

rgrinberg:
None of the uses in this PR require the `-a` flag either. So using `ls` without flags should be fine too

PR #7458, opened 2023-03-31
refactor: delete unused opens in lib.ml

anmonteiro:
While working on #7450 I noticed a few unused `open` expressions. `Resolve.Memo.O` is already `open`ed at the top.


emillon:
What does the alternative look like? That is, if we remove the global `open`, how many local opens do we have to add?

(using local opens is what we suggest in our style guide, but for cases like `Dune_file` opening `Dune_lang.Decoder` it makes sense, maybe this is one of the exceptions too)

Alizter:
I also prefer opening locally, especially when it's anything other than `Memo.O`.

rgrinberg (approved):
Feel free to experiment with different styles of maintaining the opens. In any case, this PR is a a strict improvement

anmonteiro:
This PR removes 5 `open` calls. Removing the toplevel call would result in adding 8 `open` expressions.

Since `Resolve.Memo.O` is used very pervasively throughout this module, I'd consider it to be one of the exceptions @emillon mentioned, and I personally prefer preserving its current state.

Alizter (approved):
Since we already have a top level one, I am ok with this.

PR #7461, opened 2023-03-31
[3.7] backport #7445 (Allow `package` in any location in `rule`)

emillon:
While fixing the package associations in ocaml/opam#5496 I realized that the package stanza is only accepted when it is at the end of the rule stanza.

* Add test that `packages` needs to be last, or it will be rejected

    File "dune", line 12, characters 3-10:
    12 |   (package a)
            ^^^^^^^
    Error: Unknown action or rule field.
    [1]

Signed-off-by: Marek Kubica <marek@tarides.com>
Co-authored-by: Etienne Millon <me@emillon.org>


Leonidas-from-XIV:
I was about to propose this, but I am not entirely sure this makes sense to backport.

To be able to use `package` everywhere one would need to depend on dune-lang 3.7.1, as dune-lang 3.7 would potentially pick too old of a dune version, so I feel having it only available in 3.8 would make more sense, as it would be available in all 3.8 versions.

emillon:
Ah, indeed there might be a problem. If your fix makes some dune files valid (that were invalid in older dune versions), it might make sense to version it (and not backport it).

@rgrinberg what do you think? Should the fix in #7445 be versioned or is it ok to keep it unversioned?

(my opinion is that it's unlikely to come up in the wild and that code path may be hard to version, so I'd keep it unversioned)

rgrinberg:
I think we should keep #7445 to be versioned because the issue will come up with other fields that we'll need to add to that list. Might as well bite the bullet and do it now.

emillon:
OK, let's version it then (#7471). This disqualifies it for a backport so I'm closing this one. Thanks

PR #7169, opened 2023-02-23
fix: Revert "fix: Invoke preprocessor commands from local dir"

rgrinberg:
Running the preprocessor from cwd breaks the error messages outputted by
the preprocessor.

Signed-off-by: Rudi Grinberg <me@rgrinberg.com>

PR #7057, opened 2023-02-13
Invoke preprocessor commands from local dir

gridbugs:
Prior to this change commands listed in the `(pps ...)` field were run from the workspace root. This meant that the `%{project_root}` variable couldn't be used when constructing preprocessor commands, as it returns a path relative to the directory containing the `dune` file. After this change the preprocessor commands are run from the directory containing the dune file which defines them. This behaviour is feature gated as some projects may have come to rely on the original behaviour.

Issue #7043, opened 2023-02-11
Dune tests are being executed in unexpected dir while vendoring


Kakadu:
https://github.com/Kakadu/dune-test-rel-dir/tree/v1

##### Setup

The inner project is `inner-vendored`, the top project is `dune-project-root`, and it has a symlink to inner project to emulate vendoring.

We can `(cd inner-vendored && dune test)` and it works fine. More precisely, the library `dir/lib1.cmxa` uses external preprocessor `dir/rewriter1.exe` to preprocess.

##### Issue

When we vendor inner project in the root one (right now, it is done via symlink) the tests are being executed in the context of root project, meaning from `%{workspace_root}` and not from inner `%{project_root}`. How to implement it idiomatically?

##### Possible workarounds

We can prefix invocation of rewriter executable by the difference between %{workspace_root} and
vendored %{project_root}. It seem to work, but hacky.

rgrinberg:
@gridbugs could you take a look at this?

gridbugs:
I'm going to assume that when `%{project_root}` appears in a `dune` file, it should resolve to the directory which contains a `dune-project` file which is the nearest ancestor of the `dune` file. I'll look into why that is not happening in this case.

gridbugs:
In the repro one of the commented out lines is:
```
  ;(pps ppx_inline_test -- -pp %{project_root}/dir/rewriter1.exe)
  ; works from top project but doesn't if inner is not vendored
```
I can reproduce that problem. The error is:
```
File "dir/dune", line 27, characters 2-64:
27 |   (pps ppx_inline_test -- -pp %{project_root}/dir/rewriter1.exe)
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sh: line 1: ../dir/rewriter1.exe: No such file or directory
File "dir/lib1.ml", line 1:
Error: Error while running external preprocessor
```
`%{project_root}` is getting resolved to ".." which is the relative path from the current directory (the "dir" directory) to the directory which contains the project root which is expected, but the path is then being used relative to the project root rather than "dir".

rgrinberg:
I see. Well that's a known issue. How about we start expanding the variables relative to the directory where we run the pp when the dune language in the project file is greater than 3.8. I don't think we can change the behavior without preserving it for older projects.

gridbugs:
Sounds good.

Kakadu:
[...]

PR #7463, opened 2023-03-31
Add test to show that relative paths in preprocess stanzas fail

Leonidas-from-XIV:
Preprocessing fails due to `./gen/gen.exe` being expected to be relative from the workspace root, not relative to `src/`.


rgrinberg:
I think we have something like this already. Right @gridbugs?

gridbugs:
Yes, there is a test in test/blackbox-tests/test-cases/preprocessor-directory.t. That test was written as a repro for a feature (https://github.com/ocaml/dune/pull/7057) where we changed dune so the preprocessor ran from the local directory rather than project_root but that change has since been reverted (https://github.com/ocaml/dune/pull/7169) as it broke error messages.

There is an issue about the problem of preprocessor command running from project_root here: https://github.com/ocaml/dune/issues/7043

The test I added for in https://github.com/ocaml/dune/pull/7057 remains as a kind of repro of the problem, but the test added in this PR is more comprehensive and clearer, so I'd be happy for you to delete test/blackbox-tests/test-cases/preprocessor-directory.t and add a note about the issue https://github.com/ocaml/dune/issues/7043 to your test so we remember to update the issue if/when this behaviour gets changed.

PR #6611, opened 2022-12-01
feature: make `modes` field an Ordered_set_lang instance

anmonteiro:
Allows writing `(modes :standard melange)` which will be useful when converting libraries in existing repositories.

Signed-off-by: Antonio Nuno Monteiro <anmonteiro@gmail.com>

PR #7467, opened 2023-03-31
fix: make lib mode `:standard` match defaults

jchavarri:
It seems the `:standard` mode for libraries was not including the "inherited" values for `byte` mode.

This would lead to some errors like the following, when trying to build libraries for an executable with explicit `byte` mode:

```
File "_none_", line 1:
Error: Module `Mylib' is unavailable (required by `Dune__exe__Main')
```

The redundancy of default values makes the code a bit brittle, I added comments to prevent future breakages.

### DIFF
--- a/src/dune_rules/dune_file.ml
+++ b/src/dune_rules/dune_file.ml
@@ -267,6 +267,8 @@ module Mode_conf = struct
       | Native
       | Best
 
+    let all = [ Byte; Native; Best ]
+
     let compare x y =
       match (x, y) with
       | Byte, Byte -> Eq
@@ -334,6 +336,14 @@ module Mode_conf = struct
             | Some Inherited ->
               (* this doesn't happen as inherited can't be manually specified *)
               assert false))
+
+    let to_list (t : t) : (mode_conf * Kind.t) list =
+      let get mode_conf =
+        match Map.find t mode_conf with
+        | None -> None
+        | Some k -> Some (mode_conf, k)
+      in
+      List.filter_map ~f:get all
 
     let decode =
       let decode =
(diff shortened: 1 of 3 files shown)


src/dune_rules/dune_file.ml
```diff
@@ -460,7 +461,11 @@ module Mode_conf = struct
               in
               (mode, Kind.Requested loc))
             ~eq:(fun (a, _) (b, _) -> equal a b)
-            ~standard:[ (Ocaml Best, Kind.Requested stanza_loc) ]
+            ~standard:
```
anmonteiro, line 464: I'd add a `Mode_conf.Set.Default.to_list` or something, and we'd be synced by default.

anmonteiro (approved):
LGTM pending my suggestion. Looks like a bug I introduced in https://github.com/ocaml/dune/pull/6611

rgrinberg:
Before merging, would it be possible to split the PR into two commits:

1. First commit that introduces the test and demonstrates that it fails.
2. A second commit that fixes the bug and modifies the test introduced in 1

src/dune_rules/dune_file.ml
jchavarri: Fixed in https://github.com/ocaml/dune/pull/7467/commits/b21c4a3710b86a8458a0e88c25cc4c455a467245.

jchavarri:
> would it be possible to split the PR into two commits

Done in https://github.com/ocaml/dune/pull/7467/commits/83387d183641b3336f0d085bd39007c46f4f7cc1 and https://github.com/ocaml/dune/pull/7467/commits/b21c4a3710b86a8458a0e88c25cc4c455a467245.

PR #7468, opened 2023-03-31
refactor: stop using Sys.linux

rgrinberg:
There's now a [Platform] module

### DIFF
--- a/otherlibs/stdune/src/platform.mli
+++ b/otherlibs/stdune/src/platform.mli
@@ -1,10 +1,17 @@
+(** Platform detection. *)
+
 module OS : sig
+  (** Detect the operating system. *)
+
   type t =
     | Darwin
     | Linux
+    | Windows
     | Other
 
+  (** [value] is the current os we're running on. *)
   val value : t
 end
 
+(** [assert_os os] assert that we're running on [os]. *)
 val assert_os : OS.t -> unit
--- a/src/dune_file_watcher/dune_file_watcher.ml
+++ b/src/dune_file_watcher/dune_file_watcher.ml
@@ -340,12 +340,12 @@ let fswatch_backend () =
     User_error.raise [ Pp.text "Please install fswatch to enable watch mode." ]
 
 let select_watcher_backend () =
-  if Sys.linux then (
-    assert (Ocaml_inotify.Inotify.supported_by_the_os ());
-    `Inotify_lib)
-  else if Fsevents.available () then `Fsevents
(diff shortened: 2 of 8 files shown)


otherlibs/stdune/src/platform.mli
```diff
@@ -2,6 +2,7 @@ module OS : sig
   type t =
```
Alizter, line 2: Can you add some documentation to this file?

src/dune_file_watcher/dune_file_watcher.ml
```diff
@@ -340,12 +340,13 @@ let fswatch_backend () =
...
-  if Sys.linux then (
+  match Platform.OS.value with
+  | Linux ->
     assert (Ocaml_inotify.Inotify.supported_by_the_os ());
-    `Inotify_lib)
-  else if Fsevents.available () then `Fsevents
-  else if Sys.win32 then `Fswatch_win
-  else fswatch_backend ()
+    `Inotify_lib
+  | Darwin -> if Fsevents.available () then `Fsevents else `Fswatch_win
```
Alizter, line 347: That doesn't seem correct. Perhaps you should just use `with` in the match case and point to `Fsevents like it was before. The the Other case should be a catch all.

rgrinberg: fixed now

anmonteiro (approved):

src/dune_file_watcher/dune_file_watcher.ml
```diff
@@ -340,12 +340,14 @@ let fswatch_backend () =
     User_error.raise [ Pp.text "Please install fswatch to enable watch mode." ]
 
 let select_watcher_backend () =
-  if Sys.linux then (
-    assert (Ocaml_inotify.Inotify.supported_by_the_os ());
-    `Inotify_lib)
-  else if Fsevents.available () then `Fsevents
-  else if Sys.win32 then `Fswatch_win
-  else fswatch_backend ()
+  if Fsevents.available () then `Fsevents
```
anmonteiro, line 343: my understanding is that `fsevents` is only available on macOS. 

If we were to perform that check in the `match` expression below, however, we'd need an extra clause for `Other`, that's why there's an `if` expression. Is that right?

rgrinberg: There's an if expression because we need to use `Fsevents.available` somehow. It's possible that we'd disable on some versions on macos so it could be useful

Alizter (approved):

Issue #2572, opened 2019-08-22
How to use `configurator` to fetch `pkg-config` for the cross-compiling env?


toots:
`ocaml-shine` has now been converted to use `dune` as build system!

However, when cross-compiling for `windows`, the [discover](https://github.com/savonet/ocaml-shine/blob/master/src/config/discover.ml) program used to fetch `pkg-config` information is compiled as a native binary and returns `pkg-info` related to the build host and not the target cross-compilation.

In `configure` world, this is handled by passing a host triplet and expecting the binary to use to also be prefixed with the same host triple, i.e. `x86_64-w64-mingw32-pkg-config` and etc.

Could it be possible to make this work with `configurator`?

toots:
Just wanted to bring this up again. When cross-compiling with dune, I now set `PKG_CONFIG_PATH` to the cross-compiler's path to `pkg-config` files. However, this is tricky. Since `dune` can implicitly also build binaries for the native environment if e.g. they need to be run locally, this will, in turn, trigger the `discover` script.

However, forcing `PKG_CONFIG_PATH` means that the `discover` script only sees the cross-compiling pkg-config info so we cannot configure the build to be able to build both for the native and target host using `pkg-config`.

In an ideal work, it should be possible to use either the `context` variable or, possibly, the `toolchain` variable (#3949) to write a `discover` module using `dune-configurator` that would pick the correct `pkg-config` binary and not have to worry about internals that are already pre-configured for each binary.

jeremiedimino:
We just discussed this during the dune dev meeting. It seems like adding a `pkg-config` variable to the ocamlfind toolchain definition could help here. Then configurator would call this `pkg-config` binary. WDYT?

toots:
I think that this would work great!

anmonteiro:
@toots I'm fixing this in https://github.com/ocaml/dune/pull/7469 (respecting `$PKG_CONFIG`, when set). This works for my cross-compilation workflow, and I suspect one could also set this environment variable for OPAM-cross builds.

What do you think?

toots:
In practice, `PKG_CONFIG_PATH` is usually set to the path to the cross-compilation pkg-config files. The path to `pkg-config` is not usually changed.

At the moment, all currently existing examples of cross-compiled modules do have the `PKG_CONFIG_PATH` variable set unconditionally to the cross-compilation files in the `opam-cross` CI tests.

[...]

PR #7469, opened 2023-04-02
dune-configurator: respect $PKG_CONFIG

anmonteiro:
- `PKG_CONFIG` is commonly used: there's prior art in [Nix](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/pkg-config-wrapper/setup-hook.sh#L26), [Cmake](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html)
  - in cross-compilation scenarios, the target environment can set `PKG_CONFIG` to the prefixed executable, e.g. `aarch64-unknown-linux-musl-pkg-config`

fixes https://github.com/ocaml/dune/issues/2572


anmonteiro:
This allows to remove a `pkg-config` hack in the nix-overlays, for example: https://github.com/nix-ocaml/nix-overlays/pull/716/commits/34f688d65dd267762c20a923f28eba118cebc2a5

rgrinberg:
Thanks for the PR. Some questions about nix/cmake so I understand this variable correctly.

> export PKG_CONFIG${role_post}=@targetPrefix@@baseBinName@

What's role_post in this line?

From the [cmake docs](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#variables-affecting-behavior)

Should we support the `PKG_CONFIG_EXECUTABLE` variable as well?

> The PKG_CONFIG environment variable can be used as a hint if PKG_CONFIG_EXECUTABLE has not yet been set.

What does it mean to be used as a hint?

In the current implementation, should we allow command line arguments to bw included in this variable? E.g. is this valid: `PKG_CONFIG=pkg-config --foo`?

Finally, could you make a corresponding change to pkg_config.ml in the rules? We also support pkg_config there and it would be good to be consistent.

anmonteiro:
> What's role_post in this line?

Look like it sets `_FOR_BUILD` and `_FOR_TARGET`. see [this](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/setup-hooks/role.bash). It looks only used internally, while `PKG_CONFIG` is exposed to the shell.




> Should we support the `PKG_CONFIG_EXECUTABLE` variable as well?

From what I've read, cmake will look for `PKG_CONFIG` too. [here](https://fuchsia.googlesource.com/third_party/cmake/+/v3.5.1/Modules/FindPkgConfig.cmake) they look for it first. In those cases `PKG_CONFIG_EXECUTABLE` is set from `PKG_CONFIG`. My suggestion is to support only one variable, since `_EXECUTABLE` seems specific to cmake.





> In the current implementation, should we allow command line arguments to bw included in this variable? E.g. is this valid: `PKG_CONFIG=pkg-config --foo`?

I feel like it might be valid indeed, even if there are issues with it in the [wild](https://gitlab.kitware.com/cmake/cmake/-/issues/22305). I'm not sure what to do here. My bias is to support just the binary and adapt it later based on user feedback.


> Finally, could you make a corresponding change to pkg_config.ml in the rules? We also support pkg_config there and it would be good to be consistent.

yup, will do!

anmonteiro:
Added the same logic to `Pkg_config.ml` in rules.

src/dune_rules/pkg_config.ml
```diff
@@ -1,5 +1,11 @@
 open Import
 
+let pkg_config_binary sctx =
+  let env = Super_context.context_env sctx in
+  match Env.get env "PKG_CONFIG" with
```
rgrinberg, line 5: We should probably allow overriding this variable. I recommend fetching the environment from `Super_context.env_node`

anmonteiro: Ah yeah, env_node makes sense. I changed to that.

rgrinberg (approved):
Don't forget to update the CHANGES

PR #7472, opened 2023-04-03
Fix EMLINK on Windows (again)

nojb:
This is a follow-up to #6993. Unfortunately, a typo (1142 -> 1442) meant that the fix there was useless. Also add a similar fix in the other place where `EMLINK` is handled.

For background (this is also explained in #6993), the Windows error code corresponding to `EMLINK` is `ERROR_TOO_MANY_LINKS` (= 1142). The mapping was fixed upstream https://github.com/ocaml/ocaml/pull/11991, but is only available for OCaml >= 5.1, so we need a mitigation in Dune for previous versions.

image

@emillon this should be backported to 3.7.1.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -97,6 +97,9 @@ Unreleased
 - Allow `(package ...)` in any position within `(rule ...)` stanza (#7445,
   @Leonidas-from-XIV)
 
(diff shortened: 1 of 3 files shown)


emillon:
backported in #7475, thanks

PR #7473, opened 2023-04-03
Add command to wait until the next watch-mode build completes

gridbugs:
This is a refinement of https://github.com/ocaml/dune/pull/7443. It adds an RPC command "watch_mode_wait" which only returns the next time a build completes when dune is running in watch mode. For the moment this RPC is internal and not part of the publicly-visible RPC interface. There is an accompanying command `dune internal watch-wait` which returns when the current watch-mode build completes by using the new RPC method.

### DIFF
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -73,6 +73,13 @@ let run_build_system ~common ~request =
       in
       res)
     ~finally:(fun () ->
+      let* () =
+        match Common.rpc common with
+        | `Allow server ->
+          Dune_rpc_impl.Server.Watch_mode_waiters.wake_all
+            (Dune_rpc_impl.Server.watch_mode_waiters server)
+        | `Forbid_builds -> Fiber.return ()
+      in
       Hooks.End_of_build.run ();
       Fiber.return ())
 
--- a/bin/internal.ml
+++ b/bin/internal.ml
@@ -10,4 +10,8 @@ let latest_lang_version =
 
 let group =
   Cmd.group (Cmd.info "internal")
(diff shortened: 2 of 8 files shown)


gridbugs:
I had a go at implementing a benchmark runner based on this feature and ran into a potential race condition. The runner changes a file and then waits for the next build to complete. Changing a file causes a build to start, but if the build completes before the runner starts waiting, it will miss the current build and end up waiting for the _next_ build to complete and thus waiting forever.

This is the problem I was trying to solve in https://github.com/ocaml/dune/pull/7443 with sequence numbers, but I have an alternative solution that is simpler than my previous attempt.

I'll change the RPC so that instead of waiting until the next build completes, it waits for the _current_ build to complete, and returns immediately if there is no build in progress. Additionally, rather than returning a `unit` it will return an `int` which is the number of builds that have completed since the server started. This will let the benchmark runner avoid this race condition by calling the wait RPC in a loop until the count increases. This will also simplify starting the RPC server and waiting for the initial build, as without some sort of counting mechanism it's tricky for the runner to know when the initial build is complete (there's a risk that the first wait will take place before the first build starts or after the first build finishes or during the first build).

gridbugs:
Thinking more about this, I don't think we even need and RPC to let a client wait for a build at all. The simplest thing that will work is an RPC that always returns immediately with the current count of builds that have completed, and the client can call this method in a loop (with a delay between each call) to find out when the current build completes. We're not using this RPC for measuring build times - just to tell the client when it's time to check the trace file.

gridbugs:
Closing this in favour of new PR that just adds a method for counting how many builds have finished: https://github.com/ocaml/dune/pull/7485

PR #7476, opened 2023-04-03
Only enable `package` in short-form rules on dune lang 3.8+

### LINKED ISSUE #7471: #7445 check needs to be versioned
As discussed in #7461, the check in #7445 needs to be versioned so that `(package)` is only accepted in head position if lang dune >= 3.8.

Leonidas-from-XIV:
Closes #7471, as discussed in #7461 to only allow the #7445 feature on dune-lang 3.8.

This fixes the issue that #7445 by itself fixes a language quirk that is only fixed in 3.8, but without actually requiring dune-lang 3.8. It revises the previous test that uses an older dune-lang to fail.

### DIFF
--- a/src/dune_rules/dune_file.ml
+++ b/src/dune_rules/dune_file.ml
@@ -1740,15 +1740,15 @@ module Rule = struct
           ~hints:
             (User_message.did_you_mean s
                ~candidates:(String.Map.keys atom_table))
-      | Some Field -> (
-        match s with
-        | "package" ->
-          let* () =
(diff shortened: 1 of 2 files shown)


emillon (approved):
Thanks @Leonidas-from-XIV . I refactored the implementation. so that the atom is not duplicated into the check. Does that this right to you?

Leonidas-from-XIV:
Ah yes, that's a much nicer implementation indeed!

PR #7481, opened 2023-04-03
refactor(engine): introduce a `Lookup_alias.t` type in Action_builder

anmonteiro:
- this is a follow-up to #7334, which only changed `dune_rules`
- the 1st commit refactors `Action_builder` to introduce a new intermediate `Lookup_alias` module that can be used in the `Alias_rec` traversal callback
  - the additional information provided, `allowed_build_only_subdirs`, lets the caller know about build-only subdirectories where rules are defined that dune may need to traverse in order to look up alias rules
- the 2nd commit demonstrates how this new type can be used to refactor the recursive aliases code in `dune_rules`.


src/dune_engine/action_builder.mli
```diff
@@ -113,7 +113,14 @@ end
 
 val alias : Alias.t -> unit t
 
-val dep_on_alias_if_exists : Alias.t -> Alias_status.t t
+module Lookup_alias : sig
+  type t =
+    { alias_status : Alias_status.t
+    ; allowed_build_only_subdirs : Filename.Set.t
```
snowleopard, line 119: Could you add some comments here explaining this new field?

snowleopard (approved):
Thanks! I think the change makes sense. It seems worth adding some comments for the new richer API.

@rgrinberg Please take a look too.

snowleopard:
@anmonteiro Thanks for adding the comment. After reading it, the naming started to look a bit odd though: we're calling `dep_on_alias_if_exists` but get not just an alias but some seemingly unrelated information about the build-only directories. Not a very big deal, but perhaps you have any ideas on how to make the naming a little clearer?

anmonteiro:
Good feedback. I'll think of something.

anmonteiro:
I had dropped the ball on this. Just pushed a commit:

- renamed `Lookup_alias.t` to `Alias_build_info`
- renamed `dep_on_alias_if_exists` to `dep_on_alias_build_info_if_exists`.

anmonteiro:
I plan on merging this change once CI passes, given:

- PR has been approved before
- the new naming is strictly better than what I had came up with previously

rgrinberg:
Let me have a look at it tomorrow first. I forgot I was supposed to read this.

src/dune_engine/action_builder.ml
```diff
@@ -278,30 +278,59 @@ module Alias_status = struct
...
+    in
+    { alias_status = status; allowed_build_only_subdirs }
+end
+
+let dep_on_alias_build_info_if_exists alias =
   of_thunk
     { f =
         (fun mode ->
           let open Memo.O in
-          let* definition = Load_rules.alias_exists alias in
```
rgrinberg, line 286: Do we still need `alias_exists` in the API of `Load_rules`?

rgrinberg (approved):

PR #7483, opened 2023-04-04
feature: allow opam files in opam/

rgrinberg:
When we set (opam_file_location inside_opam_directory), we are allowed
to put all the opam file in the opam/ directory relative to the project.

### DIFF
--- a/src/dune_rules/dune_project.ml
+++ b/src/dune_rules/dune_project.ml
@@ -152,6 +152,7 @@ type t =
   ; strict_package_deps : bool
   ; cram : bool
   ; expand_aliases_in_sandbox : bool
+  ; opam_file_location : [ `Relative_to_project | `Inside_opam_directory ]
   }
 
 let equal = ( == )
@@ -215,6 +216,7 @@ let to_dyn
     ; strict_package_deps
     ; cram
     ; expand_aliases_in_sandbox
+    ; opam_file_location
     } =
   let open Dyn in
   record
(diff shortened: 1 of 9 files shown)


Alizter:
Would we want this to be more configurable? There might be users that don't want an `opam/` directory but would like to give their own.

rgrinberg:
> There might be users that don't want an opam/ directory but would like to give their own.

This is the only directory that opam understands for the purposes of pinning

src/dune_rules/dune_project.ml
```diff
@@ -243,6 +245,10 @@ let to_dyn
     ; ("strict_package_deps", bool strict_package_deps)
     ; ("cram", bool cram)
     ; ("expand_aliases_in_sandbox", bool expand_aliases_in_sandbox)
+    ; ( "opam_file_lcoation"
```
emillon, line 248:
```suggestion
    ; ( "opam_file_location"
```

src/dune_rules/opam_create.ml
```diff
@@ -281,7 +282,23 @@ let generate project pkg ~template =
     | None -> ""
     | Some (_, s) -> s)
 
-let add_rule sctx ~project ~pkg =
+let add_alias_rule sctx ~project ~pkg =
```
anmonteiro, line 285: If I'm reading this correctly, then we still emit the `.opam` files in the repository root inside the build folder?

anmonteiro: to answer my own question: we don't emit `*.opam` files in `opam/`, we just expect them to be there.

anmonteiro (approved):
LGTM. 

The only question I have, for which I'm unsure of the tradeoffs, is:

the difference between telling gen_rules about this `opam/` directory as an automatic_subdir vs making `source_tree` aware of it.

rgrinberg:
> vs making source_tree aware of it.

I'm in the process of moving away as much logic as possible from Source_tree. The problem with all this partial evaluation of dune files is that makes it impossible to make anything generated by rules or interpreted differently based on the context or profile.

Alizter:
@emillon This does not appear to be documented.

emillon:
indeed. do you want to submit a PR? I think it's just a matter of adding the new field in `dune-files.rst`.

ejgallego:
Also a CHANGES entry for this seems to be missing.

ejgallego:
The lack of changes file means that we can't document to users what dune version they need in order to use this feature.

Alizter:
If its any consolation, it was introduced in 3.8.

PR #7484, opened 2023-04-04
tui: add more interactions

Alizter:
We add some more features to the `--display tui` mode.

Waiting for #7663 

Features:
- Messages to the user a separated with a rule and include a numbering together with total number of messages.
- Horizontal and vertical scroll bars which can be controlled using the arrow keys, `j` and `k` or by interacting with the mouse including scrolling and clicking. (Horizontal scrolling is done with Meta (alt) + Scroll.
- Help screen activated with `h` or `?` explaining what keys do what.
- Debug features:
  - Pressing 'd' will output some debug info about the state of the UI.
  - Pressing Meta + `l` will make dune output a lorem ipsum, allowing to quickly test any issues with scrolling. (I want to keep this in as it is harmless and out of the way).

#### How to test:
Just run
```console
./dune.exe build @install
_build/install/bin/dune build @check -w --display tui
```

#### Compatibility
- Linux
  - [x] Alacritty, works fine, this is in the screenshots
  - [x] xterm, works but meta + l is sometimes ignored
  - [x] gnome console, works fine
- [ ] macos
- [ ] windows
  - [ ] cmd
  - [ ] wsl

#### Issues
- [ ] Exceptions are not printed when in tui mode. We need to have a fallback display. For example, running two watch mode tui's will cause the second one to crash silently and return to the prompt since Dune will complain about the _build lock. We should probably add a way internal exceptions can be displayed or printed nicely.
- [x] Needs to look good on light terminals, the colors used are somewhat ad-hoc. We need to make them more configurable and then adapt based on terminal colors chosen.

### DIFF
--- a/otherlibs/stdune/src/dune
+++ b/otherlibs/stdune/src/dune
@@ -12,6 +12,6 @@
   (re_export dune_filesystem_stubs))
  (foreign_stubs
   (language c)
-  (names wait3_stubs platform_stubs copyfile_stubs))
+  (names wait3_stubs platform_stubs copyfile_stubs signal_stubs))
  (instrumentation
   (backend bisect_ppx)))
--- a/otherlibs/stdune/src/float.ml
+++ b/otherlibs/stdune/src/float.ml
@@ -10,3 +10,8 @@ let max x y =
   match compare x y with
   | Eq | Gt -> x
   | Lt -> y
+
+let min x y =
+  match compare x y with
+  | Eq | Lt -> x
+  | Gt -> y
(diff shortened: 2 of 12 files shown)


rgrinberg:
do we really need a confirmation? to me this is something more annoying than useful

Alizter:
@rgrinberg Sure I can remove it no problem. Its good to demonstrate extra state tracking for the UI though.

rgrinberg:
Can you extract resizing into a separate PR?

Alizter:
@rgrinberg It's not working completely for me, I will make a separate PR when I get it to work.

Alizter:
Here is a screenshot of the latest features:
image
Here you can see:
- The help screen
- The horizontal and vertical scrollbars (controlled with scroll and meta (alt) + scroll)
- Message separation with numbering

This needs:
- more testing, especially on edge cases.
- update help page for more info (debug and scroll)
- mouse interaction with scroll bars?

Alizter:
This is being worked on at the moment but will come as a different PR.

PR #7485, opened 2023-04-04
RPC that returns the number of completed builds

gridbugs:
Next iteration of https://github.com/ocaml/dune/pull/7473. This is much simpler and doesn't have any synchronization. It just returns the number of builds that have completed since dune was started in watch mode. The idea is that this can be used by the benchmark runner to check when a build has completed.

### DIFF
--- a/src/dune_rpc_impl/server.ml
+++ b/src/dune_rpc_impl/server.ml
@@ -208,6 +208,7 @@ type t =
       (Dep_conf.t list * Build_outcome.t Fiber.Ivar.t) Job_queue.t
   ; watch_mode_config : Watch_mode_config.t
   ; mutable clients : Clients.t
+  ; build_count : int ref
   }
 
 let ready (t : t) =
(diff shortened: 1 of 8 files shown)


src/dune_rpc_impl/server.ml
```diff
@@ -208,6 +208,7 @@ type t =
       (Dep_conf.t list * Build_outcome.t Fiber.Ivar.t) Job_queue.t
   ; watch_mode_config : Watch_mode_config.t
   ; mutable clients : Clients.t
+  ; build_count : int ref
```
rgrinberg, line 211: I think this can just be `; mutable build_count : int` and accessable via `t` in the request implementation.

rgrinberg (approved):
LGTM. Although why does it need a command? Is it just for testing?

gridbugs:
> LGTM. Although why does it need a command? Is it just for testing?

Without the command we would have to add this RPC to the public RPC interface in `dune-rpc` and do a release in order for external programs to use it. The benchmark runner is an external program and I wanted to work on it without needing to first release dune-rpc.

Also this PR doesn't even add the new RPC to the public interface. Should I add it? How much planning and discussion is normally needed to change the public RPC interface?

gridbugs:
Applied feedback in new PR https://github.com/ocaml/dune/pull/7552

rgrinberg:
> Also this PR doesn't even add the new RPC to the public interface. Should I add it? How much planning and discussion is normally needed to change the public RPC interface?

There's no need to change the public interface though. You can add it to `dune_rpc_private`.

gridbugs:
Oh cool, I didn't realize that that was possible. I'll update https://github.com/ocaml/dune/pull/7552 to add the RPC to `dune_rpc_private` and update the benchmark runner (https://github.com/ocaml-dune/ocaml-monorepo-benchmark/tree/main/dune-benchmark-runner) to use RPC rather than shelling out to the dune subcommand.

PR #7491, opened 2023-04-04
refactor: clean up imports in gen_rules

rgrinberg:

### DIFF
--- a/src/dune_rules/gen_rules.ml
+++ b/src/dune_rules/gen_rules.ml
@@ -1,6 +1,4 @@
 open Import
-module Toplevel_rules = Toplevel.Stanza
-open Dune_file
 open Memo.O
 
 module For_stanza : sig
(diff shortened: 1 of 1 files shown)


src/dune_rules/gen_rules.ml
```diff
@@ -64,9 +62,11 @@ end = struct
   let of_stanza stanza ~sctx ~src_dir ~ctx_dir ~scope ~dir_contents ~expander
       ~files_to_install =
     let dir = ctx_dir in
+    let toplevel_setup = Toplevel.Stanza.setup in
+    let open Dune_file in
     match stanza with
     | Toplevel toplevel ->
-      let+ () = Toplevel_rules.setup ~sctx ~dir ~toplevel in
+      let+ () = toplevel_setup ~sctx ~dir ~toplevel in
```
anmonteiro, line 69: if `Topleve.Stanza.setup` is only used here, do we need the `let toplevel_setup = Toplevel.Stanza.setup` expression?

anmonteiro (approved):

src/dune_rules/gen_rules.ml
rgrinberg: We need it because `let open Dune_file` shadows the `Toplevel` module.

PR #7496, opened 2023-04-05
chore: digest benchmarking - larger string

rgrinberg:
add benchmarking of file and string digests


Alizter (approved):

Alizter:
btw the commit message is wrong I think

PR #7497, opened 2023-04-05
Fix a top-level computation raising in [otherlibs/stdune/src/fpath.ml]

snowleopard:
The problem this commit is fixing is kind of obvious but if you'd like to reproduce it, you can do something like
```
mkdir dir
cd dir
rmdir ../dir
dune --help
```

### DIFF
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -316,19 +316,9 @@ module Lock = struct
     | Some path, None ->
       let repo_id = Repository_id.of_path path in
       Fiber.return @@ [ Opam_repo.of_opam_repo_dir_path ~source:None ~repo_id path ]
-    | None, Some url ->
-      let repo = Fetch.Opam_repository.of_url url in
-      let+ opam_repository = Fetch.Opam_repository.path repo in
-      (match opam_repository with
-       | Ok { path; repo_id } ->
-         [ Opam_repo.of_opam_repo_dir_path
-             ~source:(Some (OpamUrl.to_string url))
-             ~repo_id
-             path
-         ]
-       | Error _ ->
-         User_error.raise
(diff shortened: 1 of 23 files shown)


Alizter:
Can you add a cram test?

snowleopard:
Sure, I'll add a test.

Top-level computation that raise is a red flag in general, but it's pretty easy to arrange a situation where the directory where you were running your build system disappears (this happens pretty often internally where a workspace gets deleted after the user releases the corresponding branch/feature).

otherlibs/stdune/src/fpath.ml
```diff
@@ -1,6 +1,18 @@
 let is_root t = Filename.dirname t = t
 
-let initial_cwd = Stdlib.Sys.getcwd ()
+let initial_cwd : string option =
+  try Some (Stdlib.Sys.getcwd ()) with _exn -> None
```
rgrinberg, line 4: Could you include the original exception in the error message? It should be `Sys_error msg` I believe.

It's possible that one couldn't determine the cwd because of permission issues perhaps. It would be useful to diagnose that if the actual error was included.

snowleopard: Indeed, good idea.

otherlibs/stdune/src/fpath.ml
```diff
@@ -1,6 +1,18 @@
...
+
+let initial_cwd : string Lazy.t =
+  lazy
+    (match initial_cwd with
+    | Some cwd -> cwd
+    | None ->
+      User_error.raise
+        [ Pp.textf
+            "Dune could not determine the current working directory. Did you \
+             run Dune from a directory that no longer exists?"
```
Alizter, line 14: You can pass the `Did you` as a `~hint`

PR #7504, opened 2023-04-06
Update vendored code to add the LICENSE files

Leonidas-from-XIV:
This adds the missing license files as well as updates the scripts to copy the license going forward.

The only missing one is [build_path_prefix_map](https://gitlab.com/gasche/build_path_prefix_map). It doesn't have a license file — the license is specified to be MIT in its OPAM file.

Thanks for @rgrinberg to point it out and @Alizter for looking which files are missing.


Alizter:
I've filed an issue for build_path_prefix_map to add a license file: https://gitlab.com/gasche/build_path_prefix_map/-/issues/1

Alizter (approved):

Issue #5852, opened 2022-06-08
`(public_name -)` different behaviour between `executable` and `executables`


maiste:
## Expected Behavior

Hi, dear dune maintainers,
While working with `executable` and `executables` I noticed a difference in behaviour with the stanza `(public_name -)` and `(public_names - -)`.
As described in the documentation, they should work the same way:
> The executables stanza is the same as the executable stanza except that it’s used to describe several executables sharing the same configuration, so the plural executables stanza is used to represent more than one executable.

## Actual Behavior

IIUC, the two executables descriptions should work the same way as one is just a contraction of the other one:

1.
```lisp
 (executable
   (public_name -)
   (name example_1)
   (modules example_1)
   (package dune-example))

 (executable
   (public_name -)
   (name example_2)
   (modules example_2)
   (package dune-example))
```

2.
```lisp
 (executables
   (public_names - -)
   (names example_1 example_2)
   (package dune-example))
```

Whereas `2.` works without any issue and the  `-` name is taken as _ignored this field_, `1.` raised the following error:
```
Error: Multiple rules generated for _build/install/default/bin/-:
- examples/dune:3
- examples/dune:9
-> required by _build/default/dune-example.install
-> required by alias all
-> required by alias default
```

## Reproduction

The reproduction steps are:
```
git clone https://github.com/maiste/dune-example.git
cd dune-example
dune build .
```

## Specifications

- Version of `dune` (output of `dune --version`): 3.2.0
- Version of `ocaml` (output of `ocamlc --version`): 4.14.0
- Operating system (distribution and version): Ubuntu 20.04 / EndeavourOS Linux

nojb:
> one is just a contraction of the other one

Orthogonal to the issue described here, but note that this is not actually correct: `executables` allow _sharing_ modules between the different executables, while this is not possible if you replace it by multiple `executable` stanzas. Probably a good idea to fix the documentation to point this out.

> raised the following error:

This is probaly because using `-` in a single-executable stanza doesn't make a lot of sense: you may as well just supress the `public_name` field altogether. On the other hand, it may be a good idea to implement the same behaviour as for `executables` just for the sake of uniformity.

maiste:
> Orthogonal to the issue described here, but note that this is not actually correct: `executables` allow _sharing_ modules between the different executables, while this is not possible if you replace it by multiple `executable` stanzas. Probably a good idea to fix the documentation to point this out.

[...]

PR #7505, opened 2023-04-06
Document advantage of `executables` over `executable`

Leonidas-from-XIV:
One aspect of #5852 revealed that this feature is under-documented, so documenting the current behavior is a good first step.

### DIFF
--- a/doc/stanzas/executable.rst
+++ b/doc/stanzas/executable.rst
@@ -8,9 +8,9 @@ executable stanzas is as follows:
 
 .. code:: dune
 
-    (executable
-     (name <name>)
-     <optional-fields>)
+   (executable
+    (name <name>)
+    <optional-fields>)
 
 ``<name>`` is a module name that contains the executable's main entry point.
 There can be additional modules in the current directory; you only need to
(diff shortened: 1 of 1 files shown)


doc/stanzas/executable.rst
```diff
@@ -286,3 +287,32 @@ well:
...
+However, using ``executables`` the executables defined in the stanza are
+allowed to share modules.
+
+Given modules ``Foo``, ``Bar`` and ``Baz`` the usage of ``executables`` can
+simplify the code:
+
+.. code:: dune
+
+          (executables
+            (names foo bar))
```
rgrinberg, line 300: the indentation seems a bit large here. Shouldn't it be aligned with `code::`?

Leonidas-from-XIV: I copied the indentation level from the code snippet in "Linking modes", but I agree it is a bit weird. What is the correct indentation? Then I can fix all occurrences in this file.

Alizter: It should line up with`code` which is 3 spaces indented. We could also set the default syntax highlighting for the entire file to be `dune` then we wouldn't need to indent by 3 but by 2.

Alizter: @emillon knows how to set the default indentation, I couldn't remember where I saw it though.

christinerose (approved):

doc/stanzas/executable.rst
Leonidas-from-XIV: Ok, I've adjusted all the blocks to align with `code`.

emillon: yes, in that case for now we align to 3. We can't set a default highlighting here since this file for now is included from another one.

Issue #4121, opened 2021-01-14
-I option not passed during development for headers installed by a sibling package from a monorepo


aantron:
## Expected Behavior

Luv vendors libuv, and installs libuv's header `uv.h` at `lib/luv/uv.h` using an [`install` stanza](https://github.com/aantron/luv/blob/44db25c2b2669f9d1373377006a4d5850e0820e3/src/c/dune#L15) (see also #3907 about this stanza).

When a project that depends on package `luv` is compiled, Dune correctly passes `-I path/to/switch/lib/luv` when compiling the project's C files, which allows them to pull in `uv.h` with `#include <uv.h>`.

I recently [added](https://github.com/aantron/luv/commit/4cdf5f0ac9b362d4017691e653e13799b6905bf7) package `luv_unix` to the Luv repo (in branch `unix`), which depends on `luv` and needs `uv.h`. After pinning everything, doing

```
opam install luv
opam install luv_unix
```

works, because the package installations are separate, and `uv.h` is in the switch by the time `luv_unix` is installed.

The expected behavior is that during development,

```
dune build
```

or some variation on it, should also work. `luv_unix` should be able to find `uv.h`.

## Actual Behavior

```
$ dune build
         gcc src/unix/luv_unix.o (exit 1)
luv_unix.c:9:10: fatal error: uv.h: No such file or directory
    9 | #include <uv.h>
      |          ^~~~~~
compilation terminated.
```

This can be seen immediately in [Travis logs](https://travis-ci.org/github/aantron/luv/jobs/754512714#L646). The command failing there is trying to build Luv's tester, which in branch `unix` [depends](https://github.com/aantron/luv/blob/4cdf5f0ac9b362d4017691e653e13799b6905bf7/test/dune#L3) on `luv_unix`.

## Reproduction

```sh
opam pin add luv https://github.com/aantron/luv.git#unix
opam pin add luv_unix https://github.com/aantron/luv.git#unix
opam install luv luv_unix
opam remove luv luv_unix
git clone https://github.com/aantron/luv.git
cd luv
make test   # or dune build -p luv,luv_unix, etc.
```

## Specifications

- Version of `dune` (output of `dune --version`): 2.8.1
- Version of `ocaml` (output of `ocamlc --version`): 4.11.1
- Operating system (distribution and version): Ubuntu 20.04 on WSL1

bobot:
I didn't get an error with dune `2.8.2`. Could you test this version?

```
[...]

PR #7512, opened 2023-04-07
fix: preserve directory hierarchy for installed headers

rgrinberg:
If we have foo/bar.h, we'll preserve the hierarchy relative to the
library directory


src/dune_rules/install_rules.ml
```diff
@@ -279,8 +279,21 @@ end = struct
...
+        let path base =
+          Path.Build.relative dir (base ^ Foreign_language.header_extension)
+        in
+        let project = Scope.project scope in
+        if Dune_project.dune_version project < (3, 8) then fun base ->
+          make_entry Lib (path base)
+        else fun base ->
+          let sub_dir =
+            let dirname = Filename.dirname base in
+            if dirname = "." then None else in_sub_dir (Some dirname)
```
anmonteiro, line 292:
```suggestion
            if dirname = Filename.current_dir_name then None else in_sub_dir (Some dirname)
```

anmonteiro (approved):

rgrinberg:
Let's have a look at @emillon as well

rgrinberg:
On second thought, this fix is a little bit weird. For example, how would we handle something like `(install_c_headers ../foo)`?

Probably simpler to introduce the syntax from the `install` stanza and allow users to select the installed path. E.g. `(install_c_headers (./foo as baz))`. 

To make the common case simple, we could introduce a new constructor such as `(install_c_headers (with_subdir foo/bar))` that will have the same behavior as this PR.

rgrinberg:
Okay, I just switched to using normal `(foo as bar)` patterns.

It requires manual work to get the install layout we want, but at least it's obvious and consistent with the rest of dune.

rgrinberg:
Actually, I'm now leaning towards introducing a new field now. One thing I dislike about this implementation is that it makes globbing awkard. E.g. one cannot do `(install_c_headers (glob_files *.h))` for seemingly no good reason.

The new field is just going to be `public_headers` and it will allow proper filenames. As a bonus, this will allow header extensions such as `.hh` and `.hpp` that should be usable as well.

rgrinberg:
@anmonteiro take another look please

rgrinberg:
@anmonteiro I rewrote this feature again. This time with the realization that this almost exactly the same as melange's runtime deps feature.

src/dune_rules/install_rules.ml
```diff
@@ -193,6 +192,8 @@ end = struct
           in
           make_entry ?sub_dir Lib path)
     in
+    let* melange_runtime_entries = additional_deps lib.melange_runtime_deps in
+    let* public_headers = additional_deps lib.public_headers in
```
anmonteiro, line 196: perhaps this can be done in parallel? `let* ... and* ... in`

anmonteiro (approved):
Looks great. It looks very similar to runtime_deps indede.

src/dune_rules/install_rules.ml
rgrinberg: Done. There's many opportunities to do such optimizations in the install rules.

Alizter: Is this really an optimisation? I thought and* is just implemented as a bind.

src/dune_rules/lib_flags.ml
```diff
@@ -152,7 +152,35 @@ module L = struct
...
+              (* TODO handle sandboxing *)
+              deps
+            in
+            (headers :: local, external_))
+    in
+    let local =
+      let open Action_builder.O in
+      let* bindings = Action_builder.of_memo @@ Memo.all_concurrently local in
+      let+ () = Action_builder.all_unit bindings in
+      Command.Args.S []
```
emillon, line 180:
```suggestion
      Command.Args.empty
```

src/dune_rules/dune_file.ml
```diff
@@ -599,6 +600,11 @@ module Library = struct
        and+ synopsis = field_o "synopsis" string
        and+ install_c_headers =
          field "install_c_headers" (repeat (located string)) ~default:[]
+       and+ public_headers =
+         field "public_headers"
```
emillon, line 604: Can you add some doc about this in `doc/stanzas/library.rst`?

emillon (approved):
Thanks. That's pretty much what I had in mind in an early implementation.

src/dune_rules/install_rules.ml
rgrinberg: Look at the definition of `and*` in fiber.

rgrinberg:
@anmonteiro i've now forbidden all non file deps in public_headers. A similar change to melange runtime deps could be done.

anmonteiro:
Thanks for the heads up, I’ll apply the fix there too.

aantron:
As I understand it, `public_headers` as implemented here is not sufficient for Luv.

I have a library in `src/c` that needs to install headers from `src/c/vendor/include/**` such that they are accessible inside a monorepo through `-I` flags passed by Dune to sibling packages that have C code that depends on the library in `src/c` without having to include the `vendor/include` path fragment inside `#include`. That is, I need either the headers to be copied to the install directory for the library in `src/c`, or `-I that-directory/vendor/include` to be passed to the C bindings in the sibling package. Doing `dune build` shows that these flags are not passed to the C code being compiled in the sibling library, and, predictably, that library cannot find the headers.

aantron:
More information here: https://github.com/ocaml/dune/issues/4121#issuecomment-1849417965

PR #7517, opened 2023-04-07
test: unintentionally visible headers

rgrinberg:
When we have two libraries in the same package, depending on the parent
library makes the headers of the sub libraries visible.


rgrinberg:
@emillon I wonder if we should move all headers files into a special directory to avoid accidents like this

PR #7520, opened 2023-04-07
refactor: move action encoders

rgrinberg:
Action.For_shell.t encoders are only used for printing rules in `$ dune
rules`. Therefore, we can just move them to the module that implements
rule printing.


Alizter:
I was using this when experimenting with turning actions in to Nix derivations, however I guess I should use something less fragile for that.

Alizter (approved):

Alizter:
You probably also want to move the tests for this somewhere else too.

rgrinberg:
ping @snowleopard

snowleopard (approved):
Looks good, thanks!

Issue #6945, opened 2023-01-27
`dune describe pp` doesn't work with staged pps


ejgallego:
`dune describe pp foo.ml` doesn't work when a staged pps is used. Example:

```ocaml
type t = { foo : string } [@@deriving ppx_yojson]
```

with dune file:
```lisp
 (preprocess (pps ppx_deriving_yojson))
```
works fine, but with dune file:
```lisp
 (preprocess (staged_pps ppx_deriving_yojson))
```
it produces the input file

tatchi:
> A first thing we can do is add an error message when we detect that case.

Happy to give it a try 😁

ejgallego:
#7167 just adds an error message, but still `dune describe pp` doesn't work for staged pps so this bug has not been fixed.

emillon:
I'd count that as a separate fix. we've already improved the situation, let's open a new issue for this.

PR #7522, opened 2023-04-08
feat(describe pp): add support for staged_pps

tatchi:
This PR adds support for `staged_pps` for the `describe pp` command.

This should be a better fix for https://github.com/ocaml/dune/issues/6945


bin/describe.ml
```diff
@@ -747,61 +752,133 @@ module Preprocess = struct
...
+                       External_lib_deps.Ml_sources.find_origin ocaml
+                         module_name
+                     with
+                     | Some
+                         (External_lib_deps.Ml_sources.Origin.Executables exes)
+                       -> Some (`Executables exes)
+                     | Some (Library lib) -> Some (`Library lib)
+                     | None | Some (Melange _) -> None
+                   in
+                   Memo.return stanza)
```
tatchi, line 817: This is extracted from https://github.com/ocaml/dune/blob/main/src/dune_rules/top_module.ml#L16. Should we try to extract the common part into its own function?

tatchi:
What's the easiest way to use a local build version of `dune` in another project? I would like to test this command with `ppx_import`.

ejgallego:
@tatchi I often do `dune exec -- bash` , tho some people do indeed prefer to have a testing switch and pinning dune in it.

bin/describe.ml
rgrinberg: I think you can just use it directly already. You can then use `Compilation_context.preprocessing` to get the preprocessing spec.

src/dune_rules/top_module.ml
```diff
@@ -13,7 +13,7 @@ let drop_rules f =
   in
   res
 
-let find_module sctx src =
+let find_lib_or_exe sctx src =
```
tatchi, line 16: Any better suggestion for this function's name?

bin/describe.ml
tatchi: I extracted the common part in [`a4972fa` (#7522)](https://github.com/ocaml/dune/pull/7522/commits/a4972fa59a7224e5139030abd9ec589553f59531). I'm not fan of the implementation, though (see inline comments).

tatchi:
@rgrinberg any idea how we could fix the `ppx_import` case?

tatchi:
I don't think it's blocking because this PR in its current state is already an improvement, but it would be nice if we could fix the `ppx_import` case

bin/describe.ml
```diff
@@ -747,61 +753,96 @@ module Preprocess = struct
...
-          User_error.raise ~loc
-            [ Pp.text "describe pp command doesn\'t work with staged_pps" ])
-      | false ->
-        User_error.raise
-          [ Pp.textf "%s does not exist" (Path.to_string file_in_build_dir) ])
+        | Some staged_pps ->
+          let (ppx_exe, flags), (_ : Dep.Fact.t Dep.Map.t) = staged_pps in
+          let* () =
+            Build_system.build_file
+              (Path.of_string (Path.Build.to_string ppx_exe))
```
rgrinberg, line 820: You can replace the string conversion with `Path.build ppx_exe`.

src/dune_rules/top_module.ml
```diff
@@ -31,37 +31,39 @@ let find_module sctx src =
...
-    let* ocaml = Dir_contents.ocaml dir_contents in
-    let stanza =
-      match Ml_sources.find_origin ocaml module_name with
-      | Some (Executables exes) -> Some (`Executables exes)
-      | Some (Library lib) -> Some (`Library lib)
-      | None | Some (Melange _) -> None
+    let+ ocaml = Dir_contents.ocaml dir_contents in
+    match Ml_sources.find_origin ocaml module_name with
+    | Some (Executables exes) ->
+      Some (`Executables exes, module_name, dir, dir_contents)
```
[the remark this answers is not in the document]
rgrinberg: A record would be a better return type I think.

bin/describe.ml
```diff
@@ -747,61 +753,96 @@ module Preprocess = struct
...
-      let* () = Build_system.build_file pp_file in
-      let+ project = Source_tree.root () >>| Source_tree.Dir.project in
-      Ok (project, pp_file)
+      let pp_file = Path.build pp_file in
+      let+ () = Build_system.build_file pp_file in
+      Ok (`Pp_file (project, pp_file))
     | false -> (
-      Build_system.file_exists file_in_build_dir >>= function
+      let* file_exists =
+        Build_system.file_exists (Path.build file_in_build_dir)
```
rgrinberg, line 770: Instead of doing this check, can we do `find_lib_or_exe`? Then we would know if we need this .pp extension based on the preprocessing spec.

rgrinberg:
> any idea how we could fix the ppx_import case?

Once you find the `Module.t`, you can do `Module.pp_flags` to get the build request to build everything you need for ppx_import.

tatchi:
> > any idea how we could fix the ppx_import case?
> 
> Once you find the `Module.t`, you can do `Module.pp_flags` to get the build request to build everything you need for ppx_import.

Looks like by using `Top_module.find_module` I can have access to the `Module.t`. But `Module.pp_flags` only return the flags for the `ppx` itself 

```
-ppx .ppx/9e9fc5e1ce36077a8fa10e14f57446ea/ppx.exe --as-ppx --cookie 'library-name="main_prefix"'
```

What I need is to create the `cm{i,o}` files of the dependent modules (as in my example above). I see [some code](https://github.com/ocaml/dune/blob/main/bin/ocaml/top.ml#L112) in `top.ml` that uses `Top_module.find_module` and then does things related to `cm{i,o}`. Maybe something here to try here ?

rgrinberg:
Note that we aren't using `pp_flags` to compute the flags. We are using it because it's supposed to collect the dependencies required to build the module. These dependencies were discovered from running `ocamldep` with the relevant `-ppx` flag. So my idea was to just include this action builder in your build query and discard the flags themselves.

tatchi:
> So my idea was to just include this action builder in your build query and discard the flags themselves.

I [added this action builder to the existing one](https://github.com/ocaml/dune/pull/7522/commits/6c20e16fdec3f2971c14684ae8287e06a531255d#diff-ac2cca256d4c1c7da04470409c832b2983e8ec431a149d4c68aebe5d8d6411e6R817) but it still doesn't work. Is this what you had in mind or did I misunderstand?

It would be nice to have a reproduction test for this. Would it be reasonable to use `ppx_import` directly in a test to avoid recreating part of it?

PR #7523, opened 2023-04-08
refactor(top_module): add module's name to error

tatchi:


src/dune_rules/top_module.ml
```diff
@@ -55,7 +55,11 @@ let find_module sctx src =
       let module_ =
         match Modules.find modules module_name with
         | Some m -> m
-        | None -> User_error.raise [ Pp.textf "module not found" ]
+        | None ->
+          User_error.raise
+            [ Pp.textf "%s: module not found"
```
rgrinberg, line 60: Could you include the path `src` in the error message instead? Also a message like:

```
textf "Could not find module corresponding to source file %s" (Path.to_string_maybe_quoted src)
```

would be an improvement

tatchi: Done :)

PR #7526, opened 2023-04-08
test: demonstrate private library not found in project with dune-project

anmonteiro:
I'm unsure whether this is an actual bug or expected behavior, but it showed up when trying to build a melange private library inside `node_modules`.

What the test reproduces:

- Adding a private library in a sub-directory with `dune-project`
- In the root project, try to depend on said private library
- show that dune can't find it


rgrinberg:
This is expected. Private libraries aren't visible outside the project they're defined in.

anmonteiro:
Yup, makes sense. We're fixing it over there.

PR #7527, opened 2023-04-09
fix: improve locations for [installed_c_headers]

rgrinberg:
Use exact locations for [installed_c_headers] entry rules

### DIFF
--- a/src/dune_rules/install_rules.ml
+++ b/src/dune_rules/install_rules.ml
@@ -110,7 +110,7 @@ end = struct
             | None -> subdir
             | Some lib_subdir -> Filename.concat lib_subdir subdir)
       in
-      fun section ?sub_dir ?dst fn ->
+      fun section ?(loc = loc) ?sub_dir ?dst fn ->
         let entry =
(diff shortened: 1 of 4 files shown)


anmonteiro (approved):

src/dune_rules/install_rules.ml
```diff
@@ -268,8 +268,9 @@ end = struct
     in
     let+ execs = lib_ppxs ctx ~scope ~lib in
     let install_c_headers =
-      List.map lib.install_c_headers ~f:(fun base ->
-          Path.Build.relative dir (base ^ Foreign_language.header_extension))
+      List.rev_map lib.install_c_headers ~f:(fun (loc, base) ->
```
Alizter, line 271: Since you've changed this from a map to a rev_map this won't affect anything right?

Alizter (approved):

src/dune_rules/install_rules.ml
rgrinberg: makes no difference here

PR #7535, opened 2023-04-11
refactor: Make [Either.t] equal [Stdlib.Either.t]

rgrinberg:


rgrinberg:
@Leonidas-from-XIV you can also use this trick in opam to get rid of the cppo

Leonidas-from-XIV:
@rgrinberg  Interesting trick! This has led me down to this PR https://github.com/ocaml/opam/pull/5508 which simplifies that even further. But it could be potentially useful for Extlib as well.

PR #7480, opened 2023-04-03
Add "dune describe package-entries"

moyodiallo:
The need is about knowing if a private executable is going to be installed before building the project.

This PR is related to https://github.com/ocurrent/opam-dune-lint/pull/46:
The command `dune describe external-lib` does not give the package name of private executable if it going to be installed, `dune describe external-lib-deps` is used by `opam-dune-lint` to extract the external dependencies.

Issue #4347, opened 2021-03-12
Byte target can't be debugged with ocamldebug


disposedtrolley:
## Expected Behavior

`dune build` should produce a byte target for https://github.com/disposedtrolley/hello_ocaml which is able to be debugged with `ocamldebug`. A breakpoint should be able to be set on line 2 of the `helloworld` module.

## Actual Behavior

While `dune build` is able to build a `.bc` file, it seems to be missing debug symbols. For example, attempting to set a breakpoint on line 2 of `helloworld.ml` results in:

```
hello_ocaml · (main) ⟩ ocamldebug _build/default/helloworld.bc                         ~/p/o/hello_ocaml 2
        OCaml Debugger version 4.12.0

(ocd) break @helloworld 2
Loading program... done.
Can't find any event there.
(ocd)
```

When the project is built with `ocamlc -g helloworld.ml`, the resulting `a.out` file can be debugged with `ocamldebug` correctly.

## Reproduction

- PR with a reproducing test: 

1. Clone https://github.com/disposedtrolley/hello_ocaml
2. Run `dune build`
3. Run `ocamldebug _build/default/helloworld.bc`
4. Type `break @helloworld 2`
5. Observe that the debugger outputs "Can't find any event there."

## Specifications

- Version of `dune` (output of `dune --version`): 2.8.4
- Version of `ocaml` (output of `ocamlc --version`) 4.12.0
- Operating system (distribution and version): Fedora 33 (x86), macOS 11.2 (arm64)

## Additional information 

I've tried this with OCaml 4.12 (on macOS with Apple Silicon) and 4.10.2 (on macOS with Apple Silicon and Fedora 33 x86).

jeremiedimino:
Thanks for the detailed report. I had a look and found the issue; it is because of the name mangling Dune does to avoid collision between modules of different libraries and/or modules of the executable.

[...]

PR #7540, opened 2023-04-13
Support for BUILD_PATH_PREFIX_MAP

richardlford:
Modify Dune to produce `BUILD_PATH_PREFIX_MAP` mappings that map paths to package built artifacts to abstract paths that mirror the installed hierarchy. In particular, for example, if file

```
src_root/_build/default/somepaths/somefile
```

is a part of package `pkg` and will be installed in a section `sec` (e.g. `lib` or `doc`) at location

```
$prefix/sec/pkg/anotherpath/renamed_somefile
```

then the Dune `BUILD_PATH_PREFIX_MAP` will map that file to abstract path

```
/install_root/sec/pkg/anotherpath/renamed_somefile
```

where in this case, I have chosen `/install_root` as the literal prefix for mappings of things that are part of a package.

If a built artifact is not part of a package, then the mapping remains the same, namely the above file, if not part of a package, will map to

```
/workspace_root/somepaths/somefile
```

This is implemented by first using Dune's install logic to get a mapping of all installed files from their location in the build directory to the install location. Then that mapping is abbreviated, taking advantage of the fact that for many artifacts we can do a directory-level mapping rather than mapping each individual file. When the basename of a file changes when installed, the file needs its own mapping.

In addition, the `dune describe` command was expanded. 

The `dune describe workspace` command now includes package information.

There is a new `dune describe map` command that outputs information about the `BUILD_PATH_PREFIX_MAP` mappings, both full and abbreviated.


richardlford:
This is still a draft, as it still has some debugging code in it, and output that is not sanitized. Also, I'm still adding tests.
However, @rgrinberg, could you please take a look to see if I'm on the right track? Thanks.

richardlford:
I have a separate gitlab PR, https://gitlab.com/gasche/build_path_prefix_map/-/merge_requests/2#7facec255884dc55d3818e4dfbea89b1bdd7f49e, for the vendored `build_path_prefix_map` changes. This brings over changes that were merged into the ocaml trunk, with some small additions.

richardlford:
@rgrinberg, I've converted this from draft to ready for review, not because it is perfect, but because I need your help with some  memo/fiber issues. Also, for the changes to the vendored build_path_prefix_map code, I have a separate PR out to @gasche, so I realize that those changes would not be checked directly into Dune. Hopefully, that PR will be accepted and merged before this PR is ready.

I took your suggestion of looking at the `dune describe`  command and found it a handy vehicle for testing my mappings. For that purpose, I added a `dune describe map` command to call the functions to compute the forward and inverse mappings, and also to test them for consistency. 

The strategy for computing the forward mappings was to start with a map of every single file that will be installed, mapping from its relative location in the build directory to its abstract installation path. That map is then abbreviated by doing directory-level mapping when possible.

For the inverse map, we first split out the mappings of installed files that exist in the source directory (i.e. are not generated). We do that because, for the inverse mappings, we want to give precedence to the location in the source directory.

One complication that arose was related to sandboxing. Because of it, we could not use the pwd in computing the build directory paths. Instead, we had to take the root that is available in `action_exec`. That means that when doing a mapping from `action_exec`, a final adjustment to the map is needed.

All is working well as far as the mapping goes, but there were some technical challenges, and I'm not sure my solution was the correct or best way. In particular, I'm running into an assertion failure which I'll mention below.

But first, let me mention the first challenge. The existing code that does mapping is in the `dune_engine/action_exec.ml` file, in the `exec` function. It was calling code in the Build_path_prefix_map0 module to encode the desired map. But previously there was only one mapping pair, from the build root to `/workspace_root`. But now there is a more complex map that must be computed, and the information needed to compute it is in the `dune_rules/install_rules` files. But those are not accessible to `dune_engine`. My solution was to call the functions that compute the maps from `dune_rules/main.ml` in the `get` functions that returns the `Dune_rules.build_system`. At that point, I compute the maps, but store them in some references that are defined in `dune_util/build_path_prefix_map0.ml`. There they are handily available when needed.

The memo/fiber problem I'm having is showing up when trying to build the `lwt` library. The `assert false` in file `dune_engine/build_system.ml`, in function `update_build_progress_exn`, line 187 if failing. That function is only supposed to be called when the `State` is `Building`, but by adding some print statements, I was able to see that the `State` was `Initializing`. So it appears that reset_progress was never called. However, if I delete the calls to compute my maps, the problem goes away. I've spent quite a bit of time studying the Fiber and Memo code, but I think there is still a lot that I don't understand. Perhaps the solution to the problem would be apparent to you.

This could logically be divided into two PRs, one for the forward mapping, and another for the inverse (i.e. the `dune debug` command). But it made sense to code the algorithms for the forward and inverse mappings in a coordinated way, to insure they were consistent.

I'll keep doing more testing and debugging, but if you could take a look and help me with the memo/fiber problem, I'd much appreciate it.

Thanks

christinerose (approved):

src/dune_lang/pform.mli
```diff
@@ -20,6 +20,7 @@ module Var : sig
     | Ocaml_version
     | Ocaml
     | Ocamlc
+    | Ocamldebug
```
rgrinberg, line 23: This could be an independent change.

vendor/build_path_prefix_map/src/build_path_prefix_map.mli
```diff
@@ -15,10 +41,29 @@ type map = pair option list
...
-val rewrite_opt : map -> path -> path option
-(** [rewrite_opt map path] tries to find a source in [map]
+val rewrite_first : map -> path -> path option
+(** [rewrite_first map path] tries to find a source in [map]
     that is a prefix of the input [path]. If it succeeds,
     it replaces this prefix with the corresponding target.
     If it fails, it just returns [None]. *)
 
+val rewrite_all : map -> path -> path list
+(** [rewrite_all map path] finds all sources in [map]
```
rgrinberg, line 51: Are you pulling these changes from upstream? If so, this should be done separately.

bin/debug.ml
```diff
@@ -0,0 +1,319 @@
+open Stdune
+open Import
+
+let doc =
+  "Debug a bytecode program in a similar environment as if installation was \
```
rgrinberg, line 5: can you move this to `dune ocaml debug`?

src/dune_engine/action_to_sh.ml
```diff
@@ -90,7 +90,9 @@ let simplify act =
            (String.quote_for_shell target))
       :: acc
     | Pipe (outputs, l) -> Pipe (List.map ~f:block l, outputs) :: acc
-    | Extension _ -> Sh "# extensions are not supported" :: acc
+    | Extension ext ->
+      let ext_str = Dune_lang.to_string ext in
+      Sh ("# extensions are not supported=" ^ ext_str) :: acc
```
rgrinberg, line 95: Separate PR for this please.

rgrinberg:
> The memo/fiber problem I'm having is showing up when trying to build the lwt library. The assert false in file dune_engine/build_system.ml, in function update_build_progress_exn, line 187 is failing. That function is only supposed to be called when the State is Building, but by adding some print statements, I was able to see that the State was Initializing. So it appears that reset_progress was never called. However, if I delete the calls to compute my maps, the problem goes away. I've spent quite a bit of time studying the Fiber and Memo code, but I think there is still a lot that I don't understand. Perhaps the solution to the problem would be apparent to you.

Probably because of the mutation you've introduced. All memo computation must not have effects. We need to think of a different way to thread this map.

src/dune_lang/pform.mli
richardlford: Yes, I will split this out to go with the "dune debug" PR.

vendor/build_path_prefix_map/src/build_path_prefix_map.mli
richardlford: As I mentioned, I have a PR to @gasche, to get this change into his upstream repo. I've not gotten any response from him yet on that PR, however, he was the one that merged that change into the OCaml trunk. I suppose if you wanted to merge this into Dune separately you could, and then when it gets upstream you could go back to getting it there.

bin/debug.ml
richardlford: Yes, this will be split into a separate PR. However, I'd like to keep the computation of the forward and inverse maps in a single PR so we can do consistency checking of them.

src/dune_engine/action_to_sh.ml
richardlford: Will do.

richardlford:
> Probably because of the mutation you've introduced. All memo computation must not have effects. We need to think of a different way to thread this map.

I'm open to your suggestions on that. I'm not sure what you mean by effects. All my mapping computations are functions of the existing 

```
module Stanzas_to_entries : sig
  val stanzas_to_entries :
    Super_context.t -> Install.Entry.Sourced.t list Package.Name.Map.t Memo.t
end
```

They would be pure if not for the fact that stanzas_to_entries is a Memo result.

Alizter:
A few comments:
- The changes to describe you are proposing are similar to https://github.com/ocaml/dune/pull/7480/files. You should probably split it off from this PR too.
- I'm inclined to think that the command should be in the `ocaml` command group, so it would be `dune ocaml debug`. We don't have any other debuggers planned to be added (not even for Coq) so it might seem a little strange. My motivation is that it is a really ocaml-centric feature so should be in that command group.

I'm still aware that this PR is in draft mode, so I won't do many comments on the code itself for now since it might change.

richardlford:
> * The changes to describe you are proposing are similar to [Add "dune describe package-entries" #7480](https://github.com/ocaml/dune/pull/7480). You should probably split it off from this PR too.

@Alizter, Thanks for pointing that out. I will split out my "dune describe" changes to show packages. My "dune describe maps" is different and I will keep it, as it is also helpful in validating the maps.

> * I'm inclined to think that the command should be in the `ocaml` command group, so it would be `dune ocaml debug`. We don't have any other debuggers planned to be added (not even for Coq) so it might seem a little strange. My motivation is that it is a really ocaml-centric feature so should be in that command group.

OK. Or how would you feel about "dune ocamldebug"? That would be a little more concise (though only by one space, so perhaps your suggestion is better). I'll also be splitting this part into a separate PR.

Alizter:
Since command groups have a tree structure, `dune ocaml debug` is probably better.

vendor/build_path_prefix_map/src/build_path_prefix_map.mli
richardlford: I've not heard from @gasche on the build_path_prefix_map changes, so we may need to merge this as part of the PR and later vendor it when it is ready. Relative to my build_path_prefix_map I have one change, a local copy of List.find_map, in order to be able to build Dune with versions of OCaml that did not yet have List.find_map.

test/blackbox-tests/test-cases/github1946.t/run.t
```diff
@@ -7,7 +7,7 @@ in the same dune file, but require different ppx specifications
   $ dune build @all --profile release
   $ dune ocaml merlin dump-config $PWD
   Usesppx1
-  ((STDLIB /OCAMLC_WHERE)
+  ((STDLIB /install_root/lib/ocaml)
```
Alizter, line 10: It seems the behaviour of BUILD_PATH_PREFIX_MAP is broken. The setting of the env var above should make this get substituted for OCAMLC_WHERE.

richardlford:
I've converted this back to "ready to review" again. I've removed all of the code that will be split out into a separate PR, and in particular, the code for the "dune ocaml debug" command (and also some debugging code I had added). 

I have some questions for the reviewers that will guide any final changes (in addition to the comments reviewers will have):

- The goal of the current PR is that the rewritten build-time paths for files that will be installed will have the form `/install_root/<package>/<package-relative-path>`. Sometimes this mapping can be achieved at the directory level, but cases where it cannot include:
  - When the basename of the file installed is different than the file in the build directory.
  - When files in a given build directory have multiple destination directories.
- As a result of the preceding, for a large project, the mapping can be large, and in some cases we are running into a problem with exceeding the space allocated for arguments and environment variables (when doing `execve`).

Possible solutions to the above include:
- Only guarantee abstract paths that mirror the install path for selected paths. The PR currently does this by selecting only OCaml source files (`*.ml). But since Dune is supposed to be a general purpose build system, that does not seem satisfactory (though it could perhaps be OK for this PR).
- Like the preceding, but allow the user to select the filter in a stanza. Probably most users would not run into the limitation so we could only filter if requested by the user.
- Currently, if the project has multiple packages, we make a mapping that includes all of the files in all of the packages. But would it work to have a separate mapping for each package? Then when building something that is part of a package, we would use the mapping for that package. 
  - That would reduce the size of the mappings. Then we would only run into size problems when an individual package was very large.
  - Is it true that build artifacts for a given package do not reference files in a different package? If not, then perhaps we need to keep having a single mapping for all packages.
  - If it is OK to have each action use the mapping for its containing package, we need a way to know for which package a given action is being performed. I don't currently know how to do that, though if we knew the input or output files for an action we could perhaps use some maps from files to their package (we would need to make these, but we have that information).
- We could extend the mapping mechanism to allow for storing maps in a file. For example, if the value of BUILD_PATH_PREFIX_MAP started with `@`, we could interpret the rest as a path to a file containing the mapping. In order to allow user files that have a `@`, we would need to introduce an encoding for `@` similar to the way `:`, `=`, and `%` are currently encoded.

Note that the computations of the maps is now being invoked in `bin/build_cmd.ml/run_build_system` in these lines;

```ocaml
              Install_rules.Build_map.build_and_save_maps scontexts
```

This is earlier than install information would typically have been computed, but we need to have the mappings available during the build steps. This seems to be having the effect that some messages that previously were being emitted are skipped (though I do not understand exactly why--yet). 

That's all for now. I look forward to some feedback from reviewers on the issues above.

test/blackbox-tests/test-cases/github1946.t/run.t
richardlford: Actually not. The setting of the map in the Dune supersedes the existing setting. For paths not covered by the Dune mappings, the previous settings would take effect.

One difference between the current Dune mappings and the previous settings is that before, only paths under the PWD were mapped. Now paths under the opam root are also mapped (to /install_root/...). That is why the prior mappings are not effective. Perhaps these tests should have these mappings removed since they are no longer needed?

src/dune_rules/install_rules.mli
```diff
@@ -9,3 +20,53 @@ val symlink_rules :
...
+       Super_context.t Context_name.Map.t
+    -> Install.Entry.Sourced.t list Package.Name.Map.t Context_name.Map.t
+    -> Build_path_prefix_map.map String.Map.t
+
+  val build_context_map :
+       Super_context.t
+    -> Install.Entry.Sourced.t list Package.Name.Map.t
+    -> Build_path_prefix_map.map
+
+  val build_all_maps :
```
rgrinberg, line 56: Where is this function being used outside of this module? If it isn't, it shouldn't be in the mli.

src/dune_rules/install_rules.ml
```diff
@@ -499,6 +499,21 @@ end = struct
...
+      Install.Entry.Sourced.t list Package.Name.Map.t Context_name.Map.t Memo.t
+      =
+    let open Memo.O in
+    let ctxs = Context_name.Map.to_list scontexts in
+    let* pairs =
+      Memo.sequential_map ctxs ~f:(fun (ctxn, sctx) ->
+          let* packages = Stanzas_to_entries.stanzas_to_entries sctx in
+          Memo.return (ctxn, packages))
+    in
+    let maps = Context_name.Map.of_list_reduce pairs ~f:(fun old _new -> old) in
```
rgrinberg, line 513: The reduce doesn't seem right. How could you have a collision on a context name here? Shouldn't it be `Map.of_list_exn`?

rgrinberg:
> Currently, if the project has multiple packages, we make a mapping that includes all of the files in all of the packages. But would it work to have a separate mapping for each package? Then when building something that is part of a package, we would use the mapping for that package.

If this possible, I think it would be preferable. Although it does seem a bit more difficult. So what you have currently is fine too.

Note that the references introduced in `Dune_util.Build_path_prefix_map0` are a no-go. I would suggest to focus on getting rid of them. For example, by extending `Execution_parameters` to allow the necessary information to be passed without mutation.

Also, could you move as much of the logic as possible outside of install_rules and into its own module?

Those changes in Dune_util.Build_path_prefix_map probably don't belong there either. I understand you wanted a change a single code path that will work for both dune_engine and dune_rules, but this is all quite specialized to our rules. So it should live in dune_rules with sensible hooks in dune_engine.

bin/describe.ml
```diff
@@ -817,6 +1001,7 @@ module What = struct
     | External_lib_deps
     | Opam_files
     | Pp of string
+    | Map
```
rgrinberg, line 1004: "map" is a rather general name for a command that does something quite specific. How about "build-path-prefix-map"?

src/dune_rules/install_rules.ml
richardlford: I believe that of_list_reduce works correctly, but of_list_exn is more naturally the right function to use, so I will change to use it.

bin/describe.ml
richardlford: Done (in next push)

richardlford:
> Note that the references introduced in `Dune_util.Build_path_prefix_map0` are a no-go. I would suggest to focus on getting rid of them. For example, by extending `Execution_parameters` to allow the necessary information to be passed without mutation.

@rgrinberg, I'm having a difficult time finding an alternative that does not require mutation. I spent all last week trying to come up with a solution.

 Currently, the default execution parameters are set in bin/common.ml, function `init`. This function also calls `Dune_rules.Main.init` which sets the `Build_config` execution_parameters field to a function that computes the execution parameters for each directory (updating them using the information in the `dune-project` file). 

Computation of the mappings requires reading all the `dune` and `dune-project` files (as well as `opam` files), (such as is done by `Dune_load.load`, as for example in `Dune_rules.Main.get ()`). I experimented with inserting the following code at the end of `Common.init`:

```
  let build_sys =
    (* Here we make the assumption that this computation doesn't yield. *)
    Fiber.run
      (Memo.run (Dune_rules.Main.get ()))
      ~iter:(fun () -> 
        Dune_util.Log.info [Pp.text "common.ml:init, Fiber.run failed"];
        assert false)
  in
  ignore build_sys;
```

This builds, but when I run the tests I get a lot of crashes, though I don't see my message in the log. So it seems that computation of the maps must be done later, but since `Common.init` determines the `Execution_parameters`, I don't see how I can get the maps into the `Execution_parameters` without mutation.

I'm considering abandoning these changes and instead adding Dune documentation that if a package wants its code to be debuggable after installation, then it needs to arrange its source hierarchy to mirror the installation-time hierarchy.

Do you have any other suggestions?

richardlford:
I'm pushed changes to reflect some of the review comments, but have not yet solved the issue of mutation (as mentioned above).

richardlford:
I'm changing jobs and will not have time to complete this PR. It has proved more complex than expected. If someone else thinks this is worthwhile, they are welcome to continue this effort. As a simpler alternative, I have https://github.com/ocaml/dune/pull/7741. It fixes the gap in reproducibility, but the abstract paths will only work for installed libraries if the library authors choose a source layout that mirrors the installation layout.

rgrinberg:
I'm going to close this PR until there's some more demand for it. AFAIK, this would allow programs like ocamldebug to work on installed artifacts in principle. Is ocamldebug able to reverse the BUILD_PATH_PREFIX_MAP at the moment though? Can any other tool make use of this data? If not, I think it's too early for dune to add support for this.

gasche:
The nice compiler-side work of @richardlford has been taken over by @pirbo in recent months and mostly merged so it, yes, should now be in a state where Dune can make use of it. This is something that I have discussed by email with @pirbo and @Alizter. Here is what @pirbo wrote about this in said email thread:

> Here is the summary of my understanding/thoughts: I/you can make it an issue if you think that the relevant move or tell me what else to do:
> 
> - I see as a low hanging fruit the fact to slightly improve the experience of debugging a bytecode program built by dune in a terminal. I'm neither talking about native code debugging nor editor integration (at first). The minimal viable product I envision (and looks to be emerging from the comments in the issues) would be: a user would type in a terminal within a dune project `dune ocaml debug somefolder/mysuperapp.exe [args]`
> 
>   * dune would compile `somefolder/mysuperapp.exe`**in bytecode** even if `(mode bytes)` is not part of the options put in the the dune file for the executable mysuperapp
> 
>   * dune would compile the cma of all the internal libraries mysuperapp depends upon too
> 
>   * dune would launch `ocamldebug $(for a_lib in [external libraries listed as dependencies of mysuperapp] ; do $(dune_equivalent_of ocamlfind query -r -i-format a_lib)) somefolder/mysuperapp.exe [args]` with the environment variable BUILD_PATH_PREFIX_MAP set to whatever it was set end compiling the executable (and the libraries)
> 
> - No care for printers at first (but obviously this is the first follow up)
> 
> (No readliner either (so the user would better call one externally as it does for ocamldebug right now))
> 
> As far as I understand, https://github.com/ocaml/dune/pull/7750 might be a good starting point but with the difference that:
> 
> - The part about DEPLOY_PATH_PREFIX_MAP is irrelevant : 
> 
>  * what has been merged in ocaml/ocaml#14055 is that ocamldebug (version 5.5.0 and up aka current trunk only) will read BUILD_PATH_PREFIX_MAP in reverse; that should handle how ocamldebug will find sources of modules internal to the projects
> 
>  * the way ocamldebug will find sources of external modules is thanks to the `-I` added by the modern equivalent of the `ocamlfind query -i-format` (I can give a more detailed explanation but given that tree structure of installed files does not follow the one recommended by dune for the sources, there is no hope in finding files by inverting `/worktree_root/` in locations of installed files)
> 
> - I find key the aspect that when you debug """best = bytes""" for executable aka dune should know that it knows how to compile a bytecode executable even if it is not listed as a mode for the executable.
> 
> 
> Unless I'm forgetting something obvious, Once that is implemented, users should get back their setting of the pre dune ages (aka ocamldebug should find types and sources) EXCEPT FOR the possibility to set a breakpoint by giving a (non mangled) module name (aka https://github.com/ocaml/dune/issues/4347), users will still have to speak in term of source file name (or know the mangled name of things).

rgrinberg:
Thanks for the update @gasche. In that case, there are no objections to this work in principle, but it still needs some work to get it to a merge-able state. Once we have somebody willing to do that work, we can either re-open this PR or they can create a new one.

Issue #7543, opened 2023-04-13
Can't resolve `ppxlib.traverse` when ppxlib is vendored


gridbugs:
## Expected Behavior

It should be possible to build a project which depends on the `ppxlib.traverse` library when the `ppxlib` package is vendored.

## Actual Behavior

Compilation fails with the error:
```
  File "vendor/ppxlib.0.29.1/traverse/dune", line 9, characters 7-22:
  9 |   (pps ppxlib_metaquot)))
             ^^^^^^^^^^^^^^^
  Error: Library "ppxlib_metaquot" not found.
```
`ppxlib_metaquot` is part of the `ppxlib` package.

## Reproduction

- PR with a reproducing test: https://github.com/ocaml/dune/pull/7544

## Specifications

- Version of `dune` (output of `dune --version`): unreleased
- Version of `ocaml` (output of `ocamlc --version`): 4.14.1
- Operating system (distribution and version): NixOS

## Additional information 

This bug was introduced in https://github.com/ocaml/dune/commit/16a7e883ef2b2283f8a636fbf2c8bade3da92754. Relevant PR is https://github.com/ocaml/dune/pull/7415.

- Link to gist with verbose output (run `dune` with the `--verbose` flag):

anmonteiro:
fixed in https://github.com/ocaml/dune/pull/7553

PR #7544, opened 2023-04-13
Add repro for github issue 7543

gridbugs:
Reproduces https://github.com/ocaml/dune/issues/7543


anmonteiro:
fixed in https://github.com/ocaml/dune/pull/7553

PR #7547, opened 2023-04-13
fix: correctly generate .install files

rgrinberg:
Previously, the contents of the .install file would depend on whether
the .opam install file would be present in the source.

This check is wrong whenever we generate the .opam file ourselves. This
commit checks if we are generating the .opam file.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -101,6 +101,10 @@ Unreleased
 - Allow `(package ...)` in any position within `(rule ...)` stanza (#7445,
   @Leonidas-from-XIV)
 
+- Always include `opam` files in the generated `.install` file. Previously, it
+  would not be included whenever `(generate_opam_files true)` was set and the
+  `.install` file wasn't yet generated. (#7547, @rgrinberg)
+
 3.7.1 (2023-04-04)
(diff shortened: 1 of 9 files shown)


CHANGES.md
```diff
@@ -101,6 +101,10 @@ Unreleased
 - Allow `(package ...)` in any position within `(rule ...)` stanza (#7445,
   @Leonidas-from-XIV)
 
+- Always include `opam` files in the generated in `.install` file. Previously, it
```
anmonteiro, line 104:
```suggestion
- Always include `opam` files in the generated `.install` file. Previously, it
```

CHANGES.md
```diff
@@ -101,6 +101,10 @@ Unreleased
 - Allow `(package ...)` in any position within `(rule ...)` stanza (#7445,
   @Leonidas-from-XIV)
 
+- Always include `opam` files in the generated `.install` file. Previously, it
+  would be not be included whenever `(generate_opam_files true)` was set and
```
Alizter, line 105:
```suggestion
  would not be included whenever `(generate_opam_files true)` was set and
```

Issue #7413, opened 2023-03-26
Debug mapping is sometimes inconsistent with installed locations


richardlford:
## Expected Behavior
Consider a simple project with one library and a single package, `mypackage`.

Here is the source layout:

- dune-project
- lib/
  - dune
  - lib1.ml

Here is an excerpt from the dune-project file:

```
(lang dune 3.7)
(name mypackage)
(package
 (name mypackage))
```

And here is the "lib/dune" file:

```
(library
 (name lib1)
 (public_name mypackage.lib1)
 (modes byte)
 )
```
I did a `dune install --prefix _opamroot` and here is an excerpt of the layout it produced:

```
_opamroot/
_opamroot/lib
_opamroot/lib/mypackage
_opamroot/lib/mypackage/lib1
_opamroot/lib/mypackage/lib1/lib1.ml
_opamroot/lib/mypackage/lib1/lib1.cma
```

Now, when the debugger is debugging a program that used this library, we would want the debug information from the library to be:

`/workpace_root/mypackage/lib1`

so that after the debugger does a mapping from `/workspace_root` to `opamroot/lib`, that the directory will become

`opamroot/lib/mypackage/lib1`

which is where the source code and '.cma' file are located.

## Actual Behavior

The following debug directories were produced in the 'lib1.cma':

- /workspace_root
- /workspace_root/lib
- /workspace_root/lib/.lib1.objs/byte

The problem is that the debug information is missing information about which package contains the library. When building a module that is part of a package that wil be potentially installed, the mapping should include the package name, and also be as if the library location was at the root of the package. So the `BUILD_PATH_PREFIX_MAP` in this case should be:

BUILD_PATH_PREFIX_MAP="/workspace_root/mypackage=$cwd/lib"

## Reproduction

I will open a PR soon, but I think the above describes the issue adequately. Also, I plan to work on a fix for this Dune problem, and am also working on a Dune enhancement to start up `ocamldebug` with a `BUILD_PATH_PREFIX_MAP` to maps from `/workspace_root` to the actual source locations.

## Specifications

- Version of `dune` : 3.7.0
- Version of `ocaml` : 5.0.0
- Operating system: Ubuntu 22.04

rgrinberg:
> BUILD_PATH_PREFIX_MAP="/workspace_root/mypackage=$cwd/lib"

`BUILD_PATH_PREFIX_MAP` is a documented standard, I don't think we can just modify its meaning and syntax like that. Can you introduce another variable instead?

[...]

PR #7548, opened 2023-04-13
Testcase for issue #7413

richardlford:
This is a test showing the problem in issue #7413. 

The test has currently been promoted to match the existing behavior, in order not to have a failing test. When the issue is resolved the result of the grep should give a count of 1.

### DIFF
--- a/test/blackbox-tests/test-cases/rewrite-tests/library-install-paths.t
+++ b/test/blackbox-tests/test-cases/rewrite-tests/library-install-paths.t
@@ -1,0 +1,103 @@
+Test the rewriting of a minimal library, and show that it does not conform to
+the locations where the artifacts are installed.
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.7)
+  > (name minlib)
+  > (package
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/rewrite-tests/minlib.t
```diff
@@ -0,0 +1,62 @@
...
+  $ mkdir lib
+
+  $ cat > lib/dune <<EOF
+  > (library
+  >  (name minlib)
+  >  (public_name minlib)
+  >  )
+  > (rule
+  >  (target x)
+  >  (action (system "dune_cmd rewrite-path $PWD/_build/default/lib/run.ml | grep -c '^/install_root/lib/minlib/run.ml$'; touch x;")))
```
rgrinberg, line 22: This test is missing an eof?

richardlford: Interesting. Apparently CRAM supplies a missing EOF. I've added the missing EOF and rebased.

PR #7552, opened 2023-04-14
Add watch-mode benchmarks of dune monorepo

gridbugs:
This updates the monorepo benchmark to invoke the external benchmark runner [here](https://github.com/ocaml-dune/ocaml-monorepo-benchmark/tree/main/dune-benchmark-runner). It works by adding a new trace-file entry for builds in eager watch mode and adding an RPC which returns the number of builds that have completed since the RPC server started which is used by the benchmark runner to run dune in watch mode and sequentially change files and wait for builds to complete. See the commit descriptions for details about each part.

The scenarios implemented by the runner and sample output:
```
{
  "results": [
    {
      "name": "dune monorepo benchmarks",
      "metrics": [
        {
          "name": "build from scratch",
          "value": 616.247878074646,
          "units": "sec"
        },
        { "name": "null build", "value": 32.72038006782532, "units": "sec" },
        {
          "name": "watch mode: initial build",
          "value": 34.548771,
          "units": "sec"
        },
        {
          "name": "watch mode: changing file in 'base' library",
          "value": 53.681128,
          "units": "sec"
        },
        {
          "name": "watch mode: introducing error in file in 'base' library",
          "value": 29.321947,
          "units": "sec"
        },
        {
          "name": "watch mode: fixing error in file in 'base' library",
          "value": 54.447027,
          "units": "sec"
        },
        {
          "name": "watch mode: changing file in 'file_path' library",
          "value": 37.868993,
          "units": "sec"
        },
        {
          "name": "watch mode: introducing error in file in 'file_path' library",
          "value": 33.622908,
          "units": "sec"
        },
        {
          "name": "watch mode: fixing error in file in 'file_path' library",
          "value": 37.516952,
          "units": "sec"
        }
      ]
    }
  ]
}
```

Note that when running the benchmarks from the `monorepo-benchmarks` branch you'll currently get this error:
```
File "duniverse/ppxlib/traverse/dune", line 9, characters 7-22:
9 |   (pps ppxlib_metaquot)))
           ^^^^^^^^^^^^^^^
Error: Library "ppxlib_metaquot" not found.
```

This is a known issue: https://github.com/ocaml/dune/issues/7543. The temporary workaround is to revert commit 16a7e883ef2b2283f8a636fbf2c8bade3da92754.

### DIFF
--- a/bench/monorepo/Makefile
+++ b/bench/monorepo/Makefile
@@ -1,12 +1,11 @@
-# Intended to be used from inside docker containers built from monorepo-bench.Dockerfile
-RUNNER = _build/default/bench.exe
-DUNE_TO_BENCHMARK = /home/user/dune/_build/default/bin/main.exe
+# Intended to be used from inside docker containers built from bench.Dockerfile
+RUNNER = /home/user/monorepo-benchmark/dune-benchmark-runner/_build/default/src/main.exe
+DUNE_EXE_PATH = /home/user/dune/_build/default/bin/main.exe
+BUILD_TARGET = ./monorepo_bench.exe
+MONOREPO_PATH = /home/user/monorepo-benchmark/benchmark
 
-$(RUNNER): dune bench.ml
-	dune build $@ --release
-
-bench: $(RUNNER)
-	$< $(DUNE_TO_BENCHMARK) build -j auto
+bench:
+	$(RUNNER) --dune-exe-path=$(DUNE_EXE_PATH) --build-target=$(BUILD_TARGET) --monorepo-path=$(MONOREPO_PATH) --print-dune-output
 
 clean:
 	dune clean
--- a/bench/monorepo/README.md
+++ b/bench/monorepo/README.md
@@ -7,7 +7,13 @@ to be consumed by current-bench.
 
 The monorepo will be set up with opam-monorepo using the lockfile
 [here](https://github.com/ocaml-dune/ocaml-monorepo-benchmark/blob/main/benchmark/monorepo-bench.opam.locked)
-which is downloaded during `docker build`. Also during `docker build`,
-a a library is created called `monorepo` with
+which is downloaded during `docker build`. Also during `docker build`, an
+executable dune project is created called `monorepo_bench` with
 [this](https://github.com/ocaml-dune/ocaml-monorepo-benchmark/blob/main/benchmark/dune)
 dune file listing all the libraries to include in the build.
+
+The benchmark runner is a program which invokes dune in a number of
+benchmarking scenarios and prints out the results in a format understood by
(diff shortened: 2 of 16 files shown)


gridbugs:
Note that in this PR the benchmark runner interacts with watch-mode dune by shelling out to `dune internal build-count` which internally makes an RPC call. In a future version of this I intend to have the benchmark runner make the RPC call directly. I'm working on this now, but it's turning out to be more complicated than I expected so I'll add that in a future PR.

gridbugs:
The change (to this repo!) to make the runner into an RPC client was actually pretty minor so I've included it here which meant I could remove the `dune internal build-count` command.

gridbugs:
> From the description of this PR, I still don't understand why the system being benchmarked (Dune) needs be modified in order to be benchmarked. The measurements you give in the example are just times: surely, you can collect that without any changes to Dune? Could you elaborate?

The changes are so dune can be benchmarked in watch mode specfically.

Dune will rebuild the project when source files change but it doesn't produce an easily observable side-effect when a build finishes, so it's difficult to know when to stop measuring the time taken by the rebuild. Of course, there are some side effects of a build completing, such as the "Build Complete" message that gets printed out and the new binary that gets created - at least for executable projects. I elected to add to the trace file instead as it's simpler than having the benchmark runner wait for other side effects of a build completing.

The new RPC type added in this change is so that the runner can find out when a build finishes. This is necessary because the runner needs to wait until a build finishes before making the source file change that will trigger the next build. If necessary we could also use this mechanism for recording the time taken by the rebuild and remove the need for dune to add build times to the trace file.

> Also, you appear to be fixing some RPC bug in this PR. Could you factor this out, so this fix can be reviewed separately?

Those were copied from a different PR (now merged) so I could test the benchmarks in current-bench. I'll remove them from this PR and rebase now that they've been merged.

snowleopard:
@gridbugs and I had a discussion in Slack and we agreed that:
* For now, it seems simpler to just use the passive file-watching mode in the monorepo benchmarks, so we'll do that first since that doesn't require any changes in Dune.
* For the eager mode, it's worth writing up an RFC describing the desired API. It's a subtle piece of functionality, and I can imagine other systems might want it, not just the monorepo benchmarks, so it would be nice to get more eyes on it.

gridbugs:
I just discovered dune's streaming RPC interface (`Dune_rpc_lwt.V1.Client.Stream`). i did not know that this existed and it removes the need for the changes to dune in this PR (adding to the trace-file and the "build-count" RPC). I have updated the benchmark runner to use the streaming interface and I'll close this PR now and open a new one which will just update the benchmarking docker file to use the new streaming version of the runner.

gridbugs:
New PR is here https://github.com/ocaml/dune/pull/7625

PR #7553, opened 2023-04-14
test: demonstrate regression depending on a ppx by its private name

anmonteiro:
- Reproduces https://github.com/ocaml/dune/issues/7543 and is a lighter weight alternative to https://github.com/ocaml/dune/pull/7544
- the issue seems to be about depending on the private name of a ppx within the same scope
  - this was possible before #7415, and is now broken

### DIFF
--- a/src/dune_rules/expander.ml
+++ b/src/dune_rules/expander.ml
@@ -80,7 +80,7 @@ let set_local_env_var t ~var ~value =
 
 let set_dir t ~dir = { t with dir }
 
-let set_scope t ~scope = { t with scope }
+let set_scope t ~scope ~scope_host = { t with scope; scope_host }
 
 let set_bin_artifacts t ~bin_artifacts_host = { t with bin_artifacts_host }
 
--- a/src/dune_rules/super_context.ml
+++ b/src/dune_rules/super_context.ml
@@ -85,9 +85,22 @@ end = struct
 
   let external_env t ~dir = get_node t ~dir >>= Env_node.external_env
 
+  let scope_host ~scope (context : Context.t) =
+    match context.for_host with
+    | None -> Memo.return scope
+    | Some host ->
+      let dir =
+        let root = Scope.root scope in
+        let src = Path.Build.drop_build_context_exn root in
+        Path.Build.append_source host.build_dir src
+      in
+      Scope.DB.find_by_dir dir
+
   let expander_for_artifacts ~scope ~external_env ~root_expander ~dir =
+    let+ scope_host = scope_host ~scope (Expander.context root_expander) in
     Expander.extend_env root_expander ~env:external_env
-    |> Expander.set_scope ~scope |> Expander.set_dir ~dir
+    |> Expander.set_scope ~scope ~scope_host
+    |> Expander.set_dir ~dir
 
   let extend_expander t ~dir ~expander_for_artifacts =
     let* bin_artifacts_host = bin_artifacts_host t ~dir in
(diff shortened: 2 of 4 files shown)


rgrinberg (approved):

anmonteiro:
@rgrinberg I pushed a fix for this, guaranteeing that we return the same context if there's no host context.

src/dune_rules/expander.ml
```diff
@@ -51,7 +51,7 @@ type t =
   ; bin_artifacts_host : Artifacts.Bin.t
   ; bindings : value Pform.Map.t
   ; scope : Scope.t
-  ; scope_host : Scope.t
+  ; scope_host : Scope.t option
```
rgrinberg, line 54: Why does the host scope have to be an option now?

anmonteiro: oh that's a good point. It doesn't.

src/dune_rules/super_context.ml
```diff
@@ -393,13 +393,19 @@ let create ~(context : Context.t) ~host ~packages ~stanzas =
...
     in
     let* scope = Scope.DB.find_by_dir context.build_dir in
-    let* scope_host = Scope.DB.find_by_dir context_host.build_dir in
+    let* scope_host =
+      match context_host with
+      | None -> Memo.return None
+      | Some context_host ->
+        let+ scope = Scope.DB.find_by_dir context_host.build_dir in
+        Some scope
+    in
```
rgrinberg, line 408: There's another piece of code in this file that looks suspicious:

```
  let expander_for_artifacts ~scope ~external_env ~root_expander ~dir =
    Expander.extend_env root_expander ~env:external_env
    |> Expander.set_scope ~scope |> Expander.set_dir ~dir
```

It doesn't seem right to set the scope without setting the scope_host appropriately. I guess we can fix this on another occasion.

anmonteiro: I was just typing this in another comment.

anmonteiro: I think the current fix is in the wrong place.

rgrinberg: There's a simpler way to derive the scope_host from the scope that perhaps we should follow:

1. Get the root directory of `scope`.
2. Transform this directory to the analogous directory in the host's context
3. Call `Scope.DB.find_by_dir` on it.

This kind of complicates my refactoring of making source tree context based, but it's ok for now.

rgrinberg: So for this change, you would need a `; context_host : Context.t option` field rather than a `scope_host`.

rgrinberg: Or perhaps we should modify `set_scope` to take both the `scope` and the `scope_host`.  and move this logic to whoever is calling `set_scope.` It seems good to keep Expander relatively oblivious to cross compilation logic.

anmonteiro: I pushed the version where `scope_host` is derived from context. I'll explore the version where `set_scope` takes `scope_host` now and see if I like it better.

anmonteiro: Pushed the `set_scope` fix instead, which is a lot more minimal. Let me know what you think.

rgrinberg (approved):

anmonteiro:
Looks like `Expander.set_bin_artifacts` sets the host artifacts already https://github.com/ocaml/dune/blob/82c670eac095b305d9d0def10b7c19c0e697303e/src/dune_rules/expander.ml#L85

PR #7554, opened 2023-04-14
feature: blake3

rgrinberg:
Still not completely ready yet, but worth examining this early for benchmarking

### DIFF
--- a/src/dune_digest/blake3.c
+++ b/src/dune_digest/blake3.c
@@ -1,0 +1,616 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <string.h>
+
+#include "blake3.h"
+#include "blake3_impl.h"
+
+const char *blake3_version(void) { return BLAKE3_VERSION_STRING; }
+
+INLINE void chunk_state_init(blake3_chunk_state *self, const uint32_t key[8],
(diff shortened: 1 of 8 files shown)


snowleopard:
Where is the implementation coming from? It doesn't seem to have any license.

> but worth examining this early for benchmarking

Have you done any benchmarks?

rgrinberg:
> Where is the implementation coming from? It doesn't seem to have any license.

It's taken from from cryptokit.

> Have you done any benchmarks?

Yes. In my benchmarks, this digest is around 30% faster on average than md5.

snowleopard:
Nice! Not particularly spectacular but still a good win. 

How fair was the comparison? I imagine performance depends quite a lot on the bindings/IO implementation itself, not just on the underlying algorithm, so is the 30% speed-up a consequence just of switching from MD5 to Blake3, or is there anything else going on?

rgrinberg:
> so is the 30% speed-up a consequence just of switching from MD5 to Blake3, or is there anything else going on?

For strings in memory, it's around a 30% speed up across the board.

For digesting very small files, the difference is between smaller-than-30% to barely-measurable. For somewhat larger files (10kb+), the difference is around 30% again.

PR #7555, opened 2023-04-15
refactor: remove vcs from source tree

rgrinberg:
It can be implemented on top of more primitive source tree functionality
and is not worth storing as its computed quickly.


rgrinberg:
I removed all the controversial bits of this PR. Now it's just a small optimization.

PR #6606, opened 2022-11-30
fix: link time code gen

rgrinberg:
do not "guess" object names when generating link time modules.
Actually look up the generated module and find the object name

Signed-off-by: Rudi Grinberg <me@rgrinberg.com>

ps-id: 8cd946aa-9f0d-4895-8f98-9bc1320a82a1

PR #7556, opened 2023-04-15
test: demonstrate crash in link-time codegen

anmonteiro:
- demonstrates a crash in link-time codegen when there's a library called `findlib.dynload`
- this is extracted from a real case (`caqti-dynload`) when consumed within an opam monorepo, which uses https://github.com/dune-universe/lib-findlib

### DIFF
--- a/src/dune_rules/link_time_code_gen.ml
+++ b/src/dune_rules/link_time_code_gen.ml
@@ -5,16 +5,18 @@ type t = Link_time_code_gen_type.t =
   ; force_linkall : bool
   }
 
-let generate_and_compile_module cctx ~precompiled_cmi ~name ~lib ~code ~requires
-    =
+let generate_and_compile_module cctx ~precompiled_cmi ~obj_name ~name ~lib ~code
+    ~requires =
   let sctx = Compilation_context.super_context cctx in
   let open Memo.O in
   let* module_ =
     let+ modules = Dir_contents.modules_of_lib sctx lib in
     let obj_name =
-      Option.map modules ~f:(fun modules ->
-          let mli_only = Modules.find modules name |> Option.value_exn in
-          Module.obj_name mli_only)
+      match obj_name with
+      | Some _ -> obj_name
+      | None ->
+        Option.map modules ~f:(fun modules ->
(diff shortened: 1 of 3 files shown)


anmonteiro:
could be a regression from https://github.com/ocaml/dune/pull/6606

anmonteiro:
I pushed a fix that avoids the crash, but there's still an error about `Findlib_initl.cmx`.

rgrinberg:
> I pushed a fix that avoids the crash, but there's still an error about Findlib_initl.cmx.

Yes, because the fix isn't correct. If we can't find the module we whose implementation we need to generate, we're not going to produce a final executable.

src/dune_rules/link_time_code_gen.ml
```diff
@@ -12,9 +12,8 @@ let generate_and_compile_module cctx ~precompiled_cmi ~name ~lib ~code ~requires
   let* module_ =
     let+ modules = Dir_contents.modules_of_lib sctx lib in
     let obj_name =
-      Option.map modules ~f:(fun modules ->
-          let mli_only = Modules.find modules name |> Option.value_exn in
-          Module.obj_name mli_only)
+      Option.bind modules ~f:(fun modules ->
+          Option.map (Modules.find modules name) ~f:Module.obj_name)
```
rgrinberg, line 16: How about we allow the caller to provide an `obj_name` and touching `modules` if it's provided.

test/blackbox-tests/test-cases/builtin-support-override.t
```diff
@@ -0,0 +1,42 @@
...
+
+  $ mkdir findlib
+  $ cat > findlib/dune-project <<EOF
+  > (lang dune 3.7)
+  > (package (name findlib))
+  > EOF
+  $ cat > findlib/dune <<EOF
+  > (library
+  >   (name findlib_dynload)
+  >   (public_name findlib.dynload)
```
rgrinberg, line 11: I think it's because this library needs to provide a `Dynlink` and `Findlib` modules like the actual `findlib.dynload` does via its dependencies

rgrinberg:
okay, this is ready

rgrinberg (approved):

PR #7560, opened 2023-04-16
refactor: remove Source_tree.file_exists

rgrinberg:
Fdo is the only user of this function, and it's easily switched to allow
the profile file to be generated.


rgrinberg:
ping @anmonteiro

PR #7562, opened 2023-04-16
fix: remove .merlin handling from the engine

rgrinberg:
We haven't relied on .merlin files in the rules for over 2 years. It's
OK to get rid of them now.


snowleopard (approved):
Looks good!

rgrinberg:
I haven't merged this yet because @ddickstein informed me that you still rely on .merlin files in internally. Is that still true with dune?

snowleopard:
I highly doubt we depend on the deleted code internally.

Why was `.merlin` treated specially in the first place? Any stale promoted file needs to be deleted. What made `.merlin` files special?

PR #7563, opened 2023-04-16
refactor: simplify source dependencies

rgrinberg:
Remove the [Source_tree] functor in [Dep.Set] and replace it with a
function [of_source_files]

This simplifies the API everywhere, and removes an assert false.

### DIFF
--- a/src/dune_rules/source_deps.mli
+++ b/src/dune_rules/source_deps.mli
@@ -1,5 +1,6 @@
 open Import
 
 (** [files path] returns all the source dependencies starting from [path]. If
-    [path] isn't in the source, the empty dependency set is returned *)
-val files : Path.t -> Dep.Set.t Memo.t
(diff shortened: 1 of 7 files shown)


rgrinberg:
ping @snowleopard

src/dune_rules/source_deps.mli
```diff
@@ -2,4 +2,4 @@ open Import
 
 (** [files path] returns all the source dependencies starting from [path]. If
     [path] isn't in the source, the empty dependency set is returned *)
-val files : Path.t -> Dep.Set.t Memo.t
+val files : Path.t -> (Dep.Set.t * Path.Set.t) Memo.t
```
snowleopard, line 5: Should the comment here be updated?

snowleopard (approved):
Looks great, thanks!

PR #7564, opened 2023-04-16
fix: delete stale directory targets

rgrinberg:
For unsandboxed rules, directory targets weren't being cleared when
running rules were interrupted. This PR treats files and directories in
the same way.


snowleopard (approved):
Looks good!

PR #7565, opened 2023-04-17
melange: test mdx stanza

jchavarri:
It seems `mdx` stanza can't be used with Melange libraries. This test documents it.


rgrinberg:
I guess this would be expected. Mdx relies on findlib IIRC. Worth mentioning in the docs I suppose.

PR #7566, opened 2023-04-17
melange: test odoc

jchavarri:
odoc integration with Melange is only partially working, the test in the PR documents it.


jchavarri:
Merged in https://github.com/ocaml/dune/pull/7639.

PR #7572, opened 2023-04-18
refactor: simplify cwd when expanding actions

rgrinberg:
Add an explicit chdir argument when expanding actions.

This makes it unnecessary to have Preprocesisng.chdir and the other
post-processing Chdir actions.


anmonteiro (approved):
Nice. This is gonna be handy for #7373

PR #7574, opened 2023-04-18
doc: add a lexer for opam files

emillon:
It does not cover the whole opam syntax but is good enough to support the few examples we have.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -23,8 +23,10 @@ sys.path.append(os.path.abspath('exts'))
 
 from sphinx.highlighting import lexers
 from dune_lexer import DuneLexer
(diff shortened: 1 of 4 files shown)


emillon:
The lexer can be seen in action [here](https://dune--7574.org.readthedocs.build/en/7574/howto/opam-file-generation.html) and [there](https://dune--7574.org.readthedocs.build/en/7574/explanation/opam-integration.html)

Alizter (approved):
Looks good. I noticed that the highlighting in opam-file-generation for the Dune files isn't working however.

emillon:
Good point, I'll fix this too.

PR #6594, opened 2022-11-27
feature: enable (include_subdirs qualified)

rgrinberg:
`(include_subdirs qualified)` for OCaml libraries and executables

## Aliasing

We now build an alias module for every directory. To build a module `x/y/z.ml`, we need to open `-open X -open Y`. Similarly, we can write an "interface module" for every directory:

```
$ ls -R
bar/
  foo.ml
  bar.ml
```

In this example, `Bar.Foo` isn't accessible unless it's re-exported by `bar.ml`

## Ocamldep Hacks

Currently ocamldep can't give us precise deps in situations such as:
```
$ ls -R
bar/
  baz.ml
foo.ml
$ cat foo.ml
let _ = Bar.Baz.foo
```

ocamldep gives us `Bar`, even though we really want to know that it's `Lib__Bar__Baz` that we need. For now we just depend on everything in bar/. Hopefully codept will help us out here.

## TODO
- [X] aliasing scheme
- [X] dune-package support 
- [X] interface files for module groups
- [x] merlin support
- [x] ppx
- [x] fix describe

## Out of scope (for subsequent PR's):
- documentation
- per module preprocessing, virtual libraries, modules without implementation, private modules
- unwrapped libraries
- support in bootstrap

PR #7577, opened 2023-04-18
merlin: avoid touching case of filename

nojb:
Recently we upgraded from Dune 3.5 to 3.7 at LexiFi and noticed a bug under Windows that causes the Merlin configuration for files that contain uppercase letters not to be found.

The issue seems to be that under Windows, we lowercase the full file path in order to match the workspace root in a case-insensitive fashion. However, as a side effect, we "lose" the uppercase characters in the basename itself.

As a solution, this PR proposes to only do the prefix matching in case-insensitive mode, but not to touch the actual filename.

cc @voodoos @rgrinberg

### DIFF
--- a/otherlibs/stdune/src/string.ml
+++ b/otherlibs/stdune/src/string.ml
@@ -42,45 +42,60 @@ let break s ~pos = (sub s ~pos:0 ~len:pos, sub s ~pos ~len:(length s - pos))
 
 let is_empty s = length s = 0
 
-let rec check_prefix s ~prefix len i =
-  i = len || (s.[i] = prefix.[i] && check_prefix s ~prefix len (i + 1))
+module Cased_functions (X : sig
+  val normalize : char -> char
+end) =
+struct
+  let rec check_prefix s ~prefix len i =
+    i = len
+    || X.normalize s.[i] = X.normalize prefix.[i]
+       && check_prefix s ~prefix len (i + 1)
 
-let rec check_suffix s ~suffix suffix_len offset i =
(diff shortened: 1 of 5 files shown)


nojb:
(In fact, this seems to be a regression, but I don't know when it was introduced.)

otherlibs/stdune/src/string.ml
```diff
@@ -82,6 +82,50 @@ let drop_suffix_if_exists s ~suffix =
   | None -> s
   | Some s -> s
 
+module Caseless = struct
```
rgrinberg, line 85: Perhaps it would make sense to functorize these prefix functions to avoid the copy paste?

rgrinberg (approved):
I think it was https://github.com/ocaml/dune/pull/6594 that introduced this issue.

rgrinberg:
Could you add a windows only test for this perhaps?

otherlibs/stdune/src/string.ml
nojb: Yes, that's possible. Another possibility is to add an optional argument to the existing functions. Do you have a preference?

nojb:
> Could you add a windows only test for this perhaps?

I'll give it a shot.

nojb:
> > Could you add a windows only test for this perhaps?
> 
> I'll give it a shot.

This is now done.

otherlibs/stdune/src/string.ml
nojb: This is now done (with a functor).

PR #7579, opened 2023-04-18
copy_file_portable: use binary mode to avoid corruption

nojb:
A regression in #7257 meant that the "portable" file copying routine was using text mode channels, corrupting binary files when copying them (under Windows). This explains the current Windows failures in the CI.

cc @rgrinberg

### DIFF
--- a/otherlibs/stdune/src/io.ml
+++ b/otherlibs/stdune/src/io.ml
@@ -73,7 +73,7 @@ let copy_channels =
   loop
 
 let setup_copy ?(chmod = Fun.id) ~src ~dst () =
-  let ic = open_in src in
+  let ic = Stdlib.open_in_bin src in
   let oc =
     try
       let perm = (Unix.fstat (Unix.descr_of_in_channel ic)).st_perm |> chmod in


otherlibs/stdune/src/io.ml
```diff
@@ -73,7 +73,7 @@ let copy_channels =
   loop
 
 let setup_copy ?(chmod = Fun.id) ~src ~dst () =
-  let ic = open_in src in
+  let ic = Stdlib.open_in_bin src in
```
anmonteiro, line 76: why does this need the `Stdlib.` qualifier?

anmonteiro (approved):
LGTM

otherlibs/stdune/src/io.ml
nojb: It doesn't, but that's the style in the rest of the file (I guess to emphasize the use of the standard library as opposed to other functions from `stdune` itself).

anmonteiro (commented):
Looking at this again, perhaps the `open_in` function around line 203 should be moved above the `setup_copy` function. That will restore the old behavior.

nojb:
> Looking at this again, perhaps the `open_in` function around line 203 should be moved above the `setup_copy` function. That will restore the old behavior.

Unfortunately, that line is actually inside a functor and cannot be lifted out verbatim...

nojb:
CI passes, except for a failure in one of the MSVC builds that looks unrelated. Merging.

Issue #7573, opened 2023-04-18
Build failing due to existence of broken symbolic link (unused in build)


geoffder:
## Expected Behavior
Not failing when a broken symbolic link exists (without being used by the current build).

## Actual Behavior
I am building a vendored cmake project which itself has vendored thrust, which contains intentionally broken symlinks in its test directories (not involved in present build). This is resulting in failure of the dune rule which is running cmake.
```
Error: File unavailable:
c/vendor/manifold/src/third_party/thrust/dependencies/libcudacxx/libcxx/test/std/input.output/filesystems/Inputs/static_test_env/bad_symlink
Broken symbolic link
```

Deleting the symbolic links allows the build to complete (not a satisfying solution as I'd like to not be stuck patching the vendored submodules).

Alizter:
Does making the cmake project `(data_only_subdirs)` allow it to work?

geoffder:
No unfortunately `(data_only_dirs)` didn't work when I tried it in either the current ctypes dune file with `(data_only_dirs vendor)`, or in a new dune file down a directory `(data_only_dirs manifold)` I made to make sure that it wasn't just failing to apply recursively.

Alizter:
Another thing you can try is `(ignored_subdirs)`.

Alizter:
Alright, then we can probably experiment with changing `(data_only_dirs)` to allow for broken symlinks. Would you be able to make a test-case with a broken symlink in an empty dune build? Or at least give some more details about how this particular symlink is broken.

Alizter:
Actually here is a testcase:
```

  $ cat > dune-project << EOF
  > (lang dune 3.8)
  > EOF

  $ ln -s doesnt_exist bar
  $ ls
  bar
  dune-project

  $ dune build bar
  File "bar", line 1, characters 0-0:
  Error: File unavailable: bar
  Broken symbolic link
  [1]
```

geoffder:
I think that case covers the same issue, both symlinks that are being hit are simply relative paths to files/directories that don't exist

Alizter:
I had a go at handling broken symlinks however there are some issues.

https://github.com/ocaml/dune/pull/7580

In Dune we digest each tracked file. For broken symlinks there appears to be some Unix API that allows us to do this, but it doesn't appear to be exposed by OCaml's Unix library. Therefore I haven't been able to workout how to digest them yet.

PR #7580, opened 2023-04-18
WIP handling for broken symlinks

Alizter:
This is an attempt at fixing #7573. I am stuck trying to work out how to digest a symlink when we know it is broken. Unix documentation seems to indicate that I need the `O_PATH` and `O_NOFOLLOW` file open modes which don't seem to exist in the Unix Ocaml stdlib API. Therefore this is only a WIP until I can work out how to digest broken symlinks.


Alizter:
This touches too much of the engine to be worth it.

PR #7581, opened 2023-04-19
Don't attempt read from closed fd in dune_rpc_lwt

gridbugs:
This fixes a bug where RPC clients built with dune_rpc_lwt would crash while disconnecting as they attempted to read from a channel whose underlying file descriptor had been closed.

A simple program that reproduces the bug:
```ocaml
let () =
  let init =
    Dune_rpc.V1.Initialize.create ~id:(Dune_rpc.V1.Id.make (Csexp.Atom "hello"))
  in
  let where = Dune_rpc_lwt.V1.Where.default ~build_dir:"_build" () in
  Lwt_main.run
    (let open Lwt.Syntax in
    let* input_channel, output_channel = Dune_rpc_lwt.V1.connect_chan where in
    Dune_rpc_lwt.V1.Client.connect (input_channel, output_channel) init
      ~f:(fun _client -> Lwt.return ()))
```
This program just connects to an RPC server and then immediately disconnects. Prior to this change it would crash while disconnection with the error:
```
Fatal error: exception Unix.Unix_error(Unix.EBADF, "check_descriptor", "")
Raised at Lwt_unix.check_descriptor in file "src/unix/lwt_unix.cppo.ml", line 379, characters 4-64
Called from Lwt_unix.blocking in file "src/unix/lwt_unix.cppo.ml", line 384, characters 2-21
Called from Lwt_unix.read_bigarray in file "src/unix/lwt_unix.cppo.ml", line 682, characters 4-15
Called from Lwt_io.perform_io in file "src/unix/lwt_io.ml", line 236, characters 10-35
Called from Lwt_io.Primitives.read_char in file "src/unix/lwt_io.ml", line 676, characters 6-15
Called from Lwt_io.Primitives.read_char_opt.(fun) in file "src/unix/lwt_io.ml", line 686, characters 17-29
Called from Lwt.Sequential_composition.catch in file "src/core/lwt.ml", line 2008, characters 16-20
Re-raised at Lwt.Miscellaneous.poll in file "src/core/lwt.ml", line 3077, characters 20-29
Called from Lwt_main.run.run_loop in file "src/unix/lwt_main.ml", line 31, characters 10-20
Called from Lwt_main.run in file "src/unix/lwt_main.ml", line 118, characters 8-13
Re-raised at Lwt_main.run in file "src/unix/lwt_main.ml", line 124, characters 4-13
Called from Dune__exe__Hello_rpc in file "hello_rpc.ml", line 6, characters 2-260
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -125,6 +125,9 @@ Unreleased
 - On nix+macos, pass `-f` to the codesign hook to avoid errors when the binary
   is already signed (#7183, fixes #6265, @greedy)
 
+- Fix bug where RPC clients built with dune-rpc-lwt would crash when closing
+  their connection to the server (#7581, @gridbugs)
+
 3.7.1 (2023-04-04)
 ------------------
 
--- a/otherlibs/dune-rpc-lwt/src/dune_rpc_lwt.ml
+++ b/otherlibs/dune-rpc-lwt/src/dune_rpc_lwt.ml
@@ -42,33 +42,45 @@ module V1 = struct
       (struct
         type t = Lwt_io.input_channel * Lwt_io.output_channel
 
-        let read (i, _) =
+        let read (i, o) =
+          (* The input and output channels share the same file descriptor. If
+             the output channel has been closed, reading from the input channel
+             will result in an error. *)
+          let is_channel_closed () = Lwt_io.is_closed o in
           let open Csexp.Parser in
           let lexer = Lexer.create () in
           let rec loop depth stack =
-            let* res = Lwt_io.read_char_opt i in
-            match res with
-            | None ->
+            if is_channel_closed () then (
               Lexer.feed_eoi lexer;
-              Lwt.return_none
-            | Some c -> (
-              match Lexer.feed lexer c with
-              | Await -> loop depth stack
-              | Lparen -> loop (depth + 1) (Stack.open_paren stack)
-              | Rparen ->
-                let stack = Stack.close_paren stack in
-                let depth = depth - 1 in
-                if depth = 0 then
-                  let sexps = Stack.to_list stack in
-                  sexps |> List.hd |> Lwt.return_some
-                else loop depth stack
-              | Atom count ->
-                let* atom =
-                  let bytes = Bytes.create count in
-                  let+ () = Lwt_io.read_into_exactly i bytes 0 count in
-                  Bytes.to_string bytes
-                in
-                loop depth (Stack.add_atom atom stack))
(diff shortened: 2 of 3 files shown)


rgrinberg:
I think I don't understand Lwt enough to understand why this is necessary. AFAIK, this is the line that does the closing:

```
        let write (_, o) = function
          | None -> Lwt_io.close o
          | Some csexps ->
            Lwt_list.iter_s
              (fun sexp -> Lwt_io.write o (Csexp.to_string sexp))
              csexps
```

Which should wake up any readers or writers with a `Channel_closed` exception and not an `EBADF`.

Any subsequent readers or writers would get the same exception.

gridbugs:
Here's a minimal program that crashes with `Fatal error: exception Unix.Unix_error(Unix.EBADF, "check_descriptor", "")`:
```ocaml
let make_socket () =
  let open Lwt.Syntax in
  let+ sock_fd =
    let fd = Lwt_unix.socket Unix.PF_UNIX Unix.SOCK_STREAM 0 in
    let+ () = Lwt_unix.connect fd (Unix.ADDR_UNIX "mysocket") in
    fd
  in
  let fd mode = Lwt_io.of_fd sock_fd ~mode in
  (fd Input, fd Output)

let run () =
  let open Lwt.Syntax in
  let* i, o = make_socket () in
  let* () = Lwt_io.close o in
  let* (_ch : char) = Lwt_io.read_char i in
  Lwt.return_unit

(* run socket server with `nc -lkU mysocket` *)
let () = Lwt_main.run (run ())
```

Lwt's docs say:
```ocaml
exception Channel_closed of string
  (** Exception raised when a channel is closed. The parameter is a
      description of the channel. *)

```

It's not clear to me under what conditions `Channel_closed` will be raised. I don't know if lwt keeps track of the association between the input and output channels when they share an underlying socket.

In any case you are correct that we should be checking if file is open before each read. Will fix.

gridbugs:
Updated to check that the channel is open before each read.

rgrinberg:
The end result still doesn't look satisfying to me. I have no idea why `Lwt_io.close` isn't waking up existing reads/writes. Another idea would be to make the write end do a shutdown on the write end of the socket with `Unix.shutdown` and let the callers manage their own fd's.

In any case, at least this is an improvement that's unblocking you. We'll accept it as is.

gridbugs:
Agreed that the workaround is unsatisfying. I create https://github.com/ocaml/dune/issues/7624 so we can improve it when we get time. Could also be a good starter issue I think.

PR #7582, opened 2023-04-19
feature: mdx 0.4 stanza with support for `mld` files out of the box

Leonidas-from-XIV:
MDX 2.3.0 added support for MLD files, the mdx 0.4 stanza adds `*.mld` to the default globs.

Probably it would make sense to add a version check for this and it would be good to land #7569 beforehand.

### DIFF
--- a/src/dune_rules/mdx.ml
+++ b/src/dune_rules/mdx.ml
@@ -187,11 +187,19 @@ let syntax =
     [ ((0, 1), `Since (2, 4))
     ; ((0, 2), `Since (3, 0))
     ; ((0, 3), `Since (3, 2))
+    ; ((0, 4), `Since (3, 8))
     ]
 
-let default_files =
-  let has_extension ext s = String.equal ext (Filename.extension s) in
-  Predicate_lang.Glob.of_pred (has_extension ".md")
+let glob_predicate repr =
+  repr |> Dune_lang.Glob.of_string_exn Loc.none |> Predicate_lang.Glob.of_glob
+
+let default_files_of_version version =
+  let md_files = glob_predicate "*.md" in
+  let mld_files = glob_predicate "*.mld" in
+  let mld_support_since = (0, 4) in
+  match Syntax.Version.Infix.(version >= mld_support_since) with
+  | true -> Predicate_lang.union [ md_files; mld_files ]
+  | false -> md_files
 
 let decode =
   let open Dune_lang.Decoder in
@@ -199,7 +207,7 @@ let decode =
     (let+ loc = loc
      and+ version = Dune_lang.Syntax.get_exn syntax
      and+ files =
-       field "files" Predicate_lang.Glob.decode ~default:default_files
+       field "files" Predicate_lang.Glob.decode ~default:Predicate_lang.Standard
      and+ enabled_if =
        Enabled_if.decode ~allowed_vars:Any ~since:(Some (2, 9)) ()
      and+ package =
@@ -249,7 +257,8 @@ let files_to_mdx t ~sctx ~dir =
   in
   let must_mdx src_path =
     let file = Path.Source.basename src_path in
-    Predicate_lang.Glob.exec t.files ~standard:default_files file
+    let standard = default_files_of_version t.version in
+    Predicate_lang.Glob.exec t.files ~standard file
   in
   let build_path src_path =
     Path.Build.append_source (Super_context.context sctx).build_dir src_path
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -128,6 +128,9 @@ Unreleased
 - Fix bug where RPC clients built with dune-rpc-lwt would crash when closing
   their connection to the server (#7581, @gridbugs)
 
+- Introduce mdx stanza 0.4 requiring mdx >= 2.3.0 which updates the default
+  list of files to include `*.mld` files (#7582, @Leonidas-from-XIV)
+
 3.7.1 (2023-04-04)
 ------------------
 
--- a/doc/stanzas/mdx.rst
+++ b/doc/stanzas/mdx.rst
@@ -18,9 +18,10 @@ Note that this feature is still experimental and needs to be enabled in your
 
 .. code:: dune
 
-  (using mdx 0.3)
+  (using mdx 0.4)
 
-.. note:: Version ``0.2`` of the stanza requires mdx ``1.9.0``.
+.. note:: Version ``0.2`` of the stanza requires mdx ``1.9.0``. Version ``0.4``
+   of the stanza requires mdx ``2.3.0``.
 
 
 The syntax is as follows:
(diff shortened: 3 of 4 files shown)


src/dune_rules/mdx.ml
```diff
@@ -185,19 +185,32 @@ let syntax =
...
     ; ((0, 3), `Since (3, 2))
+    ; ((0, 4), `Since (3, 8))
     ]
 
 let default_files =
   let has_extension ext s = String.equal ext (Filename.extension s) in
   Predicate_lang.Glob.of_pred (has_extension ".md")
 
+let default_files_of_version version =
+  let has_extension ext s = String.equal ext (Filename.extension s) in
```
rgrinberg, line 196: There's `Filename.check_suffix`

src/dune_rules/mdx.ml
```diff
@@ -185,19 +185,32 @@ let syntax =
...
+  let is_md = has_extension ".md" in
+  let mld_support_since = (0, 4) in
+  let is_mld =
+    match Syntax.Version.compare version mld_support_since with
+    | Eq | Gt -> has_extension ".mld"
+    | Lt -> fun _ -> false
+  in
+  Predicate_lang.Glob.of_pred (fun f -> is_md f || is_mld f)
```
rgrinberg, line 204: could you use `Predicate_lang.Glob.of_glob` if possible? We're trying to get rid of arbitrary predicates in the glob language.

emillon: I second this, you could define one predicate per version range and use globs:

```ocaml
let md_files = glob ...
let mld_files = glob ...
match Syntax.Version.compare ... with
| ... -> Predicate_lang.union md_files mld_files
| ... -> md_files
```

emillon:
> it would make sense to add a version check for this

what do you mean by this?

can you add a test for this (in addition to #7569)?

Also, can you update the docs of the `mdx` stanza? it should mention the new default. You can upgrade the `0.2` version in the example and suggest the new recommended version of mdx.

Thanks!

Leonidas-from-XIV:
> what do you mean by this?

I mean, at the moment there is code that checks that MDX is at least version 1.6, but for newer `mdx` stanzas (0.4 in this case) we need the user to have MDX 2.3 at least, otherwise the build will fail.

> can you add a test for this (in addition to https://github.com/ocaml/dune/pull/7569)?

Yes, will rebase once that is merged and add it there.

emillon:
> that checks that MDX is at least version 1.6

oh, there is? what is the failure more?

I'm not sure that it's actually possible to use the mdx stanza with mdx 1.6. It would be interesting to update these bounds.

src/dune_rules/mdx.ml
```diff
@@ -185,19 +185,28 @@ let syntax =
...
     ]
 
-let default_files =
-  let has_extension ext s = String.equal ext (Filename.extension s) in
-  Predicate_lang.Glob.of_pred (has_extension ".md")
+let glob_predicate loc repr =
+  repr |> Dune_lang.Glob.of_string_exn loc |> Predicate_lang.Glob.of_glob
+
+let default_files_of_version loc version =
+  let md_files = glob_predicate loc ".md" in
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Yes, as I added the tests I realized that `.md` didn't trigger. This is actually nicer than before as it directly reflects the syntax in the `dune` file.

src/dune_rules/mdx.ml
Leonidas-from-XIV: Done. It's indeed much nicer since the code before was rather inelegant (and was reinventing globs).

Leonidas-from-XIV:
> I'm not sure that it's actually possible to use the mdx stanza with mdx 1.6. It would be interesting to update these bounds.

It actually doesn't do any version checks, it tries `ocaml-mdx deps` and if that fails for whatever reason it displays an error message suggesting to use `mdx.1.6.0`.

Maybe doing version checks of MDX depending on the stanza version would make sense (as the docs state, 0.2 needs mdx.1.9.0 and 0.4 needs 2.3.0), though if you think that's a sensible addition I would suggest adding that in a separate PR.

emillon:
I don't think that version checks (in the "error if not fulfilled") are too useful, but I'd love to have a way to push a mdx dependency to the opam generation layer (`(using mdx 0.4)` would auto-generate `mdx (>= 2.3.0)`). I tried to do that some time ago but it was harder than I expected. You can revisit that idea if you think that the experience is not great (in a different PR of course).

Leonidas-from-XIV:
I haven't thought of this but pushing the dependency to the opam file as a test dependency is indeed a much nicer UX than just falling over and telling people to install a newer MDX!

src/dune_rules/mdx.ml
```diff
@@ -247,7 +256,8 @@ let files_to_mdx t ~sctx ~dir =
   in
   let must_mdx src_path =
     let file = Path.Source.basename src_path in
-    Predicate_lang.Glob.exec t.files ~standard:default_files file
+    let standard = default_files_of_version t.loc t.version in
+    Predicate_lang.Glob.exec t.files ~standard file
```
emillon, line 260: It's weird that we're doing this in 2 places. The way I understand it:
- at parsing time, we default to `default_files`
- here, we're saying that `:standard` evaluates to `default_files`

I think that we can just do that here, and at parse time default to `Standard` in the files field. (this can be done in a separate PR, or here)

emillon: as a bonus this should make the parser applicative again (no `let*`)

src/dune_rules/mdx.ml
```diff
@@ -185,19 +185,28 @@ let syntax =
...
     ; ((0, 2), `Since (3, 0))
     ; ((0, 3), `Since (3, 2))
+    ; ((0, 4), `Since (3, 8))
     ]
 
-let default_files =
-  let has_extension ext s = String.equal ext (Filename.extension s) in
-  Predicate_lang.Glob.of_pred (has_extension ".md")
+let glob_predicate loc repr =
+  repr |> Dune_lang.Glob.of_string_exn loc |> Predicate_lang.Glob.of_glob
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Was already wondering why the globs would need a `Loc.t`. Thanks, fixed.

src/dune_rules/mdx.ml
Leonidas-from-XIV: Was already wondering why this happened in two different places, but `:standard` explains it - indeed, it is nicer to just fill it with `Predicate_lang.Standard` then parsing and resolve it to the globs in the evaluation. Added it in this PR and also extended the test to show that passing `(files :standard)` processes `*.mld` files as well.

src/dune_rules/mdx.ml
```diff
@@ -187,19 +187,28 @@ let syntax =
...
+  | false -> md_files
 
 let decode =
   let open Dune_lang.Decoder in
   fields
     (let+ loc = loc
      and+ version = Dune_lang.Syntax.get_exn syntax
      and+ files =
-       field "files" Predicate_lang.Glob.decode ~default:default_files
+       let default = Predicate_lang.Standard in
```
emillon, line 210: nitpicking but if you inline `~default:Standard` that should both save a line in the file and in the diff

Leonidas-from-XIV: I wasn't sure whether it wouldn't be too long of a line, but it did work!

emillon (approved):

PR #7587, opened 2023-04-19
test: demonstrate top level directory not marked vendored

anmonteiro:


src/dune_rules/utop.ml
```diff
@@ -84,47 +85,53 @@ let libs_and_ppx_under_dir sctx ~db ~dir =
...
-                    in
-                    let+ pps =
-                      Resolve.Memo.read_memo
-                        (Preprocess.Per_module.with_instrumentation
-                           exes.buildable.preprocess
-                           ~instrumentation_backend:
-                             (Lib.DB.instrumentation_backend (Scope.libs scope)))
-                      >>| Preprocess.Per_module.pps
+                match Source_tree.Dir.status source_dir with
+                | Vendored | Data_only -> Memo.return (acc, pps)
```
rgrinberg, line 89: Why should we skip vendored libraries?

You can have libraries defined in data_only directories using `subdir`.

anmonteiro: Isn't this only for executables?

rgrinberg: It's still suspicious. Anything in the vendored directory is never filtered by `-p`, so the behavior in that issue was expected.

PR #7588, opened 2023-04-20
odoc: put odoc support files in odoc.support

### LINKED ISSUE #7364: Bad interaction between _odoc_supports directory and github pages / jekyll
On http://ocsigen.org/js_of_ocaml/5.1.0/api/index.html, odoc.css is not found. Yet, it is available in https://github.com/ocsigen/js_of_ocaml/blob/gh-pages/5.1.0/api/_odoc_support/odoc.css.

After searching more, it could be an issue with github pages and directories with leading underscores.
- https://github.blog/2009-12-29-bypassing-jekyll-on-github-pages/
- https://github.com/mpetrovich/stylemark/issues/65

_Originally posted by @hhugo in https://github.com/ocaml/dune/issues/6913#issuecomment-1476655932_

emillon:
Fixes #7364

If the directory is named `_odoc_support` with a leading underscore, it is omitted when deploying to github pages. This is because this system also support the jekyll site generator, which uses this convention for data files.

We can not use any directory name here since we want to avoid conflicts with OCaml identifiers and names of `*.mld` files. Starting with a dot seems to fit the requirements.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -160,6 +160,10 @@ Unreleased
   This is consistent with how `(executables)` handles this field.
   (#7576 , fixes #5852, @emillon)
 
+- Change directory of odoc assets to `odoc.support` (was `_odoc_support`) so
+  that it works with Github Pages out of the box. (#7588, fixes #7364,
+  @emillon)
+
 3.7.0 (2023-02-17)
 ------------------
 
--- a/src/dune_rules/odoc.ml
+++ b/src/dune_rules/odoc.ml
@@ -85,7 +85,7 @@ let add_rule sctx =
   Super_context.add_rule sctx ~dir
 
 module Paths = struct
-  let odoc_support_dirname = "_odoc_support"
+  let odoc_support_dirname = "odoc.support"
 
   let root (context : Context.t) =
     Path.Build.relative context.Context.build_dir "_doc"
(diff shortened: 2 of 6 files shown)


Alizter:
Should we also check that gh-pages works with this name?

Alizter:
I guess we should bite the bullet and just call it `odoc_support` with a way to configure it. Another option would be Rudi's suggestion of having a " _build/doc/$context" which would avoid any name clashes.

Alizter:
Adding a `.nojekyll` file seems like a good workaround until somebody is actually using jekyll in their gh-pages site. Perhaps a better alternative would be to generate a `_config.yml` file and explicitly include `_odoc_support`. This workaround is detailed here:
https://medium.com/@juliakmdev/learning-that-github-pages-struggles-with-certain-file-names-4d867d499ca9

emillon:
> Adding a `.nojekyll` file seems like a good workaround until somebody is actually using jekyll in their gh-pages site.

When would that be the case? If I understand correctly, the issue is that if you deploy the contents of the docs folder to the root of the gh-pages branch. In that case, you're deploying a static website, there can't be any jekyll in there. Are you thinking of the case where the doc folder is a subdirectory of a gh-pages website? In that case I don't think that `_config.yml` would be honored.

dbuenzli:
> For now, how about `odoc.support`? or even `odoc.assets` or similar?

Of course! I think upstream `odoc` could simply switch to that.

PR #7590, opened 2023-04-20
refactor: extract modules from Install_rules

emillon:
`Package_paths` and `Stanzas_to_entries` are made top level in `dune_rules`, and the modules related to installation are moved to a `install` subdirectory.


rgrinberg:
It's a bit of a shame to make so many additional functions public. Can we stick to exporting only what is being used?

rgrinberg:
Yes, Package_paths is the offender. It has a lot of functions that I wouldn't want anyone using.

PR #7591, opened 2023-04-20
ci: remove MSVC 4.12.1 builds

### LINKED ISSUE #7592: Fix MSVC 4.12.1 64 bit builds
These builds are failing at the moment so they were disabled in #7591.

Initial investigation:
The difference seems to be that it does go through the "Bootstrap opam from historical release (non-Windows; Windows non-fdopen)" branch in pre/action.yml. I (somehow naively) tried to update `bootstrap_opam_version=` in the 4.12 jobs to the same version as the 4.14 build, but it then fails on both jobs.

If I understand correctly, the previous "working" jobs would go through this branch:

https://github.com/ocaml/dune/blob/4c83fce53b52bf11a41644cec920fe929bf344dc/ci/setup-dkml/gh-windows-4.12.1/pre/action.yml#L529

but I'm not sure why these files don't exist anymore. are they restored from cache?

anyway, the "actual" error is when building the compiler:

    [dkml-base-compiler-60-1e6f74.out] OCAML_FLEXLINK="./boot/ocamlrun ./flexdll/flexlink.exe" D:/.opam/dkml/src-ocaml/support/env.exe D:/.opam/dkml/src-ocaml/support/ocamloptHost-final.wrapper  -o ocamlc.opt.exe compilerlibs/ocamlcommon.cmxa compilerlibs/ocamlbytecomp.cmxa driver/main.cmx -cclib "advapi32.lib ws2_32.lib version.lib"
    [dkml-base-compiler-60-1e6f74.out] 'D' is not recognized as an internal or external command,
    [dkml-base-compiler-60-1e6f74.out] operable program or batch file.
    [dkml-base-compiler-60-1e6f74.out] File "caml_startup", line 1:
    [dkml-base-compiler-60-1e6f74.out] Error: Error during linking (exit code 1)
    [dkml-base-compiler-60-1e6f74.out] make: *** [Makefile:656: ocamlc.opt.exe] Error 2

I don't know if the issue is the var=value command syntax, or the use of driver letters with / , but some of the path comes from `opam_root` which is set in that file.

emillon:
Fixes #7592

One of these builds has been failing for some time and after some investigation it seems that the other one is only working because of some cached assets that can not be recreated easily.

Quoting @jonahbeckford, DkML is a rolling distribution, so fixes are not backported to old compiler versions. At the moment of this commit, MSVC 4.14 is the current version and works as expected.


jonahbeckford:
Thanks. Investigating if the cache is corrupted in #7596  ...

emillon:
After discussing with @jonahbeckford and @rgrinberg we agree to remove the 4.12 builds completely.

PR #7593, opened 2023-04-20
feature: allow using bigarray in bootstrap

rgrinberg:
this is in preparation for the source fetching PR

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -557,6 +557,8 @@ module Config : sig
   val output_complete_obj_arg : string
 
   val unix_library_flags : string list
+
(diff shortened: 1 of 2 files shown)


nojb (commented):
Does this work in recent versions of OCaml (where `bigarray` does not exist anymore)?

rgrinberg:
Good point. Fixed for 5.0

nojb (commented):
The `Bigarray` module was moved to the standard library in 4.07. Dune requires 4.08 and later. Do we need the `bigarray` library at all?

rgrinberg:
Oh, I guess not.

PR #7596, opened 2023-04-20
Bump the cache for MSVC

jonahbeckford:
Fix (hopefully!) build failures for MSVC 4.12.1 64-bit


Alizter:
Looks like it still fails

jonahbeckford:
This was a test. Should have marked it as such!

Issue #7600, opened 2023-04-21
Internal error with `(include_subdirs qualified)` and directories having illegal module names


Niols:
## Expected Behavior

I would expect that a project using `(include_subdirs qualified)` with a layout such as this one (note the directory `bar-baz`):

```
./
├── dune-project
└── src/
    ├── bar-baz/
    ├── dune
    └── foo.ml
```

would either:

- compile fine, ignoring the directory in question,
- fail gracefully, recommending a workaround.

(Actually, I would also expect this workaround to be `(data_only_dirs bar-baz)` but this does not fix anything.)

## Actual Behavior

```console
$ dune exec src/foo.exe
Internal error, please report upstream including the contents of _build/log.
Description:
  ("Invalid Module_name.t", { s = "bar-baz" })
Raised at Stdune__code_error.raise in file
  "otherlibs/stdune/src/code_error.ml", line 11, characters 30-62
Called from Stdlib__List.rev_map.rmap_f in file "list.ml", line 103,
  characters 22-25
Called from Stdune__list.map in file "otherlibs/stdune/src/list.ml"
  (inlined), line 5, characters 19-33
Called from Dune_rules__ml_sources.make.(fun) in file
  "src/dune_rules/ml_sources.ml", line 459, characters 23-62
Called from Stdlib__List.fold_left in file "list.ml", line 121, characters
  24-34
Called from Dune_rules__ml_sources.make in file
  "src/dune_rules/ml_sources.ml", line 457, characters 8-1023
Called from Fiber__scheduler.exec in file "otherlibs/fiber/src/scheduler.ml",
  line 73, characters 8-11
-> required by ("<unnamed>", ())
-> required by ("<unnamed>", ())
-> required by ("load-dir", In_build_dir "default/src")

I must not crash.  Uncertainty is the mind-killer. Exceptions are the
little-death that brings total obliteration.  I will fully express my cases. 
Execution will pass over me and through me.  And when it has gone past, I
will unwind the stack along its path.  Where the cases are handled there will
be nothing.  Only I will remain.
```

## Reproduction

1. Get yourself a development environment with Dune 3.7.1. I use:
   ```
   $ nix shell github:nixos/nixpkgs/0a0f9e92d#ocamlPackages.{ocaml,dune_3,findlib}
   ```

2. Create a project, eg. [this one](https://github.com/ocaml/dune/files/11295065/issue7600.zip), which I detail here:
   ```console
   $ cat dune-project
   (lang dune 3.7)
[...]

PR #7601, opened 2023-04-21
fix: directory module name checking

Alizter:
fix https://github.com/ocaml/dune/issues/7600

### DIFF
--- a/src/dune_rules/ml_sources.ml
+++ b/src/dune_rules/ml_sources.ml
@@ -456,7 +456,13 @@ let make dune_file ~dir ~scope ~lib_config ~loc ~lookup_vlib
       | Include Qualified ->
         List.fold_left dirs ~init:Module_trie.empty
           ~f:(fun acc ((dir : Path.Build.t), local, files) ->
-            let path = List.map local ~f:Module_name.of_string in
+            let path =
+              List.map local ~f:(fun m ->
+                  Module_name.parse_string_exn
+                    ( Loc.in_dir
+                        (Path.drop_optional_build_context (Path.build dir))
+                    , m ))
+            in
             let modules = modules_of_files ~dialects ~dir ~files ~path in
             match Module_trie.set_map acc path modules with
             | Ok s -> s
--- a/test/blackbox-tests/test-cases/github7600.t
+++ b/test/blackbox-tests/test-cases/github7600.t
@@ -1,0 +1,70 @@
+Reproducing bug in https://github.com/ocaml/dune/issues/7600
+
+When using (include_subdirs qualified), valid module names should be checked for
+directories too.
+
+
+  $ cat > dune-project << EOF
+  > (lang dune 3.8)
+  > EOF
+
+  $ mkdir src
+
+  $ cat > src/dune << EOF
+  > (include_subdirs qualified)
+  > (executable (name foo))
+  > EOF
+
+  $ cat > src/foo.ml
+
+  $ mkdir src/baz-bar
+
+  $ dune build @install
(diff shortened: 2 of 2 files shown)


src/dune_rules/ml_sources.ml
```diff
@@ -456,7 +456,10 @@ let make dune_file ~dir ~scope ~lib_config ~loc ~lookup_vlib
       | Include Qualified ->
         List.fold_left dirs ~init:Module_trie.empty
           ~f:(fun acc ((dir : Path.Build.t), local, files) ->
-            let path = List.map local ~f:Module_name.of_string in
+            let path =
+              List.map local ~f:(fun m ->
+                  Module_name.parse_string_exn (Loc.none, m))
```
rgrinberg, line 461: Can you set the location here using `Loc_in_dir` here?

anmonteiro (commented):
Could we mangle invalid names instead of erroring?

Alizter:
@anmonteiro If we mangle, how is the user supposed to refer to it in their file?

anmonteiro:
I think it'd be beneficial to support some of those directories and document / log the mangling scheme. For melange specifically, we'll probably want to have directories called `[id]` and `(layout)` (check https://beta.nextjs.org/docs/routing/defining-routes).

Alizter:
I suppose another solution would be to make the module name of a directory configurable from a stanza. So we could do
```
(subdir foo-bar (qualified_name foo_bar))
```
That way when Dune wants to give module names corresponding to directories, it would query a naming map which would be overridden with these stanzas.

anmonteiro:
I like a solution along those lines. We could even reuse `include_subdirs` for it:


```
(include_subdirs 
  (mode qualified)
  ((src foo) (src/invalid-[module]-name invalid_module_name)))
```

anmonteiro:
(And it goes without saying: I don't think these should be implemented in this PR). The bug fix is sensible and looks good to me.

Alizter:
@anmonteiro Can you open a ticket with this feature request then?

anmonteiro:
https://github.com/ocaml/dune/issues/7605

rgrinberg (approved):

rgrinberg (approved):

Niols:
I think the feature request in #7605 makes a lot of sense. I guess two other solutions to a directory with a malformed module name would be:

1. The directory in question contains only data and one uses `(data_only_dirs bar-baz)` (or `data_only_dirs` on an ancestor of the directory with a malformed module name).

2. The directory in question could be ignored altogether. (Is there a stanza for this?)

Should I open feature requests for those as well?

Alizter:
> 1. The directory in question contains only data and one uses `(data_only_dirs bar-baz)` (or `data_only_dirs` on an ancestor of the directory with a malformed module name).

I am not sure what exactly you mean here. Could you open an issue explaining this further?

> 2. The directory in question could be ignored altogether. (Is there a stanza for this?)

You can use the `(dirs)` stanza to remove a directory from Dunes vision. Something like `(dirs :standard \ ignored_dir)`.

Niols:
cf https://github.com/ocaml/dune/pull/7620 for two examples. One is your suggestion for `(dirs)`, which I didn't know about, and which works like a charm. The other one uses `(data_only_dirs)`, hoping that Dune will be fine with a directory then having an invalid module name. This is however not the case. I will open a feature request for this.

Niols:
Feature request in question: https://github.com/ocaml/dune/issues/7628.

rgrinberg:
Yes

PR #7603, opened 2023-04-21
test: (include_subdirs qualified) with (modules :standard \ baz)

Alizter:
repro case for https://github.com/ocaml/dune/issues/7597

Looks like:
https://github.com/ocaml/dune/blob/2a1c7e0147d81ed04593c3cdb2cfe0e2686efeab/src/dune_rules/module_trie.ml#L105-L110
is unfinished.

cc @rgrinberg


rgrinberg:
Test looks fine but it cannot be merged with the stack trace. You can try implementing the feature or filtering the stack trace.

Alizter:
@rgrinberg No harm in keeping it as draft until we fix it?

PR #7604, opened 2023-04-21
refactor: use Pp.enumerate more

Alizter:
We use Pp.enumerate more and clean up some code.

(Previously we introduced Pp.enumerate_str but then decided against).


rgrinberg:
I don't think function is useful enough.

Alizter:
@rgrinberg Do you agree that we should use Pp.enumerate at least?

rgrinberg:
Why not. If you think it helps with readability.

Alizter:
Actually nevermind it was more readable before.

Issue #7465, opened 2023-03-31
VSCode instrumentation


epatrizio:
Hello, In vscode, it seems that ocaml-lsp automatically launches the instrumentation and does it incorrectly.
I made a minimal example using `mutaml`. Here are the steps to reproduce:
``` shell-session
$ git clone https://github.com/jmid/mutaml.git
$ cd mutaml
$ opam install .
$ cd ..
$ git clone https://github.com/epatrizio/mutaml_exp.git
$ cd mutaml_exp
$ git checkout vscode
```
Then, open VSCode on `mutaml_exp`. If you open any `.ml` file (e.g. `lib/lib.ml`) and put your mouse focus on it, you'll see some files appearing in the sources.

There are two issues there:
- first, the instrumentation should not run automatically
- second, the files end up in the wrong place (you can compare to what happen when you run `dune runtest --instrument-with mutaml` not from vscode), they end-up in `lib/` when they should be somewhere in `_build/`.

Thanks a lot!

zapashcanon:
FYI I don't have the bug when using dune+merlin from neovim or my terminal directly, we only managed to trigger it from vscode.

rgrinberg:
I'm really surprised by that. Lsp/Merlin is just launching whatever dune tells it to. If anyone would like to try their hand at the fix, the issue is how we construct the `preprocess` that we provide `Merlin.make`:

```
  let preprocess =
    Preprocess.Per_module.with_instrumentation lib.buildable.preprocess
      ~instrumentation_backend:
        (Lib.DB.instrumentation_backend (Scope.libs scope))
  in
```

We should always just drop the instrumentation for merlin

PR #7606, opened 2023-04-21
fix: drop instrumentation from Merlin.make

Alizter:
We drop any instrumentation preprocessing for merlin. This also simplifies some of the code.

fixes https://github.com/ocaml/dune/issues/7465

- [x] changelog
- [x] test


CHANGES.md
```diff
@@ -1,6 +1,9 @@
 Unreleased
 ----------
 
+- Fix a bug where merlin would launch the instrumentaiton. (#7606, fixes #7465,
```
rgrinberg, line 4: How about:

```
merlin: ignore instrumentation settings for preprocessing
```

rgrinberg (approved):

rgrinberg:
> I haven't got a test case for this. Not sure how to test.

In the instrumentation tests, you can see how to add setup instrumentation.

In the merlin tests, you can see how to print merlin settings when preprocessors are present.

Alizter:
Oh I still haven't added the tests for this.

Alizter:
Since my ocamlc lives in the nix store, the path is very long, so before PATH_PREFIX_MAP replaces the path, the merlin config gets formatted for the long line. This means that the test is not reproducible depending on the location of ocamlc. I'm not sure if we can fix this since the cram test is doing the substitution.

I guess the simplest solution is just to remove the offending lines. But I am not certain how to do this easily.

Issue #6865, opened 2023-01-10
Using `menhir --explain`


thierry-martinez:
Before `dune.3.0`, with the following stanza,

```lisp
(menhir (modules ...) (flags --explain))
```

when the menhir grammars have conflicts, the reports that `menhir` generates to describe the conflicts (`.conflicts` files, one for each `.mli` file) were available in `_build/$profile/`.

Since `dune.3.0`, sandboxing prevents `dune` from keeping these files, forcing the user to run `menhir --explain` manually in the source tree to get the reports.

## Desired Behavior

The `.conflicts` files could be copied to the `_build/$profile/` directory: that would restore the previous behavior (before `dune.3.0`), which was better than the current situation, even if I would prefer never have to touch to the ``_build/` directory.

Another behavior could be to promote the `.conflicts` files in the source tree, which would be close to the experience of using `menhir` without `dune`, and which would made the `.conflicts` file easier to notice. Another possibility could be to dump these files to the standard output when they exist (which can be convenient, but with the same drawbacks than with warnings, that are not displayed again in case of rebuild).

## Example

A file `test.conflicts` should be available somewhere (or its contents should be displayed) after compiling the following file.

`test.mly`:
```
%token A
%start<unit> a
%%
let a := A; {} | A; {} 
```

`dune`:
```
(library (name test))
(menhir (modules test) (flags --explain))
```

This issue was previously discussed here: https://discuss.ocaml.org/t/how-to-use-menhir-explain-with-dune/11132

Alizter:
You can disable sandboxing with `--sandbox none`. Does that work?

rgrinberg:
Promoting it doesn't seem right as it seems like a file with error messages/diagnostics. I wonder why menhir doesn't just dump this output into stderr? @fpottier Why is the separate file necessary?

thierry-martinez:
@Alizter, thank you for the tip: indeed, `--sandbox none` works (`.conflicts` files are available in `_build/$profile`). I think I would still prefer not to have to disable sandboxing to get these files.

thierry-martinez:
@rgrinberg, indeed, these are diagnostics. Piping them to stdout/stderr would be an option, but what would be the right place with `dune` for storing such files with errors or diagnostics if not in the source tree? Storing them in `_build/` looks a bit arcane.

nojb:
[...]

PR #7607, opened 2023-04-21
menhir: support for .conflicts targets when passing --explain

Alizter:
We now generate the .conflicts target if `--explain` was passed.

This allows for `--explain` to be passed in the menhir stanza and Dune will be aware of and expect a .conflict file. However, it is forbidden to pass using the `menhir_flags` field in an `env` stanza.

Previously I tried to allow for it in the env stanza, but we will need Dune to improve on some other fronts. Specifically rules are loaded per directory which makes it difficult to have slight variations of rules with different targets in the same directory. If we had more fine-grained rule loading, then we could choose which targets we want depending on the flags in the env stanza.

fix https://github.com/ocaml/dune/issues/6865

### DIFF
--- a/src/dune_rules/dune_env.ml
+++ b/src/dune_rules/dune_env.ml
@@ -19,7 +19,31 @@ module Stanza = struct
       Option.map since ~f:(fun since ->
           Dune_lang.Syntax.since Menhir_stanza.syntax since)
     in
-    Ordered_set_lang.Unexpanded.field "menhir_flags" ?check
+    let* menhir_version = Dune_lang.Syntax.get Menhir_stanza.syntax in
+    let+ flags = Ordered_set_lang.Unexpanded.field "menhir_flags" ?check in
+    Ordered_set_lang.Unexpanded.fold_strings flags ~init:()
+      ~f:(fun _pos sw acc ->
+        match String_with_vars.text_only sw with
+        | None -> acc
+        | Some text -> (
+          if String.equal text "--explain" then
+            match menhir_version with
+            | None -> ()
+            | Some menhir_version ->
+              if menhir_version >= (2, 2) then
+                User_error.raise ~loc:(String_with_vars.loc sw)
+                  [ Pp.text
+                      "The --explain flag is not supported in the env stanza \
+                       as it currently does not allow Dune's menhir support to \
+                       produce the correct targets."
+                  ]
+                  ~hints:
+                    [ Pp.text
+                        "Add --explain to the (flags) field of the (menhir) \
+                         stanza. This will produce the correct .conflict \
+                         files."
+                    ]));
+    flags
 
   module Inline_tests = struct
     type t =
--- a/src/dune_rules/menhir/menhir_rules.ml
+++ b/src/dune_rules/menhir/menhir_rules.ml
@@ -74,10 +74,13 @@ module Run (P : PARAMS) = struct
 
   let source m = Path.relative (Path.build dir) (m ^ ".mly")
 
-  let targets m ~cmly =
-    let base = [ m ^ ".ml"; m ^ ".mli" ] in
-    List.map ~f:(Path.Build.relative dir)
-      (if cmly then (m ^ ".cmly") :: base else base)
+  let targets m ~cmly ~explain =
+    let base =
+      let cmly = if cmly then [ m ^ ".cmly" ] else [] in
+      let conflicts = if explain then [ m ^ ".conflicts" ] else [] in
+      List.concat [ conflicts; cmly; [ m ^ ".ml"; m ^ ".mli" ] ]
+    in
+    List.map base ~f:(Path.Build.relative dir)
 
   let sources ms = List.map ~f:source ms
 
@@ -170,6 +173,23 @@ module Run (P : PARAMS) = struct
                       text
                   ]))
 
+  let explain ~menhir_lang_version
+      (expanded_flags : Ordered_set_lang.Unexpanded.t) =
+    (* Given a list of [expanded_flags], which are generally provided as an
+       [Action_builder.t], we check if "--explain" ever occurs by running the
+       action. Working on [expanded_flags] is important as this will allow us to
+       check any flags appearing in (env) stanzas. *)
+    menhir_lang_version >= (2, 2)
+    &&
+    try
+      Ordered_set_lang.Unexpanded.fold_strings expanded_flags ~init:false
+        ~f:(fun _pos sw acc ->
+          match String_with_vars.text_only sw with
+          | None -> acc
+          | Some text ->
+            if String.equal text "--explain" then raise_notrace Exit else false)
+    with Exit -> true
+
   (* ------------------------------------------------------------------------ *)
 
   (* [process3 stanza] converts a Menhir stanza into a set of build rules. This
@@ -218,6 +238,9 @@ module Run (P : PARAMS) = struct
       Module_compilation.ocamlc_i ~deps cctx mock_module
         ~output:(inferred_mli base)
     in
+    let explain =
+      explain ~menhir_lang_version:stanza.menhir_lang_version stanza.flags
+    in
     (* 3. A second invocation of Menhir reads the inferred [.mli] file. *)
     menhir
       [ Command.Args.dyn expanded_flags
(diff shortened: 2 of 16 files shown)


src/dune_rules/menhir/menhir_rules.ml
```diff
@@ -170,6 +173,15 @@ module Run (P : PARAMS) = struct
...
                   ]))
 
+  let explain (stanza : stanza) =
+    Ordered_set_lang.Unexpanded.fold_strings stanza.flags ~init:false
+      ~f:(fun _pos sw explain ->
+        explain
+        ||
+        match String_with_vars.text_only sw with
+        | None -> false
+        | Some text -> String.equal "--explain" text)
```
rgrinberg, line 183: No need to go through all the flags once we find one instance.

src/dune_rules/menhir/menhir_rules.ml
```diff
@@ -74,10 +74,13 @@ module Run (P : PARAMS) = struct
...
 
-  let targets m ~cmly =
-    let base = [ m ^ ".ml"; m ^ ".mli" ] in
-    List.map ~f:(Path.Build.relative dir)
-      (if cmly then (m ^ ".cmly") :: base else base)
+  let targets m ~cmly ~explain =
+    let base =
+      let cmly = if cmly then [ m ^ ".cmly" ] else [] in
+      let conflicts = if explain then [ m ^ ".conflicts" ] else [] in
+      conflicts @ cmly @ [ m ^ ".ml"; m ^ ".mli" ]
```
rgrinberg, line 81: `List.concat`

rgrinberg (changes requested):
Can you guard  this behind a newer version of the menhir lang?

src/dune_rules/menhir/menhir_rules.ml
Alizter: Yes true, but I will have to make an iterator for Ordered_set_lang that exits early. Since the number of flags is small I didn't think it would be a big deal.

rgrinberg: It's about the intent rather than performance. We're using `fold` where we'd really like to use `find_map`.

src/dune_rules/ordered_set_lang.ml
```diff
@@ -314,6 +314,17 @@ module Unexpanded = struct
...
     loop t.ast Pos init
 
+  let exists t ~f =
+    let rec loop (t : ast) =
+      let open Ast in
+      match t with
+      | Standard | Include _ -> false
+      | Element x -> f x
+      | Union l -> List.exists l ~f:loop
+      | Diff (l, f) -> loop l <> loop f
```
rgrinberg, line 324: The `Diff` case doesn't seem right.

Looking at this again, it's ok to use `fold_strings` and just raise a `Found` exception if you find the flag in Positive position. Reimplementing this logic isn't trivial.

Alizter: OK, I've just decided to check the expanded version instead as that is much simpler and more correct. We want to also check the --explain flag appearing in the env stanza. Therefore i have added an extra test case demonstrating --explain given from an env stanza.

Alizter:
Hmm executing the action for expanding the flags didn't really work as I intended. I am not really sure how else I can get this information before passing to the targets. @rgrinberg any ideas how I might break this cycle?

Alizter:
OK now we have changed so that --explain can only be passed in the flags field of the menhir stanza. If it is attempted to be passed in the env stanza, Dune will error.

In order to allow for it to be passed in the menhir stanza, we need to have more fine grained rule loading. The issue is that if we allow --explain in env, then when the rules of a directory are being loaded, the menhir rules have to determine which targets to choose at the same time as evaluating the action producing the decision. This causes a memo cycle which can be resolved by stratifying a project across multiple directories however the memo cycle error is confusing for users. The previous version of this PR meant that menhir itself couldn't build.

This new version hides behind menhir lang 2.2 the ability to add the .conflict targets by passing --explain in the menhir stanza. And therefore should cause no issues in the reverse dependencies.

src/dune_rules/dune_env.ml
```diff
@@ -19,7 +19,28 @@ module Stanza = struct
...
-    Ordered_set_lang.Unexpanded.field "menhir_flags" ?check
+    located (Ordered_set_lang.Unexpanded.field "menhir_flags" ?check)
+    >>| fun (loc, flags) ->
+    let () =
+      Ordered_set_lang.Unexpanded.fold_strings flags ~init:()
+        ~f:(fun _pos sw acc ->
+          match String_with_vars.text_only sw with
+          | None -> acc
+          | Some text ->
+            if String.equal text "--explain" then
```
rgrinberg, line 30: Do we have the same error for cmly? If no, I would suggest to keep it uniform and add a similar error there (can be done in a separate PR).

test/blackbox-tests/test-cases/menhir/explain.t/run.t
```diff
@@ -0,0 +1,51 @@
...
+  > EOF
+
+  $ dune build ./src/test.exe
+  $ ls _build/default/src/test.exe
+  _build/default/src/test.exe
+
+Check for .conflicts files in stanza with --explain. This will fail because the
+menhir lang version is not 2.2 or above.
+
+  $ find _build/default/src -name *.conflicts
```
rgrinberg, line 20: Isn't the `*` being expanded by the shell? It should be `-name '*.conflicts'`.

test/blackbox-tests/test-cases/menhir/explain.t/run.t
```diff
@@ -0,0 +1,51 @@
...
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.8)
+  > (using menhir 2.2)
+  > EOF
+  $ dune build ./src/test.exe
+
+Check for .conflicts files in stanza with --explain.
+
+  $ find _build/default/src -name *.conflicts
```
rgrinberg, line 32: ditto

src/dune_rules/dune_env.ml
rgrinberg: Also, this check should be guarded by 2.2

src/dune_rules/dune_env.ml
```diff
@@ -19,7 +19,28 @@ module Stanza = struct
...
+              User_error.raise ~loc
```
rgrinberg, line 31: The location should come from `sw` and not the stanza

src/dune_rules/dune_env.ml
Alizter: I've made a patch making this uniform that I'll submit after this is merged.

src/dune_sexp/syntax.mli
```diff
@@ -121,4 +121,6 @@ val key : t -> Key.t Univ_map.Key.t
 
 val get_exn : t -> (Version.t, 'k) Decoder.parser
 
+val get : t -> (Version.t option, 'k) Decoder.parser
```
Alizter, line 124: I added the option version of get_exn since I want to ignore the failure during the version checks. The failure is the fact that menhir lang is not declared, and when validating menhir flags its not relevantl

src/dune_rules/dune_env.ml
```diff
@@ -19,7 +19,31 @@ module Stanza = struct
...
+    let* menhir_version = Dune_lang.Syntax.get Menhir_stanza.syntax in
+    let+ flags = Ordered_set_lang.Unexpanded.field "menhir_flags" ?check in
+    Ordered_set_lang.Unexpanded.fold_strings flags ~init:()
+      ~f:(fun _pos sw acc ->
+        match String_with_vars.text_only sw with
+        | None -> acc
+        | Some text -> (
+          if String.equal text "--explain" then
+            match menhir_version with
+            | None -> ()
```
Alizter, line 31: We do nothing if getting the menhir lang failed for whatever reason. This is fine since this is just a check for the user. This logic will be cleaned up in a later PR.

Alizter:
Hmm. I can't seem to work out how to version guard when validating in dune_lang. The way I get the version in the rules is to record it when parsing the stanzas, like we do for Coq. I am not sure what we can do in env.

emillon:
Clearing the milestone as discussed in the dev meeting.

Alizter:
This started off as a quickfix to support something that we didn't really support in the first place. It has since grown into a feature which should really be thought out by menhir people. I am unsure what exactly it is they want out of this feature so I am closing this PR. Feel free to take commits and improve it.

Issue #7026, opened 2023-02-08
`modules` field in conjunction with `private_modules`, `virtual_modules`, etc.


rgrinberg:
The issues #7022, #7023, #7024 demonstrate a similar set of issues where a module is absent from `modules`, but is present in some other module list. Dune unfortunately ignores this discrepancy. At best, this is a harmless mistake, and at worst, this leads to some confusing build errors.

I propose that we do the following:

1. We warn the user when this mistake is made for now.
2. In 3.8, we turn this into a hard error.

Wdyt @nojb @emillon

nojb:
Sounds good to me. When turning it into an error, a hint ("Have you forgotten to add it to `(modules ...)`") may be useful as well.

emillon:
I agree about the feature, but I'd push back the cycle by 1: add the warning in 3.8 and make it fatal in 3.9. It's late in the 3.7 cycle and I could see add many unforeseen warnings.

emillon:
It's still a good time to implement the warning so that it can be tested in an 3.8~alpha release.

Alizter:
Was an easy fix: https://github.com/ocaml/dune/pull/7608

PR #7608, opened 2023-04-21
feature: warn if modules is missing any mentioned modules 

Alizter:
We warn the user if modules_without_implementation, private_modules or
virtual_modules contains any modules not in the modules field.

Fixes https://github.com/ocaml/dune/issues/7026

This will be made into an error in 3.9.


rgrinberg (changes requested):
This warning should be ignored for vendored directories.

Alizter:
I've allowed the warning to be ignored when appearing inside a vendored directory and checked the vendored_dirs cases in each of the tests.

src/dune_rules/modules_field_evaluator.mli
```diff
@@ -25,5 +25,6 @@ val eval :
   -> private_modules:Ordered_set_lang.t
   -> kind:kind
   -> src_dir:Path.Build.t
+  -> is_vendored:bool
```
rgrinberg, line 28: We're already passing src_dir. Can we check if it's vendored inside eval?

src/dune_rules/modules_field_evaluator.ml
```diff
@@ -213,6 +227,21 @@ let check_invalid_module_listing ~stanza_loc ~modules_without_implementation
       (unimplemented_virt_modules |> Module_name.Path.Set.to_list
       |> List.map ~f:(fun name -> (stanza_loc, name)))
       [ Pp.text "You must provide an implementation for all of these modules." ];
+    (* Checking that (modules) incldues all declared modules *)
```
rgrinberg, line 230: `includes.

src/dune_rules/modules_field_evaluator.ml
```diff
@@ -213,6 +227,21 @@ let check_invalid_module_listing ~stanza_loc ~modules_without_implementation
...
+    let print_undelared_modules field mods =
+      (* TODO: this is a warning for now, change to an error in 3.9. *)
+      (* If we are in a vendored stanza we do nothing. *)
+      if is_vendored then ()
```
rgrinberg, line 234: `if not is_vendored ...`

src/dune_rules/modules_field_evaluator.mli
rgrinberg: I don't see an issue with that.

rgrinberg (approved):

Alizter:
@rgrinberg merging this when the CI is green.

PR #7616, opened 2023-04-23
ci: include coverage for coq and melange

Alizter:
We include the Coq and melange tests in the coverage. If it doesn't take much longer than the current coverage job, I think it will be worth doing.

Coverage job jumps from 17m to 26m. Not sure... OTOH coverage never catches any failures since it always succeeds and can be safely ignored. what is important is that the report is generated and uploaded to coveralls.

Edit: nevermind, the Coq tests were not being run.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,8 @@ dev-deps:
 
 .PHONY: coverage-deps
 coverage-deps:
-	opam install -y bisect_ppx
+  # We install all test deps including coq deps
+	opam install -y $(TEST_DEPS) bisect_ppx coq-native coq.8.16.1
 
 .PHONY: dev-switch
 dev-switch:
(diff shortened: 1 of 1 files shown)


Alizter:
Inspecting the report:

https://coveralls.io/builds/59223464

It doesn't appear to be all that useful since the tests mostly fail due to the instrumentation output/artifacts. I think it would be more useful to instrument a monorepo. So in the end I am probably against this change.

ejgallego:
> It doesn't appear to be all that useful since the tests mostly fail due to the instrumentation output/artifacts. I think it would be more useful to instrument a monorepo. So in the end I am probably against this change.

It'd be great to have coverage for Coq, however the coverage results look pretty bizarre, it seems like no Coq test was executed?

What are these failures you talk about?

Alizter:
Ah nevermind, I forgot to enable the coq tests... I will try again soon.

Alizter:
In the end, not really worth it.

PR #7618, opened 2023-04-23
fix: files recorded by sandboxed cinaps action in action builder

### LINKED ISSUE #7376: cinaps fails with Error: Unable to resolve symlink
When creating a dune project with `cinaps 1.2` and a file to run cinaps on, `@runtest` will fail with
```
Error: Unable to resolve symlink
```
on the file we apply cinaps to. `cinaps 1.0` works fine.
@hhugo suspects it's related to #6176

## Steps to reproduce
```
git clone git@github.com:smuenzel/sandbox2.git
cd sandbox2
dune build @runtest
```
Note: this is a repo with just one ml file and dune files, cut down from the original repo where I had the issue

## Specifications

```
$ dune --version
3.7.0
$ ocamlopt.opt --version
4.14.1
$ head -n1 /etc/os-release
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
```

## Additional information 

- https://gist.github.com/smuenzel/6b84b94a500fa148a40a3b814215c6f8

Alizter:
We forgot to record the source files as part of the action which was causing them to not be added to the sandbox. This fix adds them and the example in the original report becomes resolved. I was unable to reproduce this issue in a cram test so it was tested manually.

fix #7376

### DIFF
--- a/src/dune_rules/cinaps.ml
+++ b/src/dune_rules/cinaps.ml
@@ -172,7 +172,11 @@ let gen_rules sctx t ~dir ~scope =
       Dep_conf_eval.unnamed ~sandbox ~expander t.runtime_deps
     in
     let* () = runtime_deps in
-    let+ () = Action_builder.path cinaps_exe in
+    let+ () =
+      Action_builder.deps
+        (Dep.Set.of_files
+           (cinaps_exe :: List.rev_map cinapsed_files ~f:Path.build))
+    in
     Action.Full.make ~sandbox
     @@ A.chdir (Path.build dir)
          (A.progn


src/dune_rules/cinaps.ml
```diff
@@ -172,7 +172,11 @@ let gen_rules sctx t ~dir ~scope =
       Dep_conf_eval.unnamed ~sandbox ~expander t.runtime_deps
     in
     let* () = runtime_deps in
-    let+ () = Action_builder.path cinaps_exe in
+    let+ () = Action_builder.path cinaps_exe
+    and+ () =
+      Action_builder.List.iter cinapsed_files ~f:(fun x ->
+          Action_builder.path (Path.build x))
```
rgrinberg, line 178: Simpler is to use `Action_builder.deps`.

rgrinberg: And you might as well include cinaps_exe in that set as well.

rgrinberg (approved):

Issue #7605, opened 2023-04-21
include_subdirs qualified should allow mapping folders to module names


anmonteiro:
## Desired Behavior

With `include_subdirs qualified`, dune should allow configuring the name of the namespaced module

## Example

A proposal:

```lisp
(include_subdirs 
  (mode qualified)
  ((src foo) (src/invalid-[module]-name invalid_module_name)))
```

Motivation here being next.js folder names https://beta.nextjs.org/docs/routing/defining-routes and natural-feeling interop with melange.

Alizter:
Maybe we can give the field a name:
```
(include_subdirs 
 (mode qualified)
 (rename
  (src foo)
  (src/invalid-[module]-name invalid_module_name)))
```

Alizter:
Of course if this gets implemented, you can for melange include a name mangling scheme which will auto map `(foo-bar)` to `foo_bar` as you wish.

rgrinberg:
Seems ok to me. As far as I can tell, this is the same as using `copy_files#` without the extra copying step.

One could also consider the same syntax use for install stanzas `(foo-bar as foo_bar)`.

PR #7620, opened 2023-04-23
Two tests for `(include_subdirs qualified)` and directories with invalid module names

Niols:
This PR adds two tests for the `(include_subdirs qualified)` stanza, specifically when this stanza is used while a directory whose name is an invalid module name is present (`baz-baz`). The two tests circumvent this problem in two different ways:

1. The first one adds a `(dirs :standard \ baz-baz)` stanza to hide this module from Dune. This works fine as of now.
2. The second one adds a `(data_only_dirs baz-baz)` stanza to specify that this directory should not be considered as a module but as a directory of data. This fails in the current state of Dune.

These two tests are meant as an illustration for https://github.com/ocaml/dune/pull/7601#issuecomment-1518682785, which might soon lead to a feature request.


anmonteiro:
related https://github.com/ocaml/dune/issues/7605

Niols:
Added a test where the directory with a malformed module name is nested in a directory with a valid name and that one is the one mentioned in `(data_only_dirs)`.

rgrinberg:
@anmonteiro could you lead the review on this one?

test/blackbox-tests/test-cases/include-qualified/data-only-dirs-malformed.t/run.t
```diff
@@ -0,0 +1,6 @@
+The presence of a directory whose name is not a valid module name should not
+clash with (include_subdirs qualified) if the directory in question is ignored
+by a `(dirs :standard \ ...)` stanza.
```
anmonteiro, line 3: the description in some of these tests doesn't match what's being tested. You mention using `(dirs..)` but then the test uses `data_only_dirs`.

test/blackbox-tests/test-cases/include-qualified/data-only-dirs-malformed.t/dune
```diff
@@ -0,0 +1,3 @@
+(include_subdirs qualified)
+(executable (name foo))
+(data_only_dirs baz-baz)
```
anmonteiro, line 3: What is this testing? I don't think this has any effect since there are no dune rules under `baz-baz`. (`data_only_dirs` ignores dune rules in the specified directories).

There are also no modules under that directory.

anmonteiro (commented):
I think the tests would be a bit more useful if they were testing the inclusion / exclusion of the directories in question.

Right now it's not clear whether the directories are being excluded because they're specified in `data_only_dirs` / `dirs` or because they have invalid folder names.

Niols:
> I think the tests would be a bit more useful if they were testing the inclusion / exclusion of the directories in question.

Sorry, I don't understand what you mean!

> Right now it's not clear whether the directories are being excluded because they're specified in `data_only_dirs` / `dirs` or because they have invalid folder names.

When I wrote these tests, Dune was failing whenever a directory with an invalid name was present in the scope. I assumed it was a feature and therefore the test was supposed to exhibit the fact that Dune should not fail when the directories with invalid names were included in a `data_only_dirs` or `dirs` statement. Now if it has been decided that Dune was not supposed to fail on those directories to begin with but to ignore them, then yes, these tests become rather useless. I haven't played with the new versions of Dune so I don't know where we stand on this.

test/blackbox-tests/test-cases/include-qualified/data-only-dirs-malformed.t/dune
Niols: I do think that if the directories contain no Dune rules and no modules, it would make sense for Dune to simply ignore them, making the tests in this PR irrelevant. At the time of writing this PR, though, a project of mine could not compile because of a situation like this, where there were no modules and no Dune rules yet Dune failed because of a directory with invalid name. I haven't tested on newer Dune version but I have just updated my branch and we'll see what the CI thinks of this.

Niols:
> I haven't played with the new versions of Dune so I don't know where we stand on this.

The CI seems to show that this is still the case, but maybe I misunderstood something?

anmonteiro:
To provide some closure here: `data_only_dirs` only means that dune will not read dune files in those directories, so these tests work as expected: `(include_subdirs qualified)` still indexes the modules in that directory.

PR #7621, opened 2023-04-24
melange: test dune rules subcommand

jchavarri:
Due to the way melange target dirs work, it is not possible to get `dune rules <target_folder>` to return the rules inside that folder, unless it is created manually. This PR adds a test to document that behavior.

### DIFF
--- a/test/blackbox-tests/test-cases/melange/dune-rules.t
+++ b/test/blackbox-tests/test-cases/melange/dune-rules.t
@@ -1,0 +1,71 @@
+Test dune rules
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.8)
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/melange/dune-rules.t
```diff
@@ -0,0 +1,71 @@
+Test dune rules
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.8)
+  > (using melange 0.1)
+  > EOF
+
+Using flags field in melange.emit stanzas is not supported
```
anmonteiro, line 8: this description doesn't make sense in this context?

PR #7625, opened 2023-04-24
Update monorepo benchmark to use external runner

gridbugs:
The benchmark runner is now located in:
https://github.com/ocaml-dune/ocaml-monorepo-benchmark/tree/main/dune-benchmark-runner

This updates the monorepo benchmark dockerfile to download and build the benchmark runner and makes some quality of life improvements to the dockerfile.

This also changes the benchmarking scenario to include watch-mode benchmarks.


snowleopard (approved):
I didn't really follow the changes here but they are just changing the benchmarking setup, so we should just merge and see if that works.

PR #7626, opened 2023-04-25
feature: build opam packages

rgrinberg:
Building opam packages with dune. This PR is highly incomplete. It's here to start the collaboration on this feature with the relevant contributors. It does need to be complete to be merged, only that @gridbugs and @Leonidas-from-XIV understand the high level ideas and how to make contributions to it.


src/dune_rules/context_env.ml
```diff
@@ -0,0 +1,225 @@
...
+    Memo.Lazy.create ~name:"context-opam" (fun () ->
+        Bin.which ~path:(Env_path.path Env.initial) "opam" >>= function
+        | None -> Utils.program_not_found "opam" ~loc:None
+        | Some opam -> (
+          let+ version =
+            Memo.of_reproducible_fiber
+              (Process.run_capture_line ~display:Quiet Strict opam
+                 [ "--version"; "--color=never" ])
+          in
+          match Scanf.sscanf version "%d.%d.%d" (fun a b c -> (a, b, c)) with
```
samoht, line 102: Will this break if opam returns `2.2.0-alpha1`?

samoht:
Do you have a description somewhere on how this work? :-)

It seems you are tracking the `.install` files separately now (in "cookies") - do I understand correctly that you have separate rules that turn that cookie into actual file moves + context environment updates (for the local variables)?

src/dune_rules/context_env.ml
Leonidas-from-XIV: No, `Scanf.sscanf` will happily ignore everything in the input after it has finished scanning. That also means that `2.2.0~alpha1` and `2.2.0` will return the exact same `(2, 2, 0)` tuple.

rgrinberg:
> Do you have a description somewhere of how this work? :-)

I'll add some comments in the code to describe this in a bit more detail. From a high level, we're setting up the following rules for every package:

* A source fetching rule that will populate `.pkg/$name/source`. This is not done yet, but is now possible with @Leonidas-from-XIV's work
* A build rule that populates `.pkg/$name/target`. This runs the build and install steps together, and generates the cookie file.

> It seems you are tracking the .install files separately now (in "cookies") 

Yeah, that's the basic idea. The cookie stores all the artifacts produced by the package (files and variables). I could do everything without the cookie, but it would require re-parsing the .install and config files. 

> do I understand correctly that you have separate rules that turn that cookie into actual file moves + context environment updates (for the local variables)?

We don't have those rules yet, but we will eventually. For now, I compose everything together to construct the final build sandbox with environment variables.

test/blackbox-tests/test-cases/pkg/installed-binary.t
```diff
@@ -0,0 +1,39 @@
...
+  >          "\| cat >test.install <<EOF
+  >          "\| bin: [ "foo" ]
+  >          "\| lib: [ "libxxx" ]
+  >          "\| lib_root: [ "lib_rootxxx" ]
+  >          "\| share_root: [ "lib_rootxxx" ]
+  >          "\| EOF
+  >  ))
+  > EOF
+
+  $ cat >dune.lock/usetest <<EOF
```
gridbugs, line 18: Is the idea that the lock file is actually a directory containing a file for each package that describes where to get the source and how to build that package?

rgrinberg: That's exactly it. A directory was chosen over a single file to make it easier to diff lock files.

gridbugs: We discussed this at the dune-dev meeting and I think it would be helpful to discus a bit more the tradeoff between multiple files and a single file. It seems like a single file sorted by package name would also be quite easy to diff and would have additional benefits such as simplifying working with multiple lockfiles (corresponding to multiple platforms or CI experiments for example) and would also simplify some git workflows compared to one file per package.

rgrinberg: Sorting helps, but it's easily confused when we have many packages with similar metadata.

>  such as simplifying working with multiple lockfiles (corresponding to multiple platforms or CI experiments for example)

Could you explain this point a bit more? I don't see any difference here.

>  and would also simplify some git workflows compared to one file per package.

Which git workflows?

gridbugs: > 

> Sorting helps, but it's easily confused when we have many packages with similar metadata.
> 
> > such as simplifying working with multiple lockfiles (corresponding to multiple platforms or CI experiments for example)
> 
> Could you explain this point a bit more? I don't see any difference here.

For example if we want to test with ocaml 4 and ocaml 5 in CI we might create one lockfile for each ocaml version and it seems messy to need one file for each combination of lockfile x package.

> 
> > and would also simplify some git workflows compared to one file per package.
> 
> Which git workflows?

One example would be when you add a new dependency this would create a new file and this wouldn't automatically get checked in when running `git commit -a` which is what I suspect a lot of people run.

Another issue we discussed yesterday which I forgot to mention is that if a project has many dependencies having one file per dependency could cause problems on systems with a maximum number of files per directory.

rgrinberg:
@nojb you can see the windows errors in CI:

```
D:\a\dune\dune\_boot\bitfn.h(1): error C2019: expected preprocessor directive, found '1'
```

test/blackbox-tests/test-cases/pkg/installed-binary.t
rgrinberg: > For example if we want to test with ocaml 4 and ocaml 5 in CI we might create one lockfile for each ocaml version and it seems messy to need one file for each combination of lockfile x package.

I still don't get it. We can create one lock "file" for each compiler version with a directory. One scheme would be `4.14.dune.lock/`, `5.0.dune.lock/` etc.

> One example would be when you add a new dependency this would create a new file and this wouldn't automatically get checked in when running git commit -a which is what I suspect a lot of people run.

Although I don't use git this way, I grant that this is an issue, but it seems like a minor inconvenience. While diffing is a big problem every time the lock file is updated.

> Another issue we discussed yesterday which I forgot to mention is that if a project has many dependencies having one file per dependency could cause problems on systems with a maximum number of files per directory.

On such systems, we're going to run into problems with cloning opam-repository first. Which has 4k entries in its packages/ directory.

nojb:
> @nojb you can see the windows errors in CI:

See #7679.

rgrinberg:
Okay, this work is getting big enough that it makes sense to start sending fixes in individual PR's. If anyone has any questions, feel free to make them here or in slack. I'll send subsequent PR's to fix various issues.

PR #7629, opened 2023-04-25
Two tests for interactions between `(include_subdirs qualified)` and `unqualified`

Niols:
This PR adds two tests where the `(include_subdirs qualified)` and `(include_subdirs unqualified)` stanzas are both present, one nested under the other. They serve as expectations for feature request https://github.com/ocaml/dune/issues/7630.


rgrinberg:
The tests are failing in CI so they need to be promoted.

@anmonteiro do you mind taking a look as well?

Niols:
The tests are failing because they exhibit an expected behaviour of Dune, which Dune does not have at the moment. Basically, they should start passing if https://github.com/ocaml/dune/issues/7630 is fixed.

emillon:
> The tests are failing because they exhibit an expected behaviour of Dune, which Dune does not have at the moment. Basically, they should start passing if #7630 is fixed.

The way we handle this is in that case is that the output should get promoted of a note about what is actually expected. That way the tests pass, but capture what error is expected. Then, once the issue is fixed, we'll see the output change in the test.

Niols:
> the output should get promoted of a note about what is actually expected

I am not exactly sure what you mean by this. Do you have an example test I could inspire myself from?

emillon:
You can run `make test` and then `make promote`. This will make sure that `make test` will pass. (this is what we mean by "the tests need to be promoted).

Niols:
Thanks! And what about the note of what is actually expected? Where should I put that?

emillon:
You can put it as a comment in the `run.t` files, for example after the output. Something like:

```
(the correct output would be baaaz baaazo beeez quuux)
```

Niols:
Thank you very much for your patience! What about this? i rebased on a more recent `main` (as of 3062a89f356bad20cb8de17fb4088b07f899cf6c), promoted the two tests and added a note on the “correct” output and a link to the feature request.

rgrinberg:
@anmonteiro do you plan to review this?

anmonteiro (approved):
This looks good, and seems like a bug. I added a commit per test, to show that Dune doesn't seem to see the modules in the directory `bar/` for both test cases.

PR #7639, opened 2023-04-26
melange: fix odoc

tatchi:
## Based on https://github.com/ocaml/dune/pull/7566

It fixes the failing test in #7566

### DIFF
--- a/src/dune_rules/odoc.ml
+++ b/src/dune_rules/odoc.ml
@@ -222,7 +222,7 @@ let module_deps (m : Module.t) ~obj_dir ~(dep_graphs : Dep_graph.Ml_kind.t) =
     List.map deps ~f:(fun m -> Path.build (Obj_dir.Module.odoc obj_dir m)))
 
 let compile_module sctx ~obj_dir (m : Module.t) ~includes:(file_deps, iflags)
-    ~dep_graphs ~pkg_or_lnu =
+    ~dep_graphs ~pkg_or_lnu ~mode =
   let odoc_file = Obj_dir.Module.odoc obj_dir m in
   let open Memo.O in
   let+ () =
@@ -238,7 +238,12 @@ let compile_module sctx ~obj_dir (m : Module.t) ~includes:(file_deps, iflags)
           ; Target odoc_file
           ; Dep
               (Path.build
-                 (Obj_dir.Module.cmti_file ~cm_kind:(Ocaml Cmi) obj_dir m))
+                 (Obj_dir.Module.cmti_file
+                    ~cm_kind:
+                      (match mode with
+                      | Lib_mode.Ocaml _ -> Ocaml Cmi
+                      | Melange -> Melange Cmi)
+                    obj_dir m))
           ]
       in
       let open Action_builder.With_targets.O in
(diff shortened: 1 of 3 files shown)


test/blackbox-tests/test-cases/melange/odoc.t
```diff
@@ -0,0 +1,42 @@
...
+
+  $ dune build @doc
+
+  $ cat _build/default/_doc/_mlds/foo/index.mld
+  {0 foo index}
+  {1 Library foo}
+  The entry point of this library is the module:
+  {!module-Foo}.
+
+Fails for Melange-only libraries
```
jchavarri, line 27: I think this comment needs to be updated.

src/dune_rules/odoc.ml
```diff
@@ -314,6 +319,12 @@ let setup_library_odoc_rules cctx (local_lib : Lib.Local.t) =
   let ctx = Super_context.context sctx in
   let* requires = Compilation_context.requires_compile cctx in
   let info = Lib.Local.info local_lib in
+  let modes = Lib_info.modes info in
+  let mode =
```
anmonteiro, line 323: this needs to be done the other way around:

- it's only `` `Melange `` if `modes.melange == true` and the OCaml modes are false.

src/dune_rules/odoc.ml
```diff
@@ -314,6 +319,12 @@ let setup_library_odoc_rules cctx (local_lib : Lib.Local.t) =
...
+    match modes.melange with
+    | false -> `Ocaml
+    | true -> `Melange
```
jchavarri, line 326: Dune code style guide recommends keeping scope of bindings [as small as possible](https://github.com/ocaml/dune/blame/418a4e7f17b7f2e491fb0ef775fc3a7277b267c7/doc/hacking.rst#L504-L505). 

https://github.com/ocaml/dune/blob/418a4e7f17b7f2e491fb0ef775fc3a7277b267c7/doc/hacking.rst#L504-L505


So these 2 bindings could be moved to line 337, inside the definition of `compiled` binding.

jchavarri (approved):
Looks good, thanks! 2 minor requests.

src/dune_rules/odoc.ml
anmonteiro: we have `Lib_mode.Map.Set.for_merlin`. Looks like we want to share the same logic.

tatchi: I ended up just re-using `Lib_mode.Map.Set.for_merlin`. Is that fine ?

anmonteiro: Yup! that's even better. I wonder if we should rename it now that it's not for merlin only.

anmonteiro (approved):
should be OK to rename in a follow-up PR in case you can't think of a better name right now.

test/blackbox-tests/test-cases/melange/odoc.t
```diff
@@ -0,0 +1,42 @@
+Test docs generation with Melange
```
rgrinberg, line 1: Could you add a dependency on the odoc package in this test?

Alizter (approved):

test/blackbox-tests/test-cases/melange/odoc.t
tatchi: Done, assuming I correctly understood what you were asking for

test/blackbox-tests/test-cases/melange/dune
```diff
@@ -9,3 +9,7 @@
 (cram
  (deps %{bin:melc_stdlib_prefix})
  (applies_to merlin))
+
+(cram
+ (deps (package odoc))
```
Alizter, line 14: @rgrinberg The odoc tests also rely on the odoc binary, so perhaps they should also be changed too.

rgrinberg: How do you know the tests only rely on the binary?

Issue #7164, opened 2023-02-22
Missing melange tests


anmonteiro:
This is meant to be a living list of tests I'd like to see for the melange integration before we document / ship it

- [ ] public library that depends on a private library, `melange.emit` stanza that depends on the public library
- [ ] include_subdirs + public libraries get correct `--bs-package-output` path

rizo:
Any chance https://github.com/ocaml/dune/issues/7104 could be considered before the official release? The described limitation currently prevents us from starting to incrementally adopt melange.

In particular, the described workflow with virtual libraries is essential for front-end/backend code sharing and compatibility with jsoo.

For instance, it would be convenient to define a virtual JSON-handling library with different backends for the native/byte mode and the melange mode.

PR #7652, opened 2023-04-27
test(melange): incorrect path when public lib depends on private lib

anmonteiro:
- This is the first "missing test case" from https://github.com/ocaml/dune/issues/7164
- it uncovers a path resolution bug when a public library (which gets installed into `node_modules`) depends on a private library.

I think this should be fine for installed libraries (I'll add a test too), but it's currently broken for libraries in the workspace.


rgrinberg (approved):

jchavarri (approved):

test/blackbox-tests/test-cases/melange/private-lib-dep.t
```diff
@@ -0,0 +1,89 @@
...
+  >  (libraries foo)
+  >  (emit_stdlib false))
+  > EOF
+
+  $ cat > entry.ml <<EOF
+  > let () = Js.log Foo.x
+  > EOF
+
+  $ dune build @melange --display=short
+          melc priv/.priv.objs/melange/priv.{cmi,cmj,cmt}
```
rgrinberg, line 43: Anything in particular that we need to see from this output?

anmonteiro: I removed `--display=short`

Issue #7104, opened 2023-02-17
Inconsistent behaviour for virtual libraries using melange mode


rizo:
## Expected Behavior

When creating a virtual library in native or byte modes, it is possible to assign a module to be virtual and add an implementation and interface that wraps around the virtual module in the virtual library.

Specifically the following structure illustrates this:

```
- vlib/
  - virt.mli     # virtual module
  - vlib.mli     # interface of vlib
  - vlib.ml      # implementation of vlib (uses virt!)
- vlib_impl_1/
  - virt.ml      # implementation of the virt module for vlib
```

This works as expected in native and byte modes and allows providing multiple implementations for `virt.mli` while offering a high-level interface in `vlib.mli`.

This doesn't work in melange mode.

## Actual Behavior

This doesn't work when `(modes melange)` is used. Not only it is not possible to replicate the described workflow, but having `(modes melange)` also prevents the non-melange implementation to be compiled.

Error when compiling a melange target:
```
$ dune build @melange
File "vlib/vlib.ml", line 1:        
Error: Virt not found, it means either the module does not exist or it is a namespace
```

Additionally, if `(modes melange)` is present in `vlib`, the native compilation fails with:
```
File "src/dune_rules/virtual_rules.ml", line 11, characters 30-37:
Error: No rule found for vlib/.vlib.objs/byte/virt.cmi
File "src/dune_rules/virtual_rules.ml", line 11, characters 30-37:
Error: No rule found for vlib/.vlib.objs/byte/vlib.cmi
File "src/dune_rules/virtual_rules.ml", line 11, characters 30-37:
Error: No rule found for vlib/.vlib.objs/byte/vlib.cmo
File "src/dune_rules/virtual_rules.ml", line 11, characters 30-37:
Error: No rule found for vlib/.vlib.objs/native/vlib.cmx
File "src/dune_rules/virtual_rules.ml", line 11, characters 30-37:
Error: No rule found for vlib/.vlib.objs/native/vlib.o
```

## Reproduction

The following repository fully replicates the above scenario based on the [blackbox test](https://github.com/ocaml/dune/tree/main/test/blackbox-tests/test-cases/melange/virtual_lib.t) included in dune:

- https://github.com/rizo/dune-melange-virt-issue

## Specifications

- Version of `dune`:  ~3.7 `05797cd3f505beca4d6d62c576506d7608a6d29d`
- Version of `melange`: `f4ddebbc7755af5e983c4df2ac51498f286e1074`
[...]

PR #7657, opened 2023-04-28
test(melange): melange public dir depending on private lib

anmonteiro:
- demonstrates 2 bugs:
1. the `.cmi` install location should be `melange/.public_cmi_melange/priv.cmi` rather than just `melange/.public_cmi_melange/priv.cmi`
2. an issue similar to #7104 where the `.cmj` for the private library isn't in the include paths, so we can't generate the .js file for it.

### DIFF
--- a/otherlibs/stdune/src/path.ml
+++ b/otherlibs/stdune/src/path.ml
@@ -1265,6 +1265,18 @@ let chmod t ~mode = Unix.chmod (to_string t) mode
 let follow_symlink path =
   Fpath.follow_symlink (to_string path) |> Result.map ~f:of_string
 
+let drop_prefix path ~prefix =
+  String.drop_prefix (to_string path) ~prefix:(to_string prefix)
+  |> Option.map ~f:(fun p ->
+         String.drop_prefix_if_exists ~prefix:"/" p |> Local.of_string)
+
+let drop_prefix_exn t ~prefix =
+  match drop_prefix t ~prefix with
+  | None ->
+    Code_error.raise "Path.drop_prefix_exn"
+      [ ("t", to_dyn t); ("prefix", to_dyn prefix) ]
+  | Some p -> p
+
 module Expert = struct
   let drop_absolute_prefix ~prefix p =
     match
--- a/otherlibs/stdune/src/path.mli
+++ b/otherlibs/stdune/src/path.mli
@@ -447,6 +447,14 @@ val chmod : t -> mode:int -> unit
 (** Attempts to resolve a symlink. Returns [None] if the path isn't a symlink *)
 val follow_symlink : t -> (t, Fpath.follow_symlink_error) result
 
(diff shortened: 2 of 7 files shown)


anmonteiro:
I pushed a fix for 1.

2. is a bit hairier and will probably need changes in melange similar to #7104

jchavarri (approved):
Nice finding!

src/dune_rules/install_rules.ml
```diff
@@ -205,7 +205,12 @@ end = struct
         let visibility = Module.visibility m in
         let dir' = Obj_dir.cm_dir external_obj_dir cm_kind visibility in
         if Path.equal (Path.build dir) dir' then None
-        else Path.basename dir' |> Option.some
+        else
+          String.drop_prefix (Path.to_string dir')
+            ~prefix:(Path.Build.to_string dir)
+          |> Option.value_exn
+          |> String.drop_prefix_if_exists ~prefix:"/"
```
rgrinberg, line 212: The fix is fine, but this little snippet of path handling code has been duplicated a couple of times already. Would you mind extracting it into a proper path handling function? It can be done in this PR with an additional commit, or in a separate PR.

anmonteiro: fixed this in the latest push

otherlibs/stdune/src/path.mli
```diff
@@ -447,6 +447,10 @@ val chmod : t -> mode:int -> unit
 (** Attempts to resolve a symlink. Returns [None] if the path isn't a symlink *)
 val follow_symlink : t -> (t, Fpath.follow_symlink_error) result
 
+val drop_prefix_exn : t -> prefix:t -> Local.t
```
Alizter, line 450:
```
Attempts to drop the user provided [~prefix] from a local path, including any leftover `/` prefix.
```
add comments to this and below

anmonteiro: good catch, updated.

otherlibs/stdune/src/path.ml
```diff
@@ -1265,6 +1265,18 @@ let chmod t ~mode = Unix.chmod (to_string t) mode
 let follow_symlink path =
   Fpath.follow_symlink (to_string path) |> Result.map ~f:of_string
 
+let drop_prefix path ~prefix =
+  String.drop_prefix (to_string path) ~prefix:(to_string prefix)
+  |> Option.map ~f:(fun p ->
+         String.drop_prefix_if_exists ~prefix:"/" p |> Local.of_string)
```
Alizter, line 1271: I'm not certain, but would it be better to use the path separator here? I'm wondering what happens with this code on windows.

rgrinberg: No, the current code is correct. We can use `/` on windows.

rgrinberg (approved):

PR #7663, opened 2023-05-01
fix(tui): support resizing

rgrinberg:
This PR allows the tui to be rerendered on resize events. This change is twofold:

* Allow handling sigwinch in the tui thread by blocking it in every other thread.
* Calling render whenever we get the winch.

Adding @snowleopard as a reviewer for the insignificant changes to scheduler to allow us to handle sigwinch.

### DIFF
--- a/src/dune_engine/scheduler.ml
+++ b/src/dune_engine/scheduler.ml
@@ -57,7 +57,12 @@ module Shutdown = struct
       | _ -> None)
 end
 
-let blocked_signals : Signal.t list = [ Int; Quit; Term ]
+(* These are the signals that will make the scheduler attempt to terminate dune *)
+let interrupt_signals : Signal.t list = [ Int; Quit; Term ]
+
+(* In addition, the scheduler also blocks some other signals so that only
+   designated threads can handle them by unblocking *)
+let blocked_signals : Signal.t list = Winch :: interrupt_signals
 
 module Thread : sig
   val spawn : signal_watcher:[ `Yes | `No ] -> (unit -> unit) -> unit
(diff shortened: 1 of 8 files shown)


Alizter (dismissed):

Alizter (approved):
Resizing now renders reliably.

rgrinberg:
ping @snowleopard

src/dune_engine/scheduler.ml
```diff
@@ -57,7 +57,9 @@ module Shutdown = struct
       | _ -> None)
 end
 
-let blocked_signals : Signal.t list = [ Int; Quit; Term ]
+let interrupt_signals : Signal.t list = [ Int; Quit; Term ]
+
+let blocked_signals : Signal.t list = Winch :: interrupt_signals
```
snowleopard, line 62: Could you add some comments on the difference between `interrupt_signals` and `blocked_signals`?

src/dune_threaded_console/dune_threaded_console.ml
```diff
@@ -135,7 +136,7 @@ let progress () =
       (* The current console doesn't react to user events so we just sleep until
           the next loop iteration. Because it doesn't react to user input, it cannot
           modify the UI state, and as a consequence doesn't need the mutex. *)
-      let handle_user_events ~now ~time_budget _ =
+      let handle_user_events ~now ~time_budget _mutex _state =
```
snowleopard, line 139: Could you add type signatures to ignored values? Just to make sure we notice if the type changes and name becomes stale.

snowleopard (approved):
Looks good but I left a couple of comments.

PR #7666, opened 2023-05-03
dune-rpc: use loopback address on Windows

nojb:
`0.0.0.0` is a dummy address that is interpreted in different ways by the various APIs. We should use the `loopback` (= `127.0.0.1`) address which always denotes the local host.

This is needed to use the RPC protocol on Windows.


nojb:
Note: conjunction with https://github.com/ocaml/ocaml-lsp/pull/1079, this is enough to get `ocaml-lsp`/dune rpc interaction working on Windows.

nojb:
@emillon This should go into 3.8.

rgrinberg (approved):

rgrinberg:
Needs a CHANGES entry + ocamlformat

PR #7667, opened 2023-05-03
fix(opam): remove bigarray dependency

rgrinberg:
bigarray has been moved to stdlib, so mentioning it as an external
library breaks bootstrap


Leonidas-from-XIV (approved):
Looks like our lowest supported version of OCaml is 4.08 and Bigarray was moved to Stdlib in 4.07, so this should work fine.

Alizter (approved):

PR #7668, opened 2023-05-03
Vendor opam-0install

### LINKED ISSUE #7641: Vendor the 0install solver
The 0install solver will be needed to generate lock files. To use it, we must vendor it in dune.

gridbugs:
Fixes https://github.com/ocaml/dune/issues/7641

This vendors the `opam-0install` package and all of its dependencies. I tested by adding a library dependency on `opam_0install` to bin/dune and bootstrapping dune in an empty opam switch.

I modified the ocaml-dune port of opam (https://github.com/ocaml-dune/opam/tree/vendored) to remove its dependency on spdx_licenses to reduce the number of packages we need to vendor.


gridbugs:
The error on windows is:
```
cd _boot && D:\.opam\dkml\bin\ocamlopt.opt.exe -o ../_boot\dune.exe -g -I +threads unix.cmxa threads.cmxa -no-alias-deps -w -49-6 readdir.c wait3_stubs.c platform_stubs.c copyfile_stubs.c dune_digest_stubs.c xdg_stubs.c dune_flock.c spawn_stubs.c dune_stats_stubs.c csexp_rpc_stubs.c fsevents_stubs.c inotify_stubs.c fswatch_win_stubs.c winsize.c sha512_stubs.c sha256_stubs.c sha1_stubs.c -args mods_list -cclib -lshell32 -cclib -lole32 -cclib -luuid -alert -unstable

D:\a\dune\dune\_boot\sha512.h(1): error C2019: expected preprocessor directive, found '1'

D:\a\dune\dune\_boot\bitfn.h(1): error C2019: expected preprocessor directive, found '1'

D:\a\dune\dune\_boot\util.h(1): error C2019: expected preprocessor directive, found '1'

make: *** [Makefile:49: _boot/dune.exe] Error 2
```

I accidentally added:
```ocaml
; ("vendor/sha", None, false, None)
```
to boot/libs.ml which isn't part of this change. I've removed it and I'll see if that fixes the problem.

Should I update boot/libs.ml at all? I see that it gets updated automatically when running `dune build` after vendoring a new package.

rgrinberg:
> Should I update boot/libs.ml at all? I see that it gets updated automatically when running dune build after vendoring a new package.

It should be updated automatically for you.

rgrinberg:
> D:\a\dune\dune\_boot\sha512.h(1): error C2019: expected preprocessor directive, found '1'

Where is this `1` coming from? I see it in the file name but why does it appear there?

vendor/fmt/src/dune
```diff
@@ -0,0 +1,5 @@
+(library
+ (name fmt)
+ (modules fmt)
+ (flags :standard -w -3-6-27-34-50)
```
rgrinberg, line 4: How come we need the flags? We ignore all warnings in the vendored dir anyway.

vendor/fmt/src/dune
```diff
@@ -0,0 +1,5 @@
+(library
+ (name fmt)
+ (modules fmt)
```
rgrinberg, line 3: Don't think we need the `modules` field.

vendor/fmt/src/dune
gridbugs: Ah these are copied from the dune file (from the dune-universe port). Will remove

gridbugs:
> > Should I update boot/libs.ml at all? I see that it gets updated automatically when running dune build after vendoring a new package.
> 
> It should be updated automatically for you.

I mean should I check it in?

gridbugs:
> > D:\a\dune\dune_boot\sha512.h(1): error C2019: expected preprocessor directive, found '1'
> 
> Where is this `1` coming from? I see it in the file name but why does it appear there?

Not sure what's going on there.

rgrinberg:
Yes, it should be checked in.

@nojb does that error look familiar?

nojb:
> @nojb does that error look familiar?

The error seems to have gone away.

gridbugs:
The problem went away because I removed
```ocaml
; ("vendor/sha", None, false, None)
```
from boot/libs.ml, but that line will get added back automatically when running `dune build`. It's not related to this change, so how about we merge this (assuming it's otherwise fine) and I'll make an issue to fix the problem. I don't have ocaml set up on my windows machine so I can't reproduce the problem locally.

nojb:
> but that line will get added back automatically when running `dune build`.

I cannot reproduce this. Can you provide specific instructions that will make this happen?

gridbugs:
Take a look at the failing tests on https://github.com/ocaml/dune/pull/7673

vendor/update-fmt.sh
```diff
@@ -0,0 +1,28 @@
...
+    cd $TMP
+    git clone https://github.com/ocaml-dune/$PACKAGE.git
+    cd $PACKAGE
+    git checkout $version
+)
+
+SRC=$TMP/$PACKAGE
+
+cp -v $SRC/LICENSE.md $PACKAGE/
+cp -v -R $SRC/src/{fmt.ml{i,},dune} $PACKAGE/src/
```
Leonidas-from-XIV, line 25: If we use our own dune file, there is no need to copy `dune` out of here.

Leonidas-from-XIV (commented):
Mostly looked at the update scripts.

gridbugs:
> > Take a look at the failing tests on #7673
> 
> I had seen the error. What I cannot reproduce is to get Dune to somehow add `vendor/sha` to `boot/libs.ml`.

I can't reproduce it anymore either. I re-created the local opam switch I use for working on dune to get a minimal environment to test this change and since then running `dune build` doesn't automatically update boot/libs.ml.

rgrinberg:
Note that you should be able to reproduce it if you add the dependency library to the dependency closure of the dune binary (by adding it as a dependency in dune_rules, main, etc). This is what we use to scan and populate boot/libs.ml

gridbugs:
ohh that explains what I'm seeing. I had added `opam_0install` as a dependency of `bin/dune` to test out my changes locally but I didn't check in that change, but that would explain why it updated boot/libs.ml, and why I now see an automatic change to boot/libs.ml that removes all the lines added in this PR now that I've removed that dependency. I'll take out the commit that updates boot/libs.ml from this PR since we aren't adding the dependency on opam_0install yet.

gridbugs:
> @gridbugs You probably want to add the `Dune_re` patch in opam-state in the large patch in `update-opam.sh` so rerunning the vendoring script doesn't undo the vendored changes like in #7672.

Yep good idea. I think I'll wait until https://github.com/ocaml/dune/pull/7672 is merged and then rebase on top and add to the patch to avoid dealing with merge conflicts.

gridbugs:
I've rebased on top of main since https://github.com/ocaml/dune/pull/7672 was merged and added to the opam patch so `Dune_rpc` gets added to the newly-added files.

Anything else that needs changing here?

Leonidas-from-XIV:
Looks ok from my side.

vendor/0install-solver/src/solver/dune
```diff
@@ -0,0 +1,2 @@
+(library
+ (name zeroinstall_solver))
```
Alizter, line 2: Do we want to call it `dune_zeroinstall_solver` like we do for other vendored libs?

gridbugs: We don't seem to consistently apply that convention to all our vendored libs. What's the reason for renaming them with the `dune_` prefix?

rgrinberg: it's explained here https://github.com/ocaml/dune/pull/7179#issuecomment-1497848152

It's done on a per-need basis. If it's not causing conflicts, then let it be as is.

gridbugs: Ah cool. Would it be helpful to add that comment or words to the same effect to a readme file in dune's vendor directory so it's more clear what to do next time someone needs to vendor something?

PR #7672, opened 2023-05-03
Inline patch for local changes to OPAM

Leonidas-from-XIV:
Mostly about aliasing `Dune_re` that we can't do on our vendored branch of opam, lest it stops building there.

Reported by @gridbugs.


gridbugs (approved):

rgrinberg:
LGTM. Another way of working around this is to have per file flags to add `-open Foo` wherever we need, and then override `Dune_re` there. That's quite a bit involved though.

PR #7674, opened 2023-05-03
chore: make undeclared modules check an error

Alizter:
Carrying on from #7608, we make the undeclared modules check an error.
- [x] test
- [x] changelog

### DIFF
--- a/src/dune_rules/modules_field_evaluator.ml
+++ b/src/dune_rules/modules_field_evaluator.ml
@@ -153,6 +153,7 @@ let check_invalid_module_listing
   ~existing_virtual_modules
   ~allow_new_public_modules
   ~is_vendored
+  ~version
   =
   let { errors; unimplemented_virt_modules } =
     find_errors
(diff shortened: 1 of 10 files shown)


src/dune_rules/modules_field_evaluator.ml
```diff
@@ -229,10 +229,9 @@ let check_invalid_module_listing ~stanza_loc ~modules_without_implementation
       [ Pp.text "You must provide an implementation for all of these modules." ];
     (* Checking that (modules) includes all declared modules *)
     let print_undelared_modules field mods =
-      (* TODO: this is a warning for now, change to an error in 3.9. *)
       (* If we are in a vendored stanza we do nothing. *)
       if not is_vendored then
-        print ~is_error:false
+        print
```
emillon, line 234: NB this should be ~is_error:(lang >= 3.9)

Alizter: I don't have access to the version so easily. What exactly is the point of version checking here? We did a warning stage for a version and now we wish to make it an error. What do we gain by allowing for the wrong configuration?

emillon: It would break some projects, in the sense that upgrading the dune package would suddenly make them not build anymore.

Alizter:
@emillon @rgrinberg I would like this to be in 3.9, but I am unsure how to do the versioning. I would appreciate any help.

emillon:
If you can't thread the version argument back to the parsing code (looks like it would be difficult), I would recommend trying to do that check at parse time or just after.

Alizter:
@emillon I'm not sure how we can do the check at parse time due to the way the code is structured. I guess I can leave it as a warning for the time being until we work it out.

Alizter:
@emillon I've managed to thread it through. Waiting for #8119.

Alizter:
@anmonteiro I had to add a dune_version field to the melange stanza. We do the same for library and executable anyway.

Alizter:
@emillon I've added tests to this demonstrating the version guarding behaviour and I have added a changelog entry. It is ready to merge now.

Alizter:
@emillon No chance for this to be in 3.10 then?

PR #7675, opened 2023-05-03
feature(pkg): Better API for Fetch

Leonidas-from-XIV:
As discussed with @rgrinberg, this

  * Introduces a checksum type
  * Verifies the checksums
  * Adds better error handling

Unfortunately we can't use the OPAM verification because OPAM verifies the file and deletes it immediately on checksum mismatches, without ever propagating the checksum out. I'll talk to OPAM maintainers about improving the API to include better error handling but for now to get the actual checksum we have to calculate it on our side.

### DIFF
--- a/src/dune_pkg/checksum.mli
+++ b/src/dune_pkg/checksum.mli
@@ -1,0 +1,16 @@
+open Import
+
+(** [t] represents a checksum for a file that is to be fetched or has been
+    fetched already. *)
+type t
+
+include Stringlike_intf.S with type t := t
+
+(** [to_opam_hash c] converts [c] to the representation of has values used by
+    OPAM *)
+val to_opam_hash : t -> OpamHash.t
+
+(** [of_opam_hash h] converts [h] from the representation used by OPAM to Dune's *)
+val of_opam_hash : OpamHash.t -> t
+
+val pp : t -> 'a Pp.t
--- a/src/dune_pkg/fetch.ml
+++ b/src/dune_pkg/fetch.ml
@@ -63,20 +63,32 @@ module Fiber_job = struct
     run1
 end
 
-let fetch loc url ~target =
+type failure =
+  | Checksum_mismatch of Checksum.t
+  | Unavailable of User_message.t option
+
+let fetch ~checksum ~target url =
   let open Fiber.O in
   let path = Path.to_string target in
   let fname = OpamFilename.of_string path in
-  let hashes = [] in
   let label = "dune-fetch" in
-  let job = OpamRepository.pull_file label fname hashes [ url ] in
+  (* hashes have to be empty otherwise OPAM deletes the file after
+     downloading if the hash does not match *)
+  let job = OpamRepository.pull_file label fname [] [ url ] in
   let+ downloaded = Fiber_job.run job in
   match downloaded with
-  | Up_to_date () | Result () -> ()
-  | Not_available (normal, _verbose) ->
-    User_error.raise ~loc
-      ([ Pp.textf "%s not available" (OpamUrl.to_string url) ]
-      @
-      match normal with
-      | None -> []
-      | Some normal -> [ Pp.text normal ])
+  | Up_to_date () | Result () -> (
+    match checksum with
+    | None -> Ok ()
+    | Some expected -> (
+      let expected = Checksum.to_opam_hash expected in
+      match OpamHash.mismatch path expected with
+      | None -> Ok ()
+      | Some actual ->
+        (* the file is invalid, so remove it before returning to the user *)
+        Path.unlink target;
+        Error (Checksum_mismatch (Checksum.of_opam_hash actual))))
+  | Not_available (None, _verbose) -> Error (Unavailable None)
+  | Not_available (Some normal, verbose) ->
+    let msg = User_message.make [ Pp.text normal; Pp.text verbose ] in
+    Error (Unavailable (Some msg))
--- a/src/dune_pkg/fetch.mli
+++ b/src/dune_pkg/fetch.mli
@@ -1,5 +1,19 @@
 open Stdune
 
-(** [fetch loc url ~target] will fetch [url] into [target]. If there's an error
-    fetching, [loc] will be used as the location for the error. *)
-val fetch : Loc.t -> OpamUrl.t -> target:Path.t -> unit Fiber.t
+type failure =
+  | Checksum_mismatch of Checksum.t
+  | Unavailable of User_message.t option
+
+(** [fetch ~checksum ~target url] will fetch [url] into [target]. It will verify
+    the downloaded file against [checksum], unless it [checksum] is [None].
+
+    @raise Checksum_mismatch
+      When the downloaded file doesn't match the expected [checksum], this will
+      pass the actually computed checksum.
+    @raise Unavailable
+      When the file can't be retrieved, e.g. not available at the location. *)
+val fetch :
+     checksum:Checksum.t option
+  -> target:Path.t
+  -> OpamUrl.t
+  -> (unit, failure) result Fiber.t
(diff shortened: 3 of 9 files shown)


test/expect-tests/dune_pkg/dune_pkg_unit_tests.ml
```diff
@@ -85,3 +114,36 @@ let%expect_test "downloading simple file" =
...
+       ~checksum:wrong_checksum);
+  Thread.join server;
+  print_endline "Finished successfully?";
+  [%expect.unreachable]
+  [@@expect.uncaught_exn
+    {|
+  (Dune_util__Report_error.Already_reported)
+  Trailing output
+  ---------------
+  Error: Checksum was wrong, is actually "md5=cbe78b067d4739684e86edfd2cb518bd" |}]
```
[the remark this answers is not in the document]
Alizter: Also mentioning the wrong checksum might be good.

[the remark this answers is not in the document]
Alizter: I guess we will also make a candidate for promotion here eventually.

test/expect-tests/dune_pkg/dune_pkg_unit_tests.ml
```diff
@@ -46,21 +68,28 @@ let run thunk =
...
   Scheduler.Run.go config ~on_event thunk
 
-let random_port state =
+let random_port () =
+  let state = Base.Random.State.make_self_init ~allow_in_tests:true () in
   (* ephemeral port range that is not assinged by IANA to prevent collisions *)
   Base.Random.State.int_incl state 49152 65535
 
-let%expect_test "downloading simple file" =
+let setup () =
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Originally I wanted to initialize the RNG state, but then I just decided to initialize it every time in a better helper function. Inlined.

src/dune_pkg/fetch.mli
```diff
@@ -13,7 +13,7 @@ type failure =
     @raise Unavailable
       When the file can't be retrieved, e.g. not available at the location. *)
 val fetch :
-     ?checksum:Checksum.t
+     checksum:Checksum.t option
```
Alizter, line 16: The comment needs to be updated too.

test/expect-tests/dune_pkg/dune_pkg_unit_tests.ml
```diff
@@ -48,8 +48,14 @@ let download ~port ~filename ~target ?checksum () =
...
   | Error (Checksum_mismatch actual_checksum) ->
+    let expected_checksum =
+      match checksum with
+      | Some v -> v
+      | None -> assert false
+    in
     User_error.raise ~loc:Loc.none
-      [ Pp.textf "Checksum was wrong, is actually %S"
+      [ Pp.textf "Expected checksum was\n%S\n, but got\n%S"
+          (Checksum.to_string expected_checksum)
```
Alizter, line 58: Rather than adding newlines you can just add elements to the list.

Alizter: and you can remove the comma in the explanation so it looks like:
```
Expected checksum:
asdf
Actual checksum:
asdf
```

src/dune_pkg/import.ml
```diff
@@ -0,0 +1,4 @@
+include Stdune
+open Dune_util
```
Alizter, line 2: We probably don't need to do this. Having the module qualified makes it easier to track what is coming from where.

Alizter (approved):

test/expect-tests/dune_pkg/dune_pkg_unit_tests.ml
```diff
@@ -27,12 +28,40 @@ let url ~port ~filename =
...
+      | Some s -> Pp.textf "Opam says: %S" s :: errs
+    in
+    User_error.raise ~loc:Loc.none errs
+  | Error (Checksum_mismatch actual_checksum) ->
+    let expected_checksum =
+      match checksum with
+      | Some v -> v
+      | None -> assert false
+    in
+    User_error.raise ~loc:Loc.none
```
rgrinberg, line 56: You should probably take the location of the checksum as an argument. In fact, the type of `?checksum` should be `(Loc.t * Checksum.t) option`

test/expect-tests/dune_pkg/dune_pkg_unit_tests.ml
```diff
@@ -27,12 +28,40 @@ let url ~port ~filename =
...
-  print_endline "Done downloading";
-  Fiber.return ()
+  let* res = Fetch.fetch ~checksum ~target url in
+  match res with
+  | Error (Unavailable (_url, normal)) ->
+    let errs = [ Pp.text "Failure while downloading" ] in
+    let errs =
+      match normal with
+      | None -> errs
+      | Some s -> Pp.textf "Opam says: %S" s :: errs
```
rgrinberg, line 47: Does that mean we're quoting the entire error message from opam? Doesn't seem right.

test/expect-tests/dune_pkg/dune_pkg_unit_tests.ml
```diff
@@ -27,12 +28,40 @@ let url ~port ~filename =
...
+    in
+    User_error.raise ~loc:Loc.none errs
```
rgrinberg, line 49: Here, the location should be from the url we're fetching.

src/dune_pkg/fetch.ml
```diff
@@ -63,20 +63,28 @@ module Fiber_job = struct
...
+  | Up_to_date () | Result () -> (
+    match checksum with
+    | None -> Ok ()
+    | Some expected -> (
+      let expected = Checksum.to_opam_hash expected in
+      match OpamHash.mismatch path expected with
+      | None -> Ok ()
+      | Some actual ->
+        (* We don't have a standard remove command, this is what OPAM uses *)
+        OpamFilename.remove fname;
```
rgrinberg, line 88: I think we can just `Path.unlink` here.

src/dune_pkg/fetch.ml
```diff
@@ -63,20 +63,28 @@ module Fiber_job = struct
     run1
 end
 
-let fetch loc url ~target =
+type failure =
+  | Checksum_mismatch of Checksum.t
+  | Unavailable of OpamUrl.t * string option
```
rgrinberg, line 68: The url here is always the input, right? If so, we don't need to return it. So what we're left with is an error message. We represent error messages with `User_message.t` so I would just use that.

src/dune_pkg/checksum.mli
```diff
@@ -0,0 +1,14 @@
...
+type t
+
+include Stringlike_intf.S with type t := t
+
+(** [to_opam_hash c] converts [c] to the representation of has values used by
+    OPAM *)
+val to_opam_hash : t -> OpamHash.t
+
+(** [of_opam_hash h] converts [h] from the representation used by OPAM to Dune's *)
+val of_opam_hash : OpamHash.t -> t
```
rgrinberg, line 14: We could also just make this type opaque and allow users to see that `Checksum.t = Opamhash.t` I don't really have a preference here

test/expect-tests/dune_pkg/dune_pkg_unit_tests.ml
```diff
@@ -39,14 +39,12 @@ let download ~port ~filename ~target ?checksum () =
...
+  | Error (Unavailable None) ->
     let errs = [ Pp.text "Failure while downloading" ] in
-    let errs =
-      match normal with
-      | None -> errs
-      | Some s -> Pp.textf "Opam says: %S" s :: errs
-    in
     User_error.raise ~loc:Loc.none errs
+  | Error (Unavailable (Some msg)) ->
+    User_message.print msg;
```
Alizter, line 46: Does this not also fail?

src/dune_pkg/fetch.ml
Leonidas-from-XIV: That might be a point, my main reasoning was that the `Not_found` exception is always very annoying because it doesn't specify what wasn't found. But exceptions are a little different than constructors.

PR #7677, opened 2023-05-03
dune_util: record exceptions to the _build/log

Alizter:
There are times when the detailed error becomes unavailable to view in the terminal such as when working with tui. Therefore recording them to the _build/log can help record what went wrong.

I've tested this and it works.


Alizter:
Hmm needs some more work.

Issue #3431, opened 2020-04-28
Opam constraints operator precedence is not honored.


mefyl:
A `depends` stanza such as the following one in the `dune-project` file will output a wrong OPAM constraint expression because logical operator precedence seems to be ignored.

```
  (alcotest
   (and
    (or :with-test :dev)
    (>= 1.1.0)))
```

## Expected Behavior

```"alcotest" { (with-test | dev) & >= "1.1.0"}```

## Actual Behavior

```"alcotest" {with-test | dev & >= "1.1.0"}```

## Reproduction

`dune build` any project with a `dune-project` looking like

```
(lang dune 2.2)

(name coin)

(generate_opam_files true)

(package
 (name coin)
 (synopsis "COIN")
 (depends
  (alcotest
   (or
    (and (>= 1.1.0) :with-test)
    (and (>= 1.1.0) :dev))))
```

and look at the generated coin.opam.

## Specifications

- Version of `dune`: 2.5.1
- Version of `ocaml`: 4.06.0
- Operating system : `Linux 5.4.28-gentoo`

PR #7682, opened 2023-05-04
`opam_constraint t` result should respect precedence

Lucccyo:
`add_group` appends parenthesis to the result of `opam_constraint` in order to handle precedence of `opamBaseParser.mly` in `opam-file-format` (https://github.com/ocaml/opam-file-format).

Fixes [#3431](https://github.com/ocaml/dune/issues/3431)

### DIFF
--- a/src/dune_rules/package.ml
+++ b/src/dune_rules/package.ml
@@ -250,25 +250,50 @@ module Dependency = struct
     <|> let+ name = Name.decode in
         { name; constraint_ = None }
 
-  let rec opam_constraint : Constraint.t -> OpamParserTypes.FullPos.value =
-    let open OpamParserTypes.FullPos in
-    function
-    | Bvar v -> Constraint.Var.to_opam v
-    | Uop (op, x) ->
-      nopos (Prefix_relop (Op.to_relop op, Constraint.Var.to_opam x))
-    | Bop (op, x, y) ->
-      nopos
-        (Relop
-           (Op.to_relop op, Constraint.Var.to_opam x, Constraint.Var.to_opam y))
-    | And [ c ] -> opam_constraint c
-    | And (c :: cs) ->
-      nopos (Logop (nopos `And, opam_constraint c, opam_constraint (And cs)))
-    | Or [ c ] -> opam_constraint c
-    | Or (c :: cs) ->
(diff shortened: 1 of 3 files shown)


emillon:
I rebased and marked this ready.

src/dune_rules/package.ml
```diff
@@ -250,25 +250,49 @@ module Dependency = struct
...
+      | Constraint.Bvar v -> Constraint.Var.to_opam v
+      | Uop (op, x) ->
+        nopos (Prefix_relop (Op.to_relop op, Constraint.Var.to_opam x))
+      | Bop (op, x, y) ->
+        nopos
+          (Relop
+             (Op.to_relop op, Constraint.Var.to_opam x, Constraint.Var.to_opam y))
+      | And cs -> List.map cs ~f:(opam_constraint Ctx_and) |> op_list `And
+      | Or cs ->
+        let inner_ctx = Ctx_or in
```
Alizter, line 285: This can be inlined.

Alizter (approved):

src/dune_rules/package.ml
emillon: The previous version had a bug explicitly because of a copy/paste between the `And` and `Or` branches so I prefer to keep the explicit name.

src/dune_rules/package.ml
```diff
@@ -250,25 +250,49 @@ module Dependency = struct
...
-    | Or [ c ] -> opam_constraint c
-    | Or (c :: cs) ->
-      nopos (Logop (nopos `Or, opam_constraint c, opam_constraint (And cs)))
-    | And [] | Or [] ->
+  type context =
+    | Root
+    | Ctx_and
+    | Ctx_or
+
+  (** The printer in opam-file-format does not insert parentheses on its own,
```
rgrinberg, line 258: Should just be a normal comment because it's not in the mli.

src/dune_rules/package.ml
rgrinberg: Could you lower the binding closer to the use-site though? It's easier to read when introduced bindings are used quickly.

emillon: I extracted a helper function which should address both concerns.

PR #7685, opened 2023-05-05
test(melange): merlin generation

jchavarri:
While testing some stuff in merlin for melange projects, I noticed that this explicit handling of `modes melange` has no apparent effect.

I believe that after the changes in https://github.com/ocaml/dune/pull/7362, this is not necessary because `melange` is just part of the closure.

Worst case my assumption is wrong, in that case I can revert the change and add a test that exercises this functionality.

### DIFF
--- a/src/dune_rules/merlin/merlin.ml
+++ b/src/dune_rules/merlin/merlin.ml
@@ -476,27 +476,6 @@ module Unprocessed = struct
           | [] -> None
           | stdlib_dir :: _ -> Some stdlib_dir)
       in
-      let* requires =
-        match t.config.mode with
-        | Ocaml _ -> Action_builder.return requires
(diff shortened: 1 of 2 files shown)


anmonteiro (commented):
Doesn’t this break libraries in a project without melange.emit?

jchavarri:
> Doesn’t this break libraries in a project without melange.emit?

@anmonteiro You're right. I [added a test to cover that case](https://github.com/ocaml/dune/pull/7685/commits/52209d1e9f348d47899988d81aa43e80447e2ab5) and [reverted the removal of merlin handling](https://github.com/ocaml/dune/pull/7685/commits/d8b1bd83f8f73e7ab4b5a35fb7b0758bb7631c55).

PR #7702, opened 2023-05-09
Add Dune debug/trace output

richardlford:
This PR adds new debug/trace output for Dune that
currently is activated by editing the Dune source
by changing some flags from false to true.

1. src/dune_engine/action_exec.ml Changes produce output in the log showing which
actions were executed.

2. src/dune_engine/action_to_sh.ml Modified to show the extension being executed.

3. src/dune_engine/build_system.ml Tracing the state changed to the build system.

4. src/dune_engine/build_system.ml Include the location of a rule in the to_dyn output.

5. src/memo/memo.ml Add memo_trace flag that enables extra output
to facilitate understanding the memo system.

### DIFF
--- a/src/dune_engine/action_exec.ml
+++ b/src/dune_engine/action_exec.ml
@@ -556,6 +556,39 @@ let exec
       with
       | false -> env
       | true ->
+        let debug = false in
+        (if debug then
+         match targets with
+         | None -> ()
+         | Some tgts -> (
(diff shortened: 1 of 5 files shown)


shonfeder:
I think the need for tracing and debug output has been addressed in a different rout via @rgrinberg's recent work. Please free to open this again if this is still to be advanced.

PR #7713, opened 2023-05-11
feature(pkg): parse patch/substitute actions

rgrinberg:
These actions are necessary to implement opam's corresponding features


Alizter (approved):
Tests?

PR #7716, opened 2023-05-12
test(melange): demonstrate bug with nested installed runtime assets

anmonteiro:
- When a library dune file is nested under the `dune-project` file, dune was writing information to `dune-package` about `melange.runtime_deps` relative to the root.
- these paths should, however, be recorded relative to the `dune` file that defines the library.

### DIFF
--- a/src/dune_rules/dune_package.ml
+++ b/src/dune_rules/dune_package.ml
@@ -49,8 +49,7 @@ module Lib = struct
       | External paths ->
         let lib_dir = Obj_dir.dir obj_dir in
         List.map paths ~f:(fun p ->
-            Path.as_in_build_dir_exn p |> Path.Build.drop_build_context_exn
-            |> Path.append_source lib_dir)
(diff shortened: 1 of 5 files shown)


rgrinberg:
Is there a similar bug with public headers?

anmonteiro:
It's possible. I'll write a test to verify. Would you like to see that in this PR or a separate one?

anmonteiro:
There's also a bug when "escaping" the dune file dir scope, which I'll send a fix for separately.

anmonteiro:
https://github.com/ocaml/dune/pull/7718 for the separate issue (also fixes public headers).

anmonteiro:
this PR now includes 4 commits:

1. demonstrates the original bug with `melange_runtime_deps`
2. fixes the `melange_runtime_deps` bug
3. demonstrates the same bug with public_headers
4. fixes the public_headers bug

rgrinberg (approved):

PR #7718, opened 2023-05-12
fix(runtime_deps): check paths are descendant of the library root

anmonteiro:
- add a new error when `melange.runtime_deps` in public libraries refer to paths outside of the library src dir

### DIFF
--- a/src/dune_rules/install_rules.ml
+++ b/src/dune_rules/install_rules.ml
@@ -43,6 +43,25 @@ module Package_paths = struct
     Path.Build.extend_basename (meta_file ctx pkg) ~suffix:".template"
 end
 
+let check_runtime_deps_relative_path local_path ~loc ~lib_info =
+  let lib_src_dir = Lib_info.src_dir lib_info in
+  match
+    Path.Local.descendant local_path ~of_:(Path.Build.local lib_src_dir)
+  with
+  | None ->
+    User_error.raise ~loc
+      [ Pp.textf
+          "Public library `%s' depends on assets outside its source tree. This \
+           is not allowed."
+          (lib_info |> Lib_info.name |> Lib_name.to_string)
+      ]
+      ~hints:
+        [ Pp.textf
+            "Move the dependency to a descendant of the folder where the \
+             library is defined"
+        ]
+  | Some _ -> ()
+
 module Stanzas_to_entries : sig
   val stanzas_to_entries :
     Super_context.t -> Install.Entry.Sourced.t list Package.Name.Map.t Memo.t
@@ -178,7 +197,12 @@ end = struct
           ~paths:(Disallow_external lib_name)
       in
       Path.Set.to_list_map deps ~f:(fun path ->
-          let path = Path.as_in_build_dir_exn path in
+          let path =
+            let path = path |> Path.as_in_build_dir_exn in
+            check_runtime_deps_relative_path ~lib_info:info ~loc
+              (Path.Build.local path);
+            path
+          in
           let sub_dir =
             let src_dir = Path.Build.parent_exn path in
             match Path.Build.equal lib_src_dir src_dir with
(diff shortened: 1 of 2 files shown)


src/dune_rules/install_rules.ml
```diff
@@ -43,6 +43,27 @@ module Package_paths = struct
     Path.Build.extend_basename (meta_file ctx pkg) ~suffix:".template"
 end
 
+let check_runtime_deps_relative_path path ~loc ~lib_info =
+  let lib_src_dir = Lib_info.src_dir lib_info in
+  match
+    Path.Local.descendant
+      (path |> Path.as_in_build_dir_exn |> Path.Build.local)
```
rgrinberg, line 50: How about have the caller do this cast instead?

test/blackbox-tests/test-cases/melange/installed-library-with-runtime-deps-nested-errors.t
```diff
@@ -0,0 +1,35 @@
+Test errors when public library runtime dependencies escape the dune file dir
```
rgrinberg, line 1: Can the test come be a separate commit?

src/dune_rules/install_rules.ml
anmonteiro: How would you suggest getting back the original path to return from this function? (currently line 65)

Alternative I can think of:
- make this function return `unit` and just check the path?

test/blackbox-tests/test-cases/melange/installed-library-with-runtime-deps-nested-errors.t
anmonteiro: did this.

src/dune_rules/install_rules.ml
rgrinberg: > make this function return unit and just check the path?

I think that would be even better.

anmonteiro: Ok I pushed that alternative.

src/dune_rules/install_rules.ml
```diff
@@ -43,6 +43,25 @@ module Package_paths = struct
...
 
+let check_runtime_deps_relative_path local_path ~loc ~lib_info =
+  let lib_src_dir = Lib_info.src_dir lib_info in
+  match
+    Path.Local.descendant local_path ~of_:(Path.Build.local lib_src_dir)
+  with
+  | None ->
+    User_error.raise ~loc
+      [ Pp.textf
+          "Public library %s depends on a path defined in a parent directory. \
```
jchavarri, line 54: Not sure if "in a parent directory" could be refined, because afaiu it doesn't necessarily need to be a parent directory.

Alternative: "Public library %s depends on an asset that is outside the library directory tree."

jchavarri (approved):
Looks great! Afaiu this constraints a bit more the way runtime deps work, right? From being able to depend on any runtime asset in a project, to be able to depend on any runtime asset in a library? I wonder if the docs should be updated to reflect this new constraint.

anmonteiro:
> this constraints a bit more the way runtime deps work, right?

the only additional constraint this imposes is that _public_ libraries need to have runtime_deps under their "lib root" (the folder where the dune file is).

src/dune_rules/install_rules.ml
```diff
@@ -178,7 +197,13 @@ end = struct
           ~paths:(Disallow_external lib_name)
       in
       Path.Set.to_list_map deps ~f:(fun path ->
-          let path = Path.as_in_build_dir_exn path in
+          let path =
+            let local_path =
+              path |> Path.as_in_build_dir_exn |> Path.Build.local
+            in
+            check_runtime_deps_relative_path ~lib_info:info ~loc local_path;
+            Path.as_in_build_dir_exn path
```
rgrinberg, line 205: I think calling `Path.as_in_build_dir_exn` should only be called once.

anmonteiro: Good point, just fixed that.

rgrinberg (approved):

src/dune_rules/install_rules.ml
anmonteiro: Agree. Tweaked the message based on this suggestion.

PR #7721, opened 2023-05-14
test: demonstrate byte_complete executable not printing backtrace

anmonteiro:
- test shows how `.exe` and `.bc` print the backtrace for exceptions, while `.bc.exe` doesn't.

### DIFF
--- a/test/blackbox-tests/test-cases/byte-complete-locations.t
+++ b/test/blackbox-tests/test-cases/byte-complete-locations.t
@@ -1,0 +1,32 @@
+Demonstrate byte_complete executables not printing backtraces
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.8)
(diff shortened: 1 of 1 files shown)


rgrinberg:
Does setting OCAMLRUNPARAM for byte_complete change anything?

cc @nojb

anmonteiro:
> Does setting OCAMLRUNPARAM for byte_complete change anything?

No, I also tried setting `OCAMLRUNPARAM=b`, to no avail.

nojb:
`byte_complete` executables do not contain the needed debug information to print backtraces. This is an issue that needs to be fixed upstream, there is nothing Dune can do about it. https://github.com/ocaml/ocaml/issues/9369

rgrinberg:
Thanks. I guess there's nothing that we can do then.

PR #7728, opened 2023-05-14
refactor: move library path handling to single to one place

rgrinberg:


src/dune_rules/lib_info.mli
```diff
@@ -19,6 +19,8 @@ module Status : sig
 
   (** For local libraries, return the project they are part of *)
   val project : t -> Dune_project.t option
+
+  val relative_to_package : t -> Lib_name.t -> Path.Local.t option
```
Alizter, line 23: Can you add a comment?

Alizter:
I don't get this change. Before in the private case something was being done, but now it looks like you just throw an exception?

PR #7730, opened 2023-05-15
fix(opam): cleanly reject argumentless (and)/(or)

emillon:
Otherwise they would raise a code error later.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -41,6 +41,9 @@ Unreleased
 
 - Attach melange rules to the default alias (#7926, @haochenx)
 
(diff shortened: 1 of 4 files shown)


Alizter:
We could give them meaning by setting an empty or to false and and empty and to true. However this would give us redundancy with saying true and false.

rgrinberg:
Can we guard this change behind 3.8?

rgrinberg:
But they are still parsed I think.

So this situation:

> if it is not enabled, I don't think that (depends) and the like are used

Would still cause an existing project to stop building.

emillon:
Ah, of course 🤦🏻

rgrinberg (approved):

Issue #7705, opened 2023-05-09
Integration with the Solver


rgrinberg:
The goal of this ticket is to use opam's solver to produce build plans and save them to our lock files. I imagine the following steps at moment, but this is all subject to change. Our focus should be to get the pipeline working at firs.

1. Add a command `dune pkg lock` (name is tentative) to run this step. This command should not build anything and therefore not acquire the build lock.

2. The command will produce or overwrite `dune.lock` with the build plan.

3. We'll need to write some conversion functions from opam's internal metadata into our lock file format. We don't yet support all of opam's features, so this will be incomplete.

4. Any configuration for this command such as constraints we pass to the solver should be stored in the dune-workspace file.

5. For now, we can piggy back on opam's repository handling. So to update the repository used as input for lock file generation, we need to run `opam update`. This is just a temporary step to make implementing this prototype easier.

PR #7732, opened 2023-05-15
Conservative lockfile generation

gridbugs:
This is a conservative implementation of generating lockfiles (lock directories) in dune. It adds a new command `dune pkg lock` that solves the dependencies of packages listed in dune-project and creates a lock directory containing the solution.

This is missing some features which I'll add in later changes. Opening this now to get early feedback.

 - Currently you must specify a path to a local checkout of the opam-repository. I'll have it default to using the current global opam installation's repo in a later change.
 - Also currently only the bare minimum package fields are added to lockfiles.
 - There's also no way to override opam variables other than clearing them completely. 
 - Need to add tests that we can round-trip encoding/decoding lockfiles

Relates to https://github.com/ocaml/dune/issues/7705 but doesn't completely resolve it.

### DIFF
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -9,20 +9,30 @@ module Source = struct
         ; checksum : (Loc.t * Checksum.t) option
         }
 
+  module Fields = struct
+    let copy = "copy"
+
+    let fetch = "fetch"
+
+    let url = "url"
+
+    let checksum = "checksum"
+  end
+
   let decode =
     let open Dune_lang.Decoder in
     sum
-      [ ( "copy"
+      [ ( Fields.copy
         , located string >>| fun (loc, source) path ->
           External_copy
             ( loc
             , if Filename.is_relative source then
                 Path.External.relative path source
               else Path.External.of_string source ) )
-      ; ( "fetch"
+      ; ( Fields.fetch
         , enter @@ fields
-          @@ let+ url = field "url" (located string)
-             and+ checksum = field_o "checksum" (located string) in
+          @@ let+ url = field Fields.url (located string)
+             and+ checksum = field_o Fields.checksum (located string) in
              let checksum =
                match checksum with
                | None -> None
@@ -33,6 +43,18 @@ module Source = struct
              in
              fun _ -> Fetch { url; checksum } )
       ]
+
+  let encode t =
+    let open Dune_lang.Encoder in
+    match t with
+    | External_copy (_loc, path) ->
+      constr Fields.copy string (Path.External.to_string path)
+    | Fetch { url = _loc, url; checksum } ->
+      record
+        [ (Fields.url, string url)
+        ; ( Fields.checksum
+          , (option Checksum.encode) (Option.map checksum ~f:snd) )
+        ]
 end
 
 module Pkg_info = struct
@@ -51,20 +73,25 @@ module Env_update = struct
     ; value : 'a
     }
 
+  let op_by_string =
+    [ ("=", OpamParserTypes.Eq)
+    ; ("+=", PlusEq)
+    ; ("=+", EqPlus)
+    ; (":=", ColonEq)
+    ; ("=:", EqColon)
+    ; ("=+=", EqPlusEq)
+    ]
+
   let decode =
     let open Dune_lang.Decoder in
-    let env_update_op =
-      enum
-        [ ("=", OpamParserTypes.Eq)
-        ; ("+=", PlusEq)
-        ; ("=+", EqPlus)
-        ; (":=", ColonEq)
-        ; ("=:", EqColon)
-        ; ("=+=", EqPlusEq)
-        ]
-    in
+    let env_update_op = enum op_by_string in
     let+ op, var, value = triple env_update_op string String_with_vars.decode in
     { op; var; value }
+
+  let encode { op; var; value } =
+    let open Dune_lang.Encoder in
+    let env_update_op = enum op_by_string in
+    triple env_update_op string String_with_vars.encode (op, var, value)
 end
 
 module Pkg = struct
@@ -77,17 +104,33 @@ module Pkg = struct
     ; exported_env : String_with_vars.t Env_update.t list
     }
 
+  module Fields = struct
+    let version = "version"
+
+    let install = "install"
+
+    let build = "build"
+
+    let deps = "deps"
+
+    let source = "source"
+
+    let dev = "dev"
+
+    let exported_env = "exported_env"
+  end
+
   let decode =
     let open Dune_lang.Decoder in
     enter @@ fields
-    @@ let+ version = field ~default:"dev" "version" string
-       and+ install_command = field_o "install" Dune_lang.Action.decode_pkg
-       and+ build_command = field_o "build" Dune_lang.Action.decode_pkg
-       and+ deps = field ~default:[] "deps" (repeat Package_name.decode)
-       and+ source = field_o "source" Source.decode
-       and+ dev = field_b "dev"
+    @@ let+ version = field ~default:"dev" Fields.version string
+       and+ install_command = field_o Fields.install Dune_lang.Action.decode_pkg
+       and+ build_command = field_o Fields.build Dune_lang.Action.decode_pkg
+       and+ deps = field ~default:[] Fields.deps (repeat Package_name.decode)
+       and+ source = field_o Fields.source Source.decode
+       and+ dev = field_b Fields.dev
        and+ exported_env =
-         field "exported_env" ~default:[] (repeat Env_update.decode)
+         field Fields.exported_env ~default:[] (repeat Env_update.decode)
        in
        fun ~lock_dir name ->
          let info =
@@ -99,12 +142,35 @@ module Pkg = struct
            { Pkg_info.name; version; dev; source }
          in
          { build_command; deps; install_command; info; exported_env; lock_dir }
+
+  let encode
+      { build_command
+      ; install_command
+      ; deps
+      ; info = { Pkg_info.name = _; version; dev; source }
+      ; lock_dir = _
+      ; exported_env
+      } =
+    let open Dune_lang.Encoder in
+    record_fields
+      [ field Fields.version string version
+      ; field_o Fields.install Dune_lang.Action.encode install_command
+      ; field_o Fields.build Dune_lang.Action.encode build_command
+      ; field_l Fields.deps Package_name.encode deps
+      ; field_o Fields.source Source.encode source
+      ; field_b Fields.dev dev
+      ; field_l Fields.exported_env Env_update.encode exported_env
+      ]
 end
 
 type t =
   { version : Syntax.Version.t
   ; packages : Pkg.t Package_name.Map.t
   }
+
+let create_latest_version packages =
+  let version = Syntax.greatest_supported_version Dune_lang.Pkg.syntax in
+  { version; packages }
 
 let path = Path.Source.(relative root "dune.lock")
 
@@ -113,3 +179,30 @@ let metadata = "lock.dune"
 module Metadata = Dune_sexp.Versioned_file.Make (Unit)
 
 let () = Metadata.Lang.register Dune_lang.Pkg.syntax ()
+
+let encode_metadata t =
+  let open Dune_lang.Encoder in
+  list sexp
+    [ string "lang"
+    ; string (Syntax.name Dune_lang.Pkg.syntax)
+    ; Dune_lang.Syntax.Version.encode t.version
+    ]
+
+let file_contents_by_path t =
+  (metadata, [ encode_metadata t ])
+  :: (Package_name.Map.to_list t.packages
+     |> List.map ~f:(fun (name, pkg) ->
+            (Package_name.to_string name, Pkg.encode pkg)))
+
+let write_disk ~lock_dir_path t =
+  let lock_dir_path = Path.source lock_dir_path in
+  Path.rm_rf lock_dir_path;
+  Path.mkdir_p lock_dir_path;
+  file_contents_by_path t
+  |> List.iter ~f:(fun (path_within_lock_dir, contents) ->
+         let path = Path.relative lock_dir_path path_within_lock_dir in
+         Option.iter (Path.parent path) ~f:Path.mkdir_p;
+         let contents_string =
+           List.map contents ~f:Dune_lang.to_string |> String.concat ~sep:"\n"
+         in
+         Io.write_file path contents_string)
--- a/src/dune_pkg/lock_dir.mli
+++ b/src/dune_pkg/lock_dir.mli
@@ -50,8 +50,12 @@ type t =
   ; packages : Pkg.t Package_name.Map.t
   }
 
+val create_latest_version : Pkg.t Package_name.Map.t -> t
+
 val path : Path.Source.t
 
 val metadata : Filename.t
 
 module Metadata : Dune_sexp.Versioned_file.S with type data := unit
+
+val write_disk : lock_dir_path:Path.Source.t -> t -> unit
--- a/src/dune_pkg/opam.ml
+++ b/src/dune_pkg/opam.ml
@@ -1,0 +1,218 @@
+open Stdune
+module Package_name = Dune_lang.Package_name
+
+module Repo = struct
+  let ( / ) = Filename.concat
+
+  type t = { packages_dir_path : Filename.t }
+
+  let validate_repo_file opam_repo_dir_path =
+    try
+      OpamFilename.raw (opam_repo_dir_path / "repo")
+      |> OpamFile.make |> OpamFile.Repo.read |> ignore
+    with OpamSystem.Internal_error message ->
+      User_error.raise [ Pp.text message ]
+
+  let of_opam_repo_dir_path opam_repo_dir_path =
+    if not (Sys.file_exists opam_repo_dir_path) then
+      User_error.raise
+        [ Pp.textf "%s does not exist" (String.maybe_quoted opam_repo_dir_path)
+        ];
+    if not (Sys.is_directory opam_repo_dir_path) then
+      User_error.raise
+        [ Pp.textf "%s is not a directory"
+            (String.maybe_quoted opam_repo_dir_path)
+        ];
+    let packages_dir_path = opam_repo_dir_path / "packages" in
+    if
+      not
+        (Sys.file_exists packages_dir_path && Sys.is_directory packages_dir_path)
+    then
+      User_error.raise
+        [ Pp.textf
+            "%s doesn't look like a path to an opam repository as it lacks a \
+             subdirectory named \"packages\""
+            (String.maybe_quoted opam_repo_dir_path)
+        ];
+    validate_repo_file opam_repo_dir_path;
+    { packages_dir_path }
+
+  (* Return the path to an "opam" file describing a particular package
+     (name and version) from this opam repository. *)
+  let get_opam_file_path t opam_package =
+    t.packages_dir_path
+    / OpamPackage.name_to_string opam_package
+    / OpamPackage.to_string opam_package
+    / "opam"
+
+  (* Reads an opam package definition from an "opam" file in this repository
+     corresponding to a package (name and version). *)
+  let load_opam_package t opam_package =
+    let opam_file_path = get_opam_file_path t opam_package in
+    if not (Sys.file_exists opam_file_path) then
+      User_error.raise
+        [ Pp.textf
+            "Couldn't find package file for \"%s\". It was expected to be \
+             located in %s but this file does not exist"
+            (OpamPackage.to_string opam_package)
+            (String.maybe_quoted opam_file_path)
+        ];
+    OpamFile.OPAM.read (OpamFile.make (OpamFilename.raw opam_file_path))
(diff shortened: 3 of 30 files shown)


CHANGES.md
```diff
@@ -159,6 +159,9 @@ Unreleased
 - Bump minimum version of the dune language for the melange syntax extension
   from 3.7 to 3.8 (#7665, @jchavarri)
 
+- Conservatime implementation of lockfile generation (#7732 partially fixes
```
christinerose, line 162:
```suggestion
- Conservatime implementation of lock file generation (#7732 partially fixes
```

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
```diff
@@ -0,0 +1,85 @@
+Simple example of generating a lockfile with dune
```
christinerose, line 1:
```suggestion
Simple example of generating a lock file with Dune
```

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
```diff
@@ -0,0 +1,85 @@
...
+
+Generate a dune-project file.
```
christinerose, line 3:
```suggestion
Generate a `dune-project` file.
```

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
```diff
@@ -0,0 +1,85 @@
...
+  dune.lock/lockfile_generation_test:
+  
+  (version LOCAL)
+  (deps foo bar)
+  (dev true)
+  
+  ---
+  
+
+Regenerate the dune-project file introducing an unsatisfiable constraint.
```
christinerose, line 62:
```suggestion
Regenerate the `dune-project` file introducing an unsatisfiable constraint.
```

christinerose (approved):
Minor suggestions to match style guide. Approving so I don't block the merge.

src/dune_rules/pkg_rules.mli
```diff
@@ -14,3 +14,58 @@ val setup_package_rules :
   -> dir:Path.Build.t
   -> pkg_name:string
   -> Build_config.gen_rules_result Memo.t
+
+module Source : sig
```
rgrinberg, line 18: Now that #7725 is merged, you shouldn't need to expose anything from here.

src/dune_rules/package.ml
```diff
@@ -101,28 +106,29 @@ module Dependency = struct
   module Constraint = struct
     module Var = struct
       type t =
-        | QVar of string
+        | Literal of string
         | Var of string
 
       let encode = function
-        | QVar v -> Dune_lang.Encoder.string v
+        | Literal v -> Dune_lang.Encoder.string v
```
rgrinberg, line 113: If possible, could you extract the rename into a separate PR?

gridbugs: Yep! https://github.com/ocaml/dune/pull/7733

.ocamlformat-ignore
```diff
@@ -1,2 +1,3 @@
 boot/libs.ml
 src/dune_rules/assets.ml
+vendor/**
```
Alizter, line 3: Can you make this a separate change? We already have an ocamlformat ignore in vendor, but it doesn't cover all the projects.

CHANGES.md
```diff
@@ -159,6 +159,9 @@ Unreleased
 - Bump minimum version of the dune language for the melange syntax extension
   from 3.7 to 3.8 (#7665, @jchavarri)
 
+- Conservative implementation of lock file generation (#7732 partially fixes
```
rgrinberg, line 162: Since we're developing this feature in private, we shouldn't tell the user about this (yet)

src/dune_pkg/opam.ml
```diff
@@ -0,0 +1,188 @@
...
+  ; info
+  ; lock_dir = lock_dir_path
+  ; exported_env = []
+  }
+
+let solve_package_list local_packages ~env ~repo =
+  let context = Solver_context.create ~env ~repo ~local_packages in
+  let result =
+    try Solver.solve context (OpamPackage.Name.Map.keys local_packages)
+    with exn ->
```
rgrinberg, line 162: Is this reachable at all? It seems weird that this function both raises and returns `result`.

src/dune_pkg/lock_dir.mli
```diff
@@ -50,8 +50,12 @@ type t =
...
 
+val create_latest_version : Pkg.t Package_name.Map.t -> t
+
 val path : Path.Source.t
 
 val metadata : Filename.t
 
 module Metadata : Dune_sexp.Versioned_file.S with type data := unit
+
+val instantiate_on_disk : lock_dir_path:Path.Source.t -> t -> unit
```
rgrinberg, line 61: How about `write_disk`? `instantiate` sounds like we're doing more than just some serialization.

src/dune_pkg/lock_dir.ml
```diff
@@ -99,17 +142,67 @@ module Pkg = struct
...
+      ; exported_env
+      } =
+    let open Dune_lang.Encoder in
+    record_fields
+      [ field Fields.version string version
+      ; field_o Fields.install Dune_lang.Action.encode install_command
+      ; field_o Fields.build Dune_lang.Action.encode build_command
+      ; field_l Fields.deps Package_name.encode deps
+      ; field_o Fields.source Source.encode source
+      ; field Fields.dev bool dev
```
rgrinberg, line 161: Just a stylistic choice, but I would omit fields when they're equal to their default value. All the defaults are controlled by the language version, so this is safe to do.

src/dune_pkg/opam.ml
```diff
@@ -0,0 +1,188 @@
...
+  let of_opam_repo_dir_path opam_repo_dir_path =
+    if not (Sys.file_exists opam_repo_dir_path) then
+      User_error.raise [ Pp.textf "\"%s\" does not exist" opam_repo_dir_path ];
+    if not (Sys.is_directory opam_repo_dir_path) then
+      User_error.raise
+        [ Pp.textf "\"%s\" is not a directory" opam_repo_dir_path ];
+    let packages_dir_path = opam_repo_dir_path / "packages" in
+    if
+      not
+        (Sys.file_exists packages_dir_path && Sys.is_directory packages_dir_path)
```
Leonidas-from-XIV, line 18: I think this should check for the `repo` file too, which as far as I am aware needs to contin at least `opam-version: "2.0"`.

src/dune_pkg/opam.ml
```diff
@@ -0,0 +1,188 @@
...
+    then
+      User_error.raise
+        [ Pp.textf
+            "\"%s\" doesn't look like a path to an opam repository as it lacks \
+             a subdirectory named \"packages\""
+            opam_repo_dir_path
+        ];
+    { packages_dir_path }
+
+  (* Return the path to an "opam" file describing a paticular package
```
Leonidas-from-XIV, line 28:
```suggestion
  (* Return the path to an "opam" file describing a particular package
```

src/dune_pkg/opam.ml
```diff
@@ -0,0 +1,188 @@
...
+      User_error.raise
+        [ Pp.textf "\"%s\" is not a directory" opam_repo_dir_path ];
+    let packages_dir_path = opam_repo_dir_path / "packages" in
+    if
+      not
+        (Sys.file_exists packages_dir_path && Sys.is_directory packages_dir_path)
+    then
+      User_error.raise
+        [ Pp.textf
+            "\"%s\" doesn't look like a path to an opam repository as it lacks \
```
rgrinberg, line 22: Should be `%s` and `Path.to_maybe_string_quoted opam_repo_dir_path` to be consistent with our other messages.

src/dune_pkg/opam.ml
```diff
@@ -0,0 +1,188 @@
...
+
+  let user_restrictions t = Dir_context.user_restrictions t.dir_context
+
+  let filter_deps t = Dir_context.filter_deps t.dir_context
+
+  let create ~env ~repo ~local_packages =
+    let env name = Env.find_by_name env ~name in
+    let { Repo.packages_dir_path } = repo in
+    let dir_context =
+      Dir_context.create ~prefer_oldest:true
```
Leonidas-from-XIV, line 103: So we're going with MVS?

src/dune_pkg/opam.ml
```diff
@@ -0,0 +1,188 @@
...
+
+module Summary = struct
+  type t = { opam_packages_to_lock : OpamPackage.t list }
+
+  let selected_packages_message t =
+    User_message.make
+      (Pp.tag User_message.Style.Success
+         (Pp.text "Selected the following packages:")
+      :: List.map t.opam_packages_to_lock ~f:(fun package ->
+             Pp.textf "%s" (OpamPackage.to_string package)))
```
Leonidas-from-XIV, line 119:
```suggestion
             Pp.text (OpamPackage.to_string package)))
```

src/dune_pkg/opam.ml
```diff
@@ -0,0 +1,188 @@
...
+  ; lock_dir = lock_dir_path
+  ; exported_env = []
+  }
+
+let solve_package_list local_packages ~env ~repo =
+  let context = Solver_context.create ~env ~repo ~local_packages in
+  let result =
+    try Solver.solve context (OpamPackage.Name.Map.keys local_packages)
+    with exn ->
+      User_error.raise [ Pp.textf "%s" (Exn.to_dyn exn |> Dyn.to_string) ]
```
Leonidas-from-XIV, line 163:
```suggestion
      User_error.raise [ Pp.text (Exn.to_dyn exn |> Dyn.to_string) ]
```

src/dune_pkg/opam.ml
```diff
@@ -0,0 +1,188 @@
...
+  in
+  match result with
+  | Error e -> User_error.raise [ Pp.textf "%s" (Solver.diagnostics e) ]
```
Leonidas-from-XIV, line 166:
```suggestion
  | Error e -> User_error.raise [ Pp.text (Solver.diagnostics e) ]
```

src/dune_pkg/opam.mli
```diff
@@ -0,0 +1,39 @@
+open Stdune
+
+module Repo : sig
+  (** An opam repository *)
+  type t
+
+  (** Create a [t] from a path to a local directory containing an opam
+      repository. This should be a directory with a subdirectory named
+      "packages" with one subdirectory for each package name, each with
+      subdirectories for each version. *)
```
Leonidas-from-XIV, line 10: I think stating "this should be a valid opam repository" saves us from defining what an opam repository is and potentially creating our own version of opam-repositories that don't work with OPAM.

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
```diff
@@ -0,0 +1,85 @@
...
+  >   ))
+  > EOF
+
+Run the solver and generate a lock directory.
+  $ dune pkg lock --opam-env=pure --opam-repository=mock-opam-repository
+  Selected the following packages:
+  bar.0.4.0
+  baz.0.1.0
+  foo.0.0.1
+  lockfile_generation_test.LOCAL
```
Leonidas-from-XIV, line 20: I am a bit confused why the lockfile would contain its own project. Am I missing something?

.ocamlformat-ignore
gridbugs: Ah I missed that file. Here's a PR that adds opam https://github.com/ocaml/dune/pull/7744

src/dune_pkg/opam.ml
gridbugs: `Solver.solve` returns `Error` when it can't find a solution to the dependencies but it can also raise exceptions, for example if opam is unable to parse an opam file in the package repo. I'll add a comment to this effect to clarify why this is needed.

src/dune_pkg/opam.mli
```diff
@@ -0,0 +1,39 @@
...
+  val of_opam_repo_dir_path : string -> t
```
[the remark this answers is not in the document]
gridbugs: The problem with making it a `Path.t` is that I'd like to be able to call this during argument parsing, but `Path.of_string` can't be called before `Common.init`. I thought about making it a `Path.External.t` but it's useful to support relative paths to the repo for testing purposes. I'll make it a `Filename.t` instead as that communicates the intent of this function but is less opinionated than the `Path` module, and this path just ends up getting passed to opam libraries which expect a string anyway.

src/dune_pkg/lock_dir.ml
gridbugs: This was an actual bug. `decode` decodes this field as a field with no argument (ie. `(dev)`) whereas `encode` was writing a field with a name and boolean value (`(dev true)`). Switching to using the `field_b` function to match the decoder. I still need to add tests that the encoder and decoder round trip.

src/dune_pkg/opam.ml
gridbugs: (using `String.maybe_quoted` which works for `Filename.t`s)

src/dune_pkg/opam.ml
gridbugs: Is it not sufficient to let opam check the `opam-version` in the `opam` files in the repo as it's parsing them? Or would this just be to improve UX in that we could catch the problem earlier and print a more meaningful error message than the error opam would produce while parsing files? I left this file out as it's not necessary to solve dependencies but it would be easy to add. Do you happen to know if the opam libraries provide an easy way to parse this file?

src/dune_pkg/opam.ml
gridbugs: Ah good question! MVS would be my preference but this merits further discussion. @rgrinberg thoughts?

src/dune_pkg/opam.mli
gridbugs: Good point. I'll update the comment and help message. This is also a good reason to enforce that a "repo" file is present inside the opam repository even if it doesn't contribute to solving.

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
gridbugs: Ah good point there is no reason for the packages defined locally to be present in the lockfile. I'll remove them.

src/dune_pkg/opam.ml
rgrinberg: It's a good default, but we could make this configurable.

src/dune_pkg/opam.mli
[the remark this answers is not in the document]
rgrinberg: Don't we have Path.Outside_build_dir.t for that?

src/dune_pkg/lock_dir.ml
rgrinberg: Hmm, well I think that `(dev true)` or `(dev false)` should still be allowed. I'll take a look

src/dune_pkg/opam.ml
rgrinberg: Seems like a weird choice for an API to me.

In any case, if some exceptions are expected we should list them here and have proper dedicated messages for them. Doing `Exn.to_dyn exn |> Dyn.to_string` is going to show something rather ugly to users.

Some tests for these exceptions would be quite handy as well.

src/dune_pkg/opam.mli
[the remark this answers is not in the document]
gridbugs: Ah I didn't realize we had that. Minor issue I ran into with `Path.Outside_build_dir.t` is that it assumes that relative paths are in the source dir, which means you can't do:
```
dune pkg lock --opam-repository-path=../opam-repository
```
when the opam repo lives next to the dune project directory.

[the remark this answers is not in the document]
gridbugs: More specifically, calling `Path.Outside_build_dir.of_string` on a relative path outside the project (like `"../opam-repository"`) gives the error:
```
Error: path outside the workspace: ../opam-repository from .
```
Is that intentional?

src/dune_pkg/opam.ml
gridbugs: Ok cool, I'll add a list of exceptions. Would it be reasonable to keep the `Exn.to_dyn exn |> Dyn.to_string` for exceptions not in the list though, as the alternative is that it will crash dune in this case? It will be hard to be sure we're handling all possible exceptions from opam, and we might update our opam port one day and find that it added a new exception without us noticing.

src/dune_pkg/opam.mli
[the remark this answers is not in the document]
rgrinberg: I guess it is. How annoying. It should convert it to an External path at that point.

Okay let's leave it as is for now.

src/dune_pkg/opam.ml
rgrinberg: I think it would be reasonable to ask opam (or the solver) to allow us to get a readable user message from any exception that might be user facing. 

> and we might update our opam port one day and find that it added a new exception without us noticing

That would mean we have a poor error message for an error state that is reachable by the user. Which is a bug our end and a `Code_error` would be appropriate.

gridbugs: Cool. For this PR I'll catch all the exceptions from opam that I'm aware of and produce a readable `User_error` for each, and then have a catch all that produces a `Code_error` indicating that there's a user reachable error state that we haven't accounted for.

src/dune_pkg/opam.ml
Leonidas-from-XIV: I would say we should require the same format for the repo as OPAM to not introduce dialects ("this repo only works with dune"/"this repo only works with OPAM"). If I see correctly, `opam-format` has `OpamFile.Repo` which defines accessors to the metadata that can be found in the `repo` file so I would say as long as it can parse it it's a valid repo.

src/dune_pkg/opam.ml
Leonidas-from-XIV: I think it is an interesting choice for the prototype and solves a lot of "lower bounds" failures that we frequently have in OPAM, so I'm in favor of trying and collecting feedback.

gridbugs:
Applied feedback from review. I still need to add tests for error cases when a user specifies an invalid opam repo. I'll do this next week.

rgrinberg:
@gridbugs this PR doesn't need to be perfect to be merged. Just let me know when you think it's at a good point and I can merge it.

gridbugs:
> @gridbugs this PR doesn't need to be perfect to be merged. Just let me know when you think it's at a good point and I can merge it.

Ok cool, I think let's merge this now to avoid future merge conflicts. I'll add tests that the expected error message is raised for invalid opam repos in a later PR.

rgrinberg:
There seems to be a relevant failure in the windows CI:

```
./_boot/dune.exe build @install -p dune --profile dune-bootstrap
Fatal error: exception Failure("Windows only. This function isn't implemented.")
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from OpamStateConfig.default in file "vendor/opam/src/state\\opamStateConfig.ml", line 86, characters 34-73
make: *** [Makefile:[23](https://github.com/ocaml/dune/actions/runs/5054113855/jobs/9068678005?pr=7732#step:8:24)5: bootstrap] Error 2
```

gridbugs:
Heads up that I pushed a change to this branch that hacks the `opam_state` library to lazily compute the default root directory to see if it fixes the crash on windows. If this works I'll push the change upstream to dune's opam port rather than leaving the change in dune.

rgrinberg:
> Heads up that I pushed a change to this branch that hacks the opam_state library to lazily compute the default root directory to see if it fixes the crash on windows. If this works I'll push the change upstream to dune's opam port rather than leaving the change in dune.

That's fine, but can we make sure that ocaml-dune/opam contains the commit that does this? Otherwise anyone that will pull will accidentally override.

gridbugs:
> Just a general comment: we can modify upstream libraries all we want and upstream things at our own pace. There's really no pressure to do so. But we should always make sure that every change has a corresponding commit in ocaml-dune/$repo.

Make sense. My change seemed to fix the problem. I've pushed the fix to https://github.com/ocaml-dune/opam/ and updated and re-run the update script for opam and squashed the result into the existing commit in this PR that already updated the opam version.

src/dune_pkg/lock_dir.ml
```diff
@@ -99,17 +142,67 @@ module Pkg = struct
...
+
+let file_contents_by_path t =
+  (metadata, [ encode_metadata t ])
+  :: (Package_name.Map.to_list t.packages
+     |> List.map ~f:(fun (name, pkg) ->
+            (Package_name.to_string name, Pkg.encode pkg)))
+
+let write_disk ~lock_dir_path t =
+  let lock_dir_path = Path.source lock_dir_path in
+  Path.rm_rf lock_dir_path;
```
rgrinberg, line 199: Before deleting, how about we make sure this is a real lock file directory if it exists to avoid unfortunate accidents. For example, we check that it has a `dune.lock` file.

gridbugs: ooh good idea! Will implement

Issue #1492, opened 2018-10-23
Configurator needs to access opam variables and pass custom flags to pkg-config


rgrinberg:
I've found this shell script for configuration in the wild:

```
export PKG_CONFIG_PATH="$(opam config var lib)/pkgconfig"
flags="$(pkg-config --static ocaml-freestanding --cflags)"
echo "($flags)"
```

Porting this to configurator is a problem because it's not possible to access opam variables and pass `--static` to `pkg-config`. We should support these features.

diml:
Seems good, however I suppose that we don't want to call `opam` inside configurator. Should we assume that `PKG_CONFIG_PATH` is already set by opam? Or maybe try to scan the library search path?

rgrinberg:
> Seems good, however I suppose that we don't want to call opam inside configurator

So what should we do then? I don't think we can know upfront which opam files to use.

> Should we assume that PKG_CONFIG_PATH is already set by opam?

I suppose it's not being set today, but perhaps you're right that opam should improve this. 

> Or maybe try to scan the library search path?

What would scan it for? .pc files?

diml:
Ideally, yh opam should just set it. Otherwise we can use `$OPAM_SWITCH_PREFIX/lib/pkgconfig`.

> What would scan it for? .pc files?

I was thinking of just adding `dir ^ "/pkgconfig"` for every `dir` in the library search path. At least it's not specific to opam.

MisterDA:
I need to pass variables to pkg-config too; for instance this is a typical invocation to generate C files from wayland protocol descriptions in a Makefile:
```makefile
WAYLAND_FLAGS = $(shell pkg-config wayland-client --cflags --libs)
WAYLAND_PROTOCOLS_DIR = $(shell pkg-config wayland-protocols --variable=pkgdatadir)
WAYLAND_SCANNER = $(shell pkg-config wayland-scanner --variable=wayland_scanner)
```
at least dune-configurator could expose the path to pkg-config or a more generic `query` function.

anmonteiro:
I just ran into this too, trying to compile and link `libpq` statically with `postgresql-ocaml`.

In my case, the static build would need to call `pkg-config --static libpq`, which is not currently possible with Configurator.

A hack is adding `~package:"libpq --static"` but I bet that's not portable at all.

anmonteiro:
[...]

PR #7734, opened 2023-05-16
feat(pkg-config): allow reading PKG_CONFIG_ARGN

anmonteiro:
- This change proposes a fix for #1492
- The fix is borrowed from [cmake](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html) by recognizing the `PKG_CONFIG_ARGN` variable for arguments that can be passed to the pkg-config binary
- This solution should allow use cases such as:
    - setting an environment variable e.g. globally for a switch such as `--static` in a `musl+static` switch
    - setting `PKG_CONFIG_ARGN` in the `(env ...)` stanza for ad-hoc use cases


src/dune_rules/pkg_config.ml
```diff
@@ -20,8 +20,14 @@ module Query = struct
     | Libs s -> sprintf "%s.libs" s
     | Cflags s -> sprintf "%s.cflags" s
 
-  let to_args t : _ Command.Args.t list =
+  let to_args t ~env : _ Command.Args.t list =
+    let env_args : _ Command.Args.t list =
+      match Env.get env "PKG_CONFIG_ARGN" with
+      | Some s -> [ As (String.split_on_char ~sep:' ' s) ]
+      | None -> []
```
Alizter, line 27: You can use Arg.empty here and remove the S later. This doesn't need to be a list AFAICT.

src/dune_rules/pkg_config.ml
```diff
@@ -60,8 +66,9 @@ let gen_rule sctx ~loc ~dir query =
   | Error _ -> Memo.return @@ Error `Not_found
   | Ok _ as bin ->
     let command =
+      let env = Super_context.context_env sctx in
```
rgrinberg, line 69: The environment should come from the env tree I believe. So that `(env ..)` is respected.

anmonteiro: Just pushed a change that does both of those.

otherlibs/configurator/test/blackbox-tests/pkg-config-args.t/run.t
```diff
@@ -0,0 +1,34 @@
+These tests show that setting `PKG_CONFIG_ARGN` passes extra args to `pkg-config`
+
+  $ dune build 2>&1 | awk '/run:.*bin\/pkg-config/{a=1}/stderr/{a=0}a'
+  run: $TESTCASE_ROOT/_build/install/default/bin/pkg-config --print-errors gtk+-quartz-3.0
+  -> process exited with code 0
+  -> stdout:
+   | gtk+-quartz-3.0
```
rgrinberg, line 7: Not sure if this is reproducible. Personally I would just write a dummy `pkg-config` shell script that prints its args and put it in PATH for this test.

otherlibs/configurator/test/blackbox-tests/pkg-config-args.t/dune
```diff
@@ -0,0 +1,16 @@
+(executable
+ (name pkg_config)
+ (public_name pkg-config)
```
rgrinberg, line 3: You could also just put in PATH with `env` instead of having a public_name

otherlibs/configurator/test/blackbox-tests/pkg-config-args.t/run.t
anmonteiro: I copied it from `pkg-config-quoting.t`, which is. Note the redefined `pkg_config.exe` in the test that makes it reproducible.

anmonteiro: This is reproducible because of the overridden `pkg_config.ml` included in the test. I made it query / output "dummy-pkg" to make it more explicit that it's not actually calling a real `pkg-config`.

rgrinberg (approved):

PR #7741, opened 2023-05-16
Rewrite install root based paths

richardlford:
To achieve reproducible builds, currently Dune rewrites paths based on the current build directory. However, it misses paths based on the OCaml installation directory (usually ~/.opam/switch).

This change also maps paths in the installation tree to abstract paths headed by "/workspace_root".


src/dune_engine/action_exec.ml
```diff
@@ -542,6 +542,36 @@ type input =
   ; action : Action.t
   }
 
+module Install_root = struct
+  let path_sep = if Sys.win32 then ';' else ':'
+
+  let ( ^/ ) = Filename.concat
```
Alizter, line 548: just `/` should be fine for this.

src/dune_engine/action_exec.ml
```diff
@@ -542,6 +542,36 @@ type input =
...
+      if Sys.file_exists fn then Some fn else None
+
+  let find_ocaml_prog path prog =
+    List.find_map path ~f:(fun dir -> best_prog dir prog)
+
+  let get (env : Env.t) =
+    match Env.get env "OPAM_SWITCH_PREFIX" with
+    | Some dir -> Some dir
+    | None -> (
+      match Env.get env "PATH" with
```
Alizter, line 566: you can replace the rest of this with Option.O let bindings.

src/dune_engine/action_exec.ml
Alizter: Sorry I should have been clearer. I meant / as an operator rather than */ so that it is more uniform with how we do it elsewhere.

src/dune_engine/action_exec.ml
```diff
@@ -542,6 +542,34 @@ type input =
...
+    List.find_map path ~f:(fun dir -> best_prog dir prog)
+
+  let get (env : Env.t) =
+    match Env.get env "OPAM_SWITCH_PREFIX" with
+    | Some dir -> Some dir
+    | None ->
+      let open Option.O in
+      let* path_string = Env.get env "PATH" in
+      let path = String.split path_string ~on:path_sep in
+      let* ocamlc = find_ocaml_prog path "ocamlc" in
```
Alizter, line 569: this can be let+ letting you get rid of the Some

richardlford: done.

richardlford:
I promoted more tests, squashed, and rebased.

src/dune_engine/action_exec.ml
richardlford: Done.

src/dune_engine/action_exec.ml
richardlford: I understand now. You meant use `(/)` instead of `(^/)`. I made that change in the latest push.

Alizter:
Once you have diskuv installed, what you need to do is run the bootstrap command with `with-dkml` infront of it. I did this recently and did something like:
```
with-dkml ocaml boot/bootstrap.ml
```
and then
```
./dune.exe
```
should work as normal for building the dune repo.

Alizter:
You mentioned you were doing this in powershell. I wonder if just doing it in cmd will be different? When I setup windows a few weeks ago, I couldn't get it to work in powershell but it did work in cmd.

richardlford:
> You mentioned you were doing this in powershell. I wonder if just doing it in cmd will be different? When I setup windows a few weeks ago, I couldn't get it to work in powershell but it did work in cmd.

I _was_ using powershell, but the failure I showed above was with cmd.

Do you using the link I mentioned to install diskuv, or do you try to replicate what the CI is doing?

Alizter:
@richardlford I believe I installed diskuv using the installer you linked. Creating the correct environement however required the `with-dkml` wrapper. Perhaps it is the case that you need to use that for running `dune.exe` too?

richardlford:
> That makes me think that your bppm code is separating on `:` in `D:\...` rather than on `;` for windows.

I'm pretty sure you are correct. The way things are supposed to work is that Build_path_prefix_map.map's are constructed abstractly, and then when they are encoded any paths in the map are encoded such that ":" and "=" get encoded in a way that they no longer have ":" or "=". That lets them be used as the delimeters regardless of whether you are on Windows or not. 

What I think is that their are places that are not doing it that way. In particular, I know that there are test scripts that are setting the BUILD_PATH_PREFIX_MAP directly in ways that would break on windows. But this failing test does is not one of those.

To make those other tests more robust we need a way in scripts to do the appropriate encoding. If the CRAM scripts are either running on a *nix system, or on MSYS, then it shouldn't  be too hard to make a sed script that will do the appropriate encoding. 

But, as I said, this test does not do that. In particular, the windows-diff2.t test that I'm using to diagnose the problem, I've even turned off mapping but the problem is still persisting. Currently I'm thinking that the Cram sanitizing might be the problem so I'm looking at that code.

Alizter:
Yes, cram tests will understand and attempt to sanitize using bppm.

Alizter:
Yes, cram tests will understand and attempt to sanitize using bppm.

doc/tests.rst
```diff
@@ -725,11 +725,14 @@ default list of paths is:
 - The temporary directory for the current script will be replaced by ``$TMPDIR``
 
 To add additional paths to this sanitation mechanism, it's sufficient to modify
-the standard BUILD_PATH_PREFIX_MAP_ environment variable. For example:
+the standard BUILD_PATH_PREFIX_MAP_ environment variable. But one must
+be careful to encode the special delimeters (':", "=", "%") that might
+appear in the paths (e.g. on Windwos). For example:
```
Alizter, line 730: typo 9n windows

src/dune_engine/action_exec.ml
```diff
@@ -542,6 +542,34 @@ type input =
   ; action : Action.t
   }
 
+module Install_root = struct
+  let path_sep = if Sys.win32 then ';' else ':'
+
+  let ( / ) = Filename.concat
```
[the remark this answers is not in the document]
Alizter: To be clear, I mean you don't need this at all. I think it would onky be clearer if you had nested concats.

test/blackbox-tests/test-cases/display.t
```diff
@@ -2,7 +2,9 @@ Document how Dune displays various things
 =========================================
 
   $ echo '(lang dune 3.0)' > dune-project
-  $ export BUILD_PATH_PREFIX_MAP=SH=`command -v sh`
+  $
```
Alizter, line 5: empty

src/dune_engine/action_exec.ml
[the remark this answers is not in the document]
richardlford: In my upcoming push I've inlined the Filename.concat.

doc/tests.rst
richardlford: Fix in next push.

test/blackbox-tests/test-cases/display.t
richardlford: Not sure what you mean, but here is the latest code:

```
  $ ENCODED_SH=$(dune_cmd encode-prefix "`command -v sh`")
  $ export BUILD_PATH_PREFIX_MAP="SH=$ENCODED_SH"
```

I've gone through all the places that set BUILD_PATH_PREFIX_MAP manually, and made sure that all of the paths are properly encoded using the new "dune_cmd encode-prefix" command.

When I run `command -v sh` on either Linux or MSYS2's bash, I get `/usr/bin/sh`. But perhaps on some Windows environments this could give a Windows-like path that would need encoding.

richardlford:
I've pushed to reflect your latest review comments. Here is the status of debugging the Windows failures.

1. As mentioned above, I've added a "dune_cmd encode-prefix" subcommand to do the encoding of paths that are to be manually inserted into BUILD_PATH_PREFIX_MAP, and modified all the places the need to use it.
2. In the windows-diff.t/run.t test, I've inserted an echo of the BUILD_PATH_PREFIX_MAP value at the start of the test. It shows that all the Windows paths are properly encoded (so no bare "D:").
3. That test passes on "windows-latest*", but not on "MSVC*", even though for both, the BUILD_PATH_PREFIX_MAP shows as being properly encoded. The "windows-latest*" phase only fails because of the extra echo, but the "MSVC*" phases are getting the invalid map message. **What is the difference between the "windows-latest*" and the "MSVC*" environments? Do you have any idea why this would fail on one but not the other?**
4. One thing that makes this mapping somewhat confusing is that it can be nested. BUILD_PATH_PREFIX_MAP is already set on entry to a CRAM test, partly due to the CRAM code, and partly due to the code in action_exec.ml. Then, inside a CRAM test there are nested calls to dune which may add to the mapping.
5. If BUILD_PATH_PREFIX_MAP is unset at the beginning of `windows-diff.t/run.t`, the test passes. That makes it possible to get the tests to all pass, but if we don't understand why it fails, that is probably not an acceptable alternative.

Alizter:
IINM the plain windows job is really using cygwin (therefore emulating a unix system). MSVC is native windows and so the paths will be windows based.

christinerose (approved):
one very minor grammar suggestion

doc/tests.rst
```diff
@@ -725,11 +725,14 @@ default list of paths is:
 - The temporary directory for the current script will be replaced by ``$TMPDIR``
 
 To add additional paths to this sanitation mechanism, it's sufficient to modify
-the standard BUILD_PATH_PREFIX_MAP_ environment variable. For example:
+the standard BUILD_PATH_PREFIX_MAP environment variable. But one must
+be careful to encode the special delimeters (":", "=", "%") that might
+appear in the paths (e.g. on Windows). For example:
```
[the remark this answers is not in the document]
richardlford: I have made your suggest correction in the latest push.

rgrinberg:
@gridbugs could you review this one?

doc/tests.rst
```diff
@@ -725,11 +725,14 @@ default list of paths is:
 - The temporary directory for the current script will be replaced by ``$TMPDIR``
 
 To add additional paths to this sanitation mechanism, it's sufficient to modify
-the standard BUILD_PATH_PREFIX_MAP_ environment variable. For example:
+the standard BUILD_PATH_PREFIX_MAP environment variable. But one must
```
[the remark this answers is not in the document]
Alizter: shouldnt it be
```
`BUILD_PATH_PREFIX_MAP`
```
rather than
```
BUILD_PATH_PREFIX_MAP_
```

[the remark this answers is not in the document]
gridbugs: No because it's a reference to a link at the end of the file:
```
.. _BUILD_PATH_PREFIX_MAP: https://reproducible-builds.org/specs/build-path-prefix-map/
```

richardlford:
I have rebased, squashed, and force-pushed.

doc/tests.rst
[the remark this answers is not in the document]
richardlford: I'm a little confused by this. Previous code (before my change) was:

```
the standard BUILD_PATH_PREFIX_MAP_ environment variable. For example:
```

with the underscore at the end. The reference link has the underscore at the beginning. Also, in the source code there is this declaration (in dune_util/build_path_prefix_map0.ml):

```
let _BUILD_PATH_PREFIX_MAP = "BUILD_PATH_PREFIX_MAP"
```

But I did check and without my changes the link to the BUILD_PATH_PREFIX_MAP standard does work.
Maybe I don't understand the .rst conventions. Could you explain?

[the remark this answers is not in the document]
richardlford: I've made that change in my upcoming push.

src/dune_engine/action_exec.ml
```diff
@@ -558,11 +584,16 @@ let exec
...
+          ([ Some
+               { Build_path_prefix_map.source = Path.to_absolute_filename root
+               ; target = "/workspace_root"
+               }
+           ]
+          @
+          match Install_root.get env with
+          | None -> []
+          | Some install_root ->
+            [ Some { source = install_root; target = "/workspace_root" } ])
```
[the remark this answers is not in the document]
richardlford: I have incorporated this change in my latest push.

rgrinberg:
Anybody that wants to resume the work on making `$ dune ocaml debug` is welcome to revive this PR

PR #7742, opened 2023-05-16
feature(pkg): set environment in build rules

rgrinberg:
I've added the `(withenv ..)` action to replicate opam's env operations. The implementation isn't 100% done yet (we don't handle removals of env vars yet), but it's good enough for most packages.

@Leonidas-from-XIV  @gridbugs i've added you as reviewers so that you're familiar with the code and are aware of this feature.

### DIFF
--- a/src/dune_lang/action.ml
+++ b/src/dune_lang/action.ml
@@ -59,6 +59,45 @@ module File_perm = struct
     | Executable -> 0o777
 end
 
+module Env_update = struct
+  type op =
+    | Eq
+    | PlusEq
+    | EqPlus
+    | ColonEq
+    | EqColon
+    | EqPlusEq
+
+  type 'a t =
+    { op : op
+    ; var : Env.Var.t
+    ; value : 'a
+    }
+
+  let ops =
+    [ ("=", Eq)
+    ; ("+=", PlusEq)
+    ; ("=+", EqPlus)
+    ; (":=", ColonEq)
+    ; ("=:", EqColon)
+    ; ("=+=", EqPlusEq)
+    ]
+
+  let decode =
+    let open Decoder in
+    let env_update_op = enum ops in
+    let+ op, var, value = triple env_update_op string String_with_vars.decode in
+    { op; var; value }
+
+  let encode { op; var; value } =
+    let op =
+      List.find_map ops ~f:(fun (k, v) ->
+          if Poly.equal v op then Some k else None)
+      |> Option.value_exn
+    in
+    List [ atom op; atom var; String_with_vars.encode value ]
+end
+
 type t =
   | Run of String_with_vars.t * String_with_vars.t list
   | With_accepted_exit_codes of int Predicate_lang.t * t
@@ -88,6 +127,7 @@ type t =
   | Cram of String_with_vars.t
   | Patch of String_with_vars.t
   | Substitute of String_with_vars.t * String_with_vars.t
+  | Withenv of String_with_vars.t Env_update.t list * t
 
 let is_dev_null t = String_with_vars.is_pform t (Var Dev_null)
 
@@ -264,7 +304,7 @@ let cstrs_dune_file t =
 let decode_dune_file = Decoder.fix @@ fun t -> Decoder.sum (cstrs_dune_file t)
 
 let decode_pkg =
-  let cstrs_pkg =
+  let cstrs_pkg t =
     let open Decoder in
     [ ( "patch"
       , Syntax.since Pkg.syntax (0, 1)
@@ -275,9 +315,14 @@ let decode_pkg =
         >>> let+ input = sw
             and+ output = sw in
             Substitute (input, output) )
+    ; ( "withenv"
+      , Syntax.since Pkg.syntax (0, 1)
+        >>> let+ ops = enter (repeat Env_update.decode)
+            and+ t = t in
+            Withenv (ops, t) )
     ]
   in
-  Decoder.fix @@ fun t -> Decoder.sum (cstrs_dune_file t @ cstrs_pkg)
+  Decoder.fix @@ fun t -> Decoder.sum (cstrs_dune_file t @ cstrs_pkg t)
 
 let rec encode =
   let sw = String_with_vars.encode in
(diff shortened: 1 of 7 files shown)


src/dune_lang/action.ml
```diff
@@ -349,6 +394,12 @@ let rec encode =
   | Cram script -> List [ atom "cram"; sw script ]
   | Patch i -> List [ atom "patch"; sw i ]
   | Substitute (i, o) -> List [ atom "substitute"; sw i; sw o ]
+  | Withenv (ops, t) ->
+    List
+      [ atom "withenv"
```
Leonidas-from-XIV, line 399: `with-env` would probably fit better with e.g. `write-file`.

test/blackbox-tests/test-cases/pkg/withenv.t
```diff
@@ -0,0 +1,17 @@
...
+  $ mkdir dune.lock
+  $ cat >dune.lock/lock.dune <<EOF
+  > (lang package 0.1)
+  > EOF
+  $ cat >dune.lock/test <<'EOF'
+  > (build
+  >  (withenv
+  >   ((= FOO BAR)
+  >    (= BAR xxx)
+  >    (+= BAR yyy))
```
Leonidas-from-XIV, line 12: I'm wondering whether for usability a short form of the type `(withenv (ENV VALUE) ...)` might be useful.

src/dune_rules/pkg_rules.ml
```diff
@@ -195,6 +186,48 @@ module Install_cookie = struct
     | Some f -> f
 end
 
+module Env_update = struct
+  include Dune_lang.Action.Env_update
+
+  let update kind ~new_v ~old_v ~f =
+    if new_v = "" then old_v
```
Leonidas-from-XIV, line 193: Why is setting the new value to an empty string a no-op?

src/dune_rules/pkg_rules.ml
```diff
@@ -195,6 +186,48 @@ module Install_cookie = struct
...
+      match kind with
+      | `Colon ->
+        let old_v = Option.value ~default:[] old_v in
+        Some (f ~old_v ~new_v)
+      | `Plus -> (
+        match old_v with
+        | None | Some [] -> Some [ Value.String new_v ]
+        | Some old_v -> Some (f ~old_v ~new_v))
+
+  let append = update ~f:(fun ~old_v ~new_v -> old_v @ [ Value.String new_v ])
```
Leonidas-from-XIV, line 204: Maybe it would make sense not to store the env as a list but as a sort of deque and only materialize the list in the end?

OTOH, it probably doesn't matter unless people append large amounts of variables.

src/dune_rules/pkg_rules.ml
```diff
@@ -195,6 +186,48 @@ module Install_cookie = struct
...
+        match op with
+        | Eq ->
+          if new_v = "" then if Sys.win32 then None else Some [ String "" ]
+          else Some [ Value.String new_v ]
+        | PlusEq -> prepend `Plus
+        | ColonEq -> prepend `Colon
+        | EqPlus -> append `Plus
+        | EqColon -> append `Colon
+        | EqPlusEq ->
+          (* TODO nobody uses this AFAIK *)
```
Leonidas-from-XIV, line 221: Maybe remove this variant from `op` and display an error when parsing? After all it can always be added later.

src/dune_rules/pkg_rules.ml
rgrinberg: Yeah, it probably won't matter. If this is going to be optimized, I'd rather let the opam maintainers do it and give us some generic component we can replace this code with.

src/dune_rules/pkg_rules.ml
rgrinberg: I do plan to implement this before the feature is ready. So it's fine to leave this for now. You or @gridbugs are welcome to beat me to it though.

src/dune_rules/pkg_rules.ml
rgrinberg: Looks like I misunderstood the spec. Will fix.

test/blackbox-tests/test-cases/pkg/withenv.t
rgrinberg: Note that this form isn't going to be written by users so I don't mind making it verbose. If we introduce something ergonomic, I would rather move away from the cryptic symbols altogether and stick to names like `append` or `prepend`.

src/dune_lang/action.ml
Alizter: I would prefer the `-` version. We can create an issue to rename `setenv` to `set-env`.

Leonidas-from-XIV: I was looking through the docs and we mostly use `_` (which is not really typical in Lisp, but its better to be consistent than typical I'd say).

src/dune_rules/pkg_rules.ml
rgrinberg: Actually, this is the correct behavior. From the manual:

```
FOO += "", FOO := "", etc. are all ignored - i.e. opam never adds empty segments to an existing variable.
```

rgrinberg:
ping @gridbugs

test/blackbox-tests/test-cases/pkg/withenv.t
```diff
@@ -0,0 +1,22 @@
+Setting environment variables in actions
+
+  $ mkdir dune.lock
+  $ cat >dune.lock/lock.dune <<EOF
+  > (lang package 0.1)
+  > EOF
+  $ cat >dune.lock/test <<'EOF'
```
gridbugs, line 7: What's the point of the quotes around "EOF"?

rgrinberg: It is so that `$variables` aren't evaluated when writing the file. The lock file contains a shell snippet that prints an environment variable. We want it to be done when the shell script is executed, rather than when it is written.

gridbugs (approved):

Issue #7748, opened 2023-05-17
regression: coq-serapi.8.13.0+0.13.1


emillon:
Hi,
`coq-serapi.8.13.0+0.13.1` does not build with `dune.3.8.0~alpha3`, but it does with `dune.3.7.1`.
See [this build](https://toxis.caelum.ci.dev/github/ocaml/opam-repository/commit/b442a4abc1e8535be3f0490fd4ac4311c31650f3/variant/compilers,4.14,dune-configurator.3.8.0~alpha3,revdeps,coq-serapi.8.13.0+0.13.1):

```
#=== ERROR while compiling coq-serapi.8.13.0+0.13.1 ===========================#
# context              2.2.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.4.14.1 | file:///home/opam/opam-repository
# path                 ~/.opam/4.14/.opam-switch/build/coq-serapi.8.13.0+0.13.1
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p coq-serapi -j 255
# exit-code            1
# env-file             ~/.opam/log/coq-serapi-7-872efe.env
# output-file          ~/.opam/log/coq-serapi-7-872efe.out
### output ###
# Error: exception Dune_rules__coq_config.Vars.E(_)
# Raised at Dune_rules__coq_config.Vars.fail in file
#   "src/dune_rules/coq/coq_config.ml", line 35, characters 21-49
# Called from Dune_rules__coq_config.Vars.get_path in file
#   "src/dune_rules/coq/coq_config.ml" (inlined), line 44, characters 23-32
# Called from Dune_rules__coq_config.make_res in file
#   "src/dune_rules/coq/coq_config.ml", line 186, characters 25-56
# Called from Fiber__core.O.(>>|).(fun) in file "otherlibs/fiber/src/core.ml",
#   line 250, characters 36-41
# Called from Fiber__scheduler.exec in file "otherlibs/fiber/src/scheduler.ml",
#   line 73, characters 8-11
# -> required by ("<unnamed>", ())
# -> required by ("<unnamed>", ())
# -> required by ("<unnamed>", ())
# -> required by ("Super_context.all", ())
# -> required by ("toplevel", ())
# 
# I must not crash.  Uncertainty is the mind-killer. Exceptions are the
# little-death that brings total obliteration.  I will fully express my cases. 
# Execution will pass over me and through me.  And when it has gone past, I
# will unwind the stack along its path.  Where the cases are handled there will
[...]

PR #7754, opened 2023-05-17
fix(coq): make loading of theories in scope lazier 

### LINKED ISSUE #7747: regression: coq-stdlib.8.17.0
Hi,

`coq-stdlib.8.17.0` does not build correctly with `dune.3.8.0~alpha3`, while it did with `3.7.1`.
See [this build](https://toxis.caelum.ci.dev/github/ocaml/opam-repository/commit/b442a4abc1e8535be3f0490fd4ac4311c31650f3/variant/compilers,4.14,dune-configurator.3.8.0~alpha3,revdeps,coq-stdlib.8.17.0).

```
#=== ERROR while compiling coq-stdlib.8.17.0 ==================================#
# context              2.2.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.4.14.1 | file:///home/opam/opam-repository
# path                 ~/.opam/4.14/.opam-switch/build/coq-stdlib.8.17.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build ./configure -prefix /home/opam/.opam/4.14 -mandir /home/opam/.opam/4.14/man -libdir /home/opam/.opam/4.14/lib/coq -native-compiler no
# exit-code            1
# env-file             ~/.opam/log/coq-stdlib-7-f87a35.env
# output-file          ~/.opam/log/coq-stdlib-7-f87a35.out
### output ###
# cannot guess a path for Coq libraries; please use -coqlib option or ensure you have installed the package containing Coq's stdlib (coq-stdlib in OPAM) If you intend to use Coq without a standard library, the -boot -noinit options must be used.
# Error: Error while running /home/opam/.opam/4.14/bin/coqc --config
# Exit code: 1
# Output:
# 
```

I bisected that to 98de95004d753a3e30b0f4a3bbb09312541e892b (#7047).

@Alizter can you have a look?

Alizter:
We make the loading of theories even lazier in scope.ml. This required refactoring the coq db into a Memo.Lazy.t. I'm open to a better way of doing this but this at least fixes the bug observed in #7747 and #7748.

I added a test case for #7747 and showed that this patch fixes it also.

This PR fixes the following issues, however they will be closed on a new run of the opam ci:
- #7747 
- #7748

### DIFF
--- a/src/dune_rules/scope.ml
+++ b/src/dune_rules/scope.ml
@@ -4,7 +4,7 @@ open Memo.O
 type t =
   { project : Dune_project.t
   ; db : Lib.DB.t
-  ; coq_db : Coq_lib.DB.t Lazy.t
+  ; coq_db : Coq_lib.DB.t Memo.Lazy.t
   ; root : Path.Build.t
   }
 
@@ -14,7 +14,7 @@ let project t = t.project
 
 let libs t = t.db
 
-let coq_libs t = Lazy.force t.coq_db
+let coq_libs t = Memo.Lazy.force t.coq_db
 
 module DB = struct
   type scope = t
@@ -162,9 +162,6 @@ module DB = struct
     |> Coq_lib.DB.create_from_coqlib_stanzas ~find_db
          ~parent:(Some installed_theories)
 
-  let public_theories ~find_db ~installed_theories coq_stanzas =
-    lazy (public_theories ~find_db ~installed_theories coq_stanzas)
-
   (* Create a database from the public libraries defined in the stanzas *)
   let public_libs t ~installed_libs ~lib_config ~host stanzas =
     let public_libs =
@@ -228,7 +225,6 @@ module DB = struct
           (Dune_project.root project, (dir, stanza)))
       |> Path.Source.Map.of_list_multi
     in
-    let public_theories = Lazy.force public_theories in
     let parent = Some public_theories in
     let find_db dir = snd (find_by_dir_in_map db_by_project_dir dir) in
     Path.Source.Map.merge projects_by_dir coq_stanzas_by_project_dir
@@ -289,14 +285,13 @@ module DB = struct
              (project, db))
     in
     let coq_scopes_by_dir =
-      lazy
-        (coq_scopes_by_dir db_by_project_dir projects_by_dir public_theories
-           coq_stanzas)
+      Memo.Lazy.map public_theories ~f:(fun public_theories ->
+          coq_scopes_by_dir db_by_project_dir projects_by_dir public_theories
+            coq_stanzas)
     in
     let coq_db_find dir =
-      lazy
-        (let map = Lazy.force coq_scopes_by_dir in
-         Path.Source.Map.find_exn map dir)
+      Memo.Lazy.map coq_scopes_by_dir ~f:(fun x ->
+          Path.Source.Map.find_exn x dir)
     in
     Path.Source.Map.mapi db_by_project_dir ~f:(fun dir (project, db) ->
         let root =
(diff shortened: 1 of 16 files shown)


src/dune_rules/scope.ml
```diff
@@ -289,14 +290,15 @@ module DB = struct
...
+      Memo.lazy_ @@ fun () ->
+      let+ public_theories = Memo.Lazy.force public_theories in
+      coq_scopes_by_dir db_by_project_dir projects_by_dir public_theories
+        coq_stanzas
     in
     let coq_db_find dir =
-      lazy
-        (let map = Lazy.force coq_scopes_by_dir in
-         Path.Source.Map.find_exn map dir)
+      Memo.lazy_ @@ fun () ->
```
anmonteiro, line 299: I think you can use `Memo.Lazy.map` here.

src/dune_rules/scope.ml
```diff
@@ -289,14 +290,15 @@ module DB = struct
              (project, db))
     in
     let coq_scopes_by_dir =
-      lazy
-        (coq_scopes_by_dir db_by_project_dir projects_by_dir public_theories
-           coq_stanzas)
+      Memo.lazy_ @@ fun () ->
```
anmonteiro, line 293: I think you can use Memo.Lazy.map here.

src/dune_rules/scope.ml
```diff
@@ -163,7 +163,9 @@ module DB = struct
          ~parent:(Some installed_theories)
 
   let public_theories ~find_db ~installed_theories coq_stanzas =
-    lazy (public_theories ~find_db ~installed_theories coq_stanzas)
+    Memo.lazy_ @@ fun () ->
```
anmonteiro, line 166: I think you can use Memo.Lazy.map here.

ejgallego (approved):
Thanks a lot for the quick fix @Alizter , code looks good to me, and indeed the new type seems like the right one.

Sorry for the oversight in the initial implementation.

ejgallego:
> fixes https://github.com/ocaml/dune/issues/7747

should also fix the other one, tho your previous PR did also fix that one, so maybe we can mark both PRs as fixing that bug.

Alizter:
Actually, I will let @emillon close those since he will do another opam build.

ejgallego:
> Actually, I will let @emillon close those since he will do another opam build.

Changelog / PRs description should reflect bugs numbers tho, right?

Alizter:
I'll mention them in the PR description but we don't need a changelog since we are fixing a bug about a feature we just introduced in this version.

Alizter:
OK so in the first case, the first error is the one we want, however because it is very lazy it is being delayed until rule loading causing it to emit both errors. Ideally we would want only the first error like before. Not sure what the best way to fix this is. Perhaps during rule loading, we can force the database to be constructed.

The second one also looks like a straight up bug.

ejgallego:
Indeed our current approach to side-effects seems kind of fragile, I do dislike the style where we recover the scope imperatively inside coq_rules helper functions. We should make sure the library resolution is actually memoized, and we are not scanning twice. We resolve now the Coq DB a lot of times per-module, so beware of performance impact.

I would be much more comfortable if we would try to resolve a bit more in the top-level (but again this needs to be done with care)

As for the second error, that's indeed strange, is the test running with the cache enabled?

Indeed we may have a soundness bug here, as the DB build needs to depend on the contents of the directories that are in COQPATH and user contrib. (Rules should already depend on that)

Alizter:
@ejgallego I'm not sure there will be a perf impact since the database we are resolving is memod.

The other test does not have caching enabled.

ejgallego:
> I'm not sure there will be a perf impact since the database we are resolving is memod.

I meant we should make sure the DB is memoed properly.

> The other test does not have caching enabled. (as in behaves the same when explicitly disabling)

I see thanks.

> @ejgallego I'd appreciate if you could take a look since I am not sure what more I can do here any longer.

I did have a look and I have no idea what is going on.

I need to leave now, I'd try to put a few printfs in the DB creation code to try to see if the DB code is called and with what paths.

ejgallego:
Thanks @emillon , @Alizter if we don't get the cycles to look at why the DB is being too lazy in this case, IMHO it is fine to open an issue and get 3.8.0 like this, WDTY?

Alizter:
I don't mind the "multiple rules generated" error message since that is just extra noise essentially, however the other error is a bit more serious. In that case, the duplicate checking in the install directory isn't being carried out, so I would prefer to fix that.

Alizter:
OK so the missing duplicate checking was actually an incorrectly set up test. Good thing we caught it. I've fixed that and a few other tests now. I will have a look at the multiple rules issue now.

ejgallego:
Great work, thanks @Alizter . Indeed the duplicate target error is a bit annoying but really a corner case, if someone complains we can fine-tune it easily.

PR #7757, opened 2023-05-18
Make the `interface` and `implementation` fields of a dialect optional

gpetiot:
Suggested by @emillon in https://github.com/ocaml/ocaml.org/pull/1181#discussion_r1196806787

Our usecase is defining a dialect for `.eml` files, that don't have interface files.

If I had changed the `Ml_kind.Dict.intf` into an option type there would be too many changes to propagate (I tried) and it makes the code harder to follow (harder to keep the impl/intf abstraction when reading the fields).


nojb (commented):
I'm not fond of the dummy extension. The type `Ml_kind.Dict` is used in many other places apart from being used for dialects. What about defining a type specifically for dialects that has an optional `intf`?

src/ocaml/ml_kind.ml
```diff
@@ -54,3 +54,30 @@ module Dict = struct
     let open Dyn in
     record [ ("impl", f impl); ("intf", f intf) ]
 end
+
+module Dialect = struct
+  type 'a t =
+    { impl : 'a
+    ; intf : 'a option
```
nojb, line 61: Is there a reason to make `impl` mandatory? Why not keep the symmetry and make both optional?

nojb (commented):
I left a few comments. Also, as a general rule, I am not fond of infix operators such as `@@`, but that's my personal preference. I don't know if the other Dune devs have a preference. cc @rgrinberg @emillon @Alizter

emillon:
Regarding `(@@)`, indeed that's not something we use a lot in the dune code base. But in many cases it looks like you can replace it with other functions, I'll leave some comments.

emillon:
Just thinking out loud: we plan to add some formatting-only dialects (for example if you want to set a formatter for JSON files in a project). We just need to make sure that there's no syntax clash. But I don't think that these are going to use `(implementation)` or `(interface)`, so this should not interfere.

src/dune_rules/dialect.ml
```diff
@@ -54,15 +54,16 @@ let to_dyn { name; file_kinds } =
...
+    ; ("file_kinds", Ml_kind.Dialect.to_dyn File_kind.to_dyn file_kinds)
     ]
 
 let encode { name; file_kinds } =
   let open Dune_lang.Encoder in
   let file_kind_stanzas =
     let open Ml_kind in
     List.map ~f:File_kind.encode
-      [ Dict.get file_kinds Intf; Dict.get file_kinds Impl ]
+    @@ List.filter_opt
```
emillon, line 65: You should be able to refactor that as a `List.filter_map all ~f:(fun kind -> ...)`

src/dune_rules/dialect.ml
```diff
@@ -126,9 +133,9 @@ let ocaml =
           , [ ".ocamlformat"; ".ocamlformat-ignore"; ".ocamlformat-enable" ] )
     }
   in
-  let intf = file_kind Ml_kind.Intf ".mli" in
+  let intf = Option.some @@ file_kind Ml_kind.Intf ".mli" in
```
emillon, line 136:
```suggestion
  let intf = Some (file_kind Ml_kind.Intf ".mli") in
```

src/dune_rules/dialect.ml
```diff
@@ -152,9 +159,9 @@ let reason =
     ; format = Some (Loc.none, format, [])
     }
   in
-  let intf = file_kind Ml_kind.Intf ".rei" in
+  let intf = Option.some @@ file_kind Ml_kind.Intf ".rei" in
```
emillon, line 162:
```suggestion
  let intf = Some (file_kind Ml_kind.Intf ".rei") in
```

src/dune_rules/dialect.ml
```diff
@@ -179,22 +186,23 @@ let rescript =
     ; format = Some (Loc.none, format, [])
     }
   in
-  let intf = file_kind Ml_kind.Intf ".resi" in
+  let intf = Option.some @@ file_kind Ml_kind.Intf ".resi" in
```
emillon, line 189:
```suggestion
  let intf = Some (file_kind Ml_kind.Intf ".resi") in
```

nojb:
> Just thinking out loud:

Indeed what you describe is a different concept that `dialects` which are meant for "formatters" that outputs OCaml code meant to be compiled.

src/dune_rules/dialect.ml
```diff
@@ -208,18 +216,18 @@ module DB = struct
   let set_extensions_for_merlin t =
     let v =
       fold t ~init:[] ~f:(fun d s ->
-          let impl = extension d Ml_kind.Impl in
+          let impl = Option.value_exn @@ extension d Ml_kind.Impl in
```
emillon, line 219: why can't this raise?

src/dune_rules/dialect.ml
```diff
@@ -208,18 +216,18 @@ module DB = struct
...
           let intf = extension d Ml_kind.Intf in
           if
             (* Only include dialects with no preprocessing and skip default file
                extensions *)
             preprocess d Ml_kind.Impl <> None
             || preprocess d Ml_kind.Intf <> None
-            || impl = extension ocaml Ml_kind.Impl
+            || impl = Option.value_exn @@ extension ocaml Ml_kind.Impl
```
emillon, line 226: same question here, why can't this raise?

src/dune_rules/dialect.ml
```diff
@@ -247,9 +255,12 @@ module DB = struct
           ]
     in
     let by_extension =
-      add_ext
-        (add_ext by_extension dialect.file_kinds.intf.extension)
-        dialect.file_kinds.impl.extension
+      let by_extension =
+        Option.value ~default:by_extension
```
emillon, line 259: I think it will be clearer to just `match dialect.file_kinds.intf with` here

src/dune_rules/merlin/merlin.ml
```diff
@@ -171,8 +171,10 @@ module Processed = struct
         :: flags
     in
     let suffixes =
-      List.map extensions ~f:(fun { Ml_kind.Dict.impl; intf } ->
-          make_directive "SUFFIX" (Sexp.Atom (Printf.sprintf "%s %s" impl intf)))
+      List.map extensions ~f:(fun { Ml_kind.Dialect.impl; intf } ->
+          make_directive "SUFFIX"
+            (Sexp.Atom
+               (Printf.sprintf "%s %s" impl (Option.value intf ~default:""))))
```
emillon, line 177: I'm not sure what is consuming that sexp (merlin I presume) but it needs to be able to parse this correctly.

Sidenote: putting a space in atom like this does not look correct (even if there's no problem as a csexp). If we fix one issue we should fix the other one by making a dict of extensions.

src/dune_rules/merlin/merlin.ml
```diff
@@ -198,8 +200,9 @@ module Processed = struct
         printf "STDLIB %s\n" (serialize_path stdlib_dir));
     Path.Set.iter obj_dirs ~f:(fun p -> printf "B %s\n" (serialize_path p));
     Path.Set.iter src_dirs ~f:(fun p -> printf "S %s\n" (serialize_path p));
-    List.iter extensions ~f:(fun { Ml_kind.Dict.impl; intf } ->
-        printf "SUFFIX %s" (Printf.sprintf "%s %s" impl intf));
+    List.iter extensions ~f:(fun { Ml_kind.Dialect.impl; intf } ->
+        printf "SUFFIX %s"
+          (Printf.sprintf "%s %s" impl (Option.value intf ~default:"")));
```
emillon, line 205: same remark as above

src/dune_rules/module.ml
```diff
@@ -377,7 +377,10 @@ let decode_old ~src_dir =
        if exists then
          let module_basename n ~(ml_kind : Ml_kind.t) ~(dialect : Dialect.t) =
            let n = Module_name.to_string n in
-           String.lowercase n ^ Dialect.extension dialect ml_kind
+           let ext =
+             Option.value (Dialect.extension dialect ml_kind) ~default:""
```
emillon, line 381: I don't think that this case is correct - it should be an error to refer to the filename of an interface for a dialect that has no interfaces.

emillon (changes requested):

src/dune_rules/merlin/merlin.ml
gpetiot: If this is consumed by [this](https://github.com/ocaml/merlin/blob/29ad7e1cbcfcbd9eb94ab22fb7a632cf384004d3/src/kernel/mconfig_dot.ml#L70) maybe we should just write `~default:impl`

src/ocaml/ml_kind.ml
gpetiot: done

src/dune_rules/dialect.ml
gpetiot: done

src/dune_rules/dialect.ml
gpetiot: done

src/dune_rules/dialect.ml
gpetiot: done

src/dune_rules/dialect.ml
gpetiot: done

src/dune_rules/dialect.ml
gpetiot: done

src/dune_rules/dialect.ml
gpetiot: done

src/dune_rules/dialect.ml
gpetiot: done

src/dune_rules/merlin/merlin.ml
gpetiot: done

src/dune_rules/merlin/merlin.ml
gpetiot: done

src/dune_rules/module.ml
gpetiot: done

src/dune_rules/dialect.ml
```diff
@@ -45,7 +45,7 @@ end
 
 type t =
   { name : string
-  ; file_kinds : File_kind.t Ml_kind.Dict.t
+  ; file_kinds : File_kind.t Ml_kind.Dialect.t
```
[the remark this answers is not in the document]
gpetiot: done

src/dune_rules/dialect.ml
```diff
@@ -88,17 +89,23 @@ let decode =
   fields
     (let+ name = field "name" string
      and+ impl = field "implementation" (fields (kind Ml_kind.Impl))
-     and+ intf = field "interface" (fields (kind Ml_kind.Intf)) in
-     { name; file_kinds = Ml_kind.Dict.make ~intf ~impl })
+     and+ intf = field_o "interface" (fields (kind Ml_kind.Intf)) in
```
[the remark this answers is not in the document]
gpetiot: I've added a TODO placeholder for now, do you know what this should look like?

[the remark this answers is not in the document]
emillon: You can start by writing a test for this: with `(lang dune 3.8)` this should fail with an error, and accepted with `(lang dune 3.9)`.

In terms of implementation:
- add `and+ version = Dune_lang.Syntax.get_exn Stanza.syntax` to get the declared version
- add something along the lines of this after the parsing code:
```ocaml
let interface_optional_since = 3,9 in
if version < interface_optional_since && Option.is_none interface then
  Syntax.Error.since loc Stanza.syntax interface_optional_since
    ~what:"omitting (interface) in dialects"
```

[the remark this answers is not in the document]
gpetiot: Thanks! done.

nojb (commented):
Just to confirm: when, say, `(implementation)` is missing, then it is still possible to provide one by using `.ml`? I would add a test showing this.

src/dune_rules/dialect.ml
```diff
@@ -87,18 +89,35 @@ let decode =
   in
   fields
     (let+ name = field "name" string
-     and+ impl = field "implementation" (fields (kind Ml_kind.Impl))
-     and+ intf = field "interface" (fields (kind Ml_kind.Intf)) in
+     and+ impl_loc, impl =
+       located (field_o "implementation" (fields (kind Ml_kind.Impl)))
```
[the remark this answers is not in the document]
gpetiot: Makes sense, done.

gpetiot:
> Just to confirm: when, say, `(implementation)` is missing, then it is still possible to provide one by using `.ml`? I would add a test showing this.

I added a test and it works as expected.

nojb (approved):
LGTM. A Changes entry and the manual need to be updated.

doc/dune-files.rst
```diff
@@ -69,6 +69,8 @@ dialect
    .. dune:field:: implementation
 
       Details related to the implementation files (corresponding to `*.ml`).
+      
+      In Dune 3.9 and later, this field is optional.
```
Alizter, line 73: Probably should mention dune lang rather than dune since newer versions of dune will have the old behaviour if the lang version is not up to date.

gpetiot: done

nojb:
Looks OK to go to me. @emillon does it look OK to you?

emillon:
> Looks OK to go to me. @emillon does it look OK to you?

Later I think we should just document the latest version of the language (and omit changelog bits from the manual), but we're not completely there yet. In the meantime I think it's better to use directives than prose for this so I pushed some suggestions.

test/blackbox-tests/test-cases/dialects/no_impl.t/run.t
```diff
@@ -0,0 +1,6 @@
+Test the (dialect ...) stanza inside the dune-project file.
```
christinerose, line 1:
```suggestion
Test the (dialect ...) stanza inside the `dune-project` file.
```

test/blackbox-tests/test-cases/dialects/no_intf_bad.t/run.t
```diff
@@ -0,0 +1,25 @@
+Test the (dialect ...) stanza inside the dune-project file.
```
christinerose, line 1:
```suggestion
Test the (dialect ...) stanza inside the `dune-project` file.
```

christinerose (approved):

PR #7760, opened 2023-05-18
refactor: move colors into own library

### LINKED ISSUE #7759: Remove Color Handling from Rules
Currently the rules contain some color handling code that should be to its own library.

Alizter:
We move color handling code in dune_rules to a separate library called `dune_cli` which is then used in `bin/` rather than in `dune_rules` allowing `dune_cli` and `dune_rules` to be independent.

There is only a single piece of colour handling code left in `dune_rules` which tries to support colours when the compiler version is 4.02.

fix #7759

### DIFF
--- a/bin/common.ml
+++ b/bin/common.ml
@@ -2,7 +2,6 @@ open Stdune
 open Dune_config_file
 module Execution_env = Dune_util.Execution_env
 module Console = Dune_console
-module Colors = Dune_rules.Colors
 module Clflags = Dune_engine.Clflags
 module Graph = Dune_graph.Graph
 module Package = Dune_rules.Package
@@ -1044,7 +1043,17 @@ let init ?action_runner ?log_file c =
     let+ w = Dune_rules.Workspace.workspace () in
     Dune_engine.Execution_parameters.builtin_default
     |> Dune_rules.Workspace.update_execution_parameters w);
-  Dune_rules.Global.init ~capture_outputs:c.builder.capture_outputs;
+  (* Setup color support *)
+  Dune_rules.Clflags.supports_color :=
+    c.builder.capture_outputs && Lazy.force Ansi_color.stderr_supports_color;
+  (* Setup initial environment *)
+  let env =
+    if !Dune_rules.Clflags.supports_color then
+      Dune_cli.Colors.setup_env_for_colors Env.initial
+    else Env.initial
+  in
+  (* Initialize the initial enviornment of the rules *)
+  Dune_rules.Global.init env;
   let cache_config =
     match config.cache_enabled with
     | Disabled -> Dune_cache.Config.Disabled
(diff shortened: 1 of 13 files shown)


src/dune_rules/global.ml
```diff
@@ -9,7 +9,7 @@ let init ~capture_outputs =
          (not capture_outputs)
          || not (Lazy.force Ansi_color.stderr_supports_color)
```
rgrinberg, line 10: All of this logic should move to the colors module as well. `Global.init` should just take an environment.

rgrinberg:
Thanks. This is moving in the right direction. To improve this further, I would like:

1. dune_colors to be named dune_cli. We'll be able to add other cli concerns there.
2. dune_rules shouldn't depend on dune_colors (dune_cli) but rather the command line should initialize the rules with the right parameters inside bin/

Alizter:
I've renamed the library to dune_cli and done some more refactoring of color handling code.

I also tried moving Global.init to Dune_cli unless I have misunderstood you here. To me, it look like Global.init is simply a thin wrapper adding colors to an initial env.

In order to initialise the "global" environment, we need to refactor dune_rules/context.ml to take an Env.t as well as a context name in the database creation. I went ahead and attempted this but this change is propagating quite far and I am not sure this is the way to go. Therefore I have left it partially done.

src/dune_rules/context.ml
```diff
@@ -569,7 +555,7 @@ module rec Instantiate : sig
   val instantiate : Context_name.t -> instance Memo.t
 end = struct
   let instantiate_impl name : instance Memo.t =
-    let env = Global.env () in
+    let env = Dune_cli.Global.env () in
```
Alizter, line 558: How to lift this?

I generalized this whole module to use a pair of an env and a context name but the changes propagate quite far and I am unsure this is the api we want in the end.

bin/common.ml
```diff
@@ -1044,7 +1043,17 @@ let init ?action_runner ?log_file c =
...
     |> Dune_rules.Workspace.update_execution_parameters w);
-  Dune_rules.Global.init ~capture_outputs:c.builder.capture_outputs;
+  (* Setup color support *)
+  Dune_rules.Clflags.supports_color :=
+    c.builder.capture_outputs && Lazy.force Ansi_color.stderr_supports_color;
+  (* Setup initial environment *)
+  let env =
+    if !Dune_rules.Clflags.supports_color then
+      Dune_cli.Colors.setup_env_for_colors Env.initial
+    else Env.initial
```
rgrinberg, line 1053: you can move most of this logic to dune_cli, and just expose the following function:

```
val color_support : capture_outputs:bool -> (Env.t * [`Color of bool])
```

Alizter: I'm not sure what you mean here. The reason I have to initialise supports_color is because of its use in context.ml. even if I refactor like you say I don't see the point. Perhaps I am missing something?

PR #7762, opened 2023-05-19
refactor(engine): do not depend on Dune_lang.Glob

rgrinberg:
We can use Dune_glob directly if we move some functions from dune_lang
to dune_glob


snowleopard (approved):
Looks good!

PR #7763, opened 2023-05-19
fix(tui): fix supsend/resume

rgrinberg:

### DIFF
--- a/src/dune_engine/scheduler.ml
+++ b/src/dune_engine/scheduler.ml
@@ -63,7 +63,8 @@ let interrupt_signals : Signal.t list = [ Int; Quit; Term ]
 
 (* In addition, the scheduler also blocks some other signals so that only
    designated threads can handle them by unblocking *)
-let blocked_signals : Signal.t list = Winch :: interrupt_signals
+let blocked_signals : Signal.t list =
+  Dune_util.Terminal_signals.signals @ interrupt_signals
 
 module Thread : sig
   val spawn : signal_watcher:[ `Yes | `No ] -> (unit -> unit) -> unit
--- a/src/dune_threaded_console/dune_threaded_console.ml
+++ b/src/dune_threaded_console/dune_threaded_console.ml
@@ -61,8 +61,7 @@ let make (module Base : S) : (module Dune_console.Backend) =
     let start () =
       Base.start ();
       Dune_engine.Scheduler.spawn_thread @@ fun () ->
-      if not Sys.win32 then
-        ignore (Unix.sigprocmask SIG_UNBLOCK [ Signal.to_int Winch ] : int list);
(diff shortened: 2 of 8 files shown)


Alizter:
The suspend and resume appears to work, however when I suspend the cursor is missing from the terminal. It comes back if I do a `reset` however.

rgrinberg:
I suppose we shouldn't be showing the cursor at all. There's nothing the user can enter anyway

Alizter:
I mean the cursor isn't showing in the terminal session itself. i.e. start dune -> Ctrl - Z -> no cursor.

rgrinberg:
@snowleopard I added a few terminal related signals to the signal mask of the scheduler.

src/dune_threaded_console/dune_threaded_console.ml
```diff
@@ -62,7 +62,10 @@ let make (module Base : S) : (module Dune_console.Backend) =
       Base.start ();
       Dune_engine.Scheduler.spawn_thread @@ fun () ->
       if not Sys.win32 then
-        ignore (Unix.sigprocmask SIG_UNBLOCK [ Signal.to_int Winch ] : int list);
+        ignore
+          (Unix.sigprocmask SIG_UNBLOCK
+             (List.map ~f:Signal.to_int [ Winch; Cont; Tstp ])
```
snowleopard, line 67: Can we avoid writing this list twice?

rgrinberg: I don't mind, but it would have to go somewhere odd like `Dune_util`. As we don't want the console to depend on the engine.

src/dune_engine/scheduler.ml
```diff
@@ -63,7 +63,7 @@ let interrupt_signals : Signal.t list = [ Int; Quit; Term ]
 
 (* In addition, the scheduler also blocks some other signals so that only
    designated threads can handle them by unblocking *)
-let blocked_signals : Signal.t list = Winch :: interrupt_signals
+let blocked_signals : Signal.t list = Cont :: Tstp :: Winch :: interrupt_signals
```
snowleopard, line 66: Can we document some reasons for why certain signals are in this list, maybe linking to the corresponding issues?

rgrinberg: There's no particular issues (because nobody is using tui yet), but I did leave a comment for each and every single describing the need to block it.

Alizter:
Cursor is still missing for me on suspend.

snowleopard (approved):
The scheduler changes look good!

rgrinberg:
Merging. Even though this doesn't seem to be a complete fix, it seems to work in at least two setups that I've tried. Also, it's clear that it's on the right path as it's least trying to handle these things. The old code didn't even bother which is even more wrong.

PR #7764, opened 2023-05-19
refactor(engine): remove dependency on dune lang version

rgrinberg:
The dune version is the property of the dune language and the engine
shold be independent of it. Instead, we set the property
(set_should_remove_write_permissions_on_generated_files) that depends on
it explicitly


Alizter (approved):

rgrinberg:
ping @snowleopard

snowleopard (approved):
Looks good!

PR #7769, opened 2023-05-20
refactor: make Action.File_perm.t a polymorphic variant

rgrinberg:
This allows to drop the dependency on Dune_lang.Action.File_perm in the
engine

### DIFF
--- a/src/dune_engine/action.ml
+++ b/src/dune_engine/action.ml
@@ -1,12 +1,13 @@
 open Import
 module Ext = Action_intf.Ext
-module File_perm = File_perm
 module Outputs = Outputs
 module Inputs = Inputs
 
 module type T = sig
   type t
 end
+
+module File_perm = Action_intf.File_perm
 
 module Make
     (Program : T)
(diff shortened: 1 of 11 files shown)


src/dune_engine/action.ml
```diff
@@ -1,13 +1,14 @@
...
 module Ext = Action_intf.Ext
-module File_perm = File_perm
 module Outputs = Outputs
 module Inputs = Inputs
 
 module type T = sig
   type t
 end
 
+module File_perm = Action_intf.File_perm
```
Alizter, line 10: Why not put this above after Ext?

Alizter (approved):

snowleopard (commented):
I'm not quite convinced by the motivation: couldn't we simply move the type to Stdune? It seems general enough to be there.

rgrinberg:
> I'm not quite convinced by the motivation: couldn't we simply move the type to Stdune? It seems general enough to be there.

Sure, that's a valid alternative to breaking this dependency. I would even prefer `Dune_util` over `Stdune` to keep this code outside of public packages.

rgrinberg:
I'm going to change this PR to just move everything into Dune_util and break the cycle once and for all.

Issue #265, opened 2017-09-23
Diagnose missing target errors


timbertson:
I've been converting a few of my own libraries to use jbuilder. It's gone fairly well so far, so thanks :) (and sorry for the influx of issues, I'll quiet down soon I promise ;))

But a number of times I've hit a wall when jbuilder says "nope", but I can't find a way to get more information about what it tried or why it couldn't do what I asked. Here's an example:

```
$ jbuilder build _build/default/examples/todomvc/main.bc.js
Don't know how to build _build/default/examples/todomvc/main.bc.js
```

Well.. there _should_ be a build target there. In these sorts of cases I'd love some flags which enable logging about:

 - what `jbuild` files it's loading
 - what targets are defined in the given directory
 - if I'm getting really desperate, what targets are defined in the entire workspace

In this case, the only troubleshooting I could do was:

 - `ls examples/todomvc/main.ml` (yep, it's there)
 - `cat examples/todomvc/jbuild` (yep, it defines an executable called "main")

At which point I'm pretty much out of ideas.

In the end, I figured out that I had created a bad `jbuild-workspace` - it was empty (which is documented as equivalent to `(context default)`, but maybe that's not implemented in beta12?). I expect doing any sort of dump of "which targets are defined" would have more quickly pointed to a project-wide issue, as it would have been empty.

I eventually figured it out by trying other targets which had previously worked, and they could no longer be found. This process wasn't obvious either though, since top-level targets still seemed to kind of work (until I did a `clean`, then they became suspicious):

 - `jbuilder build vdoml.install` - successfully built the install file, but didn't build any of the files it depends on
  - `jbuilder build @install` - didn't appear to build anything (but didn't fail either)

diml:
Yh, the empty `jbuild-workspace` equivalent to `(context default)` is not released yet. Hopefully we should never end up with a system with 0 rules...

When you request a target that doesn't exist from the command line, jbuilder try to give a hint when it can (Did you mean ...?).

diml:
BTW, one thing that is close to what your are asking is `jbuild rules`.

timbertson:
"Did you mean" is neat, but it's probably only going to work for typos - if it only gives you zero or one suggestion, it's not likely to be a good debugging aid, where you often want to scan the list of what's available to pick out patterns or surprises.

I tried `jbuilder rules`, but it wasn't happy:

```
$ jbuilder rules
Error: External library "utop" not found.
-> required by "src/.utop/jbuild (context default)"
[...]

PR #1755, opened 2019-01-12
Add a "dune targets" command to list available targets

gpetiot:
Add a new command `dune targets` as suggested in issue #265 

- ~the `DIRS` parameters of the command allows to filter the targets, if `DIRS=d1,d2`, only lists the targets in the following directories:~
  * ~_build/default/d1/*~
  * ~_build/default/d2/*~

PR #7004, opened 2023-02-05
feature: Suggest candidates when alias not found

Alizter:
We now suggest candidates when an alias is not found. Error messages look like:
```console
$ dune build @fou
Error: Alias "fou" specified on the command line is empty.
It is not defined in . or any of its descendants.
Hint: did you mean fmt or foo?
[1]
```
- [x] changelog
- [x] tests

PR #7770, opened 2023-05-21
feature: dune show targets and dune show aliases

Alizter:
Add new `dune show targets` and `dune show aliases` commands that work similarly to `ls` displaying targets and aliases respectively in the current directory or instead any directories given as arguments. They do not show targets or aliases in subdirectories or inside directory targets. Directory targets will be printed with a trailing path separator to distinguish them however.

- fixes https://github.com/ocaml/dune/issues/265

and continuing
- https://github.com/ocaml/dune/pull/1755

part of general effort in:
- #7784 

This is quite useful in cram tests to make sure dune is willing to produce the correct targets.

- [x] changelog
- [x] docs (more cli docs too)
- [x] display directory targets ending with an `/`


Alizter:
@emillon Since this is a relatively straightforward addition, it could perhaps be considered for 3.8.

Alizter:
@nojb I have something similar giving hints to aliases: https://github.com/ocaml/dune/pull/7004. However due to the way we lookup aliases we have to do it for all contexts at the same time.

Alizter:
@nojb Actually I went ahead and implemented an `--aliases` option that shows aliases in addition. Is this what you had in mind?

I'm actually unsure if we want this in the targets command or we wish to have a separate command for displaying aliases. In any case if we are undecided we can always drop that commit.

nojb:
> @nojb Actually I went ahead and implemented an `--aliases` option that shows aliases in addition. Is this what you had in mind?

I think so, yes, thanks! What I had in mind is a beginner; I would imagine most of the time such an user will be more interested in invoking an alias than a target file.

snowleopard:
That's a very useful command but I feel like it shouldn't be top-level. Don't we usually put these things under `dune describe`? 

I don't like the name "describe" and would prefer something like `dune show` instead, but that's a detail. I think we can end up polluting the top-level namespace if we keep adding commands like `dune targets`.

I

nojb:
> I don't like the name "describe" and would prefer something like `dune show` instead, but that's a detail. I think we can end up polluting the top-level namespace if we keep adding commands like `dune targets`.

`dune show` seems like a good top-level name for the kind of functionality. Why not add `dune show` and put the command under it? (`dune show targets`). Existing commands could be migrated under it as well later: `dune show rules` instead of `dune rules`, `dune show env` instead of `dune printenv`, `dune show workspace` instead of `dune describe workspace`, etc

snowleopard:
> `dune show` seems like a good top-level name for this kind of functionality. Why not add `dune show` and put the command under it? (`dune show targets`). Existing commands could be migrated under it as well later: `dune show rules` instead of `dune rules`, `dune show env` instead of `dune printenv`, `dune show workspace` instead of `dune describe workspace`, etc

@nojb Sounds like a great plan to me!

snowleopard:
> I'm actually unsure if we want this in the targets command or we wish to have a separate command for displaying aliases. In any case if we are undecided we can always drop that commit.

@Alizter If we go with the `dune show` plan, then I'd prefer aliases to be handled by `dune show aliases`.

Alizter:
@rgrinberg We can deprecate the bad name (error) and point to the new good name.

rgrinberg:
We shouldn't. The disturbance caused by this deprecation far outweighs whatever benefit can gained from it. Going through a deprecating cycle is a ton of churn for both the developers and the users. At best, we can introduce an alias like we did for `runtest`.

nojb:
> We shouldn't. The disturbance caused by this deprecation far outweighs whatever benefit can be gained from it.

Even if we don't aggressively deprecate the existing commands and just alias them under new names, I think that, in terms of CLI usability, it would still be an improvement migrating to `dune show` as a top-level name for all these commands. We could do a "soft" deprecation by not documenting the commands under the old names anymore, so that new users would be directed to the new names only. One could even imagine going one step further (if it is possible, I haven't checked) and remove the old names from the auto-generated `dune --help` page. Even if that is not possible we could clearly indicate in the help page that the `dune show` commands are to be preferred.

Alizter:
Regarding this PR: I've added a show subcommand and added aliases and a subcommand of that. So we have `dune show aliases` and `dune show targets`. Both have tests.

I'm drafting this PR for the time being so we can continue to discuss how best to group the commands. My thinking is that we really want to add these to `dune describe` and we want to rename `dune describe` to `dune show`. But let's continue that discussion in the issue.

snowleopard:
I just want to say that this feature was requested internally (there was particular interest for `dune show aliases`). I wonder if we can make progress here, somehow. Maybe add these commands both into `dune describe` and `dune show`? Alternatively, add them only to `dune describe` for now and create a ticket to think about a possible migration story towards `dune show`.

test/blackbox-tests/test-cases/show/dune-aliases.t/run.t
```diff
@@ -0,0 +1,113 @@
...
+Aliases in subdirectories should not be picked up.
+
+  $ mkdir subdir
+  $ cat > subdir/dune << EOF
+  > (alias
+  >  (name bar))
+  > EOF
+
+  $ dune show aliases
+  .:
```
snowleopard, line 37: I think I'd prefer the `dir:` not shown in the output because it makes it harder to write automation around `dune show aliases`. I'd imagine that in most cases users will want to filter it out.

snowleopard: (Also, I think it's just confusing to see the directory in the output, particularly `.:` is a pretty obscure string.)

test/blackbox-tests/test-cases/show/dune-targets-simple.t/run.t
```diff
@@ -0,0 +1,73 @@
...
+  simple2.a
+  simple2.cma
+  simple2.cmxa
+  simple2.cmxs
+  simple2.ml-gen
+
+The command also works with files in the _build directory.
+
+  $ dune show targets _build/default/
+  _build/default:
```
snowleopard, line 49: Same for `dune show targets`: I think we shouldn't print out the directory here.

rgrinberg:
> I just want to say that this feature was requested internally (there was particular interest for dune show aliases). I wonder if we can make progress here, somehow

Externally, it's easy to make progress. We just follow existing conventions.

Internally, we no longer import the command line from upstream. Nor do we have an existing `show` or `describe` command internally, so we can name this whatever we like. However, we would need to factor this out into something that can be used by both command lines.

snowleopard:
@rgrinberg Agreed. I suppose we can move the implementation into the engine: it may be useful not only for CLI but for a future specialised Dune RPC too.

(Maybe we need a separate library for stable functionality built on top of the build engine but not sure it's worth the extra complexity.)

rgrinberg:
> it may be useful not only for CLI but for a future specialised Dune RPC too.

We already have a library where we put shared rpc components that are independent of rules. So there's no problem there.

I was going to propose a `dune_cli` library that could include shared components that we can reuse in our cli's.

I personally wouldn't want to put this into the engine.

snowleopard:
Both options seem good. I wanted to suggest `dune_cli` too but this particular functionality seems useful as an RPC as well, so maybe we're looking for something more general than just `dune_cli`.

rgrinberg:
> I wanted to suggest `dune_cli` too but this particular functionality seems useful as an RPC as well, so maybe we're looking for something more general than just `dune_cli`.

In this concrete case, I would put in RPC. Because the shared components of the cli would already depend on the shared components of the RPC (`dune rpc build`, `dune shutdown`, `dune diagnostics` are all shared and need RPC).

snowleopard:
Sounds good to me!

test/blackbox-tests/test-cases/show/dune-aliases.t/run.t
Alizter: Sure , I just copied thr behaviour of `ls` herr.

snowleopard: Hmm, not sure which `ls` you mean: mine doesn't do that :) (And thanks!)

Alizter: For me `ls` only shows the headers when requesting multiple directories. I think I will use this behaviour since in the multiple case the headers have a use.

rgrinberg (changes requested):
> test generated targets (not worth investigating at this time)

What are generated targets?

Alizter:
> > test generated targets (not worth investigating at this time)
> 
> What are generated targets?

Nevermind, I was confused about something when I wrote it.

bin/describe/aliases_cmd.ml
```diff
@@ -0,0 +1,68 @@
+open Import
+open Stdune
```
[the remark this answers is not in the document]
Alizter: apparently not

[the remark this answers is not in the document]
Alizter: I opened #7994 to fix it tho

rgrinberg (changes requested):
I started doing some cleanups and realized there's a little too much copy pasting between `aliases` and `targets` for my tastes. Can you factor the code out?

Alizter:
@rgrinberg sure, I'll have a look over the weekend.

Alizter:
@rgrinberg I've factored the common code out, I'm not so sure about the API yet. I think there is some more cleanup we can do.

bin/describe/ls_like_cmd.mli
```diff
@@ -0,0 +1,8 @@
+open Import
+
+val term :
+     (   Dune_rules.Main.build_system
+      -> Path.Source.t
+      -> Path.t
+      -> string list Action_builder.t)
```
rgrinberg, line 7: These arguments need some labeling and/or documentation.

1. What's the Path.Source.t?
2. What's the Path.t?
3. What's the list of strings being returned?

rgrinberg (changes requested):

bin/describe/ls_like_cmd.mli
Alizter: They are all hidden details now.

Alizter:
I've just added a user error for when a directory is outside of the project.;

snowleopard:
Perhaps, `dune show sources` could be useful too: for symmetry, and for making it possible to check which files Dune actually considers sources. (Not always obvious.)

rgrinberg:
> Perhaps, dune show sources could be useful too

It could also be distinguished via some visual cue in `dune show targets`. We can still add `dune show sources` of course.

Let's save that for a subsequent PR though.

snowleopard:
Sure, happy for this to go to a separate PR.

PR #7773, opened 2023-05-22
Improve JSON string quoting in Dune tracing

rleshchinskiy:
We want to include more things in the trace file and we should always produce valid JSON here.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 Unreleased
 ----------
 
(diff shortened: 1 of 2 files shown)


Alizter:
Would it be worth adding some expect tests for the escaping?

snowleopard:
> Would it be worth adding some expect tests for the escaping?

I feel like the tests would basically follow the if statements in the code. It's difficult to come up with meaningful tests that wouldn't just duplicate the logic.

PR #7778, opened 2023-05-22
Trace more data

rleshchinskiy:
This allows Dune to generate trace files which contain the same information as .jenga/debug files about processes that Dune runs. This can be quite a bit of information (including stdout and stderr) so this is controlled by the `--trace-extended` flag.

### DIFF
--- a/bin/common.ml
+++ b/bin/common.ml
@@ -535,6 +535,7 @@ module Builder = struct
     ; store_digest_preimage : bool
     ; root : string option
     ; stats_trace_file : string option
+    ; stats_trace_extended : bool
     }
 
   let set_root t root = { t with root = Some root }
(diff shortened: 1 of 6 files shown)


bin/common.ml
```diff
@@ -851,6 +857,9 @@ module Builder = struct
             [ "display-separate-messages" ]
             ~doc:"Separate error messages with a blank line.")
     in
+    if Option.is_none stats_trace_file && stats_trace_extended then
+      User_error.raise
+        [ Pp.text "Thou shalt not pass --trace-extended without --trace" ];
```
Alizter, line 862: I would suggest saying this in a less commanding manner :)

Something like:
```
--trace-extended must be passed together with --trace
```
for consistency with our other messages.

rleshchinskiy: @rgrinberg?

emillon: I'm not a huge fan of adding religious references either, so I second @Alizter's suggestion.

emillon, line 862:
```suggestion
        [ Pp.text "--trace-extended must be passed together with --trace" ];
```

emillon:
I'd prefer not to delay the release with this; in the context of JS I don't think it matters too much if it's included in the release tarball.

bin/common.ml
Alizter: It's not the reference that bothers me but rather keeping messages understandable to users, especially for those who don't have English as a first language. I'm happy to improve this message later in another PR if you prefer however.

PR #7785, opened 2023-05-22
refactor(engine): decouple from Dune_lang

rgrinberg:
This requires the following 3 general moves:

* Moving predicate_lang into its own library
* Moving various action definitions from Dune_lang to Dune_util
* Moving the definition of alias names to Dune_util.

The last refactoring of the aliases isn't yet completed. I think we'd
rather have rules specific aliases such as fmt and install to live in
dune_lang rather than in the engine, but for now this is good enough.

Also, this refactoring caught some missing validation bugs in the alias
module. I've added some TODO's to address them eventually.

### DIFF
--- a/src/dune_engine/alias.ml
+++ b/src/dune_engine/alias.ml
@@ -1,92 +1,31 @@
 open Import
 
-module Name : sig
-  type t
+module Name = struct
+  include Dune_util.Alias_name
 
-  include Dune_sexp.Conv.S with type t := t
-
-  val equal : t -> t -> bool
-
-  val hash : t -> int
-
-  val compare : t -> t -> Ordering.t
-
-  val of_string : string -> t
-
-  val parse_string_exn : Loc.t * string -> t
-
-  val to_string : t -> string
-
-  val to_dyn : t -> Dyn.t
-
-  val default : t
-
-  val runtest : t
-
-  val install : t
-
-  val fmt : t
-
-  val all : t
-
-  val parse_local_path : Loc.t * Path.Local.t -> Path.Local.t * t
-
-  include Comparable_intf.S with type key := t
-end = struct
-  include String
-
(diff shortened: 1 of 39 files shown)


src/dune_engine/alias.ml
```diff
@@ -1,92 +1,33 @@
...
     | Some s -> s
     | None -> Code_error.raise "invalid alias name" [ ("s", Dyn.string s) ]
 
-  let to_string s = s
-
-  let default = "default"
+  let default = of_string "default"
 
-  let runtest = "runtest"
+  let empty = of_string "empty"
```
Alizter, line 15: Where did this come from?

src/dune_lang/action.mli
```diff
@@ -6,27 +6,28 @@
...
 
 module Diff : sig
+  open Diff
+
   module Mode : sig
-    type t =
-      | Binary  (** no diffing, just raw comparison *)
-      | Text  (** diffing after newline normalization *)
+    type t = Mode.t =
+      | Binary
```
Alizter, line 20: Why did you get rid of the comments?

rgrinberg: By accident. It's back now.

src/dune_util/alias_name.ml
```diff
@@ -0,0 +1,29 @@
+open Stdune
+include String
+
+(* DUNE3 once we get rid the "loose" validation, implement this module using
+   [Stringlike] *)
+let of_string_opt_loose s = Option.some_if (Filename.basename s = s) s
```
Alizter, line 6: Should we update this comment? Or maybe we can use stringlike?

Alizter (approved):
Seems ok to me.

snowleopard (approved):
Looks good to me, thanks!

Issue #7598, opened 2023-04-21
[coq] [coq-lang-1.0] Ensure test suite is comprehensive. 


ejgallego:
Before `(coq lang 1.0)` we need to ensure that we are testing all possible configurations of Coq Dune projects.

It seems there are quite a few, let's study the variables:
- composition: single theory, multiple same-scope ones, multiple theories in multiple workspaces
- Coq:  not present,  in workspace,  installed
- stdlib present: no, in workspace, installed
- user-contrib: no, installed
- coqpath: yes / no
- stdlib flag: yes / no
- theories visibility: public and private, can depend in 3 different ways x 2 different configurations (same / different scope)
- native: enabled / disabled in config, manually overriden.
- plugins: no / yes single module / yes depending on other libs
- plugins: living in coqpath without findlib registration / with findlib registration
- transitive deps: both in worksspace and in user-contrib
- Coq versions: maybe 8.13 / 8.16 / 8.17 ?
- library names: different kind of overlappings

That's a lot of combos to test, but can be reduced with a bit of care.

Another point to consider is how to have a naming scheme that help us identify what part of the test matrix the test is covering.

Alizter:
Regarding the naming, if we separate tests by directories then that won't be too much of a problem.

ejgallego:
> Regarding the naming, if we separate tests by directories then that won't be too much of a problem.

How would such directories be named? What I mean by naming scheme is a directory / test name that allows us to understand what variables on the matrix are covered.

palmskog:
I think we can add the use case where I discovered #8042:
- Coq library / multiple extracted modules / multiple executables

Specifically, in [Chapar](https://github.com/coq-community/chapar/blob/6377772b288fc1684ef01bdd61914460a5711797/ml/dune), there are three different key-value store executables produced from three different Coq modules. However, those executables share the same OCaml infrastructure/libraries, so it makes sense for them to be defined in the same `dune` file.

ejgallego:
Yes @palmskog , we would like to fix #8042

PR #7790, opened 2023-05-23
coq: test and fix duplicate theory prefix in COQPATH bug

Alizter:
This bug happens when using COQPATH so it won't affect most users. When having two theories A.B and A.C installed, dune will erroneously fail saying that the theory A has been defined twice.

I think the correct fix would be to merge the maps in the case where A has no immediate .vo files and is just a prefix.


ejgallego:
Good catch @Alizter ! I think indeed it should be enough to filter empty paths in `Coq_path`

ejgallego:
In fact I'm not sure if Coq's semantics do require to merge even non empty paths.

Alizter:
The package in question are mathcomp packages. They all get installed under mathcomp.* but in different paths for Nix. There is no problem in user-contrib however.

ejgallego:
I figured mathcomp would be the culprit, what I wonder is about the case:

```
foo/A/a.vo
foo/A/B
```
```
bar/A/b.vo
bar/A/C
```
What does Coq expect us to do here?

Alizter:
That case would not be consistent since the theory would have to both be A and we would be conflicting. The only time we can have A together is if it is a prefix and in that case it will be empty.

ejgallego:
Mmm, I'm afraid that `coqc` does accept that case, so I think we should, otherwise we may risk some compat problem. With this layout:
```
./foo/A/B/b.vo
./foo/A/a.vo
./bar/A/C/d.vo
./bar/A/c.vo
```
Coq has not problem with:
```
$ COQPATH=foo:bar coqtop
> From A Require Import a b c d.
```
However if we do add a duplicate `a.v`, very funny stuff starts to happen:
```
./foo/A/B/b.vo
./foo/A/a.vo
./bar/A/C/d.vo
./bar/A/a.vo
./bar/A/c.vo
```
Then `From A Require Import a` will pick the one that is first in `COQPATH`. Maybe that's intended?

Who knows, Coq semantics are messy.

That's a nice testcase actually.

Alizter:
Yeah I didn't doubt that Coq would accept it, but from the perspective of keeping user-contrib sane, having overlapping theories breaks the contract we make with the user about having a valid user-contrib.

In the future we can enforce all installs to be done as a COQPATH like for OCaml, since then we would have an atomic garuntee. That will also mean theories will need to be picked from packages.

Solving the original issue here by merging empty theory prefixes will give some restriction on valid COQPATHs that we can defo uostream to Coq. COQPATH is rarely used in Coq nowadays so a lack of validation shouldn'g be a guiding principle for us.

ejgallego:
> breaks the contract we make with the user about having a valid user-contrib.

What's the contract we make? Note that the first example for example doesn't seem invalid to me, if the semantics of COQPATH are unionfs alike.

> the original issue here by merging empty theory prefixes will give some restriction on valid COQPATHs that we can defo uostream to Coq

If that would work I'd be fine with it, but the "empty theory" restriction makes me afraid that it could be too restrictive in the real world. I'm wary of committing to a particular semantics when several are possible.

Maybe we should clarify the semantics upstream first?

Another semantics that would make sense is to actually merge all COQPATH + user-contrib in union-like way, but duplicate-free. I believe that Haskell works that way?

The use case seems important, as in math-comp, a package wants to extend a namespace, so I'm not sure the empty restriction is important here, as long as there are no conflicts.

ejgallego:
For 3.8.1 we can indeed filter empty prefixes, but it is not clear how some other setups may work with this restriction only (thinking of compcert + floc + vst) for example.

Alizter:
I suppose then rather than merging the prefix paths we should try to do as Coq does and select the first candidate. We could emit a warning when this is done.

If we have theories A.B and A.C then a user could put both in a Coq path. If A happens to be nonempty then we choose A of A.B whenever a user depends on a theory A. Otherwise we should be able to merge them no problem.

Alizter:
I've made it so that we always merge and prefer the first coq path theory.

ejgallego (approved):

ejgallego:
By the way we should encourage dune users to contribute to inspect and contribute to the dune test suite so their use cases are fully covered.

Alizter:
So one thing that this does change is the detection of duplicate theories that we had before. I suppose for now this is the best we can do if we want to keep as faithful as possible to Coq semantics, but I worry for our users who might have a silent overload of a theory.

Alizter:
I won't write any more tests for this since we plan to write a whole bunch of tests soon anyway. https://github.com/ocaml/dune/issues/7598

PR #7796, opened 2023-05-23
refactor: change representation of Predicate_lang.Glob

rgrinberg:
Represent it directly with a glob rather than going through a predicate.

This makes it simple, serializable, and safe to use for memoization.

### DIFF
--- a/otherlibs/dune-glob/test/dune_glob_tests.ml
+++ b/otherlibs/dune-glob/test/dune_glob_tests.ml
@@ -67,3 +67,12 @@ let%expect_test _ =
   [%expect {| [pass] "*" matches "foo.ml" == true |}];
   test glob "foo/" ~expect:false;
   [%expect {| [pass] "*" matches "foo/" == false |}]
+
+let%expect_test _ =
+  let glob = Glob.of_string "[!._]*" in
+  test glob ".foo" ~expect:false;
+  [%expect {| [pass] "[!._]*" matches ".foo" == false |}];
+  test glob "foo.ml" ~expect:true;
+  [%expect {| [pass] "[!._]*" matches "foo.ml" == true |}];
+  test glob "a" ~expect:true;
+  [%expect {| [pass] "[!._]*" matches "a" == true |}]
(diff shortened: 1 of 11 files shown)


src/dune_rules/sub_dirs.ml
```diff
@@ -79,11 +79,7 @@ let status status_by_dir ~dir : Status.Or_ignored.t =
   | Some d -> Status d
 
 let default =
-  let standard_dirs =
-    Predicate_lang.Glob.of_pred (function
-      | "" -> false
-      | s -> s.[0] <> '.' && s.[0] <> '_')
-  in
+  let standard_dirs = Predicate_lang.Glob.of_glob (Glob.of_string "[!._]*") in
```
emillon, line 82: Just checking: does that match single-character directories like `a`? (I'm not sure if `*` matches the empty string in that position)

rgrinberg: I'll add a unit test for this

rgrinberg: Indeed it works. See test.

rgrinberg:
ping @snowleopard

otherlibs/dune-glob/test/dune_glob_tests.ml
```diff
@@ -67,3 +67,12 @@ let%expect_test _ =
   [%expect {| [pass] "*" matches "foo.ml" == true |}];
   test glob "foo/" ~expect:false;
   [%expect {| [pass] "*" matches "foo/" == false |}]
+
+let%expect_test _ =
```
snowleopard, line 71: It's a bit odd that we test the semantics of a particular string, rather than the semantics of `standard_dirs`. If we ever change the definition of `standard_dirs`, this test isn't going to save us.

snowleopard (approved):
Looks good, though I guess that's not the whole story? I'd expect more stuff to be deleted, like useless IDs.

otherlibs/dune-glob/test/dune_glob_tests.ml
rgrinberg: It's still a useful test to make sure we don't change the syntax of glob.

Here's a proper test for this feature though: https://github.com/ocaml/dune/pull/7808

PR #7797, opened 2023-05-24
test: demonstrate `include_subdirs qualified` crash with modules OSL

anmonteiro:
- I tried using `(include_subdirs qualified)` in a library that defined `(modules :standard \ some_module)` and dune crashed
- this change attempts to fix the crash by implementing the `base` function for merging 2 module tries

fixes https://github.com/ocaml/dune/issues/7597
closes https://github.com/ocaml/dune/pull/7603


src/dune_rules/module_trie.ml
```diff
@@ -103,14 +103,25 @@ let rec to_dyn f t =
...
+          Map.add_exn acc k (Leaf leaf)
+        | Map m -> Map.add_exn acc k (Map (base ~path ~f m)))
+  in
   let rec loop path x y =
     match (x, y) with
     | None, None -> assert false
-    | Some x, None -> base path x
-    | None, Some x -> base path x
+    | Some x, None ->
+      base x ~path ~f:(fun path x -> f path (Some x) None |> Option.value_exn)
```
rgrinberg, line 120: Given that this function is provided by the caller, I don't think we can assume that it never returns `None`.

You could modify its signature if our only uses of it always guarantee there's a value available.

rgrinberg (changes requested):

src/dune_rules/module_trie.ml
anmonteiro: Good catch, the latest push handles the option case

rgrinberg (approved):
Thanks. a CHANGES entry would be appropriate as well.

PR #7799, opened 2023-05-24
chore: set latest heading to unreleased

anmonteiro:
must be an artifact of the latest release. 3.8.0 looks confusing when there's another just below.


rgrinberg (approved):

emillon:
Thanks, looks like the merge in #7778 used the wrong header indeed.

emillon (approved):

PR #7810, opened 2023-05-26
refactor: change representation of Predicate_lang.Glob

rgrinberg:
Represent it directly with a glob rather than going through a predicate. This makes it simple, serializable, and safe to use for memoization.

Remove all the `Predicate_with_id` foot guns.

@snowleopard  Continuation of #7796 along with the suggestion we discussed.

### DIFF
--- a/bin/print_rules.ml
+++ b/bin/print_rules.ml
@@ -60,7 +60,7 @@ let rec encode : Action.For_shell.t -> Dune_lang.t =
   | With_accepted_exit_codes (pred, t) ->
     List
       [ atom "with-accepted-exit-codes"
-      ; Predicate_lang.encode Dune_lang.Encoder.int (fun _ -> assert false) pred
+      ; Predicate_lang.encode Dune_lang.Encoder.int pred
       ; encode t
       ]
   | Dynamic_run (a, xs) ->
--- a/bin/print_rules.ml
+++ b/bin/print_rules.ml
@@ -60,7 +60,7 @@ let rec encode : Action.For_shell.t -> Dune_lang.t =
   | With_accepted_exit_codes (pred, t) ->
     List
       [ atom "with-accepted-exit-codes"
-      ; Predicate_lang.encode Dune_lang.Encoder.int pred
+      ; Predicate_lang.encode Dune_lang.Encoder.int (fun _ -> assert false) pred
       ; encode t
       ]
   | Dynamic_run (a, xs) ->
--- a/src/dune_engine/dep.ml
+++ b/src/dune_engine/dep.ml
@@ -311,8 +311,7 @@ module Set = struct
      depending on [(source_tree x)]. Otherwise, we wouldn't clean up stale
      directories in directories that contain no file. *)
   let dir_without_files_dep dir =
-    file_selector
-      (File_selector.create ~dir File_selector.Predicate_with_id.false_)
+    file_selector (File_selector.of_predicate_lang ~dir Predicate_lang.any)
 
   let of_source_files ~files ~empty_directories =
     let init =
--- a/src/dune_engine/dep.ml
+++ b/src/dune_engine/dep.ml
@@ -311,7 +311,7 @@ module Set = struct
      depending on [(source_tree x)]. Otherwise, we wouldn't clean up stale
      directories in directories that contain no file. *)
   let dir_without_files_dep dir =
-    file_selector (File_selector.of_predicate_lang ~dir Predicate_lang.any)
+    file_selector (File_selector.of_predicate_lang ~dir Predicate_lang.empty)
 
   let of_source_files ~files ~empty_directories =
     let init =
--- a/src/predicate_lang/predicate_lang.ml
+++ b/src/predicate_lang/predicate_lang.ml
@@ -116,26 +116,24 @@ let to_predicate (type a) (t : a Predicate.t t) ~standard : a Predicate.t =
       exec t ~standard (fun pred -> Predicate.test pred a))
 
 module Glob = struct
-  type glob = string -> bool
+  module Glob = Dune_glob.V1
 
-  type nonrec t = glob t
+  type nonrec t = Glob.t t
 
-  let to_dyn t = to_dyn (fun _ -> Dyn.string "opaque") t
+  let to_dyn t = to_dyn Glob.to_dyn t
 
-  let exec (t : t) ~standard elem = exec t ~standard (fun f -> f elem)
+  let exec (t : t) ~standard elem = exec t ~standard (fun f -> Glob.test f elem)
 
   let filter (t : t) ~standard elems =
     match t with
     | Inter [] | Union [] -> []
     | _ -> List.filter elems ~f:(fun elem -> exec t ~standard elem)
 
-  let create_glob g = Dune_glob.V1.test g
+  let of_glob g = Element g
 
-  let of_glob g = Element (create_glob g)
+  let of_string_list s =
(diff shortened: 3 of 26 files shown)


bin/print_rules.ml
```diff
@@ -60,7 +60,7 @@ let rec encode : Action.For_shell.t -> Dune_lang.t =
   | With_accepted_exit_codes (pred, t) ->
     List
       [ atom "with-accepted-exit-codes"
-      ; Predicate_lang.encode Dune_lang.Encoder.int pred
+      ; Predicate_lang.encode Dune_lang.Encoder.int (fun _ -> assert false) pred
```
snowleopard, line 63: That looks... odd? Maybe add a comment what this does/why?

rgrinberg: indeed it's a bit ugly, but I'm not sure what can be done about it. `encode` needs a function to encode the glob only for `string Predicate_lang.t`, but I don't think we can express that with the types.

src/predicate_lang/predicate_lang.mli
```diff
@@ -22,44 +23,38 @@ val not_union : 'a t list -> 'a t
...
 val any : 'a t
 
-val decode_one : 'a Decoder.t -> 'a t Decoder.t
+val decode_one : 'a Decoder.t -> ('a -> 'b t) -> 'b t Decoder.t
 
-val decode : 'a Decoder.t -> 'a t Decoder.t
+val decode : 'a Decoder.t -> ('a -> 'b t) -> 'b t Decoder.t
 
-val encode : 'a Encoder.t -> 'a t Encoder.t
+val encode : 'a Encoder.t -> (Dune_glob.V1.t -> Dune_sexp.t) -> 'a t Encoder.t
```
snowleopard, line 30: I think this type signature is not tight enough. The second argument should be parameterised by `'a` so that you only need to provide the function for the `'a = string` case. Or is it somehow not possible?

rgrinberg: Perhaps it's possible, but I'm not sure how to do it. There's no way constrain a constructor to be provided when `'a <> string`.

snowleopard: Maybe something like this could work?

```ocaml
type 'a element = 
  | Literal : 'a -> 'a element
  | Glob : string -> string element

type 'a t = 
  | Element : 'a element -> 'a t
  | ...

val encode : ('a element -> 'a Encoder.t) -> 'a t Encoder.t
```
Then, when `'a = string`, you'll need to handle both constructors and provide a suitable encoder for globs, but when `'a = int`, you'll only need to provide one.

rgrinberg: At this point, I don't think we need the gadt at all. We can go back to the original representation:

```ocaml
type 'a t =
   | Element of 'a
   | ...
```

And then instantiate `'a t` with just:

```ocaml
type element = Glob of Glob.t | Literal of string
```

And use either `int Predicate_lang.t` or `element Predicate_lang.t` without any `assert false`.

The problem is that we lose the more "unboxed" representation that we wanted to have.

snowleopard: Ah, of course, you're right.

Could we add a second type parameter? It would probably be a little ugly but maybe workable:

```ocaml
type has_no_globs = |
type has_globs = |

type ('a, 'b) t0 =
  | Element : 'a -> ('a, 'b) t
  | Glob : Dune_glob.V1.t -> (string, has_globs) t
  | Compl : ('a, 'b) t -> ('a, 'b) t
  ...

type 'a t = ('a, has_no_globs) t

let encode encoder = ...

module With_globs = struct
  type 'a t = ('a, has_globs) t

  let encode encoder glob_to_sexp = ...
end
```

rgrinberg: Right, that's a possibility. Or we could just not have a polymorphic encoding function at all and just have encoders for `string t` and `int` because that's all we use. Anyway, let's think it over tomorrow.

src/predicate_lang/predicate_lang.ml
```diff
@@ -20,7 +20,7 @@ let union a = Union a
 
 let not_union a = compl (union a)
 
-let any = not_union []
+let any = Compl (Union [])
```
snowleopard, line 23: Could be `Compl empty` if moved below. Also: I'm wondering if this deserves a separate constructor.

snowleopard: (Also, could be `Inter []` for symmetry?)

src/dune_engine/dep.ml
```diff
@@ -311,8 +311,7 @@ module Set = struct
      depending on [(source_tree x)]. Otherwise, we wouldn't clean up stale
      directories in directories that contain no file. *)
   let dir_without_files_dep dir =
-    file_selector
-      (File_selector.create ~dir File_selector.Predicate_with_id.false_)
+    file_selector (File_selector.of_predicate_lang ~dir Predicate_lang.any)
```
snowleopard, line 314: Should this be `Predicate_lang.empty`?

snowleopard: (If yes, how come ho test caught it?)

rgrinberg: Fixed it. I think it's because we only used it on directories we already knew are empty. So it was just a little perf. optimization. Any glob can be used here.

src/predicate_lang/predicate_lang.ml
```diff
@@ -77,65 +77,111 @@ and many f k acc =
 
 and decode f = many f union []
 
-let rec encode f =
+let rec encode f t =
   let open Encoder in
-  function
+  match t with
```
snowleopard, line 82: There are still some left-over diffs here that would be nice to clean up.

src/predicate_lang/predicate_lang.mli
```diff
@@ -30,36 +30,38 @@ val encode : 'a Encoder.t -> 'a t Encoder.t
...
   val exec : t -> standard:t -> string -> bool
 
   val filter : t -> standard:t -> string list -> string list
 
-  val create_glob : Dune_glob.V1.t -> glob
-
   val of_glob : Dune_glob.V1.t -> t
 
-  val of_pred : (string -> bool) -> t
+  val of_string_list : string list -> t
```
snowleopard, line 56: What does this match? Any of the strings in the list? It would be nice to add some comments.

src/predicate_lang/predicate_lang.ml
rgrinberg: done

src/predicate_lang/predicate_lang.ml
```diff
@@ -95,47 +95,88 @@ let rec to_dyn f =
...
-let rec exec t ~standard elem =
-  match (t : _ t) with
-  | Compl t -> not (exec t ~standard elem)
-  | Element f -> elem f
-  | Union xs -> List.exists ~f:(fun t -> exec t ~standard elem) xs
-  | Inter xs -> List.for_all ~f:(fun t -> exec t ~standard elem) xs
-  | Standard -> exec standard ~standard elem
-
-let rec map t ~f =
+let rec test t ~standard ~test:check elem =
```
snowleopard, line 98: You could define it as `test_` and then have `let test = test_` to avoid renaming things back and force.

rgrinberg: Done.

snowleopard (commented):
Looks good, thanks!

snowleopard (approved):

PR #7812, opened 2023-05-26
refactor: create Diagnostic.to_user_message so it can be reused

catern:
Another patch wanted to use this function, so move it out of private bin/diagnostics.ml to somewhere it can be reused.


rgrinberg:
Could you add a DCO?

PR #7830, opened 2023-05-30
Test errors for invalid opam repositories

gridbugs:
This also expands the cases in which an opam repository is considered invalid and handles some additional exceptions that can be raised while opam is parsing the repo file.

### DIFF
--- a/src/dune_pkg/opam.ml
+++ b/src/dune_pkg/opam.ml
@@ -6,12 +6,53 @@ module Repo = struct
 
   type t = { packages_dir_path : Filename.t }
 
+  let minimum_opam_version = OpamVersion.of_string "2.0"
+
   let validate_repo_file opam_repo_dir_path =
-    try
-      OpamFilename.raw (opam_repo_dir_path / "repo")
-      |> OpamFile.make |> OpamFile.Repo.read |> ignore
-    with OpamSystem.Internal_error message ->
-      User_error.raise [ Pp.text message ]
+    let opam_repo_file_path = opam_repo_dir_path / "repo" in
+    let repo =
+      try
+        OpamFilename.raw opam_repo_file_path
+        |> OpamFile.make |> OpamFile.Repo.read
+      with
+      | OpamSystem.Internal_error message ->
+        User_error.raise [ Pp.text message ]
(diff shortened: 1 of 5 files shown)


rgrinberg:
Tests need to be promoted.

LGTM, but I don't really know why we're doing. If it's necessary, surely the opam API would do this validation for us. Otherwise, we're doing a check for something that we end up ignoring down the line anyway.

gridbugs:
> LGTM, but I don't really know why we're doing this. If it's necessary, surely the opam API would do this validation for us. Otherwise, we're doing a check for something that we end up ignoring down the line anyway.

There are two new error cases added here:
1. handling the `Bad_format _ | Bad_format_list _ | Bad_version _`
2. adding checks that the repo file contains a version number >= 2

1 is necessary to prevent dune crashing due to unhandled exception while opam parses an invalid repo file.

2 is not strictly necessary since opam is happy to parse an empty repo file. The repo file isn't actually used at all during package management. We only validate the repo file to help prevent a user accidentally pointing dune at an invalid opam repository directory. I added the opam-version constraint because of this comment: https://github.com/ocaml/dune/pull/7732#discussion_r1194852940. @Leonidas-from-XIV do you think it's reasonable to allow an empty repo file since opam can parse such a file without error?

src/dune_pkg/opam.ml
```diff
@@ -6,12 +6,46 @@ module Repo = struct
...
+    | None ->
+      User_error.raise
+        [ Pp.textf
+            "The file %s lacks an \"opam-version\" field. (Hint: add \
+             `opam-version: \"2.0\"` to the file.)"
+            (String.maybe_quoted opam_repo_file_path)
+        ]
+    | Some opam_version ->
+      if OpamVersion.compare opam_version minimum_opam_version < 0 then
+        User_error.raise
```
Leonidas-from-XIV, line 41: maybe a hint would be useful here as well.

Leonidas-from-XIV (commented):
Looks good to me as well.

rgrinberg:
Merging. My objections aren't essential, I just think we're doing something weird.

PR #7832, opened 2023-05-30
Lock directory regeneration safety

### LINKED ISSUE #7807: Lock directory regeneration safety
Before deleting, how about we make sure this is a real lock file directory if it exists to avoid unfortunate accidents. For example, we check that it has a `dune.lock` file.

_Originally posted by @rgrinberg in https://github.com/ocaml/dune/pull/7732#discussion_r1206526598_

gridbugs:
Fixes https://github.com/ocaml/dune/issues/7807

Also added a flag to `dune pkg lock` for passing a custom lock directory path rather than the default of `dune.lock` because it made it easier to test this change and we'll need it eventually anyway.

### DIFF
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -2,6 +2,42 @@ open Stdune
 open Import
 module Lock_dir = Dune_pkg.Lock_dir
 
+(* Takes a string containing a possibly absolute path and returns a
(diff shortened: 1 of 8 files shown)


rgrinberg:
I think you forgot to promote the tests.

gridbugs:
> I think you forgot to promote the tests.

Woops! Fixed

PR #7835, opened 2023-05-30
feature(pkg): lock file configuration in workspace

rgrinberg:

### DIFF
--- a/src/dune_rules/workspace.ml
+++ b/src/dune_rules/workspace.ml
@@ -260,9 +260,16 @@ module Context = struct
   end
 
   module Default = struct
-    type t = Common.t
+    type t =
+      { base : Common.t
+      ; lock : Path.Source.t option
+      }
 
-    let to_dyn = Common.to_dyn
+    let to_dyn { base; lock } =
+      Dyn.record
+        [ ("base", Common.to_dyn base)
+        ; ("lock", Dyn.(option Path.Source.to_dyn) lock)
+        ]
 
     let decode =
       let+ common = Common.decode
@@ -270,7 +277,14 @@ module Context = struct
         field_o "name"
           ( Dune_lang.Syntax.since syntax (1, 10) >>= fun () ->
             Context_name.decode )
+      and+ lock =
+        (* TODO
+           1. guard before version check before releasing
+           2. allow external paths
+        *)
+        field_o "lock" (Dpath.Local.decode ~dir:(Path.source Path.Source.root))
       in
+      let lock = Option.map lock ~f:Path.as_in_source_tree_exn in
       fun ~profile_default ~instrument_with_default ~x ->
         let common = common ~profile_default ~instrument_with_default in
         let default =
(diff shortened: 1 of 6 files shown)


src/dune_rules/workspace.ml
```diff
@@ -260,17 +260,31 @@ module Context = struct
   end
 
   module Default = struct
-    type t = Common.t
+    type t =
+      { base : Common.t
+      ; lock : Path.Source.t option
```
gridbugs, line 265: Would it be simpler to drop the `option` and default this field to `Dune_pkg.Lock_dir.path` during parsing if no value is specified in the workspace? Is it ever desirable for dune to distinguish between the case where `lock` is omitted and the case where it's explicitly set to the default path?

gridbugs (approved):

gridbugs:
Thinking about how this interacts with lockdir generation. Since lockdirs are associated with contexts, would it make sense to have `dune pkg lock` take a context name so it knows where to put the generated lockdir rather than letting the user specify the lockdir path directly?

src/dune_rules/workspace.ml
rgrinberg: > Is it ever desirable for dune to distinguish between the case where lock is omitted and the case where it's explicitly set to the default path?

There's a couple of situations where the difference is useful:

1. A lock file that hasn't been explicitly specified and is absent is not an error. While a lock file that has been explicitly specified and is absent is an error.
2. When the types reflect exactly what the user has written, you can use them for serialization.

The 2nd point is a bit more general, so I tend to writing things in this style. However, 1. is the important one here.

gridbugs:
> > Thinking about how this interacts with lockdir generation. Since lockdirs are associated with contexts, would it make sense to have dune pkg lock take a context name so it knows where to put the generated lockdir rather than letting the user specify the lockdir path directly?
> 
> I'm thinking that our current scheme works alright. There's no special location to use for a lockdir for a particular context. They all read from the default location unless explicitly specified.

Just clarifying that our current scheme is to let the user pass a lockdir path to `dune pkg lock`, and it will assume the default (`dune.lock`) if no lockdir is specified.

I think we do need to change `dune pkg lock` to be aware of the lockdirs associated with a context though. E.g. if there is only one context in the workspace file and it uses a non-default lockdir location, running `dune pkg lock` with no arguments should create the lockdir at the path specified in the context instead of the default location. Also my understanding is that `dune build` will build its targets in all contexts so maybe it makes sense to add a flag to `dune pkg lock` to cause it to (re)generate lockdirs for all contexts to match.

rgrinberg:
>  E.g. if there is only one context in the workspace file and it uses a non-default lockdir location, running dune pkg lock with no arguments should create the lockdir at the path specified in the context instead of the default location

This behavior sounds like it's trying to guess the user's intent a little too much. I'm not opposed though. I don't yet have any concrete use cases for multi context/multi lockdir builds in mind.

>  Also my understanding is that dune build will build its targets in all contexts so maybe it makes sense to add a flag to dune pkg lock to cause it to (re)generate lockdirs for all contexts to match.

Not all commands are like that. For example, `dune describe` will use the default context unless other specified.

gridbugs:
> > E.g. if there is only one context in the workspace file and it uses a non-default lockdir location, running dune pkg lock with no arguments should create the lockdir at the path specified in the context instead of the default location
> 
> This behavior sounds like it's trying to guess the user's intent a little too much. I'm not opposed though. I don't yet have any concrete use cases for multi context/multi lockdir builds in mind.

Fair enough. Happy to leave it as is for now and refine the UX when we have a clearer idea of how this will be used in practice.

rgrinberg:
Okay so let's add `dune pkg lock --all` to the TODO list. Eventually we can talk about making that default.

PR #7836, opened 2023-05-30
fix(predicate_lang): [Inter []] should be [true]

rgrinberg:
Since [Union []] is [false], [Inter []] should be [true] for symmetry.


snowleopard (approved):
Looks right to me!

PR #7839, opened 2023-05-30
refactor(predicate_lang): make abstract and rename

rgrinberg:
Make the predicate language abstract and apply some simplicifations when
[Union]/[Inter] are created.

Rename [any] and [emtpy] to [true_] and [false_]

### DIFF
--- a/src/predicate_lang/predicate_lang.ml
+++ b/src/predicate_lang/predicate_lang.ml
@@ -5,78 +5,91 @@ type 'a t =
   | Element of 'a
   | Compl of 'a t
   | Standard
-  | Union of 'a t list
-  | Inter of 'a t list
+  | Or of 'a t list
+  | And of 'a t list
+
+let element a = Element a
+
+let of_list xs = Or (List.rev_map ~f:element xs)
+
+let standard = Standard
 
 let diff a = function
-  | Union [] -> a
-  | Inter [] -> Union []
-  | b -> Inter [ a; Compl b ]
-
-let inter a = Inter a
+  | Or [] -> a
+  | And [] -> Or []
+  | b -> And [ a; Compl b ]
 
 let compl a = Compl a
 
-let union a = Union a
+let true_ = And []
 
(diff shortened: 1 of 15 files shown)


src/predicate_lang/predicate_lang.ml
```diff
@@ -8,62 +8,75 @@ type 'a t =
...
-             list and doesn't start with - or:"
+let rec decode_one =
+  let not_union a = compl (union a) in
+
+  fun f ->
+    let open Decoder in
+    let bool_ops () =
+      sum
+        [ ("or", many f union [])
+        ; ("and", many f inter [])
```
snowleopard, line 46: Looks like we should just call these combinators `or` and `and` instead of `union` and `inter`.

snowleopard (approved):
Thanks! As I wrote in a comment, we can take the renaming a bit further.

snowleopard (approved):
Thanks! As I wrote in a comment, we can take the renaming a bit further.

src/predicate_lang/predicate_lang.ml
rgrinberg: Done. Although it's `or_` and `and_` because and/or are keywords.

src/predicate_lang/predicate_lang.ml
```diff
@@ -5,103 +5,116 @@ type 'a t =
...
-      User_error.raise ~loc
-        [ Pp.text ":include isn't supported in the predicate language" ]
-    | "or" | "and" | "not" -> bool_ops ()
-    | s when s <> "" && s.[0] <> '-' && s.[0] <> ':' ->
-      User_error.raise ~loc
-        [ Pp.text
-            "This atom must be quoted because it is the first element of a \
-             list and doesn't start with - or:"
+let rec decode_one =
+  let not_or a = compl (Or a) in
```
snowleopard, line 39: You forgot `compl` -> `not_` :)

PR #7840, opened 2023-05-30
Odoc rules: Generate full docs

jonludlam:
This PR introduces new rules to build the documentation for all of the
dependent libraries used by a project, including those installed in the switch rather than in the workspace.

The primary target for this is to ensure we have up-to-date correct and complete docs for the workspace, both for browsing and for tooling to be able to query.

A few decisions made during this:

1. Only libraries that are used by the projects in the workspace are documented.

There's no attempt made to document the entire switch. This has some
interesting consequences - for example, documentation files (mlds) are associated with packages rather than libraries and may well assume the existence of all their libraries.

2. Odoc file locations

The odoc files produced from the cmti/cmt/cmi files in the switch are put into 'equivalent' dirs under `_build` - specifically if a cmti file is in `<opam-switch>/lib/foo/bar` then the odoc file will be in `_build/default/_doc/odoc/external/foo/bar`. This is done to ensure consistency between odoc's view of the world and ocaml's, as odoc uses the same strategy as ocaml for finding modules, that is, an include path.

3. Non-dune packages

There's a bare-minimum level of effort put into handling non-dune packages. Essentially I've not tried to separate things into separate libraries, and the packages are just treated like a big bag of modules.

4. Unwrapped libraries

These are also treated like the non-dune packages - this is mainly because I couldn't see a good way to get a `Modules.t` value.

5. Private libraries

Since the primary target is for the developer working on the codebase, I'm building docs for the private libraries by default.

## Still to do:

This is a draft right now - I've not updated the tests. In addition, there's other work to be done - possibly in this PR or a follow-up.

- make the index pages more interesting (ie, more metadata to document).
- silence the warnings coming from files that aren't in the workspace
- Add a symlink from the previous docs output directory (or possibly a 403 page!)
- Do a better job of stdlib


rgrinberg:
Some minor remarks:

* We use American spelling elsewhere in the project. So `Artefact` -> `Artifact`.
* Our naming convention is `Foo_bar` over `FooBar`.

A couple more questions about  the generated docs:
* Why does bin_prot have its own directory? It's a dune project so I'm surprised it has a special install layout.
* Can we exclude private libraries with inline tests somehow? It doesn't make sense to include them I think

PR #7843, opened 2023-05-30
Added XDG_STATE_HOME support to xdg library

Skyb0rg007:
The [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) was updated to include a new directory: the user state directory. This PR adds support for said directory.


rgrinberg:
I don't see what use we have for this in dune, but I suppose it's harmless.

LGTM

PR #7844, opened 2023-05-31
suggest --prefix before specific dune install options

### LINKED ISSUE #7814: Error: "Missing mandir" (dune install)
## Expected Behavior

run 
```
 dune install
``` 
in a project folder in order to install the project and its built artefact in the default locations

## Actual Behavior

```
> dune install
Error: The mandir installation directory is unknown.
Hint: It could be specified with --mandir
```

## Specifications
```
> dune --version
3.7.0
> opam --version
2.1.4
> ocaml --version
The OCaml toplevel, version 4.14.1
```

## Additional information 

My default mandir directory exists, so no need for specifying a folder I think

```
ls /usr/local/share/man
de/   ja/   man1/ man3/ man5/ man7/ man8/
```

Here some additional info on mandir:

https://www.gnu.org/prep/standards/html_node/Directory-Variables.html

> ‘mandir’
>The top-level directory for installing the man pages (if any) for this package. It will normally be /usr/local/share/man, but you >should write it as $(datarootdir)/man. (If you are using Autoconf, write it as ‘@mandir@’.)

Alizter:
When a user has not set --prefix they will get a confusing error message
about mandir. The suggestion here is to also set --mandir leading them
down a rabbit hole of specifying a complete install layout. What they
probably want to be doing is speciying --prefix so we change the error
message here to suggest that instead.

fix https://github.com/ocaml/dune/issues/7814


Alizter:
OK so the test I made is not as reproducible as I had gathered. Any ideas how I can make it so? Or should I just get rid of the test?

rgrinberg:
@emillon can you have a look at this one?

bin/install_uninstall.ml
```diff
@@ -46,9 +46,13 @@ let get_dirs context ~prefix_from_command_line ~from_command_line =
     match v with
     | Some v -> v
     | None ->
+      (* We suggest that the user sets --prefix first rather than the specific
+         missing option, since this is the most common case. *)
```
emillon, line 50:
```suggestion
      (* We suggest that the user sets --prefix first rather than the specific
         missing option, since this is the most common case. *)
```

I'd remove the comment but I'm fine with it if you'd like to keep it.

test/blackbox-tests/test-cases/install/install-no-prefix.t
```diff
@@ -0,0 +1,22 @@
...
+  > EOF
+
+  $ cat > main.ml
+  $ cat > proj.opam
+
+  $ dune build @install
+
+We should suggest to the user here that they should use --prefix.
+
+  $ dune install
```
Alizter, line 19: This only works some of the time when dune doesn't detect an install location. I am not sure how to make this reproducible. Perhaps it is better to just remove this test. WDYT @emillon?

emillon: I believe that unsetting `OPAM_SWITCH_PREFIX` will make `dune install` require an explicit `--prefix`.

emillon (approved):

PR #7847, opened 2023-05-31
fix(coq): fix crash when COQ_NATIVE_COMPILER_DEFAULT missing 

### LINKED ISSUE #7846: [Coq] Dune crashes on version 3.8
## Reproduction

1. Make sure you are on Dune 3.8.
2. `git clone git@github.com:coq-tactician/coq-tactician.git`
3. `opam install . --deps-only`
4. `dune build`

Crashes with:
```
Error: exception Dune_rules__coq_config.Vars.E(_)
Raised at file "src/dune_rules/coq/coq_config.ml", line 37, characters 21-49
Called from file "src/dune_rules/coq/coq_config.ml", line 194, characters
  10-53
Called from file "otherlibs/fiber/src/core.ml", line 250, characters 36-41
Called from file "otherlibs/fiber/src/scheduler.ml", line 73, characters 8-11
-> required by ("gen-rules", In_build_dir "default/tests")
-> required by ("load-dir", In_build_dir "default/tests")
-> required by
   ("build-alias", { dir = In_build_dir "default"; name = "default" })
-> required by ("top-level", ())

```

CC @ejgallego @Alizter 

## Specifications

- Version of `dune` (output of `dune --version`): 3.8.0
- Version of `ocaml` (output of `ocamlc --version`): 4.09.1
- Operating system (distribution and version): Arch linux

Alizter:
Coq versions < 8.13 do not have COQ_NATIVE_COMPILER_DEFAULT and were causing dune to crash when searching for this var. We make it so that if dune does not find it, it defaults to "no". The docs should also be updated accordingly.

fix #7846

- [x] changelog
- [x] docs (need to mention native comp not supported < 8.13)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Unreleased
 ----------
 
+- Fix a crash when using a version of Coq < 8.13 due to the native compiler
+  config variable being missing. We now explicitly default to `(mode vo)` for
+  these older versions of Coq. (#7847, fixes #7846, @Alizter)
+
 - Remove some compatibility code for old version of dune that generated
   `.merlin` files. Now dune will never remove `.merlin` files automatically
   (#7562)
--- a/doc/coq.rst
+++ b/doc/coq.rst
@@ -132,7 +132,8 @@ The semantics of the fields are:
   Coq theories are located by Dune.
 
 - If Coq has been configured with ``-native-compiler yes`` or ``ondemand``, Dune
-  will always build the ``cmxs`` files together with the ``vo`` files.
+  will always build the ``cmxs`` files together with the ``vo`` files. This only
+  works on Coq versions after 8.13 in which the option was introduced.
 
   You may override this by specifying ``(mode native)`` or ``(mode vo)``.
 
--- a/src/dune_rules/coq/coq_config.ml
+++ b/src/dune_rules/coq/coq_config.ml
@@ -11,8 +11,6 @@ module Vars : sig
   val get_path_opt : t -> string -> Path.t option
 
   val of_lines : string list -> (t, User_message.Style.t Pp.t) result
-
-  exception E of User_message.Style.t Pp.t
 end = struct
   open Result.O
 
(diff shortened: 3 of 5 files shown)


erikmd:
Thanks @Alizter, but I don't think the fix is optimal.

Namely, `coqc` supports `-native-compiler yes` for 8.5 ≤ Coq ≤ 8.12 as well.

The only change is that starting from Coq ≥ 8.13, is that the `coqc` flag is optional provided it has been provided to `./configure` (e.g. via the [`coq-native`](https://opam.ocaml.org/packages/coq-native/) package flag).

For more context, see:

* https://github.com/coq/ceps/blob/master/text/048-packaging-coq-native.md#readme

erikmd:
To sum up, the expected behavior (already applied by `coq-makefile`) would be something like this:

* if Coq ≥ 8.13 has been compiled with a `-native-compiler …` flag, dune would automatically generate `.coq-native/*` files
* if 8.5 ≤ Coq ≤ 8.13, it would do so as well, but only if `(mode native)` is manually provided.

Cc @proux01 @ejgallego FYI

Alizter:
This won't mean that `(mode native)` will stop working for older versions of Coq however. What this changes is the detection of native from how coq was configured which I only know to be exposed via the configure variable as you said.

proux01 (approved):
LGTM, thanks

LasseBlaauwbroek:
This patch works for me. Thanks @Alizter.

christinerose (approved):

ejgallego (approved):
Thanks @Alizter !

I think we may have a problem in the API tho, it seems to me that for variables that may not be present, we should have the config record reflect this using `option`, and the rules can actually try to figure out what the right action to perform is based on the `coq lang` version and on the rules at play. I find this a bit more robust.

So if you want to tweak that, that's fine, not a big deal otherwise.

Alizter:
@ejgallego good point, I'll move the selection to the rules to separate concerns.

Issue #7717, opened 2023-05-12
melange: expose a mechanism to enable conditional compilation


jchavarri:
## Desired Behavior

While virtual libraries are a great mechanism to support conditional compilation, and Melange already supports them, sometimes it is useful to introduce branching at the module level.

For example, [cppo](https://github.com/ocaml-community/cppo) allows to check for defined values to conditionally compile some code.

One could generate a cppo rule like:

```
(rule
 (targets foo.ml)
 (deps
  (:first-dep foo.cppo.ml))
 (action
  (run %{bin:cppo} -D MELANGE %{first-dep} -o %{targets})))
```

Some other analog rule would be defined for native, but without the `-D MELANGE` part.

Unfortunately at the moment, it is not possible to express from a `dune` file which rules are applied only for Melange targets, or only for OCaml targets.

A similar feature was added for foreign stubs in https://github.com/ocaml/dune/pull/5649.

cc @davesnx @anmonteiro

anmonteiro:
I believe this has been superseded by #10222. Please reopen if not.

PR #7848, opened 2023-06-01
feat(melange): support `enabled_if` in `melange.emit`

anmonteiro:
tangentially related to https://github.com/ocaml/dune/issues/7717


jchavarri (approved):
We should include the new field in the `melange.emit` stanza documentation, either in this PR or a new one.

test/blackbox-tests/test-cases/melange/enabled_if.t
```diff
@@ -0,0 +1,55 @@
+`enabled_if` in `melange.emit`
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.8)
+  > (using melange 0.1)
```
rgrinberg, line 5: Is this still 0.1? If so, I don't think there's a need to version any changes.

test/blackbox-tests/test-cases/melange/enabled_if.t
```diff
@@ -0,0 +1,55 @@
...
+  > (melange.emit
+  >  (target out)
+  >  (emit_stdlib false)
+  >  (enabled_if %{bin-available:melc}))
+  > EOF
+  $ cat > x.ml <<EOF
+  > let () = Js.log "hello"
+  > EOF
+
+Not available until dune lang 3.9
```
rgrinberg, line 18: If this field needs versioning, it should be using the melange syntax rather than the dune syntax.

anmonteiro:
Removed the versioning. I can add docs in a follow-up.

rgrinberg (approved):

PR #7849, opened 2023-06-01
test(melange): demonstrate melange.emit doesn't respect `-p`

anmonteiro:
- `dune build -p some-package` was trying to build `melange.emit` stanzas belonging to other packages
- the fix is to teach the `stanza_package` function about the melange package field

### DIFF
--- a/src/dune_rules/melange/melange_rules.ml
+++ b/src/dune_rules/melange/melange_rules.ml
@@ -336,8 +336,8 @@ let setup_runtime_assets_rules sctx ~dir ~target_dir ~mode ~output ~for_ mel =
   in
   ()
 
-let setup_entries_js ~sctx ~dir ~dir_contents ~scope ~compile_info ~target_dir
-    ~mode (mel : Melange_stanzas.Emit.t) =
+let modules_for_js_and_obj_dir ~sctx ~dir_contents ~scope
+    (mel : Melange_stanzas.Emit.t) =
   let open Memo.O in
   (* Use "mobjs" rather than "objs" to avoid a potential conflict with a library
      of the same name *)
@@ -345,7 +345,7 @@ let setup_entries_js ~sctx ~dir ~dir_contents ~scope ~compile_info ~target_dir
     Dir_contents.ocaml dir_contents
     >>| Ml_sources.modules_and_obj_dir ~for_:(Melange { target = mel.target })
   in
-  let* modules =
+  let+ modules =
     let version = (Super_context.context sctx).ocaml.version in
     let* preprocess =
       Resolve.Memo.read_memo
@@ -358,16 +358,24 @@ let setup_entries_js ~sctx ~dir ~dir_contents ~scope ~compile_info ~target_dir
     in
     Modules.map_user_written modules ~f:(fun m -> Memo.return @@ pped_map m)
   in
+  let modules_for_js =
+    Modules.fold_no_vlib modules ~init:[] ~f:(fun x acc ->
+        if Module.has x ~ml_kind:Impl then x :: acc else acc)
+  in
+  (modules_for_js, obj_dir)
+
+let setup_entries_js ~sctx ~dir ~dir_contents ~scope ~compile_info ~target_dir
+    ~mode (mel : Melange_stanzas.Emit.t) =
+  let open Memo.O in
+  let* modules_for_js, obj_dir =
+    modules_for_js_and_obj_dir ~sctx ~dir_contents ~scope mel
+  in
   let requires_link = Lib.Compile.requires_link compile_info in
   let pkg_name = Option.map mel.package ~f:Package.name in
   let loc = mel.loc in
   let module_systems = mel.module_systems in
   let* requires_link = Memo.Lazy.force requires_link in
   let includes = cmj_includes ~requires_link ~scope in
-  let modules_for_js =
-    Modules.fold_no_vlib modules ~init:[] ~f:(fun x acc ->
-        if Module.has x ~ml_kind:Impl then x :: acc else acc)
-  in
   let output = `Private_library_or_emit target_dir in
   let obj_dir = Obj_dir.of_local obj_dir in
   let* () =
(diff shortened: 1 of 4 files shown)


test/blackbox-tests/test-cases/melange/melange-emit-package.t
```diff
@@ -0,0 +1,38 @@
+
```
jchavarri, line 1: Could include a brief description about the test, e.g.

```suggestion
Test that melange.emit targets are not included when using -p with packages they don't belong to.
```

jchavarri (approved):
I found similar code (`match` over stanzas type) in install_rules.ml ([here](https://github.com/ocaml/dune/blob/d4ceaa91ff88c59a16b9437046e9379e8fa38d34/src/dune_rules/install_rules.ml#L354-L355) and [here](https://github.com/ocaml/dune/blob/d4ceaa91ff88c59a16b9437046e9379e8fa38d34/src/dune_rules/install_rules.ml#L416-L417)).I wonder if `Melange_emit` should be included in those cases too.

anmonteiro:
@rgrinberg FWIW this was found in a real world use case (reason-react). The setup works like this:

- `reactjs-ppx.opam` defines a ppx that only depends on ppxlib
- `reason-react` is preprocessed with the react PPX, and defines its own melange tests with `melange.emit`
- when trying to run `dune build -p react-ppx` for a release, we get an error in the JS tests (package reason-react not found)

rgrinberg:
>  So what’s the package field in the emit stanza for?

No idea tbh. In executables, it's used to attach them to packages with a public name.

> when trying to run dune build -p react-ppx for a release, we get an error in the JS tests (package reason-react not found)

Is it because you're building `@all` instead of `@install`? Otherwise, I'm not sure how the melange.emit is being picked up by the `@install` stanza.

anmonteiro:
The test in this PR still fails if I change the command to `dune build -p my-ppx @install`.

Is the issue that `melange.emit` is being picked up, then?

rgrinberg:
Actually, it's just an issue with rule loading. The failure comes from the rule loading stage. The issue is that the melange rules  use `read_memo` which blows up when searching for `@install` dependencies in `test/`

anmonteiro:
Interesting. What's the best way to address this?

rgrinberg:
`Resolve.t` is just a result type, so you can just inspect it instead of assuming it has a value. In the case where something can't be resolved, we could add some dummy targets that will fail with `Action_builder.fail`. It's not perfect, but it's the best we can do without proper target masks.

anmonteiro:
OK, I pushed an alternative fix that also satisfies the test.

I'm not sure 1. what to name the dummy target 2. what to write inside `Action_builder.fail`, so would appreciate guidance there.

src/dune_rules/melange/melange_rules.ml
```diff
@@ -438,12 +438,27 @@ let setup_emit_js_rules ~dir_contents ~dir ~scope ~sctx mel =
...
   let+ () =
-    let* requires_link =
-      Lib.Compile.requires_link compile_info
-      |> Memo.Lazy.force >>= Resolve.read_memo
+    let* requires_link_resolve =
+      Lib.Compile.requires_link compile_info |> Memo.Lazy.force
     in
-    setup_js_rules_libraries ~dir ~scope ~target_dir ~sctx ~requires_link ~mode
-      mel
+    match Resolve.is_ok requires_link_resolve with
```
rgrinberg, line 444: `Resolve.to_result` can be used here.

src/dune_rules/melange/melange_rules.ml
```diff
@@ -438,12 +438,27 @@ let setup_emit_js_rules ~dir_contents ~dir ~scope ~sctx mel =
...
+      let loc = mel.loc in
+      (* NOTE: for monorepo scenarios where single packages can be built
+       * separately (with [dune build -p <PKG>]), we can't eagerly resolve the
+       * link-time closure for the [melange.emit] stanza since melange.emit
+       * targets aren't public (i.e. part of a package.)
+       * In the cases where resolution fails, we still add dummy targets to
+       * give Dune a chance to faile if these targets end up attached to a
+       * package installation. *)
+      Super_context.add_rule sctx ~dir ~loc ~mode
+        (Action_builder.fail { fail = (fun () -> assert false) }
```
rgrinberg, line 459: You can raise the equivalent of `Resolve.read_memo` inside `fail`. You might need to add a function to `Resolve` to raise `Resolve.error` though.

rgrinberg:
> 1. what to name the dummy target 

I would pick some target names that match what we would setup if the dependencies weren't absent. So for melange.emit, that could be the entry point .js files. The intent is to convert "missing rule for target" errors into "missing dependency" errors.

src/dune_rules/resolve.ml
```diff
@@ -41,6 +41,8 @@ let to_result x = x
 
 let of_error x = Error x
 
+let raise_error ({ exn; _ } : error) = raise exn
```
rgrinberg, line 44: Hmm, it's not right to drop the stack frames. They need to be included in the error somehow. Too bad we don't have any convenient api's to do the right thing.

There's the private `Report_error.format_memo_stacks` to turn these frames into something readable. I also don't think we need a general exception type for `Resolve.error`. Keeping it a `User_message.t` would allow us to easily append some more readable information.

We don't need to solve all of these issues right away I suppose.

rgrinberg (approved):

src/dune_rules/resolve.ml
anmonteiro: I pushed a 3rd commit that takes care of this. Let me know what you think. It's a bit more invasive than I'd like, and I chose to duplicate the stack frame printer function, but it seems to work well.

rgrinberg: Looks right, but it seems to bork the formatting of some errors. See tests.

rgrinberg: On second thought, there's an easy way to solve this problem:

1. introduce `push_frames : error -> unit Memo.t`
2. call this `Resolve.push_frames` before creating the rule for the dummy target that will fail.

I still like your refactoring in Resolve.t though. But that can go in a separate PR.

anmonteiro: Makes sense. Pushed the version with `push_frames`, will move the refactoring to a separate PR.

src/dune_rules/resolve.ml
```diff
@@ -41,6 +41,18 @@ let to_result x = x
...
+
+let push_frames { stack_frames; exn = _ } =
+  let rec loop = function
+    | [] -> Memo.return ()
+    | x :: rest ->
+      Memo.push_stack_frame
+        ~human_readable_description:(fun () -> Lazy.force x)
+        (fun () -> loop rest)
+  in
+  loop stack_frames
```
rgrinberg, line 54: Can you change error_to_memo to use this function?

anmonteiro: I tried to do that but a single frame went missing in a bunch of tests

rgrinberg: I pushed the change. Let's see what CI says.

Alizter:
Weird that thr Coq tests have lost some location info in the trace.

rgrinberg:
Mystery is solved

anmonteiro:
Thanks, Rudi.

emillon:
Will do, but this needs a changelog entry.

anmonteiro:
I don’t think it needs a changes entry since melange support is still 0.1

rgrinberg:
Didn't we announce melange support in dune? If we did, I think we should include updates in  CHANGES. They're informative to users and allow them to follow the development of the feature more easily - even if we aren't yet obligated to maintain backwards compatibility.

PR #7851, opened 2023-06-01
refactor: predicate lang

rgrinberg:
Individual constructors for true/false.

This gives a more compact representation for these common forms.

### DIFF
--- a/src/predicate_lang/predicate_lang.ml
+++ b/src/predicate_lang/predicate_lang.ml
@@ -2,6 +2,8 @@ open Stdune
 open Dune_sexp
 
 type 'a t =
+  | True
+  | False
   | Element of 'a
   | Not of 'a t
   | Standard
@@ -21,9 +23,9 @@ let diff a = function
 
 let not a = Not a
 
-let true_ = And []
+let true_ = True
 
-let false_ = Or []
+let false_ = False
 
 let or_ = function
   | [] -> false_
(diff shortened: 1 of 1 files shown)


src/predicate_lang/predicate_lang.ml
```diff
@@ -94,6 +96,8 @@ and decode f = many f or_ []
 let rec encode f =
   let open Encoder in
   function
+  | True -> encode f (Or [])
```
snowleopard, line 99: That seems odd: I'd expect `True -> encode f True` here.

rgrinberg: Did you make a typo? That would just be an infinite loop then.

It is odd, but I'm afraid it's unavoidable as we don't have a way to represent `true` or `false` as literals in the predicate language. We could introduce `:true` and `:false` for example though.

snowleopard: Oh, sorry, I meant mapping `True` and `False` to the corresponding constants in the predicate language, like `:true` and `:false`, for example.

snowleopard: My thinking is: if we allow `(or ())` in the language, we better also allow `(:true)` for clarity.

rgrinberg: Agreed, but this is actually a change that needs versioning. On the rules side, we must only accept this starting from 3.9. On the engine side, would we need to bump the action digest version because we aren't serializing `Predicate_lang.true_` in the same way?

snowleopard (approved):
Thanks!

src/predicate_lang/predicate_lang.ml
snowleopard: Right, we can do this later, of course.

PR #7863, opened 2023-06-02
Generate lockdir from current switch

gridbugs:
This makes it possible for `dune pkg lock` to solve dependencies using the current opam switch in the case where no path to an opam repository directory is specified.

### DIFF
--- a/src/dune_pkg/dune
+++ b/src/dune_pkg/dune
@@ -7,10 +7,12 @@
   dune_engine
   dune_util
   dune_lang
+  dune_console
   opam_core
   opam_repository
   opam_format
   opam_state
-  opam_0install)
+  opam_0install
+  fmt)
  (instrumentation
   (backend bisect_ppx)))
--- a/src/dune_pkg/opam.ml
+++ b/src/dune_pkg/opam.ml
@@ -1,7 +1,61 @@
 open Stdune
 module Package_name = Dune_lang.Package_name
 
-module Repo = struct
+module Global : sig
+  val with_switch_state :
+       switch_name:string
+    -> f:(OpamStateTypes.unlocked OpamStateTypes.switch_state -> 'a)
+    -> 'a
+end = struct
+  let initialized = ref false
+
+  let ensure_initialized () =
+    if not !initialized then (
+      OpamSystem.init ();
+      let root = OpamStateConfig.opamroot () in
+      let (_ : OpamFile.Config.t option) =
+        OpamStateConfig.load_defaults ~lock_kind:`Lock_read root
+      in
+      OpamFormatConfig.init ();
+      OpamCoreConfig.init ~safe_mode:true ();
+      OpamRepositoryConfig.init ();
+      OpamStateConfig.init ~root_dir:root ();
+      initialized := true)
+
+  let with_switch_state ~switch_name ~f =
+    ensure_initialized ();
+    try
+      OpamGlobalState.with_ `Lock_read (fun global_state ->
+          Dune_console.print
+            [ Pp.textf "Using opam installation: %s"
+                (OpamFilename.Dir.to_string global_state.root)
+            ];
+          let switch = OpamSwitch.of_string switch_name in
+          if not (OpamGlobalState.switch_exists global_state switch) then
+            User_error.raise
+              [ Pp.textf
+                  "There is no opam switch named %s. Run `opam switch list` to \
+                   see a list of switches."
+                  (String.maybe_quoted switch_name)
+              ];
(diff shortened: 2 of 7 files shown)


gridbugs:
My intention for this feature is to make it easy to generate lockdirs while we're developing dune package management. Eventually dune will manage its own copy of the opam repository so it won't be necessary to look at the current opam switch. At this point I intend for this feature to be removed.

I've tested this locally including in cases where opam is not initialized or improperly initialized but I haven't made any automated tests for this. To test this we would need a mock opam installation. @rgrinberg think it's worth adding a mock opam installation to the dune repo for testing purposes? Long term we won't need it as this feature is just to enable easy experimentation with dune package management until dune can manage its own copy of the opam repo.

src/dune_pkg/dune
```diff
@@ -7,10 +7,12 @@
...
   dune_util
   dune_lang
+  dune_console
   opam_core
   opam_repository
   opam_format
   opam_state
-  opam_0install)
+  opam_0install
+  fmt)
```
rgrinberg, line 16: Where do we use fmt?

rgrinberg:
Could you describe the use case a little more? Are the solution produced by this solver mode dependent on the packages you have installed? Or is it just the repositories/pins attached the switch?

src/dune_pkg/dune
gridbugs: Without this package it seems like the compiler doesn't know the definition of `Fmt.t` (`type 'a t = Format.formatter -> 'a -> unit`. If I remove this dependency I get an error:
```ocaml
File "src/dune_pkg/opam.ml", line 183, characters 18-32:
183 |       | Left l -> L.pp_rejection f l
                        ^^^^^^^^^^^^^^
Error: This expression has type L.rejection Fmt.t
       This is not a function; it cannot be applied.
```

gridbugs:
> Could you describe the use case a little more? Are the solution produced by this solver mode dependent on the packages you have installed? Or is it just the repositories/pins attached the switch?

It's just the repos attached to the switch. The use case is doing manual experiments or demos of package management in dune without pointing it at a local checkout of the opam repository (prior to this change you needed to always pass `--opam-repository-path` to `dune pkg lock`).

src/dune_pkg/opam.ml
```diff
@@ -1,7 +1,52 @@
...
+      OpamSystem.init ();
+      let root = OpamStateConfig.opamroot () in
+      let (_ : OpamFile.Config.t option) =
+        OpamStateConfig.load_defaults ~lock_kind:`Lock_read root
+      in
+      OpamFormatConfig.init ();
+      OpamCoreConfig.init ~safe_mode:true ();
+      OpamRepositoryConfig.init ();
+      OpamStateConfig.init ~root_dir:root ();
+      initialized := true)
```
rgrinberg, line 21: Do any of the calls above modify something on the file system?

rgrinberg:
OK, the feature sounds reasonable.

However, dune's package management is stateless compared to opam's, perhaps we should avoid surprises and ask the user to provide the switch name? It's a little more inconvenient, but this command is a temporary band-aid until we introduce our own repo management. We can allow it to be inconvenient.

src/dune_pkg/opam.ml
gridbugs: Skimming the code it looks like the `init` functions are just setting some mutable global config records. I copied the way that opam-monorepo initializes its opam state. cc @Leonidas-from-XIV

gridbugs:
> However, dune's package management is stateless compared to opam's, perhaps we should avoid surprises and ask the user to provide the switch name? It's a little more inconvenient, but this command is a temporary band-aid until we introduce our own repo management. We can allow it to be inconvenient.

Sounds reasonable. Will update.

gridbugs:
Updated to take a `--opam-switch` argument that accepts the name or path of the switch to use for solving dependencies.

src/dune_pkg/opam.mli
```diff
@@ -22,6 +12,15 @@ module Env : sig
   val global : unit -> t
 end
 
+module Repo_selection : sig
+  (** An opam repository *)
+  type t
+
+  val switch_with_name : string -> t
+
+  val local_repo_with_env : opam_repo_dir_path:Filename.t -> env:Env.t -> t
```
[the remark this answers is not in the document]
gridbugs: This has the same problem as we discussed here: https://github.com/ocaml/dune/pull/7732#discussion_r1194850182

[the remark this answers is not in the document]
rgrinberg: Might be worth to leave the comment in the source then.

PR #7865, opened 2023-06-02
refactor: move fiber to vendor

rgrinberg:
because it's now being developed in ocaml-dune/fiber


Alizter:
You are missing the vendor script.

PR #7868, opened 2023-06-02
Implementation of `OpamSysPoll` in Dune-terms

Leonidas-from-XIV:
This branch adds an implementation of the functionality of `OpamSysPoll` but using our own concurrency primitives.

### DIFF
--- a/src/dune_pkg/opam.ml
+++ b/src/dune_pkg/opam.ml
@@ -168,6 +168,13 @@ module Env = struct
 
   let find_by_name (t : t) ~name =
     OpamVariable.Map.find_opt (OpamVariable.of_string name) t
+
+  let add ~var ~value t =
+    let variable = OpamVariable.of_string var in
+    let content = OpamVariable.string value in
+    OpamVariable.Map.add variable content t
+
+  let union = OpamVariable.Map.union (fun _left right -> right)
 end
 
 module Local_repo_with_env = struct
@@ -325,9 +332,11 @@ module Repo_state = struct
 end
 
 module Repo_selection = struct
-  type t =
+  type pre =
     | Switch_with_name of string
     | Local_repo_with_env of Local_repo_with_env.t
+
+  type t = pre
 
   let switch_with_name switch_name = Switch_with_name switch_name
 
@@ -337,6 +346,13 @@ module Repo_selection = struct
           Local_repo.of_opam_repo_dir_path opam_repo_dir_path
       ; env
       }
+
+  let add_env env' = function
+    | Switch_with_name _ as rs -> rs
+    | Local_repo_with_env ({ env; _ } as local) ->
+      let env = Env.union env env' in
+      let local = { local with env } in
+      Local_repo_with_env local
 
   (* [with_state ~f t] calls [f] on a [Repo_state.t] implied by [t] and
      returns the result. Don't let the [Repo_state.t] escape the function [f].
--- a/src/dune_pkg/opam.mli
+++ b/src/dune_pkg/opam.mli
@@ -10,14 +10,24 @@ module Env : sig
 
   (** Create an environment matching that of the global opam installation. *)
   val global : unit -> t
+
+  (** Adds a value to the environment *)
+  val add : var:Env.Var.t -> value:string -> t -> t
+
+  (** [union l r] merges two environments together. On key clashes prefers the
+      value from [r]. *)
+  val union : t -> t -> t
 end
 
 module Repo_selection : sig
   (** An opam repository *)
   type t
 
+  (** A t that does not yet have an environment added. Use [add_env]. *)
+  type pre
+
   (** An opam repo associated with a switch of a given name or directory *)
-  val switch_with_name : string -> t
+  val switch_with_name : string -> pre
 
   (** An opam repo in a local directory given by [opam_repo_dir_path]. Note that
       [opam_repo_dir_path] is a [Filename.t] rather than, say, a
@@ -40,7 +50,9 @@ module Repo_selection : sig
         to their project, and runs: `dune pkg lock --opam-repository-path
         ../opam-repository`. This rules out [Path.Outside_build_dir.t] as it
         doesn't permit relative paths outside the source directory. *)
-  val local_repo_with_env : opam_repo_dir_path:Filename.t -> env:Env.t -> t
+  val local_repo_with_env : opam_repo_dir_path:Filename.t -> env:Env.t -> pre
+
+  val add_env : Env.t -> pre -> t
 end
 
 module Summary : sig
--- a/src/dune_pkg/sys_poll.ml
+++ b/src/dune_pkg/sys_poll.ml
@@ -1,0 +1,215 @@
+open Stdune
+module Process = Dune_engine.Process
+module Display = Dune_engine.Display
+
+let norm = function
+  | "" -> None
+  | s -> Some (String.lowercase s)
+
+let path_of_string s = s |> Path.External.of_string |> Path.external_
+
+let normalise_arch raw =
+  match String.lowercase_ascii raw with
+  | "x86" | "i386" | "i486" | "i586" | "i686" -> "x86_32"
+  | "x86_64" | "amd64" -> "x86_64"
+  | "powerpc" | "ppc" | "ppcle" -> "ppc32"
+  | "ppc64" | "ppc64le" -> "ppc64"
+  | "aarch64_be" | "aarch64" -> "arm64"
+  | a
+    when a = "armv8b" || a = "armv8l"
+         || List.exists
+              ~f:(fun prefix -> String.is_prefix ~prefix a)
+              [ "armv5"; "armv6"; "earmv6"; "armv7"; "earmv7" ] -> "arm32"
+  | s -> s
+
+let normalise_os raw =
+  match String.lowercase raw with
+  | "darwin" | "osx" -> "macos"
+  | s -> s
+
+let run_capture_line ~path ~prog ~args =
+  let prog = Bin.which ~path prog in
+  match prog with
+  | None -> Fiber.return None
+  | Some prog ->
+    let open Fiber.O in
+    let failure_mode = Process.Strict in
+    let display = Display.Quiet in
+    let+ res = Process.run_capture_line ~display failure_mode prog args in
+    norm res
+
+let uname ~path args = run_capture_line ~path ~prog:"uname" ~args
+
+let lsb_release ~path args = run_capture_line ~path ~prog:"lsb_release" ~args
+
+let arch ~path =
+  let open Fiber.O in
+  let+ raw =
+    match Sys.os_type with
+    | "Unix" | "Cygwin" -> uname ~path [ "-m" ]
+    | "Win32" ->
+      if Sys.word_size = 32 && not (OpamStubs.isWoW64 ()) then
+        Fiber.return @@ Some "i686"
+      else Fiber.return @@ Some "x86_64"
+    | _ -> Fiber.return None
+  in
+  match raw with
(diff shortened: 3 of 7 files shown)


src/dune_pkg/dune
```diff
@@ -4,6 +4,7 @@
  (libraries
   stdune
   fiber
+  memo
```
rgrinberg, line 7: I would stick to fiber everywhere. Makes the API more general. Memo is only for generating the rules.

Leonidas-from-XIV: But doesn't Memo have memorizing semantics? That seems to be useful as the result will never change across runs.

rgrinberg: Right, but that's for the concrete rules to decide. Also note that turning a fiber into a memo using `Memo.of_reproducible_fiber` isn't enough to give these semantics. You need to make sure the fiber isn't recomputed every time we bind on it by creating a cell.

Leonidas-from-XIV: So if I understand you correctly we'd determine the value in the rules and would then have them as a `Memo.t` at that point?

I've adjusted the code to use fibers.

rgrinberg: Basically. In general, we push the rule creation down to dune_rules and package management is no exception. There should be a core library that's more widely usable and doesn't use memoization. The rules can finally use memoization to setup the rules, but other parts of the code (e.g. the command line), can use the api differently should they choose to.

rgrinberg:
Do you want to try using this in the rules? Or in a separate PR?

Leonidas-from-XIV:
I'll try adding this into the rules!

src/dune_rules/pkg_rules.ml
```diff
@@ -1224,13 +1263,26 @@ let gen_rules context_name (pkg : Pkg.t) =
...
 let setup_package_rules context ~dir ~pkg_name :
     Build_config.gen_rules_result Memo.t =
   match Package.Name.of_string_user_error (Loc.none, pkg_name) with
   | Error m -> raise (User_error.E m)
   | Ok name -> (
+    (* TODO: PATH *)
+    let path = [] in
     let* db = Lock_dir.get context in
-    resolve db.packages context name >>| function
+    let sys_poll =
```
rgrinberg, line 1278: I think you can probably just make this global and use the path from `Env.initial`.

rgrinberg:
Added @gridbugs to review the solver related stuff

src/dune_rules/pkg_rules.ml
```diff
@@ -717,9 +747,12 @@ end = struct
         ; exported_env = []
         }
       in
-      let+ expander = Action_expander.expander ctx t in
-      t.exported_env <-
-        List.map exported_env ~f:(Action_expander.exported_env expander);
+      let* expander = Action_expander.expander ctx t in
+      let+ exported_env =
+        Memo.sequential_map exported_env
```
rgrinberg, line 752: Let's do a parallel map here

src/dune_rules/pkg_rules.ml
```diff
@@ -535,26 +569,22 @@ module Action_expander = struct
     let dir = Path.build expander.paths.source_dir in
     match action with
     | Run (exe, args) ->
-      let+ exe, more_args = Expander.expand_exe expander exe in
-      let args =
-        more_args @ List.concat_map args ~f:(Expander.expand_pform expander)
-        |> Value.L.to_strings ~dir
+      let* exe, more_args = Expander.expand_exe expander exe in
+      let+ args =
+        Memo.sequential_map args ~f:(Expander.expand_pform expander)
```
rgrinberg, line 574: parallel_map

src/dune_rules/pkg_rules.ml
```diff
@@ -432,30 +450,40 @@ module Action_expander = struct
       | Stublibs -> Path.relative roots.lib_root "stublibs"
       | Misc -> assert false
 
+    let sys_poll_var key =
+      let+ v = Memo.Lazy.force key in
+      match v with
+      | Some v -> [ Value.String v ]
+      | None ->
+        (* TODO *)
```
rgrinberg, line 458: Could you clarify this TODO a little? What should be done here ideally? When can we expect `None` here?

src/dune_pkg/sys_poll.ml
```diff
@@ -0,0 +1,213 @@
...
+
+let sys_env ~path =
+  let open Fiber.O in
+  let entry k f =
+    let+ v = f ~path in
+    (k, v)
+  in
+  let+ mappings =
+    Fiber.all
+      [ entry "arch" arch
```
rgrinberg, line 201: Does this mean we run `uname` more than once? If so, we should avoid that.

src/dune_rules/pkg_rules.ml
```diff
@@ -569,25 +599,25 @@ module Action_expander = struct
...
+          ~f:(fun acc ({ Env_update.op = _; var; value } as update) ->
+            let* env, updates = acc in
+            let+ value =
               let expander = { expander with env } in
-              Expander.expand_pform_gen expander value ~mode:Single
-              |> Value.to_string ~dir
+              let* value =
+                Expander.expand_pform_gen expander value ~mode:Single
+              in
+              value |> Value.to_string ~dir |> Memo.return
```
Alizter, line 620:
```suggestion
              Value.to_string value ~dir
```
and make the `let*` above a `let+`.

src/dune_rules/pkg_rules.ml
Alizter, line 574: parallel?

src/dune_rules/pkg_rules.ml
```diff
@@ -387,9 +387,27 @@ module Substitute = struct
     Action.Extension (module M)
 end
 
+type sys_poll =
+  { os_version : string option Memo.Lazy.t
+  ; os_distribution : string option Memo.Lazy.t
+  ; os_family : string option Memo.Lazy.t
+  }
```
rgrinberg, line 394: I would move this record to `Sys_poll` and then have a single function there to generate it . That way we can be sure that we aren't running the same thing twice.

src/dune_rules/pkg_rules.ml
```diff
@@ -569,25 +597,25 @@ module Action_expander = struct
...
     | Withenv (updates, action) ->
-      let+ action = expand action ~expander in
-      let _env, updates =
-        List.fold_left ~init:(expander.env, []) updates
-          ~f:(fun (env, updates) ({ Env_update.op = _; var; value } as update)
-             ->
-            let value =
+      let* action = expand action ~expander in
+      let+ _env, updates =
+        List.fold_left
```
Alizter, line 608: Memo.List.fold_left

src/dune_pkg/sys_poll.ml
rgrinberg: Okay, if it works just leave a TODO for now. We can revisit this later.

src/dune_rules/pkg_rules.ml
Leonidas-from-XIV: Ah, I was looking for exactly this! I overlooked that `Memo` includes `S` which has `List` when skimming through the mli.

src/dune_rules/pkg_rules.ml
Leonidas-from-XIV: Ok, I added it there. I am a bit hesitant about having global state there, but we can always revisit this later.

src/dune_rules/pkg_rules.ml
Leonidas-from-XIV: It I understand the OPAM codebase correctly, it will eventually call `opt_eval_to_bool` which converts `None` to `false` https://github.com/ocaml/opam/blob/f67683cb49171615a475b0b7493b3bcc67a0b585/src/format/opamFilter.ml#L392-L395 for all unknown variables.

In our case we would get `None` e.g. if our platform is some kind of Unix but `uname` can't be found.

src/dune_pkg/opam.ml
```diff
@@ -315,6 +326,13 @@ module Repo_selection = struct
       ; env
       }
 
+  let with_sys_env sys_env = function
+    | Switch_with_name _ as rs -> rs
```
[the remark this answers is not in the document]
Leonidas-from-XIV: No, the problem is that in a `Switch_context` it will [always call OPAM](https://github.com/ocaml-opam/opam-0install-solver/blob/a17703ba83374975633d66bea2d40da51beabee5/lib/switch_context.ml#L22) so it is not possible to inject custom variables in a `Switch_context`. In `Dir_context` it will call a [custom function](https://github.com/ocaml-opam/opam-0install-solver/blob/a17703ba83374975633d66bea2d40da51beabee5/lib/dir_context.ml#L66).

So we could 

* do nothing
* print a warning that that this is missing functionality in this branch
* submit a PR upstream to closer align the `Dir_context` and `Switch_context`

src/dune_pkg/opam.mli
```diff
@@ -52,6 +66,6 @@ module Summary : sig
 end
 
 val solve_lock_dir :
-     repo_selection:Repo_selection.t
+     repo_selection:Repo_selection.with_sys Repo_selection.t
```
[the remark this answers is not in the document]
Leonidas-from-XIV: `Repo_selection.t` is a pure value created by parsing command line arguments, so unfortunately at this point I can't inject the environment variables, since that would require evaluating a fiber. Therefore the injection has to be done in a later step. Hence why I added `with_sys_env` that (in an admittedly rather ugly way) patches in the env later (where possible). But that means that `Repo_selection.t` with variables and without would be indistinguishable. To alleviate this I introduced a phantom type to force the user to call `with_sys_env` to turn an `empty Repo_selection.t` into a `with_sys Repo_selection.t`.

But a better solution would be more than welcome, because I am rather unhappy with the code as it is now.

src/dune_pkg/opam.mli
```diff
@@ -10,14 +10,25 @@ module Env : sig
...
+
+  type empty
+
   (** An opam repository *)
-  type t
+  type 'a t
 
   (** An opam repo associated with a switch of a given name or directory *)
-  val switch_with_name : string -> t
+  val switch_with_name : string -> empty t
```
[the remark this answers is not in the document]
Leonidas-from-XIV: It doesn't. There is currently no way to inject custom variables into a `Switch_context` so it is `empty` and in `with_sys_env` we just give up and turn it into a `with_sys` anyway.

src/dune_pkg/opam.mli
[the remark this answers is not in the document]
gridbugs: One thing I would like to preserve is the `--opam-env=pure` behaviour where the solver is run on a completely empty environment for the purpose of reproducable tests. Also I was thinking of at some point letting the user specify variables to set and I think it would be useful to give the option to set variables starting from a totally empty environment.

src/dune_rules/pkg_rules.ml
rgrinberg: Thanks. Could you include it in the comment though.

src/dune_pkg/opam.ml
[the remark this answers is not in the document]
Leonidas-from-XIV: For now we decided to not do anything. As we have learned in an in-person discussion with the OPAM maintainers, the opam-repository is not actually stored in an unpacked format by any stable version of OPAM at the moment, so we can't use `Dir_context` and point it at the unpacked opam-repository of the switch. Given this support is temporary until we implement our own opam-repository fetching this is probably good enough for now.

Leonidas-from-XIV:
I've rebased again and squashed the commits. Anything missing still or is this ready to go?

PR #7873, opened 2023-06-03
refactor: use `User_message.t` instead of `exn` for `Resolve.error`

anmonteiro:


src/dune_rules/resolve.ml
```diff
@@ -19,34 +19,58 @@ include Monad.Make (struct
...
+    match pps with
+    | [] -> None
+    | _ ->
+      Some
+        (Pp.vbox
+           (Pp.concat ~sep:Pp.cut
+              (List.map pps ~f:(fun pp ->
+                   Pp.box ~indent:3
+                     (Pp.seq (Pp.verbatim "-> ")
+                        (Pp.seq (Pp.text "required by ") pp))))))
```
rgrinberg, line 55: How come this is needed? Report_error should be able to handle this for us.

src/dune_rules/findlib/findlib.ml
```diff
@@ -703,7 +702,10 @@ end = struct
             in
             match exists with
             | Error e ->
-              Memo.return (Error (Unavailable_reason.Invalid_dune_package e))
+              Memo.return
+                (Error
+                   (Unavailable_reason.Invalid_dune_package
+                      (User_message.make [ Exn.pp e ])))
```
rgrinberg, line 708: Probably better to just make `Or_meta.load_file` return `User_message.t` too.

src/dune_rules/dune_package.ml
```diff
@@ -458,12 +458,17 @@ module Or_meta = struct
...
+           match
+             Vfile.parse_contents lexbuf ~f:(fun lang ->
+                 String_with_vars.set_decoding_env
+                   (Pform.Env.initial lang.version)
+                   (decode ~lang ~dir))
+           with
+           | x -> Ok x
+           | exception User_error.E message -> Error message
+           | exception exn ->
+             (* XXX stop catching code errors, invalid args, etc. *)
```
rgrinberg, line 470: You could take the opportunity to stop catching this stuff error. Off the top of my head, you only need to catch Sys_error, and End_of_file errors here.

rgrinberg (approved):

src/dune_rules/dune_package.ml
anmonteiro: Will do a more thorough analysis and fix in a follow-up PR.

PR #7878, opened 2023-06-05
feat: allow multiple globs in library's (stdlib (internal_modules ...))

anmonteiro:


anmonteiro:
Does this need to be versioned behind a new `experimental_building_ocaml_compiler_with_dune` 0.2?

src/dune_rules/ocaml_stdlib.ml
```diff
@@ -3,7 +3,7 @@ open! Import
 type t =
   { modules_before_stdlib : Module_name.Set.t
   ; exit_module : Module_name.t option
-  ; internal_modules : Glob.t
+  ; internal_modules : Glob.t list
```
rgrinberg, line 6: We usually favor the `Predicate_lang.Glob.t` for things like this.

anmonteiro: Thanks for the suggestion. Pushed a new version that:

- uses `Predicate_lang.Glob.t`
- versions the `experimental_building_ocaml_compiler_with_dune` syntax to 0.2

src/dune_rules/ocaml_stdlib.ml
```diff
@@ -12,20 +12,42 @@ let syntax =
...
+      and field =
+        field (repeat string >>| Predicate_lang.Glob.of_string_list)
+      in
+      (old_field, field)
+    in
+    let expected_version = (0, 2) in
+    if Syntax.Version.Infix.(version >= expected_version) then field
+    else
+      try_ old_field (fun _exn ->
+          field >>| fun _internal_modules ->
```
rgrinberg, line 37: `let+ _internal_modules = field`

src/dune_rules/ocaml_stdlib.ml
```diff
@@ -12,20 +12,42 @@ let syntax =
...
+      let old_field = field (Glob.decode >>| Predicate_lang.Glob.of_glob)
+      and field =
+        field (repeat string >>| Predicate_lang.Glob.of_string_list)
+      in
+      (old_field, field)
+    in
+    let expected_version = (0, 2) in
+    if Syntax.Version.Infix.(version >= expected_version) then field
+    else
+      try_ old_field (fun _exn ->
```
rgrinberg, line 36: If the `internal_modules` parse fails, you should re-raise the old error (`exn`).

By the way, this feature is experimental so this versioning isn't really needed. The fields are backward compatible enough that we can just tell people to move on.

rgrinberg (commented):
Also needs a CHANGES entry.

src/dune_rules/ocaml_stdlib.ml
```diff
@@ -12,20 +12,42 @@ let syntax =
...
+module Internal_modules = struct
+  let field ~loc:stanza_loc ~version =
+    let open Dune_lang.Decoder in
+    let old_field, field =
+      let field parser =
+        field "internal_modules" parser ~default:Predicate_lang.false_
+      in
+      let old_field = field (Glob.decode >>| Predicate_lang.Glob.of_glob)
+      and field =
+        field (repeat string >>| Predicate_lang.Glob.of_string_list)
```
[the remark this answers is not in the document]
anmonteiro: I’d like to pass multiple globs as `internal_modules abc* def*`

does that work with your suggestion?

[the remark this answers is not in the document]
rgrinberg: It should work. Either implicitly as `abc* def*` or explicitly as `(or abc* defe*)`

src/dune_rules/modules.ml
```diff
@@ -87,7 +87,10 @@ module Stdlib = struct
      compilation unit name of such modules, so they cannot be wrapped. *)
   let special_compiler_module (stdlib : Ocaml_stdlib.t) m =
     let name = Module.name m in
-    Glob.test stdlib.internal_modules (Module_name.to_string name)
+    let name_str = Module_name.to_string name in
+
```
Alizter, line 91:
```suggestion
```

anmonteiro:
OK the diff looks a lot simpler after the latest push.

rgrinberg (approved):

anmonteiro:
@rgrinberg @emillon do you think this could go in 3.8.2? Otherwise we'll need to wait for 3.9 for the next Melange release.

emillon:
Yes, we can aim for a way quicker 3.9.0.

PR #7886, opened 2023-06-05
refactor: remove dead code in OSL decoder

anmonteiro:
We're already in the `else` branch, we never return `` `Modes modes ``


src/dune_rules/dune_file.ml
```diff
@@ -539,20 +539,11 @@ module Library = struct
...
-                if dune_version >= expected_version then `Modes modes
-                else `Upgrade )
-              (fun _ -> raise exn))
-        >>| function
-        | `Modes modes -> modes
-        | `Upgrade ->
-          Syntax.Error.since stanza_loc Stanza.syntax expected_version
-            ~what:"Ordered set language for modes"
+           a supported version *)
+        try_ Mode_conf.Lib.Set.decode (fun _exn ->
```
jchavarri, line 543: Can the `Mode_conf.Lib.Set.decode` expression ever raise?

src/dune_rules/dune_file.ml
```diff
@@ -539,20 +539,11 @@ module Library = struct
...
+            Mode_conf.Lib.Set.decode_osl ~stanza_loc project >>| fun _modes ->
```
jchavarri, line 544: Maybe I am misreading the code, but now the order has changed: first we will try the new parser, and if it fails, nothing will catch it, so the old parser will never run.

jchavarri (commented):
I would have left the code mostly as is, except the block inside `decode_osl`, .e.g:

```diff
        try_
          (Mode_conf.Lib.Set.decode >>| fun modes -> `Modes modes)
          (fun exn ->
            try_
              ( Mode_conf.Lib.Set.decode_osl ~stanza_loc project >>| fun modes ->
-               if dune_version >= expected_version then `Modes modes
-               else `Upgrade )
+              `Upgrade)
              (fun _ -> raise exn))
        >>| function
        | `Modes modes -> modes
        | `Upgrade ->
          Syntax.Error.since stanza_loc Stanza.syntax expected_version
            ~what:"Ordered set language for modes"

File: src/dune_rules/dune_file.ml (hunk 2 above)
Comment (anmonteiro) [reply src/dune_rules/dune_file.ml:544]: I’m pretty sure we have a test for the error behavior too

File: src/dune_rules/dune_file.ml (hunk 1 above)
Comment (anmonteiro) [reply src/dune_rules/dune_file.ml:543]: It can raise if there’s an ordered set language in the field

File: src/dune_rules/dune_file.ml (hunk 2 above)
Comment (anmonteiro) [reply src/dune_rules/dune_file.ml:544]: New parser is OSL which runs in the catch callback

Comment (anmonteiro) [reply src/dune_rules/dune_file.ml:544]: Note that we’re already reading the rose branch of the version check

File: src/dune_rules/dune_file.ml (hunk 1 above)
Comment (jchavarri) [reply src/dune_rules/dune_file.ml:543]: But as the param passed to `try`, `Mode_conf.Lib.Set.decode` is not being applied to anything. How can the evaluated expression raise if it's not applied to anything? I read this code the same as:

```ml
let raise _s = failwith "foo"

let _ = try raise with | _exn -> print_endline
```

There's no point on wrapping with `try` in the example above.

anmonteiro: Because it’s inside the parser monad

jchavarri: Thanks missed that 🤦

jchavarri (approved):
sorry for the distraction

rgrinberg:
Do we have a test that covers this?

anmonteiro:
Yeah, check the tests in https://github.com/ocaml/dune/pull/6611 which also tests the error behavior.

rgrinberg (approved):

PR #7896, opened 2023-06-05
test(x-compilation): resolution failure for transitive `ppx_runtime_libraries`

anmonteiro:
- Adds a test for #7819 that will fail if #7887 is reverted

PR #7887, opened 2023-06-05
Revert #7415 and #7450

emillon:
- Revert "fix(x-compilation): delay evaluation of `ppx_runtime_deps` until context is known"
- Revert "test(x-compilation): demonstrate overlap check failure with ppx_runtime_libraries"
- Revert "fix(x-compilation): find host ppx dependencies in the host context (#7415)"

Fixes #7875

PR #7897, opened 2023-06-05
x-compilation: handle transitive ppx_runtime_libraries

anmonteiro:
1. reverts #7887 
2. showcases the test added in #7896 failing after the revert


CHANGES.md
```diff
@@ -26,9 +26,6 @@ Unreleased
 - The `interface` and `implementation` fields of a `(dialect)` are now optional
   (#7757, @gpetiot)
 
-- Revert #7415 and #7450 (Resolve `ppx_runtime_libraries` in the target context when
```
rgrinberg, line 29: Since we already released this with the revert, this PR should add a new changes entry rather this entry.

anmonteiro:
I haven't been able to find a fix for this issue yet. Here's my current investigation:

- in our test, `ppx_deriving_runtime` (public name `other-ppx.runtime`) appears in `ppx_runtime_libraries` for `other-ppx.api`, a **regular** library
- the ppx `other-ppx.eq` is a ppx deriver, which then depends on `other-ppx.api`
- `other-ppx.api` depends on `(ppx_runtime_libraries ppx_deriving_runtime)`
  - now, this ppx runtime library is resolved by the target context. but its transitive closure never reaches `other-ppx.api` since it was only reachable when we resolved the ppx from the host context

PR #6134, opened 2022-09-07
Check that a test stanza is disabled before generating the rules

voodoos:
Unlike the `executables` stanza, Dune attempted to create some rules for a `tests` even if it was disabled. This adds a similar check as the one for `executables` to abort rule generation if the stanza is disabled.

This should fix #6132

PR #5529, opened 2022-03-24
(test): skip rule creation if enabled_if is false

emillon:
When a `(test)` stanza is disabled through `(enabled_if)`, do not create any rule. This removes spurious "Library X is hidden (unsatisfied `enabled_if`)" error messages.

Issue #5505, opened 2022-03-06
enabled_if not masking a library target


avsm:
eio has the following dune file frag:

```
(test
 (name test)
 (package eio_linux)
 (enabled_if (= %{system} "linux"))
 (modules test)
 (libraries alcotest eio_linux))
```

doing a build on macos results in this error:

```
File "lib_eio_linux/tests/dune", line 30, characters 21-30:
30 |  (libraries alcotest eio_linux))
                          ^^^^^^^^^
Error: Library "eio_linux" in _build/default/lib_eio_linux is hidden
(unsatisfied 'enabled_if').
-> required by _build/default/lib_eio_linux/tests/test.bc
-> required by alias lib_eio_linux/tests/all
-> required by alias default
```

cc @talex5 @patricoferris

It would be good if the `exists_if` in the dune file stops the library target from being attempted to be resolved.

PR #7899, opened 2023-06-06
add: build_if in test stanza

### LINKED ISSUE #6938: enabled_if doesn't work in (test)
## Expected Behavior

We have:
```
(test
 (name test)
 (package eio_linux)
 (enabled_if (= %{system} "linux"))
 (modules test)
 (libraries alcotest eio_linux))
```
I expect this stanza to be ignored on macos. However, dune tries to build `test.ml` anyway. If I delete the stanza, it skips the file, as expected (reported by a macos user and tested on Linux by changing "linux" to "foobar" and confirming that it still tries to build it).

I consulted the dune documention for `(test)`:

- `(test)` says "the test stanza is the singular form of tests"
- `(tests)` says "The optional fields supported are a subset of the alias and executables fields. In particular, all fields except for public_names are supported from the [executables stanza](https://dune.readthedocs.io/en/latest/dune-files.html#shared-exe-fields). Alias fields apart from name are allowed."
- `(executables)` says "`(enabled_if <blang expression>)` is the same as the corresponding field of [library](https://dune.readthedocs.io/en/latest/dune-files.html#library)."
- `(library)` says "The condition is specified using the [Boolean Language](https://dune.readthedocs.io/en/latest/concepts.html#blang), and the field allows for the %{os_type} variable, which is expanded to the type of OS being targeted by the current build."

However, I believe `system` is supposed to work. If I try a different variable name, `dune` says `Error: Unknown variable %{baz}`, which I infer to mean that it does understand `system`.

/cc @polytypic

## Reproduction

```
$ ls
dune  dune-project  test.ml

$ cat dune
(test
 (name test)
 (enabled_if (= %{system} "foobar"))
 (modules test))

$ cat dune-project 
(lang dune 3.0)

$ cat test.ml 
BROKEN

$ dune build @all
File "test.ml", line 1, characters 0-6:
1 | BROKEN
    ^^^^^^
Error: Unbound constructor BROKEN
```

## Specifications

- Version of `ocaml` (output of `ocamlc --version`): 5.0.0
- Operating system (distribution and version): Debian 11.6

emillon:
Fixes #6938

The semantics of `(enabled_if)` in `(test)` can be confusing: `(test)` can be seen as the combination of `(executable)` and a `(rule (alias runtest))`; but `(enabled_if)` actually only controls the "running" part, not the "building" one.

This adds a new `(build_if)` field in `(test)`. When it evaluates to false, the test stanza is bypassed (no build is attempted).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -69,6 +69,9 @@ Unreleased
 - Compute digests and manage sandboxes in background threads (#7947,
   @rgrinberg)
 
+- Add `(build_if)` to the `(test)` stanza. When it evaluates to false, the
+  executable is not built. (#7899, fixes #6938, @emillon)
+
 3.8.1 (2023-06-05)
 ------------------
 
--- a/doc/stanzas/test.rst
+++ b/doc/stanzas/test.rst
@@ -37,6 +37,10 @@ In particular, all fields except for ``public_names`` are supported from the
 :ref:`executables stanza <shared-exe-fields>`. Alias fields apart from ``name``
 are allowed.
 
+The ``(enabled_if)`` field has special semantics: when present, it only applies
+to running the tests. The test executable is always built by default.
+If you need to restrict building the test executable, use ``(build_if)`` instead.
+
 By default, the test binaries are run without options.  The ``action`` field can
 override the test binary invocation, i.e., if you're using Alcotest and wish to
 see all the test failures on the standard output. When running Dune ``runtest``
--- a/src/dune_rules/dune_file.ml
+++ b/src/dune_rules/dune_file.ml
@@ -1892,6 +1892,7 @@ module Tests = struct
     ; package : Package.t option
     ; deps : Dep_conf.t Bindings.t
     ; enabled_if : Blang.t
+    ; build_if : Blang.t
     ; action : Dune_lang.Action.t option
     }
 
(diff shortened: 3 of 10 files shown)


emillon:
Remarks:
- I'm not sure about the syntax; I also contemplated extending the `enabled_if` field with something like `(enabled_if (build (...)))`
- this does not allow having a different predicate for "build if" and "test if" but I don't think we need that nuance.

emillon:
xrefs: #5529 #6134 #5505

emillon:
> How about `build_if`? It's a little more in-line with our other names. Also, it gives you strictly more flexibility by allowing you to control both conditions.

do you mean that build_if would be an extra blang field? I can try that. I think there should be a shorthand for when the condition is the same though. what do you think about `(build_if X)` and no `(enabled_if)` be a shortcut for `(build_if X) (enabled_if X)`?

emillon:
Actually, it's not even necessary: `(enabled_if)` defaults to true so replacing `(enabled_if)` by `(build_if)` does the expected thing.

rgrinberg (approved):

Alizter:
Will you update the docs?

emillon:
> Will you update the docs?

Yes, it's literally in the TODO list above :)

emillon:
This works on the ctypes port; I also created an eio branch that uses that and seems to be working.

tmattio:
This is super useful, thanks @emillon! I have the same question as @Alizter: why not change the semantics of `enabled_if` instead? It seems to me that if you don't want to run your test, it's not worth building it?

rgrinberg:
There's some situations where a test is only executable on a particular platform but it's still useful to make sure it at least builds on every platform for all developers. We have some tests like that in dune.

emillon:
Yes, exactly. For example, something that only runs if another package is available, or if you have an API key available. I would say it depends on what is in the `enabled_if` clause.

tmattio:
That makes sense, thanks!

PR #7906, opened 2023-06-06
Use threaded console, adapt for Windows

nojb:
Fixes the issue discussed in https://github.com/ocaml/dune/pull/7418#issuecomment-1577430087.

### DIFF
--- a/src/dune_threaded_console/dune_threaded_console.ml
+++ b/src/dune_threaded_console/dune_threaded_console.ml
@@ -61,7 +61,7 @@ let make (module Base : S) : (module Dune_console.Backend) =
     let start () =
       Base.start ();
       Dune_engine.Scheduler.spawn_thread @@ fun () ->
-      ignore (Unix.sigprocmask SIG_UNBLOCK [ Signal.to_int Winch ] : int list);
+      if Sys.unix then ignore (Unix.sigprocmask SIG_UNBLOCK [ Signal.to_int Winch ] : int list);
       let last = ref (Unix.gettimeofday ()) in
       let frame_rate = 1. /. 60. in
       let cleanup () =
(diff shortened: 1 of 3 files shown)


rgrinberg (approved):
LGTM. Please remember to run ocamlformat.

Note that you should also mention that we're switching to the threaded console again for everyone.

Finally, I still don't understand why the unthreaded console is broken. It works just fine outside of Windows.

src/dune_threaded_console/dune_threaded_console.ml
```diff
@@ -61,7 +61,7 @@ let make (module Base : S) : (module Dune_console.Backend) =
     let start () =
       Base.start ();
       Dune_engine.Scheduler.spawn_thread @@ fun () ->
-      ignore (Unix.sigprocmask SIG_UNBLOCK [ Signal.to_int Winch ] : int list);
+      if Sys.unix then ignore (Unix.sigprocmask SIG_UNBLOCK [ Signal.to_int Winch ] : int list);
```
rgrinberg, line 64: So cygwin doesn't emulate signals on windows?

nojb: I don't know, I never use Cygwin executables, but I can change the condition to `not Sys.win32` if that sounds better.

nojb: I did the change.

nojb:
> LGTM. Please remember to run ocamlformat.
> 
> Note that you should also mention that we're switching to the threaded console again for everyone.

Done.

> Finally, I still don't understand why the unthreaded console is broken. It works just fine outside of Windows.

Indeed; needs to be looked into.

nojb:
> Please remember to run ocamlformat.

Thanks for the reminder, done.

emillon:
3.8.1 has been released, I'm removing it from the milestone.

nojb:
> 3.8.1 has been released, I'm removing it from the milestone.

Argh, I missed that. Is a 3.8.2 planned?

Alizter:
@nojb I added it to a list of backports #7907

emillon:
Yes, we'll have to do one since there are more coq fixes.
But in any case please note that setting the milestone is only for the PR that get into the 3.8 branch (in other words, the backport), not the one that targets `main`.

PR #7911, opened 2023-06-07
doc: .opam.template files can be generated

emillon:
While trying to see if it's supported or not, I did a few tests with `(copy
a.opam.template b.opam.template)` and it worked.


rgrinberg:
Yes, this is supported.

emillon:
Great, thanks.

PR #7914, opened 2023-06-07
Lockdir encode/decode roundtrip tests

gridbugs:
This adds some tests that a `Lock_dir.t` can be written do disk and then read back from disk to obtain a `Lock_dir.t` that matches the original.

### DIFF
--- a/src/dune_lang/action.ml
+++ b/src/dune_lang/action.ml
@@ -60,38 +60,68 @@ module File_perm = struct
 end
 
 module Env_update = struct
-  type op =
-    | Eq
-    | PlusEq
-    | EqPlus
-    | ColonEq
-    | EqColon
-    | EqPlusEq
+  module Op = struct
+    type t =
+      | Eq
+      | PlusEq
+      | EqPlus
+      | ColonEq
+      | EqColon
+      | EqPlusEq
+
+    let equal a b =
+      match (a, b) with
+      | Eq, Eq
+      | PlusEq, PlusEq
+      | EqPlus, EqPlus
+      | ColonEq, ColonEq
+      | EqColon, EqColon
+      | EqPlusEq, EqPlusEq -> true
+      | _ -> false
+
+    let all =
+      [ ("=", Eq)
+      ; ("+=", PlusEq)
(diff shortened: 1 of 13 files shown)


src/dune_lang/action.ml
```diff
@@ -60,7 +60,35 @@ module File_perm = struct
...
+      | EqPlusEq
+
+    let equal a b =
+      match (a, b) with
+      | Eq, Eq -> true
+      | PlusEq, PlusEq -> true
+      | EqPlus, EqPlus -> true
+      | ColonEq, ColonEq -> true
+      | EqColon, EqColon -> true
+      | EqPlusEq, EqPlusEq -> true
```
Alizter, line 79: You can merge these all into a single ->

src/dune_pkg/lock_dir.ml
```diff
@@ -9,6 +9,27 @@ module Source = struct
         ; checksum : (Loc.t * Checksum.t) option
         }
 
+  let equal a b =
```
Alizter, line 12: Maybe this should be called equal_no_locs or something since we are explicitly ignoring the loc. Also what is the loc from/used for?

rgrinberg: You could also just add a `drop_loc` function to convert all the locations into `Loc.none`. This is the convention we use elsewhere.

gridbugs: @Alizter the loc is used in error messages. Will update to have a `drop_loc` function.

src/dune_lang/action.ml
```diff
@@ -60,38 +60,71 @@ module File_perm = struct
...
-    | EqColon
-    | EqPlusEq
+  module Op = struct
+    type t =
+      | Eq
+      | PlusEq
+      | EqPlus
+      | ColonEq
+      | EqColon
+      | EqPlusEq
```
Alizter, line 70: Could you also add comments on their meaning?

rgrinberg:
Thanks. I made all `equal` function respect locations and added `remove_locs` to the callers. Otherwise, it's a little too easy to use the wrong equality.

PR #7874, opened 2023-06-04
nix: allow entering slim shell with packages compiled from local dune

anmonteiro:
- Removes our usage of opam-nix
- Preserves a package `nix build .#dune` or `nix develop .#scope` which uses the dune package from the repository to build the OCaml nixpkgs scope

PR #7915, opened 2023-06-07
deps: use melange from opam

emillon:


Makefile
```diff
@@ -71,24 +71,14 @@ install-ocamlformat:
...
-dev-deps: melange
+dev-deps:
 	opam install -y $(TEST_DEPS)
 
 .PHONY: coverage-deps
 coverage-deps:
 	opam install -y bisect_ppx
 
-.PHONY: dev-deps-sans-melange
-dev-deps-sans-melange:
```
jchavarri, line 89: This command is used here

https://github.com/ocaml/dune/blob/bd1193de7b5b59770d01d0a3f43e6936c6271a16/ci/build-test.sh#L120

Same with `test-all-sans-melange` below.

jchavarri (approved):
Mentioned in the inline comment: I think there is some cleaning that could be done in `build-test.sh`? Not sure why the tests are not failing, maybe it's not used.

There is also some pinning happening in nix flakes, it would be nice to keep that workflow in sync with the opam one, but I have no clue how to fix those.

Makefile
emillon: I updated this by following the comments. But it's not used since we only use that file on windows.

emillon:
@anmonteiro I tried to update the config so that we pick melange >= 1.0.0 from opam instead of pinning. Is there anything to do in flake.nix?

Alizter:
@emillon https://github.com/ocaml/dune/pull/7874 takes care of that I think.

anmonteiro (approved):
LGTM, the nix changes can be done separately

PR #7919, opened 2023-06-08
split describe into cmdliner subcommands

Alizter:
As we discussed, this PR splits `dune describe` into subcommands in preparation for introducing the `dune show` command.

I've also taken this opportunity to do some misc refactoring in between, hopefully whilst preserving the functionality.

Can be squashed later, I've kept the commits separate in case I have messed anything up and I need to bisect.

- [x] changelog (it is somewhat user facing)
- [ ] update docs?
- [x] check `--help` has not broken and makes sense.


bin/describe/describe_common.mli
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  module Workspace : sig
+    type t = Item.t list
+
+    val to_dyn : options -> t -> Dyn.t
+  end
+end
+
+module Crawl : sig
```
rgrinberg, line 22: I don't understand this refactoring very well. The name is `Describe_common` and yet basically everything it contains is specific to a particular subcommand. What's the motivation for this module?

Alizter: Yeah, I didn't bother untangling it further yesterday. I'll try to do that today. There are parts used by multiple commands which I will try to leave here whilst moving other things elsewhere.

test/blackbox-tests/test-cases/describe/describe.t
```diff
@@ -1316,33 +1316,34 @@ Test other formats
 Test errors
 -----------
 
-  $ dune describe --lang 0.1 workspac
-  Error: Unknown constructor workspac
-  Hint: did you mean workspace?
+  $ dune describe workspacw --lang 0.1
```
Alizter, line 1319: This test would actually succeed since `cmdliner` can infer that `workspac` -> `workspace`. Therefore I changed it to actually be incorrect.

Alizter:
Actually, I've just noticed that I can really get rid of `describe_common`. I'll do that now.;

Alizter:
OK I've gotten rid of describe_common. There is a describe_format library left which is used by multiple commands. Since this is only about writing dyn as a csexp or sexp it might be worth including that elsewhere.

test/blackbox-tests/test-cases/describe/describe.t
```diff
@@ -1316,33 +1316,34 @@ Test other formats
...
-  Try 'dune describe --help' or 'dune --help' for more information.
+  Usage: dune describe workspace [OPTION]… [DIRS]…
+  Try 'dune describe workspace --help' or 'dune --help' for more information.
   [1]
 
 opam file listing
 -----------------
 
-  $ dune describe --lang 0.1 opam-files | dune_cmd expand_lines
+  $ dune describe opam-files | dune_cmd expand_lines
```
rgrinberg, line 1346: Why was the argument removed?

rgrinberg: We should probably just accept it everywhere we used to accept it. Otherwise it's a breaking change and I'd rather not think about break anyone's workflow for now.

Alizter: I can add a dummy option back in to `dune describe opam-files` but it cannot be in the same position. Unfortunately, command groups do not take options.

rgrinberg: Can't it be added individually to every subcommand?

Alizter: Yes, I have done that for `dune describe opam-files`. I meant that it isn't possible to keep the same position in the tests:
```
dune describe --lang 0.1 opam-files
```
Do you want me to add the lang arg to the `dune describe pp` and `dune describe external-lib-deps` too?

rgrinberg: I think that would be the safest probably. So let's do that.

bin/describe/describe_external_lib_deps.ml
```diff
@@ -0,0 +1,194 @@
...
+  let+ common = Common.term
+  and+ context_name = Common.context_arg ~doc:"Build context to use."
+  and+ _ =
+    Arg.(
+      value
+      & opt (some string) None
+      & info [ "lang" ] ~docv:"VERSION"
+          ~doc:
+            "This argument has no effect and is deprecated. It exists solely \
+             for backwards compatibility.")
```
rgrinberg, line 170: Might as well move this argument to `Describe_lang`. I see it duplicated in a few places.

Alizter: OK, I've moved it to `Describe_lang_compat`.

Issue #7907, opened 2023-06-06
backports for 3.8.2


Alizter:
- #7906 
- #7895

rgrinberg:
Let's add this one as well https://github.com/ocaml/dune/pull/7418

Alizter:
This is a regression introduced in 3.8 which highlights two issues:
- We have an issue with cycle detection (prob not important)
- There is a cycle with install rules that needs to be broken. roughly
```
coq.install -> theory -> coqc for config -> coq.install
```

emillon:
OK, thanks for the explanation. Let's put it this way: 3.8.2 is not blocked on this. A fix for this can be included in 3.8.2 but that will depend on the fix.

ejgallego:
#7908 was introduced in 3.7.0 .

Indeed none of the open Coq bugs are blockers for 3.8.2, tho the above bug is kind of serious I'm afraid, but need to investigate more.

PR #7921, opened 2023-06-08
coq: improve coq_sources overlap error 

Alizter:
Improve the error message when we have overlapping coq stanzas.

I'm waiting for 3.8.2 #7907 so that backporting is easier.

### DIFF
--- a/src/dune_rules/coq/coq_sources.ml
+++ b/src/dune_rules/coq/coq_sources.ml
@@ -58,6 +58,25 @@ let check_no_unqualified (loc, (qualif_mode : Dune_file.Include_subdirs.t)) =
 let extract t (stanza : Extraction.t) =
   Loc.Map.find_exn t.extract stanza.buildable.loc
 
+let add_module_to_rev_map ~stanza acc m =
+  Coq_module.Map.add acc m stanza |> function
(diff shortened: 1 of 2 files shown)


emillon:
> I'm waiting for 3.8.2 #7907 so that backporting is easier.

What do you mean?

Alizter:
@emillon This might contain code that will conflict with the patches I want to include in 3.8.2. I would rather that those patches get merged into main first so that they can be backported easier. If those patches get rebased on this it will be annoying for you.

emillon:
I don't understand? Nothing gets backported before it is in main. So this might conflict with another patch you want to include in main, but this should have nothing to do with 3.8.2?

emillon:
OK I think I get what you meant, sorry. This contains changes that can't be backported, so you want to include the backportable fixes first. That makes sense. But you don't have to wait for 3.8.2 to be released; just know that once this PR lands, you won't be able to ship related fixes to 3.8.x.

Alizter:
@emillon Sure there is no rush for this anyway, I just had it laying around.

Alizter:
Actually I changed my mind with this. If we care we can resurrect.

PR #7923, opened 2023-06-08
refactor(rpc): use Console.print

Alizter:


bin/rpc.ml
```diff
@@ -91,8 +91,9 @@ module Status = struct
     let+ (common : Common.t) = Common.term in
     client_term common @@ fun _common ->
     let where = active_server () in
-    printfn "Server is listening on %s" (Dune_rpc.Where.to_string where);
-    printfn "Connected clients (including this one):\n";
+    Dune_console.printf "Server is listening on %s"
```
rgrinberg, line 94: Just use `Console.print`. You can also print both messages in a single call.

Alizter: Good point.

bin/rpc.ml
```diff
@@ -91,8 +91,11 @@ module Status = struct
     let+ (common : Common.t) = Common.term in
     client_term common @@ fun _common ->
     let where = active_server () in
-    printfn "Server is listening on %s" (Dune_rpc.Where.to_string where);
-    printfn "Connected clients (including this one):\n";
+    Console.print
+      [ Pp.textf "Server is listening on %s" (Dune_rpc.Where.to_string where)
+      ; Pp.textf "Connected clients (including this one):"
+      ; Pp.newline
```
rgrinberg, line 97: Is this newline necessary? I thought it always wrote one at the end.

PR #7924, opened 2023-06-08
doc/melange.rst - make the alias attachment behavior of `melange.emit` clearer

haochenx:
- rendered: https://dune--7924.org.readthedocs.build/en/7924/melange.html
- related: https://github.com/ocaml/dune/pull/7926

Before #7926, targets declared by `melange.emit` stanza is not attached to the dune's default `@all` stanza, which can be confusing. #7926 changes that and this PR aims to make the document clearer.

<details>
<summary>original (and out-of-date) pr description</summary>
Following the introduction of the current melange document page I was confused by why running `dune build` would not produce the javascript outputs and only realize that the `melange.emit` stanza by default attach the targets to the `melange` alias instead of the default alias after a very careful examination of the documentation.

It would be nice to make the default alias attachment clear in the introduction section to avoid potential confusions of new adopters.

## (initial) diff:
```diff
-After running ``dune build output/hello.js``, Dune produces the following
-file structure:
+After running ``dune build output/hello.js`` or ``dune build @melange``
+(by default, all ``melange.emit`` stanzas will attach their targets to
+the  ``melange`` alias),
+Dune produces the following file structure:
```

that is, the following text being inserted:
"**or ``dune build @melange`` (by default, all ``melange.emit`` stanzas will attach their targets to the  ``melange`` alias)**"
</details>


haochenx:
Sorry that my previous force push (to fix a layout problem) wasn't containing the intended fix and I just force pushed again.

doc/melange.rst
```diff
@@ -57,8 +57,10 @@ Finally, add a source file to build:
 
   echo 'Js.log "hello from melange"' > hello.ml
 
-After running ``dune build output/hello.js``, Dune produces the following
-file structure:
+After running ``dune build output/hello.js`` or ``dune build @melange``
+(by default, all ``melange.emit`` stanzas will attach their targets to
+the  ``melange`` alias),
```
jchavarri, line 62: This behavior is explained in the `alias` field of the `melange.emit` stanza docs, a few paragraphs below. I am not sure it's worth adding a reference?

jchavarri (approved):
Looks good. I wonder if it's better to remove the explicit build of `output/hello.js` altogether, as it breaks as soon as there are two modules, and it's not the recommended way to build Melange projects in any case. @anmonteiro wdyt?

doc/melange.rst
haochenx: When I was following the introduction section and got `hello.js` built, I started to hacking the minimal example around and was met with surprises when `dune build` (without specifying the `melange` alias) wasn't generating the javascript files.

I personally wasted some decent time trying to debug it was the case and only realized such behavior after very carefully reading the document again.

Yes the behavior is already documented, but I think it's probably not be only me getting the impression that the targets added by `melange.emit` stanza will be attached to the `default` alias (as for most other dune stanzas) after just following the current introduction.

haochenx:
>I wonder if it's better to remove the explicit build of output/hello.js altogether

I have no preference. @anmonteiro should have a better say.

doc/melange.rst
jchavarri: I was not aware that all other targets (including jsoo ones, it seems) are added to the default alias. I am onboard with adding `melange.emit` targets to the default alias, but I'm not sure if this addition was avoided for some reason that I'm not aware of. @rgrinberg @anmonteiro do you know?

(Probably something that should be discussed in a separate issue)

rgrinberg: I think adding it to the default alias (`@all` in this case) is the correct thing to do.

anmonteiro:
@haochenx thanks for starting the discussion on this. I agree the language is confusing, but I'm concerned we might have added more confusion in this PR by explaining too much.

Given the other comment, I wonder whether this would be an appropriate course of action:

1. Open a new PR adding `melange.emit` to the default alias
2. In this PR, mention "after `dune build`, we get the following project structure" (and we can explain aliases later on)
3. in the aliases section, add a note about `melange.emit` JS rules being attached to the default alias `@all` since Dune 3.0


@haochenx would you like to tackle 1. ? I believe it'd be a matter of adding more rules to the implicit default alias here:

https://github.com/ocaml/dune/blob/0591719321fd4ce57b126f1afe652af61ed0e967/src/dune_rules/melange/melange_rules.ml#L326-L335


You can get the default alias with: `(Build_config.get ()).implicit_default_alias dir`

Alizter:
If you want to attach melange.emit to default, we can do it in time for 3.8.2.

rgrinberg:
> You can get the default alias with: (Build_config.get ()).implicit_default_alias dir

You could, but that would be a mistake. The default alias can be customized by the user so we don't want to attach to it directly and prevent the user from changing it. By "default", it's set to `@all`. So we should attach it to `@all`.

> If you want to attach melange.emit to default, we can do it in time for 3.8.2. Just open a PR or issue and add it to https://github.com/ocaml/dune/issues/7907.

We don't add features to point releases.

anmonteiro:
> By "default", it's set to `@all`. So we should attach it to `@all`.

Oh right, we should be using `Alias.Name.all`.

haochenx:
I am also in favor of adding the targets to `@all` (the default alias) by default and it seems to be the consensus here. 

I’m happy to send a PR making such a change and update this PR to reflect the point [(2) and (3) by @anmonteiro](https://github.com/ocaml/dune/pull/7924#issuecomment-1583215590). 

My concern is that this will be a breaking change to an already shipped feature. I’m not sure whether silently changing the behavior is a good thing to do. Maybe we would like to bump the plugin version? (i.e. only change the default alias if the user specify `(using melange 0.2)` instead of `(using melange 0.1)` in dune-project)

But if we agree that given the experimental status of the current melange support it’s acceptable to have such “silent” breaking changes. It’s definitely easier to implement and maintain this way. 

Also, regarding the structure of the PRs, maybe it would make sense to close this PR and address the document change in the PR that change the default alias attachment. I think this will be clearer retrospectively.

anmonteiro:
> But if we agree that given the experimental status of the current melange support it’s acceptable to have such “silent” breaking changes. It’s definitely easier to implement and maintain this way.

My opinion is that we don't care too much in this phase and we can ship this change. If others feel differently, I'm happy to defer to stronger opinions.


> Also, regarding the structure of the PRs, maybe it would make sense to close this PR and address the document change in the PR that change the default alias attachment. I think this will be clearer retrospectively.


Happy to keep the alias change and the documentation in a separate PR>

christinerose (approved):

haochenx:
Now #7296 is merged and I have updated the patch to document accordingly.
(following (2) and (3) of this suggestion: https://github.com/ocaml/dune/pull/7924#issuecomment-1583215590)

PTAL @anmonteiro @jchavarri @rgrinberg @christinerose

haochenx:
It seems that the windows CI job is stalling:

https://github.com/ocaml/dune/actions/runs/5241726516/jobs/9464209550?pr=7924

It might be a good idea to restart it (I doesn't seem to have the permission to do so)

jchavarri (approved):
Thank you! Just one suggestion, see previous comment.

jchavarri (approved):

doc/melange.rst
```diff
@@ -148,6 +148,10 @@ The resulting layout in ``_build/default/output`` will be as follows:
     is explicitly defined in the stanza, the targets from this stanza will
     be excluded from the ``melange`` alias.
 
+  - As per Dune's convention, all targets declared using this stanza will also
+    be attached to the Dune's default ``@all`` alias, regardless of whether
+    the ``(alias ...)`` field is present.
```
anmonteiro, line 153:
```suggestion
  - The targets of ``melange.emit`` are also attached to the Dune default
    alias (``@all``), regardless of whether the ``(alias ...)`` field is present.
```

anmonteiro (approved):

haochenx:
@jchavarri @anmonteiro Thank you!

christinerose (approved):

anmonteiro:
Thanks again!

haochenx:
Thank you all!

PR #7932, opened 2023-06-10
fix: ppx and reason bug

### LINKED ISSUE #7930: ppx_expect + reason is broken after dune lang 3.2
## Expected Behavior
```
File "test/Test.re", line 1, characters 0-0:
diff --git a/_build/default/test/Test.re b/_build/default/test/Test.re.corrected
index e3f7c74..d69e228 100644
--- a/_build/default/test/Test.re
+++ b/_build/default/test/Test.re.corrected
@@ -1,4 +1,4 @@
 let%expect_test "test" = {
   print_endline("hello");
-  [%expect{| |}]
+  [%expect{| hello |}]
 }
```

## Actual Behavior
```
File "test/Test.re", line 1:       
Error: I/O error: test/Test.re: No such file or directory
```

## Reproduction

test/dune
```dune
(library
 (name test)
 (inline_tests)
 (preprocess (pps ppx_expect))
 (libraries ))
```

test/test.re
```reason
let%expect_test "test" = {
  print_endline("hello");
  [%expect{| |}]
}
```

dune-project
```
(lang dune 3.3)
```

If change the line `(lang dune 3.3)` to `(lang dune 3.2)`. Everything seems ok.

## Specifications

- Version of `dune` (output of `dune --version`): 3.8.1
- Operating system (distribution and version): macOS 13.4 (22F66)

Alizter:
Fix #7930

ppx needs the original source of a file even if it is a dialect. This was missing causing the sandboxed version to error. We add the original source file as a hidden dependency in the ppx action.

I also found that inline_tests was not using the correct dialect source, which can be seen in my commits as I tried to remove my original reason dependency. I therefore have also submitted a fix for that.

- [x] changelog

### DIFF
--- a/src/dune_rules/preprocessing.ml
+++ b/src/dune_rules/preprocessing.ml
@@ -699,6 +699,10 @@ let pp_one_module sctx ~lib_name ~scope ~preprocessor_deps
                            ; Path (Path.build dst)
                            ; Command.Ml_kind.ppx_driver_flag ml_kind
                            ; Dep (Path.build src)
+                           ; Hidden_deps
+                               (Module.source m ~ml_kind |> Option.value_exn
+                              |> Module.File.path |> Dep.file
+                              |> Dep.Set.singleton)
                            ; As flags
                            ]
                          >>| Action.Full.add_sandbox sandbox))))
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,9 @@ Unreleased
 - Add `dune build --dump-gc-stats FILE` argument to dump garbage collection
   stats to a named file. (#8072, @Alizter)
 
+- Fix bug with ppx and Reason syntax due to missing dependency in sandboxed
+  action (#7932, fixes #7930, @Alizter)
+
 3.9.1 (2023-07-06)
 ------------------
 
--- a/test/blackbox-tests/test-cases/dune
+++ b/test/blackbox-tests/test-cases/dune
@@ -6,7 +6,8 @@
   (binaries
    ../utils/dune_cmd.exe
    ../utils/dunepp.exe
-   ../utils/melc_stdlib_prefix.exe)))
+   ../utils/melc_stdlib_prefix.exe
+   ../utils/refmt.exe)))
 
 (cram
  (applies_to pp-cwd)
@@ -141,5 +142,9 @@
 (cram
  (applies_to reason-expect)
  (deps
-  %{bin:refmt}
-  (package ppx_expect)))
+  (package ppx_expect)
+  %{bin:refmt}))
+
+(cram
+ (applies_to reason)
+ (deps %{bin:refmt}))
--- a/test/blackbox-tests/test-cases/dune
+++ b/test/blackbox-tests/test-cases/dune
@@ -142,5 +142,4 @@
  (applies_to reason-expect)
  (deps
   %{bin:refmt}
-  ;(package ppx_expect)
-  ))
+  (package ppx_expect)))
--- a/test/blackbox-tests/test-cases/dune
+++ b/test/blackbox-tests/test-cases/dune
@@ -137,3 +137,10 @@
  (enabled_if
   (= %{system} linux))
  (deps %{bin:strace}))
+
+(cram
+ (applies_to reason-expect)
+ (deps
+  %{bin:refmt}
+  ;(package ppx_expect)
+  ))
--- a/test/blackbox-tests/test-cases/reason-expect.t/run.t
+++ b/test/blackbox-tests/test-cases/reason-expect.t/run.t
@@ -15,43 +15,14 @@ Testing the interaction of reason syntax and ppx expect
   > EOF
 
 In https://github.com/ocaml/dune/issues/7930 there is an issue where reason
-syntax and ppx_expect break at dune lang 3.3.
-
-When dune lang is 3.2 everything should work as expected:
+syntax and ppx_expect break at dune lang 3.3 due to the sandboxing of ppx.
 
   $ cat > dune-project << EOF
-  > (lang dune 3.2)
+  > (lang dune 3.9)
   > EOF
 
   $ dune test
   File "test.re", line 1, characters 0-0:
   Error: Files _build/default/test.re and _build/default/test.re.corrected
-  differ.
-  [1]
-
-However on dune lang 3.3 this appears to break:
-
-  $ cat > dune-project << EOF
-  > (lang dune 3.3)
-  > EOF
-
-  $ dune test
-  File "test.re", line 1:
-  Error: I/O error: test.re: No such file or directory
-  [1]
-
-This appears to only happen for .re files:
-
-  $ rm test.re
-
-  $ cat > test.ml << EOF
-  > let%expect_test _ =
-  >   print_endline "Hello, world!";
-  >   [%expect {| |}]
-  > EOF
-
-  $ dune test
-  File "test.ml", line 1, characters 0-0:
-  Error: Files _build/default/test.ml and _build/default/test.ml.corrected
   differ.
   [1]
--- a/test/blackbox-tests/test-cases/reason-expect.t/run.t
+++ b/test/blackbox-tests/test-cases/reason-expect.t/run.t
@@ -1,0 +1,57 @@
+Testing the interaction of reason syntax and ppx expect
+
+  $ cat > dune << EOF
+  > (library
+  >  (name test)
+  >  (inline_tests)
+  >  (preprocess (pps ppx_expect)))
+  > EOF
+
+  $ cat > test.re << EOF
+  > let%expect_test _ = {
+  >   print_endline("Hello, world!");
+  >   [%expect {| |}]
+  > }
+  > EOF
+
+In https://github.com/ocaml/dune/issues/7930 there is an issue where reason
+syntax and ppx_expect break at dune lang 3.3.
+
+When dune lang is 3.2 everything should work as expected:
+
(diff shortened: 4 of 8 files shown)


Alizter:
When a ppx action is sandboxed, it fails for reason files since for whatever reason, the ppx executable requires the original `hello.re` source when it really should only require `hello.re.ml` after we adapt for the dialect. The sandboxed action doesn't carry the original `.re` source hence the failure.

src/dune_rules/preprocessing.ml
```diff
@@ -703,13 +703,20 @@ let pp_one_module sctx ~lib_name ~scope ~preprocessor_deps
                         let dir =
                           Path.build (Super_context.context sctx).build_dir
                         in
+                        let original_src : Dep.Set.t =
+                          match Module.source m ~ml_kind with
+                          | None -> Dep.Set.empty
```
rgrinberg, line 708: How is `None` possible?

Alizter: Not sure.

Alizter: I guess I could code error here. I don't see any way this could be None.

rgrinberg: It should be `Option.value_exn` then.

emillon:
I reproduced that without `reason` so I'll push a test which does not rely on a fake refmt.

emillon:
@Alizter I rebased and pushed a fix. Let me know what you think. This will need a changelog entry.  Thanks!

rgrinberg:
@emillon this PR needs review from me and @anmonteiro still

emillon:
sure, to be clear I was just addressing the test

Alizter:
@emillon It already has a changelog entry, is something missing?

emillon:
> @emillon It already has a changelog entry, is something missing?

ugh, coffee for me I guess. sorry!

src/dune_rules/preprocessing.ml
```diff
@@ -710,6 +710,10 @@ let pp_one_module sctx ~lib_name ~scope ~preprocessor_deps
                           ; Path (Path.build dst)
                           ; Command.Ml_kind.ppx_driver_flag ml_kind
                           ; Dep (Path.build src)
+                          ; Hidden_deps
```
anmonteiro, line 713: I don't understand why this is needed. 

IIUC, the pipeline looks something like:

```
.re -> (dialect) .re.ml -> (pps) .pp.ml
```

So isn't the original source file already a (transitive) dependency of the pps action?

```

Comment (Alizter) [reply src/dune_rules/preprocessing.ml:713]: But sandboxing does not take the transitive closure, onoy the specified deoendencies. Thats why its important to always fully specify them.

**Review by anmonteiro — APPROVED** [review]

**Comment by Alizter** [comment]
@rgrinberg I've dropped the commit modifying the targets in inline_tests.ml. As you can see this creates an exception. I can fix it by tweaking the source files:
```diff
diff --git a/src/dune_rules/inline_tests.ml b/src/dune_rules/inline_tests.ml
index 77a92acc2..941900c09 100644
--- a/src/dune_rules/inline_tests.ml
+++ b/src/dune_rules/inline_tests.ml
@@ -285,7 +285,10 @@ include Sub_system.Register_end_point (struct
       in
       List.concat l
     in
-    let source_files = List.concat_map source_modules ~f:Module.sources in
+    let source_files =
+      List.concat_map source_modules ~f:(fun module_ ->
+          Module.ml_source module_ |> Module.sources)
+    in
     Memo.parallel_iter_set
       (module Mode_conf.Set)
       info.modes
diff --git a/test/blackbox-tests/test-cases/reason-expect.t/run.t b/test/blackbox-tests/test-cases/reason-expect.t/run.t
index d1c599f90..f694d3809 100644
--- a/test/blackbox-tests/test-cases/reason-expect.t/run.t
+++ b/test/blackbox-tests/test-cases/reason-expect.t/run.t
@@ -20,18 +20,7 @@ syntax and ppx_expect break at dune lang 3.3 due to the sandboxing of ppx.
   > (lang dune 3.9)
   > EOF
   $ dune test
-  File "dune", line 3, characters 1-15:
-  3 |  (inline_tests)
-       ^^^^^^^^^^^^^^
-  Fatal error: exception Sys_error("$TESTCASE_ROOT/_build/.sandbox/31552d8346d4edb1acb27f906d1a4b1b/default/./test.re.ml: No such file or directory")
-  Raised by primitive operation at Stdlib.open_in_gen in file "stdlib.ml", line 405, characters 28-54
-  Called from Stdio__In_channel.with_file in file "src/in_channel.ml", line 18, characters 45-66
-  Called from Stdio__In_channel.read_all in file "src/in_channel.ml" (inlined), line 86, characters 21-49
-  Called from Ppx_expect_evaluator.create_group in file "evaluator/ppx_expect_evaluator.ml", line 113, characters 22-69
-  Called from Base__List.count_map in file "src/list.ml", line 462, characters 13-17
-  Called from Ppx_expect_evaluator.evaluate_tests in file "evaluator/ppx_expect_evaluator.ml", line 207, characters 2-85
-  Called from Stdlib__List.map in file "list.ml", line 92, characters 20-23
-  Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
-  Called from Ppx_inline_test_lib__Runtime.exit in file "runtime-lib/runtime.ml", line 646, characters 2-49
-  Called from Dune__exe__Inline_test_runner_test in file ".test.inline-tests/inline_test_runner_test.ml-gen", line 1, characters 9-44
+  File "test.re.ml", line 1, characters 0-0:
+  Error: Files _build/default/test.re.ml and
+  _build/default/test.re.ml.corrected differ.
   [1]
```

Alizter:
@rgrinberg Shall I split the later commits into another PR?

test/blackbox-tests/test-cases/reason-expect.t/refmt/refmt.ml
```diff
@@ -0,0 +1,52 @@
+let sprintf = Printf.sprintf
```
rgrinberg, line 1: Instead of copy pasting, what about moving this to `test-cases/utils` and making it available to both tests?

rgrinberg:
I don't understand how this diff is supposed to work:

```
+  File "test.re.ml", line 1, characters 0-0:
+  Error: Files _build/default/test.re.ml and
+  _build/default/test.re.ml.corrected differ.
```

We have a source file `test.re`, and `test.re.ml` is supposed to contain the correction for a generated file?! Why would we offer a promotion for a generated file?

rgrinberg:
Okay, the issue is that our clone of refmt is bugged because it doesn't output line directives. Unfortunately, inserting the line directive to correct the filename breaks ppx_expect. No idea what's happening but I pushed the commit that demonstrates the issue.

Anyway, we can investigate this later. Just remove the copy-pasting between the refmt's and this is good to go

test/blackbox-tests/test-cases/reason-expect.t/refmt/refmt.ml
Alizter: Its not an exact copy and paste which makes it awkward. One outputs to a file and the other outputs to stdout.

Alizter:
@rgrinberg Right, if you look at the commits, the last one I pushed fixes inline_tests. Before I get the same error as you did.

rgrinberg:
It's not the same error. The previous error was expected because the refmt error is broken. The subsequent is because ppx_expect doesn't understand directive for some reason

test/blackbox-tests/test-cases/reason-expect.t/refmt/refmt.ml
rgrinberg: Shouldn't it just copy how the original refmt binary works? The fact that it behaves differently in different tests is wrong

Alizter:
@rgrinberg I've consolidated the fake refmt and squashed your commit.

Alizter:
@rgrinberg Actually the backtrace in the bug you found isn't repro. My machine differs from the CI for some reason. Maybe we should leave the removing of the refmt dep to another PR and fix this inline tests issue?

rgrinberg:
We can't have the test suite depend on refmt

Alizter:
@rgrinberg Alright, shall I just scrub the backtrace then?

rgrinberg:
The back trace is easily fixable we just need to remove the locations.

That's not the point though, there's clearly something wrong with this fake refmt if it's making ppx_expect fail. We should fix it or it will break the test suite in other mysterious ways.

Alizter:
@rgrinberg You added this line to the fake refmt:
```
  output_string out (sprintf "# 1 %S\n" source);

```
That was intentional right?

rgrinberg:
Yes. Without this, how would ppx_expect find the original source?

anmonteiro:
What’s the refmt problem? Btw Reason can’t parse line directives, not sure if that’s related.

rgrinberg:
We have this fake `refmt` binary that we use for the tests. There's a problem with it's not working with ppx_expect for some reason (see exception). Perhaps you know what the reason is?

refmt doesn't need to understand line directives, it just needs to produce a parsetree with the correct locations. I think it does so with `--binary` which is what dune uses.

anmonteiro:
This looks like a ppx_expect bug rather than Reason? https://github.com/search?q=%22pos+%2B+len+past+end%22&type=code

anmonteiro:
My guess: the ppx_expect evaluator is trying to read the new file expecting the old file contents length?

Alizter:
Hmm I don't know how to progress with this now.

PR #7933, opened 2023-06-10
refactor(dune_lang): use fields_mutually_exclusive for aliases

Alizter:
I noticed that we have `fields_mutually_exclusive` so I've used it for the `aliases` field of the `rule` stanza.

### DIFF
--- a/src/dune_rules/dune_file.ml
+++ b/src/dune_rules/dune_file.ml
@@ -1704,27 +1704,16 @@ module Rule = struct
          field_o "package"
            (Dune_lang.Syntax.since Stanza.syntax (2, 0)
            >>> Stanza_common.Pkg.decode)
-       and+ alias =
-         field_o "alias"
-           (Dune_lang.Syntax.since Stanza.syntax (2, 0)
-           >>> Dune_lang.Alias.decode)
        and+ aliases =
-         field_o "aliases"
-           (Dune_lang.Syntax.since Stanza.syntax (3, 5)
-           >>> repeat Dune_lang.Alias.decode)
-       in
-       let aliases =
-         match alias with
-         | None -> Option.value ~default:[] aliases
-         | Some alias -> (
-           match aliases with
-           | None -> [ alias ]
-           | Some _ ->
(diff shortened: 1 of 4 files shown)


src/dune_rules/dune_file.ml
```diff
@@ -1761,8 +1748,8 @@ module Rule = struct
       | Action -> short_form
       | Since (version, inner) ->
         let what = Printf.sprintf "'%s' in short-form 'rule'" atom in
-        let* () = Dune_lang.Syntax.since ~what Stanza.syntax version in
-        interpret atom inner
+        Dune_lang.Syntax.since ~what Stanza.syntax version
+        >>> interpret atom inner
```
rgrinberg, line 1752: why did you make this change?

Alizter: Is it not the same thing? The idea was to make it more similar to the rest of the code.

Alizter: But if you are asking "why make this change here", it is part of the second commit. If you like I can split that into another PR.

rgrinberg: Yes to both points.

First, I don't see how this change is an improvement. Second, it's hard to review PR's that mix unrelated changes.

Alizter: OK I've left only the first commit.

test/blackbox-tests/test-cases/alias-multiple.t
```diff
@@ -100,6 +99,5 @@ Even if the aliases list is empty
   2 |  (alias a)
   3 |  (aliases)
   4 |  (action (echo "I have run")))
-  Error: The 'alias' and 'aliases' fields are mutually exclusive. Please use
-  only the 'aliases' field.
+  Error: fields alias, aliases are mutually exclusive
```
rgrinberg, line 102: The quotes should be back on the field names

Alizter: I have another patch doing this. Do you want me to include it here?

PR #7936, opened 2023-06-11
pkg: use Dyn.variant constructor for Op

Alizter:
Something I noticed going over the code.

### DIFF
--- a/src/dune_rules/package.ml
+++ b/src/dune_rules/package.ml
@@ -80,15 +80,8 @@ module Dependency = struct
     let map =
       [ ("=", Eq); (">=", Gte); ("<=", Lte); (">", Gt); ("<", Lt); ("<>", Neq) ]
 
-    let to_dyn =
-      let open Dyn in
-      function
-      | Eq -> string "Eq"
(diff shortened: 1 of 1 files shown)


gridbugs (approved):

src/dune_rules/package.ml
```diff
@@ -80,15 +80,13 @@ module Dependency = struct
...
-      | Lte -> string "Lte"
-      | Lt -> string "Lt"
-      | Neq -> string "Neq"
+    let to_dyn = function
+      | Eq -> Dyn.variant "Eq" []
+      | Gte -> Dyn.variant "Gte" []
+      | Lte -> Dyn.variant "Lte" []
+      | Gt -> Dyn.variant "Gt" []
+      | Lt -> Dyn.variant "Lt" []
+      | Neq -> Dyn.variant "Neq" []
```
rgrinberg, line 89: You can also just look it up from `map`. Save some code duplication this way.

Alizter: However we won't get the constructor names this way and only the symbols.;

PR #7937, opened 2023-06-11
feature: extend traces to include source scans

rgrinberg:
Source scans can be quite slow. Record them in the trace.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 Unreleased
 ----------
 
+- Include source tree scans in the traces produced by `--trace-file` (#7937,
+  @rgrinberg)
+
 - Cinaps: The promotion rules for cinaps would only offer one file at a time no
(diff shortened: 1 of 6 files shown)


rgrinberg:
Actually we no longer share the source tree code so there's nothing to review here.

snowleopard:
Can you show an example trace of some non-trivial build to see how much it takes?

rgrinberg:
So for doing a null build on dune, `dune.exe build @check` is around 2.5 seconds. 1.3 seconds of that is scanning the source tree.

snowleopard:
Oh, wow. I now want some similar tracing internally =)

Khady:
What does “source scans” refer to exactly? Is it only about checking timestamp and hashes, or does it also include the creation and/or execution of rules?

rgrinberg:
A source scan refers to traversing the entire source tree to extract all project and dune files.

PR #7940, opened 2023-06-11
fix: stop catching all exceptions in `Dune_package.Or_meta.load`

anmonteiro:
- I looked through the function calls for this part, and only explicitly noticed any `User_error.E` being raised
- Because we're parsing with a `lexbuf`, I've included `Sys_error` and `End_of_file` too, per the suggestion in https://github.com/ocaml/dune/pull/7873/files#r1225758114


src/dune_rules/dune_package.ml
```diff
@@ -466,8 +466,7 @@ module Or_meta = struct
            with
            | x -> Ok x
            | exception User_error.E message -> Error message
-           | exception exn ->
-             (* XXX stop catching code errors, invalid args, etc. *)
+           | exception ((Sys_error _ | End_of_file) as exn) ->
              Error (User_message.make [ Exn.pp exn ]))
```
rgrinberg, line 470: You can convert them to to proper error messages as well.

1. `Unexpected end of file when reading foo`
2. `[ Pp.textf "Failed to read %s." file; Pp.text sys_error ]`

anmonteiro: Good idea, fixed that in the latest push.

rgrinberg (approved):

PR #7945, opened 2023-06-12
doc: add reference info about aliases

### LINKED ISSUE #7615: Dune can be use to format non-OCaml files ! Is it an accidental feature?
Hi!

I've managed to setup dune to format files with a custom extension, using a pretty-printer defined in the repository as a binary, by adding actions to the `fmt` alias that looks like this:

```dune
(rule
 (with-stdout-to
  my-file.ext.pp.output
  (bash "%{bin:my-formatter} file %{dep:my-file.ext}")))

(rule
 (alias fmt)
 (action
  (diff my-file.ext my-file.ext.pp.output)))
```

This works just fine ! When running `dune fmt` all such files get nicely formatted and auto-promoted if needed.

I couldn't find a mention of this setup in the documentation nor in another repository, so I wonder if this is an accidental feature, or if it is reasonable to count on this behavior to keep working with future versions of dune.

## Generating rules for many files

I've gone one step further in that adding such pair of rules for each file in the dune file for each directory is a bit tedious. So in fact these rules are generated in a dune.inc file with another sub-command of the custom formatter.

```dune
(include dune.inc)

(rule
 (deps
  (glob_files *.ext))
 (action
  (with-stdout-to
   dune.inc.gen
   (pipe-stdout
    (bash "%{bin:my-formatter} gen-dune *.ext")
    (run dune format-dune-file)))))

(rule
 (alias runtest)
 (action
  (diff dune.inc dune.inc.gen)))
```

See an actual live example [here](https://github.com/mbarbin/bopkit/blob/main/test/netlist/check/dune.inc)

## Next step?

I suppose maybe someday there will be a more principal way of doing this (perhaps with customizable rules), but I wanted to report that it's very convenient that this already works with the current production dune. Thank you!

Perhaps this was made possible since the introduction of the `fmt` alias? How would you feel about documenting this workaround in the doc, and perhaps adding a test to avoid breaking this workflow? (If you think it is reasonable).

### Cinaps?

The setup to maintain `dune.gen` requires the stanza above (1 include and 2 rules) to be added to the dune files in every directory with files to format, so it is repeated many times. I would love to be able to use `cinaps` here, but I haven't managed to do so yet. I am considering whether to open another ticket regarding cinaps, so perhaps I'll talk about this some place else. I have an intuition that cinaps + generated dune files can go a long way as an alternate power of expression to a more general generic rule mechanism. What do yo think?

Thanks!

emillon:
This reworks the info found in `usage` and expands it.

Fixes #7615

### DIFF
--- a/doc/reference/aliases.rst
+++ b/doc/reference/aliases.rst
@@ -7,11 +7,9 @@ files. For example, the ``runtest`` alias corresponds to running tests.
 Model and Syntax
 ----------------
 
-Rules can be attached to an alias. When this alias is requested to be built,
-these rules fire and the actions attached to them are executed.
-
-Since rules are defined in a specific directory, aliases are attached to
-directories too.
+Dependencies and actions can be attached to an alias. When this alias is
+requested to be built, these dependencies are built and these actions are
+executed. Aliases are attached to specific directories.
 
 In commands such as ``dune build``, the syntax to refer to the ``x`` alias is
 ``@x``, for example ``dune build @x``. This is why it is common to refer to it
--- a/doc/reference/aliases.rst
+++ b/doc/reference/aliases.rst
@@ -41,7 +41,7 @@ Examples:
   the ``foo`` build context
 - ``dune build @runtest`` will run the tests for all build contexts
 
-User-defined Aliases
+User-Defined Aliases
 --------------------
 
 It is possible to use any name for alias names; it will then be available on
@@ -54,7 +54,7 @@ the command line. For example, if a Dune file contains the following, then
     (alias deploy)
     (action ./run-deployer.exe))
 
-Built-in Aliases
+Built-In Aliases
 ----------------
 
 Some aliases are defined and managed by Dune itself.
@@ -68,7 +68,7 @@ This alias corresponds to every known file target in a directory.
 ^^^^^^
 
 This alias corresponds to the set of targets necessary for development tools to
-work correctly. For example, it will build ``*.cmi``, ``*.cmt`` and ``*.cmti``
+work correctly. For example, it will build ``*.cmi``, ``*.cmt``, and ``*.cmti``
 files so that Merlin and ``ocaml-lsp-server`` can be used in the project.
 It is also useful in the development loop because it will catch compilation
 errors without executing expensive operations such as linking executables.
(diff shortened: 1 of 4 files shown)


doc/reference/aliases.rst
```diff
@@ -0,0 +1,160 @@
+Aliases
+=======
+
+:term:`Aliases <alias>` are build targets that do not correspond to specific
+files. For example, the ``runtest`` alias corresponds to running tests.
+
+Model and Syntax
+----------------
+
+Rules can be attached to an alias. When this alias is requested to be built,
```
rgrinberg, line 10: That's not completely correct. We can attach dependencies to aliases.

We can also attach aliases to rules, although the way it works under the hood is that we attach the actions rather than the rules.

doc/reference/aliases.rst
```diff
@@ -0,0 +1,160 @@
...
+- optional directory: only consider this subdirectory
+- alias name
+
+Examples:
+
+- ``dune build @_build/foo/runtest`` only runs the tests for
+  the ``foo`` build context
+- ``dune build @runtest`` will run the tests for all build contexts
+
+User-defined Aliases
```
christinerose, line 46:
```suggestion
User-Defined Aliases
```

doc/reference/aliases.rst
```diff
@@ -0,0 +1,160 @@
...
+@all
+^^^^
+
+This alias corresponds to every known file target in a directory.
+
+@check
+^^^^^^
+
+This alias corresponds to the set of targets necessary for development tools to
+work correctly. For example, it will build ``*.cmi``, ``*.cmt`` and ``*.cmti``
```
christinerose, line 73:
```suggestion
work correctly. For example, it will build ``*.cmi``, ``*.cmt``, and ``*.cmti``
```

doc/reference/aliases.rst
```diff
@@ -0,0 +1,160 @@
...
+
+@fmt
+^^^^
+
+This alias is used by formatting rules: when it is built, code formatters will
+be executed (using :doc:`promotion <../concepts/promotion>`).
+
+``dune fmt`` is a shortcut for ``dune build @fmt --auto-promote``.
+
+It is possible to build on top of this convention: if some actions are manually
```
christinerose, line 125:
```suggestion
It is possible to build on top of this convention. If some actions are manually
```

doc/reference/aliases.rst
```diff
@@ -0,0 +1,160 @@
...
+the command line. For example, if a Dune file contains the following, then
+``dune build @deploy`` will execute that command.
+
+.. code:: dune
+
+   (rule
+    (alias deploy)
+    (action ./run-deployer.exe))
+
+Built-in Aliases
```
christinerose, line 59:
```suggestion
Built-In Aliases
```

christinerose (changes requested):

PR #7946, opened 2023-06-12
create dune show command group

Alizter:
We ad a command group `dune show` which is an alias of `dune describe`. Motivation is that it is more convenient and easier to remember.

### DIFF
--- a/bin/describe/describe.ml
+++ b/bin/describe/describe.ml
@@ -7,6 +7,13 @@ open Import
 
    - duniverse people for "describe opam-files" *)
 
+let subcommands =
+  [ Describe_workspace.command
+  ; Describe_external_lib_deps.command
+  ; Describe_opam_files.command
+  ; Describe_pp.command
+  ]
+
 let group =
   let doc = "Describe the workspace." in
   let man =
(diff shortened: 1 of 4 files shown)


bin/describe/describe_pp.ml
```diff
@@ -42,11 +42,7 @@ let get_pped_file super_context file =
   let in_build_dir file =
     file |> Path.to_string |> Path.Build.relative context.build_dir
   in
-  let file_in_build_dir =
-    if String.is_empty file then
-      User_error.raise [ Pp.textf "no file is given" ]
-    else Path.of_string file |> in_build_dir |> Path.build
```
Alizter, line 48: This string is never empty since it is parsed by cmdliner.

bin/show/show.ml
rgrinberg, line 1: The extra directory seems gratuitous

bin/describe/describe_pp.ml
rgrinberg: Even when provided with quotes? E.g. `""`

bin/show/show.ml
Alizter: OK then I misunderstood, I'll revise this tomorrow.

bin/describe/describe_pp.ml
Alizter: Actually I'm not sure, I'll drop these changes.

Alizter:
@rgrinberg I've left only the creation of the command group which is exactly the same as dune describe now and always will be.

rgrinberg:
needs a CHANGES entry still.

bin/describe/describe.ml
```diff
@@ -7,6 +7,17 @@ open Import
...
 
+let subcommands =
+  [ Describe_workspace.command
+  ; Describe_external_lib_deps.command
+  ; Describe_opam_files.command
+  ; Describe_pp.command
+  ]
+
+let make_group info =
+  let default = Describe_workspace.term in
```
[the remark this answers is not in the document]
Alizter: I should note that `show` doesnt have this brhaviour. Do we want it to?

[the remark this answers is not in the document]
rgrinberg: Probably not. It was added for backwards compat with the original `dune describe` i'm assuming.

PR #7957, opened 2023-06-13
Inline tests: add node_flags

vouillon:
`node` is used to run inline tests generated by Js_of_ocaml. This allows to pass specific flags to this program.
So this adds an environment field `(js_of_ocaml (node_flags <flags>)` and a per library field `(inline_tests (node_flags <flags>))`.
Maybe that should rather be `(inline_tests (js_of_ocaml (node_flags <flags>)))` to emphasize that this only apply in `js` mode?
I still need to update the documentation.


hhugo:
`node` is not only used for inline_tests, it's also used for running tests from `(tests ..)` stanza.
```
$ ag jsoo_rules.runner
src/dune_rules/inline_tests.ml
237:        | Javascript -> Some Jsoo_rules.runner

src/dune_rules/test_rules.ml
44:              | `js -> Some Jsoo_rules.runner
```


Also, I wonder if it would be better to allow full customization of the js runner.

A user could then do `(js_of_ocaml (js_runner node --jitless %{input-file} %))`
or use another js vm.

There exist a workaround for running a js vm other than node but it's not very satisfactory.
Here is what's used in the tezos codebase.
```
(env
 (_
  (binaries (%{workspace_root}/src/tooling/node_wrapper.exe as node))))
```

vouillon:
> There exist a workaround for running a js vm other than node but it's not very satisfactory. Here is what's used in the tezos codebase.
> 
> ```
> (env
>  (_
>   (binaries (%{workspace_root}/src/tooling/node_wrapper.exe as node))))
> ```

Oh, I already do this to be able to write `run node` without having to specify options each time. But I did not realize that it affected inline tests as well. So, I don't really need this change after all.

Maybe a variable `js_runner` could be useful?

Dune seems to allow to specify flags rather than whole commands. For instance, there is a `CC` variable but only a `c_flags` environment setting. I'm wondering whether there is a strong reason for that, or whether there simply has not yet been any good reason to select a different executable.

hhugo:
@rgrinberg, do you have any opinion ?

rgrinberg:
> do you have any opinion ?

About the `js_runner` flag? Yeah it seems like a useful thing to have.

> For instance, there is a CC variable but only a c_flags environment setting. I'm wondering whether there is a strong reason for that, or whether there simply has not yet been any good reason to select a different executable.

People have asked to change the executable before. I don't think anyone would object such a way to set the  C compiler.

rgrinberg:
@hhugo what's the status of this?

vouillon:
@rgrinberg I need to understand how to add a variable `js_runner`, which would be a lot more useful than having just a `node_flags` option. Maybe I should close this PR for now?

rgrinberg:
Sure. Please re-open once you resume working on it

Issue #3449, opened 2020-05-04
[rfc] Support per-process catapult trace information


ejgallego:
As discussed in the meeting, it would be great to allow `--trace-file` to collect timing information provided by sub-processes, for example, OCaml could provide information about time spent in each phase or Coq could provide information about time spent in proofs.

I think there are two design choices to make:
- how to have the sub-processes communicate the timing information back to Dune. It seems to me that the easiest way is to make `Stats.with_process` aware of the trace output file, and include that information once the process has finished. This requires some extending of `Fiber.t` tho.
- how to instrument the corresponding call to processes when `--with-trace` is provided. As of today this info doesn't really live in the context so we'll likely have to propagate it there.

Thoughts?

Related issues:
- https://github.com/coq/coq/issues/11542
- https://github.com/ocaml/ocaml/pull/9082

snowleopard:
@rgrinberg Sounds good but I don't follow this bit:

> 1. dune will create a temporary directory for every action

We don't need a separate directory for every action. A single directory (say, `_build/.trace`) should be sufficient, right? Then, for every action we set `DUNE_TRACE_FILE=_build/.trace/<action_diges>` so that the action can write to it.

rgrinberg:
> We don't need a separate directory for every action. A single directory (say, _build/.trace) should be sufficient, right? Then, for every action we set DUNE_TRACE_FILE=_build/.trace/<action_digest> so that the action can write to it.

In my scheme, I allow for subcomponents of a particular action to write their own metrics as well. For example, if we run a `ppx` action, and different passes of the rewriter want to emit their own metrics - they're allowed to do so without worrying about the file being written to concurrently from somewhere else.

snowleopard:
I see, that makes sense. What I'd like to avoid is the need to eagerly create that directory for every action, even though most of them don't need it (at least at first). Do you think we can require the actions that want to produce custom traces to `mkdir` that directory?

rgrinberg:
I'm personally not worried about it as we create a ton of directories in dune builds and they never show up as overhead. Though allowing the actions to create them sounds fine as well to me.

PR #7960, opened 2023-06-13
feature: trace loading/marshalling files

rgrinberg:
Loading/Marshalling can be slow if the files being loaded are huge. We
record the overhead when --trace-file is enabled.

### DIFF
--- a/src/dune_util/persistent.ml
+++ b/src/dune_util/persistent.ml
@@ -49,13 +49,44 @@ module Make (D : Desc) = struct
 
   let to_string (v : D.t) = Printf.sprintf "%s%s" magic (Marshal.to_string v [])
 
-  let dump file (v : D.t) =
-    Io.with_file_out file ~f:(fun oc ->
-        output_string oc magic;
-        Marshal.to_channel oc v [])
+  let with_record stats ~name ~file ~f =
+    let start = Unix.gettimeofday () in
+    let res = Result.try_with f in
+    let event =
+      let stop = Unix.gettimeofday () in
+      let module Event = Chrome_trace.Event in
+      let module Timestamp = Event.Timestamp in
+      let dur = Timestamp.of_float_seconds (stop -. start) in
+      let common =
+        Event.common_fields ~name ~ts:(Timestamp.of_float_seconds start) ()
+      in
+      let args =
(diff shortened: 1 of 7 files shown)


rgrinberg:
@snowleopard ping on this one. should be simple.

src/dune_util/persistent.ml
```diff
@@ -72,7 +72,36 @@ module Make (D : Desc) = struct
...
+          fun file ->
+            let start = Unix.gettimeofday () in
+            let res = Result.try_with (fun () -> read_file file) in
+            let event =
+              let stop = Unix.gettimeofday () in
+              let module Event = Chrome_trace.Event in
+              let module Timestamp = Event.Timestamp in
+              let dur = Timestamp.of_float_seconds (stop -. start) in
+              let common =
+                Event.common_fields ~name:"Persistent"
```
snowleopard, line 90: Are users going to see only "Persistent" in the trace? Seems too terse/hard to guess what it is. How about "Loading persistent Dune state"?

rgrinberg: I changed it to `Loading Persistent State`. `Dune` seems obvious given the context.

snowleopard: I'm not so sure actually. There can be millions of events in the trace and which of them correspond to Dune may not be obvious. I don't see a good reason to be terse here.

snowleopard (approved):
Looks good though the current event name could be made even more explicit.

src/dune_util/persistent.ml
rgrinberg: Do you mean if we allow events to be written by other programs? Right now only dune is allowed to write to this trace file.

snowleopard: I mean, Dune runs lots of processes which can also be loading all sorts of states.

rgrinberg: Yup. But none of them can record anything in the trace file. Although we would like to change that https://github.com/ocaml/dune/issues/3449

Anyway, I can add "dune" to the name. I'm just not sure what you mean that it can be confused for some other state.

snowleopard: Thanks! 

> But none of them can record anything in the trace file.

They currently can't but Dune does that for them, right? Dune will create lots of events about other processes and distinguishing those from internal events may be non-trivial.

PR #7961, opened 2023-06-13
feature(pkg): include source fetching --trace-file

rgrinberg:


src/dune_pkg/fetch.ml
```diff
@@ -74,8 +74,27 @@ let fetch ~checksum ~target url =
   let label = "dune-fetch" in
   (* hashes have to be empty otherwise OPAM deletes the file after
```
Leonidas-from-XIV, line 75: This comment should probably move below the `event`.

Leonidas-from-XIV (commented):
Pretty useful change as downloading can take a significant amount of time when building.

PR #7965, opened 2023-06-14
Remove lock_dir field from Lock_dir.Pkg.t

gridbugs:
This field prevented us from representing package data with the lockdir path abstracted, and in cases where package data was read from a file in a lockdir, the value of the `lock_dir` field is always the path to the lockdir containing the file.

### DIFF
--- a/test/expect-tests/dune_pkg/dune_pkg_unit_tests.ml
+++ b/test/expect-tests/dune_pkg/dune_pkg_unit_tests.ml
@@ -153,25 +153,23 @@ let%expect_test "downloading, without any checksum" =
     Done downloading
     Finished successfully, no checksum verification |}]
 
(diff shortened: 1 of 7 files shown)


test/expect-tests/dune_pkg/dune_pkg_unit_tests.ml
```diff
@@ -228,88 +223,86 @@ let%expect_test "encode/decode round trip test for lockdir with complex deps" =
   let module Action = Dune_lang.Action in
   let module String_with_vars = Dune_lang.String_with_vars in
   lock_dir_encode_decode_round_trip_test ~lock_dir_path:"complex_lock_dir"
```
gridbugs, line 225: The large diffs in this file are from whitespace changes due to a function being replaced by a value.

PR #7969, opened 2023-06-14
refactor: clean up [Dune_rpc_server] api

rgrinberg:
* Add missing documentation
* Remove useless functions from the public API
* Remove pointless one liner helpers. No idea why I added those in the first place, but they're just making the code hard to follow.

### DIFF
--- a/src/dune_rpc_server/dune_rpc_server.mli
+++ b/src/dune_rpc_server/dune_rpc_server.mli
@@ -1,16 +1,6 @@
 open Stdune
 open Dune_rpc_private
 
-module Poller : sig
-  type t
-
-  val to_dyn : t -> Dyn.t
-
-  val compare : t -> t -> Ordering.t
-
-  val name : t -> Procedures.Poll.Name.t
-end
-
 module Session : sig
   module Id : Stdune.Id.S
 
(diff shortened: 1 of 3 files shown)


src/dune_rpc_server/dune_rpc_server.mli
```diff
@@ -127,11 +115,15 @@ module Handler : sig
...
+
+    type poller
+
     val implement_poll :
          's t
       -> 'a Procedures.Poll.t
-      -> on_poll:('s Session.t -> Poller.t -> 'a option Fiber.t)
-      -> on_cancel:('s Session.t -> Poller.t -> unit Fiber.t)
+      -> on_poll:('s Session.t -> poller -> 'a option Fiber.t)
+      -> on_cancel:('s Session.t -> poller -> unit Fiber.t)
```
snowleopard, line 126: This looks odd. What can you do with the `poller` if it's abstract?

rgrinberg: Nothing. This function is only used for tests and the tests don't care about the poller argument. I could specialize this function for the test and omit this argument.

src/dune_rpc_server/dune_rpc_server.mli
```diff
@@ -127,11 +115,15 @@ module Handler : sig
     -> unit
 
   module Private : sig
+    (** Raw long polling machinery only good for tests *)
```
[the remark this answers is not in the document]
rgrinberg: done.

src/dune_rpc_server/dune_rpc_server.mli
snowleopard: Yeah, I think we should drop this argument then, it's pretty confusing.

rgrinberg: Done.

snowleopard (approved):
Thanks!

Issue #7877, opened 2023-06-05
Add `dune pkg lock --all` to generate lockdirs for all contexts


gridbugs:
## Desired Behavior

Dune workspace contexts can specify a lockdir path. There should be a way to generate/update all the lockdirs in a single command `dune pkg lock --all`.

Related thread: https://github.com/ocaml/dune/pull/7835#issuecomment-1571234617

PR #7970, opened 2023-06-14
feature(pkg): Arguments to specify contexts to `dune pkg lock`

gridbugs:
Build contexts can specify custom lockdir paths. This commit adds an option --context when generating lockfiles to choose which build context to create the lockdir for, and a flag --all-contexts which generates lockdirs for all build contexts.

### DIFF
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -100,27 +100,89 @@ module Lock = struct
         (opam_package_name, opam_file))
     |> OpamPackage.Name.Map.of_list
 
+  (* Logic for choosing the lockdir path(s) to generate *)
+  let choose_lock_dir_paths ~context_name_arg ~all_contexts_arg =
+    let open Fiber.O in
+    match (context_name_arg, all_contexts_arg) with
+    | Some _, true ->
+      User_error.raise
+        [ Pp.text "--context and --all-contexts are mutually exclusive" ]
+    | context_name_opt, false -> (
+      let+ workspace = Memo.run (Workspace.workspace ()) in
+      let context_name =
+        Option.value context_name_opt ~default:Dune_engine.Context_name.default
+      in
+      let context =
+        List.find workspace.contexts ~f:(fun context ->
+            Dune_engine.Context_name.equal
+              (Workspace.Context.name context)
+              context_name)
+      in
+      match context with
+      | None ->
+        User_error.raise
+          [ Pp.textf "Unknown build context: %s"
+              (Dune_engine.Context_name.to_string context_name
+              |> String.maybe_quoted)
+          ]
+      | Some (Default { lock; _ }) ->
+        [ Option.value lock ~default:Lock_dir.default_path ]
+      | Some (Opam _) ->
+        User_error.raise
+          [ Pp.textf "Unexpected opam build context: %s"
+              (Dune_engine.Context_name.to_string context_name
+              |> String.maybe_quoted)
+          ])
+    | None, true ->
+      let+ workspace = Memo.run (Workspace.workspace ()) in
+      List.filter_map workspace.contexts ~f:(function
+        | Workspace.Context.Default { lock; _ } -> lock
+        | Opam _ -> None)
+
+  let context_term =
+    Arg.(
+      value
+      & opt (some Arg.context_name) None
+      & info [ "context" ] ~docv:"CONTEXT"
+          ~doc:
+            "Generate the lockdir associated with this context (the default \
+             context will be used if this is omitted)")
+
   let term =
     let+ (common : Common.t) = Common.term
-    and+ repo_selection = Repo_selection.term in
+    and+ repo_selection = Repo_selection.term
+    and+ context_name = context_term
+    and+ all_contexts =
+      Arg.(
+        value & flag
+        & info [ "all-contexts" ] ~doc:"Generate the lockdir for all contexts")
+    in
     let common = Common.forbid_builds common in
     let config = Common.init common in
-    Scheduler.go ~common ~config (fun () ->
-        let open Fiber.O in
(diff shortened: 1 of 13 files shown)


gridbugs:
Fixes https://github.com/ocaml/dune/issues/7877

bin/pkg.ml
```diff
@@ -101,25 +101,85 @@ module Lock = struct
...
+        ~doc:
+          "Generate the lockdir associated with this context (the default \
+           context will be used if this is omitted)"
+    and+ all_contexts =
+      Arg.(
+        value & flag
+        & info [ "all-contexts" ] ~doc:"Generate the lockdir for all contexts")
+    in
     let config = Common.init common in
     Scheduler.go ~common ~config (fun () ->
```
Alizter, line 157: You can put `@@` before `(fun` to get rid of some indentation (and then format).

bin/common.ml
```diff
@@ -1235,3 +1235,9 @@ let context_arg ~doc =
     value
     & opt Arg.context_name Dune_engine.Context_name.default
     & info [ "context" ] ~docv:"CONTEXT" ~doc)
+
+let context_arg_opt ~doc =
```
Alizter, line 1239: I'm not sure there is any point putting this in common since it is quite specialised. I would probably just inline it where you use it.

gridbugs: I was following suit since the non-optional version of the argument is defined above.

gridbugs:
Heads up that I updated this PR to add tests.

src/dune_pkg/lock_dir.mli
```diff
@@ -53,6 +53,14 @@ val metadata : Filename.t
...
-val write_disk : lock_dir_path:Path.Source.t -> t -> unit
+(** [write_disk_thunk ~lock_dir_path t] returns a thunk that when executed will
+    materialize [t] to disk at the path [lock_dir_path]. If [lock_dir_path]
+    already exists it is first deleted, provided it looks like an existing
+    lockdir. If it doesn't look like a valid lockdir an exception is raised
+    _before_ the thunk is constructed (ie. [write_disk_thunk ~lock_dir_path t]
+    will raise without the unit argument that causes the thunk to execute). This
+    allows callers to separate the validation of the existing lockdir from the
+    side effect that removes the existing lockdir and creates the new one. *)
+val write_disk_thunk : lock_dir_path:Path.Source.t -> t -> unit -> unit
```
rgrinberg, line 64: We usually use `Stage.t` for this idiom.

However, even with `Stage.t` I find this API to be confusing. I think you can just split this into functions:

```
type write
val prepare : .. -> write
val commit : write -> unit
```

src/dune_pkg/lock_dir.ml
```diff
@@ -313,18 +316,22 @@ let safely_remove_lock_dir_if_exists path =
...
+  in
+  fun () ->
+    remove_dir_if_exists_thunk ();
+    Path.mkdir_p lock_dir_path;
+    file_contents_by_path t
+    |> List.iter ~f:(fun (path_within_lock_dir, contents) ->
+           let path = Path.relative lock_dir_path path_within_lock_dir in
+           Option.iter (Path.parent path) ~f:Path.mkdir_p;
+           let contents_string =
+             List.map contents ~f:Dune_lang.to_string |> String.concat ~sep:"\n"
```
rgrinberg, line 332: Unrelated, but we have `Io.write_lines` so this concatenation isn't needed.

bin/common.ml
Alizter: Right, but the non-optional version is used in multiple places, hence it is common. From what I can see, the only reason you want this "optional" version is so that you can check mutual exclusivity with the other flag. This is really a deficiency with cmdliner so it is understandable you had to go this way. However, I don't imagine any other tools using this argument hence my original suggestion.

bin/pkg.ml
```diff
@@ -101,26 +101,84 @@ module Lock = struct
...
+
   let term =
     let+ (common : Common.t) = Common.term
-    and+ repo_selection = Repo_selection.term in
+    and+ repo_selection = Repo_selection.term
+    and+ context_name =
+      Common.context_arg_opt
+        ~doc:
+          "Generate the lockdir associated with this context (the default \
+           context will be used if this is omitted)"
```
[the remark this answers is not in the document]
Alizter: It's because of the mutual exclusivity check. Since common.context_arg defaults to default, there is no way of knowing if it wasn't set.

src/dune_pkg/lock_dir.mli
rgrinberg: We usually use `Staged.t` for optimization. E.g. `val find : corpus -> (query -> result) Staged.t`. But you're using to delay a side-effect. It's hard to tell what's the actual side effect that was done immediately versus the one that was delayed to the thunk.

bin/pkg.ml
```diff
@@ -101,26 +101,89 @@ module Lock = struct
...
-          (Dune_pkg.Opam.Summary.selected_packages_message summary);
-        Lock_dir.write_disk ~lock_dir_path lock_dir;
-        Fiber.return ())
+    Scheduler.go ~common ~config @@ fun () ->
+    let open Fiber.O in
+    let* lock_dir_paths =
+      choose_lock_dir_paths ~context_name_arg:context_name
+        ~all_contexts_arg:all_contexts
+    in
+    let* source_dir = Memo.run (Source_tree.root ()) in
```
Alizter, line 168: You can make this `let+` dropping the last `Fiber.return ()`.

rgrinberg:
@gridbugs could you rebase this one please?

rgrinberg:
Tests need to be updated

gridbugs:
Ah yep. Done.

gridbugs:
Rebased!

PR #7973, opened 2023-06-15
Monorepo benchmark tidy up

gridbugs:
Several changes aimed to make the monorepo benchmark easier to maintain:
- Use the benchmark runner's opam file to install its dependencies rather than explicitly installing them in the dockerfile.
- Stop pinning dune-rpc and dune-rpc-lwt. This was needed prior to dune 3.8 due to some recent bug fixes which have since been released.
- Document the runner command in the makefile.


gridbugs:
Oh looks like we still need to pin dune-rpc-lwt because of another recent bug fix. Will update.

gridbugs:
@rgrinberg can you take a look at this please

Issue #7974, opened 2023-06-15
Treat lockdirs as data-only


gridbugs:
## Desired Behavior

A dune lockdir contains dependency data in per-package files named after the packages themselves. This creates a problem where if the package "dune" is among the dependencies, its data will be stored in a file named "dune" which leads to dune attempting to interpret the file as a regular dune file.

To solve this we will treat lockdirs as data-only directories which will prevent dune from attempting to parse dune files inside lockdirs. This means that when determining whether a directory is "data-only" dune needs to know which directories are lockdirs. Currently custom lockdirs can be specified in a `dune-workspace` file and they can have any name. This could lead to the following scenario:
1. a user generates a lockdir for their project which depends on the package "dune"
2. they change the name of the default lockdir in `dune-workspace` without deleting the original lockdir
3. `dune build` attempts to interpret the package data in "dune.lock/dune" as a dune file and gives an error like:

```
File "dune.lock/dune", line 1, characters 1-8:
1 | (version 3.7.1)
     ^^^^^^^
Error: Unknown constructor version
```
...which is confusing.

One way of preventing this scenario is to enforce a naming convention for lockdirs: all directories whose names end in ".lock" will be treated as data-only directories, and only direcotries whose names end in ".lock" can be configured as lockdirs.

Leonidas-from-XIV:
I am a little bit averse to "magic" directories, because it might lead to some people having directories named `foo.lock` that suddenly get special treatment by accident (e.g. a lock datastructure subpackage for library `foo`). What I would suggest instead is to create a `dune` file in the folder that declares the folder as data-only.

This of course leads to a conflict with locking dune. In opam-monorepo we deal with this by renaming `dune` to `dune_`, maybe that's an option.

Alternatively, we could programmatically add such a definition on the folder containing the `.lock` directory, but I don't think there is a programmatic way to add stanzas to peoples' rules since they might be autogenerated or from an `.ml` file etc.

rgrinberg:
>  I do understand the value of convention over configuration, but in dune we usually tend towards the "explicit is better than implicit" side of the scale.

I think we tended to go the other way actually. Although I don't think this is a rule that can be consistently applied anyway.

Looking back, I do quite like my initial low tech suggestion of using file extensions. What's the problem of appending `.pkg` to every package definition file? Yeah, it's a little ugly but these things are generated anyway.

Alizter:
@gridbugs is this completed or rather not planned?

[...]

PR #7979, opened 2023-06-16
Lockdirs are data-only

gridbugs:
When a package depends on dune the lockdir will contain a file named "dune" which is not a valid dune file (instead it contains information about the package "dune"). To prevent dune attempting to interpret the file as a dune file, we make all lockdirs data-only.

To simplify this (both conceptually for users and the implementation) we introduce a convention where only directories ending in ".lock" can be specified as lockdirs, and all directories ending in ".lock" are treated as data-only.

Fixes https://github.com/ocaml/dune/issues/7974

### DIFF
--- a/src/dune_pkg/dune_pkg.ml
+++ b/src/dune_pkg/dune_pkg.ml
@@ -1,5 +1,6 @@
 module Fetch = Fetch
 module Checksum = Checksum
 module Lock_dir = Lock_dir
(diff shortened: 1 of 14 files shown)


gridbugs:
Closing following the discussion at https://github.com/ocaml/dune/issues/7974 and opting to add the .pkg extension to lockfiles instead.

PR #7981, opened 2023-06-16
test: absorb log output in action runner test

emillon:
This log was being output on dune's stderr instead.


rgrinberg:
Thanks. I think we can even drop the log for now. I only added for debugging this test. So commenting it out or guarding it behind `if false` seems better to me.

emillon:
Sure! How does that look?

rgrinberg (approved):

PR #7985, opened 2023-06-16
feature: add dune show env and make dune printenv an alias of it

Alizter:


bin/main.ml
```diff
@@ -17,7 +17,7 @@ let all : _ Cmdliner.Cmd.t list =
     ; Print_rules.command
     ; Utop.command
     ; Promotion.promote
-    ; Printenv.command
+    ; command_alias Printenv.command Printenv.term "printenv"
```
rgrinberg, line 20: I'd pick a different name. {describe,show} + print sounds very redundant. `env` alone sounds better.

rgrinberg (changes requested):
Needs a CHANGES entry

bin/main.ml
Alizter: This is the old command which was called `dune printenv`, the new one is `dune show env`.

rgrinberg (changes requested):
CHANGES please

Alizter:
> CHANGES please

done

Issue #7988, opened 2023-06-16
move glob syntax doc to dune-glob


emillon:
Once odoc 2.3.0 is released, we can use table support to move the glob syntax documentation [here](https://github.com/ocaml/dune/blob/main/doc/concepts/dependency-spec.rst#glob) to dune-glob's odoc.

PR #7989, opened 2023-06-16
doc: add index.mld for dune-glob

emillon:
As for other packages: an introduction, an example, a proper link.

The doc for `of_string` could point to a `syntax.mld` (this lives in `dependency-spec.rst` at the moment), but that would be better with table support. See #7988.

### DIFF
--- a/otherlibs/dune-glob/dune
+++ b/otherlibs/dune-glob/dune
@@ -1,0 +1,2 @@
+(documentation
+ (package dune-glob))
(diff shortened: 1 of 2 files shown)


rgrinberg:
You should probably mention that this library is not stable. It has a "versioned" api but that's only because it's needed for the DAP. Since the DAP is not stable either, this package doesn't need a stable API.

rgrinberg:
It hasn't changed in a long time so I suppose it's safe enough. But I would still mention somewhere that we don't actually have any stability guarantees like the module name might imply.

emillon:
I added a section about that.

rgrinberg (approved):

PR #7990, opened 2023-06-16
update libs in bootlib

Alizter:
This keeps being promoted for me when doing `@check`.


Alizter:
Should we think about removing the autopromotion for this actually?

rgrinberg:
It would be annoying in development. We don't want this manual promotions step usually.

I suppose what we need is a way to make dune fail if there any pending promotions.

PR #7996, opened 2023-06-17
fix(rpc): do not ignore unsent notifications

rgrinberg:
If we try to send a notification that the client doens't understand, we
just drop it on the floor. Now, we raise a code error.

This new behavior isn't ideal, but it's better than leaving this
function in this state.

### DIFF
--- a/src/dune_rpc_server/dune_rpc_server.ml
+++ b/src/dune_rpc_server/dune_rpc_server.ml
@@ -206,14 +206,11 @@ module Session = struct
   let of_stage1 (base : _ Stage1.t) handler =
     { base; handler; pollers = Dune_rpc_private.Id.Map.empty }
 
-  let notification t decl n =
-    let* () = Fiber.return () in
-    match V.Handler.prepare_notification t.handler decl with
-    | Error _ ->
-      (* cwong: What to do here? *)
-      Fiber.return ()
-    | Ok { Versioned.Staged.encode } ->
-      t.base.send (Some [ Notification (encode n) ])
+  let prepare_notification t decl =
+    V.Handler.prepare_notification t.handler decl
+
+  let send_notification t { Versioned.Staged.encode } n =
+    t.base.send (Some [ Notification (encode n) ])
 
   let request t decl id req =
     let* () = Fiber.return () in
--- a/src/dune_rpc_server/dune_rpc_server.mli
+++ b/src/dune_rpc_server/dune_rpc_server.mli
@@ -16,9 +16,17 @@ module Session : sig
   (** [get session a] sets the current state to [a].*)
   val set : 'a t -> 'a -> unit
 
-  (** [notification session n a] Send notification [a] defined by [n] to
+  (** [prepare t n] prepares a notification [n] by checking if [t] supports it.
+      If it supports it, [Ok _] is returned. Otherwise [Error _] is returned. *)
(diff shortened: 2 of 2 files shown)


snowleopard:
Is raising really better? It could only annoy/confuse users, though I guess it could help developers. If this change is aimed more at developers, should it live behind some debug flag?

If it's aimed for users, raising doesn't seem like the right tool. We could print a warning, once, instead. Wouldn't that be better?

rgrinberg:
It's aimed at users. In the case that the notification sent to the client is important, it would be better to get a loud error like this over silent breakage.

However, I could improve the behavior for developers as well. I could change the type to force developers to handle the case where the client doesn't support the notification yet.

snowleopard:
> It's aimed at users. In the case that the notification sent to the client is important, it would be better to get a loud error like this over silent breakage.

I agree with that but flooding users with lots of repeated error messages (which is what I think will happen with the current change), isn't nice. Could we do better?

snowleopard:
> However, I could improve the behavior for developers as well. I could change the type to force developers to handle the case where the client doesn't support the notification yet.

That does sound better, though you'll need to somehow indicate the problem to the user, and how to do that is still a design question.

rgrinberg:
> I agree with that but flooding users with lots of repeated error messages (which is what I think will happen with the current change), isn't nice. Could we do better?

I could flood the log I guess. I'm a bit hesitant to do any kind of remembering of already reported errors because that itself could eat a lot of memory.

> That does sound better, though you'll need to somehow indicate the problem to the user, and how to do that is still a design question.

If I change the type signature, I will force callers of this function to handle this condition better so it would be up to them to decide what's the best way to report this problem to the user - if at all it needs to be reported.

snowleopard:
> If I change the type signature, I will force callers of this function to handle this condition better so it would be up to them to decide what's the best way to report this problem to the user - if at all it needs to be reported.

Seems like this might work out better indeed. Let's give it a try?

rgrinberg:
> Seems like this might work out better indeed. Let's give it a try?

Done.

I also realized that our `request` function suffers from the same problem:

```ocaml
  let request t decl id req =
    let* () = Fiber.return () in
    match V.Handler.prepare_request t.handler decl with
    | Error error ->
      Code_error.raise "client doesn't support request"
        [ ("id", Dune_rpc_private.Id.to_dyn id)
        ; ("error", Dune_rpc_private.Version_error.to_dyn error)
        ]
```

Which means that anytime we send a request the client doesn't support, we just crash.

I'll address it separately.

snowleopard (approved):
The current change seems clearly fine but I'm confused -- surely we also need to update some some call sites? Where are they?

rgrinberg:
> surely we also need to update some some call sites? Where are they?

I checked the history and there were at some point but not anymore. I could delete this function altogether, but then we'd need to re-introduce it on the first notification we'd want to send.

snowleopard:
> I checked the history and there were at some point but not anymore. I could delete this function altogether, but then we'd need to re-introduce it on the first notification we'd want to send.

Huh, odd. Anyway, I'm happy to keep these functions, especially now that they provide a safe interface.

(I wonder if we could have some call sites left internally at Jane Street.)

rgrinberg:
I'd say let's keep it. It's bound to be re-introduced anyway.

PR #7999, opened 2023-06-17
chore: halve the number of ci runs

rgrinberg:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,8 +1,10 @@
 name: CI
 
 on:
-  - push
-  - pull_request
+  pull_request:
+  push:
+    branches:
+      - main
   - workflow_dispatch
 
 concurrency:


Alizter:
This will mean it won't run on peoples forks when they push to a branch. But I guess it can be manually dispatched instead.

Alizter (approved):

rgrinberg:
Hmm, this doesn't seem right. There's no CI on this PR.

.github/workflows/workflow.yml
```diff
@@ -1,8 +1,10 @@
 name: CI
 
 on:
-  - push
-  - pull_request
+  pull_request:
```
Alizter, line 4: I think you made the condition for pull requests empty.

Alizter: (remove the colon)

rgrinberg: Didn't make a difference. Anyway, I copied it from ocamllsp and it seems to work there.

rgrinberg:
Done by Rafal

PR #8002, opened 2023-06-17
chore: remove dune_graph.ml

Alizter:
This library is wrapped anyway.


rgrinberg:
Leave it. It's a good thing to have explicit interfaces

PR #8005, opened 2023-06-18
Validate top-module input

### LINKED ISSUE #8004: `dune ocaml top-module` crashes when file doesn't have `.ml` extension or is directory
## Expected Behavior
When passing invalid output to `top-module` I expect error message that explains the problem

## Actual Behavior
```
rafal@machine ocaml-lsp % dune ocaml top-module dune
Error: exception Invalid_argument("Filename.chop_extension")
Raised at Stdlib.invalid_arg in file "stdlib.ml", line 30, characters 20-45
Called from Main__top.Module.module_directives in file "bin/ocaml/top.ml",
  line 106, characters 17-70
Called from Main__top.Module.term.(fun) in file "bin/ocaml/top.ml", line 229,
  characters 14-48
Called from Fiber__scheduler.exec in file "otherlibs/fiber/src/scheduler.ml",
  line 73, characters 8-11

```

## Reproduction
Pass to `dune ocaml top-module` directory or file without `.ml` extension

3Rafal:
Fixes #8004

### DIFF
--- a/bin/ocaml/top.ml
+++ b/bin/ocaml/top.ml
@@ -101,13 +101,11 @@ module Module = struct
     let src = Path.Build.append_source ctx.build_dir mod_ in
     let dir = Path.Build.parent_exn src in
     let filename = Path.Build.basename src in
-    if Sys.file_exists filename then ()
-    else User_error.raise [ Pp.text "file doesn't exist" ] ;
-    if Filename.extension filename  = ".ml" then ()
-    else User_error.raise [ Pp.text "file should have '.ml' extension" ] ;
+    if Filename.extension filename = "" then
+      User_error.raise [ Pp.text "file is missing an extension" ];
     let open Memo.O in
     let module_name =
-      let name = src |> Path.Build.basename |> Filename.chop_extension in
+      let name = Filename.chop_extension filename in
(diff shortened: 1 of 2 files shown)


bin/ocaml/top.ml
```diff
@@ -100,6 +100,11 @@ module Module = struct
...
     let src = Path.Build.append_source ctx.build_dir mod_ in
     let dir = Path.Build.parent_exn src in
+    let filename = Path.Build.basename src in
+    if Sys.file_exists filename then ()
+    else User_error.raise [ Pp.text "file doesn't exist" ] ;
+    if Filename.extension filename  = ".ml" then ()
+    else User_error.raise [ Pp.text "file should have '.ml' extension" ] ;
     let open Memo.O in
     let module_name =
       let name = src |> Path.Build.basename |> Filename.chop_extension in
```
rgrinberg, line 110: This fix isn't quite right because we allow other dialects apart from OCaml. ReasonML is one example. The file existence checks are also done later so they aren't needed.

I think we should just reject files without an extension and be done with it. Could you update the PR to do that?

3Rafal: Done

PR #8006, opened 2023-06-18
refactor(dap): hide the implementation of dynamic dependencies

rgrinberg:
The representation of dynamic dependencies as they're implemented by DAP
is an implementation detail that the engine shouldn't be concerned with.

Instead, we reuse the standard representation of dependencies.

### DIFF
--- a/src/dune_engine/action_exec.mli
+++ b/src/dune_engine/action_exec.mli
@@ -1,29 +1,12 @@
 open Import
 
-(** Type for dependency requested by the dynamic action.
-
-    Must be different from [Dep.t] type because we require it to be
-    marshallable. *)
-module Dynamic_dep : sig
-  type t =
-    | File of Path.t
-    | Glob of Path.t * Glob.t
-
-  val to_dep : t -> Dep.t
-
-  val compare : t -> t -> Ordering.t
-
-  module Map : Map.S with type key := t
-
-  module Set : sig
-    include Set.S with type elt = t and type 'a map = 'a Map.t
-
-    val to_dep_set : t -> Dep.Set.t
-  end
-end
-
 module Exec_result : sig
-  type t = { dynamic_deps_stages : (Dynamic_dep.Set.t * Dep.Facts.t) List.t }
+  type t =
+    { dynamic_deps_stages :
+        (* The set can be derived from the facts by getting the keys of the
+           facts map. We don't do it because conversion isn't free *)
+        (Dep.Set.t * Dep.Facts.t) list
+    }
 end
 
 type input =
--- a/src/dune_engine/action_exec.ml
+++ b/src/dune_engine/action_exec.ml
@@ -65,7 +65,7 @@ module Dynamic_dep = struct
 end
 
 module Exec_result = struct
-  type t = { dynamic_deps_stages : (Dynamic_dep.Set.t * Dep.Facts.t) List.t }
+  type t = { dynamic_deps_stages : (Dep.Set.t * Dep.Facts.t) list }
 end
 
 type done_or_more_deps =
@@ -540,9 +540,8 @@ let exec_until_all_deps_ready ~display ~ectx ~eenv t =
     match result with
     | Done -> Fiber.return stages
     | Need_more_deps (relative_deps, deps_to_build) ->
-      let* fact_map =
-        ectx.build_deps (Dynamic_dep.Set.to_dep_set deps_to_build)
-      in
+      let deps_to_build = Dynamic_dep.Set.to_dep_set deps_to_build in
+      let* fact_map = ectx.build_deps deps_to_build in
       let stages = (deps_to_build, fact_map) :: stages in
       let eenv =
         { eenv with
--- a/src/dune_engine/rule_cache.ml
+++ b/src/dune_engine/rule_cache.ml
@@ -14,7 +14,7 @@ module Workspace_local = struct
     module Entry = struct
       type t =
         { rule_digest : Digest.t
-        ; dynamic_deps_stages : (Action_exec.Dynamic_dep.Set.t * Digest.t) list
+        ; dynamic_deps_stages : (Dep.Set.t * Digest.t) list
         ; targets_digest : Digest.t
         }
 
@@ -23,7 +23,7 @@ module Workspace_local = struct
           [ ("rule_digest", Digest.to_dyn rule_digest)
           ; ( "dynamic_deps_stages"
             , Dyn.list
-                (Dyn.pair Action_exec.Dynamic_dep.Set.to_dyn Digest.to_dyn)
+                (Dyn.pair Dep.Set.to_dyn Digest.to_dyn)
                 dynamic_deps_stages )
           ; ("targets_digest", Digest.to_dyn targets_digest)
           ]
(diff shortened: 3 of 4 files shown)


snowleopard (commented):
I'm not sure it's actually an improvement. You are hiding some details, sure, but also make the types less precise. Also, as we discussed, I'm currently doing a lot of changes around dependency representation and would prefer to avoid touching this code for now.

src/dune_engine/action_exec.mli
```diff
@@ -1,29 +1,12 @@
 open Import
 
-(** Type for dependency requested by the dynamic action.
-
-    Must be different from [Dep.t] type because we require it to be
-    marshallable. *)
-module Dynamic_dep : sig
-  type t =
```
snowleopard, line 8: You could just make this type abstract instead of removing it from the API altogether. That would keep the types precise.

rgrinberg:
We can wait, but I think this change can only help your refactoring. You have one less dependency type to worry about with this change.

src/dune_engine/action_exec.mli
rgrinberg: It's precise in the sense that it reflects a limitation of our implementation of the DAP. We could easily lift this implementation (fully or partially) and the rest of the engine should be(and is currently) oblivious to it. Removing this type from the public API actually enforces that, so for that reason I think this change is positive and is a pre-req for the DAP to be moved outside the engine entirely.

snowleopard: We may be able to lift some limitations but that's future work, which is non-trivial and isn't urgent, so who knows if/when it actually happens. In the current world, it does lose precision.

snowleopard: > pre-req for the DAP to be moved outside the engine entirely

Surely, this can't be done fully. I feel like we shouldn't rush with any of this.

rgrinberg: > We may be able to lift some limitations but that's future work, which is non-trivial and isn't urgent, so who knows if/when it actually happens

Note that I already experimented with this and it was quite easy. Just a matter of adding a few more serializers/deserializers to dap and a few more conversion functions between dap types and dune's own types. I imagine they weren't added right from the beginning since it was just a prototype.

> In the current world, it does lose precision.

Doesn't that happen with also make it abstract? Whichever way we do it, I think losing precision here is actually a good thing. It would be quite weird if someone added code that would somehow special case a feature that we haven't even made generally available yet.

> Surely, this can't be done fully. I feel like we shouldn't rush with any of this.

I did discuss the general design with you. I think we reached an agreement that it would be good to move DAP outside the engine. Did I misread something?

Indeed it can't be done fully. As far as I can tell, there are two essential components to this feature:

1. The ability of an action to request deps to be built as its being executed
2. The recording of the various deps that were requested in each stage (the "trace")

I would like the engine's interface to this feature to expose only these 2 things and leave the DAP itself to handle low level details such as serialization formats, versioning, etc.

Now our current implementation chose different representations for the deps of step 1. and 2. I can see why it was the easiest thing to do, but I don't think it's the right thing to do. So I thought it would be a good thing to address before we add some code that starts relying on this distinction.

snowleopard: > Doesn't that happen with also make it abstract?

Yeah, I suppose making it abstract is also sending a wrong signal that the structure of dynamic dependencies doesn't matter.

> Whichever way we do it, I think losing precision here is actually a good thing.

I think I disagree. Some features like caching dynamic actions are just impossible without the precision. Caching, for example, can only work if we know the list of stages at which dynamic dependencies were requested, so losing this structure in a set seems like mistake that we'll have to undo in future. (At least that's my current understanding without thinking more deeply about all the subtleties.)

> I did discuss the general design with you. I think we reached an agreement that it would be good to move DAP outside the engine. Did I misread something?

I agree that doing that is good in principle, but this doesn't contradict with it being non-trivial and non-urgent. It's not on the critical path of any of the big changes we're planning in the next year or so (or is it?). At the same time, doing changes to the code that is about to change for unrelated reasons and spending time discussing how to split the code across the build engine boundary is going to slow us down. That just makes me unenthusiastic about touching DAP right now. That's a yak best shaven in some bright future when the build engine becomes more stable.

> So I thought it would be a good thing to address before we add some code that starts relying on this distinction.

What kind of code are you thinking about? It seems like you're strongly motivated by doing some changes around DAP, and I just don't understand why given our hands are going to be full with lots of other things in the next few months.

rgrinberg: > I think I disagree. Some features like caching dynamic actions are just impossible without the precision. Caching, for example, can only work if we know the list of stages at which dynamic dependencies were requested, so losing this structure in a set seems like mistake that we'll have to undo in future. (At least that's my current understanding without thinking more deeply about all the subtleties.)

But we are maintaining the list of stages! We are simply doing so without introducing any special dependency type. Our existing `Dep.Set.t` is already perfectly adequate and there's just a simple conversion step between the two. This list of stages is what I called the "trace", and I would indeed be very much against removing it.

> What kind of code are you thinking about? It seems like you're strongly motivated by doing some changes around DAP, and I just don't understand why given our hands are going to be full with lots of other things in the next few months.

Because I believe there are people outside the core team that are willing to experiment with it and improve it further. If we allowed them to do so, they'd be able to develop the DAP concurrently and hopefully inform the design of the engine and improve the DAP itself.

I also do no think that splitting the DAP from the core will necessarily slow us down. I kind of expected it to be a win-win actually. If the DAP is reduced to a couple of simple hooks, changes to the engine no longer need to worry about the full feature set and its subtleties.

snowleopard: > But we are maintaining the list of stages!

Oh, indeed. I actually misread a diff in the `mli`, thinking that the list got collapsed into a single `Dep.Set.t`. Sorry! I'm much less worried about this PR now that this has been clarified.

> Because I believe there are people outside the core team that are willing to experiment with it and improve it further. If we allowed them to do so, they'd be able to develop the DAP concurrently and hopefully inform the design of the engine and improve the DAP itself.

> I also do no think that splitting the DAP from the core will necessarily slow us down. I kind of expected it to be a win-win actually. If the DAP is reduced to a couple of simple hooks, changes to the engine no longer need to worry about the full feature set and its subtleties.

Right. OK, I think I can get behind this motivation. Thanks for spelling this all out for me. It's worth copying some of this into the PR description, so that the motivation is clear for everyone.

src/dune_engine/action_exec.mli
```diff
@@ -1,29 +1,12 @@
 open Import
 
-(** Type for dependency requested by the dynamic action.
-
-    Must be different from [Dep.t] type because we require it to be
-    marshallable. *)
```
snowleopard, line 6: Was this comment stale? I'm confused because it claimed we need something special compared to `Dep.t` but I can't spot where this special ability was used.

snowleopard (approved):
Thanks for clearing up the motivation and my confusion about preserving the structure! The change now looks pretty uncontroversial to me. Still confused about one of the deleted comments but that's probably not important.

src/dune_engine/action_exec.mli
rgrinberg: Yes everything is marshallable ever since we got rid of the damned predicates 🙂

snowleopard: Oh, right, cool :)

rgrinberg:
> Still confused about one of the deleted comments but that's probably not important.

This was cleared up in [src/dune_engine/action_exec.mli](https://github.com/ocaml/dune/pull/8006/files/64bcd29a15eeddc4551ab7c41193928a20227cf4#diff-d604b88972c5b4ef229e02ffb2bc80c42ce74b709cbb01da9eb63b324ad87df9) in case anyone else is reading.

PR #8007, opened 2023-06-18
chore: Halve the number of CI runs

3Rafal:


.github/workflows/workflow.yml
```diff
@@ -1,9 +1,11 @@
 name: CI
 
 on:
-  - push
-  - pull_request
-  - workflow_dispatch
+  push:
+    branches:
+      - master
```
anmonteiro, line 6:
```suggestion
      - main
```

anmonteiro: The main branch is called `main` in this repo

anmonteiro (approved):
This halves the number of CI runs for branches of this repository that open PRs, which seems like the intended fix.

note that for external forks the number of CI runs was already the same as now.

3Rafal:
@anmonteiro , this PR is a fixed version of #7999. Rudi asked me to see if I can fix it, and I created this one. :)

emillon:
This means we only test the version that's merged with main, not the contents of the pushed branch, right?
Also, this removes push builds for release branches like `3.8`. Probably not a big deal, but if there's a way to glob branches `3.*` it would be useful to keep this.

anmonteiro:
> This means we only test the version that's merged with main, not the contents of the pushed branch, right?

I don't know if I follow completely, but this still tests every commit pushed to a PR, if that's what you're wondering.

emillon:
> I don't know if I follow completely, but this still tests every commit pushed to a PR, if that's what you're wondering.

Ah, OK. I remember that travis had a distinction where it would run a build with the commit pushed to a PR, and also with the commit merged with `main`, to avoid the rare situation where the branch is green but fails when merging.

PR #8010, opened 2023-06-19
fix(rpc): proper finalization everywhere

rgrinberg:
The callbacks to handle an RPC session are three fold:

1. `on_init` - called when the session is initiated
2. `on_upgrade` - called when a client and server agree on the common subset of rpc they can use
3. `on_terminate` - called when the session is terminated

We need to make sure the following two properties hold:

* When `on_init` is called, then `on_terminate` must also  be called
* When any of these raise, the appropriate finalizers are called

This PR makes sure that these properties hold. The tests demonstrate that this property didn't hold before, but now we can see everything is called correctly in the presence of exceptions.

@snowleopard this should fix the issue that we observe internally that some errors keep on looping.

### DIFF
--- a/src/dune_rpc_impl/server.ml
+++ b/src/dune_rpc_impl/server.ml
@@ -132,8 +132,8 @@ module Clients = struct
     let result = { session } in
     Session.Id.Map.add_exn t id result
 
-  let remove_session t (session : _ Session.t) =
-    let id = Session.id session in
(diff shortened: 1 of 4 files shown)


rgrinberg:
@dkalinichenko-js can you review this one since you've worked on the code recently?

dkalinichenko-js (approved):
LGTM, the forward declaration in `new_session` is clunky but can't see a better way of doing it.

PR #8013, opened 2023-06-19
doc: fix rule stanza not using the action field

MisterDA:
…and remove trailing spaces in another commit.


emillon:
It's not invalid to omit `(action)` in `(rule)` if it's the only field. Does the example trigger an error?

MisterDA:
> It's not invalid to omit `(action)` in `(rule)` if it's the only field. Does the example trigger an error?

It does not; I didn't know that was possible and assumed the worst! Sorry for the noise.

emillon:
feel free to submit the space fix separately

PR #8014, opened 2023-06-19
Lockdir package files have .pkg extension

gridbugs:
This will allow us to put a dune file in the lockdir without it conflicting with the package file for the package "dune".


Leonidas-from-XIV (approved):
Looks like a good non-magic solution to me.

PR #8015, opened 2023-06-20
feature(pkg): extra sources

rgrinberg:
Opam allows for additional file sources to be specified. We add support
for these.

### DIFF
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -2,12 +2,14 @@ open Import
 open Dune_lang
 
 module Source = struct
+  type fetch =
+    { url : Loc.t * string
+    ; checksum : (Loc.t * Checksum.t) option
+    }
+
   type t =
     | External_copy of Loc.t * Path.External.t
-    | Fetch of
-        { url : Loc.t * string
-        ; checksum : (Loc.t * Checksum.t) option
-        }
+    | Fetch of fetch
 
   let remove_locs = function
     | External_copy (_loc, path) -> External_copy (Loc.none, path)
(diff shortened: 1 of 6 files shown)


src/dune_pkg/lock_dir.ml
```diff
@@ -191,23 +213,37 @@ module Pkg = struct
        and+ exported_env =
          field Fields.exported_env ~default:[]
            (repeat Dune_lang.Action.Env_update.decode)
+       and+ extra_sources =
+         field "extra_sources" ~default:[]
```
gridbugs, line 217:
```suggestion
         field Fields.extra_sources ~default:[]
```

src/dune_pkg/lock_dir.ml
```diff
@@ -62,30 +78,22 @@ module Source = struct
...
-                 match Checksum.of_string_user_error checksum with
-                 | Ok checksum -> Some (loc, checksum)
-                 | Error e -> raise (User_error.E e))
-             in
-             fun _ -> Fetch { url; checksum } )
+        , let+ fetch = fields decode_fetch in
+          fun _ -> Fetch fetch )
       ]
 
+  let encode_fetch_field { url = _loc, url; checksum } =
```
Leonidas-from-XIV, line 85: Why is `url` bound once to `_loc` here? The previous code was doing the same, but it doesn't seem to get used.

gridbugs: The `url` field is a tuple which is being unpacked. With parens it would be `{ url = (_loc, url); ... `.

Leonidas-from-XIV: Ah! I totally missed the `,`, thanks.

PR #8018, opened 2023-06-20
refactor(dap): remove unnecessary intermediate type

rgrinberg:
We have the following type:

```ocaml
type t =
  | File of Path.t
  | Glob of Path.t * Glob.t
```

That is exactly the same type as what we use to represent
dune-action-plugin dependencies except it uses typed paths [Path.t]
instead of raw strings like the dune-action-plugin

It was used as an intermediate stage when converting DAP dependencies
into dune dependencies. So previously, the conversions would go as:

```
DAP with string paths -> DAP with Path.t -> Dep.Set.t
```

The intermediate stage is unnecessary, and now we just convert:

```
DAP with string paths -> Dep.Set.t
```

The new code is both shorter, simpler, and faster.

### DIFF
--- a/src/dune_engine/action_exec.ml
+++ b/src/dune_engine/action_exec.ml
@@ -11,58 +11,24 @@ let maybe_async =
   in
   fun f -> (Lazy.force maybe_async) f
 
-(** A version of [Dune_action_plugin.Private.Protocol.Dependency] where all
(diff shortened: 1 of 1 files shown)


Alizter:
Looks good, any numbers on the "faster"?

snowleopard (approved):
Looks good!

PR #8020, opened 2023-06-21
Prefer oldest dependencies by default while solving packages

gridbugs:
This changes the default behaviour of `dune pkg lock` to prefer the newest versions of its dependencies and adds a flag to change it to prefer the oldest versions.

I also include a change that refactors the opam solver code into its own file to help readability.

### DIFF
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -156,6 +156,13 @@ module Lock = struct
       Arg.(
         value & flag
         & info [ "all-contexts" ] ~doc:"Generate the lockdir for all contexts")
+    and+ prefer_oldest =
+      Arg.(
+        value & flag
+        & info [ "prefer-oldest" ]
+            ~doc:
+              "Use the oldest possible version of packages when solving \
+               dependencies")
     in
     let common = Common.forbid_builds common in
     let config = Common.init common in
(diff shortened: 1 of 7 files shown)


gridbugs:
I understand that this is potentially a contentious topic. I made this change based on some conversations I've been having with people about package management and I'm convinced that by default we should match the policy used by opam.

tmattio:
> I understand that this is potentially a contentious topic. I made this change based on some conversations I've been having with people about package management and I'm convinced that by default we should match the policy used by opam.

Right - as discussed, this change would have many implications for the ecosystem and the opam-repository, so we want to have a clear plan and involve the community before taking a decision.

Leonidas-from-XIV:
I am not sure we need to have this configurable, it is well possible that by people having it configured in different ways, we reach a point where neither side has any benefits, since the upsides of MVS namely that newer packages don't break older ones don't actually materialize until people use it in a significantly large set of packages.

rgrinberg:
> I understand that this is potentially a contentious topic. I made this change based on some conversations I've been having with people about package management and I'm convinced that by default we should match the policy used by opam.

It's probably best to discuss these issues in this PR.

bin/pkg.ml
```diff
@@ -156,6 +156,13 @@ module Lock = struct
       Arg.(
         value & flag
         & info [ "all-contexts" ] ~doc:"Generate the lockdir for all contexts")
+    and+ prefer_oldest =
+      Arg.(
+        value & flag
+        & info [ "prefer-oldest" ]
```
rgrinberg, line 162: I think we'd rather have a flag like `--prefer-versions={oldest,newest}` so that we have a way to run the solver regardless what the default might be.

gridbugs:
> > I understand that this is potentially a contentious topic. I made this change based on some conversations I've been having with people about package management and I'm convinced that by default we should match the policy used by opam.
> 
> It's probably best to discuss these issues in this PR.

Actually I think I'll close this PR and make an issue where we can discus this further.

gridbugs:
https://github.com/ocaml/dune/issues/8021

PR #8023, opened 2023-06-21
feature: allow arch_sixtyfour in enabled_if

### LINKED ISSUE #7997: Support the "word_size" variable in "enabled_if".
## Desired Behavior

The variables accessible as part of the `enabled_if` field of the `library` stanza (and probably others like `executable`) only support a limited range of variables that does not include `word_size`. I think it would make sense to include it so that one can restrict the compilation of a library to 64-bit architectures, which makes sense if you rely, e.g., on an `int` counters not overflowing, but you do not want to pay the extra cost of using an `Int64.t`.

Alizter:
We allow the `arch_sixtyfour` variable in `enabled_if`.

- fix #7997 
- [x] changelog
- [ ] docs (apparently we don't document enabled_if in this much detail).
- [x] tests

### DIFF
--- a/src/dune_rules/enabled_if.ml
+++ b/src/dune_rules/enabled_if.ml
@@ -18,6 +18,7 @@ let common_vars_list =
   ; "profile"
   ; "ocaml_version"
   ; "context_name"
+  ; "arch_sixtyfour"
   ]
 ;;
 
@@ -27,6 +28,7 @@ let common_vars ~since =
        ~f:(fun var ->
          match var with
          | "context_name" -> var, (2, 7)
+         | "arch_sixtyfour" -> var, (3, 11)
          | _ -> var, since)
        common_vars_list)
 ;;
--- a/test/blackbox-tests/test-cases/enabled_if/eif-arch_sixtyfour.t/run.t
+++ b/test/blackbox-tests/test-cases/enabled_if/eif-arch_sixtyfour.t/run.t
@@ -1,0 +1,22 @@
+Testing %{arch_sixtyfour} in enabled_if
+
+  $ cat > dune-project << EOF
+  > (lang dune 3.11)
+  > EOF
+
+  $ dune exec -- ./hello.exe
+  Hello, World!
+
+Testing the version guard
+
+  $ cat > dune-project << EOF
+  > (lang dune 3.10)
+  > EOF
+
+  $ dune exec -- ./hello.exe
+  File "dune", line 4, characters 5-22:
+  4 |   (= %{arch_sixtyfour} %{arch_sixtyfour})))
+           ^^^^^^^^^^^^^^^^^
+  Error: %{arch_sixtyfour} is only available since version 3.11 of the dune
+  language. Please update your dune-project file to have (lang dune 3.11).
+  [1]
--- a/doc/changes/8023.md
+++ b/doc/changes/8023.md
@@ -1,0 +1,3 @@
+
+- `enabled_if` now supports `arch_sixtyfour` variable (#8023, fixes #7997,
+  @Alizter)
--- a/test/blackbox-tests/test-cases/enabled_if/eif-arch_sixtyfour.t/dune
+++ b/test/blackbox-tests/test-cases/enabled_if/eif-arch_sixtyfour.t/dune
@@ -1,0 +1,4 @@
+(executable
+ (name hello)
+ (enabled_if
+  (= %{arch_sixtyfour} %{arch_sixtyfour})))
--- a/test/blackbox-tests/test-cases/enabled_if/eif-arch_sixtyfour.t/hello.ml
+++ b/test/blackbox-tests/test-cases/enabled_if/eif-arch_sixtyfour.t/hello.ml
@@ -1,0 +1,1 @@
+let () = Printf.printf "Hello, World!\n"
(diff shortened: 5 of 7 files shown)


Alizter:
Dune cannot version check an expression like `(= %{arch_sixtyfour} %{arch_sixtyfour}` and will complain about lists. Therefore in the test I had to reduce to a single variable in the blang so to show the check.

Alizter:
@emillon Please include for 3.9.

test/blackbox-tests/test-cases/enabled_if/eif-arch_sixtyfour.t/run.t
```diff
@@ -0,0 +1,37 @@
...
+  Hello, World!
+
+Testing the version guard
+
+  $ cat > dune-project << EOF
+  > (lang dune 3.8)
+  > EOF
+
+The parser does not understand blang expressions with unknown variables and can
+only guard on single variables.
```
emillon, line 24: I don't understand that part of the test, what does it test?

emillon: Ok. That looks like an issue with the error message, then. What I would suggest is to just use 3.8 with the same test `(= x x)`, check that it does not build (that's the important part), and you can make a note that the error message is not great.

Alizter: Alright, I've reverted to the bad error message and left a note about what it should be instead.

src/dune_rules/enabled_if.ml
```diff
@@ -18,6 +18,7 @@ let common_vars_list =
   ; "profile"
   ; "ocaml_version"
   ; "context_name"
+  ; "arch_sixtyfour"
```
emillon, line 21: I'm not sure how versioning is supposed to work here. For example, `(rule)` uses `~allowed_vars:Any` so this means that it will accept `arch_sixtyfour` even with lang dune 1.4 (this affects `context_name` too). @rgrinberg do you know how this is supposed to work?

test/blackbox-tests/test-cases/enabled_if/eif-arch_sixtyfour.t/run.t
```diff
@@ -0,0 +1,31 @@
...
+  4 |   (= %{arch_sixtyfour} %{arch_sixtyfour})))
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  Error: Unexpected list
+  [1]
+
+The error message is not ideal. It should say something like this instead:
+
+File "dune", line 3, characters 13-30:
+3 |  (enabled_if %{arch_sixtyfour}))
+Error: %{arch_sixtyfour} is only available since version 3.9 of the dune
```
emillon, line 27: I can't reproduce the failure in the error message. It's outputting the right one here (and CI agrees).

Alizter: Weird this is what I get on this branch.

Alizter: 64

src/dune_rules/enabled_if.ml
emillon: that's not correct. it means a dune file can be valid or not depending on the version of the dune binary, which we try to avoid.

Alizter:
I guess for now I will just open an issue for the difference in test output. That will require some parser cleanup that is unrelated to this. I've still got to update the documentation which I will do later.

Alizter:
@emillon I couldn't find any documentation on other allowed fields in enabled_if. I guess I don't need to document anything further?

Alizter:
We need to bump the version check and modify the changelog.

Alizter:
@emillon The changelog and version numbers have been adjusted accordingly.

CHANGES.md
```diff
@@ -9,13 +9,8 @@ Unreleased
...
   stats to a named file. (#8072, @Alizter)
 
-3.9.1 (2023-07-06)
-------------------
-
-- Disable background operations and threaded console on MacOS and other Unixes
-  where we rely on fork. (#8100, #8121, fixes #8083, @rgrinberg, @emillon)
-
-- Initialize async IO thread lazily. (#8122, @emillon)
+- `enabled_if` now supports `arch_sixtyfour` variable (#8023, fixes #7997,
```
rgrinberg, line 12: Needs to be updated

Alizter: I've updated it now, not sure what happened.

Alizter:
@rgrinberg ping

rgrinberg:
@emillon is reviewing this one

Alizter:
@rgrinberg Oh sorry, I was confused, I thought it was you.

Alizter:
@emillon ping

CHANGES.md
```diff
@@ -47,6 +47,9 @@ Unreleased
 - Fix permission errors when `sendfile` is not available (#8234, fixes #8210,
   @emillon)
 
+- `enabled_if` now supports `arch_sixtyfour` variable (#8023, fixes #7997,
```
rgrinberg, line 50: This entry is in the wrong place.

src/dune_rules/enabled_if.ml
```diff
@@ -27,6 +28,7 @@ let common_vars ~since =
        ~f:(fun var ->
          match var with
          | "context_name" -> var, (2, 7)
+         | "arch_sixtyfour" -> var, (3, 10)
```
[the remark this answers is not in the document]
Alizter: Whoops, updated all the version specific stuff now.

PR #7490, opened 2023-04-04
chore: move opam files to opam/ directory

rgrinberg:
Signed-off-by: Rudi Grinberg <me@rgrinberg.com>

PR #8027, opened 2023-06-22
chore: move .opam files into opam/ and bump dune lang

Alizter:
We had to bump the dune lang to 3.8 so I went ahead and bumped it to 3.9. Unless we need a lower bound for any reason?


Alizter:
@gridbugs I had to tweak the docker file for the bench. Could you take a look?

emillon:
If we don't need 3.9, I think we should stick to 3.8. It makes bisecting a bit easier since we can keep dune 3.8.x for the test-deps for a larger range of commits.

rgrinberg:
Isn't this the same as https://github.com/ocaml/dune/pull/7490?

Alizter:
I need some help with the Windows and MSVC CIs. @jonahbeckford could you take a look?

Alizter:
@rgrinberg Oh I completely missed that! I'll close this.

Issue #8021, opened 2023-06-21
Decide on policy for choosing package versions when solving dependencies


gridbugs:
The two competing proposals for this are:
- `prefer_newest`: the latest mutually-compatible versions of dependencies are chosen
- `prefer_oldest`: the oldest mutually-compatible versions of dependencies are chosen (this is sometimes called [MVS](https://go.dev/ref/mod#minimal-version-selection))

The solver in opam currently uses `prefer_newest`. The `opam_0install` package which we use to solve dependencies can be configured to use either policy. At the time of writing this dune is using `prefer_oldest` but this was chosen on a whim (by myself) and without proper discussion. I'm creating this issue so we can properly discus the issue and decide which policy we will use in dune.

In addition to choosing a policy we also need to decide whether to make the policy configurable by users.

## Brief description of high-level trade offs

With `prefer_newest`, packages that don't specify an upper bound on their dependencies will automatically depend on latest versions of their dependencies. This means that package maintainers don't need to explicitly update the dependencies in their package manifests when new versions of deps are released, but it also means that releasing a new version of the package requires testing that the change doesn't break any reverse dependencies of the package and possibly patching or adding upper bounds to packages that would otherwise be broken. On the other hand `prefer_oldest` would avoid situations where an update to a package would break reverse dependencies, but it would require package maintainers to explicitly update their dependencies to track the latest versions.

Some things to consider:
- opam will probably continue to use `prefer_newest`. Will it harm the package ecosystem if dune uses a different policy from opam? We really want to avoid splitting the ecosystem.
- how will this affect the maintenance burden on package maintainers and the maintainers of the opam-repository, and how does the burden scale as the number of packages grows?
- most package manifests are written with `prefer_newest` in mind and so there's little motivation to correctly set lower bounds on their dependencies. This will have to be fixed in order to correctly solve these packages with `prefer_oldest`.

tmattio:
Thanks a lot for opening the issue @gridbugs! To share more on the discussion we had offline, my main concern is that changing from resolving the newest version to the oldest is that we would incentivise developers to add unnecessary lower bounds to their packages (for instance to benefit from a performance improvement that is not strictly necessary). I am worried about the impact this will have on opam-repository: we're currently optimising for having the largest subset of co-installable packages as possible. If developers start adding stricter lower bounds, will we end up making it harder for opam-repository users to find solutions that respect their constraints?

rgrinberg:
Thanks for the summary. It looks broadly correct to me.

[...]

PR #8030, opened 2023-06-22
Prefer newest packages by default

gridbugs:
Based on the discussion at https://github.com/ocaml/dune/issues/8021, dune will prefer the newest versions of packages when solving dependencies. This policy can be configured by a command line argument to `dune pkg lock` and by a field of each context in dune-workspace.

This change includes some formatting changes to the messages printed when solving dependencies which were necessary to handle the fact that different build contexts can now have different package solutions.

### DIFF
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -86,6 +86,32 @@ module Lock = struct
         Repo_selection.local_repo_with_env ~opam_repo_dir_path ~env
   end
 
+  module Version_preference = struct
+    include Dune_pkg.Opam.Version_preference
+
+    let term =
+      let all_strings = List.map all_by_string ~f:fst in
+      let doc =
+        sprintf
+          "Whether to prefer the newest compatible version of a package or the \
+           oldest compatible version of packages while solving dependencies. \
+           This overrides any setting in the current workspace. The default is \
+           %s."
+          (to_string default)
+      in
+      let docv = String.concat ~sep:"|" all_strings |> sprintf "(%s)" in
+      Arg.(
+        value
+        & opt (some (enum all_by_string)) None
+        & info [ "version-preference" ] ~doc ~docv)
+
+    let choose ~from_arg ~from_context =
+      match (from_arg, from_context) with
+      | Some from_arg, _ -> from_arg
+      | None, Some from_context -> from_context
+      | None, None -> default
+  end
+
   (* Converts the package table found inside a [Dune_project.t] into the
      package table expected by the dependency solver *)
   let opam_file_map_of_dune_package_map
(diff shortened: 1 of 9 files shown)


src/dune_pkg/opam.ml
```diff
@@ -329,18 +352,20 @@ end
...
-        (Pp.tag User_message.Style.Success
-           (Pp.text "Selected the following packages:")
-        :: List.map opam_packages_to_lock ~f:(fun package ->
-               Pp.text (OpamPackage.to_string package)))
+      | opam_packages_to_lock ->
+        List.map opam_packages_to_lock ~f:(fun package ->
+            Pp.text (OpamPackage.to_string package))
+    in
+    User_message.make
+      (Pp.textf "Solution for %s:" (Path.Source.to_string lock_dir_path)
```
rgrinberg, line 367: to_maybe_quoted

src/dune_pkg/opam.mli
```diff
@@ -48,10 +48,30 @@ module Summary : sig
   type t
 
   (** A message listing selected packages *)
-  val selected_packages_message : t -> User_message.t
+  val selected_packages_message :
+    t -> lock_dir_path:Path.Source.t -> User_message.t
+end
+
+module Prefer_versions : sig
```
rgrinberg, line 55: I would prefer the name `Version_preference`. Although this is a personal preference that you may ignore

test/blackbox-tests/test-cases/pkg/lock-per-context.t/run.t
```diff
@@ -7,8 +7,14 @@ Create a workspace with multiple contexts, each specifying a lockdir name.
   > (context
   >  (default
   >   (name foo)
+  >   (prefer_versions newest) ; this is the default
   >   (lock bar.lock)))
   > (context
+  >  (default
+  >   (name prefers_oldest)
+  >   (prefer_versions oldest)
```
rgrinberg, line 15: Eventually, we'll want a field to group all solver settings. But this is fine for now.

rgrinberg (approved):

src/dune_pkg/opam.mli
emillon: or give an extra enterprise-ready alternative, `Version_policy`

bin/pkg.ml
```diff
@@ -86,6 +86,32 @@ module Lock = struct
...
   end
 
+  module Prefer_versions = struct
+    include Dune_pkg.Opam.Prefer_versions
+
+    let term =
+      let all_strings = List.map all_by_string ~f:fst in
+      let doc =
+        sprintf
+          "Whether to prefer the newest compatible version of a package or the \
```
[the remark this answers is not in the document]
gridbugs: The difficulty with that in dune is that we always generate lockdirs from scratch so there's no notion of updates. If it's necessary for a project to constrain its transitive dependencies I can imagine adding a field to the build context or project for these, but I don't think `--version-preference` is the right way to specify this.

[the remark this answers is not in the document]
Leonidas-from-XIV: We could have a notion of upgrades, if we look into the currently locked packages (if any) and move these to the front of the priorities to the solver, to create `--version-preference=closest` or something like it. This is what opam-monorepo does for `--minimal-update` when re-locking.

But fair point, maybe `--version-preference` is not the right place to put it.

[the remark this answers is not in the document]
gridbugs: Ah I see. Yeh I suppose one day we could add a `closest` option but I think that would require some careful design. For now let's leave it as is and if we decide to add `closest` we can always rephrase this documentation.

PR #8035, opened 2023-06-24
refactor(engine): optimize sandbox config storage

rgrinberg:
We store a sandbox configuration set for every single action we create.
For a large build, this can add up.

Our old representation was taking 6 words to represent this. This PR
changes it to use only a single word. Moreover, this new set now opaque
to the GC, speeds up comparison, hashing, etc.

### DIFF
--- a/src/dune_engine/sandbox_mode.ml
+++ b/src/dune_engine/sandbox_mode.ml
@@ -30,6 +30,14 @@ let equal a b =
   | Lt | Gt -> false
 ;;
 
+let to_dyn =
+  Dyn.option (function
+    | Symlink -> Variant ("Symlink", [])
+    | Copy -> Variant ("Copy", [])
+    | Hardlink -> Variant ("Hardlink", [])
+    | Patch_back_source_tree -> Variant ("Patch_back_source_tree", []))
+;;
+
 module Dict = struct
   type key = t
 
@@ -70,51 +78,32 @@ module Dict = struct
 end
 
 module Set = struct
-  type key = t
-  type t = bool Dict.t
+  module T = struct
+    type nonrec t = t
 
-  let compare = Dict.compare Bool.compare
-  let of_func = Dict.of_func
-  let singleton k = of_func (equal k)
+    let to_int = function
+      | None -> 0
+      | Some Copy -> 1
+      | Some Symlink -> 2
+      | Some Hardlink -> 3
+      | Some Patch_back_source_tree -> 4
+    ;;
+
+    let all =
+      [ None; Some Copy; Some Symlink; Some Hardlink; Some Patch_back_source_tree ]
+    ;;
+
+    let to_dyn = to_dyn
+  end
+
+  include Bit_set.Make (T)
 
   (* CR-someday amokhov: [Patch_back_source_tree] is a bit special in that it
      can only appear as a singleton. Perhaps, it should be treated differently
      than other sandboxing modes to make meaningless states
      non-representable. *)
   let patch_back_source_tree_only = singleton (Some Patch_back_source_tree)
-
-  let is_patch_back_source_tree_only t =
-    match compare t patch_back_source_tree_only with
-    | Eq -> true
-    | Lt | Gt -> false
-  ;;
-
-  let equal a b =
-    match compare a b with
-    | Eq -> true
-    | Lt | Gt -> false
-  ;;
-
-  let mem = Dict.get
-
-  let inter (x : t) (y : t) : t =
-    { none = x.none && y.none
-    ; copy = x.copy && y.copy
-    ; symlink = x.symlink && y.symlink
-    ; hardlink = x.hardlink && y.hardlink
-    ; patch_back_source_tree = x.patch_back_source_tree && y.patch_back_source_tree
-    }
-  ;;
-
-  let to_dyn { Dict.none; copy; symlink; hardlink; patch_back_source_tree } =
-    Dyn.Record
-      [ "none", Bool none
-      ; "copy", Bool copy
-      ; "symlink", Bool symlink
-      ; "hardlink", Bool hardlink
-      ; "patch_back_source_tree", Bool patch_back_source_tree
-      ]
-  ;;
+  let is_patch_back_source_tree_only t = t = patch_back_source_tree_only
 end
 
 (* The order of sandboxing modes in this list determines the order in which Dune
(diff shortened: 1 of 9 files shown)


src/dune_engine/sandbox_mode.ml
```diff
@@ -69,42 +76,39 @@ end
...
   type key = t
 
-  type t = bool Dict.t
+  type t = int
 
-  let compare = Dict.compare Bool.compare
+  let compare = Int.compare
 
-  let of_func = Dict.of_func
+  let singleton k = 1 lsl to_int k
```
snowleopard, line 83: I wonder if we could factor out the bit-twiddling from this module into a new Stdune module, say `Bit_dict`. Otherwise, this refactoring makes this module pretty hard to understand.

rgrinberg: Ah yes, will give that a try.

rgrinberg:
@snowleopard should be ready now

otherlibs/stdune/src/bit_set.ml
```diff
@@ -0,0 +1,41 @@
+module Common = struct
```
snowleopard, line 1: It's not clear to me what principle was used to factor out these functions into a separate module. For example, why is `let singleton x = 1 lsl Element.to_int x` not part of it? Should it be?

snowleopard: (Same question for `type t = int`.)

rgrinberg: The principle is that all of those implementations can be shared between all instantiations of the functor. In other words, they're independent of the representation of the elements. It's a small optimization that probably doesn't matter.

src/dune_engine/sandbox_mode.ml
```diff
@@ -70,51 +78,43 @@ module Dict = struct
...
+
+  let of_func =
+    let ( ++ ) = union in
+    fun f ->
+      let e x = if f x then singleton x else empty in
+      e None
+      ++ e (Some Copy)
+      ++ e (Some Symlink)
+      ++ e (Some Hardlink)
+      ++ e (Some Patch_back_source_tree)
```
snowleopard, line 109: I think there should be `Bit_set.of_func`, as this code seems fiddly enough to be hidden behind the API.

src/dune_engine/sandbox_mode.ml
```diff
@@ -70,51 +78,43 @@ module Dict = struct
...
-  type t = bool Dict.t
-
-  let compare = Dict.compare Bool.compare
-  let of_func = Dict.of_func
-  let singleton k = of_func (equal k)
+  module T = struct
+    type nonrec t = t
+
+    let to_int = function
+      | None -> 1
```
snowleopard, line 85: Can we `Make` a `Bit_set` for `t option` as well? (Just thinking about how we can move as much code as possible into `Bit_set`.)

snowleopard (commented):
Looks good but I left some suggestions/questions.

Also: have you run any benchmarks?

otherlibs/stdune/src/bit_set.ml
snowleopard: I see. I'm thinking that a part of `singleton` (that does `1 lsl`) can still be moved into `Common` though maybe that doesn't matter. Maybe add a comment explaining your principle so that others aren't wondering in the same way as me.

src/dune_engine/sandbox_mode.ml
rgrinberg: Sure, although note that the implementation will not be as good. As we'll need to fold over the list of elements instead of just inlining everything.

snowleopard: I think you could do `Array.iteri` over a precomputed array of all elements but perhaps that's still less efficient? In any case, it seems to me that this is unlikely to matter since creating a rule involves a lot more work than traversing a little array.

src/dune_engine/sandbox_mode.ml
rgrinberg: Do you mean make a functor like `Bit_set.Make` that works for `t option` rather `t`?

src/dune_engine/sandbox_mode.ml
rgrinberg: Sure. Sounds good.

otherlibs/stdune/src/bit_set.ml
rgrinberg: > I'm thinking that a part of singleton (that does 1 lsl) can still be moved into Common though maybe that doesn't matter.

Right, but that function depends on the representation of elements to be used correctly. So it's not too useful.

> Maybe add a comment explaining your principle so that others aren't wondering in the same way as me.

Will do

src/dune_engine/sandbox_mode.ml
snowleopard: I thought that maybe there should be two functors: one for `t` and another one for `t option`. But maybe I'm overcomplicating things. Feel free to ignore this comment if you'd rather keep the code as is.

rgrinberg: It can work, with a signature that looks something like this:
```
module type Element = sig
  type t

  (** [to_int t] must return a unique number for every [t] between [1] and
      [Sys.int_size - 1] (inclusive) *)
  val to_int : t -> int

  (** [all] contains all possible set elements *)
  val all : t list

  val to_dyn : t -> Dyn.t
end

module Option (T : Element) : Element with type t = T.t option

module Make (Element : Element) : sig ...
end
```

But it seems a bit overkill if you aren't using this anywhere else yet.

snowleopard: OK, agreed, let's stick to the existing `Make`.

rgrinberg:
> Also: have you run any benchmarks?

Right now the benchmarking server is broken. Once it's back up, I'll bring some numbers.

Alizter:
@rgrinberg Benchmarking is working again, can you resurrect this?

rgrinberg:
It does reduce major and promoted words by around 0.3-0.4%. Not sure if it's worth it, I'll let @snowleopard decide.

We did end up using bit_set for something else, so I will merge that at least.

Alizter:
@rgrinberg Test needs updating too.

rgrinberg:
ping @snowleopard

test/blackbox-tests/test-cases/patch-back-source-tree.t
```diff
@@ -201,7 +201,7 @@ produced in the sandbox and copied back:
 This is the internal stamp file:
 
   $ ls _build/.actions/default/blah*
-  _build/.actions/default/blah-e7a0efae1209023d2186a50341cd25fa
+  _build/.actions/default/blah-61c2a19beb7c9447302b9348604599d6
```
snowleopard, line 204: Doesn't this mean you need to bump some digest version somewhere?

snowleopard (approved):
Looks good though see one question. I haven't had time to run any benchmarks internally but it should be a small win.

PR #8037, opened 2023-06-24
dune_util: record exceptions to the _build/log

Alizter:
There are times when the detailed error becomes unavailable to view in the terminal such as when working with tui. Therefore recording them to the _build/log can help record what went wrong.


Alizter:
Not really sure what to do with this. Drafting.

rgrinberg:
This is OK. The issue is that errors raised in the console end up going to Report_error and back to the console. We need to be careful and just close the console once it fails.

Issue #8043, opened 2023-06-24
Stalls for 3.8.2 in dune build on Windows GitLab CI


jonahbeckford:
## Behavior

I do a `dune build` (exact verbose logs are at https://gist.github.com/jonahbeckford/c33118fd45812e2bf73d5f4c4266710a).

I get a stall (deadlock?) under certain conditions:

* Only Dune 3.8.2. (Versions prior to 3.8.2 work! Dune 3.6.2, 3.7.1, 3.8.0 and 3.8.1)
* Only GitLab CI machines. (Not my desktop, and apparently not on GitHub Actions)

~~Because of the above conditions, I strongly suspect the root cause is https://github.com/ocaml/dune/pull/7906 ("Use threaded console, adapt for Windows").~~ *EDIT: Tested that this wasn't the commit.*

## Reproduction

A `dune build` of many packages (quite similar to a `opam monorepo` build).

These packages:

* `lwt.5.6.1`
* `stdint.0.7.2`
* `memtrace.v0.2.3`
* `fiber.3.7.0`

and the following:

```
  astring.0.8.5+dune ;                               d950ab785805c6fb51bc48dff6380118
  base-bytes.base+dune ;                             607827e71b1a816b33e16104d966f39c
  base.v0.15.1 ;                                     f3d7f4e72096610b2f14017c351df010
  camlp-streams.5.0.1 ;                              179a0662333b2ac5c7a3cc5009e1020f
  capnp.3.5.0 ;                                      c79f2454d7d9d956f0f45f25e3c25fb2
  cppo.1.6.9 ;                                       64b7f30febf9b6c4977f537ea936992d
  csexp.1.5.2 ;                                      69b9a4107a979db5724ccdb0b1f3fbf8
  cstruct.6.2.0 ;                                    1c750a2ed4098f82b65153917b9ad475
  dum.1.0.3 ;                                        21bd91e73b7e9bccb298854f1bf47e58
  dyn.3.8.0 ;                                        e7bdb1e6bcf1cdc2c004f72faa1ecf52
  easy-format.1.3.4 ;                                e77d2a3c8294190d59006af468ef9f09
  ezjsonm.1.3.0 ;                                    1b532e39182b85e4a787f270fccf2fb4
  ; fiber.3.7.0 ;                                    excluded
  fmt.0.9.0+dune ;                                   c6bc7ec23eb7c17757f5bbf5921cbe7a
[...]

PR #8044, opened 2023-06-25
async_io: adapt self-pipe trick to Windows

nojb:
The Dune RPC server uses the "self-pipe trick" to interrupt `Unix.select`. Unfortunately, on Windows, pipes cannot be passed to the native `select` system call, which means that `Unix.select` falls back to a complicated emulation using threads to poll for data on the pipe. This, plus the fact that pipes are always blocking, seem to be causing some deadlock somewhere which makes the RPC server hang under certain circumstances.

This PR proposes an alternative form of the self-pipe trick using a self-connected UDP socket. The advantages is that no pipes are involved, so only the native `select` call is used by `Unix.select`, completely bypassing all the complexity of the emulation and the associated blocking pipes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -62,6 +62,8 @@ Unreleased
 - Add necessary parentheses in generated opam constraints (#7682, fixes #3431,
   @Lucccyo)
 
+- Fix deadlock on Windows (#8044, @nojb)
+
 3.8.2 (2023-06-16)
 ------------------
 
--- a/src/dune_async_io/async_io.ml
+++ b/src/dune_async_io/async_io.ml
@@ -153,7 +153,8 @@ let rec select_loop t =
   match t.running with
   | false ->
     Unix.close t.pipe_write;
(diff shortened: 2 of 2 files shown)


rgrinberg (approved):
Awesome work. I imagine it's worth reporting that select on windows pipes fails for this case?

Needs a CHANGES entry.

nojb:
> Awesome work. I imagine it's worth reporting that select on windows pipes fails for this case?

My understanding is that it is not that it doesn't work, but rather that the semantics of pipes on Windows do not match exactly those of Linux and the emulation in `Unix.select` does not help things, in the sense that there may be a way to make the existing approach work (probably involving further synchronization), but it becomes more and more complex and it becomes hard to know what the right solution is.

> Needs a CHANGES entry.

Thanks, will add. I also want to wait to see if by chance this PR also fixes #8043.

nojb:
@emillon if there is a bug fix release 3.8.3 this should be in it.

PR #8045, opened 2023-06-26
refactor(tui): move initialization to UI thread

rgrinberg:


Alizter:
How does making the setup lazy ensure that the UI thread will pick it up when it is evaluated?

rgrinberg:
All of those thunks will only be forced in the UI thread.

Alizter (commented):
OK, I think I get what is happening now. Another question was about the removal of the thunk on the Tui module, does this affect anything:

Alizter (approved):

rgrinberg:
The generative functor argument? No, it shouldn't affect anything because there are no longer any module side effects.

PR #8050, opened 2023-06-26
Conditional dependencies in lockdir

gridbugs:
This changes lockdir generation so that dependencies which are conditional on the variables `:with-test` and `:with-doc` are added to the lockdir.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -10,26 +10,57 @@ module Context_either (L : CONTEXT) (R : CONTEXT) :
 
   type rejection = (L.rejection, R.rejection) Either.t
 
-  let pp_rejection f = function
-    | Left l -> L.pp_rejection f l
-    | Right r -> R.pp_rejection f r
+  let pp_rejection f = Either.map ~l:(L.pp_rejection f) ~r:(R.pp_rejection f)
 
-  let candidates t name =
+  let candidates =
     let convert_rejections ~f =
       List.map ~f:(fun (version, result) ->
           (version, Result.map_error ~f result))
     in
-    match t with
-    | Left l -> L.candidates l name |> convert_rejections ~f:Either.left
-    | Right r -> R.candidates r name |> convert_rejections ~f:Either.right
+    Either.map
+      ~l:(fun l name ->
+        L.candidates l name |> convert_rejections ~f:Either.left)
+      ~r:(fun r name ->
+        R.candidates r name |> convert_rejections ~f:Either.right)
 
-  let user_restrictions = function
-    | Left l -> L.user_restrictions l
-    | Right r -> R.user_restrictions r
+  let user_restrictions =
+    Either.map ~l:L.user_restrictions ~r:R.user_restrictions
 
-  let filter_deps = function
-    | Left l -> L.filter_deps l
-    | Right r -> R.filter_deps r
+  let filter_deps = Either.map ~l:L.filter_deps ~r:R.filter_deps
+end
+
+(* Helper module for working with [OpamTypes.filtered_formula] *)
+module Filtered_formula : sig
+  open OpamTypes
+
+  (** Update a filtered formula setting the "with-test" and "with-doc" variables
+      to true. *)
+  val with_test_and_doc : filtered_formula -> filtered_formula
+end = struct
+  open OpamTypes
+
+  (* [map_formula_filters_bottom_up t ~f] calls [f] on each node of each filter
+     in t, bottom-up *)
+  let map_filters_bottom_up ~f =
+    OpamFilter.gen_filter_formula
+      (OpamFormula.partial_eval (function
+        | Filter flt -> `Formula (Atom (Filter (OpamFilter.map_up f flt)))
+        | Constraint _ as constraint_ -> `Formula (Atom constraint_)))
+
+  (* update a filtered formula replacing instances of a given variable with a given boolean value *)
+  let resolve_simple_bool_variable filtered_formula variable_to_resolve value =
+    map_filters_bottom_up filtered_formula ~f:(function
+      | FIdent ([], variable, None) as filter ->
+        if OpamVariable.equal variable variable_to_resolve then FBool value
+        else filter
+      | other -> other)
+
+  let with_test_and_doc filtered_formula =
+    List.fold_left [ "with-test"; "with-doc" ] ~init:filtered_formula
+      ~f:(fun acc variable_name ->
+        resolve_simple_bool_variable acc
+          (OpamVariable.of_string variable_name)
+          true)
 end
 
 (* Helper functor which adds a set of local packages to a [CONTEXT] *)
@@ -65,7 +96,17 @@ end = struct
 
   let user_restrictions t = Base_context.user_restrictions t.base_context
 
-  let filter_deps t = Base_context.filter_deps t.base_context
+  (* Override [filter_deps] so that local packages get test and doc dependencies *)
+  let filter_deps t package filtered_formula =
+    let package_is_local =
+      OpamPackage.Name.Map.mem (OpamPackage.name package) t.local_packages
+    in
+    let filtered_formula =
+      if package_is_local then
+        Filtered_formula.with_test_and_doc filtered_formula
+      else filtered_formula
+    in
+    Base_context.filter_deps t.base_context package filtered_formula
 end
 
 module Context_for_dune = struct
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -35,7 +35,7 @@ module Filtered_formula : sig
 
   (** Update a filtered formula setting the "with-test" and "with-doc" variables
       to true. *)
-  val with_test_and_doc : filtered_formula -> filtered_formula
+  val with_flags : Solver_env.Flag.Set.t -> filtered_formula -> filtered_formula
 end = struct
   open OpamTypes
 
@@ -55,11 +55,10 @@ end = struct
         else filter
       | other -> other)
 
-  let with_test_and_doc filtered_formula =
-    List.fold_left [ "with-test"; "with-doc" ] ~init:filtered_formula
-      ~f:(fun acc variable_name ->
+  let with_flags flags filtered_formula =
+    Solver_env.Flag.Set.fold flags ~init:filtered_formula ~f:(fun flag acc ->
         resolve_simple_bool_variable acc
-          (OpamVariable.of_string variable_name)
+          (OpamVariable.of_string (Solver_env.Flag.to_string flag))
           true)
 end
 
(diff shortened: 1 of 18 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -0,0 +1,190 @@
...
+          (Dir_context.create ~prefer_oldest
+             ~constraints:OpamPackage.Name.Map.empty ~env packages_dir_path)
+        ~local_packages
+    in
+    { base_context; env }
+
+  (* Override [filter_deps] so that local packages get test and doc dependencies *)
+  let filter_deps t =
+    let local_packages = Base_context.local_packages t.base_context in
+    filter_deps_env ~local_packages ~env:(fun pkg v ->
```
Leonidas-from-XIV, line 109: Isn't this the env function from the solver? Maybe it can be reused?

src/dune_pkg/opam_solver.ml
```diff
@@ -0,0 +1,190 @@
...
+        if
+          List.mem OpamPackageVar.predefined_depends_variables v
+            ~equal:OpamVariable.Full.equal
+        then None
+        else
+          let r = OpamPackageVar.resolve_switch ~package:pkg t.switch_state v in
+          if r = None then
+            OpamConsole.warning "Unknown variable %S"
```
Leonidas-from-XIV, line 163: It should probably not warn to the Opam console.

Leonidas-from-XIV (commented):
Looks good and a very useful change. Some nitpicking on the code from the 0install solver.

src/dune_pkg/opam_solver.ml
```diff
@@ -0,0 +1,190 @@
...
+
+(* Helper function for implementing [CONTEXT.filter_deps] where local package
+   dependencies include test and doc dependencies *)
+let filter_deps_env ~local_packages ~env pkg filtered_formula =
+  let pkg_is_local =
+    OpamPackage.Name.Map.mem (OpamPackage.name pkg) local_packages
+  in
+  filtered_formula
+  |> OpamFilter.partial_filter_formula (env pkg)
+  |> OpamFilter.filter_deps ~build:true ~post:true ~test:pkg_is_local
```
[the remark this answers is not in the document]
gridbugs: It means that dependencies marked as "post" will be resolved. I've never encountered this variable in the wild but it tells opam that the dependency can be compiled after the current package. https://opam.ocaml.org/doc/Manual.html#pkgvar-post I set it to true because that's what `Opam_0install.Dir_context` does and it seems like the most useful default setting. Perhaps one day we'll need to add support to dune to set it to false to deal with dependency cycles.

src/dune_pkg/opam_solver.ml
gridbugs: Actually yes, I've since dug a bit deeper into opam filters in another change I'm working on and I think we can simplify this by resolving the `test` and `doc` variables to true before calling the base context's `filter_deps` on the result. Will update.

src/dune_pkg/opam_solver.ml
gridbugs: Oops yes you're right.

src/dune_pkg/opam_solver.ml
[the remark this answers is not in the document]
Leonidas-from-XIV: Thanks for the explanation, I have never encountered these before. Out of curiosity I checked opam-repository and there are two non-compiler packages (only 2 versions overall) that use these: `sail.0.15` and `ctypes-foreign.0.18.0` so indeed these don't seem to be used much.

gridbugs:
I removed all the code borrowed/modified from opam_0install and instead added a module `Filtered_formula` with functions for transforming opam formulas and now `filter_deps` just uses that to replace `with-test` and `with-doc` with `true` before calling the base context's `filter_deps` on the result. This also means we could get rid of some of the `*_for_dune` modules previously in this change.

Also in the process I noticed that we were setting the `with-dev-setup` variable which I don't think we want so now we just set the `with-test` and `with-doc` variables to true.

src/dune_pkg/opam_solver.ml
```diff
@@ -0,0 +1,131 @@
...
+  (* [map_formula_filters_bottom_up t ~f] calls [f] on each node of each filter
+     in t, bottom-up *)
+  let map_filters_bottom_up ~f =
+    OpamFilter.gen_filter_formula
+      (OpamFormula.partial_eval (function
+        | Filter flt -> `Formula (Atom (Filter (OpamFilter.map_up f flt)))
+        | Constraint _ as constraint_ -> `Formula (Atom constraint_)))
+
+  (* update a filtered formula replacing instances of a given variable with a given boolean value *)
+  let resolve_simple_bool_variable ~variable_name ~value =
```
Leonidas-from-XIV, line 51: Maybe `variable_name` would be better as an `OpamVariable`, then one could use its `equal` function, thus making the named arguments less important.

src/dune_pkg/opam_solver.ml
```diff
@@ -0,0 +1,131 @@
...
+  (** Update a filtered formula setting the "with-test" and "with-doc" variables
+      to true. *)
+  val with_test_and_doc : filtered_formula -> filtered_formula
+end = struct
+  open OpamTypes
+
+  (* [map_formula_filters_bottom_up t ~f] calls [f] on each node of each filter
+     in t, bottom-up *)
+  let map_filters_bottom_up ~f =
+    OpamFilter.gen_filter_formula
```
Leonidas-from-XIV, line 45: Cool, didn't know these functions. `map_filters_bottom_up` seems quite useful.

gridbugs: Yeh `OpamFilter.map_up` is very handy. I just learnt about it recently too.

rgrinberg:
A couple of things that are missing from this PR:

* The ability to customize this behavior. In particular, it should be possible not to have docs or test deps. Also, it should be possible to include one of the two.
* The second thing we're missing is a way to propagate which deps were added because of the deps or tests flags. This is so that installing the software can use the same lock file (when applicable).

The first one needs to be done in this PR, while the second one can be done in subsequent work.

gridbugs:
>     * The second thing we're missing is a way to propagate which deps were added because of the deps or tests flags. This is so that installing the software can use the same lock file (when applicable).

Can you clarify this point. Do you mean including in package metadata whether the package was included in the lockdir because it's a test/doc dependency so that we can omit those packages when using the lockdir for installation?

gridbugs:
Cool makes sense. I'll update this PR so that `with-test` and `with-doc` can be configured per build context and make a new PR which tracks which packages were included due to these flags in a later change.

gridbugs:
@rgrinberg I've added a `Solver_env.t` type which stores variables used by the solver. Currently it just stores the flags `:with-test` and `:with-doc` but I think it would make sense to use this for all opam variables (e.g. system variables) and replace the `Opam.Env.t` type completely. I'm experimenting with that on another branch but I won't add it to this PR so it doesn't get too complicated.

src/dune_pkg/opam_solver.ml
```diff
@@ -0,0 +1,136 @@
...
+  let resolve_simple_bool_variable filtered_formula variable_to_resolve value =
+    map_filters_bottom_up filtered_formula ~f:(function
+      | FIdent ([], variable, None) as filter ->
+        if OpamVariable.equal variable variable_to_resolve then FBool value
+        else filter
+      | other -> other)
+
+  let with_flags flags filtered_formula =
+    Solver_env.Flag.Set.to_list flags
+    |> List.fold_left ~init:filtered_formula ~f:(fun acc flag ->
```
rgrinberg, line 60: You could just add a fold function to `Solver_env.Flag.Set`. It doesn't really matter much, but why favor list conversion when all you need is a fold?

src/dune_pkg/opam_solver.mli
```diff
@@ -0,0 +1,20 @@
...
+    -> packages_dir_path:Filename.t
+    -> local_packages:OpamFile.OPAM.t OpamTypes.name_map
+    -> prefer_oldest:bool
+    -> t
+
+  val create_switch_context :
+       solver_env:Solver_env.t
+    -> switch_state:OpamStateTypes.unlocked OpamStateTypes.switch_state
+    -> local_packages:OpamFile.OPAM.t OpamTypes.name_map
+    -> prefer_oldest:bool
```
rgrinberg, line 18: Don't we have a version preference type somewhere that we can use?

src/dune_pkg/opam_solver.mli
```diff
@@ -0,0 +1,20 @@
+open Stdune
+
+module Context_for_dune : sig
+  include Opam_0install.S.CONTEXT
+
+  val create_dir_context :
+       solver_env:Solver_env.t
+    -> env:(string -> OpamVariable.variable_contents option)
```
rgrinberg, line 8: Not entirely related, but I would prefer this naked `string` type turned into something abstract.

rgrinberg (approved):

src/dune_pkg/opam_solver.mli
gridbugs: I'm planning to remove this argument soon when I move all environment variables inside `Solver_env.t` so I'll leave it for now.

gridbugs:
@rgrinberg does anything more need to happen before this can be merged?

rgrinberg:
Ready from me. Waiting on @Leonidas-from-XIV to review

test/blackbox-tests/test-cases/pkg/solver-flags-per-context.t
```diff
@@ -0,0 +1,102 @@
...
+  >   (name with-doc-only)
+  >   (lock with-doc-only.lock)
+  >   (solver_env
+  >    (flags with-doc))))
+  > (context
+  >  (default
+  >   (name with-doc-and-with-test)
+  >   (lock with-doc-and-with-test.lock)
+  >   (solver_env
+  >    (flags with-doc with-test))))
```
Leonidas-from-XIV, line 52: I wonder whether it would make sense to implement the set language so `(flags :standard)` can be specified.

Leonidas-from-XIV (approved):
I think it looks quite good, some minor comments when I looked over the diff.

test/blackbox-tests/test-cases/pkg/solver-flags-per-context.t
gridbugs: Ah yes I like the idea of allowing `:standard` here. It won't be hard to add as long as we don't allow `:include` which I think is reasonable.

gridbugs: Updated to use the ordered set language for this field and added a test using the `:standard` keyword.

gridbugs:
> Yes, that's what I meant. Otherwise, we'll have to always build all the test deps.

@rgrinberg on the topic of storing additional metadata to keep track of which dependencies are test/doc-only so we don't build test/doc deps when installing, I thought about this some more and I don't think this is necessary. The solver flags apply only to dependencies of local packages define in `dune-project` and the dependency specification in `dune-project` already stores whether a package is a test or doc dependency.

PR #8052, opened 2023-06-27
refactor(stdune): improve path tables

rgrinberg:
Use a more memory efficient path table. Instead of using the variant for
the key, combine 3 tables all for the individual paths.

This makes the empty table a little more bloated (3x bigger), but gives
us a saving of 2 words for every single key we store.

### DIFF
--- a/otherlibs/stdune/src/path.ml
+++ b/otherlibs/stdune/src/path.ml
@@ -290,6 +290,8 @@ module Local : sig
 
   include Path_intf.S with type t := t
 
+  module Table : Hashtbl.S with type key = t
+
   val root : t
 
   val is_root : t -> bool
@@ -368,6 +370,8 @@ end
 
 module External : sig
   include Path_intf.S
+
+  module Table : Hashtbl.S with type key = t
 
   val relative : t -> string -> t
 
@@ -741,8 +745,6 @@ module T : sig
       Path of the same kind are compared using the standard lexical order *)
   val compare : t -> t -> Ordering.t
 
-  val equal : t -> t -> bool
-
   val hash : t -> int
 
   val in_build_dir : Local.t -> t
@@ -766,8 +768,6 @@ end = struct
     | In_build_dir _, In_source_tree _ -> Gt
     | In_build_dir x, In_build_dir y -> Local.compare x y
 
-  let equal (x : t) (y : t) = x = y
-
   let hash = Poly.hash
 
   let in_build_dir s = In_build_dir s
@@ -783,8 +783,6 @@ end = struct
 end
 
 include T
-
-let hash (t : t) = Poly.hash t
 
 let build_dir = in_build_dir Local.root
 
@@ -1167,16 +1165,6 @@ let touch ?(create = true) p =
     (* OCaml PR#8857 *)
     create ()
 
-let compare x y =
-  match (x, y) with
-  | External x, External y -> External.compare x y
-  | External _, _ -> Lt
-  | _, External _ -> Gt
-  | In_source_tree x, In_source_tree y -> Local.compare x y
-  | In_source_tree _, _ -> Lt
-  | _, In_source_tree _ -> Gt
-  | In_build_dir x, In_build_dir y -> Local.compare x y
-
 let extension t =
   match t with
   | External t -> External.extension t
@@ -1220,7 +1208,77 @@ let source s = in_source_tree s
 
 let build s = in_build_dir s
 
-module Table = Hashtbl.Make (T)
+module Table = struct
+  type key = t
+
+  type 'a t =
+    { source : 'a Source0.Table.t
+    ; build : 'a Build.Table.t
+    ; external_ : 'a External.Table.t
+    }
+
+  let create () =
+    { source = Source0.Table.create 0
+    ; build = Build.Table.create 0
+    ; external_ = External.Table.create 0
+    }
+
+  let clear { source; build; external_ } =
+    Source0.Table.clear source;
+    Build.Table.clear build;
+    External.Table.clear external_
+
+  let[@inline] mem { source; build; external_ } key =
+    match key with
+    | In_source_tree p -> Source0.Table.mem source p
+    | In_build_dir p -> Build.Table.mem build p
+    | External p -> External.Table.mem external_ p
+
+  let[@inline] set { source; build; external_ } k v =
+    match k with
+    | In_source_tree p -> Source0.Table.set source p v
+    | In_build_dir p -> Build.Table.set build p v
+    | External p -> External.Table.set external_ p v
+
+  let[@inline] remove { source; build; external_ } k =
+    match k with
+    | In_source_tree p -> Source0.Table.remove source p
+    | In_build_dir p -> Build.Table.remove build p
+    | External p -> External.Table.remove external_ p
+
+  let iter { source; build; external_ } ~f =
+    Source0.Table.iter source ~f;
+    Build.Table.iter build ~f;
+    External.Table.iter external_ ~f
+
+  let[@inline] find { source; build; external_ } = function
+    | In_source_tree p -> Source0.Table.find source p
+    | In_build_dir p -> Build.Table.find build p
+    | External p -> External.Table.find external_ p
+
+  let filteri_inplace { source; build; external_ } ~f =
+    Source0.Table.filteri_inplace source ~f:(fun [@inline] ~key ~data ->
+        f ~key:(In_source_tree key) ~data);
+    Build.Table.filteri_inplace build ~f:(fun [@inline] ~key ~data ->
+        f ~key:(In_build_dir key) ~data);
+    External.Table.filteri_inplace external_ ~f:(fun [@inline] ~key ~data ->
+        f ~key:(External key) ~data)
+
+  let filter_inplace { source; build; external_ } ~f =
+    Source0.Table.filteri_inplace source ~f:(fun [@inline] ~key:_ ~data ->
+        f data);
+    Build.Table.filteri_inplace build ~f:(fun [@inline] ~key:_ ~data -> f data);
+    External.Table.filteri_inplace external_ ~f:(fun [@inline] ~key:_ ~data ->
+        f data)
+
+  let to_dyn f { source; build; external_ } =
+    let open Dyn in
+    record
+      [ ("source", Source0.Table.to_dyn f source)
+      ; ("build", Build.Table.to_dyn f build)
+      ; ("external_", External.Table.to_dyn f external_)
+      ]
+end
 
 module L = struct
   (* TODO more efficient implementation *)
--- a/otherlibs/stdune/src/path.mli
+++ b/otherlibs/stdune/src/path.mli
@@ -84,6 +84,8 @@ module External : sig
   val of_filename_relative_to_initial_cwd : string -> t
 
   val append_local : t -> Local.t -> t
+
+  module Table : Hashtbl.S with type key = t
 end
 
 (** In the source section of the current workspace. *)
@@ -113,6 +115,8 @@ module Source : sig
   val descendant : t -> of_:t -> t option
 
   val to_local : t -> Local.t
+
+  module Table : Hashtbl.S with type key = t
 end
 
 module Permissions : sig
@@ -221,6 +225,8 @@ module Build : sig
   val lstat : t -> Unix.stats
 
   val unlink_no_err : t -> unit
+
+  module Table : Hashtbl.S with type key = t
 end
 
 type t = private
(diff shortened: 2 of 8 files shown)


snowleopard (commented):
Looks good! I am going to benchmark this change internally too.

otherlibs/stdune/src/path.mli
```diff
@@ -230,6 +236,32 @@ type t = private
 
 include Path_intf.S with type t := t
 
+module Table : sig
```
snowleopard, line 239: Why couldn't this be `module Table : Hashtbl.S with type key = path`? (Whatever the answer is, I think it's worth documenting it as a comment.)

snowleopard:
What about `Path.Set` and `Path.Map`? They are probably even bigger offenders (because we have more of them).

otherlibs/stdune/src/path.ml
```diff
@@ -1220,7 +1208,68 @@ let source s = in_source_tree s
...
+    External.Table.iter external_ ~f
+
+  let find { source; build; external_ } = function
+    | In_source_tree p -> Source0.Table.find source p
+    | In_build_dir p -> Build.Table.find build p
+    | External p -> External.Table.find external_ p
+
+  let filteri_inplace { source; build; external_ } ~f =
+    Source0.Table.filteri_inplace source ~f:(fun ~key ~data ->
+        f ~key:(In_source_tree key) ~data);
```
snowleopard, line 1259: This is likely worse than before.

snowleopard: Looks like there is just one call site (`src/dune_engine/cached_digest.ml`). Maybe we can change it to avoid the possibility of regression.

otherlibs/stdune/src/path.ml
```diff
@@ -1220,7 +1208,68 @@ let source s = in_source_tree s
...
-module Table = Hashtbl.Make (T)
+module Table = struct
+  type 'a t =
+    { source : 'a Source0.Table.t
+    ; build : 'a Build.Table.t
+    ; external_ : 'a External.Table.t
+    }
+
+  let create size =
+    { source = Source0.Table.create size
```
snowleopard, line 1219: What's the initial size of Dune tables? Shall we do something here to avoid grabbing 3x size now?

otherlibs/stdune/src/path.mli
rgrinberg: Will add a comment. The reason is that it would require implementing more functions than we actually use. Not a great concern, but it does more code that is effectively dead and might make us rely on more properties of a regular hash tables than might want to.

rgrinberg:
> What about Path.Set and Path.Map? They are probably even bigger offenders (because we have more of them).

I have an upcoming PR indeed.

otherlibs/stdune/src/path.ml
rgrinberg: We could, but I actually think it doesn't matter. We only use hash tables for global caches and we really underestimate the sizes of those. I could divide each table size by 3 if you prefer though

otherlibs/stdune/src/path.ml
rgrinberg: Indeed it's a little worse. I'll tweak the API to pass an iteration function for every type of key. That should make it the same as before.

otherlibs/stdune/src/path.mli
snowleopard: Right, agreed, it seems best not to maintain dead code.

otherlibs/stdune/src/path.ml
rgrinberg: Done. I've fixed the regression and made it even faster by doing the debug check only once.

otherlibs/stdune/src/path.ml
snowleopard: Or, maybe, use size 0 or make these tables lazy? I'm just observing some not very consistent benchmark results and wonder if this could be an explanation. Like, what if the majority of tables we have actually use just one of the path types?

src/dune_engine/cached_digest.ml
```diff
@@ -94,19 +94,26 @@ let delete_very_recent_entries () =
...
+      match Float.compare data.stats.mtime now with
+      | Lt -> true
+      | Gt | Eq -> false
+    in
+    match !Clflags.debug_digests with
+    | false -> Path.Table.filter_inplace cache.table ~f:filter
+    | true ->
+      Path.Table.filteri_inplace cache.table ~f:(fun ~key:path ~data ->
+          let filter = filter data in
+          if filter then
```
snowleopard, line 108: Did you mean `not filter`?

I'm not quite sure why we're actually doing this filtering. Do you remember?

rgrinberg: > Did you mean not filter?

Yes indeed. Thanks.

> I'm not quite sure why we're actually doing this filtering. Do you remember?

Serializing and deserializing aren't cheap and hopefully this will shrink the table before do dump the table on disk.

otherlibs/stdune/src/path.ml
rgrinberg: > I'm just observing some not very consistent benchmark results and wonder if this could be an explanation. 

Wait, I just realized that we barely use this table at all in memo. Instead, we instantiate the slow tables when we create the store 😄 

So this change does very very little in fact :)

> Like, what if the majority of tables we have actually use just one of the path types?

Don't think it would matter. The other tables would stay near their (tiny) initial sizes and the empty space in the one large table would easily dwarf the empty space in the other two.

src/dune_engine/cached_digest.ml
snowleopard: > Serializing and deserializing aren't cheap and hopefully this will shrink the table before do dump the table on disk.

I don't get it. Why are fresh digests less valuable than old ones?

otherlibs/stdune/src/path.ml
snowleopard: > The other tables would stay near their (tiny) initial sizes

Wouldn't it be better for them to stay the tiniest possible size of 0 rather than whatever `Table.t` happens to pick?

snowleopard: > Wait, I just realized that we barely use this table at all in memo. Instead, we instantiate the slow tables when we create the store

Hmm, I guess we'd like to fix that? Not sure in which order (or together)?

src/dune_engine/cached_digest.ml
rgrinberg: You're right, I have no idea. I looked through git blame and it seems like Arseniy introduced this. Might be worth asking him.

otherlibs/stdune/src/path.ml
rgrinberg: > Wouldn't it be better for them to stay the tiniest possible size of 0 rather than whatever Table.t happens to pick?

If that's the case, the caller can just pass 0 already. Perhaps we should just do that and get rid of this size argument? I don't particularly see a need for it anyway.

> Hmm, I guess we'd like to fix that? Not sure in which order (or together)?

If you confirm there's no regression from this PR, I think it's fine doing it after.

src/dune_engine/cached_digest.ml
snowleopard: Ack, I'll ask him tomorrow.

otherlibs/stdune/src/path.ml
snowleopard: > If you confirm there's no regression from this PR, I think it's fine doing it after.

The benchmarks I run are kind of all over the place memory-wise. And, weirdly, there seem to be a consistent small (0.3-1.0%) wall clock build time regression. I can try to rerun it tomorrow but it feels like something is wrong.

rgrinberg: I see. Well that's quite strange because the benchmark here shows a tiny improvement (within the noise though).

What I can also try is to add inlining annotations. I suppose it can hurt performance if there's an extra function call.

rgrinberg: It seemed to have make a difference for null builds (our only stable measurements for such small differences).

snowleopard: Nice, I'll rerun the benchmarks.

src/dune_engine/cached_digest.ml
snowleopard: So, here is the reason: we can't trust cached digest where the timestamp of the digest coincides with the current time (according to file-system's resolution granularity), so, it's important to filter out such measurements (and it was a concern on some file systems with bad granularity in the past).

src/dune_engine/cached_digest.ml
```diff
@@ -94,19 +94,25 @@ let delete_very_recent_entries () =
...
-        | Gt | Eq ->
-          if !Clflags.debug_digests then
+  | Eq | Gt -> (
+    let filter (data : file) =
+      match Float.compare data.stats.mtime now with
+      | Lt -> true
+      | Gt | Eq -> false
+    in
+    match !Clflags.debug_digests with
+    | false -> Path.Table.filter_inplace cache.table ~f:filter
```
snowleopard, line 104: This change is wrong: we want to filter our regardless of the flag. The conditional was only for printing out diagnostic measurements.

rgrinberg: Unless I'm missing something obvious, we're still doing the filtering.

src/dune_engine/cached_digest.ml
```diff
@@ -94,19 +94,25 @@ let delete_very_recent_entries () =
   let now = get_current_filesystem_time () in
   match Float.compare cache.max_timestamp now with
```
snowleopard, line 95:
```suggestion
  (* We can only trust digests with timestamps in the past. We had issues in the past with
      file systems having a slow internal clock, where we cached digests too aggressively. *)
  match Float.compare cache.max_timestamp now with
```

src/dune_engine/cached_digest.ml
snowleopard: Oh, sorry, you're right!

otherlibs/stdune/src/path.mli
```diff
@@ -230,6 +236,37 @@ type t = private
...
+
+  type path := t
+
+  type 'a t
+
+  val create : int -> 'a t
+
+  val clear : 'a t -> unit
+
+  val mem : 'a t -> path -> bool [@@inline always]
```
snowleopard, line 251: I'm getting 
```
the "inline" attribute cannot appear in this context
```
errors internally. Are you sure these attributes are placed correctly?

src/dune_engine/cached_digest.ml
snowleopard: I added a comment about this.

otherlibs/stdune/src/path.mli
rgrinberg: Weird. It must be a difference between flambda and the standard compiler. I've moved the annotations to the .ml file like we do elsewhere.

otherlibs/stdune/src/path.mli
```diff
@@ -230,6 +236,39 @@ type t = private
 
 include Path_intf.S with type t := t
 
+module Table : sig
+  (** Specialized tables for path. We do implement all of [Hashtbl_intf.S] -
```
snowleopard, line 240:
```suggestion
  (** Specialized tables for path. We don't implement all of [Hashtbl_intf.S],
```

otherlibs/stdune/src/path.mli
```diff
@@ -230,6 +236,39 @@ type t = private
...
+      only what we use in dune. *)
```
snowleopard, line 241:
```suggestion
  (** Specialized tables for path. We don't implement all of [Hashtbl_intf.S],
      only what we currently use in Dune. *)
```

otherlibs/stdune/src/path.ml
```diff
@@ -1220,7 +1208,77 @@ let source s = in_source_tree s
 
 let build s = in_build_dir s
 
-module Table = Hashtbl.Make (T)
+module Table = struct
+  type key = t
```
snowleopard, line 1212: Seems unused.

snowleopard:
The current version of this PR still shows as a slight regression in my benchmarks.

@rgrinberg What are the results on your end?

otherlibs/stdune/src/path.ml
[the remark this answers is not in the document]
snowleopard: Yeah, I figure this out in the end.

rgrinberg:
I'm not surprised that this change isn't a performance improvement, but I'm definitely surprised there's any regression. I imagined there would be no difference at all either way. I guess we should remember that OCaml hash tables aren't particularly space efficient anyway and removing a single word isn't going to change all that much.

We should note that our benchmarks are a bit biased against watch mode though. In watch mode, consuming less memory is more important because it makes subsequent operations faster. While in our measurements, we only care about the first completion build and do not care about any subsequent memory usage.

snowleopard:
Just a quick update: I've started some more benchmarks.

snowleopard:
Hah, now I'm seeing pretty consistent small improvements across most benchmarks! Maybe I botched it the first time.

Roughly: -0.1-0.2% in terms of allocations and -0.1-0.5% in terms of time.

snowleopard (approved):
Thanks!

snowleopard:
@rgrinberg Btw, don't forget to switch to `~store:(module Path.Table)` in `Action_builder.contents`. I'm making this change internally.

snowleopard:
> @rgrinberg Btw, don't forget to switch to `~store:(module Path.Table)` in `Action_builder.contents`. I'm making this change internally.

Just to add: with this change included, benchmarks now 2% speed-up for from-cache and null builds!

PR #8060, opened 2023-06-27
fix: Downloading local repo doesn't work

### LINKED ISSUE #7952: Fetching local repositories doesn't work
It seems this is triggering an error in `OpamRepository.pull_file` where it fails to check out a git repo. Will need to investigate the failure.

_Originally posted by @Leonidas-from-XIV in https://github.com/ocaml/dune/pull/7802#discussion_r1210339089_

Leonidas-from-XIV:
The OPAM API acts differently between `pull_file` (downloads a file and puts it in the desired location) and `pull_tree` (downloads a file or VCS repo, checks it out and unpacks it), so it needs to use the right function call depending on what we want to do.

Closes #7952.

### DIFF
--- a/src/dune_pkg/fetch.ml
+++ b/src/dune_pkg/fetch.ml
@@ -70,7 +70,7 @@ type failure =
 let fetch ~checksum ~target url =
   let open Fiber.O in
   let path = Path.to_string target in
-  let fname = OpamFilename.of_string path in
+  let dirname = OpamFilename.Dir.of_string path in
   let label = "dune-fetch" in
   let event =
     Dune_stats.(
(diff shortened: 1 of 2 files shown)


src/dune_pkg/fetch.ml
```diff
@@ -90,13 +105,10 @@ let fetch ~checksum ~target url =
                    ("checksum", `String (Checksum.to_string checksum)) :: args))
           }))
   in
-  (* hashes have to be empty otherwise OPAM deletes the file after
```
rgrinberg, line 93: Where did this comment go?

Leonidas-from-XIV: Upward into where it constructs the `pull` closure. It actually applies to both branches so I pulled it into a single identifier with the comment in 52548b92e.

gridbugs (approved):

Issue #7875, opened 2023-06-04
3.8.0 regression: Error: Library "ppx_deriving_runtime" not found.


hannesm:
## Expected Behavior

Compilation succeeds, exit code should be 0, and dist/caldav.hvt should exist.

## Actual Behavior

Compilation failure:

```
#=== ERROR while compiling mirage-unikernel-caldav-hvt.0.0.2 ==================#
# context              2.1.5 | freebsd/x86_64 |  | git+https://git.robur.coop/robur/unikernel-repo.git
# path                 /tmp/orb-build
# command              /bin/sh -exc cd mirage/ && mirage build
# exit-code            1
# env-file             /tmp/opam-hannes-12430/mirage-unikernel-caldav-hvt-12430-857f4b.env
# output-file          /tmp/opam-hannes-12430/mirage-unikernel-caldav-hvt-12430-857f4b.out
### output ###
# + cd mirage/
# + mirage build
# File "duniverse/ppx_deriving/src/api/dune", line 7, characters 24-44:
# 7 |  (ppx_runtime_libraries ppx_deriving_runtime)
#                             ^^^^^^^^^^^^^^^^^^^^
# Error: Library "ppx_deriving_runtime" not found.
# -> required by library "ppx_deriving.api" in
#    _build/default/duniverse/ppx_deriving/src/api
# -> required by library "caldav" in _build/solo5/duniverse/caldav/src
# -> required by executable main in dune.build:7
# -> required by _build/solo5/main.exe
# -> required by alias all (context solo5)
# -> required by alias default (context solo5)
# Generating static_caldavzap.ml
# Generating static_caldavzap.mli
# Generating static_tls.ml
# Generating static_tls.mli
# Generating static_caldavzap.ml
# Generating static_caldavzap.mli
# Generating static_tls.ml
# Generating static_tls.mli
```

No artifacts are produced, exit code is non-zero.

## Reproduction

I'm sorry it's likely this can be minimized, but I fail to understand how to best minimize the reproduction case.
Start with an empty switch (OCaml 4.14.1 here), and follow the steps:

```
$ opam install mirage ocaml-solo5
$ git clone https://github.com/roburio/caldav.git
$ cd caldav/mirage
$ mirage configure -t hvt
$ make lock pull build
```

[...]

PR #8061, opened 2023-06-27
ci: add an action to test mirage

emillon:
This gives us a known-good configuration of Mirage to test against. The vast majority of the changes on the Dune side will not affect this, so it is not useful to add this to the regular CI; instead, this gives us a manual run to trigger at release time or on changes for which that can be useful.

### DIFF
--- a/.github/workflows/mirage.yml
+++ b/.github/workflows/mirage.yml
@@ -8,19 +8,19 @@ jobs:
     name: Build caldav
     runs-on: ubuntu-latest
     steps:
-      - name: Clone caldav
-        uses: actions/checkout@v3
-        with:
-          repository: roburio/caldav
-          ref: 51f0d150542348dc259b7c9f7bc70ee592243f7f
-      - name: Use OCaml ${{ matrix.ocaml-compiler }}
-        uses: ocaml/setup-ocaml@v2
-        with:
-          ocaml-compiler: 4.14.x
-          opam-depext: false
-      - run: sed -i s/1.3/2.7/ dune-project
-      - run: opam pin add -n dune.dev git+https://github.com/ocaml/dune#$GITHUB_SHA
-      - run: sudo apt install libseccomp-dev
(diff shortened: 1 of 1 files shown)


emillon:
@TheLortex This should correspond to the testing strategy you've showed me before; let me know what you think.

(I included `pull_request:` to iterate on this, because `workflow_dispatch:` runs are only available if they're merged in `main` first)

emillon:
I introduced the cross-compilation regression fixed in 3.8.1 but mirage-www does not seem to trigger the issue.

emillon:
I updated the test to build caldav instead but still can't repro #7875 unfortunately

emillon:
I forgot to post an update: I updated this to use the caldav unikernel and was able to reproduce the issue. So it's in a reviewable state now.

.github/workflows/mirage.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Use OCaml ${{ matrix.ocaml-compiler }}
+        uses: ocaml/setup-ocaml@v2
+        with:
+          ocaml-compiler: 4.14.x
+          opam-depext: false
+      - run: sed -i s/1.3/2.7/ dune-project
+      - run: opam pin add -n dune.dev git+https://github.com/ocaml/dune#$GITHUB_SHA
+      - run: sudo apt install libseccomp-dev
+      - run: opam install mirage
+      - run: cd mirage; opam exec -- mirage configure -f config.ml -t hvt
```
Alizter, line 25: You can unindent this whole block 2 spaces to save some whitespace.

Alizter:
@emillon Can you run the action here so we can see it?

emillon:
You can see the previous runs [here](https://github.com/ocaml/dune/actions?query=workflow%3AMirage).

emillon:
it does not run in PRs. you'll be able to trigger that manually once it has landed in main.

PR #8063, opened 2023-06-27
bench: add GC stats to bench

Alizter:
### Problem
We want to have a convenient way of tracking memory gains in PRs. There is no way to do this currently apart from measuring Dune's perf by yourself.

### Solution
We start collecting Gc.stats in each of our build runs in the bench. In order to do so we introduce a `Metrics` module which is sufficiently polymorphic to support an unzipping operation which is useful when serialising to json. The mli of the `Metrics` module has some detailed documentation about the metrics that we collect about a process.

During the build runs, we garbage collect and compacitify before observing the statistics.

This PR changes the names of the benchmark jobs to be more uniform (use title case). This has the side effect of discontinuing previous runs, although the old data is still available in the website. Due to this I have bumped the version number of the bench.

We also enable sandboxing in the bench.

- [x] wait for #8072 
- [ ] Update `hacking.rst` bench documentation
- [ ] Remove unwanted metrics?

### DIFF
--- a/bench/bench.ml
+++ b/bench/bench.ml
@@ -90,31 +90,82 @@ let prepare_workspace () =
       Format.eprintf "cloning %s/%s@." pkg.org pkg.name;
       Package.clone pkg)
 
-let dune_build () =
+let dune_build ~name =
   let stdin_from = Process.(Io.null In) in
   let stdout_to = Process.Io.make_stdout Swallow in
   let stderr_to = Process.Io.make_stderr Swallow in
+  let gc_dump = Temp.create File ~prefix:"gc_stat" ~suffix:name in
   let open Fiber.O in
+  (* Build with timings and gc stats *)
   let+ times =
     Process.run_with_times dune ~display:Quiet ~stdin_from ~stdout_to ~stderr_to
-      [ "build"; "@install"; "--release" ]
+      [ "build"
+      ; "@install"
+      ; "--release"
+      ; "--dump-gc-stats"
+      ; Path.to_string gc_dump
+      ]
   in
-  times.elapsed_time
+  (* Read the gc stats from the dump file *)
+  Dune_lang.Parser.parse_string ~mode:Single ~fname:(Path.to_string gc_dump)
+    (Io.read_file gc_dump)
+  |> Dune_lang.Decoder.parse Dune_util.Gc.decode Univ_map.empty
+  |> Metrics.make times
 
 let run_bench () =
   let open Fiber.O in
-  let* clean = dune_build () in
+  let* clean = dune_build ~name:"clean" in
   let+ zero =
-    let open Fiber.O in
     let rec zero acc n =
       if n = 0 then Fiber.return (List.rev acc)
       else
-        let* time = dune_build () in
+        let* time = dune_build ~name:("zero" ^ string_of_int n) in
         zero (time :: acc) (pred n)
     in
     zero [] 5
   in
   (clean, zero)
+
+let display what units clean zero =
+  { Output.name = what
+  ; metrics =
+      [ ("[Clean] " ^ what, clean, units); ("[Null] " ^ what, zero, units) ]
+  }
+
+let results clean zero size =
+  (* tagging data for json conversion *)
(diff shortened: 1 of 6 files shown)


Alizter:
I'm now testing this with #8052 to see any memory change.

Alizter:
Now I will repush the bench commit and then push #8035.

Alizter:
This time I pushed a dummy commit allocating a pointless hashtable in a hotpath to see if it would be picked up and it appears to have done.

bench/bench.ml
```diff
@@ -95,17 +95,20 @@ let dune_build () =
...
   let stderr_to = Process.Io.make_stderr Swallow in
   let open Fiber.O in
+  (* Build with timings *)
   let+ times =
     Process.run_with_times dune ~display:Quiet ~stdin_from ~stdout_to ~stderr_to
       [ "build"; "@install"; "--release" ]
   in
-  times.elapsed_time
+  (* Run GC and compactify before querying GC stats *)
+  Gc.full_major ();
```
rgrinberg, line 104: You're handling the GC in the wrong process here. It's the dune command above that should spit out its GC stats.

To do this, you could add some `--dump-gc-stats-end=FILE` argument that would add an `at_exit` hook to dump this info.

rgrinberg (changes requested):

bench/bench.ml
Alizter: I've added that option in #8072 (which is currently also included here). The data I'm getting now makes much more sense. I'll try to test out your PRs again.

Alizter:
OK this time testing https://github.com/ocaml/dune/pull/8052 again.

Alizter:
This time trying https://github.com/ocaml/dune/pull/8035.

bench/bench.ml
```diff
@@ -90,32 +90,129 @@ let prepare_workspace () =
       Format.eprintf "cloning %s/%s@." pkg.org pkg.name;
       Package.clone pkg)
 
-let dune_build () =
+(* TODO move somewhere - can't go in Dune_util.Gc due to cycle with Dune_lang *)
+let gc_stat_decoder =
+  let open Dune_lang.Decoder in
```
rgrinberg, line 95: You can use Dune_sexp and the cycle will go away.

bench/bench.ml
```diff
@@ -90,32 +90,129 @@ let prepare_workspace () =
...
+  enter @@ fields
+  @@ let* minor_words = field "minor_words" float in
```
gridbugs, line 97: Instead of writing `let* foo = bar in let* baz = qux ...` you can write `let+ foo bar and+ baz = qux...` (this is true for all monadic modules not just `Dune_lang.Decoder`). This will result in slightly more efficient code and also match the style that most of code is written.

bench/bench.ml
```diff
@@ -90,32 +90,85 @@ let prepare_workspace () =
...
+  (* Build with timings and gc stats *)
   let+ times =
     Process.run_with_times dune ~display:Quiet ~stdin_from ~stdout_to ~stderr_to
-      [ "build"; "@install"; "--release" ]
+      [ "build"
+      ; "@install"
+      ; "--release"
+      ; "--dump-gc-stats"
+      ; Path.to_string gc_dump
+      ; "--sandbox"
```
rgrinberg, line 107: We should include sandboxing as a separate data point I think.

bench/bench.ml
```diff
@@ -90,32 +90,85 @@ let prepare_workspace () =
       Format.eprintf "cloning %s/%s@." pkg.org pkg.name;
       Package.clone pkg)
 
-let dune_build () =
+let dune_build ?(suffix = "") () =
```
rgrinberg, line 93: Instead of suffix, what about a mandatory argument called "name"

bench/bench.ml
Alizter: OK, I'll do that in another PR to separate concerns.

bench/bench.ml
Alizter: done

rgrinberg:
Awesome. Great work.

PR #8064, opened 2023-06-27
Upgrade to ocamlformat 0.26.0

Julow:
The changed formatting of `@@ let ...` (any operator) cause huge diffs in the command descriptions. The changed indentation of `if` and `let` in parentheses also has a big impact.  

The aim of this commit is to gather feedback.

Changelog can be found here: https://github.com/ocaml-ppx/ocamlformat/blob/main/CHANGES.md

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,4 +1,4 @@
-version=0.25.1
+version=0.26.0
 profile=conventional
 ocaml-version=4.08.0
 break-separators=before
--- a/bin/common.mli
+++ b/bin/common.mli
@@ -58,7 +58,7 @@ val init :
   -> t
   -> Dune_config.t
 
-(** [examples \[("description", "dune cmd foo"); ...\]] is an [EXAMPLES] manpage
+(** [examples [("description", "dune cmd foo"); ...]] is an [EXAMPLES] manpage
     section of enumerated examples illustrating how to run the documented
     commands. *)
 val examples : (string * string) list -> Cmdliner.Manpage.block
--- a/src/dune_rules/install_rules.ml
+++ b/src/dune_rules/install_rules.ml
@@ -510,9 +510,7 @@ end = struct
               for all. *)
            List.sort entries
              ~compare:(fun
-                        (a : Install.Entry.Sourced.t)
-                        (b : Install.Entry.Sourced.t)
-                      ->
+                 (a : Install.Entry.Sourced.t) (b : Install.Entry.Sourced.t) ->
                Install.Entry.compare Path.Build.compare a.entry b.entry))
 
   let stanzas_to_entries =
@@ -691,14 +689,13 @@ end = struct
               | Some entries ->
                 List.fold_left entries ~init:Lib_name.Map.empty
                   ~f:(fun
-                       acc
-                       { Dune_file.Library_redirect.old_name =
-                           old_public_name, _
-                       ; new_public_name = _, new_public_name
-                       ; loc
-                       ; _
-                       }
-                     ->
+                      acc
+                      { Dune_file.Library_redirect.old_name = old_public_name, _
+                      ; new_public_name = _, new_public_name
+                      ; loc
+                      ; _
+                      }
+                    ->
                     let old_public_name =
                       Dune_file.Public_lib.name old_public_name
                     in
@@ -986,8 +983,9 @@ include (
                 |> Install.Entry.map_dst ~f:(fun dst ->
                        Install.Entry.Dst.concat_all dst comps))
             |> List.sort
-                 ~compare:(fun (x : _ Install.Entry.t) (y : _ Install.Entry.t)
-                          -> Path.compare x.src y.src)
+                 ~compare:(fun
+                     (x : _ Install.Entry.t) (y : _ Install.Entry.t) ->
+                   Path.compare x.src y.src)
           | (dir, comps) :: dirs -> (
             match Path.Untracked.readdir_unsorted_with_kinds dir with
             | Error _ -> Code_error.raise "unable to read directory" []
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -14,9 +14,9 @@ let with_metrics ~common f =
                ; Pp.text "Timers:"
                ]
               @ List.map
-                  ~f:
-                    (fun ( timer
-                         , { Metrics.Timer.Measure.cumulative_time; count } ) ->
+                  ~f:(fun
+                      (timer, { Metrics.Timer.Measure.cumulative_time; count })
+                    ->
                     Pp.textf "%s - time spent = %.2fs, count = %d" timer
                       cumulative_time count)
                   (String.Map.to_list (Metrics.Timer.aggregated_timers ())))));
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -340,11 +340,11 @@ module Lock = struct
       let+ () = Fiber.return () in
       List.iter per_context
         ~f:(fun
-             { Per_context.solver_env = solver_env_from_context
-             ; context_common = { name = context_name; _ }
-             ; _
-             }
-           ->
+            { Per_context.solver_env = solver_env_from_context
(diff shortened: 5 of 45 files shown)


emillon:
Once 3.9.0 I'll upgrade to 0.25.0 first so we can see what's going on more precisely (I'll also enable `.git-blame-ignore-revs`).

src/dune_rules/dune_env.ml
```diff
@@ -232,8 +233,9 @@ module Stanza = struct
     enter
       (let+ pat =
          keyword "_" >>> return Any
-         <|> let+ p = Profile.decode in
-             Profile p
+         <|>
```
rgrinberg, line 236: IMO the old code looked better here.

Julow: What do you think of the similar change in cmd code, where the let-expression is much bigger and the operators are not chained ?

Julow:
> Once 3.9.0 is out I'll upgrade to 0.25.0 first so we can see what's going on more precisely (I'll also enable .git-blame-ignore-revs).

Let me know and I'll rebase this.

src/dune_rules/dune_env.ml
rgrinberg: The old version is better there too. The additional line only pays off when the let expression has some ugly wrapping due to the lack of horizontal space. There was no such issue there.

Alizter: Can ocamlformat pick the one that causes the least wrapping? (at least under the hood)

Julow: Not really, OCamlformat cannot compare two possible formatting. Only the input can be considered (though, multiple iterations on the whole code can be made).
But it can format differently whether there would be no wrapping at all and it can detect chained operators.

Such formatting could be possible:

```
<|> let+ p = Profile.decode in Profile p
<|> ...
```

falling back to this as soon as it doesn't fit on the line:

```
<|>
let+ p = Profile.decode in
(* longer code *)
Profile p
<|> ...
```

I'm mentioning chained operators because to me, the cmd code looks better with less indentation.
To me, that indentation *can* cause ugly wrapping means that it *does*.

bin/cache.ml
```diff
@@ -21,35 +21,35 @@ let trim =
...
+      | None, Some size ->
+        Result.Ok (Int64.sub (Dune_cache.Trimmer.overhead_size ()) size)
+      | _ -> Result.Error "please specify either --size or --trimmed-size"
+    in
+    Dune_cache.Trimmer.trim ~goal
+  with
+  | Error s -> User_error.raise [ Pp.text s ]
+  | Ok { trimmed_bytes } ->
+    User_message.print
+      (User_message.make [ Pp.textf "Freed %s" (Bytes_unit.pp trimmed_bytes) ])
```
Julow, line 52: The smaller indentation saved one line that was lost to wrapping here.

src/dune_rules/dune_project.ml
```diff
@@ -841,175 +841,172 @@ let parse_packages (name : Name.t option) ~info ~dir ~version packages
...
+            [ ("relative_to_project", `Relative_to_project)
+            ; ("inside_opam_directory", `Inside_opam_directory)
+            ])
+  in
+  fun (opam_packages : (Loc.t * Package.t Memo.t) Package.Name.Map.t) ->
+    let opam_file_location =
+      Option.value opam_file_location
+        ~default:(opam_file_location_default ~lang)
+    in
+    let generate_opam_files = Option.value ~default:false generate_opam_files in
```
Julow, line 918: Same here.

src/dune_rules/dune_env.ml
Julow: By looking at the diff again, I find that the reduced indentation do save lines by removing some wrapping. I also do not find a lot of chained operators so my previous suggestion is not valid.

Though, it should be possible have this formatting, as long as it fits on one line:
```ocaml
...
<|> let+ p = Profile.decode in Profile p
```

Should I add this before the next release ?

emillon: not a huge fan of saving lines at all cost, I think that the old version was better.

Julow: I went back and found the reason for this change: https://github.com/ocaml-ppx/ocamlformat/pull/2315
It's a consistency change, in the previous version, the line was broken on regular lets but not letops:
```ocaml
let _ =
  Cmd.v info
  @@ let+ trimmed_size = () in
     ()

let _ =
  ()
  <|> let+ p = Profile.decode in
      Profile p

let _ =
  Cmd.v info
  @@
  let trimmed_size = () in
  ()

let _ =
  ()
  <|>
  let p = Profile.decode in
  Profile p
```

The other form exists in the code: `bin/describe/describe_workspace.ml` line 378.

This diff can go away by keeping the line break from the source code, whether it's a let or a letop. (https://github.com/ocaml-ppx/ocamlformat/pull/2396)
We try to add as few of these as possible but the regression is large enough to justify it.

I'll merge this patch and update this preview PR after the weekend.

Julow:
I rebased and re-ran the preview with the `@@ let+` fix.

.git-blame-ignore-revs
```diff
@@ -1,2 +1,4 @@
 # ocamlformat 0.25.1
 3f01f6f3694e48cb1868018121f3959f8d23baca
+# Upgrade to OCamlformat 0.26.0
+8a1a919e3895787ae3781da1d6aff4cc41ce04c3
```
emillon, line 4: No need to do that in this PR since this will change

.ocamlformat
```diff
@@ -1,4 +1,4 @@
-version=0.25.1
+version = 0.26.0
```
emillon, line 1:
```suggestion
version=0.26.0
```

bin/common.mli
```diff
@@ -58,7 +58,7 @@ val init :
   -> t
   -> Dune_config.t
 
-(** [examples \[("description", "dune cmd foo"); ...\]] is an [EXAMPLES] manpage
+(** [examples [("description", "dune cmd foo"); ...]] is an [EXAMPLES] manpage
```
emillon, line 61: I think we meant this to be a list. I mean, the output should include a literal `[`. Is this what this does?

Julow: Yes, balanced `[]` don't need escaping.

.git-blame-ignore-revs
Julow: Right!

I'll make sure to remove that and fix the version line before I mark the PR as ready.

Julow: Though, do you think this PR could be merged unsquashed ?

emillon: for now I prefer doing this in 2 steps

Julow: Ok!

emillon (approved):
In my opinion, these are all improvements. Let's wait for @rgrinberg's opinion as well.

Julow:
Thanks :) Unless a problem is found, I intent to release today.

Julow:
OCamlformat 0.26.0 has been released! This PR can be merged or closed if it's a bit too early to upgrade.

rgrinberg (approved):

emillon:
I rebased this; this can land once #8311 is merged.

src/dune_rules/install_rules.ml
```diff
@@ -986,8 +983,9 @@ include (
                 |> Install.Entry.map_dst ~f:(fun dst ->
                        Install.Entry.Dst.concat_all dst comps))
             |> List.sort
-                 ~compare:(fun (x : _ Install.Entry.t) (y : _ Install.Entry.t)
-                          -> Path.compare x.src y.src)
+                 ~compare:(fun
+                     (x : _ Install.Entry.t) (y : _ Install.Entry.t) ->
+                   Path.compare x.src y.src)
```
Alizter, line 988: Indentation here is kind of funny. Is this expected @Julow?

Alizter (approved):
LGTM, just a non-blocking comment.

PR #8065, opened 2023-06-28
refactor: separate loc types

rgrinberg:
Introduce a Lexbuf.Loc.t that now represents the raw
locations we get from the lexer.

Dune's own locations are now abstract. This gives us an opportunity to
pick a better representation for them.


snowleopard (approved):
I haven't checked every line but the general idea seems good to me.

PR #8071, opened 2023-06-28
doc: more detail about make bench

Alizter:
I noticed that `make bench` is using the bootstrap dune. This might be a footgun really.

### DIFF
--- a/doc/hacking.rst
+++ b/doc/hacking.rst
@@ -686,7 +686,8 @@ Dune Bench
 ----------
 
 You can benchmark Dune's performance by running `make bench`. This will run a
-subset of the Duniverse.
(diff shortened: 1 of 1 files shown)


christinerose (approved):

doc/hacking.rst
```diff
@@ -686,7 +686,8 @@ Dune Bench
 ----------
 
 You can benchmark Dune's performance by running `make bench`. This will run a
-subset of the Duniverse.
+subset of the Duniverse. If you are running the bench locally, make sure that
+you bootstrap since that is the executable that the bench will run.
```
emillon, line 690: I think you need to add how to bootstrap.

Alizter: There is a whole section on bootstrapping above.

PR #8072, opened 2023-06-28
feature: add --dump-gc-stats option to dune build

Alizter:
This will dump a sexp into a file given by the --dump-gc-stats option which contains the data from Gc.stat after a collection.

We will use this in the bench in #8063.

- [x] changelog
- [x] tests


bin/common.ml
```diff
@@ -534,6 +534,7 @@ module Builder = struct
     ; dump_memo_graph_file : string option
     ; dump_memo_graph_format : Graph.File_format.t
     ; dump_memo_graph_with_timing : bool
+    ; dump_gc_stats : string option
```
rgrinberg, line 537: You can use `Path.External.t option` here and convert it with `Path.External.of_filename_relative_to_initial_cwd`.

bin/build_cmd.ml
```diff
@@ -54,6 +54,18 @@ let run_build_system ~common ~request =
...
+        | None -> Fiber.return ()
+        | Some file ->
+          let path = Path.of_filename_relative_to_initial_cwd file in
+          let+ gc_stat =
+            Gc.full_major ();
+            Gc.compact ();
+            Fiber.return (Gc.stat ())
+          in
+          Dune_util.Gc.serialize ~path gc_stat
+      in
```
rgrinberg, line 68: If instead you use an exit hook, this option will work for every command for free.

bin/common.ml
```diff
@@ -1136,6 +1148,17 @@ let init ?action_runner ?log_file c =
         in
         let event = Event.instant ~args common in
         Dune_stats.emit stats event);
+  (* Setup engine hook for printing GC stats to a file *)
+  let () =
+    Dune_engine.Hooks.End_of_build.always @@ fun () ->
```
rgrinberg, line 1153: This will run on every iteration of the watch mode. I doubt this is what you intended. Use `at_exit`.

Alizter: Actually I did this intentionally, but now I am not so sure and will change to `at_exit`.

Issue #8073, opened 2023-06-28
Cannot promote empty files over a non-existing file


Julow:
## Expected Behavior

I would expect empty files to be promotable.

## Actual Behavior

When diffing an empty file on the right and a non existing file on the left, Dune log that files differs and promotion doesn't create the file.

## Reproduction

Here's the diff in the `non-existent.t` test:

```diff
diff --git a/test/blackbox-tests/test-cases/promote/non-existent.t/dune b/test/blackbox-tests/test-cases/promote/non-existent.t/dune
index cf4a75c92..012f182b9 100644
--- a/test/blackbox-tests/test-cases/promote/non-existent.t/dune
+++ b/test/blackbox-tests/test-cases/promote/non-existent.t/dune
@@ -1,4 +1,4 @@
-(rule (with-stdout-to x.gen (echo "toto")))
+(rule (with-stdout-to x.gen (echo "")))
 
 (rule
  (alias blah-non-existent)
diff --git a/test/blackbox-tests/test-cases/promote/non-existent.t/run.t b/test/blackbox-tests/test-cases/promote/non-existent.t/run.t
index 5ca32bf3a..0adea85b2 100644
--- a/test/blackbox-tests/test-cases/promote/non-existent.t/run.t
+++ b/test/blackbox-tests/test-cases/promote/non-existent.t/run.t
@@ -7,12 +7,9 @@ Tests for promoting with non existent reference
   [1]
 
   $ dune build @blah-non-existent
-  File "x-non-existent", line 1, characters 0-0:
-  Error: Files _build/default/x-non-existent and _build/default/x.gen differ.
-  [1]
 
   $ dune promote
-  Promoting _build/default/x.gen to x-non-existent.
 
   $ cat x-non-existent
-  toto
+  cat: x-non-existent: No such file or directory
+  [1]
```

## Specifications

- Dune lastest `main`
- OCaml 4.14.1

Alizter:
Can you create a PR with the cram test with the promoted output and write in the cram the behaviour you expect? It will make it easier for us to take a look.

Julow:
I'll do that.

Though, this might not be a bug. The documentation mention that non-existing files are treated as empty files for the purpose of diffing. It's normal that diffing two empty files result in no diff, triggering no promotion.

Alizter:
Right, but in that case we could probably still register the file for promotion even if it is empty.

PR #8074, opened 2023-06-28
test: Reproduction for non-existent empty file promotion

Julow:
Related to issue https://github.com/ocaml/dune/issues/8073 It's a modified version of the test in
`test/blackbox-tests/test-cases/promote/non-existent.t`.

Promotion does not happen when diffing a non-existing file with an empty file.


Alizter (approved):
@rgrinberg can you merge this? I have a fix ready.

PR #8077, opened 2023-06-28
fix: diffing empty files against /dev/null

Alizter:
The old logic would set the first file as `Dev_null.path`. While cute, it doesn't have the correct semantics for when the file you are comparing to is also empty as reproduced in #8074. We therefore simplify the logic and only compare files when they exist, since we know when they don't exist they will never be equal.

- fixes #8074 
- [x] changelog
- [x] documentation
- [x] tests

### DIFF
--- a/test/blackbox-tests/test-cases/promote/non-existent-empty.t/run.t
+++ b/test/blackbox-tests/test-cases/promote/non-existent-empty.t/run.t
@@ -7,9 +7,11 @@ Tests for promoting with non existent reference
   [1]
 
   $ dune build @blah-non-existent
+  File "x-non-existent", line 1, characters 0-0:
+  Error: Files _build/default/x-non-existent and _build/default/x.gen differ.
+  [1]
 
   $ dune promote
+  Promoting _build/default/x.gen to x-non-existent.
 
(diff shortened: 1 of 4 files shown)


test/blackbox-tests/test-cases/promote/non-existent-empty.t/run.t
```diff
@@ -7,9 +7,11 @@ Tests for promoting with non existent reference
   [1]
 
   $ dune build @blah-non-existent
+  File "x-non-existent", line 1, characters 0-0:
+  Error: Files _build/default/x-non-existent and _build/default/x.gen differ.
```
rgrinberg, line 11: This error message should be clearer when one of the files doesn't exist.

Also, can you add a test case where we run a real diff program? I'm pretty sure this is going to give an ugly error that we're passing wrong arguments to diff at the moment. We should still be passing /dev/null to diff in such cases I imagine.

Alizter:
I am drafting this since I want to improve the error message, but to do that we need to do some refactoring of the diffing logic. While I am at it, I will try to give a more complete specification of the diffing / promotion behaviour in the tests.

PR #8078, opened 2023-06-29
refactor(pkg): use our usual style for equality

rgrinberg:
It's a little more concise while maintaining still giving us
exhaustivity checks when adding new field.


rgrinberg:
cc @gridbugs this is the style everywhere else

PR #8079, opened 2023-06-29
feature(pkg): add field to indicate ocaml package

rgrinberg:
We can now write:

```
(ocaml foo)
```

To designate the package `foo` as the one providing the ocaml toolchain.


Leonidas-from-XIV:
What is the usecase for this? Specifying a cross-compiler?

rgrinberg:
It's to tell dune which package to use to discover the ocaml toolchain. Ocaml, ocamldep, etc

gridbugs (approved):

Issue #8083, opened 2023-06-29
`dune` process hangs on macos


jchavarri:
Recently, we updated Ahrefs codebase to the latest version of Dune (to be more specific, commit 71024ef2e27491753f8790b236eb6041d5947b5f) from the previous version we were using from back in April (7bb6de7087e1a0a50e8cc9836d2a0e8bc200fc25), which was working fine.

After upgrading, all our Linux CI agents work perfectly fine, but on the macOS agents we have noticed that dune processes stay running forever, with CPU usage over 90%.

Here is some example of a dune process running for 55+minutes and with CPU at 100%:

```
$ ps aux | grep dune
USER              PID  %CPU %MEM      VSZ    RSS   TT  STAT STARTED      TIME COMMAND
user              3759 100.0  0.0 34302304    448   ??  R     3:20PM  55:12.57 /Users/user/builds/macvm100/ahrefs/monorepo-extra/_opam/bin/dune build -p dune-private-libs -j 7 @install
user              3737   0.0  1.1 34302560  93844   ??  S     3:19PM   0:04.48 /Users/user/builds/macvm100/ahrefs/monorepo-extra/_opam/bin/dune build -p dune-private-libs -j 7 @install
user             36569   0.0  0.0 34122844    856 s000  S+    4:15PM   0:00.00 grep dune
```

Looking at different occurrences of the issue, I couldn't find any pattern, on the packages where it appears. What I can say is that it happens on different versions of macos. In particular, the one in the agent used in the command above is `12.3.1`, but it also happens on `12.6.1` and `13.0.1`.

I have tried to gather some information about what the hanging `dune` process is exactly doing. Calling `sample 3759` with the process id gathered with `ps aux` above shows a lot of nested callbacks with `camlDune_engine__Process__fun_4514`, not sure if this is expected. You can find the full output of that command here:

[sample-3759.txt](https://github.com/ocaml/dune/files/11908133/sample-3759.txt)

Is there anything I could do to help diagnose the problem?

jchavarri:
Ok, I am trying to bisect now with older commits, it will take a while as the issue doesn't happen always.

> Before doing that you may want to test commit https://github.com/ocaml/dune/commit/886ff3b14adfb876c1b25c5b88c9637b59bd2c37 that caused some issues (on Windows, but you never know).

Hm, I tested this commit and its parent, and both seemed to work fine. Then I noticed the commit is not actually on the `main` branch, is that right?

nojb:
[...]

PR #8085, opened 2023-06-30
fix: block on interrupting async_io

rgrinberg:
The write end of the pipe should be blocking when interrupting async_io.
The code does not handle the interrupt returning non blocking writes.

### DIFF
--- a/src/dune_async_io/async_io.ml
+++ b/src/dune_async_io/async_io.ml
@@ -223,7 +223,6 @@ let with_io scheduler f =
         (udp_sock, udp_sock)
     in
     Unix.set_nonblock pipe_read;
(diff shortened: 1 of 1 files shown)


rgrinberg:
I'm not sure what we should do on windows here, but AFAIK this changes nothing for windows.

nojb:
> I'm not sure what we should do on windows here, but AFAIK this changes nothing for windows.

I agree.

nojb (approved):
LGTM

nojb:
Does this fix #8083?

rgrinberg:
I doubt it. Just something I realized from reading the code.

PR #8087, opened 2023-07-03
fix: dune hang on macOS

anmonteiro:
This is a naive attempt at fixing #8083.

Separating these calls seems to fix the problem we're seeing on macOS.

I'm not sure why this is the case, though, so consider this PR a starting point for discussion.


anmonteiro:
This seems to fix all previously broken packages. CI run green at https://github.com/nix-ocaml/nix-overlays/pull/916

emillon:
Once it's confirmed as the right fix I'll cut a 3.9.1 with this.

rgrinberg:
Thanks for the investigation. I disabled the feature properly in #8100 until we address the races.

PR #8088, opened 2023-07-03
chore: use ocamlformat 0.25.1

emillon:


emillon:
These look ok to me, mostly fixes with indent inside parentheses.

Alizter (approved):
Shame about the git history. Or did I misunderstand, are you able to ignore them?

emillon:
we are able to ignore them. the reason this PR is failing is because the corresponding ocamlformat version is not availble on nix, though.

Alizter:
25 was merged last week so if you update the nixpkgs version in the flake with `nix flake update` it should work.

emillon:
OK I just did that (for just the `nixpkgs` input)

Alizter:
Looks good to go now.

PR #8090, opened 2023-07-03
feature: use posix_spawn on macos

rgrinberg:

### DIFF
--- a/vendor/spawn/src/spawn_stubs.c
+++ b/vendor/spawn/src/spawn_stubs.c
@@ -13,6 +13,39 @@ CAMLextern int caml_convert_signal_number(int);
 
 #if defined(__APPLE__)
 
+# if defined(__MAC_OS_X_VERSION_MAX_ALLOWED)
+#  define USE_POSIX_SPAWN
+#  define vfork fork
+# endif
+
+#include <fcntl.h>
+#include <sys/socket.h>
+#include <sys/syscall.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#ifndef SYS___pthread_chdir
+#define SYS___pthread_chdir 348
+#endif
+#ifndef SYS___pthread_fchdir
+#define SYS___pthread_fchdir 349
+#endif
+
+static int __pthread_chdir(const char *path) {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated"
+  return syscall(SYS___pthread_chdir, path);
+#pragma clang diagnostic pop
+}
+
+static int __pthread_fchdir(int fd) {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated"
+  return syscall(SYS___pthread_fchdir, fd);
+#pragma clang diagnostic pop
+}
+
+
 CAMLprim value spawn_is_osx()
 {
   return Val_true;
@@ -28,6 +61,15 @@ CAMLprim value spawn_is_osx()
 #endif
 
 #if !defined(_WIN32)
+
+# if defined(USE_POSIX_SPAWN)
+#  include <spawn.h>
+
+#  if !defined(__APPLE__)
+#   define posix_spawn_file_actions_addchdir_np(...)  ENOSYS
+#   define posix_spawn_file_actions_addfchdir_np(...) ENOSYS
+#  endif
+# endif
 
 #include <assert.h>
 #include <string.h>
@@ -193,18 +235,26 @@ static void subprocess_failure(int failure_fd,
   _exit(127);
 }
 
-/* same as [dup] but ensures the result is >= 3. */
-static int safe_dup(int failure_fd, int fd)
+/* same as [dup] but ensures the result is -1 or >= 3. */
+static int safe_dup(int fd)
 {
   int new_fd = dup(fd);
-  if (new_fd == -1) subprocess_failure(failure_fd, "dup", NOTHING);
-  if (new_fd >= 3)
+  if (new_fd == -1 || new_fd >= 3)
     return new_fd;
   else {
-    int result = safe_dup(failure_fd, fd);
+    int result = safe_dup(fd);
     close(new_fd);
     return result;
   }
+}
+
+/* same as [safe_dup] but writes errors to a file descriptor,
+   as from a subprocess. */
+static int safe_dup_failure_fd(int failure_fd, int fd)
+{
+  int new_fd = safe_dup(fd);
+  if (new_fd == -1) subprocess_failure(failure_fd, "dup", NOTHING);
+  return new_fd;
 }
 
 enum working_dir_kind { PATH, FD, INHERIT };
@@ -262,7 +312,7 @@ static void subprocess(int failure_fd, struct spawn_info *info)
      when redirecting stdout to stderr for instance. */
 
   for (fd = 0; fd < 3; fd++)
-    tmp_fds[fd] = safe_dup(failure_fd, info->std_fds[fd]);
+    tmp_fds[fd] = safe_dup_failure_fd(failure_fd, info->std_fds[fd]);
 
   for (fd = 0; fd < 3; fd++)
     close(info->std_fds[fd]);
@@ -283,6 +333,22 @@ static void subprocess(int failure_fd, struct spawn_info *info)
 /* +-----------------------------------------------------------------+
    | Parent code                                                     |
    +-----------------------------------------------------------------+ */
+
+/* Raise a Unix error based on the contents of a [subprocess_failure] structure
+   (and some context arguments). */
+void raise_subprocess_failure(struct subprocess_failure* failure,
+                              value v_cwd,
+                              value v_prog)
+{
+  value arg = Nothing;
+  switch (failure->arg) {
+    case NOTHING: arg = Nothing;         break;
+    case CWD    : arg = Field(v_cwd, 0); break;
+    case PROG   : arg = v_prog;          break;
+  }
+  assert(memchr(failure->function, 0, sizeof(failure->function)));
+  unix_error(failure->error, failure->function, arg);
+}
 
 /* Convert a [string list] into a NULL terminated array of C
    strings.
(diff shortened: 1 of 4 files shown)


rgrinberg:
@anmonteiro @jchavarri can you check if you can reproduce the bug on this PR?

anmonteiro:
I can't reproduce the hang on macOS anymore with this PR.

anmonteiro:
I'm running it through the full set of packages too: https://github.com/nix-ocaml/nix-overlays/pull/926

anmonteiro:
This might not be that risky, the CI run succeeded: https://github.com/nix-ocaml/nix-overlays/actions/runs/5466251548/jobs/9950895919

jchavarri:
Can't repro the original issue with this PR either.

rgrinberg:
This PR has some advantages from a performance standpoint. So it might be good to report if you notice anything. I know @anmonteiro did some benchmarks on melange.

jchavarri:
I can see some significant improvements yes. From 8% to 25% faster with this branch, it seems to affect particularly Melange builds (that's where I see the 25% improvement). Note that the mac CI nodes we use have 12 cores, I imagine that the more capacity there is to parallelize, the larger the impact on build times?

anmonteiro:
@rgrinberg what's missing to get this one over the line?

rgrinberg:
If you could read over and familiarize yourself with the C to give a formal review that would be a great start.

vendor/spawn/src/spawn_stubs.c
```diff
@@ -480,18 +511,201 @@ CAMLprim value spawn_unix(value v_env,
...
+  if (info.set_pgid) {
+    e_error = posix_spawnattr_setpgroup(&attr, info.pgid);
+    if (e_error) {
+      e_function = "posix_spawnattr_setpgroup";
+      goto cleanup;
+    }
+  }
+
+  for (int fd = 0; fd < 3; fd++) {
+    int tmp_fd = tmp_fds[fd] = safe_dup(info.std_fds[fd]);
```
anmonteiro, line 590: It's probably worth it to copy the upstream concern about resetting close-on-exec, and think about whether we'd want to solve it here:

https://github.com/janestreet/spawn/pull/47/files#r930889893

rgrinberg: Yeah, we'll do it upstream. Note that I've talked to the maintainers and we agreed that this isn't an issue for dune because we don't allow anywhere else in the code to use fork.

rgrinberg:
@anmonteiro could you give this another test please? I re-enabled threads on macos since they longer should be a problem and should give us an additional perf benefit.

anmonteiro:
Running this through all the packages here: https://github.com/nix-ocaml/nix-overlays/pull/1000

anmonteiro (approved):
LGTM.

All macOS packages build on nix-overlays. They were previously broken with `fork`

jchavarri:
Just reporting some build times i measured after this PR was merged in a quite large Melange build on our codebase, using my laptop (16 cores).

Using an older version of dune (e1f3e19cb309ffebc3a7517cd8991f4582ead896):

```
time dune build @fe.all
dune build @fe.all  2539.49s user 1807.54s system 581% cpu 12:27.02 total
```

Using latest dune (https://github.com/ocaml/dune/commit/758e3705fadec3a2f1e73bf5cb38312d2e6f259e):

```
time dune build @fe.all
dune build @fe.all  2518.63s user 1663.82s system 889% cpu 7:49.98 total
```

Thanks @rgrinberg !

ixzzd:
Here are the results of compiling the same large codebase on the M2 MacBook Air on latest dune (https://github.com/ocaml/dune/commit/df4399bd14e6266eb3a3aa6daf31db1947c834fe)
```
$ time opam exec -- dune build @fe.all

________________________________________________________
Executed in  226.01 secs    fish           external
   usr time  705.37 secs    0.26 millis  705.37 secs
   sys time  412.18 secs    1.31 millis  412.18 secs

$ opam exec -- dune --version
3.10.0-259-gdf4399b
```

vs  older version of dune (https://github.com/ocaml/dune/commit/e1f3e19cb309ffebc3a7517cd8991f4582ead896)
```
$ time opam exec -- dune build @fe.all

________________________________________________________
Executed in  296.71 secs    fish           external
   usr time  675.71 secs    0.27 millis  675.71 secs
   sys time  498.06 secs    1.32 millis  498.06 secs
```

`296.71` secs vs `226.01` secs

@rgrinberg 👏 👏 👏

barracuda156:
@rgrinberg This does not seem right. Let me try fixing this and see if it builds then.

vendor/spawn/src/spawn_stubs.c
```diff
@@ -13,6 +13,39 @@ CAMLextern int caml_convert_signal_number(int);
 
 #if defined(__APPLE__)
 
+# if defined(__MAC_OS_X_VERSION_MAX_ALLOWED)
```
barracuda156, line 16: @rgrinberg What this macro is supposed to do specifically?

rgrinberg: Yeah, this is wrong. The check should be for 10.5 because we need pthread_chdir

barracuda156: I guess this is what fails for me on 10.6 ppc, which will be similar to 10.5 in this regard.

barracuda156: Let me see if with these fixes it works. Will update in few min.

rgrinberg: Okay, thanks. Note that we can't use `vfork` on macos. So it must be a leftover from the old code when we did.

barracuda156:
@rgrinberg After these https://github.com/ocaml/dune/commit/39703e071b8c4d1c6b8a8fec48074221097ca65d I got here:
```
Error: execve(../../install/default/bin/dune): Exec format error
-> required by _build/default/doc/dune-build.1
-> required by _build/install/default/man/man1/dune-build.1
-> required by _build/default/dune.install
-> required by alias install
make: *** [release] Error 1
```
How to get a proper log to see what fails exactly?

rgrinberg:
What kind of log are you looking for? There is no more information for this error.

Try running this executable maunally `../../install/default/bin/dune` and see if it works.

barracuda156:
> What kind of log are you looking for? There is no more information for this error.

Well, just in case.

> Try running this executable maunally `../../install/default/bin/dune` and see if it works.

This is literally a symlink to an empty file. So for w/e reason it is not written correctly, and build system is unaware of that.

P. S. Could you please take a look at my changes and say what should be done re `fork`? I know pretty well how macOS-specific macros work and which should be used, but I have no idea whether that `fork` define should or should not be used, and if yes, then conditional on what.

rgrinberg:
> This is literally a symlink to an empty file. So for w/e reason it is not written correctly, and build system is unaware of that.

Okay, that's the source of the error then. What command were you using to build dune and then to generate this error?

> Could you please take a look at my changes and say what should be done re fork? I know pretty well how macOS-specific macros work and which should be used, but I have no idea whether that fork define should or should not be used, and if yes, then conditional on what.

Sure, where are your changes?

Regarding posix_spawn vs fork vs vfork:

1. We should only use vfork on Linux
2. We should use posix_spawn if it's available and has all the options we need (MacOS in conjunction with pthread_chdir)
3. If none of the above are available, we fallback to fork.

So on macos, we should only use fork if our posix_spawn is good enough and pthread_chdir is available (I believe anything above 10.5, but I could be wrong).

barracuda156:
@rgrinberg Changes are here: https://github.com/ocaml/dune/commit/39703e071b8c4d1c6b8a8fec48074221097ca65d
`environ` part should be correct.
All PowerPC should use same fallbacks as 10.5 (therefore additional condition for `__ppc__`, since it exists on 10.6, but does not work exactly like 10.6 Intel).

And `fork` part is wrong, I guess :) How to fix it?

barracuda156:
@rgrinberg Changing the code to:
```
#if defined(__APPLE__)
# include <AvailabilityMacros.h>

# if (MAC_OS_X_VERSION_MAX_ALLOWED >= 1060) && !defined(__ppc__)
#  define USE_POSIX_SPAWN
# endif

# define vfork fork

#include <fcntl.h>
#include <sys/socket.h>
#include <sys/syscall.h>
#include <sys/types.h>
#include <unistd.h>
```
I still get `Error: execve(../../install/default/bin/dune): Exec format error`.

P. S. This new error does not need to be spawn-related. The last version I can say built and worked for sure was 3.7.0 and master branch as of 2022.11.18. Everything in between that and 3.11.0 has not been tried, at least yet.

Alizter:
@barracuda156 Could you try 3.10 to make sure it is this PR that is causing problems.

barracuda156:
@Alizter It looks like we have fixed `spawn` issue in https://github.com/ocaml/dune/compare/main...barracuda156:dune:darwin

`Error: execve(../../install/default/bin/dune): Exec format error` is not related to `spawn` though. In fact, I get it with 3.8.0, 3.9.0 too. Something has broken elsewhere.

rgrinberg:
Try disabling fast copying completely with `DUNE__COPY_FILE=portable`. Does that help?

barracuda156:
> Try disabling fast copying completely with `DUNE__COPY_FILE=portable`. Does that help?

Sorry for a silly question, where do I pass this? Configure does not seem to take this as an arg.

rgrinberg:
It's an environment variable, so you just need it to set when running dune. E.g. `DUNE__COPY_FILE=portable make dev`.

barracuda156:
Unfortunately, no effect:
```
--->  Building ocaml-dune
Executing:  cd "/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_ocaml_ocaml-dune/ocaml-dune/work/dune-3.11.1" && DUNE__COPY_FILE=portable /usr/bin/make -j6 release 
ocamlc -output-complete-exe -w -24 -g -o .duneboot.exe -I boot unix.cma boot/libs.ml boot/duneboot.ml
./.duneboot.exe
Error: execve(../../install/default/bin/dune): Exec format error
-> required by _build/default/doc/dune-build.1
-> required by _build/install/default/man/man1/dune-build.1
-> required by _build/default/dune.install
-> required by alias install
make: *** [release] Error 1
Command failed:  cd "/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_ocaml_ocaml-dune/ocaml-dune/work/dune-3.11.1" && DUNE__COPY_FILE=portable /usr/bin/make -j6 release 
Exit code: 2
```
But perhaps we patched out both copy_file and posix_spawn issues. This is something else, it seems.

Alizter:
@barracuda156 I think that's a typo, it should be `DUNE_CONFIG__COPY_FILE`.

barracuda156:
@Alizter @rgrinberg So yes, `DUNE_CONFIG__COPY_FILE` worked, `dune` 3.11.1 has built now!

What are the conclusions? Could we fix it properly rather than just disabling? `dune` is an important port, for OCaml ecosystem very much so, it is desirable to have it working optimally, as long as it is feasible. I can test any suggestions on my end.

barracuda156:
@Alizter OCaml does support all 32-bit platforms including powerpc in bytecode compiler (this applies to OCaml 5). That was the stance of upstream at least rather recently. Native was broken anyway, so nothing changes.

PR #8091, opened 2023-07-03
Don't solve for multiple contexts if lockdirs collide

gridbugs:
When running `dune pkg lock --all-contexts` it is now an error if there are multiple build contexts which specify the same lockdir path. Also adds whitespace between the log when solving dependencies to make it easier to read the output of `dune pkg lock --all-contexts`.

### DIFF
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -132,6 +132,7 @@ module Lock = struct
       { lock_dir_path : Path.Source.t
       ; version_preference : Version_preference.t
       ; solver_env : Dune_pkg.Solver_env.t
+      ; context_common : Dune_rules.Workspace.Context.Common.t
       }
 
     let choose ~context_name_arg ~all_contexts_arg ~version_preference_arg =
@@ -164,6 +165,7 @@ module Lock = struct
               { lock
               ; version_preference = version_preference_context
               ; solver_env
+              ; base = context_common
               ; _
               }) ->
           [ { lock_dir_path = Option.value lock ~default:Lock_dir.default_path
@@ -172,6 +174,7 @@ module Lock = struct
                   ~from_context:version_preference_context
             ; solver_env =
                 Option.value solver_env ~default:Dune_pkg.Solver_env.default
+            ; context_common
             }
           ]
         | Some (Opam _) ->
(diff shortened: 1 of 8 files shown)


bin/pkg.ml
```diff
@@ -177,14 +183,48 @@ module Lock = struct
...
+              | Some context_names -> Some (context_name :: context_names)))
+      in
+      List.find_opt (Path.Source.Map.to_list context_names_by_lock_dir_path)
+        ~f:(fun (_, context_names) -> List.length context_names > 1)
+
+    let check_for_dup_lock_dir_paths ts =
+      match contexts_with_dup_lock_dir_paths ts with
+      | None -> ()
+      | Some (lock_dir_path, context_names) ->
+        User_error.raise
```
rgrinberg, line 218: This error should include the location from the workspace file.

gridbugs: This error doesn't necessarily indicate a problem in the workspace file though. If you're just generating one lockdir at a time it's valid (at the moment at least) for all the contexts to use the same lockdir. They will just overwrite each other. Including the location in the error might mislead users to think that they need to change the workspace file.

Alternatively we could make it an error to have multiple contexts with the same lockdir path in general in which case we'd include the location in the error message.

Not sure which solution would be most appropriate.

rgrinberg: Sure, the error doesn't mean the workspace file is bad, but it's still useful to point at where dune obtained the conflicting information.

gridbugs: Fair enough. I just changed it to include the lines from the workspace file spanning the conflicting context definitions. In the process I ran into a problem where `Loc.span` would produce unexpected results when its second argument ended ended before its first argument which I fixed.

bin/pkg.ml
```diff
@@ -228,12 +271,13 @@ module Lock = struct
...
     let write_disk_list =
-      List.map per_context
-        ~f:(fun { Per_context.lock_dir_path; version_preference } ->
+      List.mapi per_context
+        ~f:(fun i { Per_context.lock_dir_path; version_preference; _ } ->
           let summary, lock_dir =
             Dune_pkg.Opam.solve_lock_dir ~version_preference ~repo_selection
               opam_file_map
           in
+          if i > 0 then Console.printf "";
```
rgrinberg, line 280: I just noticed this, but isn't it weird that print these messages before actually writing anything to disk?

rgrinberg: In any case, you can improve `summary` to return ` _ Pp.t` and then use `Pp.concat` with the appropriate `~sep` to add the spacing you need.

bin/pkg.ml
rgrinberg: Not sure if `Loc.span` is the right thing to use since it might include a lot of irrelevant stuff. 

Look for `package_defined_twice` in `dune_project.ml` for an example of how we have two locations in a single error.

bin/pkg.ml
gridbugs: Ok I've updated this to defer printing until after this loop and adding newlines with `Pp.concat`.

bin/pkg.ml
```diff
@@ -227,18 +270,22 @@ module Lock = struct
...
-          Lock_dir.Write_disk.prepare ~lock_dir_path lock_dir)
+          let summary_message =
+            Dune_pkg.Opam.Summary.selected_packages_message summary
+              ~lock_dir_path
+            |> User_message.pp
+          in
+          (Lock_dir.Write_disk.prepare ~lock_dir_path lock_dir, summary_message))
+      |> List.split
     in
+    Format.printf "%a" Pp.to_fmt (Pp.concat ~sep:Pp.newline summary_pps);
```
rgrinberg, line 288: Should be `Dune_console.print` or the like. Otherwise this will break threading.

gridbugs: Ah cool I didn't know about `Dune_console.print`. Updated.

Alizter: Or really `Console.printf`.

gridbugs: Looks like `Console` is an alias of `Dune_console` (at least in the context of this line)? Should one be preferred over the other?

Alizter: The library is called `dune_console` we make an alias `Console` so we don't have to write `Dune_` everywhere.

gridbugs: Gotcha. Updated.

gridbugs:
Are any other changes required before this is ready to merge?

bin/pkg.ml
rgrinberg: Using span is still suspicious here

bin/pkg.ml
```diff
@@ -186,18 +189,50 @@ module Lock = struct
...
+      |> List.find_opt ~f:(fun (_, context_commons) ->
+             List.length context_commons > 1)
+
+    let check_for_dup_lock_dir_paths ts =
+      contexts_with_dup_lock_dir_paths ts
+      |> Option.iter ~f:(fun (lock_dir_path, context_commons) ->
+             let loc =
+               List.map context_commons
+                 ~f:(fun (c : Dune_rules.Workspace.Context.Common.t) -> c.loc)
+               |> List.reduce ~f:Loc.span |> Option.value_exn
```
rgrinberg, line 222: Using span is wrong here.

PR #8103, opened 2023-07-04
fix: make User_message.t marshabllable

rgrinberg:


dkalinichenko-js (approved):
LGTM. I would just change `User_message.t` to have static fields, but this is more general.

PR #8104, opened 2023-07-04
fix: rally action runner errors

rgrinberg:
Whenever an action runner raises, it would not send back the response
for executing the action.

This PR changes the behavior to catch the action runner errors and pass
them along to the server as needed.


dkalinichenko-js (approved):
LGTM.

dkalinichenko-js (approved):

PR #8105, opened 2023-07-04
feature(pkg): Automatic fetching of opam-repository

Leonidas-from-XIV:
This PR is an exploration on downloading a copy of opam-repository for locking. It is surprisingly easy to implement and it works rather fast. Currently the Git repo that it downloads is 142M but an initial lock of dune takes 11s and a consecutive lock with the repo downloaded takes about 3s.

I'm honestly rather pleasantly surprised how nicely this works.

### DIFF
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -1,90 +1,116 @@
 open Import
 module Lock_dir = Dune_pkg.Lock_dir
+module Fetch = Dune_pkg.Fetch
+module Opam = Dune_pkg.Opam
+module Repo_selection = Opam.Repo_selection
+
+module Opam_repository = struct
+  type t = { url : OpamUrl.t }
+
+  let of_url url = { url }
+
+  let default =
+    of_url @@ OpamUrl.of_string "https://opam.ocaml.org/index.tar.gz"
+
+  let is_archive name =
+    List.exists
+      ~f:(fun suffix -> String.is_suffix ~suffix name)
+      [ ".tar.gz"; ".tgz"; ".tar.bz2"; ".tbz"; ".zip" ]
+
+  let path =
+    let open Fiber.O in
+    let ( / ) = Filename.concat in
+    fun { url } ->
+      Fiber.of_thunk @@ fun () ->
+      let target_dir =
+        Xdg.cache_dir (Lazy.force Dune_util.xdg) / "dune/opam-repository"
+      in
+      let target = target_dir |> Path.External.of_string |> Path.external_ in
+      let unpack = url |> OpamUrl.to_string |> is_archive in
+      let+ res = Fetch.fetch ~unpack ~checksum:None ~target url in
+      match res with
+      | Ok () -> Ok target
+      | Error _ as failure -> failure
+end
 
 module Lock = struct
-  module Repo_selection = struct
-    module Env = struct
-      module Source = struct
-        type t =
-          | Global
-          | Pure
+  module Env = struct
+    module Source = struct
+      type t =
+        | Global
+        | Pure
 
-        let to_string = function
-          | Global -> "global"
-          | Pure -> "pure"
+      let to_string = function
+        | Global -> "global"
+        | Pure -> "pure"
 
-        let default = Global
+      let default = Global
 
-        let term =
-          let all = [ Global; Pure ] in
-          let all_with_strings = List.map all ~f:(fun t -> (to_string t, t)) in
-          let all_strings = List.map all_with_strings ~f:fst in
-          let doc =
-            sprintf
-              "How to initialize the opam environment when taking the opam \
-               repository from a local directory (may only be used along with \
-               the --opam-repository-path option). Possible values are %s. \
-               '%s' will use the environment associated with the current opam \
-               switch. '%s' will use an empty environment. The default is \
-               '%s'."
-              (String.enumerate_and all_strings)
-              (to_string Global) (to_string Pure) (to_string default)
-          in
-          Arg.(
-            value
-            & opt (some (enum all_with_strings)) None
-            & info [ "opam-env" ] ~doc)
-      end
-
-      let of_source =
-        let open Dune_pkg.Opam.Env in
-        function
-        | Source.Global -> global ()
-        | Pure -> empty
+      let term =
+        let all = [ Global; Pure ] in
+        let all_with_strings = List.map all ~f:(fun t -> (to_string t, t)) in
+        let all_strings = List.map all_with_strings ~f:fst in
+        let doc =
+          sprintf
+            "How to initialize the opam environment when taking the opam \
+             repository from a local directory (may only be used along with \
+             the --opam-repository-path option). Possible values are %s. '%s' \
+             will use the environment associated with the current opam switch. \
+             '%s' will use an empty environment. The default is '%s'."
+            (String.enumerate_and all_strings)
+            (to_string Global) (to_string Pure) (to_string default)
+        in
+        Arg.(
+          value
+          & opt (some (enum all_with_strings)) None
+          & info [ "opam-env" ] ~doc)
     end
 
+    let of_source =
+      let open Dune_pkg.Opam.Env in
+      function
+      | Source.Global -> global ()
+      | Pure -> empty
+  end
+
+  module Opam_repository_path = struct
     let term =
-      let+ opam_repository_path =
-        Arg.(
-          value
-          & opt (some string) None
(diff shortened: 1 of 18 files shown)


gridbugs (approved):
Awesome! It will be nice to get rid of the opam switch logic. That will really simplify the solver. You're welcome to remove the opam switch stuff from src/dune_pkg/opam_solver.ml or one of us can do that in a later change.

bin/pkg.ml
```diff
@@ -1,90 +1,82 @@
...
-      end
+  let path { url } =
+    let open Fiber.O in
+    let ( / ) = Filename.concat in
+    Fiber.of_thunk (fun () ->
+        let target_dir = Xdg.cache_dir xdg / "dune/opam-repository" in
+        let target = target_dir |> Path.External.of_string |> Path.external_ in
+        let+ res = Fetch.fetch ~checksum:None ~target url in
+        match res with
+        | Ok () -> Ok (Path.to_string target)
```
rgrinberg, line 24: Perhaps we're stringifying the path a bit early?

Leonidas-from-XIV: It's a good point, I've converted the function to return a `Path.t` and `Repo_selection` to take a `Path.t`. Oddly enough this seems to work fine, contrary to the large explanatory comment.

@gridbugs, am I missing something?

gridbugs: Possibly it's because you're deferring the creation of the path until after the global state it depends on has been initialised.

gridbugs: Can you test that it works for relative external paths such as "../opam-repository"?

Leonidas-from-XIV: You're right, it doesn't work with relative paths:

```
Description:
  ("Path.External.of_string: relative path given",
  { t = "../../.cache/dune/opam-repository/" })
```

This looks like a deliberate restriction of `Path.External`, but in this case I think relative paths should be supported. So back to `string`?

rgrinberg: Yes, you can use a string. Or just use `Path.Outside_build_dir.t`.

Note that where I initially pointed out the issue, the stringifying wasn't being done in the command line argument parsing

Leonidas-from-XIV: Using `Path.Outside_build_dir.t` I get `Error: path outside the workspace: ../../.cache/dune/opam-repository/ from .` Oddly enough it works fine if I specify an absolute path (that is also outside the workspace), so there seems to be some discrepancy between how it is parsed between relative or absolute paths.

rgrinberg: Ah yes, I think we just need to normalize such relative paths to `Path.External.t` though.

Leonidas-from-XIV: `Path.External.of_filename_relative_to_initial_cwd` seems to be exactly what I am looking for.

rgrinberg:
For opam-repository, why do we download the git repo rather than the raw tarball? I understand using git repos for actual package sources, but here it seems like we'll always be downloading much more than we'll ever need.

Leonidas-from-XIV:
I don't have any strong opinion on it but my idea of downloading the git repo is because that makes it easier to update e.g. the next time `lock` is called, as only the differences between the last and current opam-repo need to be fetched.

rgrinberg:
That's probably worth having experimenting, but I'm not entirely sure if this is optimal for opam-repository. The issue is that we're going to fetch a bunch of intermediate git objects that will be of little use. I would imagine something rsync like on top of the http protocol would end up being much faster.

Alizter:
Is there ever a reason why we would want historical versions of an opam repository? If not, then it doesn't make much sense to use git IMO.

rgrinberg:
There's a bit more discussion to be had here, but it's quite possible that we'll have the equivalent of `opam update` on a per project basis. In which case, we will need to store more than one version.

bin/pkg.ml
gridbugs: Can you refer to paths inside the current project with that function? It's not a common use case but comes in very handy for writing tests.

Leonidas-from-XIV:
There's the question whether the lockfile should e.g. retain information about the repos and their versions and whether reproducability of lockfiles is a goal but for now I would say one global copy of opam-repository with a single version is probably going to suffice and does not prevent us from implementing a more elaborate solution later on if there is demand.

I'll talk to OPAM folks about downloading the tarball vs git. The latter has the advantage of being on Github thus profiting from their CDN, will ask what the state of the tarball is (it's hopefully also CDN'ed).

Alizter:
Just a reminder not to skip over other opam repositories like https://github.com/coq/opam-coq-archive and the OCaml beta one.

gridbugs:
If someone really wants reproducible lockdirs they could always clone the opam repo themselves and checkout a particular revision and then point dune at that. There are other factors that influence the resluting lockdir including flags/variables set in the build context or passed on the command line so I guess there's a question of whether to store all this information inside the lock.dune file in the lockdir so that lockdirs are reproducable. Kind of a lockfile for your lockfile. I'm not sure if it's something we want to fully support though as code changes in dune will also affect lockdir generation.

Leonidas-from-XIV:
@Alizter I think just the default opam-repository is a decent starting point, but you're right, it should be customizable. In the near-future we might want to extend the dune-project file with some stanza to specify the repos to use for locking, as I could see this to be a useful addition (we added this as an extra OPAM field in opam-monorepo and it has been extremely useful).

Wrt to reproducibility, I think it is an excellent idea. The issue here is that it this information is unfortunately optional. The opam-repository tarball has a field `stamp` where I assume the git hash is stored and if we clone the repo from git we can get this info, but if the repo is retrieved as tarball without `stamp` field, we can't do much. Maybe include the tarball hash as the version of the repo? In any case, if we decide on how to determine the version of a repository I'm more than happy to implement this in a future PR so we can use the version as input for the lockfile generation.

bin/pkg.ml
```diff
@@ -10,16 +10,19 @@ module Opam_repository = struct
   type t = { url : OpamUrl.t }
 
   let default =
-    { url = OpamUrl.of_string "git+https://github.com/ocaml/opam-repository.git"
-    }
+    { url = OpamUrl.of_string "https://opam.ocaml.org/index.tar.gz" }
+
+  let is_tarball name =
+    String.is_suffix ~suffix:".gz" name || String.is_suffix ~suffix:".bz" name
```
gridbugs, line 18: Nitpick but you can totally have non-tarball .gz files. Should this check for ".tag.gz" and ".tgz" instead of just ".gz"?

Leonidas-from-XIV: Yeah, and also `.tar.bz2`, `.tbz` and `.zip`, they should all be supported by OPAM. I've updated the code!

Leonidas-from-XIV:
Ok, freshly rebased and now featuring tests that show that it does download the file and store it in the dune cache.

It uses some somewhat exotic commands in the test like `shuf` and `sleep` with fractional numbers but these days these should be available in BSDs as well.

bin/pkg.ml
Leonidas-from-XIV: Yep, it seems to work fine. I was able to adjust all the test cases just fine and they work well.

test/blackbox-tests/test-cases/pkg/opam-repository-download.t
```diff
@@ -0,0 +1,100 @@
...
+  $ # wait for server to come up, this takes a bit
+  $ while [ ! -f server.pid ]; do sleep 0.01; done
+
+Run Dune and tell it to cache into our custom cache folder.
+
+  $ mkdir fake-xdg-cache
+  $ XDG_CACHE_HOME=$(pwd)/fake-xdg-cache dune pkg lock --opam-repository-url=http://localhost:$PORT/index.tar.gz
+  Solution for dune.lock:
+  bar.0.0.1
+  foo.0.0.1
```
Alizter, line 84: I guess this relies on the package version corresponding to a unique archive. What happens if you have a `.dev` version for instance? In the Coq opam archive, they have a `.dev` version which opam checks to be out of date and updates. If only the version is considered when fetching from the cache, wouldn't this cause issues?

When the opam archive updates, how do we account for this in the package cache?

src/dune_pkg/fetch.mli
```diff
@@ -13,7 +13,8 @@ type failure =
     @raise Unavailable
       When the file can't be retrieved, e.g. not available at the location. *)
 val fetch :
-     checksum:Checksum.t option
+     ?unpack:bool
```
rgrinberg, line 16: Let's make this argument mandatory to make callers always think about it.

Alizter (approved):

PR #8109, opened 2023-07-05
refactor: use Command.Args.t for Modules.pp

emillon:
This allows using the concatenation DSL instead of list and string manipulation.

### DIFF
--- a/src/dune_rules/command.mli
+++ b/src/dune_rules/command.mli
@@ -63,9 +63,6 @@ module Args : sig
   (** Create dynamic command line arguments. *)
   val dyn : string list Action_builder.t -> _ t
 
-  (** Like [dyn] but output in a single argument, quoted and concatenated. *)
-  val dyn_as_single_arg : string list Action_builder.t -> _ t
-
   (** Create an empty command line. *)
   val empty : _ t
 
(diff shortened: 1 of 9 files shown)


src/dune_rules/command.mli
```diff
@@ -63,6 +63,9 @@ module Args : sig
   (** Create dynamic command line arguments. *)
   val dyn : string list Action_builder.t -> _ t
 
+  (** Like [dyn] but output in a single argument, quoted and concatenated. *)
+  val dyn_as_single_arg : string list Action_builder.t -> _ t
```
rgrinberg, line 67: Not a fan of this function. The name is weird and the function's semantics are best understood just by reading the implementation. It's also barely useful as it's only used once in the code base.

src/dune_rules/module.mli
```diff
@@ -61,7 +61,7 @@ val path : t -> Module_name.Path.t
 
 val source : t -> ml_kind:Ml_kind.t -> File.t option
 
-val pp_flags : t -> (string list Action_builder.t * Sandbox_config.t) option
+val pp_flags : t -> (Command.Args.any Command.Args.t * Sandbox_config.t) option
```
rgrinberg, line 64: The old type was actually superior because it prevented the flags from introducing hidden targets. Command.Args.t doesn't do that. But even without that, the `string list Action_builder.t` is strictly less powerful and enough for this job. Why go for the bigger hammer?

rgrinberg (changes requested):

emillon:
Good points. The changes I wanted to do (use `Args` more) actually don't require changes to `Module` itself. How does that look now?

rgrinberg (approved):

PR #8113, opened 2023-07-05
refactor: process handling

rgrinberg:
Refactor process handling. This is has two purposes:

* Simplify the implementation and make it easier to follow
* Prepare it for making it friendly for action runners

At a high level, the refactoring does the following:

* Introuce a type for a running process [Process.t]
* Introduce a type for the result of running a process
  [Process.Result.t]
* Introduce a type for handling stdout/stderr without accidentally
  forgetting to clean it up or reading it after they've been deleted.

### DIFF
--- a/src/dune_engine/process.ml
+++ b/src/dune_engine/process.ml
@@ -139,6 +139,11 @@ module Io = struct
       else Unix.close (Lazy.force t.fd)
 
   let multi_use t = { t with status = Keep_open }
+
+  let output_on_success (out : output t) =
+    match out.kind with
+    | Terminal x -> x
+    | _ -> Print
 end
 
 type purpose =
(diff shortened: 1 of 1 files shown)


src/dune_engine/process.ml
```diff
@@ -572,6 +577,104 @@ end = struct
...
+  ; pid : Pid.t
+  ; response_file : Path.t option
+  ; stdout : Path.t option
+  ; stderr : Path.t option
+  ; stdout_on_success : Action_output_on_success.t
+  ; stderr_on_success : Action_output_on_success.t
+  }
+
+module Result = struct
+  type nonrec process = t
```
Alizter, line 591: Why nonrec?

src/dune_engine/process.ml
```diff
@@ -572,6 +577,104 @@ end = struct
...
+  module Out = struct
+    type state =
+      | No_capture
+      | File of Path.t
+      | Read of string
+      | Closed
+
+    type t =
+      { on_success : Action_output_on_success.t
+      ; mutable unexpected_output : bool
```
dkalinichenko-js, line 602: I don't like mutable fields here. For now, it's fine because this type is not exposed anywhere in the .mli, but it seems wrong to provide an interface with it.

dkalinichenko-js (approved):
Mostly LGTM, but some parts are hard to judge without seeing the .mli.

src/dune_engine/process.ml
rgrinberg: Yeah, it's a bit ugly, but I encourage you to understand the old code though :)

We handle 3 separate concerns with mutation:

1. We make sure that we never read the file unless we need the output
2. Subsequent reads are free
3. If we need to swallow the output, we make sure the output we read is always empty.

It also allows us to introduce an optimization:

To check if the output is empty, we don't need to read the file. We can just get its size using stat. That saves an open/close on empty files which should be quite common.

PR #8122, opened 2023-07-06
fix: Initialize async IO thread lazily

emillon:
The goal of that change is to avoid effects of threads when they are not actually needed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@ Unreleased
 - Disable background operations and threaded console on MacOS and other Unixes
   where we rely on fork. (#8100, #8121, fixes #8083, @rgrinberg, @emillon)
 
+- Initialize async IO thread lazily. (#8122, @emillon)
+
 - Add `dune build --dump-gc-stats FILE` argument to dump Garbage Collection
   stats to a named file. (#8072, @Alizter)
 
--- a/src/dune_async_io/async_io.ml
+++ b/src/dune_async_io/async_io.ml
@@ -222,7 +222,7 @@ module T_var : sig
 
   val get_exn : unit -> t Fiber.t
 
-  val set : t -> (unit -> 'a Fiber.t) -> 'a Fiber.t
+  val setup : t -> (unit -> 'a Fiber.t) -> 'a Fiber.t
 end = struct
   let t_var = Fiber.Var.create ()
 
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -4,6 +4,8 @@ Unreleased
 - Disable background operations on MacOS and other Unixes where we rely on
   fork. (#8100, fixes #8083, @rgrinberg)
 
+- Initialize async IO thread lazily. (#...., @emillon)
```
emillon, line 7:
```suggestion
- Initialize async IO thread lazily. (#8122, @emillon)
```

src/dune_async_io/async_io.ml
```diff
@@ -233,16 +258,10 @@ let with_io scheduler f =
...
-    Scheduler.spawn_thread (fun () ->
-        Mutex.lock t.mutex;
-        Exn.protect
-          ~f:(fun () -> select_loop t)
-          ~finally:(fun () -> Mutex.unlock t.mutex))
-  in
-  Fiber.Var.set t_var t (fun () ->
+  T_var.set t (fun () ->
       Fiber.finalize f ~finally:(fun () ->
           Mutex.lock t.mutex;
```
rgrinberg, line 266: Do we need this finalization if the thread wasn't started?

rgrinberg (changes requested):

src/dune_async_io/async_io.ml
```diff
@@ -208,10 +209,34 @@ let rec select_loop t =
...
-let t_var = Fiber.Var.create ()
+let start t =
+  let module Scheduler = (val t.scheduler : Scheduler) in
+  Scheduler.spawn_thread (fun () ->
+      Mutex.lock t.mutex;
+      Exn.protect
+        ~f:(fun () -> select_loop t)
+        ~finally:(fun () -> Mutex.unlock t.mutex))
+
+module T_var : sig
```
rgrinberg, line 220: It would be nice if this module handled the start field exclusively. But we're still missing some handling in the finalizer. So my proposal would be either one of:

1. Remove the module completely
2. Move the conditional finalization to this module.

rgrinberg (changes requested):

src/dune_async_io/async_io.ml
emillon: I just did that. the goal was to make it impossible to get an unstarted `t` from the rest of the code.

rgrinberg (approved):

rgrinberg (approved):

Issue #7829, opened 2023-05-29
Specify a specific shell to execute cram test commands


johnridesabike:
## Desired Behavior

Cram tests (and `(system ...)` commands) run in the default system shell. It would be useful to be able to specify a specific shell to use instead.

My specific use case is that I want to guarantee portability while writing tests, but I don't want to assume compatibility with my system shell (on MacOS, usually). Being able to automatically use a shell like `dash` would be an easy way to catch non-portable commands.

## Example

I think it would make sense to expose this option as an environment variable or a CLI flag, since the configuration is local and I probably don't want to commit it in a dune file.

```sh
 DUNE_SHELL=/usr/local/bin/dash dune runtest
```
or
```sh
dune runtest --shell=/usr/local/bin/dash
```

haochenx:
We ran into the same problem today and would like to see this feature.

Also it might be nice to add an option to the `cram` stanza to specify the shell to be used.
Reasonable candidates include
- `:default`, system default shell (the current behavior)
- `:bash`, force bash
- `%{bin:bash}` lookup shell binary

Absolute path might also be a reasonable candidate, but I'm not sure whether absolute paths belongs to dune files

haochenx:
I'm happy to implement this if we can agree on the exact behavior.

maroneze:
I'd also like this feature. We are starting to use more and more Cram tests, and yesterday we had an issue with a Debian machine having a dash too old for `set -o pipefail`, so now we have to choose between changing the Debian version (quite impactful) or adding explicit calls to `bash -c` everywhere in our Cram tests (which is very ugly).

Being able to choose a default shell would help keep our tests future-proof. There will always be some version issues, of course, but we'll have better control over them.

Note that, currently, from what I see in `cram_exec.ml`, it's not even the *user's default shell* that is used, but `sh` explicitly, hardcoded. In the aforementioned Debian machine, users have `bash` by default, but the Cram tests are still using `sh`.

nojb:
I think this request makes sense. Anyway, both `(system)` and CRAM test scripts are never going to be portable across systems, so we might as well provide ways to configure them as best as possible.

A PR would be welcome. (@haochenx: two years later, are you still interested in working on this?)

haochenx:
I certainly don't mind working on it as we still want to see this feature land, but I don't think I could find time to work on it until at least May. So if someone wants it fast then it's better for someone else to work on it. 

Also the previous PR was stuck with disagreements on API decisions, so we might want to fix that first.

PR #8134, opened 2023-07-06
WIP: allow specifying which shell to use for cram test

haochenx:
- related to https://github.com/ocaml/dune/issues/7829


haochenx:
@rgrinberg do you mind to take a look at the design and implementation of this PR

@johnridesabike do you think this approach satisfy your use case?

test/blackbox-tests/test-cases/cram/dune
```diff
@@ -0,0 +1,5 @@
+(cram)
+
+(cram
+ (enabled_if %{bin-available:zsh})
```
Alizter, line 4: If you do this, you will have to add zsh to the CI so we can run the tests. Probably best if you restrict to unix too.

haochenx: I think the macos CI machine should zsh available.

What's the additional benefits to add restriction to unix in addition to checking whether zsh exists?

Alizter: Well actually I'm thinking that we should be using `bin-available`. If we miss putting `zsh` in the CI then this test will never be run, and it might surprise some other devs when they have `zsh` if this test is broken. So we have two choices:

1. Either we make `zsh` mandatory (hence the unix restriction) so that the test is always seen to run.
2. We make a dummy `zsh` or any other shell, just to check how it works. (echo would probably work)
3. We don't test zsh.

haochenx: I see. I did have similar feeling when I added the test, but I ended up with adding them since I wanted to have some non-artificial tests.

But given that your concern is solid, I am in favor of doing (1) or (3) as you suggested. I don't think (2) is a good option since it seems that dune's cram test executor expects the shell at least have (a) output redirecting and (b) script sourcing features and making a convincing dummy `zsh` above the level of what is already done in `shell-opt.t` requires some significant efforts.

Regarding (1), I'm in general in favor of it, but since it involves changing the CI configuration, I don't think I should make the call.

If the maintainers of dune is not in favor of (1), then I think we should take (3) as missing this test case isn't crucial for the completeness of testing the feature added in this PR.

haochenx: (thank you for taking a look btw!)

johnridesabike:
This current implementation would certainly be helpful, but using an environment variable or CLI argument would be preferable for my use case.

My goal is to check that my tests are shell-agnostic, so I'd like to run them in something like dash locally (or in multiple different shells), but allow other users/CIs run them in whatever they have installed. I wouldn't want to commit a Dune rule that requires using a specific shell.

johnridesabike:
> For example, you can use `(shell %{env:DUNE_CRAM_SHELL=/bin/sh})` to achieve what you described.

Wouldn't this require all users to have that environment variable set in order to run cram tests?

If we can only configure it in a Dune file, then I'll just add a `(shell ...)` stanza locally and not commit it. That's not necessarily a problem, just a little clunky. It's still better than not having the feature at all though.

haochenx:
>> For example, you can use (shell %{env:DUNE_CRAM_SHELL=/bin/sh}) to achieve what you described.
>
> Wouldn't this require all users to have that environment variable set in order to run cram tests?

The `=/bin/sh` part provides the default value (in this case, `/bin/sh`) so I guess not all users need to specify that environment variable.

I agree that being able to specify the shell elegantly via environment variable is a great feature.

I will think about how the spec.

haochenx:
I'm thinking about to add two environment variables that matters for dune:

- `DUNE_CRAM_SYSTEM_SHELL`, which if set, decides the path when `Shell_spec.System_shell` is selected
- `DUNE_CRAM_BASH_SHELL`, which if set, decides the path when `Shell_spec.Bash_shell` is selected

As it comes for determining the actual path for `Shell_spec.System_shell` (and respectively for `Shell_spec.Bash_shell`), the following logic is to be used:
- check whether envar `DUNE_CRAM_SYSTEM_SHELL` exists and being non-empty, if so
  - if the given path is an absolute path, check its existence and use it as the shell program
    - if the path does not exists or is a directory, issue an error
  - if the given path is a simple name, resolve it with `Stdune.Bin.which`
    - if this fails, issue an error
  - otherwise, issue an error
- otherwise, use the current behavior (viz. resolve `"sh"` with `Stdune.Bin.which`)

Comments are welcomed.

Alizter:
I would go for something more general. We have `system` and `bash` actions that users can put in rules. We should just have `DUNE_BASH_SHELL` and `DUNE_SYSTEM_SHELL` configure these also, not just cram.

haochenx:
>I would go for something more general. We have system and bash actions that users can put in rules. We should just have DUNE_BASH_SHELL and DUNE_SYSTEM_SHELL configure these also, not just cram. However that would make the scope of this PR bigger than it needs to be.

Sounds better! But I agree that this change will go beyond the scope of this PR. I'm happy to submit another PR to implement that. If my understanding is correct, the place also need to be modify should be the following?
https://github.com/ocaml/dune/blob/4256431b1b8aa25bf9cac578286a9142c6e1c3cb/src/dune_engine/action_exec.ml#L289-L301

Alizter:
Yes that would be one place, but there are also other internal places (such as when dune talks to VCS or runs the compiler to find config stuff). I am not certain that we want to touch all of these. The design of such a feature would need some thinking. Keeping to just the cram case I think is fine for now, and if we want to generalise then we can. At the moment, I don't have a good enough use case for such broad strokes.

haochenx:
Make sense. But I still think we should probably do the same thing for `(system ..)` / `(bash ..)` and for cram. If we add `DUNE_CRAM_XXX_SHELL` and generalize later, we have to deal with the possibility of `DUNE_CRAM_XXX_SHELL` being different than the generalized config.

What about to have `DUNE_USER_ACTION_SYSTEM/BASH_SHELL` and have them to govern around bond (a) cram test and (b) `(system ..)` and `(bash ..)`?

Alizter:
In any case, I don't think adding the env var should be done in this PR. That can come at a later time.

haochenx:
>In any case, I don't think adding the env var should be done in this PR. That can come at a later time.

Agreed. I'm also not planning doing that in this PR. I will be opening another PR soon regarding env var and we should be able to continue the discussion there.

haochenx:
PR dealing with the environment variables:
- https://github.com/ocaml/dune/pull/8149

CHANGES.md
```diff
@@ -14,6 +14,9 @@ Unreleased
 - Add `dune build --dump-gc-stats FILE` argument to dump Garbage Collection
   stats to a named file. (#8072, @Alizter)
 
+- Add `(shell ..)` option to cram test stanza and user action specification.
```
Alizter, line 17:
```suggestion
- Add `(shell ..)` option to `(cram)` stanza and user action specification.
```

src/dune_lang/action.mli
```diff
@@ -120,7 +120,10 @@ type t =
   | Diff of (String_with_vars.t, String_with_vars.t) Diff.t
   | No_infer of t
   | Pipe of Outputs.t * t list
-  | Cram of String_with_vars.t
+  | Cram of
```
Alizter, line 123: I would make this a tuple rather than a record to keep it more inline with the rest of the entries.

src/dune_lang/action.ml
```diff
@@ -410,7 +419,12 @@ let rec encode =
     List
       (atom (sprintf "pipe-%s" (Outputs.to_string outputs))
       :: List.map l ~f:encode)
-  | Cram script -> List [ atom "cram"; sw script ]
+  | Cram { script; shell_spec } ->
+    let shell_spec =
+      if shell_spec = Shell_spec.default then []
+      else [ List [ atom "script"; Shell_spec.encode shell_spec ] ]
```
Alizter, line 425: This should be `shell` not `script`.

src/dune_lang/action.ml
```diff
@@ -493,7 +507,13 @@ let rec map_string_with_vars t ~f =
   | Diff diff -> Diff (Diff.map diff ~path:f ~target:f)
   | No_infer t -> No_infer (map_string_with_vars t ~f)
   | Pipe (o, ts) -> Pipe (o, List.map ts ~f:(map_string_with_vars ~f))
-  | Cram sw -> Cram (f sw)
+  | Cram { script; shell_spec } ->
+    let shell_spec =
```
Alizter, line 511: Let's put this in `Shell_spec` and call it `map`. The mli should mention how it acts on values.

src/dune_lang/shell_spec.mli
```diff
@@ -0,0 +1,16 @@
+open Dune_sexp
+
```
Alizter, line 2: Add a comment here that this is only about string_with_vars and no expansion of `exec_file_shell` is performed.

src/dune_rules/cram/cram_exec.ml
```diff
@@ -436,24 +456,38 @@ let run ~env ~script lexbuf : string Fiber.t =
...
+    { script = f spec.script; shell_spec = Shell_spec.conv f spec.shell_spec }
 
   let is_useful_to ~distribute:_ ~memoize:_ = true
 
-  let encode script path _ : Dune_lang.t =
-    List [ Dune_lang.atom_or_quoted_string "cram"; path script ]
+  let encode { script; shell_spec } path _ : Dune_sexp.t =
+    let atom = Dune_lang.atom_or_quoted_string in
+    let sh : Dune_lang.t =
+      match shell_spec with
```
Alizter, line 482: Actually rather than having 3 shells here. Why not immediately `which` them? You should also validate the custom shell as actually existing. Something else that comes to mind is that users may want to build their own shells, but then `Bin.which` will not find them. We should probably document this limitation somewhere, however I am not sure if it is even worth addressing since it is such a niche concern.

test/blackbox-tests/test-cases/cram/cram-action-shell-opt.t
```diff
@@ -0,0 +1,158 @@
...
+
+  $ echo '(lang dune 3.10)' > dune-project
+
+  $ cat > dune <<EOF
+  > (rule
+  >   (target foo.cram.corrected)
+  >   (mode (promote (until-clean)))
+  >   (action (cram foo.cram))
+  > )
+  > EOF
```
Alizter, line 11:
```suggestion
  > (rule
  >  (target foo.cram.corrected)
  >  (mode (promote (until-clean)))
  >  (action (cram foo.cram)))
  > EOF
```

Alizter: And for the rest of the stanzas. Usually we stick to a single space indent and leave no brackets on their own lines.

Alizter:
I am also inclined to suggest that you document the `cram` user action, but I am unsure if this was omitted on purpose. @emillon could probably say more.

CHANGES.md
haochenx: addressed in 2a23c30773325a1bf6086f3bfe0678009535603e

src/dune_lang/action.mli
haochenx: In an earlier iteration the `shell_spec` field was `String_with_vars.t` so I made it a record to avoid confusion. Now it makes more sense to make it a tuple for consistency.

addressed in 716b5fcc1

src/dune_lang/action.ml
haochenx: addressed in 166911519

src/dune_lang/action.ml
haochenx: addressed in 179f7daa7

src/dune_lang/shell_spec.mli
haochenx: addressed in 179f7daa78c82f8665d0dbd71c0481af437f000b

test/blackbox-tests/test-cases/cram/cram-action-shell-opt.t
haochenx: addressed in b7777bd90

src/dune_rules/cram/cram_exec.ml
haochenx: Good point. I'll see how to implement it.

test/expect-tests/dune_lang/sexp_tests.ml
```diff
@@ -353,3 +353,37 @@ comment)
...
+
+let cram_script = Dune_lang.String_with_vars.make_text loc "test.t"
+
+let%expect_test _ =
+  print_action (Cram (cram_script, System_shell));
+  [%expect {| (cram test.t) |}]
+
+let%expect_test _ =
+  print_action (Cram (cram_script, Bash_shell));
+  [%expect {| (cram test.t (shell :bash)) |}]
```
Alizter, line 385: This optional argument of the cram action will need some thinking. We don't have optional arguments in any other user action apart from ones that take lists like `progn`, `concurrency` or `cat`.

Alizter: I would be interested to hear @emillon or @rgrinberg's opinions on this however.

src/dune_rules/cram/cram_exec.ml
haochenx: addressed in 071f56c82

haochenx:
>I am also inclined to suggest that you document the cram user action, but I am unsure if this was omitted on purpose.

When adding `cram-action-shell-opt.t`, I'm actually surprised by the behavior of the `(cram)` action, where
- the only thing `(cram foo.tt)` action will do is to generate a `foo.t.corrected` file, and 
  - having a stanza like `(rule (action (cram foo.tt)))` will raise `Error: Rule has no targets specified`
    - `(rule (target foo.tt.corrected) (cram foo.tt))` works fine
  - having a stanza like `(rule (alias runtest) (action (cram foo.tt)))` will result in a correct dune config, but will always succeed even if `foo.tt` and `foo.tt.corrected` differs.
    - it seems that the "correct" way to run tests is to
      - either use two stanzas `(rule (target foo.tt.corrected) (action (cram foo.tt)))` + `(rule (alias runtest) (action (diff foo.tt.corrected foo.tt)))`
      - or use `(progn)` to combine `(cram)` and `(diff)`, like `(rule (alias runtest) (action (progn (cram foo.tt) (diff foo.tt.corrected foo.tt))))`
      - these will fail the test if there is a diff, but `dune promote` does not work
- there seems to be no way to promote `foo.tt.corrected` back to `foo.tt`.

The behavior seems a bit difficult to explain clearly (maybe not even to justify well) in documents to me.

test/blackbox-tests/test-cases/cram/shell-opt-dune-lang-version.t
```diff
@@ -0,0 +1,42 @@
...
+  $ cat > foo.cram <<EOF
+  >   $ echo "foo from foo.cram"
+  > EOF
+
+  $ dune runtest -f
+  File "dune", line 3, characters 25-38:
+  3 |   (action (cram foo.cram (shell :bash)))
+                               ^^^^^^^^^^^^^
+  Error: 'shell' is only available since version 3.10 of the dune language.
+  Please update your dune-project file to have (lang dune 3.10).
```
christinerose, line 41:
```suggestion
  Error: 'shell' is only available since version 3.10 of the Dune language.
  Please update your `dune-project` file to have `(lang dune 3.10)`.
```

christinerose: Not sure if these can be changed in the doc, as I know sometimes its the actual output formatted that way. If this is the case, please feel free to ignore or dismiss.

test/blackbox-tests/test-cases/cram/shell-opt.t
```diff
@@ -0,0 +1,252 @@
+Check that cram tests run without shell option specified
```
christinerose, line 1:
```suggestion
Check that Cram tests run without shell option specified
```

christinerose (commented):
Minor formatting suggestions :)

CHANGES.md
```diff
@@ -14,6 +14,9 @@ Unreleased
 - Add `dune build --dump-gc-stats FILE` argument to dump Garbage Collection
```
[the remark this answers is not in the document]
christinerose: Thanks for letting me know. When I get a GH notification that my review is requested, how do I know which ones to do, if some of these can be done in separate PRs?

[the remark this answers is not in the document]
christinerose: Thank you!

test/blackbox-tests/test-cases/cram/dune
haochenx: How should we reach to a consensus here?

test/blackbox-tests/test-cases/cram/shell-opt-dune-lang-version.t
haochenx: Yes, I believe this is the output of running the test so changing it here will result in this test case to fail.

If a fix is desired, I think an issue should be open and it should be addressed by another PR (as fixing this would be outside of the scope of this PR)

haochenx:
CI error after rebasing main seems to be the same as main
- this PR: https://github.com/ocaml/dune/actions/runs/5514622104/jobs/10054087615?pr=8134
- main: https://github.com/ocaml/dune/actions/runs/5511847828/jobs/10047934514

```
Run opam exec -- make test
./_boot/dune.exe runtest
File "test/blackbox-tests/test-cases/github8[15](https://github.com/ocaml/dune/actions/runs/5514622104/jobs/10054087615?pr=8134#step:11:16)8.t", line 1, characters 0-0:
/usr/local/bin/git --no-pager diff --no-index --color=always -u _build/.sandbox/883412a09d29c61c1a8403e31dfaa09b/default/test/blackbox-tests/test-cases/github8158.t _build/.sandbox/883412a09d29c61c1a8403e31dfaa09b/default/test/blackbox-tests/test-cases/github8158.t.corrected
diff --git a/_build/.sandbox/883412a09d29c61c1a8403e31dfaa09b/default/test/blackbox-tests/test-cases/github8158.t b/_build/.sandbox/883412a09d29c61c1a8403e31dfaa09b/default/test/blackbox-tests/test-cases/github8158.t.corrected
index 0e58359..54c0530 100644
--- a/_build/.sandbox/883412a09d29c61c1a8403e31dfaa09b/default/test/blackbox-tests/test-cases/github8158.t
+++ b/_build/.sandbox/883412a09d29c61c1a8403e31dfaa09b/default/test/blackbox-tests/test-cases/github8158.t.corrected
@@ -37,6 +37,6 @@ This fails, but should work (and print nothing):
 
   $ dune exec bin/e.exe 2>&1 | sed -e 's/camlppx....../camlppxXXXXXX/g' -e 's/build_......_dune/build_XXXXXX_dune/g'
   File "bin/.e.eobjs/_unknown_", line 1, characters 0-0:
-  sh: 1: ../.ppx/059965dd2a[17](https://github.com/ocaml/dune/actions/runs/5514622104/jobs/10054087615?pr=8134#step:11:18)61d0e00c111505aef5ac/ppx.exe: not found
+  sh: ../.ppx/059965dd2a1761d0e00c111505aef5ac/ppx.exe: No such file or directory
   File "bin/e.ml", line 1:
-  Error: I/O error: ../.ppx/059965dd2a1761d0e00c111505aef5ac/ppx.exe --as-ppx '/tmp/build_XXXXXX_dune/build_XXXXXX_dune/camlppxXXXXXX' '/tmp/build_XXXXXX_dune/build_XXXXXX_dune/camlppxXXXXXX'
+  Error: I/O error: ../.ppx/059965dd2a1761d0e00c111505aef5ac/ppx.exe --as-ppx '/var/folders/[24](https://github.com/ocaml/dune/actions/runs/5514622104/jobs/10054087615?pr=8134#step:11:25)/8k48jl6d249_n_qfxwsl6xvm0000gn/T/build_XXXXXX_dune/build_XXXXXX_dune/camlppxXXXXXX' '/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/build_XXXXXX_dune/build_XXXXXX_dune/camlppxXXXXXX'
```

Alizter:
@haochenx Don't worry about that error. A test demonstrating an issue #8158 and a fix #8159 were merged in succession however the failure in the test was not 100% reproducible but went away after the fix.

Alizter:
By the way, during last weeks dev meeting we discussed this feature: https://github.com/ocaml/dune/wiki/dev-meeting-20230712

I am not certain if we drew any conclusions however, perhaps @emillon can say more.

haochenx:
@Alizter Thank you for the info! I have look at the meeting notes and here are some comments of mine:

>on windows, it makes us rely on cygwin

Not completely sure, but I think the current implementation in a way expect a *nix like shell so we probably are already relying on cygwin. (and I think there are some code that detects windows and performs cygwin-specific path translation)

>possible way to make smaller feature: just allow bash?

Totally reasonable. I won't mind shredding this PR down to this.
Also it would (hypothetically) prevent writing portable tests for minimal environment (e.g. busybox)

Though I don't think taking the step to allow arbitrary shell will greatly decrease the complexity of this PR..

**Overall** I do feel this PR is unexpectedly a large diff for its scope and I am in favor of reducing the feature (i.e. restrict it to "bashism" as it's the most useful feature this PR introduce for our use case)

Alizter:
I think one thing that wasn't in the notes was the worry about giving users a foot gun with being able to select shells. I think we will probably discuss it again at the next meeting to understand what exactly we want out of this feature, since that wasn't clear for us.

haochenx:
Make sense. I can definitely wait. 

For your reference at least the motivation problem we got is that since currently cram test uses `which sh` it's is undecided what shell you get.

On our dev machines (mostly macOS) we get bash but on our CI machine we get csh. And that is causing our test cases to fail.

rgrinberg:
There's no problem with being able to select the shell. Portability isn't a universal concern and it's not the job of the build system to force it down our users. Bash isn't the be-all and end-all of shells and there's plenty of reasons to want to use something else.

I don't really like the configuration introduced in this PR though. I would imagine this feature to be configurable through the `cram` stanza. So that it's possible to set once for the entire test suite. Allowing to be customizable in the action is hardly useful in comparison.

Also, the special symbols for bash and system seem gratuitous. Is providing a binary name looking it up in PATH doesn't work for your use case?

haochenx:
re: https://github.com/ocaml/dune/pull/8134#issuecomment-1646859986

First of all, I'm open for syntactical suggestions. To answer your questions and provide more contexts:

> I would imagine this feature to be configurable through the cram stanza. So that it's possible to set once for the entire test suite.

It's possible and actually the main motivation for this PR.

> Allowing to be customizable in the action is hardly useful in comparison.

Allowing the customization in the user action is a bonus for completeness of implementation.
I wasn't [even aware](https://github.com/ocaml/dune/pull/8134#:~:text=seems%20like%20an%20undocumented%20feature%3F) about this feature before starting working on this PR.
I was thinking about omitting it at the beginning, but ended up implementing the same level of support just for the completeness.

I'd be glad to drop the user action support if that helps reaching an agreement and accelerate the review of this PR.

>Also, the special symbols for bash and system seem gratuitous. Is providing a binary name looking it up in PATH doesn't work for your use case?

Without special symbols, I'd imagine one'd use `(shell bash)` for bash and `(shell system)` for the system shell.
This raise ambiguity though since we don't know whether we should use dune's dependency language to expand `bash` (or `system`, respectively) or should lookup `bash` (or `sh`, respectively) using `which`.

I opted for special symbols to avoid the ambiguity in syntax.

- For `:bash`, one can definitely use `%{bin:bash}` instead so it's a clear favoritism towards bash. But this has been in the other places in dune, and I think wanting bash is a common enough motivating use case for this feature, so this favoritism is justifiable.
  Nevertheless I'm not too against getting rid of it.
- For `:system`, it's a bit trickier. We could convince ourselves to assume the system shell will always be equivalent to `%{bin:sh}`, but if (hypothetically) e.g. we add "proper" support for windows and use `cmd` when this option is not present, then the assumption does not hold. In this situation we do not have a way to explicitly specify the "default" behavior.

  Maybe trading `:system` with `:default` is more palatable to you. I do not really have a strong preference here.

  I chose `:system` merely because there is a user action called `system`. There is also another user action called `bash`, which made me thought choosing `:bash` and `:system` creates some good consistency.

haochenx:
>There's no problem with being able to select the shell. Portability isn't a universal concern and it's not the job of the build system to force it down our users. Bash isn't the be-all and end-all of shells and there's plenty of reasons to want to use something else.

I believe I'm holding the same view in general. As for giving bash special treatment, I think it's a good trade-off for our current time. It's almost everywhere, and you'd get a good deal of portability when specifying it over gambling what the system shell would be.

So I think it's useful to give bash some favoritism, but at the same time not against getting rid of its special treatments.

test/expect-tests/dune_lang/sexp_tests.ml
rgrinberg: How about you just introduce a different action for a cram action that uses a different shell? E.g. `cram_with_shell`?

test/blackbox-tests/test-cases/cram/cram-action-shell-opt.t
```diff
@@ -0,0 +1,153 @@
...
+
+  $ cat > foo.cram <<EOF
+  >   $ echo "foo from foo.cram"
+  > EOF
+
+  $ dune build
+  sh.ml
+  $ cat foo.cram.corrected
+    $ echo "foo from foo.cram"
+    ***** UNREACHABLE *****
```
rgrinberg, line 77: Why is this and the rest unreachable?

rgrinberg:
The problem with the special symbols is twofold:

* The dash and oil shells are just as useful as zsh and bash.
* You're not allowing users to customize the command line arguments for running the shell

> This raise ambiguity though since we don't know whether we should use dune's dependency language to expand bash (or system, respectively) or should lookup bash (or sh, respectively) using which.

Do you mean the dependency language or for allowing percent forms? I assume you meant the latter because that's the only thing that makes sense in this context. I think we should allow for expanding dune's percent forms and we should also allow command line arguments as well. For example, one should be able to:

```
(cram
 (shell %{bin:dash} -v %{script})) ;; %{script} will be expanded to the path of the shell script. like %{input-file} for ppx.
```

Dune has a single way of resolving binaries so the question regarding `which` are automatically resolved. E.g. see how `Run` is expanded in `action_unexpanded.ml`.

Once this is resolved and you remove the optional argument, this PR is good to go.

test/blackbox-tests/test-cases/cram/cram-action-shell-opt.t
emillon: My guess is that when creating `foo.cram`, the `$` is interpreted by the parent shell. You can use `<< 'EOF'` line 69 to prevent this.

christinerose (approved):

haochenx:
@rgrinberg @emillon  thanks for having a look and point out a direction. I will take sometime later this week to re-look at this issue and bring it forward.

haochenx:
Unfortunately I didn't get time to work on this PR last week.. I will see whether I can find some time this weekend.

CHANGES.md
```diff
@@ -106,6 +109,7 @@ changelog properly.
 - Add `dune build --dump-gc-stats FILE` argument to dump garbage collection
   stats to a named file. (#8072, @Alizter)
 
+=======
```
haochenx, line 112:
```suggestion
```

test/blackbox-tests/test-cases/cram/cram-action-shell-opt.t
haochenx: These tests try to see whether the PATH environment variable is regarded by making our own `bash` and `sh` compiled from `bash.ml` and `sh.ml`. But since those are nowhere near complete shell implementation, they do not handle stdout redirection directives (like `echo hello > file.txt`), which cram_exec relies on to capture the output. Therefore the cram test executor will think the cram test crashed and include `***** UNREACHABLE *****` here.

We know our own mocked version of `bash` and `sh` are invoked here by looking at https://github.com/ocaml/dune/pull/8134/files/e2913128ca6046e24b2dbfa7255be40588ce58d1#diff-0484438795287a34ad56a4a530a816caea0a5480e23fffb006408f39591b35e6R74 and https://github.com/ocaml/dune/pull/8134/files/e2913128ca6046e24b2dbfa7255be40588ce58d1#diff-0484438795287a34ad56a4a530a816caea0a5480e23fffb006408f39591b35e6R95

haochenx: This is less than ideal but I didn't come up with other ways to perform this test.

otherlibs/dune-rpc/private/types.ml
```diff
@@ -27,7 +27,7 @@ end
 module Version = struct
   type t = int * int
 
-  let latest = 3, 11
+  let latest = 3, 12
```
haochenx, line 30: not sure whether this should be in this PR

src/dune_rules/cram/cram_rules.ml
```diff
@@ -146,15 +155,56 @@ let rules ~sctx ~expander ~dir tests =
...
+                let+ shell_args = Action_builder.all (List.map args ~f:expand_arg)
+                and+ shell_prog =
+                  Expander.(
+                    With_deps_if_necessary.expand_single_path expander prog
+                    |> Deps.action_builder)
+                in
+                shell_prog, List.concat shell_args
+              in
+              return ~shell_deps ~sandbox (Some shell)
+          in
```
haochenx, line 199: not sure whether this is the best way to expand the prog path and arg here

haochenx:
I have addressed https://github.com/ocaml/dune/pull/8134#issuecomment-1712497305 except for `%{script}`, which I find a bit difficult to do.

@rgrinberg do you think `%{script}` is necessary?

otherlibs/dune-rpc/private/types.ml
rgrinberg: a separate PR would be best.

rgrinberg:
> do you think %{script} is necessary?

I can give it a try myself

haochenx:
@rgrinberg regarding `%{script}`, I could probably attempt if you could kindly
(1) point me to code that demonstrates how it could be implement
(2) confirm the exact semantics that we want: I assume that we should stop appending the `%{script}` automatically when one is specified

rgrinberg:
Thanks @haochenx.

1. You need to use the `Expander.add_bindings` function to add a script binding. See how we handling `%{input-file}` it should be pretty similar. Maybe we should just call it `%{input-file}` as well. I wonder what you think about that.
2. That's correct.

haochenx:
@rgrinberg thanks for the pointers. last time I tried to grep for `input-file` to find code examples, but the logic involved was quite complicated and not directly referable. I will check again.

Using `Expander.add_bindings` to add the binding is one thing, but suppressing the default behavior is another. When attempting, I will see which of the following approach fit the code base the best, but please review what I may do:

> Maybe we should just call it %{input-file} as well. I wonder what you think about that.

My take is that it will be confusing for the users, as it's not strictly the "input file", as the script executed is quite heavily processed snippets of the original `.t` test description files. So sticking to `%{script}` is a better idea IMO.

(I estimate that I will attempt to implement `%{script}` in a couple of days, but no guarantee.)

rgrinberg:
We ended up providing `bash` as an option. Not as general as this PR, but enough to cover our needs. Feel free to extend what we have to add more shells.

PR #8145, opened 2023-07-08
chore: dump-stats cleanups

rgrinberg:
Count stats immediately after compacting

Do not rely on prefix in tests


test/blackbox-tests/test-cases/dump-gc-stat.t
```diff
@@ -1,7 +1,7 @@
-Testing the --dump-gc-stat option
```
Alizter, line 1: I was confused why this would have worked, but I think cmdliner was completing it.

Alizter (approved):

PR #8147, opened 2023-07-08
refactor(memo): remove [node_info]

rgrinberg:
The distinction between [node] and [node_info] wasn't used anywhere.
Once it's removed, we gain two things:

* One less pointer to chase on all node operations
* Save two additional words of memory for every node

### DIFF
--- a/src/dag/dag.ml
+++ b/src/dag/dag.ml
@@ -12,23 +12,19 @@ module Make (Value : Value) () : S with type value := Value.t = struct
 
     module Id = Id.Make ()
 
-    type node_info =
+    type node =
       { id : Id.t
       ; mutable mark : mark
       ; mutable level : int
       ; mutable deps : node list
       ; mutable rev_deps : node list
       ; mutable parent : node option
-      }
-
-    and node =
-      { data : Value.t
-      ; info : node_info
+      ; value : Value.t
       }
 
     type vertex = node
 
-    let node_id { info; _ } = info.id
+    let node_id { id; _ } = id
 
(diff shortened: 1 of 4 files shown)


snowleopard (commented):
Thanks! I am going to benchmark this change internally (just curious if the improvement is visible).

src/dag/dag.ml
```diff
@@ -12,23 +12,19 @@ module Make (Value : Value) () : S with type value := Value.t = struct
...
       ; mutable level : int
       ; mutable deps : node list
       ; mutable rev_deps : node list
       ; mutable parent : node option
-      }
-
-    and node =
-      { data : Value.t
-      ; info : node_info
+      ; data : Value.t
```
snowleopard, line 22: Let's call the field `value`? (Or rename the functor parameter to `Data`?)

Alizter:
Our benchmarks don't appear to show any significant change. (We recently added GC stats to the bench). I'd be interested to know if the JS monorepo turns it into something significant.

snowleopard:
> We recently added GC stats to the bench

That's great!

I'll report back what I learn internally. Probably within noise too. (The benchmark is already running.)

rgrinberg:
This saves around 0.4 mb in the GC benchmarks. Pretty small, but it also makes the code cleaner IMO.

snowleopard:
I'm seeing consistent small (0.1-0.3%) improvements in promotions from minor to major heap across a bunch of benchmarks.

snowleopard (approved):

PR #8148, opened 2023-07-08
feature: add gc stats to --trace-file

Alizter:
- [x] doc
- [x] changelog
- [x] update tests


rgrinberg:
Could you finish this?

Alizter:
@rgrinberg I've refactored a little, updated the documentation and added a changelog entry.

Alizter:
Seems to be working correctly:
image

src/dune_util/gc.ml
```diff
@@ -94,3 +94,49 @@ let decode =
...
+    ; "live_blocks", `Int live_blocks
+    ; "free_words", `Int free_words
+    ; "free_blocks", `Int free_blocks
+    ; "largest_free", `Int largest_free
+    ; "fragments", `Int fragments
+    ; "compactions", `Int compactions
+    ; "top_heap_words", `Int top_heap_words
+    ; "stack_size", `Int stack_size
+    ]
+  in
```
rgrinberg, line 138: Can this be the same as `Dune_stats.record_gc_and_fd`

src/dune_engine/scheduler.ml
```diff
@@ -365,7 +365,10 @@ end = struct
     end
 
     let next q =
-      Option.iter q.stats ~f:Dune_stats.record_gc_and_fd;
+      Option.iter q.stats ~f:(fun stats ->
+        Dune_stats.record_fd stats;
+        (* Only [Gc.quick_stat] is querried since it is much faster. *)
+        Dune_stats.emit stats (Dune_util.Gc.event (Gc.quick_stat ())));
```
Alizter, line 371: I've made the fact that we are using `Gc.quick_stat` here more explicit.

src/dune_util/gc.ml
Alizter: I've removed the gc event from that function and replaced it's usage in Scheduler with the new one in Dune_util. The only remaining difference is that the one in scheduler does a quick_stat rather than a stat.

Alizter:
@rgrinberg This has been finished for some time, please decide if you want to keep it otherwise I will close.

PR #8151, opened 2023-07-09
refactor: simplify info type in rule

rgrinberg:
The current representation of `Rule.Info.t` is:

```
  type t =
    | From_dune_file of Loc.t
    | Internal
    | Source_file_copy of Path.Source.t
```

This representation contains redundant information:

* The location is already present as a field in `Rule.t` hence the `Loc.t` argument is unnecessary in `From_dune_file`.
* If we know the rule is copying from the source, then the source path can be easily derived from the target file.

I've removed both of those arguments and now compute their equivalents from `Rule.t` fields

### DIFF
--- a/src/dune_engine/load_rules.ml
+++ b/src/dune_engine/load_rules.ml
@@ -139,18 +139,18 @@ let describe_rule (rule : Rule.t) =
   Pp.text
   @@
   match rule.info with
-  | From_dune_file loc ->
-    let start = Loc.start loc in
+  | From_dune_file ->
+    let start = Loc.start rule.loc in
     start.pos_fname ^ ":" ^ string_of_int start.pos_lnum
   | Internal -> "<internal location>"
-  | Source_file_copy _ -> "file present in source tree"
+  | Source_file_copy -> "file present in source tree"
 ;;
 
 let report_rule_src_dir_conflict dir fn (rule : Rule.t) =
   let loc =
     match rule.info with
-    | From_dune_file loc -> loc
-    | Internal | Source_file_copy _ ->
+    | From_dune_file -> rule.loc
+    | Internal | Source_file_copy ->
       let dir =
         match Path.Build.drop_build_context dir with
         | None -> Path.build dir
@@ -183,7 +183,7 @@ let report_rule_conflict fn (rule' : Rule.t) (rule : Rule.t) =
     ]
     ~hints:
       (match rule.info, rule'.info with
-       | Source_file_copy _, _ | _, Source_file_copy _ ->
+       | Source_file_copy, _ | _, Source_file_copy ->
          [ Pp.textf
              "rm -f %s"
              (Path.to_string_maybe_quoted (Path.drop_optional_build_context fn))
@@ -351,7 +351,7 @@ end = struct
       in
       Rule.make
         ~context:None
-        ~info:(Source_file_copy path)
+        ~info:Source_file_copy
         ~targets:(Targets.File.create ctx_path)
         build)
   ;;
--- a/src/dune_engine/rule.ml
+++ b/src/dune_engine/rule.ml
@@ -3,19 +3,14 @@ module Action_builder = Action_builder0
 
 module Info = struct
   type t =
-    | From_dune_file of Loc.t
+    | From_dune_file
     | Internal
-    | Source_file_copy of Path.Source.t
-
-  let of_loc_opt = function
-    | None -> Internal
-    | Some loc -> From_dune_file loc
-  ;;
+    | Source_file_copy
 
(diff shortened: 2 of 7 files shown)


snowleopard:
> refactor: remove source path in rule info

> It's redundant because there's already a `loc` field in rules.

There seems to be some confusion between the title and description of this PR. You're in fact deleting both of them, and while `loc` indeed exists as a separate field, the source path doesn't.

On a more general note, having both `info` and `loc` seems pretty redundant to me. Removing the `loc` hidden inside the `info` (as this PR does) may be an improvement, but then maybe "info" should become "kind", "provenance" or something similar (since `loc` is also some information about the rule)?

src/dune_rules/super_context.ml
```diff
@@ -258,8 +258,12 @@ let extend_action t ~dir build =
 let make_rule t ?mode ?loc ~dir { Action_builder.With_targets.build; targets } =
   let build = extend_action t build ~dir in
   Rule.make
+    ?loc
     ?mode
-    ~info:(Rule.Info.of_loc_opt loc)
+    ~info:
+      (match loc with
+       | None -> Internal
+       | Some _ -> From_dune_file)
```
snowleopard, line 266: You wrote this as `if Loc.is_none loc then Internal else From_dune_file` and slightly differently here. Should we factor out this piece of code?

rgrinberg: Mostly because `loc` is option here while it's not where I used `Loc.is_none`. I could give the optional value a default here and unify though. E.g.

```
let make_rule ?(loc=Loc.none) .. =
  let info = if Loc.is_none then Internal else From_dune_file
```

I'm not sure how to exactly factor out this piece of code. I guess one option would be to have `Rule.make` infer the `info` if it's not provided from the location?

rgrinberg:
> There seems to be some confusion between the title and description of this PR. You're in fact deleting both of them, and while loc indeed exists as a separate field, the source path doesn't. Could you clean this up a little?

Clarified. I changed the description to just describe what the PR is doing.

> On a more general note, having both info and loc seems pretty redundant to me. Removing the loc hidden inside the info (as this PR does) may be an improvement, but then maybe "info" should become "kind", "provenance" or something similar (since loc is also some information about the rule)?

I don't mind the name `info` personally. Although `provenance` or `source` also seem fine to me.

src/dune_rules/super_context.ml
snowleopard: Ah, I didn't notice they were slightly different. Let's keep things as is then.

src/dune_engine/rule.ml
```diff
@@ -103,13 +98,20 @@ let make ?(mode = Mode.Standard) ~context ?(info = Info.Internal) ~targets actio
...
   in
   let loc =
-    match info with
-    | From_dune_file loc -> loc
-    | Internal ->
-      Loc.in_file
-        (Path.drop_optional_build_context
-           (Path.build (Path.Build.relative dir "_unknown_")))
-    | Source_file_copy p -> Loc.in_file (Path.source p)
+    match loc with
```
snowleopard, line 101: It feels a little odd to me to have this logic on the inside of `Rule.make`. It makes the logic somewhat hidden/magic from the point of view of callers.

One alternative would be to make the `loc` argument required, and provide functions `internal_rule_loc : dir:Path.t -> Loc.t` and `copy_rule_loc : Path.Source.t -> Loc.t` that factor this hidden functionality out so that it can be used explicitly when needed.

Not a very strong opinion but the current API seems odd to me.

src/dune_engine/load_rules.ml
```diff
@@ -139,18 +139,18 @@ let describe_rule (rule : Rule.t) =
...
+  | Source_file_copy -> "file present in source tree"
 ;;
 
 let report_rule_src_dir_conflict dir fn (rule : Rule.t) =
   let loc =
     match rule.info with
-    | From_dune_file loc -> loc
-    | Internal | Source_file_copy _ ->
+    | From_dune_file -> rule.loc
+    | Internal | Source_file_copy ->
```
snowleopard, line 153: If we have `rule.loc`, why do we look at `rule.info` to come up with some other `loc` here?

snowleopard (approved):
This change seems like an improvement but only marginal. I think the current `loc`/`info` API is pretty confusing and the code that looks at these fields is confusing as well. I would prefer to come up with a better design instead of merging this incremental PR but I'll let you decide what's best.

rgrinberg:
Agreed, I'm honestly not too happy about the end result. Let me think a little more on how to improve this and I'm not even sure if this PR is a marginal improvement in readability.

Issue #7308, opened 2023-03-13
Enable running two builds in separate terminals (a.k.a. `dune monitor`)


jchavarri:
https://github.com/ocaml/dune/pull/6360 disabled multiple build commands running concurrently.

However, as dune expands its scope and supports more tools, the case for running `dune build` with two or more apps in separate terminals gains strength.

For the particular use case I can speak of, at Ahrefs we have a web server written in native OCaml, and a frontend built with Melange. It is desirable to track the build status of both apps on separate terminals to not mix errors, and also because each environment might have different binaries that are executed on a successful build (e.g. the web server binary is relaunched after every build succeeds, same for webpack or other JS bundlers on the Melange side).

@rgrinberg [suggested](https://github.com/ocaml/dune/issues/6967#issuecomment-1419762781) to have a new subcommand `monitor` that would help mirror the output of a running dune process. Then, add to `dune build` the ability to delegate a job, and then attach `dune monitor` until the job is finished.

I wonder, would it be possible to solve this problem without adding a new subcommand? I.e. when `dune build` sees that another dune process is already running, it can request it to build some target, and get the output back. Essentially making the lock system transparent for users.

Another idea to sidestep the new subcommand would be to make the lock system more granular, and instead of locking the whole tree, lock the specific folders that the current build is targeting. Not sure if this is possible though 🤔

Alizter:
Currently we have no way of sending (error) messages via RPC. So if there is a Dune in watch mode, having another Dune connect by a user will mean they won't see the messages.

rgrinberg:
> I wonder, would it be possible to solve this problem without adding a new subcommand? I.e. when dune build sees that another dune process is already running, it can request it to build some target, and get the output back. Essentially making the lock system transparent for users.

Getting the output back is exactly what implementing `dune monitor` is about. Currently, we have no way of getting any useful output from a dune instance running in watch mode.

> Another idea to sidestep the new subcommand would be to make the lock system more granular, and instead of locking the whole tree, lock the specific folders that the current build is targeting. Not sure if this is possible though

It's not possible to do it per directory because there's some global database files that aren't per directory. It would be possible for the lock to be released by watch mode when it's idle though.

jchavarri:
> Getting the output back is exactly what implementing `dune monitor` is about. Currently, we have no way of getting any useful output from a dune instance running in watch mode.

[...]

PR #8152, opened 2023-07-09
feature: dune monitor

Alizter:
This PR adds the `dune monitor` command which allows the user to connect to an already running `dune build` in watch mode and view the build progress and status. It does so by connecting to a running RPC server in a workspace and listening for progress changes.

- [x] changelog
- [x] tests
- fixes https://github.com/ocaml/dune/issues/7308
- [ ] Fix formatting of error messages (something lost during RPC)
  - [ ] #8411
- [ ] Understand issue with signal watcher
- [x] RPC jobs bug
  - [x] #8410
  - [x] #8409   
- [x] #8193 
- [x] #8192
- [x] #8191
- [x] #8212 

This work is still experimental and may be subject to further changes.

In the future, all `dune build` commands may have behaviour similar to `dune monitor` allowing for multiple builds in the same workspace, all sharing the same build server. We also hope to further improve the TUI allowing for a seamless experience for users.

### DIFF
--- a/bin/monitor.ml
+++ b/bin/monitor.ml
@@ -1,0 +1,268 @@
+open Import
+open Fiber.O
+module Client = Dune_rpc_client.Client
+module Version_error = Dune_rpc_private.Version_error
+
+include struct
+  open Dune_rpc
+  module Diagnostic = Diagnostic
+  module Progress = Progress
+  module Job = Job
+  module Sub = Sub
+  module Conv = Conv
+end
+
+(** Utility module for generating [Map] modules for [Diagnostic]s and [Job]s which use
+    their [Id] as keys. *)
+module Id_map (Id : sig
+    type t
+
+    val compare : t -> t -> Ordering.t
+    val sexp : (t, Conv.values) Conv.t
+  end) =
+struct
+  include Map.Make (struct
+      include Id
+
+      let to_dyn t = Sexp.to_dyn (Conv.to_sexp Id.sexp t)
+    end)
+end
+
+module Diagnostic_id_map = Id_map (Diagnostic.Id)
+module Job_id_map = Id_map (Job.Id)
+
+module Event = struct
+  (** Events that the render loop will process. *)
+  type t =
+    | Diagnostics of Diagnostic.Event.t list
+    | Jobs of Job.Event.t list
+    | Progress of Progress.t
+end
+
+module State : sig
+  (** Internal state of the render loop. *)
+  type t
+
+  (** Initial empty state. *)
+  val init : unit -> t
+
+  module Update : sig
+    (** Incremental updates to the state. Computes increments of the state that
+        will be used for efficient rendering. *)
+    type t
+  end
+
+  val update : t -> Event.t -> Update.t
+
+  (** Given a state update, render the update. *)
+  val render : t -> Update.t -> unit
+end = struct
+  type t =
+    { mutable diagnostics : Diagnostic.t Diagnostic_id_map.t
+    ; mutable jobs : Job.t Job_id_map.t
+    ; mutable progress : Progress.t
+    }
+
+  let init () =
+    { diagnostics = Diagnostic_id_map.empty; jobs = Job_id_map.empty; progress = Waiting }
+  ;;
+
+  let done_status ~complete ~remaining ~failed state =
+    Pp.textf
+      "Done: %d%% (%d/%d, %d left%s) (jobs: %d)"
+      (if complete + remaining = 0 then 0 else complete * 100 / (complete + remaining))
+      complete
+      (complete + remaining)
+      remaining
+      (match failed with
+       | 0 -> ""
+       | failed -> sprintf ", %d failed" failed)
+      (Job_id_map.cardinal state.jobs)
+  ;;
+
+  let waiting_for_file_system_changes message =
+    Pp.seq message (Pp.verbatim ", waiting for filesystem changes...")
+  ;;
+
+  let restarting_current_build message =
+    Pp.seq message (Pp.verbatim ", restarting current build...")
+  ;;
+
+  let had_errors state =
+    match Diagnostic_id_map.cardinal state.diagnostics with
+    | 1 -> Pp.verbatim "Had 1 error"
+    | n -> Pp.textf "Had %d errors" n
+  ;;
+
+  let status (state : t) =
+    Console.Status_line.set
+      (Live
+         (fun () ->
+           match (state.progress : Progress.t) with
+           | Waiting -> Pp.verbatim "Initializing..."
+           | In_progress { complete; remaining; failed } ->
+             done_status ~complete ~remaining ~failed state
+           | Interrupted ->
+             Pp.tag User_message.Style.Error (Pp.verbatim "Source files changed")
+             |> restarting_current_build
+           | Success ->
+             Pp.tag User_message.Style.Success (Pp.verbatim "Success")
+             |> waiting_for_file_system_changes
+           | Failed ->
+             Pp.tag User_message.Style.Error (had_errors state)
+             |> waiting_for_file_system_changes))
+  ;;
+
+  module Update = struct
+    type t =
+      | Update_status
+      | Add_diagnostics of Diagnostic.t list
+      | Refresh
+
+    let jobs state jobs =
+      let jobs =
+        List.fold_left jobs ~init:state.jobs ~f:(fun acc job_event ->
+          match (job_event : Job.Event.t) with
+          | Start job -> Job_id_map.add_exn acc job.id job
+          | Stop id -> Job_id_map.remove acc id)
+      in
+      state.jobs <- jobs;
+      Update_status
+    ;;
+
+    let progress state progress =
+      state.progress <- progress;
+      Update_status
+    ;;
+
+    let diagnostics state diagnostics =
+      let mode, diagnostics =
+        List.fold_left
+          diagnostics
+          ~init:(`Add_only [], state.diagnostics)
+          ~f:(fun (mode, acc) diag_event ->
+            match (diag_event : Diagnostic.Event.t) with
+            | Remove diag -> `Remove, Diagnostic_id_map.remove acc diag.id
+            | Add diag ->
+              ( (match mode with
+                 | `Add_only diags -> `Add_only (diag :: diags)
+                 | `Remove -> `Remove)
+              , Diagnostic_id_map.add_exn acc diag.id diag ))
+      in
+      state.diagnostics <- diagnostics;
+      match mode with
+      | `Add_only update -> Add_diagnostics (List.rev update)
+      | `Remove -> Refresh
+    ;;
+  end
+
+  let update state (event : Event.t) =
+    match event with
+    | Jobs jobs -> Update.jobs state jobs
+    | Progress progress -> Update.progress state progress
+    | Diagnostics diagnostics -> Update.diagnostics state diagnostics
+  ;;
+
+  let render =
+    let f d = Console.print_user_message (Diagnostic.to_user_message d) in
+    fun (state : t) (update : Update.t) ->
+      (match (update : Update.t) with
+       | Add_diagnostics diags -> List.iter diags ~f
+       | Update_status -> ()
+       | Refresh ->
+         Console.reset ();
+         Diagnostic_id_map.iter state.diagnostics ~f);
+      status state
+  ;;
+end
+
+(* A generic loop that continuously fetches events from a [sub] that it opens a
+   poll to and writes them to the [event] bus. *)
+let fetch_loop ~(event : Event.t Fiber_event_bus.t) ~client ~f sub =
+  Client.poll client sub
+  >>= function
+  | Error version_error ->
+    let* () = Fiber_event_bus.close event in
+    User_error.raise [ Pp.verbatim (Version_error.message version_error) ]
+  | Ok poller ->
+    let rec loop () =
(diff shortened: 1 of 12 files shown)


bin/monitor.ml
```diff
@@ -0,0 +1,131 @@
...
+      and* jobs_poller =
+        poll_exn client Dune_rpc_private.Public.Sub.running_jobs
+      and* diagnostics_poller =
+        poll_exn client Dune_rpc_private.Public.Sub.diagnostic
+      in
+      Fiber.all_concurrently_unit
+        [ job_loop jobs_poller
+        ; progress_loop progress_poller
+        ; diagnostic_loop diagnostics_poller
+        ])
```
rgrinberg, line 120: This means that you will block until all 3 will have a pending message. You need to have them write to an mvar all concurrently.

bin/monitor.ml
```diff
@@ -0,0 +1,131 @@
...
+
+let poll_exn ?id client sub =
+  let+ poller = Client.poll ?id client sub in
+  match poller with
+  | Error e -> raise (Dune_rpc_private.Version_error.E e)
+  | Ok poller -> poller
+
+(** Shared state *)
+
+let job_number : int Fiber.Svar.t = Fiber.Svar.create 0
```
rgrinberg, line 45: You shouldn't create any svars

bin/monitor.ml
Alizter: How so? Are they all not looping independently?

rgrinberg: Ah yes, I misread this `foo_loop` function.

bin/monitor.ml
```diff
@@ -0,0 +1,131 @@
...
+
+let rec progress_loop (progress_poller : Dune_rpc.Progress.t Client.Stream.t) =
+  let* progress = next_exn progress_poller in
+  let job_number = Fiber.Svar.read job_number in
+  status progress job_number;
+  maybe_reset_diagnostics progress >>> progress_loop progress_poller
+
+let rec job_loop (jobs_poller : Dune_rpc.Job.Event.t list Client.Stream.t) =
+  let* jobs = next_exn jobs_poller in
+  Fiber.Svar.write job_number (List.length jobs) >>> job_loop jobs_poller
```
rgrinberg, line 79: `jobs` are start/stop events for jobs. So you need to keep an internal set of running jobs to know how many are running.

Alizter: Yes, I've noticed this. I was wondering why the running jobs count was differing so much.

bin/monitor.ml
Alizter: Removed svars.

bin/monitor.ml
Alizter: I've changed it to increment and decrement the job count now.

bin/monitor.ml
```diff
@@ -0,0 +1,197 @@
...
+  let progress p = Progress p
+
+  let jobs l = Jobs l
+end
+
+module State = struct
+  (** Internal state of the render loop. *)
+
+  type t =
+    { mutable quit : bool
```
rgrinberg, line 31: Why do you need this flag? Isn't having the event in the mvar enough?

rgrinberg:
> Make build requests

I guess this should be part of `dune build`? It should be dispatching the build to the running instance and then attaching a monitor until the build completes.

bin/monitor.ml
Alizter: If we don't finish reading all the events, the other loops will get stuck trying to write their events.

rgrinberg: Okay, I see. In that case, you should also make sure you aren't sending events past when `quit` is set in the `fetch_loop`.

bin/monitor.ml
```diff
@@ -0,0 +1,230 @@
...
+        | Remove d ->
+          state.diagnostics <- Diagnostic.Set.remove state.diagnostics d;
+          Console.reset ();
+          State.render_state state);
+      loop ()
+    (* If the poller disconnected in one of the writer loops then ping the
+       server. If it is not responding then we know the server must have
+       disconnected and we can quit. *)
+    | Poller_disconnected ->
+      (* ping the server to make sure it's still alive *)
```
rgrinberg, line 160: Is this even possible?

Alizter: This happens when the server shuts down for instance. If I have a watch mode dune and a monitor and I shut down the watch mode dune, then the fetch loops will raise an exception about the poller. That then gets turned into a `Poller_disconnected` event which we process here.

otherlibs/dune-rpc/private/exported_types.ml
```diff
@@ -43,6 +44,26 @@ module Target = struct
     | Preprocess of string list
     | Loc of Loc.t
 
+  let compare x y =
```
rgrinberg, line 47: You can just use a map keyed by the diagnostic id's

bin/monitor.ml
rgrinberg: Okay, but then how will the ping succeed?

Alizter: Ah I see what you mean. How else can I be sure that the server has disconnected and I am not getting some other error from the poll.

rgrinberg: If there's another error, you can't really recover from it anyway. So there's no difference and you should terminate the connection regardless.

You could add a `--auto-poll` mode that tries to re-establish the connection if it's dead.

Alizter: I'll leave that for a future PR for now.

otherlibs/dune-rpc/private/exported_types.ml
Alizter: Ofc!

bin/monitor.ml
Alizter: I don't have access to the state in the fetch loops so I will have to maybe do something else. There haven't been any issues with this from my testing however.

rgrinberg: Good to hear that it's not an issue, but it's still confusing to write events after quit.

You can change:

```
; mutable quit : bool
```

to:

```
; quit : bool ref
```

and then just share the quit reference with the fetch loops.

bin/monitor.ml
```diff
@@ -0,0 +1,236 @@
...
+end
+
+(** A generic loop that fetches events from a [poller] and writes them to the
+    [event] mvar. *)
+let fetch_loop (event : Event.t Fiber.Mvar.t) (poller : 'a Client.Stream.t) ~f =
+  let rec loop () =
+    let* next = Client.Stream.next poller in
+    match next with
+    | Some payload -> Fiber.Mvar.write event (f payload) >>> loop ()
+    | None -> Fiber.Mvar.write event Event.Poller_disconnected
```
rgrinberg, line 126: This should also set the quit flag. We should signal disconnects as soon as we can

bin/monitor.ml
```diff
@@ -0,0 +1,236 @@
...
+        | Remove d ->
+          state.diagnostics <- Diagnostic.Id.Map.remove state.diagnostics d.id;
+          Console.reset ();
+          State.render_state state);
+      loop ()
+    (* If one of the writer loops disconnected, then we must have disconnected from the
+       server. Therefore we tell the user that we will be exiting and loop once more so
+       that we can clear the Mvar. *)
+    | Poller_disconnected ->
+      state.quit <- true;
```
rgrinberg, line 171: Whoever sends this event should make sure this flag is set already.

otherlibs/dune-rpc/private/exported_types.mli
```diff
@@ -56,6 +56,8 @@ module Diagnostic : sig
     val hash : t -> int
     val create : int -> t
     val sexp : (t, Conv.values) Conv.t
+
+    module Map : Stdune.Map.S with type key := t
```
rgrinberg, line 60: Remove this change.

otherlibs/dune-rpc/private/exported_types.mli
```diff
@@ -138,6 +140,8 @@ module Job : sig
...
```
rgrinberg, line 144: Remove this change as well

otherlibs/dune-rpc/private/exported_types.ml
```diff
@@ -179,10 +179,17 @@ module Diagnostic = struct
   ;;
 
   module Id = struct
-    type t = int
+    module T = struct
```
rgrinberg, line 182: Should be reverted

otherlibs/dune-rpc/private/exported_types.ml
```diff
@@ -254,6 +252,14 @@ module Diagnostic = struct
...
 
   let to_dyn t = Sexp.to_dyn (Conv.to_sexp sexp t)
+  let loc t = t.loc
+  let message t = t.message
+  let severity t = t.severity
+  let promotion t = t.promotion
+  let targets t = t.targets
+  let directory t = t.directory
+  let related t = t.related
+  let id t = t.id
```
rgrinberg, line 262: Should be reverted along with all other changes to dune-rpc/private

bin/monitor.ml
```diff
@@ -0,0 +1,236 @@
...
+      let signal_watcher = `Yes in
+      let watch_exclusions = Common.watch_exclusions common in
+      Dune_config.for_scheduler
+        config
+        stats
+        ~insignificant_changes
+        ~signal_watcher
+        ~watch_exclusions
+    in
+    Scheduler.Run.go config ~on_event:(fun _ _ -> ()) exec
```
rgrinberg, line 233: This should set the file watcher to `No_watcher`.

otherlibs/dune-rpc/private/exported_types.ml
Alizter: Why? This is how we get the maps.

rgrinberg: You can move the map functor application to the caller. We want to reduce the dependence on stdune in that library.

bin/monitor.ml
```diff
@@ -0,0 +1,236 @@
...
+    Cmd.info "monitor" ~doc ~man
+  and term =
+    let open Import in
+    let+ (common : Common.t) = Common.term in
+    let common = Common.forbid_builds common in
+    let config = Common.init ~log_file:No_log_file common in
+    let stats = Common.stats common in
+    let config =
+      let insignificant_changes = Common.insignificant_changes common in
+      let signal_watcher = `Yes in
```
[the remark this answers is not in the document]
Alizter: Well I think I need it because I have no way of sending `Ctrl + C` when in the monitor. When I have it, that can be used to quit the monitor. Without it I get stuck.

bin/monitor.ml
Alizter: I believe that is the default argument, however I will be explicit about it anyway.

bin/monitor.ml
```diff
@@ -0,0 +1,236 @@
...
+      let watch_exclusions = Common.watch_exclusions common in
```
rgrinberg, line 225: This should be the empty list.

bin/monitor.ml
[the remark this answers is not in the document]
rgrinberg: Something is seriously broken then. Initializing the signal watcher is what captures the `Ctrl + C`. By default, it should just be terminating the program.

bin/monitor.ml
```diff
@@ -42,7 +42,17 @@ module Event = struct
...
+module State : sig
+  (** Internal state of the render loop together with updates that update the UI via side
+      effects. *)
+
+  type t
+
+  val init : unit -> t
+  val update_jobs : t -> Job.Event.t list -> unit
+  val update_progress : t -> Progress.t -> unit
+  val update_diagnostics : t -> Diagnostic.Event.t list -> unit
```
rgrinberg, line 54: This function is still weird. Update and render should be separate steps and rendering should be done once after all the updates.

```
(match event with
| .. -> State.update state ..);
State.render state
```

rgrinberg: If this pattern isn't efficient enough, you can make the update functions return some `update` value to help do the update more efficiently and make render:

```
type update
val render : t -> update option -> unit
```

Alizter: I've made an incremental update type for the state that gets returned by these update functions.

bin/monitor.ml
```diff
@@ -0,0 +1,281 @@
...
+module Event = struct
+  (** Events that the render loop will process. *)
+  type t =
+    | Diagnostics of Diagnostic.Event.t list
+    | Jobs of Job.Event.t list
+    | Progress of Progress.t
+
+  let diagnostics l = Diagnostics l
+  let jobs l = Jobs l
+  let progress p = Progress p
```
rgrinberg, line 43: You might as well inline these constructors. You're only using them once and they aren't offering any abstraction.

src/fiber_event_bus/fiber_event_bus.ml
```diff
@@ -0,0 +1,70 @@
...
+  let status_to_dyn = function
+    | `Open -> Dyn.variant "`Open" []
+    | `Closed -> Dyn.variant "`Closed" []
+  in
+  Dyn.record
+    [ "status", status_to_dyn status
+    ; "readers", Dyn.opaque readers
+    ; "writers", Dyn.opaque writers
+    ]
+;;
```
Alizter, line 20: @rgrinberg We ended up not using this in the tests. Shall I remove it together with the dep on dyn? I'm not sure how useful this really is anymore. We can always add a `Fiber_event_bus.status` function for inspecting the status if needed.

PR #8153, opened 2023-07-09
refactor: introduce specialized location constructors

rgrinberg:
Introduce special constructors for:

* None locations
* A location in a particular file without the line/column set
* A location where start/stop are in the same file (basically all
  locations)
* A location where start/stop are on the same line (a large chunk of
  locations)

For [Same_file], we end up saving 2 words per location.

For [Same_line], we end up saving 5 words per location.

### DIFF
--- a/otherlibs/stdune/src/lexbuf.mli
+++ b/otherlibs/stdune/src/lexbuf.mli
@@ -28,6 +28,12 @@ module Loc : sig
 
   val equal : t -> t -> bool
 
+  val map_pos : t -> f:(Position.t -> Position.t) -> t
+
+  val in_file : fname:string -> t
+
+  val is_file_only : t -> bool
+
   (** To be used with [__POS__] *)
   val of_pos : string * int * int * int -> t
 
--- a/otherlibs/stdune/src/loc0.ml
+++ b/otherlibs/stdune/src/loc0.ml
@@ -1,19 +1,133 @@
-include Lexbuf.Loc
+type t =
+  | No_loc
+  | In_file of string
+  | Lexbuf_loc of Lexbuf.Loc.t
+  | Same_file of
+      { pos_fname : string
+      ; start_lnum : int
+      ; start_bol : int
+      ; start_cnum : int
+      ; stop_lnum : int
+      ; stop_bol : int
+      ; stop_cnum : int
+      }
+  | Same_line of
+      { pos_fname : string
+      ; pos_lnum : int
+      ; pos_bol : int
+      ; start_cnum : int
+      ; stop_cnum : int
+      }
 
-let to_lexbuf_loc x = x
+module Position = Lexbuf.Position
+open Lexbuf.Loc
 
-let of_lexbuf_loc x = x
+let to_lexbuf_loc = function
+  | No_loc -> Lexbuf.Loc.none
+  | Lexbuf_loc loc -> loc
+  | In_file fname -> Lexbuf.Loc.in_file ~fname
+  | Same_file
+      { pos_fname
+      ; start_lnum
+      ; start_bol
+      ; start_cnum
+      ; stop_lnum
+      ; stop_bol
+      ; stop_cnum
+      } ->
+    let start =
+      { Lexing.pos_fname
+      ; pos_lnum = start_lnum
+      ; pos_bol = start_bol
+      ; pos_cnum = start_cnum
+      }
+    in
+    let stop =
+      { Lexing.pos_fname
+      ; pos_lnum = stop_lnum
+      ; pos_bol = stop_bol
+      ; pos_cnum = stop_cnum
+      }
+    in
+    { Lexbuf.Loc.start; stop }
+  | Same_line { pos_fname; pos_lnum; pos_bol; start_cnum; stop_cnum } ->
+    let start =
+      { Lexing.pos_fname; pos_lnum; pos_bol; pos_cnum = start_cnum }
+    in
+    let stop = { Lexing.pos_fname; pos_lnum; pos_bol; pos_cnum = stop_cnum } in
+    { Lexbuf.Loc.start; stop }
 
-let start t = t.start
+let in_file ~fname = In_file fname
 
-let stop t = t.stop
+let of_lexbuf_loc loc =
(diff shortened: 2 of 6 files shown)


rgrinberg:
All in all, this is enough to shrink the heap by about 5% for a null build and save us a single major collection.

The memory impact of locations could be significantly reduced further with slightly more invasive changes.

otherlibs/stdune/src/loc0.ml
```diff
@@ -1,19 +1,133 @@
-include Lexbuf.Loc
+type t =
+  | None_loc
```
snowleopard, line 2: How about `No_location`? Seems easier to read.

otherlibs/stdune/src/loc0.ml
```diff
@@ -1,19 +1,133 @@
...
+  | In_file of string
+  | Loc of Lexbuf.Loc.t
```
snowleopard, line 4: And this one could be `Lexbuf_loc`/`Lexbuf_location`?

otherlibs/stdune/src/loc0.ml
rgrinberg: Made it `No_loc` to match the module's name

otherlibs/stdune/src/loc0.ml
rgrinberg: Lexbuf_loc it is.

snowleopard:
> All in all, this is enough to shrink the heap by about 5% for a null build and save us a single major collection.

@rgrinberg Could you clarify where is this saving coming from? I'd expect the representation of locations matter only when errors are involved, but I guess I'm wrong. I suppose we're storing a `Loc.t` for every `Rule.t`, for example. Anything else?

test/blackbox-tests/test-cases/patch-back-source-tree.t
```diff
@@ -201,7 +201,7 @@ produced in the sandbox and copied back:
 This is the internal stamp file:
 
   $ ls _build/.actions/default/blah*
-  _build/.actions/default/blah-3209c92f18c7050c580114796b6023bd
+  _build/.actions/default/blah-b4c6197b1e46d97994280f89ef27b024
```
snowleopard, line 204: Looks like action digests changed: do we need to bump the action digest version?

rgrinberg:
> Anything else?

There's a few:

* Every single ast element in a dune file contains a location. That's every atom, every element of a list, every single quoted string.

* Rules actually contain two locations :) One in the info, and another as a field. We should get rid of this redundancy.

* Alias rules also contain a location

snowleopard:
I see, thanks! I'll benchmark this change too.

otherlibs/stdune/src/loc0.ml
```diff
@@ -1,19 +1,133 @@
...
+      { Lexing.pos_fname; pos_lnum; pos_bol; pos_cnum = start_cnum }
+    in
+    let stop = { Lexing.pos_fname; pos_lnum; pos_bol; pos_cnum = stop_cnum } in
+    { Lexbuf.Loc.start; stop }
 
-let start t = t.start
+let in_file ~fname = In_file fname
 
-let stop t = t.stop
+let of_lexbuf_loc loc =
```
snowleopard, line 62: Why do we need the `Lexbuf_loc` variant if we have conversions both ways?

snowleopard: More generally, it seems odd that `Loc.t` depends on some `Lexbuf.Loc.t` instead of the other way around. If seems weird to special case `Lexbuf` in this way.

rgrinberg: > Why do we need the Lexbuf_loc variant if we have conversions both ways?

It's still needed for locations that span more than one file. It's essentially a fallback when we can't use an optimized representation. I do have an upcoming patch where the unoptimized version will be a bit more important.

> More generally, it seems odd that Loc.t depends on some Lexbuf.Loc.t instead of the other way around. If seems weird to special case Lexbuf in this way.

`Lexbuf.Loc.t` is a bit more simple I think. It's a stable representation we use for RPC and it comes straight from the lexer. So it makes sense that our specialized location type depends on it.

snowleopard: I see. That makes sense though I think it would still be cleaner if we had a most general location type (based on two positions) isomorphic to what's currently called `Lexbuf.Loc.t`, with both `Lexbuf.Loc.t` and `Loc.t` using it.

This doesn't need to happen in this PR though.

rgrinberg: What would this general location type be called though? Right now we have

* Lexbuf.Loc.t for what the lexer produces
* Loc.t for what dune prefers to use

Presumably this general location type would have a top level name, so we'd need to pick a name that somehow distinguishes it.

snowleopard: For example, the most general type could be `Loc.General.t`, and it could be used by both `Loc.t` and `Lexbuf.t`.

otherlibs/stdune/src/lexbuf.mli
```diff
@@ -28,6 +28,12 @@ module Loc : sig
 
   val equal : t -> t -> bool
 
+  val map_pos : t -> f:(Position.t -> Position.t) -> t
+
+  val in_file : fname:string -> t
```
Alizter, line 33: Can you add some comments?

Alizter:
I'm a bit confused by the PR description. Wasn't the none location done in a separate PR?

rgrinberg:
@snowleopard did you bench this one?

snowleopard:
> @snowleopard did you bench this one?

Not yet. I'll get there!

snowleopard:
@rgrinberg Sorry, I actually realised that importing this change is tricky since it depends on some earlier changes we haven't imported yet.

We can rely on your benchmarks.

P.S.: I'm still non-convinced about relying on `Lexbuf.Loc.t` as the most general location type, as discussed in one of the comments.

rgrinberg:
Will reshuffle `Lexbuf.Loc` and `Lexbuf.Position` in a separate PR (this PR didn't introduce these names)

PR #8159, opened 2023-07-10
fix(staged_pps): expand path relative to root

### LINKED ISSUE #8158: regression: staged_pps in main
Since #8109, `(staged_pps)` is broken in some cases. This can be reproduced by trying to build `frama-c` for example.

```
File "src/libraries/utils/unicode.ml", line 1:
Error: I/O error: ../.ppx/40f2e64236f938e28b7a041b2e0f4895/ppx.exe --as-ppx --cookie 'library-name="frama_c_kernel"' '/tmp/build_2b39c9_dune/camlppx57b0a7' '/tmp/build_2b39c9_dune/camlppx5a05f1'
File "src/.frama_c_kernel.objs/_unknown_", line 1, characters 0-0:
sh: 1: ../.ppx/40f2e64236f938e28b7a041b2e0f4895/ppx.exe: not found
File "src/libraries/datatype/unmarshal_z.ml", line 1:
Error: I/O error: ../.ppx/40f2e64236f938e28b7a041b2e0f4895/ppx.exe --as-ppx --cookie 'library-name="frama_c_kernel"' '/tmp/build_2b39c9_dune/camlppx71944c' '/tmp/build_2b39c9_dune/camlppxeb3e01'
File "src/.frama_c_kernel.objs/_unknown_", line 1, characters 0-0:
sh: 1: ../.ppx/40f2e64236f938e28b7a041b2e0f4895/ppx.exe: not found
```

emillon:
Fixes #8158

This is an error introduced in #8109, not part of a release.

### DIFF
--- a/src/dune_rules/preprocessing.ml
+++ b/src/dune_rules/preprocessing.ml
@@ -635,7 +635,8 @@ let pp_one_module sctx ~lib_name ~scope ~preprocessor_deps
                Expander.expand_and_eval_set expander driver.info.as_ppx_flags
                  ~standard:(Action_builder.return [ "--as-ppx" ])
              and* () = preprocessor_deps in
-             Command.expand_no_targets ~dir:(Path.build dir)
+             Command.expand_no_targets
+               ~dir:(Path.build (Super_context.context sctx).build_dir)
                (S [ Dep (Path.build exe); As driver_flags; As flags ]))
         in
         [ "-ppx"; String.quote_list_for_shell args ]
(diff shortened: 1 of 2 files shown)


Alizter:
Doesn't have to be in this PR, but do we not have any tests touching this? It would be good to have some for the future at least.

emillon:
We have tests but only with libs at the root. This one is on me; I checked that `dir` and `ctx.dir` were equal in the test suite so I tried to simplify it further but that assertion doesn't hold in general.

rgrinberg (approved):
A test would be useful indeed

rgrinberg (approved):
A test would be useful indeed

Alizter (approved):

PR #8160, opened 2023-07-10
refactor(memo): incremental cycles

rgrinberg:
Improve performance by removing interruptible folds

 I doubt this impact anything performance wise, but this code is both shorter and clearer IMO. It's quite weird that the fold returns `Continue` at the end even though there's nothing to continue.

### DIFF
--- a/vendor/incremental-cycles/src/incremental_cycles.ml
+++ b/vendor/incremental-cycles/src/incremental_cycles.ml
@@ -28,30 +28,6 @@ struct
 open Raw_graph
 
 (******************************************************************************)
-(* Interruptible fold_left.
-
-   At each step, the client function decides whether it wants to continue (by
-   using [Continue new_accumulator]) or stop (by using [Break return_value]).
-
-   Ultimately, [interruptible_fold] returns either the last accumulator or the
-   value returned by [Break], along with a boolean indicating whether it was
-   interrupted prematurely.
-*)
-
-type ('a, 'b) interruptible_fold_step =
-  | Continue of 'a
-  | Break of 'b
-
-let rec interruptible_fold f l acc =
-  match l with
-  | [] -> Continue acc
-  | x :: xs ->
-    let res = f x acc in
-    match res with
-    | Continue acc -> interruptible_fold f xs acc
-    | Break _ -> res
-
-(******************************************************************************)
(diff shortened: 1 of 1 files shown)


rgrinberg:
The benchmarks confirm that the minor collections have been reduced from 917 to 902. Minor collections are super fast, so this isn't really noteworthy, but it does confirm that the PR does something.

vendor/incremental-cycles/src/incremental_cycles.ml
```diff
@@ -78,29 +54,24 @@ let rec visit_backward
...
+    visit_backward_aux g target mark vertex stack fuel (get_incoming g vertex)
+and visit_backward_aux g target mark vertex stack fuel = function
+  | [] -> visit_backward g target mark fuel stack
+  | y :: ys ->
+    if fuel = 0 then
+      (* There is no fuel left *)
+      VisitBackwardInterrupted
+    else if is_marked g y mark then
+      (* This vertex has already been visited, skip it *)
+      visit_backward_aux g target mark vertex stack fuel ys
```
ejgallego, line 66: Why is not `fuel` decreased here anymore?

rgrinberg: indeed it should be.

rgrinberg:
Wasn't aware this library was extracted from Coq. I guess there's no point in modifying the code manually then.

Alizter:
Do we have the original Coq code? If we can't reproduce the extraction then it seems kind of hard to maintain?

snowleopard:
> Do we have the original Coq code? If we can't reproduce the extraction then it seems kind of hard to maintain?

Yep, the true sources are in https://gitlab.inria.fr/agueneau/incremental-cycles. Maybe we should vendor them too? With Dune's support for Coq, perhaps we can even regenerate things easily :)

Alizter:
@snowleopard Thanks for the pointer! I think bringing Coq into the main Dune build would cause more complexity than good, I can't think of a non-intrusive way to do it. If @rgrinberg thinks that the perf increase here is something we want, then perhaps we can try to reproduce it in the Coq code. Not something I want to do anytime soon however.

snowleopard:
@ejgallego I'd love to have some support for (on-demand) re-extraction of code from Coq!

PR #8161, opened 2023-07-10
feature(pkg): read toolchain from built packages

rgrinberg:
This doesn't contribute anything useful because of the rule loading issues I've pointed out. But might as well get it in to avoid conflicts.

### DIFF
--- a/src/dune_rules/which.ml
+++ b/src/dune_rules/which.ml
@@ -4,19 +4,42 @@ open Memo.O
 let programs_for_which_we_prefer_opt_ext =
   [ "ocamlc"; "ocamldep"; "ocamlmklib"; "ocamlobjinfo"; "ocamlopt" ]
 
-let best_path ~dir program =
-  let exe_path program =
-    let fn = Path.relative dir (program ^ Bin.exe) in
-    let+ exists = Fs_memo.file_exists (Path.as_outside_build_dir_exn fn) in
-    if exists then Some fn else None
-  in
-  if List.mem programs_for_which_we_prefer_opt_ext program ~equal:String.equal
-  then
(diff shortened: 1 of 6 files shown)


src/dune_rules/which.ml
```diff
@@ -4,19 +4,42 @@ open Memo.O
...
+        val file_exists : Path.t -> bool Monad.t
+      end) =
+  struct
+    open Monad.O
+
+    let best_path ~dir program =
+      let exe_path program =
+        let fn = Path.relative dir (program ^ Bin.exe) in
+        let+ exists = Fs.file_exists fn in
+        if exists then Some fn else None
```
Leonidas-from-XIV, line 19: This could be `Option.some_if`, but I am on the fence myself whether it makes it better.

src/dune_rules/which.mli
```diff
@@ -2,7 +2,17 @@ open Import
 
 (** [best_path ~dir prog] if [prog] is one of the special programs that can be
```
Leonidas-from-XIV, line 3: The comment needs to be updated.

Leonidas-from-XIV (commented):
I am not sure memo is the right thing to use here, since if we happen to build an ocaml toolchain the `Best_path` to the previously found toolchain changes.

But maybe I am overthinking this. We can very much revisit this when we're at that point.

rgrinberg:
> I am not sure memo is the right thing to use here, since if we happen to build an ocaml toolchain the Best_path to the previously found toolchain changes.

This should work transparently. If the toolchain changes, it will need to be rebuilt, which will cause us to re-run the corresponding action builder. The key point here is that we no longer need the file watching machinery to do this, since its all regular targets dune manages on its own.

PR #8164, opened 2023-07-10
refactor: use [Path.Table] for build system tables

rgrinberg:
It's a little faster than creating a table from [Table.create] since it
allows for inlining.


rgrinberg:
Actually, I just made this a part of #8152.

PR #8165, opened 2023-07-10
test: add a reproduction case for #8158

emillon:


Alizter (approved):
I guess you will update this when the fix is merged.

Alizter (approved):
I guess you will update this when the fix is merged.

rgrinberg:
We have some existing tests for similar functionality written by @gridbugs that don't rely on ppxlib. I'd suggest to just follow along what he has.

emillon:
@rgrinberg I just did that, let me know what you think.

rgrinberg (approved):

PR #8166, opened 2023-07-10
refactor: halve the work of [Action_builder.contents]

rgrinberg:
I noticed that [Action_builder.contents] shares no work between [Eager]
and [Lazy]. This is wrong because its implementation just ignores this
argument.

The result is that we should do half the work reading files and store
half the table as well.

### DIFF
--- a/src/dune_engine/action_builder.ml
+++ b/src/dune_engine/action_builder.ml
@@ -79,22 +79,22 @@ let env_var s = deps (Dep.Set.singleton (Dep.env s))
 
 let alias a = dep (Dep.alias a)
 
-let contents_impl p =
-  of_thunk
-    { f =
-        (fun _mode ->
-          let open Memo.O in
-          let+ x = Build_system.read_file p ~f:Io.read_file in
-          (x, Dep.Map.empty))
-    }
-
 let contents =
-  let memo =
-    create_memo "file-contents"
-      ~input:(module Path)
-      ~cutoff:String.equal contents_impl
+  let read_file =
+    Memo.exec
+      (Memo.create "Action_builder.contents"
+         ~input:(module Path)
+         ~cutoff:String.equal
+         (fun p -> Build_system.read_file p ~f:Io.read_file))
   in
-  fun path -> exec_memo memo path
+  fun p ->
+    of_thunk
+      { f =
+          (fun _mode ->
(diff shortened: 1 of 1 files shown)


snowleopard (commented):
Whoops, that seems like a great catch! Let me benchmark it first.

rgrinberg:
Probably you won't observe much as we only use `Eager` for building things. `Lazy` is only used for somewhat specialized things.

src/dune_engine/action_builder.ml
```diff
@@ -79,22 +79,22 @@ let env_var s = deps (Dep.Set.singleton (Dep.env s))
...
-  let memo =
-    create_memo "file-contents"
-      ~input:(module Path)
-      ~cutoff:String.equal contents_impl
+  let read_file =
+    Memo.exec
+      (Memo.create "Action_builder.contents"
+         ~input:(module Path)
+         ~cutoff:String.equal
+         (fun p -> Build_system.read_file p ~f:Io.read_file))
```
snowleopard, line 88: Btw, does it make sense to run `Io.read_file` in a background thread?

snowleopard:
> Probably you won't observe much as we only use `Eager` for building things. `Lazy` is only used for somewhat specialized things.

Right, indeed, I seem to see only noise in the benchmarks. So, do we really need `Lazy`? Maybe it used to be more useful but now we're building ~everything eagerly?

rgrinberg:
We have a few uses of it in the rules. Though we could probably get rid of them fairly easily.

It seems to be important for the reflection API in `Dune_engine.Reflection`. I have no use it for myself though.

snowleopard:
> We have a few uses of it in the rules. Though we could probably get rid of them fairly easily.
> 
> It seems to be important for the reflection API in `Dune_engine.Reflection`. I have no use it for myself though.

Ah, yes. We do use reflection. I'll see if we can get rid of the `Lazy` mode there.

snowleopard (approved):

src/dune_engine/action_builder.ml
rgrinberg: Yeah, it wouldn't be such a bad idea to experiment with it. Although my experience with actions tells me we need to reduce the overhead here first.

Issue #7784, opened 2023-05-22
add dune show command


Alizter:
> I don't like the name "describe" and would prefer something like `dune show` instead, but that's a detail. I think we can end up polluting the top-level namespace if we keep adding commands like `dune targets`.

`dune show` seems like a good top-level name for this kind of functionality. Why not add `dune show` and put the command under it? (`dune show targets`). Existing commands could be migrated under it as well later: `dune show rules` instead of `dune rules`, `dune show env` instead of `dune printenv`, `dune show workspace` instead of `dune describe workspace`, etc

_Originally posted by @nojb in https://github.com/ocaml/dune/issues/7770#issuecomment-1557868979_

nojb:
Also: `dune installed-libraries` should probably become `dune show installed-libraries`.

For `dune describe` there are a number of subcommands: `dune describe opam-files`, `dune describe pp` and `dune describe external-lib-deps`. They should all be migrated under the `dune show` command.

Alizter:
@snowleopard What exactly do you want `dune show sources` to do?

snowleopard:
> @snowleopard What exactly do you want `dune show sources` to do?

Right now, if a directory contains an ignored file, a source file, and a target (generated) file, `dune show targets` would print the source and the target files.

I think there are possibly two issues with it. 

Firstly, seeing the source file in the output may be confusing: the user asked for "targets", after all.

Secondly, there is currently no way to see just sources (still skipping ignored files and targets) or just targets (skipping ignored files and sources).

I don't necessarily advocate for adding a new command, but the current design doesn't seem great.

snowleopard:
Here is an alternative design that may be an improvement:

* Change the command to `dune show files` to avoid confusion. Also works well for a possible `dune show dirs` in future.
* Support flags `dune show files --sources-only` and `--targets-only` to make it possible to see the corresponding subsets.

Alizter:
I don't like the `dune show files` idea. We are declaring every buildable target to the user, this includes the source files. I would rather have an option to exclude source files such as `dune show targets --no-sources`.

snowleopard:
> We are declaring every buildable target to the user, this includes the source files.

Are you sure the majority of Dune users considers sources to be targets? I doubt it. Perhaps we can ask them!

That Dune copies sources into `_build`, and therefore happens to have build rules for them, seems somewhat accidental to me/leaking implementation details.

> Also `dune show sources` seems kind of moot as this is essentially just ls at this point.

[...]

PR #8167, opened 2023-07-10
feature: --all option for dune show targets

Alizter:
We change `dune show targets` to omit source files by default and add an `--all` option which will show all targets including those that come from the source tree.

Addresses wish in #7784.

- [x] changelog
- [x] tests

### DIFF
--- a/bin/describe/aliases_targets.ml
+++ b/bin/describe/aliases_targets.ml
@@ -1,12 +1,16 @@
 open Import
 
-let ls_term (fetch_results : Path.Build.t -> string list Action_builder.t) =
+let common_args =
   let+ common = Common.term
   and+ paths = Arg.(value & pos_all string [ "." ] & info [] ~docv:"DIR")
   and+ context =
     Common.context_arg
       ~doc:"The context to look in. Defaults to the default context."
   in
+  (common, paths, context)
+
+let ls_term ~common ~paths ~context
+    (fetch_results : Path.Build.t -> string list Action_builder.t) =
   let config = Common.init common in
   let request (_ : Dune_rules.Main.build_system) =
     let header = List.length paths > 1 in
@@ -104,7 +108,9 @@ module Aliases_cmd = struct
     in
     List.map ~f:Dune_engine.Alias.Name.to_string alias_targets
 
-  let term = ls_term fetch_results
+  let term =
+    let+ common, paths, context = common_args in
+    ls_term ~common ~paths ~context fetch_results
 
   let command =
     let doc = "Print aliases in a given directory. Works similalry to ls." in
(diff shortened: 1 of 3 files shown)


snowleopard:
While this seems like an improvement, I find it very unintuitive that `dune show targets --all` shows targets and sources.

I guess I still prefer `dune show files` command, with some filters on top.

bin/describe/aliases_targets.ml
```diff
@@ -104,33 +108,56 @@ module Aliases_cmd = struct
...
-  let term = ls_term fetch_results
+  let term =
+    let+ common, paths, context = common_args
+    and+ all =
+      Arg.(
+        value & flag
+        & info [ "all" ]
+            ~doc:
+              "Print all targets, including those that exist in the source \
+               tree.")
```
snowleopard, line 158: "All targets including those that exist in the source tree" doesn't make much sense in either of the two worlds:
* If one considers sources to be targets, then the "including ..." part seems tautological/unnecessary.
* If one doesn't consider sources to be targets, then the "print all targets" makes no sense.

snowleopard:
> I guess I still prefer `dune show files` command, with some filters on top.

There just seem to be no way to make things non-confusing when you use "targets" or "sources" as the starting point while taking about a generic concept (such as files). Neither targets, not sources, are more general than the the other. Files are!

Alizter:
@snowleopard I think I might have a different definition of "target" to the one you are thinking about. For me a target is anything that "dune build" accepts minus aliases (that's another story).

I would be happy to remove the `--all` argument completely if we agree that inspecting the source is uninteresting. However your suggestion of `dune show files` doesn't make sense to me since it is just `ls` at this point?

snowleopard:
> For me a target is anything that "dune build" accepts minus aliases (that's another story).

This definition is pretty inconsistent with, say, what we call targets in actions. Rules have `deps` (sometimes sources, sometimes other targets) and `targets` (some of which might actually be promoted and committed as sources).

You are designing a UX feature and it's worth thinking about the least confusing way of describing all Dune concepts.

> However your suggestion of `dune show files` doesn't make sense to me since it is just `ls` at this point?

It's not `ls`, as I keep saying. There are files that Dune ignores, and they will not appear in the output. Also, there are directory targets, which shouldn't appear in `dune show files` either (but can appear in `dune show dirs`).

UX is hard! I think neither the current state of things, nor this PR in the current form, make for a good, non-confusing UX story.

snowleopard:
> UX is hard! I think neither the current state of things, nor this PR in the current form, make for a good, non-confusing UX story.

But also, just to reiterate, I'm far from being a UX expert. Sometimes it's worth talking to users to find out how they think about the sources/targets distinction. We, developers, have a very skewed view on things.

Alizter:
OK @snowleopard thanks for your valuable feedback. I think I will draft this for now and think about it more when I have some time.

snowleopard:
> OK @snowleopard thanks for your valuable feedback. I think I will draft this for now and think about it more when I have some time.

Happy to brainstorm it at some point to try to figure out a way forward!

PR #8168, opened 2023-07-10
remove linkopts(javascript) from META

hhugo:
Using `jsoo_runtime` (instead of `linkopts(javascript`) to retrieve jsoo runtime files has been advertised in the documentation since jsoo 4.1.0 (see https://ocsigen.org/js_of_ocaml/4.1.0/manual/runtime-files).

How to proceed with such cleanup wrt backward compatibility ?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+Unreleased
+----------
+
+- No longer emit linkopts(javascript) in META files (#8168, @hhugo)
+
 3.10.0 (2023-07-31)
 -------------------
 
--- a/src/dune_rules/gen_meta.ml
+++ b/src/dune_rules/gen_meta.ml
@@ -163,12 +163,7 @@ let gen_lib pub_name lib ~path ~version =
       | l ->
         let root = Pub_name.root pub_name in
         let l = List.map l ~f:Path.basename in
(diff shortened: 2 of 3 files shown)


rgrinberg:
I think it should be enough to add a conflict with older versions of the jsoo compiler. Otherwise, you also have the option to version the new behavior behind the version of the dune language in the project.

hhugo:
@rgrinberg, the change is not incompatible with older version of jsoo. It just removes an undocumented (in dune and jsoo) feature that was used before dune existed.

rgrinberg:
Okay, so we can remove it without much ceremony. The PR still needs fix the issues in CI, add a test, and a CHANGES entry.

hhugo:
I've removed dead code and added a changelog entry.
I don't know what kind of test you have in mind.

Alizter:
@hhugo I guess just a simple cram test which dumps the contents of the META file. You can use the `PATH_BUILD_PREFIX_MAP` variable to scrub any absolute paths.

hhugo:
Existing tests already capture the change of behavior

rgrinberg:
@hhugo could you rebase please?

hhugo:
now rebased

PR #8172, opened 2023-07-11
refactor: improve flushing unreliable digest

rgrinberg:
do the debug check only once

### DIFF
--- a/src/dune_engine/cached_digest.ml
+++ b/src/dune_engine/cached_digest.ml
@@ -97,19 +97,27 @@ let delete_very_recent_entries () =
      digests too aggressively. *)
   match Float.compare cache.max_timestamp now with
   | Lt -> ()
-  | Eq | Gt ->
-    Path.Table.filteri_inplace cache.table ~f:(fun ~key:path ~data ->
-        match Float.compare data.stats.mtime now with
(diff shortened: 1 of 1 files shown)


src/dune_engine/cached_digest.ml
```diff
@@ -97,19 +97,27 @@ let delete_very_recent_entries () =
...
-          if !Clflags.debug_digests then
+  | Eq | Gt -> (
+    let filter (data : file) =
+      match Float.compare data.stats.mtime now with
+      | Lt -> true
+      | Gt | Eq -> false
+    in
+    match !Clflags.debug_digests with
+    | false ->
+      Path.Table.filteri_inplace cache.table ~f:(fun ~key:_ ~data ->
```
snowleopard, line 108: How about using `Path.Table.filter_inplace` here? Or perhaps we don't have it available here?

snowleopard (approved):
Looks good!

src/dune_engine/cached_digest.ml
rgrinberg: Yup, it's not available. And it would be implemented on top of the `filteri_inplace` primitive so there should be no advantage to it

PR #8173, opened 2023-07-11
refactor: better path maps and sets

rgrinberg:

### DIFF
--- a/otherlibs/fiber/src/fiber.mli
+++ b/otherlibs/fiber/src/fiber.mli
@@ -101,6 +101,8 @@ val all_concurrently_unit : unit t list -> unit t
 (** Iter over a list in parallel. *)
 val parallel_iter : 'a list -> f:('a -> unit t) -> unit t
 
+val parallel_iter_seq : 'a Seq.t -> f:('a -> unit t) -> unit t
+
 val parallel_iter_set :
      (module Set.S with type elt = 'a and type t = 's)
   -> 's
--- a/otherlibs/stdune/src/path.ml
+++ b/otherlibs/stdune/src/path.ml
@@ -292,6 +292,14 @@ module Local : sig
 
   module Table : Hashtbl.S with type key = t
 
+  module Map : Map_intf.S with type key = t
+
+  module Set : sig
+    include Set_intf.S with type elt = t and type 'a map = 'a Map.t
+
+    val of_listing : dir:elt -> filenames:Filename.t list -> t
+  end
+
   val root : t
 
   val is_root : t -> bool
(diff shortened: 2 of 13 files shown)


snowleopard (commented):
That's... a lot of duplicated code. Do you at least get good results in benchmarks?

snowleopard:
> That's... a lot of duplicated code. 

Can we get rid of some code duplication using functors? I feel like it's going to be hard to maintain this in the long run: people will be making changes to sets/maps and forgetting to update the corresponding code for paths.

rgrinberg:
Yeah, it's definitely not optimal. Let me see if this change is at all useful first. Then I can work on removing some duplication.

snowleopard:
Oh, we're getting pretty nice figures from the benchmark bot!

Alizter:
@snowleopard The figures in the benchmark bot have been mostly garbage, I think its best to ignore it for now. Even doc PRs were getting wild numbers.

snowleopard:
> @snowleopard The figures in the benchmark bot have been mostly garbage, I think its best to ignore it for now. Even doc PRs were getting wild numbers. Hopefully it will be turned off soon.

Oh, that's sad.

> (It actually has started posting as a table).

Yay! :)

rgrinberg:
@snowleopard This PR does seem to give a significant drop to major collections for clean builds. Could you send it through a round of benchmarks internally? If the results are promising, I can clean it up and remove some of the duplication.

snowleopard:
@rgrinberg Could you import it internally? I'll kick off some benchmarks!

rgrinberg:
@snowleopard IIRC you benched this and didn't get a positive change, right?

snowleopard:
> @snowleopard IIRC you benched this and didn't get a positive change, right?

It's a positive change, particular for small builds (for some reason?). But larger builds, it's below 1% speed up, and no savings in memory. Given the size of the diff, not sure it's really worth it.

snowleopard:
Maybe we should try a different approach, and try to eradicate `Path.Map`s altogether, and instead keep multiple different maps where that's necessary?

rgrinberg:
> Maybe we should try a different approach, and try to eradicate Path.Maps altogether, and instead keep multiple different maps where that's necessary?

I fear that we'll end up having to keep just as much as code, but without the abstraction to help us. Also, the other problem is that I don't know where this change is necessary - if at all. Before writing this PR, I imagined it would be helpful everywhere and this representation is strictly superior almost everywhere. But judging by the benchmarks, the improvement seems rather underwhelming.

By the way, now that we have unboxed option, we could make empty sets/maps be 1 word. That would truly make this PR strictly superior. Although again i'm not sure if it will at all matter.

I was much more interested in seeing reduction in memory, and that didn't work out at all. I'll let you judge if the performance benefits are worth it. For the kind of workspaces I'm working with, it doesn't seem like it's worth it. Closing until there's further interest in such tiny speed-ups.

snowleopard:
I think we should come back to this at some point in future when we start to put more effort into optimisation.

snowleopard:
Thanks for the experiment! We did learn something new, so it wasn't a waste of time.

PR #8174, opened 2023-07-11
refactor: replace union with superpose

rgrinberg:
Remove an unnecessary closure allocation

### DIFF
--- a/otherlibs/stdune/src/map.ml
+++ b/otherlibs/stdune/src/map.ml
@@ -204,7 +204,9 @@ module Make (Key : Key) : S with type key = Key.t = struct
 
   let filter_opt t = filter_map t ~f:Fun.id
 
-  let superpose a b = union a b ~f:(fun _ x _ -> Some x)
+  let superpose =
+    let f _ x _ = Some x in
+    fun a b -> union a b ~f
 
   let is_subset t ~of_ ~f =
     let not_subset () = raise_notrace Exit in


rgrinberg:
There's a similar instance in `Facts.union`:

```
  let union a b =
    Map.union a b ~f:(fun _ a b ->
        assert (a = b);
        Some a)
```

But it cannot be removed because of the assertion.

otherlibs/stdune/src/map.ml
```diff
@@ -204,7 +204,9 @@ module Make (Key : Key) : S with type key = Key.t = struct
 
   let filter_opt t = filter_map t ~f:Fun.id
 
-  let superpose a b = union a b ~f:(fun _ _ y -> Some y)
+  let superpose =
+    let f _ _ y = Some y in
```
snowleopard, line 208: Why is `superpose` right biased? It seems surprising to me, and the call sites you're replacing were left-biased.

Can we make `superpose` left-biased?

snowleopard:
> There's a similar instance in `Facts.union`:
> 
> ```
>   let union a b =
>     Map.union a b ~f:(fun _ a b ->
>         assert (a = b);
>         Some a)
> ```
> 
> But it cannot be removed because of the assertion.

I suppose we could use `superpose` here if we passed a custom `f` as a parameter. But it doesn't matter: the quoted code is going to disappear soon.

otherlibs/stdune/src/map.ml
rgrinberg: > Why is superpose right-biased?

No idea, I suppose it was Jeremie's decision early on.

> and the call sites you're replacing were left-biased

I flipped the arguments in the applications though.

> Can we make superpose left-biased?

Why not, I don't mind doing it in a separate PR.  Both choices seem a little arbitrary to me and I would prefer some labels to make this clear.

rgrinberg:
> I suppose we could use superpose here if we passed a custom f as a parameter.

It would mean allocating on every element that is found in common though. Which is a little sad.

otherlibs/stdune/src/map.ml
snowleopard: > I flipped the arguments in the applications though.

Oh, I didn't notice that. I think my preference would be to first make `superimpose` left-biased and undo the swapping in this PR.

Left-biased operations are a lot more common, I think, which is why I find the current behaviour of `superimpose` surprising (and it's probably just accidental).

rgrinberg:
@snowleopard ready now

snowleopard (approved):
Looks good, though the code you're changing is going to disappear soon. I'm rewriting the whole `dep.ml` internally, so it'll just cause some conflicts in the end.

rgrinberg:
Okay, I'll just leave the optimization to `superpose` then.

snowleopard:
> Okay, I'll just leave the optimization to `superpose` then.

Sounds good, thanks!

PR #8176, opened 2023-07-11
chore: leave a comment about the weirdness of File_selector.dir

rgrinberg:

### DIFF
--- a/src/dune_engine/file_selector.ml
+++ b/src/dune_engine/file_selector.ml
@@ -1,7 +1,10 @@
 open Import
 
 type t =
-  { dir : Path.t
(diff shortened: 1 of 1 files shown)


src/dune_engine/file_selector.ml
```diff
@@ -1,7 +1,10 @@
 open Import
 
 type t =
-  { dir : Path.t
+  { (* CR rgrinberg: this [dir] is ignored when evaluating the selector with
```
snowleopard, line 4:
```suggestion
  { (* CR-someday rgrinberg: this [dir] is ignored when evaluating the selector with
```

snowleopard (approved):
Thanks!

PR #8177, opened 2023-07-11
refactor: remove useless addition of dirs

rgrinberg:
[Fact.Files.necessary_dirs_for_sandboxing] already includes the
directories in [ps.dirs], we don't need to add them again


snowleopard:
> [Fact.Files.necessary_dirs_for_sandboxing] already includes the directories in [ps.dirs], we don't need to add them again

Are you sure `Fact.Files.necessary_dirs_for_sandboxing` adds the same paths? I think it has an extra `Path.Build.parent_exn` call.

rgrinberg:
Indeed. I added a test for this in #8296

PR #8180, opened 2023-07-11
refactor: use [Filename] in odoc rules

rgrinberg:

### DIFF
--- a/src/dune_rules/odoc.ml
+++ b/src/dune_rules/odoc.ml
@@ -390,8 +390,7 @@ let setup_css_rule sctx =
         ~flags_for:None
         [ A "-o"; Path (Path.build dir) ]
     in
-    cmd
-    |> Action_builder.With_targets.add_directories ~directory_targets:[ dir ]
+    Action_builder.With_targets.add_directories ~directory_targets:[ dir ] cmd
   in
   add_rule sctx run_odoc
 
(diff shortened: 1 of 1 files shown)


src/dune_rules/odoc.ml
```diff
@@ -443,20 +442,19 @@ let libs_of_pkg ctx ~pkg =
...
       match entry with
+      | Deprecated_library_name _ -> None
       | Library lib ->
         let is_impl =
           Lib.Local.to_lib lib |> Lib.info |> Lib_info.implements
           |> Option.is_some
         in
-        Option.some_if (not is_impl) lib
-      | Deprecated_library_name _ -> None)
+        Option.some_if (not is_impl) lib)
```
Alizter, line 451: I noticed you could switch to Option.is_none instead and get rid of the not.

src/dune_rules/odoc.ml
```diff
@@ -443,20 +442,19 @@ let libs_of_pkg ctx ~pkg =
...
 
 let load_all_odoc_rules_pkg sctx ~pkg =
   let* pkg_libs = libs_of_pkg sctx ~pkg in
   let+ () =
-    Memo.parallel_iter
-      (Pkg pkg :: List.map pkg_libs ~f:(fun lib -> Lib lib))
-      ~f:(fun _ -> Memo.return ())
+    Pkg pkg :: List.map pkg_libs ~f:(fun lib -> Lib lib)
+    |> Memo.parallel_iter ~f:(fun _ -> Memo.return ())
```
Alizter, line 457: `Memo.all_concurrently`?

Alizter (approved):

src/dune_rules/odoc.ml
rgrinberg: Doesn't help. We would need to ignore the list of units in the end.

Alizter: ah I was confused. Fiber has an `all_concurrently_unit` that doesn't allocate a return list.

Alizter: In fact Memo already has it since it `include`s `Fiber`. All you need to do is add
```
val all_concurrently_unit : unit t list -> unit t
```
after `all_concurrently` after `memo.mli`.

PR #8181, opened 2023-07-11
refactor: remove extra word from stable digest

rgrinberg:
The following constructor

```
File of (string * Digest.t)
```

Is strictly worse than:

```
File of string * Digest.t
```

Because of the extra layer of boxing.

I don't know if this was intentional, but it doesn't seem useful to me.


Alizter (approved):

snowleopard (approved):
Nice!

PR #8182, opened 2023-07-11
refactor(odoc): remove dummy file

emillon:
It might have been the case that Dune wouldn't work without it but it now seems fine.

### DIFF
--- a/src/dune_rules/odoc.ml
+++ b/src/dune_rules/odoc.ml
@@ -68,16 +68,10 @@ type target =
   | Lib of Lib.Local.t
   | Pkg of Package.Name.t
 
-type source =
-  | Module
-  | Mld
-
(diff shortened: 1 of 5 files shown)


emillon:
I'm curious to know about the history here: the test suite seems fine without this, and there's an explicit hidden target in the rule, so I don't think there's an issue with removing that. Conversely, if there's an issue I think we need to add it to the test suite. Any opinions @rgrinberg @jonludlam ?

Alizter:
Isn't the correct fix to use directory targets?

emillon:
> Isn't the correct fix to use directory targets?

Yes but I'm interested to see if I can get a conservative improvement first.

rgrinberg (approved):

PR #8183, opened 2023-07-11
refactor(odoc): remove rmdir/mkdir before odoc

emillon:
The odoc binary is able to create the directory by itself, so it does not seem to be necessary anymore.


emillon:
Similar to #8182, I'm interested to see what we're missing by removing that. This can also be an oppportunity to fix the test suite to show that it's actually necessary.

rgrinberg:
Indeed there's no need to worry about removal. We should turn on sandboxing and not worry about this.

rgrinberg (approved):

PR #8185, opened 2023-07-11
refactor: use stdlib for filter_inplace

rgrinberg:
The primitive from the stdlib is a lot faster


snowleopard (approved):
Great! I suppose we're not going to notice this in benchmarks.

Alizter:
Wasn't this introduced in 4.03? Im confused to see this build.

PR #8187, opened 2023-07-11
feature(stdune): introduce compact positions

rgrinberg:
We make use of the fact that position offsets are often small integers
and store them all in a single OCaml int.

When this optimization works, we save 3-4 words per location. When it
doesn't, there's no harm and we fall back to our old "wide" offsets.

### DIFF
--- a/otherlibs/stdune/src/compact_position.ml
+++ b/otherlibs/stdune/src/compact_position.ml
@@ -1,0 +1,141 @@
+module Position = struct
+  (* We encode the position in three, 21 bit fields: [cnum][lnum][bol] *)
+  type t = int
+
+  let field_size = 21
+  let field_mask = (1 lsl field_size) - 1
+  let shift_bol = 0
+  let shift_lnum = field_size
+  let shift_cnum = 2 * field_size
+
+  let small_enough =
+    let max_size = 1 lsl field_size in
+    let test int = int <= max_size in
+    fun [@inline] { Lexing.pos_bol; pos_cnum; pos_lnum; pos_fname = _ } ->
+      test pos_bol && test pos_cnum && test pos_lnum
+  ;;
+
+  let[@inline] of_position { Lexing.pos_bol; pos_cnum; pos_lnum; pos_fname = _ } =
+    ((pos_bol land field_mask) lsl shift_bol)
+    lor ((pos_lnum land field_mask) lsl shift_lnum)
+    lor ((pos_cnum land field_mask) lsl shift_cnum)
+  ;;
+
+  let[@inline] bol t = (t lsr shift_bol) land field_mask
+  let[@inline] lnum t = (t lsr shift_lnum) land field_mask
+  let[@inline] cnum t = (t lsr shift_cnum) land field_mask
+
+  let to_position t ~fname:pos_fname =
+    let pos_bol = bol t in
+    let pos_cnum = cnum t in
+    let pos_lnum = lnum t in
+    { Lexing.pos_bol; pos_cnum; pos_lnum; pos_fname }
+  ;;
+end
+
+module Same_line_loc = struct
+  (* we encode the location in four, 15 bit chunks
+     [bol][lnum][start_cnum][stop_cnum]
+
+     Note that this leaves us with 3 spare bits. We should probably use them to
+     expand [bol] and [lnum] a little.
+
+     CR-someday jtov: Instead of [stop_cnum], we can store [stop_cnum -
+     start_cnum]. This should be smaller than [stop_cnum] and release more
+     bits for other fields.
+  *)
+  type t = int
+
+  let field_size = 15
+  let field_mask = (1 lsl field_size) - 1
+  let shift_bol = 0
+  let shift_lnum = field_size
+  let shift_start_cnum = 2 * field_size
+  let shift_stop_cnum = 3 * field_size
+
+  let create ~bol ~lnum ~start_cnum ~stop_cnum =
+    ((bol land field_mask) lsl shift_bol)
+    lor ((lnum land field_mask) lsl shift_lnum)
+    lor ((start_cnum land field_mask) lsl shift_start_cnum)
+    lor ((stop_cnum land field_mask) lsl shift_stop_cnum)
+  ;;
+
+  let[@inline] bol t = (t lsr shift_bol) land field_mask
+  let[@inline] lnum t = (t lsr shift_lnum) land field_mask
+  let[@inline] start_cnum t = (t lsr shift_start_cnum) land field_mask
+  let[@inline] stop_cnum t = (t lsr shift_stop_cnum) land field_mask
+
+  let set_start_to_stop t =
+    let bol = bol t in
+    let lnum = lnum t in
+    let stop_cnum = stop_cnum t in
+    (* this can be optimized more if necessary *)
+    create ~bol ~lnum ~start_cnum:stop_cnum ~stop_cnum
+  ;;
+
+  let small_enough =
+    let max_size = 1 lsl field_size in
+    fun [@inline] int -> int <= max_size
+  ;;
+
+  let[@inline] to_loc t ~fname:pos_fname =
+    let pos_lnum = lnum t in
+    let pos_bol = bol t in
+    let start = { Lexing.pos_fname; pos_lnum; pos_bol; pos_cnum = start_cnum t } in
+    let stop = { start with pos_cnum = stop_cnum t } in
+    { Lexbuf.Loc.start; stop }
+  ;;
+
+  let[@inline] start t ~fname:pos_fname =
+    let pos_lnum = lnum t in
+    let pos_bol = bol t in
+    { Lexing.pos_fname; pos_lnum; pos_bol; pos_cnum = start_cnum t }
+  ;;
+
+  let[@inline] stop t ~fname:pos_fname =
+    let pos_lnum = lnum t in
+    let pos_bol = bol t in
+    { Lexing.pos_fname; pos_lnum; pos_bol; pos_cnum = stop_cnum t }
+  ;;
+end
+
+include Position
+
+type of_loc =
+  | Same_line of Same_line_loc.t
+  | Loc of
+      { start : t
+      ; stop : t
+      }
+  | Loc_does_not_fit
+
+let[@inline] try_loc { Lexbuf.Loc.start; stop } =
+  if Position.small_enough start && Position.small_enough stop
+  then (
+    let start = Position.of_position start in
+    let stop = Position.of_position stop in
+    Loc { start; stop })
+  else Loc_does_not_fit
+;;
+
+let[@inline] of_loc ({ Lexbuf.Loc.start; stop } as loc) =
+  if start.pos_fname <> stop.pos_fname
+  then Loc_does_not_fit
+  else if start.pos_bol = stop.pos_bol && start.pos_lnum = stop.pos_lnum
+  then (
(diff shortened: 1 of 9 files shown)


otherlibs/stdune/src/compact_position.ml
```diff
@@ -0,0 +1,118 @@
+module Position = struct
+  type t = int
+
+  let size = 21
+
+  let mask_bol = 0x1FFFFF (* 21 bits set *)
```
Leonidas-from-XIV, line 6: Maybe compute this value to prevent desyncing with `size`?

test/blackbox-tests/test-cases/patch-back-source-tree.t
```diff
@@ -175,12 +175,15 @@ If a source file is read-only, the action sees it as writable:
 
   $ dune build
   writable
+  Error: open(x): Permission denied
+  -> required by alias default in dune:1
+  [1]
```
Leonidas-from-XIV, line 180: Is this an expected change? I would not expect anything in the expect tests to change just by using a more memory compact representation.

Leonidas-from-XIV (commented):
Interesting PR. I wasn't aware that the memory usage of the locations is a problem.

otherlibs/stdune/src/compact_position.ml
```diff
@@ -0,0 +1,118 @@
...
+
+include Position
+
+type of_loc =
+  | Same_line of Same_line_loc.t
+  | Loc of
+      { start : t
+      ; stop : t
+      }
+  | Location_does_not_fit
```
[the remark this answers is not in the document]
rgrinberg: > Of course that makes the memory usage worse in that sitaution as another cell is allocated but this would be hopefully a rare case.

That's precisely why I omitted it. This Compact_position is a low level, private API, so we can keep it a little ugly for performance's sake.

rgrinberg:
> Interesting PR. I wasn't aware that the memory usage of the locations is a problem.

Probably not the main problem, but this PR is certainly helpful. We shave off a major collection in a clean build and two major collections for a null build. Don't think we've had more impactful changes than this one memory wise.

test/blackbox-tests/test-cases/patch-back-source-tree.t
rgrinberg: Indeed, this change is wrong and not needed. Will remove.

rgrinberg:
@snowleopard bump on this one. This is good to go and the 2nd part to the location optimization PR. Might as well get them all in before pulling.

otherlibs/stdune/src/compact_position.mli
```diff
@@ -0,0 +1,47 @@
+(** Positions information that can be encoded within a single immediate *)
```
snowleopard, line 1: Why is this not just `position.mli`? The "compact" bit is just an implementation detail, right?

otherlibs/stdune/src/compact_position.mli
```diff
@@ -0,0 +1,47 @@
...
+
+type t [@@immediate]
+
+val of_position : Lexbuf.Position.t -> t
```
snowleopard, line 5: This could perhaps be `of_lexbuf_position` to avoid confusion if you agree to rename this module to just `Position`.

otherlibs/stdune/src/compact_position.ml
```diff
@@ -0,0 +1,134 @@
+module Position = struct
+  (* We encode the position in 3, 21 bit chunks: [cnum][lnum][bol] *)
```
snowleopard, line 2:
```suggestion
  (* We encode the position in three 21-bit chunks: [cnum][lnum][bol] *)
```

otherlibs/stdune/src/compact_position.ml
```diff
@@ -0,0 +1,134 @@
+module Position = struct
+  (* We encode the position in three 21-bit chunks: [cnum][lnum][bol] *)
+  type t = int
+
+  let size = 21
```
snowleopard, line 5:
```suggestion
  let chunk_size = 21
```

otherlibs/stdune/src/compact_position.ml
```diff
@@ -0,0 +1,134 @@
+module Position = struct
+  (* We encode the position in three 21-bit chunks: [cnum][lnum][bol] *)
+  type t = int
```
snowleopard, line 3: That isn't going to work 32-bit platforms. Maybe that's not a big deal because Dune was already broken on them?

otherlibs/stdune/src/compact_position.ml
```diff
@@ -0,0 +1,134 @@
...
+
+  let to_position t ~fname:pos_fname =
+    let pos_bol = bol t in
+    let pos_cnum = cnum t in
+    let pos_lnum = lnum t in
+    { Lexing.pos_bol; pos_cnum; pos_lnum; pos_fname }
+end
+
+module Same_line_loc = struct
+  (* we encode the location in 4, 15 bit [bol][lnum][start_cnum][stop_cnum]
```
snowleopard, line 39:
```suggestion
  (* we encode the location in four 15-bit chunks: [bol][lnum][start_cnum][stop_cnum]
```

otherlibs/stdune/src/compact_position.ml
```diff
@@ -0,0 +1,134 @@
+module Position = struct
+  (* We encode the position in three 21-bit chunks: [cnum][lnum][bol] *)
```
snowleopard, line 2:
```suggestion
  (* We encode the position in three 21-bit fields: [cnum][lnum][bol] *)
```

otherlibs/stdune/src/compact_position.ml
```diff
@@ -0,0 +1,134 @@
...
+
+  let to_position t ~fname:pos_fname =
+    let pos_bol = bol t in
+    let pos_cnum = cnum t in
+    let pos_lnum = lnum t in
+    { Lexing.pos_bol; pos_cnum; pos_lnum; pos_fname }
+end
+
+module Same_line_loc = struct
+  (* we encode the location in four 15-bit chunks: [bol][lnum][start_cnum][stop_cnum]
```
snowleopard, line 39:
```suggestion
  (* we encode the location in four 15-bit fields: [bol][lnum][start_cnum][stop_cnum]
```

otherlibs/stdune/src/compact_position.ml
```diff
@@ -0,0 +1,134 @@
...
+
+     Note that this leaves us with 3 spare bits. We should probably use them to
+     expand [bol] and [lnum] a little. *)
+  type t = int
+
+  let size = 15
```
snowleopard, line 45:
```suggestion
  let field_size = 15
```

snowleopard (approved):
I wonder if it is really worth fiddling with bits in this way, but if it brings significant performance benefits, then maybe it's OK. Also, see my question about 32-bit platforms.

One thing I dislike is how this PR treats the new module as second class compared to `Lexbuf.Position`. I think it should be renamed from `Compact_position` to just `Position` but maybe that's somehow complicated.

rgrinberg:
> I wonder if it is really worth fiddling with bits in this way, but if it brings significant performance benefits, then maybe it's OK.

Yeah, it's a pretty nice boost. 4% faster null builds, 20% less major collections, 2% less live words, 5% shorter builds with sandboxing. Honestly, I'm a bit surprised how impactful it is. I guess we really aren't shy about saving locations everywhere.

> Also, see my question about 32-bit platforms.

It works fine on 32 bit because I disable all the optimization on 32 bits:

```
let of_loc = if Sys.int_size = 63 then of_loc else fun _ -> Loc_does_not_fit
```

I could try a little harder, but it's probably not worth it as 32 bit isn't going to be all that performant anyway.

> One thing I dislike is how this PR treats the new module as second class compared to Lexbuf.Position. I think it should be renamed from Compact_position to just Position but maybe that's somehow complicated.

Form a high level, I view `Compact_position` as just an implementation detail. I kept as its own compilation unit mostly for the convenience of testing, and because it makes it easier to work with merlin (merlin really doesn't like `module S : sig ... end = struct .. end` when you have type errors). So it's really not even revealed anywhere in the public api. Perhaps there's another convention for private modules?

snowleopard:
> Yeah, it's a pretty nice boost. 4% faster null builds, 20% less major collections, 2% less live words, 5% shorter builds with sandboxing. Honestly, I'm a bit surprised how impactful it is.

Indeed, that sounds great, and pretty surprising indeed.

 > It works fine on 32 bit because I disable all the optimization on 32 bits:

Ah, got it. Makes sense! I think this also makes me happy with the name `Compact_position`: I didn't realise at first that sometimes we need to fall back to a non-compact representation.

snowleopard:
OK, I marked most of my earlier comments as resolved, leaving just a few less important suggestions.

snowleopard (approved):
Thanks! It would be nice to check performance improvements (if any) internally.

otherlibs/stdune/src/compact_position.ml
```diff
@@ -0,0 +1,134 @@
...
+  let mask_cnum = mask_lnum lsl field_size
+
+  let small_enough =
+    let max_size = 1 lsl field_size in
+    let test int = int <= max_size in
+    fun [@inline] { Lexing.pos_bol; pos_cnum; pos_lnum; pos_fname = _ } ->
+      test pos_bol && test pos_cnum && test pos_lnum
+  ;;
+
+  let[@inline] of_position { Lexing.pos_bol; pos_cnum; pos_lnum; pos_fname = _ } =
```
tov, line 17: I think this way of defining the bit-twiddling is unnecessarily confusing. How about something like this?:

```ocaml
  let field_size = 21
  let field_mask = (1 lsl field_size) - 1
  let shift_bol = 0
  let shift_lnum = field_size
  let shift_cnum = 2 * field_size

  let[@inline] of_position { Lexing.pos_bol; pos_cnum; pos_lnum; pos_fname = _ } =
    ((pos_bol land field_mask) lsl shift_bol)
    lor ((pos_lnum land field_mask) lsl shift_lnum)
    lor ((pos_cnum land field_mask) lsl shift_cnum)
  ;;

  let[@inline] bol t = (t lsr shift_bol) land field_mask
  let[@inline] lnum t = (t lsr shift_lnum) land field_mask
  let[@inline] cnum t = (t lsr shift_cnum) land field_mask
```

otherlibs/stdune/src/compact_position.ml
```diff
@@ -0,0 +1,134 @@
...
+
+     Note that this leaves us with 3 spare bits. We should probably use them to
+     expand [bol] and [lnum] a little. *)
+  type t = int
+
+  let field_size = 15
+  let mask_bol = (1 lsl field_size) - 1 (* 15 bits set *)
+  let mask_lnum = mask_bol lsl field_size
+  let mask_start_cnum = mask_lnum lsl field_size
+  let mask_stop_cnum = mask_start_cnum lsl field_size
```
tov, line 47: What if, instead of storing `stop_cnum`, we stored `stop_cnum - start_cnum`? That would likely require fewer bits, which would let us allocate more bits to the other field.

tov (approved):
Looks good to me. Impressive performance improvement!

otherlibs/stdune/src/compact_position.ml
rgrinberg: Left a CR-someday for this. It sounds like a good change, but I'm not sure how much impact it will have in practice. We already cover the vast majority of locations with this encoding.

PR #8188, opened 2023-07-12
Solver can solve for multiple environments in single lockdir

gridbugs:
This allows dune to generate a lockdir with control over constraints on the environment. For example it's possible to include dependencies of a package behind a range of `{ os = ... }` filters, so a single lockdir can be portable between different systems. If a user wants to constrain the solver according to a certain system they can add something like this to their `dune-workspace`:
```
 context
 (default
  ...
  (solver_env
   (sys
    (os linux)))))
```

The old behaviour (using the current system to constrain the environment) can be accessed with the flag `--use-env-from-current-system`. Also there's a new flag `--just-print-solver-env` which just prints the solver env for debugging purposes.

Both the flags (e.g. `with-test`) and env vars are now stored in the `Solver_env.t` type, and this lets us remove some of the infrastructure for handling environments and simplify solving by decoupling the repo from the environment. This PR is split into multiple commits to separate refactoring changes from functional changes.

Also this change removes the switch solver context for simplicity as dune will soon be able to download its own opam repo. Heads up that this will conflict with https://github.com/ocaml/dune/pull/8105. Also I found that I was now overriding all of the built-in behaviour of the `Dir_context` context so I stopped trying to extend it.


test/blackbox-tests/test-cases/pkg/with-test-dependencies.t
```diff
@@ -102,7 +106,9 @@ Conflicting packages can't be co-installed:
   >  (name x)
   >  (depends foo conflicts-with-foo))
   > EOF
-  Error: Can't find all required versions.
+  Error: Unable to solve dependencies in build context: default
+
```
rgrinberg, line 110: This extra space seems unnecessary

test/blackbox-tests/test-cases/pkg/just-print-solver-env.t
```diff
@@ -0,0 +1,28 @@
...
+  >   (lock dune.linux.lock)
+  >   (solver_env
+  >    (sys
+  >     (os linux)))))
+  > (context
+  >  (default
+  >   (name no-doc)
+  >   (lock dune.linux.lock)
+  >   (solver_env
+  >    (flags (:standard \ with-doc)))))
```
rgrinberg, line 21: A little weird that we write `:with-doc` in the dune-project file, but `with-doc` in the workspacefile

src/dune_pkg/solver_env.ml
```diff
@@ -49,24 +53,138 @@ module Flag = struct
...
+      [ `Var_in_both_with_different_values of T.t * string * string ]
+
+    let merge a b =
+      let b_without_dup_values =
+        Map.filteri b ~f:(fun var b_value ->
+            match get a var with
+            | None -> true
+            | Some a_value -> not (String.equal a_value b_value))
+      in
+      Map.of_list (Map.to_list a @ Map.to_list b_without_dup_values)
```
rgrinberg, line 149: Is it possible to use `Map.union` here?

src/dune_pkg/opam_solver.mli
```diff
@@ -1,20 +1,17 @@
...
+  (** A message listing selected packages *)
+  val selected_packages_message :
+    t -> lock_dir_path:Path.Source.t -> User_message.t
 end
+
+val solve_lock_dir :
+     solver_env:Solver_env.t
+  -> version_preference:Version_preference.t
+  -> repo:Opam_repo.t
+  -> OpamFile.OPAM.t OpamTypes.name_map
```
rgrinberg, line 16: This argument could use a label. On a slightly funny note, all other labels used in this signature don't seem useful to me.

src/dune_pkg/opam_solver.mli
```diff
@@ -1,20 +1,17 @@
...
+  -> (Summary.t * Lock_dir.t, [ `Diagnostic_message of string ]) result
```
rgrinberg, line 17: Can you use `Pp.t` for the error message?

src/dune_pkg/opam_solver.ml
```diff
@@ -60,83 +50,215 @@ end = struct
...
+         let summary = { Summary.opam_packages_to_lock } in
+         let lock_dir =
+           match
+             List.map opam_packages_to_lock ~f:(fun opam_package ->
+                 let pkg =
+                   opam_package_to_lock_file_pkg ~repo ~local_packages
+                     opam_package
+                 in
+                 (pkg.info.name, pkg))
+             |> Package_name.Map.of_list
```
rgrinberg, line 254: `Package_name.Map.of_list_map`

src/dune_pkg/opam_solver.ml
```diff
@@ -60,83 +50,215 @@ end = struct
...
+  let pp_rejection = Fmt.nop
+
+  let opam_version_compare t a b =
+    let comparison_int =
+      OpamPackage.Version.compare (OpamFile.OPAM.version a)
+        (OpamFile.OPAM.version b)
+    in
+    match t.version_preference with
+    | Oldest -> Ordering.of_int comparison_int
+    | Newest -> Ordering.of_int (-comparison_int)
```
rgrinberg, line 113: I think this code sort of defeats the spirit of having `Ordering` to not deal with integers. Instead, you could immediately convert your `comparison_int` to `Ordering.of_int` and then use a helper function to invert the ordering by pattern matching. Perhaps we could add this helper to `Ordering` itself?

bin/pkg.ml
```diff
@@ -227,33 +227,90 @@ module Lock = struct
...
-    let repo_selection =
-      Dune_pkg.Opam.Repo_selection.add_env sys_env repo_selection
-    in
-    (* Construct a list of thunks that will perform all the file IO side
-       effects after performing validation so that if materializing any
-       lockdir would fail then no side effect takes place. *)
-    let write_disk_list =
+
+    if just_print_solver_env then
+      Fiber.return
```
rgrinberg, line 237: `let+ () = Fiber.return () in` will save you a bunch of indentation instead.

bin/pkg.ml
```diff
@@ -227,33 +227,90 @@ module Lock = struct
...
+                 ~context_name ~solver_env_from_context
+                 ~sys_bindings_from_current_system
+             in
+             User_message.make
+               [ Pp.textf "Solver environment for context %s:\n%s"
+                   (String.maybe_quoted
+                   @@ Dune_engine.Context_name.to_string context_name)
+                   (Dune_sexp.to_string @@ Dune_pkg.Solver_env.encode solver_env)
+               ]
+             |> Console.print_user_message)
```
rgrinberg, line 256: `Console.print` will eliminate a little indentation

src/dune_pkg/solver_env.mli
```diff
@@ -8,19 +8,61 @@ module Flag : sig
...
+    val set : t -> sys_var -> string -> t
+
+    val get : t -> sys_var -> string option
+
+    type merge_error =
+      [ `Var_in_both_with_different_values of sys_var * string * string ]
+
+    (** Merge two environments returning an error if they both contain a binding
+        of the same variable to different values. *)
+    val merge : t -> t -> (t, merge_error) result
```
rgrinberg, line 53: Usually we call such a function `union`

src/dune_pkg/solver_env.mli
```diff
@@ -8,19 +8,61 @@ module Flag : sig
...
 
     val fold : t -> init:'a -> f:(flag -> 'a -> 'a) -> 'a
+
+    val mem : t -> flag -> bool
   end
 end
 
-type t = { flags : Flag.Set.t }
+module Sys_var : sig
+  type t =
```
rgrinberg, line 25: Some light docs for this module and the submodule would be appreciated

src/dune_pkg/opam_repo.mli
```diff
@@ -0,0 +1,38 @@
...
+      rules out using [Path.External.t], [Path.Source.t], and [Path.Local.t].
+    - We want to support relative paths external to the source directory to
+      handle the case where someone has the opam repository checkoud out next to
+      their project, and runs: `dune pkg lock --opam-repository-path
+      ../opam-repository`. This rules out [Path.Outside_build_dir.t] as it
+      doesn't permit relative paths outside the source directory. *)
+val of_opam_repo_dir_path : Filename.t -> t
+
+(** Returns a path to the "packages" directory of the repo *)
+val packages_dir_path : t -> Filename.t
```
rgrinberg, line 29: The type of this function is weird. `Filename.t` is only for the base name of the path. I.e. it cannot contain a `/`.

This should be `string` or preferable some sort of `Path.t`

gridbugs: > Filename.t is only for the base name of the path. I.e. it cannot contain a /.

This isn't true, for example `Filename.root : string` is defined to be `"/"` and the module defines a bunch of functions for working with entire paths.

That said I think you're right in that this can be a `Path.t` instead.

test/blackbox-tests/test-cases/pkg/with-test-dependencies.t
gridbugs: fixed

test/blackbox-tests/test-cases/pkg/just-print-solver-env.t
gridbugs: That wouldn't work currently because this field is an ordered set and the ordered set language only allows keywords to start with `:` (`:standard`, `:include`, etc). I actually think it's a little weird that when using variables in dependency conditions in `dune-project` that the `:` prefix is necessary to distinguish them from strings (e.g. it's really easy to make the mistake of filtering a dependency on `(= os "linux")` (which means the same as `(= "os" "linux")`)). I appreciate that it's hard to change now but I'd much rather we disallow unquoted strings in `dune-project` dependency specifications and treat all bare words as variables. I guess for now we can say something like "the `:` prefix dereferences a variable into its contained value".

src/dune_pkg/solver_env.ml
gridbugs: I just refactored to use `Map.union`.

src/dune_pkg/opam_solver.mli
gridbugs: Updated to change which args have labels.

src/dune_pkg/opam_solver.mli
gridbugs: done

src/dune_pkg/opam_solver.ml
gridbugs: Good suggestion. I have added a new function `Ordering.reverse` which reverses the direction of a comparison function and refactored this code to use it.

bin/pkg.ml
gridbugs: Ended up getting rid of the list and using `Console.printf`

src/dune_pkg/solver_env.mli
gridbugs: done

src/dune_pkg/opam_repo.mli
gridbugs: Turns out this function is no longer used. Still I have updated the `Opam_repo` module to use `Path.t` instead of `Filename.t` internally.

bin/pkg.ml
```diff
@@ -90,9 +49,7 @@ module Lock = struct
             ~doc:
               "Path to a local opam repository. This should be a directory \
                containing a valid opam repository such as the one at \
-               https://github.com/ocaml/opam-repository. If this option is \
-               omitted the dependencies will be locked using the current \
-               switch instead.")
+               https://github.com/ocaml/opam-repository.")
```
Leonidas-from-XIV, line 52: Ah, good catch!

bin/pkg.ml
```diff
@@ -259,6 +216,43 @@ module Lock = struct
...
+        Dune_pkg.Solver_env.Sys_var.Bindings.union
+          solver_env_from_context.Dune_pkg.Solver_env.sys
+          sys_bindings_from_current_system
+      with
+      | Ok solver_env -> solver_env
+      | Error
+          (`Var_in_both_with_different_values
+            (var, value_from_context, value_from_system)) ->
+        User_error.raise
+          [ Pp.textf "Can't create solver environment for context: %s"
```
Leonidas-from-XIV, line 237: Tiny nitpick but text should probably be `Can't create solver environment for context %s`, that is without the colon as it reads weirdly.

src/dune_pkg/opam_solver.ml
```diff
@@ -33,72 +50,218 @@ end = struct
...
     ; solver_env : Solver_env.t
     }
 
-  let create ~base_context ~local_packages ~solver_env =
-    { base_context; local_packages; solver_env }
+  let create ~solver_env ~repo ~local_packages ~version_preference =
+    { repo; version_preference; local_packages; solver_env }
 
-  type rejection = Base_context.rejection
+  type rejection = |
```
Leonidas-from-XIV, line 102: Quick question why this is an empty type now?

Leonidas-from-XIV (commented):
The code looks good, I wonder about the empty type though.

src/dune_pkg/opam_solver.ml
gridbugs: It's because there's no user constraint which causes a candidate package to be rejected and we don't currently use the `availability` field to reject package versions. The `CONTEXT` interface requires that a `rejection` type be defined.

Writing this it occurs to me that maybe we should be evaluating the `availability` field based on the solver env and rejecting candidates that are unavailable.

bin/pkg.ml
```diff
@@ -270,15 +264,42 @@ module Lock = struct
...
+        value & flag
+        & info
+            [ "just-print-solver-env" ]
+            ~doc:
+              "Print an s-expression describing the environment that would be \
+               used to solve dependencies and then exit without attempting to \
+               solve the dependencies or generate the lockfile. Intended to be \
+               used to debug situations where no solution can be found to a \
+               project's dependencies.")
+    in
```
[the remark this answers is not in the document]
gridbugs: I'm imagining the situation where a user is confused because `dune pkg lock` is producing unexpected results and so they run `dune pkg lock --help`. Since this flag is to help users debug this specific scenario I think it makes sense for it to be discoverable via `dune pkg lock --help` and I think making it a flag of `dune pkg lock` is the simplest way to achieve this.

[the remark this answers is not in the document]
Alizter: I agree with @gridbugs here. In the future when pkg support is more fleshed out, we can look at adding such a feature to `show`. But for now the indirection doesn't seem to be worth it.

src/dune_pkg/opam_solver.ml
gridbugs: I just added availability checking so now it's possible for packages to be rejected based on the `available` filter evaluating to false.

test/blackbox-tests/test-cases/pkg/just-print-solver-env.t
rgrinberg: > I appreciate that it's hard to change now but I'd much rather we disallow unquoted strings in dune-project dependency specifications and treat all bare words as variables.

Yeah, that would create a ton of inconsistency of how we handle dune files. For better or for worse, we treat `foo` and `"foo"` as equivalent in positions where both are allowed.

> That wouldn't work currently because this field is an ordered set and the ordered set language only allows keywords to start with : (:standard, :include, etc).

This isn't a fundamental limitation and we can lift it or introduce some explicit form for variables whenever there's ambiguity . E.g. `(var foo)`.

For now we can leave this as is, but this is a rather big inconsistency to expose to users.

rgrinberg:
Thanks. Merged with a little clean up commit.

PR #8189, opened 2023-07-12
Move Package.Dependency.Constraint to Dune_lang

gridbugs:
This will let us use the constraint language within lockfiles which we need in order for build and install commands be conditional on opam filters.

### DIFF
--- a/src/dune_lang/package_constraint.ml
+++ b/src/dune_lang/package_constraint.ml
@@ -1,0 +1,121 @@
+open Stdune
+
+module Op = struct
+  type t =
+    | Eq
+    | Gte
+    | Lte
+    | Gt
+    | Lt
+    | Neq
+
+  let equal a b =
+    match (a, b) with
+    | Eq, Eq | Gte, Gte | Lte, Lte | Gt, Gt | Lt, Lt | Neq, Neq -> true
+    | _ -> false
+
+  let map =
+    [ ("=", Eq); (">=", Gte); ("<=", Lte); (">", Gt); ("<", Lt); ("<>", Neq) ]
+
+  let to_dyn t =
+    Dyn.variant (fst (List.find_exn ~f:(fun (_, op) -> equal t op) map)) []
+
+  let encode x =
+    let f (_, op) = equal x op in
+    (* Assumes the [map] is complete, so exception is impossible *)
+    List.find_exn ~f map |> fst |> Dune_sexp.Encoder.string
+end
+
+module Var = struct
+  type t =
+    | Literal of string
+    | Var of string
+
+  let encode = function
+    | Literal v -> Dune_sexp.Encoder.string v
+    | Var v -> Dune_sexp.Encoder.string (":" ^ v)
+
+  let decode =
+    let open Dune_sexp.Decoder in
+    let+ s = string in
+    if String.is_prefix s ~prefix:":" then Var (String.drop s 1) else Literal s
+
+  let to_dyn = function
+    | Literal v -> Dyn.String v
+    | Var v -> Dyn.String (":" ^ v)
+end
+
+type t =
+  | Bvar of Var.t
+  | Uop of Op.t * Var.t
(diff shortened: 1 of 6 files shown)


src/dune_lang/package_constraint.mli
```diff
@@ -0,0 +1,33 @@
...
+        (** A variable name such as :version or :with-test (the leading ":" is
+            excluded from the string) *)
+end
+
+type t =
+  | Bvar of Var.t
+  | Uop of Op.t * Var.t
+  | Bop of Op.t * Var.t * Var.t
+  | And of t list
+  | Or of t list
```
Alizter, line 27: Can you add comments for these?

Leonidas-from-XIV (commented):
I think this PR makes a good deal of sense, some comments that popped into my head while I was reviewing below.

src/dune_lang/package_constraint.ml
```diff
@@ -0,0 +1,121 @@
...
+    | Var of string
+
+  let encode = function
+    | Literal v -> Dune_sexp.Encoder.string v
+    | Var v -> Dune_sexp.Encoder.string (":" ^ v)
+
+  let decode =
+    let open Dune_sexp.Decoder in
+    let+ s = string in
+    if String.is_prefix s ~prefix:":" then Var (String.drop s 1) else Literal s
```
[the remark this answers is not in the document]
gridbugs: I'll do that in a separate PR

src/dune_rules/package.mli
```diff
@@ -33,27 +33,27 @@ module Id : sig
...
-      | Gt
-      | Lt
-      | Neq
-
-    val to_relop : t -> OpamParserTypes.FullPos.relop
-  end
-
   module Constraint : sig
+    module Op : sig
+      type t = Dune_lang.Package_constraint.Op.t =
```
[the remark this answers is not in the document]
gridbugs: also `Var.t`

src/dune_lang/package_constraint.mli
gridbugs: Done

gridbugs:
In the process of documenting the different `Constraint.t` constructors I noticed it was possible to have a `Bvar` which contained a string literal. Even though the parser was incapable of producing such a value, the fact that it could be represented made the API harder for me to understand. I've changed it so that only variables can be passed to the `Bvar` constructor and made a clear distinction between the `Variable.t` and `Value.t` types (previously string literals were in a type called `Var.t` which made them seem like variables and I've always found that confusing).

rgrinberg:
The change in the description looks fine, but I see you've started pushing some additional work on top. Was that intentional?

src/dune_lang/package_constraint.mli
```diff
@@ -0,0 +1,47 @@
...
+    | Variable of Variable.t
+end
+
+(** A constraint for use in package definitions, equivalent to the filter
+    language used in opam package definitions. All expressions in this language
+    are booleans. String literal values can only appear as the argument to an
+    operator which returns a boolean. Variables are dynamically typed and can
+    represent strings and booleans. If a variable appears in a position where a
+    boolean is expected it will be assumed to represent a boolean. *)
+type t =
```
Alizter, line 29: Dumb question: Is our blang really that different from this?

gridbugs:
> The change in the description looks fine, but I see you've started pushing some additional work on top. Was that intentional?

Oops I accidentally pushed some additional work here. I just removed it.

src/dune_lang/package_constraint.mli
gridbugs: Good question. I don't know off the top of my head. Of course they both have similar functions though the package constraint language has the added requirement of being translatable to the opam filter language. At some point soon I'll be looking at actually evaluating conditions (currently we only use these to generate opam files) and at that point I'll see whether it makes sense to just replace this with blang.

PR #8193, opened 2023-07-12
fix(rpc): correct severity of error diagnostic

Alizter:
The severity of diagnostics sent over RPC seemed to be missing.

- [x] changelog


rgrinberg:
Can you add a changes entry?

PR #8194, opened 2023-07-12
rpc: add Comparable module for Diagnostics

Alizter:
I want to be able to have sets of diagnostics for the internal state of dune monitor. In order to do this, I made the `Diagnostic` type comparable. I wasn't able to finish the definition fully so I would appreciate some help doing so. I'm not even sure this is the best way to go about it.

cc @rgrinberg

### DIFF
--- a/otherlibs/dune-rpc/private/exported_types.ml
+++ b/otherlibs/dune-rpc/private/exported_types.ml
@@ -6,6 +6,8 @@ module Loc = struct
     ; stop : Lexing.position
     }
 
+  let compare = Stdune.Lexbuf.Loc.compare
+
   let start t = t.start
 
   let stop t = t.stop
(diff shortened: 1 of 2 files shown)


otherlibs/dune-rpc/private/exported_types.ml
```diff
@@ -212,16 +249,71 @@ module Diagnostic = struct
...
+        ; loc
+        ; severity
+        ; promotion
+        ; directory
+        ; related
+        } : Ordering.t =
+      let open Ordering.O in
+      let= () = List.compare ~compare:Target.compare t.targets targets in
+      let= () = Id.compare t.id id in
+      (* let= () = Pp.compare t.message message in *)
```
Alizter, line 277: For example here I don't know what to do with the messages.

rgrinberg:
this should be moved to dune monitor. I don't want to add new api's to the rpc client unless we're certain they're going to be used

Alizter:
@rgrinberg I've made it a separate PR for review and discussion, I can merge it back into the other PR later np.

PR #8195, opened 2023-07-13
changelog creator

Alizter:
This PR adds an interactive script that adds a changelog entry to `doc/changes/`. This can be invoked using
```
make changelog
```
The script will ask you some (optional) information about your change / PR and then create the entry file. It comes with a small cram test about its basic behaviour. The changelog entry is correctly formatted and laid out etc.

Another rule with alias `@compile-changelog` will compile the changes in `doc/changes` and diff them to the `CHANGES.md` changelog. This is meant to be called during each release. This can be invoked using
```
make compile-changelog
```
This will run this rule which will diff. This can be then promoted.

- [x] Document changelog process in HACKING.rst
- [x] Fix issue with `%{deps}` being unable to expand in `cat` action. (The compile rule relies on this). https://github.com/ocaml/dune/pull/8196

#### Motivation

Having a single CHANGES.md file is nice and simple, but it causes a lot of unnecessary work when rebasing and resolving conflicts.
- It's hard to have to resolve conflicts in the changelog when there are many PRs that touch a similar area.
- Older PRs might need to update where the changelog entry is being added, which can be easily missed.

This PR solves both these issues by making each entry its own file and leaving it to the release master to compile the end changelog.


Leonidas-from-XIV:
Sure, we could have a feature where dune-release looks at `UNRELEASED` and prepends it to `CHANGES.md` (potentially asking for a version) before tagging a release.

Alizter:
Please note that there is no UNRELEASED file in this PR, but rather a directory with separate files for the entries. There is an action that prepends all of them.

Leonidas-from-XIV:
Quick observation: I think it would be useful if the helper script were to save my username (or at least have it as an CLI option), given my GH username is rather long and prone to typos.

Alizter:
@Leonidas-from-XIV I've changed it so that the username is always written to `doc/changes/username` and read from there if it exists. This has been added to .gitignore. I've also made it so that it errors if a changelog entry clashes with an preexisting one.

All of this behaviour is in the cram test.

christinerose (approved):

rgrinberg:
I'm closing this until @emillon wants to use this.

PR #8196, opened 2023-07-13
test: expand %{deps} in (cat)

Alizter:
We fix a bug in which `%{deps}` with more than one dep would not expand inside of a `cat`.

- [x] changelog
- [x] test
- [x] version guard


src/dune_rules/action_unexpanded.ml
```diff
@@ -68,6 +68,9 @@ module Action_expander : sig
     (* Evaluate a path in a position of dependency, such as in [(cat <dep>)] *)
```
Alizter, line 68: This comment is kind of funny now.

src/dune_rules/action_unexpanded.ml
```diff
@@ -241,6 +244,14 @@ end = struct
         Value.to_path_in_build_or_external v
           ~error_loc:(String_with_vars.loc sw) ~dir:t.dir
 
+      let expand_paths t sw =
+        let+ v, vs = expand t ~mode:At_least_one sw in
```
Alizter, line 248: This seemed sensible to me.

src/dune_rules/action_unexpanded.ml
```diff
@@ -304,6 +315,23 @@ end = struct
       let fn = Expander.expand_path env sw in
       register_dep fn ~f:Option.some env acc
 
+    let register_deps x ~f env acc =
```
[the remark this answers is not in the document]
Alizter: I've left only `register_deps` now.

Alizter:
Do we want to version guard this? I could see it becoming an issue otherwise.

Alizter:
@rgrinberg I've managed to version guard it now. The test demonstrates this behaviour too.

Alizter:
@emillon Could you also take a look, since you recently implemented the `cat` of a list feature.

PR #8198, opened 2023-07-13
fix: update digest and incremental db versions

rgrinberg:
https://github.com/ocaml/dune/pull/8052 updated the representation for
these, but didn't bump the version.

### DIFF
--- a/src/dune_engine/cached_digest.ml
+++ b/src/dune_engine/cached_digest.ml
@@ -54,7 +54,7 @@ module P = Persistent.Make (struct
 
   let name = "DIGEST-DB"
 
-  let version = 5
+  let version = 6
 
   let to_dyn = to_dyn
 end)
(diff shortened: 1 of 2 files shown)


Alizter (approved):

snowleopard (approved):
Oops, thanks. Did you find it because the change broke something?

Alizter:
I got a few segfaults that disappeared after cleaning today, but I was tweaking rpc yesterday too so it might be unrelated.

snowleopard:
> I got a few segfaults that disappeared after cleaning today, but I was tweaking rpc yesterday too so it might be unrelated.

@Alizter I see, thanks!

@rgrinberg Could you also add a test to prevent such breakage in future? We should basically make the format change observable somewhere, forcing us to update the version when needed.

rgrinberg:
> Could you also add a test to prevent such breakage in future? We should basically make the format change observable somewhere, forcing us to update the version when needed.

Coming up soon

PR #8200, opened 2023-07-13
test: only rely on utop binary rather than package

Alizter:
The package dependency has absolute paths for the man pages causing issues on platforms such as Nix when running the test-suite. Since we are really only relying on the utop binary, we refine this dependency so that these tests can be run on Nix.


rgrinberg:
Did you confirm that this is indeed the case? You could try running `utop` with strace and observing which paths it tries to open for example.

Alizter:
Checking strace after utop is executed in a few tests tells me that only the META files of the utop package are accessed. That's probably because utop is in the findlib paths. I don't see anything else being accessed. Therefore I believe it is safe to only rely on the binary.

rgrinberg:
If the META file is accessed, why should we remove the dependency on it?

Alizter:
Looks like this isn't the way.

PR #8201, opened 2023-07-13
test: use custom wait for watching/helpers.sh

Alizter:
This allows the fs-memo.t test to succeed on Nix. Before the pid of Dune was being thrown away by a bash bug causing wait to think it was not a child process and rejecting it. We therefore wait instead for all child processes.

### DIFF
--- a/test/blackbox-tests/test-cases/watching/helpers.sh
+++ b/test/blackbox-tests/test-cases/watching/helpers.sh
@@ -22,7 +22,16 @@ with_timeout () {
 
 stop_dune () {
     with_timeout dune shutdown;
-    wait $DUNE_PID;
+    # On Linux, we may run into a bash pid aliasing bug that causes wait to
+    # reject the pid. Therefore we use tail to wait instead.
+    if [ "$(uname -s)" = "Linux" ]
+    then
+        # wait for all child processes
+        tail --pid=$DUNE_PID -f /dev/null;
+    else
+        # wait for dune to exit
+        wait $DUNE_PID;
+    fi
     cat .#dune-output;
 }
 


Leonidas-from-XIV (approved):
Amazing if this will fix the `./helpers.sh: line 25: wait: pid 79252 is not a child of this shell` bug I get on every run.

I was implementing something similar (sleep in a loop) but maybe it would be useful to automatically abort after a certain amount of time/iterations have passed to prevent e.g. CI instances from busywaiting.

Alizter:
@Leonidas-from-XIV The issue with that is it will be very system dependent. Something that takes 1s on my machine may take a lot longer on another. I chose the .1s value on purpose so to not be too slow and not keep the OS busy.

test/blackbox-tests/test-cases/watching/helpers.sh
```diff
@@ -22,7 +22,24 @@ with_timeout () {
...
+    # Instead of using wait, we loop and poll /proc/ instead. In containarised
+    # environments such as Nix, processes are vulnerable to PID aliasing. This
+    # means that implementations of bash prune their process history. Thiss
+    # causes wait to assume that the process is not a child process and hence
+    # cannot be waited on. As of writing, bash has had this bug for +10 years,
+    # so it won't be fixed anytime soon.
+    #
+    # We wait for ~ 5 min max as to not loop forever in case of a bug.
+    loop_count=0;
+    while [ -e /proc/$DUNE_PID ];
```
rgrinberg, line 36: We don't want linux only things in this part of the test suite

rgrinberg (changes requested):

Alizter:
@rgrinberg I didn't know that `/proc` wasn't available outside of linux. I've changed it now to wait on all children processes. This now works on Nix and should be MacOS compatible.

test/blackbox-tests/test-cases/watching/helpers.sh
```diff
@@ -22,7 +21,8 @@ with_timeout () {
 
 stop_dune () {
     with_timeout dune shutdown;
-    wait $DUNE_PID;
+    # wait for *any* child process
```
rgrinberg, line 24: Is this comment correct? I think this is waiting for *all* rather than *any* child processes.

Alizter, line 24:
```suggestion
    # wait for *all* child process
```

rgrinberg: We should also just restore the old behavior outside of linux. E.g. check `uname -s` and dispatch on that.

Alizter: Done.

pmwhite:
I just opened #8303 to fix this same issue, but then closed it because this already exists. I don't have a preference either way, but in case it hasn't already been considered, switching from `wait` to `tail` fixes the problem as well. I don't know if it is cross-platform enough, though - I'm just running the tests on a Linux machine.

test/blackbox-tests/test-cases/watching/helpers.sh
```diff
@@ -22,7 +22,16 @@ with_timeout () {
 
 stop_dune () {
     with_timeout dune shutdown;
-    wait $DUNE_PID;
+    # On Linux, we may run into a bash pid aliasing bug that causes wait to
+    # reject the pid. Therefore we wait for all child processes instead.
+    if [ "$(uname -s)" = "Linux" ]
+    then
+        # wait for all child processes
+        wait $(jobs -p);
```
rgrinberg, line 30: Could you use phil's tail hack instead of this. it seems like it's a more faithful emulation of waiting for `$pid` than waiting for all jobs.

rgrinberg:
>I don't know if it is cross-platform enough, though - I'm just running the tests on a Linux machine.

It's not. But it's the better workaround on Linux

Alizter:
@rgrinberg I've used `tail` instead now. Works on my machine.

PR #8204, opened 2023-07-13
refactor: describe_workspace path sanitation

Alizter:
We refactor how describe_workspace handles path sanitiation by switching from manual string manipulations to their Path equivalents. This makes the test reproducible on Nix.


rgrinberg:
@esope could you take a look at this one?

esope (approved):
This looks good to me.

PR #8205, opened 2023-07-13
test: fix github2629 repro on nix

Alizter:
This test was relying on the implicit opam install location behaviour. We make the install location explicit with the --prefix argument to make the test reproducible on Nix.


test/blackbox-tests/test-cases/github2629.t
```diff
@@ -9,7 +9,7 @@
   $ export DUNE_BUILD_DIR=$(mktemp -d -t github2629XXXXXX);
   > dune build @install;
   > cat $DUNE_BUILD_DIR/default/foo.install | sed s#$DUNE_BUILD_DIR#DUNE_BUILD_DIR#g;
-  > dune install --dry-run
+  > dune install --dry-run --prefix DUNE_BUILD_DIR
```
rgrinberg, line 12: Why are we installing into the build directory? That seems really weird. You should pass some dummy dir like `$PWD/foo`.

Alizter: Good point, it doesn't really matter. I put a dummy directory now.

PR #8207, opened 2023-07-13
fix(merlin): make merlin output more reproducible

Alizter:
I'm putting a vbox around the merlin sexp so that it always breaks. On Nix the paths are long so it was causing some of the output to break. This should make the test more reproducible.

I'm not 100% confident this will work so keeping as draft to see what CI reports.


Alizter:
This is not the way.

PR #6242, opened 2022-10-15
feature: Show number of failed jobs during build

Alizter:
Looks like this if there are failures:
```sh
Done: 83% (3217/3849, 632 left, 1 failed) (jobs: 0)
```
and looks like this if there are none:
```sh
Done: 83% (3217/3849, 632 left) (jobs: 0)
```

PR #8212, opened 2023-07-14
feature(rpc): add failed jobs to progress

Alizter:
In order to achieve feature sync with the regular dune display and monitor, we need to send the number of failed jobs which is tracked in the build engine progress. When I added this progress statistic in #6242 I purposefully didn't touch RPC.

I'm not sure if we need to bump the version for this. I've made the number of failed jobs an optional argument so perhaps that will guarantee backwards compatibility.

cc @rgrinberg what do you think?

- [x] bump version?
- [x] changelog

### DIFF
--- a/otherlibs/dune-rpc-lwt/examples/rpc_client/rpc_client.ml
+++ b/otherlibs/dune-rpc-lwt/examples/rpc_client/rpc_client.ml
@@ -32,11 +32,12 @@ let () =
          | Some Success -> "Success"
          | Some Failed -> "Failed"
          | Some Interrupted -> "Interrupted"
-         | Some (In_progress { complete; remaining }) ->
+         | Some (In_progress { complete; remaining; failed }) ->
            Printf.sprintf
-             "In_progress { complete = %d;  remaining = %d }"
+             "In_progress { complete = %d;  remaining = %d;  failed = %d }"
              complete
              remaining
+             failed
          | Some Waiting -> "Waiting"
        in
        print_endline (Printf.sprintf "Got progress_event: %s" message);
--- a/otherlibs/dune-rpc/private/procedures.ml
+++ b/otherlibs/dune-rpc/private/procedures.ml
@@ -101,13 +101,82 @@ module Poll = struct
   let name t = t.name
 
   module Progress = struct
+    module V1 = struct
+      type t =
+        | Waiting
+        | In_progress of
+            { complete : int
+            ; remaining : int
+            }
+        | Failed
+        | Interrupted
+        | Success
+
+      let sexp =
+        let open Conv in
+        let waiting = constr "waiting" unit (fun () -> Waiting) in
+        let failed = constr "failed" unit (fun () -> Failed) in
+        let in_progress =
+          let complete = field "complete" (required int) in
+          let remaining = field "remaining" (required int) in
+          constr
+            "in_progress"
+            (record (both complete remaining))
+            (fun (complete, remaining) -> In_progress { complete; remaining })
+        in
+        let interrupted = constr "interrupted" unit (fun () -> Interrupted) in
+        let success = constr "success" unit (fun () -> Success) in
+        let constrs =
+          List.map ~f:econstr [ waiting; failed; interrupted; success ]
+          @ [ econstr in_progress ]
+        in
+        let serialize = function
+          | Waiting -> case () waiting
+          | In_progress { complete; remaining } -> case (complete, remaining) in_progress
+          | Failed -> case () failed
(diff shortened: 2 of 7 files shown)


rgrinberg:
You need to use our RPC versioning mechanism to add this field. In particular, a new version of the method that provides this type is warranted to support older clients.

Alizter:
@rgrinberg I've updated the versioning code. Is this what you had in mind?

I cannot use the original type in the V1 module since the list where we register the procedures would have to be heterogeneous.

rgrinberg:
@pmwhite do you want to give this a review since you've touched this code recently?

otherlibs/dune-rpc/private/procedures.ml
```diff
@@ -118,12 +118,48 @@ module Poll = struct
...
+          | Interrupted -> case () interrupted
+          | Success -> case () success
+        in
+        sum constrs serialize
+    end
+
     let name = "progress"
 
     let v1 =
+      Decl.Request.make_current_gen ~req:Id.sexp ~resp:(Conv.option V1.sexp)
```
pmwhite, line 156: If I understand the versioning mechanism correctly, this should be a call to `Decl.Request.make_gen`, which takes additional parameters to upgrade/downgrade to/from the current generation's types. I assume that `make_current_gen` _would_ work just fine, but the disadvantage is that the serialization and upgrade logic is now intertwined. With `make_gen`, we can leave the old sexp function exactly the way it was, and have a small, isolated bit of upgrade and downgrade logic.

To be concrete, I'm suggesting that the old version of `Progress.t` live in the above `V1` module, and the extra parameters to `make_gen` should define how to go back and forth between `Progress.t` and `V1.t`.

Alizter:
FTR, I've tested an old client and new server (and vice versa) and the versioning is wokring correctly.

otherlibs/dune-rpc-lwt/examples/rpc_client/rpc_client.ml
```diff
@@ -34,9 +34,12 @@ let () =
            | Some Success -> "Success"
            | Some Failed -> "Failed"
            | Some Interrupted -> "Interrupted"
-           | Some (In_progress { complete; remaining }) ->
-             Printf.sprintf "In_progress { complete = %d;  remaining = %d }"
+           | Some (In_progress { complete; remaining; failed }) ->
+             Printf.sprintf "In_progress { complete = %d;  remaining = %d%s }"
                complete remaining
+               (match failed with
```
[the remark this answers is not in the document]
Alizter: I can reorganize the code here to be less inlined, but the point is that we don't want to keep the users eyes busy. I don't think reading `0 failed` is particularly useful or needed.

otherlibs/dune-rpc/private/procedures.ml
Alizter: I'm not sure I understand what this achieves. If we introduce the old type in V1, then we have two choices for `v1` either it is a `Progress.t` in which case now we will need the old sexp converter anyway, in addition to new conversion functions. Otherwise, we can make `v1` have `V1.t option`, but then the list of versions is no longer homogeneous.

Alizter: I guess it does separate the conversion and serialization code however.

otherlibs/dune-rpc-lwt/examples/rpc_client/rpc_client.ml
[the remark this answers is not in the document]
Alizter: Good point, fixed.

pmwhite (approved):

CHANGES.md
```diff
@@ -3,6 +3,9 @@ Unreleased
 
 - No longer emit linkopts(javascript) in META files (#8168, @hhugo)
 
+- The `progress` RPC procedure now has an extra field for the `In_progress`
```
rgrinberg, line 6: Might as well move this entry to doc/changes

otherlibs/dune-rpc/private/procedures.ml
pmwhite: Oops, just checked my email and saw these messages; github hid them after the commit was pushed. 

Yeah, the separation was all I was after, and the reason I care about that is that each time a new version is introduced, we have to go back and update all prior versions as well. With the separation, we can compose v1_to_v2 and v2_to_v3 to get v1_to_v3 for free; I think it's also less error-prone if we have the rule that we never touch the serialization code of past versions.

rgrinberg:
Thanks.

PR #8213, opened 2023-07-14
fix: value tracking mechanism for [Persistent]

rgrinberg:
Every instantce of [Persistence] will now need to provide a
[test_example] that can be constructed of the data being stored. This
data is used to record a digest number that will be committed. Anytime,
the test data type changes, the example should also change and so should
the digest. This way, we won't forget to update the digest version.

### DIFF
--- a/test/expect-tests/persistent_tests.ml
+++ b/test/expect-tests/persistent_tests.ml
@@ -1,0 +1,52 @@
+open Stdune
+module Persistent = Dune_util.Persistent
+module Digest = Dune_digest
+
+let test (type a) (module Persistent : Persistent.Desc with type t = a)
+    (example : a) =
+  let digest = Digest.generic example |> Digest.to_string in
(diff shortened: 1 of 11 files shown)


test/expect-tests/persistent_tests.ml
```diff
@@ -0,0 +1,41 @@
+open Stdune
+module Persistent = Dune_util.Persistent
+module Digest = Dune_digest
+
+let test (type a) (module Persistent : Persistent.Desc with type t = a)
+    (example : a) =
+  let digest = Digest.generic example |> Digest.to_string in
+  printfn "%s version %d\n%s\n---\n" Persistent.name Persistent.version digest
+
+let%expect_test "persistent digests" =
```
snowleopard, line 10: Could you add some comments to this file instructing what to do (bump the version) when a test fails?

snowleopard (approved):
Thanks!

PR #8217, opened 2023-07-16
makefile: slightly easier bootstrapping

Alizter:
We introduce a phony "dune.exe" target in the makefile that will always build the bootstrapped version of dune. Before the only way to bootstrap properly was to use `make _boot/dune.exe` and many times this would not work so it would have to be changed to `make -B _boot/dune.exe`.

This new target is easier to remember and write, and allows devs to quickly rebootstrap dune when needed. This is not something that needs to be done very often, but we make it easier anyway.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,10 @@ TEST_OCAMLVERSION := 4.14.1
 help:
 	@cat doc/make-help.txt
 
+.PHONY: bootstrap
+bootstrap:
+	$(MAKE) -B _boot/dune.exe
+
 .PHONY: release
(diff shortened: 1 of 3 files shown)


rgrinberg:
Instead of making a phony target named after a file, why not just make it `make bootstrap`?

Alizter:
@rgrinberg Because we already have such a target.

rgrinberg:
That one is just a clone of the `release` target so you can get rid of it

Alizter:
@rgrinberg done.

PR #8218, opened 2023-07-16
doc: fix backticks in hacking.rst

Alizter:


doc/hacking.rst
```diff
@@ -381,7 +381,9 @@ For project names, use the following capitalization:
 - **OCaml**
 - **OCamlFormat**, and ``ocamlformat`` is the command.
 - ``odoc``, always in monospace.
-- **opam**. Can be capitalised as Opam in titles and at the beginning of sentences only, as the official name is formatted opam. The command is ``opam``.
+- **opam**. Can be capitalised as Opam in titles and at the beginning of
```
christinerose, line 384:
```suggestion
- **opam**. Can be capitalised as Opam at the beginning of
```
This is a very strange one. It’s been decided that capitalization of opam is only at the beginning of sentences and not in titles/subtitles. Unless the maintainers want to change that, it was the official word I got from Thomas last year.

doc/hacking.rst
```diff
@@ -381,7 +381,9 @@ For project names, use the following capitalization:
...
+  sentences only, as the official name is formatted opam. The command is
```
christinerose, line 385:
```suggestion
  sentences only, as the official name is formatted opam. Even in titles, headers, and subheaders, it should be all lowercase: opam. The command is
```

christinerose (commented):
Looks great!
Just the Opam vs opam thing, as it’s a bizarre case.

doc/hacking.rst
Alizter: Well don't we use title case for the titles and subtitles? I think this just says write `opam` and you can capitialize it as `Opam` as you would with any other word.

Alizter: In that case, I can update the wording here. Thanks!

Alizter:
@christinerose I've pushed a new commit here that clarifies the capitalisation of opam. Thanks!

emillon (approved):

PR #8219, opened 2023-07-16
doc: fix dune-workspace example version number lagging behind

Alizter:
I don't think this was intentionally 3.2.

### DIFF
--- a/doc/dune-files.rst
+++ b/doc/dune-files.rst
@@ -781,7 +781,7 @@ containing exactly:
 
 .. code:: dune
 
-    (lang dune 3.2)
+    (lang dune 3.10)
     (context default)
 
 This allows you to use an empty ``dune-workspace`` file to mark the root of your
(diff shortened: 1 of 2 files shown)


doc/dune-files.rst
```diff
@@ -781,7 +781,7 @@ containing exactly:
 
 .. code:: dune
 
-    (lang dune 3.2)
+    (lang dune 3.10)
```
emillon, line 784: I agree it doesn't need to be 3.10, but in reality it shouldn't be `3.10` either, nor kept up to date.

For `dune-workspace` files in particular I never put anything other that `1.0` (or whatever minimum version works for the context of the doc).

That's a larger discussion, but I think `doc/test/run.t` is a bit of a distraction because it does not reflect how dune users upgrade their `dune-project` files. Encountering lang versions less than the dune binary version should be a daily experience, so I don't think we should avoid that in the docs.

Alizter: OK, in that case I'll close this PR. I thought it was left behind when looking through the workspace docs.

PR #8220, opened 2023-07-17
chore: add devcontainer

### LINKED ISSUE #7297: setup GitHub codespaces
We should setup a `.devcontainer.json` file in the repo so that we can load the dockerfiles in the repo for a dev env in the browser.

Alizter:
This will allow developers to set up a dev environment in VSCode very easily even in a web browser.

I've tested it and its possible to run the entire test suite with this.

- fix https://github.com/ocaml/dune/issues/7297

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,8 @@
+{
+	"name": "OCaml",
+	"image": "mcr.microsoft.com/devcontainers/base:bullseye",
+	"features": { "ghcr.io/avsm/ocaml-devcontainers-feature/ocaml:latest": {} },
+	"postCreateCommand": "opam init -ay --disable-sandboxing && sudo chown vscode _build && sudo apt-get update && make dev-switch && sudo apt install -y file npm",
+	"mounts": ["source=${localWorkspaceFolderBasename}-ocaml-build,target=${containerWorkspaceFolder}/_build,type=volume"],
+	"remoteUser": "vscode"
+}


Alizter:
This is from @avsm by the way. I've modified it a bit and am testing it out currently.

avsm:
I'll need to cut a stable release of my devcontainers feature if you're going to use this.  It works well, but I won't get a chance to do this until the next few weeks. (for example, I've moved the tools into a /usr/local/ocaml/version subdirectory in the devcontainer so they dont conflict with existing tools, and can compose better with other languages).

3Rafal:
I have created some simple devcontainers based on opam image. You can find them here:
https://github.com/ocaml/ocaml.org/blob/main/.devcontainer/Dockerfile
https://github.com/3Rafal/ocaml-devcontainer/

I think having stuff that is currently in post create command moved to separate dockerfile might be a bit cleaner (and better for caching).

Alizter:
Unfortunately we cannot cache the opam repository since almost every package we install depends on the Dune we just built. It is possible to instead use an older version of Dune for the opam packages, however if we run into incompatibilities between version it will be very confusing.

I'm not sure if it is worth it. We have a similar story in our Nix flake where we provide a "correct" dev env and a "fast" one with cached older versions.

I should also note that we already have a dev docker file in the repo, but I wasn't actually sure how we were meant to use it in the dev container.

Alizter:
@rgrinberg I could spend some time making this faster to setup but I think merging this now is a good starting point. If there is demand for a faster containers in the future we can look into it.

avsm:
Hold on; I still need to do a stable release of my repo; I'll take a look tomorrow.

@3Rafal the difference between our devcontainers is that I use the 'features' mechanism which is a bit newer.  Instead of a Dockerfile-per-language, features allow multiple languages to be installed in one image.

rgrinberg:
Seems like this is better than nothing. Feel free to send subsequent improvements.

PR #8223, opened 2023-07-18
fix: dune init project checks that PATH is relative

Alizter:
This should always be the case since the project directory might not exist, there is little point allowing for absolute directories here. We therefore enforce this with a nice user message instead of failing badly.
- [x] changelog


Alizter:
I don't feel like this is the route to go down.

PR #8224, opened 2023-07-18
chore: remove commented out code in Path

Alizter:
I don't know if it was intentional to keep this around. Perhaps @rgrinberg knows.


Leonidas-from-XIV (approved):
I feel commented out code is more confusing than helpful, especially if the original purpose of it becomes forgotten.

rgrinberg:
At some point I knew, I no longer remember why.

PR #8225, opened 2023-07-18
feature(pkg): Creation of files from `.in` templates

### LINKED ISSUE #7698: Building Opam Packages: Substitutions
Opam allows modifying the build step to insert "substitutions". These are `.in` files that are interpreted and their percent forms replaced with their values. We need a similar feature in dune to build these packages. To make this feature make sense from dune's POV, it's enough to add a `(subst <input> <output>)` action that can be used in the build step.

Leonidas-from-XIV:
This PR adds the support for `substs` as they exist in OPAM files.

Closes: #7698 (thanks @gridbugs)

### DIFF
--- a/src/dune_pkg/substs.ml
+++ b/src/dune_pkg/substs.ml
@@ -36,8 +36,6 @@ end
 
 module Map = Var.Map
 
-let opam_variable = Option.map ~f:(fun v -> OpamVariable.S v)
-
 let subst env self ~path ~target =
   let self' = self |> Dune_lang.Package_name.to_string |> OpamPackage.Name.of_string in
   let env full_variable =
@@ -49,8 +47,8 @@ let subst env self ~path ~target =
     in
     let key = { Var.T.package; variable } in
     match Map.find env key with
-    | Some _ as v -> opam_variable v
-    | None -> Map.find env { Var.T.package = Some self; variable } |> opam_variable
+    | Some _ as v -> v
+    | None -> Map.find env { Var.T.package = Some self; variable }
   in
   (* The OPAM API always needs .in as a suffix for the basename *)
   let target_template = Path.extend_basename target ~suffix:".in" in
--- a/src/dune_pkg/substs.ml
+++ b/src/dune_pkg/substs.ml
@@ -36,7 +36,7 @@ end
 
 module Map = Var.Map
 
-let subst env self ~path ~target =
+let subst env self ~src ~dst =
   let self' = self |> Dune_lang.Package_name.to_string |> OpamPackage.Name.of_string in
   let env full_variable =
     let variable = OpamVariable.Full.variable full_variable in
@@ -50,16 +50,7 @@ let subst env self ~path ~target =
     | Some _ as v -> v
     | None -> Map.find env { Var.T.package = Some self; variable }
   in
-  (* TODO The OPAM API always needs .in as a suffix for the basename. We should
-     fix this. *)
-  let temp_dir = Temp.create Dir ~prefix:"dune" ~suffix:"substitute" in
-  Exn.protect
-    ~finally:(fun () -> Temp.destroy Dir temp_dir)
-    ~f:(fun () ->
-      let target_template = Path.relative temp_dir (Path.basename target ^ ".in") in
-      Io.copy_file ~src:path ~dst:target_template ();
-      let path = target |> Path.to_string |> OpamFilename.Base.of_string in
-      OpamFilter.expand_interpolations_in_file env path;
-      Path.rename target_template target;
-      Temp.destroy Dir temp_dir)
+  let src = OpamFilename.of_string (Path.to_string src) in
+  let dst = OpamFilename.of_string (Path.Build.to_string dst) in
+  OpamFilter.expand_interpolations_in_file_full env ~src ~dst
 ;;
--- a/src/dune_pkg/substs.ml
+++ b/src/dune_pkg/substs.ml
@@ -50,9 +50,16 @@ let subst env self ~path ~target =
     | Some _ as v -> v
     | None -> Map.find env { Var.T.package = Some self; variable }
   in
-  (* The OPAM API always needs .in as a suffix for the basename *)
-  let target_template = Path.extend_basename target ~suffix:".in" in
-  Path.rename path target_template;
-  let basename = target |> Path.to_string |> OpamFilename.Base.of_string in
-  OpamFilter.expand_interpolations_in_file env basename
+  (* TODO The OPAM API always needs .in as a suffix for the basename. We should
+     fix this. *)
+  let temp_dir = Temp.create Dir ~prefix:"dune" ~suffix:"substitute" in
+  Exn.protect
+    ~finally:(fun () -> Temp.destroy Dir temp_dir)
+    ~f:(fun () ->
+      let target_template = Path.relative temp_dir (Path.basename target ^ ".in") in
+      Io.copy_file ~src:path ~dst:target_template ();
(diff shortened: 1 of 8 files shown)


src/dune_lang/action.ml
rgrinberg, line 1: Why do we need this change?

It's customary to specify input and outputs explicitly in dune actions.

Leonidas-from-XIV: It mirrors the opam `subst` directive, which only uses the basename and the `.in` suffix is implied.

(Relatedly, the OPAM API does the same, only accepting the basename, but of course we can do some renaming)

rgrinberg: This is a convention that we don't need to copy from opam. In dune, actions specify their inputs and outputs explicitly.

rgrinberg: I imagine you can handle multiple inputs and outputs with `(progn ..)` or `(concurrent ..)`. So I would use those until they aren't enough.

src/dune_pkg/substs.ml
```diff
@@ -0,0 +1,12 @@
+open Stdune
+
+let subst _env ~path ~target =
+  let env full_variable =
+    let _variable = OpamVariable.Full.variable full_variable in
+    None
```
Alizter, line 6: What is this about? Does the opam code work via side effects?

Leonidas-from-XIV: No, this is just temporary code as I try to hook it up with the Env.

src/dune_pkg/substs.mli
```diff
@@ -0,0 +1,25 @@
+open Stdune
+
+module Var : sig
+  type t =
+    { package : Dune_lang.Package_name.t option
+    ; variable : string
```
rgrinberg, line 6: Let's use proper types for this stuff. No strings in the public API please.

src/dune_rules/pkg_rules.ml
```diff
@@ -352,31 +351,50 @@ end
 
 module Substitute = struct
   module Spec = struct
-    type ('path, 'target) t = 'path * 'target
+    type ('path, 'target) t =
+      string Substs.Map.t * Dune_lang.Package_name.t * 'path * 'target
```
rgrinberg, line 355: This is one way to do it, but it's actually less efficient. In particular, we're doing all the work to produce this map when generating the action, but we might not necessarily run the action. Instead of generating this map, we can produce the expander when running the action in the same as we do it elsewhere in the code - by running `Action_expander.expander`.

So all we need to save into the action are the inputs required to call this function.

Or, we can even save the expander that we used at rule generation and fetch it from there.

rgrinberg (changes requested):

gridbugs:
Related issue: https://github.com/ocaml/dune/issues/7698

src/dune_pkg/substs.mli
```diff
@@ -0,0 +1,25 @@
+open Stdune
+
+module Var : sig
```
gridbugs, line 3: It seems like we're defining lots of different types of variables in different places. I'm currently looking at expanding variables in build/install commands and the set of available variables in substs is the same. I was planning to represent each variable as a pform and then use the existing pform expander to resolve variables to their values. Would that work for substitutions as well?

src/dune_rules/pkg_rules.ml
Leonidas-from-XIV: I thought about this as well. In particular `misc` is not available in the substitutions as calling it would raise an error and wrapping everything in lazy doesn't help as encoding would trigger the `Code_error` in `Paths.get` anyway when encoding.

The problem with saving the expander is that `encode`ing the expander is encoding a lot of irrelevant stuff, the issue with creating an expander is that even if I save the `Context_name.t` and `Pkg.t` to create an expander using `expander`, it will miss out on the `Withenv`'d values as these get added to the expander as it traverses the syntax tree.

src/dune_pkg/substs.mli
```diff
@@ -0,0 +1,25 @@
...
+
+  val compare : t -> t -> Ordering.t
+
+  include Comparable_intf.S with type key := t
+
+  val to_dyn : t -> Dyn.t
+end
+
+module Map : sig
+  include module type of Var.Map
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Pretty sure it is equivalent or maybe worse, I just did a lot of `include module type of` stuff so I tend to overuse it. Just the alias is nicer.

src/dune_pkg/substs.mli
Leonidas-from-XIV: Yes, I assume it would work as well, the only thing required there is a function that translates from an opam "full variable" to a string with the semantics of OPAM (like rendering bools and string lists).

src/dune_pkg/substs.mli
Leonidas-from-XIV: Done. I couldn't think of anything better so I created a wrapper around `OpamVariable.t` with some conversion functions into `Dune_lang.t` and `Dyn.t`.

src/dune_rules/pkg_rules.ml
```diff
@@ -615,6 +689,8 @@ module Action_expander = struct
             in
             env, update :: updates)
       in
+      let expander = { expander with env } in
```
rgrinberg, line 692: Isn't this a separate bug? Did you end up writing a test for this?

Leonidas-from-XIV: I didn't, because to show the expander is not updated with the bindings to execute the action it needs to access the expander bindings. I looked at it and it seems that withenv does the substitution first so by the time the action is executed it doesn't really matter what bindings the expander contains.

I can remove this drive-by fix, show it is not working in the cram test here and make a separate bug fix PR if that is preferable.

Leonidas-from-XIV: Notably I wrote a test running nested `withenv`s, hoping the inner `withenv` wouldn't see the outer bindings but it works on current `main`.

src/dune_rules/pkg_rules.ml
```diff
@@ -575,6 +575,7 @@ module Action_expander = struct
       (* values set with withenv *)
       Env.Map.map expander.env ~f:Env_update.string_of_env_values
       |> Env.Map.to_list_map ~f:(fun variable value ->
+        (* TODO why is [package = None]? *)
```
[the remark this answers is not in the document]
rgrinberg: > If you use withenv I'd say it doesn't define which package the value belongs to, (withenv ((= FOO bar)) (substitute foo.in foo)) would probably assume the variable is %{FOO}% and not %{package-name:FOO}%.

Might be best to just ask the opam maintainers what's correct. At the end of the day, we just want to replicate what they're doing accurately.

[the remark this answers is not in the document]
Leonidas-from-XIV: This is what the OPAM manpage states:

```
       OPAMVAR_var overrides the contents of the variable var when substituting `%{var}%` strings in `opam` files.

       OPAMVAR_package_var overrides the contents of the variable package:var when substituting `%{package:var}%` strings in `opam` files.
```

These are environment variables, so I've implemented the first form, a variable that is not attached to a package. I could extend the implementation that a variable of the form `PKG_<anything>` will parse as package name. My concern is that this makes sense in `OPAMVAR_package_var` because the prefix makes it clear that we're dealing with packages, but `(withenv ((= LD_LIBRARIES_PRELOAD ...) ...)` would confuse people if `LD` suddenly parsed as package name.

rgrinberg:
I ended up tweaking the OPAM api to just allow to choose the file paths explicitly

PR #8226, opened 2023-07-18
path: change is_file to be false if doesn't exist

Alizter:
Before it would be true if the file didn't exist since it was just the negation of is_directory.

I used this existence checking in #8195.


rgrinberg:
How about we just stop using this confusing function altogether and just use `stat` everywhere?

Alizter:
OK will do.

PR #8227, opened 2023-07-18
Add conditions to lockfile actions

gridbugs:
This adds an optional condition to build and install actions in lockfiles. Note that these conditions currently aren't checked. Conditions on actions is necessary to build opam packages as opam package manifests can contain build and install commands which are conditional on filters.

### DIFF
--- a/src/dune_pkg/opam.ml
+++ b/src/dune_pkg/opam.ml
@@ -213,8 +213,9 @@ let opam_package_to_lock_file_pkg ~repo_state ~local_packages opam_package =
     |> List.map ~f:(fun name ->
            (Loc.none, Package_name.of_string (OpamPackage.Name.to_string name)))
   in
-  { Lock_dir.Pkg.build_command = None
-  ; install_command = None
+  (* TODO set these from the commands in the opam file *)
+  { Lock_dir.Pkg.build_commands = []
+  ; install_commands = []
   ; deps
   ; info
   ; exported_env = []
--- a/src/dune_lang/package_constraint.ml
+++ b/src/dune_lang/package_constraint.ml
@@ -31,6 +31,8 @@ module Variable = struct
 
   let of_name name = { name }
 
+  let equal { name } t = String.equal name t.name
+
   let to_dyn { name } = Dyn.record [ ("name", Dyn.string name) ]
 
   let encode { name } = Dune_sexp.Encoder.string (":" ^ name)
@@ -50,6 +52,12 @@ module Value = struct
     | String_literal s -> Dyn.variant "String_literal" [ Dyn.string s ]
     | Variable v -> Dyn.variant "Variable" [ Variable.to_dyn v ]
 
+  let equal a b =
+    match (a, b) with
+    | String_literal a, String_literal b -> String.equal a b
+    | Variable a, Variable b -> Variable.equal a b
+    | _ -> false
+
   let encode = function
     | String_literal s -> Dune_sexp.Encoder.string s
     | Variable v -> Variable.encode v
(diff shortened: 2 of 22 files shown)


src/dune_pkg/opam.ml
```diff
@@ -213,8 +213,9 @@ let opam_package_to_lock_file_pkg ~repo_state ~local_packages opam_package =
     |> List.map ~f:(fun name ->
            (Loc.none, Package_name.of_string (OpamPackage.Name.to_string name)))
   in
-  { Lock_dir.Pkg.build_command = None
-  ; install_command = None
+  (* TODO set these from the commands in the opam file *)
```
Leonidas-from-XIV, line 216: I think this is an important next step. Should this happen in this PR?

Leonidas-from-XIV (commented):
I think adding the evaluation of the conditions in this PR would be an useful addition, but if you think that just adding the groundwork here is sufficient, I'm ok with this PR.

gridbugs:
> I think adding the evaluation of the conditions in this PR would be an useful addition, but if you think that just adding the groundwork here is sufficient, I'm ok with this PR.

My rationale for making this PR without evaluation is to try to reduce the amount of time I spend resolving merge conflicts. Also I find that one way to deal with long turnaround time caused by me being in a very different timezone from most dune devs is to make more small PRs rather than fewer larger PRs as I can be working on the next change while the current change is in review.

rgrinberg:
The syntax introduced in this PR is a little intrusive I find. We already have some keywords for doing things conditionally, let's just resuse those. For example:

```
(enabled_if (= %{os} foo) (run ...))
```

The only thing that's missing is a way to enable conditional arguments. I imagine something, we can come up with some list operations for that.

gridbugs:
I'd be happy to change the syntax to something lighter but I don't think we should re-use `enabled_if` because it takes a blang expression as its argument and it would be difficult to convert dependency constraints to blangs for reasons I describe [here](https://github.com/ocaml/dune/pull/8232#issuecomment-1643472109). Action conditions should be in the same syntax as we use in dune-project files to describe dependency conditions as this language was designed to correspond to opam filters.

Also `enabled_if` doesn't take an action as an argument in other places in dune and I think if we do use it we should be consistent in that regard.

What about something like:
```
(conditional (= :os foo) (run ...))
```

Also I don't really understand why this syntax needs to be concise. I thought in most cases lockfiles will be generated automatically so I'd rather err on the side of verbosity to help (human) readers understand their contents.

gridbugs:
That all sounds reasonable. As a prerequisite for conditional build commands we'll need to make the opam env vars available as variables in `String_with_vars`. I suggest doing so with a namespace similar to `ocaml-config`, so we could say something like `%{opam:os}` and `%{opam:arch}`. Possibly we want to use a more general name than `opam` here though. Not sure.

Also regarding booleans, I just noticed this text in the blang docs:

> After an expression is evaluated, it must be exactly the string true or false to be considered as a Boolean. Any other value will be treated as an error.

I previously missed that detail, but it looks like this already gives us the ability to represent boolean literals as the strings "true" and "false" which solves the original problem I foresaw evaluating boolean variables.

rgrinberg:
>  I suggest doing so with a namespace similar to ocaml-config, so we could say something like %{opam:os} and %{opam:arch}. Possibly we want to use a more general name than opam here though. Not sure.

If you look into `Pform`, you will notice that I've already added a bunch of opam forms. I didn't add any prefixes because some of them seemed general enough that they would be useful in general and their values would be identical to the same dune vars.

That decision might be worth revisiting. But for now, I suggest just keeping it simple.

gridbugs:
Closing this in favour of slang dsl

Issue #924, opened 2018-06-28
Adding some conditionals in the language


trefis:
This is similar to #819 but with a somewhat broader scope.
The idea would be to support `(if <cond> <then> [else])` at the toplevel of a jbuild.

This would allow encoding the example of #819, which was:
```scheme
(alias
 ((name runtest)
  (deps (foo.exe))
  (action (run ${<}))
  (enabled_if (and (<> ${os} win32) (>= ${ocaml_version} 4.0.5)))))
```

as
```scheme
(if (and (<> ${os} win32) (>= ${ocaml_version} 4.0.5))
    (alias
     ((name runtest)
      (deps (foo.exe))
      (action (run ${<})))))
```
(of course I have nothing with the sugar proposed in #819).

But would also allow doing things like:
```scheme
(if (= ${ocaml_version} "4.07._")
    (copy_files# 407/*.ml{,i}))
(if (= ${ocaml_version} "4.06._")
    (copy_files# 406/*.ml{,i}))
...
```
which are needed for merlin and currently require us to use ocaml jbuilds.

---

I'm only proposing to support this at the toplevel, but this could be pushed down to arbitrary places (after all, `(select ...)` already does similar things).

avsm:
Related to https://github.com/ocaml/dune/issues/838

trefis:
That absolutely fine by me.

PR #8229, opened 2023-07-18
feature: case statements

Alizter:
We add a `(case ...)` constructor to dune_lang for actions which when expanded will reduce down to the matching branch. Basic use looks like:

```lisp
 (rule
  (alias foo)
  (action
   (case %{read-lines:version}
    (1 (echo "version is 1"))
    (2 (echo "version is 2"))
    (3 (echo "version is 3"))
    (_ (echo "shouldn't happen, got %{read-lines:version}")))))
```
which together with a `version` file containing `2` would print `2`.

- [ ] changelog
- [x] tests
- [ ] Should the default case be optional? I think no.
- [ ] Multicase branches?
- [ ] Also implement cond?
  - will be done in followup PR, currently blocked by #8196
-  addreses part of https://github.com/ocaml/dune/issues/924

### DIFF
--- a/src/dune_lang/action.ml
+++ b/src/dune_lang/action.ml
@@ -160,6 +160,7 @@ type t =
   | Patch of String_with_vars.t
   | Substitute of String_with_vars.t * String_with_vars.t
   | Withenv of String_with_vars.t Env_update.t list * t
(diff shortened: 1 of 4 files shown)


Alizter:
I'm gonna open another PR at a later date with cond.

PR #8234, opened 2023-07-19
fix(sendfile): reuse fd before fallback

### LINKED ISSUE #8210: _build/default/bin/dune.exe: Permission denied (ecryptfs, Mint)
## Expected Behavior

I can upgrade Dune.

## Actual Behavior

I get a permission denied error.

## Reproduction

I currently have Dune 3.7.1 installed. When I ran `opam upgrade`, Dune 3.9.1 cannot build due to a permission error.

```
[ERROR] The compilation of dune.3.9.1 failed at "./_boot/dune.exe build dune.install --release --profile dune-bootstrap -j 15".

#=== ERROR while compiling dune.3.9.1 =========================================#
# context     2.1.4 | linux/x86_64 | ocaml-base-compiler.5.0.0 | https://opam.ocaml.org#7109cb08
# path        ~/.opam/5.0.0/.opam-switch/build/dune.3.9.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build ./_boot/dune.exe build dune.install --release --profile dune-bootstrap -j 15
# exit-code   1
# env-file    ~/.opam/log/dune-208893-44c9eb.env
# output-file ~/.opam/log/dune-208893-44c9eb.out
### output ###
# Error: _build/default/bin/dune.exe: Permission denied
# -> required by _build/default/bin/dune.exe
# -> required by _build/install/default/bin/dune
# -> required by _build/default/dune.install
```

When I try to upgrade to 3.8.0, 3.8.1, or 3.8.2, I get the error https://github.com/ocaml/dune/issues/8041. When I try to upgrade to 3.8.3, 3.9.0, or 3.9.1, I get this current error. https://github.com/ocaml/dune/pull/8049, which fixes the sendfile error and was included in 3.8.3, is probably the cause of the current error.

My Linux kernel version is `5.15.0-76-generic`, I am running Linux Mint 21.1.

emillon:
Fixes #8210

If the `sendfile` branch failed, it left an empty target file.
This file needs to be removed, otherwise the next `open` call can fail if the file is not writeable.

### DIFF
--- a/otherlibs/stdune/src/io.ml
+++ b/otherlibs/stdune/src/io.ml
@@ -101,7 +101,7 @@ module Copyfile = struct
     | Linux -> `Sendfile
     | Windows | Other -> `Nothing
 
-  let sendfile =
+  let sendfile_with_fallback =
     let setup_copy ?(chmod = Fun.id) ~src ~dst () =
       match Unix.openfile src [ O_RDONLY ] 0 with
       | exception Unix.Unix_error (Unix.ENOENT, _, _) -> Error `Src_missing
@@ -147,7 +147,12 @@ module Copyfile = struct
         raise (Sys_error message)
       | Ok (src, dst, src_size) ->
         Exn.protect
-          ~f:(fun () -> sendfile ~src ~dst src_size)
+          ~f:(fun () ->
+            try sendfile ~src ~dst src_size
+            with Unix.Unix_error (EINVAL, "sendfile", _) ->
+              let ic = Unix.in_channel_of_descr src in
+              let oc = Unix.out_channel_of_descr dst in
+              copy_channels ic oc)
           ~finally:(fun () ->
             Unix.close src;
             Unix.close dst)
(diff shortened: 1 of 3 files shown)


otherlibs/stdune/src/io.ml
```diff
@@ -179,6 +179,7 @@ module Copyfile = struct
   let sendfile_with_fallback ?chmod ~src ~dst () =
     try sendfile ?chmod ~src ~dst ()
     with Unix.Unix_error (EINVAL, "sendfile", _) ->
+      Unix.unlink dst;
```
Alizter, line 182: Is this the correct place to do this, or can it be done inside `sendfile` when it aborts?

Alizter: I think there is some similar cleanup for src there too.

emillon: I thought as well, but this happens in the prepare phase, not when `sendfile` itself fails.

rgrinberg, line 182: What if this call fails?

The way we handle the sendfile fallback seems error prone in general. We should should move the fallback code directly into `Copyfile.sendfile` and only catch errors from the surrounding `sendfile` syscall.

This way we can resume the fallback with the existing open fd rather than deleting and re-opening.

emillon: I just did that, let me know what you think.

rgrinberg (approved):
Thanks. This is indeed simpler.

Alizter (approved):

PR #8235, opened 2023-07-19
refactor: use Exn.protect instead of protectx

emillon:
This adds a layer of rebinding that obscures what is going on.


Alizter:
Was this done on purpose, perhaps so the values are not mutated?

rgrinberg (approved):
The previous version allocated one tuple, while this one allocates two closures.

None of this matters for this bit of code so feel free to re-arrange it according to your tastes.

Issue #8231, opened 2023-07-19
Displaying full filenames in dune-site before load


maroneze:
## Desired Behavior

I'd like `dune-site` to have more fine-grained ways to obtain debugging/logging information about which files it is loading.

More specifically, I'm trying to debug why Frama-C (WP plug-in in particular) fails to load when using a Cygwin-compiled OCaml (without MinGW), while it works when a MinGW-based installation is present. Currently, all I have is the message:

    Dynlink.Error (Dynlink.Cannot_open_dll "Dynlink.Error (Dynlink.Cannot_open_dll \"Failure(\\\"Permission denied\\\")\")")

But I have no idea which file is causing this, and I don't see an easy way to get this information. I tried using `strace`, and it seems to indicate something related to `.opam\4.14.1\lib\ocaml\nums.cmxs`, but it's hard to understand the log.

And since dune-site is the one actually responsible for defining the actual filename (after lookup and possibly some conversions, including concatenation of the directory name), when I try to debug it from Frama-C, I get no useful feedback (only that the plug-in that failed to load is WP, which I already knew).

Ideally, some environment variable, or some other setting (e.g. an optional argument) should allow outputting the specific information about which file(s) and directory(ies) are being loaded, so that if something fails later, the user can more easily understand what happened.

## Example

Assuming e.g. a `DUNE_SITE_VERBOSE=1` environment variable, when `Dune_site_plugins.V1.load` is called, it would print (to stderr or somewhere else) phrases such as `dune-site: Loading '<file>'`, so that any subsequent errors would be easier to debug.

If there is already something like that, please let me know so I can try it.

Alizter:
You can try ``--always-show-command-line` which will show a shell command approximation of the action that raised the error message. You can then inspect the path to see where the error might be coming from.

maroneze:
Sorry, I'm a bit confused about where to add such option. [Dune-site](https://dune.readthedocs.io/en/stable/sites.html) is used in Frama-C to dynamically load files during execution, so the command-line is not Dune's, but Frama-C's; it's the OCaml code that calls `dune-site` during execution. Even if I try `dune exec --always-show-command-line -- frama-c`, it doesn't seem to do anything.

For me, either an API change (adding optional `verbose` or `debug` arguments to the `load` function) or some environment variables would be required to make it work.

Alizter:
[...]

PR #8236, opened 2023-07-19
site: add debug flag

Alizter:
We add a debug flag DUNE_SITE_DEBUG which will output that a dynlinked file is being loaded. This can help users of dune site debug their developments, especially when dynlink gives unhelpful error messages.

- fix https://github.com/ocaml/dune/issues/8231
- [x] tests


otherlibs/dune-site/src/plugins/linker/toplevel/linker.ml
```diff
@@ -11,6 +11,8 @@ open Toploop [@@ocaml.warning "-33"]
 let load filename =
   let buf = Buffer.create 16 in
   let ppf = Format.formatter_of_buffer buf in
+  if Dune_site_private.debug then
+    Printf.printf "DUNE_SITE_DEBUG[toplevel]: loading %s\n" filename;
```
emillon, line 15: Can you add a test that uses that code path?

otherlibs/dune-site/test/site-debug.t
```diff
@@ -0,0 +1,83 @@
+Testing the debug flags for dune-site
+
+  $ mkdir -p b c
+
+  $ for i in b; do
```
emillon, line 5: can you inline this? (same below)

emillon (commented):
this'll need a review from someone more familiar with dune-site (@bobot / @rgrinberg maybe)

otherlibs/dune-site/test/site-debug.t
emillon: I mean, instead of `for i in b; do ... done`, can you replace the `$i` by `b` in `...`?

emillon: another option is to add `DUNE_SITE_DEBUG=1 dune exec ...` to an existing test.

Alizter: I'll change some existing tests then.

otherlibs/dune-site/src/plugins/linker/toplevel/linker.ml
emillon: If the original issue has been fixed with just the other print, I'd say we leave that part out of the first version.

Alizter: There is no issue that we are actually fixing here. We are simply giving the user of the library the option to output some extra debugging information. I don't see why it would be harmful to print stuff even if we aren't testing it here?

The use case is extremely niche and is really a quality-of-life fix.

Alizter:
Since there doesn't appear to be any interest in this I am closing.

PR #8237, opened 2023-07-20
Expose user environment when running pkg actions

gridbugs:
Currently nixos users are unable to run pkg tests that run system actions because the `$PATH` environment variable is unset in the environment in which the action runs resulting in the shell being unable to find standard unix tools such as `cat`. This is intended as a quick fix to unblock nixos users but it introduces a new problem of builds being (unintentionally) influenced by environment variables. This will require some careful design so we'll solve it later.

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -327,7 +327,14 @@ module Pkg = struct
         ]
     in
     let env = build_env t |> Env.Map.map ~f:Env_update.string_of_env_values in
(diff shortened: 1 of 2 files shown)


Alizter (approved):
Tests run for me now.

rgrinberg:
Could you add a test demonstrating this behavior? E.g. something of the following nature:

1. create some dummy shell script in `$dir`.
2. add `$dir` to `PATH`.
3. demonstrate that the dummy shell script is usable within the build rules.

gridbugs:
Updated to add a test.

PR #8241, opened 2023-07-21
melange: add `melange.dom` lib to merlin

jchavarri:
In the more recent versions, Melange modified the way its libraries are provided, so they get all wrapped (see https://github.com/melange-re/melange/pull/624/).

This PR adapts the way Dune generates merlin configs so that it can find the artifacts for the library `melange.dom`.

Remaining questions:
- The test relies on the latest version of `melange` to be available in Dune, but we don't pin `melange` anymore, should we re-add the pinning?
- Should we do something similar with `melange.node`?

### DIFF
--- a/src/dune_rules/merlin/merlin.ml
+++ b/src/dune_rules/merlin/merlin.ml
@@ -482,6 +482,22 @@ module Unprocessed = struct
     Per_item_action_builder.map t.config.preprocess
       ~f:(pp_flags sctx ~expander t.config.libname)
 
+  let add_lib ~requires ~scope lib_name =
+    let open Memo.O in
+    let libs = Scope.libs scope in
+    Lib.DB.find libs lib_name >>= function
+    | Some lib ->
+      let+ libs =
+        let linking =
+          Dune_project.implicit_transitive_deps (Scope.project scope)
+        in
+        Lib.closure [ lib ] ~linking |> Resolve.Memo.peek >>| function
+        | Ok libs -> libs
+        | Error _ -> []
+      in
+      Lib.Set.union requires (Lib.Set.of_list libs)
+    | None -> Memo.return requires
+
   let process
       ({ modules
        ; ident = _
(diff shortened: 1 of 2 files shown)


anmonteiro (approved):
Looks good to me.

I think we don't need to worry about versioning since we ignore the failure here (as before).

anmonteiro:
> Should we do something similar with melange.node?

I was planning to leave `melange.node` out, to be added explicitly if needed.

rgrinberg:
Have we considered a scheme where this information somehow originates from melange itself?

We're already running `Melange_binary.where`, why don't we have a command that gives us al the information that we need for merlin.

Also, I think I forgot the previous discussions, but why aren't we telling dune that we are compiling against these libraries? Is it because these libraries don't work on native or bytecode?

anmonteiro:
Good point. Melange indeed knows about its stdlib location. Perhaps we should use `melc --where` here too. We're already calling it to find the stdlib path. We could use the other directories too, instead of resolving the `melange` library.

jchavarri:
> We're already running `Melange_binary.where`, why don't we have a command that gives us al the information that we need for merlin.

Should this be done in this PR, or rather in another one?

> * The test relies on the latest version of `melange` to be available in Dune, but we don't pin `melange` anymore, should we re-add the pinning?

Any thoughts on the above? I know pinning melange was problematic at some point (for CI, iirc), so I'm not sure how we want to proceed.

anmonteiro:
> Should this be done in this PR, or rather in another one?
> 
> >

I'd say on this PR.



> Any thoughts on the above? I know pinning melange was problematic at some point (for CI, iirc), so I'm not sure how we want to proceed.

Yeah, I think it'd make people have to pin melange, which in its current form only works on OCaml 5.1. Not sure what to do here either.

jchavarri:
Discussed with @anmonteiro, we will remove `melange.dom` from default libs included in Melange, so this fix won't be needed, see [melange-re/melange#695](https://github.com/melange-re/melange/issues/695).

PR #8243, opened 2023-07-21
fix: disable background digests on Windows

### LINKED ISSUE #8228: Issue building packages on windows: rmdir, Directory not empty
I'm seeing issues on windows using dune 3.9.1.

I've witnessed this behavior in jsoo and zarith CI so far.
- https://github.com/ocaml/Zarith/actions/runs/5589049596/jobs/10216750778?pr=143
- https://github.com/ocsigen/js_of_ocaml/actions/runs/5588550201/jobs/10215512355?pr=1490

The error looks like this:

```
#=== ERROR while compiling sexplib0.v0.16.0 ===================================#
# context     2.0.10 | win32/x86_64 | ocaml.5.0.0 ocaml-option-mingw.1 | git+https://github.com/ocaml/opam-repository.git
# path        D:/a/js_of_ocaml/js_of_ocaml/_opam/.opam-switch/build/sexplib0.v0.16.0
# command     D:\a\js_of_ocaml\js_of_ocaml\_opam\bin\dune.exe build -p sexplib0 -j 2
# exit-code   1
# env-file    D:/.opam/log/sexplib0-5660-51cb9c.env
# output-file D:/.opam/log/sexplib0-5660-51cb9c.out
### output ###
# Error:
# rmdir(C:\Users\runneradmin\AppData\Local\Microsoft\Windows\INetCache\dune\db\temp\dune_80f647_artifacts): Directory not empty
# -> required by
#    _build/default/src/.sexplib0.objs/sexplib0__Sexp_conv_grammar.impl.d
# -> required by
#    _build/default/src/.sexplib0.objs/sexplib0__Sexp_conv_grammar.impl.all-deps
# -> required by
#    _build/default/src/.sexplib0.objs/byte/sexplib0__Sexp_conv_grammar.cmt
# -> required by
#    _build/install/default/lib/sexplib0/sexplib0__Sexp_conv_grammar.cmt
# -> required by _build/default/sexplib0.install
# -> required by alias install
```

## Specifications

- Version of `dune` (output of `dune --version`): 3.9.1 
- Version of `ocaml` (output of `ocamlc --version`): 4.14 and 5.0
- Operating system (distribution and version): windows

emillon:
Fixes https://github.com/ocaml/dune/issues/8228

In some cases when the shared cache is enabled, some temporary directories cannot be removed.

### DIFF
--- a/src/dune_config/config.ml
+++ b/src/dune_config/config.ml
@@ -116,7 +116,10 @@ let background_digests =
   let t =
     { name = "background_digests"
     ; of_string = Toggle.of_string
-    ; value = background_default
+    ; value =
+        (match Platform.OS.value with
+        | Linux -> `Enabled
+        | _ -> `Disabled)
     }
   in
   register t;
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,10 @@ Unreleased
 - Fix permission errors when `sendfile` is not available (#8234, fixes #8120,
   @emillon)
 
+- Disable background digests on Windows. This prevents an issue where
+  unremovable files would make dune crash when the shared cache is enabled.
+  (#8243, fixes #8228, @emillon)
+
 3.9.1 (2023-07-06)
 ------------------
 


emillon:
We can probably come up with a better fix but this one would mitigate the issue.

nojb:
This fix is not great; it will leave temporary directories and files laying around that will never be cleaned up. Do you understand when is the `rm_rf` called? Could the issue be solved by delaying this call somehow until other threads are done? For example, in the `Temp` module of `stdune` we register an exit hook to clean all temporary files and directories; could that be relevant here? (I haven't looked at the code, so maybe this is precisely the context in which the bug is triggered...)

Alizter:
I don't understand how we run into this issue in the first place. When creating temporary directories in Windows, it is my understanding that we ask windows for a unique temp directory name which gets put into the common temp dir location. Then we can do whatever we like with the files inside and finally delete the temp directory since the name is unqiue.

It appears in this case we are placing a non-unique directory name (from the caching stuff?) directly ourselves into the common windows temp dirs and then getting race conditions when two processes touch the same name. Wouldn't the solution be to fix how we create and delete temp directories in windows rather than ignoring the errors?

rgrinberg:
How about we just disable background digests on windows until we understand the issue better?

emillon:
> How about we just disable background digests on windows until we understand the issue better?

Sure, that's an even better workaround. I'll change the PR to do that and add that to 3.9.2.

emillon:
done

src/dune_config/config.ml
```diff
@@ -116,7 +116,10 @@ let background_digests =
   let t =
     { name = "background_digests"
     ; of_string = Toggle.of_string
-    ; value = background_default
+    ; value =
+        (match Platform.OS.value with
+        | Linux -> `Enabled
```
rgrinberg, line 121: Why did you disable it on Mac and the rest of the unixes?

rgrinberg: Oh, I remember that we have that fork bug as well.

rgrinberg (approved):

src/dune_config/config.ml
emillon: I can introduce a `Toggle.and_`  to express that as `background_default and not windows` if you think it's more explicit.

PR #8246, opened 2023-07-23
fix: use correct hash for id's in hashtable

rgrinberg:
We were using [Int.hash] rather than the specialized [hash] for id's.

### DIFF
--- a/otherlibs/stdune/src/id.ml
+++ b/otherlibs/stdune/src/id.ml
@@ -23,7 +23,6 @@ end
 module Make () : S = struct
   module Set = Int.Set
   module Map = Int.Map
-  module Table = Hashtbl.Make (Int)
 
   type t = int
 
@@ -45,4 +44,14 @@ module Make () : S = struct
   let hash (t : t) = t
 
   let to_dyn t = Dyn.Int t
+
+  module Table = Hashtbl.Make (struct
+    type nonrec t = t
+
+    let equal = equal
+
+    let hash t = t
+
+    let to_dyn = to_dyn
+  end)
 end


otherlibs/stdune/src/id.ml
```diff
@@ -45,4 +44,14 @@ module Make () : S = struct
   let hash (t : t) = t
 
   let to_dyn t = Dyn.Int t
+
+  module Table = Hashtbl.Make (struct
+    type nonrec t = t
+
+    let equal = equal
+
+    let hash t = t
```
snowleopard, line 53: Wait, surely that's a bad hash function?

Alizter: Is this not just an integer? What difference would it make if we hashed it?

snowleopard: Sorry, I couldn't resist asking AI to do some explaining for me :laughing: Pretty great advice!

> Q: I have a hashtable keyed by integers. Why is using a trivial hash function hash key = key a bad idea?

A hashtable is an advanced data structure that enables efficient key-based insertion, deletion, and access operations. It relies on a hash function to map keys to indexes of an array. The properties of a good hash function include:

1. Uniform distribution: The hash values should evenly distribute across the hashtable irrespective of the input data distribution. If we have a hash function that maps unique keys in a non-uniform way, it will lead to an uneven distribution of data in the hashtable, where some buckets contain a large number of entries and others are left empty or with minimal entries.

2. Minimize collisions: A collision occurs when two different keys map to the same index in the hashtable. If a hash function produces a high number of collisions, it undermines the effectiveness of the hashtable and performance is degraded.

A trivial hash function such as `hash key = key` would be a poor choice because:

1. Non-uniform distribution: If the keys are not uniformly distributed, the trivial hash function `hash key = key` would also lead to a non-uniform distribution in the hashtable.

2. Collisions: Although this function doesn't give any same key-value pair collisions (no two different keys will map to the same index), clustering can occur. Clusters are formed if keys are sequential or near to each other. This reduces the speed of the hash table.

3. Waste of Space: If the keys are sparse (i.e., there are large gaps in the keys' distribution), then numerous slots in the hashtable will remain empty, wasting space. This function isn't suitable for large datasets.

4. Security: Using your data as its own hash is vulnerable to targeted collision attacks, as an attacker can identify the direct correlation between the key and its hash index in a hashtable.

Hence, it's recommended to use well-designed hash functions (like Murmur, City Hash, SHA, etc., depending on use case) which distribute data more evenly across the hashtable and reduce the chance of collisions. This can enhance the performance of the hashtable's operations.

> Q: One extra detail is that the keys are actually sequentially generated ids. Does that make the situation worse?

Yes, that does make the situation significantly worse. Using sequentially generated IDs as keys with a trivial hash function (`hash key = key`) would lead to clustering, which is a close relative of collision.

Clustering is when items get concentrated in certain sections of your hash table (in this case, from the left (lower indices) towards the right (higher indices)). For sequentially generated keys, most of the data will get densely concentrated in some regions and you'll get a lot of empty slots in your hash table. This results in:

1. **Space Inefficiency**: Many slots in the hashtable will stay unused, leading to wasted memory.
2. **Performance Degradation**: As data gets more clustered, key lookup time increases. Unlike what's desirable in a hashtable, the access time will no longer be close to O(1), instead it will degrade towards O(n) in the worst case, where n is the number of keys in a particular slot/bucket.

Therefore, it's even more critical to use a good hash function when dealing with sequentially increasing keys to ensure good data distribution across the hashtable, and to maintain the desirable performance characteristics of a hashtable.

> Q: Thank you!

You're welcome! If you have any other questions, feel free to ask. Happy coding!

rgrinberg: Indeed it's a bad hash function. Although note that the purpose of this PR is to simply make the hash of `Id.Table` agree with `Id.hash`. I fixed the hash function separately in https://github.com/ocaml/dune/pull/8258

PR #8247, opened 2023-07-23
refactor(memo): replace set with hashset

rgrinberg:
The set isn't quite as small as the comment has expected. For example,
when building dune, this set can easily grow to almost thousand
elements.

Therefore we replace it with a hashtable.


Alizter:
I see some slowdown on null builds in the fermat bench. In the autumn bench, there appears to be some reduction in major collections which is good, but other values look strange.

rgrinberg:
Indeed it's WIP for now. It looks promising but I need to understand why it uses so much more memory.

PR #8249, opened 2023-07-23
fix: correctly handle absolute paths in dune ocaml top-module

### LINKED ISSUE #7370: dune ocaml top-module crashes when given an absolute path
Reproduction case in #7369.

Alizter:
- fix #7370

### DIFF
--- a/bin/ocaml/top.ml
+++ b/bin/ocaml/top.ml
@@ -189,7 +189,10 @@ module Module = struct
   let term =
     let+ builder = Common.Builder.term
     and+ module_path =
-      Arg.(required & pos 0 (some string) None & Arg.info [] ~docv:"MODULE")
+      Arg.(
+        required
+        & pos 0 (some string) None
+        & Arg.info [] ~docv:"MODULE" ~doc:"Path to an OCaml module.")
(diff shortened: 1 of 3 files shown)


rgrinberg:
Instead of disallowing them, you could just strip the initial PWD and look for corresponding source file

Alizter:
@rgrinberg What if the prefix isn't shared? I'll have to add some logic making sure its a prefix etc.

rgrinberg:
If the prefix isn't shared, then the path is outside the workspace and there's very little we can do for the user.

Alizter:
Also I'm not sure the initial PWD is the correct thing to choose. You can imagine a project `a/b/...` where you call `dune ocaml top-module` inside `b` using the absolute path for a module inside `a`. I think we rather need the project root.

rgrinberg:
Okay sure, you can use the workspace root.

Alizter:
@rgrinberg I've implemented the prefix handling logic. There is some refactoring we should do in the future: `Workspace_root` should work with `Path.t` rather than `string`.

rgrinberg:
I ran into this annoying issue and took the liberty to revive this PR.

PR #8254, opened 2023-07-23
refactor: remove Path.is_file

Alizter:
Replace with a regular stat.


Alizter:
@rgrinberg Yes, I had a look at it earlier. It is used in a few places so it will take some time to do.

PR #8258, opened 2023-07-24
fix: use good hash function for id's and integers

rgrinberg:
Also remove some duplication between the [Id] and [Int] modules.

### DIFF
--- a/otherlibs/stdune/src/int.ml
+++ b/otherlibs/stdune/src/int.ml
@@ -12,7 +12,28 @@ include Comparable.Make (T)
 
 let equal (a : t) b = a = b
 
-let hash (t : t) = t
+(* This implementation (including the comment) is taken from the Base
+   library. *)
+
+(* This hash was chosen from here: https://gist.github.com/badboy/6267743
+
+   It attempts to fulfill the primary goals of a non-cryptographic hash function:
+
+   - a bit change in the input should change ~1/2 of the output bits
+   - the output should be uniformly distributed across the output range
+   - inputs that are close to each other shouldn't lead to outputs that are close to
+     each other.
+   - all bits of the input are used in generating the output
+
+     In our case we also want it to be fast, non-allocating, and inlinable. *)
+let[@inline always] hash (t : t) =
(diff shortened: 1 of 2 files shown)


Alizter:
This doesn't actually use a good hash function. The `hash` function in `Int` is the same bad one.

rgrinberg:
Indeed, looks like we're using bad hash functions all the way down :)

Fixed it at the root here: https://github.com/ocaml/dune/pull/8259

rgrinberg:
@snowleopard this PR combines everything

otherlibs/stdune/src/int.ml
```diff
@@ -12,7 +12,27 @@ include Comparable.Make (T)
 
 let equal (a : t) b = a = b
 
-let hash (t : t) = t
+(* this implementation (including the comment) is taken from base *)
```
snowleopard, line 15:
```suggestion
(* This implementation (including the comment) is taken from the Base library. *)
```

snowleopard (commented):
Thanks! I'm starting some benchmarks.

snowleopard (approved):

otherlibs/stdune/src/int.ml
snowleopard: Did this change get undone somehow?

rgrinberg: Which change got undone? This PR https://github.com/ocaml/dune/pull/8259 will not be merged because you asked to move all the hash stuff into one PR.

snowleopard: Never mind. All good!

snowleopard:
I got mixed results in my benchmarking but I think we should still go ahead with this change, because it would improve our performance story in the long term.

* On small builds, this change looks like a small (~1%) regression.
* On a large null build, this change looks like a small (2-3%) improvement. I've got only one sample though, so it may be a fluke.

rgrinberg:
Note that the old behavior is easily restored by choosing `let hash id = Id.to_int id` wherever it's needed and is considered to be safe.

PR #8259, opened 2023-07-24
fix: use good hash for [Int.hash]

rgrinberg:

### DIFF
--- a/otherlibs/stdune/src/int.ml
+++ b/otherlibs/stdune/src/int.ml
@@ -12,7 +12,27 @@ include Comparable.Make (T)
 
 let equal (a : t) b = a = b
 
-let hash (t : t) = t
+(* this implementation (including the comment) is taken from base *)
+
+(* This hash was chosen from here: https://gist.github.com/badboy/6267743
+
+   It attempts to fulfill the primary goals of a non-cryptographic hash function:
+
+   - a bit change in the input should change ~1/2 of the output bits
+   - the output should be uniformly distributed across the output range
+   - inputs that are close to each other shouldn't lead to outputs that are close to
+     each other.
+   - all bits of the input are used in generating the output
+
+     In our case we also want it to be fast, non-allocating, and inlinable. *)
+let[@inline always] hash (t : t) =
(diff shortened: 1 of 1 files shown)


Alizter (approved):

otherlibs/stdune/src/int.ml
```diff
@@ -12,7 +12,7 @@ include Comparable.Make (T)
 
 let equal (a : t) b = a = b
 
-let hash (t : t) = t
+let hash (t : t) = Poly.hash t
```
snowleopard, line 15: Do you know what it does?

For reference, here is an implementation from Base:

```ocaml
    (* This hash was chosen from here: https://gist.github.com/badboy/6267743

       It attempts to fulfill the primary goals of a non-cryptographic hash function:

       - a bit change in the input should change ~1/2 of the output bits
       - the output should be uniformly distributed across the output range
       - inputs that are close to each other shouldn't lead to outputs that are close to
         each other.
       - all bits of the input are used in generating the output

       In our case we also want it to be fast, non-allocating, and inlinable.  *)
    let[@inline always] hash_int (t : int) =
      let t = lnot t + (t lsl 21) in
      let t = t lxor (t lsr 24) in
      let t = t + (t lsl 3) + (t lsl 8) in
      let t = t lxor (t lsr 14) in
      let t = t + (t lsl 2) + (t lsl 4) in
      let t = t lxor (t lsr 28) in
      t + (t lsl 31)
    ;;
```

rgrinberg: Here is the implementation from the stdlib:

https://github.com/ocaml/ocaml/blob/e3bf4006d2b79fa84750dd3624e7b664ec88f1b2/runtime/hash.c#L56

The implementation in base seems a lot faster (no ffi or tag dispatch overhead). I'm happy to use it instead

snowleopard: Yeah, let's use the one from Base.

snowleopard (approved):
Thanks!

rgrinberg:
Moved the change to #8258

PR #8262, opened 2023-07-24
fix: pass O_SHARE_DELETE when digesting files

### LINKED ISSUE #8268: background digests crash on Windows
> It's a solution in the sense that it prevents dune from crashing. I'll open a new issue to track "let's reenable background digests on windows". (and I'll close this one otherwise it'll keep being closed whenever a branch containing the fix is pushed, sorry about that)

_Originally posted by @emillon in https://github.com/ocaml/dune/issues/8228#issuecomment-1649486223_

emillon:
Fixes #8268

This flag corresponds to what `Stdlib.open_in` does. When passed, the files can be removed while they are still open. This condition can happen when background digests are enabled. This is a no-op everywhere else.

This allows enabling background digests on Windows.

### DIFF
--- a/src/dune_digest/dune_digest.ml
+++ b/src/dune_digest/dune_digest.ml
@@ -17,8 +17,12 @@ end
 
 module Direct_impl : Digest_impl = struct
   let file file =
+    (* On Windows, if this function is invoked in a background thread,
+       if can happen that the file is not properly closed.
+       [O_SHARE_DELETE] ensures that the main thread can delete it even if it
+       is still open. See #8243. *)
     let fd =
-      match Unix.openfile file [ Unix.O_RDONLY ] 0 with
+      match Unix.openfile file [ Unix.O_RDONLY; O_SHARE_DELETE ] 0 with
       | fd -> fd
       | exception Unix.Unix_error (Unix.EACCES, _, _) ->
         raise (Sys_error (sprintf "%s: Permission denied" file))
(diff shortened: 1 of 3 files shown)


rgrinberg (commented):
Can we leave the old code on Unix? Channels have mutexes that we don't really need here

emillon:
I'll try the other following approaches:
- use Unix.open with `_O_NOINHERIT`
- replace close by `CloseHandle(Handle_val(v))`

emillon:
Ah, even better: adding `O_SHARE_DELETE` at the open point fixes the issue (and is a noop on nonwindows). I'll update the PR.

emillon:
There's still a part I don't completely understand (why the file is closed too late in some cases) but it might be related to the details of threading and win32unix.

src/dune_digest/dune_digest.ml
```diff
@@ -18,7 +18,7 @@ end
 module Direct_impl : Digest_impl = struct
   let file file =
     let fd =
-      match Unix.openfile file [ Unix.O_RDONLY ] 0 with
+      match Unix.openfile file [ Unix.O_RDONLY; O_SHARE_DELETE ] 0 with
```
hhugo, line 21: Shoudn't this come with a comment ? maybe referencing the gh issue ..

nojb (approved):
LGTM. Understanding why the file is not closed in the first place would be ideal of course, but this fix looks solid.

emillon:
I added a comment explaining the situation. let me know what you think.

nojb:
> I added a comment explaining the situation. let me know what you think.

LGTM

hhugo:
LGTM

PR #8263, opened 2023-07-24
fix: improve bool/char hash functions

rgrinberg:
Take the implementations from base

### DIFF
--- a/otherlibs/stdune/src/bool.ml
+++ b/otherlibs/stdune/src/bool.ml
@@ -16,4 +16,6 @@ let of_string s = bool_of_string_opt s
 
 let to_dyn t = Dyn.Bool t
 
-let hash (t : bool) = Poly.hash t
+let[@inline always] hash = function
+  | true -> 1
+  | false -> 0
--- a/otherlibs/stdune/src/char.ml
+++ b/otherlibs/stdune/src/char.ml
@@ -7,3 +7,5 @@ let is_digit = function
 let is_lowercase_hex = function
   | '0' .. '9' | 'a' .. 'f' -> true
   | _non_lowercase_hex_char -> false
+
+let[@inline always] hash c = Int.hash (code c)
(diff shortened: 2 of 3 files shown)


otherlibs/stdune/src/char.ml
```diff
@@ -7,3 +7,5 @@ let is_digit = function
 let is_lowercase_hex = function
   | '0' .. '9' | 'a' .. 'f' -> true
   | _non_lowercase_hex_char -> false
+
+let hash = code
```
Alizter, line 11: We cared about uniformity for `int`, shouldn't we also pass this through `Int.hash`?

otherlibs/stdune/src/bool.ml
```diff
@@ -16,4 +16,6 @@ let of_string s = bool_of_string_opt s
 
 let to_dyn t = Dyn.Bool t
 
-let hash (t : bool) = Poly.hash t
+let hash = function
```
Alizter, line 19: Also for this, shouldn't we use a more "uniform" choice rather than 0 and 1.

otherlibs/stdune/src/char.ml
snowleopard: Agreed.

otherlibs/stdune/src/bool.ml
snowleopard: It probably doesn't make much difference either way because there can be at most 2 keys. Any two `int`s are probably equivalent in terms of uniformity.

snowleopard: We should really be using an array instead of a hash table for Boolean keys.

otherlibs/stdune/src/char.ml
rgrinberg: I changed it use `Int.hash` as well.

Might be worth asking this at base.

otherlibs/stdune/src/bool.ml
rgrinberg: > We should really be using an array instead of a hash table for Boolean keys

Indeed, but we have no such tables at the moment. I'm only adding this so that our various compound functions that use `Bool.hash` get a faster hash function.

otherlibs/stdune/src/char.ml
snowleopard: Indeed. Thanks!

snowleopard (approved):
Looks good!

PR #8265, opened 2023-07-24
test: glob_files_rec with a relative path

rgrinberg:
We demonstrate that using relative paths in [glob_files_rec] has
unintended consequences. In this test, we install some artifacts outside
the package directory just by globbing from a parent dir.

### DIFF
--- a/test/blackbox-tests/test-cases/install/install-glob/install-glob-relative.t
+++ b/test/blackbox-tests/test-cases/install/install-glob/install-glob-relative.t
@@ -1,0 +1,44 @@
+Install a glob pattern that uses a relative path.
+
+When we use a pattern such as ../foo/*, the relative path leaks into the
+installation destination and can escape the root of the installed package.
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.6)
+  > (package (name foo))
+  > EOF
+
+  $ mkdir -p stanza stuff/xy
+  $ touch stuff/foo.txt stuff/xy/bar.txt
+
+normal install stanza in the share directory of the package:
+
+  $ cat >dune <<EOF
+  > (install
+  >  (section share)
+  >  (files stuff/foo.txt))
+  > EOF
+
+faulty stanza that install things outside the package:
+
+  $ cat >stanza/dune <<EOF
+  > (install
+  >  (files (glob_files_rec ../stuff/*.txt))
+  >  (section share))
+  > EOF
+
+  $ dune build foo.install
+
+Note that the "stuff" paths from  are now going to be installed outside the
+package.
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/install/install-glob/install-glob-relative.t
```diff
@@ -0,0 +1,33 @@
+Install a glob pattern from a relative path
```
gridbugs, line 1: Can you add to this description explaining the problem this causes. Copying the commit message here will be fine.

test/blackbox-tests/test-cases/install/install-glob/install-glob-relative.t
```diff
@@ -0,0 +1,33 @@
...
+  $ dune build foo.install
+  $ cat _build/default/foo.install
+  lib: [
+    "_build/install/default/lib/foo/META"
+    "_build/install/default/lib/foo/dune-package"
+  ]
+  share: [
+    "_build/install/default/share/stuff/foo.txt" {"../stuff/foo.txt"}
+    "_build/install/default/share/stuff/xy/bar.txt" {"../stuff/xy/bar.txt"}
+  ]
```
gridbugs, line 26: It took me a minute to work out why this isn't ok. I assume that all files installed into `share` by the package `foo` must be under `.../share/foo`. Can you add a note explaining this here.

rgrinberg:
Improved the test to describe the issue better.

How should we go about the fix? I've copied @anmonteiro because he dealt with a similar issue with `runtime_deps`.

One way would be to allow users to explicitly write out the prefix. E.g. `(glob_files_rec ../foo/*.txt as foo/)` similar to how we have `as` for normal installed files.

anmonteiro:
For the Melange rules we added a check that the runtime assets must be present under the source tree where the library is defined: 

https://github.com/ocaml/dune/blob/dfb6330c5d97bb78754e25b6321ba6e0efef3922/src/dune_rules/install_rules.ml#L46-L63

gridbugs:
I'm looking into this now. I noticed that the non-recursive `glob_files` is also affected. Also it's currently possible to get this behaviour without any globs by writing `(files (../stuff/foo.txt as ../stuff/foo.txt))`. Dune will happily let you install files outside the current package if the `as` keyword specifies a relative path as the destination.

I'm going to do two things to fix this:
1. add a check that ensures the destination of an install file entry is a descendant of the package's directory.
2. as (1) will prevent an install file entry from using globs beginning with `..` I'll add some syntax for changing the prefix of files matched by a glob as @rgrinberg suggested.

PR #8275, opened 2023-07-26
Don't warn on undefined opam variables when solving

gridbugs:
Undefined variables don't indicate a problem and packages expect comparisons with undefined variables to resolve to false.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -287,9 +287,9 @@ let solve_package_list local_packages context =
     with
     | OpamPp.(Bad_format _ | Bad_format_list _ | Bad_version _) as bad_format ->
       User_error.raise [ Pp.text (OpamPp.string_of_bad_format bad_format) ]
-    (* | unexpected_exn ->
-       Code_error.raise "Unexpected exception raised while solving dependencies"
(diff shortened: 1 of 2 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -92,24 +86,9 @@ end = struct
...
-    | FIdent (_, variable, _) when is_variable_sys variable ->
-      Error (`Undefined_sys_env_var variable)
-    | _ -> (
-      match first_variable filter with
-      | Some variable -> Error (`Undefined_var variable)
-      | None -> (
-        try Ok (OpamFilter.eval_to_bool (Fun.const None) filter)
-        with Invalid_argument msg -> Error (`Not_a_bool msg)))
+    try Ok (OpamFilter.eval_to_bool ~default:false (Fun.const None) filter)
+    with Invalid_argument msg -> Error (`Not_a_bool msg)
```
rgrinberg, line 91: Just a note: it's pretty undesirable to have to rely on catching this exception. It can be thrown by just about any function in the Stdlib so it's not reliable to use it a signal for anything in particular.

gridbugs: I agree. This is a limitation of the opam api. We can obviously change the api of our vendored copy of opam but I try to keep such changes to a minimum to simplify the eventual backporting effort.

PR #8278, opened 2023-07-26
Wasm_of_ocaml support

vouillon:


src/dune_rules/jsoo/js_of_ocaml.ml
```diff
@@ -96,30 +97,33 @@ module In_buildable = struct
         (let+ flags = Ordered_set_lang.Unexpanded.field "flags"
          and+ javascript_files =
            field "javascript_files" (repeat string) ~default:[]
-         in
+         and+ wasm_files = field "wasm_files" (repeat string) ~default:[] in
```
hhugo, line 100: We shouldn't change this branch, this is for dune lang < 3.0

src/dune_rules/jsoo/js_of_ocaml.ml
```diff
@@ -96,30 +97,33 @@ module In_buildable = struct
...
+         ; wasm_files
          })
     else
       fields
         (let+ flags = Flags.decode
          and+ javascript_files =
            field "javascript_files" (repeat string) ~default:[]
-         in
-         { flags; javascript_files })
+         and+ wasm_files = field "wasm_files" (repeat string) ~default:[] in
```
hhugo, line 115: We probably need to enable this only for newer dune lang versions. see how it's done above ` if syntax_version < (3, 0) then`

src/dune_rules/jsoo/jsoo_rules.ml
```diff
@@ -137,6 +137,8 @@ end
 
 let install_jsoo_hint = "opam install js_of_ocaml-compiler"
 
+let install_wasoo_hint = "opam install wasm_of_ocaml-compiler"
```
hhugo, line 140: has this been released on opam already ?

src/dune_rules/jsoo/jsoo_rules.ml
```diff
@@ -234,12 +254,21 @@ let exe_rule cc ~javascript_files ~src ~target ~flags =
       [ Resolve.Memo.args
           (let open Resolve.Memo.O in
            let+ libs = libs in
-           Command.Args.Deps (jsoo_runtime_files libs))
-      ; Deps (List.map ~f:Path.build javascript_files)
+           Command.Args.Deps
+             ((match ctarget with
+              | JS -> []
+              | Wasm -> wasm_runtime_files libs)
+             @ jsoo_runtime_files libs))
```
hhugo, line 261: does the wasm target depend on the jsoo_runtime_files ?

src/dune_rules/jsoo/jsoo_rules.ml
```diff
@@ -411,17 +445,21 @@ let build_exe cc ~loc ~in_context ~src ~(obj_dir : Path.Build.t Obj_dir.t)
...
     >>> link_rule cc ~runtime:standalone_runtime ~target ~obj_dir
           top_sorted_modules ~flags ~linkall ~link_time_code_gen
     >>= Super_context.add_rule sctx ~loc ~dir ~mode
-  | Whole_program ->
-    exe_rule cc ~javascript_files ~src ~target ~flags
+  | Whole_program, JS ->
+    exe_rule ~ctarget cc ~javascript_files ~wasm_files:[] ~src ~target ~flags
+    >>= Super_context.add_rule sctx ~loc ~dir ~mode
+  | _, Wasm ->
+    exe_rule ~ctarget cc ~javascript_files ~wasm_files ~src ~target ~flags
```
hhugo, line 462: Shouldn't it be an error if `Separate_compilation, Wasm` ? instead of silently ignoring cmode.

src/dune_rules/jsoo/jsoo_rules.ml
vouillon: The JavaScript runtime files are parsed to get purity information about primitives. We want also to include JavaScript files with no annotation, like `lib/virtualdom.compiled.js` in package `virtual_dom`. And I have just added a way to link JavaScript code referenced by the Wasm runtime code.

src/dune_rules/jsoo/jsoo_rules.ml
vouillon: Not yet. I think we should wait for thinks (`wasm_of_ocaml`, the WebAssembly Garbage Collection proposal, and related tools) to stabilize a bit.

hhugo: The hint is a bit misleading then. Should it say something like
`opam pin add wasm_of_ocaml https://github.com/ocaml-wasm/wasm_of_ocaml.git`
@rgrinberg, any opinion ?

src/dune_rules/jsoo/jsoo_rules.ml
vouillon: The idea is that you can switch to producing Wasm code just by setting the target, without having to specify a compilation mode as well.
Do you think it would be better to change `Super_context.js_of_ocaml_compilation_mode` to not default to `Separate_compilation` when the target is `wasm`?

hhugo: Have you considered making wasm part of `compilation_mode` ?
I'm also happy to keep this as is and revisit later if needed.

hhugo: Leave it as is then.

hhugo: Maybe add a comment in the code saying that separate compilation is coming later.

src/dune_rules/super_context.mli
```diff
@@ -38,6 +38,8 @@ val js_of_ocaml_runtest_alias : t -> dir:Path.Build.t -> Alias.Name.t Memo.t
 val js_of_ocaml_compilation_mode :
   t -> dir:Path.Build.t -> Js_of_ocaml.Compilation_mode.t Memo.t
 
+val js_of_ocaml_target : t -> dir:Path.Build.t -> Js_of_ocaml.Target.t Memo.t
```
rgrinberg, line 41: Can you inline this function in the one place where it's being used?

rgrinberg:
In general the code looks really good to me. Awaiting for approval from Hugo. Also don't forget to:

* Add some tests
* Update the CHANGES entry
* Add something to the manual

src/dune_rules/jsoo/jsoo_rules.ml
hhugo: @vouillon, do you plan to support sourcemap for the wasm target ? What's the content of the js file when target is wasm ?

hhugo:
The dune lang constraint will need to be updated to 3.11.0 (dune.3.10) as been released alrady)

src/dune_rules/jsoo/jsoo_rules.ml
vouillon: It would be nice to support them eventually. Probably not a high priority, though.
What are the issues with source maps?
The contents of the js file is currently a mix of js code generated by the js_of_ocaml linker and of javascript code directly copied from a file. I should certainly improve on this, and then one could add a sourcemap to this file.

hhugo: The issue with source map is that it might need to access sources to embed them. We do not currently track this kind of dependencies. Preventing us from using sandboxing.

rgrinberg:
@hhugo do you want this to be ready for 3.11?

vouillon:
I think this can wait.

I'm wondering whether this should work like`modes`. It makes a lot of sense to be able to generate both JavaScript and Wasm code.

otherlibs/dune-rpc/private/types.ml
```diff
@@ -35,7 +35,7 @@ end
 module Version = struct
   type t = int * int
 
-  let latest = (3, 10)
```
Alizter, line 38: If you rebase you will find we are on 3.13 now.

src/dune_rules/test_rules.ml
```diff
@@ -1,5 +1,32 @@
...
+          (* We don't know how to run tests in these cases *)
+          Memo.return []
+        | Other { kind = Js; _ } ->
+          let+ js_of_ocaml = Super_context.env_node sctx ~dir >>= Env_node.js_of_ocaml in
+          let multiple_targets =
+            match js_of_ocaml.targets with
+            | Some targets -> List.length (Js_of_ocaml.Target.Set.to_list targets) > 1
+            | None -> false
+          in
+          if multiple_targets
```
hhugo, line 23: I don't really understand the logic here, why do we have Js_of_ocaml.Ext.exe when targets is wasm only ?
This kind of pattern seems fragile/ error prone if with ever add a new target

src/dune_rules/jsoo/js_of_ocaml.ml
```diff
@@ -91,6 +93,7 @@ module In_buildable = struct
   type t =
     { flags : Ordered_set_lang.Unexpanded.t Flags.t
     ; javascript_files : string list
+    ; wasm_files : string list
```
hhugo, line 96: I think it would be nice to have a comment explaining why wasm_of_ocaml reuses the `(js_of_ocaml ..)` field instead of being its own thing

src/dune_rules/test_rules.ml
vouillon: I would like to be able to simply switch between producing a JavaScript file or a Wasm binary by using a dune profile. Since we are using the JavaScript file names in dune files or in HTML files, it seems natural to use the same JavaScript file name in both cases.

Since Wasm is not widely available, one also needs to be able to generate both a Javascript file and a Wasm binary, so that a web server can send one or the other based on the browser. In this case, we need two different file names.

Do you see a better way to deal with both of these use cases?

OlivierNicole:
I found out that adding `(targets wasm)` in the `(js_of_ocaml ...)` stanza of an executable doesn’t work, only modifying environments seems possible. Wouldn’t it be good to be able to specify the `wasm` target for executables too?

vouillon:
Superseded by #10695

PR #8283, opened 2023-07-27
Add a failing test for sites plugins with two plugins

shym:
This PR is a kind of bug report: it adds a test that is a small variation on the example in the manual with two plugins. One of those plugins comes with a test that really depends on this plugin (so it is expected to succeed whether the other plugin is available or not). It shows that a test will fail if it launches the application before all the plugins have been fully built, even if they are not required.

I suppose the root cause is the fact that the `META` files for the plugins are installed before the `.cmxs` are built.

We came across this problem with `ocaml-ci` failing from time to time in a project containing various plugins but with tests that always require only one of them.

### DIFF
--- a/otherlibs/dune-site/test/two-plugins.t
+++ b/otherlibs/dune-site/test/two-plugins.t
@@ -1,0 +1,131 @@
+Test sites plugins with two plugins
+
+The tests for one plugin should be allowed to run in parallel with the build of
+other plugins.
+
+We test that by creating two plugins:
+* plugin1 contains a test that depends on plugin1 and the app,
+* plugin2 is similar, but artificially slow to build.
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.8)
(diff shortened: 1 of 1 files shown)


Alizter:
We usually put tests about sites in the otherlibs/dune_site/test.

Alizter:
Writing a PR demonstrating the bug is more than enough. You might need to sanitize the test for it to work on macos. Best bet is greping for a common string and maybe you can include the full exception as a comment if needed.

Alizter:
cc @bobot

shym:
I added a grep that will hopefully work across platforms :crossed_fingers:
But CI now fails on macos on a test this PR does not modify ?!?

Alizter:
@shym Thanks. The failing test occasionally happens out of the blue, nothing to do with you. It might be a little while before we get round to merging this, probably when we have a fix. You are more than welcome to try to fix this.

PR #8285, opened 2023-07-27
fix: [build_dir_only_sub_dirs] sanity check

rgrinberg:
[build_dir_only_sub_dirs] should not be allowed to escape the directory
we're producing in and we have a check in the code to verify this.

The current check was buggy and was checking the directory
targets map instead. The declared directory targets indeed have this
limitation, but it's already checked elsewhere.

The fix is to correctly check the right directory set. This requires
adding an api to iterate over the sub directories.


rgrinberg:
ping @anmonteiro

anmonteiro (approved):

staronj (approved):

PR #8287, opened 2023-07-27
refactor(engine): simplify rule loading

rgrinberg:
The function [load_build_directory_exn] is quite large. This PR splits
it into a few logical parts:

* Computing which source paths to ignore
* Computing the copying rules for source files
* Validating the directory targets
* Computing the descendants that artifact deletion should leave alone


snowleopard (approved):
Thanks! I haven't checked every line, but I assume that this PR just moves some code into separate functions -- I fully support that.

PR #8288, opened 2023-07-27
fix(sendfile): Flush the out_channel used in the fallback path

### LINKED ISSUE #8284: Dune upgrades to 3.9.2 but fails to build other packages (Mint, ecryptfs)
## Expected Behavior

`opam upgrade -v` successfully upgrades my packages.

## Actual Behavior

The build of packages using Dune fails.

## Reproduction

`opam update -v`

I can still revert to the previous opam state, which has Dune 3.7.1.

## Specifications

OCaml 5, Dune 3.9.2, Linux Mint 21.2, Linux kernel 5.15.0-78-generic, ecryptfs

## Additional information 

This is the output of the `opam upgrade -v` command, there are a lot of errors that are hard for me to make sense of: https://gist.github.com/alan-j-hu/7bf8fcc6ede90c6ebe604af38745334b

This is probably still related to https://github.com/ocaml/dune/issues/8041 and https://github.com/ocaml/dune/issues/8210.

alan-j-hu:
Fixes #8284

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 Unreleased
 ----------
 
+- Fix flushing when using `sendfile` fallback (#8288, @alan-j-hu)
+
 - Add `dune show rules` as alias of the `dune rules` command. (#8000, @Alizter)
 
 - Fix `%{deps}` to expand properly in `(cat ...)` when containing 2 or more
--- a/otherlibs/stdune/src/io.ml
+++ b/otherlibs/stdune/src/io.ml
@@ -145,17 +145,23 @@ module Copyfile = struct
       | Error `Src_missing ->
         let message = Printf.sprintf "%s: No such file or directory" src in
         raise (Sys_error message)
-      | Ok (src, dst, src_size) ->
-        Exn.protect
-          ~f:(fun () ->
-            try sendfile ~src ~dst src_size
-            with Unix.Unix_error (EINVAL, "sendfile", _) ->
-              let ic = Unix.in_channel_of_descr src in
-              let oc = Unix.out_channel_of_descr dst in
-              copy_channels ic oc)
-          ~finally:(fun () ->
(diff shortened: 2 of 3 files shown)


alan-j-hu:
This code is a little ugly. Because the `| exception Unix.Unix_error (EINVAL, "sendfile", _) ->` case has separate cleanup rules, I can't use `Exn.protect` to close the `src` and `dst` descriptors. I'm trying to see if there's a cleaner way to write the code.

alan-j-hu:
Made the code cleaner and made the diff smaller by adding an explicit `flush oc` instead of having two different cleanup paths.

alan-j-hu:
Added a code comment explaining why the explicit flush is necessary.

rgrinberg:
Thanks for the fix. I tweaked a little to always close with the channel api when using channels. I think that's a little safer and makes sure that all of the channel's invariants are respected.

emillon:
Nice detective work! The fix is a bit complicated but so is the problem so I'd say that's ok. I'd feel safer with a test so I added one in #8292, to merge before this PR. Thanks!
(did you manage to test that it fixes your other opam installation issues?)

emillon:
I rebased and top of the fix and it's all good to me.

alan-j-hu:
@rgrinberg Did you see the review comments I left? I'm pretty sure that the exception was forgotten to be reraised.

rgrinberg:
Nope. I didn't see them. Maybe you forgot to submit the comments? Will fix though.

alan-j-hu:
[Holy moly, "Pending" is such misleading UI. If not for today, I would not have realized that just commenting on a piece of code in review doesn't submit it.](https://github.com/orgs/community/discussions/10369)

alan-j-hu:
I can confirm that my installation issues are fixed when I pinned the dune package to the Git repo.

PR #8289, opened 2023-07-28
Add flag to `dune fmt` to disable promoting

gridbugs:
Sometimes you just want to see the diff without applying it.

### DIFF
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -209,8 +209,25 @@ let fmt =
     ]
   in
   let term =
-    let+ common = Common.term in
-    let common = Common.set_promote common Dune_engine.Clflags.Promote.Automatically in
+    let+ common = Common.term
+    and+ no_promote =
+      Arg.(
+        value
+        & flag
+        & info
+            [ "preview" ]
+            ~doc:
+              "Just print the changes that would be made without actually applying them. \
+               This takes precedence over auto-promote as that flag is assumed for this \
+               command.")
+    in
+    let common =
+      Common.set_promote
+        common
+        (if no_promote
+         then Dune_engine.Clflags.Promote.Never
+         else Dune_engine.Clflags.Promote.Automatically)
+    in
     let config = Common.init common in
     let request (setup : Import.Main.build_system) =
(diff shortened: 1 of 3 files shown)


Alizter:
Why not just do `dune build @fmt`?

bin/build_cmd.ml
```diff
@@ -202,9 +202,20 @@ let fmt =
     ]
   in
   let term =
-    let+ common = Common.term in
+    let+ common = Common.term
+    and+ no_promote =
+      Arg.(
+        value & flag
+        & info [ "no-promote" ]
```
rgrinberg, line 209: Maybe the flag should be called `preview`?

rgrinberg:
Now that we've added this subcommand, might as well make it complete.

gridbugs:
I did not know about `dune build @fmt`. I imagine many users will follow the same train of thought that led me to write this: noticing that `dune fmt` exists, then noticing that it has side effects, then checking the docs for how to make those side effects not happen.

Alizter:
By the way, there is a `--disable-promotion` flag common to even `dune fmt` that does nothing. I wonder if we should just use that.

gridbugs:
> The problem is that is means something else in `dune build` which is to avoid those auto-promoting rules.

What does it mean to promote a rule?

bin/build_cmd.ml
gridbugs: yep, changed

Alizter:
@gridbugs I mean that some rules can promote their targets automatically when you build them.

gridbugs:
Makes sense. I think let's name this flag `preview` and keep it separate from `disable-promotion` to avoid confusion (even though in a sense they ideally do the same thing).

rgrinberg:
needs a CHANGES entry

gridbugs:
> needs a CHANGES entry

Added!

rgrinberg:
Could you move the entry to doc/changes and rebase + reformat?

Sorry for the back and forth.

gridbugs:
> Could you move the entry to doc/changes and rebase + reformat?

Done!

PR #8293, opened 2023-07-28
Delete ctypes extensions 0.1 and 0.2

emillon:


Alizter (approved):

rgrinberg (approved):

Leonidas-from-XIV (approved):
Looks good.

The CI failure is on a test that shouldn't be affected and only happens on macOS, so I assume there is something untelated failing (and macOS CI is failing on other PRs as well, though with different issues).

Alizter:
That failure occasionally happens on the macos CI runners. No idea why.

PR #8294, opened 2023-07-28
feat: add a way to mark extensions as deleted

emillon:
Our extension mechanism is versioned, but we only support adding new versions.
However, for experimental extensions or 0.x versions we've always said that
these could be removed. This PR adds such a mechanism.

### DIFF
--- a/src/dune_sexp/syntax.ml
+++ b/src/dune_sexp/syntax.ml
@@ -16,6 +16,7 @@ module Version = struct
 
   include T
   module Infix = Comparator.Operators (T)
+  module Map = Map.Make (T)
 
   let equal = Infix.equal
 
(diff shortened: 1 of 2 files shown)


emillon:
See #8293 to see this in action.

src/dune_sexp/syntax.ml
```diff
@@ -270,22 +333,27 @@ let create ?(experimental = false) ~name ~desc supported_versions =
...
 let check_supported ~dune_lang_ver t (loc, ver) =
-  if
-    not (Supported_versions.is_supported t.supported_versions ver dune_lang_ver)
-  then
+  match Supported_versions.status t.supported_versions ver ~dune_lang_ver with
+  | `Supported -> ()
+  | `Deleted_in deleted_in ->
+    User_error.raise ~loc
+      [ Pp.textf
+          "Version %s of the %s extension has been deleted in Dune %s. Please \
```
rgrinberg, line 341: might be useful to mention what's the newest (or oldest) version available in this mesage

Alizter:
cc @ejgallego

src/dune_sexp/syntax.ml
Alizter: You can directly add a hint with the current dune lang version.

emillon: Done!

rgrinberg (approved):

PR #8298, opened 2023-07-29
refactor: arrays instead of lists for arguments

rgrinberg:
Should cost us about half the storage

### DIFF
--- a/src/dune_engine/action.ml
+++ b/src/dune_engine/action.ml
@@ -21,7 +21,7 @@ module Make
 struct
   include Ast
 
-  let run prog args = Run (prog, args)
+  let run prog args = Run (prog, Array.Immutable.of_list args)
 
   let chdir path t = Chdir (path, t)
 
--- a/bin/print_rules.ml
+++ b/bin/print_rules.ml
@@ -55,7 +55,10 @@ let rec encode : Action.For_shell.t -> Dune_lang.t =
   let path = Encoder.string in
   let target = Encoder.string in
   function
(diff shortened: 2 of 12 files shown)


rgrinberg:
Well, this indeed lowers memory usage essentially for free:

* Live words go down about 1.8%
* 5 minor collections less for null builds
* Major words are 1.3% less

snowleopard:
One problem with using plain arrays in APIs is their mutability. How about creating a little module in Stdune for immutable arrays (backed by plain arrays but not exposing unsafe operations)?

rgrinberg:
Good point. I added a `Array.Immutable` for this.

src/dune_engine/action.ml
```diff
@@ -21,7 +21,7 @@ module Make
 struct
   include Ast
 
-  let run prog args = Run (prog, args)
+  let run prog args = Run (prog, Array.Immutable.of_list args)
```
snowleopard, line 24: Should this change to take an array as the argument, to encourage the use of arrays in the rules?

snowleopard (approved):
Looks great, thanks! Could you import this PR, to benchmark it internally too?

src/dune_engine/action.ml
snowleopard: (Maybe this should be done in a separate PR.)

rgrinberg: I tried this, but it's a little problematic because:

* It's not clear if there's any benefits as there's so many layers of conversion for all these flags already
* Quite a bit of code will have to change.

Will leave a TODO for this.

PR #8300, opened 2023-07-30
refactor: use unboxed optionals in memo

rgrinberg:
This saves us 2 words in storage and one less pointer indirection

### DIFF
--- a/otherlibs/stdune/src/option.mli
+++ b/otherlibs/stdune/src/option.mli
@@ -60,3 +60,22 @@ module List : sig
 end
 
 val merge : 'a t -> 'a t -> f:('a -> 'a -> 'a) -> 'a t
+
+module Unboxed : sig
+  (** Poor man's unboxed option types. The value stored must not be immediate. A
+      consequence of that is that such option types cannot be nested *)
+
+  type 'a t
+
+  val get_exn : 'a t -> 'a
+
+  (** [some a] will construct the present value. If [a] is immediate, this
+      function will raise *)
+  val some : 'a -> 'a t
+
+  val none : 'a t
+
+  val is_none : 'a t -> bool
+
+  val to_dyn : ('a -> Dyn.t) -> 'a t -> Dyn.t
+end
--- a/src/memo/memo.ml
+++ b/src/memo/memo.ml
@@ -314,7 +314,13 @@ module Cache_lookup = struct
      outcome would have been [Out_of_date {old_value = None}] instead. *)
   module Failure = struct
     type 'a t =
-      | Out_of_date of { old_value : 'a option }
+      | Out_of_date of
+          { (* ['a] is instantiated to ['a Cached_value.t] below but making
+               this explicit would require us to move this type and [Result]
+               into the recursive module of [M]. We leave this comment to
+               explain why using [Option.Unboxed.t] is OK here. *)
+            old_value : 'a Option.Unboxed.t
+          }
       | Cancelled of { dependency_cycle : Cycle_error.t }
   end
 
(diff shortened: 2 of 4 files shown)


rgrinberg:
This is enough to shave off a few minor collections and reduce 1% from the major words of a clean build.

Probably this type can be used in quite a few more places, but I haven't bothered looking at everything for now.

src/memo/memo.ml
```diff
@@ -314,7 +314,7 @@ module Cache_lookup = struct
      outcome would have been [Out_of_date {old_value = None}] instead. *)
   module Failure = struct
     type 'a t =
-      | Out_of_date of { old_value : 'a option }
+      | Out_of_date of { old_value : 'a Option.Unboxed.t }
```
snowleopard, line 317: Isn't it unsafe to use `Option.Unboxed` here because `'a` can be anything?

otherlibs/stdune/src/option.mli
```diff
@@ -60,3 +60,19 @@ module List : sig
 end
 
 val merge : 'a t -> 'a t -> f:('a -> 'a -> 'a) -> 'a t
+
+module Unboxed : sig
+  (** Poor man's unboxed option types *)
```
snowleopard, line 65: Surely, you can't use this type for any `'a`? Can you document the conditions under which it's safe?

snowleopard: (Like, I'm guessing `'a` can't be an immediate?)

src/memo/memo.ml
rgrinberg: "Unsafe" in this context would mean that you just get a runtime error. You will not get a segfault. In this particular case, it's alright because the old_value is always `Cached_value.t`. I could move the definition to the recursive module `M` to specialize this type, but I'm not quite a fan of that either. Do you have a preference?

otherlibs/stdune/src/option.mli
rgrinberg: I've documented this better and improved the errors.

otherlibs/stdune/src/option.mli
```diff
@@ -60,3 +60,21 @@ module List : sig
 end
 
 val merge : 'a t -> 'a t -> f:('a -> 'a -> 'a) -> 'a t
+
+module Unboxed : sig
+  (** Poor man's unboxed option types. The value stored must not be immediate *)
```
snowleopard, line 65: It may not be obvious from this comment that you can't use `'a t t`, i.e. you can't nest such cheap unboxed options.

src/memo/memo.ml
snowleopard: Got it. I think we need to add a comment explaining why it's safe to use `Option.Unboxed.t` here.

otherlibs/stdune/src/option.mli
rgrinberg: Add a comment regarding this use case.

src/memo/memo.ml
rgrinberg: Done.

src/memo/memo.ml
```diff
@@ -314,7 +314,14 @@ module Cache_lookup = struct
      outcome would have been [Out_of_date {old_value = None}] instead. *)
   module Failure = struct
     type 'a t =
-      | Out_of_date of { old_value : 'a option }
+      | Out_of_date of
+          { (* Morally, [old_value] is always:
+                ['a Cached_value.t Option.Unboxed.t]
+               but requiring this would require us to move this type and
+               [Result] into the recursive module of [M]. We leave this
+               comment to justify why using [Option.Unboxed.t] is OK here. *)
```
snowleopard, line 322:
```suggestion
      | Out_of_date of
          { (* ['a] is instantiated to ['a Cached_value.t] below but making
               this explicit would require us to move this type and [Result]
               into the recursive module of [M]. We leave this comment to
               explain why using [Option.Unboxed.t] is OK here. *)
```

snowleopard (approved):
Thanks! I made a little tweak to one of the comments. Before merging, please import this change so that we can confirm improvements on our internal benchmarks.

snowleopard:
@rgrinberg Can you import the version we ended up releasing internally (minus `local_`?).

snowleopard:
For the record, internally, we're getting consistent small (~0.5) improvements in allocations on minor and major heaps. Build times show a small (0.5-1.0%) regression in build times on some smaller benchmarks like Core but that's likely noise.

snowleopard:
> @rgrinberg Can you import the version we ended up releasing internally (minus `local_`?).

Pinging on this. I think we should unify the code with our JS version.

PR #8304, opened 2023-07-31
refactor: replace deprecated runloop in fsevents

### LINKED ISSUE #7352: Not a bug: some functions are deprecated on MacOS >=13 (Ventura)
Just to inform you:

`FSEventStreamScheduleWithRunLoop` and `vfork` are both deprecated on OS X.

```
fsevents_stubs.c:203:3: warning: 'FSEventStreamScheduleWithRunLoop' is deprecated: first deprecated in macOS 13.0 - Use FSEventStreamSetDispatchQueue instead. [-Wdeprecated-declarations]
  FSEventStreamScheduleWithRunLoop(t->stream, runloop->runloop,
  ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/CoreServices.framework/Frameworks/FSEvents.framework/Headers/FSEvents.h:1138:1: note: 'FSEventStreamScheduleWithRunLoop' has been explicitly marked deprecated here
FSEventStreamScheduleWithRunLoop(
^
1 warning generated.
spawn_stubs.c:439:33: warning: 'vfork' is deprecated: Use posix_spawn or fork [-Wdeprecated-declarations]
  ret = Bool_val(v_use_vfork) ? vfork() : fork();
                                ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/unistd.h:604:1: note: 'vfork' has been explicitly marked deprecated here
__deprecated_msg("Use posix_spawn or fork")
^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/cdefs.h:215:48: note: expanded from macro '__deprecated_msg'
        #define __deprecated_msg(_msg) __attribute__((__deprecated__(_msg)))
                                                      ^
1 warning generated.
```

kevinji:
Replace the use of the deprecated `FSEventStreamScheduleWithRunLoop()` with `FSEventStreamSetDispatchQueue()`.

Since a dispatch queue spawns new threads, we need to call `caml_c_thread_register()` and `caml_c_thread_unregister()` for these threads. This is done indirectly via a `pthread_key_t` so that the functions only need to be called more than once if the dispatch queue switches which thread is running the callback.

We replicate the blocking nature of the existing code using a mutex and condition variable.

See git/git@b0226007f0aaf448dec1defe3e44c4e3d7513aa8 for more details about this approach.

Fixes #7352.

### DIFF
--- a/src/fsevents/fsevents_stubs.c
+++ b/src/fsevents/fsevents_stubs.c
@@ -14,30 +14,45 @@
 
 #include <CoreFoundation/CoreFoundation.h>
 #include <CoreServices/CoreServices.h>
+#include <pthread.h>
 
-typedef struct dune_runloop {
-  CFRunLoopRef runloop;
+// We use a condvar/mutex pair to allow the code calling
+// `dune_fsevents_dispatch_queue_wait_until_stopped` to be signaled by either
+//
+// - An exception raised by the callback passed to the dispatch queue
+// - Explicitly stopping via `dune_fsevents_stop`
+//
+// Since a dispatch queue uses an internal pool of threads, we need some
+// synchronization around updating `v_exn`.
+typedef struct dune_dispatch_queue {
+  dispatch_queue_t dq;
+  pthread_cond_t dq_finished;
+  pthread_mutex_t dq_lock;
   value v_exn;
-} dune_runloop;
+} dune_dispatch_queue;
 
 typedef struct dune_fsevents_t {
-  dune_runloop *runloop;
+  dune_dispatch_queue *dq;
   value v_callback;
   FSEventStreamRef stream;
 } dune_fsevents_t;
 
-#define Runloop_val(v) (*((dune_runloop **)Data_custom_val(v)))
+#define Dispatch_queue_val(v) (*((dune_dispatch_queue **)Data_custom_val(v)))
 
-void dune_fsevents_runloop_finalize(value v_runloop) {
-  dune_runloop *runloop = Runloop_val(v_runloop);
-  caml_stat_free(runloop);
+void dune_fsevents_dispatch_queue_finalize(value v_dq) {
+  dune_dispatch_queue *dq = Dispatch_queue_val(v_dq);
+  if (dq->dq)
+    dispatch_release(dq->dq);
+  pthread_cond_destroy(&dq->dq_finished);
+  pthread_mutex_destroy(&dq->dq_lock);
+  caml_stat_free(dq);
 }
 
-static struct custom_operations dune_fsevents_runloop_ops = {
-    "dune.fsevents.runloop",    dune_fsevents_runloop_finalize,
-    custom_compare_default,     custom_hash_default,
-    custom_serialize_default,   custom_deserialize_default,
-    custom_compare_ext_default, custom_fixed_length_default};
+static struct custom_operations dune_fsevents_dispatch_queue_ops = {
+    "build.dune.fsevents.dispatch_queue", dune_fsevents_dispatch_queue_finalize,
+    custom_compare_default,               custom_hash_default,
+    custom_serialize_default,             custom_deserialize_default,
+    custom_compare_ext_default,           custom_fixed_length_default};
 
 #define Fsevents_val(v) (*(dune_fsevents_t **)Data_custom_val(v))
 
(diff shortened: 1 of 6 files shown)


kevinji:
This is my first time interfacing with C from OCaml and I'm getting the following error in CI:

```
+  ../watching/helpers.sh: line 3: 85234 Illegal instruction: 4  ( dune build "$@" --passive-watch-mode > .#dune-output 2>&1 )
```

Some guidance around how I can reproduce this error locally and how I could go about debugging this would be helpful!

Alizter:
@kevinji That is a very strange error. What is your OS / arch? What kind of file system are you using?

kevinji:
@Alizter This is the macOS CI build error. I’m having trouble replicating it on a local M1 machine so I’m also asking for some help setting up my dev environment—I’ve run `make bootstrap`, `make dev`, and then `make test` but I’m getting other errors that aren’t present in CI.

I’ve modified the C bindings to fsevents so I’m wondering if I need to pass a flag for pthread support, or if I messed something up when doing FFI between C and OCaml.

Alizter:
I wonder if the C optimizer is doing some funky things. What C toolchain versions are you using?

kevinji:
On my machine both `gcc` and `clang` refer to Apple's `clang`:
```
~
❯ gcc --version
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: arm64-apple-darwin22.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

~
❯ clang --version
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: arm64-apple-darwin22.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

kevinji:
For reference, on my local machine, I get warnings that look like
```
File "test/blackbox-tests/test-cases/virtual-libraries/incorrect-archive-7027.t", line 1, characters 0-0:
diff --git a/_build/.sandbox/8a8c802dbf0064c575d0050dcd663d21/default/test/blackbox-tests/test-cases/virtual-libraries/incorrect-archive-7027.t b/_build/.sandbox/8a8c802dbf0064c575d0050dcd663d21/default/test/blackbox-tests/test-cases/virtual-libraries/incorrect-archive-7027.t.corrected
index bfdebcb23..fcc113691 100644
--- a/_build/.sandbox/8a8c802dbf0064c575d0050dcd663d21/default/test/blackbox-tests/test-cases/virtual-libraries/incorrect-archive-7027.t
+++ b/_build/.sandbox/8a8c802dbf0064c575d0050dcd663d21/default/test/blackbox-tests/test-cases/virtual-libraries/incorrect-archive-7027.t.corrected
@@ -48,3 +48,10 @@ https://github.com/ocaml/dune/issues/7027
   > EOF

   $ dune exec ./foo.exe
+  /var/folders/5d/8_6n7qx13nnbr5nmmr096gqw0000gn/T/build_1858d6_dune/build_874b91_dune/camlobj3924a9.c:1525:14: warning: a function declaration without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
+  extern value caml_get_public_method();
+               ^
+  /var/folders/5d/8_6n7qx13nnbr5nmmr096gqw0000gn/T/build_1858d6_dune/build_874b91_dune/camlobj3924a9.c:1727:14: warning: a function declaration without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
+  extern value caml_set_oo_id();
+               ^
+  2 warnings generated.
```

and failed expect tests like
```
File "test/expect-tests/persistent_tests.ml", line 1, characters 0-0:
diff --git a/_build/default/test/expect-tests/persistent_tests.ml b/_build/.sandbox/11d2f2745866f56ea24aef378f7a7bc4/default/test/expect-tests/persistent_tests.ml.corrected
index c87263c42..3042d3c43 100644
--- a/_build/default/test/expect-tests/persistent_tests.ml
+++ b/_build/.sandbox/11d2f2745866f56ea24aef378f7a7bc4/default/test/expect-tests/persistent_tests.ml.corrected
@@ -28,7 +28,7 @@ let%expect_test "persistent digests" =
     ---

     DIGEST-DB version 6
-    a4ae8e07cf52a9fb38c47c32b6d59fa6
+    a6df9e528c50debc9264b7a95489392e
     ---

     INSTALL-COOKIE version 1
@@ -40,7 +40,7 @@ let%expect_test "persistent digests" =
     ---

     COPY-LINE-DIRECTIVE-MAP version 1
-    7e311b06ebde9ff1708e4c3a1d3f5633
+    7dac5b11f6f654bb6f230392493b363f
     ---

     merlin-conf version 4
@@ -48,5 +48,5 @@ let%expect_test "persistent digests" =
     ---

     INCREMENTAL-DB version 5
-    fa67cc9b60c9f3a1b9b1ad93a56df691
+    1cc656a4502ef88e70adab1f3c9a868e
     --- |}]
```

Alizter:
Just to confirm, the test is OK before this PR?

anmonteiro:
I ran `./dune.exe build @test/blackbox-tests/test-cases/watching/path-pwd` and that only fails with this PR

anmonteiro:
Actually it doesn't repro anymore after the latest force-push.

kevinji:
I pushed a new version that fixes the C logic. The original code freed the dispatch queue fields in the wrong place (`dune_fsevents_stop`); the updated commit pushes the changes to `dune_fsevents_dispatch_queue_run` instead. I also removed some comments in `fsevents.mli` and some unused functions in `fsevents_stubs.c` to reflect what the current code actually does.

emillon:
This area had some issues with memory safety before (#6151). We were not convinced of the soundness of the bindings so it's not completely unexpected that touching this will reveal issues.

kevinji:
That's helpful to know. I pushed a new commit with the following small changes:

- Dispatch queue cleanup (including for the mutex/condvar) is now moved to the `finalize` function.
- The `custom_operations` types now are prefixed with `build.` to be Java-style per [the OCaml docs](https://v2.ocaml.org/manual/intfc.html#ss:c-custom-idents) and also to be consistent with the dispatch queue name.

src/fsevents/fsevents_stubs.c
```diff
@@ -14,30 +14,37 @@
 
 #include <CoreFoundation/CoreFoundation.h>
 #include <CoreServices/CoreServices.h>
+#include <pthread.h>
 
-typedef struct dune_runloop {
-  CFRunLoopRef runloop;
+typedef struct dune_dispatch_queue {
+  dispatch_queue_t dq;
+  pthread_cond_t dq_finished;
```
rgrinberg, line 21: Could you add some comments why we need the cv and mutex?

src/fsevents/fsevents_stubs.c
```diff
@@ -48,33 +55,37 @@ static void dune_fsevents_t_finalize(value v_t) {
...
+  dq->dq = dispatch_queue_create("build.dune.fsevents", DISPATCH_QUEUE_SERIAL);
+  dq->v_exn = Val_unit;
+  caml_register_global_root(&dq->v_exn);
+  value v_dq = caml_alloc_custom(&dune_fsevents_dispatch_queue_ops,
+                                 sizeof(dune_dispatch_queue *), 0, 1);
+  Dispatch_queue_val(v_dq) = dq;
+  CAMLreturn(v_dq);
+}
+
+CAMLprim value dune_fsevents_dispatch_queue_run(value v_dq) {
```
gridbugs, line 78: It looks like this function now just blocks via a condition variable until an exception is raised by the event handler? Previously it had the effect of running the runloop but now it looks like all it does is wait until an exception and then clean up. Perhaps this function's name should reflect the fact that it no longer "runs" anything.

kevinji: I updated the function name to `wait_until_stopped`.

kevinji:
New updates:

- The function name `dune_fsevents_dispatch_queue_run` is now `dune_fsevents_dispatch_queue_wait_until_stopped` to better reflect that it no longer "runs" anything directly.
- `pthread_cond_broadcast(&t->dq->dq_finished);` is now also called in `dune_fsevents_stop` so `dune_fsevents_dispatch_queue_wait_until_stopped` can be stopped gracefully.

rgrinberg (approved):

Alizter:
When does the dispatch queue spawn new threads?

kevinji:
According to [the macOS documentation](https://developer.apple.com/documentation/dispatch/dispatch_queue): "Work submitted to dispatch queues executes on a pool of threads managed by the system." I'm not sure at which specific point the system actually creates the threads though.

src/fsevents/fsevents_stubs.c
```diff
@@ -137,11 +157,14 @@ static void dune_fsevents_callback(const FSEventStreamRef streamRef,
...
-    t->runloop->v_exn = Extract_exception(v_res);
-    CFRunLoopStop(t->runloop->runloop);
+    pthread_mutex_lock(&t->dq->dq_lock);
+    t->dq->v_exn = Extract_exception(v_res);
+    pthread_cond_broadcast(&t->dq->dq_finished);
+    pthread_mutex_unlock(&t->dq->dq_lock);
   }
   CAMLdrop;
   caml_release_runtime_system();
+  caml_c_thread_unregister();
```
rgrinberg, line 167: We are calling unregister in every single callback invocation. Are you sure this call is cheap enough for that?

kevinji: Is there a good metric for what you mean here? From what I can tell `caml_c_thread_register` creates a thread info block and then attaches it to an existing linked list of thread info blocks, and `caml_c_thread_unregister` reverses those changes. However, I'm not sure how expensive that is.

rgrinberg: I'm not so concerned about the linked list nor the block, but rather the lock that has to be acquired to do this. Given that this callback in our case isn't doing much work (just notifying which memoization nodes are out of date and hence invalidating the build can easily add up) acquiring this lock can easily affect the latency in watch mode.

I'd be more eager if this PR added some feature in exchange for the worse performance. Are there are any concrete benefits to using dispatch queues that I'm unaware of perhaps?

kevinji: The original reason is that [`FSEventStreamScheduleWithRunLoop` is deprecated](https://developer.apple.com/documentation/coreservices/1447824-fseventstreamschedulewithrunloop) with a suggestion to use [`FSEventStreamSetDispatchQueue`](https://developer.apple.com/documentation/coreservices/1444164-fseventstreamsetdispatchqueue) instead. I think in practice, since we're using a serial queue, the background thread running the callback should remain the same, but I couldn't find an easy way to manually manage the threads used for the queue, which could allow us to only run `caml_c_thread_register` as needed.

Alternatively, we could use [`dispatch_get_main_queue`](https://developer.apple.com/documentation/dispatch/1452921-dispatch_get_main_queue), which uses a serial dispatch queue in the main thread. This would have roughly the same behavior as the original code that use [`CFRunLoopGetCurrent`](https://developer.apple.com/documentation/corefoundation/1542428-cfrunloopgetcurrent) assuming there was originally only one thread, but I think we would probably need to run the main dispatch queue somehow with either [`dispatch_main`](https://developer.apple.com/documentation/dispatch/1452860-dispatch_main) or [`CFRunLoopRun`](https://developer.apple.com/documentation/corefoundation/1542011-cfrunlooprun).

kevinji: As an alternative, I've introduced some thread-local storage that keeps track of whether `caml_c_thread_register` has been called already by the same thread, and only unregisters once the thread exits. Since a serial dispatch queue often uses the same thread, this should in principle reduce the number of register/unregister calls needed, but additional testing is probably needed.

rgrinberg: Yeah, that's a bit better. If you could do some testing measuring that we aren't spamming caml_c_thread_register/unregister then this should be good to go.

Another alternative that's a little more bulletproof but requires a bit more code is to have the code in the dispatch queue populate some sort of queue with events, and require the OCaml side to poll this queue for events in a loop. This will require some synchronization and the use of CV's to avoid busy polling, but it's probably the best we can do here.

cc @patricoferris who discussed this issue with me.

kevinji: @rgrinberg Sorry just got some time to come back to look at this. I can confirm that running the `test/blackbox-tests/test-cases/watching` test that the same background thread is being reused and `caml_c_thread_register` is only being run once. Are there some longer workflows that I can test to make sure the behavior is also as expected?

rgrinberg: You could try running longer builds on more serious projects. But your test seems enough to me.

rgrinberg: If you have some spare cycles, I would suggest looking at the queue based workaround I suggest previous comments. I think that should guarantee this is well behaved.

Issue #8238, opened 2023-07-20
dune rules --makefile does not produce a working Makefile


samoht:
The `dune rules --makefile` is a nice idea but currently doesn't seem to work.

For instance, this simple project:

```
(executable (name main))
```

```ocaml
(* main.ml *)
let () = Printf.printf "hello world\n%!"
```

will produce an error.

```
$ dune rules --makefile > Makefile
$ make
make: *** No rule to make target `_build/default/.merlin-conf/exe-main', needed by `_build/default/.merlin-exist-exe-main'.  Stop.
```

Alizter:
Actually upon further inspection, it seems you are missing the `-r` option to `dune rules --makefile`. The command `dune rules` will print all the rules for the "top level" targets. The `--makefile` option is really just a formatting option for printing these rules (versus sexp). In order for the rule set to be complete you need to print the `-r` option as well.

Alizter:
On the other hand, I think it should be the default for `-r` to be implied by `--makefile` otherwise there is really no point. I'll open a PR.

rgrinberg:
> The dune rules --makefile is a nice idea but currently doesn't seem to work.

it never worked for non trivial projects anyway.

The only use for `dune rules` is debugging, and the makefile output is not better than our sexp output. So I say let's just get rid of this flag and prevent our users from being confused by this.

jchavarri:
As a workaround, one can do (e.g. in OP example): `dune rule -m _build/default/.merlin-conf/exe-main` and get the rule for that target in Makefile format manually.

> The only use for `dune rules` is debugging

For debugging purposes it's desirable to include as much information as possible of what's happening in reality, which the Makefile format conveniently helps exercise. I understand there's some hard limitation that prevents including this kind of "implicit" or internal rules in `dune rules` output? Another example:

```
  $ cat > dune-project <<EOF
  > (lang dune 3.17)
  > EOF

  $ cat > dune <<EOF
  > (library
  >  (name lib))
  > EOF

  $ dune rules -m > Makefile

  $ make _build/default/lib.cma
  make: *** No rule to make target '_build/default/.lib.objs/byte/lib.cmo', needed by '_build/default/lib.cma'.  Stop.
  [2]
```

rgrinberg:
For debugging, how about we stick to the sexp format? We don't have to pretend the output is executable, and we can make the output adjustable so that it works for `-r`.

jchavarri:
[...]

PR #8312, opened 2023-07-31
feature: dune rules --makefile always implies -r

Alizter:
When a user passes the `--makefile` option, they always mean to also pass `-r` since the makefile will not work without recursive targets.

We add a test to test this use case and make `--makefile` always imply `-r` to demonstrate that it works.

- [x] tests
- [ ] changelog
- [x] fix tests
- fix https://github.com/ocaml/dune/issues/8238


Alizter:
We need to change the shell output to be more cross platform. `echo -n` is not valid posix. Can easily be replaced with `printf` however.

PR #8314, opened 2023-08-01
Tidy up solver env

gridbugs:
This makes it possible to configure the `make` and `jobs` opam variables through a workspace file and also to derive the `make` variable (the path to the `make` executable) from `$PATH` using the same machinery as the `%{exe:make}` pform macro. As the semantics of setting and reading build config variables differs from other types of opam variable I've taken this opportunity to refactor the solver env to reflect the different types of opam variable. This unifies the way variables are accessed which allows simplifications to the solver, also included in this change.

Also since the size of the solver environment is growing large, this change modifies the way that the environment is printed when the `--just-print-solver-env` flag is passed. The output is now more verbose and formats the variables as a nested bulleted list rather than a sexp.

### DIFF
--- a/src/dune_pkg/solver_env.mli
+++ b/src/dune_pkg/solver_env.mli
@@ -1,74 +1,89 @@
 open! Stdune
 
+module Variable : sig
+  module Flag : sig
+    (** A boolean variable *)
+    type t =
+      [ `With_test
+      | `With_doc
+      ]
+  end
+
+  module Sys : sig
+    (** Variables describing a system environment. These can be polled from the
+        current system and assigned specific values in dune-workspace but there
+        is no notion of a default value. During solving unset variables are
+        treated as wildcards so a solution can be built that works on a range of
+        systems. During building unset variables are treated as undefined
+        variables. *)
+    type t =
+      [ `Arch
+      | `Os
+      | `Os_version
+      | `Os_distribution
+      | `Os_family
+      ]
+
+    val to_string : t -> string
+    val of_string_opt : string -> t option
+
+    module Bindings : sig
+      type sys_var := t
+
+      (** A mapping from system environment variables to their values *)
+      type t
+
+      val empty : t
+      val set : t -> sys_var -> string -> t
+      val get : t -> sys_var -> string option
+
+      type union_error =
+        [ `Var_in_both_with_different_values of sys_var * string * string ]
+
+      (** Merge two environments returning an error if they both contain a
+          binding of the same variable to different values. *)
+      val union : t -> t -> (t, union_error) result
+    end
+  end
+
+  module Const : sig
+    (** Constant variables whose value can't be configured. These can be read
+        while solving and evaluating packages. *)
+    type t = [ `Opam_version ]
+  end
+
+  type t =
+    | Flag of Flag.t
+    | Sys of Sys.t
+    | Const of Const.t
+
+  val of_string_opt : string -> t option
+end
+
 (** A typed variable environment used by the dependency solver to evaluate
     package filters. Opam packages can declare conditional dependencies on other
     packages using a language made up of boolean operators and comparisons of
     strings. Variables in this language can represent booleans and strings. *)
+type t
 
-module Flag : sig
-  (** A boolean variable *)
-  type t =
-    [ `With_test
-    | `With_doc
-    ]
-
-  val to_string : t -> string
-  val of_string_opt : string -> t option
-
-  module Set : sig
-    type flag := t
-
-    (** A set flags. The presence of a flag in the set indicates that the flag
-        is set to true. *)
-    type t
-
-    val fold : t -> init:'a -> f:(flag -> 'a -> 'a) -> 'a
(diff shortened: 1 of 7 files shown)


gridbugs:
I still intend to update tests to ensure we can parse the new variables from the workspace config as thus far I've relied on deriving them from the current system. Opening this PR now for visibility and early feedback as it's quite a large change.

src/dune_pkg/solver_env.mli
```diff
@@ -1,84 +1,99 @@
...
+  type t =
+    | Bool of bool
+    | String of string
+    | Unset_sys_env
+end
 
-val default : t
+(** Look up a variable's value by name *)
+val get_by_variable_name :
+  t -> string -> (Variable_value.t, [ `Unknown_variable ]) result
```
rgrinberg, line 99: Could you introduce a type for variable names?

src/dune_pkg/solver_env.mli
```diff
@@ -1,84 +1,99 @@
...
+(** Create a [t] from a [With_opt_build_config.t] filling in any unset fields
+    from the build config with default values. The [which_make] function is used
+    to look up the path of the `make` executable if the "make" build config
+    variable is unset. The [which_make] argument is ideally a function which
+    calls [Dune_rules.Expander.which_make] for the appropriate context. That
+    function can't be called directly from this module as it would create a
+    circular dependency. *)
+val with_default_values_for_unset_build_configs :
+     With_opt_build_config.t
+  -> which_make:(unit -> Path.t option Memo.t)
```
rgrinberg, line 77: Why does this argument need to be delayed?

gridbugs: We only need to call it if there is no `make` variable configured.

src/dune_pkg/solver_env.mli
gridbugs: There is `Variable.t` inside solver_env.ml but it's not exposed. Think it would be clearer to use expose it and get rid of this helper function?

gridbugs:
Updated the `just-print-solver-env` test to set the `make` variable in its workspace configs. This turns out to be necessary for the test as otherwise it will print the path to `make` on the system running the test which will differ between different systems.

gridbugs:
There are some tests failing with:
```
[ERROR] The selected switch default is not installed.
'/opt/hostedtoolcache/opam/2.1.5/x86_64/opam config env --switch default --sexp --set-switch' failed.
[1]
```
This seems to be because we have to run that opam command to determine the location of `make` since it could be inside opam's `bin` dir. I'll debug this further tomorrow.

src/dune_pkg/solver_env.mli
gridbugs: We would still need to expose a way to convert strings (or I guess `OpamVariable.t`s) into the `Variable.t` type though. What's the purpose of avoiding raw strings? Is it just because the meaning of the string is not suggested by its type? Would replacing `string` in this api with `OpamVariable.t` be sufficient?

rgrinberg: There's two problem with using raw strings:

1. It makes reading the code harder than it needs be.
2. It becomes possible to skip the validation that is often necessary for these various types. For example, I'm pretty sure that an opam variable cannot contain the `'\0'` character. In some cases where we didn't add such validation, we ended up having to introduce painful workaround down the roads (see the code library names and module names).

src/dune_pkg/solver_env.mli
rgrinberg: Should we make `%{make}` part of the solver environment at all? Selecting a package based on the concrete path of make sounds like a very bad idea to me. Should we just forbid it?

src/dune_pkg/solver_env.mli
Leonidas-from-XIV: I would suggest against exposing `OpamVariable.t` in our API, as this ties us a bit stronger to OPAM and if we decide to change it later it would make things harder.

src/dune_pkg/solver_env.mli
gridbugs: Good point. I added them so that we could generate lockfiles with the build and install commands fully evaluated as a first step before supporting filters and substitutions in build and install commands but given the complexity added by these variables and the fact that packages _shouldn't_ be filtering based on `%{make}` or `%{jobs}` I'm inclined to remove them. Instead I'll generate lockfiles with unfiltered commands and convert variables like `%{make}` and `%{jobs}` into the appropriate pforms.

gridbugs:
Based on the fact that we really don't want packages to be able to filter based on the path to `make` or the `jobs` value I've removed the build config variables from this PR and changed the name accordingly. I've kept the solver env refactors as they simplify the solver and lock command implementation.

rgrinberg (approved):

Leonidas-from-XIV (approved):

rgrinberg:
@gridbugs please rebase (and reformat)

gridbugs:
> @gridbugs please rebase (and reformat)

Done!

PR #8319, opened 2023-08-02
chore(ocamlformat): use the janestreet profile

emillon:


rgrinberg:
cc @snowleopard. I looked over the diff and it does look pretty close to what we have internally.

snowleopard (approved):
Many thanks for doing this! That's going to make our lives so much easier when importing/exporting changes.

rgrinberg:
@julow informed us that this fork of ocamlformat is closer to what jst is using: https://github.com/ocaml-ppx/ocamlformat/pull/2314

@emillon could you give this a try and see if it gives us a diff against this? I don't mind using the fork temporarily.

otherlibs/chrome-trace/src/chrome_trace.ml
```diff
@@ -16,20 +16,18 @@ module Timestamp : sig
...
   type t = float
 
   let of_float_seconds x = x
-
   let to_float_seconds x = x
 
   let to_json f =
     let n = int_of_float @@ (f *. 1_000_000.) in
     `Int n
+  ;;
```
Alizter, line 30: What are these about?

Leonidas-from-XIV:
@Julow also mentioned that the plan is to upstream these in 0.27, so maybe we can use 0.26 in the meantime and update later? I don't know how painful this is for our existing PRs.

rgrinberg:
This PR will unblock me in syncing with upstream. I would lean against tying that work to ocamlformat's release schedule.

Alizter:
Why does jst have to always use janestreet for formatting? Can an exception for dune not be made?

rgrinberg:
We should find ways to get important work done that minimize the use of the limited resources that we have. This PR is a step towards that. Existing policies and their exceptions are of a secondary concern to us.

Alizter (approved):
Very well, LGTM as long as we add this to the ignoredrevs list too. (after this PR I guess).

emillon:
> @emillon could you give this a try and see if it gives us a diff against this? I don't mind using the fork temporarily.

I added a commit that uses ocaml-ppx/ocamlformat#2314.

> This PR is a step towards that. Existing policies and their exceptions are of a secondary concern to us.

For the part that's about using the janestreet profile, I agree, that's fine. But requiring a fork for development, I don't think that's a good tradeoff to make because this will make contributing harder for everyone.

In terms of diff distance, switching to JS changes about 37kloc. Switching to the fork adds relatively low 4kloc to that. So, just using the profile from the released version makes conflicts 10x less likely for you and changes nothing for contributors. This seems like a good tradeoff. Can we start with this?

rgrinberg:
Sure, let's give it a try. I'll try pulling and see if the conflict resolution is acceptable.

snowleopard:
Thanks everyone!

Issue #8252, opened 2023-07-23
Dune allows the creation of projects starting with numbers and the error message when building doesn't say it can't


hughesjs:
## Expected Behavior
Dune should refuse to create a project with a name that begins with a digit.
If Dune attempts to build a module whose name starts with a digit, the error message should indicate that it can't do so.

## Actual Behavior
It creates the project skeleton.
The error on build reads:
```
Error: "01_module" is an 
invalid module name. Module names must be non-empty and composed only 
of the following characters: 'A'..'Z', 'a'..'z', '_', ''' or '0'..'9'. 
Hint: M01_calorie_counting would be a correct module name
```
Which, while the hint is useful, would actually imply that the name is okay as is.

## Reproduction

```
dune init project 01_module
cd 01 module
dune build
```

## Specifications

- Version of `dune` (output of `dune --version`): 3.9.1
- Version of `ocaml` (output of `ocamlc --version`) 5.0.0
- Operating system (distribution and version): 6.4.4-arch1-1

Alizter:
It's very likely we currently do little to no validation in `dune init`. It shouldn't be a problem to do so.

hughesjs:
So I'm not 100% sure if the same restriction actually applies to the library name too, it might just be the module name, so I don't know if we want that additional validation in `dune init` or whether we should just adjust the error message on build as in #8253. I'll need some feedback from people who have used OCaml for more than a day on that one...

PR #8321, opened 2023-08-02
test(init): add repro for #8252

emillon:


Alizter:
Is it a valid library name? Could there be a case where somebody wants an unwrapped library with this kind of name? This appears to only be an issue for wrapped libraries due to the module naming convention.

emillon:
no, it's not a valid module name so it can't be a valid library name either in the context of dune-init. I'm preparing a fix to get better string names.

Alizter (approved):

PR #8324, opened 2023-08-02
feature: case and cond statements

### LINKED ISSUE #924: Adding some conditionals in the language

Alizter:
We add a `(case ...)` constructor to dune_lang for actions which when expanded will reduce down to the matching branch.

We also add a `(cond ...)` constructor for checking blang values which will reduce down to the first true branch condition.

- [x] doc
- [x] test
- [x] changelog
- fixes #924

### DIFF
--- a/test/blackbox-tests/test-cases/actions/case.t
+++ b/test/blackbox-tests/test-cases/actions/case.t
@@ -1,0 +1,170 @@
+Testing (case) action.
+
+  $ cat > dune-project << EOF
+  > (lang dune 3.11)
+  > EOF
+
+  $ cat > version << EOF
+  > 1
+  > EOF
+
+Basic usage:
+
+  $ cat > dune << EOF
+  > (rule
+  >  (alias foo)
+  >  (action
+  >   (case %{read-lines:version}
+  >    (1 (echo "version is 1"))
+  >    (2 (echo "version is 2"))
+  >    (3 (echo "version is 3"))
+  >    (_ (echo "shouldn't happen, got %{read-lines:version}")))))
+  > EOF
+
+  $ dune build @foo
+  version is 1
+
+  $ cat > version << EOF
+  > 2
+  > EOF
+  $ dune build @foo
+  version is 2
+
+  $ cat > version << EOF
+  > 3
+  > EOF
+  $ dune build @foo
+  version is 3
+
+  $ cat > version << EOF
+  > 4
+  > EOF
+  $ dune build @foo
+  shouldn't happen, got 4
+
+  $ cat > version << EOF
+  > 1
+  > EOF
+
+Missing default field:
+
+  $ cat > dune << EOF
+  > (rule
+  >  (alias foo)
+  >  (action
+  >   (case %{read-lines:version}
+  >    (1 (echo "version is 1")))))
+  > EOF
+
+  $ dune build @foo
+  File "dune", line 5, characters 4-5:
+  5 |    (1 (echo "version is 1")))))
+          ^
+  Error: The final branch must be the default one.
+  Hint: Add a (_ (...)) case at the end.
+  [1]
+
+Two default fields:
+
+  $ cat > dune << EOF
+  > (rule
+  >  (alias foo)
+  >  (action
+  >   (case %{read-lines:version}
+  >    (_ (echo "version is 1"))
+  >    (2 (echo "version is 2"))
+  >    (_ (echo "version is 3")))))
+  > EOF
+
+  $ dune build @foo
+  Error: Multiple default cases.
+  File "dune", line 5, characters 4-5:
+  5 |    (_ (echo "version is 1"))
+          ^
+  
+  File "dune", line 7, characters 4-5:
+  7 |    (_ (echo "version is 3")))))
+          ^
+  
+  [1]
+
+Last case is not default case:
+
+  $ cat > dune << EOF
+  > (rule
+  >  (alias foo)
+  >  (action
+  >   (case %{read-lines:version}
+  >    (_ (echo "version is 1"))
+  >    (2 (echo "version is 2"))
+  >    (3 (echo "version is 3")))))
+  > EOF
+
+  $ dune build @foo
+  File "dune", line 7, characters 4-5:
+  7 |    (3 (echo "version is 3")))))
+          ^
+  Error: The final branch must be the default one.
+  Hint: Add a (_ (...)) case at the end.
+  [1]
+
+Empty case:
+
+  $ cat > dune << EOF
+  > (rule
+  >  (alias foo)
+  >  (action
+  >   (case %{read-lines:version})))
+  > EOF
+
+  $ dune build @foo
+  File "dune", line 4, characters 2-30:
+  4 |   (case %{read-lines:version})))
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  Error: Not enough arguments for case
+  [1]
+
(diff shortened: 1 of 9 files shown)


test/blackbox-tests/test-cases/actions/cond.t
```diff
@@ -0,0 +1,134 @@
...
+  >  (action
+  >   (cond
+  >    ((= 1 2) (echo " 1 is 2")))))
+  > EOF
+
+  $ dune build @foo
+  File "dune", line 5, characters 4-11:
+  5 |    ((= 1 2) (echo " 1 is 2")))))
+          ^^^^^^^
+  Error: Only the default case can be at the end.
```
gridbugs, line 37: This error message would be clearer if it read `Case statements must end with a default case.` to clarify that the real problem is the lack of default case - not putting a non-default case at the end.

test/blackbox-tests/test-cases/actions/cond.t
```diff
@@ -0,0 +1,134 @@
...
+  >    ((= 2 3) (echo "2 is 3"))
+  >    (_ (echo "shouldn't happen"))
+  >    ((< 3 4) (echo "3 is less than 4")))))
+  > EOF
+
+  $ dune build @foo
+  File "dune", line 8, characters 4-11:
+  8 |    ((< 3 4) (echo "3 is less than 4")))))
+          ^^^^^^^
+  Error: Only the default case can be at the end.
```
gridbugs, line 92: If I were consuming this error message I'd prefer to be told that the default case appears somewhere other than the end of the case statement in this instance.

test/blackbox-tests/test-cases/actions/cond.t
gridbugs: Also since there is another similar feature which uses the keyword `case` we should avoid using the word "case" in error messages relating to the `cond` feature to avoid confusion between the two features.

test/blackbox-tests/test-cases/actions/case.t
```diff
@@ -0,0 +1,149 @@
...
+  [1]
+
+Two default fields:
+
+  $ cat > dune << EOF
+  > (rule
+  >  (alias foo)
+  >  (action
+  >   (case %{read-lines:version}
+  >    (_ (echo "version is 1"))
```
gridbugs, line 74: Strings in dune can be unquoted. What happens if the `version` file contains "_"? Would the `_` case be treated as the default case or matching the "_" string?

What if we changed the syntax so the default case doesn't require a condition at all. E.g.
```
(case %{read-lines:version}
 (2 (echo "version is 2"))
 (echo "version is not 2"))
```

gridbugs:
I'm looking at using this feature to convert opam commands with filters into dune actions as part of package management. Opam lets entire commands be included or excluded based on a boolean predicate.

For example:
```
build: [
  ["dune" "subst"] {dev}
  ["dune" "build" "-p" name "-j" jobs]
  ["sh" "-ex" "packaging/FreeBSD/create_package.sh"] {os = "freebsd"}
  ["sh" "-ex" "packaging/debian/create_package.sh"] {os-family = "debian"}
]
```

This could be translated a dune action. Note I'm using the `(progn)` action with no arguments as a no-op.
```clojure
(progn
 (cond
  (%{dev} (run dune subst))
  (_ (progn)))
 (run dune build -p %{pkg-self:name} -j %{jobs})
 (cond
  ((= %{os} freebsd) (run sh -ex packaging/FreeBSD/create_package.sh))
  (_ (progn)))
 (cond
  ((= %{os-family} debian) (run sh -ex packaging/debian/create_package.sh))
  (_ (progn))))
```

For improving the ergonomics of using this feature to support commands in lockfiles translated from opam I think it would be useful to allow users to omit the default case and have the conditional action act like a no-op if none of the conditions are met. Also I think it would be useful to add a no-op action to dune so we don't have to use the `(progn)` hack to simulate one.

Alizter:
@gridbugs I think defaulting to the noop will be a poor choice aimilar to the (optional) argument. I would rather a user message get raised when there is no matching branch and allow people to default to a noop if needed.

test/blackbox-tests/test-cases/actions/case.t
Alizter: If `version` contained `_` it would be compared to any other branch with pattern `_`. You can give such a pattern by quoting the `_`. For instance, here is a test I have added:

```

  $ cat > dune << EOF
  > (rule
  >  (alias foo)
  >  (action
  >   (case %{read-lines:version}
  >    ("_" (echo "version is _"))
  >    (_ (echo "version is default")))))
  > EOF

  $ dune build @foo
  version is default

  $ cat > version << EOF
  > _
  > EOF

  $ dune build @foo
  version is _

``` 

### You second point:

This makes parsing more complicated. In fact it can event be ambiguous. Imagine you had a file called `progn`. Then
```
(progn (echo foo))
```
could mean the default case or the string `progn`.

test/blackbox-tests/test-cases/actions/cond.t
Alizter: I've fixed both these issues and opted to use the word "branch".

test/blackbox-tests/test-cases/actions/cond.t
Alizter: This is trickier since we parse the list backwards to check the (reversed) first is a default case. In order to find the default case we would have to parse the rest of them. There is also no guarantee that there is even a default case.

Alizter:
@gridbugs I've added a test demonstrating what happens with `_` as a case in `case.t`. I've also improved the wording of the error message you mentioned.

test/blackbox-tests/test-cases/actions/case.t
gridbugs: Is the `_` string a special value in other places in dune? As a user I would find it confusing for there to be some values which are treated differently when they are quoted since strings without spaces don't usually need to be quoted in dune.

Alizter: It has a precedent as a catchall value in the env stanza.

Alizter:
@gridbugs Something else we can try is allowing for the omission of the default case when in a lockfile and forcing user written actions to have default cases. That way we can have the lockfile convenience without users shooting themselves in the foot.

Alizter:
Closing this as I don't think there is any wish to continue with it. Happy to reopen if needed.

PR #8325, opened 2023-08-02
bug(pkg): Add reproduction case for `cat` not working due to wrong `PATH`

Leonidas-from-XIV:
I ran across this case in #8225 and thought I can extract that into a free-standing example so we can fix it and have a regression test.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/gh8325.t
+++ b/test/blackbox-tests/test-cases/pkg/gh8325.t
@@ -1,0 +1,35 @@
+Things should be the same whether dependencies are specified or not.
+
+  $ mkdir dune.lock
+  $ cat >dune.lock/lock.dune <<EOF
+  > (lang package 0.1)
+  > EOF
+
+If we have a package we depend on
+
+  $ mkdir dependency-source
+  $ cat >dune.lock/dependency.pkg <<EOF
+  > (source (copy $PWD/dependency-source))
+  > EOF
+
+And we have a package we want to build
+
+  $ mkdir test-source
+  $ cat >dune.lock/test.pkg <<EOF
+  > (source (copy $PWD/test-source))
+  > (build
+  >  (system "command -v cat > /dev/null 2>&1 || echo no cat"))
+  > EOF
+  $ dune build .pkg/test/target/
+
+Now it fails since adding the dependency modified PATH.
+
+  $ cat >dune.lock/test.pkg <<EOF
+  > (source (copy $PWD/test-source))
+  > ; adding deps breaks cat
+  > (deps dependency)
+  > (build
+  >  (system "command -v cat > /dev/null 2>&1 || echo no cat"))
+  > EOF
+  $ dune build .pkg/test/target/
+  no cat


Leonidas-from-XIV:
I think we discussed a similar issue with @Alizter and @gridbugs before.

test/blackbox-tests/test-cases/pkg/path.t
```diff
@@ -0,0 +1,41 @@
...
+  $ mkdir dependency-source
+  $ cat >dune.lock/dependency.pkg <<EOF
+  > (source (copy $PWD/dependency-source))
+  > EOF
+
+And we have a package we want to build
+
+  $ mkdir test-source
+  $ cat >test-source/parade <<EOF
+  > (=ↀωↀ=)✧  ヾ(*ΦωΦ)ﾉ
```
rgrinberg, line 19: Let's stick to ascii if you don't mind

test/blackbox-tests/test-cases/pkg/path.t
```diff
@@ -0,0 +1,41 @@
...
+  $ cat >dune.lock/test.pkg <<EOF
+  > (source (copy $PWD/test-source))
+  > ; adding deps breaks cat
+  > (deps dependency)
+  > (build
+  >  (system "cat parade"))
+  > EOF
+  $ dune build .pkg/test/target/
+  File ".pkg/test/_unknown_", line 1, characters 0-0:
+  /usr/bin/sh: line 1: cat: command not found
```
gridbugs, line 40: This line is different on my machine:
```diff
-  /usr/bin/sh: line 1: cat: command not found
+  /bin/sh: cat: command not found
```

test/blackbox-tests/test-cases/pkg/path.t
gridbugs, line 1: Can you name this file something more specific. To run this test manually the command is `dune build @test/path` which is very vague and likely to collide with another future test in another directory.

gridbugs (changes requested):

gridbugs:
In `pkg_rules.ml` if the `build_env` function returns an env containing a `PATH` it clobbers the `PATH` variable pulled from the user's environment. I guess we need to start extending the variable adding additional paths instead.

test/blackbox-tests/test-cases/pkg/path.t
rgrinberg: Indeed, we can use `command` to make the test a little more resilient. Taken from stack overflow:

```
command -v cat >/dev/null 2>&1 || { echo "no cat" }
```

test/blackbox-tests/test-cases/pkg/path.t
Leonidas-from-XIV: Sure, done!

test/blackbox-tests/test-cases/pkg/path.t
Leonidas-from-XIV: Sure. Now that I opened the PR I can just name it after the PR number, so people can refer back to this discussion if they're confused :+1: as is the case with a few of our existing tests

Leonidas-from-XIV:
To conclude, for future readers: this issue has been fixed by #8333. Thanks @gridbugs!

PR #8326, opened 2023-08-02
refactor(init): move Use_name handling to parsing

emillon:
`dune init` supports the following pattern:

- `--public pub` will use `pub` as public name
- no `--public` argument means no public name
- `--public` with no argument will reuse the name as the public name

The third case is replaced later, requiring a custom type and some
parameter passing. Instead, we just set the right public name at parse
time.

Signed-off-by: Etienne Millon <me@emillon.org>

### DIFF
--- a/bin/init.ml
+++ b/bin/init.ml
@@ -35,17 +35,6 @@ let atom_conv = Arg.conv (atom_parser, atom_printer)
 
 let component_name_conv = Arg.conv (component_name_parser, atom_printer)
 
-let public_name_conv =
-  let open Component.Options in
-  let parser = function
-    | "" -> Ok Use_name
-    | s -> component_name_parser s |> Result.map ~f:(fun a -> Public_name a)
-  in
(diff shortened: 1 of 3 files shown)


bin/init.ml
```diff
@@ -90,15 +79,40 @@ let context_cwd : Init_context.t Term.t =
   Scheduler.go ~common:common_term ~config (fun () ->
       Memo.run (Init_context.make path))
 
-let public : Component.Options.public_name option Term.t =
+module Public_name = struct
+  type t =
+    | Use_name
+    | Public_name of Dune_lang.Atom.t
```
rgrinberg, line 85: How about using `Lib_name.t` for this?

emillon: Yes, I've got something in the works to use the proper types instead of atoms (this type is only moved in this PR)

Alizter (approved):

rgrinberg (approved):

PR #8328, opened 2023-08-02
refactor: batch async sandbox operations

rgrinberg:
We now create the sandboxing directories and link them in the same async
task


Alizter:
Looking at the sandboxed bench results, I definitely see an improvement with the memory. It does appear to have some slow down in the clean build time however.

dkalinichenko-js (approved):

rgrinberg:
> I definitely see an improvement with the memory. It does appear to have some slow down in the clean build time however.

I think that means there's something wrong with the benchmark. This PR is basically just a refactoring that shouldn't touch memory consumption at all.

Issue #8327, opened 2023-08-02
melange: introduce variable for `melange.emit` target dir


anmonteiro:
## Desired Behavior

When `melange.emit` is placed in nested subdirectories, or otherwise when folders move around, it can be hard to get a stable path to the melange output directory.

I'd like to introduce a variable to reference the target directory for these cases. One idea is `%{melange.emit:TARGET_DIR}`.

## Example

Given:

```scheme
(melange.emit
  (target out1)
  ...
)

(melange.emit
  (target out2)
  ...
)
```

The following variables would be accessible:

- `%{melange.emit:out1}`
- `%{melange.emit:out2}`

andreypopp:
Given the setup:
```
dune-project
lib/dune <- here we have (melange.emit (target ./out1) ...)
```
then the `_build` will look like:
```
_build/default/lib/out1/lib/
```

So will `%{melange.emit:out1}` refer to the `_build/default/lib/out1/` (emit root) or to the `_build/default/lib/out1/lib/` (dir corresponding to the `dune` file within that root)?

If the former then it's possible already to refer just to `./out1` from the same `dune` file, if the latter then I think the `%{melange.emit:out1}` is confusing.

I propose instead to have `%{melange.emit:out1:index}` to be able to refer to the specific module emitted by melange within the specific melange root.

The motivation is to be able configure the linker/bundler for emitted artefacts:
```
(melange.emit (target out) ...)

(rule
 (target ./bundle.js)
 (deps (source_tree out1))
 (action
  (run esbuild --bundle --outfile=%{target} %{melange.emit:out1:index})))
```

So `%{melange.emit:out1:index}` will expand here to `./out1/lib/index.js`. Why I cannot used `./out1/lib/index.js`? Because it includes `/lib/` path segment and so depends on the hierarchy my lib is placed (e.g. breaks if I vendor it within another project or just move stuff around).

anmonteiro:
> So will %{melange.emit:out1} refer to the _build/default/lib/out1/ (emit root) or to the _build/default/lib/out1/lib/ (dir corresponding to the dune file within that root)?

I was thinking the 2nd option.

> I propose instead to have %{melange.emit:out1:index}

I suppose there isn't an additional benefit of this since if you implement 1. then you can use `%{melange.emit:out1}/index.js`.

Additionally, we support multiple extensions, so it's ambiguous which file `index` would refer to in that case.

anmonteiro:
[...]

PR #8329, opened 2023-08-02
melange: add variables for melange.emit entries

anmonteiro:
- A first attempt at introducing variables for `melange.emit` JS entries
- I feel like this initial approach won't work because there's no way to mutate the expander bindings, and we'd like to introduce variables for other rules
    - I suspect this design is simply wrong, but wanted to put something up for initial discussion

fixes https://github.com/ocaml/dune/issues/8327


rgrinberg:
Why do you need to mutate the expander bindings? You can create a new expander for every rule to expand things differently.

Closing this for inactivity. Feel free to reopen if you resume working on this.

Issue #7108, opened 2023-02-17
Dune init's `--public` doesn't allow dashes


reynir:
## Expected Behavior
Dune init's `--public` can take names with dashes in them.

## Actual Behavior
An error explaining only characters `[a-zA-Z0-9_]` are allowed.

## Reproduction

```shell
$ dune init lib mirage_pair --public mirage-pair
dune: option '--public': invalid component name `mirage-pair'
      Library names must be non-empty and composed only of the
      following
      characters: 'A'..'Z', 'a'..'z', '_' or '0'..'9'.
Usage: dune init library [OPTION]… NAME [PATH]
Try 'dune init library --help' or 'dune --help' for more information.
```

## Specifications

- Version of `dune` (output of `dune --version`): 3.6.1
- Version of `ocaml` (output of `ocamlc --version`) 4.14.0
- Operating system (distribution and version): debian 11.6

Alizter:
Does putting it in quotations work?

reynir:
Sorry how do you mean?

Alizter:
I mean
```
dune init lib mirage_pair --public "mirage-pair"
```

reynir:
No, that does the same.

reynir:
The relevant code is here:
https://github.com/ocaml/dune/blob/d3d628f2eda2278bd2df6e37452d8693f367fcfd/bin/init.ml#L16-L48

The problem is the same converter for library names is used for the public library name. It uses `Lib_name.Local`

https://github.com/ocaml/dune/blob/d3d628f2eda2278bd2df6e37452d8693f367fcfd/src/dune_rules/lib_name.ml#L17-L19

I'm unsure if there's existing code that can be used for the public name or how to go forward on this.

PizieDust:
currently having this issue too

PR #8331, opened 2023-08-03
feature: check package names are valid opam names

emillon:
This adds a `Package_name.Strict` variant that uses opam conventions.
The corresponding parser is used if lang dune >= 3.11.

### DIFF
--- a/src/dune_lang/package_name.ml
+++ b/src/dune_lang/package_name.ml
@@ -10,10 +10,52 @@ include (
     let description = "package name"
     let description_of_valid_string = None
     let hint_valid = None
-
-    let of_string_opt s =
-      (* DUNE3 verify no dots or spaces *)
-      if s = "" then None else Some s
-    ;;
+    let of_string_opt s = if s = "" then None else Some s
   end) :
     Dune_util.Stringlike with type t := t)
+
+module Opam_compatible = struct
+  include Dune_util.Stringlike.Make (struct
+      type t = string
+
+      let module_ = "Package.Name.Strict"
+      let description = "opam package name"
+      let to_string s = s
+
+      let description_of_valid_string =
+        Some
+          (Pp.textf
+             "Package names can contain letters, numbers, '-', '_' and '+', and need to \
+              contain at least a letter.")
+      ;;
+
+      let is_letter = function
+        | 'a' .. 'z' | 'A' .. 'Z' -> true
+        | _ -> false
+      ;;
+
+      let is_other_valid_char = function
+        | '0' .. '9' | '-' | '+' | '_' -> true
+        | _ -> false
+      ;;
+
+      let is_valid_char c = is_letter c || is_other_valid_char c
+
+      let is_valid_string s =
+        let all_chars_valid = String.for_all s ~f:is_valid_char in
+        let has_one_letter = String.exists s ~f:is_letter in
+        all_chars_valid && has_one_letter
+      ;;
+
+      let of_string_opt s = Option.some_if (is_valid_string s) s
+
+      let make_valid s =
+        let replaced = String.map s ~f:(fun c -> if is_valid_char c then c else '_') in
+        if is_valid_string replaced then replaced else "p" ^ replaced
+      ;;
+
+      let hint_valid = Some make_valid
+    end)
+
+  let to_package_name s = s
+end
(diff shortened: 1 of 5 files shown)


emillon:
Context: that was a forgotten `DUNE3` comment; and this will be useful for #7108.

emillon:
Hmm, I made a mistake while trying to replicate the format. It should contain at least one letter, but can start with a number, see `0install`.

src/dune_lang/package_name.ml
```diff
@@ -15,8 +15,54 @@ include (
...
+
+    let description = "package name"
+
+    let to_string s = s
+
+    let description_of_valid_string =
+      Some
+        (Pp.textf
+           "Package names can contain letters, '-', '_' and '+', and need to \
+            contain at least a letter.")
```
Alizter, line 36:
```suggestion
           "Package names can contain numbers, '-', '_' and '+', and need to \
            contain at least a letter.")
```

src/dune_rules/dune_project.ml
```diff
@@ -841,175 +841,172 @@ let parse_packages (name : Name.t option) ~info ~dir ~version packages
...
-       ; explicit_js_mode
-       ; format_config
-       ; subst_config
-       ; strict_package_deps
-       ; allow_approximate_merlin
-       ; cram
-       ; expand_aliases_in_sandbox
-       ; opam_file_location
-       }
+  @@
```
Alizter, line 844: Where is this diff coming from?

src/dune_lang/package_name.ml
```diff
@@ -15,8 +15,54 @@ include (
...
+
+    let of_string_opt s =
+      let state = ref (Ok false) in
+      String.iter s ~f:(fun c ->
+          match (!state, c) with
+          | Error (), _ -> ()
+          | _, ('a' .. 'z' | 'A' .. 'Z') -> state := Ok true
+          | _, ('0' .. '9' | '-' | '_' | '+') -> ()
+          | _ -> state := Error ());
```
rgrinberg, line 45: A little weird to keep iterating if you already found an error. Also, I'm not sure why you need three status. `type state = Valid | Invalid` would do and you can start with Invalid.

I also I see a couple of good alternatives:

1. Use a regular expression to make this code easier to understand
2. Use `OpamPackage.Name.of_string`. This will require catching `Failure _`.

src/dune_rules/dune_project.ml
emillon: this is caused by the extra `let* in`

src/dune_lang/package_name.ml
emillon: I started from the state machine from `OpamPackage.Name.of_string`. 3 states are necessary because there are 2 ways to be invalid: you encountered an invalid character (you can't recover from this), or you have not yet found a letter.

But I haven't realized I could use `re` in dune. I'll switch to that.

emillon: Done. But I had to add `re` in `dune_lang/dune`. I'm not sure that's OK.

rgrinberg: It's not ideal. If the check is indeed so simple, you can just make it:

```
let valid = String.length x > 0 && String.forall ~f:is_valid_char in
```

But why can't you use the opam library directly? Opam is the source of truth for what is a valid package name. We should defer to it in such cases.

src/dune_lang/package_name.ml
```diff
@@ -10,10 +10,52 @@ include (
...
-
-    let of_string_opt s =
-      (* DUNE3 verify no dots or spaces *)
-      if s = "" then None else Some s
-    ;;
+    let of_string_opt s = if s = "" then None else Some s
   end) :
     Dune_util.Stringlike with type t := t)
+
+module Strict = struct
```
Leonidas-from-XIV, line 17: I'm not sure this is a good name. It is a stricter version than what dune-packages can be but the usecase would be opam compatibility, so maybe `Opam_compatible`?

src/dune_lang/package_name.ml
```diff
@@ -10,10 +10,52 @@ include (
...
+  include Dune_util.Stringlike.Make (struct
+      type t = string
+
+      let module_ = "Package.Name.Strict"
+      let description = "package name"
```
Leonidas-from-XIV, line 22:
```suggestion
      let description = "valid opam package name"
```

emillon:
I fixed the formatting, `@@` can be left in place actually.

rgrinberg (approved):

PR #8333, opened 2023-08-03
Pkg: Merge PATHs from user and package envs

gridbugs:
The user PATH is needed so system executables (like `cat`) can be found and the package PATH is needed to execute binaries from packages.

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -329,7 +329,14 @@ module Pkg = struct
         ; "OPAMCLI", "2.0"
         ]
     in
-    let env = build_env t |> Env.Map.map ~f:Env_update.string_of_env_values in
+    let package_env =
+      let vars =
+        build_env t
+        |> Env.Map.map ~f:Env_update.string_of_env_values
+        |> Env.Map.superpose base
+      in
+      Env.extend Env.empty ~vars
+    in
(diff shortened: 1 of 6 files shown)


src/dune_rules/pkg_rules.ml
```diff
@@ -334,7 +334,21 @@ module Pkg = struct
...
+    let path_key = "PATH" in
+    let combined_path =
+      match (Env.get user_env path_key, Env.Map.find package_env path_key) with
+      | None, other -> other
+      | Some path, None -> Some path
+      | Some user_path, Some package_path ->
+        (* Put the packge PATH before the user PATH so executables from
+           packages are preferred over locally-installed executables of the
+           same name. *)
+        Some (String.concat ~sep:":" [ package_path; user_path ])
```
rgrinberg, line 348: Would be good to move this to `Env_path`.` We try to avoid handling raw paths in here.

gridbugs: Ah yep, didn't know about `Env_path`. I've added a comment to env.ml to direct people to `Env_path` if they came to env.ml in search of operations relating to the PATH variable. Also added some documentation to clarify some of the operations for extending envs.

I added a function `Env_path.extend_env_concat_path` that merges two environments concatenating their PATHs.

PR #8336, opened 2023-08-04
Start copying commands from opam files

gridbugs:
Very early implementation of copying build and install commands from opam files into lockfiles meant to invite feedback on the general design of this feature before going deeper.

Not yet supported:
- filters on commands or arguments
- string interpolation in commands
- package-scoped variables other than `_` (the current package)
- executing the actions

Not all opam variables are supported yet.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -1,5 +1,5 @@
 open Stdune
-module Package_name = Dune_lang.Package_name
+open Dune_lang
 
 module type CONTEXT = Opam_0install.S.CONTEXT
 
@@ -230,6 +230,78 @@ module Summary = struct
   ;;
 end
 
+let opam_command_to_string_debug (args, _filter_opt) =
+  List.map args ~f:(fun (simple_arg, _filter_opt) ->
+    match simple_arg with
+    | OpamTypes.CString s -> String.quoted s
+    | CIdent ident -> ident)
+  |> String.concat ~sep:" "
+;;
+
+let opam_commands_to_actions package (commands : OpamTypes.command list) =
+  let pform_of_ident_opt ident =
+    let `Self, variable =
+      match String.split ident ~on:':' with
+      | [ variable ] | [ "_"; variable ] -> `Self, variable
+      | _ ->
+        (* TODO *)
+        Code_error.raise
+          "Evaluating package variables for non-self packages not yet implemented"
+          [ "While processing package:", Dyn.string (OpamPackage.to_string package)
+          ; "Variable:", Dyn.string ident
+          ]
+    in
+    match Pform.Var.Pkg.of_opam_variable_name_opt variable with
+    | Some pkg_var -> Ok Pform.(Var (Var.Pkg pkg_var))
+    | None -> Error (`Unknown_variable variable)
+  in
+  List.filter_map commands ~f:(fun ((args, _filter_opt) as command) ->
+    let terms =
+      List.map args ~f:(fun (simple_arg, _filter_opt) ->
+        match simple_arg with
+        | OpamTypes.CString s ->
+          (* TODO: apply replace string interpolation variables with pforms *)
+          String_with_vars.make_text Loc.none s
+        | CIdent ident ->
+          (match pform_of_ident_opt ident with
+           | Ok pform -> String_with_vars.make_pform Loc.none pform
+           | Error (`Unknown_variable variable) ->
+             (* Note that the variable name is always quoted to clarify
+                the error message in cases where the grammar of the
+                sentence would otherwise be unclear, such as:
+
+                - Encountered unknown variable type while processing...
+                - Encountered unknown variable name while processing...
+
+                In these examples, the words "type" and "name" are variable
+                names but it would be easy for users to misunderstand those
+                error messages without quotes. *)
+             User_error.raise
+               [ Pp.textf
+                   "Encountered unknown variable %S while processing commands for \
+                    package %s."
+                   variable
+                   (OpamPackage.to_string package)
+               ; Pp.text "The full command:"
+               ; Pp.text (opam_command_to_string_debug command)
+               ]))
+    in
+    match terms with
+    | program :: args -> Some (Action.run program args)
+    | [] -> None)
+;;
+
+(* returns:
+   [None] if the command list is empty
+   [Some (Action.Run ...)] if there is a single command
+   [Some (Action.Progn [Action.Run ...; ...])] if there are multiple commands *)
+let opam_commands_to_action package (commands : OpamTypes.command list) =
+  match opam_commands_to_actions package commands with
+  | [] -> None
+  | [ action ] -> Some action
+  | actions -> Some (Action.Progn actions)
+;;
+
 let opam_package_to_lock_file_pkg ~repo ~local_packages opam_package =
   let name = OpamPackage.name opam_package in
   let version = OpamPackage.version opam_package |> OpamPackage.Version.to_string in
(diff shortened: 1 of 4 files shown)


test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
```diff
@@ -0,0 +1,69 @@
...
+  >  (depends standard-dune))
+  > EOF
+  Solution for dune.lock:
+  standard-dune.0.0.1
+  
+
+  $ cat dune.lock/standard-dune.pkg
+  (version 0.0.1)
+  (install (progn (system "%{make} install")))
+  (build (progn (system "dune subst") (system "dune build -p %{name} -j %{jobs} @install @runtest @doc")))
```
rgrinberg, line 57: Is using `system` correct? I imagine it should be `(run dune subst)`.

gridbugs: I thought `(system "dune subst")` and `(run dune subst)` are equivalent. According to the dune docs:

> Execute a command using the system shell: sh on Unix and cmd on Windows.
> Example:
> `(system "command arg1 arg2")`

I deliberately chose to use `system` instead of `run` as opam commands might include `&&` or similar operators and I don't think it's possible to do `(run make && make install)` say whereas `(system "make && make install")` should work.

Alizter: Well you can turn && into a progn with a bit of parsing.

gridbugs: Sure, but for this initial implementation let's keep things as simple as possible. Parsing commands can get complicated since these can be arbitrary commands with "&&" possibly appearing in a string, etc. What's the benefit of representing commands with `Run` actions over `System` actions?

Leonidas-from-XIV: I would also suggest `system` (or `bash` even), given we probably don't want to reimplement a whole parser for shell syntax for now, especially given we support shell commands already.

rgrinberg: I see. Opam is using shell then we should copy what it's doing. However, is opam really using `sh` or `bash`? Could you double check that.

Leonidas-from-XIV: I did some digging on how OPAM runs commands. `OpamAction.build_package` collects the build commands and converts them eventually into some `OpamProcess.Job` structure and if I see it correctly to evaluate one of these it does call  `Unix.create_process_env` in the end. 

So it doesn't seem to involve a shell, thus presumably `&&` and redirections wouldn't be valid, a package would need to wrap it in `sh -c`.

By the way, we have our own runner for `OpamProcess.Job` that uses Fibers, so maybe the part that collects commands in OPAM could be leveraged. Don't even necessarily need to run them, we can also just unpack a job into its commands again and write them into the lockfile, since the `OpamProcess.Job` data structure exposes all the internal parts.

gridbugs: Thanks for looking into this @Leonidas-from-XIV. I wasn't expecting it to work that way. I'll see if the way that opam represents commands can be mapped onto dune's `run` action.

gridbugs:
Updated to use `run` instead of `system`. Also it now omits the `build` and `install` fields if there are no commands and only uses `Progn` if there were multiple commands.

test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
rgrinberg: > By the way, we have our own runner for OpamProcess.Job that uses Fibers, so maybe the part that collects commands in OPAM could be leveraged. Don't even necessarily need to run them, we can also just unpack a job into its commands again and write them into the lockfile, since the OpamProcess.Job data structure exposes all the internal parts.

We need to work with actions here otherwise the build system will not offer incremental building.

> I wasn't expecting it to work that way. I'll see if the way that opam represents commands can be mapped onto dune's run action.

FYI, I expected that it would map quite naturally. However, if it doesn't, feel free to add new constructs to the action language to match opam's semantics.

test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
```diff
@@ -0,0 +1,69 @@
...
+  >  (name x)
+  >  (depends standard-dune))
+  > EOF
+  Solution for dune.lock:
+  standard-dune.0.0.1
+  
+
+  $ cat dune.lock/standard-dune.pkg
+  (version 0.0.1)
+  (install (run %{make} install))
```
rgrinberg, line 56: Are you converting any package variables or is this saved for subsequent PR's?

rgrinberg (approved):
LGTM. A few more tests (with more forms) would be useful, but they can always be added later.

src/dune_pkg/opam_solver.ml
```diff
@@ -230,6 +230,80 @@ module Summary = struct
...
+    match simple_arg with
+    | OpamTypes.CString s -> s
+    | CIdent ident ->
+      (* Stringify the variable in the syntax of an opam string interpolation. *)
+      Printf.sprintf "%%{%s}%%" ident)
+  |> String.concat ~sep:" "
+;;
+
+let opam_commands_to_actions package (commands : OpamTypes.command list) =
+  let pform_opt_of_ident ident =
```
Leonidas-from-XIV, line 244: Likewise `pform_of_ident_opt`?

src/dune_pkg/opam_solver.ml
```diff
@@ -230,6 +230,80 @@ module Summary = struct
...
+    match terms with
+    | program :: args -> Some (Action.run program args)
+    | [] -> None)
+;;
+
+(* returns:
+   [None] if the command list is empty
+   [Some (Action.Run ...)] if there is a single command
+   [Some (Action.Progn [Action.Run ...; ...])] if there are multiple commands *)
+let opam_commands_to_action_opt package (commands : OpamTypes.command list) =
```
Leonidas-from-XIV, line 300: To break out of my complaining about `_opt`, maybe this doesn't need `_opt` at all? Since That just sort of implements hungarian notation, but for that we have the type system and `None` is a valid value and not an exceptional case (so an `_exn` variant wouldn't make sense).

src/dune_pkg/opam_solver.ml
```diff
@@ -230,6 +230,80 @@ module Summary = struct
...
+                names but it would be easy for users to misunderstand those
+                error messages without quotes. *)
+             User_error.raise
+               [ Pp.textf
+                   "Encountered unknown variable \"%s\" while processing commands for \
+                    package %s."
+                   variable
+                   (OpamPackage.to_string package)
+               ; Pp.text "The full command:"
+               ; Pp.text (opam_command_to_string_debug command)
```
Leonidas-from-XIV, line 288: I wonder whether it wouldn't be better to express this in the OPAM file syntax, because if I look at the test it states `%%{fake}%% install` but this doesn't occur in the source, instead it states `fake "install"` which inverts which parts get quoted and which ones don't.

Leonidas-from-XIV (commented):
I have some small comments about this but overall I think its good and moving to the right direction. Good chances that even without resolving of the package variables this is enough to cover 80% of all OPAM packages.

test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
gridbugs: I'll do this in a future PR. There's some design questions about namespacing package variables that we need to work out and I don't want this change to be blocked by that.

gridbugs:
Note that long-term we probably won't be able to use `Run` actions to implement opam commands due to filters on individual terms. Opam allows filters on the first term of a command, for example:
```
build: [ "make" {os-family = "linux"} "gmake" {os-family = "freebsd"} ]
```
Even though opam treats the first term of a command as the program and subsequent terms as arguments, filtering takes place before determining which term is first, so the program can be determined dynamically based on filters. In dune this filtering must happen at build time rather than solve time in order for single lockdirs to support multiple environments. I think we'll need a new construct to handle filters on terms.

This doesn't affect this PR. Just an observation.

src/dune_pkg/opam_solver.ml
```diff
@@ -230,6 +230,78 @@ module Summary = struct
...
+
+                - Encountered unknown variable type while processing...
+                - Encountered unknown variable name while processing...
+
+                In these examples, the words "type" and "name" are variable
+                names but it would be easy for users to misunderstand those
+                error messages without quotes. *)
+             User_error.raise
+               [ Pp.textf
+                   "Encountered unknown variable \"%s\" while processing commands for \
```
rgrinberg, line 281: This can be `%S` for automatic quoting

gridbugs: woah surprising, in C it's for strings of wide characters right?

Leonidas-from-XIV: I am not sure it is standard, I can only find `%S` in Microsoft reference documents, not sure it also applies to POSIX.

PR #8340, opened 2023-08-04
fix: dune describe pp no longer dumps into source root

### LINKED ISSUE #7430: `dune describe pp` creates an empty `profile.dump` file at the root of the project
## Expected Behavior

`dune describe pp some/file.ml` should print the preprocess file.

## Actual Behavior

It does print the preprocessed file but also creates a `profile.dump` file at the root of the project. This file is empty. I couldn't find a mention of this file in `_build/log`,

## Reproduction

```shell
git clone https://github.com/hhugo/merge-fmt
dune describe pp test/merge.ml
```

## Specifications

- Version of `dune` (output of `dune --version`): 3.6.1 (but also happening on an experimental version of dune)
- Operating system (distribution and version): debian 11

## Additional information 

- Link to gist with verbose output (run `dune` with the `--verbose` flag): https://gist.github.com/Khady/9660bc040c9c8fc3af90aa50e1fab0cc

Alizter:
We also fix this by running ocamlc in the _build directory and cleaning up the dump file afterwards.

- fix #7430
- [x] changelog

### DIFF
--- a/bin/describe/describe_pp.ml
+++ b/bin/describe/describe_pp.ml
@@ -18,15 +18,21 @@ let pp_with_ocamlc sctx project pp_file =
   let+ () =
     Process.run
       ~display:!Clflags.display
+      ~dir:(Path.build (Super_context.context sctx).build_dir)
       ~env:(Super_context.context_env sctx)
       Strict
       (Super_context.context sctx).ocaml.ocamlc
-      [ "-stop-after"; "parsing"; "-dsource"; Path.to_string pp_file; "-dump-into-file" ]
+      [ "-stop-after"
+      ; "parsing"
+      ; "-dsource"
+      ; Path.to_string (Path.drop_optional_build_context_maybe_sandboxed pp_file)
+      ; "-dump-into-file"
+      ]
   in
   match Path.stat dump_file with
   | Ok { st_kind = S_REG; _ } ->
     Io.cat dump_file;
-    Path.unlink_no_err dump_file
+    Path.unlink dump_file
   | _ ->
     User_error.raise [ Pp.textf "cannot find a dump file: %s" (Path.to_string dump_file) ]
 ;;
--- a/test/blackbox-tests/test-cases/describe/describe-pp/describe-pp.t/run.t
+++ b/test/blackbox-tests/test-cases/describe/describe-pp/describe-pp.t/run.t
@@ -11,4 +11,4 @@ We can also show the original source if it is not preprocessed
 We also make sure that the dump file is not present
 
   $ dune_cmd exists profile.dump
-  true
+  false
--- a/doc/changes/8340.md
+++ b/doc/changes/8340.md
@@ -1,0 +1,2 @@
+- Fix `dune describe pp` from leaving behind a dump file. (#8222, fixes #7430,
+  @Alizter)


bin/describe/describe_pp.ml
```diff
@@ -18,15 +18,22 @@ let pp_with_ocamlc sctx project pp_file =
...
+      ; Path.to_string (Path.drop_optional_build_context_maybe_sandboxed pp_file)
+      ; "-dump-into-file"
+      ]
   in
   match Path.stat dump_file with
   | Ok { st_kind = S_REG; _ } ->
     Io.cat dump_file;
-    Path.unlink_no_err dump_file
+    Path.unlink_no_err dump_file;
+    Path.rm_rf dump_file
```
Leonidas-from-XIV, line 36: What is the purpose of this double remove? If `unlink_no_err` succeeds `rm_rf` will catch `ENOENT` and return unit, it can never be a directory anyway and otherwise it will call `unlink` (without catching exceptions), so if exposing the exception is the desired effect, `Path.unlink` can be used directly instead of `Path.unlink_no_err`.

Alizter: Yes, good point. Fixed.

test/blackbox-tests/test-cases/describe/describe-pp/describe-pp.t/run.t
```diff
@@ -11,4 +11,4 @@ We can also show the original source if it is not preprocessed
 We also make sure that the dump file is not present
 
   $ dune_cmd exists profile.dump
-  true
+  false
```
Leonidas-from-XIV, line 14: I think this needs to be updated, since the cwd for running the compiler is different the compiler dumps `profile.dump` in some other place (namely `Path.build (Super_context.context sctx).build_dir`) so presumably something like

```
$ dune_cmd exists _build/default/profile.dump
false
```

Otherwise the `unlink` can still be not triggered and at the same time passing the test just fine.

Alizter: The original bug was about the dump appearing in the source tree. I'll add a test for whatever is happening in the _build dir.

Alizter: Hmm actually I confused myself here. There are two kinds of .dump files. We have `profile.dump` which is what we were seeing in the source tree and we have the individual module dumps. We can see that the module dumps are being unlinked, but in the end it doesn't actually matter since it all happens in the build directory. Therefore I think this is fine as it is.

Leonidas-from-XIV: Should we then even try unlinking them if they're in the `_build` directory? If it doesn't matter, then maybe there is not much point in wasting IO and time. What is our policy for this @rgrinberg?

rgrinberg: Tbh, it's not even a good idea to write them in the _build directory either. Ideally, we'd copy the input into some temp dir and then run it there. Yes, we should clean up after ourselves if we are going to pollute the build dir.

Alizter:
The design here is a bit flawed, but I don't really have anything better to suggest.

PR #8341, opened 2023-08-04
test: make install-libdir more repro

Alizter:
We modify the test so that the current opam prefix doesn't have an affect on the test itself. We do this by mocking up an opam prefix which is added to the env when a user does `opam env`.

This makes this test reproducible on Nix and also we avoid touching directories outside of the test.


Alizter:
Apparently not as repro as I thought.

PR #8349, opened 2023-08-06
feature: allow installing source trees

rgrinberg:
This PR allows us to install source trees with the following field in the install stanza:

```
(install
 (section doc)
 (source_tree dir))
```

### DIFF
--- a/src/dune_rules/install_rules.ml
+++ b/src/dune_rules/install_rules.ml
@@ -410,7 +410,7 @@ end = struct
           let* dirs_expanded =
             Install_entry.Dir.to_file_bindings_expanded i.dirs ~expand_str ~dir
           in
-          let+ files_from_dirs =
+          let* files_from_dirs =
             Memo.List.map dirs_expanded ~f:(fun fb ->
               let loc = File_binding.Expanded.src_loc fb in
               let src = File_binding.Expanded.src fb in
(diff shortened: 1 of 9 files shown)


src/dune_rules/install_rules.ml
```diff
@@ -1033,15 +1083,20 @@ include (
...
-          |> Install.Entry.gen_install_file
+        let open Fiber.O in
+        let+ entries =
+          let+ entries =
+            Fiber.parallel_map entries ~f:(fun (entry : _ Install.Entry.t) ->
+              match entry.kind with
+              | `File -> Fiber.return [ entry ]
+              | `Directory -> Fiber.return (read_dir_recursively entry)
+              | `Source_tree ->
+                (* these are expanded into [`File] *)
```
gridbugs, line 1094: nitpick: the `{` should be a `[`

src/dune_rules/install_rules.ml
```diff
@@ -1033,15 +1083,20 @@ include (
...
+                assert false)
```
gridbugs, line 1095: Should this be a `Code_error`?

gridbugs (approved):

src/dune_rules/install_rules.ml
rgrinberg: Did you attach your comment to the wrong line? There's no `{` on this line`.

gridbugs: ah you're right, for some reason I saw the `[` as a `{`

Issue #2123, opened 2019-05-05
Regression introduced by #2041


rgrinberg:
#2041 introduces a regression as discovered by @hannes [here](https://github.com/mirage/mirage/issues/969#issuecomment-482226514)

PR #8350, opened 2023-08-07
fix: Prevent relative install destinations leaking outside package install dir

gridbugs:
This disallows install dst paths from begining with ".." to prevent them referencing a directory outside the install directories of the package.

This new constraint would prevent globs from being able to refer to files outside the current directory and its descendants, so this also adds a new keyword which changes the prefix of paths matched by a glob in the install stanza. It's similar to the `(files (foo as bar))` syntax in that it allows installed files to be placed in a location other than the one implied by their position in the source tree. It's now possible to write:
`(files (glob_files (some/path/* with_prefix new/path)))` ...which takes all the paths matched by the glob and replaces the `some/path` component with `new/path`.

I haven't yet updated the docs or changelog or written any tests beside modifying the one defined in https://github.com/ocaml/dune/pull/8265 to demonstrate how the feature gets used to give us a chance to discuss and change things about this solution before taking the time to write docs/tests, etc.

Also note that this breaks the test `github2123`. I followed the rabbit hole starting from https://github.com/ocaml/dune/issues/2123 to understand why this is necessary but I still don't understand. The test uses this dune file:
```
(install
 (section lib)
 (files (mirage-xen.pc as ../pkgconfig/mirage-xen.pc)))
```
That file ends up in `_build/install/default/lib/pkgconfig/mirage-xen.pc` which seems like it's escaping the package's install directory by design. @rgrinberg is that right? That issue was made in 2019. Do we know if mirage still needs this feature?

### DIFF
--- a/src/dune_rules/file_binding.ml
+++ b/src/dune_rules/file_binding.ml
@@ -4,18 +4,69 @@ open Memo.O
 type ('src, 'dst) t =
   { src : 'src
   ; dst : 'dst option
+      (* The [dune_syntax] field is used for validation which has different
+         behaviour depending on the version of dune syntax in use. *)
+  ; dune_syntax : Syntax.Version.t
   }
 
-let equal f g { src; dst } t = f src t.src && Option.equal g dst t.dst
+let equal f g { src; dst; dune_syntax } t =
+  f src t.src
+  && Option.equal g dst t.dst
+  && Syntax.Version.equal dune_syntax t.dune_syntax
+;;
+
+let relative_path_starts_with_parent relative_path =
+  match String.lsplit2 relative_path ~on:'/' with
+  | None -> Filename.(equal relative_path parent_dir_name)
+  | Some (first, _) -> String.equal first Filename.parent_dir_name
+;;
+
+let validate_dst_for_install_stanza
+  ~relative_dst_path_starts_with_parent_error_when
+  ~loc
+  dst
+  dune_syntax
+  =
+  if relative_path_starts_with_parent dst
+  then (
+    match relative_dst_path_starts_with_parent_error_when with
+    | `Deprecation_warning_from_3_11 ->
+      let open Syntax.Version.Infix in
+      if dune_syntax >= (3, 11)
+      then
+        User_warning.emit
+          ~loc
+          [ Pp.textf
+              "The destination path %s begins with %s which will become an error in a \
+               future version of Dune. Destinations of files in install stanzas \
+               beginning with %s will be disallowed to prevent a package's installed \
+               files from escaping that package's install directories."
+              (String.maybe_quoted dst)
+              (String.maybe_quoted Filename.parent_dir_name)
+              (String.maybe_quoted Filename.parent_dir_name)
+          ]
+    | `Always_error ->
+      User_error.raise
+        ~loc
+        [ Pp.textf
+            "The destination path %s begins with %s which is not allowed. Destinations \
+             in install stanzas may not begin with %s to prevent a package's installed \
+             files from escaping that package's install directories."
+            (String.maybe_quoted dst)
+            (String.maybe_quoted Filename.parent_dir_name)
+            (String.maybe_quoted Filename.parent_dir_name)
+        ])
+;;
 
 module Expanded = struct
   type nonrec t = (Loc.t * Path.Build.t, Loc.t * string) t
 
-  let to_dyn { src; dst } =
+  let to_dyn { src; dst; dune_syntax } =
     let open Dyn in
     record
       [ "src", pair Loc.to_dyn Path.Build.to_dyn src
       ; "dst", option (pair Loc.to_dyn string) dst
+      ; "dune_syntax", Syntax.Version.to_dyn dune_syntax
       ]
   ;;
 
@@ -23,7 +74,7 @@ module Expanded = struct
   let dst t = Option.map ~f:snd t.dst
   let src_loc t = fst t.src
 
-  let dst_basename { src = _, src; dst } =
+  let dst_basename { src = _, src; dst; dune_syntax = _ } =
     match dst with
     | Some (_, dst) -> dst
     | None ->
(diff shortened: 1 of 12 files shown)


Alizter:
Can we get around with this constraint using symlinks?

gridbugs:
> Can we get around with this constraint using symlinks?

How would that work? Could we allow files outside their package directory provided that they are symlinks?

Leonidas-from-XIV:
> How would that work? Could we allow files outside their package directory provided that they are symlinks?

I guess if you symlink `..` to `foo` in your source tree you could write `foo/bar` and have `bar` end up in the parent folder (not that I tested it). Also not sure what happens if you write `foo/bar` to an `.install` file, whether OPAM will create symlink directories and write through symlink directories. It all feels very "exploity".

rgrinberg:
@gridbugs the fix looks right to me. However, it has to be guarded behind a dune language version. Before we remove the old behavior, we need to go through a deprecation cycle as well.

> That file ends up in _build/install/default/lib/pkgconfig/mirage-xen.pc which seems like it's escaping the package's install directory by design. @rgrinberg is that right? That issue was made in 2019. Do we know if mirage still needs this feature?

I don't know about mirage, but installing pkg-config metadata is something that I'm sure many people are going to need.

@kit-ty-kate what is the recommended way to install such files?

kit-ty-kate:
> @kit-ty-kate what is the recommended way to install such files?

I think using [lib_root](https://opam.ocaml.org/doc/Manual.html#installfield-lib_root) instead of `lib` would be the recommended way to do this.

gridbugs:
Ok so for this PR make it so that the current behaviour is maintained prior to the the next release of dune after which we'll start printing a deprecated message (but it won't be an error).

Should I mention the deprecation in the docs or should the docs just say that dst paths may not begin with ".."?

rgrinberg:
> this is in the test for installing source trees and it looks like this test is deliberately escaping its install directory (everything should be under .../default/doc/mypkg/). Is that intended?

The test is indeed intentionally testing this behavior. For `source_trees`, we can just disable escaping the section root out right since this feature hasn't been introduced.

rgrinberg:
> Should I mention the deprecation in the docs or should the docs just say that dst paths may not begin with ".."?

I think you should mention it. It's useful information for the user, so of course it can only help if it's in the manual.

I wish we more systematic about mentioning versions in the manual as well. This would save users a bunch of time.

gridbugs:
Starting a dst with `..` is now an error when installing source trees and a warning when installing files and dirs. I've added a bunch of tests of this behaviour and also of the `with_prefix` feature when using globs in the install stanza. I added docs for both features as well and also ended up adding more docs and tests for installing source trees to clarify some edge cases.

Added @christinerose to review the documentation changes.

Alizter:
i'm still a big confused about the motivation. Why do we want to restrict inatall directories? That seems like we will be more opinionated than we need to be.

src/dune_rules/file_binding.ml
```diff
@@ -4,26 +4,77 @@ open Memo.O
...
+  dune_syntax
+  =
+  if relative_path_starts_with_parent dst
+  then (
+    match relative_dst_path_starts_with_parent_error_when with
+    | `Deprecation_warning_from_3_11 ->
+      let open Syntax.Version.Infix in
+      if dune_syntax >= (3, 11)
+      then
+        User_warning.emit
```
rgrinberg, line 37: Warnings should be disabled in vendored projects.

src/dune_rules/install_entry.ml
```diff
@@ -13,11 +13,46 @@ let expand_str_with_check_for_local_path ~expand_str sw =
...
+  let decode =
+    let open Dune_lang.Decoder in
+    let install_glob_version_check = Dune_lang.Syntax.since Stanza.syntax (3, 6) in
+    let install_glob_with_prefix_version_check =
+      Dune_lang.Syntax.since Stanza.syntax (3, 11)
+    in
+    let decode_args ~recursive =
+      let decode_without_prefix =
+        String_with_vars.decode
+        >>| fun glob ->
```
rgrinberg, line 31: nitpick: this style is never an improvement over:

```
let+ glob = String_with_vars.decode in
...
```

The only time we should use `>>|` or `>>=` is whenever we get to do something point-free. Otherwise, it's just making us read `let` bindings in reverse order.

src/dune_rules/install_entry.ml
```diff
@@ -13,11 +13,46 @@ let expand_str_with_check_for_local_path ~expand_str sw =
...
+        { glob_files = { Dep_conf.Glob_files.glob; recursive }; prefix = None }
+      in
+      let decode_with_prefix =
+        enter
+          (let* () = install_glob_with_prefix_version_check in
+           let* glob = String_with_vars.decode in
+           let* _ = keyword "with_prefix" in
```
rgrinberg, line 38: Might as well make it `let* () =` here.

src/dune_rules/install_entry.ml
```diff
@@ -26,36 +61,55 @@ module File = struct
...
+        List.map (Glob_files_expand.Expanded.matches glob_expanded) ~f:(fun path ->
           let src = glob_loc, path in
-          File_binding.Unexpanded.make ~src ~dst:src)
+          let dst =
+            match prefix_loc_opt with
+            | None -> src
+            | Some (prefix, prefix_loc) ->
+              let path_without_prefix =
+                String.drop_prefix path ~prefix:glob_prefix
+                |> function
```
rgrinberg, line 93: This is the same as writing:

```
match String.drop_prefix path ~prefix:glob_prefix with
```

I'm not against using `|>`, but there's no need to favor it over basic constructs like `match`.

rgrinberg:
Looks good to me with a few minor suggestions remaining.

I don't know if this possible at this point, but I would have liked `with_prefix` to come in a subsequent PR.

Also, I think you need to add a few test cases to make sure `with_prefix` is being validated correctly. For example, absolute paths shouldn't be allowed there.

@Leonidas-from-XIV could you give this one some review as well?

gridbugs:
> While I am not sure this is a good idea from a compatibility point of view (I guess this could prevent a number of packages upgrading to dune-lang 3.11 since we take away something without providing a clear way to migrate to) the code is fine. Some minor comments attached.

Regarding compatibility I can take a look through the monorepo benchmark duniverse to see how many packages take advantage of the `as` keyword to escape their install directory. We know that some packages do this to add pkg-config entries but it would be interesting to see if there are any other use cases that will be affected by this change.

For pkg-config the migration is to use the `lib_root` section. Where's the right place to add migration notes? Should it go in the changelog?

src/dune_rules/file_binding.ml
gridbugs: I don't understand this comment. What are you suggesting I change here?

src/dune_rules/install_entry.ml
```diff
@@ -66,40 +120,68 @@ module File = struct
     ;;
   end
 
-  type t = Without_include.t Recursive_include.t
+  type t =
+    { binding : Without_include.t Recursive_include.t
```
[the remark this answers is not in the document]
gridbugs: I renamed it to `entry` as it represents a single entry in the install stanza.

gridbugs:
> I don't know if this possible at this point, but I would have liked `with_prefix` to come in a subsequent PR.

I've reduced this PR to just the deprecation of dsts starting with "..". I'll move the rest to a new PR.

gridbugs:
I moved the `with_prefix` change to here: https://github.com/ocaml/dune/pull/8416

Leonidas-from-XIV (approved):

rgrinberg:
@gridbugs I rebased the PR. I think it's still showing up warnings in vendored directories. It should be silenced there.

src/dune_rules/file_binding.ml
rgrinberg: In a project that's under a `(vendored_dirs ..)`, we omit all the warnings. This is so that projects that emit projects used by opam-monorepo (or vendored otherwise), do not spam the main projects with warnings.

gridbugs:
@rgrinberg if I'm correctly understanding you regarding vendored directories the problem would only happen if dune tried to build the install manifest for a vendored project. I don't know if it's possible for this to happen. I added a test to the end of `test/blackbox-tests/test-cases/start-install-dst-with-parent-error.t` that I thought might trigger this but it didn't work. Any idea how make dune evaluate the install stanza inside a vendored project?

rgrinberg:
You're right of course. We disable install rules for vendored packages anyway. Anyway, it's just something to keep in mind in general then.

PR #8351, opened 2023-08-07
chore(pkg): Update our vendored OPAM copy

Leonidas-from-XIV:
This updates the our vendored copy to todays OPAM master. It incorporates a small change where on Windows it potentially detects more architectures than it did before.

After talking to @rgrinberg we decided to keep our changes in ocaml-dune/opam on the master branch to be consistent with the other vendored repos (cc @gridbugs).


gridbugs (approved):

src/dune_pkg/sys_poll.ml
```diff
@@ -50,9 +50,13 @@ let arch ~path =
     match Sys.os_type with
     | "Unix" | "Cygwin" -> uname ~path [ "-m" ]
     | "Win32" ->
-      if Sys.word_size = 32 && not (OpamStubs.isWoW64 ())
-      then Fiber.return @@ Some "i686"
-      else Fiber.return @@ Some "x86_64"
+      (match OpamStubs.getArchitecture () with
```
rgrinberg, line 53: Because all the clauses do `Fiber.return`, you can factor it out to the outside:

```ocaml
Fiber.return @@ match ..
```

Leonidas-from-XIV: Indeed, I didn't see the forest for the trees.

PR #8353, opened 2023-08-08
Macros can take multiple arguments

gridbugs:
In the past we simulated macros with multiple arguments by having each macro take a single string argument which used ":" as a defacto delimiter for splitting strings into multiple arguments. This led to logic for splitting macro arguments being duplicated in various parts of our code far removed from actual macro parsing.

This change formalizes the fact that macros can take multiple arguments in the `Pform.t` type and centralizes the logic for parsing arguments delimited by ":".

### DIFF
--- a/src/dune_lang/pform.mli
+++ b/src/dune_lang/pform.mli
@@ -145,7 +145,7 @@ module Macro_invocation : sig
 
     (** Split the payload on the first ':' character raising a [User_error] if the
         payload contains no ':' characters. *)
-    val lsplit2_exn : ?loc:Loc.t -> t -> string * string
+    val lsplit2_exn : t -> Loc.t -> string * string
 
     (** Split the payload on all ':' characters *)
     val split : t -> string list
--- a/src/dune_lang/pform.mli
+++ b/src/dune_lang/pform.mli
@@ -1,5 +1,6 @@
 open Stdune
 open Dune_sexp
+module Payload = Template.Pform.Payload
 
 module Var : sig
   module Pkg : sig
@@ -131,9 +132,29 @@ module Macro : sig
   val to_dyn : t -> Dyn.t
 end
 
+module Macro_invocation : sig
+  type t =
+    { macro : Macro.t
+    ; payload : Payload.t
+    }
+
+  (** Similar to [Payload.Args] but incorporates the macro name into error messages *)
+  module Args : sig
+    (** Treat the entire payload as a single string argument. *)
+    val whole : t -> string
+
+    (** Split the payload on the first ':' character raising a [User_error] if the
+        payload contains no ':' characters. *)
+    val lsplit2_exn : ?loc:Loc.t -> t -> string * string
+
+    (** Split the payload on all ':' characters *)
+    val split : t -> string list
+  end
+end
+
 type t =
   | Var of Var.t
-  | Macro of Macro.t * string
+  | Macro of Macro_invocation.t
 
 val compare : t -> t -> Ordering.t
 val to_dyn : t -> Dyn.t
@@ -141,7 +162,7 @@ val to_dyn : t -> Dyn.t
 type encode_result =
   | Success of
       { name : string
-      ; payload : string option
+      ; payload : Payload.t option
       }
   | Pform_was_deleted
 
--- a/src/dune_rules/expander.ml
+++ b/src/dune_rules/expander.ml
@@ -297,13 +297,9 @@ let file_of_lib { Artifacts.Public_libs.context; public_libs } ~loc ~lib ~file =
   Path.relative dir file
 ;;
 
-let expand_lib_variable t source ~arg:s ~lib_exec ~lib_private =
+let expand_lib_variable t source ~lib ~file ~lib_exec ~lib_private =
   let loc = Dune_lang.Template.Pform.loc source in
-  let lib, file =
-    match String.lsplit2 s ~on:':' with
-    | None -> User_error.raise ~loc [ Pp.textf "invalid %%{lib:...} form: %s" s ]
-    | Some (lib, f) -> Lib_name.parse_string_exn (loc, lib), f
-  in
+  let lib = Lib_name.parse_string_exn (loc, lib) in
   let scope = if lib_exec then t.scope_host else t.scope in
   let p =
     let open Resolve.Memo.O in
@@ -370,7 +366,7 @@ let expand_lib_variable t source ~arg:s ~lib_exec ~lib_private =
                       (Lib_name.to_string lib))
                ]
        | _ ->
-         if (not lib_exec) || (not Sys.win32) || Filename.extension s = ".exe"
+         if (not lib_exec) || (not Sys.win32) || Filename.extension file = ".exe"
          then dep p
          else (
            let p_exe = Path.extend_basename p ~suffix:".exe" in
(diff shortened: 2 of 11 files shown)


rgrinberg:
One reason I'm not a fan of this change is that it consumes more memory for single arguments which are far more common than multiple arguments. So it seems wrong to pessimize the representation for the uncommon case.

Also, does this change how we handle `:` as arguments? e.g. `%{bin:foo:bar}` would look up the binary `foo:bar.exe`. That's almost always unintentional but we can't break it nonetheless.

Does it make sense to keep the arugment type of arguments abstract and handle arity in there? E.g.:

```ocaml
type t
val one : t -> string
val two : t -> string * string
val many : t -> Non_empty_list.t
```
Etc.

src/dune_rules/expander.ml
```diff
@@ -370,7 +378,7 @@ let expand_lib_variable t source ~arg:s ~lib_exec ~lib_private =
                       (Lib_name.to_string lib))
                ]
        | _ ->
-         if (not lib_exec) || (not Sys.win32) || Filename.extension s = ".exe"
+         if (not lib_exec) || (not Sys.win32) || Filename.extension file = ".exe"
```
Leonidas-from-XIV, line 381: This is a bit of an improvement that's generally more correct.

Leonidas-from-XIV (commented):
I was puzzled by this but then realized that `lib:library-name:file`, `lib-private:...`, `libexec` etc support these. I assume this PR is meant for future packaging work, where `pkg:foo:bar` would be useful to have.

To address Rudi's concerns, maybe let's have some common splitting code to be used in all these cases, so it would formalize `:` as separator in one place like in this PR instead of just de-factor because all of the current macros use the same splitting character, as well as common error messages if the arity is incorrect.

gridbugs:
> does this change how we handle `:` as arguments? e.g. `%{bin:foo:bar}` would look up the binary `foo:bar.exe`

I think that's what currently happens (before this change). Everything after the first `:` is treated as the macro argument and its up to the expander to split its argument on `:` if it wants to act like it supports multiple arguments.

My change should fix this behaviour of `bin`. I'll add a test.

gridbugs:
> Does it make sense to keep the arugment type of arguments abstract and handle arity in there?

Yep I think that's preferable to my initial solution. I'll update it to keep the representation abstract.

gridbugs:
> One reason I'm not a fan of this change is that it consumes more memory for single arguments which are far more common than multiple arguments.

Any change from its current representation (`string`) will consume more memory for the single argument case. An alternative to the `string Nonempty_list.t` I'm currently using would be a variant `One of string | Two of (string * string) | Many of string list` but that would also use more memory for the single argument case due to the variant discriminant though not quite so much as always using a list (lists also have discriminant words but also the reference to the rest of the list if they're not empty).

Why is it important to conserve memory here? Would it be reasonable memory-wise to use a variant to represent macro args?

gridbugs:
> Also, does this change how we handle `:` as arguments? e.g. `%{bin:foo:bar}` would look up the binary `foo:bar.exe`. That's almost always unintentional but we can't break it nonetheless.

Ah right yes this would break the case where a user has a binary `foo:bar.exe`.

gridbugs:
I think the right way forward here is to keep representing macro args as a single string but make this abstract and add functions for treating it as a single string or `:`-delimited tuple or list so at least the argument-splitting logic will become centralized. I would be in favour of making a new syntax for macros that uses sexp syntax to separate the macro from its arguments rather than the `:` syntax to be consistent with the rest of dune's syntax and to allow a ":" character to appear in strings passed to multi-argument macros.

Leonidas-from-XIV:
> I think the right way forward here is to keep representing macro args as a single string but make this abstract and add functions for treating it as a single string or `:`-delimited tuple or list so at least the argument-splitting logic will become centralized

That sounds like a good idea.

rgrinberg:
> Why is it important to conserve memory here? Would it be reasonable memory-wise to use a variant to represent macro args?

It's important because it makes dune more efficient in watch mode. However, let me just note that introducing variants here isn't some prohibitive cost. We can eat the cost without any discussion if we're fixing a bug or introducing some new feature. It's just that we don't want to do it just to make code more elegant, especially if the difference is marginal. At the very least, we should consider alternative designs that are more memory friendly. 

> I think the right way forward here is to keep representing macro args as a single string but make this abstract and add functions for treating it as a single string or :-delimited tuple or list so at least the argument-splitting logic will become centralized. I would be in favour of making a new syntax for macros that uses sexp syntax to separate the macro from its arguments rather than the : syntax to be consistent with the rest of dune's syntax and to allow a ":" character to appear in strings passed to multi-argument macros.

Sounds good.

rgrinberg:
> Ah right yes this would break the case where a user has a binary foo:bar.exe.

Also, I think we should add a test for this. You won't be able to use the executable stanza (the module name would be invalid), but you can do something like:

```
$ cat `foo.sh` <<EOF
> #!/usr/bin/env sh
> echo foo
> EOF
$ chmod +x foo.sh
$ cat >dune <<EOF
> (install
>  (section bin)
>  (name foo.sh as foo:bar))
> EOF
```

gridbugs:
> Also, I think we should add a test for this.

Done

src/dune_sexp/lexer.mll
```diff
@@ -348,7 +348,7 @@ and template_variable = parse
       (* -2 to account for the "%{" *)
       let start = { start with pos_cnum = start.pos_cnum - 2 } in
       let loc = Loc.create ~start ~stop:(Lexing.lexeme_end_p lexbuf) in
-      Template.Pform { loc ; name ; payload }
+      Template.Pform { loc ; name ; payload = Option.map ~f:(Template.Pform.Payload.create ~loc) payload }
```
rgrinberg, line 351: I think this location is incorrect. It should exclude the length of the name and the `:` that separates the two.

src/dune_rules/pkg_rules.ml
```diff
@@ -508,8 +508,8 @@ module Action_expander = struct
         let roots = Paths.install_roots paths in
         let dir = section_dir_of_root roots section in
         Memo.return [ Value.Dir dir ]
-      | Macro (Pkg, arg) ->
-        (match String.split arg ~on:':' with
+      | Macro { macro = Pkg; payload } ->
+        (match Pform.Payload.Args.split payload with
```
rgrinberg, line 512: Isn't it possible to use the Macro_invocation api here?

src/dune_sexp/lexer.mll
rgrinberg: Alternatively, you can just get rid of the location inside the payload and ask the callers to provide it.

rgrinberg: Yeah, I think that would be better actually. You would be able to use `Payload.t` when generating pforms without providing fake locations then.

src/dune_rules/pkg_rules.ml
gridbugs: Yep, fixed

src/dune_sexp/lexer.mll
gridbugs: Updated to not store a `Loc.t` inside the payload and added a function to compute the payload's loc based on the pform's loc.

gridbugs:
Rebased and it looks like the commit which adds the adversarial binary names has already been applied in another PR.

Leonidas-from-XIV (approved):

src/dune_lang/pform.mli
```diff
@@ -131,17 +132,37 @@ module Macro : sig
...
+    }
+
+  (** Similar to [Payload.Args] but incorporates the macro name into error messages *)
+  module Args : sig
+    (** Treat the entire payload as a single string argument. *)
+    val whole : t -> string
+
+    (** Split the payload on the first ':' character raising a [User_error] if the
+        payload contains no ':' characters. *)
+    val lsplit2_exn : t -> Loc.t -> string * string
```
rgrinberg, line 148: I made this non optional. Omitting locations is almost always a mistake for something like this.

Issue #8360, opened 2023-08-09
Artifact substitution tries to sign `.ml` files


anmonteiro:
## Expected Behavior

Using dune-site generates a Sites.ml file. I expect this file to not be signed during installation

## Actual Behavior

Dune signs `Sites.ml`, e.g.

```
/nix/store/q9fdiivimqczivx790w1w723jrxs7kkj-ocaml5.1.0-rc1-camomile-1.0.2/lib/ocaml/5.1.0~rc1/site-lib/camomile/lib/.#Sites.ml.dune-temp
```
(as part of Artifact_substitution.copy_file)

cc @emillon

PR #8361, opened 2023-08-09
fix: only sign executables

anmonteiro:
- It looks like macOS's `codesign` tool is a lot more lenient than Nix's `sigtool`, and skips signing files altogether if they're not binaries.
- This PR fixes the Nix part by only signing binaries and not regular files (e.g. sites modules produced by `dune-site`)

fixes https://github.com/ocaml/dune/issues/8360

### DIFF
--- a/bin/install_uninstall.ml
+++ b/bin/install_uninstall.ml
@@ -340,7 +340,8 @@ module File_ops_real (W : sig
             | Dune_package -> process_dune_package ~get_location:conf.get_location
           in
           copy_special_file ~src ~package ~ic ~oc ~f)
-    | None -> Dune_rules.Artifact_substitution.copy_file ~conf ~src ~dst ~chmod ()
+    | None ->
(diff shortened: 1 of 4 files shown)


anmonteiro:
This is a bit hard to test, but it fixes an issue I'm seeing when installing e.g. Camomile v2.0, which now uses dune-site

anmonteiro:
nix-overlays CI run: https://github.com/nix-ocaml/nix-overlays/pull/1002

previously Camomile 2.0 would fail with

> libc++abi: terminating with uncaught exception of type SigTool::NotAMachOFileException: std::exception

Leonidas-from-XIV:
I'm thinking that maybe it would be possible to make a test by overwriting `conf.sign_hook` and storing all `Path.t` passed to it (while calling the previous `sign_hook`), do some builds/installations and test that the only paths passed to the hook were executables?

PR #8362, opened 2023-08-10
Upstream Jane Street changes to Dune engine

dkalinichenko-js:
\+ minor changes to Stdune and Dune rules

### DIFF
--- a/src/dune_engine/build_config.ml
+++ b/src/dune_engine/build_config.ml
@@ -96,6 +96,7 @@ type t =
   ; execution_parameters : dir:Path.Source.t -> Execution_parameters.t Memo.t
   ; source_tree : (module Source_tree)
   ; action_runner : Action_exec.input -> Action_runner.t option
+  ; action_runners : unit -> Action_runner.t list
   ; shared_cache : (module Shared_cache_intf.S)
   }
 
@@ -103,6 +104,7 @@ let t = Fdecl.create Dyn.opaque
 
 let set
   ~action_runner
+  ~action_runners
   ~stats
   ~contexts
   ~promote_source
@@ -140,6 +142,7 @@ let set
     ; execution_parameters
     ; source_tree
     ; action_runner
+    ; action_runners
     ; shared_cache
     }
 ;;
--- a/otherlibs/dune-action-plugin/test/depends-on-its-target-by-read-dir/run.t
+++ b/otherlibs/dune-action-plugin/test/depends-on-its-target-by-read-dir/run.t
@@ -12,35 +12,10 @@
 
   $ cp ./bin/foo.exe ./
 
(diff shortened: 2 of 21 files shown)


src/dune_engine/build_config.ml
```diff
@@ -96,13 +96,15 @@ type t =
   ; execution_parameters : dir:Path.Source.t -> Execution_parameters.t Memo.t
   ; source_tree : (module Source_tree)
   ; action_runner : Action_exec.input -> Action_runner.t option
+  ; action_runners : unit -> Action_runner.t list
```
Alizter, line 99: What's the difference between this and the field above?

Alizter (commented):
Do we want to expose the `stop_on_first_error` behaviour to the user via a config option?

rgrinberg:
As discussed offline: the test suite reveals that our cycles errors aren't marshallable. We should be able to fix that.

I've glanced over the code and there's a couple of things I've noticed that are missing:

1. Stdout/Stderr truncation introduced by Jesse
2. The recording of action duration in Action_exec

I'm not sure if that was intentional, but it's fine to introduce these changes in subsequent PR's if you prefer.

src/dune_engine/build_config.ml
dkalinichenko-js: First one is for selecting the appropriate action runner based on the actions, second one is for notifying all action runners. Merging those is non-trivial since selection depends on some JS-specific code. We plan to refactor this but for now we're trying to minimize the diff between our version and upstream.

dkalinichenko-js:
> Do we want to expose the `stop_on_first_error` behaviour to the user via a config option?

That would be nice, we implemented that at JS but our front-end is different.

dkalinichenko-js:
> As discussed offline: the test suite reveals that our cycles errors aren't marshallable. We should be able to fix that.
> 
> I've glanced over the code and there's a couple of things I've noticed that are missing:
> 
> 1. Stdout/Stderr truncation introduced by Jesse
> 2. The recording of action duration in Action_exec
> 
> I'm not sure if that was intentional, but it's fine to introduce these changes in subsequent PR's if you prefer.

Yeah, both are intentional, I'll do it in later PRs.

src/dune_engine/build_config.ml
Alizter: Very well. Can you include a comment in the mli that says this will be refactored at a later date? Descriptions of the fields would also be nice but not necessary.

Alizter (approved):

Alizter:
> > Do we want to expose the `stop_on_first_error` behaviour to the user via a config option?
> 
> That would be nice, we implemented that at JS but our front-end is different.

OK, once this PR is merged we should open an issue about adding this feature in. We can discuss the details in the issue.

Alizter:
Looks like coveralls is getting confused by this branch being named `main` and also reporting the coverage here as `main`s coverage. I'll let them know upstream.

PR #8363, opened 2023-08-10
Test that the 0install solver can resolve `|` dependencies correctly

Leonidas-from-XIV:
While `depends` can't have dependencies specified via `|`, opam files can. So our dependencies do, we expect them to be resolved correctly. This extends the locking case to show that the 0install solver can pick one of two alternate solutions.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
+++ b/test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
@@ -1,5 +1,29 @@
 Simple example of generating a lock file with Dune
 
+Helper shell function that generates an opam file for a package:
+
+  $ mkpkg() {
+  >   name=$1
+  >   mkdir -p mock-opam-repository/packages/$name/$name.0.0.1
+  >   cat >mock-opam-repository/packages/$name/$name.0.0.1/opam
+  > }
+  $ emptypkg() {
+  >   mkpkg $1 <<EOF
+  > opam-version: "2.0"
+  > EOF
+  > }
+  $ versionpkg() {
+  >   name=$1
+  >   ver=$2
+  >   mkdir -p mock-opam-repository/packages/$name/$name.$ver
+  >   cat >mock-opam-repository/packages/$name/$name.$ver/opam
+  > }
+  $ emptyverpkg() {
+  >   versionpkg $1 $2 <<EOF
+  > opam-version: "2.0"
+  > EOF
+  > }
+
 Generate a `dune-project` file.
   $ cat >dune-project <<EOF
   > (lang dune 3.8)
@@ -7,11 +31,18 @@ Generate a `dune-project` file.
   >  (name lockfile_generation_test)
   >  (depends
   >    foo
-  >    (bar (>= "0.3"))
-  >   ))
+  >    (bar (>= "0.3"))))
+  > EOF
+  > mkpkg foo <<EOF
+  > opam-version: "2.0"
+  > depends: [
+  >     "baz" {>= "0.1"}
+  >     "bar" {>= "0.2"}
+  > ]
   > EOF
 
 Run the solver and generate a lock directory.
+
   $ dune pkg lock --opam-repository-path=mock-opam-repository
   Solution for dune.lock:
   bar.0.5.0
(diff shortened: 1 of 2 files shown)


Alizter (approved):

rgrinberg:
While you're at it, could you add some dependency specifications that include nesting of `|`? To see if 0install supports the more complex patterns.

Alizter:
Especially the interaction with `&` which I remember being error prone / misunderstood. (& should have higher priority over |). I guess this can be done in a later PR too.

Leonidas-from-XIV:
I've added both a package which nests `|` as well as one where we make sure that the precendence of `&` and `|` is correct.

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/mock-opam-repository/packages/priorities/priorities.0.0.1/opam
```diff
@@ -0,0 +1,4 @@
+opam-version: "2.0"
+depends: [
+    "bar" & "quux" | "baz"
```
rgrinberg, line 3: Is this equivalent to `depends: [ "bar" ("quux" | "baz") ]` if so, I'm not sure if it's testing anything additional.

It should be something like: `(foo & bar) | baz` so that it can't be simplified.

rgrinberg:
Also, this can saved for additional PR's, but we should test the not operator as well.

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/mock-opam-repository/packages/priorities/priorities.0.0.1/opam
Leonidas-from-XIV: Yes, just simple terms listed one after another automatically get an `&`.

Leonidas-from-XIV:
@rgrinberg I've added a test for negation (which only works on version formulas, since negation is only defined on verison formulas as per OPAM manual).

Turns out the dependency filter parser is oddly quiet on errors, if negation is used on packages `!"bar"` or the version is not quoted `{< 4}` it doesn't raise an error (or we're ignoring it). Anyway, that would be more of an issue to reproduce and report upstream.

rgrinberg (approved):
Thanks for the investigation. To re-iterate the point of this exercise, it's to make sure that we are able to accept all the constraints we allow users to write in dune-project files and vice-versa.

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
```diff
@@ -119,3 +149,103 @@ Run the solver again. This time it will fail.
...
+  > opam-version: "2.0"
+  > depends: [ "bar" | "baz" ]
+  > EOF
+
+After running this we expact a solution that has either `bar` or `baz` but not
+both.
+
+  $ dune pkg lock --opam-repository-path=mock-opam-repository
+  Solution for dune.lock:
+  bar.0.5.0
```
Alizter, line 172: Is it always preferred in order?

Leonidas-from-XIV: At least in my testing, this it has been 100% reproducible. Which is great, because we definitely want to be able to go from an exact input of opam files to have the exact same solution for reproducibility and even just easier debugging.

Alizter: Can we guarantee that this won't change in the future? I'm thinking that a `List.rev` might cause problems if added without thought by the opam devs. If it is not part of the opam spec should we let them know about it?

Leonidas-from-XIV: I think this is more of a theoretical problem than a practical one.

The OPAM devs don't have much to do with it as the solver is provided by 0install-solver and that is vendored, so it isn't gonna change under out feet. If the output changes in a future version of the solver, the then test needs to be adjusted (and possibly a note in a changelog added about the behavior changing) or a bug report needs to be filed if this change happened by accident.

I don't know if OPAM guarantees that in a `a | b` situation it will prefer `a` over `b` unless otherwise possible - I don't think so as it would probably make it harder to swap external solvers (cc @kit-ty-kate maybe you know more). I would expect that the only guarantee is it will return *some* valid solution if there is one.

But given that the test uses a very specific solver and it is not swappable, it seems to me that the thing it checks for is pretty accurate and is not going to cause flaky test runs.

Alizter: Thanks for the explanation, I agree that this isn't really a practical problem.

Alizter (approved):

PR #8364, opened 2023-08-10
refactor(engine): improve rule generation API

rgrinberg:
First, I've moved the Build_config interface to `Build_config_intf.Build_config`.

Then, I grouped all parts of the API that relate to rule generation to `Build_config.Gen_rules`.

I also made sure that every type has its own module where it's defined (to make it easy to include).

Finally, I added some additional constructor functions to simplify callers.


src/dune_rules/pkg_rules.ml
```diff
@@ -1370,7 +1370,7 @@ let gen_rules context_name (pkg : Pkg.t) =
     |> Action_builder.With_targets.add_directories
          ~directory_targets:[ pkg.paths.target_dir ]
   in
-  rule build_rule
+  rule ~loc:Loc.none (* TODO *) build_rule
```
Alizter, line 1373: Todo

dkalinichenko-js (approved):

rgrinberg:
@snowleopard @dkalinichenko-js if it's desired, I can also make the types in `Build_config.Gen_rules` abstract and introduce a `Build_config.Private` to access them. It's a little heavyweight, but it would really make the interface between the rules and the engine thoroughly controlled.

staronj (approved):

PR #8365, opened 2023-08-10
feature: start RPC server in every build

Alizter:
Testing this to see how it affects the bench.

****Conclusion****
Interesting, any potential overhead is firmly within noise range.


Alizter:
Something we might want to do in the future. Needs some more work however.

PR #8367, opened 2023-08-10
fix: use proper apostrophe in text

Alizter:
Noticed that this wasn't the correct character.


Leonidas-from-XIV:
Looks like someone's editor decided to use typographic apostrophes. Good catch, thanks!

PR #8368, opened 2023-08-10
feature: accept build requests in eagar watch

Alizter:
I'm not sure what good this restriction does so I am lifting it. This will let us use `dune rpc build` with an existing watch mode dune. In fact it might be useful to eventually get rid of the different watch modes and make a way to build no targets by default. (perhaps `dune build -w @empty`).


Alizter:
This causes some indefinite stalls so we should come back to this with a plan when we actually want to have this.

PR #6648, opened 2022-12-06
feature: Autogenerate man pages

Alizter:
We add a script to autogenerate man page rules which can be called using:
```
./dune.exe build @gen-man-rules
```

When these rules are present, `@check`, `@default` and `@doc` will all update the man pages if there are any changes. Currently we do not delete redundant files.

The motivation here is twofold:
1. We can have command line documentation present in the repo without the need for users to compile the executables.
2. We can observe better what the user comes across.

When all the files are out in the open, it is easy to see we have some nonsensical documentation. Some Dune commands are talking about options that don't apply to them etc.

cc @snowleopard who suggested something like this.

The patch for cmdliner: https://github.com/ocaml-dune/cmdliner/pull/2

Notes:
1. I had to modify cmdliner in order to crawl over the subcommands. cc @emillon does this patch look OK/maintainable? It will probably never make it upstream without good reason.
2. <s>In the future if we have generated includes we can skip the manual rule generation.</s> I've used directory targets instead now.
3. <s>I had to move things around so that the top Dune cmdliner command became visible in a library. Is this OK to do or should something else be done?</s> I have used the executables stanza and not exposed any new library now.

PR #8370, opened 2023-08-10
docs: fix mood and punctuation of flag doc strings

pmwhite:
The doc strings for various command options in `bin/common` were a bit inconsistently formatted and phrased, so I made them more consistent. See commit messages for more detail.

### DIFF
--- a/bin/common.ml
+++ b/bin/common.ml
@@ -196,12 +196,14 @@ module Options_implied_by_dash_p = struct
             [ "default-target" ]
             ~docs
             ~docv:"TARGET"
-            ~doc:"Set the default target that when none is specified to $(b,dune build).")
+            ~doc:
+              "Set the default target that is used when none is specified to $(b,dune \
+               build).")
     and+ always_show_command_line =
-      let doc = "Always show the full command lines of programs executed by dune" in
+      let doc = "Always show the full command lines of programs executed by dune." in
       Arg.(value & flag & info [ "always-show-command-line" ] ~docs ~doc)
     and+ promote_install_files =
-      let doc = "Promote the generated <package>.install files to the source tree" in
+      let doc = "Promote any generated <package>.install files to the source tree." in
       Arg.(
         last
         & opt_all ~vopt:true bool [ false ]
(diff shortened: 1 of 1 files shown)


Alizter:
The reason some of these were literals was due to the need for escaping of a few characters in cmdliner doc. See https://erratique.ch/software/cmdliner/doc/tool_man.html#doclang

I'll try to check the man pages by hand to see if they look ok.

bin/common.ml
```diff
@@ -938,22 +931,20 @@ module Builder = struct
...
             [ "error-reporting" ]
             ~doc:
-              "Controls when the build errors are reported.\n\
-               $(b,early) - report errors as soon as they are discovered.\n\
-               $(b,deterministic) - report errors at the end of the build in a \
-               deterministic order.\n\
-               $(b,twice) - report each error twice: once as soon as the error is \
-               discovered and then again at the end of the build, in a deterministic \
-               order.")
+              "Controls when the build errors are reported. $(b,early) - report errors \
```
Alizter, line 934: This is meant to be a list of bullets.

pmwhite: But in fact it is not formatted as a list of bullets in `dune build --watch`. I think it is better to let the string be auto-formatted instead of giving the impression that we have the ability to do formatting ourselves.

Alizter: In that case, let's remove the `-` and just list the values in sentences.

pmwhite:
> The reason some of these were literals was due to the need for escaping of a few characters in cmdliner doc.

Good point, I didn't think of that; I just assumed it was to allow manual formatting of the string. I still think the change is a good one, but I will look it over again and fix up any places where characters need to be escaped.

Alizter:
@pmwhite FTR I was working on a PR earlier this year that could help with such a change. https://github.com/ocaml/dune/pull/6648

The idea is that we autogenerate the man pages by crawling over the cmdliner commands. This would let us perform formatting / wording changes in command descriptions easily and serve as a base of documentation. I didn't continue with it because I wanted to make a promotable directory target diff for the man page outputs which I never got round to implementing.

pmwhite:
Oh, nice; we should definitely do that. As I was writing this PR, I was wishing I could also see an auto-generated diff of how the help text would change.

bin/common.ml
pmwhite: Done.

pmwhite:
Okay, I scanned over all the changes to see if anything needs to be escaped; I didn't see anything.

PR #8372, opened 2023-08-11
Use `%{pkg:...}` macro for package vars

gridbugs:
This removes the `Name` and `Version` pkg pform variables as these variables are always package scoped (possibly implicitly). This plus the fact that opam allows custom package variables means we can get rid of handling unknown variables when converting commands. It also makes it trivial to support explicitly package-scoped variables when converting commands so that's included in this change too.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -230,61 +230,35 @@ module Summary = struct
   ;;
 end
 
-let opam_command_to_string_debug (args, _filter_opt) =
-  List.map args ~f:(fun (simple_arg, _filter_opt) ->
-    match simple_arg with
-    | OpamTypes.CString s -> String.quoted s
-    | CIdent ident -> ident)
-  |> String.concat ~sep:" "
-;;
-
-let opam_commands_to_actions package (commands : OpamTypes.command list) =
-  let pform_of_ident_opt ident =
-    let `Self, variable =
-      match String.split ident ~on:':' with
-      | [ variable ] | [ "_"; variable ] -> `Self, variable
-      | _ ->
-        (* TODO *)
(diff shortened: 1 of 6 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -230,61 +230,32 @@ module Summary = struct
...
+      | None -> `Self_or_global, ident
+      | Some ("_", variable) -> `Self, variable
+      | Some (package, variable) -> `Package package, variable
     in
-    match Pform.Var.Pkg.of_opam_variable_name_opt variable with
-    | Some pkg_var -> Ok Pform.(Var (Var.Pkg pkg_var))
-    | None -> Error (`Unknown_variable variable)
+    match scope with
+    | `Package package ->
+      (* TODO: abstract the war package variables are encoded into multi-argument macros *)
```
rgrinberg, line 243: war?

rgrinberg (approved):
Kind of a shame to make all package build instructions uglier. But it does seem to simplify the code.

gridbugs:
> Kind of a shame to make all package build instructions uglier.

I plan to improve ergonomics of the `%{pkg:...}` macro in a future change.

Leonidas-from-XIV (commented):
So do I understand correctly that we will just write unknown (at lock time) variables into the lockfiles and then defer the evaluation to a later stage in the build?

Sounds ok to me.

rgrinberg:
This needs a rebase.

gridbugs:
Rebased!

src/dune_pkg/opam_solver.ml
gridbugs: should have been "var". Fixed.

PR #8374, opened 2023-08-11
refactor: change [Loc.t option] to [Loc.t]

rgrinberg:
We change [Anonymous_action.t] to be consistent with [Rule.t]

### DIFF
--- a/src/dune_engine/build_system.ml
+++ b/src/dune_engine/build_system.ml
@@ -351,7 +351,7 @@ end = struct
 
   (* The current version of the rule digest scheme. We should increment it when
      making any changes to the scheme, to avoid collisions. *)
-  let rule_digest_version = 14
+  let rule_digest_version = 15
 
   let compute_rule_digest
     (rule : Rule.t)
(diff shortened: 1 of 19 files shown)


dkalinichenko-js (approved):

src/dune_engine/build_system.ml
```diff
@@ -1164,7 +1160,8 @@ end = struct
             (Memo.Stack_frame.as_instance_of
                frame
                ~of_:execute_action_generic_stage2_memo)
-            ~f:(fun (x : Anonymous_action.t) -> x.action.loc)))
+            ~f:(fun (x : Anonymous_action.t) ->
+              Option.some_if (not @@ Loc.is_none x.action.loc) x.action.loc)))
```
Alizter, line 1164: We have Loc.is_some

Alizter, line 1164: This part is a little confusing. So we are turning a Loc back into a Loc option here? Why do we need to have an option type at all?

PR #8376, opened 2023-08-11
refactor: remove specialized aliases from engine

rgrinberg:

### DIFF
--- a/src/dune_rules/alias0.ml
+++ b/src/dune_rules/alias0.ml
@@ -1,0 +1,16 @@
+open Import
+
+let standard name =
+  let name = Alias.Name.of_string name in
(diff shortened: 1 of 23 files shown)


src/dune_rules/alias0.ml
```diff
@@ -0,0 +1,16 @@
+open Import
```
Alizter, line 1: You could include these in Import.Alias which would save having two potentially confusing modules. If it only gets used by rules, I don't see why not.

Alizter: i.e. inline this entire file into Import and then you can do `Alias.(make check)` etc.

dkalinichenko-js (approved):

PR #8377, opened 2023-08-11
Upstream Dune hooks API changes

dkalinichenko-js:
Generalize Dune hooks API over a type constructor. Introduce build start/end hooks that execute in the `Action_builder` monad. Rename pre-existing build end hooks to `Post_build`.


dkalinichenko-js:
Re-open if needed.

PR #8378, opened 2023-08-11
Upstream other Jane Street changes to Dune engine

dkalinichenko-js:


src/dune_output_truncation/jbuild
Alizter, line 1: Should this be a dune file?

dkalinichenko-js: Thanks, fixed.

Alizter:
@dkalinichenko-js Something went wrong with the rebase.

PR #8384, opened 2023-08-13
refactor(pkg): use [Dune_sexp] when possible

rgrinberg:


Alizter:
What is the benefit of reexporting Dune_sexp in Dune_lang?

rgrinberg:
No benefit. Just a temporary state of affairs to avoid transitioning everything all at once.

PR #8388, opened 2023-08-13
tests: Conv digest

pmwhite:
This is inspired by something that https://github.com/janestreet/bin_prot does; the idea is to ensure we don't accidentally break old versions of an RPC by cementing in tests some artifact of the serialization format. While this PR has some non-test changes, they should just be refactors and shouldn't change any behavior at all.

I personally think this PR is a good idea, but I will not be offended if others disagree and think it should not be merged at all.

### DIFF
--- a/otherlibs/dune-rpc/private/conv.ml
+++ b/otherlibs/dune-rpc/private/conv.ml
@@ -119,6 +119,11 @@ type version =
   }
 
 type ('a, 'kind) t =
+  | String : (string, values) t
+  | Int : (int, values) t
+  | Float : (float, values) t
+  | Unit : (unit, values) t
+  | Char : (char, values) t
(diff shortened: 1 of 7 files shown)


rgrinberg:
LGTM. I see no harm in merging this as it's giving us a little extra safety for free.

The whole conv module is something I would like to pull into its own library one day by the way.

pmwhite:
> The whole conv module is something I would like to pull into its own library one day by the way.

Seems reasonable, but out of curiosity, how come? Just because it's more generic than just an RPC serialization mechanism? Or is there something annoying that would be solved by having it separate?

rgrinberg:
It might pay off in a couple of other contexts where we just want serialization and deserialization automatically and aren't too picky about encoding things in a special way. In particular, I'm thinking of:

* Encoding and decoding dune-package files
* Storing/restoring metadata from the dune cache

Rudi.

PR #8389, opened 2023-08-14
fix: position in error for invalid char in pform

gridbugs:
Previously the error message would refer to the incorrect part of the source when reporting errors about invalid characters in pform templates, such as this error which refers to the invalid space character:
```
4 |  (action (echo %{read:foo bar.txt})))
                     ^
Error: This character is not allowed inside %{...} forms
```

This change corrects the position of the arrow indicating the invalid character and also adds the character to the error message itself:
```
4 |  (action (echo %{read:foo bar.txt})))
                             ^
Error: The character ' ' is not allowed inside %{...} forms
```

### DIFF
--- a/src/dune_sexp/lexer.mll
+++ b/src/dune_sexp/lexer.mll
@@ -352,7 +352,14 @@ and template_variable = parse
   }
   | '}' | eof
     { error lexbuf "%{...} forms cannot be empty" }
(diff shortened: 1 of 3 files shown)


Alizter (approved):
Nice!

Leonidas-from-XIV (approved):
Very nice quality of life improvement.

PR #8390, opened 2023-08-14
refactor: Unify `make` variable in pkg and non-pkg contexts

gridbugs:


src/dune_lang/pform.mli
```diff
@@ -90,6 +88,10 @@ module Var : sig
 
   val compare : t -> t -> Ordering.t
   val to_dyn : t -> Dyn.t
+
+  (** Translate an global variable from an opam file into the equivalent dune
```
Alizter, line 92:
```suggestion
  (** Translate n global variable from an opam file into the equivalent dune
```

Alizter (approved):

Leonidas-from-XIV (approved):

PR #8391, opened 2023-08-14
Variable interpolations in opam commands

gridbugs:
Adds support for replacing opam variable interpolations with dune pforms while converting opam build/install commands to dune actions.

### DIFF
--- a/src/dune_lang/string_with_vars.mli
+++ b/src/dune_lang/string_with_vars.mli
@@ -39,6 +39,10 @@ val virt_pform : ?quoted:bool -> string * int * int * int -> Pform.t -> t
 val virt_text : string * int * int * int -> string -> t
 val make_pform : ?quoted:bool -> Loc.t -> Pform.t -> t
 val make_text : ?quoted:bool -> Loc.t -> string -> t
+
+(** Concatenate a list of parts. *)
+val make : ?quoted:bool -> Loc.t -> [ `Text of string | `Pform of Pform.t ] list -> t
+
 val is_pform : t -> Pform.t -> bool
 val has_pforms : t -> bool
 
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -230,7 +230,15 @@ module Summary = struct
   ;;
 end
 
-let opam_commands_to_actions (commands : OpamTypes.command list) =
+let opam_command_to_string_debug (args, _filter_opt) =
+  List.map args ~f:(fun (simple_arg, _filter_opt) ->
+    match simple_arg with
+    | OpamTypes.CString s -> String.quoted s
+    | CIdent ident -> ident)
+  |> String.concat ~sep:" "
+;;
+
+let opam_commands_to_actions package (commands : OpamTypes.command list) =
   let pform_of_ident ident =
     let scope, variable =
       match String.lsplit2 ident ~on:':' with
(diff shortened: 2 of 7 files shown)


vendor/opam/src/format/opamFilter.mli
```diff
@@ -50,6 +51,10 @@ val fold_down_left: ('a -> filter -> 'a) -> 'a -> filter -> 'a
 (** Maps on all nodes of a filter, bottom-up *)
 val map_up: (filter -> filter) -> filter -> filter
 
+(** Regex matching string interpolation syntax (["%%"], ["%{xxx}%"], or
+    ["%{xxx"] if unclosed) *)
+val string_interp_regex : Re.re
```
Leonidas-from-XIV, line 56: Could you create a PR to OPAM to upstream this if possible?

src/dune_lang/string_with_vars.mli
```diff
@@ -39,6 +39,16 @@ val virt_pform : ?quoted:bool -> string * int * int * int -> Pform.t -> t
 val virt_text : string * int * int * int -> string -> t
 val make_pform : ?quoted:bool -> Loc.t -> Pform.t -> t
 val make_text : ?quoted:bool -> Loc.t -> string -> t
+
+(** Concatenate a list of parts separate by [sep] which is the empty string by
```
Leonidas-from-XIV, line 43:
```suggestion
(** Concatenate a list of parts separated by [sep] which is the empty string by
```

test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
```diff
@@ -36,26 +36,64 @@ Generate a mock opam repository
...
+  >     "./configure"
+  >     "--prefix=%{prefix}%"
+  >     "--docdir=%{doc}%/ocaml"
+  >   ]
+  >   [make "-j%{jobs}%"]
+  > ]
+  > install: [make "install"]
+  > EOF
+
+Make sure we don't mess up percent signs that arn't part of variable interpolation syntax:
```
Leonidas-from-XIV, line 52:
```suggestion
Make sure we don't mess up percent signs that aren't part of variable interpolation syntax:
```

src/dune_lang/string_with_vars.mli
```diff
@@ -39,6 +39,16 @@ val virt_pform : ?quoted:bool -> string * int * int * int -> Pform.t -> t
...
+    default. *)
+val make
+  :  ?sep:string
+  -> ?quoted:bool
+  -> Loc.t
+  -> [ `Text of string | `Pform of Pform.t ] list
```
Leonidas-from-XIV, line 49: I find this API a bit overloaded, as it mixes together concatenation of strings with implicit separators and creation of `String_with_vars.t` from a list of different things. I would have expected this to be multiple functions, like `make_text`, `make_pform` and a concatenation function (with possibly an explicit separator) to create a concatenated `String_with_vars.t`.

src/dune_pkg/opam_solver.ml
```diff
@@ -256,35 +256,60 @@ let opam_commands_to_actions package (commands : OpamTypes.command list) =
...
+                   "Encountered malformed variable interpolation while processing \
+                    commands for package %s."
                    (OpamPackage.to_string package)
+               ; Pp.text "The variable interpolation:"
+               ; Pp.text other
                ; Pp.text "The full command:"
                ; Pp.text (opam_command_to_string_debug command)
                ]))
+      |> List.of_seq
+      |> String_with_vars.make Loc.none
```
Leonidas-from-XIV, line 306: This creates the `String_with_vars.t` of out the split result and implicitly adds `" "` as a separator, but what I don't understand why this would be necessary — I don't see where the spaces are getting consumed so they would need to be added. Can you clarify?

gridbugs: The default separator is `""`, not `" "`. I'll look into simplifying the api as per your other suggestions and also make the separator explicit to clarify cases like this.

vendor/opam/src/format/opamFilter.mli
gridbugs: https://github.com/ocaml/opam/pull/5633

src/dune_lang/string_with_vars.mli
gridbugs: Concatenating a pair of `String_with_vars.t`s is tricky because in addition to being a list of parts (strings and pforms) they also consist of a `Loc.t` (which could just be combined with `Loc.span`) but also a whether the string is quoted which is less clear how to combine. The full type is:
```ocaml
type t =
  { quoted : bool
  ; parts : part list
  ; loc : Loc.t
  }
```
If we were to make a generic concatenation function for `String_with_vars.t` we'd need to come up with a policy for how to merge the `quoted` field so it seemed simpler to let users construct fresh values out of parts.

The `sep` argument is left over from a previous use case of this function that I ended up not using so I'll get rid of it.

src/dune_pkg/opam_solver.ml
gridbugs: Actually turns out we don't need the separator anymore so I'll remove it altogether.

src/dune_pkg/opam_solver.ml
```diff
@@ -251,13 +259,37 @@ let opam_commands_to_actions (commands : OpamTypes.command list) =
...
+             let ident = String.sub ~pos:2 ~len:(String.length interp - 4) interp in
+             `Pform (pform_of_ident ident)
+           | other ->
+             User_error.raise
+               [ Pp.textf
+                   "Encountered malformed variable interpolation while processing \
+                    commands for package %s."
+                   (OpamPackage.to_string package)
+               ; Pp.text "The variable interpolation:"
+               ; Pp.text other
```
Alizter, line 282: Rather than printing the string, could you extract the full location of the misformed string and use `Loc.pp`? That will print an excerpt with an underline.

Leonidas-from-XIV (approved):

src/dune_pkg/opam_solver.ml
```diff
@@ -251,13 +259,37 @@ let opam_commands_to_actions (commands : OpamTypes.command list) =
...
+  List.filter_map commands ~f:(fun ((args, _filter_opt) as command) ->
+    let interpolate_opam_variables s =
+      Dune_re.Seq.split_full OpamFilter.string_interp_regex s
+      |> Seq.map ~f:(function
+        | `Text text -> `Text text
+        | `Delim group ->
+          (match Dune_re.Group.get group 0 with
+           | "%%" -> `Text "%"
+           | interp
+             when String.starts_with ~prefix:"%{" interp
```
rgrinberg, line 271: We have `is_prefix` and `is_suffix` to support older versions of OCaml.

src/dune_pkg/opam_solver.ml
gridbugs: The problem with that would be that the commands are read from opam files using `OpamFile.OPAM.read` which doesn't tie the metadata fields back to locations in the file and the opam repo code inside dune doesn't give access to raw opam files. We could extend it to give access to raw files and use regexes to find the part of the file that contains the command but I'd rather do that in a separate feature if it's something we want to support.

PR #8392, opened 2023-08-14
try switching stop_on_first_error on to see if cov is fixed

Alizter:
Coverage is broken by #8362, this PR attempts to investigate why.


Alizter:
OK so it seems that the stop_on_first_error option is working correctly. I still don't understand why we are suddenly running less tests in coverage after the JS change...

PR #8399, opened 2023-08-15
Concise pkg macro

gridbugs:
This change adds a small library for encoding and decoding package variables as pforms so we can change the way they are formatted without needing to touch multiple parts of the code. It then changes the format of the `pkg` macro to be more concise in an attempt to make it more ergonomic to use.

### DIFF
--- a/src/dune_pkg/package_variable.ml
+++ b/src/dune_pkg/package_variable.ml
@@ -22,16 +22,6 @@ module Scope = struct
   type t =
     | Self
     | Package of Package_name.t
-
-  let of_string = function
-    | "_" -> Self
-    | package -> Package (Package_name.of_string package)
-  ;;
-
-  let to_string = function
-    | Self -> "_"
-    | Package package_name -> Package_name.to_string package_name
-  ;;
 end
 
 type t =
@@ -41,23 +31,36 @@ type t =
 
 let self_scoped name = { name; scope = Self }
 let package_scoped name package_name = { name; scope = Package package_name }
-let var_tag = "var"
 
-let to_payload { name; scope } =
-  Pform.Payload.of_args [ var_tag; Scope.to_string scope; Name.to_string name ]
-;;
-
-let of_macro_invocation ({ Pform.Macro_invocation.macro; _ } as macro_invocation) =
+let of_macro_invocation ~loc ({ Pform.Macro_invocation.macro; _ } as macro_invocation) =
   match macro with
+  | Pkg_self ->
+    let variable_name = Pform.Macro_invocation.Args.whole macro_invocation in
+    Ok (self_scoped (Name.of_string variable_name))
   | Pkg ->
-    (match Pform.Macro_invocation.Args.split macro_invocation with
-     | [ tag; scope_arg; variable_name ] when String.equal tag var_tag ->
-       Some { name = Name.of_string variable_name; scope = Scope.of_string scope_arg }
-     | _ -> None)
-  | _ -> None
+    let package_name, variable_name =
+      Pform.Macro_invocation.Args.lsplit2_exn macro_invocation loc
+    in
+    Ok
+      (package_scoped
+         (Name.of_string variable_name)
+         (Package_name.of_string package_name))
+  | _ -> Error `Unexpected_macro
 ;;
 
-let to_macro_invocation t = { Pform.Macro_invocation.macro = Pkg; payload = to_payload t }
+let to_macro_invocation { name; scope } =
+  match scope with
+  | Self ->
+    { Pform.Macro_invocation.macro = Pkg_self
+    ; payload = Pform.Payload.of_args [ Name.to_string name ]
+    }
+  | Package package_name ->
+    { Pform.Macro_invocation.macro = Pkg
+    ; payload =
+        Pform.Payload.of_args [ Name.to_string name; Package_name.to_string package_name ]
+    }
+;;
+
 let to_pform t = Pform.Macro (to_macro_invocation t)
 
 let of_opam_ident ident =
--- a/src/dune_pkg/package_variable.ml
+++ b/src/dune_pkg/package_variable.ml
@@ -1,0 +1,79 @@
+open Stdune
+open Dune_lang
+
+module Name = struct
+  include String
+
+  include (
+    Dune_util.Stringlike.Make (struct
+      type t = string
+
+      let to_string x = x
+      let module_ = "Package_variable.Name"
+      let description = "package variable name"
+      let description_of_valid_string = None
+      let hint_valid = None
+      let of_string_opt s = if s = "" then None else Some s
+    end) :
+      Dune_util.Stringlike with type t := t)
+end
+
+module Scope = struct
+  type t =
+    | Self
+    | Package of Package_name.t
+
+  let of_string = function
+    | "_" -> Self
+    | package -> Package (Package_name.of_string package)
+  ;;
+
+  let to_string = function
+    | Self -> "_"
+    | Package package_name -> Package_name.to_string package_name
+  ;;
+end
+
+type t =
+  { name : Name.t
+  ; scope : Scope.t
+  }
+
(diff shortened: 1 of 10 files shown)


src/dune_pkg/package_variable.ml
```diff
@@ -0,0 +1,77 @@
...
+      let hint_valid = None
+      let of_string_opt s = if s = "" then None else Some s
+    end) :
+      Dune_util.Stringlike with type t := t)
+end
+
+module Scope = struct
+  type t =
+    | Self
+    | Package of Package_name.t
```
Leonidas-from-XIV, line 24: FTR OPAM also knows of `Global`. I assume we can ignore those since this is only for package variables.

But on the other hand, should we have a case for `Self`? It is a more specific case of `Package` with the package name matching its own package. Maybe these should be normalized to regular package variables (that is, fill in `_` with the package name, as it is done already in the `macro` expansion, but earlier).

What do you think?

test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
```diff
@@ -36,17 +36,34 @@ Generate a mock opam repository
...
 
   $ cat dune.lock/standard-dune.pkg
   (version 0.0.1)
   (install (run %{make} install))
-  (build (progn (run dune subst) (run dune build -p %{pkg:var:_:name} -j %{jobs} @install @runtest @doc)))
+  (build (progn (run dune subst) (run dune build -p %{pkg:name} -j %{jobs} @install @runtest @doc)))
+
+  $ cat dune.lock/variable-types.pkg
+  (version 0.0.1)
+  (build (progn (run echo %{pkg:local_var}) (run echo %{pkg:explicit_local_var}) (run echo %{pkg:foo:package_var}) (run echo %{os_family})))
```
rgrinberg, line 69: This is certainly an improvement as the `_` convention doesn't really fit with the rest of dune, but I think it's not a simple design. To see what I mean, try to expand these percent forms into the equivalent s-expressions. You would need something like overloading based on the number of arguments to express this and I don't think that's something we'd ever want to support.

To keep things simpler, I propose we do one of:

1. Pull package local variables into the current namespace as it was before. E.g. this would make it quite similar to our existing `%{deps}` or `%{targets}` vars.
2. Introduce a different form for "self" variables. Something like `%{pkg-self:..}`.

Another approach would be to just make `%{name}` first class, and add support for nesting e.g. `%{pkg:%{name}:...}`. I think this is "simple" (using your definition), but rather ugly.

src/dune_pkg/package_variable.ml
rgrinberg: > Maybe these should be normalized to regular package variables (that is, fill in _ with the package name, as it is done already in the macro expansion, but earlier).

That would create unnecessary noise when renaming a package.

Leonidas-from-XIV: How so? If a package is renamed, we generate a new lock file with a new name anyway and the names in these files are generated by us in any case.

rgrinberg: Without this PR, git would figure out that all you've done is move `x.pkg` to `y.pkg`. With this PR, regenerating the lock file would now replace every instance of `x` by `y` inside `y.pkg` which would trick git that this file was written from scratch.

Leonidas-from-XIV: Fair enough, however how often do renames happen? I can't remember any package being renamed, because it breaks all its dependencies; the only case of that happening is when OUnit2 renamed itself using an intermediate package.

Alizter: Didn't the lock free data structures library get recently renamed to saturn? https://github.com/ocaml-multicore/saturn

It's not very common for packages to be renamed but also not unheard of.

Leonidas-from-XIV: I think `kcas` was joined with a few other packages into `saturn`, not a straight-up rename. However, I am not very involved what happened, so I might have gotten it wrong.

My actual point is that maybe the change is still worth it for a simpler semantic in Dune by just having one kind of package variables instead of optimizing for renames of packages.

rgrinberg: I do think there are solutions that are about as simple that don't interfere with renaming. In my other comment, I mentioned introducing a different form for self packages for example.

test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
gridbugs: I'm in favour of option 2 as it means the `%{foo}` syntax can remain just for a fixed set of pform variables rather than arbitrarily-named package variables. So we'll have the `%{pkg:<package>:<variable>}` macro for variables from explicitly named packages and a new `%{pkg-self:<variable>}` macro for variables from the current package.

gridbugs:
Updated to use `%{pkg-self:<var>}` syntax for local variables.

src/dune_pkg/package_variable.mli
```diff
@@ -0,0 +1,30 @@
...
+end
+
+type t =
+  { name : Name.t
+  ; scope : Scope.t
+  }
+
+val of_macro_invocation : loc:Loc.t -> Pform.Macro_invocation.t -> t option
+
+(** Parse an opam variable name. Identifiers begining with "<package>:" are
```
Leonidas-from-XIV, line 24:
```suggestion
(** Parse an opam variable name. Identifiers beginning with "<package>:" are
```

src/dune_rules/pkg_rules.ml
```diff
@@ -508,35 +508,43 @@ module Action_expander = struct
...
-                     | Some section ->
-                       let section = dune_section_of_pform section in
-                       let install_paths = Paths.install_paths paths in
-                       Memo.return [ Value.Dir (Install.Paths.get install_paths section) ]))))
-         | _ -> assert false)
+      | Macro ({ macro = Pkg | Pkg_self; _ } as macro_invocation) ->
+        let { Package_variable.name = variable_name; scope } =
+          match Package_variable.of_macro_invocation ~loc macro_invocation with
+          | Some package_variable -> package_variable
+          | None -> assert false
```
Leonidas-from-XIV, line 515: Maybe add a comment that it is a programming error to call `Package_variable.of_macro_invocation` on something that's not a `Pkg`/`Pkg_self` macro. I needed to go to the code to check why it can return `None`, but YMMV :)

src/dune_rules/pkg_rules.ml
```diff
@@ -508,35 +508,43 @@ module Action_expander = struct
...
+        let present = Option.is_some paths in
+        let variable_name = Package_variable.Name.to_string variable_name in
+        (match String.Map.find variables variable_name with
+         | Some v -> Memo.return @@ Variable.dune_value v
+         | None ->
+           (match variable_name with
+            | "pinned" -> Memo.return [ Value.String "false" ]
+            | "enable" ->
+              Memo.return [ Value.String (if present then "enable" else "disable") ]
+            | "installed" -> Memo.return [ Value.String (Bool.to_string present) ]
```
Leonidas-from-XIV, line 536: Maybe computing `present` just in this branch? It is not used anywhere else as far as I can see and I was confused why it was computed.

src/dune_rules/pkg_rules.ml
```diff
@@ -508,35 +508,43 @@ module Action_expander = struct
...
+            | _ ->
+              (match paths with
+               | None -> assert false
```
Leonidas-from-XIV, line 539: Shouldn't this have an `User_error.raise` here as well? It seems like a variable of a package that is not installed - probably something a user can do, if they make a typo in the rules.

Leonidas-from-XIV (commented):
I think it looks generally rather good, I think `%{pkg:}` and `%{pkg-self}` are a quite decent solution. Some smaller things I noted while reading the source code attached.

PR #8400, opened 2023-08-15
feature: add --stop-on-first-error CLI option

Alizter:
We expose a CLI option called --stop-on-first-error that exposes to the user the internal behaviour of stopping the build on the first encountered error.

- [x] option documentation
- [x] changelog
- [x] test


doc/changes/8400.md
```diff
@@ -0,0 +1,2 @@
+- Add `--stop-on-first-error` option to `dune build` which will terminate the build when
+  the first error message is encountered. (#8400, @Alizter)
```
rgrinberg, line 2: Let's mention the name of whoever implemented the feature in the engine as well (pmwhite I believe)

Alizter: Yes, in fact I will probably remove my name completely if @pmwhite is happy with me writing that.

rgrinberg: You did some work too, so keep it.

Alizter:
@rgrinberg Do we want this configurable from dune config? I could add an option, but we don't have any precedent for a boolean option.

rgrinberg:
I don't see the need. The feature seems very niche apart from some integrations with CI

test/blackbox-tests/test-cases/stop-on-first-error.t/run.t
```diff
@@ -0,0 +1,19 @@
+Here we demonstrate the --stop-on-first-error behaviour.
```
rgrinberg, line 1: Can you change the test not to rely on OCaml? Simple rules and actions will do here.

MisterDA:
Related issue on opam: https://github.com/ocaml/opam/issues/4629. Would be nice if both tools could use the same name for the switch. IMO Dune is a bit verbose.

PR #8408, opened 2023-08-16
feature: show number of errors when waiting in watch mode

Alizter:
This is a feature that I originally added to Dune monitor by keeping track of the cardinality of the diagnostics sent. We can do better in general by threading the correct information through the build system.

This displays:
```
Had 4 errors
```
When the watchmode build has been finished with 4 errors. For a single error it say "Had 1 error".

- [ ] changelog


src/dune_engine/build_system.ml
```diff
@@ -1279,7 +1281,7 @@ let run f =
       let final_status =
         if List.exists exns ~f:caused_by_cancellation
         then State.Restarting_current_build
-        else Build_failed__now_waiting_for_changes
+        else Build_failed__now_waiting_for_changes (List.length exns)
```
Alizter, line 1284: I'm not sure this is the correct thing to count, but it seems to work.

Alizter:
cc @pmwhite this may be useful for you.

Alizter:
@rgrinberg Because it makes the information available more widely than just through RPC. Before if I wanted to know how many jobs had failed when updating the status line in `bin/import.ml` there wasn't anything to use. Now we have a dedicated way of getting the value, which we can use in monitor if we want to or count like I was already doing.

The point is not that it makes life better for dune monitor but rather that it was inspired by dune monitor. I think having this behaviour is useful for users.

rgrinberg:
If this information is useful (I'm skeptical of that), it can already be derived from the existing machinery. Just get it from `Build_system.errors`.

Alizter:
@rgrinberg Thanks, I've dropped the other changes to the engine now and am able to get it directly from `Build_system.errors`.

rgrinberg:
Please revert all the changes to RPC. The number of errors can already be tracked by subscribing to them.

Alizter:
@rgrinberg All RPC changes reverted.

pmwhite:
As another data point to weigh, I think I would personally enjoy seeing the current error count. For instance, if I change a type that produces a to-do list of, say, 30 type errors, it'd be nice to see how that number ticks down as I fix each error.

Alizter:
Another use case that I had in mind was for when people run tests. From experience porting large test suites to Dune, a commonly asked for feature was a count for how many tests failed. Although that might not be something immediately useful to everybody I think it can be useful to some. Note that VS Code users already have a similar feature by inspecting their "problems" tab.

snowleopard (approved):
Looks good to me, though maybe we should add a test to see the "Had 2 errors" message as well?

Alizter:
@snowleopard done.

snowleopard:
> @snowleopard done.

Thanks! It's not clear to me if the test you added is deterministic though (maybe we want to add some sorting or filtering).

test/blackbox-tests/test-cases/watching/multiple-errors.t
```diff
@@ -0,0 +1,39 @@
+We test the behavior of watch mode when we have multiple errors
+
+  $ . ./helpers.sh
+
+  $ export PATH=.:$PATH
```
rgrinberg, line 5: Why do you modify the PATH?

Alizter: That was copied from a test, I'll remove it.

Alizter:
@snowleopard I think the test is deterministic, more rather I don't see why it wouldn't be.

snowleopard:
> @snowleopard I think the test is deterministic, more rather I don't see why it wouldn't be.

Because one error is racing with another?

rgrinberg:
Note that we rely on the order of error messages in many different places in our test suite. This PR isn't making it worse.

PR #8409, opened 2023-08-16
fix: executing processes

rgrinberg:
* in dune exec, the running process is already being killed by the
  polling loop. There's no need to do it twice.

* the action runner process shouldn't be restarted when the input
  changes


Alizter (approved):

rgrinberg:
Adding @gridbugs as a reviewer as well given that this is relevant to `dune exec`.

gridbugs (approved):

dkalinichenko-js (approved):

PR #8410, opened 2023-08-16
test: repro rpc jobs bug

Alizter:
This demonstrates that RPC jobs events are not stopped when dune is interrupted and restarted.

cc @rgrinberg 

I can move this test somewhere else if needed.

I can also confirm that this is fixed by #8409

### DIFF
--- a/test/expect-tests/dune_rpc_e2e/dune_rpc_jobs.ml
+++ b/test/expect-tests/dune_rpc_e2e/dune_rpc_jobs.ml
@@ -1,0 +1,84 @@
+open Stdune
+open Fiber.O
+open Dune_rpc_e2e
+module Client = Dune_rpc_client.Client
+module Dune_rpc = Dune_rpc_private
+
+include struct
+  open Dune_rpc
+  module Job = Job
+  module Conv = Conv
+end
+
+let files = List.iter ~f:(fun (f, contents) -> Io.String_path.write_file f contents)
+
+let poll_exn client decl =
+  let+ poll = Client.poll client decl in
+  match poll with
+  | Ok p -> p
+  | Error e -> raise (Dune_rpc.Version_error.E e)
+;;
+
+let print_job_events poll =
+  let+ res = Client.Stream.next poll in
+  let id_to_string id = Conv.to_sexp Job.Id.sexp id |> Sexp.to_string in
+  match res with
+  | None -> printfn "client: no more diagnostics"
+  | Some job_event_list ->
+    List.iter job_event_list ~f:(fun job_event ->
+      match (job_event : Job.Event.t) with
+      | Start job ->
+        printfn
+          "Start %s %s"
+          (id_to_string job.id)
+          (Format.asprintf "%a" Pp.to_fmt job.description)
+      | Stop id -> printfn "Stop %s" (id_to_string id))
+;;
+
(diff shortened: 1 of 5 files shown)


otherlibs/dune-rpc/private/exported_types.mli
```diff
@@ -138,6 +138,7 @@ module Job : sig
     val hash : t -> int
     val create : int -> t
     val sexp : (t, Conv.values) Conv.t
+    val to_string : t -> string
```
rgrinberg, line 141: Could you use the sexp converter insead of adding this function?

Alizter: Done.

test/expect-tests/dune_rpc_e2e/dune_rpc_jobs.ml
```diff
@@ -0,0 +1,84 @@
...
+                  (progn
+                   (write-file running "hello")
+                   (system "sleep 100")
+                   (write-file foo "foo"))))
+                |}
+               )
+             ]
+      in
+      let* poll = poll_exn client Dune_rpc.Public.Sub.running_jobs in
+      let* () = Fiber.return @@ files [ "bar", "" ] in
```
rgrinberg, line 69: what's the point of the `let* () = Fiber.return @@ ..`? Can't it just be `files [ "bar", "" ]`

test/expect-tests/dune_rpc_e2e/dune_rpc_jobs.ml
```diff
@@ -0,0 +1,84 @@
...
+          "Start %s %s"
+          (id_to_string job.id)
+          (Format.asprintf "%a" Pp.to_fmt job.description)
+      | Stop id -> printfn "Stop %s" (id_to_string id))
+;;
+
+let%expect_test "rpc jobs after rebuild" =
+  let rec wait_for_running_file () =
+    match Unix.stat "_build/default/running" with
+    | _stat -> printfn "Background process is running, lets interupt..."
```
rgrinberg, line 41: `let's interrupt`.

test/expect-tests/dune_rpc_e2e/dune_rpc_jobs.ml
Alizter: I wasn't sure about the order of evaluation so I was trying to be sure it would be sequential.

test/expect-tests/dune_rpc_e2e/dune_rpc_jobs.ml
```diff
@@ -0,0 +1,84 @@
...
+          "Start %s %s"
+          (id_to_string job.id)
+          (Format.asprintf "%a" Pp.to_fmt job.description)
+      | Stop id -> printfn "Stop %s" (id_to_string id))
+;;
+
+let%expect_test "rpc jobs after rebuild" =
+  let rec wait_for_running_file () =
+    match Unix.stat "_build/default/running" with
+    | _stat -> printfn "Background process is running, let's interupt it..."
```
rgrinberg, line 41: Still missing an `r` in `interrupt`.

Alizter: Sorry, fixed.

PR #8411, opened 2023-08-17
test: serialize and deserialize RPC error messages

Alizter:
We demonstrate a bug with how error messages are serialized. Currently there are two "Error:" prefixes after serializing and deserializing. This is because the prefix hasn't been stripped in the initial serialization. We should strip the prefix since the severity information is recorded elsewhere in the diagnostic data.

### DIFF
--- a/otherlibs/stdune/src/stdune.ml
+++ b/otherlibs/stdune/src/stdune.ml
@@ -30,6 +30,32 @@ module Pp = struct
   let compare ~compare x y =
     Ordering.of_int (Pp.compare (fun a b -> Ordering.to_int (compare a b)) x y)
   ;;
+
+  let to_dyn tag_to_dyn t =
+    match Pp.to_ast t with
+    | Error _ -> Dyn.variant "Contains Format" [ Dyn.opaque "<error>" ]
+    | Ok t ->
+      let open Dyn in
+      let rec to_dyn t =
+        match (t : _ Pp.Ast.t) with
+        | Nop -> variant "Nop" []
+        | Seq (x, y) -> variant "Seq" [ to_dyn x; to_dyn y ]
+        | Concat (x, y) -> variant "Concat" [ to_dyn x; list to_dyn y ]
+        | Box (i, t) -> variant "Box" [ int i; to_dyn t ]
+        | Vbox (i, t) -> variant "Vbox" [ int i; to_dyn t ]
+        | Hbox t -> variant "Hbox" [ to_dyn t ]
+        | Hvbox (i, t) -> variant "Hvbox" [ int i; to_dyn t ]
+        | Hovbox (i, t) -> variant "Hovbox" [ int i; to_dyn t ]
+        | Verbatim s -> variant "Verbatim" [ string s ]
+        | Char c -> variant "Char" [ char c ]
+        | Break (x, y) ->
+          variant "Break" [ triple string int string x; triple string int string y ]
+        | Newline -> variant "Newline" []
+        | Text s -> variant "Text" [ string s ]
+        | Tag (s, t) -> variant "Tag" [ tag_to_dyn s; to_dyn t ]
+      in
+      to_dyn t
+  ;;
 end
 
 module Result = Result
--- a/src/dune_engine/build_system.mli
+++ b/src/dune_engine/build_system.mli
@@ -126,6 +126,19 @@ module Error : sig
     val current : t -> error Id.Map.t
     val empty : t
   end
+
+  module For_tests : sig
+    (** Internal helpers for testing purposes. Do not use. *)
+
+    (** Construct an [Error.t] *)
+    val make
+      :  description:
+           [ `Exn of Exn_with_backtrace.t | `Diagnostic of Compound_user_error.t ]
+      -> dir:Path.t option
+      -> promotion:Diff_promotion.Annot.t option
+      -> unit
+      -> t
+  end
 end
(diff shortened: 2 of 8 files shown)


Alizter:
@rgrinberg I think the correct fix for this would be to make `user_message.ml` remember prefixes in the type. Otherwise, I don't see an easy way of removing `Error:` short of regexing.

src/dune_engine/build_system.mli
```diff
@@ -108,6 +109,14 @@ module Error : sig
 
   val promotion : t -> Diff_promotion.Annot.t option
 
+  val make
+    :  id:Id.t
+    -> description:[ `Exn of Exn_with_backtrace.t | `Diagnostic of Compound_user_error.t ]
+    -> ?dir:Path.t
+    -> ?promotion:Diff_promotion.Annot.t
```
rgrinberg, line 116: Don't use optional arguments without a concrete reason.

Also, this function should be moved to a `For_tests` submodule.

src/dune_rpc_impl/dune_rpc_impl.ml
```diff
@@ -5,6 +5,7 @@ module For_handlers = For_handlers
 module Private = Dune_rpc_client.Private
 module Watch_mode_config = Watch_mode_config
 module Where = Dune_rpc_client.Where
+module Diagnostics = Diagnostics
```
rgrinberg, line 8: Since you only need one function, you can just export the function you need. Also, it should be accessible as `Diagnostics.For_tests.xxx`.

Alizter:
@rgrinberg I've pushed another test which shows how location printing is kind of funny. I need some help scrubbing the output however.

test/expect-tests/dune_rpc/dune
```diff
@@ -5,10 +5,13 @@
  (libraries
   ocaml_config
   dune_util
+  dune_console
   dune_rpc_private
   dune_rpc_server
   dune_rpc_impl
   dune_rpc_client
+  dune_engine
+  dune_re
```
Alizter, line 14: Need to remove this, didn't use it in the end.

test/expect-tests/dune_rpc/dune_rpc_tests.ml
```diff
@@ -925,3 +925,37 @@ let%expect_test "print digests for all public RPCs" =
...
+  [%expect
+    {|
+    ---- Original ----
+    File "Bar", line 1, characters 2-3:
+    Error: An error with location!
+    ------- RPC ------
+    Error: (In directory /Foo)
+    File "/Foo/Bar", line 1, characters 2-3:
+
+    Error: An error with location! |}]
```
Alizter, line 960: @rgrinberg It seems to me like we shouldn't have to print the `In directory` part as the loc already contains this info. WDYT?

rgrinberg: Yeah, I guess it's redundant if the location is already there.

test/expect-tests/dune_rpc/dune_rpc_tests.ml
```diff
@@ -925,3 +925,37 @@ let%expect_test "print digests for all public RPCs" =
       Request: Sexp
       Response: 33528f248084297d123a6ebd4c3ddee0 |}]
 ;;
+
+let () =
+  Stdune.Path.set_root (Stdune.Path.External.of_filename_relative_to_initial_cwd ".")
+;;
+
+let test ?dir ?promotion main =
```
rgrinberg, line 933: no optional arguments

otherlibs/stdune/src/stdune.ml
```diff
@@ -30,6 +30,32 @@ module Pp = struct
   let compare ~compare x y =
     Ordering.of_int (Pp.compare (fun a b -> Ordering.to_int (compare a b)) x y)
   ;;
+
+  let to_dyn tag_to_dyn t =
```
Alizter, line 34: I've added Pp.to_dyn since I've been wanting this for a while and it make our life easier when testing.

Alizter:
@rgrinberg I've also included dyn serialisation so we can more finely compare the Pp structure that has been serialised.

test/expect-tests/dune_rpc/dune
```diff
@@ -5,10 +5,12 @@
  (libraries
   ocaml_config
   dune_util
+  dune_console
   dune_rpc_private
   dune_rpc_server
   dune_rpc_impl
   dune_rpc_client
+  dune_engine
```
rgrinberg, line 13: can you separate this into a different test library? Something like `dune_rpc_impl_tests`.

This library is specifically for testing the RPC machinery and not concrete dune rpc things

Alizter: Sure

Alizter:
@rgrinberg I've moved the tests now and added yet another example of an error message that changes. All these examples will be fixed by my upcoming PR.

src/dune_engine/build_system.mli
```diff
@@ -88,6 +88,7 @@ module Error : sig
 
     module Map : Map.S with type key = t
 
+    val gen : unit -> t
```
rgrinberg, line 91: Do you need to make this function public? Can you just make `make` generate a new id? It's only used for the tests anyway.

Alizter: Good point

PR #8414, opened 2023-08-17
feature(pkg): Write identifier of repository into metadata

Leonidas-from-XIV:
This PR adds an identifier of the repo that was passed in to the metadata so it is potentially feasible to recreate the lock with the exact same versions.

Currently supported identifiers are

1. Git commit hashes, if `git` was found and the repo is a `git` repo
2. ~~OPAM `stamp` values, if the repo has them set (opam-repository tarballs do)~~

### DIFF
--- a/src/dune_pkg/repository_id.ml
+++ b/src/dune_pkg/repository_id.ml
@@ -1,0 +1,68 @@
+open Import
+
+type t = Git_hash of string
+
+let equal a b =
+  match a, b with
+  | Git_hash a, Git_hash b -> String.equal a b
+;;
+
+let encode : t -> Dune_sexp.t = function
+  | Git_hash commitish ->
+    List
+      [ Dune_lang.atom_or_quoted_string "git_hash"
+      ; Dune_lang.atom_or_quoted_string commitish
+      ]
+;;
+
+let decode =
+  let open Dune_sexp.Decoder in
+  let+ constr, stamp = pair string string in
+  match constr with
+  | "git_hash" -> Git_hash stamp
+  | _ -> failwith "TODO: parsing failure"
+;;
+
+let to_dyn = function
+  | Git_hash commitish -> Dyn.Variant ("Git_hash", [ String commitish ])
+;;
+
+let attempt_repo_id ~dir =
+  let head_path = Path.append_local dir (Path.Local.of_string "HEAD") in
+  match Path.lstat head_path with
+  | Ok { st_kind = S_REG; _ } ->
+    let head = Io.file_line head_path 0 in
+    (match String.lsplit2 head ~on:' ' with
+     | None ->
+       (* when checking out a commit *)
+       Some (Git_hash head)
+     | Some ("ref:", reference) ->
+       (* when a reference is checked out (e.g. a branch) *)
+       let reference = Path.Local.of_string reference in
+       let rev_path = Path.append_local dir reference in
+       let commit_id = Io.file_line rev_path 0 in
+       Some (Git_hash commit_id)
+     | Some _ -> None)
+  | Ok _ | Error _ -> None
+;;
+
+let repo_id_of_git dir =
+  let res = attempt_repo_id ~dir in
+  match res with
+  | Some _ as v -> v
+  | None ->
+    (* if it is not a bare repo the repo is in .git, try again here *)
+    let git_dir = Path.Local.of_string ".git" in
+    let dir = Path.append_local dir git_dir in
+    attempt_repo_id ~dir
+;;
+
+let of_path dir =
+  match Path.stat dir with
+  | Ok { st_kind = S_DIR; _ } -> repo_id_of_git dir
+  | Ok _ | Error _ -> None
+;;
+
+module Private = struct
+  let git_hash s = Git_hash s
+end
--- a/src/dune_pkg/repository_id.mli
+++ b/src/dune_pkg/repository_id.mli
@@ -1,0 +1,15 @@
+open Import
+
+(** [t] represents an identifier of the state of an opam repository that can
+    potentially be used to reproduce the lock dir/lock file exactly. *)
+type t
+
+val encode : t -> Dune_sexp.t
+val decode : t Dune_sexp.Decoder.t
+val equal : t -> t -> bool
+val to_dyn : t -> Dyn.t
+val of_path : Path.t -> t option
+
+module Private : sig
+  val git_hash : string -> t
+end
--- a/test/blackbox-tests/test-cases/pkg/opam-repository-download.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-repository-download.t
@@ -74,6 +74,7 @@ serve our opam-repository as a mock.
   $ dune exec -- ./server.exe $PORT &
   $ # wait for server to come up, this takes a bit
   $ while [ ! -f server.pid ]; do sleep 0.01; done
+  $ rm server.pid
 
 Run Dune and tell it to cache into our custom cache folder.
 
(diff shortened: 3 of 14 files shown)


src/dune_pkg/repository_id.ml
```diff
@@ -0,0 +1,83 @@
...
+           | Variable ({ pelem = "stamp"; _ }, { pelem = String stamp; _ }) -> Some stamp
+           | _ -> None)
+       in
+       stamp |> Option.map ~f:(fun s -> Opam_stamp s) |> Fiber.return)
+;;
+
+let of_path dir =
+  let open Fiber.O in
+  let failure_mode = Process.Failure_mode.Return in
+  let display = Display.Silent in
```
rgrinberg, line 65: What's the difference between Quiet and Silent?

src/dune_pkg/repository_id.ml
```diff
@@ -0,0 +1,83 @@
...
+  let path = Env.get Env.initial "PATH" |> Option.value_exn |> Bin.parse_path in
```
rgrinberg, line 66: This can use `Env_path` somehow so that we don't hard code the environment variable names.

src/dune_pkg/repository_id.ml
```diff
@@ -0,0 +1,83 @@
...
+let equal a b =
+  match a, b with
+  | Git_hash a, Git_hash b | Opam_stamp a, Opam_stamp b -> String.equal a b
+  | _, _ -> false
+;;
+
+let encode : t -> Dune_sexp.t = function
+  | Git_hash commitish ->
+    List
+      [ Dune_lang.atom_or_quoted_string "Git_hash"
```
rgrinberg, line 18: Should be `git_hash`. We don't use uppercase names anywhere in dune files.

src/dune_pkg/repository_id.ml
```diff
@@ -0,0 +1,83 @@
...
+       in
+       stamp |> Option.map ~f:(fun s -> Opam_stamp s) |> Fiber.return)
+;;
+
+let of_path dir =
+  let open Fiber.O in
+  let failure_mode = Process.Failure_mode.Return in
+  let display = Display.Silent in
+  let path = Env.get Env.initial "PATH" |> Option.value_exn |> Bin.parse_path in
+  match Path.is_directory dir with
```
rgrinberg, line 67: This function swallows permission errors. We should avoid using it and use `Path.stat` when possible

src/dune_pkg/repository_id.mli
```diff
@@ -0,0 +1,13 @@
+open Stdune
```
rgrinberg, line 1: A sentence describing the purpose of this module would be appreciated

test/blackbox-tests/test-cases/pkg/opam-repository-download.t
```diff
@@ -99,3 +100,65 @@ containing the repository:
...
+
+  $ cat > mock-opam-repository/repo <<EOF
+  > opam-version: "2.0"
+  > stamp: "repository-identifier"
+  > EOF
+  $ cd mock-opam-repository
+  $ tar czf ../index.tar.gz *
+  $ cd ..
+
+  $ dune exec -- ./server.exe $PORT &
```
rgrinberg, line 114: Do you kill the server somewhere in this test?

src/dune_pkg/repository_id.ml
```diff
@@ -0,0 +1,83 @@
...
+  | Git_hash commitish -> Dyn.Variant ("Git_hash", [ String commitish ])
+  | Opam_stamp stamp -> Dyn.Variant ("Opam_stamp", [ String stamp ])
+;;
+
+let repo_id_of_stamp dir =
+  let local = Path.Local.of_string "repo" in
+  let repo_file = Path.append_local dir local in
+  match Path.exists repo_file with
+  | false ->
+    (* no repo file, we can't read this, bail *)
```
gridbugs, line 47: `dune pkg lock` will now behave in subtly different ways depending on whether `git` is installed. That's fine but I think we should print a warning in this case to help clarify the situation.

src/dune_pkg/repository_id.ml
Leonidas-from-XIV: `Quiet` still displays error messages when commands fail (and they output an error message), in our case that's a lot of `fatal: invalid gitfile format: $DIR/dune/_build/.sandbox/.git`. I considered changing the definition of `Quiet` but wasn't sure if that wouldn't break a lot of places that depended on printing the error message.

test/blackbox-tests/test-cases/pkg/opam-repository-download.t
Leonidas-from-XIV: The server terminates automatically after serving one request, it's specifically written that way.

src/dune_pkg/repository_id.ml
Leonidas-from-XIV: I'm not super happy about this either (though I would assume git has a high installation base and be able to fetch a git repo you needed to have git already). The alternative could be to determine the revision from the `.git` directory manually. This isn't all that difficult given the needs of this PR (parse `.git/HEAD`, look up the referenced ref, read that file for the hash) but I am not familiar with how much of the `.git` directory is a "public" API and how much are implementation details.

Let me know what you think. Removing the `git` binary avoids all the path lookups but makes me (slightly) concerned about the future-proofness of reading `.git` directory (or we could use `ocaml-git` but for such a simple task that would be extreme overkill).

src/dune_pkg/repository_id.ml
rgrinberg: Note that this error is intentional and that your test is trying to escape the sandbox (by running git without `-C` or without initializing a repo in the test)

Leonidas-from-XIV: The error in this case is intentional since the repo is not a valid repo, so it should fall back to use the `repo` file (which it does). But with `Quiet` it will always display this error message before falling back to the other code, which I would say is overly alarming.

I can either set `-C` or `~dir` to the folder, either way the result is the same. I don't think setting `-C` makes it not escape the sandbox. The git documentation makes it sound like it's basically `cd ~dir; git ...`.

However I think that maybe implementing `git rev-parse HEAD` is easy/stable enough to not use the `git` binary for it.

rgrinberg: You could also just do `git init` inside the test.

Leonidas-from-XIV: I am, in line 136?

Leonidas-from-XIV: Ah, now I understand, git traverses up and then hits that (deliberate trap) folder and breaks. It looks like this can be disabled using `--git-dir`.

src/dune_pkg/repository_id.ml
gridbugs: Since the only difference in behaviour between systems with git installed and those without it will be that the latter will lack the repo id in package metadata I think it's ok to just omit the field when git isn't installed. There won't be a functional difference as this field will just be read by humans to help understand how their lockdir was generated and as you said most users will already have git installed. In the case that git isn't installed I much prefer printing a warning indicating that the metadata will lack the repo id because git isn't installed rather than having dune try to parse the .git directory as it's simpler to implement for us and I think less surprising behaviour for users.

Leonidas-from-XIV: I've solved it by reimplementing that part parsing the `.git` data, so it works whether you have `git` installed or not, which also solves the error handling if the repo can't be parsed by `git`.

This part of git is fairly simple so I am not too worried about it breaking.

rgrinberg:
What do "stamp values" allow us to accomplish? From my brief look, it's not much. If that's the case, I'd rather just drop them and stick to git only identifiers. Especially since from our discussion we agreed on making repo fetching rely on git by default is totally fine.

Leonidas-from-XIV:
> What do "stamp values" allow us to accomplish? From my brief look, it's not much.

In the case of opam-repository, the stamp is basically `git rev-parse --short HEAD` so it would be possible to retrieve the repo at that state by `git checkout`.

rgrinberg:
In that case, let's include them as a git hash? Although note that shortened hashes aren't 100% reliable.

Skipping them still sees like a simple thing to do since we won't be fetching opam-repository this way anyway.

Leonidas-from-XIV:
I don't think you can state that it is a git hash, as the stamp value can be anything that the tarball creator considers a stamp value. It just happens to be a git hash in the case of opam-repository, but to do so we'd need to special-case the tarball.

I'll remove it then.

rgrinberg (approved):

src/dune_pkg/repository_id.ml
```diff
@@ -0,0 +1,68 @@
...
+  | Some _ as v -> v
+  | None ->
+    (* if it is not a bare repo the repo is in .git, try again here *)
+    let git_dir = Path.Local.of_string ".git" in
+    let dir = Path.append_local dir git_dir in
+    attempt_repo_id ~dir
+;;
+
+let of_path dir =
+  match Path.lstat dir with
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Ah, indeed. I think I accidentally used `lstat` when I actually meant `stat`. Fixed.

PR #8416, opened 2023-08-18
Add with_prefix keyword for installing globs

gridbugs:
This allows the `(glob_files[_rec] ...)` construct to specify an alternative prefix for destinations when installing files. Previously the prefix of the glob (the part before the `*`) would be used but this proved too inflexible in some cases such as when the glob starts with "..".

### DIFF
--- a/src/dune_rules/install_entry.ml
+++ b/src/dune_rules/install_entry.ml
@@ -13,11 +13,45 @@ let expand_str_with_check_for_local_path ~expand_str sw =
     str)
 ;;
 
+module Glob_files_with_optional_prefix = struct
+  type t =
+    { glob_files : Dep_conf.Glob_files.t
+    ; prefix : String_with_vars.t option
+    }
+
+  let decode =
+    let open Dune_lang.Decoder in
+    let install_glob_version_check = Dune_lang.Syntax.since Stanza.syntax (3, 6) in
+    let install_glob_with_prefix_version_check =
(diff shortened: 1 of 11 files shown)


gridbugs:
Note that this follows on from https://github.com/ocaml/dune/pull/8350. Merge that PR before this one.

gridbugs:
This was originally part of https://github.com/ocaml/dune/pull/8350 but I split it out into its own PR.

src/dune_rules/install_entry.ml
```diff
@@ -45,13 +69,41 @@ module File = struct
...
+            | None -> src
+            | Some (prefix, prefix_loc) ->
+              let path_without_prefix =
+                match String.drop_prefix path ~prefix:glob_prefix with
+                | Some s -> s
+                | None ->
+                  Code_error.raise
+                    ~loc:glob_loc
+                    (Printf.sprintf
+                       "This glob has a prefix %s, but the match %s does not begin with \
```
rgrinberg, line 97: In `Code_error.raise`, the first argument should be the same for all errors. This is done so that it's easy to grep for a specific type of error just by the name. The parameters of the error should be included in the second argument. So for example:

```ocaml
Code_error.raise "Glob has a prefix which does not match the prefix of the match"
  [ "glob_prefix", ..
  ; "match_prefix", ..
  ]
```

gridbugs: fixed

christinerose (approved):

PR #8421, opened 2023-08-18
refactor(engine): remove pointless intermediate function

rgrinberg:


Alizter (commented):
Wouldn't it be better to make this a `let+` and then match on it?

Leonidas-from-XIV (approved):

PR #8424, opened 2023-08-18
refactor: move scheme to own library

rgrinberg:


src/scheme/dune
```diff
@@ -0,0 +1,3 @@
+(library
```
Alizter, line 1: Missing synopsis about "internal dune lib" and bisect ppx instruementation.

rgrinberg: synopsis is only needed for public libraries

PR #8426, opened 2023-08-19
refactor: simplify diagnostic event type

Alizter:
Before we were using
```
type t =
  | Add of Diagnostic.t
  | Remove of Diagnostic.t
```
now we use
```
type t =
  | Add of Diagnostic.t
  | Remove of Diagnostic.Id.t
```

We bump the RPC version for the diagnostic procedure to account for this new type.

### DIFF
--- a/otherlibs/dune-rpc/private/procedures.ml
+++ b/otherlibs/dune-rpc/private/procedures.ml
@@ -181,13 +181,57 @@ module Poll = struct
   end
 
   module Diagnostic = struct
+    module V1 = struct
+      module Event = struct
+        type t =
+          | Add of Diagnostic.t
+          | Remove of Diagnostic.t
+
+        let sexp =
+          let diagnostic = Diagnostic.sexp in
+          let open Conv in
+          let add = constr "Add" diagnostic (fun a -> Add a) in
+          let remove = constr "Remove" diagnostic (fun a -> Remove a) in
+          sum
+            [ econstr add; econstr remove ]
+            (function
+             | Add t -> case t add
+             | Remove t -> case t remove)
+        ;;
+
+        let to_event : t -> Diagnostic.Event.t = function
+          | Add diag -> Add diag
+          | Remove diag -> Remove diag.id
+        ;;
+
+        let of_event : Diagnostic.Event.t -> t option = function
+          | Add diag -> Some (Add diag)
+          | Remove _ -> None
+        ;;
+      end
+    end
+
     let name = "diagnostic"
 
     let v1 =
+      Decl.Request.make_gen
+        ~version:1
+        ~req:Id.sexp
+        ~resp:(Conv.option (Conv.list V1.Event.sexp))
+        ~upgrade_req:Fun.id
+        ~downgrade_req:Fun.id
+        ~upgrade_resp:(Option.map ~f:(List.map ~f:V1.Event.to_event))
+        ~downgrade_resp:(fun x ->
+          let open Option.O in
+          let* x = x in
+          Option.List.all (List.map ~f:V1.Event.of_event x))
+    ;;
+
+    let v2 =
       Decl.Request.make_current_gen
+        ~version:2
         ~req:Id.sexp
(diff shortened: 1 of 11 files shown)


Alizter:
cc @pmwhite

otherlibs/dune-rpc/private/procedures.ml
```diff
@@ -181,13 +181,57 @@ module Poll = struct
...
+        ~version:1
+        ~req:Id.sexp
+        ~resp:(Conv.option (Conv.list V1.Event.sexp))
+        ~upgrade_req:Fun.id
+        ~downgrade_req:Fun.id
+        ~upgrade_resp:(Option.map ~f:(List.map ~f:V1.Event.to_event))
+        ~downgrade_resp:(fun x ->
+          let open Option.O in
+          let* x = x in
+          Option.List.all (List.map ~f:V1.Event.of_event x))
```
Alizter, line 227: I don't have a faithful way to reconstruct the full diagnostic from just the id so this is just an approximation. Add is preserved but Remove is thrown out. This would mean in practice that a new server and an old client would be missing the `Remove`s. @rgrinberg any idea how I could make this situation better?

pmwhite: I like your idea of making a dummy diagnostic with the same Id. I doubt clients need more than the Id, but it's probably worth checking (I can do that Monday). If they do, then probably this patch is not such a good idea?

src/dune_rpc_impl/server.ml
```diff
@@ -248,7 +248,8 @@ let handler (t : _ t Fdecl.t) action_runner_server handle : 'a Dune_rpc_server.H
                match prev, now with
                | None, None -> assert false
                | Some prev, None ->
-                 Some (Diagnostics.diagnostic_event_of_error_event (Remove prev))
+                 Some
+                   (Diagnostics.diagnostic_event_of_error_event (Remove (Error.id prev)))
```
pmwhite, line 252: Small nitpick: we have the id available directly from the key passed to f, so we shouldn't need to pull it out of prev.

Alizter: That seems reasonable. Shall we do that in another PR? The behaviour here should try to stay identical.

otherlibs/dune-rpc/private/procedures.ml
rgrinberg: Why can't we support both versions of the API on the server?

src/dune_rpc_impl/server.ml
pmwhite: Wouldn't behavior remain the same? I think it'd be fine to do it in this PR, but I'm also fine with doing it in a separate one.

otherlibs/dune-rpc/private/procedures.ml
Alizter: @rgrinberg Because we don't have a way of supporting the first API. In this PR, I've removed `Remove Diagnostic.t` and replaced it with `Remove Diagnostic.Id.t`. So if we want to support the older version of the API faithfully we have to somehow find the full diagnostic from the Id. (I'm not sure we keep them around for us to use like that.)

rgrinberg: Let's back off for a moment. What's preventing you today from introducing your new API as a totally separate function that doesn't touch the existing diagnostic API?

Alizter: I guess we could do that but I'm out of my depth on this one.

rgrinberg: I haven't looked closely, but it seems like it should be simple enough to keep the old API on the server side. Where you still have the set of errors for every long polling connection and you can retrieve the actual error based on the ID.

Alizter:
The version compatibility here is complete nonsense. I tried using this with ocaml-lsp and it is unable to clear older diagnostics. I'm closing this for the time being as I don't know how to make it better and I don't think it is a priority at the moment.

PR #8427, opened 2023-08-19
refactor: remove module alias from dune init

Alizter:
This module alias was supposedly because we use `Dune_lang.Cst` all over the place, but on inspection that doesn't seem to be true. Therefore it is not worth the indirection.


rgrinberg:
3 uses is enough to justify the use of this alias. Though it should come from `Dune_sexp` rather than `Dune_lang`.

Alizter:
OK, it was something that I thought would make it easier to read but it doesn't matter.

PR #8428, opened 2023-08-19
test: dune_console

Alizter:
We add tests for `dune_console` together with observing that status line clearing doesn't work properly when lines are being wrapped.

### DIFF
--- a/test/expect-tests/dune_console/dune_console_tests.ml
+++ b/test/expect-tests/dune_console/dune_console_tests.ml
@@ -1,0 +1,104 @@
+(** Creation of Dune_console is stateful so we introduce a new module for each test. *)
+module New () = Dune_console
+
+module type New_console = module type of Dune_console
+
+(** Throughout this file we use regular [" "] strings rather than [{| |}] because we want
+    to observe the escape sequences.
+
+    In order to keep tests accross different backends consistent, we create some generic
+    test scripts here that take the created [Console]. We then test these for each
+    backend. *)
+
+let test_basic_usage (module Console : New_console) =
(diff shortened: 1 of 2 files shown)


test/expect-tests/dune_console/dune_console_tests.ml
```diff
@@ -0,0 +1,104 @@
...
+   fully clear the previous lines when wrapped. *)
+let%expect_test "Status line clearing with wrapping." =
+  let module Console = New () in
+  Console.Backend.set Console.Backend.progress;
+  test_status_line_clearing_with_wrapping (module Console);
+  [%expect
+    "\n\
+     This status line is a problem because of the fact that it is especially long and \
+     therefore will not be cleared \
+     properly.\r                                                                                                                        \
```
Alizter, line 102: cc @Julow You may be interested to know that the formatting here is a bit whack. I would expect that a new `\` is inserted after `\r` and that the white space is also wrapped.

Julow: Would you expect something like this ?

```
  [%expect
    "\n\
     This status line is a problem because of the fact that it is especially long and \
     therefore will not be cleared \
     properly.\r                                                                        \
                                                     \n\
     Here is another\r               \r "]
```

Alizter: Yes, that seems more reasonable. Just wanted to give you a heads up, this test won't be around for long.

Julow: Ok, thanks!

PR #8429, opened 2023-08-19
TUI

Alizter:
In this PR we add some more features to Dune's TUI.

```
dune build -w --display tui
```

#### Description

We start by vendoring Lwd (light-weight document) and Nottui. We build on the Notty TUI that we currently have by switching components over to Nottui. This allows visual components to react to user input more locally, as in button handlers exist for each component rather than in a global loop like before. Lwd allows us to keep a persistent document state and allows components to be rerendered reactively to any changes.

This work builds on top of #7484.

#### Doc:
- [x] changelog
- [ ] (CLI) Documentation

#### Issues:
- Sometimes we might get a crash which will cause OCaml to complain about resouce deadlocks with the mutex. It happens occasionally with `dune build` and is reproducible when using `dune monitor` and connecting to a build with a diagnostic.
- In the above kind of crash, the terminal gets borked and requires a reset. We should work out how to recover.
- When we crash on startup, we don't print anything. Although, we do output to the log. (at least for `dune build`)

#### Visual issues:
- Terminals that don't have unicode don't look good. Not much we can do about this.
- When minimizing a message and getting a synopsis, the outline decorations might be over approximated. It's hard to do better than a rough approximation here at the moment.
- We cannot copy text. We should work out a way to allow the user to select and highlight text. (or even just add a copy button but thats a bit trickier).
- Make sure colours look good on light and dark terminals. See below.

---
### Demo

<details>
<summary>Click me</summary>

When running Dune (in watch mode) with `--display tui` the user will now be presented with the following screen.

image

The `?` button in the top left corner opens the help screen. Clicking this opens the help screen which looks like this:

image

Next we purposefully create some errors in our build so we can see what messages look like:

image

As you can see, scroll bars appear in the TUI which can be interacted with using the mouse. You can also use the arrow keys or vim bindings to scroll and move around.

Above each message there is a horizontal rule, which shows 3 pieces of information.
1. Whether the message can be minimized or expanded. (currently it looks to be expandable).
2. What the ordinal of this message is.
3. The total number of messages.

We now click on the first message bar in order to demonstrate the minimization feature:

image

A short synopsis of the message is included and the rest of the message is hidden from view. We can see what this looks like as more error messages accumulate:

image

We can press `m` to minimize or expand all the messages at once, allowing the user to pick and choose which messages they want to see:

image

Here is a video demonstrating these behaviours:

[Screencast from 2023-08-21 19-20-44.webm](https://github.com/ocaml/dune/assets/8614547/a367f7ca-023f-4470-a02a-a19617aa35f6)

</details>

<details>
<summary>Appearance on white background</summary>
The white background doesn't look to great. It might be impossible to please everybody with this choice, so perhaps a better route would be to make these colours customizable from dune config.

image

</details>


Alizter:
image

src/dune_tui/drawing/user_message_to_image.mli
```diff
@@ -0,0 +1,5 @@
+open Stdune
+
+(** [User_message_to_image.pp ~attr pp] converts a pretty-printer [pp] to a
+    [Notty.image] using the given [attr] for styling. *)
+val pp : ?attr:Notty.attr -> User_message.Style.t Pp.t -> Notty.image
```
rgrinberg, line 5: Avoid optional arguments please

src/dune_util/report_error.ml
```diff
@@ -200,8 +200,13 @@ let gen_report exn backtrace =
...
     in
+    Log.info_user_message msg;
     Console.print_user_message msg
 ;;
 
 let report { Exn_with_backtrace.exn; backtrace } = gen_report exn (Some backtrace)
-let report_exception exn = gen_report exn None
+
+let report_exception exn =
+  let backtrace = Printexc.get_raw_backtrace () in
```
rgrinberg, line 210: This backtrace is probably wrong. We shouldn't capture it here.

src/dune_tui/drawing/user_message_to_image.ml
```diff
@@ -0,0 +1,111 @@
+open Stdune
+
+let attr_of_ansi_color_rgb8 (c : Ansi_color.RGB8.t) =
+  let module A = Notty.A in
```
rgrinberg, line 4: Isn't this alias in Import already?

src/dune_tui/drawing/user_message_to_image.ml
```diff
@@ -0,0 +1,111 @@
...
+  | 14 -> A.lightcyan
+  | 15 -> A.lightwhite
+  | i when i <= 231 ->
+    let i = i - 16 in
+    let r = i / 36 in
+    let g = i / 6 mod 6 in
+    let b = i mod 6 in
+    A.rgb ~r ~g ~b
+  | i when i <= 255 -> A.gray (i - 232)
+  | i -> Code_error.raise "invalid 8-bit color" [ "value", Dyn.int i ]
```
rgrinberg, line 29: This isn't new, but this behavior is surprising. We shouldn't crash with a Code_error on bad input. We should probably just ignore.

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,291 @@ module Tui = struct
...
+      in
+      stabilize ()
+    in
+    (* Finally we use Notty to show the image. *)
     Term.image (term ()) image
   ;;
 
-  let resize mutex (state : Dune_threaded_console.state) =
+  (** Update any local state and finish *)
+  let finish_interaction () = Unix.gettimeofday ()
```
rgrinberg, line 309: Can you inline this function?

src/dune_tui/widgets/help_box.mli
```diff
@@ -0,0 +1,14 @@
...
+
+type t =
+  { ui : Ui.t
+  ; toggle : unit -> unit
+  }
+
+val make
+  :  helper_attr:A.t
+  -> divider_attr:A.t
+  -> help_screen_text:string list
```
rgrinberg, line 11: Is this a list of lines? If so it should be `help_screen_lines`.

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,291 @@ module Tui = struct
...
+       for separating messages. It also has a handler for clicks that minimizes the
+       message. *)
+    let horizontal_line_with_count total index =
+      let+ is_hidden = Lwd.get is_message_hidden
+      and+ w = max_message_length
+      and+ synopsis = Lwd.app (message_synopsis ~attr:helper_attr) (Lwd.return index) in
+      let index_is_hidden = is_hidden index in
+      let status =
+        I.hcat
+          [ I.uchar divider_attr (Uchar.of_int 0x169c) 1 1
```
rgrinberg, line 102: Can you make all `Uchar.of_int <magic-number>` into proper constants and add some comments for them?

src/dune_tui/drawing/user_message_to_image.ml
Alizter: However it is imossible for this Code_error to raise. It is matching on the output of Ansi_color.RGB8.to_int which can only be < 256.

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,288 @@ module Tui = struct
...
+  (** style for user feedback like message count, or scrollbar position *)
+  let user_feedback_attr = A.(fg cyan)
+
+  (** [horizontal_rule ~attr ~w] draws a horizontal line with [attr] of length [w]. *)
+  let horiztonal_rule ~attr ~w = I.uchar attr (Uchar.of_int 0x2015) w 1
+
+  (** Here we keep some persistent state about the program that we udpate each time we
+      render. This allows other components to "react" to changes using [Lwd]. *)
+
+  let term_size = Lwd.var (Term.size (term ()))
```
rgrinberg, line 51: This should be lazily initialized

Alizter: It will get initialized in the render loop anyway, so I will set it to some dummy value for now.

src/dune_util/report_error.ml
rgrinberg: Still wrong

Alizter: Yes, I haven't touched it yet.

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,289 @@ module Tui = struct
...
+  let renderer = Renderer.make ()
+
   let render (state : Dune_threaded_console.state) =
-    let messages = Queue.to_list state.messages in
-    let image = image ~status_line:state.status_line ~messages in
+    let size = Term.size (term ()) in
+    (* Update the persistant values tracked by other components. *)
+    Lwd.set term_size size;
+    Lwd.set messages (Queue.to_list state.messages);
+    Lwd.set status_line state.status_line;
```
rgrinberg, line 293: Same issue with `status_line` and `messages`.

src/dune_util/report_error.ml
```diff
@@ -200,6 +200,7 @@ let gen_report exn backtrace =
       | User -> msg
       | Developer -> append msg (i_must_not_crash ())
     in
+    Log.info_user_message msg;
```
rgrinberg, line 203: This needs to be `Log.log`

Alizter: I've also modified the type of `Log.log` to take `User_message.t` rather than `Pp.t list` which makes it more usable.

src/dune_tui/drawing/unicode.ml
```diff
@@ -0,0 +1,34 @@
...
+let ogham_feather_mark = Uchar.of_int 0x169B
+
+(** ᚜ U+169C *)
+let ogham_reversed_feather_mark = Uchar.of_int 0x169C
+
+(** ― U+2015 *)
+let horizontal_bar = Uchar.of_int 0x2015
+
+(** ═ U+2550 *)
+let box_drawings_double_horizontal = Uchar.of_int 0x2550
```
rgrinberg, line 13: This needs an `mli` and the odoc comments should be moved there.

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,304 @@ module Tui = struct
...
+    let max_message_length =
+      let+ messages = message_images
+      and+ w, _ = Lwd.get term_size in
+      let lengths = List.map messages ~f:I.width in
+      match List.max lengths ~f:Int.compare with
+      | None -> w - 1
+      | Some l -> max l w
+    ;;
+
+    (** We approximate the first line of the message. Unfortunately due to the way Notty
```
rgrinberg, line 79: This should be a normal comment. odoc comments are for mli's

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,304 @@ module Tui = struct
...
+      in
+      stabilize ()
+    in
+    (* Finally we use Notty to show the image. *)
     Term.image (term ()) image
   ;;
 
-  let resize mutex (state : Dune_threaded_console.state) =
+  (** Update any global state and finish *)
+  let finish_dirty_interaction ~mutex (state : Dune_threaded_console.state) =
```
rgrinberg, line 325: Call this function `set_dirty` and inline the `Unix.gettimeoday ()` at all the callers.

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,304 @@ module Tui = struct
...
     in
-    Notty.I.vcat (messages @ status)
+    Ui.zcat
+      [ Ui.atom
+        @@ I.zcat [ hsnap_or_leave ~width:w status; horiztonal_rule ~attr:helper_attr ~w ]
+      ; help_button
+      ]
+  ;;
+
+  (** Our document has 3 components:
```
rgrinberg, line 263: Should be a normal comment.

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,304 @@ module Tui = struct
...
+        | `Scroll dir ->
+          vscroll ~dir;
+          `Handled
+        | _ -> `Unhandled
+      in
+      ui |> Ui.keyboard_area keyboard_handler |> Ui.mouse_area mouse_handler
+    ;;
+  end
+
+  (** Help pop up contains help information for the user. It also includes a hook so that
```
rgrinberg, line 211: Normal comment

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,304 @@ module Tui = struct
...
+        collapse the message. *)
+    let line_separated_message ~total index msg =
+      let+ horizontal_line_with_count = horizontal_line_with_count total index
+      and+ toggle = Lwd.app (Lwd.get is_message_hidden) (Lwd.return index) in
+      if toggle
+      then horizontal_line_with_count
+      else Ui.vcat [ horizontal_line_with_count; Ui.atom msg ]
+    ;;
+
+    (** We take all the messages in the console and display them as
```
rgrinberg, line 158: Normal comment

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,304 @@ module Tui = struct
...
+      Ui.atom
+      @@ I.zcat
+           I.
+             [ I.hsnap ~align:`Left w (toggle_indicator <|> status <|> synopsis)
+             ; horiztonal_rule ~w ~attr:divider_attr
+             ]
+      |> Ui.mouse_area mouse_handler
+    ;;
+
+    (** This components displays a line followed by a message. Cliking on the line will
```
rgrinberg, line 148: Normal comment

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,304 @@ module Tui = struct
...
+    let is_message_hidden = Lwd.var (fun _ -> false)
 
-  let image ~status_line ~messages =
+    (** We share some work turning messages into images. *)
+    let message_images =
+      let+ messages = Lwd.get messages in
+      List.map messages ~f:(fun x -> User_message_to_image.pp (User_message.pp x))
+    ;;
+
+    (** We calculate the max message length from all the messages. This is used to keep
```
rgrinberg, line 68: Normal comment

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,304 @@ module Tui = struct
...
+
+  module Message_viewer = struct
+    (** Specialized widget for viewing messages in a list. *)
+
+    (** [is_message_hidden] tracks if a given message is hidden. [true] means it is hidden
+        and [false] means it is expanded. *)
+    let is_message_hidden = Lwd.var (fun _ -> false)
 
-  let image ~status_line ~messages =
+    (** We share some work turning messages into images. *)
```
rgrinberg, line 62: I don't understand this comment

src/dune_tui/dune_tui.ml
```diff
@@ -144,35 +33,304 @@ module Tui = struct
     fun () -> !(Lazy.force setup)
   ;;
 
-  let start () = ()
+  (** style for diving visual elements like borders or rules *)
```
rgrinberg, line 36: Normal comment

src/dune_tui/drawing/user_message_to_image.mli
```diff
@@ -0,0 +1,4 @@
+open Stdune
+
+(** [User_message_to_image.pp ~attr pp] converts a pretty-printer [pp] to a [Notty.image]. *)
+val pp : User_message.Style.t Pp.t -> Notty.image
```
rgrinberg, line 4: I question the organization of this `drawing` subdirectory. It's basically one module per function, why?

You could just keep a single drawing module and keep all the code there.

src/dune_tui/dune_tui.ml
Alizter: We are just sharing this Lwd between a few places. If we inline it, we would compute the value twice. I will get rid of the comment since it is not so helpful.

src/dune_tui/drawing/user_message_to_image.mli
Alizter: Done.

rgrinberg:
I removed the fist commit because it still needs some work. In particular, `Report_error` needs to check if the log is set before logging.

rgrinberg:
This is good enough for now. Thanks.

PR #8430, opened 2023-08-19
TUI (no nottui)

Alizter:
This WIP is an alternative to #8429 that doesn't use the nottui additions that were included there.

This is based on notty and we have basic GUI components like pop-up dialogues, scroll bars etc.

There are some issues with this such as resizing causing crashes that need to be resolved.

### DIFF
--- a/src/dune_tui/dune_tui.ml
+++ b/src/dune_tui/dune_tui.ml
@@ -164,6 +164,11 @@ module Tui = struct
     ; mutable debug : bool
     ; mutable reset_count : int
     ; mutable help_screen : bool
+    ; mutable proc_screen : bool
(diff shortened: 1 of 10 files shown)


src/dune_tui/dune_tui.ml
```diff
@@ -36,7 +36,9 @@ module Tui = struct
     fun () -> !(Lazy.force setup)
   ;;
 
-  let start () = Unix.set_nonblock Unix.stdin
+  let term_input_fds, _ = Term.fds (term ())
+
+  let start () = Unix.set_nonblock term_input_fds
```
rgrinberg, line 40: That's wrong. You can't do that anymore

Alizter: That snuck in during rebase, shouldn't be there.

PR #8447, opened 2023-08-21
refactor: introduce simpler env stanza db

rgrinberg:
This is an experiment to replace the `Env_tree`. The goal is to reduce the possibility of cycles, remove unnecessary memoization, and make it easy to introduce new values that depend on the `env` stanza without shoving everything in one module.

### DIFF
--- a/src/dune_rules/env_stanza_db.mli
+++ b/src/dune_rules/env_stanza_db.mli
@@ -1,0 +1,10 @@
+open Import
+
+val value
+  :  default:'a
+  -> dir:Path.Build.t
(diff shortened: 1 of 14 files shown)


src/dune_rules/env_stanza_db.mli
```diff
@@ -0,0 +1,9 @@
+open Import
+
+val value
```
Alizter, line 3: Doesn't look like this is used anywhere.

rgrinberg: It's there for other parts of the code to use and implement their env like components without shoving it all in Env_node

PR #8448, opened 2023-08-21
refactor: warning system

rgrinberg:
Add a [Warning] module to handle warnings in a consistent way in dune.

We'd like dune warnings to have the following two properties:

* Muted while in the vendored directory
* Possible to disable

This PR adds a warning mechanism that makes it easy to handle these two concerns.

I've converted a couple of trivial warnings we currently have to this mechanism. More can be done, but this is a start.

### DIFF
--- a/doc/dune-files.rst
+++ b/doc/dune-files.rst
@@ -689,6 +689,22 @@ map_workspace_root
    .. versionchanged:: 3.7
         Add a way to disable the mapping.
 
+.. _warnings:
+
+warnings
+--------
+
+.. dune:stanza:: warnings
+
+   .. versionadded:: 3.11
+
+   Configure Dune warnings for the project.
+
+   .. dune:field:: <name>
+      :param: <enabled | disabled>
+
+      Enable or disable the warning <name> for the current project.
+
 .. _dune-files:
 
 dune
--- a/src/dune_rules/dune_project.mli
+++ b/src/dune_rules/dune_project.mli
@@ -142,6 +142,7 @@ val default_dune_language_version : Dune_lang.Syntax.Version.t ref
 val set : t -> ('a, 'k) Dune_lang.Decoder.parser -> ('a, 'k) Dune_lang.Decoder.parser
 
 val get_exn : unit -> (t, 'k) Dune_lang.Decoder.parser
+val get : unit -> (t option, 'k) Dune_lang.Decoder.parser
 
 (** Find arguments passed to (using). [None] means that the extension was not
     written in dune-project. *)
@@ -158,6 +159,7 @@ val accept_alternative_dune_file_name : t -> bool
 val strict_package_deps : t -> bool
 val cram : t -> bool
 val info : t -> Package.Info.t
+val warnings : t -> Warning.Settings.t
 
 (** Update the execution parameters according to what is written in the
     [dune-project] file. *)
--- a/src/dune_rules/file_binding.ml
+++ b/src/dune_rules/file_binding.ml
@@ -7,21 +7,24 @@ type ('src, 'dst) t =
       (* The [dune_syntax] field is used for validation which has different
          behaviour depending on the version of dune syntax in use. *)
   ; dune_syntax : Syntax.Version.t
+  ; dir : Path.Source.t option
   }
 
-let to_dyn f g { src; dst; dune_syntax } =
+let to_dyn f g { src; dst; dune_syntax; dir } =
   let open Dyn in
   record
     [ "src", f src
     ; "dst", option g dst
     ; "dune_syntax", Syntax.Version.to_dyn dune_syntax
+    ; "dir", option Path.Source.to_dyn dir
     ]
 ;;
 
-let equal f g { src; dst; dune_syntax } t =
+let equal f g { src; dst; dune_syntax; dir } t =
   f src t.src
   && Option.equal g dst t.dst
   && Syntax.Version.equal dune_syntax t.dune_syntax
+  && Option.equal Path.Source.equal dir t.dir
 ;;
 
 let relative_path_starts_with_parent relative_path =
@@ -33,38 +36,41 @@ let relative_path_starts_with_parent relative_path =
 let validate_dst_for_install_stanza
   ~relative_dst_path_starts_with_parent_error_when
   ~loc
-  dst
-  dune_syntax
+  ~dst
+  ~dir
   =
-  if relative_path_starts_with_parent dst
-  then (
-    match relative_dst_path_starts_with_parent_error_when with
-    | `Deprecation_warning_from_3_11 ->
-      let open Syntax.Version.Infix in
-      if dune_syntax >= (3, 11)
-      then
-        User_warning.emit
-          ~loc
-          [ Pp.textf
-              "The destination path %s begins with %s which will become an error in a \
-               future version of Dune. Destinations of files in install stanzas \
-               beginning with %s will be disallowed to prevent a package's installed \
-               files from escaping that package's install directories."
-              (String.maybe_quoted dst)
-              (String.maybe_quoted Filename.parent_dir_name)
-              (String.maybe_quoted Filename.parent_dir_name)
-          ]
-    | `Always_error ->
-      User_error.raise
-        ~loc
-        [ Pp.textf
-            "The destination path %s begins with %s which is not allowed. Destinations \
-             in install stanzas may not begin with %s to prevent a package's installed \
-             files from escaping that package's install directories."
-            (String.maybe_quoted dst)
-            (String.maybe_quoted Filename.parent_dir_name)
-            (String.maybe_quoted Filename.parent_dir_name)
-        ])
+  match relative_path_starts_with_parent dst with
+  | false -> Memo.return ()
+  | true ->
+    (match relative_dst_path_starts_with_parent_error_when with
+     | `Deprecation_warning_from_3_11 ->
+       Warning_emit.emit
+         Warning.escaping_paths_in_install_stanza
+         (Warning_emit.Context.source_dir_or_enable dir)
+         (fun () ->
+           User_message.make
+             ~loc
+             [ Pp.textf
+                 "The destination path %s begins with %s which will become an error in a \
+                  future version of Dune. Destinations of files in install stanzas \
+                  beginning with %s will be disallowed to prevent a package's installed \
+                  files from escaping that package's install directories."
+                 (String.maybe_quoted dst)
+                 (String.maybe_quoted Filename.parent_dir_name)
+                 (String.maybe_quoted Filename.parent_dir_name)
+             ]
+           |> Memo.return)
+     | `Always_error ->
+       User_error.raise
+         ~loc
+         [ Pp.textf
+             "The destination path %s begins with %s which is not allowed. Destinations \
+              in install stanzas may not begin with %s to prevent a package's installed \
+              files from escaping that package's install directories."
+             (String.maybe_quoted dst)
+             (String.maybe_quoted Filename.parent_dir_name)
+             (String.maybe_quoted Filename.parent_dir_name)
+         ])
 ;;
 
 module Expanded = struct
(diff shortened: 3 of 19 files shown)


test/blackbox-tests/test-cases/install/install-glob/install-glob-relative.t
```diff
@@ -43,6 +43,14 @@ Incorrect install stanza that would place files outside the package's install di
...
   files from escaping that package's install directories.
 
+Test that we can disable the warning:
+  $ cat >>dune-project <<EOF
+  > (warnings (escaping_paths_in_install_stanza disabled))
+  > EOF
+  $ dune build foo.install
+
+Now we remove disabling the warning
+  $ sed -i '$d' dune-project
```
gridbugs, line 53: The sed on macos doesn't allow `-i` without an extension to use for backup files. Use `sed -i.bak ...`.

src/dune_rules/warning.ml
```diff
@@ -0,0 +1,78 @@
...
+end
+
+type t =
+  { name : Name.t
+  ; since : Syntax.Version.t
+  ; default : Syntax.Version.t -> Config.Toggle.t
+  }
+
+let frozen = ref false
+let all : (Name.t, t) Table.t = Table.create (module Name) 12
```
gridbugs, line 25: 12?

src/dune_rules/warning.ml
```diff
@@ -0,0 +1,78 @@
...
+
+let create ~default ~name ~since =
+  if !frozen then Code_error.raise "Warnings may not be created after we've read them" [];
+  let name = Name.of_string name in
+  let t = { name; since; default } in
+  Table.add_exn all name t;
```
gridbugs, line 31: It should probably be a `Code_error` if the name already exists in the table.

src/dune_rules/file_binding.ml
```diff
@@ -30,41 +33,53 @@ let relative_path_starts_with_parent relative_path =
   | Some (first, _) -> String.equal first Filename.parent_dir_name
 ;;
 
+let warning =
```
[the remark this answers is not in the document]
gridbugs: Oh I see that we can't define this inside the function because all warnings need to be known about by the parser so they can be disabled in a config. In that case I think this should be renamed and moved to `warning.ml`.

src/dune_rules/warning.ml
```diff
@@ -0,0 +1,78 @@
...
+type t =
+  { name : Name.t
+  ; since : Syntax.Version.t
+  ; default : Syntax.Version.t -> Config.Toggle.t
+  }
+
+let frozen = ref false
+let all : (Name.t, t) Table.t = Table.create (module Name) 12
+
+let create ~default ~name ~since =
```
gridbugs, line 27: I think we can remove the need to dynamically build up a table of warnings which would simplify the code and remove the possibility of hitting the `"Warnings may not be created after we've read them"` error. I suggest we make this `create` function private which would require that all warnings be defined centrally in this file. Then we could statically create a list in this file containing all the warnings.

src/dune_rules/warning.mli
```diff
@@ -0,0 +1,26 @@
+(** General warning mechanism for dune rules *)
+
+open Import
+
+type t
+
+val create
```
gridbugs, line 7: Can you add a comment explaining the difference between `default` and `since`. I understand it now after reading the code but I was confused at first.

src/dune_rules/warning_emit.mli
```diff
@@ -0,0 +1,13 @@
...
+
+type context =
+  | Build of Path.Build.t
+  | Source of Source_tree.Dir.t
+  | Source_dir of Path.Source.t
+  | Dune_project of Dune_project.t
+
+(** [emit w ctx f] will call [f] if [w] is enabled in [ctx]. [f] should
+    generate the warning corresponding to [w] *)
+val emit : Warning.t -> context -> (unit -> unit Memo.t) -> unit Memo.t
```
gridbugs, line 13: If the function `f` is expected to always print a warning message then why not have the `emit` function take warning message directly and print it itself? To avoid computing the message in the common case when it's not needed it could take a thunk which returns the message.

src/dune_rules/warning.ml
```diff
@@ -0,0 +1,78 @@
+open Import
+
+module Name = struct
+  include String
+
+  include Stringlike.Make (struct
+      type t = string
+
+      let module_ = "Warning"
```
gridbugs, line 9: Should the module be `"Warning.Name"`?

src/dune_rules/warning_emit.ml
```diff
@@ -0,0 +1,31 @@
+open Import
+open Memo.O
+
+type context =
```
gridbugs, line 4: Can we use a more descriptive name that `context` here? I don't have a clear sense of how to choose the `context` argument to `emit`. Is the idea that every warning is because of something the user wrote in a dune{,-project,-workspace} file and that the path to that file (or maybe the directory containing the file?) will determine whether or not it's in a vendored project?

src/dune_rules/warning_emit.mli
```diff
@@ -0,0 +1,13 @@
+(** Emit warnings that respect dune's conventions *)
+
+open Import
+
+type context =
+  | Build of Path.Build.t
+  | Source of Source_tree.Dir.t
+  | Source_dir of Path.Source.t
```
gridbugs, line 8: Why do we need both `Source` and `Source_dir` constructors? Can you rename one of them or add a comment explaining when to use one vs the other?

src/dune_rules/warning.ml
```diff
@@ -0,0 +1,78 @@
...
+  { name : Name.t
+  ; since : Syntax.Version.t
+  ; default : Syntax.Version.t -> Config.Toggle.t
+  }
+
+let frozen = ref false
+let all : (Name.t, t) Table.t = Table.create (module Name) 12
+
+let create ~default ~name ~since =
+  if !frozen then Code_error.raise "Warnings may not be created after we've read them" [];
```
gridbugs, line 28: Can you add the warning name to the list passed to `Code_error.raise`. Also if I ran into this case as a developer I'd find it more helpful if the message said something like: `"Warning.create may not be called after warning settings have been parsed. Warnings should be created at the top level (not inside function definitions) in the file src/dune_rules/warnings.ml"`.

src/dune_rules/dune_project.mli
```diff
@@ -158,6 +158,7 @@ val accept_alternative_dune_file_name : t -> bool
 val strict_package_deps : t -> bool
 val cram : t -> bool
 val info : t -> Package.Info.t
+val warnings : t -> Warning.Settings.t
```
gridbugs, line 161: NP: Can we please call this `warning_settings` instead

gridbugs (changes requested):

src/dune_rules/warning_emit.mli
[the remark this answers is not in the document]
rgrinberg: The memoization is needed to look up what project we're in and if we're in a vendored directory. To raise without memoization, I suppose we could split the computation of if we're allowed to raise the warning from actually doing it. For example:

```
type context
val in_dir : Path.Source.t -> Warning.t -> context Memo.t
val emit : context -> (unit -> unit) -> unit
```

src/dune_rules/warning.mli
```diff
@@ -0,0 +1,26 @@
...
+  type t
+
+  val to_dyn : t -> Dyn.t
+  val empty : t
+  val decode : t Dune_sexp.Decoder.t
+  val active : t -> warning -> Syntax.Version.t -> Config.Toggle.t
+end
+
+(** Warn whenever [(name )]) is missing from the [dune-project] file *)
+val missing_project_name : t
```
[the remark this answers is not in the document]
rgrinberg: I'm still not sure what should be the convention here. Do you think it's better to define all warnings in one place, or just do it at near the use site?

src/dune_rules/warning_emit.mli
rgrinberg: Different parts of the code will find it easier to provide on value over another. I don't think I'm using all of these contexts though, so I will remove the ones that are unused.

src/dune_rules/warning_emit.ml
rgrinberg: Logically, a context is what decides if a particular warning is active. E.g, we can use it in the following `val warning_active -> Warning.t -> context -> bool Memo.t`.

src/dune_rules/warning.ml
rgrinberg: `add_exn` already raises a code error.

src/dune_rules/warning.ml
rgrinberg: Any number would do. We just need some initial table size.

src/dune_rules/dune_project.mli
rgrinberg: But the field in the project file is called `warnings`, so that would be a little more confusing imo. No strong preference though.

src/dune_rules/warning.ml
rgrinberg: Eventually, it will be possible for extensions to define their own warnings. E.g. ctypes, melange, coq should all be able to do so. For now, I'm fine with keeping it this way.

Although I don't see how we can remove the table even if everything is in one place. E.g. we'll have something like:
```
let w1 = ..
let w2 = ..
let all = [w1 ; w2 ; .. ]
```

And we'll need a check that no warnings were created that aren't inside of `all`.

src/dune_rules/warning_emit.mli
gridbugs: > Different parts of the code will find it easier to provide on value over another.

That makes sense. I think renaming the constructors to more closely match the types of their arguments will help to clarify their intent.

src/dune_rules/warning_emit.mli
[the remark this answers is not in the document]
gridbugs: I don't think that would make it any easier to raise in non-memo'd places in practice though. And I guess any part of the code that has enough information to create the context is probably going to be memo'd or at least it would probably be easy to spread memo to those parts of the code so I think it's fine to leave it for now and we can always split it up later if we have a use-case for non-memo'd `emit`.

src/dune_rules/warning.mli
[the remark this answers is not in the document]
gridbugs: My preference is to define them all in one place.

src/dune_rules/warning_emit.ml
gridbugs: Right but that doesn't help me choose which path to use for the context of a particular warning. If I want to raise a warning and there are multiple paths in scope which one should I use. If there's only one path in scope how do I know it's the right one for the warning?

src/dune_rules/dune_project.mli
gridbugs: Ah ok that makes sense.

src/dune_rules/warning_emit.mli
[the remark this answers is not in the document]
rgrinberg: Emitting a context without consulting the context is definitely wrong though. Otherwise, we aren't going to respect the settings in the dune-project file for example.

rgrinberg:
@gridbugs should be better now. I've also added an API to make this usable when inside `Dune_lang.Decoder`. Many warnings come from there.

src/dune_rules/warning_emit.ml
rgrinberg: It should be clear now.

src/dune_rules/warning_emit.mli
```diff
@@ -0,0 +1,31 @@
+(** Emit warnings that respect dune's conventions *)
+
+open Import
+
+module Context : sig
+  (** a context decides whether a warning should
+      be emitted based on the settings inside the project and if the directory
+      we're is vendored *)
```
gridbugs, line 8:
```suggestion
      we're in is vendored *)
```

doc/dune-files.rst
```diff
@@ -689,7 +689,21 @@ map_workspace_root
...
+.. _warnings:
+
+warnings
+--------
+
+.. dune:stanza:: warnings
+
+   .. versionadded:: 3.11
+
+   Configure dune warnings for the project.
```
christinerose, line 701: Should this be Dune or `dune`?

christinerose (approved):

PR #8449, opened 2023-08-21
fix(tui): do not set stdin as nonblocking

rgrinberg:
That's a litte dangerous as the other end of this pipe isn't dune and
this might not handle this right.


Alizter (approved):
I also ended up doing this.

PR #8452, opened 2023-08-21
fix: handle errors in threaded console correctly

rgrinberg:
don't try to write errors to the console when the console is dead


src/dune_util/log.ml
```diff
@@ -40,16 +40,26 @@ let init ?(file = File.Default) () =
 let init_disabled () = Fdecl.set t None
 let t () = Fdecl.get t
 
+let log_oc oc msg =
```
Alizter, line 43: If you switch the order of these args you won't have to write a lambda later twice.

Issue #8291, opened 2023-07-28
Please honor CFLAGS


glondu:
## Desired Behavior

Dune should take into account the CFLAGS environment variable when compiling C code (similarly with CXXFLAGS and C++).

## Rationale

Using these environment variables is standard in build systems (e.g. make).

In Debian, we use these variables to have some options used by many (if not all) packages at once.

In particular, one of the options set by Debian is the file prefix map: `-ffile-prefix-map=/tmp/foo=.`, which is important for reproducibility.

MisterDA:
Another use-case for this just popped: GCC 15 defaults to C23 and has added many new warnings and errors, which can make some C files fail to compile (for instance, OCaml 4.14.1 cannot build with GCC 15, but 4.14.2 is fixed).
If `CFLAGS=-std=c17` is set in the environment, build systems such as Make (with the default rules), CMake, and Meson will recognize it and pass it to the C compiler invocation. The files can then be built with the C 17 semantics.
As Dune doesn't pick this variable, any project that uses `foreign_stubs` will be built with the C compiler default C standard.

PR #8464, opened 2023-08-22
Use CFLAGS and CXXFLAGS from environment to compile foreign files

glondu:
See https://github.com/ocaml/dune/issues/8291


src/dune_rules/foreign_rules.ml
```diff
@@ -127,16 +132,25 @@ let build_c
   let use_standard_flags = Dune_project.use_standard_c_and_cxx_flags project in
   let base_flags =
     match kind with
-    | Cxx -> Fdo.cxx_flags ctx
+    | Cxx ->
+       List.concat
+         [ get_env_flags "CXXFLAGS"
```
Alizter, line 137: You can use `Env.get` instead here. The `Env.t` value should come from ctx.env.

We should probably also guard this behaviour behind dune lang. cc @rgrinberg wdyt?

glondu: > Do you use a custom workspace file when building dune packages in debian?

I was not familiar with this notion, so no. The way most dune packages are built in debian is defined [here](https://salsa.debian.org/ocaml-team/dh-ocaml/-/blob/master/debhelper/ocaml_dune.pm)... Sorry, it's Perl, but should be understandable: the interesting parts are the last four functions (`build`, `test`, `install` and `clean`).

glondu: > Guarding it behind a command line flag is fine.

I've implemented that in a separate commit. I'm wondering if this is the right move, though. The feature could be useful for installing opam packages, too, and there is no way to change the command used there...?

Alizter: > What does "wdyt" mean?
"What do you think"

MisterDA:
We could also check the `CPPFLAGS` variable for flags specific to the C/C++ preprocessor. It is (at least) understood by [GNU Make](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html#index-CPPFLAGS).

rgrinberg:
@MisterDA Should we add `LDFLAGS` as well?

I discussed this with the team and it seems like this feature is the right way to go. To proceed with the PR:

* Please add support for CPPFLAGS
* Add some tests exercising this feature (see test/blackbox-tests/test-cases for examples)
* Add a change log entry in doc/changes

rgrinberg:
@MisterDA since you're interested, do you mind reviewing this one?

MisterDA:
> Should we add `LDFLAGS` as well?

Yes! Looking for inspiration, Arch Linux' [makepkg](https://wiki.archlinux.org/title/Makepkg) (the program used to build custom packages) allows user to set `CPPFLAGS`, `CFLAGS`, `CXXFLAGS`, and `LDFLAGS` globally.
`LDFLAGS` can be used to pick a linker, they show `LDFLAGS="... -fuse-ld=mold"`.

> Since you're interested, do you mind reviewing this one?

It looks good, but some tests showcasing this would be nice, indeed.

glondu:
I added support for `CPPFLAGS`, removed the `--use-env-flags` command-line argument, and rebased on current main.

rgrinberg:
Thanks @glondu. What do you think about supporting LDFLAGS as well?

glondu:
> What do you think about supporting LDFLAGS as well?

I'm not sure about this one. According to my interpretation of [make's built-in rules](https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html), they should be used in linking rules (and I would say all of them, even pure OCaml ones), which I haven't located yet (I'm new to dune's codebase).

FYI, in Debian package building environment, LDFLAGS is set to `-Wl,-z,relro` which is used for [hardening](https://wiki.debian.org/Hardening#DEB_BUILD_HARDENING_RELRO_.28ld_-z_relro.29).

rgrinberg:
Please re-open if you intend to finish this PR

PR #6966, opened 2023-01-30
Add support for eager watch mode for `dune exec`

gridbugs:
This is the second attempt at adding this feature. The problem with the first attempt was that replacing the call to `Scheduler.go` with `Scheduler.go_with_rpc_server_and_console_status_reporting` caused occasional non-deterministic seg faults on macos. There's some info about the problem on the PR which reverts the previous attempt at adding this feature: https://github.com/ocaml/dune/pull/6867. At the time of writing we don't know the cause of this problem.

The difference this time around is that we maintain the call to `Scheduler.go` when running `dune exec` not in watch mode, and only invoke `Scheduler.go_with_rpc_server_and_console_status_reporting` when running in watch mode. There is some additional refactoring done to make this split more ergonomic. We may see seg faults on macos when running exec in watch mode but at least we won't introduce the potential for seg faults into the existing use case of running exec not in watch mode (assuming of course that there is a causal link between `go_with_rpc_server_and_console_status_reporting` and the seg fault on macos, which is not necessarily the case).

Signed-off-by: Stephen Sherratt <stephen@sherra.tt>

PR #6035, opened 2022-08-03
dune exec: support pform syntax

emillon:
This supports things like `dune exec time %{bin:e}`.

The syntax is consistent with what support in `dune build` and backwards compatible in cases where no arguments start with `%`.

The resolution mechanism is slightly different for the program and the rest of the arguments:

- the program is always considered a possible dependency, either in pform syntax (`%{bin:e}` or in string syntax (`./path/to/e`, `_build/default/path/to/e`).
- arguments are only interpreted as dependencies if they are in pform syntax.

Closes #2691

PR #8474, opened 2023-08-23
dune exec: support pform syntax (rebased)

Leonidas-from-XIV:
Rebased version of #6035 to deal with #6966. Meant to be pushed on top of #6035 after the new issues have been resolved but for now it crashes.


bin/exec.ml
```diff
@@ -290,9 +290,13 @@ module Exec_context = struct
     let* get_path_and_build_if_necessary = get_path_and_build_if_necessary in
     let* env = env in
     let* expand_cli_item = expand_cli_item in
-    let* path = Build_system.run_exn get_path_and_build_if_necessary in
+    let* path, args =
+      Build_system.run_exn (fun () ->
+        let path = get_path_and_build_if_necessary () in
+        let args = args |> List.map ~f:expand_cli_item |> Memo.all in
```
rgrinberg, line 296: I think we have `Memo.concat_map`

Leonidas-from-XIV:
@rgrinberg I've added a test that pforms outside of the program name are expanded and built and the expanding works fine, but I am not entirely sure why dune builds `called` as there is no code to actually call the `get_path_and_build_if_necessary` on `args`. It seems like there is some reason `called` is getting built but I don't know how to avoid this.

rgrinberg:
> It seems like there is some reason called is getting built but I don't know how to avoid this.

Why do you want to avoid this? Seems like it's correct to build it.

The reason why it's getting build is that the `Expander` will add the correct dependencies on every form that it will expand.

bin/target.ml
```diff
@@ -165,7 +165,24 @@ let resolve_path path ~(setup : Dune_rules.Main.build_system)
...
+      ctx.Context.build_dir
+      (String.concat ~sep:Filename.dir_sep root.to_cwd)
+  in
+  let* expander = Action_builder.of_memo (Dune_rules.Super_context.expander sctx ~dir) in
+  let expander = Dune_rules.Dir_contents.add_sources_to_expander sctx expander in
+  let+ s = Dune_rules.Expander.expand_str expander sv in
+  root.reach_from_root_prefix ^ s
+;;
+
+let expand_path_from_root
```
rgrinberg, line 180: Can we somehow unify these functions?

Leonidas-from-XIV:
> Why do you want to avoid this? Seems like it's correct to build it.

Of course it is, but I did not understand that the expander is doing that. I thought I would need to call `get_path_and_build_if_necessary` to build it so I was surprised the test kept passing anyway despite me not actually triggering builds. As such I can probably just remove the function altogether and just use `expand_arg` instead.

bin/target.ml
Leonidas-from-XIV: Yup, updating the arguments make the `expand_path_from_root'` function obsolete.

Leonidas-from-XIV:
@rgrinberg I've addressed the comments, can you take a look if it is in a shape to merge?

rgrinberg (approved):

doc/changes/6035.md
```diff
@@ -0,0 +1,3 @@
+- `dune exec`: support syntax like `%{bin:program}`. This can appear anywhere
+  in the command line, so things like `dune exec time %{bin:program}` now work.
+  (#6035, fixes #2691, @emillon)
```
rgrinberg, line 3: I think you should add your own name here as well.

PR #8477, opened 2023-08-23
refactor(ansi_color): rename components

Alizter:
The "components" of RGB24 should be consistent with our `make` constructor. This will also let us include named color constants here later on.


rgrinberg:
If the components should be consistent, we should just us the full names in `make` as well. Not a fan of the shortened one letter names.

Alizter:
@rgrinberg I'm inclined to agree, but reading the API they look like colour constants. We might even add colour constants in the future.

rgrinberg:
If you want constructor, you could always add some submodule to list them.

Alizter:
Fair enough, dropping this now.

PR #8478, opened 2023-08-23
Create flakehub-publish-tagged.yml

Alizter:
We can publish our Nix flake to flakehub when there is a release.

https://flakehub.com/docs

Basically people who are interested can update their nix flake from flake hub. We just have to add this action and never think about it again.


Alizter:
But on second thoughts I'm not too sure about this.

PR #8479, opened 2023-08-24
feature(monitor): add --quit-on-disconnect option

### LINKED ISSUE #8425: Add --auto-poll mode to dune monitor
You could add a `--auto-poll` mode that tries to re-establish the connection if it's dead.

_Originally posted by @rgrinberg in https://github.com/ocaml/dune/pull/8152#discussion_r1296394772_

Alizter:
We add a --quit-on-disconnect option to Dune monitor.

Dune monitor will now automatically reconnect to an RPC server when the connection is lost. If not server is found, it will wait around until one appears.

The new --quit-on-disconnect option lets dune monitor quit rather than attempt to reconnect.

- fix https://github.com/ocaml/dune/issues/8425

[Screencast from 2023-08-24 11-37-52.webm](https://github.com/ocaml/dune/assets/8614547/ef23809b-afb9-438b-8a82-d8d203d0f690)

### DIFF
--- a/bin/monitor.ml
+++ b/bin/monitor.ml
@@ -206,7 +206,7 @@ let render_loop ~(event : Event.t Fiber_event_bus.t) =
     >>= function
     | `Closed ->
       Console.print_user_message
-        (User_error.make [ Pp.textf "Lost connection to server, exiting..." ]);
+        (User_error.make [ Pp.textf "Lost connection to server." ]);
       Fiber.return ()
     | `Next event ->
       let update = State.update state event in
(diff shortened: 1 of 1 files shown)


bin/monitor.ml
```diff
@@ -219,21 +219,59 @@ let render_loop ~(event : Event.t Fiber_event_bus.t) =
...
+let run_fetchers ~event ~client =
+  let module Sub = Dune_rpc_private.Public.Sub in
+  Fiber.all_concurrently_unit
+    [ fetch_loop ~event ~client ~f:(fun x -> Event.Jobs x) Sub.running_jobs
+    ; fetch_loop ~event ~client ~f:(fun x -> Event.Progress x) Sub.progress
+    ; fetch_loop ~event ~client ~f:(fun x -> Event.Diagnostics x) Sub.diagnostic
+    ]
+;;
+
+let repeat_while_counting ~init ~f =
```
rgrinberg, line 231: This can be implemented on top of `Fiber.repeat_while`.

bin/monitor.ml
```diff
@@ -219,21 +219,59 @@ let render_loop ~(event : Event.t Fiber_event_bus.t) =
...
-    ~f:(fun client ->
-      let event = Fiber_event_bus.create () in
-      let module Sub = Dune_rpc_private.Public.Sub in
-      Fiber.all_concurrently_unit
-        [ render_loop ~event
-        ; fetch_loop ~event ~client ~f:(fun x -> Event.Jobs x) Sub.running_jobs
-        ; fetch_loop ~event ~client ~f:(fun x -> Event.Progress x) Sub.progress
-        ; fetch_loop ~event ~client ~f:(fun x -> Event.Diagnostics x) Sub.diagnostic
-        ])
+  let* `Rejoin =
```
rgrinberg, line 262: What about just returning `unit` here?

bin/monitor.ml
Alizter: The issue is that `Fiber.repeat_while` returns `unit` when given `None`, but I want to get the value of `Where.get ()`. I could use it, but then I would have to repeat `Where.get`.

bin/monitor.ml
```diff
@@ -219,21 +219,59 @@ let render_loop ~(event : Event.t Fiber_event_bus.t) =
...
+      match Dune_rpc_impl.Where.get () with
+      | Some p -> Fiber.return (`Ok p)
+      | None when not auto_poll -> User_error.raise [ Pp.text "RPC server not running." ]
+      | None ->
+        let dots =
+          match i mod 4 with
+          | 0 -> Pp.nop
+          | 1 -> Pp.verbatim "."
+          | 2 -> Pp.verbatim ".."
+          | 3 -> Pp.verbatim "..."
```
rgrinberg, line 253: How about just `Pp.verbatim (String.make '.' (i mod 4))`

bin/monitor.ml
```diff
@@ -219,21 +219,59 @@ let render_loop ~(event : Event.t Fiber_event_bus.t) =
...
+          | _ -> assert false
+        in
+        Console.Status_line.set
+          (Console.Status_line.Live
+             (fun () -> Pp.seq (Pp.verbatim "Waiting for RPC server") dots));
+        Scheduler.sleep 0.5 >>> Fiber.return `Try_again)
```
rgrinberg, line 259: Let's do 0.30 here. 0.5 feels a bit slow

bin/monitor.ml
rgrinberg: You would only return `None` when `auto_poll = false`, right? So I don't see why `Where.get` would be repeated.

Alizter: Oh I see how to use it now, will do. I wasn't thinking about the whole function but just the where call.

PR #8481, opened 2023-08-24
fix: do not swallow [Async_io] errors

rgrinberg:
Errors would be swallowed whenever the select thread would raise while
the mutex was unlocked. The subsequent [finally] that tries to unlock
would fail, and we'd hdie the original error. We ignore the error of
unlocking the mutex twice now.

### DIFF
--- a/src/dune_async_io/async_io.ml
+++ b/src/dune_async_io/async_io.ml
@@ -167,6 +167,11 @@ let maybe_cancel table fd acc =
     drain_cancel q acc
 ;;
 
+(* This exception is raised in [select_loop] whenever it fails and the mutex is
+   unlocked. This makes sure that we don't unlock it again after [select_loop]
+   finishes *)
+exception Unlocked of Exn_with_backtrace.t
+
 let rec select_loop t =
   (match t.to_close with
    | [] -> ()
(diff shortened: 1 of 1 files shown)


src/dune_async_io/async_io.ml
```diff
@@ -224,7 +224,11 @@ let start t =
   let module Scheduler = (val t.scheduler : Scheduler) in
   Scheduler.spawn_thread (fun () ->
     Mutex.lock t.mutex;
-    Exn.protect ~f:(fun () -> select_loop t) ~finally:(fun () -> Mutex.unlock t.mutex))
+    Exn.protect
+      ~f:(fun () -> select_loop t)
+      ~finally:(fun () ->
+        try Mutex.unlock t.mutex with
```
dkalinichenko-js, line 230: Let's add a comment explaining why we are catching the error here.

rgrinberg: Something about this code didn't sit right with me and I rewrote it without this ugliness. Take another look.

src/dune_async_io/async_io.ml
```diff
@@ -201,6 +206,9 @@ let rec select_loop t =
      | exception Unix.Unix_error (Unix.(EINTR | EAGAIN), _, _) ->
        Mutex.lock t.mutex;
        select_loop t
+     | exception exn ->
+       let exn = Exn_with_backtrace.capture exn in
+       raise (Unlocked exn)
```
Alizter, line 211: Should this be `raise_notrace`?

rgrinberg: Sure, why not. Doesn't make much of a difference.

dkalinichenko-js (approved):

PR #8486, opened 2023-08-24
pkg: add `run-with-conditional-terms` action

gridbugs:
This adds an action tha is similar to `run` except that each term (including the first one) has an optional predicate which can filter out terms. This is added to match the behaviour of opam commands where each term can be optional based on a filter.

### DIFF
--- a/src/dune_lang/action.ml
+++ b/src/dune_lang/action.ml
@@ -175,6 +175,7 @@ type t =
   | Patch of String_with_vars.t
   | Substitute of String_with_vars.t * String_with_vars.t
   | Withenv of String_with_vars.t Env_update.t list * t
+  | Run_with_conditional_terms of (Blang.t option * String_with_vars.t) list
 
 let is_dev_null t = String_with_vars.is_pform t (Var Dev_null)
 
@@ -375,6 +376,23 @@ let decode_pkg =
         >>> let+ ops = enter (repeat Env_update.decode)
             and+ t = t in
             Withenv (ops, t) )
+    ; ( "run-with-conditional-terms"
+      , Syntax.since Pkg.syntax (0, 1)
+        >>>
+        let decode_term_with_condition =
+          enter
+            (let+ condition = Blang.decode
+             and+ term = String_with_vars.decode in
+             Some condition, term)
+        in
+        let decode_term_without_condition =
+          let+ term = String_with_vars.decode in
+          None, term
+        in
+        let+ terms =
+          repeat (decode_term_with_condition <|> decode_term_without_condition)
+        in
+        Run_with_conditional_terms terms )
     ]
   in
   Decoder.fix @@ fun t -> Decoder.sum (cstrs_dune_file t @ cstrs_pkg t)
@@ -430,6 +448,13 @@ let rec encode =
   | Substitute (i, o) -> List [ atom "substitute"; sw i; sw o ]
   | Withenv (ops, t) ->
     List [ atom "withenv"; List (List.map ~f:Env_update.encode ops); encode t ]
+  | Run_with_conditional_terms terms ->
+    List
+      (atom "run-with-conditional-terms"
+       :: List.map terms ~f:(fun (condition, term) ->
+         match condition with
+         | None -> sw term
+         | Some condition -> List [ Blang.encode condition; sw term ]))
 ;;
 
 (* In [Action_exec] we rely on one-to-one mapping between the cwd-relative paths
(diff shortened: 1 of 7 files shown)


Alizter:
Why can't we have conditional pforms? Something like `%{if:blang:foo}`? That way we don't need a new action.

rgrinberg:
Why do we need a separate name for this action? It seems like the same thing as `run` except without any predicates.

gridbugs:
> Why do we need a separate name for this action? It seems like the same thing as `run` except without any predicates.

That's true! No technical reason we couldn't combine them though I find the behaviour of `run-with-conditional-terms` a little strange though, especially the fact that the first term can be conditional and dynamically choose which term is treated as the program. Right now `run`'s behaviour is simple and I think will match the expectation of many users who just want to run a command and I'd worry that if we add optional behaviour (especially optional behaviour that is IMO un-intuitive) then that will make `run` harder for users to understand in general.

rgrinberg:
Fully agree with you that this is unintuitive. But we're not really eliminating it by introducing a new action, just moving it to its own action. We'll still run into the same problems when we unify these actions with normal dune files. Concretely, we'll find it awkward to explain to users why we've added a `run-with-conditional-terms` in a way that doesn't include the phrase "because opam". 

I'd like us to break this behavior into smaller and simpler components and get back opam's behavior by composing them. Even if it's a little more work upfront.

gridbugs:
Ok so that sounds like rather than adding a new action as I've done here you're suggesting adding a general mechanism for ignoring terms? Such as adding a new DSL to dune that allows (at least) filtering out terms based on predicates. (e.g. `(run (if <cond1> <term1>) (if cond2> <term2>) <uncoditional-term3>)`). Is that correct?

I don't see the problem with documenting `run-with-conditional-terms` in a way that mentions opam. I don't think this action will be used for anything other than handling opam actions. I expect people will rarely need this feature but when they find themselves trying to emulate opam or they're reading a lockfile for an opam package they can read a single piece of documentation and know exactly what it does, and if they don't use this feature then they don't need to know about it at all to use dune.

Adding a general feature for eliminating terms based on predicates would mean that we could just use the `run` action to emulate opam but I worry that people will start using it in other places and the feature will become necessary to perform other tasks besides opam emulation and it will become another general concept that people need to learn in order to read and write dune files. I'm scared we're going to start seeing this feature in places like the `(libraries ...)` field or the `(targets ...)` field and if we explicitly ban it in those places then users would have to remember which places it's allowed and which places it's not. I'd prefer to add a self-contained feature that's a little strange rather than a more elegant general feature with the potential to complicate parts of dune that used to be simple.

rgrinberg:
>  Is that correct?

I don't have a concrete suggestion that I like. I do see a few possible alternatives. But I would like to see a construct that is useful enough to be used outside of lock files for when we unify the action languages of lock files and regular dune files.

> I don't see the problem with documenting run-with-conditional-terms in a way that mentions opam

One of the main advantages of implementing package management on top of dune is that it allows our users to stick to understanding a single tool to manage their OCaml projects. This is the main reason why we didn't go along with improving opam to support a project local workflow better. To achieve this goal, we need to talk about this feature in self contained way as possible.

> I don't think this action will be used for anything other than handling opam actions. I expect people will rarely need this feature but when they find themselves trying to emulate opam or they're reading a lockfile for an opam package they can read a single piece of documentation and know exactly what it does, and if they don't use this feature then they don't need to know about it at all to use dune.

I think your assessment will be true in the short term, but not in the long term. In the beginning, yes our lock files are fairly limited and can only be generated. Later, much more will be possible through overrides, adding custom packages, generating lock files dynamically using rules, etc.

> Adding a general feature for eliminating terms based on predicates would mean that we could just use the run action to emulate opam but I worry that people will start using it in other places and the feature will become necessary to perform other tasks besides opam emulation and it will become another general concept that people need to learn in order to read and write dune files. I'm scared we're going to start seeing this feature in places like the (libraries ...) field or the (targets ...) field and if we explicitly ban it in those places then users would have to remember which places it's allowed and which places it's not. I'd prefer to add a self-contained feature that's a little strange rather than a more elegant general feature with the potential to complicate parts of dune that used to be simple.

I share the concern of complicating the dune language, but I think this is swinging it too much in the opposite direction. First, these extensions only apply to the action language and not to the dune language in general - so these terms cannot appear in `libraries` or `targets` for that reason alone. Second, we already have a massive amount of features in the action language: piping, concurrency, exit codes etc. Conditional execution is a natural addition. When we make the action language more powerful, we achieve two goals:

* We speed up builds for free because they longer need to go through an intermediate shell script for execution.
* We gain portability by allowing users to avoid using the non portable shell for actions. In fact, this is how we managed to make packages built with dune quite portable in the first place.

In summary, there's no need to shy away from extending the action language. Even if we make a mistake and introduce something we'll later regret, it will still be better than whatever alternative users would come up with. Of course we also have option to correct our mistakes in the future and evolve the action language with versioning.

gridbugs:
Ok that makes sense. Thanks for the explanation. I hadn't considered adding this feature to the action language specifically. I'll have a play around with adding conditional terms via a more general feature so we don't need a dedicated action for opam commands.

PR #8487, opened 2023-08-24
chore: update flakes

rgrinberg:

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -21,11 +21,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1689068808,
-        "narHash": "sha256-6ixXo3wt24N/melDWjq70UuHQLxGV8jZvooRanIHXw0=",
+        "lastModified": 1692799911,
+        "narHash": "sha256-3eihraek4qL744EvQXsK1Ha6C3CR7nnT8X2qWap4RNk=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "919d646de7be200f3bf08cb76ae1f09402b6f9b4",
+        "rev": "f9e7cf818399d17d347f847525c5a5a8032e4e44",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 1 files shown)


Alizter:
Doesn't work for me, but not special to this PR. I get:
```
/nix/store/dfkvqi6182s94jdf123z3rqn0i84mr1c-git-2.41.0/bin/git --no-pager diff --no-index --color=always -u _build/.sandbox/8fa9c1733b6d571a6d95fc2cac32f4b0/default/test/react.t _build/.sandbox/8fa9c1733b6d571a6d>
diff --git a/_build/.sandbox/8fa9c1733b6d571a6d95fc2cac32f4b0/default/test/react.t b/_build/.sandbox/8fa9c1733b6d571a6d95fc2cac32f4b0/default/test/react.t.corrected
index a6945b6..4fe7caf 100644
--- a/_build/.sandbox/8fa9c1733b6d571a6d95fc2cac32f4b0/default/test/react.t
+++ b/_build/.sandbox/8fa9c1733b6d571a6d95fc2cac32f4b0/default/test/react.t.corrected
@@ -38,8 +38,8 @@ Demonstrate how to use the React JSX PPX
         ocamlc .ppx/df8d19d63a600c50cc2fede8aa971a95/dune__exe___ppx.{cmi,cmo}
       ocamlopt .ppx/df8d19d63a600c50cc2fede8aa971a95/build_info__Build_info_data.{cmx,o}
-         refmt x.re.ml
       ocamlopt .ppx/df8d19d63a600c50cc2fede8aa971a95/dune__exe___ppx.{cmx,o}
+         refmt x.re.ml
       ocamlopt .ppx/df8d19d63a600c50cc2fede8aa971a95/ppx.exe
            ppx x.re.pp.ml
           melc .output.mobjs/melange/melange__X.{cmi,cmj,cmt}
Done: 99% (7564/7573, 9 left) (jobs: 9)^M
```
I've tried everything I can think of to overlay `doCheck = false` to melange but nothing seems to work.

rgrinberg:
> does the flake work for you? Maybe it is something wrong with my machine.

It works on my mac, but it does fail on Linux.

I think the problem is with the test because it relies on `--display short` to be deterministic which isn't the case.

@anmonteiro one easy way to fix this would be to add `2>&1 | sort` to all this output.

PR #8488, opened 2023-08-24
bench: no longer pin local packages

gridbugs:
Previously ew needed to pin packages defined in the dune repo such as `stdune` and `dune-rpc` as they are used by the benchmark runner and the benchmark runner was dependent on unreleased bugfixes in some of these libraries, but now these fixes have been released.


gridbugs:
cc @Alizter

PR #8490, opened 2023-08-24
test(pkg): add dependency on git where needed

rgrinberg:


Leonidas-from-XIV (approved):
Thanks, that's a good point.

PR #8499, opened 2023-08-24
test(dune_console): add escaping for clearer tests

Alizter:
We add some custom escaping functions that remove `\r` and replace it with `\\r\n` so that we can see how it is used by the console.

### DIFF
--- a/src/dune_console/progress.ml
+++ b/src/dune_console/progress.ml
@@ -3,25 +3,35 @@ open Stdune
 module No_flush = struct
   let status_line = ref Pp.nop
   let start () = ()
-  let status_line_len = ref 0
+  let status_line_height = ref 0
+  let ansi_erase = "\r\027[2K"
+  let ansi_up = "\027M"
+  let clear_code_length = String.length ansi_erase + String.length ansi_up
 
   let hide_status_line () =
-    if !status_line_len > 0 then Printf.eprintf "\r%*s\r" !status_line_len ""
+    if !status_line_height > 0
+    then (
+      let buf = Buffer.create (clear_code_length * !status_line_height) in
+      Buffer.add_string buf ansi_erase;
+      for _ = 1 to !status_line_height do
+        Buffer.add_string buf (ansi_up ^ ansi_erase)
+      done;
+      Buffer.contents buf |> Printf.eprintf "%s")
   ;;
 
-  let show_status_line () = if !status_line_len > 0 then Ansi_color.prerr !status_line
+  let show_status_line () = if !status_line_height > 0 then Ansi_color.prerr !status_line
 
   let set_status_line = function
     | None ->
       hide_status_line ();
       status_line := Pp.nop;
-      status_line_len := 0
+      status_line_height := 0
     | Some line ->
       let line = Pp.map_tags line ~f:User_message.Print_config.default in
-      let line_len = String.length (Format.asprintf "%a" Pp.to_fmt line) in
       hide_status_line ();
       status_line := line;
-      status_line_len := line_len;
+      status_line_height
+        := Format.asprintf "%a" Pp.to_fmt line |> String.split_lines |> List.length;
       show_status_line ()
   ;;
 
--- a/test/expect-tests/dune_console/dune_console_tests.ml
+++ b/test/expect-tests/dune_console/dune_console_tests.ml
@@ -1,14 +1,18 @@
-(** Creation of Dune_console is stateful so we introduce a new module for each test. *)
+open Stdune
+module Re = Dune_re
+
+let escape str =
+  str |> String.split_lines |> List.map ~f:String.escaped |> List.iter ~f:prerr_endline
+;;
+
+(* Creation of Dune_console is stateful so we introduce a new module for each test. *)
 module New () = Dune_console
 
 module type New_console = module type of Dune_console
 
-(** Throughout this file we use regular [" "] strings rather than [{| |}] because we want
-    to observe the escape sequences.
-
-    In order to keep tests accross different backends consistent, we create some generic
-    test scripts here that take the created [Console]. We then test these for each
-    backend. *)
+(* In order to keep tests accross different backends consistent, we create some generic
+   test scripts here that take the created [Console]. We then test these for each
+   backend. *)
 
 let test_basic_usage (module Console : New_console) =
   Console.printf "Hello World!";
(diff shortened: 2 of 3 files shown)


test/expect-tests/dune_console/dune_console_tests.ml
```diff
@@ -74,19 +135,27 @@ let%expect_test "basic usage" =
...
   test_status_line_clearing (module Console);
+  escape [%expect.output];
   [%expect
-    "Here is a status line\r                     \rHere is another\r               \r"]
+    {| 
+Here is a status line\r
+                     \r
+Here is another\r
+               \r
+ |}]
```
Alizter, line 158: The extra newline after the return shows that we are going back the correct amount. We need two returns when updating the status. The first so that we can clear the cells and the second after we are done.

test/expect-tests/dune_console/dune_console_tests.ml
```diff
@@ -95,10 +164,12 @@ let%expect_test "Status line clearing with wrapping." =
...
-    "\n\
-     This status line is a problem because of the fact that it is especially long and \
-     therefore will not be cleared \
-     properly.\r                                                                                                                        \
-     \rHere is another\r               \r "]
+    {| 
+This status line is a problem because of the fact that it is especially long and therefore will not be cleared properly.\r
+                                                                                                                        \r
+Here is another\r
+               \r
```
Alizter, line 173: It's interesting that we don't wrap status lines no matter what. The issue is that the console will do it for us in most cases and we won't be able to track the number of clears needed.

Alizter: Actually putting a `Pp.hovbox`, gets it wrapping again and we can see the incorrect clearing sequence.

Alizter:
I have a fix for clearing multiline status lines that I will publish after this PR is merged.

test/expect-tests/dune_console/dune_console_tests.ml
```diff
@@ -30,42 +27,107 @@ let test_status_line_clearing_with_wrapping (module Console : New_console) =
...
-           and therefore will not be cleared properly."));
+       (Pp.hovbox
+        @@ Pp.text
+             "This status line is a problem because of the fact that it is especially \
+              long and therefore will not be cleared properly."));
   Status_line.set (Status_line.Constant (Pp.text "Here is another"));
   Status_line.clear ()
 ;;
 
+module Escape = struct
```
rgrinberg, line 38: Why do we need this all this code? Why can't we just use `String.escaped` or something if it's just for tests?

Alizter: I don't want to escape all characters such as newlines and I don't see a way of making that modular.

rgrinberg: Why don't you want to escape the newlines?

Even if you don't, how about you split it by lines and print a list of lines? Or at least implement the escaping with `Re` to make it simple. You took a lot of code from our escaping functions that are bug prone and designed for performance. I don't want code like that in the test suite.

Alizter: I've implemented the escaping using Re now.

Alizter: I am purposefully not escaping newlines because I want them rendered in the test output. It makes comparing lines easier.

rgrinberg: Except you can do that with `String.split_on_char ~on:'\n' |> List.map ~f:String.escape` already.

test/expect-tests/dune_console/dune_console_tests.ml
```diff
@@ -95,10 +120,13 @@ let%expect_test "Status line clearing with wrapping." =
...
-     This status line is a problem because of the fact that it is especially long and \
-     therefore will not be cleared \
-     properly.\r                                                                                                                        \
-     \rHere is another\r               \r "]
+    {| 
+This status line is a problem because of the fact that it is especially long
+and therefore will not be cleared properly.\r
+                                                                                                                        \r
+Here is another\r
+               \r
```
Alizter, line 130: @rgrinberg The most important part of the test is this. In our progress terminal, we have this logic for clearing the status line. This assumes that the status line is a single line since it does a return, prints space for the length of the line and then returns again. Doesn't work when there are multiple lines.

Since we are not in a dumb terminal, there is no reason for us to have this limitation in general. Therefore my fix will use the correct ansi control characters to clear lines (including those before) rather than having to calculate spaces. This logic might work on a dumb terminal, but the implementation there is to just print the status line and not clear it.

test/expect-tests/dune_console/dune_console_tests.ml
rgrinberg: Just use `String.escaped`. The current test is barely readable to me. I'd rather look at escape codes than count spacing.

Alizter: Since I am going to get rid of the spacing in the fix, can we keep it this way for the time being? The reason I am asking for it is because it really isn't easy to see if the spacing is correct when it is not on a new line.

Alizter: Also, I am not putting the spacing in myself, this is done by the console. The escape chars are exactly `\r`.

rgrinberg: Okay, then push your fix to the current branch then. You're saying it isn't to see, well maybe you should compute the property you're trying to test instead of making the reader eyeball it.

Alizter: It won't make sense to compute the number of spaces later since we won't be clearing lines that way.

test/expect-tests/dune_console/dune_console_tests.ml
```diff
@@ -95,10 +113,10 @@ let%expect_test "Status line clearing with wrapping." =
...
+  escape [%expect.output];
   [%expect
-    "\n\
-     This status line is a problem because of the fact that it is especially long and \
-     therefore will not be cleared \
-     properly.\r                                                                                                                        \
-     \rHere is another\r               \r "]
+    {| 
+This status line is a problem because of the fact that it is especially long
+and therefore will not be cleared properly.\r\027[2K\027M\r\027[2K\027M\r\027[2KHere is another\r\027[2K\027M\r\027[2K
```
Alizter, line 120: @rgrinberg Do you see what I mean about adding the space after `\r` making this easier to read?

Alizter:
@rgrinberg Are you happy with having the tests and the fix in the same PR? If so I will update the PR description.

src/dune_console/progress.ml
```diff
@@ -3,25 +3,35 @@ open Stdune
...
-    if !status_line_len > 0 then Printf.eprintf "\r%*s\r" !status_line_len ""
+    if !status_line_height > 0
+    then (
+      let buf = Buffer.create (clear_code_length * !status_line_height) in
+      Buffer.add_string buf ansi_erase;
+      for _ = 1 to !status_line_height do
+        Buffer.add_string buf (ansi_up ^ ansi_erase)
+      done;
+      Buffer.contents buf |> Printf.eprintf "%s")
   ;;
```
rgrinberg, line 20: should be `prerr_string` along with a flush.

Also, there's no need for the intermediate buffer. `Stdlib.stderr` is already buffered. Just print directly.

test/expect-tests/dune_console/dune_console_tests.ml
rgrinberg: It isn't easy enough. As I've mentioned, this check is complex enough that it can't be done visually in an effective way. You should write some code that tests the assertion you're trying to eyeball.

Alizter:
After testing my fix for a bit I have come to the conclusion that it doesn't work. It currently clears more lines than I want it to. I also don't need multiline statuses anylonger so I am abandoning this.

PR #8500, opened 2023-08-25
Github action to build monorepo bunch docker image

gridbugs:


gridbugs:
I added this test because benchmarking infrastructure sometimes goes down and changes made to dune during these periods can break the monorepo benchmark without us realizing until much later. This test will build the docker image used to run the monorepo benchmark.

gridbugs:
I tried having the action upload the docker stages to a registry (`ghcr.io`) to speed up the test since otherwise each run will take about 30 minutes. I get this error:
```
denied: installation not allowed to Create organization package
```

Didn't really expect this to work without configuration. Do we have a docker registry set up which the tests can access?

gridbugs:
> Is 200 pulls per 6 hours sufficient for our purposes?

Yep I think that will be fine. Once the account is made we just have to put the password in `secrets.DOCKER_HUB_PASSWORD` in the repo settings. I'll update the PR to use dockerhub instead of github as its registry.

.github/workflows/workflow.yml
```diff
@@ -233,3 +233,17 @@ jobs:
...
           # We disable the Dune cache when running the tests
           DUNE_CACHE: disabled
+
+  monorepo_benchmark_test:
+    name: Build monorepo benchmark docker image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - uses: whoan/docker-build-with-cache-action@v5
```
Alizter, line 243: I believe v6 is the latest.

.github/workflows/workflow.yml
```diff
@@ -233,3 +233,17 @@ jobs:
         env:
           # We disable the Dune cache when running the tests
           DUNE_CACHE: disabled
+
+  monorepo_benchmark_test:
+    name: Build monorepo benchmark docker image
+    runs-on: ubuntu-latest
```
Leonidas-from-XIV, line 239: Do we think that this is a good idea? I am worried that if `ubuntu-latest` switches to a new Ubuntu release, it'll potentially break down.

gridbugs: Possibly, but note that the dockerfile itself specifies a very specific version (of debian) and its unlikely that when ubuntu gets updated it shouldn't affect how docker behaves.

PR #8502, opened 2023-08-25
pkg: convert opam command filters to dune blangs

gridbugs:
This just implements conversion of filters on entire commands using dune's `when` action. Currently it raises a code error if it a `?` unary operator is encountered (opam syntax for checking whether a variable is defined). This may be added in a later change though it's less trivial to convert than the rest of the opam filter dsl and no packages in opam-repository currently use this feature.


gridbugs:
Don't merge until https://github.com/ocaml/dune/pull/8443 is merged.

PR #5956, opened 2022-07-12
Keep rules marked promote until-clean with -p

emillon:
Closes #4401

Issue #4401, opened 2021-03-25
Bad interaction between '(mode promote-until-clean)' and tests ran with '-p <pkg>'


Gbury:
## Expected Behavior

When using a file (let's say `test`) that is "promoted until clean" as a dep for a test defined for a specific package `foo`, `dune build @runtest -p foo` should succeed.

## Actual Behavior

`dune build @runtest -p foo` fails with `Error: No rule found for test`.
The conjunction of `(mode promote-until-clean)` and `-p <pkg>` seems to make dune "forget" about how to build some targets. I don't know if that is expected and if so, what is the correct way to do things.

## Reproduction

See https://gist.github.com/Gbury/598f0b6f751767e8a66fe9c508704976

dune file:
```
(rule
  (target test)
  (mode promote-until-clean) ; without this, everything runs fine 
                             ; but with it, dune complains about "No rule found for test"
  (action (with-stdout-to %{target} (run echo foobar)))
)

(rule
  (alias runtest)
  (package foo)
  (deps (:test test))
  (action (diff %{test} reference))
)
```

0. `dune build @runtest` succeeds as can be expected
1. `dune build @runtest -p foo` fails with "Error: No rule found for test"  (make sure to `dune clean` between attempts, particularly after a successful run)
2. If the `(mode promote-until-clean)` is commented out, `dune build @runtest -p foo` succeeds

## Specifications

- Version of `dune` (output of `dune --version`): 2.8.4
- Version of `ocaml` (output of `ocamlc --version`): 4.11.1
- Operating system (distribution and version): archlinux 5.11.9-arch1-1

## Additional information 

- Link to gist with verbose output (run `dune` with the `--verbose` flag):
Output of `dune build @runtest -p foo --verbose` :
https://gist.github.com/Gbury/80ed0220ccdd729781b35d445731de90

In practice, this happens to me in the dolmen project (see e.g. [this dune file](https://github.com/Gbury/dolmen/blob/687ae8b46b0e7df3f414054b31e52a78e9185b31/src/languages/ae/dune#L34-L66 )), where a rule is defined to generate an updated version of menhir's error message file (which is promoted-until-clean, since it is meant to be used as a reference when updating the error messages after changing the grammar), and another rule is defined to use that file to check that all error states have an error message written.

jeremiedimino:
[...]

PR #8512, opened 2023-08-26
fix: ignoring promotion rules

rgrinberg:
Bug #4401 was `(promote until-clean)` rules being incorrectly ignored by
`--ignore-promoted-rules`. PR #5956 fixed it but only made the fix
available to newer projects.

This means that projects must update their dune versions before users
can actually build this project without being affected by this bug.
Since this bug fix doesn't have much the potential to break anything, we
make the fix available to all versions.


rgrinberg:
Unfortunately, some people rely on the old behavior: https://github.com/ocaml-community/sedlex/pull/154

PR #8516, opened 2023-08-26
rpc: serialize user message styles

### LINKED ISSUE #6921: Extend diagnostics RPC with display markup
Right now the diagnostics RPC strips ANSI from its error messages. This causes information loss, e.g., in expect test differences, when you have a line such as `!  [%expect {| foo |}];` -- you know something has changed on that line but without the ANSI you don't know what. The RPC should be extended to communicate its display markup so clients have this information.

Alizter:
We serialise user message styles in RPC allowing Ansi formatted error codes to be serialised faithfully. We also take the time to fix the differences between RPC messages and regular dune ones that were observed in #8411.

- fixes https://github.com/ocaml/dune/issues/6921
- [x] RPC version compliant
- [x] changlog

### DIFF
--- a/otherlibs/dune-rpc/dune_rpc.mli
+++ b/otherlibs/dune-rpc/dune_rpc.mli
@@ -86,6 +86,101 @@ module V1 : sig
     val to_string_absolute : t -> string
   end
 
+  module Ansi_color : sig
+    module RGB8 : sig
+      (** 8-bit RGB color *)
+      type t
+
+      (** [to_int t] returns the 8-bit color as an integer in the range [0, 255]. *)
+      val to_int : t -> int
+    end
+
+    module RGB24 : sig
+      (** 24-bit RGB color (true color) *)
+      type t
+
+      (** [red t] returns the red component of the 24-bit color [t]. *)
+      val red : t -> int
+
+      (** [green t] returns the green component of the 24-bit color [t]. *)
+      val green : t -> int
+
+      (** [blue t] returns the blue component of the 24-bit color [t]. *)
+      val blue : t -> int
+
+      (** [to_int t] returns the 24-bit color as an integer in the range [0, 0xFFFFFF].
+          Each color components consists of 8 bits. *)
+      val to_int : t -> int
+    end
+
+    module Style : sig
+      (** Ansi Terminal Styles *)
+      type t =
+        [ `Fg_default
+        | `Fg_black
+        | `Fg_red
+        | `Fg_green
+        | `Fg_yellow
+        | `Fg_blue
+        | `Fg_magenta
+        | `Fg_cyan
+        | `Fg_white
+        | `Fg_bright_black
+        | `Fg_bright_red
+        | `Fg_bright_green
+        | `Fg_bright_yellow
+        | `Fg_bright_blue
+        | `Fg_bright_magenta
+        | `Fg_bright_cyan
+        | `Fg_bright_white
+        | `Fg_8_bit_color of RGB8.t
+        | `Fg_24_bit_color of RGB24.t
+        | `Bg_default
+        | `Bg_black
+        | `Bg_red
+        | `Bg_green
+        | `Bg_yellow
+        | `Bg_blue
+        | `Bg_magenta
+        | `Bg_cyan
+        | `Bg_white
+        | `Bg_bright_black
+        | `Bg_bright_red
+        | `Bg_bright_green
+        | `Bg_bright_yellow
+        | `Bg_bright_blue
+        | `Bg_bright_magenta
+        | `Bg_bright_cyan
+        | `Bg_bright_white
+        | `Bg_8_bit_color of RGB8.t
+        | `Bg_24_bit_color of RGB24.t
+        | `Bold
+        | `Dim
+        | `Italic
+        | `Underline
+        ]
+    end
+  end
+
+  module User_message : sig
+    (** User Message Styles *)
+    module Style : sig
+      type t =
+        | Loc
+        | Error
+        | Warning
+        | Kwd
+        | Id
+        | Prompt
+        | Hint
+        | Details
+        | Ok
+        | Debug
+        | Success
+        | Ansi_styles of Ansi_color.Style.t list
+    end
+  end
+
   module Target : sig
     type t =
       | Path of string
@@ -121,6 +216,7 @@ module V1 : sig
 
       val loc : t -> Loc.t
       val message : t -> unit Pp.t
+      val message_with_style : t -> User_message.Style.t Pp.t
     end
 
     type t
@@ -129,6 +225,7 @@ module V1 : sig
     val loc : t -> Loc.t option
     val id : t -> Id.t
     val message : t -> unit Pp.t
+    val message_with_style : t -> User_message.Style.t Pp.t
     val severity : t -> severity option
     val promotion : t -> Promotion.t list
 
--- a/otherlibs/dune-rpc/private/exported_types.ml
+++ b/otherlibs/dune-rpc/private/exported_types.ml
@@ -34,6 +34,240 @@ module Loc = struct
   ;;
 end
 
+module Ansi_color = struct
+  module RGB8 = struct
+    include Stdune.Ansi_color.RGB8
+
+    let sexp =
+      Conv.iso Conv.char Stdune.Ansi_color.RGB8.of_char Stdune.Ansi_color.RGB8.to_char
+    ;;
+  end
+
+  module RGB24 = struct
+    include Stdune.Ansi_color.RGB24
+
+    let sexp =
+      Conv.iso Conv.int Stdune.Ansi_color.RGB24.of_int Stdune.Ansi_color.RGB24.to_int
+    ;;
+  end
+
+  module Style = struct
+    type t = Stdune.Ansi_color.Style.t
+
+    let sexp =
+      let open Conv in
+      let fg_default = constr "Fg_default" unit (fun () -> `Fg_default) in
+      let fg_black = constr "Fg_black" unit (fun () -> `Fg_black) in
+      let fg_red = constr "Fg_red" unit (fun () -> `Fg_red) in
+      let fg_green = constr "Fg_green" unit (fun () -> `Fg_green) in
+      let fg_yellow = constr "Fg_yellow" unit (fun () -> `Fg_yellow) in
+      let fg_blue = constr "Fg_blue" unit (fun () -> `Fg_blue) in
+      let fg_magenta = constr "Fg_magenta" unit (fun () -> `Fg_magenta) in
+      let fg_cyan = constr "Fg_cyan" unit (fun () -> `Fg_cyan) in
+      let fg_white = constr "Fg_white" unit (fun () -> `Fg_white) in
+      let fg_bright_black = constr "Fg_bright_black" unit (fun () -> `Fg_bright_black) in
+      let fg_bright_red = constr "Fg_bright_red" unit (fun () -> `Fg_bright_red) in
+      let fg_bright_green = constr "Fg_bright_green" unit (fun () -> `Fg_bright_green) in
+      let fg_bright_yellow =
+        constr "Fg_bright_yellow" unit (fun () -> `Fg_bright_yellow)
+      in
+      let fg_bright_blue = constr "Fg_bright_blue" unit (fun () -> `Fg_bright_blue) in
+      let fg_bright_magenta =
+        constr "Fg_bright_magenta" unit (fun () -> `Fg_bright_magenta)
+      in
+      let fg_bright_cyan = constr "Fg_bright_cyan" unit (fun () -> `Fg_bright_cyan) in
+      let fg_bright_white = constr "Fg_bright_white" unit (fun () -> `Fg_bright_white) in
+      let fg_8_bit_color =
+        constr "Fg_8_bit_color" RGB8.sexp (fun c -> `Fg_8_bit_color c)
+      in
+      let fg_24_bit_color =
+        constr "Fg_24_bit_color" RGB24.sexp (fun c -> `Fg_24_bit_color c)
+      in
+      let bg_default = constr "Bg_default" unit (fun () -> `Bg_default) in
+      let bg_black = constr "Bg_black" unit (fun () -> `Bg_black) in
+      let bg_red = constr "Bg_red" unit (fun () -> `Bg_red) in
+      let bg_green = constr "Bg_green" unit (fun () -> `Bg_green) in
+      let bg_yellow = constr "Bg_yellow" unit (fun () -> `Bg_yellow) in
+      let bg_blue = constr "Bg_blue" unit (fun () -> `Bg_blue) in
+      let bg_magenta = constr "Bg_magenta" unit (fun () -> `Bg_magenta) in
+      let bg_cyan = constr "Bg_cyan" unit (fun () -> `Bg_cyan) in
+      let bg_white = constr "Bg_white" unit (fun () -> `Bg_white) in
+      let bg_bright_black = constr "Bg_bright_black" unit (fun () -> `Bg_bright_black) in
+      let bg_bright_red = constr "Bg_bright_red" unit (fun () -> `Bg_bright_red) in
+      let bg_bright_green = constr "Bg_bright_green" unit (fun () -> `Bg_bright_green) in
+      let bg_bright_yellow =
+        constr "Bg_bright_yellow" unit (fun () -> `Bg_bright_yellow)
+      in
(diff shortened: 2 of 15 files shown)


otherlibs/dune-rpc/private/exported_types.ml
```diff
@@ -34,6 +34,468 @@ module Loc = struct
   ;;
 end
 
+let sexp_pp (conv_tag : 'a Conv.value) : 'a Pp.t Conv.value =
```
Alizter, line 37: This generalises the previous sexp_pp function, and adds it to the API for use elsewhere.

otherlibs/dune-rpc/private/exported_types.ml
```diff
@@ -34,6 +34,468 @@ module Loc = struct
...
+  module RGB24 = struct
+    type t = int
+
+    let sexp = Conv.int
+    let to_ansi_color_rgb24 t = Stdune.Ansi_color.RGB24.of_int t
+    let of_ansi_color_rgb24 t = Stdune.Ansi_color.RGB24.to_int t
+  end
+
+  module Style = struct
+    type t =
```
Alizter, line 121: Ansi_color.Style in Stdune uses polymorphic variants for this type as there is a need to add constructors for parsing reasons. I originally tried to do the same here thinking that it would allow us to avoid the conversion functions however the compiler didn't allow it. Therefore for clarity I made this a regular variant with explicit conversion functions.

otherlibs/dune-rpc/private/exported_types.ml
```diff
@@ -34,6 +34,468 @@ module Loc = struct
...
+      | `Dim -> Dim
+      | `Italic -> Italic
+      | `Underline -> Underline
+    ;;
+  end
+end
+
+module User_message = struct
+  module Style = struct
+    type t =
```
Alizter, line 408: I am unsure if we want to expose the internal details of Dune's error message formatting and make them concrete in the RPC API. @rgrinberg do you think it would be better to make this type opaque and add explicit constructors? It will be difficult to remove constructors from the Stdune version in the future if we keep this around.

otherlibs/dune-rpc/private/exported_types.ml
```diff
@@ -189,17 +585,18 @@ module Diagnostic = struct
 
   module Related = struct
     type t =
-      { message : unit Pp.t
+      { message : User_message.Style.t Pp.t
```
Alizter, line 588: Jobs also have a `unit Pp.t` description, but I will not touch it in this PR.

otherlibs/dune-rpc/private/exported_types.ml
```diff
@@ -189,17 +585,18 @@ module Diagnostic = struct
...
       ; loc : Loc.t
       }
 
-    let message t = t.message
+    let message t = t.message |> Pp.map_tags ~f:(fun _ -> ())
+    let message_with_style t = t.message
```
Alizter, line 593: We add an extra function allowing for an opt in to the stylised message. This keeps our API version compatible. The old version is just the new version with the style tags removed.

otherlibs/dune-rpc/private/exported_types.ml
```diff
@@ -256,35 +654,10 @@ module Diagnostic = struct
   let to_dyn t = Sexp.to_dyn (Conv.to_sexp sexp t)
 
   let to_user_message t =
```
Alizter, line 656: Since we have a faithful reconstruction of the styles, we get rid of the ad-hoc message styling we had before and simply use Stdune.User_message.make to get more agreement. There are still some pain points however:
- User_message concatenates hints etc. during construction which we cannot remove. However since the style is also serialised this doesn't appear different to the user. Only the pp, which in the tests can be inspected, is slightly different.

otherlibs/dune-rpc/dune_rpc.mli
```diff
@@ -86,6 +86,80 @@ module V1 : sig
...
   end
 
+  module Ansi_color : sig
+    module RGB8 : sig
+      type t
+    end
+
+    module RGB24 : sig
+      type t
+    end
```
Alizter, line 96: I'm not sure if we want to add any functions allowing us to inspect the contents. For now they are a black box to the consumer of RPC.

otherlibs/dune-rpc/private/procedures.ml
```diff
@@ -9,14 +9,115 @@ module Public = struct
   end
 
   module Diagnostics = struct
+    module V1 = struct
```
Alizter, line 12: V1 of the diagnostics API is just the previous version sharing as much with the new as possible. Basically anywhere `unit Pp.t` is mentioned, we use the lossy conversions to get them back. On the other hand it will mean that new dunes listening to old clients won't have colour support, which is fine I think.

otherlibs/stdune/src/ansi_color.ml
```diff
@@ -4,13 +4,17 @@ module RGB8 : sig
   val to_dyn : t -> Dyn.t
   val of_int : int -> t
   val to_int : t -> int
+  val of_char : char -> t
+  val to_char : t -> char
```
Alizter, line 8: I was a little hesitant to expose these. Maybe I should put them in a module. It is very unlikely that the internal representation will change for these, but on the other hand I don't want to make the consumer of the API think that it is just a char.

otherlibs/stdune/src/ansi_color.ml
```diff
@@ -30,9 +34,11 @@ module RGB24 : sig
   val red : t -> int
   val green : t -> int
   val blue : t -> int
-  val create : r:int -> g:int -> b:int -> t
+  val make : red:int -> green:int -> blue:int -> t
+  val to_int : t -> int
+  val of_int : int -> t
```
Alizter, line 39: For consistency with other functions in the code base, this was renamed to `make` and full labelled arguments given.

src/dune_rpc_impl/diagnostics.ml
```diff
@@ -52,10 +56,28 @@ let diagnostic_of_error : Build_system.Error.t -> Dune_rpc_private.Diagnostic.t
       ; loc = make_loc (Option.value_exn related.loc)
       })
   in
+  let message =
```
Alizter, line 59: Here we add in the hints the same way `User_message.pp` does it. Maybe it would be better to introduce a `hints` field in the RPC type?

test/expect-tests/dune_rpc/dune_rpc_tests.ml
Alizter, line 1: I'm not sure why the digests here have changed, I will have to look into it.

test/expect-tests/dune_rpc_e2e/dune_rpc_diagnostics.ml
```diff
@@ -521,6 +524,7 @@ let%expect_test "create and fix error" =
             ; [ "Verbatim"
               ; "This expression has type int but an expression was expected of type\n\
                 \  string\n\
+                 \n\
```
Alizter, line 527: These extra new lines should probably not be here.

src/dune_rpc_impl/diagnostics.ml
Alizter: After thinking about it, adding a hints field is kind of useless so I am happy with the way this is at the moment.

otherlibs/dune-rpc/private/exported_types.ml
Alizter: I actually had to introduce two versions of this function. The new version serialises tags as `(Tag (pair tag pp))` which is different to how it was done previously as `(Tag pp)`. Therefore I have included the old version too and made sure to version bump both diagnostics and the diagnostic poll.

test/expect-tests/dune_rpc/dune_rpc_tests.ml
Alizter: Fixed. I subtly changed the way Pp was being serialised. I've left in the old version and made diagnostics and diagnostic poll have a new version.

Alizter:
It's still possible for some messages to lose the `Error:` prefix. I have not yet been able to reproduce this in a test. I am not exactly sure what is causing this. I think it has something to do with how we use ocamlc-loc to parse error messages into compound user messages and then turn them into user messages. But I haven't been able to confirm.

pmwhite:
Sorry for the delay in starting to review this! I just started looking at it today. One thing that I don't understand is why we need to introduce new `User_message` and `Ansi_color` modules if `Stdune` already has sub-modules that look very similar to the ones introduced in this feature. Maybe this is related to what you mentioned about the compiler not letting you use polymorphic variants; I didn't follow that part - could you elaborate?

Alizter:
@pmwhite That's related but not exactly why we have to mirror the types in the dune_rpc library. We need to keep the dependencies of that library small so exposing `Stdune` in the API would require consumers to also pull in the `Stdune` library.

My hope with the poly variants was that we could mirror the API but share the implementation but that didn't work out so I ended up mirroring everything.

pmwhite:
I see, I wasn't aware of that constraint. The library _does_ depend on `stdune`, but it doesn't mention it in the interface, which I assume is useful so that users of the API can more easily use their own standard library, like `base` or something. How come mirroring the API and sharing the implementation doesn't work? I would have expected that to work out just fine.

Alizter:
@pmwhite When I tried it, I had to do something like this:
```
let to_user_message_style = function
  | `Fg_8_bit_color rgb -> RGB8.to_rgb8 rgb
...
  | x -> x
```
Now you would have thought that the final clause would go through since the type is the same structurally, however the compiler rejected this. This meant I had to write out each case anyway. Doing so meant there was little point keeping the poly variants around.

rgrinberg:
> Now you would have thought that the final clause would go through since the type is the same structurally, however the compiler rejected this.

I think you just needed to add a type annotation.

rgrinberg:
> I see, I wasn't aware of that constraint. The library does depend on stdune, but it doesn't mention it in the interface, which I assume is useful so that users of the API can more easily use their own standard library, like base or something

That's right. In particular, stdune is completely unstable so depending on it in code is going to produce churn for users of this API. Although the dune rpc api isn't completely stable, we already have production users so we do our best to maintain backwards compatibility.

pmwhite:
> @pmwhite When I tried it, I had to do something like this:
> 
> ```ocaml
> let to_user_message_style = function
>   | `Fg_8_bit_color rgb -> RGB8.to_rgb8 rgb
> ...
>   | x -> x
> ```

Is the function necessary? Could we instead expose the equality to `Stdune.User_message.Style.t` in `private/exported_types.mli`, since all we care about is not having `Stdune` show up in the public dune-rpc interface? Most of the value I was imagining sharing the types would have is that then we wouldn't have to convert back and forth between the types. (maybe I'm missing a reason why we can't share the types?)

otherlibs/dune-rpc/private/exported_types.ml
```diff
@@ -34,6 +34,402 @@ module Loc = struct
...
+      | Bg_bright_white
+      | Bg_8_bit_color of RGB8.t
+      | Bg_24_bit_color of RGB24.t
+      | Bold
+      | Dim
+      | Italic
+      | Underline
+
+    let sexp =
+      let open Conv in
```
rgrinberg, line 100: Could you introduce `Conv.enum` to make this all a bit more compact?

otherlibs/dune-rpc/private/exported_types.ml
rgrinberg: How would users use the API if this type was completely opaque?

otherlibs/dune-rpc/dune_rpc.mli
rgrinberg: How are clients supposed to use them then?

otherlibs/dune-rpc/private/exported_types.ml
Alizter: `Conv.enum` as it exists is not sufficient since I cannot handle constructor arguments.

Alizter:
@pmwhite I thought a bit harder about this and you are totally right, we can share much of the implementation. I've gone ahead and removed all the conversion functions and it has simplified things considerably.

otherlibs/dune-rpc/dune_rpc.mli
Alizter: I'll add some functions.

otherlibs/dune-rpc/private/exported_types.ml
Alizter: Fixed this.

Alizter:
I've added more functions to RGB8 and RGB24 in the `dune_rpc` API, allowing consumers to produce and consume those values easily.

otherlibs/stdune/src/ansi_color.ml
pmwhite: Seems fine to me.

otherlibs/dune-rpc/dune_rpc.mli
```diff
@@ -86,6 +86,112 @@ module V1 : sig
...
+  module Ansi_color : sig
+    module RGB8 : sig
+      (** 8-bit RGB color *)
+      type t
+
+      (** [of_int i] creates an 8-bit color from the least significant 8 bits of [i]. *)
+      val of_int : int -> t
+
+      (** [to_int t] returns the 8-bit color as an integer in the range [0, 255]. *)
+      val to_int : t -> int
```
rgrinberg, line 98: I don't think these functions are very useful. What about accessors like `red`, `green`, `blue`?

otherlibs/dune-rpc/dune_rpc.mli
```diff
@@ -86,6 +86,112 @@ module V1 : sig
...
+    end
+
+    module RGB24 : sig
+      (** 24-bit RGB color (true color) *)
+      type t
+
+      (** [make ~red ~green ~blue] creates a 24-bit color from the given RGB
+          components. Each color component should be in the range [0, 255]. *)
+      val make : red:int -> green:int -> blue:int -> t
```
rgrinberg, line 107: Don't think this constructor is necessary

otherlibs/dune-rpc/dune_rpc.mli
```diff
@@ -86,6 +86,112 @@ module V1 : sig
...
+      (** [blue t] returns the blue component of the 24-bit color [t]. *)
+      val blue : t -> int
+
+      (** [of_int i] creates a 24-bit color from the least significant 24 bits of [i].
+          Each color components consists of 8 bits. *)
+      val of_int : int -> t
+
+      (** [to_int t] returns the 24-bit color as an integer in the range [0, 0xFFFFFF].
+          Each color components consists of 8 bits. *)
+      val to_int : t -> int
```
rgrinberg, line 124: Ditto

otherlibs/dune-rpc/dune_rpc.mli
Alizter: It's not possible to do split into red, green and blue like we do for 24-bit color since not every terminal encodes 8-bit colors in the same manner. The only thing that doesn't change is that there are 256 colors.

otherlibs/dune-rpc/dune_rpc.mli
Alizter: This just makes it easier for the consumers of the API to construct terminal colors. It's not strictly necessary, but it saves people from having to bitshift themselves.

rgrinberg: Okay, but why would they need to construct these colors in the first place? They have a dune rpc client that emits error messages. The most important thing they need to do is to display. For that, accessors are enough.

Alizter: Ah yes, fair enough.

otherlibs/dune-rpc/dune_rpc.mli
Alizter: If I'm making a web client for viewing RPC messages, this might be a useful value to have since colors are hex values usually. (As an example of how this might be useful).

otherlibs/dune-rpc/dune_rpc.mli
Alizter: The only time this can happen is if `Dune` runs a program with 8-bit color output and we wish to preserve that color. At that point we are at the mercy of the client on how we quantize those values.

otherlibs/dune-rpc/private/procedures.ml
rgrinberg: Could you move this to a separate file? E.g. `diagnostics_v1.ml`?

rgrinberg (approved):
Okay, this looks ready to me. I'll give @pmwhite another chance to comment.

cc @ddickstein who was interested in this.

pmwhite:
Yeah, looks good to me, too.

PR #8518, opened 2023-08-26
fix: --ignore-promoted-rules should work on internal rules

rgrinberg:
Internal promotion rules such as generating opam files weren't being
ignored under --ignored-promoted-rules.

### DIFF
--- a/src/dune_rules/rule_mode_decoder.mli
+++ b/src/dune_rules/rule_mode_decoder.mli
@@ -15,3 +15,11 @@ end
 
 val decode : Rule.Mode.t Dune_lang.Decoder.t
 val field : Rule.Mode.t Dune_lang.Decoder.fields_parser
+
+(** [is_ignored mode ~until_clean] will return if a rule with [mode] should be
+    ignored whenever [--ignored-promoted-rules] is set.
+
+    [until_clean] is used to set if [(promote (until-clean))] is ignored as
+    considered by this function. Old versions of dune would incorrectly ignore
+    this, so we need to maintain the old behavior for now. *)
+val is_ignored : Rule.Mode.t -> until_clean:[ `Ignore | `Keep ] -> bool
--- a/doc/changes/8518.md
+++ b/doc/changes/8518.md
@@ -1,0 +1,2 @@
+- Ignore internal promote rules when `--ignore-promoted-rules` is set (#8518,
+  fix #8417, @rgrinberg)
--- a/otherlibs/dune-site/test/run.t
+++ b/otherlibs/dune-site/test/run.t
@@ -170,6 +170,7 @@ Test with an opam like installation
     ["dune" "install" "-p" name "--create-install-files" name]
   ]
 
+  $ dune build b/b.opam c/c.opam d/d.opam
   $ dune build -p a --promote-install-files=false @install
 
   $ test -e a/a.install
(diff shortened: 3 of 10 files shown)


src/dune_rules/rule_mode_decoder.mli
```diff
@@ -15,3 +15,4 @@ end
 
 val decode : Rule.Mode.t Dune_lang.Decoder.t
 val field : Rule.Mode.t Dune_lang.Decoder.fields_parser
+val is_ignored : Rule.Mode.t -> until_clean:[ `Ignore | `Keep ] -> bool
```
Leonidas-from-XIV, line 18: I think it would be useful to have a high-level doc-string to explain what the setting of `until_clean` means.

Leonidas-from-XIV (commented):
The fix is fairly straightforward and the code is actually sort of simpler than before which is nice but I would like to understand the `until_clean` setting. Or making the API a bit more self-descriptive.

src/dune_rules/rule_mode_decoder.mli
rgrinberg: Done.

rgrinberg:
Added a description. Let me know what you think of it.

src/dune_rules/rule_mode_decoder.mli
```diff
@@ -15,3 +15,9 @@ end
 
 val decode : Rule.Mode.t Dune_lang.Decoder.t
 val field : Rule.Mode.t Dune_lang.Decoder.fields_parser
+
+(** [is_ignored mode ~until_clean] will return if rule with [mode] should be
+    ignored. [until_clean] is used to set if [(promote (until-clean))] is
+    ignored as considerd by this function. Old versions of dune would
```
Leonidas-from-XIV, line 21: There is a typo here, but asides from this this makes it much easier to understand how this parameter should be used. Thanks.

(This might possibly be a case where a default argument could make sense, since in new code we don't want to use `` `Ignore`` I assume?)

Leonidas-from-XIV (approved):

src/dune_rules/rule_mode_decoder.mli
rgrinberg: I actually think we should consider just removing the old behavior and making the breaking change. The reason is that preserving the old behavior makes it impossible to build old projects without modifying the source directory. This is quite a pain for those packaging software that uses dune.

Alizter: I vaguely remember a plan to deprecate the other promote modes, would it be possible to interpret them as `(mode promote)` from now on?

rgrinberg: The only one that is deprecated is `patch_back_source_tree`

Leonidas-from-XIV: @rgrinberg You mean, without patching the dune-lang version to whatever will include this change? Maybe worth asking @glondu if this would be helpful enough to break compatibility.

rgrinberg: > You mean, without patching the dune-lang version to whatever will include this change?

Yes. Especially given that updating the dune version in the project file might run into other errors, patching isn't a reasonable workaround.

Leonidas-from-XIV: You might be right, as probably nobody depends on the existing incorrect behaviour.

rgrinberg: Unfortunately, [they do](https://github.com/ocaml/dune/pull/8512#issuecomment-1694431684) 😂 

In any case, we can just have them release a new version and then do this change.

PR #8524, opened 2023-08-27
refactor(tui): remove all unreleated code from backend specific code

Alizter:


src/dune_tui/dune_tui.ml
```diff
@@ -2,288 +2,287 @@ open Import
...
+    | `ASCII '?', _ ->
+      handle_help ();
+      `Handled
+    | _ -> `Unhandled
+  in
+  Ui.zcat [ Ui.vcat [ status_bar; message_viewer ]; help_box ]
+  |> Ui.keyboard_area keyboard_handler
+;;
 
+module Tui = struct
```
rgrinberg, line 283: A better name for this module would be `Console_backend`.

PR #8527, opened 2023-08-27
odoc: add test for `stdlib` stanza

jchavarri:
I am trying to reproduce some issue with the odoc generated html from Melange stdlib producing broken links (for example, the generated page for the [Lazy module](https://melange.re/v1.0.0/api/ml/melange/Stdlib/Lazy/index.html) has a broken link at the top to `CamlinternalLazy.t`).

When trying to reproduce with dune, I am not being able to do so, because when I add `stdlib` stanza to a library, no html is generated for any of the modules contained in the library.

Is this the expected behavior?

### DIFF
--- a/test/blackbox-tests/test-cases/odoc/stdlib-stanza.t
+++ b/test/blackbox-tests/test-cases/odoc/stdlib-stanza.t
@@ -1,0 +1,42 @@
+Demonstrate how odoc interops with the `stdlib` stanza
+
+  $ cat > dune-project << EOF
+  > (lang dune 3.9)
+  > (using experimental_building_ocaml_compiler_with_dune 0.1)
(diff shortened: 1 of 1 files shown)


rgrinberg:
cc @jonludlam

rgrinberg:
> Is this the expected behavior?

Unlikely. But Jon would know what's the plan to build the stdlib with odoc best.

jonludlam:
That's an interesting couple of bugs! Odoc is expected there to be a page generated for `CamlinternalLazy` because it's not a hidden module. We probably need to special-case this because that `Camlinternal` modules are hidden is a quirk of the stdlib.

I've no idea why the stdlib stanza causes it to stop generating _any_ documentation.

PR #8534, opened 2023-08-29
fix(tui): exit nicely when tui not supported

### LINKED ISSUE #8523: Gracefully Fail on `--display tui` on windows
We don't support windows in `--display tui` and we currently fail with an ugly stacktrace. We should instead exit properly and tell the user this mode isn't supported on windows.

Alizter:
- fixes #8523 
- [x] #8536 
- [x] Test linux
- [x] Test windows

image


Alizter:
@rgrinberg I've changed it to only single out Windows.

PR #8536, opened 2023-08-29
feature: FreeBSD, OpenBSD, NetBSD platform detection

Alizter:
We add platform detection for FreeBSD, OpenBSD and NetBSD in order to refine the `Other` constructor for Platform.OS.t. This allows us to exclude Other when we want to be more sure a feature is supported.

I could have put these all into the same constructor, however there are subtle differences which may become relevant. It would also make the C stubs a little trickier to use.

### DIFF
--- a/otherlibs/stdune/src/io.ml
+++ b/otherlibs/stdune/src/io.ml
@@ -109,7 +109,7 @@ module Copyfile = struct
     match Platform.OS.value with
     | Darwin -> `Copyfile
     | Linux -> `Sendfile
-    | Windows | Other -> `Nothing
+    | _ -> `Nothing
   ;;
 
   let sendfile_with_fallback =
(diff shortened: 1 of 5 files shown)


otherlibs/stdune/src/io.ml
```diff
@@ -109,7 +109,7 @@ module Copyfile = struct
     match Platform.OS.value with
     | Darwin -> `Copyfile
     | Linux -> `Sendfile
-    | Windows | Other -> `Nothing
+    | _ -> `Nothing
```
Alizter, line 112: Defaulting to `Nothing here until we have a demand for it.

src/dune_file_watcher/dune_file_watcher.ml
```diff
@@ -362,7 +362,7 @@ let select_watcher_backend () =
   else (
     match Platform.OS.value with
     | Windows -> `Fswatch_win
-    | Linux | Darwin | Other -> fswatch_backend ())
+    | Linux | Darwin | FreeBSD | OpenBSD | NetBSD | Other -> fswatch_backend ())
```
Alizter, line 365: I checked and fswatch has support for these.

PR #8538, opened 2023-08-29
feature: add Haiku, Serenity, Solaris detection

Alizter:
We also change the Platform.OS.t type to a polymorphic variant. The reason for doing this is it allows us to define subtypes such as `fswatch_supported` allowing us to pattern match elsewhere.

This will be useful for `posix` compliant OSes etc. later on.

We also add the following OS detection:
- Haiku
- Serenity
- Solaris
- DragonFly

### DIFF
--- a/otherlibs/stdune/src/platform.ml
+++ b/otherlibs/stdune/src/platform.ml
@@ -1,13 +1,44 @@
 module OS = struct
   (* CR-someday alizter: Include mingw32, mongw64, cygwin *)
   type t =
-    | Darwin
-    | Linux
-    | Windows
-    | FreeBSD
-    | NetBSD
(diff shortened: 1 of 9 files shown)


otherlibs/stdune/src/platform.ml
```diff
@@ -1,29 +1,68 @@
...
+    ; `OpenBSD
+    ; `DragonFly
+    ; `Haiku
+    ; `Serenity
+    ; `Solaris
+    ; `Other
+    ]
+  ;;
+
+  type fswatch_support =
```
rgrinberg, line 33: This type is unnecessary because it's the equivalent of just writing a pattern match.

Also, it should never be defined here.

PR #8539, opened 2023-08-29
fix: forbid escaping the context root

rgrinberg:

### DIFF
--- a/src/dune_rules/expander.ml
+++ b/src/dune_rules/expander.ml
@@ -771,7 +771,16 @@ let make
 
 let expand_path t sw =
   let+ v = expand t ~mode:Single sw in
-  Value.to_path v ~error_loc:(String_with_vars.loc sw) ~dir:(Path.build t.dir)
+  let loc = String_with_vars.loc sw in
+  let path = Value.to_path v ~error_loc:loc ~dir:(Path.build t.dir) in
+  let context_root = t.context.build_context.build_dir in
+  (match Path.as_in_build_dir path with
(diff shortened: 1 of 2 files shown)


src/dune_rules/expander.ml
```diff
@@ -771,7 +771,16 @@ let make
...
-  Value.to_path v ~error_loc:(String_with_vars.loc sw) ~dir:(Path.build t.dir)
+  let loc = String_with_vars.loc sw in
+  let path = Value.to_path v ~error_loc:loc ~dir:(Path.build t.dir) in
+  let context_root = t.context.build_context.build_dir in
+  (match Path.as_in_build_dir path with
+   | Some p when not (Path.Build.is_descendant p ~of_:context_root) ->
+     (* TODO consider turning these into external paths, since we already allow
+        them to be specified as absolute paths. *)
+     User_error.raise ~loc [ Pp.text "path cannot escape the context root" ]
+   | _ -> ());
```
Leonidas-from-XIV, line 782: So if the path is absolute, this check passes? Doesn't that pose potential sandboxing issues?

rgrinberg: Yes, we do not sandbox external paths. That is intended behavior because some things are impossible to sandboxing via our mechanisms. For example, the compiler is not relocatable.

Leonidas-from-XIV (approved):
Looks sensible and the error message is useful, so let's merge it :)

rgrinberg:
It might be possible to try and escape the context root in other ways. So let's keep this issue in mind and perhaps improve this check in other places.

PR #8540, opened 2023-08-29
chore: remove notty_unix dep from nottui

Alizter:
`nottui` has a small render loop that we don't use that uses `notty_unix`. We therefore modify the fetch script to remove this loop and remove the dependency.

This makes it easier to drop in another backend for notty in the future.


rgrinberg:
What does this gain? Improve the build times?

I'd rather this live as a commit in our fork rather than a sed script.

Alizter:
Yes it will improve the build time, but my motivation was for Windows. Since we are not supporting it just yet, I will close this PR.

I don't mind forking it, but it seems more work than needed when we are just removing the bottom of a file.

PR #8558, opened 2023-08-31
fix: allow shadowing builtins

rgrinberg:
Previously, builtins would always shadow librares in findlib


rgrinberg:
The benefit of this is that it allows to delay loading the compiler. Let's see if anyone actually relies on this behavior.

rgrinberg:
Actually, the fact that the builtins can't be shadowed demonstrates a bug. It means that one can define a public library that will shadow a builtin depending on whether it's installed or not. In other words, one could do the following:

1. Define two packages `str` and `foo`. Where `foo` uses the `str` library in the `str` package.
2. This will build just fine when built in the same workspace.
3. But will fail if the `str` package is installed because it will be shadowed by the builtin.

@nojb do you agree that this is an issue?

nojb:
> ping @nojb. I think it's worth having this fix in 3.11

Sorry for the delay. I will take a look in the next day or two.

nojb (approved):
LGTM

test/expect-tests/findlib_tests.ml
```diff
@@ -107,9 +107,8 @@ let%expect_test "configurator" =
 
 let%expect_test "builtins" =
   print_pkg_archives "str";
-  [%expect
-    {|
-    Hidden { byte = [ "stdlib/str.cma" ]; native = [ "stdlib/str.cmxa" ] } |}];
+  [%expect {|
+    Available { byte = []; native = [] } |}];
```
nojb, line 111: I don't understand why this test result changed; can you explain it?

rgrinberg: Sure, the test demonstrates the bug being fixed. The OCAMLPATH directory in the test contains an `str` library. Previously, it was being completely ignored and pulled from the builtin map (see the stdlib/ prefix).

Now, it's being pulled from the OCAMLPATH directory as expected.

One more little thing to note is that `Hidden` was changed to `Available`. That's because the builtin `str` contains an `exists_if` that wasn't satisfied (see the fake paths). While the library doesn't have that.

nojb: OK, thanks, but why are `byte` and `native` empty?

rgrinberg: It's just how the META file is defined for `str` in the tests.

nojb: Right, thanks!

PR #8567, opened 2023-09-01
Communicate READER to merlin for configured dialects

andreypopp:
Communicate `READER` to merlin for configured dialects.

The user visible change is the `(merlin_reader ...)` is now allowed in `dune-project` in `(dialect ...)` configuration:
```
(dialect
 (name mlx)
 (implementation
  (extension mlx)
  (merlin_reader mlx)
  (preprocess
   (run ./mlx/pp.exe %{input-file}))))
```

### DIFF
--- a/src/dune_rules/dialect.ml
+++ b/src/dune_rules/dialect.ml
@@ -11,9 +11,10 @@ module File_kind = struct
     ; preprocess : (Loc.t * Action.t) option
     ; format : (Loc.t * Action.t * string list) option
     ; print_ast : (Loc.t * Action.t) option
+    ; merlin_reader : (Loc.t * string list) option
     }
 
-  let encode { kind; extension; preprocess; format; print_ast } =
+  let encode { kind; extension; preprocess; format; print_ast; merlin_reader } =
     let open Dune_lang.Encoder in
     let kind =
       string
@@ -33,10 +34,11 @@ module File_kind = struct
                 "print_ast"
                 Action.encode
                 (Option.map ~f:(fun (_, x) -> x) print_ast)
+            ; field_o "merlin_reader" (list string) (Option.map ~f:snd merlin_reader)
             ])
   ;;
 
-  let to_dyn { kind; extension; preprocess; format; print_ast } =
+  let to_dyn { kind; extension; preprocess; format; print_ast; merlin_reader } =
     let open Dyn in
     record
       [ "kind", Ml_kind.to_dyn kind
@@ -44,6 +46,7 @@ module File_kind = struct
       ; "preprocess", option (fun (_, x) -> Action.to_dyn x) preprocess
       ; "format", option (fun (_, x, y) -> pair Action.to_dyn (list string) (x, y)) format
       ; "print_ast", option (fun (_, x) -> Action.to_dyn x) print_ast
+      ; "merlin_reader", option (fun (_, x) -> list string x) merlin_reader
       ]
   ;;
 end
@@ -85,13 +88,17 @@ let decode =
         "format"
         (map ~f:(fun (loc, x) -> loc, x, []) (located Action.decode_dune_file))
     and+ print_ast = field_o "print_ast" (located Action.decode_dune_file)
+    and+ merlin_reader =
+      field_o
+        "merlin_reader"
+        (Dune_lang.Syntax.since Stanza.syntax (3, 11) >>> located (repeat1 string))
     and+ syntax_ver = Syntax.get_exn Stanza.syntax in
     let ver = 3, 9 in
     if syntax_ver < ver && Option.is_some (String.index_from extension 1 '.')
     then (
       let what = "the possibility of defining extensions containing periods" in
       Syntax.Error.since loc Stanza.syntax ver ~what);
-    { File_kind.kind; extension; preprocess; format; print_ast }
+    { File_kind.kind; extension; preprocess; format; print_ast; merlin_reader }
   in
   fields
     (let+ name = field "name" string
@@ -143,6 +150,13 @@ let print_ast { file_kinds; _ } ml_kind =
   x.print_ast
 ;;
 
+let merlin_reader { file_kinds; _ } ml_kind =
+  let open Option.O in
+  let* x = Ml_kind.Dict.get file_kinds ml_kind in
+  let+ _, merlin_reader = x.merlin_reader in
+  merlin_reader
+;;
+
 let ocaml =
   let format kind =
     let flag_of_kind = function
@@ -180,6 +194,7 @@ let ocaml =
           , format kind
           , [ ".ocamlformat"; ".ocamlformat-ignore"; ".ocamlformat-enable" ] )
     ; print_ast = Some (Loc.none, print_ast kind)
+    ; merlin_reader = None
     }
   in
   let intf = Some (file_kind Ml_kind.Intf ".mli") in
@@ -222,6 +237,7 @@ let reason =
     ; preprocess = Some (Loc.none, preprocess)
     ; format = Some (Loc.none, format, [])
     ; print_ast = Some (Loc.none, print_ast)
+    ; merlin_reader = None
     }
   in
   let intf = Some (file_kind Ml_kind.Intf ".rei") in
@@ -251,6 +267,7 @@ let rescript =
     ; preprocess = Some (Loc.none, preprocess)
     ; format = Some (Loc.none, format, [])
     ; print_ast = None
+    ; merlin_reader = None
     }
   in
   let intf = Some (file_kind Ml_kind.Intf ".resi") in
@@ -271,7 +288,12 @@ module DB = struct
   type t =
     { by_name : dialect String.Map.t
     ; by_extension : dialect Filename.Extension.Map.t
-    ; mutable extensions_for_merlin : string option Ml_kind.Dict.t list option
+    ; for_merlin : for_merlin Lazy.t
+    }
+
+  and for_merlin =
+    { extensions : string option Ml_kind.Dict.t list
+    ; readers : string list String.Map.t
     }
 
   let fold { by_name; _ } = String.Map.fold by_name
@@ -279,35 +301,47 @@ module DB = struct
   let empty =
     { by_name = String.Map.empty
     ; by_extension = Filename.Extension.Map.empty
-    ; extensions_for_merlin = None
+    ; for_merlin = lazy { extensions = []; readers = String.Map.empty }
     }
   ;;
 
-  let set_extensions_for_merlin t =
-    let v =
-      fold t ~init:[] ~f:(fun d s ->
-        let impl = extension d Ml_kind.Impl in
-        let intf = extension d Ml_kind.Intf in
-        if (* Only include dialects with no preprocessing and skip default file
-              extensions *)
-           preprocess d Ml_kind.Impl <> None
-           || preprocess d Ml_kind.Intf <> None
-           || (impl = extension ocaml Ml_kind.Impl && intf = extension ocaml Ml_kind.Intf)
-        then s
-        else { Ml_kind.Dict.impl; intf } :: s)
-      |> List.sort ~compare:(Ml_kind.Dict.compare (Option.compare String.compare))
+  let compute_for_merlin by_name =
+    let extensions, readers =
+      String.Map.fold
+        by_name
+        ~init:([], String.Map.empty)
+        ~f:(fun d (extensions, readers) ->
+          let handle_ml_kind kind readers =
+            let ext = extension d kind in
+            if ext = extension ocaml kind
+            then (* this is standard dialect, exclude *) None, readers
+            else (
+              let reader = merlin_reader d kind in
+              match ext, reader with
+              | Some ext, Some reader -> Some ext, String.Map.add_exn readers ext reader
+              | _ ->
+                if preprocess d kind <> None
+                then (* we have preprocessor defined *) None, readers
+                else ext, readers)
+          in
+          let impl, readers = handle_ml_kind Ml_kind.Impl readers in
+          let intf, readers = handle_ml_kind Ml_kind.Intf readers in
+          let extensions =
+            match impl, intf with
+            | None, None -> extensions
+            | _ -> { Ml_kind.Dict.impl; intf } :: extensions
+          in
+          extensions, readers)
     in
-    t.extensions_for_merlin <- Some v;
-    v
+    let extensions =
+      List.sort ~compare:(Ml_kind.Dict.compare (Option.compare String.compare)) extensions
+    in
+    { extensions; readers }
   ;;
 
-  let extensions_for_merlin t =
-    match t.extensions_for_merlin with
-    | Some s -> s
-    | None -> set_extensions_for_merlin t
-  ;;
+  let for_merlin t = Lazy.force t.for_merlin
 
-  let add { by_name; by_extension; extensions_for_merlin = _ } ~loc dialect =
+  let add { by_name; by_extension; for_merlin = _ } ~loc dialect =
     let by_name =
       match String.Map.add by_name dialect.name dialect with
       | Ok by_name -> by_name
@@ -331,7 +365,7 @@ module DB = struct
     let by_extension =
       add_ext (add_ext by_extension dialect.file_kinds.intf) dialect.file_kinds.impl
     in
-    { by_name; by_extension; extensions_for_merlin = None }
+    { by_name; by_extension; for_merlin = lazy (compute_for_merlin by_name) }
   ;;
 
   let of_list dialects = List.fold_left ~f:(add ~loc:Loc.none) ~init:empty dialects
--- a/src/dune_rules/dialect.ml
+++ b/src/dune_rules/dialect.ml
@@ -91,7 +91,7 @@ let decode =
     and+ merlin_reader =
       field_o
         "merlin_reader"
-        (Dune_lang.Syntax.since Stanza.syntax (3, 11) >>> located (repeat1 string))
+        (Dune_lang.Syntax.since Stanza.syntax (3, 16) >>> located (repeat1 string))
     and+ syntax_ver = Syntax.get_exn Stanza.syntax in
     let ver = 3, 9 in
     if syntax_ver < ver && Option.is_some (String.index_from extension 1 '.')
@@ -152,8 +152,8 @@ let print_ast { file_kinds; _ } ml_kind =
 
 let merlin_reader { file_kinds; _ } ml_kind =
   let open Option.O in
-  let* x = Ml_kind.Dict.get file_kinds ml_kind in
-  let+ _, merlin_reader = x.merlin_reader in
+  let* dialect = Ml_kind.Dict.get file_kinds ml_kind in
+  let+ _, merlin_reader = dialect.merlin_reader in
   merlin_reader
 ;;
 
--- a/src/dune_rules/dialect.ml
+++ b/src/dune_rules/dialect.ml
@@ -305,38 +305,40 @@ module DB = struct
     }
   ;;
 
-  let compute_for_merlin by_name =
-    let extensions, readers =
-      String.Map.fold
-        by_name
-        ~init:([], String.Map.empty)
-        ~f:(fun d (extensions, readers) ->
-          let handle_ml_kind kind readers =
-            let ext = extension d kind in
-            if ext = extension ocaml kind
-            then (* this is standard dialect, exclude *) None, readers
-            else (
-              let reader = merlin_reader d kind in
-              match ext, reader with
-              | Some ext, Some reader -> Some ext, String.Map.add_exn readers ext reader
-              | _ ->
-                if preprocess d kind <> None
-                then (* we have preprocessor defined *) None, readers
-                else ext, readers)
-          in
-          let impl, readers = handle_ml_kind Ml_kind.Impl readers in
-          let intf, readers = handle_ml_kind Ml_kind.Intf readers in
-          let extensions =
-            match impl, intf with
-            | None, None -> extensions
-            | _ -> { Ml_kind.Dict.impl; intf } :: extensions
-          in
-          extensions, readers)
+  let compute_for_merlin =
+    let handle_ml_kind ~dialect kind readers =
+      let ext = extension dialect kind in
+      if ext = extension ocaml kind
+      then (* this is standard dialect, exclude *) None, readers
+      else (
+        match ext, merlin_reader dialect kind with
+        | Some ext, Some reader -> Some ext, String.Map.add_exn readers ext reader
+        | _ ->
+          if preprocess dialect kind <> None
+          then (* we have preprocessor defined *) None, readers
+          else ext, readers)
     in
-    let extensions =
-      List.sort ~compare:(Ml_kind.Dict.compare (Option.compare String.compare)) extensions
-    in
-    { extensions; readers }
+    fun by_name ->
+      let extensions, readers =
+        String.Map.fold
+          by_name
+          ~init:([], String.Map.empty)
+          ~f:(fun dialect (extensions, readers) ->
+            let impl, readers = handle_ml_kind ~dialect Ml_kind.Impl readers in
+            let intf, readers = handle_ml_kind ~dialect Ml_kind.Intf readers in
+            let extensions =
+              match impl, intf with
+              | None, None -> extensions
+              | _ -> { Ml_kind.Dict.impl; intf } :: extensions
+            in
+            extensions, readers)
+      in
+      let extensions =
+        List.sort
+          ~compare:(Ml_kind.Dict.compare (Option.compare String.compare))
+          extensions
+      in
+      { extensions; readers }
   ;;
 
   let for_merlin t = Lazy.force t.for_merlin
--- a/src/dune_rules/dialect.ml
+++ b/src/dune_rules/dialect.ml
@@ -292,8 +292,8 @@ module DB = struct
     }
 
   and for_merlin =
-    { extensions : string option Ml_kind.Dict.t list
-    ; readers : string list String.Map.t
+    { extensions : Filename.Extension.t option Ml_kind.Dict.t list
+    ; readers : Filename.Extension.t list String.Map.t
     }
 
   let fold { by_name; _ } = String.Map.fold by_name
--- a/src/dune_rules/merlin/merlin.ml
+++ b/src/dune_rules/merlin/merlin.ml
@@ -589,13 +589,13 @@ module Unprocessed = struct
       (* And copy for each module the resulting pp flags *)
       modules
       |> Modules.With_vlib.drop_vlib
-      |> Modules.fold_no_vlib modules ~init:[] ~f:(fun m init ->
-        Module.orig_sources m
+      |> Modules.fold ~init:[] ~f:(fun m init ->
+        Module.sources_without_pp m
         |> Path.Build.Set.of_list_map ~f:(fun src -> Path.as_in_build_dir_exn src)
         |> Path.Build.Set.fold ~init ~f:(fun src acc ->
           let config =
             { Processed.module_ = Module.set_pp m None
-            ; opens = Modules.local_open modules m
+            ; opens = Modules.With_vlib.local_open modules m
             ; reader = String.Map.find readers (Path.Build.extension src)
             }
           in
--- a/src/dune_rules/merlin/merlin.ml
+++ b/src/dune_rules/merlin/merlin.ml
@@ -600,7 +600,7 @@ module Unprocessed = struct
             }
           in
           (* we add the config with and without the extension, the latter is
-             needed for a fallback in get *)
+             needed for a fallback in this file's [get] function. *)
           let src_without_extension = remove_extension src in
           (src, config) :: (src_without_extension, config) :: acc))
       |> Path.Build.Map.of_list_reduce ~f:(fun existing _ -> existing)
--- a/src/dune_rules/merlin/merlin.ml
+++ b/src/dune_rules/merlin/merlin.ml
@@ -257,16 +257,12 @@ module Processed = struct
            (* Fallback to handle preprocessed files (where the preprocessor has
               the file extensison changed).
 
-              We choose to fallback by a lookup with file_without_ext.ml e.g.
-              stripping all possible file extensions prefixes.
+              We choose to fallback by a lookup by filename without extension.
 
               This is too rough but, really, preprocessors should emit copy
               line directives instead and then Dune should have the database
               similar to Copy_line_directive to handle this. *)
-           let file_without_ext = remove_extension file in
-           Path.Build.Map.find
-             per_file_config
-             (Path.Build.set_extension file_without_ext ~ext:".ml"))
+           Path.Build.Map.find per_file_config (remove_extension file))
     in
     let pp = Module_name.Per_item.get pp_config (Module.name module_) in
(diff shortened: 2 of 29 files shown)


rgrinberg:
cc @nojb who is the main author of the dialects feature.

Could you add some tests please? See our merlin tests for how we test the communication with merlin.

nojb:
Thanks for the ping @rgrinberg. @andreypopp: I am not familiar with the READER feature. Do you mind explaining what it is about? Thanks!

nojb:
> @nojb in merlin there's a way to plug a custom "reader" which allows to customise parsing (a few other features). The usually way to implement reader is to use https://github.com/let-def/merlin-extend SDK.

Thanks. Do I understand correctly that a reader is an executable? In any case, it seems like a coherent addition to the `(dialect)` stanza.

src/dune_rules/dialect.ml
```diff
@@ -78,13 +81,14 @@ let decode =
       field_o
         "format"
         (map ~f:(fun (loc, x) -> loc, x, []) (located Action.decode_dune_file))
+    and+ merlin_reader = field_o "merlin_reader" (located (repeat1 string))
```
nojb, line 84: This field should be versioned.

andreypopp:
> Thanks. Do I understand correctly that a reader is an executable?

The reader itself is an executable, yes, the configuration though specifies just a reader's NAME, then merlin seaches for `ocamlmerlin-NAME` in `$PATH`.

src/dune_rules/dialect.ml
```diff
@@ -78,13 +81,24 @@ let decode =
...
+        Syntax.Error.since loc Stanza.syntax ver ~what)
+    in
+    let () =
+      let ver = 3, 10 in
+      match syntax_ver < ver, merlin_reader with
+      | true, Some (loc, _) ->
+        let what = "the (merlin_reader) stanza" in
+        Syntax.Error.since loc Stanza.syntax ver ~what
+      | _ -> ()
+    in
```
nojb, line 100: The error handling check that you imitated here is specially tricky; in your case you can do something simpler, such as in https://github.com/ocaml/dune/blob/0219fad9ae3d7e45e5cd37616bfb126b9c96d802/src/dune_rules/dune_file.ml#L76-L80

andreypopp:
Actually I'm wondering if we should allow to define a reader per `dialect` or per `dialect * ml_kind`?

I've made it per `dialect * ml_kind` as merlin maps extensions to readers but, in practice I think, both implementations and interfaces will be handled by a single executable so maybe we should move this stanza up to be per `dialect`? What do you think?

src/dune_rules/dialect.ml
```diff
@@ -227,21 +251,26 @@ module DB = struct
   type t =
     { by_name : dialect String.Map.t
     ; by_extension : dialect String.Map.t
-    ; mutable extensions_for_merlin : string option Ml_kind.Dict.t list option
+    ; for_merlin : for_merlin Lazy.t
+    }
+
+  and for_merlin =
+    { extensions : string option Ml_kind.Dict.t list
+    ; extension_to_reader : (string * string list) list
```
nojb, line 259: If I understand correctly, this is supposed to map file extensions to lists of "readers". Wouldn't it be better to use a Map here? Also, I find the name a bit unintuitive; what about simply `readers`?

andreypopp: Can use map but in practice the number of dialects is negligible I think...

nojb:
> What do you think?

I think the current approach (`dialect * ml_kind`) gives a bit more flexibility and stays closer to the interface exposed by Merlin, so I would leave it as it is.

src/dune_rules/dialect.ml
nojb: I wasn't suggesting it because of the number of dialects, rather because it more clearly exposes the "map" semantics of the field (eg the keys are unique).

andreypopp: I see, that's what the field name was supposedly to clarify. Anyway, I've changed it as suggested (and the version check as well).

src/dune_rules/merlin/merlin.ml
```diff
@@ -51,17 +51,22 @@ module Processed = struct
     ; src_dirs : Path.Set.t
     ; flags : string list
     ; extensions : string option Ml_kind.Dict.t list
+    ; extension_to_reader : string list String.Map.t
```
nojb, line 54: Shall we call this field `readers` as well?

andreypopp: Yes, sorry, just pushed a commit fixing this.

src/dune_rules/dialect.ml
```diff
@@ -253,17 +275,24 @@ module DB = struct
...
     in
-    t.extensions_for_merlin <- Some v;
-    v
+    let readers =
+      String.Map.fold by_name ~init:String.Map.empty ~f:(fun d s ->
+        let add kind s =
+          match merlin_reader d kind with
+          | None -> s
+          | Some (extension, reader) ->
+            (* Ok to use [add_exn] here as we've validating below that we have
```
nojb, line 284: Do you mean `above` here instead of `below`?

nojb (approved):
I didn't looks in full detail, but generally speaking this LGTM

src/dune_rules/dialect.ml
andreypopp: The code which validates extension is registered once is [below](https://github.com/ocaml/dune/pull/8567/files/9669feffd1a3b66e27a7783e5a2c6ed41cd61f8b#diff-e77f32927aca7df771b4e00ecc3315a416133fdc68e1c9133139a60ef399fec1R310). I'll add a more specific reference though, good catch.

src/dune_rules/merlin/merlin.ml
```diff
@@ -394,7 +439,7 @@ module Unprocessed = struct
       Path.Set.singleton @@ obj_dir_of_lib `Private mode (Obj_dir.of_local obj_dir)
     in
     let flags = Ocaml_flags.get flags mode in
-    let extensions = Dialect.DB.extensions_for_merlin dialects in
+    let for_merlin = Dialect.DB.for_merlin dialects in
```
nojb, line 442: Nit: could use a record pattern here `let {Dialect.DB.extensions; for_merlin} = ...` and use punning below.

src/dune_rules/merlin/merlin.ml
```diff
@@ -257,25 +286,33 @@ module Processed = struct
...
+      in
+      let pp_module { module_; opens } =
         let name = Module.name module_ in
         let pp = Module_name.Per_item.get pp_config name in
-        let sexp = to_sexp ~opens ~pp config in
-        Pp.vbox (Pp.text (Module_name.to_string name))
-        ++ Pp.newline
-        ++ Pp.vbox (Sexp.pp sexp)
+        let sexps =
+          List.map (Module.sources module_) ~f:(fun path ->
```
andreypopp, line 304: So this is working incorrectly now as it gets `.ml` suffixed versions, e.g for `App.mlx` it will get `App.mlx.ml` and `App.mlx.mli`.

1. I'm not sure how to fix that... seems like after `Module.ml_source` we don't track the original modules so no info about the dialect or original file name is present here.
2. As I understand this command is purely for tests/debug, maybe we should make it iterate over source files and "modules"?

andreypopp: Just to clarify, in 2. above I'm suggesting to remove `print_file` completely and make `dune ocaml merlin dump-config` work by listing the directory and then querying a config for each source file found.

nojb: I'm not familiar with this part of the code, @rgrinberg or @voodoos may know.

voodoos: I think this would simplified by moving the reader info the the part of the config file that is module-specific instead of the part that is common for all modules of a stanza.

See my comment here: https://github.com/ocaml/dune/pull/8567/files#r1314880623

src/dune_rules/merlin/merlin.ml
```diff
@@ -51,17 +51,21 @@ module Processed = struct
     ; src_dirs : Path.Set.t
     ; flags : string list
     ; extensions : string option Ml_kind.Dict.t list
+    ; readers : string list String.Map.t
```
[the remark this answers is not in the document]
andreypopp: I see, this indeed simplifies things but it doesn't solve the issue with `dune ocaml merlin dump-config` — the place where we create per module configs already sees preprocessed modules only.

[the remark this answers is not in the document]
andreypopp: So in [4080abe](https://github.com/ocaml/dune/pull/8567/commits/4080abe502d60d750a4d2db482c5d36ca5463890) I've refactored it to pass source modules when configuring merlin.

This seems to me the right thing to do conceptually — as merlin works on source files and thus it makes sense to produce configuration for what's in source/user written modules.


But not sure if I broke some edge cases? From my smoke testing it works fine both for files with dialects and for regular .ml files.

andreypopp:
Made the tests pass. Let me know if there's any more I need to fix/tweak before this can be merged.

src/dune_rules/merlin/merlin.ml
```diff
@@ -1,9 +1,13 @@
...
+let strip_pp_extensions file =
   let dir = Path.Build.parent_exn file in
-  let basename, _ext = String.lsplit2_exn (Path.Build.basename file) ~on:'.' in
-  Path.Build.relative dir basename
+  let basename = Path.Build.basename file in
+  match String.split basename ~on:'.' with
+  | [] | [ _ ] | [ _; _ ] -> file
+  | name :: exts ->
+    let ext = Option.value_exn (List.last exts) in
+    Path.Build.relative dir (sprintf "%s.%s" name ext)
```
voodoos, line 10: I still don't understand why you need to change this function and it's usage at line L253/R273.
Could you document its behavior and why it's needed ?

andreypopp: This is to accommodate [this](https://github.com/andreypopp/dune/blob/10faea361ac82756ebd63a52d8ef66e45270923e/test/blackbox-tests/test-cases/merlin/server.t/run.t#L17-L20) test case.

Before this change we were tracking only config per module name (so extensions were stripped) but now, as we introduce readers which are per extension, we now track config per source file (with extensions).

We still want to treat `module_name.xx.xxx.ext` as `module_name.ext` though as the linked test case specifies, thus the function `strip_pp_extension` normalises a filename by stripping extensions added by a preprocessor.

voodoos: I think I finally understand. I initially though you had the same constraints as preprocessing specification which are per-module (that is, the same for the `ml` and the `mli` file of a module).

But readers can be different for the `ml` file and the `mli` file of a same module ?

andreypopp: > But readers can be different for the ml file and the mli file of a same module ?

You mean for implementation and interface? Yes, I've raise the same question [here](https://github.com/ocaml/dune/pull/8567#issuecomment-1702857875).

But note though that for a single module an implementation might belong to a dialect while the interface could stay pure `.ml` — e.g. `.mlx` and `.mli`. So we still need to track config per file with extension even if we change `(merlin_reader)` to be same both for implementations and interfaces within a dialect.

src/dune_rules/merlin/merlin.ml
```diff
@@ -247,35 +272,43 @@ module Processed = struct
...
        -> foo *)
     let open Option.O in
-    let+ { module_; opens } =
+    let+ { module_; opens; reader } =
       let find file =
-        let file_without_ext = remove_extension file in
-        Path.Build.Map.find per_module_config file_without_ext
+        match Path.Build.Map.find per_module_config file with
+        | Some _ as s -> s
+        | None -> Path.Build.Map.find per_module_config (strip_pp_extensions file)
```
voodoos, line 279: Why do you need this fallbacking mechanism ?

src/dune_rules/merlin/merlin.ml
voodoos: I am not very familiar with Merlin's handling of alternative readers, ~but it looks like the suffix map would allow to use a different reader for the interface and the implementation while keeping a unique configuration per module.~ In fact it's only available as a flag, not a configuration option. So having a configuration for each source file sounds like the easiest way to do it. 👍

src/dune_rules/merlin/merlin.ml
andreypopp: Modules with pp extensions `module_name.xxx.xx.ext` can be also valid modules residing in source directories (e.g. extra extension were not added by a preprocessor but present in source files). In this case it makes sense to try to resolve using non normalised filename first, I think.

voodoos: Do you have a test illustrating that ? In any case, I would like to take some time next week to pull your changes and understand better the relationship between source file / module / merlin config. It's a set of conventions that would benefit a lot from being correctly defined and documented.

The PR looks good otherwise 👍

andreypopp: No, don't have a test case...

1. I'm actually wondering if we should remove any normalisation now that we have per filename config (rather per module) and such config is built by looking at source modules (so we see "real" source files).

1. Another thing to check (add test) is we still support promoted files.

For reference the behaviour of stripping the extension was added in c36df0b73ffa98f893788b25ce68139232299e41

rgrinberg:
@voodoos are you happy with this PR being merged to be available in 3.11? Or do you want to take another look and have it for 3.12?

voodoos:
> @voodoos are you happy with this PR being merged to be available in 3.11? Or do you want to take another look and have it for 3.12?

It's probably all right but I would like to do a last pass and add a few test cases.
What is the schedule for 3.11 ? (I can do that tomorrow)

rgrinberg:
I think @emillon wants it out by end of week.

voodoos (dismissed):
I pushed a commit with a new test that focuses on Merlin's config's levels of granularity.

Most standard use cases looks good. There is one change of behavior compared to before: source files with a custom extension that changed after preprocessing are not handled anymore. For example:

Given a file, and a preprocessing action:
`modname.something.cml --pp--> modname.ml` 

One could expect to get the configuration for the module `modname` when running merlin on the `modname.something.cml`.  

Before that PR the granularity was at the module level and a simple heuristic was used to match a source-file with a module: split the filename by `.`, the first segment must be the module name.

Now the granularity is finer, and there is no way, given the source file `modname.something.cml` to guess that the appropriate config is stored under the key `modname.ml`. The new heuristic is: split the filename by `.`, the first segment must be the module name, the last segment must be the extension.

1. I agree that the correct thing to do, as you suggested, @andreypopp, is to use real source-file names as keys:
> I'm actually wondering if we should remove any normalization now that we have per filename config (rather per module) and such config is built by looking at source modules (so we see "real" source files).

2. and adding test for promoted source files sounds important too :-) 

3. Additionally, the currently provided suffixes for melange are `mlx mlx`. Isn't there a different suffix for signatures ? If `mli` are to be used, then I think the list should be `mlx mli` for Merlin jump to declaration to work as expected.

src/dune_rules/merlin/merlin.ml
voodoos: > Modules with pp extensions module_name.xxx.xx.ext can be also valid modules residing in source directories (e.g. extra extension were not added by a preprocessor but present in source files). In this case it makes sense to try to resolve using non normalised filename first, I think.

Added a test for that in my commit (at the end of the file), and it doesn't work right now.

> Another thing to check (add test) is we still support promoted files.
That's a good idea, can you do it ? (you can reuse the `merlin_config.sh` script in my new test to easily query the configuration.

dmmulroy:
Are there any updates on this merging? I would love to start seriously utilizing, advocating, and teaching mlx

andreypopp:
@voodoos I apologise for a prolonged pause, but I've just returned to this PR

I've made the tests cases you've provided pass (thank you for the test cases!) but... it's not that simple.

As I understand it's not possible now to track in dune properly sources for the preprocessed files — actions can be be of any complexity and I imagine a final .ml can have multiple deps/nputs. Not clear which one to treat as a source.

So instead I suggest the following approach:

- we track original source files in `Module.t`
- we construct merlin config by source file and not by module
- we do a look up by the proper filename (with extension)
- ... but we still have a fallback to lookup by `file_without_ext.ml` key — that's exactly to handle preprocessed files with fancy extensions

I don't think this is perfect but it seems to work... 

The ideal solution would be to build a database which would be able to do a lookup by line directives, I think. Similar to how `Copy_line_directive` work but for arbitrary modules, not just those which were `copy#`-ed.

andreypopp:
in ae596bff74803f9db227607a0619b288f8e09856 I've removed storing `Module.File.orig_path` in `dune-package` but I actually not sure what are the consequences of this change

andreypopp:
in [339d212](https://github.com/ocaml/dune/pull/8567/commits/339d21247cae01ca4630cb64ae3ce9c445bb7041) I've changed the fallback to work like this:
- store same config under original filename and filename w/o any extension as keys
- do a fallback lookup by filename w/o extension

I think this is more robust than doing a fallback `.ml` as extension (not sure it'll always be present)

src/dune_rules/dialect.ml
```diff
@@ -33,17 +34,19 @@ module File_kind = struct
                 "print_ast"
                 Action.encode
                 (Option.map ~f:(fun (_, x) -> x) print_ast)
+            ; field_o "merlin_reader" (list string) (Option.map ~f:snd merlin_reader)
```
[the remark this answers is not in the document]
anmonteiro: I think we should expect to find:

- 1 dialect per extension
- 1 reader per dialect

what am I missing in this invariant?

src/dune_rules/merlin/merlin.ml
```diff
@@ -563,22 +586,25 @@ module Unprocessed = struct
...
           let config =
             { Processed.module_ = Module.set_pp m None
             ; opens = Modules.local_open modules m
+            ; reader = String.Map.find readers (Path.Build.extension src)
             }
           in
-          (src, config) :: acc))
-      |> Path.Build.Map.of_list_exn
+          (* we add the config with and without the extension, the latter is
+             needed for a fallback in get *)
```
anmonteiro, line 602:
```suggestion
             needed for a fallback in this file's [get] function. *)
```

src/dune_rules/dialect.ml
```diff
@@ -271,43 +288,60 @@ module DB = struct
...
+                else ext, readers)
+          in
+          let impl, readers = handle_ml_kind Ml_kind.Impl readers in
+          let intf, readers = handle_ml_kind Ml_kind.Intf readers in
+          let extensions =
+            match impl, intf with
+            | None, None -> extensions
+            | _ -> { Ml_kind.Dict.impl; intf } :: extensions
+          in
+          extensions, readers)
```
anmonteiro, line 334: we can extract `handle_ml_kind` to the outside scope.
```suggestion
let compute_for_merlin =
  let handle_ml_kind ~dialect kind readers =
    let ext = extension dialect kind in
    if ext = extension ocaml kind
    then (* this is standard dialect, exclude *) None, readers
    else (
      match ext, merlin_reader dialect kind with
      | Some ext, Some reader -> Some ext, String.Map.add_exn readers ext reader
      | _ ->
        if preprocess dialect kind <> None
        then (* we have preprocessor defined *) None, readers
        else ext, readers)
  in
  fun by_name ->
    let extensions, readers =
      String.Map.fold
        by_name
        ~init:([], String.Map.empty)
        ~f:(fun dialect (extensions, readers) ->
          let impl, readers = handle_ml_kind ~dialect Ml_kind.Impl readers in
          let intf, readers = handle_ml_kind ~dialect Ml_kind.Intf readers in
          let extensions =
            match impl, intf with
            | None, None -> extensions
            | _ -> { Ml_kind.Dict.impl; intf } :: extensions
          in
          extensions, readers)
```

test/blackbox-tests/test-cases/merlin/dialect.t/dune-project
```diff
@@ -0,0 +1,10 @@
+(lang dune 3.11)
```
anmonteiro, line 1:
```suggestion
(lang dune 3.16)
```

test/blackbox-tests/test-cases/merlin/granularity.t
```diff
@@ -0,0 +1,178 @@
+  $ ocamlc_where="$(ocamlc -where)"
+  $ export BUILD_PATH_PREFIX_MAP="/OCAMLC_WHERE=$ocamlc_where:$BUILD_PATH_PREFIX_MAP"
+
+# A utility to query merlin configuration for a file:
```
anmonteiro, line 4:
```suggestion
A utility to query merlin configuration for a file:
```

test/blackbox-tests/test-cases/merlin/granularity.t
```diff
@@ -0,0 +1,178 @@
...
+  $ cat >merlin_conf.sh <<EOF
+  > #!/bin/sh
+  > FILE=\$1
+  > printf "(4:File%d:%s)" \${#FILE} \$FILE | dune ocaml-merlin \
+  >   | sed -E "s/[[:digit:]]+:/\?:/g" | sed -E "s/\)(\(+)/\n\1/g"
+  > EOF
+
+  $ chmod a+x merlin_conf.sh
+
+# Project sources
```
anmonteiro, line 14:
```suggestion
Project sources
```

test/blackbox-tests/test-cases/merlin/granularity.t
```diff
@@ -0,0 +1,178 @@
...
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.11)
```
anmonteiro, line 17:
```suggestion
  > (lang dune 3.16)
```

test/blackbox-tests/test-cases/merlin/granularity.t
```diff
@@ -0,0 +1,178 @@
...
+  > EOF
+
+  $ dune build @check
+  $ dune exec ./test.exe
+  42
+  43
+  44
+  45
+
+# We now query Merlin configuration for the various source files:
```
anmonteiro, line 92:
```suggestion
We now query Merlin configuration for the various source files:
```

test/blackbox-tests/test-cases/merlin/granularity.t
```diff
@@ -0,0 +1,178 @@
...
+
+Some configuration can be specific to a module like preprocessing.
+
+Dialects are specified by extensions so are specific to a file. This means a
+different reader might be used for the signature and the implementation.
+
+Note that suffixes provided by dialect should always be told to Merlin for
+locate to work correctly. 
+
+## Preprocessing:
```
anmonteiro, line 105:
```suggestion
Preprocessing:
```

test/blackbox-tests/test-cases/merlin/granularity.t
```diff
@@ -0,0 +1,178 @@
...
+Some configuration fields are common to all the modules of a same stanza. This
+is the case for the stdlib, build and sources directories, flags and suffixes.
+
+Some configuration can be specific to a module like preprocessing.
+
+Dialects are specified by extensions so are specific to a file. This means a
+different reader might be used for the signature and the implementation.
+
+Note that suffixes provided by dialect should always be told to Merlin for
+locate to work correctly.
```
anmonteiro, line 103:
```suggestion
Note that Merlin should always be told about dialect-provided suffixes, to make `MerlinLocate` work correctly.
```

test/blackbox-tests/test-cases/merlin/granularity.t
```diff
@@ -0,0 +1,178 @@
...
+  (?:B?:$TESTCASE_ROOT/_build/default/.test.eobjs/byte
+  (?:S?:$TESTCASE_ROOT
+  (?:FLG(?:-open?:Dune__exe)
+  (?:FLG(?:-pp?:$TESTCASE_ROOT/_build/default/pp.sh)
+  (?:FLG(?:-w?:@1..3@5..28@30..39@43@46..47@49..57@61..62@67@69-?:-strict-sequence?:-strict-formats?:-short-paths?:-keep-locs?:-no-strict-formats?:-g)
+  (?:SUFFIX?:.mlx .mlx))
+
+  $ ./merlin_conf.sh pped.mli | diff pped.out -
+
+## Melange:
```
anmonteiro, line 120:
```suggestion
Melange:
```

test/blackbox-tests/test-cases/merlin/granularity.t
```diff
@@ -0,0 +1,178 @@
...
+  7c7,8
+  < (?:SUFFIX?:.mlx .mlx))
+  \ No newline at end of file
+  ---
+  > (?:SUFFIX?:.mlx .mlx
+  > (?:READER(?:mlx)))
+  \ No newline at end of file
+  [1]
+
+## Unconventionnal file names:
```
anmonteiro, line 143:
```suggestion
Unconventional file names:
```

test/blackbox-tests/test-cases/merlin/granularity.t
```diff
@@ -0,0 +1,178 @@
...
+
+Users might have preprocessing steps that start with a non-conventionnal
```
anmonteiro, line 145:
```suggestion
Users might have preprocessing steps that start with a non-conventional
```

test/blackbox-tests/test-cases/merlin/granularity.t
```diff
@@ -0,0 +1,178 @@
...
+  (?:EXCLUDE_QUERY_DIR
+  (?:B?:$TESTCASE_ROOT/_build/default/.test.eobjs/byte
+  (?:S?:$TESTCASE_ROOT
+  (?:FLG(?:-open?:Dune__exe)
+  (?:FLG(?:-w?:@1..3@5..28@30..39@43@46..47@49..57@61..62@67@69-?:-strict-sequence?:-strict-formats?:-short-paths?:-keep-locs?:-no-strict-formats?:-g)
+  (?:SUFFIX?:.mlx .mlx))
+
+  $ ./merlin_conf.sh cppomod.ml | diff cppomod.out -
+
+Note that this means unreleated files might be given the same configuration:
```
anmonteiro, line 163:
```suggestion
Note that this means unrelated files might be given the same configuration:
```

test/blackbox-tests/test-cases/merlin/granularity.t
```diff
@@ -0,0 +1,178 @@
...
+
+  $ ./merlin_conf.sh cppomod.tralala.ml | diff cppomod.out -
+
+And with unconventionnal extension:
```
anmonteiro, line 167:
```suggestion
And with unconventional extension: 
```

test/blackbox-tests/test-cases/merlin/granularity.t
```diff
@@ -0,0 +1,178 @@
...
+Is it expected that the suffix for implementation and interface is the same ?
+  $ ./merlin_conf.sh pped.ml | tee pped.out
+  ((?:STDLIB?:/OCAMLC_WHERE
+  (?:EXCLUDE_QUERY_DIR
+  (?:B?:$TESTCASE_ROOT/_build/default/.test.eobjs/byte
+  (?:S?:$TESTCASE_ROOT
+  (?:FLG(?:-open?:Dune__exe)
+  (?:FLG(?:-pp?:$TESTCASE_ROOT/_build/default/pp.sh)
+  (?:FLG(?:-w?:@1..3@5..28@30..39@43@46..47@49..57@61..62@67@69-?:-strict-sequence?:-strict-formats?:-short-paths?:-keep-locs?:-no-strict-formats?:-g)
+  (?:SUFFIX?:.mlx .mlx))
```
anmonteiro, line 116: what does `(?:SUFFIX?:.mlx .mlx))` mean here? Is `pped.ml` getting read by merlin as if it belonged to the mlx dialect?

anmonteiro:
@andreypopp took a first look at this change. The approach seems fine, but I have a few questions throughout. I think the main thing I don't understand is why `merlin_reader` is parsed into a list of strings. Is there a case where we would pass multiple readers in that field?

test/blackbox-tests/test-cases/merlin/granularity.t
andreypopp: merlin uses `SUFFIX` config to determine if this is an implementation or an interface document

now that `.mlx .mlx` is not completely correct as mlx dialect doesn't really have an interface counterpart to `.mlx` and it is supposed to use `.mli` instead

@voodoos suggested [here](https://github.com/ocaml/dune/pull/8567#pullrequestreview-1624199713) to use `.mlx .mli` which I want to do but in a separate PR (the code is [here](https://github.com/andreypopp/dune/blob/a5fc103dda82cffff1985dcfff8955ff035b12be/src/dune_rules/merlin/merlin.ml#L144-L150)) as it seems another kind of change.

src/dune_rules/dialect.ml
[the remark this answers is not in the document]
andreypopp: the list represents a program name and a list of arguments, so it's a single reader being defined

andreypopp:
The remaining test failures look like:
```
-  (?:FLG(?:-w?:@1..3@5..28@30..39@43@46..47@49..57@61..62@67@69-?:-strict-sequence?:-strict-formats?:-short-paths?:-keep-locs?:-no-strict-formats?:-g)
+  (?:FLG(?:-w?:@1..3@5..28@31..39@43@46..47@49..57@61..62@67@69-?:-strict-sequence?:-strict-formats?:-short-paths?:-keep-locs?:-no-strict-formats?:-g)
```

note `-w ...28@30...` vs ``-w ...28@31...` and it's changed only since I've bumped dune lang to 3.16... not sure why it has changed...

src/dune_rules/module.ml
```diff
@@ -3,6 +3,9 @@ open Import
 module File = struct
   type t =
     { path : Path.t
+    ; orig_path : Path.t
```
rgrinberg, line 6: `original_path`

src/dune_rules/module.mli
```diff
@@ -88,6 +88,7 @@ module Obj_map : sig
 end
 
 val sources : t -> Path.t list
+val orig_sources : t -> Path.t list
```
rgrinberg, line 91: `original_sources`. Or better yet, `sources_without_pp`. The originality of a source isn't descriptive without context

andreypopp:
Thanks for the review, @rgrinberg! Tweaked the naming and also rebased on top of `main` to solve conflicts.

andreypopp:
> note -w ...28@30... vs -w ...28@31... and it's changed only since I've bumped dune lang to 3.16... not sure why it has changed...

Javi pointed me to: https://github.com/ocaml/dune/blob/bc57ae438eca9f22f1733dbd58df83e22ee7eaaf/src/dune_rules/ocaml_flags.ml#L34-L39 so I've promoted the tests for merlin granularity.

anmonteiro (approved):
This looks good to me now. 

@andreypopp feel free to tag me on the next PR about `mlx/mlx` (merlin suffix) too.

andreypopp:
Thanks @anmonteiro!

I've added another test case for promoted modules.

@rgrinberg @voodoos please let me know if anything else should be done with this PR or we can merge it?

rgrinberg:
No issues from me - though I haven't looked at it closely. @anmonteiro feel free to merge if you're satisfied and if the CI failures aren't related.

anmonteiro:
Let's do this. Thanks for your continued work on this, @andreypopp

PR #8577, opened 2023-09-02
fix(pkg): stop using physical equality

rgrinberg:
Makes memo go into an infinite loop


snowleopard:
> Makes memo go into an infinite loop

Just to clarify: the infinite loop here was happening only when hitting dependency cycles (according to a discussion with Rudi elsewhere).

PR #8588, opened 2023-09-04
bug: Add minimal reproduction case for invalid pform path in `dune build`

Leonidas-from-XIV:
This PR adds a minimal test case for an error noticed by @rgrinberg that when using `dune build` on a pform, it tries to build an invalid path and fails.

Initial debugging has shown that the path that is being expanded is correct, so the root case is probably some post-processing of the expanded `Path.t`.

### DIFF
--- a/test/blackbox-tests/test-cases/build-bin.t
+++ b/test/blackbox-tests/test-cases/build-bin.t
@@ -1,0 +1,72 @@
+Pform expansion should work in `exec` but also in `build`:
+
+  $ cat > dune-project << EOF
+  > (lang dune 3.10)
+  > (package (name public))
+  > EOF
+  $ cat > dune << EOF
+  > (executable (public_name public) (modules Public))
+  > (executable (name private) (modules Private))
+  > (env
+  >   (_ (binaries (private.exe as priv))))
+  > EOF
+
+The executables in question just display what they are:
+
+  $ cat > public.ml << EOF
+  > let () = print_endline "Public"
+  > EOF
+  $ cat > private.ml << EOF
+  > let () = print_endline "Private"
+  > EOF
+
+We also have a subfolder `foo`
+
+  $ mkdir foo
+  $ cat > foo/dune << EOF
+  > (executable (public_name blah))
+  > (env
+  >   (_ (binaries (blah.exe as foo))))
+  > EOF
+  $ cat > foo/blah.ml << EOF
+  > let () = print_endline "Blah"
+  > EOF
+
+Making Dune build the executables should work without error messages:
+
+  $ dune build %{bin:public}
+  Error: File unavailable:
+  $TESTCASE_ROOT/../install/default/bin/public
+  [1]
+  $ ls _build/default/public.exe
+  _build/default/public.exe
+
+Building the private executable should also work.
+
+  $ dune build %{bin:private}
+  File "command line", line 1, characters 0-14:
+  Error: Program private not found in the tree or in PATH
+   (context: default)
+  [1]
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/build-bin.t
```diff
@@ -0,0 +1,22 @@
...
+
+  $ cat > e.ml << EOF
+  > let () = print_endline "Hello"
+  > EOF
+
+Making Dune build `e` should work:
+
+  $ dune build %{bin:e}
+  Error: File unavailable:
+  $TESTCASE_ROOT/../install/default/bin/e
```
rgrinberg, line 21: I think the issue here is that the path instead of `$TESTCASE_ROOT` (`Path.root`) should be the root of the context. E.g. `Context_name.build_dir $context`.

test/blackbox-tests/test-cases/build-bin.t
```diff
@@ -0,0 +1,22 @@
+Pform expansion should work in `exec` but also in `build`:
+
+  $ cat > dune-project << EOF
+  > (lang dune 3.10)
+  > (package (name e))
+  > EOF
+  $ cat > dune << EOF
+  > (executable (public_name e))
```
rgrinberg, line 8: You should also test the case where the binary isn't public but is used in made public using `(binaries ..)` in the env stanza.

Leonidas-from-XIV: This reveals a rather interesting behavior, as `dune exec %{bin:private}` works without issues, but using it in `env` triggers an error that the target (not the source!) doesn't have a rule:

```
  File "dune", line 4, characters 30-34:
  4 |   (_ (binaries private.exe as priv)))
                                    ^^^^
  Error: No rule found for priv
```

test/blackbox-tests/test-cases/build-bin.t
```diff
@@ -0,0 +1,44 @@
...
+
+Building the private executable should also work.
+
+  $ dune build %{bin:private}
+  $ ls _build/default/private.exe
+  _build/default/private.exe
+
+Building the private executable via the env alias should work too:
+
+  $ dune build %{bin:priv}
```
rgrinberg, line 40: By the way, this particular construct should be tested more thoroughly. In particular, `dune build %{bin:priv}` should respect `binaries` in `$CWD`. So, consider the following:

```
$ cat >dune<<EOF
(.. (binaries myexe.exe as bar) ..))
> EOF
$ cat >foo/dune <<EOF
(.. (binaries blah.exe as foo) ..))
> EOF
$ dune build %{bin:bar} # should work
$ dune build %{bin:bar} --root foo/ # should work
$ dune build %{bin:foo} # doesn't work because foo is underneath us
$ dune build %{bin:foo} --root foo/ # should work
```

rgrinberg (approved):

test/blackbox-tests/test-cases/build-bin.t
rgrinberg: Also note that root discovery is disabled while inside tests. So you will need to use `--root` to set the root manually.

test/blackbox-tests/test-cases/build-bin.t
Leonidas-from-XIV: Nevermind, it was missing some parens so it got parsed incorrectly (which is problematic in some way but not relevant to this PR).

test/blackbox-tests/test-cases/build-bin.t
Leonidas-from-XIV: Yes, this was pretty much it: #8594. The other issues need some looking into.

PR #8594, opened 2023-09-05
fix: Wrong path when resolving a target, making `dune exec %{bin:foo}` fail

Leonidas-from-XIV:
This fixes one of the issues documented in #8588. The other code is not calling `Target.expand_path` so it fails on some other issue.


rgrinberg (approved):
Looks like there's some CI issues.

rgrinberg (changes requested):
Some issues in CI

rgrinberg:
Please re-open once the PR is ready

Issue #8556, opened 2023-08-31
DSL for filtering actions with blang


gridbugs:
Writing down my thoughts on a new DSL for dune files that will allow filtering the arguments to actions and actions themselves based on filters expressed with the blang DSL.

This problem comes up when translating opam files to dune lockfiles. Opam has two types of filtering which it applies to commands:

### Filtering entire commands
```
build: [
  ["dune" "subst"] {dev}
]
```

Here the entire command `dune subst` is omitted unless the variable `dev` resolves to `true`.

### Filtering individual terms
```
build: [
  [
    "dune"
    "build"
    "-p"
    name
    "-j"
    jobs
    "@install"
    "@runtest" {with-test}
    "@doc" {with-doc}
  ]
]
```

Here the `@runtest` and `@doc` terms are only included in the command if the variables `with-test` and `with-doc` are `true` respectively.

Note that this filtering even applies to the first term so the program that gets run can be dynamically chosen. E.g.

 ```
[ "make" {os-family != "freebsd"} "gmake" {os-family = "freebsd"} "install" ]
```

Since supporting this behavior in dune will require adding some type of conditional construct it would be nice if it could also handle other types of conditional construct in opam:

### String Conversions

Inside a string in an opam file you can write:
```
"%{var?string-if-true:string-if-false-or-undefined}%"
```

So one string is chosen if the variable is true and the other is chosen if it's false or undefined. The else case is mandatory.

### Package Conjunctions

Boolean package variables may be combined using the following form: `name1+name2+name3:var` is the conjunction of var for each of `name1`, `name2` and `name3`, i.e it is equivalent to `name1:var & name2:var & name3:var`.

## In dune

The most complex opam expression I can think of to exercise all the features we'll need is:
```
build: [
  [ "foo-%{pkg1+pkg2+_:var1?aaa:bbb}%-bar" {var2 & var3 = "baz"} "qux" ] {var4 | var5}
]
```

Dune does currently have a feature for interpolating variables in strings (pforms like "%{os}") but I'm against extending it to handle opam variable expressions like the one in the example involving a string conversion and a package conjunction as dune's pform syntax is too limited to express such a complex term in a readable way.

Breaking down the string interpolation in the example, the condition `pkg1+pkg2+_:var1` could be expressed today with a dune blang:
```
(and %{pkg:var1:pkg1} %{pkg:var1:pkg2} %{pkg-self:var1})
```
[...]

PR #8596, opened 2023-09-06
Experimental string list language

gridbugs:
This is a first attempt at implementing the language described in https://github.com/ocaml/dune/issues/8556. The goal is to allow dune to conditionally exclude terms in its action language which is necessary to support opam build and install commands when building opam packages.


gridbugs:
Also there's a test with some examples of the language here: https://github.com/ocaml/dune/pull/8596/files#diff-aac68abad61a6509586e3fd9c03ede431374b439fa8b1b11e73a0913e1d25d34

rgrinberg:
Looks about right to me. I see that you've introduced `Slang.t` inside `Run` for now. That's ok for now, but we'll probably need to allow it everywhere eventually. For example, I think opam supports conditional patching, so we'll at the very least need that.

src/dune_lang/slang.ml
Alizter, line 1: missing mli

gridbugs:
Tidied this up and added some tests.

src/dune_lang/blang.mli
```diff
@@ -13,19 +13,16 @@ module Op : sig
   val eval : t -> Ordering.t -> bool
 end
 
-type t =
+type 'string t =
```
rgrinberg, line 16: Do we instantiate this type with any other parameter?

src/dune_lang/action.ml
```diff
@@ -259,9 +259,8 @@ let sw = String_with_vars.decode
 let cstrs_dune_file t =
   let open Decoder in
   [ ( "run"
-    , let+ prog = sw
-      and+ args = repeat sw in
-      Run (prog, args) )
+    , let+ args = repeat Slang.decode in
```
Alizter, line 262: Why is this repeated?

src/dune_lang/blang.mli
Alizter: In the definition of Slang.form.

test/blackbox-tests/test-cases/pkg/slang.t
```diff
@@ -0,0 +1,92 @@
...
+  "true" or "false", however it evaluated to "invalid-condition".
+  [1]
+  $ test_action '(run (when (and (< 1 2) (concat complex invalid condition)) echo) foo)'
+  File "dune.lock/test.pkg", line 1, characters 33-67:
+  1 | (install (run (when (and (< 1 2) (concat complex invalid condition)) echo) foo))
+                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  Error: This expression is used as a condition and so must evaluate to either
+  "true" or "false", however it evaluated to "complexinvalidcondition".
+  [1]
+  $ test_action '(run echo (when (and (< 1 2) (or (concat t r u e) (concat f a l s e))) foo) bar)'
```
rgrinberg, line 48: When is the case of of `(or (concat t r u e) ..)` needed?

gridbugs: hopefully never, but boolean literals in blang are just the strings "true" and "false" so this works as a consequence of that.

src/dune_lang/blang.mli
gridbugs: blangs used in the slang language are `Slang.t Blang.t`. This is needed to handle the opam filters with nested conditional statements like:
```
build: [ "foo" { "%{var?bar:baz}%" = "bar" } ]
```
...where `var` is treated as a condition to choose between the strings `"bar"` and `"baz"` as part of a more complex condition `"%{var?bar:baz}%" = "bar"`.

gridbugs:
To help convince myself that this will support all opam filters I'll implement the logic for converting general opam commands into actions here.

src/dune_lang/action.ml
gridbugs: a `Slang.t` is a single term in a command - not the entire command

gridbugs:
I think I found a case that the language can't handle:
```
build: [ "echo" "%{foo+bar:var}%" ]
```
...which would print the value of the boolean expression `foo:var & bar:var`. I still need to confirm whether this is valid in opam. I think we can handle this case by adding an operator which converts blangs into strings.

src/dune_lang/blang.mli
rgrinberg: I see. If I understand this correctly, this expression corresponds to

```
(= (if (defined var) bar baz) bar)
```

rgrinberg: In any case, I do you think you should either

* Introduce your more general type as `Blang.Ast.t`
* Keep what you have, but introduce `Blang.Unexpanded.t` to contain the old type.

test/blackbox-tests/test-cases/pkg/slang.t
rgrinberg: Definitely keep the test case so that we know what's happening. However, I'm a bit concerned by this behavior. Do we really need to implicitly coerce strings to booleans? In other words, do we need this for opam compat.

src/dune_lang/slang.ml
```diff
@@ -0,0 +1,72 @@
+open! Stdune
+open Dune_sexp
+
+type t =
+  | Literal of String_with_vars.t
+  | Form of (Loc.t * form)
```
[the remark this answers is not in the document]
Alizter: IIRC string with vars has them.

test/blackbox-tests/test-cases/pkg/slang.t
gridbugs: This is already how blangs work so I think we should keep it for consistency with other occurrences of blangs in dune.

src/dune_lang/blang.mli
gridbugs: Ok makes sense. I prefer your first suggestion. Also for the record the condition in the example above would be `(= (if (var bar baz) bar)`. Confusingly opam has two different uses for `?`. When it's used as a unary operator in a filter it tests if an expression is undefined but when it's used in a variable interpolation it separates a condition from values to use when the condition is true/false.

test/blackbox-tests/test-cases/pkg/slang.t
rgrinberg: Yes, that is true, but it is something that I rather dislike and would like to change. We could keep the behavior for now, but I would like to make sure we don't rely on it in any deep way. I want us to introduce booleans (and other values) to `Value.t` and have explicit forms to convert between strings and booleans.

gridbugs: Makes sense!

gridbugs:
I've updated this so `Blang.t` just uses `String_with_vars.t` as its string type and introduced a `Blang.Ast` module with the generic versions of blang types and functions. I've updated the opam command conversion implementation to use `Slang.t` for filters and variables and added a bunch more test cases. I added a `Blang` constructor to `Slang.t` for converting bools to strings which is needed for some opam filters that coerce bools into strings.

gridbugs:
Something slightly unusual about the dune actions generated from opam commands is the ternary conditional operator (`%{var?if_true:if_false_or_undefined}%`) in opam gets converted to expressions like this:
```
      (if
       (and
        (all_pkg_vars_defined
         (and %{pkg:installed:foo} %{pkg:installed:bar} %{pkg-self:installed}))
        (and %{pkg:installed:foo} %{pkg:installed:bar} %{pkg-self:installed}))
       x
       y)
```
Note how the condition is duplicated: once to check if it depends on undefined variables with the `all_pkg_vars_defined` operator and then a second time to actually evaluate the condition. This is because the "else" case of opam's ternary operator is chosen when the condition is undefined, but I'm hesitant to add first-class undefined values to dune or to match opam's conditional operator behaviour with dune's if-statements.

To make this more readable, would it be worth adding a new operator to the language that behaves like an if-statement but which follows the "else" case if an undefined variable is encountered while evaluating the condition?

rgrinberg:
I wouldn't worry about it for now. Such expressions are fairly uncommon so the gain in readability is going to be small. Also, the same problem is elegantly addressed with `let` bindings and that's a far more useful feature.

src/dune_rules/pkg_rules.ml
```diff
@@ -601,16 +604,26 @@ module Action_expander = struct
                      (Action.Prog.Not_found.create
                         ~program
                         ~context:t.context
-                        ~loc:(Some loc)
+                        ~loc:None
```
rgrinberg, line 607: Where did the locations go?

src/dune_rules/pkg_rules.ml
```diff
@@ -537,46 +538,48 @@ module Action_expander = struct
...
+        ~dir:(Path.build t.paths.source_dir)
+    ;;
 
-    let expand_exe t sw : ((Path.t, Action.Prog.Not_found.t) result * Value.t list) Memo.t
-      =
+    let expand_exe_value t value =
       let dir = Path.build t.paths.source_dir in
-      let* prog, args = expand_pform_gen t sw ~mode:At_least_one in
       let+ prog =
-        let loc = String_with_vars.loc sw in
```
Alizter, line 575: You need to get this loc back from the value which shouldn't be too hard.

src/dune_rules/pkg_rules.ml
gridbugs: I removed it because this now takes a `Value.t` which doesn't track the location, but I just changed this function to take the location as an argument obtained from the corresponding slang expression. Also added some tests for this error message to make sure they underline the correct part of the file.

gridbugs:
Several of changes:
- added tests of end to end converting opam files to dune and then running their actions to give us more confidence that future changes to the slang encoding of opam commands doesn't change their behaviour
- implemented filtering of command terms (I forgot I hadn't done this yet!)
- change the encoding of boolean ops `&` and `|`. Previously these were transformed verbatim into blang's `and` and `or` expressions but this changed their semantics. Opam's boolean ops only act like undefined values if their value would depend on an undefined value (e.g. `undefined | true` is true but `undefined | false` is undefined). Now `a | b` in opam translates to `(or (and (all_pkg_vars_defined a) a) (and (all_pkg_vars_defined b) b) a b)`. This first checks whether a the expressions `a` and `b` contain no undefined variables before depending on their values. The repeated conditions `a` and `b` at the end are so that if all the arguments were false or undefined, and either `a` or `b` are undefined, the normal undefined variable behaviour is preserved. There's an analagous representation for `a & b`.

src/dune_lang/action.ml
Alizter: What confused me reading this is that `slang` is supposed to be about encoding lists of strings but here you are only decoding singletons?

gridbugs: Slang expressions evaluate to lists of strings but I wouldn't say slang is supposed to be about encoding lists of strings. E.g. there's no syntax for lists of strings in slang. When you write `(run echo hello world)` "echo", "hello", and "world" are 3 different slang expressions.

Do you think I should write a comment to clarify this here?

Alizter: @gridbugs That makes more sense. I think a comment in slang.mli would help clear this up, thanks!

gridbugs:
More changes:
- I've updated the language to add two additional operators `and_absorb_undefined_var` and `or_absorb_undefined_var` which have the same behaviour as opam's boolean logical operators which greatly simplifies the logic for translating opam to slang and shortens the generated code. 
- I've removed `all_pkg_vars_defined` and added `has_undefined_var` which means the opposite and I think is a better name. 
- Added `catch_undefined_var` which allows providing a fallback value to use when an undefined variable is encountered. This simplifies the generated code for the `%{var?foo:bar}%` syntax.

src/dune_lang/blang.ml
```diff
@@ -47,68 +47,74 @@ module Op = struct
   ;;
 end
 
-type t =
+type 'string ast =
```
Alizter, line 50: Why do you need this type? Can't you just inline it into `Ast.t`?

Alizter: In the mli you can move the Ast module type up with the same order as the ml file which would let you write `type t = String_with_vars.t Ast.t`

gridbugs: It's done this way so that the constructors for `Blang.t` are defined inside the `Blang` module (so you can write `Blang.And` for example). I'll add a comment explaining this.

gridbugs: Added a comment. We could also introduce functions to use instead of constructors and move the constructors to `Blang.Ast`. Let me know what you think.

Alizter: I see why you did it now.

gridbugs:
@rgrinberg is there any more work required on this PR before you think it's ready to merge?

PR #8598, opened 2023-09-06
ci: fix bench tests

jchavarri:
The benchmark tests for melange and other e2e measurements broke in https://github.com/ocaml/dune/pull/8408 as the output of the watch mode (which the tests were grepping) changed.

This PR should fix it.


jchavarri:
Works: https://github.com/jchavarri/dune/actions/runs/6100581129

PR #8601, opened 2023-09-06
feature(monitor): add Jobs tab in tui mode

Alizter:
When `dune monitor` is run with `--display tui`, we now get an extra tab that can be clicked. This displays the currently running jobs and recently finished ones.

- [ ] changelog
- [x] Some documentation about it in the help for `dune monitor`.
- [ ] multiline statuses #8619


bin/monitor.ml
```diff
@@ -57,14 +57,86 @@ module State : sig
   (** Given a state update, render the update. *)
   val render : t -> Update.t -> unit
 end = struct
+  module Job_state = struct
```
rgrinberg, line 60: I don't know how feasible it is, but it would help to keep all the code that I gave you in its own module. It will be easier to sync with upstream that way.

rgrinberg: Actually, this entire module can live in its own file. I don't see what's the issue with moving it.

bin/monitor.ml
```diff
@@ -163,21 +232,164 @@ end = struct
     | Diagnostics diagnostics -> Update.diagnostics state diagnostics
   ;;
 
+  module Jobs_tab : sig
+    val draw : t -> min_duration_sec:float -> min_display_duration_sec:float -> unit
+  end = struct
+    open Dune_tui.Import
+
+    let render (state : Job_state.t) ~min_duration_sec ~min_display_duration_sec =
```
rgrinberg, line 240: It should be easy enough to keep this outside this module with a little functor over the rendering calls as well.

Alizter:
Marking this as draft for now, as I still need to improve some things.
- [ ] Make the monitor tab prettier.
- [ ] In the simple display jobs monitor, use the jst implementation as a console status.
  - #8619 will allow for this.

bin/monitor/jobs_display_intf.ml
```diff
@@ -0,0 +1,11 @@
+module type S = sig
```
rgrinberg, line 1: Is it possible to get rid of this module?

src/dune_tui/widgets/tabs.mli
```diff
@@ -0,0 +1,16 @@
+open Import
+
+module Tab : sig
+  type t =
+    { title : string
+    ; ui : unit -> Ui.t Lwd.t
+    }
+end
+
+type t =
```
rgrinberg, line 10: How about the names `Tab` and `Tab.Bar` instead?

bin/monitor/rpc_running_jobs.mli
```diff
@@ -0,0 +1,40 @@
...
+
+  type t = private
+    { mutable running_jobs_by_id : Dune_rpc.Job.t Job_map.t
+    ; mutable displayed_jobs_by_id : Displayed_job.t Job_map.t
+    ; mutable keep_refreshing : bool
+    ; mutable free_rows : Int.Set.t
+    }
+end
+
+module Display (_ : Jobs_display_intf.S with type state = State.t) : sig
```
rgrinberg, line 33: You can easily get rid of this functor and just have a function that takes a record.

bin/monitor/tui_jobs_display.ml
```diff
@@ -0,0 +1,142 @@
...
+
+let render (state : Rpc_running_jobs.State.t) =
+  let now = Unix.gettimeofday () in
+  let jobs =
+    Rpc_running_jobs.Job_map.to_list_map
+      state.displayed_jobs_by_id
+      ~f:(fun (_ : Job.Id.t) displayed_job -> displayed_job.row, displayed_job)
+    |> Int.Map.of_list_exn
+  in
+  (* CR-someday amokhov: We should highlight jobs that were running for longer than 5s
```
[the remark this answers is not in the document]
Alizter: Yes but they are relevant.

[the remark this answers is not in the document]
Alizter: They are duplicated because we have the same issue with both jobs display backends. The coloring issues will have to be solved independently however the description shortening and anonymous action filtering could share a solution.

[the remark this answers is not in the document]
rgrinberg: Why do coloring issues have to be solved independently? We can use `Pp.t` to set the colors in both.

Alizter:
FTR, this is waiting on #8619.

PR #8602, opened 2023-09-06
test(pkg): installing OCaml via the lock dir

rgrinberg:

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/ocaml-compiler.t
+++ b/test/blackbox-tests/test-cases/pkg/ocaml-compiler.t
@@ -1,0 +1,110 @@
+Setting the compiler in the lock directory
+
+We need some data for ocamlc -config
+
+  $ mkdir stdlib && touch stdlib/Makefile.config
+  $ . ./helpers.sh
+  $ cat >ocaml.config <<EOF
+  > version: 4.14.1
+  > standard_library_default: $PWD/stdlib
+  > standard_library: $PWD/stdlib
+  > ccomp_type: cc
+  > c_compiler: clang
+  > ocamlc_cflags: -O2 -fno-strict-aliasing -fwrapv -pthread
+  > ocamlc_cppflags: -D_FILE_OFFSET_BITS=64
+  > ocamlopt_cflags: -O2 -fno-strict-aliasing -fwrapv -pthread
+  > ocamlopt_cppflags: -D_FILE_OFFSET_BITS=64
+  > bytecomp_c_compiler: clang -O2 -fno-strict-aliasing -fwrapv -pthread   -D_FILE_OFFSET_BITS=64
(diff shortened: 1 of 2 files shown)


test/blackbox-tests/test-cases/pkg/ocaml-compiler.t
```diff
@@ -0,0 +1,110 @@
...
+  > cmx_magic_number: Caml1999Y031
+  > cmxa_magic_number: Caml1999Z031
+  > ast_impl_magic_number: Caml1999M031
+  > ast_intf_magic_number: Caml1999N031
+  > cmxs_magic_number: Caml1999D031
+  > cmt_magic_number: Caml1999T031
+  > linear_magic_number: Caml1999L031
+  > EOF
+
+This is the directory for our take OCaml the test shouldn't observe. We can't
```
Leonidas-from-XIV, line 62:
```suggestion
This is the directory for our fake OCaml the test shouldn't observe. We can't
```

test/blackbox-tests/test-cases/pkg/ocaml-compiler.t
```diff
@@ -0,0 +1,110 @@
+Setting the compiler in the lock directory
+
+We need some data for ocamlc -config
+
+  $ mkdir stdlib && touch stdlib/Makefile.config
+  $ . ./helpers.sh
+  $ cat >ocaml.config <<EOF
```
Leonidas-from-XIV, line 7: This could be inlined into `mycamlsources/ocamlc`, I assume?

test/blackbox-tests/test-cases/pkg/ocaml-compiler.t
```diff
@@ -0,0 +1,110 @@
+Setting the compiler in the lock directory
+
+We need some data for ocamlc -config
+
+  $ mkdir stdlib && touch stdlib/Makefile.config
```
Leonidas-from-XIV, line 5: Why do we need this file/folder?

Leonidas-from-XIV (approved):
I have some comments on the why, but the test looks sensible.

test/blackbox-tests/test-cases/pkg/ocaml-compiler.t
rgrinberg: The `Ocaml_config` module requires it to be there. I don't know if that requirement is justified, but it is present for non OCaml installations.

test/blackbox-tests/test-cases/pkg/ocaml-compiler.t
rgrinberg: It could be, but it needs to be used in two different places:

* The fake OCaml used to shadow the system OCaml for reproducibility
* The fake OCaml installed from the lock dir

So if it's a file, we can reuse it in both instances.

PR #8603, opened 2023-09-07
fix(init): parse --public as a public name

### LINKED ISSUE #7108: Dune init's `--public` doesn't allow dashes

emillon:
Fixes #7108

### DIFF
--- a/bin/dune_init.ml
+++ b/bin/dune_init.ml
@@ -195,6 +195,37 @@ module Init_context = struct
   ;;
 end
 
+module Public_name = struct
+  include Lib_name
+  module Pkg = Dune_lang.Package_name.Opam_compatible
+
+  let is_opam_compatible l =
+    Lib_name.package_name l |> Dune_lang.Package_name.is_opam_compatible
+  ;;
+
+  let of_string_user_error (loc, s) =
+    let open Result.O in
+    let* l = of_string_user_error (loc, s) in
+    if is_opam_compatible l
+    then Ok l
+    else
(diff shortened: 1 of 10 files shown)


bin/dune_init.ml
```diff
@@ -195,6 +195,64 @@ module Init_context = struct
   ;;
 end
 
+module Public_name = struct
```
rgrinberg, line 198: Is this not the same as `Lib_name`?

emillon: not quite since it's using `Opam_compatible` and there's no `Private` version. but I'd like to make that a bit closer. maybe I can use the Lib_name parser and validate after the fact.

emillon: I did that, I think it looks better indeed.

rgrinberg (approved):

PR #8607, opened 2023-09-08
chore: add missing unix dependencies

emillon:
This suppresses warnings with ocaml >= 5.

### DIFF
--- a/src/dune_findlib/dune
+++ b/src/dune_findlib/dune
@@ -1,3 +1,3 @@
 (library
  (name dune_findlib)
- (libraries stdune dune_lang dune_engine memo ocaml dune_meta_parser))
+ (libraries stdune dune_lang dune_engine memo ocaml dune_meta_parser unix))
--- a/otherlibs/dune-rpc/private/dune
+++ b/otherlibs/dune-rpc/private/dune
@@ -1,7 +1,7 @@
 (library
  (name dune_rpc_private)
  (public_name dune-rpc.private)
- (libraries stdune ordering pp csexp dyn xdg)
+ (libraries stdune ordering pp csexp dyn xdg unix)
  (synopsis "for internal use only")
  (instrumentation
   (backend bisect_ppx)))
--- a/src/dune_digest/dune
+++ b/src/dune_digest/dune
@@ -1,6 +1,6 @@
 (library
  (name dune_digest)
- (libraries dune_metrics stdune)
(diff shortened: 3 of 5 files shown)


emillon:
2 notes:
- there's one to fix in our vendored fork of opam
- I don't understand why the one in `dune_findlib` is necessary, since `Unix` is not referenced in that library

rgrinberg:
@nojb or @dra27 why does the warning fire in this case?

In any case, I would rather disable the warning where it's not applicable rather than adding a dependency on unix.

emillon:
for that particular one we can investigate - the other ones are legitimate (the libraries use Unix).
Something that's possible is that it's triggering at a distance from another library that uses unix in an undeclared way but this is only reported when linking a dependent library.

src/dune_findlib/dune
```diff
@@ -1,3 +1,3 @@
 (library
  (name dune_findlib)
- (libraries stdune dune_lang dune_engine memo ocaml dune_meta_parser))
+ (libraries stdune dune_lang dune_engine memo ocaml dune_meta_parser unix))
```
rgrinberg, line 3: As discussed, we should disable the warning for this library because it doesn't use Unix.

emillon: done

nojb:
> Something that's possible is that it's triggering at a distance from another library that uses unix in an undeclared way but this is only reported when linking a dependent library.

I think this is the explanation: one of the dependencies must use Unix without declaring it.

emillon:
I got to the bottom of this. In this line:

https://github.com/ocaml/dune/blob/faa389a09b20d81f6f339ed1ef20d3fd4349c9bd/src/dune_findlib/config.ml#L46

the `_` has type `Dune_engine.Import.Unix_error.Detailed.t`, which expands to `Unix.error * string * string`. So while the source code does not refer to the `Unix` module, it's still "depending" on that module.

Alizter:
@emillon Does that still happen if you make it `Result.get_ok`?

emillon:
using `Result.to_option` does not work. I don't think that `_` is the problem, but rather that `Unix` appears in the type of a function we use.

rgrinberg (approved):
LGTM. I would suggest  to add a type annotation to this `_` so that it's obvious

emillon:
I added an annotation. re exporting unix from dune_engine would also work.

PR #8608, opened 2023-09-08
test(dune_console): add escaping for clearer tests

Alizter:
We split the output lines and then escape the remaining control characters.

We also make sure that the test outputs are all `{| |}` strings.

We've also simplified some of the tests so that they are not doing "2 things at once". Namely overwriting the status line and also clearing. Now the tests stick to one job.

### DIFF
--- a/test/expect-tests/dune_console/dune_console_tests.ml
+++ b/test/expect-tests/dune_console/dune_console_tests.ml
@@ -1,14 +1,17 @@
-(** Creation of Dune_console is stateful so we introduce a new module for each test. *)
+open Stdune
+
+let escape str =
+  str |> String.split_lines |> List.map ~f:String.escaped |> List.iter ~f:print_endline
+;;
+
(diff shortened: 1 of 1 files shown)


test/expect-tests/dune_console/dune_console_tests.ml
```diff
@@ -1,14 +1,17 @@
-(** Creation of Dune_console is stateful so we introduce a new module for each test. *)
+open Stdune
+
+let escape str =
+  str |> String.split_lines |> List.map ~f:String.escaped |> List.iter ~f:prerr_endline
```
rgrinberg, line 4: Any specific reason why we're printing to stderr?

Alizter: No. Shall I switch to stdout?

rgrinberg: Yes

PR #8618, opened 2023-09-09
feature: add KiB,MiB,GiB and TiB values

Alizter:
We make these the default printed values when displaying bytes. However the dune lang decoder can understand binary and decimal byte units.

We also expand the test suite to account for parsing and displaying these values.

- [x] changelog
- [x] update documentation in cmds
```
OPTIONS
      --size=BYTES
             Size to trim the cache to. BYTES is the number of bytes followed
             by a unit. Byte units can be one of B, kB, KiB, MB, MiB, GB, GiB,
             TB or TiB.  

       --trimmed-size=BYTES
           Size to trim from the cache. BYTES is the same as for --size.
```
- [x] <s>allow `123` to be parsed as `123B` or remove test.</s> not so easy to do
- [x] <s>Should we default to printing binary units or decimal units?</s> printing only decimal like before now

cc @snowleopard

### DIFF
--- a/bin/cache.ml
+++ b/bin/cache.ml
@@ -25,12 +25,25 @@ let trim =
        Arg.(
          value
          & opt (some bytes) None
-         & info ~docv:"BYTES" [ "trimmed-size" ] ~doc:"Size to trim from the cache.")
+         & info
+             ~docv:"BYTES"
+             [ "trimmed-size" ]
+             ~doc:"Size to trim from the cache. $(docv) is the same as for --size.")
      and+ size =
        Arg.(
          value
          & opt (some bytes) None
-         & info ~docv:"BYTES" [ "size" ] ~doc:"Size to trim the cache to.")
+         & info
+             ~docv:"BYTES"
+             [ "size" ]
+             ~doc:
+               (sprintf
+                  "Size to trim the cache to. $(docv) is the number of bytes followed by \
+                   a unit. Byte units can be one of %s."
+                  (String.enumerate_or
+                     (List.map
+                        ~f:(fun (units, _) -> List.hd units)
+                        Bytes_unit.conversion_table))))
      in
      Log.init_disabled ();
      let open Result.O in
--- a/otherlibs/stdune/src/bytes_unit.mli
+++ b/otherlibs/stdune/src/bytes_unit.mli
@@ -1,7 +1,8 @@
-(** Conversion table for byte suffixes and their corresponding [Int64.t] values.
-    The first element of the tuple is a list of possible suffixes for the second
-    element of the tuple which is the value. There are some static checks done
-    on this table ensuring it is ordered and well-formed.*)
+(** Conversion table for decimal byte suffixes and their corresponding [Int64.t] values.
+    The first element of the tuple is a list of possible suffixes for the second element
+    of the tuple which is the value. There are some static checks done on this table
+    ensuring it is ordered and well-formed.*)
 val conversion_table : (string list * Int64.t) list
 
+(** [pp n] pretty-prints [n] as a decimal byte suffix. *)
 val pp : Int64.t -> string
--- a/otherlibs/stdune/test/bytes_unit_tests.ml
+++ b/otherlibs/stdune/test/bytes_unit_tests.ml
@@ -15,8 +15,10 @@ let () =
   loop Bytes_unit.conversion_table
 ;;
 
-let%expect_test _ =
-  let bytes =
+let test bytes = List.iter ~f:(fun x -> Bytes_unit.pp x |> print_endline) bytes
+
+let%expect_test "Testing significant digit boundaries" =
+  test
     [ 0L
     ; 1L
     ; 12L
@@ -31,9 +33,10 @@ let%expect_test _ =
     ; 12345678901L
     ; 123456789012L
     ; 1234567890123L
-    ]
-  in
-  List.iter ~f:(fun x -> Bytes_unit.pp x |> print_endline) bytes;
+    ; 12345678901234L
+    ; 123456789012345L
+    ; 1234567890123456L
+    ];
   [%expect
     {|
       0B
@@ -49,5 +52,18 @@ let%expect_test _ =
       1.23GB
       12.35GB
       123.46GB
-      1.23TB |}]
+      1.23TB
+      12.35TB
+      123.46TB
+      1234.57TB |}]
+;;
+
+(* CR-someday alizter: we should raise an exception here rather than giving a meaningless value. *)
+(* Negative units get truncated but still printed as a negative. *)
+let%expect_test "Negative units" =
+  test [ -1L; -10L ];
+  [%expect {|
+    -0.00TB
+    -0.00TB
+    |}]
 ;;
--- a/test/blackbox-tests/test-cases/dune-cache/cache-man.t
+++ b/test/blackbox-tests/test-cases/dune-cache/cache-man.t
@@ -91,10 +91,12 @@ Testing the output of dune cache trim.
   
   OPTIONS
          --size=BYTES
-             Size to trim the cache to.
+             Size to trim the cache to. BYTES is the number of bytes followed
+             by a unit. Byte units can be one of B, kB, KiB, MB, MiB, GB, GiB,
+             TB or TiB.
   
          --trimmed-size=BYTES
-             Size to trim from the cache.
+             Size to trim from the cache. BYTES is the same as for --size.
   
   COMMON OPTIONS
          --help[=FMT] (default=auto)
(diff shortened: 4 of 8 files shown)


emillon:
I'm out of the loop on this change but I don't think we should change the default (controversial change) in the same change we're adding new accepted prefix (less controversial change).

test/expect-tests/dune_sexp/dune_sexp_tests.ml
```diff
@@ -0,0 +1,80 @@
...
+;;
+
+let%expect_test "parsing GB suffix" =
+  test_bytes "1GB";
+  [%expect {| 1_000_000_000 |}]
+;;
+
+let%expect_test "parsing GiB suffix" =
+  test_bytes_hex "1GiB";
+  [%expect {| 0x40_000_000 |}]
```
snowleopard, line 69: Same here: maybe output both hexadecimal and decimal numbers, so that I can at least quickly convince myself that the number of digits in the decimal number is as expected.

Alizter: I've replaced this with an assertion to a long power.

bin/cache.ml
```diff
@@ -21,16 +21,27 @@ let trim =
     Cmd.info "trim" ~doc ~man
   in
   Cmd.v info
-  @@ let+ trimmed_size =
+  @@ let+ size =
```
Alizter, line 24: I've swapped these round since size comes before --trimed size in the doc.

snowleopard: It's always best to move such changes unrelated to the main goal of the PR to separate PRs. Let the reviewers focus on what really matters by minimising the noise.

Alizter: @snowleopard I'll revert it for now since its not important.

bin/cache.ml
```diff
@@ -21,33 +21,44 @@ let trim =
...
-           Result.Ok (Int64.sub (Dune_cache.Trimmer.overhead_size ()) size)
-         | _ -> Result.Error "please specify either --size or --trimmed-size"
-       in
-       Dune_cache.Trimmer.trim ~goal
-     with
-     | Error s -> User_error.raise [ Pp.text s ]
-     | Ok { trimmed_bytes } ->
-       User_message.print
-         (User_message.make [ Pp.textf "Freed %s" (Bytes_unit.pp trimmed_bytes) ])
+  @@
```
Alizter, line 24: @snowleopard It seems there is noise anyway due to ocamlformat. Just know that I've only touched the documentation for `size` and `trimmed_size` in the following block.

emillon: If you put `@@` and `let+` on the same line and reformat, it will use that style and that will decrease the diff.

Alizter: Thanks! I did not know that.

otherlibs/stdune/test/bytes_unit_tests.ml
```diff
@@ -31,23 +33,37 @@ let%expect_test _ =
...
-      12.35kB
-      123.46kB
-      1.23MB
-      12.35MB
-      123.46MB
-      1.23GB
-      12.35GB
-      123.46GB
-      1.23TB |}]
+0 B
```
snowleopard, line 42: Why is this feature changing serialisation (adding a space)?

test/blackbox-tests/test-cases/dune-cache/cache-man.t
```diff
@@ -91,10 +91,13 @@ Testing the output of dune cache trim.
   
   OPTIONS
          --size=BYTES
-             Size to trim the cache to.
+             Size to trim the cache to. BYTES is the number of bytes followed
+             by a unit. Byte units can be one of B, bytes, kB, KB, kilobytes,
+             KiB, KiB, kibibytes, MB, megabytes, MiB, mebibytes, GB, gigabytes,
+             GiB, gibibytes, TB, terabytes, TiB or tebibytes.
```
snowleopard, line 97: To be honest, this help message seems pretty excessive. Noone is going to run the trimmer with `--size=1megabytes` (which, by the way, isn't even correct and we don't have a singular form).

otherlibs/stdune/test/bytes_unit_tests.ml
Alizter: I added it to be more legible.

test/blackbox-tests/test-cases/dune-cache/cache-man.t
Alizter: I agree it is excessive. Where is a better place to document what we accept? Or are you suggesting we remove the extra quantifiers?

Alizter:
My main motivation for this change was to have nice download progress indicators in package management. Typically downloads in package managers that I've seen show the units in `MiB`. This is the reason I bothered with this entire PR.

From the point of view of trimming the cache, this change is kind of useless. So maybe its best to close this for now and resurrect if its actually needed.

otherlibs/stdune/test/bytes_unit_tests.ml
snowleopard: Well, still, this PR keeps mixing different changes together which makes it hard to review and slows the review down. Why not stick to the point, which is adding new units, as the title says?

snowleopard:
I think adding units is clearly useful, and if that's the only change that this PR was attempting, it would have landed long ago. (Yet we keep being distracted by other things.)

otherlibs/stdune/test/bytes_unit_tests.ml
Alizter: You're right. I've split that change off now. Sorry about that.

test/blackbox-tests/test-cases/dune-cache/cache-man.t
Alizter: I've shortened it to just the main units now. I've updated the PR description so it is more clear too.

Alizter:
@snowleopard Sorry for the back and forth, it was not my intention to make this difficult to review. I've removed the unrelated parts and the PR now only consists of adding the new units and adding tests testing all of them.

otherlibs/stdune/src/bytes_unit.mli
```diff
@@ -1,7 +1,8 @@
-(** Conversion table for byte suffixes and their corresponding [Int64.t] values.
-    The first element of the tuple is a list of possible suffixes for the second
-    element of the tuple which is the value. There are some static checks done
-    on this table ensuring it is ordered and well-formed.*)
+(** Conversion table for decimal byte suffixes and their corresponding [Int64.t] values.
```
snowleopard, line 1:
```suggestion
(** Conversion table for decimal byte units and their corresponding [Int64.t] values.
```

otherlibs/stdune/test/bytes_unit_tests.ml
```diff
@@ -49,5 +52,17 @@ let%expect_test _ =
...
       123.46GB
-      1.23TB |}]
+      1.23TB
+      12.35TB
+      123.46TB
+      1234.57TB |}]
+;;
+
+(* Negative units get truncated but still printed as a negative. *)
+let%expect_test "Negative units" =
```
snowleopard, line 62: Thanks for adding this test but I find the behaviour surprising. I think it would be better to just raise `Invalid_argument` or something like this on negative number of bytes.

snowleopard: (So, maybe let's add a TODO to fix the current behaviour in future?)

Alizter: Would it be ok for me to do that in a follow up PR rather than this one?

snowleopard (approved):
Thanks, looks good to me now!

otherlibs/stdune/test/bytes_unit_tests.ml
snowleopard: I think adding a TODO is better in this PR (because it adds a test whose behaviour seems surprising). Addressing the TODO should better be done separately.

Alizter: Sorry, my comment was for your first comment not you second. I've added a CR now.

Alizter:
@rgrinberg This is ready to merge now.

otherlibs/stdune/test/bytes_unit_tests.ml
snowleopard: Ah, got it!

test/expect-tests/dune_sexp/dune_sexp_tests.ml
```diff
@@ -0,0 +1,92 @@
...
+;;
+
+let test_bytes ?check value = parse_and_assert ?check value |> Printf.printf "%#Ld\n"
+
+(* Hack to insert underscores for hex values. Digits must only be 0-9 *)
+let test_bytes_hex ?check value =
+  parse_and_assert ?check value
+  |> sprintf "%Lx"
+  |> Int.of_string
+  |> function
```
rgrinberg, line 30: Use a `match` instead

Alizter: Done.

PR #8619, opened 2023-09-09
fix: allow multiline status lines

Alizter:
We make status lines work on multiple lines which fixes a bug where wrapped status lines wouldn't clear properly.

- [x] changelog
- [x] finish windows stubs
- [x] regulate console size inside expect test
- [ ] truncation should react to terminal resizing (not really sure how to do this)
  - See the issue here https://github.com/ocaml/dune/pull/8619#issuecomment-1757540531
- [x] fix broken colors


src/dune_console/progress.ml
```diff
@@ -3,25 +3,44 @@ open Stdune
...
+      let buf = Buffer.create ((clear_code_length * !status_line_height) - 1) in
+      (* Return the cursor to the begining of the line. *)
+      Buffer.add_char buf '\r';
+      (* First we clear the line. *)
+      Buffer.add_string buf ansi_erase;
+      for _ = 1 to !status_line_height - 1 do
+        (* If we have multiple lines, we move the cursor up and clear. *)
+        Buffer.add_string buf (ansi_up ^ ansi_erase)
+      done;
+      Buffer.contents buf |> Printf.eprintf "%s")
```
rgrinberg, line 28: Why do you add this extra layer of buffering? You can just use print directly and it's already buffered.

test/expect-tests/dune_console/dune_console_tests.ml
```diff
@@ -1,7 +1,35 @@
...
+     the string. *)
+  let clear_count =
+    let lines_split_by_clear =
+      Re.split (Re.compile (Re.str "\027[2K")) (" " ^ str ^ " ")
+    in
+    List.length lines_split_by_clear - 1
+  in
+  Printf.printf "Clear count: %d\n" clear_count;
+  let carriage_return_count =
+    let lines_split_by_cr = Re.split (Re.compile (Re.char '\r')) (" " ^ str ^ " ") in
```
rgrinberg, line 29: Is this just `String.split_on_char`?

src/dune_console/progress.ml
```diff
@@ -3,25 +3,44 @@ open Stdune
...
   ;;
 
-  let show_status_line () = if !status_line_len > 0 then Ansi_color.prerr !status_line
+  let show_status_line () = if !status_line_height > 0 then Ansi_color.prerr !status_line
 
   let set_status_line = function
     | None ->
       hide_status_line ();
```
rgrinberg, line 35: You didn't introduce this issue, but this should be followed by a flush

test/expect-tests/dune_console/dune_console_tests.ml
```diff
@@ -1,7 +1,35 @@
...
+let escape_and_count str =
+  let lines = escape_and_split_lines str in
+  List.iter lines ~f:print_endline;
+  print_endline "-------";
+  Printf.printf "Line count: %d\n" (List.length lines);
+  (* We pad the string with spaces to ensure that we don't miss any clears at the end of
+     the string. *)
+  let clear_count =
+    let lines_split_by_clear =
+      Re.split (Re.compile (Re.str "\027[2K")) (" " ^ str ^ " ")
```
rgrinberg, line 23: You can compile the re only once outside the test.

src/dune_console/progress.ml
Alizter: Are you sure? This is the no-flush module.

src/dune_console/progress.ml
```diff
@@ -3,25 +3,37 @@ open Stdune
...
-      status_line_len := 0
+      status_line_height := 0
     | Some line ->
       let line = Pp.map_tags line ~f:User_message.Print_config.default in
-      let line_len = String.length (Format.asprintf "%a" Pp.to_fmt line) in
       hide_status_line ();
       status_line := line;
-      status_line_len := line_len;
+      status_line_height
+        := Format.asprintf "%a" Pp.to_fmt line |> String.split_lines |> List.length;
```
Alizter, line 36: This is a guess and as such breaks when the terminal width causes the status line to wrap. I need to work out a better way to calculate how high the status line will be whilst still taking into account the width of the terminal.

src/dune_console/progress.ml
```diff
@@ -3,25 +3,53 @@ open Stdune
...
       status_line := line;
-      status_line_len := line_len;
+      (status_line_height
+         := Format.asprintf "%a" Pp.to_fmt line
+            |> String.split_lines
+            |>
+            (* Notty_unix gives us a cross-platform way to get the terminal width. *)
+            match Notty_unix.winsize Unix.stderr with
+            | Some (width, _) ->
+              List.fold_left ~init:0 ~f:(fun line_count next_line ->
```
rgrinberg, line 42: All this logic looks like we're creating a problem for ourselves and then working hard to solve it. Currently, this is doing the following:

1. Pretty print the status line while wrapping long lines at a certain margin (say 80 chars)
2. Inspect the terminal settings to see if we are overflowing
3. Count the number of lines we've overflowed by.

What about just not allowing overflows in the first place.

First, if we are going to allow multiple entries in the status, let's just make that explicit by introducing:

```ocaml
val set_status : User_message.Style.t Pp.t list -> unit
```

to replace the existing `set_status_line`.

Then, whenever an entry in this status doesn't fit in one line, we just truncate it and add `...` to show it's being truncated. If the user's display is small none of this wrapping is going to look good anyway. It also makes no sense to wrap logical entries like the progress counter.

Alizter:
@rgrinberg
- Status messages that are too long are now truncated and "..." added.
- Status_line has now become status

src/dune_config/config.ml
```diff
@@ -168,3 +168,18 @@ let threaded_console =
   register t;
   t
 ;;
+
+let console_width_fallback =
```
Alizter, line 172: This is a fallback value for the console width, mainly used during testing.

rgrinberg:
> truncation should react to terminal resizing (not really sure how to do this)

When the terminal is resized, the appropriate signal is sent and re-rendering should occur. Is that not working at the moment?

Alizter:
@rgrinberg I think the issue is that we calculate the clearing based on the recent terminal size rather than the one from when we rendered the display. However I'm not fully satisfied with this reason.

To reproduce the issue easily, build the dune in this PR and run something like `dune monitor`. That will display a status line which you can use to demonstrate the issue with resizing the window.

Here is what I observe:

[Screencast from 2023-10-11 12-58-59.webm](https://github.com/ocaml/dune/assets/8614547/653a2423-629f-4ee5-8865-24af1ecb2963)

Alizter:
Actually, this appears to only be an issue when the terminal emulator wraps lines. I've tried this in another terminal emulator that doesn't wrap lines and it works correctly.

I think I have a better idea of what is going on. We calculate the truncated length based on a terminal width that we query. When we resize the width shrinks causing the line to momentarily wrap and causes this issue. I am not really sure what we can do about this issue.

src/dune_console/dune
```diff
@@ -1,5 +1,5 @@
 (library
  (name dune_console)
- (libraries stdune)
+ (libraries stdune dune_notty_unix dune_config)
```
rgrinberg, line 3: We can't depend on `dune_notty_unix` in this library unfortunately. I suggest that you move the console size computation to stdune (or to its own library) and then access it from both dune_notty_unix and dune_console.

rgrinberg (changes requested):

otherlibs/stdune/src/pp0.mli
```diff
@@ -0,0 +1,11 @@
...
+
+(** This version of [Pp.compare] uses [Ordering.t] rather than returning an [int]. *)
+val compare : compare:('a -> 'a -> Ordering.t) -> 'a t -> 'a t -> Ordering.t
+
+val to_dyn : ('a -> Dyn.t) -> 'a t -> Dyn.t
+
+(** [truncate length pp] truncates [pp] to at most [length] characters. If the [pp]
+    would go over the [length] then an ellipsis is added at the end (still within
+    length). *)
+val truncate : int -> 'a t -> 'a t
```
rgrinberg, line 11: `t` comes first.

otherlibs/stdune/src/pp0.mli
```diff
@@ -0,0 +1,11 @@
+include module type of Pp with type 'a t = 'a Pp.t
+
+(** This version of [Pp.compare] uses [Ordering.t] rather than returning an [int]. *)
+val compare : compare:('a -> 'a -> Ordering.t) -> 'a t -> 'a t -> Ordering.t
```
rgrinberg, line 4: We don't use this label consistently, so you can just drop it.

rgrinberg:
I added a test where your truncation outputs more than the hard limit.

> We could go back to having the complicated logic that checks if a line would wrap and calculate how to clear the wraped line correctly. But that seems like overkill.

I think that's probably the right way to go.

otherlibs/stdune/test/pp_tests.ml
```diff
@@ -0,0 +1,188 @@
+open Stdune
+
+let test_truncate length pp =
+  let pp = Pp.truncate length pp in
+  Format.printf "%s" (Pp.to_dyn Dyn.opaque pp |> Dyn.to_string)
+;;
+
+let test_truncate_and_print length pp =
+  let pp = Pp.truncate length pp in
+  Format.printf "%a" Pp.to_fmt pp
```
rgrinberg, line 10: Can you write an assert that makes sure we don't print:

1. More than one line
2. We don't go over the length hard limit

src/dune_tui/dune_tui.ml
```diff
@@ -230,10 +230,10 @@ let status_bar =
     in
     Button.of_ (Ui.atom image) toggle
   and+ status =
-    Lwd.get status_line
+    Lwd.get status
     >>| function
-    | None -> I.empty
-    | Some message -> Drawing.pp_to_image message
+    | [] -> I.empty
+    | messages -> I.hcat (List.map ~f:Drawing.pp_to_image messages)
```
rgrinberg, line 236: Shouldn't there be some truncation here as well?

test/expect-tests/dune_console/dune_console_tests.ml
```diff
@@ -1,7 +1,36 @@
...
+   - Number of line clears "\027[2K".
+   - Number of carriage returns.
+*)
+let escape_and_count =
+  let clear_re = Re.compile (Re.str "\027[2K") in
+  fun str ->
+    let lines = escape_and_split_lines str in
+    List.iter lines ~f:print_endline;
+    print_endline "-------";
+    Printf.printf "Line count: %d\n" (List.length lines);
```
rgrinberg, line 27: `printfn`

test/expect-tests/dune_console/dune_console_tests.ml
```diff
@@ -1,7 +1,36 @@
...
+    ((* We pad the string with spaces to ensure that we don't miss any clears at the end
+        of the string. *)
+     let lines_split_by_clear = Re.split clear_re (" " ^ str ^ " ") in
+     List.length lines_split_by_clear - 1 |> Printf.printf "Clear count: %d\n");
+    let lines_split_by_cr = String.split_on_char ~sep:'\r' str in
+    List.length lines_split_by_cr - 1 |> Printf.printf "Carriage return count: %d\n"
```
rgrinberg, line 33: Could you write an assertion based on these numbers. E.g. `assert (clear_count = carriage_return_count)` or something.

test/expect-tests/dune_console/dune_console_tests.ml
```diff
@@ -1,7 +1,36 @@
...
+let escape_and_count =
+  let clear_re = Re.compile (Re.str "\027[2K") in
+  fun str ->
+    let lines = escape_and_split_lines str in
+    List.iter lines ~f:print_endline;
+    print_endline "-------";
+    Printf.printf "Line count: %d\n" (List.length lines);
+    ((* We pad the string with spaces to ensure that we don't miss any clears at the end
+        of the string. *)
+     let lines_split_by_clear = Re.split clear_re (" " ^ str ^ " ") in
```
rgrinberg, line 30: Weird that you need to do this. Can't you just `Re.full_split` and count the delimiters?

src/dune_config/config.mli
```diff
@@ -54,6 +54,10 @@ val threaded_console : Toggle.t t
 (** The number of frames per second for the threaded console. *)
 val threaded_console_frames_per_second : [ `Default | `Custom of int ] t
 
+(** The console width that dune will fall back on if it cannot determine the
+    console width. *)
+val console_width_fallback : int t
```
rgrinberg, line 59: It would be nice to introduce an option to disable truncation altogether.

rgrinberg, line 59: Instead of calling this option the "fallback", make it:

```
val console_width : int option t
```

And if it's set, always use it instead of asking for the terminal size.

rgrinberg:
As discussed earlier,

1. we can remove the truncation from this into a subsequent PR
2. I would suggest that we use `hbox` for every line in the status line. This might not necessarily hold in the future, as we could make this customizable. But for now, it seems like a good default.

src/dune_console/progress.ml
```diff
@@ -1,40 +1,60 @@
...
-      show_status_line ()
+  let show_status () =
+    if List.length !status > 0
+    then (
+      let width =
+        match Notty_unix.winsize Unix.stderr with
+        | Some (width, _) -> width
+        | None -> Dune_config.Config.(get console_width_fallback)
+      in
+      List.map ~f:(fun line -> Pp.truncate width line |> Pp.hbox) !status
```
Alizter, line 32: @rgrinberg We are currently hboxing every line.

Issue #8626, opened 2023-09-12
`#use_output "dune ocaml top"` fails in a project where `dune utop` works


Khady:
In ahrefs' codebase we can't use `#use_output "dune ocaml top"` in utop because it fails with an error. This is surprising as `dune utop` works flawlessly (if we don't consider the long build time)

## Expected Behavior

`#use_output "dune ocaml top"` should work in all projects where `dune utop` is working.

## Actual Behavior

```
utop # #use_output "dune ocaml top";;
File "(command-output)", line 1:
Error: Reference to undefined global `Dune_site__Dune_site_data'
^CFatal error: exception Stdlib.Sys.Break
^CFatal error: exception Stdlib.Sys.Break
```

## Reproduction

Unfortunately I don't have an open source project to reproduce the error currently.

## Specifications

- Version of `dune` (output of `dune --version`): e1f3e19cb309ffebc3a7517cd8991f4582ead896
- Version of `ocaml` (output of `ocamlc --version`): 4.14.0
- Operating system (distribution and version): debian 11

nojb:
Thanks. Relevant to the discussion in #6894

zbaylin:
This also happens (with a different error) when any library requiring `dune-build-info` is used (or if it is `#require`d itself). I imagine this is because the `Build_info_data` module [has no strict implementation](https://github.com/ocaml/dune/blob/main/otherlibs/dune-build-info/src/build_info_data.mli).

This can be easily tested by running:
```
$ utop
utop # #require "dune-build-info"
Error: Reference to undefined global `Build_info__Build_info_data'
```

Khady:
I just found out that the failure is only in utop, it works fine in `ocaml` or in a jupyter kernel.

```
$ ocaml
OCaml version 4.14.0
Enter #help;; for help.

Findlib has been successfully loaded. Additional directives:
  #require "package";;      to load a package
  #list;;                   to list the available packages
  #camlp4o;;                to load camlp4 (standard syntax)
  #camlp4r;;                to load camlp4 (revised syntax)
  #predicates "p,q,...";;   to set these predicates
  Topfind.reset();;         to force that packages will be reloaded
  #thread;;                 to enable threads

# #use_output "dune ocaml top .";;
Entering directory '/home/me/Code/ahrefs/monorepo2'
File "(command-output)", line 1:       
Error: Reference to undefined global `Dune_site__Dune_site_data'
# #show Url;;
module Url :
  sig
    type scheme = Http | Https
[...]

PR #8631, opened 2023-09-12
dune utop: do not do custom linking

### LINKED ISSUE #6894: Revamp dune utop for greater speed
After a [public](https://discuss.ocaml.org/t/dune-wish-list-for-2023/11083/49?u=grayswandyr), then private discussion with @rgrinberg , I'm submitting this issue about `dune utop`. `dune utop` is very handy for our students but it's slow (> 5s on my 2019 laptop for a project only printing "hello world") when it needs to be rebuilt. The slow part is the linking phase building an ad hoc utop.exe.

While `dune utop` is perhaps (?) not that useful to an everyday developer, this is the command our students use the most and it seems  important, in this respect, to make it fast.

@rgrinberg  suggested using `TM=$(mktemp) && dune ocaml top > "$TM" && utop -init "$TM"` to have a fast experience. If this is indeed equivalent to `dune utop` in all cases, it might be interesting to redesign `dune utop` to perform these steps rather than rebuilding utop.exe.

nojb:
In #6894 it is mentioned that `dune utop` is very slow. @emillon [mentions](https://github.com/ocaml/dune/issues/6894#issuecomment-1713353584) that this is because we link `utop` in custom mode (probably to avoid having to deal with loading shared C stubs, see [this comment](https://github.com/ocaml/dune/blob/0f555db0dc111c50088dfbedbd1c2d9d2dc6ea66/src/dune_rules/utop.ml#L7-L13)). This PR proposes switching to a "normal" bytecode link, instead of custom mode. The only complication is that we need to set `CAML_LD_LIBRARY_PATH` suitably so that shared C stubs can be found.

Another approach that was discussed in #6894 was to pass the output of `dune ocaml top` to `utop` (cc @rgrinberg). However, this has a number of limitations, see eg #8626; also, PPX does not seem to be supported by `dune ocaml top`. These limitations could/should be lifted, but this PR is meant to be a low-effort fix for the slow loading times in `utop` that can be done quickly right now.

I did some quick tests and this PR does seem to give a nice speedup, but I don't have any "large" examples to try it on. Maybe @grayswandyr can try it out and report back? Thanks.

Fixes #6894


grayswandyr:
Thanks for addressing this. I will report back. Technically, to get your modified dune, it's enough for me to pull opam-pin this commit, right?

nojb:
> Technically, to get your modified dune, it's enough for me to pull opam-pin this commit, right?

I believe so.

rgrinberg:
LGTM. I just did a minor tweak.

Do we have a test to make sure that stubs are still working?

grayswandyr:
> I did some quick tests and this PR does seem to give a nice speedup, but I don't have any "large" examples to try it on. Maybe @grayswandyr can try it out and report back? Thanks.

I also observe a good speedup (on rather small codebases). Thanks.

nojb:
> LGTM. I just did a minor tweak.

Thanks!

> Do we have a test to make sure that stubs are still working?

I added one.

rgrinberg (approved):
LGTM. Just needs a doc/changes entry.

CHANGES.md
```diff
@@ -5,6 +5,9 @@ If you're a contributor, please include your CHANGES entry in a file
 `doc/changes/$PR_NAME.md`. At release time, it will be incoporated into the
 changelog properly.
 
+- `dune utop` no longer links `utop` in "custom" mode, which should make this
```
rgrinberg, line 8: notice the comment above this message :)

nojb:
> LGTM. Just needs a doc/changes entry.

Thanks, I added a changelog entry. I didn't touch the docs otherwise since this change is an implementation detail.

CHANGES.md
nojb: Argh, will fix. Thanks.

gasche:
Thanks!

"A nice speedup" and "a good speedup" sound good, but very vague. Could you actually provide real numbers for one of your tests, so that we can get a ballpark estimate of the difference?

grayswandyr:
On my examples, I got a 3x-4x speed up but these are *few*, rather *small* examples, on a basic laptop, so I don't think it's very meaningful.  But it's very noticeable (almost instantaneous) when running lab-session exercises, while students used to complain of the several seconds it took to run their code each time it's modeified.

emillon:
This also completely removes utop tests from the list of slowest tests in dune. Huge thanks.

Khady:
We have one lib at ahrefs where the speedup is 20x. Thanks

nojb:
> r approach that was discussed in #6894 was to pass the output of `dune ocaml top` to `utop` (cc @rgrinberg). However, this has a number of limitations, see eg #8626; also, PPX does not seem to be supported by `dune ocaml top`. These limitations could/should be lifted, but this PR is meant to be a low-effort fix for the slow loadi

Thanks, good to know!

Issue #8093, opened 2023-07-03
Patch files in lockdir


gridbugs:
Opam files have a `patches` field which lists patches to apply to the source before building. The patch files are stored in the opam repository itself (not within the package source). In order to build a dune package from a lockdir without consulting the opam repository the lockdir will need to store patches for any locked packages which require them. Patch files will be stored in the directory `<package>.files` inside the lockdir (see https://github.com/ocaml/dune/issues/8095 for more on the `<package>.files` directory).

Not all the files in `<package>.files` will necessarily be patches, and opam allows patches to be conditionally applied based on filters so we'll need a field in the `<package>.pkg` metadata file which includes a list of packages and the conditions under which they should be applied.

rgrinberg:
This is implemented in the build rules (see `pkg/patch.t`) but we still need to convert .opam files to the appropriate build rules.

rgrinberg:
@gridbugs assigning the conversion step to you

Alizter:
@rgrinberg This is done now.

PR #8635, opened 2023-09-12
test(pkg): lock file generation from opam file with patch

Alizter:
We show that the generation of lock files does not take into account the patch step in an opam file.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/dune
+++ b/test/blackbox-tests/test-cases/pkg/dune
@@ -4,7 +4,7 @@
 
 (cram
  (deps %{bin:patch})
- (applies_to patch))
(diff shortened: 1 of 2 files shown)


gridbugs:
What's the context for this? Is this a feature we thought was already implemented but it's not working correctly? Linking the issue for applying opam patches in dune: https://github.com/ocaml/dune/issues/8093

rgrinberg:
> What's the context for this? Is this a feature we thought was already implemented but it's not working correctly?

It just wasn't implemented yet. Ali wanted to contribute on package management so I recommended that he'd start working on that issue.

gridbugs (approved):

PR #8646, opened 2023-09-13
fix: add original error to spurious error message

rgrinberg:
When computing digests for the cash, the algorithm was as follows:

1. Compute the digest for every target
2. If there's a failure, recompute all the digests to find the failure

The 2nd was introducing a race because it's not necessarily the case that trying to compute the digest the second time would fail.

The new algorithm does the same thing but without trying to recompute any digests. If there's an error, it will always report it.

### DIFF
--- a/src/dune_engine/cached_digest.ml
+++ b/src/dune_engine/cached_digest.ml
@@ -181,68 +181,63 @@ let set path digest =
 ;;
 
 module Digest_result = struct
-  type t =
-    | Ok of Digest.t
-    | No_such_file
-    | Broken_symlink
-    | Cyclic_symlink
-    | Unexpected_kind of File_kind.t
-    | Unix_error of Unix_error.Detailed.t
-    | Error of exn
+  module Error = struct
+    type t =
+      | No_such_file
+      | Broken_symlink
+      | Cyclic_symlink
+      | Unexpected_kind of File_kind.t
+      | Unix_error of Unix_error.Detailed.t
+      | Unrecognized of exn
 
-  let equal x y =
-    match x, y with
-    | Ok x, Ok y -> Digest.equal x y
-    | Ok _, _ | _, Ok _ -> false
-    | No_such_file, No_such_file -> true
-    | No_such_file, _ | _, No_such_file -> false
-    | Broken_symlink, Broken_symlink -> true
-    | Broken_symlink, _ | _, Broken_symlink -> false
-    | Cyclic_symlink, Cyclic_symlink -> true
-    | Cyclic_symlink, _ | _, Cyclic_symlink -> false
-    | Unexpected_kind x, Unexpected_kind y -> File_kind.equal x y
-    | Unexpected_kind _, _ | _, Unexpected_kind _ -> false
-    | Unix_error x, Unix_error y ->
-      Tuple.T3.equal Unix_error.equal String.equal String.equal x y
-    | Unix_error _, _ | _, Unix_error _ -> false
-    | Error x, Error y ->
-      (* Falling back to polymorphic equality check seems OK for this rare case.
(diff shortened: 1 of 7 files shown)


src/dune_engine/cached_digest.ml
```diff
@@ -181,68 +181,63 @@ let set path digest =
...
-      x = y
-  ;;
+  module Error = struct
+    type t =
+      | No_such_file
+      | Broken_symlink
+      | Cyclic_symlink
+      | Unexpected_kind of File_kind.t
+      | Unix_error of Unix_error.Detailed.t
+      | Error of exn
```
snowleopard, line 191: Now that it's inside the `Error` module, I'd prefer to call this variant `Unknown` or `Unrecognized`.

src/dune_shared_cache/dune_shared_cache.ml
```diff
@@ -208,23 +208,23 @@ struct
...
     match
-      Targets.Produced.Option.mapi produced_targets ~f:(fun target () ->
-        compute_digest target |> Cached_digest.Digest_result.to_option)
+      Targets.Produced.Result.mapi produced_targets ~f:(fun target () ->
+        compute_digest target)
     with
-    | Some result -> result
-    | None ->
+    | Ok result -> result
+    | Error error ->
```
snowleopard, line 215: As discussed, let's try to combine all errors above, to avoid the race.

src/dune_engine/targets.ml
```diff
@@ -263,6 +263,32 @@ module Produced = struct
...
 
+  (* Dummy digest because we want to continue discoverign all the errors
+     and we need some value to return in [mapi]. It will never be returned *)
+  let dummy_digest = Digest.generic ""
+
+  let collect_digests
+    { files; dirs }
+    ~(f : Path.Build.t -> 'a -> Cached_digest.Digest_result.t)
+    =
+    let errors = ref [] in
```
staronj, line 274: We can get rid of `errors : _ list ref` and `dummy_digest` if we return `(_, _) either` from `f` and pipe it through a `List.partition_map ~f:Fn.id`

rgrinberg: Yup, but this means that we'll need create the intermediate list and then throw it away to create the map again. With the current code we:

1. Avoid an intermediate data structure
2. Re-create the map while relying on the existing sorted order

Given that this error case is quite error, it didn't seem right to make the whole code slower just for this.

staronj (approved):

src/dune_shared_cache/dune_shared_cache.ml
```diff
@@ -250,15 +249,16 @@ struct
...
+          Nonempty_list.to_list errors |> List.fold_left ~init:([], []) ~f:process_target
+        in
+        List.rev missing, List.rev errors
       in
       (match missing, errors with
        | [], [] ->
          Code_error.raise
-           "compute_target_digests_or_raise_error: spurious target digest failure"
+           "compute_target_digests_or_raise_error: this is impossible because we should \
+            at least be showing the original error"
```
snowleopard, line 261: I'm surprised we haven't managed to get rid of this unreachable code.

rgrinberg: We probably could if we had a function like:

```
Nonempty_list.partition_map
:   'a Nonempty_list.t
-> f:('a -> ('b, 'c) Either.t) 
-> ('a Nonempty_list.t * 'b list, 'a list * 'b Nonempty_list.t) Either.t
```

But it seems overkill

rgrinberg: I did simplify this code further with `List.partition_map` and added a comment as to why this case is hard to eliminate statically https://github.com/ocaml/dune/pull/8731

PR #8647, opened 2023-09-13
test(pkg): copying opam repository files

Alizter:
This test demonstrates that we don't currently copy the files folder in the opam repository to the lock file.


test/blackbox-tests/test-cases/pkg/opam-package-copy-files.t
```diff
@@ -0,0 +1,46 @@
+  $ . ./helpers.sh
```
rgrinberg, line 1: Could you add a test description?

PR #8648, opened 2023-09-14
pkg: copy files from opam repository to lock dir

Alizter:
The opam solver now copies any files in the `files` directory inside a package directory in an opam repository.

### DIFF
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -393,7 +393,14 @@ module Write_disk = struct
 
   type t = unit -> unit
 
-  let prepare ~lock_dir_path lock_dir =
+  module Files_entry = struct
+    type t =
+      { original_file : Path.t
+      ; local_file : Path.Local.t
+      }
+  end
+
+  let prepare ~lock_dir_path ~files lock_dir =
     let lock_dir_path = Path.source lock_dir_path in
     let remove_dir_if_exists = safely_remove_lock_dir_if_exists_thunk lock_dir_path in
     fun () ->
@@ -408,7 +415,19 @@ module Write_disk = struct
             Dune_lang.Ast.add_loc ~loc:Loc.none sexp |> Dune_sexp.Cst.concrete)
         in
         let pp = Dune_lang.Format.pp_top_sexps ~version:(3, 11) cst in
-        Format.asprintf "%a" Pp.to_fmt pp |> Io.write_file path)
+        Format.asprintf "%a" Pp.to_fmt pp |> Io.write_file path;
+        Package_name.Map.iteri files ~f:(fun package_name files ->
+          let files_dir =
+            Path.relative lock_dir_path (Package_name.to_string package_name ^ ".files")
+          in
+          Path.mkdir_p files_dir;
+          List.iter files ~f:(fun { Files_entry.original_file; local_file } ->
+            let dst = Path.append_local files_dir local_file in
+            Path.mkdir_p (Path.parent_exn dst);
+            Io.copy_file
+              ~src:original_file
+              ~dst:(Path.append_local files_dir local_file)
+              ())))
   ;;
 
   let commit t = t ()
--- a/src/dune_pkg/lock_dir.mli
+++ b/src/dune_pkg/lock_dir.mli
@@ -71,7 +71,19 @@ module Write_disk : sig
   type lock_dir := t
   type t
 
-  val prepare : lock_dir_path:Path.Source.t -> lock_dir -> t
+  module Files_entry : sig
+    type t =
+      { original_file : Path.t
+      ; local_file : Path.Local.t
+      }
+  end
+
+  val prepare
+    :  lock_dir_path:Path.Source.t
+    -> files:Files_entry.t Package_name.Map.Multi.t
+    -> lock_dir
+    -> t
+
   val commit : t -> unit
 end
 
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -458,28 +458,29 @@ let solve_package_list local_packages context =
 (* Scan a path recursively down retrieving a list of all files together with their
    relative path. *)
 let scan_files_entries ~repo_id path =
-  let rec read acc path =
+  (* TODO Add some cycle detection *)
+  let rec read acc dir =
+    let path = Path.append_local path dir in
     match Path.readdir_unsorted_with_kinds path with
-    | Ok files ->
-      List.concat_map files ~f:(fun (filename, kind) ->
+    | Ok entries ->
+      List.fold_left entries ~init:acc ~f:(fun acc (filename, kind) ->
+        let local_path = Path.Local.relative dir filename in
         match (kind : Unix.file_kind) with
-        | S_REG -> [ acc @ [ filename ] ]
-        | S_DIR -> read (acc @ [ filename ]) (Path.relative path filename)
-        | _ -> [])
-    | Error (Unix.ENOENT, _, _) -> []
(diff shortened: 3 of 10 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -482,5 +482,40 @@ let solve_lock_dir solver_env version_preference (repo, repo_id) ~local_packages
       | Ok pkgs_by_name ->
         Lock_dir.create_latest_version pkgs_by_name ~ocaml:None ~repo_id
     in
-    summary, lock_dir)
+    let files =
```
Alizter, line 485: @rgrinberg Is there a better way to do this?

src/dune_pkg/opam_solver.ml
```diff
@@ -482,5 +482,40 @@ let solve_lock_dir solver_env version_preference (repo, repo_id) ~local_packages
...
+      match
+        Package_name.Map.of_list_map opam_packages_to_lock ~f:(fun opam_package ->
+          let files_path = Opam_repo.get_opam_package_files_path repo opam_package in
+          ( Package_name.of_string
+              (OpamPackage.Name.to_string (OpamPackage.name opam_package))
+          , read files_path ))
+      with
+      | Ok files -> files
+      | Error _ ->
+        (* What to do here? *)
```
Alizter, line 517: Not sure what to do in this situation yet.

src/dune_pkg/opam_solver.ml
```diff
@@ -482,5 +482,40 @@ let solve_lock_dir solver_env version_preference (repo, repo_id) ~local_packages
...
+        let rec read acc path =
+          match Path.readdir_unsorted_with_kinds path with
+          | Ok files ->
+            List.concat_map files ~f:(fun (filename, kind) ->
+              match (kind : Unix.file_kind) with
+              | S_REG -> [ acc @ [ filename ] ]
+              | S_DIR -> read (acc @ [ filename ]) (Path.relative path filename)
+              | _ -> [])
+          | Error _ ->
+            (* TODO *)
```
Alizter, line 496: Not sure what to do in this situation either.

src/dune_pkg/lock_dir.mli
```diff
@@ -71,7 +71,12 @@ module Write_disk : sig
   type lock_dir := t
   type t
 
-  val prepare : lock_dir_path:Path.Source.t -> lock_dir -> t
+  val prepare
+    :  lock_dir_path:Path.Source.t
+    -> files:(Path.t * Path.Local.t) Package_name.Map.Multi.t
```
Alizter, line 76: There is some redundant information here since the `Path.t` contains the local path too. Perhaps it would be better to just keep a root and a local path.

src/dune_pkg/lock_dir.ml
```diff
@@ -408,7 +408,19 @@ module Write_disk = struct
...
-        Format.asprintf "%a" Pp.to_fmt pp |> Io.write_file path)
+        Format.asprintf "%a" Pp.to_fmt pp |> Io.write_file path;
+        Package_name.Map.iteri files ~f:(fun package_name files ->
+          let files_dir =
+            Path.relative lock_dir_path (Package_name.to_string package_name ^ ".files")
+          in
+          if List.is_non_empty files then Path.mkdir_p files_dir;
+          List.iter files ~f:(fun (original_file, local_file) ->
+            let dst = Path.append_local files_dir local_file in
+            Path.mkdir_p (Path.parent_exn dst);
```
Alizter, line 419: This might be a bit wasteful, however creating all the subdirectories is a bit annoying otherwise.

src/dune_pkg/lock_dir.ml
```diff
@@ -408,7 +408,19 @@ module Write_disk = struct
             Dune_lang.Ast.add_loc ~loc:Loc.none sexp |> Dune_sexp.Cst.concrete)
         in
         let pp = Dune_lang.Format.pp_top_sexps ~version:(3, 11) cst in
-        Format.asprintf "%a" Pp.to_fmt pp |> Io.write_file path)
+        Format.asprintf "%a" Pp.to_fmt pp |> Io.write_file path;
+        Package_name.Map.iteri files ~f:(fun package_name files ->
+          let files_dir =
+            Path.relative lock_dir_path (Package_name.to_string package_name ^ ".files")
+          in
+          if List.is_non_empty files then Path.mkdir_p files_dir;
```
rgrinberg, line 416: How can this list be empty? If there are no files, why is the package even included in the map?

src/dune_pkg/lock_dir.ml
rgrinberg: No, it's totally fine.

src/dune_pkg/lock_dir.mli
rgrinberg: I don't think it matters if there's a redundancy. However, I do think that the tuple is making it harder to read than necessary. Can you add a proper record to represent this?

src/dune_pkg/opam_solver.ml
rgrinberg: When is this error possible?

rgrinberg: Okay, do we have a test for this? If no, could you please add one?

Alizter: There doesn't appear to be a test for it however I am not quite sure how it could be triggered in the first place. The relevant function is `solve_lock_dir` in `opam_solver`. The only way this error can be triggered is if the solver returns duplicate opam packages.

src/dune_pkg/opam_solver.ml
Alizter: This situation happens when we are unable to read a directory. It returns a Unix error so I guess the best we can do is raise it.

src/dune_pkg/lock_dir.ml
Alizter: Good point.

src/dune_pkg/opam_solver.ml
Alizter: I've added a test demonstrating what happens if we get a unix error whilst trying to read the `files/` directory.

test/blackbox-tests/test-cases/pkg/opam-package-files-unix-error.t
```diff
@@ -0,0 +1,49 @@
...
+  $ cat >$opam_repo/files/$fname2 <<EOF
+  > bar
+  > EOF
+We remove the read permissions for dir/
+
+  $ chmod -r $opam_repo/files/dir
+
+The error message should have a location for the opam repository.
+
+This does not currently seem to be the case.
```
Alizter, line 34: This will be fixed once this todo is resolved:
https://github.com/ocaml/dune/blob/b982acf9a7fb52474720bc5257b812f824b87616/bin/pkg.ml#L368-L369

cc @Leonidas-from-XIV

src/dune_pkg/opam_solver.mli
```diff
@@ -13,4 +13,6 @@ val solve_lock_dir
   -> Version_preference.t
   -> Opam_repo.t * (Loc.t * Repository_id.t) option
   -> local_packages:OpamFile.OPAM.t OpamTypes.name_map
-  -> (Summary.t * Lock_dir.t, [ `Diagnostic_message of _ Pp.t ]) result
+  -> ( Summary.t * Lock_dir.t * Lock_dir.Write_disk.Files_entry.t Package_name.Map.Multi.t
```
Alizter, line 16: Shall I make this a record?

rgrinberg: Not in this PR

PR #8654, opened 2023-09-14
feature(pkg): add patching support to dune pkg

Alizter:
We add support for the patches field in an opam file. We create a custom patch action that makes sure we can patch the files in quesiton by materializing any symlinks.

### DIFF
--- a/src/dune_patch/dune_patch.ml
+++ b/src/dune_patch/dune_patch.ml
@@ -1,0 +1,139 @@
+open Stdune
+
+module Re = struct
+  include Dune_re
+
+  module Group = struct
+    include Group
+
+    let get_opt group n = if Group.test group n then Some (get group n) else None
+  end
+end
+
+include struct
+  open Dune_engine
+  module Action = Action
+  module Process = Process
+  module Utils = Utils
+  module Action_builder = Action_builder
+end
+
+include struct
+  open Dune_lang
+  module Value = Value
+end
+
+let re =
+  let line xs = Re.seq ((Re.bol :: xs) @ [ Re.eol ]) in
+  let followed_by_line xs = Re.seq [ Re.str "\n"; line xs ] in
+  Re.compile
+  @@ Re.seq
+       [ Re.alt
+           [ line [ Re.str {|--- a/|}; Re.group (Re.rep1 Re.notnl) ]
+           ; line [ Re.str {|--- |}; Re.group (Re.rep1 Re.notnl) ]
+           ]
+       ; Re.alt
+           [ followed_by_line [ Re.str {|+++ b/|}; Re.group (Re.rep1 Re.notnl) ]
+           ; followed_by_line [ Re.str {|+++ |}; Re.group (Re.rep1 Re.notnl) ]
+           ]
+       ]
+;;
+
+module Patch = struct
+  (* CR-someday alizter: more parsed infromation about the patch should go here.
+     Eventually we wish to replace the patch command inside the patch action with a pure
+     OCaml implementation. *)
+  type t =
+    | New of Path.Local.t
+    | Delete of Path.Local.t
+    | Replace of Path.Local.t
+end
+
+let patches_of_string patch_string =
+  Re.all re patch_string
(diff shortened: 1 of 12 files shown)


src/dune_patch_parser/dune_patch_parser.mli
```diff
@@ -0,0 +1,4 @@
+open Stdune
+
+val files_of_patch : dir:Path.t -> string -> Path.t list
+val files_of_patch_file : dir:Path.t -> Path.t -> Path.t list
```
rgrinberg, line 4: Probably you could just move the entire implementation of the patch action here (minus the engine scaffolding) and write even more extensive unit tests.

Alizter:
I've added a new test that demonstrates multiple patches which contain multiple modified files.

src/dune_patch_parser/dune_patch_parser.mli
Alizter: OK I'll call this library `dune_patch` too.

test/expect-tests/dune_patch/dune_patch_tests.ml
```diff
@@ -0,0 +1,46 @@
+open Stdune
+
+let () = Dune_tests_common.init ()
+
+let basic =
```
rgrinberg, line 5: You should add tests testing the entire action here. This is where it will be useful to expose `Spec.action` without all the scaffolding.

Alizter: OK, I'll write some tests.

src/dune_patch/dune_patch.ml
```diff
@@ -0,0 +1,90 @@
...
+    (* 2. Materialize those files. *)
+    List.iter files ~f:(fun file ->
+      Io.copy_file ~src:file ~dst:(Path.extend_basename file ~suffix:".for_patch") ());
+    List.iter files ~f:(fun file ->
+      Unix.rename (Path.to_string file ^ ".for_patch") (Path.to_string file));
+    (* 3. Apply the patch. *)
+    (* TODO opam has a preprocessing step that we should probably apply *)
+    Process.run
+      ~dir:eenv.working_dir
+      ~display:Dune_engine.Display.Quiet
```
rgrinberg, line 69: I think this should be inherited from `Clflags.display` - not in this module, but in whoever is calling this function.

src/dune_patch/dune_patch.mli
```diff
@@ -0,0 +1,4 @@
+open Stdune
+
+val files_of_patch : dir:Path.t -> string -> Path.t list
+val action : input:Path.t -> Dune_engine.Action.t
```
rgrinberg, line 4: This should return `Dune_engine.Action.t Action_builder.t` with the appropriate dependencies on the input and the patch binary.

src/dune_patch/dune_patch.ml
Alizter: Would it make sense to add display to eenv?

rgrinberg: Might be. Let's not worry about this now.

src/dune_patch/dune_patch.ml
```diff
@@ -0,0 +1,90 @@
...
+  let encode (p : (_, _) t) input _ : Dune_lang.t =
+    List [ Dune_lang.atom_or_quoted_string name; input p ]
+  ;;
+
+  let action p ~(ectx : Action.Ext.context) ~(eenv : Action.Ext.env) =
+    let open Fiber.O in
+    let* () = Fiber.return () in
+    let input = Value.to_string ~dir:eenv.working_dir (Value.Path p) in
+    let patch =
+      let path = eenv.env |> Env_path.path in
```
rgrinberg, line 48: This is wrong. It should be be provided by the action and looked up exactly from where it's being looked up now.

Alizter: Fixed.

rgrinberg:
I ended up simplifying things a bit so that we can replace the patch binary with an OCaml implementation transparently.

PR #8656, opened 2023-09-15
test(pkg): test substs field in opam file

Alizter:
We test how the substs field in an opam file gets translated to dune.lock. Currently nothing is done.

I have a fix for this that can be PRed after #8654

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/opam-package-with-subst.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-package-with-subst.t
@@ -1,0 +1,54 @@
+We test how opam files with substs fields are translated into the dune.lock file.
+ 
+  $ . ./helpers.sh
+
+Generate a mock opam repository
+  $ mkdir -p mock-opam-repository
+  $ cat >mock-opam-repository/repo <<EOF
+  > opam-version: "2.0"
+  > EOF
+
+Make a package with a substs field 
+  $ mkpkg with-substs <<EOF
+  > opam-version: "2.0"
+  > substs: ["foo.ml"]
+  > build: [ "sh" "-c" "[ -e foo.ml ] && cat foo.ml" ]
+  > EOF
+
+  $ opam_repo=mock-opam-repository/packages/with-substs/with-substs.0.0.1
+
+  $ solve_project <<EOF
+  > (lang dune 3.8)
+  > (package
+  >  (name x)
+  >  (allow_empty)
+  >  (depends with-substs)) 
+  > EOF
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/pkg/opam-package-with-subst.t
```diff
@@ -0,0 +1,54 @@
...
+  $ mkdir -p mock-opam-repository
+  $ cat >mock-opam-repository/repo <<EOF
+  > opam-version: "2.0"
+  > EOF
+
+Make a package with a substs field 
+  $ mkpkg with-substs <<EOF
+  > opam-version: "2.0"
+  > substs: ["foo.ml"]
+  > build: [ ["[" "-e" "foo.ml" "]"] [ "cat" "foo.ml"] ]
```
rgrinberg, line 15: How about just `"-e foo.ml && cat foo.ml"`?

test/blackbox-tests/test-cases/pkg/opam-package-with-subst.t
```diff
@@ -0,0 +1,54 @@
...
+  > EOF
+
+The lockfile should contain the substitute action. The generation step currently doesn't
+add this in.
+
+  $ cat dune.lock/with-substs.pkg 
+  (version 0.0.1)
+  
+  (build
+   (run [ -e foo.ml ]))
```
rgrinberg, line 41: By the way, this translation step seems like a bug. This should be a system action

test/blackbox-tests/test-cases/pkg/opam-package-with-subst.t
Alizter: Sure let me try that.

Alizter: Well it will have to be `[ -e foo.ml ] && cat foo.ml` however since run is being used that doesn't work. Let me also submit a fix for that.

test/blackbox-tests/test-cases/pkg/opam-package-with-subst.t
Alizter: Opened #8663.

rgrinberg: The bug is in your opam file. It should use `sh -c` to execute the shell command.

Alizter: Fixed.

PR #8662, opened 2023-09-15
test: missing opam file now causes build failure

emillon:
Since #8518, we fail in the case where `(generate_opam_files)` is true but the generated file is not part of the release tarball.

### DIFF
--- a/test/blackbox-tests/test-cases/missing-opam-generated-file.t
+++ b/test/blackbox-tests/test-cases/missing-opam-generated-file.t
@@ -1,0 +1,21 @@
+This shows a difference since #8518: we set up a project with a generated opam
+file. If the resulting file is not present in the repository (or if it has been
+renamed to just `opam` as in this example), we now fail.
+
+A related issue happens when `pkg.opam` exists but is not up to date. The
+instruction in the opam file would previously consider the original version but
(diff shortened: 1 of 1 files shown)


emillon:
@rgrinberg this is technically a regression introduced by #8518 (there are 94 failures in the alpha1 run related to that). I think we can argue that these packages have an issue, but I wonder if we can do better than just failing.

rgrinberg:
I suppose we could guard the check behind a project version. Although it makes the fix sort of useless because upstream maintainers don't want to upgrade the dune language in every single package to make the builds reproducible.

emillon:
I agree it doesn't seem ideal. Removing duplicates, it's 17 releases (12 unique packages) so it's something we can easily communicate and patch.

emillon:
I think we can set bounds + send patches upstream but this can be discussed at the next team meeting.

rgrinberg:
Upstream patches that allows for this PR to work https://github.com/ocaml/dune/pull/8512 would be welcome as well.

It's the same issue just with different rules.

rgrinberg:
Now that I think about it, we could adjust the feature as following:

Instead of ignoring the rule completely, we could make it a normal instead of a promote rule.

emillon:
I'll merge this as a regression test and open an issue to track this idea.

PR #8663, opened 2023-09-15
fix(pkg): use system instead of run when translating build action

Alizter:
We were using the `run` action to run arbitrary shell commands in the opam build step. This is wrong because it doesn't allow us to run certain commands. We therefore change back to `system` which is more correct, however in the future we might want to refine this depending on the kind of command.

For context, it was noticed in #8656 that this was incorrect.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
+++ b/test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
@@ -117,41 +117,41 @@ Package which has boolean where string was expected. This should be caught while
   (version 0.0.1)
   
   (install
-   (run %{make} install))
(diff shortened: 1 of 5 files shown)


test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
```diff
@@ -184,54 +184,54 @@ Package which has boolean where string was expected. This should be caught while
    (progn
     (when
      %{pkg-self:foo}
-     (run echo a))
+     (system "echo a"))
```
Alizter, line 187: For example, we could improve an echo step and translate it into the echo action.

PR #8666, opened 2023-09-15
chore(ci): switch to OCaml 5.2 trunk and intro 5.1 jobs

Alizter:
Not sure if we want so many jobs. @rgrinberg WDYT?

I think that dropping OCaml 5.0 might be better than not adding 5.1.


rgrinberg:
We can get rid of 5.0

Alizter:
@rgrinberg Done.

PR #8668, opened 2023-09-16
fix(pkg): add support for %{make}

rgrinberg:


test/blackbox-tests/test-cases/pkg/make.t
Alizter, line 1: Do you need to add make to the test deps?

PR #8669, opened 2023-09-16
feature(pkg): translate substs field of opam file into build action

### LINKED ISSUE #8640: Full Opam Support: substs
Opam files may contain the `substs` field:

```
substs: ["sexp-dir.patch"]
```

This should be converted to the `substitute` action.

See the `substitute.t` test for an example.

Alizter:
This adds support for the translation of the substs field of an opam file.
- fixes #8640 

We have two tests: one with a basic substitution and another with a substitution that modifies a patch file. This is in order to make explicit the order of substitution and patching implicitly present in opam.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -386,6 +386,12 @@ let opam_commands_to_actions package (commands : OpamTypes.command list) =
     | [] -> None)
 ;;
 
+let make_action = function
+  | [] -> None
+  | [ action ] -> Some action
+  | actions -> Some (Action.Progn actions)
+;;
+
 let opam_package_to_lock_file_pkg ~repo ~local_packages opam_package =
   let name = OpamPackage.name opam_package in
   let version = OpamPackage.version opam_package |> OpamPackage.Version.to_string in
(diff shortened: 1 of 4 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -424,27 +419,30 @@ let opam_package_to_lock_file_pkg ~repo ~local_packages opam_package =
...
+    in
+    let build_step =
+      opam_commands_to_actions opam_package (OpamFile.OPAM.build opam_file) |> progn_opt
     in
-    opam_commands_to_action opam_package (OpamFile.OPAM.build opam_file)
-    |> Option.map ~f:(fun action ->
-      match patch_step with
-      | None -> action
-      | Some patch_step -> Action.Progn [ patch_step; action ])
+    List.filter_opt [ subst_step; patch_step; build_step ] |> progn_opt
```
rgrinberg, line 442: You could leave all the `{subst,patch,build}_step` as lists and use `List.concat`. That way you could inline `progn_opt` which is a bit of an ugly function to keep around.

Alizter: Sure thing.

src/dune_pkg/opam_solver.ml
```diff
@@ -386,17 +386,6 @@ let opam_commands_to_actions package (commands : OpamTypes.command list) =
     | [] -> None)
 ;;
 
-(* returns:
-   [None] if the command list is empty
-   [Some (Action.Run ...)] if there is a single command
-   [Some (Action.Progn [Action.Run ...; ...])] if there are multiple commands *)
-let opam_commands_to_action package (commands : OpamTypes.command list) =
```
rgrinberg, line 393: You could keep this function and not have to do the pattern match twice.

Alizter: Well not really since this matches on `opam_commands_to_actions`. That's why I introduced `prog_opt` before. If you like I can bring that back and use it instead.

src/dune_pkg/opam_solver.ml
```diff
@@ -424,27 +413,34 @@ let opam_package_to_lock_file_pkg ~repo ~local_packages opam_package =
...
-        |> List.map ~f:(fun (x, _) ->
-          Action.Patch
-            (String_with_vars.make_text Loc.none (OpamFilename.Base.to_string x)))
-      with
-      | [] -> None
-      | [ x ] -> Some x
-      | xs -> Some (Action.Progn xs)
+      OpamFile.OPAM.patches opam_file
+      |> List.map ~f:(fun (x, _) ->
+        Action.Patch (String_with_vars.make_text Loc.none (OpamFilename.Base.to_string x)))
```
gridbugs, line 429: FYI `Action.When` should work here to apply the filter (happy for this to come in a later change though)

gridbugs (approved):

src/dune_pkg/opam_solver.ml
Alizter: Let's do that in a later change. Thanks!

PR #8670, opened 2023-09-16
fix(pkg): improve unknown package error

rgrinberg:


src/dune_rules/pkg_rules.ml
```diff
@@ -861,8 +861,11 @@ end = struct
     fun db ctx (loc, name) ->
       Memo.exec memo (db, ctx, name)
       >>| function
-      | None -> User_error.raise ~loc [ Pp.text "Unknown package" ]
       | Some s -> s
+      | None ->
+        User_error.raise
+          ~loc
```
Alizter, line 867: When do we have a loc?

PR #8672, opened 2023-09-16
feature(pkg): translate extra-source field of opam file into lock file

### LINKED ISSUE #8639: Support conversion of extra-sources
Opam files may contain the following specification for additional sources:

```
extra-files: [
  ["patches/fallback.patch" "md5=e5080d62e298c8002d7d4a63dc3a71db"]
  ["src/bytes0.ml" "md5=a3d9f0482f11047341319b3eabba7270"]
]
```

These should be converted to equivalent `extra_sources` fields in the definition of lock_dir.

Alizter:
The extra-source field in an opam file gets translated.

- fixes #8639 

Are there any meaningful locations we can use here?

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -401,18 +401,31 @@ let opam_package_to_lock_file_pkg ~repo ~local_packages opam_package =
   let name = OpamPackage.name opam_package in
   let version = OpamPackage.version opam_package |> OpamPackage.Version.to_string in
   let dev = OpamPackage.Name.Map.mem name local_packages in
+  let opam_file =
+    match OpamPackage.Name.Map.find_opt name local_packages with
+    | None -> Opam_repo.load_opam_package repo opam_package
+    | Some local_package -> local_package
+  in
+  let extra_sources =
+    OpamFile.OPAM.extra_sources opam_file
+    |> List.map ~f:(fun (opam_basename, opam_url) ->
+      ( Path.Local.of_string (OpamFilename.Base.to_string opam_basename)
+      , let url = Loc.none, OpamUrl.to_string (OpamFile.URL.url opam_url) in
+        let checksum =
+          match OpamFile.URL.checksum opam_url with
+          | [] -> None
+          (* opam discards the later checksums, so we only take the first one *)
+          | checksum :: _ -> Some (Loc.none, Checksum.of_opam_hash checksum)
+        in
+        Lock_dir.Source.Fetch { Lock_dir.Source.url; checksum } ))
+  in
   let info =
     { Lock_dir.Pkg_info.name = Package_name.of_string (OpamPackage.Name.to_string name)
     ; version
     ; dev
     ; source = None
-    ; extra_sources = []
+    ; extra_sources
     }
-  in
-  let opam_file =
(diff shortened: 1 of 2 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -401,19 +401,35 @@ let opam_package_to_lock_file_pkg ~repo ~local_packages opam_package =
...
+      ( Path.Local.of_string (OpamFilename.Base.to_string opam_basename)
+      , (* TODO loc? *)
+        let url = Loc.none, OpamUrl.to_string (OpamFile.URL.url opam_url) in
+        let checksum =
+          match OpamFile.URL.checksum opam_url with
+          | [] -> None
+          (* opam discards the later checksums, so we only take the first one *)
+          | checksum :: _ ->
+            (* TODO: any meaningful loc we can give this? *)
+            Some (Loc.none, Checksum.of_opam_hash checksum)
```
rgrinberg, line 421: The process of translation opam files to the lock directory can't really fail if the opam file is valid. So the location is meaningless.

src/dune_pkg/opam_solver.ml
```diff
@@ -401,19 +401,35 @@ let opam_package_to_lock_file_pkg ~repo ~local_packages opam_package =
...
+    match OpamPackage.Name.Map.find_opt name local_packages with
+    | None -> Opam_repo.load_opam_package repo opam_package
+    | Some local_package -> local_package
+  in
+  let extra_sources =
+    OpamFile.OPAM.extra_sources opam_file
+    |> List.map ~f:(fun (opam_basename, opam_url) ->
+      ( Path.Local.of_string (OpamFilename.Base.to_string opam_basename)
+      , (* TODO loc? *)
+        let url = Loc.none, OpamUrl.to_string (OpamFile.URL.url opam_url) in
```
rgrinberg, line 414: ditto

test/blackbox-tests/test-cases/pkg/opam-package-with-extra-source.t
```diff
@@ -31,8 +31,15 @@ Make a package with an extra-source field
...
 
   $ cat dune.lock/with-extra-source.pkg 
   (version 0.0.1)
+  
+  (extra_sources
+   (some/file
+    (fetch
+     (url https://some-url)
+     (checksum
+      sha256=8beda92f97cde6d4a55a836ca6dc9f860bb5f1a6b765b80be4594943288571cf))))
```
rgrinberg, line 44: I forgot we haven't fixed the frontend for this yet. Definitely a little ugly.

Could you add an example with md5 as well?

rgrinberg:
Also, are there any tests with multiple checksums?

Alizter:
@rgrinberg I've added packages with multiple cheksums and md5.

rgrinberg:
@Leonidas-from-XIV any idea what multiple checksums are for?

For now, we can just leave a comment that we aren't doing anything with them.

Leonidas-from-XIV:
@Alizter It is used in `validate_and_add_to_cache`, I linked the relevant source code in #8639. Its a bit strange but I guess if the checksums were validated when the file was downloaded and cached, then for retrieving it from the cache just using the first hash as an integrity and less of a security check is acceptable.

PR #8680, opened 2023-09-18
test(pkg): patch with filter

Alizter:
We demonstrate that adding a filter to a patches field of an opam file does nothing in the translation to our lock file.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/opam-package-with-patch-filter.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-package-with-patch-filter.t
@@ -1,0 +1,61 @@
+We test the translation of an opam package with a patches field with a filter into a dune
+lock file.
+
+  $ . ./helpers.sh
+
+Generate a mock opam repository
+  $ mkdir -p mock-opam-repository
+  $ cat >mock-opam-repository/repo <<EOF
+  > opam-version: "2.0"
+  > EOF
+
+Make a package with a patch behind a filter
+  $ mkpkg with-patch-filter <<EOF
+  > opam-version: "2.0"
+  > patches: ["foo.patch" {with-test}]
+  > build: ["cat" "foo.ml"]
+  > EOF
+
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/pkg/opam-package-with-patch-filter.t
```diff
@@ -0,0 +1,59 @@
+  $ . ./helpers.sh
```
rgrinberg, line 1: Can you add a description to this test?

test/blackbox-tests/test-cases/pkg/opam-package-with-patch-filter.t
```diff
@@ -0,0 +1,61 @@
...
+Generate a mock opam repository
+  $ mkdir -p mock-opam-repository
+  $ cat >mock-opam-repository/repo <<EOF
+  > opam-version: "2.0"
+  > EOF
+
+Make a package with a patch behind a filter
+  $ mkpkg with-patch-filter <<EOF
+  > opam-version: "2.0"
+  > patches: ["foo.patch" {with-test}]
```
Alizter, line 15: @rgrinberg I've removed the `os` stuff that isn't in opam and replaced it with `with-test` that I am sure is always here. Since we don't build the package in this mode, it should work when we add the filters.

rgrinberg: The problem with this test is that it's not obvious that you're solving with the test filter turned. Why don't you use os instead?

Alizter: Does it matter what `with-test` is when we solve? It should always give a `when` anyway.

rgrinberg: Oh yeah, we only evaluate filters for dependencies.

PR #8688, opened 2023-09-18
fix(pkg): expand with-test into %{with-test}

Alizter:
We fix the expansion of {with-test} in an opam file to %{with-test} as a pfrom. For now this is always set to be true.

I've left some CRs mentioning that this temporary default should be changed to something more appropriate.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
+++ b/test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
@@ -209,7 +209,7 @@ Package which has boolean where string was expected. This should be caught while
      (run echo g))
     (when
      (and
-      %{pkg-self:with-test}
+      %{with-test}
       (< %{pkg:ocaml:version} 5.0.0))
      (run echo h))
     (when
--- a/test/blackbox-tests/test-cases/pkg/opam-var/opam-var-pkg.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-var/opam-var-pkg.t
@@ -103,10 +103,10 @@ corresponding Dune version.
     (run echo 32 %{pkg:foo:dev})
     (run echo 33 %{pkg-self:opamfile})
     (run echo 34 %{pkg:foo:opamfile})
-    (run echo 35 %{pkg-self:with-test})
+    (run echo 35 %{with-test})
     (run echo 36 %{pkg-self:with-test})
     (run echo 37 %{pkg:foo:with-test})
-    (run echo 38 %{pkg-self:with-doc})
+    (run echo 38 %{with-doc})
     (run echo 39 %{pkg-self:with-doc})
     (run echo 40 %{pkg:foo:with-doc})
     (run echo 41 %{pkg-self:with-dev-setup})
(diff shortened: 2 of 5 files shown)


test/blackbox-tests/test-cases/pkg/convert-opam-commands.t
```diff
@@ -214,7 +214,7 @@ Package which has boolean where string was expected. This should be caught while
      (run echo g))
     (when
      (and
-      %{pkg-self:with-test}
+      %{with-test}
```
rgrinberg, line 217: Is there no analogous test for `with-doc`?

Alizter: I could add one, I wasn't sure if we wanted to go ahead with this.

rgrinberg: It's worth improving the test to include this.

Alizter: I won't improve this test, but the other opam var tests I added do test for that. I've rebased this PR now so you should see we are tested both with-doc and with-test properly.

Alizter:
@rgrinberg I should note however that `with-test` and `with-doc` are package variables and not global variables like this PR pretends. I don't think this is the correct way to do this.

cc @gridbugs

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-pkg.t
```diff
@@ -103,10 +103,10 @@ corresponding Dune version.
     (run echo 32 %{pkg:foo:dev})
     (run echo 33 %{pkg-self:opamfile})
     (run echo 34 %{pkg:foo:opamfile})
-    (run echo 35 %{pkg-self:with-test})
```
Alizter, line 106: This actually might be more correct than the new version. It all depends on how we want to handle with-test.

Alizter:
Marking as draft for now, as I am not happy with it.

gridbugs:
I think it makes sense for these to be package variables expanding to `%{pkg-self:with-test}%` and `%{pkg:foo:with-test}%` the same way we treat `name` or `version`.

rgrinberg:
When are these variables useful though? They aren't useful for building packages in opam-repository. Without a concrete reason, I'd rather not spend the cycles supporting them.

gridbugs:
> They aren't useful for building packages in opam-repository.

Many packages have a build command like:
```
    "dune"
    "build"
    "-p"
    name
    "-j"
    jobs
    "@install"
    "@runtest" {with-test}
    "@doc" {with-doc}
```

rgrinberg:
> > They aren't useful for building packages in opam-repository.
> 
> Many packages have a build command like:
> 
> ```
>     "dune"
>     "build"
>     "-p"
>     name
>     "-j"
>     jobs
>     "@install"
>     "@runtest" {with-test}
>     "@doc" {with-doc}
> ```

I can only think of two use cases allowed by the command above:

1. Running tests for a dependency (rarely useful)
2. Building the documentation for a particular package (somewhat more useful)

Both can be easily supported without having package specific variables.

PR #8689, opened 2023-09-18
feature(pkg): initialize context with packages

### LINKED ISSUE #7707: Initialize Scope/Toolchain/Context from Lock File
To be able to use binaries and libraries provided by packages built by dune we need to appropriately initialize:

* The OCaml toolchain. It should be possible to discover the toolchain from the "ocaml" package.

* The public scope. We should be able to initialize it from the `OCAMLPATH` exported by the packages in the lock file. This will require re-working findlib to make it usable with build paths.

* The context should now be able to locate binaries using the set of installed packages.

rgrinberg:
This PR does the following:

* Initialize the context with the OCaml toolchain from the lock dir
* Initialize the findlib database with all the packages installed from the lock dir
* Introduce the environment variable `DUNE_PKG_OVERRIDE_OCAML` to prevent dune from installing OCaml. This is useful to make sure we can run the test suite in a reasonable time.

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -105,6 +105,11 @@ module Lock_dir = struct
   ;;
 
   let get (ctx : Context_name.t) : t Memo.t = get_path ctx >>= Load.load
+
+  let has_lock ctx =
+    let* path = get_path ctx in
+    Fs_memo.dir_exists (In_source_dir path)
+  ;;
 end
 
 module Paths = struct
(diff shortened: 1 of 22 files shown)


src/dune_rules/pkg_rules.ml
```diff
@@ -1415,8 +1455,18 @@ module Gen_rules = Build_config.Gen_rules
...
-  let+ pkg = resolve db.packages context (Loc.none, name) in
+  let* db = DB.get context in
+  let+ pkg =
+    Resolve.resolve db context (Loc.none, name)
+    >>| function
+    | `Inside_lock_dir pkg -> pkg
+    | `System_provided ->
+      User_error.raise
+        [ Pp.textf
+            "no rules for %s because it's set as provided by the system"
```
gridbugs, line 1466: Can you change this error to "There are no rules for ..." for clarity (assuming I've interpreted it correctly)

gridbugs (approved):

PR #8692, opened 2023-09-18
test(pkg): modify test to use switch instead

Alizter:
<s>This makes this test more portable.</s>

This avoids `with-test` as it might be treated in a special way.


rgrinberg:
The reason why we want to use `switch` over `with-test` isn't portability but because `with-test` might be treated in a special way.

PR #8695, opened 2023-09-18
test(pkg): clean up mock opam repo creation

Alizter:


test/blackbox-tests/test-cases/pkg/helpers.sh
```diff
@@ -18,6 +18,14 @@ show_pkg_cookie() {
   $dune internal dump $pkg_root/$1/target/cookie
 }
 
+if [ ! -d mock-opam-repository ]
```
rgrinberg, line 21: Could you make this an explicit function instead?

Alizter: That would mean we have to call it everywhere as well. Is that ok?

rgrinberg: Yeah, that's how we do it for all other setup functions in the tests.

PR #8697, opened 2023-09-18
test(pkg): translation of setenv field of opam file into lock file

Alizter:


test/blackbox-tests/test-cases/pkg/opam-package-with-setenv.t
```diff
@@ -0,0 +1,53 @@
...
+  > setenv: [
+  >  [EXPORTED_ENV_VAR = "Hello from the other package!"]
+  > ]
+  > EOF
+
+Make another package that depends on that and outputs the exported env var
+  $ mkpkg deps-on-with-setenv <<EOF
+  > opam-version: "2.0"
+  > depends: [ "with-setenv" ]
+  > build: [ "printenv" "EXPORTED_ENV_VAR" ]
```
rgrinberg, line 18: Let's just use the shell to print an environment variable

PR #8701, opened 2023-09-19
feature(pkg): translate build-env from opam file into lock dir

### LINKED ISSUE #8642: Full Opam Support: build-env
Opam allows to set the environment as part of the package build. For example:

```
build-env: [
  [CI = ""]
]
```

In dune, we have a corresponding `setenv` action.

Alizter:
- Comes after #8700 
- Fixes #8642

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -386,6 +386,22 @@ let opam_commands_to_actions package (commands : OpamTypes.command list) =
     | [] -> None)
 ;;
 
+let opam_env_update_to_env_update ((var, env_op, value_string, _) : OpamTypes.env_update)
+  : String_with_vars.t Action.Env_update.t
+  =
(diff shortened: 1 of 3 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -450,12 +474,15 @@ let opam_package_to_lock_file_pkg ~repo ~local_packages opam_package =
     let build_step =
       opam_commands_to_actions opam_package (OpamFile.OPAM.build opam_file)
     in
-    List.concat [ subst_step; patch_step; build_step ] |> make_action
+    List.concat [ subst_step; patch_step; build_step ]
+    |> make_action
+    |> Option.map ~f:build_env
```
rgrinberg, line 479: You should add a test for an opam file that has no install/build and yet has `build-env`.

Alizter: Done.

PR #8706, opened 2023-09-19
fix: make ignored rules fallback

### LINKED ISSUE #8703: missing opam file causes build failure
Now that I think about it, we could adjust the feature as following:

Instead of ignoring the rule completely, we could make it a normal instead of a promote rule.

_Originally posted by @rgrinberg in https://github.com/ocaml/dune/issues/8662#issuecomment-1723020909_

emillon:
Fixes #8703

This is a lighter change than the original in #8518.

### DIFF
--- a/src/dune_rules/super_context.ml
+++ b/src/dune_rules/super_context.ml
@@ -259,31 +259,29 @@ let extend_action t ~dir action =
 ;;
 
 let make_rule t ?mode ?loc ~dir { Action_builder.With_targets.build; targets } =
-  match mode with
-  | Some mode when Rule_mode_decoder.is_ignored mode ~until_clean:`Keep -> None
-  | _ ->
-    let build = extend_action t build ~dir in
-    Some
-      (Rule.make
-         ?mode
-         ~info:(Rule.Info.of_loc_opt loc)
-         ~context:(Some (Context.build_context (Env_tree.context t)))
-         ~targets
-         build)
+  let (mode : Rule.Mode.t option) =
+    match mode with
+    | Some mode when Rule_mode_decoder.is_ignored mode ~until_clean:`Keep -> Some Fallback
+    | _ -> mode
+  in
+  let build = extend_action t build ~dir in
+  Rule.make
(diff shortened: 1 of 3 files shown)


rgrinberg:
To be clear, making `--ignore-promoted-rules` depend on the version in the dune-project is a mistake. This option should work for all rules (internal and user defined) across all versions of the dune language.

emillon:
Ok. We can remove that version check. It's likely to surface some issues in some projects, though. What do we do for 3.11? Is the fix in this PR enough, and we'll remove the version check for user rules in 3.12?

rgrinberg:
> It's likely to surface some issues in some projects, though. What do we do for 3.11? Is the fix in this PR enough, and we'll remove the version check for user rules in 3.12?

I'd just do it in one go. It becomes hard to keep a map in your head of how things work between different versions of dune if we split a logical change in chunks like that.

Tbh, if this change is so painful to include in 3.11, you could always revert it in that branch and we could keep it in `main` until we decide to make another release.

emillon:
OK. The next round of testing will reveal what breaks. I'll do the other change in a separate PR to ease reverting.

emillon:
Other fix is in #8721

src/dune_rules/super_context.ml
```diff
@@ -259,31 +259,29 @@ let extend_action t ~dir action =
...
 
 let add_rule_get_targets t ?mode ?loc ~dir build =
-  match make_rule t ?mode ?loc ~dir build with
-  | None -> Memo.return None
-  | Some rule ->
-    let+ () = Rules.Produce.rule rule in
-    Some rule.targets
+  let rule = make_rule t ?mode ?loc ~dir build in
+  let+ () = Rules.Produce.rule rule in
+  Some rule.targets
```
rgrinberg, line 284: This function no longer needs to return `option`

rgrinberg: https://github.com/ocaml/dune/pull/8725

PR #8712, opened 2023-09-20
test(pkg): split out and expand env update tests for setenv 

Alizter:
This reverts commit 977a943858ae1a32e3cbd0e5708952371cccf229.

We also add in the tests for each env update operation. Together with #7742 it highlights some shortcomings of our env var handling.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/env-update.t/run.t
+++ b/test/blackbox-tests/test-cases/pkg/env-update.t/run.t
@@ -1,0 +1,79 @@
+We have 3 opam packages depending on each other in order. We test how setenv interacts
+with the various env update operators.
+
+  $ . ./env-update-helpers.sh
+
+  $ test 
+  package1
+    [ override = "package1" ]
+    [ prepend_sep := "package1" ]
+    [ prepend += "package1" ]
+    [ append_sep =: "package1" ]
+    [ append =+ "package1" ]
+  package2
+    [ override = "package2" ]
+    [ prepend_sep := "package2" ]
+    [ prepend += "package2" ]
+    [ append_sep =: "package2" ]
+    [ append =+ "package2" ]
+  package3
+    [ override = "package3" ]
+    [ prepend_sep := "package3" ]
+    [ prepend += "package3" ]
+    [ append_sep =: "package3" ]
+    [ append =+ "package3" ]
+  
+  package1:
+     override: initial value
+  prepend_sep: initial value
+      prepend: initial value
+   append_sep: initial value
+       append: initial value
+  
+  package2:
+     override: initial value
+  prepend_sep: initial value
+      prepend: initial value
+   append_sep: initial value
+       append: initial value
+  
+  package3:
+     override: initial value
+  prepend_sep: initial value
+      prepend: initial value
+   append_sep: initial value
+       append: initial value
+
+Here are some properties the values above should have:
+
+= override
+The following should hold:
+- package1 should have the initial value
+- package(n) shuld have value package(n-1)
+This does not currently hold.
+ 
+:= prepend seperator
+The following should hold:
+- All values should have a trailing seperator :
+- package1 should have the initial value
+- package3 should output "package2:package1:initial value:"
+This does not currently hold.
+
++= prepend
+The following should hold:
+- package1 should have the initial value
+- package3 should output "package2:package1:initial value"
+This does not currently hold.
+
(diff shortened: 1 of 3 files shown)


rgrinberg:
Can you add the tests that are going to replace these in this PR?

test/blackbox-tests/test-cases/pkg/helpers.sh
```diff
@@ -50,3 +50,17 @@ cat >dune.lock/lock.dune <<EOF
 (lang package 0.1)
 EOF
 }
+
+# [mkenvtestpkg <package> <depends> <env_update_op>] makes an opam <package> that depends
+# on <depends> and updated "env_var" using <env_update_op>
+mkenvtestpkg() {
+  mkpkg $1 <<EOF
+opam-version: "2.0"
+depends: [ "${2:-}" ]
```
rgrinberg, line 59: How does the default value for `$2` help here?

Alizter: Thats the package that this one will depend on. We have 3 of them in the tests: package1->package2->package3. We are testing how setenv affects the env of the next package.

Alizter: Originally this was a 'EOF' so passing "" wouldn't be empty. I've removed it since so Ill remove the default value too.

test/blackbox-tests/test-cases/pkg/helpers.sh
```diff
@@ -52,3 +52,17 @@ cat >dune.lock/lock.dune <<EOF
 (lang package 0.1)
 EOF
 }
+
+# [mkenvtestpkg <package> <depends> <env_update_op>] makes an opam <package> that depends
+# on <depends> and updated "env_var" using <env_update_op>
+mkenvtestpkg() {
```
rgrinberg, line 58: This looks like a function that isn't useful for all tests. So how about:

1.You move all of these related tests into a sub directory
2. Introduce a shell script with only this function.

test/blackbox-tests/test-cases/pkg/env-update-prepend-sep.t
```diff
@@ -0,0 +1,34 @@
+We have 3 opam packages depending on each other in order. We test how setenv interacts
+with the := operator.
+
+  $ . ./helpers.sh
+  $ mkrepo
+
+  $ mkenvtestpkg package1 "" :=
+  $ mkenvtestpkg package2 package1 :=
+  $ mkenvtestpkg package3 package2 :=
```
rgrinberg, line 9: how about printing the relevant command after making the package? Otherwise, it's actually hard to understand what the test is doing.

Alizter: Printing the command trace is quite verbose, therefore I will just echo what it is doing in a readable form.

rgrinberg: Yeah, you don't need the entire command. Just the subset relevant for the test. So the line that updates the environment basically.

Alizter:
@rgrinberg I've moved the tests to their own directory and shared much more.

test/blackbox-tests/test-cases/pkg/env-update/env-update-helpers.sh
```diff
@@ -0,0 +1,38 @@
...
+mktestpkg() {
+  mkpkg $1 <<EOF
+opam-version: "2.0"
+depends: [ "$2" ]
+setenv: [
+  [ env_var $3 "$1" ]
+]
+build: [ "sh" "-c" "echo \"$1: \$env_var\"" ]
+EOF
+  echo "Created $1 depending on $2 updating env_var using $3"
```
rgrinberg, line 14: How about just printing the setenv line in this messasge?

test/blackbox-tests/test-cases/pkg/env-update.t/run.t
rgrinberg, line 1: What's the difference between this test and `opam-package-with-setenv`? Any reason why they can't be merged?

Alizter: This is explicitly testing the environment updates. I made them separate in the beginning in case I wanted to also add in build-env.

Alizter:
I've decided to keep the env updates in the set env test.

PR #8714, opened 2023-09-20
test(pkg): opam package with root variable

Alizter:
We give a better error message when the solver encounters a package with the `root` opam variable. We can't just fail during parsing as we need more information for a good error message.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -237,6 +237,20 @@ let opam_command_to_string_debug (args, _filter_opt) =
   |> String.concat ~sep:" "
 ;;
 
+let package_variable_of_opam_ident package ident =
+  let err_not_supported var package =
+    User_error.raise
+      [ Pp.textf
+          "Variable %S occuring in opam package %S is not supported."
+          var
+          (OpamPackage.to_string package)
+      ]
+  in
+  match Package_variable.pform_of_opam_ident ident with
+  | Ok pform -> pform
+  | Error `Root_unsupported -> err_not_supported "root" package
+;;
+
 let filter_to_blang package filter =
   let fident_to_blang packages variable string_converter =
     if Option.is_some string_converter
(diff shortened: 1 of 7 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -237,6 +237,20 @@ let opam_command_to_string_debug (args, _filter_opt) =
...
+let package_variable_of_opam_ident package ident =
+  let err_not_supported var package =
+    User_error.raise
+      [ Pp.textf
+          "Variable %S occuring in opam package %S is not supported."
+          var
+          (OpamPackage.to_string package)
+      ]
+  in
+  match Package_variable.pform_of_opam_ident ident with
```
rgrinberg, line 249: What about just moving the check for `"root"` here. That would allow us not to modify the type.

src/dune_pkg/opam_solver.ml
```diff
@@ -237,6 +237,20 @@ let opam_command_to_string_debug (args, _filter_opt) =
   |> String.concat ~sep:" "
 ;;
 
+let package_variable_of_opam_ident package ident =
+  let err_not_supported var package =
```
rgrinberg, line 241: Any reason why this is a function? I think it can just be inlined.

src/dune_pkg/opam_solver.ml
Alizter: I would have to pass the package name. Would that be acceptable?

src/dune_pkg/opam_solver.ml
Alizter: Sure I'll do that.

src/dune_pkg/opam_solver.ml
rgrinberg: I think so. My thinking is that this value of `Pform` can't be used correctly under any circumstances. So no need to allow people to build values from it.

Alizter:
This is subsumed by #8724

PR #8715, opened 2023-09-20
test(pkg): using %{root}% opam var in substs

Alizter:
This demonstrates an issue with using `%{root}%` in an ".in" file for a substs step.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/opam-package-subst-root-var.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-package-subst-root-var.t
@@ -1,0 +1,50 @@
+The root opam var occuring in a substition is not supported. We make sure that this has a
+good error message. This cannot happen during the solving stage, however we can catch it
+during building.
+
+  $ . ./helpers.sh
+  $ mkrepo
+
+Make a package with a substs field 
+  $ mkpkg with-substs-root <<EOF
+  > substs: ["foo.ml"]
+  > build: [ "echo" "foo.ml" ]
+  > EOF
+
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/pkg/opam-package-subst-root-var.t
```diff
@@ -0,0 +1,51 @@
...
+  > EOF
+
+The file foo.ml should have been built:
+
+  $ build_pkg with-substs-root 2>&1 | head -n6
+  Internal error, please report upstream including the contents of _build/log.
+  Description:
+    ("unable to serialize exception",
+    { exn =
+        "OpamSystem.File_not_found(\"$TESTCASE_ROOT/_build/.sandbox/32b841facd505f4493dd4a22d339f953/_private/default/.pkg/with-substs-root/source/foo.ml.in\")"
```
Alizter, line 50: This digest is probably not repro.

rgrinberg: Yes, that doesn't look repro. You can either push the fix in this PR or filter this out.

Alizter: I'll do this later when we have a better idea for a fix. This is really a bug about the .in file substitution but I won't be touching that any time soon.

gridbugs:
Do you happen to know for what reason a package would use the `%{root}%` variable in a substitution?

Alizter:
@gridbugs I don't have an example, but its not easy to find. The point of this test is that we are missing a validation step for aubsts.

PR #8717, opened 2023-09-20
fix(pkg): resolve lock dir packages in (package ..)

rgrinberg:
Resolve `(package ..)` by looking them up in the lock directory.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/pkg-deps.t
+++ b/test/blackbox-tests/test-cases/pkg/pkg-deps.t
@@ -8,7 +8,7 @@ We should be able to specify (package ..) deps on locally built packages.
 
   $ make_lockdir
   $ cat >dune.lock/foo.pkg <<EOF
-  > (build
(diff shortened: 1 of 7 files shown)


test/blackbox-tests/test-cases/pkg/pkg-deps.t
```diff
@@ -18,16 +18,12 @@ We should be able to specify (package ..) deps on locally built packages.
   > (dirs :standard \ external_sources)
   > (rule
   >  (alias foo)
-  >  (action (system "command -v foo"))
+  >  (action (which foo))
```
Alizter, line 21: Did you mean `run`?

Alizter: Does this get interpreted as a `run`? I wasn't aware `rule` was able to do that.

gridbugs (approved):

PR #8719, opened 2023-09-20
pkg: make --use-env-from-current-system hint a hint

Alizter:
This makes the solver failure error message less verbose.


gridbugs:
Awesome, I didn't know about the hints feature!

gridbugs (approved):

Issue #8676, opened 2023-09-17
dune fails to build on termux due to missing wait3/pthreads


seanbreckenridge:
## Expected Behavior

`opam install dune` installs `dune`

Perhaps related #7385?

## Actual Behavior

```
$ opam install dune
The following actions will be performed:
  ∗ install dune 3.10.0

<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[dune.3.10.0] found in cache

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[ERROR] The compilation of dune failed at
        "/data/data/com.termux/files/home/.opam/5.0.0/bin/ocaml boot/bootstrap.ml
        -j 1".

#=== ERROR while compiling dune.3.10.0 ========================================#
# context     2.0.4 | linux/arm64 | ocaml-base-compiler.5.0.0 | https://opam.ocaml.org#b61304c6
# path        ~/.opam/5.0.0/.opam-switch/build/dune.3.10.0
# command     ~/.opam/5.0.0/bin/ocaml boot/bootstrap.ml -j 1
# exit-code   2
# env-file    ~/.opam/log/dune-18226-d2c37b.env
# output-file ~/.opam/log/dune-18226-d2c37b.out
### output ###
# extern value caml_set_oo_id();
# [...]
# /data/data/com.termux/files/home/.opam/5.0.0/lib/ocaml/caml/mlvalues.h:423:18: note: conflicting prototype is here
# CAMLextern value caml_set_oo_id(value obj);
#                  ^
# 2 warnings generated.
# ./.duneboot.exe -j 1
# cd _boot && /data/data/com.termux/files/home/.opam/5.0.0/bin/ocamlopt.opt -o ../_boot/dune.exe -g -I +unix -I +threads unix.cmxa threads.cmxa -no-alias-deps -w -49-6 readdir.c wait3_stubs.c signal_stubs.c platform_stubs.c copyfile_stubs.c dune_digest_stubs.c xdg_stubs.c spawn_stubs.c dune_stats_stubs.c dune_flock.c csexp_rpc_stubs.c fsevents_stubs.c inotify_stubs.c fswatch_win_stubs.c sha512_[...]
# otherlibs/stdune/src/wait3_stubs.c:59:9: error: call to undeclared function 'wait3'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
[...]

PR #8720, opened 2023-09-21
fix: replace obsolete wait3 with wait4

purarue:
Related: https://github.com/ocaml/dune/issues/8676

`wait3` is the predecessor to `wait4`: https://www.gnu.org/software/libc/manual/html_node/BSD-Wait-Functions.html

When `-1` is passed as the first argument to `wait4`, it behaves the same as `wait3`

main reason for switching is that `wait3` is not declared on android, but `wait4` is

Testing on termux: `Linux localhost 4.19.269-g8728c337137c-ab10161573 #1 SMP PREEMPT Thu May 18 11:19:41 UTC 2023 aarch64 Android`

```
git clone https://github.com/seanbreckenridge/dune
cd dune
[ /data/data/com.termux/files/usr/tmp/Cx3VGdysNf/dune | main ] $ make release
ocamlc -output-complete-exe -w -24 -g -o .duneboot.exe -I boot -I +unix unix.cma boot/libs.ml boot/duneboot.ml
./.duneboot.exe
Done: 143/1349 (jobs: 8)cd _boot && /data/data/com.termux/files/home/.opam/5.1.0/bin/ocamlopt.opt -c -g -I +unix -I +threads wait3_stubs.c
otherlibs/stdune/src/wait3_stubs.c:59:9: error: call to undeclared function 'wait3'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  pid = wait3(&status, cv_flags, &ru);
        ^
1 error generated.
[ /data/data/com.termux/files/usr/tmp/Cx3VGdysNf/dune | main ] $ git checkout replace-wait3-wait4
Switched to branch 'replace-wait3-wait4'
[ /data/data/com.termux/files/usr/tmp/Cx3VGdysNf/dune | replace-wait3-wait4 ] $ git branch --set-upstream-to=origin/replace-wait3-wait4 replace-wait3-wait4 && git pull -q
branch 'replace-wait3-wait4' set up to track 'origin/replace-wait3-wait4'.
[ /data/data/com.termux/files/usr/tmp/Cx3VGdysNf/dune | replace-wait3-wait4 ] $ make release
ocamlc -output-complete-exe -w -24 -g -o .duneboot.exe -I boot -I +unix unix.cma boot/libs.ml boot/duneboot.ml
./.duneboot.exe
Done: 146/1349 (jobs: 8)cd _boot && /data/data/com.termux/files/home/.opam/5.1.0/bin/ocamlopt.opt -c -g -I +unix -I +threads spawn_stubs.c
vendor/spawn/src/spawn_stubs.c:706:3: error: call to undeclared function 'pthread_setcancelstate'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &cancel_state);
  ^
vendor/spawn/src/spawn_stubs.c:706:26: error: use of undeclared identifier 'PTHREAD_CANCEL_DISABLE'
```

Ends up with a different error (solved by https://github.com/janestreet/spawn/pull/52), but point is the 'undeclared function wait3' error is gone

No more references to `wait3`:

```
[ ~/Repos/dune | replace-wait3-wait4 ] $ rg 'wait3'    
[ ~/Repos/dune | replace-wait3-wait4 ] $
```

### DIFF
--- a/otherlibs/stdune/src/wait4_stubs.c
+++ b/otherlibs/stdune/src/wait4_stubs.c
@@ -3,9 +3,9 @@
 #ifdef _WIN32
 #include <caml/fail.h>
 
-void dune_wait3(value flags) {
+void dune_wait4(value flags) {
   (void)flags;
-  caml_failwith("wait3: not supported on windows");
+  caml_failwith("wait4: not supported on windows");
 }
 
 #else
@@ -45,7 +45,7 @@ static value alloc_process_status(int status) {
 
 static int wait_flag_table[] = {WNOHANG, WUNTRACED};
 
-value dune_wait3(value flags) {
+value dune_wait4(value flags) {
   CAMLparam1(flags);
   CAMLlocal2(times, res);
 
@@ -56,11 +56,11 @@ value dune_wait3(value flags) {
   struct rusage ru;
 
   caml_enter_blocking_section();
-  pid = wait3(&status, cv_flags, &ru);
+  pid = wait4(-1, &status, cv_flags, &ru);
   gettimeofday(&tp, NULL);
   caml_leave_blocking_section();
   if (pid == -1)
-    uerror("wait3", Nothing);
+    uerror("wait4", Nothing);
 
   times = caml_alloc_small(2 * Double_wosize, Double_array_tag);
   Store_double_field(times, 0, ru.ru_utime.tv_sec + ru.ru_utime.tv_usec / 1e6);
--- a/otherlibs/stdune/src/wait4_stubs.c
+++ b/otherlibs/stdune/src/wait4_stubs.c
@@ -3,7 +3,7 @@
 #ifdef _WIN32
 #include <caml/fail.h>
 
-void dune_wait4(value flags) {
+void dune_wait4(value v_pid, value flags) {
   (void)flags;
   caml_failwith("wait4: not supported on windows");
 }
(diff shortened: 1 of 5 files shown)


otherlibs/stdune/src/wait4_stubs.c
```diff
@@ -56,11 +56,11 @@ value dune_wait3(value flags) {
   struct rusage ru;
 
   caml_enter_blocking_section();
-  pid = wait3(&status, cv_flags, &ru);
+  pid = wait4(-1, &status, cv_flags, &ru);
```
rgrinberg, line 59: Instead of `-1,` how about adding this as an argument to `dune_wait4` and propagating it everywhere?

seanbreckenridge: Ah, hm, I tried using the `Pid.t`, I think defined [here](https://github.com/ocaml/dune/blob/b326c30331b5fdb039e06b02b88bdff96bf17da3/otherlibs/stdune/src/pid.ml#L8-L11), but when converting from `Pid.of_int -1` (`-1` being any child process), this would end up failing. I guess two options are, change the implementation of the `pid.ml` file be changed to allow `-1`, or change the type signatures to accept a `int` instead?

src/dune_engine/scheduler.ml
```diff
@@ -597,7 +597,8 @@ end = struct
 
   let wait_unix t =
     Mutex.unlock t.mutex;
-    let proc_info = Proc.wait [] in
+    (* -1 means wait for any child process *)
```
seanbreckenridge, line 600: I dont think `wait_unix`s function signature should be changed to accept the `pid`, right? Not sure if the inline `(Pid.of_int (-1))` is fine here, am still relatively new to ocaml. let me know if that should be declared above and then re-used here or something

otherlibs/stdune/src/wait4_stubs.c
```diff
@@ -3,9 +3,9 @@
 #ifdef _WIN32
 #include <caml/fail.h>
 
-void dune_wait3(value flags) {
+void dune_wait4(value flags) {
```
Alizter, line 6: This signature here needs updating to take the extra argument.

rgrinberg:
cc @pmwhite if you'd like to pull this change.

pmwhite:
This has been pulled.

PR #8721, opened 2023-09-21
change: never ignore promote until-clean under -p

emillon:
This extends #5956 to all versions of `(lang dune)`, since this is a property of the CLI.

### DIFF
--- a/doc/changes/promote-fallback-always.md
+++ b/doc/changes/promote-fallback-always.md
@@ -1,0 +1,3 @@
+- If `--ignore-promoted-rules` is set, do not change `(promote (until-clean))`
+  rules and make affected rules fallback instead of removing them. This was
+  previously done depending on `(lang dune)`. (#8721, @emillon)
--- a/promote-until-clean-always.md
+++ b/promote-until-clean-always.md
@@ -1,3 +1,0 @@
-- Never ignore `(promote (until-clean))` if `--ignore-promoted-rules` is
-  passed. This was previously done depending on `(lang dune)`. (#8721,
-  @emillon)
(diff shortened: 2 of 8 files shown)


rgrinberg:
Isn't this exactly what I've tried here? https://github.com/ocaml/dune/pull/8512 It didn't work because it indeed broke stuff downstream.

Instead of filtering, I thought we agreed to turn them into fallback instead.

emillon:
Ah yes I was not sure if you meant to make these rules fallback too. I just updated the PR but there's a test failing:

```
;; More complex test

(rule
 (target into+ignoring)
 (mode (promote (into subdir)))
 (action (with-stdout-to %{target} (echo "hello"))))

(rule
 (target into+ignoring)
 (enabled_if %{ignoring_promoted_rules})
 (action (copy subdir/into+ignoring into+ignoring)))
```

I think it's invalid now, but it's probably a smaller repro of some existing package. WDYT?

rgrinberg:
> I think it's invalid now, but it's probably a smaller repro of some existing package. WDYT?

Isn't this basically a manually created fallback rule?

rgrinberg:
In any case, given that this variable exists (though it seems to be used) I feel like this change is big enough that perhaps we should save this all until 4.0?

emillon:
Agree. Let's keep #8706 though for 3.11.

rgrinberg:
What's the rush to get it out for 3.11? The bug fix is basically incomplete so this fix helps nobody still. Moreover, it introduces a confusing behavior

emillon:
What about generated opam files that now fail because of #8518? Do we call that an intended breakage?

rgrinberg:
I'm reverting that as well.

PR #8724, opened 2023-09-21
test(pkg): opam variable tests

Alizter:
We try to exhaustively test all the possible opam variables. We attempt to translate a dummy opam package that echos the variable into a dune lock file. Where it makes sense we also try to build this dummy package.

We also fix the incorrect macro expansion of package variables and remove the unsupported misc; in addition giving better error messages to all unsupported variables.

### DIFF
--- a/src/dune_pkg/package_variable.ml
+++ b/src/dune_pkg/package_variable.ml
@@ -57,26 +57,40 @@ let to_macro_invocation { name; scope } =
   | Package package_name ->
     { Pform.Macro_invocation.macro = Pkg
     ; payload =
-        Pform.Payload.of_args [ Name.to_string name; Package_name.to_string package_name ]
+        Pform.Payload.of_args [ Package_name.to_string package_name; Name.to_string name ]
     }
 ;;
 
 let to_pform t = Pform.Macro (to_macro_invocation t)
 
-let of_opam_ident ident =
-  match String.lsplit2 ident ~on:':' with
-  | Some ("_", variable) -> `Package_variable (self_scoped (Name.of_string variable))
-  | Some (package, variable) ->
+let package_variable package variable =
+  match package, variable with
+  | _, ("root" | "hash" | "build-id" | "misc" | "opam-version") ->
+    `Unsupported_variable variable
+  | "_", variable -> `Package_variable (self_scoped (Name.of_string variable))
+  | package, variable ->
     `Package_variable
       (package_scoped (Name.of_string variable) (Package_name.of_string package))
+;;
+
+let of_opam_ident ident =
+  match String.lsplit2 ident ~on:':' with
+  | Some (package, variable) -> package_variable package variable
   | None ->
     (match Pform.Var.of_opam_global_variable_name ident with
      | Some var -> `Global_variable var
-     | None -> `Package_variable (self_scoped (Name.of_string ident)))
+     | None -> package_variable "_" ident)
 ;;
 
-let pform_of_opam_ident ident =
+let pform_of_opam_ident ~package_name ident =
   match of_opam_ident ident with
   | `Package_variable t -> to_pform t
   | `Global_variable var -> Pform.Var var
+  | `Unsupported_variable name ->
+    User_error.raise
+      [ Pp.textf
+          "Variable %S occuring in opam package %S is not supported."
+          name
+          package_name
+      ]
 ;;
--- a/test/blackbox-tests/test-cases/pkg/opam-var/dune
+++ b/test/blackbox-tests/test-cases/pkg/opam-var/dune
@@ -1,0 +1,25 @@
+(cram
+ (deps opam-var-helpers.sh)
+ (applies_to :whole_subtree))
+
+(cram
+ (deps opam-vars)
+ (applies_to opam-var-os))
+
+(data_only_dirs fake_opam_root)
+
+; We trick opam into thinking it has a config file with fake_opam_root so that we can
+; access the os variables without having to do opam init
+
+(rule
+ (target opam-vars)
+ (deps fake_opam_root/config)
+ (action
+  (with-stdout-to
+   %{target}
+   (progn
+    (run %{bin:opam} var --root fake_opam_root arch)
+    (run %{bin:opam} var --root fake_opam_root os)
+    (run %{bin:opam} var --root fake_opam_root os-distribution)
+    (run %{bin:opam} var --root fake_opam_root os-family)
+    (run %{bin:opam} var --root fake_opam_root os-version)))))
--- a/test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
@@ -1,0 +1,56 @@
+  $ . ./opam-var-helpers.sh
+
+Here we test the translation and implementation of global opam variables. OS specific
+variables can be found in `opam-var-os.t`.
+
+  $ mkrepo
+  > mkpkg testpkg << EOF
+  > opam-version: "2.0"
+  > build: [
+  >   [ "echo" jobs ]
+  >   [ "echo" make ]
+  >   [ "echo" user ] 
+  >   [ "echo" group ]
+  > ]
+  > EOF
+  > solve testpkg
+  Solution for dune.lock:
+  testpkg.0.0.1
+  
+  $ cat dune.lock/testpkg.pkg
+  (version 0.0.1)
+  
+  (build
+   (progn
+    (run echo %{jobs})
+    (run echo %{make})
+    (run echo %{user})
+    (run echo %{group})))
+
+The implementation of %{user} uses Unix.getlogin which doesn't work in our Linux CI job.
+Therefore we modify the lockfile here to remove that from the opam file:
+
+  $ mkpkg testpkg << EOF
+  > opam-version: "2.0"
+  > build: [
+  >   [ "echo" jobs ]
+  >   [ "echo" make ]
+  >   [ "echo" group ]
+  > ]
+  > EOF
+  > solve testpkg
+  Solution for dune.lock:
+  testpkg.0.0.1
+  
+If this is fixed, we should use $(whoami) to compare the output.
+
+The value for "jobs" should always be 1.
+
+  $ MAKE="$(which make)"
+  > GROUP="$(id -gn)"
+  > build_pkg testpkg 2>&1 \
+  > | sed "s/$GROUP/GROUP/g" | sed "s#$MAKE#MAKE#g"
+  File "dune.lock/testpkg.pkg", line 5, characters 12-19:
+  5 |   (run echo %{jobs})
+                  ^^^^^^^
+  Error: Unknown variable %{jobs}
--- a/test/blackbox-tests/test-cases/pkg/opam-var/opam-var-os.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-var/opam-var-os.t
@@ -1,0 +1,45 @@
+  $ . ./opam-var-helpers.sh
+
+Here we test global opam variables that are system specific. Since these values change
+between systems, we can't hardcode them in the test. Instead, we use the opam var command
+to compare their values.
+
+# arch os os-distribution os-family os-version user group
+
+
+  $ mkrepo
+  > mkpkg testpkg <<EOF
+  > opam-version: "2.0"
+  > build: [
+  >   ["echo" arch]
+  >   ["echo" os]
+  >   ["echo" os-distribution]
+  >   ["echo" os-family]
+  >   ["echo" os-version]
+  > ]
+  > EOF
+  > solve testpkg
+  Solution for dune.lock:
+  testpkg.0.0.1
+  
+  $ cat dune.lock/testpkg.pkg 
+  (version 0.0.1)
+  
+  (build
+   (progn
+    (run echo %{arch})
+    (run echo %{pkg-self:os})
+    (run echo %{os_distribution})
+    (run echo %{os_family})
+    (run echo %{os_version})))
+
+We write all the dune values to a file and then diff them with the output of opam var.
+
+  $ build_pkg testpkg 2> dune-vars
+  [1]
+
+The two files should be identical.
+
+  $ diff -q --label="opam-vars"  opam-vars --label="dune-vars" dune-vars 
+  Files opam-vars and dune-vars differ
+  [1]
--- a/test/blackbox-tests/test-cases/pkg/opam-var/opam-var-pkg.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-var/opam-var-pkg.t
@@ -1,0 +1,227 @@
+  $ . ./opam-var-helpers.sh
+
+Here we test the translation and implementation of opam package variables.
+
+We echo each package variable. 
+
+  $ mkrepo
+  > mkpkg "testpkg" <<'EOF' 
+  > opam-version: "2.0"
+  > depends: [ "foo" ]
+  > build: [
+  >   [ "echo" "1"     name ]
+  >   [ "echo" "2"   _:name ]
+  >   [ "echo" "3" foo:name ]
+  >   [ "echo" "4"     version ]
+  >   [ "echo" "5"   _:version ]
+  >   [ "echo" "6" foo:version ]
+  >   [ "echo" "7"   _:depends ]
+  >   [ "echo" "8" foo:depends ]
+  >   [ "echo" "9"    _:installed ]
+  >   [ "echo" "10" foo:installed ]
+  >   [ "echo" "11"   _:enable ]
+  >   [ "echo" "12" foo:enable ]
+  >   [ "echo" "13"   _:pinned ]
+  >   [ "echo" "14" foo:pinned ]
+  >   [ "echo" "15"   _:bin ]
+  >   [ "echo" "16" foo:bin ]
+  >   [ "echo" "17"   _:sbin ]
+  >   [ "echo" "18" foo:sbin ]
+  >   [ "echo" "19"   _:lib ]
+  >   [ "echo" "20" foo:lib ]
+  >   [ "echo" "21"   _:man ]
+  >   [ "echo" "22" foo:man ]
+  >   [ "echo" "23"   _:doc ]
+  >   [ "echo" "24" foo:doc ]
+  >   [ "echo" "25"   _:share ]
+  >   [ "echo" "26" foo:share ]
+  >   [ "echo" "27"   _:etc ]
+  >   [ "echo" "28" foo:etc ]
+  >   [ "echo" "29"   _:build ]
+  >   [ "echo" "30" foo:build ]
+  >   [ "echo" "31"   _:dev ]
+  >   [ "echo" "32" foo:dev ]
+  >   [ "echo" "33"   _:opamfile ]
+  >   [ "echo" "34" foo:opamfile ]
+  >   [ "echo" "35"     with-test ]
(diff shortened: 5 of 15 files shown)


test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
```diff
@@ -0,0 +1,138 @@
...
+os-version
+  $ test os-version
+  (build
+   (run echo %{os_version}))
+  File "dune.lock/testpkg.pkg", line 4, characters 11-24:
+  4 |  (run echo %{os_version}))
+                 ^^^^^^^^^^^^^
+  Error: Unknown variable %{os_version}
+  [1]
+switch
```
rgrinberg, line 72: All of these path variables can be combined into a single test. For example, the output could be:

```
%{switch}: dune
%{prefix}: ../target
%{lib}: ../target/lib
...
```

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-pkg.t
```diff
@@ -0,0 +1,232 @@
...
+_:build
+  $ test _:build
+  (build
+   (run echo %{pkg-self:build}))
+  File "dune.lock/testpkg.pkg", line 4, characters 11-28:
+  4 |  (run echo %{pkg-self:build}))
+                 ^^^^^^^^^^^^^^^^^
+  Error: Unknown macro %{pkg-self:..}
+  [1]
+_:hash
```
rgrinberg, line 149: All explicitly unsupported variables like this one can be moved to their own test file

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
Alizter: Done. Note that %{switch} isn't a directory but rather the name of the current "switch".

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-pkg.t
Alizter: The `pkg-self` macro not working doesn't tell me if we are not intending to support it. So I'm not sure exactly which ones we are intending to support. Is there a list somewhere?

Alizter:
I've generalised the test script to take multiple arguments allowing us to do certain tests at once.

Alizter:
cc @gridbugs @Leonidas-from-XIV take a look and let me know if the outputs are as you expect.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-pkg.t
Alizter: I've moved `root` and `_:hash` to an unsupported variables file anyway.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
```diff
@@ -0,0 +1,107 @@
...
+  4 |  (run echo %{os_distribution}))
+                 ^^^^^^^^^^^^^^^^^^
+  Error: Unknown variable %{os_distribution}
+  [1]
+os-family
+  $ test os-fmaily
+  (build
+   (run echo %{pkg-self:os-fmaily}))
+  File "dune.lock/testpkg.pkg", line 4, characters 11-32:
+  4 |  (run echo %{pkg-self:os-fmaily}))
```
rgrinberg, line 54: `os-family`

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
```diff
@@ -0,0 +1,107 @@
...
+  4 |  (run echo %{pkg-self:os}))
+                 ^^^^^^^^^^^^^^
+  Error: Unknown macro %{pkg-self:..}
+  [1]
+os-distribution
+  $ test os-distribution
+  (build
+   (run echo %{os_distribution}))
+  File "dune.lock/testpkg.pkg", line 4, characters 11-29:
+  4 |  (run echo %{os_distribution}))
```
rgrinberg, line 45: `os-distribution`

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
```diff
@@ -0,0 +1,107 @@
...
+  4 |  (run echo %{pkg-self:os-fmaily}))
+                 ^^^^^^^^^^^^^^^^^^^^^
+  Error: Unknown macro %{pkg-self:..}
+  [1]
+os-version
+  $ test os-version
+  (build
+   (run echo %{os_version}))
+  File "dune.lock/testpkg.pkg", line 4, characters 11-24:
+  4 |  (run echo %{os_version}))
```
rgrinberg, line 63: `os-version`

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
```diff
@@ -0,0 +1,107 @@
...
+  ../target/bin
+  ../target/sbin
+  ../target/share
+  ../target/doc
+  ../target/etc
+  ../target/man
+  ../target/lib/toplevel
+  ../target/lib/stublibs
+user
+  $ current_user=$(whoami)
```
rgrinberg, line 97: `"$(whoami)"`

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
```diff
@@ -0,0 +1,107 @@
...
+  $ test user 2>&1 | sed "s/$current_user/current_user/g"
+  (build
+   (run echo %{user}))
+  current_user
+group
+  $ export current_user_primary_group=$(id -g -n)
```
rgrinberg, line 103: `"$(id -g -n)"`

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
Alizter: Do you want me to fix how these are translated? The correct opam variable is being tested, its just our translation that is wrong.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
```diff
@@ -0,0 +1,102 @@
...
+  (build
+   (run echo %{os-distribution}))
+  
+os-family
+TODO: seems to be broken (on NixOS)
+  $ test os-family
+  (build
+   (run echo %{os-family}))
+os-version
+TODO: seems to be broken (and hard to make x-platform) (on NixOS returns Nix verison?!)
```
Alizter, line 49: This is broken.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
```diff
@@ -0,0 +1,102 @@
...
+   (run echo %{arch}))
+  ARCH
+os
+  $ OS=$(uname -s | awk '{print tolower($0)}')
+  > test os 2>&1 | sed "s/^$OS/OS/g"
+  (build
+   (run echo %{os}))
+  OS
+os-distribution
+TODO: seems to be broken (on NixOS)
```
Alizter, line 38: This is broken.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
```diff
@@ -0,0 +1,102 @@
...
+  $ test os-distribution
+  (build
+   (run echo %{os-distribution}))
+  
+os-family
```
Alizter, line 43: This is broken.

Alizter (commented):
Some issues with Sys_poll.

gridbugs (approved):

rgrinberg:
> TODO (might be NixOS issues)

You should confirm that these are not issues in opam itself. Our goal with these variables is to reproduce opam's behavior faithfully. If opam itself is buggy, there's not much we can do about it.

src/dune_pkg/package_variable.ml
```diff
@@ -63,20 +63,33 @@ let to_macro_invocation { name; scope } =
 
 let to_pform t = Pform.Macro (to_macro_invocation t)
 
-let of_opam_ident ident =
-  match String.lsplit2 ident ~on:':' with
-  | Some ("_", variable) -> `Package_variable (self_scoped (Name.of_string variable))
-  | Some (package, variable) ->
+let package_variable package variable =
+  match package, variable with
+  | _, ("root" | "hash" | "build-id" | "misc") -> `Unsupported_variable variable
```
rgrinberg, line 68: So we're only excluding unsupported variables inside the translation, but not inside the substitution, right?

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
rgrinberg: You should describe the broken behavior on NixOS inside the test.

rgrinberg:
All the `%{os*}` family of tests are very similar. The following would be a good way to handle them:

* Stop printing their values. The values are going to be completely unportable and making them portable requires re-implementing their functionality in shell scripts. As long as they resolve this is going to be enough.
* Group them all into a single package

src/dune_pkg/package_variable.ml
Alizter: I think so. That's something #8715 shows. I'll reconfirm after this PR is merged.

rgrinberg:
Thinking about this again, we should probably translate the variable names to dune's convention. So we want `%{os_family}` rather than `%{os-family}`.

Alizter:
@rgrinberg Actually, if our goal is to be on par with opam, then maybe we can just compare the output of `opam var VAR` with the variable we want in the tests?

rgrinberg:
Using `opam var` to verify the results would be fine. Although you would need to initialize opam inside the tests

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
```diff
@@ -0,0 +1,102 @@
+  $ . ./opam-var-helpers.sh
+
+Here we test the translation and implementation of global opam variables.
+This also includes switch specific variables since these are all trivial for Dune.
+opam-version
+  $ test opam-version
+  (build
+   (run echo %{pkg-self:opam-version}))
+  File "dune.lock/testpkg.pkg", line 4, characters 11-35:
+  4 |  (run echo %{pkg-self:opam-version}))
```
rgrinberg, line 10: opam-version isn't a package variable

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
```diff
@@ -0,0 +1,24 @@
...
+Here we test the translation and implementation of global opam variables. OS specific
+variables can be found in `opam-var-os.t`.
+
+# jobs
+Should always be 1
+  $ test jobs 
+  (build
+   (run echo %{jobs}))
+  File "dune.lock/testpkg.pkg", line 4, characters 11-18:
+  4 |  (run echo %{jobs}))
```
Alizter, line 12: I need to investigate this some more, but for some reason this won't evaluate.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-os.t
```diff
@@ -0,0 +1,27 @@
...
+  > $(opam var group)
+  > EOF
+ 
+  $ test arch os os-distribution os-family os-version user group >/dev/null 2> dune-vars 
+  [1]
+
+The two files should be identical.
+  $ diff -q -U0 --label="opam"  opam-vars --label="dune" dune-vars 
+  Files opam and dune differ
+  [1]
```
Alizter, line 27: This will take some more time to fix as our syspoll isn't agreeing with opam on my NixOS machine. I also want to try running this tests on some other platforms.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-unsupported.t
```diff
@@ -0,0 +1,30 @@
+  $ . ./opam-var-helpers.sh
+
+opam variables that are explicitly unsupported in dune
+
+These should all have nice error messages explaining that they are not supported.
+
+# opam-version
+  $ test_lockfile opam-version
+  (build
+   (run echo %{pkg-self:opam-version}))
```
Alizter, line 10: Do we want to explicitly not support this or give it a dummy value? grepping the opam repo doesn't appear to show it actually being used other than being set at the top of the opam file (which we don't care about).

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-os.t
```diff
@@ -0,0 +1,19 @@
...
+Here we test global opam variables that are system specific. Since these values change
+between systems, we can't hardcode them in the test. Instead, we use the opam var command
+to compare their values.
+
+# arch os os-distribution os-family os-version user group
+
+We write all the dune values to a file and then diff them with the output of opam var.
+
+  $ test arch os os-distribution os-family os-version >/dev/null 2> dune-vars 
+  [1]
```
Alizter, line 12: This fails for the moment as the os variables don't all work during the translation step. This will be fixed by #8738.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
```diff
@@ -0,0 +1,35 @@
...
+# make
+There is a difference here as dune returns the absolute directory to make whereas
+opam returns `make`.
+  $ BUILD_PATH_PREFIX_MAP="MAKE_DIR=$(dirname $(which make)):$BUILD_PATH_PREFIX_MAP"
+  > test make 
+  (build
+   (run echo %{make}))
+  MAKE_DIR/make
+
+# user group
```
Alizter, line 26: This is actually tied to the switch config and is created from scratch for every new opam switch. This makes it tricky to get from our `opam var` trick that we used for os therefore we use standard posix methods of getting this information.

test/blackbox-tests/test-cases/pkg/opam-var/dune
```diff
@@ -0,0 +1,25 @@
...
+; We trick opam into thinking it has a config file with fake_opam_root so that we can
+; access the os variables without having to do opam init
+
+(rule
+ (target opam-vars)
+ (deps fake_opam_root/config)
+ (action
+  (with-stdout-to
+   %{target}
+   (progn
```
Alizter, line 20: I experimented with making this `concurrent` and then sorting however `opam var` locks the directory so concurrent queries error. We would also had to bump our dune project version.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-global.t
Alizter: Not going to fix here.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-os.t
Alizter: Not going to fix in this PR.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-pkg.t
```diff
@@ -0,0 +1,214 @@
+  $ . ./opam-var-helpers.sh
+
+Here we test the translation and implementation of opam package variables.
```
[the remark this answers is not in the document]
Alizter: Done.

Alizter:
The tests run fine on NixOS and on macos. I can't work out why they run differently in the CI.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-pkg.t
```diff
@@ -0,0 +1,71 @@
...
+
+name _:name version _:version _:depends _:installed _:enable _:pinned _:bin _:sbin _:lib
+_:man _:doc _:share _:etc _:build _:dev _:opamfile with-test _:with-test with-doc
+_:with-doc _:with-dev-setup
+
+  $ test_lockfile name _:name version _:version _:depends _:installed _:enable _:pinned \
```
rgrinberg, line 9: Could you write the package out instead of using this function?

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-switch.t
```diff
@@ -0,0 +1,41 @@
...
+there to be a single switch named "dune" and all the installation locations should be in
+the package building directories.
+
+# switch
+  $ test switch
+  (build
+   (run echo %{switch}))
+  dune
+# build prefix lib libexec bin sbin share doc etc man toplevel stublibs
+  $ test build prefix lib libexec bin sbin share doc etc man toplevel stublibs
```
rgrinberg, line 14: Again you can just write this stuff out

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-switch.t
```diff
@@ -0,0 +1,41 @@
+  $ . ./opam-var-helpers.sh
+
+These opam variables are known as "switch variables" in opam, but since in Dune we don't
+have switches, we consider them glboal variables. To keep inline with opam we consider
+there to be a single switch named "dune" and all the installation locations should be in
+the package building directories.
+
+# switch
+  $ test switch
```
rgrinberg, line 9: ditto

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-unsupported.t
rgrinberg: If nobody is using it, we don't need to support it. I don't see of a reasonable way of using this variable.

Alizter:
@rgrinberg There are now two commits here. The first commit is the original PR which consists of just tests, however there are some obviously in incorrect tests for the `opam-var-pkg` as the order of the macro has been muddled. This causes some `assert false`s to trigger leading to ugly backtraces in the code.

I've therefore included a second commit which is the fix for the order of the maco. I can separate the second PR into anther PR if you really want, however I don't really want to spend effort hiding a backtrace.

test/blackbox-tests/test-cases/pkg/opam-var/opam-var-pkg.t
```diff
@@ -0,0 +1,227 @@
...
+    (run echo 39 %{pkg-self:with-doc})
+    (run echo 40 %{pkg:with-doc:foo})
+    (run echo 41 %{pkg-self:with-dev-setup})
+    (run echo 42 %{pkg:with-dev-setup:foo})))
+  
+  (deps foo)
+
+The values here are not important, but Dune should be able to interpret the variables.
+
+  $ build_pkg testpkg
```
Alizter, line 119: This will be fixed after this PR. The order of the macro is wrong.

rgrinberg: Could you either:

1. filter out the backtrace
2. push the fix as a separate commit so there's no error.

Otherwise, this will make every other PR that touches pkg_rules break.

rgrinberg:
Okay, thanks. The tests are much clearer and I can follow along easily now.

The backtrace needs to be hidden or fixed though. Otherwise it's too much of a disturbance to everyone else.

Alizter:
@rgrinberg I've hidden `user` due to the buggy nature of `Unix.getlogin` on the Linux CI. I've also pushed the commit fixing the package variables which fixes the backtrace from being displayed.

PR #8725, opened 2023-09-21
refactor: simplify left over type signature

rgrinberg:


rgrinberg:
actually, this fix is really not good enough for 3.11

PR #8726, opened 2023-09-21
Revert "fix: make ignored rules fallback (#8706)"

rgrinberg:
@emillon Let's wait for a proper fix.


emillon:
Thanks, that's safer like this.

PR #8728, opened 2023-09-22
Warn if a descendant of source_tree is ignored

gridbugs:
This prints a warning if any directory which is a descendant of a rule target specified with `(source_tree ...)` is not copied to the build directory.


src/dune_rules/dep_conf_eval.ml
```diff
@@ -162,6 +162,46 @@ let rec dep expander : Dep_conf.t -> _ = function
...
+           List.exists
+             descendant_file_paths_in_build_dir
+             ~f:(fun descendant_file_path_in_build_dir ->
+               Path.Source.is_descendant
+                 descendant_file_path_in_build_dir
+                 ~of_:descendant_dir_path)
+         in
+         if not at_least_one_build_path_descends_from_source_path
+         then
+           User_warning.emit
```
rgrinberg, line 185: We have a warning module to make sure we don't emit warnings that can't be turned off.

gridbugs: Ah of course, will update!

rgrinberg:
Please re-open once if you resume work on this PR.

PR #8732, opened 2023-09-22
feature(memo): introduce variables

rgrinberg:
cc @snowleopard and @staronj who reviewed/authored this change internally


snowleopard (approved):
Thanks! It would be nice to add tests, so we can improve the current implementation further. Also looking forward to switching to using variables (or tables of variables?) in Fs_memo.

Alizter:
What's the use case for this? Debugging?

rgrinberg:
Added some simple tests.

It's useful to replace mutable variables.

PR #8738, opened 2023-09-22
fix(pkg): os variables

Alizter:
- Added missing `os` opam variable.
- fixed some incorrect names

### DIFF
--- a/src/dune_pkg/sys_poll.ml
+++ b/src/dune_pkg/sys_poll.ml
@@ -102,7 +102,7 @@ let os_release_field field =
         | Error _ -> None
         | Ok (key, v) ->
           (match Scanf.sscanf v "\"%s@\"" Fun.id with
-           | Error _ -> None
+           | Error _ -> Some (key, v)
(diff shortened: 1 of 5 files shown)


src/dune_pkg/sys_poll.ml
```diff
@@ -102,7 +102,7 @@ let os_release_field field =
         | Error _ -> None
         | Ok (key, v) ->
           (match Scanf.sscanf v "\"%s@\"" Fun.id with
-           | Error _ -> None
+           | Error _ -> Some (key, v)
```
Alizter, line 105: @Leonidas-from-XIV This was mistakenly set to none meaning that fields with values not wrapped with `""` were being ignored. I had a look at the opam code and changed it to match. On my system, `ID=nixos` was being ignored since it didn't have quotes around the name. Now it should pick up such fields fine.

rgrinberg:
Probably the correct way to test this is with unit tests and mock `uname` output

PR #8749, opened 2023-09-23
fix(pkg): jobs var

Alizter:
We fix the expansion of the jobs var.

### DIFF
--- a/src/dune_lang/pform.ml
+++ b/src/dune_lang/pform.ml
@@ -571,6 +571,7 @@ module Env = struct
         ; "user", User
         ; "group", Group
         ; "arch", Arch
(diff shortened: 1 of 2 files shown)


Alizter:
The CI is failing because occasionally our Linux runners have gmake installed causing it to be preferred over make.

This highlights a bug that we have and I have created a test for it in #8750. I also have a fix ready. If I rerun the CI it will most likely go away.

rgrinberg:
@Alizter what's going on with this PR?

Alizter:
@rgrinberg I've fixed it. It was waiting for the removal of the user var which has now happened.

rgrinberg:
Thanks.

PR #8750, opened 2023-09-24
test(pkg): show precedence of make and gmake

Alizter:
This test demonstrates that we currently prefer `gmake` over `make` when resolving `%{make}`.

opam won't resolve make in the PATH but rather just use `make` literally. This means that `make` will always be preferred over `gmake`.

Typically if `make` and `gmake` are both present, `make` is just a symlink to `gmake` anyway.

---

#### Idea how to fix:

The correct thing to do here IMO would be to replicate opams behaviour and choose `make` over `gmake` in this situation. This test would therefore be update to prefer `make` at the end.

Here is how we choose make in our implementation:
```ocaml
let which loc context ~path =
  (let which = Which.which ~path in
   match Sys.unix with
   | false -> which "make"
   | true ->
     which "gmake"
     >>= (function
     | None -> which "make"
     | Some _ as s -> Memo.return s))
  >>| function
  | Some p -> p
  | None -> Utils.program_not_found ~context ~loc:(Some loc) "make"
;;
```

If we are on a unix-like system we immediately prefer `gmake`. We should therefore swap the order and prefer `make` over `gmake` using `gmake` as a fallback instead.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/opam-var/make-over-gmake.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-var/make-over-gmake.t
@@ -1,0 +1,24 @@
+  $ . ./opam-var-helpers.sh 
+
+If both make and gmake are available, we should prefer make.
+
+opam doesn't resolve make but rather runs make literally so it is up to whatever is in
+PATH to resolve it. This does mean that make is preferred over gmake so we should
+replicate that behaviour.
+
+We create a dummy package that will output the value of the opam make variable:
+  $ mkrepo
+  > mkpkg testpkg << EOF
+  > opam-version: "2.0"
+  > build: [ "echo" make ]
+  > EOF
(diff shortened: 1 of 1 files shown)


Alizter:
@rgrinberg The reason I noticed this at all is because some of our CI runners have gmake present which makes our make detection tests unrepro. I could try to make the tests also prefer gmake but I thought this would be better since it is closer to what opam is doing.

rgrinberg:
Modifying the tests to accommodate this is the right way to go at this time. Changing the meaning of `%{make}` in the rules is not something to done lightly.

Just adding `gmake` to the PATH alias to `make` should be simple enough

Alizter:
OK, I'll update the test to be more resilient. There is no point having this test anymore so closing.

rgrinberg:
This test is still useful I think. It would be helpful to note the difference in behavior in the test suite.

PR #8752, opened 2023-09-25
fix(nix): fix utop tests

Alizter:
Nix was compressing the man pages for utop which meant the dune-package file was pointing to non-existant files. Overriding utop and disabling the compression of man pages fixes this issue.

We can also use this fix for other `(package)` dependencies that are problematic with nix.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -76,7 +76,11 @@
         let
           pkgs = nixpkgs.legacyPackages.${system}.appendOverlays [
             (self: super: {
-              ocamlPackages = self.ocaml-ng.ocamlPackages_4_14;
+              ocamlPackages = self.ocaml-ng.ocamlPackages_4_14.overrideScope' (oself: osuper: {
+                utop = osuper.utop.overrideAttrs {
+                  dontGzipMan = true;
+                };
+              });
             })
             melange.overlays.default
           ];


flake.nix
```diff
@@ -221,13 +225,24 @@
                 mdx
                 odoc
                 ppx_expect
+
```
rgrinberg, line 228: did this sneak in accidentally?

rgrinberg:
It seems like a decent workaround. The problem with it is that it requires every single package with manpages that is depended on with `(package ..)` to be modified this way.

I don't know enough about nix, but surely there's other instances it tries to modify build artifacts and has to modify some metadata as well. Can we make nix modify the dune-package file to use the gz filename?

Alizter:
@rgrinberg That should be possible. I'll ask around.

We can make the overlays a bit easier to work with by separating them out of the flake. It's already gotten pretty big.

anmonteiro:
Probably easy to change `buildDunePackage` itself. that's what we did for `--man-dir` and other stuff before.

Alizter:
@anmonteiro Yes, I've spoken to Ulrik about supporting this in buildDunePackage. For the moment, this PR solves our own issues with the test suite.

anmonteiro (approved):
makes sense to do it here while it's not fixed upstream

PR #8758, opened 2023-09-25
test(pkg): build package dep in release mode

Alizter:


test/blackbox-tests/test-cases/pkg/ignore-lock-release.t
```diff
@@ -0,0 +1,21 @@
...
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.11)
+  > (package
+  >  (name foo)
+  >  (allow_empty)
+  >  (depends test))
+  > EOF
+
+  $ dune build @install --release 2>&1 | tail -n1
```
Alizter, line 20: We use tail because `--release` implies `--always-show-command-line` which makes this test noisier.

PR #8759, opened 2023-09-25
Relative paths with pkg-config

frejsoya:
This slightly, but not much, improves tests wrt. to pkg-config as it avoids rewriting .pc files awk.

Relies on dune 3.8 and passing --define-prefix through PKG_CONFIG_ARGN to pkg-config. The prefix defined will be the grandparent dir of .pc file. IE. /foo/bar/libdir/pkgconfig/libexample.pc -> prefix=/foo. The location of .pc file must be in a dir named 'pkgconfig'


rgrinberg:
LGTM. Could you add a DCO by signing off your commit and revert the globs that CI doesn't like?

PR #8761, opened 2023-09-25
fix(pkg): dune.lock is ignored in --release

### LINKED ISSUE #8684: Ignore Lock Directory in Release Mode
Projects that are going to use our dependency locking mechanism will likely commit the lock directory. This means that it will be included in the package sources when building and install via opam. To make sure we don't break such packages, we can ignore the lock directory by default. This will be implied by `--release`

Alizter:
We add a new --ignore-lock-directory option that will always skip the lockfile. We also add this to --release (and -p).

- fix #8684

### DIFF
--- a/bin/common.ml
+++ b/bin/common.ml
@@ -102,6 +102,7 @@ module Options_implied_by_dash_p = struct
     ; always_show_command_line : bool
     ; promote_install_files : bool
     ; require_dune_project_file : bool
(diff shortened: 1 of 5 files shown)


rgrinberg:
Does this work when the lock directory is configured via the workspace file as well?

Alizter:
@rgrinberg I'll add a test for the contexts soon.

PR #8766, opened 2023-09-26
fix(pkg): remove wildcard behaviour for sys vars

### LINKED ISSUE #8765: Remove wildcard behaviour for solver system environment variables
The current behaviour of the dependency solver is to treat comparisons against any unset system environment variables as "true". The goal of this behaviour is to produce lockdirs that will work on as many systems as possible, but it creates some strange behaviour in some cases.

E.g.
```
depends: [
  "uring" {os = "linux"}
  "luv" {!(os = "linux")}
]
```

The current behaviour would always pick `uring` and never pick `luv`.

Generic lockdirs isn't part of the MVP and since trying to support it is leading to this problem let's simplify things by requiring each system environment variable to be assigned a specific value while solving dependencies. The value should be derived from the current system by default but it should be possible to override variables in the build context (ie. configuration in the `dune-workspace` file).

gridbugs:
Fixes https://github.com/ocaml/dune/issues/8765

### DIFF
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -158,72 +158,30 @@ module Per_context = struct
 end
 
 module Print_solver_env = struct
-  (* Add the opam system variables derived from the current system into a
-     solver environment configured in a build context. If variables derived
-     from the current system are also present in the solver environment
-     configured in the build context, an error is raised unless they both have
-     the same value (in which case that value is used). *)
-  let merge_current_system_bindings_into_solver_env_from_context
-    ~context_name
+  (* The system environment variables used by the solver are taken from the
+     current system by default but can be overridden by the build context. *)
+  let override_solver_env_variables
     ~solver_env_from_context
     ~sys_bindings_from_current_system
-    ~use_env_from_current_system
     =
-    let sys =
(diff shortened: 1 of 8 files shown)


bin/pkg.ml
```diff
@@ -438,22 +390,15 @@ module Lock = struct
...
-              "Set opam system environment variables based on the current system when \
-               solving dependencies. This will restrict packages to just those which can \
-               be installed on the current system. Note that this will mean that the \
-               generated lockdir may not be compatible with other systems. If the build \
-               context(s) specify system environment variables then the solver will use \
-               the union of the environment variables inferred from the current system \
-               and the environment variables set in the build context(s). If the current \
-               system and the build context(s) disagree on the value of an environment \
-               variable then an error is raised.")
+              "Don't derive system solver variables from the current system. Values \
```
rgrinberg, line 400: Could you explain the option a bit more?

What happens when the solver needs to evaluate `%{os}` and we passed this option.

gridbugs: I've updated the description. Hopefully it is clearer now.

PR #8767, opened 2023-09-26
test(pkg): remove version boilerplate

rgrinberg:


Alizter (approved):
I'm guessing the opam file parser doesn't actually check the version then.

rgrinberg:
It does check it. But it's an implementation detail of opam we just acknowledege.

PR #8769, opened 2023-09-26
refactor(pkg): add module type to Context_for_dune

Alizter:
This module type was being implicitly assumed. We make this explicit.


Alizter:
Actually this makes things more annoying to edit so I'll close this.

PR #8771, opened 2023-09-26
Add [Stdune.Bit_set]

rgrinberg:
This allows us to represent small sets as a single word


Alizter:
Were you going to use this somewhere?

rgrinberg:
It's already being used internally

PR #8773, opened 2023-09-26
feature(pkg): dune pkg outdated

### LINKED ISSUE #8031: Print All Outdated Packages
[Version selection](https://github.com/ocaml/dune/issues/8021) aside, it would be independently useful to know which packages are currently out of date with respect to the opam-repository. I propose that we add a command: `dune pkg outdated` that would print all immediate dependencies that are out of date. It may also accept a `--transitive` flag to print all outdated transitive dependencies as well. This command should not modify anything in the workspace.

@Leonidas-from-XIV @gridbugs do you think this will be useful?

Alizter:
We add a `dune pkg outdated` package that will display outdated package dependencies with respect to the configured opam repositories.

It comes with a --transitive flag that shows all outdated transitive deps.

- fix #8031 

<details>
<summary>Screenshots</summary>

image
image
image

</details>

### DIFF
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -4,16 +4,8 @@ module Fetch = Dune_pkg.Fetch
 module Opam_repo = Dune_pkg.Opam_repo
 module Repository_id = Dune_pkg.Repository_id
 
-let context_term =
-  Arg.(
-    value
-    & opt (some Arg.context_name) None
-    & info
-        [ "context" ]
-        ~docv:"CONTEXT"
-        ~doc:
-          "Generate the lockdir associated with this context (the default context will \
-           be used if this is omitted)")
+let context_term ~doc =
+  Arg.(value & opt (some Arg.context_name) None & info [ "context" ] ~docv:"CONTEXT" ~doc)
 ;;
 
 module Version_preference = struct
@@ -215,7 +207,11 @@ module Print_solver_env = struct
 
   let term =
     let+ (common : Common.t) = Common.term
-    and+ context_name = context_term
+    and+ context_name =
+      context_term
+        ~doc:
+          "Generate the lockdir associated with this context (the default context will \
+           be used if this is omitted)"
     and+ all_contexts =
       Arg.(
         value
@@ -356,6 +352,25 @@ module Lock = struct
                ]))
   ;;
 
+  let find_local_packages =
+    let open Fiber.O in
+    let+ project =
+      let+ source_dir = Memo.run (Source_tree.root ()) in
+      Source_tree.Dir.project source_dir
+    in
+    Dune_project.packages project
+    |> Package.Name.Map.map ~f:(fun pkg ->
+      let opam_file = Package.to_opam_file pkg in
+      let file =
+        Path.source
+        @@
+        match pkg.has_opam_file with
+        | Generated | Exists false -> Dune_project.file project
+        | Exists true -> pkg.opam_file
+      in
+      { Opam_repo.With_file.opam_file; file })
+  ;;
+
   let solve
     per_context
     ~opam_repository_path
@@ -368,23 +383,7 @@ module Lock = struct
     (* a list of thunks that will perform all the file IO side
        effects after performing validation so that if materializing any
        lockdir would fail then no side effect takes place. *)
-    (let* local_packages =
-       let+ project =
-         let+ source_dir = Memo.run (Source_tree.root ()) in
-         Source_tree.Dir.project source_dir
-       in
-       Dune_project.packages project
-       |> Package.Name.Map.map ~f:(fun pkg ->
-         let opam_file = Package.to_opam_file pkg in
-         let file =
-           Path.source
-           @@
-           match pkg.has_opam_file with
-           | Generated | Exists false -> Dune_project.file project
-           | Exists true -> pkg.opam_file
-         in
-         { Opam_repo.With_file.opam_file; file })
-     in
+    (let* local_packages = find_local_packages in
      let+ solutions =
        Fiber.parallel_map
          per_context
@@ -476,7 +475,11 @@ module Lock = struct
     let+ (common : Common.t) = Common.term
     and+ opam_repository_path = Opam_repository_path.term
     and+ opam_repository_url = Opam_repository_url.term
-    and+ context_name = context_term
+    and+ context_name =
+      context_term
+        ~doc:
+          "Generate the lockdir associated with this context (the default context will \
+           be used if this is omitted)"
     and+ all_contexts =
       Arg.(
         value
@@ -529,6 +532,133 @@ module Lock = struct
   let command = Cmd.v info term
 end
 
+module Outdated = struct
+  let find_outdated_packages
+    ~context_name_arg
+    ~all_contexts_arg
+    ~opam_repository_path
+    ~opam_repository_url
+    ~transitive
+    ()
+    =
+    let open Fiber.O in
+    let+ pps, not_founds =
+      Per_context.choose ~context_name_arg ~all_contexts_arg ~version_preference_arg:None
+      >>= Fiber.parallel_map
+            ~f:
+              (fun
+                { Per_context.lock_dir_path
+                ; version_preference = _
+                ; repos
+                ; solver_env = solver_env_from_context
+                ; context_common = _
+                }
+              ->
+              let solver_env =
+                Print_solver_env.override_solver_env_variables
+                  ~solver_env_from_context
+                  ~sys_bindings_from_current_system:
+                    Dune_pkg.Solver_env.Variable.Sys.Bindings.empty
+              in
+              let+ repos =
+                Lock.get_repos repos solver_env ~opam_repository_path ~opam_repository_url
+              and+ local_packages = Lock.find_local_packages in
+              let lock_dir = Lock_dir.read_disk lock_dir_path in
+              let results =
+                Dune_pkg_outdated.find ~repos ~local_packages lock_dir.packages
+              in
+              ( Dune_pkg_outdated.pp ~transitive ~lock_dir_path results
+              , ( Dune_pkg_outdated.packages_that_were_not_found results
+                  |> Package_name.Set.of_list
+                  |> Package_name.Set.to_list
+                , lock_dir_path
+                , repos ) ))
+      >>| List.split
+    in
+    (match pps with
+     | [ _ ] -> Console.print pps
+     | _ -> Console.print [ Pp.enumerate ~f:Fun.id pps ]);
+    let error_messages =
+      List.filter_map not_founds ~f:(function
+        | [], _, _ -> None
+        | packages, lock_dir_path, repos ->
+          Some
+            [ Pp.textf
+                "When checking %s, the following packages:"
+                (Path.Source.to_string_maybe_quoted lock_dir_path)
+            ; Pp.enumerate packages ~f:(fun name ->
+                Dune_lang.Package_name.to_string name |> Pp.verbatim)
+            ; Pp.text "were not found in the following opam repositories:"
+            ; Pp.enumerate repos ~f:(fun repo ->
+                Opam_repo.serializable repo
+                |> Dyn.option Opam_repo.Serializable.to_dyn
+                |> Dyn.pp)
+            ])
+    in
+    match error_messages with
+    | [] -> ()
+    | [ error_message ] -> User_error.raise error_message
+    | error_messages ->
+      User_error.raise
+        [ Pp.text "Some packages could not be found."
+        ; Pp.enumerate ~f:(Pp.concat ~sep:Pp.newline) error_messages
+        ]
+  ;;
+
+  let term =
+    let+ (common : Common.t) = Common.term
+    and+ context_name_arg =
+      context_term ~doc:"Check for outdated packages in this context"
+    and+ all_contexts_arg =
+      Arg.(
+        value
+        & flag
+        & info [ "all-contexts" ] ~doc:"Check for outdated packages in all contexts")
+    and+ opam_repository_path = Lock.Opam_repository_path.term
+    and+ opam_repository_url = Lock.Opam_repository_url.term
+    and+ transitive =
+      Arg.(
+        value
+        & flag
+        & info
+            [ "transitive" ]
+            ~doc:"Check for outdated packages in transitive dependencies")
+    in
+    let common = Common.forbid_builds common in
+    let config = Common.init common in
+    Scheduler.go ~common ~config
+    @@ find_outdated_packages
+         ~context_name_arg
+         ~all_contexts_arg
+         ~opam_repository_path
+         ~opam_repository_url
+         ~transitive
+  ;;
+
+  let info =
+    let doc = "Check for outdated packages" in
+    let man =
+      [ `S "DESCRIPTION"
+      ; `P
+          "List packages in from lock directory that have newer versions available. By \
+           default, only direct dependencies are checked. The $(b,--transitive) flag can \
+           be used to check transitive dependencies as well."
+      ; `P "For example:"
+      ; `Pre "    \\$ dune pkg outdated"
+      ; `Noblank
+      ; `Pre "    1/2 packages in dune.lock are outdated."
+      ; `Noblank
+      ; `Pre "    - ocaml 4.14.1 < 5.1.0"
+      ; `Noblank
+      ; `Pre "    - dune 3.7.1 < 3.11.0"
+      ]
+    in
+    Cmd.info "outdated" ~doc ~man
+  ;;
+
+  let command = Cmd.v info term
+end
+
 let info =
   let doc = "Experimental package management" in
   let man =
@@ -540,4 +670,4 @@ let info =
   Cmd.info "pkg" ~doc ~man
 ;;
 
-let group = Cmd.group info [ Lock.command; Print_solver_env.command ]
+let group = Cmd.group info [ Lock.command; Print_solver_env.command; Outdated.command ]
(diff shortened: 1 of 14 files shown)


test/blackbox-tests/test-cases/pkg/outdated.t
```diff
@@ -0,0 +1,53 @@
...
+
+If we remove packages from the repository then we should get a nice error.
+  $ rm -rf mock-opam-repository/packages/bar
+  $ outdated 
+  Error: bar is not available in any of the configured repositories
+  [1]
+
+When printing both successes and failures, the errors should appear afterwards.
+  $ outdated --transitive
+  Error: bar is not available in any of the configured repositories
```
Alizter, line 52: I changed some things around and I missed that this has now swallowed its output. Will fix.

src/dune_pkg/opam_solver.ml
```diff
@@ -101,10 +101,10 @@ module Context_for_dune = struct
     { repos; version_preference; local_packages; solver_env }
   ;;
 
-  type rejection = Unavailable
+  type rejection = [ `Unavailable ]
```
rgrinberg, line 104: Why is this changed?

src/dune_pkg/opam_solver.ml
```diff
@@ -101,10 +101,10 @@ module Context_for_dune = struct
...
 
   let pp_rejection f = function
-    | Unavailable -> Fmt.string f "Availability condition not satisfied"
+    | `Unavailable -> Fmt.string f "Availability condition not satisfied"
```
rgrinberg, line 107: Note for later: stop using the `Fmt` library in our code.

src/dune_pkg/opam_solver.ml
Alizter: I had tried something but reverted it later, looks like I missed this.

rgrinberg:
I like how the output can separate transitive vs. immediate dependencies, but I don't think this is something users will find obvious. So how about making the normal mode explain this. E.g.
```
12/15 immediate and 140/200 transitive dependencies are out of date. Displaying immediate dependencies only, add `--transitive` to display everything.
```

bin/pkg.ml
```diff
@@ -505,6 +518,129 @@ module Lock = struct
...
+                ]
+            (* If the package is missing from the repositories, we print an error. *)
+            | Error `Package_not_found ->
+              User_error.raise
+                [ Pp.textf
+                    "%s is not available in any of the configured repositories"
+                    (Dune_lang.Package_name.to_string pkg.info.name)
+                ])
+        in
+        match errors with
```
rgrinberg, line 617: `Fiber.reraise_all`

bin/pkg.ml
```diff
@@ -505,6 +518,129 @@ module Lock = struct
...
+                           (Pp.verbatim pkg.info.version)
+                       ; Pp.text " < "
+                       ; Pp.tag
+                           (User_message.Style.Ansi_styles [ `Fg_bright_green ])
+                           (Pp.verbatim (OpamPackage.Version.to_string version))
+                       ]
+                ]
+            (* If the package is missing from the repositories, we print an error. *)
+            | Error `Package_not_found ->
+              User_error.raise
```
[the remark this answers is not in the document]
Alizter: Rather than catching exceptions I've opted to just use `Result` on which I can manually match and filter and then raise errors later when I need to.

Alizter:
@rgrinberg I've refactored the error handling to use result instead rather than messing around with fibers. At the end I use `Fiber.parallel_iter` to raise all of them as user errors. I could combine this into a single raise if that's better.

I've also added support for the `--all-contexts` arg and updated the test so that it works for lockfiles in different contexts also. We also now print which lock file we are talking about to be clearer to the user.

I'll see what I can do for adding more information about immediate vs transitive deps in the message and also some counting.

Alizter:
I've merged the errors into a single error message, however they are still separate errors per context. See the test for how this looks in practice.

Alizter:
I've pushed a change that gives informative statuses based on the information present. Notably there is now a module `Dune_pkg.Better_candidate` in charge of dealing with all this status printing business. It was getting too long and complex to include in pkg.ml so I opted to move it to its separate module where its internal workings are hidden.

This also lets us test the messages extensively, so I have gone ahead and done that and enumerated all the combinations of 4 packages both transitive and immediate to test how the message ought to change.

I haven't yet settles on the forms of the status, notably I didn't incorporate @rgrinberg's suggestion of counts in the statuses yet.

I also have some cleanup to do in the cram test since there is some redundant information with the context.

I will iterate on this tomorrow.

Alizter:
I've pushed some more changes.

The number of packages that need updating is now reported. Have a look at the screenshots spoiler in the original post.

src/dune_pkg/better_candidate.mli
```diff
@@ -0,0 +1,21 @@
...
+  | Package_not_found
+  | Package_is_best_candidate
+
+type t =
+  { is_immediate_dep_of_local_package : bool
+  ; result : result
+  ; pkg : Lock_dir.Pkg.t
+  }
+
+val outdated_packages : t list -> (bool * Import.Package_name.t * string * string) list
```
rgrinberg, line 14: It would be easier to read if the tuple was replaced with a record.

src/dune_pkg/better_candidate.mli
```diff
@@ -0,0 +1,21 @@
+open Import
```
rgrinberg, line 1: A comment about the purpose of this module would be useful

src/dune_pkg/better_candidate.mli
```diff
@@ -0,0 +1,21 @@
...
+  | Package_is_best_candidate
+
+type t =
+  { is_immediate_dep_of_local_package : bool
+  ; result : result
+  ; pkg : Lock_dir.Pkg.t
+  }
+
+val outdated_packages : t list -> (bool * Import.Package_name.t * string * string) list
+val packages_that_were_not_found : t list -> Import.Package_name.t list
```
rgrinberg, line 15: `Import` should be opened

src/dune_pkg/better_candidate.mli
```diff
@@ -0,0 +1,21 @@
...
+
+val outdated_status
+  :  transitive:bool
+  -> Path.Source.t
```
rgrinberg, line 19: It's not clear what the `Path.Source.t` argument is

src/dune_pkg/opam_repo.mli
```diff
@@ -22,6 +22,7 @@ val of_opam_repo_dir_path
   -> t
 
 val repo_id : t -> Repository_id.t option
+val source : t -> string option
```
rgrinberg, line 25: We don't have a type for sources?

src/dune_pkg/opam_solver.mli
```diff
@@ -22,3 +22,27 @@ val solve_lock_dir
...
+
+    - [Package_is_best_candidate] indicates that the given pkg is already the best
+      available candidate.
+
+    - [Package_not_found] indicates that the given pkg cannot be found in any of the given
+      repositories. *)
+val better_candidate
+  :  solver_env:Solver_env.t
+  -> repos:Opam_repo.t list
+  -> version_preference:Version_preference.t
```
rgrinberg, line 45: Why is version preference being passed? The `Better_candidate.t` generated should be independent from this option used to direct the solver. In fact, I'm not sure why the solver environment is needed either.

test/expect-tests/dune_pkg/dune_pkg_better_candidate_test.ml
```diff
@@ -0,0 +1,360 @@
...
+    ; pkg = create_dummy_package "bar" "2.0.0"
+    })
+  @ List.init number_of_transitive ~f:(fun _ ->
+    { Better_candidate.is_immediate_dep_of_local_package = false
+    ; result = Better_candidate { version = OpamPackage.Version.of_string "3.0.0" }
+    ; pkg = create_dummy_package "bar" "2.0.0"
+    })
+;;
+
+let find_color_of_line line =
```
rgrinberg, line 49: How about you just strip the colors from the output in these tests? It doesn't seem like it's worth testing.

Alizter: I included them so that we can see exactly what colours the messages are, as there are a lot of them.

bin/pkg.ml
```diff
@@ -491,6 +505,148 @@ module Lock = struct
...
+        value
+        & flag
+        & info
+            [ "transitive" ]
+            ~doc:"Check for outdated packages in transitive dependencies")
+    in
+    let common = Common.forbid_builds common in
+    let config = Common.init common in
+    Scheduler.go ~common ~config
+    @@ fun () ->
```
rgrinberg, line 531: Could you move the argument to `Scheduler.go` to its own function?

src/dune_pkg/opam_solver.mli
Alizter: This is because we are using the `candidates` function from the `CONTEXT` we pass to 0install. Unfortunately this requires us to initialize some dummy parameters.

Version preference is still something that is relevant for us however as it suggests what a better candidate will be. I suppose this will always be newer packages.

 Solver env could probably be factored out but I didn't want to touch the 0install context code too much. Happy to do this if you prefer.

src/dune_pkg/opam_repo.mli
Alizter: We do not, this is probably why this was hidden to begin with.

src/dune_pkg/better_candidate.mli
Alizter: It's the lock directory. I've cleared this up now.

test/expect-tests/dune_pkg/dune_pkg_better_candidate_test.ml
Alizter: Its also the kind of thing that we will never see during development but users will see all the time. I therefore think its important to have some idea of what users will end up seeing.

src/dune_pkg/opam_solver.mli
Alizter: I've removed all the usage of 0install now.

Alizter:
@rgrinberg I've moved some stuff around and removed the usage of 0install. The old `Better_candidate` module was renamed to `Outdated_packages` and the new API allows you to `find`, `pp` and `raise_errors` that occurred when searching. Some backdoors for testing were left in a `For_tests` module.

src/dune_pkg/outdated_packages.mli
gridbugs, line 1: Thank you for documenting this file so thoroughly.

src/dune_pkg/outdated_packages.ml
```diff
@@ -0,0 +1,226 @@
...
+;;
+
+let better_candidate ~repos ~local_packages (pkg : Lock_dir.Pkg.t) : result =
+  let pkg_name = pkg.info.name |> Package_name.to_string |> OpamPackage.Name.of_string in
+  let is_immediate_dep_of_local_package =
+    Package_name.Map.exists local_packages ~f:(fun opam_file ->
+      OpamFile.OPAM.depends opam_file
+      |> OpamFilter.filter_deps
+           ~build:true
+           ~post:true
```
rgrinberg, line 121: Should not include post dependencies

src/dune_pkg/outdated_packages.mli
```diff
@@ -0,0 +1,50 @@
...
+  -> t
+  -> User_message.Style.t Pp.t list
+
+(** [raise_errors ~raise ~lock_dir_path t] raises an errors that occured whilst searching
+    for outdated packages. This allows the errors to be delayed so that other results can be
+    printed first. *)
+val raise_errors
+  :  default:User_message.Style.t Pp.t list
+  -> t
+  -> User_message.Style.t Pp.t list
```
rgrinberg, line 35: Why does have a function that raises an error have a non polymorphic return type?

src/dune_pkg/outdated_packages.ml
```diff
@@ -0,0 +1,226 @@
...
+;;
+
+let packages_that_were_not_found l =
+  List.filter_map l ~f:(function
+    | Package_not_found name -> Some name
+    | _ -> None)
+;;
+
+(* Helpers for writing messages irregardless of the number of packages. *)
+let the_amount_out_of total amount =
```
rgrinberg, line 40: You can just write `[x/y]` in every single one of these cases.

src/dune_pkg/outdated_packages.ml
```diff
@@ -0,0 +1,226 @@
...
+           ; "outdated."
+           ]
+    ]
+    @ transitive_helper ~all_of
+  in
+  match transitive_status with
+  (* If there are no outdated transitive deps then everything is up to date. *)
+  | `No_transitive_deps_outdated ->
+    [ Pp.tag User_message.Style.Success
+      @@ Pp.textf "%s is up to date!" (Path.Source.to_string lock_dir_path)
```
rgrinberg, line 107: can you remove the exclamation mark?

src/dune_pkg/outdated_packages.ml
```diff
@@ -0,0 +1,226 @@
...
+    ]
+  | `All_transitive_deps_outdated -> packages_in_lockdir_are true number_of_outdated_deps
+  | `Some_transitive_deps_outdated ->
+    packages_in_lockdir_are false number_of_outdated_deps
```
rgrinberg, line 111: without a label, this `false` argument is hard to read.

src/dune_pkg/outdated_packages.ml
```diff
@@ -0,0 +1,226 @@
...
+  | pkg_names ->
+    User_error.raise
+      ([ Pp.text "The following packages were not found in the configured repositories:"
+       ; Pp.enumerate pkg_names ~f:(fun name ->
+           Dune_lang.Package_name.to_string name |> Pp.verbatim)
+       ]
+       @ default)
+;;
+
+let find ~repos ~local_packages { Lock_dir.packages; _ } =
```
rgrinberg, line 206: If this function only needs the packages, it would be better if it just took the packages as an argument.

src/dune_pkg/outdated_packages.mli
Alizter: I've mixed up some concerns here. It is printing spaces when there are no errors.

rgrinberg:
All the pretty printing code should be moved from `dune_pkg` and into the binary. All of that stuff are implementation details of the command itself.

Alizter:
@rgrinberg Can I propose splitting up `pkg.ml` then? Otherwise that command is getting too long.

rgrinberg:
How about you just add your own command as `pkg_outdated` and split the existing stuff in a subsequent PR?

Alizter:
@rgrinberg Sorry, I didn't mean splitting all the other commands, I just meant outdated.

bin/pkg.ml
```diff
@@ -504,6 +510,104 @@ module Lock = struct
...
+        let+ repos =
+          Lock.get_repos repos solver_env ~opam_repository_path ~opam_repository_url
+        and+ local_packages = Lock.find_local_packages in
+        let lock_dir = Lock_dir.read_disk lock_dir_path in
+        let results = Outdated_packages.find ~repos ~local_packages lock_dir in
+        Console.print (Outdated_packages.pp ~transitive ~lock_dir_path results);
+        let default_spacing =
+          if per_context_index < List.length per_context - 1 then [ Pp.nop ] else []
+        in
+        Console.print (Outdated_packages.raise_errors results ~default:default_spacing))
```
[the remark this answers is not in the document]
Alizter: Sure. I'll do all the printing at once. The only issue is how to have the errors show for each context separately.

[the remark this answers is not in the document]
Alizter: I've accumulated both the messages and the errors and now print and raise at the end.

bin/pkg.ml
```diff
@@ -529,6 +534,134 @@ module Lock = struct
...
+           be used to check transitive dependencies as well."
+      ; `P "For example:"
+      ; `Pre "    \\$ dune pkg outdated"
+      ; `Noblank
+      ; `Pre "    Outdated packages in dune.lock:"
+      ; `Noblank
+      ; `Pre "    ocaml 5.1.1 < 5.2.0"
+      ; `Noblank
+      ; `Pre "    dune 3.7.1 < 3.11.0"
+      ]
```
Alizter, line 657: This description needs updating.

test/expect-tests/dune_pkg/dune
```diff
@@ -1,5 +1,6 @@
 (library
  (name dune_pkg_unit_tests)
+ (modules dune_pkg_unit_tests)
```
rgrinberg, line 3: don't think you need this

test/expect-tests/dune_pkg_outdated/dune
```diff
@@ -0,0 +1,19 @@
+(library
+ (name dune_pkg_outdated_test)
+ (modules dune_pkg_outdated_test)
```
rgrinberg, line 3: not necessary

test/expect-tests/dune_pkg_outdated/dune
```diff
@@ -0,0 +1,19 @@
...
+ (inline_tests)
+ (libraries
+  dune_tests_common
+  stdune
+  dune_pkg_outdated
+  dune_console
+  threads.posix
```
rgrinberg, line 10: where do you use threads?

Alizter: I don't, I got confused it confused with the other transitive deps.

Alizter:
@rgrinberg I've removed the

PR #8774, opened 2023-09-26
test(pkg): building with empty dune.lock/

Alizter:
Having an empty dune.lock directory should not prevent a regular dune build.

This can happen if a lock file was created and then the git changes discarded for example.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/empty-lock.t
+++ b/test/blackbox-tests/test-cases/pkg/empty-lock.t
@@ -1,0 +1,11 @@
+An empty dune.lock directory should prevent dune from building. We should also
+give a nice error message telling the user what is wrong and how it can be
+fixed.
+
+  $ mkdir dune.lock
+  $ cat > dune-project <<EOF
+  > (lang dune 3.11)
+  > EOF
+  $ dune build
+  Error: dune.lock/lock.dune: No such file or directory
+  [1]


Alizter:
I have a fix for this. @gridbugs could you take a look?

gridbugs (approved):

rgrinberg:
Could you describe what the fix is? To me, having an empty directory like this is indeed a malformed lock directory.

test/blackbox-tests/test-cases/pkg/empty-lock.t
```diff
@@ -0,0 +1,9 @@
+An empty dune.lock directory should not prevent dune from building.
```
rgrinberg, line 1: should be `should prevent`

rgrinberg:
> A user deletes the files in the lock directory using git.

How can this be done without deleting the directory?

> Switches branches which adds and cleans out the lock directory.

I just tried this and directory was deleted. Can you demonstrate the steps that make this possible?

gridbugs:
How does dune know that the `dune.lock` directory is a lockdir? That's just the default name, right? - the lockdir could be named anything configured in the workspace.

Alizter:
I could not reproduce this again. I guess I will change the test description as Rudi says and create an issue for a better error message here.
```
  Error: dune.lock/lock.dune: No such file or directory
```
Should really be:
```
  Error: dune.lock is malformed. Please delete or recreate the lock directory.
  Hint: dune pkg lock
```

Alizter:
This has been resolved in another way and I don't think there is much use in keeping this test around.

PR #8776, opened 2023-09-27
test: dune-workspace and --release

Alizter:
We test the interaction of dune-workspace files and --release. The current behaviour appears to be to not ignore them in --release mode.


rgrinberg:
Thanks. This behavior is probably not desirable. As it means that whoever is installing the package cannot configure settings that were set by dune-workspace.

Issue #108, opened 2017-05-30
bytecode + c stubs not working as expected


rgrinberg:
See this repo:

https://github.com/rgrinberg/jbuilder-c-stubs

```
[rgrinberg:~/tmp/c-ocaml] 1 % jbuilder build hello_world.exe
[rgrinberg:~/tmp/c-ocaml] % _build/default/hello_world.exe
Hello
[rgrinberg:~/tmp/c-ocaml] % _build/default/hello_world.bc
Fatal error: cannot load shared library dllc_lib_stubs
Reason: dllc_lib_stubs.so: cannot open shared object file: No such file or directory

```

Am I doing something wrong here?

diml:
`ocamlrun` looks up shared libraries for C stubs using CAML_LD_LIBRARY_PATH. So you'd need to set it. `jbuulder exec ...` does that for you

rgrinberg:
Hm, but `CAML_LD_LIBRARY_PATH` seems correct to me:

```
[rgrinberg:~/tmp/c-ocaml] % echo $CAML_LD_LIBRARY_PATH
/home/rgrinberg/.opam/4.04.1/lib/stublibs:/home/rgrinberg/.opam/4.04.1/lib/ocaml/stublibs:/home/rgrinberg/.opam/4.04.1/lib/ocaml
```

opam switch is supposed to set it right?

But regardless, my executable isn't public, can I still somehow run it? I can make it public, but it seems annoying to have to make executables just to run them.

diml:
Well, it only points to things that are installed. In your example the dll for the stubs is in _build/...

As long as the library with the stubs is public you'll be able to run it with `jbuilder exec ...`. It's a limitation of shared libraries really, jbuilder can't do much about it. If you want to call the executable inside a rule, you should just use thr .exe

rgrinberg:
Hmm, I'm still unable to run the bytecode binary this way. Even after setting CAML_LD_LIBRARY_PATH to what `jbuilder exec` would. See:

```
[rgrinberg:~/tmp/c-ocaml] master(+2/-1)* 2 ± jbuilder exec -- env sh -c 'echo $CAML_LD_LIBRARY_PATH'
[...]

PR #8784, opened 2023-09-27
Add link flags ocamlmklib when using ctypes stubs.

frejsoya:
Add link flags for external dependencies when using ctypes.

  Required when the external lib is not statically linked, in which case
  linker needs to be aware at runtime what libs to call. This is the case
  for

  - dune exec <file>.bc
  - dune utop
  - mdx tests

  That is add to ocamlmklib, link flags such as -l<libname> -L<libdir>
  from ctypes if info is available.

  If dependent library are not installed in expected locations, env
  variable such as LD_LIBRARY_PATH must be extended when required program
  is run, depending on os. This may be fixed by adding relocatable paths to
  link flags.

### DIFF
--- a/doc/changes/8784.md
+++ b/doc/changes/8784.md
@@ -1,0 +1,1 @@
+- Add link flags to to ocamlmklib for ctypes stubs (#8784, @frejsoya)
--- a/src/dune_rules/lib_rules.ml
+++ b/src/dune_rules/lib_rules.ml
@@ -328,7 +328,14 @@ let build_stubs lib ~cctx ~dir ~expander ~requires ~dir_contents ~vlib_stubs_o_f
       | _ -> Action_builder.return []
     in
     let c_library_flags =
-      Expander.expand_and_eval_set expander lib.c_library_flags ~standard
+      let open Action_builder.O in
+      let+ c_lib = Expander.expand_and_eval_set expander lib.c_library_flags ~standard
+      and+ ctypes_lib =
+        (* CR rgrinberg: Should we add these flags to :standard? to make
+           it possible for users to remove these *)
+        Ctypes_rules.ctypes_cclib_flags sctx ~expander ~buildable:lib.buildable
+      in
+      c_lib @ ctypes_lib
     in
     let lib_o_files_for_all_modes = Mode.Map.Multi.for_all_modes o_files in
     let for_all_modes = List.rev_append vlib_stubs_o_files lib_o_files_for_all_modes in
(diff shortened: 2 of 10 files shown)


frejsoya:
This change simply copies what is done in exe_rules.ml

It took some time to figure out that lib_rules.ml has to check explicitly what ctypes believes are the link flags vs. it being contained within in ctype_rules.ml:gen_rules

lukstafi:
I face the problem of `ocamlmklib` not getting the library that it's supposed to link with for both the manual "vendored" setup, and using `pkg_config` (where I generate the `.pc` file manually).

rgrinberg:
@lukstafi does this PR fix the problem for you?

lukstafi:
> @lukstafi does this PR fix the problem for you?

Yes! For `(build_flags_resolver pkg_config)`, this PR solves the problem completely. For `(build_flags_resolver (vendored ...))`, with this PR (not synced/rebased) I regress to issue #108. Then, I can ameliorate that one with e.g. `CAML_LD_LIBRARY_PATH=./_build/default/lib:$CAML_LD_LIBRARY_PATH  dune exec examples/tut01_hello_world.bc`.
Strangely/unfortunately, `(env (_ (env-vars (CAML_LD_LIBRARY_PATH "./_build/default/lib:..."))))` does not work.

In my case, I would go ahead using `(build_flags_resolver pkg_config)` (it's cleaner despite the extra complexity).

frejsoya:
Note i have little clue about dune internals and even more so with ctypes. Similar its been decades since i dealt with linkers/shared libraries.

There are a few things that need to be correct  if i recall correctly. 

1) dllstub.so must be locatable by ocamlrun https://v2.ocaml.org/manual/runtime.html#s:ocamlrun-dllpath
2) afaik ocamlrun does dlopen(dllstub.so). So any C libraries needed by dllstub needs to be resolvable by the os linker. (Not ocamlrun). But i did not verify this with anyone.


For the vendored libraries (non pkg-config) either manually setting LD_LIBRARY_PATH at runtime or adding the right rolcoatable (rpath) linker flags could resolve that part.

(See rpath discussion below wrt to CMAKE.

https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling

lukstafi:
> For the vendored libraries (non pkg-config) either manually setting LD_LIBRARY_PATH at runtime or adding the right rolcoatable (rpath) linker flags could resolve that part.
> 

So it looks to me `rpath` is not helpful in case of `dllstubs` because it would require different build flags / binaries for development and for install.

lukstafi:
Another complication, perhaps not related to this PR: bytecode expect-test framework tests: `(library (inline_tests) (modes byte) ...)` are not run by `dune test` -- it only triggers a build.

rgrinberg:
> Another complication, perhaps not related to this PR: bytecode expect-test framework tests: `(library (inline_tests) (modes byte) ...)` are not run by `dune test` -- it only triggers a build.

Would you like to report a separate issue for this? Preferably with a minimal example that reproduces the issue.

PR #8791, opened 2023-09-28
refactor(engine): replace collect_errors with finalize

rgrinberg:
The behavior should be equivalent but [finalize] is a little cleaner and
faster.

### DIFF
--- a/src/dune_engine/build_system.ml
+++ b/src/dune_engine/build_system.ml
@@ -380,53 +380,47 @@ end = struct
          | None -> root
          | Some sandbox -> Sandbox.map_path sandbox root)
     in
-    let* exec_result =
-      Fiber.collect_errors
-      @@ fun () ->
-      with_locks locks ~f:(fun () ->
-        let build_deps deps = Memo.run (build_deps deps) in
-        let* action_exec_result =
-          let input =
-            { Action_exec.root
-            ; context
-            ; env
-            ; targets = Some targets
-            ; rule_loc = loc
-            ; execution_parameters
-            ; action
-            }
-          in
-          match (Build_config.get ()).action_runner input with
-          | None -> Action_exec.exec input ~build_deps
-          | Some runner -> Action_runner.exec_action runner input
-        in
-        let* action_exec_result = Action_exec.Exec_result.ok_exn action_exec_result in
-        let+ produced_targets =
-          match sandbox with
-          | None -> Targets.Produced.produced_after_rule_executed_exn ~loc targets
-          | Some sandbox ->
-            (* The stamp file for anonymous actions is always created outside
-               the sandbox, so we can't move it. *)
-            let should_be_skipped =
-              match rule_kind with
-              | Normal_rule -> fun (_ : Path.Build.t) -> false
-              | Anonymous_action { stamp_file; _ } -> Path.Build.equal stamp_file
+    Fiber.finalize
+      ~finally:(fun () ->
+        match sandbox with
+        | Some sandbox -> Sandbox.destroy sandbox
+        | None ->
+          (* All went well, these targets are no longer pending. *)
+          Pending_targets.remove targets;
+          Fiber.return ())
+      (fun () ->
+        with_locks locks ~f:(fun () ->
+          let build_deps deps = Memo.run (build_deps deps) in
+          let* action_exec_result =
+            let input =
+              { Action_exec.root
+              ; context
+              ; env
+              ; targets = Some targets
+              ; rule_loc = loc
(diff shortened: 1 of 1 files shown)


Alizter (approved):

snowleopard:
> The behavior should be equivalent but [finalize] is a little cleaner and faster.

I agree that it's cleaner but how is it faster? `Fiber.finalize` calls `collect_errors` on `finally`, whereas the original code doesn't.

What do benchmarks say?

snowleopard:
I also doubt the "should be equivalent" part. Surely the error-collection behaviour is changing?

rgrinberg:
> I agree that it's cleaner but how is it faster? Fiber.finalize calls collect_errors on finally, whereas the original code doesn't.

I thought `finalize` was a primitive in fiber. Indeed it's not and there's no performance difference.

> What do benchmarks say?

They say there's no difference. Not surprisingly.

> I also doubt the "should be equivalent" part. Surely the error-collection behaviour is changing?

Yes, and actually I think this new behavior is improved. In the code before this PR, an IO error deleting the sandbox or deleting the pending targets would cause the original errors from running the actions to be discarded.

snowleopard:
> > I also doubt the "should be equivalent" part. Surely the error-collection behaviour is changing?
> 
> Yes, and actually I think this new behavior is improved. In the code before this PR, an IO error deleting the sandbox or deleting the pending targets would cause the original errors from running the actions to be discarded.

Yep, that sounds right. Maybe it's worth adding a test that for this case? (Feel free to skip if it's difficult to write such a test.)

snowleopard (approved):

rgrinberg:
It ended up being a little difficult to write a test. The only scenario I could think of is the user accidentally meddling with the sandbox.

snowleopard:
> The only scenario I could think of is the user accidentally meddling with the sandbox.

...which is not really uncommon and providing good error messages in this case seems useful.

snowleopard:
Imported this PR internally.

@rgrinberg What's the meaning of the `upstream-sync` label? Shall I remove it now that I've imported the PR?

src/dune_engine/build_system.ml
```diff
@@ -380,53 +380,47 @@ end = struct
...
-            let should_be_skipped =
-              match rule_kind with
-              | Normal_rule -> fun (_ : Path.Build.t) -> false
-              | Anonymous_action { stamp_file; _ } -> Path.Build.equal stamp_file
+    Fiber.finalize
+      ~finally:(fun () ->
+        match sandbox with
+        | Some sandbox -> Sandbox.destroy sandbox
+        | None ->
+          (* All went well, these targets are no longer pending. *)
```
snowleopard, line 388: Btw, the "all went well" part is a little confusing (the action might have failed actually). Internally, I've just deleted this comment as it seems unnecessary.

snowleopard:
> Imported this PR internally.

Actually, importing this PR turned out pretty annoying because internally we have the code below. Note that we run `Js_corrected_files.rename_corrected_files` only in the error case. This makes it necessary to change `Fiber.finalize` to communicate the execution result to `finally`, which is a somewhat invasive change.

```
    let () =
      match exec_result, sandbox with
      | Error _, Some sandbox -> Js_corrected_files.rename_corrected_files sandbox
      | _ -> ()
    in
    let* () =
      match sandbox with
      | Some sandbox -> Sandbox.destroy sandbox
      | None ->
        (* All went well, these targets are no longer pending. *)
        Pending_targets.remove targets;
        Fiber.return ()
    in
    match exec_result with
    | Ok res -> Fiber.return res
    | Error l -> Fiber.reraise_all l
```

snowleopard:
OK, never mind the last point. I figured out where to move the `Js_corrected_files.rename_corrected_files` bit without introducing a more complicated version of `Fiber.finalize`.

rgrinberg:
> What's the meaning of the upstream-sync label? Shall I remove it now that I've imported the PR?

Yup. That's the goal. Just a list we can use to manage the PR's that are still pending.

src/dune_engine/build_system.ml
rgrinberg: Would you like to submit a PR to do the same here?

snowleopard:
> Yup. That's the goal. Just a list we can use to manage the PR's that are still pending.

Good idea. Thanks!

src/dune_engine/build_system.ml
snowleopard: Done: https://github.com/ocaml/dune/pull/8808

PR #8793, opened 2023-09-29
test(pkg): demonstrate that lockfiles in workspace are not ignored

Alizter:
Following up from #8684, it appears that lockfiles in the workspace are not ignored. This could be solved by just ignoring the workspace.


rgrinberg:
Isn't this redundant? We don't ignore the workspace for other settings either in release mode.

Alizter:
But we ignore the lock file, as seen above in the test. In --release mode we should ignore the lock file in all contexts.

rgrinberg:
I don't think that makes sense. Either we ignore the workspace file or we don't.

PR #8795, opened 2023-09-29
feature: Haiku support

### LINKED ISSUE #8551: unable to bootstrap on Haiku
Haiku doesn't have `sys/syscalls.h` so our current implementation of spawn will not work. Specifically `spawn_stubs.c`.

Seems to be related to https://github.com/ocaml/dune/issues/5577

@rgrinberg is `sys/syscalls.h` even needed in that file? If the cygwin build works without it, then maybe it is better to remove?

Alizter:
This patch adds Haiku detection to our platform stubs letting us give a better hint for `fswatch` when compiling on Haiku.

It also introduces a linking flag for `dune` that allows `wait4` to be linked correctly Haiku. 

- fixes #8551 
- [x] bootstraps on Haiku
- [x] builds on Haiku
- [x] update notty fork (noctl has different location, fixed in https://github.com/ocaml-dune/notty/pull/5)
- [x] changelog

Here is TUI:
image

### DIFF
--- a/src/dune_file_watcher/dune_file_watcher.ml
+++ b/src/dune_file_watcher/dune_file_watcher.ml
@@ -358,6 +358,7 @@ let fswatch_backend () =
   | None ->
     let hints =
       match Platform.OS.value with
+      | Haiku -> [ Pp.text "fswatch is available on HaikuPorts" ]
       | FreeBSD -> [ Pp.text "pkg install fswatch-mon" ]
       | _ -> []
     in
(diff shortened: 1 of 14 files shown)


src/dune_file_watcher/dune_file_watcher.ml
```diff
@@ -358,6 +358,7 @@ let fswatch_backend () =
   | None ->
     let hints =
       match Platform.OS.value with
+      | Haiku -> [ Pp.text "fswatch is available on HaikuPorts" ]
```
waddlesplash, line 361: Binary packages are available, too. `pkgman install devel:libfswatch` will install the development files, or `pkgman install cmd:fswatch` for just the command.

Alizter: Thanks for taking a look! I'll try seeing which one is needed and update the message here accordingly.

PR #8796, opened 2023-09-29
Avoid a warning of the C compiler on Windows

shym:
The macro `unix_error` is also defined in `caml/unixsupport.h` (at least when `CAML_BUILDING_UNIX` is not set with ≥ 5.x compiler) so the C compiler generates a warning when it is redefined in `fswatch_win_stubs`.
This PR simply avoids this unnecessary warning.


Alizter (approved):
Looks sensible.

nojb (approved):
LGTM

PR #8798, opened 2023-09-29
fix: always use "portable" hardlinks

rgrinberg:
Our portable_hardlink fallbacks to copying whenever hardlinking fails
for whatever reason (too many links, windows, etc)

We now use it consistently when sandboxing & caching.

### DIFF
--- a/src/dune_cache/local.ml
+++ b/src/dune_cache/local.ml
@@ -98,14 +98,16 @@ module Artifacts = struct
      the result is [Error] with the corresponding exception. Otherwise, the
      result is [Ok ()]. *)
   let store_targets_to ~temp_dir ~targets ~mode : unit Or_exn.t =
+    let portable_hard_link_or_copy =
+      match (mode : Dune_cache_storage.Mode.t) with
+      | Hardlink -> Io.portable_hardlink
+      | Copy -> fun ~src ~dst -> Io.copy_file ~src ~dst ()
+    in
     Result.List.iter targets ~f:(fun { Target.path; _ } ->
       let path_in_build_dir = Path.build path in
       let path_in_temp_dir = Path.relative temp_dir (Path.basename path_in_build_dir) in
       Result.try_with (fun () ->
-        Dune_cache_storage.Util.link_or_copy
-          ~mode
-          ~src:path_in_build_dir
-          ~dst:path_in_temp_dir))
+        portable_hard_link_or_copy ~src:path_in_build_dir ~dst:path_in_temp_dir))
   ;;
 
   (* Step II of [store_skipping_metadata].
--- a/src/dune_cache_storage/util.ml
+++ b/src/dune_cache_storage/util.ml
@@ -16,12 +16,6 @@ module Optimistically = struct
   ;;
 end
 
-let link_or_copy ~mode ~src ~dst =
-  match (mode : Mode.t) with
-  | Hardlink -> Path.link src dst
-  | Copy -> Io.copy_file ~src ~dst ()
-;;
-
 module Write_result = struct
   type t =
     | Ok
@@ -41,6 +35,8 @@ end
 let add_atomically ~mode ~src ~dst : Write_result.t =
   match (mode : Mode.t) with
   | Hardlink ->
+    (* The [src] comes from a fresh temporary file, so we should never get
+       [EMLINK] *)
     (match Optimistically.link ~src ~dst with
      | () -> Ok
      | exception Unix.Unix_error (Unix.EEXIST, _, _) -> Already_present
--- a/src/dune_cache_storage/util.mli
+++ b/src/dune_cache_storage/util.mli
@@ -17,9 +17,6 @@ end
     function returns [Ok] instead of [Already_present]. *)
 val write_atomically : mode:Mode.t -> content:string -> Path.t -> Write_result.t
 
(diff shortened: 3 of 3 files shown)


src/dune_cache/local.ml
```diff
@@ -98,15 +67,18 @@ module Artifacts = struct
      If any of the targets couldn't be stored in the temporary directory, then
      the result is [Error] with the corresponding exception. Otherwise, the
      result is [Ok ()]. *)
-  let store_targets_to ~temp_dir ~targets ~mode : unit Or_exn.t =
+  let store_targets_to ~temp_dir ~targets ~(mode : Dune_cache_storage.Mode.t)
+    : unit Or_exn.t
+    =
+    let copy =
+      match mode with
+      | Hardlink -> fun ~src ~dst -> Io.portable_hardlink ~src ~dst
```
rgrinberg, line 75: I inlined this function from `util` because it was the only place where it was used.

src/dune_cache/local.ml
```diff
@@ -43,37 +43,6 @@ module Target = struct
...
-  | Unix.Unix_error (Unix.EUNKNOWNERR -1142, _, _) (* Needed for OCaml < 5.1 *) ->
-    Temp.with_temp_file ~dir:temp_dir ~prefix:"dune" ~suffix:"copy" ~f:(function
-      | Error e -> raise e
-      | Ok temp_file ->
-        Io.copy_file ~src ~dst:temp_file ();
-        (* This replaces [src], which has too many links already, with a fresh
-           copy we've just created in the [temp_file]. *)
-        Path.rename temp_file src;
-        (* This should now succeed. *)
-        Path.link src dst)
```
rgrinberg, line 74: The difference between `Io.portable_hardlink` is that it will just copy when hard linking fails rather than copying and then linking.

snowleopard: If you read the above comment carefully, you'll see that the old code was doing more than just falling back to copying. I don't think this place needs to change because it's not really broken. You are replacing it with something simpler but also less optimal.

src/dune_cache/local.ml
```diff
@@ -98,15 +67,18 @@ module Artifacts = struct
      If any of the targets couldn't be stored in the temporary directory, then
      the result is [Error] with the corresponding exception. Otherwise, the
      result is [Ok ()]. *)
-  let store_targets_to ~temp_dir ~targets ~mode : unit Or_exn.t =
+  let store_targets_to ~temp_dir ~targets ~(mode : Dune_cache_storage.Mode.t)
+    : unit Or_exn.t
+    =
+    let copy =
```
snowleopard, line 73: Calling this function `copy` is pretty confusing.
```suggestion
    let hardlink_or_copy =
```

src/dune_cache/local.ml
rgrinberg: Sure, this function allows for more sharing. But the only files where we get "too many links" are small anyway. So I didn't think it was worth it. I could restore it though.

snowleopard: We can discuss simplifying this code (you may be right that it's not worth it) but it seems wrong to conflate a bug fix and a possibly controversial change that removes an optimisation in the same PR.

src/dune_cache/local.ml
snowleopard: Anyway, below is the code I wrote internally. Let's use it here too, to avoid divergence.
```
let store_targets_to ~temp_dir ~targets ~mode : unit Or_exn.t =
  let portable_hard_link_or_copy =
    match (mode : Dune_cache_storage.Mode.t) with
    | Hardlink -> Io.portable_hardlink
    | Copy -> fun ~src ~dst -> Io.copy_file ~src ~dst ()
  in
```

rgrinberg: Nope, just didn't push the change yet. Pushed it now though.

snowleopard: Thanks! Apologies for being too pedantic, but could you use the code snippet I copied above? We've already released that version internally, and it'd be best to keep the diff empty.

src/dune_cache_storage/util.ml
```diff
@@ -35,6 +29,10 @@ let add_atomically ~mode ~src ~dst : Write_result.t =
     (match Optimistically.link ~src ~dst with
      | () -> Ok
      | exception Unix.Unix_error (Unix.EEXIST, _, _) -> Already_present
+     | exception Unix.Unix_error (Unix.EMLINK, _, _) ->
```
snowleopard, line 32: Btw, I think this bit is actually unnecessary because the only caller passes a fresh temp file as `src`. I didn't import this bit internally. (And according to https://github.com/ocaml/dune/pull/8816 this function isn't used anywhere else.)

rgrinberg: Added a comment in the code to reflect this.

src/dune_cache/local.ml
rgrinberg: I copied your code. Btw, `hard_link` is not consistent with `hardlink` which we use everywhere else.

snowleopard: Oops, indeed, my bad. Thanks!

snowleopard: Feel free to fix this (and I'll fix it internally too).

PR #8803, opened 2023-09-30
New odoc rules

jonludlam:
This supercedes PR #7840.

This is a _new_ set of rules that lives _alongside_ the original odoc rules. At present it builds docs for your packages, their libraries, your private libraries and all of their dependencies. I've tested this on dune itself, ocaml.org, odoc and tezos-shell, though because this is a new target, we can and should view this as a bit experimental for a bit. 

It's invoked with `dune build @doc-new` and the results are index by the main index page at `_build/default/_doc_new/html/docs/index.html`


jonludlam:
Almost 1000 PRs in between this and the previous incarnation!

jonludlam:
The test failures are interesting:

```
@@ -114,7 +114,7 @@ true
 let%expect_test _ =
   is_descendant (e "/foo/bar") ~of_:(e "/foo");
   [%expect {|
-false
+true
 |}]
 ;;
 
@@ -135,14 +135,14 @@ false
 let%expect_test _ =
   is_descendant (e "/foo/bar/") ~of_:(e "/foo/bar");
   [%expect {|
-false
+true
 |}]
 ;;
 
 let%expect_test _ =
   is_descendant (e "/foo/bar") ~of_:(e "/");
   [%expect {|
-false
+true
 |}]
 ;;
```

On the surface it seems the new results are an improvement - but I assume there's a good reason that these tests are there?

otherlibs/stdune/src/path.ml
```diff
@@ -862,6 +862,12 @@ let external_of_in_source_tree x = external_of_local x ~root:(Lazy.force abs_roo
 
 let reach t ~from =
   match t, from with
+  | External t, External from ->
+    (match
+       String.drop_prefix ~prefix:(External.to_string from ^ "/") (External.to_string t)
```
rgrinberg, line 867: Why is this needed? Can't you pass external paths as absolute paths?

jonludlam: It's for `Odoc_new.local_path_of_findlib_path` - there's a comment explaining what it does. Is there a better way to do that?

src/dune_rules/odoc_new.ml
```diff
@@ -0,0 +1,1959 @@
...
+  ;;
+
+  let alias ctx ~all index = alias ~dir:(Index.odoc_dir ctx ~all index)
+
+  let setup_deps ctx ~all m files =
+    Rules.Produce.Alias.add_deps (alias ctx ~all m) (Action_builder.path_set files)
+  ;;
+end
+
+let contains_double_underscore s =
```
rgrinberg, line 666: Which packages require this hack to work?

jonludlam: Dream is one - the sub libraries are named with double underscores.

src/dune_rules/odoc_new.ml
```diff
@@ -1397,7 +1397,8 @@ let index_info_of_pkg_def =
       pkg_artifacts sctx index pkg_name
     in
     let+ index_infos =
-      let+ findlib_paths = Context.findlib_paths ctx in
+      let* findlib_paths = Context.findlib_paths ctx in
+      let+ dwms = Valid.filter_dune_with_modules ctx ~all dwms in
```
Alizter, line 1401: You can combine this into a single `let+ and+`.

src/dune_rules/odoc_new.ml
jonludlam: Oh, and eio too: ocaml/odoc#1013

src/dune_rules/odoc_new.ml
```diff
@@ -0,0 +1,1960 @@
...
+    let local =
+      let local_full =
+        Lib.info lib
+        |> Lib_info.obj_dir
+        |> Obj_dir.dir
+        |> Paths.local_path_of_findlib_path findlib_paths
+      in
+      String.split ~on:'/' local_full
+    in
+    let init = [ Package (ty, List.hd local) ] in
```
Alizter, line 179: This won't work in general

otherlibs/stdune/src/path.ml
Alizter: We have `Path.drop_prefix` which might be more sensible. It also gives a `Local.Path.t` which I believe is what you actually want. I'm still a bit confused about the `Odoc_new.local_path_of_findlib_path` function however. It seems everywhere you use it, you split on `/` and then take the head, so I think there is some other property you are trying to extract here.

src/dune_rules/odoc_new.ml
Alizter: But see my other comments as I am dubious to the utility of the "local_path_of_findlib_path" function.

otherlibs/stdune/src/path.ml
jonludlam: eurgh - Path.drop_prefix is not always doing what I want!

```
# obj_dir = '/Users/jon/devel/odoc/_opam/lib/ocaml-compiler-libs/shadow'
# local = '-compiler-libs/shadow'
```

This is because there are two findlib paths - `/Users/jon/devel/odoc/_opam/lib` and `/Users/jon/devel/odoc/_opam/lib/ocaml` - and it's the second is definitely a string prefix of the lib above, but not a _path_ prefix.

jonludlam: I don't - you're spot on, I really want the path prefix, and getting back a `Path.Local.t` would be great. Sorry, I wasn't clear above. The above is what `Path.drop_prefix` is giving me, not what I actually want - ie, I think `Path.drop_prefix` is broken. For the above, what I'd like is to return the local path representing `ocaml-compiler-libs/shadow`.

Alizter: OK, let me try to repro in a unit test.

Alizter: Yep, that's definitely a bug. This is what happens when unit tests are not written! :)

Alizter: Or actually maybe intended. I guess what we really need is a `Path.drop_path_prefix` function which doesn't exist.

Alizter: I've created a test https://github.com/ocaml/dune/pull/8870 so that we can discuss this behaviour there.

jonludlam:
OK, the dodgy `Odoc_new.local_path_of_findlib_path` has gone, replaced by a new type `External_location` in `Dune_package` that is effectively the split I wanted. This was good because it highlighted something I hadn't considered - that with a system switch the stdlib and associated libs are not necessarily in a findlib dir. 

Secondly, @rgrinberg pointed out that there might potentially be a directory name clash for some nasty cases, so added a layer of hierarchy to the output to ensure this can never happen.

Thirdly, this prompted me to push through with some more logical simplifications -- essentially to structure the code more around libraries than packages, meaning mostly that I'm more treating things on a per-directory basis than a per-package-tree-of-directories basis.

Sorry it took some time, it ended up being a bit of a big change.

Alizter:
FTR `Path.drop_prefix` was buggy and a fix can be found here:

- https://github.com/ocaml/dune/pull/8965

Sorry about closing the PR I misclicked.

jonludlam:
I think this is now ready for review again.

rgrinberg:
Great Work. Sorry for all the delays.

PR #8805, opened 2023-09-30
refactor: centralize workaround for OCaml bug

rgrinberg:
OCaml older than 5.1 will not correctly raise EMLINK. We make
[Path.link] transalte the error correctly and remove the two other cases
where it was done.

### DIFF
--- a/otherlibs/stdune/src/io.ml
+++ b/otherlibs/stdune/src/io.ml
@@ -491,10 +491,9 @@ let portable_hardlink ~src ~dst =
           filter out the duplicates first. *)
        Path.unlink dst;
        Path.link src dst
-     | Unix.Unix_error (Unix.EMLINK, _, _)
-     | Unix.Unix_error (Unix.EUNKNOWNERR -1142, _, _) (* Needed for OCaml < 5.1 *) ->
+     | Unix.Unix_error (Unix.EMLINK, _, _) ->
        (* If we can't make a new hard link because we reached the limit on the
-          number of hard links per file, we fall back to copying. We expect that
-          this happens very rarely (probably only for empty files). *)
+          number of hard links per file, we fall back to copying. We expect
+          that this happens very rarely (probably only for empty files). *)
        copy_file ~src ~dst ())
 ;;
--- a/otherlibs/stdune/src/path.ml
+++ b/otherlibs/stdune/src/path.ml
@@ -1130,7 +1130,16 @@ let is_directory t =
 
 let rmdir t = Unix.rmdir (to_string t)
 let unlink t = Fpath.unlink (to_string t)
-let link x y = Unix.link (to_string x) (to_string y)
+
+let link src dst =
+  match Unix.link (to_string src) (to_string dst) with
+  | exception Unix.Unix_error (Unix.EUNKNOWNERR -1142, syscall, arg)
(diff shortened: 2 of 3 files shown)


Alizter (approved):

otherlibs/stdune/src/path.ml
```diff
@@ -1130,7 +1130,19 @@ let is_directory t =
...
 let rmdir t = Unix.rmdir (to_string t)
 let unlink t = Fpath.unlink (to_string t)
-let link x y = Unix.link (to_string x) (to_string y)
+
+let link x y =
+  match Unix.link (to_string x) (to_string y) with
+  | exception Unix.Unix_error (Unix.EUNKNOWNERR -1142, syscall, arg)
+  (* Needed for OCaml < 5.1 *) ->
+    (* If we can't make a new hard link because we reached the limit on the
+       number of hard links per file, we fall back to copying. We expect that
```
snowleopard, line 1139: The bit about falling back to copying is out of place now.

snowleopard (approved):
Nice! Please restore a part of the comment at the call site.

snowleopard:
> OCaml older than 5.1 will not correctly raise EMLINK.

To clarify: that's only true on Windows, right?

rgrinberg:
Yes. It's only needed on Windows.

otherlibs/stdune/src/io.ml
```diff
@@ -491,10 +491,5 @@ let portable_hardlink ~src ~dst =
           filter out the duplicates first. *)
        Path.unlink dst;
        Path.link src dst
-     | Unix.Unix_error (Unix.EMLINK, _, _)
-     | Unix.Unix_error (Unix.EUNKNOWNERR -1142, _, _) (* Needed for OCaml < 5.1 *) ->
-       (* If we can't make a new hard link because we reached the limit on the
-          number of hard links per file, we fall back to copying. We expect that
-          this happens very rarely (probably only for empty files). *)
-       copy_file ~src ~dst ())
+     | Unix.Unix_error (Unix.EMLINK, _, _) -> copy_file ~src ~dst ())
```
snowleopard, line 494: I think the above comment was useful here, could you restore it?

otherlibs/stdune/src/path.ml
rgrinberg: It's back. Accidentally deleted it.

otherlibs/stdune/src/path.ml
```diff
@@ -1130,7 +1130,16 @@ let is_directory t =
...
 let unlink t = Fpath.unlink (to_string t)
-let link x y = Unix.link (to_string x) (to_string y)
+
+let link src dst =
+  match Unix.link (to_string src) (to_string dst) with
+  | exception Unix.Unix_error (Unix.EUNKNOWNERR -1142, syscall, arg)
+  (* Needed for OCaml < 5.1 on windows *) ->
+    Exn.reraise (Unix.Unix_error (Unix.EMLINK, syscall, arg))
+  | s -> s
+  | exception e -> Exn.reraise e
```
nojb, line 1140: Why is this case needed?

rgrinberg: https://github.com/ocaml/dune/pull/8807

PR #8806, opened 2023-09-30
rpc status: fix for Win32

### LINKED ISSUE #8799: windows: dune rpc status not working
## Expected Behavior

`dune rpc status` should work correctly on windows.

## Actual Behavior

`duen rpc status` reports that there is no running RPC server. This is not true as ocamllsp is able to connect to it just fine.

## Reproduction

1. On Windows with dkml ocaml, start dune in watch mode in a project.
2. From another terminal do `dune rpc status`
3. Dune will complain nothing is running.

## Specifications

- Version of `dune` (output of `dune --version`): 3.10.0
- Operating system (distribution and version): Windows 11

nojb:
Beginning of a fix for #8799, but not complete because `dune rpc status` does not seem to be fully working:

image

(no data is reported for the connected client)

cc @Alizter

### DIFF
--- a/doc/changes/8806.md
+++ b/doc/changes/8806.md
@@ -1,1 +1,1 @@
-- Fix `dune rpc status` on Windows (#8806, fixes #8799, @nojb)
+- Fix `dune rpc` commands on Windows (#8806, fixes #8799, @nojb)
--- a/doc/changes/8806.md
+++ b/doc/changes/8806.md
@@ -1,0 +1,1 @@
+- Fix `dune rpc status` on Windows (#8806, fixes #8799, @nojb)
--- a/otherlibs/dune-rpc/private/where.ml
+++ b/otherlibs/dune-rpc/private/where.ml
@@ -151,20 +151,22 @@ module Make
          | Ok s -> Ok (Some s)
          | Error exn -> Error exn)
     | None ->
-      let of_file f =
-        let+ contents = IO.read_file f in
-        match contents with
-        | Error e -> Error e
-        | Ok contents ->
-          (match of_string contents with
+      (match default ~build_dir () with
+       | `Ip _ as x -> Fiber.return (Ok (Some x))
+       | `Unix file as x ->
+         let of_file f =
+           let+ contents = IO.read_file f in
+           match contents with
            | Error e -> Error e
-           | Ok s -> Ok (Some s))
-      in
-      let file = Filename.concat build_dir rpc_socket_relative_to_build_dir in
-      let** analyze = IO.analyze_path file in
-      (match analyze with
-       | `Other -> Fiber.return (Ok None)
-       | `Normal_file -> of_file file
-       | `Unix_socket -> Fiber.return (Ok (Some (`Unix file))))
+           | Ok contents ->
+             (match of_string contents with
+              | Error e -> Error e
+              | Ok s -> Ok (Some s))
+         in
+         let** analyze = IO.analyze_path file in
+         (match analyze with
+          | `Other -> Fiber.return (Ok None)
+          | `Normal_file -> of_file file
+          | `Unix_socket -> Fiber.return (Ok (Some x))))
   ;;
 end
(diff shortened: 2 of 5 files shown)


nojb:
After further testing, it seems to be working:

image

@Alizter, can you give it a try on your side to confirm? Thanks!

Alizter (approved):
Works great, thanks!

cc @emillon any chance we can get this backported for 3.9-3.11? Or at least add it to a list in case another point release is decided.

otherlibs/dune-rpc/private/where.ml
```diff
@@ -151,20 +151,22 @@ module Make
          | Ok s -> Ok (Some s)
          | Error exn -> Error exn)
     | None ->
-      let of_file f =
-        let+ contents = IO.read_file f in
-        match contents with
-        | Error e -> Error e
-        | Ok contents ->
-          (match of_string contents with
+      (match default ~build_dir () with
```
rgrinberg, line 154: I don't know what the issue is yet, but this fix doesn't seem right to me. Previously, the code would look at the socket inside the build directory to connect, and now it's just assuming the address is the default address.

rgrinberg (changes requested):

otherlibs/dune-rpc/private/where.ml
nojb: Nothing has changed here for Unix systems: `default ~build_dir ()` still returns the socket inside the build directory:

https://github.com/ocaml/dune/blob/071cb4718d4c23f9b3645c8c254028df6d1a35c2/otherlibs/dune-rpc/private/where.ml#L130-L134

rgrinberg: Okay, but even on Windows, the workflow to connect should be:

1. Inspect the file _build/.rpc/dune
2. If it's a socket, connect to it (will never happen on windows)
3. If it's a normal file, read it and get the connection string from here. On windows, that connection string will have the address we want to connect on.

nojb: Oops, you are right. I hadn't grasped the logic behind the client/server split. Now I pushed a commit with the right fix I think.

src/dune_rpc_client/where.ml
```diff
@@ -47,10 +47,7 @@ let to_socket = function
   | `Ip (`Host host, `Port port) -> Unix.ADDR_INET (Unix.inet_addr_of_string host, port)
 ;;
 
-let to_string = function
-  | `Unix p -> sprintf "unix://%s" p
-  | `Ip (`Host host, `Port port) -> sprintf "%s:%d" host port
-;;
+let to_string t = Dune_rpc_private.Where.to_string t
```
nojb, line 50: Use the right "dbus" format so that it can be parsed back in https://github.com/ocaml/dune/blob/071cb4718d4c23f9b3645c8c254028df6d1a35c2/otherlibs/dune-rpc/private/where.ml#L154-L162

doc/changes/8806.md
```diff
@@ -0,0 +1 @@
+- Fix `dune rpc status` on Windows (#8806, fixes #8799, @nojb)
```
rgrinberg, line 1: This fixes a little more than `status`. It should fix all `dune rpc` commands in fact.

rgrinberg (approved):

doc/changes/8806.md
Alizter: Looks like `monitor` too.

nojb: Thanks for the review. I amended the Changes entry.

emillon:
> Works great, thanks!
> 
> cc @emillon any chance we can get this backported for 3.9-3.11? Or at least add it to a list in case another point release is decided.

Sure. To be clear, this is a regression introduced in 3.9.0?

nojb:
> > Works great, thanks!
> > cc @emillon any chance we can get this backported for 3.9-3.11? Or at least add it to a list in case another point release is decided.
> 
> Sure. To be clear, this is a regression introduced in 3.9.0?

I suspect this has never worked on Windows, until this PR.

emillon:
OK. Note a huge rush then. I'll backport that to 3.11.1.

PR #8807, opened 2023-10-01
refactor: remove pointless exception catch-all

rgrinberg:
If we don't match on the exception, it will be re-raised anyway


nojb (approved):
LGTM

PR #8812, opened 2023-10-02
dune_console: make refresh rate configurable and default INSIDE_EMACS to 15

Alizter:
We make the refresh rate of the dune console configurable from dune_config. The default value is 60 frames per second like before but now we also default to 15 if `INSIDE_EMACS` is set. I've also validated the range between `0 < x <= 1000` which is arbitrary but I don't see any use case for such large frame rates. This is also very useful for testing dune console.

### DIFF
--- a/src/dune_config/config.ml
+++ b/src/dune_config/config.ml
@@ -168,3 +168,19 @@ let threaded_console =
   register t;
   t
 ;;
+
+let threaded_console_frames_per_second =
+  let t =
+    { name = "threaded_console_frames_per_second"
+    ; of_string =
+        (fun x ->
+          match Int.of_string x with
+          | Some x when x > 0 && x <= 1000 -> Ok (`Custom x)
+          | Some _ -> Error (sprintf "value must be between 1 and 1000")
+          | None -> Error (sprintf "could not parse %S as an integer" x))
+    ; value = `Default
+    }
+  in
+  register t;
+  t
+;;
--- a/src/dune_config_file/display.ml
+++ b/src/dune_config_file/display.ml
@@ -40,7 +40,9 @@ let console_backend = function
        Dune_console.Backend.dumb
      | true ->
        (match Config.(get threaded_console) with
-        | `Enabled -> Dune_threaded_console.progress ()
+        | `Enabled ->
+          Dune_threaded_console.progress
+            ~frames_per_second:(Dune_util.frames_per_second ())
         | `Disabled ->
           Dune_util.Terminal_signals.unblock ();
           Dune_console.Backend.progress))
--- a/src/dune_threaded_console/dune_threaded_console.mli
+++ b/src/dune_threaded_console/dune_threaded_console.mli
@@ -1,4 +1,11 @@
 include module type of Dune_threaded_console_intf
 
-val make : (module S) -> Dune_console.Backend.t
-val progress : unit -> Dune_console.Backend.t
+(** [make ~frames_per_second (module T)] is a backend that renders the user interface in a
+    separate thread. The module [T] must implement the [Threaded] interface. There are
+    special functions included to handle various functions of a user interface.
+
+    The [frames_per_second] argument controls how often the user interface is updated. *)
+val make : frames_per_second:int -> (module S) -> Dune_console.Backend.t
+
+(** Threaded variant of [Dune_console.Backend.progress]. *)
+val progress : frames_per_second:int -> Dune_console.Backend.t
--- a/src/dune_tui/dune_tui.ml
+++ b/src/dune_tui/dune_tui.ml
@@ -378,7 +378,12 @@ module Console_backend = struct
 end
 
 let backend =
-  let t = lazy (Dune_threaded_console.make (module Console_backend)) in
+  let t =
+    lazy
+      (Dune_threaded_console.make
+         ~frames_per_second:(Dune_util.frames_per_second ())
+         (module Console_backend))
+  in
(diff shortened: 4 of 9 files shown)


src/dune_config/config.ml
```diff
@@ -168,3 +168,24 @@ let threaded_console =
...
+    ; of_string =
+        (fun x ->
+          match Int.of_string x with
+          | Some x ->
+            (match x > 0 && x <= 1000 with
+             | true -> Ok x
+             | false -> Error (sprintf "value must be between 1 and 1000"))
+          | None -> Error (sprintf "could not parse %S as an integer" x))
+    ; value =
+        (match Env.get Env.initial "INSIDE_EMACS" with
```
rgrinberg, line 184: It's not good to have to copy paste this variable

Alizter: `INSIDE_EMACS` lives inside `dune_util` and I didn't want to have to depend on it.

rgrinberg: You can make `Dune_threaded_console.make` take this parameter and then thread it pass it from whoever is initializing the console.

Alizter: Done.

src/dune_threaded_console/dune_threaded_console.mli
```diff
@@ -1,4 +1,4 @@
 include module type of Dune_threaded_console_intf
 
-val make : (module S) -> Dune_console.Backend.t
-val progress : unit -> Dune_console.Backend.t
+val make : refresh_rate:int -> (module S) -> Dune_console.Backend.t
+val progress : refresh_rate:int -> Dune_console.Backend.t
```
rgrinberg, line 4: How about the name "frames_per_second"?

src/dune_config_file/display.ml
```diff
@@ -40,7 +40,12 @@ let console_backend = function
        Dune_console.Backend.dumb
      | true ->
        (match Config.(get threaded_console) with
-        | `Enabled -> Dune_threaded_console.progress ()
+        | `Enabled ->
+          Dune_threaded_console.progress
+            ~refresh_rate:
+              (match Dune_util.Execution_env.inside_emacs with
```
rgrinberg, line 46: Kind of weird that being inside emacs overrides an explicit setting.

Alizter: Not really sure how else to handle it. I'm not overly fond of these emacs specific behaviours to begin with.

rgrinberg: How about this:

1. if the config is set, do not override it
2. if it's not set, default to 60 outside of emacs. 15 inside of emacs.

Alizter: Done.

src/dune_tui/dune_tui.ml
```diff
@@ -378,7 +378,15 @@ module Console_backend = struct
...
 
 let backend =
-  let t = lazy (Dune_threaded_console.make (module Console_backend)) in
+  let t =
+    lazy
+      (Dune_threaded_console.make
+         ~frames_per_second:
+           (match Dune_config.Config.(get threaded_console_frames_per_second) with
+            | `Custom fps -> fps
+            | `Default -> 60)
```
rgrinberg, line 387: Why not respect the emacs config here as well?

Alizter: Initially because I was unsure if TUI would work inside emacs.

rgrinberg: Sure, but why create this inconsistency? I don't see what benefit it could have other than confusing people.

Alizter: Fixed.

rgrinberg:
Can you add a CHANGES entry?

src/dune_config_file/display.ml
```diff
@@ -40,7 +40,13 @@ let console_backend = function
        Dune_console.Backend.dumb
      | true ->
        (match Config.(get threaded_console) with
-        | `Enabled -> Dune_threaded_console.progress ()
+        | `Enabled ->
+          Dune_threaded_console.progress
+            ~frames_per_second:
+              (match Config.(get threaded_console_frames_per_second) with
```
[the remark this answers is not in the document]
Alizter: Not really sure where to put it. Any suggestions?

Alizter:
I've added a chnagelog entry. I don't really know what to do about the "copy pasted code" as I am not sure where it should live.

src/dune_config_file/display.ml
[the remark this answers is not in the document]
rgrinberg: in this very module, you could introduce a function called:

```
val frames_per_second : unit -> int
```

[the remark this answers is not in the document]
Alizter: But I can't use that from `Dune_tui` since this module depends on it. It would also not make much sense to put it in `Dune_tui`.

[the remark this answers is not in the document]
rgrinberg: Okay, we need to figure this out somehow. The compelling reason is that you're copy pasting code. And no, there's no coincidence, we should keep the fps consistent for all the UI's.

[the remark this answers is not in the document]
Alizter: Alright, moved to `dune_util` which seems to work.

Alizter:
@rgrinberg I've moved `frames_per_second` to `Dune_util` and now the code is shared between TUI and progress.

PR #8814, opened 2023-10-02
test: sandbox deletion

rgrinberg:
This test demosntrates that errors from deleting the sandbox do not
shadow rule errors.

### DIFF
--- a/test/blackbox-tests/test-cases/interfere-sandbox-deletion.t
+++ b/test/blackbox-tests/test-cases/interfere-sandbox-deletion.t
@@ -1,0 +1,32 @@
+Create a bad rule that doesn't allow the sandbox to be deleted.
+
+Note that this rule also fails. We make sure that we see both the error from
+the rule and the sandbox cleanup.
+
+  $ cat >dune <<EOF
+  > (rule
+  >  (target foo)
+  >  (action (system "\| touch foo && mkdir bar && touch bar/x && chmod -w bar &&
+  >                  "\| echo failed action && exit 1
+  >           )))
+  > EOF
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.11)
+  > EOF
+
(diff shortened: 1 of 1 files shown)


Alizter:
I've triggered this when making cram tests that remove read permissions from files.

Alizter:
@rgrinberg Doesn't look like `-R` is in posix `chmod`.

test/blackbox-tests/test-cases/interfere-sandbox-deletion.t
```diff
@@ -0,0 +1,33 @@
+Create a bad rule that doesn't allow the sandbox to be deleted.
+
+Note that this rule also fails. We make sure that we see both the error from
+the rule and the sandbox cleanup.
+
+  $ cat >dune <<EOF
+  > (rule
+  >  (target foo)
+  >  (action
+  >   (with-accepted-exit-codes 1
```
snowleopard, line 10: Why? Was this command supposed to fail too? (It doesn't appear to, at the moment.)

rgrinberg: Yes, the intention is for the command to produce the target _and_ fail. It also does fail, see `Command exited with code 0.` in the error message.

Alizter (approved):

test/blackbox-tests/test-cases/interfere-sandbox-deletion.t
snowleopard: Oh, I see. Why not `exit 1` instead? Seems way simpler.

snowleopard: (And maybe `echo` something to confirm the error gets captured and displayed.)

rgrinberg: Should be good now.

snowleopard (approved):
Thank! Looks great now.

PR #8815, opened 2023-10-02
refactor(pkg): use portable hardlinking

rgrinberg:

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -1047,12 +1047,6 @@ module Install_action = struct
       |> Section.Map.of_list_multi
     ;;
 
-    let hardlink_or_copy ~src ~dst =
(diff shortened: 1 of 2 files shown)


src/dune_rules/pkg_rules.ml
```diff
@@ -1089,7 +1083,8 @@ module Install_action = struct
       then None
       else (
         let dst = prepare_copy ~install_file ~target_dir entry in
-        hardlink_or_copy ~src ~dst;
+        (let src = Path.to_string src |> Unix.readlink |> Path.of_string in
```
Alizter, line 1086: So the src is whatever the existing symlink was pointing to? How do we know if this is valid?

rgrinberg: if it's not valid, we're going to fail installing regardless.

Alizter (approved):

PR #8816, opened 2023-10-02
refactor(cache): Make [Util.add_atomically] private

rgrinberg:
It doesn't seem to be used outside of [Util]

This needs confirmation that it isn't used anywhere upstream.


snowleopard (approved):
Looks good, thanks!

PR #8819, opened 2023-10-02
fix(pkg): move solver env printing to own command

rgrinberg:
Instead of [--just-print-solver-env], introduce [dune pkg
just-print-solver-env]

I didn't remove some of the duplication between the two commands,
because some of these duplicated options will be removed.

### DIFF
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -4,197 +4,159 @@ module Fetch = Dune_pkg.Fetch
 module Opam_repo = Dune_pkg.Opam_repo
 module Repository_id = Dune_pkg.Repository_id
 
-module Lock = struct
-  module Opam_repository_path = struct
-    let term =
-      let dune_path =
-        let parser s =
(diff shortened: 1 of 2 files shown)


bin/pkg.ml
```diff
@@ -288,6 +238,125 @@ module Lock = struct
           ])
   ;;
 
+  let print_solver_env
```
Leonidas-from-XIV, line 241: I think it might be better for readability not to shadow `print_solver_env`.

Leonidas-from-XIV (approved):
Come to think of it, it is a better UX.

Github does a terrible job displaying the diff but I trust this is mostly reshuffling, so I concentrated on the parts that I think are new. Looks good to me.

rgrinberg:
> Come to think of it, it is a better UX.

Almost always the case. Now printing the environment doesn't accept command line arguments it doesn't understand.

PR #8823, opened 2023-10-03
test(pkg): cycles in opam repo

Alizter:
The solver does not check for cyles. This might lead to an invalid build
plan so something to keep in mind.

We also add a test for a situation where there is a cycle in the repo,
but an alternative valid solution is still available.

In both tests the solver ignores the cycle.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/opam-package-cycle-with-or.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-package-cycle-with-or.t
@@ -1,0 +1,30 @@
+We test an repository with a package cycle that has an alternative solution which avoids a
+cycle. We have the following packages:
+
+>      a --> b --> c
+
+Now c depends either on a or on a fourth package d. Therefore there is a valid solution
+available that avoids a cycle.
+
+  $ . ./helpers.sh
+  $ mkrepo
+
+  $ mkpkg a <<EOF
+  > depends: [ "b" ]
+  > EOF
+  $ mkpkg b <<EOF
+  > depends: [ "c" ]
+  > EOF
+  $ mkpkg c <<EOF
+  > depends: [ "a" | "d" ]
+  > EOF
+  $ mkpkg d
+
+Solver finds the invalid solution as it doesn't check cycles.
+
+  $ solve c
+  Solution for dune.lock:
+  a.0.0.1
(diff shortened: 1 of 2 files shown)


test/blackbox-tests/test-cases/pkg/opam-package-cycle-with-or.t
```diff
@@ -0,0 +1,29 @@
...
+>      a --> b --> c
+
+Now c depends either on a or on a fourth package d. Therefore there is a valid solution
+available that avoids a cycle.
+
+  $ . ./helpers.sh
+  $ mkrepo
+
+  $ mkpkg a <<EOF
+  > depends: [ b ]
```
rgrinberg, line 13: Shouldn't it be `[ "b" ]`, otherwise opam interprets it as a variable?

Alizter: You're totally right. The actual behaviour seems to be to just ignore the cycle.

rgrinberg:
@Leonidas-from-XIV is that the expected behavior from 0install? Does opam do any cycle checking?

Leonidas-from-XIV:
I took a look and there is apparently some handling of `0install-solver/src/solver/solver_core.ml` so it looks like it should at least detect that the solver is back at the beginning of the loop. Possibly a bug?

I've also checked what happens in OPAM when you have cycles:
```
[ERROR] Package conflict!
The actions to process have cyclic dependencies:
  - ∗ a.1 → ∗ b.1 → ∗ a.1

No solution found, exiting
```

However in @Alizter's case, OPAM does successfully choose the solution without cycles:
```
The following actions will be performed:
  ∗ install d 1 [required by c]
  ∗ install c 1
===== ∗ 2 =====
```

I am however not sure that this test should live in Dune. It seems like the test that cyclic dependencies are resolved properly would better be suited to live in 0install-solver, since it would need to get solved there.

Alizter:
@Leonidas-from-XIV I think that this test should still live in Dune since we are using a solver (0install atm) to find packages and if we aren't doing the same as opam then that's probably something we should be aware of.

rgrinberg:
Even if this isn't our bug, it's a good idea to document this limitation in our test suite.

PR #8826, opened 2023-10-03
feature(pkg): add status line for solving step

rgrinberg:


Alizter (approved):
LGTM, we can probably improve this at a later point with an animation to indiciate that dune is still working.

PR #8827, opened 2023-10-03
refactor(pkg): move repo fetching to own function

rgrinberg:


Alizter:
I did something similar in #8773 but ended up not needing it so exploded.

Alizter (approved):

PR #8828, opened 2023-10-03
fix(pkg): improve locations of conversion errors

rgrinberg:
We don't have the exact locations, but we can at least point to the
relevant files.

### DIFF
--- a/src/dune_pkg/package_variable.ml
+++ b/src/dune_pkg/package_variable.ml
@@ -94,15 +94,10 @@ let of_opam_ident ident =
         Result.map variable ~f:(fun var -> `Package_variable (self_scoped var)))
 ;;
 
-let pform_of_opam_ident ~package_name ident =
+let pform_of_opam_ident (loc, ident) =
   match of_opam_ident ident with
   | Ok (`Package_variable t) -> to_pform t
   | Ok (`Global_variable var) -> Pform.Var var
   | Error (`Unsupported_variable name) ->
-    User_error.raise
-      [ Pp.textf
-          "Variable %S occuring in opam package %S is not supported."
-          name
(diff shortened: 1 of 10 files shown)


Alizter:
Can you add a test so we can see what it looks like?

src/dune_pkg/package_variable.ml
```diff
@@ -94,15 +94,10 @@ let of_opam_ident ident =
...
   | Ok (`Package_variable t) -> to_pform t
   | Ok (`Global_variable var) -> Pform.Var var
   | Error (`Unsupported_variable name) ->
-    User_error.raise
-      [ Pp.textf
-          "Variable %S occuring in opam package %S is not supported."
-          name
-          package_name
-      ]
+    User_error.raise ~loc [ Pp.textf "Variable %S occuring is not supported." name ]
```
Alizter, line 102: What about "Occurance of variable %S is not supported."?

Reads a bit better IMO.

rgrinberg: I just removed the word "occuring"

The addition of "Occurance of" sounds redundant to me.

rgrinberg:
ping @Leonidas-from-XIV

Alizter:
I still think it might be good to provide the package name and version in the message and not just the location.

rgrinberg:
How would that information help? It's irrelevant to the error itself, therefore distracting. More is not better when it comes to error messages.

Leonidas-from-XIV (approved):
Looks good. The only concern I have is that the filename to the offending file might not be a particularly useful one, especially if it is read from a checkout in a cache directory or from a git object.

But lets not get future concerns prevent current improvements.

PR #8831, opened 2023-10-03
fix(pkg): Make the repos use `Ordered_set_lang`

Leonidas-from-XIV:
As mentioned in the previous PR on multiple OPAM repos, the list should use `Ordered_set_lang` as it gets support for set operations and more importantly the definition of `:standard` for free.

This fixes also the support for downloading multiple OPAM repos and adds a test that this works as desired.

### DIFF
--- a/src/dune_pkg/fetch.ml
+++ b/src/dune_pkg/fetch.ml
@@ -135,10 +135,16 @@ let fetch ~unpack ~checksum ~target (url : OpamUrl.t) =
 ;;
 
 module Opam_repository = struct
-  type t = { url : OpamUrl.t }
+  type t = Workspace.Repository.t
 
-  let of_url url = { url }
-  let default = of_url @@ OpamUrl.of_string "https://opam.ocaml.org/index.tar.gz"
+  let of_url url =
+    let name = Workspace.Repository.Name.of_string "unknown" in
+    let source = OpamUrl.to_string url in
+    Workspace.Repository.create ~name ~source
+  ;;
+
+  let of_workspace_repo v = v
+  let default = Workspace.Repository.default
 
   let is_archive name =
     List.exists
@@ -154,11 +160,19 @@ module Opam_repository = struct
   let path =
     let open Fiber.O in
     let ( / ) = Filename.concat in
-    fun { url } ->
+    fun repo ->
+      let url = Workspace.Repository.opam_url repo in
+      let name =
+        repo |> Workspace.Repository.name |> Workspace.Repository.Name.to_string
+      in
+      let path_digest =
+        url |> OpamUrl.to_string |> Dune_digest.string |> Dune_digest.to_string
+      in
+      let folder = sprintf "%s-%s" name path_digest in
       Fiber.of_thunk
       @@ fun () ->
       let target_dir =
-        Xdg.cache_dir (Lazy.force Dune_util.xdg) / "dune/opam-repository"
+        Xdg.cache_dir (Lazy.force Dune_util.xdg) / "dune/opam-repositories" / folder
       in
       let target = target_dir |> Path.External.of_string |> Path.external_ in
       let unpack = url |> OpamUrl.to_string |> is_archive in
--- a/src/dune_pkg/solver_env.ml
+++ b/src/dune_pkg/solver_env.ml
@@ -206,8 +206,16 @@ let default =
   { flags = Variable.Flag.Set.all
   ; sys = Variable.Sys.Bindings.empty
   ; const = Variable.Const.bindings
-  ; repos = [ Workspace.Repository.Name.of_string ":standard" ] (* TODO *)
+  ; repos = [ Workspace.Repository.Name.of_string "default" ]
   }
+;;
+
+let repos_of_ordered_set ordered_set =
+  Dune_lang.Ordered_set_lang.eval
+    ordered_set
+    ~parse:(fun ~loc string -> Workspace.Repository.Name.parse_string_exn (loc, string))
+    ~eq:Workspace.Repository.Name.equal
+    ~standard:default.repos
 ;;
 
 let decode =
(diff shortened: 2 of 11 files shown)


test/blackbox-tests/test-cases/pkg/multiple-opam-repos.t
```diff
@@ -0,0 +1,134 @@
...
+  $ git add -A
+  $ git commit -m "Initial commit" > /dev/null
+  $ cd ..
+
+We have to define both repositories in the workspace, but will only use `new`.
+
+  $ cat > dune-workspace <<EOF
+  > (lang dune 3.10)
+  > (repository
+  >   (name new)
```
rgrinberg, line 33: Single space is our standard indentation.

src/dune_pkg/solver_env.ml
```diff
@@ -224,19 +224,26 @@ let default =
...
   ; const = Variable.Const.bindings
-  ; repos = [ Workspace.Repository.Name.of_string ":standard" ] (* TODO *)
+  ; repos = [ Workspace.Repository.Name.of_string ":standard" ]
   }
 ;;
 
+let repos_of_ordered_set ordered_set =
+  Dune_lang.Ordered_set_lang.eval
+    ordered_set
+    ~parse:(fun ~loc:_ -> Workspace.Repository.Name.of_string)
```
rgrinberg, line 234: I think you can use `parse_string_exn` here to make use of the location.

Leonidas-from-XIV: Ah cool, that exists indeed. Bit of a pity it takes a tuple but then again the currying made it harder to read than it needed to.

src/dune_pkg/solver_env.ml
```diff
@@ -224,19 +224,26 @@ let default =
   { flags = Variable.Flag.Set.all
   ; sys = Variable.Sys.Bindings.empty
   ; const = Variable.Const.bindings
-  ; repos = [ Workspace.Repository.Name.of_string ":standard" ] (* TODO *)
+  ; repos = [ Workspace.Repository.Name.of_string ":standard" ]
```
rgrinberg, line 227: How can the name of a repo be `:standard`? I imagine this should be something like `default`? The same way we have a "default" context.

test/blackbox-tests/test-cases/pkg/multiple-opam-repos.t
```diff
@@ -0,0 +1,134 @@
+We want to test that support for multiple opam repositories works.
+
+  $ . ./helpers.sh
+  $ mkrepo
+  $ mkpkg foo 1.0 <<EOF
+  > EOF
+  $ cd mock-opam-repository
+  $ git init --quiet
```
Alizter, line 8: You should add `%{bin:git}` to the cram stanza.

src/dune_pkg/solver_env.ml
Alizter: Shouldn't it be the ":standard" set of repositories rather than a repo itself?

src/dune_pkg/fetch.ml
```diff
@@ -135,10 +135,16 @@ let fetch ~unpack ~checksum ~target (url : OpamUrl.t) =
 ;;
 
 module Opam_repository = struct
-  type t = { url : OpamUrl.t }
+  type t = Workspace.Repository.t
 
-  let of_url url = { url }
-  let default = of_url @@ OpamUrl.of_string "https://opam.ocaml.org/index.tar.gz"
+  let of_url url =
+    let name = Workspace.Repository.Name.of_string "unknown" in
```
Alizter, line 141: What happen when multiple repos become "unknown"? Won't their download directories overlap?

gridbugs (approved):

src/dune_pkg/fetch.ml
Leonidas-from-XIV: Yes, and it would create an issue. Currently the only place where this happens is when a repo is specified via command line, so maybe it would make sense to hash the path instead of using the name. Downside is that the repo names in the folder become strange hex character strings but it might be a better solution overall.

src/dune_pkg/solver_env.ml
Leonidas-from-XIV: It's because the name of a repository is just a string and I predefined a repository that's called `:standard` (that's how it worked before the Ordered set lang so if someone specified `:standard` it would pick just that). I can rename it into `default` now that we are using Ordered set lang.

Leonidas-from-XIV: > Shouldn't it be the ":standard" set of repositories rather than a repo itself?

This is the ordered set of desired repos, the joining with the list of repos comes later (especially since this is `solver_env`. `repositories` lives in a different stanza I don't think joining it here makes it more readable).

src/dune_pkg/fetch.ml
Leonidas-from-XIV: I'm now building `name-digest(source)` directories, so this should avoid this case at the expense of having the same repo cached multiple time with the same name, but I don't think it is a big risk.

src/dune_pkg/solver_env.ml
Leonidas-from-XIV: Renamed the repo to `default`.

src/dune_pkg/fetch.ml
```diff
@@ -154,11 +160,17 @@ module Opam_repository = struct
...
     let open Fiber.O in
     let ( / ) = Filename.concat in
-    fun { url } ->
+    fun repo ->
+      let url = Workspace.Repository.opam_url repo in
+      let name =
+        repo |> Workspace.Repository.name |> Workspace.Repository.Name.to_string
+      in
+      let path_digest = url |> OpamUrl.to_string |> Dune_digest.string |> Dune_digest.to_string in
+      let folder = Printf.sprintf "%s-%s" name path_digest in
```
rgrinberg, line 169: we don't need to qualify `sprintf` it's already in `Stdune`.

rgrinberg (approved):
Needs to be formatted

Leonidas-from-XIV:
Reformatted and rebased.

PR #8841, opened 2023-10-04
feature(pkg): dune describe pkg lock

### LINKED ISSUE #8031: Print All Outdated Packages

### LINKED ISSUE #8833: Make `dune describe` work on lock directories
`$ dune describe pkg lock dune.lock` should give us a nice pretty printed list of packages inside `lock.pkg`. The list should also include their versions and perhaps differentiate transitive vs immediate dependencies somehow.

Alizter:
We add a new `dune describe` command that displays the packages in a lock file. In order to do this, we refactored pkg.ml to share its printing function.

- fix https://github.com/ocaml/dune/issues/8833

### DIFF
--- a/bin/describe/describe_pkg.ml
+++ b/bin/describe/describe_pkg.ml
@@ -24,7 +24,7 @@ module Lock = struct
       let lock_dir = Lock_dir.read_disk lock_dir_path in
       Pp.concat
         ~sep:Pp.space
-        [ Pp.hbox
+        [ Pp.hovbox
           @@ Pp.textf "Contents of %s:" (Path.Source.to_string_maybe_quoted lock_dir_path)
         ; Pkg.Lock.pp_packages lock_dir.packages
         ]
--- a/bin/describe/describe_pkg.ml
+++ b/bin/describe/describe_pkg.ml
@@ -27,7 +27,8 @@ module Lock = struct
         [ Pp.hbox
           @@ Pp.textf "Contents of %s:" (Path.Source.to_string_maybe_quoted lock_dir_path)
         ; Pkg.Lock.pp_packages lock_dir.packages
-        ])
+        ]
+      |> Pp.vbox)
   ;;
 
   let command =
--- a/bin/describe/describe_pkg.ml
+++ b/bin/describe/describe_pkg.ml
@@ -23,10 +23,10 @@ module Lock = struct
     @@ List.map lock_dir_paths ~f:(fun lock_dir_path ->
       let lock_dir = Lock_dir.read_disk lock_dir_path in
       Pp.concat
-        ~sep:Pp.newline
+        ~sep:Pp.space
         [ Pp.hbox
           @@ Pp.textf "Contents of %s:" (Path.Source.to_string_maybe_quoted lock_dir_path)
-        ; Pkg.Lock.pp_packages lock_dir |> Pp.enumerate ~f:Fun.id
+        ; Pkg.Lock.pp_packages lock_dir.packages
         ])
   ;;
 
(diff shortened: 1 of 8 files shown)


bin/describe/describe_pkg.ml
```diff
@@ -0,0 +1,41 @@
...
+             directory specified in the context."
+          ~docv:"LOCKDIRS"
+          []
+    in
+    let lock_dir_paths =
+      match List.map lock_dir_paths ~f:Path.Source.of_string with
+      | [] -> [ Lock_dir.default_path ]
+      | lock_dir_paths -> lock_dir_paths
+    in
+    List.iteri lock_dir_paths ~f:(fun index lock_dir_path ->
```
rgrinberg, line 22: As a rule  you should use `Pp.enumerate` or `Pp.concat` to print a list of things.

bin/describe/describe_pkg.ml
```diff
@@ -0,0 +1,41 @@
...
+      let lock_dir = Lock_dir.read_disk lock_dir_path in
+      Console.print
+        ((Pp.textf "Contents of %s:" (Path.Source.to_string_maybe_quoted lock_dir_path)
+          :: Lock_dir.pp_packages lock_dir)
+         @ if index < List.length lock_dir_paths - 1 then [ Pp.nop ] else []))
```
rgrinberg, line 27: If you use the advice above, then this index check will be unnecessary

src/dune_pkg/lock_dir.mli
```diff
@@ -52,6 +52,7 @@ type t =
 val remove_locs : t -> t
 val equal : t -> t -> bool
 val to_dyn : t -> Dyn.t
+val pp_packages : t -> 'a Pp.t list
```
rgrinberg, line 55: Can you do without this function for now?

1. The type of this function is wrong. `pp_packages` should return `Pp.t` based on the name.
2. We don't need pretty printing functions like this in the dune_pkg API

bin/describe/describe_pkg.ml
Alizter: Sure, but this will require changing how we print solutions to the lock file. I purposefully avoided this to keep changes to a minimum.

rgrinberg: I don't see how using `Pp.concat_map` here will need changes anywhere else in the code.

Alizter: Sorry, I misunderstood the location.

Alizter:
@rgrinberg This is ready for another round of review. I will follow this PR up with another that improves the output of `dune pkg lock` since we can use Pp.enumerate and some color there.

bin/pkg.mli
```diff
@@ -1,3 +1,9 @@
 open Import
 
+module Lock : sig
+  (** [pp_packages lock_dir] returns a list of pretty-printed packages
+      occuring in [lock_dir]. *)
+  val pp_packages : Dune_pkg.Lock_dir.t -> 'a Pp.t list
```
rgrinberg, line 6: This function shouldn't exist in the public api.

rgrinberg (changes requested):

bin/pkg.mli
Alizter: This isn't accessible since it's in the executable.

rgrinberg: Okay, well it's type should be `Pp.t` then.

rgrinberg:
You can introduce `Pp.enumerate` in a subsequent PR.

PR #8842, opened 2023-10-04
chore(pkg): stop using Fmt

Alizter:
@rgrinberg I don't know if this is what you had in mind.


rgrinberg:
Yeah, that's what I had in mind.

PR #8847, opened 2023-10-05
Solver env in lockdir metadata

gridbugs:


src/dune_pkg/solver_env.ml
```diff
@@ -54,6 +54,18 @@ module Variable = struct
...
+      let encode t =
+        Dune_sexp.List
+          (to_list t
+           |> List.map ~f:(fun flag ->
+             (* Note that flags are quoted to make the output a valid Ordered
+                Set Language expression so that [encode] and [decode] can round
+                trip. Technically only the first flag needs to be quoted to
+                form a valid OSL expression but all are quoted for
+                consistency.*)
+             Dune_sexp.Quoted_string (to_string flag)))
```
Leonidas-from-XIV, line 66: I wonder whether it would make sense to implement `encode` on `Ordered_set_lang` to be able to write the (unevaluated) expression back to the lockfile.

On one hand it is nice to see what exactly the OSL has evaluated to, but on the other hand there is a not so nice part where we expose `:standard` and the expression in the lockfile doesn't match the expression from the dune-workspace.

This is very much a thinking out loud thing, let me know if you have any arguments for or against either of these behaviours.

Leonidas-from-XIV (commented):
Looks sensible. I think it's worth discussing what the behavior of the serialized OSL should be.

Alizter:
I think I missed the motivation for encoding the solver env in the lockdir. Can you remind us?

src/dune_pkg/solver_env.ml
rgrinberg: `:standard` is specific to the version of the dune language (used in the workspace file). While the lock directory should be stable wrt to its own version. For that reason, we should evaluate it.

src/dune_pkg/solver_env.ml
```diff
@@ -131,6 +143,13 @@ module Variable = struct
             ]
       ;;
 
+      let encode t =
+        let open Encoder in
+        Map.to_list t
```
rgrinberg, line 148: `Map.to_list_map`

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
```diff
@@ -119,6 +127,14 @@ Run the solver again preferring oldest versions of dependencies:
    (complete false)
    (used))
   
+  (solver_env
+   (flags
+    ("with-doc" "with-test"))
```
rgrinberg, line 132: do these need quoting?

test/blackbox-tests/test-cases/pkg/opam-generate-ocaml-package.t
```diff
@@ -18,3 +18,11 @@
   (repositories
    (complete false)
    (used))
+  
+  (solver_env
+   (flags
+    ("with-doc" "with-test"))
+   (sys)
```
rgrinberg, line 25: If it's empty, we might as well omit it.

rgrinberg:
Let's focus on talking about the implementation in this PR. If you want to discuss the motivation, go to the original issue.

src/dune_pkg/solver_env.ml
rgrinberg, line 66: Should be `atom_or_quoted_string` to avoid the quotes.

src/dune_pkg/solver_env.ml
```diff
@@ -168,9 +198,39 @@ module Variable = struct
...
+      (* Tests that each field matches the constant value for that field. This
+         is included so that [encode] and [decode] can round trip. *)
+      let decode =
+        let open Decoder in
+        let+ loc, opam_version =
+          fields
+          @@ located
+          @@ field Fields.opam_version ~default:Values.opam_version string
+        in
+        if not (String.equal opam_version Values.opam_version)
```
rgrinberg, line 216: Why do we have this check?

src/dune_pkg/solver_env.ml
gridbugs: I think it's ok that the expression in the lockfile doesn't match the expression in the workspace file. This is really just so human readers can get a sense for the conditions under which the lockdir was generated.

test/blackbox-tests/test-cases/pkg/opam-generate-ocaml-package.t
gridbugs: I thought about removing empty sections but figured it would be useful to tell users explicitly that no system variables were set. Otherwise in such a case they might think that system variables aren't supposed to show up here.

src/dune_pkg/solver_env.ml
gridbugs: The quotes are necessary because the first element of an OSL expressions must be quoted if it's a string literal (see the comment above this line).

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
gridbugs: Yes because it's an OSL and so if the first term is a string literal it must be quoted. Subsequent values are quoted for consistency.

src/dune_pkg/solver_env.ml
gridbugs: It's so that users can't encode an env, change a constant field's value, then decode an env with a different value to the contsant's value.

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
rgrinberg: But why is it OSL? I expected that we would expand the OSL to an actual ordered set when writing the flags here.

src/dune_pkg/solver_env.ml
rgrinberg: Hmm, but I think all this ends up doing is reducing compatibility. Suppose that we update this version in dune because we updated our version of opam. Now we can't build any old build plans even though we really should.

I need to think about this a little more.

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
gridbugs: But then `encode` and `decode` wouldn't round-trip, and the syntax for the solver env in the lockdir won't match the syntax used for the solver env in the workspace config.

src/dune_pkg/solver_env.ml
gridbugs: Ah that's a good point, but I think allowing arbitrary values here is also not the right solution. Do you see a use-case where this data needs to be parsed back into a `Solver_env.t`? I think we're mainly including it here for human-readability. In that case could generate the sovler env in the lockfile with something other than `Solver_env.encode` and not worry about needing to decode it. This would also mean we aren't bound by OSL syntax when serializing the flags field.

gridbugs:
@rgrinberg and I discussed this offline. We can't think of a reason to store variables that weren't used in the solve in the lockdir metadata. I'll close this PR and make a new one that instruments the solver to keep track of the variables used in the solve, and record just those variables in the metadata.

PR #8849, opened 2023-10-05
Fix inline_tests action when partition list is empty

### LINKED ISSUE #8848: Inline_tests with empty partition
When no inline test is declared (aka empty partition list) in a library using (inline_tests), dune complains with
```
- Error: Rule dependencies are configured to require sandboxing, but the rule
- has no actions that could potentially require sandboxing.
```

hhugo:
fix #8848 

This is a 3 lines fix if your ignore white space

### DIFF
--- a/src/dune_rules/inline_tests.ml
+++ b/src/dune_rules/inline_tests.ml
@@ -347,15 +347,18 @@ include Sub_system.Register_end_point (struct
                List.map partitions_flags ~f:(fun p -> action mode (flags p))
                |> Action_builder.all
              and+ () = Action_builder.paths source_files in
-             let run_tests = Action.concurrent actions in
-             let diffs =
-               List.map source_files ~f:(fun fn ->
-                 Path.as_in_build_dir_exn fn
-                 |> Path.Build.extend_basename ~suffix:".corrected"
-                 |> Action.diff ~optional:true fn)
-               |> Action.concurrent
-             in
-             Action.Full.make ~sandbox @@ Action.progn [ run_tests; diffs ]))
+             match actions with
+             | [] -> Action.Full.empty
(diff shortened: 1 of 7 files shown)


src/dune_rules/inline_tests.ml
```diff
@@ -347,15 +347,18 @@ include Sub_system.Register_end_point (struct
...
+             | _ :: _ ->
+               let run_tests = Action.concurrent actions in
+               let diffs =
+                 List.map source_files ~f:(fun fn ->
+                   Path.as_in_build_dir_exn fn
+                   |> Path.Build.extend_basename ~suffix:".corrected"
+                   |> Action.diff ~optional:true fn)
+                 |> Action.concurrent
+               in
+               Action.Full.make ~sandbox @@ Action.progn [ run_tests; diffs ]))
```
Alizter, line 361: Doesn't have to be in this PR, but wouldn't it make more sense to concurrently run and diff rather than concurrently run and then concurrently diff? At the moment, all the diffs have to wait for all the tests to run.

hhugo: There is nothing allowing us to know what diff to perform after a given run

Alizter: Oh yes of course.

Alizter (approved):
Fix looks good to me. Of course, we wouldn't need it if `Action.Full.make` was smart enough to avoid sandboxing if an action doesn't need it rather than complaining.

rgrinberg:
Thanks. a CHANGES entry would be appropriate

hhugo:
> a CHANGES entry would be appropriate

done

rgrinberg:
DCO please :)

PR #8852, opened 2023-10-05
feature(pkg): include local test deps

rgrinberg:
This will include all the dependencies marked with `:with-test` from the packages inside the workspace.

For now, there's no way to turn this off, but at least this is the better default.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/common-filters-deps.t
+++ b/test/blackbox-tests/test-cases/pkg/common-filters-deps.t
@@ -1,0 +1,23 @@
+Demonstrate that local dependencies that are marked as {with-test} can be
+included.
+
+  $ . ./helpers.sh
+  $ mkrepo
+
+  $ mkpkg post <<EOF
+  > EOF
+  $ mkpkg build <<EOF
+  > EOF
+  $ mkpkg dev <<EOF
+  > EOF
+  $ mkpkg test <<EOF
+  > EOF
+  $ mkpkg doc <<EOF
+  > EOF
+
+  $ solve "(test :with-test) (doc :with-doc) (dev :with-dev) (build :build) (post :post)"
+  Solution for dune.lock:
+  build.0.0.1
+  doc.0.0.1
+  test.0.0.1
+  


Alizter:
I tried locking the dune repo and it found a solution, however when building it seems the `dune` package itself is missing. Was this done on purpose?

rgrinberg:
Yeah, it's expected. I have a PR to fix it that will come after this.

Alizter (approved):

src/dune_pkg/opam_solver.ml
```diff
@@ -15,10 +15,11 @@ module Context_for_dune = struct
     ; version_preference : Version_preference.t
     ; local_packages : OpamFile.OPAM.t OpamPackage.Name.Map.t
     ; solver_env : Solver_env.t
+    ; enable_tests : bool
     }
 
-  let create ~solver_env ~repos ~local_packages ~version_preference =
-    { repos; version_preference; local_packages; solver_env }
+  let create ~solver_env ~repos ~local_packages ~version_preference ~enable_tests =
+    { repos; version_preference; local_packages; solver_env; enable_tests }
```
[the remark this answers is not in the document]
rgrinberg: Good point. I've changed the default value and also used the flags from the Solver_env for filtering.

src/dune_pkg/opam_solver.ml
```diff
@@ -132,27 +132,21 @@ module Context_for_dune = struct
...
     in
     map_filters filtered_formula ~f:(resolve_solver_env solver_env)
-    |> OpamFilter.filter_deps
-         ~build:true
-         ~post:false
-         ~dev:false
-         ~default:false
-         ~test:false
-         ~doc:false
+    |> OpamFilter.filter_deps ~build:true ~post:false ~dev:false ~default:false ~test ~doc
```
gridbugs, line 149: I'm surprised it's necessary to explicitly pass `~test` and `~doc` here. The `resolve_solver_env` function should have already substituted all occurrences of `with-test` and `with-doc` with `true` or `false` so those variables shouldn't be present in the filters passed to `filter_deps` here. Does the behaviour change if you leave `~test` and `~doc` as `false`?

rgrinberg: You're right. It doesn't do anything. I wonder why we need the `filter_deps` step at all then. We could include build, dev, etc all in the Solver_env.t

rgrinberg: I changed the PR to just include the tests.

PR #8858, opened 2023-10-05
Add dkml-workflows to TEST_DEPS

gridbugs:
When setting up the dune codebase on a new machine it's recommended to run `make dev-deps` to install development dependencies but you also need to install the `dkml-workflows` package in order to run (say) `dune build` in the root of this project. This change adds `dkml-workflows` to the packages installed with `make dev-deps` to remove the need to manually install it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,8 @@ ppxlib \
 ctypes \
 "utop>=2.6.0" \
 "melange>=1.0.0" \
-"rescript-syntax"
+"rescript-syntax" \
+"dkml-workflows"
 # Dependencies recommended for developing dune locally,
 # but not wanted in CI
 DEV_DEPS := \


Alizter:
Is this only for windows? Also I don't think it needs a changelog since it is not user facing.

gridbugs:
I bootstrapped dune on a mac the other day and I got a bunch of errors about dkml not being installed when running `make test` but I can't reproduce it today so I'll close this.

gridbugs:
Ah actually I can reproduce the problem on macos and linux by running `dune build` in the dune repo. The output is:
```
File "ci/setup-dkml/pc/dune", line 43, characters 8-21:
43 |    (run pc-setup-dkml --output-darwin_x86_64 %{target}))))
             ^^^^^^^^^^^^^
Error: Program pc-setup-dkml not found in the tree or in PATH
 (context: default)
File "ci/setup-dkml/pc/dune", line 58, characters 8-21:
58 |    (run pc-setup-dkml --output-linux_x86 %{target}))))
             ^^^^^^^^^^^^^
Error: Program pc-setup-dkml not found in the tree or in PATH
 (context: default)
File "ci/setup-dkml/pc/dune", line 73, characters 8-21:
73 |    (run pc-setup-dkml --output-linux_x86_64 %{target}))))
             ^^^^^^^^^^^^^
Error: Program pc-setup-dkml not found in the tree or in PATH
 (context: default)
File "ci/setup-dkml/pc/dune", line 13, characters 8-21:
13 |    (run pc-setup-dkml --output-windows_x86 %{target}))))
             ^^^^^^^^^^^^^
Error: Program pc-setup-dkml not found in the tree or in PATH
 (context: default)
File "ci/setup-dkml/pc/dune", line 28, characters 8-21:
28 |    (run pc-setup-dkml --output-windows_x86_64 %{target}))))
             ^^^^^^^^^^^^^
Error: Program pc-setup-dkml not found in the tree or in PATH
 (context: default)
File "ci/setup-dkml/gh-windows/post/dune", line 12, characters 8-26:
12 |    (run gh-dkml-action-yml --phase post --output-windows %{target}))))
             ^^^^^^^^^^^^^^^^^^
Error: Program gh-dkml-action-yml not found in the tree or in PATH
 (context: default)
File "ci/setup-dkml/gh-windows/pre/dune", line 12, characters 8-26:
12 |    (run gh-dkml-action-yml --phase pre --output-windows %{target}))))
```

Installing `dkml-workflows` solves the problem.

gridbugs:
Removed the changelog file.

Alizter:
@gridbugs What is strange is that these are specific targets for windows.

gridbugs:
Right, maybe a better solution would be to only enable those targets on windows then.

Alizter:
Another question is why do you want to do `dune build` on its own? Typically something like `dune build @runtest` or `dune build @check` is chosen instead.

gridbugs:
My typical workflow is to run `dune build -w` in one terminal and edit code in another terminal. Should I be running something else instead?

Alizter:
@gridbugs running `dune build @check` is faster since it only does bytecode compilation and some merlin stuff. I typically do `dune build @check -w` in my terminal whilst editing elsewhere. If I am running tests then I do `dune build @name_of_cram_test -w` or `dune build @directory/to/tests -w`. `--auto-promote` is also convenient in those cases.

These are just my own recommendations since `dune build` is trying to build `@@default` and its not always clear what that exactly is in a given situation.

The added benefit is that you won't accidentally run unrelated stuff like the dkml helpers.

gridbugs:
Ok that makes sense, thanks for the tip. I'll run `dune build @check -w` from now on. I see that the `@default` alias builds everything by default but that means that building the default alias is usually not what you want to do (at least in this repo). Since running "dune build" feels like the obvious first thing to try when attempting to build a project would it make sense to change the default to something more likely to match the user's intent?

Also reflecting on the statement:

> Typically something like dune build @runtest or dune build @check is chosen instead

If this is true then I question why we even have a default alias. If one usually specifies a specific target then having a default will trick people into thinking that the default is what they usually want (at least that's what happened to me). Maybe in dune 4 we should make `dune build` require a target and running it without a target could print a list of targets or something. That way it would quickly teach people who think `dune build` that 1) dune has a concept of aliases and 2) which alias they probably want to use.

Alizter:
@rgrinberg can perhaps say more but the entire reason we have a `@default` alias is because users asked for it. It is also configurable.

The reason we don't use a `@default` alias for Dune itself is that it is not immediately clear what should be the default. Do we want to run tests? Do we want to build all the libraries? AFAIK there is no good answer to this that will satisfy everybody.

I should note that I don't have an aversion to using `dune build` on its own. For my other projects I use it all the time, but there it is immediately more clear what should be built. For Dune itself, there are just far too many targets, and unless you have a beefy computer you're going to be wasting a lot of time building stuff you don't need.

PR #8862, opened 2023-10-06
fix(pkg): workaround the ocaml configure script not accepting relative paths

rgrinberg:
Absolutize the %{prefix} and package section paths for now.

This requires us to disable sandboxing, because the absolute paths
always point outside the sandbox.

This also makes the dune cache kind of useless for packages, but this
isn't an issue as it doesn't work anyway.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/patch.t
+++ b/test/blackbox-tests/test-cases/pkg/patch.t
@@ -29,5 +29,4 @@ Applying patches
 Demonstrate that the original source shouldn't be modified:
 
   $ cat _build/_private/default/.pkg/test/source/foo.ml
-  cat: _build/_private/default/.pkg/test/source/foo.ml: No such file or directory
-  [1]
+  Hello World
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -486,6 +486,14 @@ module Action_expander = struct
         [ Value.String "" ]
     ;;
 
+    let encode_path path =
+      (* This hack is needed because the ocaml configure script requires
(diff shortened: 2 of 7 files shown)


test/blackbox-tests/test-cases/pkg/patch.t
```diff
@@ -29,5 +29,4 @@ Applying patches
 Demonstrate that the original source shouldn't be modified:
 
   $ cat _build/_private/default/.pkg/test/source/foo.ml
-  cat: _build/_private/default/.pkg/test/source/foo.ml: No such file or directory
-  [1]
+  Hello World
```
Alizter, line 32: I don't understand why this has changed.

Alizter:
I ran this on a package:
```
configure: creating cache config.cache
configure: Configuring OCaml version 5.2.0+dev0-2023-04-11
checking build system type... x86_64-pc-linux-gnu
checking host system type... i386-apple-darwin13.2.0
checking target system type... i386-apple-darwin13.2.0
checking for i386-apple-darwin13.2.0-ld... no
checking for i386-apple-darwin13.2.0-link... no
```
I have no idea why, but I am getting detected as Darwin?!

Alizter:
OK the configure script does run, but we have an issue with how the target is declared:
```
Error: Is a directory
-> required by _build/_private/default/.pkg/ocaml-base-compiler/source
-> required by _build/_private/default/.pkg/ocaml-base-compiler/target/cookie
```
I'm guessing something here needs to be a directory target and isn't declared as such.

rgrinberg:
A better version of this hack that doesn't break sandboxing is available here: https://github.com/ocaml/dune/pull/8921

PR #8865, opened 2023-10-06
refactor(pkg): better bindings for action expansion

rgrinberg:

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -673,21 +673,23 @@ module Action_expander = struct
     let dir = Path.build expander.paths.source_dir in
     match action with
     | Run (exe, args) ->
-      let* exe, more_args = Expander.expand_exe expander exe in
-      let+ args = Memo.parallel_map args ~f:(Expander.expand_sw expander) in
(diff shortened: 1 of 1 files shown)


src/dune_rules/pkg_rules.ml
```diff
@@ -673,21 +673,23 @@ module Action_expander = struct
...
       Action.Run (exe, Array.Immutable.of_list args)
     | Progn t ->
       let+ args = Memo.parallel_map t ~f:(expand ~expander) in
       Action.Progn args
     | System arg ->
-      let+ arg = Expander.expand_pform_gen ~mode:Single expander arg in
-      let arg = arg |> Value.to_string ~dir in
+      let+ arg =
+        Expander.expand_pform_gen ~mode:Single expander arg >>| Value.to_string ~dir
+      in
```
Alizter, line 686: Why is this better? It seems we now have two maps rather than just the one from before?

Alizter: You could also do that by inlining the arg but doesn't matter. I was simply interested if there was a deeper reason.

Alizter (approved):

Issue #8067, opened 2023-06-28
Caching of Directory Targets


rgrinberg:
Currently, rules which produce directories are ignored by the cache. This is unfortunate, because such rules are often quite heavy and would benefit a lot from caching.

It should be straightforward to extend the cache to support directories. The only requirement is that we store them in the cache which reuses equivalent files between directories targets produced by different rules.

There's a couple of approaches we can take. Our current representation of cache entries is essentially `Filename.t list`. The simplest way to extend it would be to make it `Path.Local.t list`. Which would allow for entries relative to sub directories. This is already enough for sharing all files between targets and doesn't require us to change the format of the cache.

If we we wanted to share entire directories between entries, we could imagine imagine changing the format to `(kind * Filename.t) list` where `type kind = File | Directory`. The advantage of this representation is that we don't need to repeat the same file lists if they're reused. The disadvantage is that it requires changing our entries format.

cc @snowleopard

snowleopard:
We are planning to take a look at this internally, as part of an intern project that starts next week.

snowleopard:
The intern project didn't happen unfortunately, so as far as I know noone is currently working on this.

PR #8868, opened 2023-10-06
test: directory targets and shared cache

emillon:
See #8067.

### DIFF
--- a/test/blackbox-tests/test-cases/directory-targets/cache.t
+++ b/test/blackbox-tests/test-cases/directory-targets/cache.t
@@ -55,3 +55,11 @@ displayed, but the files should be created.
   true
   $ dune_cmd exists _build/default/out/b
   true
+
+If part of a directory target is removed, we expect it to be rebuilt using the
+cache.
+
+  $ dune_cmd wait-for-fs-clock-to-advance
+  $ rm _build/default/out/a
+  $ dune build out
+  Running create.sh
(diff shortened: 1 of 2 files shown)


Alizter:
Would it be useful to have a regular file that also prints something on creation so that we can compare the behaviour for files side by side with the behaviour for directories?

rgrinberg:
There is some overlap with what `main.t` is testing. Could you move (or delete if it's redundant) what's being tested there into your test?

emillon:
@rgrinberg done.
@Alizter presumably this is already being tested elsewhere; this test is specific to the interaction between directory targets and the shared cache.

rgrinberg (approved):

test/blackbox-tests/test-cases/directory-targets/cache.t
```diff
@@ -0,0 +1,65 @@
...
+  $ dune_cmd exists _build/default/out/b
+  true
+  $ dune cache size
+  0B
+
+Now we create another project with the same contents.
+
+  $ cd ..
+  $ mkdir b
+  $ cp -t b a/dune-project a/dune a/create.sh
```
rgrinberg, line 46: I see `-t` doesn't exist on macos.

Btw, why not just `mv`? I see you're not using the old project anyway.

emillon: Ah I thought it was a posix thing; I switched to the other form.

PR #8870, opened 2023-10-06
test(stdune): Path.drop_prefix doesn't respect path boundaries

Alizter:
This is perhaps an unexpected result of using `Path.drop_prefix`. Perhaps another function `Path.drop_path_prefix` is needed instead if this behaviour is intentional.

### DIFF
--- a/otherlibs/stdune/test/path_tests.ml
+++ b/otherlibs/stdune/test/path_tests.ml
@@ -39,6 +39,10 @@ let drop_build_context p =
 
 let local_part p = Path.local_part p |> Path.Local.to_dyn |> print_dyn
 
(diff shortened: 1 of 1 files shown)


Alizter:
@anmonteiro since you implemented this function I am interested to hear what you think about this issue.

anmonteiro:
This looks like a bug in `drop_prefix` instead. I think we want to drop the prefix if it's matched exactly as a path segment.

I think that wouldn't break anything existing usage, but I'm eager to see the results of that test run.

Alizter:
Reopening as we don't have an issue. But the tests here have been subsumed by #8961.

PR #8876, opened 2023-10-07
fix(pkg): correctly verify tarball checksums

rgrinberg:
The changes I've made to the opam libraries are rather ugly, but it does seem like their api needs to be redesigned a bit to allow more customized error handling. This isn't a priority for now though.

### DIFF
--- a/test/expect-tests/dune_pkg/fetch_tests.ml
+++ b/test/expect-tests/dune_pkg/fetch_tests.ml
@@ -54,7 +54,7 @@ Content-Length: %d
   port, thread
 ;;
 
-let download ~unpack ~port ~filename ~target ?checksum () =
+let download ?(reproducible = true) ~unpack ~port ~filename ~target ?checksum () =
   let open Fiber.O in
   let url = url ~port ~filename in
   let* res = Fetch.fetch ~unpack ~checksum ~target url in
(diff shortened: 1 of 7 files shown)


Leonidas-from-XIV:
What is the fix here? When I wrote the original code I tried to make sure to avoid having to add additional patches to our OPAM fork, because the way it is currently computed in OPAM is very much unusable.

Alizter:
@Leonidas-from-XIV Now that we unpack downloads the file being checksummed is a directory which breaks the opam code.

test/expect-tests/dune_pkg/fetch_tests.ml
```diff
@@ -186,5 +191,8 @@ let%expect_test "downloading, tarball" =
   (Dune_util__Report_error.Already_reported)
   Trailing output
   ---------------
-  Error: Is a directory |}]
+  Error: Expected checksum was
+  md5=c533195dc4253503071a19d42f08e877
+  but got
+  <REDACTED> |}]
```
Alizter, line 197: @Leonidas-from-XIV  See the test here.

Leonidas-from-XIV:
Ah, but then shouldn't the OPAM patches be applied to our fork first and then updated here? Otherwise the one who updates it next will accidentally undo these patches.

Leonidas-from-XIV:
The problem is that this changes the API so whatever works here might need additional changes in our OPAM fork to continue building (e.g. handling of the new variant).

Leonidas-from-XIV (commented):
I think it would be nice to have something that we could eventually upstream (shouldn't be too controversial) so we don't have to maintain a patch like this.

rgrinberg:
Yes, it would be nice to have something that we can upstream. For now, I’m afraid we’ll have to work with this hot fix. Without this, the entire feature is not usable in practice so it’s a bit urgent to get something working.

Alizter:
@rgrinberg Were you going to update the fork?

PR #8877, opened 2023-10-07
refactor(engine): improve execution parameter handling

rgrinberg:
* Raise an error on invalid paths
* Replace the catch wildcard pattern with explicit constructors for
  exhaustiveness checking

### DIFF
--- a/src/dune_engine/build_system.ml
+++ b/src/dune_engine/build_system.ml
@@ -442,7 +442,8 @@ end = struct
       match Dpath.Target_dir.of_target dir with
       | Regular (With_context (_, dir)) | Anonymous_action (With_context (_, dir)) ->
         (Build_config.get ()).execution_parameters ~dir
-      | _ -> Execution_parameters.default
+      | Anonymous_action Root | Regular Root | Invalid _ ->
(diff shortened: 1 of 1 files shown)


snowleopard (approved):
Looks good, thanks!

src/dune_engine/build_system.ml
```diff
@@ -442,7 +442,11 @@ end = struct
       match Dpath.Target_dir.of_target dir with
       | Regular (With_context (_, dir)) | Anonymous_action (With_context (_, dir)) ->
         (Build_config.get ()).execution_parameters ~dir
-      | _ -> Execution_parameters.default
+      | Anonymous_action Root | Regular Root -> Execution_parameters.default
```
snowleopard, line 445: Could you add a comment explaining the difference in behaviour in the two cases above?

rgrinberg: Actually, I just forbade the execution parameters for `Anonymous Root` and `Regular Root`. There can be no rules in those directories, so there's no need for execution parameters either.

snowleopard (approved):
Looks good!

PR #8882, opened 2023-10-07
chore: split melange off into own CI job

Alizter:
This splits off the melange tests into their own separate CI job like for Coq. I will update the nix flake after this PR to separate concerns.

And now that we are doing this for multiple jobs, we take the opportunity to make uniform how deps for these jobs are installed.


rgrinberg:
We don't want to use enabled_if

PR #8887, opened 2023-10-09
feature(cram): allow overriding default alias

rgrinberg:
Introduce a [(default_aliases ..)] that allows to set the default alias
for a cram test.

Unlike, the existing [alias] field, we're only allowed to set the
default once. If the default isn't set, it's just [runtest].

### DIFF
--- a/doc/changes/8887.md
+++ b/doc/changes/8887.md
@@ -1,0 +1,2 @@
+- Introduce the `runtest_alias` field to the `cram` stanza. This allows
+  removing default `runtest` alias from tests. (@rgrinberg, #8887)
--- a/doc/tests.rst
+++ b/doc/tests.rst
@@ -674,6 +674,9 @@ The ``cram`` stanza accepts the following fields:
 - ``deps`` - dependencies of the test
 - ``(package <package-name>)`` - attach the tests selected by this stanza to the
   specified package
+- ``(runtest_alias <true|false>)`` - when set to ``false``, do not add the
+  tests to the ``runtest`` alias. The default is to add every cram test to
+  ``runtest``, but this is not always desired.
 
 A single test may be configured by more than one ``cram`` stanza. In such cases,
 the values from all applicable ``cram`` stanzas are merged together to get the
--- a/src/dune_rules/cram/cram_stanza.ml
+++ b/src/dune_rules/cram/cram_stanza.ml
@@ -28,6 +28,7 @@ type t =
   ; enabled_if : Blang.t
   ; locks : Locks.t
   ; package : Package.t option
+  ; runtest_alias : (Loc.t * bool) option
   }
 
 type Stanza.t += T of t
@@ -42,6 +43,10 @@ let decode =
      and+ locks = Locks.field ~check:(Dune_lang.Syntax.since Stanza.syntax (2, 9)) ()
      and+ package =
        Stanza_common.Pkg.field_opt ~check:(Dune_lang.Syntax.since Stanza.syntax (2, 8)) ()
+     and+ runtest_alias =
+       field_o
+         "runtest_alias"
+         (Dune_lang.Syntax.since Stanza.syntax (3, 11) >>> located bool)
      in
-     { loc; alias; deps; enabled_if; locks; applies_to; package })
+     { loc; alias; deps; enabled_if; locks; applies_to; package; runtest_alias })
 ;;
--- a/src/dune_rules/cram/cram_rules.ml
+++ b/src/dune_rules/cram/cram_rules.ml
@@ -14,7 +14,7 @@ module Spec = struct
 
   let empty =
     { loc = Loc.none
-    ; alias = Alias.Name.Set.singleton Alias0.runtest
+    ; alias = Alias.Name.Set.empty
     ; enabled_if = [ Blang.true_ ]
     ; locks = Path.Set.empty
     ; deps = []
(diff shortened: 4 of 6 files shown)


Alizter:
It would have probably been better to have an `(aliases :standard)` field IMO. But I don't know a good way to get people to transition to using it if we ever switch `(aliases foo)` -> `(aliases :standard foo)`.

rgrinberg:
@gridbugs do you want to review this one?

test/blackbox-tests/test-cases/cram/default-aliases.t
```diff
@@ -0,0 +1,48 @@
...
+  $ dune build @this
+  File "foo.t", line 1, characters 0-0:
+  Error: Files _build/default/foo.t and _build/default/foo.t.corrected differ.
+  [1]
+  $ dune build @blah
+  File "foo.t", line 1, characters 0-0:
+  Error: Files _build/default/foo.t and _build/default/foo.t.corrected differ.
+  [1]
+
+Now we try setting the defualt aliases twice. This should be impossible:
```
gridbugs, line 33:
```suggestion
Now we try setting the default aliases twice. This should be impossible:
```

gridbugs:
Can you please update the docs with the new field. Also it's not clear to me what it means to change the default alias for a test. It looks like you always need to specify the name of the alias in order to run the test so in what sense is it a default? From looking at the example it seems like the behaviour of this feature is that tests can be grouped together by specifying the same `(default_aliases <name>)` an then you can build that alias to run all the associated tests. Is that right? Maybe the field should be named `group_alias` and a separate flag should allow a test to be excluded from `@runtest`.

rgrinberg:
> It looks like you always need to specify the name of the alias in order to run the test so in what sense is it a default?

By default, the alias is set to `runtest`. This feature allows to override this deafult.

> From looking at the example it seems like the behaviour of this feature is that tests can be grouped together by specifying the same (default_aliases <name>) an then you can build that alias to run all the associated tests. Is that right? Maybe the field should be named group_alias

This property applies to all fields in the `(cram ..)` stanza. The `(applies_to ..)` field determines to which fields the stanza applies.

>  and a separate flag should allow a test to be excluded from @runtest.

Yeah, that could work. We could add a flag like `(default_alias <true|false>)`.  It would be very similar to what is done here already, so I'm not sure if it's an improvement.

gridbugs:
> By default, the alias is set to `runtest`. This feature allows to override this deafult.

Do you mean all tests defined in the same directory as the `dune` file will be associated with the specified alias rather than with `@runtest`?

rgrinberg:
Yes. Although instead of a single alias, it's a list. With a list, we can also set the default to be the empty list and therefore remove all default aliases.

gridbugs:
Ok that makes sense. I find the naming a bit confusing because there's now a notion of a "default default" alias (`@runtest`). Like there are two levels of defaults - the default aliases associated with tests and the choice of default alias to use if none is specified.

rgrinberg:
Wait, there's only one level of defaults that I can see. To recap, there's two fields:

* The existing `(alias ..)` field that can only be used to _add_ a new alias to a set of tests. This field isn't able to remove any aliases which is why it cannot be used to solve the problem of removing the default `runtest` alias.
* The new `(default_aliases ..)` field that can be used to _set_ the default alias instead of `runtest`. Only this one is a real default.

gridbugs:
If `default_aliases` is not specified it effectively takes the value `runtest`, no? That would make `runtest` the default value for `default_aliases`.

rgrinberg:
Okay, I see what you mean. What if the name was `base_aliases` instead?

I suppose your idea of just having a flag `(default_alias <true|false>)` is also ok.

Any preferences?

gridbugs:
I like the name `base_aliases`. Any reason to not do both though? I see this change as two distinct features that happen to share an implementation:
- allowing a test to be excluded from the set of tests run by `dune runtest`
- allowing a group of related tests to be grouped together under a new alias (this could be particularly useful with pkg - we could add an alias that just runs the pkg tests without removing those tests from `@runtest`).

So what if we made it so regardless of the value of `base_aliases`, the tests remain in `@runtest` unless explicitly excluded with a flag. And rather than calling the flag `(default_alias <true|false>)` I think it would be clearer to have a `(exclude_from_runtest)` flag.

rgrinberg:
> allowing a group of related tests to be grouped together under a new alias (this could be particularly useful with pkg - we could add an alias that just runs the pkg tests without removing those tests from @runtest).

If you need this, we can already have this:

```
$ cat test/blackbox-tests/test-cases/pkg/dune
(cram
 (applies_to :whole_subtree)
 (alias pkg))
```

The only thing that's missing is excluding things from `runtest` by default.

gridbugs:
It's also not possible to include a test in multiple custom aliases (ie. aliases besides `@runtest`), right?

rgrinberg:
Nope, it's possible. You just need multiple `cram` stanzas.

```
$ cat a/dune
(cram (alias foo))
(cram (alias bar))
```

Will make all the tests in the directory `a/` buildable under 3 aliases:
1. `runtest`
2. `foo`
3. `bar`

gridbugs:
Ah cool. In that case I'd prefer we add a flag that excludes the tests in a directory from `@runtest`. I would find that UI clearer than setting an empty list of base aliases, since it's not immediately obvious that the reason a test is included in `@runtest` is because `@runtest` is the default base alias for all cram tests.

rgrinberg:
Okay, I'll call the flag `(runtest_alias <enable|disable>)`. It will also not be possible to set it more than once per test.

rgrinberg:
@gridbugs updated. If this is OK, I will update the docs.

doc/changes/8887.md
```diff
@@ -0,0 +1,2 @@
+- Introduce the `default_aliases` to the `cram` stanza. This allows overriding
```
gridbugs, line 1: This is now out of date

gridbugs (commented):
Looks good to me

rgrinberg:
docs are in

doc/tests.rst
```diff
@@ -674,6 +674,8 @@ The ``cram`` stanza accepts the following fields:
 - ``deps`` - dependencies of the test
 - ``(package <package-name>)`` - attach the tests selected by this stanza to the
   specified package
+- ``(runtest_alias <true|false>)`` - when set to ``false``, do not add the
+  tests to the ``runtest`` alias by default.
```
gridbugs, line 678: I don't get why "by default" is there. Can you add a note that if this field is omitted then the tests will be included in the `@runtest` alias. Also what's the use case for `(runtest_alias true)`? Is there ever a time when the field should be explicitly set to true?

rgrinberg: > I don't get why "by default" is there. Can you add a note that if this field is omitted then the tests will be included in the @runtest alias

I clarified it.

> Also what's the use case for (runtest_alias true)? Is there ever a time when the field should be explicitly set to true?

Mostly to allow things like `(runtest_alias %{lib-available:foo})` or blang expressions if there is ever a need. 

Experience tells me that flag-like fields (e.g. `no_runtest_alias`) end up being annoying to use. E.g. suppose that we change the default to not add the runtest alias to cram by default in a later version of the dune language, how would the user be able to switch it back on?

gridbugs: Makes sense. Thanks for clarifying.

gridbugs (approved):

rgrinberg:
@Alizter feel free to use this to move melange/coq to separate CI jobs.

src/dune_rules/cram/cram_stanza.ml
```diff
@@ -42,6 +43,10 @@ let decode =
      and+ locks = Locks.field ~check:(Dune_lang.Syntax.since Stanza.syntax (2, 9)) ()
      and+ package =
        Stanza_common.Pkg.field_opt ~check:(Dune_lang.Syntax.since Stanza.syntax (2, 8)) ()
+     and+ runtest_alias =
+       field_o
+         "runtest_alias"
+         (Dune_lang.Syntax.since Stanza.syntax (3, 11) >>> located bool)
```
emillon, line 49: this should be 3.12 since it got added in 3.12.0

PR #8902, opened 2023-10-10
fix: ignore [bigarray] correctly

### LINKED ISSUE #8901: Local libraries named bigarray are incorrectly removed
This change is incorrect and introduces a bug that forbids local libraries named `bigarray` from being used on OCaml < 5.0.

_Originally posted by @rgrinberg in https://github.com/ocaml/dune/pull/5526#discussion_r1350998122_

rgrinberg:
The way it was ignored previously, it would completely ignore locally
named libraries named bigarray.

### DIFF
--- a/doc/changes/8902.md
+++ b/doc/changes/8902.md
@@ -1,0 +1,2 @@
+- Do not ignore libraries named `bigarray` when they are defined in conjunction
+  with OCaml 5.0 (#8902, fixes #8901, @rgrinberg)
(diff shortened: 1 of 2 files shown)


rgrinberg:
Actually, the most correct approach would be to introduce an explicit `stdlib` for `Lib.t` and then just redirect `bigarray` to do that.

Anyway, for now this is good enough for a fix.

emillon (approved):

PR #8904, opened 2023-10-10
fix: add fake opam switch to opam-var tests

gridbugs:
Prior to this change the tests in pkg/opam-var would fail on older versions of opam with the error:
```
[ERROR] No switch is currently set. Please use 'opam switch' to set or
install a switch
```

This change adds an opam switch to the fake opam root used by these tests.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/opam-var/dune
+++ b/test/blackbox-tests/test-cases/pkg/opam-var/dune
@@ -4,12 +4,19 @@
 
 (data_only_dirs fake_opam_root)
 
+; Without this, no copying rules will be setup for the .opam-switch directory
+
+(subdir
+ fake_opam_root/default
+ (dirs :standard .opam-switch))
+
 ; We trick opam into thinking it has a config file with fake_opam_root so that we can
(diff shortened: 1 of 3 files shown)


test/blackbox-tests/test-cases/pkg/opam-var/dune
```diff
@@ -4,12 +4,19 @@
 
 (data_only_dirs fake_opam_root)
 
+; Without this, the .opam-switch directory won't be copied to the build directory.
+
+(subdir
+ fake_opam_root/default
+ (dirs :standard .opam-switch))
```
Alizter, line 11: Erm. That's strange. We mark `fake_opam_root` as a data_only_dirs. Shouldn't everything inside be copied? Seems like a bug @rgrinberg.

rgrinberg: The comment should read:

```
;; Without this, no copying rules will be setup for the .opam-switch directory
```

Things are actually copied when there are dependencies on them.

rgrinberg (approved):

PR #8905, opened 2023-10-10
fix: use portable shell shebang in exec-bin test

gridbugs:


gridbugs:
Some systems don't have `/bin/bash` (e.g. nixos) but `/bin/sh` is guaranteed by posix.

emillon:
Tempted to create label `i-use-nix-btw` :)

PR #8909, opened 2023-10-10
fix: remove code error in copying sandbox

rgrinberg:
When there's a type of file we can't copy, do not raise a code error.
This situation is possible if the user sneaks in a socket file in the
build directory.

### DIFF
--- a/src/dune_engine/sandbox.ml
+++ b/src/dune_engine/sandbox.ml
@@ -52,9 +52,15 @@ let rec copy_recursively (src_kind : Unix.file_kind) ~src ~dst =
        List.iter contents ~f:(fun (name, kind) ->
          copy_recursively kind ~src:(Path.relative src name) ~dst:(Path.relative dst name)))
   | _ ->
-    Code_error.raise
-      "Can not copy file of this kind"
-      [ "src_kind", File_kind.to_dyn src_kind; "src", Path.to_dyn src ]
+    User_error.raise
+      ~hints:
+        [ Pp.text "re-run dune to delete the stale artifact, or manually delete this file"
+        ]
+      [ Pp.textf
+          "Failed to copy file %s of kind %s while creating a copy sandbox"
+          (Path.to_string_maybe_quoted src)
+          (File_kind.to_string_hum src_kind)
+      ]
 ;;
 
 let create_dirs t ~deps ~rule_dir =


Alizter:
What should the user do when they encounter this error? With the code error at least it mentions that this is something Dune didn't account for and should be reported.

rgrinberg:
That's the whole point. There's nothing to report. Dune cannot sandbox these other file types.

src/dune_engine/sandbox.ml
```diff
@@ -52,9 +52,12 @@ let rec copy_recursively (src_kind : Unix.file_kind) ~src ~dst =
        List.iter contents ~f:(fun (name, kind) ->
          copy_recursively kind ~src:(Path.relative src name) ~dst:(Path.relative dst name)))
   | _ ->
-    Code_error.raise
-      "Can not copy file of this kind"
-      [ "src_kind", File_kind.to_dyn src_kind; "src", Path.to_dyn src ]
+    User_error.raise
+      [ Pp.textf
+          "Can not copy file %s of kind %s"
```
Alizter, line 57:
```suggestion
          "Cannot copy file %s of kind %s"
```

Alizter (approved):
I still think that some more guidance for users in this situation would be warranted but I also understand that it is not a very common one. Your call.

rgrinberg:
It would help if I understood what kind of guidance would you expect the user to see. As far as I can tell, we're giving the users all the information that we have.

Alizter:
This error only happens when a rule that uses say a socket file is sandboxed. In which case if a user is running their build and they come across this error, how will they know that it comes from sandboxing? I would have expected something along the lines of:
```
Dune cannot sandbox said file because it cannot be copied reliably. Try disabling sandboxing or removing the file.
```
or something to that effect. If I have misunderstood the whole situation, then ignore me and sorry for the noise.

rgrinberg:
Okay, if it's the error message I'm happy to improve it.

Your suggestion is wrong on two accounts though:

> Dune cannot sandbox said file because it cannot be copied reliably
It cannot be copied at all.
> Try disabling sandboxing or removing the file.
Disabling sandboxing isn't going to help. The file was created there after the rule was already executed.

I still don't see how the previous code error was better. It signalled a bug in dune which didn't actually exist.

emillon (approved):
I agree that it should be a user error and that it's one where it's difficult to have a good error message. This is a clear improvement.

emillon:
We can add a test case with `mknod` or `mkfifo` in a user action but that's not strictly necessary.

Alizter:
As a user encountering the message:
```
"Cannot copy file %s of kind %s"
```
here are some questions that might pop into my head:
1. Why is something trying to copy a file? (A user shouldn't have to know the internal details of how Dune works)
2. Is there something I can do about this? (The message could say, "Dune cannot work with files of this kind")
3. Where can I find more information about this? (The Code_error at least had the suggestion of a bug report which could be useful to find more information).

Anyway, I don't think its necessary to improve this user message further, but I just wanted to explain my thoughts on it more clearly.

rgrinberg:
> We can add a test case with mknod or mkfifo in a user action but that's not strictly necessary.

It's actually harder than it sounds. Because we already make sure the targets are all normal files after running the rules. We would need to introduce one of these commands after this check is done and before it's copied while sandboxing.

> The Code_error at least had the suggestion of a bug report which could be useful to find more information

This is the exact thing this PR is trying to avoid. We don't want incorrect bug reports from users.

PR #8911, opened 2023-10-11
test: easier to read permissions from dune_cmd

Alizter:
This adds a `dune_cmd ls` command that prints output like this:
```
-rw-r--r-- README.md
drwxr-xr-x _boot
drwxr-xr-x _build
```

This is more portable than using `ls` which can have some subtle differences in implementation across platforms. It also allows us to inspect the permissions of files in a more readable way.

@rgrinberg This might be useful for #8908

### DIFF
--- a/test/blackbox-tests/test-cases/directory-targets/copy-sandboxing.t
+++ b/test/blackbox-tests/test-cases/directory-targets/copy-sandboxing.t
@@ -8,16 +8,7 @@ mode:
 
   $ cat >print-permissions.sh <<EOF
   > /usr/bin/env sh
-  > echo permissions of output/
-  > dune_cmd stat permissions output
-  > echo permissions of output/y
-  > dune_cmd stat permissions output/y
-  > echo permissions of output/x
-  > dune_cmd stat permissions output/x
-  > echo permissions of output/subdir
-  > dune_cmd stat permissions output/subdir
-  > echo permissions of output/subdir/z
-  > dune_cmd stat permissions output/subdir/z
+  > dune_cmd ls -d output output/y output/x output/subdir output/subdir/z
   > echo ""
   > EOF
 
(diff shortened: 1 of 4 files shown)


rgrinberg:
Could you use it in some test so we can see how it works?

test/blackbox-tests/test-cases/directory-targets/copy-sandboxing.t
```diff
@@ -8,16 +8,11 @@ mode:
...
-  > dune_cmd stat permissions output
-  > echo permissions of output/y
-  > dune_cmd stat permissions output/y
-  > echo permissions of output/x
-  > dune_cmd stat permissions output/x
-  > echo permissions of output/subdir
-  > dune_cmd stat permissions output/subdir
-  > echo permissions of output/subdir/z
-  > dune_cmd stat permissions output/subdir/z
+  > dune_cmd ls output
```
rgrinberg, line 11: I think this is equivalent to:

```
dune_cmd ls output/ output/subdir
```

test/blackbox-tests/test-cases/directory-targets/copy-sandboxing.t
```diff
@@ -46,38 +37,31 @@ mode:
...
-  776
-  permissions of output/subdir/z
-  444
+  drwxrwxrwx output
+  -r--r--r-- output/y
+  -r--r--r-- output/x
+  drwxrwxrw- output/subdir
+  -r--r--r-- output/subdir/z
   
+  $ ( cd _build/default && dune_cmd ls output output/subdir )
```
Alizter, line 60: I will remove this. Just to demo. Resolve the comment when you've seen it.

test/blackbox-tests/test-cases/directory-targets/copy-sandboxing.t
Alizter: I've replaced it with an explicit `dune_cmd ls -d` so that we can see the perms on the dir too.

rgrinberg (changes requested):

test/blackbox-tests/utils/dune_cmd.ml
```diff
@@ -59,6 +59,101 @@ module Stat = struct
...
+
+  let print_stat filename (stat : Unix.stats) =
+    Printf.printf
+      "%c%s %s\n"
+      (kind_to_char stat.st_kind)
+      (permission_to_string stat.st_perm)
+      filename
+  ;;
+
+  let print_stats_of_dir file =
```
rgrinberg, line 104: Why is the argument called file when this is a directory?

test/blackbox-tests/utils/dune_cmd.ml
```diff
@@ -59,6 +59,101 @@ module Stat = struct
...
+    | S_DIR ->
+      (match Path.readdir_unsorted file with
+       | Ok files ->
+         List.map files ~f:(fun x -> x, Path.lstat_exn (Path.relative file x))
+         |> List.sort ~compare:(fun (x, _) (y, _) -> String.compare x y)
+         |> List.iter ~f:(fun (x, stat) ->
+           print_stat (Path.relative file x |> Path.to_string) stat)
+       | Error unix_err -> Unix_error.Detailed.raise unix_err)
+    | _ ->
+      raise (Arg.Bad (sprintf "%s is not a directory" (Path.to_string_maybe_quoted file)))
```
rgrinberg, line 116: How is this possible?

test/blackbox-tests/utils/dune_cmd.ml
```diff
@@ -59,6 +59,101 @@ module Stat = struct
...
+    | S_DIR -> 'd'
+    | S_CHR -> 'c'
+    | S_BLK -> 'b'
+    | S_LNK -> 'l'
+    | S_FIFO -> 'p'
+    | S_SOCK -> 's'
+  ;;
+
+  let print_stat filename (stat : Unix.stats) =
+    Printf.printf
```
rgrinberg, line 97: `printfn` here and elsewhere. It doesn't need the `Printf` qualifier either.

PR #8914, opened 2023-10-11
refactor(pkg): remove pointless module type alias

rgrinberg:


Alizter (approved):
I think the intention was the enforce the type later, but when I tried that it became to finicky to work with.

PR #8916, opened 2023-10-11
Refactor Common.init

rleshchinskiy:
Previously, we would obtain a [Common.t] from [Common.term] but it would be only partially initialised, until we called [Common.init] on it. This seemed unnecessarily unsafe and also made some upcoming changes to tracing harder than they should be.

This PR refactors things such that [Common.init] now takes a [Builder.t] (which was partially exposed but not really used externally before) and produces a [Common.t]. [Builder.t] encapsulates how Dune should be initialised and [Common.t] is the result of that initialisation. This seems cleaner.

This also allows us to move some config settings into [Builder.t] which were previously sprinkled through the code. The upcoming tracing settings will be able to go in the same spot.

### DIFF
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -143,9 +143,9 @@ let runtest_info =
 
 let runtest_term =
   let name_ = Arg.info [] ~docv:"DIR" in
-  let+ common = Common.term
+  let+ builder = Common.Builder.term
   and+ dirs = Arg.(value & pos_all string [ "." ] name_) in
-  let config = Common.init common in
+  let common, config = Common.init builder in
   let request (setup : Import.Main.build_system) =
     Action_builder.all_unit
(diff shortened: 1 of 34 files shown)


snowleopard (approved):
Reviewed internally. @rgrinberg could you please double-check that this change makes sense externally?

Alizter (approved):
LGTM, I think it was always strange that you had to call `init` when using common but I like that it now falls out from the typing.

snowleopard:
@Alizter Thanks for taking a look!

rgrinberg:
The change is appreciated, but perhaps you forgot that we no longer sync the command line?

snowleopard:
> The change is appreciated, but perhaps you forgot that we no longer sync the command line?

@rgrinberg If I understand correctly, some of Roman's further changes might depend on this one.

Issue #8652, opened 2023-09-14
Overlapping Dependencies in Lock Directory


rgrinberg:
Apologies for the title, better suggestions are welcome. The issue is best understood by an example.

Consider the following setup:

```
$ cat dune.lock/b.pkg
(depends c)
$ cat dune-project
...
(package
 (name a)
 (deps b))
(package
 (name c))
```

So `a` and `c` exist in the workspace while `b` is fetched via the lock directory. If you follow the dependency edges of `a`, you will go through the lock directory and back to the workspace. Such dependencies edges are currently not allowed in dune.

The reason is the following cycle:
1. To build anything in the workspace, we need to lookup some library, say `x`.
2. Looking up `x` requires us to build every single package in the lock directory
3. If a package in the lock directory depends on a package in the workspace, that will immediately result in a cycle.

We could do better by limiting the search to a subset of packages in some situations. For example, if we're building `foo.lib` in package `foo`, we look at the package dependencies of `foo` to determine the set of packages that need to be built to resolve the library.

This of course only works for building libraries or executables inside packages. A private library or executable belongs to no package so the trick above fails (However, implementing #8650 would allow this to work anyway).

Anyway, if we manage to get over the cycles, we'll need some additional setup to make the build work. In the example above, package `b` needs the installed layout of `c`. We essentially already have this in _build/install, but it needs to be subset of files needed by `b` (otherwise we'll run into different cycles). In particular, we only need the install rules for `c`. That's not a problem, we just need to setup the install rules again for every package in the workspace in `_private/` individually. So in practice, we'll just have rules in `_private/default/.pkg/$pkg` for every package in the workspace.

rgrinberg:
This issue actually prevents dune's package management to be used to work with dune itself.

Since our repo includes the dune-configurator project, we cannot use any package that itself depends on dune-configurator. That ends up being just about every package.

rgrinberg:
I talked about this @gridbugs and we have a plan on how to make progress on this issue. The first step of this issue is to make it possible for dune to build a package inside the workspace without loading every single package in the lock directory. To do this, we need to make use of two sources of information:

1. The list of dependencies of a package. This list is available in the dune-project file after it's evaluated against a lock dir
2. The association between a subset of the workspace and the package that it belongs to.

[...]

PR #8918, opened 2023-10-11
fix: record contents of installed directories

emillon:
Previously we would just record the directory name, with no information that this is a directory; this would cause an issue when depending against the package.


doc/changes/package-deps-dirs.md
```diff
@@ -0,0 +1,3 @@
+- Record contents of installed directories in `dune-package` files. This fixes
+  `(package)` dependencies against packages that contain such directories.
+  (#...., fixes #8915, @emillon)
```
emillon, line 3:
```suggestion
  (#8918, fixes #8915, @emillon)
```

src/dune_rules/install_rules.ml
```diff
@@ -412,10 +420,18 @@ end = struct
...
-      Memo.List.map dirs_expanded ~f:(fun fb ->
+      Memo.List.concat_map dirs_expanded ~f:(fun fb ->
         let loc = File_binding.Expanded.src_loc fb in
-        let+ entry = make_entry ~kind:`Directory fb in
-        Install.Entry.Sourced.create ~loc entry)
+        let* entry = make_entry ~kind:`Directory fb in
+        let* rule_opt = Load_rules.get_rule (Path.build entry.src) in
+        let+ entries =
+          match rule_opt with
+          | None -> Memo.return [ entry ]
```
emillon, line 429: I think it's dead code but I'm not too familiar with that part.

src/dune_rules/install_rules.ml
```diff
@@ -412,10 +420,18 @@ end = struct
...
+          | Some rule ->
+            let+ { targets; _ } = Build_system.execute_rule rule in
```
emillon, line 431: There's a dependency loop here: the set of installed files is necessary to populate the contents of the install file; but dune-package is also installed so we can't easily get the set of installed files when creating the contents of `dune-package`. I think we could try to make `gen_install_file` in this module generate `dune-project` as well but I'm not 100% sure. Anyway, if I understand correctly we're going to execute this rule just a bit later so I don't think it's too bad to call `execute_rule` directly here to get the produced targets.

rgrinberg: There's another problem as well. With the current code, we now need to build all directory targets to just generate the rule for the dune-package file. This has two practical disadvantages:

1. Depending on `(package foo)` is going to build the directory targets of all other packages. 
2. It will make the implementation of https://github.com/ocaml/dune/issues/8652 more difficult for similar reasons. We'll now won't be able to list the entries of a package without building things.

So I'm afraid this doesn't work. I think we should bite the bullet and just add `(dirs )` to dune-package.

emillon: ok. I did so in #8953

emillon:
Closed in favor of #8953

PR #8920, opened 2023-10-11
fix: make copying sandbox files writeable

rgrinberg:
Currently, the `copy` sandbox just uses whatever permissions are set in the workspace local cache.

With this PR, it will now make all files and directories writeable.

### DIFF
--- a/src/dune_engine/sandbox.ml
+++ b/src/dune_engine/sandbox.ml
@@ -41,26 +41,60 @@ type t =
 let dir t = t.dir
 let map_path t p = Path.Build.append t.dir p
 
-let rec copy_recursively (src_kind : Unix.file_kind) ~src ~dst =
-  match src_kind with
-  | S_REG -> Io.copy_file ~src ~dst ()
-  | S_DIR ->
-    (match Path.Untracked.readdir_unsorted_with_kinds src with
-     | Error e -> Unix_error.Detailed.raise e
-     | Ok contents ->
-       Path.mkdir_p dst;
-       List.iter contents ~f:(fun (name, kind) ->
-         copy_recursively kind ~src:(Path.relative src name) ~dst:(Path.relative dst name)))
-  | _ ->
-    User_error.raise
-      ~hints:
-        [ Pp.text "re-run dune to delete the stale artifact, or manually delete this file"
+module Item = struct
+  type t =
+    | File
+    | Directory of { perms : int }
+    | Other of Unix.file_kind
+
+  let of_path path =
+    let { Unix.st_kind; st_perm; _ } = Path.Untracked.stat_exn path in
+    match st_kind with
+    | S_DIR -> Directory { perms = st_perm }
+    | S_REG -> File
+    | kind -> Other kind
+  ;;
+
+  let of_kind path (kind : Unix.file_kind) =
+    match kind with
+    | S_DIR -> Directory { perms = (Path.Untracked.stat_exn path).st_perm }
+    | S_REG -> File
+    | _ -> Other kind
+  ;;
+end
+
+let copy_recursively =
+  let chmod_file = Path.Permissions.add Path.Permissions.write in
+  let chmod_dir p =
+    Path.Permissions.add Path.Permissions.execute p
+    |> Path.Permissions.add Path.Permissions.write
+  in
+  let rec loop item ~src ~dst =
+    match (item : Item.t) with
+    | File -> Io.copy_file ~chmod:chmod_file ~src ~dst ()
+    | Directory { perms } ->
+      (match Path.Untracked.readdir_unsorted_with_kinds src with
+       | Error e -> Unix_error.Detailed.raise e
+       | Ok contents ->
+         let perms = chmod_dir perms in
+         Path.mkdir_p ~perms dst;
+         List.iter contents ~f:(fun (name, kind) ->
+           let src = Path.relative src name in
+           let item = Item.of_kind src kind in
+           loop item ~src ~dst:(Path.relative dst name)))
+    | Other kind ->
+      User_error.raise
+        ~hints:
+          [ Pp.text
+              "Re-run Dune to delete the stale artifact, or manually delete this file"
+          ]
+        [ Pp.textf
+            "Failed to copy file %s of kind %S while creating a copy sandbox"
+            (Path.to_string_maybe_quoted src)
+            (File_kind.to_string_hum kind)
         ]
-      [ Pp.textf
-          "Failed to copy file %s of kind %s while creating a copy sandbox"
-          (Path.to_string_maybe_quoted src)
-          (File_kind.to_string_hum src_kind)
-      ]
+  in
+  loop
 ;;
 
 let create_dirs t ~deps ~rule_dir =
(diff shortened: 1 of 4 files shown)


src/dune_engine/sandbox.ml
```diff
@@ -41,26 +41,39 @@ type t =
...
+             | S_DIR -> `Dir (Path.Untracked.lstat_exn src).st_perm
+             | S_REG -> `File
+             | _ -> `Other kind
+           in
+           loop what ~src ~dst:(Path.relative dst name)))
+    | `Other kind ->
+      User_error.raise
+        ~hints:
+          [ Pp.text
+              "re-run dune to delete the stale artifact, or manually delete this file"
```
snowleopard, line 68:
```suggestion
              "Re-run Dune to delete the stale artifact, or manually delete this file."
```

src/dune_engine/sandbox.ml
```diff
@@ -41,26 +41,39 @@ type t =
...
+           in
+           loop what ~src ~dst:(Path.relative dst name)))
+    | `Other kind ->
+      User_error.raise
+        ~hints:
+          [ Pp.text
+              "Re-run Dune to delete the stale artifact, or manually delete this file."
+          ]
+        [ Pp.textf
+            "Failed to copy file %s of kind %s while creating a copy sandbox"
```
snowleopard, line 71:
```suggestion
            "Failed to copy file %s of kind %S while creating a copy sandbox"
```

src/dune_engine/sandbox.ml
```diff
@@ -90,15 +103,22 @@ let link_function ~(mode : Sandbox_mode.some) =
...
+         in
+         copy_recursively what ~src ~dst
      | Hardlink ->
        (match Sys.win32 with
         | true -> win32_error mode
         | false -> fun src dst -> Io.portable_hardlink ~src ~dst)
      | Patch_back_source_tree ->
        (* We need to let the action modify its dependencies, so we copy
           dependencies and make them writable. *)
+       (* CR-rgrinberg: this doesn't work with directory targets *)
```
snowleopard, line 121:
```suggestion
       (* CR-someday rgrinberg: this doesn't work with directory targets *)
```

test/blackbox-tests/test-cases/directory-targets/copy-sandboxing.t
```diff
@@ -60,13 +60,13 @@ mode:
...
   755
   permissions of output/y
-  444
+  644
   permissions of output/x
-  444
+  644
   permissions of output/subdir
-  755
+  754
```
snowleopard, line 67: I thought this PR only adds permissions. Here we are losing some?

src/dune_engine/sandbox.ml
```diff
@@ -90,15 +103,22 @@ let link_function ~(mode : Sandbox_mode.some) =
         | false -> fun src dst -> Io.portable_symlink ~src ~dst)
      | Copy ->
        fun src dst ->
-         let { Unix.st_kind; _ } = Path.Untracked.stat_exn src in
-         copy_recursively st_kind ~src ~dst
+         let what =
+           let { Unix.st_kind; st_perm; _ } = Path.Untracked.stat_exn src in
+           match st_kind with
+           | S_DIR -> `Dir st_perm
```
snowleopard, line 109: Let's create a regular type for this variant and put it in a module with something like `of_unix_stats` to avoid duplicating this code.

src/dune_engine/sandbox.ml
```diff
@@ -90,15 +103,22 @@ let link_function ~(mode : Sandbox_mode.some) =
...
+         in
+         copy_recursively what ~src ~dst
      | Hardlink ->
        (match Sys.win32 with
         | true -> win32_error mode
         | false -> fun src dst -> Io.portable_hardlink ~src ~dst)
      | Patch_back_source_tree ->
        (* We need to let the action modify its dependencies, so we copy
           dependencies and make them writable. *)
+       (* CR-someday rgrinberg: this doesn't work with directory targets *)
```
snowleopard, line 121: I'm confused: doesn't `copy_recursively` exist solely for supporting directory targets (or, more accurately, for supporting dependencies on them)? It seems odd to go through all the trouble and not support directory targets in the end.

snowleopard (commented):
Thanks! Generally looks good but I've added some suggestions/questions.

src/dune_engine/sandbox.ml
rgrinberg: I think @emillon just forgot to add full support for them here. I have upcoming PR's that will do so though. This PR is just focusing on the permissions.

src/dune_engine/sandbox.ml
rgrinberg: I could do a regular variant, but it actually won't prevent duplicating the code. That's because:

1. At this call site, we get the file kind from `Unix.stat`
2. In the other call sites, we get the file kind from `Path.readdir_with_kinds`

Would you still like the regular variant?

test/blackbox-tests/test-cases/directory-targets/copy-sandboxing.t
rgrinberg: It's because we are now copying the permissions of directories correctly. Previously, we'd just do `Path.mkdir`, now we make sure we create the directory with the correct permissions.

src/dune_engine/sandbox.ml
```diff
@@ -41,26 +41,39 @@ type t =
...
+  let chmod = Path.Permissions.add Path.Permissions.write in
+  let rec loop what ~src ~dst =
+    match what with
+    | `File -> Io.copy_file ~chmod ~src ~dst ()
+    | `Dir perms ->
+      (match Path.Untracked.readdir_unsorted_with_kinds src with
+       | Error e -> Unix_error.Detailed.raise e
+       | Ok contents ->
+         let perms = chmod perms in
+         Path.mkdir_p ~perms dst;
```
rgrinberg, line 54: @snowleopard this is the line that makes it appear as if we're losing permissions. But we're actually just copying them accurately. Previously, we would just do `Path.mkdir_p` which always created things as `755`.

I could move this into a separate PR if you prefer, but to me it seemed like a similar enough of a bug (incorrect permissions) to just stuff it here.

snowleopard: Ah, I see, thanks.

src/dune_engine/sandbox.ml
snowleopard: Yes, I think so. Converting stats around is boring code, and it's best to keep it in a little helper module.

src/dune_engine/sandbox.ml
rgrinberg: @snowleopard actually, should we also add executable permissions to the directory? Without executable permissions, it's not possible to write anything to the directory anyway.

snowleopard: Yeah, we might as well do that. I guess it doesn't matter much in practice because unreadable directories just don't show up often.

rgrinberg:
@snowleopard addressed all the comments

snowleopard (approved):
Looks good, thanks!

PR #8922, opened 2023-10-12
fix(pkg): improve invalid substitute error

rgrinberg:
When substitute tries to escape the build directory, give the user a
proper error.

### DIFF
--- a/src/dune_rules/expander0.ml
+++ b/src/dune_rules/expander0.ml
@@ -11,3 +11,16 @@ let isn't_allowed_in_this_position ~(source : Dune_lang.Template.Pform.t) =
   in
   raise (User_error.E exn)
 ;;
+
+let as_in_build_dir ~what ~loc p =
+  match Path.as_in_build_dir p with
+  | Some p -> p
+  | None ->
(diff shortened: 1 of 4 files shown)


src/dune_rules/expander0.ml
```diff
@@ -11,3 +11,16 @@ let isn't_allowed_in_this_position ~(source : Dune_lang.Template.Pform.t) =
...
 ;;
+
+let as_in_build_dir ~what ~loc p =
+  match Path.as_in_build_dir p with
+  | Some p -> p
+  | None ->
+    User_error.raise
+      ~loc
+      [ Pp.textf
+          "%s %s is outside the build directory. This is not allowed."
```
Alizter, line 22: So the error will say:
```
susbstitute foo is outside the build directory. This is not allowed.
```
Perhaps it would read better if it said "sustitute: foo is outside the build directory".

rgrinberg: It's inconsistent with the error messages. This function should be modified if the error message isn't good. Although it seems good enough for me.

PR #8923, opened 2023-10-12
fix(pkg): require copying sandbox for build rules

rgrinberg:

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -805,15 +805,18 @@ module Action_expander = struct
     }
   ;;
 
-  let expand context (pkg : Pkg.t) action =
-    let* expander = expander context pkg in
-    let+ action =
-      expand action ~expander >>| Action.chdir (Path.build pkg.paths.source_dir)
-    in
-    (* TODO copying is needed because patch/substitute might be present *)
-    Action.Full.make ~sandbox:Sandbox_config.needs_sandboxing action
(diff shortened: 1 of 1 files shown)


src/dune_rules/pkg_rules.ml
```diff
@@ -805,15 +805,17 @@ module Action_expander = struct
...
-    |> Action_builder.return
-    |> Action_builder.with_no_targets
+  let expand =
+    let sandbox = Sandbox_mode.Set.singleton Sandbox_mode.copy in
+    fun context (pkg : Pkg.t) action ->
+      let* expander = expander context pkg in
+      let+ action =
+        expand action ~expander >>| Action.chdir (Path.build pkg.paths.source_dir)
+      in
+      (* TODO copying is needed because patch/substitute might be present *)
```
Alizter, line 815: TODO can be removed but the rest left as a comment. It's not just patch/substitute to but weird self modifying build system setups like OCaml's makefiles.

Alizter (approved):

src/dune_rules/pkg_rules.ml
rgrinberg: Indeed. I'm going to use the copying sandbox under the following circumstances:
1. the build system of the package isn't dune.
2. there's an explicit install step

Alizter: That seems sensible. We'll see if any exceptions arise in practice.

PR #8924, opened 2023-10-12
fix: stale directory target check

rgrinberg:
The current directory check swallows permission and other errors and
just says the directory doesn't exist.

We improve the check to emit a correct error every time.

We should probably reimplement `Path.exists` to stop being relying on `Sys.file_exists` and use `Unix`.

### DIFF
--- a/src/dune_engine/sandbox.ml
+++ b/src/dune_engine/sandbox.ml
@@ -282,6 +282,13 @@ let apply_changes_to_source_tree t ~old_snapshot =
        | Lt | Gt -> copy_file p))
 ;;
 
+let hint_delete_dir =
+  [ Pp.text
+      "delete this file manually or check the permissions of the parent directory of \
+       this file"
+  ]
+;;
+
 let move_targets_to_build_dir t ~loc ~should_be_skipped ~(targets : Targets.Validated.t)
   : unit Targets.Produced.t Fiber.t
   =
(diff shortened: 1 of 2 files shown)


src/dune_engine/sandbox.ml
```diff
@@ -295,13 +295,25 @@ let move_targets_to_build_dir t ~loc ~should_be_skipped ~(targets : Targets.Vali
...
+               ]
+             [ Pp.textf "unable to stat %s" (Path.Build.to_string_maybe_quoted target)
+             ; Pp.text "reason:"
+             ; Pp.text (Unix_error.Detailed.to_string_hum e)
+             ]
+         | Ok { Unix.st_kind; _ } ->
+           (* We clean up all targets (including directory targets) before running an
+              action, so this branch should be unreachable. *)
+           Code_error.raise
+             "Stale directory target in the build directory"
```
rgrinberg, line 315: Should this be a Code error? I feel like this is an issue poorly behaved rules can introduce.

snowleopard: Can you demonstrate that with a test?

Alizter: We are hitting this error when building the compiler, for some reason the sandbox is still around. I still don't fully understand the issue so we haven't been able to make a minimal reproduction yet.

rgrinberg: A test has been added. I also changed it to a User_error because otherwise the test isn't reproducible.

snowleopard: Thanks! The test build doesn't seem to fail though?

Alizter: I think its failing correctly in the CI, just isn't updated.

rgrinberg: Ah yep. It's updated now.

src/dune_engine/sandbox.ml
```diff
@@ -295,13 +302,26 @@ let move_targets_to_build_dir t ~loc ~should_be_skipped ~(targets : Targets.Vali
...
+             ; Pp.text (Unix_error.Detailed.to_string_hum e)
+             ]
+         | Ok { Unix.st_kind; _ } ->
+           (* We clean up all targets (including directory targets) before
+              running an action, so this branch should be unreachable unless
+              the rule somehow escaped the sandbox *)
+           User_error.raise
+             ~hints:hint_delete_dir
+             [ Pp.textf
+                 "Target %s of type %s already exists in the build directory"
```
snowleopard, line 321:
```suggestion
                 "Target %s of kind %S already exists in the build directory"
```

src/dune_engine/sandbox.ml
```diff
@@ -282,6 +282,13 @@ let apply_changes_to_source_tree t ~old_snapshot =
        | Lt | Gt -> copy_file p))
 ;;
 
+let hint_delete_dir =
+  [ Pp.text
+      "delete this file manually or check the permissions of the parent directory of \
+       this file"
```
snowleopard, line 288:
```suggestion
      "Delete this file manually or check the permissions of the parent directory of \
       this file."
```

snowleopard (approved):
Looks good, thanks! Made a little suggestion.

PR #8926, opened 2023-10-12
Add a cram alias for the pkg tests

gridbugs:
This allows all the tests defined in test/blackbox-tests/test-cases/pkg to be run with a single command (the command is `dune build @test/pkg`).

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/dune
+++ b/test/blackbox-tests/test-cases/pkg/dune
@@ -1,4 +1,8 @@
 (cram
+ (alias pkg)
+ (applies_to :whole_subtree))
+
+(cram
  (deps helpers.sh)
  (applies_to :whole_subtree))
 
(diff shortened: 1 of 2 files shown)


gridbugs:
My main motivation for this is to help me develop on macos. Some of the test don't work on macos so I can't run `make test` but all of the pkg tests work fine and those are usually the tests I'm most interested in anyway.

Alizter:
A little verbose but you can do the same with:
```
./dune.exe build @test/blackbox-tests/test-cases/pkg/runtest @test/expect-tests/dune_pkg/runtest
```
In addition to running the expect tests.

Also the alias stanza depending on those two aliases recursively is probably what you want.

Another thing we could do is add an `(include)` stanza to our `dune-file` then people could drop their own stanzas in as they please. This doesn't work when the file doesn't exist though so a bit of a pain.

rgrinberg (commented):
Even less verbose is `$ dune runtest test/blackbox-tests/test-cases/pkg`.

Maybe this alias should add the pkg expect tests so that it's not redundant?

gridbugs:
Good point about the expect tests. I've added those to the alias too.

rgrinberg (approved):

PR #8927, opened 2023-10-12
Command.run: remove comment about API

emillon:
After trying to change the type, it turns out that adding `S` everywhere is not going in the right direction.

### DIFF
--- a/src/dune_rules/command.mli
+++ b/src/dune_rules/command.mli
@@ -71,8 +71,6 @@ module Args : sig
   val as_any : without_targets t -> any t
 end
 
-(* TODO: Using list in [with_targets t list] complicates the API unnecessarily:
-   we can use the constructor [S] to concatenate lists instead. *)
 val run
   :  dir:Path.t
   -> ?sandbox:Sandbox_config.t


emillon:
I wanted to address the TODO item as a PR. Personally I think it's good to use `Args.t` instead of `Args.t list` and list manipulation, but if we don't think it's the right solution I'm fine with just removing the TODO item instead.

src/dune_rules/command.mli
```diff
@@ -72,21 +72,19 @@ module Args : sig
...
 
-(* TODO: Using list in [with_targets t list] complicates the API unnecessarily:
-   we can use the constructor [S] to concatenate lists instead. *)
 val run
   :  dir:Path.t
   -> ?sandbox:Sandbox_config.t
   -> ?stdout_to:Path.Build.t
   -> Action.Prog.t
-  -> Args.any Args.t list
+  -> Args.any Args.t
```
Alizter, line 80: I definitely prefer this as it now encourages the right way to think about arguments. It's pretty evident that everybody was confused whether to use `Args.t list` or `Args.t` before.

Alizter: But maybe I spoke to soon since it makes calling `run` almost always more awkward.

rgrinberg:
I don't think this is a good change tbh. The point of taking the list was to be symmetric with `Process.run`.

We can see this by how much noise this PR adds.

rgrinberg:
Yes, that sounds good.

PR #4605, opened 2021-05-13
Replace Dune_engine.Dep_path by a more general mechanism in Memo

jeremiedimino:
The goal of `Dep_path` was to accumulate the path from the main goal to an error. It was working by wrapping exceptions and adding a `Dep_path.Entry.t list` to them. At certain strategic points, we were catching and re-raising exceptions with an augmented dependency path.

This feature is now redundant with the memo stack. Given that, this PR moves the wrapping to `Memo` and removes the forward references `Memo.unwrap_exn`. It also allows to attach "human readable description" to memo nodes. To print the dependency path, we now simply need to extract human readable description of stack frames from the memo stack. These descriptions are also used when displaying dependency cycles. A few error messages changed as a result of these changes. The new messages seem fine to me but they could be tweaked if needed.

Since exceptions raised by the `Memo` module now systematically have a stack attached to them, I removed the stack from `Memo.Cycle_error.t`.

### Recovering locations

Another feature of `Dep_path` was to "augment" error messages with a location recoved from the dependency path for messages that didn't have a location already.

~To preserve this feature, it is also possible to attach a location to stack frames. In the end, both the human readable description and the location are captured in a new `Memo.Stack_frame.Info.t` record:~

I replaced this feature by a simpler one where we insert the location at the point where we execute a rule. This seems to produce better error message, as the previously recovered location where pointing to the use site rather than the definition site.

### Less exceptions related functions in `Memo.Build`

After these changes, the only place in the code where we observe exceptions in the memo monad is in `merlin.ml` where we ignore failures related to computing the compilation flags. So I removed the various `with_error_handler`, `collect_error`, ... functions from `Memo.Build` and replaced them by a single `swallow_errors` one:

```ocaml
  val swallow_errors : (unit -> 'a t) -> ('a, unit) Result.t t
```

This should give us more freedom to deal with exceptions inside Memo. And if we drop this last use case in merlin.ml, we could remove this last function, which would give us even more freedom inside Memo.

PR #8928, opened 2023-10-12
Remove Command.Args.Fail

emillon:
This has not been used for a while.


emillon:
I don't know. I could dig up a use removed in #4605, not sure it's the last one.

rgrinberg (approved):

PR #8929, opened 2023-10-12
test(dune_console): colors and styles

Alizter:
## Problem

Displaying the correct styles and colors in the terminal can have subtle bugs that can take a while to notice. Especially since it is something that end users will see the most, but they might not understand that something is a bug.

We have very few tests specifying how we encode our styles into ansi escape codes. And we don't have an easy way to manually check them either.

## Solution

We create a test binary called `dune_console_styles.exe` that outputs a "document" exercising the various styles and colors we support. When passing `--tui` it also whirls up the scheduler so you can interact with TUI and view the same document this time via the NoTTY renderer. This has the following advantages:

- We can quickly debug visual issues in different terminal emulators.
- We can manually inspect how we are translating styles into ansi codes. (There are bugs!)
- The ansi codes for our User_message styles are recorded in a cram test `console-styles.t`.
- We can easily compare the output of TUI vs regular dune console. (There are bugs!)

## Bugs

Having this test proved very useful and allowed us to identify the following bugs:
- We incorrectly translate both foreground and background 8-bit and 24-bit colors into background ansi codes.
  - A fix has been prepared and will be submitted after this PR. 
- The "dim" style in NoTTY is incorrect. 
  - An upstream fork fix was made here: https://github.com/ocaml-dune/notty/pull/8. This will be pulled in soon.

## Screenshots
image

image

### DIFF
--- a/test/blackbox-tests/utils/dune
+++ b/test/blackbox-tests/utils/dune
@@ -22,3 +22,17 @@
 (executable
  (name refmt)
  (modules refmt))
+
+(executable
+ (name dune_console_styles)
+ (modules dune_console_styles)
+ (libraries
+  stdune
+  fiber
+  dune_config_file
+  dune_config
+  dune_console
+  cmdliner
+  dune_util
+  dune_engine
+  dune_tui))
--- a/test/blackbox-tests/test-cases/console-styles.t
+++ b/test/blackbox-tests/test-cases/console-styles.t
@@ -1,0 +1,74 @@
+We test the different styles that dune_console can output. This is done by running a
+special binary that can be run separately for manual inspection. It will simply dump the
+ansi codes used here so that any changes to how styles are rendered can be easily seen.
+
+Trick dune into thinking this is a color terminal:
+  $ CLICOLOR_FORCE=1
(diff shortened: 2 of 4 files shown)


test/blackbox-tests/utils/dune
```diff
@@ -22,3 +22,17 @@
 (executable
  (name refmt)
  (modules refmt))
+
+(executable
+ (name dune_console_styles)
+ (modules dune_console_styles)
```
rgrinberg, line 28: what about juts using an expect test rather than introducing a dummy binary?

rgrinberg (changes requested):

test/blackbox-tests/utils/dune
Alizter: I prefer the binary because then I can run that manually.

rgrinberg: Okay, that’s not enough of a reason. You can run inline tests manually as well.

Alizter:
There are some disagreements about the design here that I don't know how to resolve so I am closing this. If we want to add something like this in the future we can always reopen.

PR #8930, opened 2023-10-12
Move more trace file maintenance into Dune_stats

rleshchinskiy:
This is just a small refactoring to move more logic for dealing with trace
files into Dune_stats. This is in preparation for more significant changes
in this area to enable trace files by default.


src/dune_stats/dune_stats.ml
```diff
@@ -141,23 +143,33 @@ let set_global t =
...
+  let close_once f () =
+    if not !closed
+    then (
+      f ();
+      closed := true)
   in
-  let close =
+  let print, close_, flush =
+    let mkout out =
+      Stdlib.output_string out, (fun () -> Stdlib.close_out out), fun () -> flush out
```
Alizter, line 155: Probably better to use `Io` equivalents here.

Alizter: I'll do that in a later PR then.

rgrinberg:
Has been subsumed by later PR's and we're quite close to enabling the trace file by default in main.

Issue #7049, opened 2023-02-12
Engine: Contexts


rgrinberg:
This is another idea from a discussion with @snowleopard. I thought I would write it down for the record and to give an opportunity for other curious souls to participate. The end goal of this proposal is to demonstrate why contexts can, and should be implemented outside of the engine, with a plan on how to get there.

# Past

Contexts have a long history and have existed since the earliest versions of jbuilder. Therefore it will be helpful to talk about why they were introduced and what were the problems they were meant to solve.

Multi contexts builds were introduced to solve the simple problem of testing dune against multiple versions of the compiler. As such, they were nothing more than parallel builds of the same rules with different configurations, where each configuration was an opam switch. Shortly after, it was realized that they also neatly solved the problem of cross compilation. In a cross compilation setup, the "target" context would delegate the running of all build tools defined in the workspace to the "host" context automatically.

Eventually, a bunch of other things were grafted on. Everything from setting profiles per context, designating a merlin context, setting instrumentation. Fast forward to today, there's demand for customizing ignored directories per context, dune automatically constructing contexts using opam, overriding compiler configuration, and a dozen of other features I might have missed.

# Present

The problem with contexts is that it's one of the only features that partially exists in both the rules in the engine. One must understand how contexts work to make sense of the engine, but all the powerful features offered by contexts such as cross compilation are actually provided by the rules. Worst of all, it's not possible to opt out of contexts at all, even though that truthfully, they're a rarely used feature that consumes a lot of memory when used to its full power.

# Future

Our goal should be to make the engine context agnostic and have the context implementation live in one place (the rules, or a separate library). To do that, we need to consider the different ways the engine is aware of different contexts today.

## Rule Loading

Our rule loading scheme has special context aware implicit rules for copying from the source directory. However, there's no reason why the build configuration callback can't setup the equivalent copying rules.

## Promotions, Error Messages

When registering promotions, the engine derives the source path of the file from the build path by stripping the build context from the path. Similarly, when reporting errors, the engine strips the context to translate build paths to source paths to simplify error messages. If the engine were to be context oblivious, we would need to provide a callback to translate build paths to source paths when appropriate.

## Directory Targets

Currently, it's not possible for dune to depend on directories that were copied from the source even though it's possible to depend on directories generated by rules. One of the reasons why this is hard to fix is all the special casing of source directories in the rule loading. If this special casing was to be removed, I imagine it would be much simpler to fix this bug.

snowleopard:
[...]

PR #8932, opened 2023-10-12
feature(engine): allow customizing execution parameters per context

rgrinberg:
Contexts are a way for users to customize the rules in a workspace and build all these workspaces concurrently. This PR makes it possible to customize the execution parameters per context, giving us additional customization possibilities for free.

In addition, it's also good to make the engine as agnostic to contexts as possible. It gives the rules more flexibility and makes the engine assume less about the rules.

Here's a related ticket that describes the motivation more: https://github.com/ocaml/dune/issues/7049

### DIFF
--- a/src/dune_engine/build_config.ml
+++ b/src/dune_engine/build_config.ml
@@ -77,7 +77,7 @@ type t =
   ; cache_config : Dune_cache.Config.t
   ; cache_debug_flags : Cache_debug_flags.t
   ; implicit_default_alias : Path.Build.t -> unit Action_builder.t option Memo.t
-  ; execution_parameters : dir:Path.Source.t -> Execution_parameters.t Memo.t
+  ; execution_parameters : dir:Path.Build.t -> Execution_parameters.t Memo.t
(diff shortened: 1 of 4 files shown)


rgrinberg:
@pmwhite could you review this one as well?

pmwhite:
Looks good. I don't see where the behavior changed, though. Would it be easy enough to write a test that demonstrates using different execution parameters in different contexts? Or if that's too much of a pain, maybe just elaborate in this PR a bit more. What you said seems clear enough to someone with more context, but I don't have very much - at least not about this corner of Dune's feature set.

pmwhite (approved):

rgrinberg:
>  Would it be easy enough to write a test that demonstrates using different execution parameters in different contexts?

It should be easy, but there's no way to do this yet. We can only modify the execution context via project files and there's no way to do that yet.

I'll update the description to include more context.

pmwhite:
Ahh, I see now. We're expanding what the engine can do, but not changing behavior for existing use-cases. Makes sense! Thanks for explaining.

PR #8934, opened 2023-10-13
Remove flags from solver_env

gridbugs:
We want it to be possible for users to choose whether to build test dependencies at build time rather than at solve time. This means that when generating a lockdir we must effectively set the opam flag "with-test" to true and include all test dependencies of the local package (but not of its dependencies).

Also, there doesn't appear to be a use case for including the documentation-only dependencies in the lockdir.

For these two reasons, the "with-test" and "with-doc" flags are removed from the solver env.


test/blackbox-tests/test-cases/pkg/just-print-solver-env.t
```diff
@@ -30,28 +27,27 @@ Add some build contexts with different environments
   >   (name no-doc)
   >   (lock dune.linux.lock)
   >   (solver_env
-  >    (flags (:standard \ with-doc)))))
+  >    (sys
```
Leonidas-from-XIV, line 30: It's a bit strange that this appears here now, given only the flags were removed?

Leonidas-from-XIV (commented):
The removal is pretty uncontroversial, however the changes in the variables are a surprise to me, why is this happening?

rgrinberg:
> Also, there doesn't appear to be a use case for including the documentation-only dependencies in the lockdir.

I would say that this isn't quite precise. There's no use case for including documentation only dependencies in a lock dir that's used for development or release, but people could find a use for a lock directory that includes documentation dependencies to build and share documentation websites for example. 

In any case, we need to think about the documentation workflow from scratch before we implement anything.

By the way, I still think that users should be able to override dune settings and provide whatever flags they want. E.g. they should be able to set things like `with-dev-setup`.

I think we should modify this PR by just unifying "sys" with all the other variables set in the workspace. I don't see why the user should care about distinguishing them. So for example:

```
(solver_env
 (variables
  with-dev-setup
  (arch bar)
  (os foo)))
```

@gridbugs do you see a reason for users to distinguish these?

test/blackbox-tests/test-cases/pkg/just-print-solver-env.t
gridbugs: This test exercises the code for printing the solver environment and it previously did this by setting some flags in the environment and then printing them. With the removal of flags this test would be left with an empty environment so I set some other variables so the printing code still gets tested.

test/blackbox-tests/test-cases/pkg/just-print-solver-env.t
```diff
@@ -30,28 +27,27 @@ Add some build contexts with different environments
...
-    - arch (unset)
-    - os (unset)
-    - os-version (unset)
-    - os-distribution (unset)
-    - os-family (unset)
+    - arch = x86_64
+    - os = linux
+    - os-version = 22.04
+    - os-distribution = ubuntu
+    - os-family = ubuntu
```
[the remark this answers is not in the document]
gridbugs: It's just so the variable environment isn't empty for this test.

gridbugs:
> ```
> 
> ```
> 
> 
>     
>   
> 
> @gridbugs do you see a reason for users to distinguish these?

There's no reason to distinguish them. The only reason they're currently separate is because they're in separate fields of `Solver_env.t` and the decoder for that type follows its structure.

rgrinberg (approved):
We can add a way to override arbitrary variable in a subsequent PR.

I do think we should get rid of the distinction between "sys", "flags" and other variables in the stanza.

gridbugs:
Makes sense. There's no reason to expose the distinction to users. Will update in a future PR.

gridbugs:
Also it might make sense to remove the distinction altogether. Originally I thought it would be useful to only give the solver access to specific variables rather than allowing arbitrary variables to be configured but now I feel like that's too specific to the variables in opam and we should just allow arbitrary variables to be set in the solver variable environment. I'll think a bit more about this.

gridbugs:
@rgrinberg I've written down my thoughts regarding setting solver variables in the workspace file here: https://github.com/ocaml/dune/issues/8969

PR #8935, opened 2023-10-13
Refactor solver workspace fields

gridbugs:
This renames the fields in the workspace config that affect the solver
    environment. Namely, "sys" is renamed to "override_solver_sys_vars" to
    reflect the fact that they override the sys vars obtained by polling the
    system, and "repositories" is renamed to "default_repositories" to
    reflect the fact that those repositories are only used if no other repos
    were specified.

### DIFF
--- a/src/dune_rules/workspace.ml
+++ b/src/dune_rules/workspace.ml
@@ -279,20 +279,32 @@ module Context = struct
       { base : Common.t
       ; lock : Path.Source.t option
       ; version_preference : Dune_pkg.Version_preference.t option
-      ; solver_env : Dune_pkg.Solver_env.t option
+      ; solver_sys_vars : Dune_pkg.Solver_env.Variable.Sys.Bindings.t option
+      ; repositories : Dune_pkg.Pkg_workspace.Repository.Name.t list
       }
 
-    let to_dyn { base; lock; version_preference; solver_env } =
+    let to_dyn { base; lock; version_preference; solver_sys_vars; repositories } =
       Dyn.record
         [ "base", Common.to_dyn base
         ; "lock", Dyn.(option Path.Source.to_dyn) lock
         ; ( "version_preference"
           , Dyn.option Dune_pkg.Version_preference.to_dyn version_preference )
-        ; "solver_env", Dyn.option Dune_pkg.Solver_env.to_dyn solver_env
+        ; ( "solver_sys_vars"
+          , Dyn.option Dune_pkg.Solver_env.Variable.Sys.Bindings.to_dyn solver_sys_vars )
+        ; ( "repositories"
+          , Dyn.list Dune_pkg.Pkg_workspace.Repository.Name.to_dyn repositories )
         ]
     ;;
 
     let decode =
+      let repositories_of_ordered_set ordered_set =
+        Dune_lang.Ordered_set_lang.eval
+          ordered_set
+          ~parse:(fun ~loc string ->
+            Dune_pkg.Pkg_workspace.Repository.Name.parse_string_exn (loc, string))
+          ~eq:Dune_pkg.Pkg_workspace.Repository.Name.equal
+          ~standard:[ Dune_pkg.Pkg_workspace.Repository.Name.of_string "default" ]
+      in
       let+ common = Common.decode
       and+ name =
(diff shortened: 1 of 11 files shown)


gridbugs:
Note that this subsumes https://github.com/ocaml/dune/pull/8934 so please merge that first. I split this into its own PR as its more controversial than its parent and I didn't want to block it.

src/dune_rules/workspace.ml
```diff
@@ -279,20 +279,36 @@ module Context = struct
...
     ;;
 
     let decode =
+      let default_repositories_of_ordered_set ordered_set =
+        Dune_lang.Ordered_set_lang.eval
+          ordered_set
+          ~parse:(fun ~loc string ->
+            Dune_pkg.Pkg_workspace.Repository.Name.parse_string_exn (loc, string))
+          ~eq:Dune_pkg.Pkg_workspace.Repository.Name.equal
+          ~standard:[ Dune_pkg.Pkg_workspace.Repository.Name.of_string "default" ]
```
Leonidas-from-XIV, line 310: While at it, it would be good to unify `:standard` with the `default` repositories, to avoid the case that `:standard` and `` (no stanza) end up having different defaults.

(One could also refactor this where the name would be taken from the predefined repository, but I don't think hoisting more work onto this PR is sensible, better get iterative improvements).

test/blackbox-tests/test-cases/pkg/env-conditional-dependencies.t
```diff
@@ -33,16 +33,14 @@ Create a workspace config that defines separate build contexts for macos and lin
   >  (default
   >   (name linux)
   >   (lock dune.linux.lock)
-  >   (solver_env
-  >    (sys
-  >     (os linux)))))
+  >   (override_solver_sys_vars
```
Leonidas-from-XIV, line 36: I think my concern would that these names are getting increasingly complex and hard to understand. A user doesn't really know (nor should they really need to care) what solver sys vars are.

test/blackbox-tests/test-cases/pkg/multiple-opam-repos.t
```diff
@@ -38,8 +38,7 @@ We have to define both repositories in the workspace, but will only use `new`.
   > (context
   >  (default
   >   (name default)
-  >   (solver_env
-  >    (repositories new))))
+  >   (fallback_repositories new)))
```
Leonidas-from-XIV, line 41: Likewise. What are fallback repositories? What are they falling back from? Where can I specify normal repositories? Even as someone involved in dune development, this confuses me.

Leonidas-from-XIV (commented):
I think the code is fine, but the new naming makes things very confusing I feel.

test/blackbox-tests/test-cases/pkg/env-conditional-dependencies.t
Alizter: But should a user be reading this in the first place?

rgrinberg:
Same comments as Marek's basically. "override" and "fallback" aren't helpful. For the same reason our "flags" fields aren't called "override_flags".

> A user doesn't really know (nor should they really need to care) what solver sys vars are.

Agreed. Though I think that Stephen is already planning simplify this in https://github.com/ocaml/dune/issues/8969

gridbugs:
Thanks for the feedback. I'll get rid of the name changes.

gridbugs:
Ok now this change just refactors the code without changing the names of fields. UI-wise the only change is to remove the `solver_env` field of the workspace conf and to move the user-configurable fields into the workspace directly. For now the only user-configurable variables are system variables but this will change as part of https://github.com/ocaml/dune/issues/8969.

src/dune_rules/workspace.ml
gridbugs: > While at it, it would be good to unify `:standard` with the `default` repositories

@Leonidas-from-XIV mind if I leave this to you? I don't feel like I have context for doing that at the moment.

rgrinberg (approved):

src/dune_rules/workspace.ml
Leonidas-from-XIV: Yeah, sure, it's not urgent at all and not related to this PR really.

PR #8936, opened 2023-10-13
Merge Dep_node_without_state into Dep_node

rleshchinskiy:
It seems that [Dep_node_without_state] doesn't actually need to exist so this merges it into [Dep_node]. The main drawback is that more modules need to be mutually recursive now so the bulk of this PR just moves things around to make that happen. The resulting code still seems clearer to me and also seems to be slightly more efficient since we remove 2 words (I think) and 1 indirection from each [Dep_node].


snowleopard (approved):
Reviewed, battletested and benchmarked internally. Thanks!

PR #8937, opened 2023-10-13
WIP: cache directory targets

emillon:
- Display cache size
- Store Path.Local
- WIP: cache directory targets


emillon:
@rgrinberg just asking if this seems to be the right approach before getting this in good shape. I modified `Metadata_entry` so that it stores more than just the basename and then I try to route the right paths in there. Is this what you had in mind in #8067?

src/dune_cache/local.ml
```diff
@@ -5,8 +5,8 @@ open Import
 
 module Store_artifacts_result = struct
   type t =
-    | Stored of (Path.Build.t * Digest.t) list
-    | Already_present of (Path.Build.t * Digest.t) list
+    | Stored of (Path.Build.t * Digest.t * Path.Build.t option) list
+    | Already_present of (Path.Build.t * Digest.t * Path.Build.t option) list
```
rgrinberg, line 9: It's hard to understand what this argument is. I suggest turning this tuple into a record.

src/dune_engine/targets.mli
```diff
@@ -90,6 +90,11 @@ module Produced : sig
       error if the list contains duplicates. *)
   val of_file_list_exn : (Path.Build.t * Digest.t) list -> Digest.t t
 
+  val make_exn
+    :  files:(Path.Build.t * Digest.t) list
+    -> dirs:Digest.t Filename.Map.t Path.Build.Map.t
```
rgrinberg, line 95: A directory can have subdirectories, so shouldn't this be `Path.Local.Map.t`?

rgrinberg (commented):
Looks like you're on the right track. I would suggest that you add tests for the following two cases:

1. A directory target that produces a file that is already in the cache (as a target of another rule for exampe)
2. Two rules producing directory targets that share a whole subdirectory branch. To demonstrate that we don't (or will) support sharing at the level of entire directories.

emillon:
I added the first test in #8984 .

emillon:
And the second one is in #8986.

src/dune_cache/local.ml
emillon: I turned that into an `Artifact` module which seems useful.

emillon:
@rgrinberg this is ready for another round of reviews. This seems to work but some details are over my head. Some commits are fairly independent so we can also land them separately if necessary.

rgrinberg:
@staronj could you take a look at this one?

rgrinberg:
@emillon AFAIK, @rleshchinskiy is going to take over this and re-export once this is done. Is that ok?

emillon:
We discussed this last week with @snowleopard. JS is going to provide an implementation of this feature. Thanks!

PR #8938, opened 2023-10-13
fix(engine): small optimization for Patch_back_source_tree sandboxes

rgrinberg:
We don't need to [stat] directories when snapshotting the sandbox in
[Patch_back_source_tree].

### DIFF
--- a/src/dune_engine/sandbox.ml
+++ b/src/dune_engine/sandbox.ml
@@ -123,14 +123,14 @@ let snapshot t =
   (* CR-someday jeremiedimino: we do this kind of traversal in other places.
      Might be worth trying to factorise the code. *)
   let rec walk dir acc =
-    match Path.Untracked.readdir_unsorted dir with
+    match Path.Untracked.readdir_unsorted_with_kinds dir with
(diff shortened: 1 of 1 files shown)


rgrinberg:
@dkalinichenko-js could you review this one?

dkalinichenko-js (approved):

snowleopard:
@dkalinichenko-js Could you please downstream this one?

dkalinichenko-js:
> @dkalinichenko-js Could you please downstream this one?

This change is originally from @pmwhite fyi, it's already in the code.

rgrinberg:
@dkalinichenko-js This change has been downstreamed by @pmwhite already? Am I getting that right?

If so, whoever did should remove the `upstream-sync` label.

pmwhite:
@dkalinichenko-js I think you are referring to https://github.com/ocaml/dune/pull/8705. It's a very similar piece of code to this one, but in a different part of this file. I don't think we have this change internally yet. (for what it's worth, I came to the same conclusion at first and was very confused, because I didn't remember importing this change)

PR #8940, opened 2023-10-14
feature: allow custom actions to be dynamic

rgrinberg:
Currently, custom actions are less powerful than normal actions in one specific way:

They aren't allowed to build dependencies as part of their execution. In particular, a custom action cannot implement `dynamic-run`.

Fortunately, extending custom actions with this ability ends up being quite simple.


snowleopard (approved):
Looks good!

PR #8943, opened 2023-10-15
chore: improve error message for temp dir failure

rgrinberg:
I improve the error messages for temporary directory creation in the following way:

1. If we try to create inside a non existing directory, the directory name that we failed to create will be included in the error.
2. If the suffix or prefix contain path elements (e.g. `/` on unix), the directory will be rejected. This isn't allowed because it will end up pointing to subdirectories in the temp directory which haven't been created yet.

### DIFF
--- a/otherlibs/stdune/src/temp.ml
+++ b/otherlibs/stdune/src/temp.ml
@@ -44,7 +44,9 @@ let create_temp_dir ?perms path =
   | Created -> Ok ()
   | Already_exists -> Error `Retry
   | Missing_parent_directory ->
-    Code_error.raise "[Temp.create_temp_dir] called in a non-existing directory" []
+    Code_error.raise
+      "[Temp.create_temp_dir] called in a nonexistent directory"
+      [ "dir", Path.to_dyn path ]
 ;;
 
 let set = function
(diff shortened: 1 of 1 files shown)


rgrinberg:
@pmwhite do you want to review this one?

otherlibs/stdune/src/temp.ml
```diff
@@ -44,7 +44,9 @@ let create_temp_dir ?perms path =
   | Created -> Ok ()
   | Already_exists -> Error `Retry
   | Missing_parent_directory ->
-    Code_error.raise "[Temp.create_temp_dir] called in a non-existing directory" []
+    Code_error.raise
+      "[Temp.create_temp_dir] called in a non-existing directory"
```
pmwhite, line 48: While we're at it, perhaps we should also improve the grammar. "Non-existing" should be "nonexistent".

otherlibs/stdune/src/temp.ml
```diff
@@ -89,6 +93,16 @@ let create ?perms what ~prefix ~suffix =
        Perhaps, we should use something like [_build/.temp] instead? *)
     Filename.get_temp_dir_name () |> Path.of_filename_relative_to_initial_cwd
   in
+  if contains_path_sep prefix
+  then
+    Code_error.raise
+      "Temp.create: invalid prefix contains path elements"
```
pmwhite, line 99: The wording is a bit off. The prefix is invalid _because_ it contains path elements, but the error makes it sound like it is an invalid prefix that also contains path elements. I'd suggest "prefix must not contain path elements". Same for the suffix message as well.

pmwhite (approved):
Approving, with 2 small suggestions. Feel free to merge after you consider them.

PR #8950, opened 2023-10-16
feature(pkg): Single repo source

Leonidas-from-XIV:
Currently a work in progress as the integration into the solver returns a `Fiber.t` and the solver needs to be updated to accept a monadic value.

### DIFF
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -317,18 +317,10 @@ module Lock = struct
       let repo_id = Repository_id.of_path path in
       Fiber.return @@ [ Opam_repo.of_opam_repo_dir_path ~source:None ~repo_id path ]
     | None, Some url ->
-      let repo = Fetch.Opam_repository.of_url url in
-      let+ opam_repository = Fetch.Opam_repository.path repo in
-      (match opam_repository with
-       | Ok { path; repo_id } ->
-         [ Opam_repo.of_opam_repo_dir_path
-             ~source:(Some (OpamUrl.to_string url))
-             ~repo_id
-             path
-         ]
-       | Error _ ->
-         User_error.raise
-           [ Pp.text "Can't determine the location of the opam-repository" ])
+      let source = url.OpamUrl.path in
+      let git = Lazy.force Dune_vcs.Vcs.git in
+      let+ opam_repo = Opam_repo.of_git_repo ~git ~repo_id:None source in
+      [ opam_repo ]
     | None, None ->
       repositories
       |> Fiber.parallel_map ~f:(fun name ->
@@ -341,19 +333,9 @@ module Lock = struct
             ]
         | Some repo ->
           let url = Dune_pkg.Pkg_workspace.Repository.opam_url repo in
-          let repo = Fetch.Opam_repository.of_workspace_repo repo in
-          let+ opam_repository = Fetch.Opam_repository.path repo in
-          (match opam_repository with
-           | Ok { path; repo_id } ->
-             Opam_repo.of_opam_repo_dir_path
-               ~source:(Some (OpamUrl.to_string url))
-               ~repo_id
-               path
-           | Error _ ->
-             User_error.raise
-               [ Pp.textf "Can't determine the location of the opam-repository '%s'"
-                 @@ Dune_pkg.Pkg_workspace.Repository.Name.to_string name
-               ]))
+          let source = url.OpamUrl.path in
+          let git = Lazy.force Dune_vcs.Vcs.git in
+          Opam_repo.of_git_repo ~git ~repo_id:None source)
   ;;
 
   let find_local_packages =
@@ -581,7 +563,7 @@ module Outdated = struct
                 ; repositories
                 }
               ->
-              let+ repos =
+              let* repos =
                 Lock.get_repos
                   repos
                   ~opam_repository_path
@@ -589,7 +571,7 @@ module Outdated = struct
                   ~repositories
               and+ local_packages = Lock.find_local_packages in
               let lock_dir = Lock_dir.read_disk lock_dir_path in
-              let results =
+              let+ results =
                 Dune_pkg_outdated.find ~repos ~local_packages lock_dir.packages
               in
               ( Dune_pkg_outdated.pp ~transitive ~lock_dir_path results
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -316,10 +316,8 @@ module Lock = struct
     | Some path, None ->
       let repo_id = Repository_id.of_path path in
       Fiber.return @@ [ Opam_repo.of_opam_repo_dir_path ~source:None ~repo_id path ]
-    | None, Some url ->
-      let source = url.OpamUrl.path in
-      let git = Lazy.force Dune_vcs.Vcs.git in
-      let+ opam_repo = Opam_repo.of_git_repo ~git ~repo_id:None source in
+    | None, Some (url : OpamUrl.t) ->
+      let+ opam_repo = Opam_repo.of_git_repo ~repo_id:None url.path in
       [ opam_repo ]
     | None, None ->
       repositories
@@ -333,9 +331,7 @@ module Lock = struct
             ]
         | Some repo ->
           let url = Dune_pkg.Pkg_workspace.Repository.opam_url repo in
-          let source = url.OpamUrl.path in
-          let git = Lazy.force Dune_vcs.Vcs.git in
-          Opam_repo.of_git_repo ~git ~repo_id:None source)
+          Opam_repo.of_git_repo ~repo_id:None url.path)
   ;;
 
   let find_local_packages =
--- a/bin/pkg.ml
+++ b/bin/pkg.ml
@@ -317,7 +317,7 @@ module Lock = struct
       let repo_id = Repository_id.of_path path in
       Fiber.return @@ [ Opam_repo.of_opam_repo_dir_path ~source:None ~repo_id path ]
     | None, Some (url : OpamUrl.t) ->
-      let+ opam_repo = Opam_repo.of_git_repo ~repo_id:None url.path in
+      let+ opam_repo = Opam_repo.of_git_repo ~repo_id:None ~source:url.path in
       [ opam_repo ]
     | None, None ->
       repositories
@@ -331,7 +331,7 @@ module Lock = struct
             ]
         | Some repo ->
           let url = Dune_pkg.Pkg_workspace.Repository.opam_url repo in
-          Opam_repo.of_git_repo ~repo_id:None url.path)
+          Opam_repo.of_git_repo ~repo_id:None ~source:url.path)
   ;;
 
   let find_local_packages =
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -232,17 +232,6 @@ let scan_files_entries path =
   read [] Path.Local.root
 ;;
 
-module File_entry = struct
-  type source =
-    | Path of Path.t
-    | Content of string
-
-  type t =
-    { local_file : Path.Local.t
-    ; original : source
-    }
-end
-
 let get_opam_package_files t opam_package =
   let open Fiber.O in
   let name = opam_package |> OpamPackage.name |> OpamPackage.Name.to_string in
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -385,7 +385,7 @@ let load_all_versions ts opam_package_name =
 ;;
 
 module Private = struct
-  let create ?source ?repo_id () =
+  let create ~source ~repo_id =
     let packages_dir_path = Path.of_string "/" in
     let serializable =
       Option.map source ~f:(fun source -> { Serializable.repo_id; source })
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -153,10 +153,10 @@ let xdg_repo_location =
   lazy (Xdg.cache_dir (Lazy.force Dune_util.xdg) / "dune/git-repo" |> Path.of_string)
 ;;
 
-let of_git_repo ~git ~repo_id source =
+let of_git_repo ~repo_id source =
   let open Fiber.O in
   let dir = Lazy.force xdg_repo_location in
-  let* repo = Rev_store.load_or_create ~git ~dir in
+  let* repo = Rev_store.load_or_create ~dir in
   let* remote = Rev_store.add_repo repo ~source in
   let+ at_rev, computed_repo_id =
     match repo_id with
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -253,16 +253,13 @@ let get_opam_package_files t opam_package =
        in
        Fiber.return entries)
   | Repo at_rev ->
-    let package_root =
+    let files_root =
       Path.Local.L.relative
         Path.Local.root
-        [ "packages"; name; OpamPackage.to_string opam_package ]
+        [ "packages"; name; OpamPackage.to_string opam_package; "files" ]
     in
-    let* dir_entries =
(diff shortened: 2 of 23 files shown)


bin/pkg.ml
```diff
@@ -401,20 +391,22 @@ module Lock = struct
...
-                 Dune_pkg.Opam_solver.solve_lock_dir
-                   solver_env
-                   version_preference
-                   repos
-                   ~local_packages
-                   ~experimental_translate_opam_filters)
-           with
+           let overlay =
+             Console.Status_line.add_overlay (Constant (Pp.text "Solving for Build Plan"))
+           in
```
Alizter, line 399: Isn't it better to put this in the scope of solution so it only happens when it is run? @rgrinberg please correct me if I'm wrong.

rgrinberg: Yes. It's indeed better that way.

src/dune_pkg/fetch.mli
```diff
@@ -19,6 +19,31 @@ val fetch
   -> OpamUrl.t
   -> (unit, failure) result Fiber.t
 
+module Git_repo : sig
```
rgrinberg, line 22: Do you think you could move this to a separate library and add some unit tests?

bin/pkg.ml
```diff
@@ -302,6 +302,14 @@ module Lock = struct
     ;;
   end
 
+  let get_git =
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Good point!

src/dune_pkg/fetch.ml
```diff
@@ -260,6 +261,123 @@ let fetch ~unpack ~checksum ~target (url : OpamUrl.t) =
...
+    { git : Path.t
+    ; dir : Path.t
+    }
+
+  type rev = Rev of string
+
+  let equal { git; dir } t = Path.equal git t.git && Path.equal dir t.dir
+  let display = Display.Quiet
+  let failure_mode = Process.Failure_mode.Strict
+  let run { git; dir } = Process.run ~dir ~display failure_mode git
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Yep, I just didn't want to deal with it while I was getting out the kinks. Maybe we should really have a new display mode where the commands are silent? @rgrinberg what do you think?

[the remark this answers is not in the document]
Alizter: I think that would necessarily have to be user facing and I am not sure the utility in having no output from Dune. I definitely wouldn't encourage it, but it could be useful for other programs wishing to run dune.

src/dune_pkg/fetch.mli
Leonidas-from-XIV: Will do! But first I'd like to rebase this upon #8962, then I can squash all the intermediate mess.

src/dune_pkg/fetch.ml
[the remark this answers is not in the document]
Leonidas-from-XIV: I don't mean making dune silent, I mean adding a display mode where processes don't write to the users terminal. There is no need for the user to even know `git` is called in the background do create a bare repo on demand.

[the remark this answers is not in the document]
rgrinberg: > Maybe we should really have a new display mode where the commands are silent?

The display mode is something that corresponds to `--display`, so I'm sure if that's the right place. We do have flags like `--action-stdout-on-success`. Maybe they should be the default.

> I mean adding a display mode where processes don't write to the users terminal.

Isn't that what you already did with `make_stdout` and `make_stderr`?

> There is no need for the user to even know git is called in the background do create a bare repo on demand.

If the operation requires network access, I think it's useful to tell the user what is running. I agree that other commands don't need to be displayed.

In any case, the PR doesn't need to fix these issues before it's good to go.

src/dune_pkg/fetch.ml
```diff
@@ -260,6 +261,149 @@ let fetch ~unpack ~checksum ~target (url : OpamUrl.t) =
...
+        let open Fiber.O in
+        (* TODO future: cache ls-tree *)
+        let+ all_files =
+          run_capture_zero_separated_lines
+            repo
+            [ "ls-tree"; "-z"; "--name-only"; "-r"; rev ]
+        in
+        List.filter_map all_files ~f:(fun entry ->
+          let path_entry = Path.of_string entry in
+          Path.descendant path_entry ~of_:path)
```
rgrinberg, line 367: is it possible for this function to return `None`?

src/dune_pkg/opam_repo.ml
```diff
@@ -132,7 +143,45 @@ let of_opam_repo_dir_path ~source ~repo_id opam_repo_dir_path =
...
+    | Some repo_id' ->
+      let+ at_rev = Fetch.Git_repo.Remote.rev_of_repository_id remote repo_id' in
+      at_rev, repo_id
+    | None ->
+      let* branch = Fetch.Git_repo.Remote.default_branch remote in
+      let name =
+        match branch with
+        | Some name -> name
+        | None ->
+          Code_error.raise
```
rgrinberg, line 170: If this is possible, it shouldn't be a `Code_error`. `User_error` would be more appropriate.

src/dune_pkg/opam_repo.ml
```diff
@@ -132,7 +143,45 @@ let of_opam_repo_dir_path ~source ~repo_id opam_repo_dir_path =
...
+            []
+      in
+      let+ at_rev = Fetch.Git_repo.Remote.rev_of_name remote ~name in
+      (match at_rev with
+       | Some at_rev ->
+         Some at_rev, Some (Fetch.Git_repo.Remote.At_rev.repository_id at_rev)
+       | None -> None, None)
+  in
+  match at_rev with
+  | None -> Code_error.raise "Could not find revision in repository" []
```
rgrinberg, line 181: Same. This seems like something that is possible.

src/dune_pkg/opam_repo.ml
```diff
@@ -186,27 +218,108 @@ end
...
-        |> Option.map ~f:(fun (with_file : With_file.t) -> with_file.opam_file)))
+    let pkgs = List.concat pkgs in
+    let* found_versions =
+      Fiber.parallel_map pkgs ~f:(fun opam_pkg ->
+        let+ found_in_repo =
+          Fiber.parallel_map ts ~f:(fun t -> load_opam_package t opam_pkg)
+        in
+        List.find_map found_in_repo ~f:Fun.id)
+    in
+    let finds = List.filter_map found_versions ~f:Fun.id in
```
rgrinberg, line 319: Not related to this PR, but this should be `List.filter_opt`.

src/dune_pkg/fetch.ml
```diff
@@ -260,6 +261,149 @@ let fetch ~unpack ~checksum ~target (url : OpamUrl.t) =
...
+      { repo : t
+      ; handle : string
+      }
+
+    let head_branch = Re.(compile (seq [ str "HEAD branch: "; group (rep1 any); eol ]))
+    let update { repo; handle } = run repo [ "fetch"; handle; "--no-tags" ]
+
+    let default_branch { repo; handle } =
+      let open Fiber.O in
+      let+ contents = run_capture_lines repo [ "remote"; "show"; handle ] in
```
rgrinberg, line 337: `git ls-remote $remote HEAD` is a little more direct.

src/dune_pkg/opam_solver.ml
```diff
@@ -434,17 +435,21 @@ let opam_package_to_lock_file_pkg
...
       | None ->
-        let opam_files =
-          List.filter_map repos ~f:(fun repo ->
-            Opam_repo.load_opam_package repo opam_package)
+        let+ opam_files =
+          let+ pkgs =
+            Fiber.parallel_map repos ~f:(fun repo ->
+              Opam_repo.load_opam_package repo opam_package)
+          in
+          List.filter_map ~f:Fun.id pkgs
```
rgrinberg, line 452: `List.filter_opt`

src/dune_pkg/fetch.ml
```diff
@@ -260,6 +261,149 @@ let fetch ~unpack ~checksum ~target (url : OpamUrl.t) =
...
+    Process.run_capture_zero_separated ~dir ~display failure_mode git
+  ;;
+
+  let show { git; dir } (Rev rev) path =
+    let open Fiber.O in
+    let failure_mode = Dune_vcs.Vcs.git_accept () in
+    let command = [ "show"; sprintf "%s:%s" rev (Path.to_string path) ] in
+    let stderr_to = make_stderr () in
+    let+ res =
+      Process.run_capture_lines ~dir ~display ~stderr_to failure_mode git command
```
rgrinberg, line 303: `run_capture` should be enough to avoid having to concatenate things manually

src/dune_pkg/fetch.mli
```diff
@@ -19,6 +19,33 @@ val fetch
...
+  module Remote : sig
+    type t
+
+    val update : t -> unit Fiber.t
+    val default_branch : t -> string option Fiber.t
+
+    module At_rev : sig
+      type t
+
+      val content : t -> Path.t -> string option Fiber.t
```
rgrinberg, line 35: Perhaps the signature should be changed to `Path.Local.t`? To make sure that we only provide paths inside the repository

src/dune_pkg/fetch.ml
```diff
@@ -260,6 +261,149 @@ let fetch ~unpack ~checksum ~target (url : OpamUrl.t) =
...
+
+    let equal { repo; handle } t = equal repo t.repo && String.equal handle t.handle
+
+    module At_rev = struct
+      type nonrec t =
+        { remote : t
+        ; revision : rev
+        }
+
+      let content { remote = { repo; handle = _ }; revision } path =
```
rgrinberg, line 351: Here's a potential way to speed this function up for our use case. Change the signature to:

```
type size = int
type entries = size Path.Local.Map.t
val entries : t -> Path.t -> size Path.Local.Map.t
val contents : t -> entries -> string Path.Local.Map.
```

To do this, you will need to change you directory listing to include the sizes with `--long`. Then, you can batch all the entries into a single `$ git show` command. Since you know the size of every object, you can parse the output this way.

Now this won't work well on windows because of stupid command line limits.

Other approaches that can be used to optimize this:
1. sparse checkouts
2. using libgit or the OCaml equivalent.

You don't need implement this optimization in this PR, but it's helpful to include the summary of this discussion in the comments.

rgrinberg: Here's a 3rd way to do it:

```
$ git archive $rev:$dir -o $dir.tar
```

And then untar & read the contents of `$dir.tar`.

This way would be quite good once we import an OCaml implementation of tar.

src/dune_pkg/fetch.ml
Leonidas-from-XIV: It doesn't do the same though. `ls-remote` will tell me the HEAD revision of the remote *as it is on the remote* at that moment, but looking up the main branch and rev-parsing `remote/branch-name` will tell me what my local remote's HEAD is (no network interaction). So to be on the safe side I'd need to `ls-remote` and then fetch to make sure I have that revision locally.

src/dune_pkg/fetch.ml
Leonidas-from-XIV: Yes, this iterates over all the files in the repo, so also files like `repo`, for which this would return `None`.

src/dune_pkg/fetch.mli
Leonidas-from-XIV: I've added some expect tests. Do you think unit tests should also be added?

Leonidas-from-XIV:
@rgrinberg I got it in a shape where it seems to work pretty well, I'd appreciate if you could take another look. There are for sure some points that can be optimized (e.g. caching of `ls-tree`, `git archive`, usage of `libgit`/`ocaml-git`) but conceptually these can be added fairly easily if the current implementation turns out to be a bottleneck.

src/dune_pkg/opam_repo.mli
```diff
@@ -21,6 +21,9 @@ val of_opam_repo_dir_path
   -> Path.t
   -> t
 
+(** [of_git_repo git source] loads the data through git *)
+val of_git_repo : git:Path.t -> repo_id:Repository_id.t option -> string -> t Fiber.t
```
Alizter, line 25: What's the point of the `repo_id` argument here?

src/dune_pkg/rev_store.ml
```diff
@@ -0,0 +1,157 @@
...
+    { repo : t
+    ; handle : string
+    }
+
+  let head_branch = Re.(compile (seq [ str "HEAD branch: "; group (rep1 any); eol ]))
+  let update { repo; handle } = run repo [ "fetch"; handle; "--no-tags" ]
+
+  let default_branch { repo; handle } =
+    let open Fiber.O in
+    let+ contents = run_capture_lines repo [ "remote"; "show"; handle ] in
```
Alizter, line 76: Is this meant to be `--show`?

src/dune_pkg/rev_store.mli
Alizter, line 1: I have some questions about the revision store:

How does it handle opam repositories with conflicting entries? Say package `foo` has a different version `1.0` file in repo `A` vs repo `B`?

Alizter:
I left some questions and also double checked that all the `git` commands are valid across all versions of `git` and that seems to be the case.

src/dune_pkg/fetch.ml
```diff
@@ -260,6 +261,149 @@ let fetch ~unpack ~checksum ~target (url : OpamUrl.t) =
...
+        let+ type' = run_capture_line repo [ "cat-file"; "-t"; rev ] in
+        (match type' with
+         | "commit" -> Some { At_rev.remote; revision = Rev rev }
+         | _ -> None)
+    ;;
+  end
+
+  let add_repo t ~source =
+    let open Fiber.O in
+    let handle = source |> Dune_digest.string |> Dune_digest.to_string in
```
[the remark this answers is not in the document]
rgrinberg: Okay, I see. I do think that it would be nice to have remote names that are readable. If nothing else, it would help debugging.

src/dune_pkg/fetch.mli
rgrinberg: I don't see any expect tests, do you mean cram tests?

In any case, yes I think that unit tests (in the form expect tests) should be added. It doesn't have to be in this PR though.

src/dune_pkg/fetch.mli
rgrinberg: `Path.Local.t` is a relative path that doesn't start with `..`. Exactly what we need here. I've modified the code to use it.

rgrinberg:
I removed the `git` field from `Rev_store.t`. It made the equality function on the repo depend on the location of the git binary. That seemed rather surprising to me so I removed it.

src/dune_pkg/rev_store.mli
rgrinberg: There's no issue with it. We store git objects and not concrete files.

src/dune_pkg/opam_repo.mli
rgrinberg: We save it into the lock dir

Leonidas-from-XIV: > What's the point of the `repo_id` argument here? It doesn't seem to be used.

It is used, you can use it to check out a specific revision of the repository. This is useful if you want to reproduce a very specific version of the repo and run the solver again.

src/dune_pkg/rev_store.mli
Leonidas-from-XIV: Exactly. The files are identified by the hash, so `packages/foo/foo.1.0` can be different depending on the revision, and each revision is said to be unique due to git using SHA1 (so indeed we never need to specify which remote a particiular revision is from since there is never any overlap).

Leonidas-from-XIV:
Ah, thanks, I completely missed that `Vcs.git` exists. Saves an annoying lookup step too :-)

src/dune_pkg/rev_store.mli
Alizter: Got it, thanks!

PR #8952, opened 2023-10-17
fix: extra space in fake xdg-open

emillon:


emillon:
cc @rgrinberg @EmileTrotignon

PR #8953, opened 2023-10-17
feat: record installed directories in dune-package

### LINKED ISSUE #8915: (package) dependencies do not work when directory targets are used
When the contents of a package are recorded in `dune-package` files, we don't make a distinction between files and directory targets: just the edge of the directory target is recorded in the `(files)` field.

Later, when `(package p)` is expanded, this causes the following error:

```
Error: File unavailable:
/path/to/share/X
This is not a regular file (S_DIR)
```

There are several ways we could try to fix this:
1. expand the list of all installed files in dune-package (rather than just storing the directory name)
1. recursively copy the contents when something depends on an installed "thing" that happens to be a directory (not sure that's possible or a good idea)
1. add an extra field (dirs) in dune-package files which contains "opaque" directories

The first one seems the most reasonable.

emillon:
Fixes #8915

We add a `(dir PATH)` construct to the `(files)` field in `dune-package` files. It is used by `(package)` dependencies: they are expanded to the recursive contents of the installed directory.


Alizter:
This doesn't record the contents of the directories right? Meaning that you could in theory remove files from the declared directory and dune-package would be OK with that until a directory target (as a dep) is missing a target?

src/dune_rules/dune_package.mli
```diff
@@ -46,14 +46,18 @@ module Entry : sig
   val to_dyn : t Dyn.builder
 end
 
+type file =
```
rgrinberg, line 49: Weird to call this a file when it can be a directory. `path` is less misleading. Or even change the type to:

```
files : (Section.t * [`Directory | `File] * Path.t) list
```

rgrinberg (approved):
LGTM apart from the naming issue.

PR #8954, opened 2023-10-17
doc: document DUNE_CACHE_ROOT in the manpages

### LINKED ISSUE #5678: Document DUNE_CACHE_ROOT
It doesn't seem to be in the man pages but appears in the refman.

emillon:
Fixes #5678

### DIFF
--- a/bin/common.ml
+++ b/bin/common.ml
@@ -1363,6 +1363,9 @@ let envs =
     ; info
         ~doc:"If different than $(b,0), ANSI colors should be enabled no matter what."
         "CLICOLOR_FORCE"
+    ; info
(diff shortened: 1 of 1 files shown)


Alizter:
This begs the question why this isn't a command line option in the first place? Is it being used internally by JS?

Alizter (approved):
Anyway, LGTM. Would be good to be more consistent on these options but that's not for this PR.

emillon:
I think that this boils down to the fact that `dune_cache_storage` is its own library and `Dune_cache_storage.Layout.root_dir` is a plain `Path.t`. So the initialization is done eagerly. I suppose we could hook that to `dune_config` but as you said that's a different issue.

PR #8957, opened 2023-10-17
fix: stop using Pp.newline

rgrinberg:
If we want vertical spacing, we can create a vbox.

This also makes the output of $ dune pkg outdated more uniform. It was
possible to preserve the old output, but it would require even more
special casing. So I just get rid of it all at once.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/outdated.t
+++ b/test/blackbox-tests/test-cases/pkg/outdated.t
@@ -90,7 +90,8 @@ If we remove packages from the repository then we should get a nice error.
   $ outdated 
   1/2 packages in dune.lock are outdated.
   Showing immediate dependencies, use --transitive to see the rest.
-  Error: When checking dune.lock, the following packages:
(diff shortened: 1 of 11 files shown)


test/blackbox-tests/test-cases/pkg/outdated.t
```diff
@@ -140,14 +143,14 @@ With multiple contexts, the errors should also be printed for each context.
...
-    - bar
-    - foo
-    were not found in the following opam repositories:
-    - None
-  - When checking dune.lock, the following packages:
-    - bar
-    - foo
-    were not found in the following opam repositories:
-    - None
+  When checking dune.workspace.lock, the following packages:
```
rgrinberg, line 146: The enumeration here is used for things that don't really belong in a list. I got rid of it.

Alizter: My intention with that was to make it easier to read multiple error messages, but I guess this can be improved in other ways.

PR #8961, opened 2023-10-18
test(stdune): Path.drop_prefix tests

Alizter:
Reproduction for https://github.com/ocaml/dune/pull/8870. A fix for this is also ready after this PR.

### DIFF
--- a/otherlibs/stdune/test/path_tests.ml
+++ b/otherlibs/stdune/test/path_tests.ml
@@ -460,3 +460,37 @@ let%expect_test _ =
 None
 |}]
 ;;
+
+let%expect_test "drop prefix" =
+  Path.drop_prefix ~prefix:(r "foo/bar") (r "foo/bar/baz")
+  |> Dyn.option Path.Local.to_dyn
+  |> print_dyn;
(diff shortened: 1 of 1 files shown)


otherlibs/stdune/test/path_tests.ml
```diff
@@ -460,3 +460,37 @@ let%expect_test _ =
...
+let%expect_test "drop prefix" =
+  Path.drop_prefix ~prefix:(r "foo/bar") (r "foo/bar/baz")
+  |> Dyn.option Path.Local.to_dyn
+  |> print_dyn;
+  [%expect {| Some "baz" |}]
+;;
+
+let%expect_test "drop external prefix" =
+  Path.External.drop_prefix
+    ~prefix:(e "foo/bar" |> Path.as_external_exn)
```
rgrinberg, line 473: How about just creating the external paths directly instead of relying on converting them with this function?

rgrinberg (changes requested):

otherlibs/stdune/test/path_tests.ml
Alizter: Done.

PR #8962, opened 2023-10-18
feature(pkg): thread monad through 0install solver

rgrinberg:
This is a rather dumb translation that does not try to take advantage of any concurrency. It should be enough for our needs however.


Leonidas-from-XIV (approved):
I assume this is this the same code that you pushed in #8950? That seems to be working fine and I looked through it as part of that PR.

Alizter:
Does this allow for eager error messages during solving?

rgrinberg:
What are eager error messages?

rgrinberg:
> I assume this is this the same code that you pushed in #8950? That seems to be working fine and I looked through it as part of that PR.

Yes, it's the exact same stuff but cleaned up and isolated.

PR #8966, opened 2023-10-18
fix: failing coq --config

### LINKED ISSUE #8958: [Coq] `(stdlib no)` broken since 3.8.0
Starting with Dune 3.8.0, due to https://github.com/ocaml/dune/pull/6563, dune can no longer build in `(stdlib no)` mode when the stdlib is not installed. This is because it is checking `coqc --config`, which gives an exit code of 1. It has always done that, but Dune didn't seem to check it before 3.8.0

## Reproduction

We can reproduce the problem with Dune's own tests.

1. Make sure that you have `coq-core.8.18.0` installed, **but not `coq-stdlib`!**
2. `DUNE_COQ_TEST=enable ./dune.exe build @no-stdlib` outputs:

```
File "test/blackbox-tests/test-cases/coq/no-stdlib.t/run.t", line 1, characters 0-0:
------ test/blackbox-tests/test-cases/coq/no-stdlib.t/run.t
++++++ test/blackbox-tests/test-cases/coq/no-stdlib.t/run.t.corrected
File "test/blackbox-tests/test-cases/coq/no-stdlib.t/run.t", line 7, characters 0-1:
 |Test that when `(stdlib no)` is provided, the standard library is not bound to `Coq`
 |and the prelude is not imported
 |
 |  $ dune build --display=short foo.vo
 |  Warning: Coq Language Versions lower than 0.8 have been deprecated in Dune
 |  3.8 and will be removed in an upcoming Dune version.
-|        coqdep .basic.theory.d
-|          coqc foo.{glob,vo}
+|  cannot guess a path for Coq libraries; please use -coqlib option or ensure you have installed the package containing Coq's stdlib (coq-stdlib in OPAM) If you intend to use Coq without a standard library, the -boot -noinit options must be used.
+|  cannot guess a path for Coq libraries; please use -coqlib option or ensure you have installed the package containing Coq's stdlib (coq-stdlib in OPAM) If you intend to use Coq without a standard library, the -boot -noinit options must be used.
+|  [1]
 |
 |  $ dune build --display=short bar.vo
 |  Warning: Coq Language Versions lower than 0.8 have been deprecated in Dune
 |  3.8 and will be removed in an upcoming Dune version.
-|          coqc bar.{glob,vo} (exit 1)
-|  File "./bar.v", line 1, characters 20-23:
-|  Error: The reference nat was not found in the current environment.
-|  
+|  cannot guess a path for Coq libraries; please use -coqlib option or ensure you have installed the package containing Coq's stdlib (coq-stdlib in OPAM) If you intend to use Coq without a standard library, the -boot -noinit options must be used.
+|  cannot guess a path for Coq libraries; please use -coqlib option or ensure you have installed the package containing Coq's stdlib (coq-stdlib in OPAM) If you intend to use Coq without a standard library, the -boot -noinit options must be used.
```
3. When you do `git checkout 3.7.1` and run the tests again, you will see that they succeed.

CC @ejgallego @Alizter

Alizter:
The first commit adds a test detailing how Dune reacts to `coqc --version` and `coqc --config` failing. We write down what we expect to happen in each situation.

The second commit does a few things:
- Cleans up how `coq_config` is handled, shares more code, streamlines the API (removing `make_exn`).
- We start passing `-boot` to `coqc --print-version` to make it work when only `coq-core` is around.
- We make sure that the failure of `coqc --config` does not inhibit the build. There is a check for the config in the installed theories composition for example. We emit a warning if config fails and do a best effort without any installed theories.
- We improve error messages elsewhere where config or version are requested but could fail.

I don't think this needs a version bump as it affects a very small minority of users who build with only `coq-core`.

- [x] changelog
- Fixes #8958

### DIFF
--- a/doc/changes/coq_config_fix.md
+++ b/doc/changes/coq_config_fix.md
@@ -1,0 +1,6 @@
+- [coq rules] Be more tolerant when coqc --print-version / --config
+  don't work properly, and fallback to a reasonable default. This
+  fixes problems when building Coq projects with `(stdlib no)` and
+  likely other cases. (#8966, fix #8958, @Alizter, reported by Lasse
+  Blaauwbroek)
+
--- a/src/dune_rules/coq/coq_config.ml
+++ b/src/dune_rules/coq/coq_config.ml
@@ -67,6 +67,24 @@ end = struct
   ;;
 end
 
+let get_output_from_config_or_version ~(coqc : Action.Prog.t) ~what memo =
+  match coqc with
+  | Ok coqc_path ->
+    let open Memo.O in
+    let+ output, exit_code = Memo.exec memo coqc_path in
+    let sbin = Path.to_string coqc_path in
+    if exit_code <> 0
+    then
+      Error
+        (Pp.vbox
+         @@ Pp.concat_map
+              ~sep:Pp.space
+              ~f:(fun x -> Pp.hbox (Pp.text x))
+              (sprintf "%s %s failed with exit code %d." sbin what exit_code :: output))
+    else Ok output
+  | Error e -> Action.Prog.Not_found.raise e
+;;
+
 module Version = struct
   module Num = struct
     type t =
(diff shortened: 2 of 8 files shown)


ejgallego:
Thanks a lot Ali! I will review tomorrow (if you need a review)

Alizter:
@ejgallego A look over would be nice. I think just reading what I've written in the test and whether that makes sense would be enough.

LasseBlaauwbroek:
Thanks @Alizter !

ejgallego (approved):
LGTM, thanks Ali! I let you merge.

I wonder if maybe a general mechanism to query this kind of programs, and handle errors in the same way would be of interest to Dune?

We could have helpers to implement some common operations, for example, to provide default values.

Alizter:
@ejgallego If you have a look, its easy to see that coq_config is inspired by the handling in OCaml. I don't see much utility in sharing code here as each has very specific needs. `ocamlc --config` is also much stabler compared to Coq. It will probably end up being more of a maintenance burden to share code here, especially if in the far future Coq support becomes a plugin.

Alizter:
Before we merge, I need to push a changelog entry.

ejgallego:
> Before we merge, I need to push a changelog entry.

Indeed, sorry I forgot about this!

Regarding the handling of the commands, I was thinking more about making an extension of `Memo.exec` that is a bit more convenient to use w.r.t. error handling.

But indeed not a lot to gain right now I think? It seems dune has about 50 uses of it, so I'm unsure.

ejgallego:
@Alizter what's blocking this? I'd suggest amending a bit the changes entry, otherwise I suggest we merge.

Alizter:
@ejgallego I can't amend the changelog just yet. So if you want to do it feel free.

ejgallego:
@Alizter sounds good, I will amend and merge if I get a minute, otherwise you feel free to go ahead.

PR #8972, opened 2023-10-19
fix(engine): remove context from action digest

rgrinberg:
This PR removes the context name from the rules and actions. The context can always be derived from the action directory so it's just redundant.

### DIFF
--- a/src/dune_engine/build_system.ml
+++ b/src/dune_engine/build_system.ml
@@ -235,7 +235,7 @@ end = struct
 
   (* The current version of the rule digest scheme. We should increment it when
      making any changes to the scheme, to avoid collisions. *)
-  let rule_digest_version = 16
+  let rule_digest_version = 17
 
   let compute_rule_digest
     (rule : Rule.t)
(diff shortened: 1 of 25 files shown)


pmwhite (approved):

src/dune_engine/build_system.ml
```diff
@@ -393,7 +392,7 @@ end = struct
           let* action_exec_result =
             let input =
               { Action_exec.root
-              ; context
+              ; context (* can be derived from the root *)
```
snowleopard, line 395: I'm a bit confused by this comment. Why "can be" if we've already derived it, and why from "root" if we derived it from the `dir`?

snowleopard (approved):
Awesome, thanks! Just left a little comment. 

Also, the PR title is a little misleading: the actual change is more than just removing the context from action digest.

snowleopard:
@pmwhite Do you mind importing this change? Seems nice to lose the `context` field internally too.

src/dune_engine/build_system.ml
rgrinberg: "root" refers to the field inside `Action_exec.input`

rgrinberg: I removed the comment though

snowleopard: Ah, I see. Thanks!

Issue #8730, opened 2023-09-22
Solver variables in package lockfiles


gridbugs:
Some opam variables are both used to determine a package's dependencies during solving and also used to determine which commands to run while building and installing the package. To make sure both the solving step and building/installing step see the same variables the solver variables used to generate a lockdir should also be used during the build/installation. I think it would be appropriate to store the value of solver valiables in the `lock.dune` file and then read them from that file to make the variables available during build/installation.

rgrinberg:
This only applies to a subset of variables. Otherwise how would it work for `%{os_family}` for example?

Could you list the variables you intend for this to work?

gridbugs:
I had the flags defined in `Solver_env.Variable.Flag` (`with-test` and `with-doc`) in mind when I wrote this but you raise an interesting point about `os-family` and other system variables. If a lockfile is generated with `os = linux` for example, what should the behaviour be when a build is attempted on a system where `os = macos`?

gridbugs:
Thinking about this more and we only want `with-test` and `with-doc` to be set for packages which are defined in the current project. So I think they should be set in individual package lockfiles rather than `lock.dune` as they'll have different values for different packages.

gridbugs:
But for system variables like `os-family` we might want to record the variables that were set when the solve was done so on systems that are incompatible with those variables we can detect it and possibly print a warning or something. For system variables it makes sense to record them in `lock.dune` as they'll be the same for all packages.

rgrinberg:
>  If a lockfile is generated with os = linux for example, what should the behaviour be when a build is attempted on a system where os = macos?

I would imagine the build instructions to be evaluated with `os = macos` and the build having a reasonable chance of succeeding. Especially if `%{os}` isn't used anywhere.

> Thinking about this more and we only want with-test and with-doc to be set for packages which are defined in the current project

Yeah, that seems like a reasonable first approximation. However, the following somewhat niche use cases also exist:
1. Building docs for dependencies
2. Verifying that the dependencies are working by running their test

> But for system variables like os-family we might want to record the variables that were set when the solve was done so on systems that are incompatible with those variables we can detect it and possibly print a warning or something. For system variables it makes sense to record them in lock.dune as they'll be the same for all packages.

I think that makes sense

gridbugs:
@rgrinberg and I discussed this and the plan is to abstract the `with-test` and `with-doc` variables away from users so we won't store these in lockfiles. I'll just add the system variables for now.

[...]

PR #8973, opened 2023-10-20
pkg: store solver variables in lockdir metadata

gridbugs:
Adds a "solver_vars" field to lockdir metadata which holds the values of variables which were evaluated while solving dependencies.

### DIFF
--- a/src/dune_pkg/dune_pkg.ml
+++ b/src/dune_pkg/dune_pkg.ml
@@ -7,6 +7,7 @@ module Opam_solver = Opam_solver
 module Package_variable = Package_variable
 module Repository_id = Repository_id
 module Solver_env = Solver_env
+module Solver_stats = Solver_stats
 module Substs = Substs
 module Sys_poll = Sys_poll
 module Version_preference = Version_preference
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -311,6 +311,7 @@ type t =
   ; packages : Pkg.t Package_name.Map.t
   ; ocaml : (Loc.t * Package_name.t) option
   ; repos : Repositories.t
+  ; expanded_solver_variable_bindings : Solver_stats.Expanded_variable_bindings.t
   }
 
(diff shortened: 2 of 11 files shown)


gridbugs:
@rgrinberg we've discussed in the past the possibility of evaluating variables in commands and command filters at solve-time rather than build-time (at least for now). If/when we start doing that any variables that are expanded should be added to the solver_vars stored in the lockdir metadata.

gridbugs:
fixes https://github.com/ocaml/dune/issues/8730

gridbugs:
@rgrinberg can you take a look at this please

rgrinberg (approved):
Not a fan of having fields such as "string_variables". Why can't this just be "variables"? And why are we limiting this to strings?

I think what we're missing is something similar to OSL but for maps. The following is a short sketch:

1. `:standard` refers to the default set of bindings
2. `(:standard (foo bar))` adds the variable `foo` with value `bar`.
3. `(:standard \ baz)` removes the binding with the key `baz`.

Perhaps reusing the syntax of OSL is a mistake, but I think this is still conceptually cleaner than `unset_variables` and `string_variables`.

gridbugs:
I'm happy to rename `string_variables` to `variables`. It's limited to strings because those are the only types of variables that we currently support. When we add flags we'll need a boolean type, though we may want to just stick to the current convention in dune and opam of using sentinel string values "true" and "false" to represent bools.

I don't see how using "OSL for maps" would let us remove the unset variables field. That field contains the set of variables that were read by the solver but which have no value. I suppose we could add them to the map with a special value "undefined" but adding an "undefined" value feels like a can of worms to me.

gridbugs:
Discussed this with @rgrinberg on slack. We'll change the syntax of how variables are encoded in the lockdir to:
```
 (os linux)
 (os-distribution :unset)
 (with-test true)
```
...rather than having a separate field for unset variables.

I'll make that change in a future PR since another feature I'm working on depends on encoding the solver variables in the lockdir.

PR #8974, opened 2023-10-20
Show dune cache location info in log

nojb:
I often find myself struggling to find where the cache is located (especially on Windows). What do you think about printing its location in the log?

image


Alizter:
I was also thinking about maybe adding the location to `dune cache size` so it says something like:
```
Dune cache at /path/to/cache
12.3 GB
```

Alizter (approved):
This patch seems like a sensible addition. I don't know if there are any concerns with sharing sensitive directories in the log when users paste their log during bug reports however.

nojb:
> @Alizter I share your concern, however at least in this screenshot the users home directory and thus username is already revealed via `INSIDE_DUNE` (same for me, I just checked).

The home directory and the path to the compiler/OPAM/etc are all printed in the log already.

bin/common.ml
```diff
@@ -1249,6 +1249,11 @@ let init (builder : Builder.t) =
         }
   in
   Log.info [ Pp.textf "Shared cache: %s" (Config.Toggle.to_string config.cache_enabled) ];
+  Log.info
```
rgrinberg, line 1252: Perhaps you should print only if the cache is enabled?

nojb: We could; but is it worth it? Also, the location of the cache is still valid, even if the cache is disabled _right now_.

rgrinberg (approved):

bin/common.ml
rgrinberg: Sure, I don't think it matters that much. Was just wondering if showing the path to the cache might confuse people that the cache is enabled if they aren't reading carefully.

PR #8975, opened 2023-10-20
Add `dune cache clear` command

### LINKED ISSUE #8971: feature: clear dune cache
It is pretty common to use `dune cache trim` to completely remove the cache. The right invocation is `dune cache trim --size 0B` but it requires finding about `--size` vs `--trimmed-size`, adding a `B` suffix, etc.

I suggest adding a dedicated way to do that. Either `dune cache trim --all`, or `dune cache clear`, to be determined.

nojb:
Fixes #8971

Of course, if someone does `DUNE_CACHE_ROOT=/ dune cache clear` or similar they will zap their whole system...

### DIFF
--- a/bin/cache.ml
+++ b/bin/cache.ml
@@ -80,16 +80,25 @@ let size =
     Cmd.info "size" ~doc ~man
   in
   Cmd.v info
-  @@ let+ machine_readble =
+  @@ let+ machine_readable =
        Arg.(
          value
          & flag
          & info [ "machine-readable" ] ~doc:"Outputs size as a plain number of bytes.")
      in
      let size = Dune_cache.Trimmer.overhead_size () in
-     if machine_readble
+     if machine_readable
      then User_message.print (User_message.make [ Pp.textf "%Ld" size ])
      else User_message.print (User_message.make [ Pp.textf "%s" (Bytes_unit.pp size) ])
+;;
+
+let clear =
+  let info =
+    let doc = "Clear the Dune cache." in
+    let man = [ `P "Remove any traces of the Dune cache." ] in
+    Cmd.info "clear" ~doc ~man
+  in
+  Cmd.v info @@ Term.(const Dune_cache_storage.clear $ const ())
 ;;
 
 let command =
@@ -104,5 +113,5 @@ let command =
     in
     Cmd.info "cache" ~doc ~man
   in
-  Cmd.group info [ trim; size ]
+  Cmd.group info [ trim; size; clear ]
 ;;
--- a/doc/changes/8975.md
+++ b/doc/changes/8975.md
@@ -1,0 +1,2 @@
+- Add command `dune cache clear` to completely delete all traces of the Dune
+  cache. (#8975, @nojb)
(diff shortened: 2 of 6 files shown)


Alizter:
Maybe we should prompt the user saying:
```
This will delete ALL contents directory /foo/bar, are you sure?
```
Saves anybody accidentally zapping something.

Alizter:
We could also perform a simple check to make sure it is Dune cache we are deleting and not some other files. I'm a bit worried about giving users a blanketed way to shoot themselves.

rgrinberg:
What's the difference between clearing the cache and doing dune clean? I don't see a need for a prompt for either. 

A nice improvement would be to see how many files were deleted, but it's not required for this PR.

@pmwhite could you review this one?

emillon:
I think `Path.rm_rf ~external:true` is a bit too unsafe for my taste. Is it possible to iterate on `Dune_cache_storage.Layout.Versioned.list_*_entries` instead?

nojb:
> Is it possible to iterate on `Dune_cache_storage.Layout.Versioned.list_*_entries` instead?

We could. An alternative could be to change the interpretation of `DUNE_CACHE_ROOT` as follows. Currently, `DUNE_CACHE_ROOT` is (by default) `~/.cache/dune/db` and the cache creates a number of directories inside this directory. Instead we could consider that `DUNE_CACHE_ROOT` is `~/.cache/dune` by default and the Dune cache will create all its directories inside `$DUNE_CACHE_ROOT/db`. The `dune cache clear` command could then just zap the `$DUNE_CACHE_ROOT/db` directory which would be safer than deleting `$DUNE_CACHE_ROOT`.

Opinions?

pmwhite:
@nojb The most significant part of that alternative to me is that we require `DUNE_CACHE_ROOT` to contain a `db` sub-directory, thus providing a little bit of validation that the variable actually does point to a cache. Whether we delete the cache directory itself or `db` is an orthogonal decision, I think; that is, we could check that there exists a `db` sub-directory before deleting the whole cache. Are there other bits of structure that we could check before deleting anything?

Separately, a few bikeshedding thoughts: `dune cache clear` does sound to me like we're removing the contents of the cache, rather than deleting the cache itself. So I would support either changing the behavior to delete just the contents, or renaming the command to `dune cache delete`. Maybe `dune delete-cache` is clearer because `dune cache delete` might seem like it lets you delete individual cache entries, but it's also a bit less discoverable. I don't have strong feelings about any of these naming things, to be clear; do what you think is best.

Alizter:
On the topic of naming, `dune clean` and `dune cache clear` sounds very similar and it might confuse future users trying to remember `clean` vs `clear`. I don't have a good solution for this, so you can disregard this comment if you wish.

nojb:
> I find `Path.rm_rf ~external:true` a bit too unsafe. Is it possible to iterate on `Dune_cache_storage.Layout.Versioned.list_*_entries` instead?

I pushed a commit with this approach; I agree that it is an improvement.

pmwhite (approved):

nojb:
Sorry, this one slipped through the cracks :) I rebased it, added a test and an entry to the changelog.

It's been already approved by @pmwhite, so am planning to merge it once the CI passes.

emillon:
Thanks!

PR #8976, opened 2023-10-20
melange: show test depending on alias

jchavarri:
Demonstrates issue #8977, when trying to depend on the alias of a `melange.emit` stanza.


rgrinberg:
I think we have tests for this that don't depend on melange. Btw, thanks or finding a bug here.

PR #8978, opened 2023-10-20
Reduce allocations while loading rules

pmwhite:
This PR includes a few miscellaneous changes that should reduce the amount of allocation done while loading rules. I attempted to have reasonable separation of changes by commit, so you might find it helpful to read the changes one commit at a time, in addition to viewing all the changes at once.


src/dune_engine/load_rules.ml
```diff
@@ -715,20 +715,13 @@ end = struct
...
-      ; copy_rules = []
-      }
-    ;;
+    let empty = { source_files = Path.Source.Set.empty; source_dirs = Filename.Set.empty }
   end
 
   (* Compute all the copying rules from the source directory. *)
-  let rules_from_source_dir source_paths_to_ignore (context_name : Context_name.t) sub_dir
-    =
+  let rules_from_source_dir source_paths_to_ignore sub_dir =
```
Alizter, line 729: This doesn't seem to make sense any longer, same with the comment above.

src/dune_engine/load_rules.ml
```diff
@@ -402,10 +402,12 @@ end = struct
       Appendable_list.to_list expansions)
   ;;
 
-  let filter_out_fallback_rules ~source_files rules =
-    List.filter rules ~f:(fun (rule : Rule.t) ->
+  let disjoint set1 set2 = not (Path.Source.Set.exists set1 ~f:(Path.Source.Set.mem set2))
```
rgrinberg, line 405: Since this function is generic, it could be moved `Stdune.Set`.

src/dune_engine/load_rules.ml
```diff
@@ -688,47 +689,44 @@ end = struct
...
-      ; copy_rules = []
-      }
-    ;;
+    let empty = { source_files = Path.Source.Set.empty; source_dirs = Filename.Set.empty }
   end
 
   (* Compute all the copying rules from the source directory. *)
-  let rules_from_source_dir source_paths_to_ignore (context_name : Context_name.t) sub_dir
-    =
+  let rules_from_source_dir source_paths_to_ignore sub_dir =
```
rgrinberg, line 729: a name like `source_files_and_dirs` is more appropriate now

src/dune_engine/load_rules.ml
```diff
@@ -688,47 +689,44 @@ end = struct
...
+           in
+           iter
+             ~files:(Path.Build.Set.union files target_files)
+             ~dirnames:(Filename.Set.union dirnames (Lazy.force target_dirnames))
+             rules)
+    in
+    iter ~files:Path.Build.Set.empty ~dirnames:Filename.Set.empty rules
   ;;
 
   module Source_rules = struct
```
rgrinberg, line 719: This module name could be improved

rgrinberg (approved):

rgrinberg:
@pmwhite could you add a DCO to every commit? Or squash and then DCO the one commit if that's easier?

src/dune_engine/load_rules.ml
pmwhite: Yeah, might as well do that. We could go even further and unify this type with the `source_paths_to_ignore` type, since they are roughly the same. But I don't think that change belongs in the PR.

pmwhite:
Ahh, thanks, I forgot about signing off on commits. They should each be marked correctly now.

PR #8980, opened 2023-10-20
fix: add break hints to Dyn.pp

Alizter:
When working on another feature I noticed that Dyn.pp behaved very poorly when pretty printing variant constructor arguments. Turns out a Pp.char ',' separator was being used however there was no break hint. This patch adds a Pp.space afterwards which seems to make the output much more pleasant.

### DIFF
--- a/otherlibs/dyn/dyn.ml
+++ b/otherlibs/dyn/dyn.ml
@@ -112,7 +112,11 @@ let rec pp =
   | Variant (v, xs) ->
     Pp.hvbox
       ~indent:2
-      (Pp.concat [ Pp.verbatim v; Pp.space; Pp.concat_map ~sep:(Pp.char ',') xs ~f:pp ])
+      (Pp.concat
+         [ Pp.verbatim v
+         ; Pp.space
+         ; Pp.concat_map ~sep:(Pp.seq (Pp.char ',') Pp.space) xs ~f:pp
+         ])
 ;;
 
 let to_string t = Format.asprintf "%a" Pp.to_fmt (pp t)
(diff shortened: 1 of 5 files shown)


otherlibs/dyn/dyn.ml
```diff
@@ -112,7 +112,11 @@ let rec pp =
   | Variant (v, xs) ->
     Pp.hvbox
       ~indent:2
-      (Pp.concat [ Pp.verbatim v; Pp.space; Pp.concat_map ~sep:(Pp.char ',') xs ~f:pp ])
+      (Pp.concat
+         [ Pp.verbatim v
+         ; Pp.space
+         ; Pp.concat_map ~sep:(Pp.seq (Pp.char ',') Pp.space) xs ~f:pp
```
rgrinberg, line 118: Shouldn't it be `Pp.seq Pp.space (Pp.char ',')`? We use a leading `,` and `;` everywhere in our code.

Alizter: That works too, I was being consistent with what was being done for Tuple.

rgrinberg: Okay, then leave it. We can change the style in subsequent PR's.

Alizter: We have a choice here.
```
Pp.custom_break ~fits:(",", 1, "") ~breaks:("", 0, ", ")
```
Gives:
```
a, b, c
```
when fitting on a line and
```
a
, b
, c
```
when breaking.

We can reduce the space in ~fits to 0 to preserve the tight constructor tuples we already had but I think it looks clearer with a space.

I can also make the same break modification for tuples.

Alizter: I'll leave it for later however.

PR #8984, opened 2023-10-23
test: use a cached file in a directory target

emillon:
We cache a file in a file target and use it in a directory target.

### DIFF
--- a/test/blackbox-tests/test-cases/directory-targets/cache-file-and-dir.t
+++ b/test/blackbox-tests/test-cases/directory-targets/cache-file-and-dir.t
@@ -2,6 +2,7 @@ This checks what happens when a file available in the cache is used in a directo
 
   $ export DUNE_CACHE_ROOT=$PWD/.cache
   $ export DUNE_CACHE=enabled
+  $ . ./helpers.sh
 
   $ cat > dune-project << EOF
   > (lang dune 3.11)
@@ -32,11 +33,6 @@ This checks what happens when a file available in the cache is used in a directo
 
 We will check whether an entry is linked from the cache. This corresponds to a
 file with more than one link.
-
-  $ is_linked() {
-  >   nlinks=$(dune_cmd stat hardlinks $1)
-  >   [ $nlinks -gt 1 ] && echo linked || echo not linked
-  > }
 
 We prime the cache with the file:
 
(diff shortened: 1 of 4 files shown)


test/blackbox-tests/test-cases/directory-targets/cache-file-and-dir.t
```diff
@@ -0,0 +1,56 @@
...
+  >  (target (dir dir_out))
+  >  (deps gen.exe)
+  >  (action
+  >   (no-infer
+  >    (progn
+  >     (run mkdir dir_out)
+  >     (with-stdout-to dir_out/a (run ./gen.exe))
+  >     (write-file dir_out/b contents_b)))))
+  > 
+  > (executable
```
rgrinberg, line 25: Instead of building an executable, you could:

1. execute it with `ocaml gen.ml`
2. use a shell script (we already depend on a shell script elsewhere in this test anyway)

I feel it's a bit necessary to make use of dune's OCaml rules for this test.

test/blackbox-tests/test-cases/directory-targets/cache-file-and-dir.t
```diff
@@ -0,0 +1,56 @@
...
+
+  $ cat > gen.ml << EOF
+  > let () = prerr_endline "running command"
+  > let () = print_endline "contents"
+  > EOF
+
+We will check whether an entry is linked from the cache. This corresponds to a
+file with more than one link.
+
+  $ is_linked() {
```
rgrinberg, line 37: I've noticed there's a similar function in the other test.

It would be nice to reuse this function in both tests. See `helpers.sh` for some examples.

test/blackbox-tests/test-cases/directory-targets/cache-file-and-dir.t
```diff
@@ -0,0 +1,56 @@
...
+
+We prime the cache with the file:
+
+  $ dune build file_out
+  running command
+  $ is_linked _build/default/file_out
+  linked
+
+Then use it in the directory target. We expect the command to run (because we
+can not know in advance that it is going to generate that fail); but to link
```
rgrinberg, line 50: that "file"

rgrinberg (approved):

PR #8986, opened 2023-10-23
test: directory targets sharing a subdirectory

emillon:


test/blackbox-tests/test-cases/directory-targets/cache-shared-subdir.t
```diff
@@ -0,0 +1,54 @@
...
+  > (lang dune 3.11)
+  > (using directory-targets 0.1)
+  > EOF
+
+  $ cat > dune << EOF
+  > (rule
+  >  (target (dir d1))
+  >  (deps gen.sh)
+  >  (action
+  >   (no-infer
```
rgrinberg, line 16: It might be clearer to move `no-infer` around the `write-file`. As far as I can tell, you just don't want the d1/x target from being inferred.

Then, you can remove the `deps` field

test/blackbox-tests/test-cases/directory-targets/cache-shared-subdir.t
```diff
@@ -0,0 +1,54 @@
+We create 2 directory targets which share a whole subdirectory.
```
rgrinberg, line 1: Another sentence to describe the property we're observing would be useful.

rgrinberg (approved):

PR #8990, opened 2023-10-24
fix: include expand aliases in digest

rgrinberg:
Action digests should depend on whether the aliases have been expanded in the sandbox.

Technically, this value should only matter if the action is sandboxed and there are alias dependencies, but it's likely not worth the effort to make the digest this precise.

cc @jchavarri who discovered this issue

### DIFF
--- a/test/blackbox-tests/test-cases/depend-on/expand-aliases-rerun.t
+++ b/test/blackbox-tests/test-cases/depend-on/expand-aliases-rerun.t
@@ -25,3 +25,8 @@ Now we set (expand_aliases_in_sandbox), and re-run the action.
   > (expand_aliases_in_sandbox)
   > EOF
   $ dune build @foo
+  .
+  ./alias-dep
+
+The above output should include the files that we depend on via the alias
+expansion.
--- a/doc/changes/8990.md
+++ b/doc/changes/8990.md
@@ -1,0 +1,2 @@
+- Re-run actions whenever `(expand_aliases_in_sandbox)` changes (#8990,
+  @rgrinberg)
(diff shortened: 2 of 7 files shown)


test/blackbox-tests/test-cases/depend-on/expand-aliases-rerun.t
```diff
@@ -25,3 +25,5 @@ Now we set (expand_aliases_in_sandbox), and re-run the action.
   > (expand_aliases_in_sandbox)
   > EOF
   $ dune build @foo
+  .
+  ./alias-dep
```
rgrinberg, line 29: This test demonstrates the issue being fixed

snowleopard: I don't follow. Could you make a separate PR where this test fails, so it's easy to see the problem and that this PR actually fixes it?

snowleopard:
The change seems OK but I don't understand the problem that is being solved. Is there an GitHub issue with an exaplanation?

test/blackbox-tests/test-cases/depend-on/expand-aliases-rerun.t
rgrinberg: This test demonstrates the issue https://github.com/ocaml/dune/pull/8991

snowleopard: Thanks, could you rebase this PR to pick it up?

rgrinberg: I've rebased and added a sentence that explains the output. Let me know if it's clear now.

snowleopard: That helps, thanks!

snowleopard (approved):
Looks good!

PR #8992, opened 2023-10-24
melange: remove ppx flag from merlin

jchavarri:
In https://github.com/melange-re/melange/pull/583 Melange moved its preprocessor outside of the compiler so that it can be invoked like any other ppx.

This PR adapts merlin config to not set this flag any longer.

Regarding backwards compatibility, the plan is to wait until a newer Dune version with this patch is released, and once it's published, add an upper bound on the `dune` dep in the `melange.1.0.0` opam so that it never uses that version.

### DIFF
--- a/test/blackbox-tests/test-cases/melange/merlin.t
+++ b/test/blackbox-tests/test-cases/melange/merlin.t
@@ -43,15 +43,11 @@ Paths to Melange stdlib appear in B and S entries without melange.emit stanza
   S /MELC_STDLIB
   S $TESTCASE_ROOT
 
-All 3 modules (Foo, Foo__ and Bar) contain a ppx directive
-
-  $ dune ocaml merlin dump-config $PWD | grep -i "ppx"
-  [1]
-
   $ target=output
   $ cat >dune <<EOF
   > (melange.emit
   >  (target "$target")
+  >  (compile_flags :standard -bs-D DEBUG=true )
   >  (modules main))
   > EOF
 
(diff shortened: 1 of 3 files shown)


test/blackbox-tests/test-cases/melange/merlin.t
```diff
@@ -46,9 +46,7 @@ Paths to Melange stdlib appear in B and S entries without melange.emit stanza
 All 3 modules (Foo, Foo__ and Bar) contain a ppx directive
 
   $ dune ocaml merlin dump-config $PWD | grep -i "ppx"
```
jchavarri, line 48: Not sure what to do with this test and the one below. They don't make much sense anymore so we can prob remove them?

rgrinberg: Seems like it. Do we have tests for regular ppx + melange?

jchavarri: I cleaned up the test file in https://github.com/ocaml/dune/pull/8992/commits/53c8a93d9c29c1eee2a6d6a31090a38255dde73a.

> Do we have tests for regular ppx + melange?

Yes, see https://github.com/ocaml/dune/blob/53c8a93d9c29c1eee2a6d6a31090a38255dde73a/test/blackbox-tests/test-cases/melange/merlin.t#L105-L165

anmonteiro (approved):

PR #8997, opened 2023-10-25
Simplify Cache_lookup type in Memo

snowleopard:
I think the code is much simpler than before, plus we get rid of some unnecessary indirections in the `Failure` case.

Also removed two stale comments about variables, and a duplicated comment on `Value`.

Benchmarks show small (<1%) improvements across most build scenarios.

----

This has been reviewed and used internally for weeks with no issues.

### DIFF
--- a/src/memo/memo.ml
+++ b/src/memo/memo.ml
@@ -231,8 +231,7 @@ module M = struct
     type 'a t =
       | Cached_value of 'a Cached_value.t
       | Out_of_date of { old_value : 'a Cached_value.t Option.Unboxed.t }
-      | Restoring of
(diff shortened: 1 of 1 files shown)


rgrinberg (approved):

Alizter:
This is a 5% improvement for clean build time!

snowleopard:
> This is a 5% improvement for clean build time!

Cool! Where does this number come from? The monorepo benchmark seems to be still running.

Alizter:
Here is the bench job: https://autumn.ocamllabs.io/ocaml/dune/pull/8997/base/main?worker=autumn&image=bench.Dockerfile

It's possible that the jobs have finished and the GitHub statuses just haven't updated.

PR #9002, opened 2023-10-26
feature: run some rule operations in the background

rgrinberg:
* cram tests execution

### DIFF
--- a/src/dune_config/config.mli
+++ b/src/dune_config/config.mli
@@ -54,6 +54,9 @@ val threaded_console : Toggle.t t
 (** The number of frames per second for the threaded console. *)
 val threaded_console_frames_per_second : [ `Default | `Custom of int ] t
 
+(** Controls whether we use background threads in the dune rules *)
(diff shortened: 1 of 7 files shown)


src/dune_config/config.mli
```diff
@@ -54,6 +54,9 @@ val threaded_console : Toggle.t t
 (** The number of frames per second for the threaded console. *)
 val threaded_console_frames_per_second : [ `Default | `Custom of int ] t
 
+(** Controls whether we use background threads in the dune rules *)
+val background_dune_rules : Toggle.t t
```
Alizter, line 58: Can you move this next to the other `background_*` options?

Alizter (approved):

Alizter:
It appears there is some noticeable overhead in the build times however.

I wonder if it makes any difference when running cram tests, but we don't bench that.

rgrinberg:
For now I've toggled this off. It will come in useful once we have faster Async IO primitives

PR #9003, opened 2023-10-26
fix: do not use non-memoized io rules

rgrinberg:
without using Fs_memo, dune will not work correctly in watch mode


emillon (approved):
Thanks, I assumed it was a performance issue but I didn't think about watch mode semantics.

PR #9009, opened 2023-10-27
fix: allow reuse of dispatch queues between fsevents in macOS

### LINKED ISSUE #8910: fsevents expect tests fail on macos
## Expected Behavior

The fsevents expect tests defined in test/expect-tests/fsevents/ should all pass. Running `dune test test/expect-tests/fsevents` should produce no output.

## Actual Behavior

The output of `dune test test/expect-tests/fsevents` is:
```diff
File "test/expect-tests/fsevents/fsevents_tests.ml", line 1, characters 0-0:
diff --git a/_build/default/test/expect-tests/fsevents/fsevents_tests.ml b/_build/.sandbox/fdbd0c5af99815b26eb99e93065ce1e3/default/test/expect-tests/fsevents/fsevents_tests.ml.corrected
index bda3e030c..4520aef04 100644
--- a/_build/default/test/expect-tests/fsevents/fsevents_tests.ml
+++ b/_build/.sandbox/fdbd0c5af99815b26eb99e93065ce1e3/default/test/expect-tests/fsevents/fsevents_tests.ml.corrected
@@ -232,7 +232,7 @@ let%expect_test "file create event" =
   test_with_operations (fun () -> Io.String_path.write_file "./file" "foobar");
   [%expect
     {|
-    > { action = "Unknown"; kind = "File"; path = "$TESTCASE_ROOT/file" } |}]
+    > { action = "Create"; kind = "File"; path = "$TESTCASE_ROOT/file" } |}]
 ;;

 let%expect_test "dir create event" =
@@ -248,8 +248,8 @@ let%expect_test "move file" =
     Unix.rename "old" "new");
   [%expect
     {|
-    > { action = "Unknown"; kind = "File"; path = "$TESTCASE_ROOT/old" }
-    > { action = "Unknown"; kind = "File"; path = "$TESTCASE_ROOT/new" } |}]
+    > { action = "Create"; kind = "File"; path = "$TESTCASE_ROOT/old" }
+    > { action = "Rename"; kind = "File"; path = "$TESTCASE_ROOT/new" } |}]
 ;;

 let%expect_test "raise inside callback" =
@@ -278,14 +278,13 @@ let%expect_test "set exclusion paths" =
   run Filename.concat;
   [%expect
     {|
-    > { action = "Create"; kind = "Dir"; path = "$TESTCASE_ROOT/ignored" }
-    > { action = "Unknown"; kind = "File"; path = "$TESTCASE_ROOT/ignored/old" } |}];
+    > { action = "Create"; kind = "Dir"; path = "$TESTCASE_ROOT/ignored" } |}];
   (* but relative paths do not *)
   run (fun _ name -> name);
   [%expect
     {|
     > { action = "Create"; kind = "Dir"; path = "$TESTCASE_ROOT/ignored" }
-    > { action = "Unknown"; kind = "File"; path = "$TESTCASE_ROOT/ignored/old" } |}]
+    > { action = "Create"; kind = "File"; path = "$TESTCASE_ROOT/ignored/old" } |}]
 ;;

 let%expect_test "multiple fsevents" =
```

## To Reproduce

On macos run `dune test test/expect-tests/fsevents`. I've reproduced this on an M1 mac running Ventura, and an intel mac running Montery.

### LINKED ISSUE #9004: Watch mode crashes on macos on `main` branch
On my machine, watch mode is broken on the main branch.
@emillon mentioned it might be linked to https://github.com/ocaml/dune/pull/8304

## Specifications

- Version of `dune`: 3.11.0-221-g5b6c8f4 (main)
- Version of `ocaml`: 4.14.1 (and 5.1.0)
- Operating system (distribution and version): MacOS Sonoma (M1 pro)

## Reproduction

- Clean copy at 5b6c8f4430b134cff9b19b524ea4fc97b1c7e963
- `make dev`
- `./dune.exe build @install -w`

```
$ ./dune.exe build @install -w     
Description:
  ("Fsevents.start: dispatch queue stopped", {})
Called from Stdlib__Fun.protect in file "fun.ml", line 33, characters 8-15
Re-raised at Stdlib__Fun.protect in file "fun.ml", line 38, characters 6-52
Called from Stdlib__Fun.protect in file "fun.ml", line 33, characters 8-15
Re-raised at Stdlib__Fun.protect in file "fun.ml", line 38, characters 6-52
Called from Dune_file_watcher.add_watch in file
  "src/dune_file_watcher/dune_file_watcher.ml", line 729, characters 13-50
Called from Dune_engine__Fs_memo.Watcher.watch_path in file
  "src/dune_engine/fs_memo.ml", line 69, characters 10-50
Called from Dune_engine__Fs_memo.Watcher.memo_for_watching_via_parent.(fun)
  in file "src/dune_engine/fs_memo.ml", line 130, characters 8-58
  line 76, characters 8-11
-> required by
   ("fs_memo_for_watching_via_parent",
   External "/Library/TeX/texbin/latexmk/hg")
-> required by ("which", <opaque>)
-> required by ("which", <opaque>)
-> required by
   ("load-dir", In_build_dir "default/test/blackbox-tests/test-cases/cram")

```

kevinji:
Previously, the code waiting on `dq->dq_finished` would get notified in `dune_fsevents_dispatch_queue_wait_until_stopped` that the dispatch queue had completed running. This would lead to the dispatch queue being marked as `Stopped` on the OCaml side, even though the dispatch queue could be passed into `Fsevents.start` again.

This change now tracks the number of fsevents streams in a `fsevents_streams` counter so that `Dispatch_queue.wait_until_stopped` will only return when an exception is raised or if all streams have been stopped, restoring the original behavior in the code that used a runloop.

Fixes #9004. Fixes #8910.

### DIFF
--- a/src/fsevents/fsevents_stubs.c
+++ b/src/fsevents/fsevents_stubs.c
@@ -28,6 +28,7 @@ typedef struct dune_dispatch_queue {
   dispatch_queue_t dq;
   pthread_cond_t dq_finished;
   pthread_mutex_t dq_lock;
+  uint32_t fsevent_streams;
   value v_exn;
 } dune_dispatch_queue;
 
@@ -76,6 +77,7 @@ CAMLprim value dune_fsevents_dispatch_queue_create(value v_unit) {
   pthread_mutex_init(&dq->dq_lock, NULL);
   pthread_cond_init(&dq->dq_finished, NULL);
   dq->dq = dispatch_queue_create("build.dune.fsevents", DISPATCH_QUEUE_SERIAL);
+  dq->fsevent_streams = 0;
   dq->v_exn = Val_unit;
   caml_register_global_root(&dq->v_exn);
   v_dq = caml_alloc_custom(&dune_fsevents_dispatch_queue_ops,
@@ -91,10 +93,10 @@ CAMLprim value dune_fsevents_dispatch_queue_wait_until_stopped(value v_dq) {
   caml_release_runtime_system();
   pthread_mutex_lock(&dq->dq_lock);
   pthread_cond_wait(&dq->dq_finished, &dq->dq_lock);
+  v_exn = dq->v_exn;
   pthread_mutex_unlock(&dq->dq_lock);
   caml_acquire_runtime_system();
   caml_remove_global_root(&dq->v_exn);
-  v_exn = dq->v_exn;
   if (v_exn != Val_unit) {
     caml_raise(v_exn);
   }
@@ -187,10 +189,12 @@ static void dune_fsevents_callback(const FSEventStreamRef streamRef,
   v_res = caml_callback_exn(t->v_callback, v_events_xs);
   if (Is_exception_result(v_res)) {
     v_res = Extract_exception(v_res);
+    caml_release_runtime_system();
     pthread_mutex_lock(&dq->dq_lock);
     dq->v_exn = v_res;
     pthread_cond_broadcast(&dq->dq_finished);
     pthread_mutex_unlock(&dq->dq_lock);
+    caml_acquire_runtime_system();
   }
   CAMLdrop;
   caml_release_runtime_system();
@@ -239,6 +243,7 @@ CAMLprim value dune_fsevents_create(value v_paths, value v_latency,
       &context, paths, kFSEventStreamEventIdSinceNow, Double_val(v_latency),
       flags);
   CFRelease(paths);
+  t->dq = NULL;
   caml_register_global_root(&t->v_callback);
   t->v_callback = v_callback;
   t->stream = stream;
@@ -267,6 +272,11 @@ CAMLprim value dune_fsevents_start(value v_t, value v_dq) {
   dune_fsevents_t *t = Fsevents_val(v_t);
   dune_dispatch_queue *dq = Dispatch_queue_val(v_dq);
   t->dq = dq;
+  caml_release_runtime_system();
+  pthread_mutex_lock(&dq->dq_lock);
+  dq->fsevent_streams++;
+  pthread_mutex_unlock(&dq->dq_lock);
+  caml_acquire_runtime_system();
   FSEventStreamSetDispatchQueue(t->stream, dq->dq);
   bool res = FSEventStreamStart(t->stream);
   if (!res) {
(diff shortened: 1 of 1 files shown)


Alizter:
Why does this only trigger on M1?

kevinji:
I think this should apply to any Mac running Dune, not just an M-series chip, but I don’t have an Intel Mac to confirm this.

gridbugs:
Confirming that I can reproduce https://github.com/ocaml/dune/issues/9004 and the fix in this PR on an Intel mac running Monterey and an M1 mac running Sonoma.

gridbugs (approved):

kevinji:
I've pushed a new version that has slightly altered behavior from the original PR: now, the dispatch queue wait call will also return if all fsevents streams have been removed.

I'm running into what seems like a race condition where if the callback returns an exception which then notifies the `dune_fsevents_dispatch_queue_wait_until_stopped` function via a condition variable, the dispatch queue will sometimes (non-deterministically) still execute another item in the queue before stopping.

Ideally if an exception is seen, the queue should stop processing immediately. I tried inserting
```ocaml
  FSEventStreamStop(t->stream);
  FSEventStreamInvalidate(t->stream);
  FSEventStreamRelease(t->stream);
```
into the body of `dune_fsevents_callback` after an exception is seen, but still seem to see some race conditions, so any suggestions would be appreciated.

gridbugs:
> I'm running into what seems like a race condition...
@kevinji can you clarify whether that issue was present before your recent changes to this PR? If not, would it make sense to merge the original form of this PR for the release of 3.12 while we work out the issues with the new code?

I'm trying to get an understanding of how the new changes work but it looks like there are several changes unrelated to the description (e.g. some seem stylistic). Do you mind splitting those into a separate commit so I can be sure I'm paying attention to the right stuff.

kevinji:
@gridbugs Apologies for the late response. I've pulled out refactoring changes into #9288 and the test changes into #9289. Once those PRs are merged I can rebase this PR on top of main to clean up the diff.

This PR fixes the bugs mentioned in the initial comment as well as segfaults that I saw when running the tests in `test/expect-tests/fsevents`. Currently, the race condition I'm experiencing is in the "raise inside callback" test, where the `[EXIT]` in the output often (but not always) fails to show up.

Let me know if I can provide any other clarification.

gridbugs:
I'm happy with this. One issue is that the race condition could lead to flaky CI. I now understand why it happens. In `dune_fsevents_callback` (fsevents_stubs.c) we call the user-provided callback and then if the callback raised an exception we store the exception in `dq->v_exn` and broadcast `dq->dq_finished` which causes `wait_until_stopped` to stop waiting and either return or raise the exception in `dq->v_exn` if there's an exception stored there. The problem happens when `dune_fsevents_stop` is called while a callback is being processed. `dune_fsevents_stop` also broadcasts on `dq->dq_finished` which wakes the thread that called `wait_until_stopped`. If the callback raises but doesn't do so until after `wait_until_stopped` checks for exceptions, the exception never gets re-raised from the c code to the caller of `wait_until_stopped`.

So the question is what should happen to exceptions that are raised by callbacks after the watcher has been stopped (note that the watcher may have been stopped while the callback was running)? I think it's reasonable to continue the current behaviour of dropping such exceptions, in which case a quick fix for the test would be to delay stopping the watcher until the callback is likely to have completed (I don't feel too bad about sleeping in this test since this test already sleeps in another place):
```diff
diff --git a/test/expect-tests/fsevents/fsevents_tests.ml b/test/expect-tests/fsevents/fsevents_tests.ml
index fbc5c6551..ff2fc5ee5 100644
--- a/test/expect-tests/fsevents/fsevents_tests.ml
+++ b/test/expect-tests/fsevents/fsevents_tests.ml
@@ -258,7 +258,8 @@ let%expect_test "raise inside callback" =
       raise Exit)
     (fun () ->
       Io.String_path.write_file "old" "foobar";
-      Io.String_path.write_file "old" "foobar");
+      Io.String_path.write_file "old" "foobar";
+      Unix.sleepf 1.0);
   [%expect {|
     [EXIT]
     exiting. |}]

```

emillon:
Thanks for the review and the fix. In terms of releasing a fix I'll take a flaky test against a regression, so I'll merge this now. We can discuss the fix for the flakiness separately.

PR #9011, opened 2023-10-27
chore(pkg): cleanup imports

Alizter:
- Use `open Import` everywhere in `dune_pkg`
- Replace Dune_lang with Dune_sexp where appropriate
- Use `Dune_sexp.Encoder/Decoder` where appropriate
- Remove unnecessary module qualifiers


rgrinberg:
Thanks

PR #9014, opened 2023-10-27
test(pkg): show inconsistency with with-doc in build vs lock situation

Alizter:


test/blackbox-tests/test-cases/pkg/opam-package-with-doc.t
```diff
@@ -0,0 +1,35 @@
...
+  - with-doc.0.0.1
+
+The filter is translated correctly.
+  $ cat dune.lock/with-doc.pkg 
+  (version 0.0.1)
+  
+  (build
+   (run
+    echo
+    (when %{pkg-self:with-doc} "Building with-doc")))
```
rgrinberg, line 27: This is an issue with our lock generation. This `when` should not be included b/c we evaluated `with-doc` to false when solving.

Alizter: This is with the `--experimental-filter-translation` option. When that is not set the filter is removed but the step stays.

PR #9015, opened 2023-10-27
test(pkg): show inconsistency with with-test in build vs lock solution

Alizter:
We demonstrate that the lock file solution does not contain test dependencies even though the build plan executes the with-test steps.

I've actually later changed this to enable translation of the filters (for whatever reason this is a separate option). Now all that is left is for a value to be given to `pkg-self:with-test`.


test/blackbox-tests/test-cases/pkg/opam-package-with-test.t
```diff
@@ -0,0 +1,36 @@
...
+  - with-test.0.0.1
+
+The build instructions have a conditional action as expected.
+  $ cat dune.lock/with-test.pkg 
+  (version 0.0.1)
+  
+  (build
+   (run
+    echo
+    (when %{pkg-self:with-test} "Building with-test")))
```
rgrinberg, line 28: There's a bug here, but this test describes it incorrectly. `{with-test}` should indeed be false when solving. We're just forgetting to simplify the build and install instructions according to this value.

Alizter:
The test description is incorrect. We can write a similar test with a correct description in the future. No need to keep this PR open.

PR #9019, opened 2023-10-27
fix(pkg): Check that git repo dir is accessible

Leonidas-from-XIV:
Instead of going ahead and running git against something that is not an accessible directory, this will at least raise an exception.

Are there better ways to solve it? Except for showing a possibly better error message, not really - if someone has replaced the folder with something we can't access there is no other option than to fail in some way or other.

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -46,26 +46,18 @@ let show { dir } (Rev rev) path =
   >>| Result.to_option
 ;;
 
-let dir_exists dir =
-  match Path.exists dir with
-  | false -> false
-  | true ->
-    (match Path.stat dir with
-     | Ok stats ->
-       (match stats.st_kind with
-        | S_DIR -> true
-        | _ -> false)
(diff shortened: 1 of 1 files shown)


Alizter:
One option would be to raise a user error and add a hint about deleting the directory, especially if its only Dune that is using it.

src/dune_pkg/rev_store.ml
```diff
@@ -48,13 +48,18 @@ let show { dir } (Rev rev) path =
...
-  (* TODO might as well double check it's a directory *)
-  match Path.exists dir with
-  | true -> Fiber.return t
-  | false ->
-    Path.mkdir_p dir;
-    let+ () = run t [ "init"; "--bare" ] in
-    t
+  let+ () =
+    match Fpath.mkdir_p (Path.to_string dir) with
+    | Already_exists -> Fiber.return ()
```
Alizter, line 53: Am I correct in thinking that the directory was being created before the fiber was being run before? Does putting the creation in the `let+` make it have a more predictable outcome?

rgrinberg:
Thanks. Merged with a few tweaks.

PR #6355, opened 2022-10-29
stdune: check if existing dir is actually a dir

Alizter:
We check that if a directory already exists that it is in fact a directory and not a file.

PR #9020, opened 2023-10-27
fix(pkg): Manage remotes without shelling out

Leonidas-from-XIV:
A mutex had to be added to prevent one fiber `git init --bare` while the other fiber happily loaded the repo and failed because the repo was not yet done initializing.

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -295,31 +295,46 @@ module Remote = struct
   ;;
 end
 
-let remote_exists dir ~name =
-  (* TODO read this directly from .git/config *)
-  let stdout_to = make_stdout () in
-  let stderr_to = make_stderr () in
-  let command = [ "remote"; "show"; name ] in
-  let+ (), exit_code =
-    let git = Lazy.force Vcs.git in
-    Process.run ~dir ~display ~stderr_to ~stdout_to Return git command
+let remote_header =
+  Re.(compile (seq [ bol; str "[remote \""; group (rep1 any); str "\"]"; eol ]))
+;;
+
+let remote_exists ~dir ~name =
+  Path.relative dir "config"
+  |> Io.lines_of_file
+  |> List.find_opt ~f:(fun line ->
+    match Re.exec_opt remote_header line with
+    | Some groups ->
+      let remote_name = Re.Group.get groups 1 in
+      String.equal remote_name name
+    | None -> false)
+  |> Option.is_some
+;;
+
+let remote_add ~dir handle source =
+  let git_config = Path.relative dir "config" in
+  let existing = Io.read_file git_config in
+  let stanza =
+    sprintf
+      {|%s
+
+  [remote "%s"]
+    url = %s
+    fetch = +refs/heads/*:refs/remotes/%s/*
+|}
+      existing
+      handle
+      source
+      handle
   in
-  match exit_code with
-  | 0 -> true
-  | 128 | _ -> false
+  Io.write_file git_config stanza
 ;;
 
 let add_repo ({ dir } as t) ~source =
-  (* TODO add this directly using .git/config *)
   let handle = source |> Dune_digest.string |> Dune_digest.to_string in
   let lock = lock_path t in
   with_flock lock ~f:(fun () ->
-    let* exists = remote_exists dir ~name:handle in
-    let* () =
-      match exists with
-      | true -> Fiber.return ()
-      | false -> run t [ "remote"; "add"; handle; source ]
-    in
+    if not (remote_exists ~dir ~name:handle) then remote_add ~dir handle source;
     let remote : Remote.t = { repo = t; handle } in
     let+ () = Remote.update remote in
     remote)


Leonidas-from-XIV:
Note to myself: this is most likely due to `load_or_create` called twice for two repos and the first one creates the directory, but isn't done with repo creation by the time the second runs, so the second one gets a `Rev_store.t` before the first one has finished running `git init --bare` and when the second one tries to add a remote, it fails.

bin/pkg.ml
```diff
@@ -320,8 +320,15 @@ module Lock = struct
         ]
   ;;
 
+  let xdg_repo_location =
+    let ( / ) = Filename.concat in
```
Alizter, line 324: This is kind of funny since it will differ on windows but later you use `/`. I think you should just convert the Xdg to a path directly and use `Path.L.relative` which takes in a list of args rather than a single on like `Path.relative`.

Alizter: I guess you didn't introduce this in this PR however, so we can do it later.

src/dune_pkg/rev_store.ml
```diff
@@ -128,30 +138,51 @@ module Remote = struct
...
-  let remote : Remote.t = { repo = t; handle } in
-  let+ () = Remote.update remote in
-  remote
+  Fiber.Mutex.with_lock lock ~f:(fun () ->
+    let exists = remote_exists ~dir ~name:handle in
+    let () =
+      match exists with
+      | true -> ()
+      | false -> remote_add ~dir handle source
+    in
```
Alizter, line 184:
```suggestion
    if not (remote_exists ~dir ~name:handle) then remote_add ~dir handle source;
```

bin/pkg.ml
Leonidas-from-XIV: You're right. Might as well bite the bullet here.

src/dune_pkg/rev_store.ml
```diff
@@ -177,11 +177,7 @@ let add_repo ({ lock; dir } as t) ~source =
   let handle = source |> Dune_digest.string |> Dune_digest.to_string in
   Fiber.Mutex.with_lock lock ~f:(fun () ->
     let exists = remote_exists ~dir ~name:handle in
-    let () =
-      match exists with
-      | true -> ()
-      | false -> remote_add ~dir handle source
-    in
+    if not (remote_exists ~dir ~name:handle) then remote_add ~dir handle source;
```
Alizter, line 180: The suggestion I made before didn't let me edit the line above. You need to also remove that.

src/dune_pkg/rev_store.ml
```diff
@@ -53,24 +52,20 @@ let show { dir; lock = _ } (Rev rev) path =
...
-      | Created -> run t [ "init"; "--bare" ]
-      | exception Unix.Unix_error (e, x, y) ->
-        User_error.raise
-          [ Pp.textf "%s isn't a directory" (Path.to_string_maybe_quoted dir)
-          ; Pp.textf "reason: %s" (Unix_error.Detailed.to_string_hum (e, x, y))
-          ]
-          ~hints:[ Pp.text "delete this file or check its permissions" ]
-    in
-    t)
+  let t = { dir; lock } in
```
Alizter, line 55: Remove line above and
```suggestion
  let t = { dir; lock = Fiber.Mutex.create () } in
```

src/dune_pkg/rev_store.ml
```diff
@@ -53,24 +52,20 @@ let show { dir; lock = _ } (Rev rev) path =
...
+  let* () = Fiber.return () in
+  let+ () =
+    match Fpath.mkdir_p (Path.to_string dir) with
+    | Already_exists -> Fiber.return ()
+    | Created -> run t [ "init"; "--bare" ]
+    | exception Unix.Unix_error (e, x, y) ->
+      User_error.raise
+        [ Pp.textf "%s isn't a directory" (Path.to_string_maybe_quoted dir)
+        ; Pp.textf "reason: %s" (Unix_error.Detailed.to_string_hum (e, x, y))
```
Alizter, line 64:
```suggestion
    | exception Unix.Unix_error unix_err ->
      User_error.raise
        [ Pp.textf "%s isn't a directory" (Path.to_string_maybe_quoted dir)
        ; Pp.textf "reason: %s" (Unix_error.Detailed.to_string_hum unix_err)
```

Leonidas-from-XIV: This seems to be one of the cases where constructors of arity 3 and 3-tuples are not treated the same, so unfortunately it has to be reconstructed.

Alizter: I'm trying to think how this can even occur. Would `Fpath.mkdir_p` return `Already_exists`? Have a look at `mkdir_p` inside `install_uninstall.ml`. There is even a change request for `Fpath` to handle this case better, and I even tried to solve it previously https://github.com/ocaml/dune/pull/6355 but it got reverted due to a perf bug.

Alizter:
`Rev_store.Remote.update` is unused it appears.

src/dune_pkg/rev_store.ml
```diff
@@ -5,39 +5,43 @@ module Display = Dune_engine.Display
 module Re = Dune_re
 open Fiber.O
 
-type t = { dir : Path.t }
+type t =
+  { dir : Path.t
+  ; lock : Fiber.Mutex.t
```
rgrinberg, line 10: What about using a file lock instead? We have an `Flock` module in `Dune_util` that's pretty useful.

Leonidas-from-XIV: The issue why I went for an in-memory solution is because the cleanup is simple: the due process ends (cleanly or otherwise - the mutex disappears).

With file-based solutions you have to deal with what happens if the lock file is not cleaned up. I took a look at flock and it seems the lock management is in the kernel, so I assume the kernel would know when to release the lock, which sidesteps that issue. Our API for it is blocking, is there a way to wrap a Fiber to await for the lock being released?

bin/pkg.ml
```diff
@@ -320,8 +320,17 @@ module Lock = struct
         ]
   ;;
 
+  let xdg_repo_location =
```
rgrinberg, line 323: I think this already exists in `opam_repo`, so I would suggest exporting it out of there and using here.

src/dune_pkg/rev_store.ml
rgrinberg: You can run the locking in a background thread. But tbh, that's probably not necessary. Feel free to block with a timeout waiting on the lock.

src/dune_pkg/rev_store.mli
```diff
@@ -31,5 +30,11 @@ module Remote : sig
 end
 
 val content_of_files : t -> File.t list -> string list Fiber.t
-val load_or_create : dir:Path.t -> t Fiber.t
+val load_or_create : lock:Dune_util.Flock.t -> dir:Path.t -> t Fiber.t
```
rgrinberg, line 33: I think the lock should stay an implementation detail of `Rev_store`. In other words, we shouldn't mention it in the mli.

src/dune_pkg/rev_store.ml
```diff
@@ -3,42 +3,60 @@ open Dune_vcs
...
 type rev = Rev of string
 
-let equal { dir } t = Path.equal dir t.dir
+let rec lock_limited flock lock timeout =
+  let sleep_duration = 0.1 in
+  match Flock.lock_non_block flock lock with
+  | Ok `Failure as v ->
+    if timeout > 0.0
+    then (
+      let timeout = timeout -. sleep_duration in
```
rgrinberg, line 22: You should make sure `timeout` is non-negative here. E.g. `let timeout = max 0. timeout`.

src/dune_pkg/rev_store.ml
```diff
@@ -2,43 +2,101 @@ open Stdune
...
+let with_flock lock_path ~f =
+  let open Fiber.O in
+  let parent = Path.parent_exn lock_path in
+  Path.mkdir_p parent;
+  let fd = Unix.openfile (Path.to_string lock_path) [ Unix.O_CREAT; O_RDONLY ] 0o644 in
+  let flock = Flock.create fd in
+  let max_duration = 5. in
+  Fiber.finalize
+    ~finally:(fun () ->
+      (* closing the fd releases the flock *)
```
rgrinberg, line 53: Should we remove the lock file as well? If it's present, it's a bit hard to say if any program is taking the lock at the moment.

rgrinberg:
LGTM but this needs some tests (the part about adding/removing remotes). Preferably some regular expect tests that don't pull all the package management machinery.

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: Implemented it via polling on a nonblocking lock, I think it's not too bad.

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: We can but existence and non-existence of the file doesn't indicate holding the lock as it isn't an atomic operation, for that one needs to query e.g. `lsof`. But we can remove it as a hint to the user that no lock is meant to be held.

src/dune_pkg/rev_store.ml
```diff
@@ -232,32 +289,48 @@ module Remote = struct
   ;;
 end
 
-let remote_exists { dir } ~name =
-  (* TODO read this directly from .git/config *)
```
emillon, line 236: I'm not convinced that this is a good idea to reimplement git logic and parsing its files in dune. It usually has good machine-readable interfaces.

Leonidas-from-XIV: Ok, I've rebased the PR and took out all the locking code, so we can discuss it on the merit of just the `git remote` handling.

The point of removing the `git` invocation is mostly that this is a bit faster and makes us a bit more independent from the `git` CLI (which granted, is rather unlikely to break). The `.git/config` format is fairly stable so just adding the lines in OCaml seems to be a fairly low-risk change and removes 1-2 `git` calls per configured repository.

Leonidas-from-XIV:
Closing the PR due to lack of interest, will remove the `TODO` comment in some PR in the future.

Alizter:
This PR went through a lot of changes. Is it only the inter-process locking that is left?

Leonidas-from-XIV:
@Alizter The inter-process locking that developed in this branch was split out and merged as #9140, so this branch just contains the originally planned feature of avoiding a few `git` calls.

PR #9021, opened 2023-10-27
fix(pkg) Update `default` repository to the git version of opam-repository

Leonidas-from-XIV:
We don't support the tarball anymore, but forgot to update the `:standard` repository.

Reported by @Alizter


Leonidas-from-XIV:
Since I ripped out OpamUrl, there is a discussion to be had what the format of the URLs/paths for the repos should be.

Should we support "whatever git accepts as remote"? Should we support "whatever OPAM supports as a repo-url" should we have our own format? What about passing options on the command line?

Alizter (approved):
This has been working for me, but needs a rebase.

Leonidas-from-XIV:
This was one of the simpler rebases :-)

test/blackbox-tests/test-cases/pkg/multiple-opam-repos.t
```diff
@@ -31,10 +31,10 @@ We have to define both repositories in the workspace, but will only use `new`.
   > (lang dune 3.10)
   > (repository
   >  (name new)
-  >  (source "git+file://$(pwd)/mock-opam-repository"))
+  >  (source "$(pwd)/mock-opam-repository"))
   > (repository
   >  (name old)
-  >  (source "git+file://$(pwd)/old-mock-opam-repository"))
+  >  (source "$(pwd)/old-mock-opam-repository"))
```
rgrinberg, line 37: How do we know that this is a git repo now?

Leonidas-from-XIV: We only support git repos here.

rgrinberg: Shouldn't we support repositories defined on the file system as well?

Leonidas-from-XIV: You can do that via the CLI, do you think it would be sensible to do that in `dune-workspace` as well?

rgrinberg: > Edit: The only way to use paths is via the CLI.

I see. I guess that's TODO. Since we need to support paths here, I wonder if we should just leave that `git+https` prefix. Otherwise, I don't see how we'll differentiate the two.

Leonidas-from-XIV: Yes, that's a good point. Another option would be to check if it is a path and then check if the path is a git repo and if not use it as regular path, but it is a bit magical and I am not sure if that wouldn't lead to a confusing UX.

rgrinberg: Yes, that would definitely be confusing. Especially given that one can point to a local git repo and expect one of the two behaviors:

1. Ignore untracked/unstaged changes. Corresponds to treating it like a git repo
2. Take those changes into account. Corresponds to treating it like a normal directory

So this should be explicit

Leonidas-from-XIV: Ok, makes sense; I'll add the distinction back. I'm sure I can coerce OpamUrl to give me more useful outputs or just parse it myself :)

Leonidas-from-XIV: Ok, I believe this should work. If OPAM suggests HTTP as backend, we pass it to git anyway, so `https://github.com/ocaml/opam-repository.git` will be treated the same as `git+https://github.com/ocaml/opam-repository.git`.

rgrinberg (approved):

PR #9025, opened 2023-10-27
fix: keep menhir .conflicts files (#6865)

Tchou:
Add the .conflicts file to targets of menhir stanzas when the --explain flag is present and menhir is not called with --only-tokens.


Alizter:
Could you add a test demonstrating that putting --explain in `env` doesn't work?

Tchou:
I don't understand. I added a test demonstrating that it does work in `env`. Or maybe the test is not complex enough ? like there should be several distinct `env` ?
The test is in `explain-env.t` with two sub-directories and two parsers where `--explain` works for both.

Note at this moment I can only test with `env` in a `dune` file. When I try to mention `menhir_flags` in `dune-workspace`, `dune` crashes (I opened an issue for that).

Or else I don't understand what not working means (which may very well be the case).

Alizter:
@Tchou I apologise, I read a little fast earlier. Let me take a closer look. You can disregard my comment from earlier.

test/blackbox-tests/test-cases/menhir/explain-env.t/run.t
```diff
@@ -0,0 +1,10 @@
+  $ dune build a/test.exe --debug-dependency-path
+  Warning: one state has shift/reduce conflicts.
+  Warning: one shift/reduce conflict was arbitrarily resolved.
+  $ ls _build/default/a/parser.conflicts
+  _build/default/a/parser.conflicts
```
Alizter, line 5: Could you do `dune build _build/default/a/parser.conflicts` directly so that we can see Dune recognizes it as a target and isn't producing it by accident.

Alizter: The reason I ask is that if Dune isn't actually aware of the .conflicts file in this situation then it cannot be depended on, or expected to persist between builds.

src/dune_rules/menhir/menhir_rules.ml
```diff
@@ -263,26 +265,32 @@ module Run (P : PARAMS) = struct
...
+     [.conflicts] file is always generated (if Menhir succeeds in creating a
+     [.ml/.mli] file).
+  *)
 
   let process (stanza : stanza) : unit Memo.t =
     let base = Option.value_exn stanza.merge_into in
-    let ocaml_type_inference_disabled, cmly =
+    let ocaml_type_inference_disabled, cmly, conflicts =
       Ordered_set_lang.Unexpanded.fold_strings
         stanza.flags
```
Alizter, line 279: Upon reading it seems that `stanza.flags` doesn't contain `--explain` yet if it comes from the env. --explain from the env only becomes available in process3 it seems. It's tricky to know --explain is in env without first expanding and I got stuck with my PR because it is too early to expand here.

Tchou: I'm not sure I get the subtlety, but here is what I could test:
- dune does not seem to recognize .conflicts as targets (is it because they are `Hidden_targets` ?) 
```
Error: Don't know how to build _build/default/a/parser.conflicts
```
- it does not seem to be related to `env`. The other test I added (which uses `flags` in the menhir stanza directly) has the same problem (.conflicts not being a target)
- This is also the case for cmly files (Don't know how to build).
- This is also the case for .s files produced by `ocamlopt`. Indeed, if I put `(ocamlopt_flags -S)`, the .s assembly files are present in _build/.../native/foo__bar.s but I have the same error when I try to `dune build` them.

In any case, I don't think this is an issue for most menhir users (or at least no more than the other two cases that don't "work" either and are unrelated to this flag). 
These .conflicts files are really transient files that are used to debug one's grammar, they are not to be committed to a projects repository (at least no more than .s files generated by ocamlopt). One exception would be `menhir` developers that would have a use for that, but this is another, far less common use case.

Alizter: Right, but you get the same behaviour with menhir today if you pass `--explain` before your PR. What your PR tries to do is declare `.conflict` files as targets of the menhir rule when `--explain` is passed in the stanza. Being a hidden target doesn't matter here, Dune should still be able to build it.

ocamlopt_flags -S is also not a good example because Dune does not know anything about -S and so does not produce the .s files as targets of the OCaml module rules. This means that those files that are produced will not be known by Dune hence they can be deleted or cleaned up causing users to miss them.

I definitely agree that this is a useful feature to have, but it has proven to be tricky to implement correctly. In my old PR #7607 you should see that `.conflicts` files are at least recognized when `--explain` is passed in the stanza. For env the situation is a bit trickier, but I would be OK with having a partial solution where `--explain` works in a stanza but not in env.

Tchou: I don't get the same behavior. I have a more complex project where the .conflicts files are not present in _build. My understanding is that it is due to sandboxing but when I use --sandbox=none the file still isn't there (and I still see in the logs, cd .sandbox/... before invocations of menhir).

If I build my project with this PR, the .conflicts files appears (regardless of --sandbox). Or is it possible I picked up a commit between 3.11.1 and main that fixes this ?

Tchou: Arrrg. Scracth that. I understand now what you mean (sorry).
So both my test cases were buggy because they use `(land dune 3.0)` but sandboxing is forced starting with `(lang dune 3.5)` for menhir. And indeed if I change the dune version, the simple test passes but not the env. (Which is still an improvement I would say).
I'll investigate further.

Alizter: @Tchou no worries. Let me know if you have any questions about the code.

Tchou:
New attempt (the diff with main is small so I squashed and force pushed since the solution I ended up with handle both cases).

I think this works well but the intricacies and differences of `Memo.t` vs `Action_builder.t` are lost on me. I added a few more test cases to also handle `.cmly` files. With this commit, everything should work as intended:
- .cmly and .conflicts files are preserved in _build
- they are correctly recognized as targets (`dune build _build/default/foo.{conflicts,cmly}` is happy)
- it works whether the flags are passed through the `menhir` stanza or through `env`.

Alizter:
@Tchou What's left to undraft this?

Tchou:
Hi,
sorry it slipped by. So it's not satisfactory at the moment since I don't understand how things could work in all generality. So long story short:
- if --explain files are passed in env or in the dune file directly all is well (with current patch)
- BUT if the flags are given with an :include file that fails (dune actually report some internal error due to me doing crazy stuff). And this breaks menhir CI too since in the test cases of the menhir tools, include files are used to test menhir with various flags (what I mean is that in my branch, will musing around trying to understand how I could test the content 
of :include files, i must have broken it, I can of course backtrack but it feels silly to not handle this case).

I have gone through dune sources and I cannot understand how to do get the *expanded* include file, from within the menhir plugin. Ideally it should go like this:
- wait until the include file is expanded as actual flags
- at that moment parse the flags and add new targets (the .conflict file) like it is done now.
My limited understanding is that an :included file is also a dependency of the rule (so that the rule is recomputed if the file changes) but then expansion seem to occur pretty late.

I couldn't understand dune's architecture well enough to know whether this is possible.
If there is some documentation or examples of other plugins/rules doing it I could try to look into it again.

nojb:
> Ideally it should go like this:
> 
> * wait until the include file is expanded as actual flags
> * at that moment parse the flags and add new targets (the .conflict file) like it is done now.
>   My limited understanding is that an :included file is also a dependency of the rule (so that the rule is recomputed if the file changes) but then expansion seem to occur pretty late.
> 
> I couldn't understand dune's architecture well enough to know whether this is possible.

Hello Kim! @rgrinberg or @snowleopard will correct me, but I understand that at the moment targets must be statically known by Dune, so it is not possible to have the set of targets depend on things like `:include` which need Dune to build stuff.

rgrinberg:
Replaced by #9512

Issue #9017, opened 2023-10-27
pkg: unable to depend on `seq` package


Alizter:
There are legacy library compatibility packages for OCaml such as `seq`. When we translate their opam files we have no build plan:
```
(version base)

(deps ocaml)
```
which is fine however this leads to the following problems when building:
```
DUNE_PKG_OVERRIDE_OCAML=1 ~/dune/_build/install/default/bin/dune build
Error:
open(_build/_private/default/.pkg/seq/source/seq.install): No such file or directory
-> required by _build/_private/default/.pkg/seq/target/cookie
```
Ideally in this situation, since there is no install step, we should not try to depend on seq.install or at least generate a dummy value.

Alizter:
Specifically it seems that `seq` includes `seq.install` in its `extra-files` field.

Alizter:
Here is the repro:
- https://github.com/ocaml/dune/pull/9026

Alizter:
We are getting further with this build, now the permissions are read-only causing issue:
```
Error:
open(_build/_private/default/.pkg/seq/source/META.seq): Permission denied
-> required by _build/_private/default/.pkg/seq/target/cookie
```

rgrinberg:
Should be fixed. Please re-open with a test case if something is still broken.

Alizter:
I can confirm that this now builds.

PR #9026, opened 2023-10-27
test(pkg): demo files not being copied in empty builds

Alizter:
Reproduction case for #9017

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/opam-package-with-files-install.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-package-with-files-install.t
@@ -1,25 +1,18 @@
-We make sure that file like foo.install are correctly included from the files
-subdirectory of an opam repository. Notably, the build step here is empty. The
-library seq is such an example in the wild with this behaviour.
+This test demonstrates a package where the .install file being created by the
+file copying step rather than the build step.
 
   $ . ./helpers.sh
-  $ mkrepo
 
-  $ mkpkg foo
-  $ mkdir -p $mock_packages/foo/foo.0.0.1/files
-  $ cat > $mock_packages/foo/foo.0.0.1/files/foo.install
+  $ mkdir -p dune.lock/foo.files
+  $ cat >dune.lock/lock.dune <<EOF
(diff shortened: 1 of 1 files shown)


Alizter:
I have a fix for this ready, though I am not 100% I understand why it works.

test/blackbox-tests/test-cases/pkg/opam-package-with-files-install.t
```diff
@@ -0,0 +1,22 @@
...
+  $ solve foo
+  Solution for dune.lock:
+  - foo.0.0.1
+
+The build action copying files/ is missing.
+  $ cat dune.lock/foo.pkg
+  (version 0.0.1)
+
+The foo.install file in files/ should have been copied over.
+  $ build_pkg foo 2> /dev/null
```
rgrinberg, line 21: Don't think this step is necessary.

test/blackbox-tests/test-cases/pkg/opam-package-with-files-install.t
```diff
@@ -0,0 +1,22 @@
...
+
+  $ mkpkg foo
+  $ mkdir -p $mock_packages/foo/foo.0.0.1/files
+  $ cat > $mock_packages/foo/foo.0.0.1/files/foo.install
+
+  $ solve foo
+  Solution for dune.lock:
+  - foo.0.0.1
+
+The build action copying files/ is missing.
```
rgrinberg, line 16: There is no "build action" for copying files/

Alizter: You're right, I've rewritten it to something clearer now.

test/blackbox-tests/test-cases/pkg/opam-package-with-files-install.t
Alizter: I've added a check for the .files directory. You can now see that it is in the lock file.

rgrinberg:
The solving step wasn't necessary so I got rid of it.

Thanks for the PR.

PR #9032, opened 2023-10-28
refactor(pkg): simplify package loading

rgrinberg:
There's no difference between Ok [] and Error `Package_not_found. So we
just use the empty list to represent that we haven't found any packages.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -132,24 +132,20 @@ module Context_for_dune = struct
       in
       Fiber.return [ version, Ok opam_file ]
     | None ->
-      let+ all_versions = Opam_repo.load_all_versions t.repos name in
-      (match all_versions with
-       | Error `Package_not_found ->
-         (* The CONTEXT interface doesn't give us a way to report this type of
-            error and there's not enough context to give a helpful error message
-            so just tell opam_0install that there are no versions of this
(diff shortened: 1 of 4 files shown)


Leonidas-from-XIV (approved):
Yep, the weird conversion of `option`, `result` and empty/non-empty lists was a mess that accumulated over multiple PRs, good to clean it up finally.

src/dune_pkg/opam_solver.ml
```diff
@@ -132,24 +132,20 @@ module Context_for_dune = struct
...
-         let opam_files_in_priority_order =
-           List.sort opam_files ~compare:(opam_version_compare t)
-         in
-         List.map opam_files_in_priority_order ~f:(fun opam_file ->
-           let opam_file_result =
-             if is_opam_available t opam_file then Ok opam_file else Error Unavailable
-           in
-           OpamFile.OPAM.version opam_file, opam_file_result))
+      let+ opam_files = Opam_repo.load_all_versions t.repos name in
+      (* The CONTEXT interface doesn't give us a way to report this type of
```
Alizter, line 136: Is this comment still helpful/relevant?

rgrinberg: Not particularly, but I already have other PR's in this file, so removing it would cause conflicts.

PR #9033, opened 2023-10-28
refactor: re-implement [Appendable_list]

rgrinberg:
This seems to give us a small reduction in allocations and a corresponding reduction in build times: https://autumn.ocamllabs.io/ocaml/dune/pull/9033/base/main?worker=autumn&image=bench.Dockerfile

cc @snowleopard

### DIFF
--- a/otherlibs/stdune/src/appendable_list.ml
+++ b/otherlibs/stdune/src/appendable_list.ml
@@ -1,13 +1,67 @@
-type 'a t = 'a list -> 'a list
+type 'a t =
+  | Empty
+  | Singleton of 'a
+  | Cons of 'a * 'a t
+  | List of 'a * 'a * 'a list
+  | Append of 'a t * 'a t
+  | Concat of 'a t list
 
-let empty k = k
-let singleton x k = x :: k
-let to_list l = l []
-let ( @ ) a b k = a (b k)
-let cons x xs = singleton x @ xs
+let empty = Empty
+let singleton x = Singleton x
 
-let rec concat l k =
-  match l with
-  | [] -> k
-  | t :: l -> t (concat l k)
+let ( @ ) a b =
+  match a, b with
+  | Empty, _ -> b
+  | _, Empty -> a
+  | Singleton a, Singleton b -> List (a, b, [])
+  | Singleton a, _ -> Cons (a, b)
+  | _, _ -> Append (a, b)
+;;
+
+let cons x xs =
+  match xs with
+  | Empty -> Singleton x
+  | List (y, z, rest) -> List (x, y, z :: rest)
+  | _ -> Cons (x, xs)
+;;
+
+let to_list_rev =
+  let rec loop acc stack =
+    match stack with
+    | [] -> acc
+    | t :: stack ->
+      (match t with
+       | Empty -> loop acc stack
+       | Singleton x -> loop (x :: acc) stack
+       | Cons (x, xs) -> loop (x :: acc) (xs :: stack)
+       | List (x, y, xs) -> loop (List.rev_append xs (y :: x :: acc)) stack
+       | Append (xs, ys) -> loop acc (xs :: ys :: stack)
+       | Concat [] -> loop acc stack
+       | Concat (x :: xs) -> loop acc (x :: Concat xs :: stack))
+  in
+  fun t -> loop [] [ t ]
+;;
+
+let to_list xs = List.rev (to_list_rev xs)
+
+let is_empty = function
+  | Empty -> true
+  | _ -> false
+;;
+
+let rec concat = function
+  | [] -> Empty
+  | [ x ] -> if is_empty x then Empty else x
+  | x :: xs as list -> if is_empty x then concat xs else Concat list
+;;
+
+let of_list = function
+  | [] -> Empty
+  | [ x ] -> Singleton x
+  | x :: y :: xs -> List (x, y, xs)
+;;
+
+let rec of_list_concat = function
+  | [] -> Empty
+  | x :: xs as list -> if is_empty x then of_list_concat xs else Concat list
 ;;
--- a/otherlibs/stdune/src/appendable_list.mli
+++ b/otherlibs/stdune/src/appendable_list.mli
@@ -9,3 +9,7 @@ val ( @ ) : 'a t -> 'a t -> 'a t
 val cons : 'a -> 'a t -> 'a t
 val concat : 'a t list -> 'a t
 val to_list : 'a t -> 'a list
+val to_list_rev : 'a t -> 'a list
+val of_list : 'a list -> 'a t
+val is_empty : _ t -> bool
+val of_list_concat : 'a t list -> 'a t
--- a/otherlibs/stdune/test/appendable_list_tests.ml
+++ b/otherlibs/stdune/test/appendable_list_tests.ml
@@ -55,3 +55,17 @@ let%expect_test "concat" =
     8
     9 |}]
 ;;
+
+let%expect_test "is_empty" =
+  let assert_empty l = assert (Al.is_empty l) in
+  assert_empty Al.empty;
+  [%expect {||}];
+  assert_empty @@ Al.concat [];
+  [%expect {||}];
+  assert_empty @@ Al.concat [ Al.empty ];
+  [%expect {||}];
+  assert_empty @@ Al.concat [ Al.empty; Al.empty ];
+  [%expect {||}];
+  assert_empty @@ Al.of_list [];
+  [%expect {||}]
+;;
--- a/src/dune_engine/load_rules.ml
+++ b/src/dune_engine/load_rules.ml
@@ -380,8 +380,8 @@ end = struct
   ;;
 
   let compute_alias_expansions ~(collected : Rules.Dir_rules.ready) ~dir =
-    let aliases = collected.aliases in
     let+ aliases =
+      let aliases = collected.aliases in
       if Alias.Name.Map.mem aliases Alias.Name.default
       then Memo.return aliases
       else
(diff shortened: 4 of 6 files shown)


otherlibs/stdune/src/appendable_list.ml
```diff
@@ -1,13 +1,38 @@
...
-let cons x xs = singleton x @ xs
-
-let rec concat l k =
-  match l with
-  | [] -> k
-  | t :: l -> t (concat l k)
+type 'a t =
+  | Empty
+  | Singleton of 'a
+  | Cons of 'a * 'a t
```
snowleopard, line 4: I wonder if it's also worth having a `List of 'a list` constructor to provide a cheap `of_list` function. Are there any call sites that end up calling ``Cons` a bunch of times to turn a list into an `Appendable_list`?

snowleopard (approved):
Nice, I was about to do that myself :) Let me run some benchmarks internally.

otherlibs/stdune/src/appendable_list.ml
snowleopard: I checked, and it looks like there are no such places. (There are also no calls for `Appendable_list.cons`, so not sure, maybe we could nuke it too... or add `List` anyway.)

snowleopard (changes requested):
Actually, benchmarks don't look very good. I'm seeing a slight regression in build times.

rgrinberg:
I think I understand the issue. There's a place where we're using it where it isn't really appropriate. Will send a PR, one second.

snowleopard:
Apparently @pmwhite has been refactoring appendable lists too, so I'm adding him as a reviewer. His implementation seems to show slightly better results on our benchmarks.

rgrinberg:
@snowleopard I improved the performance a little further at the cost of some refactoring and using the reverse order when computing the alias expansion. The reverse order still produces the same build artifacts, but changes the order of actions running.

otherlibs/stdune/src/appendable_list.ml
rgrinberg: I added it anyway. There's no cost to it, and it would be bad if we'd get a performance penalty just for using this.

pmwhite:
Yeah, I've been refactoring this module too. My goal was not to make it faster or take less memory, but just to make it support the `is_empty` function efficiently; the function representation has to first be converted to a list before it can be inspected.

The reason I want `is_empty` is that I think `Dune_engine.Rules.t` would be better to represent as a `Appendable_list.t` rather than a `Id.Map.t`, so that merging two maps is cheaper. My experiments seems to demonstrate that it is worthwhile, but I haven't gotten around to cleaning it up into a series of PRs (the first of which was going to be something similar to this).

Anyway, the representation I used was just `Empty | Cons | Append`, which effectively left the module's behavior the same - I merely replaced closure allocation with explicit constructor allocation. It may also have been a very slight performance improvement due to fewer indirect function calls.

rgrinberg:
Okay, so I think the important parts of our changes are independent. My representation is basically the same as yours with an extra optimization with taking the union of a list of appendable lists and removing an unnecessary list reversal

otherlibs/stdune/src/appendable_list.ml
```diff
@@ -1,13 +1,57 @@
...
+  | xs -> Concat xs
+;;
+
+let of_list = function
+  | [] -> Empty
+  | [ x ] -> Singleton x
+  | xs -> List xs
+;;
+
+let of_list_concat xs = Concat xs
```
snowleopard, line 57: I think we should check the empty and singleton list cases, like above, so we can always check for emptiness cheaply.

otherlibs/stdune/src/appendable_list.ml
snowleopard: Ack, thanks!

otherlibs/stdune/src/appendable_list.ml
```diff
@@ -1,13 +1,57 @@
-type 'a t = 'a list -> 'a list
+type 'a t =
+  | Empty
+  | Singleton of 'a
+  | Cons of 'a * 'a t
+  | Append of 'a t * 'a t
+  | List of 'a list
+  | Concat of 'a t list
```
snowleopard, line 7: How about
```suggestion
  | List of 'a * 'a * 'a list
  | Concat of 'a t * 'a t * 'a t list
```
to rule out lists shorter than 2 elements? We've got no use for them, I think.

I wanted to write: 

> This has the advantage that it's obvious from the type that there is only one way to create an empty list -- using the `Empty` constructor.

but then realised that it's unfortuantely not true due to the `Concat Empty Empty []` case, which is a bit sad.

otherlibs/stdune/src/appendable_list.ml
rgrinberg: Done.

otherlibs/stdune/src/appendable_list.ml
rgrinberg: We've got no use for short lists, but these short lists aren't going to be more efficient. There's no real difference between `List (x :: y :: xs)` and `Cons (x, Cons (y, List xs))`.

> but then realised that it's unfortuantely not true due to the Concat Empty Empty Empty [] case, which is a bit sad.

Fixed this. Now we always use `Empty` for the empty list.

otherlibs/stdune/src/appendable_list.mli
```diff
@@ -9,3 +9,6 @@ val ( @ ) : 'a t -> 'a t -> 'a t
 val cons : 'a -> 'a t -> 'a t
 val concat : 'a t list -> 'a t
 val to_list : 'a t -> 'a list
+val to_list_rev : 'a t -> 'a list
+val of_list : 'a list -> 'a t
```
pmwhite, line 13: I like the addition of `to_list_rev`, since it makes the function tail-recursive without paying the extra cost of reversing the list at the end. Do we even need `to_list` now? I realize its usefulness depends on the context, but it seems kinda nice to delete it altogether so that we can't accidentally use it instead of `to_list_rev` without thinking.

otherlibs/stdune/src/appendable_list.ml
```diff
@@ -1,13 +1,65 @@
...
+  in
+  fun t -> loop [] [ t ]
+;;
+
+let to_list xs = List.rev (to_list_rev xs)
+
+let concat = function
+  | [] -> Empty
+  | [ x ] -> x
+  | xs -> Concat xs
```
pmwhite, line 48: The `[ Empty; Empty ]` case doesn't get converted to `Empty`, which makes `is_empty` incorrect.

pmwhite (changes requested):

pmwhite:
I think we should add some tests in this PR. I've already written some for the interface that existed before; give me a few minutes and I'll have a PR ready for you...

pmwhite:
See https://github.com/ocaml/dune/pull/9043. Could we merge that, rebase this on top of it, and then add more tests, especially around the interaction between `is_empty` and other functions?

otherlibs/stdune/src/appendable_list.ml
snowleopard: I don't quite follow your example with lists. Surely `List (x, y, xs)` is a better representation than either the one you get from standard lists or from `Cons` cells. It we've already gone through the trouble of eliminating short list cases, it seems nice to benefit from this in the types and in terms of removing a couple of unnecessary memory indirections.

otherlibs/stdune/src/appendable_list.mli
rgrinberg: No, we don't use it. I just added it for completeness. I'll get rid of it.

otherlibs/stdune/src/appendable_list.ml
rgrinberg: Fixed now.

otherlibs/stdune/src/appendable_list.ml
rgrinberg: Ah yeah, indeed. I misread your suggestion. Modified `List` to be a little more efficient. We could also introduce a constructor for `Tuple of 'a * 'a` (in the same spirit as `Singleton`). To really make two elements as efficient as possible.

otherlibs/stdune/src/appendable_list.ml
```diff
@@ -1,13 +1,67 @@
...
+
+let of_list = function
+  | [] -> Empty
+  | [ x ] -> Singleton x
+  | x :: y :: xs -> List (x, y, xs)
+;;
+
+let rec of_list_concat = function
+  | [] -> Empty
+  | x :: xs as list -> if is_empty x then of_list_concat xs else Concat list
```
snowleopard, line 66: It seems possibly useful to filter out all `Empty` elements instead of just those in the beginning of the list.

Also, we can handle the singleton case more efficiently (i.e. return `x` for `[x]` after filtering).

otherlibs/stdune/src/appendable_list.ml
snowleopard: Thanks! Not sure how useful `Tuple` is going to be in practice but yeah, we could add it too.

Note that `Concat` is similar to `List`: we can guarantee that it has at least two elements.

otherlibs/stdune/src/appendable_list.ml
```diff
@@ -1,13 +1,67 @@
...
-let singleton x k = x :: k
-let to_list l = l []
-let ( @ ) a b k = a (b k)
-let cons x xs = singleton x @ xs
-
-let rec concat l k =
-  match l with
-  | [] -> k
-  | t :: l -> t (concat l k)
+type 'a t =
```
snowleopard, line 1: I may be overengineering things, but with GADTs we could ensure that all recursive `'a t`'s are non-empty with no performance cost.

Alizter:
There appears to be a regression in the build times according to the bench.

https://autumn.ocamllabs.io/ocaml/dune/pull/9033/base/main?worker=autumn&image=bench.Dockerfile

Monorepo bench mark seems to agree

https://autumn.ocamllabs.io/ocaml/dune/pull/9033/base/main?worker=fermat&image=bench/monorepo/bench.Dockerfile

Is this reproducible in the Jst benchmarks?

rgrinberg:
I think the cause of this regression is the fact that we're putting too much work in maintaining a compact representation of this data structure. An `Appendable_list.t` is a transient thing that is only used to accumulate something and then throw it away once it's turned into a list.

Issue #8686, opened 2023-09-18
`file-depends` support


rgrinberg:
Opam is allowed to include files and their checksums in `<pkgname.config>`. E.g.:

```
file-depends: [
  [ "/foo/bar" "checksum" ]
]
```

This is basically a dynamic dependency mechanism like in our dynamic action plugin. We can discard the hash and just add a dependency on the file in question.

Alizter:
When do we add the dependency? The <pkg>.config files exist in the sources of the repos themselves, sometimes even being generated using .in files.

This means that we cannot add this dependency in the lock file for the package. The build step could generate the config files so it might be possible to include the dependency then. Or the install step might be an ideal place to do it.

rgrinberg:
> When do we add the dependency?

Right after the .config file is parsed. Which I think is done when we create the build cookie.

Alizter:
As an update I had a go at this, but from what I understand there doesn't seem to be a way to add dynamic dependencies to custom actions. Notably, dune_action_plugin is using it's own thing. There appears to be a CR in action_intf about adding such a "needs more dependencies" result in the action part of a custom action.

PR #9035, opened 2023-10-28
test(pkg): file-depends

Alizter:
We test the file-depends field of the .config file of an opam package and demonstrate that currently it doesn't work as expected.

Demonstrates expected behaviour that a solution to #8686 should have.


rgrinberg:
Thanks

PR #9036, opened 2023-10-28
refactor: use [Unix.rename] over [Sys.rename]

rgrinberg:
It gives better errors

### DIFF
--- a/otherlibs/stdune/src/path.ml
+++ b/otherlibs/stdune/src/path.ml
@@ -1342,7 +1342,7 @@ let set_of_build_paths_list =
 ;;
 
 let set_of_external_paths set = External.Set.to_list set |> Set.of_list_map ~f:external_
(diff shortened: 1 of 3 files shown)


snowleopard (approved):
The change looks good though it's hard to be sure you found all the call sites handling exceptions.

Also: you mentioned better error messages but no test changed. Could you give an example?

rgrinberg:
> The change looks good though it's hard to be sure you found all the call sites handling exceptions.

It's not too bad to look over all of them since they're just `Sys_error` and we're only using this exception sporadically. Going further with this, I think we should get rid of all instances of `Sys_error` by not using the Io functions from Sys. It's just a less precise version of `Unix_error`

> Also: you mentioned better error messages but no test changed. Could you give an example?

Indeed, let me try to create a test case.

PR #9037, opened 2023-10-29
refactor(pkg): deforest rule/aliases collection

rgrinberg:
Small simplification of rule/alias collection that also makes the code a little shorter


snowleopard (approved):
Looks good, thanks!

snowleopard:
I've imported this change internally.

Issue #8854, opened 2023-10-05
rfc(engine): sandbox only when needed rather than throwing an error


Alizter:
## Problem
When a rule is configured to be sandboxed, however the action of the rule does not need sandboxing, we raise the following error message:
```ocaml
            User_error.raise
              ~loc
              [ Pp.text
                  "Rule dependencies are configured to require sandboxing, but the rule \
                   has no actions that could potentially require sandboxing."
              ]
```
This can appear in some rather cryptic locations such as:
- #8848 
  - which led to a fix like https://github.com/ocaml/dune/pull/8849
- https://github.com/ocaml/dune/pull/8443#issuecomment-1689391531

In this situation, it doesn't really make alot of sense that raise a user error for the following reasons:
1. Because incorrect sandbox configurations are generally out of the users control and can be internal to dune.
2. We have a better alternative solution which is to simply not sandbox.

## Proposal 1

We should modify the behaviour in this case to simply conform to the sandboxing needs. If an action doesn't need to be sandboxed, then we should simply not sandbox it.

## Proposal 2

In the aforementioned cases, the action in question was in fact the "empty actions" which is just `(progn)`. Perhaps it is vacuously useful to sandbox in this case and we should just change it?

cc @rgrinberg @snowleopard

Alizter:
This has appeared again when trying to fix #9026. The situation is that an empty `(progn)` can't be sandboxed, but that's ok and having it default to no sandboxing seems sensible.

snowleopard:
I think we should just sandbox such actions. It doesn't seem particularly useful to insist on some sort of optimality. The current behaviour seems pretty annoying (imagine an action often flipping between requiring and not requiring sandboxing).

Alizter:
A conservative fix for the issues I've been encountering would be to mark `Prog []` as needing sandboxing. This would leave the error for other situations.

snowleopard:
I don't understand why we need to be conservative. I'm inclined to just remove the error and do the sandboxing if the sandboxing configuration asks for it.

PR #9039, opened 2023-10-29
fix(pkg): fix to allow seq to build

Alizter:
Here is a way of fixing #9017. It seems that the `expand` step is necessary for the copied `.install` file to be available. I'm not entirely sure why. In order to allow the `expand` step to always work, I needed to use the empty action rather than not calling `expand` at all. Unfortunately due to strict requirements with sandboxing (see #8854) we need a workaround to disable the sandbox in that case.

@rgrinberg would you be able to lend a hand and explain how to fix this properly?

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -805,22 +805,28 @@ module Action_expander = struct
     }
   ;;
 
-  let expand =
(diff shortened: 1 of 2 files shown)


rgrinberg:
Well this turned out to be a lot more involved than I expected. Should be fixed now.

PR #9041, opened 2023-10-30
fix: allow internal dune actions to be sandboxed

rgrinberg:
Internal actions that that don't need to be sandboxed would prevent the
entire rule from being sandboxed. Now we just sandbox them even if it's
not necessary.

### DIFF
--- a/src/dune_engine/action.ml
+++ b/src/dune_engine/action.ml
@@ -319,7 +319,6 @@ let is_useful_to distribute memoize =
     | false -> Clearly_not
 ;;
 
-let is_useful_to_sandbox = is_useful_to false false
 let is_useful_to_distribute = is_useful_to true false
 let is_useful_to_memoize = is_useful_to true true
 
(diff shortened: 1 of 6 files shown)


src/dune_engine/build_system.ml
```diff
@@ -470,22 +470,10 @@ end = struct
       let* () = Targets.maybe_async (fun () -> Path.mkdir_p (Path.build dir)) in
       let is_action_dynamic = Action.is_dynamic action.action in
       let sandbox_mode =
-        match Action.is_useful_to_sandbox action.action with
```
rgrinberg, line 473: The function `Action.is_useful_sandbox` can be deleted. Or should we keep it?

snowleopard: Let's delete it.

snowleopard (approved):
Thanks, the changes look good to me but let's also delete `Action.is_useful_sandbox`.

src/dune_engine/action.ml
```diff
@@ -319,7 +319,6 @@ let is_useful_to distribute memoize =
     | false -> Clearly_not
 ;;
 
-let is_useful_to_sandbox = is_useful_to false false
```
snowleopard, line 322: Looks like the lines below can now be simplified (the first argument can just be dropped).

rgrinberg: done here https://github.com/ocaml/dune/pull/9058

PR #9043, opened 2023-10-30
tests: Add tests for Appendable_list

pmwhite:
This was reviewed internally by @dkalinichenko-js. Feel free to review this yourself, but I think we could also just merge it straightaway.


Alizter (approved):
LGTM in context of #9033

pmwhite:
Thanks. Yeah, there's definitely a lot more stuff we could test. This PR wasn't really meant to be a full-fledged test suite for the module. I just figured since I had this bit of code lying around, we might as well merge it before doing any refactoring to the module itself. (which is what is happening in #9033)

PR #9044, opened 2023-10-30
test(pkg): make test more repro

Alizter:
This was out of order for me when I ran the test, so I switched it to test for the existence of the files directly.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/opam-repository-download.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-repository-download.t
@@ -35,9 +35,8 @@ Make a mock repo tarball that will get used by dune to download the package
 Our custom cache folder should be populated with the unpacked tarball
 containing the repository:
 
-  $ find fake-xdg-cache | grep HEAD
-  fake-xdg-cache/dune/git-repo/HEAD
-  fake-xdg-cache/dune/git-repo/FETCH_HEAD
+  $ [ -f fake-xdg-cache/dune/git-repo/HEAD ]
+  > [ -f fake-xdg-cache/dune/git-repo/FETCH_HEAD ]
 
 Automatically download the repo and make sure lock.dune contains the repo hash
 (exit code 0 from grep, we don't care about the output as it is not


test/blackbox-tests/test-cases/pkg/opam-repository-download.t
```diff
@@ -35,9 +35,8 @@ Make a mock repo tarball that will get used by dune to download the package
 Our custom cache folder should be populated with the unpacked tarball
 containing the repository:
 
-  $ find fake-xdg-cache | grep HEAD
-  fake-xdg-cache/dune/git-repo/HEAD
-  fake-xdg-cache/dune/git-repo/FETCH_HEAD
+  $ [ -f fake-xdg-cache/dune/git-repo/HEAD ]
+  > [ -f fake-xdg-cache/dune/git-repo/FETCH_HEAD ]
```
rgrinberg, line 39: Shouldn't there be `&&` to make sure the last check isn't ignored?

Alizter: I suppose I could change the `>` to a `$` if you are worried that they will intefere.

rgrinberg: > It currently should be equivalent to ; rather than &&.

That's exactly the problem. Add the following line to your test before the check:
```
$ rm fake-xdg-cache/dune/git-repo/HEAD
```
And note that your test is still passing.

Leonidas-from-XIV, line 39: Oops, thanks for noticing this.

I don't think it makes sense for to test the `FETCH_HEAD`. It was just in the original test because `HEAD` found it and there wasn't enough reasons speaking for excluding it.

But I'll ask a different question: why not just plop in `sort`? I forgot to add it when I revised the test, but it's a rather decent solution that keeps the shell commands fairly readable and the cram test meaningful.

Alizter: @Leonidas-from-XIV sort also works too. I thought this would be simpler as it directly checks the existence rather than relying on `find`.

PR #9046, opened 2023-10-30
fix(pkg): improve version loading

rgrinberg:
* do not load overriden versions
* load all versions rather than the first found in th repo


Leonidas-from-XIV (approved):
I think this is indeed an improvement. I was wondering whether it makes sure that each version is only found in the first repo that is mentioned in `ts` but following that logic this seems to be the case!

PR #9048, opened 2023-10-30
pkg: fix architecture detection on windows

### LINKED ISSUE #8866: pkg: unable to lock on windows
Locking on Windows gives the following:
```
PS C:\Users\ali\dune> with-dkml .\dune.exe pkg lock
Error: exception Failure("Windows only. This function isn't implemented.")
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from OpamStubs.that's_a_no_no in file
  "vendor/opam/src/core\\opamStubs.ml" (inlined), line 13, characters 23-80
Called from Dune_pkg__Sys_poll.arch in file "src/dune_pkg\\sys_poll.ml", line
  55, characters 15-43
Called from Dune_pkg__Sys_poll.sys_bindings.entry in file
  "src/dune_pkg\\sys_poll.ml", line 210, characters 13-20
Called from Dune_pkg__Sys_poll.sys_bindings in file
  "src/dune_pkg\\sys_poll.ml", line 217, characters 8-24
Called from Main__Pkg.Lock.lock in file "bin\\pkg.ml", line 462, characters
  11-82
Called from Fiber__Scheduler.exec in file "vendor/fiber/src\\scheduler.ml",
  line 76, characters 8-11

PS C:\Users\ali\dune>
```

Clearly we are using some platform specific code here.

Alizter:
The opam libs have an alternative implementation of opamStubs.ml for windows which contains `external`s to C stub. I've gone ahead and extracted only the parts we need for `Dune_pkg.Sys_poll` to work.

There was a difficult to resolve issue with one of the stubs:
```C
CAMLprim value OPAMW_GetWindowsVersion(value unit)
{
  value result;
  Field(result, 0) = Val_int(caml_win32_major);
  Field(result, 1) = Val_int(caml_win32_minor);
  Field(result, 2) = Val_int(caml_win32_build);
  Field(result, 3) = Val_int(caml_win32_revision);

  return result;
}
```
this is different to the one that can be found upstream as I have omitted the `result = caml_alloc_small(4, 0);` line. This simply caused dune to exit silently when called leading me to think that this was doing something incorrect. Not having it is a bit worrying too since I might be overwriting a random location in memory. cc @dra27 @nojb if you have any ideas what might be the problem.

In the end, this patch together with #9021 allows for `dune pkg lock` to work on Windows. I have not yet tested a package build but I don't see any reason why it wouldn't work.

Fixes #8866 

Screenshot from 2023-10-30 23-09-06


nojb:
You were missing an include (=> MSVC was assuming a wrong signature for `caml_alloc_small`). I pushed a fix directly to your branch; plus another small fix in the call to `GetModuleHandle`.

Alizter:
I managed to work out the linking issue during bootstrapping. I had called the new C stubs `opamStubs.c` however this conflicted with the existing `opamStubs` library in the CI when you have opam installed. I renamed it `custom_opamStubs.c` and it now appears to work.

rgrinberg:
@Leonidas-from-XIV could you have a look at the stubs? How do they differ from what opam is doing upstream?

Leonidas-from-XIV (approved):
@rgrinberg I took a look and the detection code looks exactly the same as in OPAM itself (so presumably correct and if anything at least matching the exact same results as OPAM), the main difference is basically how this is compiled. OPAM does some build-time dispatching in dune to determine which version of the stubs to build (build the Windows-specific stuff) whereas this uses CPP to compile them on Windows and just throw an exception in that case. I personally think this version is a bit nicer.

I wasn't sure whether the `caml_failwith` call was correct, but it does seem like it returns control to the runtime immediately, so it looks correct.

rgrinberg (changes requested):
LGTM. This needs a corresponding patch in ocaml-dune/opam and an update of the vendor script.

Alizter:
Waiting for changes to fork to be merged:
- https://github.com/ocaml-dune/opam/pull/2
- https://github.com/ocaml-dune/opam/pull/3

Alizter:
Another question this PR brings up is if we want to test pkg on Windows?

rgrinberg:
Thanks

PR #9050, opened 2023-10-31
refactor: re-implement [Appendable_list]

rgrinberg:
This is the same as https://github.com/ocaml/dune/pull/9033 without any additional book-keeping for empty checks or using the most compact constructors. Performance is looking better now.

### DIFF
--- a/otherlibs/stdune/src/appendable_list.mli
+++ b/otherlibs/stdune/src/appendable_list.mli
@@ -9,3 +9,7 @@ val ( @ ) : 'a t -> 'a t -> 'a t
 val cons : 'a -> 'a t -> 'a t
 val concat : 'a t list -> 'a t
 val to_list : 'a t -> 'a list
+val to_list_rev : 'a t -> 'a list
+val of_list : 'a list -> 'a t
+val is_empty : _ t -> bool
+val of_list_concat : 'a t list -> 'a t
--- a/otherlibs/stdune/src/appendable_list.mli
+++ b/otherlibs/stdune/src/appendable_list.mli
@@ -7,9 +7,8 @@ val empty : 'a t
 val singleton : 'a -> 'a t
 val ( @ ) : 'a t -> 'a t -> 'a t
 val cons : 'a -> 'a t -> 'a t
-val concat : 'a t list -> 'a t
 val to_list : 'a t -> 'a list
 val to_list_rev : 'a t -> 'a list
 val of_list : 'a list -> 'a t
 val is_empty : _ t -> bool
-val of_list_concat : 'a t list -> 'a t
+val concat : 'a t list -> 'a t
--- a/otherlibs/stdune/src/appendable_list.mli
+++ b/otherlibs/stdune/src/appendable_list.mli
@@ -10,5 +10,7 @@ val cons : 'a -> 'a t -> 'a t
 val to_list : 'a t -> 'a list
 val to_list_rev : 'a t -> 'a list
 val of_list : 'a list -> 'a t
-val is_empty : _ t -> bool
 val concat : 'a t list -> 'a t
+
+(** The current implementation is slow, don't use it on a hot path. *)
+val is_empty : _ t -> bool
--- a/src/dune_engine/load_rules.ml
+++ b/src/dune_engine/load_rules.ml
@@ -398,6 +398,8 @@ end = struct
             }
     in
     Alias.Name.Map.map aliases ~f:(fun { Rules.Dir_rules.Alias_spec.expansions } ->
+      (* CR-soon rgrinberg: hide this reversal behind the interface from
+         [Alias_spec] *)
       Appendable_list.to_list_rev expansions)
   ;;
 
--- a/src/dune_engine/load_rules.ml
+++ b/src/dune_engine/load_rules.ml
@@ -392,10 +392,13 @@ end = struct
           Alias.Name.Map.set
             aliases
             Alias.Name.default
-            (Rules.Dir_rules.Alias_spec.singleton
-               (Loc.none, Rules.Dir_rules.Alias_spec.Deps expansion))
+            { expansions =
+                Appendable_list.singleton
+                  (Loc.none, Rules.Dir_rules.Alias_spec.Deps expansion)
+            }
     in
-    Alias.Name.Map.map aliases ~f:Rules.Dir_rules.Alias_spec.to_list
+    Alias.Name.Map.map aliases ~f:(fun { Rules.Dir_rules.Alias_spec.expansions } ->
+      Appendable_list.to_list_rev expansions)
   ;;
 
   let add_non_fallback_rules ~init ~source_files rules =
--- a/src/dune_engine/load_rules.ml
+++ b/src/dune_engine/load_rules.ml
@@ -399,7 +399,8 @@ end = struct
     in
     Alias.Name.Map.map aliases ~f:(fun { Rules.Dir_rules.Alias_spec.expansions } ->
       (* CR-soon rgrinberg: hide this reversal behind the interface from
-         [Alias_spec] *)
+         [Alias_spec]. The order doesn't really matter, as we're just
+         collecting the dependencies that are attached to the alias *)
       Appendable_list.to_list_rev expansions)
   ;;
 
--- a/src/dune_engine/load_rules.ml
+++ b/src/dune_engine/load_rules.ml
@@ -380,8 +380,8 @@ end = struct
   ;;
 
   let compute_alias_expansions ~(collected : Rules.Dir_rules.ready) ~dir =
-    let aliases = collected.aliases in
     let+ aliases =
+      let aliases = collected.aliases in
       if Alias.Name.Map.mem aliases Alias.Name.default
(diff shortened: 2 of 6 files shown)


pmwhite (approved):

pmwhite:
Looks good to me. I'm not suggesting we put it in this PR, but I'm curious if providing built-in iteration functions for `Appendable_list.t` would be worthwhile. They would save the extra iteration and allocation that `to_list` does, but each iteration would probably be more expensive because of all the indirection in the data structure. I guess it depends on how many times we iterate over the list of aliases.

pmwhite:
Actually, it occurs to me that we can reduce allocation in `to_list_rev` a bit:
```ocaml
let to_list_rev =
  let rec loop1 acc t stack =
    match t with
    | Empty -> loop0 acc stack
    | Singleton x -> loop0 (x :: acc) stack
    | Cons (x, xs) -> loop1 (x :: acc) xs stack
    | List xs -> loop0 (List.rev_append xs acc) stack
    | Append (xs, ys) -> loop1 acc xs (ys :: stack)   
    | Concat [] -> loop0 acc stack
    | Concat (x :: xs) -> loop1 acc x (Concat xs :: stack)
  and loop0 acc stack =                                 
    match stack with
    | [] -> acc
    | t :: stack -> loop1 acc t stack
  in
  fun t -> loop1 [] t []    
;;
```

rgrinberg:
> Looks good to me. I'm not suggesting we put it in this PR, but I'm curious if providing built-in iteration functions for `Appendable_list.t` would be worthwhile. They would save the extra iteration and allocation that `to_list` does, but each iteration would probably be more expensive because of all the indirection in the data structure. I guess it depends on how many times we iterate over the list of aliases.

One issue that makes this idea less attractive is that we are specifically converting this into a list because the list is a better representation for memoization (uses less space, easy to compare/hash).

rgrinberg:
> Actually, it occurs to me that we can reduce allocation in `to_list_rev` a bit:
> 
> ```ocaml
> let to_list_rev =
>   let rec loop1 acc t stack =
>     match t with
>     | Empty -> loop0 acc stack
>     | Singleton x -> loop0 (x :: acc) stack
>     | Cons (x, xs) -> loop1 (x :: acc) xs stack
>     | List xs -> loop0 (List.rev_append xs acc) stack
>     | Append (xs, ys) -> loop1 acc xs (ys :: stack)   
>     | Concat [] -> loop0 acc stack
>     | Concat (x :: xs) -> loop1 acc x (Concat xs :: stack)
>   and loop0 acc stack =                                 
>     match stack with
>     | [] -> acc
>     | t :: stack -> loop1 acc t stack
>   in
>   fun t -> loop1 [] t []    
> ;;
> ```

Thanks. I applied this patch.

pmwhite (approved):

rgrinberg:
@pmwhite @snowleopard let me know once this is benchmarked internally.

otherlibs/stdune/src/appendable_list.mli
```diff
@@ -7,5 +7,8 @@ val empty : 'a t
 val singleton : 'a -> 'a t
 val ( @ ) : 'a t -> 'a t -> 'a t
 val cons : 'a -> 'a t -> 'a t
-val concat : 'a t list -> 'a t
 val to_list : 'a t -> 'a list
+val to_list_rev : 'a t -> 'a list
+val of_list : 'a list -> 'a t
+val is_empty : _ t -> bool
```
snowleopard, line 13:
```suggestion

(** The current implementation is slow, don't use it on a hot path. *)
val is_empty : _ t -> bool
```

src/dune_engine/rules.ml
```diff
@@ -3,15 +3,24 @@ module Action_builder = Action_builder0
...
 
 module Dir_rules = struct
-  module Alias_spec = struct
+  module Alias_spec : sig
     type item =
       | Deps of unit Action_builder.t
       | Action of Rule.Anonymous_action.t Action_builder.t
 
-    type t = { expansions : (Loc.t * item) Appendable_list.t } [@@unboxed]
+    type t = (Loc.t * item) Appendable_list.t
```
snowleopard, line 11: All these changes seem unrelated to me. Why are we doing them here? They would make it hard to benchmark just the change to the implementation of the `Appendable_list`.

snowleopard (changes requested):
I think we should disentangle the changes to the `Appendable_list` implementation from the rest of the refactoring, to make it easier to benchmark.

src/dune_engine/rules.ml
snowleopard: (Also, I'm actually changing `src/dune_engine/rules.ml` to make it faster and that moves the ground underneath, seemingly unnecessarily.)

rgrinberg: Without this change it's impossible to use `Appendable_list.concat` which reduces allocation.

rgrinberg: Wait, actually I'm not using `Appendable_list.concat anymore`. Let me see if I can get rid of it.

src/dune_engine/rules.mli
```diff
@@ -29,7 +29,10 @@ module Dir_rules : sig
         Action of
           Rule.Anonymous_action.t Action_builder.t
 
-    type t = { expansions : (Loc.t * item) Appendable_list.t } [@@unboxed]
+    type t
```
rgrinberg, line 32: @snowleopard I think we should keep this refactoring. It's not strictly necessary, but it's important to remember that we need to get the list of rules in reverse order (using `to_list_rev`) to avoid an unnecessary allocation. I think that detail should stay private to `Rules`.

snowleopard: Changing `to_list` to `to_list_rev` seems fine but the current change is quite a bit bigger.

rgrinberg: Okay, reverted as well.

src/dune_engine/rules.ml
```diff
@@ -47,9 +46,12 @@ module Dir_rules = struct
...
     let aliases =
-      Alias.Name.Map.of_list_multi aliases
-      |> Alias.Name.Map.map ~f:(fun specs ->
-        List.fold_left specs ~init:Alias_spec.empty ~f:Alias_spec.union)
+      let add_item what = function
+        | None -> Some what
+        | Some base -> Some (Alias_spec.union what base)
+      in
+      List.fold_left aliases ~init:Alias.Name.Map.empty ~f:(fun acc (name, item) ->
+        Alias.Name.Map.update acc name ~f:(add_item item))
```
snowleopard, line 54: This change seems unnecessary too?

src/dune_engine/load_rules.ml
```diff
@@ -398,7 +398,9 @@ end = struct
             }
     in
     Alias.Name.Map.map aliases ~f:(fun { Rules.Dir_rules.Alias_spec.expansions } ->
-      Appendable_list.to_list expansions)
+      (* CR-soon rgrinberg: hide this reversal behind the interface from
+         [Alias_spec] *)
+      Appendable_list.to_list_rev expansions)
```
snowleopard, line 403: Could you also add a note why using the `_rev` version is fine here?

src/dune_engine/rules.mli
snowleopard: Thanks!

src/dune_engine/rules.ml
rgrinberg: It's a performance improvement. `Alias.Name.Map.of_list_multi` does work to maintain the order (`List.rev` at the end) and this code avoids it.

snowleopard: I see. I'm just concerned that by combining multiple improvements, we are going to misunderstand the performance characteristics of the new appendable list implementation.

rgrinberg: It's possible to isolate this change, but then it's going to add a bunch of spurious changes to the tests. If you think it's worth it, I can do it.

snowleopard: Wait, are you saying this change is undoing the list reversal due to `to_list_rev`?

I didn't realise that. At minimum, this deserves a comment, similar to the one you added in `src/dune_engine/load_rules.ml`.

snowleopard: (And then maybe we could keep it indeed.)

rgrinberg: > Wait, are you saying this change is undoing the list reversal due to to_list_rev?

Yes. This is why there are no spurious changes in the tests now.

> I didn't realise that. At minimum, this deserves a comment, similar to the one you added in src/dune_engine/load_rules.ml.

Added a comment.

snowleopard: Thanks!

snowleopard (approved):
OK, I think this one is good to go, though I don't know how our internal benchmarking went.

snowleopard:
OK, I got some benchmark results for this PR applied internally: it's a consistent but small improvement (0.3-0.4% time, 0.1% memory). Let's merge it.

@pmwhite I'm assuming you're going to import it as part of your feature.

PR #9051, opened 2023-10-31
fix(pkg): add a candidates cache

rgrinberg:
This cache makes sure we never load candidates for a single package more
than once

This isn't going to solve the performance problems, but it's an optimization that will be useful regardless.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -24,6 +24,7 @@ module Context_for_dune = struct
   type 'a monad = 'a Monad.t
   type filtered_formula = OpamTypes.filtered_formula
   type filter = OpamTypes.filter
+  type rejection = Unavailable
 
   let local_package_default_version =
     OpamPackage.Version.of_string Lock_dir.Pkg_info.default_version
@@ -36,6 +37,10 @@ module Context_for_dune = struct
     ; solver_env : Solver_env.t
     ; dune_version : OpamPackage.Version.t
     ; stats_updater : Solver_stats.Updater.t
+    ; candidates_cache :
+        ( Package_name.t
+        , (OpamTypes.version * (OpamFile.OPAM.t, rejection) result) list )
+        Table.t
     }
 
   let create ~solver_env ~repos ~local_packages ~version_preference ~stats_updater =
@@ -43,10 +48,16 @@ module Context_for_dune = struct
       let major, minor = Dune_lang.Stanza.latest_version in
       OpamPackage.Version.of_string @@ sprintf "%d.%d" major minor
     in
-    { repos; version_preference; local_packages; solver_env; dune_version; stats_updater }
+    let candidates_cache = Table.create (module Package_name) 1 in
+    { repos
+    ; version_preference
+    ; local_packages
+    ; solver_env
+    ; dune_version
+    ; stats_updater
+    ; candidates_cache
+    }
   ;;
-
-  type rejection = Unavailable
 
   let pp_rejection f = function
     | Unavailable -> Format.pp_print_string f "Availability condition not satisfied"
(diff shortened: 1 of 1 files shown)


Leonidas-from-XIV (commented):
Definitely useful to cache them, but I wonder whether it makes sense to have the cache here or in the Rev_store. Retrieving the candidates from the file system is pretty fast (and possibly involves the OS caching files), so this might cache things that don't necessarily need to be cached.

rgrinberg:
The cache is quite useful for regular directories. Even if the OS cache was reliable (it's not), the cost of the syscalls itself is far greater than retrieving this information from the table.

Alizter:
Here are my benchmarks. I modified both main and this PR to avoid networking. For a small repo like `dune` there doesn't appear to be a noticeable difference. I'll try a large repo soon.

## Main using local repo

```
$ hyperfine -w 10 -r 40  "_build/install/default/bin/dune pkg lock --opam-repository-path ~/repos/opam-repository/" -c "rm -r dune.lock/"
Benchmark 1: _build/install/default/bin/dune pkg lock --opam-repository-path ~/repos/opam-repository/
  Time (mean ± σ):     331.8 ms ±   8.2 ms    [User: 290.8 ms, System: 32.9 ms]
  Range (min … max):   316.0 ms … 358.8 ms    40 runs
```

## Rudi's PR using local repo

```
$ hyperfine -w 10 -r 40  "_build/install/default/bin/dune pkg lock --opam-repository-path ~/repos/opam-repository/" -c "rm -r dune.lock/"
Benchmark 1: _build/install/default/bin/dune pkg lock --opam-repository-path ~/repos/opam-repository/
  Time (mean ± σ):     337.9 ms ±   7.8 ms    [User: 296.1 ms, System: 34.5 ms]
  Range (min … max):   322.1 ms … 356.1 ms    40 runs
```


## Main using git-repo with no networking

```
$ hyperfine -w 10 -r 40  "_build/install/default/bin/dune pkg lock --disable-networking" -c "rm -r dune.lock/"
Benchmark 1: _build/install/default/bin/dune pkg lock --disable-networking
  Time (mean ± σ):      7.241 s ±  0.067 s    [User: 10.167 s, System: 3.090 s]
  Range (min … max):    7.132 s …  7.459 s    40 runs
```

## Rudi's PR using git-repo with no remote update

```
$ hyperfine -w 10 -r 40  "_build/install/default/bin/dune pkg lock --disable-networking" -c "rm -r dune.lock/"
Benchmark 1: _build/install/default/bin/dune pkg lock --disable-networking
  Time (mean ± σ):      7.235 s ±  0.047 s    [User: 10.165 s, System: 3.085 s]
  Range (min … max):    7.140 s …  7.389 s    40 runs
```

gridbugs:
The code looks fine but I'm hesitant to merge this until we can demonstrate that it improves performance.

rgrinberg:
Note that this cache:

1. Will not increase performance in all cases. In many simple cases, all candidates are loaded exactly once anyway.
2. The cache is needed for correctness. It guarantees that we always return a consistent result for a package. This matters when the repository is a directory and we cannot control from editing it.
3. The cache will be used in another upcoming optimization I have.

For successful solves, I'm not sure the cache is being hit. But for failures, it is being hit. See this test in `lockfile-generation.t`:

```
Run the solver again. This time it will fail.
  $ dune pkg lock --opam-repository-path=mock-opam-repository
```

With the cache is enabled and you will see it get getting hit.

rgrinberg:
> The cache will be used in another upcoming optimization I have.

To expand on this, we need the candidates cache to go from the package version to the repository it came from to load the "extra-files" without guessing the repo via expensive loading operations.

Alizter (approved):
Seems ok to me.

gridbugs (approved):

PR #9052, opened 2023-10-31
fix: remove [loc] from rule digest

rgrinberg:
The location of rules is only used for error messages. Therefore, we don't need to include it when computing digests.

### DIFF
--- a/test/blackbox-tests/test-cases/rule/digest.t/run.t
+++ b/test/blackbox-tests/test-cases/rule/digest.t/run.t
@@ -48,23 +48,41 @@ Now the same but with an alias.
   running...
   digest: $2
 
-Let's add a comment to the dune file. One might think that it doesn't affect
-the rule digest, but it does because all the locations gets shifted.
+Let's add a comment to the dune file. Dune does not re-run the rule because it
+has the same digest:
+
+  $ changelocation() {
+  > sed -i.bak '1s/^/;; spurious location change\n/' dune
+  > }
+
+  $ changelocation
+
+Now we make sure that failed rules re-run when the location changes:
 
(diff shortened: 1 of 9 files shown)


ejgallego:
Will that mean that if I reformat a `dune` file that contains errorss, the errors will display the wrong position?

[That is to say, are errors in `dune` files sometime memoized?]

rgrinberg:
There will be no such issue. The key to remember is that we don't cache errors - only successes. So if your dune file contains errors, the rules derived from it will never be produced, hence never executed, hence their digests will never be calculated.

snowleopard:
It's easy to add a test to demonstrate this property, and to make sure we don't regress.

test/blackbox-tests/test-cases/rule/digest.t/run.t
```diff
@@ -66,5 +66,3 @@ make them more resilient to non semantic changes.
 # need to update it when rule digest version changes.
 
   $ dune build @default
-  running...
```
snowleopard, line 69: Please remove the CR-someday above.

snowleopard: Actually, it's not just the CR-somday that needs updating after your change. Please update the text above too.

snowleopard (approved):
Looks but please update the text of the test which demonstrates the improvement. You could also add a new test there to demonstrate that failing actions do get correctly rerun and produce errors with correct locations.

rgrinberg:
Updated the test and added a new one.

snowleopard:
> Updated the test and added a new one.

Awesome, thanks!

PR #9056, opened 2023-10-31
fix: check if existing dir is actually a dir

Alizter:
Resurrecting #6355. Now that we have better benchmarks, let's check the performance impact if any.

I've also gone ahead and improved the tests for `Fpath` from that PR and addressed some remaining comments.

### DIFF
--- a/bin/install_uninstall.ml
+++ b/bin/install_uninstall.ml
@@ -374,17 +374,10 @@ module File_ops_real (W : sig
   ;;
 
   let mkdir_p p =
-    (* CR-someday amokhov: We should really change [Path.mkdir_p dir] to fail if
-       it turns out that [dir] exists and is not a directory. Even better, make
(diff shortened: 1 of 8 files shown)


Alizter:
OK, the monorepo benchmark has shown us that there is a time regression in the null build. There appear to be no regressions in the smaller benchmark however that are visible. The issue stems from the part that Andrey pointed out where we do an extra syscall when `Unix.mkdir` fails in order to determine if it is a directory. We'll need to think a bit harder how we can absorb that overhead elsewhere.

Alizter:
One idea I've had to speed this up now that we do an extra syscall is to start caching some of the mkdirs. This might help the speed but obviously memory usage will increase.

I've haven't been able to think of another way to remove the extra syscall without really breaking the correctness.

I've also tried to have a go at writing my own C stubs implementing mkdir_p but I got a headache trying to link to OCaml's unix library in a cross platform way.

For now this appears to be stuck.

Issue #1645, opened 2018-12-12
Odoc: valid module name clashes cause issues


jonludlam:

diml:
Hmm yh, I believe it's because odoc works with packages rather than libraries. I don't know how we can fix this in Dune :/

jonludlam:
I've updated the description to make it clear that this is for sub-libraries (repro.one and repro.two)

jonludlam:
This is related to odoc#148

avsm:
I'm still seeing this in ocamllabs/ocaml-ctypes#588 with odoc 1.4.0, need to investigate more:


```
 $ dune build @doc
Error: Conflicting implementations for virtual library "ctypes-foreign":
- "ctypes-foreign.threaded" in _build/default/src/ctypes-foreign-threaded
- "ctypes-foreign.unthreaded" in _build/default/src/ctypes-foreign-unthreaded
Multiple rules generated for _build/default/_doc/_html/ctypes-foreign/Foreign/index.html:
- <internal location>
- <internal location>
```

rgrinberg:
@jonludlam is there a corresponding odoc bug? I don't think this can be fixed until odoc addresses this.

jonludlam:
@avsm which version of dune are you using? This ought to be addressed by https://github.com/ocaml/dune/pull/2141

@rgrinberg there's no issue on the odoc repo yet. I'll make one when I've got a moment.

avsm:
@jonludlam twas with the released dune 1.9.3 -- I'll try with the master branch shortly

davesnx:
Facing this issue and found https://github.com/ocaml/odoc/issues/148 has been resolved. Is there any chance to relook into this?

Khady:
Is it correct to say that this is (at least partially) fixed by the `@doc-new` rule?

emillon:
ping @jonludlam for this question
(though I wouldn't close issues fixed by doc-new until we know how these new rules are going to be enabled)

jonludlam:
Yes, `doc-new` should not have this problem. See the test cases for [doc rules](https://github.com/ocaml/dune/blob/main/test/blackbox-tests/test-cases/odoc/github1645.t) and [doc-new rules](https://github.com/ocaml/dune/blob/main/test/blackbox-tests/test-cases/odoc/new/github1645.t)

PR #9057, opened 2023-10-31
Collect rules with Appendable_list.t instead of Id.Map.t

pmwhite:
Based on #9050, the main goal is to replace the unioning of a bunch of maps with what I expect to be a cheaper operation. I've benchmarked a similar patch on Jane Street's code base and saw enough improvement to think that this is a worthwhile refactor. My thinking was to have review done first here on GitHub; once it is able to be merged I'll import it, run benchmark it again and report back with a summary of the results.

### DIFF
--- a/otherlibs/stdune/src/appendable_list.ml
+++ b/otherlibs/stdune/src/appendable_list.ml
@@ -1,13 +1,67 @@
-type 'a t = 'a list -> 'a list
+type 'a t =
+  | Empty
+  | Singleton of 'a
+  | Cons of 'a * 'a t
+  | List of 'a * 'a * 'a list
+  | Append of 'a t * 'a t
+  | Concat of 'a t list
 
-let empty k = k
-let singleton x k = x :: k
-let to_list l = l []
-let ( @ ) a b k = a (b k)
-let cons x xs = singleton x @ xs
+let empty = Empty
+let singleton x = Singleton x
 
-let rec concat l k =
-  match l with
-  | [] -> k
-  | t :: l -> t (concat l k)
+let ( @ ) a b =
+  match a, b with
+  | Empty, _ -> b
+  | _, Empty -> a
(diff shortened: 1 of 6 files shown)


pmwhite:
It's probably not worth reviewing until #9050 gets merged, so that the diff in this PR isn't as large. I wish GitHub knew how to handle PR chains.

pmwhite:
Converting to draft because I was silly and forgot to run tests, which are broken.

Alizter:
The bench shows some very promising results!


Metric | Last PR value | Last main value | Delta
-- | -- | -- | --
[Clean Sandbox] Build Time | 299.965 | 316.930 | -5.35289%
[Clean] Build Time | 298.799 | 310.847 | -3.87600%
[Null Sandbox] Build Time | 3.58720 | 3.63885 | -1.41939%
[Null] Build Time | 3.57499 | 3.66382 | -2.42454%

Alizter:
However it seems there are times when the union is eliminating duplicates:
```
/odoc-wrapped-lib.t/run.t
   $ dune build @doc
+  Error: Multiple rules generated for
+  _build/default/_doc/_html/l/index.html.json:
+  - <internal location>
+  - <internal location>
+  -> required by alias doc
+  [1]
```
I wonder if these can be resolved by not duplicating them to begin with.;

pmwhite:
@Alizter Awesome, thanks for the benchmark results! Regarding the test failures, some of them are just re-orderings of rules, which is expected.

Other failures, like the one you highlighted, replace one error message with another. I suspect that the duplicate rules error message isn't new, but has just been hidden by the first error message - the re-ordering of rules changed which one actually gets displayed. Notice that the error you pointed out is very similar to #1645 , which is yet to be fixed.

rgrinberg:
@pmwhite please re-open once you make some progress on this

PR #9058, opened 2023-10-31
refactor: drop [~distribute] in [is_useful_to]

rgrinberg:

### DIFF
--- a/src/dune_engine/action.ml
+++ b/src/dune_engine/action.ml
@@ -287,7 +287,7 @@ type is_useful =
   | Clearly_not
   | Maybe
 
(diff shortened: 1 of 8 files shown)


snowleopard (approved):
Thanks!

Alizter:
Perhaps not for this PR, but the function `is_useful_to ~memoize` is kind of confusing. We could probably replace the boolean value with
```
[ `Memoize | `Distribute ]
```
at no extra cost.

snowleopard:
> Perhaps not for this PR, but the function `is_useful_to ~memoize` is kind of confusing.

Agreed. I think we should just split it in two separate functions.

PR #9059, opened 2023-11-01
refactor(pkg): simplify extra file loading

rgrinberg:
Instead of repeating the lookup in the repositories, we take advantage
of the fact that we know where every version is coming from.

### DIFF
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -311,10 +311,13 @@ let load_all_versions ts opam_package_name =
     if OpamPackage.Version.Map.mem version acc
     then acc
     else OpamPackage.Version.Map.add version (repo, package) acc)
-  |> OpamPackage.Version.Map.values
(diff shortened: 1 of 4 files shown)


src/dune_pkg/opam_repo.ml
```diff
@@ -360,10 +360,13 @@ let load_all_versions ts opam_package_name =
     if OpamPackage.Version.Map.mem version acc
     then acc
     else OpamPackage.Version.Map.add version (repo, package) acc)
-  >>| OpamPackage.Version.Map.values
-  >>= Fiber.parallel_map ~f:(fun (repo, pkg) -> load_opam_package repo pkg)
+  >>| OpamPackage.Version.Map.to_seq
```
Alizter, line 363: Could you explain why you are using seq here?

rgrinberg: Because there's no `Map.to_list` function.

rgrinberg:
Merging this PR as it's blocking other work and is obviously good. Once we resolve a package, it's obviously unnecessary to look it up again just to check its repository.

Alizter (approved):

PR #9060, opened 2023-11-01
pkg: use new type instead of string for versions

gridbugs:
This adds the `Package_version.t` type which represents the type of a version of a package. This replaces the use of `string` for this purpose.


Alizter:
Wouldn't it be better for `Package_version` to live in `Dune_lang` since its not just package specific?

rgrinberg (approved):

PR #9061, opened 2023-11-01
pkg: add sort to test for deterministic output

gridbugs:


Alizter (approved):

Leonidas-from-XIV (approved):
This is the fix I suggested in #9044 :-)

(Also sorry that it was non-reproducible)

PR #9062, opened 2023-11-01
pkg: add quotes to command in error message

gridbugs:
Without the quotes it's harder than it needs to be to tell which part of the sentence is the command it's hinting you to run.

### DIFF
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -564,7 +564,7 @@ struct
     match Path.exists (Path.source lock_dir_path) with
     | false ->
       User_error.raise
-        ~hints:[ Pp.text "Run dune pkg lock to generate it." ]
+        ~hints:[ Pp.text "Run `dune pkg lock` to generate it." ]
         [ Pp.textf "%s does not exist." (Path.Source.to_string lock_dir_path) ]
     | true ->
       (match Path.is_directory (Path.source lock_dir_path) with


src/dune_pkg/lock_dir.ml
```diff
@@ -564,7 +564,7 @@ struct
     match Path.exists (Path.source lock_dir_path) with
     | false ->
       User_error.raise
-        ~hints:[ Pp.text "Run dune pkg lock to generate it." ]
+        ~hints:[ Pp.text "Run `dune pkg lock` to generate it." ]
```
Alizter, line 567: to be more consistent with other messages you can do a `Pp.textf` and use `%S` which will double quote the string and then put `"dune pkg lock"` as an arg.

Leonidas-from-XIV: I just did a `rg hints` to get a rough overview of what we use as a convention and the convention is that there is no convention. Sometimes we state commands with backticks (`coqc --config`), sometimes we specify a (variable) program with `%S`, sometimes we just state to install fswatch without any quotes (or dune subst, or dune build).

But I don't think `%S` is really the way to go, since that quotes string according to OCaml conventions which is not really what we need or want here. In the case where we do so it makes a bit more sense, since the command is filled in and not a constant, but even there it's a little bit abuse of `%S`.

Personal opinion: I like backticks. Maybe we should have them everywhere.

Leonidas-from-XIV (approved):

src/dune_pkg/lock_dir.ml
rgrinberg: Why should we have them? They're awkward for copy pasting and there's no markdown formatting here.

gridbugs: They're no more awkward than any other type of quote and they're the established convention for denoting commands in many places.

PR #9063, opened 2023-11-01
fix(pkg): Avoid duplicate `ls-tree` calls

Leonidas-from-XIV:
For a particular revision the output of `ls-tree` will never change, so we can just get the file list when the `At_rev.t` is created.

### DIFF
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -208,7 +208,7 @@ let get_opam_package_files t opam_package =
         [ "packages"; name; OpamPackage.to_string opam_package; "files" ]
     in
     Rev_store.Remote.At_rev.directory_entries at_rev files_root
-    >>= Fiber.parallel_map ~f:(fun remote_file ->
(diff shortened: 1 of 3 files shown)


Alizter:
Testing the locking of `patdiff` with no networking before and after this patch. 

# Before
```
Benchmark 1: ~/dune/_build/install/default/bin/dune pkg lock --disable-networking
  Time (mean ± σ):     32.887 s ±  0.182 s    [User: 40.304 s, System: 13.059 s]
  Range (min … max):   32.601 s … 33.124 s    10 runs
```
# After
``
Benchmark 1: ~/dune/_build/install/default/bin/dune pkg lock --disable-networking
  Time (mean ± σ):      6.792 s ±  0.069 s    [User: 14.942 s, System: 8.382 s]
  Range (min … max):    6.716 s …  6.923 s    10 runs
```

rgrinberg:
Thanks. I converted the list to a set. This simplifies the implementation somewhat and also avoids equality having that slow sorting step.

PR #9065, opened 2023-11-01
refactor(pkg): use a single git call to load opam files

rgrinberg:
Rather than calling $ git show for every file, we introduce an API that
allows us to fetch mutliple files in a single command.

### DIFF
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -3,6 +3,22 @@ open Fiber.O
 
 let ( / ) = Path.relative
 
+let rev_store =
+  let store = ref None in
+  Fiber.of_thunk (fun () ->
+    match !store with
+    | Some s -> Fiber.return s
+    | None ->
+      let dir =
+        Path.L.relative
+          (Path.of_string (Xdg.cache_dir (Lazy.force Dune_util.xdg)))
+          [ "dune"; "git-repo" ]
+      in
+      let+ rev_store = Rev_store.load_or_create ~dir in
+      store := Some rev_store;
+      rev_store)
+;;
+
 module Serializable = struct
   type t =
     { repo_id : Repository_id.t option
@@ -97,18 +113,10 @@ let of_opam_repo_dir_path ~source ~repo_id opam_repo_dir_path =
   { source = Directory packages_dir_path; serializable }
 ;;
 
-let xdg_repo_location =
-  let ( / ) = Filename.concat in
-  lazy (Xdg.cache_dir (Lazy.force Dune_util.xdg) / "dune/git-repo" |> Path.of_string)
-;;
-
 let of_git_repo ~repo_id ~source =
   let+ at_rev, computed_repo_id =
     let* remote =
-      let* repo =
-        let dir = Lazy.force xdg_repo_location in
-        Rev_store.load_or_create ~dir
-      in
+      let* repo = rev_store in
       Rev_store.add_repo repo ~source
     in
     match repo_id with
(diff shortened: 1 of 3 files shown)


src/dune_pkg/opam_repo.ml
```diff
@@ -265,59 +274,110 @@ let load_opam_package t opam_package =
...
+  List.map2 opam_packages contents ~f:(fun file opam_file_contents ->
+    let path = Rev_store.File.path file in
+    let opam_file =
+      let filename =
+        (* the filename is used to read the version number *)
+        Path.Local.to_string path |> OpamFilename.of_string |> OpamFile.make
+      in
+      OpamFile.OPAM.read_from_string ~filename opam_file_contents
+    in
+    (* TODO the [file] here is made up *)
```
Leonidas-from-XIV, line 288: I don't quite understand this TODO. What is the action here? Probably not to create the file, since we deliberately don't do that.

src/dune_pkg/rev_store.ml
```diff
@@ -37,10 +37,15 @@ let run_capture_zero_separated_lines { dir } =
...
-let show { dir } (Rev rev) path =
+let show { dir } revs_and_paths =
   let git = Lazy.force Vcs.git in
   let failure_mode = Vcs.git_accept () in
-  let command = [ "show"; sprintf "%s:%s" rev (Path.Local.to_string path) ] in
+  let command =
+    "show"
+    :: List.map revs_and_paths ~f:(function
+      | `Object o -> o
+      | `Path (Rev r, path) -> sprintf "%s:%s" r (Path.Local.to_string path))
```
Leonidas-from-XIV, line 47: I worry that this probably needs to run `git show` in batches since the command line length on some platforms is [8191 characters](https://learn.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/command-line-string-limitation), on others it is a more generous 2097152 characters.

src/dune_pkg/opam_repo.ml
```diff
@@ -265,59 +274,110 @@ let load_opam_package t opam_package =
...
-    then acc
-    else OpamPackage.Version.Map.add version (repo, package) acc)
-  |> OpamPackage.Version.Map.to_seq
-  |> List.of_seq
-  |> Fiber.parallel_map ~f:(fun (version, (repo, pkg)) ->
-    load_opam_package repo pkg
-    >>| Option.map ~f:(fun (pkg : With_file.t) -> version, (repo, pkg)))
-  >>| List.filter_opt
-  >>| OpamPackage.Version.Map.of_list
+  let from_git, from_dirs =
```
Leonidas-from-XIV, line 343: I do not understand what the reason is that the `all_package_versions` are to be distinguished between git and directory and require to be handled differently in `load_all_versions`. Why this particular choice?

Leonidas-from-XIV (commented):
Seems like a good thing but we probably need to make sure to implement batching for platforms with low character limits for process calls. But I believe this can be fitted in without changing the API.

src/dune_pkg/opam_repo.ml
rgrinberg: Because fetching from git can be batched into a single process execution, while fetching from directories cannot. This code is just preparing the batch.

src/dune_pkg/rev_store.ml
rgrinberg: Ah yeah, that's annoying. I've added split batches to mitigate this.

src/dune_pkg/opam_repo.ml
rgrinberg: The file is used as a location in case of an error. Unfortunately, this file is totally bogus since it only exists as a git object. I'm not sure how to write a proper location here yet unfortunately.

Leonidas-from-XIV: If this location is only for error messages maybe we should potentially have our own error type here with its own `pp` function that could tell the user whether it is an path in their filesystem or a path in some revision of a repo? 

Anyway, this comment was not introduced in this PR nor has the fake file been introduced here so this is just a drive-by remark and can be improved at a later point in time.

Leonidas-from-XIV (approved):
Yeah, the batching looks like a fairly solid solution now.

PR #9067, opened 2023-11-02
pkg: remove dev field from lockfile

gridbugs:
In opam the "dev" package variable is true for packages that weren't build from a release archive. Currently we only have support for building released packages so removing it to simplify the code.


rgrinberg:
I guess this will have to be resurrected shortly because we need to get pinning to work. I'd lean towards keeping it. Pinning is a key feature that's at the top of the todo list.

gridbugs:
Makes sense. Closing.

PR #9068, opened 2023-11-02
pkg: check version constraints for lockfile deps

gridbugs:
Identifying lockfile deps now works by searching for a set of dependencies that both satisfies the dependency formula of the package and is also satisfied by the package versions present in the lockdir. In the process some common operations on opam formulae is moved into a separate module. This is in preparation for a later change that will depend on some of these operations.

See the individual commits for more detail.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -22,7 +22,6 @@ end
 
 module Context_for_dune = struct
   type 'a monad = 'a Monad.t
-  type filtered_formula = OpamTypes.filtered_formula
   type filter = OpamTypes.filter
   type rejection = Unavailable
 
@@ -84,24 +83,6 @@ module Context_for_dune = struct
     | Invalid_argument msg -> Error (`Not_a_bool msg)
   ;;
 
-  (* Substitute variables with their values *)
-  let resolve_solver_env
-    (solver_env : Solver_env.t)
-    (stats_updater : Solver_stats.Updater.t)
-    : OpamTypes.filter -> OpamTypes.filter
-    =
-    OpamFilter.map_up (function
-      | FIdent ([], variable, None) as filter ->
-        (match Solver_env.Variable.of_string_opt (OpamVariable.to_string variable) with
-         | None -> filter
-         | Some variable ->
-           Solver_stats.Updater.expand_variable stats_updater variable;
-           (match Solver_env.get solver_env variable with
-            | Unset_sys -> filter
-            | String string -> FString string))
-      | other -> other)
-  ;;
-
   let is_opam_available =
     (* The solver can call this function several times on the same package. If
        the package contains an invalid `available` filter we want to print a
@@ -112,7 +93,10 @@ module Context_for_dune = struct
       let available = OpamFile.OPAM.available opam in
       match
         let available_vars_resolved =
-          resolve_solver_env t.solver_env t.stats_updater available
+          Resolve_opam_formula.substitute_variables_in_filter
+            ~stats_updater:(Some t.stats_updater)
+            t.solver_env
+            available
         in
         eval_to_bool available_vars_resolved
       with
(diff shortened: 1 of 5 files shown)


test/blackbox-tests/test-cases/pkg/lockfile-deps-respect-constraints.t
```diff
@@ -0,0 +1,45 @@
...
+  $ mkpkg a 0.0.2
+  $ mkpkg b
+
+The package "c" can be satisfied by b or by a.0.0.1.
+  $ mkpkg c <<EOF
+  > depends: [ "b" | "a" {= "0.0.1"} ]
+  > EOF
+
+The package "d" can only be satisfied by "a.0.0.2" which will force the package
+"c" to depend on "b" rather than "a" and ensure "a.0.0.2" is in the solution
```
rgrinberg, line 18: Note that `|` isn't exclusive in the constraint language. If you want to force `b` instead of `a`, you need an appropriate conflict between the two packages.

src/dune_pkg/opam_solver.ml
```diff
@@ -518,15 +492,22 @@ let opam_package_to_lock_file_pkg
...
+      Resolve_opam_formula.filtered_formula_to_package_names
+        ~with_test:false
+        solver_env
+        version_by_package_name
+        opam_file.depends
+    with
+    | Ok dep_package_names ->
+      List.map dep_package_names ~f:(fun package_name -> Loc.none, package_name)
+    | Error (`Formula_could_not_be_satisfied hints) ->
+      Code_error.raise
```
rgrinberg, line 506: Isn't this possible? why is it a code error?

src/dune_pkg/resolve_opam_formula.mli
```diff
@@ -0,0 +1,63 @@
...
+(** Update a filter by substituting variables from a [Solver_env.t] *)
+val substitute_variables_in_filter
+  :  ?stats_updater:Solver_stats.Updater.t
+  -> Solver_env.t
+  -> OpamTypes.filter
+  -> OpamTypes.filter
+
+(** Evaluate the filters in a filtered formula returning the resulting formula. *)
+val filtered_formula_to_formula
+  :  ?stats_updater:Solver_stats.Updater.t
```
rgrinberg, line 12: Unless, there's a specific need we try to avoid optional arguments. They introduce bugs, and their main utility which is improving an API in a backwards compatible way isn't relevant to our internal code.

src/dune_pkg/resolve_opam_formula.mli
```diff
@@ -0,0 +1,63 @@
...
+
+(** Update a filter by substituting variables from a [Solver_env.t] *)
+val substitute_variables_in_filter
+  :  ?stats_updater:Solver_stats.Updater.t
+  -> Solver_env.t
+  -> OpamTypes.filter
+  -> OpamTypes.filter
+
+(** Evaluate the filters in a filtered formula returning the resulting formula. *)
+val filtered_formula_to_formula
```
rgrinberg, line 11: This name is a bit awkward to read. What about just `apply_filter`?

test/blackbox-tests/test-cases/pkg/lockfile-deps-respect-constraints.t
gridbugs: Here I force `b` by having a second package (`d`) which depends on an incompatible version of `a`. The only way to solve for both `b` and `c` is for d to depend on `a.0.0.2` which leaves `b` as the only possible way for `a` to be satisfied.

src/dune_pkg/opam_solver.ml
gridbugs: It shouldn't be possible. This runs immediately following the solver and the solution should contain packages that satisfy the dependency formulae of each package in the solution.

rgrinberg (approved):

PR #9074, opened 2023-11-02
Improve [Action_builder.memoize]

snowleopard:
Improve `Action_builder.memoize`:

* Remove one of the `Lazy.t` cells. Most `Eager` parts of `Action_builder.t`s end up being forced in every build, so the `Lazy.t` wrapper is unnecessary.

* Add the "(lazy)" suffix to the names of `Lazy` parts to be able to distinguish `Eager` and `Lazy` parts in profiling.

Our internal benchmarks show a consistent improvement across all scenarios, both in terms of time (1-2%) and memory allocations (a tiny improvement but still: 0-0.2%).

Note that I also benchmarked a similar change where I also remove the `Lazy.t` cell in the `Lazy` part of `Action_builder.t`s. That was a clear regression both in terms of time and memory. 

@rgrinberg I'm assuming this is a good change externally as well. We use `Action_builder.run Eager` in all `Action_builder.run` call sites in `build_system.ml`: `execute_rule_impl`, `build_alias_impl` and `dep_on_anonymous_action`, which is the main place where `Action_builder`s are evaluated. One notable (and the only, internally) user of `Action_builder.run Lazy` is `reflection.ml`, and I don't think we should worry about slightly increasing allocations here (because we usually combine reflection checks with building the actual targets). There is only one other call site with `Lazy` externally (`src/dune_rules/lib_file_deps.ml`) and I think it contributes much less to the overall memory usage compared to the big stuff in `build_system.ml`. Please let me know if you agree. If not, we can make this change internal-only.

### DIFF
--- a/src/dune_engine/action_builder0.ml
+++ b/src/dune_engine/action_builder0.ml
@@ -130,24 +130,31 @@ let force_lazy_or_eager
   : type a b.
     a eval_mode
     -> (b * Dep.Set.t) Memo.Lazy.t Lazy.t
-    -> (b * Dep.Facts.t) Memo.Lazy.t Lazy.t
+    -> (b * Dep.Facts.t) Memo.Lazy.t
     -> (b * a Dep.Map.t) Memo.t
   =
   fun mode lazy_ eager ->
   match mode with
   | Lazy -> Memo.Lazy.force (Lazy.force lazy_)
-  | Eager -> Memo.Lazy.force (Lazy.force eager)
+  | Eager -> Memo.Lazy.force eager
 ;;
 
 let memoize ?cutoff name t =
-  let cutoff ~equal =
-    Option.map cutoff ~f:(fun eq -> Tuple.T2.equal eq (Dep.Map.equal ~equal))
+  let lazy_ : ('a * Dep.Set.t) Memo.Lazy.t Lazy.t =
+    lazy
+      (let cutoff =
+         Option.map cutoff ~f:(fun equal -> Tuple.T2.equal equal Dep.Set.equal)
+       in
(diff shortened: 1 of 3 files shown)


rgrinberg (approved):
The benchmark results are here: https://autumn.ocamllabs.io/ocaml/dune/pull/9074/base/main?worker=autumn&image=bench.Dockerfile

The results are a bit of a wash. No real difference in build time and there's more allocation.

On the public side, little performance tweaks like this aren't very impactful and keeping up with upstream seems more important.

src/dune_engine/action_builder0.ml
```diff
@@ -130,24 +130,31 @@ let force_lazy_or_eager
...
+       in
+       Memo.lazy_ ?cutoff ~name:(name ^ "(lazy)") (fun () -> t.f Lazy))
   in
-  let memo ~equal eval_mode =
-    Memo.lazy_ ?cutoff:(cutoff ~equal) ~name (fun () -> t.f eval_mode)
+  (* Unlike [lazy_], [eager] doesn't have the outer [Lazy.t] wrapper because most [Eager]
+     nodes end up getting forced during every build. *)
+  let eager : ('a * Dep.Facts.t) Memo.Lazy.t =
+    let cutoff =
+      Option.map cutoff ~f:(fun equal -> Tuple.T2.equal equal Dep.Facts.equal)
```
rgrinberg, line 154: Not sure if this is worth doing in this PR, but this closure (`~f`) can be hoisted up to reduce an allocation.

snowleopard: I was under impression that such functions are floated out automatically but let me give that a try. Btw, by "up" do you mean one level or two levels?

PR #9075, opened 2023-11-03
Add test that creates a local switch

gridbugs:


.github/workflows/workflow.yml
```diff
@@ -247,3 +247,23 @@ jobs:
           push_image_and_stages: on:push
           username: ocamldune
           password: "${{ secrets.DOCKER_HUB_PASSWORD }}"
+
+  create-local-opam-switch:
+    name: Create local opam switch
+    strategy:
+      fail-fast: false
```
rgrinberg, line 254: why not `true`? AFAIK, this is to stop running jobs after new pushes?

gridbugs: changed it to `true`

rgrinberg (approved):

PR #9078, opened 2023-11-03
fix: dune-package parsing bug

rgrinberg:
(dir ..) wasn't being parsed correctly

### DIFF
--- a/src/dune_rules/dune_package.ml
+++ b/src/dune_rules/dune_package.ml
@@ -361,15 +361,17 @@ type path = [ `File | `Dir ] * Install.Entry.Dst.t
 
 let decode_path =
   let open Dune_lang.Decoder in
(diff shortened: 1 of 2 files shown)


doc/changes/9078.md
```diff
@@ -0,0 +1,2 @@
+- Fix incorrect parsing of `dune-package` files in the presence of installed
```
emillon, line 1: Since the feature is not part of a released version I think that this PR can just amend `doc/changes/package-dirs.md`

emillon (approved):
Thanks. I got a bit too fast when refactoring the feature.

doc/changes/9078.md
rgrinberg: I just got rid of this. If it wasn't released, no need to mention it.

PR #9079, opened 2023-11-03
chore: Use `User_message.command` to denote command line items

Leonidas-from-XIV:
Follow up to #9064. I did some searching to find all the occurrences but due to the nature of it, I might have missed out some places.


bin/ocaml/ocaml_merlin.ml
```diff
@@ -99,7 +99,7 @@ end = struct
     | None ->
       Path.Build.drop_build_context_exn file
       |> Path.Source.to_string_maybe_quoted
-      |> Printf.sprintf "No config found for file %s. Try calling `dune build`."
+      |> Printf.sprintf "No config found for file %s. Try calling 'dune build'."
```
gridbugs, line 102: Why not use `User_message.command` here?

src/dune_rules/merlin/merlin.ml
```diff
@@ -130,7 +130,7 @@ module Processed = struct
       Error
         "The current Merlin configuration has been generated by another, incompatible, \
          version of Dune. Please rebuild the project. (Using the same version of Dune as \
-         the one running the `ocaml-merlin` server.)"
+         the one running the 'ocaml-merlin' server.)"
```
gridbugs, line 133: `User_message.command`?

rgrinberg (approved):

bin/ocaml/ocaml_merlin.ml
Leonidas-from-XIV: Because it gets used as input to `Merlin_conf.make_error` which does not get raised as `User_error.raise` does, but instead serialized to an S-Expression (and probably transmitted somewhere, and I didn't want to do the research of what breaks if we send formatted data there).

src/dune_rules/merlin/merlin.ml
Leonidas-from-XIV: Same as above, this doesn't get printed to the terminal but serialized into an S-Expression. Presumably something is printing this somewhere but hardcoding this will make the output of dune a bit more consistent.

Given this doesn't prevent a better solution in the future, I don't think it is worth overthinking this too much in this PR.

PR #9090, opened 2023-11-06
refactor(pkg): split remote from at_rev

rgrinberg:
revisions are indepedent from remotes

### DIFF
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -69,12 +69,12 @@ end
 module Backend = struct
   type t =
     | Directory of Path.t
-    | Repo of Rev_store.Remote.At_rev.t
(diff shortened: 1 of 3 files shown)


Leonidas-from-XIV (approved):
They are, but I found it easier to have a "Git repo -> Remote -> Files at a revision of the remote" abstraction, as that matches more what my mental model is when interacting with git repos. The fact that revisions and remotes are unrelated wasn't even clear to me until I started implementing the storage.

rgrinberg:
Indeed. But it turns out to be useful to handle revisions in bulk while ignoring which remote they're originating from. This allows implementing an upcoming optimization in a simpler way.

Leonidas-from-XIV:
> But it turns out to be useful to handle revisions in bulk while ignoring which remote they're originating from. 

Ok, yes, that's a good point!

PR #9092, opened 2023-11-06
refactor(pkg): move all paths to a module

rgrinberg:
Also, I made the `Directory` constructor just take the opam repository root rather than the `packages` subdirectories. That's consistent with how we treat git repos and we can use the same functions for paths everywhere.


Leonidas-from-XIV (approved):
Making the path calculation in one place makes for more readable code :+1:

PR #9095, opened 2023-11-06
pkg: detect test-only dependencies in lockdir

gridbugs:
Adds a new command `dune pkg list-locked-dependencies` which lists the locked dependencies of local packages, identifying which dependencies will only be included if tests are enabled.

### DIFF
--- a/src/dune_pkg/package_universe.mli
+++ b/src/dune_pkg/package_universe.mli
@@ -3,12 +3,22 @@ open! Import
 (** All of the packages in a dune project including local packages and packages
     in a lockdir. The lockdir is guaranteed to contain a valid dependency
     solution for the local packages. *)
-type t = private
-  { local_packages : Local_package.t Package_name.Map.t
-  ; lock_dir : Lock_dir.t
-  }
+type t
 
 val create
   :  Local_package.t Package_name.Map.t
   -> Lock_dir.t
   -> (t, User_message.t) result
+
+(** Returns the dependencies of the specified package within teh package
+    universe *)
+val opam_package_dependencies_of_package
+  :  t
+  -> Package_name.t
+  -> which:[ `All | `Non_test | `Test_only ]
+  -> traverse:[ `Immediate | `Transitive ]
+  -> OpamPackage.t list
+
+(** Returns the opam package whose name is the given package name and whose
+    version is the version of that package within the package universe *)
+val opam_package_of_package : t -> Package_name.t -> OpamPackage.t
--- a/bin/describe/describe_pkg.ml
+++ b/bin/describe/describe_pkg.ml
@@ -75,8 +75,135 @@ module Dependency_hash = struct
   let command = Cmd.v info term
 end
 
+module List_locked_dependencies = struct
+  module Package_universe = Dune_pkg.Package_universe
+  module Lock_dir = Dune_pkg.Lock_dir
+  module Opam_repo = Dune_pkg.Opam_repo
+  module Package_version = Dune_pkg.Package_version
+  module Opam_solver = Dune_pkg.Opam_solver
+
+  let info =
+    let doc = "List the dependencies locked by a lockdir" in
+    let man = [ `S "DESCRIPTION"; `P "List the dependencies locked by a lockdir" ] in
+    Cmd.info "list-locked-dependencies" ~doc ~man
+  ;;
+
+  let package_deps_in_lock_dir_pp package_universe package_name ~transitive =
+    let traverse, traverse_word =
+      if transitive then `Transitive, "Transitive" else `Immediate, "Immediate"
+    in
+    let opam_package =
+      Package_universe.opam_package_of_package package_universe package_name
+    in
+    let list_dependencies which =
+      Package_universe.opam_package_dependencies_of_package
+        package_universe
+        package_name
+        ~which
+        ~traverse
+    in
+    Pp.concat
+      ~sep:Pp.cut
+      [ Pp.hbox
+          (Pp.textf
+             "%s dependencies of local package %s"
+             traverse_word
+             (OpamPackage.to_string opam_package))
+      ; Pp.enumerate (list_dependencies `Non_test) ~f:(fun opam_package ->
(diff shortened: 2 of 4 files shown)


Alizter:
What's the difference between this and `dune describe pkg lock`?

gridbugs:
> What's the difference between this and `dune describe pkg lock`?

Will update this next week to just add "test-only" annotations to the output of `dune describe pkg lock` rather than adding a new command. I forgot we had that.

Alizter:
It might be helpful to include the command "describe pkg lock "both in the pkg and describe command group. That way users won't need to remember which one it is in.

rgrinberg:
@gridbugs please undraft once this is updated

gridbugs:
This PR will be greatly simplified by https://github.com/ocaml/dune/pull/9172 which consolidates all the lockdir validation. I'll rebase on top of that and undraft once https://github.com/ocaml/dune/pull/9172 is merged.

gridbugs:
I've updated this to work with the changes introduced in https://github.com/ocaml/dune/pull/9172. This is now ready for review.

gridbugs:
@Alizter I looked at `describe pkg lock` but that prints entire lockdirs whereas the new command in this change just prints the dependencies of individual packages according to the lockdirs. The distinction is important because a package may be a test-only dependency of one package but a non-test dependency of another.

bin/pkg/list_locked_dependencies.ml
```diff
@@ -0,0 +1,113 @@
...
+  in
+  let list_dependencies which =
+    Package_universe.opam_package_dependencies_of_package
+      package_universe
+      package_name
+      ~which
+      ~traverse
+  in
+  Pp.concat
+    ~sep:Pp.newline
```
rgrinberg, line 30: This should be `Pp.cut`. We should almost never use `Pp.newline` because it breaks that are already open.

bin/pkg/list_locked_dependencies.ml
```diff
@@ -0,0 +1,113 @@
...
+    [ Pp.verbatim
```
rgrinberg, line 31: How come it's not `Pp.text`?

src/dune_pkg/package_universe.mli
```diff
@@ -3,12 +3,26 @@ open! Import
...
+type t
 
 val create
   :  Local_package.t Package_name.Map.t
   -> Lock_dir.t
   -> (t, User_message.t) result
+
+(** Raises a [User_error.E] if the lockdir doesn't contain a solution to the
+    local packages *)
+val create_exn : Local_package.t Package_name.Map.t -> Lock_dir.t -> t
```
rgrinberg, line 15: I see this function is only used once. So I would just inline it since it's just as well implemented on top of the existing `create`.

rgrinberg:
I pushed a commit that uses `Top_closure`. It seems to work as we'd expect.

rgrinberg:
You could also consider adding this command under `dune describe pkg list-locked-dependencies`.

rgrinberg (approved):

bin/pkg/list_locked_dependencies.ml
gridbugs: `Pp.text` wraps the text in a way that I find hard to read. With `verbatim` the line doesn't get broken whereas switching to `Pp.text` it wraps like:
```
- Immediate dependencies of local package
   local_2.dev
   - baz.0.0.1
   - qux.0.0.1 (test only)
```
I find that hard to read because it looks to me like the subheading is `Immediate dependencies of local package` but actually the subheading is `Immediate dependencies of local package local_2.dev`.

rgrinberg: I think that's because you just forgot the box around the text. E.g. it should be:

```
Pp.hovbox (Pp.text ..)
```

We're getting `Pp.paragraph` to make this a little easier to remember.

bin/pkg/list_locked_dependencies.ml
gridbugs: It looks like `Pp.cut` doesn't guarantee that the line will be broken though. My intention here is to print a nested list of titled lists of the form:
```
- Dependencies of local package a.dev
  - dep1.0.0.1
  - dep2.0.0.1
  - ...
- Dependencies of local package b.dev
  - dep1.0.0.1
  - ...
- ...
```
I can mostly achieve this with `Pp.enumerate`  but the title of the sublists (the `Dependencies of local package ...` part) isn't an element of the list so I need another way to print a line of text above the list.

What does it mean to "break boxes that are already open"?

src/dune_pkg/package_universe.mli
gridbugs: done

bin/pkg/list_locked_dependencies.ml
gridbugs: I've got something that I think works using `Pp.vbox` to ensure that all the `cut`s cause a break. Let me know what you think.

bin/pkg/list_locked_dependencies.ml
gridbugs: Ah cool I just found `Pp.hbox` which prevents line breaks.

bin/pkg/list_locked_dependencies.ml
rgrinberg: >  It looks like Pp.cut doesn't guarantee that the line will be broken though

If you want to try guarantee that a line will be broken, make a vertical box and then every space will be interpreted as a newline. See the implementation of `Pp.enumerate` for an example.

> What does it mean to "break boxes that are already open"?

Whenever you do `Pp.{box,vbox,hbox,hovbox,..} foo` we allow whoever is rendering `foo` to choose how to interpret `Pp.space` and `Pp.cut`. That allows us our errors to be well wrapped no matter the size of the user's terminal or if they're displayed as LSP diagnostics, to be wrapped there appropriately as well.

Therefore whenever you output a newline, you're breaking the layout engine's assumptions on how much more its allowed to render on this line before wrapping. Essentially, making your `Pp.t` value uncomposable since it cannot be wrapped with boxes.

PR #9096, opened 2023-11-06
refactor(pkg): fetch all package extra files in a single call

rgrinberg:
Previously, we'd call `git show` essentially for every single package to try and see if it has `pkg.files`. Now we do all this work in a single call.

### DIFF
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -213,96 +213,127 @@ let scan_files_entries path =
   read [] Path.Local.root
 ;;
 
-let get_opam_package_files t opam_package =
-  match t.source with
-  | Directory path ->
-    let file_path = Path.append_local path (Paths.files_dir opam_package) in
-    scan_files_entries file_path
-    |> List.map ~f:(fun local_file ->
-      let original = File_entry.Path (Path.append_local file_path local_file) in
-      { File_entry.local_file; original })
-    |> Fiber.return
-  | Repo at_rev ->
-    let files_root = Paths.files_dir opam_package in
-    Rev_store.At_rev.directory_entries at_rev files_root
-    |> Rev_store.File.Set.to_list
-    |> Fiber.parallel_map ~f:(fun remote_file ->
-      let remote_file_path = Rev_store.File.path remote_file in
-      Rev_store.At_rev.content at_rev remote_file_path
-      >>| function
-      | None ->
-        Code_error.raise
-          "Enumerated file in directory but file can't be retrieved"
-          [ "local_file", Rev_store.File.to_dyn remote_file ]
-      | Some content ->
-        let local_file =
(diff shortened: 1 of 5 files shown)


src/dune_pkg/opam_repo.ml
```diff
@@ -213,96 +213,127 @@ let scan_files_entries path =
...
+  let indexed = List.mapi with_files ~f:(fun i w -> i, w) |> Int.Map.of_list_exn in
+  let from_dirs, from_git =
+    Int.Map.partition_map indexed ~f:(fun (with_file : With_file.t) ->
+      match with_file.extra_files with
+      | Git_files files -> Right (with_file.package, files)
+      | Inside_files_dir ->
+        let dir =
+          match with_file.repo.source with
+          | Directory root -> Path.append_local root (Paths.files_dir with_file.package)
+          | Repo _ -> assert false
```
Leonidas-from-XIV, line 271: This assert looks rather ugly as it represents a case that shouldn't happen yet with the types modelled as it, it does. I would instead suggest adding an argument to this constructor that has the `root`, which would avoid this case.

(Or convert this assert to a `Code_error.raise` since there is definitely some bug if this is happening).

src/dune_pkg/opam_repo.ml
```diff
@@ -378,11 +419,8 @@ let load_all_versions ts opam_package_name =
     | [] -> Fiber.return []
     | packages ->
       let* rev_store = rev_store in
-      let+ opam_files =
-        List.map packages ~f:(fun (repo, file, _) -> repo, file)
-        |> load_packages_from_git rev_store
-      in
-      List.map2 opam_files packages ~f:(fun opam_file (_, _, pkg) -> pkg, opam_file)
+      let+ opam_files = load_packages_from_git rev_store packages in
+      List.map2 opam_files packages ~f:(fun opam_file (_, _, pkg, _) -> pkg, opam_file)
```
Leonidas-from-XIV, line 423: Maybe it would make sense to change `load_packages_from_git` to save this additional `map2`, given it already does a `map2` over `opam_files`.

rgrinberg: It's possible but it would be a bit weird. We'd be complicating the type of `load_packages_from_git` by making it "know" the shape the caller wants. Since the perf of this `map2` isn't important, i'd rather leave it simpler.

src/dune_pkg/opam_repo.ml
rgrinberg: I've converted this to an assert false. I have an idea on how to get rid of the assert completely, but it would take quite a bit more refactoring.

rgrinberg:
@Leonidas-from-XIV is this PR good?

Leonidas-from-XIV:
Ah, sorry. Yes, it's good! I'll rebase and merge :)

PR #9097, opened 2023-11-07
refactor: lazy target checking

rgrinberg:
Allow checking if a dierctroy is a target without generating all
directory targets.

### DIFF
--- a/src/dune_engine/build_config.ml
+++ b/src/dune_engine/build_config.ml
@@ -17,12 +17,41 @@ module Gen_rules = struct
     let union a b = Path.Build.Map.union a b ~f:(fun _ a b -> Some (Subdir_set.union a b))
   end
 
+  module Directory_targets = struct
+    include Build_config_intf.Directory_targets
+
+    let empty =
+      { all = Memo.return Path.Build.Map.empty; mem = (fun _ -> Memo.return false) }
+    ;;
+
+    let of_map all =
+      { all = Memo.return all; mem = (fun p -> Memo.return (Path.Build.Map.mem all p)) }
+    ;;
+
+    let union_exn t { all; mem } =
+      let open Memo.O in
+      let all =
+        Memo.Lazy.create (fun () ->
+          let+ left = t.all
+          and+ right = all in
+          Path.Build.Map.union_exn left right)
+      in
+      let mem p =
+        t.mem p
+        >>= function
+        | true -> Memo.return true
+        | false -> mem p
+      in
+      { all = Memo.Lazy.force all; mem }
+    ;;
+  end
+
   module Rules = struct
     include Build_config_intf.Rules
 
     let empty =
       { build_dir_only_sub_dirs = Path.Build.Map.empty
-      ; directory_targets = Path.Build.Map.empty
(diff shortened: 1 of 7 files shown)


src/dune_rules/gen_rules.ml
```diff
@@ -618,7 +618,10 @@ let gen_rules_standalone_or_root
     Rules.union rules rules'
   in
   let* build_config =
-    let+ directory_targets = collect_directory_targets ~dir ~init:directory_targets in
+    let+ directory_targets =
+      let+ directory_targets = collect_directory_targets ~dir ~init:directory_targets in
+      Gen_rules.Directory_targets.of_map directory_targets
```
rleshchinskiy, line 623: If the goal is to construct these things lazily why are we calling `collect_directory_targets` eagerly here? What is it exactly that's supposed to be lazy? Just the unions?

src/dune_engine/build_config_intf.ml
```diff
@@ -13,13 +13,20 @@ end
 
 type rules = Rules.t
 
+module Directory_targets = struct
+  type t =
+    { all : Loc.t Path.Build.Map.t Memo.t
+    ; mem : Path.Build.t -> bool Memo.t
+    }
+end
```
rleshchinskiy, line 21: Since this type has invariants I think it would be better to not allow clients to constructs instances which break them. So it might be better to keep it abstract and expose `all` and `mem` as functions.

src/dune_engine/build_config.ml
```diff
@@ -17,12 +17,41 @@ module Gen_rules = struct
...
+      in
+      let mem p =
+        t.mem p
+        >>= function
+        | true -> Memo.return true
+        | false -> mem p
+      in
+      { all = Memo.Lazy.force all; mem }
+    ;;
+  end
```
rleshchinskiy, line 47: Just so I understand correctly: we only ever construct this through `of_map` and `union_exn`, right? So what this basically does it provide O(1) union?

If so, then it might be clearer to just have a generic data structure which does just that. One thing I'm worried about is that if we union multiple `Directory_targets` and then evaluate `all`, we will hang on to both the materialised union (in `all`) and the individual bits (in `mem`), effectively doubling the memory requirements. Here is a sketch (not type checked) of a data structure which provides O(1) union but doesn't have that problem:

```
type ('a,'b) t = { all : ('a,'b) Map.t Lazy.t; mem : ('a -> bool) ref }

let union_exn t u =
  let mem = ref (fun x -> !t.mem x || !u.mem x) in
  let all = lazy (
    let all = Map.union_exn (force t.all) (force u.all) in
    mem := (fun x -> Map.mem all x);
    all)
  in
  { all; mem }
```

Another potential issue is that `mem` now does a linear search over the unioned components where before, it was O(log N). Do you expect that to matter in practice?

rleshchinskiy (changes requested):

rgrinberg:
Thanks for the feedback. I'm shelfing this for now. I've managed to work around the issue by refactoring the declaration of directory targets. There's still some possibility of cycles, and I will revisit this PR once they come up.

PR #9100, opened 2023-11-07
refactor: Move Package_version to Dune_lang

gridbugs:


src/dune_rules/link_time_code_gen.ml
```diff
@@ -140,7 +140,7 @@ let build_info_code cctx ~libs ~api_version =
         match Lib_info.version (Lib.info lib) with
         | Some v ->
           Memo.return
-            (sprintf "Some %S" (Dune_pkg.Package_version.to_string v), placeholders)
+            (sprintf "Some %S" (Dune_lang.Package_version.to_string v), placeholders)
```
rgrinberg, line 143: To avoid invasive diffs in the future, I recommend that you put `Package_version` in `dune_rules/import.ml` and then just use the short version everywhere. No need to fully qualify this module as it's very unlikely we'll have another one with the same name elsewhere.

rgrinberg (approved):

src/dune_rules/link_time_code_gen.ml
gridbugs: Good idea. Updated.

PR #9103, opened 2023-11-07
pkg: add Dune_pkg.Local_package module

gridbugs:
Thus far we've been using `OpamFile.OPAM.t` to represent local packages in our package management code, mainly because dune's `Package.t` type is in the `dune_rules` library which `dune_pkg` can't depend on (to avoid a dependency cycle). This change adds a minimal `Dune_pkg.Local_package.t` type containing the information about a local package that is relevant to package management from which a `OpamFile.OPAM.t` can be derived. This also replaces the `OpamSolver.local_package` type which was opam-specific and contained an unused `file` field (removed in this change).

Having our own local package type rather than depending on opam's will help clarify the intent of code through types and reduce the degree to which pkg's api is opam-specific. It also means we can easily implement `to_dyn` and other helpful functions without having to worry about how to represent values of `OpamFile.OPAM.t`.

Also in this change the `Package_version.*_opam` conversion functions are renamed to `Package_version.*_opam_package_version` to clarify that they convert between `Package_version.t` and `OpamPackage.Version.t` (rather than `OpamVersion.t`) and for consistency with the `Package_name.*_opam_package_version` functions.

### DIFF
--- a/src/dune_pkg/local_package.mli
+++ b/src/dune_pkg/local_package.mli
@@ -1,0 +1,13 @@
+open! Import
+
+(** Information about a package that's relevant for solving dependencies *)
+type t =
+  { name : Package_name.t
+  ; version : Package_version.t option
+  ; dependencies : Package_dependency.t list
+  }
+
+(** [to_opam_file t] returns an [OpamFile.OPAM.t] whose fields are based on the
+    fields of [t]. Note that this does not actually create a corresponding file
+    on disk. *)
+val to_opam_file : t -> OpamFile.OPAM.t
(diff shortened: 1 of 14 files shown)


rgrinberg (approved):

src/dune_pkg/local_package.mli
```diff
@@ -0,0 +1,13 @@
+open! Import
```
Leonidas-from-XIV, line 1: I like the PR description as it adds context and reasoning for this, so I think adding a documentation comment to this module what a `Local_package` is and why it exists would be good!

Leonidas-from-XIV (approved):
I have rerun that flaky macOS test, let's hope it will work this time :slightly_smiling_face:

src/dune_pkg/local_package.mli
gridbugs: Will add in a later change.

PR #9114, opened 2023-11-08
pkg: add local packages hash to lockdir

gridbugs:
If the package definitions in dune-project are changed after a lockdir is generated then the lockdir may no longer contain a valid solution for the local packages in the project. To help detect when this has happened this change adds a new field to the lockdir containing a hash of all the local package metadata used for generating it. When the local package metadata would change in a way that may affect the solution we can detect this by computing a new hash of the local packages and comparing it against the one stored in the lockdir.

### DIFF
--- a/src/dune_lang/package_constraint.ml
+++ b/src/dune_lang/package_constraint.ml
@@ -9,12 +9,30 @@ module Op = struct
     | Lt
     | Neq
 
-  let equal a b =
+  (* Define an arbitrary ordering on [t] to allow a package constraint to be
+     used as the key of a map or set. The order from lowest to highest is:
+     [Eq, Gte, Lte, Gt, Lt, Neq] *)
+  let compare a b =
     match a, b with
-    | Eq, Eq | Gte, Gte | Lte, Lte | Gt, Gt | Lt, Lt | Neq, Neq -> true
-    | _ -> false
+    | Eq, Eq -> Ordering.Eq
+    | Eq, _ -> Lt
+    | _, Eq -> Gt
+    | Gte, Gte -> Eq
+    | Gte, _ -> Lt
+    | _, Gte -> Gt
+    | Lte, Lte -> Eq
+    | Lte, _ -> Lt
+    | _, Lte -> Gt
+    | Gt, Gt -> Eq
+    | Gt, _ -> Lt
+    | _, Gt -> Gt
+    | Lt, Lt -> Eq
+    | Lt, _ -> Lt
+    | _, Lt -> Gt
+    | Neq, Neq -> Eq
   ;;
 
+  let equal a b = Ordering.is_eq (compare a b)
   let map = [ "=", Eq; ">=", Gte; "<=", Lte; ">", Gt; "<", Lt; "<>", Neq ]
   let to_dyn t = Dyn.variant (fst (List.find_exn ~f:(fun (_, op) -> equal t op) map)) []
 
@@ -24,7 +42,7 @@ module Op = struct
     List.find_exn ~f map |> fst
   ;;
 
-  let encode x = Dune_sexp.Encoder.string (to_string x)
+  let encode x = to_string x |> Dune_sexp.Encoder.string
 end
 
 module Variable = struct
@@ -38,6 +56,8 @@ module Variable = struct
     Dune_sexp.Decoder.atom_matching ~desc:"variable" (fun s ->
       if String.is_prefix s ~prefix:":" then Some (of_name (String.drop s 1)) else None)
   ;;
+
+  let compare { name } t = String.compare name t.name
 end
 
 module Value = struct
@@ -48,6 +68,14 @@ module Value = struct
   let to_dyn = function
     | String_literal s -> Dyn.variant "String_literal" [ Dyn.string s ]
     | Variable v -> Dyn.variant "Variable" [ Variable.to_dyn v ]
+  ;;
+
+  let compare a b =
+    match a, b with
+    | String_literal a, String_literal b -> String.compare a b
+    | String_literal _, _ -> Lt
+    | _, String_literal _ -> Gt
+    | Variable a, Variable b -> Variable.compare a b
   ;;
 
   let encode = function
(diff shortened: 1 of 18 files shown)


gridbugs:
Making this PR as a strawman to start a conversation about whether this is the right way to detect when dune-project becomes out of sync with the lockdir. Now that I've thought a little more about this in the process of making this change my preference for solving this problem would be to check that all dependencies of local packages can be satisfied from the lockdir. That would have the benefit of not needing to store any more information in the lockdir, would allow us to produce richer error messages than just "the hashes don't match", and also detect the case where a user has tampered with their lockdir (which hashing local packages obviously won't catch).

rgrinberg:
> . Now that I've thought a little more about this in the process of making this change my preference for solving this problem would be to check that all dependencies of local packages can be satisfied from the lockdir. That would have the benefit of not needing to store any more information in the lockdir, would allow us to produce richer error messages than just "the hashes don't match", and also detect the case where a user has tampered with their lockdir (which hashing local packages obviously won't catch).

This doesn't really help when the user updates their constraints, right?

Leonidas-from-XIV:
Personally I think creating some kind of checksum of the input to the solver and printing a warning that the lock folder might be out of date seems fine - a simple and understandable solution that is easy to explain to users.

> Now that I've thought a little more about this in the process of making this change my preference for solving this problem would be to check that all dependencies of local packages can be satisfied from the lockdir.

How would you do this? The obvious solution to me to run the solver with only the currently selected packages as candidates works but I worry that running the solver all the time would be a bit time-consuming and not necessarily correct as the packages metadata in the repo might have changed (e.g. a package was marked as non-available etc).

Leonidas-from-XIV:
Generally: a good idea. I am always annoyed when I get my `dune-project` and the generated opam files out of sync by accident and with lock files this is even easier as you need to lock specifically.

Bike shedding: I find the naming "local package hash" confusing as as an user I would call this "dependency hash", as this describes the dependencies that I have declared.

Also, if we use such a hash (or any method of checking that the lock dirs and `dune-project` are in sync), I don't think it should be a separate command, as that basically requires users to check manually - and if they remember to do, they could also remember to re-lock. It should be happening any time dune would use the lock dir as this is the place where things go wrong. This of course means that any such check would need to be fast enough to be run automatically.

gridbugs:
I've updated this PR so that it only hashes the set of non-local dependencies of local packages. This way certain operations such as adding a package which only depends on existing dependencies (including their constraints) doesn't affect the hash and so won't require regenerating the lockdir.

gridbugs:
> I don't think it should be a separate command, as that basically requires users to check manually

Agreed. I've done it this way for now as we currently don't have any operations that require the lockdir and local packages to be in sync. The first such operation would be https://github.com/ocaml/dune/pull/9095 which is what motivated me to make this PR.

gridbugs:
> > Now that I've thought a little more about this in the process of making this change my preference for solving this problem would be to check that all dependencies of local packages can be satisfied from the lockdir.
> 
> How would you do this? The obvious solution to me to run the solver with only the currently selected packages as candidates works but I worry that running the solver all the time would be a bit time-consuming and not necessarily correct as the packages metadata in the repo might have changed (e.g. a package was marked as non-available etc).

There's enough information already in the lockdir to compute the transitive dependency closure of all local packages and detect when a dependency is absent. There's no need to run the solver again.

gridbugs:
Converted to draft while https://github.com/ocaml/dune/pull/9172 is in review which validates lockdirs against local packages by analyzing dependencies of local packages and checking if they are contained in the lockdir. That change is needed regardless of the dependency hash in order to gracefully handle the case where files are added/removed/changed in the lockdir in a way which invalidates it w.r.t the local packages (which should be rare but we still need to handle it). Said PR steps on this one's toes a bit so I'll wait for the dust to settle before dealing with all the merge conflicts.

gridbugs:
Updated on top of my recent changes to lockdir validation. This is now ready for review.

src/dune_lang/package_constraint.ml
```diff
@@ -9,12 +9,27 @@ module Op = struct
...
+    | Lte, Lte -> Eq
+    | Lte, _ -> Lt
+    | _, Lte -> Gt
+    | Gt, Gt -> Eq
+    | Gt, _ -> Lt
+    | _, Gt -> Gt
+    | Lt, Lt -> Eq
+    | Lt, _ -> Lt
+    | _, Lt -> Gt
+    | Neq, Neq -> Eq
```
Leonidas-from-XIV, line 29: What is the semantics of this pattern match? Maybe writing down a comment on the ordering would be helpful, since I don't think there's a natural ordering for comparison operators.

src/dune_pkg/dune
```diff
@@ -21,6 +21,7 @@
   build_info
   zeroinstall_solver
   fmt
-  xdg)
+  xdg
+  sha)
```
Leonidas-from-XIV, line 25: Careful, the `sha` package is only vendored because of OPAM, I would be a bit hesitant to just use it directly in other code and probably go for `Digest` instead.

Or at least decide what kinds of hashes we want to use in Dune and deliberately vendor that hash library.

src/dune_pkg/local_package.ml
```diff
@@ -7,15 +8,91 @@ type t =
...
-let to_opam_file { name; version; dependencies; loc = _ } =
-  OpamFile.OPAM.empty
-  |> OpamFile.OPAM.with_name (Package_name.to_opam_package_name name)
-  |> OpamFile.OPAM.with_version_opt
-       (Option.map version ~f:Package_version.to_opam_package_version)
-  |> OpamFile.OPAM.with_depends
-       (Package_dependency.list_to_opam_filtered_formula dependencies)
-;;
+module Dependency_hash = struct
+  module Hash = Sha256
```
Leonidas-from-XIV, line 12: Another thing to be aware of is how to deal with hash migration. Maybe a scheme where the hash is prefixed by its algorithm like OPAMs `sha1=deadcafeb0ba` would make sense in places where the user is exposed to the hash?

src/dune_pkg/local_package.mli
```diff
@@ -1,17 +1,68 @@
...
+      both constraints will form the constraint associated with that
+      dependency. Package constraints are de-duplicated by comparing them only
+      on their syntax. *)
+  type t
+
+  (** Returns a hash of all dependencies in the set or [`Empty] if the set is
+      empty. The reason for behaving differently when the set is empty is so
+      that callers are forced to explicitly handle the case where there are no
+      dependencies which will likely lead to better user experience. *)
+  val hash_or_empty : t -> (Dependency_hash.t, [ `Empty ]) result
```
Leonidas-from-XIV, line 41: I think it is a good idea to distinguish this, but wouldn't `Dependency_hash.t option` suffice? The `None` case is mostly used if something can't be sensibly constructed, exactly like `Empty` here, but with the added benefit of `Option.O`.

src/dune_pkg/dune
gridbugs: switched to use `Digest`

src/dune_pkg/local_package.ml
gridbugs: changed the format to `md5=<hash>`

rgrinberg: I'm sorry I'm a bit late to the party, but I think the hash algorithm should be controlled by the version of the lock directory set in `lock.dune`. The whole hash thing is a complete implementation detail of dune and there's no need to expose any details to the user.

rgrinberg (approved):

Issue #9024, opened 2023-10-27
Dune crashes when menhir_flags is in dune-workspace


Tchou:
## Expected Behavior
Dune should not crash with an unhandled exception when the `(env ...)` stanza of dune-workspace mentions `menhir_flags`

## Actual Behavior
Dune crashes when `menhir_flags` appears in `(env ...)`

## Reproduction

1. create an empty directory `mkdir foo`
2. add these two files, `dune-project`:
```
(lang dune 3.0)
(using menhir 2.0)
```
and `dune-workspace`:
```
(lang dune 3.0)
(env
  (test (menhir_flags --table)))
```
3. run `dune build`, dune immediately crashes with a stack trace.

## Specifications

- Version of `dune` (output of `dune --version`): 3.11.1
- Version of `ocaml` (output of `ocamlc --version`): 4.14.1
- Operating system (distribution and version): Ubuntu 23.04, x86_64

## Additional information 

- Link to gist with verbose output (run `dune` with the `--verbose` flag):
https://gist.github.com/Tchou/bfdcbccc048d394c9ad8e2ec1f23c32f

emillon:
Thanks. I added a reproduction case in #9120.
This bug has been present since the introduction of `menhir_flags` in 2.2.0, so we can first add a proper error saying it's not supported. The issue is that the decoding of this field is versioned in term of the menhir extension, and extensions are not available in workspace files. We could probably version this (in workspace files) against `lang dune`.

rgrinberg:
Fixed

PR #9121, opened 2023-11-08
Better message for menhir_flags in dune-workspace

emillon:
This adds a context variable to determine whether we are parsing a dune-workspace file or a dune file.

See #9024


rgrinberg:
Did you look into allowing enabling syntax extensions in the workspace file? This limitation isn't something we want long term.

PR #9129, opened 2023-11-09
tests: use %{cc} instead of cc

emillon:
There is no guarantee of a `cc` executable in scope.


Leonidas-from-XIV (approved):
Sounds very sensible, probably makes the tests also more portable.

PR #9132, opened 2023-11-09
Remove various helper functions from Action_builder modules

snowleopard:
@rgrinberg As discussed, I'm cleaning up the implementation of `Action_builder` modules, so I'm deleting all inessential helpers that can be easily implemented on the rules side.

Could you take over from here, re-adding the helpers in Dune rules? It's probably easiest to extend the `Action_builder` module when importing `Dune_engine` in the rules by just copying the helpers I deleted.


snowleopard:
Btw, this clean up includes simplifying the implementation of `paths_matching` (previously called `paths_matching_unit`) by noticing that the result of the computation is just ignored. With all the unused code around, it was not obvious that the code could be made simpler. Now it is.

Internal benchmarks show consistent improvements in build times (1-2%) and allocations (up to 1%).

snowleopard:
Actually, I've just realised that `Action_builder.path_matching` was just a trivial special case of `Action_builder.record`, so I nuked it too.

snowleopard:
I've removed `Action_builder.contents path` which can now be replaced with `Action_builder.of_memo (Memo.of_thunk_apply Build_system.read_file path)`.

snowleopard:
`Action_builder.*file_exists` are gone too. They can be trivially implemented via `Build_system.file_exists`. 

I noticed that internally we had a bunch of incorrect uses of `Action_builder.memo` where we passed a prematurely forced computation, so I made `Action_builder.memo` accept a thunk, so one can't mess up.

rgrinberg:
@snowleopard do you have more changes pending or should I update the rules?

snowleopard:
> @snowleopard do you have more changes pending or should I update the rules?

There are a couple of more removals I'm working on. I'll ping you in the next couple of days!

snowleopard:
@rgrinberg OK, I think I'm done. My next step is changing the implementation of `Action_builder.t` but that will be done in a separate PR. Please take over from here, and let me know if I messed up anything.

rgrinberg:
@snowleopard I think you forgot to push `Dep.Facts.record_facts` and the `visibility_override` in `dune_engine/action.ml`

snowleopard:
> @snowleopard I think you forgot to push `Dep.Facts.record_facts` and the `visibility_override` in `dune_engine/action.ml`

Oops. `visibility_override` is just an internal thing. 

I added `Dep.Facts.record_facts`, though a few other changes happened in `dep.ml` so I won't be surprised if I'm missing something else... Let me know!

rgrinberg:
Completed in a series of PR's culminating in https://github.com/ocaml/dune/pull/9559

PR #9140, opened 2023-11-10
fix(pkg): Add a lock around the revision store

Leonidas-from-XIV:
This prevents multiple processes to modify the git repo in an invalid way concurrently.

Thanks to @emillon for helping to figure out the puzzle of the ignored lock (morale of the story: mixing blocking and nonblocking code leads to odd behavior due to unexpected order of execution).

This is the lock part that #9020 required as a separate PR, so it just contains the lock change (which I believe is not controversial) and I can rebase #9020 on top of this. I threw in the test for `add_repo` to show that adding the lock didn't break anything.

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -2,11 +2,72 @@ open Stdune
 open Dune_vcs
 module Process = Dune_engine.Process
 module Display = Dune_engine.Display
+module Scheduler = Dune_engine.Scheduler
 module Re = Dune_re
+module Flock = Dune_util.Flock
 open Fiber.O
 
 type t = { dir : Path.t }
+
+let lock_path { dir } =
+  let parent = dir |> Path.parent_exn in
+  Path.relative parent "rev-store.lock"
+;;
+
 type rev = Rev of string
+
+(* Async-inspired variation of [Fiber.repeat_until] *)
+let rec repeat_until_finished state f =
+  let* computation = f state in
+  match computation with
+  | `Repeat state -> repeat_until_finished state f
+  | `Finished result -> Fiber.return result
+;;
+
+let attempt_to_lock flock lock ~max_tries =
+  let sleep_duration = 0.1 in
+  repeat_until_finished max_tries (function
+    | 0 -> Fiber.return @@ `Finished (Ok `Failure)
+    | retry ->
+      (match Flock.lock_non_block flock lock with
+       | Ok `Success as ok -> Fiber.return @@ `Finished ok
+       | Ok `Failure ->
+         let+ () = Scheduler.sleep sleep_duration in
+         `Repeat (retry - 1)
+       | err -> Fiber.return @@ `Finished err))
+;;
+
+let with_flock lock_path ~f =
+  let open Fiber.O in
+  let parent = Path.parent_exn lock_path in
+  Path.mkdir_p parent;
+  let fd = Unix.openfile (Path.to_string lock_path) [ Unix.O_CREAT; O_RDONLY ] 0o644 in
+  let flock = Flock.create fd in
+  let max_tries = 50 in
+  Fiber.finalize
+    ~finally:(fun () ->
+      (* closing the fd releases the flock automatically *)
+      match Unix_error.Detailed.catch Unix.close fd with
+      | Ok () ->
+        (* delete the lock to signal to the user we don't hold a lock *)
+        Fiber.return @@ Path.unlink_no_err lock_path
+      | Error detailed -> Unix_error.Detailed.raise detailed)
+    (fun () ->
+      let* acquired = attempt_to_lock flock Flock.Exclusive ~max_tries in
+      match acquired with
+      | Ok `Success -> f ()
+      | Ok `Failure ->
+        Code_error.raise
+          (sprintf "Couldn't acquire lock after %d attempts to lock" max_tries)
+          []
+      | Error error ->
+        User_error.raise
+          [ Pp.textf
+              "Failed to get a lock for the revision store at %s: %s"
+              (Path.to_string_maybe_quoted lock_path)
+              (Unix.error_message error)
+          ])
+;;
 
 let equal { dir } t = Path.equal dir t.dir
 let display = Display.Quiet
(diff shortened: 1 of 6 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -2,43 +2,97 @@ open Stdune
 open Dune_vcs
 module Process = Dune_engine.Process
 module Display = Dune_engine.Display
+module Scheduler = Dune_engine.Scheduler
 module Re = Dune_re
+module Flock = Dune_util.Flock
 open Fiber.O
 
-type t = { dir : Path.t }
+type t =
```
emillon, line 10: Since `lock` and `dir` are always relative to each other I would argue to keep just `dir` in `t` and compute the lock path from it.

src/dune_pkg/rev_store.ml
```diff
@@ -2,43 +2,97 @@ open Stdune
...
+  let timeout = max timeout 0. in
+  let sleep_duration = 0.1 in
+  repeat_until_finished timeout (fun timeout ->
+    if timeout > 0.
+    then (
+      match Flock.lock_non_block flock lock with
+      | Ok `Success as ok -> Fiber.return @@ `Finished ok
+      | Ok `Failure ->
+        let+ () = Scheduler.sleep sleep_duration in
+        `Repeat (timeout -. sleep_duration)
```
emillon, line 35: One alternative is to keep a count of retries. That avoids float math weirdness like "can repeated substraction of 0.1 from 5. yield a negative number?". In that case the timeout is only used once so I think we could remove the parameter and hardcode 50 retries of 0.1 seconds. Just an idea

src/dune_pkg/rev_store.ml
```diff
@@ -2,43 +2,97 @@ open Stdune
...
+  let open Fiber.O in
+  let parent = Path.parent_exn lock_path in
+  Path.mkdir_p parent;
+  let fd = Unix.openfile (Path.to_string lock_path) [ Unix.O_CREAT; O_RDONLY ] 0o644 in
+  let flock = Flock.create fd in
+  let max_duration = 5. in
+  Fiber.finalize
+    ~finally:(fun () ->
+      (* closing the fd releases the flock automatically *)
+      Unix.close fd;
```
emillon, line 50: What happens if `close` fails? (`close_noerr` is probably good enough?)

src/dune_pkg/rev_store.ml
```diff
@@ -2,43 +2,97 @@ open Stdune
...
+    (fun () ->
+      let* acquired = lock_limited flock Flock.Exclusive max_duration in
+      match acquired with
+      | Ok `Success -> f ()
+      | Ok `Failure ->
+        Code_error.raise
+          (sprintf "Couldn't acquire lock after %.1f seconds" max_duration)
+          []
+      | Error error ->
+        Code_error.raise "lock" [ "error", Dyn.string (Unix.error_message error) ])
```
emillon, line 62: I think we just want to reraise it as `Unix.Unix_error error`. This is what will happen on filesystems that don't support flock. I suppose NFS is going to throw errors here.
Not strictly needed but you can test that by injecting an error with `strace`, I did that in a sendfile test (grepping for `sendfile` in the test suite should find it).

src/dune_pkg/rev_store.ml
rgrinberg: If `close` fails, it's very likely that it's `EBADFD` and that means we have a bug somewhere. We shouldn't ignore it.

src/dune_pkg/rev_store.ml
rgrinberg: A unix error is better than a code error, but a user error with the context would be even better.

src/dune_pkg/rev_store.ml
```diff
@@ -84,18 +138,21 @@ let show =
...
-      User_error.raise
-        [ Pp.textf "%s isn't a directory" (Path.to_string_maybe_quoted dir)
-        ; Pp.textf "reason: %s" (Unix_error.Detailed.to_string_hum (e, x, y))
-        ]
-        ~hints:[ Pp.text "delete this file or check its permissions" ]
+    with_flock lock ~f:(fun () ->
+      match Fpath.mkdir_p (Path.to_string dir) with
+      | Already_exists -> Fiber.return ()
+      | Created -> run t [ "init"; "--bare" ]
+      | exception Unix.Unix_error (e, x, y) ->
```
gridbugs, line 150: `(e, x, y)` are meaningless names and this code just passes the tuple on to `Unix_error.Detailed.to_string_hum` so I'd suggest not destructuring the tuple here.

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: Indeed, this is better. I had this left over from when the lock was an actual sort of lock but now that it is just a `Path.t` it makes no sense. Thanks for pointing it out!.

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: Sure. I thought a time was a bit better of an API but it's unpleasantly continuous. This approach has the advantage that we can guarantee 50 iterations, but 50 times sleeping 0.1 doesn't mean that we stopped at 5s, it might've been more.

The main ugly thing is that this is still trying for some randomly picked period, but I can't think of an immediately better solution.

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: I thought so too, but in this case `Unix_error` is an constructor of arity 3, and not a constructor of arity 1 with a tuple, so it can't be bound to a tuple value:

```ocaml
# type a = A of (int * string) | B of int * string;;
# let x = A (3, "42");;
val x : a = A (3, "42")
# let x = B (3, "42");;
val x : a = B (3, "42")
# match x with | A v -> () | B v -> ();;
Error: The constructor B expects 2 argument(s),
       but is applied here to 1 argument(s)
```

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: I've wrapped it with some error handling code. Unfortunately I don't think I can use the same `strace` trick since injecting into `close` is would require knowing the FD otherwise everything breaks.

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: Ah yes, the `sendfile` trick is pretty cool, I wasn't aware that `strace` can also inject failures. But I guess it makes sense if it intercepts all syscalls :+1:

src/dune_pkg/rev_store.ml
```diff
@@ -2,12 +2,76 @@ open Stdune
...
+  let flock = Flock.create fd in
+  let max_tries = 50 in
+  Fiber.finalize
+    ~finally:(fun () ->
+      (* closing the fd releases the flock automatically *)
+      match Unix_error.Detailed.catch Unix.close fd with
+      | Ok () ->
+        (* delete the lock to signal to the user we don't hold a lock *)
+        Fiber.return @@ Path.unlink_no_err lock_path
+      | Error detailed ->
```
gridbugs, line 54: Why is this a `Code_error` and not a `User_error`?

Leonidas-from-XIV: If there is some error that prevents the lock being released, it is probably a bug in our codebase. Like the FD is wrong or something, in any case not something that the user can do much about except for reporting it as a bug. Maybe if the lockfile was on NFS and the server went away or something.

test/blackbox-tests/test-cases/pkg/dune
```diff
@@ -12,7 +12,17 @@
 
 (cram
  (deps %{bin:git})
- (applies_to git-source opam-repository-download multiple-opam-repos))
+ (applies_to
+  git-source
```
emillon, line 16: I'd bite the bullet and got with `applied_to :whole_subtree`, we often do that in other subsystems

emillon (approved):

src/dune_pkg/rev_store.ml
emillon: You can also go with `Unix_error.Detailed.raise detailed`

test/blackbox-tests/test-cases/pkg/dune
Leonidas-from-XIV: Yeah, with the rev-store using git under the hood, its fairly likely to pull in a dependency on it unknowingly.

PR #9141, opened 2023-11-10
Refactor and enable Memo counters

snowleopard:
Upstreaming internal changes that improve Memo metrics (move them out of `memo.ml` completely, clean up the naming, enable unconditionally as they are cheap enough).

We've reviewed, tested and benchmarked this change internally and found no issues.


snowleopard:
Merging as reviewed internally.

PR #9148, opened 2023-11-12
revert #7263

rgrinberg:
We aren't allowed to load the coq scope in projects where coq isn't
enabled.


Alizter:
I don't see how this can be trigerred without a coq stanza present. And for a coq stanza you need coq lang to be enabled.

rgrinberg:
Yes, this wasn't an issue. This was work was replaced with #9149 btw

PR #9172, opened 2023-11-14
Validate lockdir against local packages

gridbugs:
This adds a type `Dune_pkg.Package_universe.t` made up of the set of local packages in a project and a lockdir whose constructor validates that the lockdir contains a valid solution for the dependencies of local packages. As we add more pkg operations we'll need a way of talking about the set of packages that makes up a project, both local and not.

Also this adds a command `dune pkg validate-lockdir` which validates lockdirs against local packages in order to test this feature.

### DIFF
--- a/bin/pkg/validate_lock_dir.ml
+++ b/bin/pkg/validate_lock_dir.ml
@@ -1,0 +1,84 @@
+open! Import
+module Package_universe = Dune_pkg.Package_universe
+module Lock_dir = Dune_pkg.Lock_dir
+module Opam_repo = Dune_pkg.Opam_repo
+module Package_version = Dune_pkg.Package_version
+module Opam_solver = Dune_pkg.Opam_solver
+
+let info =
+  let doc = "Validate that a lockdir contains a solution for local packages" in
+  let man = [ `S "DESCRIPTION"; `P doc ] in
+  Cmd.info "validate-lockdir" ~doc ~man
+;;
+
+let enumerate_lock_dirs_by_path ~context_name_arg ~all_contexts_arg =
+  let open Fiber.O in
+  let+ per_contexts =
+    Pkg_common.Per_context.choose
+      ~context_name_arg
+      ~all_contexts_arg
+      ~version_preference_arg:None
+  in
+  List.filter_map per_contexts ~f:(fun { Pkg_common.Per_context.lock_dir_path; _ } ->
+    if Path.exists (Path.source lock_dir_path)
+    then (
+      try Some (Ok (lock_dir_path, Lock_dir.read_disk lock_dir_path)) with
+      | User_error.E e -> Some (Error (lock_dir_path, `Parse_error e)))
+    else None)
+;;
+
+let validate_lock_dirs ~context_name_arg ~all_contexts_arg =
+  let open Fiber.O in
+  let+ lock_dirs_by_path = enumerate_lock_dirs_by_path ~context_name_arg ~all_contexts_arg
+  and+ local_packages = Pkg_common.find_local_packages in
+  if List.is_empty lock_dirs_by_path then print_endline "No lockdirs to validate.";
+  let errors_by_path =
+    List.filter_map lock_dirs_by_path ~f:(function
+      | Error e -> Some e
+      | Ok (path, lock_dir) ->
+        (match Package_universe.create local_packages lock_dir with
+         | Ok _ -> None
+         | Error e -> Some (path, `Lock_dir_out_of_sync e)))
+  in
+  if not (List.is_empty errors_by_path)
+  then (
+    List.iter errors_by_path ~f:(fun (path, error) ->
+      match error with
+      | `Parse_error error ->
+        User_message.prerr
+          (User_message.make
+             [ Pp.textf
+                 "Failed to parse lockdir %s:"
+                 (Path.Source.to_string_maybe_quoted path)
+             ; User_message.pp error
+             ])
+      | `Lock_dir_out_of_sync error ->
+        User_message.prerr
+          (User_message.make
+             [ Pp.textf
+                 "Lockdir %s does not contain a solution for local packages:"
+                 (Path.Source.to_string path)
+             ]);
+        User_message.prerr error);
+    User_error.raise
+      [ Pp.text "Some lockdirs do not contain solutions for local packages:"
+      ; Pp.enumerate errors_by_path ~f:(fun (path, _) ->
+          Pp.text (Path.Source.to_string path))
+      ])
+;;
+
+let term =
+  let+ builder = Common.Builder.term
+  and+ context_name =
+    Pkg_common.context_term ~doc:"Validate the lockdir associated with this context"
+  and+ all_contexts =
+    Arg.(value & flag & info [ "all-contexts" ] ~doc:"Validate all lockdirs")
+  in
+  let builder = Common.Builder.forbid_builds builder in
+  let common, config = Common.init builder in
+  Scheduler.go ~common ~config
+  @@ fun () ->
+  validate_lock_dirs ~context_name_arg:context_name ~all_contexts_arg:all_contexts
+;;
+
+let command = Cmd.v info term
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -705,3 +705,32 @@ module Load_immediate = Make_load (struct
   end)
 
 let read_disk = Load_immediate.load
+
+let transitive_dependency_closure t start =
+  let missing_packages =
+    let all_packages_in_lock_dir = Package_name.Set.of_keys t.packages in
+    Package_name.Set.diff start all_packages_in_lock_dir
+  in
+  match Package_name.Set.is_empty missing_packages with
+  | false -> Error (`Missing_packages missing_packages)
+  | true ->
+    let to_visit = Queue.create () in
+    let push_set = Package_name.Set.iter ~f:(Queue.push to_visit) in
+    push_set start;
+    let rec loop seen =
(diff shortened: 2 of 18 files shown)


bin/pkg/validate_lock_dir.ml
```diff
@@ -0,0 +1,79 @@
...
+  let open Fiber.O in
+  let+ per_contexts =
+    Per_context.choose ~context_name_arg ~all_contexts_arg ~version_preference_arg:None
+  in
+  List.filter_map per_contexts ~f:(fun { Per_context.lock_dir_path; _ } ->
+    if Path.exists (Path.source lock_dir_path)
+    then (
+      try Some (lock_dir_path, Lock_dir.read_disk lock_dir_path) with
+      | User_error.E e ->
+        User_warning.emit
```
Leonidas-from-XIV, line 25: If the lock dir is to be validated and it can't be read, should this be a warning? It seems like an error.

bin/pkg/validate_lock_dir.ml
```diff
@@ -0,0 +1,79 @@
...
+        None)
+    else None)
+;;
+
+let validate_lock_dirs ~context_name_arg ~all_contexts_arg =
+  let open Fiber.O in
+  let+ lock_dirs_by_path = enumerate_lock_dirs_by_path ~context_name_arg ~all_contexts_arg
+  and+ local_packages = find_local_packages in
+  if List.is_empty lock_dirs_by_path
+  then User_error.raise [ Pp.text "No lockdirs to validate." ];
```
Leonidas-from-XIV, line 40: It's a good question what the result should be if there is no lock dirs. Is the absence of any lock dirs an error or is it a success because all 0 lock dirs are valid?

I lean towards the latter but maybe one wouldn't call `validate-lock-dir` if one did not expect lock dirs to exist. OTOH, having to enable or disable validation explicitly in e.g. scripts is annoying.

bin/pkg/validate_lock_dir.ml
```diff
@@ -0,0 +1,79 @@
+open! Import
+open Pkg_common
```
Leonidas-from-XIV, line 2: I find this a bit confusing to follow (since there's a bunch of names that are unqualified in the code) but that's just me.

src/dune_pkg/lock_dir.ml
```diff
@@ -705,3 +731,30 @@ module Load_immediate = Make_load (struct
   end)
 
 let read_disk = Load_immediate.load
+
+let dependency_transitive_closure t start =
```
Leonidas-from-XIV, line 735: `transitive_dependency_closure`?

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
```diff
@@ -61,7 +61,7 @@ Print the contents of each file in the lockdir:
   
   (version 0.0.1)
   
-  (deps baz bar)
+  (deps bar baz)
```
Leonidas-from-XIV, line 64: How come these are flipped now? I am a bit worried that if this is due to using a map, we might get non-reproducible lockfiles due to using an unordered map. Maybe it would make sense to sort them just to be sure.

test/blackbox-tests/test-cases/pkg/lockdir-validation.t
```diff
@@ -50,3 +50,17 @@ package graph then it's caught when loading the lockdir.
...
+  > EOF
+
+  $ dune describe pkg lock
+  File "dune.lock/a.pkg", line 1, characters 8-9:
+  1 | (deps b b)
+              ^
+  Error: Package "a" in lockdir dune.lock contains duplicate dependencies on
+  package "b".
+  Hint: This could indicate that the lockdir is corrupted. Delete it and then
+  regenerate it by running: 'dune pkg lock'
```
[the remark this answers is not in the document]
gridbugs: Mostly I'm interested in detecting user modifications to the lockdir that cause it to no longer contain a solution to local package dependencies. I agree that this check is possibly too strict. I added it because we keep track of the `Loc.t` of each dependency in the lockdir and I didn't want to think about how to handle the case where a dependency appears twice. Also if a dependency appears twice then the lockfile won't round-trip after this change. Neither of those are big deals though.

bin/pkg/validate_lock_dir.ml
gridbugs: The problem is that if this command is run with `--all-contexts` I still want it to detect as many invalid lockdirs as it can. I'll change this section to treat a lockdir that can't be parsed as the same type of error as an invalid lockdir.

bin/pkg/validate_lock_dir.ml
gridbugs: Good point about scripts. I'll update it to print a message but exit with success.

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
gridbugs: Indeed they are flipped due to representing dependencies with a map.

> I am a bit worried that if this is due to using a map, we might get non-reproducible lockfiles due to using an unordered map. Maybe it would make sense to sort them just to be sure.

Non-reproducible how? All lockfiles generated by dune will have their dependencies in order (following this PR) due to using a map.

rgrinberg: Note that we should preserve the order in which the dependencies appear in the opam file. Packages can define the same binaries (unlike in opam) and the order would decide the shadowing rules.

test/blackbox-tests/test-cases/pkg/lockdir-validation.t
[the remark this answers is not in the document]
Leonidas-from-XIV: Yeah, I agree, after I wrote the comment I saw the code that computes it and indeed the validation comes sort of "for free".

I don't know, what do the others think?

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
gridbugs: Although thinking more about this I think it's a bit dangerous to encode meaning in the order of dependencies. I know I have a habit of sorting dependencies alphabetically in other build systems and it would be a nasty surprise for users if doing so had an effect on the build. Is this behaviour specific to lockfiles or is it established already in other parts of dune?

test/blackbox-tests/test-cases/pkg/lockdir-validation.t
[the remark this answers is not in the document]
gridbugs: To preserve the order of dependencies I'll be changing the representation of lockfile dependencies back to a list so there's no structural reason to not allow duplicates. I'll get rid of this check for now.

test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
rgrinberg: It's inherited form OCaml. For example, the linking order of libraries determines the order of evaluation of module effects. It's certainly not ideal, so that's why we don't re-order things unless it's really needed.

src/dune_pkg/lock_dir.ml
```diff
@@ -705,3 +705,32 @@ module Load_immediate = Make_load (struct
...
+
+let transitive_dependency_closure t start =
+  let missing_packages =
+    let all_packages_in_lock_dir = Package_name.Set.of_keys t.packages in
+    Package_name.Set.diff start all_packages_in_lock_dir
+  in
+  match Package_name.Set.is_empty missing_packages with
+  | false -> Error (`Missing_packages missing_packages)
+  | true ->
+    let to_visit = Queue.create () in
```
rgrinberg, line 717: There's a `Top_closure` module you could take a look at. Perhaps it's usable here.

src/dune_pkg/package_universe.mli
```diff
@@ -0,0 +1,26 @@
...
+(** All of the packages in a dune project including local packages and packages
+    in a lockdir. The lockdir is guaranteed to contain a valid dependency
+    solution for the local packages. *)
+type t = private
+  { local_packages : Local_package.t Package_name.Map.t
+  ; lock_dir : Lock_dir.t
+  }
+
+module Error : sig
+  type t =
```
rgrinberg, line 12: Do we distinguish this error anywhere? If not, just making it a `User_message.t` is simpler

rgrinberg (approved):

src/dune_pkg/package_universe.mli
gridbugs: I wanted to make explicit in the api the different ways that a lockdir can be invalid w.r.t local packages.

rgrinberg: Sure, it's already explicit in the implementation. The callers clearly don't care (at the moment), so might as well hide that detail from them.

src/dune_pkg/lock_dir.ml
gridbugs: It does look handy. I had a look for something like this a while back but didn't find it (I searched for names like "dag"). I'll merge this as is but look at switching to `Top_closure` in a later change. One possible issue is that `Top_closure` doesn't allow for dependency cycles but my understanding is that opam allows dependencies cycles as long as some of the deps are marked as `post`. Is that right? I've been intentionally allowing for dependency cycles in the dependency graph traversals I've implemented thus far for this reason.

rgrinberg: We actually ignore post deps for precisely this reason. Dune's engine cannot allow for cycles so we can assume that they aren't an issue.

PR #9174, opened 2023-11-14
fix(pkg): proper error messages for bad config files

rgrinberg:


test/blackbox-tests/test-cases/pkg/variables.t
```diff
@@ -40,3 +40,23 @@ Test that we can set variables
...
+  >  ))
+  > EOF
+
+  $ build_pkg test 2>&1 | sed 's/File .*:/File $REDACTED:/'
+  Error:
+  File $REDACTED:
+  1 | this is dummy text
+           ^^
+  Error parsing test.config
+  Reason: Parse error
```
Alizter, line 60: This seems kind of redundant. Would it make sense to drop the reason?

PR #9189, opened 2023-11-15
melange: repro of paths issues with include_subdirs

jchavarri:
A repro for issue #9190.


anmonteiro (approved):
yup, this is a bug. taking a look

PR #9191, opened 2023-11-15
fix(pkg): Avoid IO when determining the default branch

Leonidas-from-XIV:
This moves the default branch determination to when the remote is first added, it determines the HEAD branch and uses this sole branch as a tracking branch.

When loading a repo, it determines the default branch from the list of tracking branches (of which there is now only one, thus it must be the default).

This approach that we came up with @Alizter makes it possible to avoid network I/O when loading remotes so if we make the `update` optional (future PR) we can lock without any network IO, only using the cached git revisions.

### DIFF
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -150,23 +150,7 @@ let of_git_repo ~repo_id ~source =
       at_rev, Some repo_id
     | None ->
       let+ at_rev =
-        let* name =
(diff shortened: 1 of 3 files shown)


rgrinberg (approved):

Leonidas-from-XIV:
(Small cleanup with only one change - one of the errors is a code error: it is not the users fault if we set up the rev-store in an invalid way)

PR #9194, opened 2023-11-15
test: formatting settings from context

rgrinberg:
@emillon is this behavior intentional? It seems like `env` settings for formatting in the workspace file are always ignored.


rgrinberg:
ping on this one @emillon I have a fix for this if it's indeed a bug

rgrinberg:
I'm going to assume this was unintentional as it makes context and workspace settings useless

emillon:
Yes, I don't think it's intentional indeed.

PR #9202, opened 2023-11-16
feature(pkg): Implement offline mode for Rev Store

Leonidas-from-XIV:
This allows the rev store to be loaded without doing a `git fetch` thus allowing offline functionality as long as the cache is populated.

The behavior is opt-in to not lead to people accidentally locking old versions.

### DIFF
--- a/bin/pkg/lock.ml
+++ b/bin/pkg/lock.ml
@@ -37,6 +37,7 @@ let solve
   per_context
   ~opam_repository_path
   ~opam_repository_url
+  ~update_opam_repositories
   ~sys_bindings_from_current_system
   ~experimental_translate_opam_filters
   =
@@ -67,7 +68,12 @@ let solve
                   ~sys_bindings_from_current_system)
          in
          let* repos =
-           get_repos repos ~opam_repository_path ~opam_repository_url ~repositories
+           get_repos
+             repos
+             ~opam_repository_path
+             ~opam_repository_url
+             ~repositories
+             ~update_opam_repositories
          in
          let overlay =
            Console.Status_line.add_overlay (Constant (Pp.text "Solving for Build Plan"))
@@ -128,6 +134,7 @@ let lock
   ~version_preference
   ~opam_repository_path
   ~opam_repository_url
+  ~update_opam_repositories
   ~experimental_translate_opam_filters
   =
   let open Fiber.O in
@@ -145,6 +152,7 @@ let lock
     per_context
     ~opam_repository_path
     ~opam_repository_url
+    ~update_opam_repositories
     ~sys_bindings_from_current_system
     ~experimental_translate_opam_filters
 ;;
@@ -186,6 +194,15 @@ let term =
              enabled by default but is currently opt-in as we expect to make major \
              changes to it in the future. Without this flag all conditional commands and \
              terms in Opam files are included unconditionally.")
+  and+ skip_update =
+    Arg.(
+      value
+      & flag
+      & info
+          [ "skip-update" ]
+          ~doc:
+            "Do not fetch updates of opam repositories, will use the cached opam \
+             metadata. This allows offline use if the repositories are cached locally.")
   in
   let builder = Common.Builder.forbid_builds builder in
   let common, config = Common.init builder in
@@ -197,6 +214,7 @@ let term =
       ~version_preference
       ~opam_repository_path
       ~opam_repository_url
+      ~update_opam_repositories:(not skip_update)
       ~experimental_translate_opam_filters)
 ;;
 
--- a/src/dune_pkg/opam_repo.mli
+++ b/src/dune_pkg/opam_repo.mli
@@ -21,8 +21,16 @@ val of_opam_repo_dir_path
   -> Path.t
   -> t
 
-(** [of_git_repo git source] loads the data through git *)
-val of_git_repo : repo_id:Repository_id.t option -> source:string -> t Fiber.t
+(** [of_git_repo git ~repo_id ~update ~source] loads the opam repository located at [source] from git.
+    [source] can be any URL that [git remote add] supports.
+
+    Set [update] to true to update the source to the newest revision, otherwise it will use the latest
+    data available in the cache (if any). *)
+val of_git_repo
+  :  repo_id:Repository_id.t option
+  -> update:bool
+  -> source:string
+  -> t Fiber.t
 
 val repo_id : t -> Repository_id.t option
 val source : t -> string option
--- a/src/dune_pkg/rev_store.mli
+++ b/src/dune_pkg/rev_store.mli
@@ -21,10 +21,21 @@ module At_rev : sig
 end
 
 module Remote : sig
+  (** [uninit] represents an uninitialized remote. Use [update] or
+      [don't_update] to get a remote [t] *)
+  type uninit
+
+  (** handle representing a particular git repository *)
   type t
 
   val equal : t -> t -> bool
-  val update : t -> unit Fiber.t
+
+  (** [update remote] will fetch the most current revisions from the remote *)
+  val update : uninit -> t Fiber.t
+
+  (** [don't_update] signals that the remote should not be updated *)
+  val don't_update : uninit -> t
+
   val default_branch : t -> string
   val rev_of_name : t -> name:string -> At_rev.t option Fiber.t
   val rev_of_repository_id : t -> Repository_id.t -> At_rev.t option Fiber.t
@@ -32,4 +43,12 @@ end
 
 val content_of_files : t -> File.t list -> string list Fiber.t
 val load_or_create : dir:Path.t -> t Fiber.t
-val add_repo : t -> source:string -> Remote.t Fiber.t
+
+(** [add_repo t ~source] idempotently registers a git repo to the rev store.
+    [source] is any URL that is supported by [git remote add].
+
+    This only adds the remote metadata, to get a remote you need to either
+    use [Remote.update] if you want to fetch from the remote (thus potentially
+    triggering network IO) or if you are sure the [t] already contains all
+    required revisions (e.g. from a previous run) then use [don't_update]. *)
+val add_repo : t -> source:string -> Remote.uninit Fiber.t
--- a/test/blackbox-tests/test-cases/pkg/opam-repository-download.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-repository-download.t
@@ -13,7 +13,7 @@ Make a mock repo tarball that will get used by dune to download the package
   $ cd mock-opam-repository
   $ git init --quiet
   $ git add -A
-  $ git commit -m "Initial commit" > /dev/null
+  $ git commit -m "Initial commit" --quiet
   $ REPO_HASH=$(git rev-parse HEAD)
   $ cd ..
   $ mkdir fake-xdg-cache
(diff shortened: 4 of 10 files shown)


bin/pkg/lock.ml
```diff
@@ -37,6 +37,7 @@ let solve
   per_context
   ~opam_repository_path
   ~opam_repository_url
+  ~update
```
Alizter, line 40: Perhaps a more descriptive nae like `update_opam_repositories` would be better.

src/dune_pkg/opam_repo.mli
```diff
@@ -21,8 +21,12 @@ val of_opam_repo_dir_path
   -> Path.t
   -> t
 
-(** [of_git_repo git source] loads the data through git *)
-val of_git_repo : repo_id:Repository_id.t option -> source:string -> t Fiber.t
+(** [of_git_repo git ~repo_id ~update ~source] loads the opam repository from git. *)
+val of_git_repo
+  :  repo_id:Repository_id.t option
+  -> update:bool
+  -> source:string
```
gridbugs, line 28: can you document what `source` means or use a type other than `string`

gridbugs (approved):

src/dune_pkg/rev_store.mli
```diff
@@ -32,4 +32,8 @@ end
 
 val content_of_files : t -> File.t list -> string list Fiber.t
 val load_or_create : dir:Path.t -> t Fiber.t
-val add_repo : t -> source:string -> Remote.t Fiber.t
+
+(** [add_repo t ~update ~source] idempotently adds a git repo to the rev store.
+    If [update] is [false], it avoids fetching the remote repo, thus if the repo
+    was already previously added this can be done without network I/O *)
+val add_repo : t -> update:bool -> source:string -> Remote.t Fiber.t
```
rgrinberg, line 39: For a good API, I think updating should be a separate function from adding a repository.

One of the issues that we've ran into with an opam API is that it tries to cram every type behavior into a function flag. I'd like to avoid that and keep smaller, simpler functions that can be composed.

src/dune_pkg/opam_repo.mli
Alizter: That's meant to be the url or path I believe.

Leonidas-from-XIV: Good point. I went with documenting it because besides doing a dummy type that represents "whatever `git remote add` accepts" with very little validation that we can do on our side.

Leonidas-from-XIV: I will probably be reworking this in the near future to have a source specification format with repo/branch/tags.

src/dune_pkg/rev_store.mli
Leonidas-from-XIV: That's a good point, I removed it from the rev store. I kept it mostly so users of the API don't accidentally forget to run `update`. However, I ended up solving it via types in a nicer way.

src/dune_pkg/opam_repo.mli
```diff
@@ -21,8 +21,16 @@ val of_opam_repo_dir_path
   -> Path.t
   -> t
 
-(** [of_git_repo git source] loads the data through git *)
-val of_git_repo : repo_id:Repository_id.t option -> source:string -> t Fiber.t
+(** [of_git_repo git ~repo_id ~update ~source] loads the opam repository located at [source] from git.
+    [source] can be any URL that [git remote add] supports.
```
Alizter, line 25: Not just URLs right? Local paths should also work.

test/expect-tests/dune_pkg/rev_store_tests.ml
```diff
@@ -50,13 +50,19 @@ let%expect_test "adding remotes" =
     let remote_path = Path.relative cwd "git-remote" in
     let* () = create_repo_at remote_path in
     let source = Path.to_string remote_path in
-    let* _remote = Rev_store.add_repo rev_store ~source in
+    let* remote = Rev_store.add_repo rev_store ~source in
+    let* (_ : Rev_store.Remote.t) = Rev_store.Remote.update remote in
     print_endline "Creating first remote succeeded";
```
Alizter, line 55: You can put an `[%expect]` here and below the other places you print to get more immediate feedback in the test without having to decipher it all at the end.

test/blackbox-tests/test-cases/pkg/opam-repository-download.t
```diff
@@ -77,3 +77,30 @@ The repository can also be injected via the dune-workspace file
...
   $ grep "git_hash $REPO_HASH" dune.lock/lock.dune > /dev/null
+
+A new packages is released in the repo:
+
+  $ mkpkg bar 1.0.0 <<EOF
+  > depends: [ "foo" ]
+  > EOF
+  $ cd mock-opam-repository
+  $ git add -A
+  $ git commit -m "bar.0.1.0" > /dev/null
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Ah, cool. It's `--quiet` but yes.

src/dune_pkg/rev_store.mli
```diff
@@ -21,15 +21,33 @@ module At_rev : sig
 end
 
 module Remote : sig
+  (** [uninit] represents an uninitialized remote. Use [update] or
+      [don't_update] to get a remote [t] *)
+  type uninit
```
rgrinberg, line 26: I wonder why do we need this type?

src/dune_pkg/rev_store.mli
```diff
@@ -21,15 +21,33 @@ module At_rev : sig
...
+  (** handle representing a particular git repository *)
   type t
 
   val equal : t -> t -> bool
-  val update : t -> unit Fiber.t
+
+  (** [update remote] will fetch the most current revisions from the remote *)
+  val update : uninit -> t Fiber.t
+
+  (** [don't_update] signalizes that the remote should not be updated *)
```
gridbugs, line 36:
```suggestion
  (** [don't_update] signals that the remote should not be updated *)
```

src/dune_pkg/rev_store.mli
gridbugs: It looks like it's so people don't forget to update the remote (returned by `add_repo`) or explicitly declare that they won't update it (`don't_update`).

src/dune_pkg/rev_store.mli
```diff
@@ -21,15 +21,33 @@ module At_rev : sig
...
 end
 
 val content_of_files : t -> File.t list -> string list Fiber.t
 val load_or_create : dir:Path.t -> t Fiber.t
-val add_repo : t -> source:string -> Remote.t Fiber.t
+
+(** [add_repo t ~source] idempotently registers a git repo to the rev store.
+    [source] is any URL that is supported by [git remote add].
+
+    Remember to run [Remote.update] if you want the remote repository to be
```
gridbugs, line 50: The instructions in this comment are now enforced via `Remote.uninit`. Is it worth mentioning that here?

Alizter: That's only if you use the remote however.

Leonidas-from-XIV: Good point, I wrote it before I enforced it. So I changed to wording a bit to be more clear and give people pointers what to do with a `Remote.uninit`, because it is still helpful to understand.

src/dune_pkg/rev_store.mli
Leonidas-from-XIV: Exactly this. I didn't like that it it was easily possible to accidentally forget to update and then be surprised why the `Rev_store` would never have any content.

Originally started with a phantom type as described in the classic Jane Street post about access control, but given there is no need to ever switch back and forth between the unupdated and the updated state I simplified it, to just have `uninit` and force the API consumer to use `update` or `don't_update` deliberately. It could use a `update_again : t -> t/unit Fiber.t` but given we don't ever do that this can be added if we need it.

Alizter:
This does not appear to work:
```
~/dune/_build/install/default/bin/dune pkg lock --skip-update
fatal: unable to access 'https://github.com/ocaml/opam-repository.git/': Could not resolve host: github.com
```
with Wi-fi off.

src/dune_pkg/rev_store.mli
rgrinberg: As I see it, `Rev_store` is basically just a database. A database isn't limited to being updated or not updated after it is initialized. It can be updated whenever and those operations can be interleaved with other operations. It doesn't seem like a limitation for us at the moment, but it does seem like something gratuitous. In any case, it's not blocking the PR.

Leonidas-from-XIV:
@Alizter Thanks, that was indeed an issue, where we try to determine whether the remote exists but `git remote show <handle>` would try to connect to the remote repo. Unfortunately `git remote show -n` doesn't work to determine whether the remote exists as it will happily do whatever and exit with 0, so now I'm parsing `git remote --verbose` instead.

I've also updated the test to move the repo away in the `--skip-update` test, so even if git attempted to talk to the repo it would fail since the repo is gone (basically simulating being offline, or the repo being down).

Alizter (approved):
Working now!

rgrinberg:
By the way, the workflow that this PR allow is unnatural. Deciding whether a remote is to be updated isn't a property to be set by a command line flag, but is determined by the user's configuration. There's basically two modes:

* If the user specified a commitish, then we check if the corresponding git object is in the store. If it's not, there's no need to update. Otherwise, we update.

* If the user doesn't specify a commitish, then we should always use the latest HEAD of that remote.

In neither of these cases this flag is useful.

Alizter:
@rgrinberg If you don't have an internet connection, how do you check the latest HEAD of the remote?

rgrinberg:
You go to your existing lock directory and copy paste the hash of the opam repository into your workspace.

Alizter:
It seems a little unnecessary to edit sources in order to get around a network issue. It seems more ergonomic to allow the user to disable updates from the CLI rather than editing their existing configuration, given that this is meant to be a temporary work around.

rgrinberg:
Why don't we step back and explain in what situation this flag can actually be useful for. The premise is that the user would like to regenerate the lock directory without an internet connection. Okay, what's the user going to do with this lock directory once it's generated? Nothing can be built until there's an internet connection to fetch the actual sources.

The entire scenario sounds completely artificial and there's no need to provide any "ergonomic" solution for it.

Alizter:
The sources and the opam repository can still be cached however. I disagree that it is an artificial situation. Let me share the use case that I have in mind, and you can determine if this is too unlikely to be worth supporting.

I am working on my ocaml project whilst travelling without an internet connection. On my computer is an existing lock directory, opam repo cache and package cache. I remove a dependency from my project leading me to have to relock my file.

I can do as you suggested and manually target an opam repo hash just so I can locally build or I could do as I suggested and pass a CLI flag to avoid networking. Passing this flag allows me to relock everything and rebuild since everything I need is already fetched.

If you don't think this is worth supporting, then I guess I can't do much about it. But at least I can make my case.

rgrinberg:
> Passing this flag allows me to relock everything and rebuild since everything I need is already fetched.

There's no way to guarantee that everything you need is already fetched. That makes your workflow not usable in general and not recommended to users.

In the rare situation where you somehow know that your new lock directory is going to be buildable without downloading sources, you can go through the process I've outlined above.

Alizter:
Very well. Sorry about suggesting this @Leonidas-from-XIV.

rgrinberg:
The flag itself can be kept. Someone finds it useful and it does no harm when it's not enabled. We should just make sure it interacts with the two workflows in https://github.com/ocaml/dune/pull/9202#issuecomment-1823177053 work correctly.

For example, what is the interaction of setting this flag and then pointing to a commit that doesn't yet exist in the git repo?

gridbugs:
> > Passing this flag allows me to relock everything and rebuild since everything I need is already fetched.
> 
> There's no way to guarantee that everything you need is already fetched. That makes your workflow not usable in general and not recommended to users.
> 
> In the rare situation where you somehow know that your new lock directory is going to be buildable without downloading sources, you can go through the process I've outlined above.

Is this actually that rare? In the situation @Alizter described they removed a dependency and assuming that they built the project at least once beforehand then all the dependencies of the new solution will almost certainly be cached.

Leonidas-from-XIV:
> For example, what is the interaction of setting this flag and then pointing to a commit that doesn't yet exist in the git repo?

The same one as pointing to a commit and being offline I would expect, both are a user error of some sort. But telling dune not to update the repo has the advantage that workflows like a CI locking from a known commit without network access are possible, which they were not before and coupled with a tarball cache the whole packaging can be tested.

Of course, the other option is to specify a opam-repo as path, but then we don't record the repo id in the lockfile, since the repository used is assumed to be unversioned and unreproducible.

PR #9205, opened 2023-11-16
refactor(pkg): reuse our variable types

rgrinberg:

### DIFF
--- a/src/dune_pkg/package_variable.mli
+++ b/src/dune_pkg/package_variable.mli
@@ -3,6 +3,14 @@ open Import
 module Name : sig
   type t
 
+  val to_opam : t -> OpamVariable.t
+  val of_opam : OpamVariable.t -> t
+  val compare : t -> t -> Ordering.t
+  val to_dyn : t -> Dyn.t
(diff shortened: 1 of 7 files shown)


src/dune_pkg/package_variable.mli
```diff
@@ -1,7 +1,12 @@
 open Import
 
 module Name : sig
-  type t
+  type t = OpamVariable.t
```
gridbugs, line 4: Does this need to be opam-specific?

gridbugs (approved):

src/dune_pkg/package_variable.mli
gridbugs: My preference would be to internally represent `Name.t` with an `OpamVariable.t` to maintain compatibility with opam but not expose this fact in the api and make the conversion to opam explicit. This is the convention that most of our pkg code follows and it helps clarify which parts of pkg are opam-specific and which parts are dune's own pkg policy.

Leonidas-from-XIV (approved):

Issue #9154, opened 2023-11-13
Error while building melange for nix flakes development shell


gridbugs:
## Expected Behavior

It should be possible to start a dev shell by running `nix develop`.

## Actual Behavior

Running `nix develop` gives this error:
```
$ nix develop
error: builder for '/nix/store/pjyxkiy3k5jxvjqkh6sah1ipfsr9igw3-ocaml4.14.1-melange-dev.drv' failed with exit code 1;
       last 10 log lines:
       >    $ dune build @mel --display=short
       >          ocamlc .ppx/df8d19d63a600c50cc2fede8aa971a95/dune__exe___ppx.{cmi,cmo}
       >        ocamlopt .ppx/df8d19d63a600c50cc2fede8aa971a95/build_info__Build_info_data.{cmx,o}
       > -         refmt x.re.ml
       >        ocamlopt .ppx/df8d19d63a600c50cc2fede8aa971a95/dune__exe___ppx.{cmx,o}
       > +         refmt x.re.ml
       >        ocamlopt .ppx/df8d19d63a600c50cc2fede8aa971a95/ppx.exe
       >             ppx x.re.pp.ml
       >            melc .output.mobjs/melange/melange__X.{cmi,cmj,cmt}
       >
       For full logs, run 'nix log /nix/store/pjyxkiy3k5jxvjqkh6sah1ipfsr9igw3-ocaml4.14.1-melange-dev.drv'.
error: 1 dependencies of derivation '/nix/store/bbyidqs08yxyykikvhb2aiav67drl937-nix-shell-env.drv' failed to build
```

Full output of `nix log <hash>-ocaml4.14.1-melange-dev.drv`: https://gist.github.com/gridbugs/681ec26e516e56fca9acb57c4331cba3

emillon:
@gridbugs should this be closed now that #9206 is merged?

PR #9206, opened 2023-11-16
fix(nix): fix melange build and update flake

Alizter:
We need to use melange 1.0 because later versions don't work with OCaml 4.14. Annoyingly, this version of melange has a test that is run by the flake which is non reproducible between differing versions of Dune due to the order of execution. This leads to the failure observed in https://github.com/ocaml/dune/issues/9154.

Updating melange is not an option, so I've gone ahead and forked melange 1.0 and manually removed the offending test. This lets `nix develop` work again which was broken for a while.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,46 +1,15 @@
 {
   "nodes": {
-    "flake-compat": {
-      "flake": false,
-      "locked": {
-        "lastModified": 1627913399,
-        "narHash": "sha256-hY8g6H2KFL8ownSiFeMOjwPC8P0ueXpCVEbxgda3pko=",
-        "owner": "edolstra",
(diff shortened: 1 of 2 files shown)


anmonteiro:
Isn't `melange.overrideAttrs (_: { doCheck = false; })` enough?

Alizter:
@anmonteiro I seem to recall trying that before and it not working. Let me give it another go.

Alizter:
@anmonteiro Is there any issue with continuing like this from your end?

anmonteiro:
feel free to go ahead. We should remember to update Melange once Dune runs its tests on 5.1 instead.

Alizter:
ping @rgrinberg

PR #9212, opened 2023-11-17
chore: update dune-project version to 3.11

Alizter:
In order to use (cram (runtest_alias ...)) we need to bump our dune-project version.


Alizter:
Not sure how to update the bench.

PR #9213, opened 2023-11-17
test: duplicate rules being loaded on rebuild

Alizter:
Here is a minimal test case for a bug that occurs when using watch mode.

@emillon I would suggest that this is blocking for 3.12 as it renders watch mode unusable.

### DIFF
--- a/src/dune_rules/gen_rules.ml
+++ b/src/dune_rules/gen_rules.ml
@@ -398,6 +398,7 @@ let gen_project_rules =
     ()
   in
   fun sctx source_dir ->
+    let* () = Memo.return () in
     let project = Source_tree.Dir.project source_dir in
     match
       Path.Source.equal (Source_tree.Dir.path source_dir) (Dune_project.root project)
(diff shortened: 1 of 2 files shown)


Alizter:
I managed to bisect this to 18789795216fa22c21ee6f2542a59888cfffd360 from #9160.

rgrinberg:
Thanks for the repro

src/dune_rules/gen_rules.ml
```diff
@@ -398,6 +398,7 @@ let gen_project_rules =
     ()
   in
   fun sctx source_dir ->
+    let* () = Memo.return () in
```
emillon, line 401: Was this intended to be merged?

emillon:
> I would suggest that this is blocking for 3.12 as it renders watch mode unusable.

No problem for these kinds of fixes since you've identified it as a regression.
Though in the future I would recommend opening a separate issue for the problem itself. Here we have the bug report, the repro and the fix in the same PR, which is harder to work with.

PR #9215, opened 2023-11-17
fix(memo): delay implicit output collector

### LINKED ISSUE #9225: multiple rules generated for ccomp
Watch mode in the dune repo produces this:

```
Error: Multiple rules generated for _build/default/.dune/ccomp/ccomp:                                                                                                                                 
- <internal location>                                                                                                                                                                                 
- <internal location>                                                                                                                                                                                 
-> required by _build/default/otherlibs/dune-rpc-lwt/test/realpath_stubs.o                                                                                                                            
-> required by alias otherlibs/dune-rpc-lwt/test/check
```

This is similar to https://github.com/ocaml/dune/pull/9213.

### LINKED ISSUE #9226: multiple rules generated for .opam in watch mode
```
Error: Multiple rules generated for _build/default/xdg.opam:                                                                                                                                          
- <internal location>                                                                                                                                                                                 
- <internal location>                                                                                                                                                                                 
```

similar to
- https://github.com/ocaml/dune/issues/9225
- https://github.com/ocaml/dune/pull/9213

rgrinberg:
When doing:

```
let x = Memo.Implicit_output.collect (fun () -> ..) in
let* _ = x in
let* _ = x in
```

Would share the collecting cell between the binds. That's inconsistent
with delaying the side effects until binds.

### DIFF
--- a/src/memo/implicit_output.ml
+++ b/src/memo/implicit_output.ml
@@ -104,21 +104,22 @@ let produce_opt t v =
 ;;
 
 let collect (type o) (type_ : o t) f =
-  (* If we don't delay the computation here, [output] becomes shared between future runs
-     of the resulting fiber, causing the collected output to be duplicated. *)
-  let* () = Fiber.return () in
-  let output = ref None in
-  Fiber.map
-    (Fiber.Var.set
-       current_handler
-       (module struct
-         type nonrec o = o
+  (* If we don't delay the computation here, [output] becomes shared between
+     future runs of the resulting fiber, causing the collected output to be
+     duplicated. *)
+  Fiber.of_thunk (fun () ->
+    let output = ref None in
+    Fiber.map
+      (Fiber.Var.set
+         current_handler
+         (module struct
+           type nonrec o = o
 
-         let type_ = type_
-         let so_far = output
-       end)
-       f)
-    ~f:(fun res -> res, !output)
+           let type_ = type_
+           let so_far = output
+         end)
+         f)
+      ~f:(fun res -> res, !output))
 ;;
 
 let forbid f = Fiber.Var.unset current_handler f
--- a/src/memo/implicit_output.ml
+++ b/src/memo/implicit_output.ml
@@ -104,6 +104,8 @@ let produce_opt t v =
 ;;
 
 let collect (type o) (type_ : o t) f =
+  (* If we don't delay the computation here, [output] becomes shared between future runs
+     of the resulting fiber, causing the collected output to be duplicated. *)
   let* () = Fiber.return () in
   let output = ref None in
   Fiber.map
--- a/src/memo/implicit_output.ml
+++ b/src/memo/implicit_output.ml
@@ -104,6 +104,7 @@ let produce_opt t v =
 ;;
 
 let collect (type o) (type_ : o t) f =
+  let* () = Fiber.return () in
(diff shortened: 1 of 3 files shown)


rgrinberg:
@snowleopard I don't think this affects you, b/c you don't use implicit output, but it's a bit of a footgun.

snowleopard (approved):
Looks good but I think internally we do rely on implicit output, so I'd prefer to test/benchmark this change internally first. Please hold on with merging.

Alizter:
Fixes both bugs I observed:
- https://github.com/ocaml/dune/issues/9225
- https://github.com/ocaml/dune/issues/9226

emillon:
Is it possible to add a test for this?

rgrinberg:
Added it here https://github.com/ocaml/dune/pull/9247

emillon:
@snowleopard have you been able to test this internally? This is the last issue blocking the release of 3.12.0.

snowleopard:
> @snowleopard have you been able to test this internally? This is the last issue blocking the release of 3.12.0.

Started the benchmarks, will report results later today. Sorry for the delay!

src/memo/implicit_output.ml
```diff
@@ -104,6 +104,7 @@ let produce_opt t v =
 ;;
 
 let collect (type o) (type_ : o t) f =
+  let* () = Fiber.return () in
```
snowleopard, line 107:
```suggestion
  (* If we don't delay the computation here, [output] becomes shared between future runs
     of the resulting fiber, causing the collected output to be duplicated. *)
  let* () = Fiber.return () in
```

snowleopard: Also, let's use `Fiber.of_thunk` instead, which is a more explicit mechanism.

emillon:
I have a quick question. In many places we have `let* () = Memo.return () in` and I'm never too sure about what this means. Does this change to `Memo` itself mean that we can remove (some of) these lines?

test/expect-tests/memo/main.ml
```diff
@@ -2164,5 +2164,5 @@ let%expect_test "implicit output bug. storage cell is reused between runs" =
   set_invalidate_print_run ();
   [%expect {| Some [ "x" ] |}];
   set_invalidate_print_run ();
-  [%expect {| Some [ "x"; "x" ] |}]
+  [%expect {| Some [ "x" ] |}]
```
snowleopard, line 2167: Let's also edit the name of the test now that it's now a bug. Internally, I wrote this:

```
let%expect_test "Ensure that implicit output storage cell is not reused between runs" =
```

emillon:
Thanks for the explanation.

emillon:
I'll address the review and merge.

rgrinberg:
One reason I write `let*() = Memo.return () in` instead of `Memo.of_thunk` is because the latter indents the entire of the function to the right. In this case, it doesn't really matter, but sometimes it causes a bunch of ugly wrapping.

snowleopard:
> One reason I write `let*() = Memo.return () in` instead of `Memo.of_thunk` is because the latter indents the entire of the function to the right. In this case, it doesn't really matter, but sometimes it causes a bunch of ugly wrapping.

Yeah, it's a little annoying but I think I still prefer the explicit version despite the formatting awkwardness.

PR #9216, opened 2023-11-17
`dune cache trim`: also print number of files removed

nojb:
What it says on the tin. May be useful to get some more insight when `dune cache trim` takes a long time. cc @snowleopard


snowleopard (approved):
Looks good, thanks!

PR #9219, opened 2023-11-18
refactor: annotate some lazy values to simplify debugging

rgrinberg:


Alizter (approved):

Alizter:
Do you think this is a good practice that should always be done? We could add a note in hacking.md or Memo.Lazy about it. Even making the optional arg required might help.

rgrinberg:
don't see a reason to always do it, but it's a good first step to debug cycles. That might be mentioned in the hacking document on how to debug cycles

PR #9221, opened 2023-11-19
chore: update vendored opam

Alizter:
I've also modified the vendor script slightly so that we do a more efficient "git clone" by simply `fetch`ing the commit we want straight from the remote and checking it out locally. Downloading opam takes a long time, so I think this is a good idea.

We can eventually consolidate a lot of the common functions in the vendor update scripts and do the same everywhere.

### DIFF
--- a/vendor/update-opam.sh
+++ b/vendor/update-opam.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-version=dcb5165badc5fe87f7f34518c13b7d8e69a79f6f
+version=0984de6a2def6ddf2b3beab7857c062db7e0319d
 
 set -e -o pipefail
 


vendor/update-opam.sh
```diff
@@ -12,8 +12,10 @@ mkdir -p opam/src
 
 (
     cd $TMP
-    git clone https://github.com/ocaml-dune/opam.git
+    mkdir opam
     cd opam
+    git init 
+    git fetch https://github.com/ocaml-dune/opam.git $version --depth 1
```
rgrinberg, line 18: Can you revert this? All the vendoring scripts should have the same pattern

Alizter:
@rgrinberg I've reverted the cloning changes. I still think its a good idea, but I'll do them in another PR with some more sharing between the scripts so that we don't have so much copy-paste and they are fatser. The reason I noticed with opam is because the download is very large and I have a slow internet connection at the moment.

PR #9224, opened 2023-11-20
test: demonstrate a bunch of bugs in path.ml

Alizter:
We demonstrate a bunch of existing bugs in path.ml with respect to windows paths. Mainly `C:\foo\bar` is not recognized as an absolute directory, causing issues later on.

### DIFF
--- a/otherlibs/stdune/test/path_tests.ml
+++ b/otherlibs/stdune/test/path_tests.ml
@@ -69,10 +69,26 @@ let%expect_test _ =
   |}]
 ;;
 
+(* CR-someday alizter: This is a bug *)
+let%expect_test _ =
+  is_descendant (r "foo") ~of_:(r "foo\\");
+  [%expect {|
+  false
+  |}]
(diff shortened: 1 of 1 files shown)


nojb:
> Mainly `C:\foo\bar` is not recognized as an absolute directory, causing issues later on.

What environment (compiler) did you use to build Dune and which environment are you running this on?

nojb:
> This is all on linux, does it matter which one is used?

Indeed! The `Filename` module of the OCaml standard library depends on the system: you will only get correct answers if you run on a matching system. In other words, you need to use one of the native Windows ports (MSVC or Mingw64) if you want `C:\`-style paths to be correctly recognized.

Alizter:
Hmm in which case I'll move these to a separate module and make this a windows only test. Thanks!

Alizter:
I couldn't work out how to run an expect test only on windows so I've pretty much abandoned this PR.

PR #9228, opened 2023-11-20
test: demo dupe .opam rule

Alizter:
Repro case for #9226


Alizter:
Windows failure is unrelated.

rgrinberg:
How about combining them all to a single test file? They're all basically the same issue

Alizter:
Done. See #9229.

PR #9231, opened 2023-11-20
fix(coq): unescape \: to :

### LINKED ISSUE #9218: dune does not recognise paths starting with `C:/` as absolute on Windows
## Expected Behavior

dune produces correct paths in Windows cygwin MinGW cross environments

## Actual Behavior

Path normalization in dune produces invalid paths. The likely cause is that paths starting with C:/ are not recognized as absolute paths. Please note that Windows file APIs do accept / as path separator (since pretty much ever) without any pre processing (say by MinGW C library). So dune should support this as well.

Example paths produced in a Coq Platform 2023.11 / 8.18 build are:
```
src/C/bin/cygwin64_coq_platform/home/Michael/.opam/CP.2023.11.0~8.18~2023.11/lib/coq/theories/extraction/ExtrOcamlString.vo
src/C/bin/cygwin64_coq_platform/home/Michael/.opam/CP.2023.11.0~8.18~2023.11/lib/coq/theories/ZArith/ZArith.vo
theories/C/bin/cygwin64_coq_platform/home/Michael/.opam/CP.2023.11.0~8.18~2023.11/lib/coq-core/plugins/ltac/ltac_plugin.cmxs
```
these paths result in error messages like:
```
# 10 |  (modules :standard)
# 11 |  (flags -noinit -indices-matter -color on))
# Error: No rule found for
# theories/C/bin/cygwin64_coq_platform/home/Michael/.opam/CP.2023.11.0~8.18~2023.11/lib/coq-core/plugins/ltac/ltac_plugin.cmxs
```

A few notes:
- I am using a cygwin MinGW cross build environment. The produced dune is a MinGW exe, not a cygwin exe.
- paths usually use a mix of / and \ as path separators in such an environment

## Reproduction

- follow these [instructions](https://github.com/coq/platform/blob/main/doc/README_Windows.md) to setup a Coq Platform build environment on Windows - select extent "basic".
- open a shell of the newly created cygwin (which has opam)
- opam install dune 3.10.0 and any of these packages:
  - coq-simple-io.1.8.0
  - coq-mathcomp-word.2.1
  - coq-hott.8.18

## Specifications

```
$ dune --version
3.10.0

$ ocamlc --version
4.14.1

cygwin 64 bit on Windows 10, MinGW cross (that is using https://github.com/ocaml-opam/opam-repository-mingw)
```

Alizter:
coqdep outputs escaped paths for makefile which means on Windows that `C:\foo\bar` gets escaped to `C\:\\foo\\bar` causing Dune to interpret escaped absolute Windows directories as relative ones.

This patch searches for `\:` in the deps output of coqdep and replaces it with `:` allowing the paths to be interpreted correctly.

I couldn't write any tests for this since we don't run Coq tests on Windows and I don't plan to.

fix #9218

### DIFF
--- a/src/dune_rules/coq/coq_rules.ml
+++ b/src/dune_rules/coq/coq_rules.ml
@@ -450,6 +450,15 @@ let coqdep_invalid phase line =
     [ "phase", Dyn.string phase; "line", Dyn.string line ]
 ;;
 
+(* Handle the case where the path contains ":" and coqdep escapes this
+   as "\:" causing Dune to misinterpret the path. We revert
+   the escaping, which allows dune to work on Windows.
+
+   Note that coqdep escapes a few more things, including spaces, $, #,
+   [], ?, %, homedir... How to handle that seems tricky.
+*)
+let unescape_coqdep string = Re.replace_string (Re.compile (Re.str "\\:")) ~by:":" string
+
 let parse_line ~dir line =
   match String.lsplit2 line ~on:':' with
   | None -> coqdep_invalid "split" line
@@ -463,7 +472,9 @@ let parse_line ~dir line =
     in
     (* let depname, ext = Filename.split_extension ff in *)
     let target = Path.relative (Path.build dir) target in
-    let deps = String.extract_blank_separated_words deps in
+    (* EJGA: XXX using `String.extract_blank_separated_words` works
+       for OCaml, but not for Coq as we don't use `-modules` *)
+    let deps = unescape_coqdep deps |> String.extract_blank_separated_words in
     (* Add prelude deps for when stdlib is in scope and we are not actually
        compiling the prelude *)
     let deps = List.map ~f:(Path.relative (Path.build dir)) deps in
--- a/doc/changes/9231_coq.md
+++ b/doc/changes/9231_coq.md
@@ -1,0 +1,2 @@
+- Fixed a bug where Dune was incorrectly parsing the output of coqdep when it was escaped,
+  as is the case on Windows. (#9231, fixes #9218, @Alizter)


Alizter:
@MSoegtropIMC I've tested this and it is able to build `coq-mathcomp-analysis`, `coq-mathcomp-word` and `coq-hott`.

src/dune_rules/coq/coq_rules.ml
```diff
@@ -448,7 +448,14 @@ let parse_line ~dir line =
     in
     (* let depname, ext = Filename.split_extension ff in *)
     let target = Path.relative (Path.build dir) target in
-    let deps = String.extract_blank_separated_words deps in
+    let deps =
+      (* Before we split the deps on spaces, we do some preprocessing to handle the case
+         where the path contains ":" and coqdep escapes this as "\:" causing Dune to
+         misinterpret the path. We simply revert that escaping alllowing this to work on
+         Windows. *)
+      Re.replace_string (Re.compile (Re.str "\\:")) ~by:":" deps
```
ejgallego, line 456: Should we replace `\\` by `\` too?

ejgallego: Should we enable this only on Windows?

ejgallego: Maybe move the logic to its own function `unquote_make_path_on_win` ?

ejgallego (approved):

src/dune_rules/coq/coq_rules.ml
Alizter: I discussed replacing `\\` but decided against it to keep the patch small. It's harmless on Windows anyway.

Alizter: @ejgallego I don't think its worth making into its own function, nor worth enabling only on windows.

Alizter: I don't think it is going to be such a hotpath as we aren't parsing coqdep all the time.

Alizter: Even if we don't use this patch, we could allow the Coq platform maintainers to use it and come up with something better. At the end of the day, we are fighting hacks with hacks so making this slicker seems like dressing a pig. The more correct solution would be to have a better format for coqdep to communicate with Dune.

MSoegtropIMC:
@Alizter : I don't see a CI run triggered by you in Coq Platform. Can you please point me to the CI results?

I will test it today evening with local runs.

Alizter:
@MSoegtropIMC I didn't trigger the Coq Platform CI since I ran out of time to look into it yesterday. I had tested this myself on a Windows VM with Coq platform.

src/dune_rules/coq/coq_rules.ml
ejgallego: I think we should use the patch; it is fine to run it unconditionally, but I think I'd like to have the code modular and have a top-level function `unescape_coqdep_path` so the different implementation concerns are separated.

ejgallego:
This also needs a changelog (do we have a `needs: changes entry` tag?)

Alizter:
@ejgallego I'll add a changelog entry.

ejgallego:
I'm happy with un-escaping `:` only for now (what about spaces in paths?)

Un-escaping logic should be placed into their own function tho.

Alizter:
@ejgallego Feel free to push any improvements here. I won't spend much more time on this for now.

PR #9241, opened 2023-11-21
feature(pkg): Support for specifying particular branches/commits for repos

Leonidas-from-XIV:
This adds support for checking out specific revisions that are different from the default branch. This also refactors the `source` argument which used to be a pretty nondescript string and is not a distinct type that is constructed from an `OpamUrl.t`.

### DIFF
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -34,6 +34,80 @@ module Paths = struct
   let opam_file package = Path.Local.relative (package_dir package) "opam"
 end
 
+module Source = struct
+  type commitish =
+    | Commit of string
+    | Branch of string
+    | Tag of string
+
+  let commitish_to_dyn = function
+    | Commit c -> Dyn.variant "Commit" [ Dyn.string c ]
+    | Branch b -> Dyn.variant "Branch" [ Dyn.string b ]
+    | Tag t -> Dyn.variant "Tag" [ Dyn.string t ]
+  ;;
+
+  let commitish_equal a b =
+    match a, b with
+    | Commit x, Commit x' | Branch x, Branch x' | Tag x, Tag x' -> String.equal x x'
+    | _, _ -> false
+  ;;
+
+  type t =
+    { url : string
+    ; commit : commitish option
+    }
+
+  module Private = struct
+    let of_opam_url rev_store ({ OpamUrl.hash; _ } as opam_url) =
+      (* fairly ugly to pull the rev-store out of thin air *)
+      let url = OpamUrl.base_url opam_url in
+      let+ commit =
+        match hash with
+        | None -> Fiber.return None
+        | Some ref ->
+          (* OpamUrl doesn't distinguish between branches/tags and commits, so we need to look up *)
+          let* member = Rev_store.mem rev_store ~rev:ref in
+          (match member with
+           | true -> Fiber.return @@ Some (Commit ref)
+           | false ->
+             let+ type' = Rev_store.ref_type rev_store ~source:url ~ref in
+             (match type' with
+              | Some `Tag -> Some (Tag ref)
+              | Some `Head -> Some (Branch ref)
+              | None ->
+                User_error.raise
+                  ~hints:
+                    [ Pp.text
+                        "Make sure the URL is correct and the repository contains the \
+                         branch/tag"
+                    ]
+                  [ Pp.textf
+                      "Opam repository at '%s' does not have a reference '%s'"
+                      url
+                      ref
+                  ]))
+      in
+      { commit; url }
+    ;;
+  end
+
+  let of_opam_url opam_url =
+    (* fairly ugly to pull the rev-store out of thin air *)
+    let* rev_store = rev_store in
+    Private.of_opam_url rev_store opam_url
+  ;;
+
+  let to_string { url; commit = _ } = url
+
+  let to_dyn { url; commit } =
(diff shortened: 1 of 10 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -297,8 +303,49 @@ module Remote = struct
   ;;
 end
 
+module Source = struct
```
rgrinberg, line 306: If this type is necessary, it should probably live in the `Opam_repo`. It's not really modelling a property of the revision store as it doesn't care about the association between remotes and commits

Leonidas-from-XIV: I would disagree, in my opinion `Source.t` represents a thing that you can add to the revision store and as the store is backed by Git, it needs to know how to represent things in the revision store and not in the `Opam_repo`. 

One might disagree that the commit that is chosen is not really a feature of the source but rather a choice on what revision to check out, though if the user picks a branch we might want to use that branch as the `default_branch`.

Also, it can't be added to `Opam_repo` as that would create a cyclic dependency, so it would need to live in its own place.

src/dune_pkg/opam_repo.mli
```diff
@@ -13,27 +13,39 @@ end
 
 val equal : t -> t -> bool
 
-(** [of_opam_repo_dir_path opam_repo_dir] creates a repo representedy by a local
+module Source : sig
+  type committish = Commit of string
+
+  type t =
+    { url : string
```
rgrinberg, line 20: What about just using `OpamUrl.t` instead of string? Sure, this type isn't ideal, but it's better than a string.

test/expect-tests/dune_pkg/dune_pkg_unit_tests.ml
```diff
@@ -284,7 +289,7 @@ let%expect_test "encode/decode round trip test for lockdir with complex deps" =
             Some
               [ opam_repo_serializable
                   Some Git_hash "95cf548dc",
-                  "well-known-repo"
+                  { url = "https://github.com/ocaml/dune"; commit = None }
```
rgrinberg, line 292: How about a test case when `commit = Some _` as well?

test/blackbox-tests/test-cases/pkg/opam-repository-download.t
```diff
@@ -57,8 +57,27 @@ other systems and thus shouldn't be included.
...
+We also test that it is possible to specify a specific commit when locking a
+repo, in this case the one from an older revision. So we create a new package
+in the repo and make sure it locks the older version.
 
-The repository can also be injected via the dune-workspace file
+  $ mkpkg foo 0.1.0 <<EOF
+  > EOF
+  $ cd mock-opam-repository
+  $ git add -A
+  $ git commit -m "new release of foo" > /dev/null
```
rgrinberg, line 68: `--quiet`

src/dune_pkg/opam_repo.ml
```diff
@@ -34,33 +34,77 @@ module Paths = struct
...
+    (* CR-someday Leonidas-from-XIV: encode commit and remove repo id? *)
+    let open Dune_sexp.Encoder in
+    record_fields [ field "url" string url ]
+  ;;
+
+  let decode =
+    let open Dune_sexp.Decoder in
+    fields
+    @@ let+ url = field "url" string in
+       { url; commit = None }
```
rgrinberg, line 75: Shouldn't we try to extract the commit from the specified url?

src/dune_pkg/opam_repo.ml
```diff
@@ -34,33 +34,77 @@ module Paths = struct
...
       "opam_repo_serializable"
-      [ option Repository_id.to_dyn repo_id; string source ]
+      [ option Repository_id.to_dyn repo_id; Source.to_dyn source ]
   ;;
 
   let encode { repo_id; source } =
     let open Encoder in
     record_fields
-      [ field "source" string source; field_o "repo_id" Repository_id.encode repo_id ]
+      [ field_l "source" Fun.id (Source.encode source)
```
rgrinberg, line 99: I think this ends up being problematic in the situation where the user specifies a branch, a tag, or a shortened commit hash. We should really record full git object hash in all of these situations.

rgrinberg:
Okay, this is looking better. One thing to keep in mind is that this PR tries to maintain the symmetry between what we record in the lock directory and what we write in the workspace file. That doesn't work because we should only record complete git objects in the lock directory.

Also, we should make sure that we update the repository when "committish" specified is a tag or a branch (both might point to a different commit after an update).

src/dune_pkg/opam_repo.mli
Leonidas-from-XIV: The problem is that an `OpamUrl.t` is not really the kind of URL that we want, since it also includes hashes which are used for branches/commits so I want to have the "convert OpamURL to a thing that the Rev Store understands" in one place. With commits its easy so far, but OpamURL doesn't distinguish between branches and commits so some kind of lookup will be necessary.

Leonidas-from-XIV:
I've added the lookup for branches/tags but yet to add some tests to make sure locking with branches and tags works as expected.

This has the sad side-effect that parsing an `OpamUrl.t` into a `Source.t` returns a fiber, since it needs to potentially connect to the repo to determine whether the hash is a branch or a tag (this could even be normalized into a commit hash, since `ls-remote` also returns that, potentially a cleaner approach).

Leonidas-from-XIV:
@rgrinberg I've updated the tests now.

rgrinberg (approved):

PR #9243, opened 2023-11-21
chore: add lang dune 3.12

emillon:
Nothing is attached to it, but this ensures that we are future proof:
Assuming we do not do that in 3.12.0, (lang dune 3.12) will be implicitly be created when bumping to 3.13. Then `(lang dune 3.12)` would incorrectly create a `>= 3.12.0` opam bound.

### DIFF
--- a/doc/coq.rst
+++ b/doc/coq.rst
@@ -417,7 +417,7 @@ Let us start with a simple project. First, make sure we have a
 
 .. code:: dune
 
-  (lang dune 3.11)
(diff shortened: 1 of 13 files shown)


Alizter:
You should also update the release checklist to make sure we bump the version each time.

Alizter (approved):

emillon:
Yes, I've added that too.

rgrinberg (approved):

Issue #8772, opened 2023-09-26
Building Private Libraries/Executables with Lock Directory


rgrinberg:
Private libraries and executables are those that do not belong to any particular package. However, they still need to be built and may depend on public packages. There should be a way to specify which packages are necessary to build such libraries and executables.

Alizter:
What if we allowed for `(package...)` in a deps like field?

rgrinberg:
Where would the field go?

rgrinberg:
Here's a simple suggestion:

1. Allow users to define a "private" package. That's a package for which we don't setup any install rules
2. Users can now attach additional dependencies/conflicts to this private package to build their private libraries/executables

Alizter:
I was thinking of adding it to the library/executable stanza and make it mutually exclusive with package/public_name.

rgrinberg:
See the discussion in https://github.com/ocaml/dune/pull/8356 for the problems with such an approach

rgrinberg:
I don't think private packages are misleading and they have other uses.

Creating additional stanzas in the dune project file isn't great because it requires more work for conflicts and just about any other thing that can be attached to packages.

Alizter:
I'm confused about the suggestion with private packages then. Where would you declare them?

rgrinberg:
it would be the exact same thing as a normal package, except that it would have `(private)` set inside the dune-project file.

For such a package, there would be installation rules.

Alizter:
How do we handle packages that depend on a private package? Will we need a similar setup like we have for private libraries and executables?

rgrinberg:
For packages, we don't need separate private name and public name namespaces. We can just say that a public package cannot depend on a private package.

Alizter:
Does this mean private packages can depend on other private packages? I feel like this is not a useful behaviour and I would rather have no package being able to depend on private packages.

rgrinberg:
That's inconsistent with how we handle libraries and such dependencies are in fact useful. They can be used to group shared constraints between multiple private packages.

Alizter:
How do we handle `(package ...)` dependencies in the dependency specification? This is really relying on the install layout so my instinct is to forbid this. Currently (with my PR) we get an error like this:
```
  Error: No rule found for alias .foo-files
  -> required by alias foo in dune:1
  [1]
```
since we are avoiding the install rules.

rgrinberg:
[...]

PR #9244, opened 2023-11-21
feature: private packages

Alizter:
We add a (private) field to the (package) stanza allowing for packages to be declared that do not generate any install rules. This is useful for adding private executables and libraries to a package so that their package dependencies can be resolved without generating install rules.

Private executables and libraries can be added to private libraries without needing to make them public.

grants #8772 

- [x] update versioning
- [ ] add documentation
- [ ] changelog
- [x] Add test of public package depending on private package.
- [x] https://github.com/ocaml/dune/pull/9250

### DIFF
--- a/src/dune_rules/dune_file.ml
+++ b/src/dune_rules/dune_file.ml
@@ -1316,6 +1316,11 @@ module Executables = struct
                     project
                     (pluralize "executable" ~multi)
               }
+        | Some (_loc, package), None when package.private_ ->
+          Some
+            { package
+            ; public_names = List.map names ~f:(fun (loc, name) -> loc, Some name)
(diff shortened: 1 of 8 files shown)


src/dune_rules/dune_file.ml
```diff
@@ -1316,6 +1316,11 @@ module Executables = struct
                     project
                     (pluralize "executable" ~multi)
               }
+        | Some (_loc, package), None when package.private_ ->
+          Some
+            { package
+            ; public_names = List.map names ~f:(fun (loc, name) -> loc, Some name)
```
Alizter, line 1322: Should this just be empty?

PR #9252, opened 2023-11-21
fix(pkg): fix downloads on windows

Alizter:
We make the invocation of `curl` Windows-friendly.

### DIFF
--- a/src/dune_pkg/fetch.ml
+++ b/src/dune_pkg/fetch.ml
@@ -2,7 +2,18 @@ open Import
 open Fiber.O
 
 module Curl = struct
-  let bin = lazy (Bin.which ~path:(Env_path.path Env.initial) "curl")
+  let bin =
+    lazy
+      (match Bin.which ~path:(Env_path.path Env.initial) "curl" with
+       | Some p -> p
+       | None ->
+         User_error.raise
+           [ Pp.concat
+               ~sep:Pp.space
+               [ User_message.command "curl"; Pp.text "not available in PATH" ]
+             |> Pp.hovbox
+           ])
+  ;;
 
   let user_agent =
     lazy
@@ -12,30 +23,81 @@ module Curl = struct
        | Some v -> base ^ "." ^ Build_info.V1.Version.to_string v)
   ;;
 
+  module Fiber_lazy : sig
+    type 'a t
+
+    val create : (unit -> 'a Fiber.t) -> 'a t
+    val force : 'a t -> 'a Fiber.t
+  end = struct
+    type 'a t =
+      { value : 'a Fiber.Ivar.t
+      ; mutable f : (unit -> 'a Fiber.t) option
+      }
+
+    let create f = { f = Some f; value = Fiber.Ivar.create () }
+
+    let force t =
+      match t.f with
+      | None -> Fiber.Ivar.read t.value
+      | Some f ->
+        Fiber.of_thunk (fun () ->
+          t.f <- None;
+          let* v = f () in
+          let+ () = Fiber.Ivar.fill t.value v in
+          v)
+    ;;
+  end
+
+  let curl_features_regex =
+    (* If these features are present, then --compressed is supported *)
+    let features = [ "libz"; "brotli"; "zstd" ] in
+    Re.compile
+      Re.(seq [ bol; str "Features:"; rep (first (alt (List.map ~f:str features))) ])
+  ;;
+
+  let compressed_supported =
+    (* We check if curl supports --compressed by running curl -V and checking if
+       the output contains the features we need. *)
+    Fiber_lazy.create (fun () ->
+      let+ lines, _ =
+        let stderr_to =
+          Process.Io.make_stderr
+            ~output_on_success:Swallow
+            ~output_limit:Dune_engine.Execution_parameters.Action_output_limit.default
+        in
+        Process.run_capture_lines
+          Return
+          ~stderr_to
+          ~display:Quiet
+          (Lazy.force bin)
+          [ "-V" ]
+      in
+      match List.find_map lines ~f:(Re.exec_opt curl_features_regex) with
+      | Some group -> Re.Group.test group 0
+      | None -> false)
+  ;;
+
   let run ~url ~temp_dir ~output =
-    let bin =
-      match Lazy.force bin with
-      | Some p -> p
-      | None -> User_error.raise [ Pp.text "curl not available in PATH" ]
-    in
+    let* compressed_supported = Fiber_lazy.force compressed_supported in
     let args =
-      [ "-L"
-      ; "-s"
-      ; "--compressed"
-      ; "--user-agent"
-      ; Lazy.force user_agent
-      ; "--write-out"
-      ; "%{http_code}\\n"
-      ; "-o"
-      ; Path.to_string output
-      ; "--"
-      ; url
-      ]
+      List.flatten
+        [ [ "-L"
+          ; "-s"
+          ; "--user-agent"
+          ; Lazy.force user_agent
+          ; "--write-out"
+          ; "\"%{http_code}\"" (* This arg must be quoted to work on windows *)
+          ; "-o"
+          ; Path.to_string output
+          ]
+        ; (if compressed_supported then [ "--compressed" ] else [])
+        ; [ "--"; url ]
+        ]
     in
     let stderr = Path.relative temp_dir "curl.stderr" in
     let+ http_code, exit_code =
       let stderr_to = Process.Io.file stderr Out in
-      Process.run_capture_line Return ~stderr_to ~display:Quiet bin args
+      Process.run_capture_line Return ~stderr_to ~display:Quiet (Lazy.force bin) args
     in
     if exit_code <> 0
     then (
@@ -56,7 +118,12 @@ module Curl = struct
            ([ Pp.textf "curl returned an invalid error code %d" exit_code ] @ stderr)))
     else (
       Path.unlink_no_err stderr;
-      match Int.of_string http_code with
+      match
+        let open Option.O in
+        let suffix = {|"|} in
+        let prefix = suffix in
+        String.drop_prefix_and_suffix ~prefix ~suffix http_code >>= Int.of_string
+      with
       | None ->
         Error
           (User_message.make


Alizter:
Hmm trying this out locally, I am getting some errors with tar.stderr being expected but not being found. I think I have misunderstood something in the implementation.
```
Error:
open(_build/_private/default/.pkg/ocaml-variants/source/tar.stderr): No such file or directory
-> required by _build/_private/default/.pkg/ocaml-variants/source
-> required by _build/_private/default/.pkg/ocaml-variants/target/cookie
```

rgrinberg:
Did you get an opinion from upstream about this?

Alizter:
@rgrinberg Upstream being opam? I haven't asked anybody else about this, just tested myself on Windows. I'm not sure how opam handles tar, maybe our vendored copy is out of date in that respect. cc @rjbou

src/dune_pkg/fetch.ml
```diff
@@ -24,8 +24,8 @@ module Curl = struct
       ; "--compressed"
       ; "--user-agent"
       ; Lazy.force user_agent
-      ; "--write-out"
-      ; "%{http_code}\\n"
+      ; "--write-out" (* This arg must be quoted to work on windows *)
+      ; "\"%{http_code}\""
```
Leonidas-from-XIV, line 28: Weirdly enough probably due to ocamlformat, this looks like it is attached to the wrong line. I really dislike end-of-line comments :(

src/dune_pkg/fetch.ml
```diff
@@ -56,16 +56,59 @@ module Curl = struct
...
+  let bin = lazy (Bin.which ~path:(Env_path.path Env.initial) "tar")
+
+  let run ~temp_dir input output =
+    let bin =
+      match Lazy.force bin with
+      | Some p -> p
+      | None -> User_error.raise [ Pp.text "tar not available in PATH" ]
+    in
+    let args =
+      [ "-xf"; Path.to_string input; "--force-local"; "-C"; Path.to_string output ]
```
Leonidas-from-XIV, line 85: I think this only works if the copy of `tar` is smart enough to recognize the archive format automatically (GNU tar does, not sure about other implementations), it would probably be safer to specify the appropriate archive option or use the unarchiving tool directly.

src/dune_pkg/fetch.ml
```diff
@@ -56,16 +56,59 @@ module Curl = struct
...
+    in
+    let stderr = Path.relative temp_dir "tar.stderr" in
+    let+ _, exit_code =
+      let stderr_to = Process.Io.file stderr Out in
+      Process.run_capture_lines Return ~stderr_to ~display:Quiet bin args
```
Leonidas-from-XIV, line 90: I wonder if it wouldn't be better to use `Strict` here, that way the whole validation and command output handling will be handled by `Process` in the same consistent way all other processes are handled.

src/dune_pkg/fetch.ml
```diff
@@ -56,16 +56,59 @@ module Curl = struct
            ([ Pp.textf "curl returned an invalid error code %d" exit_code ] @ stderr)))
     else (
       Path.unlink_no_err stderr;
-      match Int.of_string http_code with
+      (* Since we had to quote the return http_code, we need to strip it *)
+      match
+        let open Option.O in
+        String.drop_prefix_and_suffix ~prefix:"\"" ~suffix:"\"" http_code
```
Leonidas-from-XIV, line 62: Here's an alternative, possibly more readable (except apparently to the Github syntax highlighting)

```ocaml
let prefix = {|"|} in
let suffix = prefix in
String.drop_prefix_and_suffix ~prefix ~suffix http_code
```

Leonidas-from-XIV (commented):
I will concur with @rgrinberg, since opam 2.2 is supposed to properly support Windows, isn't this happening with the upstream code as well?

But I don't mind replacing the OPAM extraction code with our own implementation if we can make it work well on all supported platforms, it certainly makes it easier to see what is happening.

src/dune_pkg/fetch.ml
Alizter: Yes, it is in the "wrong" place, but I'm not sure it matters too much as there is only one quoted argument in proximity anyway.

src/dune_pkg/fetch.ml
Alizter: I checked freebsd tar and that appears to also determine the archive format automatically, from what I read in the manual. (At least as long as it is .gz I suppose). This also is the case on windows from testing.

src/dune_pkg/fetch.ml
Alizter: We use return in order to display any stderr we get from the process when we fail to extract. We do the same for curl too. I think giving the original command error in this case is more useful than a blanket Dune one, or even reconstructing the error.

src/dune_pkg/fetch.ml
Alizter: I'll move it anyway.

src/dune_pkg/fetch.ml
Alizter: I noticed that it's easier just to get the substring of the output.

src/dune_pkg/fetch.ml
Alizter: With quiet, we can't know what was outputted since we discarded it.

src/dune_pkg/fetch.ml
Leonidas-from-XIV: Can you also check `.bz2` on BSDs? topkg and dune-release create these (not that I am particularly fond of bzip2 but there is a significant amount of bzipped tarballs).

Alizter: For now, I will leave it as is. I can make a follow up PR where we can do the detection of the archive format by inspecting the header bytes. This is probably better than relying on file extensions. However it would bloat this PR too much to include it here.

rjbou:
> I'm not sure how opam handles tar, maybe our vendored copy is out of date in that respect.

cf [tar handling](https://github.com/ocaml/opam/blob/master/src/core/opamSystem.ml#L911), opam linuxify pathes, as it uses cygwin tar.

For the curl part, if it is portable, it can be upstreamed.

rgrinberg (changes requested):
Also the curl changes need to be split from the tar changes. They should include justification why upstream does or doesn't do this and why.

Alizter:
@rjbou I assume you are also using cygwin curl for opam?

@rgrinberg I've updated this fix to only touch our curl call. The explanation in the commit message is:

```
    opam invokes the cygwin version of curl which means passing % in a unix
    shell is fine. We are however calling the Windows version of curl on
    Windows which means that % is not a valid token in a command line
    enviornment.
    
    To get around this and make it more cross platform, we quote the
    outputted error code on our invocation of curl in with double quotes.
    This allows the command to work on Windows and all the platforms it
    worked on before.
```

Alizter:
I can improve our use of `tar` (via an opam) lib, in a follow up PR.

Leonidas-from-XIV (commented):
I appreciate the research that went into fixing the difference between Windows and Cygwin!

src/dune_pkg/fetch.ml
```diff
@@ -56,11 +56,15 @@ module Curl = struct
...
-      match Int.of_string http_code with
+      (* Since we had to quote the return http_code, we need to strip it *)
+      match
+        Int.of_string (String.sub http_code ~pos:1 ~len:(String.length http_code - 2))
+      with
       | None ->
         Error
           (User_message.make
              [ Pp.textf "curl returned an HTTP code we don't understand: %S" http_code ])
+      | Some 200 -> Ok ()
```
rgrinberg, line 67: This is redundant.

src/dune_pkg/fetch.ml
```diff
@@ -56,11 +56,15 @@ module Curl = struct
            ([ Pp.textf "curl returned an invalid error code %d" exit_code ] @ stderr)))
     else (
       Path.unlink_no_err stderr;
-      match Int.of_string http_code with
+      (* Since we had to quote the return http_code, we need to strip it *)
+      match
+        Int.of_string (String.sub http_code ~pos:1 ~len:(String.length http_code - 2))
```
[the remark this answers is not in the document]
Alizter: When do unexpected inputs occur however? I don't expect curl to not respect the `--write-out` argument.

src/dune_pkg/fetch.ml
Alizter: I meant to also change the line following, let me do that.

rgrinberg: Please don't add pointless cosmetic changes to this PR.

src/dune_pkg/fetch.ml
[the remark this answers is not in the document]
Leonidas-from-XIV: Indeed, that's rare; however the added validation is just an additional point.

My main point is that "remove leading and trailing double quotes" is easier to understand in code than "skip the first character and then read the length of the http_code in bytes minus two", for which I would need to reconstruct what the high-level purpose of doing so is.

src/dune_pkg/fetch.ml
Alizter: OK removing.

rgrinberg:
> @rjbou I assume you are also using cygwin curl for opam?
> 
> @rgrinberg I've updated this fix to only touch our curl call. The explanation in the commit message is:
> 
> ```
>     opam invokes the cygwin version of curl which means passing % in a unix
>     shell is fine. We are however calling the Windows version of curl on
>     Windows which means that % is not a valid token in a command line
>     enviornment.
>     
>     To get around this and make it more cross platform, we quote the
>     outputted error code on our invocation of curl in with double quotes.
>     This allows the command to work on Windows and all the platforms it
>     worked on before.
> ```

@nojb could you confirm this makes sense?

Leonidas-from-XIV (approved):

Leonidas-from-XIV:
@rjbou Seems like the approach in this PR would also be a solution for OPAM on Windows :)

Alizter:
Marking as draft, as I tested it again after some parts were removed and it no longer works. I need to workout why.

Alizter:
Here is some proof of it working on windows:

First we create a lock directory:
```
Solution for dune.lock:
- base-threads.base
- base-unix.base
- csexp.1.5.2
- dune.3.11.1
- ocaml.5.2.0
- ocaml-config.3
- ocaml-variants.5.2.0+trunk
```

Next we run dune build on main:
```
PS C:\Users\ali\test> with-dkml bash .\build
File "dune.lock/ocaml-variants.pkg", line 43, characters 7-58:
43 |   (url https://github.com/ocaml/ocaml/archive/trunk.tar.gz)))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: curl returned an HTTP code we don't understand: "%http_coden"
```

Now we run dune build on this PR:
```
PS C:\Users\ali\test> with-dkml bash .\build
Error: CreateProcess(): No such file or directory
-> required by _build/_private/default/.pkg/ocaml-variants/target/cookie
-> required by - package csexp
```

This is failing as it has successfully downloaded, but the opam call to tar is problematic.

See the final entry in the log:
```
PS C:\Users\ali\test> cat .\_build\log
# C:\Users\ali\dune\_build\install\default\bin\dune.exe build
# OCAMLPARAM: unset
# Shared cache: disabled
# Shared cache location:
# C:\Users\ali\AppData\Local\Microsoft\Windows\INetCache\dune\db
# Workspace root: "C:\\Users\\ali\\test"
# Auto-detected concurrency: 4
$ C:\Users\ali\AppData\Local\Programs\DISKUV~1\tools\MSYS2\usr\bin\curl.exe -L -s --compressed --user-agent dune --write-out "\"%{http_code}\"" -o C:\Users\ali\AppData\Local\Temp\dune_af0b96_trunk.tar.gz\download -- https://github.com/ocaml/ocaml/archive/trunk.tar.gz > C:\Users\ali\AppData\Local\Temp\dune_200746_output 2> C:\Users\ali\AppData\Local\Temp\dune_af0b96_trunk.tar.gz\curl.stderr
$ C:\Users\ali\AppData\Local\Programs\DISKUV~1\tools\MSYS2\usr\bin\tar.exe -xf C:\Users\ali\AppData\Local\Temp\dune_af0b96_trunk.tar.gz\download --force-local -C _build/_private/default/.pkg/ocaml-variants/source > C:\Users\ali\AppData\Local\Temp\dune_9527b0_output 2> C:\Users\ali\AppData\Local\Temp\dune_af0b96_trunk.tar.gz\tar.stderr
PS C:\Users\ali\test>
```

Alizter:
The issue seems to be the `--compressed` argument.

Leonidas-from-XIV:
Looking at the man page of `curl` it seems that it is essentially just adding `deflate` to the accept encoding. Given we're using it to download already compressed tarballs, removing it shouldn't cause any significant differences (I would expect any halfway decent HTTP server to not recompress tarballs anyway).

Alizter:
I've removed `--compressed`. FTR curl.exe has it in the --help page, however when passing, it complains that this version of libcurl doesn't support it. The solution online seems to be to get a more recent version of curl, but it looks like out of the box, --compressed isn't an option for us. I'm also not sure how we could test for it and that's probably too complicated anyway.

Leonidas-from-XIV:
It could be that for `--compressed` curl needs to be linked with `zlib`, but for some reason Microsoft omitted that.

Alizter:
@rgrinberg I don't see anyway to do it other than parsing the output of curl when passing --compressed.

nojb:
> Anyway to detect if this feature is present by running curl? Most tarballs are compressed by it's certainly not a law and it seems a little pessimistic to disable it everywhere.

If memory serves, the usual way to do this is to parse `curl -V` and see if `libz` and/or `brotli` is listed in the `Features` line.

Leonidas-from-XIV:
I took a quick and dirty look at opam-repository (at a random halfway-current revision) and `rg 'https://.*\.tar' -c | wc -l` got me 18839 tarballs, however `rg 'https://.*\.tar"' -c | wc -l` (aka where the URL just ends with a `.tar` without a compression suffix) yields zero results.

I would say that `--compressed` does very little difference and potentially even harm given it adds the accept-encoding to all requests and requires the server to determine whether the file should be compressed.

Alizter:
@nojb Having a look through curl's sources we have:
```
      case 'j': /* --compressed */
        if(toggle && !(feature_libz || feature_brotli || feature_zstd)) {
          err = PARAM_LIBCURL_DOESNT_SUPPORT;
          break;
        }
        config->encoding = toggle;
        break;
```
Which means we only need to test for those features. The `PARAM_LIBCURL_DOESNT_SUPPORT` macro is the error message that we get on Windows.

Alizter:
I've added some `curl -V` detection so that we can be sure the `--compressed` flag is supported.

Alizter:
I've tested it, and it correctly omits --compressed on Windows.

Leonidas-from-XIV:
Did you push your changes? I can't see the detection code in the PR.

Alizter:
@Leonidas-from-XIV I accidentally dropped a commit on that push, should be together now.

src/dune_pkg/fetch.ml
```diff
@@ -2,7 +2,12 @@ open Import
 open Fiber.O
 
 module Curl = struct
-  let bin = lazy (Bin.which ~path:(Env_path.path Env.initial) "curl")
+  let bin =
+    lazy
+      (match Bin.which ~path:(Env_path.path Env.initial) "curl" with
+       | Some p -> p
+       | None -> User_error.raise [ Pp.text "curl not available in PATH" ])
```
Leonidas-from-XIV, line 9: If you want to, you can use the new function to denote `curl` is a command; you can see that in #9079. Also maybe add a hint to install `curl`?

src/dune_pkg/fetch.ml
```diff
@@ -12,30 +17,80 @@ module Curl = struct
...
+          v)
+    ;;
+  end
+
+  let compressed_supported =
+    (* We check if curl supports --compressed by running curl -V and checking if
+       the output contains the features we need. *)
+    Fiber_lazy.create (fun () ->
+      let+ lines, _ =
+        let stderr_to = Process.Io.null Out in
```
Leonidas-from-XIV, line 50: Alternatively `Process.Io.make_stderr`.

src/dune_pkg/fetch.ml
```diff
@@ -12,30 +17,80 @@ module Curl = struct
...
+          Return
+          ~stderr_to
+          ~display:Quiet
+          (Lazy.force bin)
+          [ "-V" ]
+      in
+      (* If these features are present, then --compressed is supported *)
+      let features = [ "libz"; "brotli"; "zstd" ] in
+      match lines with
+      | [ _; _; _; curl_features ] ->
```
Leonidas-from-XIV, line 61: Careful, this will break if curl adds or removes any of these lines (e.g. the user has a git version of curl that might not have the second line which contains the release date), I'd say this is a bit too fragile.

I'd suggest a `find_map` with `seq [bol; str "Features:"; …]` instead.

rgrinberg (changes requested):
Could you add a test to avoid the bug that was reported https://github.com/ocaml/dune/pull/9321

Alizter:
@rgrinberg That was due to touching tar. This PR does not touch tar any longer, only curl. I can add that test in a follow up PR before I submit my tar changes.

src/dune_pkg/fetch.ml
Alizter: AFAIK every platform we support should have curl. Do you have a platform in mind where this is not the case?

Alizter: If a user is using a system that doesn't come preinstalled with curl, then I think there is a good chance they know what they are doing and we can't really give a helpful hint. I'm open to improving this if you have a suggestion for a hint, but I cannot think of one here.

Alizter:
@rgrinberg I've added that test in https://github.com/ocaml/dune/pull/9364.

src/dune_pkg/fetch.ml
```diff
@@ -12,30 +23,84 @@ module Curl = struct
...
+          ~display:Quiet
+          (Lazy.force bin)
+          [ "-V" ]
+      in
+      (* If these features are present, then --compressed is supported *)
+      let features = [ "libz"; "brotli"; "zstd" ] in
+      match
+        List.find_map lines ~f:(fun line ->
+          Re.exec_opt
+            (Re.compile
```
rgrinberg, line 73: The regex should be compiled only once.

Alizter: Done.

PR #9255, opened 2023-11-21
fix(pkg): open rev lock with same settings as dune lock

rgrinberg:
It would be better to share the code, but this is good enough for now.

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -33,7 +33,12 @@ let with_flock lock_path ~f =
   let open Fiber.O in
   let parent = Path.parent_exn lock_path in
   Path.mkdir_p parent;
-  let fd = Unix.openfile (Path.to_string lock_path) [ Unix.O_CREAT; O_RDONLY ] 0o644 in
+  let fd =
+    Unix.openfile
+      (Path.to_string lock_path)
+      [ Unix.O_CREAT; O_WRONLY; O_SHARE_DELETE; Unix.O_CLOEXEC ]
+      0o600
+  in
   let flock = Flock.create fd in
   let max_retries = 49 in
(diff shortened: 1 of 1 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -33,7 +33,13 @@ let with_flock lock_path ~f =
...
   let parent = Path.parent_exn lock_path in
   Path.mkdir_p parent;
-  let fd = Unix.openfile (Path.to_string lock_path) [ Unix.O_CREAT; O_RDONLY ] 0o644 in
+  let fd =
+    Unix.openfile
+      (Path.to_string lock_path)
+      [ Unix.O_CREAT; O_RDONLY; O_SHARE_DELETE ]
+      0o600
+  in
+  Unix.set_close_on_exec fd;
```
Leonidas-from-XIV, line 42: I think `O_CLOEXEC` would be better in that case, since this also avoids potential race conditions.

Leonidas-from-XIV (commented):
Any reason why these flags & modes in particular were chosen?

src/dune_pkg/rev_store.ml
rgrinberg: ah yes indeed.

rgrinberg:
> Any reason why these flags & modes in particular were chosen?

The mysterious flag is `O_SHARE_DELETE` and I think @nojb added it for Windows

nojb:
> > Any reason why these flags & modes in particular were chosen?
> 
> The mysterious flag is `O_SHARE_DELETE` and I think @nojb added it for Windows

Indeed, this flag is needed so that the file can be deleted even if it is open, matching the usual semantics on Unix.

Leonidas-from-XIV (approved):

Issue #8969, opened 2023-10-19
Allow users to set arbitrary solver variables


gridbugs:
The dependency solver uses variables to make decisions about which dependencies to include in the solution. Currently users can set these variables in the `(solver_env ...)` field of a build context configuration in the `dune-workspace` file, but only a small number of prescribed variables may be set this way based on opam's global variables. Since opam packages can refer to arbitrary variables in filtered formulae, and since we don't want package management to be opam-specific anyway, we should allow arbitrary variables to be set here.

Allowing arbitrary variables implies removing the current distinction that dune makes between flags and system variables. This distinction is based on the types of variables expected by opam, but in order to remove opam-specific policy the types of solver variables must be dynamic. The solver (opam or otherwise) will be free to impose type constraints on variables but dune shouldn't prescribe any type policy of its own.

We also need to think about how user-set variables interact with variables which have default values. Two cases where variables are set automatically in the context of the opam solver are:
 - deriving system variables (e.g. `os-family`, `arch`) from the currently running system. Note that this is different from regular default values as the value of system variables will be different on different machines. User-set variables should override the derived values.
 - setting `with-test` to `true` by default so that the decision of whether to build test deps can be deferred to build-time (rather than choosing at solve-time). If a user was to override `with-test` to `false` it would not be possible to dynamically enable tests at build-time as the requisite packages won't be in the lockdir, but a user may want to force it to be false to workaround a situation where no solution can otherwise be found due to a packaging error.

In both cases, variables set in the workspace config override their default values. However it would also be useful to allow users to clear a variable which is set by default (e.g. if they are on linux where `os-family` is set to their distro's name but they want to solve for macos as though `os-family` was unset). For variables derived from the current system, I'd prefer the default value be a special value indicating that the value is derived rather than dynamically choosing the default based on the system.

Here's an example of how setting variables in a build context could look:
```
(solver_env
 (with-doc true)   ; flag
 (os "macos")       ; string variable
 (os-family :unset) ; override a derived system variable, clearing its value
 (arch "%{arch}")  ; string variable with pform
)
```

[...]

PR #9256, opened 2023-11-22
pkg: allow user to set arbitrary solver variables

gridbugs:
First attempt at adding user-controlled solver variables. Fixes https://github.com/ocaml/dune/issues/8969. I intend to add checks that prevent the user from setting user variables whose names are the same as system variables or constants and also prevent the user from setting the `with-test` variable but opening this for review now to get early feedback.

### DIFF
--- a/src/dune_pkg/solver_env.ml
+++ b/src/dune_pkg/solver_env.ml
@@ -83,6 +83,51 @@ module Variable = struct
     end
   end
 
+  module User = struct
+    type t = Package_variable.Name.t
+
+    module Map = Package_variable.Name.Map
+
+    module Bindings = struct
+      type t = string Map.t
+
+      let empty = Map.empty
+      let to_dyn = Map.to_dyn Dyn.string
+      let equal = Map.equal ~equal:String.equal
+      let get = Map.find
+      let set = Map.set
+
+      let decode =
+        let open Decoder in
+        let+ loc, bindings =
+          located (repeat (pair Package_variable.Name.decode string))
+        in
+        match Map.of_list bindings with
+        | Ok t -> t
+        | Error (duplicate_key, a, b) ->
+          User_error.raise
+            ~loc
+            [ Pp.textf
+                "Duplicate entries for user variable %s (%s, %s)"
+                (String.maybe_quoted (Package_variable.Name.to_string duplicate_key))
+                (String.maybe_quoted a)
+                (String.maybe_quoted b)
+            ]
+      ;;
+
+      let pp t =
+        if Map.is_empty t
+        then Pp.text "(none)"
+        else
+          Pp.enumerate (Map.to_list t) ~f:(fun (variable, value) ->
+            Pp.textf
+              "%s = %s"
+              (Package_variable.Name.to_string variable)
+              (String.maybe_quoted value))
+      ;;
+    end
+  end
+
   module Const = struct
     type t = [ `Opam_version ]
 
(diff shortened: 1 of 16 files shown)


src/dune_pkg/solver_env.mli
```diff
@@ -38,6 +38,24 @@ module Variable : sig
     end
   end
 
+  module User : sig
+    (** A custom variable that can be configured by a user *)
+    type t
```
rgrinberg, line 43: Shouldn't this be the same as `Package_variable.Name.t`?

gridbugs: Yes you're right. Updated.

src/dune_pkg/solver_stats.ml
```diff
@@ -91,16 +91,18 @@ module Expanded_variable_bindings = struct
   let to_solver_env { variable_values; unset_variables = _ } =
     (* TODO currently this only supports system variables but this will be
        generalized when the solver gains support for arbitrary variables *)
```
Leonidas-from-XIV, line 93: I guess the comment should be updated now?

Leonidas-from-XIV (approved):

src/dune_pkg/solver_env.mli
```diff
@@ -38,6 +38,24 @@ module Variable : sig
     end
   end
 
+  module User : sig
+    (** A custom variable that can be configured by a user *)
+    type t = Package_variable.Name.t
```
rgrinberg, line 43: I'd remove this type alias and the `User` module.

rgrinberg (approved):
LGTM. Just some minor comments.

On a related note, the whole distinction between "sys" and "user" vars is not something to advertise. It's strictly an implementation detail of ours and it should be hidden from the lock directory and the API outside of `dune_pkg`.

I see it being mentioned everywhere, but I don't really see the point.

src/dune_pkg/solver_env.ml
```diff
@@ -158,35 +209,41 @@ end
...
-let default = create ~sys:Variable.Sys.Bindings.empty
+let create ~sys ~user = { sys; user; const = Variable.Const.bindings }
+let default = create ~sys:Variable.Sys.Bindings.empty ~user:Variable.User.Map.empty
 
 let decode =
   let open Decoder in
   fields
   @@
-  let+ sys = field Fields.sys ~default:default.sys Variable.Sys.Bindings.decode in
+  let+ sys = field Fields.sys ~default:default.sys Variable.Sys.Bindings.decode
```
rgrinberg, line 229: No need to keep this field as separate. Users don't need to care about this distinction. They can set/override all variables.

gridbugs:
The difference between sys and user variables is that sys variables have default values that depend on what system you are on. When setting a sys variable in your config you are _overriding_ some default value for that variable, whereas when setting a user variable you are creating a new variable and assigning it a value. I wanted to make this distinction clear in the ui.

rgrinberg:
We are free to set choose defaults for any other variable and the user would override it in an equivalent way. For example, we may choose to set the default to `with-dev` to `true` one day. Although we're likely to ever to do this, it's within our rights to even choose this default based on the system the user is on.

But disregarding the above, if we are to make a distinction in the UI, we are doing it for some concrete use case. Just as an example, some variables might not be allowed to vary per context, so we'd split them between two groups to make sure the user is aware. Here, we're just drawing the user's attentions to facts that are mostly relevant to us some developers.

gridbugs:
Makes sense. I'll update this change to remove the distinction between different types of solver variable.

gridbugs:
Closing in favour of https://github.com/ocaml/dune/pull/9325

PR #9259, opened 2023-11-22
fix(pkg): improve error checking of lock dir

rgrinberg:
* Use a single stat call for all the checks
* Handle unreadable directories

### DIFF
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -516,19 +516,18 @@ module Write_disk = struct
      values indicate that it's unsafe to remove the existing directory and lock
      directory regeneration should not proceed. *)
   let check_existing_lock_dir path =
-    match Path.exists path with
-    | false -> Ok `Non_existant
-    | true ->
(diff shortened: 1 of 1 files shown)


src/dune_pkg/lock_dir.ml
```diff
@@ -516,19 +516,18 @@ module Write_disk = struct
...
   let check_existing_lock_dir path =
-    match Path.exists path with
-    | false -> Ok `Non_existant
-    | true ->
-      (match Path.is_directory path with
-       | false -> Error `Not_directory
+    match Path.stat path with
+    | Ok { st_kind = S_DIR; _ } ->
+      let metadata_path = Path.relative path metadata_filename in
+      (match Path.exists metadata_path && not (Path.is_directory metadata_path) with
```
Alizter, line 522: This should probably also just be a stat rather than two stats. The negation of Path.is_directory can be any other file type too.

Alizter (approved):

gridbugs (approved):

PR #9266, opened 2023-11-22
fix: use [O_CLOEXEC] when digesting files

rgrinberg:
@snowleopard AFAIK, you don't use this code upstream. But you might still want to make a similar change there.


rgrinberg:
@dkalinichenko-js I've added you as a reviewer on this PR and a couple of others that are like it.

dkalinichenko-js (approved):

PR #9268, opened 2023-11-22
fix(pkg): better lock cleanup

rgrinberg:
* We write the pid to the lock file
* We delete the lock file before release the lock to avoid a race

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -39,20 +39,29 @@ let with_flock lock_path ~f =
       [ Unix.O_CREAT; O_WRONLY; O_SHARE_DELETE; Unix.O_CLOEXEC ]
       0o600
   in
+  let out = Unix.out_channel_of_descr fd in
   let flock = Flock.create fd in
   let max_retries = 49 in
   Fiber.finalize
     ~finally:(fun () ->
-      (* closing the fd releases the flock automatically *)
-      match Unix_error.Detailed.catch Unix.close fd with
-      | Ok () ->
-        (* delete the lock to signal to the user we don't hold a lock *)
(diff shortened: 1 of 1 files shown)


Leonidas-from-XIV (approved):
Looks good, however what is the point of the PID in the lock file? To be able to trace who holds it, if anyone?

Good point on deleting the lock file before releasing the FD, otherwise it would indeed be possible to remove the lockfile of another process (or fail doing exactly that) I would probably solve this by not deleting the lockfile at all, which is avoids the problem just as well.

Alizter:
@Leonidas-from-XIV We add the pid to the dune lock file during a build, which allows us to report which pid has locked the build directory allowing the user to manually kill it or delete the lock as necessary. Writing the pid helps disambiguate between a running process having locked it, and a lock file being left over.

Leonidas-from-XIV:
@Alizter However which PID has the lock can also be retrieved from `lsof`, which is somewhat more authoritative as the process with the original PID can terminate and process ids wrap around etc.

Alizter:
We could do that, but we would also have to make sure the solution is cross-platform. Windows probably has a similar capability since by default you cannot delete "open" files.

Leonidas-from-XIV:
We set `O_SHARE_DELETE` on all locks I think :)

Alizter:
@Leonidas-from-XIV Sorry, I didn't mean that it was a problem on Windows, just that it should be possible to find an `lsof` equivalent.

Leonidas-from-XIV:
Ah yeah, you're right — fair enough.

Leonidas-from-XIV:
I think an error message mentioning the PID would be a good addition, but that can be added later as well.

PR #9269, opened 2023-11-22
test(pkg): invalid package dependency bug

Alizter:
See https://github.com/ocaml/dune/issues/9270

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/invalid-version.t
+++ b/test/blackbox-tests/test-cases/pkg/invalid-version.t
@@ -1,0 +1,14 @@
+Having an invalid package dependency should give a good user message rather than raising
+an uncaught exception. It is very likely that users will type foo.1.2.3 for a package
+version due to the convention in opam. In this case we could also give a hint how to write
+it in a dune-project file.
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.13)
+  > (package
+  >  (name invalid)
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/pkg/invalid-version.t
```diff
@@ -0,0 +1,14 @@
...
+it in a dune-project file.
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.13)
+  > (package
+  >  (name invalid)
+  >  (depends foo.1.2.3))
+  > EOF
+
+  $ dune pkg lock 2>&1 | head -n1
```
rgrinberg, line 13: What's the error message when we try to build the project?

If we have a good error message, it should come regardless if the user is using this feature or not.

Alizter: There is no error message, only a complaint that the package is empty. When passing `(allow_empty)` there is still no error.

rgrinberg: Okay, then please include this in a separate test outside of `pkg/`. I think this is where we should fix the issue first.

Alizter: OK I've done that in https://github.com/ocaml/dune/pull/9271. I'm keeping this test since it's specifically about `dune pkg lock`.

PR #9273, opened 2023-11-23
fix(pkg): remove build warning

rgrinberg:
Removes the following warning:

```
custom_opamStubs.c:92:1: warning: non-void function does not return a value [-Wreturn-type]
}
^
custom_opamStubs.c:97:1: warning: non-void function does not return a value [-Wreturn-type]
}
^
```

@Alizter can you upstream these?


Alizter:
https://github.com/ocaml-dune/opam/pull/6

PR #9282, opened 2023-11-24
fix: copy sandboxes should copy symlinks as well

rgrinberg:
Add support for copying sandboxes with symlinks. We do this just by copying whatever the symlink is pointing to. We also copy broken symlinks.

### DIFF
--- a/src/dune_engine/sandbox.ml
+++ b/src/dune_engine/sandbox.ml
@@ -45,6 +45,7 @@ module Item = struct
   type t =
     | File
     | Directory of { perms : int }
+    | Link
     | Other of Unix.file_kind
 
   let of_path path =
@@ -59,6 +60,7 @@ module Item = struct
     match kind with
     | S_DIR -> Directory { perms = (Path.Untracked.stat_exn path).st_perm }
     | S_REG -> File
+    | S_LNK -> Link
     | _ -> Other kind
   ;;
 end
@@ -71,6 +73,12 @@ let copy_recursively =
   in
   let rec loop item ~src ~dst =
     match (item : Item.t) with
+    | Link ->
+      (match Path.Untracked.stat_exn src with
+       | { Unix.st_kind = S_REG; _ } -> Io.copy_file ~chmod:chmod_file ~src ~dst ()
+       | { Unix.st_kind = S_DIR; st_perm = perms; _ } ->
+         loop (Directory { perms }) ~src ~dst
+       | { Unix.st_kind; _ } -> loop (Other st_kind) ~src ~dst)
     | File -> Io.copy_file ~chmod:chmod_file ~src ~dst ()
     | Directory { perms } ->
       (match Path.Untracked.readdir_unsorted_with_kinds src with
--- a/doc/changes/9282.md
+++ b/doc/changes/9282.md
@@ -1,0 +1,1 @@
+- Copying mode for sandboxes will now follow symbolic links (#9282, @rgrinberg)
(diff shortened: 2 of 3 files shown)


src/dune_engine/sandbox.ml
```diff
@@ -71,6 +73,15 @@ let copy_recursively =
...
   let rec loop item ~src ~dst =
     match (item : Item.t) with
+    | Link ->
+      (match Path.Untracked.stat_exn src with
+       | { Unix.st_kind = S_REG; _ } -> Io.copy_file ~chmod:chmod_file ~src ~dst ()
+       | { Unix.st_kind = S_DIR; st_perm = perms; _ } ->
+         loop (Directory { perms }) ~src ~dst
+       | { Unix.st_kind; _ } -> loop (Other st_kind) ~src ~dst
+       | exception Unix.Unix_error (ENOENT, _, _) ->
+         let contents = Unix.readlink (Path.to_string src) in
```
snowleopard, line 83: What's this case doing? Trying to recreate a broken link in the sandbox? Shouldn't this be an error instead?

src/dune_engine/sandbox.ml
```diff
@@ -59,6 +60,7 @@ module Item = struct
     match kind with
     | S_DIR -> Directory { perms = (Path.Untracked.stat_exn path).st_perm }
     | S_REG -> File
+    | S_LNK -> Link
```
snowleopard, line 63: Where is `S_LNK` coming from? Seems like only `stat_exn`s are around but we need an `lstat_exn`, right?

src/dune_engine/sandbox.ml
Alizter: Packages may create their own broken links for their own tests for example. I think its better to ignore this rather than let dune get stuck on it.

rgrinberg: Yes, that ends up being an issue in practice. Especially for packages that require broken symlinks to run their test suites. We could fork the few packages that do this, but package builds are complete black boxes to us so I think it will be easier to just support it.

Alizter: I should also note that we don't even support broken symlinks in data_only_dirs which is something I looked into before. It's not clear how to handle that case in fs_memo.

src/dune_engine/sandbox.ml
rgrinberg: It comes from listing the directory with `Path.Untracked.readdir_unsorted_with_kinds`.

snowleopard: Ah, got it, thanks.

src/dune_engine/sandbox.ml
snowleopard: Can we force the users to request this behaviour rather than it being the default? Such automagic, I think, always comes back to bite later.

snowleopard: In other words, I think seeing a failure and reacting to it by picking a special sandboxing mode wastes a lot less time compared to the build silently succeeding with unexpected results.

rgrinberg: Umm, how about I'll split this behavior from the current PR if it's controversial.

> Can we force the users to request this behaviour rather than it being the default?

Sure, let's think about how. Would it be a sandboxing option? Seems like it's a bit more general and can apply to non sandboxed rules that produce directory targets. E.g. "allow_broken_symlinks_in_target_directories".

rgrinberg:
Broken symlinks support is gone until we rework it.

snowleopard (approved):
OK, this looks like a clear improvements, thanks! Agreed that we can consider the controversial bit separately.

PR #9292, opened 2023-11-27
fix(tests): adjust to setup-ocaml passing OPAMCONFIRMLEVEL

emillon:
This ensures our test suite is more future proof wrt opam CLI.

As an immediate benefit, this allows the test suite to run with setup-ocaml 2.1.5: it sets `OPAMCONFIRMLEVEL=unsafe-yes`, which emits a warning if the cli version is 2.0.

See ocaml/setup-ocaml#728.


emillon:
This solution is not perfect as there are still many warnings but at least this fixes the CI.

Alizter (approved):
Seems good to me. A bit annoying that this has become an issue now however.

PR #9295, opened 2023-11-27
feature(pkg): Print PID of process that created the lock file

Leonidas-from-XIV:
Follow up to #9268 to print the PID that we write.

This changes the error to an user error since it gives the user some way to deal with this problem, though it is a bit of a grey zone.

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -63,9 +63,17 @@ let with_flock lock_path ~f =
               Unix_error.Detailed.create ue ~syscall:"flock" ~arg:"unlock"
               |> Unix_error.Detailed.raise)
       | Ok `Failure ->
-        Code_error.raise
-          (sprintf "Couldn't acquire lock after %d attempts to lock" max_retries)
-          []
+        let pid = Io.read_file lock_path in
+        User_error.raise
+          ~hints:
+            [ Pp.textf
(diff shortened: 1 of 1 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -63,9 +63,11 @@ let with_flock lock_path ~f =
               Unix_error.Detailed.create ue ~syscall:"flock" ~arg:"unlock"
               |> Unix_error.Detailed.raise)
       | Ok `Failure ->
-        Code_error.raise
-          (sprintf "Couldn't acquire lock after %d attempts to lock" max_retries)
-          []
+        let pid = Io.read_file lock_path |> Int.of_string_exn in
+        User_error.raise
+          ~hints:[ Pp.textf "Check if process %d is running and close it" pid ]
```
Alizter, line 68: This isn't the hint I would give usually. It could be that another process is running and rightfully so, in which case it doesn't make sense to close it. We should already have a message like this for dune.lock:
```
        User_error.raise
          [ Pp.textf
              "A running dune%s instance has locked the build directory. If this is not \
               the case, please delete %s"
              (match Io.read_file (Path.build lock_file) with
               | exception _ -> ""
               | pid -> sprintf " (pid: %s)" pid)
              (Path.Build.to_string_maybe_quoted lock_file)
          ])
;;
```

rgrinberg (approved):

PR #9296, opened 2023-11-27
refactor: introduce [Fpath.traverse_files]

rgrinberg:
To remove duplication between the various traversal functions we have.

### DIFF
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -178,26 +178,24 @@ let if_exists p =
 (* Scan a path recursively down retrieving a list of all files together with their
    relative path. *)
 let scan_files_entries path =
-  (* TODO Add some cycle detection *)
-  let rec read acc dir =
-    let path = Path.append_local path dir in
-    match Path.readdir_unsorted_with_kinds path with
-    | Ok entries ->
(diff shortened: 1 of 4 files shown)


src/dune_pkg/opam_repo.ml
```diff
@@ -178,26 +178,24 @@ let if_exists p =
 (* Scan a path recursively down retrieving a list of all files together with their
    relative path. *)
 let scan_files_entries path =
-  (* TODO Add some cycle detection *)
```
Alizter, line 181: Should we leave a CR for this in fpath?

test/blackbox-tests/test-cases/pkg/opam-package-files-unix-error.t
```diff
@@ -27,7 +27,7 @@ The error message should have a location for the opam repository.
 This does not currently seem to be the case.
 
   $ solve with-patch
-  File "$TESTCASE_ROOT/mock-opam-repository/packages/with-patch/with-patch.0.0.1/files/dir", line 1, characters 0-0:
+  File "$TESTCASE_ROOT/mock-opam-repository/packages/with-patch/with-patch.0.0.1/files", line 1, characters 0-0:
```
Alizter, line 30: I'm a little unsure about this change. Isn't the correct location really `files/dir` like before?

PR #9306, opened 2023-11-28
test(rocq): Demonstrate bug in rocq dep parsing when paths contain spaces

ejgallego:


Alizter:
Yeah that's a bug. I don't suggest merging until we have a fix ready.

ejgallego:
> Yeah that's a bug. I don't suggest merging until we have a fix ready.

It seems we need to split before doing the unescaping, I am not sure what is the best approach to this, would a regex suffice?

ejgallego:
Updated for rocq.

rgrinberg:
Do you plan to merge this @ejgallego? I'd prefer not to leave PR's hanging like this.

ejgallego:
@rgrinberg , indeed, I think I'd prefer to merge, @Alizter would that be OK?

ejgallego:
I guess fixing this bug is a pretty easy papercut (handle quoting properly in rocq dep parsing) , so the idea was to address the bug "on the fly"

Alizter:
> would that be OK?

That's fine by me.

Maybe @rlepigre can take a look at fixing this at some point?

PR #9310, opened 2023-11-29
Add delay to address flaky fsevent test

### LINKED ISSUE #9299: Flaky test in fsevents
I'm happy with this. One issue is that the race condition could lead to flaky CI. I now understand why it happens. In `dune_fsevents_callback` (fsevents_stubs.c) we call the user-provided callback and then if the callback raised an exception we store the exception in `dq->v_exn` and broadcast `dq->dq_finished` which causes `wait_until_stopped` to stop waiting and either return or raise the exception in `dq->v_exn` if there's an exception stored there. The problem happens when `dune_fsevents_stop` is called while a callback is being processed. `dune_fsevents_stop` also broadcasts on `dq->dq_finished` which wakes the thread that called `wait_until_stopped`. If the callback raises but doesn't do so until after `wait_until_stopped` checks for exceptions, the exception never gets re-raised from the c code to the caller of `wait_until_stopped`.

So the question is what should happen to exceptions that are raised by callbacks after the watcher has been stopped (note that the watcher may have been stopped while the callback was running)? I think it's reasonable to continue the current behaviour of dropping such exceptions, in which case a quick fix for the test would be to delay stopping the watcher until the callback is likely to have completed (I don't feel too bad about sleeping in this test since this test already sleeps in another place):
```diff
diff --git a/test/expect-tests/fsevents/fsevents_tests.ml b/test/expect-tests/fsevents/fsevents_tests.ml
index fbc5c6551..ff2fc5ee5 100644
--- a/test/expect-tests/fsevents/fsevents_tests.ml
+++ b/test/expect-tests/fsevents/fsevents_tests.ml
@@ -258,7 +258,8 @@ let%expect_test "raise inside callback" =
       raise Exit)
     (fun () ->
       Io.String_path.write_file "old" "foobar";
-      Io.String_path.write_file "old" "foobar");
+      Io.String_path.write_file "old" "foobar";
+      Unix.sleepf 1.0);
   [%expect {|
     [EXIT]
     exiting. |}]

```

_Originally posted by @gridbugs in https://github.com/ocaml/dune/issues/9009#issuecomment-1829229989_

gridbugs:
The test checks that exceptions raised in in event handlers cause the watcher to terminate and the exception is passed to the thread waiting on the watcher. The test is flaky because the watcher is stopped directly after some period of time and there's a race between this and the exception being raised. This change delays stopping the watcher directly so it's practically guaranteed that there will be time for the exception to be raised.

Fixes https://github.com/ocaml/dune/issues/9299

### DIFF
--- a/test/expect-tests/fsevents/fsevents_tests.ml
+++ b/test/expect-tests/fsevents/fsevents_tests.ml
@@ -258,7 +258,10 @@ let%expect_test "raise inside callback" =
       raise Exit)
     (fun () ->
       Io.String_path.write_file "old" "foobar";
-      Io.String_path.write_file "old" "foobar");
+      Io.String_path.write_file "old" "foobar";
+      (* Delay to allow the event handler callback to catch the exception
+         before stopping the watcher. *)
+      Unix.sleepf 1.0);
   [%expect {|
     [EXIT]
     exiting. |}]


emillon:
Is there a way to trigger the exception at a different time, for example by triggering a GC? (I don't know if that makes sense)

gridbugs:
I don't understand this suggestion. The test is testing that exceptions raised by event handlers are returned by `wait_until_stopped`. How would triggering a GC help here?

emillon:
nevermind, I was confused about the cause of the race condition.

emillon (approved):

PR #9322, opened 2023-11-29
test(pkg): solver ignores conflicts

Alizter:
We demonstrate that the solver ignores conflicts.


Alizter:
Sorry, I've pushed again it should be fixed now.

PR #9323, opened 2023-11-29
test(pkg): duplicate repository definition

rgrinberg:


Leonidas-from-XIV (approved):
Yeah, we should fix this, thanks for the repro!

PR #9325, opened 2023-11-30
pkg: simplify representation of solver variables

gridbugs:
This removes the distinctions between system solver variables and constants and switches to using a single map to store all solver variables. This also allows users to set arbitrary variables in the workspace context.

This was originally meant as fixes to https://github.com/ocaml/dune/pull/9256 but I ended up rewriting most of it so I'm opening a new PR.

### DIFF
--- a/bin/pkg/lock.ml
+++ b/bin/pkg/lock.ml
@@ -38,7 +38,7 @@ let solve
   ~opam_repository_path
   ~opam_repository_url
   ~update_opam_repositories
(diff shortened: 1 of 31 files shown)


gridbugs:
Fixes https://github.com/ocaml/dune/issues/8969

rgrinberg:
Thanks. Rebased and merged this myself since my PR caused the conflicts.

Issue #7383, opened 2023-03-22
Dune gets stuck when promoting a directory that used to be a file


Alizter:
When a previously promoted file changes from a file to a directory, this causes Dune to error and get stuck. The solution is to manually delete the promotion staging area.

Reproduction case:
- #9329

PR #9327, opened 2023-11-30
fix(engine): remove old target even if it is a dir

### LINKED ISSUE #6575: Clean needed when turning a directory into a file
Changing a directory test to a single-file test with the same name requires a `dune clean`:

```shell
Given a working directory cram test:
  $ mkdir sometest.t
  $ cat >sometest.t/run.t <<EOF
  >   $ echo "foobar"
  >   foobar
  > EOF

  $ dune build @sometest

We turn it into a single-file test:
  $ mv sometest.t sometest.t.bak
  $ mv sometest.t.bak/run.t sometest.t
  $ rm -r sometest.t.bak

FIXME: Dune should detect the change:
  $ dune build @sometest
  Error: _build/default/sometest.t: Is a directory
  -> required by _build/default/sometest.t
  -> required by alias sometest
  [1]

After a clean it works as expected:
  $ dune clean
  $ dune build @sometest
```

## Specifications

- Version of `dune`: main (3.6.0-85-gdd6f976)
- Version of `ocaml`: 4.14
- Operating system: Macos (M1)

emillon:
Fixes #6575


Alizter:
Does this address https://github.com/ocaml/dune/issues/7383?

src/dune_engine/build_system.ml
```diff
@@ -539,10 +539,12 @@ end = struct
...
                 try Path.Build.unlink path with
-                | Unix.Unix_error (EISDIR, _, _) ->
-                  (* If target changed from a directory to a file, delete
-                     in anyway. *)
-                  remove_target_dir path
+                (* If target changed from a directory to a file, delete
+                   in anyway. *)
+                | Unix.Unix_error (error, _, _) ->
+                  (match error, Platform.OS.value with
+                   | EISDIR, _ | EPERM, Darwin -> remove_target_dir path
```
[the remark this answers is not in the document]
emillon: This is because `unlink` actually errors with `EPERM` on macos when it's passed a directory, not `EISDIR`. Shall I abstract that to a new function that returns a variant for this case?

emillon:
> what about when a file is changed to a directory

Because we use `rm_rf` for directories, I think that it's going to work fine, but it's worth adding a test case.

src/dune_engine/build_system.ml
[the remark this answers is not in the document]
snowleopard: I think this deserves a comment.

otherlibs/stdune/src/fpath.mli
```diff
@@ -29,6 +29,15 @@ val follow_symlinks : string -> string option
...
 val unlink : string -> unit
 val unlink_no_err : string -> unit
+
+type unlink_status =
+  | Success
+  | Is_a_directory
+  | Other_error
+
+(** Unlink and return error, if any. *)
+val unlink_status : string -> unlink_status
```
snowleopard, line 39: I'm tempted to call this function `unlink` and rename the current `unlink` to `unlink_exn`.

src/dune_engine/build_system.ml
```diff
@@ -550,8 +550,18 @@ end = struct
...
-              Path.Build.Set.iter targets.dirs ~f:(fun dir -> Path.rm_rf (Path.build dir)))
+              let remove_target_dir dir = Path.rm_rf (Path.build dir) in
+              let remove_target_file path =
+                match Path.Build.unlink_status path with
+                | Success -> ()
+                | Is_a_directory ->
+                  (* If target changed from a directory to a file, delete
+                     in anyway. *)
+                  remove_target_dir path
+                | Other_error -> ()
```
snowleopard, line 561: Maybe worth leaving a trace in the log when hitting this case, mentioning the specific exception?

src/dune_engine/build_system.ml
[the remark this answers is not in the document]
snowleopard: Actually, never mind: now that it's a separate function, it seems clear enough.

otherlibs/stdune/src/fpath.ml
```diff
@@ -115,6 +115,21 @@ let win32_unlink fn =
 
 let unlink = if Stdlib.Sys.win32 then win32_unlink else Unix.unlink
 
+type unlink_status =
+  | Success
+  | Is_a_directory
+  | Other_error
```
snowleopard, line 121: Our think the convention we follow elsewhere is use `Error of Unix_error.Detailed.t` to preserve information, e.g. for logging.

src/dune_engine/build_system.ml
emillon: yes, done!

otherlibs/stdune/src/fpath.ml
emillon: Done (but we also ignore other kinds of errors so the argument has type `exn`)

otherlibs/stdune/src/fpath.mli
emillon: I just pushed that too.

emillon:
ping @snowleopard - I renamed the functions as requested, I think this is ready to go but let me know if you need to test this.

snowleopard (approved):
Thanks, looks good!

snowleopard:
cc @rleshchinskiy who recently implemented something similar internally (I believe he simply called `rm_rf` in both cases).

emillon:
thanks. I pushed a change in order not to log if the file does not exist (which is what was breaking the coq test)

PR #9332, opened 2023-11-30
pkg: ignore more ocaml packages when DUNE_PKG_OVERRIDE_OCAML

Alizter:
@rgrinberg These are all dependencies of the `ocaml` package, so we should ignore them as well when passing `DUNE_PKG_OVERRIDE_OCAML`.

Fixes:
- https://github.com/ocaml/dune/issues/9334
- https://github.com/ocaml/dune/issues/9335

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -910,6 +910,18 @@ end
 
 let ocaml_package_name = Package.Name.of_string "ocaml"
 
+let ocaml_package_dependencies =
+  List.map
+    ~f:Package.Name.of_string
+    (* These are all dependencies of the "ocaml" package *)
+    [ "ocaml-config"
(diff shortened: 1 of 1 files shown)


rgrinberg:
What does this fix?

Alizter:
When testing package builds, we don't want these packages to also be built since they are for the `ocaml` package only. Previously, the lock file would contain build rules for these packages too, which meant testing the build of other packages causes these packages to be build in addition. When passing our override, there is no good way to build these packages.

rgrinberg:
I think you should start with a bug ticket explaining the issue you're observing first.

Alizter:
@rgrinberg I've created two issues describing the problems that I encountered that this PR is attempting to fix.

rgrinberg:
Hardcoding a list of packages that provide the compiler is wrong. The opam repository already provides us this information.

PR #9337, opened 2023-11-30
feature(pkg): additional package constraints

rgrinberg:
It's often useful to add additional constraints when solving lock
directories. The additional constraints can be usde to force different
configuration for testing.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -48,14 +48,28 @@ module Context_for_dune = struct
          warning, but only once per package. This field will keep track of the
          packages for which we've printed a warning. *)
       mutable available_cache : bool OpamPackage.Map.t
+    ; constraints : OpamTypes.filtered_formula Package_name.Map.t
     }
 
-  let create ~solver_env ~repos ~local_packages ~version_preference ~stats_updater =
+  let create
+    ~solver_env
+    ~repos
+    ~local_packages
+    ~version_preference
+    ~stats_updater
+    ~constraints
+    =
     let dune_version =
       let major, minor = Dune_lang.Stanza.latest_version in
       OpamPackage.Version.of_string @@ sprintf "%d.%d" major minor
     in
     let candidates_cache = Table.create (module Package_name) 1 in
+    let constraints =
+      List.map constraints ~f:(fun (constraint_ : Package_dependency.t) ->
+        constraint_.name, constraint_)
+      |> Package_name.Map.of_list_multi
+      |> Package_name.Map.map ~f:Package_dependency.list_to_opam_filtered_formula
+    in
     { repos
     ; version_preference
     ; local_packages
@@ -64,6 +78,7 @@ module Context_for_dune = struct
     ; stats_updater
     ; candidates_cache
     ; available_cache = OpamPackage.Map.empty
+    ; constraints
     }
   ;;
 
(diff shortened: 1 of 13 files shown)


rgrinberg:
At the moment, adding constraints cannot introduce new packages to the lock directory. I'm wondering if I should do that.

src/dune_pkg/opam_solver.ml
```diff
@@ -170,11 +185,20 @@ module Context_for_dune = struct
...
-      |> Package_name.Map.mem t.local_packages
+    let name = OpamPackage.name package |> Package_name.of_opam_package_name in
+    let filtered_formula =
+      OpamFormula.map_up_formula
+        (fun formula ->
+          match formula with
+          | Atom (pkg, _) ->
+            let name = Package_name.of_opam_package_name pkg in
+            (match Package_name.Map.find t.constraints name with
+             | None -> formula
```
gridbugs, line 196: Is this how we prevent additional dependencies from being added by the workspace constraints? If so I think this code deserves a comment.

gridbugs:
I don't understand the motivation for this change. Currently it looks like a way of specifying additional version constraints on packages if those packages happen to also be dependencies of local packages. Is that right? When would someone want to add constraints to the workspace rather than adding dependency constraints to local packages?

Also what's the function of adding a package to the `constraints` field without a version number? Is it a noop? If so I think we should make it an error. Would it make sense to name the field `version_constraints` instead?

rgrinberg:
The motivation is as follows: you would like to test your package against multiple versions of OCaml (or any other dependency for that matter). With this field, you can add additional constraints to select particular versions of the dependencies that you would like to test.

> When would someone want to add constraints to the workspace rather than adding dependency constraints to local packages?

It's very likely that the project is compatible with multiple versions of OCaml, but we would like to test and make sure. Therefore we specify the additional constraints in the workspace file to avoid such overly specific constraints to others who might be pinning our package.

> Also what's the function of adding a package to the constraints field without a version number? Is it a noop? If so I think we should make it an error. Would it make sense to name the field version_constraints instead?

For now it's a no-op, but we could make it so that packages listed in the `constraints` field are also installed. Then, it would be a useful feature. For example, one could use this to install some additional dependencies that aren't exactly test dependencies but could be situationally useful. Instrumentation dependencies such as `ppx_bisect` come to mind.

gridbugs:
I think there are a few different objectives we're trying to achieve with this change:
1. testing with different versions of dependencies than what is chosen by default by the solver
2. installing situationally useful development tools

For 1, there's value in allowing the user to specify exact versions of the package to test against, but I don't see a strong reason for allowing arbitrary constraints here. I'd argue this field should be a list of (package, version) tuples rather than re-using the constraint syntax from dune-project to help communicate the intention of this field to users and to clarify that this is not where they should be putting their dependencies.

For 2, why do these need to be specified per-lockdir, and why should the affect the solution? I think these types of tools should just be installed with opam.

Leonidas-from-XIV:
> I'd argue this field should be a list of (package, version) tuples rather than re-using the constraint syntax from dune-project to help communicate the intention of this field to users and to clarify that this is not where they should be putting their dependencies.

If the idea is to make e.g. CI test a specific configuration, I agree that this is better. It would also make sense to have this as a command line option instead of having to write to the workspace file and then run dune, since this is much more direct. One could test whether this conflicts with the constraints or use that as an override over the constraints, not sure what would be more sensible, I could see usecases for both.

> For 2, why do these need to be specified per-lockdir, and why should the affect the solution? I think these types of tools should just be installed with opam.

I think if I have a project that doesn't use an OPAM switch and uses exclusively dune package management, I wouldn't want to go ahead and also install `ocamlfmt`, `merlin`, `lsp-server` etc. etc. separately in a switch, so I think being able to install dev tools is a great feature. However adding dev tools via a lock dir constraint seems extremely obscure to me and no user would consider this discoverable.

gridbugs:
Ok I think that sounds reasonable.

> There are other use cases where specifying the constraint version isn't desirable. For example, I want to test different `depopt` configurations which requires me to pick or disable some package. Specifying a concrete version isn't good here.

In this case I think all the packages listed in the constraints field should be locked (and downloaded), even when the version number is omitted. Generally I prefer that behaviour over making them no-ops anyway.

rgrinberg:
Thinking about this again, being able to introduce constraints without pulling in the packages does sound useful. It can be used to guide the selection of depopts for example.

I'm leaning towards pulling in additional packages a separate field.

gridbugs:
Agreed that that would also be useful and should be part of a separate feature to adding additional dependencies to a workspace.

rgrinberg:
Alright, so I think this is ready for now. I'll add a separate feature to modify the set of "root packages" we use to compute the lock dir.

gridbugs (approved):

PR #9339, opened 2023-12-01
pkg: prevent user from setting with-test variable

gridbugs:
This variable is used internally by the solver to implement dune's policy of including test dependencies of local packages only. We may want to allow users some degree of control over this process eventually but that should come from a higher-level feature than setting the with-test variable directly.


rgrinberg (approved):
It's OK to add stuff like this, but I'd recommend that we focus on expanding the functionality rather than implementing safety belts. We're gonna have plenty of ideas on what to do here once we get some real world experience from users in the upcoming alpha releases.

PR #9346, opened 2023-12-01
pkg: substitute custom solver vars in lockfiles

gridbugs:
Previsouly only "with-test" and "with-doc" were substituted. This generalizes partial evaluation of filters to include all solver vars. To allow users to set "with-doc" but to maintain the current behaviour where it defaults to "false", it's added to the solver env by default with value "false".

Also added a case to the logic for simplifying slang expressions so the generated code is more readable as previously it would generate expressions like `(when true ...)` and `(when false ...)` when a filter is applied to individual terms in a command.


bin/pkg/pkg_common.mli
```diff
@@ -4,8 +4,8 @@ val context_term : doc:string -> Context_name.t option Term.t
 
 (** Create a [Dune_pkg.Solver_env.t] by combining variables taken from the
     current system and variables taken from the current context, with priority
-    being given to the latter. Also adds a binding from the variable
-    "opam-version" to the version of opam vendored by dune. *)
+    being given to the latter. Also adds some sensible default values for some
+    variables (e.g. "opam-version") which can be overridden. *)
```
Leonidas-from-XIV, line 8: I liked the mention that "opam-version" would match our vendored opam.

src/dune_lang/slang.ml
```diff
@@ -217,6 +222,11 @@ let rec simplify = function
 
 and simplify_blang = function
   | Blang.Const b -> Blang.Const b
+  | Expr (Literal s) ->
+    (match String_with_vars.text_only s with
+     | Some "true" -> Const true
+     | Some "false" -> Const false
+     | _ -> Expr (Literal s))
```
Leonidas-from-XIV, line 229: Maybe bind the `Expr (Literal s)` in the match clause and reuse it here to show that it is the same. Potentially saves an allocation, but mainly a bit simpler to grasp that it doesn't simplify in that case.

src/dune_pkg/opam_solver.ml
```diff
@@ -351,15 +351,20 @@ let filter_to_blang ~package ~loc filter =
...
   OpamFilter.partial_eval (fun var ->
     match OpamVariable.Full.scope var with
     | Global ->
-      let name = OpamVariable.Full.variable var in
-      if OpamVariable.equal name (OpamVariable.of_string "with-doc")
-         || OpamVariable.equal name (OpamVariable.of_string "with-test")
-      then Some (B false)
-      else None
+      let name = OpamVariable.Full.variable var |> Variable_name.of_opam in
+      if Variable_name.(equal name with_test)
```
Leonidas-from-XIV, line 359: A bit nitpicky but `name` could be shadowed if `Variable_name` introduces `name` (which doesn't sound all that unlikely). There would probably be a warning, but I feel writing out `Variable_name.equal` and `Variable_name.with_test` makes it easier to understand.

src/dune_pkg/solver_env.mli
```diff
@@ -14,8 +14,7 @@ val get : t -> Variable_name.t -> Variable_value.t option
     existing values of those variables in [a]. *)
 val extend : t -> t -> t
 
-(** A [t] with a single variable "opam-version" set to the version of the opam
-    library dependen on by dune *)
-val with_opam_version_set_to_current : t
+(** A [t] with default values for some variables *)
+val with_defaults : t
```
Leonidas-from-XIV, line 18: I actually like this name a lot better! Maybe mention the defaults in the documentation string?

src/dune_pkg/variable_value.ml
```diff
@@ -13,6 +13,9 @@ open! Import
 *)
 type t = String of string
```
Leonidas-from-XIV, line 14: Not really related to this PR, but given there's just one constructor, maybe it can be removed?

Leonidas-from-XIV (approved):

src/dune_lang/slang.ml
```diff
@@ -195,7 +195,12 @@ let rec simplify = function
          let combined_sw = String_with_vars.make ~quoted loc parts in
          Literal combined_sw)
        else Form (loc, Concat (List.map xs ~f:simplify))
-     | When (condition, t) -> Form (loc, When (simplify_blang condition, simplify t))
+     | When (condition, t) ->
+       let simplified_condition = simplify_blang condition in
+       (match simplified_condition with
+        | Blang.Const true -> t
+        | Const false -> Nil
```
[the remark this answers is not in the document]
Julow: Unfortunately OCamlformat cannot change what it formats. Maybe what you need is a linter based on [semgrep](https://github.com/semgrep/semgrep) ?
There's no hope of performing such transformation automatically, for example, if `A.t` is a GADT.

src/dune_pkg/solver_env.mli
gridbugs: Added

rgrinberg (approved):

PR #9348, opened 2023-12-01
[ci] Test Coq 8.18

ejgallego:
This is an experiment to see if we should test in CI a lower and an upper bound for Coq.

Note that starting with Coq 8.17, Coq has been split into `coq-core` and `coq-stdlib`, with `coq-stdlib` being the heavy package.

Unfortunately we still need to install `coq-stdlib` for our tests, as we need to check that stdlib / installed theories are working properly, but we could at some point just have a very light `coq-prelude` package that would take 2 seconds to build and would work for us too.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -212,7 +212,7 @@ jobs:
       - uses: cachix/install-nix-action@v18
       - run: nix develop .#doc -c make doc
 
-  coq:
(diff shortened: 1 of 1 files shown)


Alizter:
Good opportunity to make the output of the tests a bit more version independent.

ejgallego:
Superseeded by #12035

PR #9352, opened 2023-12-01
fix(pkg): Only parse `blob` objects into files

### LINKED ISSUE #9328: pkg: rev_store can't handle symlinks
Some opam repositories such as [dune-universe/opam-overlays](https://github.com/dune-universe/opam-overlays/tree/master/test/rwo.t) contain symlinks. This causes the rev_store to fail since the parser cannot understand the output of `git-ls`. Dune raises 
```
Error: exception Not_found
Raised at Dune_re__Core.exec in file "vendor/re/src/core.ml", line 937,
  characters 20-35
Called from Dune_pkg__Rev_store.File.parse.(fun) in file
  "src/dune_pkg/rev_store.ml", line 227, characters 14-29
```
in this case which under inspection gives the following line from `git-ls` as problematic:
```
160000 commit 0f8ea69ad66f745bd52ecae7189b72f8b6d50da8       -	test/rwo.t/book
```

Presumably this is due to the fact that `book` is a symlink in that repo and therefore changes the output of `git-ls`.

In this particular case the symlink occurs outside of `packages/` so the case could be made to ignore it, however it seems sensible after discussing with @Leonidas-from-XIV that symlinks could be supported in the `files/` directory of packages. Therefore the rev_store should be modified to handle them correctly.

To reproduce my error above, use the following `dune-workspace` file:
```
(lang dune 3.11)

(repository
 (name opam-overlays)
 (source "git+https://github.com/dune-universe/opam-overlays.git"))

(context
 (default
  (name default)
  (repositories opam-overlays)))
```

Leonidas-from-XIV:
While investigating #9328 it turns out that the tree has an object that's of type `commit`. I originally thought that this was a symlink to nowhere (since the revision it points to does not exist) but it turns out it is something different as symlinks are represented as blobs after all.

The code should discard these non-blob objects. However I don't yet know how to create a `commit` object for the repro case so for now it only tests broken symlinks. Any git-fu appreciated!

Fixes #9328

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -211,11 +211,11 @@ module File = struct
       let perm = Re.(rep1 digit) in
       let hash = Re.(rep1 alnum) in
       let type_ = Re.(rep1 alpha) in
-      let size = Re.(rep1 digit) in
+      let size = Re.(alt [ rep1 digit; str "-" ]) in
       let path = Re.(rep1 any) in
       [ perm
       ; space
-      ; type_
+      ; Re.group type_
       ; space
       ; Re.group hash
       ; space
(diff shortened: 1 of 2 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -224,11 +224,17 @@ module File = struct
...
     fun line ->
-      let m = Re.exec re line in
-      { hash = Re.Group.get m 1
-      ; size = Int.of_string_exn @@ Re.Group.get m 2
-      ; path = Path.Local.of_string @@ Re.Group.get m 3
-      }
+      Re.exec_opt re line
+      |> Option.bind ~f:(fun m ->
+        match Re.Group.get m 1 with
+        | "blob" ->
```
rgrinberg, line 230: If we only want blobs, how about we modify the regex to accept only blobs?

rgrinberg:
Could you add a test case in a separate PR so that we can record the issue existing and then being fixed?

Leonidas-from-XIV:
@rgrinberg Sure, that actually prompted me into building a repo that has this exact issue that Ali described: #9372. I will rebase this PR on top of that when it is merged, since that repro test is much better than the one here.

Leonidas-from-XIV:
Rebased on top of the repro case and it shows that locking works now even in the presence of a `commit` object.

rgrinberg (approved):

PR #9357, opened 2023-12-03
refactor: remove physical equality from dialects

rgrinberg:

### DIFF
--- a/src/dune_rules/dune_project.ml
+++ b/src/dune_rules/dune_project.ml
@@ -146,7 +146,8 @@ type t =
   ; warnings : Warning.Settings.t
   ; use_standard_c_and_cxx_flags : bool option
   ; file_key : File_key.t
-  ; dialects : Dialect.DB.t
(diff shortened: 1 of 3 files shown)


src/dune_rules/dune_project.ml
```diff
@@ -692,9 +692,10 @@ let encode : t -> Dune_lang.t list =
       [ string "lang"; string "dune"; Dune_lang.Syntax.Version.encode dune_version ]
   in
   let dialects =
-    if Dialect.DB.is_default dialects
-    then []
-    else Dialect.DB.fold ~f:(fun d ls -> Dialect.encode d :: ls) ~init:[] dialects
```
nojb, line 697: Instead of the `dialects` value into an option, why not simply remove `is_default` and the `then` case here? As far as I can see the `builtin` values appear in the result of this function as soon as there is at least one extra dialect.

nojb (commented):
I think the PR is correct, but I think something simpler can be done.

PR #9363, opened 2023-12-03
test: dropping entire prefix for Path.drop_prefix

Alizter:
We add a test demonstrating that dropping the entire path as the refix in both Path.drop_prefix and Path.External.drop_prefix has a set behaviour. This is in preperation of #8965.

### DIFF
--- a/otherlibs/stdune/test/path_tests.ml
+++ b/otherlibs/stdune/test/path_tests.ml
@@ -494,3 +494,15 @@ let%expect_test "drop external prefix as substring" =
   |> print_dyn;
   [%expect {| Some "baz" |}]
 ;;
+
+let%expect_test "drop entire path" =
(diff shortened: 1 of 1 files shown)


otherlibs/stdune/test/path_tests.ml
```diff
@@ -494,3 +494,19 @@ let%expect_test "drop external prefix as substring" =
...
+
+let%expect_test "drop entire path" =
+  Path.drop_prefix ~prefix:(r "foo/bar") (r "foo/bar")
+  |> Dyn.option Path.Local.to_dyn
+  |> print_dyn;
+  [%expect {| Some "." |}]
+;;
+
+let%expect_test "drop entire external path" =
+  Path.External.drop_prefix
```
rgrinberg, line 506: Same here

Issue #9365, opened 2023-12-03
select doesn't work in test stanza


Alizter:
Select doesn't appear to work in the `test` stanza, compared to `library` and `executable`.

Alizter:
I have a fix ready.

PR #9366, opened 2023-12-03
test: reproduction case for #9365

Alizter:
We also test the other stanzas while reproducing the bug with the test stanza.

### DIFF
--- a/test/blackbox-tests/test-cases/select-in-test.t
+++ b/test/blackbox-tests/test-cases/select-in-test.t
@@ -1,0 +1,32 @@
+We test the (select) field of the (libraries) field of the test stanza.
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.11)
+  > EOF
+
+  $ cat > dune <<EOF
+  > (test
+  >  (name test)
+  >  (modules test)
+  >  (libraries
+  >   (select test.ml from
+  >    (unix -> test.unix.ml)
+  >    (!unix -> test.nounix.ml))))
+  > EOF
+
+  $ cat > test.unix.ml <<EOF
+  > let () = print_endline "Test: Unix was found!"
+  > EOF
+  $ cat > test.nounix.ml <<EOF
+  > let () = print_endline "Test: Unix was not found!"
+  > EOF
+
+The select field does not pick up the module sources for the test stanza
+correctly. This is a bug.
+
+  $ dune runtest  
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/select.t/run.t
```diff
@@ -37,3 +49,56 @@
   $ dune runtest
   bar has unix
   foo has no fake
+
+Then we test that the rule stanza works as expected in the library stanza:
```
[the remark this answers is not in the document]
Alizter: I wrote a new test and removed the changes to the existing one.

rgrinberg:
If the issue is about the test stanza, why does the test include library and executable stanzas? Shouldn't it be possible to reproduce the issue only with the test stanza?

Alizter:
@rgrinberg It's just about the test stanza, but this ensures we are also testing the library and executable stanzas.

rgrinberg:
Don't we have other select tests that make this redundant?

Alizter:
I could only find a test for executable, this additionally has a test for library.

rgrinberg:
It would be better not to mix adding a regression test and re-organizing the existing tests. So I suggest that you a regression test and then create a subsequent PR to re-organize things.

Alizter:
I've reduced the test down to testing only the bugged behaviour of `select` in `test`.

test/blackbox-tests/test-cases/select-in-test.t
```diff
@@ -0,0 +1,32 @@
+We test the (select) field of the (libraries) field of the test stanza.
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.11)
+  > EOF
+
+  $ cat > dune <<EOF
+  > (test
+  >  (name test)
+  >  (modules test)
```
rgrinberg, line 10: Do you need the `modules` line to reproduce the bug?

rgrinberg:
By the way, there might a similar bug with `melange.emit`

test/blackbox-tests/test-cases/select-in-test.t
Alizter: No, that was from when the other stanzas were there. I'll remove it.

rgrinberg:
Thanks. Next time, try to keep the test demonstrating the bug as simple as possible. Improvements can always be made later.

PR #9368, opened 2023-12-03
fix(coq): delay scope loading in install rules

rgrinberg:
No idea if this helps, but I did say I would at least give it a try.

@Alizter @ejgallego can you see if it helps?

### DIFF
--- a/src/dune_rules/coq/coq_rules.ml
+++ b/src/dune_rules/coq/coq_rules.ml
@@ -952,11 +952,14 @@ let coqtop_args_theory ~sctx ~dir ~dir_contents (s : Coq_stanza.Theory.t) coq_mo
 (* This is here for compatibility with Coq < 8.11, which expects plugin files to
    be in the folder containing the `.vo` files *)
 let coq_plugins_install_rules ~scope ~package ~dst_dir (s : Coq_stanza.Theory.t) =
-  let lib_db = Scope.libs scope in
   let+ ml_libs =
     (* get_libraries from Coq's ML dependencies *)
-    Resolve.Memo.read_memo
-      (Resolve.Memo.List.map ~f:(Lib.DB.resolve lib_db) s.buildable.plugins)
+    Resolve.Memo.List.map s.buildable.plugins ~f:(fun lib ->
+      let open Resolve.Memo.O in
+      let* scope = Resolve.Memo.lift_memo scope in
+      let lib_db = Scope.libs scope in
+      Lib.DB.resolve lib_db lib)
+    |> Resolve.Memo.read_memo
   in
   let rules_for_lib lib =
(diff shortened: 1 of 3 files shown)


src/dune_rules/coq/coq_rules.ml
```diff
@@ -988,12 +991,6 @@ let install_rules ~sctx ~dir s =
   | { Coq_stanza.Theory.package = Some package; buildable; _ } ->
     let loc = s.buildable.loc in
     let* mode = select_native_mode ~sctx ~dir buildable in
-    let* scope = Scope.DB.find_by_dir dir in
-    (* We force the coq scope for this DB as to fail early in case of
-       some library conflicts that would also generate conflicting
-       install rules. This is needed as now install rules less lazy
-       than the theory rules. *)
-    let _ = Scope.coq_libs scope in
```
rgrinberg, line 996: This forcing is an issue. It requires computing the coq scope in the install rules, which itself may force the install rules by depending on public executables.

src/dune_rules/coq/coq_rules.ml
```diff
@@ -952,11 +952,14 @@ let coqtop_args_theory ~sctx ~dir ~dir_contents (s : Coq_stanza.Theory.t) coq_mo
 (* This is here for compatibility with Coq < 8.11, which expects plugin files to
    be in the folder containing the `.vo` files *)
 let coq_plugins_install_rules ~scope ~package ~dst_dir (s : Coq_stanza.Theory.t) =
-  let lib_db = Scope.libs scope in
   let+ ml_libs =
     (* get_libraries from Coq's ML dependencies *)
-    Resolve.Memo.read_memo
-      (Resolve.Memo.List.map ~f:(Lib.DB.resolve lib_db) s.buildable.plugins)
+    Resolve.Memo.List.map s.buildable.plugins ~f:(fun lib ->
```
rgrinberg, line 957: The point of this shuffling is that you will now no longer load the OCaml library database unless you use plugins. It's possibly unnecessary, but it seems like a good thing to have anyway.

Alizter:
This didn't change anything. Btw you can run 
```
Testing memo cycle in dune coq rules

  $ cat > dune-project << EOF
  > (lang dune 3.9)
  > (using coq 0.8)
  > (package (name foo))
  > EOF

  $ cat > dune << EOF
  > (executable
  >  (public_name coqc)
  >  (package foo))
  > 
  > (coq.theory
  >  (name bar)
  >  (package foo))
  > EOF

  $ cat > coqc.ml << EOF
  > asdfasdf
  > EOF
  $ cat > bar.v

  $ timeout s 2 --preserve-status 1 dune build @install

```
without having Coq installed.

PR #9372, opened 2023-12-04
test: reproduction case for #9328

Leonidas-from-XIV:
This reproduces the case in #9328 where the repo contains a `commit` object (not that this serves any purpose, but such cases exist in the wild) and our code breaks.

It comments the steps on how to build such a structure in git as this is using some of the git "plumbing" tools that is not particularly well-known, however the [Git Book: Git Objects](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) has some information on how to build a commit by hand (they use files in the book but fortunately it also works with commits).

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/git-repo.t
+++ b/test/blackbox-tests/test-cases/pkg/git-repo.t
@@ -98,3 +98,40 @@ that by hashing with the checksum that we expected in the OPAM file
 
   $ cmp -s dune.lock/foo.files/$FILES_NAME "$FILES_FOLDER/$FILES_NAME" && echo "The contents match"
   The contents match
+
+The git repos support repo should also be able to handle unusual objects in our
+tree, which can e.g. happen with submodules.
+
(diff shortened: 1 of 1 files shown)


rgrinberg (approved):

Alizter (approved):

test/blackbox-tests/test-cases/pkg/git-repo.t
```diff
@@ -98,3 +98,38 @@ that by hashing with the checksum that we expected in the OPAM file
 
   $ cmp -s dune.lock/foo.files/$FILES_NAME "$FILES_FOLDER/$FILES_NAME" && echo "The contents match"
   The contents match
+
+The git repos support repo should also be able to handle unusual objects in our tree, which can e.g. happen with submodules.
```
rgrinberg, line 102: This paragraph seems a bit long. Wrapping it to 80 chars would make it a little nicer.

PR #9373, opened 2023-12-04
feat(pkg): Remove `opam-repository-url` option

Leonidas-from-XIV:
Since the repositories are specified in the workspace file where more information like URL, branch or tag can be stored this makes for a better set-up.

### DIFF
--- a/bin/pkg/lock.ml
+++ b/bin/pkg/lock.ml
@@ -36,7 +36,6 @@ let check_for_dup_lock_dir_paths ts =
 let solve
   per_context
   ~opam_repository_path
-  ~opam_repository_url
   ~update_opam_repositories
   ~solver_env_from_current_system
   ~experimental_translate_opam_filters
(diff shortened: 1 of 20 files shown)


rgrinberg (approved):
LGTM. I think you can remove the opam repository path option as well.

gridbugs (approved):

Leonidas-from-XIV:
>  I think you can remove the opam repository path option as well.

I tried but that will currently leave us without a way to specify directories as opam-repos, as the `dune-workspace` currently only supports git repos. So maybe the way forward would be to allow that first and then remove that option?

Leonidas-from-XIV:
Ok, also removed the `opam-repository-path`, but that required a lot of changes and modifying `dune-workspace` with Unix tools is quite painful.

Alizter:
In #9343 I am removing all the --context arguments and replacing them with positional lock directories. Maybe we should hold off on the opam-repository-path changes until thats done, since it will simplify the current workspace wrangling.

rgrinberg:
I'll leave it to Marek to decide, but I would suggest not to delay merging this PR for other work. There's some work I'd like to do that is waiting for this PR to be merged first.

Leonidas-from-XIV:
I've merged this because at the end of the day, someone will have to rebase (and I know it sucks, I just did and more tests broke). Like Rudi I also have some work that is blocked on this, so if merging this unblocks both of us that's some compromise.

Changing the CLI is always painful because it touches a lot of tests. Though I assume this will get better once we have a post-1.0 CLI interface.

Issue #3961, opened 2020-11-19
Use `dune exec` for testing failing command


bobot:
When an action fails we currently print a little sh script which should run the command in the same way:

```
(cd _build/default/foo/bar && ../my.exe arg1 arg2 > file1 )
```

However we have to go to great length to translate some action features such as redirection, pipe, that we are not always sure to translate correctly. Moreover dune adds also a lot of environment variables such as `OCAML_LOADPATH`, and also dune specific one for the sites and plugin mechanism. So running executable directly from `_build` is a bad habit and will fail in not simple cases.

`dune exec` execute correctly executable with the needed environment. We can perhaps make it able to execute actions so that the output of failed command is simpler to write and more truthful but still easy to modify from the command line:

```
dune exec --in-build-dir --chdir foo/bar --context default --with-stdout-to file1 -- ../my.exe arg1 arg2
```

avsm:
This makes sense. Perhaps best to drop default values (e.g. `--context default) to try to minimise it a little, but the approach  sounds good to me.

rgrinberg:
Sounds good to me as well. Though I'd prefer an sexp based syntax that somewhat reflects how we write the actions in the dune files.

> Moreover dune adds also a lot of environment variables such as OCAML_LOADPATH, and also dune specific one for the sites and plugin mechanism. So running executable directly from _build is a bad habit and will fail in not simple cases.

One idea that was on my mind before was a `--debug` mode for builds. It would allow you to debug failure by launching a shell in the build directory where the command failed with all the environment variables set. One could then replay the failed command with some shell function. This is possibly overkill, but this could have been handy for me on a number of occasions.

nojb:
> One idea that was on my mind before was a `--debug` mode for builds. It would allow you to debug failure by launching a shell in the build directory where the command failed with all the environment variables set. One could then replay the failed command with some shell function. This is possibly overkill, but this could have been handy for me on a number of occasions.

This sounds interesting.

Personally am not sure the approach proposed in this PR is the best solution. When debugging build failures oftentimes it is useful to remove `dune` from the equation, as the problem may be in `dune` itself, or you need to debug the issue at a lower level than what `dune` exposes, eg individual compilation commands. In these cases case, it is useful to have available and be able to execute the command independently of `dune`, for which you need the "shell" command line.

bobot:
>  In these cases case, it is useful to have available and be able to execute the command independently of dune, for which you need the "shell" command line.

[...]

PR #9375, opened 2023-12-04
add option to keep sandboxes after use

Alizter:
### Problem

When running into errors or unexpected outputs in our rules it can be helpful to inspect the _build directory to see what has happened. Unfortunately if these actions are sandboxed, we do not retain the sandbox after execution which makes debugging them difficult.

### Solution

We add a hidden developer option `DUNE_CONFIG__CLEAR_SANDBOXES` that defaults to `enabled` which when `disabled` will skip the cleanup step of the sandboxing mechanism. This allows for manual inspection of a _build directory and allows us to better rerun command line output for sandboxed actions.

This was useful in debugging #9374 for example.

### DIFF
--- a/src/dune_config/config.ml
+++ b/src/dune_config/config.ml
@@ -192,3 +192,9 @@ let threaded_console_frames_per_second =
   register t;
   t
 ;;
+
+let clear_sandboxes =
+  let t = { name = "clear_sandboxes"; of_string = Toggle.of_string; value = `Enabled } in
+  register t;
+  t
+;;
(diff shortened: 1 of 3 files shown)


snowleopard:
I'm not convinced, to be honest. I don't mind merging this since it's not a lot of code but what's the intended workflow? Run a build, then search through a ton of sandboxes, and then manually clean up? Doesn't seem very user-friendly.

What happened to the `dune debug`/`dune shell` ideas? Not sure it's the best reference but here is a ticket I found after a quick search: https://github.com/ocaml/dune/issues/3961.

I'd prefer we invested time into developing a proper solution.

Alizter:
There is no need to search for sandboxes since they are recorded in the build log. If you don't think this is worth it then thats enough reason for me to close. I found it useful when working on other things, so I will keep it as a local hack.

PR #9376, opened 2023-12-04
refactor: make all stanzas define comparisons/hashing

rgrinberg:
We require stanzas to provide comparisons and hashing. This gets rid of physical equality. Eventually, we'll also get rid of polymorphic comparisons once we can use ppx.

@nojb could you take a brief look at stanza.ml and let me know if I'm over complicating things.


rgrinberg:
gentle ping on this one @nojb

nojb:
> gentle ping on this one @nojb

Thanks for the ping. I'll take a look later today.

nojb (commented):
I am not en expert on the techniques used in `stanza.mli`, but looks reasonable to me.

I am not formally approving the PR because I didn't look at all at the rest of it :)

PR #9378, opened 2023-12-04
feature(pkg): add our package overlay to default repos

rgrinberg:
Our newly created overlay repository (https://github.com/ocaml-dune/opam-overlays) is added to the list of repos. I've also renamed the repos to clarify things:

* `upstream` is ocaml/opam-repository
* `overlay` is our overlay

### DIFF
--- a/bin/pkg/pkg_common.ml
+++ b/bin/pkg/pkg_common.ml
@@ -68,7 +68,10 @@ module Per_context = struct
     let repositories =
       Option.map lock_dir ~f:(fun lock_dir -> lock_dir.repositories)
       |> Option.value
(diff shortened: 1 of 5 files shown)


Alizter:
Does this mean that users cannot name their own repos overlay? We could try `dune-overlay` to avoid this.

gridbugs (approved):

Leonidas-from-XIV (approved):

rgrinberg:
> (e.g. you couldn't do :standard \ upstream) is a different discussion for which we should probably have a ticket to discuss.

Why shouldn't one do `:standard \ upstream`? It seems like a reasonable thing to want and the point of using the OSL is to allow such stuff in the first place.

PR #9381, opened 2023-12-04
refactor: unify relational operators

rgrinberg:
Relational operators for blang and package constraints are equivalent.
We don't need two types.

### DIFF
--- a/src/dune_lang/relop.mli
+++ b/src/dune_lang/relop.mli
@@ -1,0 +1,15 @@
+type t =
+  | Eq
+  | Gte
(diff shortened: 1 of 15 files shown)


src/dune_lang/relop.mli
```diff
@@ -0,0 +1,15 @@
+type t =
+  | Eq
```
rgrinberg, line 2: We could even make this a polymorphic variant and avoid yet another set of conversions.

gridbugs (approved):

PR #9384, opened 2023-12-05
pkg: compute checksums for packages without checksums

### LINKED ISSUE #9336: Support packages without checksums
Some packages in the opam repository do not include checksums. One such example are the `+trunk` OCaml compiler packages.

We should download such packages as part of the solving stage and include the newly computed checksums.

gridbugs:
fixes https://github.com/ocaml/dune/issues/9336

### DIFF
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -46,6 +46,61 @@ module Source = struct
         ]
   ;;
 
+  let fetch_and_hash_archive url =
+    let open Fiber.O in
+    let temp_dir = Temp.create Dir ~prefix:"dune" ~suffix:"archive" in
+    Fiber.finalize ~finally:(fun () ->
+      Temp.destroy Dir temp_dir;
+      Fiber.return ())
+    @@ fun () ->
+    let target = Path.relative temp_dir "archive" in
+    Fetch.fetch ~unpack:false ~checksum:None ~target (OpamUrl.of_string url)
+    >>| function
+    | Ok () -> Some (Dune_digest.file target |> Checksum.of_dune_digest)
+    | Error (Checksum_mismatch _) ->
+      Code_error.raise "Checksum mismatch when no checksum was provided" []
+    | Error (Unavailable message_opt) ->
+      let message =
+        match message_opt with
+        | Some message -> message
+        | None ->
+          User_message.make [ Pp.textf "Failed to retrieve source archive from: %s" url ]
+      in
+      User_warning.emit_message message;
+      None
+  ;;
+
+  let compute_missing_checksum_of_fetch
+    ({ url = url_loc, url; checksum } as fetch)
+    package_name
+    =
+    let open Fiber.O in
+    match checksum with
+    | Some _ -> Fiber.return fetch
+    | None ->
+      User_message.print
+        (User_message.make
+           [ Pp.textf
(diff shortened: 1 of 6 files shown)


gridbugs:
It occurs to me that maybe we should add some indication that packages have had their hashes added by dune in this way. Hashes in opam have some air of legitimacy since they've at least gone through a manual code review process when the package was added to the repo, but if dune is just going to add its own hashes to packages that otherwise lack them people might start to trust our hashes to the same degree that they trust opam's.

rgrinberg:
I suppose we could print the hashes that we've had to compute to let the user know we're doing this. FYI, opam installs such packages without checksums without any ceremony either. I'm not even sure if it saves the hash it ends up downloading.

gridbugs:
I added a message about how dune is generating its own checksum. It also prints the source archive url so users can vet suspicious-looking urls. Also I changed the test to use a text file rather than an archive since creating the archive (tar) was producing different results on different platforms.

src/dune_pkg/lock_dir.ml
```diff
@@ -46,6 +46,61 @@ module Source = struct
...
+    let open Fiber.O in
+    let temp_dir = Temp.create Dir ~prefix:"dune" ~suffix:"archive" in
+    Fiber.finalize ~finally:(fun () ->
+      Temp.destroy Dir temp_dir;
+      Fiber.return ())
+    @@ fun () ->
+    let target = Path.relative temp_dir "archive" in
+    Fetch.fetch ~unpack:false ~checksum:None ~target (OpamUrl.of_string url)
+    >>| function
+    | Ok () -> Some (Dune_digest.file target |> Checksum.of_dune_digest)
```
rgrinberg, line 59: FYI, the internals of `Fetch.fetch` can already give you the newly calculated checksum. It should be enough to just tweak the API to get it somehow.

rgrinberg (approved):
LGTM for a start, but I see a couple things that are fishy:

1. Why do we include checksums for sources computed from the local file system? Seems a bit pointless since they can't be shared anyway.

2. The build step is going to refetch the sources again. That seems wasteful. We need to reuse the build rules somehow when fetching here to avoid this.

Leonidas-from-XIV:
> We could introduce this cache some other directory, but it seems a bit premature to do this.

Wouldn't `$XDG_CACHE_LOCATION/dune` be the right place for it, potentially even reusing the same convention as opam? Another option would be to share OPAMs cache, but I am not sure whether that's a good move.

rgrinberg:
Btw, you can set `dev = true` for packages that are fetched this way

rgrinberg:
> Deduplication would happen automatically if you use <hash-algo>/<first two chars of hash>/<rest-of-hash> for the filename as the OPAM cache does. Which works rather well with packages sharing a single-tarball (e.g. all multi-package projects using dune-release).

The dune cache is much better than that. It deduplicates on a per file basis, so tarballs of different versions of the same package will end up sharing the majority of the contents.

But even without that, creating a secondary cache system parallel to the one we already have is somewhat of a last resort.

Leonidas-from-XIV:
That sounds reasonable to me.

Out of curiosity, how many packages do not have checksums in opam-repository?

rgrinberg:
> Out of curiosity, how many packages do not have checksums in opam-repository?

Not sure how many, but the `+trunk` OCaml packages use it so it's important to support.

Issue #9371, opened 2023-12-04
Use `Dune_lang.Package_name.Opam_compatible.t` for package names in package management code.


gridbugs:
It's possible to use invalid opam package names for dune package names in general (e.g. `foo.0.1`). When interoperating with opam we enforce opam's package naming rules but this is done inconsistently in our package management code as not all operations need to call into opam. It would be a more consistent user experience if all package management related operations require that package names be compatible with opam. The type `Dune_lang.Package_name.Opam_compatible.t` requires that the name be opam-compatible when values are constructed, so I think this is the type we should use to represent package names rather than the `Dune_lang.Package_name.t` type we currently use which doesn't require opam-compatibility.

gridbugs:
I agree that would be better but is there a risk that someone is depending on the current behaviour? I guess we could add the additional checks with a version check. @emillon I see you added `Package_name.Opam_compatible` quite recently. Is there a reason you added the new type rather than adding checks to the existing one?

rgrinberg:
> I agree that would be better but is there a risk that someone is depending on the current behaviour?

It could break people's projects, but given that the old behavior is useless anyway and no packages can be submitted to opam without a valid name, I don't see any issue with doing it.

How about we just export the type equality between our own `Package_name.t` and opam's? All this wrapping and unwrapping we currently have is really not buying us much in my opinion. We can still have our own proper validation on top. There's no contradiction here.

emillon:
> It could break people's projects, but given that the old behavior is useless anyway and no packages can be submitted to opam without a valid name, I don't see any issue with doing it.

When `(generate_opam_files)` is set, I agree but there's all the projects which use `(package)` and public names but without opam files. These can become broken.

rgrinberg:
Public names are always validated, and if they are invalid, then they would be rejected by findlib anyway when building something against those packages once they are installed.

Btw, there's no reason to use packages or public names unless the project has components that will get published to opam. So it seems highly unlikely that such packages exist.

Anyway, running an experiment to confirm or deny such suspicions seems like a good next step.

gridbugs:
[...]

PR #9385, opened 2023-12-05
Added rudimentary check on package names in dependencies

### LINKED ISSUE #9270: Writing foo.1.2.3 as a package dependency should give a nice error and a hint
In opam the convention is to write `foo.1.2.3` for a package version. For new users, it is not obvious that this doesn't work in Dune immediately. They may try something like this in `dune-project`:
```
(lang dune 3.13)
(package
 (name invalid)
 (depends foo.1.2.3))
```
Currently we raise an ugly uncaught exception when runnning `dune pkg lock`:
```
  Error: exception Failure("Invalid character in package name \"foo.1.2.3\"")
```
the reproduction case is in #9269. We should instead raise a nice user message with a hint with something like:
```
Hint: replace "foo.1.2.3" with "(foo (= 1.2.3))".
```

We don't even validate the package name at all when running `dune build`. The reproduction case can be found in #9271.

ElectreAAS:
Resolves #9270.

This PR adds a simple check to all the `<pkgname>` listed in the dependencies of a `dune-project`.

I just saw #9371, and have no opinions on it, but since they look closely related, I'm linking it here too.

Pinging @Leonidas-from-XIV & @emillon for a review.

### DIFF
--- a/doc/changes/cram-runtest-alias-version.md
+++ b/doc/changes/cram-runtest-alias-version.md
@@ -1,0 +1,1 @@
+- Fix version check in `runtest_alias` for `cram` stanza (#9454, @emillon)
--- a/src/dune_rules/cram/cram_stanza.ml
+++ b/src/dune_rules/cram/cram_stanza.ml
@@ -66,7 +66,7 @@ let decode =
      and+ runtest_alias =
        field_o
          "runtest_alias"
-         (Dune_lang.Syntax.since Stanza.syntax (3, 11) >>> located bool)
+         (Dune_lang.Syntax.since Stanza.syntax (3, 12) >>> located bool)
      in
      { loc; alias; deps; enabled_if; locks; applies_to; package; runtest_alias })
 ;;
--- a/test/blackbox-tests/test-cases/cram/runtest_alias.t
+++ b/test/blackbox-tests/test-cases/cram/runtest_alias.t
@@ -1,7 +1,7 @@
 Control the default runtest alias for cram tests
 
   $ cat >dune-project <<EOF
-  > (lang dune 3.11)
+  > (lang dune 3.12)
   > EOF
 
   $ cat >dune <<EOF


test/blackbox-tests/test-cases/invalid-package-version.t
```diff
@@ -13,6 +13,10 @@ opam.
 Building doesn't raise any error about the invalid version.
```
Leonidas-from-XIV, line 13: This should be updated now.

test/blackbox-tests/test-cases/invalid-package-version.t
```diff
@@ -13,6 +13,10 @@ opam.
...
 
   $ dune build
+  Error: Invalid character in package name "foo.1.2.3"
+  Hint: Package names may only contain letters, numbers, hyphens (-) and
+  underscores (_). Maybe you wrote "foo.1.2.3" instead of "(foo (= 1.2.3))".
+  [1]
 
 Even generating .opam files doesn't raise any error about the version.
```
Leonidas-from-XIV, line 21: Likewise.

Leonidas-from-XIV (changes requested):
Great to have your first PR! I've added a few comments, convenient that there are already some cram tests to test the change.

src/dune_rules/package.ml
```diff
@@ -415,6 +415,22 @@ let decode_name ~version =
   if version >= (3, 11) then Name.decode_opam_compatible else Name.decode
 ;;
 
+let check_pkgname (dep : Dune_lang.Package_dependency.t) =
+  let name = Name.to_string dep.name in
+  if String.exists name ~f:(function
+       | 'a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | '-' | '_' -> false
+       | _ -> true)
```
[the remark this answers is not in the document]
ElectreAAS: Realized a bit late that `decode_name` in the same file already does the same thing, I just needed to a bit of glue in between.

ElectreAAS:
Note: I only added the 'opam compatible' check in the `depends` field, but not the similarly formed `conflicts` and `depopts`. For the latter it seems to make sense, but the former?

Alizter:
@ElectreAAS We use `conflicts` as inputs to the solver IIRC so they should also be validated the same as `depends`.

ElectreAAS:
I just rebased the PR on main, sorry for the noise.

I have updated the expect part of [test/blackbox-tests/test-cases/pkg/invalid-version.t](https://github.com/ElectreAAS/dune/blob/main/test/blackbox-tests/test-cases/pkg/invalid-version.t), but not the text yet, as I'm quite confused as to what is the difference in tested behaviour in this test and [test/blackbox-tests/test-cases/invalid-package-version.t](https://github.com/ElectreAAS/dune/blob/main/test/blackbox-tests/test-cases/invalid-package-version.t).

Leonidas-from-XIV:
However should we care about `(depends)` if we're not generating opam files and never use the data?

Not sure how many packages specify depends without generating opam files, because without it it is fairly pointless at the moment.

PR #9388, opened 2023-12-06
refactor: allow config values to be defined anywhere

rgrinberg:
Expose a [Config.make] function to allow settings to be defined
anywhere.

Move dune rules specific configuration to [Dune_rules].

@snowleopard this should reduce some potential for syncing by allowing internal/external dune to keep private config options outside of the main module.


snowleopard (approved):
Looks good, thanks!

Issue #7680, opened 2023-05-04
[RFC] Dune Package Management


tmattio:
This document serves as a Request for Comments (RFC) on adding package management support to Dune.

At present, the project is in the prototyping stage, and we are exploring different approaches. Consequently, this document is intended to provide a guideline rather than a definitive specification for the implementation of package management in Dune. As we continue to prototype features and make progress, we will update this document accordingly.

The primary aim of this document is to stimulate community feedback and invite discussions about our plans.

We will generally stay high-level and describe the user experience rather than how we intend to implement each part. The specific implementation details can be discussed in separate issues and Pull Requests.

Outlined below are the high-level principles that guide our integration of package management into Dune:

1. **Seamless integration**: Package management features should be fully integrated. Users shouldn't have to read the opam manual to learn how to use package management features in Dune.
2. **Compatibility with existing features**: The package management features should work seamlessly with existing Dune features, such as watch mode. These features should work in an obvious way with package management.
3. **Windows support**: The first stable version of the package management system must include Windows support; it will not be considered stable until Windows is supported.
4. **Cross-compilation**: Support for cross-compilation should be a first-class citizen.
5. **Build directory isolation**: No state should exist outside the `_build` directory.
6. **Explicit state modification**: Commands that modify the state implicitly, such as pinning, are forbidden.
7. **Performance**: Features should be as fast as they can be.

**TL;DR**: Here are some elements from the RFC we'd like to highlight:

- We will cache the build of dependencies, including the compiler, eliminating the need to recompile the compiler for every project.
- Lockfiles will contain dependencies source and build instructions, with `dune build` using the lockfile by default, it will result in faster compilation as we don't need the solver or the opam repository.
- We will support cross-compilation for arbitrary opam packages, removing the need to port packages to Dune to leverage cross-compilation.
- `dune-workspace` files will be used to configure compiler versions through contexts, allowing seamless testing on different compiler versions.
- Windows support will be provided as a first-class citizen.
- We will support local sources for dependencies, enabling concurrent work on multiple projects. Changes in local sources will be detected, and dependent projects will be automatically rebuilt.

## Overview

Package management in Dune aims to improve the user experience on the OCaml Platform by addressing the friction caused by using separate opam and dune CLI tools.

From the user's perspective, downloading and building dependencies become an integral part of building a project. Dune becomes the primary tool that users install on their systems to work in OCaml. The workflow to go from a clean environment to a compiled project looks like:

```
$ apt install dune # or winget install dune
[...]

PR #9392, opened 2023-12-06
chore: disable windows cygwin job

Alizter:
Until https://github.com/ocaml/setup-ocaml/issues/731 is fixed, there is little we can do to allow this CI job to succeed.

The issue is that opam on Windows (cygwin) is stuck on version 2.0.10 which incorrectly pins opam files in an opam/ subdirectory. The fix exists in 2.1.x so setup-ocaml needs to update their opam version. However this is not any easy task, as it requires maintaining a no longer maintained cygwin opam repo. Therefore a fix won't happen soon.

We already run windows tests, of which there is only a single test, in the MSVC jobs. Therefore I think it is OK to disable this job for now.

Edit: The original issue I reported got moved, here is the new ticket:
- https://github.com/ocaml/setup-ocaml/issues/733

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -24,7 +24,7 @@ jobs:
         os:
           - macos-latest
           - ubuntu-latest
-          - windows-latest
+          #- windows-latest # renable once https://github.com/ocaml/setup-ocaml/issues/731 is fixed
         # Please keep the list in sync with the minimal version of OCaml in
         # dune-project, opam/dune.opam.template and bootstrap.ml
         #


emillon:
I'm not sure about this. We also have another option, which is to keep the opam files in the root directory until setup-ocaml works. 
The benefits of having opam files in a subdirectory are cosmetic, while having cygwin builds is actually useful to find problems on this configuration. With opam 2.2 around the corner and https://github.com/ocaml/dune/issues/7680 committing to first class support for package management, I think it's important to keep this tested.

emillon:
Closing in favor of #9412

PR #9398, opened 2023-12-06
fix(pkg): simplify associative slang operations

rgrinberg:

### DIFF
--- a/src/dune_lang/slang.ml
+++ b/src/dune_lang/slang.ml
@@ -136,12 +136,12 @@ let catch_undefined_var ?(loc = Loc.none) value ~fallback =
   Form (loc, Catch_undefined_var { value; fallback })
 ;;
 
-let and_absorb_undefined_var ?(loc = Loc.none) blangs =
-  Form (loc, And_absorb_undefined_var blangs)
-;;
-
 let or_absorb_undefined_var ?(loc = Loc.none) blangs =
   Form (loc, Or_absorb_undefined_var blangs)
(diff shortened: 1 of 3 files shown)


Alizter (approved):

src/dune_lang/slang.ml
```diff
@@ -137,10 +137,22 @@ let catch_undefined_var ?(loc = Loc.none) value ~fallback =
 ;;
 
 let and_absorb_undefined_var ?(loc = Loc.none) blangs =
+  let blangs : blang list =
```
gridbugs, line 140: This function is named after the `And_absorb_undefined_var` constructor which suggests that all it does is invoke the constructor (at least that was my intention when naming it). I think that doing additional processing to its arguments (even if it's benign as far as the value of the expression is concerned) may lead to confusion when debugging generated slang expressions. I think the `simplify` function in this module would be a more suitable place for this logic.

rgrinberg: If you look at `Predicate_lang`, we put similar simplifications in such smart constructors. I personally like that it makes it impossible to construct non-canonical values (at least if we make the type private). I can move this to `simplify` though

rgrinberg: Moved everything to simplify. I still like making the "smart" constructors a little smarter, as it's helpful to know that you're always dealing with simplified expressions, but for this is more consistent.

gridbugs (approved):

PR #9405, opened 2023-12-07
chore: replace print_endline with Console.print

Alizter:


rgrinberg:
Actually the way it is now is wrong. The output of `dune describe` should be written to stdout to make this command usable in common shell pipelines. If we're using the console to print this output in other describe commands, we should fix those.

Alizter:
@rgrinberg I will close this then. I don't know how useful it will be to adapt Console to print to stdout optionally, seems like an unnecessary complication.

PR #9407, opened 2023-12-07
Refactor Dune_engine.Targets

rleshchinskiy:
This is a bunch of refactoring around `Targets` and especially `Targets.Produced`. The main changes are:

- We had similar but subtly different code for sandboxed and non-sandboxed directory targets which is now unified. As part of this, we no longer promote empty (sub)directories. Previously, this was only the case for sandboxed directory targets because we didn't move empty (sub)directories from the sandbox to the build dir.
- We generalise `Targets.Produced.collect_digests` which allows us to remove `Targets.Produced.Option`.
- We move everything to do with async file operations in `Build_system` which really is the only user of those, anyway.
- We remove all dependencies on `Dune_engine` from `Targets`. A later PR will move those into a separate library which will allow us to use them in the cache as well.

### DIFF
--- a/otherlibs/stdune/src/fpath.ml
+++ b/otherlibs/stdune/src/fpath.ml
@@ -161,7 +161,7 @@ let rm_rf fn =
 ;;
 
 let traverse_files =
-  let rec loop root stack acc f =
+  let rec loop root stack acc on_file on_dir =
     match stack with
     | [] -> acc
     | dir :: dirs ->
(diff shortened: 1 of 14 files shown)


snowleopard (approved):
Thanks! All reviewed, benchmarked and battle-tested internally.

rgrinberg (approved):
I've replaced the use of symlinking to create a directory target with just normal copying as an experiment and it didn't seem to help. I've also experimented with sandboxing to make sure that we specified all dependencies correctly and again without success.

The fact that doing a clean build gives us the right answer in the install rules is rather concerning and my first assumption is that there's some bug in the engine.

Unfortunately, I didn't get anywhere investigating. I'll keep investigating, but I don't think this should block the PR.

I've left a commit (which I later reverted) switching the symlinking to copying in case anyone wants to experiment.

PR #9410, opened 2023-12-07
fix(pkg): remove experimental filter translation option

rgrinberg:
Nothing works without this option is enabled, so we might as well make
it the default.


test/blackbox-tests/test-cases/pkg/helpers.sh
```diff
@@ -84,9 +84,7 @@ solve_project() {
 solve_project_translate_opam_filters() {
   cat >dune-project
   add_mock_repo_if_needed
-  dune pkg lock \
-    --dont-poll-system-solver-variables \
-    --experimental-translate-opam-filters
+  dune pkg lock --dont-poll-system-solver-variables
 }
```
Alizter, line 88: This can all be removed

gridbugs (approved):

PR #9418, opened 2023-12-08
pkg: read local package metadata from opam files

### LINKED ISSUE #9341: Support dune projects with only opam metadata
Currently, we read all package metadata from the dune-project file. While that's correct thing to do when dune-project contain all the metadata, we completely fail for projects where this metadata only lives in the .opam file. Instead, we should fallback to .opam files when the project files don't define the relevant packages.

cc @tmattio

gridbugs:
Fixes https://github.com/ocaml/dune/issues/9341


rgrinberg:
Merged with a few improvements.

PR #9421, opened 2023-12-08
doc: move reference info about cram test stanza

### LINKED ISSUE #9211: Documentation of cram stanza should be moved to stanza reference
Currently the documentation of the cram stanza lives in a how-to guide rather than in the main stanza reference. This should be moved to the stanza reference. The how-to guide should just be about examples and discussion rather than a point of reference.

emillon:
It goes with the rest of the stanza documentation and we can use the domain.

Fixes #9211

### DIFF
--- a/doc/stanzas/cram.rst
+++ b/doc/stanzas/cram.rst
@@ -11,8 +11,6 @@ cram
    cases, the values from all applicable ``cram`` stanzas are merged together
    to get the final values for all the fields.
 
-   .. seealso:: :doc:`reference/cram`
-
    .. dune:field:: deps
       :param: <dep-spec>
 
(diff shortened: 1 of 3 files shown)


doc/stanzas/cram.rst
```diff
@@ -0,0 +1,88 @@
...
+
+   .. dune:field:: package
+      :param: <name>
+
+      Attach the tests selected by this stanza to the specified package.
+
+   .. dune:field:: runtest_alias
+      :param: <true|false>
+
+      .. versionadded:: 3.12
```
Alizter, line 84: This should be 3.11

emillon: No, the field has actually been added in 3.12. The version check is fixed in #9454.

christinerose (approved):

PR #9422, opened 2023-12-08
doc: add reference info for cram tests

emillon:
This document explains cram tests themselves, not only the stanza and how to set them up.

### DIFF
--- a/doc/reference/cram.rst
+++ b/doc/reference/cram.rst
@@ -1,0 +1,97 @@
+Cram Tests
+==========
+
+Syntax
+------
+
+Cram tests are interpreted line by line, depending on the first characters of
+each line:
+
+- if a line starts with two spaces, it is part of a command or its output.
+
+  - if the next two characters are a dollar sign ``$`` and a space character,
+    the rest is a command.
+
+  - if the next two characters are a greater-than sign ``>`` and a space
+    character, the rest is the continuation of a command.
+
+  - otherwise (line starts with two spaces and something else), this is the
+    expected output of the command just before.
+
+- otherwise (line does not start with two spaces), this is a comment.
+
+Ignoring comments, a cram test is composed of a list of commands (1 line), with
+optional continuation lines, and with optional output lines.
+
+.. code:: cram
+
+   Plain paragraphs that do not start with spaces are ignored by cram tests.
+   They are often used to introduce commands. For example, this is a command:
+
+     $ touch this-file.txt
+
+   The above is the simplest command: it has no continuation lines and no output.
+   Some commands have an expected output:
+
+     $ ls
+     this-file.txt
+
+   There can be several output lines if the command is expected to print
+   several lines.
+   Also, note that if a command has no output, the next one can come in the
+   next line.
+
+     $ touch other-file.txt
+     $ ls
+     other-file.txt
+     this-file.txt
+
+   Continuation lines are used when a command fits on several lines. This can
+   happen in all the cases where pressing Enter would not run the command. For
+   example, when passing a backslash character to escape the line ending.
+   In that case, all the continuation lines are grouped together as a single
+   command.
+   This syntax mimics the PS2 prompt in shells - the ">" character is not
+   passed to the command.
+
+     $ echo \
+     >   a \
+     >   b \
+     >   d \
+     >   c
+     a b c d
+
+Semantics
+---------
+
+Execution and Promotion
+^^^^^^^^^^^^^^^^^^^^^^^
+
+When a cram test is executed, the commands it contains are executed, and a
+corrected file is created where the outputs of the commands are inserted after
+each command. This corrected file is then offered for
+:doc:`promotion <../concepts/promotion>` by Dune.
+
+Concretely, this means that ``dune runtest`` will display the difference
+between the current contents of the cram test, and the output of the latest
+run. This diff can be applied by running ``dune promote``, as usual.
+
+.. code:: diff
+
+   $ touch changed-name.txt
+   $ ls
+  -other-file.txt
+  +changed-name.txt
+   this-file.txt
+
+File and Directory Tests
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+There are two types of cram tests: file tests and directory tests. File tests
+are files with a ``.t`` extension. Directory tests are files named ``run.t``
+within a directory with a name that ends with ``.t``.
+
+A cram test begins its execution in a temporary directory where its
+dependencies (as listed in the corresponding :ref:`cram stanzas <cram-stanza>`,
+if any) are available. In the case of a directory test, the contents of the
+directory are also available.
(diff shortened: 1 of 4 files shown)


doc/reference/cram.rst
```diff
@@ -0,0 +1,97 @@
+Cram Tests
+==========
+
+Syntax
+------
+
+Cram tests are interpreted line by line, depending on the first characters of
+each line:
+
+- if a line starts with two spaces, it is part of a command or its output.
```
Alizter, line 10: Shouldn't these start with a capital letter?

Alizter:
Should we also add an example of testing a hello world binary? This would be good to show demonstrate the cram stanza and how binaries are added to `env`.

rgrinberg:
There's a bit of redundancy here with the description of cram tests in tests.rst.

emillon:
> Should we also add an example of testing a hello world binary? This would be good to show demonstrate the cram stanza and how binaries are added to `env`.

That's more in scope for an howto guide. What we have in `tests.rst` with the `wc` example looks closer to that and can recycled into a good howto guide. This is more about how the syntax works and how cram tests are executed. I initially wanted to describe heredocs etc here but this are more howto territory.

emillon:
> There's a bit of redundancy here with the description of cram tests in tests.rst.

I replaced that with a xref

christinerose (approved):

createyourpersonalaccount:
@Alizter Thanks I will respond to this soon.

createyourpersonalaccount:
I'm commenting on @emillon's branch. I wrote everything I could come up with, but the usefulness of these comments is up to you to decide:

# On [doc/hacking.rst](https://github.com/emillon/dune/blob/c95087ba59e42722ee36b8a57ba1e1c1cc320ddd/doc/hacking.rst):

Under [Bootstrapping](https://github.com/emillon/dune/blob/c95087ba59e42722ee36b8a57ba1e1c1cc320ddd/doc/hacking.rst#bootstrapping), the last paragraph starting with `Note that tests are currently written for` is not an ideal situation. It would be better if a test failed with assert message "version X.Y.Z required" or if the test is skipped for other versions.

Under [Writing Tests](https://github.com/emillon/dune/blob/c95087ba59e42722ee36b8a57ba1e1c1cc320ddd/doc/hacking.rst#writing-tests) the entire text from `Most of our tests are written ...` up to `We write both our unit tests and integration tests in this way.` could be omitted: developers understand this information. The less-familiar Cram tests will be introduced shortly and they are explained there.

The sentence

> For unit tests, we use the [ppx_expect](https://github.com/janestreet/ppx_expect) framework, where we introduce tests via `let%expect_test`, and `[%expect ...]` nodes capture expectations: 

could be improved to

> For unit tests, we use the [ppx_expect](https://github.com/janestreet/ppx_expect) framework, where we introduce tests via `let%expect_test`, and where we use `[%expect ...]` nodes to capture expectations: 

In

> For integration tests, we use a system similar to [Cram tests](https://bitheap.org/cram/) for testing shell commands for their expected output.

In the immediate Cram test example, I would split into three examples, with comments (in the rst document, not cram-style comments) describing each test very briefly:

> Test the output of `echo`:
>
> ```
> $ echo 'Hello, world!'
> Hello, world!
> ```
>
> Test the return value of `false`:
> 
> ```
> $ false
> [1]
> ```
>
> A multi-line shell invocation:
>
> ```
> $ cat <<EOF
> > multi
> > line
> > EOF
> multi
> line
> ```

Is bash syntax supported? This should be clarified at this point, and even mentioned every time Cram tests are introduced in the docs (i.e. with "Cram tests are ... with bash-like syntax").

At this point in the document, it might be fruitful to link to a simple and a complex Cram test in the source tree, and a bit earlier it might be useful to link to a `ppx_expect` use in the source tree.  (e.g. "for good examples the developer is invited to see this file for unit tests and this file for integration tests")

# On [doc/reference/cram.rst](https://github.com/emillon/dune/blob/c95087ba59e42722ee36b8a57ba1e1c1cc320ddd/doc/reference/cram.rst)

The Syntax/Semantics headings are a bit artificial: semantics are described in syntax, and the cram test example definitely shows off the semantics. I would suggest simply removing these headings, or restricting the Syntax heading to only the list of descriptions of characters.

The document begins by describing how cram tests are parsed, however, it would be useful to first give a synopsis of what cram tests are, which as of my current understanding, are integration tests that compare actual shell command output to expected output.

The sentence

> Otherwise (line does not start with two spaces), this is a comment.

could be improven to

> Otherwise if the line does not start with two spaces it is a comment.

The meaning of the next sentence is not clear to me:

> Ignoring comments, a Cram test is composed of a list of commands (1 line), ...

I think it means that Cram tests have one command *per* line. I would just delete this sentence (it is already described in the list above) and instead I would simply say that the expected output is *optional*; and I would describe a scenario, or a class of situations, in which this might be useful. Personally, I think this is useful for setting up various things in the Cram test environment for example.

The Cram test example that follows should be prefaced with `An example:` or similar.

Under [Execution and Promotion](https://github.com/emillon/dune/blob/c95087ba59e42722ee36b8a57ba1e1c1cc320ddd/doc/reference/cram.rst#execution-and-promotion)

When the corrected file semantics are described, it may be good to highlight this with a bold NOTE or something similar that draws attention. It's quite unusual in my opinion. I think I was getting these promotions applied automatically with `dune build`, which is the recommended way to run Cram tests (as of when I last saw the documentation on them). Speaking of which, this file does not describe how to run cram tests with Dune. Where is this described? **Please make sure that invocation of Cram tests with Dune is described somewhere** (I may have just missed it).

Under [Files and Directory Tests](https://github.com/emillon/dune/blob/c95087ba59e42722ee36b8a57ba1e1c1cc320ddd/doc/reference/cram.rst#file-and-directory-tests) it would be good to explain briefly what a directory test would be useful for. Why would one want files to be available? E.g. it allows for test assets or whatnot.

emillon:
- hacking.rst (not this PR):
  - [ ] remove "note that tests..."
  - [ ] add a test that asserts on test-version of ocaml
  - [ ] add an alias/makefile target to run only said test (will be useful for support requests
  - [ ] reword part about `[%expect]`
  - [ ] split cram example
  - [ ] clarify whether bash is supported
  - [ ] link to a `ppx_expect` example
  - [ ] link to a cram example
- cram.rst (this PR):
  - [ ] better syntax/semantics split
  - [ ] add synopsis first
  - [ ] "Otherwise..." part
  - [ ] 1 line per command part
  - [ ] add "an example" header
  - promotion
    - [ ] highlight part about accepting promotions
    - [ ] how to run a cram test
  - [ ] why would you use a directory test
  - [ ] document return values

christinerose (approved):

emillon:
I incorporated @createyourpersonalaccount's feedback into that PR and created #9734 for the rest. thanks

PR #9427, opened 2023-12-08
refactor: replace Result.to_dyn with Dyn.result

Alizter:


rgrinberg:
This isn't necessary. The old form is fine

PR #9430, opened 2023-12-09
feature(pkg): depopts

### LINKED ISSUE #9415: Depopts Are Broken
It appears that there's no way to make 0install select depopts at the moment

Alizter:
fix #9415 
- [ ] tests
- [ ] update fork

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -466,6 +466,7 @@ let opam_package_to_lock_file_pkg
   solver_env
   version_by_package_name
   opam_package
+  (opam_packages_to_lock : OpamPackage.t list)
   ~(candidates_cache : (Package_name.t, Context_for_dune.candidates) Table.t)
   =
   let name = OpamPackage.name opam_package in
@@ -519,14 +520,14 @@ let opam_package_to_lock_file_pkg
     ; extra_sources
     }
   in
-  let deps =
+  let deps_from_filtered_formula deps =
     match
       Resolve_opam_formula.filtered_formula_to_package_names
         ~stats_updater:None
         ~with_test:false
         solver_env
         version_by_package_name
-        opam_file.depends
+        deps
     with
     | Ok dep_package_names ->
       List.map dep_package_names ~f:(fun package_name -> Loc.none, package_name)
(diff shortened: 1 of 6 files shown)


rgrinberg:
For now, you can try desugaring optional dependencies the way I was told by Kate. E.g.

```
depends : [ "x" ]
depopts : [ "a" "b" ]
```

Is equivalent to:

```
depends: [
  ("x" & "a") | ( "x" & "b") | ("x" & "a" & "b")
]
```

Alizter:
@rgrinberg Since the solver reads the opam files directly, what is the best way to do this desugaring? I can do it for local packages, but I don't see how to do it for the solver.

vendor/opam-0install/lib/model.ml
```diff
@@ -193,7 +192,8 @@ module Make (Monad : S.Monad) (Context : S.CONTEXT with type 'a monad = 'a Monad
                   |> list_deps ~context ~importance
                 in
                 make_deps `Essential ensure OpamFile.OPAM.depends @
-                make_deps `Restricts prevent OpamFile.OPAM.conflicts
+                make_deps `Restricts prevent OpamFile.OPAM.conflicts @
+                make_deps `Recommended ensure OpamFile.OPAM.depopts
```
rgrinberg, line 196: @kit-ty-kate is it ok to support depopts like this?

src/dune_pkg/opam_solver.ml
```diff
@@ -537,6 +538,19 @@ let opam_package_to_lock_file_pkg
         ; "hints", Dyn.list Resolve_opam_formula.Unsatisfied_formula_hint.to_dyn hints
         ]
   in
+  let deps = deps_from_filtered_formula opam_file.depends in
+  let depopts =
+    (* We need to filter out depopts that don't appear in the build plan. *)
+    OpamFormula.map
+      (fun (name, x) ->
+        match
+          List.find_opt ~f:(fun p -> OpamPackage.name p = name) opam_packages_to_lock
```
kit-ty-kate, line 547: `OpamVersion.compare` is different from the polymorphic `Stdlib.compare`
```suggestion
          List.find_opt ~f:(fun p -> OpamPackage.Name.equal (OpamPackage.name p) name) opam_packages_to_lock
```

vendor/opam-0install/lib/model.ml
kit-ty-kate: I don’t think so. That sounds like a change in semantics from what opam does. See the note in https://opam.ocaml.org/doc/Manual.html#opamfield-depopts

kit-ty-kate (commented):
By curiosity, if you’re using opam-0install, why are you not using https://github.com/ocaml/opam/blob/master/src/solver/opamBuiltin0install.ml directly?

vendor/opam-0install/lib/model.ml
rgrinberg: I find the following line hard to interpret in the case of dune:

> It contains packages that will be used, if present, by the package being defined, either during build or runtime, but that are not required for its installation

The only packages that are present when building the packages that are only those requested by the user (e.g. the `depends` field). So there's no way for a depopt to be opportunistically present this way.

kit-ty-kate: mmh, thinking about it again, the change of semantics should be fine for dune actually.

rgrinberg:
I think this has been addressed.

PR #9431, opened 2023-12-09
test: make sure depopts are optional

Alizter:
This adds a simple test that ensures we are treating depopts optionally in the solver. When we support depopts in #9430, this test shouldn't change. (And it doesn't)

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/depopts-optional.t
+++ b/test/blackbox-tests/test-cases/pkg/depopts-optional.t
@@ -1,0 +1,21 @@
+This tests that depopts are considered optional by the solver.
+
+  $ . ./helpers.sh
+  $ mkrepo
+  $ mkpkg foo
+  $ mkpkg bar
+
+We create a package that depends on foo and has an optional dependency on bar
+but also a conflict with bar. This should ensure that bar is never selected by
+the solver.
+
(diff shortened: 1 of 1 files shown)


rgrinberg:
I think this test would be more useful if it used version constraints for `bar`. With the following situations:

1. There's a depopt on `bar` but we conflict with the latest version. Therefore, the older version is still available to us.
2. There's a depopt on `bar` but we conflict with all available versions. Therefore, the depopt should be omitted.

Alizter:
@rgrinberg I have such a test case in #9430 but it appears it doesn't work correctly. This simple case is useful on its own however, since it shouldn't change regardless of how we support depopts.

rgrinberg:
Aren't all the tests useful on their own? Why merge only a subset of them?

Alizter:
@rgrinberg I'm not suggesting merging a subset, I'm just saying this test is doing something very particular and a test demonstrating the behaviour you are asking for will appear in another PR.

rgrinberg:
Okay, let's see them in a subsequent PR. Also, could you take the opportunity to re-organize the depopts into a single directory with a single test file per test case? That will make it much easier for us to follow.

Alizter:
Moved in https://github.com/ocaml/dune/pull/9433

PR #9434, opened 2023-12-09
test: depopts with constraints

Alizter:
This test demonstrates a situation where a depopt should eventually be avoided when a more up to date but conflicting (with the depopt) build plan is available.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/depopts/depopts-with-constraints.t
+++ b/test/blackbox-tests/test-cases/pkg/depopts/depopts-with-constraints.t
@@ -1,0 +1,39 @@
+We test how the solver chooses a solution for depopts with constraints.
+
+  $ . ../helpers.sh
+  $ mkrepo
+
+We create a package "foo" that is the dependency of two packages "optional" and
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/pkg/depopts/depopts-with-constraints.t
```diff
@@ -0,0 +1,38 @@
...
+  > EOF
+
+The package "bar" depends on "foo" and depends optionally on "optional".
+
+  $ mkpkg bar <<'EOF'
+  > depends: [ "foo" ]
+  > depopts: [ "optional" ]
+  > EOF
+
+Here we expect the solver to choose all 3 packages as a build plan.
```
rgrinberg, line 24: We have no such expectation. It's within our right to ignore the optional dependency. It's best to describe what we're doing here rather than speculating how it should work.

Alizter: I've updated the comment.

PR #9439, opened 2023-12-09
[coq] Allow to disable Coq language deprecation warning.

ejgallego:
I had to adapt the warnings infra a little bit, IMHO that is fine, tho it would be nice to see if we can recover the vendor behavior.

I also corrected the hint message a little bit, as it was a bit confusing IMHO, but will place that in the next commit as to keep diffs tidy.

### DIFF
--- a/doc/changes/9439_coq_disable_08_warning.md
+++ b/doc/changes/9439_coq_disable_08_warning.md
@@ -1,0 +1,1 @@
+- Allow to disable Coq 0.8 deprecation warning (#9439, @ejgallego)
--- a/src/dune_rules/coq/coq_stanza.ml
+++ b/src/dune_rules/coq/coq_stanza.ml
@@ -18,16 +18,31 @@ let coq_syntax =
 
 let already_warned = ref false
 
+let deprecated_coq_lang_lt_08 =
+  Warning.make
+    ~default:(fun _version -> `Enabled)
+    ~name:"deprecated_coq_lang_lt_08"
+    ~since:(3, 8)
+;;
+
+let deprecated_coq_lang_lt_08_msg =
+  User_message.make
+    [ Pp.text
+        "Coq Language Versions lower than 0.8 have been deprecated in Dune 3.8 and will \
+         be removed in an upcoming Dune version."
+    ]
+;;
+
 let get_coq_syntax () =
   let* version = Dune_lang.Syntax.get_exn coq_syntax in
+  let* project = Dune_project.get_exn () in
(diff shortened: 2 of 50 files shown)


ejgallego:
An alternative to this is to add a `Deprecated_in` constructor to `Syntax.Supported_versions`, tho the logic would seem mainly orthogonal, as we would still need to switch the deprecation warning to the `Warning_emit` interface.

Alizter:
Why do we want users to disable this warning?

ejgallego:
> Why do we want users to disable this warning? i.e. what's the point of the warning if it just gets ignored?

I am not sure I understand your question @Alizter , doesn't what you say apply to all warnings?

There are many reasons to disable this warning, in particular for example in Coq we still require a lower bound of 3.6

Alizter:
@ejgallego Then I am a bit confused why we marked it as deprecated to begin with?

ejgallego:
We marked it as deprecated because it is going a) broken b) going to be removed in a future version.

rgrinberg (approved):
LGTM. The code looks fine. I didn't delve into the details of what is the warning about or if it's needed, but the way it's introduced and used is correct.

ejgallego:
Thanks! I'll ask Ali in private about his concerns with the warning and merge once he's OK.

src/dune_rules/warning.ml
```diff
@@ -95,3 +95,10 @@ let escaping_paths_in_install_stanza =
 ;;
 
 let no_keep_locs = create ~default:(fun _ -> `Enabled) ~name:"no_keep_locs" ~since:(3, 11)
+
+let deprecated_coq_lang_lt_08 =
```
rgrinberg, line 99: Actually, one small request: could you move this to somewhere in `dune_rules/coq`. I'd like to keep all the extensions separate

ejgallego: I can move it, but note that I will have to export `Warning.create` too.

rgrinberg: I think that's fine. We need a public API for defining warnings anyway.

ejgallego:
Pushed the changes to infra to other PRs, @Alizter I've sent you a message to try to understand what your concerns are.

Other than that, I think this is ready.

ejgallego:
Ali point was "why people cannot upgrade to Coq 0.8", and indeed, most people should, but in the real world a few people is stuck due to poor reasons. So we let them opt out of the warning for now, at their own risk.

PR #9440, opened 2023-12-09
test: conflicting depopts

Alizter:
@rgrinberg This test demonstrates the behaviour of depopts when there is a conflict between them. It appears that 0install eagerly discards them after #9430. There are <s>3</s> 2 situations that could come out of this test:

1. We choose no depopts, as is currently (after #9430) the behaviour.
2. We choose 1 depopt, probably the latest version. This behaviour seems the most sensible, but would require changes in 0install.
3. <s>We choose both depopts by allowing for "local dependencies". This would make the depopts as only being visible to the package being built, allowing multiple versions to be installed at once. This would be a separate feature, and probably has some of its own difficulties.</s> This won't work.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/depopts/depopts-with-conflicting-constraints.t
+++ b/test/blackbox-tests/test-cases/pkg/depopts/depopts-with-conflicting-constraints.t
@@ -1,0 +1,23 @@
+We test depopts with conflicting constraints to see which one the solver will
+prefer if any:
+
+  $ . ../helpers.sh
+  $ mkpkg foo 1
+  $ mkpkg foo 2
+
+  $ mkpkg bar <<'EOF'
+  > depopts: [ "foo" {= "1"} ]
+  > EOF
+
+  $ mkpkg baz <<'EOF'
+  > depopts: [ "foo" {= "2"} ]
+  > EOF
+
+We don't currently support depopts so they are both omitted.
+
+  $ solve bar baz
+  Solution for dune.lock:
+  - bar.0.0.1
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/pkg/depopts/depopts-with-conflicting-constraints.t
```diff
@@ -0,0 +1,27 @@
...
+  > EOF
+
+We don't currently support depopts so they are both omitted by omission.
+
+  $ solve bar baz
+  Solution for dune.lock:
+  - bar.0.0.1
+  - baz.0.0.1
+
+In the future, we could support "local depopts" that are only in scope for
```
rgrinberg, line 23: This description is wrong. It will never be possible to install multiple depopts side by side and all 4 packages. What is possible is selecting a version of `foo` to satisfy `bar` (or `baz`), and then building `baz` (or `bar`) without the depopt.

Alizter: OK that's 3, in my suggestions above. I think I'll update the description to 2, as that is what you are describing.

rgrinberg: 3. is not possible. You will get libraries that cannot be linked together

Alizter: OK, I've updated the description to mention 2 instead.

PR #9447, opened 2023-12-10
fix(init): correct behavior of PATH argument for `dune init project`

### LINKED ISSUE #9209: Initialize new project in current directory
## Desired Behavior

I would like to be able to run `dune init proj NAME` and have it create my project in place (ie, not a subdirectory ./NAME). 

### Current Behavior / Problem Space
Currently `dune init proj NAME` requires a NAME parameter and creates a subdirectory.  Every time I've used dune to initialize a project, I am starting from an empty git repo, and running dune from the repo's root.  This creates the dune project nested down a level which I don't want.  I work around this by moving the contents of the newly created subdirectory up to the repo's root, and then deleting the now empty `./NAME` directory that was created by dune.   This is uncomfortable for someone who's new to the language and is spinning up new projects that are very small in scope just to learn (ie: me). 

## Example
Intended functionality would be similar to [`nix flake initi`](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake-init).
Example workflow for initializing project:
```
# Create a new bare repo on github.com
git clone my/repo
nix flake init # included to show how I expected dune to work
dune init proj MY_PROJ --in-place
# write some ocaml
```
## Initial Considerations
My initial intuition when using dune was just to run `dune init proj` (omitting the name parameter) as this is how other tooling like [`nix flake init`](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake-init) works.  Because the name parameter gets used for more then just the path in which to create the dune project, my suggestion is to add an optional flag that specifies that I want my dune project created in-place.   Something like `--in-place` is narrow in scope and behavior, or `--path=/my/custom/path`  which could give more flexibility to the tool at the cost of intuitiveness.

shonfeder:
Fixes #9209

The logic around the `PATH` argument for the `dune init project` subcommand was incorrect. This was creating an unnecessary limitation on the behavior of the sub command, leading to #9209, and inconsistent behavior, diagnosed in https://github.com/ocaml/dune/issues/9209#issuecomment-1817510486.

This was the error: when the `PATH` argument was give to `dune init project`, the `NAME` argument was being appended as the final part of the path. As a result, instead of initializing the project in the specified directory, we were initializing it in a subdirectory of the given directory, bearing `NAME`. As an illustrative example, this led to the confusing behavior that both `dune init proj foo` and `dune init proj foo .` initialized projects in a newly created directory `./foo/`.

This error went undetected for so long in part because it was not being exercised in test coverage and in part because the intended behavior was underspecified in the documentation. I have tried to fix both of these oversights.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -77,7 +77,31 @@ the file and directory, if necessary):
      (preprocess
       (pps ppx_inline_tests)))
 
-Consult the manual page using the ``dune init --help`` command for more details.
+Initializing Components in a Specified Directory
+------------------------------------------------
+
+All ``init`` subcommands take an optional ``PATH`` argument, which should be a
+path to a directory. When supplied, the component will be created in the
+specified directory. E.g., to initialize a project in the current working
+directory, run
+
+.. code:: console
+
+    $ dune init proj my_proj .
+
+To initialize a project in a directory in some nested path, run
+
+.. code:: console
+
+    $ dune init proj my_proj path/to/my/project
+
+If the specified directory does not already exist, it will be created.
+
+Learning More About the ``init`` Commands
+-----------------------------------------
+
+Consult the manual page using the ```dune init --help`` command for more
+details.
 
 .. _finding-root:
 
@@ -309,8 +333,8 @@ must be prefixed by the shortest one.
 
 .. _dune-subst:
 
-dune subst
-==========
+``dune subst``
+==============
 
 One of the features ``dune-release`` provides is watermarking; it replaces
 various strings of the form ``%%ID%%`` in all your project files
--- a/test/blackbox-tests/test-cases/dune-init.t/run.t
+++ b/test/blackbox-tests/test-cases/dune-init.t/run.t
@@ -9,7 +9,7 @@ Setup mutable files
   >   (name test_lib))
   > EOF
 
-Adding a library
+Adding a Library
 ----------------
 
 Can init a public library
@@ -61,7 +61,7 @@ Clean up library with inlines tests
 
   $ rm -rf ./_inline_tests_lib
 
-Adding an executable
+Adding an Executable
 --------------------
 
 Can init a public executable
@@ -90,7 +90,7 @@ Clean up the executable tests
 
   $ rm -rf ./_test_bin_dir
 
-Adding tests
+Adding Tests
 ------------
 
 Can init tests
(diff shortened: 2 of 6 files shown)


test/blackbox-tests/test-cases/dune-init.t/run.t
```diff
@@ -144,6 +144,16 @@ Clean up from the dune file created at an absolute path
 
   $ rm -rf $PWD/_test_dir
 
+Add a library in the current working directory
+
+  $ dune init lib test_cwd_lib .
```
christinerose, line 149: Should these be in code block? If so, I can do that.

shonfeder: This is the correct syntax for our cram-style tests. Thank you tho :)

shonfeder:
Looks like you'll need to sign off on your commits @christinerose :)

christinerose:
> Looks like you'll need to sign off on your commits @christinerose :)

Oh no! I forgot! Thanks for the reminder.

christinerose:
> Looks like you'll need to sign off on your commits @christinerose :)

I'm so sorry, but I can't figure out how to do that! It's probably easier to just delete the commits and I'll redo it. It didn't take me long. 🙁

shonfeder:
I can fix it on my side if you don't me adding my sign off.

christinerose:
> I can fix it on my side if you don't mind me adding my sign off.

Not at all! Please do!

doc/usage.rst
```diff
@@ -9,9 +9,6 @@ This section describes using ``dune`` from the shell.
 Initializing Components
 =======================
 
-NOTE: The ``dune init`` command is still under development and subject to
-change.
-
```
Leonidas-from-XIV, line 14: Why is this being removed?

shonfeder: It's been there for years now, and afaik the init subcommand is not liable to any more radical changes at this point then most other parts of the CLI.

But if you all have changes planned to justify this continued earning, I'm happy to revert this part.

Leonidas-from-XIV: I think there's indeed some plans on what to do with `dune init` in the future, so I think we should't consider this done and set in stone, especially as the post-1.0 parts of dune have rather strong backwards-compatibility guarantees.

Leonidas-from-XIV (approved):
I think this behavior makes more sense than the one before and aligns better with user expectations.

doc/usage.rst
shonfeder: Ok! I've reverted this.

shonfeder:
Thanks for the review, @Leonidas-from-XIV!

shonfeder:
This should be ready for merge.

PR #9460, opened 2023-12-11
Update opam-file-generation.rst - Added line about generating only the opam file

rikimasan:
Wanted to add a line about generating only the opam file since this is very useful for projects that are in a state that can't build without the dependency being installed among other situations where this can be useful. Seemed like something that should be stated in the docs.


emillon:
Thanks!

Issue #9456, opened 2023-12-11
Incorrect dependency handling in new odoc rules.


jonludlam:
## Expected Behavior

The new odoc rules should build the odoc files in dependency order, with correct include paths.

## Actual Behavior

When building docs for a package containing sublibraries, if the main library depends upon the sublibraries, these are not built in the correct order. For example, `core` depends upon `core.filename_base`, but when building the docs for core, this is not reflected in the build order nor include path.

## Reproduction

Build the docs for core and check the include paths for the odoc invocations. Indirectly the simplest way to observe this is to create a package with a sublibrary, and in the main library create a module whose type is found via `module type of` a module in the sublibrary. The module in the main library will not have an expansion. PR with this test coming soon!

PR #9461, opened 2023-12-11
Fix for #9456

jonludlam:


rgrinberg:
Thanks. Could you add a CHANGES entry?

PR #9466, opened 2023-12-12
chore(pkg): clean up imports in bin/

Alizter:


rgrinberg:
Exposes too many pkg modules where they aren't needed.

PR #9467, opened 2023-12-12
doc: start documenting package management features internally

Alizter:
This is the start of some documentation on package management features. Suggestions to add topics to this are welcome. The goal is to collect documentation as the feature has grown quite large and we may forget some details in the future when attempting to write user facing documentation.

Let me know if there is anything important that you deem missing.

### DIFF
--- a/doc/dev/pkg.md
+++ b/doc/dev/pkg.md
@@ -1,0 +1,129 @@
+# Package Management in Dune
+
+- [Package Management in Dune](#package-management-in-dune)
+  - [Introduction](#introduction)
+  - [Reference](#reference)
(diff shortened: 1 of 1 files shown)


emillon:
I'm fine with merging this but I would recommend using the wiki for low-barrier, non-user-facing, collaborative writing.

Alizter:
@emillon I'm not a big fan of keeping such documentation in the Wiki because its easy to forget about it. Having it in the repo has the advantage of seeing documentation changes next to PRs that change behaviours.

I also don't think we have a requirement for this to be low barrier since it is an internal developer documentation that will be replaced by a user facing one in due time.

PR #9470, opened 2023-12-12
Forbid empty directory targets

rleshchinskiy:
We no longer allow directory targets to be empty. This wasn't well supported before and interferes with the upcoming caching support.

### DIFF
--- a/src/dune_engine/targets.ml
+++ b/src/dune_engine/targets.ml
@@ -108,6 +108,7 @@ module Produced = struct
   module Error = struct
     type t =
       | Missing_dir of Path.Build.t
+      | Empty_dir of Path.Build.t
       | Unreadable_dir of Path.Build.t * Unix_error.Detailed.t
       | Unsupported_file of Path.Build.t * File_kind.t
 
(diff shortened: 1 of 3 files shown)


rleshchinskiy:
@Alizter Could you please advise whether the Coq test failure is an issue.

ejgallego:
> @Alizter Could you please advise whether the Coq test failure is an issue.

Yes, it is a real failure; the test can be fixed by adding a dummy `.v` file, for example one that contains `Definition a := 3.`.

This is actually interesting, as we already have a weird path in the coqdoc rules when `modules = []`, so we will make this case an error directly in the rules.

rleshchinskiy:
Thanks, @ejgallego. From what you say, I assume this restriction won't be a problem for Coq in practice?

ejgallego:
> Thanks, @ejgallego. From what you say, I assume this restriction won't be a problem for Coq in practice?

I think it won't indeed, I am not sure people have come up with use cases for theories without any attached module, and even so, we could always generate a dummy `index.html` in that case.

rgrinberg:
Out of curiosity, what makes it hard to support empty directory targets?

rleshchinskiy:
> Out of curiosity, what makes it hard to support empty directory targets?

The main user-facing issue is that if you promote them, your source control will (likely) ignore them. We decided that they are worth flagging explicitly for that reason.

rgrinberg:
Given that justification, I would imagine it makes sense to error on such directories only if `mode = Promote`. Although that might be a little ad-hoc.

rleshchinskiy:
I'll defer to @snowleopard on this. Internally, empty directories in `Targets.Produced.t` were wobbly before and are no longer allow with the (merged and upcoming) caching changes so actually supporting them properly would need more work.

rgrinberg (approved):

snowleopard:
> Given that justification, I would imagine it makes sense to error on such directories only if `mode = Promote`. Although that might be a little ad-hoc.

@rgrinberg Promotion shouldn't turn build successes into failures, I think.

PR #9472, opened 2023-12-12
Added check on package names in dependencies

ElectreAAS:
This PR is roughly the same as #9385 which was closed by mistake. Sorry for the inconvenience.
The last commit here wasn't in the original, and solves my question in https://github.com/ocaml/dune/pull/9385#issuecomment-1842561593.

cc @Leonidas-from-XIV


ElectreAAS:
> https://github.com/ocaml/dune/pull/9385#issuecomment-1846922522
> However should we care about `(depends)` if we're not generating opam files and never use the data?
> 
> Not sure how many packages specify depends without generating opam files, because without it it is fairly pointless at the moment.

Regardless of whether or not we use the data, the documentation states that package names have to be valid wrt. a specific syntax, so we probably should check it.

Leonidas-from-XIV:
I feel like this is good but what would be even better is (as hinted in one of the cram tests) is detect the `<valid-package-name>.<version>` case specifically and give a hint that is not `<valid-package-name>_<version-with-underscores>` as a hint but rather `(<valid-package-name> (= "<version>"))` which is what the user probably meant here and fall back to the current hint if the the package name is nonsense.

test/blackbox-tests/test-cases/invalid-package-version.t
```diff
@@ -19,7 +19,7 @@ Building doesn't work as the dependency is invalid.
   Error: "foo.1.2.3" is an invalid opam package name.
   Package names can contain letters, numbers, '-', '_' and '+', and need to
   contain at least a letter.
-  Hint: foo_1_2_3 would be a correct opam package name
+  Hint: (foo (= 1.2.3)) would be a correct opam package name
```
Alizter, line 22: This is no longer an "opam package name" but a "package dependency".

Alizter: Any other time we validate a package name would also give a hint like `(foo (= 1.2.3))` even if that syntax is not accepted.

Leonidas-from-XIV: This is a good point, it should be constrained to places where we can potentially generate opam files from.

src/dune_lang/package_name.ml
```diff
@@ -49,11 +49,24 @@ module Opam_compatible = struct
...
     let of_string_opt s = Option.some_if (is_valid_string s) s
 
-    let make_valid s =
+    let escape s =
       let replaced = String.map s ~f:(fun c -> if is_valid_char c then c else '_') in
       if is_valid_string replaced then replaced else "p" ^ replaced
     ;;
 
+    let make_valid s =
+      match index_opt s '.' with
```
Leonidas-from-XIV, line 58: Better use `String.lsplit2 ~on:'.' s` here, as this will get you the `before`/`after` part immediately.

src/dune_lang/package_name.ml
```diff
@@ -49,11 +49,24 @@ module Opam_compatible = struct
...
+      | None -> escape s
+      | Some i ->
+        let before = sub s ~pos:0 ~len:i in
+        let s_len = length s in
+        if is_valid_string before && s_len > i + 1
+        then (
+          let after = sub s ~pos:(i + 1) ~len:(s_len - i - 1) in
+          Printf.sprintf "(%s (= %s))" before after)
```
Leonidas-from-XIV, line 66: Just a stylistic convention, we don't usually use `Printf.sprintf` in dune, it is already included so `sprintf` should suffice.

src/dune_lang/package_name.ml
```diff
@@ -49,11 +49,24 @@ module Opam_compatible = struct
 
     let of_string_opt s = Option.some_if (is_valid_string s) s
 
-    let make_valid s =
+    let escape s =
       let replaced = String.map s ~f:(fun c -> if is_valid_char c then c else '_') in
       if is_valid_string replaced then replaced else "p" ^ replaced
     ;;
 
+    let make_valid s =
```
Leonidas-from-XIV, line 57: As @Alizter suggests, I think this needs a separate `make_valid_for_dependency` (or whatever you want to call it) function.

Leonidas-from-XIV (changes requested):

src/dune_lang/package_name.ml
ElectreAAS: Didn't know this one, as it's in stdune but not in the stdlib :)

src/dune_lang/package_name.ml
ElectreAAS: I believe using a simpler `Dependency.decode` which uses the old `make_valid` under the hood, as per my last commit, solves this. The old way of doing it is preserved.

test/blackbox-tests/test-cases/invalid-package-version.t
ElectreAAS: I believe my last commit solves all of these comments.

Issue #9474, opened 2023-12-12
pkg: config reading not supported on Windows


Alizter:
When building packages on Windows, we get the following error:

```
PS C:\Users\ali\test> C:\Users\ali\dune\_build\install\default\bin\dune.exe build
Internal error, please report upstream including the contents of _build/log.
Description:
  ("unable to serialize exception",
  { exn = "Failure(\"Windows only. This function isn't implemented.\")" })
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml",
  line 31, characters 17-27
Re-raised at CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml",
  line 36, characters 4-11
Called from OpamFilename.prettify_path in file
  "vendor/opam/src/core/opamFilename.ml", line 497, characters 33-54
Called from OpamFile.MakeIO.read_opt in file
  "vendor/opam/src/format/opamFile.ml", line 125, characters 19-42
Called from OpamFile.MakeIO.read in file
  "vendor/opam/src/format/opamFile.ml", line 149, characters 10-20
Called from Dune_rules__Pkg_rules.Install_action.Spec.read_variables in file
  "src/dune_rules/pkg_rules.ml", line 1258, characters 16-45
Called from Dune_rules__Async.async in file "src/dune_rules/async.ml", line
  6, characters 30-36
Called from Dune_rules__Pkg_rules.Install_action.Spec.action in file
  "src/dune_rules/pkg_rules.ml", line 1384, characters 25-75
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
Re-raised at Stdune__Exn.raise_with_backtrace in file
  "otherlibs/stdune/src/exn.ml", line 38, characters 27-56
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
Re-raised at Stdune__Exn.raise_with_backtrace in file
  "otherlibs/stdune/src/exn.ml", line 38, characters 27-56
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
-> required by
   ("execute-rule",
   { id = 0
   ; info =
       From_dune_file
         { pos_fname = "<none>"
[...]

PR #9486, opened 2023-12-13
fix(pkg): add stub for shGetFolderPath

Alizter:
Building packages on windows results in error described here:
- https://github.com/ocaml/dune/issues/9474

This is due to a missing stub. This PR adds that missing stub.

- [ ] update fork

@dra27 Could you take a look at this? I've plucked your code from the opam stubs, but there were some warnings about `WCHAR` being incompatible with the expected type for `SHGetFolderPath`. Turns out there is a variant called `SHGetFolderPathW` specifically for "W"ide characters which seems to work.

When this patch is used, the build of a lock file gets further however invocation of `ocamlc` seems to get confused about Windows paths in `BUILD_PATH_PREFIX_MAP`. Would you know if this is a known problem in OCaml?


vendor/opam/src/core/custom_opamStubs.c
```diff
@@ -78,6 +79,26 @@ CAMLprim value OPAMW_GetArchitecture(value unit)
     return get_PE_cpu_architecture(NativeMachine);
 }
 
+/*
+ * Somewhat against my better judgement, wrap SHGetFolderPath rather than
+ * SHGetKnownFolderPath to maintain XP compatibility. OPAM already requires
```
Leonidas-from-XIV, line 84: But then, wouldn't we be perfectly fine to use `SHGetKnownFolderPath` in our stubs? We don't really have to stay compatible with an OS from 2001.

I don't even think OCaml 5+ has a native compiler that supports 32-bit systems like XP machines would predominantly be.

Alizter: I didn't write this comment, but took it from the opam stubs.

Leonidas-from-XIV: Yes, I know, but the point of the comment is a valid one and while for OPAM it might've been the right call (I assume), it might not be for us.

rgrinberg:
Before there's any discussion of the code, can you update the description to include:

1. What's the problem that you're address
2. How this PR is addressing this problem

Alizter:
@rgrinberg I've updated the PR description with the issue that this is meant to fix: https://github.com/ocaml/dune/issues/9474

rgrinberg:
Either I'm missing some context, or the description is still incomplete. How is this stub supposed to help with reading a config file? Why would we even need this platform specific code to read a config file?

Alizter:
@rgrinberg I'm not entirely sure why opam wants to read the home directory in this case. This PR simply stops that from being an error which it currently is. You can check the backtrace in that issue to see that this function eventually gets called and because it doesn't have an implementation it errors.

We could probably do this in a better way, but it was my understanding that we didn't want to modify the opam code for now and get something working.

rgrinberg:
Yes, we don't want to modify the opam code. Unfortunately, this is exactly what your PR is doing.

Looking at the stacktrace, here's an alternative that does not modify upstream:

1. Read the config file using our own IO functions
2. Parse it using one of the string parsers.

Alizter:
I somewhat disagree that this is what my PR is doing, but I agree that your alternative is perhaps a better option.

PR #9488, opened 2023-12-13
Use Targets.Produced.t in dune_cache

rleshchinskiy:
We replace various occurrences of `('a * Path.Build.t) list` with `'a Targets.Produced.t`.  This allows us to maintain various invariants throughout and should also make handling directory targets here a lot more straightforward.

This requires moving `Dune_engine.Targets` into a separate library.


snowleopard (approved):
Thanks! Reviewed and benchmarked internally.

Issue #7908, opened 2023-06-06
[coq] Composition with (boot) theories hangs with a cycle in Dune > 3.7 when coq lang > 0.6


ejgallego:
I've discovered some issues when testing composition with Coq itself in Dune 3.8.

First issue is as follows, given:
- a composed Coq build (see [1] how to setup)
- a `(lang coq 0.7`) or 0.8

Then Dune will fail with:
```
Error: Dependency cycle between:  
   Computing installable artifacts for package coqide-server
-> required by Computing installable artifacts for package coq-core
-> required by _build/default/coq-core.install
-> required by alias default in dune:22
```
This seems to be related to the `Coq_config` resolution in boot. Note that the issue doesn't occur for lower versions of the Coq language.

cc @Alizter 

[1] To reproduce, you can use Coq's main branch:
```
$ git clone https://github.com/coq/coq.git
$ cd coq && cp theories/dune.disabled theories/dune
$ # edit dune-project to raise `(lang coq ...)`
$ dune build @default
```

ejgallego:
@rgrinberg it seems to the that the problem here is as follows:
- we have two packages `coq-core` and `coq-stdlib`, `coqc` is in `coq-core`, `coq-stdlib` has all the .v files
- to compute the list of files to install for `coq-stdlib`, Dune needs to call `coqc --config` to see if Coq has native compiler enabled
- the resolution of `coqc` in this case seems to depend on all the .install files in scope.

And boom! Here we got a cycle. Is there any hope to avoid this?

It would also be fine for us to use a weaker resolution method.

rgrinberg:
Can you post the trace of the cycle?

ejgallego:
> I suppose we can change the computation of the artifacts db to do binaries and libraries separately.

Happy to follow your lead on this.

For a short-term fix, do we have a method that will resolve the binary more weakly ?

Alizter:
Well there is `Which` but that won't be able to find it from a super_context.

ejgallego:
If that doesn't trigger the cycle we could live with it, no?

The only downside is that the Coq composed build wouldn't detect native or version, but that's not a big deal I think.

ejgallego:
I am surprised tho the cycle doesn't trigger even in a non-composed scenario wil multiple packages tho.

We may likely have a problem there too, but maybe people haven't bumped the `(lang coq )` variable yet so it is not bitting them.

rgrinberg:
[...]

PR #9496, opened 2023-12-13
fix: resolve installed binaries to local paths

rgrinberg:
Previously, we'd always binaries from the install context
(_build/install/$context/bin). This would unnecessarily load the install
rules.

Now, we add an argument that allows us to resolve to the original paths.
Reducing the amount of rules that need to be loaded.

We use this argument in a few cases where we don't need to build the
path in _build/install

Ideally, we'd use this everywhere as it's basically strictly superior. The only issue is that it doesn't preserve backward compat in cases like:

```
(rule
 (dep %{bin:foo})
 (action (run which foo)))
```

Resolving it in _build/install/default will make the `which` above succeed because we add the bin path in the install directory to the PATH when executing actions. There's a way to get around that of course, but it's not trivial.

### DIFF
--- a/doc/changes/9496.md
+++ b/doc/changes/9496.md
@@ -1,0 +1,3 @@
+- Resolve various public binaries to their build location, rather than to where
+  they're copied in the `_build/install` directory (#9496, fixes #7908,
+  @rgrinberg).
--- a/src/dune_rules/melange/melange_binary.ml
+++ b/src/dune_rules/melange/melange_binary.ml
@@ -1,7 +1,13 @@
 open Import
 
 let melc sctx ~loc ~dir =
-  Super_context.resolve_program_memo sctx ~loc ~dir ~hint:"opam install melange" "melc"
+  Super_context.resolve_program_memo
+    sctx
+    ~loc
+    ~dir
+    ~where:Original_path
+    ~hint:"opam install melange"
+    "melc"
 ;;
 
 let where =
--- a/otherlibs/stdune/src/map.ml
+++ b/otherlibs/stdune/src/map.ml
@@ -44,6 +44,12 @@ module Make (Key : Key) : S with type key = Key.t = struct
   let merge a b ~f = merge a b ~f
   let union a b ~f = union a b ~f
 
+  let union_all maps ~f =
+    match maps with
+    | [] -> empty
+    | init :: maps -> List.fold_left maps ~init ~f:(fun acc map -> union acc map ~f)
+  ;;
+
   let union_exn a b =
     union a b ~f:(fun key _ _ ->
       Code_error.raise
--- a/otherlibs/stdune/src/map_intf.ml
+++ b/otherlibs/stdune/src/map_intf.ml
@@ -20,6 +20,7 @@ module type S = sig
   val add_multi : 'a list t -> key -> 'a -> 'a list t
   val merge : 'a t -> 'b t -> f:(key -> 'a option -> 'b option -> 'c option) -> 'c t
   val union : 'a t -> 'a t -> f:(key -> 'a -> 'a -> 'a option) -> 'a t
+  val union_all : 'a t list -> f:(key -> 'a -> 'a -> 'a option) -> 'a t
 
   (** Like [union] but raises a code error if a key appears in both maps. *)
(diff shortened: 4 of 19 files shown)


src/dune_rules/melange/melange_binary.ml
```diff
@@ -1,7 +1,13 @@
...
 
 let melc sctx ~loc ~dir =
-  Super_context.resolve_program_memo sctx ~loc ~dir ~hint:"opam install melange" "melc"
+  Super_context.resolve_program_memo
+    sctx
+    ~loc
+    ~dir
+    ~where:Original_path
+    ~hint:"opam install melange"
+    "melc"
```
rgrinberg, line 10: @anmonteiro I tweaked the resolution of melange here as well. It shouldn't affect anything, but could in theory help with some odd cycles

rgrinberg:
I just realized that we also resolve `(enabled_if ..)` a little too early for public binaries. I'll fix that in a subsequent PR.

rgrinberg:
ping @ejgallego

ejgallego:
> ping @ejgallego

Thanks, I'm still on holidays but I'll try to test this ASAP.

rgrinberg:
Sure no problem. Although the window for getting this in for 3.13 is closing.

ejgallego:
> Sure no problem. Although the window for getting this in for 3.13 is closing.

I just came back to Paris today, I'll keep that in mind.

I have a first question tho, is this really a refactoring? Seems to me like an observable behavior change?

rgrinberg:
I did a CHANGES entry to the observable bit. I'll change the title since it's indeed not a refactoring.

ejgallego:
@rgrinberg I can confirm that this PR fixes the base cases I have for #7908 .

However we still have a problem in the case a Coq theory using plugins is in the workspace. In this case, Dune hangs hard.

I guess this is due to `Coq_rules.ml_pack_and_meta_rule` , however I don't see how the install rules are forced.

What the code does is as follows, assume a setup such as:

```lisp
(library
 (name bar)
 (public_name foo.bar)
 (libraries coq-core.vernac))

(coq.theory
 (name bar)
 (package foo)
 (modules moo)
 (plugins foo.bar))
```

Then, the rule to generate `bar.v.d` using coqdep needs to be as follows:
```
(rule
  (targets moo.v.d)
  (deps %{install}/lib/foo/META moo.v)
  (action (run coqdep -Q bar . moo.v)))
```

As `coqdep` needs `META` for foo in place to resolve the `.cmxs` file that will appear as a dependency on the `.vo` rule.

But indeed I'm puzzled as why this makes Dune hang, I don't see any forcing in computing the `META` path.

rgrinberg:
Those META issues need to be addressed separately from this PR.

> But indeed I'm puzzled as why this makes Dune hang, I don't see any forcing in computing the META path.

Probably there's a cycle that makes it hang. AFAIK @snowleopard is working on upstreaming some improvements to the cycle detection

ejgallego:
> Probably there's a cycle that makes it hang. AFAIK @snowleopard is working on upstreaming some improvements to the cycle detection

In fact I think what hangs is the printing of the cycle error, dune gets stuck like this:
```
$ dune build -p coq-core,coq-stdlib,foo
   Computing installable artifacts for package coq-core
[hangs]
```
Note that it gets stuck with 0% CPU use, so it is a deadlock.

I've pushed a tree [here](https://github.com/ejgallego/coq/tree/dune_hang_bug) if you folks are curious, to reproduce see commit message.

ejgallego:
Actually never mind, I messed up my testing and I was not using the right `dune` binary!

Things seem to work fine even with my extended test cases, as they should be.

doc/changes/9496.md
```diff
@@ -0,0 +1,2 @@
+- Resolve various public binaries to their build location, rather than to where
+  they're copied in the `_build/install` directory (@9496, @rgrinberg).
```
ejgallego, line 2: Fixes #7908

ejgallego:
Another comment @rgrinberg : should this change be guarded by dune lang version?

ejgallego:
LGTM

PR #9505, opened 2023-12-14
Switch to Filename.Set in various places

rleshchinskiy:
This switches various things to use `Filename.Set` and `Filename_set` instead of `Path.Set`.

Most of the code is by @snowleopard


snowleopard (approved):
Thanks for upstreaming these changes!

rgrinberg (approved):

rgrinberg:
> @rgrinberg Could you double-check changes in the rules?

LGTM

PR #9508, opened 2023-12-14
fix: stop passing -j jobs in opam file to ocaml

Alizter:
This seems to be ignored, but it definitely isn't correct.


rgrinberg:
It's not ignored.

Alizter:
Sorry I misunderstood.

PR #9509, opened 2023-12-15
pkg: version field in lockfile is mandatory

gridbugs:


test/blackbox-tests/test-cases/pkg/slang.t
```diff
@@ -8,7 +8,10 @@
 Helper function to create a lockfile with a given install action and then build it, running the action.
   $ test_action() {
   >   dune clean || true
-  >   echo "(install $1)" > dune.lock/test.pkg
+  >   cat > dune.lock/test.pkg <<EOF
+  > (version 0.0.1)
+  > (install $1)
```
Leonidas-from-XIV, line 13: Could be swapped to avoid the renumbering in the output, but this is just nit-picking.

Leonidas-from-XIV (approved):

rgrinberg (approved):

test/blackbox-tests/test-cases/pkg/slang.t
gridbugs: I did that at first but changed it to have the version first because I wanted the lockfiles in the tests to be more realistic.

PR #9510, opened 2023-12-15
pkg: allow default local package version to be set

gridbugs:


test/blackbox-tests/test-cases/pkg/helpers.sh
```diff
@@ -80,8 +80,14 @@ solve_project() {
...
 
 make_lockdir() {
-  mkdir dune.lock
-  cat >dune.lock/lock.dune <<EOF
+  if [ "$#" -eq "0" ]
+  then
+    lockdir="dune.lock"
+  else
+    lockdir="$1"
+  fi
```
Leonidas-from-XIV, line 88: Not sure if this is POSIX-compliant but you could probably use `${1:-dune.lock}` here.

Leonidas-from-XIV (commented):
The code looks good and straightforward, however I would like to know what the use case for this is? Every feature comes with a cost (in this case, the API requiring more arguments) and I wonder when would be a use case where a user would want to set their version of local packages to something else.

rgrinberg:
> The code looks good and straightforward, however I would like to know what the use case for this is? Every feature comes with a cost (in this case, the API requiring more arguments) and I wonder when would be a use case where a user would want to set their version of local packages to something else.

The simplest way to motivate this feature is that it's needed for full control of the solver. Packages in opam repository should be able to depend on local packages in the workspace, and therfore are able to express constraints based on the versions of local packages. So we need control over these version numbers to guide the solver into picking the solutions we want. It's the same as opam allowing you to select the version when pinning a package.

rgrinberg:
The field that this PR introduces is rather weird IMO. There should be a way to just set the version per package.

gridbugs:
Closing this as I don't think it's needed. I'll make a new pr for overriding the versions of individual packages.

PR #9512, opened 2023-12-15
Add (explain) field to (menhir) stanza

### LINKED ISSUE #6865: Using `menhir --explain`

nojb:
A simple-minded attempt at resolving #6865: add an `(explain)` field to the `(menhir)` stanza.

Fixes #6865


gasche:
Nice!

It would be nice to also update the documentation.

Would there be a way for users to access the `.conflicts` file that does not involve foraging into the build artifact directory? (For example, is it the case that `dune exec -- less foo/Parser.conflicts` would work?)

nojb:
> It would be nice to also update the documentation.

Yes, will do.

> For example, is it the case that `dune exec -- less foo/Parser.conflicts` would work?

This seems to work.

> Would there be a way for users to access the `.conflicts` file that does not involve foraging into the build artifact directory?

One can also add `(mode (promote (until-clean)))` to "promote" the `.conflicts` file to the current directory, but this will also promote the generated `.ml`/`.mli` files (Dune does not appear to be able to promote only a subset of the generated files by a rule at the moment).

nojb:
I pushed another commit https://github.com/ocaml/dune/pull/9512/commits/0b46891d03ab6b3f23d79c81b4245daffeb90943 detecting the `--explain` flag in the `(flags)` field. I'm in two minds whether this one is better than `(explain)`.

gasche:
I don't like the idea of seeing the generated .ml and .mli in my build directory. For the time being I would rather document `dune exec -- less`.

gasche:
I wonder whether `explain` should be the default. This sounds like a very convenient thing to have, and I don't see any immediate cost -- presumably this has a performance cost, but this is neglectible for not-too-large grammars.

cc @fpottier: do you see an obvious issue with Dune passing `--explain` by default to Menhir? (Are there realistic use-cases where this is problematic?)

Does your current implementation support `(explain false)` to disable the option (should be a no-op for now)? Having this working today would be convenient if `(explain)` becomes the default now or in the future.

ejgallego:
Explain being the default would make sense, note however the case where the user does something like `(flags :standard \ --explain)` should be either documented as not supported, or supported.

For the Coq mode, understanding flags would have been very convenient, but there were too many corner cases in the end.

rgrinberg:
> Dune does not appear to be able to promote only a subset of the generated files by a rule at the moment

There's actually `(promote (only ..))`. Although I don't know whether it's good to use this in the end. Also, I guess it can't be combined with `until-clean`?

> Would there be a way for users to access the .conflicts file that does not involve foraging into the build artifact directory? (For example, is it the case that dune exec -- less foo/Parser.conflicts would work?)

We can also add a specialized dune subcommand for menhir. I don't know if using `dune exec` is all that obvious.

A couple of improvements I would like to suggest:

1. We should add a warning or error if the user passes `--explain` manually and point them to this feature (only when the newer version is enabled)
2. We should allow blang expressions inside `(explain ..)` to allow conditionally enabling this feature. It can always be added later, but it creates complications due to versioning.

src/dune_rules/menhir/menhir_stanza.ml
```diff
@@ -31,14 +33,21 @@ let decode =
...
+     and+ explain = field_o "explain" (Dune_lang.Syntax.since syntax (2, 2) >>> bool) in
      let infer =
        match infer with
        | Some infer -> infer
        | None -> menhir_syntax >= (2, 0)
      in
-     { merge_into; flags; modules; mode; loc; infer; enabled_if })
+     let explain =
+       match explain with
+       | None -> dune_syntax >= (3, 13)
```
rgrinberg, line 47: I think you should use the version of the menhir syntax to control this behavior.

nojb: Good point, done.

rgrinberg (approved):

nojb:
> 2\. We should allow blang expressions inside `(explain ..)` to allow conditionally enabling this feature. It can always be added later, but it creates complications due to versioning.

This should not be working.

nojb:
> 1. We should add a warning or error if the user passes `--explain` manually and point them to this feature (only when the newer version is enabled)

I am not sure this is going to be very useful if the default

> 1. We should add a warning or error if the user passes `--explain` manually and point them to this feature (only when the newer version is enabled)

I pushed something along these lines, let me know if it looks OK.

nojb:
@rgrinberg just to confirm before this gets merged: do you think it is OK to switch the default to __generate__ the `.conflicts` file?

rgrinberg:
If the change is guarded behind a version flag, I see no issues from the rules side. I don't know enough about menhir to know whether this is a good idea though. I'll trust Gabriel's opinion on this one.

nojb:
I pushed one more commit that:

- Adds a new `(menhir (flags ...) (explain ...))` field to the `(env ...)` field allowing to set the new `(explain)` setting using `(env)` as well.
- Removes the old `(menhir_flags ...)` field of `(env)` (as it is replaced by `(menhir (flags ...))`).

The CRAM test shows these features in action.

Alizter:
@nojb could you test that this works with a dune-workspace file?

rgrinberg:
There's no need for such a test, there's already one that demonstrates the menhir flags cannot be set in the workspace file.

src/dune_lang/blang.ml
```diff
@@ -15,6 +15,20 @@ module Ast = struct
   let true_ = Const true
   let false_ = Const false
 
+  let rec equal f t1 t2 =
+    match t1, t2 with
+    | Const b1, Const b2 -> Bool.equal b1 b2
+    | Not t1, Not t2 -> equal f t1 t2
+    | Expr x1, Expr x2 -> f x1 x2
+    | And tl1, And tl2 | Or tl1, Or tl2 ->
+      (match List.for_all2 ~f:(equal f) tl1 tl2 with
```
rgrinberg, line 24: Isn't `List.equal` good here?

rgrinberg:
> I pushed one more commit that:
> 
> * Adds a new `(menhir (flags ...) (explain ...))` field to the `(env)` stanza allowing to set the new `(explain)` setting using `(env)` as well.
> * Removes the old `(menhir_flags ...)` field of `(env)` (as it is replaced by `(menhir (flags ...))`).
> 
> The CRAM test shows these features in action.

The addition looks good. Given that the new version of the menhir syntax isn't strictly backward compatible, how about bumping it to `(3, 0)`?

src/dune_lang/blang.ml
nojb: Yes, it is. Amended as suggested, thanks.

nojb:
> The addition looks good. Given that the new version of the menhir syntax isn't strictly backward compatible, how about bumping it to `(3, 0)`?

Done, thanks.

nojb:
(This is ready from my side.)

rgrinberg:
This is ready from my end as well.

rgrinberg:
Will take the liberty to merge this one to avoid a merge conflict with some of my own work. Thanks @nojb!

nojb:
Thanks!

gasche:
Thanks as well!

PR #9515, opened 2023-12-16
refactor: introduce inherited fields in stanza db

rgrinberg:
This allows us to define the constituents of `Env_node` separately. The goals are as follows:

1. Make it possible for rules to define their own handling of their `env` fields individually. This allows for better isolation of subsystem specific parts to their own part of the rules. E.g. the coq stuff now lives separate from everything else.

2. Remove the manual shuffling of scope/artifacts that's required when creating the env node. Getting rid of this also fixes a bug with the vendored flags as we can see from the test. Also this PR doesn't completely get rid of this.

3. Improve the performance of incremental compilation. Now, we don't recompute all the effective values derived from `env` when a single one changes. The recomputation is reserved to only what needs to be recomputed. Although no cutoffs have been added yet.

### DIFF
--- a/test/blackbox-tests/test-cases/github7034.t/run.t
+++ b/test/blackbox-tests/test-cases/github7034.t/run.t
@@ -106,23 +106,16 @@ But when lang dune is 3.3 or higher the warning becomes an error:
     -short-paths
     -keep-locs))
   Leaving directory 'outer'
+
   $ dune build --root=outer
   Entering directory 'outer'
-  File "vendored/inner/inner.ml", line 6, characters 11-18:
-  6 | type t = { x : int }
-                 ^^^^^^^
-  Error (warning 69 [unused-field]): unused record field x.
   Leaving directory 'outer'
-  [1]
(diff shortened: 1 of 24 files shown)


test/blackbox-tests/test-cases/github7034.t/run.t
```diff
@@ -106,23 +106,16 @@ But when lang dune is 3.3 or higher the warning becomes an error:
     -short-paths
     -keep-locs))
   Leaving directory 'outer'
+
   $ dune build --root=outer
   Entering directory 'outer'
-  File "vendored/inner/inner.ml", line 6, characters 11-18:
```
rgrinberg, line 111: @gridbugs could you take a look at this test? Does this PR fix the issue?

gridbugs: yep!

rgrinberg:
@gridbugs do you want to look over this one?

gridbugs (approved):

PR #9523, opened 2023-12-18
Store Targets.Validated and Produced as root dir + basenames

rleshchinskiy:
These represent targets for one rule and are thus rooted in one directory. This PR changes the representation to reflect this fact and uses this to refactor a few places. The resulting code seems cleaner, is slightly more efficient and simplifies the implementation of caching for directory targets in a subsequent PR.


snowleopard (approved):
Thanks! All reviewed, benchmarked and tested internally.

PR #9528, opened 2023-12-19
Add test for parse error when using run action

gridbugs:
Repro for https://github.com/ocaml/dune/issues/9529


rgrinberg:
I think you need to promote this test and add the appropriate incantation to remove the stack trace (search for `awk` in the repo)

gridbugs:
I'm planning to wait until https://github.com/ocaml/dune/pull/9530 is merged at which point the test will no longer fail.

rgrinberg (approved):
Usually, the other order is preferred. First demonstrate the failure with a failing test, and then the subsequent commit demonstrates how the issue is fixed in the test.

PR #9531, opened 2023-12-19
fix(pkg): prevent blang/slang infinite recursion

gridbugs:
Each of the parsers for the blang and slang DSLs attempt to invoke the other when encountering a form they can't interpret. Blang invokes slang as a slang expression may return a blang literal (a string "true" or "false"), and slang invokes blang to convert the result of a blang expression into a string "true" or "false". When a form couldn't be interpreted by either parser each parser would continuously invoke the other leading to infinite recursion. This fixes the problem by checking the parser's location in the file and making sure that progress has been made before recurring.

### DIFF
--- a/src/dune_lang/slang.ml
+++ b/src/dune_lang/slang.ml
@@ -25,9 +25,16 @@ and form =
   | Or_absorb_undefined_var of blang list
   | Blang of blang
 
+let decode_literal =
+  let open Decoder in
+  let+ x = String_with_vars.decode in
+  Literal x
+;;
+
 let decode =
   let open Decoder in
   fix (fun decode ->
+    let decode_blang = Blang.Ast.decode ~override_decode_bare_literal:None decode in
     let decode_form =
       sum
         ~force_parens:true
@@ -35,11 +42,11 @@ let decode =
           , let+ x = repeat decode in
             Concat x )
         ; ( "when"
-          , let+ condition = Blang.Ast.decode decode
+          , let+ condition = decode_blang
             and+ t = decode in
             When (condition, t) )
         ; ( "if"
-          , let+ condition = Blang.Ast.decode decode
+          , let+ condition = decode_blang
             and+ then_ = decode
             and+ else_ = decode in
             If { condition; then_; else_ } )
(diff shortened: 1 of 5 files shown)


src/dune_lang/slang.ml
```diff
@@ -58,12 +58,37 @@ let decode =
...
+               attempting to parse the blang expression. If the blang decoder
+               invokes the slang decoder at the same location we know that it's
+               not a valid blang expression. Since we only invoke the blang
+               decoder when the slang decoder has already failed to interpret
+               the form, the form is invalid so raise the original exception
+               from when the slang decoder failed to interpret it. *)
+            let* string_start = loc in
+            if Loc.equal string_start blang_start then raise slang_exn else decode
+          in
+          let+ loc, x = located (Blang.Ast.decode decode_blang_string) in
```
Leonidas-from-XIV, line 88: It feels weird to me to solve this with a comparison of Loc. Maybe this could be solved by a parameter `recurse_blang`/`recurse_slang` to detect a Blang->Slang->Blang or Slang->Blang->Slang case?

gridbugs: Good point. I've added a `Blang.Ast.decode_without_bare_literals` decoder for blang which doesn't parse boolean literals to break cycles like this without resorting to comparing locations.

Leonidas-from-XIV: That seems like a much less "magical" way, much better! However it seems it triggers new test failures.

rgrinberg (approved):

rgrinberg (changes requested):
Seems like this fix breaks parsing elsewhere (see CI)

gridbugs:
Ah yep, without bare literals at all we can't parse literal bools in blangs that are being implicitly converted to slangs, such as `(run echo (and true))`. Updated it so that the the `'string` parser used to decode blang literals can be overiden, and the slang parser overrides it to only parse slang literals when parsing blangs in the position where a slang is expected (ie. the implicit blang->slang conversion).

rgrinberg (approved):

Issue #9522, opened 2023-12-18
Add `cond` statement to slang dsl


gridbugs:
The immediate motivation is allowing the lockdir to be chosen dynamically by a slang expression in a more ergonomic way than nested if-statements (see https://github.com/ocaml/dune/issues/9451#issuecomment-1851363963).

The proposed syntax is:
```
(cond
 (<condition1> <expr1>)
 (<condition2> <expr2>)
 ...
 (<conditionN> <exprN)
 <exprDefault>)
```

gridbugs:
We didn't end up needing this feature for package management.

Issue #9451, opened 2023-12-11
Choose an appropriate lockdir for the current system


gridbugs:
If a project has different dependencies on different systems it will require separate lockdir for each system. For the purpose of this discussion assume that a project can be built on both linux and windows, but has different dependencies on each system which are locked in directories "dune.linux.lock" and "dune.windows.lock" respectively.

Currently each build context is associated with a single lockdir and each dune project has a single default build context. If the default build context uses its lockdir as "dune.linux.lock" then windows users will need to manually specify an alternative build context which uses "dune.windows.lock". More generally, each dune project with system-specific dependencies will need to nominate a particular system to be favoured by the default context and require developers on other systems to manually override the context (via command line args) to the one compatible with their system. This will likely lead to frustration as it means the default behaviour of dune (without choosing an alternative context) doesn't work for some users, depending on their system.

To address this dune will need to choose which lockdir to use by interrogating the current system. To motivate discussion, here's my initial proposal for how it should look:

```
; dune-workspace
...
(context
  (default
    (name default)
    (lockdir_choice
     (dune.linux.lock (= %{os} linux)
     (dune.windows.lock (= %{os} windows)))
```

I'm proposing to add a new `lockdir_choice` field to `context`, mutually-exclusive with the `lockdir` field, which defines a list of (path, blang) pairs encoding the conditions under which each lockdir can be chosen. I'm against using a more general select statement or the slang language here so that in the event that the user's system isn't in the list, dune can print an appropriate error message. The reason the predicates are expressed as blangs rather than just naming the system is so that users can discriminate on other aspects of the system such as arch and version. The reason I propose dynamically choosing the lockdir of a context rather than dynamically choosing the context itself is because the other fields of the context don't depend on the current system.

nojb:
Sorry for naïve question, but what is the obstacle to Dune taking care of a per-system dependencies using a single metadata directory/file and doing so automatically (without needing intervention from the developers or the users of the project)?

rgrinberg:
The suggestion is fine, but I dislike adding such specific conditional constructs to the dune language. I'd prefer a more general construct, even if it's only available in a single location at the start. If we had many constructs like this, I fear that the dune language will become impossible to learn or remember.

> but what is the obstacle to Dune taking care of a per-system dependencies using a single metadata directory/file and doing so automatically (without needing intervention from the developers or the users of the project)?

[...]

PR #9533, opened 2023-12-19
pkg: add cond expression to slang dsl

gridbugs:
Fixes https://github.com/ocaml/dune/issues/9522

### DIFF
--- a/src/dune_lang/action.ml
+++ b/src/dune_lang/action.ml
@@ -533,6 +533,14 @@ let rec slang_map_string_with_vars ~f = function
         When
           ( blang_map_string_with_vars condition ~f:(slang_map_string_with_vars ~f)
           , slang_map_string_with_vars t ~f )
+      | Cond { cases; default } ->
+        Cond
+          { cases =
+              List.map cases ~f:(fun (condition, slang) ->
+                ( blang_map_string_with_vars condition ~f:(slang_map_string_with_vars ~f)
+                , slang_map_string_with_vars slang ~f ))
+          ; default = slang_map_string_with_vars default ~f
+          }
       | If { condition; then_; else_ } ->
         If
           { condition =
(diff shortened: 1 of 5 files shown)


gridbugs:
Heads up that the tests will fail until https://github.com/ocaml/dune/pull/9531 is merged. I've tested this locally cherry-picking the fix.

gridbugs:
More context for the motivation for this change is in the discussion here: https://github.com/ocaml/dune/issues/9451

rgrinberg:
Why are we adding this construct to slang though? IIUC, slang is our WIP for translating opam build commands. While we need `cond` to evaluate conditionals in workspace files.

gridbugs:
> Why are we adding this construct to slang though? IIUC, slang is our WIP for translating opam build commands.

That's just one application of slang. The intention is that slang is a general purpose string maniplation dsl for dune. Slang already has the ability to resolve conditionals and work with strings which as all we need to do to choose a lockdir by inspecting the current system.

rgrinberg:
There's some features of slang that I wouldn't want the rest of the dune language to inherit. For example, the handling of undefined variables. In any case, I'd like to see how slang is actually integrated into the evaluation of dune files first.

gridbugs:
Ah fair enough. What would you suggest to address https://github.com/ocaml/dune/issues/9451 then? Would it make sense to have a cond-like statement that's specific to the `lockdir` field? I guess we could add cond to the `lockdir` field first and then make it available in other parts of dune as appropriate over time?

gridbugs:
More generally speaking, I think there's room for a general string manipulation language like slang in other parts of dune but I agree with your concerns about the handling of undefined variables. Maybe we should separate out the core of the language from parts that are only there to handle opam's idiosyncrasies and then use that core language for the `lockdir` field. That way users wouldn't need to learn yet another mini dsl.

gridbugs:
Closing this. We can solve https://github.com/ocaml/dune/issues/9451 with a less general dsl than slang.

PR #9534, opened 2023-12-19
doc(pkg): Add documentation on how the revision store works

Leonidas-from-XIV:
This documents the rev store, which will probably make it easier to understand the idea behind it.

### DIFF
--- a/doc/dev/rev-store.md
+++ b/doc/dev/rev-store.md
@@ -1,0 +1,106 @@
+The Revision Store
+==================
+
+The revision store is the place where Git data that is relevant to the Dune
+package management is cached.
+
+The Concepts
+------------
+
+The revision store uses Git in the way of its original slogan, as a
+content-adressable file system. A lot of data (code) and meta-data (opam files)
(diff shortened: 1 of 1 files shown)


doc/dev/rev-store.md
```diff
@@ -0,0 +1,124 @@
...
+
+The implementation of these features is a mix of calling the `git` binary and
+implementing parts in OCaml. This means that the `git` binary is required on
+the system. Possible future improvements could be using
+[ocaml-git](https://github.com/mirage/ocaml-git) to avoid the dependency.
+
+Getting the Contents
+--------------------
+
+The `Rev_store` module contains modules/types that are reflecting specific
```
rgrinberg, line 111: I feel like details of this contents are going to change and are best documented in `Rev_store` itself.

rgrinberg (approved):

doc/dev/rev-store.md
Leonidas-from-XIV: Good point, I'll remove this section. The `mli` should be clear enough and if not then it can be documented there.

PR #9536, opened 2023-12-19
repro(pkg): Test the behavior with submodules

Leonidas-from-XIV:
I wanted to write a repro case to show whether it works or not with submodules (not that I expected it to) but turns out that it triggers some issue where the default branch can't be determined. I think it is useful as a repro case either way to solve all the issues that are on the way of supporting submodules.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/opam-repository-download.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-repository-download.t
@@ -224,3 +224,56 @@ So we should get `bar.1.0.0` when locking.
   Solution for dune.lock:
   - bar.1.0.0
   - foo.0.1.0
+
+We want to make sure locking works even with submodules:
+
+  $ rm -r mock-opam-repository
+  $ mkrepo
+  $ mkpkg foo <<EOF
+  > EOF
+  $ mkpkg bar <<EOF
+  > depends: [ "foo" ]
+  > EOF
+
+We move `packages` into a different folder
+
+  $ mkdir remote-submodule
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/pkg/opam-repository-download.t
```diff
@@ -224,3 +224,55 @@ So we should get `bar.1.0.0` when locking.
...
+  $ cat > dune-workspace <<EOF
+  > (lang dune 3.10)
+  > (repository
+  >  (name mock)
+  >  (source "git+file://$(pwd)/mock-opam-repository"))
+  > (lock_dir
+  >  (repositories mock))
+  > EOF
+
+We should be able to successfully the project with `foo` and `bar`:
```
gridbugs, line 274:
```suggestion
We should be able to successfully solve the project with `foo` and `bar`:
```

test/blackbox-tests/test-cases/pkg/opam-repository-download.t
```diff
@@ -224,3 +224,55 @@ So we should get `bar.1.0.0` when locking.
...
+
```
gridbugs, line 275: Maybe mention here that this is currently not supported.

gridbugs (approved):

rgrinberg (approved):

test/blackbox-tests/test-cases/pkg/opam-repository-download.t
Leonidas-from-XIV: Good point, I added a note that this is not the desired behavior.

Issue #9511, opened 2023-12-15
Validate solver vars in lockdir against those in the workspace


gridbugs:
The solver variables that were read while generating a lockdir are stored in the lockdir. Some operations on lockdirs use the variables in the lockdir while others read the whole solver variable environment from the workspace. To avoid inconsistencies we should validate that all the solver variables in the lockdir are assigned the same values in workspace file. The two ways inconsistencies can be introduced is by a user changing a variable in the workspace file in a way that would affect the dependency solution, and by directly tampering with the lockdir.

Note that there will likely be variables in the lockdir with no counterpart in the workspace file. These would have been polled from the system where the lockdir was generated and should be ignored by this validation.

rgrinberg:
I think this is done?

gridbugs:
Yep

Issue #9406, opened 2023-12-07
Warnings about Lock Directory Portability


rgrinberg:
Since we now record the value of all the variables evaluated during solving, we can now detect if one of these variables does not match the build time environment. For example, if we used `%{os}% = "linux"` to compute some dependencies, we can warn the user trying to run the lock directory on macos.

gridbugs:
Yep, for each combination of os, arch, distro, etc that affects the dependency solution there will be a separate lockdir. 

I'm not sure we should even attempt to proceed with a build if the recorded solver env differs from the current env, at least not by default. We attempt to filter build/install commands at solve time, and if a necessary command for your env was filtered out at solve time it won't be in the lockdir at all and you'll have no way of knowing that this was the case without manually checking the original opam file. This may lead to strange errors which are hard to debug. Since reliable builds is one of the selling points of lockdirs I think dune should be more conservative about attempting builds where this is a possibility.

PR #9546, opened 2023-12-20
pkg: validate vars in lockdir against workspace

gridbugs:
Adds a check that the variables in lockdir metadata agree with the variable assignments in the workspace file.

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -112,8 +112,22 @@ module Lock_dir = struct
     | Some (Opam _) -> None
   ;;
 
+  let get_workspace_lock_dir ctx =
+    let* workspace = Workspace.workspace () in
+    let+ path = get_path ctx >>| Option.value_exn in
+    Workspace.find_lock_dir workspace path
+  ;;
+
   let get (ctx : Context_name.t) : t Memo.t =
-    get_path ctx >>| Option.value_exn >>= Load.load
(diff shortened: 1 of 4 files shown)


gridbugs:
Fixes https://github.com/ocaml/dune/issues/9511

src/dune_rules/pkg_rules.ml
```diff
@@ -112,8 +112,22 @@ module Lock_dir = struct
...
+
   let get (ctx : Context_name.t) : t Memo.t =
-    get_path ctx >>| Option.value_exn >>= Load.load
+    let* lock_dir = get_path ctx >>| Option.value_exn >>= Load.load in
+    let+ workspace_lock_dir = get_workspace_lock_dir ctx in
+    (match workspace_lock_dir with
+     | None -> ()
+     | Some workspace_lock_dir ->
+       Solver_stats.Expanded_variable_bindings.validate_against_solver_env
+         lock_dir.expanded_solver_variable_bindings
```
rgrinberg, line 128: Does this include the variables set by default? Like `os`, etc.

gridbugs: No, only the variables set in the workspace. Otherwise this would be enforcing that lockdirs only build on systems compatible with solver variables stored in their metadata (ie. https://github.com/ocaml/dune/issues/9406).

rgrinberg (approved):

PR #9390, opened 2023-12-06
fix(pkg): Avoid `git` translating its CLI

Leonidas-from-XIV:
Turns out that some Git commands translate their messages e.g. when running with a `fr_FR` locale. Then the regex fails and as such the code fails to determine the default branch.

This PR makes sure to always run Git in English, though a better solution would be to find a lower-level command that determines the same info without having to parse high-level commands that get translated.

PR #9549, opened 2023-12-20
fix(pkg): Read the main branch correctly even if tags exist

Leonidas-from-XIV:
The issue was that `git remote show -n <handle>` will return tags as branches, so if a repo has tags, it will fail to determine the default branch. This has been possible since #9471.

`git remote show` is problematic in many ways (e.g. it localizes output, thus already having lead to issues, see #9390), so instead this code reads the branch from the fetch `refspec`, which is locale-independent.

This does not fix the issue that submodules are not supported yet, but it will determine the branch correctly on the way to determining submodules and not fail with an unrelated error.

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -407,18 +407,31 @@ let query_head_branch =
       | Some m -> Some (Re.Group.get m 1))
 ;;
 
+let branch_of_refspec refspec =
+  refspec
+  |> String.drop_prefix_if_exists ~prefix:"+"
+  |> String.lsplit2 ~on:':'
+  |> Option.bind ~f:(fun (remote_ref, _local_ref) ->
+    String.drop_prefix remote_ref ~prefix:"refs/heads/")
+;;
+
 let read_head_branch =
-  let headline = Re.(compile @@ seq [ bol; rep space; str "Remote branch:" ]) in
+  let fetch_line = Re.(compile @@ seq [ str "fetch = "; group (rep1 any); eol ]) in
   fun t handle ->
-    let+ lines = run_capture_lines t [ "remote"; "show"; "-n"; handle ] in
-    let rec inspect = function
-      | [] | [ _ ] -> None
-      | heading :: branch :: rest ->
-        (match Re.exec_opt headline heading with
-         | None -> inspect (branch :: rest)
-         | Some _ -> Some (String.trim branch))
-    in
-    inspect lines
+    let headline = sprintf {|[remote "%s"]|} handle in
+    let path = Path.relative t.dir "config" in
+    let lines = Io.lines_of_file path in
+    let _front, back = List.split_while lines ~f:(String.equal headline) in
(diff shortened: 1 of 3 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -407,18 +407,40 @@ let query_head_branch =
...
+    | line :: lines ->
+      (match Re.exec_opt fetch_line line with
+       | None -> inspect_in_section lines
+       | Some m ->
+         let refspec = Re.Group.get m 1 in
+         (match branch_of_refspec refspec with
+          | Some b -> Some b
+          | None -> inspect_in_section lines))
+  in
+  let rec inspect = function
```
rgrinberg, line 436: I think this is just `List.find_map`?

src/dune_pkg/rev_store.ml
```diff
@@ -407,18 +407,40 @@ let query_head_branch =
...
+    String.drop_prefix remote_ref ~prefix:"refs/heads/")
+;;
+
+let read_head_branch t handle =
+  let headline =
+    Re.(compile @@ seq [ bol; str (sprintf {|[remote "%s"]|} handle); eol ])
+  in
+  let fetch_line = Re.(compile @@ seq [ str "fetch = "; group (rep1 any); eol ]) in
+  let path = Path.relative t.dir "config" in
+  let lines = Io.read_file path |> String.split ~on:'\n' in
```
rgrinberg, line 424: We have `Io.lines_of_file`

src/dune_pkg/rev_store.ml
```diff
@@ -407,18 +407,40 @@ let query_head_branch =
...
+  let rec inspect_in_section = function
```
rgrinberg, line 425: This can also be replaced by `List.find_map`

src/dune_pkg/rev_store.ml
```diff
@@ -407,18 +407,40 @@ let query_head_branch =
...
+  |> String.lsplit2 ~on:':'
+  |> Option.bind ~f:(fun (remote_ref, _local_ref) ->
+    String.drop_prefix remote_ref ~prefix:"refs/heads/")
+;;
+
+let read_head_branch t handle =
+  let headline =
+    Re.(compile @@ seq [ bol; str (sprintf {|[remote "%s"]|} handle); eol ])
+  in
+  let fetch_line = Re.(compile @@ seq [ str "fetch = "; group (rep1 any); eol ]) in
```
rgrinberg, line 422: The regex can be hoisted up so they are only compiled once

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: It is not, as it pops of lines from the list until the heading is found. But turns out it is fairly similar to `List.split_while`, so I rewrote it to use that.

rgrinberg (approved):

gridbugs (approved):

PR #9550, opened 2023-12-20
refactor: move file contents memoization to [Build_system]

rgrinberg:
and remove it from the rules

Continuation of the splitting of https://github.com/ocaml/dune/pull/9132

### DIFF
--- a/src/dune_rules/expander.ml
+++ b/src/dune_rules/expander.ml
@@ -260,15 +260,11 @@ let[@inline never] invalid_use_of_target_variable
        assert false)
 ;;
 
-let expand_read_macro ~dir ~source s ~read ~pack =
+let expand_read_macro ~dir ~source s ~read =
   let path = relative ~source dir s in
   let read =
     let open Memo.O in
-    let+ x =
-      Build_system.read_file path ~f:(fun a -> Async.async (fun () -> read a))
-      >>= Memo.of_reproducible_fiber
-    in
-    pack x
+    Build_system.read_file path >>| read
   in
   Need_full_expander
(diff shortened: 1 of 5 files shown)


src/dune_rules/recursive_include.ml
```diff
@@ -63,7 +63,7 @@ let decode ~base_term ~include_keyword ~include_allowed_in_versions ~non_sexp_be
 
 let load_included_file config path ~context =
   let open Memo.O in
-  let+ contents = Build_system.read_file (Path.build path) ~f:Io.read_file in
+  let+ contents = Build_system.with_file (Path.build path) ~f:Io.read_file in
```
snowleopard, line 66: Wait, why not use the memoized version here?

src/dune_rules/expander.ml
```diff
@@ -265,7 +265,7 @@ let expand_read_macro ~dir ~source s ~read ~pack =
   let read =
     let open Memo.O in
     let+ x =
-      Build_system.read_file path ~f:(fun a -> Async.async (fun () -> read a))
+      Build_system.with_file path ~f:(fun a -> Async.async (fun () -> read a))
```
snowleopard, line 268: Not sure what `read` is, but maybe worth memoizing it.

snowleopard (approved):
Thanks! I left a couple of suggestions.

src/dune_rules/recursive_include.ml
rgrinberg: Ah yeah, good call. I memoized all the reads now.

PR #9553, opened 2023-12-21
refactor: move more [Action_builder] functions to rules

rgrinberg:
Continuation of the splitting of https://github.com/ocaml/dune/pull/9132

This PR includes more trivial definitions that can be moved to the rules without much modification

### DIFF
--- a/src/dune_rules/action_builder.ml
+++ b/src/dune_rules/action_builder.ml
@@ -3,6 +3,20 @@ include Dune_engine.Action_builder
 open O
 module With_targets = With_targets
 
+type fail = { fail : 'a. unit -> 'a }
+
+let fail x =
+  let+ () = return () in
+  x.fail ()
+;;
+
+let delayed f =
+  let+ () = return () in
+  f ()
+;;
+
+let of_memo_join f = of_memo f >>= Fun.id
+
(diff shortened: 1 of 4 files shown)


src/dune_rules/action_builder.ml
```diff
@@ -3,6 +3,20 @@ include Dune_engine.Action_builder
...
+  let+ () = return () in
+  x.fail ()
+;;
+
+let delayed f =
+  let+ () = return () in
+  f ()
+;;
+
+let of_memo_join f = of_memo f >>= Fun.id
```
rgrinberg, line 18: This new implementation is suboptimal. Perhaps we should introduce `Action_builder.join` to make this a little more natural?

snowleopard: Are you worried about an extra `union` with the empty map? It doesn't seem worth optimising to be honest, especially given that we plan to change the implementation soon.

snowleopard (approved):
Thanks!

src/dune_rules/action_builder.ml
rgrinberg: Nah, I just find the lack of `val join : 'a t t -> 'a t` to be unnatural for a monadic interface.

PR #9559, opened 2023-12-21
refactor: hide [Action_builder.eval_mode] and remove [of_thunk]

rgrinberg:
This PR finalizes https://github.com/ocaml/dune/pull/9132

We ended up keeping `Action_builder.push_stack_frame` because there's no way to implement it externally


snowleopard (approved):
Looks good!

PR #9561, opened 2023-12-22
pkg: store vars in commnad filters in metadata

### LINKED ISSUE #9544: Solver vars used to filter commands are not added to lockdir metadata
Currently we only add variables used to filter dependencies, but we also filter commands at solve-time with solver variables and if a variable is only used to filter commands and not dependencies then it's not added to the lockdir metadata.

gridbugs:
Prior to this change we only stored variables that were used in dependency filters. This change adds variables that were used in command filters.

Fixes https://github.com/ocaml/dune/issues/9544.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -384,7 +384,7 @@ let filter_to_blang ~package ~loc filter =
   filter_to_blang filter
 ;;
 
-let simplify_filter solver_env =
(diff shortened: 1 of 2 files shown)


Leonidas-from-XIV (approved):

rgrinberg (approved):

rgrinberg:
Gonna merge this one to avoid myself conflicts down the line.

PR #9563, opened 2023-12-22
fix(boot): make bootstrap reproducible by default

### LINKED ISSUE #9507: bootstrap is not reproducible wrt -j value
While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds), I found that our [`ocaml-dune` 3.12.1 package](https://github.com/bmwiedemann/openSUSE/blob/2675aba344755ee2d684010e1430dbff0f86cedc/packages/o/ocaml-dune/ocaml-dune.spec#L80) varies depending on the amount of parallelism.

## Expected Behavior
Build output should be deterministic - as it still was in the previous version 3.11.1

## Actual Behavior
Binaries produced by `-j 1` and `-j 4` vary even in length.
A diff of `strings` output looks thus:
```diff
 camlDune_rules__Package_db__fun_2964                                           
-camlOpamSystem__303
 camlDune_engine__Action_exec__fun_5160
 camlDune_engine__Action_builder0__code_begin
+camlStdune__Signal__3410
 camlDune_re__Core__witness_2586
 camlDune_rules__Modules__fun_4126
+camlNotty_uucp_data__16639
 camlDune_rpc_impl__fun_970
[...snipped...]
```
one created a `_boot/compiled_ml_files` and the other a `_boot/mods_list`. Build logs also differ a lot.

## Reproduction

- PR with a reproducing test: 

1. on openSUSE or Debian
```bash
#!/bin/sh
osc co openSUSE:Factory/ocaml-dune && cd $_ || exit 1
for N in 1 2 ; do
    osc build -j $N --vm-type=kvm --noservice --no-preinstallimage --keep-pkg=RPMS.$N
    unrpm RPMS.$N/ocaml-dune-*.x86_64.rpm
    strings usr/bin/dune > $N.strings
done
diff -u {1,2}.strings
```

## Specifications

- Operating system (distribution and version): openSUSE Tumbleweed 20231213

## Additional information 

https://gist.github.com/bmwiedemann/c05ab97a162d966ae40b7a39065efd6e

emillon:
Fixes #9507

In `duneboot.ml`, there are 2 strategies to build `dune.exe`:
- separate compilation (build all modules and link them together in several commands)
- single command compilation (pass all modules to the compiler driver)

Separate compilation can be executed in parallel when `-j` is set accordingly, which speeds up the build.
But when only one job can be executed at once (`-j 1`), it is faster to use a single command.

So, we would switch to single command build when `-j 1` was passed. However, this is surprising because it produces a different binary depending on the number of parallel jobs (see #9507). So this disables this optimization by default; it can be enabled by hand by passing `--use-single-command`.

(NB: this does not apply to Windows, which always uses single command compilation)

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -1,6 +1,13 @@
 (** {2 Command line} *)
 
-let concurrency, verbose, _keep_generated_files, debug, secondary, force_byte_compilation =
+let ( concurrency
+    , verbose
+    , _keep_generated_files
+    , debug
+    , secondary
+    , force_byte_compilation
+    , use_single_command )
+  =
   let anon s = raise (Arg.Bad (Printf.sprintf "don't know what to do with %s\n" s)) in
   let concurrency = ref None in
   let verbose = ref false in
@@ -9,6 +16,7 @@ let concurrency, verbose, _keep_generated_files, debug, secondary, force_byte_co
   let debug = ref false in
   let secondary = ref false in
   let force_byte_compilation = ref false in
+  let use_single_command = ref Sys.win32 in
   Arg.parse
     [ "-j", Int (fun n -> concurrency := Some n), "JOBS Concurrency"
     ; "--verbose", Set verbose, " Set the display mode"
@@ -18,6 +26,10 @@ let concurrency, verbose, _keep_generated_files, debug, secondary, force_byte_co
     ; ( "--force-byte-compilation"
       , Set force_byte_compilation
       , " Force bytecode compilation even if ocamlopt is available" )
+    ; ( "--use-single-command"
+      , Set use_single_command
+      , " Build with a single command instead of separate compilation. This is faster \
+         compared to -j1, but produces a different binary." )
     ]
     anon
     (Printf.sprintf "Usage: %s <options>\nOptions are:" prog);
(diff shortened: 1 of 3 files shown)


boot/duneboot.ml
```diff
@@ -9,6 +16,7 @@ let concurrency, verbose, _keep_generated_files, debug, secondary, force_byte_co
   let debug = ref false in
   let secondary = ref false in
   let force_byte_compilation = ref false in
+  let use_single_command = ref Sys.win32 in
```
rgrinberg, line 19: Could you leave a comment to explain why windows is special?

rgrinberg (approved):
LGTM, but I think a name like `--build-single-command` is more appropriate. "use" is a bit too general, and we're still allowed to run `ocamldep` with multiple commands (for example)

rgrinberg:
a CHANGES entry is necessary as well please

rgrinberg (changes requested):
Actually, don't you need to accept (and ignore) this argument in `boot/bootstrap.ml`? Otherwise it will complain that the argument is unknown and won't display it in `ocaml bootstrap.ml --help`.

emillon:
Closed in favor of #9735.

PR #9568, opened 2023-12-25
Remove warning 30 from the standard set.

gasche:
Warning 30 (same constructor/label name used in two mutually-recursive declarations) is an annoying warning that makes no sense since we introduced type-based disambiguation of fields and constructors. It was disabled by default in OCaml 4.10

  https://github.com/ocaml/ocaml/pull/9046

and there is no reason to keep it in Dune.

Note: I wasn't sure how warning changes are supposed to be versioned (or not). 3eff8102d299bc6f408736ad641b0392957a4ac1 implements versioning for the new warnings 67, 69, but then c967d8a4880732f2f7cba39d4f283154c5ef914e earlier removed a bunch of warning without any specific versioning code. I went with the simplest approach of the two, but happy to take feedback. (This could benefit from a comment in the source besides `(* New warnings should be introduced here *)`.

### DIFF
--- a/src/dune_rules/ocaml_flags.ml
+++ b/src/dune_rules/ocaml_flags.ml
@@ -13,6 +13,8 @@ let dev_mode_warnings =
       (Int.Set.of_list
          [ 4; 29; 40; 41; 42; 44; 45; 48; 58; 59; 60; 63; 64; 65; 66; 67; 68; 69; 70 ])
   in
+  let add n range = Int.Set.add range n in
+  let remove n range = Int.Set.remove range n in
   let warnings_range ws =
     let wrange_to_flag (x, y) =
       if x = y then sprintf "@%d" x else sprintf "@%d..%d" x y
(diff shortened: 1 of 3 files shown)


rgrinberg:
> but then https://github.com/ocaml/dune/commit/c967d8a4880732f2f7cba39d4f283154c5ef914e earlier removed a bunch of warning without any specific versioning code

This is just an issue with our suboptimal warning handling code. A new version of OCaml that introduces new warnings will have them enabled in dune by default. So the commit above simply disabled these warnings.

I agree with the change this PR proposes, but it needs to be version guarded.

gasche:
@rgrinberg I edited the PR to add versioning. I had to massage the code a bit to make it read nicely enough with 3 different cases instead of 2. (If there was a simple way to ask dune to memoize functions, we could use it instead of the `lazy` soup, but I stayed true to the existing code style.)

src/dune_rules/ocaml_flags.ml
```diff
@@ -25,10 +27,28 @@ let dev_mode_warnings =
...
+    lazy (warnings_range (Lazy.force lazy_range)) in
+  let range_pre_3_3, pre_3_3 =
+    and_warnings @@ lazy all in
+  let range_pre_3_13, pre_3_13 =
+    and_warnings @@ lazy (
+      Lazy.force range_pre_3_3
+      |> add 67
+      |> add 69
+    ) in
+  let range_later, later =
```
rgrinberg, line 41: CI says that `range_later` isn't used anywhere

gasche: Fixed, thanks!

gasche:
@rgrinberg I see that you fixed a conditional, `dune_version <= foo` should have been `dune_version < foo`, but this occurs twice in the code and both places should be fixed. Let me force-push a more complete fix.

rgrinberg:
Sure, please take over.

gasche:
Thanks!

PR #9574, opened 2023-12-26
Optimise Dep.Fact.Files.t

snowleopard:
Importing some internal optimisations (already reviewed, tested and benchmarked).

This PR optimises `Dep.Fact.Files.t` in two ways:
* Stop storing per-file digests: they are not needed -- the aggregate `t.digest` is sufficient.
* Stop storing `dirs`: it makes no sense to record the facts about the whole directory target when depending only on a subset of files inside via a `File_selector`.
* Stop using `File_selector.to_dyn` when computing digests of file selectors.

I also import a few comments with future suggestions, turn some tuples into records (for clarity), and add `action_digest_version` to force rerunning the actions after the digesting scheme changes (not sure why it was missing).

### DIFF
--- a/src/dune_engine/build_system.ml
+++ b/src/dune_engine/build_system.ml
@@ -899,20 +899,18 @@ end = struct
             let+ digest = build_file path in
             path, digest)
         in
-        Dep.Fact.Files.make ~files:(Path.Map.of_list_exn files)
+        Dep.Fact.Files.create ?dir:(Path.as_in_build_dir dir) (Path.Map.of_list_exn files)
       | Build_under_directory_target { directory_target_ancestor = _ } ->
         let+ path_map = build_dir dir in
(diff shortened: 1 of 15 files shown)


snowleopard:
The testsuite here caught a bug in `Dep.Facts.paths` that we missed internally, which I fixed in the last commit. Oops! I'm going to import this bugfix internally and add a test for it.

rgrinberg (approved):

snowleopard:
Not merging yet, since the latest commit seems to be a big regression in terms of allocations internally.

snowleopard:
OK, I pushed a more lightweight fix for the problem. Hopefully, it'll work.

snowleopard:
The testsuite passed, so merging! Thanks the the help, @rgrinberg.

PR #9578, opened 2023-12-28
Allow expansion in `(modules)` field

nojb:
Currently `(modules)` has to be an "ordered set" consisting of literal string elements. This PR proposes to allow expansion to take place in this field, allowing in particular the use of `(:include)` and `%{read-lines:}` to specify lists of modules that depend on other artifacts built by Dune. Apart from representing a real improvement on the expresiveness of the build language, I have a feeling this change would also cover a considerable portion of use-cases of the OCaml syntax.

I suggest to start by looking at the test to understand what is being done in the PR.

The first commit is preparatory: it moves a submodule to avoid a dependency cycle in the second commit. The second commit contains the meat of the change:

- Replace `Ordered_set_lang.t` by `Ordered_set_lang.Unexpanded.t` for all `(modules)` fields (eg `(private_modules)`, `(modules_without_implementation)`, etc).
- Introduce an "expand" step in the function `Modules_field_evaluator.eval` (which now takes and `Expander.t` as argument). This means that we enter the `Action_builder.t` monad in this code; however, immediately after expansion we exit back to `Memo` by calling `Action_builder.evaluate_and_collect_facts`.
- Currently the `fake_modules` logic in `Modules_field_evaluator` is disabled; I didn't know what this was for and the existing implementation used side-effects that wouldn't work well in the `Memo` monad.
- Note that it is not possible to have the build-time dependencies of the `(modules)` field in the same directory as the containing stanza (see the test); however, the error (a dependency cycle in `Memo`) is not currently caught and Dune brutally crashes. How do I handle this error gracefully?

Suggestions for improvements or if I missed anything are warmly welcome! Thanks.

### DIFF
--- a/src/dune_rules/expander.ml
+++ b/src/dune_rules/expander.ml
@@ -55,7 +55,7 @@ type t =
   ; scope : Scope.t
   ; scope_host : Scope.t
   ; context : Context.t
-  ; lookup_artifacts : (dir:Path.Build.t -> Ml_sources.Artifacts.t Memo.t) option
+  ; lookup_artifacts : (dir:Path.Build.t -> Artifacts.Objs.t Memo.t) option
   ; expanding_what : Expanding_what.t
   }
 
@@ -165,7 +165,7 @@ let expand_artifact ~source t a s =
        let name =
          Module_name.of_string_allow_invalid (Dune_lang.Template.Pform.loc source, name)
        in
-       (match Ml_sources.Artifacts.lookup_module artifacts name with
+       (match Artifacts.Objs.lookup_module artifacts name with
         | None ->
           does_not_exist
             ~loc:(Dune_lang.Template.Pform.loc source)
@@ -177,7 +177,7 @@ let expand_artifact ~source t a s =
            | Some path -> dep (Path.build path)))
      | Lib mode ->
        let name = Lib_name.parse_string_exn (Dune_lang.Template.Pform.loc source, name) in
-       (match Ml_sources.Artifacts.lookup_library artifacts name with
+       (match Artifacts.Objs.lookup_library artifacts name with
         | None ->
           does_not_exist
             ~loc:(Dune_lang.Template.Pform.loc source)
--- a/src/dune_rules/expander.ml
+++ b/src/dune_rules/expander.ml
@@ -909,6 +909,16 @@ let expand_and_eval_set t set ~standard =
   Ordered_set_lang.eval set ~standard ~eq:String.equal ~parse:(fun ~loc:_ s -> s)
 ;;
 
+module Unordered (Key : Ordered_set_lang.Key) = struct
+  module Unordered = Ordered_set_lang.Unordered (Key)
+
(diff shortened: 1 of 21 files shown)


rgrinberg:
I haven't reviewed the entire PR, but I can answer this one:

> Currently the fake_modules logic in Modules_field_evaluator is disabled; I didn't know what this was for and the existing implementation used side-effects that wouldn't work well in the Memo monad.

The `fake_modules` logic is to prevent things like:

```
(library
 (modules foo)
 (modules_without_implementation bar))
```

Here `modules_without_implementation` refers to a module that isn't part of the library. The fake module check warns the user about that.

nojb:
> The `fake_modules` logic is to prevent things like:

Thanks! I pushed a commit ("Handle fake modules") to do this without using a reference; seems to be working again now.

test/blackbox-tests/test-cases/modules-expansion.t
```diff
@@ -0,0 +1,177 @@
...
+  Description:
+    ("internal dependency cycle",
+    { frames =
+        [ ("build-file", In_build_dir "default/lst")
+        ; ("<unnamed>", ())
+        ; ("<unnamed>", ())
+        ; ("load-dir", In_build_dir "default")
+        ]
+    })
+  Raised at Memo.Exec.exec_dep_node.(fun) in file "src/memo/memo.ml", line
```
rgrinberg, line 147: We don't include stack traces in the test output (they make the tests fail due to spurious reasons). See how we filter such stacktraces elsewhere with `awk`.

nojb: The stack trace was meant to document the issue to aid reviewing, but I hope we will be able to get rid of it before the PR is merged :)

nojb: I pushed a commit with an attempt for a more user-friendly stack trace ("Human readable stack trace in case of cycle").

nojb:
The feature is now versioned for Dune >= 3.13 and I pushed some (brief) documentation. This is now ready for a first round of review. Thanks!

src/dune_rules/expander.ml
```diff
@@ -909,6 +909,24 @@ let expand_and_eval_set t set ~standard =
   Ordered_set_lang.eval set ~standard ~eq:String.equal ~parse:(fun ~loc:_ s -> s)
 ;;
 
+module Unordered (Key : Ordered_set_lang.Key) = struct
```
rgrinberg, line 912: Can this code be moved to where it's being used? E.g. `modules_field_evaluator`? I don't see how it depends on any private api's of expander.

rgrinberg (approved):
LGTM. It would be nice to add `(mode byte)` to the tests to make them a little faster. I don't see a need for native compilation to test the specific features here.

nojb:
> It would be nice to add `(mode byte)` to the tests to make them a little faster.

Fided, thanks.

Planning to merge once CI passes. Thanks for the review!

nojb:
Merged, thanks!

PR #9586, opened 2023-12-29
Memoize the last unmemoized computation in [build_dep]

snowleopard:
Addresses a CR-soon. Reviewed and benchmarked internally.

### DIFF
--- a/otherlibs/dune-action-plugin/test/depends-on-its-target-by-read-dir/run.t
+++ b/otherlibs/dune-action-plugin/test/depends-on-its-target-by-read-dir/run.t
@@ -12,10 +12,10 @@
 
   $ cp ./bin/foo.exe ./
 
-  $ dune build some_file
-  Error: Dependency cycle between:
-     _build/default/some_file
-  [1]
+  $ dune build some_file 2>&1 | awk '/Internal error/,/unable to serialize/'
+  Internal error, please report upstream including the contents of _build/log.
+  Description:
+    ("unable to serialize exception",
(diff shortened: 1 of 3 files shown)


snowleopard:
@rgrinberg Do you like the new entries in error stack traces? If yes, please accept the changes. If not, feel free to remove the `human_readable_description` from the new Memo table.

I personally think the new entries are helpful (they tell the user why a certain file was built -- it came from a glob).

Btw, internally, our message is like this: `Evaluating a glob in directory %S`. I didn't change yours because I wasn't sure you want it. But I think our message is possibly clearer. How about also including the glob itself?

snowleopard:
Maybe we should simplify the message so that
```
   -> required by _build/default/impl/impl.cma
+  -> required by Evaluating predicate in directory _build/default/impl
   -> required by alias impl/all
```
becomes
```
   -> required by _build/default/impl/impl.cma
+  -> required by _build/default/impl/*.cma
   -> required by alias impl/all
```
? That seems like a clear improvement to me.

rgrinberg (approved):
LGTM. Although there's a regression in the error message of dune action plugin test. It doesn't really matter, but it seems like we're not converting exceptions properly and that can be observed in RPC as well.

snowleopard:
@rgrinberg Thanks! I ended up undoing the change that introduced the extra stack frame, because it felt odd to bundle it with this PR. Feel free to do it separately if you like (I've added a CR-someday for that).

Issue #9567, opened 2023-12-24
Dune init proj should create a .gitignore containing _build


yawaramin:
## Desired Behavior

`dune init proj` should create a `.gitignore` file as part of the new project. The `.gitignore` file should contain `_build`.

Related: #207

yawaramin:
Another idea–let's also add an empty `.ocamlformat` file to the initial project to enable formatting it, since ocamlformat won't work unless this file is present in the project.

chshersh:
Yes to both default `.gitignore` and `.ocamlformat` 🙏🏻 

Having a fully-configured project from the start enables you to focus on relevant parts. I can't tell enough how important it is!

Leonidas-from-XIV:
@yawaramin It does not if `dune init` is run with `.` since #9447 was merged.

yawaramin:
These defaults will be a massive win for the vast majority of users who are using git and the OCaml Platform recommended ocamlformat. Also I don't think we need to tip-toe around the fact that git is the de facto VCS of the OCaml ecosystem.

If you are a darcs and topiary user, these files will have absolutely zero impact on your project, aside from adding a few bytes of overall file size, unless you notice and delete the files before checking in. Which I am fairly sure darcs and topiary users are smart enough to do.

(Also, isn't darcs a dead project and its successor Pijul taking over?)

yawaramin:
>  dune--unlike git, darcs, pijul, nix, npm, etc.--does not init in the $PWD but creates a new directory

Fixed since #9447 

> Neither I or anyone else will know if git and ocamlformat will maintain hegemony in the next decade.

We are not making an irreversible decision here, let's worry about the software of the next decade if and when it actually happens.

> adding a prompt shouldn’t be seen as annoying

Prompts are annoying though, if I tell dune 'make a new project for me', I don't want it to check with me on every single trivial decision, I just want it to do the right thing automatically.

yawaramin:
Personally I don't think dune should have logic to check every possible VCS (what about Fossil?); imho it's not worth the effort to add so much checking to avoid creating a single .gitignore file with a few bytes in it that can immediately be deleted, but ymmv.

yawaramin:
[...]

PR #9588, opened 2023-12-29
Adding .gitignore and .ocamlformat to init #9567

gabyfle:
Hi !

This is my very first time contributing to such a big open-source project in OCaml. As stated in the issue, that would be nice to have a `.gitignore` with with `_build` added as well as an empty `.ocamlformat` file once a project is created with `dune init proj`

This commit simply adds two files to the list of targets to create inside the `proj` function of the `dune-init.ml` `Make` module.

Related to: https://github.com/ocaml/dune/issues/9567


Leonidas-from-XIV:
Hi, thanks for the contribution. As mentioned in #9567 we need to test the behavior when `.gitignore` and `.ocamlformat` already exist and in that case, presumably, keep the old content.

yawaramin:
We just need a way to 'create if not exists, do nothing if exists'.

gabyfle:
> Hi, thanks for the contribution. As mentioned in #9567 we need to test the behavior when `.gitignore` and `.ocamlformat` already exist and in that case, presumably, keep the old content.

Hi, the current behavior is this one when `.ocamlformat` and `.gitignore` files already exists. I bet what you mean is that we should avoid telling the `dune init` command to create those files so that warnings are gone ? Is that the expected behavior ?

```
dune init proj foo .
Warning: File .ocamlformat was not created because it already exists
Warning: File .gitignore was not created because it already exists
Success: initialized project component named foo
```

yawaramin:
The warnings actually make sense here, we are alerting the user of something unusual and perhaps unexpected. This is a good UX imho.

gabyfle:
> The warnings actually make sense here, we are alerting the user of something unusual and perhaps unexpected. This is a good UX imho.

Woups, just pushed some code that prevents creating the files if they already exists.

For reference, running `cargo init .` inside a folder already containing a `.gitignore` adds this to the file:

```.gitignore
# Added by cargo

/target
```

When I'm then trying to make a `cargo init .` in a directory where a `.gitignore` file already have the `/target` directory inside, here is what the file become:

```.gitignore

# Added by cargo

/target


# Added by cargo
#
# already existing elements were commented out

#/target
```

That would implies ""parsing"" the `.gitignore` file but seems like a usefull feature.

rgrinberg:
@Julow wdyt? I thought `.ocamlformat` is no longer necessary in newer versions of ocamlformat

Leonidas-from-XIV:
@gabyfle What I mean is that this PR should have an expect test that checks the behavior. The warning seems fine to me, but I would like to have the test to make sure that it 1) works and creates the files 2) doesn't overwrite the files if they already exist. This way we can be sure that this functionality is not broken by accident in future releases without us noticing.

gabyfle:
> @gabyfle What I mean is that this PR should have an expect test that checks the behavior. The warning seems fine to me, but I would like to have the test to make sure that it 1) works and creates the files 2) doesn't overwrite the files if they already exist. This way we can be sure that this functionality is not broken by accident in future releases without us noticing.

Ho ok, sure thing, I'll try to create a specific expect-test for this feature and push it here !

bin/dune_init.ml
```diff
@@ -519,7 +519,19 @@ module Component = struct
...
-        { dir; files = dune_project_file dir opts :: package_files }
+        let default_files =
+          let exists =
+            Path.exists (Path.relative dir ".ocamlformat")
+            && Path.exists (Path.relative dir ".gitignore")
+          in
+          if exists
+          then []
+          else
+            [ File.make_text dir ".ocamlformat" ""
```
Julow, line 530: It would be useful to set the version, if ocamlformat is installed at this time.

```
version = <output of ocamlformat --version>
```

Julow (commented):
> @Julow wdyt? I thought .ocamlformat is no longer necessary in newer versions of ocamlformat

We did not merge this change. The `.ocamlformat` file is needed to activate ocamlformat.

bin/dune_init.ml
Leonidas-from-XIV: This could be annoying if you create a project, then a local switch and that switch then uses a different ocamlformat.

yawaramin: Yeah, we are aiming for the basics to get started here, it's fine to not lock things doing at this point. Later if the user needs a specific version they can always set it. We are not making an irreversible decision here.

Julow: The version mismatch is explicitly annoying but without specifying the version, the code might be formatted differently on the different switches.

We recommend always specifying a version in `.ocamlformat` files.

yawaramin: If we specify a version in the `.ocamlformat` file, then we upgrade to a newer version of ocamlformat, it will refuse to run because of the mismatched version. Then the options are to either change the version number in the file and reformat the code, or to downgrade ocamlformat and keep the formatting unchanged.

If we don't specify a version and we upgrade to a newer version, it will run and reformat the code, in which case we will notice and either accept the reformat and check it in, or downgrade ocamlformat and keep the formatting unchanged.

I don't see any significant difference between the options.

Julow: In the second scenario, it might be hard to not accept the new formatting as new code have been written. You'd have to remember the previous version of ocamlformat you used and install that.
The change in formatting might also be unnoticed and pollute the history and that might be frequent when switching back and forth between two switches or working on other projects that use different versions.
Formatting check in CI needs to know the version.
As this can affect the entire source tree, the ocamlformat team thinks that the explicit version check is better.

gabyfle: Then I suppose that adding the version is the more logical choice if it's recommended by the `ocamlformat` team. I'll try to push a commit as soon as possible to add this feature.

Thanks you all for your time.

bin/dune_init.ml
```diff
@@ -524,10 +524,41 @@ module Component = struct
...
+                      ++ Pp.textf ": couldn't get the ocamlformat version number."
+                    ]);
+               Console.print_user_message
+                 (User_message.make
+                    [ Pp.tag User_message.Style.Hint (Pp.verbatim "Format")
+                      ++ Pp.textf ": Are you sure you have ocamlformat installed ?"
+                    ]));
+              ""
+          in
           if exists
```
gabyfle, line 554: Should I keep this close or not then ? What do you think ? Removing it make `dune` display a nice warning when files already exists in the target directory (when using `dune init proj .` for example)

test/expect-tests/dune_init_proj/dune_init_proj_tests.ml
gabyfle, line 1: Unfortunately, for some reasons, I'm unable to launch `make test` even tho I ran `make dev-deps`. I'm not sure what's the reason, because all the dependencies are installed inside my current `opam switch` but I still got lots of `library not found` when trying to run the command.

So warning, because this test hasn't been tested and it may not work. This is still "work in progress" until I can run the tests.

rgrinberg: Are you using a local switch or a global switch? You might want to try running command inside the switch using `opam exec -- make test`

test/expect-tests/dune_init_proj/dune_init_proj_tests.ml
```diff
@@ -0,0 +1,46 @@
+open Stdune
```
Leonidas-from-XIV, line 1: Sorry, I mentioned expect thest but this is better as a cram test, since it only calls commands, for which `cram` is way better suited.

tests_result
```diff
@@ -0,0 +1 @@
+./_boot/dune.exe runtest
```
Leonidas-from-XIV, line 1: This should presumably not be part of this PR?

bin/dune_init.ml
```diff
@@ -519,7 +519,43 @@ module Component = struct
             [ File.make_text (Path.parent_exn opam_file) (Path.basename opam_file) "" ]
           | Esy -> [ File.make_text dir "package.json" "" ]
         in
-        { dir; files = dune_project_file dir opts :: package_files }
+        let default_files =
+          let fmt_version =
+            try
+              let ic = Unix.open_process_in "ocamlformat --version" in
```
Leonidas-from-XIV, line 525: Instead of `open_prcess_in`, I think it would be better to use Dune's `Process`.

yawaramin:
Please keep comments specifically related to reviewing the PR. For general comments about the feature please comment on the related issue.

bin/dune_init.ml
```diff
@@ -519,7 +519,43 @@ module Component = struct
...
+              let version = input_line ic in
+              let _ = Unix.close_process_in ic in
+              (let open Pp.O in
+              Console.print_user_message
+                (User_message.make
+                   [ Pp.tag User_message.Style.Hint (Pp.verbatim "Format")
+                     ++ Pp.textf ": using ocamlversion number %s" version
```
shonfeder, line 532: Is this meant to be

```suggestion
                     ++ Pp.textf ": using ocamlformat version number %s" version
```

I'd ask us to be careful about adding output, and push back a bit to ask why we need to output anything at all here. The reasons to be careful here is that this kind of thing can quickly adds up to a wall of spam. I don't know whether it is an explicit part of the design philosophy, but dune tends to mostly follow the [Unix Rule of Silence](https://www.linfo.org/rule_of_silence.html), and I suggest we do that here as well.

This output could be moved into the debugging logs, but I'm not evens sure why we'd need it there. WDYT?

bin/dune_init.ml
```diff
@@ -519,7 +519,43 @@ module Component = struct
...
+                   ]));
+              version
```
shonfeder, line 534: Hello! Sorry to be coming late to this PR.

The design of the `init` subcommand centers around the values of type `Component.t`, that help abstract over the kinds of stuff it will create. See

https://github.com/ocaml/dune/blob/581f852/bin/dune_init.mli#L23-L25

Rather than creating these files unconditionally, and rather than extending the project component directly, I suggest we add a new component (or two) to support this kind of tooling config file. This would offer at least the following benefits:

- We could add a subcommand (or flag) to enable creation of just these files, which could be useful as the amount of such files grows (e.g., maybe we also want to include generation of CI configs).
- It provides a natural place for providing options for the kind of customization and specialization that @toastal has been advocating for, since each component is parameterized on a set of options.
- It will keep the design of `init` more composable and easier to read about, which should help maintenance in the future: rather than creating files 

The concrete steps for this change would be:

1. Specify an aptly named component (maybe like `Tooling_configs` or do we want a different abstraction?)
2. Write a function that will take the options that configure the component to a list of [targets](https://github.com/ocaml/dune/blob/581f852/bin/dune_init.ml#L310.-L314).
3. Put those pieces together to extend the existing targets being generated by the project component in https://github.com/ocaml/dune/blob/581f852/bin/dune_init.ml#L552-L559

This feels like the right approach to me given the existing design, but I'd be willing to discuss if you (or others) feel it isn't in keeping with the aims here.

bin/dune_init.ml
```diff
@@ -519,7 +519,43 @@ module Component = struct
...
+                (User_message.make
+                   [ Pp.tag User_message.Style.Warning (Pp.verbatim "Format")
+                     ++ Pp.textf ": couldn't get the ocamlformat version number."
+                   ]);
+              Console.print_user_message
+                (User_message.make
+                   [ Pp.tag User_message.Style.Hint (Pp.verbatim "Format")
+                     ++ Pp.textf ": Are you sure you have ocamlformat installed ?"
+                   ]));
+              ""
```
shonfeder, line 548: I'm not sure what to do here, but:

- I don't think we should have a catch-all exception handler, lest we hide unrelated errors
- If I have chosen not to use ocamlformat as part of my workflow, then I don't want this noise every time I initialize a project.

shonfeder (changes requested):
Thanks for adding this new functionality! Great work so far :)

I didn't see this WIP until now, so sorry for the late edition comments. I'd be happy to talk my suggestions thru and offer any other support that you'd find useful.

bin/dune_init.ml
```diff
@@ -519,7 +519,43 @@ module Component = struct
...
+          in
+          (* get the current ocamlversion *)
+          [ File.make_text
+              dir
+              ".ocamlformat"
+              (if fmt_version = "" then "" else "version = \"" ^ fmt_version ^ "\"\n")
+          ; File.make_text dir ".gitignore" "_build/"
```
shonfeder, line 555: Maybe we should probably also add the `_opam` directory here?

test/expect-tests/dune_init_proj/dune_init_proj_tests.ml
shonfeder: @gabyfle most of the `cram` test for this subcommand are in https://github.com/ocaml/dune/tree/main/test/blackbox-tests/test-cases/dune-init.t/run.t

bin/dune_init.ml
shonfeder: One other thought here: currently `Dune_init` is designed to push all the I/O operations to the edge, and only have pure computations deriving all of the components. The main benefit is that it makes it easy to reason about compositions of different component generation logic. The addition of this interaction with the `ocamlformat` executable in the middle of deriving the project violates this design approach. Perhaps we can find a way to move this kind of action into a configuration phase, rather then during component derivation? I'd even suggest moving this kind of I/O into https://github.com/ocaml/dune/tree/main/bin/init.ml , as this seems analogous to our creation of the new project directory and initilization of the build context to my mind. But I wonder if core maintainers think that is the right place?

bin/dune_init.ml
shonfeder: Actually, on further thought, we can always add this on refactor. No need to complicate your PR. Sorry for the noise!

rgrinberg:
Please re-open if you ever resume work on this PR.

PR #9591, opened 2023-12-30
feature(coq): extra_sources field for coq.theory stanza

Alizter:
We add an extra_sources field to the coq-theory and coq.extraction stanzas that allows the decleration of extra sources. This allows Coq features such as Load and Extra Dependency to be used correctly.

We add two tests that demonstrates how it can be used for those features.

The extra_sources field finds the declared files and makes sure that Dune declares a dependency on them before calling coqdep for the theory being built.

- [ ] docs
- [ ] changelog

### DIFF
--- a/src/dune_rules/coq/coq_rules.ml
+++ b/src/dune_rules/coq/coq_rules.ml
@@ -463,6 +463,7 @@ let setup_coqdep_for_theory_rule
   ~ml_flags
   ~mlpack_rule
   ~coq_lang_version
+  ~extra_sources
   coq_modules
   =
   let* boot_type =
@@ -474,7 +475,9 @@ let setup_coqdep_for_theory_rule
     | m :: _ -> Bootstrap.make ~dir ~use_stdlib ~wrapper_name ~coq_lang_version m
   in
   (* coqdep needs the full source + plugin's mlpack to be present :( *)
-  let sources = List.rev_map ~f:Coq_module.source coq_modules in
+  let sources =
+    List.map ~f:Path.build extra_sources @ List.rev_map ~f:Coq_module.source coq_modules
+  in
   let file_flags =
     [ Command.Args.S
         (coqc_file_flags ~dir ~theories_deps ~wrapper_name ~boot_type ~ml_flags)
@@ -718,13 +721,13 @@ let coq_modules_of_theory ~sctx lib =
     Coq_sources.library coq_sources ~name |> List.rev_map ~f:Coq_module.source
 ;;
 
-let source_rule ~sctx theories =
+let source_rule ~sctx extra_sources theories =
   (* sources for depending libraries coqdep requires all the files to be in the
      tree to produce correct dependencies, including those of dependencies *)
   Action_builder.dyn_paths_unit
     (let open Action_builder.O in
      let+ l = Action_builder.List.map theories ~f:(coq_modules_of_theory ~sctx) in
-     List.concat l)
+     List.map ~f:Path.build extra_sources @ List.concat l)
 ;;
 
 let setup_coqdoc_rules ~sctx ~dir ~theories_deps (s : Coq_stanza.Theory.t) coq_modules =
@@ -901,7 +904,10 @@ let setup_theory_rules ~sctx ~dir ~dir_contents (s : Coq_stanza.Theory.t) =
     in
     let open Action_builder.O in
     let* theories = Resolve.Memo.read theories in
-    source_rule ~sctx theories
+    source_rule
+      ~sctx
+      (List.map ~f:(Path.Build.relative dir) s.buildable.extra_sources)
+      theories
   in
   let coqc_dir = Super_context.context sctx |> Context.build_dir in
   let* mode = select_native_mode ~sctx ~dir s.buildable in
@@ -917,6 +923,7 @@ let setup_theory_rules ~sctx ~dir ~dir_contents (s : Coq_stanza.Theory.t) =
     ~ml_flags
     ~mlpack_rule
     ~coq_lang_version
+    ~extra_sources:(List.map ~f:(Path.Build.relative dir) s.buildable.extra_sources)
     coq_modules
   >>> Memo.parallel_iter
         coq_modules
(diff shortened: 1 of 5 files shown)


gares:
I tested it, and I made it to work, but it lacks some features and it also does something weird.
From the log I see the list of extra_sources is passed  to coqdep, which is not what coq_makefile does (but does not hurt, apparently). At the same time, since `Extra Dependency` needs the files to be in a logical path, I had to add the mapping by hand. Note that if the files are mapped, then one can omit passing them to coqdep.

Here my dune file:
```
(coq.theory
  (name elpi)
  (package coq-elpi)
  (flags -Q elpi/ elpi)
  (extra_sources  ../elpi/coq-arg-HOAS.elpi   ...omitted.... )
  (plugins coq-elpi.elpi))
```

I have two remarks:
- I could use support for glob expressions, or entire directories, since the list of extra dependencies tends to be large
- as you can see I had to write `-Q elpi/ elpi` and not `-Q ../elpi/ elpi` since coqc is run from the root. It is a bit odd that I'm talking about the same `../elpi/` directory by I don't put the `..`

For my use case `extra_sources_from_dir elpi ../elpi/` would be ideal. 

Of course I could put the .elpi files next to the .v ones, and that would save me the -Q, but I prefer to keep them in a separate directory since I've too many of these. Glob expressions would be kind of necessary anyway.

ejgallego:
Thanks for the PR @Alizter , actually there is something I hadn't realized about elpi files, and is that they are bound by loadpath search.

> We add an extra_sources field to the coq-theory and coq.extraction stanzas that allows the decleration of extra sources. This allows Coq features such as Load and Extra Dependency to be used correctly.

At first I thought that this is what we wanted, but actually elpi files are found via loadpath so they have a logical name.

So indeed, what we need here is an `(elpi_modules ...)` field . In fact, we should actually update the scanning code so we also detect `.elpi` files bound in a `(coq.theory ...)`, so things work out of the box, by dune interpreting the absence of `.elpi` files as `(elpi_modules :standard)`.

Actually if we had a nice syntax for it, we could re-use `(modules ...)` for both, using `(modules ((name foo) (kind elpi)))`, but it is a bit verbose.

So I'd suggest:

- we extend `Coq_source.t` notion of module to support `.v` and `.elpi` files.
- we update the scanning code to detect `.elpi` files.

Rest should be easy, as thing will work out of the box (except for indeed filtering sources that produce a .vo file and those that don't, so we only load the source rules for them)

ejgallego:
> Here my dune file:
> 
> ```
> (coq.theory
>   (name elpi)
>   (package coq-elpi)
>   (flags -Q elpi/ elpi)
>   (extra_sources  ../elpi/coq-arg-HOAS.elpi   ...omitted.... )
>   (plugins coq-elpi.elpi))
> ```

Using `-Q` is flags is not allowed in Dune. We don't enforce this restriction, but we will. Dune owns handling of logical paths and users cannot modify it. Instead, you should define two theories, each one with their own logical path, if you want to use two logical paths.

> I have two remarks:
> 
> * I could use support for glob expressions, or entire directories, since the list of extra dependencies tends to be large

Indeed, as I suggested above, as `.elpi` modules have a logical path, Dune should scan those automatically as it does for .v files. Then they can be manipulated with the ordered set language.

> * as you can see I had to write `-Q elpi/ elpi` and not `-Q ../elpi/ elpi` since coqc is run from the root. It is a bit odd that I'm talking about the same `../elpi/` directory by I don't put the `..`

Yes, this is not supposed to work.

> Of course I could put the .elpi files next to the .v ones, and that would save me the -Q, but I prefer to keep them in a separate directory since I've too many of these. Glob expressions would be kind of necessary anyway.

Note that file location is not the real problem here, you can also put them in a subdir, did you forgot `(include_subdirs qualified)` in your setup?

The main restriction here is that there is only one logical path (and its children) for each `(coq.theory ...)`, so using a single `(coq.theory ...)` with name `elpi`, the `.elpi` modules have logical path `elpi.elpi.module`.

gares:
Note that extra depends are not elpi specific, it would be weird to have an elpi_soirces or an elpi kind, nothing here is elpi specific.

I think I tried to make a separate theory, but it does not help much since depending on a theory is not enough to have it's extra sources be copied to _build, as far as I can remember.

ejgallego:
> I think I tried to make a separate theory, but it does not help much since depending on a theory is not enough to have it's extra sources be copied to _build, as far as I can remember.

Indeed, the problem is that we only recognize .v files; fixing that should make everything work.

We could recognize all kind of files in the source directory, so that would be complete w.r.t. `Extra Depends`, I think for now it is fine just to also scan `.elpi` files, recognizing every file as a source seems like it could create loads of problems.

ejgallego:
Extra Depends design, as it is, is indeed problematic for people wanting hygienic builds, as it basically forces copying everything on the source dir to the build directory, _before dependency discovery_ (so the build system cannot know what to copy)

`coqdep -modules` would solve this, but meanwhile recognizing `.elpi` fine seems OK to me.

ejgallego:
Many sources happen to be generated, or often manipulated, so the distinction is not immediate.

Moreover caching and sandboxing do require identifying precisely the sources too.

But indeed a design where Dune can match extra deps when they are explicit would be nice, with auto-scanning for Elpi modules.

src/dune_rules/coq/coq_rules.ml
```diff
@@ -718,13 +721,13 @@ let coq_modules_of_theory ~sctx lib =
     Coq_sources.library coq_sources ~name |> List.rev_map ~f:Coq_module.source
 ;;
 
-let source_rule ~sctx theories =
+let source_rule ~sctx extra_sources theories =
```
ejgallego, line 724: I think we should also include the `extra_sources` of the theories we depend on, maybe time to introduce a `coq_sources_of_theory` function?

Alizter:
@ejgallego Feel free to take over this PR. We probably record some duplciated deps somewhere btw from looking at the console commands. I don't have a lot of time to work on this any more.

rlepigre:
@ejgallego if you could summarize what needs to be done, I can probably help to push this trough soonish.

I'm also wondering: do the `.elpi` files need to be installed? If yes, is already done here?

gares:
No, they don't need to be installed since they are included in the .vo files that accumulate them. Still, I think it is nice to have them installed.

gares:
Ping @ejgallego

gares:
any news on this front? I'd really like to use dune (and hence ppxes) in coq-elpi!

gares:
Ping @ejgallego

gares:
Is there any other dune developer that can take care of this PR other then you, @ejgallego ?
I don't know who else could be pinged.

ejgallego:
I don't know.

gares:
@Alizter could you ping some other dune dev to assign?

ejgallego:
Had some recent discussion with @rlepigre about this, I think we can go ahead once we have tests setup.

IIANM, we can do an elpi-free test of this by using the extra dependency command for Coq.

gares:
That would be great.

ejgallego:
This should be reopened for rocq once #12035 is merged.

PR #9600, opened 2024-01-02
refactor: remove [chop_extension] from [Stdune]

rgrinberg:
This function raises [Invalid_argument ..] on error which is almost
never what we want. We shadow it and recommend using [remove_extension]
instead.

@snowleopard I don't think this function is being used internally, but copying you just in case it is.


snowleopard:
@rgrinberg Indeed, it's unused internally.

snowleopard (approved):
Looks good to me.

PR #9606, opened 2024-01-02
fix(pkg): include all error messages

rgrinberg:


bin/pkg/lock.ml
```diff
@@ -89,16 +89,19 @@ let solve
                version_preference
                ~update_opam_repositories
                solver_env_from_current_system)
+     >>| List.partition_map ~f:Result.to_either
    in
-   Result.List.all solutions)
```
rgrinberg, line 93: `Result.List.all` terminates when it finds the first `Error` constructor. We'd like to report all of them instead.

gridbugs (approved):

PR #2854, opened 2019-11-05
New bootstrap

dra27:
OK, here goes!

The good:
- There are now no .boot.ml files 🎉
- It's faster on everything I've tried it on for `-j 1`
- It's comparable and usually faster for `-j 4`
- It integrates my previous work on the secondary compiler, so this dune can be installed in a switch running OCaml 4.02+:

```bash
opam switch create dune-secondary --repos=secondary=git+https://github.com/dra27/opam-repository#secondary ocaml-base-compiler.4.02.3
opam install ocamlfind-secondary
```

The bad:
- The scheduler is not as good on high performance machines - on our 72 core machines, this bootstrap's taking 9sec or so where the old one took 6sec.
- The scheduler seems to suck on Windows (it's actually faster not to parallelise) - this isn't true of Dune itself (although the `-j 1` build on Windows of this bootstrap is comparable to a parallel build on Windows). Not sure whether it's the scheduler, Windows or both. Some attention should at some point be put to the fact that a 32-core high performance Windows VM takes six times as long to compile and link dune as 4-core Linux VM on my laptop...

The still-to-do:
- Work out suitable parallelism
- `setup.ml` is written to the tree at the moment (i.e. `configure.ml` support needs to go back in)
- bootstrap info needs to be fed into the script properly
- I have no story at the moment for the build_info

I've tried various hybrid schemes on the way to this - apologies in advance for my generally low standard of code.

PR #9613, opened 2024-01-03
boot: use several processes to bootstrap on Win32

emillon:
This workaround does not seem to be necessary (anymore?); the mechanism with several commands work, including when `concurrency = 1`.

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -1193,7 +1193,7 @@ let main () =
        | None -> []
        | Some flags -> flags)
   in
-  let build = if concurrency = 1 || Sys.win32 then build_with_single_command else build in
+  let build = if concurrency = 1 then build_with_single_command else build in
   build ~ocaml_config ~dependencies ~c_files ~link_flags task
 ;;
 


emillon:
@dra27 you wrote this bootstrap code; do you remember why it was useful to have windows-specific behavior here?
(context is #9507 and we're wondering if we should remove that code path).

I'll also try to find a way to benchmark how much memory that saves.

dra27:
I have a vague recollection it was sufficiently faster to justify the extra code path

dra27:
Indeed - it seemed to be just empirical evidence (see https://github.com/ocaml/dune/pull/2854) - if it’s not the case any more, then marvellous!

emillon:
I ran the numbers:

OS | strategy | time
-- | -- | --
linux | single | 33.59
linux | j1 | 41.11
linux | j2 | 26.76
linux | j3 | 21.36
linux | j4 | 19.55
windows | single | 380.29
windows | j1 | 302.25
windows | j2 | 280.99
windows | j3 | 323.05
windows | j4 | 346.51

<summary>
Bootstrap times for linux
<details>
image
</details>
</summary>

<summary>
Bootstrap times for windows
<details>
image
</details>
</summary>

I agree that there's something weird going on with Windows, and it's probably pretty noisy (I only took one measurement here) but all in all it doesn't that the separate code path is worth it.

(I tried to benchmark memory usage, but did not fine any difference either using cygwin's `/usr/bin/time` nor with [this approach](https://stackoverflow.com/questions/69784728/powershell-how-can-i-get-memory-consumption-peak-for-my-exe-program-from-power) using powershell).

emillon:
Closed in favor of #9735.

PR #9615, opened 2024-01-03
Pull Dep.Facts.paths computation outside of the Sandbox module

snowleopard:
Upstreaming a part of a larger internal change whose motivation was to avoid calling `Dep.Fact.paths` twice inside the `execute_action_for_rule` function. This isn't particularly important for upstream but let's keep our patch small.

### DIFF
--- a/src/dune_rules/install_rules.ml
+++ b/src/dune_rules/install_rules.ml
@@ -1047,7 +1047,7 @@ let package_deps (pkg : Package.t) files =
         let* res = Dune_engine.Build_system.execute_rule rule in
         loop_files
           rules_seen
-          (Dep.Facts.paths ~expand_aliases:true res.deps
+          (Dep.Facts.paths ~expand_aliases:true res.facts
            |> Path.Set.to_list
            |> (* if this file isn't in the build dir, it doesn't belong to any
                  package and it doesn't have dependencies that do *)
--- a/src/dune_engine/build_system.ml
+++ b/src/dune_engine/build_system.ml
@@ -120,7 +120,7 @@ end
 let () = Hooks.End_of_build.always Metrics.reset
 
 type rule_execution_result =
-  { deps : Dep.Fact.t Dep.Map.t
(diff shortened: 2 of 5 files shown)


src/dune_rules/install_rules.ml
```diff
@@ -1047,7 +1047,7 @@ let package_deps (pkg : Package.t) files =
         let* res = Dune_engine.Build_system.execute_rule rule in
         loop_files
           rules_seen
-          (Dep.Facts.paths ~expand_aliases:true res.deps
+          (Dep.Facts.paths ~expand_aliases:true res.facts
```
snowleopard, line 1050: @rgrinberg Are you sure passing `~expand_aliases:true` here is correct? Shouldn't we be passing `Execution_parameters.expand_aliases_in_sandbox` instead of expanding aliases unconditionally?

rgrinberg: These aren't user rules - so dune has the freedom to pick the correct behavior. I think `expand_aliases:true` is correct, since we are using an alias to collect all the dependency of a particular package.

snowleopard: But... wouldn't a user expect changing this configuration setting affect what install rules are doing?

rgrinberg: The install rules aren't really configurable in this way. E.g. when one writes:

```
(rule
 (deps (package foo))
 (..))
```

The expectation is that the package `foo` is built *and* its artifacts are present in _build/install. There's an alias that introduces the layer of indirection between these two facts, and for it to work as expected, `expand_aliases` has to be set. Now we could get rid of using the alias this way, but it we'll need to reimplement the install rules and I recall running into problems doing it without the level of indirection from the alias.

snowleopard: I see, thanks! I was mostly wondering if we can avoid calling `Dep.Facts.paths` in two different ways. It sounds like we can't.

rgrinberg (approved):

PR #9616, opened 2024-01-03
refactor(pkg): move global rev_store to [Rev_store]

rgrinberg:


Leonidas-from-XIV (approved):
I would probably call it something else than `get` (maybe `singleton`) but I do think that keeping it in `Rev_store` is a better place, as everything that needs it can access it now.

PR #9618, opened 2024-01-04
fix(pkg): remove branch from remote names

rgrinberg:
This way we can reuse the same remote if we're using different branches
(or tags, etc)

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -583,7 +583,6 @@ end
 module Remote = struct
   type nonrec t =
     { repo : t
-    ; handle : string
     ; source : string
     ; default_branch : string
     ; add_remote : string -> unit Fiber.t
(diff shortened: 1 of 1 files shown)


Leonidas-from-XIV:
The reason why the branch is part of the handle is a deliberate choice. The problem is if we remove the branch from the remote name, we can't determine the default branch from the handle anymore (as it would have multiple branches associated with it) without an online connection to the remote, thus making offline use impossible.

Given a remote will share commits with any other remote, adding the same remote multiple times with different branches should not create much overhead, basically just the `config` entries and a few files pointing at refs.

rgrinberg:
> The problem is if we remove the branch from the remote name, we can't determine the default branch from the handle anymore (as it would have multiple branches associated with it) without an online connection to the remote, thus making offline use impossible.

I'm not sure I understand this. The "handle" is a digest, so there's no way invert it in anyway to determine the default branch. Or do you mean some entry .git/config? In that case, the same URL will have the same default branch anyway.

Also, if a user writes `git+https://github.com/foo/bar` as an opam source, we should indeed check the default branch remotely on every fetch. Otherwise, we won't stay up to date if the default branch changes.

Issue #9593, opened 2024-01-02
dune subst Error: stat(test): No such file or directory on semgrep


jmikedupont2:
## Expected Behavior
should work as part of semgrep build in opam

## Actual Behavior

"Error: stat(test): No such file or directory"

```
dune subst  --debug
Error: stat(test): No such file or directory
Raised by primitive operation at Stdune__Path.stat_exn in file
  "otherlibs/stdune/src/path.ml", line 1322, characters 17-40
Called from Main__Subst.is_a_source_file in file "bin/subst.ml", line 19,
  characters 5-25
Called from Main__Subst.subst.(fun) in file "bin/subst.ml", line 360,
  characters 7-28
Called from Stdlib__List.iter in file "list.ml", line 110, characters 12-15               Called from Fiber__Core.O.(>>|).(fun) in file "vendor/fiber/src/core.ml",                   line 252, characters 36-41
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
```

## Reproduction

- PR with a reproducing test: 

1.
1.
1.

## Specifications

- Version of `dune` (output of `dune --version`):
- Version of `ocaml` (output of `ocamlc --version`)
- Operating system (distribution and version):

## Additional information 

- Link to gist with verbose output (run `dune` with the `--verbose` flag):

jmikedupont2:
So my thoughts are that we can just add a check for a broken symlink and skip it.

nojb:
In order to investigate this issue, reproduction instruction would be needed, ideally in an example as small as possible. Thanks!

PR #9637, opened 2024-01-05
test: subst should not fail on broken symlinks

emillon:
See #9593

### DIFF
--- a/test/blackbox-tests/test-cases/subst/broken-symlink.t
+++ b/test/blackbox-tests/test-cases/subst/broken-symlink.t
@@ -1,0 +1,23 @@
+dune subst should not fail when encountering broken symlinks.
+See #9593.
+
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/subst/broken-symlink.t
```diff
@@ -0,0 +1,19 @@
...
+
+  $ git init -q
+  $ cat > dune-project << EOF
+  > (lang dune 1.0)
+  > (name proj)
+  > (package
+  >  (name proj))
+  > EOF
+  $ ln -s nonexistent broken 
+  $ git add dune-project broken
```
rgrinberg, line 12: This git repo is needed to get a version for `dune subst` right? Would be good to clarify in the test.

rgrinberg (approved):

PR #9640, opened 2024-01-05
refactor(pkg): leave CR for untimely update

rgrinberg:

### DIFF
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -172,6 +172,9 @@ let of_git_repo ~update (source : Source.t) =
         | _ -> None
       in
       let* remote = Rev_store.add_repo repo ~source:source.url ~branch in
+      (* CR-rgrinberg: We need to update before deciding whether a ref is a tag
+         or a branch, yet we already have [source], so we've determined this
+         [source] with out of date info? *)
       match update with
       | true -> Rev_store.Remote.update remote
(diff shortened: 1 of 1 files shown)


src/dune_pkg/opam_repo.ml
```diff
@@ -172,6 +172,9 @@ let of_git_repo ~update (source : Source.t) =
         | _ -> None
       in
       let* remote = Rev_store.add_repo repo ~source:source.url ~branch in
+      (* CR-rgrinberg: We need to update before deciding whether a ref is a tag
+         or a branch, yet we already have [source], so we've determined this
+         [source] with out of date info? *)
```
Leonidas-from-XIV, line 177: Not quite: we get a source before updating, because the source could be a commit, in which case we don't need to update (provided the commit is part of the repo). In the other cases we talk to the remote repo (using `ls-remote`).

rgrinberg: Does that mean `--skip-update` isn't respected then?

Leonidas-from-XIV: It is, but it doesn't work for to meaningfully use as an offline mode anymore (offline mode only works in very specific circumstances - you need to specify a revision that exists in the rev store and also pass `--skip-update`). At this point, it should probably be removed since offline mode isn't really feasible in most cases anymore.

Leonidas-from-XIV: Created #9683 for that.

PR #9642, opened 2024-01-05
fix(pkg): pass locations to [Opam_repo.Source] and to command line

rgrinberg:


src/dune_pkg/opam_repo.mli
```diff
@@ -25,13 +25,12 @@ module Source : sig
 
   val commit : t -> Commitish.t option
   val url : t -> string
-  val of_opam_url : OpamUrl.t -> t Fiber.t
-  val to_string : t -> string
```
rgrinberg, line 29: I removed `to_string` from the public API. It's not clear what this string representation is good for (serialization? pretty printing?. We have better types for things like that `Pp.t`, `Dune_sexp.Encoder.t`, etc.

src/dune_pkg/opamUrl0.mli
```diff
@@ -4,7 +4,7 @@ val equal : t -> t -> bool
 val hash : t -> int
 val to_string : t -> string
 val of_string : string -> t
-val decode : t Dune_sexp.Decoder.t
+val decode : (Stdune.Loc.t * t) Dune_sexp.Decoder.t
```
Leonidas-from-XIV, line 7: Is it common that `decode` would return a tuple of `loc * t`?

Leonidas-from-XIV (approved):

src/dune_pkg/opamUrl0.mli
rgrinberg: I suppose we call it `decode_loc` elsewhere. I renamed it

Issue #9517, opened 2023-12-16
dune build -w  exclusion does not work in Mac OS


camlspotter:
## Expected Behavior

The file event monitoring of `dune build -w` should ignore the file creation of an Emacs temp file like `.#foobar.ml`

## Actual Behavior

Generation of `.#foobar.ml` triggers the rebuild, which should not.

## Reproduction

1. Use Mac OS
2. Execute `dune build -w` in your project
3. See the message `waiting for filesystem changes...`
4. Open a terminal and execute `touch .#foobar.ml`
5. `dune build -w` restarts building, which is not an expected behavior.

## Specifications

- `dune.3.11.1`
- Mac OS

## Additional information 

The issue does not reproduce in Linux.

emillon:
Hi,
The core of the issue is that the logic for excluded files is different for each backend, and that for fsevents (the backend used on macos), the logic is partially implemented.

There are 2 ways to do this kind of filtering:
- either when initializing the backend, we tell it to ignore certain files or pattern
- or whenever an event fires, we check the path against the list of exclusions

In the case of fsevents, there's an exclusion list of the first type defined here:

https://github.com/ocaml/dune/blob/013ae8a776155cf0a3a68f109bab34712c3f5b4f/src/dune_file_watcher/dune_file_watcher.ml#L588

But the function that sets exclusions cannot be called with more that 8 entries:
https://github.com/ocaml/dune/blob/013ae8a776155cf0a3a68f109bab34712c3f5b4f/src/fsevents/fsevents.ml#L318

But these are just directories, not patterns, so I think that the length check is there to prevent a solution like evaluating a glob and passing that to the macos APIs.

Regarding the other kind of filtering, we have a list of "standard exclusions" defined here:

https://github.com/ocaml/dune/blob/013ae8a776155cf0a3a68f109bab34712c3f5b4f/src/dune_config_file/dune_config_file.ml#L256

It's the default of `--watch-exclusions`, which is passed when the watcher backend is initialized here:

https://github.com/ocaml/dune/blob/013ae8a776155cf0a3a68f109bab34712c3f5b4f/src/dune_file_watcher/dune_file_watcher.ml#L677

Then, each backend is able to postprocess events using that list.
As you can see, in the Fsevents case, neither `~should_exclude` nor `~watch_exclusions` is passed.

So I think that one way to fix that is to add an extra argument to `create_fsevents` and do the filtering in that function when an event arrives.

[...]

PR #9643, opened 2024-01-05
fix: use standard_watch_exclusions with fsevents

PoorlyDefinedBehaviour:
Closes: https://github.com/ocaml/dune/issues/9517

### DIFF
--- a/doc/changes/9643.md
+++ b/doc/changes/9643.md
@@ -1,1 +1,2 @@
-- Use watch exclusions in watch mode on MacOS (#9643, fixes #9517, @rgrinberg)
+- Use watch exclusions in watch mode on MacOS (#9643, fixes #9517,
+  @PoorlyDefinedBehaviour)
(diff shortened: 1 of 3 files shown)


rgrinberg:
Thanks for your contribution. We require commits to be signed with real names and not pseudonyms. See https://github.com/ocaml/dune/blob/main/CONTRIBUTING.md for more details

rgrinberg:
I did a couple of changes to this PR:

1. The test was moved to its own PR https://github.com/ocaml/dune/pull/9654
2. I modified the exclusion to exist directly in the fsevents callback rather than passing it all the way to the fsevents constructor. This is how we handle it on the inotify side already.

if these changes are OK with you, I can merge.

Thank you for your contribution!

PoorlyDefinedBehaviour:
> I did a couple of changes to this PR:
> 
> 1. The test was moved to its own PR [test: make sure .#name.ml patterns are ignored #9654](https://github.com/ocaml/dune/pull/9654)
> 2. I modified the exclusion to exist directly in the fsevents callback rather than passing it all the way to the fsevents constructor. This is how we handle it on the inotify side already.
> 
> if these changes are OK with you, I can merge.
> 
> Thank you for your contribution!

That's ok :+1:

PR #9669, opened 2024-01-08
fix(pkg): directory_entries bug

rgrinberg:
Whenever "foo" is a file, [directory_entires "foo"] would return itself.
That's not the correct behavior for things that are directories. The fix
is so to stop considering itself as a directory entry.

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -307,7 +307,10 @@ module At_rev = struct
        1. using libgit or ocamlgit
        2. possibly using [$ git archive] *)
     File.Set.filter files_at_rev ~f:(fun (file : File.t) ->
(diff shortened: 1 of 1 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -307,7 +307,8 @@ module At_rev = struct
        1. using libgit or ocamlgit
        2. possibly using [$ git archive] *)
     File.Set.filter files_at_rev ~f:(fun (file : File.t) ->
-      Path.Local.is_descendant file.path ~of_:path)
+      (not (Path.Local.equal file.path path))
```
Leonidas-from-XIV, line 310: Could you add a comment here why this check exists? Otherwise its hard to understand what it is for.

Leonidas-from-XIV (approved):

PR #9670, opened 2024-01-08
fix(pkg): improve locations for local path repos

rgrinberg:

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/invalid-opam-repo-errors.t/run.t
+++ b/test/blackbox-tests/test-cases/pkg/invalid-opam-repo-errors.t/run.t
@@ -14,7 +14,18 @@ Test the error cases for invalid opam repositories
   >  (source "file://$(pwd)/directory-that-does-not-exist"))
   > EOF
 
-  $ dune pkg lock
+  $ lock() {
+  > out="$(dune pkg lock 2>&1)"
+  > local code="$?"
+  > echo "$out" | sed 's/character.*:/characters X-X:/g' \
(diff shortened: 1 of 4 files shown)


test/blackbox-tests/test-cases/pkg/invalid-opam-repo-errors.t/run.t
```diff
@@ -14,11 +14,18 @@ Test the error cases for invalid opam repositories
   >  (source "file://$(pwd)/directory-that-does-not-exist"))
   > EOF
 
-  $ dune pkg lock
+  $ lock() {
+  > dune pkg lock 2>&1 | sed 's/character.*:/characters X-X:/g' \
+  >   | sed 's/source ".*"/source ../g' \
+  >   | grep -v "\^"
+  > }
```
Leonidas-from-XIV, line 21: This should return the error code of `dune pkg lock`, otherwise all calls of `lock` succeed (all the `[1]` disappear even when the locking failed).

I think that's only really possible to do using bash/zsh--isms in an elegant way (`PIPESTATUS`) but also possible with temporary files in POSIX.

Leonidas-from-XIV (commented):
Great change, only thing missing is `lock` returning the correct exit code.

test/blackbox-tests/test-cases/pkg/invalid-opam-repo-errors.t/run.t
rgrinberg: I added back the status codes manually

PR #9679, opened 2024-01-09
Add cond statement for chossing lockdir

gridbugs:


rgrinberg:
Please re-open if still relevant

PR #9680, opened 2024-01-09
Raise code error when git command fails

gridbugs:
This is intended to help developers faster understand what's happening when dune invokes git and git fails with an error.

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -100,21 +100,16 @@ let make_stderr () = Process.Io.make_stderr ~output_on_success:Swallow ~output_l
 (* to avoid Git translating its CLI *)
 let env = Env.add Env.initial ~var:"LC_ALL" ~value:"C"
 
-let git_code_error dir args exit_code maybe_output_dyn =
+let git_code_error ~dir ~args ~exit_code ~output =
   let git = Lazy.force Vcs.git in
-  let output_list =
-    match maybe_output_dyn with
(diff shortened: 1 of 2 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -100,26 +100,61 @@ let make_stderr () = Process.Io.make_stderr ~output_on_success:Swallow ~output_l
 (* to avoid Git translating its CLI *)
 let env = Env.add Env.initial ~var:"LC_ALL" ~value:"C"
 
+let handle_git_error git command dir fiber =
+  let stderr_path = Temp.create File ~prefix:"dune" ~suffix:"output" in
+  let stderr_to = Process.Io.file stderr_path Out in
+  let out =
+    Fiber.with_error_handler
```
rgrinberg, line 107: Doesn't this hide the original exception?

What about just changing `failure_mode` to just return the status and we can fail ourselves with a proper error.

gridbugs:
> Isn't the invocation already logged to `_build/log`? This is what I check when there's some problem with subprocesses.

I think for the sake of ux it should be a `Code_error` though. Currently the only output is the error message from git with no context.

gridbugs:
I rewrote this to set `failure_mode` to `Return`

rgrinberg:
Merging this after some simplifications.

Note that `==` is physical equality and while it works as you'd expect on integers, it's unlikely what you want. I removed it.

PR #9683, opened 2024-01-09
Remove `--skip-update` and make it implied when possible

### LINKED ISSUE #9354: pkg: remove --skip-update
After thinking about it some more, it might have been a mistake to add `--skip-update` at this point in time. Since we didn't have a user mandate for it, we could remove it to simplify our pkg implementation and make it easier to work with. In the future, if there turns out to be such a demand, we can always design something.

It should also be noted that the same behaviour is already possible using dune-workspace files.

Leonidas-from-XIV:
The only place where offline mode would work is if the repositoy URL didn't specify any hash or specified a hash that exists in the rev store.

It is easier to make the distinction on whether to update dependent on the `Source` than an CLI option (which if used wrong will just make dune fail).

I've fixed `of_opam_url` to update the repo if the revision was not found locally and is not a tag or branch, that way specifying a revision can work offline if the rev store is up to date but will transparently go online if it is out of date.

Closes #9354

### DIFF
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -174,7 +174,7 @@ let of_opam_repo_dir_path loc opam_repo_dir_path =
   { source = Directory opam_repo_dir_path; serializable = None; loc }
 ;;
 
-let of_git_repo ~update (source : Source.t) =
+let of_git_repo (source : Source.t) =
   let+ at_rev =
     let* remote =
       let* repo = Rev_store.get in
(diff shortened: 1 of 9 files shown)


src/dune_pkg/opam_repo.ml
```diff
@@ -58,25 +58,36 @@ module Source = struct
...
-                        "Make sure the URL is correct and the repository contains the \
-                         branch/tag"
-                    ]
-                  [ Pp.textf
-                      "Opam repository at '%s' does not have a reference '%s'"
-                      url
-                      ref
-                  ]))
+                (* we have to update the local repo and see if the commit exists *)
+                let* _ =
```
rgrinberg, line 71: It's important to annotate the types of anything that we're ignoring. Although I'm not sure how useful is `Remote.update` anymore.

rgrinberg (approved):

gridbugs (approved):

src/dune_pkg/opam_repo.ml
Leonidas-from-XIV: Good point.

I would argue it is still useful to have a way to `don't_update`. It could be merged into `add_repo`, but it feels like mixing responsibilities. Maybe a refactor that would change `add_repo` to imply working with a remote source instead of implying it will add a repository.

PR #9689, opened 2024-01-10
pkg: enforce opam-compatible package names

gridbugs:
This refactors package-management code to use
`Dune_lang.Package_name.Opam_compatible.t` for all package names. Previously we were checking for opam compatibility in some package management operations but not others leading to inconsistent behaviour. This removes the inconsistency by making it impossible to represent an opam incompatible package name in package management code.

Note that there's still an inconsistency in the treatment of package names in the rest of dune but we can't remove that currently as people may depend on it. There's a discussion of this problem at: https://github.com/ocaml/dune/issues/9371


gridbugs:
fixes https://github.com/ocaml/dune/issues/9371

src/dune_pkg/package_name.mli
```diff
@@ -1,8 +1,11 @@
 open! Stdune
```
Leonidas-from-XIV, line 1: Since this is getting used in the rest of the source code, maybe it would make sense to rename this to `Opam_package_name`, because reading the diff a lot of the changes are replacing `Package.Name` with `Package_name` which seems like a rename but it actually has different semantics as it requires Opam-compatible names.

src/dune_util/dune_util.ml
```diff
@@ -3,6 +3,7 @@ module Log = Log
 module Persistent = Persistent
 module Report_error = Report_error
 module Stringlike = Stringlike
+module Located_stringlike = Located_stringlike
```
Leonidas-from-XIV, line 6: This seems like a useful addition! :+1:

src/dune_util/located_stringlike.mli
```diff
@@ -0,0 +1,6 @@
+open! Stdune
+
+(** Implement [Stringlike_intf.S] with a value accompanied by a [Loc.t] for
+    string that carry around their own source location for easier error
+    reporting. *)
+module Make (S : Stringlike_intf.S_base) : Stringlike_intf.S with type t = Loc.t * S.t
```
rgrinberg, line 6: We've had something like this in the past, but it turned out to be problematic. The issue is that it makes it too easy not to think about locations whens doing comparisons. When we do `Foo.equal` we should really think about whether we need to take the location into account when comparing identifiers. For that reason we don't include the locations in such type definitions.

gridbugs: What if we changed the functor to take an additional argument specifying whether comparisons should include the loc or not so you're forced to consciously make that decision when instantiating it?

Alternatively we could just include the loc in `Package_name.t` and treat it as a special (and hopefully temporary) case.

rgrinberg: Isn't it easy enough to just pass `Loc.t * Package_name.t` wherever you need the location? That's what we do everywhere else (library names, repository names, etc.)

gridbugs: The `Loc.t` is needed whenever we convert from a `Package_name.t` to a `Package_name.Opam_compatible.t` and I figured it would be a big refactor to add `Loc.t` to all the relevant places.

rgrinberg: Tbh, it doesn't seem so bad - I've done it just about everywhere else without paying attention to it. In any case, I'm fine with this code as well. Although CI says it's not working yet.

rgrinberg: Oh btw, could you move this module closer to where it's being used. `Dune_util` is being used by the engine so ideally we wouldn't anything that isn't needed there to avoid syncing issues. I realize it already has some stuff that probably shouldn't live there, but we should avoid adding more.

gridbugs: I'll do an experiment to see how bad the refactoring would actually be and if it turns out fine then I'll just delete this module. Otherwise yep I'll move it to dune_lang.

src/dune_pkg/package_name.mli
gridbugs: Renamed to `Opam_compatible_package_name`.

src/dune_util/located_stringlike.mli
gridbugs: It would be a big refactor to add `Loc.t` to all the places where it's needed so I've instead moved `Located_stringlike` into the `dune_lang` library.

rgrinberg (approved):

rgrinberg:
Could you also make a ticket to get rid of all this stuff for 4.0? This is definitely a breaking change we want to make.

rgrinberg (changes requested):
After Looking at the code again, my opinion shifted against this PR.

From what I can tell, the only benefit of this code is that we no longer allow incorrect package names in the package management code.

But the package management code is unstable anyway, and by the time it will stabilize, we'll be ready to reconsider 4.0 anyway. So we really aren't going to gain much from this extra validation during this period. Moreover, it's such a huge diff. Doesn't seem to be worth it.

gridbugs:
Fair enough. Closing this, and I've made an issue to unify the package name types for 4.0: https://github.com/ocaml/dune/issues/9731

PR #9696, opened 2024-01-10
refactor(engine): check config once in maybe_async

emillon:
This moves the `maybe_async` functions to their own module and extract a
pattern where:

- the config key is only checked once
- the resulting function is polymorphic


rgrinberg:
The changes look OK to me. I added @snowleopard as a reviewer since it touches the engine

emillon:
Thanks. Also, I'm not completely satisfied with the names so I'm happy to take suggestions.

snowleopard:
> The changes look OK to me. I added @snowleopard as a reviewer since it touches the engine

@rgrinberg Could you import this change so we can benchmark it and possibly tweak some things given that we've diverged from upstream quite a bit?

PR #9697, opened 2024-01-10
refactor: move Artifacts.Objs to own module

rgrinberg:
cc @nojb. This was interfering with some PR's that I have coming so I moved it to its own module


nojb:
> cc @nojb. This was interfering with some PR's that I have coming so I moved it to its own module

Looks good! Thanks.

PR #9698, opened 2024-01-10
refactor(pkg): make remote reading a little more resilient

rgrinberg:
Previously, the algorithm of looking up inside a remote would treat
everything after [remote $name] as belonging to the remote. That's not
correct because we can have other remotes after what we found.

This PR makes sure to first get the section that belongs to a remote and
limits the search to this section.

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -449,25 +449,49 @@ let branch_of_refspec refspec =
     String.drop_prefix remote_ref ~prefix:"refs/heads/")
 ;;
 
+let find_section =
+  let re =
+    Re.seq
+      [ Re.(rep space)
+      ; Re.str "[remote "
+      ; Re.char '"'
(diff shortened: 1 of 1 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -449,25 +449,50 @@ let branch_of_refspec refspec =
...
+let find_section =
+  let re =
+    Re.seq
+      [ Re.(rep space)
+      ; Re.str "[remote"
+      ; Re.(rep space)
+      ; Re.char '"'
+      ; Re.group (Re.rep (Re.diff Re.any (Re.char '"')))
+      ; Re.char '"'
+      ; Re.(rep space)
```
Leonidas-from-XIV, line 461: I checked and git doesn't support spaces between the closing double quote and the `]`.

Leonidas-from-XIV (approved):

PR #9699, opened 2024-01-10
refactor: remove some optional arguments

rgrinberg:
They provide no benefit here


src/dune_rules/dune_project.mli
```diff
@@ -123,12 +123,7 @@ val load
```
Alizter, line 123: This comment needs to be updated.

PR #9702, opened 2024-01-11
feature: enable directory targets for everyone

rgrinberg:
The [(using directory-targets 0.1)] extension has been removed and
starting from 3.13, directory targets are fully supported.


snowleopard:
Actually, I wonder if we should wait a bit. We are doing some implementation/semantic clean up internally (cc @rleshchinskiy) and might change the behaviour in some edge cases. It would be best to upstream those changes while this feature still has an experimental status.

Specifically, we'd like to do the following:

* Forbid requesting subdirectories of a directory target from command line (the current implementation has some weird behaviour in those cases). This allows simplifying `build_system.ml` quite a bit.

 * Make it possible to declare dependencies on whole directory targets (instead of having to use globs). This seems like an important feature to have to make directory targets first class.

emillon:
Yes, waiting sounds good. We can always check back for 3.14. This will avoid supporting the old version forever when the semantics change.

bin/describe/describe_format.ml
```diff
@@ -23,7 +23,7 @@ let print_as_sexp dyn =
     |> Dune_lang.Ast.add_loc ~loc:Loc.none
     |> Dune_lang.Cst.concrete
   in
-  let version = Dune_lang.Syntax.greatest_supported_version Stanza.syntax in
+  let version = Dune_lang.Syntax.greatest_supported_version_exn Stanza.syntax in
```
emillon, line 26: Can you move that change to its PR?

doc/stanzas/rule.rst
```diff
@@ -128,19 +128,14 @@ Note that in Dune, targets must always be known statically. For instance, this
 Directory targets
 ~~~~~~~~~~~~~~~~~
 
-Note that at this time, Dune officially only supports user rules with targets in
-the current directory. However, starting from Dune 3.0, we provide an
-experimental support for *directory targets*, where an action can produce a
-whole tree of build artifacts. To specify a directory target, you can use the
-``(dir <dirname>)`` syntax. For example, the following stanza describes a rule
-with a file target ``foo`` and a directory target ``bar``.
+Starting from Dune 3.13, we provide an support for *directory targets*, where
```
emillon, line 131: You can use `.. versionadded:: 3.13` to move the history out of the prose.

src/dune_rules/stanzas/rule_conf.ml
```diff
@@ -82,27 +82,24 @@ let short_form =
...
 
-let directory_targets_extension =
+let _directory_targets_extension =
   let syntax =
     Dune_lang.Syntax.create
       ~name:"directory-targets"
+      ~experimental:true
       ~desc:"experimental support for directory targets"
-      [ (0, 1), `Since (3, 0) ]
+      [ (0, 1), `Deleted_in (3, 13) ]
```
emillon, line 91: Can you add a test of a project that refers to the deleted extension?

test/blackbox-tests/test-cases/start-install-dst-with-parent-error.t
```diff
@@ -200,6 +198,33 @@ Test that on older versions of dune we don't get warnings in this case:
   >  (dirs (bar as ..)))
   > EOF
   $ dune build foo.install && cat _build/default/foo.install
+  File "dune", line 15, characters 15-17:
```
emillon, line 201: Is this warning appearing because of the bump to 3.13?

doc/stanzas/rule.rst
hhugo, line 131:
```suggestion
Starting from Dune 3.13, we provide support for *directory targets*, where
```

rgrinberg:
Holding off sounds fine to me

> Make it possible to declare dependencies on whole directory targets (instead of having to use globs). This seems like an important feature to have to make directory targets first class.

Isn't this already possible?

Issue #9678, opened 2024-01-09
Pointing a package source to a directory doesn't work


rgrinberg:
Having a repo with a package's source set to a local dir:

```
url {
  src: "file://foo/bar"
}
```

doesn't work when `foo/bar` is a directory. What ends up happening is that we try to compute a checksum for a file and fail.

There's no need to compute this checksum.

cc @gridbugs

PR #9703, opened 2024-01-11
pkg: don't generate checksums for local sources

### LINKED ISSUE #9677: Do not include checksums for "local" sources
Currently, we try to compute checksums for sources that exist on the local FS when computing the lock directory. E.g. if we have a package with:

```
url {
  src: "file://foo/bar"
}
```

We will compute and include the checksum of `foo/bar`. There's no need for checksums to exist in such a case because we are using such a source for local development and it's expected to change. Also, the checksums doesn't help with reproducibility since the source is private to us anyway.

Similar deal for:

```
url {
  src: "git+file://foo/mygitrepo"
}
```

There's no need for a checksum here. Although currently we don't support checksums for VCS sources at all.

cc @gridbugs

gridbugs:
fixes https://github.com/ocaml/dune/issues/9677

### DIFF
--- a/src/dune_pkg/opamUrl0.ml
+++ b/src/dune_pkg/opamUrl0.ml
@@ -18,3 +18,5 @@ let is_version_control t =
   | #version_control -> true
   | _ -> false
 ;;
+
+let is_local t = String.equal t.transport "file"
--- a/test/blackbox-tests/test-cases/pkg/compute-checksums-when-missing.t
+++ b/test/blackbox-tests/test-cases/pkg/compute-checksums-when-missing.t
@@ -1,64 +1,91 @@
 Test that dune will add checksums to lockfiles when the package has a source
-archive but no checksum.
+archive but no checksum. This test uses an http server to serve packages to
+test checksum generation, since we only generate checksums for packages
+downloaded from non-local sources.
   $ . ./helpers.sh
   $ mkrepo
 
-  $ strip_pwd() {
-  >   sed -e "s#$PWD#<pwd>#"
+  $ strip_transient() {
+  >   sed -e "s#$PWD#<pwd>#" | \
+  >   # strip the entire address as the port number may also appear in the hash
+  >   sed -e "s#http://.*$PORT#<addr>#"
   > }
 
+A file that will comprise the package source:
   $ echo "Hello, World!" > foo.txt
+
+Run the server in the background:
+  $ webserver_oneshot --content-file foo.txt --port-file port.txt &
+  $ until test -f port.txt; do sleep 0.1; done
+  $ PORT=$(cat port.txt)
 
   $ mkpkg foo <<EOF
   > url {
-  >  src: "$PWD/foo.txt"
+  >  src: "http://0.0.0.0:$PORT"
   > }
   > EOF
 
-  $ solve foo | strip_pwd
+  $ solve foo | strip_transient
   Solution for dune.lock:
   - foo.0.0.1
   Package "foo" has source archive which lacks a checksum.
-  The source archive will be downloaded from:
-  file://<pwd>/foo.txt
+  The source archive will be downloaded from: <addr>
   Dune will compute its own checksum for this source archive.
 
 Replace the path in the lockfile as it would otherwise include the sandbox
 path.
-  $ cat dune.lock/foo.pkg | strip_pwd
+  $ cat dune.lock/foo.pkg | strip_transient
   (version 0.0.1)
   
   (source
    (fetch
-    (url
-     file://<pwd>/foo.txt)
+    (url <addr>)
     (checksum md5=bea8252ff4e80f41719ea13cdf007273)))
   
   (dev)
 
 Now make sure we can gracefully handle the case when the archive is missing.
 
-  $ rm foo.txt
-  $ solve foo 2>&1 | strip_pwd
+  $ rm port.txt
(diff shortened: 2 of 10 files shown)


src/dune_pkg/opamUrl0.ml
```diff
@@ -18,3 +18,5 @@ let is_version_control t =
   | #version_control -> true
   | _ -> false
 ;;
+
+let is_file t = String.equal t.transport "file"
```
rgrinberg, line 22: Will this work for `git+file` as well?

test/blackbox-tests/test-cases/pkg/compute-checksums-when-missing.t
```diff
@@ -1,87 +1,131 @@
...
+  $ PORT=$(shuf -i 10000-65535 -n 1)
+
+  $ strip_transient() {
+  >   sed -e "s#$PWD#<pwd>#" | \
+  >   # strip the entire address as the port number may also appear in the hash
+  >   sed -e "s#http://.*$PORT#<addr>#"
   > }
 
+Make a webserver that serves a single file once before terminating:
+  $ cat >oneshot_webserver.ml <<EOF
```
rgrinberg, line 19: How about turning this into a proper binary and using it as follows:

```
$ serve_once --file FILE --port-file FILE
```

This will serve `--file` and then terminate. The program will select a port randomly and write the selection `--port-file`.

rgrinberg: Oh and this way we won't need to depend on ocamlfind for this test

gridbugs:
I just added a new test that we can use local source archives and directories as package sources which fixes https://github.com/ocaml/dune/issues/9678.

src/dune_pkg/opamUrl0.ml
gridbugs: Maybe eventually. Do we currently have any support for using a git repo as a package source (other than treating it as a directory)? Do we plan to support that as part of the pkg mvp?

rgrinberg: I suppose it's needed to pin packages to local git repos. Sounds pretty useful, but I wouldn't be too worried if it wasn't available in the MVP.

test/blackbox-tests/test-cases/pkg/compute-checksums-when-missing.t
gridbugs: I moved the server to test/blackbox-tests/utils

rgrinberg (approved):

test/blackbox-tests/utils/webserver_oneshot.ml
```diff
@@ -0,0 +1,61 @@
...
+  Unix.setsockopt sock Unix.SO_REUSEADDR true;
+  Unix.bind sock addr;
+  Unix.listen sock 1;
+  let port =
+    match Unix.getsockname sock with
+    | Unix.ADDR_INET (_, port) -> port
+    | ADDR_UNIX _ -> failwith "unreachable"
+  in
+  Out_channel.with_open_text port_file (fun out_channel ->
+    Out_channel.output_string out_channel (Printf.sprintf "%d" port));
```
Leonidas-from-XIV, line 47: Probably would make sense to add a final newline here, this way it doesn't mess up terminals when `cat`-ed and capturing it with `$()` should still work.

Leonidas-from-XIV (commented):
Looks good, I've added some minor comments to make it faster and clearer I hope.

test/blackbox-tests/test-cases/pkg/compute-checksums-when-missing.t
```diff
@@ -1,87 +1,98 @@
...
+  >   sed -e "s#http://.*$PORT#<addr>#"
   > }
 
+A file that will comprise the package source:
   $ echo "Hello, World!" > foo.txt
 
+Run the server in the background:
+  $ webserver_oneshot --content-file foo.txt --port-file port.txt &
+  $ until test -f port.txt; do sleep 1; done
+  $ PORT=$(cat port.txt)
```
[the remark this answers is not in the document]
gridbugs: Writing the port to a specified file lets us keep stdout free for debugging messages if needed and it also lets us use the existence of the port file as a convenient way to detect when the server has started.

src/dune_pkg/opamUrl0.ml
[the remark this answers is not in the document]
gridbugs: renamed to `is_local`

test/blackbox-tests/test-cases/pkg/dune
```diff
@@ -31,3 +31,7 @@
 (cram
  (deps %{bin:which})
  (applies_to pkg-deps))
+
+(cram
+ (deps %{bin:webserver_oneshot})
+ (applies_to :whole_subtree))
```
rgrinberg, line 37: how about just making it a dependency of the test that uses it?

test/blackbox-tests/test-cases/pkg/compute-checksums-when-missing.t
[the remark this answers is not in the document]
Leonidas-from-XIV: Yes, you need the file but debugging messages can be printed to stderr and the port to stdout.

PR #9705, opened 2024-01-11
Add `dune promotion list` to display the current changes files.

moyodiallo:
It is a starting point of `dune promotion list`. One of the concern would be the print format.
An example of Dyn format printed directly.
```
{ src = In_build_dir "default/bin/.formatted/main.ml"
; staging = None
; dst = In_source_tree "bin/main.ml"
}
{ src = In_build_dir "default/src/dune_patch/.formatted/dune_patch.ml"
; staging = None
; dst = In_source_tree "src/dune_patch/dune_patch.ml"
}
```

### DIFF
--- a/src/dune_engine/diff_promotion.mli
+++ b/src/dune_engine/diff_promotion.mli
@@ -18,6 +18,8 @@ module File : sig
 
   val to_dyn : t -> Dyn.t
   val in_staging_area : Path.Source.t -> Path.Build.t
+  val compare : t -> t -> ordering
+  val source : t -> Path.Source.t
 
   (** Register an intermediate file to promote. The build path may point to the
       sandbox and the file will be moved to the staging area. *)
@@ -43,6 +45,8 @@ type files_to_promote =
   | All
   | These of Path.Source.t list * (Path.Source.t -> unit)
 
+val load_db : unit -> File.t list
+val filter_db : files_to_promote -> File.t list -> File.t list
+val diff_for_file : File.t -> (Print_diff.Diff.t, User_message.t) result Fiber.t
 val promote_files_registered_in_last_run : files_to_promote -> unit
 val display : files_to_promote -> unit Fiber.t
-val display_files : files_to_promote -> unit Fiber.t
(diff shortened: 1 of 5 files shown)


rgrinberg:
Could you add some tests to the new command?

moyodiallo:
Added in https://github.com/ocaml/dune/commit/f2857317b5d58f5ae1e4ce00d792e6203aeed9e0.

What about the print format ? Do we need a sort of harmonization like `dune describe` format(sexp).

rgrinberg:
I would just print the filename in the source directory to start. AFAIK, this is mostly for end users to get an idea for what promotions they can promote. No need to add anything more advanced for now.

moyodiallo:
Fair enough, done in https://github.com/ocaml/dune/pull/9705/commits/239ad083a201487ffcbeeb2da80445ab2b1be9cf.

rgrinberg (changes requested):
The implementation is fine, but we need the implementation in bin/promotion.ml. I realize that will require exposing some functions in `diff_promotion.mli`, but that should be preferable to adding more CLI code to the engine.

src/dune_engine/diff_promotion.mli
```diff
@@ -45,3 +45,4 @@ type files_to_promote =
 
 val promote_files_registered_in_last_run : files_to_promote -> unit
 val display : files_to_promote -> unit Fiber.t
```
rgrinberg, line 47: @snowleopard is `Diff_promotion.display` used internally? If not, I can move it to the CLI.

moyodiallo:
> The implementation is fine, but we need the implementation in bin/promotion.ml. I realize that will require exposing some functions in `diff_promotion.mli`, but that should be preferable to adding more CLI code to the engine.

I agree done in https://github.com/ocaml/dune/pull/9705/commits/c3032a1bf1f84550882909acfda1d01712c35102.

src/dune_engine/diff_promotion.mli
snowleopard: I think it's unused, so, yes let's move it. Thanks!

rgrinberg: Done in https://github.com/ocaml/dune/pull/9931

rgrinberg:
Thanks

PR #9706, opened 2024-01-11
refactor: set directory targets as experimental

rgrinberg:

### DIFF
--- a/src/dune_rules/stanzas/rule_conf.ml
+++ b/src/dune_rules/stanzas/rule_conf.ml
@@ -87,6 +87,7 @@ let directory_targets_extension =
     Dune_lang.Syntax.create
       ~name:"directory-targets"
       ~desc:"experimental support for directory targets"
+      ~experimental:true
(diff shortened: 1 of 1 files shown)


src/dune_rules/stanzas/rule_conf.ml
```diff
@@ -87,6 +87,7 @@ let directory_targets_extension =
     Dune_lang.Syntax.create
       ~name:"directory-targets"
       ~desc:"experimental support for directory targets"
+      ~experimental:true
```
emillon, line 90: What are the semantics changed? I assumed that 0.x and experimental were equivalent.

rgrinberg: 0.x implies experimental, but the the converse doesn't hold. Experimental versions can start from 1.x. 

In this case, we just added the experimental flag sometime later and forgot to update everywhere where it should be used.

emillon (approved):

PR #9707, opened 2024-01-11
fix: optional binaries handling

rgrinberg:
We wait until the binaries are resolved to determine if they are optional or not. This allows things to be a little more lazy as we avoid evaluating whether a binary is available until it's used.

Moreover, the handling of `enabled_if` on install sections is no longer ignored when generating the artifacts map


rgrinberg:
@gridbugs you think you could try reviewing this one?

gridbugs (approved):

PR #9708, opened 2024-01-11
fix: use original paths in (run ..)

rgrinberg:
Starting from 3.13, we resolve [foo] in [(run foo ..)] to where the
binary is defined rather than in the _build/install/default directory.

### DIFF
--- a/src/dune_rules/action_unexpanded.ml
+++ b/src/dune_rules/action_unexpanded.ml
@@ -397,7 +397,15 @@ end = struct
                Action_builder.return (Ok (Path.relative dir s))
              | In_path ->
                Action_builder.of_memo
-                 (Artifacts.binary ~loc:(Some loc) (Expander.artifacts env.expander) s))
+               @@
+               let open Memo.O in
+               let* where =
+                 let+ project = Dune_load.find_project ~dir:env.dir in
+                 if Dune_project.dune_version project >= (3, 14)
+                 then Artifacts.Original_path
+                 else Install_dir
+               in
+               Artifacts.binary ~loc:(Some loc) ~where (Expander.artifacts env.expander) s)
         in
         let prog = Result.map prog ~f:(Expander.map_exe env.expander) in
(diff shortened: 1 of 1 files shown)


rgrinberg:
@ejgallego this change is indeed versioned since we're affecting user defined rules

rgrinberg:
ping @ejgallego do you want to review this one?

ejgallego:
Is there a way to assess the breakage this change would produce once the users migrate to `(lang dune 3.13)` ?

rgrinberg:
I don't think there is. Although it seems very unlikely that this should cause problems. I'm really only versioning this behavior because it can technically be observed by users looking at `Sys.argv.(0)` - which doesn't seem like a good thing to do in general.

ejgallego:
> I don't think there is. Although it seems very unlikely that this should cause problems. I'm really only versioning this behavior because it can technically be observed by users looking at `Sys.argv.(0)` - which doesn't seem like a good thing to do in general.

Indeed my fear is that a few packages may look into their `argv[0]` or directory the binary lives in, so I dunno what the default behavior should be here. Can in some cases the `install` layout be preferred by the users?

I am unsure what to say more. Documentation states "resolved locally", so indeed, this was never specified.

Maybe this would be worth documenting so users know what to expect of `(run ...)` ?

rgrinberg:
The potential breakage aside, the change itself is rather desirable. If there are issues down the line, we can add an option to disable this behavior.

ejgallego:
> I actually prefer to leave it vague. This way we don't have to commit to one or the other.

I think it is fine not to commit to a particular implementation, but maybe we should document the semantics of run better, as what happens in `run` is been a frequent source of user confusion.

Indeed, many programs are just not designed to work with sandboxing, as they do depend on a lot of data files, and try to search for them. So this tends to lead to non-reproducible workings as we don't provide an easy way to control the layout before run, and the documentation is IMHO not easy to access to users.

I'm not sure how `run` interacts with sites actually.

PR #9712, opened 2024-01-11
refactor: optimize [Fpath.is_root]

rgrinberg:
On unix, the old implementation of Filename.dirname was exactly
equivalent to x = "." || x = "/", but was implemented in a way that
allocated a string only to throw it away every time.

On windows, we leave the old implementation as I'm unable to verify the
above.

### DIFF
--- a/otherlibs/stdune/src/fpath.ml
+++ b/otherlibs/stdune/src/fpath.ml
@@ -1,4 +1,12 @@
-let is_root t = Filename.dirname t = t
+let is_root =
+  if Sys.unix
+  then fun x -> x = "/" || x = "."
+  else
+    (* CR-someday rgrinberg: can we do better on windows? *)
+    fun s ->
+    Filename.dirname s = s
+;;
+
 let initial_cwd = Stdlib.Sys.getcwd ()
 
 type mkdir_result =


snowleopard:
Thanks! Please post benchmark results here.

rgrinberg:
As usual, the benchmarking job is acting up just when you need it :).

Here's my own measurements:

Doing a zero build on dune with this PR is 1.3642s (averaged over 5 runs). Without this PR, it's 1.4036. For a relative decrease of 2.8%. When the benchmarking job was working, the decrease was more modest (around 1%), with a corresponding 1% decrease to promoted words.

I can add micro benchmarks to this function if this isn't convincing enough.

rgrinberg:
Here's the results of the micro benchmarks:

```
┌────────────────────────────────────────────────────────────────────────────┬──────────┬─────────┬────────────┐
│ Name                                                                       │ Time/Run │ mWd/Run │ Percentage │
├────────────────────────────────────────────────────────────────────────────┼──────────┼─────────┼────────────┤
│ [bench/micro/path_bench.ml] is_root:root                                   │   8.80ns │         │      1.82% │
│ [bench/micro/path_bench.ml] is_root:short path                             │   8.19ns │         │      1.70% │
│ [bench/micro/path_bench.ml] is_root:long path                              │   8.20ns │         │      1.70% │
```

So it's between 2-7x faster

snowleopard:
Nice, thanks for the benchmarks! I'm curious to see if we notice this in our internal benchmarks, so importing this change now.

otherlibs/stdune/src/fpath.ml
```diff
@@ -1,4 +1,10 @@
-let is_root t = Filename.dirname t = t
+let is_root =
+  if Sys.unix
+  then fun x -> x = "/" || x = "."
+  else (* CR-rgrinberg: can we do better on windows? *)
```
snowleopard, line 4: (In case you were following our CR convention.)
```suggestion
  else (* CR-someday rgrinberg: can we do better on windows? *)
```

otherlibs/stdune/src/fpath.ml
```diff
@@ -1,4 +1,10 @@
-let is_root t = Filename.dirname t = t
+let is_root =
+  if Sys.unix
```
snowleopard, line 2: I wonder: don't we have something better for OS deterction in Stdune?

snowleopard: Oh, yes, it seems we do: `Platform.OS.value`.

snowleopard: Why? It seems nicer, especially given that some future implementations might only, say, work on Windows but not on other platforms.

I've started internal benchmarks with the following version:

```
let is_root =
  match Platform.OS.value with
  | Linux | Darwin -> fun x -> x = "/" || x = "."
  | Windows | Other ->
    (* CR-someday rgrinberg: Can we do better on Windows? *)
    fun s -> Filename.dirname s = s
;;
```

rgrinberg: `Other` are almost always Unix-like so they should be included for the optimization. This is why `Sys.unix` is a bit more precise, it's true for every OS with `/` at the root.

snowleopard: Hmm, I see. That makes sense though I don't like mixing different styles (i.e. sometimes using `Sys`, sometimes using `Platform`). Maybe `Platform` should expose `slash_is_root : bool` or something like that.

rgrinberg: The rest of the module is already using Sys so perhaps we should just it all at once to use Platform (along with introducing this function)

snowleopard: OK, for now, I've switched to `Sys.unix` internally as well, but yes, it seems useful to do a separate clean up in future.

snowleopard:
Didn't the old implementation return `true` for the empty string too?

snowleopard:
I feel like maybe be we want some expect tests actually, which test a few corner cases?

rgrinberg:
> Didn't the old implementation return `true` for the empty string too?

It did not. `Filename.dirname "" = "."` which isn't equal to itself. You might be thinking `Path.Local.is_root` and its variants. We could make them the same, but my implementation preserves the previous semantics.

>  I feel like maybe be we want some expect tests actually, which test a few corner cases?

Added some tests here https://github.com/ocaml/dune/pull/10089

snowleopard:
Thanks for adding the test! Sorry, I confused myself when staring at the implementation of `Filename.dirname`.

snowleopard (approved):
Thanks, looks good! I didn't see much in our internal benchmarks. There seems to be some slight improvement in build times around 0.2% but it's well below the usual noise margin. Anyway, it seems fine to merge if only because it actually makes the implementation much clearer.

rgrinberg:
That's a bit disappointing, but at least this new implementation is as simple as the old one.

PR #9713, opened 2024-01-11
refactor: optimize some comparisons in some hot spots

rgrinberg:
@snowleopard I don't think it will make much of a difference internally since you have your own frontend for sexps. The benchmarks do show that it shaves off some allocations for us


rgrinberg:
This is only used in the public dune

PR #9718, opened 2024-01-12
pkg: simple end to end test of package management

gridbugs:
Adds a test that solves dependencies to make a lockdir and then builds a project that depends on a library in its lockdir.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/e2e.t
+++ b/test/blackbox-tests/test-cases/pkg/e2e.t
@@ -1,0 +1,80 @@
+Exercises end to end locking and building a simple project.
+
+  $ . ./helpers.sh
+  $ mkrepo
+  $ add_mock_repo_if_needed
+
+Make a library:
+  $ mkdir foo
+  $ cd foo
+  $ cat > dune-project <<EOF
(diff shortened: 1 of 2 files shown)


test/blackbox-tests/test-cases/pkg/e2e.t
```diff
@@ -0,0 +1,72 @@
...
+  >     name
+  >     "-j"
+  >     jobs
+  >     "@install"
+  >     "@runtest" {with-test}
+  >     "@doc" {with-doc}
+  >   ]
+  > ]
+  > url {
+  >  src: "file://$PWD/foo.tar.gz"
```
Leonidas-from-XIV, line 42: Can you add the checksum as well to make sure the validation works?

Leonidas-from-XIV (approved):
Very useful, good to see a test that makes sure packages can depend on dune-built dependencies!

test/blackbox-tests/test-cases/pkg/e2e.t
gridbugs: done

rgrinberg (approved):

PR #9720, opened 2024-01-12
pkg: support write-file in expander

emillon:


rgrinberg (approved):
Seems fine. Note that the goal is to unify this with regular actions. We don't want to maintain two expansion mechanisms.

gridbugs (approved):

emillon:
Thanks!

PR #9735, opened 2024-01-15
boot: remove single-command bootstrap

### LINKED ISSUE #9507: bootstrap is not reproducible wrt -j value

emillon:
Fixes #9507

The bootstrap process can use 2 strategies:
- parallel: run compile commands in parallel and link the rest
- single-command: run ocamlopt with a long list of arguments

The single-command strategy is used if win32 or if `-j 1` is set (implicitly or explicitly).

One problem is that: parallel and single-command create different binaries (#9507).

The assumption was that single-command would be faster than `-j 1` on Linux and any parallel build on Windows. However, from a quick benchmark the assumption is true on Linux but false on Windows.

The tiny savings in the case of Linux where only a single core is available are not enough to justify the extra code path and reproducibility gotcha.

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -1141,27 +1141,6 @@ let build
        ])
 ;;
 
-let build_with_single_command
-  ~ocaml_config:_
-  ~dependencies
-  ~c_files
-  ~link_flags
-  { target = name, main; external_libraries; _ }
-  =
-  let external_libraries, external_includes = resolve_externals external_libraries in
-  write_args "mods_list" (sort_files dependencies ~main);
-  Process.run
-    ~cwd:build_dir
-    Config.compiler
-    (List.concat
-       [ common_build_args name ~external_includes ~external_libraries
-       ; [ "-no-alias-deps"; "-w"; "-49-6" ]
-       ; c_files
-       ; [ "-args"; "mods_list" ] @ link_flags
-       ; allow_unstable_sources
-       ])
-;;
-
 let rec rm_rf fn =
   match Unix.lstat fn with
   | { st_kind = S_DIR; _ } ->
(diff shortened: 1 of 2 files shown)


emillon:
As discussed in last dev meeting.

rgrinberg (approved):
Worth a CHANGES entry

PR #9742, opened 2024-01-15
test(pkg): wait for jobs to finish

rgrinberg:
When running our test webserver, we should wait make sure that it
terminates after we're doing using it. The reasons are twofold:

1. Make sure that we indeed downloaded something form it and it
   terminated
2. To make sure it doensn't leak outside the current test if it doesn't
   terminate.


gridbugs (approved):
TIL about `wait`!

PR #9749, opened 2024-01-16
test(pkg): share http code

rgrinberg:
The unit tests and cram tests both have their own http server that
duplicates some code. This PR introduces an HTTP module to start sharing
some of it.

### DIFF
--- a/test/http/http.ml
+++ b/test/http/http.ml
@@ -1,0 +1,52 @@
+open Stdune
+
+module Server = struct
+  type t =
+    { sock : Unix.file_descr
+    ; addr : Unix.sockaddr
+    }
+
+  type session = out_channel
+
+  let make addr =
+    let sock = Unix.socket ~cloexec:true Unix.PF_INET Unix.SOCK_STREAM 0 in
+    { sock; addr }
+  ;;
+
+  let port t =
+    match Unix.getsockname t.sock with
+    | Unix.ADDR_INET (_, port) -> port
+    | ADDR_UNIX _ -> failwith "no port defined"
+  ;;
+
+  let start t =
+    Unix.setsockopt t.sock Unix.SO_REUSEADDR true;
+    Unix.bind t.sock t.addr;
+    Unix.listen t.sock 1
+  ;;
+
+  let accept t ~f =
+    let descr, _sockaddr = Unix.accept ~cloexec:true t.sock in
+    let out = Unix.out_channel_of_descr descr in
+    f out;
+    close_out out
+  ;;
+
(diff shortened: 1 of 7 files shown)


test/http/http.ml
```diff
@@ -0,0 +1,48 @@
...
+    { sock; addr }
+  ;;
+
+  let port t =
+    match Unix.getsockname t.sock with
+    | Unix.ADDR_INET (_, port) -> port
+    | ADDR_UNIX _ -> failwith "no port defined"
+  ;;
+
+  let serve_once t =
```
gridbugs, line 20: Is this function actually limited to serving only once? I thought the reason why the server only responds to a single request is that it doesn't call `Unix.accept` in a loop.

test/http/http.mli
```diff
@@ -0,0 +1,15 @@
+(** Http helpers used for tests *)
+
+module Server : sig
+  (* Dummy http server that accepts a single request *)
```
gridbugs, line 4: It's really up to the client whether the server accepts a single request. Since this module exposes `accept` then a client may implement a more sophisticated server by calling `accept` multiple times.

rgrinberg: Revised.

test/http/http.ml
gridbugs: That's confusing because it doesn't actually call `Unix.accept` and this module already had a function `accept` which does call `Unix.accept`.  This function sets up the socket but doesn't wait for a request (`Unix.accept` waits for a request).

rgrinberg: Indeed I renamed the old `accept` to be `start`.

test/http/http.mli
```diff
@@ -0,0 +1,15 @@
...
+
+  type t
+
+  val make : Unix.sockaddr -> t
+  val accept : t -> out_channel
+  val stop : t -> unit
+  val port : t -> int
+  val start : t -> unit
+  val respond : out_channel -> status:[ `Ok | `Not_found ] -> content:string -> unit
+  val respond_file : out_channel -> file:string -> unit
```
gridbugs, line 14: It's not a huge deal but I think this would be clearer if `accept` and `respond*` were combined. This way a client would call a function which waits for a request and is passed a callback whose job it is to reply to the request. The problem with the current api is that a client might attempt to write arbitrary data to the `out_channel` returned by `accept` and they might forget to call one of `respond` or `respond_file` on it or they might call both, all of which would result in erroneous behaviour.

rgrinberg: I changed `accept` to take a callback.

gridbugs (approved):

PR #9750, opened 2024-01-16
pkg: add cond statement for choosing lockdir

gridbugs:
We expect projects with system-specific packages to include multiple lockdirs - one for each system. This change adds a dsl for selecting the appropriate lockdir for the current system.

### DIFF
--- a/src/dune_pkg/sys_poll.ml
+++ b/src/dune_pkg/sys_poll.ml
@@ -48,33 +48,51 @@ let run_capture_line ~path ~prog ~args =
   | Some prog -> Process.run_capture_line ~display:Quiet Strict prog args >>| norm
 ;;
 
+module Config_override_variables = struct
+  let string_option_config name =
+    let config =
+      Dune_config.Config.make ~name ~of_string:(fun s -> Ok (Some s)) ~default:None
+    in
+    fun () -> Dune_config.Config.get config
+  ;;
+
+  let os = string_option_config "os"
+  let arch = string_option_config "arch"
+end
+
 (* CR-rgrinberg: do we need to call [uname] for every single option? Can't we
    call [uname -a] and extract everything from there *)
 let uname ~path args = run_capture_line ~path ~prog:"uname" ~args
 let lsb_release ~path args = run_capture_line ~path ~prog:"lsb_release" ~args
 
 let arch ~path =
-  (match Sys.os_type with
-   | "Unix" | "Cygwin" -> uname ~path [ "-m" ]
-   | "Win32" ->
-     Fiber.return
-     @@
-       (match OpamStubs.getArchitecture () with
-       | OpamStubs.AMD64 -> Some "x86_64"
-       | ARM -> Some "arm32"
-       | ARM64 -> Some "arm64"
-       | IA64 -> Some "ia64"
-       | Intel -> Some "x86_32"
-       | Unknown -> None)
-   | _ -> Fiber.return None)
-  >>| apply_or_skip_empty normalise_arch
+  match Config_override_variables.arch () with
+  | Some arch -> Fiber.return (Some arch)
+  | None ->
+    (match Sys.os_type with
+     | "Unix" | "Cygwin" -> uname ~path [ "-m" ]
+     | "Win32" ->
+       Fiber.return
+       @@
+         (match OpamStubs.getArchitecture () with
+         | OpamStubs.AMD64 -> Some "x86_64"
+         | ARM -> Some "arm32"
+         | ARM64 -> Some "arm64"
+         | IA64 -> Some "ia64"
+         | Intel -> Some "x86_32"
+         | Unknown -> None)
+     | _ -> Fiber.return None)
+    >>| apply_or_skip_empty normalise_arch
 ;;
 
 let os ~path =
-  (match Sys.os_type with
-   | "Unix" -> uname ~path [ "-s" ]
-   | s -> Fiber.return (norm s))
-  >>| apply_or_skip_empty normalise_os
+  match Config_override_variables.os () with
+  | Some os -> Fiber.return (Some os)
+  | None ->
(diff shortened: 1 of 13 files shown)


gridbugs:
Opening this PR to get feedback on my approach. If there are no major issues with how this works then I'll add some comments for posterity.

test/blackbox-tests/test-cases/pkg/lock-directory-selection.t
```diff
@@ -0,0 +1,114 @@
...
+  - macos-only.0.0.1
+  $ dune pkg lock dune.macos.lock
+  Solution for dune.macos.lock:
+  - macos-only.0.0.1
+  $ dune pkg lock dune.linux.lock
+  Solution for dune.linux.lock:
+  - linux-only.0.0.1
+
+Demonstrate that the correct lockdir is being chosen by building packages that
+are only dependen on on certain systems.
```
Leonidas-from-XIV, line 74: "dependent"?

src/dune_rules/expander.ml
```diff
@@ -426,28 +422,6 @@ let lib_config_var (var : Pform.Var.t) (lib_config : Lib_config.t) =
   ]
 ;;
 
-let ocaml_config_var (var : Pform.Var.t) (ocaml_config : Ocaml_config.t) =
```
Leonidas-from-XIV, line 429: Where did this code go? Was it duplicated?

Leonidas-from-XIV (approved):
Seems reasonable to have a mechanism to automatically assign the lockdir.

src/dune_rules/expander.ml
gridbugs: ah oops I forgot to add a file. It's in https://github.com/ocaml/dune/pull/9750/files#diff-5e4f430877c0a281dea7b3d55853b8a5743863c1528cc4b64c9fccebb25bc9b6

test/blackbox-tests/test-cases/pkg/lock-directory-selection.t
```diff
@@ -0,0 +1,115 @@
...
+  >   (os linux)))
+  > (lock_dir
+  >  (repositories mock))
+  > (repository
+  >  (name mock)
+  >  (source "file://$(pwd)/mock-opam-repository"))
+  > (context
+  >  (default
+  >   (lock_dir (cond
+  >    ((and (= %{architecture} arm64) (= %{system} macos)) dune.macos.arm64.lock)
```
rgrinberg, line 47: Do you think the extra pairs of paren is are necessary? It could be:

```
(cond
 (and (= %{architecture} arm64) (= %{system} macos)) dune.macos.arm64.lock
 ..)
```

So we make `cond` accept a list of an even number of elements. I know this isn't a how lisp does it, but clojure does it and it seems a little neater.

rgrinberg:
The implementation of the syntax looks good, but the data used to expand these pforms is suspicious. Using the compiler from the ambient environment to choose the lock directory is likely to cause unexpected behavior. For example, users aren't expected to have a compiler installed at all to use lock directory, and dune should install the compiler for them. Therefore, I suggest that we limit the pforms to whatever is computable without the compiler. See `Sys_poll` for such options.

test/blackbox-tests/test-cases/pkg/lock-directory-selection.t
rgrinberg: I also like how clojure uses `:else` to mark the catch-all clause.

gridbugs: I prefer to keep the extra parens as the conditions and lockfiles are more explicitly tied together by syntax. If we alternated interpreting terms as conditions and lockdirs then I'd worry about the types of syntax error that would become possible if dune accidentally interpreted a lockdir as a condition or vice versa.

I will add a catch-all clause.

gridbugs:
I've updated this to implement the pform expander with sys_poll rather than `ocamlc -config` and added a default case to `cond` statements with a `default` keyword.

src/dune_rules/lock_dir.ml
```diff
@@ -25,17 +25,40 @@ module Load = Make_load (struct
     ;;
   end)
 
+let select_lock_dir lock_dir_selection =
+  let* workspace = Workspace.workspace () in
+  let* raw_expander =
+    Memo.of_non_reproducible_fiber
+      (Sys_poll.make ~path:(Env_path.path Env.initial)
```
rgrinberg, line 32: You can share this computation with what we have in `pkg_rules` I think. That would make it so that we don't evaluate these variables more than once per run.

src/dune_pkg/sys_poll.ml
```diff
@@ -48,33 +48,51 @@ let run_capture_line ~path ~prog ~args =
...
+module Config_override_variables = struct
+  let string_option_config name =
+    let config =
+      Dune_config.Config.make ~name ~of_string:(fun s -> Ok (Some s)) ~default:None
+    in
+    fun () -> Dune_config.Config.get config
+  ;;
+
+  let os = string_option_config "os"
+  let arch = string_option_config "arch"
```
rgrinberg, line 60: I wonder if we should just allow overriding these variables at build rime rather than just during solving. That would replace this config and and would likely be a useful feature in general.

rgrinberg (approved):

src/dune_rules/lock_dir.ml
gridbugs: Moved the logic into lock_dir.ml and reused in pkg_rules.ml

src/dune_pkg/sys_poll.ml
gridbugs: This should already be the case. The overrides are applied in functions in `Sys_poll` such as `Sys_poll.os` which are used to compute variables both at solve time and build time.

PR #9756, opened 2024-01-16
Dune should execute inline tests whose only mode is `byte`

lukstafi:
Intended for testing behavior under the bytecode compiler.

See issue: _`dune runtest` compiles but does not execute `(inline_tests) (modes byte)`_


rgrinberg:
Thank your for the test. Could you sign your commit as per https://github.com/ocaml/dune/blob/main/CONTRIBUTING.md

Alizter:
Thanks for the test. Superseded by #13199.

PR #9761, opened 2024-01-17
Faster `Dune_file.t` equality check

### LINKED ISSUE #9738: Performance regression after #8447
While testing most recent commits from `main` I noticed that noop builds were sustantially slower in the last weeks.

After dissecting for the past few months, I could pin down to commit 9105a43b719cc66aefd7c7f84816bd1a77f236c7, from https://github.com/ocaml/dune/pull/8447, cc @rgrinberg @Alizter as PR author and reviewer.

The regression makes the noop build in Ahrefs case almost 2x slower:

```
# using commit 9105a43b719cc66aefd7c7f84816bd1a77f236c7
$ time ../dune/_build/default/bin/main.exe build @fe.all
                                             
real    0m43.125s
user    0m41.054s
sys     0m2.092s

# using commit 9c41108359ac93d369a6e78da88ffbde8b1f8ca9
$ time ../dune/_build/default/bin/main.exe build @fe.all
                                             
real    0m22.828s
user    0m21.099s
sys     0m1.760s
```

All other things in the measurements above are equal. I am building the dune exe using `dune build bin/main.exe` and calling the exe directly as shown in the `time` commands above.

I will investigate more tomorrow to see if I can identify which particular rules are slowing down, in case there are any other info I can provide, please lmk.

jchavarri:
Fixes #9738, bringing down the build times measured in that issue to ~21s.

Makes the assumption that given two `Dune_file.t` values, if their `path` exists and it is the same, then they are the same. I am not 100% sure this is correct.

### DIFF
--- a/src/dune_rules/source_tree.ml
+++ b/src/dune_rules/source_tree.ml
@@ -86,9 +86,10 @@ module Dune_file = struct
   ;;
 
   let equal { path; kind; plain } t =
-    Option.equal Path.Source.equal path t.path
-    && equal_kind kind t.kind
-    && Plain.equal plain t.plain
+    match path, t.path with
+    | Some a, Some b -> Path.Source.equal a b
+    | Some _, None | None, Some _ -> false
+    | None, None -> equal_kind kind t.kind && Plain.equal plain t.plain
   ;;
 
   let get_static_sexp t = t.plain.contents.sexps


rgrinberg:
This change isn't correct. First of all it breaks the following invariant `x = y => hash x = hash y`. But even if we were to fix hashing, it would still be incorrect. In particular, it would break in watch mode when the dune file needs to be recomputed as per the following sequence of events:

1. The first call to `in_dir dir` will encounter no memo node and create one for `dir`. So far so good.
2. The dune file in `dir` is edited, and then dune tries to recompute the dune file in it.
3. The recomputation will cause `in_dir dir` to be re-evaluated.
4. At this point, memo will find the memo node we created in step 1.
5. This reused memo node still has the old input value of type `Source_tree.Dune_file.t`, so invalidating and recomputing isn't going to produce the most up to date result.

There's two ways to fix the issue. One is to make `equal` cheap and maintain its correctness. That was my suggestion of adding an md5 digest to the input. Another way is to simply avoid the reliance on this quality function. I did that in https://github.com/ocaml/dune/pull/9769

I hope this explanation helps.

jchavarri:
Makes sense 👍 Thanks for the explanation.

PR #9763, opened 2024-01-17
refactor: unify package variables

rgrinberg:
Use a single representation for package variables in [Dune_lang] and
[Dune_pkg].

It doesn't make sense for these things to be represented differently. If
there's any drift between these types, there's a bug in our code.


rgrinberg:
This pulls in `opam_format` in `dune_lang`. I'm not sure what was the objection to doing that was before, but I don't see much downside to it.

gridbugs (approved):

PR #9769, opened 2024-01-18
fix: performance regression from #8447

rgrinberg:
Get rid of the slow dune file comparison in #9738


anmonteiro:
Explanation of this fix in https://github.com/ocaml/dune/pull/9761#issuecomment-1897776677

Issue #9746, opened 2024-01-15
Re-downloading the same sources when computing pins or packages without checksums


rgrinberg:
If we have two packages such as:

```
$ cat lwt/1.0.0/opam
..
url {
  src: "http://github.com/1.0.0/lwt.tar.gz"
}
$ cat lwt-unix/1.0.0/opam
..
url {
  src: "http://github.com/1.0.0/lwt.tar.gz"
}
```

We are going to download them twice during the solve phase. A similar problem exists with `pin-depends`. This is all rather wasteful, especially for projects that define multiple packages - common for dune-projects.

We should maintain the already downloaded sources as part of the solve phase and not redo this work.

@gridbugs could you have a look at this one?

gridbugs:
@rgrinberg is this issue just about reusing downloaded files within a single run of dune, or across multiple invocations of dune?

Leonidas-from-XIV:
I've seen #9771 but I'm wondering that if we're adding a cache, wouldn't it be better to store the artifacts persistently? This is what the previous OPAM code did: download the file and add it to the cache if the checksum matches, then pull it out of the cache the next time the file is requested (and reverify the checksums in case of data corruption).

rgrinberg:
It's an option, but not my first one. We'd have to re-implement the dune cache without deduplication, sharing, the UI to manage it. If we realize that the dune cache can't be used for this purpose, then we can consider creating our own parallel implementation.

Issue #9701, opened 2024-01-10
Re-downloading the same sources when building packages


rgrinberg:
Our source fetching rules do a bunch of redundant work. For example:

* Fetching the same package sources in different contexts
* Fetching package sources for packages that live in the same repository

We should adjust the rules to deduplicate this work.

This will also make the source fetching rules work much better with the dune cache

rgrinberg:
Here's a concrete approach to fix the problem:

1. We download all sources into `_build/_source/$checksum`
2. To setup the rules in `_build/_source/$checksum`, we iterate over all the sources in all the lock directories, and build a reverse map between checksums and their source definitions
3. When constructing package sources, we copy them from the directory in 1.

The downside of this scheme is that there's a bunch of pointless copying here due to sandboxing and all the intermediate artifacts.

Oh and there's no need to use this scheme for local sources (obvious reasons) or git b/c there's no downloading as everything exists in the rev store.

rgrinberg:
Reproduced in https://github.com/ocaml/dune/pull/9741

PR #9771, opened 2024-01-18
pkg: reuse downloaded archives within single run

gridbugs:
Archives downloaded over http by package management code will be cached in a temporary directory that persists during a single invocation of dune. This is intended to reduce the number of times that dune downloads the same file during package management operations. Note that it's common in the ocaml ecosystem for a single source archive to contain multiple packages.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/compute-checksums-when-missing.t
+++ b/test/blackbox-tests/test-cases/pkg/compute-checksums-when-missing.t
@@ -100,3 +100,49 @@ Check that no checksum is computed for a local source directory:
   $ solve foo 2>&1 | strip_transient
   Solution for dune.lock:
   - foo.0.0.1
+
+
+Create 3 packages that all share the same source url with no checksum. Dune
+will need to download each package's source archive to compute their hashes.
+Test that dune only downloads the file a single time since the source url is
+identical among the packgaes. The fact that the download only occurs once is
+asserted by the fact that the webserver will only serve the file a single time.
+  $ webserver_oneshot --content-file foo.txt --port-file port.txt &
+  $ until test -f port.txt; do sleep 0.1; done
+  $ PORT=$(cat port.txt)
+
+  $ mkpkg foo <<EOF
+  > url {
+  >  src: "http://0.0.0.0:$PORT"
+  > }
+  > EOF
+
+  $ mkpkg bar <<EOF
+  > url {
+  >  src: "http://0.0.0.0:$PORT"
+  > }
+  > EOF
+
+  $ mkpkg baz <<EOF
+  > url {
+  >  src: "http://0.0.0.0:$PORT"
+  > }
+  > EOF
+
+  $ solve foo bar baz | strip_transient
+  Solution for dune.lock:
+  - bar.0.0.1
+  - baz.0.0.1
+  - foo.0.0.1
+  Package "bar" has source archive which lacks a checksum.
+  The source archive will be downloaded from: <addr>
+  Dune will compute its own checksum for this source archive.
+  Package "baz" has source archive which lacks a checksum.
+  The source archive will be downloaded from: <addr>
+  Dune will compute its own checksum for this source archive.
+  Package "foo" has source archive which lacks a checksum.
+  The source archive will be downloaded from: <addr>
+  Dune will compute its own checksum for this source archive.
+
+  $ wait
+
--- a/src/dune_pkg/single_run_file_cache.ml
+++ b/src/dune_pkg/single_run_file_cache.ml
@@ -1,0 +1,50 @@
+open! Import
+open! Stdune
+
+type entry =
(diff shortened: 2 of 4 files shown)


gridbugs:
fixes https://github.com/ocaml/dune/issues/9746

src/dune_pkg/fetch.ml
```diff
@@ -188,18 +188,78 @@ type failure =
...
+     invocation of dune and is deleted when dune terminates *)
+  type t =
+    { temp_dir : Path.t
+    ; url_to_filename : entry String.Map.t ref
+    }
+
+  let create () =
+    let temp_dir = Temp.create Dir ~prefix:"dune" ~suffix:"fetch-archive-cache" in
+    at_exit (fun () -> Temp.destroy Dir temp_dir);
+    { temp_dir; url_to_filename = ref String.Map.empty }
```
Leonidas-from-XIV, line 207: If this is a `Map ref`, maybe `Hashtbl` could be an option. It has slightly different semantics.

src/dune_pkg/fetch.ml
```diff
@@ -188,18 +188,78 @@ type failure =
...
+  ;;
+
+  (* [with_ t ~url ~f] checks whether [url] corresponds to an entry in the
+     cache. If an entry is present then the name of the cached file is
+     returned. Otherwise [f] is called on a path and is expected to download
+     the file at [url] and store it at that path. If [f] returns [Ok ()] but it
+     fails to do create a file at the given path then a [Code_error] is raised.
```
Leonidas-from-XIV, line 214: I wonder, would it not be better if `f` returned either the path to which the file was downloaded (and then the caller moves it to the cache) or the contents? That way instead of `()` the return value can be something meaningful.

test/blackbox-tests/test-cases/pkg/compute-checksums-when-missing.t
```diff
@@ -100,3 +100,49 @@ Check that no checksum is computed for a local source directory:
...
+  - foo.0.0.1
+  Package "bar" has source archive which lacks a checksum.
+  The source archive will be downloaded from: <addr>
+  Dune will compute its own checksum for this source archive.
+  Package "baz" has source archive which lacks a checksum.
+  The source archive will be downloaded from: <addr>
+  Dune will compute its own checksum for this source archive.
+  Package "foo" has source archive which lacks a checksum.
+  The source archive will be downloaded from: <addr>
+  Dune will compute its own checksum for this source archive.
```
Leonidas-from-XIV, line 145: I think the cache code should only trigger if a checksum was specified, because if the URL is specified multiple times it doesn't mean that the data returned there needs to be identical.

rgrinberg:
Would it be possible to move this logic away from `Fetch` and into `compute_missing_checksums`?

A couple of reasons  for this:

1. It won't be needed for building once I'm done with https://github.com/ocaml/dune/issues/9701
2. I'm going to need a data structure that keeps track of what's already been fetched for pins. This is to have cycles when there are circular references.

src/dune_pkg/fetch.ml
```diff
@@ -188,18 +188,78 @@ type failure =
 
 let label = "dune-fetch"
 
+module Single_run_cache = struct
+  type entry =
+    { filename : Filename.t
+    ; download_complete : unit Fiber.t
```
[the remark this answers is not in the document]
gridbugs: done

src/dune_pkg/fetch.ml
gridbugs: done

src/dune_pkg/fetch.ml
gridbugs: I think it's simpler to have `f` create the file at a specified path rather than letting `f` choose its own path. This allows the cached files to be stored in a centralized location of the cache's choosing and in some cases clients of the cache will be able to use the cached files without needing to copy them elsewhere (such as when computing their checksums).

test/blackbox-tests/test-cases/pkg/compute-checksums-when-missing.t
gridbugs: That's a valid concern but one of the things we want to achieve with this cache is avoiding downloading files multiple times when downloading files for the purposes of computing their checksum (when the package lacks a checksum).

gridbugs:
@rgrinberg  I've moved the cache implementation into its own file and use it directly when computing checksums rather than in all downloads. It would be trivial to add a function for interrogating the contents of the cache. It would be nice if we could reuse files downloaded when computing their checksums when building packages however. It might make sense to use https://github.com/ocaml/dune/issues/9701 for caching files that dune downloads to compute checksums rather than having the separate caching mechanism introduced in this PR.

rgrinberg (approved):

Leonidas-from-XIV (approved):

PR #9793, opened 2024-01-22
Add --files to dune describe opam-files

reynir:
This flag makes dune describe opam-files only print the opam file names, each on their own line. This is useful for shell scripts where parsing sexp can be somewhat complicated.

The flag unfortunately ignores the `--format` argument. Another option could be to add a NUL-byte delimited format assuming none of the output would include NUL bytes (which I'm not sure is a reasonable assumption).

- [x] changelog

### DIFF
--- a/bin/describe/describe_opam_files.ml
+++ b/bin/describe/describe_opam_files.ml
@@ -3,7 +3,13 @@ open Import
 let term =
   let+ builder = Common.Builder.term
   and+ format = Describe_format.arg
-  and+ _ = Describe_lang_compat.arg in
+  and+ _ = Describe_lang_compat.arg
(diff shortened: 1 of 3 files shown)


emillon:
Actually I'm wondering if we should make `format` per-command so that your PR could be expressed as just `--format=lines`.

(also, you'll need to sign-off your contribution as described in https://github.com/ocaml/dune/blob/main/CONTRIBUTING.md)

Alizter:
@reynir I've renamed the flag to `--files`, done some cleanup and rebased with a changelog. Thanks for the contribution!

rgrinberg (approved):

PR #9798, opened 2024-01-22
feature(pkg): Submodule support

Leonidas-from-XIV:
This adds support for repos with submodules. If submodules are defined in the repo and commit objects exist at the specified location, the specified repo will be added as a remote and the revisions will get added to the rev-store.

When accessing the revisions, the `At_rev` will look up directories and content in the submodule `At_rev`s.

### DIFF
--- a/src/dune_pkg/git_config_parser.ml
+++ b/src/dune_pkg/git_config_parser.ml
@@ -1,0 +1,70 @@
+open Import
+
+type bindings = string * string
+
+type section =
+  { name : string
+  ; arg : string option
+  ; bindings : bindings list
+  }
+
+type t = section list
+
+let strip =
+  let re = Re.(compile (alt [ seq [ bos; rep blank ]; seq [ rep blank; eos ] ])) in
+  Re.replace_string ~all:true re ~by:""
+;;
+
+let parse_section_bindings =
+  let binding_re =
+    Re.(
+      compile
+      @@ seq [ rep space; group (rep1 (diff any (char '='))); char '='; group (rep1 any) ])
+  in
+  fun lines ->
+    List.filter_map lines ~f:(fun line ->
+      Re.exec_opt binding_re line
+      |> Option.map ~f:(fun m ->
+        let name = Re.Group.get m 1 |> strip in
+        let value = Re.Group.get m 2 |> strip in
+        name, value))
+;;
+
+let parse_section_header =
+  let identifier = Re.(rep1 (diff any blank)) in
+  let section_re =
+    Re.(
+      compile
+      @@ seq
+           [ char '['
+           ; group identifier
+           ; alt [ char ']'; seq [ rep1 blank; char '"'; group (rep1 any); str "\"]" ] ]
+           ])
+  in
+  fun line ->
+    Re.exec_opt section_re line
+    |> Option.map ~f:(fun m -> Re.Group.get m 1, Re.Group.get_opt m 2)
+;;
+
+let rec parse_sections acc lines =
+  match lines with
+  | [] -> List.rev acc
+  | header :: rest ->
+    (match parse_section_header header with
+     | Some (name, arg) ->
+       let this, rest =
+         List.split_while rest ~f:(fun line ->
+           not @@ String.is_prefix ~prefix:"[" @@ strip line)
+       in
+       let bindings = parse_section_bindings this in
+       let section = { name; arg; bindings } in
+       parse_sections (section :: acc) rest
+     | None ->
+       (* not a header, skip line *)
+       parse_sections acc rest)
+;;
+
+let parse s =
+  let lines = String.split ~on:'\n' s in
+  parse_sections [] lines |> Result.ok
+;;
--- a/src/dune_pkg/git_config_parser.ml
+++ b/src/dune_pkg/git_config_parser.ml
@@ -16,22 +16,45 @@ let strip =
 ;;
 
 let parse_section_bindings =
-  let binding_re =
+  let binding_name = Re.(seq [ alpha; rep alnum ]) in
+  let binding_assignment_re =
     Re.(
       compile
-      @@ seq [ rep space; group (rep1 (diff any (char '='))); char '='; group (rep1 any) ])
+      @@ seq
+           [ bol
+           ; rep space
+           ; group binding_name
+           ; rep space
+           ; char '='
+           ; rep space
+           ; group (rep1 any)
+           ; rep space
+           ; eol
+           ])
+  in
+  let binding_boolean_re =
+    Re.(compile @@ seq [ bol; rep space; group binding_name; rep space; eol ])
   in
   fun lines ->
     List.filter_map lines ~f:(fun line ->
-      Re.exec_opt binding_re line
-      |> Option.map ~f:(fun m ->
-        let name = Re.Group.get m 1 |> strip in
-        let value = Re.Group.get m 2 |> strip in
-        name, value))
+      let binding_assignment =
+        Re.exec_opt binding_assignment_re line
+        |> Option.map ~f:(fun m ->
+          let name = Re.Group.get m 1 in
+          let value = Re.Group.get m 2 in
+          name, value)
+      in
+      match binding_assignment with
+      | Some _ as v -> v
+      | None ->
+        Re.exec_opt binding_boolean_re line
+        |> Option.map ~f:(fun m ->
+          let name = Re.Group.get m 1 in
+          name, "true"))
 ;;
 
 let parse_section_header =
-  let identifier = Re.(rep1 (diff any blank)) in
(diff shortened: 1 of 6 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -301,55 +339,188 @@ module File = struct
...
+      let locate_by_re re line =
+        Re.exec_opt re line |> Option.map ~f:(fun m -> Re.Group.get m 1)
+      in
+      let path = List.find_map section ~f:(locate_by_re path_re) in
+      let url = List.find_map section ~f:(locate_by_re url_re) in
+      match path, url with
+      | Some path, Some source ->
+        let path = Path.Local.of_string path in
+        { path; source }, rest
+      | _, _ -> failwith "Submodule definition missing path & url"
```
rgrinberg, line 387: Should be either Code_error or User_error

src/dune_pkg/rev_store.ml
```diff
@@ -301,55 +339,188 @@ module File = struct
...
+    in
+    let rec parse submodules = function
+      | [] -> submodules
+      | line :: lines ->
+        (match String.is_prefix ~prefix:"[submodule" line with
```
rgrinberg, line 392: The config parser had a few bugs with respect to parsing, can we switch to something a bit more principled: E.g.

```
type section =
  { name : string
  ; arg : string option
  ; bindings : bindings
  }
val parse : string -> section list
```

Btw, I find it a bit strange that for remotes we chose to use `git remote` to get the information, but for submodules we're parsing `.gitmodules`. Might as well stick do the same thing for both (I'd recommend just parsing the config format, which is more probably more stable than the output of git commands)

src/dune_pkg/rev_store.ml
```diff
@@ -16,6 +16,9 @@ let lock_path { dir } =
 
 type rev = Rev of string
 
+let rev_compare (Rev a) (Rev b) = String.compare a b
+let rev_to_dyn (Rev r) = Dyn.variant "Rev" [ Dyn.string r ]
```
rgrinberg, line 20: Might as well introduce a `Rev` submodule to group these.

src/dune_pkg/rev_store.ml
```diff
@@ -251,22 +254,57 @@ let load_or_create ~dir =
...
+
+    let compare a b =
+      match a, b with
+      | Blob a, Blob b -> compare_blob a b
+      | Commit a, Commit b -> compare_commit a b
+      | Blob _, Commit _ -> Ordering.Lt
+      | Commit _, Blob _ -> Ordering.Gt
+    ;;
+
+    let commit_to_dyn { path; rev } =
```
rgrinberg, line 293: `dyn_of_commit` or introduce a `Commit` submodule (preferable)

rgrinberg (changes requested):
Looks like a good start. I left some comments.

The biggest omission that I see is that we're not support submodules for the sources of packages themselves. I think that's quite a bit more common than using submodules for opam repositories themselves.

So I see two ways around this:
1. Implement submodule fetching in `fetch.ml`
2. Just switch `fetch.ml` to use the revision store for git repos

In either case, I suggest that we add some tests for building packages that use sources with submodules (in a separate PR)

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: I believe I had a branch where I replaced a number of git calls with parsing the config files, but if I remember it correctly there was some concern about replacing git commands with parsing the files. Unfortunately I can't seem to find the discussion, @emillon do you remember this? However, which commands are plumbing is not well defined.

Anyway, yes to a simple parser, we seem to need this every now and then. Replacing `git remote` would however be a different PR.

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: Indeed, looks like I overlooked this when submitting. However, I wonder about using `Loc.t` for non-dune files: would that be a good idea?

rgrinberg: If the error corresponds to something that comes from a source file, that would be a good idea. I wouldn't use `Loc.t` for things like internal config files of git for example though.

src/dune_pkg/rev_store.ml
rgrinberg: Yes, there's no need to touch `git remote` here. I'm just questioning why we're not handling all the metadata consistently.

I don't know why Etienne prefers to use `git remote`, but it seems worse than reading the config file. Git's config files maintain at least a certain level of backwards compatibility b/c newer versions of git need to be able to read them. While the output of individual git commands is not guaranteed to be stable anywhere (although in practice it's quite stable)

test/expect-tests/dune_pkg/git_config.ml
```diff
@@ -0,0 +1,137 @@
...
+
+let%expect_test "parsing named section" =
+  let config = {|[foo "name"]
+    bar = baz
+    |} in
+  print_or_fail config;
+  [%expect
+    {|
+  [ { name = "foo"; arg = Some "name"; bindings = [ ("bar", "baz") ] } ]
+  |}]
```
rgrinberg, line 136: Is there a test for config that doesn't start with a section? E.g
```
foo = bar
[section]
x = y
```

In the above, the binding `foo = bar` should belong to the anonymous section.

Leonidas-from-XIV: According to [the documentation that I just found](https://git-scm.com/docs/git-config)

> Each variable must belong to some section, which means that there must be a section header before the first setting of a variable.

Thus there are no anonymous sections. But I will update the parser on the syntax restrictions mentioned in the document, there's a number of things not supported yet.

src/dune_pkg/rev_store.ml
emillon: my opinion was that it's not dune's job to read other programs' config files (at one point, we said we wouldn't parse `.ocamlformat` files for example). but I'm pretty far from this area of the code, and I don't know what guarantees `git` makes for its output and for its config file, so I don't have a strong opinion on this.

rgrinberg: Avoiding depending on the implementations of external projects is certainly desirable, but there's no advantage on depending on the output of an external command vs the format of a file. There are no stability guarantees on the output of the command as far as I know.

src/dune_pkg/git_config_parser.ml
gridbugs, line 1: I think the regexes in this file would be easier to understand and maintain if they were written using strings rather than seqs. What's the reason for using seqs here?

src/dune_pkg/rev_store.ml
emillon: what about accessing this data through `git config`? you can list the bindings using `git config --list`, and get one using for example `git config --get remote.origin.url`? I'm pretty sure that this is easier to parse than git config files (and you don't have to know the logic of where to look for files, etc)

rgrinberg: Wasn't aware of it. Seems like a good option as well. Can it be used to edit the config as well? Can we use it to access submodules?

If yes, seems like this could have saved us some parsing work. The parsing is already done, so it probably wouldn't be our first priority to rewrite it though.

emillon: Yes, this can be used to edit the config files as well. It returns precise exit codes too (see `git-config(1)`), so I think it would be a useful tool to use.

rgrinberg: and submodules?

emillon: do you mean, how to parse `.gitmodules`? yes, you can do that using `git config --file .gitmodules --list` for example. you can also refer to a blob like `--blob main:.gitmodules` if you don't have a working tree.

rgrinberg:
@Leonidas-from-XIV I pushed some changes. I think the simplest way to make this feature reliable is to just resolve all files in submodules into their correct blobs. This way, the code for fetching multiple files can continue working transparently and not have to worry about the existence of submodules. We lose some efficiency this way, as we don't really need to fetch the submodules until we access a path that is under the submodule, but the old code didn't have this property either.

Some other issues I've noted are in the code. I'll list a few other issues as well. They don't need to be addressed right now, but it's good to put them on the TODO list.

1. We need tests for submodules that have submodules
2. Additionally, we need to make sure we don't allow cycles while fetching submodules.
3. `check_out` on repos with submodules is untested

src/dune_pkg/rev_store.ml
rgrinberg: Very cool. Seems like that's all we need then.

rgrinberg (approved):

rgrinberg:
Going to merge this to save ourselves the rebase grief and following can work can be done in subsequent PR's. Thanks for the hard work!

PR #9811, opened 2024-01-23
fix(subst): skip large files on 32-bit

### LINKED ISSUE #9538: dune subst crashes on large files (32-bit systems)
`dune subst` crashes on 32-bit systems when a large file is present:

(true as of 3.12.1)

```
  Error: exception Invalid_argument("Bytes.create")
  Raised by primitive operation at Stdune__Io.Make.eagerly_input_string in file
    "otherlibs/stdune/src/io.ml", line 273, characters 14-30
  Called from Stdune__Io.Make.read_all.(fun) in file
    "otherlibs/stdune/src/io.ml", line 308, characters 16-40
  Called from Stdune__Exn.protectx in file "otherlibs/stdune/src/exn.ml", line
    10, characters 8-11
  Re-raised at Stdune__Exn.protectx in file "otherlibs/stdune/src/exn.ml", line
    16, characters 4-11
  Called from Dune__exe__Subst.subst_file in file "bin/subst.ml", line 113,
    characters 10-27
  Called from Stdlib__List.iter in file "list.ml", line 110, characters 12-15
  Called from Fiber__Core.O.(>>|).(fun) in file "vendor/fiber/src/core.ml",
    line 252, characters 36-41
    line 76, characters 8-11
    line 76, characters 8-11
```

emillon:
Fixes #9538

When `subst` tries to work on a file, it first loads it in memory as an OCaml `string`. This type has a maximum size, `Sys.max_string_length`. This is not an issue on 64-bit systems, but the 32-bit limit can be below the size of actual files.

However, it is unlikely that these files need substitutions, so we just log a warning in this case.

### DIFF
--- a/bin/subst.ml
+++ b/bin/subst.ml
@@ -120,15 +120,27 @@ let subst_string s path ~map =
 ;;
 
 let subst_file path ~map =
-  let s = Io.read_file path in
-  let s =
-    if Path.is_root (Path.parent_exn path) && Package.is_opam_file path
-    then "version: \"%%" ^ "VERSION_NUM" ^ "%%\"\n" ^ s
-    else s
-  in
-  match subst_string s ~map path with
-  | None -> ()
-  | Some s -> Io.write_file path s
+  match Io.with_file_in path ~f:Io.read_all_unless_large with
+  | Error () ->
+    let hints =
+      if Sys.word_size = 32
+      then
+        [ Pp.textf
+            "Dune has been built as a 32-bit binary so the maximum size \"dune subst\" \
+             can operate on is 16MiB."
+        ]
+      else []
+    in
+    User_warning.emit ~hints [ Pp.textf "Ignoring large file: %s" (Path.to_string path) ]
+  | Ok s ->
+    let s =
+      if Path.is_root (Path.parent_exn path) && Package.is_opam_file path
+      then "version: \"%%" ^ "VERSION_NUM" ^ "%%\"\n" ^ s
+      else s
+    in
+    (match subst_string s ~map path with
+     | None -> ()
+     | Some s -> Io.write_file path s)
 ;;
 
 (* Extending the Dune_project APIs, but adding capability to modify *)
--- a/doc/changes/9811.md
+++ b/doc/changes/9811.md
@@ -1,0 +1,1 @@
+- subst: do not fail on 32-bit systems when large files are encountered. Just log a warning in this case. (#9811, fixes #9538, @emillon)
--- a/test/blackbox-tests/test-cases/subst/32bit.t
+++ b/test/blackbox-tests/test-cases/subst/32bit.t
@@ -26,10 +26,7 @@ This test uses subst, which needs a git repository:
    create mode 100644 dune-project
    create mode 100644 large.dat
 
-  $ dune subst 2>&1 | head -n 6
-  Error: exception Invalid_argument("Bytes.create")
-  Raised by primitive operation at Stdune__Io.Make.eagerly_input_string in file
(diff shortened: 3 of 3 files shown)


bin/subst.ml
```diff
@@ -110,15 +110,18 @@ let subst_string s path ~map =
...
-    if Path.is_root (Path.parent_exn path) && Package.is_opam_file path
-    then "version: \"%%" ^ "VERSION_NUM" ^ "%%\"\n" ^ s
-    else s
-  in
-  match subst_string s ~map path with
-  | None -> ()
-  | Some s -> Io.write_file path s
+  match Io.read_file path with
+  | exception Io.Large_file ->
+    User_warning.emit [ Pp.textf "Ignoring large file: %s" (Path.to_string path) ]
```
Leonidas-from-XIV, line 115: The warning is very useful, that way we can also monitor via bug reports if this is a problem that we would need to solve in a different way, e.g. by streaming the file.

I think it would be useful to add a `hint` to tell the user that the file is too large for 32 bit platforms.

bin/subst.ml
```diff
@@ -110,15 +110,18 @@ let subst_string s path ~map =
...
+  | s ->
+    let s =
+      if Path.is_root (Path.parent_exn path) && Package.is_opam_file path
```
Leonidas-from-XIV, line 118: I know that this is from the original code but this looks like it needs to be adopted if we generate OPAM files in the `opam/` subdirectory.

otherlibs/stdune/src/io.ml
```diff
@@ -270,6 +272,7 @@ struct
      This can be detected by checking that the length of the resulting string is
      less than [len]. *)
   let eagerly_input_string ic len =
```
Leonidas-from-XIV, line 274: Add `_exn` as s suffix here?

otherlibs/stdune/src/io.mli
```diff
@@ -9,6 +9,10 @@ val input_lines : in_channel -> string list
     unrelated channels because it uses a statically-allocated global buffer. *)
 val copy_channels : in_channel -> out_channel -> unit
 
+(** This is raised when [read_all] or similar functions try to operate on files larger than [Sys.max_string_length]. This should be an issue only on 32-bit systems where this is 16 MiB. *)
+exception Large_file
+
+(** Read all of a channel. Can raise [Large_file]. *)
 val read_all : in_channel -> string
```
Leonidas-from-XIV, line 16: Likewise, should probably be `read_all_exn`.

emillon: Yes, even if in most cases we don't have to handle it, it might be a good nudge. I think I'll make that change in a separate PR.

rgrinberg:
The change looks fine, but we need to be careful with introducing new exceptions. The issue is that exceptions don't get marshalled automatically over RPC so you can just break things internally when this error is raised. Instead, I propose the following:

1. Introduce a function `val read_file_unless_large : path -> (string, unit) result`
2. use this function to implement the current `read_file`. The `Error` case can be turned into a Code_error which is a safe exceptions to use
3. In the subst code, use `read_file_unless_large` directly.

bin/subst.ml
emillon: done. not too sure about the wording, let me know what you think

emillon:
Thanks. I adapted that to the new API.

bin/subst.ml
Leonidas-from-XIV: Looks good to me!

Leonidas-from-XIV (approved):

PR #9828, opened 2024-01-24
refactor: detect large files in Io functions

emillon:
`Io.read_all` and related functions read the contents of a file in a string, which has a size limit (`Sys.max_string_length`) and can be an issue in 32-bit systems. This makes an explicit check and raises a `Code_error` in these situations.


emillon:
I tried to write some tests when the channel is not seekable; in that case `Buffer` throws exceptions. This will probably be a lot of work to try to detect this correctly, so I instead documented this as a limitation.

rgrinberg (approved):

PR #9829, opened 2024-01-24
doc: explain the bootstrap process

emillon:


doc/explanation/bootstrap.rst
```diff
@@ -0,0 +1,45 @@
+How Dune Uses Dune to Build Dune
```
christinerose, line 1: I love this header!

christinerose (approved):

doc/explanation/bootstrap.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Step 2: Bootstrap Dune
+----------------------
+
+``_boot/dune.exe`` is the bootstrap Dune. Since it has been built from
+the Dune sources, it will act like Dune. However, since it has not been produced by
+Dune, it might not be reproducible or might contained unused modules (like
+platform-specific code).
+
+So the next step is to compile the Dune sources using the bootstrap Dune:
```
rgrinberg, line 27: Note that no compilation is does in this step. We just copy the bootstrapped dune into `_build/install`. This step is only used to generate the .install file.

emillon: - wait, shouldn't we? this means that the installed dune has been produced by the bootstrap process, not by dune?
- Also, when we run `make dev`, we see that after the bootstrap, everything is re-built using an actual dune. is that just for the libs?

doc/explanation/bootstrap.rst
emillon: that's a reference to https://zachholman.com/talk/how-github-uses-github-to-build-github/

emillon: though we miss the triple meanint in the original ("how github (the company) uses github (the piece of software) to build github (the service)")

doc/explanation/bootstrap.rst
rgrinberg: > wait, shouldn't we? this means that the installed dune has been produced by the bootstrap process, not by dune?

We used to do that. It doubled the build times so users were relieved that we got rid of it.

> Also, when we run make dev, we see that after the bootstrap, everything is re-built using an actual dune. is that just for the libs?

For the libs, for merlin support, and of course for incremental builds.

emillon: thanks a lot, I had missed that distinction. I'll incorporate that into the doc

rgrinberg (approved):

rgrinberg (approved):
LGTM. The bootstrapping section in hacking.rst needs to be merged with this. There's some redundancy between the two now.

PR #9839, opened 2024-01-25
Add test to show lib name collision in contexts

jchavarri:
I am checking Dune workspaces as part of some exploration to find better ways to configure projects that use both Melange and OCaml libraries with some shared code.

My assumption was that a workspace has its own library namespace, and it is not shared with other workspaces. So it would be possible to define a library `foo` in two workspaces A and B and avoid collisions through the `enabled_if` field.

As this test shows, this is not the case, and Dune will error out because the library is defined twice.

Is the assumption incorrect? If it isn't, I would be happy to contribute a fix for it to continue this exploration, thanks!


rgrinberg:
Your test indeed demonstrates a limitation of dune's rules. You're welcome to try and lift this limitation. See `scope.ml` and `lib.ml` for how the library database is constructed. It shouldn't be too hard to allow more than one entry per name as long as only one is enabled.

Note that after you lift it, likely it will not be enough because `enabled_if` on a library accepts very few arguments. That's a different limitation that I would also like to lift.

jchavarri:
Thanks for the pointers, will take a look.

> likely it will not be enough because `enabled_if` on a library accepts very few arguments. That's a different limitation that I would also like to lift

What kind of new arguments do you have in mind?

jchavarri:
Sharing here some notes of what I've been able to find out.

Currently, there is a uniform "view" (for lack of a better name) of library dbs that is shared across all contexts. It is created in the [Scope.DB.all](https://github.com/ocaml/dune/blob/07ba1765e0ffae2716eb7a8983609e2c8bf5af8d/src/dune_rules/scope.ml#L327-L340) function, which is called at the very beginning of each Dune run.

Having a uniform view of the library dbs makes it impossible to keep checking for duplication of libraries when `Scope.DB.all` is called, because at that point we don't have the information yet on the "expander" that is needed to resolve the `enabled_if` field. So it seems necessary to move the check to a further stage in order to pass the context. Is that right? Otherwise, I think we should change the map used in `Scope` (which is based on `Lib_name`) to consider more things besides the library name when finding items in it.

I wonder if rather than using `enabled_if` for this, there could be a new field `for_context` added to the `library` stanza. That way, we could just add the libraries to the DB of the context they belong to "statically", without having to wait for a later stage, or mess with Blang expanders. Similarly to what is done for filtering by packages:

https://github.com/ocaml/dune/blob/07ba1765e0ffae2716eb7a8983609e2c8bf5af8d/src/dune_rules/scope.ml#L336

The Dune docs only mention `os_type` in the variables section of `enabled_if`:

https://github.com/ocaml/dune/blob/07ba1765e0ffae2716eb7a8983609e2c8bf5af8d/doc/stanzas/library.rst#L193-L196

I think `enabled_if` design is a bit lose, that's why I propose the `for_context` field, which would settle down conditionally available libraries per context within a more stable design.

rgrinberg:
`Scope.DB.all` is computed for all contexts, but it does not create a database for all the contexts, it creates a database for every context. Note the context keyed map it generates.

> Otherwise, I think we should change the map used in Scope (which is based on Lib_name) to consider more things besides the library name when finding items in it.

You need to change the map, but it's enough to just make it list valued. Right now it just returns a single `Found_or_redirect.t` per name, but we should be able to return multiple items here and then filter the list based on the enabled_if value.

Adding specialized fields would make the whole thing more confusing.

jchavarri:
I'm leaving here some notes, as I have very limited idea of what i'm doing, so maybe they can serve to surface if I'm going in the wrong direction: 

- Originally tried to start passing `context` to the `resolve` field in `Lib.t`. This turned out to touch a lot of places, so I suspected it would not be the right way to go, and reverted it
- Found out there's a way to know if a library is enabled using `Lib_info.enabled info`, so I'm filtering out the results in `create_db_from_stanzas` `map` value to only consider the libraries that are enabled.

The second approach seems to be working out, I had to disable some odoc rules due to some internal errors:

```
Description:
  ("[gen_rules] returned rules in a directory that is not a descendant of the directory it was called for",
  { dir = In_build_dir "alt-context/a"
  ; example =
      Alias
        { dir = In_build_dir "alt-context/b/.foo.objs/byte"
        ; name = ".odoc-all"
        }
  })
```

I will continue down this path tomorrow, fixing the odoc rules and any other potential breakage after the change.

rgrinberg:
> The second approach seems to be working out, I had to disable some odoc rules due to some internal errors:

It will work out, but it will make the errors worse. Our rules distinguish the following two errors:
1. A library name cannot be resolved because it's not found
2. A library name cannot be resolved because it's not available

Your change would convert all errors in type 2. to type 1.

A better approach is the following:

1. Allow `Lib.DB.create` to return more than one library per name when the libraries have enabled_if on them (requires modifying `Lib.Resolve_result.t`)
2. before we call `Lib.instantiate_impl`, we evaluate the enabled_if of every library that corresponds to our name.
3. If we find that only one `enabled_if` is true, then we proceed as usual. Otherwise, we complain that more than one library is available.

For inspiration, you can take a look at `artifcats.ml` where I handle `enabled_if` correctly but for binaries

jchavarri:
@rgrinberg Thanks for the pointers, they are super helpful.

I made some progress on the new handling of lib dbs following the steps mentioned (see https://github.com/ocaml/dune/pull/9839/commits/1b0c2c33f9c92cf51c7d345d6112927f0fb4f870).

Also fixed a cryptic runtime error coming from some `Option.value_exn` usage (wasn't able to figure out where it was coming from), by moving `Odoc.setup_private_library_doc_alias` to be called when we know the library is enabled (https://github.com/ocaml/dune/pull/9839/commits/857ec39ad9c22a57abc27f5eb29acf8efec8eade). I'm not 100% sure about this, as there was a comment in the `main` branch, so probably it was kept that way for some reason:

https://github.com/ocaml/dune/blob/c11c1f12e6f52860ad171188a139c78ec5167d64/src/dune_rules/gen_rules.ml#L105-L106

Tomorrow, I'll try to figure out why the `.odoc-all` alias triggers a `returned rules in a directory that is not a descendant of the directory it was called for` error, it seems the rules for odoc get added to libraries that should be disabled. I saw that the code in `Lib_rules.rules` doesn't use `resolve_result`, but rather `Lib.Status.t` for some reason, so I might have to look into adapting that type as well to the new situation.

There are also some errors that have regressed, from `is hidden (unsatisfied 'enabled_if')` to `not found`. I will look at that later, should be easy to fix 🤞

rgrinberg:
I've nothing useful to say about odoc issues. It's been a while since I've looked at that stuff. Seems like my comment
>  (* XXX why are we setting up private doc rules for disabled libraries? *) 
is giving us a useful hint. I'd push `Odoc.setup_private_library_doc_alias` into the subsequent `if_available_buildable` and see what happens. This can come in a separate PR.

> I saw that the code in Lib_rules.rules doesn't use resolve_result, but rather Lib.Status.t for some reason, so I might have to look into adapting that type as well to the new situation.

`Lib.Status.t` doesn't need to change. Because whoever is constructing the status (`resolve_name`), should evaluate the `enabled_if` and error out (by constructing an appropriate `Status.Invalid`) if more than one evaluates to true.

jchavarri:
I have been looking into a weird issue. When I defined a test using a public library using the code in this PR:

```
Show behavior of doc-private alias when it's part of an unavailable library

  $ cat > dune-project << EOF
  > (lang dune 3.13)
  > (package (name foo))
  > EOF

  $ cat > dune << EOF
  > (library
  >  (name bar)
  >  (public_name foo.bar)
  >  (enabled_if (= %{context_name} "unavailable")))
  > EOF
  $ cat > foo.ml <<EOF
  > let x = "foo"
  > EOF

  $ dune build
```

The `dune build` command never ends, because `resolve_name` function enters an endless loop. Looking at the logic in `db.parent`, I am not sure I understand how it is supposed to work.

There are two usages of `Lib.DB.create`:
- in `create_db_from_stanzas` the parent is `public_libs`:
https://github.com/ocaml/dune/blob/c11c1f12e6f52860ad171188a139c78ec5167d64/src/dune_rules/scope.ml#L249
- in `public_libs` the parent is `installed_libs`:
https://github.com/ocaml/dune/blob/c11c1f12e6f52860ad171188a139c78ec5167d64/src/dune_rules/scope.ml#L182

What exactly is `db.parent` and is there some invariant where it should reflect the parent-child relationship between folders or some other kind of hierarchy? If that's the case, I am not entirely sure it does so at the moment.

jchavarri:
Unrelated to the comment above (sorry for the unordered stream of comments).

I have some suspicions that this calculation of `enabled_if` is not correct:

https://github.com/ocaml/dune/blob/21557d8d0b68650d4997c5ec94a1a782a2847a87/src/dune_rules/gen_rules.ml#L117

The reason is that it does not take the current context into account, so in the test that is included in this PR, it always returns `true`, even for libraries that should not be enabled under that context. I think this is what is causing the `[gen_rules] returned rules in a directory that is not a descendant of the directory it was called for` errors in the test.

The reason is that in the case of the test, where there are two libraries with the same name (one for each context) there will always exist some library available with the given name. It seems the accurate way to check if a library is enabled is to get that information using `Lib_info.enabled`?

jchavarri:
In https://github.com/ocaml/dune/pull/9839/commits/548fabd70d13d9e68d6fc9c851335e42dcda6694, I updated the test to show the issue I've been trying to figure out for the last few days.

rgrinberg:
> The reason is that it does not take the current context into account, so in the test that is included in this PR, it always returns true, even for libraries that should not be enabled under that context

A library database is context specific, so it will should account for the when evaluating the availability. See the call to `Library.to_lib_info` in `scope.ml` to confirm.

rgrinberg:
Pushed a fix. The following check for availability was no longer correct:

```
let* enabled_if = Lib.DB.available (Scope.libs scope) (Library.best_name lib) in
```

Using your test as an example:

* We are checking the availability of the name `foo` before building a/foo or b/foo
* The name `foo` is always available in every context because one of the stanzas will evaluate to true (either a/foo or b/foo)
* So both stanzas are always going to be generated in every context.

The solution is to make the availability check more specific. Instead of checking if the name `foo` is available, we need to check if `foo` is available *and* it resolves to the stanza which we are generating the rules for.

The simplex fix I did addresses the problem, but we should improve the API to make it harder to make this error. E.g.

```
val available_library_stanza : Lib.DB.t -> dir:Path.Build.t -> Lib_name.t -> bool Memo.t
```

src/dune_rules/lib.mli
```diff
@@ -123,8 +123,16 @@ module DB : sig
...
   val find_even_when_hidden : t -> Lib_name.t -> lib option Memo.t
+
+  (** [available db lib_name] returns [true] if there is any library available
+      in the database with name [lib_name], regardless if it is enabled or not. *)
   val available : t -> Lib_name.t -> bool Memo.t
 
+  (** [available_library_stanza db ~dir lib_name], similar to [available], but only
+      returns [true] if there is a library available with name [lib_name], and the stanza
+      directory where the library is defined matches [dir]. *)
+  val available_library_stanza : t -> dir:Path.Build.t -> Lib_name.t -> bool Memo.t
```
rgrinberg, line 134: `Lib.DB.get_compile_info` needs to have this `~dir` argument as well. Otherwise we might return the compilation information of the wrong library.

jchavarri:
@rgrinberg I hit another issue with the current approach we're taking. The problem is that now we are checking for duplicated libraries at DB read time (`Lib.DB.find`) rather than DB creation time as before. You can see how the last test in `eif-library-name-collision.t` to check behavior for duplicated libs in the same context now doesn't error out:

https://github.com/jchavarri/dune/blob/e53cbdf988de089a841fde175d20fbfb38fca18e/test/blackbox-tests/test-cases/enabled_if/eif-library-name-collision.t#L62-L74

With the way DB reads are being handled, `Invalid` status is just ignored, e.g in `find`:

https://github.com/ocaml/dune/blob/9ea459016edcd01471ff27a5d75e142d813cc06d/src/dune_rules/lib.ml#L1846

I understand why is it this way, because these errors would have been caught before in the past (in `create_db_from_stanzas`). I guess I should try to handle these `Invalid` status errors at read time now? Do you think this will be too invasive?

rgrinberg:
We can have two types of duplicated library errors:

1. A group of libraries have the same name but one of these libraries does not have an enabled_if. This should be an error at database construction time.

2. A group of libraries have the same name, they all have an `enabled_if` and more than one of them evaluates to true when we look up  the library. In such a case, we should immediately error on library instantiation and not bother returning anything to the user or delaying the error. Since there's an issue with the dune file that needs to be fixed.

src/dune_rules/lib.mli
jchavarri: I tried to get something done about this in https://github.com/ocaml/dune/pull/9839/commits/28585d63a9dcd23d5ceb1f4fe096faea60fd174a.

jchavarri:
@rgrinberg I think I got the PR to a point where the feature and error handling seems to be working, as far as I can tell. Regarding my last comment about error handling, I realized the test only had two private libraries, that's why the error was never trigger. As soon as there is a consumer of the libraries, the error shows, as the test now reflects.

If there are any more tests to add or refactors, I am happy to tackle them. Thanks!

jchavarri:
Hm there seems to be some issue with the recent change to update `Lib.DB.get_compile_info` to take the lib `dir` as input. When we generate the rules for melange emit stanza, we iterate over all the libraries:

https://github.com/ocaml/dune/blob/109e5f56c0652b1994030453af720cc83a32b9b6/src/dune_rules/melange/melange_rules.ml#L426-L431

At that point, `dir` is not the library dir but rather the lib where the `melange.emit` stanza is declared. Not sure if the fix it to avoid using `get_compile_info` or getting the lib dir from somewhere 🤔

rgrinberg:
I think you can just add a version of `get_compile_info` that takes `Lib.t` instead of the name + path. That should work for both melange and the regular library rules.

jchavarri:
I have added another test file in https://github.com/ocaml/dune/pull/9839/commits/4f1b4a71e415bad881f251ae0670d12cde5d3cc2 for the "ultimate" use case I had in mind: defining two libraries with the same name in the same `dune` file, and use `enabled_if` to have them be added to separate contexts. 

I realized that at the moment this use case is not possible because of the way `Ml_sources.Modules.t` defines the map of libraries:

https://github.com/ocaml/dune/blob/339f529cb5e9664588802ab4dfa2d7de2f490bfb/src/dune_rules/ml_sources.ml#L66

It creates a map using `Lib_name.t` as key, which makes it impossible to satisfy the use case I have in mind. I am not sure if this is something that could be improved, nor if it should be part of this PR or a new one?

jchavarri:
Otherwise, I think the PR is just missing the patch from https://github.com/ocaml/dune/pull/10010, once that's merged, all tests should pass fine.

src/dune_rules/lib.ml
```diff
@@ -1894,11 +1976,38 @@ module DB = struct
...
+    >>= function
+    | false -> Memo.return false
+    | true ->
+      find t name
+      >>| (function
+       | None -> assert false
+       | Some lib ->
+         (match Local.of_lib lib with
+          | None -> false
+          | Some info -> Path.Build.equal dir (Lib_info.src_dir (Local.info info))))
```
jchavarri, line 1991: This works at the moment because the code assumes two libs with the same name must exist in different folders, but this might not be necessarily true in the future, see related comment: https://github.com/ocaml/dune/pull/9839#issuecomment-1941386688

rgrinberg:
I suppose we could do something a bit more reliable than the directory. The triple `Loc.t * Path.Build.t * Lib_name.t` should be enough. We could call this triple `Library_id.t` and then just use it uniformly everywhere.

jchavarri:
> The triple `Loc.t * Path.Build.t * Lib_name.t` should be enough.

I have instead taken a different approach. I started thinking why does `Ml_sources.make` gets called at all for a library that is disabled? So in the [last commit](https://github.com/ocaml/dune/pull/9839/commits/7f19d29cd5fc62ab6c79130253c190c3e0971cee), I explored just disabling all the rules added in `Lib_rules.rules` if the library is not enabled for that context, including merlin rules.

It seems to have worked, as the test now passes successfully. The only remaining part is to fix the case for the duplication of library names in the same folder, under the same context, where `get_compile_info` is called now for libraries that don't really exist. If this approach makes sense, I will continue working in this direction tmr.

rgrinberg:
What happens when you do:

```
$ dune build %{cmo:foo}
```

When `Foo` is part of a disabled library? Ideally there should be a decent error instead of just "module doesn't exist".

Probably worth adding a test for this case in a separate PR.

jchavarri:
@rgrinberg I added a test for the case you mentioned in #10026.

Re: the issues with `get_compile_info` failing, here is what I understand is happening.

Let's say there are two contexts `default` and `alt`, and a library `foo` that is only defined in `default` context through the `enabled_if` field.

`Gen_rules.of_stanza` will still call `Lib_rules.rules` for the stanza that defines `foo` under `alt` context, and this is where currently `get_compile_info` is breaking, because it will not be able to find a library.

I am not sure what `Gen_rules.of_stanza` is supposed to do when a stanza is disabled for the current context. The function is supposed to return some compilation context + merlin configuration, but what's the point if the stanza is disabled altogether? Shouldn't Dune just don't add any rules in that case and skip the processing of the stanza altogether?

src/dune_rules/lib_rules.ml
```diff
@@ -570,86 +571,112 @@ let library_rules
...
-      ~flags
-      ~modules
-      ~source_dirs:Path.Source.Set.empty
-      ~preprocess:(Preprocess.Per_module.without_instrumentation lib.buildable.preprocess)
-      ~libname:(Some (snd lib.name))
-      ~obj_dir
-      ~dialects:(Dune_project.dialects (Scope.project scope))
-      ~ident:(Lib.Compile.merlin_ident compile_info)
-      ~modes:(`Lib (Lib_info.modes lib_info)) )
+  match enabled with
```
jchavarri, line 587: This check might look redundant, but if we don't conditionally return merlin config, the `eif-library-name-collision-same-folder` test fails with multiple rules for cmxs:

```
Error: Multiple rules generated for _build/alt-context/foo.cmxs:
- dune:4
- dune:1
Error: Multiple rules generated for _build/default/foo.cmxs:
- dune:4
- dune:1
```

src/dune_rules/lib.ml
```diff
@@ -2092,39 +2180,3 @@ let to_dune_lib
   in
   Dune_package.Lib.of_dune_lib ~info ~main_module_name
 ;;
-
-module Local : sig
```
jchavarri, line 2096: This module was just moved up to make `Local` available in scope for `get_compile_info`.

jchavarri (commented):
@rgrinberg I think I got things to be in a place where all the use cases I wanted to support are working, and nothing existing seems to have regressed (at least from the functionality being tested).

I added some inline notes to help reviewing, the diff might look less noisy [without the whitespace](https://github.com/ocaml/dune/pull/9839/files?diff=unified&w=1).

test/blackbox-tests/test-cases/enabled_if/eif-library-name-collision-same-folder.t
```diff
@@ -0,0 +1,97 @@
+Showcase that using same library name in two workspaces inside the same folder
```
moyodiallo, line 1:
```suggestion
Showcase that using same library name in two contexts inside the same folder
```

test/blackbox-tests/test-cases/enabled_if/eif-library-name-collision.t
```diff
@@ -0,0 +1,103 @@
+Showcase that using same library name in two workspaces is possible for private libraries
```
moyodiallo, line 1:
```suggestion
Showcase that using same library name in two contexts is possible for private libraries
```

rgrinberg:
@anmonteiro could you review this one?

src/dune_rules/ml_sources.ml
```diff
@@ -389,29 +389,43 @@ let modules_of_stanzas =
...
-            ~modules
-            ~lib
-            ~include_subdirs
-            ~version:lib.dune_version
-          >>= Resolve.read_memo
-        in
-        let obj_dir = Library.obj_dir lib ~dir in
-        `Library (lib, sources, modules, obj_dir)
+        let* lib_config = lib_config in
+        let info = Library.to_lib_info lib ~dir ~lib_config in
```
anmonteiro, line 405:
```suggestion
        let* info =
          let+ lib_config = lib_config in
          Library.to_lib_info lib ~dir ~lib_config
        in
```

src/dune_rules/ml_sources.ml
```diff
@@ -389,29 +389,43 @@ let modules_of_stanzas =
...
+        let* enabled_if = Lib_info.enabled info in
```
anmonteiro, line 406: refactoring suggestion:

- rename `Lib_info.enabled` to `Lib_info.to_enabled_if`
- add `Lib_info.enabled` which is `val enabled: Lib_info.t -> bool Memo.t`
  - implementation is your conversion from the variant to bool (saw that a few times in the PR)

src/dune_rules/lib_rules.ml
```diff
@@ -570,86 +571,112 @@ let library_rules
...
-  in
-  let* () =
-    Memo.Option.iter vimpl ~f:(Virtual_rules.setup_copy_rules_for_impl ~sctx ~dir)
-  in
-  let* () = Check_rules.add_cycle_check sctx ~dir top_sorted_modules in
-  let* () = gen_wrapped_compat_modules lib cctx
-  and* () = Module_compilation.build_all cctx
-  and* expander = Super_context.expander sctx ~dir
-  and* lib_info =
+  let* lib_info =
```
anmonteiro, line 574: are there unpredictable consequences of getting the lib info before doing the cycle check on its modules?

src/dune_rules/gen_rules.ml
```diff
@@ -111,14 +111,11 @@ end = struct
       let+ () = Toplevel.Stanza.setup ~sctx ~dir ~toplevel in
       empty_none
     | Library.T lib ->
-      let* enabled_if = Lib.DB.available (Scope.libs scope) (Library.best_name lib) in
-      if_available_buildable
-        ~loc:lib.buildable.loc
-        (fun () ->
-          let+ () = Odoc.setup_private_library_doc_alias sctx ~scope ~dir:ctx_dir lib
```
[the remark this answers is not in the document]
jchavarri: Extracted this part to https://github.com/ocaml/dune/pull/10161.

src/dune_rules/lib.ml
```diff
@@ -1896,15 +1978,21 @@ module DB = struct
 
   let available t name = Resolve_names.available_internal t name
 
-  let get_compile_info t ~allow_overlaps name =
+  let get_compile_info t ~allow_overlaps ~dir name =
     let open Memo.O in
-    find_even_when_hidden t name
```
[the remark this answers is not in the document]
jchavarri: If we don't skip hidden libs, then a bunch of tests will fail with this error:

```
  Description:
    ("modules_and_obj_dir: failed lookup", { keys = []; for_ = Library "foo" })
```

The errors seem to come from the usage in `Ml_sources.modules`, I have started a separate PR to discuss this removal: https://github.com/ocaml/dune/pull/10171

test/blackbox-tests/test-cases/enabled_if/eif-library-name-collision.t
```diff
@@ -0,0 +1,103 @@
...
+
+  $ cat > b/dune << EOF
+  > (library
+  >  (name foo)
+  >  (public_name baz.foo)
+  >  (enabled_if (= %{context_name} "alt-context")))
+  > EOF
+
+  $ dune build
+  Error: Library foo is defined twice:
```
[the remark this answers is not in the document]
jchavarri: I have started an attempt in https://github.com/ocaml/dune/pull/9839/commits/619ea9dfccde0057dad3eaf0ee34346ec480e42d.

This doesn't work yet, as it seems the result from querying the DB to find a public library is ignored. There's a call to `find` here:

https://github.com/ocaml/dune/blob/98c576e643450659e2893ced9ae8e076b6129d95/src/dune_rules/scope.ml#L392

But this function already skips all invalid results to just return `None`:

https://github.com/ocaml/dune/blob/98c576e643450659e2893ced9ae8e076b6129d95/src/dune_rules/lib.ml#L1866

And then back in the call point, there's a comment that shows that somehow this should be dealt with:

https://github.com/ocaml/dune/blob/98c576e643450659e2893ced9ae8e076b6129d95/src/dune_rules/scope.ml#L395-L397

It seems if the result is `Invalid` (because there are 2 libraries defined) we should do something else?

src/dune_rules/scope.ml
```diff
@@ -63,13 +63,15 @@ module DB = struct
...
+          Library.best_name conf, Found_or_redirect.found [ info ])
       |> Lib_name.Map.of_list_reducei ~f:(fun name (v1 : Found_or_redirect.t) v2 ->
         let res =
           match v1, v2 with
-          | Found info1, Found info2 -> Error (Lib_info.loc info1, Lib_info.loc info2)
+          | Found info1, Found info2 ->
+            Ok (Found_or_redirect.found (List.rev_append info1 info2))
           | Found info, Redirect (loc, _) | Redirect (loc, _), Found info ->
-            Error (loc, Lib_info.loc info)
+            (* todo: should this not be an error? *)
```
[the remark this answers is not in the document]
jchavarri: there's some `Nonempty_list` module. Maybe I could add it eventually.

src/dune_rules/lib.ml
```diff
@@ -1130,19 +1142,53 @@ end = struct
...
     | Redirect (db', (_, name')) -> find_internal db' name'
-    | Found info -> instantiate db name info ~hidden:None
+    | Found libs ->
+      (match libs with
+       | [] | _ :: [] ->
+         (* In case we have 0 or 1 results found, convert to [Status.t] directly.
+            This allows to provide better errors later on,
+            e.g. `Library "foo" in _build/default is hidden (unsatisfied 'enabled_if') *)
+         to_status ~db ~name libs
+       | _ :: _ :: _ ->
```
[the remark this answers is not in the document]
jchavarri: This is what I tried to reflect in the inline comments. Would you change / improve anything to make it clearer?

src/dune_rules/lib.ml
```diff
@@ -132,6 +132,18 @@ module Error = struct
       ]
   ;;
 
+  let duplicated ~loc ~name ~dir_a ~dir_b =
```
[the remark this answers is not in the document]
jchavarri: We'll need it for public libs as well (see failed attempt in https://github.com/ocaml/dune/pull/9839/commits/619ea9dfccde0057dad3eaf0ee34346ec480e42d).

test/blackbox-tests/test-cases/enabled_if/eif-library-name-collision-same-folder.t
```diff
@@ -0,0 +1,97 @@
...
+  $ cat > dune << EOF
+  > (library
+  >  (name foo))
+  > (library
+  >  (name foo))
+  > EOF
+
+If no public lib is available, the build finishes fine as there are no consumers of the libraries
+
+  $ dune build
```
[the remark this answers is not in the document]
jchavarri: I agree it's strange. I still not fully understand why this happens yet.

jchavarri:
Thanks for the reviews and guidance. This PR got a bit long and messy, it has served its purpose as many parts of it were extracted into separate PRs.

Starting again with a much smaller diff (for now) in https://github.com/ocaml/dune/pull/10179.

PR #9845, opened 2024-01-26
fix(coq): delay boot type detection

### LINKED ISSUE #9818: regression on missing Coq
## Expected Behavior

Consider the following toy demo repository:

https://github.com/vzaliva/coq-dune-examlpe

In an opam switch **without** Coq installed, with dune 3.12.2 it builds without errors using following command:

`dune build -p example-ocaml`

(this is expected)

## Actual Behavior

However, with dune 3.13.0 it gives build error:

```
$ dune build -p example-ocaml
Error: Couldn't find Coq standard library, and theory is not using (stdlib
no)
```

(this is a regression)

It is related to the following bug https://github.com/ocaml/dune/issues/8099
See the discussion there.

## Specifications
```
$ dune --version
3.13.0
$ ocaml --version
The OCaml toplevel, version 4.13.1
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 23.04
Release:	23.04
Codename:	lunar
```
## Additional information 

```
dune build -p example-ocaml --verbose
Shared cache: disabled
Shared cache location: /home/lord/.cache/dune/db
Workspace root: /home/lord/src/coq-dune-examlpe
Auto-detected concurrency: 12
Dune context:
 { name = "default"
 ; kind = "default"
 ; profile = Release
 ; merlin = true
 ; fdo_target_exe = None
 ; installed_env =
     map
       { "INSIDE_DUNE" : "/home/lord/src/coq-dune-examlpe/_build/default"
       ; "OCAML_COLOR" : "always"
       ; "OPAMCOLOR" : "always"
       }
 ; instrument_with = []
 }
Actual targets:
- recursive alias @install
Error: Couldn't find Coq standard library, and theory is not using (stdlib
no)
```

rgrinberg:

Fixes #9818

### DIFF
--- a/src/dune_rules/coq/coq_rules.ml
+++ b/src/dune_rules/coq/coq_rules.ml
@@ -200,15 +200,15 @@ module Bootstrap : sig
     | Stdlib of Coq_lib.t (** Regular case in >= 0.8 (or in < 0.8
                               (boot) was used *)
 
-  val empty : t Memo.t
+  val empty : t
 
   val make
-    :  dir:Path.Build.t
+    :  scope:Scope.t
     -> use_stdlib:bool
     -> wrapper_name:string
     -> coq_lang_version:Syntax.Version.t
     -> Coq_module.t
-    -> t Memo.t
+    -> t Resolve.Memo.t
 
   val flags : t -> 'a Command.Args.t
 end = struct
@@ -222,7 +222,7 @@ end = struct
 
   (* For empty set of modules, we return Prelude which is kinda
      conservative. *)
-  let empty = Memo.return No_stdlib
+  let empty = No_stdlib
 
   (* Hack to know if a module is a prelude module *)
   let check_init_module bootlib wrapper_name coq_module =
@@ -235,12 +235,12 @@ end = struct
 
   (* [Bootstrap.t] determines, for a concrete Coq module, how the Coq
      "standard library" is being handled. See the main modes above. *)
-  let make ~dir ~use_stdlib ~wrapper_name ~coq_lang_version coq_module =
-    let* scope = Scope.DB.find_by_dir dir in
-    let+ boot_lib =
+  let make ~scope ~use_stdlib ~wrapper_name ~coq_lang_version coq_module =
+    let open Resolve.Memo.O in
+    let* boot_lib =
       Scope.coq_libs scope
+      |> Resolve.Memo.lift_memo
       >>= Coq_lib.DB.resolve_boot ~coq_lang_version
-      |> Resolve.Memo.read_memo
     in
     if use_stdlib
     then (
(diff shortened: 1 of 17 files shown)


src/dune_rules/coq/coq_rules.ml
```diff
@@ -235,25 +235,29 @@ end = struct
...
-  let make ~dir ~use_stdlib ~wrapper_name ~coq_lang_version coq_module =
-    let* scope = Scope.DB.find_by_dir dir in
-    let+ boot_lib =
+  let make ~scope ~use_stdlib ~wrapper_name ~coq_lang_version coq_module =
+    let open Resolve.Memo.O in
+    let* boot_lib =
       Scope.coq_libs scope
+      |> Resolve.Memo.lift_memo
       >>= Coq_lib.DB.resolve_boot ~coq_lang_version
-      |> Resolve.Memo.read_memo
```
rgrinberg, line 243: This is a no-no because it break the rules in the entire directory if the library isn't found.

src/dune_rules/coq/coq_rules.ml
```diff
@@ -235,25 +235,29 @@ end = struct
...
       match boot_lib with
       | None ->
         if coq_lang_version >= (0, 8)
         then
-          User_error.raise
-            [ Pp.text
-                "Couldn't find Coq standard library, and theory is not using (stdlib no)"
-            ]
-        else Implicit
+          Resolve.Memo.fail
```
rgrinberg, line 251: Instead of raising, we save the error.

src/dune_rules/coq/coq_rules.ml
```diff
@@ -289,7 +293,8 @@ let coqc_file_flags ~dir ~theories_deps ~wrapper_name ~boot_type ~ml_flags
     ; A wrapper_name
     ]
   in
-  [ Bootstrap.flags boot_type; S file_flags ]
+  let boot_flags = Resolve.Memo.read boot_type |> Action_builder.map ~f:Bootstrap.flags in
+  [ Dyn boot_flags; S file_flags ]
```
rgrinberg, line 297: And here we resurface the error, but this time only when we try to execute the relevant rule that needs this.

emillon:
FTR, this does fix #9818.

ejgallego (approved):
Thanks a lot @rgrinberg !

It would be good if we try to push the test case, however I'm unsure how can we make a test such that we ensure `coqc` is not available, maybe we could mock `coqc` to fail, but that's not exactly the same.

emillon:
@ejgallego you can make a `path` directory with just `dune` and `ocamlc` copied from the PATH, and set `PATH` to just that directory. See https://github.com/ocaml/dune/blob/3.13.0/test/blackbox-tests/test-cases/melange/missing-melc.t for example.

ejgallego:
Thanks @emillon ! Will add this test, as this seems like an easy to get wrong case.

@rgrinberg , PR looks ready for merging to me, however a tweak on Coq's test-suite is needed due to the output being more verbose now on error.

rgrinberg:
Sure, feel free to update the output if it looks good to you. I don't have coq installed at the moment.

PR #9861, opened 2024-01-30
fix(boot): sort readdir output for reproducibility

### LINKED ISSUE #9794: reproducible builds: filesystem readdir order
While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds), I found that our ocaml-dune package varies through non-determinism from filesystem readdir order.

## Expected Behavior
dune build should be reproducible

## Actual Behavior

```diff
--- strings RPMS.1/usr/bin/dune
+++ strings RPMS.2/usr/bin/dune
[...]
 padding.0
-sha1_do_chunk
 sha512_do_chunk
+sha1_do_chunk
 scanmult
 shutdown_happened
```

Maybe related: ocaml-dune-3.12.2/_boot/source_files produced by `./.duneboot.exe --verbose` also has order variations.

## Reproduction

build on two different filesystems or use disorderfs or randomize readdir order

## Specifications

- Version of `dune` (output of `dune --version`): ocaml-dune-3.12.2 with #9735
- Version of `ocaml` (output of `ocamlc --version`) ocaml-4.14.1
- Operating system (distribution and version): openSUSE Tumbleweed

emillon:
Fixes #9794

`_boot/dune.exe` is installed, so it needs to be reproducible. This
change ensures that the source files are scanned in an order that is
independent from the underlying directory entries.

This has been tested with `disorderfs --shuffle-dirents=yes`: two runs
of `make bootstrap` create the same binary.

Signed-off-by: Etienne Millon <me@emillon.org>

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -102,12 +102,12 @@ module Status_line = struct
   let () = at_exit (fun () -> Printf.printf "\r%*s\r" (String.length !displayed) "")
 end
 
-(* Return list of entries in [path] as [path/entry] *)
+(* Return a sorted list of entries in [path] as [path/entry] *)
 let readdir path =
-  Array.fold_right
-    ~f:(fun entry dir -> (path ^/ entry) :: dir)
-    ~init:[]
-    (Sys.readdir path)
+  Sys.readdir path
+  |> Array.to_list
+  |> List.map ~f:(fun entry -> path ^/ entry)
+  |> List.sort ~cmp:String.compare
 ;;
 
 let open_out file =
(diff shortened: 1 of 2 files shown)


bmwiedemann (approved):
With #9735 this produced bit-identical binaries in my tests for openSUSE.

Leonidas-from-XIV (approved):
Given this is `duneboot` I assume the performance impact of adding a second iteration over the entries plus the sorting is afterwards is not a hot path.

TIL: `disorderfs` - cool tool.

emillon:
Yes, this is only used to build dune itself.

PR #9863, opened 2024-01-30
doc: add "How to Install Dune"

emillon:
This was buried in the quickstart document (actually a collection of howtos). This is intended to be the central place that documents opam env, etc, so the other "if this does not work" paragraphs are removed.


doc/howto/install-dune.rst
```diff
@@ -0,0 +1,39 @@
...
+
+.. code:: console
+
+  $ dune --version
+  3.12.1
+
+.. note::
+
+   In most cases, when using Opam you will not need to install Dune by hand.
+   Installing the project's dependencies will install it in the Opam switch.
```
Leonidas-from-XIV, line 39:
```suggestion
   Installing the projects dependencies will install it in the Opam switch.
```
I think, since it is the possessive form.

doc/howto/install-dune.rst
```diff
@@ -0,0 +1,39 @@
...
+===================
+
+Dune is available as an Opam package. First, make sure that Opam is installed:
+
+  .. code:: console
+
+   $ opam --version
+   2.1.5
+
+Any version greater than 2.0.0 is supported, though at least 2.1.0 is preferred.
```
Leonidas-from-XIV, line 11: To me these read a bit better. However, newer vs greater in versions is a bit of a grey zone, since "newer" reads better but "greater" is technically more accurate (2.0.45 might be newer than 2.1.0 after all), but greater sounds a bit like a value assesment. @christinerose what would you say, I'm honestly curious!

```suggestion
Any version newer than 2.0.0 is supported, though preferably at least 2.1.0.
```

or 

```suggestion
Any version newer than 2.0.0 is supported, though for the best experience at least 2.1.0 is recommended.
```

Leonidas-from-XIV (commented):
I think this is a very sensible reorganisation. The troubleshooting notes at the end of the quick start seemed bolted on and would make more sense in a separate document about troubleshooting.

(Also, recommending to install `ocp-indent` on issues seems extremely dubious advice)

doc/howto/install-dune.rst
emillon: "higher"?

Leonidas-from-XIV: Yes, that's perfect!

doc/howto/install-dune.rst
emillon: That's the dependencies of the project, so the project's dependencies? (there's only one project here)

Leonidas-from-XIV (approved):

PR #9865, opened 2024-01-30
doc: the Dune Mental Model

emillon:


doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+alias (``dune runtest`` is a shortcut for ``dune build @runtest``).
+
+To build the target *t*, Dune will first check is *t* is in the source tree. In
+that case, it will copy it to the ``_build`` directory. Otherwise, it will
+check if it is the target of a rule *r*. In that case, it will build all
+dependencies of *r* and if any dependency has changed, it will run the action
+of *t*. (Dune will raise an error in other cases: if *t* is both in the source
+tree and the target of a rule, or if it is neither)
+
+In the case of our example, if we call ``dune runtest``, Dune will consider all
```
emillon, line 135: this paragraph is a mouthful, but a bit on purpose. I wonder if it could work as a nested bullet list, but this feels too technical

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+
+- one rule to compile the ``tool.exe`` executable:
+
+  - it will read its dependencies: ``main.cmx`` and ``config.cmx``
+  - and write its output: ``tool.exe``
+  - using an action: ``(run ocamlopt -o tool.exe main.cmx config.cmx``)
+
+Note that in this example, some files are targets of a rule and dependencies of
+another (``.cmx`` files). We are unlikely to ever interact with them directly,
+so it can also be useful to think of the ``(executable)`` stanza as acting as a
```
snowleopard, line 91: I'm curious why you find it useful to think of this as just one rule. From the point of view of Dune's performance/cache model, it's pretty important that these rules are separate. (On the other hand, bundling these separate rules into one is sometimes appropriate and internally we have some optimisations that are actually working this way.)

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+big rule with ``main.ml`` and ``config.ml`` as inputs.
+
+Running the Tests
+-----------------
+
+Some rules do not produce any output file, but we're still interested in
+running an action. A test is a good example: we want the build process to exit
```
snowleopard, line 98:
```suggestion
running their actions. A test is a good example: we want the build process to exit
```

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+   (deps tool.exe testdata.txt)
+   (alias runtest)
+   (action
+    (run tool.exe testdata.txt)))
+
+This rule:
+
+- reads its dependencies: ``tool.exe`` and ``testdata.txt``
+- writes no targets
+- by running its action: ``(run tool.exe testdata.txt)``
```
snowleopard, line 115: "by running" reads odd here, maybe "and runs"?

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+- (but it is attached to ``runtest``)
```
snowleopard, line 116:
```suggestion
- (and it is attached to ``runtest``)
```

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+big rule with ``main.ml`` and ``config.ml`` as inputs.
+
+Running the Tests
+-----------------
+
+Some rules do not produce any output file, but we're still interested in
+running an action. A test is a good example: we want the build process to exit
+with an error code if the action fails.
+In that case, the rule does not have targets, but we "attach" it to an
+:term:`alias`, ``runtest`` in this case.
```
snowleopard, line 101: Maybe it's worth saying why we attach such actions to aliases: if we didn't, there would be no way to force executing them, since they have no targets. We could also remind the readers here that Dune is lazy: it only executes actions that are required to build the requested build goals.

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+- by running its action: ``(run tool.exe testdata.txt)``
+- (but it is attached to ``runtest``)
+
+How Dune Interprets Rules
+-------------------------
+
+Now that we have seen how Dune sets up rules for a project, let's see how it
+executes a ``dune build`` command.
+
+Dune can build targets: either files, or aliases. ``dune build tool.exe`` will
```
snowleopard, line 124:
```suggestion
Dune can build *goals*: either files or aliases. ``dune build tool.exe`` will
```
I think "goal" is the usual term to denote things that, e.g. can appear on the command line.

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+build ``tool.exe``, while ``dune build @runtest`` will build the ``runtest``
+alias (``dune runtest`` is a shortcut for ``dune build @runtest``).
+
+To build the target *t*, Dune will first check is *t* is in the source tree. In
+that case, it will copy it to the ``_build`` directory. Otherwise, it will
```
snowleopard, line 129: I think "it will copy it to the `_build` directory" may be leaking some unnecessary implementation details. Without thinking deeply about how Dune builds are working, copying sources into the `_build` directory might seem unnecessary/confusing. In fact, internally, this will soon not be true in some cases.

snowleopard: So, maybe, "In this case, there is nothing to do" instead?

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+check if it is the target of a rule *r*. In that case, it will build all
+dependencies of *r* and if any dependency has changed, it will run the action
```
snowleopard, line 131: > and if any dependency has changed

This is a bit vague (changed since when?). I find it easier to separate caching into a completely separate concern, which only applies to executing actions. Like this:

When Dune is about to execute an action, it checks (in various caches) if it executed it before on the same set of dependencies, and, if yes, it can skip executing it and reuse the previous result.

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+of *t*. (Dune will raise an error in other cases: if *t* is both in the source
+tree and the target of a rule, or if it is neither)
+
+In the case of our example, if we call ``dune runtest``, Dune will consider all
+rules attached to the ``runtest`` alias. This is just the integration test
+rule. It needs to build its dependencies, ``tool.exe`` and ``testdata.txt``.
+The latter is present in the source tree, so it just needs to be copied.
```
snowleopard, line 138: > so it just needs to be copied

Again, I think this detail can be skipped.

doc/explanation/mental-model.rst
snowleopard: I've found this paragraph too mechanistic and not very insightful. Maybe we can even skip it? Or, hide it and include the bullet list into a hidden section if anyone would like to step though it.

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
+The Dune Mental Model
```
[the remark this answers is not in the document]
gridbugs: I like "The Dune Mental Model".

doc/explanation/mental-model.rst
gridbugs: Agreed that it would be helpful to explain why the alias is necessary. My question reading this part is: Since dune doesn't run actions from rules whose targets all exist, and this rule has no targets, why does it ever run actions of rules with no targets? Is this a special case for rules with no targets?

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+.. note::
+
+   This document is a simplification of the reality: the actual rules might be
+   different, it does not touch rule loading and glosses over how caching
+   works, but should be a useful tool to build an understanding of Dune.
+
+How Dune Works
+--------------
+
+The building brick of Dune is the *rule*:
```
gridbugs, line 17: I think "building block" is the more commonly used term.

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+what actions it needs to execute to build the required targets.
+
+An Example
+----------
+
+Let's take the following example.
+
+- there's a CLI tool written in OCaml.
+- it has some build-time configuration stored in ``config.json``.
+- it has an integration test which consists in running in it with
```
gridbugs, line 32:
```suggestion
- it has an integration test which consists of running in it with
```

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+
+In contrast, to describe the compilation of the executable, we would write:
+
+.. code:: dune
+
+  (executable
+   (name tool)
+   (modules main config))
+
+Here, we use Dune's abstractions. It knows about the OCaml compilation model:
```
gridbugs, line 67: To me this reads more clearly if the first "it" is replaced with "dune".
```suggestion
Here, we use Dune's abstractions. Dune knows about the OCaml compilation model:
```

doc/explanation/mental-model.rst
gridbugs: I've never encountered the word "goal" in dune or other build systems. To avoid introducing a new piece of terminology, what about rewording it to:
:
```suggestion
Dune can build files and aliases. ``dune build tool.exe`` will
```

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+-------------------------
+
+Now that we have seen how Dune sets up rules for a project, let's see how it
+executes a ``dune build`` command.
+
+Dune can build targets: either files, or aliases. ``dune build tool.exe`` will
+build ``tool.exe``, while ``dune build @runtest`` will build the ``runtest``
+alias (``dune runtest`` is a shortcut for ``dune build @runtest``).
+
+To build the target *t*, Dune will first check is *t* is in the source tree. In
```
gridbugs, line 128: I think it would help to clarify that this paragraph is only about files and not aliases. It might also be helpful to explain under what conditions does dune actually build an alias (I don't actually know the answer to this). What if rule has both an alias and target files and all the target files are present?

doc/explanation/mental-model.rst
gridbugs: I personally find explicitly stepped examples like this very helpful when learning how to use new tools.

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+change only part of the project (say the ``main.ml`` file), only a small number
+of rules will be evaluated, the ones that depend on ``main.ml``.
+
+Conclusion
+----------
+
+Dune's underlying model is based on rules. Stanzas are high-level constructs
+that can generate multiple rules, that are not always visible.
+
+To build a target, Dune looks for the rule that produces that target and makes
```
gridbugs, line 169:
```suggestion
To build a target or alias, Dune looks for the rule that produces that target and makes
```
Not sure this change is necessary, but since the `(target ...)` syntax is used for file targets and `(alias ...)` is used for aliases, using the word "target" here might suggest that the behaviour when building an "alias" is different.

gridbugs: Oh actually this paragraph reads as though it's only about file targets, in which case I'd say leave it as such and add an additional sentence summarizing how dune builds aliases.

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+its way back to source files. This explains why it is generally fine with
+"dynamic" dependencies, but needs "static" targets.
```
gridbugs, line 171: I think the detail about dynamic dependencies and static targets should either be skipped here or elaborated upon further outside the conclusion. This page doesn't talk about dynamically generating the dependencies of a rule so I think it's a bit out of place here, and a fairly high-level concept to introduce at this point in the page.

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+
+Rules define a graph which models precise relations between files. Most of the
```
gridbugs, line 173:
```suggestion
Rules define a directed acyclic graph which models precise dependencies between files. Most of the
```

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+- reads its dependencies: ``tool.exe`` and ``testdata.txt``
+- writes no targets
+- by running its action: ``(run tool.exe testdata.txt)``
+- (but it is attached to ``runtest``)
+
+How Dune Interprets Rules
+-------------------------
+
+Now that we have seen how Dune sets up rules for a project, let's see how it
+executes a ``dune build`` command.
```
gridbugs, line 122: I think it would also be helpful to explain what happens when running `dune build` with no arguments. What targets/aliases does it attempt to build in that case?

gridbugs:
Thanks for making this @emillon! I think this stands to have a big impact on helping people understand dune.

doc/explanation/mental-model.rst
Leonidas-from-XIV: Likewise I don't think I ever saw them referred to as goals. I was thinking of what Make calls them and they are usually "Makefile targets". Which are after all very close to files that we build (targets) and aliases (what Make calls "phony targets" and hides behind terrible syntax).

(On the other hand "Dune can build [...] aliases" is oddly worded, maybe "build files" and "evaluate aliases"?)

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
+The Dune Mental Model
+=====================
+
+It is not strictly necessary to understand Dune's underlying model to use it;
+but knowing how it works under the hood will help understand some errors and
+what's possible with Dune.
```
Leonidas-from-XIV, line 6: I think it also makes writing rules easier if you know what you're doing instead of (I assume many like me started out just copying boilerplate from somewhere else).

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+   works, but should be a useful tool to build an understanding of Dune.
+
+How Dune Works
+--------------
+
+The building brick of Dune is the *rule*:
+
+    A *rule* reads *dependencies* and writes *targets* by executing an *action*.
+
+When ``dune build`` is executed, it will first read ``dune`` files to determine
```
Leonidas-from-XIV, line 21:
```suggestion
When ``dune build`` is executed, it will first read the project's ``dune`` files to determine
```

As it feels a bit like there's a word missing.

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+  - and write its output: ``main.cmx``
+  - using an action: ``(run ocamlopt -c main.ml)``
+
+- one rule to compile the ``Config`` module:
+
+  - it will read its dependency: ``config.ml``
+  - and write its output: ``config.cmx``
+  - using an action: ``(run ocamlopt -c config.ml)``
+
+- one rule to compile the ``tool.exe`` executable:
```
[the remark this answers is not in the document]
Leonidas-from-XIV: In theory "compile" is always correct in the broader sense of the word but I agree that as a reader I would expect "link" here, even if this is not necessarily what `ocamlopt` is doing.

But maybe this opinion is influenced too much by the C point of view.

doc/explanation/mental-model.rst
Leonidas-from-XIV: As a second opinion, I find it useful to think about it that way, since as a Dune user one rarely if ever need to think about `.cm*` files, so the intermediate rules aren't all that interesting.

I would even consider making this thinking the starting point since that's sort of the intuition and then explain the workings in more detail later.

doc/explanation/mental-model.rst
Leonidas-from-XIV: Doesn't it actually have some sort of hidden target? After all, if the test successfully finishes, it doesn't get re-run again.

I tripped up on this with e.g. property-based tests which don't have a reproducible outcome, once it succeeds you need `--force` to re-run it as there's no change in dependencies.

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+
+To build the target *t*, Dune will first check is *t* is in the source tree. In
+that case, it will copy it to the ``_build`` directory. Otherwise, it will
+check if it is the target of a rule *r*. In that case, it will build all
+dependencies of *r* and if any dependency has changed, it will run the action
+of *t*. (Dune will raise an error in other cases: if *t* is both in the source
+tree and the target of a rule, or if it is neither)
+
+In the case of our example, if we call ``dune runtest``, Dune will consider all
+rules attached to the ``runtest`` alias. This is just the integration test
```
Leonidas-from-XIV, line 136:
```suggestion
rules attached to the ``runtest`` alias. In this case it is just the integration test
```

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,175 @@
...
+its dependencies available, so we can execute its action to produce its target,
+``config.ml``. Now the module compilation rule for ``Config`` can be executed,
+producing ``config.cmx``; and in turn the linking rule can be executed,
+producing ``tool.exe``. Finally, ``tool.exe`` can be executed with
+``testdata.txt`` as its argument.
+
+In a nutshell: we recursively copied all the dependencies of the test rule, and
+executed the rules in the correct order.
+
+This is a "cold build", where nothing was previously built. But note that if we
```
Leonidas-from-XIV, line 159:
```suggestion
This is a "cold build", where there were no previous build artifacts. Note that if we
```

Leonidas-from-XIV (commented):
I think this would easily be the page of the documentation I would suggest people to read the most.

Added some of my comments and :+1: to the previous comments that this document has gotten.

doc/explanation/mental-model.rst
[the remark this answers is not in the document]
emillon: yes, I meant link here

doc/explanation/mental-model.rst
emillon: This comes from this graph from the presentation (omitting the graphs in a first version is a deliberate design decision, so that the text has to be clear and stays accessible)

```mermaid
graph LR
  tool.exe;
  config.ml;
  main.ml;
  subgraph "executable: tool"
    bm[/compile Main/] --> main.cmx --> l[/link executable/]
    bc[/compile Config/] --> config.cmx
  end
  
  l[/link executable/] --> tool.exe
  config.cmx --> l
  config.ml --> bc
  main.ml --> bm
```

Here what I meant is that the collection of rules from the executable stanza act as an opaque box with the `ml` files as dependencies and `tool.exe` as a target. I think that's useful to think it that way but I shouldn't call that a rule.

doc/explanation/mental-model.rst
emillon: too much lego at home

doc/explanation/mental-model.rst
emillon: I believe that "consist of" defined the components of something , while "consist in" introduces a definition?
Maybe
"it has an integration test, in which the tool is executed with `testdata.txt` as input"?

doc/explanation/mental-model.rst
emillon: I like the DAG part, but if we say "dependencies between files", this reuses the "dependency" word for something else. Maybe "which models dependency relations"?

doc/explanation/mental-model.rst
emillon: I'm not sure about the details, but I think that the implementation of aliases is based on "tokens" (stamp files) whose name is based on the digest of their dependencies

snowleopard: We shouldn't include such implementation details in the mental model. Stamp files are likely to disappear at some point (maybe even soon).

snowleopard: (I'm thinking of nuking stamp files became writing a file on disk is a pretty inefficient way to cache an action execution.)

doc/explanation/mental-model.rst
snowleopard: OK, let's keep it. Looks like most people want it.

doc/explanation/mental-model.rst
snowleopard: I like the graph! My feeling is that it's sufficient to say that the highlighted group of rules corresponds to the `executable` stanza and leave it at that (i.e. no need to talk about thinking about them as one rule).

doc/explanation/mental-model.rst
snowleopard: I like @gridbugs's version. We use "goals" in implementation but happy to leave it there.

I also think that saying "build files and aliases" is fine because it directly corresponds to the CLI `dune build file @alias`.

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,174 @@
...
+   different, it does not touch rule loading and glosses over how caching
+   works, but should be a useful tool to build an understanding of Dune.
+
+How Dune Works
+--------------
+
+The building block of Dune is the *rule*:
+
+    A *rule* reads *dependencies* and writes *targets* by executing an *action*.
+
```
ejgallego, line 20: Sometimes when I have been explaining the mental model to people I've used:

```
{ deps } ---[action]---> { targets }
```

doc/explanation/mental-model.rst
emillon: that's intended to be a callback to the blockquote at the beginning, "a rule reads dependencies and writes targets using an action". I tried to rephrase it.

emillon:
Folks, thanks a lot for your feedback. I've reworked the document a bit, in particular:

- I added a "what to build" section. the goal is to explain that we have (A) a set of rules derived from dune files (B) a "goal" (I've not used the term in the doc) and that (C) using A and B as inputs, an algorithm will do the right thing
- in part "C" (the build algorithm) I've detailed what it means to build a file, build an alias, and execute a rule.

Let me know if you have additional remarks.

doc/explanation/mental-model.rst
```diff
@@ -0,0 +1,197 @@
...
+Running the Tests
+^^^^^^^^^^^^^^^^^
+
+Some rules do not produce any output file, but we're still interested in
+running their actions. A test is a good example: we want the build process to exit
+with an error code if the action fails.
+In that case, the rule does not have targets, but we "attach" it to an
+:term:`alias`, ``runtest`` in this case. This gives us a way of requesting this
+rule to be executed. As we are about to see, rules are executed lazily by
```
Leonidas-from-XIV, line 103: While this sort of implies it is possible I think it would be nice to state directly that more than one rule can be attached to an alias and in fact `runtest` has lots of rules attached to them.

(That's one way my understanding of aliases as "phony Makefile targets" did not match up)

Leonidas-from-XIV (approved):
I think it looks good and further improvements can be added when necessary in the future anyway.

doc/explanation/mental-model.rst
emillon: a "dune for make users" guide could be useful indeed!

PR #9875, opened 2024-01-31
refactor: conditionally add odoc alias rules

jchavarri:
As discussed in https://github.com/ocaml/dune/pull/9839#issuecomment-1917934343.

The previous code was causing trouble when introducing conditionally builds of libraries using `enabled_if` (what's being done in #9839), because Dune would try to add the alias rules for a library that didn't exist.

### DIFF
--- a/src/dune_rules/gen_rules.ml
+++ b/src/dune_rules/gen_rules.ml
@@ -112,12 +112,13 @@ end = struct
       let+ () = toplevel_setup ~sctx ~dir ~toplevel in
       empty_none
     | Library.T lib ->
-      (* XXX why are we setting up private doc rules for disabled libraries? *)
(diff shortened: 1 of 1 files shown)


emillon:
I'm confused, does this PR cause a change in behavior? If that's the case, it's not a refactor and a test is preferable.

jchavarri:
Sorry @emillon you're right that it's not a refactor. I'm afraid I am not able to include a test, given that I can only reproduce the problem under certain context that I can only replicate with the changes applied in the referenced PR. So it might be better to include this change as part of the other PR as originally intended. I will reopen if I can find a way to isolate a test for this change.

rgrinberg:
I believe you should be able to reproduce this as follows:

1. Create a private library (not attached to a package)
2. Put an `enabled_if` clause on it that evaluates to false.
3. Try building the `doc-private` alias on this library

PR #9876, opened 2024-01-31
refactor: do not build a record in Install.Paths

### LINKED ISSUE #4308: `Install.Section.Paths` is inefficient
- `get_local_location` creates a big record and immediately takes a single field, throwing out the rest
- there are various call sites where we do the same: create a `Paths.t` record, immediately take a single field and throw out the rest

emillon:
Fixes #4308

As stated in the issue, we always build a full record for a single field.


rgrinberg:
I don't think the issue is true in general. It's only true for `get_local_location` where it can be fixed locally by either saving the record or just returning the path without constructing it. There's plenty of uses where we reuse the path fields. Also, this new representation is rather annoying as it now has a function and isn't pure data.

emillon:
That's fair. I changed that to just compute a single field. And to simplify the representation, I'm using a map instead of a record. What do you think?

rgrinberg:
I'm not sure I would call the end result an "optimization". We've made artifact substitution faster at the expense of making the install and package management rules equally slower. Seems like the wrong trade-off to make.

You could just as easily fix the problem without any of the downsides by adding your new `get_field` function while keeping the old code. Moreover, the problem with the throwaway record still exists with the `Roots.t` that is created.

rgrinberg:
One easy way to revive this PR would be to just memoize the construction of the record itself. That can be done with a simple hash table to avoid the Memo overhead, and the result should be without downsides.

Issue #9728, opened 2024-01-14
Synopsis & description characters are escaped


toastal:
## Expected Behavior

Dune format does nothing:

```scm
(package
 (name cafe_menu_mgr)
 (synopsis "Menu manager for a café named, กาแฟ")
 (description "Add & remove items from กาแฟ’s menu, manage prices, etc."))
```

## Actual Behavior

Dune formats does:

```scm
(package
 (name cafe_menu_mgr)
 (synopsis
  "Menu manager for a caf\195\169 named, \224\184\129\224\184\178\224\185\129\224\184\159")
 (description
  "Add & remove items from \224\184\129\224\184\178\224\185\129\224\184\159\226\128\153s menu, manage prices, etc."))
```

## Reproduction

- PR with a reproducing test: 

1. Init a project
1. Add non-ASCII characters to synopsis / description
1. Dune Format

## Specifications

- Version of `dune` (output of `dune --version`): 3.12.1
- Version of `ocaml` (output of `ocamlc --version`): 4.14.1
- Operating system (distribution and version): Fedora Linux 39

## Additional information 

Discovered here: https://discuss.ocaml.org/t/why-can-t-i-create-a-project-with-non-ascii-characters/13865/12

@yawaramin says this shouldn’t happen for certain fields so they can still be human-readable.

yawaramin:
I'd certainly expect dune to not touch any values inside string fields. opam doesn't.

emillon:
I had a look at this a while ago, so let me comment on this (actually, even some of the Dune authors have non-ascii characters in their names!).

In short, this is a CST vs AST issue: assuming a UTF-8 encoded `dune` file, Dune will parse `"é"` and `"\195\169"` the same. It will pass the same string down the program, and in particular it should produce the same opam file (probably one with escapes).

To support that properly in `dune fmt`, the formatter needs to be able to tell the difference between these concrete syntaxes. For example, its String constructor would contain the unescaped contents of strings.

We already have 2 parsers (for AST and CST), but instead of doing AST -> CST (which would be a lossy operation), we instead materialize CSTs from ASTs (where we "invent" which concrete syntax was used) for performance reasons.

The situation is described here:

https://github.com/ocaml/dune/blob/3.12.2/src/dune_sexp/parser.ml#L5-L15

It describes several options: do the conversion the other way around (probably too slow), write another parser, GADT tricks.

yawaramin:
[...]

PR #9877, opened 2024-01-31
test: for non-ASCII characters (#9728)

moyodiallo:
See #9728.

### DIFF
--- a/test/blackbox-tests/test-cases/enabled_if/non-ascii.t
+++ b/test/blackbox-tests/test-cases/enabled_if/non-ascii.t
@@ -1,0 +1,13 @@
+Afer the parsing non-ascii characters could be compared through expressions.
+Linked to the issue #9728
+
+  $ echo '(lang dune 3.0)' > dune-project
+  $ cat >dune <<"EOF"
+  > (rule
+  >  (alias default)
+  >  (enabled_if (= "É" "\195\137"))
+  >  (action (echo "Hello, world!\n")))
+  > EOF
+
+  $ dune build
+  Hello, world!
--- a/test/blackbox-tests/test-cases/formatting/non-ascii-characters.t
+++ b/test/blackbox-tests/test-cases/formatting/non-ascii-characters.t
@@ -8,7 +8,6 @@ How the non-ASCII characters are handled, this is also related to the issue #972
   $ dune format-dune-file <<EOF
   > (run foo %{bin:é})
   > EOF
-  > (run foo %{bin:é})
   File "", line 1, characters 15-16:
   Error: The character '\195' is not allowed inside %{...} forms
   run: not found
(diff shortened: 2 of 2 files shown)


moyodiallo:
Why the check failing only on macos with this error https://github.com/ocaml/dune/actions/runs/7728961114/job/21071024492?pr=9877. AFAIK even though there's no execution involve with the dune file only formatted.

test/blackbox-tests/test-cases/formatting/non-ascii-characters.t
```diff
@@ -0,0 +1,20 @@
...
+  > (run foo %{bin:é})
+  File "", line 1, characters 15-16:
+  Error: The character '\195' is not allowed inside %{...} forms
+  run: not found
+  [127]
+
+  $ dune format-dune-file <<EOF
+  > (= "É" "\195\137")
+  > EOF
+  (= "\195\137" "\195\137")
```
emillon, line 20: This part is interesting to add, but not just as a formatting check. We want to make sure that this evaluates to `true`. Maybe as `test-cases/enabled_if/non-ascii.t`?

moyodiallo: Seeing the result of after parsed, the evaluation will be a comparison between string. Is it really necessary to add that evaluation ?

emillon: yes, this shows that we do the comparison after interpreting escape directives.

moyodiallo: I get it.

moyodiallo: Done in https://github.com/ocaml/dune/pull/9877/commits/9d2d404e7b8ba44b4b302edac91b8102cff1506a.

test/blackbox-tests/test-cases/enabled_if/non-ascii.t
```diff
@@ -0,0 +1,13 @@
+Afer the parsing non-ascii characters could be compared through expressions.
+Linked to the issue #9728
+
+  $ echo '(lang dune 3.0)' > dune-project
+  $ cat >dune <<"EOF"
+  > (rule
+  >  (alias default)
```
emillon, line 7: nice trick to use the default alias here

emillon:
Thanks. I removed the superfluous test

test/blackbox-tests/test-cases/enabled_if/non-ascii.t
moyodiallo: Thanks, but It was inspired by `/test/blackbox-tests/test-cases/enabled_if/read-macro.t`.

PR #9884, opened 2024-02-01
test(melange): remove rescript-syntax tests from CI

rgrinberg:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,7 @@ ppx_inline_test \
 ppxlib \
 ctypes \
 "utop>=2.6.0" \
-"melange>=1.0.0" \
-"rescript-syntax"
+"melange=1.0.0"
 # Dependencies recommended for developing dune locally,
 # but not wanted in CI
 DEV_DEPS := \
--- a/test/blackbox-tests/test-cases/melange/dune
+++ b/test/blackbox-tests/test-cases/melange/dune
@@ -4,7 +4,8 @@
 
 (cram
  (deps %{bin:rescript_syntax})
- (applies_to rescript-syntax))
+ (applies_to rescript-syntax)
+ (runtest_alias false))
 
 (cram
  (deps %{bin:melc_stdlib_prefix})
--- a/chrome-trace.opam
+++ b/chrome-trace.opam
@@ -10,7 +10,7 @@ homepage: "https://github.com/ocaml/dune"
 doc: "https://dune.readthedocs.io/"
 bug-reports: "https://github.com/ocaml/dune/issues"
 depends: [
-  "dune" {>= "3.5"}
+  "dune" {>= "3.12"}
   "ocaml" {>= "4.08.0"}
   "odoc" {with-doc}
 ]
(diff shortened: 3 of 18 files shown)


rgrinberg:
@gridbugs does the monorepo benchmark need to be updated to 3.12?

gridbugs:
> @gridbugs does the monorepo benchmark need to be updated to 3.12?

I don't think so. Looks like it ran successfully.

gridbugs:
> @gridbugs does the monorepo benchmark need to be updated to 3.12?

The error is:
```
Step 24/28 : RUN . ~/.profile && dune build bin/main.exe --release
 ---> Running in 1113cef47d40
File "dune-project", line 1, characters 11-15:
1 | (lang dune 3.12)
               ^^^^
Error: Version 3.12 of the dune language is not supported.
Supported versions of this extension in version 3.12 of the dune language:
- 1.0 to 1.12
- 2.0 to 2.9
- 3.0 to 3.11
```

That error message doesn't make sense to me. Which extension is it talking about? Is the dune language itself an extension of dune?

rgrinberg:
>  Is the dune language itself an extension of dune?

That's what it claims but it doesn't really work like that. We should amend that. Anyway, the version of dune used to build the benchmarks is too old to understand `(lang dune 3.12)` after this PR.

gridbugs:
Ah so the failing step is when it builds dune in the benchmark docker container. I think it's using a version of dune that's too old. The installed version of dune in the cached docker image is too old.

gridbugs:
https://github.com/ocaml/dune/pull/9891 will hopefully fix the problem.

anmonteiro (approved):

test/blackbox-tests/test-cases/melange/dune
```diff
@@ -4,7 +4,8 @@
 
 (cram
  (deps %{bin:rescript_syntax})
- (applies_to rescript-syntax))
+ (applies_to rescript-syntax)
+ (runtest_alias false))
```
jchavarri, line 8: What's the upside of this vs removing the `cram` stanza altogether?

jchavarri (approved):
lgtm. Is this part still needed?

https://github.com/ocaml/dune/blob/e06a128a52b47b884e85c0470ba4173ed64b9501/src/dune_engine/action.ml#L77-L78

rgrinberg:
> lgtm. Is this part still needed?
> 
> https://github.com/ocaml/dune/blob/e06a128a52b47b884e85c0470ba4173ed64b9501/src/dune_engine/action.ml#L77-L78

Indeed we can remove it. Although I'd rather think of a way of removing all the hints from the engine.

Makefile
```diff
@@ -25,8 +25,7 @@ ppx_inline_test \
 ppxlib \
 ctypes \
 "utop>=2.6.0" \
-"melange>=1.0.0" \
-"rescript-syntax"
+"melange=1.0.0"
```
emillon, line 28: should have been >=1.0.0, right?

emillon: ah no, the tests are using 1.0.0 only features.

Issue #9862, opened 2024-01-30
dune subst and opam/ subfolder


emillon:
I know that this is from the original code but this looks like it needs to be adopted if we generate OPAM files in the `opam/` subdirectory.

_Originally posted by @Leonidas-from-XIV in https://github.com/ocaml/dune/pull/9811#discussion_r1463363517_

PR #9894, opened 2024-02-01
test: add repro for #9862 (subst in opam/ subdir)

emillon:


test/blackbox-tests/test-cases/subst/opam-files/at-root.t
```diff
@@ -0,0 +1,26 @@
+  $ cat > dune-project << EOF
```
rgrinberg, line 1: This test could use a toplevel description.

rgrinberg (approved):
We don't do this consistently, but adding a runtime dependency on `%{bin:git}` for these two tests would be welcome.

PR #9895, opened 2024-02-01
fix(subst): subst for opam files in opam/ subdir

### LINKED ISSUE #9862: dune subst and opam/ subfolder

emillon:
Fixes #9862

### DIFF
--- a/bin/subst.ml
+++ b/bin/subst.ml
@@ -119,7 +119,7 @@ let subst_string s path ~map =
     Some (Buffer.contents buf)
 ;;
 
-let subst_file path ~map =
+let subst_file path ~map opam_package_files =
   match Io.with_file_in path ~f:Io.read_all_unless_large with
   | Error () ->
     let hints =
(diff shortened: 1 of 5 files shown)


rgrinberg (approved):
The fix looks fine, but the implementation details of opam file locations are now split across 3 modules (`Subst`, `Opam_create`, and `Package`). Ideally, it would all exits in a single module, but at least you can make this less fragile but collecting all the opam files for the project with the help of `Dune_project.packages` and `Package.opam_file` and then get rid of this `Package.is_opam_file` function altogether.

emillon:
I was wondering about this as well. But I'm not sure that `Dune_project` is the right place to do that: we want to run the substitutions only at `Path.root`, not for opam files of vendored packages. Should we use `Scope` for that? Or move the functions to `Dune_project` as you said but only act if the project is defined at `Path.root`?

rgrinberg:
We're already going to run substitutions only for the root project as we're loading the project only in `Path.root`.

Note that my suggestion does not involve modifying `dune_project.ml` at all. Everything can be implemented on top of existing api's in `bin/subst.ml`. All you should need is `Dune_project.packages` and `Package.opam_file`.

rgrinberg:
Ended up pushing the improvement. Btw, the test did not define a package in the project file which was wrong. We only support opam files in `opam/` with corresponding `package` stanzas.

emillon:
Thanks Rudi!

Alizter:
FTR this broke Coq's CI upstream. See https://coq.zulipchat.com/#narrow/stream/237656-Coq-devs-.26-plugin-devs/topic/.5BERROR.5D.20The.20comp**.20of.20coq-core.2Edev.20failed.20at.20.22dune.20subst.22.2E

Issue #9773, opened 2024-01-18
Ctypes and foreign bindings won't bundle library properly


mgree:
I'm working on [libdash](https://github.com/mgree/libdash), which is offers OCaml and Python bindings to the dash POSIX shell parser. dune is being uncooperative around packaging the library up properly (i.e., with both bindings and code). My work is in the ~~[`ocaml-static-ctypes` branch](https://github.com/mgree/libdash/tree/ocaml-static-ctypes)~~ [`main` branch](https://github.com/mgree/libdash) (since OCaml bindings don't work right now anyway). (I originally posted this [on discuss.ocaml.org](https://discuss.ocaml.org/t/packaging-foreign-archives-with-bindings-using-ctypes/13685), but wasn't able to resolve things there.)

The structure is:

- an autotools build constructs `libdash.a` and `dlldash.so` from the (lightly modified dash source)
- some static ctypes bindings connects the C interface to OCaml
- a bunch of OCaml converts the C ASTs into more sensible structures
- some executables use the OCaml ASTs to do some work

A critical property here is that the C library is not a common one, and must be distributed _with_ the OCaml source. The produced OCaml native library should include all of the C code.

# dune woes

I---after several years of trying off and on---have gotten dune to build the project locally and have tests pass. (I posted here [_three years ago_](https://discuss.ocaml.org/t/opam-packaging-questions-around-ocaml-bindings-to-custom-c-library/5120).) But there's a problem: the build fails with multiple uses of `libdash.a`. Here's the output:

```ShellSession
$ dune build
Error: Multiple rules generated for
_build/install/default/lib/libdash/libdash.a:
- ocaml/dune:8
- ocaml/dune:8
-> required by _build/default/libdash.install
-> required by alias all
-> required by alias default
```

The offending clause:

```dune
(library
  (name libdash)
  (public_name libdash)
  (modes native)
  (modules (:standard \ json_to_shell shell_to_json ast_json))
  (libraries ctypes ctypes.foreign)
  (foreign_archives ../dash)
  (ctypes
    (external_library_name dash)
    (build_flags_resolver (vendored (c_flags :standard) (c_library_flags :standard)))
    (deps (glob_files ../src/*.h) ../src/builtins.h ../src/nodes.h ../src/syntax.h ../src/token.h ../src/token_vars.h)
    (headers (preamble
              "\
             \n#include \"../src/shell.h\"\
             \n#include \"../src/memalloc.h\"\
[...]

PR #9896, opened 2024-02-01
test: add repro for #9773 (name collision)

emillon:


rgrinberg (approved):
LGTM. I think you can simplify the test to produce a dummy archive/shared object. Right now the test depends on `%{cc}` and `ar` and it doesn't necessarily need to.

emillon:
By dummy, do you mean empty? I think that works on linux but not on macos, unfortunately.

rgrinberg:
Yes, I meant empty. I don't see why it would make a difference on macos, the test her doesn't execute any commands as it fails to generate the rules. In any case, the way it is is fine now as well

emillon:
Ah, I see. The problem is that once the problem that the test demonstrates is fixed, we'll have an error message.

PR #9897, opened 2024-02-01
fix: don't add odoc private rules in unavailable libs

jchavarri:
As discussed in https://github.com/ocaml/dune/pull/9839#issuecomment-1917934343.

The previous code was causing trouble when introducing conditionally builds of libraries using `enabled_if` (what's being done in #9839), because Dune would try to add the alias rules for a library that didn't exist.

The test uses the guidelines shared by @rgrinberg in https://github.com/ocaml/dune/pull/9875#issuecomment-1920223246.

Can check the [first commit](https://github.com/ocaml/dune/commit/6d7cab9999e4624a70415278794d3812840787de) in the PR to see current behavior in `main`.

### DIFF
--- a/src/dune_rules/gen_rules.ml
+++ b/src/dune_rules/gen_rules.ml
@@ -112,12 +112,13 @@ end = struct
       let+ () = toplevel_setup ~sctx ~dir ~toplevel in
       empty_none
     | Library.T lib ->
-      (* XXX why are we setting up private doc rules for disabled libraries? *)
-      let* () = Odoc.setup_private_library_doc_alias sctx ~scope ~dir:ctx_dir lib
-      and+ enabled_if = Lib.DB.available (Scope.libs scope) (Library.best_name lib) in
+      let* enabled_if = Lib.DB.available (Scope.libs scope) (Library.best_name lib) in
       if_available_buildable
         ~loc:lib.buildable.loc
-        (fun () -> Lib_rules.rules lib ~sctx ~dir ~scope ~dir_contents ~expander)
+        (fun () ->
+          let+ () = Odoc.setup_private_library_doc_alias sctx ~scope ~dir:ctx_dir lib
+          and+ rules = Lib_rules.rules lib ~sctx ~dir ~scope ~dir_contents ~expander in
+          rules)
         enabled_if
     | Foreign.Library.T lib ->
       let+ () = Lib_rules.foreign_rules lib ~sctx ~dir ~dir_contents ~expander in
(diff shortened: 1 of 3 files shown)


rgrinberg (approved):
LGTM. This deserves a CHANGES entry

test/blackbox-tests/test-cases/enabled_if/eif-lib-unavailable-doc-private.t
```diff
@@ -0,0 +1,16 @@
+Show behavior of doc-private alias when it's part of an unavailable library
```
rgrinberg, line 1: There's an odoc directory for odoc tests. This test can go there.

jchavarri:
Thanks for the review.

> There's an odoc directory for odoc tests. This test can go there.
> LGTM. This deserves a CHANGES entry

Both done.

src/dune_rules/gen_rules.ml
```diff
@@ -112,12 +112,13 @@ end = struct
...
-      (* XXX why are we setting up private doc rules for disabled libraries? *)
-      let* () = Odoc.setup_private_library_doc_alias sctx ~scope ~dir:ctx_dir lib
-      and+ enabled_if = Lib.DB.available (Scope.libs scope) (Library.best_name lib) in
+      let* enabled_if = Lib.DB.available (Scope.libs scope) (Library.best_name lib) in
       if_available_buildable
         ~loc:lib.buildable.loc
-        (fun () -> Lib_rules.rules lib ~sctx ~dir ~scope ~dir_contents ~expander)
+        (fun () ->
+          let+ () = Odoc.setup_private_library_doc_alias sctx ~scope ~dir:ctx_dir lib
+          and+ rules = Lib_rules.rules lib ~sctx ~dir ~scope ~dir_contents ~expander in
```
emillon, line 120: style question for @rgrinberg: is there an advantage in doing this compared to the following (I've noticed that pattern in odoc rules in particular)

```ocaml
(fun () ->
  let* () = Odoc.setup_private_library_doc_alias sctx ~scope ~dir:ctx_dir lib in
  Lib_rules.rules lib ~sctx ~dir ~scope ~dir_contents ~expander)
```

rgrinberg: The former has the opportunity to do something concurrently while the the latter does not. Usually it doesn't matter when generating rules as we rarely do work where we can take advantage of it. However, as we make dune more concurrent and allow for more dynamism (e.g. dune file generation), the latter becomes more important.

PR #9905, opened 2024-02-02
chore(pkg): Remove rudimentary git config parser in favor of `git config`

Leonidas-from-XIV:
This removes the rudimentary config parser in favor of using `git config --list --blob` which conveniently can just parse from any blob in the repo.

This makes our parser match the git parser, thus no need for the parser tests anymore.


src/dune_pkg/rev_store.ml
```diff
@@ -391,22 +391,88 @@ module Entry = struct
...
+      include Comparable.Make (T)
+    end
+
+    let config { dir } revision path : t Fiber.t =
+      let (Rev.Rev rev) = revision in
+      let location = [ "--blob"; sprintf "%s:%s" rev (Path.Local.to_string path) ] in
+      let command = [ "config"; "--list" ] @ location in
+      let failure_mode =
+        Process.Failure_mode.Accept
+          (Predicate.create (fun x -> Int.equal x 0 || Int.equal x 128))
```
emillon, line 432: What does 128 map to?

src/dune_pkg/rev_store.ml
```diff
@@ -391,22 +391,88 @@ module Entry = struct
...
+      in
+      let git = Lazy.force Vcs.git in
+      let+ content =
+        let stderr_to = make_stderr () in
+        Process.run_capture_lines ~dir ~display:Quiet ~stderr_to failure_mode git command
+      in
+      match content with
+      | Error _ -> []
```
emillon, line 440: should we say something here?

src/dune_pkg/rev_store.ml
```diff
@@ -391,22 +391,88 @@ module Entry = struct
...
+        content
+        |> List.fold_left ~init:KV.Map.empty ~f:(fun acc line ->
+          match String.lsplit2 ~on:'=' line with
+          | None -> acc
+          | Some (key, value) ->
+            (match String.lsplit2 ~on:'.' key with
+             | None -> acc
+             | Some (section, key) ->
+               let arg, binding =
+                 match String.rsplit2 ~on:'.' key with
```
emillon, line 451: I suppose that the `rsplit2` takes care of parsing things like `branch.compat-5.0-dune-2.9.remote=origin`, but tests for these kinds of things would be useful

src/dune_pkg/rev_store.ml
```diff
@@ -391,22 +391,88 @@ module Entry = struct
...
+        let to_dyn = Tuple.T2.to_dyn Dyn.string (Dyn.option Dyn.string)
+      end
+
+      include Comparable.Make (T)
+    end
+
+    let config { dir } revision path : t Fiber.t =
+      let (Rev.Rev rev) = revision in
+      let location = [ "--blob"; sprintf "%s:%s" rev (Path.Local.to_string path) ] in
+      let command = [ "config"; "--list" ] @ location in
```
emillon, line 429: I don't have the big picture, but instead of reading all the config at once, we could just read each key as needed. Especially if we're going to `--set` things, this could avoid some race conditions if the config file changes.

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: This is what `git` returns when the blob is not found (e.g. because the revisions just doesn't have a `.gitmodules`).

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: I don't think so, it's not actually an error (the `failure_mode` maps it to `Error` but it is `Error 128` which is emitted when there's no `.gitmodules`), in which case it it an empty list of submodules.

emillon: `git-config.1` lists some error cases, such as when the config file is invalid. I think we should display something in these situations.

src/dune_pkg/rev_store.ml
emillon: it would be useful to test for that situation separately, to avoid handling everything at once. do we have a working tree at that point? if that's the case, we can check for existence (and use `--file`). Otherwise, we can use `git cat-file -t main:.gitsubmodules`.

Or actually, maybe just use a failure mode of `Return` and deal with it at that point (`Error 128 -> []`).

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: The problem with this is that for each submodule we'd need to call `git config` at least twice (for `submodule.<name>.path` and `submodule.<name>.url`) and to get the list of `<name>` submodules we have to call `git config --list` anyway, at which point we get all the parsed data.

I don't think this is a big issue, since we probably can't `--set` on `--blob` anyway and the `.gitmodule` files that we parse are usually fairly short.

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: Good point, it's a bit of a shame that in any case the code is 128. I've split it up to check for the existence of the file with `cat-file -e` and used `run_capture_lines` which displays errors on failures.

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: Fair enough, I dropped that approach and split it into checking for existence of the file + followed by the parsing if it exists.

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: Indeed. I've added some tests for this. It's sort of ugly that this needs to be exposed to be tested with expect tests, would be nice to avoid that if there was a better way (I could stuff it into a `Private` module but that's also just a workaround).

emillon: One alternative is to make a test that has a fake `git` binary in scope. But that's not necessarily better.

Leonidas-from-XIV: True; haven't thought of that. I think the code is now fine-ish. If the processing was more complicated to the point that returning a tuple wouldn't be feasible a test `git` binary would indeed be the better alternative.

emillon (approved):

PR #9908, opened 2024-02-02
fix: [map_result Accept] is broken

rgrinberg:
It's hard coded to only accept 0 as the error code instead of consulting
the predicate.


rgrinberg:
Actually, this isn't a bug but just a really weird behavior.

PR #9910, opened 2024-02-03
fix(pkg): filter ls-remote

rgrinberg:
We can limit ls-remote to with --heads and --tags to not fetch unwated
refs.


Leonidas-from-XIV (approved):
Good point. I checked what else there would be and for larger projects like OPAM `ls-remote` also returns a *massive* amount of PRs that we don't care about, so cutting that out is useful.

PR #9913, opened 2024-02-03
feature: dynamic_include stanza

rgrinberg:
Introduce the [dynamic_include] stanza. It's like [(include ..)], but
doesn't allow us to generate static stanzas and a few other stanzas that can affect the entire context. Those include:

* libraries, coq theories, and other library like stanzas
* public executables
* install stanzas with `(section bin)`.
* plugin stanzas

The limitation on generating private libraries is possible to lift, but it will take quite a bit more effort. I will do so in subsequent PR's.

# TODO

- [x] `dynamic_include` in `dynamic_include`
- [x] `include` in `dynamic_include`.
- [x] cycle checking for the above.

### DIFF
--- a/doc/stanzas/dynamic_include.rst
+++ b/doc/stanzas/dynamic_include.rst
@@ -1,0 +1,26 @@
+dynamic_include
+---------------
+
+The ``dynamic_include`` stanza allows including the contents of another file in
+the current dune file like the ``include`` stanza.  However, the
+``dynamic_include`` stanza allows the included file to be the target of a rule
+and disallows generating some stanzas.
+
+For instance:
+
+.. code:: dune
+
+   (subdir b (dynamic_include ../a/foo.inc))
+   (subdir a
+    (with-stdout-to foo.inc
+     (echo "(rule (with-stdout-to file) (echo bar))")))
+
+In the example above, the dynamic rule loading and generation are split into
+different directories to avoid rule loading cycles as rules are loaded per
+directory.
+
+The following stanzas cannot be dynamically generated:
+
+* Libraries, coq theories, library redirects
+* Public executables or install section with the ``bin`` section
+* Plugin stanzas
--- a/src/dune_rules/dune_file.ml
+++ b/src/dune_rules/dune_file.ml
@@ -3,11 +3,13 @@ open Import
 type t =
   { dir : Path.Source.t
   ; project : Dune_project.t
-  ; stanzas : Stanza.t list
+  ; stanzas : Stanza.t list Memo.Lazy.t
+  ; static_stanzas : Stanza.t list
   }
 
 let dir t = t.dir
-let stanzas t = t.stanzas
+let stanzas t = Memo.Lazy.force t.stanzas
+let static_stanzas t = t.static_stanzas
 let project t = t.project
 
 module Mask = struct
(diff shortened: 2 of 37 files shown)


nojb:
Can you say a few words for the motivation behind this addition?

About the name `dynamic_include`: typically my first feeling is that a "dynamic" X is something more powerful than a "plain" X, but in this case it seems to be _less_ powerful than a plain X. In any case, I don't understand the feature well enough yet to propose a different name.

rgrinberg:
The main motivation is a straightforward one. Allow a way of generating rules without promoting them to the source.

Why a different name? Because this include has different semantics and limitations that are (likely) impossible to lift. So I think a different name is appropriate.

nojb:
> Allow a way of generating rules without promoting them to the source.

Sorry for the naive question, but what does "promoting them to the source" mean here? I looked at the test, but couldn't figure out what is special about the new form (vs a normal `(include)`).

rgrinberg:
When you write (include foo), the file foo must exist in the source directory. To make foo be generated by a rule, we workaround it by making the rule promotable.

If you use this new stanza, you can skip the promotion step and use a regular rule to generate stanzas provided the limitations in the description are acceptable to you.

nojb:
> If you use this new stanza, you can skip the promotion step and use a regular rule to generate stanzas provided the limitations in the description are acceptable to you.

I had missed that, thanks. Indeed, this is a cool addition!

Being able to define executables in this way is already a big advance; eg defining a test suite for every file in a directory.

src/dune_rules/stanzas/stanzas.ml
```diff
@@ -117,6 +133,10 @@ let stanzas : constructors =
       , let+ () = Dune_lang.Syntax.since Stanza.syntax (2, 0)
         and+ t = Deprecated_library_name.decode in
         [ Deprecated_library_name.make_stanza t ] )
+    ; ( "dynamic_include"
+      , let+ () = Dune_lang.Syntax.since Stanza.syntax (3, 13)
```
nojb, line 137:
```suggestion
      , let+ () = Dune_lang.Syntax.since Stanza.syntax (3, 14)
```

src/dune_rules/dune_file.ml
```diff
@@ -288,47 +346,36 @@ module Script = struct
...
+    | Install_conf.T { section = loc, Section Bin; _ } ->
+      User_error.raise ~loc [ Pp.text "binary section cannot be generated dynamically" ]
+    | Coq_stanza.Theory.T { buildable = { Coq_stanza.Buildable.loc; _ }; _ }
+    | Library.T { buildable = { loc; _ }; _ }
+    | Install_conf.T { section = _, Site { loc; _ }; _ }
+    | Executables.T
+        { buildable = { loc; _ }; install_conf = Some { section = _, Section Bin; _ }; _ }
+    | Deprecated_library_name.T { Library_redirect.loc; _ }
+    | Plugin.T { site = loc, (_, _); _ } -> User_error.raise ~loc err
+    | _ -> ()
```
nojb, line 368: Just for my understanding, if one removed this check and used one of the "forbidden" stanzas inside a dynamic include, how would Dune react?

rgrinberg: It would ignore processing them in some key places. Such as adding them to the library database or the database we use for %{bin:..}.

nojb: OK, thanks.

nojb (approved):
LGTM (but I only looked at the code changes briefly)

src/dune_rules/dune_file.ml
```diff
@@ -288,47 +346,36 @@ module Script = struct
...
-  match mask with
-  | None -> dune_files
-  | Some visible_pkgs ->
-    List.map dune_files ~f:(fun dune_file ->
-      { dune_file with
-        stanzas = filter_out_stanzas_from_hidden_packages ~visible_pkgs dune_file.stanzas
-      })
+let check_dynamic_stanza =
+  (* CR-rgrinberg: unfortunately this needs to kept in sync with the rules
+     manually *)
```
nojb, line 355: What about making this boolean flag part of the definition of the stanza (ie an argument of `Stanza.Make`)? That way the programmer is forced to provide this logic when adding new stanzas.

doc/stanzas/dynamic_include.rst
```diff
@@ -0,0 +1,26 @@
...
+For instance:
+
+.. code:: dune
+
+   (subdir b (dynamic_include ../a/foo.inc))
+   (subdir a
+    (with-stdout-to foo.inc
+     (echo "(rule (with-stdout-to file) (echo bar))")))
+
+In the example above, the dynaminc rule loading and generation are split into
```
emillon, line 18:
```suggestion
In the example above, the dynamic rule loading and generation are split into
```

src/dune_rules/dune_file.ml
rgrinberg: I tried this, but it ends up being a pretty massive change. I do agree that something like this would be good and we should continue using `Stanza.Make` to make the stanza interface more uniform. Will leave a TODO for this.

PR #9918, opened 2024-02-05
pkg: absolute paths for section pforms

gridbugs:
Some packages expect these variables to resolve to absolute paths rather than relative paths. `ocaml-variants` is an example of such a package.

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -587,7 +587,9 @@ module Action_expander = struct
       | Sys_ocaml_version ->
         sys_poll_var (fun { sys_ocaml_version; _ } -> sys_ocaml_version)
       | Build -> Memo.return [ Value.Dir (Path.build paths.source_dir) ]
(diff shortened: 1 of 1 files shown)


rgrinberg:
As the tests show, this breaks sandboxing. The way this needs to be done is how I did with the run action, delaying the expansion of paths until rule execution.

Which part of ocaml-variants relies on absolute paths btw? I thought it was just `run` and I already fixed the issue there.

gridbugs:
The `configure` script complains if its `--prefix` and `--docdir` arguments aren't given absolute paths.

gridbugs:
Strange. I see in our tests that we do indeed emit absolute paths in run actions. Not sure why it's producing relative paths when I attempt to build a project in the wild. I'll keep investigating.

gridbugs:
Looks like it doesn't emit absolute paths when the variable is in a string interpolation. `(run echo %{prefix})` prints an absolute path but `(run echo prefix=%{prefix})` prints "prefix=<relative path>". Closing this.

PR #9921, opened 2024-02-05
chore: add missing unix dependencies

emillon:
See #8607 for a recent example.

OCaml 5 adds a `ocaml_deprecated_auto_include` alert when the `unix` library is referenced but without an explicit dependency.

This commit fixes these alerts by adding a dependency to `unix` in places that refer to `Unix` directly (ie, it makes the metadata correct thus removing the alert).


ElectreAAS (approved):
Thanks, that's exactly the kind of annoyances that slow down adoption of OCaml 5. I just tested this and can confirm this PR fixes the problem.

rgrinberg (approved):

PR #9928, opened 2024-02-05
Repro: dune fmt should not fail to format when depopt isn't available

ElectreAAS:
Initially reported by @EmileTrotignon & @art-w.
The use-case here is in the CI:
Opam CI assumes no dependencies are needed to format the project. In most cases that's valid.
However, adding `(enabled_if (and %{lib-available:nope} (= %{version:nope} 9.9.9)))` should not prevent the formatting from taking place, even if `nope` isn't installed.
Also the `and` here isn't lazy? That's unusual...


rgrinberg:
Indeed this is an issue. Simplest way to address it would be to change the expansion monad to be `Resolve.Memo.t` rather than `Resolve.t`.

emillon:
Thanks for the repro.
In terms of tracking, you can:
- open an issue with the description of the problem
- turn this PR into a reproduction case that references the issue
- (otherwise, if we just merge this PR, we don't have anything open to track the issue)

ElectreAAS:
- [x] open an issue with the description of the problem
- [x] turn this PR into a reproduction case that references the issue

test/blackbox-tests/test-cases/lazy-and.t
```diff
@@ -0,0 +1,28 @@
+Running `dune fmt` when a dependency is not available should not crash this way.
+  $ cat > dune-project <<EOF
+  > (lang dune 3.13)
+  > EOF
+
+  $ cat > dune << EOF
+  > (test
+  >  (name foo)
```
emillon, line 8: does the test work when you remove the `enabled_if` part? if not, it would be useful to add an empty `foo.ml`, to make sure that the PR fixing the issue will actually make the test pass.

ElectreAAS: I added a second dune file without the enabled_if part. In that case everything works as intended (I think)

ElectreAAS:
I just rebased on main. Assuming the CI is green, I think this is good to go.

test/blackbox-tests/test-cases/lazy-and.t
```diff
@@ -0,0 +1,55 @@
...
+
+  $ dune fmt
+  File "foo.ml", line 1, characters 0-0:
+  Error: Files _build/default/foo.ml and _build/default/.formatted/foo.ml
+  differ.
+  Promoting _build/default/.formatted/foo.ml to foo.ml.
+  [1]
+
+  $ cat foo.ml
+  let () = Fun.id ()
```
Leonidas-from-XIV, line 55: Does this not require that `ocamlformat` is installed for the test?

Leonidas-from-XIV (commented):
It seems to me that should better be two tests:

1. the short-circuiting `and`
2. the `ocamlformat` issue (e.g. with `enabled_if false`)

rgrinberg:

PR #9929, opened 2024-02-05
Chore: Explicitly call Stdlib.Mutex in file with open Base

ElectreAAS:
When compiling with OCaml 5.0 or 5.1 we get this error:

```
File "test/expect-tests/dune_file_watcher/dune_file_watcher_tests_lib.ml", line 7, characters 2-12:
7 |   Mutex.lock mutex;
      ^^^^^^^^^^
Error (alert deprecated): module Base.Mutex
[2016-09] this element comes from the stdlib distributed with OCaml.
Referring to the stdlib directly is discouraged by Base. You should either
use the equivalent functionality offered by Base, or if you really want to
refer to the stdlib, use Stdlib.Mutex instead

File "test/expect-tests/dune_file_watcher/dune_file_watcher_tests_lib.ml", line 8, characters 37-49:
8 |   Exn.protect ~f ~finally:(fun () -> Mutex.unlock mutex)
                                         ^^^^^^^^^^^^
Error (alert deprecated): module Base.Mutex
[2016-09] this element comes from the stdlib distributed with OCaml.
Referring to the stdlib directly is discouraged by Base. You should either
use the equivalent functionality offered by Base, or if you really want to
refer to the stdlib, use Stdlib.Mutex instead
```

This PR explicitly calls the Stdlib and removes the error.

### DIFF
--- a/test/expect-tests/dune_file_watcher/dune_file_watcher_tests_lib.ml
+++ b/test/expect-tests/dune_file_watcher/dune_file_watcher_tests_lib.ml
@@ -4,8 +4,14 @@ open Base
 open Stdune
 
 let critical_section mutex ~f =
-  Stdlib.Mutex.lock mutex;
-  Exn.protect ~f ~finally:(fun () -> Stdlib.Mutex.unlock mutex)
+  (* Since 5.0, using "Mutex" with Base open rings an alert and suggests
+     we use "Stdlib.Mutex" instead.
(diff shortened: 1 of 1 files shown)


ElectreAAS:
> I think that the CI error is because Mutex is not a submodule of Stdlib in 4.14.

Thanks for the tip! I just rebased on main, and added 17478b3e935d2e99de227d40f5a1a7ce5e6041ef which should fix the CI and explain things further.

emillon:
Thanks!

PR #9930, opened 2024-02-05
Correct spelling errors throughout the project

kupolak:
As mentioned in the title - address typos.


vendor/0install-solver/src/solver/diagnostics.ml
emillon, line 1: source code in the `vendor/` directory is not being developed here (it's pulled from other projects), so we can't fix it this way. can you remove this part from this PR?

emillon (commented):
Thanks a lot. I checked the changes and it's only typos that can be changed without complications. But the `vendor/` parts cannot be fixed this way.

vendor/0install-solver/src/solver/diagnostics.ml
kupolak: Sure, I removed it.
Does it make sense to use git submodules for these other projects? That could be a potential improvement I suppose but I'm not sure if it's doable.

emillon: thanks for the suggestion, but it's not directly mapping to an existing repository so we don't use submodules in dune.

PR #9932, opened 2024-02-06
fix(pkg): do not redownload the same url's

### LINKED ISSUE #9701: Re-downloading the same sources when building packages

rgrinberg:
Download them all to _build/_fetch/$checksum

### DIFF
--- a/src/dune_rules/fetch_rules.ml
+++ b/src/dune_rules/fetch_rules.ml
@@ -1,0 +1,315 @@
+open Import
+open Memo.O
+module Gen_rules = Build_config.Gen_rules
+
+include struct
+  open Dune_pkg
+  module Checksum = Checksum
+  module Pkg = Lock_dir.Pkg
+end
+
+let context_name = Context_name.of_string "_fetch"
+let context = Build_context.create ~name:context_name
+let digest_of_url url = OpamUrl.to_string url |> Digest.string
+
+let make_target ~kind url_or_checksum =
+  let basename =
+    match kind with
+    | `File -> "file"
+    | `Directory -> "dir"
+  in
+  let dir1, dir2 =
+    match url_or_checksum with
+    | `Url url -> "url", digest_of_url url |> Digest.to_string
+    | `Checksum checksum -> "checksum", Checksum.to_string checksum
+  in
+  Path.Build.L.relative context.build_dir [ dir1; dir2; basename ]
+;;
+
+type kind =
+  [ `File
+  | `Directory
+  ]
+
+module Spec = struct
+  type ('path, 'target) t =
+    { target : 'target
+    ; url : Loc.t * OpamUrl.t
+    ; checksum : (Loc.t * Checksum.t) option
+    ; kind : kind
+    }
+
+  let name = "source-fetch"
+  let version = 1
+  let bimap t _ g = { t with target = g t.target }
+  let is_useful_to ~memoize = memoize
+
+  let encode_loc f (loc, x) =
+    Dune_lang.List
+      (* TODO use something better for locs here *)
+      [ Dune_lang.atom_or_quoted_string (Loc.to_file_colon_line loc); f x ]
+  ;;
+
+  let encode { target; url; checksum; kind } _ encode_target : Dune_lang.t =
+    List
+      ([ Dune_lang.atom_or_quoted_string name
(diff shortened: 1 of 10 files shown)


src/dune_rules/fetch_rules.ml
```diff
@@ -0,0 +1,275 @@
+open Import
+open Memo.O
+module Gen_rules = Build_config.Gen_rules
+
+include struct
```
Leonidas-from-XIV, line 5: What is the purpose of including it instead of writing this on the top-level?

src/dune_rules/fetch_rules.ml
```diff
@@ -0,0 +1,275 @@
...
+      Rules.collect_unit (fun () ->
+        url_of_checksum ctx checksum
+        >>= function
+        | None ->
+          User_error.raise
+            [ Pp.textf "unknown checksum %s" (Checksum.to_string checksum) ]
+        | Some (url, checksum) -> gen_rules_checksum url checksum)
+    in
+    Gen_rules.make rules ~directory_targets |> Memo.return
+  | _ -> Memo.return Gen_rules.no_rules
```
Leonidas-from-XIV, line 192: What is `compontents` here? Is silently no-oping in that case the right thing to do?

src/dune_rules/fetch_rules.ml
```diff
@@ -0,0 +1,275 @@
...
+  | `Directory ->
+    Copy.action ~src_dir:src ~dst_dir:target
+    |> Action.Full.make
+    |> Action_builder.With_targets.return
+    |> Action_builder.With_targets.add_directories ~directory_targets:[ target ]
+;;
+
+let fetch ~target kind url checksum =
+  match checksum with
+  | Some (_, checksum) -> copy ~target kind checksum
```
Leonidas-from-XIV, line 264: Will this not fail when a checksum is specified but not present in the cache?

src/dune_rules/fetch_rules.ml
rgrinberg: You get to avoid the `Dune_pkg` prefix. E.g. in

```
module Checksum = Dune_pkg.Checksum
module Pkg = Dune_pkg.Lock_dir.Pkg
```

the `Dune_pkg` is repeated.

src/dune_rules/fetch_rules.ml
rgrinberg: components would be any other build path `_build/_fetch/` that doesn't match the pattern above. So it's correct to return no rules for these unknown paths. There's not much of a real difference though. We load the rules in a directory to build a particular target in it, so we could either:

1. Get an error loading some directory rules
2. Depending on that target and then dune telling us the target doesn't exist.

src/dune_rules/fetch_rules.ml
```diff
@@ -0,0 +1,275 @@
...
+let context_name = Context_name.of_string "_fetch"
+let context = Build_context.create ~name:context_name
+
+let make_target ~kind checksum =
+  let basename =
+    match kind with
+    | `File -> "file"
+    | `Directory -> "dir"
+  in
+  Path.Build.L.relative context.build_dir [ Checksum.to_string checksum; basename ]
```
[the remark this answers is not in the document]
rgrinberg: Yes, we could do that. I'll leave a TODO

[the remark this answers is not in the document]
rgrinberg: Actually, if we need to do this for the source fetching, we'd also need to do this for the package rules themselves. Those are going to have at least as many directories in `.pkg/$pkg_name`

src/dune_rules/fetch_rules.ml
rgrinberg: It will not. It will introduce a dependency on `_fetch/$checksum/<file|dir>` which will load the rules in that directory and generate the rule for this checksum (if it's a valid checksum).

gridbugs (approved):

Leonidas-from-XIV (approved):

PR #9936, opened 2024-02-06
chore: update version of melange to v3

### LINKED ISSUE #9935: melange: update test version to 3.0.0
Hi,
We're trying to upgrade the OCaml version that we use for tests to 5.1, but at the moment the test suite requires melange `1.0.0` which is 4.14 only.

Usually I'd do it but in looks like it's a fairly involved change that's not just replacing `bs.raw` by `mel.raw`. Can we get a hand @anmonteiro or @jchavarri ? Thanks!

jchavarri:
Fixes #9935.

- Removes `rescript-syntax` tests, as a continuation of https://github.com/ocaml/dune/pull/9884#discussion_r1474482961.
- Replaces usages of `bs-D` with either `mel-g` when possible, or just warnings for merlin tests
- Replace `bs.raw` with `mel.raw`
- Add `melange.node` when needed

The changes in `emit-with-runtime-deps-edge-cases.t` (there's a whole bunch of artifacts listed now) might be controversial.


Makefile
```diff
@@ -25,7 +25,7 @@ ppx_inline_test \
 ppxlib \
 ctypes \
 "utop>=2.6.0" \
-"melange=1.0.0"
+"melange=3.0.0-414"
```
emillon, line 28:
```suggestion
"melange>=3.0.0"
```

to easily allow `-51`, we'll add a bound if this becomes necessary

test/blackbox-tests/test-cases/melange/emit-with-runtime-deps-edge-cases.t
```diff
@@ -137,17 +142,87 @@ Test depending on external paths
...
   $ cat > external/main.ml <<EOF
-  > let dirname = [%bs.raw "__dirname"]
+  > let dirname = [%mel.raw "__dirname"]
   > let file_path = "./assets/file.txt"
   > let file_content = Node.Fs.readFileSync (dirname ^ "/" ^ file_path) \`utf8
   > let () = Js.log file_content
   > EOF
 
   $ dune build @mel --display=short
+          melc external/external-output/node_modules/melange.js/caml.js
```
emillon, line 157: since you mentioned it, is this list useful or should we pipe that into `grep -v node_modules`?

jchavarri: I think we can remove `--display` flag altogether, if @anmonteiro is ok with it.

anmonteiro (approved):

emillon:
that's great, thanks!

Issue #9919, opened 2024-02-05
In pkg files, pforms inside string interpolations should be substituted with absolute paths


gridbugs:
Currently section pforms in string interpolations such as `--prefix=%{prefix}` are substituted with relative paths while section pforms on their own are substituted with absolute paths. They should both be substituted by absolute paths.

gridbugs:
Repro: https://github.com/ocaml/dune/pull/9920

rgrinberg:
There's still some premature stringification going on in `String_with_vars.t`. I think you can fix this by modifying `String_with_vars.expand_result` to add a `Concat` constructor like i did with slang.

Issue #7502, opened 2023-04-05
[RFC] Shared Build Sandboxes


rgrinberg:
This RFC is created in the context of package management, but the proposal here could be used to address issues such as https://github.com/ocaml/dune/issues/6005 as well.

## Problem

When using dune as a package manager, our goal is to build opam packages with dune. This is easy enough to do in a dune workspace, we just pick a special directory and build all the packages there. The problem is evident when we try to share built packages between workspaces. Would like to share these artifacts by using the dune cache and not creating our own ad-hoc caching mechanism. Unfortunately, this isn't possible because all workspace local rules leak the path of dune workspace itself, which makes all the package artifacts non-reusable.

## Solution

To solve this problem, all that we really need is a way to setup the package build rules in a way that they produce the same paths no matter where the workspace is from.

We have a mechanism that allows us to relocate executing actions to directories already, it's called sandboxing. All that we need to do is to prepare sandbox paths that are shareable across all workspaces. Consider the following scheme:

1. We maintain all sandboxes in a single directory: `$HOME/.dune/sandboxes/<hash>`
2. The hash is the digest of the action and its dependencies. This is so that no matter what dune instance is used to build the package, it will have the same hash if it's building the same package.
3. Before building anything in the sandbox, we acquire a file lock inside this sandbox.
4. If we succeed acquiring the lock, we execute the action and populate the dune cache. Then we release the lock.
5. If we don't acquire the lock, we can just wait until we get the lock and then check if it was written to the cache. If it wasn't, the action must have failed and we'll re-run to see the failure.

### `chroot`

Although it's somewhat portable, it's dismissed because it requires elevated privileges

### cgroups

Using cgroups (or any of its API wrappers such as bubblewrap) works very well on Linux, but it's not portable enough.

PR #9940, opened 2024-02-07
pkg: fix bug preventing paths from being sandboxed

gridbugs:
Pform variables which appear in build/install commands that refer to paths (e.g. `%{target}`) are resolved in two steps. First the normal pform substitution is applied, replacing the variable with some path, then the path is further transformed to refer to the corresponding path within the sandbox specific to the package being built/installed.

Path pform variables which appeared in string interpolations (such as `--target=%{target}`) were erroneously not subject to the second transformation, meaning that they were left as the original, unsandboxed path. This was due to string interpolations being collapsed into strings too early so that path information was lost at the point when the sandboxing transformation is applied.

This fixes this by introducing the concept of "deferred concatenations" which represent a sequence of values to be concatenated in the future. This lets the path-iness of the path variable components of string interpolations to be preserved until the point when the sandboxing transformation is applied.

### DIFF
--- a/src/dune_lang/string_with_vars.ml
+++ b/src/dune_lang/string_with_vars.ml
@@ -132,17 +132,34 @@ let concat_rev = function
 ;;
 
 module Mode = struct
-  type _ t =
-    | Single : Value.t t
-    | Many : Value.t list t
-    | At_least_one : (Value.t * Value.t list) t
+  type (_, _) t =
+    | Single : (Value.Deferred_concat.t, Value.t) t
+    | Many : (Value.Deferred_concat.t list, Value.t list) t
+    | At_least_one
+        : ( Value.Deferred_concat.t * Value.Deferred_concat.t list
+            , Value.t * Value.t list )
+            t
 
-  let string : type a. a t -> string -> a =
+  let string
+    : type deferred_concat value. (deferred_concat, value) t -> string -> deferred_concat
+    =
+    let deferred_concat_string s = Value.Deferred_concat.singleton (Value.String s) in
     fun t s ->
+      match t with
+      | Single -> deferred_concat_string s
+      | Many -> [ deferred_concat_string s ]
+      | At_least_one -> deferred_concat_string s, []
+  ;;
+
+  let deferred_concat
+    : type deferred_concat value.
+      (deferred_concat, value) t -> Value.Deferred_concat.t -> deferred_concat
+    =
+    fun t v ->
     match t with
-    | Single -> Value.String s
-    | Many -> [ Value.String s ]
-    | At_least_one -> Value.String s, []
+    | Single -> v
+    | Many -> [ v ]
+    | At_least_one -> v, []
   ;;
 
   let invalid_multivalue ~source l ~what =
@@ -158,14 +175,34 @@ module Mode = struct
       ]
   ;;
 
-  let value : type a. source:Template.Pform.t -> a t -> Value.t list -> a =
+  let values_from_pform
+    : type deferred_concat value.
+      source:Template.Pform.t
+      -> (deferred_concat, value) t
+      -> Value.t list
+      -> deferred_concat
+    =
     fun ~source t x ->
-    match t, x with
+    match t, List.map x ~f:Value.Deferred_concat.singleton with
     | Many, x -> x
     | Single, [ x ] -> x
     | At_least_one, x :: l -> x, l
     | Single, _ -> invalid_multivalue ~source x ~what:"a single"
     | At_least_one, [] -> invalid_multivalue ~source x ~what:"at least one"
+  ;;
+
+  let force
+    : type deferred_concat value.
+      (deferred_concat, value) t -> deferred_concat -> dir:Path.t -> value
+    =
+    fun t v ~dir ->
+    match t with
+    | Single -> Value.Deferred_concat.force ~dir v
+    | Many -> List.map v ~f:(Value.Deferred_concat.force ~dir)
+    | At_least_one ->
+      let x, xs = v in
+      ( Value.Deferred_concat.force ~dir x
+      , List.map xs ~f:(Value.Deferred_concat.force ~dir) )
   ;;
 end
 
@@ -247,14 +284,25 @@ let is_prefix t ~prefix:want =
 module type Expander = sig
   type 'a app
 
-  val expand : t -> mode:'a Mode.t -> dir:Path.t -> f:Value.t list app expander -> 'a app
+  val expand
+    :  t
+    -> mode:(_, 'value) Mode.t
+    -> dir:Path.t
+    -> f:Value.t list app expander
+    -> 'value app
 
   val expand_result
     :  t
-    -> mode:'a Mode.t
+    -> mode:(_, 'value) Mode.t
     -> dir:Path.t
     -> f:(Value.t list, 'error) result app expander
-    -> ('a, 'error) result app
+    -> ('value, 'error) result app
+
+  val expand_result_deferred_concat
+    :  t
+    -> mode:('deferred_concat, _) Mode.t
+    -> f:(Value.t list, 'error) result app expander
+    -> ('deferred_concat, 'error) result app
 
   val expand_as_much_as_possible
     :  t
@@ -266,27 +314,34 @@ end
 module Make_expander (A : Applicative) : Expander with type 'a app := 'a A.t = struct
   open A.O
 
-  let expand_result
-    : type a.
+  let expand_result_deferred_concat
+    : type deferred_concat value.
       t
-      -> mode:a Mode.t
-      -> dir:Path.t
+      -> mode:(deferred_concat, value) Mode.t
       -> f:(Value.t list, 'error) result A.t expander
-      -> (a, 'error) result A.t
+      -> (deferred_concat, 'error) result A.t
     =
-    fun t ~mode ~dir ~f ->
+    fun t ~mode ~f ->
     match t.parts with
+    (* Special case where a list of values may be returned if [mode] is [Multi] or [At_least_one] *)
+    | [ Pform (source, p) ] when not t.quoted ->
+      let+ v = f ~source p in
+      Result.map v ~f:(fun v -> Mode.values_from_pform mode ~source v)
     (* Optimizations for some common cases *)
     | [] -> A.return (Ok (Mode.string mode ""))
     | [ Text s ] -> A.return (Ok (Mode.string mode s))
-    | [ Pform (source, p) ] when not t.quoted ->
-      let+ v = f ~source p in
-      Result.map v ~f:(Mode.value mode ~source)
+    (* General case covering quoted variables and strings made up of text parts
+       and variable parts. *)
     | _ ->
+      (* Quoted strings with vars concatenate variable contents with spaces
+         and unquoted ones concatenate variable contents with no spaces (unless
+         the entire string with vars is a single variable in which case it
+         resolves to multiple values and no concatenation occurs). *)
+      let inner_sep = if t.quoted then Some " " else None in
       let+ chunks =
         A.all
           (List.map t.parts ~f:(function
-            | Text s -> A.return (Ok [ s ])
+            | Text s -> A.return (Ok (Value.Deferred_concat.singleton (Value.String s)))
             | Error (_, msg) ->
               (* The [let+ () = A.return () in ...] is to delay the error until
                  the evaluation of the applicative *)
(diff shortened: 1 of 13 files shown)


gridbugs:
Fixes https://github.com/ocaml/dune/issues/9919

src/dune_lang/value.mli
```diff
@@ -34,6 +34,33 @@ module L : sig
...
   val deps_only : t list -> Path.t list
   val dirs : Path.t list -> t list
+
+  (** Concatenate values to a string, separating each value by a space. *)
   val concat : t list -> dir:Path.t -> string
+
   val to_strings : t list -> dir:Path.t -> string list
 end
+
+module Deferred_concat : sig
```
Leonidas-from-XIV, line 44: Given that deferring can happen also for other reason, maybe `Deferred` is a better name?

src/dune_lang/string_with_vars.ml
```diff
@@ -295,14 +310,27 @@ module Make_expander (A : Applicative) : Expander with type 'a app := 'a A.t = s
...
-                  let vs = Mode.value Many v ~source in
-                  List.map ~f:(Value.to_string ~dir) vs))))
+                Value.Deferred_concat.concat_values v ~sep:inner_sep)))
       in
       Result.map (Result.List.all chunks) ~f:(fun chunks ->
-        Mode.string mode (String.concat (List.concat chunks) ~sep:""))
+        `Deferred_concat (Value.Deferred_concat.concat chunks ~sep:None))
+  ;;
+
+  let expand_result
```
Leonidas-from-XIV, line 319: I wonder, could maybe `expand_result` be replaced by `expand_result_deferred_concat`? To avoid introducing the need to distinguish between `Value.t` and `Value.Deferred_concat.t` (so `Value.t` becomes the deferred version and is then forced). I would prefer not adding additional concept of deferred values, so if values are deferred by default and always have to be forced that could keep the mental model simpler (since `expand_result_deferred_concat` called under the hood anyway).

Leonidas-from-XIV (commented):
To me the two phase resolution seems like a good idea! My comments are about decreasing the mental burden of having to distinguish between values and deferred values.

src/dune_lang/string_with_vars.mli
```diff
@@ -114,6 +114,20 @@ module type Expander = sig
...
+      pforms and strings made up of a mix of pforms and text) then a
+      [`Deferred_concat _] is returned containing each part of the result. This
+      preserves the [Value.t] of each pform rather than collapsing path values
+      to strings (as done by [expand_result] and [expand]). This is why
+      [expand_result_deferred_concat] does not need a [dir] argument. *)
+  val expand_result_deferred_concat
+    :  t
+    -> mode:'a Mode.t
+    -> f:(Value.t list, 'error) result app expander
+    -> ([ `Values of 'a | `Deferred_concat of Value.Deferred_concat.t ], 'error) result
```
rgrinberg, line 128: The type signature doesn't look right to me. The index in `'a Mode.t` shouldn't be discarded in `Value.Deferred_concat.t`. We should be able to only expand `Value.Deferred_concat.t` to a list of values or a single value based on the `'a Mode.t`.

gridbugs: > The index in 'a Mode.t shouldn't be discarded in Value.Deferred_concat.t
It's safe to discard the `'a` because regardless of `'a`, `Deferred_concat` will only be returned in cases where only a single value would have been returned anyway.

> We should be able to only expand Value.Deferred_concat.t to a list of values or a single value based on the 'a Mode.t.
That doesn't sound right to me. A `Deferred_concat.t` always represents a single value. If `'a` is `Multi` then `%{targets}` would expand to all the targets as separate values (say `foo.txt` and `bar.txt`) whereas `"%{targets}` would expand to the single value `foo.txt bar.txt`. All cases involving concatenation expand to a single value and the index `'a` shouldn't affect the treatment of such values.

src/dune_lang/string_with_vars.ml
gridbugs: Interesting idea. Making values deferred by default might tidy up some of the messiness introduced in this change.

src/dune_lang/value.mli
rgrinberg: How about the name `Builder`? It's quite similar to a string builder from Java.

src/dune_lang/string_with_vars.mli
rgrinberg: Okay, indeed `Deferred_concat.t` is possible for all modes. It's rather confusing though.

src/dune_lang/value.ml
```diff
@@ -79,3 +79,40 @@ module L = struct
...
 end
+
+module Deferred_concat = struct
+  type value = t
+  type t = value list
+
+  let singleton = List.singleton
+
+  (* Helper to place a value [sep] between every value in a list *)
+  let rec list_intersperse xs ~sep =
```
rgrinberg, line 90: Could you move  this function to stdune?

src/dune_lang/string_with_vars.mli
gridbugs: Yeh this function's type is a bit messy. I think making values deferred by default will simplify it.

gridbugs:
I've updated `String_with_vars` so that expanding produces deferred values and updated the codebase so all the places where we expand sws now also explicitly force the deferred result. To help with this I added a function `String_with_vars.Mode.force` which forces a deferred result with the arity implied by the mode.

If requiring callers to explicitly force whenever expanding a `String_with_vars.t` is too much of a hastle, we could add some functions to `Expander` that return the forced values.

src/dune_lang/value.mli
gridbugs: I've gone with `Deferred` for now. @rgrinberg are you worried that people might think it's got something to do with `Async`?

gridbugs: This type could just as well be implemented as `type t = Path.t -> Value.t`. It lets you commit to doing operations on lists of values but the operations don't occur until a path is provided so stringification of paths in values can occur relative to the given path. I think it's apt to think of this as deferring those operations even though internally it's implemented as a list of values to concatenate.

rgrinberg:
Not a fan of the changes to the rest of the rules. We should remember that this is all just a workaround for configure scripts that refuse to accept relative paths. We shouldn't have to care about this stuff anywhere else in the rules.

src/dune_lang/string_with_vars.ml
```diff
@@ -132,17 +132,26 @@ let concat_rev = function
 ;;
 
 module Mode = struct
-  type _ t =
-    | Single : Value.t t
-    | Many : Value.t list t
-    | At_least_one : (Value.t * Value.t list) t
+  type (_, _) t =
+    | Single : (Value.Deferred.t, Value.t) t
+    | Many : (Value.Deferred.t list, Value.t list) t
```
rgrinberg, line 137: Is it ever possible to expand into a `Value.Deferred.t list` anywhere?

src/dune_lang/value.mli
```diff
@@ -35,6 +35,40 @@ module L : sig
...
   val deps_only : t list -> Path.t list
   val dirs : Path.t list -> t list
-  val concat : t list -> dir:Path.t -> string
   val to_strings : t list -> dir:Path.t -> string list
 end
+
+module Deferred : sig
+  type value = t
+
+  (** A computation which produces a [value] when forced. Forcing requires a
```
rgrinberg, line 44: The description it's a computation, but it really is just a value. It can be compared, serialized, etc. A computation is a function which thankfully it is not.

src/dune_lang/value.mli
rgrinberg: The type `type t = Path.t -> Value.t` could be used implement far more than the behavior in this module. That's the main problem with this name - it's too general for the relatively specific functionality it represents. In any case, you can leave it as `Deferred`. It's only needed for the package management rules, no use to worry too much about it

gridbugs:
> Not a fan of the changes to the rest of the rules. We should remember that this is all just a workaround for configure scripts that refuse to accept relative paths. We shouldn't have to care about this stuff anywhere else in the rules.

This was my attempt to string with vars expansion return a deferred value by default. I think what we should do instead is introduce a new variant of expand that returns a deferred value as I orignally had and leave the existing functions returning a regular value.

src/dune_lang/string_with_vars.ml
gridbugs: Yep, see the `Action_expander.slang_expander` function

src/dune_lang/value.mli
gridbugs: Would the original name `Deferred_concat` be more suitable in your opinion?

rgrinberg:
Yes, I think that was better. You could also re-implement the old API in terms of the new one.

It's worth remembering that this hack isn't the only way to solve the problem. With the shared sandbox proposal, we could safely expand absolute paths and not worry about relocation. All in all, I'm not ready to commit to this approach yet. If there were other advantages to deferring concatenation, this change would be more acceptable.

gridbugs:
I don't really see this as a hack.

It clarifies the distinction between evaluating `String_with_vars.t` and stringification of values. The only reason `expand` takes a `dir:Path.t` argument is so that it can stringify paths in values it needs to concatenate, and having a version of `expand` that doesn't take a path and returns a deferred value hopefully clarifies this.

Previously you added a similar feature to slang evaluation (which I've updated to use `Value.Deferred.t`) so this isn't the first time we've needed this feature. That change was also to allow sandboxing to rewrite paths IIRC. Even if we implemented shared sandboxing (https://github.com/ocaml/dune/issues/7502) I don't see how we avoid needing to rewrite paths to refer into the sandbox.

> We should remember that this is all just a workaround for configure scripts that refuse to accept relative paths.

I disagree. This is about treating pform variables the same regardless of whether they appear on their own or in a string interpolation. Even if all configure scripts accepted relative paths I would still consider it an error if `%{prefix}` on its own expands to `_private/default/.pkg/test/target` while the string interpolation `--prefix=%{prefix}` expands to `--prefix=../target`.

> With the shared sandbox proposal, we could safely expand absolute paths and not worry about relocation.

I don't understand how this would allow us to expand the path in `--prefix=%{prefix}` to an absolute path if the path-ness is removed by the concatenation. Or are you saying that the pform itself would expand to an absolute path?

rgrinberg:
I think I might be not understanding this PR fully.

> This is about treating pform variables the same regardless of whether they appear on their own or in a string interpolation. Even if all configure scripts accepted relative paths I would still consider it an error if %{prefix} on its own expands to _private/default/.pkg/test/target while the string interpolation --prefix=%{prefix} expands to --prefix=../target.

Where is this possible outside of the package management rules and the `Run_with_path` action? If this is fixing bugs elsewhere in the rules, I'm indeed underestimating the value of this new API.

gridbugs:
> Where is this possible outside of the package management rules and the Run_with_path action? If this is fixing bugs elsewhere in the rules, I'm indeed underestimating the value of this new API.

The bug happens when dune transforms a path produced by expanding a `String_with_vars.t`. As far as I know this only happens in package management rules. 

But the fact that expanding a `String_with_vars.t` returns `Value.t`s with the `Path` and `Dir` constructors only if the string is a single unquoted pform and the `String` constructor (stringifying paths) in all other cases is a bit confusing to me. It could lead to people writing code that checks the constructor of a returned `Value.t` and does something special for paths when encountering `Path` and `Dir`, and incorrectly assume that the `String` constructor doesn't contain paths (as we did in package management). Expanding `String_with_vars.t` should either always stringify its output (which would make sandboxing via path rewriting impossible), or always preserve path information appear in string interpolations (this PR).

gridbugs:
I've renamed the module back to `Deferred_concat` and made `expand` and `expand_result` returned `Value.t`s, undoing all the changes to other rules.

src/dune_lang/value.mli
gridbugs: Reworded this comment

src/dune_lang/value.ml
gridbugs: done

rgrinberg:
Okay, that's more clear, thanks. To evaluate the new API, I consider these two points:

1. How general is the API? The API introduced here is so far only useful for a single action for package building rules.
2. Is the API simpler or more complex? At first glance, the API is more complex because it offers a whole module to handle what was a single parameter.

To only say the API is more clear is rather vague to me. Yes, it separates the expansion from the stringification of the paths. But before I count that as a good thing, I need to see where this distinction is useful. Sure, it helps us with the package management rules, which as you've pointed out I've introduced myself. However, I did that in a localized way that did not disturb any existing code because I wasn't convinced if this distinction is useful to track in general. I'm not yet certain whether the distinction is useful, so I'm trying to look for more examples where we need to care.

A couple more statements that deserve some clarification:

> I disagree. This is about treating pform variables the same regardless of whether they appear on their own or in a string interpolation. Even if all configure scripts accepted relative paths I would still consider it an error if %{prefix} on its own expands to _private/default/.pkg/test/target while the string interpolation --prefix=%{prefix} expands to --prefix=../target.

If the configure scripts accepted relative paths, there would never be a need to delay any expansion in the first place. We could eagerly stringify all paths relative to our directory and it would always be correct.

> Expanding String_with_vars.t should either always stringify its output (which would make sandboxing via path rewriting impossible)

Sandboxing is only impossible if executing the action rejects relative paths as we have in package management.

By the way, another issue I was curious about was the performance impact. In theory, it should be faster to expand a template and stringify the paths right away rather than preserve the intermediate representation. But looking at the benchmarks, it seems like the extra allocation is not impactful. So perhaps I overreacted and we could consider this new API.

rgrinberg (approved):
Just noticed that you reverted the API change. As a bug fix, this PR looks good to go. For the new API, a separate PR would be welcome to discuss the benefits. I did not think that the new API had no merit, just that it was unobvious to me at first glance.

gridbugs:
Cheers. By API change do you mean changing `expand` to returned a deferred value by default and require callers to explicitly force it?

rgrinberg:
> By API change do you mean changing expand to returned a deferred value by default and require callers to explicitly force it?

Yes, this the part that I'm suspicious of.

PR #9941, opened 2024-02-07
refactor: optimize [Path.reach]

rgrinberg:

### DIFF
--- a/otherlibs/stdune/src/path.ml
+++ b/otherlibs/stdune/src/path.ml
@@ -57,25 +57,9 @@ end = struct
   let to_string t = t
   let hash = String.hash
   let compare = String.compare
+  let equal = String.equal
   let root = "."
   let is_root t = Ordering.is_eq (compare t root)
-
-  let to_list =
-    let rec loop t acc i j =
-      if i = 0
-      then String.take t j :: acc
-      else (
-        match t.[i - 1] with
-        | '/' -> loop t (String.sub t ~pos:i ~len:(j - i) :: acc) (i - 1) (i - 1)
-        | _ -> loop t acc (i - 1) j)
-    in
-    fun t ->
-      if is_root t
-      then []
-      else (
-        let len = String.length t in
-        loop t [] len len)
-  ;;
 
   let parent t =
     if is_root t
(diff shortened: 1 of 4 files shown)


snowleopard:
Could you do some microbenchmarks to show the difference? Seems like quite a lot of fiddly code to add unless there is a solid win.

snowleopard:
Added @tov as a reviewer.

rgrinberg:
Here are the results from micro benchmarking:
```
┌────────────────────────────────────────────────────────────────────────────┬──────────┬─────────┬────────────┐
│ Name                                                                       │ Time/Run │ mWd/Run │ Percentage │
├────────────────────────────────────────────────────────────────────────────┼──────────┼─────────┼────────────┤
│ [bench/micro/path_bench.ml] reach:from root long path                      │   6.37ns │         │      1.33% │
│ [bench/micro/path_bench.ml] reach:from root short path                     │   6.35ns │         │      1.33% │
│ [bench/micro/path_bench.ml] reach:reach root from short path               │  22.05ns │         │      4.60% │
│ [bench/micro/path_bench.ml] reach:reach root from long path                │ 393.12ns │  11.00w │     82.00% │
│ [bench/micro/path_bench.ml] reach:reach long path from similar long path   │ 479.44ns │   4.00w │    100.00% │
│ [bench/micro/path_bench.ml] reach:reach short path from similar short path │  61.28ns │   4.00w │     12.78% │
└────────────────────────────────────────────────────────────────────────────┴──────────┴─────────┴────────────┘
```

Which should be  compared to https://github.com/ocaml/dune/pull/10074

As we can see, this new implementation is dramatically faster across the board.

snowleopard:
Awesome improvement! I wonder how this compares to Jenga's implementation of `Path.reach`. I remember doing some optimisation there a couple of years ago... We should double check!

snowleopard:
Btw, there is a lot of finicky code here. We should definitely add some thorough tests to catch possible bugs.

rgrinberg:
> Btw, there is a lot of finicky code here. We should definitely add some thorough tests to catch possible bugs.

Agreed. Before this PR, I've added 17 unit tests to make sure all the various cases are checked. Those tests plus the monorepo benchmark successfully completing make me fairly confident the semantics are the same.

tov (approved):

PR #9946, opened 2024-02-08
Add test for `String_with_vars.t`

gridbugs:
I found it useful to have a test for all the `String_with_vars.t` cases in one place.


rgrinberg:
Don't think this belongs in `pkg/` as this is just expanding regular rules

gridbugs:
> Don't think this belongs in `pkg/` as this is just expanding regular rules

Agreed. Fixed.

rgrinberg (approved):

PR #9956, opened 2024-02-09
favour odocls that are not dependencies

EmileTrotignon:
When calling sherlodoc on with @doc-new, all the .odocl are indexed by sherlodoc. It makes sense to index the dependencies, but the entries from the dependencies should be given less priority as search results. 

This PR passes a `--favoured` flag to sherlodoc, this flag identifies the .odocls that are not external. In turn, sherlodoc ranks these entries higher.


jonludlam:
Looks correct to me. Just needs a rebase.

jonludlam (approved):

rgrinberg:
Also, please DCO

PR #9975, opened 2024-02-12
feature(pkg): proper build instructions for dune packages

rgrinberg:
When pinning a dune package, we don't need to record the build command
because we are going to interpret it ourselves based on the version in
the dune-project file.

In the future, this will give us ability to build such packages without
spawning any additional dune binaries.

### DIFF
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -42,9 +42,58 @@ module Pkg_info = struct
   let default_version = Package_version.of_string "dev"
 end
 
+module Build_command = struct
+  type t =
+    | Action of Action.t
+    | Dune
+
+  let equal x y =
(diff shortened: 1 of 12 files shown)


gridbugs (approved):
Sounds useful! What's the plan for packages that build with a different version of dune than the one currently installed? In particular what if the project transitively depends on a version of dune newer than the current one (e.g. if the user hasn't updated their dune in a while)?

rgrinberg:
At the moment we don't support it. The issue is that newer versions of dune can generate metadata that older versions of dune cannot understand. So this cannot work in general unless we agree on some sort of scheme to maintain backwards compat. I wouldn't be particularly worried about this however. Users already need to install dune in the first place to get started. They might as well continue with this same process to keep it updated.

PR #9978, opened 2024-02-12
fix typo in dynamic_include message

emillon:


ElectreAAS (approved):
Obviously correct, thanks :)

PR #9980, opened 2024-02-12
Bump ocaml version in tests, update hashes

ElectreAAS:
This PR contains fixes for the first 9 items in the list in #9954.

While it works and compiles alright, the other tests mentionned in the tracking issue haven't been changed and hence don't pass.

Issue #9954, opened 2024-02-09
Tracking: Upgrading the OCaml version we use in tests to 5.1.1


ElectreAAS:
Right now, we use OCaml 4.14.1 in the tests of dune :point_down: 
https://github.com/ocaml/dune/blob/e0bd14d154020e369df5e091ff79f0f31681799a/Makefile#L35

There's been an effort to bump that number to 5.1.1. I'll let other maintainers fill in the rationale below.
This is the list of tests that pass on `main` with 4.14.1, but don't with 5.1.1.

- [ ] Bump the version number everywhere needed in the docs / comments
- [ ] Bump the version number in the Makefile.
- [ ] Update hashes in
  - [ ] `test/expect-tests/persistent_tests.ml`
  - [ ] `test/blackbox-tests/test-cases/patch-back-source-tree.t`
  - [ ] `test/blackbox-tests/test-cases/dune-cache/mode-copy.t`
  - [ ] `test/blackbox-tests/test-cases/dune-cache/mode-hardlink.t`
  - [ ] `test/blackbox-tests/test-cases/dune-cache/repro-check.t`
  - [ ] `test/blackbox-tests/test-cases/dune-cache/trim.t`
- [ ] `error_messages_separated`: Since ocaml/ocaml#12024 (included in 5.1.0), there are newlines inserted in between errors for a given file. We just need to update the comments and promote.
  - [ ] Thanks to the above, `test/blackbox-tests/test-cases/wrapped-transition.t/run.t` just needs some newlines added in the expect portion.
  - [ ] Same for `test/blackbox-tests/test-cases/vendor/alerts.t/run.t`
- [x] `test/blackbox-tests/test-cases/bigarray.t/run.t` incorrectly checks whether bigarray is available or not. In 5.0+, bigarray is part of the stdlib and not a separate library anymore, but isn't detected by the call to `select`. See also ocaml/ocaml#10896. Would be fixed by #10011.
- [ ] `test/blackbox-tests/test-cases/describe/{describe-workspace-pp.t & describe.t}`: the diff says `(source_dir /FINDLIB/ocaml/compiler-libs) -> (source_dir /FINDLIB/compiler-libs)` in several places. I have no clue what this means.
- [ ] `test/blackbox-tests/test-cases/findlib-dynload.t/run.t`: we have something that also looks findlib-related: a line about findlib finding `threads.posix` is removed.
[...]

PR #9981, opened 2024-02-12
Future bug report: `select` is confused by bigarray on ocaml 5.1.1

ElectreAAS:
This PR includes #9980, but the interesting part is the second commit.
The difference between `test/.../c/dune` and `test/.../d/dune` is only the order of the branches of the `select`. That should not change anything, and yet it does.
`bigarray` is both available and not available.

If we find a fix that would solve item 13 in #9954.


ElectreAAS:
Fixed in #10011.

PR #10011, opened 2024-02-13
fix: `select` with `bigarray` in OCaml 5

moyodiallo:
This is related to #9981.

It turns out the fact that `bigarray` is ignored is not known to `select`.


rgrinberg:
Thanks for the fix. Sorry for the delay, but changes to the library database always require a very careful look.

PR #10026, opened 2024-02-14
enabled_if: add test that targets cmo of a disabled lib

jchavarri:
Extracted from https://github.com/ocaml/dune/pull/9839#issuecomment-1942829730.


rgrinberg:
Thanks for the test. The current behavior is definitely not great. The error should include that the target is missing due to a disabled library.

Issue #9979, opened 2024-02-12
Dune links against wrong cstubs in byte mode


talex5:
When using byte mode, I see odd build errors, like this:

```
$ dune build ./examples/hello/main.bc
File "_none_", line 1:                 
Error: Error while linking lib_eio/unix/eio_unix.cma(Eio_unix__Private):
       The external function `eio_unix_readlinkat' is not available
```

stracing the build, I see:

```
175082 newfstatat(AT_FDCWD, "dlleio_unix_stubs.so", 0x7ffe35250a50, 0) = -1 ENOENT (No such file or directory)
175082 newfstatat(AT_FDCWD, "/home/user/.opam/5.1.1/lib/cstruct/../stublibs/dlleio_unix_stubs.so", {st_mode=S_IFREG|0755, st_size=33456, ...}, 0) = 0
175082 openat(AT_FDCWD, "/home/user/.opam/5.1.1/lib/cstruct/../stublibs/dlleio_unix_stubs.so", O_RDONLY|O_CLOEXEC) = 4
```
If I remove the installed version of `dlleio_unix_stubs.so` then it does find the local one instead.

## Reproduction

```
> docker run --rm -it ocaml/opam:debian-11-ocaml-5.1

$ opam-2.1 install eio_main.0.14
$ git clone https://github.com/ocaml-multicore/eio.git
$ cd eio
$ cat >examples/hello/dune <<EOF
(executable
 (name main)
 (modes byte)
 (libraries eio_main))
EOF

$ dune build ./examples/hello/main.bc
File "_none_", line 1:                 
Error: Error while linking lib_eio/unix/eio_unix.cma(Eio_unix__Private):
       The external function `eio_unix_readlinkat' is not available

$ opam-2.1 remove eio
$ dune build ./examples/net/main.exe
[ success ]
```

## Specifications

- Version of `dune` (output of `dune --version`): 3.13.1
- Version of `ocaml` (output of `ocamlc --version`): 5.1.1
- Operating system (distribution and version): Debian 11.8

nojb:
Are you able to paste here a copy of the build log of a failing build? You can find it in the file `_build/log` after the build fails.

talex5:
Here's the log that results from running the reproduction commands above (I ran the build twice, so the log only has the failing part):

```
$ cat _build/log
# dune build ./examples/hello/main.bc
# OCAMLPARAM: unset
# Shared cache: disabled
# Shared cache location: /home/opam/.cache/dune/db
[...]

PR #10028, opened 2024-02-14
Add failing test for #9979

nojb:
See #9979


rgrinberg (approved):
LGTM. Why is the name of the test `install-stubs-stubs.t`?  Not sure why the extra `-stubs`?

nojb:
> LGTM. Why is the name of the test `install-stubs-stubs.t`? Not sure why the extra `-stubs`?

No reason; I renamed it with a better name. Thanks.

Issue #9952, opened 2024-02-09
Uniform Caching API in Dune_pkg


rgrinberg:
We currently have four caches in dune_pkg:

1. candidates cache in the solver
2. availability cache in the solver
3. mount cache in the pin handling
4. download cache in single_run_file_cache

They all have slight differences, but could probably use the same caching API. In particular, they should all create ivars before they do no any work (caches 1. and 2. don't do that).

@gridbugs could you take a look at this one?

gridbugs:
Taking a look at the different caches we have in package management and I'm not convinced it's worth it to unify them with a single api. Possibly it makes sense to unify the two caches used in the solver since they are both just maps (well a map and a table). But `single_run_file_cache` serves a very different purpose than the caches used in the solver - it doesn't cache ocaml values but rather files on disk created as a side effect of calling a function and expects users to open the files and read the contents themselves. I don't see a sensible through line between `single_run_file_cache` and the solver caches.

Also I'm not sure what you mean by the "mount cache in the pin handling". Are you referring to how all the functions in `mount.ml` create a fiber node before doing any work so that the result of the computation is cached in the ivar? This is different to the other types of cache mentioned since it's not associating keys with cache entries.

rgrinberg:
You can start with unifying 1 and 2 as a first step.

> Also I'm not sure what you mean by the "mount cache in the pin handling". Are you referring to how all the functions in mount.ml create a fiber node before doing any work so that the result of the computation is cached in the ivar? This is different to the other types of cache mentioned since it's not associating keys with cache entries.

I mean the code that populates the cache stored in this type:

```
  type state =
    { mutable traversed :
        (DB.t * Dune_lang.Package.t Package_name.Map.t) option Fiber.Ivar.t OpamUrl.Map.t
    }
```

gridbugs:
Even between 1 and 2 it's not clear what common logic could be moved to a common api. 1 is updated asynchronously with a fiber so its cache would need fiber's threaded through it (e.g. `val find_or_add : cache -> 'key -> f:(unit -> 'value Fiber.t) -> 'value Fiber.t)`). But 2 is updated synchronously and it doesn't seem worth it to thread fibers through the logic for checking if a package is available just to share a caching api.

[...]

PR #10034, opened 2024-02-15
pkg: add cache of asyncronously-computed values

gridbugs:
This replaces the ad-hoc cache of solver candidates with the added benifit that it guarantees that each value will be computed only once.

Part of https://github.com/ocaml/dune/issues/9952

### DIFF
--- a/src/dune_pkg/fiber_cache.ml
+++ b/src/dune_pkg/fiber_cache.ml
@@ -1,0 +1,51 @@
+open! Import
+
+type ('k, 'v) t =
+  { (* The cache stores results so that if an exception was raised while
+       computing a value, concurrent accesses to the cache can find out. *)
+    table : ('k, ('v, Exn_with_backtrace.t list) result Fiber.Ivar.t) Table.t
+      (* This module is stored so it can be reused to create a table from the cache. *)
+  ; key_module : (module Table.Key with type t = 'k)
+  }
+
+let create key_module = { table = Table.create key_module 1; key_module }
+
+let find_or_add t key ~f =
+  let open Fiber.O in
+  let* () = Fiber.return () in
+  match Table.find t.table key with
+  | Some ivar ->
+    let* value_result = Fiber.Ivar.read ivar in
+    (match value_result with
+     | Ok value -> Fiber.return value
+     | Error exns -> Fiber.reraise_all exns)
+  | None ->
+    let ivar = Fiber.Ivar.create () in
+    (* Add the empty ivar to the table before calling [f] so that if the key
+       requested again while [f] is running then the caller will wait on the
+       ivar that will eventually contain the result. *)
+    Table.set t.table key ivar;
+    let* value_result = Fiber.collect_errors f in
+    let* () = Fiber.Ivar.fill ivar value_result in
+    (match value_result with
+     | Ok value -> Fiber.return value
+     | Error errors -> Fiber.reraise_all errors)
+;;
+
+let to_table t =
+  let open Fiber.O in
+  let* () = Fiber.return () in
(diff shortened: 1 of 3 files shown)


src/dune_pkg/fiber_cache.ml
```diff
@@ -0,0 +1,42 @@
...
+       called in the same fiber note so there's no possibility for a
+       conflicting entry to be added to the table in between those calls. *)
+    Table.add_exn t.table key ivar;
+    let* value = f () in
+    let+ () = Fiber.Ivar.fill ivar value in
+    value
+;;
+
+let to_table t =
+  let open Fiber.O in
```
rgrinberg, line 31: This needs a `let* () = Fiber.return ()` as well

rgrinberg:
LGTM. I think you can also use this to replace a similar table in `pin_stanza`

src/dune_pkg/fiber_cache.ml
gridbugs: Updated. Can you remind me why this is necessary. Is it just so that no work happens after the function is called until the fiber is scheduled? Why is that important?

gridbugs:
> LGTM. I think you can also use this to replace a similar table in `pin_stanza`

I think I can also replace some of `single_run_file_cache` with this module. I'll do both in a separate PR.

src/dune_pkg/fiber_cache.ml
rgrinberg: It's important because it's allowed to bind on the fiber more than once. So the side effect should always be repeated once per bind.

src/dune_pkg/fiber_cache.ml
```diff
@@ -0,0 +1,43 @@
...
+    let ivar = Fiber.Ivar.create () in
+    (* Add the empty ivar to the table before calling [f] so that if the key
+       requested again while [f] is running then the caller will wait on the
+       ivar that will eventually contain the result.
+
+       Note that this won't raise because [Table.find] and [Table.add_exn] are
+       called in the same fiber note so there's no possibility for a
+       conflicting entry to be added to the table in between those calls. *)
+    Table.add_exn t.table key ivar;
+    let* value = f () in
```
rgrinberg, line 25: To be extra safe, you can use `Fiber.with_error_handler` to evict `key` in case of an error. That way subsequent calls to the same will not deadlock.

gridbugs: Updated to handle the case where [f] raises. Concurrent calls (while the ivar is present in the table) will raise, and subsequent calls (after the ivar is removed) will attempt to recompute the value.

src/dune_pkg/fiber_cache.ml
```diff
@@ -0,0 +1,70 @@
...
+  let ivar = Fiber.Ivar.create () in
+  Table.set t.table key ivar;
+  let* value_result = Fiber.collect_errors f in
+  let* () = Fiber.Ivar.fill ivar value_result in
+  match value_result with
+  | Ok value -> Fiber.return value
+  | Error errors ->
+    (* Remove the ivar from the table so that future calls to [find_or_add]
+       attempt to recompute the value. *)
+    Table.remove t.table key;
```
rgrinberg, line 33: I think this behavior is weird b/c it is non deterministic. We should always either recompute on error, or we raise the same errors.

gridbugs: Fair enough. Updated so that it always raises. I thought about having it attempt to recompute but it was getting complicated and in most cases I would expect the recomputation to fail in the same way.

rgrinberg (approved):

PR #10035, opened 2024-02-15
merlin: show behavior with non-default context

jchavarri:
The test shows how `dune ocaml merlin` errors out when trying to get artifacts generated in a non-default context. Not sure if this is the expected behavior, but it might be helpful to have it documented.

### DIFF
--- a/test/blackbox-tests/test-cases/merlin/multi-context.t
+++ b/test/blackbox-tests/test-cases/merlin/multi-context.t
@@ -1,0 +1,47 @@
+Showcase dune ocaml merlin behavior with libraries built in the non-default context
+
+  $ ocamlc_where="$(ocamlc -where)"
+  $ export BUILD_PATH_PREFIX_MAP="/OCAMLC_WHERE=$ocamlc_where:$BUILD_PATH_PREFIX_MAP"
+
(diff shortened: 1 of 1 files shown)


rgrinberg:
Yes, this is expected. There's a `merlin` field that can be used to determine which context is used for merlin.

jchavarri:
I see, thanks for the pointers. I could find the `test/blackbox-tests/test-cases/merlin/default-based-context.t` test, which contains similar checks than the ones added in this PR.

For additional context 😅 I am investigating how to make dune choose one context or another more dynamically, as we would like to have ocaml-lsp query dune for artifacts in a specific context (as part of the work in #10222). For this, I understand something like the Dune context configuration would not be enough because it can't be changed from the clients. What I'm not sure about is if the change would need to be added in the interface between ocaml-lsp and dune (I understand, the `ocaml-merlin` server).

I could find the commands exposed by the ocaml-merlin server:

https://github.com/ocaml/dune/blob/6c40cb12deb7b05dc1d947330324ecbd0e5b61c7/bin/ocaml/ocaml_merlin.ml#L27-L29

Would it be necessary to add some parameter to the `File` command to specify the context merlin is querying for?

PR #10060, opened 2024-02-17
fix: checking out git repos via rev_store

rgrinberg:
We must create the target directory before extracting.

We can also use -C to instruct tar which directory to use rather than
doing chdir ourselves.

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -619,6 +619,10 @@ module At_rev = struct
     let git = Lazy.force Vcs.git in
     let tar = Lazy.force tar in
     let temp_dir = Temp.create Dir ~prefix:"rev-store" ~suffix:rev in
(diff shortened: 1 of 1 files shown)


Leonidas-from-XIV (approved):
The code looks good, hoewever the PR name doesn't seem to match what the code is doing :)

(Are you sure `tar -C` is portable though?)

rgrinberg:
`-C` is supported by both gnu and bsd.

In what way doesn't the description match?

I suppose another change that I did not describe is that I now untar into the temp dir and then move it into the target directory. This is to avoid leaving stray empty directories if the tar fails.

rgrinberg:
I see. The function's name is "check_out" so I thought this is the terminology we use. Maybe we should rename it to unpack.

Leonidas-from-XIV:
Fair enough, I don't think the naming is something worth splitting hairs about.

PR #10064, opened 2024-02-17
refactor(pkg): remove [Remote.equal]

rgrinberg:


Leonidas-from-XIV (approved):
Good catch! I wish there was a way to get a warning for things that are in interfaces that are not in libraries, since it is very easy to have unused code as long as it is declared in an `mli`.

PR #10066, opened 2024-02-17
refactor(pkg): fetch with --no-write-fetch-head

rgrinberg:
This file is going to be over-written by other fetches anyway


Leonidas-from-XIV (approved):
Indeed, no point in writing it.

Issue #10076, opened 2024-02-20
Non-deterministic crash invoking mkdir during package build


gridbugs:
During some package builds I see a non-deterministic code error inside `Fpath.mkdir_p` when creating a directory inside the sandbox for building a package.

E.g.
```
Internal error, please report upstream including the contents of _build/log.
Description:
  ("failed to create parent directory",
  { t_s =
      "_build/.sandbox/8cc1e2f4b23197a3c96d399a8e8423d5/_private/default/.pkg"
  })
```

I tried modifying dune to print the path of the missing parent directory and can manually confirm that the directory is present.

Here's a repo containing a small project for reproducing the problem: https://github.com/gridbugs/dune-mkdir-crash-repro

gridbugs:
I'm unable to reproduce this issue when running `dune build` with `-j1`.

gridbugs:
Similarly, if I remove sources of parallelism by replacing calls to `Memo.parallel_map` with `Memo.sequential_map` in pkg_rules.ml, or setting `DUNE_CONFIG__BACKGROUND_SANDBOXES=disabled` the problem appears to go away.

gridbugs:
Also, if I add a call to `Unix.stat` on the directory immediately before printing the error, that branch stops being taken. This could just be changing the interleaving of threads such that this issue never comes up. Makes it hard to debug though.

```patch
diff --git a/otherlibs/stdune/src/fpath.ml b/otherlibs/stdune/src/fpath.ml
index eab87535c..e6d01a972 100644
--- a/otherlibs/stdune/src/fpath.ml
+++ b/otherlibs/stdune/src/fpath.ml
@@ -40,6 +40,7 @@ let rec mkdir_p ?perms t_s =
          | Created -> Created
          | Already_exists -> Already_exists
          | Missing_parent_directory ->
+           let _stats = Unix.stat t_s in
            (* But we just successfully created the parent directory. So it was
               likely deleted right now. Let's give up *)
            Code_error.raise "failed to create parent directory" [ "t_s", Dyn.string t_s ]))
```

gridbugs:
Also I'm unable to reproduce the problem when running dune with `strace`.

rgrinberg:
Are you able to reproduce this in the main branch?

gridbugs:
Nope, this appears to now be fixed.

PR #10077, opened 2024-02-20
Add info to code error in `Fpath.mkdir_p`

gridbugs:
Intended to help debug https://github.com/ocaml/dune/issues/10076

### DIFF
--- a/otherlibs/stdune/src/fpath.ml
+++ b/otherlibs/stdune/src/fpath.ml
@@ -19,30 +19,44 @@ type mkdir_p_result =
   | Already_exists
   | Created
 
-let rec mkdir_p ?perms t_s =
-  match mkdir ?perms t_s with
-  | Created -> Created
-  | Already_exists -> Already_exists
-  | Missing_parent_directory ->
-    if is_root t_s
-    then
-      Code_error.raise
-        "Impossible happened: [Fpath.mkdir] refused to create a directory at the root, \
-         allegedly because its parent was missing"
-        []
-    else (
-      let parent = Filename.dirname t_s in
-      match mkdir_p ?perms parent with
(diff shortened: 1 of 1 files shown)


otherlibs/stdune/src/fpath.ml
```diff
@@ -19,30 +19,44 @@ type mkdir_p_result =
...
+                likely deleted right now. Let's give up *)
+             Code_error.raise
+               "failed to create parent directory"
+               [ "tried to create directory", Dyn.string t_s
+               ; "missing parent", Dyn.string parent
+               ; "parent mkdir_p_result", mkdir_p_result_to_dyn parent_mkdir_p_result
+               ; "while creating direcotry", Dyn.string full_t_s
+               ]))
+  in
+  loop full_t_s
```
emillon, line 59: it looks like in addition to the extra logging, you also refactored the function to use `loop`. can you remove this part to make the diff lighter if possible?

otherlibs/stdune/src/fpath.ml
```diff
@@ -19,30 +19,44 @@ type mkdir_p_result =
...
+             to create the parent directory concurrently. *)
+          (match mkdir t_s ?perms with
+           | Created -> Created
+           | Already_exists -> Already_exists
+           | Missing_parent_directory ->
+             (* But we just successfully created the parent directory. So it was
+                likely deleted right now. Let's give up *)
+             Code_error.raise
+               "failed to create parent directory"
+               [ "tried to create directory", Dyn.string t_s
```
emillon, line 53: I'm not sure that these strings are intended to be descriptive; usually they're just the variable names, like `["t_s", Dyn.string t_s;...`

PR #10078, opened 2024-02-20
fix(nix): override mdx's propagatedBuildInputs to avoid logs conflicts

anmonteiro:


flake.nix
```diff
@@ -221,10 +227,6 @@
                 js_of_ocaml
                 utop
                 core_bench
-                mdx
-                odoc
-                ppx_expect
-                ctypes
```
anmonteiro, line 227: callout: these are already buildInputs to `makeDuneDevShell`

rgrinberg (approved):

PR #10085, opened 2024-02-21
fix(pkg): stop using opam's untar code

rgrinberg:
It does a bunch of illegal things like chdir that are illegal in dune


Leonidas-from-XIV (approved):
Makes sense. I wondered whether `Fiber_job` is still used, but apparently `fetch_others` still calls OPAM which returns a `Job`.

rgrinberg:
Right, that’s a good point. Maybe we should just get rid of fetch_others then. I doubt it’s going to work correctly anyway, given all the problems we found.

PR #10086, opened 2024-02-21
test(pkg): Test resolution of ambiguous hashes

Leonidas-from-XIV:
I was curious what happens and it looks like we're not picking up the branch as a branch but also not failing given the ref doesn't exist as a revision either.

Thought it was worth a repro case.


rgrinberg (approved):
Indeed it's worth a repro. Probably worth a bug ticket as well to document this limitation. I doubt users will ever run into this, but it's useful to keep this in mind when we re-design url's.

PR #10095, opened 2024-02-21
test(pkg): [exported_env] and [PATH]

rgrinberg:
Demonstrate that appending to [PATH] using [exported_env] doesn't allow
us to access additional binaries.


rgrinberg:
cc @gridbugs this is an issue you've discovered as well. That we need to translate the environment when moving actions to the sandbox.

gridbugs (approved):

PR #10103, opened 2024-02-22
fix(pkg): correctly set [OCAMLPATH] for context

rgrinberg:
Previously, we'd set the [OCAMLPATH] using the full environment, but
that would convert all the findlib paths to [Path.External.t] which
isn't correct, as those paths exist in the build dir.


rgrinberg:
The failure in CI reveals another bug which I will address separately

gridbugs (approved):

PR #10113, opened 2024-02-22
fix: handle utf8 characters in the dune files(#9728)

moyodiallo:
This PR is about principally handling utf8 characters in quoted strings.


rgrinberg:
Why do we need to copy paste code from the stdlib?

moyodiallo:
> Why do we need to copy paste code from the stdlib?

No need, we could write our own functions. But I did this because I used functions introduced in OCaml.4.14 (later version).

rgrinberg:
The issue isn’t the copy pasting of some code itself, it’s just that I think we’re copying too much. Don’t we already have a subset of Uchar present since our minimum version is 4.08? Could we just add only the code we need?

moyodiallo:
I've just tested `uutf`, I'm going to use `vendor/uutf` present in Dune project. That is sufficient for this PR.

moyodiallo:
Indeed `uutf` use Uchar which is present in OCaml 4.08.

moyodiallo:
The copy paste is removed and `uutf` is used instead.

rgrinberg:
It's a bit sad, but this PR has a non trivial increase in pressure on the GC. Since this fix only affects a tiny fraction of users, is it possible to implement this in a less intrusive way?

moyodiallo:
Yes it is possible. I'll try to fix it.

emillon:
I don't think that the lexer changes are useful for this change. Just changing `Escape` will fix the formatter, which was the original issue. The parsing error for `%{...}` will continue refering to individual bytes, but I think that's fine compared to changing the lexer.

src/dune_sexp/escape.ml
```diff
@@ -1,26 +1,47 @@
 open! Stdune
 
+let utf_8_byte_length u =
```
emillon, line 3: Is that `Uchar.utf_8_byte_length`?

moyodiallo: Yes

moyodiallo:
> I don't think that the lexer changes are useful for this change. Just changing `Escape` will fix the formatter, which was the original issue. The parsing error for `%{...}` will continue refering to individual bytes, but I think that's fine compared to changing the lexer.

I get it but it won't be coherent, accepting utf8 char and printing escaped characters.

src/dune_sexp/escape.mli
```diff
@@ -1,2 +1,7 @@
+module Utf8 : sig
```
emillon, line 1: If these functions are from the standard library, can you move them to `Stdune.Uchar`? The goal is that we can effectively remove that once we bump our compatibility to ocaml, so if the rest of the code can look like it's written against the stdlib it's better.

src/dune_sexp/lexer.mll
```diff
@@ -352,13 +355,30 @@ and template_variable = parse
...
-  | (varname_char* as skip) (_ as other)
-  | (varname_char+ ':' ((':' | varname_char)*) as skip) (_ as other)
+  | (varname_char* as skip) (non_ascii* as maybe_utf) (_ as other)
+  | (varname_char+ ':' ((':' | varname_char)*) as skip) (non_ascii* as maybe_utf) (_ as other)
   {
-    error
-      ~delta:(String.length skip)
-      lexbuf
-      (Printf.sprintf "The character %C is not allowed inside %%{...} forms" other)
+    let utf_len = String.length maybe_utf in
```
emillon, line 361: This looks better but I'd still feel safer about making the lexer changes in a separate PR.
(concretely, this would have been merged already without the lexer changes getting in the way)

test/blackbox-tests/test-cases/formatting/non-ascii-characters.t
```diff
@@ -1,13 +1,30 @@
...
+  $ dune format-dune-file <<EOF
+  > (echo "É")
+  > EOF
+  (echo "É")
+
+  $ dune format-dune-file <<EOF
+  > (Écho "hello")
+  > EOF
+  File "", line 1, characters 1-1:
+  Error: Invalid . file
```
emillon, line 29: where does the ` . ` part come from?

emillon (commented):
this will also require a changelog entry

src/dune_sexp/escape.mli
moyodiallo: Those functions are not coming from the standard library. The functions are inspired from `uutf` with care and adapted with our use case.

src/dune_sexp/lexer.mll
moyodiallo: Ok I will do that.

test/blackbox-tests/test-cases/formatting/non-ascii-characters.t
moyodiallo: This is coming from the lexer https://github.com/ocaml/dune/blob/1f80d72a2afdf49ff9bd32e340727d1e0babcdb8/src/dune_sexp/lexer.mll#L30
This is because there's no file created with current test. 
Changing it by giving a file, it gives the name of file instead of `.`:
```
$ cat > dune <<EOF
> (Écho "hello")
> EOF
> dune format-dune-file dune
File "$TESTCASE_ROOT/dune", line 1, characters 1-1:
(Écho "hello")

Error: Invalid dune file
[1]
```

moyodiallo:
> this will also require a changelog entry

The changelog entry added.

test/blackbox-tests/test-cases/formatting/non-ascii-characters.t
emillon: ok, thanks!

src/dune_sexp/escape.ml
```diff
@@ -1,26 +1,92 @@
...
+  let unsafe_get s i = String.unsafe_get s i |> Char.code
+  let next_utf8_length s i = utf8_byte_length (String.unsafe_get s i)
+
+  let is_utf8_valid s i l =
+    assert (String.length s >= l);
+    match l with
+    | 1 -> true
+    | 2 ->
+      let b1 = unsafe_get s (i + 1) in
+      if b1 lsr 6 != 0b10 then false else true
```
rgrinberg, line 31: Could you replace `if x then false else true` with `not x`. Here and everywhere else

rgrinberg (changes requested):

emillon:
Thanks. I'll adapt the API so that it's more in line with the stdlib API and thus easier to remove later.

src/dune_sexp/escape.ml
moyodiallo: Done in https://github.com/ocaml/dune/commit/b52cef49ae33015dc202a5f4adc671063bd15234

emillon:
I tried to use the stdlib way, but doing so and adding polyfills for 4.08 meant that stdune would depend on `uutf`. So instead I adapted the code to use `uutf` decoder (that work on 4.08) in `dune_sexp` with a note on how to migrate to stdlib when we bump the compat.

emillon:
@rgrinberg is the library structure (`dune_sexp` depending on `dune_uutf`) sensible?

rgrinberg (approved):
Seems fine to me. Make sure to double check the benchmarks before merging.

emillon:
Benchmarks look good so I'm merging this, thanks.

PR #10115, opened 2024-02-22
fix(coq): Fix performance regression in coqdep unescaping

### LINKED ISSUE #10088: [performance] Zero-build regression 3.11 - 3.14
Hi folks,

I'm forwarding this [perf regression](https://github.com/HoTT/Coq-HoTT/issues/1866#issuecomment-1955048836) from Coq users, as it seems interesting.

The relevant bit seems this zero-build data:
```
dune 3.14.0:  5.9s
dune 3.11.1:  3.3s
make 4.3:     0.6s 
```
That's almost 2 seconds so worth having a look IMHO.

ejgallego:
Regression from #9231

Fixes #10088

### DIFF
--- a/doc/changes/10115.md
+++ b/doc/changes/10115.md
@@ -1,0 +1,2 @@
+- coq: fix performance regression in coqdep unescaping (#10115, fixes
+  #10088, @ejgallego, thanks to Dan Christensen for the report)
--- a/src/dune_rules/coq/coq_rules.ml
+++ b/src/dune_rules/coq/coq_rules.ml
@@ -525,7 +525,8 @@ let coqdep_invalid phase line =
    Note that coqdep escapes a few more things, including spaces, $, #,
    [], ?, %, homedir... How to handle that seems tricky.
 *)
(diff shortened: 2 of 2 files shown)


ejgallego:
With this PR, zero build for Coq-HoTT:
```
main:
real	0m5,952s
user	0m5,857s
sys	0m0,083s

3.11.1:
real	0m5,543s
user	0m5,420s
sys	0m0,132s
egallego@mobiverse:~/external/coq/HoTT$ time ~/external/dune/_build/install/default/bin/dune build
```
so it seems like reasonable as to consider the original regression closed.

I was curious and I did time the `.v.d` file parsing as to understand what was the total for the zero build.

And indeed, it seems we are not memoizing the parsing of the .v.d. file, that's gonna be terrible! (cc: @Alizter )

I seem to recall we had this properly covered, I wonder what happened? But def we should store the `Dep_map` into a memo node.

ejgallego:
This PR is pretty trivial so I will merge soon, anyways the regression fixed is not observable after #10116 is merged, which solves the real culprit.

PR #10122, opened 2024-02-23
fix(pkg): correctly extract tarballs

rgrinberg:
Previously, we'd extract sources as [source/$basename-of-tar]. We should
instead extract the sources into [source/] directly.


rgrinberg:
@gridbugs this demonstrates the fix in the e2e test

gridbugs (approved):

test/expect-tests/dune_pkg/fetch_tests.ml
```diff
@@ -202,8 +202,7 @@ let%expect_test "downloading, tarball with no checksum match" =
     Done downloading
     Finished successfully, no checksum verification
     ------
-    files in target dir:
-    plaintext.md |}]
+    files in target dir: |}]
```
Leonidas-from-XIV, line 205: Is this not missing the file that it should unpack?

rgrinberg: Does opam support them? All the tarballs I've checked do have a root directory. If opam does indeed support them, we can add support for them as well.

Leonidas-from-XIV: I haven't seen them either, because I assume `dune-release` and `opam-publish` create them in a sensible way, but I was curious and yes `opam` supports tarballs without a root: 

If I create a tarball that contains `install.ml` with

```ocaml
let () =
  print_endline "Installation failed";
  exit 10
```

And a repo with `rootless.1`  opam file
```
opam-version: "2.0"
build: [
  ["ocaml" "install.ml"]
]
url {
  src: "file://<path-to>/rootless.tar.gz"
}
```
Then upon `opam install rootless` I get

```
#=== ERROR while compiling rootless.1 =========================================#
# context     2.1.5 | linux/x86_64 | ocaml-base-compiler.4.14.1 | file://<path-to>/rootless-repo
# path        ~/<path-to-switch>/.opam-switch/build/rootless.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build ocaml install.ml
# exit-code   10
# env-file    ~/.opam/log/rootless-602822-ae8c0f.env
# output-file ~/.opam/log/rootless-602822-ae8c0f.out
### output ###
# Installation failed
```

So opam successfully unpacked and failed with the right error message and code as my script.

rgrinberg:
@Leonidas-from-XIV should be ready now

gridbugs:
I just tried out this change locally and it seems to introduce errors during `dune build`:
```
Error:
rename(/run/user/1000/dir_d4279f_download/ppx_derivers-1.2.1): Invalid cross-device link
-> required by
   _build/_fetch/checksum/md5=5dc2bf130c1db3c731fe0fffc5648b41/dir
-> required by _build/_private/default/.pkg/ppx_derivers/source
-> required by _build/_private/default/.pkg/ppx_derivers/target/cookie
-> required by - package ppxlib
-> required by - package ppx_cold
-> required by - package ppx_base
-> required by - package base_quickcheck
-> required by - package ppx_jane
-> required by - package abstract_algebra
```
for a bunch of packages. This was from building an empty package dependent on `incr_map` on nixos.

rgrinberg:
Sigh. The temp directory has to exist in the build directory I suppose.

test/expect-tests/dune_pkg/dune
```diff
@@ -29,7 +29,7 @@
  (target tarball.tar.gz)
  (deps plaintext.md)
  (action
-  (run tar -czf %{target} %{deps})))
+  (run tar -czf %{target} -C .. dune_pkg/%{deps})))
```
Leonidas-from-XIV, line 32: I would've probably moved `plaintext.md` into a subfolder to avoid `tar` escaping out of the folder, but this also works.

Leonidas-from-XIV (approved):

test/expect-tests/dune_pkg/dune
rgrinberg: Good point. I moved the inputs to a subdir and also added another file into the archive just for thoroughness.

PR #10130, opened 2024-02-23
test(pkg): Extend the test to use recursive submodules

### LINKED ISSUE #9377: Submodules Support
Make sure that we support submodules in the following two contexts:

1. Pulling package sources from git
2. Using opam repositories from git that contain submodules

cc @Leonidas-from-XIV

Leonidas-from-XIV:
An end-to-end test to make sure nested submodules work as well. In this case it is `mock-repo` which references a submodule `packages`  which contains a direct package `foo` and another submodule `bar` which contains the `bar` package. The solver tries to find a solution for bar and succeeds.

Fixes #9377


rgrinberg:
Could you move this to a separate test case? I know it's a little more work, but it's getting hard to read everything inside `opam-repository-download.t`

rgrinberg (approved):

PR #10131, opened 2024-02-23
fix(pkg): correctly rebuild packages when default branch changes

rgrinberg:

### DIFF
--- a/src/dune_rules/fetch_rules.ml
+++ b/src/dune_rules/fetch_rules.ml
@@ -5,7 +5,9 @@ module Gen_rules = Build_config.Gen_rules
 include struct
   open Dune_pkg
   module Checksum = Checksum
+  module Rev_store = Rev_store
   module Pkg = Lock_dir.Pkg
+  module OpamUrl = OpamUrl
 end
 
 let context_name = Context_name.of_string "_fetch"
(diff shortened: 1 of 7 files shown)


src/dune_rules/fetch_rules.ml
```diff
@@ -31,6 +33,36 @@ type kind =
   | `Directory
   ]
 
+let resolve_url =
+  (* Before we fetch any git repo, we make sure to convert the URL to fetch the
+     git object directly. The object is used to compte the fetch action digest
```
Leonidas-from-XIV, line 38: compute?

src/dune_rules/fetch_rules.ml
```diff
@@ -169,19 +204,21 @@ let gen_rules_for_checksum_or_url url checksum =
...
         Rules.Produce.rule (Rule.make ~info ~targets build)
     in
     let make_target = make_target checksum_or_url in
     let action ~target ~kind =
-      action ~url ~checksum ~target ~kind
+      action ~url:(loc_url, url) ~checksum ~target ~kind
       |> Action.Full.make
-      |> Action_builder.With_targets.return
+      |> Action_builder.return
+      |> Action_builder.with_no_targets
```
Leonidas-from-XIV, line 221: I'm trying to understand, why is this without targets?

rgrinberg: They the targets are added in the functions below. It's done separately because there's either a file or a directory target.

Leonidas-from-XIV (approved):

PR #10142, opened 2024-02-26
doc: add a cookbook

emillon:
This creates a "cookbook" section with some self-contained examples.

Theses examples are extracted from `quick-start.rst`.

One nice aspect is that these samples are linkable, and that the previous/next links work.

I tried to convert some of the examples to a style that encourages tinkering ("Run it!" / "Change it!" in `hello-world` for example), but most examples are just moved as is.

Questions / follow-up work:
- in terms of diataxis, this is tutorial material. should this leave under `/tutorials/cookbook`?
- the "building a hello world program from scratch" is the closest we have to a tutorial. it could be recycled into a proper one.
- also, this tutorial is the only place we explain that the various files need to be created with a text editor and put in the same directory. this could be added to `cookbook/index`.
- some recipes are not self-contained (ex: `define-library`), are not formatted (ex: `generator`), do not follow best practices (ex: `c-stubs`, useless `(name)`, no `:standard`), or might be useless (`toplevel`). This can be addressed separately.
- what about our `examples` directory? should it be moved to the cookbook?


emillon:
Curious to hear your thoughts about this @ElectreAAS

ElectreAAS:
> in terms of diataxis, this is tutorial material. should this leave under /tutorials/cookbook?

Maybe not? At least not in their current form. Right now I'd say it's more a 'gallery' of [Thing you might want to do -> How to do it], but without any explanations. They could however be recycled into proper tutorials, but that'd be more involved work.

In general I agree with the idea of this PR, these examples are small and sometimes useful pieces, a cookbook sounds like a good place for them.

christinerose (approved):

rgrinberg:
Please re-open if you ever resume work

PR #10152, opened 2024-02-27
doc: remove dune:action directive

emillon:
The goal of this PR is to remove the custom python code to handle `dune:`
directives.

This plugin (in `doc/exts/sphinx_dunedomain.py`) defines directives that do several things:

- they generate nice markup that describe what we're documenting (the headers with blue or gray backgroud)
- they support cross-references
- they generate an index

But, this is too complicated for our use case:

- cross-references are unnecessary now that we have one document per action
- the index was not linked, but we now have a nicer one in `reference/actions/index`
- crucially, there's a `describe` directive in Sphinx that generates the same markup as the one we were generating.

So this PR does 2 things:

- switch from the high-level `dune:action::` to a low-level `describe::` (which renders identically)
- remove the now unused support code in `sphinx_dunedomain.py`


emillon:
(as followup PRs I have similar ones that do the same for `dune:field` and `dune:stanza`, thus removing `sphinx_dunedomain` completely)

ElectreAAS:
While this looks obviously correct, and I'm generally in favor of removing python code, do you know more about the *why* was this code added there in the first place? And what do you mean by unneeded cross-references here?

ElectreAAS (approved):

emillon:
I wrote the python code in question. At the time I thought that this was the only way of getting this kind of markup (I didn't know about `describe::`). Indices and cross-references are nice sphinx features, but they're not really required: instead of using a xref to link to the `cat` action, we can just link to the `cat` document (now that it exists).

emillon:
Thanks!

PR #10154, opened 2024-02-27
doc: add an howto about rule generation

### LINKED ISSUE #10057: Better documentation of `dynamic_include`
The documentation for `dynamic_include` is too brief and should include a meaningful example (also the rule stanza is incorrect): https://dune.readthedocs.io/en/stable/dune-files.html#dynamic-include

I have a typical usecase, which is to generate a list of rules to build doc files based on the original source files and I cannot find a pattern to make it work.

Here's what I have so far, based on the current doc:

top-level `dune`:
```dune
(subdir
 doc
 (dynamic_include ../gendune/doc/dune.inc))

(subdir
 gendune/doc
 (rule
  (target dune.inc)
  (deps
    (source_tree ../../doc))
  (action
   (with-stdout-to
    %{target}
    (run ./gen_dune.exe ../../doc)))))
```

And for `gendune/doc/dune`:
```dune
(executable
 (name gen_dune)
 (modules gen_dune))
```

But this fails with some circular dependency issue:
```
   dynamic_include gendune/doc/dune.inc in directroy doc
-> _build/default/gendune/doc/dune.inc
-> dynamic_include gendune/doc/dune.inc in directroy doc
-> required by alias default in dune:19
```

It's not obvious to me how to make this work since the `gen_dune.exe` executable needs all the files to be present. If I remove the `deps` above, the files are not present when `gen_dune.exe` is executed.

As an aside, not being able to co-locate these rules is rather unfortunate.

emillon:
Fixes #10057

This covers both the legacy "generate-include-commit" pattern as well as the new one based on `(dynamic_include)`.

### DIFF
--- a/doc/howto/rule-generation.rst
+++ b/doc/howto/rule-generation.rst
@@ -1,0 +1,266 @@
+Using Rule Generation
+=====================
+
+Sometimes it can be useful to generate Dune rules that depend on the file system
+layout or on the content of configuration files. This often happens for
+integration tests.
+
+In this document, we will see two ways to encode this behavior.
+
+We suppose that we are testing an executable named ``tool``. There are some
+input files named ``*.input``, output files named ``*.output``, and we want to
+ensure that when running ``tool`` on ``x.input``, the standard output
+corresponds to ``x.output``.
+
+The Generate-Include-Commit Pattern
+-----------------------------------
+
+.. note::
+
+   This is the most common way to do this. It has a couple drawbacks listed
+   below, but you should start with this pattern.
+
+What we are going to do is:
+
+- generate a ``dune.inc`` file;
+- include it in our main ``dune`` file;
+- commit the generated code in the source repository.
+
+This creates a loop: a program (the "generator") looks at the file system and
+creates a ``dune.inc`` file. Changes to the file system (for example, if a test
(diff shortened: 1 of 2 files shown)


doc/howto/rule-generation.rst
```diff
@@ -0,0 +1,266 @@
...
+    (executable
+     (name gen))
+
+This defines an executable that lists ``*.input`` files in the current
+directory and outputs rules on its standard output.
+
+.. note::
+
+   It is important to sort the input files to ensure that the output is
+   independent on the order the files are returned by ``Sys.readdir``.
```
christinerose, line 81:
```suggestion
   It is important to sort the input files to ensure that the output is
   independent from the order ``Sys.readdir`` returns the files .
```

ElectreAAS (approved):
Love it!

doc/howto/rule-generation.rst
emillon: Don't we need to say
> from the order *in which* `Sys.readdir` returns the files?

(the context is that `Sys.readdir` returns a list of files, but in a seemingly random order)

christinerose: Good call! I updated the suggestion.

PR #10155, opened 2024-02-27
refactor(coq): add cutoff to dep map computation

rgrinberg:
Remove the unnecessary [Dep_map] module. It's just a copy of
[Path.Map]

Then specialize the map to be [Path.Build.Map]. It improves the
performance of comparison which speeds up memoization.


rgrinberg:
Yeah, this doesn't work.

PR #10157, opened 2024-02-28
Add quotes around field name in an error message

gridbugs:
There are several places where a dune config will use a field named "name", and if such a field is omitted, prior to this change dune would print the error 'field name missing' which is ambiguous. This change puts quotes around the field name (ie. 'field "name" missing') so it's clear that a field named "name" is missing, and not the name of a field.


emillon (commented):
Sounds good. It looks like this is related to the TODO comment above, so can you make sure that it's addressed and remove it?

gridbugs:
I've added quotes in a few more places. I think the errors in this file are consistent now so I've removed the TODO.

Leonidas-from-XIV:
I wonder, wouldn't it be better to introduce a formatting directive for fields like we do with commands? That way we can change the rendering later in one go and also the source code will note in code that we consider `license` in this context a field name.

emillon:
@Leonidas-from-XIV 's suggestion is a good one I think but the current situation of the PR already improves the situation so I'm rebasing and merging this PR. @Leonidas-from-XIV, feel free to open an issue to track this if you want to work on the improvement. Thanks!

PR #10164, opened 2024-02-28
test: sort readdir output for reproducibility

emillon:


Leonidas-from-XIV (approved):
Good catch; I'm surprised it managed through CI on macOS and Linux without breaking.

emillon:
until recently it was a single file I think

PR #10165, opened 2024-02-28
fix: use the correct expander for cram stanzas

rgrinberg:
We should use the expander where the stanza is defined, rather than the
expander of the test.

### DIFF
--- a/src/dune_rules/cram/cram_rules.ml
+++ b/src/dune_rules/cram/cram_rules.ml
@@ -60,6 +60,8 @@ let test_rule
     Memo.parallel_iter aliases ~f:(fun alias ->
       Alias_rules.add sctx ~alias ~loc (missing_run_t test))
   | Ok test ->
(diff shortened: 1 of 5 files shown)


emillon:
Concretely, how does this problem manifest itself? relative locks are interpreted in the wrong directory?

rgrinberg:
I realize my explanation above is a little opaque. The programming discipline that I'm trying to convey is pretty important for keeping the dune language consistent, so it deserves a better explanation. It can be stated as:

Suppose we have a percent form `%{foo}` in `foo/dune`. We should always expand this percent form using the expander obtained from `Super_context.expander ~dir:foo`. Using an expander from a different directory is incorrect because the evaluation of the percent form depends on the directory in which it is defined. That's all there is it to it really.

rgrinberg:
Something is borked with this PR so I replaced it with https://github.com/ocaml/dune/pull/10185

PR #10169, opened 2024-02-29
fix: allow looking up artifacts everywhere

rgrinberg:
Before this PR, we had a function to fetch the expansion for a directory:

```
Super_context.expander : dir:Path.Build.t -> Expander.t
```

The expander returned was incomplete however. In particular, it wasn't able to expand forms such as `%{cma:..}`. The end result is that whether forms such as `%{cma:..}` were expandable in a particular field in a dune file was basically random.

After this PR, we the expander returned by the function above is always able to expand these forms. So this makes the dune language both more powerful and more consistent.


nojb:
Can you say a few words about what the PR is doing? Thanks!

rgrinberg:
Updated the description

src/dune_rules/expander.ml
```diff
@@ -156,45 +156,40 @@ let expand_version { scope; _ } ~(source : Dune_lang.Template.Pform.t) s =
...
-    let name = Path.Build.basename path in
-    let dir = Path.Build.parent_exn path in
-    let does_not_exist ~loc ~what name =
-      User_error.raise ~loc [ Pp.textf "%s %s does not exist." what name ]
+  let path = Path.Build.relative t.dir s in
+  let loc = Dune_lang.Template.Pform.loc source in
+  let name = Path.Build.basename path in
+  let dir = Path.Build.parent_exn path in
+  if Path.Build.is_root dir
+  then User_error.raise ~loc [ Pp.text "cannot escape the workspace root directory" ];
```
nojb, line 164: Is this check as general as it could be? What happens if we write `%{cma:../../../foo}` in the test (so that we go way outside the root directory)?

nojb (approved):
Looks like a nice cleanup. LGTM, modulo a question (below).

src/dune_rules/expander.ml
rgrinberg: I added a test - It doesn't go through.

By the way, there's still a way to do something illegal here. One can write:

```
%{cma:../another-context/foo}
```

And use the relative path to escape to another context. We should definitely forbid that as well. It's a much larger change though.

PR #10170, opened 2024-02-29
test: add tests for library name collisions

jchavarri:
In #9839, I ended up running into several scenarios that can occur when two libraries have the same name, and the existing test infrastructure was not enough for my needs (there are only some coq tests that cover this functionality). I am extracting all the tests into a separate PR so it gets easier to see how the behavior changes in that PR, vs the current behavior in `main`.


rgrinberg:
Thanks. Could you move the lib-collision-* tests to a single folder? That would make it easier to run all of them.

anmonteiro (approved):

PR #10171, opened 2024-02-29
Skip hidden libs in `get_compile_info` and `ml_sources`

jchavarri:
Extracted from #9839.

The fact that Dune tries to get compile info for hidden libraries is problematic for the new use cases that that PR enables.

From what this code says:

https://github.com/ocaml/dune/blob/98c576e643450659e2893ced9ae8e076b6129d95/src/dune_rules/dune_package.mli#L47-L53

I infer that Dune should not consider hidden libs when building rules or gathering the modules from a given library, because it never has to build them?

I looked for the original commit where `find_even_when_hidden` was introduced, but it doesn't provide a lot of information, besides "This makes everything else simpler": https://github.com/ocaml/dune/commit/17f4567014d533ae0135e940e3613bfa247e2f67

### DIFF
--- a/src/dune_rules/lib.ml
+++ b/src/dune_rules/lib.ml
@@ -1898,7 +1898,7 @@ module DB = struct
 
   let get_compile_info t ~allow_overlaps name =
     let open Memo.O in
-    find_even_when_hidden t name
+    find t name
     >>| function
     | Some lib -> lib, Compile.for_lib ~allow_overlaps t lib
     | None ->
--- a/src/dune_rules/ml_sources.ml
+++ b/src/dune_rules/ml_sources.ml
@@ -304,7 +304,7 @@ let make_lib_modules
       let* resolved =
         let* libs = libs in
         Library.best_name lib
-        |> Lib.DB.find_even_when_hidden libs
+        |> Lib.DB.find libs
         (* can't happen because this library is defined using the current
            stanza *)
         >>| Option.value_exn


rgrinberg:
The case where `find_even_hidden` is supposed to help is thus:

1. Library `x` is hidden because of some unsatisfied `enabled_if`
2. The user attempts to build `x.cma`
3. Dune should tell the user that `x.cma` cannot be built b/c it's hidden.

The above works similarly if we `x` is hidden b/c it depends on some dependency that is missing because of an unsatisfied `exists_if`.

So as long as we are still giving good error messages in the cases above, we don't need to use `find_even_when_hidden`.

rgrinberg:
A similar property needs to hold for `%{cma:lib}`, which explains why `Ml_sources` uses `find_even_when_hidden`

jchavarri:
Thanks for the extra context. Should the cases mentioned be added as tests?

I didn't know there were different cases for "disabled" libs, I am not sure if both would be considered "hidden":

- libraries relying on `META` files using `exists_if` (side question: does Dune build libs with `META` files?)
- libraries relying on the `enabled_if` field of the `library` stanza

I have restarted the work in #9839 from scratch in https://github.com/ocaml/dune/pull/10179, this time trying to avoid replacing the usages of `find_even_when_hidden` with `find.

rgrinberg:
> Thanks for the extra context. Should the cases mentioned be added as tests?

I imagine. I'd be very surprised if they aren't already. You could try breaking the behavior, and see if a test fails. If not, adding more tests for this would be very welcome.

> I didn't know there were different cases for "disabled" libs, I am not sure if both would be considered "hidden":

What would be the distinction from dune's point of view? To dune, hidden and disabled are treated in the same way.

jchavarri:
> I imagine. I'd be very surprised if they aren't already. You could try breaking the behavior, and see if a test fails.

This is what this PR is doing right? But the tests in CI passed.

> What would be the distinction from dune's point of view? To dune, hidden and disabled are treated in the same way.

Maybe no difference. I assumed Dune wouldn't build libraries that use `META` files? Or in other words, would Dune call `get_compile_info` for libs using `META` files?

rgrinberg:
> This is what this PR is doing right? But the tests in CI passed.

Yup, and that is disturbing :)

> Maybe no difference. I assumed Dune wouldn't build libraries that use META files? Or in other words, would Dune call get_compile_info for libs using META files?

Dune uses META files for libraries that don't have dune metadata. This is important for interoperability.

PR #10181, opened 2024-03-01
refactor: move coq scopes to own module

rgrinberg:
cc @ejgallego and @Alizter . I need to refactor OCaml scope handling in dune and it would be helpful to have the coq stuff live on its own. No semantic changes here, but I did remove some intermediate lazy cells whose purpose I did not see.

### DIFF
--- a/src/dune_rules/scope.ml
+++ b/src/dune_rules/scope.ml
@@ -4,14 +4,14 @@ open Memo.O
 type t =
   { project : Dune_project.t
   ; db : Lib.DB.t
-  ; coq_db : Coq_lib.DB.t Memo.Lazy.t
+  ; coq_db : Coq_lib.DB.t Memo.t
   ; root : Path.Build.t
   }
 
 let root t = t.root
 let project t = t.project
 let libs t = t.db
-let coq_libs t = Memo.Lazy.force t.coq_db
+let coq_libs t = t.coq_db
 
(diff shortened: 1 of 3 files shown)


src/dune_rules/scope.ml
```diff
@@ -4,14 +4,14 @@ open Memo.O
 type t =
   { project : Dune_project.t
   ; db : Lib.DB.t
-  ; coq_db : Coq_lib.DB.t Memo.Lazy.t
+  ; coq_db : Coq_lib.DB.t Memo.t
```
Alizter, line 7: Since this is no longer lazy, wouldn't this mean that we construct the coq_db even if it is not being used?

rgrinberg: It's still lazy. This `Memo.t` is just forcing the cell constructed in `Coq_scope.t`

rgrinberg: The work that is no longer saved is the map lookup in the scope database. However such a map lookup has less overhead than memo's machinery so there's no issue.

ejgallego:
Lazy handling was indeed pretty subtle on the original code, I'll review the changes ASAP

rgrinberg:
I'm going to merge this as this is getting in the way of making further improvements to Scope. Please review at your own convenience.

PR #10208, opened 2024-03-04
refactor: rewrite trivial guards as patterns

emillon:
Trivial guards meaning "lhs = rhs" with either side a constant.


rgrinberg (approved):
LGTM. We need @snowleopard to confirm the engine side though

snowleopard (approved):
Looks good to me, thanks!

PR #10211, opened 2024-03-05
test: eif, add collision tests for exes and melange.emit

jchavarri:
Adds similar tests to those included in #10170, but applied to `executable` and `melange.emit` stanzas.

Suggested in https://github.com/ocaml/dune/pull/10179#issuecomment-1977431804.

I have some ideas to explore, but I'd rather have the tests merged first to reflect the behavior in `main`, so any further work is easier to share and review.


test/blackbox-tests/test-cases/enabled_if/eif-exe-name-collision-same-folder.t
```diff
@@ -0,0 +1,59 @@
...
+  5 |  (name foo)
+  6 |  (enabled_if (= %{context_name} "alt-context")))
+  Error: Executable "foo" appears for the second time in this directory
+  [1]
+
+For public exes
+
+  $ cat > dune << EOF
+  > (executable
+  >  (public_name foo)
```
rgrinberg, line 43: No need to test public exes separately. It's the exact same thing.

test/blackbox-tests/test-cases/enabled_if/eif-exe-name-collision.t
```diff
@@ -0,0 +1,57 @@
...
+  $ dune build
+
+For public exes
+
+  $ cat > a/dune << EOF
+  > (executable
+  >  (public_name foo)
+  >  (package bar)
+  >  (enabled_if (= %{context_name} "default")))
+  > EOF
```
rgrinberg, line 48: Same thing. Public executables do not change the test in any way.

test/blackbox-tests/test-cases/enabled_if/eif-exe-name-collision-same-folder.t
jchavarri: I copied over from library tests 😅  cleaned up in https://github.com/ocaml/dune/pull/10211/commits/0f9bbadda48585eb2e4c9aa9186ba154e0f82cde.

PR #10214, opened 2024-03-05
fix(pkg): Create temporary directories on same FS

### LINKED ISSUE #10213: bug(pkg): Fetching fails when temp dir is on different FS
## Expected Behavior

`Fetch.fetch_git` working.

## Actual Behavior

```
 let download_git rev_store url ~target =
@@ -231,5 +223,11 @@ let%expect_test "downloading via git" =
     let* (_commit : string) = Rev_store_tests.create_repo_at source in
     let+ () = download_git rev_store url ~target in
     print_endline (Io.read_file entry));
-  [%expect {| just some content |}]
+  [%expect.unreachable]
+[@@expect.uncaught_exn {|
+  Trailing output
+  ---------------
+  Error:
+  rename(/tmp/build_f3d1d7_dune/rev-store_b3e2d7_8b22a987d9344502845afcaedefbdd1e4a49d477/dir): Invalid cross-device link |}]
 ;;
```
Looking into the code `Path.rename` uses `Unix.rename` and this seemingly only works if the source and destination are on the same file system.

## Reproduction

1. have `/tmp` on a different filesystem
1. `dune runtest`

## Specifications

- Version of `dune` (output of `dune --version`): `main` 
- Operating system (distribution and version): Fedora Linux.

Leonidas-from-XIV:
The problem in #10213 is that `Temp.create` creates a temporary directory and then `Path.rename` is used to move the files to the actual target directory. This unfortunately only works if the directory that `Temp.create` picks and `~target` are on the same file system.

Multiple solutions exist:

  1. Attempt to write to the same filesystem, e.g. by hoping that the parent directory of `target` is also on the same file system as `~target`. This isn't necessarily true, but given we write to `_build` it should be true in most cases.
  2. Write to a temporary directory but then copy the files manually. This makes the whole operation non-atomic and incurs potentially a lot of file I/O

This PR attempts #1. An unfortunate consequence of this is that `Path.mkdir_p` is necessary to create the parent directory and if it didn't exist, in the case of failure it is left behind and not deleted. We could probably solve this by keeping track of directories that are created in such case, if it proves to be an issue. Input welcome!

Fixes #10213

### DIFF
--- a/src/dune_pkg/tar.ml
+++ b/src/dune_pkg/tar.ml
@@ -11,13 +11,6 @@ let bin =
 
 let output_limit = 1_000_000
 
-let temp_dir_in_build =
-  lazy
-    (let dir = Path.relative (Path.build Path.Build.root) ".temp" in
-     Path.mkdir_p dir;
-     dir)
-;;
-
 let extract ~archive ~target =
   let* () = Fiber.return () in
   let tar = Lazy.force bin in
@@ -26,8 +19,10 @@ let extract ~archive ~target =
     let suffix = Path.basename archive in
     match target with
     | In_build_dir _ ->
-      Temp.temp_in_dir Dir ~dir:(Lazy.force temp_dir_in_build) ~prefix ~suffix
-    | _ -> Temp.create Dir ~prefix ~suffix
+      Temp.temp_in_dir Dir ~dir:(Lazy.force Temp_dir.in_build) ~prefix ~suffix
+    | _ ->
+      let parent = Path.parent_exn target in
+      Temp.temp_in_dir Dir ~dir:parent ~prefix ~suffix
   in
   Fiber.finalize ~finally:(fun () ->
     Temp.destroy Dir target_in_temp;
--- a/src/dune_pkg/tar.ml
+++ b/src/dune_pkg/tar.ml
@@ -17,12 +17,7 @@ let extract ~archive ~target =
   let target_in_temp =
     let prefix = Path.basename target in
     let suffix = Path.basename archive in
-    match target with
-    | In_build_dir _ ->
-      Temp.temp_in_dir Dir ~dir:(Lazy.force Temp_dir.in_build) ~prefix ~suffix
-    | _ ->
-      let parent = Path.parent_exn target in
-      Temp.temp_in_dir Dir ~dir:parent ~prefix ~suffix
+    Temp_dir.dir_for_target ~target ~prefix ~suffix
   in
   Fiber.finalize ~finally:(fun () ->
     Temp.destroy Dir target_in_temp;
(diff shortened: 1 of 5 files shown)


rgrinberg:
Wouldn't it be possible to use the same trick that I did for `tar`? Create the temporary directory inside the build directory if the target is inside the build directory. I think that should be enough.

Leonidas-from-XIV:
Yes, that's a good solution, I'll rework it.

rgrinberg:
> I've changed it in the places where this was possible but for Fetch the target can be anywhere, so assuming _build and the target are on the same FS isn't really possible.

Yes, they're not always going to match, but we can just check if `target` is in the build dir. That should be good enough in almost all cases.

> I dislike duplicating temp_dir_in_build, maybe it should go in Temp and add some atexit hook to be deleted?

You can create a new module for it in dune_pkg. Note that there's no need to add a hook manually - there's one added automatically.

By the way, I think we are creating too many temporary directories for some of these jobs. For example, we should reuse a temp directory for downloading and untaring.

rgrinberg:
Is this ready for review now?

gridbugs:
@Leonidas-from-XIV Just making sure we don't forget about this change. This is necessary for package management to work on machines where `/tmp` is a separate fs. I need to manually apply this patch whenever I want to test dune package management on my machine.

Leonidas-from-XIV:
Yes, sorry, I've been busy with other stuff. I just looked through it and I think it can be reviewed. Just rebased it on `main`.

> By the way, I think we are creating too many temporary directories for some of these jobs. For example, we should reuse a temp directory for downloading and untaring.

Possibly, but creating temporary directories is cheap (1 inode) and fast and the PR as-is solves the failure I get on every `dune runtest`. Optimizing the amount of temporary directories could be done in a future PR, without breaking anything.

src/dune_pkg/tar.ml
```diff
@@ -11,23 +11,13 @@ let bin =
...
-     dir)
-;;
-
 let extract ~archive ~target =
   let* () = Fiber.return () in
   let tar = Lazy.force bin in
   let target_in_temp =
     let prefix = Path.basename target in
     let suffix = Path.basename archive in
-    match target with
```
rgrinberg, line 27: I think this now assumes that `target` always lives in `Path.Build.t`? We should do one of the following:

1. changing target to be `Path.Build.t`
2. assert that this is the case
3. revert back to the old code for handling non build directories.

#3 would be my preference as we need code like this in other parts of dune.

Leonidas-from-XIV: Good point, there was indeed an issue where it would write to the wrong place. I changed it to check for the build folder and only create it in the build folder in that case. However in the other case we cannot use `Temp.create` with `rename` as `Temp.create` will (on my system) always create in `/tmp` and it will always fail unless the target is in `/tmp` as well.

So I used the "subdir of parent dir" approach here as well.

rgrinberg:
LGTM. I just made some minor adjustment to reduce repetitive code.

PR #10215, opened 2024-03-05
doc: remove dune:field directive

emillon:
This is a followup to #10152, which does the same for `dune:field`. The next and final step is to do the same for `dune:stanza`.


ElectreAAS (approved):
:+1:

PR #10217, opened 2024-03-05
feat(boot): specify platform-specific cflags

emillon:
These flags have the same semantics as `link_flags` but are used when building C code.

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -1067,6 +1067,7 @@ let build
   ~ocaml_config
   ~dependencies
   ~c_files
+  ~build_flags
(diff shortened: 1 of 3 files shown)


rgrinberg:
Why do we need this feature?

emillon:
The context is #10212, we can take the discussion there.

rgrinberg:
Okay, so this PR isn't useful on its own and the feature cannot even be exercises in it. I'd say just submit it as a part of #10212 then

rgrinberg:
Closing as it's not a refactoring and as a feature, it's not usable  on its own.

PR #10219, opened 2024-03-05
doc: remove dune:stanza directive

emillon:
Followup to #10152 and #10215. This removes uses of `dune:stanza` and the need for the custom dune domain.


ElectreAAS (approved):
Always nice to have less python :)

christinerose (approved):

PR #10220, opened 2024-03-05
eif: fix name collision in same folder for exes and melange emits

jchavarri:
This applies (somehow) the idea mentioned in https://github.com/ocaml/dune/pull/10179#issuecomment-1977431804.

Part of the fix for https://github.com/ocaml/dune/issues/10222.

In order to keep the duplication check active for actual cases where 2 libs are enabled in the same folder, a new field is added to the `group_part` type to keep track of the `enable` state of either libs, exes or melange-emits.

For cases when a collision is detected when both stanzas are enabled, the previous behavior remains.

If one of the stanzas is disabled, then we filter out the disabled stanzas from the list before proceeding. ~I guess we could filter the stanzas upfront, but I believe keeping the disable ones as much as possible might enable use cases where the user is trying to target the artifact of a disabled stanza, so Dune can error out appropriately (see related https://github.com/ocaml/dune/pull/10026).~ Edit: in the end, the stanzas are filtered upfront as there were not many upsides to do it lazily at the moment.


src/dune_rules/modules.ml
```diff
@@ -1264,3 +1264,13 @@ let canonical_path t (group : Group.t) m =
   | Impl { impl = { modules = Wrapped w; _ }; _ } | Wrapped w -> w.group.name :: path
   | _ -> Module.path m
 ;;
+
+type enabled =
```
jchavarri, line 1268: The only reason why I added this type and fun here is because both `Ml_sources` and `Artifacts_obj` depend on it. `Artifacts_obj` didn't seem like the right place, and adding it to `Ml_sources` would create a cycle.

rgrinberg: We have a `Toggle.t` type that's the same as this. I'd say you can just use it

src/dune_rules/ml_sources.ml
```diff
@@ -53,41 +54,95 @@ module Modules = struct
     }
 
   let make { libraries = libs; executables = exes; melange_emits = emits } =
-    let libraries =
+    let libraries, libs =
```
jchavarri, line 57: The only reason why I have to carry over `libs` here (and `exes` and `emits` below) is because otherwise the code that checks for the "Module %S is used in several stanzas:" errors will still be triggered for cases where two libs have the same name but on of them is disabled.

src/dune_rules/artifacts_obj.mli
```diff
@@ -7,8 +7,8 @@ val empty : t
 val make
   :  dir:Path.Build.t
   -> lib_config:Lib_config.t Memo.t
-  -> libs:(Library.t * _ * Modules.t * Path.Build.t Obj_dir.t) list
-  -> exes:(_ * _ * Modules.t * Path.Build.t Obj_dir.t) list
+  -> libs:(Library.t * _ * Modules.t * Path.Build.t Obj_dir.t * Toggle.t) list
+  -> exes:(_ * _ * Modules.t * Path.Build.t Obj_dir.t * Toggle.t) list
```
rgrinberg, line 11: It seems like you pass the toggle to artifacts_obj but you don't do anything with it, any reason for that?

jchavarri: I used it originally, but I guess I can remove it now and just pass the right list of libs, exes and emits directly from the caller, if that's the better approach.

jchavarri:
@rgrinberg Sorry, this PR is not yet ready. The CI failures helped me understand there is something borked on the way `setup_emit_js_rules` is called (from `Gen_rules.gen_rules`), because the right context is not taken into account, and so there is a mismatch between that function and `Lib_rules.rules` (which sets the rules for melange `cmj`). I still don't fully understand the problem.

src/dune_rules/artifacts_obj.mli
rgrinberg: I'm not sure if it's the better approach, so I'm looking at the code right to find out. I'll list my thought process below.

First, I need to check where are the callers of `Artifacts_obj` that use a library name. Looks like there's only one caller of:

```
val lookup_library : t -> Lib_name.t -> Lib_info.local option
```

That lives in Expander:

```
    (match Artifacts_obj.lookup_library artifacts name with
     | None -> does_not_exist ~what:"Library" (Lib_name.to_string name)
     | Some lib ->
       Mode.Dict.get (Lib_info.archives lib) mode
       |> Action_builder.List.map ~f:(fun fn ->
         let fn = Path.build fn in
         let+ () = Action_builder.path fn in
         Value.Path fn))
```

So now I think about what's going to happen if we pretend that a disabled library does not exist. Since this code is part of the expander, to reproduce this behavior I'd run the following command `dune build %{cma:dir/disabled-lib}`.

From the perspective of the user, dune telling me this is disabled would be rather undesirable. So we need to improve the situation somehow. To do so, we need `Artifacts_obj.lookup_library` to propagate the library being disabled.

I make the conclusion that your decision is indeed correct and we need to pass enabled/disabled to `Artifacts_obj`

Hope that helps.

jchavarri: Just to be clear, this process would help on the eventual case that in some future Dune sets "fake" rules for all the disabled libraries in the current context right? As we confirmed the other day that there's no need to set rules for disabled libs right now.

I also wonder, in this hypothetic future where there are fake rules for disabled libraries (for better user errors), how would we deal with the case where two libraries are defined in the same folder with the same name, but one defined on each context? I assume we would skip the creation of fake rules in that case?

rgrinberg: Yes, we discussed using fake rules to make the error messages better. And indeed the decision was to forego them for now, as it's quite complex to implement. However, there's other cases where we can improve the error messages without much complexity. Expanding percent forms like we have here is an obvious candidate.

In your hypothetical case we would skip the fake rules. As they would overlap with actual rules.

test/blackbox-tests/test-cases/test-build-if/feature.t
```diff
@@ -1,4 +1,4 @@
-enabled_if has a limitation: it attempts building even if enabled_if evaluates to false.
+enabled_if and build_if have similar behavior
```
jchavarri, line 1: cc @emillon. This PR seems to fix the case found in #7899.

test/blackbox-tests/test-cases/melange/enabled_if.t
```diff
@@ -34,7 +34,13 @@
   > EOF
 
   $ dune build @melange --display short
+  Error: Alias "melange" specified on the command line is empty.
+  It is not defined in . or any of its descendants.
+  [1]
```
jchavarri, line 39: These errors might be more meaningful to users than the previous behavior (not building anything because the alias is "empty", but not erroring either).

src/dune_rules/ml_sources.ml
```diff
@@ -53,13 +54,19 @@ module Modules = struct
     }
 
   let make { libraries = libs; executables = exes; melange_emits = emits } =
+    let keep_enabled t =
+      List.filter t ~f:(fun (_, _, _, _, enabled) -> Toggle.enabled enabled)
+    in
```
anmonteiro, line 59:
```suggestion
  let make = 
    let keep_enabled t =
      List.filter t ~f:(fun (_, _, _, _, enabled) -> Toggle.enabled enabled)
    in
    fun { libraries = libs; executables = exes; melange_emits = emits } ->

```

src/dune_rules/ml_sources.ml
```diff
@@ -393,6 +400,8 @@ let modules_of_stanzas =
     Memo.parallel_map stanzas ~f:(fun stanza ->
       match Stanza.repr stanza with
       | Library.T lib ->
+        let* enabled = Expander.eval_blang expander lib.enabled_if in
+        let enabled = Toggle.of_bool enabled in
```
anmonteiro, line 404:
```suggestion
        let* enabled = 
          let+ enabled = Expander.eval_blang expander lib.enabled_if in
          Toggle.of_bool enabled
        in
```

src/dune_rules/ml_sources.ml
```diff
@@ -411,9 +420,11 @@ let modules_of_stanzas =
           >>= Resolve.read_memo
         in
         let obj_dir = Library.obj_dir lib ~dir in
-        `Library (lib, sources, modules, obj_dir)
+        `Library (lib, sources, modules, obj_dir, enabled)
       | Executables.T exes | Tests.T { exes; _ } ->
         let obj_dir = Executables.obj_dir ~dir exes in
+        let* enabled = Expander.eval_blang expander exes.enabled_if in
+        let enabled = Toggle.of_bool enabled in
```
anmonteiro, line 427:
```suggestion
        let* enabled = 
          let+ enabled = Expander.eval_blang expander exes.enabled_if in
          Toggle.of_bool enabled
        in
```

src/dune_rules/ml_sources.ml
```diff
@@ -434,9 +445,11 @@ let modules_of_stanzas =
           then Modules_group.make_wrapped ~obj_dir ~modules `Exe
           else Modules_group.exe_unwrapped modules ~obj_dir
         in
-        `Executables (exes, sources, modules, obj_dir)
+        `Executables (exes, sources, modules, obj_dir, enabled)
       | Melange_stanzas.Emit.T mel ->
         let obj_dir = Obj_dir.make_melange_emit ~dir ~name:mel.target in
+        let* enabled = Expander.eval_blang expander mel.enabled_if in
+        let enabled = Toggle.of_bool enabled in
```
anmonteiro, line 452:
```suggestion
        let* enabled =
          let+ enabled = Expander.eval_blang expander mel.enabled_if in
          Toggle.of_bool enabled
        in
```

src/dune_rules/ml_sources.ml
anmonteiro: or otherwise `let enabled = Expander.eval_blang expander lib.enabled_if >>| Toggle.of_bool`

anmonteiro (approved):

src/dune_rules/ml_sources.ml
```diff
@@ -45,6 +45,7 @@ module Modules = struct
     * (Loc.t * Module.Source.t) Module_trie.t
```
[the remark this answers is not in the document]
jchavarri: I agree. I can do so in a follow-up PR.

test/blackbox-tests/test-cases/test-build-if/feature.t
rgrinberg: The old behavior was in fact correct and this PR breaks it. The description of this test might be the source of confusion. There's no limitation anywhere, it's just that the different semantics were chosen for `enabled_if` on tests.

src/dune_rules/artifacts_obj.mli
rgrinberg: I added a test case for this in https://github.com/ocaml/dune/pull/10245

Looks like we don't handle disabled libraries correctly

rgrinberg (changes requested):

src/dune_rules/artifacts_obj.mli
jchavarri: The change requests didn't specify if this is something that should be fixed in this PR. Is that the case? In any case, the PR doesn't seem to change the current behavior.

test/blackbox-tests/test-cases/test-build-if/feature.t
jchavarri: Just as a side note, if I'm understanding the current behavior properly, it is quite confusing from a user perspective. `enabled_if` semantics are now different depending on where it's used:

- In `executable` stanzas, the field means the executable is not built if it evaluates to `false`
- In `test` stanzas:
  - when `all` alias is used, the field means the test executable is always built, regardless its evaluated value
  - but for `runtest` alias it means the executable will not be built if it evaluates to `false` (like `executable` stanzas)

I found this comment that explains the use case for it: https://github.com/ocaml/dune/pull/7899#issuecomment-1581552347.

It's probably too late, but wonder if it would have been possible to leave the behavior of `enabled_if` consistent, and allow advanced use cases to use a combination of `rule` and `executable`, or add a new field `test_if`.

src/dune_rules/artifacts_obj.mli
rgrinberg: No, there's no need to fix it here.

test/blackbox-tests/test-cases/test-build-if/feature.t
rgrinberg: Yes, in hindsight it would be better to do it your way. We could revisit it for 4.0

rgrinberg:
I improved the PR as follows:

1. stop propagating enabled_if when they aren't being used. we should propagate them to improve error messages in the future, but we can save that until its actually needed.
2. correctly handle build_if for tests
3. avoid evaluating enabled_if on melange.emit unless it's needed (to avoid possible cycles)
4. do not compute modules for disabled stanzas.

Should be good once you add a CHANGES entry

anmonteiro (approved):

PR #10226, opened 2024-03-07
Test error cases of OSL and predicate langugae

gridbugs:


rgrinberg (approved):
LGTM. Could you add descriptions to the tests? A couple of sentences per test is fine.

PR #10231, opened 2024-03-07
refactor: treat deprecated lib case explicitly

jchavarri:
This change will help with enabling multi-context libraries in #10179.

The reason is that with the changes in that PR, the checks in `create_db_from_stanzas` change a bit, with two redirects being a potentially valid situation (e.g. if there are two public libraries with the same name defined in different contexts).

Having a way to explicitly recognize the deprecated library case allows to keep the behavior of that feature untouched while unlocking multi-context public libraries.

### DIFF
--- a/src/dune_rules/lib.ml
+++ b/src/dune_rules/lib.ml
@@ -412,6 +412,7 @@ and resolve_result =
   | Ignore
   | Redirect_in_the_same_db of (Loc.t * Lib_name.t)
   | Redirect of db * (Loc.t * Lib_name.t)
+  | Deprecated_library_name of (Loc.t * Lib_name.t)
 
 let lib_config (t : lib) = t.lib_config
 let name t = t.name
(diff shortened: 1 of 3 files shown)


rgrinberg:
I'd keep it in #10231 . Off the top of my head, I don't see how this is going to help to implement enabled_if for libraries.

jchavarri:
The main reason is that to enable multi-context libs with the same name, we need to allow `Redirect` with duplicated names to "go through" in the database to support the case with public libs.

Essentially, the following line needs to change to remove the `Lib_name.equal` change and instead add both to the (newly added) list of libs:

https://github.com/ocaml/dune/blob/1f8042665aa62476b7a8d15add2f744fdd731219/src/dune_rules/scope.ml#L74

But if we do that change without introducing the new variant, we will break the deprecated libs use case.

jchavarri:
> I'd keep it in #10231 .

10231 is this PR. Which one are you referring to?

rgrinberg:
I meant #10179.

It could be very well the case that this PR is useful, but it's impossible to tell without seeing the final use of it. On its own, it does not yet offer any value (not in readability, performance, etc)

jchavarri:
ftr applied the changes in https://github.com/ocaml/dune/pull/10179/commits/845ac40d62115f2b466bfa2f4f43cdf69c101c3f.

Issue #10232, opened 2024-03-08
Changes made to non-dune pinned packages are not visible until `dune clean` is run


gridbugs:
After dune attempts to build a non-dune pinned package, changes to the pinned package are not picked up by dune until `dune clean` is run, even if the attempt to build the pinned package failed.

Changes to the pinned package should be picked up immediately regardless of the state of the current project and trigger rebuilds as necessary.

Repro test added in: https://github.com/ocaml/dune/pull/10233

Issue #9700, opened 2024-01-10
Remove the `copy` constructor from `Dune_pkg.Source.t`


rgrinberg:
Writing `(source (copy ..))` should be the same as `(source (fetch file://..))` so `(copy ..)` is redundant.

rgrinberg:
cc @gridbugs @Leonidas-from-XIV any objections to doing this?

gridbugs:
No objection from me.

Leonidas-from-XIV:
Nor from me.

PR #10233, opened 2024-03-08
Add repro for gh10232

gridbugs:
Add a test reproducing https://github.com/ocaml/dune/issues/10232


rgrinberg:
Okay, I understand the issue. I think this will be fixed once we address https://github.com/ocaml/dune/issues/9700

What we'd like is that `file://` is to be implemented like `copy` is right now, and for `copy` to go away.

rgrinberg (approved):

Issue #10234, opened 2024-03-08
Changes made to the build/install commands of a pinned non-dune package are not picked up until `dune pkg lock` is run


gridbugs:
Normally users can change files in the source of a pinned package and expect the changes to be visible to the current project (which pins said package), though currently running `dune clean` is necessary (see https://github.com/ocaml/dune/issues/10232). This enables the workflow of testing local changes to a library by cloning one of its reverse dependencies and pinning the local source of the library.

However, if the build/install commands in the pinned package's opam file are changed, the changes are not visible to the current project until `dune pkg lock` is run.

Repro: TODO

This happens because the build/install commands from the pinned package are copied to the lockdir during `dune pkg lock` and are subsequently used by dune to build the pinned package. One way to avoid this issue would be deferring parsing the build/install commands of the pinned package until dune attempts to build the package.

PR #10235, opened 2024-03-08
Add repro for gh10234

gridbugs:
Adds repro test for https://github.com/ocaml/dune/issues/10234

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/pin-stanza/update-non-dune-local-pin-opam-file.t
+++ b/test/blackbox-tests/test-cases/pkg/pin-stanza/update-non-dune-local-pin-opam-file.t
@@ -1,0 +1,105 @@
+This demonstrates an issue when pinning a non-opam package. Changes made to the
+pinned package's source are visible in the current project _unless_ the pinned
+package's opam file is changed, in which case it's necessary to run `dune pkg
+lock`. This is because the pinned package is built using build commands copied
+to the package's lockfile in the current project, and they are only re-copied
+when running `dune pkg lock`. This is inconsistent with how dune treats other
+files in the pinned package.
+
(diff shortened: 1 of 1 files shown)


rgrinberg:
This should be same as https://github.com/ocaml/dune/issues/9700 as well - but on the lock file generation side. We're currently generating `file://`, but it's buggy. So we need to make it robust like `copy`.

gridbugs:
I don't understand how this can be fixed by removing the `copy` constructor. The issue here is that out of date build commands are being run from the lockfile. We would need a way of checking if the opam file in the pinned package had changed and regenerating its lockfile if necessary, or a way of running the build commands from the pinned package's opam file directly rather than from its lockfile.

rgrinberg:
Good point. Why are we even recording the build command for pins again? If we have the checksum for the source, that already fixes the build command. If we don't have the cehcksum, the build command is useless anyway

test/blackbox-tests/test-cases/pkg/pin-stanza/update-non-dune-local-pin-opam-file.t
```diff
@@ -0,0 +1,105 @@
+This demonstrates an issue when pinning a non-opam package. Changes made to the
+pinned package's source are visible in the current project _unless_ the pinned
+package's opam file is changed, in which case it's necessary to run `dune pkg
+lock`. This is because the pinned package is built using build commands copied
+to the package's lockfile in the current project, and they are only re-copied
+when running `dune pkg lock`. This is inconsistent with how dune treats other
+files in the pinned package.
+
+  $ . ../helpers.sh
```
Alizter, line 9: Needs to be removed, it is handled by the parent cram stanza with the `setup_scripts` field.

Issue #10236, opened 2024-03-08
dune exec path shouldn't change when using watch mode


Khady:
## Expected Behavior

PWD of a command ran by `dune exec` should be the same in and out of watch mode.

## Actual Behavior

PWD in watch mode is always at the root of the dune project. While in normal mode it is where dune was launched.

## Reproduction

```
/tmp$ dune init project demo
Entering directory '/tmp/demo'
Success: initialized project component named demo
/tmp$ cd demo
/tmp/demo$ dune exec -- pwd
/tmp/demo                   <-------- this is where dune was launched
/tmp/demo$ cd bin
/tmp/demo/bin$ dune exec -- pwd
Entering directory '/tmp/demo'
Leaving directory '/tmp/demo'
/tmp/demo/bin                   <-------- this is where dune was launched
/tmp/demo/bin$ dune exec -w -- pwd
Entering directory '/tmp/demo'

/tmp/demo <---- this is NOT where dune was launched
```

## Specifications

- Version of `dune` (output of `dune --version`): 3.14.0
- Version of `ocaml` (output of `ocamlc --version`): 4.14.1
- Operating system (distribution and version): linux

gridbugs:
Should be fixed by https://github.com/ocaml/dune/pull/10262

Khady:
Thanks!

PR #10241, opened 2024-03-08
add reproduction test for #10236

tatchi:
This adds a reproduction test for #10236. Unfortunately, watch tests are disabled due to some issues on CI (+macos)

https://github.com/ocaml/dune/blob/c41993f7c03878017c5fe390117a013d79795f3d/test/blackbox-tests/test-cases/exec-watch/dune#L1-L15

Perhaps this one would cause less problems as we don't actually need to watch anything? In any case, I hope this will help to fix the problem.


rgrinberg:
@gridbugs could you have a look at this one?

gridbugs:
Happy for this to be merged. I've got a PR open which fixes this issue (https://github.com/ocaml/dune/pull/10262) which I will merge after the current PR is merged.

tatchi:
Thanks for the fix @gridbugs 🙏

PR #10247, opened 2024-03-11
refactor: high-level dune-package rewriting

emillon:
`dune install` rewrites `dune-package` files on the fly to store sections as computed by the installation step.

The previous implementation used to rely on sexp manipulation, which is brittle and does not evolve with the `dune-package` codecs.

This uses a more end-to-end approach where the data structure is decoded, modified, and encoded again.
    
Some aspects are not as easy as they seem because there are different encodings for paths; some are relative to the package root, other from the `Stublibs` section, so the API is not as simple as it could be.

### DIFF
--- a/bin/install_uninstall.ml
+++ b/bin/install_uninstall.ml
@@ -206,6 +206,7 @@ module File_ops_real (W : sig
   let print_line = print_line ~verbosity
   let get_vcs p = Dune_rules.Vcs_db.nearest_vcs p
 
+  (* CR-emillon rework this API or at least remove the optional argument *)
   type load_special_file_result =
     { need_version : bool
     ; callback : ?version:string -> Format.formatter -> unit
--- a/bin/install_uninstall.ml
+++ b/bin/install_uninstall.ml
@@ -217,7 +217,7 @@ module File_ops_real (W : sig
 
   let copy_special_file ~src ~package ~ic ~oc ~f =
     let open Fiber.O in
-    match f ic with
+    match f ic ~src with
     | None -> Fiber.return Use_plain_copy
     (* XXX should we really be catching everything here? *)
     | exception _ ->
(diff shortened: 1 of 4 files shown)


bin/install_uninstall.ml
```diff
@@ -269,69 +269,32 @@ module File_ops_real (W : sig
...
+    match Dune_package.Or_meta.parse src lb |> User_error.ok_exn with
+    | Use_meta ->
+      let callback ?version:_ ppf = Dune_package.Or_meta.pp_use_meta ~dune_version ppf in
+      Some { need_version = false; callback }
+    | Dune_package dp ->
+      (* replace sites with external path in the file *)
+      let dp, replace_info = Dune_package.replace_site_sections ~get_location dp in
+      (* replace version if needed in the file *)
+      let need_version = Option.is_none dp.version in
+      let callback ?version ppf =
```
rgrinberg, line 284: We should get rid of this optional argument. Not necessarily in this PR, but at least leave a note.

rgrinberg (approved):
LGTM. Let's have @anmonteiro have a look as well so he familiarizes himself with the code

emillon:
Thanks @anmonteiro.
It's definitely something lacking in the current situation. I would very much prefer to have a plain `encode` and `decode` that would allow an easy round trip. The problem is that paths in dune-package files are relative (sometimes to the package root, sometimes to the stublibs dir), so they're not easily loadable from arbitrary paths. This led to this weird API; I still believe that it's an improvement compared to the sexp edition logic we had before.
One alternative I've considered is to deserialize paths as sexps and do the conversion in a separate step. This would imply an extra type, or at least extra type parameters here and there. I'm not sure it's worth it; what do you think?
And in terms of evolution, I'd like to see if it's possible to remove that substitution step completely. I'm not sure why it's necessary to do that, compared to generating the right `.install` file directly, but I'm not yet familiar enough with `dune-sites` to see that.

PR #10248, opened 2024-03-11
refactor(install): use Path.of_local

emillon:
`append_local` creates source paths that look like `_build/install/...`. While these are "correct", `Path.Source.is_in_build_dir` returns true for these. In contrast, `Path.of_local` will detect the `_build` part and make a build path.

### DIFF
--- a/bin/install_uninstall.ml
+++ b/bin/install_uninstall.ml
@@ -606,8 +606,7 @@ let run
       let entries_per_package =
         List.map install_files ~f:(fun (package, install_file) ->
           let entries =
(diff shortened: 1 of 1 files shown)


emillon:
You're right. Initially I had named that `refactor(install)` because indeed there's no visible change in behavior (neither intended nor observed).
The context is #10247. Without this change, we can't parse dune-package files and reprint them because the paths won't match if they're "buggy" (`Source` but in `_build`).
I'll rename the commit.

rgrinberg (approved):

PR #10250, opened 2024-03-11
feature: remove limitations on [enabled_if] on libraries

rgrinberg:
Previously, the `enabled_if` field on libraries had a bunch of limitations. Basically, only variables that could be computed from the context could be used. E.g. `profile`, `context_name`, etc. This PR removes this limitation and now every variable is allowed provided that it doesn't introduce a memoization cycle.

Normally, such a feature would be trivial, but here we some module dependency cycles that prevent the natural solution from working. If it was possible, we would just call `Super_context.expander ~dir sctx |> Expander.eval_blang` everywhere and be done with it. Unfortunately, we have to solve two problems:

1. Expander depends on scope, and we need `enabled_if` to compute the scope
2. Super_context depends on scope, and therefore on various `enabled_if` through 1.

So we break this cycle by introducing an `Expander0.t` that doesn't depend on scope and has an expander db that is the equivalent of `Super_context.expander`.

### DIFF
--- a/src/dune_rules/expander0.ml
+++ b/src/dune_rules/expander0.ml
@@ -24,3 +24,26 @@ let as_in_build_dir ~what ~loc p =
           (Path.to_string_maybe_quoted p)
       ]
 ;;
+
+module type S = sig
+  type t
+
+  val project : t -> Dune_project.t
+  val eval_blang : t -> Blang.t -> bool Memo.t
+end
+
+open Memo.O
+
+type t = E : 'a Memo.t * (module S with type t = 'a) -> t
+
+let db = Fdecl.create Dyn.opaque
+let set_db = Fdecl.set db
+let create e m = E (e, m)
+let project (E (e, (module E))) = Memo.map e ~f:E.project
+
+let eval_blang (E (e, (module E))) blang =
+  let* e = e in
+  E.eval_blang e blang
+;;
+
+let get ~dir = (Fdecl.get db) ~dir
--- a/src/dune_rules/scope.ml
+++ b/src/dune_rules/scope.ml
@@ -52,7 +52,7 @@ module DB = struct
   end
 
   let create_db_from_stanzas ~instrument_with ~parent ~lib_config stanzas =
-    let (map : Found_or_redirect.t Lib_name.Map.t) =
+    let map =
       List.map stanzas ~f:(fun stanza ->
         match (stanza : Library_related_stanza.t) with
         | Library_redirect s ->
@@ -62,7 +62,10 @@ module DB = struct
           let old_public_name = Deprecated_library_name.old_public_name s in
           Found_or_redirect.redirect old_public_name s.new_public_name
         | Library (dir, (conf : Library.t)) ->
-          let info = Library.to_lib_info conf ~dir ~lib_config |> Lib_info.of_local in
+          let info =
+            let expander = Expander0.get ~dir in
+            Library.to_lib_info conf ~expander ~dir ~lib_config |> Lib_info.of_local
+          in
           Library.best_name conf, Found_or_redirect.found info)
       |> Lib_name.Map.of_list_reducei ~f:(fun name (v1 : Found_or_redirect.t) v2 ->
         let res =
@@ -100,11 +103,12 @@ module DB = struct
       ~parent:(Some parent)
       ~resolve:(fun name ->
         Memo.return
-          (match Lib_name.Map.find map name with
-           | None -> Lib.DB.Resolve_result.not_found
-           | Some (Redirect lib) -> Lib.DB.Resolve_result.redirect_in_the_same_db lib
-           | Some (Found lib) -> Lib.DB.Resolve_result.found lib))
-      ~all:(fun () -> Lib_name.Map.keys map |> Memo.return)
+        @@
+        match Lib_name.Map.find map name with
+        | None -> Lib.DB.Resolve_result.not_found
+        | Some (Redirect lib) -> Lib.DB.Resolve_result.redirect_in_the_same_db lib
+        | Some (Found lib) -> Lib.DB.Resolve_result.found lib)
+      ~all:(fun () -> Memo.return @@ Lib_name.Map.keys map)
       ~lib_config
       ~instrument_with
   ;;
--- a/src/dune_rules/super_context.ml
+++ b/src/dune_rules/super_context.ml
@@ -382,7 +382,13 @@ let () =
     let* ctx = Context.DB.by_dir dir in
     let* t = find_exn (Context.name ctx) in
     let* expander = expander t ~dir in
-    Expander.expand_str expander sw |> Action_builder.evaluate_and_collect_facts >>| fst)
+    Expander.expand_str expander sw |> Action_builder.evaluate_and_collect_facts >>| fst);
+  Expander0.set_db (fun ~dir ->
+    Context.DB.by_dir dir
+    >>| Context.name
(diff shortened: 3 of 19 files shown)


rgrinberg:
@jchavarri since you've expressed interest in this PR and have looked at this code lately, I've added you as a reviewer. I've also improved the description of the implementation.

src/dune_rules/super_context.ml
```diff
@@ -382,7 +382,13 @@ let () =
     let* ctx = Context.DB.by_dir dir in
     let* t = find_exn (Context.name ctx) in
     let* expander = expander t ~dir in
-    Expander.expand_str expander sw |> Action_builder.evaluate_and_collect_facts >>| fst)
+    Expander.expand_str expander sw |> Action_builder.evaluate_and_collect_facts >>| fst);
+  Expander0.set_db (fun ~dir ->
```
jchavarri, line 386: I'm not sure I understand why `Expander0` needs its own db. I wonder if there's a way to introduce that module without having to add more state / side effects?

test/blackbox-tests/test-cases/enabled_if/eif-library-name-collision.t
```diff
@@ -40,6 +40,7 @@ For private libraries
   Error: Library foo is defined twice:
   - a/dune:1
   - b/dune:1
+  -> required by alias default
```
jchavarri, line 43: Can't track why this line appears now by looking at the code. Do you know?

src/dune_rules/scope.ml
```diff
@@ -52,59 +52,64 @@ module DB = struct
...
-          let main_message =
-            Pp.textf "Library %s is defined twice:" (Lib_name.to_string name)
-          in
-          let annots =
-            let main = User_message.make ~loc:loc2 [ main_message ] in
-            let related =
-              [ User_message.make ~loc:loc1 [ Pp.text "Already defined here" ] ]
+    let open Memo.O in
+    let map =
+      Memo.lazy_ (fun () ->
```
jchavarri, line 57: Is this performance optimization related in some way to the main purpose of the PR?

src/dune_rules/expander0.ml
```diff
@@ -24,3 +24,26 @@ let as_in_build_dir ~what ~loc p =
...
+module type S = sig
+  type t
+
+  val project : t -> Dune_project.t
+  val eval_blang : t -> Blang.t -> bool Memo.t
+end
+
+open Memo.O
+
+type t = E : 'a Memo.t * (module S with type t = 'a) -> t
```
jchavarri, line 37: Why are GADTs / witness necessary? Would it be possible to implement it with simpler types?

jchavarri (commented):
I asked some questions, mostly trying to follow what's going on. Probably this change needs a changelog entry.

> So we break this cycle by introducing an `Expander0.t` that doesn't depend on scope and has an expander db that is the equivalent of `Super_context.expander`.

I took a look at removing the cycle by moving `lookup_artifacts` out of `Expander`, but more cycles surfaced:

```
   _build/default/src/dune_rules/.dune_rules.objs/dune_rules__Scope.intf.all-deps
-> _build/default/src/dune_rules/.dune_rules.objs/dune_rules__Expander.intf.all-deps
-> _build/default/src/dune_rules/.dune_rules.objs/dune_rules__Library.intf.all-deps
-> _build/default/src/dune_rules/.dune_rules.objs/dune_rules__Library_redirect.intf.all-deps
-> _build/default/src/dune_rules/.dune_rules.objs/dune_rules__Deprecated_library_name.intf.all-deps
-> _build/default/src/dune_rules/.dune_rules.objs/dune_rules__Scope.intf.all-deps
```

I wonder if it'd be worth exploring that path more, as increasing the amount of code in `Expander0` seems like a short term solution rather than something maintainable in the long term to me? Probably this feeling comes from code duplication and new state being added in `set_db`, which I understand could lead to situations where `Expander` and `Expander0` could get "out of sync".

src/dune_rules/super_context.ml
rgrinberg: It is not its own db, it's just a forward declaration for the db in super context. I do not know how to break cycles without using forward references. Open to suggestions though

test/blackbox-tests/test-cases/enabled_if/eif-library-name-collision.t
rgrinberg: Since the evaluation is a bit delayed, we're running into the when executing the rule rather than loading them. It's not an issue. It's usually an issue when we do the opposite, and raise too eagerly.

rgrinberg:
> I asked some questions, mostly trying to follow what's going on. Probably this change needs a changelog entry.
> 
> > So we break this cycle by introducing an `Expander0.t` that doesn't depend on scope and has an expander db that is the equivalent of `Super_context.expander`.
> 
> I took a look at removing the cycle by moving `lookup_artifacts` out of `Expander`, but more cycles surfaced:
> 
> ..
> 
> I wonder if it'd be worth exploring that path more, as increasing the amount of code in `Expander0` seems like a short term solution rather than something maintainable in the long term to me? Probably this feeling comes from code duplication and new state being added in `set_db`, which I understand could lead to situations where `Expander` and `Expander0` could get "out of sync".

You can try exploring it, but there are fundamental issues here that make it impossible to factor our code enough to prevent cycles. There's three different concerns:

1. Expanding a percent form - adds a dependency on the public API of the expander (pforms, templates, projects, etc)
2. Implementing an expander - requires knowing all the types of data that can exist inside an expander (scopes, artifacts, etc)
3. Constructing a percent form - requires knowing how to make an expander and populate it with data. The data can come from just about everywhere (super_context, env_node, dir_contents, etc)

You can see how steps 2. and 3. are limiting. Anytime you want to expand something, pulling in 2 or 3 might not be an option.
So in the end we do need a way to expand things with only 1.

Another thing we can try is to pass an expansion function around to those that need to expand. That ends up creating a bunch of inconsistency bugs because this exapnsion function is being constructed in a slightly different way everywhere.

src/dune_rules/super_context.ml
jchavarri, line 386: Are there other places where we potentially should set this db? E.g. `Target.expand_path_from_root`, `Artifacts_db.get_installed_binaries`... (i'm just grepping for usages of `Expander.expand_str`).

test/blackbox-tests/test-cases/enabled_if/enabled-if-lib-other-vars.t
```diff
@@ -0,0 +1,16 @@
+Here we demonstrate that we expand any variable in enabled_if on libraries
```
jchavarri, line 1: Are there other tests that were showing the previous behavior? If that's the case, might be worth to move this case over there or put them together in some way?

jchavarri (approved):

src/dune_rules/super_context.ml
rgrinberg: It's a forward declaration, so it can only be set once. However, there are more cases where we can use this db. For example, in `Artifacts`. I have a PR coming up to do that.

test/blackbox-tests/test-cases/enabled_if/enabled-if-lib-other-vars.t
rgrinberg: Yes, see the other tests updated in this PR. The previous behavior is still preserved (although the error message is tweaked)

PR #10256, opened 2024-03-12
fix(spawn): fix headers issues on non-glibc

emillon:
`caml/signals.h` is not public API; it does not include `caml/config.h` and so it will not include `signal.h` because it does not know about `POSIX_SIGNAL`. This issue is visible on musl as "unknown type `sigset_t`".

### DIFF
--- a/vendor/update-spawn.sh
+++ b/vendor/update-spawn.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-version=d5991f7eb4073c85b440789587936d14b4e0417f
+version=6c752122070f377c2607f6969a821f166a43bd5e
 
 set -e -o pipefail
 
--- a/doc/changes/10256.md
+++ b/doc/changes/10256.md
@@ -1,0 +1,2 @@
+- fix compilation on non-glibc systems due to `signal.h` not being pulled in
+  spawn stubs. (#10256, @emillon)
--- a/vendor/spawn/src/spawn_stubs.c
+++ b/vendor/spawn/src/spawn_stubs.c
@@ -15,16 +15,14 @@
 #endif
 #endif
 
-/* for [caml_convert_signal_number] */
-#include <caml/signals.h>
-
-#undef CAML_INTERNALS
-
 #include <caml/mlvalues.h>
 #include <caml/memory.h>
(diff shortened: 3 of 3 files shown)


emillon:
I tested that patch on @anmonteiro 's nix-overlays using `nix build -L .#hydraJobs.x86_64-linux.musl_4_14.tls`.

vendor/update-spawn.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-version=d5991f7eb4073c85b440789587936d14b4e0417f
+version=6c752122070f377c2607f6969a821f166a43bd5e
```
Leonidas-from-XIV, line 3: Is this part of `janestreet/spawn`? It seems to be [a different fork](https://github.com/janestreet/spawn/commit/6c752122070f377c2607f6969a821f166a43bd5e) so this will fail on others who haven't pulled your branch. I suggest changing the URL to your fork temporarily for this issue until the fix is merged.

Leonidas-from-XIV (commented):
Great investigation; could you make a PR to `janestreet/spawn` so we won't accidentally lose this fix in the future?

I remember there was some discussion on guarding `caml/signals.h` against accidental misuse in the future, maybe worth opening a ticket/PR while it is fresh in our minds?

The patch itself is good, I have no qualms.

vendor/update-spawn.sh
emillon: We point at ocaml-dune/spawn. I've made this PR by running the update script so it will work for others too.
(in addition to this patch, the branch has a macos fix which is an unmerged upstream PR)

emillon:
> I remember there was some discussion on guarding `caml/signals.h` against accidental misuse in the future, maybe worth opening a ticket/PR while it is fresh in our minds?

the version in `trunk` looks like it doesn't need `config.h` so I think it's immune to that problem.

Leonidas-from-XIV (approved):

vendor/update-spawn.sh
Leonidas-from-XIV: Ah, fair point. I looked at the wrong place.

PR #10262, opened 2024-03-13
Exec watch mode spawns processes with initial cwd

gridbugs:
Fixes a bug where commands run from `dune exec -w` would have the cwd of the project root rather than the directory where the command was run.

Fixes https://github.com/ocaml/dune/issues/10236

I haven't added any tests as a good test is added in https://github.com/ocaml/dune/pull/10241.


rgrinberg (approved):
Needs a CHANGES entry

emillon:
Can you first comment on #10241 so we can merge it before this one?

Issue #8931, opened 2023-10-12
pkg: Zarith doesn't build


Alizter:
Building Zarith fails with:
```
ocamlfind install -destdir "/home/ali/HoTT/outputs/out/lib/ocaml/4.14.1/site-lib/" zarith META zarith.cma libzarith.a z.cmi q.cmi big_int_Z.cmi zarith_top.cma z.mli zarith.cmxa zarith_version.cmx z.cmx q.cmx big_int_Z.cmx zarith.cmxs zarith.h q.mli big_int_Z.mli zarith.a z.cmti q.cmti big_int_Z.cmti -optional dllzarith.so
ocamlfind: Bad configuration: Cannot mkdir /home/ali/HoTT/outputs/out/lib/ocaml/4.14.1/site-lib/zarith because a path component does not exist or is not a directory
```
This is using `DUNE_PKG_OVERRIDE_OCAML=1` for the lock and build. And the following dune-project:
```
(lang dune 3.11)

(package
 (name foo)
 (dependencies
  (zarith (= 1.13))))
```

Alizter:
The issue here is really that `ocamlfind install` doesn't work.

rgrinberg:
Yeah, that's a shame. There's a few packages like that we're going to add workarounds for. I think the simplest thing to do for now is to introduce our own overlay system like duniverse. We will need very few overlays in comparison to duniverse.

Also, it would be nice if we had a way to mark packages as being incompatible with dune's package management.

cc @Leonidas-from-XIV who worked with something like this in duniverse already.

Alizter:
Zarith is able to be packages up for Nix, but obviously that doesn't use the opam file.

https://github.com/NixOS/nixpkgs/blob/397aa97b0eb578f15534580461f7d3e9e2eb0c93/pkgs/development/ocaml-modules/zarith/default.nix#L7

They configure with the correct paths which is perhaps something we could do. Or at least modify the OPAM file for Zarith to allow.

rgrinberg:
Or we can use the dune build system Hugo write. Any of those options will require an overlay until the upstream package is updated.

Leonidas-from-XIV:
> cc @Leonidas-from-XIV who worked with something like this in duniverse already.

Overall I'd say the UX for this is pretty bad at large. 

What we would need is 

1. A way to mark packages as not buildable in the dune package management system (in `opam-monorepo` we required that the build instructions use `dune`, but here we'd need to know through some other means)
[...]

PR #10267, opened 2024-03-14
pkg: Set OCAMLFIND_DESTDIR for install actions

### LINKED ISSUE #10252: pkg: support building packages with ocamlfind
Once the issue with relocating ocamlfind (https://github.com/ocaml/ocamlfind/pull/72) we'll still need to to have it build packages in a way that dune expects. There are two parts to this:
 - ocamlfind must be able to find other libraries built with dune
 - libraries/exes built with ocamlfind must be placed in a location dune expects so other packages can depend on them

gridbugs:
If a package uses ocamlfind to install its files then ocamlfind will consult its config file to determine where in the filesystem the installed files ought to go. In dune, each package is built in an isolated sandbox, and so by default ocamlfind won't install files to the correct location. The OCAMLFIND_DESTDIR environment variable can be used to override ocamlfind's default behaviour for determining where to install files.

This change sets the OCAMLFIND_DESTDIR variable to the "target/lib" directory within a package's build sandbox when executing a package's install command. Build commands are unaffected by this change.

Fixes https://github.com/ocaml/dune/issues/10252.

Necessary for fixing https://github.com/ocaml/dune/issues/8931, but we also need to merge https://github.com/ocaml/ocamlfind/pull/72 (relocatable ocamlfind) to fix that issue.

In addition to the simple test included in this change, I've also manually tested this change by building zarith using a patched version of ocamlfind (patched with https://github.com/ocaml/ocamlfind/pull/72).

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -456,6 +456,7 @@ module Run_with_path = struct
     type ('path, 'target) t =
       { prog : Action.Prog.t
       ; args : 'path arg Array.Immutable.t
+      ; ocamlfind_destdir : 'path
       }
 
     let name = "run-with-path"
@@ -467,10 +468,16 @@ module Run_with_path = struct
         | Path p -> Path (f p))
     ;;
 
-    let bimap t f _g = { t with args = Array.Immutable.map t.args ~f:(map_arg ~f) }
+    let bimap t f _g =
+      { t with
+        args = Array.Immutable.map t.args ~f:(map_arg ~f)
+      ; ocamlfind_destdir = f t.ocamlfind_destdir
+      }
+    ;;
+
     let is_useful_to ~memoize:_ = true
 
-    let encode { prog; args } path _ : Dune_lang.t =
+    let encode { prog; args; ocamlfind_destdir } path _ : Dune_lang.t =
       let prog =
         Dune_lang.atom_or_quoted_string
         @@
@@ -485,10 +492,17 @@ module Run_with_path = struct
               | String s -> Dune_lang.atom_or_quoted_string s
               | Path p -> path p)))
       in
-      List ([ Dune_lang.atom_or_quoted_string name; prog ] @ args)
+      List
+        [ List ([ Dune_lang.atom_or_quoted_string name; prog ] @ args)
+        ; path ocamlfind_destdir
+        ]
     ;;
 
-    let action { prog; args } ~(ectx : Action.Ext.context) ~(eenv : Action.Ext.env) =
+    let action
+      { prog; args; ocamlfind_destdir }
+      ~(ectx : Action.Ext.context)
+      ~(eenv : Action.Ext.env)
+      =
       let open Fiber.O in
       match prog with
       | Error e -> Action.Prog.Not_found.raise e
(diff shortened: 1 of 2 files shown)


src/dune_rules/pkg_rules.ml
```diff
@@ -467,10 +472,16 @@ module Run_with_path = struct
...
+      { t with
+        args = Array.Immutable.map t.args ~f:(map_arg ~f)
+      ; set_ocamlfind_destdir = Option.map ~f t.set_ocamlfind_destdir
+      }
+    ;;
+
     let is_useful_to ~memoize:_ = true
 
-    let encode { prog; args } path _ : Dune_lang.t =
+    let encode { prog; args; set_ocamlfind_destdir = _ } path _ : Dune_lang.t =
```
rgrinberg, line 484: might as well add it to the digest just in case

Leonidas-from-XIV:
I wonder whether it wouldn't make more sense to set `OPAM_SWITCH_PREFIX` instead of `OCAMLFIND_DESTDIR`? That way in theory more than just `ocamlfind` will know where to install to.

(The relocatable patch should probably make the `findlib.conf` file entirely optional, defaulting to `$PREFIX/lib` in case it is not found and if `OPAM_SWITCH_PREFIX` is set it should probably use that as `$PREFIX`)

gridbugs:
I just noticed that when I use dune with this patch to build zarith it still installs its files to the wrong location. My little test library that installs with ocamlfind installs as expected but zarith does not. I'll do more testing with zarith next week.

rgrinberg:
Did you confirm that zarith uses `ocamlfind install`? I recall it had a weird makefile that worked without ocamlfind

gridbugs:
The problem with zarith turns out to be that it explicitly passes `-destdir` to `ocamlfind install` and it gets the destdir by running `ocamlfind printconf destdir` during its _build_ command (it's part of the `configure` script). We might be able to fix this by setting `OCAMLFIND_DESTDIR` during the build command as well as during the install command.

src/dune_rules/pkg_rules.ml
```diff
@@ -958,7 +986,13 @@ module Action_expander = struct
   ;;
 
   let install_command context (pkg : Pkg.t) =
-    Option.map pkg.install_command ~f:(expand context pkg)
+    Option.map pkg.install_command ~f:(fun action ->
+      let* expander = expander context pkg in
+      let set_ocamlfind_destdir =
+        Some
+          (Path.relative (Path.build expander.paths.target_dir) (Section.to_string Lib))
```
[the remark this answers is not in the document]
gridbugs: Updated

src/dune_rules/pkg_rules.ml
gridbugs: done

gridbugs:
I've applied review feedback and fixed the failing test. @Leonidas-from-XIV regarding setting `OPAM_SWITCH_PREFIX`, I think we should set that in addition to `OCAMLFIND_DESTDIR`. There's an issue for this here: https://github.com/ocaml/dune/issues/10294

rgrinberg:
We already set OPAM_SWITCH_PREFIX. The issue is that it's not relocated and some packages do not respect it

gridbugs:
> We already set OPAM_SWITCH_PREFIX. The issue is that it's not relocated and some packages do not respect it

Makes sense.

I realized the test still had some stuff specific to my machine in it. That's now fixed. Does anything else need to be changed about this PR?

rgrinberg (approved):

PR #10270, opened 2024-03-15
feat(melange): support `esm` for module_systems too

anmonteiro:
In Melange, we're starting to rename `es6` to `esm` (for **E**CMA**S**cript **M**odules). This change will have to be very gradual, over at least one deprecation cycle. For now, let's start accepting `esm` in addition to `es6` too.

See https://github.com/melange-re/melange/issues/1066.

### DIFF
--- a/src/dune_rules/melange/melange.ml
+++ b/src/dune_rules/melange/melange.ml
@@ -2,14 +2,14 @@ open Import
 
 module Module_system = struct
   type t =
-    | Es6
-    | CommonJs
+    | ESM
+    | CommonJS
 
-  let default = CommonJs, ".js"
+  let default = CommonJS, ".js"
 
   let to_string = function
-    | Es6 -> "es6"
-    | CommonJs -> "commonjs"
+    | ESM -> "es6"
+    | CommonJS -> "commonjs"
   ;;
 end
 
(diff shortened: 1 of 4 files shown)


src/dune_rules/melange/melange.ml
```diff
@@ -2,13 +2,13 @@ open Import
...
   type t =
-    | Es6
+    | ESM
     | CommonJs
 
   let default = CommonJs, ".js"
 
   let to_string = function
-    | Es6 -> "es6"
+    | ESM -> "es6"
```
anmonteiro, line 11: still printing to `"es6"` for the command-line option since that's what all Melange versions support.

src/dune_rules/melange/melange.ml
```diff
@@ -2,13 +2,13 @@ open Import
 
 module Module_system = struct
   type t =
-    | Es6
+    | ESM
```
jchavarri, line 5: OCD note: why `ESM` but not `CommonJS` nor `ES6`? 😄 
```suggestion
    | Esm
```

jchavarri (approved):

src/dune_rules/melange/melange.ml
anmonteiro: I agree we should use `CommonJS` instead, but this page uses `ESM` all over the place https://nodejs.org/api/esm.html#modules-ecmascript-modules

rgrinberg (approved):
LGTM. Though it's not necessary to ask for my review for melange things unless there are specific questions for me.

PR #10278, opened 2024-03-16
refactor: remove pointless conversion in [Command.t]

rgrinberg:

### DIFF
--- a/src/dune_rules/command.ml
+++ b/src/dune_rules/command.ml
@@ -28,11 +28,14 @@ module Args0 = struct
   let dyn args = Dyn (Action_builder.map args ~f:(fun x -> As x))
   let empty = S []
 
-  let rec as_any : without_targets t -> any t = function
+  let as_any : without_targets t -> any t = function
     | A _ as x -> (x :> any t)
     | As _ as x -> (x :> any t)
-    | S l -> S (List.map l ~f:as_any)
-    | Concat (sep, l) -> Concat (sep, List.map l ~f:as_any)
+    (* We can't convince the type checker to do a cast, so we force it.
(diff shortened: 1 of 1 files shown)


src/dune_rules/command.ml
```diff
@@ -28,18 +28,24 @@ module Args0 = struct
...
+           | Dep _ as x -> (x :> any t)
+           | Deps _ as x -> (x :> any t)
+           | Path _ as x -> (x :> any t)
+           | Paths _ as x -> (x :> any t)
+           | Hidden_deps _ as x -> (x :> any t)
+           | Dyn _ as x -> (x :> any t)
+           | Expand _ as x -> (x :> any t)
+         ;;
+       ]} *)
+    Obj.magic
```
emillon, line 48: It's unfortunate that it's impossible to convince the typechecker that `_ t` is covariant (which would allow as_any to be just a coercion); but it seems dangerous to just `Obj.magic` our way out of this. Maybe we can do that just for the `Concat` and `S` cases?

Another suggestion is, can we use TRMC to do this? it looks like we could use this to convince the compiler not to allocate the new list.

emillon: (more precisely, by "dangerous" above I mean that it would be good to have this function break when a constructor is added)

rgrinberg: > Maybe we can do that just for the Concat and S cases?

We would still have a pattern match that has a runtime cost.

I did bring back the old implementation for compilation and exhaustive pattern matching purposes.

rgrinberg: Wasn't aware of that, thanks. If you inspect the generated code and check that it's equivalent, we can keep the pattern match.

PR #10280, opened 2024-03-17
test: `(modules_without_implementation)` expansion

anmonteiro:
- a bit of a follow-up to #9578, testing that `(modules_without_implementation)` is also expanded
- I needed this for `(modules_without_implementation)` so I decided to write a failing test. I was pleasantly surprised that it already worked!


nojb (approved):
LGTM, thanks!

Issue #10281, opened 2024-03-18
pkg: a package may be built before its depopts


gridbugs:
To illustrate the problem with an example, consider the package `cohttp-async` which depends on both 
the `logs` and `fmt` packages. The `logs` package has a depopt on `fmt` and `logs`'s build script checks whether the `fmt` package is installed and only builds the `logs.fmt` library if `fmt` is installed (specifically by evaluating `%{fmt:installed}%` in its opam file).

When building `cohttp-async` with dune it's possible for the `logs` package to be built before the `fmt` package, in which case the `logs.fmt` library won't be built because at the time that `logs` was built, the `fmt` package wasn't installed.

A simple way to fix this issue would be adding a package's available depopts to its `depends` field in its lockfile. By "available" I mean just add depopts that appear in the lockdir for the project.

PR #10282, opened 2024-03-18
pkg: add depopts to depends when available

gridbugs:
When a package's optional dependency is among the packages in the dependency solution for the project, we want that package's build take place after its optional dependency has been built. This allows the package's build commands to observe that its optional dependency is installed. This can be achieved by adding a package's optional dependency to its lockfile's `depends` field if the optional dependency is part of the dependency solution.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -510,6 +510,33 @@ let make_action = function
   | actions -> Some (Action.Progn actions)
 ;;
 
+(* Returns the set of depopts of a package which are part of the solution
+   represented by [version_by_package_name] *)
+let available_depopts solver_env version_by_package_name (opam_package : OpamFile.OPAM.t) =
+  let formula =
+    Resolve_opam_formula.apply_filter
+      (Solver_env.to_env solver_env)
+      ~with_test:false
+      opam_package.depopts
+  in
+  let atoms = OpamFormula.to_dnf formula |> List.concat in
+  List.filter_map atoms ~f:(fun (name, version_constraint) ->
+    let name = Package_name.of_opam_package_name name in
+    Package_name.Map.find version_by_package_name name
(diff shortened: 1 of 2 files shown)


gridbugs:
Fixes https://github.com/ocaml/dune/issues/10281. Raising now to get early feedback, but I intend to add tests if there are no major issues with my approach.

src/dune_pkg/opam_solver.ml
```diff
@@ -573,15 +601,25 @@ let opam_package_to_lock_file_pkg
...
         [ "package", Dyn.string (OpamFile.OPAM.package opam_file |> OpamPackage.to_string)
         ; "hints", Dyn.list Resolve_opam_formula.Unsatisfied_formula_hint.to_dyn hints
         ]
   in
+  let depopts = available_depopts solver_env version_by_package_name opam_file in
+  print_endline
+    (sprintf
+       "included depopts from package %S: %s"
+       (Package_name.to_string name)
+       (Package_name.Set.to_dyn depopts |> Dyn.to_string));
```
Leonidas-from-XIV, line 617: Leftover debug print (I guess you're aware of it, just to not forget it)

Leonidas-from-XIV (commented):
It seems like a sensible approach to me and a good catch, otherwise it could lead to fairly badly broken builds.

src/dune_pkg/opam_solver.ml
gridbugs: ah good catch

rgrinberg:
The approach looks right. This does need tests though.

Also there's some failures where existing dependencies are getting re-ordered. That seems suspicious

gridbugs:
I fixed the issue where dependencies would be re-ordered but I still need to add tests.

gridbugs:
Added tests

rgrinberg (approved):

PR #10283, opened 2024-03-18
feat(describe): support dialects in describe pp

emillon:


rgrinberg (approved):

test/blackbox-tests/test-cases/describe/describe-pp/dune
```diff
@@ -0,0 +1,3 @@
+(cram
+ (applies_to describe-pp)
+ (deps %{bin:refmt}))
```
Leonidas-from-XIV, line 3: Why does this test depend on `refmt`?

bin/describe/describe_pp.ml
```diff
@@ -29,6 +19,28 @@ let pp_with_ocamlc env ~ocamlc dialects pp_file =
...
 
+let files_for_source file dialects =
+  let base, ext = Path.split_extension file in
+  let dialect, kind =
+    match Dune_rules.Dialect.DB.find_by_extension dialects ext with
+    | None -> User_error.raise [ Pp.textf "unsupported extension: %s" ext ]
+    | Some x -> x
+  in
+  let new_ext =
+    ext ^ Option.value (Dune_rules.Dialect.ml_suffix dialect kind) ~default:""
```
Leonidas-from-XIV, line 30: I think for readability `sprintf "%s%s"` would be nicer. Or using `Path.extend_basename` which does exactly this under the hood, e.g. `Path.extend_basename file ~suffix:(Option.value ..)`.

test/blackbox-tests/test-cases/describe/describe-pp/describe-pp.t/run.t
```diff
@@ -12,3 +12,8 @@ We also make sure that the dump file is not present
 
   $ dune_cmd exists profile.dump
   true
+
+This also works for reason code
+
+  $ dune describe pp src/main_re.re
+  ;;Util.log "Hello, world!"
```
Leonidas-from-XIV, line 19: Since I was confused why this works (explanation: reason dialect is enabled by default, no need to enable it in `dune-project`) I think it would be nice to disable it and show that the dialect setting is being respected.

Leonidas-from-XIV (approved):
I think this is a reasonable change. I've added some suggestions, but these are fairly minor and can also be incorporated later.

test/blackbox-tests/test-cases/describe/describe-pp/describe-pp.t/run.t
emillon: yes. as discussed elsewhere, we'll make that more obvious in the docs.

anmonteiro, line 19: @emillon it would be nice to print the result with the dialect printer, if there's one.

PR #10284, opened 2024-03-18
Document that Reason dialect is supported out of the box

Leonidas-from-XIV:
Today I learned that we support Reason code out of the box with no additional configuration, which is fairly nice but doesn't appear to be documented so this PR is a suggestion to add a paragraph about that.

Its a fairly neat feature and I assume it could be used more if people were aware of how easy it is to use.

@emillon also suggested that a general howto on how to use reason in a dune project could be useful and that could be a good next step. It's fairly easy after all but if it is not documented, people won't know.

### DIFF
--- a/doc/reference/files/dune-project/dialect.rst
+++ b/doc/reference/files/dune-project/dialect.rst
@@ -65,3 +65,17 @@ dialect
       This field supports the same sub-fields as ``implementation``.
 
       .. versionchanged:: 3.9 This field is made optional.
+
+
+Default dialects
+----------------
+
+Dune ships with two dialects pre-configured and enabled:
+
+* ``ocaml`` for the default OCaml syntax which consumes `.ml` and `.mli` files
+  and uses ``ocamlformat`` for formatting.
+* ``reason`` for the Reason syntax and enabled in `.re`/`.rei` files. ``refmt``
+  is used for formatting.
+
+A third dialect, ``rescript``, is added when Melange support (see :doc:`/melange`)
+is enabled in the project.


doc/reference/files/dune-project/dialect.rst
```diff
@@ -65,3 +65,9 @@ dialect
       This field supports the same sub-fields as ``implementation``.
 
       .. versionchanged:: 3.9 This field is made optional.
+
+Dune comes with some dialects pre-set and enabled. The most common one is
```
emillon, line 69: Thanks. I think this should go into a new "Default Sialects" section.

Also, we're in a reference section so we should be precise: the default dialects are exactly `ocaml` and `reason`. In addition, if `(using melange)` is used, the `rescript` dialect is enabled (I just learned this!). Maybe this should be mentioned.

Leonidas-from-XIV: Good point, I've added these. I think switching to an enumeration of dialects works better than a freeform paragraph as it was before.

doc/reference/files/dune-project/dialect.rst
```diff
@@ -65,3 +65,20 @@ dialect
...
+* ``ocaml`` for the default OCaml syntax which consumes `.ml` and `.mli` files
+  and uses ``ocamlformat`` for formatting.
+* ``reason`` for the Reason syntax and enabled in `.re`/`.rei` files. For
+  formatting ``refmt`` is used.
+
+When ``(using melange)`` is enabled in the `dune-project` a third dialect is
+enabled:
+
+* ``rescript`` for the ReScript syntax which is used in `.res` and `.resi`
+  files. The formatter in this case is ``rescript_syntax``.
```
davesnx, line 84: That's not true right? rescript_syntax is only available for melange 1.0.0

davesnx, line 84: This should be specified, rescript_syntax only works in Melange v1 and it doesn't support the entire rescript language (as newer versions) in any form in dune

Leonidas-from-XIV: As someone who is a bit out of touch with Rescript and Melange, does that mean that `rescript_syntax` can only be used in Melange v1 (and not v2 or v3) and even then only supports an older subset of the language?

I don't know to which extent Melange v1 + ReScript are used but this sounds like a serious restriction and makes me wonder if supporting something of progressively lower relevance as time goes on and Rescript evolves makes sense overall.

davesnx: Exactly, rescript_syntax (https://opam.ocaml.org/packages/rescript-syntax/) was a tool to help people migrate from ReScript to Melange. It's not a feature dune should keep up, since rescript-syntax isn't up-to-date with https://github.com/rescript-lang/rescript-compiler in any form.

doc/reference/files/dune-project/dialect.rst
```diff
@@ -65,3 +65,17 @@ dialect
       This field supports the same sub-fields as ``implementation``.
 
       .. versionchanged:: 3.9 This field is made optional.
+
+
+Default dialects
+----------------
+
+In the default configuration Dune comes with two dialects pre-set and enabled:
```
anmonteiro, line 73:
```suggestion
Dune ships with two dialects pre-configured and enabled:
```

doc/reference/files/dune-project/dialect.rst
```diff
@@ -65,3 +65,17 @@ dialect
...
+
+* ``ocaml`` for the default OCaml syntax which consumes `.ml` and `.mli` files
+  and uses ``ocamlformat`` for formatting.
+* ``reason`` for the Reason syntax and enabled in `.re`/`.rei` files. For
+  formatting ``refmt`` is used.
```
anmonteiro, line 78:
```suggestion
* ``reason`` for the Reason syntax and enabled in `.re`/`.rei` files. ``refmt``
  is used for formatting.
```

doc/reference/files/dune-project/dialect.rst
```diff
@@ -65,3 +65,17 @@ dialect
...
+
+A third dialect called ``rescript`` is added when Melange (see :doc:`/melange`)
```
anmonteiro, line 80:
```suggestion
A third dialect, ``rescript``, is added when Melange support (see :doc:`/melange`)
```

anmonteiro (commented):
optional language tweaks

anmonteiro (approved):

PR #10288, opened 2024-03-20
test(melange): add `(emit_stdlib false)` to supported melange tests

anmonteiro:
- `(emit_stdlib false)` disables emitting all the JS targets for the melange stdlib
- this should speed up the tests quite a bit (tens of seconds)


anmonteiro:
This is around 15-20s faster on linux. hard to tell on macOS

jchavarri:
Do you know if there's a way to set this at the env variable level? So we can set it once and any future tests added (where we/I might forget about adding `emit_stdlib false`) don't run into the same performance costs.

anmonteiro:
I'm not sure. I also don't know whether we'd want this to be the default in tests, but probably yes? 🤔

rgrinberg (approved):

PR #10289, opened 2024-03-20
fix: symlink `.impl.d` files for virtual library modules too

anmonteiro:
- when setting up rules for implementation files in a virtual library, their artifacts are symlinked to the object directory for the concrete implementation too
- this change adds rules to symlink the ocamldep resulting files too, as they may be needed in the object directory
- specifically, this fixes the failing test in https://github.com/ocaml/dune/pull/10286 as will be demonstrated once I cherry-pick this commit in that PR.


rgrinberg:
You say that they may be needed, but more concretely this is needed for melange only? I don't think this is needed anywhere else because I just merge the dependency graphs for vlibs and impls everywhere else and use that.

rgrinberg:
Okay, then I'd just include this as a commit in #10286. If it doesn't fix anything on its own, a separate PR isn't really needed.

anmonteiro:
Sounds good, will incorporate it there.

PR #10292, opened 2024-03-20
chore: remove unused plugin/dune file

emillon:
The `jbuild_plugin.*` are never used as a library, they are used in a `(mode promote)` rule in `dune_rules`.


rgrinberg:
It's build as a library to provide merlin support and syntax checking when we do `dune build @check`

emillon:
makes sense, thanks

PR #10295, opened 2024-03-20
test(melange): missing dependency during emission with sandbox enabled

anmonteiro:
based on the suggestion in https://github.com/ocaml/dune/pull/10286#issuecomment-2010776851

enabling sandboxing confirms there's a missing dependency

### DIFF
--- a/test/blackbox-tests/test-cases/melange/melange-sandboxing-deps.t
+++ b/test/blackbox-tests/test-cases/melange/melange-sandboxing-deps.t
@@ -30,7 +30,4 @@ Show missing dependency in Melange when sandboxing is enabled
   > EOF
 
   $ dune build @mel
-  File "_none_", line 1:
-  Error: Foo__Bar not found, it means either the module does not exist or it is a namespace
-  [1]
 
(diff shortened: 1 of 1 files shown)


anmonteiro:
not sure if this is needed anymore since #10286 fixed it.

anmonteiro:
> The only cost is in performance, so it could really be enabled everywhere else in the test suite as well.

I don't understand the implication here. This PR has a performance impact? But enabling more sandboxing will have less perf impact?

anmonteiro:
IIRC there's also something preventing us from enabling sandboxing altogether for `melange.emit` because of the runtime_deps stuff?

rgrinberg:
Sure, let me clarify. Sandboxing has a performance cost, but it makes the rules more reliable. For the test suite, that seems like a good trade-off because we are building trivial projects anyway.

Sandboxing works with runtime deps if the dependencies are specified correctly. I don't see any specific issues there.

I suppose one issue with sandboxing is that rules can often rely on it in a way that will make them break in non-sandboxed mode. E.g. `-I` in sandboxed mode is guaranteed only to observe the `.cmi` files you depend on. But in non sandboxed mode, it's more loose, so you need to be careful in your compiler.

anmonteiro:
I'd love to run the test suite through sandboxing and see if it holds up for Melange. Is there a way to enable sandboxing for the entire test suite other than e.g. setting the environment variable in the dune file?

rgrinberg:
Maybe you could use the setenv field in the env stanza to turn it on for melange tests? `(setenv DUNE_SANDBOX symlink)`

rgrinberg (approved):

anmonteiro:
I'd rather merge #10312. Just need to disable that odoc test.

PR #10297, opened 2024-03-22
fix(melange): account for preprocessing when getting library's Modules.t during emission

anmonteiro:
- this is a regression introduced in #10286, caused by reading `immediate_deps_of` the current cmj file
- ~`Ocamldep.read_immediate_deps_of` needs to account for the fact that the file may have been pre-processed by a dialect, and use the final filename instead.~
- we need to account for preprocessing (which changes the module target filenames in the build directory) when getting the `Modules.t` structure where we consult `.impl.d` files
- this diff is best reviewed [without whitespace](https://github.com/ocaml/dune/pull/10297/files?w=1)

the error can be seen in the first commit:

```
  Error: ocamldep returned unexpected output for _build/default/lib/foo.myd:
  > lib/foo.myd.ml: Bar
  -> required by _build/default/output/lib/foo.js
  -> required by alias mel
  [1]
```

### DIFF
--- a/src/dune_rules/melange/melange_rules.ml
+++ b/src/dune_rules/melange/melange_rules.ml
@@ -40,8 +40,21 @@ let make_js_name ~js_ext ~output m =
     Path.Build.relative dst_dir basename
 ;;
 
-let impl_only_modules_defined_in_this_lib sctx lib =
-  let+ modules = Dir_contents.modules_of_lib sctx lib in
+let modules_in_obj_dir ~sctx ~scope ~preprocess modules =
+  let* ocaml = Context.ocaml (Super_context.context sctx) in
+  let version = ocaml.version in
+  let* preprocess =
+    Resolve.Memo.read_memo
+      (Preprocess.Per_module.with_instrumentation
+         preprocess
+         ~instrumentation_backend:(Lib.DB.instrumentation_backend (Scope.libs scope)))
+  in
+  let pped_map = Staged.unstage (Preprocessing.pped_modules_map preprocess version) in
+  Modules.map_user_written modules ~f:(fun m -> Memo.return @@ pped_map m)
+;;
+
+let impl_only_modules_defined_in_this_lib ~sctx ~scope lib =
+  let* modules = Dir_contents.modules_of_lib sctx lib in
   match modules with
   | None ->
     User_error.raise
@@ -52,8 +65,12 @@ let impl_only_modules_defined_in_this_lib sctx lib =
           (Lib.name lib |> Lib_name.to_string)
       ]
   | Some modules ->
+    let info = Lib.info lib in
+    let+ modules =
+      let preprocess = Lib_info.preprocess info in
+      modules_in_obj_dir ~sctx ~scope ~preprocess modules
+    in
     let () =
-      let info = Lib.info lib in
       let modes = Lib_info.modes info in
       match modes.melange with
       | false ->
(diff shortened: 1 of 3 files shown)


anmonteiro:
I tested this in a real world project impacted by the regression and it fixed the issue too.

src/dune_rules/melange/melange_rules.ml
```diff
@@ -133,11 +152,11 @@ let js_targets_of_modules modules ~module_systems ~output =
   |> Path.Set.union_all
 ;;
 
-let js_targets_of_libs sctx libs ~module_systems ~target_dir =
+let js_targets_of_libs ~sctx ~scope ~module_systems ~target_dir libs =
   Resolve.Memo.List.concat_map module_systems ~f:(fun (_, js_ext) ->
     let open Memo.O in
     let of_lib lib =
-      let+ modules = impl_only_modules_defined_in_this_lib sctx lib in
+      let+ _, modules = impl_only_modules_defined_in_this_lib ~sctx ~scope lib in
```
jchavarri, line 159: As in this case the first part of the tuple is unused, would it make sense to just return the `Modules.t` from `impl_only_modules_defined_in_this_lib` (unprocessed), and then call `modules_in_obj_dir` manually when it's needed in `setup_js_rules_libraries`? Just wondering if that work can be skipped when possible.

src/dune_rules/melange/melange_rules.ml
```diff
@@ -195,17 +214,24 @@ let build_js
...
             let+ module_deps =
               Dep_rules.immediate_deps_of m modules ~obj_dir ~ml_kind:Impl
             in
-            List.map module_deps ~f:(fun dep_m ->
-              Obj_dir.Module.cm_file_exn obj_dir dep_m ~kind:(Melange Cmj) |> Path.build)
-          | None -> Action_builder.return []
-        in
-        Action_builder.dyn_paths_unit local_library_paths >>> command)
+            List.fold_left module_deps ~init:[] ~f:(fun acc dep_m ->
+              if Module.has dep_m ~ml_kind:Impl
```
jchavarri, line 224: Why is this check necessary if `immediate_deps_of` is being called with `~ml_kind:Impl`?

src/dune_rules/melange/melange_rules.ml
```diff
@@ -403,18 +421,7 @@ let modules_for_js_and_obj_dir ~sctx ~dir_contents ~scope (mel : Melange_stanzas
...
-    let* preprocess =
-      Resolve.Memo.read_memo
-        (Preprocess.Per_module.with_instrumentation
-           mel.preprocess
-           ~instrumentation_backend:(Lib.DB.instrumentation_backend (Scope.libs scope)))
-    in
-    let pped_map = Staged.unstage (Preprocessing.pped_modules_map preprocess version) in
-    Modules.map_user_written modules ~f:(fun m -> Memo.return @@ pped_map m)
-  in
+  let+ modules = modules_in_obj_dir ~sctx ~scope ~preprocess:mel.preprocess modules in
```
jchavarri, line 432: I don't understand why do we account for `mel.preprocess` here and not `Lib_info.preprocess info` like above?

jchavarri (approved):

src/dune_rules/melange/melange_rules.ml
anmonteiro: This handles melange.emit and the other handles (library)

src/dune_rules/melange/melange_rules.ml
anmonteiro: Because the immediate dependencies could be intf-only modules which don’t have cmj

src/dune_rules/melange/melange_rules.ml
anmonteiro: Work wouldn’t be skipped. Modules.t is used to compute the Module.t list.

PR #10298, opened 2024-03-22
test: add eif library cycle test

jchavarri:
As mentioned in https://github.com/ocaml/dune/pull/10296#discussion_r1534710211.


rgrinberg (approved):
LGTM. Notice how the cycle reported here is completely spurious should be fixed rather than reinforced.

PR #10299, opened 2024-03-22
doc: remove extra files/ layer and use sections

emillon:
The extra layer nests the pages too deeply and requires extra clicks. Instead
we put this directly in the reference hierarchy and highlight them in a
separate section.


christinerose (approved):

doc/reference/foreign.rst
```diff
@@ -154,7 +153,7 @@ The ``archive_name`` field specifies the archive's name. You can refer
 to the same archive name from multiple OCaml libraries and executables, so a
 foreign archive is a bit like a foreign library, hence the name of the stanza.
 The ``enabled_if`` field has the same meaning as in the
-:doc:`/reference/files/dune/library` stanza.
```
Leonidas-from-XIV, line 157: Maybe make this in into a relative path, to match the other relative paths.

Leonidas-from-XIV (approved):
Some minor comments; however none affect the rendering of the document.

PR #10302, opened 2024-03-22
refactor: invert control in install_uninstall

emillon:
Instead of returning a `load_special_file_result`, we change the API so that functions get a `~get_version` and `oc` argument and use them.


anmonteiro (approved):
LGTM. Nice cleanup

PR #10307, opened 2024-03-26
feat: support libraries with the same name in multiple contexts

anmonteiro:
this is #10179 + my commits

### DIFF
--- a/src/dune_rules/dune_package.ml
+++ b/src/dune_rules/dune_package.ml
@@ -231,9 +231,7 @@ module Lib = struct
        let entry_modules = Modules.entry_modules modules |> List.map ~f:Module.name in
        let info : Path.t Lib_info.t =
          let src_dir = Obj_dir.dir obj_dir in
-         let sentinel =
-           Lib_info.Sentinel.external_ ~loc ~src_dir ~enabled_if:Blang.true_ name
-         in
+         let sentinel = Lib_info.Sentinel.external_ ~loc ~src_dir name in
          let enabled = Memo.return Lib_info.Enabled_status.Normal in
          let status =
            match Lib_name.analyze name with
--- a/src/dune_rules/dune_package.ml
+++ b/src/dune_rules/dune_package.ml
@@ -231,6 +231,9 @@ module Lib = struct
        let entry_modules = Modules.entry_modules modules |> List.map ~f:Module.name in
        let info : Path.t Lib_info.t =
          let src_dir = Obj_dir.dir obj_dir in
+         let sentinel =
+           Lib_info.Sentinel.external_ ~loc ~src_dir ~enabled_if:Blang.true_ name
+         in
          let enabled = Memo.return Lib_info.Enabled_status.Normal in
          let status =
            match Lib_name.analyze name with
@@ -255,6 +258,7 @@ module Lib = struct
            ~path_kind:External
            ~loc
            ~name
+           ~sentinel
            ~kind
            ~status
            ~src_dir
@@ -361,12 +365,12 @@ end
 module Entry = struct
   type t =
     | Library of Lib.t
-    | Deprecated_library_name of Deprecated_library_name.t
+    | Deprecated_library_name of Path.t * Deprecated_library_name.t
     | Hidden_library of Lib.t
 
   let name = function
     | Library lib | Hidden_library lib -> Lib_info.name (Lib.info lib)
-    | Deprecated_library_name d -> d.old_public_name
+    | Deprecated_library_name (_, d) -> d.old_public_name
   ;;
 
   let version = function
@@ -376,7 +380,14 @@ module Entry = struct
 
   let loc = function
     | Library lib | Hidden_library lib -> Lib_info.loc (Lib.info lib)
-    | Deprecated_library_name d -> d.loc
+    | Deprecated_library_name (_, d) -> d.loc
+  ;;
+
+  let sentinel = function
+    | Library lib | Hidden_library lib ->
+      let info = Lib.info lib in
+      Lib_info.sentinel info
+    | Deprecated_library_name _ -> assert false
   ;;
 
   let cstrs ~lang ~dir =
@@ -386,7 +397,7 @@ module Entry = struct
         Library lib )
     ; ( "deprecated_library_name"
       , let+ x = Deprecated_library_name.decode in
-        Deprecated_library_name x )
+        Deprecated_library_name (dir, x) )
     ]
   ;;
 
@@ -394,7 +405,7 @@ module Entry = struct
     let open Dyn in
     match x with
     | Library lib -> variant "Library" [ Lib.to_dyn lib ]
-    | Deprecated_library_name lib ->
+    | Deprecated_library_name (_, lib) ->
       variant "Deprecated_library_name" [ Deprecated_library_name.to_dyn lib ]
     | Hidden_library lib -> variant "Hidden_library" [ Lib.to_dyn lib ]
   ;;
@@ -536,7 +547,7 @@ let encode ~encoding ~dune_version { entries; name; version; dir; sections; site
       match e with
       | Entry.Library lib ->
         list (Dune_lang.atom "library" :: Lib.encode lib ~package_root:dir ~stublibs)
-      | Deprecated_library_name d ->
+      | Deprecated_library_name (_, d) ->
         list (Dune_lang.atom "deprecated_library_name" :: Deprecated_library_name.encode d)
       | Hidden_library lib ->
         Code_error.raise
--- a/src/dune_rules/dune_package.ml
+++ b/src/dune_rules/dune_package.ml
@@ -365,12 +365,12 @@ end
 module Entry = struct
   type t =
     | Library of Lib.t
-    | Deprecated_library_name of Path.t * Deprecated_library_name.t
+    | Deprecated_library_name of Deprecated_library_name.t
     | Hidden_library of Lib.t
 
   let name = function
     | Library lib | Hidden_library lib -> Lib_info.name (Lib.info lib)
-    | Deprecated_library_name (_, d) -> d.old_public_name
+    | Deprecated_library_name d -> d.old_public_name
   ;;
 
   let version = function
@@ -380,7 +380,7 @@ module Entry = struct
 
   let loc = function
     | Library lib | Hidden_library lib -> Lib_info.loc (Lib.info lib)
-    | Deprecated_library_name (_, d) -> d.loc
+    | Deprecated_library_name d -> d.loc
   ;;
 
   let sentinel = function
@@ -397,7 +397,7 @@ module Entry = struct
         Library lib )
     ; ( "deprecated_library_name"
       , let+ x = Deprecated_library_name.decode in
-        Deprecated_library_name (dir, x) )
+        Deprecated_library_name x )
     ]
   ;;
 
@@ -405,7 +405,7 @@ module Entry = struct
     let open Dyn in
     match x with
     | Library lib -> variant "Library" [ Lib.to_dyn lib ]
-    | Deprecated_library_name (_, lib) ->
+    | Deprecated_library_name lib ->
       variant "Deprecated_library_name" [ Deprecated_library_name.to_dyn lib ]
     | Hidden_library lib -> variant "Hidden_library" [ Lib.to_dyn lib ]
   ;;
@@ -547,7 +547,7 @@ let encode ~encoding ~dune_version { entries; name; version; dir; sections; site
       match e with
       | Entry.Library lib ->
         list (Dune_lang.atom "library" :: Lib.encode lib ~package_root:dir ~stublibs)
-      | Deprecated_library_name (_, d) ->
+      | Deprecated_library_name d ->
         list (Dune_lang.atom "deprecated_library_name" :: Deprecated_library_name.encode d)
       | Hidden_library lib ->
         Code_error.raise
--- a/src/dune_rules/dune_package.ml
+++ b/src/dune_rules/dune_package.ml
@@ -382,9 +382,7 @@ module Entry = struct
   ;;
 
   let library_id = function
-    | Library lib | Hidden_library lib ->
-      let info = Lib.info lib in
-      Lib_info.library_id info
+    | Library lib | Hidden_library lib -> Lib_info.library_id (Lib.info lib)
     | Deprecated_library_name _ -> assert false
   ;;
 
--- a/src/dune_rules/dune_package.ml
+++ b/src/dune_rules/dune_package.ml
@@ -381,11 +381,6 @@ module Entry = struct
     | Deprecated_library_name d -> d.loc
   ;;
 
-  let library_id = function
-    | Library lib | Hidden_library lib -> Lib_info.library_id (Lib.info lib)
-    | Deprecated_library_name _ -> assert false
-  ;;
-
   let cstrs ~lang ~dir =
     let open Dune_lang.Decoder in
     [ ( "library"
--- a/src/dune_rules/dune_package.ml
+++ b/src/dune_rules/dune_package.ml
@@ -231,7 +231,7 @@ module Lib = struct
        let entry_modules = Modules.entry_modules modules |> List.map ~f:Module.name in
        let info : Path.t Lib_info.t =
          let src_dir = Obj_dir.dir obj_dir in
-         let lib_id = Lib_id.external_ ~loc ~src_dir name in
+         let lib_id = Lib_id.External (loc, name) in
          let enabled = Memo.return Lib_info.Enabled_status.Normal in
          let status =
            match Lib_name.analyze name with
--- a/src/dune_rules/dune_package.ml
+++ b/src/dune_rules/dune_package.ml
@@ -231,7 +231,7 @@ module Lib = struct
        let entry_modules = Modules.entry_modules modules |> List.map ~f:Module.name in
        let info : Path.t Lib_info.t =
          let src_dir = Obj_dir.dir obj_dir in
-         let library_id = Lib_info.Library_id.external_ ~loc ~src_dir name in
+         let lib_id = Lib_id.external_ ~loc ~src_dir name in
          let enabled = Memo.return Lib_info.Enabled_status.Normal in
          let status =
            match Lib_name.analyze name with
@@ -256,7 +256,7 @@ module Lib = struct
            ~path_kind:External
            ~loc
            ~name
-           ~library_id
+           ~lib_id
            ~kind
            ~status
            ~src_dir
--- a/src/dune_rules/dune_package.ml
+++ b/src/dune_rules/dune_package.ml
@@ -231,7 +231,7 @@ module Lib = struct
        let entry_modules = Modules.entry_modules modules |> List.map ~f:Module.name in
        let info : Path.t Lib_info.t =
          let src_dir = Obj_dir.dir obj_dir in
-         let sentinel = Lib_info.Sentinel.external_ ~loc ~src_dir name in
+         let library_id = Lib_info.Library_id.external_ ~loc ~src_dir name in
          let enabled = Memo.return Lib_info.Enabled_status.Normal in
          let status =
            match Lib_name.analyze name with
@@ -256,7 +256,7 @@ module Lib = struct
            ~path_kind:External
            ~loc
            ~name
-           ~sentinel
+           ~library_id
            ~kind
            ~status
            ~src_dir
@@ -381,10 +381,10 @@ module Entry = struct
     | Deprecated_library_name d -> d.loc
   ;;
 
-  let sentinel = function
+  let library_id = function
     | Library lib | Hidden_library lib ->
       let info = Lib.info lib in
-      Lib_info.sentinel info
+      Lib_info.library_id info
     | Deprecated_library_name _ -> assert false
   ;;
 
--- a/src/dune_rules/findlib.ml
+++ b/src/dune_rules/findlib.ml
@@ -207,12 +207,12 @@ let to_dune_library (t : Findlib.Package.t) ~dir_contents ~ext_lib ~external_loc
     in
     let modules = Lib_info.Source.External None in
     let name = t.name in
-    let library_id = Lib_info.Library_id.external_ ~loc ~src_dir name in
+    let lib_id = Lib_id.external_ ~loc ~src_dir name in
     Lib_info.create
       ~loc
       ~path_kind:External
       ~name
-      ~library_id
+      ~lib_id
       ~kind
       ~status
       ~src_dir
--- a/src/dune_rules/findlib.ml
+++ b/src/dune_rules/findlib.ml
@@ -207,7 +207,7 @@ let to_dune_library (t : Findlib.Package.t) ~dir_contents ~ext_lib ~external_loc
     in
     let modules = Lib_info.Source.External None in
     let name = t.name in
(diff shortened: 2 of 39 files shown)


anmonteiro:
@rgrinberg @jchavarri let's continue here. This is:

- a do-over of #10179 and jchavarri/dune#12 after:
    - renaming "library ID" to "Lib_info.Sentinel.t"
    - putting back the old signatures in `Lib.DB`
    - minimizing the diff
    - auditing whether we should be querying the DB by name or sentinel
    - fixing remaining tests

Should be ready for a serious round of review now.

src/dune_rules/scope.ml
```diff
@@ -26,158 +26,320 @@ module DB = struct
...
+  let resolve_name =
+    let module Resolve_result = Lib.DB.Resolve_result in
+    let module With_multiple_results = Resolve_result.With_multiple_results in
+    fun ~resolve_sentinel id_map name ->
+      match
+        Lib_name.Map.find id_map name |> Option.map ~f:Lib_info.Sentinel.Set.to_list
+      with
+      | None ->
+        Memo.return (With_multiple_results.resolve_result Resolve_result.not_found)
+      | Some [] -> assert false
```
jchavarri, line 78: I think there's something like `Code_error` that could be used here instead of plain assert?

src/dune_rules/dune_package.ml
```diff
@@ -373,7 +377,14 @@ module Entry = struct
...
     | Library lib | Hidden_library lib -> Lib_info.loc (Lib.info lib)
-    | Deprecated_library_name d -> d.loc
+    | Deprecated_library_name (_, d) -> d.loc
+  ;;
+
+  let sentinel = function
+    | Library lib | Hidden_library lib ->
+      let info = Lib.info lib in
+      Lib_info.sentinel info
+    | Deprecated_library_name _ -> assert false
```
jchavarri, line 387: Why is this impossible?

src/dune_rules/lib_info.mli
```diff
@@ -86,9 +86,25 @@ module Main_module_name : sig
   type t = Module_name.t option Inherited.t
 end
 
+(** What's the subset of fields that uniquely identifies this stanza? *)
+module Sentinel : sig
+  type t
```
rgrinberg, line 91: What meaning is the name Sentinel is desired to convey here?

src/dune_rules/lib.mli
```diff
@@ -113,23 +126,27 @@ module DB : sig
...
     -> unit
     -> t
 
   val find : t -> Lib_name.t -> lib option Memo.t
   val find_even_when_hidden : t -> Lib_name.t -> lib option Memo.t
-  val available : t -> Lib_name.t -> bool Memo.t
+  val find_sentinel : t -> Lib_info.Sentinel.t -> lib option Memo.t
+  val find_sentinel_even_when_hidden : t -> Lib_info.Sentinel.t -> lib option Memo.t
+  val available : t -> Lib_info.Sentinel.t -> bool Memo.t
+  val available_by_name : t -> Lib_name.t -> bool Memo.t
```
rgrinberg, line 142: You can flip these definitions around and preserve more existing code. E.g. `available` and `available_by_sentinel`.

src/dune_rules/lib.ml
```diff
@@ -1116,7 +1166,92 @@ end = struct
   ;;
 
   let instantiate db name info ~hidden = (Lazy.force db.instantiate) name info ~hidden
-  let find_internal db (name : Lib_name.t) = resolve_name db name
+
+  let resolve_hidden db ~info hidden =
+    let open Memo.O in
+    (match db.parent with
+     | None -> Memo.return Status.Not_found
```
jchavarri, line 1173: ~Shouldn't we return `Hidden` here?~ Edit: i see now that this is how existing code worked 🤷

jchavarri (commented):
Thanks a lot for your work on this! ❤️ 

I left a few questions, the most relevant one being in `lib-collision-public-same-folder.t`. 

If you want me to work on any of this, please lmk!

test/blackbox-tests/test-cases/lib-collision/lib-collision-public-same-folder.t
```diff
@@ -19,9 +19,9 @@ the same folder.
 Without any consumers of the libraries
 
   $ dune build
-  Error: Library foo is defined twice:
-  - dune:6
-  - dune:3
+  Error: Multiple rules generated for _build/default/foo.cmxs:
```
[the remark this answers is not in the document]
jchavarri: Attempt to fix in https://github.com/anmonteiro/dune/pull/10 (still requires the introduction of `Library.private_name`, I am not able to see a way around this).

jchavarri:
In https://github.com/ocaml/dune/pull/10307/commits/86bd2453c7e85366e9750358bf4dff3f2cb9230d I removed some changes that were copied over from early work, but are not necessary. The diff in `dune_package.ml(i)` is now cleaner and the total diff size of the PR went down a bit (from `+935 −305` to `+921 −292`).

src/dune_rules/lib_info.mli
anmonteiro: I called it "sentinel" since this is effectively something that can tell different libraries apart, and uniquely identifies them. in other words, it acts as an indicator of this being the library we're looking for without ambiguity.

test/blackbox-tests/test-cases/lib-collision/lib-collision-public-same-public-name.t
```diff
@@ -0,0 +1,47 @@
...
+  > let () = Foo.x
+  > EOF
+
+  $ dune build
+  File "a/dune", line 1, characters 0-44:
+  1 | (library
+  2 |  (name foo)
+  3 |  (public_name bar.foo))
+  Error: Library with name "foo" is defined in two folders (_build/default/b
+  and _build/default/a) and shares a name with library "bar". Either change one
```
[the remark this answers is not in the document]
anmonteiro: well, the libraries are defined in 2 folders, and share a name. that's what I tried to convey in the new error message. the shared name is `bar.foo` which we don't surface, but that doesn't make the error message incorrect

src/dune_rules/lib_info.mli
rgrinberg: What's the difference between an id and a sentinel then?

anmonteiro: no difference. I can change it back if you prefer.

rgrinberg: I find Id the clearer name, and it seems like other people agree. I'd prefer a name people are more familiar with.

anmonteiro: absolutely. just changed it back to `library_id`.

test/blackbox-tests/test-cases/lib-collision/lib-collision-public-same-folder.t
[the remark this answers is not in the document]
anmonteiro: I think we'll want @rgrinberg to review those changes. I don't know exactly what semantics we want here.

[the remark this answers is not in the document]
rgrinberg: I think the old error was good. The user clearly made a mistake here.

test/blackbox-tests/test-cases/lib-collision/lib-collision-public-same-public-name.t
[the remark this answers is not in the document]
rgrinberg: I agree the message could be rewarded. Perhaps something like:

> This library name "bar.foo" is already defined in b/dune:3

test/blackbox-tests/test-cases/lib-collision/lib-collision-public-same-folder.t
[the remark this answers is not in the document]
anmonteiro: what do you think about comparing by obj_dir instead of exposing the private name (https://github.com/anmonteiro/dune/pull/10#discussion_r1540119767)? I guess that wouldn't cover private names in different directories.

test/blackbox-tests/test-cases/lib-collision/lib-collision-public-same-public-name.t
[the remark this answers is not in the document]
jchavarri: > but that doesn't make the error message incorrect

I got confused by this part:

> Library with name "foo" is defined in two folders (_build/default/b and _build/default/a)

I saw no library "foo" defined in those two folders, but I was looking at the private names. I guess the source of my confusion comes from never looking at the second part of the public lib name (after the dot) as the library name, but rather the private one.

test/blackbox-tests/test-cases/lib-collision/lib-collision-public-same-folder.t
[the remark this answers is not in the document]
jchavarri: https://github.com/anmonteiro/dune/pull/10/files#r1540920892

src/dune_rules/lib.ml
```diff
@@ -396,10 +431,11 @@ end
 
 type db =
   { parent : db option
-  ; resolve : Lib_name.t -> resolve_result Memo.t
+  ; resolve : Lib_name.t -> resolve_result_with_multiple_results Memo.t
+  ; resolve_library_id : Lib_info.Library_id.t -> resolve_result Memo.t
   ; instantiate :
       (Lib_name.t -> Path.t Lib_info.t -> hidden:string option -> Status.t Memo.t) Lazy.t
-  ; all : Lib_name.t list Memo.Lazy.t
+  ; all : Lib_info.Library_id.t list Memo.Lazy.t
```
rgrinberg, line 438: Is this a correct use of id's? we aren't supposed to list disabled libraries here

anmonteiro: thanks for catching that. I changed it back to return a Lib_name.t list

rgrinberg:
There's still something fishy about we're handling `Ml_sources.t`. E.g. consider this:
```
diff --git a/test/blackbox-tests/test-cases/lib-collision/lib-collision-private-same-folder.t b/test/blackbox-tests/test-cases/lib-collision/lib-collision-private-same-folder.t
index 9c7391ac9..706869002 100644
--- a/test/blackbox-tests/test-cases/lib-collision/lib-collision-private-same-folder.t
+++ b/test/blackbox-tests/test-cases/lib-collision/lib-collision-private-same-folder.t
@@ -7,7 +7,9 @@ the same folder.
 
   $ cat > dune << EOF
   > (library
-  >  (name foo))
+  >  (name foo)
+  >  (libraries xxx)
+  >  (optional))
   > (library
   >  (name foo))
   > EOF
@@ -15,9 +17,9 @@ the same folder.
 Without any consumers of the libraries
 
   $ dune build
-  File "dune", line 3, characters 0-21:
-  3 | (library
-  4 |  (name foo))
+  File "dune", line 5, characters 0-21:
+  5 | (library
+  6 |  (name foo))
   Error: Library "foo" appears for the second time in this directory
   [1]
```

I would imagine that this should work. Not necessarily blocking this PR, but this code still needs tuning.

src/dune_rules/lib_info.mli
```diff
@@ -86,9 +86,25 @@ module Main_module_name : sig
   type t = Module_name.t option Inherited.t
 end
 
+(** What's the subset of fields that uniquely identifies this stanza? *)
```
rgrinberg, line 89: Could you move this to its own file? Also `Lib_id` sounds just as good and shorter to me.

src/dune_rules/scope.ml
```diff
@@ -26,158 +26,307 @@ module DB = struct
...
+        let+ enabled =
+          let+ toggle = Memo.Lazy.force enabled in
+          Toggle.enabled toggle
+        in
+        if enabled
+        then Lib.DB.Resolve_result.redirect_in_the_same_db (loc, to_)
+        else Lib.DB.Resolve_result.not_found
+      | Some (Found lib) -> Memo.return (Lib.DB.Resolve_result.found lib)
+      | Some (Deprecated_library_name lib) ->
+        Memo.return (Lib.DB.Resolve_result.redirect_in_the_same_db lib)
```
rgrinberg, line 209: Now that I think about it, this seems wrong. We should be redirecting the `new_public_name` into the public lib DB rather then the same db. This isn't an issue introduced by this PR though.

src/dune_rules/scope.ml
```diff
@@ -26,158 +26,307 @@ module DB = struct
...
+                  , Name s.new_public_name
+                  , Deprecated_library_name.to_library_id ~src_dir s )
+            in
+            match candidate with
+            | None -> libname_map, id_map, library_id_map
+            | Some (public_name, r2, library_id) ->
+              let libname_map' =
+                Lib_name.Map.update libname_map public_name ~f:(function
+                  | None -> Some (library_id, r2)
+                  | Some (sent1, _r1) ->
```
rgrinberg, line 274: `sent1` should be renamed.

src/dune_rules/stanzas/deprecated_library_name.ml
```diff
@@ -47,3 +47,10 @@ let decode =
      in
      { Library_redirect.loc; project; old_name; new_public_name })
 ;;
+
+let to_library_id ~src_dir (t : t) =
+  let lib, _ = t.old_name in
+  let loc = Public_lib.loc lib
+  and enabled_if = Blang.true_ in
```
rgrinberg, line 54: This seems fishy, how can a deprecated library name have an id?

test/blackbox-tests/test-cases/lib-collision/lib-collision-private.t
```diff
@@ -17,14 +17,19 @@ different folders.
...
-
-  $ dune build
-  Error: Library foo is defined twice:
-  - a/dune:1
-  - b/dune:1
-  -> required by alias default
-  [1]
+Without any consumers of the libraries (both are built in separate folders)
+
+  $ dune build --display short
```
rgrinberg, line 22: This can just be `dune build a/foo.cma b/foo.cma`

src/dune_rules/findlib.ml
```diff
@@ -206,10 +206,13 @@ let to_dune_library (t : Findlib.Package.t) ~dir_contents ~ext_lib ~external_loc
                     | Error e -> Error e))))
     in
     let modules = Lib_info.Source.External None in
+    let name = t.name in
+    let library_id = Lib_info.Library_id.external_ ~loc ~src_dir name in
```
rgrinberg, line 210: I think difference for id's of external vs local libraries is that external id's really are identified by their name. That's because we never keep searching for external libraries after we find the first name. So I think the id type should be more like:

```
type t =
  | External of Lib_name.t
  | Local of what_is_library_id_now
```

rgrinberg: In fact, something like:

```
$ cat lib_id.mli
module Local : sig
  type t
end
type t = External of Lib_name.t | Local of Local.t
```
And you can keep using `Lib_id.Local.t` in places like `Ml_sources` where it only makes sense to talk about local libraries.

anmonteiro:
Thanks for your latest round of review. I addressed your comments in the latest commits.

I made a note to fix both those cases (1. the `(optional)` case; 2. the case where we should be redirecting to the public DB) once this diff is merged.

anmonteiro:
Removing library IDs from `Deprecated_library_name` causes some test failures, because we're using `name -> id -> resolved_library` for name resolution. I believe the simplest approach is adding them back.

rgrinberg:
Which tests fail? I believe they'd be incorrect. Consider this dune file:

```
(deprecated_library_name (old_public_name a) (new_public_name b))
(library (public_name b) (enabled_if false))
(library (name x) (libraries a))
```

Where we have an external library `b` that is found. Where should the `a` resolve`? If you're following the `id` it should be the disabled library in the dune file - which is wrong.

anmonteiro:
> Which tests fail?

the ones around deprecated_library_name in `test/blackbox-tests/test-cases/deprecated-library-name/features.t`

anmonteiro:
> Where we have an external library b that is found. Where should the a resolve? If you're following the id it should be the disabled library in the dune file - which is wrong.

I believe what's failing is the simplest case possible: `(deprecated_library_name (old_public_name a) (new_public_name b))` and `(libraries a)` somewhere.

Since we resolve name -> library_id, and there's no such entry in the map for deprecated libraries, we stopped learning how to resolve `a`.

anmonteiro:
> Since we resolve name -> library_id, and there's no such entry in the map for deprecated libraries, we stopped learning how to resolve a.

In the current implementation, we keep :
1. a map of `name -> id`
2. a map of `id -> found_or_redirect`

we use the result of 1 to consult 2. I suspect that I could change this implementation to track:

1. a map of `name -> found_or_redirect`
2. a map of `id -> found_or_redirect`

and that would resolve deprecated libraries too.

anmonteiro:
> we use the result of 1 to consult 2. I suspect that I could change this implementation to track:
> 
> 1. a map of `name -> found_or_redirect`
> 2. a map of `id -> found_or_redirect`
> 
> and that would resolve deprecated libraries too.

I implemented this approach in the most recent push. Now there's one failing test (in the same `features.t` file) that I believe is testing for a restriction that we're removing in this PR: just like in `test/blackbox-tests/test-cases/lib-collision/lib-collision-private.t`, we're now tolerating libraries with the same name at scope creation time, until something requests to depend on it.

src/dune_rules/lib.mli
```diff
@@ -100,8 +100,20 @@ module DB : sig
...
-    val redirect : db -> Loc.t * Lib_name.t -> t
+    val redirect : db -> Lib_id.t -> t
     val redirect_in_the_same_db : Loc.t * Lib_name.t -> t
+
+    module With_multiple_results : sig
+      type resolve_result := t
+
+      type t = private
+        | Resolve_result of resolve_result
+        | Multiple_results of resolve_result Nonempty_list.t
```
rgrinberg, line 111: Don't think this type gains us much over just using a list. The empty list can just correspond to `not_found` as well.

src/dune_rules/lib.ml
```diff
@@ -1116,7 +1139,85 @@ end = struct
...
+              | false ->
+                let name =
+                  if Lib_name.equal a.name b.name then a.name else Lib_id.name a_id
+                and loc_a = Lib_info.loc a.info
+                and loc_b = Lib_info.loc b.info in
+                Status.Invalid (Error.duplicated ~loc_a ~loc_b ~name))
+           | Invalid _, _ -> acc
+           | (Found _ as lib), (Hidden _ | Ignore | Not_found | Invalid _)
+           | (Hidden _ | Ignore | Not_found), (Found _ as lib) -> lib
+           | (Hidden _ | Ignore | Not_found), (Hidden _ | Ignore | Not_found | Invalid _)
```
rgrinberg, line 1212: This is rather hairy. Maybe try cleaning this up by pattern matching first on `acc` and then on `status`?

rgrinberg:
> > Where we have an external library b that is found. Where should the a resolve? If you're following the id it should be the disabled library in the dune file - which is wrong.
> 
> I believe what's failing is the simplest case possible: `(deprecated_library_name (old_public_name a) (new_public_name b))` and `(libraries a)` somewhere.
> 
> Since we resolve name -> library_id, and there's no such entry in the map for deprecated libraries, we stopped learning how to resolve `a`.

Right, the new code you've pushed looks good to me now. The only id based redirects are those that are introduced by `public_name`. Only for those we know the target stanza (and therefore the id).

The failing test in features.t needs to be adjusted to force using those duplicated names somehow. I suspect that should be easy to do.

src/dune_rules/lib.ml
anmonteiro: I simplified it in 2f43f38dafebf2f2b8434901d909eef820254575

anmonteiro:
> The failing test in features.t needs to be adjusted to force using those duplicated names somehow. I suspect that should be easy to do.

I think I hadn't paid enough attention to what we were checking when constructing the scope from stanzas. I re-added the check that was present in a previous version of this PR in f14a0549139df53f2025aca760b7cef2c8347dc9

src/dune_rules/lib.mli
```diff
@@ -100,8 +100,20 @@ module DB : sig
...
-    val redirect : db -> Loc.t * Lib_name.t -> t
+    val redirect : db -> Lib_id.t -> t
     val redirect_in_the_same_db : Loc.t * Lib_name.t -> t
+
+    module With_multiple_results : sig
+      type resolve_result := t
+
+      type t = private
+        | Resolve_result of resolve_result
+        | Multiple_results of resolve_result list
```
rgrinberg, line 111: You can go further. Given that `Resolve_result x = Multiple_results [ x ]`, there's no need for the individual constructors anymore. You can just have:

```
type result
type resolve_result = result list
```

anmonteiro: indeed. just did that in [b6ddab6](https://github.com/ocaml/dune/pull/10307/commits/b6ddab6dc5c20915d5987508865ac60a9e9860d8)

src/dune_rules/scope.ml
```diff
@@ -84,20 +81,20 @@ module DB = struct
...
 
   let resolve_lib_id lib_id_map lib_id =
     match Lib_id.Map.find lib_id_map lib_id with
-    | None -> Memo.return Resolve_result.not_found
+    | None -> Memo.return Lib.DB.Resolve_result.not_found
     | Some found_or_redirect -> resolve_found_or_redirect found_or_redirect
   ;;
 
-  let not_found = With_multiple_results.resolve_result Resolve_result.not_found
+  let not_found = [ Lib.DB.Resolve_result.not_found ]
```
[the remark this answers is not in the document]
anmonteiro: ah yeah, that's even better. will change

rgrinberg (approved):

src/dune_rules/scope.ml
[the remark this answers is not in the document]
anmonteiro: wait that breaks some tests, e.g. `@test/blackbox-tests/test-cases/enabled_if/eif-melange-emit-name-collision`:

```diff
 $ dune build
+  File "a/dune", line 1, characters 0-72:
+  1 | (melange.emit
+  2 |  (target foo)
+  3 |  (enabled_if (= %{context_name} "default")))
+  Error: Library "melange" not found.
+  -> required by _build/default/a/foo/a/foo.js
+  -> required by alias a/all
+  -> required by alias default
+  File "b/dune", line 1, characters 0-76:
+  1 | (melange.emit
+  2 |  (target foo)
+  3 |  (enabled_if (= %{context_name} "alt-context")))
+  Error: Library "melange" not found.
+  -> required by _build/alt-context/b/foo/b/foo.js
+  -> required by alias b/all (context alt-context)
+  -> required by alias default (context alt-context)
+  [1]
```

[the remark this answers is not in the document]
rgrinberg: Doesn't that mean we're treating `[ not_found ]` and `[]` differently somewhere? That sounds like a bug.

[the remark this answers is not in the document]
anmonteiro: ok, all good. fixed in b2e63f66497ceeeab4f3b3a27455eb701cb673ab. I needed to also handle resolving in the parent DB in the empty list case.

anmonteiro:
Thank you @rgrinberg @jchavarri for the guidance and reviews here. I've added a changelog entry and will merge upon a green CI build.

PR #10312, opened 2024-03-27
test(melange): enable the dune sandbox for melange tests

anmonteiro:
based on the suggestion in https://github.com/ocaml/dune/pull/10295#issuecomment-2014612841


anmonteiro:
`./dune.exe build @test/blackbox-tests/test-cases/melange/odoc` is the only failure. Looks like there might be a missing dependency for the melange odoc rules somewhere?

PR #10319, opened 2024-03-27
doc-new: Omit docs for 'empty libraries'

### LINKED ISSUE #10056: `dune build @doc-new` treats dependency errors as fatal
As reported by @mbarbin, when running `dune build @doc-new` with fatal warnings enabled in the dune env:

```
(env (dev (odoc (warnings fatal))))

(library (public_name foo) (libraries eio))
```
```
(* foo.ml *)
let f = Eio.Switch.run
```

As of `eio.0.14`, we happen to get a warning for that dependency which blocks the html documentation generation of our local `foo` library:

```
$ dune build @doc-new
File "eio.mld", line 10, characters 0-11:
Error: '{!modules ...}' should not be empty.
ERROR: Warnings have been generated.

$ ls _build/default/_doc_new/html/docs/
findlib-0  findlib-1  local  odoc.support  stdlib

$ ls _build/default/_doc_new/html/docs/local/
index.html   # missing `foo` folder
```

Note that when `sherlodoc` is installed, there is even less directories/files generated in the `_doc_new` folder:

```
$ opam install sherlodoc
$ dune build @doc-new
$ ls _build/default/_doc_new/html/docs/  
odoc.support  sherlodoc.js  # missing all html files/folders
```

Removing the fatal warnings configuration enables the production of our local `Foo` documentation:

```
$ tree _build/default/_doc_new/html/docs/local/
_build/default/_doc_new/html/docs/local/foo
├── Foo
│   └── index.html
└── index.html
```

It appears that [warnings from external dependencies were expected to be quietly ignored](https://github.com/art-w/sherlodoc/issues/29#issuecomment-1946142240) rather than interpreted as fatal:

> @jonludlam : This should already be implemented, modulo bugs - we're supposed to pass the 'quiet' flag to external libs (things that come from opam rather than in your tree), and this should suppress the --warn-error flag: https://github.com/ocaml/dune/blob/main/src/dune_rules/odoc.ml#L237-L246 - if that's not working please file an issue!

(tested with dune 3.14.0 and ocaml 5.1.0)

jonludlam:
In fact this is likely not an empty library, but a package that either doesn't contain libraries or only contains sublibraries. In either case, it's not at all useful to say a library exists and has no modules!

Fixes #10056

### DIFF
--- a/doc/changes/10319.md
+++ b/doc/changes/10319.md
@@ -1,0 +1,2 @@
+- Don't try to document non-existent libraries in doc-new target (#10319,
+  fixes #10056, @jonludlam)
--- a/src/dune_rules/odoc_new.ml
+++ b/src/dune_rules/odoc_new.ml
@@ -1619,21 +1619,24 @@ let standard_index_contents b entry_modules =
   entry_modules
   |> List.sort ~compare:(fun (x, _) (y, _) -> Lib_name.compare x y)
   |> List.iter ~f:(fun (lib, modules) ->
-    Printf.bprintf b "{1 Library %s}\n" (Lib_name.to_string lib);
-    Buffer.add_string
-      b
-      (match modules with
-       | [ x ] ->
-         sprintf
-           "The entry point of this library is the module:\n{!%s}.\n"
-           (Artifact.reference x)
-       | _ ->
-         sprintf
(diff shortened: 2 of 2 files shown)


rgrinberg (approved):

rgrinberg:
CHANGES entry please

PR #10322, opened 2024-03-28
feat(describe-pp): print reason files with `refmt`

anmonteiro:
- this initial diff has `refmt` hardcoded in `bin/describe_pp.ml` 
- It might be better to add a new field to the dialect and model this use case in the dialect directly too


test/blackbox-tests/test-cases/describe/describe-pp/describe-pp.t/run.t
```diff
@@ -16,4 +16,6 @@ We also make sure that the dump file is not present
 This also works for reason code
 
   $ dune describe pp src/main_re.re
-  ;;Util.log "Hello, world!"
+  # 1 "_build/default/src/main_re.pp.re.ml"
+  # 1 "src/main_re.pp.re"
```
rgrinberg, line 20: How come we're printing the location directives now?

anmonteiro: the tests have a dedicated `refmt` executable that I changed to match the real `refmt` by printing to stdout

anmonteiro:
> * It might be better to add a new field to the dialect and model this use case in the dialect directly too

I changed the initial implementation to this:
- added a field `dump_ast` in dialects
- described those actions for OCaml and Reason
- in `describe pp`, execute the dialect's `dump_ast` if it exists, falling back to OCaml's, known to exist

rgrinberg (approved):
LGTM. @nojb could you take a look at the dialect changes?

anmonteiro:
I renamed `dump_ast` to `print_ast`, since `dump_ast` was ambiguous (could also be interpreted as "dump AST to a file")

anmonteiro:
perhaps `print_ast` should be versioned?

nojb (approved):
The dialect changes LGTM. Thanks!

PR #10324, opened 2024-03-28
merlin: add new contexts commands

### LINKED ISSUE #10222: Support "multi-context" libraries
## Introduction

At Ahrefs, we are now able to share more code between Melange and OCaml, thanks largely to Dune.

One of the main drivers for increasing code sharing is the UI for our products. This UI is comprised of React components implemented with Melange and [ReasonReact](https://reasonml.github.io/reason-react/), the Melange bindings to [ReactJS](https://react.dev/). Recently, we began using [server-reason-react](https://github.com/ml-in-barcelona/server-reason-react) to migrate some of these components to the server, which is built in OCaml.

As more components are migrated to be compilable and executable in OCaml to become "shared components", it is useful to rewrite existing Melange libraries in OCaml, preserving their structural layout (modules, functions) but using the native platform in the implementation. This allows for these migrations to proceed with fewer changes in the components that are becoming shared.

One way to implement these structurally uniform layouts is by leveraging [Dune contexts](https://dune.readthedocs.io/en/stable/overview.html#term-build-context), and have the Melange implementations live in one context and the OCaml ones in another.

Unfortunately, Dune currently does not support defining two libraries with the same name in different contexts using `enabled_if`. This feature would be very useful to support multi-context libraries, particularly for the UI components we are currently migrating.

## Motivation

When it comes to shared Melange + OCaml code, the current approach we use involves building all code under the same Dune context and using `copy_files` to place the shared modules in the right folders.

This approach has a few downsides. For example, it is impossible to use the same name for both implementations of the same library. When attempting to define two libraries with the same name in the same context, Dune will rightly complain.

The current workaround is to use different names for each implementation and make all these multi-context libraries unwrapped, with the disadvantages that unwrapped libraries entail at scale, such as module collisions.

In addition to the above, `copy_files` brings its own issues with build performance, editor breakages, and problems when using PPX, which sometimes lead to "Multiple rules generated" Dune errors.

## Technical details

@andreypopp originally explained some of these issues in depth in this [Melange discussion](https://github.com/melange-re/melange/discussions/694). As mentioned in that discussion, our initial idea was to introduce a new Dune stanza for Melange libraries, and also introduce a separate PPX pipeline, exclusively for Melange.

Over time, we realized this is precisely what Dune build contexts are for. Dune contexts allow us to define completely separate builds, each with its own PPX pipelines. Libraries can also be set conditionally on a given context using `enabled_if` and the `%{context_name}` variable. For example:

```clojure
(library
 (name foo)
 (enabled_if (= %{context_name} "alt-context")))
```

So it seems like the perfect solution for our use case.

Currently, support in Dune for this use case is limited. Dune makes a lot of assumptions that only one context will be used on a given build. Things like library duplication checks or source duplication checks don’t take the `enabled_if` field into account.

Our plan is to fix this with the help of the Dune maintainers team. We have started writing tests and prototyping some solutions in [ocaml/dune#9839](https://github.com/ocaml/dune/pull/9839) and [ocaml/dune#10179](https://github.com/ocaml/dune/pull/10179), and a few others have been extracted from these PRs.

After making Dune capable of handling multi-context libraries, we should fix editor integration too, so that users can choose in which build context they want to work. We also realized that the Dune-Merlin integration doesn’t support multiple contexts (see [ocaml/dune#10035](https://github.com/ocaml/dune/pull/10035)), and we will certainly need to add some integration in vscode-ocaml.

Eventually, this work might be useful for other alternate tooling supported by Dune, like Coq, Js_of_ocaml, or even cross-compilation. One can imagine users of these tools wanting to have a library with the same name defined in an OCaml native context and a more specific context.

## Drawbacks

Using two Dune contexts might lead to relatively slower builds than using a single context, as code that is 100% shared across Melange and OCaml will have to be built twice. But arguably, this is kind of the case already, as even in the same context, Dune has to generate and execute rules to run both `melc` and `ocamlc/opt`, so the performance impact should, in theory, be limited.

## Alternatives

Adding a new stanza like `melange.library` is the main alternative proposed, which can be found in the [Melange discussion](https://github.com/melange-re/melange/discussions/694) mentioned above.

## Unresolved Questions

Most of the unknowns remain in what specific parts of Dune have to be modified to make this work. We discover new cases as we go and add tests, but we also need guidance from Dune maintainers before proceeding as some changes might be too invasive, or might break the way in which Dune handle specific errors.

Another part that is not 100% clear yet is how editor integration will work, but it should be easy to configure Dune to provide paths to artifacts based on a chosen context.

jchavarri:
Fixes #10222.

This work is part of the multi-context support for vscode (see https://github.com/ocamllabs/vscode-ocaml-platform/issues/1432).

## Current version

This PR prepares Dune to support multi-context config retrieval from Merlin by doing the following:
- Add an optional flag `--context` to `dune ocaml-merlin`: if passed, this flag will be used to return data to subsequent requests to the `ocaml-merlin` server
- Add `dune describe contexts` subcommands, which returns the list of Dune contexts available (1 per line)
- It also adds a new field `generate_merlin_rules` to the contexts declared in the workspace (originally implemented in #10328). This new field allows to generate merlin rules in any case, even if the context is not selected using `(merlin)`. This addition is useful to dynamically change the context for Merlin via the `SetContext` command mentioned above.

---

## Original version (now obsolete)

This PR adds two new commands to the ocaml-merlin API:

- `GetContexts`: returns the list of contexts available in the workspace (only ctxt names)
- `SetContext(name)`: sets the current context to get merlin data from

It also adds a new field `generate_merlin_rules` to the contexts declared in the workspace (originally implemented in #10328). This new field allows to generate merlin rules in any case, even if the context is not selected using `(merlin)`. This addition is useful to dynamically change the context for Merlin via the `SetContext` command mentioned above.


jchavarri:
@voodoos Thanks for the feedback. I wasn't aware of how ocaml-lsp and Dune communicated through Dune's RPC feature, I understand you are referring to [this `Dune` module](https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/src/dune.mli)?

I am not sure I understand how the suggested approach would work in terms of storing the information about the selected context. In the current implementation, the `SetContext` command will change the value of the current context:

https://github.com/ocaml/dune/blob/7a5395eb621a01e700ad2bab7897a446f527383d/bin/ocaml/ocaml_merlin.ml#L215

which is stored as a `ref` that is initialized as soon as `dune ocaml-merlin` starts:

https://github.com/ocaml/dune/blob/7a5395eb621a01e700ad2bab7897a446f527383d/bin/ocaml/ocaml_merlin.ml#L181

This is handy because this value can also be used by `dune ocaml-merlin` where the editor information is requested:

https://github.com/ocaml/dune/blob/7a5395eb621a01e700ad2bab7897a446f527383d/bin/ocaml/ocaml_merlin.ml#L186

As can be seen, this state handling happens only a few lines apart and doesn't spread to other modules outside `Ocaml_merlin`.

If the new commands were moved to Dune's RPC, then we would need to find a way to sync between the two processes (`rpc` and `ocaml-merlin`), because we would be writing the state through Dune's RPC, but we would still have to read it from the `ocaml-merlin` process. Or is there an alternative to avoid that?

> Adding new features to `dune ocaml-merlin` would only make a transition to the RPC more complex.

Considering the above, I think this addition doesn't change much the current state of things, and moving all the Merlin configuration handling and state at once to the RPC will be easier than spreading state across processes. I would be happy to eventually work on transitioning everything to the RPC if/when time permits, but that sounds like a much larger task and scope than the current goal: to allow users to select the context used by `print_merlin_conf`.

jchavarri:
> Since we might switch every plugins to using lsp in the future, and you seem to mostly target vscode, that might not be an issue ?

Unfortunately, I would need to support users that don't use vscode / ocaml-lsp and query merlin directly via vim or emacs.

test/blackbox-tests/test-cases/merlin/get-set-contexts.t
```diff
@@ -0,0 +1,86 @@
...
+  >   (name alt)
+  >   (generate_merlin_rules)))
+  > EOF
+
+  $ dune build
+
+Request config for file in alt context before calling SetContext
+
+  $ printf '(4:File%d:%s)' ${#FILE2} $FILE2 | dune ocaml-merlin | grep -i "$lib2" | sed 's/^[^:]*:[^:]*://'
+  No config found for file bar.ml. Try calling 'dune build'.))
```
voodoos, line 69: I think we should improve the error messages now that the user might be using the wrong context since `dune build` won't solve that issue.

jchavarri:
@voodoos Thanks for the review. I wonder what your thoughts are about the suggestion in https://github.com/ocamllabs/vscode-ocaml-platform/issues/1432#issuecomment-2034676788? In particular, this part:

> add `--context CONTEXT` option `dune ocaml ocaml-merlin`

Maybe that would make easier to transition to RPC in the future? We would just need to move the `--context` flag from the `dune ocaml-merlin` executable to the RPC one. Besides that, I find interesting about the suggestion that it removes the need to patch `dot-merlin-reader`, which I was preparing in https://github.com/ocaml/merlin/pull/1743/.

jchavarri:
@voodoos I have updated the PR to follow our recent decision to use a `--context` flag rather than a new protocol command to set the context. I also added a new subcommand `dune describe contexts` to list the contexts, which I will use from ocaml-lsp later on to offer clients the list of contexts available (to show in e.g. vscode a menu to choose from).

The work on [ocaml-lsp](https://github.com/ocaml/ocaml-lsp/pull/1238) is still ongoing, but any feedback you might have on this one is already much appreciated. Thanks!

bin/ocaml/ocaml_merlin.ml
```diff
@@ -129,48 +133,71 @@ end = struct
...
     | Ok path ->
       let files = get_merlin_files_paths path in
       Merlin.Processed.print_generic_dot_merlin files
   ;;
 
-  let start () =
+  let start ~ctx_name () =
+    let open Fiber.O in
+    let* setup = Import.Main.setup () in
+    let* setup = Memo.run setup in
```
voodoos, line 179: I am wondering about the cost of these additional initializations. Is it really useful to check the existence of the context ? If it does not exist a "No config found for file %s." would be printed anyway.

jchavarri: As it was a one-time thing (this code will run once when the ocaml-merlin process starts) I assumed it wouldn't have a lot of impact. Removed in https://github.com/ocaml/dune/pull/10324/commits/f21fbf837d177bedb8d332d236b8d2e97ef8a720.

jchavarri:
The downstream PRs that depend on this feature are ready for review:

- [ocaml-lsp#1238](https://github.com/ocaml/ocaml-lsp/pull/1238/)
- [vscode-ocaml-platform#1449](https://github.com/ocamllabs/vscode-ocaml-platform/pull/1449)

The latter includes a short screen capture that helps visualize what the features added in this PR enable, from the user perspective.

jchavarri:
@rgrinberg @voodoos Apologies for pinging, I wonder if you have any thoughts on this PR? Thanks.

bin/ocaml/ocaml_merlin.ml
```diff
@@ -129,48 +133,61 @@ end = struct
...
   ;;
 
-  let start () =
+  let start ~ctx_name () =
+    let open Fiber.O in
+    let selected_context =
+      match ctx_name with
+      | None -> Default
+      | Some ctx_name ->
+        (match Context_name.of_string_opt ctx_name with
```
anmonteiro, line 182: I was wondering what the valid names for contexts were and incidentally found a crash: https://github.com/ocaml/dune/pull/10472

bin/ocaml/ocaml_merlin.ml
```diff
@@ -129,48 +133,61 @@ end = struct
...
+         | None -> User_error.raise [ Pp.textf "Invalid context name %S" ctx_name ]
```
anmonteiro, line 183: we could provide a hint (maybe it could even live in `context_name.ml`) about what's a valid context name here.

bin/ocaml/ocaml_merlin.ml
```diff
@@ -129,48 +133,61 @@ end = struct
...
-    to_local s
+    to_local ~selected_context:Default s
     >>| function
     | Error mess -> Printf.eprintf "%s\n%!" mess
     | Ok path -> get_merlin_files_paths path |> List.iter ~f:Merlin.Processed.print_file
   ;;
 
   let dump_dot_merlin s =
-    to_local s
+    to_local ~selected_context:Default s
```
anmonteiro, line 168: wouldn't this also benefit from printing the config for a given context argument?

doc/changes/10324.md
```diff
@@ -0,0 +1,5 @@
+- Add new flag `--context` to `ocaml-merlin`, which allows to select a Dune
```
anmonteiro, line 1:
```suggestion
- Add new flag `--context` to `dune ocaml-merlin`, which allows to select a Dune
```

doc/reference/dune-workspace/context.rst
```diff
@@ -21,6 +21,9 @@ the description of an opam switch, as follows:
 
 - ``(merlin)`` instructs Dune to use this build context for Merlin.
 
+- ``(generate_merlin_rules)`` instructs Dune to generate Merlin rules for this
+  context, even if it is not the one selected for Merlin.
```
anmonteiro, line 25:
```suggestion
  context, even if it is not the one selected via ``(merlin)``.
```

src/dune_rules/workspace.ml
```diff
@@ -384,7 +406,8 @@ module Context = struct
           "instrument_with"
           (Dune_lang.Syntax.since syntax (2, 7) >>> repeat Lib_name.decode)
       and+ loc = loc
-      and+ merlin = field_b "merlin" in
+      and+ merlin = field_b "merlin"
+      and+ generate_merlin_rules = field_b "generate_merlin_rules" in
```
anmonteiro, line 410: should this have a `since 3.16` check?

anmonteiro (commented):
This looks good overall and my suggestions are mostly minor. The important ones are:

- should we gate this on `lang dune 3.16`?
- should we make the other commands in `dune ocaml-merlin` take `--context` too, not just `print-conf`?

src/dune_rules/workspace.ml
jchavarri: 👍 done in https://github.com/ocaml/dune/pull/10324/commits/d384104f3f2872b1c16c2f5a61c556292675e190.

bin/ocaml/ocaml_merlin.ml
jchavarri: yup, added in https://github.com/ocaml/dune/pull/10324/commits/eb09dbaa6a1539822bcc0a06d650925590867d2b.

bin/ocaml/ocaml_merlin.ml
jchavarri: i agree. maybe this can be done in a separate PR (worth including a fix for your repro in #10472 as well).

jchavarri:
@anmonteiro thanks for the review!

>     * should we gate this on `lang dune 3.16`?
>     * should we make the other commands in `dune ocaml-merlin` take `--context` too, not just `print-conf`?

I thought the answer to these 2 questions is yes, so just included them in the PR, pls lmk if I'm missing anything :)

bin/ocaml/ocaml_merlin.ml
```diff
@@ -264,6 +291,14 @@ module Dump_dot_merlin = struct
             ~doc:
               "The path to the folder of which the configuration should be printed. \
                Defaults to the current directory.")
+    and+ ctx_name =
```
anmonteiro, line 294: I missed this at first, but I think this is where you should be parsing the context, instead of `arg_to_ctx`.

```ocaml
Arg.(value & opt Selected_context.conv Selected_context.Default & ...)
```

where `Selected_context.conv` is 


```ocaml
let conv =
  let parse ctx_name =
    match Context_name.of_string_opt ctx_name with
    | None -> User_error.raise [ Pp.textf "Invalid context name %S" ctx_name ]
    | Some ctx_name -> Custom ctx_name
  in
  let print = function Default -> "default" | Custom ctx -> ctx in
  Arg.conv ~docv:"context" (parse, print)
```

bin/ocaml/ocaml_merlin.ml
```diff
@@ -264,6 +291,14 @@ module Dump_dot_merlin = struct
...
+      Arg.(
+        value
+        & opt (some string) None
+        & info
+            [ "context" ]
+            ~docv:"CONTEXT"
+            ~doc:"The Dune context in which the command will return information for")
```
anmonteiro, line 301: What do you think about this phrasing? I can't exactly point out what, but the current sentence reads sort of weird.

```suggestion
            ~doc:"Dune will return information for this build context")
```

anmonteiro (approved):
looks good. I just think we need to improve the context arg parsing. I made a suggestion that avoids creating the intermediate option type.

bin/ocaml/ocaml_merlin.ml
jchavarri: Added in https://github.com/ocaml/dune/pull/10324/commits/666e5250015e19f5f5d391b7e35ac64e13e6963b.

anmonteiro (approved):

anmonteiro:
I tried resolving conflicts on this but `ls -a ./_build/cross/.merlin-conf` and `ls -a ./_build/default/.merlin-conf` don't seem to exist anymore in tests (`test/blackbox-tests/test-cases/merlin/default-based-context.t/run.t`).

I wonder if it's because of #8567. Bisecting....

anmonteiro:
nevermind, I just messed up the rebase.

anmonteiro:
I also just pushed a commit that:

- removes `Selected_context.t` (`Context_name.t` and associated functions have all we need)
- add the `--context` arg to `dump` too
- reuse the context arg as `Selected_context.arg`

anmonteiro:
Thank you, let's do it

PR #10325, opened 2024-03-28
merlin: add rules always, regardless the configuration of `(merlin)` field

jchavarri:
This change is related to the introduction of selected context for merlin controlled via ocaml-lsp (see related issue https://github.com/ocamllabs/vscode-ocaml-platform/issues/1432, and ongoing dune work in https://github.com/ocaml/dune/pull/10324).

The change consists on always adding merlin rules, regardless the user choice of context through the `(merlin)` field in `dune-workspace`.

The reason for this change is that it unlocks dynamically switching the context used for merlin without the need to rebuild again, nor changing the `(merlin)` value in `dune-workspace` (which kind of defeats the goal of dynamically switching context).

In theory, the change should not introduce any regressions, besides the expected extra time to create and execute the new rules added, because each context is independent from the others.

It shouldn't change any external behavior from the Dune user point of view either, because `Ocaml_merlin` isn't reading the information from `Context` but from the workspace value directly:

https://github.com/ocaml/dune/blob/9265e2c584b8c9e9e3333da83a5c37a162dd8d28/bin/ocaml/ocaml_merlin.ml#L138


rgrinberg:
It's not great that everyone will have to pay for this performance penalty for this relatively niche feature. What about just adding a setting for enabling/disabling the merlin per context?

jchavarri:
> for this performance penalty

It is worth noting that the penalty for users with a single Dune context (which I am guessing should be the majority) will be non existent.

> for this relatively niche feature

The feature doesn't exist yet, so it definitely has very few users 😄

> What about just adding a setting for enabling/disabling the merlin per context?

You mean some new field in context like `generate_merlin_rules`? What would happen if user configures some context like:

```clojure
(context
 (default
  (name foo)
  (generate_merlin_rules false)
  (merlin)))
```
?

To avoid such inconsistencies, maybe we can set some field in Dune config that applies to all contexts? Similar to `sandboxing_preference`:

```
(generate_merlin_rules <setting>)
```

where `<setting>` can be one of:
- `current`: Only generates Merlin rules for the current context selected. This is the default value.
- `all`: Generates Merlin rules for all contexts.

jchavarri:
Some random metrics to gather usage of `context` from GitHub open source repos:

- 263 `dune-workspace` files found ([link](https://github.com/search?q=path%3A**%2Fdune-workspace&type=code&ref=advsearch))
- of those, 106 files use `context` stanza ([link](https://github.com/search?q=context+path%3A**%2Fdune-workspace&type=code&ref=advsearch&p=2))
- ~26 of those have more than one context defined

jchavarri:
I also searched for `dune-project` and there are [~9500 files found](https://github.com/search?q=path%3A**%2Fdune-project&type=code&ref=advsearch). So at least based on this metric from open source OCaml projects hosted on GitHub, less than 0.3% of projects would be affected by the additional work.

jchavarri:
> What about just adding a setting for enabling/disabling the merlin per context?

I implemented my interpretation of this suggestion in #10328.

jchavarri:
@rgrinberg Any thoughts on the above comments or the alt approach suggested that was implemented in #10328? Considering how niche the usage of multi-context is, I tend to think introducing this extra work will not have a lot of impact of users (as opposed to adding yet another configuration flag to Dune). cc @anmonteiro

rgrinberg:
If we had a way to make cross compilation work without contexts, maybe this would be acceptable. Your usage search omits all the uses of opam-cross-{windows,android}, which use environment variables to setup a cross compilation build environment. Essentially all OCaml cross compilation is done through these repos, so it's pretty important. #10328 is the right way to go.

Issue #10317, opened 2024-03-27
gettext-camomile fails to build on nix (3.15.0~alpha1)


emillon:
https://github.com/nix-ocaml/nix-overlays/actions/runs/8438106703/job/23120699845?pr=1349

```
Running phase: checkPhase
File "test/test-camomile/dune", line 2, characters 7-11:
2 |  (name test)
           ^^^^
(cd _build/default/test/test-camomile && ./test.exe -test-dir ../testdata)
Fatal error: exception File "src/configImpl.ml", line 1, characters 58-64: Assertion failed
                                   
error: builder for '/nix/store/pww9d1y40cp8im0ggsaw051j2v6j163q-ocaml4.14.2-gettext-camomile-0.4.2.drv' failed with exit code 1;
       last 15 log lines:
       > Running phase: unpackPhase
       > unpacking source archive /nix/store/rz68lchjc6v1i9p1cahqlhvif4acfpsr-source
       > source root is source
       > Running phase: patchPhase
       > Running phase: updateAutotoolsGnuConfigScriptsPhase
       > Running phase: configurePhase
       > no configure script, doing nothing
       > Running phase: buildPhase
       > Running phase: checkPhase
       > File "test/test-camomile/dune", line 2, characters 7-11:
       > 2 |  (name test)
       >            ^^^^
       > (cd _build/default/test/test-camomile && ./test.exe -test-dir ../testdata)
       > Fatal error: exception File "src/configImpl.ml", line 1, characters 58-64: Assertion failed
       >
       For full logs, run 'nix log /nix/store/pww9d1y40cp8im0ggsaw051j2v6j163q-ocaml4.14.2-gettext-camomile-0.4.2.drv'.
```

emillon:
Definitely a dune-site issue:
```ocaml
let get_path site = match site with [path] -> path | _ -> assert false
```

emillon:
Maybe not a dune-site issue actually! bisected to 2e11cbd8079ca44de92ed778e40ae2ef34501da4 (#10275)

emillon:
debugging notes:
- this is reproducible on nix, but not on opam, because `dune install` is not used by opam
- changing `if executable` to `if true` does not fix it
- adding a `seek_in` in the fallback branch does not fix it

next step is trying to see the diff in installed files to see if it's truncation or something else.

emillon:
Actually removing `if executable` works, I'll send a PR

jchavarri:
[...]

PR #10326, opened 2024-03-28
test: sites modules substitution with dune install

emillon:
Repro for #10317

### DIFF
--- a/otherlibs/dune-site/test/install-subst.t
+++ b/otherlibs/dune-site/test/install-subst.t
@@ -1,0 +1,44 @@
+This checks that generated sites files correctly substituted with `dune install`.
+
+See #10317.
+
+  $ cat > dune-project << EOF
+  > (lang dune 3.0)
+  > (using dune_site 0.1)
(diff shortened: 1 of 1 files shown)


otherlibs/dune-site/test/install-subst.t
```diff
@@ -0,0 +1,56 @@
...
+  > EOF
+
+  $ dune build
+
+  $ has_placeholder() {
+  >   name=$1;
+  >   if grep -q DUNE_PLACEHOLDER "$name"; then
+  >     printf "%-30s has placeholders\n" "$name"
+  >   else
+  >     printf "%-30s has no placeholders\n" "$name"
```
Leonidas-from-XIV, line 29: Maybe it should not print anything in the "success" case, that way files that are not that relevant are not included in the output thus making the test less fragile.

Leonidas-from-XIV (approved):
Its a good test showing that we're accidentally installing something that has escaped expansion.

PR #10328, opened 2024-03-28
merlin: introduce `generate_merlin_rules`

jchavarri:
Alternate implementation to #10325.

Instead of enabling merlin rules for all contexts, this PR applies the suggestion from https://github.com/ocaml/dune/pull/10325#issuecomment-2025302408, by adding a `generate_merlin_rules` field that can be used from each context to opt-in for this behavior.


rgrinberg:
This looks like the right approach. I think you should combine this PR with #10328. It will be easier for me to asses the final result this way and the value of all of these configuration options. It would also make it easier for me to test it.

jchavarri:
Thanks for the feedback. I incorporated these changes in https://github.com/ocaml/dune/pull/10324.

PR #10327, opened 2024-03-28
fix(install): do substitution for sites files

emillon:
Fixes #10317

PR #10332, opened 2024-03-29
[3.15] Backport #10327

emillon:
- test: sites modules substitution with dune install (#10326)
- fix(install): do substitution for sites files (#10327)
- Changelog

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+3.15.0~alpha2 (unreleased)
+--------------------------
+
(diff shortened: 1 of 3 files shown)


Leonidas-from-XIV (approved):
Very nice. A shame this is probably going to make it (slightly) slower, but we might be able to track that the file needs substitution at some other earlier point through the flow.

emillon:
to be clear, this is still strictly faster than 3.14.2 (which used slow copy everywhere, even in the fallback case)

PR #10333, opened 2024-03-29
fix: use `Long_val` for sendfile() parameters to fix file copying in docker

tonyfettes:
I was using dune in docker, and every it copies some large file in a data_only_dirs it reports that `sendfile: No such file or directory`. I tried digging into the source and I think converting the ocaml integer to 32-bit version will make it impossible to copy any file that is larger than 4 GB. I have tested in a docker, but I cannot really confirm whether it really works on real linux machine (I'm using macOS).

### DIFF
--- a/otherlibs/stdune/src/copyfile_stubs.c
+++ b/otherlibs/stdune/src/copyfile_stubs.c
@@ -74,7 +74,7 @@ CAMLprim value stdune_copyfile(value v_from, value v_to) {
 }
 
 static int dune_sendfile(int in, int out, size_t length) {
-  int ret;
+  ssize_t ret;
   while (length > 0) {
     ret = sendfile(out, in, NULL, length);
     if (ret < 0) {
--- a/otherlibs/stdune/src/copyfile_stubs.c
+++ b/otherlibs/stdune/src/copyfile_stubs.c
@@ -73,7 +73,7 @@ CAMLprim value stdune_copyfile(value v_from, value v_to) {
   caml_failwith("copyfile: only on macos");
 }
 
-static int dune_sendfile(int in, int out, int length) {
+static int dune_sendfile(int in, int out, size_t length) {
   int ret;
   while (length > 0) {
     ret = sendfile(out, in, NULL, length);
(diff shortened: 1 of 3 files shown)


emillon:
Thanks, I'll try to test this.

otherlibs/stdune/src/copyfile_stubs.c
```diff
@@ -73,7 +73,7 @@ CAMLprim value stdune_copyfile(value v_from, value v_to) {
   caml_failwith("copyfile: only on macos");
 }
 
-static int dune_sendfile(int in, int out, int length) {
+static int dune_sendfile(int in, int out, size_t length) {
   int ret;
```
emillon, line 77:
```suggestion
  ssize_t ret;
```

(since that's the return type of `sendfile`)

otherlibs/stdune/src/copyfile_stubs.c
```diff
@@ -90,7 +90,7 @@ CAMLprim value stdune_sendfile(value v_in, value v_out, value v_size) {
   caml_release_runtime_system();
   /* TODO Use copy_file_range once we have a good mechanism to test for its
    * existence */
-  int ret = dune_sendfile(FD_val(v_in), FD_val(v_out), Int_val(v_size));
+  int ret = dune_sendfile(FD_val(v_in), FD_val(v_out), Long_val(v_size));
```
emillon, line 93: FTR, I checked whether it's safe to use `size_t` with `Long_val`; the ocaml runtime does that so it seems good.

emillon (commented):
I checked that it fixes the issue with the test in #10334

emillon:
@dra27 does this look ok or are there any considerations for using `Long_val` here?

rgrinberg (approved):

anmonteiro (approved):

dra27:
This will fix files over 4GB on 64-bit systems, but it will still overflow on 32-bit systems. It's a slightly confusingly-named macro, but `Long_val` converts an OCaml integer value to a value with the same size as a pointer (so a 64-bit integer on 64-bit systems) whereas `Int_val` converts to a C `int` (which is always 32bits, as already noted).

The only way to fix this for 32-bit systems is to use an `int64` or, with slightly heavier-weight C code, an `int63`.

I'm not sure how much that level of 32-bit support really matters for Dune, though?

rgrinberg:
Yes, there's no need to bother 32 bit support for dune itself.

PR #10334, opened 2024-03-29
test: demonstrate overflow with sendfile

emillon:
See #10333

### DIFF
--- a/test/blackbox-tests/test-cases/dune
+++ b/test/blackbox-tests/test-cases/dune
@@ -43,6 +43,13 @@
  (applies_to fdo)
  (enabled_if false))
 
+; This test copies a large file, which is too slow to be enabled all the time.
+; The source file is actually sparse, but sendfile seems to do a full copy
+; nevertheless.
+(cram
+ (applies_to sendfile-large-file)
+ (enabled_if false))
+
 ; This test is flaky
 
 (cram
--- a/test/blackbox-tests/test-cases/sendfile-large-file.t
+++ b/test/blackbox-tests/test-cases/sendfile-large-file.t
@@ -1,0 +1,22 @@
+We create a large file and check that it is copied completely.
+
+  $ cat > dune-project << EOF
+  > (lang dune 1.0)
+  > EOF
+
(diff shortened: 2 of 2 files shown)


test/blackbox-tests/test-cases/dune
```diff
@@ -43,6 +43,10 @@
  (applies_to fdo)
  (enabled_if false))
 
+(cram
+ (applies_to sendfile-large-file)
+ (enabled_if false))
```
rgrinberg, line 48: Would be helpful to describe in a comment why this test is disabled

test/blackbox-tests/test-cases/sendfile-large-file.t
```diff
@@ -0,0 +1,22 @@
+We create a large file and check that it is copied completely.
+
+  $ cat > dune-project << EOF
+  > (lang dune 1.0)
+  > EOF
+
+  $ cat > create.ml << EOF
+  > let () = Unix.truncate "file.dat" 0x1_00_00_00_03
```
Leonidas-from-XIV, line 8: Can we make sure somehow that the file that is being created here is getting created on a file system that supports sparse files?

Looking at [this comparison list](https://en.wikipedia.org/wiki/Comparison_of_file_systems) it seems like it should work on Windows assuming its running on NTFS and not some FAT-varaiant and it works on MacOS as long as the FS is APFS and not HFS+ (BSDs and Linux seem to be pretty well-covered wrt to sparse file FS support).

emillon: The test is disabled by default because it's pretty slow in its current implementation. If we can make it use sparse files in the usual test targets, I think we can enable it by default. In any case I'll add a note about that. If it's complicated to use sparse files I'll disable the test.

emillon:
ok so actually truncate makes a sparse file, but `sendfile` doesn't preserve holes in files (at least on ext4), so the test takes ~4s. The benefit / time balance doesn't feel in favor of enabling the test by default, I'll add a note about that.

Issue #10300, opened 2024-03-22
3.15.0 release tracking


emillon:
## pre-branch blockers

(none)

## known blockers

- [x] All 3.14 changelogs merged
- [x] test mirage
- [x] test nix-overlays
* [x] #10327
  * [x] backport: #10332 
* [x] #10314
* [x] #10315
* [x] #10336
  * [x] backport: #10337

## alpha1 (2024-03-25)

- [x] release https://github.com/ocaml/opam-repository/pull/25585
- [x] triage
- [x] go/no go

## alpha2 (2024-03-29)

- [x] release https://github.com/ocaml/opam-repository/pull/25605
  - [x] test nix-overlays: https://github.com/nix-ocaml/nix-overlays/pull/1356
- [x] triage
- [x] go/no go

## release

- [x] changelog
- [x] release https://github.com/ocaml/opam-repository/pull/25615
- [x] triage

## post-release

- [X] ocaml.org changelog
  - https://github.com/ocaml/ocaml.org/pull/2316
  - https://ocaml.org/changelog/2024-04-03-dune-3.15.0 once ocaml.org updates
- [X] discuss post
 - https://discuss.ocaml.org/t/ann-dune-3-15/14438
- [X] close milestone

PR #10336, opened 2024-03-29
chore: Avoid result < 1.5 when installing, as this breaks on 4.08+

Leonidas-from-XIV:
Detected as part of #10300, the alternate solution in https://github.com/ocaml/opam-repository/pull/25603 is unlikely to go forward at this point so this implements the policy of opam-repository and adds the conflict on our side.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -156,7 +156,9 @@ understood by dune language."))
   (csexp (>= 1.5.0))
   (lwt (>= 5.3.0))
   base-unix)
- (description "Specialization of dune-rpc to Lwt"))
+ (description "Specialization of dune-rpc to Lwt")
+ (conflicts
+   (result (< 1.5))))
 
 (package
  (name dyn)
--- a/dune-rpc-lwt.opam
+++ b/dune-rpc-lwt.opam
@@ -16,6 +16,9 @@ depends: [
   "base-unix"
   "odoc" {with-doc}
 ]
+conflicts: [
+  "result" {< "1.5"}
+]
 dev-repo: "git+https://github.com/ocaml/dune.git"
 build: [
   ["dune" "subst"] {dev}


emillon (approved):

anmonteiro:
Is this needed after https://github.com/ocaml/dune/pull/10261 ?

Leonidas-from-XIV:
This is necessary because of #10261. Before we required `result.1.5` and everything worked, but after removing it (which is a good thing!) package solutions could going forward pick result < 1.5 (in the case of the CI failure linked it was `result.1.0`) which defines a result type that is not compatible with that one in 4.08+, so when `dune-rpc-lwt` would use the 4.08-builtin `result` and another package would use `result.1.0`-provided result they wouldn't be compatible.

My first thought was to restrict result to only have `result.1.5` on 4.08+ so I posted https://github.com/ocaml/opam-repository/pull/25603 but as you can read in the issues linked it wasn't the first time that this has bit someone. For now packages are supposed to conflict with `result` until compatibility with versions older than 4.08 is dropped from opam-repository.

In practice this mostly affects the low-version check in opam-repository, because users would get `result.1.5` as that is the newest version and a number of other packages already conflict on `result < 1.5`.

anmonteiro:
understood, thanks for the details!

dune-project
```diff
@@ -156,7 +156,9 @@ understood by dune language."))
   (csexp (>= 1.5.0))
   (lwt (>= 5.3.0))
   base-unix)
- (description "Specialization of dune-rpc to Lwt"))
+ (description "Specialization of dune-rpc to Lwt")
+ (conflicts
+   (result (< 1.5))))
```
rgrinberg, line 161: Should be 1 space here.

PR #10340, opened 2024-03-30
fix(describe-pp): show output on successive runs

anmonteiro:
- #10322 introduced `Build_system.execute_action` which is cached
- this diff adds the `(universe)` to the anonymous action obtained in the dialect so that output is always shown after running `dune describe pp`

### DIFF
--- a/bin/describe/describe_pp.ml
+++ b/bin/describe/describe_pp.ml
@@ -45,6 +45,13 @@ let print_pped_file sctx file pp_file =
     in
     Action_builder.evaluate_and_collect_facts build
   in
+  let observing_facts =
+    (* We add `(universe)` to the dependencies of this action so that `dune
+       describe pp` always prints output *)
+    match Dep.Map.add observing_facts Dep.universe Dep.Fact.nothing with
+    | Ok x -> x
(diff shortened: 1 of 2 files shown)


bin/describe/describe_pp.ml
```diff
@@ -42,6 +42,13 @@ let print_pped_file sctx file pp_file =
     in
     Action_builder.evaluate_and_collect_facts build
   in
+  let observing_facts =
+    (* We add `(universe)` to the dependencies of this action so that `dune
+       describe pp` always prints output *)
+    match Dep.Map.add observing_facts Dep.universe Dep.Fact.nothing with
+    | Ok x -> x
+    | Error _ -> observing_facts
```
rgrinberg, line 50: I think you're better off just using `Action_exec` directly if you don't want to observe the result. I'd try doing that first.

rgrinberg (approved):
I suppose this is good enough for a hot fix

anmonteiro:
@rgrinberg I pushed a version that uses `Action_exec.exec`. Looks a bit better IMO

rgrinberg (approved):

PR #10342, opened 2024-03-30
fix: forward-compatible fix for OCaml 5.2

anmonteiro:
- Upstream PR https://github.com/ocaml/ocaml/pull/12451 fixed how warning 53 (misplaced attribute)
- this change fixes those instances in dune to avoid reporting warning 53 in builds


src/memo/memo.mli
```diff
@@ -101,7 +101,6 @@ module Make_map_traversals (Map : Map.S) : sig
   val parallel_iter : 'a Map.t -> f:(Map.key -> 'a -> unit t) -> unit t
   val parallel_map : 'a Map.t -> f:(Map.key -> 'a -> 'b t) -> 'b Map.t t
 end
-[@@inline always]
```
anmonteiro, line 104: callout: `@inline` annotations don't belong in interfaces. they are a property of the implementation: https://github.com/ocaml/ocaml/issues/12972#issuecomment-1938287079

anmonteiro: prior work: https://github.com/ocaml-dune/fiber/pull/25

otherlibs/stdune/src/stdune.mli
```diff
@@ -0,0 +1,145 @@
...
+module Hashtbl = Hashtbl
+module Table = Table
+module Int = Int
+module Id = Id
+module Io = Io
+module List = List
+module Map = Map
+module Option = Option
+module Or_exn = Or_exn
+module Ordering = Ordering
```
rgrinberg, line 25: Do we need this mli? It's a bit of a copy-paste that doesn't really define the interface in a more useful way than the .ml file?

anmonteiro: I think we need it if we want to include the unstable alert with it. Otherwise putting it in the `.ml` file triggers warning 53.

rgrinberg: Hmm, but why? Are we not allowed to use alerts in .ml files?

anmonteiro: You can check how they're rejected by 5.2 here: https://github.com/ocaml/ocaml/blob/1f3c26c89b87d9c610699253b605b3fbace2f070/testsuite/tests/warnings/w53.ml#L26

I pushed an alternative version that disables it for the alert attribute in the .ml file

anmonteiro:
Looks like this might be fixed upstream (see https://github.com/ocaml/ocaml/issues/13054), so I'll revert the stdune.ml changes altogether.

anmonteiro:
This is now just a removal of `[@inline ..]` in the interface file (see https://github.com/ocaml-dune/fiber/pull/25), so merging.

PR #10351, opened 2024-04-01
chore: remove bench/micro/fiber_bench*.ml

anmonteiro:
these are being tracked in https://github.com/ocaml-dune/fiber


rgrinberg:
`runner.sh` needs updating too

rgrinberg (approved):

PR #10353, opened 2024-04-01
test: reproduce wrong link flag if pkg-config not in $PATH

anmonteiro:
this was originally spotted in https://github.com/ygrek/ocaml-zstd/pull/9 when I couldn't understand why `zstd` wouldn't build.

- in this test, we have `(external_library_name libexample)` and `(build_flags_resolver pkg_config)`, however the `pkg-config` executable is not in `$PATH`
- dune seems to derive the linker flags from `external_library_name`, which is a bit weird with this failure mode. Is that expected?
    - should this rather be an error that Dune failed to resolve `pkg-config`?


test/blackbox-tests/test-cases/ctypes/lib-pkg_config-not-in-path.t/run.t
```diff
@@ -0,0 +1,28 @@
...
+  File "stubgen/dune", line 1, characters 0-384:
+   1 | (library
+   2 |   (name examplelib)
+   3 |   (flags (:standard -w -9-27))
+  ....
+  12 |       (instance Functions)
+  13 |       (functor Function_description))
+  14 |     (generated_entry_point C)))
+  ld: library not found for -llibexample
+  clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
```
anmonteiro, line 27: clang-16 is likely to be specific to my system

rgrinberg:
> should this rather be an error that Dune failed to resolve pkg-config?

Sure that would be fine. Would there be another way of specifying a default in case a package isn't found in pkg-config?

rgrinberg:
Closing for inactivity. But feel free to re-open or even just merge this test case if it's ready.

PR #10354, opened 2024-04-01
fix: redirect deprecated libraries resolution to the public lib DB

anmonteiro:
addresses this [comment](https://github.com/ocaml/dune/pull/10307#discussion_r1544964403)

### DIFF
--- a/src/dune_rules/lib.ml
+++ b/src/dune_rules/lib.ml
@@ -424,7 +424,7 @@ and resolve_result =
   | Invalid of User_message.t
   | Ignore
   | Redirect_in_the_same_db of (Loc.t * Lib_name.t)
-  | Redirect of db * Lib_id.t
+  | Redirect of db * (Loc.t * Lib_name.t)
 
 let lib_config (t : lib) = t.lib_config
 let name t = t.name
@@ -1153,7 +1153,7 @@ end = struct
   let handle_resolve_result db ~super = function
     | Ignore -> Memo.return Status.Ignore
     | Redirect_in_the_same_db (_, name') -> find_internal db name'
-    | Redirect (db', lib_id') -> resolve_lib_id db' lib_id'
+    | Redirect (db', (_, name')) -> find_internal db' name'
     | Found info ->
       let name = Lib_info.name info in
       instantiate db name info ~hidden:None
@@ -1173,7 +1173,7 @@ end = struct
       Memo.parallel_map candidates ~f:(function
         | Ignore -> Memo.return (Some Status.Ignore)
         | Redirect_in_the_same_db (_, name') -> find_internal db name' >>| Option.some
-        | Redirect (db', lib_id') -> resolve_lib_id db' lib_id' >>| Option.some
+        | Redirect (db', (_, name')) -> find_internal db' name' >>| Option.some
         | Found info ->
           Lib_info.enabled info
           >>= (function
@@ -1866,7 +1866,7 @@ module DB = struct
       | Invalid of User_message.t
       | Ignore
       | Redirect_in_the_same_db of (Loc.t * Lib_name.t)
-      | Redirect of db * Lib_id.t
+      | Redirect of db * (Loc.t * Lib_name.t)
 
     let found f = Found f
     let not_found = Not_found
@@ -1881,7 +1881,7 @@ module DB = struct
       | Found lib -> variant "Found" [ Lib_info.to_dyn Path.to_dyn lib ]
       | Hidden h -> variant "Hidden" [ Hidden.to_dyn (Lib_info.to_dyn Path.to_dyn) h ]
       | Ignore -> variant "Ignore" []
-      | Redirect (_, lib_id) -> variant "Redirect" [ Lib_id.to_dyn lib_id ]
+      | Redirect (_, (_, name)) -> variant "Redirect" [ Lib_name.to_dyn name ]
       | Redirect_in_the_same_db (_, name) ->
         variant "Redirect_in_the_same_db" [ Lib_name.to_dyn name ]
     ;;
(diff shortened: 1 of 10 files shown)


test/blackbox-tests/test-cases/lib-collision/lib-collision-public-same-public-name.t
```diff
@@ -23,12 +23,12 @@ different folders.
 Without any consumers of the libraries
 
   $ dune build
```
anmonteiro, line 25: this error becomes better: we now use the `loc` for the `(public_name ..)` field, which is the actual conflict.

jchavarri (approved):
I'm afraid I don't fully understand what the change in redirection to parent is fixing (besides the better locations in errors mentioned). But otherwise lgtm.

src/dune_rules/lib.ml
```diff
@@ -424,7 +424,7 @@ and resolve_result =
   | Invalid of User_message.t
   | Ignore
   | Redirect_in_the_same_db of (Loc.t * Lib_name.t)
-  | Redirect of db * Lib_id.t
+  | Redirect of db * (Loc.t * Lib_name.t)
```
rgrinberg, line 427: Isn't this constructor also used for redirecting public names to private names?

src/dune_rules/scope.ml
```diff
@@ -85,13 +85,13 @@ module DB = struct
       else Lib.DB.Resolve_result.not_found
     | Found lib -> Memo.return (Lib.DB.Resolve_result.found lib)
     | Deprecated_library_name lib ->
-      Memo.return (Lib.DB.Resolve_result.redirect_in_the_same_db lib)
+      Memo.return (Lib.DB.Resolve_result.redirect parent lib)
```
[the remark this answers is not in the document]
anmonteiro: I believe this change is just saving us 1 redirect, because the new public name will exist in the public libs DB. We'd redirect anyway at resolution time (following 2 redirects, now just 1)

anmonteiro:
> Isn't this constructor also used for redirecting public names to private names?

I think this might be true. We may have to specify whether we're redirecting to a name or a library ID.

rgrinberg:
Just to make sure we're on the same page, I think we need two constructors:

1. `Redirect of Lib_name.t * DB.t` used for implemented `(deprecated_library_stanza ..)`. The target db here should be the public db
2. `Redirect of Lib_id.t` used for implementing public to private redirects.

anmonteiro:
OK I added a new `Redirect_to_lib_id` constructor that handles the redirection for local stanzas.

I also added `visibility` in `Lib_id` to get the "best loc" for a stanza. This way we make sure to keep showing the correct loc for the public name clash case in tests.

src/dune_rules/lib.mli
```diff
@@ -100,7 +100,8 @@ module DB : sig
     val not_found : t
     val found : Lib_info.external_ -> t
     val to_dyn : t Dyn.builder
-    val redirect : db -> Lib_id.t -> t
+    val redirect : db -> Loc.t * Lib_name.t -> t
```
rgrinberg, line 103: How about renaming this one too for symmetry? E.g. `redirect_by_name`

src/dune_rules/lib.mli
```diff
@@ -100,7 +100,8 @@ module DB : sig
...
+    val redirect_to_lib_id : db -> Lib_id.t -> t
```
rgrinberg, line 104: `redirect_by_id` reads a bit better to me

src/dune_rules/lib_id.ml
```diff
@@ -18,7 +23,7 @@ module Local = struct
       | x -> x
     ;;
```
rgrinberg, line 24: I missed this from before, but this seems wrong. How is the comparison function ignoring the other elements?

src/dune_rules/scope.ml
rgrinberg, line 88: Is `parent` always the public library database here?

anmonteiro: yes. I'll move these functions more closely together and make that explicit in a comment

anmonteiro: it adds a bit of whitespace diff noise, but I did the above in https://github.com/ocaml/dune/pull/10354/commits/3e730834ed182f5bfaed4cc16c904abd23f795c7

src/dune_rules/lib_id.ml
anmonteiro: My thought was that, since this refers to `Local` library IDs, the `loc` would be more than enough to distinguish different lib_ids.

anmonteiro: The only field we're not comparing is `enabled_if`. We could, but I couldn't come up with a `compare` function for `Blang.t` -- there's only `Blang.equal`.

rgrinberg: > My thought was that, since this refers to Local library IDs, the loc would be more than enough to distinguish different lib_ids.

In that case, I would drop the `enabled_if` from the id as well.

src/dune_rules/scope.ml
```diff
@@ -73,131 +73,133 @@ module DB = struct
...
-                    , Redirect { loc = loc1; to_ = lib1; _ } )
-                  | ( Redirect { loc = loc1; to_ = lib1; _ }
-                    , Deprecated_library_name (loc2, lib2) )
-                  | ( Deprecated_library_name (loc1, lib1)
-                    , Deprecated_library_name (loc2, lib2) ) ->
-                    if Lib_name.equal lib1 lib2 then Ok r1 else Error (loc1, loc2)
+  let create_db_from_stanzas =
+    (* Here, [parent] is always the public_libs DB. Check the call to
+       [create_db_from_stanzas] below. *)
+    let resolve_found_or_redirect ~parent fr =
```
rgrinberg, line 79: How about just rename `~parent` to `~public_db` then?

rgrinberg (approved):

PR #10355, opened 2024-04-02
fix(ml-sources): check that a library is available before tracking it

anmonteiro:
Since #10307, dune started allowing libraries to share the same name.

Thus, checking `enabled_if` isn't enough for libraries; dune now needs to check whether the library is available before tracking it in ml_sources.

addresses https://github.com/ocaml/dune/pull/10307#issuecomment-2027828644

### DIFF
--- a/src/dune_rules/ml_sources.mli
+++ b/src/dune_rules/ml_sources.mli
@@ -27,13 +27,17 @@ type for_ =
       }
   | Melange of { target : string }
 
-val modules_and_obj_dir : t -> for_:for_ -> Modules.t * Path.Build.t Obj_dir.t
+val modules_and_obj_dir
+  :  t
+  -> libs:Lib.DB.t
+  -> for_:for_
+  -> (Modules.t * Path.Build.t Obj_dir.t) Memo.t
 
 (** Modules attached to a library, executable, or melange.emit stanza.*)
-val modules : t -> for_:for_ -> Modules.t
+val modules : t -> libs:Lib.DB.t -> for_:for_ -> Modules.t Memo.t
 
 (** Find out the origin of the stanza for a given module *)
-val find_origin : t -> Module_name.Path.t -> Origin.t option
+val find_origin : t -> libs:Lib.DB.t -> Module_name.Path.t -> Origin.t option Memo.t
 
 val empty : t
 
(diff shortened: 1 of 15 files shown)


anmonteiro:
This diff is way better without whitespace https://github.com/ocaml/dune/pull/10355/files?w=1

jchavarri (approved):

rgrinberg:
One issue with this PR is that it's introducing a library database look up before we're able to resolve the sources of a library. That now has the potential for cycles. E.g:

```
(rule
 (with-stdout-to foo "m_%{lib-available:bar"))
(library
 (name foo)
 (modules %{read:foo}))
```

It would be better to delay this availability check until the source is needed by somebody

anmonteiro:
> One issue with this PR is that it's introducing a library database look up before we're able to resolve the sources of a library. That now has the potential for cycles. E.g:
> 
> ```
> (rule
>  (with-stdout-to foo "m_%{lib-available:bar"))
> (library
>  (name foo)
>  (modules %{read:foo}))
> ```
> 
> It would be better to delay this availability check until the source is needed by somebody

This cycle is present even if I revert my diff. Are you suggesting that we're adding to it rather than causing it?

anmonteiro:
I changed the implementation to delay querying `Lib.DB.available_by_id` until we query for `modules_and_obj_dir`. The implementation is fairly minimal, but it makes `modules_and_obj_dir` return a `Memo.t` and an extra `Lib.DB.t` argument, which make the diff a bit larger.

anmonteiro:
@rgrinberg I think this is still waiting for another look.

rgrinberg:
I'll try and take a look again this week

src/dune_rules/ml_sources.mli
```diff
@@ -27,13 +27,17 @@ type for_ =
       }
   | Melange of { target : string }
 
-val modules_and_obj_dir : t -> for_:for_ -> Modules.t * Path.Build.t Obj_dir.t
+val modules_and_obj_dir
+  :  t
+  -> db:Lib.DB.t
```
rgrinberg, line 32: How about `libs`? `db` is too generic.

rgrinberg (approved):

PR #10273, opened 2024-03-15
chore: use 4.14.2 to run tests

emillon:
The CI uses `4.14.x` so will pick it up as well. No tests are broken.

PR #10358, opened 2024-04-02
Bump recommend OCaml version to 4.14.2

samoht:
Otherwise, the black-box tests are spammed with compiler messages on macOS due to newer versions of XCode. See: https://github.com/ocaml/ocaml/issues/12641


anmonteiro:
I think this is a duplicate of https://github.com/ocaml/dune/pull/10273 cc @EmileTrotignon

emillon:
duplicate of #10273

PR #10361, opened 2024-04-02
test: fix truncate on macOS

samoht:
Otherwise the test fails with:

```
$ truncate --size=-10 _build/.db
+|  truncate: illegal option -- -
+|  usage: truncate [-c] -s [+|-|%|/]size[K|k|M|m|G|g|T|t] file ...
+|         truncate [-c] -r rfile file ...
+|  [1]
```

### DIFF
--- a/test/blackbox-tests/test-cases/corrupt-persistent.t
+++ b/test/blackbox-tests/test-cases/corrupt-persistent.t
@@ -10,11 +10,10 @@
 
 Delete last 10 chars of the .db file to corrupt it
 
-  $ truncate --size=-10 _build/.db
+  $ truncate -s -10 _build/.db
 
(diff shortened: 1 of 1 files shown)


rgrinberg:
Could you describe the issue you're fixing a little more? I'm not sure what is being fixed here.

anmonteiro:
I believe BSD truncate supports only `-s`, while GNU truncate supports `-s` / `--size=`

samoht:
Woops, I've edited the description of the PR to add the error message that I got. Indeed, only `-s` (or not `--size`) is supported on macOS (and on I think on BSD*)

rgrinberg:
Sure, no harm in this. Although note that you won't get very far without the gnu tools anyway

PR #10366, opened 2024-04-02
fix: display both line numbers in error message location ranges.

pmwhite:
Upstreaming a fix we made internally


rgrinberg:
Changes look good. Could you add a CHANGES entry and promote the tests?

PR #10369, opened 2024-04-03
fix: `Impl` case of `Modules.map_user_written`

anmonteiro:
I'm not 100% sure if this is right, but I spotted that `fold_user_written` a few lines above (line 1129) ignores `t.vlib` while `map_user_written` ignores `t.impl`. One of them surely isn't right?


rgrinberg:
The change looks right to me. But given the lack of impact on the test suite, I'm now worried this is dead code anyway. Could you double check that we're executing this code path in at least some test?

anmonteiro:
I don't think it's dead code because there are places where we call `map_user_written`. We may just not have a test for it.

anmonteiro:
Might actually be dead-code after all. Check #10383 and its description.

rgrinberg (approved):

PR #10370, opened 2024-04-03
test: demonstrate a dependency cycle between modules field and rule

anmonteiro:
this might be the "rules in the same dir" / masking limitation thing that prevents us from generating melange js in the same directory, but I'm not sure?

### DIFF
--- a/test/blackbox-tests/test-cases/lib-available-dynamic-modules.t
+++ b/test/blackbox-tests/test-cases/lib-available-dynamic-modules.t
@@ -5,39 +5,37 @@ generates them dynamically in the same directory
   > (lang dune 3.13)
   > EOF
 
-  $ cat > dune << EOF
-  > (rule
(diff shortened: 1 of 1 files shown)


rgrinberg:
Yes, you should move the `rule` stanza to a different directory and then copy the file into your desired directory with a copy rule.

jchavarri:
Should this change be consolidated with `eif-library-cycle.t`?

rgrinberg:
Btw, I think we need to refine the repro further still. Roughly, what we need is:

1. libraries `x` and `y` in the same dir.
2. library `x` needs a dynamic list of modules. The dynamic list should exist in some separate directory to avoid the issue Etienne mentioned.
3. the dynamic list of modules needs to be generated by resolving `y` somehow. E.g. by running an executable that uses `y`

anmonteiro:
@rgrinberg I think the test now illustrates the case you described. Could you check again, please?

rgrinberg (approved):

Issue #10301, opened 2024-03-22
Cannot befriend Menhir and ocamllex with `include_subdirs unqualified`


hirrolot:
My `lib/dune` is the following:

```
(library (name foo))

(include_subdirs unqualified)

(menhir
 (modules parser)
 (flags --explain --dump))

(ocamllex lexer)
```

But the issue is that I have `parser.mly` and `lexer.mll` in the `surface/` directory. When running `dune build`, I receive the following error:

```
File "lib/dune", line 16, characters 0-52:
16 | (menhir
17 |  (modules parser)
18 |  (flags --explain --dump))
Error: No rule found for lib/parser.mly
File "lib/dune", line 20, characters 0-16:
20 | (ocamllex lexer)
     ^^^^^^^^^^^^^^^^
Error: No rule found for lib/lexer.mll
```

How do I tell Dune that it should search for `parser.mly` and `lexer.mll` in `lib/surface/`, not `lib/`?

My Dune version is `3.14.2`.

emillon:
I'm a bit surprised to see menhir pop up here because it should have been fixed in #8949. I'll try to repro, thanks.

anmonteiro:
added a reproduction here https://github.com/ocaml/dune/pull/10371

johnyob:
I also think `inline_tests` doesn't play nicely with `include_subdirs unqualified`. I encountered this while running tests [here](https://github.com/johnyob/dromedary/tree/main/test/typing). 

I get the following error:
```
File "test/typing/dune", line 4, characters 1-15:
4 |  (inline_tests)
     ^^^^^^^^^^^^^^
Uncaught exception:
  
  (Sys_error
   "typing/test_whitington.ml: No such file or directory")

Raised by primitive operation at Stdlib.open_in_gen in file "stdlib.ml", line 405, characters 28-54
Called from Stdlib.open_in in file "stdlib.ml" (inlined), line 410, characters 2-45
Called from Ppx_expect_runtime__Write_corrected_file.f in file "runtime/write_corrected_file.ml", line 35, characters 21-44
Called from Base__List.map in file "src/list.ml", line 433, characters 15-18
Called from Ppx_expect_runtime.(fun) in file "runtime/ppx_expect_runtime.ml", lines 90-99, characters 4-18
Called from Stdlib__List.map in file "list.ml", line 87, characters 15-19
[...]

PR #10371, opened 2024-04-03
test: repro #10301

anmonteiro:
reproes #10301


anmonteiro:
at least for the ocamllex case it's easy to see why the target is expected to be in the same dir as the dune file: https://github.com/ocaml/dune/blob/main/src/dune_rules/stanzas/rule_conf.ml#L215-L220

anmonteiro:
and for menhir:

https://github.com/ocaml/dune/blob/2a4a5bf6b904549dccd580904bff52e49a8ec8f2/src/dune_rules/menhir/menhir_rules.ml#L100-L103

emillon (commented):
thanks for the repro. 2 remarks:
- can you add a pointer to #10301?
- if these are 2 separate issues, can you create 2 separate tests?
thanks!

anmonteiro:
@emillon changed this to be just about menhir, and renamed the test case to point to #10301.

anmonteiro:
ping @emillon

emillon (approved):

PR #10379, opened 2024-04-03
doc: tour of the dune codebase

emillon:
This is a new document (well, group of documents) where we go through various places of the code base. The main path goes from the CLI, to decoding, to rule generation, to the engine (of which we don't say much); and then a detour by the libraries and tests.

### DIFF
--- a/doc/explanation/tour/decoding.rst
+++ b/doc/explanation/tour/decoding.rst
@@ -1,0 +1,69 @@
+Parsing of Dune Files
+---------------------
+
+Parsing ``dune`` files is done in two steps:
+
+- They are parsed as S-expressions using :file:`src/dune_sexp/parser.mli`;
+- Then they are decoded using :file:`src/dune_sexp/decoder.mli`. The result of
+  this decoding step is added to an extensible variant using a mechanism in
+  :file:`src/dune_lang/stanza.mli`.
+
+Instead of writing a parser or using pattern matching, we define decoders,
+which are abstract values of type ``'a Decoder.t`` (returning a value of type
+``'a``). These decoders are assembled using combinators. For example, we can
+use simple decoders to write a decoder for a record type. This decoder
+abstraction is monadic, but the applicative subset is sufficient for most
+decoders.
+
+As an example, here is how ``(copy_files)`` is parsed:
+
+`src/dune_rules/stanzas/copy_files.ml <https://github.com/ocaml/dune/blob/3.15.0/src/dune_rules/stanzas/copy_files.ml#L31-L50>`_
+  .. code-block:: ocaml
+    :linenos:
+    :lineno-start: 31
+
+    let long_form =
+      let check = Dune_lang.Syntax.since Stanza.syntax (2, 7) in
+      let+ alias = field_o "alias" (check >>> Dune_lang.Alias.decode)
+      and+ mode = field "mode" ~default:Rule.Mode.Standard (check >>> Rule_mode_decoder.decode)
+      and+ enabled_if = Enabled_if.decode ~allowed_vars:Any ~since:(Some (2, 8)) ()
+      and+ files = field "files" (check >>> String_with_vars.decode)
+      and+ only_sources =
+        field_o
+          "only_sources"
+          (Dune_lang.Syntax.since Stanza.syntax (3, 14) >>> decode_only_sources)
+      and+ syntax_version = Dune_lang.Syntax.get_exn Stanza.syntax in
+      let only_sources = Option.value only_sources ~default:Blang.false_ in
+      { add_line_directive = false
+      ; alias
+      ; mode
+      ; enabled_if
+      ; files
+      ; only_sources
+      ; syntax_version
+      }
+
+The fields are queried individually, and a record is built using all the
+intermediate results. This will automatically take care of generating "unknown
+field X," "duplicate field X," and similar error messages.
+
+Another interesting thing to note is that the fields are not decoded directly,
+but use the following pattern:
+
+.. code:: ocaml
+
+   Syntax.since Stanza.syntax (x, y) >>> decoder
+
+Let's unpack this: ``(>>>)`` will run a ``unit Decoder.t`` on the input before
+passing the input to an actual decoder. The first decoder can be used to
+implement a check and trigger an error in some cases.
+
+Here, it is used for versioning. For example the ``(copy_files)`` stanza
+started supporting ``(enabled_if``) in version 2.8. Decoding this field is
+protected by this ``since`` call: it means that if the language version in
+:doc:`/reference/dune-project/index` file is greater than 2.8. In particular,
+this ensures that the project can not be built with Dune versions older than
+``2.8.0``.
+
+Once decoding succeeds, various stanzas are turned into various types defined
+in :file:`src/dune_rules/stanzas/`.
--- a/doc/explanation/tour/engine.rst
+++ b/doc/explanation/tour/engine.rst
@@ -1,0 +1,15 @@
+The Engine
+==========
+
+The engine is the core, reusable part of Dune. It contains all the composable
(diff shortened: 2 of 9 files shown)


emillon:
rendered version here: https://dune--10379.org.readthedocs.build/en/10379/explanation/tour/index.html

anmonteiro (commented):
I think the explanations are clear and the structure sets up a good foundation to keep evolving. I left a few minor suggestions as I was reading through. Thanks for writing this up!

doc/explanation/tour/rule-generation.rst
```diff
@@ -0,0 +1,155 @@
...
+Here, the ``mdx_action`` that is set up is not just a single
+``Command.run_dyn_prog`` call: it is assembled using combinators from
+``Action_builder.With_targets``. This is another monad used in Dune. It
+corresponds to what can happen at build time, like running commands or creating
+files, or more complex actions such as reading a file that needs to be built by
+another rule. It is also used to track dependencies and targets. The "thing"
+that we register to the Dune engine using ``Super_context.add_rule`` has type
+``Action.Full.t Action_builder.With_targets.t``.
+
+This is different from ``Memo``, which corresponds to what happens within
```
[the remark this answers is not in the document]
emillon: thanks. I tried with a `note::` admonition (as well as the one on Memo before). what do you think?

doc/explanation/tour/rule-generation.rst
```diff
@@ -0,0 +1,155 @@
...
+---------------
+
+Using these parsed stanzas, the next step is to generate rules. This work
+starts in :file:`src/dune_rules/gen_rules.ml`, which dispatches to various
+modules in :file:`src/dune_rules/`.
+
+Often, modules will contain a ``gen_rules`` function taking:
+
+- the parsed stanza,
+- a ``Super_context.t`` value,
```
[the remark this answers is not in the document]
emillon: I reworked that part, let me know what you think

doc/explanation/tour/engine.rst
```diff
@@ -0,0 +1,13 @@
+The Engine
+==========
+
+The engine is at the core of Dune: it keeps track of the various directories
```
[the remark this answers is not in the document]
snowleopard: Not just "in theory"! That's exactly what we did at Jane Street, where we replaced `dune_rules` (and other parts of the "frontend" like CLI) with a completely different codebase, while keeping the backend.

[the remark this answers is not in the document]
emillon: thanks. I added this. we can expand later if necessary.

Leonidas-from-XIV:
I am leaving out some (minor) copyediting remarks. I think the structure is pretty good and the parts that are written are extremely useful, especially the rules part.

What I think I would like to know:

 * More about the libraries. Why are `stdune`, `dyn` etc separate libraries? `dune-configuratior` is meant to be used but what would guide us to make e.g. `xdg` a separate library?
 * Some more information on vendored libraries and how to deal with the forked versions. What policies do we have on forking, upstreaming etc
 * Expanders. I think the expanders are rather complex to understand and getting more complicated (partly in an effort to avoid cycles)
 * What kind of tests should go where? What is the reasoning behind why some tests are unit tests, some are cram tests etc.
 * How to keep this document up to date? Some things are probably going to change over time, what would be the best way to make sure that changes to the code will keep this document useful in the future?
 * Maybe something about the bootstrap process?

emillon:
Thanks for the remarks.
I'll try to define a scope for this PR, since we can always expand on this later. I think that's actually a good sign that a document prompts new ones!

I'll expand the index page with an exhaustive map of the code base, which points to the other pages of the tour, or adequate pages (for example, we already have a document dedicated to the bootstrap process, which would be pointed out by `boot`).

> How to keep this document up to date?

This is an important question. I think we should resist trying to update it at each version. My impression is that it should be enough to update it every year or so. To keep make it useful and easy to update I suggest the following:
- it is clear that it is more tied to a particular version than the rest of the documentation (admonition on index.rst)
- links are permalinks to the tagged release
- keep file names easy to change - I initially wanted all `:file:` links to be actual hyperlinks but this is going to improve friction when updating

Once we have this, we can have a round of copyediting and I believe that the first version could be merged.

emillon:
@christinerose this is all new content, readable in a rendered form [here](https://dune--10379.org.readthedocs.build/en/10379/explanation/tour/index.html). I think the content is ready, I'm interested to hear your comments regarding copy-editing, etc. Thanks!

doc/explanation/tour/rule-generation.rst
```diff
@@ -0,0 +1,168 @@
...
+defined in :file:`src/dune_rules/command.mli` where ``A`` refers to a plain
+string, and ``Dep`` to a string that should be interpreted as a dependency.
+Between that, and the ``~stdout_to`` parameter, it is enough for Dune to know
+about the rule's dependencies (what it will read) and its target (what it will
+produce).
+
+The second rule, that generates ``mdx_gen.ml-gen``, is similar: it is also done
+by calling ``Command.run_dyn_prog``.
+
+The third rule, to build the executable, calls ``Exe.build_and_link`` that is a
```
christinerose, line 90:
```suggestion
The third rule, to build the executable, calls ``Exe.build_and_link`` that is a
```
The third rule builds the executable and calls ``Exe.build_and_link`` that is a

doc/explanation/tour/rule-generation.rst
```diff
@@ -0,0 +1,168 @@
...
+corresponds to what can happen at build time, like running commands or creating
+files, or more complex actions such as reading a file that needs to be built by
+another rule. It is also used to track dependencies and targets. The "thing"
+that we register to the Dune engine using ``Super_context.add_rule`` has type
+``Action.Full.t Action_builder.With_targets.t``.
+
+.. note::
+
+  This is different from ``Memo``, which corresponds to what happens within
+  ``Dune`` itself. But it is also possible to use ``Memo`` from an
```
christinerose, line 134:
```suggestion
  Dune itself. But it is also possible to use ``Memo`` from an
```

If it happens within Dune itself, we’re talking about the build system in general, so I don’t think it belongs in monospace.

doc/explanation/tour/rule-generation.rst
```diff
@@ -0,0 +1,168 @@
...
+  ``Action_builder`` context. In that sense, ``Action_builder`` is more
+  powerful (at execution time, ``Action_builder`` will manage what happens in
```
christinerose, line 136:
```suggestion
  powerful! At execution time, ``Action_builder`` will manage what happens in
```
I think it has more impact taken out of the parenthesis, if you agree. It’s not wrong being in them, so your choice!

doc/explanation/tour/rule-generation.rst
```diff
@@ -0,0 +1,168 @@
...
+  the ``_build`` directory, while ``Memo`` is only concerned with
+  what happens in memory).
```
christinerose, line 138:
```suggestion
  what happens in memory.
```

doc/explanation/tour/decoding.rst
```diff
@@ -0,0 +1,68 @@
+Parsing of Dune Files
+---------------------
+
+Parsing ``dune`` files is done in two steps: first, they are parsed as
```
christinerose, line 4: There’s nothing incorrect about the sentence structure here, but I think it would be more readable/scannable if it was in a list: 

Parsing `dune` files is done in two steps:
1. They are parsed…
2. Then they are decoded…

But it’s totally up to you! Just my two cents 🙂

christinerose (commented):
Thanks for asking me to review!

It looks really good overall! I made some minor line editing changes, and I left a few comments/suggestions.

emillon:
Thanks everyone for your feedback. This now looks in good enough shape to be published.

PR #10380, opened 2024-04-03
add reproduction for regression with exec + watch

tatchi:
This demonstrates a regression when calling the `exec' command in watch mode from a child folder of the root folder.

So in short the following works fine

```
dune exec --watch /path/to/bin/main.exe
```

But not:

```
cd path && dune exec --watch ./to/bin/main.exe
```

In this case I get this error on Linux:

```
execve(_build/default/path/to/bin/main.exe): : No such file or directory
```

On MacOS, and according to the regression test, the error is different:

```
Error: posix_spawn(): No such file or directory
```


test/blackbox-tests/test-cases/exec-watch/exec-watch-basic.t/run.t
```diff
@@ -18,7 +18,7 @@ between each change to its code.
   1 | let () = print_endline "baz
                              ^
   Error: String literal not terminated
-  Had errors, waiting for filesystem changes...
+  Had 1 error, waiting for filesystem changes...
```
tatchi, line 21: Nothing to do with this PR, this is just a change that was never promoted

test/blackbox-tests/test-cases/exec-watch/exec-watch-multi-levels.t/run.t
```diff
@@ -0,0 +1,27 @@
...
+  $ ../wait-for-file.sh $DONE_FLAG
+  $ kill $PID
+
+
+It's broken when invoked in a child folder
+  $ cd bin && dune exec --root .. --watch ./bin/main.exe &
+  Entering directory '..'
+  Error: posix_spawn(): No such file or directory
+  Leaving directory '..'
+  $ sleep 0.5;
```
tatchi, line 21: We could probably come up with something better than `sleep`, but my point was to show the problem

tatchi:
You can see the error in Linux on CI: https://github.com/ocaml/dune/actions/runs/8545271412/job/23413184499?pr=10380#step:11:26

tatchi:
Reverting https://github.com/ocaml/dune/pull/10262 seems to fix the issue cc @gridbugs

gridbugs:
Investigating...

gridbugs:
Fixed in https://github.com/ocaml/dune/pull/10386. The problem was caused by relative paths to executables in the _build directory being resolved relative to the user's cwd rather than the project root. See the the commit message of https://github.com/ocaml/dune/pull/10386/commits/5038bbccb5f29c75e67600d344fd56302f89e7af for more info.

I've squashed your commits into a single commit with your signature attached and included it in my PR, mostly to add the tests you wrote.

Thanks for the report!

tatchi:
Thanks for the quick fix @gridbugs 🤗

Issue #10178, opened 2024-03-01
Feature: DUNE_PROJECT_SOURCE_PREFIX_MAP


jonahbeckford:
This is already implemented in https://github.com/jonahbeckford/dune/commit/812af6c58603ba77c7e26d4711f8ae1f189b441e. I'd like to get a feature review (and I have to wait for other things before I submit a PR for this). I can also demo this at a later date to the Dune team.

## Desired Behavior

The DUNE_PROJECT_SOURCE_PREFIX_MAP environment variable behaves similarly to the build path prefix map. However, it focuses on source path modifications rather than build path modifications.

The use cases I have for this are:
1. (today) autogenerating the `dune` build files into a single git-ignored directory tree, for a tidy project hygiene. (It would be nicer if that directory tree didn't have to exist at all, but Dune's `Path.t` abstraction forces `dune` files to be in the project directory). Those `dune` build files have `(copy)` stanzas to the original source code.
2. (soon) Major transformations of the `.ml` files. For example, a system (being implemented right now) where naming a file `xyz_windows_x86_64.ml` would be considered to be `xyz.ml` for a Windows 64-bit build. This is just a slight extension to the `(copy)` of the first use case, where filenames are adjusted during the `(copy)`.

This commit focuses on the Merlin experience.

## Example

Given a tiny project:

```
<project>/src/X/A.ml
<project>/dune-project <-- auto-generated and git-ignored
<project>/#s/X/dune <-- auto-generated and git-ignored
<project>/_build/default/#s/X/A.ml <-- (copy) of src/X/A.ml
```

Merlin should see a Merlin configuration for `src/X/A.ml` even though the `(executable)` was for `#s/X/A.ml`.

A `#s=src` DUNE_PROJECT_SOURCE_PREFIX_MAP accomplishes that.

Notice a few things:
1. The user only needs to enter `src/X/A.ml`. No dune file needs to be written.
2. Obvious drawback that there is an extra copy.
3. As the project grows large with `.ml` files, the generated `dune` files all stay in a single tree `#s` that is git-ignored.
4. I haven't shown it in that example but my use-cases are multiple script "roots" (like `src/`, `tests/`, `assets/bitmaps/`, etc.) to hold content.

rgrinberg:
If I'm getting this right, the major use case for this feature is to provide your own metadata that desugars into dune and dune-project files. If so, I don't think this proposal is enough because the error messages emitted by dune would still contain the generated paths in the error messages.

[...]

PR #10381, opened 2024-04-03
Add merlin locate tests and add build dirs of modules to Merlin so copy# is visible.

jonahbeckford:
This is a partial follow up to #10178

+bugfix: Adds the build directory of modules to the Merlin source directories so copy# can be followed. Without this modification (`merlin.ml:Unprocessed.process`) the test https://github.com/ocaml/dune/blob/2f47fed767aa191a9d8fc448b940c32421acf959/test/blackbox-tests/test-cases/merlin/merlin-locate.t/run.t#L39-L49
would be
> ```
>  $ cat ./$LIBNAME/b.ml | ocamlmerlin single locate -look-for ml -position 1:10 -filename b.ml | jq .value
>  "'Copy_of_a' seems to originate from 'Copy_of_a' whose ML file could not be found"
> ```

+Adds a test for basic [locate] command

+Adds a "copyroundtrip" test for a [locate] command whose source is from a [copy#] stanza
  * Upstream issue: `ocamlmerlin` should read copy# line header and report the original file not the copied file.

+Adds `merlin` package to TEST_DEPS so that `ocamlmerlin` binary can be used in tests

### DIFF
--- a/test/blackbox-tests/test-cases/melange/merlin.t
+++ b/test/blackbox-tests/test-cases/melange/merlin.t
@@ -40,6 +40,7 @@ Paths to Melange stdlib appear in B and S entries without melange.emit stanza
   S /MELC_STDLIB
   S /MELC_STDLIB
   S $TESTCASE_ROOT
+  S $TESTCASE_ROOT/_build/default
 
   $ target=output
   $ cat >dune <<EOF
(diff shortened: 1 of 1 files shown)


jonahbeckford:
What is missing: the last step needs to be something like making sure the final file is not `In_build_dir`.

Today, Merlin does not seem to respect the line header, so it give back a location in `In_build_dir`. That path should be skipped.

Tomorrow, when Merlin respects the line header and gives back a location in the source tree, that path should be used.

I *think* that is reasonable logic (although I'm not sure where it goes).

voodoos:
Adding directories with the copied source in Merlin configuration is not a correct way to solve that issue. The `S` configuration directive should only be used for directories containing source files directly written (and writable) by the users.

rgrinberg:

PR #10367, opened 2024-04-03
Allow rules with directory targets to be disabled

gridbugs:
Previously attempting to disable a rule with a directory target would cause dune to crash.

Fixes https://github.com/ocaml/dune/issues/10310

PR #10382, opened 2024-04-04
Allow rules with directory targets to be disabled (again)

### LINKED ISSUE #10310: Crash with directory targets and multiple workspace contexts

gridbugs:
Previously attempting to disable a rule with a directory target would cause dune to crash.

Fixes https://github.com/ocaml/dune/issues/10310

Follow up from https://github.com/ocaml/dune/pull/10367 which has been reverted. To address the problems with that fix:
 - `extract_directory_targets` now takes a `Expander.t Memo.t` and only evaluates it if the `enabled_if` field is present.
 - I remembered to include a changelog entry.

### DIFF
--- a/src/dune_rules/dir_status.ml
+++ b/src/dune_rules/dir_status.ml
@@ -85,23 +85,20 @@ let error_no_module_consumer ~loc (qualification : Include_subdirs.qualification
     ]
 ;;
 
-let extract_directory_targets ~dir ~expander stanzas =
+let extract_directory_targets ~dir stanzas =
   Memo.List.fold_left stanzas ~init:Path.Build.Map.empty ~f:(fun acc stanza ->
     match Stanza.repr stanza with
     | Rule_conf.T { targets = Static { targets = l; _ }; loc = rule_loc; enabled_if; _ }
       ->
-      let+ available =
-        match enabled_if with
-        | Blang.Const const -> Memo.return const
-        | _ ->
-          (* Only evaluate the expander if the enabled_if field is
-             non-trivial to avoid memo cycles. If the enabled_if field is
-             absent from the "rule" stanza then its value will be [Const
-             true]. *)
-          let* expander = expander in
-          Expander0.eval_blang expander enabled_if
-      in
-      (match available with
+      (match enabled_if with
+       | Blang.Const const -> Memo.return const
+       | _ ->
+         (* Only evaluate the expander if the enabled_if field is
+            non-trivial to avoid memo cycles. If the enabled_if field is absent
+            from the "rule" stanza then its value will be [Const true]. *)
+         let* expander = Expander0.get ~dir in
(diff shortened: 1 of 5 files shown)


src/dune_rules/dir_status.ml
```diff
@@ -85,32 +85,46 @@ let error_no_module_consumer ~loc (qualification : Include_subdirs.qualification
...
+        | Blang.Const const -> Memo.return const
+        | _ ->
+          (* Only evaluate the expander if the enabled_if field is
+             non-trivial to avoid memo cycles. If the enabled_if field is
+             absent from the "rule" stanza then its value will be [Const
+             true]. *)
+          let* expander = expander in
+          Expander.eval_blang expander enabled_if
+      in
+      if not available
```
rgrinberg, line 104: how about `match available`? I find reading `if not x else` to be a little difficult

gridbugs:
> The fix looks correct now. I think it can be improved a little more though. If you use `Expander0` instead of expander, you will avoid doing a little more work when we have no `enabled_if` on the rules which is the vast majority of rules.

I've made this suggested change, but can you explain why this helps. In the case when there is no `enabled_if` on a rule, the branch that evaluates the blang isn't taken, so why would using an `Expander0.t` make a difference?

Also I couldn't find any documentation on the difference between `Expander0` and `Expander`. In general when should one be preferred over the other?

anmonteiro:
hey @gridbugs. There's at least some context here: https://github.com/ocaml/dune/pull/10250

rgrinberg:
> I've made this suggested change, but can you explain why this helps. In the case when there is no enabled_if on a rule, the branch that evaluates the blang isn't taken, so why would using an Expander0.t make a difference?

The way the code is written, I suppose it doesn't make a difference. I imagined it to call `Expander0.get` from within `Dir_status` itself so that you don't even have to pass anything when there's no `enabled_if`.

> Also I couldn't find any documentation on the difference between Expander0 and Expander. In general when should one be preferred over the other?

You can use `Expander0` to expand things with introducing a module dependency on `Expander` and `Super_context`. That ends up being pretty useful to avoid OCaml's module dependency cycles. Otherwise, there's no difference.

rgrinberg (approved):

PR #10383, opened 2024-04-04
refactor: introduce `Modules.With_vlib.t`

anmonteiro:
testing an alternative for https://github.com/ocaml/dune/pull/10369:

- adding a `Modules.With_vlib.t`:
    1. reading a modules field and source modules produces `Modules.t`
    2. augmenting a `Modules.t` with virtual module implementations (`Vimpl.impl_modules`) produces a `Modules.With_vlib.t`
- most functions migrate to operate on the new type `Modules.With_vlib.t`
    - the distinction is roughly: `Compilation_context.t` operates on `Modules.With_vlib.t`; `Ml_sources.t` and `Dir_contents.t` operate on `Modules.t`
- the main benefit of this change is that `{map,fold}_user_written` / `fold_user_available` always operate on `Modules.t` and not on `Modules.With_vlib.t`.
    - which does make the code changed in #10369 effectively dead code
- we don't need `Modules.fold_no_vlib` anymore, so it's rather called `Modules.fold`
- `fold_no_vlib{,_with_no_aliases}` are only functions of the `Modules.With_vlib` module.

### DIFF
--- a/src/dune_rules/modules.ml
+++ b/src/dune_rules/modules.ml
@@ -1158,11 +1158,12 @@ module With_vlib = struct
     exists ~f:has
   ;;
 
-  let fold_no_vlib t ~init ~f =
-    match t with
-    | Modules t -> fold t ~init ~f
-    | Impl { vlib = _; impl; _ } -> fold impl ~f ~init
+  let drop_vlib = function
+    | Modules t -> t
+    | Impl { vlib = _; impl; _ } -> impl
   ;;
+
+  let fold_no_vlib t ~init ~f = fold (drop_vlib t) ~init ~f
 
   let fold_no_vlib_with_aliases =
     let group_of_alias t m =
--- a/src/dune_rules/modules.ml
+++ b/src/dune_rules/modules.ml
@@ -1163,8 +1163,6 @@ module With_vlib = struct
     | Impl { vlib = _; impl; _ } -> impl
   ;;
 
-  let fold_no_vlib t ~init ~f = fold (drop_vlib t) ~init ~f
-
   let fold_no_vlib_with_aliases =
     let group_of_alias t m =
       match t.modules with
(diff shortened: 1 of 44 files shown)


src/dune_rules/modules.ml
```diff
@@ -1119,17 +899,15 @@ let is_user_written m =
...
 
-let rec fold_user_written t ~f ~init =
-  let f m acc = if is_user_written m then f m acc else acc in
+let fold_user_available t ~f ~init =
   match t.modules with
   | Stdlib w -> Stdlib.fold w ~init ~f
   | Singleton m -> f m init
   | Unwrapped modules -> Unwrapped.fold modules ~init ~f
-  | Wrapped { group; _ } -> Group.fold group ~init ~f
-  | Impl { impl; vlib = _ } -> fold_user_written impl ~f ~init
```
anmonteiro, line 1129: dead code

src/dune_rules/modules.ml
```diff
@@ -1146,122 +924,437 @@ let rec map_user_written t ~f =
     | Wrapped ({ group; wrapped_compat = _; wrapped = _; toplevel_module = _ } as w) ->
       let+ group = Group.Memo_traversals.parallel_map group ~f in
       Wrapped { w with group }
-    | Impl t ->
```
anmonteiro, line 1149: dead code

rgrinberg (commented):
Seems OK to me in general. I find the choice of API makes things unnecessarily verbose for callers. It's almost always the case that we need to deal with virtual libraries, so it's a bit of a shame that we have to repeat this everywhere.

Also, is it possible to implement the various `_no_vlib` functions in `Modules.With_vlib` by adding a function:

```ocaml
module Modules : sig
  type t
  module With_vlib : sig
    type modules := t
    val drop_vlib  : t -> modules
  end
end
```

And then doing `Modules.With_vlib.drop_vlib |> Modules.fold`?

anmonteiro:
@rgrinberg Addressed the code review here:

- stopped exposing `Modules.With_vlib.fold` and changed it to `modules |> Modules.drop_vlib |> Modules.fold`
    - `Modules.With_vlib.fold_no_vlib_with_aliases` is still implemented using `Modules.fold` but the fucntion itself still needs to exist because it operates on the vlib module.
- I'm not sure what to do with the different constructors, let me know if you'd still like to see further changes

rgrinberg (approved):

PR #10385, opened 2024-04-04
pkg: user error when non-existent repo specified

gridbugs:
Previously it would be a code error if a user specified an invalid git repo url when pinning a package. This change makes it a user error instead.

### DIFF
--- a/src/dune_pkg/fetch.mli
+++ b/src/dune_pkg/fetch.mli
@@ -16,11 +16,11 @@ val fetch
   :  unpack:bool
   -> checksum:Checksum.t option
   -> target:Path.t
-  -> OpamUrl.t
+  -> url:Loc.t * OpamUrl.t
   -> (unit, failure) result Fiber.t
 
 val fetch_git
   :  Rev_store.t
   -> target:Path.t
-  -> OpamUrl.t
+  -> url:Loc.t * OpamUrl.t
   -> (unit, failure) result Fiber.t
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -149,6 +149,32 @@ let git_code_error ~dir ~args ~exit_code ~output =
     ]
 ;;
 
+module Git_error = struct
+  type t =
+    { dir : Path.t
+    ; args : string list
+    ; exit_code : int
+    ; output : string list
+    }
+
+  let raise_code_error { dir; args; exit_code; output } =
+    let git = Lazy.force Vcs.git in
+    Code_error.raise
+      "git returned non-zero exit code"
+      [ "exit code", Dyn.int exit_code
+      ; "dir", Path.to_dyn dir
+      ; "git", Path.to_dyn git
+      ; "args", Dyn.list Dyn.string args
+      ; "output", Dyn.list Dyn.string output
+      ]
+  ;;
+
+  let result_get_or_code_error = function
+    | Ok x -> x
+    | Error t -> raise_code_error t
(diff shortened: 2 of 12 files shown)


boot/libs.ml
```diff
@@ -1,3 +1,5 @@
+let executables = [ "main" ]
+
```
Leonidas-from-XIV, line 2: This change looks to be unrelated?

src/dune_pkg/rev_store.ml
```diff
@@ -157,28 +183,29 @@ let run_with_exit_code { dir; _ } ~allow_codes ~display args =
...
-  then exit_code
-  else git_code_error ~dir ~args ~exit_code ~output:[]
+  then Ok exit_code
+  else Error { Git_error.dir; args; exit_code; output = [] }
 ;;
 
 let run t ~display args =
-  run_with_exit_code t ~allow_codes:(Int.equal 0) ~display args >>| ignore
+  run_with_exit_code t ~allow_codes:(Int.equal 0) ~display args
+  >>| Result.map ~f:(fun (_ : int) -> ())
```
Leonidas-from-XIV, line 192: I know there is some use in annotating the expected type to be ignored but it's also adding a lot of noise, so maybe

```suggestion
  >>| Result.map ~f:ignore
```

src/dune_pkg/rev_store.ml
```diff
@@ -724,12 +758,32 @@ let remote =
...
         Fiber_lazy.create (fun () ->
-          let+ hits = run_capture_lines t ~display:!Dune_engine.Clflags.display command in
+          let+ hits =
+            run_capture_lines t ~display:!Dune_engine.Clflags.display command
+            >>| function
+            | Ok lines -> lines
+            | Error git_error ->
+              (match git_error.exit_code with
+               | 128 ->
+                 let command_string = String.concat ~sep:" " command in
```
Leonidas-from-XIV, line 773: This will generate an invalid command if any of the items in `command` have a space (which I guess a URL could have, unless it is URL-encoded). Maybe better always quote the second item to be on the safe side.

Leonidas-from-XIV (commented):
It's a useful change, my main suggestion is to have the URL and the Loc tied closer together.

boot/libs.ml
Leonidas-from-XIV: Yes, I know the problem :-) Usually I can get this fixed when I rebootstrap from `main` with `make bootstrap`.

src/dune_pkg/fetch.mli
```diff
@@ -17,10 +17,12 @@ val fetch
   -> checksum:Checksum.t option
   -> target:Path.t
   -> OpamUrl.t
+  -> url_loc:Loc.t
```
rgrinberg, line 20: When it's not clear what which argument the location pertains to, dune's convention is usually to tuple it. So in this case, it should be `(Loc.t * OpamUrl.t)`.

src/dune_pkg/rev_store.ml
rgrinberg: It's pretty useful to know what's being ignored. I think the noise makes it worth it here. You could also do `Result.map ~f:(ignore : int -> unit)` I suppose

gridbugs: Updated to use the `ignore` function as I think it's clearer.

rgrinberg (approved):

Leonidas-from-XIV (approved):

PR #10386, opened 2024-04-04
Fix locating program in exec watch mode

gridbugs:
This fixes a bug where running an executable from the current project
with `dune exec --watch` would be unable to find the executable unless
the command was run from the project's root directory.

The problem was introduced when we started setting the cwd of processes
spawned by exec in watch mode to the user's current directory
(https://github.com/ocaml/dune/pull/10262). If the program argument to
exec refers to a file to be built in the current project (such as an
executable implemented in the current project) then the path to the
executable to spawn will be a path inside the _build directory relative
to the project root. Since the cwd of the new process was set to the
user's current directory, this relative path was being resolved within
the current directory, whereas it should have been resolved relative to
the project root.

The fix was to convert relative paths into absolute paths relative to
the project root (this was already being done for exec when not in watch
mode).

### DIFF
--- a/bin/import.ml
+++ b/bin/import.ml
@@ -235,14 +235,13 @@ module Scheduler = struct
   ;;
 end
 
-let restore_cwd_and_execve (common : Common.t) prog argv env =
-  let prog =
-    if Filename.is_relative prog
-    then (
-      let root = Common.root common in
-      Filename.concat root.dir prog)
-    else prog
-  in
+let resolve_relative_path_within_root_dir (root : Workspace_root.t) path =
+  if Filename.is_relative path then Filename.concat root.dir path else path
+;;
+
+let restore_cwd_and_execve common prog argv env =
+  let root = Common.root common in
+  let prog = resolve_relative_path_within_root_dir root prog in
   Proc.restore_cwd_and_execve prog argv ~env
 ;;
 
--- a/bin/exec.ml
+++ b/bin/exec.ml
@@ -77,22 +77,25 @@ module Command_to_exec = struct
 
   (* Helper function to spawn a new process running a command in an
      environment, returning the new process' pid *)
-  let spawn_process path ~args ~env =
+  let spawn_process root prog ~args ~env =
     let pid =
-      let path = Path.to_string path in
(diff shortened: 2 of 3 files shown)


test/blackbox-tests/test-cases/exec-watch/exec-watch-multi-levels.t/run.t
```diff
@@ -0,0 +1,34 @@
+Test the behaviour of exec watch mode when the current directory is not the
```
emillon, line 1: This is a repro for this issue, right? can you land it first in its own PR?

bin/import.ml
```diff
@@ -239,14 +239,16 @@ module Scheduler = struct
...
 
-let restore_cwd_and_execve (common : Common.t) prog argv env =
-  let prog =
-    if Filename.is_relative prog
-    then (
-      let root = Common.root common in
-      Filename.concat root.dir prog)
-    else prog
-  in
+let resolve_relative_path_within_root_dir (common : Common.t) path =
```
anmonteiro, line 242: you're threading `Common.t` all the way here just to get `Common.root`. how about we just take `~root` instead?

gridbugs:
> what happens with `cd path && dune exec -- echo hello`?
> 
> IIUC, `echo` would pass the `Filename.is_relative` check and get a prefix when it shouldn't?

`dune exec` converts paths with no slashes into absolute paths by looking them up in PATH, so any program names will be represented by their absolute path here.

test/blackbox-tests/test-cases/exec-watch/exec-watch-multi-levels.t/run.t
gridbugs: Made a new pr here: https://github.com/ocaml/dune/pull/10391

bin/import.ml
gridbugs: fixed

gridbugs:
Just making sure that we don't forget to merge this. Are any other changes necessary?

rgrinberg:
Fixed in another PR

PR #10391, opened 2024-04-05
Add repro for exec watch mode bug

gridbugs:
This adds a test reproducing a bug with exec watch mode where dune is unable to find an executable if it's specified with a relative path when the user's cwd is not the project root.

Note that the watch mode tests are disabled by default as they are flakey.


test/blackbox-tests/test-cases/exec-watch/exec-watch-multi-levels.t/bin/main.ml
```diff
@@ -0,0 +1,9 @@
+let () =
+  let touch path =
+    let fd = Unix.openfile path [ Unix.O_CREAT ] 777 in
```
emillon, line 3:
```suggestion
    let fd = Unix.openfile path [ Unix.O_CREAT ] 0o777 in
```

gridbugs:
@emillon I think this slipped through the cracks. Think it's good to merge?

emillon (approved):

PR #10402, opened 2024-04-07
refactor: evaluate directory targets enabled_if in parallel map

rgrinberg:


test/blackbox-tests/test-cases/coq/coqdoc-dir-target-clash.t/run.t
```diff
@@ -1,20 +1,18 @@
 We try to build the documentation but there will be a clash between the
 directory targets. Notice how the tex one fails before html one.
```
anmonteiro, line 2: > Notice how the tex one fails before html one.

This is no longer true?

Will this test be deterministic with `parallel_map`?

anmonteiro (approved):

PR #10407, opened 2024-04-10
pkg: add test updating PATH with withenv

gridbugs:
Adds a repro for https://github.com/ocaml/dune/issues/10408


rgrinberg (approved):

gridbugs:
@rgrinberg can you clarify whether this behaviour is expected or if it's a bug

rgrinberg:
I think it's a bug. This behavior differs from opam, right? In which case, it would certainly be unintentional on my part.

gridbugs:
Confirming that this is not how opam behaves.

PR #10411, opened 2024-04-11
test(melange): show dependency missing bug with virtual libraries

anmonteiro:
reproduces https://github.com/melange-re/melange/issues/1105

PR #10412, opened 2024-04-11
fix(melange): during JS emission, depend on virtual lib implementations

anmonteiro:
fixes the test in https://github.com/ocaml/dune/pull/10411

### DIFF
--- a/src/dune_rules/melange/melange_rules.ml
+++ b/src/dune_rules/melange/melange_rules.ml
@@ -484,6 +484,22 @@ let setup_js_rules_libraries =
   in
   fun ~dir ~scope ~target_dir ~sctx ~requires_link ~mode (mel : Melange_stanzas.Emit.t) ->
     let build_js = build_js ~sctx ~mode ~module_systems:mel.module_systems in
+    let with_vlib_implementations =
+      let vlib_implementations =
+        (* vlib_name => concrete_impl *)
+        List.fold_left requires_link ~init:Lib_name.Map.empty ~f:(fun acc dep ->
+          match Lib_info.implements (Lib.info dep) with
+          | None -> acc
+          | Some (_, vlib_name) -> Lib_name.Map.add_exn acc vlib_name dep)
+      in
+      fun lib deps ->
+        (* Depend on the concrete implementations of virtual libraries so
+           that Melange can find their `.cmj` files. *)
+        List.fold_left deps ~init:deps ~f:(fun acc dep ->
+          match Lib_name.Map.find vlib_implementations (Lib.name dep) with
+          | None -> acc
+          | Some sub -> if Lib.equal sub lib then acc else sub :: acc)
+    in
     Memo.parallel_iter requires_link ~f:(fun lib ->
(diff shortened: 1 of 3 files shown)


src/dune_rules/melange/melange_rules.ml
```diff
@@ -502,6 +509,13 @@ let setup_js_rules_libraries =
       let* includes =
         let+ requires_link =
           Memo.Lazy.force (Lib.Compile.requires_link lib_compile_info)
+          |> Resolve.Memo.map ~f:(fun deps ->
```
jchavarri, line 512: iiuc using `map` will effectively replace the vlib dep with its implementation, for the sake of the cmj deps. But what would happen if the vlib itself contains some `ml` files?

Edit: never mind, I see below that we start from the existing `deps` and the addition of the implementations is purely additive.

src/dune_rules/melange/melange_rules.ml
```diff
@@ -484,6 +484,13 @@ let setup_js_rules_libraries =
   in
   fun ~dir ~scope ~target_dir ~sctx ~requires_link ~mode (mel : Melange_stanzas.Emit.t) ->
     let build_js = build_js ~sctx ~mode ~module_systems:mel.module_systems in
+    let vlib_implementations =
```
jchavarri, line 487: Maybe instead of getting the `vlib_implementations` here and below define some inline callback that uses them, I would create a `with_vlib_implementations` function here that takes the `deps` arg, so that it can be used directly below, e.g.:

```ml
          |> Resolve.Memo.map ~f:with_vlib_implementations
```

jchavarri (approved):

src/dune_rules/melange/melange_rules.ml
anmonteiro: but then you'd be creating it on every iteration of the loop, no?

jchavarri: No, I mean something like this, in line 486-487 (outside the loop):

```ml
  fun ~dir ~scope ~target_dir ~sctx ~requires_link ~mode (mel : Melange_stanzas.Emit.t) ->
    let build_js = build_js ~sctx ~mode ~module_systems:mel.module_systems in
    let with_vlib_implementations deps = ...all the new stuff..
```

Then in the loop just this

```ml
          |> Resolve.Memo.map ~f:with_vlib_implementations
```

jchavarri: It's pretty minor in any case.

anmonteiro: check what I've pushed. is that what you had in mind?

jchavarri: Exactly! 👍

anmonteiro:
@emillon @Leonidas-from-XIV may this be considered for a patch release, if you're planning on that?

anmonteiro:
Strictly a fix

emillon:
good for 3.15.1 in my book, then!

PR #10413, opened 2024-04-11
[xdg] provide macOS known directories

### LINKED ISSUE #10398: Xdg support: directories used for macOS
Currently, XDG uses the same behavior for macOS and other Unix-like operating system. However, if one refers to the [Apple Guidelines on directories organization](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html#//apple_ref/doc/uid/TP40010672-CH10-SW1), this should not be the case.

Xdg provides (I excluded home and runtime dir which are not important here):
- `XDG_DATA_HOME` (as `data_dir`) -> `~/.local/share`
- `XDG_CONFIG_HOME` (as `config_dir`) -> `~/.config`
- `XDG_CACHE_HOME` (as `cache_dir`) -> `~/.cache`
- `XDG_STATE_HOME`  (as `state_dir`) -> `~/.local/state`

The macOS hierarchy is quite different and in fact, it is closer to the behavior that is currently implemented for Windows. As far as I understand the XDG standard and the Apple Documentation, we should have:

- `XDG_DATA_HOME` (as `data_dir`) -> `~/Library/Application Support` (W: LocalAppData)
- `XDG_CONFIG_HOME` (as `config_dir`) -> `~/Library/Application Support` (W: LocalAppData)
- `XDG_CACHE_HOME` (as `cache_dir`) -> `~/Library/Application Support` (W: LocalAppData)
- `XDG_STATE_HOME`  (as `state_dir`) -> `~/Library/Caches` (W: InternetCache)

The main problem is that fixing this would basically break packages that use XDG. On the other side, it would improve compatibility with macOS tools (for example, automatic backups on macOS ignore `~/Library/Caches`). Also, I am wondering what is the best way to detect that we use macOS.

AllanBlanchard:
Fix #10398 

Add `?macos:bool` argument to `Xdg.create`. When enabled, `Xdg` uses known macos directories, that is:
- `~/Library/Caches` for cache,
- `~/Application Support` otherwise.

Implication of each default:
- false: preserves current behavior but inconsistent with `?win32` default behavior,
- true: consistent with `?win32` default, but macOS users of packages that use `Xdg` will potentially lose their config.


AllanBlanchard:
@rgrinberg I let you decide of the default, or even whether you want to introduce the feature or not.

otherlibs/xdg/xdg.mli
```diff
@@ -21,6 +21,9 @@ val state_dir : t -> string
 val runtime_dir : t -> string option
 
 (** Constructor of type {!t}. [~win32] (default: {!Sys.win32}) determines
-    whether to use Win32-specific APIs. [~env] is the function to get
+    whether to use Win32-specific APIs. [~macos] (default: [false]) determines
+    whether to use macOS-specific directories. [~env] is the function to get
     environment variables, typically {!Sys.getenv_opt}. *)
-val create : ?win32:bool -> env:(string -> string option) -> unit -> t
+val create :
+  ?win32:bool -> ?macos:bool ->
```
rgrinberg, line 28: It's weird to allow passing `~win32:true` and `~macos:true` at the same. Could you add a separate function for macos instead?

AllanBlanchard: Ok, I just realize that I didn't understand *at all* how the library is meant to be used :sweat_smile: . Now I get it, I'll change that. Thank you.

Leonidas-from-XIV:
While I understand the appeal of using the macOS standard directories and it might be sensible for Dune itself, it seems incorrect in the context of XDG specifically.

Quoting the specification:

>  `$XDG_CACHE_HOME` defines the base directory relative to which user-specific non-essential data files should be stored. If `$XDG_CACHE_HOME` is either not set or empty, a default equal to `$HOME/.cache` should be used.

Thus if we return `$HOME/Library/Caches` on unset `$XDG_CACHE_HOME` that would violate the specification.

jonahbeckford:
I get @Leonidas-from-XIV 's concern, although I think the use of this library has been to map to the conventions on the operating systems. If not, you could not use the XDG library on Windows.

Maybe we could disambiguate: use a `~strict` option if you need strict compliance of XDG? Of course, it would be highly inappropriate to do a strict reading of the spec on Windows, but at least if something really needs full XDG compliance they can get it.

*I am really interested in the conventional macOS directories; it is 74 apps+variants in `~/Library/Application Support` versus 19 in `~/.config` on my machine ... and 3 of those 19 are (wrongly IMHO) influenced by the `Xdg` library.*

rgrinberg:
"strict" I think is too vague and open to interpretation. I wouldn't mind constructors for `Xdg.t` that interpret the spec differently. In the end, all that matters for us is how dune interacts with the spec, and not what other users of the xdg library want their programs to do. So if it's useful for them to tweak the standard for their uses, I don't see a problem with it.

rgrinberg:
I'm closing this PR since it seems like you abandoned it. FTR, the changes seem fine to me and I don't mind changing the default on macOS.

Feel free to reopen if you still want this.

PR #10414, opened 2024-04-11
Show the context name for errors happening in "alt" contexts

### LINKED ISSUE #10378: Include the context name for errors that happen in the non-default context
Consider this example (it uses Melange but that's only for demo purposes, could happen in any multi-context setup):

```
  $ cat > dune-project << EOF
  > (lang dune 3.13)
  > (using melange 0.1)
  > (package (name bar) (allow_empty))
  > (package (name baz) (allow_empty))
  > EOF

  $ cat > dune-workspace << EOF
  > (lang dune 3.13)
  > 
  > (context default)
  > 
  > (context
  >  (default
  >   (name melange)))
  > EOF
  $ cat > dune << EOF
  > (library
  >  (name foo)
  >  (public_name bar.foo)
  >  (enabled_if (= %{context_name} "default")))
  > (library
  >  (name foo)
  >  (public_name baz.foo)
  >  (modes melange)
  >  (enabled_if (= %{context_name} "melange")))
  > EOF

  $ cat > foo.ml <<EOF
  > let t = Str.regexp
  > EOF

  $ dune build
  File "foo.ml", line 1, characters 8-18:
  1 | let t = Str.regexp
              ^^^^^^^^^^
  Error: Unbound module Str
  [1]
```

The error message does not allow the user to know under which context the error is happening. This can be quite confusing when using more than one context at scale.

I wonder if Dune could include the context name for "alt contexts" (non-default), similarly to what is done currently for `display=short`, e.g.:

```
  $ dune build
  [melange]
  File "foo.ml", line 1, characters 8-18:
  1 | let t = Str.regexp
              ^^^^^^^^^^
  Error: Unbound module Str
  [1]
```

cc @davesnx

jchavarri:
Fixes #10378.

Initially I started looking into adding the changes at a later stage (`Build_cmd`, `Report_error`) but at that moment it seems it's already "too late" and I couldn't find a way to get the context where the process is being executed from. In the end I am resorting 

The print function doesn't add any new information if the context name is `default`. Maybe there could be some opt-in configuration to enable default logging in all cases added in a future PR cc @davesnx.

Given the pervasiveness of `User_message.make`, I assume this initial PR is missing a lot of places where it'd be interesting to print the context name as well. But I thought it'd make things easier for reviewers to start with a small PR, and then follow up later on with other PRs that will include any improvements over it.


rgrinberg:
We're actually trying to get rid of contexts from the engine and make them strictly rule only, so this PR would make it more more difficult for us.

I think we pass the directory with some user messages. So you could try extracting in `Report_error` and deriving the context from that.

jchavarri:
@rgrinberg is https://github.com/ocaml/dune/pull/10414/commits/63f17d133763693cfd77c048ae7aa960c136554a what you had in mind? I tried moving the `with-directory` annotation function to `User_message` to simplify things further, but this creates a cycle between `User_message` and `Path`. I assume that's the reason why it was in `Process` to begin with.

jchavarri:
> We're actually trying to get rid of contexts from the engine and make them strictly rule only, so this PR would make it more more difficult for us.

Ftr: https://github.com/ocaml/dune/issues/7049.

otherlibs/stdune/src/user_message.ml
```diff
@@ -226,6 +227,12 @@ let pp { loc; paragraphs; hints; annots = _ } =
            (Pp.textf "File %S, %s, characters %d-%d:" start.pos_fname lnum start_c stop_c))
       :: paragraphs
   in
+  let paragraphs =
+    match context with
+    | None | Some "default" | Some ".sandbox" -> paragraphs
```
emillon, line 232: If you can pass a `Context_name.t`, you could use `Context_name.is_default` here. There are other non-context directories in `_build`, like `.actions`, not sure if they need special handling.

jchavarri: I don't think I can use `Context_name` in `User_message`, because the latter exists in stdune and the former in dune_engine. Unless we move `Context_name` from dune_engine to stdune, but I am assuming if the plan is to get rid of `Context_name` in [the engine](https://github.com/ocaml/dune/pull/10414#issuecomment-2050294373), then it doesn't make sense to keep it in stdune? I'm not sure i'm understanding the whole thing.

emillon: Ah, sorry I missed this.
Maybe it could work to define a type like `No_context | Default | Other_context of string` and use it here? But then it means that `dune_util` needs to parse the context name.

jchavarri: Sorry, I am not sure I understand. `dune_util` depends on `stdune`. If we define a type in `dune_util`, how could it be used in `stdune` without getting cycles? Is the suggestion to move `User_message` over to `dune_util`?

> Maybe it could work to define a type like

I am defining a similar type in the related PR #10324: 

https://github.com/ocaml/dune/blob/6c2baf925e9bd07377443aecac1c540f8e756b0f/bin/ocaml/ocaml_merlin.ml#L14-L16

emillon: ok, nevermind the suggestion then. this is not a hard blocker for me

rgrinberg:
> @rgrinberg is [63f17d1](https://github.com/ocaml/dune/commit/63f17d133763693cfd77c048ae7aa960c136554a) what you had in mind? I tried moving the `with-directory` annotation function to `User_message` to simplify things further, but this creates a cycle between `User_message` and `Path`. I assume that's the reason why it was in `Process` to begin with.

I'm going to double check this tomorrow, but I think you can just get rid of that annotation and just include `dir` inside `User_message`.

jchavarri:
> I think you can just get rid of that annotation and just include `dir` inside `User_message`.

I tried adding a new field `dir : Path.t option` to `User_message.t`, but I get some dep cycle error:

```
Error: dependency cycle between modules in
_build/default/otherlibs/stdune/src:
   User_warning
-> Path
-> Path
-> required by _build/default/otherlibs/stdune/src/stdune.cmxa
-> required by
   _build/default/otherlibs/dune-rpc-lwt/examples/rpc_client/rpc_client.exe
-> required by _build/install/default/bin/dune-rpc-lwt-example-client
-> required by _build/default/dune-rpc-lwt.install
-> required by alias install
```

I don't get it very well because `User_warning` doesn't really use `Path`.

emillon:
I think there's a cycle even without `User_warning` since in the original code:

- `Path` depends on `User_error` through `User_error.raise` calls
- `User_error` depends on `User_message` through the definiton of `E`

```mermaid
graph LR
User_error -- User_error.raise --> Path
User_message -- User_error.t --> User_error
```

rgrinberg:
That's a bit of a pain. I think you can just use a string path though.

jchavarri:
> I think you can just use a string path though.

Tried something in https://github.com/ocaml/dune/pull/10414/commits/550d20e5f33efdc58ff238cca727927ea30321db.

jchavarri:
Tests are failing for unrelated reason, see https://github.com/ocaml/dune/pull/10484.

anmonteiro (approved):

rgrinberg:
@pmwhite can you check if this change is going to work internally?

pmwhite:
@rgrinberg I just wrote a patch to import this PR internally. It mostly applies without changes, except for a few places (e.g. `action_runners.ml`, `process.ml`, and `build_system_error.ml`) where we already had access to the directory through a `job_summary` field that has yet to be upstreamed. These places are not very significant, and all our internal tests pass, so this PR is :+1: with respect to not conflicting with internal changes.

rgrinberg:
Ok, thanks. In general, I'm quite happy to get rid of this "annots" map and just shove things directly into the user error type. Once the internal patch is merged, could you sync this PR with it and we can merge?

PR #10415, opened 2024-04-11
fix(melange): treat private libraries with `(package ..)` as public libs

anmonteiro:
part of #7164 

fixes the test added in https://github.com/ocaml/dune/pull/7652


test/blackbox-tests/test-cases/melange/emit-private.t
```diff
@@ -47,8 +47,8 @@ Test dependency on a private library in the same package as melange.emit
   > EOF
 
   $ OCAMLPATH=$PWD/prefix/lib/:$OCAMLPATH dune build @dist --display=short 2>&1 | grep -v melange
-          melc b/dist/a/a.js
-          melc b/dist/a/foo.js
+          melc b/dist/node_modules/a.__private__.a/a.js
+          melc b/dist/node_modules/a.__private__.a/foo.js
```
jchavarri, line 51: These paths were a bit confusing for me originally because the pkg name is the same as the library. Maybe the package could be renamed to `pkg` for clarity.

src/dune_rules/compilation_context.ml
```diff
@@ -130,7 +130,7 @@ let create
   ?stdlib
   ~js_of_ocaml
   ~package
-  ?public_lib_name
+  ~melange_package_name
```
jchavarri, line 133: What was behind the decision of changing from optional param to non-optional?

jchavarri (approved):

src/dune_rules/compilation_context.ml
anmonteiro: To make it explicit where we’re passing the argument

PR #10416, opened 2024-04-12
fix(melange): add `.cmj` include dirs for installed private libs

anmonteiro:
Melange still needs to find the module paths for these, which are stored in the `.cmj` files.


anmonteiro:
OK I guess this stops hiding private modules. I guess we can hide the `.cmj` from compilation and make it available at emission time. I'll change `Lib_mode.t` to read ``| Melange of [ | `Compilation | `Emission ]`` and add the obj_dir directories accordingly.

test/blackbox-tests/test-cases/melange/installed-private-lib-dep.t
```diff
@@ -65,11 +65,8 @@ An issue similar to #7104 still present because the `.cmj` is not visible.
   $ OCAMLPATH=$PWD/prefix/lib/:$OCAMLPATH dune build @melange --root app --display=short
```
jchavarri, line 65: The comment in line 63 can be updated.

jchavarri (approved):

anmonteiro:
I made the fix a lot less invasive and local to `Lib_flags` instead.

PR #10419, opened 2024-04-12
fix(melange): better error message for nested `melange.emit` stanzas

anmonteiro:
in the hint, use simpler language and suggest something concrete.

### DIFF
--- a/test/blackbox-tests/test-cases/melange/nested-melange-emit-stanzas.t
+++ b/test/blackbox-tests/test-cases/melange/nested-melange-emit-stanzas.t
@@ -30,5 +30,5 @@ Make sure an error is returned if trying to nest `melange.emit` stanzas
   Error: melange.emit stanzas cannot be nested
   - a/dune:1
   - a/output/b/dune:1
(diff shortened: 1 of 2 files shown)


test/blackbox-tests/test-cases/melange/nested-melange-emit-stanzas.t
```diff
@@ -30,5 +30,5 @@ Make sure an error is returned if trying to nest `melange.emit` stanzas
   Error: melange.emit stanzas cannot be nested
   - a/dune:1
   - a/output/b/dune:1
-  Hint: Move the melange.emit stanza from a/output/b to at least the level of a
+  Hint: Move both `melange.emit' stanzas to the same dune file
```
jchavarri, line 33: The could also coexist in sibling folders right? Not sure if worth mentioning.

jchavarri (approved):

test/blackbox-tests/test-cases/melange/nested-melange-emit-stanzas.t
anmonteiro: that's a good point. I'll change it up.

PR #6762, opened 2022-12-22
[WIP] Support for indexing the codebase

voodoos:
This PR is still a work in progress and is part of an effort to support project-wide occurrences when working with OCaml.

The current plan to support project-wide occurrences has three distinct parts:
- A new tool that can index values in a `cmt`: [`ocaml-uideps`](https://github.com/voodoos/ocaml-uideps)
- New Dune rules that orchestrate the calls to `ocaml-uideps` to generate an index of an entire project (this PR)
- Merlin / Editor mode support to query that index

The heart of this PR is the `uideps.ml` module, that exposes two functions, `cctx_rules` and `project_rule`, that respectively provide rules to generate indexes for each compilation context and for the global project.

The generation of the indexes can be summarized as:
- Index every CMT file in the project
- For each lib and exe stanza, aggregate their cmt's indexes
- Aggregate all these indexes into the projects's global index  

```
                                        ──┐
              project.uideps              │
                    ▲                     │
                    │                     │
          ┌─────────┴─────────┐           │
          │                   │           │ "project rule"
          │                   │           │
          │                   │           │
                                          │
     cctx.uideps         cctx.uideps    ──┤
                                          │
          ▲                   ▲           │
          │                   │           │
     ┌────┴───┐               │           │
     │        │               │           │
     │        │               │           │
 a.uideps  b.uideps      main.uideps      │   "ctx rules"
                                          │
     ▲       ▲                ▲           │
     │       │                │           │
┌────┼───────┼────┐     ┌─────┼────┐      │
│    │       │    │     │     │    │      │
│  a.cmt    b.cmt │     │ main.cmt │      │
│  a.ml     b.ml  │     │ main.ml  │      │
│                 │     │          │    ──┘
│       Lib       │     │    Exe   │
└─────────────────┘     └──────────┘
```

Here a a few of the points that I think would be worth investigating / discussing:

1. Indexing is, in essence, a slow process. There is a lot of room for performance improvement right now, and that is the main reason this PR is a draft. However it is to be expected that even after optimization, the indexing process might remains too slow to be considered a standard Dune target, especially on big projects. I can see several options:
    - We could remove this target from the `@all` alias. That means the user must knowingly trigger the index build.
    - Dune could treat progress distinctly: tell that the actual build is over but indexing is still in progress.
2. I used a new "syntax extension" (with no actual syntax) to make the feature opt-in, but a dune-project option might be more suitable. 
[...]

PR #10422, opened 2024-04-15
[5.2] Build occurrences index for Merlin

voodoos:
This PR adds new rules that drive the `ocaml-index` tool to aggregate a workspace index. The rules are a bit simpler as the ones in https://github.com/ocaml/dune/pull/6762 since now part of the indexation will is done by the compiler in 5.2.

Building the index involves three steps:
- Use the `-bin-annot-occurrences` flag to have the compiler store partial indexes in the `cmt` files.
- Run the `ocaml-index` tool to make an index for each stanza (actually each compilation-context).
- Run the `ocaml-index` tool to aggregate every stanza's index into a global project index.
  
I ran simple benchmarks on Dune itself. These are cold builds ran after `./dune.exe clean`:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./dune.exe build @check` (without `-bin-annot-occurrences`) | 8.632 ± 0.167 | 8.454 | 8.950 | 1.00 |
| `./dune.exe build @check` | 8.637 ± 0.108 | 8.433 | 8.796 | 1.00 |
| `./dune.exe build @check @ocaml-index` | 13.597 ± 0.173 | 13.384 | 13.904 | 1.00 |

- There are no perceptible performance cost of using the `-bin-annon-occurrences` flag when writing `cmt` files, so the best option is probably to always use that flag to simplify the rules.
- Actually building and gathering the index is costly (especially from a cold build), so it should be opt-in. (We have several optimizations in mind that should improve that situation in the future.)

There are several moving parts that I would like to discuss in the next dune-dev meeting:
1. The indexing rules have specific dependencies that I am unsure how best to check/enforce:
   - The external `ocaml-index` tool must be installed
   - The `-bin-annot-occurrences` flag requires OCaml 5.2
2. I am unsure of the best way to make indexation opt-in (it can be expensive on massive projects)
   - I initially added a `dune-workspace`-level option  that I remove in cd34ae18a860a3e7aec6b6fe0f798d891cb22617 following a discussion with @emillon 
   - I currently think we could simply keep the `@ocaml-index` alias but remove it from `@default` so that users have to explicitly use it.

Other issues to consider:
- The Melange compiler is not yet accepting the `-bin-annot-occurrences` flag

image


Makefile
```diff
@@ -25,7 +25,8 @@ ppx_inline_test \
 ppxlib \
 ctypes \
 "utop>=2.6.0" \
-"melange"
+"melange" \
+"ocaml-index"
```
emillon, line 29: it's best if `ocaml-index` is not a test-dep. you can write a fake binary that still allows you to test the behavior (I can help with that)

otherlibs/stdune/src/stdune.ml
```diff
@@ -1,5 +1,4 @@
-[@@@alert unstable "The API of this library is not stable and may change without notice."]
-[@@@alert "-unstable"]
```
emillon, line 2: just so we don't forget: you can work around that by passing `-alert -unstable`

src/dune_rules/alias0.ml
```diff
@@ -14,5 +14,6 @@ let doc_json = standard "doc-json"
 let doc_new = standard "doc-new"
 let check = standard "check"
 let install = standard "install"
+let ocaml_index = standard "ocaml-index"
```
emillon, line 17: if we end up making a new standard index, it'll need to be documented in `doc/reference/aliases.rst`

voodoos:
> For a first release, I would consider the following strategy to select whether this feature is enabled:
> 
>     * attach these rules to `@ocaml-index`
> 
>     * if `ocaml-index` is available, and OCaml >= 5.2 (which I suppose is an opam constraint already), attach `@ocaml-index` to `@check`
> 
> 
> (This is similar to how we set up `sherlodoc`, though the situation is a bit different)
> 
> That way:
> 
>     * you can test this in dune's test suite, by providing a fake `ocaml-index` binary (check how sherlodoc is tested)
> 
>     * people can use it in an opt-in way in their workspaces
> 
>     * we can find later a way to configure this in a more precise way.

I wonder whether we should attach `@ocaml-index` to `@check` to make it really "opt-in" ?

voodoos:
@rgrinberg removing the aggregation step does have an impact on cold build time, which is probably more noticeable on rebuilds. `@check @ocaml-index` is still 46% slower than `@check`, but it was 57% before. 

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./dune.exe build @check` | 9.090 ± 0.089 | 8.959 | 9.222 | 1.00 |
| `./dune.exe build @check @ocaml-index` | 13.302 ± 0.174 | 12.964 | 13.449 | 1.00 |

But now I realize that my measurement are wrong: the `@ocaml-index` appears to build much more targets (and not only indexes) than `@check` so it is not a good comparison. (and maybe I do something wrong in my rules ?)

If I compare with `@check @install` the results are much "better", only a 16% increase in built time:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./dune.exe build @check @install` | 18.552 ± 0.441 | 18.206 | 19.745 | 1.00 |
| `./dune.exe build @check @install @ocaml-index` | 21.517 ± 0.219 | 21.224 | 21.976 | 1.00 |

But even then, there is some additional building happening (in vendored dirs notably). I find this a bit surprising since the dependencies of `@ocaml-index` are "every cmt of every stanza of the project" which I would expect to be the result of building `@check` ?

It would be great if someone could have a look :-)
Additionally, is there some way to list the targets built for some given aliases ? And to have Dune output time spent running commands (ideally total time per tool) ?

In any case, the more I think about it the more I am convinced that the "don't merge" strategy is the best. I will update the merlin config and see how it goes.

otherlibs/stdune/src/stdune.ml
```diff
@@ -1,5 +1,4 @@
...
+
```
rgrinberg, line 1: an accident?

src/dune_rules/merlin/ocaml_index.ml
```diff
@@ -0,0 +1,119 @@
+open Import
+module CC = Compilation_context
+module SC = Super_context
+
+let ocaml_index sctx ~dir =
+  Super_context.resolve_program_memo ~loc:None ~dir sctx "ocaml-index"
+;;
+
+let index_path_in_obj_dir ?for_cmt obj_dir =
```
rgrinberg, line 9: Make this argument required please.

src/dune_rules/merlin/ocaml_index.ml
```diff
@@ -0,0 +1,119 @@
...
+      ~f:(fun module_ ->
+        Obj_dir.Module.cmt_file obj_dir ~ml_kind:Intf ~cm_kind module_
+        |> Option.map ~f:(fun cmti -> Path.build cmti))
+      modules
+  in
+  let* ocaml_index = ocaml_index sctx ~dir in
+  let context_dir =
+    CC.context cctx |> Context.name |> Context_name.build_dir |> Path.build
+  in
+  let* additionnal_libs =
```
rgrinberg, line 51: `additional_libs`

src/dune_rules/merlin/ocaml_index.ml
```diff
@@ -0,0 +1,119 @@
...
+    let open Resolve.Memo.O in
+    let* req_link = CC.requires_link cctx in
+    let+ req_compile = CC.requires_compile cctx in
+    let a =
+      List.fold_left req_link ~init:[] ~f:(fun acc l ->
+        if List.exists req_compile ~f:(Lib.equal l)
+        then acc
+        else (
+          let dir = Lib.info l |> Lib_info.obj_dir |> Obj_dir.byte_dir in
+          Command.Args.(A "-I" :: Path dir :: acc)))
```
rgrinberg, line 68: Maybe you can can use `Lib_flags.L.include_flags` here?

src/dune_rules/merlin/ocaml_index.ml
```diff
@@ -0,0 +1,119 @@
...
+    in
+    Command.Args.S a
+  in
+  let fn = index_path_in_obj_dir obj_dir in
+  let includes =
+    Resolve.peek additionnal_libs |> Result.value ~default:Command.Args.empty
+  in
+  let aggregate =
+    Command.run
```
rgrinberg, line 77: This should be `Command.run_dyn_prog` and the resolution of the binary should be delayed into action builder

rgrinberg:
As discussed, the performance penalty currently is too large to build this by default. A separate alias sounds like a good start to me. When this feature becomes fast enough for general use, do you think we will still need the ocaml-index alias?

Also, could you comment on the stability of this feature for the users? When this feature is shipped, is it recommended that everyone start using it right away? Or is it more for wider experimentation?

voodoos:
Thanks for your review @rgrinberg. I was focused on experimenting with performance improvement but I will answer your comments soon.

We spent some time with @emillon to make the indexer faster and we did found some important optimizations.

However indexing only the libraries/executables which are part of `@check` still takes  increases cold-build time by 18% .
Moreover most of the additional time is spent at the end of the build when indexing the `dune_rules` library which features a large number of modules. I notably tried things such as indexing each `cmt` independently to get finer grained dependencies. This improves partial re-build time a lot but also increase significantly cold-build time because it doesn't allow the indexer to cache results that are the same for every `cmt`.

In short, I don't think we can afford to enable the feature by default right now. I propose to add two additional targets: 
- `@check-index` which only indexes cmt generated by the `@check` target
- `@index-all` which perform full indexation of the project, which might trigger the build of additionnal targets like unused vendored dependencies.

Eventually, our best option is probably to have LSP ask dune via RPC to update the index when needed.

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./dune.exe build @check` | 8.789 ± 0.094 | 8.673 | 8.920 | 1.00 |
| `./dune.exe build @check-index` | 10.422 ± 0.174 | 10.227 | 10.838 | 1.00 |
| `./dune.exe build @index-all` | 11.520 ± 0.239 | 11.177 | 11.877 | 1.00 |

Results are actually better on a codebase such as Irmin where there is no single very large library such as `dune_rules`: the time increase between `@check` ad `@check-index`is 9.5%.

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `dune build @check` | 15.718 ± 0.091 | 15.618 | 15.928 | 1.00 |
| `dune build @check-index` | 17.220 ± 0.129 | 17.000 | 17.420 | 1.00 |
| `dune build @ocaml-index` | 17.437 ± 0.120 | 17.228 | 17.601 | 1.00 |

There is a tension between: more granularity for smaller incremental rebuild VS less granularity that allow for greater use of cache which speeds up cold builds but slow down incremental rebuilds. We could probably adapt the indexer to cache on the disk results for each cmt so that we could get the best of both worlds, but I don't think the benefit would be big enough to make it a priority.

Makefile
voodoos: Could you point me to such "fake binary" in the tree that I could take inspiration from ?

otherlibs/stdune/src/stdune.ml
voodoos: It was the quickest hack to make my build pass with OCaml 5.2, I pushed it by mistake.

otherlibs/stdune/src/stdune.ml
voodoos: I am not familiar with alerts, I tried to add `-alert` but it still fails:
```
File "otherlibs/stdune/src/stdune.ml", line 2, characters 0-29:
2 | [@@@alert "-alert -unstable"]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error (warning 47 [attribute-payload]): illegal payload for attribute 'alert'.
Ill-formed list of alert settings

File "otherlibs/stdune/src/stdune.ml", line 1, characters 4-9:
1 | [@@@alert unstable "The API of this library is not stable and may change without notice."]
        ^^^^^
Error (warning 53 [misplaced-attribute]): the "alert" attribute cannot appear in this context
```

src/dune_rules/alias0.ml
voodoos: todo @voodoos: I will add documentation when we will have settled on the additional alias(es) we want.

src/dune_rules/merlin/ocaml_index.ml
voodoos: Right, it is not even used anymore, I removed it.

src/dune_rules/merlin/ocaml_index.ml
voodoos: Done in 92c986567cd4ca1317b96aa3e3d2672aa29ae452

Do you know if there is a better way for me to get the list of implicit transitive dependencies than doing the ugly `requires_link \ requires_compile` intersection ?

Makefile
emillon: `test/blackbox-tests/utils/refmt.ml`

otherlibs/stdune/src/stdune.ml
emillon: taking a step back: why did you make that change? in other words, what breaks when you leave these attributes?

Makefile
voodoos: I tried to do something similar but I had to hack some workaround because the binary doesn't have a valid module name (`ocaml-index`). Is there a better way to do it ?

It's in that commit: https://github.com/ocaml/dune/pull/10422/commits/cc7c74e29f61480b5d93cb782b8782027c49bd77

rgrinberg:
> Additionally, is there some way to list the targets built for some given aliases ? And to have Dune output time spent running commands (ideally total time per tool) ?

Yes, there's the `--trace-file` flag

rgrinberg:
Could you stick to `check-index` or `index-all` for now? The only reason vendored sources aren't included in `@check` is because of a bug.

src/dune_rules/alias0.ml
voodoos: done in [4d13b4d](https://github.com/ocaml/dune/pull/10422/commits/4d13b4d1d00e3f49166bddc0e5a94d52c68a1044)

voodoos:
I rebased the PR and made a few more changes like adding documentation and removing `@check-index`.

I think the PR is now ready for another round of review. 

@rgrinberg I still wonder about the hidden transitive deps issue. I think now that the compiler supports the `-H` flag and Merlin has corresponding configuration directives we should pass it the hidden deps using these directives Right now to get the set of hidden deps I use `requires_link \ requires_compile` and it appears to work, but is that correct ? Is there a better way to get these ?

src/dune_rules/merlin/merlin.ml
```diff
@@ -224,6 +244,7 @@ module Processed = struct
       printf "STDLIB %s\n" (serialize_path stdlib_dir));
     Path.Set.iter obj_dirs ~f:(fun p -> printf "B %s\n" (serialize_path p));
     Path.Set.iter src_dirs ~f:(fun p -> printf "S %s\n" (serialize_path p));
+    List.iter indexes ~f:(fun p -> printf "INDEX %s\n" (serialize_path p));
```
rgrinberg, line 247: What would happen to older versions of merlin if they'd encounter this directive?

src/dune_rules/merlin/ocaml_index.ml
```diff
@@ -0,0 +1,120 @@
...
+      List.filter req_link ~f:(fun l -> not (List.exists req_compile ~f:(Lib.equal l)))
+    in
+    Lib_flags.L.include_flags non_compile_libs (Lib_mode.Ocaml Byte)
+  in
+  let dir = Compilation_context.dir cctx in
+  let aggregate =
+    let obj_dir = Compilation_context.obj_dir cctx in
+    let fn = index_path_in_obj_dir obj_dir in
+    let includes =
+      Resolve.peek additional_libs |> Result.value ~default:Command.Args.empty
```
rgrinberg, line 41: Is the peek here intentional? In case we can't resolve one of the libs, we just pretend all of them don't exist without letting the user know?

src/dune_rules/merlin/merlin.ml
voodoos: Since Merlin 4.6 unknown configuration tags are ignored when `dune ocaml-merlin` is used as a configuration reader, so it should not trigger any error.

src/dune_rules/merlin/ocaml_index.ml
voodoos: That's a good catch, I removed the peek in [4c2aee3](https://github.com/ocaml/dune/pull/10422/commits/4c2aee34457b9c4e2ec78dc7f405e748a5558add)

src/dune_rules/merlin/ocaml_index.ml
```diff
@@ -0,0 +1,109 @@
...
+        let cmts =
+          [ Intf; Impl ]
+          |> List.filter_map ~f:(fun ml_kind ->
+            Obj_dir.Module.cmt_file obj_dir ~ml_kind ~cm_kind module_
+            |> Option.map ~f:Path.build)
+        in
+        List.rev_append cmts acc)
+    in
+    Command.run_dyn_prog
+      ~dir:context_dir
```
rgrinberg, line 60: Why do you not run ocaml-index from `dir`? We only run compilation commands from context_dir for good error messages. Does ocaml-index output error messages that refer to source code?

src/dune_rules/merlin/ocaml_index.ml
```diff
@@ -0,0 +1,109 @@
...
+      (ocaml_index sctx ~dir)
+      [ A "aggregate"
+      ; A "--root"
+      ; A Path.(Source.root |> source |> to_absolute_filename)
```
rgrinberg, line 64: Why is it needed to absolutize the path here?

src/dune_rules/merlin/ocaml_index.ml
voodoos: `ocaml-index` relies on the load-path stored in the `cmt` files. These paths are relative to the directory in which the compiler was called.

If you would prefer the indexer to run in `dir` we can have dune pass the complete load-path via the CLI.

src/dune_rules/merlin/ocaml_index.ml
voodoos: The indexer aggregate parsetree locations. These locations contain a path that is relative to the call-site of the compiler. When reading the index, Merlin needs a way to find the correct source file of each occurrences. The current way of doing that is that the indexer make the relative path of location absolute using the `--root` argument. 

It was the most straightforward approach, but if you'd prefer we can also keep relative paths in the locations and either store the source root path into the index or pass it to Merlin via configuration.

src/dune_rules/merlin/ocaml_index.ml
rgrinberg: No, that's fine to leave it this way. It makes sense to leave a comment about this though.

src/dune_rules/merlin/ocaml_index.ml
rgrinberg: > When reading the index, Merlin needs a way to find the correct source file of each occurrences. The current way of doing that is that the indexer make the relative path of location absolute using the --root argument.

Shouldn't the source directory be communicated through the .merlin file? Or the equivalent configuration that we have via the `dune ocaml-merlin` command.

> It was the most straightforward approach, but if you'd prefer we can also keep relative paths in the locations and either store the source root path into the index or pass it to Merlin via configuration.

With absolute paths, we're making the rule non relocatable and unusable with the dune cache. So I would say that using a relative path is the minimum. But see my first question. I'd like to keep relying on the configuration command that merlin runs to communicate the source directory.

rgrinberg (approved):
LGTM.

Please add a CHANGES entry before merging

PR #10425, opened 2024-04-16
refactor: Remove duplicated match when calculating digest

Leonidas-from-XIV:
As I was trying to understand the caching code I came across the fact that the match inside the loop and outside of it seemed to have the exact condition, so I removed the outermost.


Leonidas-from-XIV:
Missed the fact that `S_LNK` is forbidden on the toplevel but allowed in the `loop`.

PR #10426, opened 2024-04-16
fix: Don't fail on symlinks to folders

Leonidas-from-XIV:
Trying to figure out why it the code was forbidding symlinks to folders.

Locally the tests seem to succeed.

### DIFF
--- a/src/dune_digest/cached_digest.ml
+++ b/src/dune_digest/cached_digest.ml
@@ -283,7 +283,7 @@ let refresh_and_remove_write_permissions ~allow_dirs path =
       (match stats.st_kind with
        | S_LNK ->
          (match Path.stat_exn path with
-          | stats -> refresh stats ~allow_dirs:false path
+          | stats -> refresh stats ~allow_dirs path
           | exception Unix.Unix_error (ELOOP, _, _) -> Error Cyclic_symlink
           | exception Unix.Unix_error (ENOENT, _, _) -> Error Broken_symlink)
        | S_REG ->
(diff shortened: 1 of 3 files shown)


Leonidas-from-XIV:
@Sudha247 reminded me of this PR so out of curiosity I rebased it and it still seems to just work, so I'm marking it as ready to review. Maybe the review will reveal what the idea behind disallowing directories originally was.

Alizter:
Things that I would check:

1. What happens to symlinks to absolute directories? They probably aren't portable.
2. What happens if a directory target contains a broken symlink?
3. What happens to symlinks to places outside of the workspace.

Sudha247:
> What happens to symlinks to absolute directories? They probably aren't portable.

My first reaction to this is, it's probably okay..? If I have a local project that I'm building with dune and have added an absolute path on my local system as a symlink, it might break in a different scenario, but I'd still expect it to work on the current setup. It'd be good to have a test for this, though.

> What happens to symlinks to places outside of the workspace.

Same thoughts as above.

> What happens if a directory target contains a broken symlink?

Good point, will be good to check.

Alizter:
More questions here:
4. What happens when a directory target contains a symlink to somewhere outside of a directory target.

This is probably the reason for the restriction today.

Leonidas-from-XIV:
Closing with the same reason as #12613:

> After some discussion about this it seems supporting directory symlinks inside directory targets might need a little bit more thinking and design than we can spare at the moment.

PR #10436, opened 2024-04-17
chore: prepare 3.15.1

emillon:


CHANGES.md
```diff
@@ -1,3 +1,17 @@
+3.15.1 (2024-04-17)
+-------------------
+
```
Leonidas-from-XIV, line 3: I would add a "Fixed" subheader for this. Though looking at the older entries in the changelog, it's not really all that consistent.

emillon: good idea!

Leonidas-from-XIV (approved):
LGTM!

PR #10438, opened 2024-04-17
always block signals and attempt to shutdown gracefully

pmwhite:
We should always block signals so that we have a chance to shutdown properly instead of violently. However, it's not always appropriate to print the warning message. I'm not claiming that all the decisions in this commit are correct, but they are as close to behavior-preserving as possible.


rgrinberg:
Looks like the failures are related to action runners. We should just get rid of them here as they're a useless feature to anyone external

PR #10439, opened 2024-04-18
feature: Read creation time of files

Leonidas-from-XIV:
This could be used to e.g. evict files from the dune cache by whichever files are the oldest (instead of `ctime`, which is updated when symlinks are added or removed).

Currently a WIP since it was only tested on a recent Linux version.


rgrinberg:
This is an interesting idea, but crime is not the criteria we'd use ideally. Just because something is old, doesn't mean it's a poor cache entry. A poor cache entry is one that isn't being hit very often.

Feel free to reopen if you have improvements on this idea.

Issue #10440, opened 2024-04-18
The colon variants of withenv env update operators do not adding leading/trailing separators


gridbugs:
The `:=` and `=:` operators inside the `withenv` action do not add leading/trailing separators when operating on unset variables. These operators are documented in opam's manual as:

> += or := prepend. They differ when the variable is unset of empty, where := adds a trailing separator.
> =+ or =: append. They differ when the variable is unset of empty, where =: adds a leading separator.

Todo: add repro

PR #10441, opened 2024-04-18
Add repro bug where leading and trailing colons are omitted from variables in withenv

gridbugs:
Reproduces https://github.com/ocaml/dune/issues/10440


rgrinberg (approved):
This code is delicate enough you might want to extract the environment management into a library and unit test it.

PR #10442, opened 2024-04-18
fix: lazier directory target extraction

rgrinberg:
If no directory targets are defined, then do not evaluate enabled_if

might fix the bug reported by @anmonteiro 

cc @gridbugs

### DIFF
--- a/src/dune_rules/dir_status.ml
+++ b/src/dune_rules/dir_status.ml
@@ -85,10 +85,38 @@ let error_no_module_consumer ~loc (qualification : Include_subdirs.qualification
     ]
 ;;
 
-let extract_directory_targets ~dir stanzas =
-  Memo.parallel_map stanzas ~f:(fun stanza ->
-    match Stanza.repr stanza with
-    | Rule_conf.T { targets = Static { targets; _ }; loc = rule_loc; enabled_if; _ } ->
+let directory_targets_of_rule ~dir { Rule_conf.targets; loc = rule_loc; enabled_if; _ } =
+  match targets with
+  | Infer ->
+    (* we don't infer directory targets *)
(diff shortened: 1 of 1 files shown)


src/dune_rules/dir_status.ml
```diff
@@ -85,42 +85,53 @@ let error_no_module_consumer ~loc (qualification : Include_subdirs.qualification
     ]
 ;;
 
+let directory_targets_of_rule
+  ~dir
+  { Rule_conf.targets = Static { targets; _ }; loc = rule_loc; enabled_if; _ }
```
anmonteiro, line 90: there's a matching failure for this pattern in CI.

src/dune_rules/dir_status.ml
```diff
@@ -85,10 +85,36 @@ let error_no_module_consumer ~loc (qualification : Include_subdirs.qualification
     ]
 ;;
 
-let extract_directory_targets ~dir stanzas =
-  Memo.parallel_map stanzas ~f:(fun stanza ->
-    match Stanza.repr stanza with
-    | Rule_conf.T { targets = Static { targets; _ }; loc = rule_loc; enabled_if; _ } ->
+let directory_targets_of_rule ~dir { Rule_conf.targets; loc = rule_loc; enabled_if; _ } =
+  match targets with
+  | Infer -> (* we don't infer directroy targets *) Memo.return Path.Build.Map.empty
```
anmonteiro, line 90:
```suggestion
  | Infer -> (* we don't infer directory targets *) Memo.return Path.Build.Map.empty
```

anmonteiro (approved):
tested this change and it fixes the build for ocaml-protoc-plugin

Issue #10294, opened 2024-03-20
Environment Variables aren't relocated to the Sandbox


rgrinberg:
When we run actions inside the sandbox, we set various environment variables. The variables we need to relocate are:

* `PATH`
* `OPAM_SWITCH_PREFIX`
* `OCAMLPATH`
* `OCAMLTOP_INCLUDE_PATH`
* `MANPATH`
* `CAML_LD_LIBRARY_PATH`
* `OCAMLFIND_IGNORE_DUPS_IN`

All the variables above are computed automatically by dune when computing the environment from the set of packages. Since we manage these variables, we should take care to relocate them.

We should be modifying such paths into paths in the sandbox.

cc @gridbugs

gridbugs:
This sounds similar to https://github.com/ocaml/dune/pull/10267 so I'll work on fixing this too (in a separate PR).

Alizter:
This should be easier to do after #14373.

PR #10444, opened 2024-04-19
Fix gh10408

### LINKED ISSUE #10408: pkg: adding a path to the `PATH` variable with the `withenv` action overwrites the existing `PATH` value rather than prepending to it
Not sure if this behaviour is intentional or not. I expected that adding a value to `PATH` would result in the existing `PATH` being prepended to rather than clobbered.

Repro here: https://github.com/ocaml/dune/pull/10407

gridbugs:
When updating the PATH variable with the `withenv` action, dune would previously clobber the existing PATH from the environment rather than updating it. Withenv applies environment updates iteratively starting from some base environment. Prior to this change, the base environment did not include the global environment, so as far as withenv knew the PATH was originally empty. The fix is to start withenv with an environment which includes the global environment.

Fixes https://github.com/ocaml/dune/issues/10408

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -144,6 +144,47 @@ module Install_cookie = struct
   ;;
 end
 
+module Value_list_env = struct
+  (* A representation of an environment where each variable can hold a
+     list of [Value.t]. Each variable will be encoded into a delimited
+     string (in the style of the PATH variable). *)
+  type t = Value.t list Env.Map.t
+
+  let parse_strings s = Bin.parse s |> List.map ~f:(fun s -> Value.String s)
+  let of_env env : t = Env.to_map env |> Env.Map.map ~f:parse_strings
+
+  let string_of_env_values values =
+    List.map values ~f:(function
+      | Value.String s -> s
+      | Dir s | Path s -> Path.to_absolute_filename s)
+    |> Bin.encode_strings
+  ;;
+
+  let to_env (t : t) = Env.Map.map t ~f:string_of_env_values |> Env.of_map
+  let get_path t = Env.Map.find t Env_path.var
+
+  (* [extend_concat_path a b] adds all variables from [b] to [a]
+     overwriting any existing values of those variables in [a] except for PATH
+     which is set to the concatenation of the PATH variables from [a] and [b]
+     with the PATH entries from [b] preceding the PATH entries from
+     [a]. If only one of the arguments contains a PATH variable then
+     its value will be the value of PATH in the result, however if
+     neither argument contains a PATH variable then PATH will be unset
+     in the result. *)
+  let extend_concat_path a b =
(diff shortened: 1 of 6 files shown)


src/dune_rules/pkg_rules.ml
```diff
@@ -144,6 +144,40 @@ module Install_cookie = struct
...
+  let get_path t = Env.Map.find t Env_path.var
+
+  (* [extend_concat_path a b] adds all variables from [b] to [a]
+     overwriting any existing values of those variables in [a] except for PATH
+     which is set to the concatenation of the PATH variables from [a] and [b]
+     with the PATH entries from [b] preceding the PATH entries from [a] *)
+  let extend_concat_path a b =
+    let path =
+      match get_path a, get_path b with
+      | None, None -> []
```
rgrinberg, line 173: Does this mean if neither side has `PATH` set, you will set it to the empty path?

gridbugs: That's correct, though thinking more about this I think what it should be doing instead is not setting `PATH` at all in such a case. Will update.

gridbugs: Updated

src/dune_rules/pkg_rules.ml
```diff
@@ -144,6 +144,47 @@ module Install_cookie = struct
...
+
+  (* [extend_concat_path a b] adds all variables from [b] to [a]
+     overwriting any existing values of those variables in [a] except for PATH
+     which is set to the concatenation of the PATH variables from [a] and [b]
+     with the PATH entries from [b] preceding the PATH entries from
+     [a]. If only one of the arguments contains a PATH variable then
+     its value will be the value of PATH in the result, however if
+     neither argument contains a PATH variable then PATH will be unset
+     in the result. *)
+  let extend_concat_path a b =
```
rgrinberg, line 174: There's a few other variables where we need this treatment - basically any variable that we expand automatically by adding a package dep. This is a good start though.

rgrinberg (approved):

src/dune_rules/pkg_rules.ml
gridbugs: Ah that makes sense. I'll make an issue for generalizing this to other variables. Any idea what variables might be affected or under what conditions this might fail?

gridbugs: Actually @rgrinberg would you mind making an issue for this. Can you think of a problem that could be caused by dune clobering environment variables containing lists other than PATH? Possibly systems where libraries are stored in an unusual place dictated by an environment variable? I guess dune needs to establish a convention for which variables get passed through and which ones are sandboxed.

Relevant: https://github.com/ocaml/dune/pull/8237

rgrinberg: Isn't this the same as https://github.com/ocaml/dune/issues/10294? I've expanded the list of variables we need to relocate there.

PR #9552, opened 2023-12-20
refactor: switch to new API for recording deps

rgrinberg:
Continuation of the splitting of https://github.com/ocaml/dune/pull/9132

PR #10446, opened 2024-04-20
stop recording deps in Action_builder.contents

### LINKED ISSUE #10149: dune always rebuilds all coq files as soon as any dependency changes
## Expected Behavior

Only `leaf.vo` gets rebuild.

## Actual Behavior

All files get rebuilt, including `other_root.vo` (the added dependency) as well as `root.v` (the existing dependency).

## Reproduction

- PR with a reproducing test: 
https://github.com/ocaml/dune/pull/10150

The steps below can be used to reproduce exactly the behavior described above but the reproducing PR has a smaller test case that is better suited to debugging this.

1. Extract [bug.zip](https://github.com/ocaml/dune/files/14418639/bug.zip)
2. `cd bug`
3. `dune build coq`
4. `echo "Require Import bug.other_root." >> coq/leaf.v`
5. `dune build coq`

## Specifications

- Version of `dune` (output of `dune --version`): 3.13.0, master
- Operating system (distribution and version): linux

## Additional information 

`git bisect` points to:

```
2876a55d2646cdd8fa130398f3fdfb7f8acdbd9f is the first bad commit
commit 2876a55d2646cdd8fa130398f3fdfb7f8acdbd9f
Author: Rudi Grinberg <me@rgrinberg.com>
Date:   Wed Dec 20 18:52:18 2023 -0600

    refactor: switch to new API for recording deps (#9552)
    
 src/dune_engine/action_builder.ml                  | 93 +++++++++++++++++-----
 src/dune_engine/action_builder.mli                 | 41 ++++++++--
 src/dune_engine/build_system.ml                    | 37 +++------
 src/dune_engine/build_system.mli                   |  3 +
 src/dune_engine/load_rules.ml                      | 47 +++++------
 src/dune_rules/action_builder.ml                   | 14 +---
 src/dune_rules/alias_builder.ml                    | 47 ++++-------
 .../test-cases/special_files.t/run.t               |  3 +-
 8 files changed, 164 insertions(+), 121 deletions(-)
bisect found first bad commit
```

The corresponding PR is https://github.com/ocaml/dune/pull/9552

CC @rlepigre

Alizter:
PR https://github.com/ocaml/dune/pull/9552 changed the semantics of `Action_builder.contents` so that it records a dependency. The Coq rules were relying on the previous behaviour of not recording a dependency in order to generate a build plan. This caused issue https://github.com/ocaml/dune/issues/10149 to appear.

This PR fixes https://github.com/ocaml/dune/issues/10149 by removing the dependency being recorded in `Action_builder.contents` matching the previous semantics.

cc @rgrinberg 

Here is the semantic change in #9552 
https://github.com/ocaml/dune/pull/9552/files#r1573098164

- [x] changelog

### DIFF
--- a/doc/changes/10446.md
+++ b/doc/changes/10446.md
@@ -1,0 +1,2 @@
+- Fix a bug where Coq projects were being rebuilt from scratch each time the
+  dependency graph changed. (#10446, fixes #10149, @alizter) 
--- a/src/dune_rules/action_builder.ml
+++ b/src/dune_rules/action_builder.ml
@@ -37,12 +37,7 @@ let paths ps = deps (Dep.Set.of_files ps)
 let path_set ps = deps (Dep.Set.of_files_set ps)
 let dyn_paths paths = dyn_deps (paths >>| fun (x, paths) -> x, Dep.Set.of_files paths)
 let dyn_paths_unit paths = dyn_deps (paths >>| fun paths -> (), Dep.Set.of_files paths)
-
-let contents p =
-  let* () = Dep.file p |> Dep.Set.singleton |> Build_system.record_deps in
-  of_memo (Build_system.read_file p)
-;;
-
+let contents p = of_memo (Build_system.read_file p)
 let lines_of p = contents p >>| String.split_lines
 
 let read_sexp p =
--- a/test/blackbox-tests/test-cases/coq/no-rebuild-on-dep-change.t
+++ b/test/blackbox-tests/test-cases/coq/no-rebuild-on-dep-change.t
@@ -11,8 +11,7 @@
 This test makes sure that a full rebuild is not triggered when the output of
 coqdep is changed.
 
-Currently this is the case and a bug:
+This is as expected:
   $ dune build --display=short
         coqdep .bug.theory.d
-          coqc root.{glob,vo}
           coqc leaf.{glob,vo}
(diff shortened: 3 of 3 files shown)


snowleopard:
Wow, that seems like a pretty serious bug. It probably made incremental builds much worse in general, not just for Coq. Thanks for the fix!

Do we have any benchmarks that show an improvement after this commit?

It may be worth patching some earlier releases.

rgrinberg:
Thanks, could you add a change log entry?

We do not have any incremental benchmarks.

Alizter:
@rgrinberg I've added a changelog.

I don't see exactly how this would make incremental builds worse outside of Coq. The Coq support appears to be the only serious consumer of `Action_builder.contents` in terms of using it to generate a build plan. But that also means we have more motivation to keep the original semantics.

I've only mentioned Coq in the changelog since it is the only user-visible effect I am aware of.

ejgallego:
Thanks for the fix @Alizter , a couple of additional comments:
- I can see how the semantics of `contents` and dep recording can be confusing some times, maybe it would be worth to make the documentation of `contents` more precise in the `.mli` file?
- I have seen various criteria for naming PRs as "refactor", maybe we should clarify the rules of what do we consider as a "refactoring"? [I fear that indeed when we tag stuff as refactor, we tend to review less carefully as most of those are boring / straighforward]

PR #10449, opened 2024-04-22
[3.15] backport #10442

emillon:
- refactor: evaluate directory targets enabled_if in parallel map (#10402)
- fix: lazier directory target extraction (#10442)


emillon:
I had to pick the refactor which changes some error messages but that looks safe enough to me.

Leonidas-from-XIV (approved):
It's a bit unusual for messages to change with point-releases but it's probably not worth overcomplicating things and backporting just the fix.

PR #10455, opened 2024-04-23
Clarify some env handling for pkg rules

gridbugs:
The logic for handling environments used when building packages is complicated. This change breaks it up into smaller functions, simplifies some of the logic, and adds some comments explaining what each one does.

A consequence of this change is that the order in which the bin directories of dependencies appear in the PATH variable while building a package is the same as the order in which the dependencies appear in the package's lockfile (previously the order was reversed).

---

This mostly consists of notes I made for myself and experimental changes I made to help me understand how this logic works. I think it's worth checking it in to help the next person better understand this code.

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -153,6 +153,8 @@ module Value_list_env = struct
   let parse_strings s = Bin.parse s |> List.map ~f:(fun s -> Value.String s)
   let of_env env : t = Env.to_map env |> Env.Map.map ~f:parse_strings
 
+  (* Concatenate a list of values in the style of lists found in
+     environment variables, such as PATH *)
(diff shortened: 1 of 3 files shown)


rgrinberg (approved):

rgrinberg (changes requested):
LGTM, but the tests are failing

gridbugs:
Fixed a bug causing some variables to be concatenated rather than replaced in a situation where they should be replaced. I've promoted the failing test (opam-package-with-setenv) as the behaviour of dune after this change is closer to the behaviour of opam (confirmed by setting up some local opam packages matching the packages from this test). Dune's behaviour still doesn't perfectly match opam's because of https://github.com/ocaml/dune/issues/10440.

rgrinberg (approved):

PR #10456, opened 2024-04-24
Remove action runner commands.

pmwhite:
Right now this is only a useful feature for Jane Street. Removing it to avoid adding extra maintenance burden to external Dune. (motivated by #10438)


pmwhite:
I could be mistaken, but the test failures seem unrelated to this PR.

PR #10458, opened 2024-04-25
Look for META files rather than just directories to list plugins

### LINKED ISSUE #10457: An empty directory in the plugin directory makes `load_all` fail
The dune-site plugin system expects all the entries of the plugin directory to be directories, each of which should contain a `META` file.
Unfortunately, we can end up with an empty directory there in a rather normal course of actions, after removing a package that provides a plugin.

Here is a small scenario that shows the problem. First, the following commands exhibit the intended behaviour:

```
$ opam install ortac-core
[...]

$ ortac
ortac: error: no plugin is available, please install at least one

$ opam install ortac-qcheck-stm
[...]

$ ortac
ortac: error: missing command
Usage: ortac [COMMAND] …
Try 'ortac --help' for more information.
```

But removing the second package breaks the first:

```
$ opam remove ortac-qcheck-stm
[...]

$ ortac
Fatal error: exception The plugin "qcheck-stm" can't be found in the search paths "%{prefix}%/lib/ortac-core/plugins".
```

where one would expect the initial behaviour (`no plugin is available...`).

shym:
This PR proposes a possible fix for #10457.

If the plugin directory contains anything but directories that each contain a `META` file, loading plugins fails. This PR proposes to check additionally for the presence of the `META` files and ignore other entries in the plugin directory.


otherlibs/dune-site/src/plugins/plugins.ml
```diff
@@ -1,10 +1,16 @@
 open Dune_site.Private_
 module Data = Dune_site_plugins_data
 
+let meta_fn = "META"
+
 let readdir dirs =
+  let ( / ) = Filename.concat in
```
anmonteiro, line 7:
```suggestion
let readdir =
  let ( / ) = Filename.concat in
  fun dirs ->
  
```

otherlibs/dune-site/src/plugins/plugins.ml
```diff
@@ -1,10 +1,16 @@
...
   List.concat
     (List.map
-       (fun dir -> Array.to_list (Sys.readdir dir))
+       (fun dir ->
```
anmonteiro, line 10: this can probably be a `List.filter_map` instead.

It's available since OCaml 4.08, so it should be safe to use inside dune (https://github.com/ocaml/ocaml/blob/8b6b7cfbf26ae50ae87d5849c827e609a1309c98/stdlib/list.mli#L195-L200)

anmonteiro (commented):
works for me. it'd be nice to have a test case for this

shym:
Thank you for your review!

I wasn’t sure about the best way to test this behaviour. I have added such a possible test, where the empty directory is explicitly created to mimick what happens with OPAM.

Your suggestion to go with `filter_map` sent me on another exploration. Checking that `Sys.file_exists` on each candidate directory is not enough to ensure `readdir` will not fail. So I ended up with the following proposition: just try to `Sys.readdir` each plugins directory and handle the errors as empty directories.

rgrinberg:
ping @anmonteiro

otherlibs/dune-site/test/opam-uninstall.t
```diff
@@ -0,0 +1,96 @@
...
+  >  (libraries app.register dune-site dune-site.plugins))
+  > 
+  > (library
+  >  (public_name app.register)
+  >  (name registration)
+  >  (modules registration))
+  > 
+  > (generate_sites_module
+  > (module sites)
+  > (plugins (app plugins)))
```
anmonteiro, line 31:
```suggestion
  > (generate_sites_module
  >  (module sites)
  >  (plugins (app plugins)))
```

anmonteiro (approved):

emillon:
Thanks. Can you add a changelog entry in `doc/changes/`?

PR #10463, opened 2024-04-28
test(lib-collision): show dune uses the relevant loc for collision errors

anmonteiro:


test/blackbox-tests/test-cases/lib-collision/lib-collision-public-name.t
```diff
@@ -0,0 +1,42 @@
...
+  >  (public_name baz.foo))
+  > EOF
+
+Without any consumers of the libraries
+
+  $ dune build
+  File "dune", line 5, characters 7-10:
+  5 |  (name foo)
+             ^^^
+  Error: Library "foo" appears for the second time in this directory
```
anmonteiro, line 25: uses the relevant library for the error message, which in this case is `foo` (the colliding one.)

rgrinberg (approved):

Issue #8203, opened 2023-07-13
Getting tests working on Nix


Alizter:
Currently the tests don't all work on Nix. This issue serves as a tracker to fixing them. Eventually we will be able to run the test suite in the Nix CI jobs.

Here are the ones that don't:
- [x] `otherlibs/dune-site/test`
  + Installation directory was relying on implicit opam. Adding --prefix fixes this.
  + Fix: https://github.com/ocaml/dune/pull/8206
  * `otherlibs/dune-site/test/run.t`
  * `otherlibs/dune-site/test/run_2_9.t`
- `test/blackbox-tests/test-cases`
  * [x] `test/blackbox-tests/test-cases/describe`
    + Fix: https://github.com/ocaml/dune/pull/8204
    * `test/blackbox-tests/test-cases/describe/describe-workspace-pp.t`
    * `test/blackbox-tests/test-cases/describe/describe.t`
  * [x] `test/blackbox-tests/test-cases/github2629.t`
    + Fix: https://github.com/ocaml/dune/pull/8205
  * [x] `test/blackbox-tests/test-cases/install-dir/install-libdir.t/run.t`
    + Fix: https://github.com/ocaml/dune/pull/8341
    + Fix: https://github.com/ocaml/dune/pull/8878
  * [ ] `test/blackbox-tests/test-cases/melange`
    * `test/blackbox-tests/test-cases/melange/merlin-compile-flags.t`
    * `test/blackbox-tests/test-cases/melange/merlin.t`
  * `test/balckbox-tests/test-cases/merlin`
    * [x] `test/blackbox-tests/test-cases/merlin/instrumentation.t/run.t`
      + Fix: https://github.com/ocaml/dune/pull/8880 
    * [x] `test/blackbox-tests/test-cases/merlin/merlin-tests.t/run.t`
      + Fix: https://github.com/ocaml/dune/pull/8881
  * [x] `test/blackbox-tests/test-cases/pkg`
    + Fix: https://github.com/ocaml/dune/pull/8237
    + Appears to be an issue about PATH and sh on Nix. For example, `cat` is not being found.
    * `test/blackbox-tests/test-cases/pkg/extra-sources.t`
    * `test/blackbox-tests/test-cases/pkg/install-action-dirs.t`
    * `test/blackbox-tests/test-cases/pkg/install-action.t`
    * `test/blackbox-tests/test-cases/pkg/installed-binary.t`
    * `test/blackbox-tests/test-cases/pkg/package-files.t`
[...]

PR #10465, opened 2024-04-29
fix: handle trailing / in Path.External.drop_prefix

emillon:
In a nutshell:

    # drop_prefix ~prefix:"/a/b/c/" "/a/b/c/d/e"
    Some "d/e"

This improves the test situation on nix (#8203) in the default devshell, because the paths in `OCAMLPATH` can have trailing slashes.

More precisely, the `--sanitize-for-tests` options uses `Path.External.drop_prefix` to replace elements of `OCAMLPATH` by the `FINDLIB` string. If `OCAMLPATH` contains paths with trailing slashes, these prefixes are not sanitized properly.

### DIFF
--- a/otherlibs/stdune/src/path.ml
+++ b/otherlibs/stdune/src/path.ml
@@ -1457,7 +1457,10 @@ let drop_prefix path ~prefix =
   if prefix = path
   then Some Local.root
   else (
-    let prefix = to_string prefix ^ "/" in
+    let prefix_s = to_string prefix in
+    let prefix =
+      if String.is_suffix ~suffix:"/" prefix_s then prefix_s else prefix_s ^ "/"
+    in
     let open Option.O in
     let+ suffix = String.drop_prefix (to_string path) ~prefix in
     Local.of_string suffix)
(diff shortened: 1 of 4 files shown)


rgrinberg:
The tests are failing.

Also, this needs a better description on why it's better to modify this function rather than the tests themselves.

emillon:
I changed to a portable function and amended the description.

otherlibs/stdune/src/path.ml
```diff
@@ -505,7 +505,7 @@ end = struct
     if prefix = path
     then Some Local.root
     else (
-      let prefix = prefix ^ "/" in
+      let prefix = if String.is_suffix ~suffix:"/" prefix then prefix else prefix ^ "/" in
```
rgrinberg, line 508: Can you make sure the `drop_prefix` works in a consistent way for non external paths as well? I think it might rely on the fact that `Path.Local.t` never contains a leading `/`, but it's good to double check.

emillon: I checked and this was failing, so I pushed a fix for this as well

rgrinberg (approved):
LGTM. The implementations of drop_prefix were basically copy-pasted so I got rid of the one in `Path.External`. The performance wasn't worth the duplication.

emillon:
thanks!

Issue #10362, opened 2024-04-02
Add an option to enable libraries on specific contexts based on the `modes` field


jchavarri:
In a project with both native and Melange libraries, one might want to keep the former in `default` context, and move all Melange libraries to a different context `melange`, so that library name collisions can be avoided (see related issue #10222).

To do this, Dune currently requires going through all libraries that include `(modes melange)`, and add `(enabled_if (= %{context_name} "melange"))`. 

Besides that, if one wants to properly separate namespaces, all existing OCaml libraries should be annotated with the equivalent `(enabled_if (= %{context_name} "default"))`. 

This is quite noisy for projects with tens or hundreds of libraries. Besides, the `enabled_if` conditions are mostly redundant with the information already available in the library `modes` field.

I wonder if there could be some setting in such scenarios where a single addition to the `dune-project` or `dune-workspace` file could include the `(= %{context_name} "xxx")` condition to each library, depending on its mode, e.g.:

- If the library has `(modes melange)`: adds `(enabled_if (= %{context_name} "melange"))`
- If the library doesn't have `(modes melange)`: adds `(enabled_if (= %{context_name} "default"))`
- If the library has `(modes :standard melange)`: doesn't add any condition (library is available in both contexts)

This setting could be activated through a new stanza `(melange_use_mode_as_context true)`.

cc @anmonteiro @davesnx @andreypopp

PR #10471, opened 2024-04-30
test: show that universal libraries can't share the public_name

anmonteiro:
Inspired by a question from @andreypopp.

This is a bit of a sad limitation, because it means `enabled_if` just proliferates when depending on installed libraries.


test/blackbox-tests/test-cases/enabled_if/eif-public-name-collision.t
```diff
@@ -0,0 +1,64 @@
+Public libraries using the same `public_name`, in different contexts
```
rgrinberg, line 1: Could you describe the test a bit more here? It should be possible to know what it's about without knowing any context.

jchavarri:
I tried using `(switch)` in the `melange` context defined in the test so that public libraries get installed in separate switches, but the result is the same. 😕 I guess Dune duplication checks come before actually checking if the public libraries are being installed in separate switches?

> because it means `enabled_if` just proliferates when depending on installed libraries.

I fail to see how the premise of this PR would work. How can we avoid the proliferation of `enabled_if` without something like https://github.com/ocaml/dune/issues/10362? As a consumer of a universal library, how can I use it without having two contexts? In reality, a "universal" library consists of two implementations, so I can't imagine a way in which these two implementations are exposed under a single name. For this, Dune already provides a solution: virtual libraries. But they require to hide the implementation details behind abstract types to make it work. Universal libraries allow you to expose the implementations to consumers, at the cost of defining two build contexts.

anmonteiro:
@rgrinberg thanks, I made the description a bit more clear.

jchavarri (approved):

test/blackbox-tests/test-cases/enabled_if/eif-public-name-collision.t
```diff
@@ -0,0 +1,65 @@
+Show that public library names can't be defined twice, even in different
+contexts
```
rgrinberg, line 2: It's worth adding a statement on if this is a bug or an intentional limitation.

rgrinberg (approved):

test/blackbox-tests/test-cases/enabled_if/eif-public-name-collision.t
anmonteiro: to be honest, we're not sure. It depends a bit on the usage, and how they're installed:

- for Melange libraries that clash, it's not a problem if they have the same public name, as we install melange artifacts under `<lib_name>/melange` anyway. 
- for other libraries, if we decided to install them in the same context, they would be overwritten.

We're working on an installation RFC with the aim of making a special case for a "melange context" that can be installed in the same prefix as native libraries with the same name.

rgrinberg: Okay, well I think this particular issue is easily fixed. We just ignore libraries with `enabled_if` for the purposes of `libname_conflict_map`.

Issue #10473, opened 2024-05-01
Bonsai no longer builds with dune package management


gridbugs:
There's a regression resulting in the following error when using dune to build the dependencies of the package "bonsai" (which we've been using to evaluate dune package management):
```
ocamlfind: Config file not found - neither /Users/s/src/depend-on-bonsai/_build/.sandbox/afd0d45a7d1cb969e78f1a490724f4c3/_private/default/.pkg/ocamlfind/target/lib/findlib.conf nor the directory /Users/s/src/depend-on-bonsai/_build/.sandbox/afd0d45a7d1cb969e78f1a490724f4c3/_private/default/.pkg/ocamlfind/target/lib/findlib.conf.d
make: *** [install] Error 2
-> required by _build/_private/default/.pkg/zarith/target
```

The regression was introduced in https://github.com/ocaml/dune/commit/3ea5444cd7b2874a10add601b9043b5d38662a96

PR which changes back the offending line: https://github.com/ocaml/dune/pull/10474

I don't yet understand why that change caused the regression.

Leonidas-from-XIV:
Complete shot in the dark, but maybe it has something to do with `OCAMLFIND_DESTDIR` that is set to the wrong place? However, I am wondering why it would then complain about the config file, maybe somehow `OCAMLFIND_CONF` is being set to something invalid?

gridbugs:
This was fixed by https://github.com/ocaml/dune/pull/10474.

PR #10474, opened 2024-05-01
pkg: Undo change reversing order of env updates

gridbugs:
In 3ea5444cd the order of env updates performed by packages was reversed. This brought the behaviour of dune closer to opam but introduced a regression which prevents ocamlfind from building with dune package management.

Here's an issue for tracking investigating why this change caused the regression: https://github.com/ocaml/dune/issues/10473

I'm proposing we merge this to revert to the old behaviour in the short term and then investigate why changing the order of updates caused the regression.

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -330,14 +330,12 @@ module Pkg = struct
   (* Given a list of packages, construct an env containing variables
      set by each package. Variables containing delimited lists of
      paths (e.g. PATH) which appear in multiple package's envs are
-     concatenated in the order of their associated packages in the
-     input list. Environment updates via the `exported_env` field
+     concatenated in the reverse order of their associated packages in
+     the input list. Environment updates via the `exported_env` field
      (equivalent to opam's `setenv` field) are applied for each
-     package in reverse order to the argument list so that packages
-     appearing earlier can overwrite the values of variables set by
-     packages appearing later. *)
+     package in the same order as the argument list. *)
   let build_env_of_deps ts =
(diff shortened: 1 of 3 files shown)


src/dune_rules/pkg_rules.ml
```diff
@@ -337,7 +337,7 @@ module Pkg = struct
      appearing earlier can overwrite the values of variables set by
      packages appearing later. *)
```
Leonidas-from-XIV, line 338: I think this comment needs to be adapted now because it states the reverse of what is happening and that's very confusing.

Leonidas-from-XIV (commented):
This is in fact very strange, especially as the ocamlfind build definitions don't do anything specific with the environment.

test/blackbox-tests/test-cases/pkg/opam-package-with-setenv.t
```diff
@@ -109,10 +109,10 @@ We currently have the following issues:
   > append_without_leading_sep="foo:bar" \
   > append_with_leading_sep="foo:bar" \
   > build_pkg deps-on-with-setenv-2
-  Hello from the other package!
-  Prepended without trailing sep:Prepended 2nd time without trailing sep
-  Prepended with trailing sep:Prepended 2nd time with sep
-  Appended 2nd time without leading sep:Appended without leading sep
-  Appended 2nd time with leading sep:Appended with leading sep
+  Hello from the second package!
```
rgrinberg, line 112: Could you leave a comment that this output is incorrect?

rgrinberg (approved):

src/dune_rules/pkg_rules.ml
gridbugs: Good catch. Fixed!

PR #10475, opened 2024-05-01
Download and use toolchain while building packages

gridbugs:
Download, build, and install the necessary toolchains when building packages that depend on versions of ocaml supported by the toolchains feature.

### DIFF
--- a/src/dune_pkg/toolchain.ml
+++ b/src/dune_pkg/toolchain.ml
@@ -19,6 +19,7 @@ end
 
 module Dir = struct
   let make_and_check_dir path =
+    let path = Path.outside_build_dir path in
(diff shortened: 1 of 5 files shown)


gridbugs:
I've tested this by building bonsai after applying the patches in https://github.com/ocaml/dune/pull/10476. I still needed to add the `ocaml-system` constraint to prevent the compiler from being downloaded and built by dune. This meant that for the initial solve I needed to have an opam switch present with the correct compiler installed. To test this feature, I renamed the switch so that there was no `ocaml` in my path, and I was still able to build the bonsai package with dune using the compiler from the toolchain.

gridbugs:
Closing this as it's subsumed by https://github.com/ocaml/dune/pull/10478.

PR #10476, opened 2024-05-01
Cherry-pick fixes to toolchains branch

gridbugs:
Two commits fixing some issues with package management that are yet to be merged into main, but which are necessary for package management to be usable in practice.


gridbugs:
Merging this without review to simplify subsequent changes, as it should be uncontraversial.

PR #10479, opened 2024-05-03
ci: add an action to build binaries

emillon:
This adds a manual workflow to get binaries for:

- linux (intel)
- mac (intel)
- mac (arm)

In addition, a combined zip is available.

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -1,0 +1,34 @@
+name: Binaries
+
+on:
+  workflow_dispatch:
+
+jobs:
+  binary:
+    name: Create
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: macos-13
+            name: mac-intel
+          - os: macos-14
+            name: mac-arm
+          - os: ubuntu-22.04
+            name: linux-intel
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: cachix/install-nix-action@v22
+      - run: nix build
+      - uses: actions/upload-artifact@v4
+        with:
+          path: result/bin/dune
+          name: dune-${{ matrix.name }}
+  combine:
+    runs-on: ubuntu-latest
+    needs: binary
+    steps:
+      - uses: actions/upload-artifact/merge@v4
+        with:
+          separate-directories: true


.github/workflows/binaries.yml
```diff
@@ -0,0 +1,34 @@
...
+    name: Create
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: macos-13
+            name: mac-intel
+          - os: macos-14
+            name: mac-arm
+          - os: ubuntu-latest
```
Leonidas-from-XIV, line 17:
```suggestion
          - os: ubuntu-20.04
```
to avoid surprises if `latest` changes and suddenly our binaries stop working for people overnight.

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: cachix/install-nix-action@v22
+      - run: nix build
+      - uses: actions/upload-artifact@v4
+        with:
+          path: result/bin/dune
+          name: dune-${{ matrix.name }}
+  merge:
```
Leonidas-from-XIV, line 28: Since I was surprised what it merges (turns out: merge not the git sense but rather a "merge actions into one" way)
```suggestion
  upload:
```

Leonidas-from-XIV (commented):
Cool that nix can be used to deal with both OSes!

.github/workflows/binaries.yml
emillon: the first build uploads a single-arch zip, and the this one uploads a combination zip, so i used `combine:`

Leonidas-from-XIV (approved):

emillon:
> I wonder if that'd allow distributing dune as a standalone binary, which could be especially useful for dune 4 and dune-pkg.

I should have mentioned that, but yes that's the motivation behind that workflow: this way we can distribute binaries so that people can easily test that without installing anything else.

> I wonder if it'd be interesting to build a statically linked binary too, e.g. dune-musl64 on x86_64 linux.

Sounds like a good idea, is there a "standard" way to do this with nix?

anmonteiro:
> Sounds like a good idea, is there a "standard" way to do this with nix?

It should be really easy with my overlays to build `pkgsCross.musl64.ocaml-ng.ocamlPackages_4_14.dune`.

emillon:
done in #10485

PR #10481, opened 2024-05-03
ci: run build on macOS x86_64 again

anmonteiro:
- macOS runners in GitHub Actions were upgraded to M1 macs (they migrated the `macos-latest` label)
- https://github.com/ocaml/dune/pull/10468 adapted the test suite to m1 macs, but we're not running tests on macOS x86 anymore

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -44,6 +44,10 @@ jobs:
           - ocaml-compiler: 5.1.x
             os: macos-latest
             skip_test: true
+          # macOS x86_64 (Intel)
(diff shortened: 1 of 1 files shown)


Leonidas-from-XIV (approved):
Seems reasonable to me, the only thing I don't like is that macOS builds usually take a long time as there are few runners available.

However, the ARM and AMD64 builds run in parallel and I could imagine with `macos-latest` moving that there's less users of the AMD64 builds thus faster to get a runner.

rgrinberg (approved):

emillon (approved):

PR #10484, opened 2024-05-06
chore(nix): update flakes

anmonteiro:


anmonteiro:
I don't get why warning 14 starts happening all of a sudden in the tests that disable it. Warning 26 is also disabled and is not showing up.

emillon:
We're installing melange through a git branch, maybe it's just that it has changed in the meantime?

anmonteiro:
I did update that git branch, so it must be that; I’ll revert the update when I get a chance and stage the update on a different branch.

I think until I do that I’ve broken dune tests.

emillon:
great, thanks. please use a hash instead of the branch name when you do :)

anmonteiro:
alright, I reverted the update and will be taking a look into it in melange.

also changed from branch to sha

emillon (approved):

Makefile
```diff
@@ -76,7 +76,7 @@ dev-depext:
 
 .PHONY: melange
 melange:
-	opam pin add -n melange.dev https://github.com/melange-re/melange.git#v4-414-dev
+	opam pin add -n melange.dev https://github.com/melange-re/melange.git#24e21cc42
```
jchavarri, line 79: Is it worth adding a comment to explain to which branch the hash is pointing to? Like in the nix case. For those of us that are not hip 😄

anmonteiro:
FWIW I fixed the issue we found here in https://github.com/melange-re/melange/pull/1118

PR #10485, opened 2024-05-06
ci: build a static linux binary

emillon:
The dynamic one uses a dynamic loader under a nix path so we can't really use it.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -13,6 +13,11 @@
       inputs.nixpkgs.follows = "nixpkgs";
       inputs.flake-utils.follows = "flake-utils";
     };
+    ocaml-overlays = {
+      url = "github:nix-ocaml/nix-overlays";
+      inputs.nixpkgs.follows = "nixpkgs";
+      inputs.flake-utils.follows = "flake-utils";
+    };
   };
   outputs =
     { self
@@ -20,6 +25,7 @@
     , nixpkgs
     , ocamllsp
     , melange
+    , ocaml-overlays
     }:
     flake-utils.lib.eachDefaultSystem (system:
     let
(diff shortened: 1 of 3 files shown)


emillon:
@anmonteiro I tried using pkgsCross to build a static version of dune, but it's building a dynamic version:

```
$ nix build .#dune-static
$ file result/bin/dune 
result/bin/dune: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /nix/store/73xmadm08bhnjq47vk5shw6h0rm2b7c6-musl-x86_64-unknown-linux-musl-1.2.3/lib/ld-musl-x86_64.so.1, not stripped
```

At least it's using musl, so I think it's just a matter of passing the right flags. I had a look at how `ocaml-option-static` works and it's just passing `LIBS=-static` to the compiler, which the static overlay seems to do. But since it's a cross-compilation I'm not sure whether there's something extra to do.

emillon:
Ah thanks that did it. We just make a bootstrap binary so we can't set flags directly from a profile, so I had to be creative.

flake.nix
```diff
@@ -37,6 +43,27 @@
         melange.overlays.default
         ocamllsp.overlays.default
       ];
+      use-this-dune-static = self: super: {
+        ocaml-ng =
+          super.ocaml-ng // {
+            ocamlPackages_4_14 = super.ocaml-ng.ocamlPackages_4_14.overrideScope (pself: psuper: {
```
emillon, line 49: is there a simpler way to do this override?

anmonteiro: maybe `ocamlPackages = super.ocaml-ng.ocamlPackages_4_14.overrideScope ...` and then use `pkgs.ocamlPackages` throughout?

PR #10487, opened 2024-05-06
Automatically lock on `build` if there is no lock dir

Leonidas-from-XIV:
This is more of an exploration of the toolchains work as it would automatically lock whenever any user that does not yet have a lock file calls `dune build`.

For an inclusion in the main dune this would need to be configurable on... something. The right criterion is up for discussion.

### DIFF
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -47,7 +47,21 @@ let run_build_system ~common ~request =
       Cached_digest.invalidate_cached_timestamps ();
       let* setup = Import.Main.setup () in
       let request =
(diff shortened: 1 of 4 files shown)


gridbugs (approved):
This is a good start but I think we also need to handle the case where the dependencies in dune-project are updated such that they no longer match what's in the lockdir. It should never be necessary to explicitly create a lockdir.

Happy for that to come in a later change though.

Leonidas-from-XIV:
Closing in favor of #10546 which takes the suggestion from @emillon and requires an existing lock to trigger re-locking.

PR #10491, opened 2024-05-08
refactor(pkg): keep the files dir in dune_pkg

rgrinberg:
Since this directory is created along with the lock directory, we might as well keep the path there.

### DIFF
--- a/src/dune_pkg/lock_dir.mli
+++ b/src/dune_pkg/lock_dir.mli
@@ -31,6 +31,7 @@ module Pkg : sig
 
   val equal : t -> t -> bool
   val decode : (lock_dir:Path.Source.t -> Package_name.t -> t) Decoder.t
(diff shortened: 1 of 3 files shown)


src/dune_pkg/lock_dir.mli
```diff
@@ -31,6 +31,7 @@ module Pkg : sig
 
   val equal : t -> t -> bool
   val decode : (lock_dir:Path.Source.t -> Package_name.t -> t) Decoder.t
+  val files_dir : Package_name.t -> Path.Source.t -> Path.Source.t
```
gridbugs, line 34: This needs a comment or a label on the `Path.Source.t` arg. Otherwise it's not clear what to pass as the `Path.Source.t` arg.

gridbugs (approved):

PR #10494, opened 2024-05-09
feat(melange): support lint in `melange.emit`

### LINKED ISSUE #7353: melange: support `lint` in `melange.emit`
At the moment, `lint` is not supported in `melange.emit` stanza.

anmonteiro:
support lint in `melange.emit`

closes https://github.com/ocaml/dune/issues/7353


test/blackbox-tests/test-cases/melange/lint.t/correct/dune
jchavarri, line 1: Why not using the "single file" approach like all other Melange tests? It seems Dune tests are moving in that direction (and imo it makes easier to understand what's going on in the tests quickly).

jchavarri (approved):
Only a comment about the cram test shape.

test/blackbox-tests/test-cases/melange/lint.t/correct/dune
anmonteiro: I copied it from the other `lint` test in the codebase, and it has the same structure.

PR #10495, opened 2024-05-09
doc: js targets & the `@all` alias

rgrinberg:
Describe that the `all` alias doesn't include js targets by default and what to do to get back the old behavior.

### DIFF
--- a/doc/reference/aliases.rst
+++ b/doc/reference/aliases.rst
@@ -64,6 +64,11 @@ Some aliases are defined and managed by Dune itself.
 
 This alias corresponds to every known file target in a directory.
 
+Since version 2.0 of the dune language, JS targets of executables are no longer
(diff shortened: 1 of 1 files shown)


anmonteiro (approved):
I didn't know this. nice

doc/reference/aliases.rst
```diff
@@ -64,6 +64,11 @@ Some aliases are defined and managed by Dune itself.
 
 This alias corresponds to every known file target in a directory.
 
+Since version 2.0 of the dune language, JS targets of executables are no longer
+included in the `all` alias by default. To get back the old behavior of
+including the js targets in `all`, one can add the ``js`` target to the
```
anmonteiro, line 69:
```suggestion
including the JS targets in `all`, one can add the ``js`` target to the
```

gridbugs (approved):

PR #10497, opened 2024-05-09
Add toolchain to env in more places

gridbugs:
This fixes the issue where we couldn't build `ocamlformat` with toolchains.

In the process of debugging that issue I noticed that if a dune project doesn't depend on any opam packages (e.g. if it has no deps or all its deps are pinned) then the lockdir won't pin a compiler version and toolchains won't work which I think is undesirable as the initial state of a project is to have no deps. I've addressed this for now by assuming the latest known version of the toolchain in such cases, but we may want to revisit this decision.

I've also updated the way that pkg_rules determine the ocmal toolchain such that they never produce a toolchain backed by the regular compiler package, as we know this will never work (at least until the compiler is relocatable). Now it either returns a dune toolchain directory or tries to use the system compiler.

### DIFF
--- a/bin/toolchain.ml
+++ b/bin/toolchain.ml
@@ -4,7 +4,7 @@ module Toolchain = Dune_pkg.Toolchain
 module Version = struct
   include Toolchain.Version
 
-  let conv = Arg.enum all_by_string
+  let conv = Arg.enum (List.map all ~f:(fun version -> to_string version, version))
 end
 
 module Get = struct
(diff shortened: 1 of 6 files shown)


Leonidas-from-XIV (approved):
> I've addressed this for now by assuming the latest known version of the toolchain in such cases, but we may want to revisit this decision.

This seems entirely reasonable I would say. In the future any non-test release would be what I as a user would expect.

Issue #10272, opened 2024-03-15
pkg: Pinning a package to a non-existent github repo prompts for a github login


gridbugs:
If the url of a pin stanza is a non-existent repo, running `dune pkg lock` will print `Username for 'https://github.com':`. Dune will then hang and not respond to ctrl+c.

E.g.
```
(pin
 (url "git+https://github.com/gridbugs/not-a-repo")
 (package
  (name foo)
  (version 1.0)))
```

rgrinberg:
I suppose we could use: https://stackoverflow.com/questions/23563062/how-do-i-force-git-not-to-prompt-for-credentials for now

emillon:
This happens because there's no way to tell if the repo does not exist, or if it is a private one. So I guess a related question is, do we support private repositories in pin?

rgrinberg:
It's possible to configure one's git client to authenticate itself automatically. It would be nice to offer the user that option interactively as part of a dune build, but it's not so easy, so I'd say the priority is low.

emillon:
Do you know if opam supports this?

rgrinberg:
I don't know but perhaps @kit-ty-kate does

kit-ty-kate:
I believe opam just leaves stdin open so it should be possible to authenticate but nothing is done automatically

PR #10498, opened 2024-05-09
fix(pkg): do not prompt for passwords with git

rgrinberg:
Simple fix for https://github.com/ocaml/dune/issues/10272


gridbugs (approved):
Awesome!

PR #10501, opened 2024-05-09
test(melange): add a test showing dependent re-compilation

anmonteiro:
this is the last test case we were missing from https://github.com/ocaml/dune/issues/7164

I feel like there's a missing optimization here, where only `lib/foo.ml` should be recompiled, but instead all its dependents end up getting recompiled, too.


jchavarri (approved):
We might be adding deps on `ml` files when it should be only `mli`. I wonder if there's a similar test for `executable`s?

anmonteiro:
This is likely because we depend on the `.cmj` rather than any source dependencies.

PR #10512, opened 2024-05-10
fix(pkg): remove External_copy

### LINKED ISSUE #9700: Remove the `copy` constructor from `Dune_pkg.Source.t`

### LINKED ISSUE #10232: Changes made to non-dune pinned packages are not visible until `dune clean` is run

rgrinberg:
Make it the same as file://. Both of these constructors mean the same thing, so let's keep one of them. This simplification also fixes dune reacting to changes made to pinned packages.

There's still some ugliness around the types. In particular, we should be more careful about where we expect the url to point to a file vs a directory.

Note that I still haven't yet removed the separate constructors in the frontend. That's going to conflict with some PR's I have, so I will do that later.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/pin-stanza/update-non-dune-local-pin.t
+++ b/test/blackbox-tests/test-cases/pkg/pin-stanza/update-non-dune-local-pin.t
@@ -1,6 +1,6 @@
-This demonstrates an issue when pinning a non-opam package. After attempting to
-build the package, changes to the package are only picked up by dune after
-running `dune clean`, even if building the package failed.
+This demonstrates pinning a non-opam package and then modifying its sources.
+Whenever the sources are modified, dune should rebuild the package in the
+workspace where it's locked.
 
   $ . ../helpers.sh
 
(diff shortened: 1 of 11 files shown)


test/blackbox-tests/test-cases/pkg/pin-stanza/update-non-dune-local-pin.t
```diff
@@ -52,8 +52,8 @@ Update the message that gets printed while building foo:
 
 The change to the package is ignored:
   $ dune build 2>&1 | grep -v make
-  echo aaa
-  aaa
+  echo bbb
```
gridbugs, line 55: Looks like this fixes https://github.com/ocaml/dune/pull/10512. Can you update the comments in this test as they are now incorrect.

gridbugs (approved):

PR #10513, opened 2024-05-10
Make scheduler type mutable instead of containing fields with ref types

pmwhite:
For better or worse, the current state of the scheduler is that at any point in time, it has exactly one conceptual "current build". Putting the cancellation token inside a fiber-local variable means there could be more than one cancellation token flying around, which is misleading.

This patch does not preclude us from eventually making the scheduler and cancellation flow more build-agnostic, but until that happens, we I think this is a better state to be in.

@rgrinberg We have an analogous internal patch to this one, so this won't cause the scheduler to diverge more than it has already. I'll wait to merge the internal patch until you get the opportunity to object to this direction. 

From our in-person discussion, I think you were imagining we would the cancellation token would get taken out of the scheduler altogether in favor of being set in its won fiber-local var in all the places that care about it. I'm not against ultimately getting there, but I think it is a larger jump than this PR.

### DIFF
--- a/src/dune_engine/scheduler.ml
+++ b/src/dune_engine/scheduler.ml
@@ -816,41 +816,18 @@ end = struct
   ;;
 end
 
-(** All fields of [t] must be immutable. This is because we re-create [t] every
-    time we start a new build to locally set the [cancel] field. However, all
-    instances of [t] must share all other fields, in particular the references
-    such as [status].
-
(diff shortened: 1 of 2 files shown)


pmwhite:
The failing tests are demonstrating that I need to think this through a bit more.

rgrinberg (approved):

PR #10518, opened 2024-05-12
scheduler: extract memo invalidation from the status field

pmwhite:
The accumulation of the invalidation is orthogonal to what the build status is. Regardless of how the build progresses, we should always continually combine new build input change into the current invalidation, and we should only set the invalidation to empty when we also call Memo.reset with the current invalidation.

At the same time, it also makes sense to cleanup the handling of the `build_input_change` ivar. There is no need for it to be an option that is initialized on-demand, since it is easy and cheap to just create an initial ivar. In addition, we shouldn't care what the current build status is when waiting for a file to change. (I suspect that some of the reason for the complication is due to the `--react-to-insignificant-file changes` flag that no longer exists.) I think it should be handled in a similar way to the memo invalidation: it should be accumulated and reset at the same time. Here is an invariant: `build_inputs_changed` should be empty if and only if the memo invalidation is empty.

This patch doesn't claim to fix any bugs; it's just intended as a code cleanup.

### DIFF
--- a/src/dune_engine/scheduler.ml
+++ b/src/dune_engine/scheduler.ml
@@ -668,12 +668,12 @@ end
 type status =
   | (* We are not doing a build. Just accumulating invalidations until the next
        build starts. *)
-    Standing_by of { invalidation : Memo.Invalidation.t }
+    Standing_by
   | (* Running a build *)
     Building of Fiber.Cancel.t
   | (* Cancellation requested. Build jobs are immediately rejected in this
        state *)
-    Restarting_build of Memo.Invalidation.t
+    Restarting_build
 
 module Build_outcome = struct
   type t =
@@ -782,17 +782,49 @@ end = struct
   ;;
 end
 
+module Trigger : sig
+  (** Conceptually, a [unit Fiber.Ivar.t] that can be filled idempotently.
+      Ivars cannot, in general, have idempotent fill operations, since the second
+      fill might have different data in it than the first. Since this type only
+      contains unit data, we can be sure that every fill will be (). In fact, we
+      don't even both accepting data as the parameter of [trigger], since we
+      already know what it must be. *)
+  type t
+
+  val create : unit -> t
+  val trigger : t -> Fiber.fill list
+  val wait : t -> unit Fiber.t
+end = struct
+  type t =
+    { ivar : unit Fiber.Ivar.t
+    ; mutable filled : bool
+    }
+
+  let create () = { ivar = Fiber.Ivar.create (); filled = false }
+
+  let trigger t =
+    if t.filled
+    then []
+    else (
+      t.filled <- true;
+      [ Fiber.Fill (t.ivar, ()) ])
+  ;;
+
+  let wait t = Fiber.Ivar.read t.ivar
+end
+
 type t =
   { config : Config.t
   ; alarm_clock : Alarm_clock.t Lazy.t
   ; mutable status : status
+  ; mutable invalidation : Memo.Invalidation.t
   ; handler : Handler.t
   ; job_throttle : Fiber.Throttle.t
   ; events : Event.Queue.t
   ; process_watcher : Process_watcher.t
   ; file_watcher : Dune_file_watcher.t option
   ; fs_syncs : unit Fiber.Ivar.t Dune_file_watcher.Sync_id.Table.t
-  ; mutable wait_for_build_input_change : unit Fiber.Ivar.t option
+  ; mutable build_inputs_changed : Trigger.t
   ; mutable cancel : Fiber.Cancel.t
   ; thread_pool : Thread_pool.t
   }
@@ -908,6 +940,7 @@ let prepare (config : Config.t) ~(handler : Handler.t) ~events ~file_watcher =
          here because then the non-polling mode would run in "Standing_by"
          mode, which is even weirder. *)
       Building cancel
+  ; invalidation = Memo.Invalidation.empty
   ; job_throttle = Fiber.Throttle.create config.concurrency
   ; process_watcher
   ; events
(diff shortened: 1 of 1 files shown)


src/dune_engine/scheduler.ml
```diff
@@ -782,17 +782,44 @@ end = struct
   ;;
 end
 
+module Trigger : sig
+  (** Conceptually, a [unit Fiber.Ivar.t] that may be filled more than once. *)
```
rgrinberg, line 786: More precisely, fills past the first one are ignored. So I would call this idempotent instead.

src/dune_engine/scheduler.ml
```diff
@@ -782,17 +782,44 @@ end = struct
...
+  type t
+
+  val create : unit -> t
+  val trigger : t -> Fiber.fill list
```
rgrinberg, line 790: How about returning an option? Would make the interface more precise.

src/dune_engine/scheduler.ml
```diff
@@ -782,17 +782,44 @@ end = struct
...
+end = struct
+  type t =
+    { ivar : unit Fiber.Ivar.t
+    ; mutable filled : bool
+    }
+
+  let create () = { ivar = Fiber.Ivar.create (); filled = false }
+
+  let trigger t =
+    if not t.filled
```
rgrinberg, line 801: If you invert this `if`, it will be easier to read and the shorter clause will be on top.

src/dune_engine/scheduler.ml
pmwhite: I wouldn't say using an option is more precise. Option and list communicate different things. In this case, the user shouldn't care whether zero, one, or many fills in the list; all it is supposed to do is just forward it along to the fiber scheduler. If it were an option, we would have to convert it to a list later.

src/dune_engine/scheduler.ml
pmwhite: Yes, good point. Note that since the data is unit, ignoring a fill is indistinguishable from not ignoring it. I've expanded the comment with more nuance. Feel free to edit it a bit if you prefer a different wording.

src/dune_engine/scheduler.ml
pmwhite: Oops, yup, I don't know why I did it this way.

src/dune_engine/scheduler.ml
rgrinberg: Why would we convert it to a list? We'd just cons it to an existing list if something was there?

This issue isn't crucial but I'm not convinced by your argument for using a list. If the user shouldn't care about zero, one or many fills, then returning the list goes against your intention. You're now making it possible for the user to discriminate multiple fills from a single one. To actually get to your stated goal, you would need to introduce some abstract type that only allows for appending fills. Now I don't see that as needed because the code is trivial, but returning a richer type to the user gives more rope to hang himself, not less.

rgrinberg:
> The accumulation of the invalidation is orthogonal to what the build status is. Regardless of how the build progresses, we should always continually combine new build input change into the current invalidation, and we should only set the invalidation to empty when we also call Memo.reset with the current invalidation.

Previously, we had the invariant that when we're building something we have no pending invalidation. That seems kind of an important invariant to me.

src/dune_engine/scheduler.ml
pmwhite: EDIT: Oh, I see you resolved this comment. Feel free to ignore below, but I'm posting it in cases it's enough to convince you. :)

`option` says "there is one fill or there is no fill." `list` says "I won't tell
you how fills there are; be prepared for any amount." In that sense, `list` is a
bit more abstract because the user gets less information about the data being
returned.

It's true that now `list` gives the user more freedom to misunderstand the argument, but it would look pretty weird if they did, and there is only one callsite, so I'm not too concerned about that.

Perhaps the main reason I picked `list` is that it has a convenient append operation. Unless I've misunderstood the suggested change, using an `option` would require a match expression when calling `trigger`; it's not so bad, but it seems a bit noisy to me.

pmwhite:
> Previously, we had the invariant that when we're building something we have no pending invalidation. That seems kind of an important invariant to me.

That invariant is still upheld, unless I'm mistaken. I think what you're saying that the types used to enforce the invariant, and with this change they don't; is that right?

But I don't think the types were enforcing anything. Just because `t.status` is in the `Building` state doesn't mean there were no invalidations that we should respond to; we might have just ignored some.

I'll admit that previously the types made it natural to update both the invalidation and the status at the same time, since it was in the same variable, but it also required a little bit more case analysis that pleasantly goes away with this change.

My ultimate plan is to get rid of the `status` field so that all these invariants are upheld by definition: rather than having to keep `status` in sync with `invalidation`, and `cancel` , we'll define "when we're building something" to mean "the cancel hasn't fired and the invalidation is empty."

The refactor I'm trying to make is similar to going from `'a list option` to `'a list` because the `None` case always maps to `[]` (at least in this hypothetical scenario).

rgrinberg (approved):
I tried to think about how we could misuse an invalidation when `status = Building` but couldn't think of anything concrete.W ould it be helpful to add an assertion that the invalidations are empty when transitioning from `Building` to `Standing_by`?

pmwhite:
That might be a good assertion to have, but if we merge the [sequel to this PR](https://github.com/ocaml/dune/pull/10519) or [the alternative sequel](https://github.com/ocaml/dune/pull/10548), then it either wouldn't make sense because the status field would be gone, or it wouldn't be true anymore, because we'd actually leave the build in the `Building` state after an invalidation happens. How about we wait to see how the next PR ends up before adding the assertion to this PR?

PR #10520, opened 2024-05-13
scheduler: remove `status` field

pmwhite:
Based off of https://github.com/ocaml/dune/pull/10519

Previous changes have moved all the useful information from the status field into other fields, such as `cancel` and `invalidation`, so now we can get rid of the field altogether.


pmwhite:
I included the commits in https://github.com/ocaml/dune/pull/10519, so closing this PR.

PR #10521, opened 2024-05-13
pkg: don't use "compiler" flag to identify compiler

gridbugs:
Opam's "compiler" flag is used too broadly to accurately identify packages containing ocaml compilers. For example this flag is set for compiler options packages such as ocaml-option-flambda which configure the ocaml-variants package, effectively preventing the use of compiler options packages, as ocaml-variants also has the "compiler" flag set, and dune only permits a single compiler package in a solution.

This change fixes this problem by using the presence of the "ocaml-core-compiler" conflict class to identify compiler packages.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -537,6 +537,17 @@ let available_depopts solver_env version_by_package_name (opam_package : OpamFil
         else None))
 ;;
 
+(* Heuristic to determine whether a package is an ocaml compiler *)
+let opam_file_is_compiler (opam_package : OpamFile.OPAM.t) =
+  (* Identify compiler packages by using the fact that all compiler
+     Packages declare conflicts with other compiler packages. note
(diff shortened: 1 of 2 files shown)


Leonidas-from-XIV (commented):
The implementation and reasoning sounds very sensible, probably our generated test compiler packages need to gain this conflict class (which makes sense, to be more accurate to what opam-repository actually does) and then its good to merge.

gridbugs:
Ok I've fixed the test

rgrinberg:
> and dune only permits a single compiler package in a solution.

I suppose we could also lift this limitation. This fix should work as well though

rgrinberg (approved):

Leonidas-from-XIV:
> I suppose we could also lift this limitation.

I'm curious, how would that work? My fear is that if you mix and match compilers in your build, you get artifacts that aren't compatible with each other.

rgrinberg:
The conflict class already prevents invalid build plans, so that could never happen regardless.

PR #10534, opened 2024-05-15
Stop treating toolchains as packages in pkg rules

gridbugs:
This grew out of the work I'm doing to read toolchain information from the opam repo rather than hardcoding it (https://github.com/tarides/team-build-system/issues/11), but the main benefit of this change is that it removes a bunch of special cases I needed to add so that packages containing toolchains were ignored by most of the package rules.

It's also going to make it easier to thread the extra information needed to represent the `ocaml-variants` toolchain to the parts of the code where it's needed in a later change.

### DIFF
--- a/src/dune_engine/scheduler.mli
+++ b/src/dune_engine/scheduler.mli
@@ -143,9 +143,9 @@ val cancel_current_build : unit -> unit Fiber.t
 
 val inject_memo_invalidation : Memo.Invalidation.t -> unit Fiber.t
 
-(** [sleep duration] wait for [duration] to elapse. Sleepers are checked for
-    wake up at a rate of once per 0.1 seconds. So [duration] should be at least
-    this long. *)
+(** [sleep duration] wait for [duration] seconds to elapse. Sleepers
+    are checked for wake up at a rate of once per 0.1 seconds. So
+    [duration] should be at least this long. *)
 val sleep : float -> unit Fiber.t
 
 val stats : unit -> Dune_stats.t option Fiber.t
--- a/src/dune_pkg/flock.ml
+++ b/src/dune_pkg/flock.ml
@@ -1,0 +1,149 @@
+open! Stdune
+module Flock = Dune_util.Flock
+module Scheduler = Dune_engine.Scheduler
+
+type t =
+  { flock : Flock.t
+  ; lock_path : Path.t
+  }
+
+(* Global mutable set of names, used to prevent printing "waiting for
+   lock" messages multiple times when multiple concurrent fibers try
+   to take a lock at the same time. *)
+module Global_waiting_names = struct
+  let state = lazy (String.Table.create 1)
+
+  (* add a name to the set, returning [true] iff the name wasn't
+     already in the set *)
+  let add name =
+    let state = Lazy.force state in
+    String.Table.add state name () |> Result.is_ok
+  ;;
+
+  let remove name =
+    let state = Lazy.force state in
+    String.Table.remove state name
+  ;;
+end
+
+let attempt_to_lock { flock; lock_path } ~name_for_messages ~timeout_s =
+  let open Fiber.O in
+  let current_dune_pid = Unix.getpid () in
+  let rec loop timeout_s =
+    match Flock.lock_non_block flock Flock.Exclusive with
+    | Error e -> Fiber.return @@ Error e
+    | Ok `Success ->
+      Global_waiting_names.remove name_for_messages;
+      Fiber.return (Ok `Success)
+    | Ok `Failure -> handle_failure timeout_s
+  and handle_failure timeout_s =
+    let locked_by_pid = int_of_string (Io.read_file lock_path) in
+    let sleep_duration_s = 0.1 in
+    let remaining_duration_s = timeout_s -. sleep_duration_s in
+    if remaining_duration_s <= 0.0
+    then Fiber.return (Ok `Timeout)
+    else (
+      if locked_by_pid <> current_dune_pid && Global_waiting_names.add name_for_messages
+      then
+        (* Only print this message if the dune process that holds the
+           lock isn't the current process and this is the first time
+           that the current process has failed to take the lock since
+           the last time it successfully took the lock. This prevents
+           the situation where multiple fibers all attempt to take the
+           lock concurrently while it's held by another process from
+           causing the following message from being printed multiple
+           times. *)
+        User_message.print
+          (User_message.make
+             [ Pp.textf
+                 "Waiting for another instance of dune (pid %d) to release the lock for \
+                  the resource %S..."
+                 locked_by_pid
+                 name_for_messages
+             ]);
+      let* () = Scheduler.sleep sleep_duration_s in
+      loop remaining_duration_s)
+  in
+  loop timeout_s
+;;
+
+let with_flock lock_path ~name_for_messages ~timeout_s ~f =
+  let open Fiber.O in
+  let parent = Path.parent_exn lock_path in
+  Path.mkdir_p parent;
+  let fd =
+    Unix.openfile
+      (Path.to_string lock_path)
+      [ Unix.O_CREAT; O_WRONLY; O_SHARE_DELETE; O_CLOEXEC ]
+      0o600
+  in
+  let out = Unix.out_channel_of_descr fd in
+  let flock = Flock.create fd in
+  let current_dune_pid = Unix.getpid () in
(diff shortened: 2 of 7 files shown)


src/dune_rules/pkg_rules.ml
```diff
@@ -355,38 +357,11 @@ module Pkg = struct
...
-    if is_compiler t
-    then Dune_pkg.Toolchain.Version.of_package_version t.info.version
-    else None
-  ;;
-
-  let is_compiler_with_version_in_toolchain_dir t =
-    Option.is_some (as_compiler_with_version_in_toolchain_dir t)
-  ;;
-
   (* Returns the version of the compiler toolchain dependen on by this
```
Leonidas-from-XIV, line 360: Tangentially to this code change but there's a typo here. Worth fixing :-)

src/dune_rules/pkg_rules.ml
```diff
@@ -264,6 +265,7 @@ module Pkg = struct
     ; build_command : Build_command.t option
     ; install_command : Dune_lang.Action.t option
     ; depends : t list
+    ; toolchain_depends : Dune_pkg.Toolchain.Version.t list
```
Leonidas-from-XIV, line 268: Can you explain why this is a list? Naïvely I would expect this to be one single compiler that was picked.

src/dune_pkg/toolchain.ml
```diff
@@ -278,42 +282,45 @@ let install version =
...
-    let* () = build version in
-    log_print Details @@ Pp.text "Installing...";
-    let+ () = install version in
-    log_print Success
-    @@ Pp.textf
-         "Success! Compiler toolchain version %s installed to %s."
-         (Version.to_string version)
-         (Version.target_dir version |> Path.Outside_build_dir.to_string))
+let get =
+  let mutex = Fiber.Mutex.create () in
```
Leonidas-from-XIV, line 286: This probably needs an even better mutex to guard against multiple processes trying to create the same compiler. You could take a look at the `git-repo` mutex, which guards against exactly this.

src/dune_rules/pkg_rules.ml
```diff
@@ -1658,35 +1595,34 @@ module Compiler_dependency = struct
...
+       | `Toolchain toolchain_version -> Toolchain toolchain_version
        | `System_provided -> System_provided
-       | `Inside_lock_dir pkg -> Pkg pkg)
+       | `Inside_lock_dir pkg -> Inside_lock_dir pkg)
   ;;
 
   let toolchain_version = function
+    | Toolchain toolchain_version -> Some toolchain_version
+    | Inside_lock_dir pkg ->
+      let module Toolchain = Dune_pkg.Toolchain in
```
Leonidas-from-XIV, line 1606: I would probably put this on the toplevel, since in the Dune codebase we rarely do local module bindings. Minor point.

src/dune_rules/pkg_rules.ml
gridbugs: It's a list to represent the rare but possible situation where a package depends on both `ocaml-base-compiler` and `ocaml-variants`. However I think elsewhere in the code we make this case an error, so I'll change it to be an option instead for simplicity.

src/dune_rules/pkg_rules.ml
```diff
@@ -1167,38 +1133,47 @@ end = struct
...
+           ; install_command
+           ; depends
+           ; toolchain_depends
+           ; paths
+           ; info
+           ; files_dir
+           ; exported_env = []
+           }
+         in
+         let+ exported_env =
```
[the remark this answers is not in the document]
gridbugs: I'm not entirely sure what this code does to be honest. It's in the diff because I changed its indentation.

src/dune_pkg/toolchain.ml
gridbugs: Solved by moving the flock used in the rev store to its own file and reusing it here.

gridbugs:
@Leonidas-from-XIV can you take a second look at this please

src/dune_engine/scheduler.mli
```diff
@@ -143,9 +143,9 @@ val cancel_current_build : unit -> unit Fiber.t
 
 val inject_memo_invalidation : Memo.Invalidation.t -> unit Fiber.t
 
-(** [sleep duration] wait for [duration] to elapse. Sleepers are checked for
-    wake up at a rate of once per 0.1 seconds. So [duration] should be at least
-    this long. *)
+(** [sleep duration] wait for [duration] seconds to elapse. sleepers
+    are checked for wake up at a rate of once per 0.1 seconds. so
```
Leonidas-from-XIV, line 147: Despite the useful explanation that the duration is actually seconds, it seems some "convert to lower case" went wrong here.

src/dune_pkg/flock.ml
```diff
@@ -0,0 +1,75 @@
...
+module Flock = Dune_util.Flock
+module Scheduler = Dune_engine.Scheduler
+
+let rec attempt_to_lock flock lock ~timeout_s =
+  let open Fiber.O in
+  match Flock.lock_non_block flock lock with
+  | Error e -> Fiber.return @@ Error e
+  | Ok `Success -> Fiber.return (Ok `Success)
+  | Ok `Failure ->
+    let sleep_duration_s = 0.1 in
```
Leonidas-from-XIV, line 11: This could possibly be pulled out of the function. I don't know, even in the original code it is an admittedly ugly bit.

src/dune_pkg/flock.ml
```diff
@@ -0,0 +1,75 @@
...
+;;
+
+let with_flock lock_path ~name_for_messages ~timeout_s ~f =
+  let open Fiber.O in
+  let parent = Path.parent_exn lock_path in
+  Path.mkdir_p parent;
+  let fd =
+    Unix.openfile
+      (Path.to_string lock_path)
+      [ Unix.O_CREAT; O_WRONLY; O_SHARE_DELETE; Unix.O_CLOEXEC ]
```
Leonidas-from-XIV, line 27: That second `Unix` there can be removed.

src/dune_pkg/toolchain.ml
```diff
@@ -279,41 +287,46 @@ let install version =
...
-    let+ () = install version in
-    log_print Success
-    @@ Pp.textf
-         "Success! Compiler toolchain version %s installed to %s."
-         (Version.to_string version)
-         (Version.target_dir version |> Path.Outside_build_dir.to_string))
+  Flock.with_flock
+    (Version.flock_path version)
+    ~name_for_messages:(sprintf "toolchain version %s" (Version.to_string version))
+    ~timeout_s:infinity
```
Leonidas-from-XIV, line 293: While it is hard to know how long it should wait for another dune building the compiler, I do think that erroring out if the other instance gets stuck on something could be the better approach. Maybe it should not attempt to acquire the lock multiple times at and just print something along the lines of "well, some other instance (pid %d) is building the toolchain compiler".

WDYT?

Leonidas-from-XIV (commented):
Looks fairly good with the new generalized lock.

src/dune_engine/scheduler.mli
gridbugs: How strange. Good catch! Fixed

src/dune_pkg/toolchain.ml
gridbugs: I'll definitely update this to print a message if acquiring the lock failed the first time around. I tried this earlier but gave up because it printed a bunch of seemingly spurious messages, but I just realized that it was because multiple fibers within the same instance of dune try to take the lock at the same time. I'll make it so that it only prints a message if the lock is currently taken by a different process.

gridbugs: As for timing out, I'm happy to just let the user decide to kill it if it takes too long. I'm not sure about erroring out if another instance of dune is taking a long time to install the compiler. I don't want to make the experience worse for people with slow computers or slow internet where installing dune will take longer than expected. We could set a really conservative time limit like an hour, but my preference would be to make it clear to the user that this instance of dune is waiting for another instance of dune to finish building the compiler, and leave it up to them whether to kill the current instance of dune or wait for the other instance to finish.

Leonidas-from-XIV (approved):

src/dune_pkg/flock.ml
gridbugs: I've split this into two mutually recursive closures to avoid needing to pass as much info from one recursive call to the next.

Issue #9333, opened 2023-11-30
Add support for the new -H <dir> argument in OCaml 5.2


kit-ty-kate:
Would fix:
* https://github.com/ocaml/dune/issues/2733
* https://github.com/ocaml/dune/issues/4963

https://github.com/ocaml/ocaml/pull/12246 added support for the `-H <dir>` argument, which, similarly to `-I <dir>` adds the given directory to the list of directories the compiler will look into for interfaces. However contrary to `-I`, `-H` forbids explicit calls to the modules and instead only resolves their type.

Using this would fix a foundamental flaw in the `(implicit_transitive_deps false)` option which causes regular annoyance for users and opam-repository maintainers alike.

Dune should be able to detect which version of the OCaml compiler it’s using, and if it’s OCaml >= 5.2, then instead of not adding any of the implicit directories in `(implicit_transitive_deps false)` mode, it would instead use `-H`, the same way `-I` is used with `(implicit_transitive_deps true)`

nojb:
I took a quick look. A clean way to approach this would be if the calculation of the transitive closure of a set of libraries (eg `Lib.closure`) could return a list `(bool * lib) list` indicating whether each `lib` is in the root set of simply reachable by transitiveness. Maybe this is easy, I haven't checked.

PR #10535, opened 2024-05-15
Use new hidden deps Merlin support for handling implicit-transitive-deps false

voodoos:
With release `5.2` the compiler introduces a new flag `-H` to introduce dependencies that are required for linking but should be hidden from the currently built module. Merlin for 5.2 follows with two new configuration directives: `BH` and `SH` that allows to provide hidden build and source paths. 

This PR effectively takes advantage of these new directives to correctly communicate hidden dependencies to Merlin.

All versions of Merlin since 4.6 will accept that flag, but only the preview versions for 5.2 can actually take advantage of this additional information.


nojb:
Hello @voodoos, thanks for this PR. Note that #9333 is being worked on at the moment (cc @MA0010). Once that is done, you will be able to simplify the patch because the lists of requires given by `Compilation_context.requires` will be annotated with a flag indicating whether each dependency is direct or transitive (= hidden), and you won't need to do this computation "just" for Merlin. Moreover, I think it would be great if we could merge this PR at the same time that Dune itself is able to support `-H`.

Accordingly, I suggest to suspend this PR until #9333 is done; does that sound OK? I hope that there will be a PR for #9333 in the next week or so.

nojb:
@voodoos #10644 has been merged; so it should now be possible to finalize this PR by using `Compilation_context.requires_hidden`. Thanks for waiting!

Leonidas-from-XIV:
Hi @voodoos could you rebase the PR? We were talking about this in the last dune-dev meeting and it would be nice to have it as part of the 3.17 release.

test/blackbox-tests/test-cases/merlin/implicit-transitive-deps.t/run.t
```diff
@@ -3,31 +3,36 @@
...
   $ FILE=$PWD/bin/main.ml
   $ printf "(4:File%d:%s)" ${#FILE} $FILE | dune ocaml-merlin |
-  > sed -E "s/[[:digit:]]+:/\?:/g" | tr '(' '\n' | grep ":S?"
+  > sed -E "s/[[:digit:]]+:/\?:/g" | tr '(' '\n' | grep -E ":[BS]H?\?"
+  ?:B?:$TESTCASE_ROOT/_build/default/bin/.main.eobjs/byte)
+  ?:B?:$TESTCASE_ROOT/_build/default/src/lib1/.lib1.objs/byte)
   ?:S?:$TESTCASE_ROOT/bin)
   ?:S?:$TESTCASE_ROOT/src/lib1)
-  ?:S?:$TESTCASE_ROOT/src/lib2)
-  ?:S?:$TESTCASE_ROOT/src/lib_dep)
```
voodoos, line 20: @Leonidas-from-XIV  I will have a look on Monday, but I think this is a regression for users of OCaml <5.2

voodoos: Actually, since `lib2` is hidden, it should not be a part of the standard source path for Merlin, this was a regression introduced with the indexing rules, and the current behavior should be the correct one. The only correct way to handle these are via the new `SH` and `BH` directives, which are specific to 5.2.

voodoos:
@Leonidas-from-XIV rebased and signed-off

src/dune_rules/merlin/merlin.ml
```diff
@@ -556,6 +613,21 @@ module Unprocessed = struct
       ~f:(pp_flags ctx ~expander t.config.libname)
   ;;
 
+  let add_lib_dirs sctx mode ~init libs =
```
Leonidas-from-XIV, line 616: Can you give `init` a better (outside) name? It is used as a named argument to `List.fold_left` but in the context of `val add_lib_dirs : Super_context.t -> Mode.t -> init: -> ...`  is quite nondescript.

Leonidas-from-XIV (commented):
Could you please add a changelog entry? This is one of the highlights of the release, would be good for people to be aware of it :)

src/dune_rules/merlin/merlin.ml
voodoos: I now make the union later in [b4641d9](https://github.com/ocaml/dune/pull/10535/commits/b4641d937e68d578e6417de90416a731656e1f29), and since `source_dir` was always empty I removed it altogether.

test/blackbox-tests/test-cases/overlapping-deps.t
```diff
@@ -81,7 +81,7 @@ We can fix the error by allow overlapping dependencies:
   > EOF
   $ OCAMLPATH=$PWD/../external/_build/install/default/lib dune build %{cma:proj2/bar} --root .
 
-Strangely, the error disappears if we remove the source for the bar lib:
+The error does not disappears if we remove the source for the bar lib:
```
Leonidas-from-XIV, line 84:
```suggestion
The error does not disappear if we remove the source for the bar lib:
```

test/blackbox-tests/test-cases/overlapping-deps.t
```diff
@@ -90,6 +90,16 @@ Strangely, the error disappears if we remove the source for the bar lib:
...
+  Error: Conflict between the following libraries:
+  - "some_package2" in _build/default/proj1
+  - "some_package2" in
+    $TESTCASE_ROOT/use/../external/_build/install/default/lib/some_package2
+    -> required by library "some_package1" in
+       $TESTCASE_ROOT/use/../external/_build/install/default/lib/some_package1
+  -> required by _build/default/proj2/.merlin-conf/lib-bar
+  -> required by _build/default/proj2/bar.cma
+  -> required by %{cma:proj2/bar} at command line:1
+  [1]
```
Leonidas-from-XIV, line 102: Interesting that this triggers the error now, this could be a regression for the release but it seems like the error is what should've happened here anyway.

test/blackbox-tests/test-cases/deprecated-library-name/features.t
```diff
@@ -96,8 +96,7 @@ that wasn't found:
   1 | (executable (name prog) (libraries a))
                                          ^
   Error: Library "a" not found.
-  -> required by _build/default/c/.prog.eobjs/byte/dune__exe__Prog.cmi
-  -> required by _build/default/c/.prog.eobjs/native/dune__exe__Prog.cmx
+  -> required by _build/default/c/.merlin-conf/exe-prog
```
Leonidas-from-XIV, line 99: My main concern is that this changes so many tests, but given the dependency that is getting reported is an internal one and stays an internal one I believe this is not an issue.

voodoos: It looks like the PR changes some evaluation order that makes Dune errors after following a different branch. Not sure what causes this however...

Leonidas-from-XIV: @rgrinberg Can you please weight in whether this is an issue? It doesn't seem to be wrong, but maybe the order is important in some way I am not aware of?

voodoos: I found the difference: before we only "peeked" (with [`Resolve.peek`](https://github.com/ocaml/dune/pull/10535/files#diff-d738a29d0c19436c2bb3dbf5358d8528183ba8558ed380f1bd6167a915e52c06L440-L444)) at the dependencies and now we ["read"](https://github.com/ocaml/dune/pull/10535/files#diff-d738a29d0c19436c2bb3dbf5358d8528183ba8558ed380f1bd6167a915e52c06R665-R666) them.

I will revert that change as is seems wrong to introduce dependencies here.

Leonidas-from-XIV (approved):

rgrinberg:
@nojb I think you've assigned yourself to this PR. Do you mind taking a final look, and if everything is OK, merging?

doc/changes/10535.md
```diff
@@ -0,0 +1,2 @@
+- Use Merlin new configuration directives `BH` and `SH` for hidden dependencies.
```
rgrinberg, line 1: I would mention that hidden dependencies correspond to the `-h` flag instead of internal merlin directives which most people are unaware of.

rgrinberg (approved):

doc/changes/10535.md
nojb: Indeed, it is always better to write Changes entries from the point of view of the user, not the developer. In this case, something like
```
- Make Merlin/OCaml-LSP aware of "hidden" dependencies used by
  `(implicit_transitive_deps false)` via the `-H` compiler flag.
```
would be more useful for the occasional reader.

nojb (approved):
LGTM

doc/changes/10535.md
nojb: I took the liberty of tweaking the Changes entry in this manner. Planning to merge once CI passes. Thanks!

voodoos:
Thank you @nojb  !

PR #10544, opened 2024-05-17
pkg: dedup package names in dependencies

### LINKED ISSUE #10542: pkg: duplicate package names in lockfile depends field of packages with conjunctions in their dependency constraints
Running `dune pkg lock` creates a lockfile whose `depends` field contains duplicate entries for packages with conjunctions in their dependency constraints, e.g.
```
depends: [
 "foo" {  >= "0.0.0" & < "1.0.0" }
]
```
...would translate into a lockfile with:
```
(depends foo foo)
```

Repro: https://github.com/ocaml/dune/pull/10543

gridbugs:
Fixes a bug where depending on a package with a constraint involving a conjunction would lead to the package appearing in the "depends" lockfile field multiple times.

Fixes https://github.com/ocaml/dune/issues/10542

### DIFF
--- a/src/dune_pkg/resolve_opam_formula.ml
+++ b/src/dune_pkg/resolve_opam_formula.ml
@@ -79,6 +79,19 @@ end
 type unsatisfied_formula =
   [ `Formula_could_not_be_satisfied of Unsatisfied_formula_hint.t list ]
 
+let dedup_package_names package_names =
+  let ret_reversed, _ =
+    List.fold_left
+      package_names
+      ~init:([], Package_name.Set.empty)
+      ~f:(fun (ret_reversed, seen) x ->
+        if Package_name.Set.mem seen x
+        then ret_reversed, seen
+        else x :: ret_reversed, Package_name.Set.add seen x)
+  in
+  List.rev ret_reversed
+;;
+
 let formula_to_package_names version_by_package_name opam_formula =
   let check_conjunction conjunction =
     Result.List.map conjunction ~f:(fun (opam_package_name, version_constraint_opt) ->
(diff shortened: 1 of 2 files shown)


src/dune_pkg/resolve_opam_formula.ml
```diff
@@ -79,6 +79,19 @@ end
...
+    List.fold_left
+      package_names
+      ~init:([], Package_name.Set.empty)
+      ~f:(fun (ret_reversed, seen) x ->
+        if Package_name.Set.mem seen x
+        then ret_reversed, seen
+        else x :: ret_reversed, Package_name.Set.add seen x)
+  in
+  List.rev ret_reversed
+;;
```
Leonidas-from-XIV, line 93: It could potentially make sense to have a

```
val uniq : equal:('a -> 'a -> bool) -> 'a list -> 'a t
```

style function in `Stdune.List`, but this is fine too.

Leonidas-from-XIV (approved):

src/dune_pkg/resolve_opam_formula.ml
Leonidas-from-XIV: Ah, I just noticed that `List.sort_uniq` already exists. It will obviously reshuffle the order, but might be ok for the usecase.

rgrinberg: Note that we aren't allowed to re-order the dependencies as specified by the user. The order has a semantic meaning.

rgrinberg (approved):

src/dune_pkg/resolve_opam_formula.ml
Leonidas-from-XIV: In the case of `libraries` yes, but do OPAM dependencies have a semantic ordering?

gridbugs: I think we should preserve the order of dependencies in lockfiles if for no other reason than to simplify the mental model for the correspondence between lockfiles and the opam files they are generated from. If some fields are copied but reordered then it's more work to understand how the opam file maps to the lockfile.

PR #10545, opened 2024-05-17
pkg: set with-dev-setup opam variable by default

gridbugs:


gridbugs:
Copy of https://github.com/ocaml/dune/pull/10540 for the toolchains branch, merging without review.

PR #10546, opened 2024-05-17
Implicit relock if the dependencies have changed

Leonidas-from-XIV:
If lockfiles exist *and* these are out of date, this PR runs the solver again to re-create the lock directory.

Obsoletes #10487 as this only does the automatic locking only if the user opted in by running `dune pkg lock` before.

### DIFF
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -1,5 +1,13 @@
 open Import
 
+(** ad-hoc feature flag that automatically re-creates lock files if the
+    contents of the dependency list in the dune-project goes out of date
+
+    Supposed to be enabled in the developer preview and disabled in the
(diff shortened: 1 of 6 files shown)


bin/build_cmd.ml
```diff
@@ -1,5 +1,7 @@
 open Import
 
+let use_autorelock = true
```
gridbugs, line 3: Can you add a comment about the plan for this variable, e.g. will it always be true in the developer preview and false otherwise?

gridbugs (approved):

bin/build_cmd.ml
Leonidas-from-XIV: Good point, done. Added a note that if this were to be set to true on main the whole switch should just be removed.

PR #10547, opened 2024-05-17
fix: don't take build lock for `coq top --no-build`

### LINKED ISSUE #7671: Running "dune coq top --no-build ..." tries to take the lock
## Expected Behavior

Running `dune coq top --no-build ...` should not fail if there is a concurrently running `dune build` command.

## Actual Behavior

You get: `Error: A running dune instance has locked the build directory`.

Is there anything we can do to fix this? Or does `dune coq top --no-build ...` possibly need to build stuff?

lzy0505:
This fixes #7671 where `dune coq top` tries to take the global build lock even if `--no-build` is enabled.

The problem was that `Common.Builder.term` had `allow_builds` set to `true` by default and didn't get updated when `--no-build` was specified. This led to the creation of an RPC server at initialization, which triggers the lock.

The fix was to update `allow_builds` to `false` with `Common.Builder.forbid_builds`.

### DIFF
--- a/bin/coq/coqtop.ml
+++ b/bin/coq/coqtop.ml
@@ -18,7 +18,7 @@ let man =
 let info = Cmd.info "top" ~doc ~man
 
 let term =
-  let+ builder = Common.Builder.term
+  let+ default_builder = Common.Builder.term
   and+ context =
     let doc = "Run the Coq toplevel in this build context." in
     Common.context_arg ~doc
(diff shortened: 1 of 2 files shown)


rlepigre:
> @rlepigre could you test this?

@Alizter I gave this PR a quick try, and it seems to do the trick. Thanks for the fix @lzy0505!

Alizter (approved):

Alizter:
@lzy0505 Could you add a changelog entry? Have a look at doc/changes to see how it is done. Once that is done, we can probably merge.

lzy0505:
@Alizter Done.

Alizter:
Thanks this looks good. @rgrinberg @emillon this is ready for merging.

PR #10548, opened 2024-05-18
scheduler: simplify build status handling

pmwhite:
Based off https://github.com/ocaml/dune/pull/10518

This is an alternative to https://github.com/ocaml/dune/pull/10519, based on some discussion in that PR. This one keeps the build status intact, but merges `Standing_by` and `Restarting_build` into a single case. I recommend reading the diff commit-by-commit.

### DIFF
--- a/src/dune_engine/scheduler.ml
+++ b/src/dune_engine/scheduler.ml
@@ -1015,9 +1015,11 @@ end = struct
         match t.status with
         | Not_building -> []
         | Building cancellation ->
-          t.handler t.config Build_interrupted;
-          t.status <- Not_building;
-          Fiber.Cancel.fire' cancellation
+          if Fiber.Cancel.fired cancellation
+          then []
+          else (
+            t.handler t.config Build_interrupted;
+            Fiber.Cancel.fire' cancellation)
       in
       let fills = Trigger.trigger t.build_inputs_changed @ fills in
       match Nonempty_list.of_list fills with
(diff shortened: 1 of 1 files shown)


src/dune_engine/scheduler.ml
```diff
@@ -1015,9 +1015,11 @@ end = struct
         match t.status with
         | Not_building -> []
         | Building cancellation ->
-          t.handler t.config Build_interrupted;
-          t.status <- Not_building;
-          Fiber.Cancel.fire' cancellation
+          if Fiber.Cancel.fired cancellation
+          then []
+          else (
+            t.handler t.config Build_interrupted;
```
rgrinberg, line 1027: Before, we'd call this callback unconditionally. Is there a change of behavior here?

pmwhite: Before, it was impossible to hit this case twice in the same build because of the `t.status <- Not_building` line above. But without that line, we can hit this case multiple times, and we only want to sent the `Build_interrupted` notification on the first interruption (instead of sending it on every subsequent interruption).

pmwhite:
I just ran this PR against our internal test suite and hit the following bug: In passive watch mode, if we cancel a build at about the same time as a file changes, then our explicit cancellation will be ignored because it look like the build was cancelled in response to a memo invalidation.

I'm going to close this PR for now because I'm no longer sure this is getting me any closer to my goal.

PR #10549, opened 2024-05-19
fix(lib-collision): tolerate `public_name` collisions in different contexts

anmonteiro:
- fixes the test added in https://github.com/ocaml/dune/pull/10471
- we use the same strategy for public libraries that we use elsewhere in `scope.ml`: we store and `enabled` `Toggle.t Memo.t` that we evaluate when resolving a public library by its name.

### DIFF
--- a/src/dune_rules/scope.ml
+++ b/src/dune_rules/scope.ml
@@ -206,25 +206,30 @@ module DB = struct
     | Project of
         { project : Dune_project.t
         ; lib_id : Lib_id.Local.t
+        ; enabled : Toggle.t Memo.t
         ; loc : Loc.t
         }
     | Name of (Loc.t * Lib_name.t)
 
   (* Create a database from the public libraries defined in the stanzas *)
   let public_libs =
-    let public_loc_of_redirect_to = function
-      | Project { loc; _ } | Name (loc, _) -> loc
-    in
     let resolve_redirect_to t rt =
       match rt with
-      | Project { project; lib_id; _ } ->
-        let scope = find_by_project (Fdecl.get t) project in
-        Lib.DB.Resolve_result.redirect_by_id scope.db (Local lib_id)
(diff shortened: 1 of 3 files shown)


jchavarri (approved):
Thanks!

src/dune_rules/scope.ml
```diff
@@ -473,15 +460,22 @@ module DB = struct
              | Some lib ->
                let name = Package.name pkg in
                (name, Lib_entry.Library (Lib.Local.of_lib_exn lib)) :: acc)
-          | Library.T { visibility = Public pub; _ } ->
-            let+ lib = Lib.DB.find public_libs (Public_lib.name pub) in
+          | Library.T { visibility = Public pub; enabled_if; _ } ->
```
rgrinberg, line 463: Why don't private libraries have an enabled_if check?

src/dune_rules/scope.ml
```diff
@@ -473,15 +460,22 @@ module DB = struct
...
+            let* lib = Lib.DB.find public_libs (Public_lib.name pub) in
```
rgrinberg, line 464: You should use the id rather than the name for this check.

src/dune_rules/scope.ml
anmonteiro: that's a good point. I'll make these changes shortly.

src/dune_rules/scope.ml
anmonteiro: I tried `find_by_lib_id` but I got a crash about adding a duplicate name in https://github.com/ocaml/dune/blob/main/src/dune_rules/install_rules.ml#L697

anmonteiro: I'm not sure what that means yet.

anmonteiro: Ah, it's likely because it's finding the right lib now.

PR #10550, opened 2024-05-20
feat: allow sandboxing for compilation rules

anmonteiro:


rgrinberg:
LGTM. We can disable the sandboxing benchmarks for now

PR #10555, opened 2024-05-22
doc: remove outdated info about variants

emillon:
Some of our docs about virtual libraries mention "variants", which is a feature that doesn't exist anymore (since 2020).

This removes the history part and uses consistent naming.

### DIFF
--- a/doc/howto/index.rst
+++ b/doc/howto/index.rst
@@ -16,7 +16,7 @@ These guides will help you use Dune's features in your project.
    ../instrumentation
    ../jsoo
    ../melange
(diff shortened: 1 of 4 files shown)


rgrinberg:
Given the fact that people often use this terminology in the wild and not only in the context of dune, it makes sense to me mention in the manual as something deprecated and historical. Without the current mention, users wouldn't have a way t resolve their confusion.

emillon:
Good point. What do you think about this new text?

rgrinberg (approved):

PR #10556, opened 2024-05-22
pkg: use our unpack code for rsync urls

emillon:
This removes an important use of `fetch_others`.

(the test was printing the expected hash instead of the computed one)


rgrinberg (approved):
LGTM.

Note that we should probably explicitly error on URL's of the form `rsync://`. We don't yet properly support using rsync.

emillon:
Thanks. I added some checks for this and `unpack`.

PR #9904, opened 2024-02-02
doc: split files doc into smaller documents

emillon:
First, apologies for the large PR, but actually it is mostly moving things around.

The "previous" situation is that we have a single `dune-files.rst` which contains documentation for all of our files, also including `dune-project`, `dune-workspace`, and `config`.

To get it organized, some pieces are extracted into `doc/stanzas/*` but that's just textual inclusion, it still builds a single page. Also, we rely on references (`:ref:`) to do deep links (for example, to link to the documentation of a stanza or a field).

What this PR does is switch to a hierarchical structure, where individual stanzas are in their own document, grouped by the file type they can be found it (concretely, there is one directory for `dune` stanzas, one directory for `dune-project` stanzas, etc). And instead of relying on references, we link to documents (`:doc:`) directly.

Here's how the PR is structured in terms of commits:

- for each file type (config, dune, dune-project, dune-workspace):
  - move the section of `dune-files.rst` to `/reference/files/X/index.rst`
  - split the subsections into `/reference/files/X/*.rst`
  - remove the references and turn them into `:doc:` links.
- remove `dune-files.rst`

We can discuss the approach, and if there are concrete remarks to do I'm happy to split the PR in smaller steps, its structure should allow it easily.

Thanks

PR #10559, opened 2024-05-22
doc: split docs for builtin aliases

emillon:
This is the same kind of work as #9904, but for aliases: we give each builtin alias its own document, and add cross-references to these new pages.


doc/reference/aliases/default.rst
```diff
@@ -0,0 +1,26 @@
+@default
+========
+
+This alias corresponds to the default argument for ``dune build``: ``dune
+build`` is equivalent to ``dune build @@default``. Similarly, ``dune build
```
Leonidas-from-XIV, line 5: I think it's surprising that the documentation about `@default` suddenly talks about `@@default`.

doc/reference/aliases/doc.rst
```diff
@@ -0,0 +1,4 @@
+@doc
+====
+
+This alias builds documentation for public libraries as HTML pages.
```
Leonidas-from-XIV, line 4: I think it would be useful to link to the "Generating documentation" document here.

doc/reference/aliases/install.rst
```diff
@@ -0,0 +1,6 @@
+@install
+========
+
+This alias depends on the ``*.install`` files used by the :doc:`opam
```
Leonidas-from-XIV, line 4: While "depends" is correct, I think a more understandable high-level description would be saying that this will create the install files which in turn...

doc/reference/aliases/runtest.rst
```diff
@@ -0,0 +1,7 @@
+@runtest
+========
+
+Actions that run tests are attached to this alias. For example this convention
+is used by the ``(test)`` stanza.
+
+``dune runtest`` is a shortcut for ``dune build @runtest``.
```
Leonidas-from-XIV, line 7: Likewise, linking to "Writing and running tests" could be a good link to add.

Leonidas-from-XIV (approved):
I think this is a good improvement and linking the mentions of the aliases to the documentation makes it much easier to traverse the document to find information about them. I've added a few notes how I think it could be make it a bit more user-friendly still.

emillon:
Thanks. I added your suggestions.

PR #10561, opened 2024-05-23
pkg: disable hg/darcs fetch code

emillon:
This marks hg and darcs backends as unsupported, and removes the code path that uses opam libs to download sources.


rgrinberg (approved):

emillon:
(FTR, there are no hg or darcs packages on opam-repository)

PR #10567, opened 2024-05-24
Read toolchain source url from opam repo

gridbugs:
Rather than hardcoding the source urls and checksums of ocaml compilers, this information is taken from the opam repo. Currently the only supported compiler package is ocaml-base-compiler, but this change attempts to make it simpler to introduce support for the ocaml-variants package in the near future.

We used to pass around toolchain versions as a proxy for toolchains to keep track of which toolchain to use, mostly in pkg_rules.ml. I've replaced this with what I'm calling a "compiler package" which contains a bit more information than just the version, mostly so that when we add support for ocaml-variants, that the compiler options can be added to this type.

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -198,11 +198,11 @@ module Value_list_env = struct
       Some (Value.Dir path :: paths))
   ;;
 
-  (* Prepends the bin dir of a given toolchain version to the PATH
+  (* Prepends the bin dir of a given toolchain compiler to the PATH
      variable in the environment. *)
-  let add_toolchain_bin_dir_to_path t toolchain_version =
+  let add_toolchain_bin_dir_to_path t compiler =
     let toolchain_bin_dir =
-      Toolchain.Version.bin_dir toolchain_version |> Path.outside_build_dir
+      Toolchain.Compiler.bin_dir compiler |> Path.outside_build_dir
     in
     add_path t Env_path.var toolchain_bin_dir
   ;;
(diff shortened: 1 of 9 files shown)


src/dune_rules/pkg_rules.ml
```diff
@@ -1603,84 +1615,88 @@ let source_rules (pkg : Pkg.t) =
...
+         depend on the compiler but a compiler is still needed to
+         build the project.
+
+         In this case, default to the latest version of the base
+         compiler. *)
+      let latest_compiler_package =
+        Toolchain.Available_compiler_packages.latest_base_compiler_package
+          db.available_compiler_packages
+      in
+      Memo.return (Toolchain latest_compiler_package)
```
Leonidas-from-XIV, line 1642: Are we sure this is the right thing to do? Maybe someone uses dune to not build ocaml code? Like, for example they only use Melange.

gridbugs: In that case wouldn't they still need the ocaml compiler to build melange?

That said, since we generate a lockdir when running `dune build` now, this case should never get hit as there will always be a lockdir present when building the project. I think I will remove this logic since I don't think it will make it upstream.

gridbugs: We do need to make sure that we don't break the hello-world scenario in the interviews. The project initially doesn't have any dependencies, right, and in that situation we still need to be able to compile ocaml code I think. Removing this code won't break it (I'll double check) because we hard code a dependency on ocaml-base-compiler, but that's a hack that I'll remove soon, at which point we will need some other way of either making sure that all projects implicitly depend on the compiler, or making sure that the compiler is available even if it's not an explicit dependency.

gridbugs: Removing this also meant that I could remove a bunch of code related to finding the latest version of the compiler.

gridbugs:
I just renamed `compiler_package` to `compiler` in more places. Let me know if there is anything more you'd like me to change in this PR.

Leonidas-from-XIV (approved):

PR #10574, opened 2024-05-24
chore(pkg): remove pointless channel when writing lock

rgrinberg:
No need to allocate a huge buffer to write an integer. Also, we already have a pid writing function, so we can just reuse it.


rgrinberg:
ping @Leonidas-from-XIV

Leonidas-from-XIV (approved):

PR #10575, opened 2024-05-24
fix: truncate lock file

rgrinberg:
Truncate it after successful locks and before an unlock.

This makes sure that we aren't left with garbage from previous runs.


rgrinberg:
This has been reviewed internally

PR #10579, opened 2024-05-24
Reproduction case for `dune build @fmt` erroring when ocamlformat is not installed

shonfeder:
Shows the current behavior of `dune build @fmt` exiting with an error when ocamlformat is not installed.


test/blackbox-tests/test-cases/formatting/ocamlformat-not-installed.t
```diff
@@ -0,0 +1,20 @@
...
+
+We neeed an ocaml file and a dune rule to trigger the logic that would hit ocamlformat
+
+  $ touch dune foo.ml
+
+TODO: Check if ocamlformat is installed and figure out how to remove or mask it if so
+
+  $ ocamlformat
+
+  $ which ocamlformat
```
shonfeder, line 18: I haven't figure out a nice clean way to ensure that ocamlformat is not installed yet. I'll keep thinking on it, but any suggestions would be welcome.

Leonidas-from-XIV: Maybe you could try cleaning out `PATH` before calling Dune? It's admittedly much easier to add things in the `PATH` than to remove them.

PR #10589, opened 2024-05-27
merlin: do not break lines in dump-config

emillon:
`dune ocaml merlin dump-config` is meant for testing and debugging.

Breaking horizontal boxes is not useful here and it makes postprocessing more difficult.

Setting a large margin will make some boxes render as a single line.

This also helps with reproducibility, since sometimes long paths could break differently.


emillon:
The context is that some melange tests have been failing locally and that's due to a difference in path length.

This can be reviewed using `git diff --word-diff`.

voodoos (approved):
Not sure about the macos CI failure, but the changes look good to me.

PR #10590, opened 2024-05-27
fix: typo: exist_if -> exists_if

emillon:
The ocamlfind field is called `exists_if`, so let's use that.


Leonidas-from-XIV (approved):
A quick check in `ocamlfind` reveals that that's indeed true. Good catch!

Issue #10593, opened 2024-05-28
pkg: dune ignores the `avoid-version` flag in opam packages


gridbugs:
This can lead to problems such as https://github.com/ocaml/dune/issues/10592 where the solver chooses development versions of packages.

There is an issue to address this upstream in opam-0install-solver (which dune vendors): https://github.com/ocaml-opam/opam-0install-solver/issues/23

I'm making this issue to document how this affects dune package management, though when we fix this we should upstream the fix and address the linked issue as well.

Reproduced in https://github.com/ocaml/dune/pull/10594 and https://github.com/ocaml/dune/pull/10595

gridbugs:
Correct!

PR #10594, opened 2024-05-28
pkg: test exercising package disjunctions

gridbugs:
Reproduces https://github.com/ocaml/dune/issues/10593

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/opam-solver-or.t
+++ b/test/blackbox-tests/test-cases/pkg/opam-solver-or.t
@@ -3,10 +3,10 @@ Demonstrate the generation of the lock directory in the presence of "|"
   $ . ./helpers.sh
   $ mkrepo
 
-  $ mkpkg a1 <<EOF
+  $ mkpkg a1 0.0.1 <<EOF
   > conflicts: [ "a2" ]
   > EOF
-  $ mkpkg a2 <<EOF
+  $ mkpkg a2 0.0.1 <<EOF
   > conflicts: [ "a1" ]
   > EOF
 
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/pkg/disjunction.t
```diff
@@ -0,0 +1,51 @@
...
+Release a new version of the second package in the disjunction to demonstrate
+that relative version numbers don't affect the choice.
+  $ mkpkg b 0.2
+  $ solve dep-on-disjunction
+  Solution for dune.lock:
+  - a.0.1
+  - dep-on-disjunction.0.1
+
+New version of dep-on-disjunction specifying version numers of deps. Note that
+only "b" exists with the specified version. If the solver chooses the new
```
Leonidas-from-XIV, line 30: Could you also add `c.0.2` or make it depend on `c.0.1`? Because otherwise `b.0.2` is the only solution and that just tests that the solver can find a solution.

test/blackbox-tests/test-cases/pkg/disjunction.t
```diff
@@ -0,0 +1,51 @@
...
+  > EOF
+  $ solve dep-on-disjunction
+  Solution for dune.lock:
+  - b.0.2
+  - dep-on-disjunction.0.2
+
+Mark "b.0.2" as avoid-version so the solver will try to find a solution which doesn't include it.
+  $ mkpkg b 0.2 << EOF
+  > flags: [avoid-version]
+  > EOF
```
Leonidas-from-XIV, line 43: To me this seems correct, as `b.0.2` is the only version that can be picked, even if it is to be avoided otherwise.

In opam-monorepo we reshuffle the candidates to demote avoid-version packages: https://github.com/tarides/opam-monorepo/blob/141d351b79715c99d66bfd898a2a64115a2e8769/lib/opam_solve.ml#L179-L189 

This is of course only a partial solution because if all versions are demoted, but the package is first in the disjunction, the solver will still pick it over the next one in the disjunction (even if these packages are not marked as `avoid-version`).

gridbugs: That's also the approach I took with this fix https://github.com/ocaml/dune/pull/10596

rgrinberg:
The tests look fine but I remember adding a bunch of tests along these lines already. Did you make sure there's no duplication?

test/blackbox-tests/test-cases/pkg/disjunction.t
gridbugs: I don't think that's true. For example the solver is free to choose `deps-on-disjunction.0.1` rather than `deps-on-disjunction.0.2`.

gridbugs:
> The tests look fine but I remember adding a bunch of tests along these lines already. Did you make sure there's no duplication?

Good point, I've changed this PR to update the existing disjunction test with new cases.

rgrinberg (approved):

PR #10595, opened 2024-05-28
pkg: test of how the solver chooses compiler packages

gridbugs:
Demonstrates several problems where dune chooses unstable compiler packages by default due to the fact that it ignores the avoid-version flag.


Leonidas-from-XIV (approved):
Yep, this exactly mirrors my understanding of why we pick alpha compilers.

Issue #10592, opened 2024-05-28
pkg: the solver prefers unreleased version of ocaml-variants over ocaml-base-compiler


gridbugs:
Sometimes versions of the compiler that are not officially released are released on opam. At the time of writing, the latest officially released version of ocaml is 5.2.0, however the opam package `ocaml.5.3.0` exists. It depends on the disjunction of `ocaml-base-compiler.5.3.0` (which does not exist), and `ocaml-variants.5.3.0+trunk` which does exist but is marked `avoid-version` (dune currently ignores this flag due to https://github.com/ocaml-opam/opam-0install-solver/issues/23).

The consequence of this is that if a project depends on `ocaml`, dune will find a solution which includes `ocaml-variants`, as this is the only way that the latest version of the `ocaml` can also be chosen.

The output from the solver at the time of writing is:
```
$ dune pkg lock
Solving for Build PlanPackage "ocaml-variants" has source archive which lacks a checksum.
The source archive will be downloaded from:
https://github.com/ocaml/ocaml/archive/trunk.tar.gz
Dune will compute its own checksum for this source archive.
Solution for dune.lock:
- ocaml.5.3.0
- ocaml-config.3
- ocaml-variants.5.3.0+trunk
```

Repro: https://github.com/ocaml/dune/pull/10595

PR #10596, opened 2024-05-28
pkg: avoid unstable compiler packages

gridbugs:
This introduces two workarounds to problems caused by the fact that opam-0install-solver - the library underlying dune's dependency solver - does not understand the avoid-version flag which is set on opam packages to indicate that the solver should choose an alternative package if it's able to do so.

The first problem happens when ocaml-base-complier has an alpha, beta, or rc package available with a higher version number than the latest stable release of ocaml-base-compiler. In this case, dune would choose the higher versioned (unstable) package rather than the latest stable version.

The workaround is to reorder the list of candidate versions for each package such that packages that have the avoid-version flag set appear later than those that do not have this flag set. (Note that this is done per-package name, not between different package names.)  This solution will fix the problem for all packages - not just ocaml-base-compiler.

The second problem happens when a version of the meta package named "ocaml" is available with a higher version than the latest version of ocaml-base-compiler, and there is a development version of ocaml-variants available which corresponds to the version of the latest "ocaml" package. In this case, dune would try to choose the latest version of "ocaml" and end up depending on the development version of ocaml-variants rather than ocaml-base-compiler. All versions of ocaml-variants have the avoid-version flag set.

The workaround here is to determine which version of ocaml-base-compiler is the latest stable release (the version with the highest version number that does not have avoid-version set). Then, when considering the version candidates for the "ocaml" meta package, reorder the list of candidates such that any versions that are higher than the latest release of ocaml-base-compiler are moved to the end of the list.

It's worth noting that in both cases, the solution does not restrict which package solutions are possible, it just changes the way that the solver prioritizes packages relative to one another; if a user specifies a dependency on a development version of a package, or the ocaml-variants package, it will still be included in the solution.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -65,7 +65,7 @@ module Context_for_dune = struct
          packages for which we've printed a warning. *)
       available_cache : (OpamPackage.t, bool) Table.t
     ; constraints : OpamTypes.filtered_formula Package_name.Map.t
-    ; latest_released_base_compiler_version : Package_version.t option
+    ; latest_released_base_compiler : OpamFile.OPAM.t option
     }
 
   let create
@@ -76,7 +76,7 @@ module Context_for_dune = struct
     ~version_preference
     ~stats_updater
     ~constraints
-    ~latest_released_base_compiler_version
+    ~latest_released_base_compiler
     =
     let dune_version =
       let major, minor = Dune_lang.Stanza.latest_version in
@@ -108,7 +108,7 @@ module Context_for_dune = struct
     ; candidates_cache
     ; available_cache
     ; constraints
-    ; latest_released_base_compiler_version
+    ; latest_released_base_compiler
     }
   ;;
 
@@ -219,24 +219,33 @@ module Context_for_dune = struct
      released version of ocaml-base-compiler is not known, return the
      list unchanged. *)
   let move_packages_newer_than_latest_released_base_compiler_to_end t packages =
-    match t.latest_released_base_compiler_version with
-    | Some latest_released_base_compiler_version ->
-      let latest_released_base_compiler_version =
-        Package_version.to_opam_package_version latest_released_base_compiler_version
-      in
+    match t.latest_released_base_compiler with
+    | Some latest_released_base_compiler ->
       let earlier_or_equal, later =
         List.partition packages ~f:(fun package ->
-          let package_version =
-            OpamFile.OPAM.version (Resolved_package.opam_file package)
-          in
-          match
-            opam_version_compare t package_version latest_released_base_compiler_version
-          with
+          let opam_file = Resolved_package.opam_file package in
+          match opam_version_compare t opam_file latest_released_base_compiler with
           | Lt -> false
           | Eq | Gt -> true)
       in
       earlier_or_equal @ later
     | None -> packages
+  ;;
+
+  (* It's common for the latest version of
+     the ocaml opam package to be greater than the latest released
+     ocaml-base-compiler, in which case the ocaml package will depend
+     development releases of the underlying compiler, such as +trunk releases
+     of ocaml-variants.
+
+     Development releases of packages are marked with the avoid-version flag,
+     but opam-0install-solver does not implement this flag.
+
+     Consequently, for when the package is called "ocaml" and the version to
+     to be picked a hack is needed. *)
+  let requires_latest_version_hack t name =
+    Version_preference.equal t.version_preference Version_preference.Newest
+    && Package_name.(equal (of_opam_package_name name) (of_string "ocaml"))
   ;;
 
   let repo_candidate t name =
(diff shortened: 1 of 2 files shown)


gridbugs:
Fixes https://github.com/ocaml/dune/issues/10592
Related to https://github.com/ocaml/dune/issues/10593, though doesn't fix it in general.

src/dune_pkg/opam_solver.ml
```diff
@@ -177,19 +183,82 @@ module Context_for_dune = struct
...
+           sorting for clarity. *)
+        |> List.sort ~compare:(fun p1 p2 ->
+          opam_version_compare
+            t
+            (Resolved_package.opam_file p1 |> OpamFile.OPAM.version)
+            (Resolved_package.opam_file p2 |> OpamFile.OPAM.version))
+      in
+      (if Package_name.(equal (of_opam_package_name name) (of_string "ocaml"))
+       then
+         (* This is a hack to avoid versions of the the ocaml package
```
rgrinberg, line 244: This hack is only needed when preferring upper bounds, right? In which case, it should be enabled when preferring upper bounds.

src/dune_pkg/opam_solver.ml
```diff
@@ -177,19 +183,82 @@ module Context_for_dune = struct
...
+            which are later than the latest release of
+            ocaml-base-compiler. It's common for the latest version of
+            the ocaml opam package to be greater than the latest
+            released ocaml-base-compiler, in which case the ocaml
+            package will depend development releases of the underlying
+            compiler, such as +trunk releases of
+            ocaml-variants. Development releases of packages are
```
rgrinberg, line 251: Isn't this PR implementing support for the avoid-version flag? That is, it's taking this flag into account when returning the list of available versions.

rgrinberg (changes requested):
It would be good to split the compiler specific hack with the support for `avoid-version`.

Also, this implementation with a bunch of post processing steps is rather confusing. Is it possible to just put all of this logic into to the comparison function we pass to the sort? This a much simpler way of thinking about what version we're going to get.

src/dune_pkg/opam_solver.ml
gridbugs: No not completely. After this change the solver will respect `avoid-version` between different versions a given package (e.g. avoiding beta releases of the compiler) however it will not respect `avoid-version` between different packages in general. If `foo` depends on the disjunction `bar | baz` and `bar` is `avoid-version` while `baz` is not, `bar` may still be chosen to satisfy the constraint.

gridbugs:
> It would be good to split the compiler specific hack with the support for `avoid-version`.

I'll move the part of this change that reorders package versions by `avoid-version` into its own PR and update this one to just include the compiler-specific hack.

src/dune_pkg/opam_solver.ml
gridbugs: Good point. Will update.

gridbugs:
Here is a PR that reorders package versions by `avoid-version` but doesn't add the compiler-specific hack: https://github.com/ocaml/dune/pull/10642

src/dune_pkg/opam_solver.ml
Leonidas-from-XIV: I've moved the condition into a predicate function along with @rgrinberg's suggestion.

src/dune_pkg/opam_solver.ml
```diff
@@ -195,19 +198,85 @@ module Context_for_dune = struct
...
   ;;
 
+  (* Move all packages with the avoid-version flag beyond any package
+     without that flag, otherwise preserving the order. This is needed
+     so that development versions of packages (such as alpha releases
+     of the compiler) are not favoured over older, stable releases of
+     the package. This is a hack to work around the fact that
+     opam-0install-solver does not implement the avoid-versions
+     flag. *)
+  let move_avoid_versions_packages_to_end packages =
```
rgrinberg, line 208: I don't think this is needed? This is already handled by the new comparison function.

rgrinberg (changes requested):

src/dune_pkg/opam_solver.ml
Leonidas-from-XIV: It is still needed, as we need to make sure the most recent version of the `ocaml` package that depends on a non-`avoid-version` compiler package is located in front of versions of `ocaml` that have only packages that are marked `avoid-version`.

E.g. the current state of the [`ocaml.5.3.0`](https://github.com/ocaml/opam-repository/blob/0b240d2960133fd3d8aa8f008d7aa79534caa3b9/packages/ocaml/ocaml.5.3.0/opam) depends on `ocaml-base-compiler.5.3.0` (doesn't exist) or `ocaml-variants {>= "5.3.0~" & < "5.3.1~"}` which can be fulfilled by [`ocaml-variants.5.3.0+trunk`](https://github.com/ocaml/opam-repository/blob/0b240d2960133fd3d8aa8f008d7aa79534caa3b9/packages/ocaml-variants/ocaml-variants.5.3.0%2Btrunk/opam).

Thus without this function the solver would attempt to pick `ocaml.5.3.0`; see that there are no candidates for `ocaml-base-compiler` and skip; see that there is one candidate for `ocaml-variants` (we have sorted it to the back due to `avoid-version` but the candidates list only contains one entry anyway) and pick that one.

This is a consistent pattern in opam-repository (that there is a new `ocaml` package before the final release of `ocaml-base-compiler` exists) that in many cases Dune would always pick `ocaml-variants.<latest-development-release>+trunk` that this hack might be worth it.

gridbugs: I'm going to break the logic for restricting the ocaml version to the latest stable ocaml-base-compiler version into its own PR for clarity, then close this PR.

gridbugs:
I've made a PR that splits out the workaround for the issue where the solver always chooses unstable versions of the compiler, moving the logic into the version comparison function as @rgrinberg suggested. https://github.com/ocaml/dune/pull/10668

PR #3721, opened 2020-08-18
[coq] Add .aux & .glob files as targets

rgrinberg:
Fix #3437

@gares @ejgallego should `.aux` files be installed as well?

PR #10602, opened 2024-05-30
[coq] [fix] Install .glob files

ejgallego:
This seems like an omission from the beginning, I'm surprised this wasn't a problem for anyone until now.

### DIFF
--- a/doc/changes/10602.md
+++ b/doc/changes/10602.md
@@ -1,0 +1,1 @@
+- install `.glob` files for Coq theories too (#10602, @ejgallego)
--- a/src/dune_rules/coq/coq_module.ml
+++ b/src/dune_rules/coq/coq_module.ml
@@ -76,6 +76,17 @@ let glob_file x ~obj_dir =
   Path.Build.relative vo_dir (x.name ^ ".glob")
 ;;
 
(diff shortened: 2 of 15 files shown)


Alizter:
These were added as targets back here:
- https://github.com/ocaml/dune/pull/3721

ejgallego:
Thanks for finding out about that PR @Alizter , indeed it was mostly an accident as it can be seen from the discussion; the plan was to install `.glob` files when we added coqdoc support, however we did forget.

JasonGross:
> I'm surprised this wasn't a problem for anyone until now.

I guess nobody had tried to use coqbot to minimize a development that both (a) relied on dune, and (b) did not qualify project-local imports.  (The minimizer is fairly robust to missing .glob files, in that it will attempt to reconstruct the .glob file via coqc, even for installed files, using the same flags that are used to compile the development being minimized.  Thus only projects that rely on binding their own files with `-R` instead of `-Q` fail to have .glob files remade.)

ejgallego:
Indeed @JasonGross , actually I noted because I did implement jump to definition in coq-lsp and I was like "where are the glob files) !

PR #10603, opened 2024-05-31
rpc: is building aliases possible?

jchavarri:
I am trying to build an alias through RPC but I don't seem to be able to. I include a small repro as a test in the PR.


rgrinberg:
Yes, building through RPC has limitations.

Alizter:
To clarify the situation, it is possible to build an alias through RPC, but the syntax is different. You will need to do `dune rpc build '(alias_rec foo)'` instead. However `dune build`, when connecting to an RPC server, will work fine.

PR #10604, opened 2024-06-03
ci: only the static binary is needed.

moyodiallo:
It could be confusing also because the non-static binary doesn't works in general and  only works on a specific environment (nix config).

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -18,9 +18,6 @@ jobs:
             installable: .#
           - os: ubuntu-22.04
             name: linux-intel
-            installable: .#
-          - os: ubuntu-22.04
-            name: linux-intel-static
             installable: .#dune-static
     runs-on: ${{ matrix.os }}
     steps:


emillon (approved):

gridbugs:
I'm trying to better understand the problem with the gnu target. Does anyone have more info about why the binary only runs on nix (just confirming that you mean nixos)?

anmonteiro:
Likely he means nix, not just nixOS. The reason being that Nix has its own libraries in the nix store (usually at `/nix/store`, in every Nix installation), so dynamically linked libraries will point to it

emillon:
Yes, both the dynamic linker entry (`ld-linux.so` and the dynamic dependencies will use nix conventions, which isn't portable to all linux systems.

So this leaves 2 approaches:
- build a dynamic binary on a old glibc system so that it can be dynamically loaded on most systems
- build a static binary

We already have a set up to do the latter and it'll be faster due to better caching, so we use a static binary (since we don't have external C dependencies on linux, the drawbacks of static binaries, like size, don't apply here).

gridbugs:
> build a dynamic binary on a old glibc system so that it can be dynamically loaded on most systems

Why an "old" glibc system? Is there a problem with the binaries resulting from building dune on a modern glibc system?

emillon:
yes, the dynamic symbols have a version associated with them, and this means that a binary built on a system with a newer glibc might not run on one with an older glibc.

PR #10614, opened 2024-06-05
Support ocaml-variants with toolchains

gridbugs:
This changes the toolchains mechanism to accept the ocaml-variants package, and allow the build commands of a package to check whether certain packages are installed which allows the ocaml-variants package to be configured by the installation of its depopts.

### DIFF
--- a/src/dune_pkg/toolchain.ml
+++ b/src/dune_pkg/toolchain.ml
@@ -16,18 +16,48 @@ module Make = struct
   ;;
 end
 
-module Names = struct
-  let ocaml_base_compiler = Package_name.of_string "ocaml-base-compiler"
-end
-
 module Compiler = struct
   (* Value of the PATH variable *)
   let env_path = Env_path.path Env.initial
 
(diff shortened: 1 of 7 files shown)


src/dune_pkg/toolchain.ml
```diff
@@ -16,18 +16,48 @@ module Make = struct
...
+       their dependencies. This means that the list of dependencies must be
+       present in the name of the directory where the compiler gets
+       installed.
+
+       An example identifier string for ocaml-variants with multiple
+       dependencies is:
+       ocaml-variants.5.2.0+options.with_deps.ocaml-option-flambda,ocaml-option-no-flat-float-array *)
+    let identifier_string t =
+      let deps_suffix =
+        if List.is_empty t.deps
```
Leonidas-from-XIV, line 45: Probably nicer as a pattern match on `[]` than `if/else`

Leonidas-from-XIV (approved):
Looks good

PR #10620, opened 2024-06-07
Allow toolchains to be disabled with flag

gridbugs:
Also fixes a bug that was introduced with the initial toolchains implementation that leads to a memo cycle when resolving the ocaml-variants package. Breaking the memo cycle also let us remove the code that tracks the compiler dependency of individual packages which was adimitedly a hack to work around the same circular dependency that caused the memo cycle. Removing this code should make it less contraversial to merge the toolchains branch upstream.

### DIFF
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -1,13 +1,5 @@
 open Import
 
-(** ad-hoc feature flag that automatically re-creates lock files if the
-    contents of the dependency list in the dune-project goes out of date
-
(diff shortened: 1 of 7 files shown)


gridbugs:
I've tested this by building bonsai and making sure that it fails due to the compiler's relocation issue, and that forcing the ocaml-system compiler solves the problem.

Leonidas-from-XIV (approved):

PR #10622, opened 2024-06-07
melange: the compiler does not support -bin-annot-occurrences

voodoos:
The Melange compiler does not accept the `-bin-annot-occurrences` flag.

The test relying on OCaml's version is therefore incorrect when building with melange compiler.
This PR adds a special case to prevent use of this flag when building with melange.

### DIFF
--- a/src/dune_rules/module_compilation.ml
+++ b/src/dune_rules/module_compilation.ml
@@ -183,23 +183,20 @@ let build_cm
        (Action_builder.map module_deps ~f:(other_cm_files ~opaque ~cm_kind ~obj_dir))
    in
    let other_targets, cmt_args =
-     match cm_kind with
-     | Ocaml Cmx -> other_targets, Command.Args.empty
-     | Ocaml (Cmi | Cmo) | Melange (Cmi | Cmj) ->
-       if Compilation_context.bin_annot cctx
-       then (
-         let fn =
-           Option.value_exn (Obj_dir.Module.cmt_file obj_dir m ~cm_kind ~ml_kind)
-         in
-         let annots =
-           [ "-bin-annot" ]
-           @
-           if Version.supports_bin_annot_occurrences ocaml.version
-           then [ "-bin-annot-occurrences" ]
-           else []
-         in
-         fn :: other_targets, As annots)
-       else other_targets, Command.Args.empty
+     match Compilation_context.bin_annot cctx, cm_kind with
+     | false, _ | true, Ocaml Cmx -> other_targets, Command.Args.empty
+     | true, Ocaml (Cmi | Cmo) ->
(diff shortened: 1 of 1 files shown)


emillon:
cc @anmonteiro - do we want this, or fix this on the melange side?
(@gridbugs we might need to backport this to 3.16 if we go with the patch route)

anmonteiro:
We’ll have to fix it in both sides I suppose. Where can I read more about this flag?

src/dune_rules/module_compilation.ml
```diff
@@ -183,23 +183,20 @@ let build_cm
...
-         in
-         fn :: other_targets, As annots)
-       else other_targets, Command.Args.empty
+     match Compilation_context.bin_annot cctx, cm_kind with
+     | false, _ | true, Ocaml Cmx -> other_targets, Command.Args.empty
+     | true, Ocaml (Cmi | Cmo) ->
+       let fn = Option.value_exn (Obj_dir.Module.cmt_file obj_dir m ~cm_kind ~ml_kind) in
+       let annots =
+         if Version.supports_bin_annot_occurrences ocaml.version
+         then [ "-bin-annot"; "-bin-annot-occurrences" ]
```
rgrinberg, line 192: spurious change? the previous version did not duplicate the flag name here

src/dune_rules/module_compilation.ml
```diff
@@ -183,23 +183,20 @@ let build_cm
...
-     | Ocaml (Cmi | Cmo) | Melange (Cmi | Cmj) ->
-       if Compilation_context.bin_annot cctx
-       then (
-         let fn =
-           Option.value_exn (Obj_dir.Module.cmt_file obj_dir m ~cm_kind ~ml_kind)
-         in
-         let annots =
-           [ "-bin-annot" ]
-           @
-           if Version.supports_bin_annot_occurrences ocaml.version
```
rgrinberg, line 197: How about just maintaining the old code but introducing:

```
let supports_bin_annot_occurences = Version.supports_bin_annot_occurences ocaml.version && not cm_kind <> Melange in
```

rgrinberg (changes requested):

anmonteiro:
actually, since melange supports 5.2, this is just a matter of supporting the flag. 

I bet it'd be nice to have occurrences support in Melange projects, too, so let me release a Melange version with support for this flag.

anmonteiro:
released melange 4.0.1-52 with this change: https://github.com/ocaml/opam-repository/pull/26043

voodoos:
Thanks @anmonteiro

I guess that means this PR is no more useful ?

anmonteiro:
I’d vote to close it

PR #10623, opened 2024-06-07
Indexing rules: correctly declare indexes dependencies

voodoos:
I overlooked dependencies in the indexing rules: I though that passing build folders of transitive deps as command args to the indexer was also declaring dependencies on these libs. This is actually not the case.

This means that indexes are not correctly rebuilt when one of their dependencies is updated.

This PR adds a set of `Hidden_deps` that correspond to all the index files of the direct dependencies local to the build.

@emillon sadly this means that the rules released in 3.16 are not working properly. Do you think there will be a patch release ?

### DIFF
--- a/src/dune_rules/merlin/ocaml_index.ml
+++ b/src/dune_rules/merlin/ocaml_index.ml
@@ -20,11 +20,10 @@ let cctx_rules cctx =
      definitions are used by all the cmts of modules in this cctx. *)
   let sctx = Compilation_context.super_context cctx in
   let dir = Compilation_context.dir cctx in
-  let open Memo.O in
-  let* aggregate =
+  let aggregate =
     let obj_dir = Compilation_context.obj_dir cctx in
     let fn = index_path_in_obj_dir obj_dir in
-    let* additional_libs =
+    let additional_libs =
       let open Resolve.Memo.O in
       let+ non_compile_libs =
         (* The indexer relies on the load_path of cmt files. When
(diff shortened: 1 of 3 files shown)


emillon:
3.16 is not out yet, so we can backport the change or revert it.

src/dune_rules/merlin/ocaml_index.ml
```diff
@@ -63,7 +81,8 @@ let cctx_rules cctx =
       ; A "-o"
       ; Target fn
       ; Deps modules_deps
-      ; Dyn (Resolve.Memo.read additional_libs)
```
rgrinberg, line 66: Could you go back to using `Dyn`?

rgrinberg (changes requested):

src/dune_rules/merlin/ocaml_index.ml
voodoos: `Resolve.args`  felt more idiomatic, but I can switch back of course. Is there a non-cosmetic reason which make using ` Dyn`  better ? Because it pushes the memos a bit deeper ?

rgrinberg: There's indeed a non cosmetic reason. In the previous code, we'd compute the dependencies when we needed to run the action. While in your code, we compute the dependencies when loading the rules. A simplified summary of how dune executes a rule will be helpful here. It goes like this:

1. Given the target we're building, find the directory of this target
2. In this directory, load the rules.
3. Find the rule that corresponds to the target
4. Compute the dependencies of the rule
5. Build the dependencies of the rule
6. Execute the rule to build the target.

Your change pushes work from step 4 to step 2. This is unnecessary work when we're loading the rules in the directory to build some other rule.

voodoos: Thanks a lot for the detailed explanation, I made the change in [cf0a087](https://github.com/ocaml/dune/pull/10623/commits/cf0a0873a636d829e12b7b8f122a4ff10d88ac46) and I will keep it in mind for my future contributions.

rgrinberg (approved):

voodoos:
Just merged, thanks for the review @rgrinberg

PR #10624, opened 2024-06-07
Implement support for multiple checksums

Leonidas-from-XIV:
OPAM supports multiple checksums by validating every checksum against the file, thus allowing migrations to stronger checksums, so this PR adds the support to `Fetch` and makes the toolchains use it.

### DIFF
--- a/src/dune_pkg/fetch.mli
+++ b/src/dune_pkg/fetch.mli
@@ -17,7 +17,7 @@ type failure =
     available at the location. *)
 val fetch
   :  unpack:bool
-  -> checksums:Checksum.t Nonempty_list.t option
+  -> checksums:Checksum.t list
   -> target:Path.t
   -> url:Loc.t * OpamUrl.t
   -> (unit, failure) result Fiber.t
--- a/src/dune_pkg/fetch.mli
+++ b/src/dune_pkg/fetch.mli
@@ -1,20 +1,23 @@
 open Import
 
 type failure =
-  | Checksum_mismatch of Checksum.t
+  | Checksum_mismatch of
+      { expected : Checksum.t
+      ; got : Checksum.t
+      }
   | Unavailable of User_message.t option
 
-(** [fetch ~checksum ~target url] will fetch [url] into [target]. It will verify
-    the downloaded file against [checksum], unless it [checksum] is [None].
+(** [fetch ~checksums ~target url] will fetch [url] into [target]. It will verify
+    the downloaded file against [checksums], unless it [checksums] is [None].
 
     return [Error (Checksum_mismatch _)] When the downloaded file doesn't match
-    the expected [checksum], this will pass the actually computed checksum.
+    the expected [checksums], this will return the actually computed checksum.
 
     return [Error (Unavailable _))] When the file can't be retrieved, e.g. not
     available at the location. *)
 val fetch
   :  unpack:bool
-  -> checksum:Checksum.t option
+  -> checksums:Checksum.t Nonempty_list.t option
   -> target:Path.t
   -> url:Loc.t * OpamUrl.t
   -> (unit, failure) result Fiber.t
--- a/otherlibs/stdune/src/nonempty_list.ml
+++ b/otherlibs/stdune/src/nonempty_list.ml
@@ -8,8 +8,3 @@ let of_list = function
 ;;
 
 let to_list (x :: xs) = List.cons x xs
-
-let fold_left ~f ~init (x :: xs) =
-  let init = f init x in
-  List.fold_left ~f ~init xs
-;;
--- a/otherlibs/stdune/src/nonempty_list.ml
+++ b/otherlibs/stdune/src/nonempty_list.ml
@@ -8,3 +8,8 @@ let of_list = function
 ;;
 
 let to_list (x :: xs) = List.cons x xs
+
+let fold_left ~f ~init (x :: xs) =
+  let init = f init x in
+  List.fold_left ~f ~init xs
+;;
--- a/otherlibs/stdune/src/nonempty_list.mli
+++ b/otherlibs/stdune/src/nonempty_list.mli
@@ -5,3 +5,4 @@ type 'a t = ( :: ) of 'a * 'a list
 val hd : 'a t -> 'a
 val of_list : 'a list -> 'a t option
 val to_list : 'a t -> 'a list
+val fold_left : f:('acc -> 'a -> 'acc) -> init:'acc -> 'a t -> 'acc
--- a/otherlibs/stdune/src/nonempty_list.mli
+++ b/otherlibs/stdune/src/nonempty_list.mli
@@ -5,4 +5,3 @@ type 'a t = ( :: ) of 'a * 'a list
 val hd : 'a t -> 'a
 val of_list : 'a list -> 'a t option
 val to_list : 'a t -> 'a list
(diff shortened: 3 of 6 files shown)


gridbugs (approved):

gridbugs:
My only suggestion is that it might be worth targeting main rather than toolchains with this PR to simplify merging the toolchains branch into main (either way is fine though!).

Leonidas-from-XIV:
That's a good point. I was thinking about it and wasn't sure which way would make it easier. But generally I do prefer targetting `main`, will try to rebase and see what happens.

christinerose (approved):

christinerose (approved):

src/dune_pkg/fetch.mli
```diff
@@ -1,20 +1,23 @@
...
     return [Error (Checksum_mismatch _)] When the downloaded file doesn't match
-    the expected [checksum], this will pass the actually computed checksum.
+    the expected [checksums], this will return the actually computed checksum.
 
     return [Error (Unavailable _))] When the file can't be retrieved, e.g. not
     available at the location. *)
 val fetch
   :  unpack:bool
-  -> checksum:Checksum.t option
+  -> checksums:Checksum.t Nonempty_list.t option
```
rgrinberg, line 20: I think this is equivalent to just a list of checksums

rgrinberg (commented):
TBH, I don't really see the point. Moreover, this only works for opam packages from the opam repository and I would hate to clutter dune's lock files for something without a defined benefit. What about just using the "best" checksum when it's available and not bothering with the rest.

src/dune_pkg/fetch.mli
Leonidas-from-XIV: I know, but what I want to express is that either there are checksums in which case there should be at least one or there is specifically no checksum. This avoids the case where "there are checksums" but the checksums in question are an empty list.

Leonidas-from-XIV:
>  this only works for opam packages from the opam repository

I don't understand this comment, as all packages are from the OPAM repository?

> What about just using the "best" checksum when it's available and not bothering with the rest.

The issue I see is that there is no good metric for "best" checksum as what is "best" is not an objective property but one that is dependent on ongoing research (which initially got me down the rabbit hole of implementing this). A breakthrough in SHA2 might as well make SHA1 "better" down the line, thus supporting all checksums has the advantage of not taking sides on the Dune end (thus deferring the choice to OPAM-Repo maintainers and package authors) and also make for a better migration strategy (if e.g. opam-repo adds SHA2 to all MD5 packages the lockfiles now would all change checksums which is easier to audit if you know that the MD5 in the lockfiles stayed the same).

Another point is that it just matches what `opam` is doing, thus all concerns wrt to checksums in `opam` would apply to `dune`, but if Dune does something different it might require a different approach. In my opinion matching what `opam` does in areas where doing things differently does not give us an advantage is a good thing.

src/dune_pkg/fetch.mli
rgrinberg: I don't understand the difference between

> specifically no checksum

and

>  "there are checksums" but the checksums in question are an empty list.

And it's not for the lack of trying. I suggest that I as a reader don't see the difference, it's not going to be obvious to many other reads either. In which case, no additional clarity is gained here.

rgrinberg:
> I don't understand this comment, as all packages are from the OPAM repository?

I don't think they are. Some package sources come from git pins, where we rely on the sha signature of git commits.

But more importantly, we mostly verify checksums when we're building from a lock directory. In a lock directory, we're still saving only one checksum right?

Leonidas-from-XIV:
> I don't think they are. Some package sources come from git pins, where we rely on the sha signature of git commits.

For the majority future users I would assume the amount of pins to be very few to none. I investigated the amount of local switches I personally have (37 at time of writing) and the histogram shows that the majority (27) of them don't use any pins at all:

```sh
$ sort < switch-pin-counts | uniq -c
     27 0
      5 1
      3 2
      1 3
      1 6
```

Even in the package with 6 pins the amount of opam dependencies outnumbers the amount of pins.

In the case of git repos, we only have one checksum, but that use case is still covered by this PR.

> In a lock directory, we're still saving only one checksum right?

Yes, I am working on commits atop of this PR (where at the momet I simply wrap the checksum into a 1-element list) that extend the support to read and write multiple checksums in lockfiles. This can happen as part of this PR or in a future one, depending on whether we include the support or not.

src/dune_pkg/fetch.mli
Leonidas-from-XIV: The distinction between "no content" and "there is at least one thing" is in my eyes a pretty core part of `option` types. Because we can of course always replace `None` with `[]` and `Some x` with `[x]` but the option type signifies that the absence of content triggers a different behavior than its presence. In the dune codebase I have the impression that we try to be explicit that something is missing, e.g. through `Loc.none` in the case of location info.

If we just have a list, it is easy to assume you have checksums but the list of checksums is 0 thus you don't actually have checksums. Empty lists operate like lists with content under most list operators so it is easy to pass an empty list where you didn't mean to because the previous function that computed you the checksums served you with an empty list (e.g. incorrectly as result of a bug).

This is also why I use `Nonempty_list.t` because otherwise it would be hard to understand whether there is a semantic difference between `None` and `Some []`.

Leonidas-from-XIV: I removed it now, but we now have `fetch ~checksums:[]` and `fetch_without_checksum` which seem to be doing roughly the same. Maybe the right way would be to preserve the `Nonempty_list.t` and use `fetch_without_checksum` in places where we don't expect checksums.

rgrinberg:
Pins are scarcely used because they are unnecessarily complicated to use in opam. It's not going to be the case in the future.

IMO, it's not worth working on this feature. Taking it to the extreme would mean polluting all the lock dirs with a bunch of extra checksums that are going to make diffs readable. There's also some non trivial complexity with `fetch_rules.ml` that you will have to address. There we setup rules keyed on their digest to make sure we don't re-download things. You will need to make sure account for multiple hashes now. IMHO there's more important things to do.

Leonidas-from-XIV:
> It's not going to be the case in the future.

It's much easier to specify that you want this and that package than to figure out the git repo of the package and pinning a tag, which also opts you out of all the improvements on the package metadata that opam-repo maintainers are doing in the meantime.

Plus uploading packages with pins to opam-repository is a no-go, thus while slightly more convenient I expect pins to stay a rarely used feature.

src/dune_pkg/fetch.mli
rgrinberg: I think I understand where you were going with this, but I think that clarifying the distinction can only be done with an explicit constructor name. E.g. `No_checksums | At_least of Nonempty_list.t`. Otherwise, you just replaced Cons and Nil with Some, and None.

FTR, this is my interpretation of how the function works:

It takes a list of checksums and verifies all of them.
If there are no checksums in the list, there are no checksums to verify.
If there's one checksum in the list, there is 1 checksum to verify.
If there's two checksums in the list, ...

Leonidas-from-XIV: Fair enough point about the constructor, I agree that that would make it more explicit than `Some (Cons ...)`. In my mind one of the reasons we had a distinction between `fetch` and `fetch_without_checksums` is to make sure that if the user picked `fetch` would have (at least one) checksum(s) passed.

PR #10631, opened 2024-06-09
[coq] make expand "coq.version" and co not depend on coq-stdlib

### LINKED ISSUE #10629: [coq] expand coq.version & co without using coq-lib
The file `coq_config.ml` is able to expand variables by two means:
- `coqc --print-version`
- `coqc -config`

The former works fine with just coq-core being built (if one also passes `-boot -noinit`). The use case is to compose with other projects using the coq.version variable (for example to feed ppx_optcomp) and run `dune build @check`.

Unfortunately the file `coq_config.ml` first computes all data and then checks which variable to expand. Hence, both commands are run, and the latter requires a more "complete" build of coq to work.

It would be nice if the `by_name` dispatching could be used upfront, and only call the command needed in order to fulfill the request.

gares:
fix #10629

### DIFF
--- a/doc/changes/10631.md
+++ b/doc/changes/10631.md
@@ -1,0 +1,2 @@
+- Don't depend on coq-stdlib being installed when expanding variables
+  of the `coq.version` family (#10631, fixes #10629, @gares)
--- a/src/dune_rules/coq/coq_config.ml
+++ b/src/dune_rules/coq/coq_config.ml
@@ -154,7 +154,7 @@ module Version = struct
          Return
          bin
          (* we pass -boot since this means it will work with just coq-core *)
-         [ "--print-version"; "-boot" ]
+         [ "--print-version"; "-boot"; "-noinit" ]
   ;;
 
   let version_memo = Memo.create "coq-and-ocaml-version" ~input:(module Path) impl_version
--- a/src/dune_rules/coq/coq_config.ml
+++ b/src/dune_rules/coq/coq_config.ml
@@ -195,61 +195,58 @@ module Version = struct
   ;;
 end
 
-module Config = struct
-  type t =
-    { coqlib : Path.t
-    ; coqcorelib : Path.t option (* this is not available in Coq < 8.14 *)
-    ; coq_native_compiler_default :
-        string option (* this is not available in Coq < 8.13 *)
-    }
+type t =
+  { coqlib : Path.t
+  ; coqcorelib : Path.t option (* this is not available in Coq < 8.14 *)
+  ; coq_native_compiler_default : string option (* this is not available in Coq < 8.13 *)
+  }
 
-  let impl_config bin =
-    let* _ = Build_system.build_file bin in
-    Memo.of_reproducible_fiber
-    @@ Process.run_capture_lines
-         ~display:Quiet
-         ~stderr_to:
-           (Process.Io.make_stderr
-              ~output_on_success:Swallow
-              ~output_limit:Execution_parameters.Action_output_limit.default)
-         Return
-         bin
-         [ "--config" ]
-  ;;
+let impl_config bin =
+  let* _ = Build_system.build_file bin in
+  Memo.of_reproducible_fiber
+  @@ Process.run_capture_lines
+       ~display:Quiet
+       ~stderr_to:
+         (Process.Io.make_stderr
+            ~output_on_success:Swallow
+            ~output_limit:Execution_parameters.Action_output_limit.default)
+       Return
+       bin
+       [ "--config" ]
+;;
 
-  let config_memo = Memo.create "coq-config" ~input:(module Path) impl_config
+let config_memo = Memo.create "coq-config" ~input:(module Path) impl_config
 
-  let make ~(coqc : Action.Prog.t) =
-    let open Memo.O in
-    let+ config_output =
-      get_output_from_config_or_version ~coqc ~what:"--config" config_memo
-    in
-    let open Result.O in
-    let* config_output = config_output in
-    match Vars.of_lines config_output with
-    | Ok vars ->
-      let coqlib = Vars.get_path vars "COQLIB" in
-      (* this is not available in Coq < 8.14 *)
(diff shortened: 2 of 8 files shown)


src/dune_rules/coq/coq_config.mli
```diff
@@ -1,24 +1,5 @@
 open Import
 
-(** Data of a Coq configuration. *)
-type t
```
ejgallego, line 4: I think that using our coding style this still should be `Coq_config.t` , @Alizter ?

ejgallego (commented):
Thanks for the patch, indeed it seems very useful!

I'll let @Alizter review first, as he is the expert on this code.

Can you add changelog meanwhile (add a new file in `doc/changes`)

gares:
@Alizter please take a look at
https://github.com/ocaml/dune/pull/10631/files#diff-3d90c4b942eed2beaba48d8cbbd71cbe1ce22b6b7a61ea76999e92cf0a9a13e8
you are already testing this for coq 8.16

Alizter:
@gares Thanks, I forgot all the details there. I will take a closer look at this PR soon.

gares:
ping

doc/changes/10631.md
```diff
@@ -0,0 +1,2 @@
+- Don't depend con coq-stdlib being installed when expanding variables
```
Alizter, line 1:
```suggestion
- Don't depend on coq-stdlib being installed when expanding variables
```

src/dune_rules/coq/coq_config.mli
Alizter: Yes, but I think it would be too confusing to split this further into separate files. I'm fine with it like this.

test/blackbox-tests/test-cases/coq/failed-config.t/run.t
```diff
@@ -104,22 +104,16 @@ Here we query the version of Coq. Due to the expansion of %{coq:_} macros we nee
 Fails for now, but could be improved in the future.
```
Alizter, line 104: Has improved now, we can update the comment. We can say something like: Querying the Coq version is not affected by the inability to query the Coq configuration.

src/dune_rules/coq/coq_config.ml
```diff
@@ -183,112 +183,107 @@ module Version = struct
...
-  let s = Pform.Macro_invocation.Args.whole macro in
-  let open Memo.O in
-  let* coqc = Artifacts.binary artifacts_host ~where:Original_path ~loc:None "coqc" in
-  let+ t = make ~coqc in
-  match t with
-  | Error msg ->
-    User_error.raise
-      ~loc:(Dune_lang.Template.Pform.loc source)
-      [ Pp.textf "Could not expand %%{coq:%s} as running coqc --config failed." s; msg ]
-      ~hints:
```
Alizter, line 279: I think it still makes sense to keep this hint right?

test/blackbox-tests/test-cases/coq/failed-config.t/run.t
```diff
@@ -139,10 +133,8 @@ Should fail.
   File "dune", line 4, characters 8-21:
   4 |   (echo %{coq:coqlib})))
               ^^^^^^^^^^^^^
-  Error: Could not expand %{coq:coqlib} as running coqc --config failed.
+  Error: Could not expand %{coq:coqlib} as running coqc failed.
   $TESTCASE_ROOT/bin/coqc --config failed with exit code 1.
-  Hint: coqc --config requires the coq-stdlib package in order to function
```
Alizter, line 144: I would have preferred for this hint to stay, but I see that it's not so easy to reuse the error in both the version and config cases so I guess this is fine for now.

test/blackbox-tests/test-cases/coq/failed-config.t/run.t
gares: done

gares:
Thanks for the review, I think I did implement the requested changes

Alizter (approved):
@ejgallego I am happy with these changes. Would you like to take another look now.

src/dune_rules/coq/coq_config.mli
ejgallego: I think we should not split, but keep `Coq_config.t` (as is now on `main`) , and add a `Version` module so we have `Coq_config.Version.t` too, both on the same file. WDYT?

src/dune_rules/coq/coq_config.mli
```diff
@@ -31,22 +12,57 @@ module Value : sig
...
+      - version.major
+      - version.minor
+      - version.revision
+      - version.suffix
+      - version
+      - ocaml-version *)
+  val by_name : t -> string -> Value.t Option.t
+end
+
+module Config : sig
```
ejgallego, line 43: This module is redundant, as it is already inside `Coq_config` , which will became `Coq.Config` at some point when we update Dune's sources to use qualified subdirs.

cc: @Alizter @gares

ejgallego:
> @ejgallego I am happy with these changes. Would you like to take another look now.

Thanks @Alizter , I will process this likely early next week, thanks for the help.

I think tho the PR needs a minor tweak to about the redundant `Coq_config.Config`, other than that, seems good to me.

gares:
done

ejgallego (approved):
Great!

test/blackbox-tests/test-cases/coq/failed-config.t/run.t
ejgallego: Indeed let's not delay the PR on this, but for sure it would be helpful to users to be more specific about errors in this code. This kind of small stuff tends to matter a lot w.r.t. usability.

ejgallego:
@gares format CI job failed (you can also rebase on the way as github is complaining about it)

src/dune_rules/coq/coq_config.mli
Alizter: Doesn't make a difference to me.

gares:
hoefully done

PR #10634, opened 2024-06-10
pkg(fix): use upstream or overlay as repository name (#10605)

### LINKED ISSUE #10605: pkg: `dune pkg lock` raises a code error if you try to name a repository "upstream"
My dune-workspace:
```
(lang dune 3.15)

(lock_dir
 (repositories upstream))

(repository
 (name upstream)
 (source "git+https://github.com/ocaml/opam-repository"))
```

Running `dune pkg lock` gives the error:
```
Description:
  ("Map.of_list_exn", { key = "upstream" })
Called from Dune__exe__Lock.solve_lock_dir in file "bin/pkg/lock.ml", line
  53, characters 4-39
  line 76, characters 8-11

```

moyodiallo:
Fixes #10605

### DIFF
--- a/src/dune_rules/workspace.ml
+++ b/src/dune_rules/workspace.ml
@@ -856,7 +856,11 @@ let step1 clflags =
        in
        let defined_names = ref Context_name.Set.empty in
        let env = Lazy.force env in
-       let repos = default_repositories @ List.map ~f:Lazy.force repos in
+       let repos =
+         if List.is_empty repos
+         then default_repositories
+         else List.map ~f:Lazy.force repos
+       in
        let merlin_context =
          List.fold_left contexts ~init:None ~f:(fun acc ctx ->
            let name = Context.name ctx in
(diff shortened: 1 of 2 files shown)


src/dune_rules/workspace.ml
```diff
@@ -856,7 +856,11 @@ let step1 clflags =
        in
        let defined_names = ref Context_name.Set.empty in
        let env = Lazy.force env in
-       let repos = default_repositories @ List.map ~f:Lazy.force repos in
+       let repos =
+         match List.map ~f:Lazy.force repos with
+         | [] -> default_repositories
```
emillon, line 861:
```
if List.is_empty repos then default_repositories else List.map ~f:Lazy.force repos
```

rgrinberg (changes requested):
Could you add a description to the PR so that I can understand the intention of the fix?

moyodiallo:
> Could you add a description to the PR so that I can understand the intention of the fix?

My bad, this PR is not the good fix. And overriding won't be the good fix too, because we could have two different lock_dirs, one overrides the name `upstream`  and the other don't.

PR #10639, opened 2024-06-11
Add toolchains feature

gridbugs:
Toolchains refers to a feature where dune downloads and builds ocaml compiler packages using an alternative mechanism to regular dune package management. Toolchains are installed into
$XDG_CACHE_DIR/dune/toolchains/<name.version>. Both the ocaml-base-compiler and ocaml-variants compilers are supported. In the case of the latter, multiple different configurations may be installed concurrently, by concatenating the names of enabled options to the end of the name of the directory containing the toolchain.

The toolchains feature is a temporary workaround for the fact that the compiler packages are not currently relocatable. That is, the artifacts from the compiler packages built inside dune's build sandbox cannot be used after moving them into the dune cache or their eventual location within the _build directory. Work is currently being done to make the compiler relocatable, at which point dune will be able to build compiler packages in its sandbox and the toolchains feature can be deleted.

### DIFF
--- a/src/dune_pkg/checksum.mli
+++ b/src/dune_pkg/checksum.mli
@@ -20,6 +20,3 @@ val pp : t -> 'a Pp.t
 val equal : t -> t -> bool
 
 module Map : Map.S with type key = t
-
-(** Returns the hash whose kind is the strongest (e.g. SHA512 is stronger than MD5) *)
-val choose_strongest : t list -> t option
--- a/src/dune_pkg/checksum.mli
+++ b/src/dune_pkg/checksum.mli
@@ -20,3 +20,6 @@ val pp : t -> 'a Pp.t
 val equal : t -> t -> bool
 
 module Map : Map.S with type key = t
+
+(** Returns the hash whose kind is the strongest (e.g. SHA512 is stronger than MD5) *)
+val choose_strongest : t list -> t option
--- a/src/dune_pkg/flock.ml
+++ b/src/dune_pkg/flock.ml
@@ -11,35 +11,28 @@ type t =
    lock" messages multiple times when multiple concurrent fibers try
    to take a lock at the same time. *)
 module Global_waiting_names = struct
-  let state = lazy (String.Table.create 1)
+  let state = String.Table.create 1
 
-  (* add a name to the set, returning [true] iff the name wasn't
+  (* Add a name to the set, returning [true] iff the name wasn't
      already in the set *)
-  let add name =
-    let state = Lazy.force state in
-    String.Table.add state name () |> Result.is_ok
-  ;;
-
-  let remove name =
-    let state = Lazy.force state in
-    String.Table.remove state name
-  ;;
+  let add name = String.Table.add state name () |> Result.is_ok
+  let remove name = String.Table.remove state name
 end
 
-let attempt_to_lock { flock; lock_path } ~name_for_messages ~timeout_s =
+let attempt_to_lock { flock; lock_path } ~name_for_messages ~timeout_seconds =
   let open Fiber.O in
   let current_dune_pid = Unix.getpid () in
-  let rec loop timeout_s =
+  let rec loop timeout_seconds =
     match Flock.lock_non_block flock Flock.Exclusive with
     | Error e -> Fiber.return @@ Error e
     | Ok `Success ->
       Global_waiting_names.remove name_for_messages;
       Fiber.return (Ok `Success)
-    | Ok `Failure -> handle_failure timeout_s
-  and handle_failure timeout_s =
+    | Ok `Failure -> handle_failure timeout_seconds
+  and handle_failure timeout_seconds =
     let locked_by_pid = int_of_string (Io.read_file lock_path) in
     let sleep_duration_s = 0.1 in
-    let remaining_duration_s = timeout_s -. sleep_duration_s in
+    let remaining_duration_s = timeout_seconds -. sleep_duration_s in
     if remaining_duration_s <= 0.0
     then Fiber.return (Ok `Timeout)
     else (
@@ -64,10 +57,10 @@ let attempt_to_lock { flock; lock_path } ~name_for_messages ~timeout_s =
       let* () = Scheduler.sleep ~seconds:sleep_duration_s in
       loop remaining_duration_s)
   in
-  loop timeout_s
+  loop timeout_seconds
 ;;
 
-let with_flock lock_path ~name_for_messages ~timeout_s ~f =
+let with_flock lock_path ~name_for_messages ~timeout_seconds ~f =
   let open Fiber.O in
   let parent = Path.parent_exn lock_path in
   Path.mkdir_p parent;
@@ -84,7 +77,7 @@ let with_flock lock_path ~name_for_messages ~timeout_s ~f =
       let+ () = Fiber.return () in
       Unix.close fd)
     (fun () ->
-      attempt_to_lock { flock; lock_path } ~name_for_messages ~timeout_s
+      attempt_to_lock { flock; lock_path } ~name_for_messages ~timeout_seconds
       >>= function
       | Ok `Success ->
         Fiber.finalize
@@ -108,8 +101,8 @@ let with_flock lock_path ~name_for_messages ~timeout_s ~f =
               Unix_error.Detailed.create ue ~syscall:"flock" ~arg:"unlock"
               |> Unix_error.Detailed.raise)
       | Ok `Timeout ->
-        let locked_by_pid = int_of_string (Io.read_file lock_path) in
-        if locked_by_pid == current_dune_pid
+        let locked_by_pid = Int.of_string_exn (Io.read_file lock_path) in
+        if Int.equal locked_by_pid current_dune_pid
         then
           Code_error.raise
             "timeout while waiting for flock, but flock was currently held by the \
@@ -133,7 +126,7 @@ let with_flock lock_path ~name_for_messages ~timeout_s ~f =
             [ Pp.textf
                 "Timed out after %.2f seconds while waiting for another instance of dune \
                  (pid %d) to release the lock on the resource %S."
-                timeout_s
+                timeout_seconds
                 locked_by_pid
                 name_for_messages
             ]
--- a/src/dune_pkg/flock.ml
+++ b/src/dune_pkg/flock.ml
@@ -1,0 +1,148 @@
+open! Stdune
+module Flock = Dune_util.Flock
+module Scheduler = Dune_engine.Scheduler
+
+type t =
+  { flock : Flock.t
+  ; lock_path : Path.t
+  }
+
+(* Global mutable set of names, used to prevent printing "waiting for
+   lock" messages multiple times when multiple concurrent fibers try
+   to take a lock at the same time. *)
+module Global_waiting_names = struct
+  let state = lazy (String.Table.create 1)
+
+  (* add a name to the set, returning [true] iff the name wasn't
+     already in the set *)
+  let add name =
+    let state = Lazy.force state in
+    String.Table.add state name () |> Result.is_ok
+  ;;
+
+  let remove name =
+    let state = Lazy.force state in
+    String.Table.remove state name
+  ;;
+end
+
+let attempt_to_lock { flock; lock_path } ~name_for_messages ~timeout_s =
+  let open Fiber.O in
+  let current_dune_pid = Unix.getpid () in
+  let rec loop timeout_s =
+    match Flock.lock_non_block flock Flock.Exclusive with
+    | Error e -> Fiber.return @@ Error e
+    | Ok `Success ->
+      Global_waiting_names.remove name_for_messages;
+      Fiber.return (Ok `Success)
+    | Ok `Failure -> handle_failure timeout_s
+  and handle_failure timeout_s =
+    let locked_by_pid = int_of_string (Io.read_file lock_path) in
+    let sleep_duration_s = 0.1 in
+    let remaining_duration_s = timeout_s -. sleep_duration_s in
+    if remaining_duration_s <= 0.0
+    then Fiber.return (Ok `Timeout)
+    else (
+      if locked_by_pid <> current_dune_pid && Global_waiting_names.add name_for_messages
+      then
+        (* Only print this message if the dune process that holds the
+           lock isn't the current process and this is the first time
+           that the current process has failed to take the lock since
+           the last time it successfully took the lock. This prevents
+           the situation where multiple fibers all attempt to take the
+           lock concurrently while it's held by another process from
+           causing the following message from being printed multiple
+           times. *)
+        User_message.print
+          (User_message.make
+             [ Pp.textf
+                 "Waiting for another instance of dune (pid %d) to release the lock for \
+                  the resource %S..."
+                 locked_by_pid
+                 name_for_messages
+             ]);
+      let* () = Scheduler.sleep ~seconds:sleep_duration_s in
+      loop remaining_duration_s)
+  in
+  loop timeout_s
+;;
+
+let with_flock lock_path ~name_for_messages ~timeout_s ~f =
+  let open Fiber.O in
+  let parent = Path.parent_exn lock_path in
+  Path.mkdir_p parent;
+  let fd =
+    Unix.openfile
+      (Path.to_string lock_path)
+      [ Unix.O_CREAT; O_WRONLY; O_SHARE_DELETE; O_CLOEXEC ]
+      0o600
+  in
+  let flock = Flock.create fd in
+  let current_dune_pid = Unix.getpid () in
+  Fiber.finalize
+    ~finally:(fun () ->
+      let+ () = Fiber.return () in
+      Unix.close fd)
+    (fun () ->
+      attempt_to_lock { flock; lock_path } ~name_for_messages ~timeout_s
+      >>= function
+      | Ok `Success ->
+        Fiber.finalize
+          (fun () ->
+            Dune_util.Global_lock.write_pid fd;
+            f ())
+          ~finally:(fun () ->
+            let+ () = Fiber.return () in
+            match Flock.unlock flock with
+            | Ok () ->
+              (* Note that after the lock has been released, we
+                 deliberately don't delete the lock file to avoid a race
+                 condition where other processes or fibers still need to
+                 read the file to determine the process that held the
+                 lock. Even though the lock has been released, other
+                 parties may be in between timing out waiting for the
+                 lock and reading the lock file to get the pid to
+                 include in their error message. *)
+              ()
+            | Error ue ->
+              Unix_error.Detailed.create ue ~syscall:"flock" ~arg:"unlock"
+              |> Unix_error.Detailed.raise)
+      | Ok `Timeout ->
+        let locked_by_pid = int_of_string (Io.read_file lock_path) in
+        if locked_by_pid == current_dune_pid
+        then
+          Code_error.raise
+            "timeout while waiting for flock, but flock was currently held by the \
+             current process"
+            [ "name_for_messages", Dyn.string name_for_messages ]
+        else
+          User_error.raise
+            ~hints:
+              [ Pp.textf
+                  "Another instance of dune (pid %d) currently holds the lock for the \
+                   resource %S. If this is unexpected, terminate that process and re-run \
+                   the command."
+                  locked_by_pid
+                  name_for_messages
+              ; Pp.textf
+                  "As a last resort, if the other instance of dune (pid %d) is no longer \
+                   running, manually delete the lock file %s."
+                  locked_by_pid
+                  (Path.to_string_maybe_quoted lock_path)
+              ]
+            [ Pp.textf
+                "Timed out after %.2f seconds while waiting for another instance of dune \
+                 (pid %d) to release the lock on the resource %S."
+                timeout_s
+                locked_by_pid
+                name_for_messages
+            ]
+      | Error error ->
+        User_error.raise
+          [ Pp.textf
+              "Failed to get a lock for the resource %S with lock file %s: %s"
+              name_for_messages
+              (Path.to_string_maybe_quoted lock_path)
+              (Unix.error_message error)
+          ])
+;;
--- a/src/dune_pkg/flock.mli
+++ b/src/dune_pkg/flock.mli
@@ -1,0 +1,23 @@
+open! Stdune
+
+(** [with_flock path ~name_for_messages ~timeout_s ~f] ensures mutual
+    exclusion for the function [f] across multiple concurrent
+    instances of dune, using the lock file at [path] to coordinate
+    between different dune instances. If the lock is not acquired
+    after [timeout_s] seconds then a [User_error] is raised. Pass
+    [infinity] to keep trying to take the lock
+    forever. [name_for_messages] is the name used to refer to this
+    lock in error messages.
+
+    Within the a dune process, this function also ensures mutual
+    exclusion between fibers. Note that if this function times out
+    waiting for the lock while the lock is held by a different fiber
+    of the same dune process, a [Code_error] is raised rather than a
+    [User_error]. If a timeout is possible, avoid allowing multiple
+    fibers to concurrently attempt to take a flock. *)
+val with_flock
+  :  Path.t
+  -> name_for_messages:string
+  -> timeout_s:float
+  -> f:(unit -> 'a Fiber.t)
+  -> 'a Fiber.t
--- a/src/dune_pkg/flock.mli
+++ b/src/dune_pkg/flock.mli
@@ -1,11 +1,11 @@
 open! Stdune
 
-(** [with_flock path ~name_for_messages ~timeout_s ~f] ensures mutual
-    exclusion for the function [f] across multiple concurrent
+(** [with_flock path ~name_for_messages ~timeout_seconds ~f] ensures
+    mutual exclusion for the function [f] across multiple concurrent
     instances of dune, using the lock file at [path] to coordinate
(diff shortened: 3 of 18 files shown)


src/dune_rules/ocaml_toolchain.mli
```diff
@@ -26,6 +26,12 @@ val of_env_with_findlib
 
 val of_binaries : path:Path.t list -> Context_name.t -> Env.t -> Path.Set.t -> t Memo.t
 
+val of_toolchain_compiler
```
Leonidas-from-XIV, line 29: Add a note that this is where it will attempt to download and build the compiler (it it doesn't exist).

Leonidas-from-XIV (commented):
Need to fix the tests as well. One of them needs to be promoted and the other ones are faulty (they remove `git` from the `PATH` where they should not), but also with the `toolchain` flag off we shouldn't be calling `git`

@moyodiallo mentions that the tests are slower which is probably due to `Available_compiler` shelling out to `git` through `Opam_repo`.

src/dune_pkg/toolchain.ml
```diff
@@ -0,0 +1,530 @@
...
+
+     Installation happens in two steps. First, run the package's
+     install command passing "compilerIR=..." to install the toolchain
+     into a temporary directory. Then the target directory from the
+     temporary directory is copied into the final installation
+     directory. This allows us to use the fact that the final
+     installation directory exists to check that the toolchain is
+     installed. *)
+  let install t =
+    let open Fiber.O in
```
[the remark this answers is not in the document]
Leonidas-from-XIV: This is called from a place that holds the (f)lock so its fine.

src/dune_rules/pkg_rules.ml
```diff
@@ -193,7 +195,114 @@ module Value_list_env = struct
...
+    in
+    add_path t Env_path.var toolchain_bin_dir
+  ;;
+end
+
+module DB = struct
+  type t =
+    { all : Lock_dir.Pkg.t Package.Name.Map.t
+    ; system_provided : Package.Name.Set.t
+    ; available_compilers : Toolchain.Available_compilers.t
```
moyodiallo, line 215: Why do we need `available_compilers` here ? We are loading this from `opam-repository` when it is not used.

Leonidas-from-XIV: Good point, I've removed it from the DB. It's a better approach than to make it an `option` as then the semantics of `DB.equal` are not dependent on `use_toolchains`.

src/dune_pkg/toolchain.mli
```diff
@@ -0,0 +1,33 @@
...
+open! Stdune
+
+(* [Compiler.t] represents a reference to a compiler installation in a shared
+   location.
+
+   It can be retrieved using [get] at which point it will either be returned
+   if it is installed or downloaded and built.
+*)
+module Compiler : sig
+  type uninit
```
rgrinberg, line 11: How about just combining find and get so that you don't need this?

gridbugs: Personally I like the fact that logic for looking up a compiler by name/version is separated from the logic for installing the compiler.

What do you see as the advantage of not exposing the uninstalled and installed types?

src/dune_pkg/toolchain.ml
```diff
@@ -0,0 +1,530 @@
...
+       must be possible to install multiple instances of the same version
+       of ocaml-variants at the same time, provided that they differ in
+       their dependencies. This means that the list of dependencies must be
+       present in the name of the directory where the compiler gets
+       installed.
+
+       An example identifier string for ocaml-variants with multiple
+       dependencies is:
+       ocaml-variants.5.2.0+options.with_deps.ocaml-option-flambda,ocaml-option-no-flat-float-array *)
+    let identifier_string t =
```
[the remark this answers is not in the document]
gridbugs: We discussed this further on slack and agreed to keep the support for ocaml-variants.

src/dune_rules/ocaml_toolchain.mli
gridbugs: I assume this comment was left before introducing a witness type into `Toolchain.Compiler`, as now `of_toolchain_compiler` no longer installs the compiler (it takes a compiler that's already been installed).

src/dune_rules/pkg_rules.ml
```diff
@@ -193,7 +195,112 @@ module Value_list_env = struct
...
+    let toolchain_bin_dir =
+      Toolchain.Compiler.bin_dir compiler |> Path.outside_build_dir
+    in
+    add_path t Env_path.var toolchain_bin_dir
+  ;;
+end
+
+module DB = struct
+  type t =
+    { all : Lock_dir.Pkg.t Package.Name.Map.t
```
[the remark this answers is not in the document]
gridbugs: I agree with that suggestion though note that `DB` module was just moved from below in this file and not introduced as part of the toolchains feature.

src/dune_rules/pkg_rules.ml
```diff
@@ -193,7 +195,112 @@ module Value_list_env = struct
...
+    ; system_provided : Package.Name.Set.t
```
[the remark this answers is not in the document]
gridbugs: I propose we make a new PR implementing the suggested improvements to DB.

gridbugs:
I added a few more doc comments to the `Toolchain` module and revised some of the existing comments. I also renamed `Compiler.uninit` to `Compiler.t` and renamed `Compiler.t` to `Compiler.installed` as I find that clearer. Now if an interface accepts a `Compiler.installed` it's clear from the type name that it expects to be passed an installed compiler.

There are a few more unaddressed comments on this PR that I will address tomorrow.

src/dune_pkg/checksum.mli
```diff
@@ -20,3 +20,6 @@ val pp : t -> 'a Pp.t
 val equal : t -> t -> bool
 
 module Map : Map.S with type key = t
+
+(** Returns the hash whose kind is the strongest (e.g. SHA512 is stronger than MD5) *)
+val choose_strongest : t list -> t option
```
rgrinberg, line 25: Did this make it in by accident? Doesn't seem related.

src/dune_pkg/toolchain.ml
```diff
@@ -0,0 +1,549 @@
...
+        (Package_version.to_string t.version)
+        deps_suffix
+    ;;
+
+    module Paths = struct
+      (* The path to the directory that will contain all toolchain
+         versions. Creates the directory if it doesn't already exist. *)
+      let toolchain_base_dir () =
+        let cache_dir =
+          Xdg.create ~env:Sys.getenv_opt ()
```
rgrinberg, line 73: There's already an instance of Xdg.t in Dune_util I believe

src/dune_pkg/toolchain.mli
rgrinberg: When adding new code, IMHO we should justify why we're adding the code rather than looking for reasons not to add it. So I would like to invert your question:

* Why do we need both an installed and uninstalled type?
* Can we do anything useful to an uninstalled toolchain?

If there are no clear answers to the questions above, then I would suggest the interface is not as simple as it could be. I'm usually biased towards minimalism, so I'm willing to be flexible and entertain the idea that I'm missing the point. However, I would like to point an example where I managed to eliminate such a seemingly unnecessary type and greatly simplify and clarify the interface in the process: https://github.com/ocaml/dune/pull/10071 (See `Remote.uninit`). I'm not saying this situation is analogous, but I have my suspicions.

If you are convinced that this type is necessary, feel free to resolve this comment.

gridbugs: Strictly speaking having a separate type for installed and uninstalled compilers is not necessary. However it allows us to separate the logic for choosing which compiler to install from the logic for installing a compiler, and I believe that this justifies the extra code. When consuming APIs I find that separating concerns like this helps me understand how to use them, and also helps me understand how to debug when things go wrong. The failure mode for choosing which compiler to install is very different from the failure mode for installing the compiler, and I expect it to aid debugging if these tasks are performed by different functions.

src/dune_pkg/checksum.mli
gridbugs: It's included deliberately. This function is called in toolchain.ml. It lets us validate the compiler package source archive against the strongest checksum in its opam file. Ideally we'd validate it against all its checksums but our `Fetch.fetch` function doesn't currently support this (support will be added in https://github.com/ocaml/dune/pull/10624).

src/dune_pkg/flock.ml
```diff
@@ -0,0 +1,148 @@
...
+                 parties may be in between timing out waiting for the
+                 lock and reading the lock file to get the pid to
+                 include in their error message. *)
+              ()
+            | Error ue ->
+              Unix_error.Detailed.create ue ~syscall:"flock" ~arg:"unlock"
+              |> Unix_error.Detailed.raise)
+      | Ok `Timeout ->
+        let locked_by_pid = int_of_string (Io.read_file lock_path) in
+        if locked_by_pid == current_dune_pid
```
rgrinberg, line 112: No physical equality please

src/dune_pkg/checksum.mli
rgrinberg: Do we choose the "strongest" checksum for regular packages? Why is it special cased for toolchains?

src/dune_rules/pkg_rules.ml
```diff
@@ -193,7 +195,114 @@ module Value_list_env = struct
...
+
+  let get =
+    let dune = Package.Name.Set.singleton (Package.Name.of_string "dune") in
+    fun context ->
+      let+ all = Lock_dir.get context in
+      { all = all.packages; system_provided = dune }
+  ;;
+end
+
+module Compiler = struct
```
rgrinberg, line 230: Can this module be extracted to its own file?

src/dune_pkg/flock.ml
```diff
@@ -0,0 +1,148 @@
...
+type t =
+  { flock : Flock.t
+  ; lock_path : Path.t
+  }
+
+(* Global mutable set of names, used to prevent printing "waiting for
+   lock" messages multiple times when multiple concurrent fibers try
+   to take a lock at the same time. *)
+module Global_waiting_names = struct
+  let state = lazy (String.Table.create 1)
```
rgrinberg, line 14: no point in making it lazy. A lazy cell is about the same size as a table with one element.

src/dune_pkg/flock.ml
```diff
@@ -0,0 +1,148 @@
...
+  let current_dune_pid = Unix.getpid () in
+  let rec loop timeout_s =
+    match Flock.lock_non_block flock Flock.Exclusive with
+    | Error e -> Fiber.return @@ Error e
+    | Ok `Success ->
+      Global_waiting_names.remove name_for_messages;
+      Fiber.return (Ok `Success)
+    | Ok `Failure -> handle_failure timeout_s
+  and handle_failure timeout_s =
+    let locked_by_pid = int_of_string (Io.read_file lock_path) in
```
rgrinberg, line 40: `Int.of_string`.

src/dune_pkg/flock.mli
```diff
@@ -0,0 +1,23 @@
...
+    Within the a dune process, this function also ensures mutual
+    exclusion between fibers. Note that if this function times out
+    waiting for the lock while the lock is held by a different fiber
+    of the same dune process, a [Code_error] is raised rather than a
+    [User_error]. If a timeout is possible, avoid allowing multiple
+    fibers to concurrently attempt to take a flock. *)
+val with_flock
+  :  Path.t
+  -> name_for_messages:string
+  -> timeout_s:float
```
rgrinberg, line 21: Sometimes we use `_s` and other times we use `seconds`. Let's just stick to the latter naming convention.

src/dune_pkg/checksum.mli
gridbugs: Looks like we just use the first checksum for regular packages. Ok I'll change the toolchains logic to also just use the first with the expectation that it will soon be changed to use all the checksums when https://github.com/ocaml/dune/pull/10624 is merged.

src/dune_rules/ocaml_toolchain.mli
Leonidas-from-XIV: Yep, I addressed some of the notes of the review we did with @moyodiallo and @maiste but not all of them.

rgrinberg (approved):
I looked over the key parts in dune_rules and some of the rest. I'll let @Leonidas-from-XIV finish the review.

My only remaining request before merging is to move the toolchain additions in the rules into their own module in dune_rules. We shouldn't automatically default into doing everything in pkg_rules anymore. It made sense in the beginning, but the components of the rules are mature enough that we can afford some modularity.

src/dune_pkg/flock.ml
gridbugs: Good to know!

gridbugs:
Thanks for the review @rgrinberg!

src/dune_rules/pkg_rules.ml
rgrinberg: The value would be that it would make it easier to localize the impact of this feature to one module. Following that, a written interface for it that would give us a complete picture of how this feature interacts with the rules.

Looking at the code, it needs a bit of improvement to extract out such an interface. For example, this excerpt is an implementation detail of this feature and the rules have no business knowing anything about it:

```ocaml
      Memo.of_reproducible_fiber
        (let open Fiber.O in
         let* available_compilers =
           Toolchain.Available_compilers.load_upstream_opam_repo ()
         in
         let* toolchain_compiler =
           Toolchain.Available_compilers.find available_compilers name version ~deps
         in
         match toolchain_compiler with
         | None -> Fiber.return None
         | Some toolchain_compiler ->
           let+ toolchain_compiler =
             Toolchain.Compiler.ensure_installed
               ~log_when:`Install_only
               toolchain_compiler
           in
           Some toolchain_compiler)
```

I can make the changes myself if you aren't sure what I'm trying to get at.

src/dune_rules/pkg_rules.ml
```diff
@@ -193,7 +195,114 @@ module Value_list_env = struct
...
+      Some (Value.Dir path :: paths))
+  ;;
+
+  (* Prepends the bin dir of a given toolchain compiler to the PATH
+     variable in the environment. *)
+  let add_toolchain_bin_dir_to_path t compiler =
+    let toolchain_bin_dir =
+      Toolchain.Compiler.bin_dir compiler |> Path.outside_build_dir
+    in
+    add_path t Env_path.var toolchain_bin_dir
```
rgrinberg, line 207: This is also an implementation detail that should stay hidden from the rules if possible

src/dune_rules/pkg_rules.ml
```diff
@@ -1094,38 +1189,48 @@ end = struct
...
-        { Pkg.id
-        ; build_command
-        ; install_command
-        ; depends
-        ; paths
-        ; info
-        ; files_dir
-        ; exported_env = []
-        }
+      let* compiler =
```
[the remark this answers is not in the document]
rgrinberg: I agree with the comment above. Moreover, this behavior is actually dangerous because it makes the builds we compute from lock directories depend on non workspace related data. In particular, if a toolchain is installed that coincides with the name of a package in any existing lock directory, it will now be treated in a different way. If a package corresponds to a toolchain it should correspond to that in the lock directory.

I retract my approval until we see what can be done to address this issue.

rgrinberg (changes requested):

src/dune_rules/pkg_rules.ml
rgrinberg: I did this change myself. A separate module wasn't necessary

src/dune_rules/pkg_rules.ml
```diff
@@ -1094,38 +1169,47 @@ end = struct
...
+        Compiler.find_and_install_toolchain_compiler
```
rgrinberg, line 1173: Doesn't this mean that we access the opam repo when building the lock directory? That's not allowed, and also it's very slow.

src/dune_rules/pkg_rules.ml
```diff
@@ -1533,6 +1617,7 @@ let source_rules (pkg : Pkg.t) =
 ;;
 
 let build_rule context_name ~source_deps (pkg : Pkg.t) =
+  let* compiler = Compiler.of_context context_name in
```
rgrinberg, line 1620: Some packages do not require the compiler

src/dune_pkg/toolchain.mli
rgrinberg: Sure, feel free to keep the separate functions. However, when integrating two different components, the amount of implementation details visible to each component needs to be minimized. I think the current API does not achieve that goal. If this is because some other API quality was prioritized ahead of that, then I think it's regrettable.

rgrinberg:
Looking over the integration with Pkg_rules again, I pointed out a few flaws I didn't notice earlier. I know you've dismissed this earlier, but I can't help but think your job could have been made much easier by making a toolchain a pseudo package that just installs the toolchain and creates an install cookie file. Otherwise I'm not sure how we can even maintain the same semantics we've had before.

src/dune_rules/pkg_rules.ml
```diff
@@ -1606,7 +1691,10 @@ let build_rule context_name ~source_deps (pkg : Pkg.t) =
   Action_builder.deps deps
   |> Action_builder.with_no_targets
   (* TODO should we add env deps on these? *)
-  >>> add_env (Pkg.exported_env pkg) build_action
+  >>> add_env
+        (Compiler.add_path_to_env compiler ~env:(Pkg.exported_value_env pkg)
+         |> Value_list_env.to_env)
```
rgrinberg, line 1696: Isn't the order wrong? The toolchain shouldn't override the environment of the packages. Its position in the environment should come from where it came in the package closure.

src/dune_rules/pkg_rules.ml
```diff
@@ -1726,8 +1832,11 @@ let which context =
 ;;
 
 let ocamlpath context =
-  let+ all_packages = all_packages context in
-  let env = Pkg.build_env_of_deps all_packages in
+  let* all_packages = all_packages context in
+  let+ env =
+    Compiler.of_context context
+    >>| Compiler.add_path_to_env ~env:(Pkg.build_env_of_deps all_packages)
```
rgrinberg, line 1838: I don't think you need this because the toolchain cannot set OCAMLPATH (Maybe?)

src/dune_rules/pkg_rules.ml
```diff
@@ -1739,8 +1848,11 @@ let lock_dir_active = Lock_dir.lock_dir_active
 let lock_dir_path = Lock_dir.get_path
 
 let exported_env context =
-  let+ all_packages = all_packages context in
-  let env = Pkg.build_env_of_deps all_packages in
+  let* all_packages = all_packages context in
+  let+ env =
+    Compiler.of_context context
+    >>| Compiler.add_path_to_env ~env:(Pkg.build_env_of_deps all_packages)
```
rgrinberg, line 1854: Order is wrong here too. The fact that there are two places where we must keep the order in mind is concerning.

src/dune_rules/ocaml_toolchain.mli
rgrinberg: I moved this entire function to where we setup the toolchain modules in Pkg_rules. Now all the implementation details for the rules live in one place (more or less)

rgrinberg:
The review is a bit scattered, so let me just summarize the state of affairs so that the priorities are clear.

* The interaction with opam repo when building needs to go away. This is critical.
* The remaining bugs can we can make a note of and move on. If we're lucky, the users will never run into them.
* The design questions can be stopped for now. Let's just concentrate on keeping things in one place in dune_rules, or moving the code to Dune_pkg otherwise.

gridbugs:
> Looking over the integration with Pkg_rules again, I pointed out a few flaws I didn't notice earlier. I know you've dismissed this earlier, but I can't help but think your job could have been made much easier by making a toolchain a pseudo package that just installs the toolchain and creates an install cookie file. Otherwise I'm not sure how we can even maintain the same semantics we've had before.

Can you elaborate a bit on this. Are you proposing to represent a toolchain with a regular package (that gets a lockfile like any other package), where dune would recognize the package as representing a toolchain, and install the compiler into the toolchain directory when building the package? That way we could use the existing machinery in pkg_rules for updating paths.

gridbugs:
@rgrinberg I've started experimenting with a more conservative implementation of toolchains based on your suggestion. The only difference is in pkg_rules, where the toolchain is installed as if it was the install command of compiler packages. It's not completely working yet as I'm not providing the new `Ocaml_toolchain.t`. Can you take a look at https://github.com/ocaml/dune/pull/10698/files#diff-ed8949e9f1cea7cfb9eee6f20d456401dc3755949ed4d24268d1c861f24d69b7 and let me know if I'm on the right track please.

rgrinberg:
> Can you elaborate a bit on this. Are you proposing to represent a toolchain with a regular package (that gets a lockfile like any other package), where dune would recognize the package as representing a toolchain, and install the compiler into the toolchain directory when building the package? That way we could use the existing machinery in pkg_rules for updating paths.

Basically, there's nothing special between a toolchain and any regular package apart from how we build it. A regular package is installed by running an action that builds the package in `_build`. A non relocatable package on the other hand, builds it in a global directory. The package rules could work equally well with both types of packages as long as the relevant types are updated to reflect that. In particular, `Paths.t`, `Install_cookie.t`, `Pkg.t` are of interest. Some of these types will need to be updated to allow external paths by generalizing `Path.Build.t` to `Path.t`. Others, you might need to introduce explicit variants to special case things.

> Can you take a look at [#10698 (files)](https://github.com/ocaml/dune/pull/10698/files#diff-ed8949e9f1cea7cfb9eee6f20d456401dc3755949ed4d24268d1c861f24d69b7) and let me know if I'm on the right track please.

Yeah, that seems to be on the right track:

Fetching a non relocatable package should work in the same way. Or at least I don't see why should it be any different. So that work seems unnecessary.

The meat of the work is implementing a custom build/install action that is going to reuse your custom building/caching logic instead of reusing dune's. A few things to keep in mind:

1. For modifying the build/install actions, instead of adding an "is_compiler" check. Which as I mentioned earlier, is an imprecise way of naming a non-relocatable package, you could expand `Build_command.t` (the install command will need similar treatment) to:
```
  type t =
    | Action of Action.t
    | Dune
    | Non_relocatable of Action.t (* you might want to pick something simpler than Action.t as the argument *)
```

All that should remain is implementing a custom action to run your build/install commands using your existing code

2. Note that you will still need to produce the install cookie file in `_build`, but this cookie file should point to your external paths.

3. You will also need to copy the source paths to your global directory manually as well. As dune will not understand that you need them outside the build directory.

4. Finally, if your action encounters a compiler that has already been built for the configuration that you're looking for, it should do no work and just produce a cookie pointing to the existing install. A good way to do this is to maintain a cache of non relocatable packages keyed by `md5 (source_checksum, action_digest)` in your state directory. But something simpler will work too if don't care about caching reliably.

6. You should disable sandboxing for the build/install rule. Those don't make sense for non-relocatable packages. When another package depends on a non-relocatable package, we exploit the fact that external paths `Path.External.t` never get copied to the sandbox even if they are dependencies. Otherwise, we'd need to disable sandboxing everywhere.

To do all this, I suggest you start by defining the simplest non relocatable package:

```
(version 0.0.1)
(relocatable false)
;; when you depend on this package and print `%{lib}/foo`, you should see a path pointing to your global dir
(install (run bash -i "mkdir -p %{lib}; echo $PWD > %{lib}/foo"))
```

If you can make it work for the above, you should be able to make it work for the complier.

gridbugs:
Hey thanks for the detailed suggestions.

I want to share a bit of context from inside the build system team. Previously we had a long-lived branch that includes the toolchains feature, and the goal of upstreaming this feature was so we no longer have to keep it up to date with main by periodically merging main into our long-lived branch. It's a temporary (🤞) workaround to the compiler's relocation issue, and we just needed this feature to unblock some developer interviews evaluating other more user-visible features, so we prioritized implementing it quickly.

We (the build system team) don't want to spend much more time upstreaming this feature. We're going to start using the long-lived toolchains branch again internally (including changes from this review) for our new feature work rather than focusing on upstreaming toolchains. I don't want anyone on this review to feel like it was a waste of time. I'm very grateful for all the feedback and improvements made to the branch!

I'm still interested in seeing if we can upstream a version of this feature, though given it's a temporary (🤞) workaround and I'm planning to spend most of my time on new feature work, I can't justify implementing the solution you're proposing @rgrinberg. My plan is to instead:
 - remove the interaction with the opam repo from toolchains and by using the source urls and hashes from the compiler package lockfile rather than looking them up in the opam repo
 - install the compiler into the toolchains dir as if doing so was a regular package install command (ie. just from the `install_command` function in pkg_rules) rather than ensuring that it's installed in other parts of pkg_rules
 - add the compiler bin directory to PATH that same way as we update PATH for regular packages

The general idea would be to make this change more conservative by reducing its footprint on the pkg_rules module. Of course, all of this functionality will be disabled with a feature flag anyway, but making it more conservative will still help make sure we didn't subtly break something and also keep the contract between pkg_rules and the rest of dune from getting too complex.

rgrinberg:
Sounds fine with me. If you wait long enough, maybe relocation will arrive and you will never need to implement this feature properly anyway. I'd say if this feature maintains the "illusion" that the compiler is first class. To maintain the illusion, I think this requires:

1. No interaction with opam repo during build/install time
2. Allowing users to pin the compiler

I think this is good enough for alphas and betas.

gridbugs:
Thanks for the feedback everyone. Closing this in favour of a new PR that re-implements this functionality based on the feedback. https://github.com/ocaml/dune/pull/10719.

PR #10641, opened 2024-06-11
Implicit relock if the dependencies have changed

gridbugs:
If lockfiles exist and these are out of date, this PR runs the solver again to re-create the lock directory.

This is a copy of @Leonidas-from-XIV's PR https://github.com/ocaml/dune/pull/10546, but targeting main rather than the toolchains branch. I've changed it to use the same feature flags module introduced in https://github.com/ocaml/dune/pull/10639.

### DIFF
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -47,7 +47,48 @@ let run_build_system ~common ~request =
       Cached_digest.invalidate_cached_timestamps ();
       let* setup = Import.Main.setup () in
       let request =
-        Action_builder.bind (Action_builder.of_memo setup) ~f:(fun setup -> request setup)
+        let open Action_builder.O in
+        let autorelock =
+          match Dune_pkg.Feature_flags.use_autorelock with
+          | false -> Memo.return ()
+          | true ->
+            Memo.of_thunk (fun () ->
+              let open Memo.O in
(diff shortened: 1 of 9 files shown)


bin/build_cmd.ml
```diff
@@ -47,7 +47,48 @@ let run_build_system ~common ~request =
       Cached_digest.invalidate_cached_timestamps ();
       let* setup = Import.Main.setup () in
       let request =
-        Action_builder.bind (Action_builder.of_memo setup) ~f:(fun setup -> request setup)
+        let open Action_builder.O in
```
rgrinberg, line 50: This implementation should be factored more cleanly. We'd like the experimental code not to overlap with stable code.

rgrinberg:
There's some problems with this implementation from what I can tell:

1. It adds the overhead of reading the lock directory and checking if it's up to date on every single iteration of the watch mode. Has this overhead been measured?

2. Auto re-locking does not get triggered when local opam repositories are updated.

3. It does not use a lock so a user manually running `dune pkg lock` can race against the auto lock.

Apart from that, I have some questions:

1. How does one force dune to update remote opam repositories before running the auto lock?

2. What happens when one doing is multiple updates to project files in succession? Does it mean this triggers multiple auto locks that all get staggered?

maiste:
This PR can be closed as it was done in but in a different way in #10851.

PR #10642, opened 2024-06-11
pkg: respect avoid-versions between versions of pkg

gridbugs:
Dune vendors the opam-0install-solver package and uses it to solve package dependencies, however opam-0install-solver does not respect the avoid-versions flag of opam packages. This can lead to situations where unstable releases of packages (notably of the compiler packages) are chosen in favor of the latest stable release of a package.

This change implements a partial workaround for this limitation where the solver will prefer versions of packages that lack the avoid-version flag.

Note that this does not fully implement the avoid-versions flag. If a package depends on a disjunction of different packages then the avoid-versions flag is not considered when choosing which package in the disjunction to choose.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -32,6 +32,10 @@ let add_self_to_filter_env package env variable =
     else env variable
 ;;
 
+let opam_file_is_avoid_version (opam_file : OpamFile.OPAM.t) =
+  List.mem opam_file.flags Pkgflag_AvoidVersion ~equal:Poly.equal
+;;
+
 module Context_for_dune = struct
   type 'a monad = 'a Monad.t
   type filter = OpamTypes.filter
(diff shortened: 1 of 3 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -109,18 +113,33 @@ module Context_for_dune = struct
...
+      (* If both packages have the same setting of their avoid-version
+         flag then compare their versions. *)
+      let ordering =
+        opam_package_version_compare (OpamFile.OPAM.version a) (OpamFile.OPAM.version b)
+      in
+      (match t.version_preference with
+       | Oldest -> ordering
+       | Newest -> Ordering.opposite ordering)
+    | false, true -> Lt
+    | true, false -> Gt
```
rgrinberg, line 142: Is this the same as comparing `(opam_file_is_avoid_version, OpamPackage.Version.t)`? If it is, then `Tuple.T2.compare` would make this more readable.

rgrinberg (approved):
LGTM - this is much clearer. One minor note about combining the comparison to avoid the pattern matching on Ordering (which is easy to get wrong)

src/dune_pkg/opam_solver.ml
Leonidas-from-XIV: I've pushed onto the PR with this suggestion, it does make it a bit nicer.

PR #10643, opened 2024-06-11
Update changelog for 3.16.0~alpha2

gridbugs:


CHANGES.md
```diff
@@ -66,6 +66,9 @@
 - virtual libraries: fix an issue where linking an executable involving several
   virtual libries would cause an error. (#10581, fixes #10460, @rgrinberg)
 
+- Correctly declare dependencies of indexes so that they are rebuilt when
+  needed. (#10623, @voodoos)
+
```
emillon, line 71: this is a bit messy, but this is an entry of alpha2, not alpha1. we'll collapse these at release time (and this entry will be removed since it's not relevant to released versions)

gridbugs: ah, didn't realize that we keep them separate. Fixed.

emillon (approved):

PR #10644, opened 2024-06-11
Add support for -H compiler flag

### LINKED ISSUE #2733: Debugging build failures with implicit_transitive_deps false
I'm wondering if there is standard advice on debugging build failures introduced by setting `implicit_transitive_deps false`? For example, I added that config to a project that includes a ppx rewriter and see:
```
Error: The module Selected_ast is an alias for module Ppxlib_ast__.Import.Selected_ast, which is missing
```
I suppose that by digging through the metadata files for ppxlib I would eventually find some library name to add to the stanza:
```
(library
 (name ppx_trace)
 (kind ppx_rewriter)
 (preprocess no_preprocessing)
 (libraries ppxlib))
```
But is there a better message that dune can give, since without disabling `implicit_transitive_deps` the build succeeds?

### LINKED ISSUE #4963: ignore ppx runtime dependencies when implicit transitive deps = false
## Desired Behavior

When `(implicit_transitive_deps false)` is specified, it would be good if dependencies required by ppx rewriters were excluded from it. For example, `ppx_deriving.runtime` should be left implicit even in this case, because in all likelihood it's only depended upon by automatically generated code, for which the dependency is already explicit in `(pps …)`.

related: #2733

### LINKED ISSUE #9333: Add support for the new -H <dir> argument in OCaml 5.2

nojb:
This PR adds support to Dune to use the `-H` compiler flag introduced in https://github.com/ocaml/ocaml/pull/12246, which provides good semantics to the `(implicit_transitive_deps false)`.

This work was carried out by @MA0010 as part of an internship at LexiFi. I am opening this PR but @MA0010 will take the lead during the review process.

- The first commit contains the meat of the change: we keep track of which dependencies are "direct" (= mentioned in the Dune file) and which dependencies are purely transitive. For direct dependencies we use `-I` and for transitive dependencies we use `-H` (of course, only for OCaml >= 5.2). The use of `-H` is also guarded by Dune version 3.17 (this is to minimize the possibility of breakage, also there are some corner cases where the change is not completely backwards-compatible). 

- The second commit adds an explicit `unix` dependency to some vendored libraries. This is needed because `unix` is mentioned in the source of these libraries, but the library is not mentioned in the Dune stanzas. The libraries are able to build before this PR because OCaml adds `-I +unix` implicitly (for recent enough versions of OCaml where `unix` is installed in the separate directory `+unix`). However, after this PR, the compiler passes `-H +unix` which disables the implicit addition of `-I +unix` breaking compilation (again, because `Unix` __is__ a direct dependency of these libraries).

- The third commit contains some tests. Writing the tests was a bit involved due to the need to have them pass both with versions of OCaml supporting `-H` and those that do not. Suggestions for improvement welcome.

- The last commit bumps the Dune version to 3.17. We could consider not bumping the Dune language version since the new mode is almost always a strict improvement for users of `(implicit_transitive_deps false)` (modulo corner cases such as the one mentioned in the second bullet point). Opinions welcome.

Future work:

- Adapt the Merlin machinery to pass `-H` as needed (there is already an open PR for this: #10535).
- Maybe adapt the `ocaml-index` machinery accordingly (cc @voodoos)

Fixes #9333 
Fixes ocsigen/tyxml#274 
Fixes #2733
Fixes #4963


emillon:
Hi, thanks for this PR and the detailed description. There are a couple things we can land first:
- the 3.17 language version (we'll need it anyway and 3.16 has branched)
- the fix for unix is useful anyway because it triggers a warning on 5.x and we're going to bump the test suite to use this version soon


Regarding your question:
> Writing the tests was a bit involved due to the need to have them pass both with versions of OCaml supporting -H and those that do not. Suggestions for improvement welcome.

Yes. the reason it's painful is that we're only testing a single version of the compiler in CI, but you can add tests that require 5.x and mark them with `(enabled_if (>= %{ocaml_version} 5.2))`. This means that `make test` will work for them locally, and will work on 5.x in CI when we bump the CI version. Similarly for tests that rely on 4.x. They'll become dead tests when we bump the tested version but that can still be useful for some time.

MA0010:
Hello, thanks for the comment.

> -   the 3.17 language version (we'll need it anyway and 3.16 has branched)
> - the fix for unix is useful anyway because it triggers a warning on 5.x and we're going to bump the test suite to use this version soon.

I have opened a separate PR to land the changes you proposed (see #10645). As you mentioned, those can be merged independently of the feature to be added through this PR. I will remove them from this PR as soon as they get merged successfully.

src/dune_rules/lib.mli
```diff
@@ -235,3 +235,5 @@ module Local : sig
 
   include Comparable_intf.S with type key := t
 end
+
+module Tbl : Hashtbl.S with type key := t
```
rgrinberg, line 239: I don't think you need this since our `Table` module is polymorphic. `let table = Table.create (module Lib)` should work

src/dune_rules/lib_flags.mli
```diff
@@ -24,7 +24,14 @@ module L : sig
 
   val to_iflags : Path.Set.t -> _ Command.Args.t
   val include_paths : ?project:Dune_project.t -> t -> Lib_mode.t -> Path.Set.t
-  val include_flags : ?project:Dune_project.t -> t -> Lib_mode.t -> _ Command.Args.t
+
+  val include_flags
+    :  ?project:Dune_project.t
+    -> ?direct:(Lib.t -> bool)
```
rgrinberg, line 30: Why can't we just pass a list: `([ Direct | Indirect] * Lib.t) list` as an argument? I'm not a fan of passing around callbacks.

src/dune_rules/compilation_context.ml
```diff
@@ -139,12 +141,24 @@ let create
...
+  let context = Super_context.context super_context in
+  let sandbox = Sandbox_config.no_special_requirements in
+  let* ocaml = Context.ocaml context in
+  let requires, direct =
     if Dune_project.implicit_transitive_deps project
-    then Memo.Lazy.force requires_link
-    else requires_compile
+    then Memo.Lazy.force requires_link, Resolve.Memo.return (fun _ -> true)
+    else if Version.supports_hidden_includes ocaml.version
+            && Dune_project.dune_version project >= (3, 17)
```
rgrinberg, line 151: I think this mode still needs to be guarded behind `implicit_transitive_deps`. Once we experiment with it more, we can enable it by default for newer compilers perhaps.

rgrinberg:
@voodoos could you review the implementation as well? I think you needed something similar for merlin. It would be good to see what can be shared.

src/dune_rules/compilation_context.ml
MA0010: Sorry, but I do not seem to have understood well your point. The new feature is invoked only when `implicit_transitive_deps` is false, according to the written logic, or am I missing something?

Thanks for the review.

voodoos (commented):
I think the way dependencies are exposed right now is too restricting. The `Lib.Compile.t` API only provide us with either `require_compile`, the list of direct dependencies, or  `require_link`, the list of all transitive dependencies  including the direct ones.

This means that we have to compute the disjunction `require_link \ require_compile` to get the actual "hidden" or "extra" dependencies. Right now this is needed in 3 places: rules for indexation (#10422), rules for merlin config (#10535) and this PR.

One issue is that `require_compile` and `require_link` are lists which prevent an efficient computation of the disjunction. In #10422 and #10535 this is done in a somewhat inefficient way using multiple calls to `List.mem`. This PR introduces a more efficient approach by using a set (the (lib, unit) Hashtbl.t) to perform more efficient membership tests.

I think we should update `Lib.Compile.t` to either provide an additional `requires_extra` list of libs that is computed in an efficient way and share the result, or have `requires_compile` and `requires_link` be Sets so that consumers can efficiently compute the disjunction without too much code duplication

This would allow the different places using the "extra deps" to do it in a uniform way.

src/dune_rules/lib_flags.mli
MA0010: Hello,
I have removed the passed callback. Instead, now two lists are passed, one containing the direct dependencies (`requires_direct`) and the other contains the hidden dependencies (`requires_hidden`). The `requires_hidden` is also exposed in `Compilation_context.t`, and so can be used in other places when needed. 

The code have changed a little bit, and the diff in `lib_flags.ml` is now smaller.

Review appreciated!

MA0010:
Concerning the last two commits, they contain edits based on the reviews and comments (cc @voodoos @rgrinberg), resting the main goal of keeping track of hidden deps unchanged. Those dependencies are now tracked by a new list `hidden_requires` exposed in the `Compilation_context.t` type. In short:

- The first of the new commits keeps the previous approach of using hashtables to compute the hidden dependencies, but now instead of being passed by a callback, they are stored in list `requires_hidden` and exposed in `Compilation_context.t` for external use cc (@voodoos). This list is then used to insert the `-H` flag for the correct deps.

- The second one simply removes the custom hashtable module from `lib.ml`, in favor of the already defined `Table` module (@rgrinberg). Also, the `ocaml_index.ml` file is adapted to use the new `requires_hidden` list, however in a naive way, and maybe can be used to adapt the machinery better (cc @voodoos )

src/dune_rules/lib_flags.ml
```diff
@@ -155,8 +162,16 @@ module L = struct
...
+  let include_flags ?project ts_direct ts_hidden mode =
+    let hidden_includes =
+      to_hflags
+        (include_paths ?project ts_hidden { lib_mode = mode; melange_emit = false })
+    in
+    let direct_includes =
+      to_iflags
+        (include_paths ?project ts_direct { lib_mode = mode; melange_emit = false })
+    in
+    Command.Args.S [ direct_includes; hidden_includes ]
```
rgrinberg, line 174: Isn't this re-ordering the include flags? e.g. if we have `(libraries x y)`, we'd have `closure(x) closure(y)`, and now we have `x y closure(x) closure(y)`?

nojb: If I understand things correctly, I don't think anything is being re-ordered; instead some extra `-H` flags are being added at the "end" of the command-line when `implicit_transitive_deps` is `false`:

- If `implicit_transitive_deps` is `true`, then `direct_includes` is `requires_link` (from `Compilation_context`), and `hidden_includes` is empty; this is the same as today, and nothing changes.
- If `implicit_transitive_deps` is `false` (and OCaml is recent enough, etc) `direct_includes` is `requires_compile` (from `Compilation_context`), and `hidden_includes` contains some extra flags relative to today, but the order of `direct_includes` does not change.

Do you agree with this understanding @MA0010?

src/dune_rules/lib_flags.ml
```diff
@@ -155,8 +162,16 @@ module L = struct
...
-    to_iflags (include_paths ?project ts { lib_mode = mode; melange_emit = false })
+  let include_flags ?project ts_direct ts_hidden mode =
+    let hidden_includes =
+      to_hflags
+        (include_paths ?project ts_hidden { lib_mode = mode; melange_emit = false })
+    in
+    let direct_includes =
+      to_iflags
+        (include_paths ?project ts_direct { lib_mode = mode; melange_emit = false })
+    in
```
nojb, line 173: Let's share the code for the two definitions.

src/dune_rules/lib_flags.ml
```diff
@@ -92,6 +92,13 @@ module L = struct
        |> List.rev)
   ;;
 
+  let to_hflags dirs =
+    Command.Args.S
+      (Path.Set.fold dirs ~init:[] ~f:(fun dir acc ->
+         Command.Args.Path dir :: A "-H" :: acc)
+       |> List.rev)
+  ;;
+
```
nojb, line 101: Let's share the code with `to_iflags` above.

src/dune_rules/compilation_context.ml
```diff
@@ -3,36 +3,51 @@ open Import
 module Includes = struct
   type t = Command.Args.without_targets Command.Args.t Lib_mode.Cm_kind.Map.t
 
-  let make ~project ~opaque ~requires : _ Lib_mode.Cm_kind.Map.t =
+  let make
+    ~project
+    ~opaque
+    ~direct_requires
+    ~(hidden_requires : Lib_flags.L.t Resolve.Memo.t)
```
nojb, line 10:
```suggestion
    ~hidden_requires
```

src/dune_rules/lib_flags.ml
```diff
@@ -85,10 +85,10 @@ let link_deps sctx t mode =
 module L = struct
   type nonrec t = Lib.t list
 
-  let to_iflags dirs =
+  let to_iflags ?(flag = "-I") dirs =
```
Leonidas-from-XIV, line 88: In our codebase we often try to avoid default flags, as the defaults are often easy to miss and in this case "to_iflags" which generates `-H` flags is somewhat strange.

As such I suggest making this a labelled `to_flag` function and 
```ocaml
let to_ifags = to_flags "-I"
let to_hflags = to_flags "-H"
```

This prevents having to change all the code while allowing an easy way to get an `-H` flag.

Leonidas-from-XIV: You might not even need to expose `to_hflags` in the `mli` if nobody outside the module is going to use it. That way the compiler can warn about unused functions in case the code stops using `to_hflags`.

MA0010: You are right for now `to_hflags` is not used anywhere outside `lib_flags.mli`, but I exposed it for the sake of symmetry (since if `to_iflags` is to be used somewhere in some later work, `to_hflags` would be also needed).

But maybe it is better for now to remove it from the interface.

src/dune_rules/compilation_context.ml
```diff
@@ -139,12 +150,26 @@ let create
...
+        let+ requires = requires_compile
+        and+ requires_link = Memo.Lazy.force requires_link in
+        let requires_table = Table.create (module Lib) 500 in
+        let () = List.iter ~f:(fun lib -> Table.set requires_table lib ()) requires in
+        List.filter requires_link ~f:(fun l -> not (Table.mem requires_table l))
+      in
+      requires_compile, requires_hidden)
+    else requires_compile, Resolve.Memo.return []
   in
-  let sandbox = Sandbox_config.no_special_requirements in
```
nojb, line 147: Is this change needed?

src/dune_rules/compilation_context.ml
```diff
@@ -139,12 +150,26 @@ let create
...
-    else requires_compile
+    then Memo.Lazy.force requires_link, Resolve.Memo.return []
+    else if Version.supports_hidden_includes ocaml.version
+            && Dune_project.dune_version project >= (3, 17)
+    then (
+      let requires_hidden =
+        let open Resolve.Memo.O in
+        let+ requires = requires_compile
+        and+ requires_link = Memo.Lazy.force requires_link in
+        let requires_table = Table.create (module Lib) 500 in
```
nojb, line 166: `500` seems too large; I suggest `5`.

doc/changes/10644.md
```diff
@@ -0,0 +1,6 @@
+- Add support for the -H flag (introduced in Ocaml compiler 5.2) in dune
+  (only supported for lang versions 3.17 and higher). This  adaptation gives
+  the correct semantics for `implicit_transitive_deps = false` mode, since then
+  transitive dependencies are included with -H flag and so are not directly accessible
+  but visible for the compiler. (#10644, fixes (#9333, #ocsigen/tyxml#274, #2733, #4963),
+  @MA0100)
```
nojb, line 6:
```suggestion
- Add support for the -H flag (introduced in OCaml compiler 5.2) in dune
  (only supported for lang versions 3.17 and higher). This adaptation gives
  the correct semantics for `implicit_transitive_deps = false` mode.
  (#10644, fixes #9333, ocsigen/tyxml#274, #2733, #4963, @MA0100)
```

src/dune_rules/merlin/ocaml_index.ml
```diff
@@ -32,11 +32,9 @@ let cctx_rules cctx =
 
            The implicit transitive libs correspond to the set:
            (requires_link \ req_compile) *)
-        let* req_link = Compilation_context.requires_link cctx in
-        let+ req_compile = Compilation_context.requires_compile cctx in
-        List.filter req_link ~f:(fun l -> not (List.exists req_compile ~f:(Lib.equal l)))
+        Compilation_context.requires_hidden cctx
```
nojb, line 35: This doesn't seem quite right. For new versions of OCaml (>= 5.2) and version of the Dune language >= 3.17, Dune will pass `-H` to the compiler and these flags will be recorded in the `.cmt` files and will, in turn, be read by `ocaml-index`. Accordingly, in these cases we do not need to do anything special here. It is only for backwards compatibility that we can keep the above logic (ie if OCaml is < 5.2 _or_ if the Dune language veresion if < 3.17). Also, the flags should all be passed with `-I`, not `-H` in this case.

nojb: @voodoos sounds OK?

nojb:
Friendly ping @anmonteiro: are you planning to review the PR? I feel this is getting ready to be merged, so do not hesitate to leave a note if you are planning to do so (which would be appreciated, of course!).

src/dune_rules/merlin/ocaml_index.ml
voodoos: Yes, that's completely right. If Dune >= 3.17 there is no more additional flags to pass to the indexer. (Note that indexing is not available prior to OCaml 5.2, so there is no need to check that here.)

nojb: Thanks for the confirmation! So only the check for OCaml lang version < 3.17 is needed.

MA0010: Done!

src/dune_rules/merlin/ocaml_index.ml
```diff
@@ -24,19 +24,21 @@ let cctx_rules cctx =
...
+          (* This is only needed for the case where OCaml < 5.2 or the dune language version
+             < 3.17, since in the latter cases those additional libs will be already included in the compilation
+             with -H flag.*)
+          let* req_link = Compilation_context.requires_link cctx in
+          let+ req_compile = Compilation_context.requires_compile cctx in
+          List.filter req_link ~f:(fun l ->
+            not (List.exists req_compile ~f:(Lib.equal l)))
+        in
+        Lib_flags.L.include_flags non_compile_libs [] (Lib_mode.Ocaml Byte)
+      else Resolve.Memo.return Command.Args.empty
```
nojb, line 41: I suggest tweaking the placement and wording of the comment as follows:
```suggestion
      let scope = Compilation_context.scope cctx in
      (* Dune language >= 3.17 correctly passes the `-H` flag to the compiler. *)
      if Dune_project.dune_version (Scope.project scope) < (3, 17)
      then
        let open Resolve.Memo.O in
        let+ non_compile_libs =
          let* req_link = Compilation_context.requires_link cctx in
          let+ req_compile = Compilation_context.requires_compile cctx in
          List.filter req_link ~f:(fun l ->
            not (List.exists req_compile ~f:(Lib.equal l)))
        in
        Lib_flags.L.include_flags non_compile_libs [] (Lib_mode.Ocaml Byte)
      else Resolve.Memo.return Command.Args.empty
```

doc/changes/10644.md
```diff
@@ -0,0 +1,4 @@
+- Add support for the -H flag (introduced in OCaml compiler 5.2) in dune
+  (only supported for lang versions 3.17 and higher). This adaptation gives
```
nojb, line 2:
```suggestion
  (requires lang version 3.17). This adaptation gives
```

src/dune_rules/compilation_context.ml
```diff
@@ -139,10 +150,24 @@ let create
...
+    else if Version.supports_hidden_includes ocaml.version
+            && Dune_project.dune_version project >= (3, 17)
+    then (
+      let requires_hidden =
+        let open Resolve.Memo.O in
+        let+ requires = requires_compile
+        and+ requires_link = Memo.Lazy.force requires_link in
+        let requires_table = Table.create (module Lib) 5 in
+        let () = List.iter ~f:(fun lib -> Table.set requires_table lib ()) requires in
+        List.filter requires_link ~f:(fun l -> not (Table.mem requires_table l))
```
nojb, line 167:
```suggestion
        let+ requires_compile = requires_compile
        and+ requires_link = Memo.Lazy.force requires_link in
        let requires_table = Table.create (module Lib) 5 in
        let () = List.iter ~f:(fun lib -> Table.set requires_table lib ()) requires_compile in
        List.filter requires_link ~f:(fun l -> not (Table.mem requires_table l))
```

nojb (commented):
LGTM, modulo a couple of suggestions.

I would prefer a second official approval before signing this off.

src/dune_rules/compilation_context.ml
```diff
@@ -3,36 +3,45 @@ open Import
...
               ; Hidden_deps
                   (if opaque
                    then
-                     List.map libs ~f:(fun lib ->
-                       ( lib
-                       , if Lib.is_local lib
-                         then [ Lib_file_deps.Group.Ocaml Cmi ]
-                         else [ Ocaml Cmi; Ocaml Cmx ] ))
+                     List.map
+                       (List.concat [ direct_libs; hidden_libs ])
```
emillon, line 35:
```suggestion
                       (direct_libs @ hidden_libs)
```

src/dune_rules/compilation_context.ml
```diff
@@ -3,36 +3,45 @@ open Import
...
+                       ~f:(fun lib ->
+                         ( lib
+                         , if Lib.is_local lib
+                           then [ Lib_file_deps.Group.Ocaml Cmi ]
+                           else [ Ocaml Cmi; Ocaml Cmx ] ))
                      |> Lib_file_deps.deps_with_exts
                    else
                      Lib_file_deps.deps
-                       libs
+                       (List.concat [ direct_libs; hidden_libs ])
```
emillon, line 44:
```suggestion
                       (direct_libs @ hidden_libs)
```

src/dune_rules/lib_flags.mli
```diff
@@ -24,7 +24,7 @@ module L : sig
 
   val to_iflags : Path.Set.t -> _ Command.Args.t
   val include_paths : ?project:Dune_project.t -> t -> Lib_mode.t -> Path.Set.t
-  val include_flags : ?project:Dune_project.t -> t -> Lib_mode.t -> _ Command.Args.t
+  val include_flags : ?project:Dune_project.t -> t -> t -> Lib_mode.t -> _ Command.Args.t
```
emillon, line 27: Since it takes 2 arguments of type `t`, I think it'd be useful to label these.

test/blackbox-tests/test-cases/hidden-deps.t/bar.ml
```diff
@@ -0,0 +1,3 @@
+let x = 5
+
+let y = Foo.v
```
emillon, line 3: the newline at eof in this test (and others) need fixing

test/blackbox-tests/test-cases/hidden-deps.t/run.t
```diff
@@ -0,0 +1,36 @@
+Should include Foo with -H:
+
+  $ getincludes () {
+  > dune build --verbose ./run.exe 2>&1 | grep run.ml | grep -E -o "\-$1\s(.foo)\S*" | sed s/\-$1//g | tr -d '[:space:]'
+  > }
+
+  $ supports_H=$(if ocamlc -H x --help >/dev/null 2>&1; then echo true; else echo false; fi)
```
emillon, line 7: I would recommend doing that in 2 tests:
- one that assumes ocaml >= 5.2 (guarded with `(enabled_if (>= %{ocaml_version} 5.2))`)
- one that assumes ocaml < 5.2 (similarly guarded)

And then you can use "normal" cram tests, which would consist in just running `getincludes` and displaying the output.

test/blackbox-tests/test-cases/hidden-deps.t/tyxml/run.t
```diff
@@ -0,0 +1,5 @@
+Test if the issue #274 is solved: the tyxml.functor is included with -H flag in
+ITD = false case, and thus no type abstraction happens when it is used.
+
+  $ ocamlc -H x --help > /dev/null
```
emillon, line 4: same here, you can guard the test behind `(enabled_if)`.

emillon (commented):
Thanks. It seems you're mostly there so I left comments on style and tests. You might also want to add a quick note in `implicit_transitive_deps.rst` to explain that on ocaml 5.2 it's implemented using `-H`.

src/dune_rules/compilation_context.ml
rgrinberg: Yeah, that wasn't very clear at all. Let me explain the behavior I would like to achieve:

When we're using `(implicit_transitive_deps false)` and OCaml >= 5.2.0, we should use use your new implementation with the `-H` flag.

When we're using `(implicit_transitive_deps false)` and OCaml < 5.2.0, we should go back to using the old implementation of hiding `-I` flags.

When we're using `(implicit_transitive_deps true)`, it should disable using `-H` flags completely.

The issue with the current behavior is that users do expect upgrading to 3.17 to be relatively easy. With this change, they will now need to look at all their transitive deps. While I agree that it's a good thing, it's too much of a breakage to introduce in a minor version bump.

Therefore, I suggest that we guard this behind a feature flag. We already have a feature flag though `implicit_transitive_deps`, so we can just reuse it.

rgrinberg (changes requested):
As stated, I think this change is a bit too aggressive to be enabled by default at the moment.

nojb:
> Let me explain the behavior I would like to achieve:

@rgrinberg Unless I am missing something, I think this is already the implemented behaviour: `-H` is only used when `implicit_transtiive_deps` is `false` (and OCaml >= 5.2); see condition on line 156 in the file in question:

https://github.com/ocaml/dune/blob/8e1757ffba6f5fff3584fbda94a3e4a7f478bf60/src/dune_rules/compilation_context.ml#L156-L157

test/blackbox-tests/test-cases/hidden-deps.t/run.t
MA0010: Hello, thanks for the review.
I am familiar with gaurding executables and libraries with the `enabled_if` (essentially making them hidden to the test). However, this would not be very useful here.

I have tried to add the `enabled_if` to a `cram` stanza in the dune file, hoping that would disable the whole test. This seems to not work since when I try to build the test it still runs irrespective of the condition mentioned with `enabled_if`.

Do you have any suggestions on how one would use `enabled_if` in order to disable the whole test?

emillon: sure! you can have a look at this for example:

https://github.com/ocaml/dune/blob/3.16.0/test/blackbox-tests/test-cases/dune#L103-L107

MA0010:
I have completed the suggested edits. However, I request a final look at the tests before merging. For now, I have separated the `hidden-deps` tests into two tests, each one guarded by the correct OCaml version. However, this causes code duplication, and there may be a way to guard individual components of the same test behind the OCaml version.

src/dune_rules/compilation_context.ml
nojb: @MA0010 this can be committed, right?

src/dune_rules/compilation_context.ml
```diff
@@ -139,10 +147,26 @@ let create
...
+            && Dune_project.dune_version project >= (3, 17)
+    then (
+      let requires_hidden =
+        let open Resolve.Memo.O in
+        let+ requires_compile = requires_compile
+        and+ requires_link = Memo.Lazy.force requires_link in
+        let requires_table = Table.create (module Lib) 5 in
+        let () =
+          List.iter ~f:(fun lib -> Table.set requires_table lib ()) requires_compile
+        in
```
nojb, line 165:
```suggestion
        List.iter ~f:(fun lib -> Table.set requires_table lib ()) requires_compile;
```

test/blackbox-tests/test-cases/hidden-deps-supported.t/tyxml/run.ml
```diff
@@ -0,0 +1,2 @@
+open Tyxml.Html
+let _ = p [ a [ txt "a" ] ]
```
nojb, line 2: Missing newline at EOF.

test/blackbox-tests/test-cases/hidden-deps-supported.t/tyxml/dune
```diff
@@ -0,0 +1,3 @@
+(executable
+  (name run)
+  (libraries tyxml))
```
nojb, line 3: Missing newline at EOF.

test/blackbox-tests/test-cases/hidden-deps-supported.t/run.ml
```diff
@@ -0,0 +1 @@
+let _ = Bar.y
```
nojb, line 1: Missing newline at EOF.

test/blackbox-tests/test-cases/hidden-deps-supported.t/run.t
```diff
@@ -0,0 +1,45 @@
...
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.17)
+  > (implicit_transitive_deps false)
+  > EOF
+
+  $ echo "$(getincludes H)"
+  .foo.objs/byte.foo.objs/byte.foo.objs/native
+
+
```
nojb, line 24:
```suggestion
```

test/blackbox-tests/test-cases/hidden-deps-supported.t/run.t
```diff
@@ -0,0 +1,45 @@
...
+  > dune build --verbose ./run.exe 2>&1 | grep run.ml | grep -E -o "\-$1\s(.foo)\S*" | sed s/\-$1//g | tr -d '[:space:]'
+  > }
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.17)
+  > (implicit_transitive_deps true)
+  > EOF
+  $ echo "$(getincludes I)"
+  .foo.objs/byte.foo.objs/byte.foo.objs/native
+
```
nojb, line 14:
```suggestion
```

test/blackbox-tests/test-cases/hidden-deps-supported.t/run.t
```diff
@@ -0,0 +1,45 @@
...
+                      ^^^^^
+  Error: Unbound module Foo
+
+Test if the issue #274 is solved: the tyxml.functor is included with -H flag in
+ITD = false case, and thus no type abstraction happens when it is used.
+
+  $ dune build --root=./tyxml
+  Entering directory 'tyxml'
+  Leaving directory 'tyxml'
+
```
nojb, line 44:
```suggestion
```

test/blackbox-tests/test-cases/hidden-deps-unsupported.t/run.t
```diff
@@ -0,0 +1,25 @@
...
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.17)
+  > (implicit_transitive_deps false)
+  > EOF
+
+  $ echo "$(getincludes I)"
+  
+
+
```
nojb, line 25:
```suggestion

```

test/blackbox-tests/test-cases/hidden-deps-unsupported.t/run.t
```diff
@@ -0,0 +1,25 @@
...
+  > dune build --verbose ./run.exe 2>&1 | grep run.ml | grep -E -o "\-$1\s(.foo)\S*" | sed s/\-$1//g | tr -d '[:space:]'
+  > }
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.17)
+  > (implicit_transitive_deps true)
+  > EOF
+  $ echo "$(getincludes I)"
+  .foo.objs/byte.foo.objs/byte.foo.objs/native
+
```
nojb, line 15:
```suggestion
```

nojb:
I think all issues have been addressed and the PR is ready to go.

@emillon: would you like to do the honours? (as I have been supervising the PR, I prefer not to merge it myself).

nojb:
> I think all issues have been addressed and the PR is ready to go.
> 
> @emillon: would you like to do the honours? (as I have been supervising the PR, I prefer not to merge it myself).

Tha PR should be squash-merged by the way.

emillon:
@nojb yes, I'll give it a last look, squash and merge.

doc/changes/10644.md
```diff
@@ -0,0 +1,4 @@
+- Add support for the -H flag (introduced in OCaml compiler 5.2) in dune
+  (requires lang versions 3.17). This adaptation gives
```
emillon, line 2:
```suggestion
  (requires dune lang 3.17). This adaptation gives
```

emillon:
I can't push to the original branch but that should be fine.

nojb:
> I can't push to the original branch but that should be fine.

I think this was my fault (I opened the PR from a branch in @MA0010's fork which disables this mechanism).

@MA0010 can you fix the remaining issue so that we can merge? Thanks!

emillon:
the last issues are just cosmetic - newlines etc and the changelog can be fixed at release time.
Thanks a lot to everyone involved here. This is the consequence of a multi-year effort that started in dune 1.7 (5 years ago), made its way to the compiler, and back here!

kit-ty-kate:
Thank you so much for this work!

Alizter:
Is tyxml supposed to be a test dependency here?

nojb:
> Is tyxml supposed to be a test dependency here?

Yes, there is a test that depends on it, so I guess so.

rgrinberg:
I think we should modify the test to avoid this dependency

nojb:
> I think we should modify the test to avoid this dependency

This specific test was added simply to exhibit the resolution of https://github.com/ocsigen/tyxml/issues/274. But there are other tests for `-H`, so we can just remove this one test.

nojb:
See #11728

PR #10649, opened 2024-06-14
copy_files: error out if no files found

### LINKED ISSUE #10646: `copy_files`: should Dune error out if no files are found?
Right now, Dune is lenient in the way `copy_files` work. If I do:

```
(copy_files foo/bar.txt)
```

But if the file `foo/bar.txt` doesn't exist, Dune will proceed just fine, as far as I can tell.

I wonder if it should error out if no files to copy are found given a glob? So that users can also get some help to maintain their `dune` files clean and avoid wasteful work for Dune (parsing, resolving globs, looking for inexistent files...)

jchavarri:
Fixes #10646.


src/dune_rules/simple_rules.ml
```diff
@@ -227,6 +227,8 @@ let copy_files sctx ~dir ~expander ~src_dir (def : Copy_files.t) =
     in
     Build_system.eval_pred (File_selector.of_glob ~dir glob)
   in
+  if Filename_set.is_empty files
```
rgrinberg, line 230: This has to be guarded by a version check since older versions can rely on the old behavior.

src/dune_rules/simple_rules.ml
```diff
@@ -227,6 +227,8 @@ let copy_files sctx ~dir ~expander ~src_dir (def : Copy_files.t) =
...
+  then User_error.raise ~loc [ Pp.textf "Cannot find any files to copy" ];
```
rgrinberg, line 231: How about `[ Pp.text "does not match any files" ]`?

jchavarri:
Thanks for the review @rgrinberg, I applied the suggested changes.

rgrinberg (approved):
LGTM. Could you add a CHANGES entry?

PR #10650, opened 2024-06-17
Fix generating sites modules failure when implicit_transitive_deps set to false

### LINKED ISSUE #9661: Modules generated for sites use an implicit transitive dep
## Expected Behavior
Compiling modules created with `generate_sites_module` when `implicit_transitive_deps` is set to `false` should work.

## Actual Behavior
It doesn't, failing with:

```
     ~section:Dune_section.Share
              ^^^^^^^^^^^^^^^^^^
Error: Unbound module Dune_section
```

From what I understand, the `dune-site` lib implicitly depends on `dune-private-libs.dune-section`.

As a workaround the library that uses sites can directly depend on `dune-private-libs.dune-section`.

## Specifications

- Version of `ocaml` (output of `ocamlc --version`): 5.1.0
- Operating system (distribution and version): macOS 14

MA0010:
This PR proposes an edit to allow building dune projects with `generate_sites_module` stanza, even when `implicit_transitive_deps` is set to false, which is the desired behavior.

In short, when using `genereate_sites_modules`, the module `sites.ml` is generated having `dune-site` library as a direct dependency, which in turn has `dune-section` private library as a transitive dep. Since `dune-section` is used explicitly in `sites.ml`, the build fails when `implicit_transitive_deps` is set to false (even with dune supporting hidden deps, see #10644). 

Therefore, dune-section is re_exported in the stanza of dune-site in order to avoid this failure.

Fixes #9661


doc/changes/10650.md
```diff
@@ -0,0 +1,3 @@
+- Re_export dune-section private library in the dune-site library stanza,
+    in order to avoid failure when generating and building sites modules
+    with implicit_transitive_deps = false. (#10650, @MA0100)
```
nojb, line 3: Please remove the extra leading whitespace.

MA0010: Done.

nojb (approved):
LGTM (also seen by @rgrinberg offline)

doc/changes/10650.md
```diff
@@ -0,0 +1,3 @@
+- Re_export dune-section private library in the dune-site library stanza,
+  in order to avoid failure when generating and building sites modules
+  with implicit_transitive_deps = false. (#10650, @MA0100)
```
nojb, line 3:
```suggestion
  with implicit_transitive_deps = false. (#10650, fixes #9661, @MA0100)
```

nojb: Friendly ping; if you make this change we can merge the PR.

PR #10655, opened 2024-06-19
use setup-ocaml v3

emillon:
The main feature is support for opam 2.2 on windows. This also means we can unify the `install deps` stage of the CI. For now, we keep `@runtest-windows`; more tests can be switched over later.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -70,10 +70,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Use OCaml ${{ matrix.ocaml-compiler }}
-        uses: ocaml/setup-ocaml@v2
+        uses: ocaml/setup-ocaml@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/workflow.yml
```diff
@@ -70,10 +70,10 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Use OCaml ${{ matrix.ocaml-compiler }}
-        uses: ocaml/setup-ocaml@v2
+        uses: ocaml/setup-ocaml@v3
         with:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
-          opam-depext: false
+          allow-prerelease-opam: true
```
emillon, line 76:
```suggestion
          allow-prerelease-opam: true
```

Leonidas-from-XIV (approved):
Nice, I appreciate the simplification!

PR #10657, opened 2024-06-19
fix(test): correctly scrub coq and coq/theories

emillon:
In `[^\s\f]*`, the `\` is interpreted literally, matching a backslash character. Instead we use `[:space:]` which corresponds to `\s`, but works in a `[]` context.

In addition, we have to fix the pattern so that it only matches directories that end with `coq` or `coq/theories`, otherwise it replaces too many directories and `$TESTCASE_ROOT` gets eaten.

This has the nice benefit of making the coq test suite work under nix, where `coq/theories` is installed elsewhere.


ejgallego (approved):
Thanks @emillon !

emillon:
Cool, this looks like it works on bsd sed as well. Otherwise, rewriting this as ocaml would have been an option.

PR #10658, opened 2024-06-19
test: do not install core_bench

emillon:
It is only used in benchmarks so unnecessary in most of CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,6 @@ BIN := ./_boot/dune.exe
 TEST_DEPS := \
 lwt \
 cinaps \
-core_bench \
 "csexp>=1.3.0" \
 js_of_ocaml \
 js_of_ocaml-compiler \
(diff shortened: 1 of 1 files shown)


emillon:
The "Install deps on Unix" phase does from 4m40s to 3m20s.

Leonidas-from-XIV (approved):
Very useful change though I would prefer if the "scrub num" change were a separate PR.

emillon:
Good idea, done in #10661.

rgrinberg:
does this mean that we somebody can make a change to our benchmarks and CI wouldn't catch the breakage?

emillon:
They're built in separate workflow, so I assume that it would be caught in the benchmark workflow.

rgrinberg:
The microbenchmarks don't run in any workflow as far as I can tell.

emillon:
Indeed; I've filed #10682 for this. Installing them separately will make it possible to install the "normal" test deps on windows in #10655 and we can add a dedicated linux-only (maybe nix) build that checks that.

PR #10660, opened 2024-06-19
ci: use nix for coq job

emillon:
We have to override the nixpkgs definition to get `coq-native`.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -187,20 +187,9 @@ jobs:
     name: Coq 8.16.1
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-      - name: Use OCaml 4.14.x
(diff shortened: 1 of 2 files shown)


emillon:
Before:  11m 21s
Now:  8m 46s

ejgallego:
Great, thanks!

ejgallego:
Indeed Nix provides very good build times for Coq, it is used heavily in some Coq projects.

emillon:
Thanks. Do you know if some projects use nix to build coq-native? I had to define it in here so it's not cached at the moment.

emillon:
Not ideal but an improvement already so I'm going for it.
cc @Alizter - if you know of a way to use coq-native from nix we can use it.

Alizter:
@emillon Unfortunately there is no way to use Coq native from nixpkgs. The way you've done it seems fine to me.

PR #10662, opened 2024-06-20
pkg: simplify the display logs

maiste:
Hi,
This PR is an attempt to make the _toolchain_ logs less verbose.  As the packages are evaluated as common rules, they printed a lot of lots. To overcome such behavior, the idea is to redirect the error logs into a file and print them in case of errors.

This PR is structured in 3 parts:
1. A modification of the `User_message` module to embedded `headers`. This information is printed before any other logs. It allows displaying the name of the package or a piece of information to give a bit of context. Maybe, this part is too much for this PR. I can either extract it or remove it.
2. Propagate the information such as the `package name` and the `location` to the function in charge of handling it.
3. Provide a `Output` module to sink the error logs into a file and display them as : Error if it fails and, warning if the command succeeds, but in verbose mode.

:warning: I left the feature flags turn to true for the review, but it might need to be removed before merging.

Note that I'm not award of most of the functionalities provided by the dune helper modules (`Stdune`, `User_message`, etc). So, feel free to suggest better implementation if it can improve the readability and fit better in the dune codebase! :pray: 

Closes tarides/team-build-system#30

### DIFF
--- a/bin/common.ml
+++ b/bin/common.ml
@@ -567,6 +567,7 @@ module Builder = struct
     ; debug_artifact_substitution : bool
     ; debug_load_dir : bool
     ; debug_digests : bool
+    ; debug_package_logs : bool
     ; wait_for_filesystem_clock : bool
     ; only_packages : Only_packages.Clflags.t
     ; capture_outputs : bool
@@ -666,6 +667,16 @@ module Builder = struct
             [ "debug-digests" ]
             ~docs
             ~doc:"Explain why Dune decides to re-digest some files")
+    and+ debug_package_logs =
+      let doc = "Always print the standard logs when building packages" in
+      Arg.(
+        value
+        & flag
+        & info
+            [ "debug-package-logs" ]
+            ~docs
+            ~doc
+            ~env:(Cmd.Env.info ~doc "DUNE_DEBUG_PACKAGE_LOGS"))
     and+ no_buffer =
       let doc =
         "Do not buffer the output of commands executed by dune. By default dune buffers \
@@ -970,6 +981,7 @@ module Builder = struct
     ; debug_artifact_substitution
     ; debug_load_dir
     ; debug_digests
+    ; debug_package_logs
     ; wait_for_filesystem_clock
     ; only_packages
     ; capture_outputs = not no_buffer
@@ -1236,6 +1248,7 @@ let init (builder : Builder.t) =
   Dune_engine.Clflags.debug_load_dir := c.builder.debug_load_dir;
   Dune_engine.Clflags.debug_fs_cache := c.builder.cache_debug_flags.fs_cache;
   Dune_digest.Clflags.debug_digests := c.builder.debug_digests;
+  Dune_rules.Clflags.debug_package_logs := c.builder.debug_package_logs;
   Dune_digest.Clflags.wait_for_filesystem_clock := c.builder.wait_for_filesystem_clock;
   Dune_engine.Clflags.capture_outputs := c.builder.capture_outputs;
   Dune_engine.Clflags.diff_command := c.builder.diff_command;
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -9,6 +9,7 @@ include struct
   module Checksum = Checksum
   module Source = Source
   module Build_command = Lock_dir.Build_command
+  module Display = Dune_engine.Display
 end
 
 module Variable = struct
@@ -405,7 +406,8 @@ module Expander0 = struct
   include Expander0
 
   type t =
-    { paths : Paths.t
+    { name : Dune_pkg.Package_name.t
+    ; paths : Paths.t
     ; artifacts : Path.t Filename.Map.t
     ; depends : (Variable.value Package_variable_name.Map.t * Paths.t) Package.Name.Map.t
     ; context : Context_name.t
@@ -514,6 +516,63 @@ module Substitute = struct
 end
 
 module Run_with_path = struct
+  module Output : sig
+    type error
+
+    val io : error -> Process.Io.output Process.Io.t
+
+    val with_error
+      :  accepted_exit_codes:int Predicate.t
+      -> pkg:Dune_pkg.Package_name.t * Loc.t
+      -> display:Display.t
+      -> (error -> 'a)
+      -> 'a
+
+    val prerr : rc:int -> error -> unit
+  end = struct
+    type error =
+      { pkg : Dune_pkg.Package_name.t * Loc.t
+      ; filename : Dpath.t
+      ; io : Process.Io.output Process.Io.t
+      ; accepted_exit_codes : int Predicate.t
+      ; display : Display.t
+      }
+
+    let io t = t.io
+
+    let with_error ~accepted_exit_codes ~pkg ~display f =
+      let filename = Temp.create File ~prefix:"dune-pkg" ~suffix:"stderr" in
+      let io = Process.Io.(file filename Out) in
+      let t = { filename; io; accepted_exit_codes; display; pkg } in
+      let result = f t in
+      Temp.destroy File filename;
+      result
+    ;;
+
+    let to_paragraphs t error =
+      let pp_pkg =
+        let pkg_name = Dune_pkg.Package_name.to_string (fst t.pkg) in
+        Pp.textf "Logs for package %s" pkg_name
+      in
+      let loc = snd t.pkg in
+      [ pp_pkg; Pp.verbatim error ], loc
+    ;;
+
+    let prerr ~rc error =
+      match Predicate.test error.accepted_exit_codes rc, error.display with
+      | false, _ ->
+        let paragraphs, loc = Stdune.Io.read_file error.filename |> to_paragraphs error in
+        User_warning.emit ~loc ~is_error:true paragraphs
+      | true, Display.Verbose ->
+        let content = Stdune.Io.read_file error.filename in
+        if not (String.is_empty content)
+        then (
+          let paragraphs, loc = to_paragraphs error content in
+          User_warning.emit ~loc paragraphs)
+      | true, _ -> ()
+    ;;
+  end
+
   module Spec = struct
     type 'path chunk =
       | String of string
@@ -525,6 +584,7 @@ module Run_with_path = struct
       { prog : Action.Prog.t
       ; args : 'path arg Array.Immutable.t
       ; ocamlfind_destdir : 'path
+      ; pkg : Dune_pkg.Package_name.t * Loc.t
       }
 
     let name = "run-with-path"
@@ -545,7 +605,7 @@ module Run_with_path = struct
 
     let is_useful_to ~memoize:_ = true
 
-    let encode { prog; args; ocamlfind_destdir } path _ : Dune_lang.t =
+    let encode { prog; args; ocamlfind_destdir; pkg = _ } path _ : Dune_lang.t =
       let prog =
         Dune_lang.atom_or_quoted_string
         @@
@@ -567,11 +627,12 @@ module Run_with_path = struct
     ;;
 
     let action
-      { prog; args; ocamlfind_destdir }
+      { prog; args; ocamlfind_destdir; pkg }
       ~(ectx : Action.Ext.context)
       ~(eenv : Action.Ext.env)
       =
       let open Fiber.O in
+      let display = !Clflags.display in
       match prog with
       | Error e -> Action.Prog.Not_found.raise e
       | Ok prog ->
@@ -589,31 +650,37 @@ module Run_with_path = struct
             ~var:"OCAMLFIND_DESTDIR"
             ~value:(Path.to_absolute_filename ocamlfind_destdir)
         in
-        Process.run
-          (Accept eenv.exit_codes)
-          prog
-          args
-          ~display:!Clflags.display
-          ~metadata
-          ~stdout_to:eenv.stdout_to
-          ~stderr_to:eenv.stderr_to
-          ~stdin_from:eenv.stdin_from
-          ~dir:eenv.working_dir
-          ~env
-        >>= (function
-         | Error _ -> Fiber.return ()
-         | Ok () -> Fiber.return ())
+        Output.with_error ~accepted_exit_codes:eenv.exit_codes ~pkg ~display (fun error ->
+          let stdout_to =
+            match !Clflags.debug_package_logs, display with
+            | true, _ | false, Display.Verbose -> eenv.stdout_to
+            | _ -> Process.Io.(null Out)
+          in
+          Process.run
+            Return
+            prog
+            args
+            ~display
+            ~metadata
+            ~stdout_to
+            ~stderr_to:(Output.io error)
+            ~stdin_from:eenv.stdin_from
+            ~dir:eenv.working_dir
+            ~env
+          >>= fun (_, rc) ->
+          Output.prerr ~rc error;
+          Fiber.return ())
     ;;
   end
 
-  let action prog args ~ocamlfind_destdir =
+  let action ~pkg prog args ~ocamlfind_destdir =
     let module M = struct
       type path = Path.t
       type target = Path.Build.t
 
       module Spec = Spec
 
-      let v = { Spec.prog; args; ocamlfind_destdir }
+      let v = { Spec.prog; args; ocamlfind_destdir; pkg }
     end
     in
     Action.Extension (module M)
@@ -742,7 +809,7 @@ module Action_expander = struct
     ;;
 
     let expand_pform
-      { env = _; paths; artifacts = _; context; depends; version = _ }
+      { name = _; env = _; paths; artifacts = _; context; depends; version = _ }
       ~source
       (pform : Pform.t)
       : (Value.t list, [ `Undefined_pkg_var of Package_variable_name.t ]) result Memo.t
@@ -877,7 +944,7 @@ module Action_expander = struct
          let ocamlfind_destdir =
            (Lazy.force expander.paths.install_roots).lib_root |> Path.build
          in
-         Run_with_path.action exe args ~ocamlfind_destdir)
+         Run_with_path.action ~pkg:(expander.name, prog_loc) exe args ~ocamlfind_destdir)
     | Progn t ->
       let+ args = Memo.parallel_map t ~f:(expand ~expander) in
       Action.Progn args
@@ -1007,6 +1074,7 @@ module Action_expander = struct
         (Pkg_info.variables pkg.info, pkg.paths)
     in
     { Expander.paths = pkg.paths
+    ; name = pkg.info.name
     ; artifacts = binaries
     ; context
     ; depends
--- a/test/blackbox-tests/test-cases/pkg/build-package-logs.t
+++ b/test/blackbox-tests/test-cases/pkg/build-package-logs.t
@@ -1,0 +1,48 @@
+Test the error message when installing package that fails.
+
+  $ . ./helpers.sh
+  $ make_lockdir
+  $ export DUNE_DEBUG_PACKAGE_LOGS=0
+
+Make a project with two packages, one successful and one that fails:
+
+  $ cat > dune-project << EOF
+  > (lang dune 3.12)
+  > EOF
+
+Create a package with a failing command that throws an error:
+
+  $ make_lockpkg x << EOF
+  > (version 0.0.1)
+  > (build
+  >    (progn
+  >       (run cat i_dont_exist)))
+  > EOF
+
+Building the package should fail and print an error:
+
+  $ build_pkg x 2>&1 | sed -E 's#/.*/cat#cat#g'
+  File "dune.lock/x.pkg", line 4, characters 11-14:
+  4 |       (run cat i_dont_exist)))
+                 ^^^
+  Error: Logs for package x
+  cat: i_dont_exist: No such file or directory
+  
+
+Create a package with a succeeding command that displays some text:
+
+  $ make_lockpkg y << EOF
+  > (version 0.0.1)
+  > (build
+  >    (progn
+  >       (run echo "Success!")))
+  > EOF
+
+Building the package should succeed and print no output:
+
+  $ build_pkg y
+
+Checks the package is installed:
+
+  $ show_pkg_cookie y
+  { files = map {}; variables = [] }
--- a/test/blackbox-tests/test-cases/pkg/pin-stanza/update-non-dune-local-pin.t
+++ b/test/blackbox-tests/test-cases/pkg/pin-stanza/update-non-dune-local-pin.t
@@ -41,7 +41,10 @@ Attempt to build the package the first time:
   echo aaa
   aaa
   false
-  -> required by _build/_private/default/.pkg/foo/target/cookie
+  File "dune.lock/foo.pkg", line 4, characters 6-13:
+            ^^^^^^^
+  Error: Logs for package foo
+  
 
 Update the message that gets printed while building foo:
   $ cat >foo/Makefile <<EOF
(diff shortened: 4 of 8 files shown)


emillon:
Thanks. It's good that you managed to use the existing process code to do this.
In terms of design, the most important thing to address is whether we can implement that without adding `headers` in `User_message`. If that's not possible, we could try to share the "add a message before" logic with the "Context:" message added in #10414. My impression is that it could be implemented by prepending a `paragraph`.
For the general state of the PR, it will be useful to add tests for that. I don't have a great deal of experience with testing pkg code, but I think we have some helpers for that - @Leonidas-from-XIV should be able to help you. This will be useful to make sure that the feature keeps working as we're iterating on it.
I'll add short remarks on the implem.

otherlibs/stdune/src/user_message.ml
```diff
@@ -226,13 +228,24 @@ let pp { loc; paragraphs; hints; annots = _ } =
            (Pp.textf "File %S, %s, characters %d-%d:" start.pos_fname lnum start_c stop_c))
       :: paragraphs
   in
+  let paragraphs = List.append headers paragraphs in
```
emillon, line 231: does that yield to double printing of headers in some cases?

otherlibs/stdune/src/user_message.ml
```diff
@@ -226,13 +228,24 @@ let pp { loc; paragraphs; hints; annots = _ } =
...
   Pp.vbox (Pp.concat_map paragraphs ~sep:Pp.nop ~f:(fun pp -> Pp.seq pp Pp.cut))
 ;;
 
 let print ?(config = Print_config.default) t =
   Ansi_color.print (Pp.map_tags (pp t) ~f:config)
 ;;
 
+let print_headers ?(config = Print_config.default) t =
+  let headers = Pp.concat_map t.headers ~sep:Pp.nop ~f:(fun pp -> Pp.seq pp Pp.cut) in
```
emillon, line 240: `Pp.nop` is the default value of `~sep` (here and below)

src/dune_pkg/feature_flags.ml
```diff
@@ -1,4 +1,4 @@
 open! Import
 
-let use_toolchains = false
-let use_autorelock = false
+let use_toolchains = true
+let use_autorelock = true
```
emillon, line 4: You mentionned that it's not necessary, but let's not forget to remove this. Also, the tests should work without these flags.
(note to self: we really need a way to enable that at runtime or compile time, since we're bound to forget at some point)

src/dune_rules/pkg_rules.ml
```diff
@@ -621,6 +623,109 @@ module Substitute = struct
...
 
 module Run_with_path = struct
+  module Output : sig
+    type t
+
+    val io : t -> Process.Io.output Process.Io.t
+
+    val v
+      :  ?accepted_exit_codes:int Predicate.t
+      -> ?pkg_name:Dune_lang.Package_name.t
```
emillon, line 633: we avoid optional arguments in new code - you can make `accepted_exit_codes` a labelled argument and `pkg_name` an option.

src/dune_rules/pkg_rules.ml
```diff
@@ -621,6 +623,109 @@ module Substitute = struct
...
+
+    let consume_and_print_error t display ~code ~loc =
+      match Predicate.test t.accepted_exit_codes code, display with
+      | false, _ ->
+        let msg = read t |> error_msg ~loc `Error t.pkg_name in
+        close t;
+        raise (User_error.E msg)
+      | true, Display.Verbose ->
+        let error = read t in
+        if has_output error
```
emillon, line 720:
```suggestion
        if not (String.is_empty error)
```

src/dune_rules/pkg_rules.ml
```diff
@@ -621,6 +623,109 @@ module Substitute = struct
...
+      ; accepted_exit_codes : int Predicate.t
+      ; mutable state : state
+      }
+
+    let prefix = "dune-pkg"
+    let exit_code_zero = Predicate.create (fun x -> x = 0)
+    let has_output error = error <> ""
+
+    let v ?(accepted_exit_codes = exit_code_zero) ?pkg_name suffix =
+      let fn = Temp.create File ~prefix ~suffix in
```
emillon, line 656: I think we can improve the resource management here:
- `v` kinds of sound like a constructor (we tend to use the `create` name for this), but it's actually opening a file.
- if you can restructure that to a "context manager" this will be easier to reason about. something like `val with_ : ?accepted_exit_codes -> ?pkg_name -> string -> f:(t -> 'a) -> 'a`, a bit like `In_channel.with_open`. that way you don't have to deal with `state`, `close` etc.

src/dune_rules/pkg_rules.ml
```diff
@@ -621,6 +623,109 @@ module Substitute = struct
...
+    ;;
+
+    let read t =
+      match t.state with
+      | Open ->
+        let content = Stdune.Io.read_file t.fn in
+        content
+      | Closed ->
+        raise
+          (User_error.E
```
emillon, line 676: This part is likely to go away but we prefer `Code_error.raise` for this kind of defensive check

src/dune_rules/pkg_rules.ml
```diff
@@ -621,6 +623,109 @@ module Substitute = struct
...
+        | `Warning -> Pp.text "Warning", User_message.Style.Warning
+        | `Error -> Pp.text "Error", User_message.Style.Error
+      in
+      let pp_package =
+        match pkg_name with
+        | None -> Pp.nop
+        | Some pkg_name ->
+          let pkg_name = Dune_pkg.Package_name.to_string pkg_name in
+          Pp.(
+            tag report_style (text "<><><>")
```
emillon, line 694: no need to mimic what opam is doing here. this clutters output and is less accessible than just printing the error message.

src/dune_pkg/feature_flags.ml
maiste: I agree.

src/dune_rules/pkg_rules.ml
maiste: - Yes, the `v` is an old `Irmin` habits... 
- It would make the code much simpler! Thanks

src/dune_pkg/feature_flags.ml
emillon: yes, I'll take care of this

emillon:
Let me clarify a bit what I mean:

- the first thing to do is to add tests for this. this way, we can discuss concretely about how the messages look
- this feature, as I understand it, is not strictly related to toolchains, is it? it's about suppressing outputs when building packages? (if I misunderstood, please correct me - this changes the constraints)
  - if that's indeed a pkg-generic feature, you can use the existing testing mechanisms for pkg rules. we don't refer to existing packages in the tests, but we'd define a package that fails. I didn't find a test that does exactly that, but for example `curl-not-installed.t` attempts to build a package without curl installed and shows how it fails.
- the context argument from #10414 means a `Context.t`, not a general "context" in the english sense. so you can't reuse it verbatim, but my point was that if we go that route you would be able to remove `context` in favor of a `header`. but let's first see if we can get rid of this `header` mechanism altogether.

src/dune_pkg/feature_flags.ml
Leonidas-from-XIV: I think a better way to go about it could be to target `main`. The log printing happens there as well and adding the PR to `toolchains` at this point only to upstream toolchains and having to upstream this afterwards anyway seems like it would require dealing with landing this PR essentially twice.

emillon: ah I had missed this. Yes, this should target `main`.

maiste:
Thanks for the clarifications! :pray:
> the first thing to do is to add tests for this. this way, we can discuss concretely about how the messages look

You are totally right. I'll add some of the corrections and add the tests then to help with the discussion.

> this feature, as I understand it, is not strictly related to toolchains, is it? it's about suppressing outputs when building packages? (if I misunderstood, please correct me - this changes the constraints) 

Sorry, my mistake: I still confuse the toolchain part and the packages part :sweat_smile: It's about suppressing the logs when building packages when there are no error, display only errors if we are in non-verbose and displaying all of them (warning in this case) when we are in verbose mode.

> the context argument from https://github.com/ocaml/dune/pull/10414 means a Context.t

I'm not sure to fully understand the point here. In the PR you linked, it seems that it's using `string option` not `Context.t`. 

To be fair, the `headers` part is only related to "how the messages look". If multiple packages fail, you have many errors, and it's hard to "humanly parse" displayed errors. Maybe this is not something that should land in this PR: this one could only be about printing logs in case of error or verbose mode and not care about the error style. Another PR could be more about refactoring the styling part.

src/dune_pkg/feature_flags.ml
maiste: Ok, I can rebase the work against `main` then

emillon: you'll need to change the base branch of the PR too (edit button next to title)

emillon:
> I'm not sure to fully understand the point here. In the PR you linked, it seems that it's using `string option` not `Context.t`.

that's to avoid dependency cycles, this is indeed a context name.

src/dune_rules/pkg_rules.ml
```diff
@@ -621,6 +623,109 @@ module Substitute = struct
...
+    type t =
+      { pkg_name : Dune_pkg.Package_name.t option
+      ; fn : Dpath.t
+      ; io : Process.Io.output Process.Io.t
+      ; accepted_exit_codes : int Predicate.t
+      ; mutable state : state
+      }
+
+    let prefix = "dune-pkg"
+    let exit_code_zero = Predicate.create (fun x -> x = 0)
```
Leonidas-from-XIV, line 652: You can also use `Int.equal 0` here.

src/dune_rules/pkg_rules.ml
```diff
@@ -621,6 +623,109 @@ module Substitute = struct
...
+
+    val consume_and_print_error : t -> Display.t -> code:int -> loc:Loc.t -> unit
+  end = struct
+    type state =
+      | Open
+      | Closed
+
+    type t =
+      { pkg_name : Dune_pkg.Package_name.t option
+      ; fn : Dpath.t
```
Leonidas-from-XIV, line 645: `fn` is a bit of a poor name, as it reminds me more of "function" than `filename`.

src/dune_rules/pkg_rules.ml
```diff
@@ -652,7 +759,7 @@ module Run_with_path = struct
 
     let is_useful_to ~memoize:_ = true
 
-    let encode { prog; args; ocamlfind_destdir } path _ : Dune_lang.t =
+    let encode { prog; args; ocamlfind_destdir; _ } path _ : Dune_lang.t =
```
Leonidas-from-XIV, line 762: I would suggest to specifically state the arguments that are being `_`'ed in such cases; otherwise you opt out of useful compiler errors that will tell you that you haven't dealt with an argument. Even if the case you have now might be legit, it will make it harder for future modifications.

(But in any case I believe `pkg_name` should also be encoded)

Leonidas-from-XIV (commented):
This is a good starting point! Happy to see it land soon.

I am a bit torn on including `Loc.t` as a value into the `Spec`. Usually we deal with them by having `Loc.t * 'a` types where `'a` is the value the `Loc.t` is related to. I've looked at some other action specs and the other `Spec`s don't include `Loc`s, maybe someone can pitch in how to do that in the best way?

src/dune_rules/pkg_rules.ml
```diff
@@ -514,6 +516,80 @@ module Substitute = struct
...
+      ; filename : Dpath.t
+      ; io : Process.Io.output Process.Io.t
+      ; accepted_exit_codes : int Predicate.t
+      ; display : Display.t
+      }
+
+    let io t = t.io
+
+    let with_error ~accepted_exit_codes ~pkg_name ~display f =
+      let filename = Temp.create File ~prefix:"dune-pkg" ~suffix:"stderr" in
```
Leonidas-from-XIV, line 544: This is okay in this PR, yet I think it could be useful to write these files into `_build` instead, similarly how we write `_build/log` into the build directory and not to a temporary file. That way users have the option to navigate to the folder containing the log output and see what the outputs were.

Similarly, it would also be interesting to have the `stdout` captured in the same way, for example output of `./configure` style scripts (to answer "was the library successfully detected etc.)

Leonidas-from-XIV (commented):
I was thinking of this and this is not a blocker but I thought it could be nice to have. But it can happen in a future PR as well.

src/dune_rules/pkg_rules.ml
maiste: I agree with your proposal, as it would be nice to be able to get the error logs after the `build`. However, it might be something we want to do in another PR as the scope of this one would be too wide.

src/dune_rules/pkg_rules.ml
```diff
@@ -589,31 +670,41 @@ module Run_with_path = struct
...
+               ~dir:eenv.working_dir
+               ~env
+             >>= fun (_, rc) ->
+             Output.prerr ~loc ~rc error;
+             Fiber.return ())
     ;;
   end
 
-  let action prog args ~ocamlfind_destdir =
+  let action ~loc ?pkg_name prog args ~ocamlfind_destdir =
```
rgrinberg, line 700: No optional arguments please

rgrinberg (changes requested):

rgrinberg:
If this PR is adding a file that isn't a target of a rule, what prevents the file from being deleted by dune in subsequent runs during stale artifact deletion?

rgrinberg:
> A modification of the User_message module to embedded headers. This information is printed before any other logs. It allows displaying the name of the package or a piece of information to give a bit of context. Maybe, this part is too much for this PR. I can either extract it or remove it.

I would suggest that you try to accomplish what your PR needs without this change. Such headers are not transmitted through RPC nor are they interpreted by any of the editors. This makes our editor integration incomplete. It seems like a rather general feature, but I can't find a use for it anywhere else in the codebase.

maiste:
I have removed the `headers` related parts to make this PR more atomic. There is now a cram test to ensure the mechanism would be break on purpose. @Leonidas-from-XIV, I follow your suggestion and add the location in the type as a tuple, as it's done everywhere.

It should be much simpler to review now, as it temporarily redirects the errors in a file and displays them or not, depending on the result of the command.

src/dune_rules/pkg_rules.ml
```diff
@@ -514,6 +516,65 @@ module Substitute = struct
...
+    let with_error ~accepted_exit_codes ~pkg ~display f =
+      let filename = Temp.create File ~prefix:"dune-pkg" ~suffix:"stderr" in
+      let io = Process.Io.(file filename Out) in
+      let t = { filename; io; accepted_exit_codes; display; pkg } in
+      let result = f t in
+      Temp.destroy File filename;
+      result
+    ;;
+
+    let error_msg ~is_critical t error =
```
emillon, line 552: the `is_critical` bool is a bit weird, I wonder if that would be clearer to pass the message instead

maiste: This issue is that in one case it's a `Warning` and in another one it's an `Error`. We need to stick to the behaviour in `Process` and raise only on critical error. I can change it by adding a `kind` instead of  `is_critical` using variant.

emillon: Ah ok I missed that part. My main issue was the `critical` bit, indeed. What you're suggesting is fine by me, but as a final suggestion, maybe we can use `User_warning.emit` and the `is_error` flag, with just `Logs for package X` as message in both cases. I wonder if that would be enough?

maiste: As it would display `Error` and `Warning` in front of the message, I think his is enough for now :+1:

test/blackbox-tests/test-cases/pkg/build-package-logs.t
```diff
@@ -0,0 +1,44 @@
...
+  [1]
+
+We create a package with a succeeding command that displays some text:
+
+  $ make_lockpkg y << EOF
+  > (version 0.0.1)
+  > (build 
+  >    (progn
+  >       (run echo "Success!")
+  >       (run true)))
```
Leonidas-from-XIV, line 39: This is not necessary, echo will already return 0.

test/blackbox-tests/test-cases/pkg/build-package-logs.t
```diff
@@ -0,0 +1,44 @@
...
+  > EOF
+
+Building the package should succeed and print no output:
+
+  $ build_pkg y
```
Leonidas-from-XIV, line 44: It would be nice if you could make sure that the package `y` succeeded to install. I think `helpers.sh` has a function to display the "cookie" of a package and that should presumably exist when the package was successfully installed.

```sh
  $ show_pkg_cookie y
```

or something like this.

Leonidas-from-XIV (commented):
This PR breaks some other tests (which relied on the fact that the output was always printed). Could you go and fix them? I assume that requires passing `--verbose` to where this is necessary.

maiste:
@Leonidas-from-XIV, I'm not sure about the path I should take with this. Indeed, if I add a verbose flag, the tests would pass, but the logging result is pretty big... Is OK to do so?

Leonidas-from-XIV:
You could prefix the tests to output a specific output and do `dune build | grep <THAT-PREFIX>`. Alternatively, make the tests output their output to a file that's getting installed and then `cat` that file to show the result. The latter might be a tiny bit cleaner but also a tiny bit more involved.

rgrinberg (approved):
I'll give others a chance to review. This is good to go from me.

rgrinberg:
To make it easy to fix the tests, you could add restore the old behavior with `INSIDE_DUNE` or another variable.

maiste:
Thanks, @rgrinberg!
For the tests, I'm updating them to use a file as @Leonidas-from-XIV suggested. It would make the ones concerned more resilient to output changes. Indeed, the goal of the tests that currently break is to test the behavior, not the output. This way, the test I added would ensure the output stay consistent and the other ones the behavior.

christinerose (commented):
Changed make to `make`, but if there is a different way to do that in `.t` files please let me know, and I'll fix it.

emillon (commented):
The implementation looks good, but I'm concerned with the changes in tests TBH.
The problem isn't really the diff size, but rather that what we're testing doesn't match the reality (real build instructions don't write to files; they output on stdout and stderr).

Since it looks like `--display verbose` is too verbose, I suggest starting with `INSIDE_DUNE` so that we can at least land this PR. (I can offer you guidance about how to pretend we're not `INSIDE_DUNE` to write `build-package-logs.t`)

Then, in a future PR, we can see how we can make better tests. The pain we're seeing is because we're testing through too many layers, so maybe we could try to define a helper executable that directly executes a package build, or something like that.

Leonidas-from-XIV:
> real build instructions don't write to files; they output on stdout and stderr

I'd say the opposite, real build instructions write to files and all output to stdout is a side-effect (or execute commands based on filters). In the case of `opam` the output gets hidden away by default as well, thus if the output was important for builds `opam` would be breaking them.

maiste:
I agree with @Leonidas-from-XIV: in the "real build instruction" we write to files (many times it's just a call to `dune build`). Also, relying on `INSIDE_DUNE` would not reflect the reality of what happens at the logging level. Maybe, truly rewriting the test (not only changing the output to a file), would be something we want to do once we have a stable structure for pkg and toolchain?

emillon:
> real build instructions write to files and all output to stdout is a side-effect

Let me clarify what I meant; there are some packages where `configure` outputs things like paths to stdout.
We also have tests that mimic that with `echo %{lib}%`, those don't write to files.

> Maybe, truly rewriting the test (not only changing the output to a file), would be something we want to do once we have a stable structure for pkg and toolchain?

Yes, that's what I was suggesting; we disagree on the intermediate step. If you use `INSIDE_DUNE`, you'll have to change the tests only once; by writing to temporary files you'll have to change them twice, with an intermediate state that's a bit messy IMO.

maiste:
> Let me clarify what I meant; there are some packages where configure outputs things like paths to stdout.
We also have tests that mimic that with echo %{lib}%, those don't write to files.

Out of curiosity, is it a test that I updated? :confused: 

Considering your comment, if we have some packages relying on `stdout` in the `pkg` context, suppressing `stdout` logging would break them. In this scenario, it is not a question of testing, it means that my fix would break some behaviors and make it impossible to have in `dune pkg`. Using `INSIDE_DUNE` would only hide this problem "under the carpet", no? Packages would break at installation time?

> Yes, that's what I was suggesting; we disagree on the intermediate step. If you use INSIDE_DUNE, you'll have to change the tests only once; by writing to temporary files you'll have to change them twice, with an intermediate state that's a bit messy IMO.

Technically, the changes are already here: it would be changing them once now ^^ Joke aside, I don't mind removing them: my only concern here is that we would put `inside_dune` for a behavior that is not only related to `dune` internally. It would  "unsynchronized" the tests from the behavior that happens in "the real world". I agree that the test proposition here is not perfect, and it means we will have to change them at some point. However, this intermediate state allows the tests to break for the right reason: if another PR changes the output but not the internal behavior, it would only break 1 test, not the 30 others.

maiste:
After an offline discussion with @emillon, I agree with his suggestion of adding a flag for printing the output when present, `--debug-package-logs`. 

This way, it provides access to the output when building packages without making `dune` in the `inside_dune` mode. The conclusion was to add this so we can merge this PR. 

I'll open a new one with the `cat` version of the tests, so we'll have a proper place to discuss it. As a result, we would be able to merge this PR and use the test PR to sanitize the tests (if we reach the conclusion that this is what we need to do).

@emillon @Leonidas-from-XIV, I let you merge if this is good for you this way.

maiste:
@emillon @Leonidas-from-XIV
I updated the tests and rebase on master. If that's good for you, it should be possible to merge it :+1:

bin/common.ml
```diff
@@ -666,6 +667,14 @@ module Builder = struct
...
             ~docs
             ~doc:"Explain why Dune decides to re-digest some files")
+    and+ debug_package_logs =
+      Arg.(
+        value
+        & flag
+        & info
+            [ "debug-package-logs" ]
+            ~docs
+            ~doc:"Always print the standard logs when building packages")
```
emillon, line 677: If you add this, you won't have to explicitly pass the flag in tests:

```
~env:(Cmd.Env.info ~doc "DUNE_DEBUG_PACKAGE_LOGS"))
```

emillon (approved):
Thanks. You can accept the env var directly in the cmdliner term to simplify the setup and reduce the diff again.
Please add a changelog entry as well.

test/blackbox-tests/test-cases/pkg/helpers.sh
```diff
@@ -5,7 +5,7 @@ dune="dune"
 pkg_root="_build/_private/default/.pkg"
 
 build_pkg() {
-  $dune build $pkg_root/$1/target/
+    $dune build $pkg_root/$1/target/
```
Leonidas-from-XIV, line 8: This change can probably be removed. Or not. Of little importance anyway.

Leonidas-from-XIV (approved):
Looks good to me, can be (squashed &) merged if tests pass.

emillon:
I'll cleanup the diff and merge. Thanks!

test/blackbox-tests/test-cases/pkg/pin-stanza/update-non-dune-local-pin.t
```diff
@@ -51,8 +54,11 @@ Update the message that gets printed while building foo:
...
 The change to the package is picked up:
-  $ dune build 2>&1 | grep -v make
+  $ dune build --debug-package-logs 2>&1 | grep -v make
   echo bbb
   bbb
   false
-  -> required by _build/_private/default/.pkg/foo/target/cookie
+  File "dune.lock/foo.pkg", line 4, characters 6-13:
+            ^^^^^^^
+  Error: Logs for package foo
```
emillon, line 63: The loc doesn't seem right here (`^^^` points at nothing). Let's move that to a new issue.

PR #10664, opened 2024-06-21
test(windows): set eol=lf on sh files

emillon:
This ensures that these files are created with LF endings, even when checking out on windows. Otherwise, some tests using these will fail.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,6 +2,7 @@
 *.rst             text eol=lf
 *.c               text eol=lf
 *.t               text eol=lf -linguist-detectable
+*.sh              text eol=lf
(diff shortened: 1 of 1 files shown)


emillon:
(this is consistent with e.g. the ocaml/opam repository)

Leonidas-from-XIV (approved):
Seems reasonable to me.

Maybe it would be even better to mark all files as `eol=lf` and only specify `.ps1` as differing?

From `man gitattributes` it sounds to me like git will walk all lines and use the last line that matches for the actual value so something like `* text eol=lf` could be defined as the first one. WDYT?

PR #10666, opened 2024-06-21
test(win): use bash instead of system when needed

emillon:
On Windows, `(system)` uses `cmd` and fails on these commands.
Using `(bash)` ensures that it is used even on Windows.

### DIFF
--- a/test/blackbox-tests/test-cases/directory-targets/dune
+++ b/test/blackbox-tests/test-cases/directory-targets/dune
@@ -1,3 +1,7 @@
 (cram
  (deps helpers.sh)
  (applies_to :whole_subtree))
+
+(cram
+ (applies_to remove-write-permissions)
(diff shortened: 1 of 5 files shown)


maiste (commented):
In the `blackbox-tests/test-cases` directory, it seems that `system` is used at many places. Is there a reason why you targeted these three files in particular and not the other ones?

emillon:
Yes: these are the ones I saw fail with "invalid syntax" in #10655. There's a bunch of cases that are valid uses of `(system)`, and maybe there will be more cases to fix as we fix more things.

maiste (approved):
Maybe other fixes should land on another PR and this one can be merged?

rgrinberg:
These tests deserve a `%{bin:bash}` dependency as well

PR #10668, opened 2024-06-25
pkg: workaround to avoid unstable compilers

### LINKED ISSUE #10592: pkg: the solver prefers unreleased version of ocaml-variants over ocaml-base-compiler

gridbugs:
Fixes https://github.com/ocaml/dune/issues/10592.

Solves an issue where the solver would choose unstable releases of the compiler. When the latest version of the "ocaml" package can only be satisfied by unstable versions of "ocaml-variants", dune would still include this unstable compiler in package solutions. This leads to users being given unstable versions of the compiler by default. This is only a problem because dune doesn't yet fully implement the avoid-version flag, and compiler packages are released under the assumption that the solver respects this flag.

The workaround introduced in this change is to determine the latest stable version of "ocaml-base-compiler" (the latest version lacking the avoid-version flag), and have the solver prefer versions of the "ocaml" package that are no later than this version.

### DIFF
--- a/otherlibs/stdune/src/tuple.ml
+++ b/otherlibs/stdune/src/tuple.ml
@@ -20,4 +20,13 @@ module T3 = struct
   let to_dyn = Dyn.triple
   let hash f g h (a, b, c) = Poly.hash (f a, g b, h c)
   let equal f g h (a1, b1, c1) (a2, b2, c2) = f a1 a2 && g b1 b2 && h c1 c2
+
+  let compare f g h (a1, b1, c1) (a2, b2, c2) =
+    match f a1 a2 with
+    | (Ordering.Lt | Gt) as x -> x
+    | Eq ->
+      (match g b1 b2 with
+       | (Ordering.Lt | Gt) as x -> x
+       | Eq -> h c1 c2)
+  ;;
 end
--- a/otherlibs/stdune/src/tuple.ml
+++ b/otherlibs/stdune/src/tuple.ml
@@ -20,13 +20,4 @@ module T3 = struct
   let to_dyn = Dyn.triple
   let hash f g h (a, b, c) = Poly.hash (f a, g b, h c)
   let equal f g h (a1, b1, c1) (a2, b2, c2) = f a1 a2 && g b1 b2 && h c1 c2
-
-  let compare f g h (a1, b1, c1) (a2, b2, c2) =
-    match f a1 a2 with
-    | (Ordering.Lt | Gt) as x -> x
-    | Eq ->
-      (match g b1 b2 with
-       | (Ordering.Lt | Gt) as x -> x
-       | Eq -> h c1 c2)
-  ;;
 end
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -36,6 +36,13 @@ let opam_file_is_avoid_version (opam_file : OpamFile.OPAM.t) =
   List.mem opam_file.flags Pkgflag_AvoidVersion ~equal:Poly.equal
 ;;
 
+let opam_file_is_ocaml_package (opam_file : OpamFile.OPAM.t) =
+  Option.equal
+    OpamPackage.Name.equal
+    opam_file.name
+    (Some (OpamPackage.Name.of_string "ocaml"))
+;;
+
 module Context_for_dune = struct
   type 'a monad = 'a Monad.t
   type filter = OpamTypes.filter
@@ -65,6 +72,7 @@ module Context_for_dune = struct
          packages for which we've printed a warning. *)
       available_cache : (OpamPackage.t, bool) Table.t
     ; constraints : OpamTypes.filtered_formula Package_name.Map.t
+    ; latest_released_base_compiler_version : OpamPackage.Version.t option
     }
 
   let create
@@ -75,6 +83,7 @@ module Context_for_dune = struct
     ~version_preference
     ~stats_updater
     ~constraints
+    ~latest_released_base_compiler_version
     =
     let dune_version =
       let major, minor = Dune_lang.Stanza.latest_version in
@@ -106,6 +115,7 @@ module Context_for_dune = struct
     ; candidates_cache
     ; available_cache
     ; constraints
+    ; latest_released_base_compiler_version
     }
   ;;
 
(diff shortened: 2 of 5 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -750,6 +862,23 @@ module Solver_result = struct
...
 end
 
+(* Returns the latest version of the package named
+   "ocaml-base-compiler" which does not have the avoid-version flag
+   set. This will be used to allow the solver to prefer non-development
+   versions of compiler packages, working around the lack of support for
+   the avoid-version flag in opam-0install-solver. *)
+let latest_released_base_compiler_version repos =
+  let+ all_versions =
+    OpamPackage.Name.of_string "ocaml-base-compiler" |> Opam_repo.load_all_versions repos
```
rgrinberg, line 872: Seems a bit annoying to do this work if `version_preference = Oldest`. Maybe just set this to None in that case?

gridbugs: done

Leonidas-from-XIV (approved):

src/dune_pkg/opam_solver.ml
```diff
@@ -36,6 +36,13 @@ let opam_file_is_avoid_version (opam_file : OpamFile.OPAM.t) =
   List.mem opam_file.flags Pkgflag_AvoidVersion ~equal:Poly.equal
 ;;
 
+let opam_file_is_ocaml_package (opam_file : OpamFile.OPAM.t) =
+  Option.equal
+    OpamPackage.Name.equal
+    opam_file.name
+    (Some (OpamPackage.Name.of_string "ocaml"))
```
rgrinberg, line 43: Don't we have a conflict class for OCaml packages? Would be good to leave a comment why we can't use it here.

src/dune_pkg/opam_solver.ml
```diff
@@ -106,26 +115,123 @@ module Context_for_dune = struct
...
 
+  let opam_version_compare a b = OpamPackage.Version.compare a b |> Ordering.of_int
+
+  (* Returns a bool which is true iff the following conditions are met:
+     - the version preference is [Newest]
+     - the given package is the ocaml compiler metapackage
+     - the version of the given package is later than the latest
+       version of the ocaml-base-compiler package lacking the
+       avoid-version flag *)
+  let opam_file_is_ocaml_newer_than_latest_base_compiler_if_prefer_newest
```
rgrinberg, line 134: I still find this code hard to read tbh. I think the conversion to booleans to define and order is what's making it hard to read. Will try to get rid of it.

src/dune_pkg/opam_solver.ml
gridbugs: I assume you're referring to the `ocaml-core-compiler` conflict class. The `ocaml` package isn't part of that class. It seems to only contain compiler implementations such as `ocaml-base-compiler` and `ocaml-variants`. How were you imagining to use this class to help here? Just as a way to avoid hard-coding the "ocaml" package name?

Leonidas-from-XIV: Even if using the conflict class, we then would just hardcode one name (`ocaml-core-compiler`) for another (`ocaml`).

rgrinberg: It's not about hard coding one particular name over the other. It's about making sure our code stays consistent with recognizing the compiler. I recall that I hard coded the name of the compiler package in another feature and it turned out wrong. Another reason for doing this in a principled way is that dune should support compiler forks just as well as it supports the original compiler.

src/dune_pkg/opam_solver.ml
```diff
@@ -106,26 +115,123 @@ module Context_for_dune = struct
...
+    t
+    (opam_file : OpamFile.OPAM.t)
+    =
+    match t.version_preference with
+    | Oldest -> false
+    | Newest ->
+      (match opam_file_is_ocaml_package opam_file with
+       | false -> false
+       | true ->
+         (match Option.both opam_file.version t.latest_released_base_compiler_version with
```
rgrinberg, line 144: When can the versions be absent here?

gridbugs: For the lhs, only if the package being compared is named "ocaml" but lacks a version, and for the rhs, only if the opam repo has no package named "ocaml-base-compiler". Neither of these cases should happen in practice but I'm trying to minimize the assumptions dune makes about the opam repo.

Though I notice that elsewhere in this file we call `OpamFile.OPAM.version` which raises if its argument doesn't have a version, so there's little point checking for the version field here.

src/dune_pkg/opam_solver.ml
gridbugs: To clarify, @rgrinberg are you planning to do this work or are you asking me to take a look at it. Is there any harm in merging it as is and making some readability changes in a later PR? The build system team is planning to do some more developer interviews soon and this change is necessary for that work so it would help us if it could be merged soon.

rgrinberg: I will try to simplify the code a bit

rgrinberg:
After thinking about, I've implemented a different workaround for this problem. It's imperfect, but at least it's simple to explain: we just skip avoid-versions packages and just treat them as unavailable.

The previous workaround was a little too complicated to think about. As a rule of thumb, if something is difficult to explain to other devs on the dune team, it's going to be even more painful to the wider user base. For now, let's keep it simpler and re-evaluate once we get more feedback.

If we end up needing packages with avoid-version, we can just introduce a special solver mode that allows for them (perhaps asking the users for a package white list?)

rgrinberg:
I kept a slightly cleaned up version of the original workaround in the commit history. Just in case we want to revisit things.

otherlibs/stdune/src/tuple.ml
```diff
@@ -20,4 +20,13 @@ module T3 = struct
   let to_dyn = Dyn.triple
   let hash f g h (a, b, c) = Poly.hash (f a, g b, h c)
   let equal f g h (a1, b1, c1) (a2, b2, c2) = f a1 a2 && g b1 b2 && h c1 c2
+
```
gridbugs, line 23: This function is no longer used.

gridbugs:
Looks good to me

Issue #10670, opened 2024-06-26
Dependency solver attempts to choose system-mingw dependency of ocaml-base-compiler on macos and linux


gridbugs:
Solving on linux or macos with a dependency on `(ocaml-base-compiler (= 5.2.0))` now gives the error:
```
Error: Unable to solve dependencies for the following lock directories:
Lock directory dune.lock:
Can't find all required versions.
Selected: ... ocaml-base-compiler.5.2.0 system-mingw
- system-mingw -> (problem)
    No usable implementations:
      system-mingw.1: Availability condition not satisfied
```

This seems to have been introduced when the ocaml packages in the opam repo gained native windows support in https://github.com/ocaml/opam-repository/commit/0b240d2960133fd3d8aa8f008d7aa79534caa3b9.

Repro: https://github.com/ocaml/dune/pull/10672

gridbugs:
This happens because dune solves dependencies with the `post` variable set to false, so the following disjunction in ocaml-base-compiler's dependencies cannot be resolved:
```
(("arch-x86_64" {os = "win32" & arch = "x86_64"} & "system-mingw" &
    "mingw-w64-shims" {os-distribution = "cygwin" & build}) |
   ("arch-x86_32" {os = "win32"} & "ocaml-option-bytecode-only" &
    "system-mingw" &
    "mingw-w64-shims" {os-distribution = "cygwin" & build}) |
   "host-system-other" {os != "win32" & post})
```

rgrinberg:
Anyway to talk to upstream about not using post for this? post dependencies don't really make sense in dune.

gridbugs:
Yeh sure I'll chat with upstream. Can you elaborate a little on why post deps don't make sense in dune though. Is the idea that post deps in opam only make sense for executables that need a dep at runtime but not at compile time, and dune package management is only intended to install library dependencies (with the exception of the compiler, ocamlfind, ocamlbuild, and other build tools)?

rgrinberg:
I don't recall the exact reasons, but I think it went something like this: all packages in dune need to define their dependencies explicitly. So there's no point to "install" a post dependency since it will not be available to anyone unless they explicitly ask for it. So if we have a package `foo` with a post dep `bar`, we can have two situations:

- a package depends on `foo` and `bar` - requires building bar with `post = true`
- a package depends on `bar` only - requires building bar with `post = false`

Or something along those lines. I could have misunderstood how post dependencies worked though. It would be good if you could revisit that.

[...]

PR #10672, opened 2024-06-26
pkg: repro for gh10670

gridbugs:
Repro for https://github.com/ocaml/dune/issues/10670


test/blackbox-tests/test-cases/pkg/gh10670.t
```diff
@@ -0,0 +1,44 @@
+Reproduces https://github.com/ocaml/dune/issues/10670
+
+Dune solves packages with `` set to false, so the disjunction in
```
Leonidas-from-XIV, line 3: Is this missing a bit of text?

rgrinberg, line 3: I think you mean `"post"` here. There's no markdown in cram files by the way.

rgrinberg (approved):

Issue #10675, opened 2024-06-27
Incompatibility between `enabled_if` and `foreign_stubs`?


shym:
Using dune version 3.16.0, I’ve tried to define 2 ways to build an executable depending on the context, once adding a foreign stub (to override the default `main`). This makes dune crash.
Only the `dune` file contains something really relevant I think, but here is the full setup:

```
$ cat > dune << EOF
(executable
 (enabled_if
  (<> %{context_name} other))
 (name main))

(executable
 (enabled_if
  (= %{context_name} other))
 (name main)
 (foreign_stubs
  (language c)
  (names startup)))
EOF

$ cat > dune-project << EOF
(lang dune 3.0)
EOF

$ cat > dune-workspace << EOF
(lang dune 2.0)

(context (default))

(context
 (default
  (name other)))
EOF

$ cat > main.ml << EOF
let () = Printf.printf "Hello from OCaml!\n%!"
EOF

$ cat > startup.c << EOF
#include <stdio.h>
#include <caml/callback.h>

int main(int argc, char **argv) {
    printf("Hello from C!\n");
    caml_startup(argv);
    return 0;
}
EOF
```

With that setup, `dune build` crashes with:

```
Internal error, please report upstream including the contents of _build/log.
Description:
  ("Map.of_list_map_exn", { key = "main" })
Raised at Stdune__Code_error.raise in file
  "otherlibs/stdune/src/code_error.ml", line 10, characters 30-62
Called from Dune_rules__Foreign_sources.make in file
  "src/dune_rules/foreign_sources.ml", lines 245-247, characters 4-19
Called from Fiber__Core.O.(>>|).(fun) in file "vendor/fiber/src/core.ml",
  line 253, characters 36-41
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
-> required by ("<unnamed>", ())
-> required by ("<unnamed>", ())
-> required by ("load-dir", In_build_dir "other")
-> required by
   ("build-alias", { dir = In_build_dir "other"; name = "default" })
-> required by ("toplevel", ())

I must not crash...
```

with the following build log (redacted with pseudo-variables for legibility):

```
# dune build
# OCAMLPARAM: unset
# Shared cache: disabled
# Shared cache location: $HOME/.cache/dune/db
# Workspace root: $PWD
# Auto-detected concurrency: 16
# Dune context:
#  { name = "other"
#  ; kind = "default"
#  ; profile = Dev
[...]

PR #10676, opened 2024-06-27
test: repro for #10675

emillon:


shym:
Nicely shorter than my original example, thank you!

PR #10681, opened 2024-07-01
Use target triples to describe action binaries

maiste:
Hi
This PR changes the name of the binary targets to make them use _triple targets_ instead.

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -11,16 +11,16 @@ jobs:
       matrix:
         include:
           - os: macos-13
-            name: mac-intel
+            name: x86_64-apple-darwin
             installable: .#
           - os: macos-14
-            name: mac-arm
+            name: aarch64-apple-darwin
             installable: .#
           - os: ubuntu-22.04
-            name: linux-intel
+            name: x86_64-unknown-linux-gnu
             installable: .#
           - os: ubuntu-22.04
-            name: linux-intel-static
+            name: x86_64-unknow-linux-musl
             installable: .#dune-static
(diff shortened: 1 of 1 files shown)


.github/workflows/binaries.yml
```diff
@@ -11,16 +11,16 @@ jobs:
...
-            name: mac-arm
+            name: aarch64-apple-darwin
             installable: .#
           - os: ubuntu-22.04
-            name: linux-intel
+            name: x86_64-unknown-linux-gnu
             installable: .#
           - os: ubuntu-22.04
-            name: linux-intel-static
+            name: x86_64-unknow-linux-musl
```
emillon, line 23: typo here

.github/workflows/binaries.yml
```diff
@@ -11,16 +11,16 @@ jobs:
...
         include:
           - os: macos-13
-            name: mac-intel
+            name: x86_64-apple-darwin
             installable: .#
           - os: macos-14
-            name: mac-arm
+            name: aarch64-apple-darwin
             installable: .#
           - os: ubuntu-22.04
```
emillon, line 19: I thought we got rid of that one, maybe that was only on the toolchains branch

maiste: Do you want me to remove this target or we leave it here?

emillon: can you do it in a separate PR so we don't forget?

maiste: Sure, no problem :+1: 
Done in #10683

PR #10692, opened 2024-07-03
refactor: remove field build_if from test stanzas

vouillon:
Reuse field exes.enabled_if instead

### DIFF
--- a/src/dune_rules/stanzas/tests.mli
+++ b/src/dune_rules/stanzas/tests.mli
@@ -6,7 +6,6 @@ type t =
   ; package : Package.t option
   ; deps : Dep_conf.t Bindings.t
   ; enabled_if : Blang.t
-  ; build_if : Blang.t
   ; action : Dune_lang.Action.t option
   }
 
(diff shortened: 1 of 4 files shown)


src/dune_rules/stanzas/tests.mli
```diff
@@ -6,7 +6,6 @@ type t =
   ; package : Package.t option
   ; deps : Dep_conf.t Bindings.t
   ; enabled_if : Blang.t
-  ; build_if : Blang.t
```
rgrinberg, line 9: Don't we use `build_if` and `enabled_if` for different things for tests?

1. enabled_if is used to determine whether running the executable is attached to the alias
2. build_if is used to determine if the test executable is built at all

vouillon: Right. The point is that we have a field `enabled_if` in type `Executables.t` to indicate whether to build the executables, so it makes a lot of sense to use this field rather than an additional field in type `Tests.t`.

PR #10694, opened 2024-07-04
test: Exercise the current semantics of checksum selection

Leonidas-from-XIV:
It might be useful to have a test that shows how the checksum selection works, which checksum is selected and how it is recorded.

The test shows that at the moment we always pick the first checksum.


test/blackbox-tests/test-cases/pkg/multiple-checksums.t
```diff
@@ -0,0 +1,42 @@
...
+  > EOF
+
+Locking should work
+
+  $ add_mock_repo_if_needed
+  $ dune pkg lock
+  Solution for dune.lock:
+  - multiple-checksums.0.0.1
+
+The generated lock file should contain at least one checksum:
```
maiste, line 34: Are we sure it is going to take the `md5` first? If it takes the first one in the _opam_ list, maybe it should be reflected in this sentence.

Leonidas-from-XIV: In our code at the moment it definitely does always take the first: https://github.com/ocaml/dune/blob/425743ccaa353a65a36fa86f2551faf3afa6fa3b/src/dune_pkg/opam_solver.ml#L598-L599 (`extra-sources`) and https://github.com/ocaml/dune/blob/425743ccaa353a65a36fa86f2551faf3afa6fa3b/src/dune_pkg/opam_solver.ml#L608-L610 (`url`) but I am indeed mystified by the comment that `opam` discards the other checksums. I'll investigate.

Leonidas-from-XIV: So reading `OpamRepository.validate_and_add_to_cache` (as pointed to by @rjbou) then opam validates all the checksums, it doesn't take the first. I've removed that comment, because it was misleading.

In the cases where Opam needs a single checksum (e.g. to pick the location where to save it in the cache it calls `OpamHash.sort` which sorts by hashing algorithm https://github.com/ocaml/opam/blob/e85673526bac320ff849f756ffbd54086bde2811/src/core/opamHash.ml#L20 and and uses that as "main" checksum.

maiste: This is something you wanted to change at some point, IIRC? Because if opam takes all the checksum, we also want such a behavior, no?

Leonidas-from-XIV:
The `hash-algos.t` test is actually nicer because it spends less time with setup, I've ended up just copying the `extra-sources` bit from this PR into it (and replaced the checksums for easier readability), because the other test was already more comprehensive and tested all the algorithms.

gridbugs (approved):

maiste (approved):

PR #10696, opened 2024-07-04
fix(10685): display hint when no action field in rule

### LINKED ISSUE #10685: Point to (alias) when action field is absent
Sometimes, I use `rule` stanzas to consolidate two inner targets, e.g.:

```clojure
(rule
 (alias parent)
 (deps
  (alias child_one)
  (alias child_two)))
```

But I get `Error: field action missing` from Dune, so I end up using something like this:

```clojure
(rule
 (alias parent)
 (deps
  (alias child_one)
  (alias child_two))
 ; empty action
 (action (progn)))
```

I wonder why is the `action` field required? Is there anything wrong with having a rule with no action, but just putting together a bunch of `deps`? And as a side question: is there a better way for empty actions than using `progn`?

maiste:
This PR allows `dune` to display a hint when trying to use `rule` without an `action` stanza.

Closes #10685

### DIFF
--- a/src/dune_rules/stanzas/rule_conf.ml
+++ b/src/dune_rules/stanzas/rule_conf.ml
@@ -158,10 +158,10 @@ let long_form =
        | Some action -> action
        | None ->
          let hints =
-           if List.is_non_empty aliases
-           then
+           if List.is_empty aliases
+           then []
+           else
              [ Pp.text "You can use the (alias) stanza to add dependencies to an alias." ]
-           else []
          in
          field_missing ~hints loc "action"
      in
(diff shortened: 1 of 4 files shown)


src/dune_rules/stanzas/rule_conf.ml
```diff
@@ -153,6 +153,18 @@ let long_form =
              ];
          Standard, true
      in
+     let action =
+       match action_o with
+       | Some action -> action
+       | None ->
+         let hints =
+           if List.is_non_empty aliases
```
rgrinberg, line 161: you can reverse this condition is `List.is_empty`

test/blackbox-tests/test-cases/github10685.t
```diff
@@ -0,0 +1,55 @@
+
```
emillon, line 1:
```suggestion
```

emillon (approved):

PR #10699, opened 2024-07-05
fix: pass -w -53 in wrapped_compat modules

### LINKED ISSUE #10689: ocaml 5.2: warning 53 with (wrapped (transition))
When using `(wrapped (transition))` with ocaml 5.2, warning 53 (misplaced-attribute) triggers. This seems to be a bug in the ocaml compiler (see https://github.com/ocaml/ocaml/issues/13054#issuecomment-2205839590) but we might be able to work around it in dune.

emillon:
On OCaml 5.2, this feature triggers warning 53, which is justified.

Quoting @nojb:

> Dune first builds bar.cmi and bar.cmo from bar.ml-gen and this part
> works well (bar.ml-gen contains a top-level attribute but no .cmi).
> However, when it is the turn of the native-code compiler, bar.cmi
> already exists (and contains the expected deprecation attribute), so the
> top-level attribute in Bar.ml-gen is correctly ignored by the compiler,
> hence the warning.

Fixes #10689

### DIFF
--- a/src/dune_rules/compilation_context.ml
+++ b/src/dune_rules/compilation_context.ml
@@ -304,7 +304,11 @@ let for_module_generated_at_link_time cctx ~requires ~module_ =
   }
 ;;
 
-let for_wrapped_compat t = { t with includes = Includes.empty; stdlib = None }
+let for_wrapped_compat t =
+  (* See #10689 *)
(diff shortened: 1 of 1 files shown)


nojb (approved):
LGTM (if it works :))

PR #10705, opened 2024-07-09
test: fix oneshot webserver for large binary files

gridbugs:
Fixes several issues where the oneshot webserver used for testing package downloading would exit before the transfer was complete. This involves transefering data as byte arrays rather than as strings, and properly shutting down the socket before exiting the server.

The only remaining use case for the `respond_string` function was in sending 404 errors, so I've replaced it with a specialized function that only sends 404 errors.

### DIFF
--- a/test/http/http.ml
+++ b/test/http/http.ml
@@ -29,24 +29,61 @@ module Server = struct
     let descr, _sockaddr = Unix.accept ~cloexec:true t.sock in
     let out = Unix.out_channel_of_descr descr in
     f out;
+    Out_channel.flush out;
+    Unix.shutdown descr Unix.SHUTDOWN_SEND;
     close_out out
   ;;
 
   let stop t = Unix.close t.sock
 
-  let respond out ~status ~content =
-    let status =
+  let header status mime content_length =
+    let status_string =
       match status with
       | `Ok -> "200 Ok"
       | `Not_found -> "404 Not Found"
     in
-    let content_length = String.length content in
-    Printf.fprintf out "HTTP/1.1 %s\nContent-Length: %d\n\n" status content_length;
-    Out_channel.output_string out content
+    let mime_string =
+      match mime with
+      | `Text -> "text/plain"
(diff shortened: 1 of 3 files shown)


gridbugs:
The context for this change is that it fixed the issue downloading package tarballs in tests, as can be seen in https://github.com/ocaml/dune/pull/10706.

test/http/http.ml
```diff
@@ -29,24 +29,61 @@ module Server = struct
...
+    In_channel.with_open_bin file (fun in_channel ->
+      let length = In_channel.length in_channel |> Int64.to_int in
+      let out_fd = Unix.descr_of_out_channel out in
+      let header_bytes = Bytes.of_string (header `Ok `Binary length) in
+      send_bytes out_fd header_bytes (Bytes.length header_bytes);
+      let buf_size = 4096 in
+      let buf = Bytes.create buf_size in
+      let pos = ref 0 in
+      try
+        while true do
```
Leonidas-from-XIV, line 81: I think it would be nicer to make the `while` loop condition the thing that makes the loop termnate instead of building `break` out of exceptions.

Leonidas-from-XIV (commented):
What was the reason the previous code did not work? Was there some issue with `output_string`?

test/http/http.ml
gridbugs: Sounds good to me.

gridbugs:
> What was the reason the previous code did not work? Was there some issue with `output_string`?

The main problem was actually that we weren't shutting down the socket properly, which meant that the server process would sometimes exit before the file had been completely transfered. I added the binary transfer stuff as I thought it was related and decided to leave it in as it's still more correct this way.

Leonidas-from-XIV (approved):

rgrinberg:
I think this fix is generally incorrect and needs to reverted. The main issue is that we're just not flushing the buffered writes. There's no need to go and use low level socket API's to fix this issue.

PR #10706, opened 2024-07-09
Fix rule target conflict with extra source

gridbugs:
This is a quick fix that addresses the "Error: Multiple rules generated for <target>" when building a package with the extra-source field. This isn't expected to work in cases where a package has multiple extra-source entries, which will come in a later change.

Note that the test is currently complaining about `curl` returning an unexpected result. This is fixed in https://github.com/ocaml/dune/pull/10705, so we'll merge that first.

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -1589,7 +1589,7 @@ let source_rules (pkg : Pkg.t) =
         | `Directory_or_archive src ->
           loc, Action_builder.copy ~src:(Path.external_ src) ~dst:extra_source
         | `Fetch ->
-          let rule = Fetch_rules.fetch ~target:pkg.paths.source_dir `File fetch in
+          let rule = Fetch_rules.fetch ~target:extra_source `File fetch in
           loc, rule
       in
       Path.build extra_source, rule)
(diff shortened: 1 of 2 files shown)


Leonidas-from-XIV (approved):
Looks good to me, so once #10705 is merged and this is rebased fixing the test I recommend merging.

gridbugs:
Hmm looks like the error I thought I fixed in https://github.com/ocaml/dune/pull/10705 is still happening, though I can no longer reproduce it locally. I'll rerun the test to help get a sense of whether it fails non-deterministically and see if I get the failure to happen on my machine tomorrow.

Leonidas-from-XIV:
Yes, I have the same issue, it works for me locally on Fedora and it seems to work on the mac and windows CI builds, but somehow Ubuntu fails. I'll investigate as well.

gridbugs:
Just had a thought, that since the problem with the oneshot webserver only happens when the file is sufficiently large, we can avoid the issue for the etrxa-sources test by just not including the .git directory in the tarball, which will reduce its size substantially. I've pushed a commit to this effect. Unless the problem with the webserver is easy to find, I think we should just make an issue and repro for it and move on.

Leonidas-from-XIV:
Good thinking; looks like this is an ok workaround that lets us move forward. While it would be certainly nice to fix this particular issue (and as I noticed, #10714), this passes CI (and is unlikely to break randomly unless the size of the tar file increases).

PR #10708, opened 2024-07-09
test: Exercise the case that there are multiple `extra-sources`

Leonidas-from-XIV:
As mentioned by @gridbugs in #10706 the fix might solve #10691 as it only ever tests the case where there is one `extra-source` but there is no such limitation in OPAM so this test adds a second `extra-source`.

In addition it makes sure that the extra-sources are applied in the right order as they depend upon each other, otherwise patching should fail.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/extra-sources.t
+++ b/test/blackbox-tests/test-cases/pkg/extra-sources.t
@@ -55,17 +55,24 @@ First we need a project that will have the patch applied:
   > let msg = "Patch successfully applied"
   > EOF
   $ git diff > ../required.patch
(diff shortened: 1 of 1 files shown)


gridbugs (approved):

gridbugs:
Looks like the fix in https://github.com/ocaml/dune/pull/10706 also works for this case. I missed the detail that each extra source is downloaded into a separate dir based on its name.

Leonidas-from-XIV:
Even better, if fixing this was just fixing the typo its a good outcome. I'll rebase the test (because I think checking multiple sources is a good idea in any case, even if it works).

PR #10709, opened 2024-07-09
doc: use describe in library.rst

emillon:
This replaces a bullet list with proper markup.


Leonidas-from-XIV (approved):
I assume the text didn't change aside from capitalization. It seemed that way at least for the first few items, so by induction I assume that this is true for the rest. :upside_down_face: 

Yeah, looks nicer in the rendered documentation. And might work better with search?

emillon:
Correct, I haven't changed the content.

> And might work better with search?

Maybe not immediately. It might be possible to turn the "library" page into an index like https://dune.readthedocs.io/en/stable/reference/actions/index.html linking to individual pages with examples, but there's a tradeoff between large page vs more clicks.

PR #10710, opened 2024-07-09
Enable dune cache by default

ElectreAAS:
This PR enables the dune cache by default, without having to use `DUNE_CACHE=enabled`. The idea is to lay a groundwork to enable caching of actions that are safe to cache and have them be cached out of the box in the future.

To do so the default of whether actions can be cached is moved a new `Clflag` (`Clfags.can_go_into_shared_cache_default`). This value is set to `false` as most actions can't be cached, thus it could break code to enable caching of them by default. This flag can be overridden via the `DUNE_CACHE_RULES=enabled` environment variable/toggle, which will then allow to cache all actions (as it did before when you enabled `DUNE_CACHE=enabled`).

Some adjustments for the creation of the cache directories needed to be made if Dune is run in environments where the cache directory cannot be written (e.g. when packaging for Nix or Debian), it will output a warning in such case and continue with the cache disabled as before. The warning can be silenced by specifically setting `DUNE_CACHE=disabled`.

The tests that require the rules to be cached have been updated to include `DUNE_CACHE_RULES=enabled` and exhibit the same behavior wrt. caching, hashes and trimming as before.

In a future PR, some of the actions to that are deemed to be safe for caching (e.g. downloading package sources) will have their default value set as such.


src/dune_config_file/feature_flags.ml
```diff
@@ -0,0 +1 @@
+let cache_enabled_by_default = true
```
Leonidas-from-XIV, line 1: For `main` this should be `false`.

test/blackbox-tests/test-cases/default_cache.t
```diff
@@ -0,0 +1,32 @@
...
+  >  (name foo))
+  > EOF
+
+  $ cat > foo.ml << EOF
+  > let f x y = x + y
+  > EOF
+
+Set up cache directory
+  $ rm -rf /tmp/dune_test_cache
+  $ mkdir /tmp/dune_test_cache
```
Leonidas-from-XIV, line 15: Do not create files outside the test directory. $(pwd)/dune_test_cache` is a better choice.

test/blackbox-tests/test-cases/default_cache.t
```diff
@@ -0,0 +1,32 @@
...
+
+  $ DUNE_CACHE=disabled DUNE_CACHE_ROOT=/tmp/dune_test_cache dune build
```
Leonidas-from-XIV, line 17: Probably makes sense to export `DUNE_CACHE_ROOT` so it doesn't need to be specified every time.

test/blackbox-tests/test-cases/default_cache.t
```diff
@@ -0,0 +1,32 @@
+The dune cache should be enabled by default
```
Leonidas-from-XIV, line 1: I think our tests usually use `kebab-case` for the test cases.

src/dune_config_file/feature_flags.ml
ElectreAAS: I thought the point was to enable by default on `main`?
Or should I open a separate PR targeting the toolchains branch setting this up to true?

Leonidas-from-XIV: I don't think we can just enable the dune cache by default on `main`, because in such case there would be no point to have a feature flag.

ElectreAAS: I updated the feature flag. At this point I don't see how we can satisfy all these criteria:
- all unstable features are guarded by feature flags
- these features have tests that actually test the feature
- when we want to enable this feature on main, the commit to do so will only need a `- let x = false` `+ let x = true`
Maybe there's something I missed...
This PR should be mergeable in the meantime__

src/dune_config_file/feature_flags.ml
```diff
@@ -0,0 +1 @@
+let cache_enabled_by_default = false
```
rgrinberg, line 1: What is the point of "feature flags" when compared to the runtime configuration we already have? Is it that they are only set at compile time?

rgrinberg:
Before this default is flipped, we need to have a story for how this interact with user rules. As a first approximation, all user rules are unsafe to cache this way. So users should be able to opt-in to enable caching for their own rules.

maiste:
FMU, as this PR is only changing the default, if the user has a configuration, it will take it instead of the default. Or am I missing something? :confused:

rgrinberg:
The problem lies in that most users do not have such a configuration. Moreover, most project contain user rules of the form:
```
(rule
 ..)
```
That are absolutely not safe to cache. Therefore, flipping the default like is done in this PR is highly likely to cause a lot of unintended breakage.

src/dune_config_file/feature_flags.ml
ElectreAAS: Update: I removed the feature flag as this is something that we want to merge on main

ElectreAAS:
Incorporated (very helpful) comments by Rudi on #10789, we now have a different environment variable named (for now) DUNE_CACHE_USER_RULES defaulting to false, the old DUNE_CACHE is now on by default, and `Action.Full.t` are now uncacheable by default.
What remains to be done is picking which of the 36 uses of `Action.Full.make` can be safely cached.

ElectreAAS:
The intended behaviour is now, as far as I understand correctly _(Writing mostly for my future self here)_ :
By default the cache is now available, however most stuff won't go into it
The only things that do go into it are the uses 'blessed' by that `can_go_in_shared_cache` parameter (looking at package rules obviously, but maybe more later?)
What the users can do:
- disable the cache if they don't want the new default (is that considered a breaking change?)
- enable caching anything not 'blessed' by using the new env variable

rgrinberg:
> disable the cache if they don't want the new default (is that considered a breaking change?)

I don't think so. Since disabling the cache will only affect the package management rules which aren't stable.

Leonidas-from-XIV:
> disabling the cache will only affect the package management rules which aren't stable.

I think I am misunderstanding something because as we see in the tests, a lot of them fail if `DUNE_CACHE_USER_RULES=enabled` is not set, so to get the previous behavior of `DUNE_CACHE=enabled` you also need `DUNE_CACHE_USER_RULES=enabled`. This seems like a breaking change?

nojb:
> That are absolutely not safe to cache.

@rgrinberg can you remind me why user rules are not safe to cache?

Leonidas-from-XIV:
I don't think the Nix failures have anything to do with the PR. I'll add the overrides for rules from package management in a second PR so this is just laying down the groundwork for it.

rgrinberg:
As mentioned on slack, the error is relevant. Dune is trying to create the cache directory at $HOME and nix is rightfully forbidding that.

rgrinberg:
One way to work around this would be to disable the cache if the cache dir isn't writeable.

Maybe just emit a warning?

Leonidas-from-XIV:
@rgrinberg I've added some code that detects if the directories can't be created, it disables the cache and warns.

I had to add a case to `mkdir_p_result` which has the downside that `mkdir_p_result` now has success and error cases (mirroring `mkdir_result`). I think that's not super nice, as it probably would be better to split the cases into an "there's a directory" (`Ok`) and "couldn't be created" (`Error`) category but on the other hand I didn't want to overengineer this.

src/dune_cache/config.ml
```diff
@@ -43,4 +43,5 @@ type t =
   | Enabled of
       { storage_mode : Dune_cache_storage.Mode.t
       ; reproducibility_check : Reproducibility_check.t
+      ; cache_user_rules : bool
```
rgrinberg, line 46: Whether the user rules are cached is not something that the cache itself should be aware of. It's a property of our rules and how we choose to create them.

src/dune_cache/shared.ml
```diff
@@ -65,28 +65,27 @@ struct
...
-    | Enabled { storage_mode = mode; reproducibility_check } ->
-      (match Config.Reproducibility_check.sample reproducibility_check with
-       | true ->
-         (* CR-someday amokhov: Here we re-execute the rule, as in Jenga. To make
-            [check_probability] more meaningful, we could first make sure that
-            the shared cache actually does contain an entry for [rule_digest]. *)
-         Fiber.return (Hit_or_miss.Miss Miss_reason.Rerunning_for_reproducibility_check)
-       | false -> try_to_restore_from_shared_cache ~mode ~rule_digest ~targets)
+    | Enabled { storage_mode = mode; reproducibility_check; cache_user_rules } ->
+      if can_go_in_shared_cache || cache_user_rules
```
rgrinberg, line 72: Again, the cache shouldn't be aware of user rules.

src/dune_cache_storage/layout.ml
```diff
@@ -82,7 +82,12 @@ let value_storage_dir = Versioned.value_storage_dir Version.Value.current
 let value_path = Versioned.value_path Version.Value.current
 
 let create_cache_directories () =
-  List.iter
-    [ temp_dir; metadata_storage_dir; file_storage_dir; value_storage_dir ]
-    ~f:(fun path -> ignore (Fpath.mkdir_p (Path.to_string path) : Fpath.mkdir_p_result))
+  [ temp_dir; metadata_storage_dir; file_storage_dir; value_storage_dir ]
+  |> List.fold_left ~init:(Ok ()) ~f:(fun result path ->
+    match result with
+    | Error _ as e -> e
```
rgrinberg, line 88: A little suspect to use a fold here but to terminate on the very first failure

src/dune_file_watcher/dune_file_watcher.ml
```diff
@@ -397,7 +397,9 @@ let prepare_sync () =
   | Cleared -> ()
   | Directory_does_not_exist ->
     (match Fpath.mkdir_p dir with
-     | Already_exists | Created -> ())
+     | Already_exists | Created -> ()
+     | Permission_denied ->
+       Code_error.raise "Directory could not be created" [ "dir", Dyn.string dir ])
```
rgrinberg, line 402: Isn't it possible for the user to run into this?

src/dune_rpc_impl/server.ml
```diff
@@ -94,7 +94,10 @@ module Run = struct
               in
               Registry.Config.register registry_config dune
             in
-            let (_ : Fpath.mkdir_p_result) = Fpath.mkdir_p (Filename.dirname path) in
+            (match Fpath.mkdir_p (Filename.dirname path) with
+             | Created | Already_exists -> ()
+             | Permission_denied ->
+               User_error.raise [ Pp.textf "Could not create directory %s" path ]);
```
rgrinberg, line 100: Don't think this new error is better than the old.

rgrinberg:
Nix is still showing a failure. I think you can write a test for this issue by the way (in a separate PR preferably). You can set XDG_CACHE_HOME to some non-existent or read only directory and you'll be able to demonstrate the failure.

src/dune_cache_storage/layout.ml
Leonidas-from-XIV: I want to only continue if all of them have been successfully created. The other option I could think of would be to `map` and then something like `Result.List.all_unit` but there is no point in continuing to create folders if one of them has failed. Likewise, I also don't think there is a point to delete them again if all but the last have succeeded.

(Looking at `Stdune.Result` we have `Result.List.fold_left` which I guess inlines the match on the result, so that can be made a bit nicer but it does essentially still the same).

rgrinberg: I think you want `Result.List.iter` here. If we don't have that, you can add it.

I'm not sure what you mean by inlining, but my point is that using `List.fold_left` will always traverse the entire list even though you trying to step after the first failure.

```ocaml
let x = [ false ; true ]
let f x = if x then Error () else Ok ()
let () = List.fold_left x ~init:(Ok ()) ~f:(fun x -> print_endline "foo"; f x)
let () = Result.List.fold_left x ~init:() ~f:(fun x -> print_endline "foo"; f x)
```

The `List.fold_left` prints `foo` twice, while `Result.List.fold_left` prints `foo` only once. I think you intend to have the latter behavior.

Leonidas-from-XIV: Sure, but that's why the first statement in the body is a `match` so that a previous `Error` will short-circuit into returning `Error` and not call any more `mkdir`.

Ah yes, `Result.List.iter` is exactly what I wanted, thus I can avoid match in `~f`.

src/dune_cache/config.ml
Leonidas-from-XIV: Good point, it the user rules should just state `can_go_in_shared_cache = true` in that case.

otherlibs/stdune/src/fpath.mli
```diff
@@ -5,12 +5,14 @@ type mkdir_result =
   | Created (** The directory was created. *)
   | Missing_parent_directory
   (** No parent directory, use [mkdir_p] if you want to create it too. *)
+  | Permission_denied (** Could not create the directory *)
```
rgrinberg, line 8: I check our code internally, and it turns that we have many uses of this function that would need to be modified. Instead of modifying this function, can you add a new function such as `val mkdir_p_with_permissions : (mkdir_p_result, [`Permission_denied]) result` and use that? That would make syncing a lot simpler for us.

src/dune_engine/build_config.ml
```diff
@@ -110,10 +110,19 @@ let set
...
         ~f:(fun ((ctx : Build_context.t), ctx_type) -> ctx.name, (ctx, ctx_type)))
   in
-  let () =
+  let cache_config =
     match (cache_config : Dune_cache.Config.t) with
-    | Disabled -> ()
-    | Enabled _ -> Dune_cache_storage.Layout.create_cache_directories ()
+    | Disabled -> cache_config
+    | Enabled _ ->
+      (match Dune_cache_storage.Layout.create_cache_directories () with
```
rgrinberg, line 117: Can you move this check to the caller of this function? This should be either in the rules or in the command line (if my memory serves correctly)

test/blackbox-tests/test-cases/dune-cache/mode-hardlink.t
```diff
@@ -35,9 +36,9 @@ never built [target1] before.
   $ dune build --config-file=config target1 --debug-cache=shared,workspace-local \
   >   2>&1 | grep '_build/default/source\|_build/default/target'
   Workspace-local cache miss: _build/default/source: never seen this target before
-  Shared cache miss [3ad1761950da90e34e52b4c065db1504] (_build/default/source): not found in cache
+  Shared cache miss [e2f27218eba545659418b9c738e421ff] (_build/default/source): not found in cache
```
rgrinberg, line 39: Is the digest different because you bumped the version? If so, I'd recommend moving the version bump into a separate PR. Your feature on its own should not modify the digests, otherwise there's a bug somewhere.

Leonidas-from-XIV: I assumed the hashes changed due to the change of the default value of `can_go_in_shared_cache` in https://github.com/ocaml/dune/pull/10710/commits/f9c0af97001967e08f634f3e76aa88df38b4d737 hence I promoted these changes.

But looking at it now I think the default should be `true` and just in the case of user rules it should be set to false (or whatever the configuration says) and they should, for the most part, flip back to the previous digests.

src/dune_cache/shared.ml
```diff
@@ -65,28 +65,27 @@ struct
...
+      else Fiber.return (Hit_or_miss.Miss Miss_reason.Cannot_go_in_shared_cache)
   ;;
 
   let lookup ~can_go_in_shared_cache ~rule_digest ~targets
     : Digest.t Targets.Produced.t option Fiber.t
     =
     let open Fiber.O in
-    let+ result =
-      match can_go_in_shared_cache with
-      | false -> Fiber.return (Hit_or_miss.Miss Miss_reason.Cannot_go_in_shared_cache)
```
rgrinberg, line 87: What does this change do? As far as I can tell, it just makes the function return `Cache_disabled` rather than `Cannot_go_in_shared_cache`. Does that have an impact?

test/blackbox-tests/test-cases/dune-cache/mode-hardlink.t
rgrinberg: Shouldn't the tests enable the caching of user rules because they rely on user rules? In which case, I would assume the digests to stay the same.

src/dune_rules/simple_rules.ml
```diff
@@ -59,15 +59,25 @@ let interpret_and_add_locks ~expander locks action =
   | locks -> Action_builder.map action ~f:(Action.Full.add_locks locks)
 ;;
 
+let mark_as_unsafe_for_shared_cache action =
+  let open Action_builder.O in
+  let+ action = action in
+  (* TODO: need to set can_go_in_shared_cache to the config value *)
+  Action.Full.add_can_go_in_shared_cache false action
```
rgrinberg, line 66: I think once this respects your configuration option, the digests should be restored.

src/dune_cache_storage/layout.mli
```diff
@@ -12,8 +12,11 @@ open Import
 val root_dir : Path.t
 
 (** Create a few subdirectories in [root_dir]. We expose this function because
-    we don't want to modify the file system when the cache is disabled. *)
-val create_cache_directories : unit -> unit
+    we don't want to modify the file system when the cache is disabled.
+
+    Returns whether creation has succeeded or in the case of error which directory
+    could not be created. *)
+val create_cache_directories : unit -> (unit, Path.t) result
```
rgrinberg, line 19: I think the error should include the underlying unix error so that you can actually print what went wrong.

src/dune_cache_storage/layout.ml
```diff
@@ -82,7 +82,10 @@ let value_storage_dir = Versioned.value_storage_dir Version.Value.current
...
 let create_cache_directories () =
-  List.iter
-    [ temp_dir; metadata_storage_dir; file_storage_dir; value_storage_dir ]
-    ~f:(fun path -> ignore (Fpath.mkdir_p (Path.to_string path) : Fpath.mkdir_p_result))
+  [ temp_dir; metadata_storage_dir; file_storage_dir; value_storage_dir ]
+  |> Result.List.iter ~f:(fun path ->
+    match Fpath.mkdir_p (Path.to_string path) with
+    | Already_exists | Created -> Ok ()
+    | (exception Unix.Unix_error (EACCES, _, _))
+    | (exception Unix.Unix_error (EROFS, _, _)) -> Error path)
```
rgrinberg, line 90: I looked at the man page for `mkdir` and the amount of error codes possible is massive. What about just catching every unix error? All of them can have the same fallback.

test/blackbox-tests/test-cases/dune-cache/mode-hardlink.t
Leonidas-from-XIV: You're right. I've bumped the version to 3.17 to be able use the `cache-user-rules` stanza, to be consistent with the tests using sexps to configure the cache but setting `DUNE_CACHE_USER_RULES=enabled` and making actions cacheable by default again restores the hashes to their previous values.

src/dune_rules/simple_rules.ml
Leonidas-from-XIV: They are now, indeed. Reading the config out of the workspace configuration which seems to me fairly ok.

src/dune_engine/action.mli
```diff
@@ -151,7 +151,7 @@ module Full : sig
   val make
     :  ?env:Env.t (** default [Env.empty] *)
     -> ?locks:Path.t list (** default [[]] *)
-    -> ?can_go_in_shared_cache:bool (** default [true] *)
+    -> ?can_go_in_shared_cache:bool (** default [false] *)
```
rgrinberg, line 154: Comment is stale?

I'm trying to think of how to make this change less painful for you, and I think that without modifying this default. It's going to be rather painful. The issue is that what we call "user rules" is not just whatever is present in simple_rules, but it includes the ocaml compilation rules, C/C++ rules, cram, ocamlformat, etc. So for all of those rules you want `~can_go_in_shared_cache:DUNE_CACHE_RULES`. The only rules for which you want `can_go_in_shared_cache:true` are those in pkg_rules and fetch_rules.

So I think that you indeed should modify this default. In particular, it should be set to `DUNE_CACHE_RULES`.  This will require setting some sort of mutable reference in `Dune_engine.Clflags.can_go_in_shared_cache_default` and then setting it based on this environment variable.

It's somewhat intrusive, but I don't see how else to do this without pulling one's hair out.

src/dune_rules/simple_rules.ml
```diff
@@ -59,15 +59,27 @@ let interpret_and_add_locks ~expander locks action =
...
 let add_user_rule
   sctx
   ~dir
   ~(rule : Rule_conf.t)
-  ~(action : _ Action_builder.With_targets.t)
+  ~(action : Action.Full.t Action_builder.With_targets.t)
   ~expander
   =
+  let open Memo.O in
+  let* config = Workspace.workspace_config () in
```
rgrinberg, line 76: Note that once the cache setting is set, it can no longer be modified in watch mode. There's a few options like that in the workspace that are incompatible with the watch mode.

So this is not better than just using a global var. It's slightly confusing in fact as it pretends to be something that it's not (a value that can change in watch mode)

src/dune_engine/build_config.ml
Leonidas-from-XIV: I did that in 984f8904b but that means that the call to `Dune_cache_storage.Layout.create_cache_directories` now lives in the `dune_rules`. Is that the right place?

I didn't want to make `Build_config.set` return an error that the caller has to understand and call `set` again with the cache disabled.

src/dune_engine/action.mli
Leonidas-from-XIV: I've updated it to read the `ref` which is filled from the environment. As it it shouldn't be reset at runtime, I will remove the `cache-user-rules` field, since it can't be changed in watch mode anyway.

src/dune_cache_storage/layout.ml
Leonidas-from-XIV: Ok, I'm passing the error to the caller so the caller can display the error reason (and the path that failed).

src/dune_engine/build_config.ml
rgrinberg: > I did that in https://github.com/ocaml/dune/commit/984f8904b790de94abe8bf4d445b6d6d5428d3e6 but that means that the call to Dune_cache_storage.Layout.create_cache_directories now lives in the dune_rules. Is that the right place?

Ideally, it would live in the command line but having it live in the rules is acceptable for now.

> I didn't want to make Build_config.set return an error that the caller has to understand and call set again with the cache disabled.

Yeah, `Build_config.set` should never fail. It is only used to set some hooks to configure the engine and the rules.

src/dune_cache/shared.ml
rgrinberg: You might have missed this comment? Don't think the changes in this file are necessary.

src/dune_engine/build_config.ml
```diff
@@ -110,11 +110,6 @@ let set
         contexts
         ~f:(fun ((ctx : Build_context.t), ctx_type) -> ctx.name, (ctx, ctx_type)))
   in
-  let () =
-    match (cache_config : Dune_cache.Config.t) with
```
rgrinberg, line 114: @snowleopard. Is moving this to the caller acceptable? Seems like this is an improvement to allow the caller to create the cache directories and handle any possible errors.

src/dune_cache/shared.ml
Leonidas-from-XIV: I think the code used to read whether to override the value in the config, but I agree that with the current state this seems to not be necessary. I've reverted it and will rebase and squash it later.

rgrinberg:
The changes look fine from my end. This needs a look from @snowleopard though.

Leonidas-from-XIV:
Glad to hear; I added @snowleopard as a reviewer.

Leonidas-from-XIV:
That's a good point. I have updated the description with where the PR has ended up. That should make it clearer what it does on a higher level and why the changes have been done the way they are.

nojb:
> To do so the default of for whether Actions can be changed is moved a new `Clflag` to set the default and set to `false` as most Actions can't be changed, thus it could break code to enable caching of them. This flag can be overridden via the `DUNE_CACHE_RULES` environment variable/toggle, which will then cache everything (as it did before).

Sorry, but I can't parse this paragraph, something appears to have been lost along the way :)

Leonidas-from-XIV:
@nojb Sorry, my fingers somehow wrote "changed" when my mind meant "cached". I've fixed these, rewrote it a bit and I hope it is more understandable now.

nojb:
> I've fixed these, rewrote it a bit and I hope it is more understandable now.

Thanks, it is clearer now.

src/dune_rules/main.ml
```diff
@@ -75,6 +75,32 @@ let init
       let download ~rule_digest:_ = Fiber.return ()
     end)
   in
+  let cache_config =
```
snowleopard, line 78: It seems odd for this code to live in `src/dune_rules`, since these aren't rules. Maybe it should be in the command line somewhere.

snowleopard: (Not a blocker for this PR, just an observation.)

ejgallego:
Hi folks, should a change log be added to this PR?

ElectreAAS:
> Hi folks, should a change log be added to this PR?

Indeed, see #10913 for the changelog entry PR

samoht:
> As a first approximation, all user rules are unsafe to cache this way.

@rgrinberg do you know which packages have this issue? I've been running `(cache enabled)` for a while now without problems. Could we have an easier way to opt-in than an environment variable? That seems quite a big regression, especially when the `rule` runes are at the root of the dependency graph on some projects!

rgrinberg:
Why is it a big regression? You set an environment variable and move on.

I don't know which packages are affected which is why I cannot just enable this option by default. A good way to look for such packages would be to turn on sandboxing and see which rules fail.

ejgallego:
> Why is it a big regression? You set an environment variable and move on.

Setting env vars across different build setups can be pretty time-consuming and often not trivial.

I agree with @samoht that this is a regression for users.

snowleopard:
I also agree that this is a regression. I hope "set some new environment variable and move on" is not a typical experience when upgrading to a new version of Dune. We should find a better way!

rgrinberg:
We tried to find a better way but this is all we could come up with.

The cache has to be enabled for package management, but doing so while opting in to cache all user rules would be even more disruptive. At least this way the disruption is limited to the small subset of users using the cache.

Suggestions are welcome.

snowleopard:
Well, package management is a new feature, and it seems fair to require early adopters to, say, set `DUNE_CACHE_RULES=false`. And tell them to bug the rule authors of rules that require that.

snowleopard:
In general, penalising rules that do be the right thing seems surprising. Surely, the incentives should go the other way around.

rgrinberg:
> Well, package management is a new feature, and it seems fair to require early adopters to, say, set DUNE_CACHE_RULES=false. And tell them to bug the rule authors of rules that require that.

This suggestion is with conjunction to enabling the dune cache by default?

It might be worth recalling why we never enabled the cache by default in the first place. Our own OCaml rules had some correctness issues we never got to the bottom of (around C compilation and linking IIRC). User rules were in an even worse spot. Those were mostly unsandboxed and thus often broken. I don't think the situation has changed much since.

I don't want to punish users that write correct rules, but I also know it would be way too optimistic to assume such rules are correct by default either.

People who enabled the cache are a minority of advanced users, so I think putting the burden of maintaining config on them is more fair than on the users of package management. Some of those are going to be beginners that are just getting started with dune.

samoht:
> People who enabled the cache are a minority of advanced users, so I think putting the burden of maintaining config on them is more fair than on the users of package management.

I've been using dune-cache extensively for years now. The subset of the ecosystem I am working with does not seem to have any broken rules. So I'm surprised that you are so conservative here (even if I understand the rationale, I'm not sure I agree with the proposed workaround). In any case, I would much rather edit ~/.config/dune/config` once and for all with `(dune_super_cache enabled)` than remember to set up an env variable in my shell :-)

ejgallego:
Another thing to consider is what will happen to `ocaml` github action.

IIANM it uses the cache, so it would require adaptation?

Indeed there is large use of dune cache (and it is often why Coq people migrate from make to dune, to have a cache), so IMHO the user experience of this change could be not super good, if that silently stops working.

There is now a changelog entry proposed tho.

Maybe a good compromise would be to lead towards a setup more friendly to the "status quo" of users, we could have a setup such that:

- `(cache enabled)`: old behavior, all rules are cached
- `(cache safe)`: new behavior, only cache "safe" rules

Dune package management could emit a warning if users are running with `(cache enabled)`.

WDYT folks?

ElectreAAS:
I appreciate all the discussion going on here, hearing differing viewpoints is always nice.

I think I am convinced by your arguments @ejgallego & @samoht.

My idea of what to do next would be almost the same as what was proposed above :
1. Any tinkering should be doable both by env vars and by config file editing
2. Make 3 modes for the cache: completely disabled (the old default), completely enabled, and the new "we only cache things we are 100% sure are cacheable for all cases".
3. We move the default to the safe mode, while not requiring users change their existing config. That way it wouldn't break anyone's workflow

ejgallego:
Sounds great to me @ElectreAAS ; maybe we could actually deprecate `(cache enabled)` too, in favor of something like `(cache unsafe)` ?

This can be done properly using `(dune lang ...)` I think, modulo layer issues (I hope not)

I like the idea of using `(dune lang ...)` in config to control this kind of behavior, however it may not be always doable.

rgrinberg:
> Another thing to consider is what will happen to `ocaml` github action.
> 
> IIANM it uses the cache, so it would require adaptation?
> 
> Indeed there is large use of dune cache (and it is often why Coq people migrate from make to dune, to have a cache), so IMHO the user experience of this change could be not super good, if that silently stops working.
> 
> There is now a changelog entry proposed tho.
> 
> Maybe a good compromise would be to lead towards a setup more friendly to the "status quo" of users, we could have a setup such that:
> 
> * `(cache enabled)`: old behavior, all rules are cached
> * `(cache safe)`: new behavior, only cache "safe" rules
> 
> Dune package management could emit a warning if users are running with `(cache enabled)`.
> 
> WDYT folks?

I'm lukewarm about it. I think having the cache enabled or disabled vs selecting which rules are cacheable are orthogonal things and mixing them up in a single config will lead to confusion and lost flexibility down the road. On the other hand, if it really is so difficult to move to a different way to enable the cache for the coq devs (how did you get them on all on board to use the cache in the first place?), then so be it.

ejgallego:
> I'm lukewarm about it. I think having the cache enabled or disabled vs selecting which rules are cacheable are orthogonal things and mixing them up in a single config will lead to confusion and lost flexibility down the road.

That is a very interesting observation. Indeed I assumed the opposite, that enabling the cache and understanding what gets cached were very related concepts.

My thought process is like this: imagine I do `(cache enabled)`, but then no rule gets cached because all my rules are unsafe. That seems pretty confusing to me.

We could indeed split the config in something like `(cache enabled/disabled)` and `(cache_config safe/full)`, however I am not sure how this is better?

What would an example of "confusion" or "lost flexibility" in the scenario look like?

> On the other hand, if it really is so difficult to move to a different way to enable the cache for the coq devs (how did you get them on all on board to use the cache in the first place?), then so be it.

I am not so worried about the Coq devs, that's a small set of users, but more about a lot of users (mainly "industrial") that rely on the cache. In general, this is a de facto breaking change and may require significant fiddling with things like the github ocaml action etc... 

> how did you get them on all on board to use the cache in the first place?

I am not sure I understand the question, I guess once they saw it in action most enabled it, a full Coq compilation is very expensive, so caching pays off immensely.

nojb:
Why not proceed as follows:

- `DUNE_CACHE=enabled` -> cache is fully enabled (like before this PR)
- `DUNE_CACHE=partial` -> cache is enabled but not for user-defined rules (like in this PR)
- `DUNE_CACHE=disabled` -> cache is fully disabled (like before this PR)

Then this PR should amount to changing _the default_ to `DUNE_CACHE=partial`. This is completely backwards compatible (meaning that people who have already set `DUNE_CACHE` do not observe any difference in behaviour), while enabling the Dune cache by default, as in this PR.

rgrinberg:
> but then no rule gets cached because all my rules are unsafe. That seems pretty confusing to me.

That's the situation today, but I think we can start marking some rules as safe to cache gradually. I'm thinking ocamldep rules, user rules that are sandboxed, some of the compilation rules.

> What would an example of "confusion" or "lost flexibility" in the scenario look like?

Here's the main difference: when you mark a rule as unsafe for caching, this bit ends up being included in the action digest. The consequence of doing this is that once this bit is flipped (via the environment variable for example), the action will re-run and the cache entry will be populated (or possibly trimmed if going the other way). On the other hand, enabling/disabling the cache has no effect on the rules themselves. So it can never trigger recompilation.

Maybe this distinction isn't particularly important or the two types of settings can be unified somehow in the future?

As I said, we can go along with your suggestion if this will make transitioning easier for you. It's good enough for us to proceed on the package management side.

> I am not sure I understand the question, I guess once they saw it in action most enabled it, a full Coq compilation is very expensive, so caching pays off immensely.

My point is that you somehow told the team that a config file needs to be updated to enable the cache. Well, you can tell them that a new config is needed in the same way. Or, if you have some makefile you all share, you can just set the config there.

rgrinberg:
It sounds quite similar to Emilio's proposal. Seems fine as well.

nojb:
> It sounds quite similar to Emilio's proposal. Seems fine as well.

See #10944

ejgallego:
> My point is that you somehow told the team that a config file needs to be updated to enable the cache. Well, you can tell them that a new config is needed in the same way. Or, if you have some makefile you all share, you can just set the config there.

Indeed, this is why I was requesting for some setting in `.config`, vs an env var.

In the Coq world there a multitude of different CI / testing deployements, and not only Coq devs, but many other users would be affected by this change.

Unfortunately there is nothing close to a homogeneous Coq build CI setup (we are trying to improve that, check your mail if you are interested ;)), but indeed I don't know how to write a guide on adding a new env var effectively, there are just too many combinations of shells and dev setups, it seems to me. The config file is nicer to implement I feel.

A bit orthogonal to this is the topic of backwards compat, IMO dune has been doing superb on this aspect, so changes like this are better controlled by `(dune lang ...)` version so users can control breakage better.

I am often bitten by rules that are cached wrongly, so indeed it is great to progress on this front, so we get better tools to debug

rgrinberg:
Another thing that you might find worth doing is to mark all the coq rules as being explicitly safe. If you're confident about those, might as well enable them under `enabled-except-user-rules` as well.

ejgallego:
> Another thing that you might find worth doing is to mark all the coq rules as being explicitly safe. If you're confident about those, might as well enable them under `enabled-except-user-rules` as well.

Indeed thanks! I've added this to the list of TODOs to the `(coq lang 1.0)` effort.

One reason we didn't mark them as safe, is that while Coq rules are "cache safe", they are not "sandbox" safe, due to the "hack" of including hashes of deps in .vo files.

PR #10712, opened 2024-07-10
ci: explicitly define mingw and msvc builds

emillon:
This replaces the "default" windows-latest (which seems to default to MSVC) by explicit MSVC and mingw variants.

See #10715


maiste (approved):
LGTM :+1:

PR #10714, opened 2024-07-10
tests: Automatically shut down the one-shot server after some time

Leonidas-from-XIV:
The one-shot server automatically terminates after serving one single HTTP request, which is useful for tests. Unfortunately, if no such request happens, e.g. due to something else in a test failing so the request never happens the tests currently leave a one-shot werserver instance running indefinitely. This has happened to me while testing #10706, where I had dozens of leftover servers running on random ports for a while.

It is hard to know whether a test is still going to happen or not but this PR adds a timeout for the server that if nobody connects for a set amount of time the server will terminate on its own.

This limits the left-behind servers which will automatically terminate on their own after a while when nobody connects.

### DIFF
--- a/test/http/http.ml
+++ b/test/http/http.ml
@@ -6,6 +6,8 @@ module Server = struct
     ; addr : Unix.sockaddr
     }
 
+  let auto_shutdown_seconds = 30.
+
   type session = out_channel
 
   let make addr =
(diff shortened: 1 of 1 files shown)


test/http/http.ml
```diff
@@ -1,5 +1,7 @@
 open Stdune
 
+let server_auto_shutdown = 30.
+
```
maiste, line 4: Isn't it better to have it in the `Server.t` type? It could be pass as a value to make with a fixed default value. WDYT?

maiste (approved):
One quick question, but otherwise LGTM :+1:

test/http/http.ml
```diff
@@ -6,6 +6,8 @@ module Server = struct
     ; addr : Unix.sockaddr
     }
 
+  let server_auto_shutdown = 30.
```
gridbugs, line 9: Can you add a unit to this name (e.g. `server_auto_shutdown_seconds`).

gridbugs (approved):

test/http/http.ml
maiste: Indeed, it can still be modified later if we need to.

rgrinberg:
All this is doing is hiding away buggy tests. The servers should just clean up correctly.

Leonidas-from-XIV:
The one-shot server can't clean up if it never gets contacted, e.g. if a test fails and never issues the HTTP request. It then just waits forever for a request that will not come. Hence I added a timeout so it shuts down automatically after some reasonable time.

PR #3928, opened 2020-11-09
dune formatting: style tweaks

nojb:
This PR tweaks the formatting style for `dune` files in two small ways:

- reduces to one (from two) the number of spaces when indenting certain nested lists. I assume this was unintended because it seems incoherent with the one-space indent for top-level lists.

- uses an `hv` box rather than an `hov` box for these lists. The rationale is that if you have long lists in your `dune` files the "wrapped" style produced using `hov` boxes is unreadable, `diff`-unfriendly, etc.

I realize that the specific formatting style is not versioned, so am not sure how to go about with this change.

Issue #2291, opened 2019-06-17
dune format-dune-file on dune-project is suboptimal


avsm:
The results of running `dune format-dune-file < dune-project` ends up looking a bit terrible:

```
(lang dune 1.10)

(name github)

(generate_opam_files true)

(license MIT)

(maintainers "Anil Madhavapeddy <anil@recoil.org>")

(authors "Anil Madhavapeddy" "David Sheets" "Andy Ray" "Jeff Hammerbacher"
  "Thomas Gazagnaire" "Rudi Grinberg" "Qi Li" "Jeremy Yallop" "Dave Tucker")

(source
 (github mirage/ocaml-github))

(documentation "https://mirage.github.io/ocaml-github/")

(package
 (name github)
 (tags
  (org:mirage org:xapi-project git))
 (depends
  (ocaml
   (>= 4.03.0))
  (dune
   (>= 1.10))
  (uri
   (>= 1.9.0))
  (cohttp
   (>= 0.99.0))
  (cohttp-lwt
   (>= 0.99))
  (lwt
   (>= 2.4.4))
  (atdgen
   (>= 2.0.0))
  (yojson
   (>= 1.6.0))
  stringext)
 (synopsis "GitHub APIv3 OCaml library")
 (description
   "This library provides an OCaml interface to the\n[GitHub APIv3](https://developer.github.com/v3/) (JSON).\n\nIt is compatible with [MirageOS](https://mirage.io) and also compiles to pure\nJavaScript via [js_of_ocaml](http://ocsigen.org/js_of_ocaml)."))

(package
 (name github-jsoo)
 (tags
  (org:mirage org:xapi-project git))
 (depends
  (ocaml
   (>= 4.03.0))
  (dune
   (>= 1.10))
  (github
   (= :version))
  (cohttp
   (>= 0.99.0))
  (cohttp-lwt-jsoo
   (>= 0.99.0))
  (js_of_ocaml-lwt
   (>= 3.4.0)))
 (synopsis "GitHub APIv3 JavaScript library")
 (description
   "This library provides an OCaml interface to the [GitHub APIv3](https://developer.github.com/v3/)\n(JSON). This library installs the JavaScript version, which uses [js_of_ocaml](http://ocsigen.org/js_of_ocaml)."))
```

vs the original which was much more compact:

```
(lang dune 1.10)
(name github)

(generate_opam_files true)

(license MIT)
(maintainers "Anil Madhavapeddy <anil@recoil.org>")
[...]

PR #10716, opened 2024-07-10
Auto format `dune-project` files with `dune fmt`

### LINKED ISSUE #3223: dune-project not formatted with @fmt
When I run `dune build @fmt`, I would expect `dune-project` to be checked as well as other `dune` files.  Any reason for it not to be checked?  I understand that the first line of that file (e.g. `(lang dune 2.3)`) is a bit of an exception, but the rest could probably benefit from auto formatting, no?

## Specifications

- Version of `dune` (output of `dune --version`): 2.3.1
- Version of `ocaml` (output of `ocamlc --version`): 4.09.0
- Operating system (distribution and version): Arch Linux

maxRN:
Hi!

This PR enables automatic formatting of `dune-project` files when running `dune fmt`. This should close #3223. Note that there were some concerns raised on that issue regarding the formatting, but those should have been addressed by #3928, see also #2291.

Note: I'm new to OCaml, especially working on such a mature codebase so there will probably a lot of mistakes in my submission. I hope it can still be of use and I'm more than happy to learn and address any problems with this PR.

### DIFF
--- a/src/dune_lang/format_config.ml
+++ b/src/dune_lang/format_config.ml
@@ -62,18 +62,25 @@ end
 module Enabled_for = struct
   type t =
     | Only of Language.Set.t
+    | Dune_2_All
     | All
 
   let to_dyn =
     let open Dyn in
     function
     | Only l -> variant "only" [ Language.Set.to_dyn l ]
+    | Dune_2_All -> string "dune_2_all"
     | All -> string "all"
   ;;
 
   let includes t =
     match t with
     | Only l -> Language.Set.mem l
+    | Dune_2_All ->
+      fun l ->
+        (match l with
+         | Language.Dialect _ | Language.Dune -> true
+         | _ -> false)
     | All -> fun _ -> true
   ;;
 
@@ -140,20 +147,12 @@ let dune2_dec =
 
 let enabled_for_all = { loc = Loc.none; enabled_for = Enabled_for.All }
 let disabled = { loc = Loc.none; enabled_for = Enabled_for.Only Language.Set.empty }
-
-let enabled_dune_2 =
-  { loc = Loc.none
-  ; enabled_for =
-      Enabled_for.Only
-        (Language.Set.of_list
-           [ Language.Dialect "ocaml"; Language.Dialect "reason"; Language.Dune ])
-  }
-;;
-
+let enabled_dune_2 = { loc = Loc.none; enabled_for = Enabled_for.Dune_2_All }
 let field ~since = field_o "formatting" (Syntax.since Stanza.syntax since >>> dune2_dec)
 
 let is_empty = function
   | { enabled_for = Enabled_for.Only l; _ } -> Language.Set.is_empty l
+  | { enabled_for = Enabled_for.Dune_2_All; _ } -> false
   | { enabled_for = All; _ } -> false
 ;;
 
@@ -177,6 +176,7 @@ let encode_explicit conf =
 let to_explicit { loc; enabled_for } =
   match enabled_for with
   | Enabled_for.All -> None
+  | Enabled_for.Dune_2_All -> Some { loc; enabled_for = Language.in_ext_1_1 }
   | Only l -> Some { loc; enabled_for = l }
 ;;
 
--- a/src/dune_lang/format_config.ml
+++ b/src/dune_lang/format_config.ml
@@ -14,19 +14,19 @@ module Language = struct
     type t =
       | Dialect of string
       | Dune
-      | DuneProject
+      | Dune_project
 
     let compare t1 t2 =
       match t1, t2 with
       | Dialect s1, Dialect s2 -> String.compare s1 s2
       | Dialect _, Dune -> Gt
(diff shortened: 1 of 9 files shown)


src/dune_rules/format_rules.ml
```diff
@@ -133,16 +147,17 @@ let gen_rules_output
...
-            >>> add_diff sctx loc alias_formatted ~dir ~input ~output)))
+          |> Memo.Option.iter
+               ~f:(add_diff_rule ~sctx ~loc ~alias_formatted ~dir ~output_dir ~version)))
+  and* () =
+    match Format_config.includes config Dune with
+    | false -> Memo.return ()
+    | true ->
+      Dune_project.file project
+      |> Memo.Option.iter ~f:(fun path ->
+        if String.equal "_build/default" (Path.Build.to_string dir)
```
maxRN, line 158: I'm not happy with this if check for the hard-coded build directory, but I didn't find a way to check if we are in the "_build/default" directory other than this.

maxRN:
I'm also not sure why the Format CI step fails. It runs `nix develop .#fmt -c make fmt` right? When I run this on my machine I don't get any errors so I'm not sure why it's failing in CI.

emillon:
Thanks.
I can't give a full review at the moment but here are a few remarks in the meantime:
- In terms of design, I don't think that this should be tied to the formatting of `dune` files. Maybe it's useful to be able to disable the formatting of just `dune-project` files, for example.
- This will need to be versioned: if `(lang dune)` is < (3, 17), do not format these files, otherwise do format them by default.
- I'm not sure the formatting has been fixed enough to work well with dune-project files, but that's something we can try and improve later.

src/dune_rules/format_rules.ml
rgrinberg: How about checking it for equality against `Context.build_dir`?

maxRN:
> Thanks. I can't give a full review at the moment but here are a few remarks in the meantime:
> 
> * In terms of design, I don't think that this should be tied to the formatting of `dune` files. Maybe it's useful to be able to disable the formatting of just `dune-project` files, for example.
> * This will need to be versioned: if `(lang dune)` is < (3, 17), do not format these files, otherwise do format them by default.
> * I'm not sure the formatting has been fixed enough to work well with dune-project files, but that's something we can try and improve later.

Thanks for the feedback! I'll take another look

test/blackbox-tests/test-cases/formatting/feature.t/run.t
```diff
@@ -190,7 +190,8 @@ Sometimes, the suggestion is to just remove the configuration.
   5 | (using fmt 1.2)
       ^^^^^^^^^^^^^^^
   Error: Starting with (lang dune 2.0), formatting is enabled by default.
-  To port it to the new syntax, you can delete this part.
+  To port it to the new syntax, you can replace this part by:
+  (formatting (enabled_for dune ocaml reason))
```
maxRN, line 194: Unsure whether we should extend the help message by mentioning the differences between > 3.17 dune and 3.17+ 
i.e. that you can remove the stanza, but that it will also format `dune-project` files.

maxRN:
@emillon I've added a version check so it will only automatically format from 3.17 (?) onwards. There's no explicit way to disable just formatting for `dune-project` files, but just like before users can just specify the file types that they want to be formatted automatically and not include `dune-project`.
One more thing I thought about while implementing this: I think it would make sense to also include `dune-workspace` files in this PR. I don't know what the state on their formatting is, but it would be a lot easier to add support for both file types in one change then having to go back later and add support.

src/dune_lang/format_config.ml
```diff
@@ -14,27 +14,35 @@ module Language = struct
     type t =
       | Dialect of string
       | Dune
+      | DuneProject
```
rgrinberg, line 17: `Dune_project` is the convention

rgrinberg:
Can you enable this feature for dune itself? IIRC, there were some problems with preserving multi line quoted strings in the dune language.

maxRN:
> Can you enable this feature for dune itself? IIRC, there were some problems with preserving multi line quoted strings in the dune language.

Sorry, I don't think I understand the question. What feature exactly are you referring to?

rgrinberg:
Dune itself has a [dune-project file](https://github.com/ocaml/dune/blob/main/dune-project). It would be good to see how this PR works to auto-format it.

maxRN:
Ahh, I see what you mean. I added the formatted `dune-project` file for reference. **NOT** intended to be included in the final PR. You are right, there are some problems with multiline strings - the newlines are output verbatim instead of being actual newline characters. 
My gut feeling is that this should be an easy fix, do you want me to tackle this in this PR as well or create a new one so that we can discuss this separately?

rgrinberg:
If you could fix it in a separate PR that would be great. It's really an issue for all dune files, but project files include large description strings fairly often.

maxRN:
Alright, I'm investigating some possible solutions and will try to create a separate draft PR in the coming days.

PR #10717, opened 2024-07-10
chore: leave comment about hackery

rgrinberg:

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -1589,6 +1589,7 @@ let source_rules (pkg : Pkg.t) =
         | `Directory_or_archive src ->
           loc, Action_builder.copy ~src:(Path.external_ src) ~dst:extra_source
         | `Fetch ->
+          (* This is wrong, but we're doing a temporary workaround. See #10706 *)
           let rule = Fetch_rules.fetch ~target:extra_source `File fetch in
           loc, rule
       in


gridbugs:
What's the hackery here? Are you referring to my claim about lack of support for multiple external sources? Turns out I was wrong about that (see this test: https://github.com/ocaml/dune/blob/a855dd7bb445b74fa5d9fd1e57f8db53cbafd1de/test/blackbox-tests/test-cases/pkg/extra-sources.t#L124). Is there a different issue with extra sources?

rgrinberg:
Because the PR claims that this the fix broke a different use case. So it's important to leave a comment about such situations as readers will always be left wondering if it was an accident.

> Turns out I was wrong about that (see this test:

So does that confirm your fix does not cause any other issues?

Leonidas-from-XIV:
The thing that #10706 mentions is broken is curl failing when attempting to download a large tarball from the oneshot webserver and there have been some attempts to fix the oneshot webserver (weirdly only failing in CI but working fine locally) but the PR sidestepped the issue by reducing the size of the tarball instead.

Handling `extra-sources` is entirely fixed by #10706, I don't think there are any issues that are created by it. It works with multiple `extra-sources` as well and this is ensured by the test added in #10708. So I don't think there is any hackery to document.

The only issue is the mysterious failure of the oneshot webserver on large files on CI but it is just used for the tests and has been worked around already.

rgrinberg:
> Turns out I was wrong about that (see this test:

Okay, if you can confirm that your fix did not introduce any issues you can close this PR.

PR #10718, opened 2024-07-11
doc: split foreign.rst

emillon:
This splits the monolithic `foreign.rst` into 4 pages:
- `foreign_stubs`
- `foreign_archives` (and `foreign_library`)
- `extra_objects`
- flags in foreign code

This is given its own toc in the index, and some xrefs are added.

### DIFF
--- a/doc/reference/extra-objects.rst
+++ b/doc/reference/extra-objects.rst
@@ -1,0 +1,30 @@
+Extra Objects
+-------------
+
+It's possible to specify native object files to be packaged with OCaml
+libraries or linked into OCaml executables. Do this by using the
+``extra_objects`` field of the ``library`` or ``executable`` stanzas.
+For example:
+
+.. code:: dune
+
+    (executable
+     (public_name main)
+     (extra_objects foo bar))
+
+    (rule
+     (targets foo.o bar.o)
+     (deps foo.c bar.c)
+     (action (run ocamlopt %{deps})))
+
+This example builds an executable which is linked against a pair of native
+object files, ``foo.o`` and ``bar.o``. The ``extra_objects`` field takes a list
+of object names, which correspond to the object file names with their path and
+extension omitted (in contrast, ``(rule)`` manipulates file names, so the
+extension needs to be provided).
+
+In this example, the sources corresponding to the objects (``foo.c`` and
+``bar.c``)  are assumed to be present in the same directory as the OCaml source
+code, and a custom ``rule`` is used to compile the C source code into object
+files using ``ocamlopt``. This is not necessary; one can instead compile foreign
+object files manually and place them next to the OCaml source code.
--- a/doc/reference/foreign-archives.rst
+++ b/doc/reference/foreign-archives.rst
@@ -1,0 +1,49 @@
+Foreign Archives
+----------------
+
+Foreign archives are a way to link archives that are separately compiled.
+
+They are particularly useful when embedding a library written in a foreign
+language and/or built with another build system. See :ref:`foreign-sandboxing`
+for more details.
+
+To use this feature, use the ``foreign_archives`` field of the corresponding
+``library`` or ``executable`` stanza. For example:
+
+.. code:: dune
+
+    (library
+     (name lib)
+     (foreign_stubs (language c) (names src1 src2))
+     (foreign_stubs (language cxx) (names src3) (flags -O2))
+     (foreign_archives arch1 some/dir/arch2))
+
+Here, in addition to :doc:`foreign-stubs`, we also specify foreign archives
+``arch1`` and ``arch2``, where the latter is stored in a subdirectory
+``some/dir``.
+
+You can build a foreign archive manually, e.g., using a custom ``rule`` as
+described in :ref:`foreign-sandboxing`, or ask Dune to build it via the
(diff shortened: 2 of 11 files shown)


emillon:
@Leonidas-from-XIV the only content changes are the added xrefs in the flags document and the removed intro in the now-deleted document.

christinerose (approved):

doc/reference/extra-objects.rst
```diff
@@ -0,0 +1,29 @@
...
+
+    (rule
+     (targets foo.o bar.o)
+     (deps foo.c bar.c)
+     (action (run ocamlopt %{deps})))
+
+This example builds an executable which is linked against a pair of native
+object files, ``foo.o`` and ``bar.o``. The ``extra_objects`` field takes a
+list of object names, which correspond to the object file names with their path
+and extension omitted.
```
Leonidas-from-XIV, line 23: Not necessary in this PR but when reading I am wondering why `extra_objects` doesn't use the ending (probably because it uses whatever the native object file extension is) but the `rule` does (because rules don't have any implied filename rules), which makes this a bit strange.

doc/reference/foreign-archives.rst
```diff
@@ -0,0 +1,49 @@
+Foreign Archives
+----------------
+
+You can also specify archives of separately compiled foreign object files
```
Leonidas-from-XIV, line 4: "also" is weird here, as there is nothing in this text to relate to now that it was split. Maybe a short introduction what foreign archives are and what they can be used for would be a better motivation paragraph.

doc/reference/foreign-archives.rst
```diff
@@ -0,0 +1,49 @@
...
+     (language c)
+     (enabled_if true)
+     (names src4 src5)
+     (include_dir headers))
+
+This asks Dune to compile C source files ``src4`` and ``src5`` with
+headers tracked in the ``headers`` directory and put the resulting
+object files into an archive ``arch1``, whose full name is typically
+``libarch1.a`` for static linking and ``dllarch1.so`` for dynamic
+linking.
```
Leonidas-from-XIV, line 38: We might need to revisit that with Windows in mind at a later point in time.

doc/reference/foreign-archives.rst
```diff
@@ -0,0 +1,49 @@
...
+The ``foreign_library`` stanza supports all :doc:`foreign-stubs` fields.
+The ``archive_name`` field specifies the archive's name. You can refer
+to the same archive name from multiple OCaml libraries and executables, so a
+foreign archive is a bit like a foreign library, hence the name of the stanza.
+The ``enabled_if`` field has the same meaning as in the :doc:`dune/library`
+stanza.
+
+Foreign archives are particularly useful when embedding a library written in
+a foreign language and/or built with another build system. See
+:ref:`foreign-sandboxing` for more details.
```
Leonidas-from-XIV, line 49: This paragraph is great but would be better at the top of the file to explain to the user what they are looking at.

doc/reference/foreign-flags.rst
```diff
@@ -0,0 +1,22 @@
...
+  :doc:`dune-workspace/index` file
+- In the :doc:`per-context env <dune-workspace/context>` definitions in a
+  :doc:`dune-workspace/index` file
+- In the :doc:`dune/env` definition of a :doc:`dune/index` file
+- In a :doc:`foreign_stubs <foreign-stubs>`, :doc:`foreign_library
+  <foreign-archives>` or :doc:`extra_objects <extra-objects>` field of an
+  executable or a library.
+
+The ``%{cc}`` :doc:`variable <../concepts/variables>` will contain the flags
+from the first three levels only.
```
Leonidas-from-XIV, line 22: I think this is quite abstract without examples. I think adding an example of what is discussed would be a big improvement to explain at which level which variables flags are set.

Leonidas-from-XIV (commented):
I am not sure splitting the document into multiple pages is an improvement since I liked that the topic of dealing with foreign code was fairly contained on one page, but will defer to your judgement on that.

Thanks for specifying the content changes, that has made it significantly nicer to review! I wrote some comments as I looked through the newly split pages though.

emillon:
> I am not sure splitting the document into multiple pages is an improvement since I liked that the topic of dealing with foreign code was fairly contained on one page, but will defer to your judgement on that.

Thanks for the feedback, there's a balance to find here:
- on the one hand, the monolithic page was the only place where `foreign_stubs` (and friends) were documented, and they contain unrelated documentation (if you're looking for the syntax of `(foreign_stubs)`, it's not likely you're interested in knowing about the history of `use_standard_c_and_cxx_flags`
- on the other hand, the "here are all your options regarding foreign code in Dune" aspect is now lost.

What I suggest is creating an issue about creating the following doc: an how-to guide which would be "how to use non-ocaml code in my project", with sections corresponding to foreign_stubs (you're in control of this C code) and foreign_archives with optionally foreign_library (this is already built somewhere / there's another build system in there). How does that sound?

doc/reference/foreign-flags.rst
emillon: let's do that in a separate issue, I think there's a nice diagram to be reused here.

Issue #10070, opened 2024-02-19
Don't generate a lockfile for the dune package


gridbugs:
When generating a lockdir dune currently makes a lockfile for the dune package if it is among the project's dependencies, however the dune package is treated specially in that it's never built and dune assumes that whenever a package build command invokes a `dune` command, that it's referring to the executable for the instance of dune currently orchestrating the build.

Since the dune package is never built, and the version of dune which is "locked" is implied by the version of dune currently installed, the lockfile is unnecessary. The presence of a lockfile for dune may also suggest to users that the dune package is treated the same as other packages which would be misleading.

See this thread for more context: https://github.com/ocaml/dune/issues/10019

PR #10719, opened 2024-07-12
Toolchains feature (second iteration)

gridbugs:
This is a reimplementation of the toolchains feature based on the feedback on its [original PR](https://github.com/ocaml/dune/pull/10639). The main difference is that compiler packages are now built using the build commands from their lockfiles rather than from their corresponding opam files. This has a number of benefits:
- it's not necessary for dune to download the opam repo to build a toolchain package - all the information dune needs is already present in the lockdir
- a lockfile can be manually edited to have different build commands, and these will be respected when building the toolchain
  - toolchain install directories include a hash of the package's metadata, nix style, so changes to a lockfile in one project won't affect the toolchain used in another project

Note that all this functionality is disabled by default, so this change should have no visible effect unless the feature flags are explicitly enabled (in feature_flags.ml, introduced in this patch).

I recommend reviewing this as separate commits, as there were three distinct changes required to make this work which. The commit messages contain detailed descriptions of each change.

Also @rgrinberg, you mentioned [here](https://github.com/ocaml/dune/pull/10639#issuecomment-2218762292) that it's important that users be able to pin the compiler. In my testing I was unable to pin compiler packages on dune without this feature. I cloned  https://github.com/ocaml/ocaml.git and checked out the 5.2 branch, then tried running `dune pkg lock` with my project pinning ocaml-variants to the checkout, and I got this error:
```
$ dune pkg lock
File "ocaml-variants.opam", line 1, characters 0-0:
Error: Expected formula to be a conjunction of atoms but encountered non-atom
term 'ocaml-option-32bit | ocaml-option-afl | ocaml-option-bytecode-only |
ocaml-option-no-flat-float-array | ocaml-option-flambda | ocaml-option-fp |
ocaml-option-musl | ocaml-option-leak-sanitizer |
ocaml-option-address-sanitizer | ocaml-option-static | ocaml-option-tsan'
```
Is this a regression, or have we never tried pinning the compiler until now?

Either way, I don't see how this change would preclude anyone from pinning the compiler packages when it becomes possible to do in general, so I don't think fixing that issue should block the merging of this PR.


src/dune_rules/pkg_rules.ml
```diff
@@ -259,11 +273,251 @@ module Pkg = struct
...
+               ; Dune_lang.Action.Run
+                   [ Slang.text "mkdir"
+                   ; Slang.text "-p"
+                   ; Slang.text @@ Path.to_string @@ Path.parent_exn prefix
+                   ]
+               ; Dune_lang.Action.Run
+                   [ Slang.text "mv"
+                   ; (* Prevents mv from replacing the destination if it
+                        already exists. This can happen if two dune
+                        instances race to install the toolchain. Note
```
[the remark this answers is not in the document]
gridbugs: That would be ideal but I don't think it's possible when specifying the build with `Dune_lang.Action`s. I couldn't find a way of composing fibers and actions so instead I'm allowing multiple instances of dune to race to build compiler packages with only the first to complete the build able to move the artifacts into the toolchain directory.

src/dune_rules/pkg_rules.ml
```diff
@@ -259,11 +273,251 @@ module Pkg = struct
     ; info : Pkg_info.t
     ; paths : Paths.t
     ; files_dir : Path.Build.t
+    ; system_provided : bool
     ; mutable exported_env : string Env_update.t list
     }
 
   module Top_closure = Top_closure.Make (Id.Set) (Monad.Id)
 
+  let is_system_provided t = t.system_provided
```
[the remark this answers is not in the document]
gridbugs: Neither. Currently the only "system provided" package is `dune`. There is a hard-coded list here: https://github.com/ocaml/dune/blob/f3845bd54eb2b3d47cbe7066f0367903333827ec/src/dune_rules/pkg_rules.ml#L1144. These are packages that dune assumes to already be installed, and won't try to install at all.

src/dune_rules/pkg_rules.ml
```diff
@@ -259,11 +273,251 @@ module Pkg = struct
...
+    let install_cookie t =
+      let bin_dir = bin_dir t in
+      Fs_memo.dir_contents bin_dir
+      >>| function
+      | Error _ -> { Install_cookie.files = Section.Map.empty; variables = [] }
+      | Ok files ->
+        let bin_paths =
+          Fs_cache.Dir_contents.to_list files
+          |> List.filter_map ~f:(fun (filename, kind) ->
+            match kind with
```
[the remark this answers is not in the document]
gridbugs: Just in case there are other types of file in the bin directory (e.g. directories or devices), it makes sure we don't resolve `which` queries to those files.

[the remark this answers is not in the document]
Leonidas-from-XIV: But why would there be such files? Does building the compiler create such entries?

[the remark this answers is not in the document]
gridbugs: There shouldn't be such files when building the compiler. It's probably safe to remove this check.

src/dune_rules/pkg_rules.ml
```diff
@@ -748,17 +1015,27 @@ module Action_expander = struct
       | Sys_ocaml_version ->
         sys_poll_var (fun { sys_ocaml_version; _ } -> sys_ocaml_version)
       | Build -> Memo.return [ Value.Dir (Path.build paths.source_dir) ]
-      | Prefix -> Memo.return [ Value.Dir (Path.build paths.target_dir) ]
+      | Prefix ->
+        Memo.return
+          [ (match override_pform.prefix with
+             | None -> Value.Dir (Path.build paths.target_dir)
```
rgrinberg, line 1021: Why can't you just set `paths.target_dir` for this package instead?

src/dune_pkg/feature_flags.mli
```diff
@@ -0,0 +1,29 @@
...
+    by dune's package management features as the compiler is not
+    relocatable, and this flag allows dune to workaround this problem
+    providing an experience to users that is almost identical to dune
+    installing the compiler packgae.
+
+    When this flag is disabled, users of dune package management need to
+    manage their compiler installation with opam or a system package
+    manager, as compilers packages that would be installed by dune will
+    not work correctly. *)
+val use_toolchains : bool
```
rgrinberg, line 23: We already have `Dune_config` for this. I don't think "feature flags" are different enough

rgrinberg:
There's probably no regression. I would suggest that you take the opam file in that repo and attempt to turn it into a test we can check in. You might be able to minimize the opam file to see what is the problematic bit.

Either way, it does not prevent this PR from being merged.

src/dune_rules/pkg_rules.ml
gridbugs: Because `paths.target_dir` is a `Path.Build.t` but I need to override it with a path outside the build directory. I guess we could change `Paths.t` to use `Path.t` rather than `Path.Build.t`, but then we would lose the guarantee that all the paths are (usually) within the build directory. Do you think it would be worth it to make this change to `Paths.t`?

src/dune_pkg/feature_flags.mli
gridbugs: Good point! I think that will suit my needs here. Will update.

gridbugs: done

gridbugs:
Made an issue for the problem pinning the compiler. Seems to only happen when the pinned package has both depopts in its opam file and a dune-project file. https://github.com/ocaml/dune/issues/10739

src/dune_rules/pkg_rules.ml
```diff
@@ -1564,8 +1900,13 @@ let source_rules (pkg : Pkg.t) =
       Lock_dir.source_kind source
       >>= (function
        | `Local (`File, _) | `Fetch ->
-         let fetch = Fetch_rules.fetch ~target:pkg.paths.source_dir `Directory source in
-         Memo.return (Dep.Set.of_files [ Path.build pkg.paths.source_dir ], [ loc, fetch ])
+         let+ fetch =
+           let target = pkg.paths.source_dir in
+           if Pkg.Toolchain.is_compiler_and_toolchains_enabled pkg
+           then Pkg.Toolchain.fetch_action pkg ~target ~source
```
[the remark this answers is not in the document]
gridbugs: Take a look at the `Pkg.Toolchain.dummy_fetch` function above. When the compiler is already installed, this avoids the need to download it but still creates a dummy source directory for the package so dune can act as though the package is installed. I think it's just an optimization though so I can take it out if necessary.

src/dune_rules/pkg_rules.ml
gridbugs: I had a go at changing the types in `Paths.t` to be `Path.t` but ran into an issue where a `Paths.t` is used to create actions, and a lot of the actions constructors expect `Path.Build.t`. I don't want to convert from `Path.t` to `Path.Build.t` in these places because then we'd have to think about what to do when the conversion fails which would complicate the logic.

src/dune_rules/pkg_rules.ml
```diff
@@ -1062,9 +1352,11 @@ module Action_expander = struct
     ;;
   end
 
-  let expander context (pkg : Pkg.t) =
+  let expander context (pkg : Pkg.t) ~override_pform =
     let+ { Artifacts_and_deps.binaries; dep_info } =
-      Pkg.deps_closure pkg |> Artifacts_and_deps.of_closure
+      Pkg.deps_closure pkg
+      |> List.filter ~f:(Fun.negate Pkg.is_system_provided)
```
rgrinberg, line 1358: how come this filter is needed now?

src/dune_rules/pkg_rules.ml
```diff
@@ -1744,40 +2049,43 @@ let setup_rules ~components ~dir ctx =
...
   |> Memo.parallel_map ~f:(fun (package : Lock_dir.Pkg.t) ->
     let package = package.info.name in
-    Resolve.resolve db context (Loc.none, package)
-    >>| function
-    | `Inside_lock_dir pkg -> Some pkg
-    | `System_provided -> None)
-  >>| List.filter_opt
+    Resolve.resolve db context (Loc.none, package))
   >>| Pkg.top_closure
+  >>| List.filter ~f:(Fun.negate Pkg.is_system_provided)
```
rgrinberg, line 2088: doesn't seem need to do this filtering in two different places. can we unify this?

src/dune_rules/pkg_toolchain.mli
```diff
@@ -0,0 +1,28 @@
...
+      manage their compiler installation with opam or a system package
+      manager, as compilers packages that would be installed by dune will
+      not work correctly. *)
+  val enabled : bool Config.t
+
+  (** When building the compiler with toolchains, build it with [make
+      -j] rather than [make -j1], allowing more parallelism. This can
+      theoretically lead to build failures, but these are extremely rare in
+      practice. *)
+  val build_compiler_in_parallel : bool Config.t
```
rgrinberg, line 27: Why do we need this workaround for the compiler? If there are build failures, then they exist in regular opam as well. Or they also have an override?

src/dune_rules/pkg_rules.ml
```diff
@@ -1149,11 +1458,7 @@ module DB = struct
 end
 
 module rec Resolve : sig
-  val resolve
-    :  DB.t
-    -> Context_name.t
-    -> Loc.t * Package.Name.t
-    -> [ `Inside_lock_dir of Pkg.t | `System_provided ] Memo.t
+  val resolve : DB.t -> Context_name.t -> Loc.t * Package.Name.t -> Pkg.t Memo.t
```
rgrinberg, line 1461: I don't really see why this needed to change, but it doesn't seem like an improvement. An explanation on why this was needed would be helpful.

src/dune_rules/pkg_rules.ml
```diff
@@ -1214,16 +1517,13 @@ end = struct
...
-      then Memo.return `System_provided
-      else
-        Memo.exec memo (db, ctx, name)
-        >>| function
-        | Some s -> `Inside_lock_dir s
-        | None ->
-          User_error.raise
-            ~loc
-            [ Pp.textf "Unknown package %S" (Package.Name.to_string name) ]
+      Memo.exec memo (db, ctx, name)
```
rgrinberg, line 1520: The old behavior did not pointlessly memoize the look up in `db.system_rpovided`. It doesn't really matter, but I'd like to understand why this was changed.

src/dune_rules/pkg_rules.ml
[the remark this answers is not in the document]
rgrinberg: Yeah, I don't think this is necessary. Or rather, the mechanism to address the problem that you're stating already exists for regular packages through the shared cache. So we might as well just use it everywhere.

src/dune_rules/pkg_rules.ml
```diff
@@ -1819,12 +2127,12 @@ let find_package ctx pkg =
...
-    >>| (function
-           | `System_provided -> Action_builder.return ()
-           | `Inside_lock_dir pkg ->
-             let open Action_builder.O in
-             let+ _cookie = (Pkg_installed.of_paths pkg.paths).cookie in
-             ())
-    >>| Option.some
+    let+ pkg = Resolve.resolve db ctx (Loc.none, pkg) in
+    Some
+      (if pkg.system_provided
```
rgrinberg, line 2132: Again this change seems gratuitous. With the old code, the return type forced callers to handle "system provided" packages, now one has to remember to look up some field.

src/dune_rules/pkg_rules.ml
rgrinberg: Okay, I will try doing something to improve this myself.

> because then we'd have to think about what to do when the conversion fails which would complicate the logic.

I think that's exactly the kind of a problem that's already present in this code. One has to remember if the package is a toolchain to access the package fields correctly, and it's very error prone.

src/dune_rules/pkg_toolchain.mli
```diff
@@ -0,0 +1,56 @@
...
+  -> installation_prefix:Path.Outside_build_dir.t
+  -> suffix:string
+  -> Dune_lang.Action.t
+
+module Override_pform : sig
+  (** Allows various pform values to be overriden when expanding pforms
+      inside package commands. *)
+  type t =
+    { prefix : Path.t
+    ; doc : Path.t
```
rgrinberg, line 48: Why was doc included and other install sections were not?

src/dune_rules/pkg_rules.ml
```diff
@@ -259,11 +260,107 @@ module Pkg = struct
     ; info : Pkg_info.t
     ; paths : Paths.t
     ; files_dir : Path.Build.t
+    ; system_provided : bool
```
rgrinberg, line 263: The reason why this flag doesn't make sense is because if a package is "system provided", how can it even have a valid `Pkg.t`? For example, if OCaml is provided by the system, how do you set `paths` correctly?

src/dune_rules/pkg_toolchain.mli
gridbugs: Prior to https://github.com/ocaml/dune/pull/10755, dune would always build all packages with `-j1`. The compiler takes particularly long to compile with `-j1`, and I've been told that building with `-j` might not work but I've never encountered any problems with it.

src/dune_rules/pkg_rules.ml
gridbugs: Copied from the commit message of https://github.com/ocaml/dune/pull/10719/commits/48ed844f53629999193121d116c7db4de752775f:

The value of the PATH environment variable when building a pacage is
set such that all the executables in that package's dependency cone
are visible. This includes the ocaml compiler itself. In the past,
system-provided packages (such as dune) were omitted from this
dependency cone, along with their transitive dependencies.

This creates a problem for packages that depend on dune and don't
directly depend on the compiler. Such packages typically build with
`dune build -p ...`, and dune assumes that it's being run in an
environment where the ocamlc compiler is visible. Although the dune
package depends on the compiler, due to its exclusion from the
dependency cone used to set the PATH variable dune is unable to find
the compiler when building the package.

We haven't noticed this problem until now because it's rare to use
dune in environments where the compiler is not globally installed or
present in the active opam switch, however running dune on systems
without a global ocaml installation is something we're trying to
enable in the near future.

The fix is to add all packages, system-provided or not, to the graph
of [Pkg.t] objects that is used to compute the build environment for
each package. To prevent dune from attempting to build system-provided
packages, the check whether a package is system-provided is deferred
to the point where dune computes the list of packages to build.

src/dune_rules/pkg_toolchain.mli
gridbugs: Because the build command of `ocaml-base-compiler` only specifically refers to `doc` (and `prefix`).

src/dune_rules/pkg_rules.ml
gridbugs: The only reason I added system provided packages was so that their transitive dependencies could be tracked. How about we make this type a variant with a system-provided case and a regular case?

src/dune_rules/pkg_rules.ml
```diff
@@ -318,18 +318,17 @@ module Pkg = struct
...
-            ; Package_name.of_string "ocaml-variants"
-            ]
-        in
-        fun t -> Package_name.Set.mem compiler_package_names t.info.name
-      else Fun.const false
+    let is_compiler_and_toolchains_enabled t =
+      Config.get Toolchains_config.enabled
+      &&
+      let module Package_name = Dune_pkg.Package_name in
+      let compiler_package_names =
```
gridbugs: In my original PR this set was computed once ever and then used in each call to `is_compiler_and_toolchains_enabled`. There's no point using a set here if it's being computed on each call - we might as well just linearly search the list.

gridbugs: I replaced the use of a set here with a linear search of the list.

src/dune_rules/pkg_rules.ml
gridbugs: Otherwise dune would attempt to install system provided packages, which we don't want it to do.

src/dune_rules/pkg_rules.ml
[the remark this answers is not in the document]
gridbugs: I removed the custom fetch rule for toolchain packages.

gridbugs:
The main change I made today was removing the custom fetch rule for toolchain packages. I modified the build command in the case where the toolchain package is already installed such that it now generates an empty config.cache for the compiler package, as without this file other packages can't see that the compiler is installed. Formerly this was done as part of the custom fetch action for toolchain packages.

src/dune_rules/pkg_rules.ml
rgrinberg: If a package is system provided, why do we need to track its (direct or transitive) deps? You would think those deps would system provided as well. Or are they runtime deps of sort?

src/dune_rules/pkg_rules.ml
rgrinberg: Thanks for the explanation. 

> This creates a problem for packages that depend on dune and don't directly depend on the compiler.

This points to an issue with such packages and not with dune. It is in fact possible to use dune without an OCaml compiler and we should encourage people to do so when appropriate (for building coq code for example). So I would much prefer if we could just fix the metadata for these packages in opam-repository rather than to add workarounds in dune.

If a workaround is truly necessary in dune, I would rather it be done in the least intrusive way possible. For example, by preprocessing all deps of such packages that depend on dune and not OCaml to also include OCaml.

src/dune_rules/pkg_rules.ml
rgrinberg: Sure, that's fine. Note that I only changed your code because it was evaluating the config as a module side effect. That's not allowed because the config needs to be initialized first. You can keep using a set but you need to delay the check with lazy.

In any case, it will not matter as binary search and linear search are pretty close when the size of the collection is 2.

src/dune_rules/pkg_toolchain.mli
rgrinberg: Ack. Let's just follow what opam does and if this is indeed an issue, it's between opam repository and the compiler maintainers.

rgrinberg:
Okay, this is shaping up to look good. I'd say the hack about injecting the transitive deps of system provided packages is my last point of contention. I would suggest one of the following courses of actions:

1. Just get rid of this workaround and send some PR's to opam-repository to fix the offending packages.
2. Explicitly re-export whatever you need from a system provided package. So for example, you can say that dune will re-export OCaml and add it as a dependency when resolving the `depends` field of any package that includes dune

While we are including dune itself in the lock directory at the moment, this is temporary until https://github.com/ocaml/dune/issues/10070 is fixed. So the current workaround will need to go away anyway.

avsm:
Adding a direct dependency to such packages in the ocaml/opam-repository seems reasonable. They do, after all, do directly use `ocaml` and should have a dependency on it.  Please do send in such a PR rather than introducing a workaround into dune.

gridbugs:
Closing now that https://github.com/ocaml/dune/pull/10769 is merged.

PR #10720, opened 2024-07-12
doc: rework (alias) page

emillon:
This adds `describe::` markup and removes outdated information about (action).

### DIFF
--- a/doc/reference/dune/alias.rst
+++ b/doc/reference/dune/alias.rst
@@ -1,55 +1,70 @@
+.. highlight:: dune
+
 alias
 -----
 
-The ``alias`` stanza adds dependencies to an alias or specifies an action to run
-to construct the alias.
+.. describe:: (alias ...)
 
-The syntax is as follows:
+   Add dependencies to an alias.
 
-.. code:: dune
+   Aliases do not need to be explicitly created, adding to a new name will
+   "create" an alias.
+   An alias with name ``x`` can be built by running ``dune build @x``.
+   See :doc:`/reference/aliases`.
 
-    (alias
-     (name    <alias-name>)
-     (deps    <deps-conf list>)
-     <optional-fields>)
+   The common use of the ``alias`` stanza is make an alias depend on other ones::
 
-``<name>`` is an alias name such as ``runtest``.
+       (alias
+        (name runtest)
+        (deps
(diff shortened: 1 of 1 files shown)


doc/reference/dune/alias.rst
```diff
@@ -1,55 +1,58 @@
+.. highlight:: dune
+
 alias
 -----
 
-The ``alias`` stanza adds dependencies to an alias or specifies an action to run
-to construct the alias.
+.. describe:: (alias ...)
+
+   Add dependencies to an alias.
```
Leonidas-from-XIV, line 8: The wording here is a bit unclear whether or not you can create new aliases or can just add dependencies to existing ones. I think it would be better to preempt the users question and state this here directly.

doc/reference/dune/alias.rst
```diff
@@ -1,55 +1,58 @@
...
+   an action to run to construct the alias. Please use a :doc:`rule` stanza
+   with the ``alias`` field instead.
+
+   .. describe:: (name <name>)
 
-The syntax is as follows:
+      An alias name such as ``runtest``.
 
-.. code:: dune
+      This field is required.
```
Leonidas-from-XIV, line 26: Maybe `runtest` is not a good example because this can create confusion between `dune runtest` (not an alias but people might think it is) and `dune build @runtest` (an alias). Maybe `install` (`dune install` exists too but isn't widely used) or `check`?

Leonidas-from-XIV (approved):
I think the change are good and while I am aware that the wording is mostly taken from the previous version, I have some suggestions for improvements.

doc/reference/dune/alias.rst
emillon: I'm not sure. Adding to `runtest` is arguably the most common use of `(alias)`, for example to depend on `.exe` file to be built but not executed. In contrast, adding to `install` does nothing (since it won't generate install entries in the `.install` file).

Leonidas-from-XIV: Then maybe explaining that this will modify what runs as part of `dune build @runtest` (which is equivalent to `dune runtest`) will clear things up.

The thing I want to avoid is for users to think that aliases turn into subcommands.

PR #10723, opened 2024-07-12
fix: name suggestion on opam context

### LINKED ISSUE #10721: Add a hint to context name error
## Context

Hi,
When you specify an `Opam` switch with a _path_ instead of a `name`:
```dune
(context
  (opam
    (switch /path/to/switch)))
```
you get the following error message:
```
Error: Generated context name "/path/to/switch" is invalid
Please specify a context name manually with the (name ..) field
```

## Modification proposed

There are two modifications (discussed on the ocamllabs slack), we could make to improve the readability:
- Add `The generated context name "..." from the switch name is invalid` to the error message
- Provide a hint about it: `Hint: you may want to add (name switch-name)`

The message would be the following:
```
Error: Generated context name
"/path/to/switch" is invalid
The generated context name "/path/to/switch" from the switch name is invalid.
Please specify a context name manually with the (name ..) field

Hint: you may want to add (name switch-name) 
```

maiste:
Hi,
This PR fixes #10721. It provides more information when someone set a switch using an absolute path.
The new error message is:
```
Error: Generated context name
"/path/to/switch" is invalid
The generated context name "/path/to/switch-name" from the switch name is invalid.
Please specify a context name manually with the (name ..) field
Hint: you may want to add (name switch-name) 
```

### DIFF
--- a/src/dune_rules/workspace.ml
+++ b/src/dune_rules/workspace.ml
@@ -462,6 +462,14 @@ module Context = struct
       Common.equal base t.base && Opam_switch.equal switch t.switch
     ;;
 
+    let name_hint_opt name =
+      if String.is_prefix ~prefix:"/" name
+      then (
+        let context_name = Filename.basename name in
+        Some [ Pp.textf "(name %s) would be a valid context name" context_name ])
+      else None
+    ;;
+
     let decode =
       let+ loc_switch, switch = field "switch" (located string)
       and+ name = field_o "name" Context_name.decode
@@ -479,9 +487,10 @@ module Context = struct
              | None ->
                User_error.raise
                  ~loc:loc_switch
-                 [ Pp.textf "Generated context name %S is invalid" name
-                 ; Pp.text
-                     "Please specify a context name manually with the (name ..) field"
+                 ?hints:(name_hint_opt name)
+                 [ Pp.textf
+                     "The name generated from this switch can not be used as a context \
+                      name. Please use (name) to disambiguate."
                  ])
         in
         let base = { base with targets = Target.add base.targets x; name } in
--- a/test/blackbox-tests/test-cases/github10721.t
+++ b/test/blackbox-tests/test-cases/github10721.t
@@ -6,13 +6,14 @@
   > (lang dune 3.0)
   > (context
   >  (opam
-  >   (switch /absolute/path/to/opam/switch)))
+  >   (switch /absolute/path/to/opam/switchname)))
   > EOF
 
   $ dune build
-  File "dune-workspace", line 4, characters 10-39:
-  4 |   (switch /absolute/path/to/opam/switch)))
-                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-  Error: Generated context name "/absolute/path/to/opam/switch" is invalid
-  Please specify a context name manually with the (name ..) field
+  File "dune-workspace", line 4, characters 10-43:
+  4 |   (switch /absolute/path/to/opam/switchname)))
+                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  Error: The name generated from this switch can not be used as a context name.
+  Please use (name) to disambiguate.
+  Hint: (name switchname) would be a valid context name
   [1]


emillon:
Thanks. Can you add a test for this?

src/dune_rules/workspace.ml
```diff
@@ -462,6 +462,14 @@ module Context = struct
       Common.equal base t.base && Opam_switch.equal switch t.switch
     ;;
 
+    let name_hint_opt name =
+      if name <> "" && String.contains name '/'
```
emillon, line 466: The test for `""` is unnecessary, but what do you think about just checking at the beginning for absolute paths? (`String.is_prefix ~prefix:"/"`)

src/dune_rules/workspace.ml
```diff
@@ -479,7 +487,10 @@ module Context = struct
              | None ->
                User_error.raise
                  ~loc:loc_switch
-                 [ Pp.textf "Generated context name %S is invalid" name
+                 ?hints:(name_hint_opt name)
```
emillon, line 490: There's an opportunity to refactor to use `Context_name.of_string_user_error` instead of `of_string_opt` (see `Stringlike_intf`).
Then you can use the `Stringlike` API to implement that hint, by defining a `hint_valid` function (which can use your basename implementation if the string looks like an absolute path, or something less elegant otherwise)

src/dune_rules/workspace.ml
maiste: Is it necessary an absolute or could it work with relative path?

emillon: I think that opam switch names are either absolute paths (local switches) or strings without `/` (global switches), but there might be other code paths that try to build context names from user input.

maiste: I think we can keep the heuristic that it starts with a `/` then as the last scenario seems uncommon.

src/dune_rules/workspace.ml
maiste: After investigating, I will finally keep the error like this. Indeed, if I take this path, the error would look like: 
```
File "dune-workspace", line 5, characters 12-45:
5 |     (switch /path/to/the/opam/switch/named/dune)))
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: "/path/to/the/opam/switch/named/dune" is an invalid context name.
It must not be empty, be the exact string "log", contain slashes, contain
backslashes or start with the symbol  ".".
Hint: (name dune) would be a correct context name
```
It makes the error message more obscure as it totally eclipse the context where it happens...

emillon: Sorry, what's specifically wrong about that error message? You don't have to define `description_of_valid_string` if it doesn't make sense.

test/blackbox-tests/test-cases/github10721.t
```diff
@@ -0,0 +1,21 @@
+
+  $ cat > dune-project << EOF
+  > (lang dune 3.0)
+  > EOF
+
+  $ cat > dune-workspace << EOF
+  > (lang dune 3.0)
+  > (context
+  >   (opam
```
emillon, line 9: we use a 1 space indent in the test suite

test/blackbox-tests/test-cases/github10721.t
```diff
@@ -0,0 +1,21 @@
...
+  >     (switch /absolute/path/to/opam/switch)))
+  > EOF
+
+  $ dune build
+  File "dune-workspace", line 4, characters 12-41:
+  4 |     (switch /absolute/path/to/opam/switch)))
+                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  Error: The generated context name "/absolute/path/to/opam/switch" from the
+  switch name is invalid.
```
emillon, line 18: How about: this switch name can not be used as a context name?

src/dune_rules/workspace.ml
maiste: Yes, I can remove it, but, IMHO, this is not the problem here. If we remove the `match ... with` we lost the information about the fact that the user should add `(name ...)` field in case we don't have it in the `(opam (switch ...))` context.

I don't know a way to specify this using the `Stringlike` API as the `hint_valid` name, insert a suggestion into an already existing string: `Format.sprintf "%s would be a correct context name" (f_extracted_from_hint_valid s)`.  I attempt to modify the `hint_valid` function with my `name_hint_opt` function, but the result does not look informative enough (as you can see in the output above).

Furthermore, I don't see a way to specify the additional message, as it was suggested in the original discussion on Slack:
```
The generated context name "..." from the switch name is invalid, Please specify a context name manually with the (name ..) field.
```
This message makes sens because the user has the information that the name what generated from the `switch` and that it should manually add the `(name ...)` field. Maybe I'm misunderstanding how the `Stringlike` API work and there is a more elegant way to do it? :confused:

test/blackbox-tests/test-cases/github10721.t
maiste: I think this is the same issue we have in https://github.com/ocaml/dune/pull/10723#discussion_r1678073928: users won't understand why Dune is talking about `context  name` whereas we never express a context name. The information that it generated and that we need a `(name ...)` to prevent the automatic generation should be mandatory IMO.

emillon: Ok, then how about something like this:

> This switch name can not be used as a context name. Please use (name) to disambiguate
> Hint: (name xxx) would be a valid context name

Something like that.

My points are:
- we don't need to repeat the switch name because it's pointed by the loc in the error message
- It's clear that we're defining a context. We're in a `(context)` field.

src/dune_rules/workspace.ml
emillon: let's scratch that idea, it's just an implementation detail

emillon:
Can you open a PR with just the test? it will show how this PR changes the error message.

test/blackbox-tests/test-cases/github10721.t
maiste: I agree with this message, but I would still emphases the fact it is auto generated. I would add:

> The name generated from this switch can not be used as a context name. Please use (name) to disambiguate
> Hint: (name xxx) would be a valid context name

WDYT?

I agree on both of your points :+1:

emillon: Yes, that would work

maiste:
I pushed the fixes and opened the test PR :+1:

emillon (approved):

emillon:
Thanks!

PR #10724, opened 2024-07-12
feat: configure-time toggles

emillon:
This is a demo of configure-time toggles, a way to set some configuration values at configure time.

A `party_mode` toggle is included, that displays a colored message when the build system starts.

By default, it is off. It can be enabled either:

- at runtime, by setting `DUNE_CONFIG__PARTY_MODE=enabled`
- at configure time, by running `./configure --toggles party_mode`

### DIFF
--- a/bin/main.ml
+++ b/bin/main.ml
@@ -101,6 +101,7 @@ let exit_and_flush code =
 ;;
 
 let () =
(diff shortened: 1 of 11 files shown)


emillon:
(note to self, the implementation isn't great, if we switch to generating an init function in `Setup` we can avoid the GADT stuff.)

Leonidas-from-XIV:
Closing in favor of #10843.

PR #10726, opened 2024-07-15
Implement post dependencies

gridbugs:
When computing the dependencies of individual packages, distinguish between post and regular dependencies. This is done by first evaluating a package's dependency formula with the same environment as was used for solving (with post=true) to compute all dependencies, then computing non-post dependencies by evaluating it a second time with post=false, ignoring any additional packages added to the formula or any terms that now have no solutions as a consequence of this change to the environment.

Note that this change also sets post=true during solving which matches the behavior of Opam's solver.

Fixes https://github.com/ocaml/dune/issues/10670


rgrinberg:
I think this is missing the check to make the post deps available in the build environment? So I'm surprised if this is enough to make things work for the compiler.

Also, does this include post dependencies in the lock directory? IIUC, this will include such deps but will leave as "dangling" and unused. Either way, a test case to demonstrate the behavior is appreciated.

gridbugs:
Currently it doesn't include dependencies in the lock directory.

What do you mean by:
> the check to make the post deps available in the build environment

gridbugs:
Oh I guess you mean allowing the build commands for a package to access executables installed as part of the package's post dependencies. I don't think this should be allowed, as due to them being post dependencies, their executables might not have been built yet.

rgrinberg:
This what I (and I suppose David since he agreed with my suggestion in the bug report) imagine it to look like:

```
$ cat foo.pkg
(post bar)
$ cat baz.pkg
$ (deps foo) 
```

Now when building `baz`, the effective set of deps we depend on is `(foo bar)`, since `bar` is a post dep of `foo`.

rgrinberg:
> Currently it doesn't include dependencies in the lock directory.

Is there a test to demonstrate this?

gridbugs:
Ah good point, currently this change doesn't account for transitive dependencies that happen to be post dependencies when building packages. To address this I'll add post dependencies to lockfiles and add them to the build environment.

gridbugs:
Ok I've updated this PR such that post dependencies are now tracked by a new lockfile field, and when building a package, the post dependencies of all package's within its dependency closure (but not its own post dependencies) are available in its build environment.

src/dune_pkg/package_universe.ml
```diff
@@ -55,7 +55,7 @@ let concrete_dependencies_of_local_package t local_package_name ~with_test =
          (Solver_env.to_env t.solver_env)
          t.version_by_package_name
   with
-  | Ok s -> s
+  | Ok { regular; _ } -> regular
```
Leonidas-from-XIV, line 58: Nitpick, but maybe better use `post=_` to discard it in case new fields get added as we want the compiler to complain.

test/blackbox-tests/test-cases/pkg/transitive-post-deps.t
```diff
@@ -0,0 +1,94 @@
...
+  Bar installer running foo.sh:
+  This is foo!
+  Done installing bar!
+
+  $ cat _build/default/x.txt
+  This is bar!
+
+  $ dune clean
+
+A package with a regular dependenc on "foo".
```
Leonidas-from-XIV, line 60:
```suggestion
A package with a regular dependency on "foo".
```

src/dune_rules/pkg_rules.ml
```diff
@@ -277,7 +284,7 @@ module Pkg = struct
         ]
   ;;
 
-  let deps_closure t = top_closure t.depends
+  let deps_closure t = top_closure (t.depends @ t.inherited_post_depends)
```
Leonidas-from-XIV, line 287: Aren't `t.inherited_post_depends` added twice now? First here and then in `top_closure`?

Leonidas-from-XIV (approved):
Most questions I had solved themselves while I was reading this PR so that's great.

I much prefer this approach of specifically listing post-depends instead of encoding them in a magic way in the dependency formula.

src/dune_rules/pkg_rules.ml
```diff
@@ -1175,13 +1189,40 @@ end = struct
...
           (Dune_pkg.Lock_dir.Pkg.files_dir info.name ~lock_dir)
       in
+      let* inherited_post_depends =
+        (* Deduplicate package names, preserving the association
+           between package names and (any) one of the locations of the
+           package name (for error messages). *)
+        let all_locs_by_name =
+          List.concat_map depends ~f:(fun (pkg : Pkg.t) ->
+            List.map pkg.post_depend_names ~f:(fun (loc, name) -> name, loc))
+          |> Package.Name.Map.of_list_reduce ~f:(fun x _ -> x)
```
rgrinberg, line 1199: This step does re-ordering of deps

rgrinberg (changes requested):

src/dune_rules/pkg_rules.ml
```diff
@@ -1175,13 +1189,40 @@ end = struct
...
+        in
+        Package.Name.Map.to_list all_locs_by_name
+        |> Memo.parallel_map ~f:(fun (name, loc) ->
+          if Package.Name.equal name info.name
+          then
+            (* It's possible for a package to be a post
```
rgrinberg, line 1205: I don't really get this problem, is there a test for it?

src/dune_rules/pkg_rules.ml
```diff
@@ -1175,13 +1189,40 @@ end = struct
           (Context_name.build_dir ctx)
           (Dune_pkg.Lock_dir.Pkg.files_dir info.name ~lock_dir)
       in
+      let* inherited_post_depends =
+        (* Deduplicate package names, preserving the association
+           between package names and (any) one of the locations of the
+           package name (for error messages). *)
+        let all_locs_by_name =
+          List.concat_map depends ~f:(fun (pkg : Pkg.t) ->
```
rgrinberg, line 1197: This step looks wrong to me. You should be traversing the post deps when computing the closure

src/dune_rules/pkg_rules.ml
```diff
@@ -266,7 +270,10 @@ module Pkg = struct
 
   let top_closure depends =
     match
-      Top_closure.top_closure depends ~key:(fun t -> t.id) ~deps:(fun t -> t.depends)
+      Top_closure.top_closure
+        depends
+        ~key:(fun t -> t.id)
+        ~deps:(fun t -> t.depends @ t.inherited_post_depends)
```
rgrinberg, line 276: This looks wrong to me. I would imagine this to look like:

```ocaml
~deps:(fun t -> t.depends @ List.concat_map t.depends ~f:(fun d -> d.post))
```

rgrinberg:
The test suite seems to always start by executing the solver. I don't think this should be done for exercising the behavior of post deps in the lock directory. You should write those lock files manually and describe the behavior we expect.

By the way, we need a test to make sure we preserve the order of post deps as we computed it from the opam file. We should not re-order `a {post} b {post}` and `b {post} a {post}` aren't equivalent

src/dune_rules/pkg_rules.ml
gridbugs: This is covered by the `transitive-post-deps` test. `foo` has a post dep on `bar`, and `bar` has a regular dep on `foo`. This loop is enumerating the post deps of each regular dep of a package, and `bar` is a post dep of `foo`, which is a regular dep of `bar`. Without this check, we'd try to resolve `bar` while computing the inherited post dependencies of `bar` which would be a memo loop.

gridbugs:
I can write some more tests where the lockfiles are created by hand, but can you explain a bit about what value you hope to gain by including such tests.

src/dune_rules/pkg_rules.ml
gridbugs: How is that different than what I've written? The `t.inherrited_post_depends` field contains the union of post dependencies of each dependency of `t` (apart from `t` itself).

src/dune_rules/pkg_rules.ml
gridbugs: Are you suggesting to compute post deps when computing the `depends` field, just above this in the file? I deliberately separated the logic for computing inherited post deps from computing regular deps for clarity. Also note that we can't traverse post dependencies in the same way that we traverse regular dependencies as that would lead to cycles.

Why is it necessary to traverse post deps when computing the dependency closure?

src/dune_rules/pkg_rules.ml
gridbugs: Ah good point. Will fix.

src/dune_rules/pkg_rules.ml
gridbugs: Oh and I should clarify that the reason I added a field with the union of post dependencies of all deps rather than just adding a field with the package's own post deps is it simplifies handling dependency cycles broken by post dependencies.

src/dune_rules/pkg_rules.ml
gridbugs: Ok I changed this to preserve the order in which post dependencies appear in the lockfile.

gridbugs:
Fixed the issue where inherited post deps would be reorderd and added a test that the order of post deps in lockfiles is respected when installing packages.

rgrinberg:
> I can write some more tests where the lockfiles are created by hand, but can you explain a bit about what value you hope to gain by including such tests.

Thanks for asking. I don't believe this was ever discussed elsewhere, but it's quite an important topic. I have two reasons for doing this. The first is a general principle, and the second is specific to this type of feature.

First, we must keep in mind that not all tests suites are created equal. Good test suites have the following property:

1. For any potential breaking change, there should be a test that breaks when this change is introduced.
2. This test should be as resistant as possible to breaking when unrelated changes are made.

On the other hand, bad suites lack the above property. Therefore, what they end up causing is that any breaking change - no matter how small, breaks everything. This makes debugging a nightmare and the greatly reduces the utility of the test suite.

Notice how I've laboriously introduced tests for every single feature of the lock directories in isolation. I don't know if you've ever taken advantage of this, but I have found it invaluable to know exactly what impact on the package build rules my changes have made. I would like the test suite to remain useful for this purpose.

The second argument is specific to the fact that you're introducing a feature to the lock directory spec. Although we can remain oblivious to this for now, we must always remember that the lock directories are versioned independently of the solver step. In practice, this means that we are allowed to modify the solver to produce a new type of lock directory, but we must always guarantee that we can interpret and build all lock directories that we may have produced in the past. To do this end, we should at the very least make sure that we're testing every single lock directory feature we support. Irrespective of whether we generate it or not.

src/dune_rules/pkg_rules.ml
rgrinberg: When you say "memo loop", I assume you mean a cycle during the _resolution_ of the packages (as opposed to a dependency cycle). If that is the case, then there's no need to introduce a workaround and we can just modify resolution not to cycle.

src/dune_rules/pkg_rules.ml
rgrinberg: I mentioned that it was "wrong" before, but that is probably too strong of a statement. I rather meant "not obviously correct". I do not yet see the gain in clarity from splitting the computation of the closure into two steps, but I do understand why you've needed to do so to avoid the resolution cycle. It's not the most direct way to solve this problem though.

If you look at lib.ml, you will see that all of these problems have already been solved for libraries. In particular, post dependencies are analogous to dependencies marked with `re_exports`. My initial instinct is to solve the exact same problem in the exact same way, rather than introducing an approach with its own pros and cons.

I can take a look at doing this, since I'm not sure how familiar you are with lib.ml

src/dune_rules/pkg_rules.ml
rgrinberg: > How is that different than what I've written? The t.inherrited_post_depends field contains the union of post dependencies of each dependency of t (apart from t itself).

I'd like to think it's the same, but I'm not sure - that's the problem. Doing it two stages converted a change that is simple to specify into something I need to think about. Let me get back to you on that.

rgrinberg:
I hacked on this a bit tonight. Didn't manage to get it working, but will try again this week.

src/dune_rules/pkg_rules.ml
```diff
@@ -255,19 +255,32 @@ module Pkg = struct
...
+
+  let deps t =
+    let* depends = Memo.Lazy.force t.depends in
+    Memo.List.concat_map depends ~f:(fun dep ->
+      let+ post_depends =
+        Memo.Lazy.force dep.post_depends
+        (* It's possible for a package to be a post dependency of one of its
+           dependencies but we avoid tracking that dependency here to prevent
+           cycles in the dependency hierarchy. *)
+        >>| List.filter ~f:(fun dep -> not (Id.equal t.id dep.id))
```
rgrinberg, line 276: I still don't understand why this check is correct. If the intent is to make sure post deps never introduce cycles, this isn't enough as it only removes cycles of length 1.

gridbugs: Hmm I need to think about this some more. I see that since your changes the tests are now failing with:
```
     - package bar
  [1]
```
which is the type of memo cycle that this check was trying to prevent.

But as you point out, memo cycles may still happen if the loop would be longer than 1. I'll think about this some more today.

gridbugs:
Ok thinking about this a bit more, our current approach of saying that a package will depend on the post dependencies of all of its dependencies can't be correct. If `a` depends on `b`, and `b` post-depends on `a`, this would imply that `a` depends on `a`. I initially thought this was fine, and we could just ignore such cases (that was the point of my check for cycles of length 1) but as @rgrinberg points out, longer cycles are still possible, such as if `a` depends on `b`, `b` post-depends on `c`, and `c` depends on `a`, which would result in a cycle between `a` and `c`.

I've left a comment on the original issue (https://github.com/ocaml/dune/issues/10670#issuecomment-2235223905) so we can discus this further with @dra27.

gridbugs:
@rgrinberg while we work out the correct way to build/install post dependencies in dune, I'd like to merge this PR minus the changes to pkg_rules (so it would add post dependencies in a lockfile field but not attempt to install them). Currently dune isn't able to solve most projects due to the post dependency in the ocaml-base-compiler package, and merging the changes to the solver from this PR would fix that problem and unblock some of the work the build-system team wants to do.

gridbugs:
I just added a test I forgot to check in yesterday that checks the order in which post deps are installed.

rgrinberg:
I merged the solving bit to unlock other work while we figure out how post deps actually affect the rules.

rgrinberg:
This can be revisited once we are more certain how post affect the build rules

PR #10730, opened 2024-07-16
test: add for opam switch name with absolute path

maiste:
This PR provides a test to display the error input when trying to use an absolute path in a opam switch context.

Linked to #10723


test/blackbox-tests/test-cases/github10721.t
```diff
@@ -0,0 +1,19 @@
+
```
emillon, line 1:
```suggestion
```

emillon (approved):

PR #10733, opened 2024-07-16
chore: Add Cache Directory Tagging Specification

### LINKED ISSUE #9054: Conform to the Cache Directory Tagging Specification
This is a convention that allows cache directories to be marked as cache directories allowing programs such as archival and backup utilities to ignore them safely.

The convention consists of adding a file `CACHEDIR.TAG` with the following content:
```
Signature: 8a477f597d28d172789f06886806bc55
```
which is just the md5 of the string ".IsCacheDirectory".

Ideally we would just add this to `$XDG_CACHE_HOME/dune/`

https://bford.info/cachedir/

H-ANSEN:
This pull requests closes #9054 implementing the format specification found at https://bford.info/cachedir/.

I opted to follow @Alizter's specification in the issue only creating the `CACHEDIR.TAG` file when the default directory for the dune cache is being used. So if the environment variable `DUNE_CACHE_ROOT` is set this file will not be generated! Additionally, I omitted adding comments to the file but the specification does say that comments beginning with a `#` character are permitted after the signature, giving the following example:

```
Signature: 8a477f597d28d172789f06886806bc55
# This file is a cache directory tag created by (application name).
# For information about cache directory tags, see:
#	http://www.brynosaurus.com/cachedir/
```

### DIFF
--- a/src/dune_cache_storage/layout.ml
+++ b/src/dune_cache_storage/layout.ml
@@ -81,8 +81,28 @@ let file_path = Versioned.file_path Version.File.current
 let value_storage_dir = Versioned.value_storage_dir Version.Value.current
 let value_path = Versioned.value_path Version.Value.current
 
+let is_default_root_dir () =
+  let root_dir_str = Path.to_string root_dir in
+  String.is_suffix root_dir_str ~suffix:"dune/db"
+;;
+
+(** Write a cache directory tagging file to the cache directory if and only if 
+    the default cache directory `$XDG_CACHE_HOME/dune/` is being used. This
+    convention allows achival and backup utilities to ignore the cache. See more
+    info at: https://bford.info/cachedir/ *)
+let write_cache_directory_tag () =
+  if is_default_root_dir () 
+  then (
+    let root_dir_str = Path.to_string root_dir in
+    let root_str = String.drop_suffix_if_exists root_dir_str ~suffix:"/db" in
+    let cache_tag_path = (Path.of_string root_str) / "CACHEDIR.TAG" in
+    let cache_tag_content = "Signature: 8a477f597d28d172789f06886806bc55" in
+    Io.write_file ~binary:false cache_tag_path cache_tag_content)
+;;
+
 let create_cache_directories () =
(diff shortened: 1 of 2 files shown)


emillon:
out of curiosity, what tool do you use that honors this specification?

Alizter:
My original motivation was rsync

emillon:
the spec says rsync could one day use this, but it doesn't seem that it has ever been implemented.

H-ANSEN:
> out of curiosity, what tool do you use that honors this specification?

I used `tar` on Linux with each of the following commands to package up my whole cache folder:
- `--exclude-caches`
- `--exclude-caches-all`
- `--exclude-caches-under`

emillon:
I'm not opposed to that feature if there's actual demand for it and it fixes a use case, but at the moment it looks like the spec is only implemented by gnu tar, and there are other ways to achieve this without modifying dune (in particular, since it's only added when the cache is in `XDG_CACHE_HOME`, why not ignore the whole cache dir?).

To put it differently, the issue is tagged `good-first-issue` but I think this requires some validation first.

H-ANSEN:
Yea.. I totally see what you mean. There could be an argument made that it makes more sense to include the tagging file when the cache directory is moved from its default. I don't think I can provide much more validation, I mostly went after this issue since it was marked `good-first-issue`.

madroach:
Popular backup tools like restic ([since 2017](https://github.com/borgbackup/borg/commit/1d5410a3cfaf8ae3127d3b664c15a65e8bba143a)), borg ([since 2014](https://github.com/restic/restic/commit/dbda89254204e53efe3d7ce4abf283d5dcca56d0)) and duplicati ([since 2018](https://github.com/duplicati/duplicati/commit/83950e78c62f0b03f3e207cbd3db3a15b64c4896)) support `CACHEDIR.TAG`.

It should always be added when the dune cache directory is created, no matter whether in the default or custom location.

rgrinberg:
No objection to this feature if anybody cares enough to revive it.

PR #10737, opened 2024-07-17
test: do not use unix in implicit-transitive-deps

emillon:
Using the unix library will cause a warning on 5.2.
Instead we use a library that is defined in the test.


nojb:
What is the warning?

emillon:
it's the "unix library has been implicitly added to the search path" message.

PR #10738, opened 2024-07-17
feat: Special Variable %{file-available:<path>}

### LINKED ISSUE #6016: Special variables: add %{file-available:<file>} syntax
## Desired Behavior

In some situations, especially when mixing dune and non-dune subprojects, we would like to be able to conditionally depend on the existence of a certain file, that is, write:

    (enabled_if %{file-available:<file>})

Which would return `true` if `<file>` exists, or `false` otherwise.

I could not find a way to emulate this behavior with the current boolean functions in Dune, at least not in a simple way (at the end I mention a possible workaround, which seems full of complications).

In my particular use case, this would decide whether a given test (which is built using a dune rule) should be run. This test depends on the existence of a `node_modules` directory, that is, whether the user ran `npm install` before running the test. The `npm install` part is completely optional: users may not want to run it (too costly); or they may be unable to do so (lack of internet connectivity); or it may be an optional component only available in certain configurations.

In any case, there is currently no library or executable associated with the subproject; so I cannot readily use `lib-available` or `bin-available`.

I also tried using `read:`, since the documentation says *`read:<path>` expands to the contents of the given file*, in case a file containing the string *true* might be interpreted as the `true` value. This did not work, however.

Using the existence of a file to conditionally enable some feature seems like it could be useful in several scenarios. For instance, the [stamp idiom](https://www.technovelty.org/tips/the-stamp-idiom-with-make.html) is based on creating files to indicate whether certain stages were reached. It could be abused, though.

To check the feasibility of the required feature, I quickly hacked the dune source code to test, and it seems to work. But seeing the memoization that takes place inside Dune made me realize that there is an issue if the file in question could be created or erased *during* the build. The user should be duly notified that the file should *not* be modified by the build, otherwise all bets are off. Despite that restriction, the code to support it seems simple (except for relative path issues, which require a bit more care).

Overall, this works like an extension of `bin-available`.

### Workaround in the absence of this feature

This request does smell a bit of feature creep. It is probably always possible to replace the existence check of a file with a library, especially if the directory containing it is writable; for instance, after running `npm install` in my example, which would create a `node_modules` directory, I might as well run:

    echo "(library (name node_modules))" > dune

To create an empty library and then use `%{lib-available:node_modules}` to check its existence. But if the directory already has a `dune` file, it might require creating a subdirectory; otherwise, if the directory is *not* handled by dune (e.g. not present in a `dirs` stanza), that might also be a problem. Overall, for each file to check, I might have to create a new library.
So, it is *possible* to sidestep the need for `file-available`, but it is very *inconvenient* to do so. Unless there are already some better techniques to emulate this behavior.

## Example

```
(enabled_if %{file-available:node_modules})
```

The above rule should run only if the `node_modules` file/directory exists.

H-ANSEN:
Special variable expanding to a boolean value depending on the existence of a file at the given path. This pull is in reference to the feature request discussed at #6016. I was able to implement tests confirming that generated files could also be detected.

Closes #6016


test/blackbox-tests/test-cases/file-available.t
```diff
@@ -0,0 +1,77 @@
...
+  $ dune build @runtest0
+  dune-project available: true
+  bin/foo.ml available: true
+  bar.ml available: true
+
+Test to see that some files that do not exist are not available.
+
+  $ cat > dune <<EOF
+  > (rule
+  >   (alias runtest1)
```
rgrinberg, line 33: We use 1 space of indentation in our tests

src/dune_rules/expander.ml
```diff
@@ -663,6 +663,13 @@ let expand_pform_macro
            let* artifacts_host = t.artifacts_host in
            let+ b = Artifacts.binary_available artifacts_host s in
            b |> string_of_bool |> string))
+  | File_available ->
+    Direct (
+      Without
+        (let open Memo.O in
+         let file_path = Path.of_string s in
+         let+ available = Fs.file_exists file_path in
```
rgrinberg, line 671: This isn't going to work if `file_path` is a target of a rule. Is that intentional?

test/blackbox-tests/test-cases/file-available.t
```diff
@@ -0,0 +1,77 @@
...
+  > (rule
+  >   (with-stdout-to gen.txt (echo "generated file text")))
+  > (rule
+  >   (write-file gen1.txt "another generated file")) 
+  > (rule
+  >   (alias runtest2)
+  >   (action
+  >     (progn
+  >       (echo "gen.txt available: %{file-available:_build/default/gen.txt}\n")
+  >       (echo "gen1.txt available: %{file-available:_build/default/gen1.txt}\n"))))
```
rgrinberg, line 56: You should not have to qualify the context with _build/default. The rules must do that for you.

rgrinberg (changes requested):

test/blackbox-tests/test-cases/file-available.t
H-ANSEN: Could you expand upon this a bit, I'm not sure I fully understand what you mean. How I interpret this is, by default the macro invocation should be looking for files staring in the base directory `_build/default`?

rgrinberg: All rules have the following property: they work in any "context". The path `_build/default/gen.txt` that you referred is a path from the `default` context. So it means that your rule cannot possibly work for other contexts. This is because you have implemented `file-available` to only work for source files.

src/dune_rules/expander.ml
H-ANSEN: I think I might have this working I implemented added a simple test to the same `file-available.t` test file.

src/dune_rules/expander.ml
```diff
@@ -663,6 +663,13 @@ let expand_pform_macro
            let* artifacts_host = t.artifacts_host in
            let+ b = Artifacts.binary_available artifacts_host s in
            b |> string_of_bool |> string))
+  | File_available ->
+    Direct
+      (Without
+        (let open Memo.O in
+         let path = relative ~source dir s in
+         let+ available = Fs.file_exists path in
```
rgrinberg, line 671: I think you just want `Build_system.file_exists` here. You should never have a source path here. Source paths are transparently copied to _build/default for the rules to consume. If you can observe a source path here, then we have a problem.

rgrinberg:
Thanks

Issue #10741, opened 2024-07-18
Oneshot webserver used in some pkg tests fails to send large files


gridbugs:
This issue seems to be related to the webserver shutting down before the transfer is complete. I thought I fixed the issue in https://github.com/ocaml/dune/pull/10705, but it still happens for sufficiently large files. As this is a race between the transfer completing and the server shutting down, the maximum size for a successfully transferred file is platform-specific. On my linux machine, I can consistently reproduce this issue attempting to transfer a 20MB file.

Reproduced in https://github.com/ocaml/dune/pull/10742

rgrinberg:
Maybe we should switch to a more reliable HTTP server in our tests. No reason to develop one inside the repo.

gridbugs:
Yeh fair point. Maybe there's something in opam that fits our needs.

moyodiallo:
I've tried reducing the buffer of `webserver_oneshot` from 4096 to 512 and it turns out working super well. I even tried to test it with 1000M #10742 and still working. Maybe the transfer is faster when the buffer is lower.

PR #10742, opened 2024-07-18
Add repro for gh10741

gridbugs:
Repro for https://github.com/ocaml/dune/issues/10741

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/dune
+++ b/test/blackbox-tests/test-cases/pkg/dune
@@ -44,7 +44,8 @@
   e2e
   source-caching
   tarball
-  extra-sources))
+  extra-sources
+  oneshot-webserver-large-file-curl-error))
 
 (cram
  (deps %{bin:md5sum})
--- a/test/blackbox-tests/test-cases/pkg/oneshot-webserver-large-file-curl-error.t
+++ b/test/blackbox-tests/test-cases/pkg/oneshot-webserver-large-file-curl-error.t
@@ -1,0 +1,17 @@
+Reproduction for an issue with the oneshot webserver used for testing
+package management, where attempting to send a large file causes the
+connection to be reset before the transfer completes.
+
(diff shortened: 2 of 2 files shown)


test/blackbox-tests/test-cases/pkg/oneshot-webserver-large-file-curl-error.t
```diff
@@ -0,0 +1,17 @@
+Reproduction for an issue with the oneshot webserver used for testing
+package management, where attempting to send a large file causes the
+connection to be reset before the transfer completes.
+
+Make a large file:
+  $ dd if=/dev/zero of=./large-file iflag=fullblock,count_bytes status=none count=20M
```
moyodiallo, line 6: I have tested on my machine, and it starts failing from 14M, yeah 20M could be better number.

moyodiallo (approved):

rgrinberg (changes requested):
Can you disable this test by default? Writing an additional 20 megs every time we run the test seems a bit annoying.

test/blackbox-tests/test-cases/pkg/dune
```diff
@@ -44,7 +44,8 @@
   e2e
   source-caching
   tarball
-  extra-sources))
+  extra-sources
+  oneshot-webserver-large-file-curl-error))
```
rgrinberg, line 48: This test depends on curl as well

rgrinberg:
Actually, the file size didn't need to be 20M to repro the problem. If it's 1M, then it should be enough to leave on by default.

gridbugs:
Ok I'll change it to 1M. On my machine I can't reproduce the problem with a 1M file, but github actions seem to be more sensitive to this issue than my machine is.

gridbugs:
This is no longer needed as the issue was fixed in https://github.com/ocaml/dune/pull/10753

PR #10647, opened 2024-06-13
pkg: build and make ocamlformat dev-tool available

moyodiallo:
Attempts to implement #10688.

Reopen the PR #10613 to target `main` branch instead of toolchain one.

To avoid dune capturing a dev-tool like `ocamlformat` when using `dune build @fmt`. This PR is about to build this tool independently from dune-project dependencies, to avoid conflict. After a successful build it will be directly available whenever `dune build @fmt` or `dune fmt` is invoked. 

- [X] ocamlformat: build and make it available
- [x] ~~prioritize `dev tools` to `PATH` when `dune.lock` dir is present in the project.~~
- [x] Solve ocamlformat when `dune fmt` invoke it. (you're right here @Leonidas-from-XIV, it is needed to keep the pkg tests using `dune pkg lock` intact or facilitate adding new test for this PR)
- [x] Fix the fetch of ocamlformat dev tool dependencies ( It turns out that the fetch rules are missing those packages)
- [x] add a cram test

The next works that would be done:
- The ocamlformat stanza `(ocamlformat <version>)` could be added in another PR in order to make this one simple. 
- A PR for `odoc` tool will the same as this PR.

PR #10744, opened 2024-07-18
chore: use with-dev-setup to manage opam test deps

### LINKED ISSUE #6939: Issues to compile dune from scratch
When opening an issue, following issue instructions i began the process to add a test & open PR. I followed the [hacking](https://github.com/ocaml/dune/blob/master/doc/hacking.rst) instruction from an empty switch, but wasn't able to easily compile.

## Expected Behavior

Update guidelines/makefile to ease install from scratch, on an empty *environment*.

### hacking.rst
Following hacking guidelines, i went for the simple (on an empty switch).
```
$ make dev
$ ./dune.exe build @check
$ ./dune.exe runtest
```

`make dev` builds `dune.exe`, but fails because of missing dependencies
```
$ make dev
ocamlc -output-complete-exe -w -24 -g -o .duneboot.exe -I boot unix.cma boot/libs.ml boot/duneboot.ml
./.duneboot.exe
./_boot/dune.exe build @install
File "otherlibs/dune-rpc-lwt/src/dune", line 4, characters 38-46:
4 |  (libraries result csexp dune_rpc lwt lwt.unix unix)
                                          ^^^^^^^^
Error: Library "lwt.unix" not found.
-> required by library "dune-rpc-lwt" in
   _build/default/otherlibs/dune-rpc-lwt/src
-> required by _build/default/META.dune-rpc-lwt
-> required by _build/install/default/lib/dune-rpc-lwt/META
-> required by _build/default/dune-rpc-lwt.install
-> required by alias install
make: *** [Makefile:53: dev] Error 1
```
However, it builds  `dune.exe`, but checking that everything compile or launching test (the 2 other commands) fail as there is other missing dependencies (ppx_inline_test.config, ppx_inline_test.config, core_bench.inline_benchmarks, lwt, stdio, etc.).

#### Reproduction
```
$ opam switch create test-sw --empty # or an already empty switch
$ git clone git@github.com:ocaml/dune.git
$ cd dune
$ git reset --hard e8ad4cd5db8b9b2
$ make dev
$ ./dune.exe build @check
$ ./dune.exe runtest
```

### Using `make dev-switch`
After discussion with @emillon, i was advised to use `make dev-switch`. Here it fails too, but for another reason: opam can't install dependencies at switch creation as they depend on dune.
```
test "test-sw" = "/tmp/dune" || \
	opam switch create -y . 4.14.0 --deps-only --with-test

<><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><><><>
Switch invariant: ["ocaml-base-compiler" {= "4.14.0"} | "ocaml-system" {= "4.14.0"}]
The following actions will be performed:
=== install 38 packages
  ∗ base                  v0.15.1 [required by ppx_expect]
[...]
#=== ERROR while compiling cppo.1.6.9 =========================================#
# context     2.2.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.4.14.0 | https://opam.ocaml.org
# path        /tmp/dune/_opam/.opam-switch/build/cppo.1.6.9
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p cppo -j 6
# exit-code   1
# env-file    ~/.opam/log/cppo-78510-577e06.env
# output-file ~/.opam/log/cppo-78510-577e06.out
### output ###
# bwrap: execvp dune: No such file or directory


<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
┌─ The following actions failed
│ λ build cppo                1.6.9
│ λ build csexp               1.5.1
│ λ build jane-street-headers v0.15.0
│ λ build ocaml-compiler-libs v0.12.4
│ λ build pp                  1.1.2
│ λ build ppx_derivers        1.2.1
│ λ build re                  1.10.4
│ λ build result              1.5
│ λ build sexplib0            v0.15.1
│ λ build stdlib-shims        0.3.0
[...]
```
#### Reproduction
```
$ opam switch create test-sw --empty # or an already empty switch
$ git clone git@github.com:ocaml/dune.git
$ cd dune
$ git reset --hard e8ad4cd5db8b9b2
$ make dev-switch
```

### Using `make dev-switch` on an empty local switch
Finally, I decided to create a local empty switch to not let the makefile create one, and go directly to dependencies install. It worked well! I was able to build dune, check everything, and run testsuite. The only remaining issue is a missing binary on a test: `node`
```
$ ./dune.exe build @check
File "test/blackbox-tests/test-cases/dune", line 18, characters 8-19:
18 |   (deps %{bin:node})
             ^^^^^^^^^^^
Error: Program node not found in the tree or in PATH
```
#### Reproduction
```
$ opam switch create test-sw --empty # or an already empty switch
$ git clone git@github.com:ocaml/dune.git
$ cd dune
$ git reset --hard e8ad4cd5db8b9b2
$ opam switch create . --empty
$ make dev-switch
$ make dev
$ ./dune.exe build @check
$ ./dune.exe runtest
```
However, it wasn't a blocker for new test addition.

## Specifications

- Version of `dune` (output of `dune --version`): main branch, e8ad4cd5db8b9b2
- Version of `ocaml` (output of `ocamlc --version`): irrelevant
- Operating system (distribution and version): debian

## Additional information 

Full log can be found [here](https://gist.github.com/rjbou/bb480986eee45dae9e0699513d9c8dd7).

Thanks!

emillon:
The goal is to simplify setting up the test dependencies. We use `{with-dev-setup}` in `dune.opam`. By default, these dependencies are not installed, but passing `--with-dev-setup` will install them.

Because the test dependencies are now expressed in an opam file, this is more expressive, and it makes it possible to use opam filters to install different dependencies on different configurations for example (#10712 is an example).

The source of truth for the ocamlformat version is moved to the opam file, so the parsing code in `make install-ocamlformat` and `flake.nix` is adapted. This means that `ocamlformat` will not be able to check its version at runtime anymore, but CI will do it and it's likely that Dune is going to manage `.ocamlformat` itself in the future (#10647).

`make dev-deps` continues working as before, but it can now be expressed as an opam command. `make dev-switch` is now an idempotent way to keep the switch in sync (which should cover #6939 as it was reported).

Fixes #6939

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,25 +6,6 @@ DESTDIR_ARG := $(if $(DESTDIR),--destdir $(DESTDIR),)
 INSTALL_ARGS := $(PREFIX_ARG) $(LIBDIR_ARG) $(DESTDIR_ARG)
 BIN := ./_boot/dune.exe
 
-# Dependencies used for testing dune, when developed locally and
-# when tested in CI
-TEST_DEPS := \
-lwt \
-cinaps \
-"csexp>=1.3.0" \
-js_of_ocaml \
-js_of_ocaml-compiler \
-"mdx>=2.3.0" \
-menhir \
-ocamlfind \
-ocamlformat.$$(awk -F = '$$1 == "version" {print $$2}' .ocamlformat) \
-"odoc>=2.4.0" \
-"ppx_expect>=v0.16.0" \
-ppx_inline_test \
-ppxlib \
-ctypes \
-"utop>=2.6.0" \
-"melange>=4.0.0-414"
 # Dependencies recommended for developing dune locally,
 # but not wanted in CI
 DEV_DEPS := \
@@ -68,31 +49,33 @@ uninstall:
 .PHONY: reinstall
 reinstall: uninstall install
 
+.PHONY: install-ocamlformat
 install-ocamlformat:
 	opam install -y ocamlformat.$$(awk -F = '$$1 == "version" {print $$2}' .ocamlformat)
 
-dev-depext:
-	opam depext -y $(TEST_DEPS)
-
 .PHONY: dev-deps
-dev-deps:
-	opam install -y $(TEST_DEPS)
+dev-deps: install-ocamlformat
+	opam install -y . --deps-only --with-dev-setup
 
 .PHONY: coverage-deps
 coverage-deps:
 	opam install -y bisect_ppx
 
 .PHONY: dev-deps-sans-melange
-dev-deps-sans-melange:
-	opam install -y $(TEST_DEPS)
+dev-deps-sans-melange: dev-deps
 
 .PHONY: dev-switch
 dev-switch:
 	opam update
 # Ensuring that either a dev switch already exists or a new one is created
-	test "$(shell opam switch show)" = "$(shell pwd)" || \
-		opam switch create -y . $(TEST_OCAMLVERSION) --deps-only --with-test
-	opam install -y --update-invariant ocaml.$(TEST_OCAMLVERSION) $(TEST_DEPS) $(DEV_DEPS)
+	if test -d _opam ; then \
+		opam install -y --update-invariant ocaml.$(TEST_OCAMLVERSION); \
+	else \
+		opam switch create -y . $(TEST_OCAMLVERSION) --no-install ; \
+	fi
+	opam install -y . --deps-only --with-test --with-dev-setup
+	$(MAKE) install-ocamlformat
+	opam install -y $(DEV_DEPS)
 
 .PHONY: test
 test: $(BIN)
--- a/doc/hacking.rst
+++ b/doc/hacking.rst
@@ -24,6 +24,12 @@ To create a directory-local opam switch with the dependencies necessary to build
 
   $ make dev-switch
 
+This can also be used to keep the switch updated when dependencies change.
+
+The ``Makefile`` also has a ``make dev-deps`` which will install just the
+dependencies used by tests. These are marked ``{ with-dev-setup }`` in Dune's
+opam file.
+
 Bootstrapping
 =============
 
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -68,31 +49,34 @@ uninstall:
 .PHONY: reinstall
 reinstall: uninstall install
 
-install-ocamlformat:
-	opam install -y ocamlformat.$$(awk -F = '$$1 == "version" {print $$2}' .ocamlformat)
+OCAMLFORMAT_VERSION := $(shell opam show ./dune.opam --just-file |grep ocamlformat|cut -d'"' -f4 )
```
Leonidas-from-XIV, line 52: I think you can use `--raw` and `-fdepends` here to select the right section of the OPAM file. Personally I'd probably use `grep` to get the version instead of `cut` since it feels a it roundabout to use `"` as delimiter but well, it does work.

Leonidas-from-XIV (approved):
I think that's a nice step to make the set-up more streamlined. The less `Makefile` the better.

Makefile
emillon: `--raw` is implied by `--just-file`. `-fdepends` will restrict a bit the input but it's still necessary to select the line that contains ocamlformat, so it will be one `grep` followed by one `grep -o`, is that what you had in mind?

rgrinberg:
> likely that Dune is going to manage .ocamlformat itself in the future

I don't think this is likely. The version in the .ocamlformat should be restored unless there's a recommendation from the ocamformat team to add it back.

emillon:
> The version in the .ocamlformat should be restored unless there's a recommendation from the ocamformat team to add it back.

Do you mean *to remove it*?

In any case, we can leave ocamlformat from this change if that's a bit controversial (see #10745).

Makefile
Leonidas-from-XIV: Indeed, but by specifying `-fdepends` you make sure that the ocamlformat mentioned is in the depends field and not e.g. in `conflicts` or something. It's not a big deal either way as we control the opam file for the most part.

moyodiallo:
> This means that ocamlformat will not be able to check its version at runtime anymore, but CI will do it and it's likely that Dune is going to manage .ocamlformat itself in the future (https://github.com/ocaml/dune/pull/10647).

This does not change anything whatever we decide about `.ocamlformat`. We could adapt dune easily.

PR #4327, opened 2021-03-08
Ensure Dune can build itself in CI

dra27:
The testsuite is already failing on the 2.8 branch - this additional step fails the 4.08 and 4.09 test builds as well.

The problem doesn't show in the CI jobs which run tests because they install `ocamlfind`

Signed-off-by: David Allsopp <david.allsopp@metastack.com>

PR #10746, opened 2024-07-18
ci: do not bootstrap twice

emillon:
`make bootstrap` does exactly `make -B _boot/dune.exe`, which is done just before, so we bootstrap twice in the workflow.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -95,9 +95,6 @@ jobs:
       # away this makes it possible to see build errors as soon as possible
       - run: opam exec -- make _boot/dune.exe
 
-      # Ensure Dune can build itself
-      - run: opam exec -- make bootstrap
-
       - name: Install deps
         run: |
           opam install . --deps-only --with-test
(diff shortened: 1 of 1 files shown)


emillon:
I'm not sure what the history is here. This second bootstrap has been added in #6308; the original command has been added in #4327 but with this profile selected, dune is not really rebuilt anyway.

rgrinberg:
I think the purpose of this command is to ensure that the .install file is generated correctly. Could you confirm that the .install file is generated without this command?

rgrinberg (approved):
LGTM.

.github/workflows/workflow.yml
```diff
@@ -93,10 +93,7 @@ jobs:
 
       # dune doesn't have any additional dependencies so we can build it right
       # away this makes it possible to see build errors as soon as possible
-      - run: opam exec -- make _boot/dune.exe
```
rgrinberg, line 96: Good to see another path eliminated from the yaml file. These are always a pain to keep in sync

PR #10747, opened 2024-07-18
bench: use setup ocaml v3

emillon:


emillon:
(merging to unblock `main`)

PR #10669, opened 2024-06-25
chore: use 5.2.0 for tests

emillon:

PR #10749, opened 2024-07-19
ci: remove DKML builds

emillon:
We now have MSVC and mingw tests (see #10712) so we can use the native opam 2.2 builds instead of dkml. This also allows us to move the test version forward (#10669).


emillon:
Thanks a lot @jonahbeckford, these builds have been useful in the meantime.

Leonidas-from-XIV (approved):
Nice, good to have one common OPAM path for all platforms now.

jonahbeckford:
Excellent!

PR #10750, opened 2024-07-19
fix(pkg): webserver-oneshot when downloading a large file

moyodiallo:
Fixes https://github.com/ocaml/dune/issues/10741

We should merge this PR only after the merge of #10742.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/gh10741.t
+++ b/test/blackbox-tests/test-cases/pkg/gh10741.t
@@ -1,9 +1,9 @@
-Reproduction for an issue with the oneshot webserver used for testing
+The issue with the oneshot webserver used for testing
 package management, where attempting to send a large file causes the
-connection to be reset before the transfer completes.
+connection to be reset before the transfer completes, is now fixed.
 
 Make a large file:
-  $ dd if=/dev/zero of=./large-file iflag=fullblock,count_bytes status=none count=20M
+  $ dd if=/dev/zero of=./large-file iflag=fullblock,count_bytes status=none count=64M
 
 Run the oneshot webserver to serve the large file:
   $ webserver_oneshot --content-file ./large-file --port-file port.txt &
@@ -11,7 +11,3 @@ Run the oneshot webserver to serve the large file:
 
 Download the large file with curl:
   $ curl -sS -o large-file-copy http://localhost:$(cat port.txt) > /dev/null
-  curl: (56) Recv failure: Connection reset by peer
-  [56]
-
-  $ wait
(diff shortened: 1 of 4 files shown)


test/blackbox-tests/test-cases/pkg/gh10741.t
```diff
@@ -0,0 +1,13 @@
+The issue with the oneshot webserver used for testing
+package management, where attempting to send a large file causes the
+connection to be reset before the transfer completes, is now fixed.
+
+Make a large file:
+  $ dd if=/dev/zero of=./large-file iflag=fullblock,count_bytes status=none count=64M
```
moyodiallo, line 6: It was tested with count=10000M and it works.

test/blackbox-tests/test-cases/pkg/gh10741.t
```diff
@@ -0,0 +1,13 @@
...
+
+Run the oneshot webserver to serve the large file:
+  $ webserver_oneshot --content-file ./large-file --port-file port.txt &
+  $ until test -f port.txt ; do sleep 0.1; done
```
maiste, line 10: I'm not sure to understand: why you need this file?

moyodiallo: Which file ?

moyodiallo: If it is the test file `gh10741.t`, it coming from https://github.com/ocaml/dune/pull/10742 which reproduced the issue.

Leonidas-from-XIV: I think we should merge #10742 first and rebase the fix on it.

moyodiallo: For sure, we could go this.

maiste: Sorry, my question wasn't well written and clear. Why do you need to test the presence of the `port.txt` file in the test?
```
$ until test -f port.txt ; do sleep 0.1; done
```
If it is executed sequentially, we shouldn't have to wait, no?

Leonidas-from-XIV: The server is started in the background (`&`) thus by the time the next line runs, the file might not yet be created, so the next line waits until the server is up and creates the port file (which is its way of signalling "I'm ready to accept a connection").

The server can't be started in the foreground, as then the test never terminates.

rgrinberg:
Fixed in #10753

PR #10752, opened 2024-07-19
feature(pkg): set [post = true] when solving

rgrinberg:


test/blackbox-tests/test-cases/pkg/post-deps-solving.t
```diff
@@ -0,0 +1,86 @@
...
+
+  $ mkpkg bar
+
+  $ mkpkg foo <<EOF
+  > depends: [ "bar" {post} ]
+  > EOF
+
+We don't need bar, but we still include it:
+
+  $ solve foo
```
rgrinberg, line 14: We shouldn't lock packages whose sources we aren't actually using for the build. We can keep this undesirable property for now until we figure out how post deps actually affect the build.

PR #10753, opened 2024-07-20
Revert "test: fix oneshot webserver for large binary files (#10705)"

rgrinberg:
This reverts commit 5905292de45d62a8c204bac2f86498fbde4ac1f9.

Also properly implement request flushing to fix the actual issue and gets rid of workarounds for buggy tests.


rgrinberg:
Actually, just reverting #10750 also fixes the repro in https://github.com/ocaml/dune/pull/10742/files. So I think we need a proper repro for the actual bug.

rgrinberg:
Okay, so the real problem was that our server wasn't reading the HTTP requests. Should be good after this PR.

Leonidas-from-XIV:
It looks like this also removed the improvement from #10714?

PR #10754, opened 2024-07-21
fix(pkg): pinning depopts

rgrinberg:
depopts aren't conjunctions, but disjunctions


test/blackbox-tests/test-cases/pkg/pin-stanza/pin-depopts.t
```diff
@@ -37,7 +37,5 @@ https://github.com/ocaml/dune/issues/10739
   > (lang dune 3.16)
```
gridbugs, line 37: Remember to update the comment above this part of test claiming that this doesn't work correctly.

PR #10759, opened 2024-07-22
test(site): cat whole install file

emillon:
This output changes slightly in 5.2 (since a file is compressed), so the goal of this PR is to make the diff cleaner afterwards.

(without expanding the install file, the upgrade to 5.2 shows a change in the count, which is harder to interpret)

### DIFF
--- a/otherlibs/dune-site/test/run.t
+++ b/otherlibs/dune-site/test/run.t
@@ -177,11 +177,30 @@ Test with an opam like installation
 
   $ dune install -p a --create-install-files a --prefix "_destdir"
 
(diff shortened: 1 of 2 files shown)


Leonidas-from-XIV:
Is the order of the entries fixed? I'm a little bit worried that this might introduce flakiness.

emillon:
Yes, it's built from install entries which are sorted there:

https://github.com/ocaml/dune/blob/3.16.0/src/dune_rules/install_rules.ml#L569-L577

Leonidas-from-XIV (approved):

PR #10760, opened 2024-07-22
doc: add an example for variable name :version

### LINKED ISSUE #10758: Dune generated opam files are not correct when the version is `= version`
## Expected Behavior
When `generate_opam_files` is enabled and a dependency has the constraint `(= version)`, the opam file should contain the constraint `{= version}`

## Actual Behavior
The constraint is currently `{= "version"}`.

## Reproduction

By using the following `dune-project` file:
```
(lang dune 3.13)

(generate_opam_files true)

(package
  (name bug)
  (depends (bug-lib (= version)))
)
```
and the following `dune` file:
```
(executable
 (name bug))
```
with an empty `bug.ml` file, the command `dune build` will generate the `bug.opam` file.

## Specifications

- Operating system (distribution and version): Debian Unstable

emillon:
Fixes #10758

### DIFF
--- a/doc/howto/opam-file-generation.rst
+++ b/doc/howto/opam-file-generation.rst
@@ -113,9 +113,10 @@ General Notes and Tips
   keep the corresponding opam fields in a ``pkg.opam.template`` file. See
   :doc:`../reference/packages`.
 - It is not necessary to specify ``(version)``, this will be added at release
-  time if you use `dune-release`_.
-
-.. _dune-release: https://github.com/tarides/dune-release
+  time if you use `dune-release <https://github.com/tarides/dune-release>`_.
+- To generate an opam variable such as ``version``, use a colon ``:`` followed
+  by the name of the variable. For example, to generate ``a { = version }`` in
+  the opam file, use ``(a (= :version))`` in ``dune-project``.
 
 Generating Opam Files
 ---------------------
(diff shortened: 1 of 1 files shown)


doc/howto/opam-file-generation.rst
```diff
@@ -113,9 +113,10 @@ General Notes and Tips
   keep the corresponding opam fields in a ``pkg.opam.template`` file. See
   :doc:`../reference/packages`.
 - It is not necessary to specify ``(version)``, this will be added at release
-  time if you use `dune-release`_.
-
-.. _dune-release: https://github.com/tarides/dune-release
+  time if you use `dune-release <https://github.com/tarides/dune-release>`_.
+- To generate an opam variable such as ``version``, use a colon ``:`` followed
```
maiste, line 117: Do we have a page in the manual that specifies the variable behavior (such as `:with-test`, ...) and how they are used in dune? If yes, it could be useful to have a reference to it?
WDYT?

emillon: there's https://github.com/ocaml/dune/blob/main/doc/explanation/opam-integration.rst which explains the dune/opam integration. we could add in as a general `..seealso::` at the end. sounds good?

maiste: Yes, definitely :+1: 

However, I realized the first part of my suggestion is not relevant in this PR and should be address in another one. IIRC, there is no place (apart from examples) that `:with-test ` would be converted into `{with-test}` isn't it?

emillon: Actually the `version` example is already in the reference for `(package)` but a xref doesn't hurt. I'll add both.

emillon:
Thanks!

PR #10763, opened 2024-07-23
chore: use 5.1.1 for tests

emillon:
This bumps the version of OCaml used in tests to 5.1.1.

The changes should be straightforward to review:
- version numbers are changed
- CI is adjusted
- some hashes are changed
- the layout of the compiler is changed
- some error messages are changed

Interesting changes to note:
- on OCaml 5, with the dev profile, jsoo triggers a warning
- the location of cmt files in install files is changed, because they are compressed and binary substitution does not happen

At the time this PR is opened, 5.2 is available but requires more changes around warnings and alerts in alias modules so it will be updated in a second step.

Also, there are a few components that we're leaving behind:
- the monorepo benchmarks can be updated separately
- the benchmark in `bench.Dockerfile` can be updated separately
- coq cannot use OCaml 5 at the moment so it's forced to use 4.x
- 5.1.1 doesn't support MSVC so we leave the build on 4.14 for now.


emillon:
@ElectreAAS this is now ready, the individual commits should make sense for review.

otherlibs/dune-site/test/run_2_9.t
```diff
@@ -181,7 +181,7 @@ Test with an opam like installation
     "_build/install/default/lib/a/a__.cmx"
     "_build/install/default/lib/a/a__.ml"
     "_build/install/default/lib/a/a__Sites.cmi"
-    "_destdir/_destdir/lib/a/a__Sites.cmt"
+    "_build/install/default/lib/a/a__Sites.cmt"
```
rgrinberg, line 184: Is this change expected?

emillon: Yes, I put the reason in the commit message for squashing:

> the source is different because since cmt files are compressed in 5.1, they can be copied straight from the build directory (the placeholder is not detected)

christinerose (approved):

.github/workflows/workflow.yml
```diff
@@ -30,33 +30,29 @@ jobs:
...
         ocaml-compiler:
-          - 4.14.x
+          - 5.1.x
         include:
           # OCaml trunk:
           - ocaml-compiler: ocaml-variants.5.3.0+trunk
             os: ubuntu-latest
             skip_test: true
-          # OCaml 5:
+          # OCaml 4:
```
ElectreAAS, line 39: Isn't this a typo? Or should we have a separate 4.14 CI run here?

ElectreAAS (approved):
Bar my minuscule comment, everything looks good to me! Very happy to see this done :)

.github/workflows/workflow.yml
emillon: Yes indeed the plan was to add a 4.14 build-only CI instead of the 4.13 one.

rgrinberg (approved):

PR #10766, opened 2024-07-24
Organization In Source Stanza

### LINKED ISSUE #6723: Allow `(source (gitlab org/project/repo))` in dune-project
## Expected Behavior

`(source (gitlab org/project/repo))` should be a valid source.

## Actual Behavior

```
6 | (source (gitlab nomadic-labs/cryptography/ocaml-chia-vdf))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Gitlab repository must be of form user/repo
```

## Reproduction

Update the dune-project here: https://gitlab.com/nomadic-labs/cryptography/ocaml-chia-vdf.

## Specifications

- Version of `dune` (output of `dune --version`):
3.6.1
4.14
- Operating system (distribution and version):
Ubuntu 22.04

H-ANSEN:
Adds support to the `source` stanza for Gitlab style organization repositories. This PR is in reference to and would close #6723 if merged.

Putting this out there as a draft for feedback. Right now its implemented so that a 'link' of the type `organization/project/repo` can be prefixed with any of the supported platforms *(github, bitbucket, sourcehut, or gitlab)* in the source stanza despite the fact that I think only Gitlab uses this convention. It would be possible to present an error in the case that this format is used with a platform which does not support it but I was unsure if I should add this.

### DIFF
--- a/test/blackbox-tests/test-cases/source-stanza.t
+++ b/test/blackbox-tests/test-cases/source-stanza.t
@@ -1,0 +1,114 @@
+Test 'source' stanza compatibility with both user and organization paths from
+the supported 'github', 'gitlab', 'sourcehut', and 'bitbucket'.
+
+Test a generated 'github' user repo
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.17)
+  > (name foo)
+  > (generate_opam_files true)
+  > (source (github user/repo))
+  > (package
+  >  (allow_empty)
+  >  (name foo))
+  > EOF
+
+  $ dune build
+  $ cat foo.opam | grep -i github
+  homepage: "https://github.com/user/repo"
+  bug-reports: "https://github.com/user/repo/issues"
+  dev-repo: "git+https://github.com/user/repo.git"
+
+Test a generated 'gitlab' user repo
+
+  $ sed -i -e '4s|.*|(source (gitlab user/repo))|' dune-project
+  $ dune build
+  $ cat foo.opam | grep -i gitlab
+  homepage: "https://gitlab.com/user/repo"
+  bug-reports: "https://gitlab.com/user/repo/-/issues"
+  dev-repo: "git+https://gitlab.com/user/repo.git"
+
+Test a generated 'sourcehut' user repo
+
+  $ sed -i -e '4s|.*|(source (sourcehut user/repo))|' dune-project
+  $ dune build
+  $ cat foo.opam | grep -i sr.ht
+  homepage: "https://sr.ht/~user/repo"
+  bug-reports: "https://todo.sr.ht/~user/repo"
+  dev-repo: "git+https://git.sr.ht/~user/repo"
+
+Test a generated 'bitbucket' user repo
+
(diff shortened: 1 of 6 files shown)


rgrinberg:
I think you should definitely limit this format to gitlab. Moreover, the types should reflect and it shouldn't be possible to construct say a github repo with a `organization * project * repo` triple. Basically, these helpers are here to prevent us from making mistakes when entering this data into the project file. It would be rather unfortunate if one could make a different type of mistake by mixing the different types of forges.

H-ANSEN:
Is there an established common practice for using `Syntax.since`? In [source_kind.ml](src/dune_lang/source_kind.ml) the function `enum` uses `Syntax.since` to denote the minimum version which supports a given forge, per the conversation at #6723 I was going to add a limitation to the organization style gitlab syntax but I'm unsure which version should be used? Should a future version be used `3.17`, the current version `3.16`, or something else?

rgrinberg:
The new format should be enabled starting from 3.17

H-ANSEN:
This should pass CI as long as all dependencies install correctly. Not sure why it failed to fetch that one dependency but I can't run CI again without pushing.

rgrinberg:
It's a transient failure that can safely be ignored

test/blackbox-tests/test-cases/source-stanza.t
```diff
@@ -0,0 +1,114 @@
+Test 'source' staza compatibility with both user and organization paths from
```
Leonidas-from-XIV, line 1:
```suggestion
Test 'source' stanza compatibility with both user and organization paths from
```

test/blackbox-tests/test-cases/source-stanza.t
```diff
@@ -0,0 +1,114 @@
...
+  Error: Sourcehut repository must be of form user/repo
+  Hint: The provided form 'org/proj/repo' is specific to Gitlab projects
+  [1]
+
+So far we have been using '(lang dune 3.17)' which supports gitlab organization
+style syntax, we will bump the version down and check to make sure an error is
+thrown telling us we need a more recent version of dune to use orginaziton
+syntax.
+
+  $ sed -i -e '1s|.*|(lang dune 3.0)|' dune-project
```
Leonidas-from-XIV, line 100:
```suggestion
  $ sed -i -e '1s|.*|(lang dune 3.16)|' dune-project
```
To make sure it's added in 3.17 and not, say, 3.11.

Leonidas-from-XIV (commented):
I've rerun the test in CI now, it passed.

Leonidas-from-XIV:
@H-ANSEN It's shaping up! Can you also update the `doc/reference/dune-project/generate_opam_files.rst` to add the newly supported gitlab syntax and that it's been added in Dune 3.17?

Leonidas-from-XIV (approved):

Leonidas-from-XIV:
Thanks for the contribution!

PR #10769, opened 2024-07-28
feature(pkg): toolchains

rgrinberg:
This is a replacement for #10719 without some of the undesirable parts.

There's a few issues remaining:
1. There's no cookie file being written for toolchains. We should that to keep things as uniform as possible with regular packages.
2. The digest doesn't include dependencies of the resolved packages. Again, we should include a digest of the transitive closure of deps (but without the random id's).
3. The check for installation is being done at rule generation time. This is wrong, and should be done at action execution time.

@gridbugs could you address these issues in separate PR's?

### DIFF
--- a/otherlibs/stdune/src/path.ml
+++ b/otherlibs/stdune/src/path.ml
@@ -488,6 +488,7 @@ module External : sig
   include Path_intf.S
   module Table : Hashtbl.S with type key = t
 
+  val root : t
   val relative : t -> string -> t
   val mkdir_p : ?perms:int -> t -> unit
   val initial_cwd : t
--- a/otherlibs/stdune/src/path.mli
+++ b/otherlibs/stdune/src/path.mli
@@ -71,6 +71,7 @@ end
 module External : sig
   include Path_intf.S
 
+  val root : t
   val initial_cwd : t
   val cwd : unit -> t
   val relative : t -> string -> t
--- a/src/dune_rules/dune
+++ b/src/dune_rules/dune
@@ -37,7 +37,8 @@
   dune_patch
   scheme
   fs
-  unix)
+  unix
+  xdg)
  (synopsis "Internal Dune library, do not use!")
  (instrumentation
   (backend bisect_ppx)))
--- a/src/dune_rules/ocaml_toolchain.mli
+++ b/src/dune_rules/ocaml_toolchain.mli
@@ -24,6 +24,13 @@ val of_env_with_findlib
   -> which:(Filename.t -> Path.t option Memo.t)
   -> t Memo.t
 
+val make
+  :  Context_name.t
+  -> which:(string -> Path.t option Memo.t)
+  -> env:Env.t
+  -> get_ocaml_tool:(dir:Path.t -> string -> Path.t option Memo.t)
+  -> t Memo.t
+
 val of_binaries : path:Path.t list -> Context_name.t -> Env.t -> Path.Set.t -> t Memo.t
 
 (** Return the compiler needed for this compilation mode *)
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -1168,7 +1168,9 @@ end = struct
   let resolve_impl ((db : DB.t), ctx, (name : Package.Name.t)) =
     match Package.Name.Map.find db.all name with
     | None -> Memo.return None
-    | Some { Lock_dir.Pkg.build_command; install_command; depends; info; exported_env } ->
+    | Some
+        ({ Lock_dir.Pkg.build_command; install_command; depends; info; exported_env } as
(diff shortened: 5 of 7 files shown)


gridbugs:
How did you test this? It doesn't seem to work for me. Building a dune project with a lockdir and toolchains enabled gives me this error:
```
/home/s/.cache/dune/toolchains/ocaml-base-compiler.5.2.0-863680cbf6cf46f0d359eef4882252c5/target/cookie
...
```

This wasn't an issue in https://github.com/ocaml/dune/pull/10719, but I'm guessing it will be fixed by implementing:

> There's no cookie file being written for toolchains. We should that to keep things as uniform as possible with regular packages.

Which digest do you mean in:

> The digest doesn't include dependencies of the resolved packages. Again, we should include a digest of the transitive closure of deps (but without the random id's).

Do you mean the hash at the end of toolchains directories (e.g. `ocaml-base-compiler.5.2.0-863680cbf6cf46f0d359eef4882252c5`) or something else? What random id's are you referring to?

gridbugs:
I spend some time today trying to understand how to create a cookie file for a toolchain package. For non-toolchain package it looks like the rule that produces cookies depends on the package having been downloaded, built, and installed, but for toolchain packages these rules aren't being created. I can see that something is setting up package rules by calling `Pkg_rules.setup_rules` for `_build/_private`, `_build/_private/default`, `_build/_private/default/.pkg`, `_build/_private/default/.pkg/<some package>`, etc, but I can't figure out what is driving that process.

@rgrinberg can you help me understand where the logic lives for choosing which directories to set up package rules inside of?

Also from the type of `setup_rules` and related functions it looks like they are only intended to create rules inside the project directory (ie. paths are represented with `Path.Build.t`). That suggests I won't be able to use rules to create a cookie for toolchain packages since they are external to any project. Would it be reasonable to instead manually create the cookie file for a toolchain package at the point where it's needed (e.g. https://github.com/ocaml/dune/blob/1f447cae96fc7a4b1fbabd957335a87a0b33e6e1/src/dune_rules/pkg_rules.ml#L417) rather than creating it with a rule?

rgrinberg:
> How did you test this? It doesn't seem to work for me. Building a dune project with a lockdir and toolchains enabled gives me this error:

Yes, I didn't get it working yet. As you've guessed, the cookie file support isn't there.

> Do you mean the hash at the end of toolchains directories (e.g. ocaml-base-compiler.5.2.0-863680cbf6cf46f0d359eef4882252c5) or something else? What random id's are you referring to?

Yes, that's the digest. [In your code](https://github.com/gridbugs/dune/blob/12e30f89fd67899cbff76e95c5c9c7bbfe1a4ed4/src/dune_rules/pkg_rules.ml#L336), you had:

```ocaml
let pkg_hash t = Digest.generic (t.build_command, t.install_command, t.depends, t.info)
```

Note that every `t` in `t.depends` has an `id : Id.t` field that's re-generated on every invocation. It's not meaningful and shouldn't be a part of the hash.

> I spend some time today trying to understand how to create a cookie file for a toolchain package. For non-toolchain package it looks like the rule that produces cookies depends on the package having been downloaded, built, and installed, but for toolchain packages these rules aren't being created

I don't think toolchains are being special cased in anyway for creating the rules. The problem is probably that executing their rules isn't producing anything dune understands (b/c there's no cookie). If you run `$ dune build _build/_private/default/.pkg/ocaml-base-compiler/target`, what do you see? Are any toolchain commands being executed? I believe this should do something at least.

For a toolchain package, the cookie should be created with roughly the following logic:

```ocaml
if not toolchain_is_installed then install_toolchain;
let files = read_files_from_installation () in
let cookie = make_cookie files in
write_cookie cookie
```

This should be your install action. Take a look at the existing install action for regular packages. It's important to understand how it works, because installing a toolchain is what should do most of the work.

> Also from the type of setup_rules and related functions it looks like they are only intended to create rules inside the project directory (ie. paths are represented with Path.Build.t). That suggests I won't be able to use rules to create a cookie for toolchain packages since they are external to any project. Would it be reasonable to instead manually create the cookie file for a toolchain package at the point where it's needed 

Creating the cookie file at the point where it's needed is exactly the same as having a rule for it. Note that the cookie file should exist on a per project basis in the _build directory (yes it will always be the same, but this is the easiest). It is just that the paths in this cookie file that need to point to your external toolchain installation.

I added some docs [here](https://github.com/ocaml/dune/pull/10781). Feel free to ask more questions.

I think it's important to understand the two different execution stages:
1. Rule generation
2. Action Execution

For toolchains, these two stages have these responsibilites:
1. During rule generation, we determine if the package is a toolchain or not. We determine the path where it must exist.
2. During action execution, we do the actual building and installing. However, the only build path we produce is the cookie file. Everything else is written to the toolchain dir.

PR #10770, opened 2024-07-28
refactor(engine): move some dap code into separate module

rgrinberg:
This code has been eliminated in the JS fork. Action execution is easier to sync if the removed bits live in a separate file.

cc @snowleopard 


snowleopard (approved):
Good idea, thanks!

PR #10777, opened 2024-07-29
Jsoo: make dune aware of sourcemap generation 

### LINKED ISSUE #10673: Dune does not recognise separate source map files from js_of_ocaml
## Expected Behavior
Hello, I am compiling a module using jsoo using separate sourcemap files. After passing --sourcemap in both flags and link_flags, js_of_ocaml will emit both .bc.js and .bc.map file into the build folder.

We expect that Dune is aware of the .js and .map files, and we should be able to reference these in other rules. For example, this dune file should install both the compiled Javascript and its source map into the bin directory.

```lisp
(executable
  (name builder)
  (modes js)
  (modules builder)

  (js_of_ocaml (flags --sourcemap :standard \ --source-map-inline)
               (link_flags --sourcemap :standard \ --source-map-inline)))

; static files for distribution
(install
 (files builder.bc.js builder.map.js)
 (section bin))
```

## Actual Behavior
Dune does not recognise the .map file ("no rule found") and fails the build.
```python
File "web/dune", line 28, characters 22-36:
28 |  (files builder.bc.js builder.bc.map index.html)
                           ^^^^^^^^^^^^^^
Error: No rule found for web/builder.bc.map
```

## Reproduction

- PR with a reproducing test: 

1. Attempt a build with a dune file similar to the above. For convenience, I have created a minimal dune project here: [jsoo_sourcemap_missing.tar.gz](https://github.com/user-attachments/files/16011498/jsoo_sourcemap_missing.tar.gz).
1. Observe the error when running `dune build`.
3. Note the source map is present in the build directory: `ls _build/default/web/`
3. Observe a successful build when omitting the .map.js from the install clause of web/dune.

## Specifications

- Version of `ocaml` (output of `ocamlc --version`): 4.14.2
- Operating system (distribution and version): EndeavourOS Rolling Linux x86_64

hhugo:
Attempt to fix https://github.com/ocaml/dune/issues/10673
One can now configure the jsoo behavior for sourcemap in the dune stanza directly.

It can be configure in the `env` stanza or per executable.

In addition, compilation_mode can now also be configured per executable.

Todo:
 - [x] update doc
 - [x] find a better name for "sourcemap in a separate file"


hhugo:
@katrinafyi, does that work for you ?

katrinafyi:
This looks excellent! Thanks for working on it, and I'll give it a try sometime this week.

christinerose (approved):

katrinafyi:
It works well after a basic test :) Thanks!

hhugo:
@vouillon, does this look ok in the context of wasm_of_ocaml ?

vouillon:
This looks ok to me. When generating some wasm code, one can just ignore this option.

christinerose (approved):

rgrinberg:
Thanks @hhugo. How about a CHANGES entry?

hhugo:
> Thanks @hhugo. How about a CHANGES entry?

added

hhugo:
This causes a small regression with dune 3.17.
If one is passing explicit sourcemap flags with a dune-lang < 3.17 , one may end up with duplicated (or incompatible)  flags that jsoo will reject.
I'll see if I can fix this.

PR #10780, opened 2024-07-29
Fix multi line formatting

maxRN:


src/dune_sexp/ast.mli
```diff
@@ -5,6 +5,7 @@ open Stdune
 type t =
   | Atom of Loc.t * Atom.t
   | Quoted_string of Loc.t * string
+  | Block_string of Loc.t * string
```
rgrinberg, line 8: Is it possible to introduce the Block_string only in the CST?

maxRN: Not without further work: `Dune_sexp.parse` parses all values into a list of `Encoded.t` which are actually `Ast.t` values and turns those back into `Cst.t`. So if we don't have a `Block_string` in the AST we lose precision so to speak, because we would turn a `Cst.Block_string` into a `Ast.Quoted_string` and then we can no longer know if it's really a `Quoted_string` or a `Block_string`.

Maybe we could somehow adjust the parsing function and add an option to directly return it as `Cst.t` values.

nojb: Why not change `string` by a richer abstract type that would remember whether it came from a quoted string or a blocked string, as well as having a `to_string` to get the underlying string. The type could also include the `Loc.t`.

PR #10782, opened 2024-07-30
Create cookie file for toolchain packages

gridbugs:
This fixes two issues with cookies for toolchain packages:

Prior to this change, dune expected toolchain cookies to be created inside the toolchains directory, ie. outside of the project. The machinery for creating cookies only works within the build directory, so cookies for toolchain packages were never created. The location of the cookie had to be the same as the location of the installed package, in what dune referred to as the package's "target_dir". The fix is to decouply these two locations by adding a new path "prefix" to packages, corresponding to the "--prefix" argument of most configure scripts. This allows toolchain packages to have a "target_dir" within the build directory, and for a cookie file to be created within its "target_dir", while still allowing the package to be installed outside the project inside its "prefix", inside the toolchains directory.

The second issue is that the contents of the cookie was based on the package's "target_dir", and thus cookies for toolchain packages would be empty (toolchain packages are now installed in the package's "prefix" dir instead). Dune makes the distinction between the "paths" and "write_paths" of a package, the latter being the unsandboxed final install paths of the package, and it's the "target_dir" within a package's "write_paths" that is traversed (in the absence of a <package>.install file) to determine the contents of the package's cookie. The fix was to detect the case where the "paths" of a package contains a prefix referring outside the build directory, and in such a case, traverse the directory at that path rather than the one at "target_dir".

Toolchains still don't completely work, as in some cases build tools are unable to find binaries installed by the compiler package, however manual inspection shows that the cookie is being created for toolchain packages and contains the expected contents. Fixes for the remaining issues will come in later changes.

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -68,6 +68,7 @@ module Paths = struct
     ; name : Package.Name.t
     ; install_roots : 'a Install.Roots.t Lazy.t
     ; install_paths : 'a Install.Paths.t Lazy.t
+    ; prefix : 'a
     }
 
   let map_path t ~f =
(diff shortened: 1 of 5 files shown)


rgrinberg (approved):

src/dune_rules/pkg_rules.ml
```diff
@@ -1198,16 +1205,37 @@ end = struct
           let doc =
             Path.outside_build_dir @@ Path.Outside_build_dir.relative prefix "doc"
           in
+          let* build_command =
+            match build_command with
+            | None | Some Dune -> Memo.return build_command
+            | Some (Action action) ->
+              let+ installed = Fs_memo.dir_exists prefix in
```
rgrinberg, line 1212: Note that doing this check at rule generation time at is wrong as I've mentioned. Doesn't prevent this PR from being merged, but you should work on pushing these checks to action execution.

src/dune_rules/pkg_rules.ml
```diff
@@ -1198,16 +1205,37 @@ end = struct
...
+              let action =
+                if installed
+                then
+                  (* If the toolchain is already installed, just create an empty
+                     config.cache file so other packages see that the compiler package
+                     is installed. *)
+                  Dune_lang.Action.Run
```
rgrinberg, line 1219: Any reason why you moved this from `Pkg_toolchain`? I'd really prefer to keep this feature as isolated as possible. I realize, we can't fully accomplish that, but we should at least try.

gridbugs: Not a strong reason, no. I'll move it back.

Issue #10729, opened 2024-07-15
[Pkg] Enable cache by default


maiste:
For the Dune Developer Preview program, we would like to have the cache enabled by default to save bandwidth when you are trying to install packages. This modification would respect the following:

1. If there is a value in the configuration, the CLI or in the environment, we want to make sure this one is used.
2. Otherwise, the cache would be enabled by default at configure time only for the developer preview executable.

A solution to this would be to parametrize the default solution using some flags and enabled them (it?) at configure time.

The first proposition can be found on #10710

maiste:
@rgrinberg I opened this PR on @emillon suggestion so we can discuss a possible solution for this.

>  The problem lies in that most users do not have such a configuration. Moreover, most project contain user rules of the form:
> (rule
 > ..)
> That are absolutely not safe to cache. Therefore, flipping the default like is done in this PR is highly likely to cause a lot of unintended breakage.

IIUC the problem, for many users, their rules depend on the fact that they are not caching anything? As the goal of the PRs it to avoid the package to be download multiple times, could a specific variable for this cache be a solution?

rgrinberg:
The end goal is correct, but we need to adjust the plan on how to get there. So what we'd like is the ability to cache only package rules. To get there, we can:

1. Introduce a shared cache toggle in build rules themselves (we have one internally called `can_go_in_shared_cache`)
2. Set this toggle to `false` for all rules by default
3. Set this toggle to true for package rules
4. Enable the shared cache
5. Introduce a way to change the default for normal rules to get back the old behavior for users that want to cache their own rules

maiste:
Ok, IIUC, the idea is to split the shared cache toggle, in multiple toggles, so we can have specific caches per rule?

rgrinberg:
We'd like the following behavior:

1. DUNE_CACHE=enabled everything is cached like the current behavior
2. DUNE_CACHE=disabled nothing is cached
3. DUNE_CACHE is not set. Only the package rules are cached.

A per rule toggle would be a reasonable way to implement this

maiste:
I see. If I understand correctly how it would work, it would mean rewrite the shared cache from one unique cache to a split version of it?

emillon:
> I see. If I understand correctly how it would work, it would mean rewrite the shared cache from one unique cache to a split version of it?

I don't think so. You keep a single shared cache, but you change the logic to determine if a rule can use it.

Leonidas-from-XIV:
Solved by #10710 and #10850.

PR #10789, opened 2024-07-31
Move 'can_go_in_shared_cache' from Action to Rule, with the same default

ElectreAAS:
This is my second attempt at going in the direction of caching by default, and caching downloads, following the discussion over at #10729.
I realized that the field `can_go_in_shared_cache` in `Dune_Engine.Action.Full.t` already existed, had a default to true, and was never set to false anywhere. I then hijack this by moving the field to `Rule.t` instead. This makes it so that all calls to `Rule.make` can now set this variable, and the goal is to change the call in `pkg_rule` to true when the PR is ready.

cc @Leonidas-from-XIV

### DIFF
--- a/src/dune_engine/build_system.ml
+++ b/src/dune_engine/build_system.ml
@@ -257,7 +257,6 @@ end = struct
     let { Action.Full.action
         ; env
         ; locks
-        ; can_go_in_shared_cache
         ; sandbox = _ (* already taken into account in [sandbox_mode] *)
         }
       =
@@ -273,7 +272,7 @@ end = struct
       , Dep.Facts.digest facts ~env
       , target_paths rule.targets.files @ target_paths rule.targets.dirs
       , Action.for_shell action
-      , can_go_in_shared_cache
+      , true (* TODO: don't forget to remove this *)
       , List.map locks ~f:Path.to_string
       , Execution_parameters.action_stdout_on_success execution_parameters
       , Execution_parameters.action_stderr_on_success execution_parameters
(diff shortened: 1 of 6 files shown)


src/dune_engine/build_system.ml
```diff
@@ -273,7 +272,7 @@ end = struct
       , Dep.Facts.digest facts ~env
       , target_paths rule.targets.files @ target_paths rule.targets.dirs
       , Action.for_shell action
-      , can_go_in_shared_cache
+      , true (* TODO: don't forget to remove this *)
```
ElectreAAS, line 275: TODO comment intentionnaly left here, as removing the `true` in the tuple changes the digest, and trips the tests. I wanted to make a separate commit that only removes this line and changes the hash to make it clean for the CI.

rgrinberg:
How does it help you to move the flag from the rule to the action? Every rule has a corresponding action, so I think you can just set the flag on the action in the rule.

ElectreAAS:
> How does it help you to move the flag from the rule to the action? Every rule has a corresponding action, so I think you can just set the flag on the action in the rule.

As you can probably guess I'm a little bit confused by the situation.
The field `can_go_in_shared_cache` in `Action.t` sounds like it should do the trick, and I can just set this flag to true in more places than currently. However, it is set by default to true, and I can't find a place in the code where that default value is not used. Is this field really doing the thing it should?
Moreover, if I want to enable caching package downloads, can I trust that by adding a variable in `dune_rules/pkg_rules` it's going to make its way to `dune_engine/build_system`? That file is the one place where the field is actually used, but I don't know if package rules actually make it there...

rgrinberg:
You're on the right track. To recap:

1. We want to enable the dune cache by default
2. But doing so will break a lot of existing code that might not be cacheable
3. So we flip the default, but set this flag for all rules to say that they're incompatible with the dune cache.
4. We make an exception for the rules defined by pkg_rules.ml. Those should now be explicitly compatible.
5. But flipping 3, we've made it impossible for users to get back the old behavior of caching everything. So we can introduce a dune flag, say `DUNE_CACHE_RULE_DEFAULT=false|true` to allow user to get back to the old world where everything would be cached.

Hope that makes sense. Feel free to ask more.

ElectreAAS:
Thanks for the explanation. I think I made some progress back on #10710.

Issue #10788, opened 2024-07-31
Error building the package "seq" as a dependency with dune package management


gridbugs:
To reproduce this, clone https://github.com/ocaml-ppx/ocamlformat.git (which depends on seq - other packages depending on seq may also have this problem).

The run:
```
$ dune pkg lock
$ dune build
```

The error will be:
```
Error:                              
open(_build/_private/default/.pkg/seq/source/META.seq): No such file or directory
-> required by _build/_private/default/.pkg/seq/target/cookie
-> required by - package re
-> required by - package alcotest
```

Inspecting the contents of _build/_private/default/.pkg/seq/source/META.seq:
```
$ ls _build/_private/default/.pkg/seq/
extra_source
```

The lockfile for seq hints at the issue:
```
 $ cat dune.lock/seq.pkg 
(version base)

(depends ocaml)

(extra_sources
 (META.seq
  (fetch
   (url
    https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/seq/META.seq)
   (checksum
    sha256=e95062b4d0519ef8335c02f7d0f1952d11b814c7ab7e6d566a206116162fa2be)))
 (seq.install
  (fetch
   (url
    https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/seq/seq.install)
   (checksum
    sha256=fff926c2c4d5a82b6c94c60c4c35eb06e3d39975893ebe6b1f0e6557cbe34904))))
```

This corresponds with the opam file:
```
$ opam show seq.base --raw
opam-version: "2.0"
name: "seq"
version: "base"
synopsis:
  "Compatibility package for OCaml's standard iterator type starting from 4.07."
maintainer: " "
authors: " "
homepage: " "
bug-reports: "https://caml.inria.fr/mantis/main_page.php"
depends: [
  "ocaml" {>= "4.07.0"}
]
dev-repo: "git+https://github.com/ocaml/ocaml.git"
extra-source "META.seq" {
  src:
    "https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/seq/META.seq"
  checksum: [
    "sha256=e95062b4d0519ef8335c02f7d0f1952d11b814c7ab7e6d566a206116162fa2be"
    "md5=b33c8a1a6c7ed797816ce27df4855107"
  ]
}
extra-source "seq.install" {
  src:
    "https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/seq/seq.install"
  checksum: [
[...]

PR #10790, opened 2024-08-01
Add repro for gh10788

gridbugs:
Reproduces https://github.com/ocaml/dune/issues/10788, as well as a related issue where dune would attempt to copy extra_source files into a location where they already exist, leading to a permission error.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/exclusively-extra-sources.t
+++ b/test/blackbox-tests/test-cases/pkg/exclusively-extra-sources.t
@@ -1,0 +1,24 @@
+Test for packages with no source field but with extra_sources.
+
+  $ . ./helpers.sh
+  $ make_lockdir
+
+  $ cat > dune.lock/foo.pkg <<EOF
+  > (version 1)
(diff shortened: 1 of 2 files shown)


gridbugs:
Note the failure on macos, where the output of one test is different from on linux, and another test doesn't fail at all while it does fail on linux. Since this issue is fixed in https://github.com/ocaml/dune/pull/10791, (at which point both tests will pass and print no error messages) I recommend we merge that PR first, then rebase this on top (updating its documentation to be consistent with the test now passing).

rgrinberg (approved):

gridbugs:
I've updated the tests to split the two tests into separate files, and I changed the second test scenario to be when a package's source has a filename in common with an extra-source. I verified that this produced the same error prior to https://github.com/ocaml/dune/pull/10791 being merged as the original test, but it's a way of exercising the code that closer matches what we'll see in the wild than what this test originally did (previously the test would manually create a file in the source directory with the same name as the extra source prior to building the project).

PR #10791, opened 2024-08-01
Fix some issues with extra-sources

### LINKED ISSUE #10788: Error building the package "seq" as a dependency with dune package management

gridbugs:
Before copying a file from a package's "extra_sources" directory to the "source" directory, make sure that:
 - the package's "source" directory exists
 - the destination file doesn't already exist

Fixes https://github.com/ocaml/dune/issues/10788

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -1757,9 +1757,27 @@ let build_rule context_name ~source_deps (pkg : Pkg.t) =
         in
         copy_action
         @ List.map pkg.info.extra_sources ~f:(fun (local, _) ->
+          (* If the package has extra sources, they will be
+             initially stored in the extra_sources directory for that
+             package. Prior to building, the contents of
+             extra_sources must be copied into the package's source
+             directory. *)
           let src = Paths.extra_source pkg.paths local in
           let dst = Path.Build.append_local pkg.write_paths.source_dir local in
-          Action.copy src dst |> Action.Full.make |> Action_builder.With_targets.return)
+          Action.progn
+            [ (* If the package has no source directory (some
+                 low-level packages are exclusively made up of extra
+                 sources), the source directory is first created. *)
+              Action.mkdir pkg.write_paths.source_dir
+            ; (* It's possible for some extra sources to already be at
+                 the destination. If these files are write-protected
+                 then the copy action will fail if we don't first remove
+                 them. *)
+              Action.remove_tree dst
+            ; Action.copy src dst
+            ]
+          |> Action.Full.make
+          |> Action_builder.With_targets.return)
       and+ build_action =
         match Action_expander.build_command context_name pkg with
         | None -> Memo.return []


Leonidas-from-XIV:
I don't think this is the right solution. The issue is that if there is no `build` field, it doesn't add the `Action.chdir` thus no dependency on the `source_dir` is recorded, thus the copy rules are not executed by the engine. If they are executed, they work fine, the directory is created.

You can see that `seq` works if you add a

```
(build (run true))
```
into the lock file.

gridbugs:
Interesting! I'll dig deeper into this tomorrow.

Leonidas-from-XIV:
I've been discussing this with help from @emillon but just recording a dependency is not enough, as then it fails because there's no target that generates `source_dir`. I'll also be looking at this because this is what at every single package that (transitively) depends on `seq` fails on when building.

emillon:
do we have a test for that?
also, if setting `(build (run true))` works we can use that in the meantime.

Leonidas-from-XIV (approved):
After spending entirely too much of both my and @emillon's time trying to understand what is happening and why `(build (run true))` or even just `(build (progn))` is fixing it:

If there is a build action, the expander will add a `Action.chdir <source-dir>` to the build. This will implicitly `mkdir` `<source-dir>` if if doesn't exist thus when the copy rules are executed they succeed because `<source-dir>` exists at that point. So the `(build (progn))` is actually a `(build (chdir <source-dir> (progn))`.

If there are no build instructions in the lock file, this `chdir` doesn't happen and thus no `<source-dir>` folder will be created, at which case the copy rules will fail (I guess silently) and the next thing expecting the copy rules to have succeeded will fail because the files didn't appear where they were supposed to be.

Thus creating the `<source-dir>` explicitly is the right solution here, however maybe it would make sense to do that conditionally if it doesn't exist, like above with `Fs_memo.dir_exists`. Then again, if the `mkdir` action is idempotent, it might be faster to attempt to create it than issuing a `stat()`.

Anyway I am now confident that this is the right solution and explains the rather unexpected behavior that I've observed.

src/dune_rules/pkg_rules.ml
```diff
@@ -1757,9 +1757,27 @@ let build_rule context_name ~source_deps (pkg : Pkg.t) =
...
+          Action.progn
+            [ (* If the package has no source directory (some
+                 low-level packages are exclusively made up of extra
+                 sources), the source directory is first created. *)
+              Action.mkdir pkg.write_paths.source_dir
+            ; (* It's possible for some extra sources to already be at
+                 the destination. If these files are write-protected
+                 then the copy action will fail if we don't first remove
+                 them. *)
+              Action.remove_tree dst
```
rgrinberg, line 1776: Why are removing existing sources though? Shouldn't we merge all the sources?

Leonidas-from-XIV:
Given `Copy` just maps to `cp` https://github.com/ocaml/dune/blob/0a8af6e1c8eb701e43cefa92bcfd131ec89ed6d7/src/dune_engine/action_to_sh.ml#L62

I guess you're right, it's `cp` being cryptic what exactly is missing:

```sh
$ cp exists.ml missing/new.ml
cp: cannot create regular file 'missing/new.ml': No such file or directory
```

src/dune_rules/pkg_rules.ml
emillon: that confused me as well, apparently this is to encode the semantics that extra-sources override the contents of the tarball.
I think a dedicated test for this would be useful.
In addition, maybe we can skip unpacking files present in `extra_sources` and skip this `remove_tree`, or something like that?

emillon:
Copy is not executed by shelling out to `cp`, it's just how we represent that in `dune rules -m`. It's implemented by `Io.copy_file` (see the `Action_exec` module).

src/dune_rules/pkg_rules.ml
Leonidas-from-XIV: If the source tarball has a file (or directory) and the extra-source has the same path I would expect that the extra-source would replace the original file as it is meant to patch the source tarball.

rgrinberg: Ok let's just make sure we have a test for this and we are matching opam's behavior.

gridbugs: Tests are added in https://github.com/ocaml/dune/pull/10790. I updated that PR with a test for the scenario where an extra-source file replaces a file in the source dir and did a manual experiment to confirm that this is also how opam behaves.

PR #10794, opened 2024-08-03
refactor: remove action runners from public dune

rgrinberg:
Action runners in this dune are a rather old version that it is very much out of date with the state of the actual feature and has many bugs and limitations that need a lot of upstreaming to fix. In its current state, it does not provide any value to users of public dune.

If we ever want to upstream this feature, it will be easier to just do it from scratch anyway instead of resolving the conflicts against some old version. This is due to the fact that the action runner API with the engine is rather small and well defined.

Once we introduce a better abstraction for action execution (as we are indeed planning), it will be probably be even simpler to upstream action runners.

All in all, we gain nothing on either team by keeping this dead weight around.

cc @snowleopard 


src/dune_engine/build_system.ml
```diff
@@ -1106,12 +1103,7 @@ let report_early_exn exn =
     let+ () = State.add_errors errors
     and+ () =
       match !Clflags.stop_on_first_error with
-      | true ->
-        let* () =
-          (Build_config.get ()).action_runners ()
-          |> Fiber.parallel_iter ~f:Action_runner.cancel_build
-        in
-        Scheduler.cancel_current_build ()
```
rgrinberg, line 1114: This bit is completely broken anyway

snowleopard (commented):
Thanks!

PR #10798, opened 2024-08-05
Add test of the toolchains mechanism

gridbugs:


gridbugs:
The test is failing because the randomly-generated test sandbox directory ends up in the lockfile, causing the hash in the toolchain directory name to be different every time the test runs. Will fix.

gridbugs:
> The test is failing because the randomly-generated test sandbox directory ends up in the lockfile, causing the hash in the toolchain directory name to be different every time the test runs. Will fix.

This is now fixed.

gridbugs:
I'll do that in a separate PR as it's unrelated to this change.

test/blackbox-tests/test-cases/pkg/toolchain-installation.t
```diff
@@ -0,0 +1,95 @@
...
+  $ . ./helpers.sh
+  $ make_lockdir
+
+  $ mkdir fake-compiler
+
+A configure script for the fake compiler that writes the prefix path
+to a file so the fake compiler can be installed to the specified
+location, testing that dune respects the "prefix" variable for
+toolchain packages.
+  $ cat > fake-compiler/configure << EOF
```
Leonidas-from-XIV, line 13: Nitpick but if you [quote EOF](https://linuxize.com/post/bash-heredoc/) you don't have to escape `$` which will make it a bit easier to read and extend since one doesn't have to remember to add the escapes.

test/blackbox-tests/test-cases/pkg/toolchain-installation.t
```diff
@@ -0,0 +1,95 @@
...
+  $ mkdir -p fake-compiler/target/bin
+  $ cat > fake-compiler/target/bin/ocamlc << EOF
+  > #!/bin/sh
+  > echo "Hello from fake ocamlc!"
+  > EOF
+  $ chmod a+x fake-compiler/target/bin/ocamlc
+
+The makefile for the fake compiler copies the fake ocamlc script into
+the appropriate location, respecting the DESTDIR variable.
+  $ cat > fake-compiler/Makefile << EOF
```
Leonidas-from-XIV, line 31: Similarly here. Makefiles are notoriously bad for quoting and unquoting.

Leonidas-from-XIV (approved):
I added a few stylistic comments but when it comes to the content of the test it looks good and sensible to me!

PR #10802, opened 2024-08-06
Progress indication for lockdir generation

gridbugs:
First pass at progress indication for lockdir generation, where the status line displays a message while the package repos are being updated, and a second message during solving. This might be updated in the future to display finer-grained status updates.

### DIFF
--- a/bin/pkg/lock.ml
+++ b/bin/pkg/lock.ml
@@ -5,6 +5,53 @@ module Opam_repo = Dune_pkg.Opam_repo
 module Lock_dir = Dune_pkg.Lock_dir
 module Pin_stanza = Dune_pkg.Pin_stanza
 
+module Progress_indicator = struct
+  module Per_lockdir = struct
+    module State = struct
+      module Repository = Dune_pkg.Pkg_workspace.Repository
+
+      type t =
+        | Updating_repos of Repository.Name.t list
+        | Solving
+
+      let pp = function
+        | Updating_repos repo_names ->
+          Pp.textf
+            "Updating package repos %s..."
+            (List.map repo_names ~f:(fun repo_name ->
+               Repository.Name.to_string repo_name |> String.quoted)
+             |> String.enumerate_and)
+        | Solving -> Pp.text "Solving..."
+      ;;
+    end
+
+    type t =
+      { lockdir_path : Path.Source.t
+      ; state : State.t option ref
+      }
+
+    let create lockdir_path = { lockdir_path; state = ref None }
+  end
+
+  (* The progress indicator for the entire lock operation, which may
+     involve generating multiple lockdirs *)
+  type t = Per_lockdir.t list
+
+  let pp (t : t) =
+    (* Only display the first non-done lockdir state, since the status
+       line can only consist of a single line. *)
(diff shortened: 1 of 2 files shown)


bin/pkg/lock.ml
```diff
@@ -46,30 +107,25 @@ let solve_lock_dir
...
         (unset_solver_vars_of_workspace workspace ~lock_dir_path)
   in
-  let overlay =
-    Console.Status_line.add_overlay (Constant (Pp.text "Solving for Build Plan"))
-  in
   let* repos =
-    repositories_of_workspace workspace
-    |> get_repos ~repositories:(repositories_of_lock_dir workspace ~lock_dir_path)
+    let repo_map = repositories_of_workspace workspace in
+    let repo_names = Dune_pkg.Pkg_workspace.Repository.Name.Map.keys repo_map in
```
Leonidas-from-XIV, line 112: I think we should make sure to sort them to make sure that the order is always consistent.

bin/pkg/lock.ml
```diff
@@ -5,6 +5,66 @@ module Opam_repo = Dune_pkg.Opam_repo
 module Lock_dir = Dune_pkg.Lock_dir
 module Pin_stanza = Dune_pkg.Pin_stanza
 
+module Progress_indicator = struct
+  module Per_lockdir = struct
+    module State = struct
+      module Repository = Dune_pkg.Pkg_workspace.Repository
+
+      type t =
```
Leonidas-from-XIV, line 13: To me it would make more sense if these were more aligned with the solving process. `Initializing` is not a step of the solver, it's just a dummy value to put into the state `ref` (it is never assigned outside of initialization).

Will the user ever see `Done`?

bin/pkg/lock.ml
```diff
@@ -5,6 +5,66 @@ module Opam_repo = Dune_pkg.Opam_repo
...
+    type t =
+      { lockdir_path : Path.Source.t
+      ; state : State.t ref
+      }
+
+    let create lockdir_path = { lockdir_path; state = ref State.Initializing }
+
+    let pp { lockdir_path; state } =
+      Pp.concat
+        [ Pp.textf "Locking %s: " (Path.Source.to_string lockdir_path |> String.quoted)
```
Leonidas-from-XIV, line 41: The progress displays `"dune.lock"` with double quotes, when it shows the solution it says `Solution for dune.lock`, so, without the quotes. Either way is fine but I think we should be consistent. Maybe have a `pp` for lockdir names.

Leonidas-from-XIV (commented):
I like the overlay a lot and in my experience it works really well and looks nice. My points here are fairly minor.

The mutable state is a bit iffy, but other solutions seem a bit heavy-handed for the usecase so I don't have a better suggestion.

bin/pkg/lock.ml
gridbugs: Users will probably never see `Initializing` or `Done`. I'll update `State.t` to just represent the updating and solving steps, and make the state optional so a `None` value can represent the state before and after locking.

bin/pkg/lock.ml
gridbugs: I removed the quotes for consistency.

Leonidas-from-XIV (approved):

PR #10803, opened 2024-08-06
Print package build progress status messages

gridbugs:
First pass at printing messages to the console indicating when a package is being downloaded and built. This is disabled by default and enabled with a config variable.

### DIFF
--- a/src/dune_rules/pkg_build_progress.ml
+++ b/src/dune_rules/pkg_build_progress.ml
@@ -1,0 +1,82 @@
+open! Import
+
+let enabled = Config.make_toggle ~name:"pkg_build_progress" ~default:`Disabled
+
+module Status = struct
+  type t =
+    [ `Downloading
+    | `Building
+    ]
+
+  let to_string = function
+    | `Downloading -> "Downloading"
+    | `Building -> "Building"
+  ;;
+end
+
+module Message = struct
+  type t =
+    { package_name : Package.Name.t
+    ; package_version : Package_version.t
+    ; status : Status.t
+    }
+
+  let user_message { package_name; package_version; status } =
+    let status_tag = User_message.Style.Ok in
+    User_message.make
+      [ Pp.concat
+          [ Pp.tag status_tag (Pp.text (Status.to_string status))
+          ; Pp.textf
+              " %s.%s"
+              (Package.Name.to_string package_name)
+              (Package_version.to_string package_version)
+          ]
+      ]
+  ;;
+
+  let display t =
+    match Config.get enabled with
+    | `Enabled -> Console.print_user_message (user_message t)
+    | `Disabled -> ()
+  ;;
+
+  let encode { package_name; package_version; status } =
+    Sexp.List
+      [ Sexp.Atom (Package.Name.to_string package_name)
+      ; Sexp.Atom (Package_version.to_string package_version)
+      ; Sexp.Atom (Status.to_string status)
+      ]
+  ;;
+end
+
+module Spec = struct
+  type ('path, 'target) t = Message.t
+
+  let name = "progress-action"
+  let version = 1
+  let is_useful_to ~memoize:_ = true
+  let bimap t _f _g = t
+
+  let encode t _ _ =
+    Sexp.List [ Sexp.Atom name; Sexp.Atom (Int.to_string version); Message.encode t ]
+  ;;
+
+  let action t ~ectx:_ ~eenv:_ =
+    let open Fiber.O in
+    let+ () = Fiber.return () in
+    Message.display t
+  ;;
+end
+
+let progress_action package_name package_version status =
+  let module M = struct
+    type path = Path.t
+    type target = Path.Build.t
+
+    module Spec = Spec
+
+    let v = { Message.package_name; package_version; status }
+  end
+  in
+  Action.Extension (module M)
+;;
--- a/src/dune_rules/pkg_build_progress.mli
+++ b/src/dune_rules/pkg_build_progress.mli
@@ -1,0 +1,14 @@
+open! Import
+
+(** An action which prints a progress message about a package to
+    the console so users can be informed about which of their
+    project's dependencies are currently being installed.
+
+    The message will only print if the
+    DUNE_CONFIG__PKG_BUILD_PROGRESS config variable is "enabled"
+    (it's "disabled" by default). *)
+val progress_action
+  :  Package.Name.t
+  -> Package_version.t
+  -> [ `Downloading | `Building ]
+  -> Action.t
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -1675,6 +1675,16 @@ let source_rules (pkg : Pkg.t) =
        | `Local (`File, _) | `Fetch ->
          let fetch =
            Fetch_rules.fetch ~target:pkg.write_paths.source_dir `Directory source
+           |> With_targets.map
+                ~f:
+                  (Action.Full.map ~f:(fun action ->
+                     let progress =
+                       Pkg_build_progress.progress_action
+                         pkg.info.name
+                         pkg.info.version
+                         `Downloading
+                     in
+                     Action.progn [ progress; action ]))
          in
          Memo.return (Dep.Set.of_files [ pkg.paths.source_dir ], [ loc, fetch ])
        | `Local (`Directory, source_root) ->
@@ -1713,7 +1723,7 @@ let source_rules (pkg : Pkg.t) =
 
 let build_rule context_name ~source_deps (pkg : Pkg.t) =
   let+ build_action =
-    let+ build_and_install =
+    let+ copy_action, build_action, install_action =
       let+ copy_action =
         let+ copy_action =
           Fs_memo.dir_exists
@@ -1784,7 +1794,7 @@ let build_rule context_name ~source_deps (pkg : Pkg.t) =
           in
           [ mkdir_install_dirs; install_action ]
       in
-      List.concat [ copy_action; build_action; install_action ]
+      copy_action, build_action, install_action
     in
     let install_file_action =
       let prefix_outside_build_dir = Path.as_outside_build_dir pkg.paths.prefix in
@@ -1798,7 +1808,20 @@ let build_rule context_name ~source_deps (pkg : Pkg.t) =
       |> Action_builder.return
       |> Action_builder.with_no_targets
     in
-    Action_builder.progn (build_and_install @ [ install_file_action ])
+    (* Action to print a "Building" message for the package if its
+       target directory is not yet created. *)
+    let progress_building =
+      Pkg_build_progress.progress_action pkg.info.name pkg.info.version `Building
+      |> Action.Full.make
+      |> Action_builder.return
+      |> Action_builder.with_no_targets
+    in
+    Action_builder.progn
+      (copy_action
+       @ [ progress_building ]
+       @ build_action
+       @ install_action
+       @ [ install_file_action ])
   in
   let deps = Dep.Set.union source_deps (Pkg.package_deps pkg) in
   let open Action_builder.With_targets.O in


Leonidas-from-XIV:
I've tried it and one of the concerns I have with it is that it will print all of the "Downloading" and "Building" messages even as it is not actually doing these things as the actions resolve to no-ops. E.g. running `dune build` a second time should not print anything, yet it prints all the packages that it has already downloaded and built.

From a personal perspective I wish it was easier to add the current action to the `Done: 66% (66/100, 34 left)` indicator. That would avoid spamming the console with output and still provide a way to see that something is happening. Maybe creating an API to extend that would be a good idea?

gridbugs:
Good point about messages showing up even when the package doesn't need to be re-downloaded or rebuilt. I'll see what it will take to only print those messages when the action actually happens.

Regarding updating the `Done: X%...` line, my preference would be to leave that showing the overall progress while printing more detailed information to regular stdout.

gridbugs:
I made two improvements to this:
- it now only prints a status messages when the action happens. I couldn't figure out how to directly check when an action happens in a non-invasive way so I'm using the existence of the source and target directories as a proxy for the download and build actions having not been run.
- initially it would print a package's "Building" status before its "Downloading" status. This is because the download action starts after the build action, as the build action is what triggers the download action to run. Using the existence of a package's source and target directories as proxies for an action having completed gives more precise control over when a message is printed, which allows me to print the "Building" status after the download has completed.

src/dune_rules/pkg_build_progress.ml
```diff
@@ -0,0 +1,55 @@
...
+      ]
+  in
+  Console.print_user_message user_message
+;;
+
+let display_build_progress package_name package_version ~source_dir ~target_dir =
+  let open Action_builder.O in
+  let source =
+    match source_dir with
+    | `No_package_source -> Action_builder.return ()
```
Leonidas-from-XIV, line 30: This might need to be extended because while we don't download source tarballs in some packages (like the classic `seq.base`) we still download extra-sources so I would expect them to show up in as "Downloading".

Leonidas-from-XIV (commented):
Yes, that's a good point, it would be nice if there was a way to ask Dune whether an action is done or not. Maybe there's an API for it (then this could potentially be extended to any action), but anyway I think monitoring the source and target dirs is an ok workaround when it comes to packages.

src/dune_rules/pkg_build_progress.mli
```diff
@@ -0,0 +1,14 @@
...
+    to the console. This must be enabled by the setting the
+    DUNE_CONFIG__PKG_BUILD_PROGRESS config variable to "enabled". If
+    this variable is "disabled" (the default) then this function
+    returns a no-op instead. *)
+val display_build_progress
+  :  Package.Name.t
+  -> Package_version.t
+  -> source_dir:[ `Some of Path.Build.t | `No_package_source ]
+  -> target_dir:[ `Some of Path.Build.t | `No_build_or_install_command ]
+  -> unit Action_builder.t
```
rgrinberg, line 14: I don't think this is right. Action_builder is for generating actions (like the name says), while progress reporting should occur when the actions are being executed.

gridbugs: Fair enough, the current implementation does feel a bit hacky. Does dune have an API for registering a function to call when an action executes or is that something I'll need to add?

rgrinberg: No such API exists. I suppose you can just wrap existing actions with something like:
```ocaml
progn [ start_report_progress ; action ; stop_report_progress ]
```

A good first step would be to move the toolchain stuff to execution.

gridbugs: Wrapping actions makes sense. I think for a first version we'd only need to print a message before starting a build action.

> A good first step would be to move the toolchain stuff to execution.

How is this related to printing progress status messages. We'll be printing status messages for all packages, not just toolchain packages.

rgrinberg: >  Or is there some way to make an Action.t that just calls an ocaml function when it runs?

There is. You can create a custom action that will execute whatever OCaml code you'd like. See the various `exec` functions of the custom actions that we create.

rgrinberg: > How is this related to printing progress status messages. We'll be printing status messages for all packages, not just toolchain packages.

If you were to add progress reporting during action execution, it would only for normal packages and not not for toolchains because toolchains are built and installed during rule generation.

However, once toolchains are installed during action execution, then progress reporting would work for both types of packages.

You can do both of these tasks in any order. I just suggested doing the simpler one first.

gridbugs: Toolchains aren't built or installed during rule generation. The only unusual thing that happens with toolchains at rule generation time is the check whether the necessary toolchain package is already installed.

rgrinberg: Not sure how I confused those two things.

 I just realized that using `progn` for progress start/progress end will not work because it will not run termination if the action fails. I think you're better off introducing a dedicated action like:

```ocaml
val with_progress_report : Action.t -> Action.t
```

That will run the underlying action with the `Fiber.finalize` to make sure progress termination is always reported. You might need to add the action into `Dune_engine.Action` directly instead of using a custom action. Custom actions do not allow referring to other actions at the moment.

gridbugs: What do you mean by "it will not run termination if the action fails"?

I don't think we need to report the completion of an action right now though. All I want for the first version of this is for dune to print a message "Downloading foo..." or "Building foo..." immediately before the download or build of a package begins.

rgrinberg: What can kind of progress reporting do you expect when you do run the following command:

```
$ dune build _build/_private/default/.pkg/foo/target
```

I would expect a message telling me that the build of package `foo` has started and then this message to be eliminated once the build is done. This is consistent with how we manage progress updates for regular build notifications and it would require updating the status line twice.

If you only care to print a log message, you don't need a termination action. However, that's a poorer UI and is inconsistent with the rest of dune.

rgrinberg: Just to be clear, if you want to write a console message as a stepping stone, that's alright. But I do think that you will need to do what I described about eventually. Otherwise, you will not have a consistent experience and it will not work well in watch mode.

gridbugs: The specific problem I'm trying to solve right now is that when a user runs `dune build` in a project with a lockdir they get no output for the usually several minutes it takes for all the dependencies to be downloaded and built.

I'll have it write a console message at first since that's easy to implement and we can quickly get it in front of people to try it out.

Long term I agree that it would be nice to have a more sophisticated UI that displays the transition of each package from "downloading" to "building" to "done". This will be harder to do so I'll defer this work for a future change. Dune has the ability to render a dynamically-updating terminal UI with `Console.Status_line` but it only works for a single line so I don't see how we'd be able to show the status of all concurrently-building packages. Does dune have another UI mechanism that would be more appropriate or is that something we'd have to build?

rgrinberg: You can start with messages indeed.

We could extend the status line to work with more than one line. Although I think you should be able to make it work by putting all the packages being downloaded or installed in one line.

src/dune_rules/pkg_build_progress.ml
```diff
@@ -0,0 +1,55 @@
...
+         inside the _build directory. *)
+      if Path.Untracked.exists (Path.build source_dir)
+      then Action_builder.return ()
+      else (
+        display package_name package_version `Downloading;
+        Action_builder.path (Path.build source_dir))
+  in
+  let target =
+    match target_dir with
+    | `No_build_or_install_command -> Action_builder.return ()
```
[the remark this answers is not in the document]
gridbugs: I updated it so that if a package has no source archive we still print a "Downloading" message prior to running its copy actions. This means that packages with only `extra_sources` or no sources at all still get a "Downloading" message so users don't get confused as to why some of their dependencies are being built without first being downloaded.

src/dune_rules/pkg_build_progress.ml
gridbugs: Fixed

src/dune_rules/pkg_build_progress.mli
gridbugs: Ok I updated this PR so that now there is an `Action.t` that prints a log message when it runs that gets injected before downloading and building a package.

src/dune_rules/pkg_build_progress.ml
```diff
@@ -0,0 +1,75 @@
...
+  type ('path, 'target) t =
+    { message : Message.t
+    ; output : Path.Build.t
+    }
+
+  let name = "progress-action"
+  let version = 1
+  let is_useful_to ~memoize:_ = true
+  let bimap { message; output } _f _g = { message; output }
+  let encode _ _ _ = Sexp.List []
```
rgrinberg, line 54: Why did you leave this unimplemented?

src/dune_rules/pkg_build_progress.ml
```diff
@@ -0,0 +1,75 @@
...
+
+  let action t ~ectx:_ ~eenv:_ =
+    (match Path.Untracked.exists (Path.build t.output) with
+     | false -> Message.display t.message
+     | true -> ());
```
rgrinberg, line 59: Did you check if this clause ever executes? Isn't the target directory created every time a package is built?

src/dune_rules/pkg_rules.ml
```diff
@@ -1798,7 +1809,46 @@ let build_rule context_name ~source_deps (pkg : Pkg.t) =
       |> Action_builder.return
       |> Action_builder.with_no_targets
     in
-    Action_builder.progn (build_and_install @ [ install_file_action ])
+    (* This action will print a "Downloading" message for the package,
+       only if its source directory is not yet present. Not all
+       packages have source archives and the "Downloading" message that
+       prints while fetching the source arcive won't have been printed
+       for these packages. This is to avoid the situation where dune
```
rgrinberg, line 1816: You're saying that this situation is confusing, but I think it reflects exactly what dune is doing and the attempt to workaround as confusing. It's indeed the case that dune will have to rebuild a package without re-downloading the sources. Why should we pretend otherwise?

src/dune_rules/pkg_build_progress.ml
rgrinberg: I now see why this can work. It's because the path in this action isn't being sandboxed.

src/dune_rules/pkg_build_progress.ml
rgrinberg: In particular, the implementation of the action digest should reflect:

1. The name of the action
2. The version

This one is debatable, but perhaps you should also include the message of the action? It will cause the full action to re-run if the message changes, but that's unlikely to occur often.

src/dune_rules/pkg_rules.ml
rgrinberg: I should add that there's a far more common case for rebuilding a package without re-dowloading its sources. Consider upgrade a dependency of package. You will need download the dependency, build it, and then download the package itself without re-downloading.

> Not all packages have source archives and the "Downloading" message that prints while fetching the source arcive won't have been printed for these packages

A better way to solve this would be to modify the fetch rules to insert a "download" action before fetching.

src/dune_rules/pkg_build_progress.ml
gridbugs: Good spot. I forgot to write this. Fixed.

src/dune_rules/pkg_rules.ml
gridbugs: For the sake of simplicity/accuracy I've removed this message for now. Now dune will only print a "downloading" message when a package's source archive is downloaded. We'll be testing this feature in some user interviews so if it turns out to confuse people we can find a way to solve that.

Regarding extra sources, for the initial version of this feature I don't think it's necessary to print anything. I'm hesitant to add a "download" message before fetching as we do with the source archive as it's common for packages to have multiple extra sources and I don't want to print the message for each one. There's a bunch of ways we could solve that problem but I want to get the basic feature merged before thinking about that.

src/dune_rules/pkg_build_progress.ml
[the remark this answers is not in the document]
gridbugs: Heads up that I changed this back to not print "Downloading" for packages without a source archive for simplicity/accuracy. I'm happy to explore adding messages for extra sources in a later change.

src/dune_rules/pkg_rules.ml
rgrinberg: It will be even simpler and more correct if you remove the `output` parameter. I don't really understand this explanation:

> The [output] argument is the path of a file that will be created by the operation that the progress message is about so that the message can be elided if the file already exists.

Why do you want to not print the message if the target directory already exists? This is not the criteria for rebuilding packages. Essentially, you're omitting this progress message whenever the package is rebuilt after the first time.

gridbugs: In a previous implementation of this feature I found that it would print the progress messages for each dependency, even if they didn't get (re)built/downloaded, and the `output` param was a workaround to prevent this. However I just checked and now that printing status messages is implemented as an action this doesn't happen, so there's no reason to keep the `output` parameter. I've removed it.

rgrinberg (approved):

PR #10804, opened 2024-08-06
Bump oneshot webserver timeout

gridbugs:
This is timing out more frequently in CI recently, so this increases the timeout from 5s to 30s.

### DIFF
--- a/test/http/http.ml
+++ b/test/http/http.ml
@@ -38,7 +38,7 @@ module Server = struct
 
   let auto_shutdown_seconds =
     match Sys.getenv_opt "DUNE_WEBSERVER_TIMEOUT" with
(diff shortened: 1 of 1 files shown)


Leonidas-from-XIV (approved):
Looks good but can you also update `test/blackbox-tests/test-cases/pkg/compute-checksums-when-missing.t` to remove the environment variable there as it just sets the timeout to be lower?

Leonidas-from-XIV:
Ah this timeout stuff is annoying, I've created #10808 to remove the whole thing (and speed it up by removing the need for `wait`).

gridbugs:
Closing in favour of https://github.com/ocaml/dune/pull/10808.

PR #10808, opened 2024-08-06
Replace `oneshot_webserver` with fake curl

Leonidas-from-XIV:
This solves a lot of the waiting, leftover HTTP processes etc. by replacing the HTTP calls with a fake `curl` binary that copies the inputs to wherever Dune wants `curl` to output the downloaded files.

Original idea by @emillon.


test/blackbox-tests/test-cases/pkg/unavailable-package-source.t
```diff
@@ -34,10 +34,11 @@ Git
...
-on 35000
-  $ runtest "(fetch (url \"https://0.0.0.0:35000\"))" 2>&1 | sed -ne '/Error:/,$ p'
-  Error: curl returned an invalid error code 7
-         
-         
+HTTP
+
+  $ runtest "(fetch (url \"https://0.0.0.0:35000\"))" 2>&1 | sed -ne '/Error:/,$ p' | sed -e 's/".*"/"<PATH>"/'
+  Error:
+  failed to unpackage archive downloaded from https://0.0.0.0:35000
```
gridbugs, line 41:
```suggestion
  failed to unpack archive downloaded from https://0.0.0.0:35000
```

gridbugs:
You can probably also remove the `test/http` directory as it was only used by the oneshot webserver.

gridbugs (approved):

Leonidas-from-XIV:
> You can probably also remove the `test/http` directory as it was only used by the oneshot webserver.

I thought so too, but it is used by the unit tests and I didn't want to spend too much time on injecting the fake curl into them, so I'd leave it around until we run into issues there.

test/blackbox-tests/test-cases/pkg/unavailable-package-source.t
Leonidas-from-XIV: This was from `Fetch` not the test, but you're right the wording was odd. Fixed it!

PR #10810, opened 2024-08-08
Use [Config.Toggle] for toolchains

gridbugs:
This changes the variable that enables toolchains from DUNE_CONFIG__TOOLCHAINS_ENABLED=[true|false] to
DUNE_CONFIG__TOOLCHAINS=[enabled|disabled], following the established convention for such variables in dune, and for compatibility with [upcoming support for compile-time configuration](https://github.com/ocaml/dune/pull/10724).

### DIFF
--- a/src/dune_config/config.ml
+++ b/src/dune_config/config.ml
@@ -85,6 +85,7 @@ let make ~name ~of_string ~default =
   t
 ;;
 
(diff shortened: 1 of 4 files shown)


Leonidas-from-XIV (approved):
More consistency is great!

PR #10811, opened 2024-08-09
feat: configure-time-toggles (cont)

Leonidas-from-XIV:
This is a continuation of @emillon's work in #10724; I've rebased that PR and removed the witnesses by adding the toggles to a second-non heterogenous list of config options that only contains the toggles.

I am not sure this was the direction Etienne wanted to take it, so I am creating a new PR for it.

### DIFF
--- a/src/dune_config/config.mli
+++ b/src/dune_config/config.mli
@@ -22,6 +22,8 @@ module Toggle : sig
   val to_dyn : t -> Dyn.t
 end
 
+val set_configure_time_toggles : names:string list -> unit
+
 (** [make ~name ~of_string ~default] registers a config value called [name],
     parsed using [of_string], defaulting to [default]. *)
 val make : name:string -> of_string:(string -> ('a, string) result) -> default:'a -> 'a t
@@ -67,3 +69,5 @@ val threaded_console_frames_per_second : [ `Default | `Custom of int ] t
     Note that environment variables take precedence over the values passed here
     for easy overriding. *)
 val init : (Loc.t * string) String.Map.t -> unit
+
+val party_mode : Toggle.t t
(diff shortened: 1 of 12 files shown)


rgrinberg:
Are you planning on modifying the defaults for any existing configuration values? If not, I don't see why any modifications to dune_config are necessary.

gridbugs (approved):

gridbugs:
> Are you planning on modifying the defaults for any existing configuration values? If not, I don't see why any modifications to dune_config are necessary.

For the developer preview we'll be enabling `toolchains` by default. There's are a few more features currently in progress that will also be disabled by default on the `main` branch and opam releases of dune but enabled by default in the developer preview.

rgrinberg:
There's couple of things here that should have never been merged:
1. the changes to dune_config
2. the changes to dune_engine
can you revert those?

src/dune_config/config.mli
```diff
@@ -67,3 +69,5 @@ val threaded_console_frames_per_second : [ `Default | `Custom of int ] t
     Note that environment variables take precedence over the values passed here
     for easy overriding. *)
 val init : (Loc.t * string) String.Map.t -> unit
+
+val party_mode : Toggle.t t
```
rgrinberg, line 73: This is just a dummy config, do you still need it after testing?

rgrinberg:
In particular, here's a way to accomplish what you need without touching dune_config (which we are using internally):

1. Introduce individual flags for all the options you would like to set the default for at configure time
2. Use the options defined in 1. as the defaults when defining your config values

Everything else seems unnecessary

src/dune_config/config.mli
Leonidas-from-XIV: No, this is just for testing, it can safely be removed. I haven't considered it to be of particular importance but it can be reverted without any problem.

PR #10814, opened 2024-08-10
fix(pkg): do not lock unreachable packages

rgrinberg:
unreachable packages are those that aren't reachable via the dependencies of local packages

typically, such packages are included when solving via post deps

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -758,6 +758,53 @@ module Solver_result = struct
     }
 end
 
+let reject_unreachable_packages =
+  let reachable deps_of_package ~roots =
+    let seen = ref Package_name.Set.empty in
+    let rec loop = function
+      | [] -> ()
+      | pkg :: rest ->
+        if Package_name.Set.mem !seen pkg
+        then loop rest
+        else (
+          seen := Package_name.Set.add !seen pkg;
+          let deps =
+            match Package_name.Map.find deps_of_package pkg with
+            | None -> []
+            | Some deps -> deps
+          in
+          loop (List.rev_append deps rest))
+    in
+    loop roots;
+    !seen
+  in
+  fun ~local_packages ~pkgs_by_name ->
+    let roots = Package_name.Map.keys local_packages in
+    let pkgs_by_name =
+      Package_name.Map.merge pkgs_by_name local_packages ~f:(fun name lhs rhs ->
+        match lhs, rhs with
+        | None, None -> assert false
+        | Some _, Some _ ->
+          Code_error.raise
+            "package is both local and returned by solver"
+            [ "name", Package_name.to_dyn name ]
+        | Some (pkg : Lock_dir.Pkg.t), None -> Some (List.map pkg.depends ~f:snd)
+        | None, Some (pkg : Local_package.For_solver.t) ->
+          let deps =
+            List.map pkg.dependencies ~f:(fun (d : Package_dependency.t) -> d.name)
+          in
+          let depopts =
+            List.filter_map pkg.depopts ~f:(fun (d : Package_dependency.t) ->
+              Option.some_if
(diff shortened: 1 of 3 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -860,6 +880,33 @@ let solve_lock_dir
                       (Package_name.to_string name)
                       (Package_name.to_string dep_name)
                   ]));
+        let reachable =
+          let roots = Package_name.Map.keys local_packages in
+          let pkgs_by_name =
+            Package_name.Map.merge pkgs_by_name local_packages ~f:(fun _name lhs rhs ->
+              match lhs, rhs with
+              | None, None -> assert false
+              | Some _, Some _ -> assert false
```
gridbugs, line 889: This case should be a code error as it's conceivable that there is a package name in common between local and locked packages, _and_ due to a bug dune failed to detect this case prior to this point.

src/dune_pkg/opam_solver.ml
```diff
@@ -860,6 +880,33 @@ let solve_lock_dir
                       (Package_name.to_string name)
                       (Package_name.to_string dep_name)
                   ]));
+        let reachable =
+          let roots = Package_name.Map.keys local_packages in
+          let pkgs_by_name =
```
gridbugs, line 885: The `solve_lock_dir` function is getting too long. Can you move this logic for computing the name->deps map to a separate function or move it into the `reachable` function itself.

src/dune_pkg/opam_solver.ml
```diff
@@ -758,6 +758,26 @@ module Solver_result = struct
     }
 end
 
+let reachable pkgs ~roots =
```
gridbugs, line 761: The `pkgs` argument maps package names to their dependencies. Can you rename it to reflect this.

gridbugs:
Looks like a useful change. I have a few readability suggestions but generally happy with this.

gridbugs (approved):

src/dune_pkg/opam_solver.ml
```diff
@@ -758,6 +758,26 @@ module Solver_result = struct
...
+      if Package_name.Set.mem !seen pkg
+      then loop rest
+      else (
+        seen := Package_name.Set.add !seen pkg;
+        let deps =
+          match Package_name.Map.find pkgs pkg with
+          | None -> []
+          | Some deps -> deps
+        in
+        loop (List.rev_append deps rest))
```
gridbugs, line 775: What's the point of reversing the dependencies before pre-pending them to the list of names to expand? Is it just that it's more efficient that using `@` and has no impact on the correctness of this search?

src/dune_pkg/opam_solver.ml
```diff
@@ -758,6 +758,26 @@ module Solver_result = struct
     }
 end
 
+let reachable pkgs ~roots =
+  let seen = ref Package_name.Set.empty in
+  let rec loop = function
+    | [] -> ()
+    | pkg :: rest ->
+      if Package_name.Set.mem !seen pkg
```
gridbugs, line 766: Nitpick: Checking if a package is seen only after popping it from the stack to expand means that common packages like `ocaml` are going to be added to the stack many times. It would be more efficient to perform this check and update `seen` at the point where a package would be pushed to the stack.

src/dune_pkg/opam_solver.ml
rgrinberg: It's for performance, though it can be serious. `@` reverses the list on every application, therefore it requires traversing the entire list.

src/dune_pkg/opam_solver.ml
rgrinberg: I think this does mean we might traverse the same package more than once though. For example, if `ocaml` is listed as a dependency twice.

gridbugs (approved):

PR #10824, opened 2024-08-13
feature(pkg): enable/disable toolchains at compile time

rgrinberg:
For now, please start with something simple like this. We use Dune_config internally, and would rather not have its implementation changed. Changes to the engine should be proposed first.


rgrinberg:
This was just a demonstration. Feel free to do it however you want within the parameters we've agreed on

PR #10825, opened 2024-08-13
refactor: Use `Stdlib.String(Labels)` implementations if available

Leonidas-from-XIV:
This uses the Stdlib implementations if the stdlib supports these. Has the advantage that if the stdlib has an optimized implementation, we automatically get to benefit of it.

This uses a trick to define the fallback functions and overwrite them if they're available in the included module, otherwise they will be used. Needs to silence warning 32 but that's a small price to pay to avoid conditional compilation.

### DIFF
--- a/otherlibs/stdune/src/string.ml
+++ b/otherlibs/stdune/src/string.ml
@@ -1,5 +1,27 @@
 (* Because other the syntax s.[x] causes trouble *)
 module String = Stdlib.String
+
+module StringLabels = struct
+  (* functions potentially in the stdlib, depending on OCaml version *)
+
+  let[@warning "-32"] exists =
+    let rec loop s i len f =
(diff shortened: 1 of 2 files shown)


otherlibs/stdune/src/string.ml
```diff
@@ -221,12 +231,6 @@ let enumerate_one_of = function
   | s -> "One of " ^ enumerate_or s
 ;;
 
-let concat ~sep = function
-  | [] -> ""
-  | [ x ] -> x
-  | xs -> concat ~sep xs
-;;
```
Leonidas-from-XIV, line 228: This function exists in the stdlib in all OCaml versions we support with the same semantics  and only works because it accidentally calls `Stdlib.StringLabels.concat` (missing the `rec` flag); otherwise it never uses the `sep` argument.

rgrinberg (approved):

PR #10831, opened 2024-08-17
Map `(re_export)` to `exports` field in `META` files

### LINKED ISSUE #10830: Map `(re_export)` to `exports` field in `META` files
There is a proposal https://discuss.ocaml.org/t/proposal-a-new-exports-field-in-findlib-meta-files/13947/40 for a convention to extend the notion that we call `(re_export)` in  Dune to general `META` files via a new field called `exports`. In order to play well with the ecosystem, Dune should map `(re_export)` to the `exports` file in `META` and vice-versa.

I suspect these aspects will become more relevant as time passes and `(implicit_transitive_deps false)` is adopted more widely, now that it is correctly implemented by `-H`.

nojb:
Fixes #10830 

See #10830 and the links there (especially https://discuss.ocaml.org/t/proposal-a-new-exports-field-in-findlib-meta-files/13947) for the context. Each commit contains a corresponding test.

- The first commit makes Dune output an `exports` field in the generated `META` containing any library dependencies which appear in `(re_export)`.
- The second commit makes Dune interpret `exports` field in `META` files by mapping the libraries listed in it to `(re_export)`. Note that `exports` is semantically interpreted as a list of boolean flags on the existing `requires` field. This means: libraries listed in `exports` but not in `requires` are ignored, and `exports` is assumed not to be "predicated" (in the Findlib sense).
- The third commit adapts `(deprecated_library_name)` to use `(exports)` (this is actually one of the motivating use-cases mentioned in https://discuss.ocaml.org/t/proposal-a-new-exports-field-in-findlib-meta-files/13947)


nojb:
> * `exports` is assumed not to be "predicated" (in the Findlib sense).

Actually, I am not sure about this point. Maybe we should interpret predicates in the new `exports` field.

rgrinberg:
I think you need to guard the Meta generation behind a 3.17 check.

nojb:
> I think you need to guard the Meta generation behind a 3.17 check.

Done in https://github.com/ocaml/dune/pull/10831/commits/c2cdbb3f376fe6c6b22e2140dffd790309e1aad8. Thanks!

nojb:
> Are there any tests for how we interpret this field from META files? I see how we generate them, but we should test how they're read (whenever there's no dune-package file)

Yes, there is one expect test: https://github.com/ocaml/dune/pull/10831/files#diff-b575837081f645adcc29f1302fc07bb1b2044801712369ccbe094aea0e41f97b

rgrinberg:
I see a unit test that we parse them correctly. But it would be nice to demonstrate that we can actually build against something that such a META. This isn't strictly necessary, but the only tangible benefit this feature provides to dune users is the ability to use Daniel's libraries correctly. So I do think this is the most important criteria for us.

nojb:
> it would be nice to demonstrate that we can actually build against something that such a META

OK, I will add a test showing this later today. Thanks!

nojb:
> > it would be nice to demonstrate that we can actually build against something that such a META
> 
> OK, I will add a test showing this later today. Thanks!

I pushed a new dedicated test `meta-exports.t` that does this now: https://github.com/ocaml/dune/pull/10831/files#diff-0f4411a3b78fcb4907a6dd8530f238a7593c81dbf0bd1bb9069e9cb117a4f4f3. Since the new test checks both the ability to consume and generate the `exports` fields, I reverted the changes to the existing test `meta-gen.t`.

rgrinberg (approved):

nojb:
Thanks for the approval @rgrinberg. I pushed a Changelog entry in preparation for merging.

@dbuenzli: does the PR look OK from your side? You can take a look at the CRAM test `meta-exports.t` to get an idea of what was implemented.

dbuenzli:
> @dbuenzli: does the PR look OK from your side? You can take a look at the CRAM test `meta-exports.t` to get an idea of what was implemented.

I did that. I would say yes.

nojb:
> > @dbuenzli: does the PR look OK from your side? You can take a look at the CRAM test `meta-exports.t` to get an idea of what was implemented.
> 
> I did that. I would say yes.

Thanks for the confirmation!

PR #10835, opened 2024-08-20
feat: Config defaults for dune-project

### LINKED ISSUE #10027: Config options for default author, maintainer etc
## Desired Behavior

Running `dune init project myproject` uses values from ~/.config/dune/config to populate values in `dune-project` such as `authors`, `maintainers` and possibly more.

## Example

`dune init project myproject` generates, among other files, a `dune-project` file with stanzas such as `(authors "Author Name")` and `(maintainers "Maintainer Name")` which has to be edited manually every time even if my name is the same between invocations.

H-ANSEN:
This is a pull request in reference to and would close #10027 by adding the ability to configure defaults for the `authors`, `maintainers`, and `license` stanzas of the `dune-project` file. Values are sourced from the config file when using the `dune init project` command allowing for preferred preset values to be added. Values can be specified by adding the following stanza to the dune config file
```
(project_defaults
 (authors <string(s)>)
 (maintainers <string(s)>)
 (license <string(s)>))
```


H-ANSEN:
This is ready for review @Leonidas-from-XIV ! Tagging you here sense I never got any initial response and you reviewed my previous pull :)

src/dune_config_file/dune_config_file.ml
```diff
@@ -16,6 +16,35 @@ module Dune_config = struct
      simplicity *)
   let syntax = Stanza.syntax
 
+  module Project_defaults = struct
+    type t =
+      { authors : string list option
```
rgrinberg, line 21: Can't this be `string list`? If you aren't allowing the empty list, then `[]` can just mean the field is left unspecified. `string list option` is only useful if `(authors)` is allowed.

rgrinberg (changes requested):

src/dune_config_file/dune_config_file.ml
H-ANSEN: It could be changed to `string list`. I think my original intention was to just copy the allowed behavior in the `dune-project` file, when I was testing it seems that `(authors)` is allowed there and didn't cause any issue.

Also in [src/dune_lang/package_info.mli](https://github.com/ocaml/dune/pull/10835/files) the function `example` takes a `string list option` where a value of `[]` literally means there were no arguments provided and a value of `None` uses the default value of `Author Name <author@example.com>`.

Let me know what you think, and if needed I can change it!

H-ANSEN: I see what you mean by just having the empty list represent the absence of the stanza, I also figured allowing something like `(authors)` was a bit less error prone for a user especially when it came to throwing errors when someone is trying to make a new project.

rgrinberg: In `Package_info.decode` don't we allow the empty list for these fields? In that case, `string list option` indeed makes sense there. My objection is that your code pretends that `None` and `Some []` are possible, but validation in fact forbids that.

H-ANSEN: In `package_info.ml` authors, license, and maintainers are all decoded as a `string list option`. The encoder also uses the same scheme defaulting to the empty list if a value of `None` is provided.

H-ANSEN: Also to further clarify, each of the sub stanzas of `project_defaults` is optional so you could do something like this
```
(project_defaults
 (authors)
 (license MIT))
```
in this case authors would have the empty list value and maintainers would have a value of `None`

rgrinberg: > is optional so you could do something like this

Looking at your code, I don't think your example will work. You're using `repeat1` which will forbid `(authors)`.

H-ANSEN: I guess your right! What way should we have it then? Should we allow `(authors)` or should I simplify it down to just use `string list` over `string list option`

rgrinberg: Both are fine I think. On one hand, we allow the empty file in the `dune-project` file. So it would be more consistent to allow it here as well. On the other hand, the empty list is fairly useless as a default in this case.

H-ANSEN: I think we should allow it just to keep constancy, that way you can edit your `config` file as if you were editing your `dune-project` file!

doc/reference/config/project_defaults.rst
```diff
@@ -0,0 +1,50 @@
+project_defaults
+----------------
+
+Specify default values for stanzas ``authors``, ``maintainers``, and ``license``
+of the dune-project file when initializing a project with ``dune init proj``.
```
Leonidas-from-XIV, line 5: `dune-project` needs to be monospaced I think. Or link to the `dune-project` reference.

doc/reference/config/project_defaults.rst
```diff
@@ -0,0 +1,50 @@
...
+   .. code:: dune
+
+       (project_defaults
+        (maintainers
+         "Jane Doe <jane.doe@example.com>"
+         "John Doe <john.doe@example.com>"))
+
+.. describe:: (license <string(s)>)
+
+   Specify license.
```
Leonidas-from-XIV, line 43: Maybe worth noting that SPDX format is preferred.

doc/reference/config/project_defaults.rst
```diff
@@ -0,0 +1,50 @@
...
+.. code:: dune
+
+   (project_defaults
+    <optional-fields>)
+
+``<optional-fields>`` are:
+
+.. describe:: (authors <string(s)>)
+
+   Specify authors.
```
Leonidas-from-XIV, line 17: I wonder, if this could link to the definition of `authors` in the `dune-project` reference?

Leonidas-from-XIV (commented):
Thanks for the PR, I appreciate it comes with a test and documentation updates. The code looks fine, I only have a few nitpicks about the documentation.

Could you add a short changelog entry as well?

rgrinberg (approved):

doc/reference/config/project_defaults.rst
H-ANSEN: I couldn't find any relevant definition of `authors` in the reference, I'm sure I am just glancing over it though. I made all the other requested changes as well as adding a `.. versionadded 3.17` to the top of the doc page!

PR #10841, opened 2024-08-21
RFC: Make it possible to disable line wrapping in tests

Leonidas-from-XIV:
Sometimes cram tests are not reproducible if different platforms output different length errors, in which case even editing out the errors with common tools like `sed` does not help as the wrapping happens before the `sed` invocation runs, thus the message is split over multiple lines.

This PR suggests adding an environment variable to disable the line wrapping globally by setting the margin to a high value.

Some follow-up UX improvements might me of interest, e.g. reading the width of the terminal automatically and setting the value to the width of the users terminal. But these are outside of the scope of this proposal.


src/dune_rules/colors.ml
```diff
@@ -41,6 +41,13 @@ let mark_open_stag = function
   | _ -> ""
 ;;
 
+let setup_terminal_width () =
+  let env = Env.initial in
+  match Env.mem ~var:"DUNE_TEST__DISABLE_LINE_BREAK" env with
+  | false -> ()
+  | true -> Format.set_margin Int.max_int
```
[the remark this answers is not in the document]
Leonidas-from-XIV: It's only available starting from OCaml 5.2, so for a nicer implementation we could wrap `Format` in Stdune and add it there.

Leonidas-from-XIV:
I've added a test for this but at the moment it shows that my implementation doesn't work. Nevertheless I am curious to hear people's opinion on the feature itself to see whether it would be worth getting this to work properly.

I run into the wrapping issue every now and then and it is annoying to work-around, thus I'd like to have a good way to deal with this issue going forward.

src/dune_rules/colors.ml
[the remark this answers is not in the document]
maiste: My bad, I didn't see! Maybe, and we could make it evolved with the OCaml version we use to build dune.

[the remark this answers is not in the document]
Leonidas-from-XIV: Added it.

gridbugs:
I've never run into this issue personally but it sounds useful.

rgrinberg:
> Some follow-up UX improvements might me of interest, e.g. reading the width of the terminal automatically and setting the value to the width of the users terminal. But these are outside of the scope of this proposal.

How does this help with making the tests reproducible? You would think that making the test output depend on the terminal width is going to make things less reproducible. Or is this an orthogonal UX suggestion?

src/dune_rules/colors.ml
```diff
@@ -41,6 +41,13 @@ let mark_open_stag = function
   | _ -> ""
 ;;
 
+let setup_terminal_width () =
+  let env = Env.initial in
+  match Env.mem ~var:"DUNE_TEST__DISABLE_LINE_BREAK" env with
+  | false -> ()
+  | true -> Format.set_margin Format.pp_infinity
```
rgrinberg, line 48: I think you need to set this for stderr as well as stdout

rgrinberg:
Adding a hack to make it easier to write tests seems fine to me. You should use `Dune_config.make` to define this constant though

Leonidas-from-XIV:
> How does this help with making the tests reproducible? You would think that making the test output depend on the terminal width is going to make things less reproducible.

This PR/RFC is just about setting the wrapping to "infinity" to avoid any automatic box wrapping.

> Or is this an orthogonal UX suggestion?

Yes, that's something else I thought about because a lot of output I see from all kinds of tools that use `Fromat` looks like this

```
Here is some
text that
has been wrapped
despite there
being 262 characters
in my terminal
```

But that's a separate concern.

Leonidas-from-XIV:
> You should use `Dune_config.make` to define this constant though

That was my initial implementation, but the issue is that `Dune_config.get` will fail because `init` hasn't been called yet and I want to override the values as early in the execution as possible to avoid the `Format` being used before. Where would be a good way to call `Dune_rules.Colors.setup_terminal_width` where the config is guaranteed to be initialized?

src/dune_rules/colors.ml
Leonidas-from-XIV: I needed to change/replace `User_message.prerr`, indeed, though this code turned out to not do what I hoped for since apparently the global `Format` margin is not respected when printing to strings.

otherlibs/stdune/src/ansi_color.ml
```diff
@@ -494,7 +494,7 @@ let rec tag_handler buf current_styles ppf (styles : Style.t list) pp =
     (Style.escape_sequence_buf buf (current_styles :> Style.t list))
 ;;
 
-let make_printer supports_color ppf =
+let make_printer ~line_break ~supports_color ppf =
```
maiste, line 497: Is there a reason for using a positional argument instead of an `optional` with a `true` as default value?

Leonidas-from-XIV: We generally try to avoid optional arguments in the Dune code base as opting users unknowingly into default behaviors is not great. In this case, the outcome of this is fairly limited as `make_printer` is only really used by `print` and `prerr`.

Leonidas-from-XIV:
Turns out it is actually not safe to ask for `Dune_config` in `Dune_console`, at least not in the general case.

otherlibs/stdune/src/ansi_color.mli
```diff
@@ -94,7 +94,11 @@ module Style : sig
   val escape_sequence : t list -> string
 end
 
-val make_printer : bool Lazy.t -> Format.formatter -> (Style.t list Pp.t -> unit) Staged.t
+val make_printer
+  :  line_break:bool
```
rgrinberg, line 98: How about we avoid changing the public API here for the tests? Not that the change itself is bad, but when we're changing the API just to accommodate 1 test out of hundreds, it's hard to justify this trade-off. If you can't use Dune_config, then so bet it. Feel free to use `Sys.getenv` directly. It's just for the tests anyway.

rgrinberg:
The only reason Dune_config has this limitation is that every config value in it can be set in the dune-config or workspace files. That requires an explicit initialization step after those files are read and the config values cannot be used until those are initialized. In practice, I don't think it turned out to be very useful and we can consider getting rid of this feature if it will make Dune_config usable in some contexts where it's currently not. This is a developer, and not a user facing feature, so we can just do whatever's more convenient for us.

otherlibs/stdune/src/ansi_color.mli
Leonidas-from-XIV: In this particular case `make_printer` is only public for a test, the regular code uses `print` or `prerr`, thus I considered it not too invasive.

Anyway, I removed `Dune_config` and used `Sys.getenv` and reverted all these changes.

Leonidas-from-XIV:
> In practice, I don't think it turned out to be very useful and we can consider getting rid of this feature if it will make Dune_config usable in some contexts where it's currently not.

That's probably useful, perhaps through a two stage system where parts that don't care about the values from the config can read it early, whereas parts that want the full config will retain the current behavior but definitely the scope of a different PR.

So I'm reading the variables out of the environment directly, while keeping a `Dune_config`-like interface, in case we want to migrate to that in the future.

rgrinberg (approved):

PR #10843, opened 2024-08-21
feat: Adding configure-time toggles

ElectreAAS:
This PR is a new attempt at doing what was started in various PRs already (see #10710, #10724, #10810, #10811, #10824), but without the problematic changes to `dune_config/config.ml` or the engine.
The toggles introduced with this PR are `toolchains` and `pkg_build_progress`.
Both of them are disabled by default, but they can be modified either at compile time with `./configure --toggles name1,name2`, or at run time with `DUNE_CONFIG__NAME1=enabled`.


ElectreAAS:
I would've liked to add tests that showcase this new feature, but I'm very unsure as to how to make them...

src/dune_config/compile_time.ml
```diff
@@ -0,0 +1,19 @@
...
+let initialized = ref false
+let toggled = ref []
+
+let init ~names =
+  if !initialized then Code_error.raise "Compile_time.init: already initialized" [];
+  toggled := names;
+  initialized := true
+;;
+
+let set name =
```
gridbugs, line 12: Should this raise an error if it's called before `init`?

boot/configure.ml
```diff
@@ -70,6 +74,10 @@ let () =
       , Arg.String (set_dir datadir)
       , "DIR where files for the share_root section are installed for the default build \
          context" )
+    ; ( "--toggles"
+      , Arg.String set_toggles
+      , "NAMES comma-separated list of configuration options to be set to 'enabled'. See \
+         the list of options in dune_config/compile_time." )
```
gridbugs, line 80: When referring to source filenames in messages like this my preference is to refer to the full filename including its extension.

gridbugs: Also this takes the string name of the variable and these are defined in the .ml file, but the documentation about what each variable does is in the .mli file. Something about that doesn't feel right to me. Like we have a well-documented interface in the mli file but users still have to look at the ml file in order to find out the strings to pass here.

boot/configure.ml
```diff
@@ -90,5 +98,7 @@ let () =
   pr "  ; sbin = %s" (option string !sbindir);
   pr "  ; libexec_root = %s" (option string !libexecdir);
   pr "  }";
+  pr "";
+  pr "let init () = %s" (toggle_inits !toggles);
```
gridbugs, line 102: It would be nice if typos in config variable names or attempts to enable non-existent variables could be caught at compile time (currently these cases are ignored if I understand correctly). One way to do this might be be introducing a variant type for the config options and requiring the user to pass the constructor names to `--toggles`. The constructors could also be documented in the mli file avoiding the issue for users to look inside the ml file to use this feature.

boot/configure.ml
```diff
@@ -70,6 +74,10 @@ let () =
       , Arg.String (set_dir datadir)
       , "DIR where files for the share_root section are installed for the default build \
          context" )
+    ; ( "--toggles"
```
gridbugs, line 77: I think "--enable-toggles" would be a more accurate name for this argument.

rgrinberg: As discussed in the meeting, the `--help` argument should be useful. In particular, we want:

1. a separate command line flag (or argument) for every toggle
2. a description of what this flag does
3. a note that states that this flag is experimental and packagers should not rely on it

rgrinberg (changes requested):

boot/configure.ml
ElectreAAS: > As discussed in the meeting, the `--help` argument should be useful. In particular, we want:
> 
>     1. a separate command line flag (or argument) for every toggle
> 
>     2. a description of what this flag does
> 
>     3. a note that states that this flag is experimental and packagers should not rely on it

Done!

ElectreAAS:
> As discussed in the meeting, the --help argument should be useful. In particular, we want:
> 
>  - a separate command line flag (or argument) for every toggle
>  - a description of what this flag does
>  - a note that states that this flag is experimental and packagers should not rely on it

Done!

boot/configure.ml
ElectreAAS: This should be resolved now, but I did it the way Rudi suggested, with `--enable-toolchains`.
So if someone tries `--enable-doesntexist` or `--enable-tolochains`, the error is going to be picked up

src/dune_config/compile_time.ml
Leonidas-from-XIV, line 12: I think `make` would be a more appropriate name.

ElectreAAS: Done!

rgrinberg:
There's a couple of things I adjusted in this PR:

* Moved the new module to the rules. No reason for it to live in Dune_config as it would be just one more unnecessary thing to sync. If you need it outside the rules, feel free to move it up to its own library.

* I made all the defaults explicitly listed in defaults.mli. This is the convention that we have for all configure time defaults. We just generate simple values instead of code. As a consequence, all the set/make stuff becomes unnecessary and we have less strings that we can typo.

This is good enough for now.

ElectreAAS:
> I made all the defaults explicitly listed in defaults.mli. This is the convention that we have for all configure time defaults. We just generate simple values instead of code. As a consequence, all the set/make stuff becomes unnecessary and we have less strings that we can typo.

Ah yes I see what you meant with the new code. That was code I adapted from one of the original PRs, but didn't know why it was made that way. The new way looks simpler indeed
Thanks!

PR #10847, opened 2024-08-27
Return [result] when loading lockdir

gridbugs:
Adds versions of functions that load lockdirs that return [Error] when loading fails rather than raising an exception.

### DIFF
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -634,32 +634,35 @@ struct
     let open Io.O in
     Io.stats_kind lock_dir_path
     >>| function
-    | Ok S_DIR -> ()
+    | Ok S_DIR -> Ok ()
     | Error (Unix.ENOENT, _, _) ->
-      User_error.raise
-        ~hints:
-          [ Pp.concat
-              ~sep:Pp.space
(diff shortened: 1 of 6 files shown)


gridbugs:
Context is here: https://github.com/ocaml/dune/pull/10647#issuecomment-2309581105

rgrinberg (approved):

src/dune_pkg/lock_dir.ml
```diff
@@ -634,32 +634,35 @@ struct
...
+      Error
+        (User_error.make
+           [ Pp.textf "%s is not accessible" (Path.Source.to_string lock_dir_path)
+           ; Pp.textf "reason: %s" (Unix_error.Detailed.to_string_hum e)
+           ])
     | _ ->
-      User_error.raise
-        [ Pp.textf "%s is not a directory." (Path.Source.to_string lock_dir_path) ]
+      Error
+        (User_error.raise
```
Leonidas-from-XIV, line 659: Shouldn't this be `User_error.make`? Or, if we keep this as something that `raises` then without the `Error` constructor?

gridbugs: Ah yep good catch, I missed that one. Fixed.

PR #10849, opened 2024-08-27
Subtle hint that topics is not a keyword

punchagan:
Number of projects in the opam-repository have the word topics as one of the tags, presumably because of the structure of the `tags` sexp in the dune-project file generated on `dune init`. See ocurrent/opam-ci-check#26

### DIFF
--- a/bin/dune_init.ml
+++ b/bin/dune_init.ml
@@ -391,7 +391,7 @@ module Component = struct
             ~dir
             ~synopsis:(Some "A short synopsis")
             ~description:(Some "A longer description")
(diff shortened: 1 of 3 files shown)


Leonidas-from-XIV (approved):
I think that's a good, pragmatic change and the preset value does make "topics" seem like it has a special meaning due to the way that the formatter omits quotes on bare strings.

rgrinberg:
CHANGES entry please

punchagan:
> CHANGES entry please

Done!

PR #10850, opened 2024-08-27
feat(pkg): Enable cache on fetch actions

Leonidas-from-XIV:
This enables caching of downloaded files as part of the fetch action, if the file to be retrieved is retrieved via a cryptographic hash (and thus presumably the exact same as the first time it was downloaded).

@rgrinberg  I've added a new function to the engine to force-set the cache flag, as `add_can_go_in_shared_cache` can only be flipped to `false` but never back to `true`. It's not used at the moment, so another option would be to remove that `&&` but just in case I made it a separate function.

### DIFF
--- a/src/dune_rules/fetch_rules.ml
+++ b/src/dune_rules/fetch_rules.ml
@@ -206,7 +206,7 @@ let gen_rules_for_checksum_or_url (loc_url, (url : OpamUrl.t)) checksum =
     let make_target = make_target checksum_or_url in
     let action ~target ~kind =
       action ~url:(loc_url, url) ~checksum ~target ~kind
-      |> Action.Full.make
+      |> Action.Full.make ~can_go_in_shared_cache:true
       |> Action_builder.return
       |> Action_builder.with_no_targets
     in
@@ -327,7 +327,7 @@ let fetch ~target kind (source : Source.t) =
            .fetch context. This would just add pointless additional overhead. *)
         action ~url:source.url ~checksum:source.checksum ~target ~kind
     in
-    Action.Full.make action
+    Action.Full.make ~can_go_in_shared_cache:true action
     |> Action_builder.With_targets.return
     |> Action_builder.With_targets.add_directories ~directory_targets:[ target ]
 ;;
--- a/test/blackbox-tests/test-cases/pkg/fetch-cache.t
+++ b/test/blackbox-tests/test-cases/pkg/fetch-cache.t
@@ -1,0 +1,61 @@
+Testing that files are only fetched once.
+
+  $ . ./helpers.sh
+
+No need to set DUNE_CACHE (enabled by default) nor DUNE_CACHE_RULES as the
+fetch rules are always considered safe to cache, but we'll set a custom
+directory for the shared cache.
+
+  $ export DUNE_CACHE_ROOT=$(pwd)/dune-cache
+  $ unset DUNE_CACHE
+  $ unset DUNE_CACHE_RULES
+
+Set up a project that depends on a package that is being downloaded
+
+  $ make_lockdir
+  $ echo "Contents" > tar-contents
+  $ CONTENT_CHECKSUM=$(md5sum tar-contents | cut -f1 -d' ')
+  $ tar cf test.tar tar-contents
+  $ echo test.tar > fake-curls
+  $ SRC_PORT=1
+  $ SRC_CHECKSUM=$(md5sum test.tar | cut -f1 -d' ')
+  $ cat >dune.lock/test.pkg <<EOF
+  > (version 0.0.1)
+  > (source
+  >  (fetch
+  >   (url http://localhost:$SRC_PORT)
+  >   (checksum md5=$SRC_CHECKSUM)))
+  > EOF
+  $ cat > dune-project <<EOF
+  > (lang dune 3.17)
+  > (package (name my) (depends test) (allow_empty))
+  > EOF
+
+The first build should succeed, fetching the source, populating the cache and
+disabling the download of the source a second time.
+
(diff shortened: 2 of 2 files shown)


gridbugs (approved):

src/dune_engine/action.mli
```diff
@@ -172,5 +172,8 @@ module Full : sig
   val add_sandbox : Sandbox_config.t -> t -> t
   val add_can_go_in_shared_cache : bool -> t -> t
 
+  (** Marks action as safe to be cached *)
+  val allowed_in_shared_cache : t -> t
```
rgrinberg, line 176: I'd say just use the existing function for now.

src/dune_rules/fetch_rules.ml
```diff
@@ -206,6 +206,7 @@ let gen_rules_for_checksum_or_url (loc_url, (url : OpamUrl.t)) checksum =
     let action ~target ~kind =
       action ~url:(loc_url, url) ~checksum ~target ~kind
       |> Action.Full.make
```
rgrinberg, line 208: Actually, you can just pass the `can_go_in_shared_cache` argument to `Action.Full.make`.

src/dune_rules/fetch_rules.ml
```diff
@@ -327,6 +328,7 @@ let fetch ~target kind (source : Source.t) =
         action ~url:source.url ~checksum:source.checksum ~target ~kind
     in
     Action.Full.make action
+    |> Action.Full.allowed_in_shared_cache
```
rgrinberg, line 331: ditto

src/dune_engine/action.mli
Leonidas-from-XIV: Removed it again as it wasn't needed after all.

src/dune_rules/fetch_rules.ml
Leonidas-from-XIV: True, that's a better solution. Completely missed that.

maiste (approved):

src/dune_rules/fetch_rules.ml
```diff
@@ -182,10 +182,10 @@ let find_checksum, find_url =
 ;;
 
 let gen_rules_for_checksum_or_url (loc_url, (url : OpamUrl.t)) checksum =
-  let checksum_or_url =
+  let checksum_or_url, can_go_in_shared_cache =
     match checksum with
-    | Some (_, checksum) -> `Checksum checksum
-    | None -> `Url url
+    | Some (_, checksum) -> `Checksum checksum, Some true
+    | None -> `Url url, None
```
rgrinberg, line 188: Why do we rely on the default in this case? Either this is useful to cache or it's not.

I would say it's probably useful to cache since this step will de-duplicate the files being stored among all the workspaces. If you think this will consume too much space, then just set it to false.

test/blackbox-tests/test-cases/pkg/fetch-cache.t
```diff
@@ -0,0 +1,61 @@
...
+  $ echo "Contents" > tar-contents
+  $ CONTENT_CHECKSUM=$(md5sum tar-contents | cut -f1 -d' ')
+  $ tar cf test.tar tar-contents
+  $ echo test.tar > fake-curls
+  $ SRC_PORT=1
+  $ SRC_CHECKSUM=$(md5sum test.tar | cut -f1 -d' ')
+  $ cat >dune.lock/test.pkg <<EOF
+  > (version 0.0.1)
+  > (source
+  >   (fetch
```
rgrinberg, line 25: Should be a single space here

src/dune_rules/fetch_rules.ml
Leonidas-from-XIV: I am concerned about caching it via just the URL because there's nothing else preventing the URL from changing, so caching a source that keeps changing seems potentially incorrect. I wanted to let people override that with `DUNE_CACHE_RULES=enabled` which enables all these potentially unsafe behaviors instead of introducing another config option.

For package management it doesn't matter since just about all packages have checksums.

If you think this is alright to cache in all cases, I will gladly switch it to always `true`.

rgrinberg: > I am concerned about caching it via just the URL because there's nothing else preventing the URL from changing, so caching a source that keeps changing seems potentially incorrect

Yeah, that's an issue. Why is this related to the shared cache though? Is it because you want `dune clean` to re-download?

> I wanted to let people override that with DUNE_CACHE_RULES=enabled which enables all these potentially unsafe behaviors instead of introducing another config option.

The `DUNE_CACHE_RULES` is mostly there for backwards compat. I don't think we should rely on it for future rules. When rules are useful and safe to cache we should do it, otherwise, we disable it.

> If you think this is alright to cache in all cases, I will gladly switch it to always true.

As mentioned earlier, I think it has the potential to save users a bunch of space if they are expanding the same source that comes from a git revision in multiple workspaces.

I think using either `true` or ` false` is fine, but I don't like letting the user fiddle with `DUNE_CACHE_RULES`.

Leonidas-from-XIV: > Yeah, that's an issue. Why is this related to the shared cache though?

My understanding is that if the URL changes upstream we would continue to use the old data from the shared cache with no easy way to even invalidate that particular URL. This happens quite a lot when people depend on Github tarballs which get regenerated in in certain cases.

> The `DUNE_CACHE_RULES` is mostly there for backwards compat. I don't think we should rely on it for future rules. When rules are useful and safe to cache we should do it, otherwise, we disable it.

I see, yes. That makes sense.

rgrinberg (approved):

Issue #10842, opened 2024-08-21
[dune-pkg] Relocking files with `dune build`


maiste:
Following the discussion during the _Dune Dev Meeting_ ([cf notes](https://github.com/ocaml/dune/wiki/dev-meeting-2024-08-21)), there was the discussion about re-locking the file with `dune build` and `dune build --watch`. The retain solution was:

- `dune build` would print an error message when it is run, and it detects the dependencies have changed. The message would ask the user to re-run `dune pkg lock`. The user could then re-run `dune build` and the error about the dependency must be gone.
- `dune build --watch` would pause and display a message asking for the user to run the `dune pkg lock. Once it is done, it would unpause automatically as the _watch mode_ would detect the dependencies have changed. Thus, it would recompute the dependency graph and re-build what needs to be re-built. This would be done by checking the dependency hash and check when it changes.

This solution is a compromise between _auto-locking_ and _full manual locking_, as it would prevent the user to have their dependencies updated without noticing. 

There are some questions left here:
- Should we add a config flag to have this as an experimental feature, or should it automatically detect the lock file and be available without it?
- Is there a way in the current behavior of Dune to ""pause"" the _watch mode_?
- @rgrinberg, you mentioned the fact that people could hard-code the _opam_ commit in their workspace file, and we could take it into account when rebuilding. Is this something we want to consider here, or should we exclude this for now? IIUC, running `dune pkg lock` will ~`opam update` our package set: do we want to warn the user if the _commit_ is hard-coded?

rgrinberg:
> Should we add a config flag to have this as an experimental feature, or should it automatically detect the lock file and be available without it?

It seems like the behavior is a strict improvement so there's no need for a flag.

> Is there a way in the current behavior of Dune to ""pause"" the watch mode?

You don't need to pause anything manually. It's enough to just verify the lock file before generating any rules. An error during this stage, would prevent dune from doing anything.

>  you mentioned the fact that people could hard-code the opam commit in their workspace file, and we could take it into account when rebuilding. Is this something we want to consider here, or should we exclude this for now? IIUC, running dune pkg lock will ~opam update our package set: do we want to warn the user if the commit is hard-coded?

I'd say let's not do anything for now. Would like to hear some demand from users before developing this workflow further.

maiste:
Closing this issue as it has been done in #10851 and #10852.

PR #10851, opened 2024-08-27
feat: dune build asks for relocking

maiste:
This PR superseeds #10641. It displays an error to the user if the lock dir is out of sync with the `dune-project` when they try to execute `dune build`. It works with both the watch mode and the standard mode.

It requires #10852 to be merged first before merging this one. However, the code can be review before as they don't have direct dependencies.

It solves  #10842.

### DIFF
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -30,34 +30,43 @@ let with_metrics ~common f =
     Fiber.return ())
 ;;
 
-let raise_on_lock_dir_out_of_sync () =
-  Memo.of_thunk (fun () ->
-    let open Memo.O in
-    let lock_dir_path = Dune_pkg.Lock_dir.default_path in
-    let lock_dirs = Pkg_common.Lock_dirs_arg.of_path lock_dir_path in
-    let* per_contexts =
-      Workspace.workspace () >>| Pkg_common.Lock_dirs_arg.lock_dirs_of_workspace lock_dirs
-    in
-    let lock_dirs =
-      List.filter_map per_contexts ~f:(fun lock_dir_path ->
-        match Path.exists (Path.source lock_dir_path) with
-        | true -> Some (Dune_pkg.Lock_dir.read_disk lock_dir_path)
-        | false -> None)
-    in
-    match lock_dirs with
-    | [] -> Memo.return ()
-    | lock_dirs ->
-      let* local_packages = Pkg_common.find_local_packages in
-      let locks =
-        List.map lock_dirs ~f:(fun lock_dir ->
+let raise_on_lock_dir_out_of_sync profile =
+  let skip_check =
+    match profile with
+    | None -> false
+    | Some profile -> Profile.is_release profile
+  in
+  if skip_check
+  then Memo.Lazy.of_val ()
+  else
+    Memo.lazy_ (fun () ->
+      let open Memo.O in
+      let lock_dir_path = Dune_pkg.Lock_dir.default_path in
+      let lock_dirs = Pkg_common.Lock_dirs_arg.of_path lock_dir_path in
+      let* per_contexts =
+        Workspace.workspace ()
+        >>| Pkg_common.Lock_dirs_arg.lock_dirs_of_workspace lock_dirs
+      in
+      let lock_dirs =
+        List.filter_map per_contexts ~f:(fun lock_dir_path ->
+          match Path.exists (Path.source lock_dir_path) with
+          | true -> Some (Dune_pkg.Lock_dir.read_disk_exn lock_dir_path)
+          | false -> None)
+      in
+      match lock_dirs with
+      | [] -> Memo.return ()
+      | lock_dirs ->
+        let* local_packages = Pkg_common.find_local_packages in
+        let f lock_dir =
           match Dune_pkg.Package_universe.up_to_date local_packages lock_dir with
           | `Valid -> Memo.return ()
           | `Invalid _ ->
             let hints = Pp.[ text "run dune pkg lock" ] in
-            User_error.raise ~hints [ Pp.text "The lock dir is not sync with your dune-project" ])
-      in
-      let+ (_ : unit list) = Memo.all_concurrently locks in
-      ())
+            User_error.raise
+              ~hints
+              [ Pp.text "The lock dir is not sync with your dune-project" ]
+        in
+        Memo.parallel_iter ~f lock_dirs)
 ;;
 
 let run_build_system ~common ~request =
@@ -77,7 +86,9 @@ let run_build_system ~common ~request =
       Cached_digest.invalidate_cached_timestamps ();
       let* setup = Import.Main.setup () in
       let setup_with_check =
-        Memo.bind (raise_on_lock_dir_out_of_sync ()) ~f:(fun () -> setup)
+        Memo.bind
+          (Memo.Lazy.force (raise_on_lock_dir_out_of_sync @@ Common.profile common))
+          ~f:(fun () -> setup)
       in
       let request =
         Action_builder.bind (Action_builder.of_memo setup_with_check) ~f:(fun setup ->
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -30,6 +30,36 @@ let with_metrics ~common f =
     Fiber.return ())
 ;;
 
+let raise_on_lock_dir_out_of_sync () =
+  Memo.of_thunk (fun () ->
+    let open Memo.O in
+    let lock_dir_path = Dune_pkg.Lock_dir.default_path in
+    let lock_dirs = Pkg_common.Lock_dirs_arg.of_path lock_dir_path in
+    let* per_contexts =
+      Workspace.workspace () >>| Pkg_common.Lock_dirs_arg.lock_dirs_of_workspace lock_dirs
+    in
+    let lock_dirs =
+      List.filter_map per_contexts ~f:(fun lock_dir_path ->
+        match Path.exists (Path.source lock_dir_path) with
+        | true -> Some (Dune_pkg.Lock_dir.read_disk lock_dir_path)
+        | false -> None)
+    in
+    match lock_dirs with
+    | [] -> Memo.return ()
+    | lock_dirs ->
+      let* local_packages = Pkg_common.find_local_packages in
+      let locks =
+        List.map lock_dirs ~f:(fun lock_dir ->
+          match Dune_pkg.Package_universe.up_to_date local_packages lock_dir with
+          | `Valid -> Memo.return ()
+          | `Invalid _ ->
+            let hints = Pp.[ text "run dune pkg lock" ] in
+            User_error.raise ~hints [ Pp.text "The lock dir is not sync with your dune-project" ])
+      in
+      let+ (_ : unit list) = Memo.all_concurrently locks in
+      ())
+;;
+
 let run_build_system ~common ~request =
   let run ~(toplevel : unit Memo.Lazy.t) =
     with_metrics ~common (fun () -> build (fun () -> Memo.Lazy.force toplevel))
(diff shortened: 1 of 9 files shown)


bin/build_cmd.ml
```diff
@@ -30,6 +30,38 @@ let with_metrics ~common f =
     Fiber.return ())
 ;;
 
+let raise_on_lock_dir_out_of_sync () =
+  Memo.of_thunk (fun () ->
```
rgrinberg, line 34: This one needs to be `Memo.lazy_` to avoid recomputation I believe

maiste: I don't have enough knowledge on `Memo`: what is the difference from `Memo.lazy_` and `Memo.thunk` (apart the fact that the first one wrap the result in a Lazy structure)? From my previous understanding, the role of `Memo` was to ensure there isn't some "action duplication". What do I miss here? :confused:

rgrinberg: The difference is as how bind behaves on them. When you bind on `Memo.of_thunk f`, `f` will always be recomputed on demand. While if you bind on `Memo.Lazy.force (Memo.lazy_ f))`, you will only recompute `f` when necessary.

maiste: Thanks for the explanation :pray:

bin/build_cmd.ml
```diff
@@ -30,6 +30,38 @@ let with_metrics ~common f =
...
+        List.map lock_dirs ~f:(fun lock_dir ->
+          match Dune_pkg.Package_universe.up_to_date local_packages lock_dir with
+          | `Valid -> Memo.return ()
+          | `Invalid _ ->
+            let hints = Pp.[ text "run dune pkg lock" ] in
+            User_error.raise
+              ~hints
+              [ Pp.text "The lock dir is not sync with your dune-project" ])
+      in
+      let+ (_ : unit list) = Memo.all_concurrently locks in
```
rgrinberg, line 61: I think you can just use `Memo.parallel_iter`.

rgrinberg (changes requested):
There's some issues with the tests. We don't want to invoke the solver when we're just testing the specification of the language for specifying dune packages.

I guess these tests no longer work b/c they're failing your newly added check? I wonder why is that, we've already had this check before - it just didn't work in watch mode.

maiste:
I can exclude the checks when we are in the testing context. Do you think it's better to introduce an environment variable, or using the `INSIDE_DUNE` variable is enough? 

The only test I see as failed in the CI is the Nix one, not the tests related to this PR. Which test(s) are you referring to? Maybe just adding the guard would fix the issue.

test/blackbox-tests/test-cases/pkg/extra-source-overlap-with-source.t
```diff
@@ -2,20 +2,27 @@ Test for packages with an extra-source file with the same name as a
...
-  > (version 1)
-  > (source
-  >  (copy $PWD/foo-source))
-  > (extra_sources
-  >  (foo.txt
-  >   (fetch
-  >    (url file://$PWD/foo.txt))))
-  > EOF
 
+Create a fake package archive
```
rgrinberg, line 6: This change and others is what I'm referring to. We should be able to test how `extra_sources` works without invoking the solver

rgrinberg:
I'm first trying to understand why this check is only firing now. @gridbugs didn't we have such a check firing in non watch mode before?

> Do you think it's better to introduce an environment variable, or using the INSIDE_DUNE variable is enough?

How about we just disable the check when there's no hash in the lock directory? In such a case, we are fairly confident dune didn't generate the lock directory, so we can skip the check altogether.

maiste:
From my understanding, the PR that was supposed to do this wasn't merged (#10641) as it was doing the auto relooking, not only the warning. Indeed, this is a good idea: if the user generates the code by hand, it is their responsibility to ensure it work. I'll work on a new version of the code then.

rgrinberg:
There's a failure:

```
File "test/blackbox-tests/test-cases/pkg/ignore-lock-release.t", line 1, characters 0-0:
/usr/bin/git --no-pager diff --no-index --color=always -u _build/.sandbox/407077920e9f248353c[15](https://github.com/ocaml/dune/actions/runs/10722163469/job/29732440933?pr=10851#step:9:16)66b968f13a7/default/test/blackbox-tests/test-cases/pkg/ignore-lock-release.t _build/.sandbox/407077920e9f248353c1566b968f13a7/default/test/blackbox-tests/test-cases/pkg/ignore-lock-release.t.corrected
diff --git a/_build/.sandbox/407077920e9f248353c1566b968f13a7/default/test/blackbox-tests/test-cases/pkg/ignore-lock-release.t b/_build/.sandbox/407077920e9f248353c1566b968f13a7/default/test/blackbox-tests/test-cases/pkg/ignore-lock-release.t.corrected
index e2e7e09..d874cac 100644
--- a/_build/.sandbox/407077920e9f248353c1566b968f13a7/default/test/blackbox-tests/test-cases/pkg/ignore-lock-release.t
+++ b/_build/.sandbox/407077920e9f248353c1566b968f13a7/default/test/blackbox-tests/test-cases/pkg/ignore-lock-release.t.corrected
@@ -[18](https://github.com/ocaml/dune/actions/runs/10722163469/job/29732440933?pr=10851#step:9:19),3 +18,5 @@ When building a project in release mode we should ignore the lock directory.
   > EOF
 
   $ dune build @install --release
+  Error: Field "version" is missing
+  [1]
```
The error message is valid, but we shouldn't be parsing anything in the `--release` profile.

rgrinberg:
Unless there's already such a test, it seems like a good idea.

gridbugs:
> I'm first trying to understand why this check is only firing now. @gridbugs didn't we have such a check firing in non watch mode before?

The only such check I'm aware of is in a separate command `dune pkg validate-lockdir` (added in https://github.com/ocaml/dune/pull/9172). That command mostly just serves to check that the validation logic works in our tests, and if we're moving the validation into `dune build` we can probably get rid of `dune pkg validate-lockdir` altogether since I don't expect any user to run that command.

maiste:
As we discussed, I've made the following changes:
- Add a regression test for the locking message behavior.
- Update the code to prevent it to be run in two situations:
  - If the lock.dune does not have a hash. It means we are in the situation where the user manually touched or generated the file.
  - If the release profile is on, we don't want to run the checks, either.

rgrinberg:
Thanks! A few issues remain:

1. You're creating the lazy value for every iteration, which defeats the purpose of using lazy for saving recomputations.
2. If there's an error in one context, it shouldn't raise if you're building another context. Similarly, the profile is set per context. So skipping should also be checked per context.

To address both, I think it would be enough to just to use `Per_context` and create the check for every context. To make sure you aren't raising too eagerly, the check can be moved to `Gen_rules.gen_rules`. That way, it will be done only when loading the rules for the appropriate context.

maiste:
Thanks for the new review round. I have questions regarding your suggestions:
1. If I create the `unit Memo.Lazy.t` in `run_build_command` and pass it as an argument to the other functions, would it work to solve (1) ?
2. Is there a lock dir per context? Indeed, in the current implementation, we extract the _lock dir_ from the command line and/or the workspace.
   2.1. What does multiple contexts imply for the _lock dirs_? 
   2.2. I'm lost in terms of semantics here. In the workspace, if I have a context other than `default`, would I build both?

I'm asking all of this because moving the code to `Gen_rules.gen_rules` would imply to loose many helpers from `Pkg_common`. I want to be sure that I can mimic the semantic of the current check in `Gen_rules.gen_rules` where I don't have access to the _lock dir_ arg as I currently have in `build_cmd.ml`. It would mean that I need to get this value from somewhere else.

rgrinberg:
> If I create the unit Memo.Lazy.t in run_build_command and pass it as an argument to the other functions, would it work to solve (1) ?

Yes, but I don't think you should do that because the `Memo.Lazy.t` can just easily be created in the toplevel. See the other places where we access the profile (`Per_context.profile` for example).

> Is there a lock dir per context? Indeed, in the current implementation, we extract the lock dir from the command line and/or the workspace.

Every context may have its own lock directory. Lock directories are associated to contexts in the workspace file. It is contexts that are concerned with having a particular lock dir active.

>  What does multiple contexts imply for the lock dirs?

Nothing. Lock dirs are oblivious to contexts.

>  I'm lost in terms of semantics here. In the workspace, if I have a context other than default, would I build both?

Yes you would. For example, `dune build @install` will be the install alias in every single context defined in your workspace.

> I'm asking all of this because moving the code to Gen_rules.gen_rules would imply to lose many helpers from Pkg_common. I want to be sure that I can mimic the semantic of the current check in Gen_rules.gen_rules where I don't have access to the lock dir arg as I currently have in build_cmd.ml. It would mean that I need to get this value from somewhere else.

You can move things to dune_rules or dune_pkg as you wish to reuse code.

> Also, as the Gen_rules.gen_rules is directly used by the Dune_engine is it a good idea to introduce such logic at its core?

In this case, I think it's a good idea. As this is the simplest way to maintain context isolation with lock directories.

maiste:
Thanks for the clarifications :pray: Regarding your answer, I have few questions left:

> Yes, but I don't think you should do that because the Memo.Lazy.t can just easily be created in the toplevel. See the other places where we access the profile (Per_context.profile for example).

IIFC, I can create the lazy value at top level and `Lazy.force` the result when I'm using it `gen_rules`: it would do the re-computation when needed, right?

> Yes you would. For example, dune build @install will be the install alias in every single context defined in your workspace.

OK, by moving the check in `gen_rules`, I don't have to "deal with this" as the function would be triggered for every context. Am I correct here? :confused: Also, FMU, I only need to run the check when I'm not in `install`, `_private` or `fetch` but for every other context.

> Nothing. Lock dirs are oblivious to contexts.

I have spent time into the code to understand better how the different modules interacts with each others. However, I don't know how I can extract a lock dir from a context, as it seems not be part of the context type. If try to overcome this issue by matching `Context.t` with `Workspace.Context.t`.

 However, I have a pain point here. Indeed, I am able to retrieve the `Lock_dir_selection.t` from both, but I don't know how to evaluate it as a `Path.Source.t`. The blocking point is when I have a `Cond of Dune_lang.Cond.t`, I don't know how to evaluate it in the context of `gen_rules`. I have seen there is the `Lock_dir_selection.eval` function, but I don't know what I need to provide to `~f` of type `Value.t list Memo.t String_with_vars.expander `.  Maybe it is a false issue as in the context of `gen_rules`, this would be already evaluated. What would be your suggestion for this? 

> In this case, I think it's a good idea. As this is the simplest way to maintain context isolation with lock directories.
I understand why you suggest moving it to `Gen_rules.gen_rules` and it's far clearer about the why, thanks :smile: 

 I'll try to ship something as soon as possible. There might be more questions coming :sweat_smile:

rgrinberg:
> IIFC, I can create the lazy value at top level and Lazy.force the result when I'm using it gen_rules: it would do the re-computation when needed, right?

Yes. 

> OK, by moving the check in gen_rules, I don't have to "deal with this" as the function would be triggered for every context. Am I correct here? 😕 Also, FMU, I only need to run the check when I'm not in install, _private or fetch but for every other context.

fetch isn't context specific. Install or _private could also do the check, but there's no strong requirement to do so.

 > I have spent time into the code to understand better how the different modules interacts with each others. However, I don't know how I can extract a lock dir from a context, as it seems not be part of the context type. If try to overcome this issue by matching Context.t with Workspace.Context.t.

Have a look at how `pkg_rules.ml` generates package rules for a particular context. To generate those rules, the associated lock directory for the context must be looked up.

> However, I have a pain point here. Indeed, I am able to retrieve the Lock_dir_selection.t from both, but I don't know how to evaluate it as a Path.Source.t. The blocking point is when I have a Cond of Dune_lang.Cond.t, I don't know how to evaluate it in the context of gen_rules. I have seen there is the Lock_dir_selection.eval function, but I don't know what I need to provide to ~f of type Value.t list Memo.t String_with_vars.expander . Maybe it is a false issue as in the context of gen_rules, this would be already evaluated. What would be your suggestion for this?

See above. `pkg_rules` obtains the lock dir for a context. You can just reuse the code (`Lock_dir.get`)

maiste:
OK, first thanks for all the explanations :pray: 
I have moved the code to `Gen_rules.gen_rules` as you suggested. As I can extract the lock dir from the context directly is far less code, which is great! One last blocker I encounter is that I can't define the `Memo.lazy_` at top level as it requires the `ctx` to build the function. I'm wondering if there is a way to cache this efficiently: is there a data structure I can use?

rgrinberg:
> One last blocker I encounter is that I can't define the Memo.lazy_ at toplevel as it requires the ctx to build the function. I'm wondering if there is a way to cache this efficiently: is there a data structure I can use?

You can use `Per_context.create_by_name`. It handles the low level details of this for you.

maiste:
Thanks for the pointer. I have updated the code. It's ready for another review round :+1: 

Should I add a CHANGELOG entry for this?

rgrinberg:
There's no need. Package management is private thus far. We'll start recording changes officially once we have a stable release.

maiste:
Ok, thanks! I have rebased the branch on top of `main`. The CI errors are not related to this PR :+1:

PR #10852, opened 2024-08-27
feat: update prepare function for atomic lock

maiste:
This PR is a proposition for updating the `dune.lock` directory in an atomic way. Its goal is to prevent the user of unwanted behavior while running the `dune build --watch` command. It consists of 4 steps:
1. Rename the `dune.lock` into `.dune.lock` (in case of failure, it can still be restored)
2. Create a temporary directory and solve the new constraints
3. Rename the temporary directory to `dune.lock`
4. Erase the `.dune.lock` (as it has been replaced)

This is the first piece of the solution for  #10842.

### DIFF
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -498,6 +498,32 @@ module Write_disk = struct
     | Ok _ -> Error `Not_directory
   ;;
 
+  let raise_user_error_on_check_existance path e =
+    let error_reason_pp =
+      match e with
+      | `Unreadable -> Pp.text "Unable to read lock directory"
+      | `Not_directory -> Pp.text "Specified lock dir path is not a directory"
+      | `No_metadata_file ->
+        Pp.textf "Specified lock dir lacks metadata file (%s)" metadata_filename
+      | `Failed_to_parse_metadata (path, exn) ->
+        Pp.concat
+          ~sep:Pp.cut
+          [ Pp.textf
+              "Unable to parse lock directory metadata file (%s):"
+              (Path.to_string_maybe_quoted path)
+            |> Pp.hovbox
+          ; Exn.pp exn |> Pp.hovbox
+          ]
+        |> Pp.vbox
+    in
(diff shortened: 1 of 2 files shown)


src/dune_pkg/lock_dir.ml
```diff
@@ -555,15 +597,24 @@ module Write_disk = struct
...
             match original with
             | Path src -> Io.copy_file ~src ~dst ()
-            | Content content -> Io.write_file dst content)))
+            | Content content -> Io.write_file dst content)));
+      safely_rename_lock_dir_thunk ~dst:lock_dir_path_external lock_dir_path ();
+      remove_hidden_dir_if_exists ()
+    in
+    fun () ->
+      Temp.with_temp_dir
+        ~parent_dir:(Path.source Path.Source.root)
```
rgrinberg, line 614: I think the directory should be derived from the directory where the lock directory is being placed rather than hardcoded to be the source root.

src/dune_pkg/lock_dir.ml
```diff
@@ -498,6 +498,36 @@ module Write_disk = struct
     | Ok _ -> Error `Not_directory
   ;;
 
+  let raise_user_error_on_check_existance path e =
+    let error_reason_pp =
```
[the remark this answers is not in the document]
maiste: As I just moved the function, I imagine that it was historically because of the `module Pp`. I changed it :+1:

src/dune_pkg/lock_dir.ml
maiste: I have updated the code following your suggestion. It puts the temporary directory in the lock dir parent. In most case it will be `.` as the lock dir is `./dune.lock`

rgrinberg (approved):

Leonidas-from-XIV (approved):
Looks good. I would've probably renamed the lockdir to a hidden temporary name to avoid clashing with potentially existing hidden lock dirs that weren't cleaned up properly but there's pros and cons to each approach.

PR #10853, opened 2024-08-28
Typo in preprocessing-spec.rst

ttamttam:
(preprocessing) → (preprocess)


Leonidas-from-XIV (approved):
Thanks, this is indeed more correct.

PR #10858, opened 2024-08-29
feat: create an alias from pkg to package

maiste:
From the interviews conducted by @leostera it appears that, intuitively, people tend to write `package` instead of `pkg`. To respond to this, I have added a `dune package` command, with the same sub commands as `dune pkg` and a test to make sure we keep track of both. 

It doesn't seem that _Cmdliner_ provides a way to do aliases, so I had to introduce a new command.

### DIFF
--- a/bin/main.ml
+++ b/bin/main.ml
@@ -38,6 +38,7 @@ let all : _ Cmdliner.Cmd.t list =
     ; Init.group
     ; Promotion.group
     ; Pkg.group
+    ; Pkg.Alias.group
     ]
   in
   terms @ groups
--- a/bin/pkg/pkg.ml
+++ b/bin/pkg/pkg.ml
@@ -1,22 +1,23 @@
 open Import
 
-let info =
-  let doc = "Experimental package management" in
-  let man =
-    [ `S "DESCRIPTION"
-    ; `P {|Commands for doing package management with dune|}
-    ; `Blocks Common.help_secs
-    ]
-  in
-  Cmd.info "pkg" ~doc ~man
+let man =
+  [ `S "DESCRIPTION"
+  ; `P {|Commands for doing package management with dune|}
+  ; `Blocks Common.help_secs
+  ]
 ;;
 
-let group =
-  Cmd.group
-    info
-    [ Lock.command
-    ; Print_solver_env.command
-    ; Outdated.command
-    ; Validate_lock_dir.command
-    ]
+let subcommands =
+  [ Lock.command; Print_solver_env.command; Outdated.command; Validate_lock_dir.command ]
 ;;
+
+let info name =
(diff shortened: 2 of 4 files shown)


christinerose (approved):
Just added monospace around `pkg` command in last file, although since it's a `.t` file, I'm not sure I did it correctly.

test/blackbox-tests/test-cases/pkg/package-command-exists.t
```diff
@@ -0,0 +1,4 @@
+The package command is an alias for the pkg command. The test ensures it breaks if
+the package command is not available.
```
rgrinberg, line 2: Is there really a point to this? If you remove the command, `doc/dune.inc` will already get updated.

maiste: I didn't know if it would be enough to detect a regression. If you think this is enough, I can definitely remove this test :+1:

rgrinberg: It should be enough. If you remove any command, the doc file gets updated. So yeah, I think you can safely drop this test.

maiste: I have reverted the change.

bin/main.ml
```diff
@@ -38,6 +38,7 @@ let all : _ Cmdliner.Cmd.t list =
     ; Init.group
     ; Promotion.group
     ; Pkg.group
+    ; Pkg.Alias.group
```
Leonidas-from-XIV, line 41: I personally would rather create a `Package` module and define the commands there (exposing `subcommands` and `info`) but I can see the point in not cluttering the public interface.

bin/pkg/pkg.ml
```diff
@@ -1,22 +1,23 @@
...
-  let man =
-    [ `S "DESCRIPTION"
-    ; `P {|Commands for doing package management with dune|}
-    ; `Blocks Common.help_secs
-    ]
-  in
-  Cmd.info "pkg" ~doc ~man
+let man =
+  [ `S "DESCRIPTION"
+  ; `P {|Commands for doing package management with dune|}
```
Leonidas-from-XIV, line 5: Maybe "Commands for OCaml package management with Dune"

bin/main.ml
maiste: This idea is to not duplicate the code and one of them is going to disappear at some point ^^

bin/pkg/pkg.ml
maiste: I just extract the code, but I can modify that :wink:

christinerose (approved):

Issue #10856, opened 2024-08-28
Build coverage fails on cstruct


ElectreAAS:
I've tried to build several highly depended-upon packages with the new package management features to improve coverage, and here's an interesting one:

Note: according to check.ci.ocaml.org, `cstruct.6.2.0` has 4340 reverse dependencies, which is almost the maximum.

```
---> Building package n°12: package "cstruct.6.2.0":
Successfully extracted to /tmp/pkg-coverage-9df431-builder/cstruct.6.2.0
Solution for dune.lock:
- afl-persistent.1.4
- alcotest.1.8.0
- astring.0.8.5
- async.v0.16.0
- async_kernel.v0.16.0
- async_rpc_kernel.v0.16.0
- async_unix.v0.16.0
- base.v0.16.3
- base-bigarray.base
- base-bytes.base
- base-threads.base
- base-unix.base
- base_bigstring.v0.16.0
- base_quickcheck.v0.16.0
- bin_prot.v0.16.0
- cmdliner.1.3.0
- conf-bash.1
- core.v0.16.2
- core_kernel.v0.16.0
- core_unix.v0.16.0
- cppo.1.6.9
- crowbar.0.2.1
- csexp.1.5.2
- dune-configurator.3.16.0
- expect_test_helpers_core.v0.16.0
- fieldslib.v0.16.0
- fmt.0.9.0
- host-arch-x86_64.1
- host-system-other.1
- int_repr.v0.16.0
- jane-street-headers.v0.16.0
- jst-config.v0.16.0
- lwt.5.7.0
- lwt_ppx.2.1.0
- num.1.5-1
- ocaml.4.14.2
- ocaml-base-compiler.4.14.2
- ocaml-compiler-libs.v0.12.4
- ocaml-config.2
- ocaml-migrate-parsetree.2.4.0
- ocaml-options-vanilla.1
- ocaml-syntax-shims.1.0.0
- ocaml_intrinsics.v0.16.1
- ocamlbuild.0.15.0+dune
- ocamlfind.1.9.6+dune
- ocplib-endian.1.2
- ounit.2.2.7
- ounit2.2.2.7
- parsexp.v0.16.0
- ppx_assert.v0.16.0
- ppx_base.v0.16.0
- ppx_bench.v0.16.0
- ppx_bin_prot.v0.16.0
- ppx_cold.v0.16.0
- ppx_compare.v0.16.0
- ppx_custom_printf.v0.16.0
- ppx_derivers.1.2.1
[...]

PR #10859, opened 2024-08-29
Repro: patch doesn't like timestamps

ElectreAAS:
Reproduction test for the bug found in #10856.
We assume the patch format is that of `git diff`, but it happens (like in our case on opam) that the GNU diff tool is used instead.
This produces a timestamp on the same line as the filename, which isn't caught.

### DIFF
--- a/test/expect-tests/dune_patch/dune_patch_tests.ml
+++ b/test/expect-tests/dune_patch/dune_patch_tests.ml
@@ -161,13 +161,13 @@ let%expect_test "patching a deleted file" =
 ;;
 
 let%expect_test "Using a patch from 'diff' with a timestamp" =
-  test [ "foo.ml", "This is wrong\n" ] ("foo.patch", unified);
-  check "foo.ml";
-  [%expect.unreachable]
-[@@expect.uncaught_exn {|
-  (Dune_util__Report_error.Already_reported)
-  Trailing output
-  ---------------
-  Error: /tmp/build_b1b383_dune/dune_71088f_patch_test/foo.ml	2024-08-29
-  17:38:00.243088256 +0200: No such file or directory |}]
+  try
+    test [ "foo.ml", "This is wrong\n" ] ("foo.patch", unified);
+    check "foo.ml";
+    [%expect.unreachable]
+  with
+  | Dune_util.Report_error.Already_reported ->
+    if String.ends_with ~suffix:"No such file or directory\n" [%expect.output]
+    then [%expect ""]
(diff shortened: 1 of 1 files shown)


test/expect-tests/dune_patch/dune_patch_tests.ml
```diff
@@ -147,3 +159,15 @@ let%expect_test "patching a deleted file" =
...
+
+let%expect_test "Using a patch from 'diff' with a timestamp" =
+  test [ "foo.ml", "This is wrong\n" ] ("foo.patch", unified);
+  check "foo.ml";
+  [%expect.unreachable]
+[@@expect.uncaught_exn {|
+  (Dune_util__Report_error.Already_reported)
+  Trailing output
+  ---------------
+  Error: /tmp/build_b1b383_dune/dune_71088f_patch_test/foo.ml	2024-08-29
```
rgrinberg, line 171: We don't want to leak this temp path in the test

Leonidas-from-XIV: Maybe this could be better tested with a cram test, then it is easier to post-process the output?

rgrinberg: You can post process output in expect tests as well. See `[%output]`.

ElectreAAS: Should be cleaner now

test/expect-tests/dune_patch/dune_patch_tests.ml
```diff
@@ -147,3 +159,17 @@ let%expect_test "patching a deleted file" =
...
 ;;
+
+let%expect_test "Using a patch from 'diff' with a timestamp" =
+  try
+    test [ "foo.ml", "This is wrong\n" ] ("foo.patch", unified);
+    check "foo.ml";
+    [%expect.unreachable]
+  with
+  | Dune_util.Report_error.Already_reported ->
+    String.drop [%expect.output] 53 |> print_endline;
```
Leonidas-from-XIV, line 170: I don't think it is good to hardcode such a constant, given the temporary path locations aren't necessarily fixed length. Better use something like `String.lsplit2_exn` and `Filename.basename` to normalize the path.

ElectreAAS: True, I just didn't want to spend too much time on this since it's going to be removed by the fix anyway, but having a green CI is always nice

PR #10865, opened 2024-08-30
Better patch format parsing

ElectreAAS:
## What
Fixes a bug found in #10856, with repro pending in #10859:
A fair amount of packages fail to build when there was a patch made to their entry in the opam-repo, and that patch didn't follow a specific format.

## How
The regexp matching a filename was a simple `.+` (or `rep1 notnl` in Re terms). This has the unfortunate side effect of capturing the entire rest of the line, including a potential timestamp attached to the filename.
The new regexp is a still simple `[^\s]+` (or `rep1 (compl [space])`), which will stop when encountering whitespace. The rest of the line is caught and not used.
This has the unfortunate side effect of failing on filenames containing spaces.

I think this tradeoff is a good one for this scenario, since we expect patches to be applied to normal dev-created files, not adversarial edge-case crafters.
We could absolutely try to cover everything here, but the range of different formats that GNU patch can emit, along with the fact that filenames can contain almost anything, makes that a lot more effort.
Maybe we still should though?

### DIFF
--- a/src/dune_patch/dune_patch.ml
+++ b/src/dune_patch/dune_patch.ml
@@ -15,18 +15,15 @@ include struct
 end
 
 let re =
-  let line xs = Re.seq ((Re.bol :: xs) @ [ Re.eol ]) in
-  let followed_by_line xs = Re.seq [ Re.str "\n"; line xs ] in
-  Re.compile
-  @@ Re.seq
-       [ Re.alt
-           [ line [ Re.str {|--- a/|}; Re.group (Re.rep1 Re.notnl) ]
-           ; line [ Re.str {|--- |}; Re.group (Re.rep1 Re.notnl) ]
-           ]
-       ; Re.alt
-           [ followed_by_line [ Re.str {|+++ b/|}; Re.group (Re.rep1 Re.notnl) ]
-           ; followed_by_line [ Re.str {|+++ |}; Re.group (Re.rep1 Re.notnl) ]
-           ]
+  let open Re in
+  let line xs = seq ((bol :: xs) @ [ eol ]) in
+  let filename = group @@ rep1 @@ compl [ space ] in
+  let timestamp = rep notnl in
+  compile
+  @@ seq
+       [ line [ str {|--- |}; opt (str "a/"); filename; timestamp ]
+       ; str "\n"
+       ; line [ str {|+++ |}; opt (str "b/"); filename; timestamp ]
        ]
 ;;
 
@@ -44,25 +41,22 @@ let patches_of_string patch_string =
   Re.all re patch_string
   |> List.filter_map ~f:(fun group ->
     let open Option.O in
-    (* If these are [Some] then they are [/dev/null] *)
-    let old_file_dev_null = Re.Group.get_opt group 2 in
-    let new_file_dev_null = Re.Group.get_opt group 4 in
-    match old_file_dev_null, new_file_dev_null with
-    | Some _dev_null, Some _ ->
+    (* A match failure means a file name couldn't be parsed. *)
+    let* old_file = Re.Group.get_opt group 1 in
+    let* new_file = Re.Group.get_opt group 2 in
+    match old_file = "/dev/null", new_file = "/dev/null" with
+    | true, true ->
       (* when both files are /dev/null we don't care about the patch. *)
       None
-    | Some _, None ->
+    | true, false ->
       (* New file *)
-      let+ new_file = Re.Group.get_opt group 3 in
-      Patch.New (Path.Local.of_string new_file)
-    | None, Some _ ->
+      Some (Patch.New (Path.Local.of_string new_file))
+    | false, true ->
       (* Delete file *)
-      let+ new_file = Re.Group.get_opt group 1 in
-      Patch.Delete (Path.Local.of_string new_file)
-    | None, None ->
+      Some (Patch.Delete (Path.Local.of_string old_file))
+    | false, false ->
       (* Replace file *)
-      let+ new_file = Re.Group.get_opt group 3 in
-      Patch.Replace (Path.Local.of_string new_file))
+      Some (Patch.Replace (Path.Local.of_string new_file)))
 ;;
 
 let prog =
--- a/src/dune_patch/dune_patch.ml
+++ b/src/dune_patch/dune_patch.ml
@@ -15,16 +15,15 @@ include struct
 end
 
 let re =
-  let open Re in
-  let line xs = seq ((bol :: xs) @ [ eol ]) in
-  let filename = group @@ rep1 @@ compl [ space ] in
+  let line xs = Re.seq ((Re.bol :: xs) @ [ Re.eol ]) in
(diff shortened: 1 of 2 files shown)


src/dune_patch/dune_patch.ml
```diff
@@ -15,18 +15,15 @@ include struct
...
-           ; line [ Re.str {|--- |}; Re.group (Re.rep1 Re.notnl) ]
-           ]
-       ; Re.alt
-           [ followed_by_line [ Re.str {|+++ b/|}; Re.group (Re.rep1 Re.notnl) ]
-           ; followed_by_line [ Re.str {|+++ |}; Re.group (Re.rep1 Re.notnl) ]
-           ]
+  let open Re in
+  let line xs = seq ((bol :: xs) @ [ eol ]) in
+  let filename = group @@ rep1 @@ compl [ space ] in
+  let timestamp = rep notnl in
```
Leonidas-from-XIV, line 21: Can you convert this into matching a timestamp format or rename this into something that describes that this is just some leftovers data and we don't really care what it is?

In general it would be good to test this with paths that contain spaces. However that might be out of scope, given we don't support files with spaces anyway.

src/dune_patch/dune_patch.ml
```diff
@@ -15,18 +15,15 @@ include struct
...
+  compile
+  @@ seq
+       [ line [ str {|--- |}; opt (str "a/"); filename; timestamp ]
+       ; str "\n"
+       ; line [ str {|+++ |}; opt (str "b/"); filename; timestamp ]
```
Leonidas-from-XIV, line 26: This matches
```
--- a/foo
+++ foo
```
And I assume that the previous code would (rightly) reject this. Maybe worth keeping this behavior?

test/expect-tests/dune_patch/dune_patch_tests.ml
```diff
@@ -58,6 +58,18 @@ index ea988f6bd..000000000
 |}
 ;;
 
+(* Use GNU diff 'unified' format instead of 'git diff' *)
```
Leonidas-from-XIV, line 61: This should move to #10859 and then when it is rebased it should show that the code parses it properly now.

src/dune_patch/dune_patch.ml
ElectreAAS: The previous code would trip over this case pretty badly.
`+++ foo` would get interpreted as a dev/null, which means deleting the file `a/foo`.
Now it is understood as a name change, and `foo` will get updated with the contents of the patch

src/dune_patch/dune_patch.ml
ElectreAAS: > rename this into something that describes that this is just some leftovers data and we don't really care what it is?

Yeah that's a good point
> In general it would be good to test this with paths that contain spaces. However that might be out of scope, given we don't support files with spaces anyway.

Special characters in filenames are very finnicky and platform & tool dependent, I would much rather not chase a 100% full-compliance here.
I looked it up, and over at ocaml/opam-source-archives (the place they store the patches), there's over 300 patches, and not a single one has a file with a space in it.

ElectreAAS: > rename this into something that describes that this is just some leftovers data and we don't really care what it is?

Done!

src/dune_patch/dune_patch.ml
```diff
@@ -15,18 +15,16 @@ include struct
...
-  @@ Re.seq
-       [ Re.alt
-           [ line [ Re.str {|--- a/|}; Re.group (Re.rep1 Re.notnl) ]
-           ; line [ Re.str {|--- |}; Re.group (Re.rep1 Re.notnl) ]
-           ]
-       ; Re.alt
-           [ followed_by_line [ Re.str {|+++ b/|}; Re.group (Re.rep1 Re.notnl) ]
-           ; followed_by_line [ Re.str {|+++ |}; Re.group (Re.rep1 Re.notnl) ]
-           ]
+  let open Re in
```
rgrinberg, line 18: Can you remove the style changes? For example the local open, and the conversion of parens to `@@`. It makes it hard to understand the diff.

src/dune_patch/dune_patch.ml
```diff
@@ -44,25 +42,22 @@ let patches_of_string patch_string =
...
     let open Option.O in
-    (* If these are [Some] then they are [/dev/null] *)
-    let old_file_dev_null = Re.Group.get_opt group 2 in
-    let new_file_dev_null = Re.Group.get_opt group 4 in
-    match old_file_dev_null, new_file_dev_null with
-    | Some _dev_null, Some _ ->
+    (* A match failure means a file name couldn't be parsed. *)
+    let* old_file = Re.Group.get_opt group 1 in
+    let* new_file = Re.Group.get_opt group 2 in
+    match old_file = "/dev/null", new_file = "/dev/null" with
```
rgrinberg, line 48: Is this a change relevant to the bug fix?

ElectreAAS: They're part of fixing this bug and part of fixing a related bug found while reading the old code: any file not starting with "a/" is interpreted as "/dev/null"

I can make a repro PR if you want

rgrinberg:
@Alizter do you have a moment to glance over over this?

Alizter:
Why are we not using `git diff`? I seem to recall that we always assumed `git` was available which would make the patch format expected. I don't think I considered implementing a unified diff format because of this. (I don't know if there is a standard specification).

src/dune_patch/dune_patch.ml
Leonidas-from-XIV: If the file doesn't start with some prefix, be it `a/`/`b/` or some other prefix, we can't apply it anyway since we pass `-p1` to `patch` and `patch` will not apply it.

Thus I think the right thing would be to make sure that there is a prefix (or the file is `/dev/null`) or call `patch` with `-p0` if there was no prefix.

Alizter:
But if the objective is to tweak the parser so it can do:
```
---filename timestamp
```
then this seems fine to me. My previous question is really about the motivation which I don't fully understand.

rgrinberg:
How does `git diff` help? We don't need to generate diffs from the file tree - we need to read them.

Leonidas-from-XIV:
@Alizter The issue is due to #10856 where we get the patch from opam-repository ([this one](https://github.com/ocaml/opam-source-archives/blob/main/patches/afl-persistent/config.patch)) and it is patch in the unified diff format.

Alizter:
OK this LGTM then.

rgrinberg (changes requested):
As mentioned. Some unnecessary style changes snuck in.

The additional bug fix can stay, but we do need a test for it.

test/expect-tests/dune_patch/dune_patch_tests.ml
ElectreAAS: The repro PR has been merged, and this PR has been rebased on that.

rgrinberg (approved):

PR #10868, opened 2024-09-01
makefile: add watch rule

Khady:
It is convenient to have a quick way to compile in watch mode.


rgrinberg:
DCO is missing

Khady:
Sorry, I forgot about that. Should be fixed now.

PR #10871, opened 2024-09-02
fix: update printing of boot/libs.ml

emillon:
It has been changed in #10864.


rgrinberg (approved):

rgrinberg:
cc @tov some unnecessary parens were introduced

PR #10873, opened 2024-09-02
tests: Honor `DUNE_CONFIG__SKIP_LINE_BREAK`

Leonidas-from-XIV:
A follow up to #10589 and #10841 this PR attempts to unify the line-break handling a little bit.

At the moment this doesn't really skip the line-break but it sets it to the same value as it was before, to not break all tests but could be changed to use `Format.pp_infinity`, in which case a little bit of special casing could be removed.

### DIFF
--- a/src/dune_rules/merlin/merlin.ml
+++ b/src/dune_rules/merlin/merlin.ml
@@ -327,7 +327,15 @@ module Processed = struct
         |> Pp.concat_map ~sep:Pp.cut ~f:pp_one
         |> Pp.vbox
       in
-      Format.printf "%a%a@." Format.pp_set_margin 1000 Pp.to_fmt pp
+      let pp_format =
+        match Lazy.force Ansi_color.skip_line_break with
+        | false -> Pp.to_fmt
+        | true ->
+          fun ppf ->
+            Format.pp_set_margin ppf 1000;
+            Pp.to_fmt ppf
+      in
+      Format.printf "%a@." pp_format pp
   ;;
 
   let print_generic_dot_merlin paths =
--- a/test/blackbox-tests/test-cases/github2206.t/run.t
+++ b/test/blackbox-tests/test-cases/github2206.t/run.t
@@ -1,4 +1,7 @@
 copy_files would break the generation of the preprocessing flags
+
+  $ export DUNE_CONFIG__SKIP_LINE_BREAK=enabled
+
   $ dune build copy_files/.merlin-conf/exe-foo
   $ dune ocaml merlin dump-config $PWD/copy_files |
   > grep -B 1 -A 0 "pp"
--- a/otherlibs/stdune/src/ansi_color.mli
+++ b/otherlibs/stdune/src/ansi_color.mli
@@ -108,6 +108,9 @@ val stdout_supports_color : bool Lazy.t
 val stderr_supports_color : bool Lazy.t
 val output_is_a_tty : bool Lazy.t
 
+(* Whether to avoid breaking long lines *)
+val skip_line_break : bool Lazy.t
+
 (** Filter out escape sequences in a string *)
 val strip : string -> string
 
--- a/test/blackbox-tests/test-cases/melange/merlin-compile-flags.t
+++ b/test/blackbox-tests/test-cases/melange/merlin-compile-flags.t
@@ -1,5 +1,7 @@
 Show that the merlin config knows about melange.compile_flags
 
+  $ export DUNE_CONFIG__SKIP_LINE_BREAK=enabled
+
   $ cat >dune-project <<EOF
   > (lang dune 3.8)
   > (using melange 0.1)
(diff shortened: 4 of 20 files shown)


Leonidas-from-XIV:
I'm totally open to use `pp_infinity`, I just wanted to show first that the tests are reproducible if `DUNE_CONFIG__SKIP_LINE_BREAK` is set in all of those that require it.

test/blackbox-tests/test-cases/github2206.t/run.t
```diff
@@ -1,4 +1,7 @@
 copy_files would break the generation of the preprocessing flags
+
+  $ export DUNE_CONFIG__SKIP_LINE_BREAK=enabled
```
emillon, line 3: I didn't follow the original feature, but maybe it would be simpler to make that just an `INSIDE_DUNE` thing?

src/dune_rules/merlin/merlin.ml
```diff
@@ -327,7 +327,15 @@ module Processed = struct
         |> Pp.concat_map ~sep:Pp.cut ~f:pp_one
         |> Pp.vbox
       in
-      Format.printf "%a%a@." Format.pp_set_margin 1000 Pp.to_fmt pp
+      let fmt =
```
emillon, line 330: style remark: I'd expect `fmt` to be either a format string or a formatter (though `ppf` is preferred for that) but it's a formatting function applied to `pp`, so how about renaming it to `format_pp`?

emillon (approved):

test/blackbox-tests/test-cases/github2206.t/run.t
Leonidas-from-XIV: In the long run I would like to enable this by default (or at least configure the terminal width to be be margin with by default), I think Dune is much better with this option even outside of the tests.

src/dune_rules/merlin/merlin.ml
Leonidas-from-XIV: I know, normally this would be called `pp(_type)` but the conventions in `Pp` are a bit odd and the code calls the value to be printed `pp` already.

I started with renaming one the `pp` and the other one to `v` but then there was one place in the code where the conventions would be broken and it didn't seem worth it being inconsistent. I can name it `format_pp` to reduce confusion and convention-breaking.

test/blackbox-tests/test-cases/github2206.t/run.t
emillon: (sorry for the off-topic discussion)

I think it's a good idea to adapt to the size of the terminal, but this will require tinkering with SIGWINCH etc - I think that notty already handles that so it's possible that the terminal UI already works that way.

Exposing this to the user might be confusing so I'm wondering if there's a heuristic we can use. When will the user want to set this variable, and when won't they?

rgrinberg:
I don't really get the point of this PR from the description. Aren't the tests already reproducible as we can see from CI passing? Or are there some situations where they are failing?

test/blackbox-tests/test-cases/github2206.t/run.t
Leonidas-from-XIV: My thinking is that the user (a fancy way of saying "me") basically never wants wrapping to happen. Even if we wrap at the width of the terminal, we're wrapping at a place where the terminal would've wrapped anyway and we introduce a hard-newline. So if the user resizes the terminal afterwards, it can't rewrap the text anyway, there will always be the hard-newline. If OTOH we just leave it up to the terminal it works.

The only place where wrapping makes some sense IMHO is when we want to print some hierarchical structure with indentation and long lines and then the terminal softwrapping would break it. Even in these cases at least I personally prefer to disable line-wrapping in the terminal and have side-scrolling.

However, this might be a better setup for a dedicated RFC to explain the idea to a bigger audience and solicit more perspectives.

src/dune_rules/merlin/merlin.ml
```diff
@@ -327,7 +327,15 @@ module Processed = struct
         |> Pp.concat_map ~sep:Pp.cut ~f:pp_one
         |> Pp.vbox
       in
-      Format.printf "%a%a@." Format.pp_set_margin 1000 Pp.to_fmt pp
+      let pp_format =
+        match Lazy.force Ansi_color.skip_line_break with
```
rgrinberg, line 331: Don't think there's much point to this configuration. Isn't this function only used for tests? Might as well just make it produce the output the tests want unconditionally.

Leonidas-from-XIV:
@rgrinberg The original #10589 introduced the 1000 character wrapping to improve the tests (if I understand it correctly), but didn't make it an opt-in feature. Upon seeing #10841 @emillon commented that it would be nice to unify the solutions to have the Merlin dumping be controlled by the same variable, thus making the wrapping test-only.

rgrinberg:
Is there a use for writing merlin debug output with wrapping?

rgrinberg:
Here's a summary of this PR as far as I understand it:

1. Debug output disabled debug output to make the tests reproducible
2. Now debug output is only reproducible if a variable is set.
3. All new tests need to remember to set this environment variable to get reproducible debug output

Doesn't seem like an improvement to me. The variable was added to control wrapping for user visible output b/c clearly it can't be the default outside the tests there. What's the motivation here?

Leonidas-from-XIV:
The motivation for this PR was to make `DUNE_CONFIG__SKIP_LINE_WRAPPING` control line wrapping in more places in Dune, so it's less confusing that it doesn't seem to have an influence on Merlin debug line wrapping because that place was not configurable. So more a "improve consistency of the UI" kind of feature.

rgrinberg:
Okay well I hope it's obvious why this PR is making things more confusing. The least confusing behavior is just to do the right thing when we know what the right thing is. More confusing is to make this behavior configurable. Still more confusing is to make it configurable and set the wrong default.

test/blackbox-tests/test-cases/github2206.t/run.t
rgrinberg: That sounds useful. Another reason to customize this is that some people prefer a different margin anyway. From what I remember, 120 chars was asked before.

PR #10879, opened 2024-09-04
test: Spaces in filenames repro case

Leonidas-from-XIV:
Another edge case I could think of is spaces in filenames to be patched.

The `git diff` format just considers everything after the prefix (`a/`/`b/`) as filename so as long as the filename has no newlines it can be named anything. The unified diff format on the other hand will detect spaces and quote the filename with doublequotes (thus potentially creating another edge case but `foo"bar".ml` seems more like a pathological case to make tools fail rather than a file that would exist in a source tree).

This PR comes with a small utility function to postprocess the error output and display the error message including the path that also improves another test case.

### DIFF
--- a/test/expect-tests/dune_patch/dune_patch_tests.ml
+++ b/test/expect-tests/dune_patch/dune_patch_tests.ml
@@ -80,6 +80,30 @@ let no_prefix =
 |}
 ;;
 
+(* The file is called "foo bar" *)
+let spaces =
+  {|
+diff --git a/foo bar b/foo bar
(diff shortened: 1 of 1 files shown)


rgrinberg (approved):

ElectreAAS (approved):
Thanks for the utility functions :)

Leonidas-from-XIV:
I really wanted to avoid messing with the error message but I couldn't avoid removing the linebreaks in the error reason as well as the file name thus I just bit the bullet and removed all `\n`.

PR #10883, opened 2024-09-05
Always pass `-fdiagnostics-color=always` to `cc`-style compilers to avoid useless recompilations when Dune's stderr is redirected

### LINKED ISSUE #10882: Redirecting the stderr of Dune causes recompilation of C stubs
Dune decides whether to pass `-fdiagnostics-color` to the (GCC-style) C compiler or not depending on whether `stderr` is a terminal or not. This means that if one does `dune build 2>/dev/null` followed by `dune build`, the second command will rebuild all C stubs, which seems wrong.

One way to approach this may be to always make pass the `-fdiagnostics-color` flag to GCC, but filter out the ANSI codes in Dune, like it does in other cases (if I understand correctly).

nojb:
Fixes #10882

### DIFF
--- a/doc/changes/10883.md
+++ b/doc/changes/10883.md
@@ -1,0 +1,2 @@
+- Fix an issue where C stubs would be rebuilt whenever the stderr of Dune was
+  redirected. (#10883, @nojb)
--- a/src/dune_rules/cxx_flags.ml
+++ b/src/dune_rules/cxx_flags.ml
@@ -22,8 +22,7 @@ let base_cxx_flags ~for_ cc =
 ;;
 
 let fdiagnostics_color = function
-  | (Gcc | Clang) when Lazy.force Ansi_color.stderr_supports_color ->
-    [ "-fdiagnostics-color=always" ]
+  | Gcc | Clang -> [ "-fdiagnostics-color=always" ]
   | _ -> []
 ;;
 
--- a/test/blackbox-tests/test-cases/foreign-stubs/c-flags-diagnostics-color.t/run.t
+++ b/test/blackbox-tests/test-cases/foreign-stubs/c-flags-diagnostics-color.t/run.t
@@ -45,11 +45,7 @@ test that we correctly filter out the color codes
   >   (language c) (names stub)))
   > EOF
 
-  $ dune build
-  File "dune", line 4, characters 22-26:
-  4 |   (language c) (names stub)))
-                            ^^^^
-  stub.c:1:2: error: #error "error message"
-      1 | #error "error message"
-        |  ^~~~~
+We check there is no ESC [ sequence.
+
+  $ dune build 2>&1 | grep `printf '\033\\['`
   [1]
--- a/test/blackbox-tests/test-cases/foreign-stubs/c-flags-diagnostics-color.t/run.t
+++ b/test/blackbox-tests/test-cases/foreign-stubs/c-flags-diagnostics-color.t/run.t
@@ -1,9 +1,6 @@
 This test won't work with MSVC as it's using GCC and clang CLI
 parameters.
 
-We have to force Dune outputting colors, for that we use
-`CLICOLOR_FORCE=1`.
-
 The flag should be present in the command, as we want it in the
 :standard set.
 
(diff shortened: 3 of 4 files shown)


emillon:
cc @MisterDA do you remember why there's a guard here?

Leonidas-from-XIV:
@nojb Does your PR filter out the color sequences when writing to `_build/log`? I could see a point in always forcing color and filtering it out depending on where we write to (log, stdout) but we'd need to make sure it doesn't write color to logfiles.

nojb:
> @nojb Does your PR filter out the color sequences when writing to `_build/log`? I could see a point in always forcing color and filtering it out depending on where we write to (log, stdout) but we'd need to make sure it doesn't write color to logfiles.

In general Dune knows how to strip ANSI color sequences from the output of commands if Dune's output is itself not a terminal. This works if one redirects the output of Dune, eg `dune 2>...`. However, there seems to be something not quite right with `_build/log`:
```
$ (cd _build/default && /home/nojebar/mlfi/mlfi-ins/bin/ocamlopt.opt -w @1..3@5..28@30..39@43@46..47@49..57@61..62-40 -strict-sequence -strict-formats -short-paths -keep-locs -w -49 -nopervasives -nostdlib -g -I .lib.objs/byte -I .lib.objs/native -intf-suffix .ml-gen \
-no-alias-deps -opaque -o .lib.objs/native/lib.cmx -c -impl lib.ml-gen)
$ (cd _build/default && /usr/bin/gcc -O2 -fno-strict-aliasing -fwrapv -pthread -fPIC -D_FILE_OFFSET_BITS=64 -fdiagnostics-color=always -g -I /home/nojebar/mlfi/mlfi-ins/lib -o stubs.o -c stubs.c)
>
>     1 | #
>       |
```
but this is independent of this PR, the behaviour is the same on `main`.

MisterDA:
Why does Dune decide to rebuild the stubs? Is it because the set of flags changed that it invalidates the artifacts? If so, I think Dune should rather ignore this flag.

I think this is the wrong fix; the original idea was to pass the `-fdiagnostics-color=always` flag to the C compiler if and only if Dune decides it should _itself_ use colored output.
If Dune strips the ANSI color escape codes from the C compiler when Dune chooses _not_ to output colors, then we can pass `-fdiagnostics-color=always` unconditionally to the C compiler and I'm fine with this patch.

As a reminder, the end goal was to implement [Standard for ANSI Colors in Terminals](http://bixense.com/clicolors/) (quite informal, inspired by the BSD world). Tools should detect whether to output colors based on a CLI flags, an handful of environment variables, or if the output is a terminal (using [`isatty`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/isatty.html)). As the output of the tools Dune invokes is buffered through pipes, they don't output colored text by default. If they were to use a single, standardized environment variable, such as `CLICOLOR_FORCE`, Dune could simply set it for all its subprocesses. Unfortunately that's not the case, so we have to resort to passing flags.

nojb:
> > If Dune strips the ANSI color escape codes from the C compiler when Dune chooses _not_ to output colors, then we can pass `-fdiagnostics-color=always` unconditionally to the C compiler and I'm fine with this patch.
> 
> This is the behaviour Dune has always had, as far as I know, and also used with the OCaml compiler.

Also cc @rgrinberg which will know about this.

rgrinberg:
Always choosing to use colors and stripping the colors ourselves is the consistent thing to do.

nojb:
> Always choosing to use colors and stripping the colors ourselves is the consistent thing to do.

Thanks for the quick reply! Do you confirm that this is already how Dune works and there isn't anything more to do in this PR to handle colors coming from the C compiler correctly?

rgrinberg:
I confirm. Some testing would of course be appreciated though.

FYI, this might break `--no-buffer` in some ways, but I'm not particularly concerned as that mode is rather dangerous anyway.

nojb:
> I confirm. Some testing would of course be appreciated though.

Thanks. I tested locally as well as adapting some existing tests that check that the flag is passed correctly and that the color codes are filtered out when dune's stderr is not a terminal.

Planning to merge once CI passes.

PR #10884, opened 2024-09-05
perf: add names to source tree events

### LINKED ISSUE #10880: Performance: Trace file shows `Source tree scan` stage takes 38+ seconds
We continue trying to figure out in which ways we could optimize our Dune build setup at Ahrefs.

One of the things that quickly stands out is that when enabling `--trace-file`, we can see that the "Source tree scan" step of folder `.` takes more than 38s.

One of the instances of the `Source tree scan` events shows the folder `backend/api/src`, which takes ~37s. But after that event, there's only very short lived events which don't add much information.

Is there any way in which we could modify Dune to get more granular information about where the time is being spent? For example, if would be interesting to know if this time is spent parsing rules (of which `dune` files), executing commands that generate `sexp` files, etc. If those things is something that can happen during the `Source tree scan` stage, of course.

Please find below the trace file (obtained with `dune build @fe.all @aa --trace-file fe.all-aa.txt`) and a screenshot loading the trace in Brave (Chrome-like browser):

[dune-build-fe.all-aa-trace.txt](https://github.com/user-attachments/files/16868672/dune-build-fe.all-aa-trace.txt)

image

## Specifications

- Version of `dune`: `3.15.0`
- Version of `ocaml`: `4.14.0`
- Operating system (distribution and version): Debian 12, kernel 6.5.0

jchavarri:
Partially fixes #10880.

There is a trace event being registered in `map_reduce` function of `Source_tree`, but I noticed this function is used from different places, with different traversal functions and logic.

This PR:
- passes around a label to be used as the event name, so that viewers can notice distinctions between different traversals
- includes the dir being evaluated into the event name as well. It was already added to the `args` of the event, but having it upfront makes things easier when visualizing the traces too.

Here are some demo traces and screenshots, visualized in [Firefox profiler](https://profiler.firefox.com/) (which also support Chrome traces).

The build profiled goes through a lot of rules but none of them has to be built, except for a huge executable at the end (see the long `ocamlc.opt` event on the right).

Before ([trace file]((https://github.com/user-attachments/files/16894320/main-build.txt))):

image

After ([trace file](https://github.com/user-attachments/files/16894354/add-event-names-build.txt)):

image

I am a bit confused as to why "Alias builder" traversal is running twice, with the first time not including the large `ocamlc.opt` chunk, but the second time including it.

### DIFF
--- a/bin/target.ml
+++ b/bin/target.ml
@@ -53,6 +53,7 @@ let all_direct_targets dir =
     Source_tree_map_reduce.map_reduce
       root
       ~traverse:Source_dir_status.Set.all
+      ~trace_event_name:"All direct targets"
       ~f:(fun dir ->
         Dune_engine.Load_rules.load_dir
           ~dir:
--- a/doc/changes/10884.md
+++ b/doc/changes/10884.md
@@ -1,0 +1,1 @@
+- Add names to source tree events in performance traces (#10884, @jchavarri)
--- a/src/dune_rules/alias_rec.ml
+++ b/src/dune_rules/alias_rec.ml
@@ -70,6 +70,10 @@ include Alias_builder.Alias_rec (struct
       >>= function
       | None -> Action_builder.return Alias_builder.Alias_status.Not_defined
       | Some src_dir ->
-        Map_reduce.map_reduce src_dir ~traverse:Source_dir_status.Set.normal_only ~f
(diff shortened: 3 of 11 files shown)


rgrinberg:
Thanks, can you add a CHANGES entry?

PR #10885, opened 2024-09-05
Support patches without a prefix by determining the prefix depth

Leonidas-from-XIV:
This reads the prefix depth from the file. We currently only support only `1` but this adds support for `0` which the regex already supported.

For arbitrary depth prefixes this would require a more complex parsing regex but also these are presumably rather rare in the real world.

### DIFF
--- a/src/dune_patch/dune_patch.ml
+++ b/src/dune_patch/dune_patch.ml
@@ -22,8 +22,8 @@ let re =
   let junk = Re.rep Re.notnl in
   Re.compile
   @@ Re.seq
-       [ line [ Re.str {|--- |}; Re.opt (Re.str "a/"); filename; junk ]
-       ; followed_by_line [ Re.str {|+++ |}; Re.opt (Re.str "b/"); filename; junk ]
+       [ line [ Re.str {|--- |}; filename; junk ]
+       ; followed_by_line [ Re.str {|+++ |}; filename; junk ]
        ]
 ;;
 
@@ -31,10 +31,15 @@ module Patch = struct
   (* CR-someday alizter: more parsed information about the patch should go here.
      Eventually we wish to replace the patch command inside the patch action with a pure
      OCaml implementation. *)
-  type t =
+  type operation =
     | New of Path.Local.t
     | Delete of Path.Local.t
     | Replace of Path.Local.t
+
+  type t =
+    { prefix : int
+    ; op : operation
+    }
 end
 
 let patches_of_string patch_string =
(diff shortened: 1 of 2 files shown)


src/dune_patch/dune_patch.ml
```diff
@@ -22,41 +22,81 @@ let re =
...
 let patches_of_string patch_string =
   Re.all re patch_string
   |> List.filter_map ~f:(fun group ->
     let open Option.O in
     (* A match failure means a file name couldn't be parsed. *)
-    let* old_file = Re.Group.get_opt group 1 in
-    let* new_file = Re.Group.get_opt group 2 in
+    let* old_prefix = Re.Group.get_opt group 1 in
+    let* old_file = Re.Group.get_opt group 2 in
+    let* new_prefix = Re.Group.get_opt group 3 in
```
rgrinberg, line 67: Is a prefix always a relative path?

Leonidas-from-XIV: That's a good question and I never thought about it so tried what the tools do. `diff -u` will gladly generate a patch with an absolute path:

```diff
--- /tmp/diffy/old	2024-09-06 09:48:02.146670822 +0200
+++ /tmp/diffy/new	2024-09-06 09:48:10.369718133 +0200
@@ -1 +1 @@
-This is wrong.
+This is right.
```

However if I have the files at the locations and use `patch -p0` it will refuse to apply the patch:

```sh
$ patch -p0 < absolute.patch
Ignoring potentially dangerous file name /tmp/diffy/old
Ignoring potentially dangerous file name /tmp/diffy/new
can't find file to patch at input line 3
Perhaps you used the wrong -p or --strip option?
```

If I instead call it from /tmp/diffy and strip the 2 components it will still not do it correctly:

```sh
$ patch -p2 < absolute.patch
can't find file to patch at input line 3
Perhaps you used the wrong -p or --strip option?
```

However, if I turn the absolute path `/tmp/diffy/...` into a relative `tmp/diffy/...` it applies the patch correctly with `-p2` so I assume that `patch` can only handle relative paths.

In our case the only prefixes that the regex can ever match are `""`, `"a/"` and `"b/"` but the calculation function is a little bit more robust.

rgrinberg: Okay, then I think we can assume only a relative path is valid then. what about a trailing slash on the relative path, is that accepted? Also if trailing slashes are allowed, how is just `/` interpreted?

Reason why I'm asking is that I'm not sure using string operations on slashes is the robust approach here. I would much prefer to see things converted to `Path.Local.t` and operated on using path functions.

Leonidas-from-XIV: > Reason why I'm asking is that I'm not sure using string operations on slashes is the robust approach here. I would much prefer to see things converted to `Path.Local.t` and operated on using path functions.

Yes you're right, that is nicer. I've converted it to use `Path.Local` functions, there are some pretty useful ones to deal with this exact case. I've moved the handling of the prefix from the regex to the OCaml code, that way the prefix handling is in one place.

rgrinberg (approved):

ElectreAAS (approved):

PR #10888, opened 2024-09-05
refactor: expand ocamlformat rules without action ast

rgrinberg:
support for ocamlformat is hardcoded anyway, so just setup the rules with the extra expansion step for actions defined using the dune lang.

Extracted from #10647 


gridbugs (approved):
I fixed some typos in comments and rebased onto main.

PR #10890, opened 2024-09-06
Change colors to be more readable in light themes

gridbugs:
Some of the text colors used by dune tend not to be readable on default light-themes terminals. Of course it will always be possible to configure your terminal so that some colors are too light/dark to read, but some of the preset themes on macos are unable to display some colors commonly used by dune.

The three types of text in question are `Ok` seen below as the green program names in the output of `dune build --display=short`, `Details` seen below as the dim text on the right side of the output of `dune build --display=short`, and `Hint` seen below as the inverted colors or italic text displaying the word "Hint" in the output of `dune pkg validate`.

The light theme is MacOS Terminal's "Novel" theme, the dark theme is MacOS Terminal's "Basic" theme, and the purple themes are "Catpuccin Mocha" themes of WezTerm (top) and Emacs (bottom).

@leostera you've mentioned issues with being able to read the output of dune on light themes a couple of times. Can you share the terminals and themes where you run into this issue and confirm whether this change fixes it.

Before:
Screenshot 2024-09-06 at 19 11 06

After:
Screenshot 2024-09-06 at 19 13 36

### DIFF
--- a/doc/changes/10890.md
+++ b/doc/changes/10890.md
@@ -1,0 +1,2 @@
+- Change some colors to improve readability in light-mode terminals
+  (#10890, @gridbugs)
--- a/otherlibs/stdune/src/user_message.ml
+++ b/otherlibs/stdune/src/user_message.ml
@@ -159,9 +159,9 @@ module Print_config = struct
     | Kwd -> [ `Bold; `Fg_blue ]
     | Id -> [ `Bold; `Fg_yellow ]
     | Prompt -> [ `Bold; `Fg_green ]
-    | Hint -> [ `Italic; `Fg_white ]
-    | Details -> [ `Dim; `Fg_white ]
-    | Ok -> [ `Dim; `Fg_green ]
+    | Hint -> [ `Italic ]
+    | Details -> [ `Dim ]
+    | Ok -> [ `Fg_green ]
     | Debug -> [ `Underline; `Fg_bright_cyan ]
     | Success -> [ `Bold; `Fg_green ]
     | Ansi_styles l -> l


rgrinberg (approved):
No preference towards any particular colors. I do think we more empirical input from some more devs. The new scheme should work by default in a wide variety setups.

gridbugs:
Tried this out on a few more terminals. Here are the default themes for rxvt-unicode, xterm, emacs and vscode before and after this change. I suspect most people change their themes from the defaults, especially for the first 3 whose default theme is a white background, but I think it's important for dune to be useable in the default themes for these terminals regardless. In vscode (which defaults to a dark theme) dune is readable before and after this change, but I included it to demonstrate that I didn't break readability in this case.

Before:
dune-old1

After:
dune-new1

gridbugs:
Here's how it looked before and after on the default terminal theme in ubuntu. Again it was fine before this change, but just demonstrating that it remains fine after this change:

Before:
dune-old

After:
dune-new

maiste (approved):
It is displayed nicely on my side with `foot` term with both `Catppuccin-mocha` and `Catppuccin-latte` :+1:

Nice colors! :smile:

rgrinberg:
@nojb could you try this on windows?

nojb:
> @nojb could you try this on windows?

Sure; a bit busy at the moment, but should be able to do it in the next day or so.

nojb:
> > @nojb could you try this on windows?
> 
> Sure; a bit busy at the moment, but should be able to do it in the next day or so.

I tried with several of the built-in themes of Windows Terminal; seems fine to me:

image

image

image

image

image

image

PR #10891, opened 2024-09-06
Perf: Emit trace events from `load_build_directory_exn`

jchavarri:
This is an attempt to get more granular information on traces about the rule loading process. It adds events for all the steps performed in `load_build_directory_exn`.

Regular profiles look contain more information:

image

I must be doing something wrong though, because when one looks at the results using [Firefox Profile](https://profiler.firefox.com/) (which doesn't collapse events), the durations of similar events happening in separate build folders always take the same time. Maybe it is related to memoization?

This is visible in the animation below, here is the trace file used for it:

[dune_trace_load_rules.json](https://github.com/user-attachments/files/16906081/dune_trace_load_rules.json)

firefox-profile

### DIFF
--- a/src/dune_engine/load_rules.ml
+++ b/src/dune_engine/load_rules.ml
@@ -790,83 +790,162 @@ end = struct
         ])
   ;;
 
+  let write_event ~start ~dir name =
+    let () =
+      match Dune_stats.global () with
+      | None -> ()
+      | Some stats ->
+        let stop = Unix.gettimeofday () in
+        let module Event = Chrome_trace.Event in
+        let module Timestamp = Event.Timestamp in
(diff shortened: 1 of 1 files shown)


jchavarri:
Easier reviewing [with whitespace enabled](https://github.com/ocaml/dune/pull/10891/files?diff=unified&w=1).

rgrinberg:
Try moving this instrumentation to `dune_rules/gen_rules` for now. This will give you a more accurate idea on how much work is to done to actually generate the rules.

The way you're doing it now is recording a lot of overhead from the engine itself. While that's useful, you will need to separate it from the rule callback later anyway, and some of the events you've added aren't going to make sense once we've upstreamed some of our changes.

Issue #3516, opened 2020-05-30
dune format-dune-file doesn't respect the formatting stanza


Khady:
In a project with a dune-project that says `(formatting (enabled_for ocaml reason))`, running `dune format-dune-file` will still format the dune file despite the formatting stanza saying that it shouldn't happen.

The problem mostly appears when using ocamllsp as `dune format-dune-file` is a low level command that one usually doesn't call directly.

dune 2.5

jeremiedimino:
/cc @rgrinberg @emillon 

My understanding is that `dune format-dune-file` is indeed a low-level command that doesn't read the `dune-project` file. Though it seems like it should given that the formatting of `dune` files is versioned as per [this discussion](https://github.com/ocaml/dune/wiki/dev-meeting-20200429#dune-fmt).

rgrinberg:
Yeah, it’s indeed intended to be a low level command. The problem is that we don’t have a suitable high level way to format dune files. I suppose this could be added as a use case in the RPC issue

jeremiedimino:
If we start doing the formatting in-process, do we still need `dune format-dune-file` to be low-level?

emillon:
To confirm, yes, `dune format-dune-file` is supposed to be a low level command and always format its input. A high level frontend to `@fmt` to check (or fix) formatting of all (or some) files based on the `dune-project` configuration would probably be enough to fix this problem.

jeremiedimino:
We should probably compare what dune does to what ocamlformat does. My understand is that right now, the editor calls `ocamlformat <file>` and this command will read the ocamlformat configuration files to decide whether and how to format the file. Is that correct?

jberdine:
> My understand is that right now, the editor calls ocamlformat <file> and this command will read the ocamlformat configuration files to decide whether and how to format the file. Is that correct?

Yes, with the caveat that (at least the emacs integration) dumps the possibly-unsaved contents of the editing buffer to a temporary file and passes ocamlformat a -name flag to tell it where to look for configuration files.

jeremiedimino:
Do we think this mode of operation is the right one? From a distance, it seems good to me.

If we think it is, then we should add a similar feature in Dune. i.e. add a dune command to format a dune file taking into account the dune configuration files, and taking a similar `-name` flag.

[...]

PR #10892, opened 2024-09-06
Tweak dune file formatting: avoid pure vertical layout when formatting wrapped lists

### LINKED ISSUE #10860: auto formatting of `run` shouldn't split its content into one word per line
## Desired Behavior

When using `run` in a rule stanza we can have long command with many arguments which quickly end up being split into one word per line. It would tend to look better if `run` could keep everything on a single line as much as possible. And when it splits things into multiple lines it would put as much content as possible on every line.

## Example

The current

```
(rule
 (target sql_file1.ml)
 (deps
  gen_file2.sql
  gen_file3.sql)
 (mode promote)
 (action
  (with-stdout-to
   %{target}
   (run
    sqlgg
    -static-header
    -gen
    caml_io
    -name
    Make
    -params
    unnamed
    -category
    none
    gen_file2.sql
    -category
    all
    gen_file3.sql))))
```

could look more like this

```lisp
(rule
 (target sql_file1.ml)
 (deps 
  gen_file2.sql
  gen_file3s.sql)
 (mode promote)
 (action
  (with-stdout-to
   %{target}
   (run  sqlgg -static-header -gen caml_io -name Make -params unnamed -category none gen_file2.sql -category all gen_file3.sql))))
```

or if the line is longer than the limit, it could look like this
```
   (run  sqlgg -static-header -gen caml_io -name Make -params unnamed -category none 
            gen_file2.sql -category all gen_file3.sql))))
```

nojb:
Fixes #10860

The effect of the change is illustrated in the tests.

cc @Khady

### DIFF
--- a/doc/changes/10892.md
+++ b/doc/changes/10892.md
@@ -1,0 +1,2 @@
+- Format long lists in s-expressions to fill the line instead of formatting
+  them in a vertical way (#10892, fixes #10860, @nojb)
--- a/src/dune_lang/format.ml
+++ b/src/dune_lang/format.ml
@@ -26,7 +26,10 @@ let print_wrapped_list ~version x =
   let inner = Pp.concat_map ~sep:Pp.space ~f:pp_simple x in
   if version < (2, 8)
   then Pp.char '(' ++ Pp.hovbox ~indent:1 inner ++ Pp.char ')'
-  else Pp.hvbox ~indent:1 (Pp.char '(' ++ inner ++ Pp.char ')')
+  else
+    (if version < (3, 17) then Pp.hvbox else Pp.hovbox)
+      ~indent:1
+      (Pp.char '(' ++ inner ++ Pp.char ')')
 ;;
 
 let pp_comment_line l = Pp.char ';' ++ Pp.verbatim l
--- a/test/blackbox-tests/test-cases/describe/describe-workspace-pp.t
+++ b/test/blackbox-tests/test-cases/describe/describe-workspace-pp.t
@@ -104,20 +104,13 @@ not stable across different setups.
    (executables
     ((names (exe))
      (requires
-      (c39d8e11db2363236e69af7750ce7b9a
-       c9367091ddd9a70d99fc22ede348f17c
-       1f2b5eb300ea716920494385a31bb5fb
-       5014e215e204cf8da6c32644cda1b31e
-       249b2edaf3cc552a247667041bb5f015
-       ba85adfb1c97e7d7af3df35b16b2fc0d
-       2363fd46dac995a1c79679dfa1a9881b
-       43b7cbe1f93f4f502ec614971027cff9
-       e68a558facd1546b51c7abdbf6aed1cb
-       24f4eb12e3ff51b310dbf7443c6087be
-       449445be7a24ce51e119d57e9e255d3f
-       5ae836dcdead11d5c16815297c5a1ae6
-       2c61db8e94cb08e0fe642152aee8121a
-       6fb5d46437c55abca48c8b995f8afa51
+      (c39d8e11db2363236e69af7750ce7b9a c9367091ddd9a70d99fc22ede348f17c
+       1f2b5eb300ea716920494385a31bb5fb 5014e215e204cf8da6c32644cda1b31e
+       249b2edaf3cc552a247667041bb5f015 ba85adfb1c97e7d7af3df35b16b2fc0d
+       2363fd46dac995a1c79679dfa1a9881b 43b7cbe1f93f4f502ec614971027cff9
+       e68a558facd1546b51c7abdbf6aed1cb 24f4eb12e3ff51b310dbf7443c6087be
+       449445be7a24ce51e119d57e9e255d3f 5ae836dcdead11d5c16815297c5a1ae6
+       2c61db8e94cb08e0fe642152aee8121a 6fb5d46437c55abca48c8b995f8afa51
        f9851d3f8ae32391e7594cf97332a78c))
      (modules
       (((name Exe)
@@ -147,8 +140,7 @@ not stable across different setups.
      (uid f9851d3f8ae32391e7594cf97332a78c)
      (local true)
      (requires
-      (ba85adfb1c97e7d7af3df35b16b2fc0d
-       2c61db8e94cb08e0fe642152aee8121a
+      (ba85adfb1c97e7d7af3df35b16b2fc0d 2c61db8e94cb08e0fe642152aee8121a
        6fb5d46437c55abca48c8b995f8afa51))
      (source_dir _build/default/dummy_ppx)
      (modules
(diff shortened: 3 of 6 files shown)


rgrinberg (approved):
LGTM. Changes entry please

Khady:
Thanks! It would definitely help with `run`. I don't have a real opinion about the impact on other wrapped lists. From the diff it generally looks good. Two points where it could look worse are `modules` and `libraries` where it can be useful to have things kept in a alphabetical order to ease reading. I tend to think that the horizontal layout would be fine anyway.

nojb:
> I don't have a real opinion about the impact on other wrapped lists.

Unfortunately, the formatter works at the level of s-expressions; so we don't really know the difference between "run" and other lists, so using different formatting for different types of wrapped lists is not straightforward at the moment.

Khady (approved):
thanks :)

Leonidas-from-XIV:
I added a changelog entry and merged it to move this a bit forward :-) @nojb feel free to adjust the wording if that's not what you had in mind.

nojb:
@Leonidas-from-XIV Seems fine, thanks. The reason I had not merged it is that I was not 100% sure that the new style is always better than the old style, but I guess there's no harm to try it out. Thanks!

maiste:
While releasing [3.17]() we found that this PR seems to break odoc `@runtest` (see the [CI](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/f1d49be945320c030eb896dc36ee563859d291dc/variant/compilers,4.14,dune-configurator.3.17.0~alpha0,revdeps,odoc.2.4.3)). @nojb is the version here related to the `dune` version or the version from the `lang dune <version>` in `dune-project`? If this is the executable version, it would break dune for people doing some tests with `sexpr` as long as they upgrade to `dune.3.17.0`.

nojb:
> While releasing [3.17](https://github.com/ocaml/dune/issues/11143#issuecomment-2502349062) we found that this PR seems to break odoc `@runtest` (see the [CI](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/f1d49be945320c030eb896dc36ee563859d291dc/variant/compilers,4.14,dune-configurator.3.17.0~alpha0,revdeps,odoc.2.4.3)). @nojb is the version here related to the `dune` version or the version from the `lang dune <version>` in `dune-project`? If this is the executable version, it would break dune for people doing some tests with `sexpr` as long as they upgrade to `dune.3.17.0`.

I'm away from my computer at the moment. I'll take a look later and post back here any findings.

nojb:
> > While releasing [3.17](https://github.com/ocaml/dune/issues/11143#issuecomment-2502349062) we found that this PR seems to break odoc `@runtest` (see the [CI](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/f1d49be945320c030eb896dc36ee563859d291dc/variant/compilers,4.14,dune-configurator.3.17.0~alpha0,revdeps,odoc.2.4.3)). @nojb is the version here related to the `dune` version or the version from the `lang dune <version>` in `dune-project`? If this is the executable version, it would break dune for people doing some tests with `sexpr` as long as they upgrade to `dune.3.17.0`.
> 
> I'm away from my computer at the moment. I'll take a look later and post back here any findings.

From a quick glance: the rule in question

https://github.com/ocaml/odoc/blob/822d266232fccdffbd4922434c81c45ab6d583f4/test/generators/dune#L11-L16

uses `dune format-dune-file` without a `--dune-version` argument, so the latest available version of the Dune language will be used, in which the case the observed behaviour is the expected one.

nojb:
> Is this behaviour specific to `dune format-dune-file` where you have to specify the version or it will use the executable version?

Yes, it is specific to this command, which does not assume to be in a Dune workspace at all to begin with.

Leonidas-from-XIV:
Thanks for the investigation, you're right that `format-dune-file` has this documented:

```
       --dune-version=VERSION (absent=3.17)
           Which version of Dune language to use.
```

I personally would argue that this is quite unhelpful default as then the editor needs to parse the dune lang version or have something that's changing depening on which dune version your package management tool has selected.

It's also different from how the rest of Dune works.

nojb:
> It's also different from how the rest of Dune works.

Not really, as there are no analogs of this command in Dune: Dune is not typically supposed to call Dune in a user rule, and indeed `dune format-dune-file` was mostly added as a debugging aid, not meant to be used from user rules. Instead, we could/should expose a new type of action (ie `(format_dune_file <path>)`) that could be used in user rules and that would take into account the current version of the Dune language.

I can look into adding this, but it won't fix the immediate issue, of course.

maiste:
@nojb would it be a problem if we remove the fix from `3.17.0` and reintroduce it in a future dune release? It would give us more time to "fix" `odoc` and make them ready to support this change. WDYT?

nojb:
> @nojb would it be a problem if we remove the fix from `3.17.0` and reintroduce it in a future dune release? It would give us more time to "fix" `odoc` and make them ready to support this change. WDYT?

Sounds good to me.

Leonidas-from-XIV:
The documentation states that this is also supposed to be used for editors, but if an editor does not use the `--dune-version` command, then it will still format the files in an inconsistent way.

And while I can see the point of being independent from the workspace, I would expect that in the presence of the workspace there would be a way to use the dune-lang version, without having to parse `dune-project` manually.

The reason why I think it is wrong is because I am thinking what what PR to send upstream to odoc to fix this:

1. I can add `--dune-version=3.16` to keep it at a fixed version, but this will break the lower-bounds check if OPAM selects Dune 3.7
2. I can specify `--dune-version=3.7` but this is fairly ugly because now odoc needs to update two places. I thought that maybe `--dune-version=%{dune_lang_version}` could be used, but the variable does not exist
3. We could introduce `--read-dune-version-from-workspace` but I really don't like flags of the `--do-the-right-thing` kind as people need to be aware of them so it pushes the duty to do the right thing onto users.
4. We could, as you suggest expose an action, but this doesn't help editors, so if they use this command they still need to determine the dune-lang version to use it properly.

My suggestion would be skipping this from 3.17 and reintroducing this in 3.18 with a change that if `--dune-version` is not specified, it tries to read the version from the project and if that fails then it falls back to using the dune-version. WDYT?

nojb:
> > with a change that if `--dune-version` is not specified, it tries to read the version from the project and if that fails then it falls back to using the dune-version.
> 
> Sounds reasonable.

Thinking more about this: reading the version from *which* project? The one corresponding to the current working directory, or the one containing the target Dune file? I guess the latter, as otherwise the result of the command would depend on the current directory.

Leonidas-from-XIV:
The one that the file would be belonging to, a little bit like vendoring a project creates a barrier for automatic formatting.

Khady:
if `dune format-dune-file` starts to read some of the dune config for the project, could it go all the way and solve https://github.com/ocaml/dune/issues/3516?

nojb:
> if `dune format-dune-file` starts to read some of the dune config for the project, could it go all the way and solve #3516?

I guess so. I think it would be the same as calling `dune build @fmt` but restricting the effect to a single file. Needs to be checked that this behaviour is OK for existing users of `dune format-dune-file`. You mentioned OCaml-LSP in #3516. Are you aware of any others?

Incidentally, in #11166 there is a first (orthogonal) step which is to expose a `(format-dune-file)` action to be used within Dune files instead of calling `dune format-dune-file`.

Khady:
> Are you aware of any others?

Nothing that comes directly to my mind

mbarbin:
> You mentioned OCaml-LSP in #3516. Are you aware of any others?

I'll add [dunolint](https://github.com/mbarbin/dunolint) to the list of `dune format-dune-file` users.

PR #10894, opened 2024-09-09
Return [result] from [Workspace_root.create]

gridbugs:
This makes it possible to check whether the user is currently inside a dune project. This is needed for dev-tools to have fallback behaviour when run outside of a dune project which is necessary to allow users to configure their editors to invoke dev-tools via dune. For example, if a user configures their editor to start ocamllsp with the command `dune ocamllsp`, dune needs to be able to handle the case where the user edits an ocaml file outside of a dune project, for example by attempting to invoke ocamllsp from PATH.

### DIFF
--- a/bin/workspace_root.mli
+++ b/bin/workspace_root.mli
@@ -1,3 +1,5 @@
+open! Stdune
+
 (** Finding the root of the workspace *)
 
 module Kind : sig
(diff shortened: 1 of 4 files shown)


bin/workspace_root.mli
```diff
@@ -16,4 +16,5 @@ type t =
   ; kind : Kind.t
   }
 
-val create : default_is_cwd:bool -> specified_by_user:string option -> t
+val create : default_is_cwd:bool -> specified_by_user:string option -> (t, exn) result
```
rgrinberg, line 19: How about just returning `User_error.t`? Usually it's a mistake to delay raising other types of exceptions (code errors for example)

gridbugs: Good point, fixed.

rgrinberg (approved):

PR #10897, opened 2024-09-10
Fix url opened by "dune ocaml doc"

gridbugs:
Prior to this change, "dune ocaml doc" would attempt to open an incorrect url in browsers: "file://_build/default/_doc/_html/index.html". This is incorrect because it appears to be a path relative to the project root, but browsers interpret "file://foo/some/path" as the path "/some/path" on the host referred to by "foo". In my experiments however, the "foo" component is ignored and the browser redirects to "file:///some/path" instead. Indeed, when I run "dune ocaml doc" I end up at "file:///default/_doc/_html/index.html", with the "_build" component of the path being ignored, and the remainder of the path treated as an absolute path (which obviously doesn't refer to the correct file).

It seems like droppping the "file://" prefix of the url is enough to get browers to accept the relative path and correctly navigate to "file:///absolute/path/to/index.html". Alternatively we could fix the logic for computing an absolute path to index.html and keep the "file://" prefix of the url, but that seems more complicated. I've updated a variable name from "absolute_toplevel_index_pathh" to "relative_toplevel_index_path" to reflect that it is in fact a path relative to the project root and not an absolute path.

### DIFF
--- a/bin/ocaml/doc.ml
+++ b/bin/ocaml/doc.ml
@@ -23,7 +23,7 @@ let term =
       Alias.in_dir ~name:Dune_rules.Alias.doc ~recursive:true ~contexts:setup.contexts dir
       |> Alias.request
     in
-    let absolute_toplevel_index_path =
+    let relative_toplevel_index_path =
       let toplevel_index_path =
         let is_default ctx = ctx |> Context.name |> Dune_engine.Context_name.is_default in
         let doc_ctx = List.find_exn setup.contexts ~f:is_default in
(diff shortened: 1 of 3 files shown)


gridbugs:
On MacOS I found that without the "file://" component of the URL the "-u" argument to the "open" command no longer works, but just running "open path/to/index.html" opens a browser and navigates to the absolute path to the file.

Leonidas-from-XIV (approved):
Looks good to me, `xdg-open` on my system does accept relative paths and opens them in the browser just fine.

rgrinberg (approved):

Issue #9294, opened 2023-11-27
Handle depexts field of opam files in a principled way


Alizter:
When an opam file has a `depexts` field, opam is instructed to use it's opam-depext library, historically a separate program, to find a good error message or suggestion of commands to run, later versions even running those commands.

In the context of Dune package management, we have access to the depexts field in an opam file, however currently we just ignore it. I propose that instead of ignoring it, we raise a user error with some detailed information about which commands should be run.

Here are my proposals with their strengths and drawbacks:

1. Run `opam-depext`
  + Advantages:
    - We can reuse opam-depext command suggestions and logic.
    - making it very easy for users to install external dependencies.
  + Disadvantages:
    - We rely on an external program or more internal opam libraries.
    - We shouldn't really involve ourselves with system installs, as that might introduce permission issues into the workflow with less experienced users (already (or used to be) a problem with opam).

2. Warn the user of all the depexts that were found during locking. We can even give system specific ones by using the filters on the depexts field.
  + Advantages:
    - Users will no longer get cryptic error messages during the building of the package.
  + Disadvantages:
    - Users not locking and just running the build won't get the messages.
    - It's not always obvious what commands to run from just the depexts entries. opam-depext solves this by having os-distribution specific suggestions for the external library or program.
    - This is noisy during locking, especially if the user already has the external dependencies installed. We cannot really determine if a depext is resolved.

3. Record the depexts in the lock file, and report them as hints during an unsuccessful build.
  + Advantages:
    - This doesn't affect locking.
    - When the user encounters a build error from a package, we give them a useful starting point to resolve it.
  + Disadvantages: 
    - Not clear what to do with depexts during locking, do we keep filters or not?
    - As in 2, it might not be useful simply to report the depexts field.

4. Keep the status quo.
   + Advantages:
     - We don't need to design or maintain any special handling of depexts.
   + Disadvantages:
     - User will get cryptic errors for missing dependencies in package builds, with no clear way to move forward.

These are the proposals on how to handle this issue. My personal opinion would be 3 with a mix of 1. Which would become the following proposal:

5. Do 3, but use opam-depext, if available, to give better hints.

I'm interested to hear other proposals and opinions on how to handle this situation.

rgrinberg:
Solved by @moyodiallo

PR #10900, opened 2024-09-10
pkg: add support for opam `depexts`

moyodiallo:
`dune` will just print a warning message whenever a package fails to build. The warning message would be a list of the `depexts` dependency without checking if they exist in the current OS or not.

This is one of the way for supporting depexts that mentioned in https://github.com/ocaml/dune/issues/9294.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/depexts-error-message.t
+++ b/test/blackbox-tests/test-cases/pkg/depexts-error-message.t
@@ -1,4 +1,4 @@
-Exercises end to end locking and building a simple project.
+ When a package fails to build, dune will print opam depexts warning.
 
   $ . ./helpers.sh
   $ mkrepo
@@ -6,12 +6,8 @@ Exercises end to end locking and building a simple project.
 
 Make a library that would fail when building it:
   $ mkdir foo
-  $ cd foo
-  $ cat > dune-project <<EOF
-  > (lang dune 3.13)
-  > (
+  $ cat > foo/dune-project <<EOF
   > EOF
-  $ cd ..
   $ tar -czf foo.tar.gz foo
   $ rm -rf foo
 
@@ -24,9 +20,6 @@ Make a package for the library with depexts:
   >     "build"
   >     "-p"
   >     name
-  >     "-j"
-  >     jobs
-  >     "@install"
   >   ]
   > ]
   > url {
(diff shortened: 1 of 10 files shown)


test/blackbox-tests/test-cases/pkg/depexts.t
```diff
@@ -0,0 +1,104 @@
+Exercises end to end locking and building a simple project.
```
rgrinberg, line 1: Can you split this test in two parts

1. One that demonstrates how we convert the `depext` field in opam to lock directory packages. This test should use the solver.
2. One that demonstrates how this new field is affecting the error messages. This test should not use the solver.

rgrinberg (changes requested):

test/blackbox-tests/test-cases/pkg/depexts.t
moyodiallo: Done.

test/blackbox-tests/test-cases/pkg/depexts-error-message.t
```diff
@@ -0,0 +1,61 @@
...
+  >  (depends foo))
+  > EOF
+  $ cat > dune <<EOF
+  > (executable
+  >  (public_name bar)
+  >  (libraries foo))
+  > EOF
+
+Lock, build, and run the executable in the project:
+  $ dune pkg lock
```
rgrinberg, line 47: Instead of relying on the solver, you can just write `foo.pkg` directly.

rgrinberg:
I tweaked the error message a bit. The rest looks good but testing the `depexts` field in the lock dir shouldn't involve the solver.

test/blackbox-tests/test-cases/pkg/depexts-error-message.t
```diff
@@ -0,0 +1,61 @@
+ When a package fails to build, dune will print opam depexts warning.
```
rgrinberg, line 1: Can you move these two tests into a `depexts` sub-directory?

anmonteiro:
we tried this with `reason-react` and it resulted in a dune crash. check https://github.com/ocaml/dune/issues/10930

anmonteiro:
this is likely because we use `npm-version` which is a variable interpreted by an opam-plugin (check https://github.com/reasonml/reason-react/blob/c07e413750c5ca1632e6d3405a3cfbdf8d24c84e/reason-react.opam#L45-L48).

moyodiallo:
I'm for reverting it until we fix the subsequent bug.

rgrinberg:
Do you want to add a repro test to demonstrate the issue before you revert?

moyodiallo:
A question: is `npm-version` an opam variable ? I'm not sure which test to add.

anmonteiro:
`npm-version` is [interpreted](https://github.com/ahrefs/opam-check-npm-deps/blob/dffa1f2c9e787420e71dc6507e825305c4ac8e1d/src/main.ml#L33-L38) by `opam-check-npm-deps`

moyodiallo:
This variable is specific to `depexts` on my understanding so far. I think the better way to go, is to maintain the PR and fix issue. If anyone has another proposition I'm open to it.

PR #10904, opened 2024-09-11
Add `codeberg` Source Kind

nlordell:
Add a new kind to the `source` stanza for Codeberg projects.


test/blackbox-tests/test-cases/codeberg.t
```diff
@@ -0,0 +1,17 @@
...
+  > (package
+  >  (allow_empty)
+  >  (name foo))
+  > EOF
+
+  $ dune build
+  $ cat foo.opam | grep -i codeberg.org
+  homepage: "https://codeberg.org/john/doe"
+  bug-reports: "https://codeberg.org/john/doe/issues"
+  dev-repo: "git+https://codeberg.org/john/doe.git"
```
Leonidas-from-XIV, line 17: It would be good to have a test that shows that this doesn't work on dune-lang < 3.17.

nlordell: Added in 207cae89ccb769241f8d7f31f913f0398b095193

Leonidas-from-XIV:
@nlordell Could you add a changelog entry for this? I think it would be nice to know about it.

PR #10906, opened 2024-09-12
refactor: share code for formatting aligned messages

gridbugs:


rgrinberg:
Thanks. We should probably remove all of this display logic from the engine. But this is good enough for now.

Issue #10910, opened 2024-09-12
`dune pkg lock -x <context>` should lock the associated lock dir


maiste:
With Dune package management, when you are trying to lock a project with a specified context, it locks the default context instead of locking the lock directory associated with the context (if it exists). The following example should put the locked files to `foo.lock` but instead, it puts them in `dune.lock` when running `dune pkg lock -x foo`:

```lisp
 (lock_dir
  (path foo.lock)
  (repositories mock))
 (context
  (default
   (name foo)
   (lock_dir foo.lock)))
```

rgrinberg:
The only bug as far as I can tell is that we accept the `-x` parameter for `dune pkg lock`. `dune pkg lock` should be completely oblivious to contexts.

maiste:
I see! I'm still familiarizing with build context in `Dune`. Sorry for the noise.

The issue here is that it comes with the `Common.t term` argument. Would there be a way to remove this argument?

rgrinberg:
The name `Common.t` is a little too optimistic. It assumes that all of its flags are going to be useful to all commands. That's certainly not the case, as most of the flags from `Common.t` are in fact useless for `dune pkg`. This `-x` flag, for example, is only useful for build and install commands. So it should really not exist in common.

I don't really have concrete suggestions other than considering every flag in `Common.t` on a case by case basis and then making a decision on whether it really is "common".

maiste:
OK, I understand! It is more a question of legacy here, then. It would need another issue to discuss such a redesign, if required :sweat_smile: From offline discussion, suggestions are to improve the content of the help message from the CLI and to add more information in the cross compilation page to explain the meaning of this flag.

On a second hand, the behavior that triggers this issue was that you can define a `(lock_dir name.lock)` stanza in the `(context ...)` stanza. From offline discussion, I learned it was made this way so you can define a lock directory "per system" (x86_64, Linux, Windows, etc) as they are not stable per system. Considering this stanza, it would make sens to have a way to build a lock dir associated with a context (for example, a `--context` flag). I know that you said context are oblivious of lock dir, but this stanza kind of breaks this rule :confused:

rgrinberg:
> Considering this stanza, it would make sens to have a way to build a lock dir associated with a context (for example, a --context flag). I know that you said context are oblivious of lock dir, but this stanza kind of breaks this rule 😕

You could indeed create such a flag. I don't think it would be very useful though. For a couple of reasons:

[...]

PR #10911, opened 2024-09-12
test: reproduction case for issue #10910

maiste:
This is a reproduction test for issue #10910. It shows that locking with a context lock the wrong directory.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/gh10910.t
+++ b/test/blackbox-tests/test-cases/pkg/gh10910.t
@@ -1,0 +1,44 @@
+This issue demonstrates a bug when we have a context which specifies a lock dir
+in the `dune-workspace` file and try to lock it using the same context. It
+should pick the right context but instead it lock all the directories.
+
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/pkg/gh10910.t
```diff
@@ -0,0 +1,44 @@
...
+  > (lang dune 3.16)
+  > (package
+  >  (name test)
+  >  (allow_empty)
+  >  (depends foo))
+  > EOF
+
+It tries to lock `dune.lock` whereas it should lock `foo.lock` as specified in
+the `dune-workspace` file:
+  $ dune pkg lock -x foo
```
rgrinberg, line 41: The bug here is that the `-x` flag is meaningless for creating a lock directory. `-x` only makes sense for cross compilation.

rgrinberg (changes requested):

maiste:
See discussions in  #10910

PR #10912, opened 2024-09-12
Add links between tutorial sections

cuihtlauac:
Up to my understanding, tutorial introduced in PR #10711 does
not provide user discoverable means to navigate between its
parts. This add a links to the next section.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -8,6 +8,7 @@ libraries, run tests, and much more.
    :maxdepth: 2
 
    getting-started/index
+   tutorials/index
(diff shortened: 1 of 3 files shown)


emillon:
That's interesting. There are usually next and prev links generated (see screenshot) but they're not generated, presumably because the tutorial, is in WIP form, is an orphaned page. I'll try to massage that instead of having to insert manual links.

image

emillon:
The tutorial is now listed but there's the WIP banner and the nav links.

PR #10913, opened 2024-09-12
Changelog entry for the change of dune cache default (#10710)

ElectreAAS:


doc/changes/10710.md
```diff
@@ -0,0 +1 @@
+- Enable dune's cache by default. (#10710, @ElectreAAS, joint work with @Leonidas-from-XIV)
```
rgrinberg, line 1: Probably worth mentioning the new `DUNE_CACHE_RULES` environment variable.

ejgallego:
Thanks for the update @ElectreAAS !

I have a last question, is it possible with the new code to recover the old behavior when the user had used `(cache enabled)` in the config file, without using any env variables?

doc/caching.rst
```diff
@@ -15,19 +15,26 @@ or simply undoing some changes within the same workspace.
...
 systems and ``%LOCALAPPDATA%\Microsoft\Windows\Temporary Internet Files\dune`` on Windows.
 You can change the default location by setting the environment variable
 ``DUNE_CACHE_ROOT``.
 
+Caching User Rules
+==================
+
+Not all rules are cached by default: user-created ``(rule _)`` actions are not to prevent mistakes.
+You can enable the caching of these rules by setting the environment variable ``DUNE_CACHE_RULES`` to ``enabled``.
+
```
Khady, line 37: to be clear, this is about caching the output of such rules right? Not the rules themselves

doc/caching.rst
```diff
@@ -15,19 +15,26 @@ or simply undoing some changes within the same workspace.
 Configuration
 =============
 
-For now, Dune cache is an opt-in feature. There are three ways to enable it.
-Choose the one that is more convenient for you:
+.. TODO (ElectreAAS): update version number here when releasing
+Since %version%, Dune cache is now an opt-out feature: it is enabled by default.
```
Leonidas-from-XIV, line 19: I think this can just be set to 3.17 already.

Leonidas-from-XIV:
> I have a last question, is it possible with the new code to recover the old behavior when the user had used `(cache enabled)` in the config file, without using any env variables?

No, not with the current code, you need to explicitly opt-in to caching the user rules.

doc/caching.rst
ElectreAAS: I don't really know what caching a rule without caching its result would even mean. So yes I believe this is about caching outputs.

rgrinberg:
> is it possible with the new code to recover the old behavior when the user had used (cache enabled) in the config file, without using any env variables?

As Marek mentioned, it's not possible at the moment. If you'd like to add a config option for this in addition to the env var, you are welcome to.

ejgallego:
> As Marek mentioned, it's not possible at the moment. If you'd like to add a config option for this in addition to the env var, you are welcome to.

Thanks folks for the confirmation.

I think indeed an option to enable this in `./config/dune` should be added, otherwise quite a few users I know (virtually all Coq devs + many users) will regress as they have a setup where they enable the cache in the config file (their rules are safe), however it is far from trivial for them to switch to using environment variables.

I guess that would be a regression, but likely this setup did fall under the "experimental" category, right?

christinerose (approved):

doc/changes/10710.md
ElectreAAS: Fixed now :)

doc/caching.rst
ElectreAAS: Fixed now :)

christinerose (approved):

nojb:
> I am aware of the discussion happening in #10710 but I think this PR accurately describes the current state of `main`. If any changes will be done that will happen in a separate PR so I think merging this one in the meantime is sensible.

The discussion in #10710 is converging, so you may want to hold off before merging this one. See #10944.

ElectreAAS:
#10944 has been merged, and it includes a correct changelog.
Closing this one.

PR #10914, opened 2024-09-13
Add odoc dev tool

gridbugs:
The command `dune ocaml doc` will now lock, build, and run odoc as a dev tool if the dev tools feature is enabled.

### DIFF
--- a/src/dune_rules/odoc.ml
+++ b/src/dune_rules/odoc.ml
@@ -245,14 +245,33 @@ let odoc_base_flags quiet build_dir =
   | Nonfatal -> S []
 ;;
 
+let odoc_dev_tool_lock_dir_exists () =
+  let path = Dune_pkg.Lock_dir.dev_tool_lock_dir_path Odoc in
+  Fs_memo.dir_exists (Path.Outside_build_dir.In_source_dir path)
+;;
+
+let odoc_dev_tool_exe_path_building_if_necessary () =
+  let open Action_builder.O in
+  let path = Path.build (Pkg_dev_tool.exe_path Odoc) in
+  let+ () = Action_builder.path path in
+  Ok path
+;;
+
 let odoc_program sctx dir =
-  Super_context.resolve_program
-    sctx
-    ~dir
-    ~where:Original_path
-    "odoc"
-    ~loc:None
-    ~hint:"opam install odoc"
+  let open Action_builder.O in
+  let* odoc_dev_tool_lock_dir_exists =
+    Action_builder.of_memo (odoc_dev_tool_lock_dir_exists ())
+  in
+  match odoc_dev_tool_lock_dir_exists with
+  | true -> odoc_dev_tool_exe_path_building_if_necessary ()
+  | false ->
+    Super_context.resolve_program
+      sctx
+      ~dir
+      ~where:Original_path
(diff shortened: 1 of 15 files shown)


rgrinberg:
Doesn't odoc have the same limitation as lsp? I.e it needs to be compiled with the same version of OCaml as the project.

src/dune_rules/odoc.ml
```diff
@@ -245,14 +245,33 @@ let odoc_base_flags quiet build_dir =
   | Nonfatal -> S []
 ;;
 
+let odoc_dev_tool_lock_dir_exists () =
+  let path = Dune_pkg.Lock_dir.dev_tool_lock_dir_path Odoc in
+  Fs_memo.dir_exists (Path.source path |> Path.as_outside_build_dir_exn)
```
rgrinberg, line 250: Don't think the `as_outside_build_dir_exn` is necessary. You can just use of the constructors for `Path.Outside_build_dir` directly.

rgrinberg (changes requested):

test/blackbox-tests/test-cases/pkg/odoc/dev-tool-odoc-basic.t
```diff
@@ -0,0 +1,35 @@
...
+  > (package
+  >  (name foo)
+  >  (allow_empty))
+  > EOF
+
+Lock and build the fake odoc executable, and then use the fake
+executable to attempt to generate documentation. The fake executable
+doesn't actually do anything so the command fails but this
+demonstrates that it was run.
+  $ DUNE_CONFIG__PKG_BUILD_PROGRESS=enabled DUNE_CONFIG__LOCK_DEV_TOOL=enabled dune ocaml doc
```
moyodiallo, line 24:
```suggestion
  $ DUNE_CONFIG__LOCK_DEV_TOOL=enabled dune ocaml doc
```

moyodiallo (changes requested):

test/blackbox-tests/test-cases/pkg/odoc/dev-tool-odoc-basic.t
gridbugs: Good point, fixed.

src/dune_rules/odoc.ml
gridbugs: Good point

gridbugs:
> Doesn't odoc have the same limitation as lsp? I.e it needs to be compiled with the same version of OCaml as the project.

You're totally right. I didn't realize this but that explains some of the weird errors I've run into with odoc recently:

E.g.
```
File "src/.climate.objs/byte/_unknown_", line 1, characters 0-0:
ERROR: File "climate__.cmt":
not an implementation
```

I'll borrow some of the code from my [ocamllsp dev-tool pr](https://github.com/ocaml/dune/pull/10915#pullrequestreview-2312160014) to make sure that the same compiler is used to build odoc as is used to build the project.

gridbugs:
I updated this so that it always uses the same version of the compiler to build odoc as it does to build the project and added some tests that it relocks the odoc dev tool to keep the compiler dependency in sync with the project.

rgrinberg (approved):

PR #10915, opened 2024-09-13
Add ocamllsp dev tool

gridbugs:
Adds a command `dune ocamllsp` which locks, builds and runs ocamllsp from within the _build directory. It makes sure that the version of ocamlc used to compile ocamllsp is the same as the one in the project's lockdir since ocamllsp needs to be built with the same version of ocamlc as the code that it's analyzing.

This command will exit with an error message unless it's run within a dune project with a dune.lock directory containing a lockfile for the package "ocaml" so it's not ready for use by editors. Support for falling back to a system or opam-managed installation of ocamllsp will come in a later change.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/ocamllsp/dev-tool-ocamllsp-relock-on-ocaml-version-change.t
+++ b/test/blackbox-tests/test-cases/pkg/ocamllsp/dev-tool-ocamllsp-relock-on-ocaml-version-change.t
@@ -1,0 +1,63 @@
+Test that if the version of the "ocaml" package in the project's
+lockdir changes then the ocamllsp dev tool is re-locked to be built
+with the version of the ocaml compiler now in the project's
+lockdir. This is necessary because ocamllsp must be compiled with the
+same version of the ocaml compiler as the code that it's analyzing.
+
+  $ . ../helpers.sh
+  $ . ./helpers.sh
+
+  $ mkrepo
+  $ make_mock_ocamllsp_package
+  $ mkpkg ocaml 5.2.0
+  $ mkpkg ocaml 5.1.0
+
+  $ setup_ocamllsp_workspace
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.16)
+  > 
+  > (package
+  >  (name foo)
+  >  (allow_empty))
+  > EOF
+
+  $ make_lockdir
+  $ cat > dune.lock/ocaml.pkg <<EOF
+  > (version 5.2.0)
+  > EOF
+
+Initially ocamllsp will be depend on ocaml.5.2.0 to match the project.
+  $ dune tools exec ocamllsp
+  Solution for dev-tools.locks/ocaml-lsp-server:
+  - ocaml.5.2.0
+  - ocaml-lsp-server.0.0.1
+       Running 'ocamllsp'
+  hello from fake ocamllsp
+  $ cat dev-tools.locks/ocaml-lsp-server/ocaml.pkg
+  (version 5.2.0)
+
+We can re-run "dune tools exec ocamllsp" without relocking or rebuilding.
+  $ dune tools exec ocamllsp
+       Running 'ocamllsp'
+  hello from fake ocamllsp
+
+Change the version of ocaml that the project depends on.
+  $ cat > dune.lock/ocaml.pkg <<EOF
+  > (version 5.1.0)
+  > EOF
+
+Running "dune tools exec ocamllsp" causes ocamllsp to be relocked and rebuilt
+before running. Ocamllsp now depends on ocaml.5.1.0.
+  $ dune tools exec ocamllsp
+  The version of the compiler package ("ocaml") in this project's lockdir has
+  changed to 5.1.0 (formerly the compiler version was 5.2.0). The dev-tool
+  "ocaml-lsp-server" will be re-locked and rebuilt with this version of the
+  compiler.
+  Solution for dev-tools.locks/ocaml-lsp-server:
+  - ocaml.5.1.0
+  - ocaml-lsp-server.0.0.1
+       Running 'ocamllsp'
+  hello from fake ocamllsp
+  $ cat dev-tools.locks/ocaml-lsp-server/ocaml.pkg
+  (version 5.1.0)
--- a/bin/lock_dev_tool.ml
+++ b/bin/lock_dev_tool.ml
@@ -208,3 +208,4 @@ let lock_ocamlformat () =
 ;;
 
 let lock_odoc () = lock_dev_tool Odoc None
+let lock_ocamllsp () = lock_dev_tool Ocamllsp None
--- a/bin/lock_dev_tool.mli
+++ b/bin/lock_dev_tool.mli
@@ -3,3 +3,4 @@ open! Import
 val is_enabled : bool Lazy.t
 val lock_ocamlformat : unit -> unit Memo.t
 val lock_odoc : unit -> unit Memo.t
+val lock_ocamllsp : unit -> unit Memo.t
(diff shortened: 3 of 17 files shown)


bin/ocamllsp.ml
```diff
@@ -0,0 +1,133 @@
...
+      | Error failure_reason -> raise (User_error.E (Failure.message failure_reason))
+      | Ok ocaml_compiler_version ->
+        let* () = Lock_dev_tool.lock_ocamllsp ~ocaml_compiler_version in
+        let+ () = build_ocamllsp common in
+        run_ocamllsp common ~args)
+;;
+
+let info =
+  let doc = "Run ocamllsp, installing it as a dev tool if necessary." in
+  Cmd.info "ocamllsp" ~doc
```
leostera, line 130:
```suggestion
  Cmd.info "lsp" ~doc
```

Small change but important to keep the DX consistent with `fmt`, `doc`, etc.

bin/ocamllsp.ml
```diff
@@ -0,0 +1,133 @@
...
+     dev-tool. *)
+  let message reason =
+    User_message.make
+      [ Pp.textf
+          "Unable to run %s as a dev-tool because %s."
+          ocamllsp_exe_name
+          (Reason.to_string reason)
+      ; Pp.nop
+      ; Pp.textf
+          "To run %s as a dev-tool, the following conditions mest be met:"
```
leostera, line 38:
```suggestion
          "To run %s as a dev-tool, the following conditions must be met:"
```

moyodiallo:
Testing, I found.

` DUNE_CONFIG__PKG_BUILD_PROGRESS=enabled DUNE_CONFIG__TOOLCHAINS=enabled dune ocamllsp`
works well.

When `dune fmt` follows, I'm getting this error message:
```
$ DUNE_CONFIG__PKG_BUILD_PROGRESS=enabled DUNE_CONFIG__LOCK_DEV_TOOL=enabled dune fmt
Error: Multiple rules generated for
_build/_private/default/.pkg/ocaml-base-compiler/source:
- dune.lock/ocaml-base-compiler.pkg:24
- dune.lock/ocaml-base-compiler.pkg:13
-> required by - package cmdliner
```

bin/ocamllsp.ml
gridbugs: The reason I think it should remain `dune ocamllsp` is this command is equivalent to running `ocamllsp`. `dune fmt` and `dune doc` both perform some higher-level function that inolves running `ocamlformat` and `odoc` respectively, but `dune ocamllsp` just builds `ocamllsp` and then runs it.

bin/ocamllsp.ml
gridbugs: Nobody is going to run `dune lsp` - it's always going to be started by your text editor. In a later PR I'm going to make it so that if the conditions necessary for ocamllsp to be installed as a dev tool are not met, that dune will try to fall back to a system or opam-managed instance of ocamllsp. This is necessary so that users can put `dune ocamllsp` in their editor configs but still be able to edit ocamlc code without a lockfile or even outside a dune project. Without this I don't expect many people to start using `dune ocamllsp`.

bin/ocamllsp.ml
rgrinberg: For now, just make it `dune ocaml lsp`. That's where we put commands that aren't often useful to the user.

bin/ocamllsp.ml
```diff
@@ -0,0 +1,133 @@
...
+  let open Memo.O in
+  let context =
+    (* Dev tools are only ever built with the default context. *)
+    Context_name.default
+  in
+  let* result = Dune_rules.Lock_dir.get context in
+  match result with
+  | Error _ -> Memo.return @@ Error Failure.Reason.No_lock_dir
+  | Ok { packages; _ } ->
+    (match Package_name.Map.find packages compiler_package_name with
```
rgrinberg, line 95: The way this works is by matching the version number of OCaml inside the project with the version number we're going to solve for. While this will usually work, it will fail for custom versions of the compiler.

The more correct thing to do would be to "pin" the compiler in LSP to be the compiler from the project. Don't think this is strictly necessary here, but we should leave a comment + add a test case for this situation.

bin/ocamllsp.ml
```diff
@@ -0,0 +1,133 @@
...
+let term =
+  let+ builder = Common.Builder.term
+  and+ args = Arg.(value & pos_all string [] (info [] ~docv:"ARGS")) in
+  match is_in_dune_project builder with
+  | false -> raise (User_error.E (Failure.message Failure.Reason.Not_in_dune_project))
+  | true ->
+    let common, config = Common.init builder in
+    Scheduler.go ~common ~config (fun () ->
+      let open Fiber.O in
+      locked_ocaml_compiler_version ()
```
rgrinberg, line 118: Did you really need to delay the error from this function? All you did is re-raise it in the caller anyway.

bin/ocamllsp.ml
gridbugs: @rgrinberg what do you think about `dune tools ocamllsp`? There was some discussion yesterday about introducing a `dune tools` command group for commands that just build and run some other tool.

rgrinberg: I guess that's fine too. Maybe it should be `dune tools exec ocamllsp`? I imagine something like `dune tools describe` being useful as well.

gridbugs: I changed the command to `dune tools exec ocamllsp`.

bin/ocamllsp.ml
gridbugs: I added a comment explaining this.

bin/ocamllsp.ml
gridbugs: You're right, this was left over from a previous iteration of this change where the delayed errors made more sense. Fixed.

test/blackbox-tests/test-cases/pkg/ocamllsp/dev-tool-ocamllsp-relock-on-ocaml-version-change.t
```diff
@@ -0,0 +1,64 @@
...
+  >  (allow_empty))
+  > EOF
+
+  $ make_lockdir
+  $ cat > dune.lock/ocaml.pkg <<EOF
+  > (version 5.2.0)
+  > EOF
+
+Initially ocamllsp will be depend on ocaml.5.2.0 to match the project.
+  $ DUNE_CONFIG__PKG_BUILD_PROGRESS=enabled dune tools exec ocamllsp
```
moyodiallo, line 31:
```suggestion
  $ dune tools exec ocamllsp
```
Do we need to print the progress ?  I think the way we print progress may change.

test/blackbox-tests/test-cases/pkg/ocamllsp/dev-tool-ocamllsp-relock-on-ocaml-version-change.t
```diff
@@ -0,0 +1,64 @@
...
+  hello from fake ocamllsp
+
+Change the version of ocaml that the project depends on.
+  $ cat > dune.lock/ocaml.pkg <<EOF
+  > (version 5.1.0)
+  > EOF
+
+Running "dune tools exec ocamllsp" causes ocamllsp to be relocked and rebuilt
+before running. Ocamllsp now depends on ocaml.5.1.0.
+  $ DUNE_CONFIG__PKG_BUILD_PROGRESS=enabled dune tools exec ocamllsp
```
moyodiallo, line 53:
```suggestion
  $ dune tools exec ocamllsp
```
Same as previous comment.

moyodiallo (changes requested):

rgrinberg (approved):

PR #10916, opened 2024-09-13
Add changelog entry for ocamlformat dev-tool feature

moyodiallo:
Due to a mistake when pushing things, the file was remove from https://github.com/ocaml/dune/pull/10647 before the merge.


rgrinberg:
We aren't recording changes to package management until the feature is stabilized.

PR #10918, opened 2024-09-13
pkg: Use `filtered_formula` to represent dependencies

### LINKED ISSUE #10837: Error solving project with deps specified in opam file when deps contain a disjunction
If a project defines a local package in an opam file instead of its dune-project file, and the opam file specifies a disjunction of packages among its dependencies, dune prints an error while solving:
```
Error: Expected formula to be a conjunction of atoms but encountered non-atom
term 'extlib {build} | extlib-compat {build}'
```

To reproduce:
```
$ opam source lutin.2.71.15
$ cd lutin.2.71.15
$ dune pkg lock
File "lutin.opam", line 1, characters 0-0:
Error: Expected formula to be a conjunction of atoms but encountered non-atom
term 'extlib {build} | extlib-compat {build}'
```

The package `lutin` contains this disjunction in its dependencies:
```
"extlib" {build} | "extlib-compat" {build}
```

Note that there's no way to describe a package disjunction in the DSL for specifying dependencies in dune-project, but opam files allow it. Dune can handle the case when an external package (ie. a dependency of the project) has a package disjunction in its dependencies, but not when the current project itself has a package disjunction in its dependencies.

I think it would be valid to just improve the error message here to make it clear that this case is not supported.

Leonidas-from-XIV:
At the moment when an OPAM package in the local packages we only accept a list of `Atom`s, so if it has an disjunction in the dependencies we fail when loading as we attempt to parse the list of `Atom`s into a `Dependency_set.t` structure. This structure only supports a sequence of conjunctions (as that is the only syntax that `dune-project` files support at the moment).

This PR changes the internal representation of the dependencies to be `OpamTypes.filtered_formula` as when reading OPAM files this is what we get from the OPAM file parser. When we pass the data to the solver it is also `OpamTpyes.filtered_formula` that we have to encode back to. Thus I suggest keeping `OpamTypes.filtered_formula` as dependency specification and compute the information we need from it directly. The information is fairly limited as we only need one example package name for an error message and a hash of the dependencies.

The downside of this is that all existing dependency hashes change (see tests), as the representation of the package constraints the hash is calculated over is changing, but I consider this an acceptable compromise.

Fixes #10837


ElectreAAS (commented):
Some tests expectedly fail with hash incongruences, but some others fail with this:
```diff
-  File "x.opam", line 1, characters 0-0:
-  Error: Expected formula to be a conjunction of atoms but encountered non-atom
-  term 'a | b'
+  Error: Unable to solve dependencies for the following lock directories:
+  Lock directory dune.lock:
+  Can't find all required versions.
+  Selected: x.dev a
+  - a -> (problem)
+      No known implementations at all
```

Shouldn't these tests now pass with a proper build?

src/dune_pkg/local_package.ml
```diff
@@ -108,16 +64,82 @@ module For_solver = struct
...
+
+    let to_dyn = Dyn.pair Package_name.to_dyn Opam_dyn.condition
+
+    let compare (name, condition) (other_name, other_condition) =
+      let open Ordering.O in
+      let= () = Package_name.compare name other_name in
+      let compare_filters filter filter' =
+        let get_vars = function
+          | OpamTypes.Constraint _ -> []
+          | Filter filter -> Stdlib.List.sort Stdlib.compare (OpamFilter.variables filter)
```
rgrinberg, line 80: Why do we need to sort?

src/dune_pkg/local_package.ml
```diff
@@ -108,16 +64,82 @@ module For_solver = struct
...
+        in
+        match get_vars filter, get_vars filter' with
+        | v :: _, v' :: _ -> Stdlib.compare v v'
+        | [], _ :: _ -> 1
```
rgrinberg, line 84: We don't use numbers for orderings in the dune code, just the constructors of `Ordering`.

rgrinberg (changes requested):

src/dune_pkg/local_package.ml
Leonidas-from-XIV: This is basically taken taken from `vendor/opam/src/state/opamFileTools.ml` as a whole, because I wanted a way to create a `Set`/`Map` without having to implement my own `compare` function.

I assume this is because the ordering of the filters doesn't matter, if the filters are in a different ordering they should still be considered equal.

src/dune_pkg/local_package.ml
```diff
@@ -108,16 +64,82 @@ module For_solver = struct
...
+      in
+      Ordering.of_int
+        (OpamFormula.compare_formula compare_filters condition other_condition)
+    ;;
+  end
+
+  module Map = Map.Make (T)
+  module Set = Set.Make (T) (Map)
+
+  let maximum_package_set formula =
```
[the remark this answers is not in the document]
Leonidas-from-XIV: That is true, but bar of evaluating the formula with all the filters (which would need access to all of the OPAM variables and lead to potentially different solutions on different systems) this is a simple way to determine an overapproximation of the non-local packages. In the worst case this would lead to

1. Reporting the non-local package error with non-local package that wouldn't get picked (as we only ever pick one at random for the error message)
2. The hash of non-local packages would change if a non-local package changes its dependencies even if we don't actually end up depending on it.

While only an approximation I think the upside is a stable solution across systems. The solver still gets the whole dependency formula and will compute an accurate lock file.

Leonidas-from-XIV:
@ElectreAAS The failure you quoted makes sense, because the test in `opam-file-errors.t` does not define a package `a` or `b`. To merge this PR I think the best way would be to change it to to define `a` (and maybe `b`) and show that these are supported now. The other tests in the file can be deleted to as we don't convert `filtered_formula`s into dune conditions anymore, so the error case that this is exercising would not happen anymore.

The hash changes are an unfortumate side-effect as the non-local dependency set is calculated from a different representation and bar of implementing the OPAM filters into Dune's condition language and then hashing that there is no good way of preserving the previous hashes (or running a hybrid approach of encoding them one way if they can and another if they cannot be represented using dune-lang conditions). I don't think we guarantee that the hashes are stable between versions of Dune.

src/dune_pkg/local_package.ml
Leonidas-from-XIV: That's true and I personally also prefer this, but this function gets passed to `OpamFormula.compare_formula` which requires the integer way of comparing. I can reimplement `compare_formula` to use `Ordering` or add a new function to `Ordering` that would convert comparison functions between the representations. Which of those do you think would be the best approach?

Leonidas-from-XIV:
I've replaced the failing test with one that shows that disjunctions in opam files lead to valid lock directory solutions using either side of the disjunction.

src/dune_pkg/local_package.ml
rgrinberg: Instead of copying this code, is it possible to just expose this code in the opamFileTools mli?

src/dune_pkg/local_package.mli
```diff
@@ -76,11 +57,14 @@ module For_solver : sig
...
-  (** Returns the set of dependencies of all given local packages excluding
-      dependencies which are packages in the provided list. Pass this the list
-      of all local package in a project to get a set of all non-local
-      dependencies of the project. *)
-  val list_non_local_dependency_set : t list -> Dependency_set.t
+  (** Returns a hash of the dependencies that are not local to the project (if any) *)
+  val non_local_dependency_hash : t list -> Dependency_hash.t option
+
+  (** Returns the name of the first non-local dependency it can find *)
+  val any_non_local_dependency_name : t list -> Package_name.t
```
rgrinberg, line 64: This function should return `option` or have an `_exn` suffix

src/dune_pkg/local_package.ml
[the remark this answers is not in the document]
rgrinberg: Btw, any folds in the opam library you could reach for to write this?

src/dune_pkg/opam_solver.ml
```diff
@@ -800,9 +800,7 @@ let reject_unreachable_packages =
             [ "name", Package_name.to_dyn name ]
         | Some (pkg : Lock_dir.Pkg.t), None -> Some (List.map pkg.depends ~f:snd)
         | None, Some (pkg : Local_package.For_solver.t) ->
-          let deps =
-            List.map pkg.dependencies ~f:(fun (d : Package_dependency.t) -> d.name)
-          in
+          let deps = Local_package.For_solver.dependency_names pkg in
```
rgrinberg, line 803: Isn't this wrong? You should not consider packages that haven't been selected for the reachability check. This would mean that you're including packages in the lock directory that are not needed for the actual build. This is a good thing to add a test for as well.

rgrinberg (changes requested):

rgrinberg:
Looked over the approach again. I agree that emitting a dependency that wasn't selected for the hash check as an example isn't so bad. It's also an improvement to have the dependency hash reflect the formula. There's an issue introduced to the reachability check though. That one should most definitely only consider effective dependency sets. Doing otherwise would include useless packages in the lock dir

src/dune_pkg/opam_solver.ml
Leonidas-from-XIV: Yes, you're completely right, thanks for pointing this out. This code needs to be changed because it uses the input to the solver as the output instead of using the solution that the solver outputs.

I'll add a test that checks that dependencies that don't get picked by the solver don't end up in the lock directory.

Leonidas-from-XIV: I'm looking what the semantics of `reachable` is and according to #10814 it is about removing `post` dependencies (and also `post-deps-solving.t` is the only test failing if I comment out the reachability test).

Since this is basically about filtering the solver output via the possible deps from the input, how about still collecting all possible packages (since they are potentially reachable) except those that are `post` and using that for reachability?

That way packages that the solver is not picking will not be part of the lock dir (as the solution just does not contain them) and `post` packages won't be picked either (as they will be filtered out by the reachability check). This makes `post` a bit special but `post` is already a magic variable in `opam`.

src/dune_pkg/local_package.ml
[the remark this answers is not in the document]
Leonidas-from-XIV: Good point. Yes, ended up doing exactly that. Also given `"foo" | "bar"` and `"bar" | "foo"` lead to different solutions.

src/dune_pkg/local_package.ml
Leonidas-from-XIV: That code just went away completely, fortunately.

src/dune_pkg/local_package.ml
Leonidas-from-XIV: Gone now.

Leonidas-from-XIV:
> That one should most definitely only consider effective dependency sets. Doing otherwise would include useless packages in the lock dir

In my tests this seems to work, as the reachability is tested by filtering over a solver solution, so if the solver doesn't pick it, it isn't ever checked for reachability. That works even in the presence of `post` dependencies, the `post` dependency tests all pass.

src/dune_pkg/opam_solver.ml
rgrinberg: Added a test showing an unnecessary post dependencies being eliminated.

> how about still collecting all possible packages (since they are potentially reachable) except those that are post and using that for reachability?

I think it's just simpler to do this without any hacks. We've already resolved the `or` deps when writing the dependency list in the lock directory. Might as well just use that here instead of the formula

src/dune_pkg/dependency_formula.ml
```diff
@@ -0,0 +1,168 @@
...
+    let l = condition_to_sexp l in
+    let r = condition_to_sexp r in
+    Dune_sexp.List [ Dune_sexp.atom "and"; l; r ]
+  | Or (l, r) ->
+    let l = condition_to_sexp l in
+    let r = condition_to_sexp r in
+    Dune_sexp.List [ Dune_sexp.atom "or"; l; r ]
+;;
+
+let rec to_sexp v =
```
rgrinberg, line 91: A bit of a shakem to introduce this much boilerplate. I just noticed we have some converters in `Opam_dyn`. How about we try using that instead of the sexp?

src/dune_pkg/opam_solver.ml
rgrinberg: In particular, I'm thinking of just using this snippet to evaluate the formula:

```
      Resolve_opam_formula.filtered_formula_to_package_names
        ~with_test:false
        (add_self_to_filter_env opam_package (Solver_env.to_env solver_env))
        version_by_package_name
        what
```

We already do this for external packages. No reason why we can't do it for local packages for the reachability check. You might want to wait until #11022 is merged though, as I'm going to move this snippet somewhere where it can be shared.

src/dune_pkg/dependency_formula.ml
Leonidas-from-XIV: That works too, I've replaced it with converting to `Dyn.t` and hashing that. I currently hash it via `to_string` but maybe a nicer solution would be using `Dyn.hash` and converting the `int` into a hex digest?

src/dune_pkg/opam_solver.ml
Leonidas-from-XIV: Ah, I missed the comment. Ok, I'll try to evaluate the formula instead.

Leonidas-from-XIV: Looking at the documentation of `filtered_formula_to_package_names` makes me question whether this is the right thing to use:

> Given a map associating package names with the version of that package in a solution, this function evaluates an `OpamFormula.t` to a concrete list of package names all contained in the solution with versions which satisfy the formula, or an error if the the version constraints or packages in the formula can't be satisfied with the packages in the solution. If the formula contains disjunctions (the `|` operator in opam) then there may be multiple sets of packages that satisfy the formula. In this case the choice of which set to return is made arbitrarily.

This sounds to me that if the solver chooses one side of the disjunction and `filtered_formula_to_package_names` the other, this will lead to all the solver-selected legitimate dependencies being filtered out. Or, if I pass in the result from the solver as the package name map, the function will fail to find a solution.

rgrinberg:
LGTM. I just replaced the recursion with some existing opam library functions.

PR #10920, opened 2024-09-13
doc: `dune pkg` tutorial

Leonidas-from-XIV:
This lays the groundwork for an "Introduction to Dune package management" tutorial that we can point people to so they can form an understanding how to work with package mangement.

It's currently not linked and also not finished. My plan is to eventually show this:

  * Create a project
  * Add a dependency
  * Change the version that is locked
  * Remove dependency
 
To not overcomplicate things in an introduction document my plan is to use `fmt` as an example dependency, as it is fairly well-known (thus the example code isn't distracting) and also has the same module/`findlib`/`opam` name.

### DIFF
--- a/doc/tutorials/dune-package-management/dependencies.md
+++ b/doc/tutorials/dune-package-management/dependencies.md
@@ -1,0 +1,140 @@
+# Managing Dependencies
+
+The OCaml ecosystem has a wealth of third-party packages that are available for
+use. In this section we will look into how to use them with Dune.
+
+## Adding Dependencies
+
+Much like in regular projects, to add a library we need to add a dependency to
+it. For simplicity we will use the popular `fmt` library as an example, but any
+package from the [package repository](https://ocaml.org/packages) can be used.
+
+First we update the `dune-project` file to add a dependeny on the opam package.
+
+::::{dropdown} `dune-project`
+:icon: file-code
+
+:::{literalinclude} dependencies/dune-project
+:language: dune
+:emphasize-lines: 8
+:::
+
+::::
+
+After this change to our project dependencies, we need to relock dependencies
+to update our lock directory with the new packages.
+
+```sh
+$ dune pkg lock
+Solution for dune.lock:
+- base-unix.base
+- fmt.0.9.0
+- ocaml.5.2.0
+- ocaml-base-compiler.5.2.0
+- ocaml-config.3
+- ocamlbuild.0.15.0+dune
+- ocamlfind.1.9.6+dune
+- topkg.1.0.7
+```
+
+You can see a lot of new dependencies, among these `fmt`.
+
+:::{note}
+The list of packages being output includes all dependencies of your project,
+including transitive dependencies.
+:::
+
+This will take care of installing the dependencies, but we still need to add it to
+our build as a library, as usual:
+
+::::{dropdown} `dune`
+:icon: file-code
+
+:::{literalinclude} dependencies/dune
+:language: dune
+:emphasize-lines: 3
+:::
+
+Adding a library dependency to our `dune` file via the `libraries` stanza.
+
+::::
+
+This will allow us to use the `Fmt` module in our OCaml code.
+
+::::{dropdown} `test.ml`
+:icon: file-code
+
+:::{literalinclude} dependencies/test.ml
+:language: ocaml
+:emphasize-lines: 4
+:::
+
+We update the code to define an `Fmt.t` pretty-printer for the list of strings
+and then use it to print the value.
+
+::::
+
+To build it we just call `build` again.
+
+```sh
+$ dune build
+```
+
+which will download and install the new dependencies and build our project as
+before.
+
+As we see, the code works and uses `fmt` to do the pretty-printing:
+
+```sh
+$ dune exec ./test.exe
+Hello, OCaml, Rust!
+```
+
+### Dependency Constraints
+
+Packages are often only compatible with some versions of dependencies. To
+specify a version range, use the regular Dune dependency syntax
+used for opam dependencies in the `dune-project` file.
+
+::::{dropdown} `dune-project`
+:icon: file-code
+
+:::{literalinclude} dependencies/constraints
+:language: dune
+:emphasize-lines: 7-8
+:::
+
+::::
+
+This ensures the `fmt` package to install will be compatible with
+our request. These constraints will be taken into account the next time the
+package is locked:
+
+```sh
+$ dune pkg lock
+Solution for dune.lock:
+- base-unix.base
+- fmt.0.9.0
+- ocaml.5.2.0
+- ocaml-base-compiler.5.2.0
+- ocaml-config.3
+- ocamlbuild.0.15.0+dune
+- ocamlfind.1.9.6+dune
+- topkg.1.0.7
+```
+
+The version of `fmt` picked is indeed between `0.6` and `1.0`.
+
+## Removing Dependencies
+
+Given all dependencies are defined in the `dune-project` file, removing a
+dependency means to remove the dependency from the `depends` field of your
+`dune-project` and relocking the project.
+
+The new lock directory will not depend on the package anymore, and in future
+builds, the package will not be accessible as `library` anymore.
+
+:::{note}
+The removed dependency might still be part of the lock directory if some other
+dependency of your project depends on it.
+:::
--- a/doc/tutorials/dune-package-management/index.md
+++ b/doc/tutorials/dune-package-management/index.md
@@ -1,0 +1,27 @@
+---
+author: Marek Kubica
+---
+
+OCaml Package Management With Dune
+==================================
+
+:::{warning}
+Dune Package Management is not final yet and details are still subject to
+change.
+:::
+
+In this tutorial we will be looking at how to use Dune for managing project
+dependencies. This enables users to install the compiler as well as third-party
+dependencies using a single tool which takes care of building code and
+dependencies.
+
+To get started you only need Dune. Head to {doc}`setup` to begin the setup.
+
+:::{toctree}
+:hidden:
+:maxdepth: 1
+setup
+dependencies
+pinning
+repos
+:::
--- a/doc/tutorials/dune-package-management/pinning.md
+++ b/doc/tutorials/dune-package-management/pinning.md
@@ -1,0 +1,50 @@
+# Pinning Projects
+
+When Dune is looking up packages to lock, it uses the (pre)configured OCaml
+package repositories. However it is also possible to manually specify the
+sources for packages; for example, if the package is not released in a package
+repository. This is called "pinning."
+
+## Installing Packages From a Pin
+
+To pin a package, a new `pin` has to be declared in the `dune-project` file.
+
+::::{dropdown} `dune-project`
+:icon: file-code
+
+:::{literalinclude} pinning/dune-project
+:language: dune
+:emphasize-lines: 4-6
+:::
+
+This will create a pin on the `fmt` package and use the specified Git repository
+URL to retrieve the sources.
+
+::::
+
+The next time the package is locked, Dune will use this repository instead of
+the information from the selected package repositories.
+
+```sh
+$ dune pkg lock
+Solution for dune.lock:
+- base-unix.base
+- fmt.dev
+- ocaml.5.0.0
+- ocaml-base-compiler.5.0.0
+- ocaml-config.3
+- ocamlbuild.0.15.0+dune
+- ocamlfind.1.9.6+dune
+- topkg.1.0.7
+```
+
+Unlike previously, the version of the `fmt` library that is picked is `dev`, to
+signify a development version.
+
+The next build will check out the sources from that repository instead of
+downloading the release tarball:
+
+```sh
+$ dune exec ./test.exe
+Hello, OCaml, Rust!
+```
--- a/doc/tutorials/dune-package-management/repos.md
+++ b/doc/tutorials/dune-package-management/repos.md
@@ -1,0 +1,52 @@
+# Custom Repositories
+
+By default when locking package versions, Dune looks up packages from two
+sources:
+
+  1. The upstream, community maintained `opam-repository` at
+     [ocaml/opam-repository](https://github.com/ocaml/opam-repository) for most
+     packages
+  2. An overlay repository with patched software to allow usage in Dune at
+     [ocaml-dune/opam-overlays](https://github.com/ocaml-dune/opam-overlays)
+
+To change the presets, the `dune-workspace` file has to be edited (and created
+if it didn't exist):
+
+::::{dropdown} `dune-workspace`
+:icon: file-code
+
+:::{literalinclude} repos/dune-workspace
+:language: dune
+:::
+
+::::
+
+In this case, we want to select a specific revision of the community repository
+instead of always using the most recent one, as it would do by default. We
+define a new repository and configure the lock directory to use this
+repository.
+
+When relocking the dependencies, the list of packages that are found as
+dependencies changes accordingly:
+
+```sh
+$ dune pkg lock
+Solution for dune.lock:
+- base-unix.base
+- fmt.0.9.0
+- ocaml.5.0.0
+- ocaml-base-compiler.5.0.0
+- ocaml-config.3
+- ocamlbuild.0.15.0+dune
+- ocamlfind.1.9.6+dune
+- topkg.1.0.
+```
+
+Compared to before, the OCaml compiler version is older, which shows
+that we did indeed pick an older version of the package repository for locking.
+
+:::{note}
+This feature can also be used to make sure the locked dependencies are
+reproducible, as fixing all the package repository versions will lead to
+equivalent locking results.
+:::
--- a/doc/tutorials/dune-package-management/setup.md
+++ b/doc/tutorials/dune-package-management/setup.md
@@ -1,0 +1,109 @@
+# Setting Up Package Management With Dune
+
+The idea of package management with Dune has been as unobtrusive as
+possible. Thus most projects can easily be built with just the minimum of
+changes.
+
+In this tutorial we will create a simple project to use the integrated package
+management feature for the very first time.
+
+## Declare Dependencies
+
+The best way to work with the package management is to declare your
+dependencies in the `dune-project` file.
+
+::::{dropdown} `dune-project`
+:icon: file-code
+
+:::{literalinclude} setup/dune-project
+:language: dune
+:emphasize-lines: 6-7
+:::
+
+We define a project called `test` and define that to build it we need an OCaml
+compiler that is at least version 4.14.
+
+This is exactly the same information that is used to generate opam files using
+the `generate_opam_files` stanza as described in
+{doc}`/howto/opam-file-generation`.
+
+::::
+
+::::{dropdown} `test.ml`
+:icon: file-code
+
+:::{literalinclude} setup/test.ml
+:language: ocaml
+:::
+
+To show that the build works, this simple program will be built and executed.
+
+::::
+
+
+::::{dropdown} `dune`
+:icon: file-code
+
+:::{literalinclude} setup/dune
+:language: dune
+:::
+
+To declare our module an executable we need a little bit of configuration, so we
+just define the module as an executable.
+
+::::
+
+After our project skeleton is set up, we can proceed to the next step.
+
+## Locking Dependencies
+
+After declaring the dependencies, you will need to tell Dune which package
+versions to use for your project. This is done by creating a lock directory.
+This is easily done with a new Dune command:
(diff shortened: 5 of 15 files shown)


doc/tutorials/dune-package-management/concepts.md
```diff
@@ -0,0 +1,40 @@
+# Concepts
+
+## Motivation
```
leostera, line 3: I don't think we need to motivate package management, but we could have a section that explains the history of this in the ecosystem.

Newcomers and most people that want to get started with it don't _need_ to know what opam is, just how to use package management with dune.

doc/tutorials/dune-package-management/setup.md
```diff
@@ -0,0 +1,65 @@
...
+
+```dune
+(lang dune 3.17)
+(name test)
+
+(generate_opam_files true)
+
+(package
+  (name test)
+  (depends
```
leostera, line 23: We should explain that dependencies go in `depends` and what the format is with a examples – I've seen enough people fumble with this already.

Some examples would be:

* For the latest version, just use the package name: `cmdliner`
* For an exact version, use `(cmdliner (= 1.2.3))`
* For a constraint, use `(cmdliner (>= 1.2.0))`

doc/tutorials/dune-package-management/setup.md
```diff
@@ -0,0 +1,65 @@
...
+    (ocaml (>= 4.14))))
+```
+
+## Create lock directory
```
leostera, line 27:
```suggestion
## Locking dependencies
```

The fact that its a lock dir is an implementation detail and the intent the user has here is not to create a dir, but to lock the version of the dependencies they need.

doc/tutorials/dune-package-management/concepts.md
Leonidas-from-XIV: I think describing why we're doing this and why this is an improvement is important, given this tutorial is not only meant for newcomers but also for existing users of OCaml that use OPAM.

doc/tutorials/dune-package-management/setup.md
Leonidas-from-XIV: I think that's a good point! I've added a section on this in a later part.

Leonidas-from-XIV:
I've reworked it a bit and improved the structure to make it less of a work in progress and more something that can be read. As such it would be nice to have your input wrt. to the wording, the markup or the content.

In generally I tried to keep it vague enough to apply to `main` with or without the developer preview features turned on; this way it can be part of the documentation of both regular Dune as well as linked from the developer preview info page without creating confusion (hopefully).

samoht:
An important concept (useful to existing users, not for newcomers) is that `dune pkg` is (or will be) able to build any package even if they are not packaged with Dune if they follow a few simple rules (mostly, they need to be relocatable). This means we do not mandate all the dependencies to use `dune`, nor do we need to patch existing packages (like `opam-monorepo` used to do) to build with Dune.

I think it would also be useful to explain this in a bit more detail - i.e. the way Dune see a package is separated into a few "macro" rules (that can be cached as all the other Dune rules): fetching and untaring the sources, running the opam build instruction in a "black box" sandbox with the right environment setup, etc. This is all possible as we re-use `opam-lib` infrastructure as much as possible (to download the package, to setup the required env, etc). 

You could also explain why we are using 0install solver (instead of re-using the default opam solver): we always find a solution from an empty set of installed packages (while the opam solver tries to minimize recompilation as it comes from the Debian ecosystem) -- and 0install-solver is optimised for this (with slightly worse error messages unfortunately).

etc, etc. Ie. I do think explaining the mental model of how things actually work is helpful to convince experienced opam/dune users that this is going in the right direction: i.e. there are many more things to add (including stop looking at the packages being built with Dune as a black box) but the architecture is pretty solid (which I believe it is) to allow all of this with enough time and people working on it.

emillon:
Thanks. Without going into details, I recommend reading https://diataxis.fr/tutorials/ - this is the compass we're using for our docs.

I think it's useful to have a tutorial for dune package management, and its goal should be that people that follow it learn by doing. The rest is important, but can be put in a different place in the documentation. As a litmus test, I think that everything we would show in a demo should be part of the tutorial. This will make the tutorial leaner and easier to follow. The resulting tutorial will be linear: set up, lock, build, add a dependency, build, remove a dependency, etc.

In my mind there are 3 types of content here:
- the "concepts" page is explanatory. I agree it's not useful in a tutorial. But it can be a great "Why is Dune dealing with package management?" doc in the explanations section.
- the "common dependency constraints" examples are useful for some how-to material. This is useful for users that will use that in the context of their project. How about a "How-to: add an external dependency"? this is exactly the workflow we'll simplify by adding an `add`, so this justifies a dedicated howto
- the linear tutorial itself

Hope that makes sense. I'll add notes in the doc

doc/tutorials/dune-package-management/index.md
```diff
@@ -0,0 +1,29 @@
+---
+author: Marek Kubica
+orphan: true
```
emillon, line 3: the tutorial is linked in a toctree, so it's not an orphan. you can remove this line (sphinx doesn't warn about this)

doc/tutorials/dune-package-management/setup.md
```diff
@@ -0,0 +1,197 @@
+# Setting up Package Management with Dune
+
+The idea of package management with Dune has been to be as unobtrusive as
+possible. Thus most projects can easily be built with just the minimum of
+changes.
+
+In this tutorial we'll create a simple project to use the OCaml package
+management for the very first time.
```
emillon, line 8: In the goal of reproducibility, we should make it clear how to get a dune binary that can be used in this tutorial (I assume it doesn't work with a released version), or at least clarify that it doesn't work with the dune that people have on their machines.

doc/tutorials/dune-package-management/setup.md
```diff
@@ -0,0 +1,197 @@
...
+$ dune build
+```
+
+which will download and install the new dependencies and build our project as
+before.
+
+As we see, the code works and uses `fmt` to do the prettyprinting:
+
+```sh
+$ dune exec ./test.exr
```
emillon, line 187:
```suggestion
$ dune exec ./test.exe
```

doc/tutorials/dune-package-management/setup.md
```diff
@@ -0,0 +1,197 @@
...
+- ocaml-config.3
+- ocamlbuild.0.15.0+dune
+- ocamlfind.1.9.6+dune
+- topkg.1.0.7
+```
+
+You can see a lot of new dependencies, among these `fmt`.
+
+:::{note}
+The list pf packages being output includes all dependencies of your project,
```
emillon, line 140:
```suggestion
The list of packages being output includes all dependencies of your project,
```

doc/tutorials/dune-package-management/concepts.md
```diff
@@ -0,0 +1,53 @@
...
+## Improvements
+
+Dune changes the imperative process of managing an OCaml installation (switch),
+installing, updating and uninstalling packages with a declarative process where
+all the dependencies are declared in one location and then kept in sync
+automatically.
+
+The advantages of this process are as follows:
+
+  - No need to manually update the opam-repository
```
samoht, line 26: Looking at that list without giving more context, it's unclear why any of these items are negative. 

So you can probably turn that list into a more positive feature list -- ie. focus on the positive things that `dune pkg` allows:

- automatic update of opam-repository by default, but can still control the exact opam-repo version by specifying the commit hash in `dune-project` for extra reproducibility
- package (and compilers!) are now cached - giving the illusion of having one local switch per project, but with maximal sharing (like a global switch on steroids)
- edit a file to record new dependencies (instead of using the CLI) -- this is super helpful when mixed with dune's watch mode to automatically install packages that the projects need.
- only one file to edit (`dune-project`) which becomes the source of truth -- all the rest: the opam files, the build environment, etc. is automatically kept in sync

(ie. I've tried to explain why each of your item is now a better UX with `dune pkg` instead of listing a list of opam features without context)

doc/tutorials/dune-package-management/concepts.md
rgrinberg: Maybe the comparison to opam should be a separate section?

doc/tutorials/dune-package-management/setup.md
```diff
@@ -0,0 +1,197 @@
...
+The idea of package management with Dune has been to be as unobtrusive as
+possible. Thus most projects can easily be built with just the minimum of
+changes.
+
+In this tutorial we'll create a simple project to use the OCaml package
+management for the very first time.
+
+## Declare dependencies
+
+The best way to work with the package management is to declare your
```
maiste, line 12: Maybe a link to how to declare opam file in Dune could add some value here?

doc/tutorials/dune-package-management/setup.md
```diff
@@ -0,0 +1,197 @@
...
+Hello, OCaml, Rust!
+```
+
+## Adding libraries
+
+Much like in regular projects to add a library we need to add a dependency to
+it. For simplicity we'll use the popular `fmt` library as an example, but any
+library on the OPAM repository can be used.
+
+First we update the `dune-project` file to add a dependeny on the OPAM package.
```
maiste, line 109: Do we need to mention opam here? Isn't it confusing that we are learning how to get rid of opam but still talk about opam? :confused:

doc/tutorials/dune-package-management/concepts.md
Leonidas-from-XIV: Yes, sleeping over it and reading the Diátaxis document I think the right thing is to remove it from the tutorial. There is probably a point to describe the how package management works in the documentation somewhere else but it is not in the tutorial.

doc/tutorials/dune-package-management/setup.md
Leonidas-from-XIV: I think `opam` and OPAM repository and OPAM packages are different things:

  1. `opam` is the tool to install OPAM packages. I'd prefer not to use it since the tutorial shouldn't require users having access to an `opam` binary.
  2. OPAM packages are packages in the OPAM repository. I could call them "packages" or "OCaml packages" but I think "OPAM package" is for better or worse a common way to refer to them.
  3. OPAM repository is here to stay, we use it internally so if I were to call it "package repository" it would probably cause more confusion than improve things.

doc/tutorials/dune-package-management/management.md
```diff
@@ -0,0 +1,150 @@
+# Managing Packages
+
+This section will introduce some more features of the package management.
```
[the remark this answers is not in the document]
Leonidas-from-XIV: I've removed the whole file, because it felt like a grab-bag of random things, so I moved these things into the place in the tutorial where they fit the story better.

doc/tutorials/dune-package-management/concepts.md
```diff
@@ -0,0 +1,53 @@
+# Concepts
+
+## Motivation
+
+As the OCaml ecosystem is growing, installing dependencies has become more and
+more common. As such the need for package management arose and around 2012
+[opam](https://opam.ocaml.org/) was created. `opam` enabled the creation of a
+central repository of packages called `opam-repository` which contains all
+published OCaml packages. This repository has over the years grown to about
+5000 distinct packages today.
```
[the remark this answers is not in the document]
Leonidas-from-XIV: I think it is a nice idea, I didn't add a link because I didn't know what the "canonical" source for opam-repository was. I think https://ocaml.org/packages is good, but the section is gone. If I add something like this I'll make sure to add that link :+1:

doc/tutorials/dune-package-management/concepts.md
Leonidas-from-XIV: I think they are negative because these are all the things that make me go "aww, finally, that's gonna be so nice".

The section is removed now because it's not really the right thing for a tutorial. When we add a reference on dune package management we should definitely write it in the way you propose.

doc/tutorials/dune-package-management/setup.md
maiste: It makes sense! Thanks for the clarification :+1:

doc/tutorials/dune-package-management/index.md
Leonidas-from-XIV: It was an orphan at first but then I realized that without linking it up to the toctree it wasn't really feasible to jump between sections. Fixed.

doc/tutorials/dune-package-management/concepts.md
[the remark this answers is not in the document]
Leonidas-from-XIV: I've changed my mind on this and agree. It's gone now, we can have an in-depth explanation how things work somewhere else in the documentation (and probably also update the reference to include `pins` and `lock_dir` etc).

doc/tutorials/dune-package-management/management.md
```diff
@@ -0,0 +1,150 @@
...
+:emphasize-lines: 7-8
+:::
+
+::::
+
+This will require that `fmt` will never be updated to version `1.0`. These
+constraints will be taken into account the next time the package is locked:
+
+```sh
+$ dune pkg lock
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Agreed; I believe this is outdated now since the section on versions are in the same page now.

doc/tutorials/dune-package-management/setup.md
Leonidas-from-XIV: Good point, I originally wanted to add it but couldn't figure out the syntax. But after asking @emillon how to do it, it's linked now :+1:

Leonidas-from-XIV:
> I do think explaining the mental model of how things actually work is helpful to convince experienced opam/dune users that this is going in the right direction

I agree as well, but after thinking about it I realized that these should be two different types documents. This PR is a tutorial to describe how to get things going, whereas the mental model document would be describing how things work.

To quote the Diátaxis document:

> All learning moves in one direction: from the concrete and particular, towards the general and abstract. The latter will emerge from the former.

(Which also matches my experience learning e.g. `git`). I expect that a lot of more experienced people will be interested in how it works under the hood, so I think it will make sense to create a second PR after this to explain what is happening under the hood when using these commands.

doc/tutorials/dune-package-management/repos.md
```diff
@@ -0,0 +1,50 @@
...
+
+  1. The upstream, community maintained opam repository at
+     [ocaml/opam-repository](https://github.com/ocaml/opam-repository) for most
+     packages
+  2. An overlay opam repository with patched software to allow usage in Dune at
+     [ocaml-dune/opam-overlays](https://github.com/ocaml-dune/opam-overlays)
+
+This list is built-in but can be configured. To change the repositories to be
+used the `dune-workspace` file has to be edited (and created if it didn't
+exist):
```
christinerose, line 13:
```suggestion
This list is built-in but can be configured. To change the repositories, the `dune-workspace` file has to be edited (and created if it didn't
exist):
```

is "to be used" necessary here? Could it be inferred that the repositories they need to change are the ones they want to use?

Leonidas-from-XIV: I think it isn't necessary. I mostly added it because configuration consists of adding them and then configuring them to be used, but people will surely figure that out.

doc/tutorials/dune-package-management/dependencies.md
```diff
@@ -0,0 +1,140 @@
...
+:icon: file-code
+
+:::{literalinclude} dependencies/constraints
+:language: dune
+:emphasize-lines: 7-8
+:::
+
+::::
+
+This will ensure that the `fmt` package to be installed will be compatible with
```
christinerose, line 109:
```suggestion
This ensures the `fmt` package to install will be compatible with
```
Can we tighten this sentence and retain accuracy?

doc/tutorials/dune-package-management/dependencies.md
```diff
@@ -0,0 +1,140 @@
...
+The version of `fmt` picked is indeed between `0.6` and `1.0`.
+
+## Removing Dependencies
+
+Given all dependencies are defined in the `dune-project` file, removing a
+dependency means to remove the dependency from the `depends` field of your
+`dune-project` and relocking the project.
+
+The new lock directory will not depend on the package anymore and in future
+builds the package will not be accessible as `library` anymore.
```
christinerose, line 135:
```suggestion
The new lock directory will not depend on the package anymore, and in future
builds, the package will not be accessible as `library` anymore.
```
Consider adding these commas for clarity. I had to read the sentence 5x to understand, as I kept reading "in future" as "in the future" rather than "in future builds"

doc/tutorials/dune-package-management/pinning.md
```diff
@@ -0,0 +1,49 @@
+# Pinning Projects
+
+When Dune is looking up packages it uses opam repository to look up packages.
```
christinerose, line 3: Should this maybe be `opam` repository or `opam-repository`?

doc/tutorials/dune-package-management/repos.md
```diff
@@ -0,0 +1,49 @@
+## Custom Repositories
+
+In normal usage Dune looks up packages from two sources:
+
+  1. The upstream, community maintained opam repository at
```
christinerose, line 5: If this is supposed to be `opam-repository`, I will go through and change them all.

doc/tutorials/dune-package-management/setup.md
```diff
@@ -0,0 +1,110 @@
...
+:::{note}
+The versions that get locked might be different from this tutorial, as we only
+specified the lower bound of `ocaml`; barring any additional configuration, Dune
+will pick the newest possible version for each dependency.
+:::
+
+## Build Project
+
+The build the project you can just use the regular Dune commands to build your
+project.
```
christinerose, line 85:
```suggestion
To build the project, you can just use the regular Dune commands.
```

doc/tutorials/dune-package-management/repos.md
Leonidas-from-XIV: I replaced it with "package repository" given we link to `opam-repository` right afterwards, that way it should be less repetitive while minimizing confusion (I hope).

doc/tutorials/dune-package-management/setup.md
christinerose: No worries! It happens to all of us!

doc/tutorials/dune-package-management/repos.md
```diff
@@ -0,0 +1,52 @@
...
+- ocaml.5.0.0
+- ocaml-base-compiler.5.0.0
+- ocaml-config.3
+- ocamlbuild.0.15.0+dune
+- ocamlfind.1.9.6+dune
+- topkg.1.0.
+```
+
+Compared to before, the version of the OCaml compiler is older, which shows
+that for locking we did indeed pick an older version of the package repository.
```
christinerose, line 46:
```suggestion
Compared to before, the OCaml compiler version is older, which shows
that we did indeed pick an older version of the package repository for locking.
```
Is this still accurate?

Leonidas-from-XIV: Yes, thanks!

ElectreAAS (approved):
I love how simple it all looks!
LGTM :+1:

leostera (approved):
LGTM 🚀 We can refine/iterate as we start getting feedback 🙏🏼

doc/tutorials/dune-package-management/dependencies.md
```diff
@@ -0,0 +1,140 @@
+# Managing Dependencies
+
+The OCaml ecosystem has a wealth of third-party packages that are available for
+use. In this section we will look into how to use them with Dune.
+
+## Adding Dependencies
+
+Much like in regular projects, to add a library we need to add a dependency to
+it. For simplicity we'll use the popular `fmt` library as an example, but any
```
maiste, line 9:
```suggestion
it. For simplicity we will use the popular `fmt` library as an example, but any
```

maiste (approved):
Being nitpicking here, but all good otherwise :+1:

doc/tutorials/dune-package-management/setup.md
```diff
@@ -0,0 +1,109 @@
+# Setting Up Package Management With Dune
+
+The idea of package management with Dune has been as unobtrusive as
+possible. Thus most projects can easily be built with just the minimum of
+changes.
+
+In this tutorial we'll create a simple project to use the OCaml package
+management for the very first time.
```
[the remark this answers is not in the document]
Leonidas-from-XIV: I've reworded it a bit to avoid "dune package" as a term, so people don't conflate opam packages with dune packages.

doc/tutorials/dune-package-management/dependencies.md
Leonidas-from-XIV: Thanks, applied!

christinerose (approved):

PR #10923, opened 2024-09-15
Fix error/warnings for config file

H-ANSEN:
The following pull request seeks to resolve the issue listed below!

## Expected Behavior
The configuration file for dune makes use of the same version numbering scheme as dune-project files *`(lang dune <ver>)`*. When attempting to use dune with a configuration file that has a mismatch between a stanza and it's supported dune version an error indicating this mismatch in the configuration file should be displayed:
```cmd
File "dune/config", line 2, characters 0-15:
2 | (cache enabled)
Error: 'cache' is only available since version 2.0 of the dune language.
Please update your dune config file to have (lang dune 2.0).
```
## Actual Behavior
The actual recommends changing the version in the `dune-project` file as opposed to the users config file:
```cmd
File "dune/config" line 2, characters 0-15:
2 | (cache enabled)
Error: 'cache' is only available since version 2.0 of the dune language.
Please update your dune-project file to have (lang dune 2.0).
```
## Reproduction
```cmd
printf "(lang dune 1.0)\n(cache enabled)" > dune-config
dune init proj test --config-file=dune-config
```

### DIFF
--- a/doc/changes/10923.md
+++ b/doc/changes/10923.md
@@ -1,0 +1,3 @@
+- Fix the file referred to in the error/warning message displayed due to the
+  dune configuration version not supporting a particular configuration
+  stanza in use. (#10923, @H-ANSEN)
--- a/src/dune_sexp/syntax.ml
+++ b/src/dune_sexp/syntax.ml
@@ -250,18 +250,22 @@ module Key = struct
 end
 
 module Error_msg = struct
-  let since t ver ~what =
-    let lang_or_using = if t.name = "dune" then "lang" else "using" in
+  let fmt_error_msg t ver ~what ~file =
+    let lang_or_using name = if name = "dune" then "lang" else "using" in
     Printf.sprintf
-      "%s is only available since version %s of %s. Please update your dune-project file \
-       to have (%s %s %s)."
+      "%s is only available since version %s of %s. Please update your %s file to have \
+       (%s %s %s)."
       what
       (Version.to_string ver)
       t.desc
-      lang_or_using
+      file
+      (lang_or_using t.name)
       t.name
       (Version.to_string ver)
   ;;
+
+  let since t ver ~what = fmt_error_msg t ver ~what ~file:"dune-project"
+  let since_config t ver ~what = fmt_error_msg t ver ~what ~file:"dune config"
 end
 
 module Error = struct
@@ -520,16 +524,20 @@ let renamed_in t ver ~to_ =
     Error.renamed_in loc t ver ~what ~to_
 ;;
 
-let since ?what ?(fatal = true) t ver =
+let since_fmt ?(fatal = true) ~fmt t ver loc =
   let open Version.Infix in
   let* current_ver = get_exn t in
   if current_ver >= ver
   then return ()
-  else
-    let* loc, what_ctx = desc () in
-    let what = Option.value what ~default:what_ctx in
+  else (
     if fatal
-    then Error.since loc t ver ~what
-    else User_warning.emit ~loc [ Pp.text (Error_msg.since t ver ~what) ];
-    return ()
+    then User_error.raise ~loc [ Pp.text (fmt t ver) ]
+    else User_warning.emit ~loc [ Pp.text (fmt t ver) ];
+    return ())
+;;
+
+let since ?what ?(fatal = true) t ver =
+  let* loc, what_ctx = desc () in
+  let what = Option.value what ~default:what_ctx in
+  since_fmt ~fatal ~fmt:(Error_msg.since ~what) t ver loc
 ;;
--- a/src/dune_sexp/syntax.mli
+++ b/src/dune_sexp/syntax.mli
@@ -29,7 +29,14 @@ end
 type t
 
 module Error_msg : sig
+  (** [since t ver what] formats an error string indicating that the syntax
+      [t] described by [what] is only available since [ver] of the dune-project
+      file. *)
   val since : t -> Version.t -> what:string -> string
+
+  (** Like [since] but formats an error message relating to the dune config file
+      rather than the dune-project file. *)
+  val since_config : t -> Version.t -> what:string -> string
 end
 
 module Error : sig
@@ -71,6 +78,9 @@ val create
 (** Return the name of the syntax. *)
 val name : t -> string
 
+(** Indicate the location and kind of value being parsed *)
+val desc : unit -> (Loc.t * string, 'a) Decoder.parser
+
 (** Check that the given version is supported and raise otherwise. *)
 val check_supported : dune_lang_ver:Version.t -> t -> Loc.t * Version.t -> unit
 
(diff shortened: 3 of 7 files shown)


doc/changes/10923.md
```diff
@@ -0,0 +1,3 @@
+- Fix error/warning messages displayed when the specified dune configuration
```
Leonidas-from-XIV, line 1: I would say what it fixes is the file that the error refers to, maybe worth being explicit about it in the changelog.

src/dune_sexp/syntax.ml
```diff
@@ -250,15 +250,28 @@ module Key = struct
...
       what
       (Version.to_string ver)
       t.desc
-      lang_or_using
+      (lang_or_using t.name)
+      t.name
+      (Version.to_string ver)
+  ;;
+
+  let since_config t ver ~what =
```
Leonidas-from-XIV, line 267: Is there a point in having this be two separate functions? To me it seems like what it does is

```ocaml
sprintf "%s is only available since version %s of %s. Please update your %s file to have (%s %s %s)."

i.e. the file to be mentioned can also just be an argument.
```

src/dune_sexp/syntax.mli
```diff
@@ -71,6 +75,8 @@ val create
 (** Return the name of the syntax. *)
 val name : t -> string
 
+val desc : unit -> (Loc.t * string, _) Decoder.parser
```
Leonidas-from-XIV, line 78: Can you add the type of `_` here as well? `_` is useful for writing code but looking at an `mli` file `_` is really unhelpful. Also a short docstring would be appreciated.

src/dune_sexp/syntax.mli
```diff
@@ -98,6 +104,16 @@ val renamed_in : t -> Version.t -> to_:string -> (unit, _) Decoder.parser
...
 
+(** Like [since] but accepts a function [fmt] allowing custom formatting of the
+    entire error/warning message. See [Error_msg] for format functions. *)
+val since_fmt
+  :  ?fatal:bool
+  -> fmt:(t -> Version.t -> string)
+  -> t
+  -> Version.t
+  -> Loc.t
+  -> (unit, _) Decoder.parser
```
Leonidas-from-XIV, line 115: Likewise here for `_`.

test/blackbox-tests/test-cases/config/config-version.t
```diff
@@ -0,0 +1,56 @@
...
+Attempt to initialize a new project while an invaild stanza due to versioning
+exists in the config.
+
+  $ dune init proj test --config-file=dune-config
+  File "$TESTCASE_ROOT/dune-config", line 2, characters 0-15:
+  2 | (cache enabled)
+      ^^^^^^^^^^^^^^^
+  Error: 'cache' is only available since version 2.0 of the dune language.
+  Please update your dune config file to have (lang dune 2.0).
+  [1]
```
Leonidas-from-XIV, line 23: Can you add a test that adds `(lang dune 2.0)` and shows this works? There is probably a test that makes sure of it, but it is also useful to show that it works.

test/blackbox-tests/test-cases/config/config-version.t
```diff
@@ -0,0 +1,56 @@
...
+
+Swap out the invaild stanza for another and try again.
+
+  $ sed -i -e '$s/.*/(cache-check-probability 0.5)/' dune-config
```
Leonidas-from-XIV, line 27: This can be done with `echo "(cache-check-probability 0.5)" >> dune-config` which is a bit simpler to read and achieves a similar result.

test/blackbox-tests/test-cases/config/config-version.t
```diff
@@ -0,0 +1,56 @@
...
+  $ dune init proj test --config-file=dune-config
+  File "$TESTCASE_ROOT/dune-config", line 2, characters 0-29:
+  2 | (cache-check-probability 0.5)
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  Error: 'cache-check-probability' is only available since version 2.7 of the
+  dune language. Please update your dune config file to have (lang dune 2.7).
+  [1]
+
+Swap out the invaild stanza for another and try again.
```
Leonidas-from-XIV, line 36: Correct me if I'm misunderstanding something but past this point I don't see a purpose of the other tests. We already established that it displays the right file when it errors and it works when deeding the advice and adding a new error shows the error message again with the right error location.

I'd remove the rest of the tests as they don't exercise any more interesting edge cases (and running extra tests takes time that could be better spent).

Leonidas-from-XIV (changes requested):
Thanks for the contribution, I appreciate the extensive tests. The code is fine, see my comments on what would be worth improving to get it merged.

src/dune_sexp/syntax.ml
H-ANSEN: Yea! Everything here can be super simplified if we take the file name as an argument. When deciding how to fix this my initial thought was to alter the `since` function as follows:
```ocaml
let since ?(file = "dune-project") ?(fatal = true) ?what t ver = ...
```
Due to the fact that `since` is used all over the place I felt that it might be worth proposing and implementation which preserved the original signature and also provided a way to control the entire error message. If you think that altering the signature as shown above would be okay I'd be happy to move towards this type of representation, this would also have the added benefit simplifying/resolving some of the other changes you mention below!

test/blackbox-tests/test-cases/config/config-version.t
H-ANSEN: Your correct here, my intention was to simply cover multiple stanzas with differing supported versions of dune but 2 tests does cover this case.

test/blackbox-tests/test-cases/config/config-version.t
H-ANSEN: Here I am using sed to **swap** out the first invalid stanza *'(cache enabled)'*. `>>` has the effect of appending rather than swapping so the resulting error would be caused on the same *'(cache enabled)'* stanza from the first test instead of the new *cache-check-probability* stanza!

src/dune_sexp/syntax.ml
H-ANSEN: Your correct! It's functionally the same thing just with a different string describing a different file. I can condense this into a single function.

test/blackbox-tests/test-cases/config/config-version.t
H-ANSEN: Ahh sorry, I did not understand what you initially meant, thank you for clarifying! I've cleaned up the test cases let me know what you think.

Leonidas-from-XIV (changes requested):
Thanks for the revisions, the tests look pretty good now. I have one clarification and one small thing to make the tests safer, after that I think it's good to go!

test/blackbox-tests/test-cases/config/config-version.t
Leonidas-from-XIV: They have a good flow demonstrating the issue with the error and the fix, I like it.

Leonidas-from-XIV (approved):
Looks good, thanks!

PR #10926, opened 2024-09-16
chore(issue): package management report

maiste:
As we are starting to promote Dune Package Management to get some feedback from users (to improve the experience), there is a risk to have bug report in Dune. This PR is a proposition to filter the result by having a dedicated form for Package Management issue. It would directly assign the right label. WDYT?

### DIFF
--- a/.github/ISSUE_TEMPLATE/package_management.md
+++ b/.github/ISSUE_TEMPLATE/package_management.md
@@ -1,0 +1,21 @@
+---
+name: Package Management bug report
+about: Report a problem related to Dune Package Management
+title: '[Package Management]'
+labels: 'package management build failures'
+assignees: ''
+---
+
+## Specification
+
+<!-- output of `dune --version` --> 
+- Version of `dune`: 
+- Operating system (distribution and version):
+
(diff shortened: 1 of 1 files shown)


emillon:
2 remarks about that:
- generally I think it's better to wait until triage is an issue because of the volume or missing info (we'll know what info to ask for)
- it's now possible to define a form instead of just markdown sections if you prefer, see https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

maiste:
I agree with your first remark. However, as Dune Package Management will be soon in `ocaml.org`, I wanted to make sure we don't have more triage work to do than expected. If people are coming from the Dune Preview (from ocaml.org), I wouldn't expect this to be mixed with the general dune maintenance :confused: 

About the second point, I have seen it but as it was in beta I was a bit cold to introduce it in Dune. WDYT,

maiste:
After some reflexion, as we already have a bug report, it's easier to do the triage on our own with the package management label.

PR #10928, opened 2024-09-16
refactor(pkg): tweak experimental flag names

rgrinberg:
If only have "enable" flags, it will be impossible to disable existing flags once we change the defaults. With this change, we can tweak the defaults without adding/removing more flags.


ElectreAAS (approved):
These changes make sense imo, and make it clearer for users as well.
Approving :)

anmonteiro (commented):
I believe this PR needs to change the `flake.nix` file, too, where we build a version of dune with these features enabled

maiste (approved):

PR #10931, opened 2024-09-17
[fix] [directory targets] Fix promotion of directory targets in some cases

### LINKED ISSUE #10609: [regression] promotion of directory targets has trouble with directories starting in @
Dear dune devs,

updating to Dune 3.15, I've found the following problem with this simple rule (that used to work):
```
(rule
 (targets
   (dir node_modules))
 (deps package.json)
 (mode promote)
 (action
  (bash "npm i")))
```

it seems in 3.15.3, dune chokes with:
```
Error: Cannot promote files to "node_modules/@babel/parser".
Reason: opendir(node_modules/@babel/parser): No such file or directory
-> required by _build/default/node_modules
```
File is correctly present in `_build` dir.

Any package.json file that depends on a package starting with `@` seems to do the trick, for example:
```
{
  "name": "test",
  "version": "0.17.0",
  "type": "module",
  "description": "NPM test",
  "main": "index.js",
  "dependencies": {
    "@octokit/core": "^4.2.1"
  }
}
```

Bisect points to b1c339b868e00fd5a1a3e72db90e57b4f3285208 , I've verified this to be the culprit cc @rleshchinskiy

ejgallego:
When the promoted directory has sub-directories with only other sub-directories inside, the map will miss that directory and it will not be properly re-created, failing at children creation time.

I am not sure I fully follow the logic here (cc: @rleshchinskiy).

The current test case works for me (TM), if we revert the fix we get:

```
Error: Cannot promote files to "node_modules/node-cmake/node_modules/ansi-regex".
Reason: opendir(node_modules/node-cmake/node_modules/ansi-regex): No such file or directory
```

Regression in b1c339b / #9407

Closes #10609 .

### DIFF
--- a/src/dune_targets/dune_targets.ml
+++ b/src/dune_targets/dune_targets.ml
@@ -243,9 +243,7 @@ module Produced = struct
                 acc_filenames, Path.Local.Map.union_exn acc_dirs dir
               | _ -> Error (Unsupported_file (Path.Build.relative dir filename, kind)))
         in
-        if not (Filename.Map.is_empty filenames)
-        then Path.Local.Map.add_exn dirs local filenames
-        else dirs
+        Path.Local.Map.add_exn dirs local filenames
     in
     let directory root dirname =
       let open Result.O in
--- a/doc/changes/10931.md
+++ b/doc/changes/10931.md
@@ -1,0 +1,3 @@
+- Fix promotion of directory targets when the promoted directory has
+  sub-directories with only other sub-directories inside; regression
+  in b1c339b / #9407 (@ejgallego, closes #10609)
--- a/test/blackbox-tests/test-cases/directory-targets/github10609/run.t
+++ b/test/blackbox-tests/test-cases/directory-targets/github10609/run.t
@@ -1,0 +1,24 @@
+We test that a directory target with only other subdirs can be
+properly promoted.
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.16)
+  > (using directory-targets 0.1)
+  > EOF
+
+  $ cat > dune <<EOF
+  > (rule
+  >  (targets (dir node_modules))
+  >  (mode (promote (until-clean)))
+  >  (action
+  >    (progn
+  >      (system "mkdir -p node_modules/node-cmake/")
+  >      (system "touch    node_modules/node-cmake/.jshintrc")
+  >      (system "mkdir -p node_modules/node-cmake/node_modules/ansi-regex")
+  >      (system "touch    node_modules/node-cmake/node_modules/ansi-regex/index.js")
+  >    )))
+  > EOF
+
+  $ dune build node_modules
+  $ ls node_modules
+  node-cmake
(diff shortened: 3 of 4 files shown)


src/dune_targets/dune_targets.ml
```diff
@@ -243,9 +243,7 @@ module Produced = struct
                 acc_filenames, Path.Local.Map.union_exn acc_dirs dir
               | _ -> Error (Unsupported_file (Path.Build.relative dir filename, kind)))
         in
-        if not (Filename.Map.is_empty filenames)
-        then Path.Local.Map.add_exn dirs local filenames
-        else dirs
+        Path.Local.Map.add_exn dirs local filenames
```
moyodiallo, line 246: I'm not sure for this fix, I think this code is fine. The issue is coming from somewhere else.

moyodiallo (changes requested):
After digging a bit, I found the issue could come from:

https://github.com/ocaml/dune/blob/1a8b43f449cf00996281b5c73b2a0f3400817a00/src/dune_engine/target_promotion.ml#L210-L225

The function is supposed to remove the stale files and sub-directories, but at the line `224`, we are not supposed to report the error when the source directory doesn't exist which is fine anyway.

You should observe the effect replacing the line by
```ocaml
  | Error unix_error -> ()
```
but it's not end fix, because we could have other type of errors.

src/dune_targets/dune_targets.ml
ejgallego: Thanks a lot for the help. My impression is that indeed it matters that we add the binding `{ dir -> [] }` to the map, as even if the directory doesn't have files, it can have subdirs.

But I didn't track down exactly the part that makes the current code break.

ejgallego: So indeed you are correct that the crash happens at `remove_state_files_and_subdirectories`, but my impression is that we break the invariant about source representation that the above function requires.

moyodiallo: I get it, that's fair point. 

The question I have now is, are we going to allow promoting empty directories ?

ejgallego: Maybe the confusion is about what we call "empty directory", empty directories are OK not to promote in my view, however here the directories in question are not empty, they just have no files, however they do have other subdirectories, so definitively they need to be considered part of the sources.

moyodiallo: I tested the fix, it is promoting even empty directories.
This is not fixing exactly the test-case, it's fixing it indirectly by allowing an empty directory.

ejgallego: Indeed after this, empty directories are promoted. I don't see why they shouldn't be promoted?

We could optimize this away, but I am not sure it is worth it.

Note that the fix doesn't exactly allow "empty directories", it allows too, crutially, directories that have no files but are not empty, such as the test case.

moyodiallo: Looking to this further. I agree with you, this where we need to fix the issue.

moyodiallo: This is one of the solution https://github.com/moyodiallo/dune/commit/b00b92e1fbc1007736677436d3b9b56be3f089fc. 

The thing is adding all the parent directories of the collected files, it will ends up with no empty directory.

ElectreAAS:
Closing in favor of the now-merged #11226

ejgallego:
Thanks a lot @ElectreAAS for your hard work on fixing this.

PR #10936, opened 2024-09-19
fix: enabled_if with `env` variable

### LINKED ISSUE #10905: `enabled_if` field does not support environment variables on executable stanza
In the [docs](https://dune.readthedocs.io/en/stable/reference/dune/executable.html#executable), it is claimed that the field `enabled_if` behaves for executables the same as libraries. [As of dune 3.15](https://ocaml.org/changelog/2024-04-03-dune.3.15.0), `enabled_if` should allow to enable/disable a library (and therefore executable) based on environment variables.

## Expected Behavior

Given a directory with

`dune-project` file:

```
(lang dune 3.15)
(name dune-test)
(package
 (name dune-test))
```

`dune` file:

```
(executable
 (name main)
 (public_name dune_test)
 (modules main)
 (package dune-test)
 (modes exe))

(executable
 (name main_2)
 (public_name dune_test_2)
 (enabled_if (= %{env:MYVAR=disabled} enabled))
 (modules main_2)
 (package dune-test)
 (modes exe))
```

and files `main.ml` and `main_2.ml` (omitted as irrelevant), running `dune build` should result in producing either `main.exe`, or both `main.exe` and `main_2.exe`, based on the value of `$MYVAR`.

## Actual Behavior

`dune build` fails with the following error:

```
File "dune", line 11, characters 16-37:
11 |  (enabled_if (= %{env:MYVAR=disabled} enabled))
                     ^^^^^^^^^^^^^^^^^^^^^
Error: Only architecture, system, model, os_type, ccomp_type, profile,
ocaml_version, context_name and arch_sixtyfour variables are allowed in this
'enabled_if' field. Please upgrade your dune language to at least 3.15.
```

Note that `dune-project` already states `(lang dune 3.15)`. Setting the language version to 3.16 results in the same error.

Also note that the same `enabled_if` expression has the desired effect if used on `library`.

## Specifications

- Version of `dune`: 3.16.0
- Version of `ocaml`: 4.12.1
- Operating system: macOS 13.5.1, Darwin Kernel Version 22.6.0 arm64

moyodiallo:
Fixes https://github.com/ocaml/dune/issues/10905.

### DIFF
--- a/src/dune_rules/enabled_if.ml
+++ b/src/dune_rules/enabled_if.ml
@@ -56,23 +56,23 @@ let decode_value ~allowed_vars ?(is_error = true) () =
       ~override_decode_bare_literal:None
       (String_with_vars.decode_manually (fun env var ->
          let name = Dune_lang.Template.Pform.name var in
-         match Dune_lang.Template.Pform.payload var with
-         | _ when not @@ List.exists ~f:(String.equal name) common_vars_list ->
+         if List.exists ~f:(String.equal name) common_vars_list
+         then (
+           match List.assoc allowed_vars name with
+           | None ->
+             emit_warning allowed_vars is_error var;
+             Pform.Env.parse env var
+           | Some min_ver ->
+             let current_ver = Pform.Env.syntax_version env in
+             if min_ver > current_ver
+             then (
+               let loc = Dune_lang.Template.Pform.loc var in
+               let what = Dune_lang.Template.Pform.describe var in
+               Dune_lang.Syntax.Error.since loc Stanza.syntax min_ver ~what)
+             else Pform.Env.unsafe_parse_without_checking_version env var)
+         else (
            emit_warning allowed_vars is_error var;
-           Pform.Env.parse env var
-         | _ ->
-           (match List.assoc allowed_vars name with
-            | None ->
-              emit_warning allowed_vars is_error var;
-              Pform.Env.parse env var
-            | Some min_ver ->
-              let current_ver = Pform.Env.syntax_version env in
-              if min_ver > current_ver
-              then (
-                let loc = Dune_lang.Template.Pform.loc var in
-                let what = Dune_lang.Template.Pform.describe var in
-                Dune_lang.Syntax.Error.since loc Stanza.syntax min_ver ~what)
-              else Pform.Env.unsafe_parse_without_checking_version env var)))
+           Pform.Env.parse env var)))
 ;;
 
 let decode ~allowed_vars ?is_error ~since () =
--- a/src/dune_rules/enabled_if.ml
+++ b/src/dune_rules/enabled_if.ml
@@ -19,6 +19,7 @@ let common_vars_list =
   ; "ocaml_version"
   ; "context_name"
   ; "arch_sixtyfour"
+  ; "env"
   ]
 ;;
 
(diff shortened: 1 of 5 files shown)


test/blackbox-tests/test-cases/enabled_if/eif-env-vars.t
```diff
@@ -0,0 +1,31 @@
...
+  >  (modules main_2)
+  >  (package dune-test)
+  >  (modes exe))
+  > EOF
+
+  $ cat > main.ml <<EOF
+  > let () = print_string "Hello world"
+  > EOF
+
+  $ MYVAR=disabled dune exec -- dune_test
```
Leonidas-from-XIV, line 30: Can you also add a test for `MYVAR=enabled`?

src/dune_rules/enabled_if.ml
```diff
@@ -29,6 +30,7 @@ let common_vars ~since =
          match var with
          | "context_name" -> var, (2, 7)
          | "arch_sixtyfour" -> var, (3, 11)
+         | "env" -> var, (1, 14)
```
Leonidas-from-XIV, line 33: Why is this since version 1.14? Given it didn't work before, shouldn't it be 3.17 or as a bugfix to a feature in 3.15 be `(3, 15)`?

Leonidas-from-XIV (commented):
Nice to get this fixed!

src/dune_rules/enabled_if.ml
moyodiallo: This is related to `env` itself. Like `context_name` start from 2.7.

moyodiallo: surely we need a backport on 3.15 after.

rgrinberg: I don't understand why this is 1.14 either

src/dune_rules/enabled_if.ml
```diff
@@ -53,12 +55,12 @@ let decode_value ~allowed_vars ?(is_error = true) () =
     Blang.Ast.decode
       ~override_decode_bare_literal:None
       (String_with_vars.decode_manually (fun env var ->
+         let name = Dune_lang.Template.Pform.name var in
          match Dune_lang.Template.Pform.payload var with
```
rgrinberg, line 59: Why do we need this pattern match at all?

rgrinberg (changes requested):

src/dune_rules/enabled_if.ml
moyodiallo: Yes, there's no need,  I'll refactor.  I just kept it because it was there before.

src/dune_rules/enabled_if.ml
moyodiallo: My bad.  I've got a misunderstanding here, reading the error message. I was thinking it was specific to the variable itself. The `1.14` is coming from the doc, when `env` is introduced. 

```
Error: %{env:--} is only available since version <version> of the dune
  language. Please update your dune-project file to have (lang dune <version>).
```

 Sorry @Leonidas-from-XIV you're right.

moyodiallo: I add on `3.17`.

moyodiallo:
Added `env` on 3.15 dune lang to keep it consistent with the previous upgrade of `enabled_if` https://github.com/ocaml/dune/pull/10250, obviously a fix for dune 3.15.

ping @Leonidas-from-XIV @rgrinberg

Leonidas-from-XIV (approved):

Issue #5514, opened 2022-03-10
[RFC] Order-only dependencies


snowleopard:
It is sometimes useful to specify dependencies of a build action in two steps:

* Before executing the action, specify the set of its *possible dependencies* P. The set P is used only for scheduling, i.e., everything in P must be built before running the action. Such dependencies are often called *order-only dependencies*.

* After executing the action, specify the set of its *actual dependencies* A, where A must be a subset of P. These dependencies are used to determine when the action will need to be rebuilt.

Here are a couple of use-cases:

* It is common to depend on all C headers in a project (i.e. `headers/*.h`), to make sure they are up-to-date, and then compile C sources using `gcc` with the flags `-MMD -MF` to obtain the set of actual headers that were needed for compilation. 

* A recently discussed use-case in OCaml is to improve build incrementality with respect to library dependencies. For example, if a library B depends on a library A, we can use order-only dependencies to ensure that all modules of A have been built before compiling modules of B, and then use `ocamldep` to find actual dependencies of each module of B, to avoid recompiling all B's modules whenever any module of A changes.

Order-only dependencies are not new, for example, the Shake build system provides [orderOnly](https://hackage.haskell.org/package/shake-0.19.6/docs/Development-Shake.html#v:orderOnly) and [needed](https://hackage.haskell.org/package/shake-0.19.6/docs/Development-Shake.html#v:needed) primitives to specify order-only and actual dependencies, respectively. See Section 3.5 of [this paper](https://github.com/ndmitchell/shake-paper/releases/download/hs-preprint/hs-2016-preprint.pdf) for more details.

For now, this RFC only describes the idea but I'll update it with more details after we figure out how to add this feature to Dune.

bobot:
Do you have an idea of the rough API that would give?

snowleopard:
> Do you have an idea of the rough API that would give?

At the level of the build engine, we already have the `goal` primitive that allows expressing order-only dependencies in build rules: 

https://github.com/ocaml/dune/blob/4bc7629a5e767623f35f0a52c2d4d147e18cf7ac/src/dune_engine/action_builder0.mli#L70-L74

This API in principle seems sufficient, although one could imagine packaging it up into a more explicit combinator, with some associated documentation/examples. I haven't yet had time to think about how that combinator would look like -- any suggestions are welcome!

We could also consider exposing order-only dependencies at the level of the `dune` language, e.g. adding `order_only_deps` to the `rule` stanza, but I'm not sure it's really worth it.

[...]

PR #10943, opened 2024-09-20
[RFC] Add support for order-only and post-use dependencies to Dune engine

MA0010:
This PR (written as part of an internship at LexiFi) is an attempt to add support for order-only and post-use dependencies to the Dune engine. For a discussion of these notions and further background, see #5514. However, for simplicity, we explain the notions briefly:

- **Order-only** dependencies: an order-only dependency of a rule is a dependency that is guaranteed to be up-to-date when the rule is refreshed, but that changes to the dependency **do not** retrigger the rule.

- **Post-use** dependencies: a post-use dependency of a rule is an (ordinary) dependency that is declared **after** the rule action executes. This may sound a bit weird but it becomes interesting when the dependency is also declared as an order-only dependency of the rule.

The prototypical use of these notions is as follows: you have a rule for which the precise set of actual dependencies is hard to specify (eg the set of headers needed to compile a C file), so you declare an overapproximation (eg: all `*.h` files in a given directory) of the dependency set as order-only dependencies of the rule. When the rule is triggered, it produces as a side-effect the set of precise dependencies (eg `gcc -MD`) which can then be declared as "actual" (post-use) dependencies. Dune is then able to use this information to avoid retriggering the rule if a non-actual order-only dependency changes.

The PR contains two commits:

**First commit**

Here, we add support for "order-only" dependencies. This notion is actually already supported by the Dune engine (via the `Action_builder.goal` combinator), but not currently exposed in the Dune language. This commit exposes this notion in the Dune language via a `(order_only <dep>)` constructor. There is a small test to exhibit its effect.

**Second commit**

In this commit we add support for "post-use" dependencies. Concretely we add an **action** `(needed_deps <paths>)` that can be used to declare "actual" dependencies as part of the action (the idea being that `<paths>` are created as a side-effect of previous actions). Each `<path>` is supposed to contain zero or more S-expressions describing dependencies (following a grammar corresponding almost exactly to a subset of the usual dependency language). A few remarks about the implementation are below. A test for this feature is similarly included.

Both features are documented in the PR, but the intent is not necessarily to expose the features to the user at this time; the documentation was mainly to help understand the features for review. If we want to keep these features, we may decide not to expose them to the user in this way at this time.

**What is missing.**

- When one declares "post-use" dependencies in the way described above, it is necessary for correctness that these dependencies be a subset of the declared dependencies of the rule. Otherwise the dependencies in question may not be up-to-date when the action that needs them executes. In other words, the whole point of "post-use" dependencies is that they should already be up-to-date when the action begins to execute and `(needed_deps)` is only there to help Dune refine the condition for retriggering the rule. This means that `(needed_deps)` should never actually trigger any building; and it is an error if actual building is necessary. However, this is not checked at this time. I have a patch more or less ready for this, but it seems better to leave this for later once an initial review of the general approach is done.

- For now, rules using `(needed_deps)` are not stored in the shared cache. However, in principle this should be possible to do, but it was left for later for simplicity. (Some changes to the semantics of the shared cache appear to be required.)

**Indications about the implementation of the second commit**

The goal is to use the knowledge of actual/post-use dependencies to optimize rebuilding of a rule. To do that, we reuse the code path used to transmit "dynamic" dependencies (as in the Action plugin) back to the build system. We extend the relevant type to also transmit back the set of needed dependencies gathered during action execution. This set of actual dependencies so obtained are then stored in the local workspace cache entry for the rule. When Dune looks up a rule in the local cache to decide whether to re-execute it (using `Rule_cache.Workspace_local.lookup`), the digest of the needed deps is checked to decide if the rule is up-to-date or not. If these dependencies are additionally declared as order-only, then the net effect is that the rule will not be retriggered if an order-only, non-actual dependency changes.

The work in this PR was supervised by @nojb.

Assigning to engine expert @snowleopard, but any and all comments warmly welcome.

### DIFF
--- a/bin/print_rules.ml
+++ b/bin/print_rules.ml
@@ -99,6 +99,7 @@ let rec encode : Action.For_shell.t -> Dune_lang.t =
   | Mkdir x -> List [ atom "mkdir"; target x ]
   | Pipe (outputs, l) ->
     List (atom (sprintf "pipe-%s" (Outputs.to_string outputs)) :: List.map l ~f:encode)
+  | Needed_deps xs -> List (atom "needed_deps" :: List.map xs ~f:path)
   | Extension ext -> List [ atom "ext"; Dune_sexp.Quoted_string (Sexp.to_string ext) ]
 ;;
 
--- a/doc/concepts/dependency-spec.rst
+++ b/doc/concepts/dependency-spec.rst
@@ -46,6 +46,9 @@ Dependencies in ``dune`` files can be specified using one of the following:
 - ``(include <file>)`` read the s-expression in ``<file>`` and interpret it as
   additional dependencies. The s-expression is expected to be a list of the
   same constructs enumerated here.
+- ``(order_only <deps>)`` declares ``<deps>`` as "order only" dependencies.
+  This means that these dependencies will be guaranteed to be up-to-date before
+  building the target, but changes in them will not cause the target to be rebuilt.
 
 In all these cases, the argument supports :doc:`variables`.
 
--- a/doc/reference/actions/needed_deps.rst
+++ b/doc/reference/actions/needed_deps.rst
@@ -1,0 +1,38 @@
+needed_deps
+----
+
+.. highlight:: dune
+
+.. describe:: (needed_deps <file> ...)
+
+   Declare the contents of ``<file>`` assumed to as dependencies of the rule.
+   Each ``<file>`` is assumed to contain a valid dependency specification (see
+   below).
+
+   This is used to declare that certain "order only" dependencies (see
+   :docs:`concepts/dependency-spec`) are actual dependencies of the rule.
+
+   The file passed as an argument is assumed to contain zero or more
+   S-expressions using the following constructors which cover a subset of the
+   usual dependency sepecification:
+
+   - ``(file <filename>)``
+   - ``(alias <alias_name>)``
+   - ``(file_selector <glob>)``
+   - ``(universe)``
+
+   (see :docs:`concepts/dependency-spec` for the details of each constructor).
+
+   In the following example, both ``a`` and ``b`` are declared as "order only"
+   dependencies. The file ``b`` is in addition declared as a needed dependency
+   for the rule. The rule will only be reran if ``b`` changes, but not if ``a``
+   changes.
+
+   Example::
+
+   (rule
+    (deps (order_only (file a) (file b)))
+    (action
+     (progn
+     (with-stdout-to deps (echo "(file b)"))
+     (needed_deps deps))))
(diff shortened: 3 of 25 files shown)


snowleopard:
@MA0010 Hi there, thank you for the prototype!

I am about to go on parental leave so I'm afraid I won't be able to review this PR any time soon, but I'd like to say that @rgrinberg and I are planning to upstream our internal changes to the engine in November. Until we do that, reviewing this PR seems hard since the engine has undergone a major rewrite to make it more scalable. We do still have many of the same concepts but I would very much prefer reviewing a patch on top of "actual" code rather than what Dune engine was a year ago (from my perspective).

Perhaps, Rudi can give you some comments while I'm away?

nojb:
> @rgrinberg and I are planning to upstream our internal changes to the engine in November.

That's good to hear!

> I would very much prefer reviewing a patch on top of "actual" code rather than what Dune engine was a year ago (from my perspective).

Makes sense.

> I am about to go on parental leave

Congratulations!

rgrinberg:
@nojb do you have a specific use for this feature? If we are to accept this feature and work towards merging it, it would be good to know that you have some specific motivation in mind. I don't plan to work on general purpose engine features until we have a convincing use case in mind.

nojb:
> @nojb do you have a specific use for this feature? If we are to accept this feature and work towards merging it, it would be good to know that you have some specific motivation in mind. I don't plan to work on general purpose engine features until we have a convincing use case in mind.

Thanks for the ping @rgrinberg. Yes, there is a specific motivation for this feature, which is mentioned at the top of the issue.

> The prototypical use of these notions is as follows: you have a rule for which the precise set of actual dependencies is hard to specify (eg the set of headers needed to compile a C file), so you declare an overapproximation (eg: all *.h files in a given directory) of the dependency set as order-only dependencies of the rule. When the rule is triggered, it produces as a side-effect the set of precise dependencies (eg gcc -MD) which can then be declared as "actual" (post-use) dependencies. Dune is then able to use this information to avoid retriggering the rule if a non-actual order-only dependency changes.

Concretely, the motivation for this work (which was done during an internship at LexiFi) was to support fine-grained dependencies _between_ OCaml libraries in Dune, starting from today's rough over-approximation (all modules depend on all modules across libraries), and refining the dependency tree after the initial compilation (as in the paragraph above), using either `ocamlobjinfo` (in a prototype) or using information which is already available to the compiler (tweaking it to make this information available to Dune or other clients).

I still think this is an important direction to explore and that the motivation is solid, but I am not pushing to integrate this code as-is, which was more of a POC than anything else.

rgrinberg:
Thanks for the response. For gcc, I think we can get precise dependencies compilation with `gcc -MM` without compilation and therefore can be taken advantage of without any changes to the engine. For OCaml, ocamldep would be precise enough in the absence of module aliases. I've been told that ocamldep already has a `-map` mode that tries to do something similar, but it's not quite convenient enough for dune. So I would lean towards that before considering any extensions to the engine.

PR #10946, opened 2024-09-23
Add fake curl exe to ocamlformat tests

gridbugs:
The tests would sometimes try to run the system curl executable which would cause the tests to fail.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/ocamlformat/dune
+++ b/test/blackbox-tests/test-cases/pkg/ocamlformat/dune
@@ -1,3 +1,7 @@
 (cram
  (deps helpers.sh)
  (applies_to :whole_subtree))
+
+(cram
+ (deps %{bin:curl})
+ (applies_to ocamlformat-e2e))


moyodiallo (approved):

rgrinberg (approved):

test/blackbox-tests/test-cases/pkg/ocamlformat/dune
```diff
@@ -1,3 +1,3 @@
 (cram
- (deps helpers.sh)
+ (deps helpers.sh %{bin:curl})
```
rgrinberg, line 2: Does every test really depend on curl?

gridbugs: Not every test. I can make it more fine grained.

rgrinberg: Not completely necessary, but it is an improvement IMO. I just found it surprising that there are no tests that don't use the network.

maiste (approved):

PR #10950, opened 2024-09-23
doc(pkg): Explain package management

Leonidas-from-XIV:
The explanation part of dune package management, split out following the feedback of #10920, aimed at users who want to have a deeper understanding on how the system works under the hood.

### DIFF
--- a/doc/explanation/package-management.md
+++ b/doc/explanation/package-management.md
@@ -1,0 +1,233 @@
+# How Package Management Works
+
+This document explains how Dune's package management works under the hood. It
+requires a bit of familiarity with how opam repositories work and how Dune
+builds packages. Thus it is aimed at people who want to understand how the
+feature works, not how it is used.
+
+For a tour on how to apply package management to a project, refer to the
+{doc}`/tutorials/dune-package-management/index` tutorial.
+
+## Motivation
+
+A core part of modern programming is using existing code to save time. The
+OCaml package ecosystem has quite a long history with many projects building
+upon each other over many years. A significant step forward was the creation of
+the OCaml Package Manager, opam, along with the establishment of a public
+package repository which made it a lot more feasible to share code between
+people and projects.
+
+Over time, best practices have evolved, and while opam has incorporated some
+changes, it couldn't adopt all the modern workflows due to its existing user
+base and constraints.
+
+Thus the Dune Package Management has been designed with a few core goals in
+mind:
+
+* No global state visible to users, everything is local to projects
+* Package management is configured through files (`dune-project` and optionally
+  `dune-workspace`)
+* Repositories are automatically kept up to date unless explicitly configured
+  to use specific versions
+* Builds can only access packages they have declared dependencies on
+* Reproducible builds through lockfiles
+
+Dune plays well with the existing OCaml ecosystem and does not introduce a new
+type of packages. Rather, it uses the same package repository and Dune packages
+stay installable with opam.
+
+## Package Management in a Project
+
+This section describes what happens in a Dune project using the package
+management feature.
+
+## Dependency Selection
+
+The first step is to determine which packages need to be installed.
+Traditionally this has been defined in the `depends` field of a project's opam
+file(s).
+
+Since version 1.10 Dune has supported {doc}`opam file generation
+</howto/opam-file-generation>` by specifying the package dependencies in the
+`dune-project`.
+
+The package management feature uses the same metadata, as Dune will determine
+the list of packages to install from the `depends` field in the `dune-project`
+file. This allows projects to completely omit generation of `.opam` files, as
+long as they use Dune for package management. Thus all dependencies on OCaml
+packages are only declared in one single file.
+
+For compatibility with a larger amount of existing projects, Dune will also
+collect dependencies from `.opam` files in the project. So while recommended,
+there is no obligation to switch to declaring dependencies in the
+`dune-project`. Likewise the generation of `.opam` files continues to be
+supported.
+
+## Locking
+
+Given the list of the project's dependencies and their version
+constraints, the next steps are:
+
+1. Find the transitive dependencies and figure out a version for each
+   dependency that satisifies the constraints
+2. For each dependency, download it, build it, and make it available to the
+   project
+
+In opam, `opam install` does both of these.
+
+In Dune, these are separate steps: the first one is `dune pkg lock`, and the
+second one happens implicitly as part of [building](#building).
+
+The idea of doing the first step and recording it for later is popular in other
+programming language package managers like NPM and is usually called locking.
+Creating a lock file ensures that the dependencies to be installed are
+always the same - unless that lock file is updated of course.
+
+:::{note}
+`opam` also supports creating lock files. However, these are not as central to
+the opam workflow as they are in the case of package management in Dune, which
+always requires a set of locked packages.
+:::
+
+In the most general sense, a lock file is just a set of specific packages
+and their versions to be installed.
+
+Instead of a lock file, Dune writes this information to a directory (the "lock
+directory") with files that describe the dependencies. It includes the
+package's name and version. Unlike many other package managers, the files
+include a lot of other information as well, such as the location of the source
+archives to download (since there is no central location for all archives), the
+build instructions (since each package can use its own way of building), and
+additional metadata like the system packages it depends upon.
+
+The information is stored in a directory (`dune.lock` by default) as separate
+files, to reduce potential merge conflicts and simplify code review. Storing
+additional files like patches is also simpler this way.
+
+### Package Repository Management
+
+To find a valid solution that allows a project to be built, it is necessary to
+know what packages exist, what versions of these packages exist, and what other
+packages these depend on, etc.
+
+In opam, this information is tracked in a central repository called
+[`opam-repository`](https://github.com/ocaml/opam-repository), which contains
+all the metadata for published packages.
+
+It is managed using Git; opam typically uses a snapshot to find the
+dependencies when searching for a solution that satisfies the constraints.
+
+Likewise, Dune uses the same repository; however, instead of snapshots of the
+contents, it uses the Git repository directly.
+
+:::{note}
+Dune maintains a shared internal cache containing all Git repositories that
+projects use. This way updates and checkouts are very fast because only new
+revisions have to be retrieved. The downside is that to be included in the
+cache, all the Git repos have to be cloned first which depending on the size of
+the repositories can take a bit of time.
+:::
+
+On every call to `dune pkg lock`, Dune will update the metadata repository
+first (hence why efficiently updating that repository matters). This means that
+each `dune pkg lock` will use the newest set of packages available.
+
+However, it is also possible to declare specific revisions of the repositories,
+to get a reproducible solution. Due to using Git, any previous revision of the
+repository can be used by specifying a commit hash.
+
+Dune uses two repositories by default:
+
+* `upstream` refers to the default branch of `opam-repository`, which contains
+  all the publicly released packages.
+* `overlay` refers to
+  [opam-overlay](https://github.com/ocaml-dune/opam-overlays), which defines
+  packages patched to work with package management. The long-term goal is to
+  have as few packages as possible in this repository as more and more packages
+  work within Dune Package Management upstream. Check the
+  [compatibility](#compatibility) section for details.
+
+### Solving
+
+After Dune has read the constraints and loaded set of candidate packages, it is
+necessary to determine which packages and versions should be selected for the
+package lock.
+
+To do so, Dune uses
+[`opam-0install-solver`](https://github.com/ocaml-opam/opam-0install-solver),
+which is a variant of the [`0install`](https://github.com/0install/0install)
+solver to find solutions for opam packages.
+
+Contrary to opam, the Dune solver always starts from a blank slate; it assumes
+nothing is installed and everything needs to be installed. This has the
+advantage that solving is now simpler, and previous solver solutions don't
+interfere with the current one. Thus, given the same inputs, it should always
+come up with the same result; no state is held between the solver runs.
+
+This can lead to more packages being installed (as opam won't install new
+package versions by default if the existing versions satisfy the constraints),
+but it avoids interference from already installed packages that lead to
+potentially different solutions.
+
+After solving is done, the solution gets written into the lock directory with
+all the metadata necessary to build and install the packages. From this point
+on, there is no need to access the package metadata repositories.
+
+:::{note}
+Solving and locking does not download the package sources. These are downloaded
+in the build step.
+:::
+
+(building)=
+## Building
+
+When building, Dune will read the information from the lock directory and set
+up rules for the packages. Check {doc}`/explanation/mental-model` for details
+about rules.
+
+The rules that the package management sets up include:
+
+* Fetch rules to download and unpack the source archives, and also download any
+additional sources such as patches
+* Build rules to execute the build instructions stored in the lock directory
+* Install rules to put the artifacts that were built into the appropriate
+  Dune-managed folders
+
+Creating these processes as rules mean that they will only be executed on
+demand, so if the project has already downloaded the sources, it does not need
+to download them again. Likewise, if packages are installed, they stay
+installed.
+
+The results of the rules are stored in the project's `_build` directory and
+managed automatically by Dune. Thus, when cleaning the build directory, the
+installed packages are cleaned as well and will be reinstalled at the next
+build.
+
+(compatibility)=
+## Packaging for Dune Compatibility
+
+Dune can build and install most packages as dependencies, even if they are not
+built with Dune themselves. Dune will execute the build instructions from the
+lock directory, very similar to opam.
+
+However, packages must adhere to certain rules to be compatible with Dune.
+
+The most important one is that the packages must not use absolute paths to
+refer to files. That means they cannot read the path they are being built or
+installed in and expect this path to remain the same. Dune builds packages in a
+sandbox location, and after the build has finished, it moves the files to the
+actual destination.
+
+:::{note}
+Unlike opam, Dune at the moment does not wrap the build in sandboxing tools
+like [Bubblewrap](https://github.com/containers/bubblewrap).
+:::
+
+To comply with these restrictions the usual solution is to use relative paths,
+as Dune guarantees that packages installed into different sections are
+installed in a way where their relative location stays the same.
+
+The `overlay` repository exists specifically to make currently non-compliant
+packages compatible with Dune's package management. It does so by supplying
+releases of packages where the current upstream releases don't support Dune
+package management yet.
--- a/doc/explanation/index.rst
+++ b/doc/explanation/index.rst
@@ -9,6 +9,7 @@ the rest of the OCaml ecosystem.
 
    preprocessing
    ocaml-ecosystem
+   package-management
    opam-integration
    bootstrap
    mental-model


doc/explanation/package-management.md
```diff
@@ -0,0 +1,220 @@
...
+the modern workflows.
+
+Dune package management attempts to take the parts of the OPAM ecosystem that
+have stood the test of time and couple them with a modern workflow. Some of the
+improvements include:
+
+* Automatic package repository updates
+* Easily reproducible dependencies
+* All package dependencies declared in a single file that is kept in-sync
+* Per-project dependencies
```
samoht, line 32: I don't see what's new in that list from what opam can already do (apart from automatic package repository updates, which can be seen as quite bad from a reproducibility point of view if you don't have a lockfile). 

Instead of comparing opam vs. dune pkg, I would stress the design principle. A random list of stuff:
- making the OCaml environment setup trivial;
- promoting a lockfile-first approach to address reproducibility use cases;
- unifying the configuration files (and removing hidden global states managed by a CLI tool);
- improving cross-packages and vendoring workflows;
and much more.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,220 @@
...
+dependencies when trying to find a solution that satisfies the constraints.
+
+Likewise, Dune uses the same repository, however instead of snapshots of the
+files it uses the Git repository directly. In fact, Dune maintains a shared
+internal cache containing all Git repositories that projects use. This has the
+advantage that updates to the metadata are very fast since only the newest
+revisions have to be retrieved. The downside is that for the creation of the
+cache the entire repository has to be cloned first.
+
+Given a priorization of fast updates, whenever Dune needs to determine the
```
samoht, line 119: Does it mean `dune pkg lock` always needs network access?

doc/explanation/package-management.md
```diff
@@ -0,0 +1,220 @@
...
+However, there are some important restrictions that packages must adhere to to
+be able to work with Dune.
+
+The most important one is that the packages must not use absolute paths to
+refer to files. That means, they cannot read the path they are being built or
+installed into and expect this path to be stable. Dune builds packages in a
+sandbox location and after the build has finished it moves the files to the
+actual destination.
+
+The reason for this is that one hand enables building without messing up the
```
samoht, line 206: one hand enables -> missing word?

doc/explanation/package-management.md
```diff
@@ -0,0 +1,220 @@
...
+lock file, very similar to `opam`.
+
+However, there are some important restrictions that packages must adhere to to
+be able to work with Dune.
+
+The most important one is that the packages must not use absolute paths to
+refer to files. That means, they cannot read the path they are being built or
+installed into and expect this path to be stable. Dune builds packages in a
+sandbox location and after the build has finished it moves the files to the
+actual destination.
```
samoht, line 204: Maybe a good place to explain the difference between Dune and Opam sandboxing models?

samoht:
I've left a few comments - I'm also wondering if we should highlight the Dune cache a bit more here - that's one of the killer feature of `dune pkg`: instant build of most of your dependencies (including the compiler) as they could be re-used between projects.

doc/explanation/package-management.md
Leonidas-from-XIV: It needs network access unless the repos you use are pinned to a revision that's in the rev store.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,220 @@
...
+To go from a set of dependency constraints for a project to a set of installed
+packages and versions there needs to be a step to determine the right packages
+and their versions to be installed.
+
+In `opam` this process happens as part of `opam install`, thus linking finding
+a solution that satifies the given constrains and installation into one step.
+Dune on the other hand moves finding a solution and installing into separate
+steps. First a solution is found and then packages are installed separately.
+The idea of finding a solution and recording it for later is popular in many
+package managers and is usually called locking.
```
maiste, line 76: I think it can be nice to give some examples here

doc/explanation/package-management.md
```diff
@@ -0,0 +1,220 @@
...
+
+A Dune lockfile extends this to a directory with files that describe the
+dependencies. It includes the name of the package and the version. Unlike many
+other package managers, the files include a lot of other information as well,
+such as the location of the source archives to download (since there is no
+central location for all archives), the build instructions (since each package
+can use its own way of building) as well as additional metadata like the system
+packages it depends upon.
+
+The files are stored in a directory as separate files, because that makes it
```
maiste, line 95: It could be interesting to mention the `dune.lock` dir name.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,220 @@
...
+  [opam-overlay](https://github.com/ocaml-dune/opam-overlays), which defines
+  packages patched to work with package management. The long-term goal is to
+  have as few packages as possible in this repository as more and more packages
+  work within Dune package management upstream. Check the
+  [compatibility](#compatibility) section for details.
+
+### Solving
+
+After Dune has retrieved the constraints and the set of possible packages it is
+necessary to determine which packages & versions should be selected for the
```
maiste, line 141:
```suggestion
necessary to determine which packages and versions should be selected for the
```

doc/explanation/package-management.md
```diff
@@ -0,0 +1,220 @@
...
+
+This can lead to more packages being installed (as OPAM won't by default
+install new versions of packages if the existing versions satisfy the
+constraints) but it avoids interference from already installed packages that
+lead to potentially different solutions.
+
+After solving is done, the solution gets written into the lock directory with
+all the data that is necessary to build and install the packages. From this
+point, the project contains all necessary information and does not need to
+access any package repositories.
```
maiste, line 162: It might be interesting to have a terminology specification somewhere. When I read the sentence, it seems wrong, as we will later install packages. The distinction between package repositories and the packages we installed should be somewhere in an index, IMO.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Building
+
+When building, Dune will read the information from the lock directory and set
+up build rules to use the packages. The rules include
```
maiste, line 167: Could be nice to point to Dune mental model here :smile:

doc/explanation/package-management.md
```diff
@@ -0,0 +1,220 @@
...
+The results of the rules are stored in the `_build` directory of the project
+and managed automatically by Dune. Thus when cleaning the build directory, the
+installed packages are cleaned as well and will be reinstalled at the next
+build.
+
+When building the users project, the installed packages are added to the
+necessary search paths, thus user code can use the dependencies without any
+additional configuration.
+
+(compatibility)=
```
maiste, line 190:
```suggestion
```
Is it Sphinx syntax or a typo?

doc/explanation/package-management.md
```diff
@@ -0,0 +1,220 @@
...
+  Dune-managed folders
+
+Creating these processes as rules means that they will only be executed on
+demand, thus if the project has already downloaded the sources, it does not
+need to download them again. Likewise, if packages are installed, they stay
+installed.
+
+The results of the rules are stored in the `_build` directory of the project
+and managed automatically by Dune. Thus when cleaning the build directory, the
+installed packages are cleaned as well and will be reinstalled at the next
```
maiste, line 183: Shouldn't we talk about the fact that they are cached, so it will be much faster the second time?

doc/explanation/package-management.md
```diff
@@ -0,0 +1,220 @@
...
+
+The reason for this is that one hand it enables building without messing up the
+current state and on the other hand it allows for caching artifacts across
+projects.
+
+To sidestep these restructions in many cases the solution is to use relative
+paths, as Dune guarantees that packages installed into different sections are
+installed in a way where their relative location stays the same.
+
+A minor difference is that Dune does not support packages installing themselves
```
maiste, line 214: I'm not sure to understand this paragraph, could you elaborate?

doc/explanation/package-management.md
Leonidas-from-XIV: But they're not cached (by default) at the moment, so it would seem a bit weird to describe that as a feature.

doc/explanation/package-management.md
Leonidas-from-XIV: It's MyST (the markdown plugin for reST/RTD) syntax, so I can refer to that section as `#compatibility` elsewhere. There's a whole stack of Markdown, MyST and RTD things overlapping so determining what syntax to use can be a bit surprising.

doc/explanation/package-management.md
maiste: My bad, I thought they were with the recent discussions on the cache :see_no_evil:

doc/explanation/package-management.md
Leonidas-from-XIV: Some (very few) opam packages install themselves in the compilers `stdlib` folder, where you don't need to declare them as dependency in `(libraries)` (so they work like `List` or `String` etc).

This doesn't work in Dune, as Dune does not allow packages to install themselves into other package's folders. It is a bit hacky and only works in OPAM because the compiler's stdlib folder has a known location; OPAM does not expose that location as a section for `install` files.

Do you have a suggestion how to describe this in an understandable way?

maiste: Out of curiosity, do you have a package name doing this? :smile: 

Otherwise, the explanation you gave me is pretty understandable. I was lacking the context that some packages are actually doing this! I think providing the context would make it more understandable. 
```
Some packages install themselves in the Stdlib directory
-> Allow by opam because of a known location 
-> Dune is per project so don't have a global location
-> [rest of you explanation]
```
:+1:

Leonidas-from-XIV: Yes, we encountered this issue with the `num` package: https://github.com/ocaml/opam-repository/pull/26178. I think the only packages that would do this are packages that have been spun out of the stdlib (thus were previously accessible without a `libraries` entry thus the `compat` packages preserve the behavior) but don't install a into a package directory. In the case of `num` we solved this by installing into both places and that seems to be a best-of-both-worlds situation.

However, this is a fairly rare case, as `num` was the only occurrence, so maybe not even worth mentioning.

doc/explanation/package-management.md
samoht: Do you have a way to control this? Ie. what if I want to rebuild a package in a train (with all the dependencies already cached locally because I've built the project in the past)

Leonidas-from-XIV: If you have built the project in the past, then you don't need to relock and it works offline (as fetch rules are cached out of the box since #10710), even if you remove `_build`.

If you want to relock without internet it requires a bit if configuration. You can define the opam-repos to use for locking to use an exact revision [like here](https://dune.readthedocs.io/en/latest/tutorials/dune-package-management/repos.html), in which case Dune will first check whether the revision is available in the revision cache, and only pull if the revision was not found yet.

doc/explanation/package-management.md
Leonidas-from-XIV: The main difference is that we build in a different path than the one we install to. As far as I know the fact that we don't wrap things in `bwrap`/`sandbox-exec` are not design decisions, just something we haven't implemented yet. It is good idea and shouldn't be too difficult, especially with `opam` having made sure that packages generally work in its sandbox.

doc/explanation/package-management.md
Leonidas-from-XIV: I've added NPM, any other package managers worth mentioning? Cargo?

doc/explanation/package-management.md
Leonidas-from-XIV: I was a bit hesitant as `dune.lock` feels a bit like an implementation detail but I guess in case we change it we can change this document as well.

doc/explanation/package-management.md
Leonidas-from-XIV: I am unsure what you consider untrue about it? I've elaborated that section, also noting that it does not download sources, is it better now?

doc/explanation/package-management.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Dependency Selection
+
+The first step is to determine which packages need to be installed.
+Traditionally this has been defined in the `depends` field of a projects opam
+file(s).
+
+Since a while Dune has also supported {doc}`opam file generation
+</howto/opam-file-generation>` by specifying the package dependencies in the
+`dune-project`. Outside of this feature, Dune had not used the `depends` stanza.
```
christinerose, line 50: There's something off with the beginning of this sentence. Do you mean: "For a while, Dune has also supported..." or "Since Dune has supported...."

doc/explanation/package-management.md
```diff
@@ -0,0 +1,226 @@
...
+
+Dune can build and install most packages as dependencies, even if they are not
+built with Dune themselves. Dune will execute the build instructions from the
+lock file, very similar to opam.
+
+However, packages must meet certain requirements to be compatible with Dune.
+
+The most important one is that the packages must not use absolute paths to
+refer to files. That means, they cannot read the path they are being built or
+installed into and expect this path to be stable. Dune builds packages in a
```
christinerose, line 208:
```suggestion
refer to files. This means they cannot rely on the path they are being built or installed in, as it may not remain consistent. Dune builds packages in a
```
Is this still accurate? If so, it improves readability and clarity. If not, let's work on clarifying it together.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,226 @@
...
+sandbox location, and after the build has finished, it moves the files to the
+actual destination.
+
+The reason for this is clear. On one hand it enables building without messing up the
+current state, and on the other hand it allows for caching artifacts across
+projects.
+
+To sidestep these restructions in many cases the solution is to use relative
+paths, as Dune guarantees that packages installed into different sections are
+installed in a way where their relative location stays the same.
```
christinerose, line 218:
```suggestion
To work around these restrictions, a common solution is to use relative paths. Dune ensures that packages installed in different sections maintain the same relative location, allowing for more flexibility.
```
Is this still accurate?

doc/explanation/package-management.md
```diff
@@ -0,0 +1,226 @@
...
+
+A minor difference is that Dune does not support packages installing themselves
+into the standard library, thus being available without having to be declared a
```
christinerose, line 221:
```suggestion
into the standard library, so they're available without having to be declared a
```
Does this work better?

christinerose (commented):
A few suggestions for potentially clarifying sentences.

doc/explanation/package-management.md
v-gb: It wouldn't say "side step" or "work around" to refer to the expected way of satisfying a condition. Maybe:

To comply with this requirement, the usual solution ...

Leonidas-from-XIV: I like the "comply" wording because it makes it sound positive, whereas "work around" and "side step" is a sounds like there is a bug that people have to try to avoid. I would argue that avoiding absolute paths is a good thing in general and makes the code better even without Dune package management.

doc/explanation/package-management.md
maiste: I think NPM is a good one :+1:

doc/explanation/package-management.md
maiste: Sorry for the delay, as it was marked as outdated I didn't see it. The `package repositories` can be confusing: we won't access opam-repository for the metadata, but it will need to access the server where you host the package to fetch the tarball. It could be considered as a package repository too. Hence, my question about the terminology ^^

doc/explanation/package-management.md
rgrinberg: Here's a few principles that I've had in mind when designing this feature:

1. No global state visible to the user
2. All package configuration must be done via config file updates (dune-project or dune-workspace). No stateful commands such as `opam pin`.
3. "Automatic` package repository updates" also applies to pins. For example, we re-fetch branches & tags.
4. Package builds can only access packages which are listed as dependencies.
5. Users do not need to learn yet another file format to configure their workspace. Everything should be doable via workspaces.
6. Build plans are independently versioned so that they can be interpreted in the same way between different versions of dune

doc/explanation/package-management.md
```diff
@@ -0,0 +1,226 @@
...
+
+It is managed using Git, and opam typically uses a snapshot to find the
+dependencies when searching for a solution that satisfies the constraints.
+
+Likewise, Dune uses the same repository; however, instead of file snapshots, 
+it uses the Git repository directly. In fact, Dune maintains a shared
+internal cache containing all Git repositories that projects use. The
+advantage is that updates to the metadata are very fast because only the newest
+revisions have to be retrieved. The downside is that for the creation of the
+cache, the entire repository has to be cloned first.
```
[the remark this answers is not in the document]
maiste: It's the metadata repository here: the `opam-repository` one.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,226 @@
...
+
+Given a priorisation of fast updates, whenever Dune needs to determine the
+available packages, it will update the repository first. Thus each locking
+process by default will use the newest set of packages available.
+
+However, it is also possible to specify specific revisions of the repositories,
+to get a reproducible solution. Due to using Git, any previous revision of the
+repository can be used by specifying a commit hash.
```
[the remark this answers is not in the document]
maiste: Here it is in the context of `opam-repository` so metadata repository or "package repository".

doc/explanation/package-management.md
Leonidas-from-XIV: Ok, I added the word "metadata" (and you can see there is also a note that sources are not being downloaded).

I personally don't consider the package sources a repository per-se as for a repository to me implies a sort of centralized collection of things (like a git repository is a collection of files that belong to a project or opam repository is a collection of opam files or the NPM registry is a repository because it contains the packages themselves), whereas the source files are just a loose collection of URLs. While it's true that a lot of tarballs are on Github, that seems more an emergent outcome than some deliberate effort in creating a repository of package tarballs.

Anyway, long story short, I'm marking this as resolved unless people disagree.

doc/explanation/package-management.md
Leonidas-from-XIV: Thanks for pointing it out. The "a while" was my attempt to weasel out of the fact that I don't know since when this existed. This pushed me to open #10984 so I can write "1.10" here and it makes more sense.

doc/explanation/package-management.md
Leonidas-from-XIV: I've updated it a bit to remove the commas. I don't think "consistent" is the right word in the context as that would imply the path is "inconsistent" but that doesn't make that much sense. I've paraphrased it to be "stay the same" which might be simpler to understand. WDYT?

doc/explanation/package-management.md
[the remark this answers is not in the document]
Leonidas-from-XIV: This is in the "package repository management" section, so I (hoped) it was clear that it is about the opam metadata.

It will normally use the js_of_ocaml release tarball if that's what your lockfile refers to. But if you use a pin to specify the jsoo git repo it will also add that repo to the git revision cache (thus multiple projects doing this will not have to clone jsoo twice).

doc/explanation/package-management.md
```diff
@@ -0,0 +1,226 @@
...
+installed.
+
+The results of the rules are stored in the project's `_build` directory
+and managed automatically by Dune. Thus, when cleaning the build directory, the
+installed packages are cleaned as well and will be reinstalled at the next
+build.
+
+When building the users project, the installed packages are added to the
+necessary search paths, so user code can use the dependencies without any
+additional configuration.
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Fair enough, I've removed it.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,226 @@
...
+The package management feature changes this, as Dune now determines the list of
+packages to install from the `depends` stanza in the `dune-project` file. This
+allows projects to completely omit generation of `.opam` files, as long as they
+use Dune for package management. Thus all dependencies are only declared in one
+file.
+
+For compatibility with a larger amount of existing projects, Dune will also
+collect dependencies from `.opam` files in the project. So while recommended,
+there is no obligation to switch to declaring dependencies in the
+`dune-project`. Likewise the generation of `.opam` files will still work.
```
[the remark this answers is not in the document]
Leonidas-from-XIV: What do you mean?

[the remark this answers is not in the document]
v-gb: I mean the user docs pointed to by https://dune.ci.dev/ only say that the dependencies are specified in the dune-project file. If the opam file is possible as well, maybe the user docs is where this information should live.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,226 @@
+# How Package Management Works
+
+This document gives an explanation on how the new package management
+feature introduced in Dune works under the hood. It requires a bit of
+familiarity with how opam repositories work and how Dune builds packages. Thus
+it is aimed at people who want to understand how the feature works, not how it
+is used.
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Good point, better to avoid a The New iPad (2012) situation.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,226 @@
...
+
+Since a while Dune has also supported {doc}`opam file generation
+</howto/opam-file-generation>` by specifying the package dependencies in the
+`dune-project`. Outside of this feature, Dune had not used the `depends` stanza.
+
+The package management feature changes this, as Dune now determines the list of
+packages to install from the `depends` stanza in the `dune-project` file. This
+allows projects to completely omit generation of `.opam` files, as long as they
+use Dune for package management. Thus all dependencies are only declared in one
+file.
```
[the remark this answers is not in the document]
Leonidas-from-XIV: I've changed this section a bit. I think the core confusion comes from the fact that we have two types of dependencies:

 * OCaml packages (or OPAM packages?) which are now in `dune-project`
 * Libraries (or findlib libraries?) which are typically shipped in OCaml packages, which continue to be declared in `dune` files.

To me this is fairly clear since we're talking about an opam replacement (and opam has no notion of the findlib packages), but maybe it should go to some kind of glossary?

doc/explanation/package-management.md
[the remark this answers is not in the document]
Leonidas-from-XIV: I've been looking at where this is stated and I can't come up with any place; could you point me to it?

The tutorial does puts the dependencies in `dune-project` but given it is a tutorial it deliberately glosses over a number of minor things.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,226 @@
...
+  patches and unpack them
+* Build rules to evaluate the build instructions from the build instructions
+  stored in the lock directory
+* Install rules to put the artifacts that were built into the appropriate
+  Dune-managed folders
+
+Creating these processes as rules mean that they will only be executed on
+demand, so if the project has already downloaded the sources, it does not
+need to download them again. Likewise, if packages are installed, they stay
+installed.
```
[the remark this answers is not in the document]
Leonidas-from-XIV: This is definitely the plan, however at the moment by default only downloading the dependencies is cached, thus I didn't want to introduce information that would require additional disclaimers (aka "this works but at the moment only if you manually set `(dune-cache enabled)` but we're working on it"). Thus I would wait with the inclusion of a paragraph like this until the feature lands as default and this would be true without any qualifiers.

christinerose (approved):

doc/explanation/package-management.md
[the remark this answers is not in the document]
v-gb: Ok, this was not clear, because shortly above, the text says:

> In fact, Dune maintains a shared internal cache containing all Git repositories that projects use.

where "all the repository that projects use" is still only referring to metadata repositories, but it seems like it's talking about all repositories involved, including source repositories.

doc/explanation/package-management.md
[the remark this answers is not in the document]
v-gb: The dune.ci.dev site seems to be gone?
The document was not referred to as "tutorial" in my memory, this was referred to as "doc" or something like that. This bit of text seemed important enough to mention to me, but maybe not.

doc/explanation/package-management.md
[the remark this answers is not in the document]
v-gb: I admit that I have no clue why we have library names, ocamlfind names, and opam names. I have never seen cases where the ocamlfind and opam names are of any use, but that could be ignorance. On the other hand, I get confused semi regularly because I know the lib name I want to use but I have no idea how to figure what package it comes from.

The dune files contains things like `core_kernel.foo`, so I have the impression that the `core_kernel` bit is the opam bit. I certainly generate my dune-project from my dune files by dropping the parts of the name after the dots:

```sh
$ dune describe external-lib-deps | sexp select external_deps | sexp query 'each (index 0)' | cut -d . -f 1 | sort -u | grep -xv dynlink | xargs echo
```

doc/explanation/package-management.md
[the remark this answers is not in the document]
v-gb: To me, the sentence " In fact, Dune maintains a shared internal cache containing all Git repositories that projects use. " means you're not talking only about opam metadata anymore. You're explicitly generalizing

doc/explanation/package-management.md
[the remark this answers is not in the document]
Leonidas-from-XIV: It's a bit of a complicated historical thing because these concepts separate but unrelated. ocamlfind names are names of libraries that are declared in META files (this is what Dune uses in `library` stanzas). These are used to tell the compiler in which directory to look at to find the compiled modules of a particular library, they basically extend the search patch with a few `-I` options. There was no means of distribution at first, then game GODI/WODI.

OPAM came along later and introduced OPAM packages as a way to distribute tarballs + their build instructions. However you can package anything in your tarball (multiple `ocamlfind` libraries in one OPAM package is totally valid) thus the names of packages are not linked to the names of libraries.

Dune adds a bit on top of that by forcing public libraries start with the name of the opam package that it is supposed to go in, but this is a Dune-specific convention and does not apply to all packages in OPAM.

Anyway, I think this is somewhat of a deep dive into something else that's not really related to package management.

doc/explanation/package-management.md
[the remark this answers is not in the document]
Leonidas-from-XIV: It's now at [preview.dune.build](https://preview.dune.build/) (with a redirect from the old URL), but also there I don't see any mention.

doc/explanation/package-management.md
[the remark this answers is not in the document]
Leonidas-from-XIV: Fair enough, I've updated the text to pull the info into a `note` and explain what is happening.

doc/explanation/package-management.md
Leonidas-from-XIV: I've updated the section to describe the features of the package management without explicit comparison with OPAM.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+Thus the Dune Package Management has been designed with a few core goals in
+mind:
+
+* No global state visible to users, everything is local to projects
+* Package management is configured through files (`dune-project` and optionally
+  `dune-workspace`)
+* Repositories are automatically kept up to date unless explicitely configured
+  to use specific versions
+* Builds can only access packages they have declared dependencies on
+* Reproducable builds through lockfiles
```
v-gb, line 33:
```suggestion
* Reproducible builds through lockfiles
```

Leonidas-from-XIV: Thanks :)

doc/explanation/package-management.md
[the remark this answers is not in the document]
v-gb: My point was precisely that this text was not in the user facing doc, and that maybe it should. But now that the site is available, well the site points to the dune docs, which then point the package management tutorial. So, fair enough, this mention isn't tutorial-worthy.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+changes, it couldn't adopt all the modern workflows due to its existing user
+base and constraints.
+
+Thus the Dune Package Management has been designed with a few core goals in
+mind:
+
+* No global state visible to users, everything is local to projects
+* Package management is configured through files (`dune-project` and optionally
+  `dune-workspace`)
+* Repositories are automatically kept up to date unless explicitely configured
```
v-gb, line 30:
```suggestion
* Repositories are automatically kept up to date unless explicitly configured
```

doc/explanation/package-management.md
leostera: I think its worth highlighting that we aren't bwrapping stuff – we can give a shoutout to Opam there as well.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+refer to files. That means they cannot read the path they are being built or
+installed in and expect this path to remain the same. Dune builds packages in a
+sandbox location, and after the build has finished, it moves the files to the
+actual destination.
+
+The reason for this is clear. On one hand it enables building without messing
+up the current state, and on the other hand it allows for caching artifacts
+across projects.
+
+To comply with these restructions the usual solution is to use relative paths,
```
ElectreAAS, line 224:
```suggestion
To comply with these restrictions the usual solution is to use relative paths,
```

ElectreAAS (approved):
Thanks a lot for this!

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Package Management in a Project
+
+This section describes what happens in a Dune project using the package
+management feature.
+
+## Dependency Selection
+
+The first step is to determine which packages need to be installed.
+Traditionally this has been defined in the `depends` field of a projects opam
```
gridbugs, line 47:
```suggestion
Traditionally this has been defined in the `depends` field of a project's opam
```

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+file(s).
+
+Since version 1.10 Dune has supported {doc}`opam file generation
+</howto/opam-file-generation>` by specifying the package dependencies in the
+`dune-project`.
+
+The package management feature uses the same metadata, as Dune will determine
+the list of packages to install from the `depends` stanza in the `dune-project`
```
gridbugs, line 55:
```suggestion
the list of packages to install from the `depends` field in the `dune-project`
```
I think "stanza" has a specific definition within dune as a top-level sexp in a `dune` file (not a `dune-project` file).

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+For compatibility with a larger amount of existing projects, Dune will also
+collect dependencies from `.opam` files in the project. So while recommended,
+there is no obligation to switch to declaring dependencies in the
+`dune-project`. Likewise the generation of `.opam` files continues to be
+supported.
+
+## Locking
+
+Given the list of the project's transitive dependencies and their version
+constraints, the next steps are:
```
gridbugs, line 69: This makes it sound like the list of transitive dependencies is the input to the solver, but it's really just the list of dependencies from dune-project. The solver computes the transitive dependencies at the same time as choosing versions, as different choices of version may imply the inclusion/exclusion of packages from the transitive dependencies.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+
+1. Figure out a version for each dependency that satisifies the constraints
+2. For each dependency, download it, build it, and make it available to the
+   project
+
+In `opam`, `opam install` does both of these.
```
gridbugs, line 75: The way "opam" is used here feels to me like we're referring to the concept of opam rather than the opam executable, so I think the backticks are unnecessary.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+
+In `dune`, these are separate steps: the first one is `dune pkg lock`, and the
```
gridbugs, line 77: Same as above, we're referring the "dune way" of doing things, so it should be "Dune" rather than "`dune`".

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+each `dune pkg lock` will use the newest set of packages available.
+
+However, it is also possible to specify specific revisions of the repositories,
+to get a reproducible solution. Due to using Git, any previous revision of the
+repository can be used by specifying a commit hash.
+
+Dune defines two repositories by default:
+
+* `upstream` refers to the default branch of `opam-repository`, which
+  contains all the publically released packages.
```
gridbugs, line 141:
```suggestion
  contains all the publicly released packages.
```
Technically "publically" is also correct but much less commonly used. https://english.stackexchange.com/questions/45136/difference-between-publicly-and-publically

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+* `overlay` refers to
+  [opam-overlay](https://github.com/ocaml-dune/opam-overlays), which defines
+  packages patched to work with package management. The long-term goal is to
+  have as few packages as possible in this repository as more and more packages
+  work within Dune Package Management upstream. Check the
+  [compatibility](#compatibility) section for details.
+
+### Solving
+
+After Dune has retrieved the constraints and the set of possible packages, it
```
gridbugs, line 151: This suggests to me that dune has to do something to "retrieve" the packages and their version constraints, but that info is just parsed from dune-project.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+is necessary to determine which packages and versions should be selected for
+the package lock.
+
+To do so, Dune uses
+[`opam-0install-solver`](https://github.com/ocaml-opam/opam-0install-solver),
+which is a variant of the `0install` solver to find solutions for opam
```
gridbugs, line 157:
```suggestion
which is a variant of the [`0install`](https://github.com/0install/0install) solver to find solutions for opam
```

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+packages.
+
+Contrary to opam, the Dune solver always starts from a blank slate. It assumes
```
gridbugs, line 160:
```suggestion
Contrary to opam, the Dune solver always starts from a blank slate; it assumes
```

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+advantage that solving is now simpler, and previous solver solutions don't
+interfere with the current one. Thus, given the same inputs, it should always
+have the same outputs; no state is held between the solver runs.
+
+This can lead to more packages being installed (as opam won't install new
+package versions by default if the existing versions satisfy the constraints),
+but it avoids interference from already installed packages that lead to
+potentially different solutions.
+
+After solving is done, the solution gets written into the lock directory with
```
gridbugs, line 171: This is the first instance of the term "lock directory" on this page. Previous places on the page the same concept is referred to as the "package lock". I think you refer to it as a "lock directory" througout the page and introduce the term towards the beginning of the "Locking" section.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+:::{note}
+Solving and locking does not download the package sources. These are downloaded
+in a later step.
+:::
+
+(building)=
+## Building
+
+When building, Dune will read the information from the lock directory and set
+up rules to use the packages. Check {doc}`/explanation/mental-model` for
```
gridbugs, line 184:
```suggestion
up rules for the packages. Check {doc}`/explanation/mental-model` for
```

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+details about rules.
+
+The rules that the package management sets up include:
+
+* Fetch rules to download the sources as well as any additional sources like
+  patches and unpack them
```
gridbugs, line 190:
```suggestion
* Fetch rules to download and unpack the source archives, and also download any
additional sources such as patches
```

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+:::{note}
+`opam` also supports creating lock files. However, these are not as central to
+the opam workflow as they are in the case of package management in Dune, which
+always requires a set of locked packages.
+:::
+
+In the most general sense, a package lock is just a set of specific packages
+and their versions to be installed.
+
+A Dune lock file extends this to a directory with files that describe the
```
gridbugs, line 94:
```suggestion
A Dune lock file extends this to a directory (a "lock directory") with files that describe the
```

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+managed automatically by Dune. Thus, when cleaning the build directory, the
+installed packages are cleaned as well and will be reinstalled at the next
+build.
+
+(compatibility)=
+## Packaging for Dune Compatibility
+
+Dune can build and install most packages as dependencies, even if they are not
+built with Dune themselves. Dune will execute the build instructions from the
+lock file, very similar to opam.
```
gridbugs, line 210:
```suggestion
lock directory, very similar to opam.
```
We haven't introduced what a "lock file" is in the parlance of dune, so I think it's clearer to just refer to the lock directory here instead.

doc/explanation/package-management.md
```diff
@@ -0,0 +1,229 @@
...
+
+However, packages must meet certain requirements to be compatible with Dune.
+
+The most important one is that the packages must not use absolute paths to
+refer to files. That means they cannot read the path they are being built or
+installed in and expect this path to remain the same. Dune builds packages in a
+sandbox location, and after the build has finished, it moves the files to the
+actual destination.
+
+The reason for this is clear. On one hand it enables building without messing
```
gridbugs, line 220:
```suggestion
There are several reasons for building packages in a sandbox. On one hand it enables building without messing
```
I don't think claiming the reason is clear adds to the article.

gridbugs (commented):
Nice explanation! I've left some minor suggestions here and there.

doc/explanation/package-management.md
Leonidas-from-XIV: I don't really like to include things that can change easily within the scope of one PR as that will date the document and needs to be updated but I added a note on this.

doc/explanation/package-management.md
Leonidas-from-XIV: That's true. I've rewritten it a bit to be more accurate while (hopefully) be still readable.

doc/explanation/package-management.md
Leonidas-from-XIV: Technically it is kind of sort of true, but I agree that it sounds more grand than it is. The retrieved mostly refers to the candidate set; I've tried clearing that up a little.

doc/explanation/package-management.md
Leonidas-from-XIV: I think with your suggestion to introduce "lock directory" this seems good now.

PR #10947, opened 2024-09-23
test(pkg): reproduce not finding the patch file (`dune fmt`)

moyodiallo:
It reproduce https://github.com/ocaml/dune/issues/10903.

Issue #10903, opened 2024-09-11
Race condition building project with package management


gridbugs:
Sometimes when building a project with package management I see this error:
```
    Building ocaml-base-compiler.5.2.0  
    Building ocaml-config.3      
    Building base-unix.base      
    Building ocaml.5.2.0         
    Building seq.base              
    Building ocamlfind.1.9.6+dune  
Error:                             
_build/.sandbox/bed9cbdb5ceececaf0c3aaddc1db1624/_private/default/.dev-tool/odoc/ocamlfind/source/0001-Harden-test-for-OCaml-5.patch:
No such file or directory
-> required by _build/_private/default/.dev-tool/odoc/ocamlfind/target/cookie
-> required by - package topkg
-> required by - package astring
-> required by - package odoc
    Building ocamlbuild.0.15.0+dune
```

I've seen this on a range of projects, but so far I've only seen it on this specific patch from ocamlfind's package. I don't have a reliable reproduction case yet. It happens very infrequently, and the project always builds as expected when I rerun the command.

moyodiallo:
> This is not linked to race condition. I've tried to print the file path, running `dune fmt`. I assume you're testing with #10647. It seems the problem is coming from there.
> 
> It prints: `extra_files: _build/_private/default/.pkg/ocamlfind/source/0001-Harden-test-for-OCaml-5.patch`
> 
> instead of `.dev-tool` it is copying inside `.pkg`.

My bad, this is not true.

moyodiallo:
I noticed every time you do an initial `dune fmt` the error appears and after that any time you try to rebuild  without cleaning(removing the dev-tool locks), it's works.

I think there's a rule that coming before another one and it's not supposed to do that.

gridbugs:
I haven't seen this issue in over a year so there's a good chance it's been fixed accidentally. Optimistically closing.

PR #10951, opened 2024-09-23
fix(pkg): `dune fmt` missing extra files from the locks.

moyodiallo:
This PR follows https://github.com/ocaml/dune/pull/10947  in which we could see the reproduction of the bug from the issue https://github.com/ocaml/dune/issues/10903.

It turns out, the issue is not a race condition, the auto-locking/solve for the dev-tool save the lock files after the setup of the build. Since `dune` setup once in the same run, the patch files ends up missing during the build.

### DIFF
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -82,40 +82,12 @@ let run_build_system ~common ~request =
       Fiber.return ())
 ;;
 
-let lock_ocamlformat () =
-  if Lazy.force Lock_dev_tool.is_enabled
-  then
-    (* Note that generating the ocamlformat lockdir here means
-       that it will be created when a user runs `dune fmt` but not
-       when a user runs `dune build @fmt`. It's important that
-       this logic remain outside of `dune build`, as `dune
-       build` is intended to only build targets, and generating
-       a lockdir is not building a target. *)
-    Lock_dev_tool.lock_ocamlformat () |> Memo.run
-  else Fiber.return ()
-;;
-
-let run_build_command_poll_eager
-  ~with_lock_ocamlformat
-  ~(common : Common.t)
-  ~config
-  ~request
-  : unit
-  =
-  let open Fiber.O in
+let run_build_command_poll_eager ~(common : Common.t) ~config ~request : unit =
   Scheduler.go_with_rpc_server_and_console_status_reporting ~common ~config (fun () ->
-    Scheduler.Run.poll
-      (let* () = if with_lock_ocamlformat then lock_ocamlformat () else Fiber.return () in
-       run_build_system ~common ~request))
+    Scheduler.Run.poll (run_build_system ~common ~request))
 ;;
 
-let run_build_command_poll_passive
-  ~with_lock_ocamlformat
-  ~(common : Common.t)
-  ~config
-  ~request:_
-  : unit
-  =
+let run_build_command_poll_passive ~(common : Common.t) ~config ~request:_ : unit =
   (* CR-someday aalekseyev: It would've been better to complain if [request] is
      non-empty, but we can't check that here because [request] is a function.*)
   let open Fiber.O in
@@ -127,20 +99,16 @@ let run_build_command_poll_passive
   Scheduler.go_with_rpc_server_and_console_status_reporting ~common ~config (fun () ->
     Scheduler.Run.poll_passive
       ~get_build_request:
-        (let* () =
-           if with_lock_ocamlformat then lock_ocamlformat () else Fiber.return ()
-         in
-         let+ (Build (targets, ivar)) = Dune_rpc_impl.Server.pending_build_action rpc in
+        (let+ (Build (targets, ivar)) = Dune_rpc_impl.Server.pending_build_action rpc in
          let request setup =
            Target.interpret_targets (Common.root common) config setup targets
          in
          run_build_system ~common ~request, ivar))
 ;;
 
-let run_build_command_once ~with_lock_ocamlformat ~(common : Common.t) ~config ~request =
+let run_build_command_once ~(common : Common.t) ~config ~request =
   let open Fiber.O in
   let once () =
-    let* () = if with_lock_ocamlformat then lock_ocamlformat () else Fiber.return () in
     let+ res = run_build_system ~common ~request in
     match res with
     | Error `Already_reported -> raise Dune_util.Report_error.Already_reported
@@ -154,18 +122,6 @@ let run_build_command ~(common : Common.t) ~config ~request =
    | Yes Eager -> run_build_command_poll_eager
    | Yes Passive -> run_build_command_poll_passive
    | No -> run_build_command_once)
-    ~with_lock_ocamlformat:false
-    ~common
-    ~config
-    ~request
-;;
-
-let run_build_command_fmt ~(common : Common.t) ~config ~request =
-  (match Common.watch common with
-   | Yes Eager -> run_build_command_poll_eager
-   | Yes Passive -> run_build_command_poll_passive
-   | No -> run_build_command_once)
-    ~with_lock_ocamlformat:true
     ~common
     ~config
     ~request
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -82,12 +82,40 @@ let run_build_system ~common ~request =
       Fiber.return ())
 ;;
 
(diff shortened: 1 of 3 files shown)


bin/build_cmd.ml
```diff
@@ -82,12 +82,18 @@ let run_build_system ~common ~request =
       Fiber.return ())
 ;;
 
-let run_build_command_poll_eager ~(common : Common.t) ~config ~request : unit =
+let run_build_command_poll_eager ~pre_request ~(common : Common.t) ~config ~request : unit
```
rgrinberg, line 85: no "hooks" please. They make the code unreadable.

rgrinberg (changes requested):

gridbugs:
I think a simpler fix is to run a separate fiber scheduler to lock ocamlformat before running the build system. E.g.:

```diff
diff --git a/bin/build_cmd.ml b/bin/build_cmd.ml
index fec619125..66cf397cb 100644
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -230,20 +230,17 @@ let fmt =
       Common.Builder.set_promote builder (if no_promote then Never else Automatically)
     in
     let common, config = Common.init builder in
+    if Lazy.force Lock_dev_tool.is_enabled
+    then
+      (* Note that generating the ocamlformat lockdir here means that
+         it will be created when a user runs `dune fmt` but not when a
+         user runs `dune build @fmt`. It's important that this logic
+         remain outside of `dune build`, as `dune build` is intended
+         to only build targets, and generating a lockdir is not
+         building a target. *)
+      Scheduler.go ~common ~config (fun () ->
+        Memo.run (Lock_dev_tool.lock_ocamlformat ()));
     let request (setup : Import.Main.build_system) =
-      let open Action_builder.O in
-      let* () =
-        if Lazy.force Lock_dev_tool.is_enabled
-        then
-          (* Note that generating the ocamlformat lockdir here means
-             that it will be created when a user runs `dune fmt` but not
-             when a user runs `dune build @fmt`. It's important that
-             this logic remain outside of `dune build`, as `dune
-             build` is intended to only build targets, and generating
-             a lockdir is not building a target. *)
-          Action_builder.of_memo (Lock_dev_tool.lock_ocamlformat ())
-        else Action_builder.return ()
-      in
       let dir = Path.(relative root) (Common.prefix_target common ".") in
       Alias.in_dir ~name:Dune_rules.Alias.fmt ~recursive:true ~contexts:setup.contexts dir
       |> Alias.request

```

moyodiallo:
> I think a simpler fix is to run a separate fiber scheduler to lock ocamlformat before running the build system. E.g.:

I came with this idea at some point and it turns out running twice the scheduler wasn't accepted.

rgrinberg:
> I think a simpler fix is to run a separate fiber scheduler to lock ocamlformat before running the build system. E.g.:

You can do this in `dune fmt` as long as you don't change `dune build`.

moyodiallo:
A refactoring that avoid `dune build @fmt` to lock ocamlformat and also avoid to run the scheduler twice.

moyodiallo:
Another last review would be great. The hooks are removed and the scheduler runs only once.

@gridbugs @rgrinberg.

bin/build_cmd.ml
```diff
@@ -82,12 +82,40 @@ let run_build_system ~common ~request =
       Fiber.return ())
 ;;
 
-let run_build_command_poll_eager ~(common : Common.t) ~config ~request : unit =
+let lock_ocamlformat () =
+  if Lazy.force Lock_dev_tool.is_enabled
+  then
+    (* Note that generating the ocamlformat lockdir here means
+             that it will be created when a user runs `dune fmt` but not
```
gridbugs, line 89: nitpick: can you fix the indentation here so the comment text lines up on each line

bin/build_cmd.ml
```diff
@@ -82,12 +82,40 @@ let run_build_system ~common ~request =
...
+             when a user runs `dune build @fmt`. It's important that
+             this logic remain outside of `dune build`, as `dune
+             build` is intended to only build targets, and generating
+             a lockdir is not building a target. *)
+    Lock_dev_tool.lock_ocamlformat () |> Memo.run
+  else Fiber.return ()
+;;
+
+let run_build_command_poll_eager
+  ~with_lock_ocamlformat
```
gridbugs, line 99: I'm not a fan of introducing special behaviour into all these functions when a flag is passed. I think it muddies what was once a fairly clean interface with a special case which makes it harder to conceptualize what these functions do.

rgrinberg: Agreed. A flag is certainly better than a hook, but we should try to do better.

moyodiallo: I think we could at least agree upon https://github.com/ocaml/dune/pull/10951/commits/69ea9cf041bcea3e2d3ec1d2de5ebebd3adb7aaa at least or start from that.

gridbugs: What was the issue with running the scheduler multiple times? It seems like that would also solve this problem and IMO would be preferable to either a hook or a flag.

moyodiallo: I'm also OK to run the scheduler once because it avoid loading the project twice than having a refactoring preference.

gridbugs: If we're not going to solve this with a hook, a flag, or with running the scheduler twice, then I'm not sure what other options are possible. If/when we make solving dependencies into a build rule/action then this will become simpler. In the meantime it sounds like a flag is the least bad option. @rgrinberg how do you feel about merging this with the flag for now?

gridbugs (approved):
I'm happy with the flag solution provided that there's a comment explaining why it's necessary.

rgrinberg:
I'll take a look tomorrow

bin/build_cmd.ml
rgrinberg: I looked at the code again and I don't think this is adequate. Here's a plan that addresses most of my concerns:

1. Remove watch mode support from `fmt`. There's no point to it as it's essentially racing against the user to edit source files
2. Implement `dune fmt` by running the scheduler and then running the lock + build inside.

rgrinberg (changes requested):

bin/build_cmd.ml
moyodiallo: > There's no point to it as it's essentially racing against the user to edit source files

Is it because no one using `fmt` on watch mode, this is not addressed before ?

rgrinberg: Yes, probably nobody is using fmt in watch mode. It would be rather frustrating as it's always racing to update the file as you're editing it, possibly overwriting your changes.

moyodiallo:
>  1. Remove watch mode support from fmt. There's no point to it as it's essentially racing against the user to edit source files
> 2. Implement dune fmt by running the scheduler and then running the lock + build inside

The 2.  won't solve this issue because what we want is to run the lock and then run the scheduler. 
But I've got what the suggestions 1 & 2 means in the case of the PR and some of the work is done here https://github.com/ocaml/dune/pull/11044.

The suggestions are applied.

moyodiallo:
@gridbugs  What do you think about this simplification ?

ping @rgrinberg.

test/blackbox-tests/test-cases/pkg/ocamlformat/ocamlformat-patch-extra-files.t
```diff
@@ -3,17 +3,21 @@ ocamlformat and its dependencies during the same run of the "dune fmt" command.
 loaded before the lock files are produced, this is why any 'patch' file inside
 'dev-tools.locks/ocmalformat' is not copied inside the 'build' directory when a rule depends on it.
 
-In the case of this issue, there is a rule that depends on an 'patch' file in order to copy the file
+The issue was that there is a rule that depends on an 'patch' file in order to copy the file
 inside '_private/default/..' directory, since the file could not be copied, the rule is not activated.
 So it fails when 'dune' trying to apply the patch. After any follwing run of 'dune fmt', it works
 because the 'patch' file is already present.
 
+The issue is now fixed.
```
gridbugs, line 11: I don't think we should write this sort of comment in our tests. The test should describe the behaviour we're trying to preserve, but it's not necessary to describe the original problem or the fact that it was fixed. If anything, include a link to the github issue so people can read more about it there.

gridbugs:
I'm happy with the state of this PR. I've left some comments about some minor stylistic fixes to the test.

rgrinberg (approved):

PR #10952, opened 2024-09-24
refactor(pkg test): Avoid `gzip` on tarballs

Leonidas-from-XIV:
This avoids gzipping the tarballs that are created to test end to end functionality. Compressing and decompressing them takes time for minimal space saving so this PR just removes it.

It retains the compression in the tarballs test, to make sure compressed tarballs continue to work.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/depexts/error-message.t
+++ b/test/blackbox-tests/test-cases/pkg/depexts/error-message.t
@@ -8,7 +8,7 @@ Make a library that would fail when building it:
   $ mkdir foo
   $ cat > foo/dune-project <<EOF
   > EOF
-  $ tar -czf foo.tar.gz foo
+  $ tar cf foo.tar foo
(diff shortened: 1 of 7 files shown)


doc/dune.inc
```diff
@@ -269,6 +269,15 @@
...
 
+(rule
+ (with-stdout-to dune-tools.1
+  (run dune tools --help=groff)))
+
+(install
+ (section man)
+ (package dune)
+ (files   dune-tools.1))
+
```
moyodiallo, line 280:
```suggestion
```

This seems to be added by mistake.

moyodiallo (approved):
LGTM, just one thing about `doc/dune.inc` that added by mistake.

doc/dune.inc
Leonidas-from-XIV: Indeed a mistake; thanks, I've removed it.

PR #10953, opened 2024-09-24
fix(haiku): execve not allowed if other pthreads have already been created

Sylvain78:
Compiling for Haiku

### DIFF
--- a/otherlibs/stdune/src/proc.ml
+++ b/otherlibs/stdune/src/proc.ml
@@ -11,7 +11,7 @@ let restore_cwd_and_execve prog argv ~env =
   (* run at_exit before changing the working directory *)
   Stdlib.do_at_exit ();
   Sys.chdir (Path.External.to_string Path.External.initial_cwd);
-  if Sys.win32
+  if Sys.win32 || Platform.OS.value = Platform.OS.Haiku
   then (
     let pid = Unix.create_process_env prog argv env Unix.stdin Unix.stdout Unix.stderr in
     match snd (Unix.waitpid [] pid) with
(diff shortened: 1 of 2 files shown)


otherlibs/stdune/src/proc.ml
```diff
@@ -11,7 +11,7 @@ let restore_cwd_and_execve prog argv ~env =
   (* run at_exit before changing the working directory *)
   Stdlib.do_at_exit ();
   Sys.chdir (Path.External.to_string Path.External.initial_cwd);
-  if Sys.win32
+  if Sys.win32 || Platform.OS.value == Platform.OS.Haiku
```
anmonteiro, line 14:
```suggestion
  if Sys.win32 || Platform.OS.value = Platform.OS.Haiku
```

vendor/spawn/src/spawn_stubs.c
```diff
@@ -344,7 +344,6 @@ static void subprocess(int failure_fd, struct spawn_info *info)
   }
 
   pthread_sigmask(SIG_SETMASK, &info->child_sigmask, NULL);
-
```
anmonteiro, line 347: can you revert this whitespace diff?

anmonteiro (approved):

anmonteiro:
@rgrinberg do you think this needs a changelog entry?

rgrinberg:
I don't see why not. If this change is useful to anyone using Haiku, the change log entry would also be useful

Alizter:
@Sylvain78 Thank you for the context. I recall now seeing something similar to do this when I had a look at Haiku support before. What the Haiku designers will do upstream shouldn't be something we rely on obviously, so the fix that you've given here seems sensible and rigid.

Please add a changelog entry together with links to the Haiku issue and this should be good to go.

Issue #10959, opened 2024-09-27
Assertion failure in "dune exec -w" with package management enabled


liate7:
In the most recent dune developer preview build, `dune exec -w …` raises an assertion failure.

## Expected Behavior
Run the program, then restart/run the program on a change to the dependencies

## Actual Behavior
Prints the following stack trace, then exits 1.
```
File "src/dune_engine/build_system.ml", line 67, characters 11-17:
File "src/dune_engine/build_system.ml", line 67, characters 11-17: Assertion
failed
Raised at Dune_engine__Build_system.State.update_build_progress_exn in file
  "src/dune_engine/build_system.ml" (inlined), line 67, characters 11-23
Called from Dune_engine__Build_system.State.start_rule_exn in file
  "src/dune_engine/build_system.ml" (inlined), line 76, characters 4-122
Called from Dune_engine__Build_system.Exported.execute_rule_impl in file
  "src/dune_engine/build_system.ml", line 462, characters 41-66
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
Re-raised at Stdune__Exn.raise_with_backtrace in file
  "otherlibs/stdune/src/exn.ml" (inlined), line 38, characters 27-56
Called from Stdune__Exn_with_backtrace.reraise in file
  "otherlibs/stdune/src/exn_with_backtrace.ml", line 20, characters 33-71
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
Re-raised at Stdune__Exn.raise_with_backtrace in file
  "otherlibs/stdune/src/exn.ml" (inlined), line 38, characters 27-56
Called from Stdune__Exn_with_backtrace.reraise in file
  "otherlibs/stdune/src/exn_with_backtrace.ml", line 20, characters 33-71
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
Re-raised at Stdune__Exn.raise_with_backtrace in file
  "otherlibs/stdune/src/exn.ml" (inlined), line 38, characters 27-56
Called from Stdune__Exn_with_backtrace.reraise in file
  "otherlibs/stdune/src/exn_with_backtrace.ml", line 20, characters 33-71
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
[...]

PR #10960, opened 2024-09-27
test: reproduce `dune exec -w` crash with pkg management

anmonteiro:
reproduces the crash in #10959

### DIFF
--- a/bin/exec.ml
+++ b/bin/exec.ml
@@ -67,12 +67,16 @@ module Command_to_exec = struct
   (* A command to execute, which knows how to (re)build the program and then
      run it with some arguments in an environment *)
 
+  type command_env =
+    { path : Path.t
+    ; env : Env.t
+    }
+
   type t =
-    { get_path_and_build_if_necessary :
-        string -> (Path.t, [ `Already_reported ]) result Fiber.t
+    { get_env_and_build_if_necessary :
+        string -> (command_env, [ `Already_reported ]) result Fiber.t
     ; prog : string
     ; args : string list
-    ; env : Env.t
     }
 
   (* Helper function to spawn a new process running a command in an
@@ -90,12 +94,10 @@ module Command_to_exec = struct
 
   (* Run the command, first (re)building the program which the command is
      invoking *)
-  let build_and_run_in_child_process
-    ~root
-    { get_path_and_build_if_necessary; prog; args; env }
-    =
-    get_path_and_build_if_necessary prog
-    |> Fiber.map ~f:(Result.map ~f:(spawn_process ~root ~args ~env))
+  let build_and_run_in_child_process ~root { get_env_and_build_if_necessary; prog; args } =
+    get_env_and_build_if_necessary prog
+    |> Fiber.map
+         ~f:(Result.map ~f:(fun { path; env } -> spawn_process ~root ~args ~env path))
   ;;
 end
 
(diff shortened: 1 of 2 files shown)


Alizter:
```
  let update_build_progress_exn ~f =
```
is initialising when this happens.

Alizter:
For some reason this stems from the `let*` binding of `env` in `run_eagar_watch`. Not sure how it is being triggered yet however.

test/blackbox-tests/test-cases/pkg/gh10959.t
```diff
@@ -0,0 +1,42 @@
...
+
+  $ cat >dune <<EOF
+  > (executable
+  >  (name x))
+  > EOF
+  $ cat >x.ml <<EOF
+  > let () = print_endline "x"
+  > EOF
+
+  $ dune pkg lock
```
rgrinberg, line 35: Is this necessary? Shouldn't you be able to just write the lock directory in the test.

anmonteiro: thanks, I simplified the test without running `dune pkg lock`.

anmonteiro:
I just pushed a [commit](https://github.com/ocaml/dune/pull/10960/commits/084621fb3319872a3d7994f73bfbc9f42a387815) that delays evaluating `Env.t Memo.t` until `get_path_and_build_if_necessary`

test/blackbox-tests/test-cases/pkg/gh10959.t
```diff
@@ -34,6 +34,11 @@ Repro `dune exec --watch` crash with pkg management
   >  (system "echo building test"))
   > EOF
 
-  $ dune exec -w ./x.exe 2>&1 | grep -io "I must not crash"
-  I must not crash
-
+  $ dune exec -w ./x.exe > output.log 2>&1 &
+  $ PID=$!
+  $ sleep 2
```
Alizter, line 39: `2` seems long, what about `0.1`? The other `exec-watch` tests appear to pick something like that. Although killing processes this was seems very fragile, there ought to be a more principled way to test these. Not this PRs problem however.

bin/exec.ml
```diff
@@ -69,10 +69,9 @@ module Command_to_exec = struct
 
   type t =
     { get_path_and_build_if_necessary :
```
Alizter, line 71: Perhaps call this `get_env_and_build_if_necessary`.

test/blackbox-tests/test-cases/pkg/gh10959.t
Alizter: The watching/ tests use some better helpers. Up to you if you want to improve things. Also I would move this from `pkg` to `exec-watch`.

Alizter: They have these `start_dune` `stop_dune` helpers from a shared script (see the dune file for where it is). A test example is: https://github.com/ocaml/dune/blob/main/test/blackbox-tests/test-cases/watching/copy-rules.t

The logic in the script handles stopping the watch mode build and the output. Presumably something like that could be adapted for `exec`. But if this is too much work that is also fine, since the other `exec-watch/` tests are written this way.

anmonteiro:
before addressing the outstanding code review items, I'd like some confirmation that the fix is in the right direction, or even correct.

rgrinberg:
The fix looks good to me

test/blackbox-tests/test-cases/pkg/gh10959.t
anmonteiro: so this actually takes a bit longer so I added a loop that sleeps for 0.1 seconds (with a timeout) until we get the expected number of lines in the output file

anmonteiro:
should be ready for another look

rgrinberg (approved):
A bit worried about this test being flaky. I suppose we can disable if it proves to be a problem.

anmonteiro:
Absolutely. In my tests it was always under the allocated time budget so we can disable it if it regresses.

maiste:
> before addressing the outstanding code review items, I'd like some confirmation that the fix is in the right direction, or even correct.
Does this PR also fix the problem, or is it the test to find the problem?

anmonteiro:
this PR should also have fixed the issue. is it still reproducible for you?

maiste:
Sorry, it was a misunderstanding on my side with the title of the PR. I tried it and it definitely fixes the problem, thanks :pray:

Issue #10674, opened 2024-06-27
Internal error: Invalid Package_version.t


OlivierNicole:
## Expected Behavior

`dune build @install -j auto --profile dev` should work.

## Actual Behavior

```
Internal error, please report upstream including the contents of _build/log.
Description:
  ("Invalid Package_version.t", { s = "" })
Raised at Stdune__Code_error.raise in file
  "otherlibs/stdune/src/code_error.ml", line 10, characters 30-62
Called from Stdune__Option.O.(>>|) in file "otherlibs/stdune/src/option.ml",
  line 9, characters 21-26
Called from Stdune__Option.map in file "otherlibs/stdune/src/option.ml"
  (inlined), line 22, characters 15-28
Called from Dune_rules__Package.load_opam_file_with_contents in file
  "src/dune_rules/package.ml" (inlined), line 74, characters 35-74
Called from Dune_rules__Package.load_opam_file_with_contents in file
  "src/dune_rules/package.ml", line 74, characters 13-75
Called from Fiber__Core.O.(>>|).(fun) in file "vendor/fiber/src/core.ml",
  line 253, characters 36-41
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
-> required by ("find-dir-raw", In_source_tree ".")
-> required by ("dune_load", ())
-> required by ("Super_context.all", ())
-> required by ("toplevel", ())
```

build.log:
```
# dune build @install -j auto --profile dev
# OCAMLPARAM: unset
# Shared cache: disabled
# Shared cache location: /home/olivier/.cache/dune/db
# Workspace root: /home/olivier/hacking_days/pyre-check/source
# Auto-detected concurrency: 8
# Dune context:
#  { name = "default"
#  ; kind = "default"
#  ; profile = Dev
#  ; merlin = true
#  ; fdo_target_exe = None
#  ; build_dir = In_build_dir "default"
#  ; instrument_with = []
#  }
```

## Reproduction

- Clone https://github.com/OlivierNicole/pyre-check/tree/dune_bug (the repo is large, you may want to shallow clone)
- `cd pyre-check`
- `./scripts/setup.sh --local`

## Specifications

- Version of `dune` (output of `dune --version`): 3.16.0
- Version of `ocaml` (output of `ocamlc --version`): 5.2.0
- Operating system (distribution and version): NixOS 24.05

OlivierNicole:
[...]

PR #10965, opened 2024-09-28
reproduce bug in #10674 with empty .opam version field

Alizter:
This reproduces the code error in #10674 and fixes it with a nicer user error stating the issue.


test/blackbox-tests/test-cases/package-version-empty.t/run.t
```diff
@@ -0,0 +1,9 @@
+Testing dune #10674 where an empty version in an opam file caused a code error
+in dune. We should make sure that this case is handled gracefully.
+
+  $ dune build
+  File "foo.opam", line 2, characters 0-7:
+  2 | version: ""
+      ^^^^^^^
+  Error: "" is an invalid package version.
```
rgrinberg, line 8: In the future, perhaps you could tell the user what's a valid version. E.g. a non empty string.

PR #10969, opened 2024-09-29
update nixpkgs in flake

Alizter:
This allows for `./dune.exe build @check` to run again in the `nix develop`er mode.

I tried updating the entire flake, but there are build errors in ocaml-lsp so I stuck to nixpkgs for the time being.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -121,11 +121,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1726142289,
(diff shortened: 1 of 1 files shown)


Alizter:
@rgrinberg We did it at the same time. :laughing:

Alizter:
@rgrinberg Let's go ahead with this for the time being so we can at least get a usable dev env.

Issue #10872, opened 2024-09-02
CRAM tests on windows


pirbo:
## Expected Behavior
I'm not sure... Low hanging fruit: a warning saying "CRAM tests skipped, `sh` cannot be found in the path."

## Actual Behavior
```
Internal error, please report upstream including the contents of _build/log.
Description:
  ("Option.value_exn", {})
Raised at Stdune__Code_error.raise in file
  "otherlibs/stdune/src\\code_error.ml", line 10, characters 30-62
Called from Dune_rules__Cram_exec.run in file
  "src/dune_rules\\cram\\cram_exec.ml", line 427, characters 6-45
Called from Fiber__Scheduler.exec in file "vendor/fiber/src\\scheduler.ml",
  line 76, characters 8-11
Re-raised at Stdune__Exn.raise_with_backtrace in file
  "otherlibs/stdune/src\\exn.ml" (inlined), line 38, characters 27-56
[...]
```

Line 427 of src/dune_rules/cram/cram_exec.ml is `Option.value_exn (Bin.which ~path "sh")` :)

## Reproduction

@sudha247 discovered the bug while testing if the "ocaml platform" was windows yet (and I investigated it). It is raised by, at least, ocamlformat.0.26.0 and odoc.2.4.2 but I mean any cram test anywhere would raise it

1. open a Powershell or cmd.exe
1. opam source ocamlformat
1. cd ocamlformat.0.26.0
1. dune runtest

CAUTION: dont be confused as I've been at first: `opam install -t ocamlformat` is ~working~ failing with a different set of errors (a ton of diffs where `/`s have been replaced by some `\`s...) I mean (cygwin) `sh` IS found in this case because "opam install" extends your environment with enough of the cygwin_opam_installs for that)

PS: nope sadly `opam exec -- dune runtest` is not sufficient to get (cygwin) `sh` in your path...  If/When cygwin executable are added to the path by opam is a mystery to me, I would redirect the question to @dra27 ;)

## Specifications

- Version of `dune` (output of `dune --version`): 3.16.0 and 3.15.2
- Version of `ocaml` (output of `ocamlc --version`): 5.2.0
- Operating system (distribution and version): Windows (11 VM) opam.2.2.1 install by winget

nojb:
> "CRAM tests skipped, `sh` cannot be found in the path."

Sounds reasonable. PR welcome!

emillon:
[...]

PR #10978, opened 2024-09-30
fix(cram): improve error message on sh unavailable

maiste:
This PR improves the message of the error mentioned in #10872. It is a simple improvement to make the error more readable and provide a hint when users forget to add the `sh` dependency.

However, as it happens on Windows, I can't test the behaviour. @rgrinberg do you think it is fine to not add a test here as it only improves an error message?

### DIFF
--- a/src/dune_rules/cram/cram_exec.ml
+++ b/src/dune_rules/cram/cram_exec.ml
@@ -424,7 +424,12 @@ let run ~env ~script lexbuf : string Fiber.t =
   let+ () =
     let sh =
       let path = Env_path.path Env.initial in
-      Option.value_exn (Bin.which ~path "sh")
+      match Bin.which ~path "sh" with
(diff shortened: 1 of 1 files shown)


src/dune_rules/cram/cram_exec.ml
```diff
@@ -424,7 +424,12 @@ let run ~env ~script lexbuf : string Fiber.t =
   let+ () =
     let sh =
       let path = Env_path.path Env.initial in
-      Option.value_exn (Bin.which ~path "sh")
+      match Bin.which ~path "sh" with
+      | Some sh -> sh
+      | None ->
+        let hints = Pp.[ text "Try to add (deps %{bin:sh} ) to the (cram) stanza" ] in
```
rgrinberg, line 430: Does this suggestion do anything?

PR #10979, opened 2024-09-30
Fix doc/reference/dune/executable.rst: add missing executables libraries

edwinans:


doc/reference/dune/executable.rst
```diff
@@ -304,7 +304,8 @@ simplify the code:
 .. code:: dune
 
    (executables
-     (names foo bar))
+     (names foo bar)
+     (libraries baz))
```
nojb, line 308: This is actually correct as it is, even if it is not written as clearly as it could be.

The stanza
```
(executables
 (names foo bar))
```
means "define two executables, `foo.exe` and `bar.exe`, using _all_ modules present in the present directory (namely, `foo.ml`, `bar.ml` and `baz.ml`)", where it is implicit that both `foo.ml` and `bar.ml` refer to `baz.ml`.

Since it is not possible to use modules (like `baz.ml`) in more than one stanza, if one wants to "unroll" the `executables` stanza into individual `executable` stanzas, then one needs to put `baz.ml` into a separate library and depend on it from both individual `executable` stanzas, which is what is illustrated below.

edwinans: Thanks for the explanation, it was indeed not trivial at first sight.
Consequently, I'm closing this PR.

Issue #10839, opened 2024-08-20
Dependency cycle when building the `lwt` project with dune package management


gridbugs:
To reproduce, run:
```
$ opam source lwt.5.7.0
$ cd lwt.5.7.0
$ dune pkg lock
$ DUNE_CONFIG__TOOLCHAINS=enabled dune build
Error: Dependency cycle between:
   - package base-bytes
-> _build/_private/default/.pkg/ocamlfind/target/cookie
-> - package base-bytes
```

There's no obvious dependency cycle between lockfiles. Projects that depend on `lwt` build fine with dune package management. This problem only happens when building the `lwt` project itself.

I'll update this issue as I find more projects with this problem.

rgrinberg:
I did get a different cycle though:
```
Dependency cycle between:
   - package react
-> _build/_private/default/.pkg/ocamlfind/target/cookie
-> - package topkg
-> - package react
```

Leonidas-from-XIV:
After removing the `ocamlfind` dependency from `base-bytes` I got the same error as Rudi:

```
Error: Dependency cycle between:
   - package react
-> _build/_private/default/.pkg/ocamlfind/target/cookie
-> - package topkg
-> - package react
```

The only other dependencies on `ocamlfind` are `topkg` and `react` (which depends on `ocamlfind` and `topkg`) and removing them unlocks the dependency cycle but fails to build. It seems there's something weird about `ocamlfind`, but we don't special-case it in the package rule code and the lock result looks totally correct.

PR #10980, opened 2024-09-30
test(pkg): repro 10839

rgrinberg:
OCaml syntax doesn't work with package management


gridbugs:
Nice find. Linking to https://github.com/ocaml/dune/issues/10839.

gridbugs (approved):

PR #10982, opened 2024-10-01
Fix exec -w for relative paths with --root argument

gridbugs:
Passing --root to `dune build` and `dune exec` causes relative paths to files to be resolved relative to the workspace root rather than the working directory (in addition to its main function of explicitly setting the workspace root directory). This was not implemented correctly for exec watch mode, where relative paths would be resolved relative to the working directory instead.

There was already a test for this which was failing, however the test is disabled in CI as it is known to be flaky.

### DIFF
--- a/bin/exec.ml
+++ b/bin/exec.ml
@@ -77,22 +77,26 @@ module Command_to_exec = struct
 
   (* Helper function to spawn a new process running a command in an
      environment, returning the new process' pid *)
-  let spawn_process path ~args ~env =
+  let spawn_process (common : Common.t) path ~args ~env =
     let pid =
-      let path = Path.to_string path in
+      let prog = string_path_relative_to_specified_root common (Path.to_string path) in
       let env = Env.to_unix env |> Spawn.Env.of_list in
-      let argv = path :: args in
+      let argv = prog :: args in
       let cwd = Spawn.Working_dir.Path Fpath.initial_cwd in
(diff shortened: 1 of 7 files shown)


bin/import.ml
```diff
@@ -237,14 +237,16 @@ module Scheduler = struct
...
-    if Filename.is_relative prog
-    then (
-      let root = Common.root common in
-      Filename.concat root.dir prog)
-    else prog
-  in
+let string_path_relative_to_specified_root (common : Common.t) path =
+  if Filename.is_relative path
+  then (
+    let root = Common.root common in
```
rgrinberg, line 243: This function is taking `common` as an argument but is only using one field from it. How about just making it take `root`?

rgrinberg (changes requested):

bin/import.ml
gridbugs: done

bin/exec.ml
```diff
@@ -77,22 +77,28 @@ module Command_to_exec = struct
 
   (* Helper function to spawn a new process running a command in an
      environment, returning the new process' pid *)
-  let spawn_process path ~args ~env =
+  let spawn_process (common : Common.t) path ~args ~env =
```
rgrinberg, line 80: Same thing here. A function taking `common` as an argument is essentially function that takes every single configuration value and command line flag. It should just take `Common.root` as that' the only thing it needs from this record.

 I can fix this one

PR #10983, opened 2024-10-01
Clear console before running command in exec -w

### LINKED ISSUE #10922: Missing newline in "dune exec -w"
## Expected Behavior

When running "dune exec" with watch mode on (`-w/--watch`), it would be better to show program's output on a new line. Currently it's appended to the message "Success, waiting for filesystem changes...".

Similar to #7142 but with watch mode on.

```
Success, waiting for filesystem changes...
Hello, dune!
```

## Actual Behavior

```
Success, waiting for filesystem changes...Hello, dune!
```

## Reproduction

```
$ dune init project test
Entering directory '/private/tmp/test'
Success: initialized project component named test

$ cd test
$ dune exec -w test
Success, waiting for filesystem changes...Hello, World!
Hello, World!
Success, waiting for filesystem changes...
Success, waiting for filesystem changes...Hello, dune!
```

## Specifications

- Operating system (distribution and version): macOS 14.6.1 (darwin23.0)

gridbugs:
This fixes an issue where the build progress stays on the console after the build completes, and the first line of the output of the command being exec'd is printed on teh same line as the build progress.

A side effect of this change is that some of the "Build Successful" messages in tests get truncated.

Fixes https://github.com/ocaml/dune/issues/10922

### DIFF
--- a/bin/exec.ml
+++ b/bin/exec.ml
@@ -94,10 +94,17 @@ module Command_to_exec = struct
 
   (* Run the command, first (re)building the program which the command is
      invoking *)
-  let build_and_run_in_child_process ~root { get_env_and_build_if_necessary; prog; args } =
+  let build_and_run_in_child_process
+    ~root
+    ~config
+    { get_env_and_build_if_necessary; prog; args }
+    =
     get_env_and_build_if_necessary prog
     |> Fiber.map
-         ~f:(Result.map ~f:(fun { path; env } -> spawn_process ~root ~args ~env path))
+         ~f:
+           (Result.map ~f:(fun { path; env } ->
+              Scheduler.maybe_clear_screen ~details_hum:[] config;
+              spawn_process ~root ~args ~env path))
   ;;
 end
 
@@ -144,18 +151,18 @@ module Watch = struct
 
   (* Kills the currently running process, then runs the given command after
      (re)building the program which it will invoke *)
-  let run ~root state ~command_to_exec =
+  let run ~root ~config state ~command_to_exec =
     let open Fiber.O in
     let* () = Fiber.return () in
     let* () = kill_currently_running_process state in
     let* command_to_exec = command_to_exec () in
-    Command_to_exec.build_and_run_in_child_process ~root command_to_exec
+    Command_to_exec.build_and_run_in_child_process ~root ~config command_to_exec
     >>| Result.map ~f:(fun pid -> state.currently_running_pid := Some pid)
   ;;
 
-  let loop ~root ~command_to_exec =
+  let loop ~root ~config ~command_to_exec =
     let state = init_state () in
-    Scheduler.Run.poll (run ~root state ~command_to_exec)
+    Scheduler.Run.poll (run ~root ~config state ~command_to_exec)
   ;;
 end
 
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/exec-watch/exec-watch-basic.t/run.t
```diff
@@ -10,16 +10,16 @@ between each change to its code.
   > EOF
 
   $ dune exec --watch ./foo.exe &
-  Success, waiting for filesystem changes...
+   changes...
```
maiste, line 13: Why does it remove the entire sentence, minus the last word, here?

gridbugs: No idea! It only happens in cram tests - not when running dune from a terminal. I've tried various `Console` functions to see if I could fix the test behaviour while also maintaining the behaviour when you run the command interactively but nothing works. Maybe something to do with how cram tests capture the output of commands? @rgrinberg any ideas?

rgrinberg: No idea, sorry. But shouldn't you restore the terminal after the process finishes?

maiste (approved):

bin/exec.ml
```diff
@@ -95,7 +95,11 @@ module Command_to_exec = struct
     { get_path_and_build_if_necessary; prog; args; env }
     =
     get_path_and_build_if_necessary prog
-    |> Fiber.map ~f:(Result.map ~f:(spawn_process ~root ~args ~env))
+    |> Fiber.map
+         ~f:
+           (Result.map ~f:(fun prog ->
+              Console.reset ();
```
Alizter, line 101: Rather than doing this directly, let's be consistent with `dune build` watch mode.
```
              Scheduler.maybe_clear_screen ~details_hum:[] config;
```
You will need to thread `config` from `run_eagar_watch`.

Alizter:
Are we aware that all the `exec-watch` tests are disabled? i.e. `(enable_if false)`.

gridbugs:
> Are we aware that all the `exec-watch` tests are disabled? i.e. `(enable_if false)`.

Yep, there's a big comment [here](https://github.com/ocaml/dune/blob/d27724b4f0ee27607f78be39191cb9cda0306645/test/blackbox-tests/test-cases/exec-watch/dune#L3) explaining why (they're flakey).

bin/exec.ml
gridbugs: Thanks for this suggestions. This somehow fixed the issues in tests where the success messages was being truncated.

gridbugs: Oh I see, because `maybe_clear_screen` has different behaviour when running inside the dune repo than when running in the wild.

test/blackbox-tests/test-cases/exec-watch/exec-watch-basic.t/run.t
gridbugs: It's non-trivial to restore the terminal after the process finishes because dune doesn't wait for the process to finish and continues to run in watch mode, printing to the terminal itself while the new process runs in the background. This means that running programs that change terminal settings (such as programs with tuis) from exec watch mode will have a bad ux, especially because after every change dune will kill the process and launch a new process, possibly leaving the terminal in a bad state. It's definitely something we could improve in the future but not necessary to solve this issue.

PR #10984, opened 2024-10-02
doc: Add versions when opam-file related stanzas were added

Leonidas-from-XIV:
There is no note about `generate_opam_files` in the changelog so the only reference seems to be the source. This PR adds that information.

It's per-stanza as the versions where each of the stanzas were added is quite varied, some were even added before opam file generation existed.

### DIFF
--- a/doc/reference/dune-project/generate_opam_files.rst
+++ b/doc/reference/dune-project/generate_opam_files.rst
@@ -3,6 +3,8 @@ generate_opam_files
 
 .. describe:: (generate_opam_files ...)
 
+   .. versionadded:: 1.10
+
    Use metadata specified in the ``dune-project`` file to generate ``.opam``
    files.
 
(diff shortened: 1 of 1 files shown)


maiste:
How do you find the versions when it was introduced?

moyodiallo:
> How did you find the versions when it was introduced? (apart from generate_opam_files whose the first mention was used)

I'm also noticing different versions.

Leonidas-from-XIV:
I created an empty project with `(lang dune 1.0)` and let `dune build` tell me the version the dune-lang that the offending stanzas required. I could've also checked the source but this approach is *TDD*! :sweat_smile:

maiste (approved):
Try it on my side and get the same number. If this is OK, we could definitely merge it.

ElectreAAS (approved):

moyodiallo (approved):

Issue #10985, opened 2024-10-02
`dune build @ocaml-index` fails when depending on a package with package management enabled


liate7:
This only happens when there's a `dune.lock` and one of the dune files in the project depend on a library from outside the project.

## Expected Behavior
`dune build @ocaml-index` works and builds the index.

## Actual Behavior
`dune build @ocaml-index` returns 1 with error messages complaining about rules defining directory targets when the rules' actions didn't produce them.  For a dummy project only depending on `containers`, this was:
```
Error: This rule defines a directory target "default/.pkg/containers/target"
that matches the requested path
"default/.pkg/containers/target/lib/containers/cctx.ocaml-index" but the
rule's action didn't produce it
-> required by _build/default/lib/.test.objs/cctx.ocaml-index
-> required by alias ocaml-index
Error: This rule defines a directory target "default/.pkg/either/target" that
matches the requested path
"default/.pkg/either/target/lib/either/cctx.ocaml-index" but the rule's
action didn't produce it
-> required by _build/default/lib/.test.objs/cctx.ocaml-index
-> required by alias ocaml-index
Error: This rule defines a directory target "default/.pkg/containers/target"
that matches the requested path
"default/.pkg/containers/target/lib/containers/monomorphic/cctx.ocaml-index"
but the rule's action didn't produce it
-> required by _build/default/lib/.test.objs/cctx.ocaml-index
-> required by alias ocaml-index
```

## Reproduction

```
mkdir test
cd test
cat > dune-project <<EOF
(lang dune 3.17)
(package
  (name test)
  (depends ocaml dune containers))
EOF
mkdir bin
cat >bin/dune <<EOF
(executable
  (name main)
  (libraries containers))
EOF
touch bin/main.ml
dune pkg lock
dune build @ocaml-index
```

Edit: I've written a pull request for the bug: #10986 

## Specifications

- Version of `dune` (output of `dune --version`): "Dune Developer Preview: build 2024-10-02T07:47:55+00:00, git revision 451e5b4463c4fe5b4f5427c0a40a5962c38ccc0f"
- Version of `ocaml` (output of `ocamlc --version`)
- Operating system (distribution and version): Linux 6.10.7, Guix System

## Additional information 

- Link to gist with verbose output (run `dune` with the `--verbose` flag): [`dune build @ocaml-index --verbose`](https://gist.github.com/liate7/efc3bef907c5b2426daca65bd9a44fc9)

PR #10986, opened 2024-10-03
test: reproduce `@ocaml-index` failing with package management

liate7:
reproduces the build failure in #10985


test/blackbox-tests/test-cases/ocaml-index/gh10985/run.t
```diff
@@ -0,0 +1,23 @@
+This test attempts to build the ocaml index while depending on a library
+installed through a lock file.
+
+This is based off of the pkg/libraries.t test, just taking advantage of
+being in its own directory.
+
+This can't be just written to disk, I /think/ so it cat get an absolute path
+in the source stanza.
```
Leonidas-from-XIV, line 8: I don't understand this paragraph.

test/blackbox-tests/test-cases/ocaml-index/gh10985/run.t
```diff
@@ -0,0 +1,23 @@
+This test attempts to build the ocaml index while depending on a library
+installed through a lock file.
+
+This is based off of the pkg/libraries.t test, just taking advantage of
+being in its own directory.
```
Leonidas-from-XIV, line 5: I thing generally we try to have issues in a single cram file unless it is too complicated. This looks fairly straightforward.

Leonidas-from-XIV (changes requested):
Thanks for the repro case, these are a great way to test potential bugfixes. Here's my observations on the PR.

test/blackbox-tests/test-cases/ocaml-index/gh10985/run.t
liate7: The new commit is in a single file

test/blackbox-tests/test-cases/ocaml-index/gh10985.t
```diff
@@ -0,0 +1,55 @@
...
+  > EOF
+
+  $ cat >dune.lock/mypkg.pkg <<EOF
+  > (version 0.0.1)
+  > (source (copy $PWD/external_sources))
+  > (build (run dune build --release --promote-install-file=true . @install))
+  > EOF
+
+  $ cat >dune <<EOF
+  > (dirs (:standard \ external_sources))
```
Leonidas-from-XIV, line 38: You can probably avoid this by putting `external_sources` in one folder, the repo that uses it in another and that refer to it via `(copy $PWD/../external_sources)`.

Leonidas-from-XIV (approved):

test/blackbox-tests/test-cases/ocaml-index/gh10985.t
liate7: Right now, this file is the same as `test/blackbox-tests/test-cases/pkg/libraries.t`, just with a slightly different comment and building the index instead of `foo.cma`.  I literally started by copying over that file.  I can definitely still make this change, it just seems weird to change this from the directly analogous working test.

(If I don't change it, maybe it's worth noting this in the file, instead of just in the commit description?)

liate7: I just made this change.  If this really matters, I can also send a separate pull request to update `pkg/libraries.t`.

rgrinberg:
Merged in https://github.com/ocaml/dune/pull/10986. Thanks

PR #10990, opened 2024-10-04
pkg: fix dev-tool bug where `dune fmt` would revert file

gridbugs:
The rule for running ocamlformat on source files when dev-tools are in use did not depend on the input file, so changes to the input file wouldn't cause ocamlformat to be re-run on the updated file. The consequence of this is that `dune fmt` would promote stale versions of files.

### DIFF
--- a/src/dune_rules/format_rules.ml
+++ b/src/dune_rules/format_rules.ml
@@ -97,7 +97,9 @@ module Ocamlformat = struct
     let open Action_builder.With_targets.O in
     (* Depend on [extra_deps] so if the ocamlformat config file
        changes then ocamlformat will run again. *)
-    extra_deps dir >>> action
+    extra_deps dir
+    >>> action
+    |> With_targets.map ~f:(Action.Full.add_sandbox Sandbox_config.needs_sandboxing)
   ;;
 
   let action_when_ocamlformat_isn't_locked ~input kind =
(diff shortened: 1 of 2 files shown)


gridbugs:
Fixes https://github.com/ocaml/dune/issues/10991

moyodiallo:
Would you like add the from https://github.com/ocaml/dune/pull/10993 ?

moyodiallo (approved):

rgrinberg:
Fix looks good, but I think we need to improve things further to avoid mistakes like this. I would suggest the following steps:

1. Demonstrate the bug in a test
2. Make sure the formatting action is sandboxed when using package management
3. Apply this fix

It might be easier to do 2. followed by 1.

maiste (approved):

gridbugs:
>     2. Make sure the formatting action is sandboxed when using package management

How would sandboxing the formatting action prevent mistakes like this?

gridbugs:
> If sandboxing was an enabled for the rule, the action would always fail with some error saying that it was unable to find the input file. Without sandboxing, the rule sometimes succeeds b/c the input was already in _build due to being needed for some other rule.

Ok I added sandboxing and confirmed that without the fix the formatting action can't find its input file.

rgrinberg (approved):

PR #11007, opened 2024-10-09
fix(pkg): add location hash mismatch error

rgrinberg:
We use the lock directory as the location rather than pointing at the hash, as the user shouldn't really touch the hash.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/lock-out-of-sync.t
+++ b/test/blackbox-tests/test-cases/pkg/lock-out-of-sync.t
@@ -41,6 +41,7 @@ We add the bar dependency to the test package
 
 It fails as we have not regenerated the lock:
   $ dune build
+  File "dune.lock/lock.dune", line 1, characters 0-0:
   Error: The lock dir is not sync with your dune-project
   Hint: run dune pkg lock
   [1]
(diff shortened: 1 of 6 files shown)


test/blackbox-tests/test-cases/pkg/lock-out-of-sync.t
```diff
@@ -41,6 +41,7 @@ We add the bar dependency to the test package
 
 It fails as we have not regenerated the lock:
   $ dune build
+  File "dune.lock", line 1, characters 0-0:
```
maiste, line 44: Isn't it more confusing to have a pointer to the file with specific line and characters instead of just the instruction to run `dune pkg lock`? I would understand with just the filename (as a dir) but here, it might create the inverted effect and push the user to have a look at the directory. Wouldn't it be possible to have `Directory "dune.lock"` instead of `File "dune.lock", line 1, characters 0-0`?

rgrinberg: I could pick the `lock.dune` file inside `dune.lock`.

maiste: You are right. It might be more understandable from a user perspective.

maiste (approved):

Issue #10987, opened 2024-10-03
pkg: inscrutable error when depending on a non-existent package


v-gb:
## Expected Behavior

If the dune-project file mentions a package that doesn't exist in the `depends` field, I'd expect an error saying so when calling `dune pkg lock`.

## Actual Behavior

Instead, dune prints screenfuls of confusing information that mention various problems, almost all of which is irrelevant to the problem (but seems like it could be relevant).

## Reproduction

Given a dune-project with:

```
(lang dune 3.15)
(name abc)
(package
 (name abc)
 (synopsis "A short synopsis")
 (description "A longer description")
 (depends
   ocaml
   dune
   base
   re
   unix
   core
   core_unix
   ocamlformat
   ocamlformat-lib
 )
 (tags
  (topics "to describe" your project)))
```

I get this error when running `dune pkg lock`:

```sh
$ dune --version
"Dune Developer Preview: build 2024-09-28T01:30:13+00:00, git revision
17071ec30d10390badcb6cb1f6a43984b1be54a6"
$ dune pkg lock
Error: Unable to solve dependencies for the following lock directories:
Lock directory dune.lock:
Can't find all required versions.
Selected: base.v0.17.1 base-threads.base base-unix.base
          base_bigstring.v0.17.0 base_quickcheck.v0.17.0 bin_prot.v0.17.0
          capitalization.v0.17.0 cmdliner.1.3.0 core.v0.17.1
          core_kernel.v0.17.0 core_unix.v0.17.0 csexp.1.5.2
          dune-configurator.3.16.0 expect_test_helpers_core.v0.17.0
          fieldslib.v0.17.0 gel.v0.17.0 int_repr.v0.17.0
          jane-street-headers.v0.17.0 jst-config.v0.17.0 migration.dev
          num.1.5-1 ocaml.5.4.0 ocaml-config.3
          ocaml_intrinsics_kernel.v0.17.1 ocamlformat.0.26.2 parsexp.v0.17.0
          ppx_assert.v0.17.0 ppx_base.v0.17.0 ppx_bench.v0.17.0
          ppx_bin_prot.v0.17.0 ppx_cold.v0.17.0 ppx_compare.v0.17.0
          ppx_custom_printf.v0.17.0 ppx_diff.v0.17.0
          ppx_disable_unused_warnings.v0.17.0 ppx_enumerate.v0.17.0
          ppx_expect.v0.17.0 ppx_fields_conv.v0.17.0
[...]

Issue #10838, opened 2024-08-20
Solver error when one local package depends on another local package with version constraints


gridbugs:
Consider this example from https://github.com/shonfeder/alg_structs:
```
(version 0.1.3)

(package
 (name alg_structs)
 (depends
  (dune (>= 1.11.3))
  (ocaml (>= 4.08.0))
  (ppx_deriving (>= 4.4))
))

(package
 (name alg_structs_qcheck)
 (depends
  (dune (>= 1.11.3))

  (ocaml (>= 4.08.0))
  (alg_structs (and (>= 0.1.3) (< 0.2.0)))
  (qcheck (>= 0.11))

  (alcotest (and :with-test (>= 0.8.5)))
  (qcheck-alcotest (and :with-test (>= 0.11)))
))
```

The second package `alg_structs_qcheck` depends on the first package with version constraints `(alg_structs (and (>= 0.1.3) (< 0.2.0)))`. Solving this project with dune gives the error:
```
 $ dune pkg lock
Error: Unable to solve dependencies for the following lock directories:
Lock directory dune.lock:
Can't find all required versions.
Selected: alg_structs.dev base-threads.base base-unix.base cppo.1.6.9
          ocaml.5.3.0 ocaml-config.3 ocamlfind.1.9.6+dune ppx_derivers.1.2.1
          ppx_deriving.6.0.2 alg_structs&alg_structs_qcheck ocaml
          ocaml-base-compiler ocaml-base-compiler
- alg_structs_qcheck -> (problem)
    Rejected candidates:
      alg_structs_qcheck.dev: Requires alg_structs >= 0.1.3 & < 0.2.0
- dune -> dune.3.16.0
    User requested = 3.17
- ocaml-base-compiler -> (problem)
    Rejected candidates:
      ocaml-base-compiler.5.2.0: Requires ocaml = 5.2.0
      ocaml-base-compiler.5.1.1: Requires ocaml = 5.1.1
      ocaml-base-compiler.5.1.0: Requires ocaml = 5.1.0
      ocaml-base-compiler.5.0.0: Requires ocaml = 5.0.0
      ocaml-base-compiler.4.14.2: Requires ocaml = 4.14.2
      ...
- ppxlib -> (problem)
    ppx_deriving 6.0.2 requires >= 0.32.0
    Rejected candidates:
      ppxlib.0.33.0: Requires ocaml >= 4.04.1 & < 5.3.0
[...]

PR #11015, opened 2024-10-16
[refactor] Simplify our 0install vendor to remove unused features

ElectreAAS:
Our vendored dependency, `opam-0install`, is our only interface with `0install` and doesn't use `command`s or 'machine groups' at all. Here I remove them entirely, without any feature loss.

The goal is to simplify the codebase such that future modifications are easier to make. In particular, looking at things like #10987, #10838, maybe even help with #10837.


ElectreAAS:
Note the this PR includes two commits that run ocamlformat on both the 0install-solver and the opam-0install codebases. Working with a formatted file is a lot easier.
We might want to add those to `.git-blame-ignore-revs` afterwards.
I could also remove the commits if we deem them unnecessary.

rgrinberg:
Thanks for getting started on this. To properly fork these projects and develop them as internal dune libraries, I suggest we take the following preliminary steps:

1. Move them from vendor to src
2. Remove the import scripts
3. Run ocamlformat

Once those 3 are done, the removal of dead code should commence as it will be a lot easier to tell what is being deleted. You followed the same steps but in a different order, which makes review a little bit harder. It's not a big deal though, as the changes are still simple. However, it would be helpful to push our changes after preliminary work

ElectreAAS:
Thanks for the pointers @rgrinberg, I'll work on cleaning up my butchery after stabilizing my changes, I have other ideas still

Relatedly, my last commit pretty much solves #10987, although in a very opiniated way.
I'll look into patching that on its own though

ElectreAAS:
My above comment is now outdated, this PR is ready for review. The commit solving the weird errors is in another PR.

PR #11019, opened 2024-10-18
pkg: relax version constraints for ocamlformat dev tool

gridbugs:
Previously dune would install the exact version of ocamlformat specified in the .ocamlformat file in the project's root. This prevents alternative distributions of the ocamlformat package from being used, as these typically append a suffix to the version number.

The motivation for this change is supporting a binary dev tools distribution where the ocamlformat package is named like "ocamlformat.0.26.2+binary", and we want that package to be installable when the .ocamlformat file of a project specifies "version=0.26.2".

### DIFF
--- a/bin/lock_dev_tool.ml
+++ b/bin/lock_dev_tool.ml
@@ -13,6 +13,31 @@ let is_enabled =
      | `Disabled -> false)
 ;;
 
+(* Returns a version constraint accepting (almost) all versions whose prefix is
+   the given version. This allows alternative distributions of packages to be
+   chosen, such as choosing "ocamlformat.0.26.2+binary" when .ocamlformat
+   contains "version=0.26.2". *)
+let relaxed_version_constraint_of_version version =
+  let open Dune_lang in
+  let min_version = Package_version.to_string version in
+  (* The goal here is to add a suffix to [min_version] to construct a version
+     number higher than than any version number likely to appear with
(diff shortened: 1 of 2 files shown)


bin/lock_dev_tool.ml
```diff
@@ -13,6 +13,31 @@ let is_enabled =
...
+  (* The goal here is to add a suffix to [min_version] to construct a version
+     number higher than than any version number likely to appear with
+     [min_version] as a prefix. "_" is the highest ascii symbol that can appear
+     in version numbers, excluding "~" which has a special meaning. It's
+     conceivable that one or two consecutive "_" characters may be used in a
+     version, so this appends "___" to [min_version].
+
+     Read more at: https://opam.ocaml.org/doc/Manual.html#Version-ordering
+  *)
+  let max_version = String.concat ~sep:"" [ min_version; "___" ] in
```
Leonidas-from-XIV, line 32:
```suggestion
  let max_version = min_version ^ "___"  in
```

Leonidas-from-XIV (commented):
Seems sensible to me, I would assume `ocamlformat.0.26.2.1`should qualify if one requests `0.26.2`.

moyodiallo:
Would you like add a test ?

gridbugs:
Added a test

Leonidas-from-XIV (approved):
LGTM! Nice and clear test.

PR #11020, opened 2024-10-18
pkg: add binary dev tools repo

gridbugs:
Adds a repo of binary dev tools to dune so that dev tools can be quickly installed without the need to download and compile their entire dependency cone on supported platforms. Currently ocamlfind and ocamllsp are included in the repo.

### DIFF
--- a/src/dune_pkg/workspace.ml
+++ b/src/dune_pkg/workspace.ml
@@ -52,6 +52,15 @@ module Repository = struct
     }
   ;;
 
+  let binary_packages =
+    { name = "binary-packages"
+    ; source =
+        ( Loc.none
+        , OpamUrl.of_string "git+https://github.com/ocaml-dune/ocaml-binary-packages.git"
+        )
+    }
+  ;;
+
   let decode =
     let open Decoder in
     fields
--- a/src/dune_rules/workspace.ml
+++ b/src/dune_rules/workspace.ml
@@ -682,6 +682,8 @@ let find_lock_dir t path =
   List.find t.lock_dirs ~f:(fun lock_dir -> Path.Source.equal lock_dir.path path)
 ;;
 
+let add_repo t repo = { t with repos = repo :: t.repos }
+
 include Dune_lang.Versioned_file.Make (struct
     type t = unit
   end)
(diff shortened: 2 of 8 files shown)


src/dune_pkg/workspace.ml
```diff
@@ -52,6 +52,15 @@ module Repository = struct
     }
   ;;
 
+  let binary_dev_tools =
```
Leonidas-from-XIV, line 55: Does it only have to be just `dev-tools`? From what I understand, nothing (but relocatability) is preventing us from shipping any compiled packages there, like `Lwt` or any package that we choose to precompile. Having e.g. the top 100 OPAM packages there could be a significant speedup.

src/dune_rules/workspace.ml
```diff
@@ -2,7 +2,9 @@ open Import
 open Dune_lang.Decoder
 module Repository = Dune_pkg.Pkg_workspace.Repository
 
-let default_repositories = [ Repository.overlay; Repository.upstream ]
+let default_repositories =
+  [ Repository.binary_dev_tools; Repository.overlay; Repository.upstream ]
```
Leonidas-from-XIV, line 6: Given this is repo is very new, I think it would be nice to have this behind a feature flag and just enabled in the preview.

Personally I don't really want to pull binaries from the internet, so having an easy way to opt out of it (that is not putting `dune-workspace` files everywhere) seems important to me. Maybe an option in `dune-config` or so?

rgrinberg:
This repository is not (yet?) usable outside of dev tools? If that's the case, we shouldn't make it visible in the workspace at all. Ideally, this repository should only be used when installing dev tools.

src/dune_pkg/workspace.ml
gridbugs: Good point. I'll rename the repo to be more general to give us the option to add more packages to it in the future.

src/dune_rules/workspace.ml
gridbugs: Agreed, but do you mind if this comes in a later change?

gridbugs:
Updated so that the binary repo is only used when solving dev tools.

Leonidas-from-XIV (approved):

PR #11022, opened 2024-10-18
refactor(pkg): simplify depopts handling

rgrinberg:
The simplification comes from us resolving the depopts formula exactly in the same way as we do for regular deps.

In theory, this is also a bug fix because we are now using the same environment in both depopts and regular deps. However, I was unable to reproduce this as we lack full support for depopts.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -539,33 +539,6 @@ let make_action = function
   | actions -> Some (Action.Progn actions)
 ;;
 
-(* Returns the set of depopts of a package which are part of the solution
-   represented by [version_by_package_name] *)
-let available_depopts solver_env version_by_package_name depopts =
-  let formula =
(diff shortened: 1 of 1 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -577,6 +550,28 @@ let opam_file_is_compiler (opam_package : OpamFile.OPAM.t) =
...
 ;;
 
+let resolve_depopts ~resolve depopts =
+  (let rec collect acc depopts =
+     match (depopts : OpamTypes.filtered_formula) with
+     | Or ((Atom (_, _) as dep), depopts) -> collect (dep :: acc) depopts
+     | Atom (_, _) -> depopts :: acc
+     | Empty -> acc
+     | _ ->
+       (* We rely on depopts always being a of or'ed package names. Opam
```
gridbugs, line 560: missing word?

gridbugs (approved):

PR #11028, opened 2024-10-19
fix: allow [$ dune subst] to work without a repo

### LINKED ISSUE #10927: dune subst silently ignores %%ID%% string when outside a git-controlled directory
Note this impact file-pinned opam packages in particular
## Reproduction

```sh
$ mkdir test

$ cat > dune-project << EOF
(lang dune 3.16)
(name test)
EOF

$ echo "%%NAME%%" > README.md
```

## Expected Behavior

```sh
$ dune subst
$ cat README.md
test
```

## Actual Behavior

```sh
$ dune subst
$ cat README.md
%%NAME%%
```

## Specifications

- dune 3.16.0
- ocaml 5.2
- debian 12

rgrinberg:

### DIFF
--- a/bin/subst.ml
+++ b/bin/subst.ml
@@ -299,111 +299,143 @@ let make_watermark_map ~commit ~version ~dune_project ~info =
 
 let subst vcs =
   let open Memo.O in
-  let* (version, commit), files =
-    Memo.fork_and_join
-      (fun () ->
-        Memo.fork_and_join (fun () -> Vcs.describe vcs) (fun () -> Vcs.commit_id vcs))
-      (fun () -> Vcs.files vcs)
-  in
-  let+ (dune_project : Dune_project.t) =
-    (let files =
-       (* Filter-out files form sub-directories *)
-       List.fold_left files ~init:String.Set.empty ~f:(fun acc fn ->
-         let fn = Path.source fn in
-         if Path.is_root (Path.parent_exn fn)
-         then String.Set.add acc (Path.to_string fn)
-         else acc)
-     in
-     Dune_project.load ~dir:Path.Source.root ~files ~infer_from_opam_files:true)
-    >>| function
-    | Some dune_project -> dune_project
-    | None ->
-      User_error.raise
-        ~loc:(Loc.in_dir (Path.source Path.Source.root))
-        [ Pp.text
-            "There is no dune-project file in the current directory, please add one with \
-             a (name <name>) field in it."
-        ]
-        ~hints:
-          [ Pp.concat
-              ~sep:Pp.space
-              [ User_message.command "dune subst"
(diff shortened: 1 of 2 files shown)


rgrinberg:
@anmonteiro do you want to review this one?

bin/subst.ml
```diff
@@ -299,111 +299,142 @@ let make_watermark_map ~commit ~version ~dune_project ~info =
...
+  (match vcs with
+   | Some vcs ->
+     let+ version = Vcs.describe vcs
+     and+ commit_id = Vcs.commit_id vcs
+     and+ files = Vcs.files vcs in
+     Some (version, commit_id, files)
+   | None ->
+     (* We have to do this because scanning the source tree evaluates [-p].
+        That's because [-p] is needed to interpret packages in dune projects
+        correctly. It should not be necessary, so we shuold probably make the
```
anmonteiro, line 311:
```suggestion
        correctly. It should not be necessary, so we should probably make the
```

bin/subst.ml
```diff
@@ -299,111 +299,142 @@ let make_watermark_map ~commit ~version ~dune_project ~info =
...
+        package loading lazier. *)
+     Dune_rules.Only_packages.Clflags.set No_restriction;
+     let* root = Source_tree.root () in
+     let project = Source_tree.Dir.project root in
+     if Dune_project.dune_version project < (3, 17)
+     then Memo.return None
```
anmonteiro, line 317: should we print a warning / hint here or would that be too noisy?

bin/subst.ml
```diff
@@ -299,111 +299,142 @@ let make_watermark_map ~commit ~version ~dune_project ~info =
...
+      then subst_file path ~map:watermarks opam_package_files))
 ;;
 
 let subst () =
-  match
-    Sys.readdir "." |> Array.to_list |> String.Set.of_list |> Vcs.Kind.of_dir_contents
-  with
-  | None -> Fiber.return ()
-  | Some kind -> Memo.run (subst { kind; root = Path.root })
+  Sys.readdir "."
```
anmonteiro, line 431: why do we need `Sys.readdir "."` at all if we get the source root in the function above?

Would it be better to find the root first and then try to find the VCS kind? what happens if `subst` is not run from the VCS root?

anmonteiro (approved):

bin/subst.ml
rgrinberg: Too noisy I'd say. Essentially every single package would emit this warning now.

bin/subst.ml
rgrinberg: Yeah, we can call `Source_tree.nearest_vcs`. Will leave a TODO

btjorge:
The fix introduce a regression when an opam file is present, I just opened #11045 reporting that

PR #11031, opened 2024-10-20
refactor: remove [Ocaml_index.project_index]

rgrinberg:
doesn't seem to be used anywhere


voodoos (approved):
Good catch, it must have been a remnant from a past refactor.

Issue #11001, opened 2024-10-07
pkg: a command line that installs all the dev-tools at once.


moyodiallo:
## Desired Behavior

Right now, a dev-tool is installed when running the command that related to it. So it could be useful to have a command line that installs all the dev-tools for a user at once. 

Some of the use cases could be:
- A user want to have all the tools before being offline.
- Caching on a CI step.

## Example

Implementing `dune tools install` could be solution which will trigger the build of every tool that is available.

shonfeder:
Closing this in favor of https://github.com/ocaml/dune/issues/12135, since that has more context and discussion.

PR #11032, opened 2024-10-20
pkg: Install all the dev-tools that are available

moyodiallo:
This is supposed to close https://github.com/ocaml/dune/issues/11001. 

But instead of having `dune tools install` I wonder if `dune build @pkg-tools` could be a better alternative, because this is equivalent building/invoking some aliases.

### DIFF
--- a/bin/lock_dev_tool.ml
+++ b/bin/lock_dev_tool.ml
@@ -229,9 +229,10 @@ let lock_ocamlformat () =
 let lock_odoc () = lock_dev_tool Odoc None
 let lock_ocamllsp () = lock_dev_tool Ocamllsp None
 
-let lock_tools (tools : Dune_pkg.Dev_tool.t list) =
-  List.map tools ~f:(function
-    | Dune_pkg.Dev_tool.Odoc -> lock_odoc ()
+let lock_tools tools =
+  List.map tools ~f:(fun (tool : Dune_pkg.Dev_tool.t) ->
+    match tool with
+    | Odoc -> lock_odoc ()
     | Ocamllsp -> lock_ocamllsp ()
     | Ocamlformat -> lock_ocamlformat ())
 ;;
--- a/bin/lock_dev_tool.ml
+++ b/bin/lock_dev_tool.ml
@@ -228,3 +228,10 @@ let lock_ocamlformat () =
 
 let lock_odoc () = lock_dev_tool Odoc None
 let lock_ocamllsp () = lock_dev_tool Ocamllsp None
+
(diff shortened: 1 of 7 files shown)


maiste:
> I've added @maiste to weight in on this.

I would advocate for the opposite here. The aliases make sense when they are related to building targets in the context of Dune. Here, we are talking about development tools, so I'm afraid it would break the semantics as they are external tools. Also, in terms of accessibility, it would be easier for users to get access to the command than to the alias as with @gridbugs work, they have access to auto-completion.

My recommendation would be to stick to the command instead of the alias, but in the end, it is just my opinion :smile:

bin/lock_dev_tool.ml
```diff
@@ -228,3 +228,10 @@ let lock_ocamlformat () =
 
 let lock_odoc () = lock_dev_tool Odoc None
 let lock_ocamllsp () = lock_dev_tool Ocamllsp None
+
+let lock_tools (tools : Dune_pkg.Dev_tool.t list) =
+  List.map tools ~f:(function
+    | Dune_pkg.Dev_tool.Odoc -> lock_odoc ()
```
Leonidas-from-XIV, line 234: Can you add a type annotation so this prefix is not needed?

bin/tools/install_tools.ml
```diff
@@ -0,0 +1,41 @@
...
+  let open Fiber.O in
+  let+ result =
+    let dev_tools_path =
+      List.map dev_tools ~f:(fun path -> Pkg_dev_tool.exe_path path |> Path.build)
+    in
+    Build_cmd.run_build_system ~common ~request:(fun _build_system ->
+      List.map dev_tools_path ~f:Action_builder.path |> Action_builder.all_unit)
+  in
+  match result with
+  | Error `Already_reported -> raise Dune_util.Report_error.Already_reported
```
Leonidas-from-XIV, line 14: I wonder, shouldn't this be `Code_error.raise`?

moyodiallo: If there is an error, it is already reported during the build and `Code_error.raise` is about raising with a message.

gridbugs:
On the topic of installing dev tools with a command vs with an alias, I agree with @maiste that it breaks the semantics of aliases and is less accessible in terms of shell completion (at least until we can complete aliases which won't happen any time soon).

Looks like I accidentally duplicated this PR with https://github.com/ocaml/dune/pull/11146, which I closed because the overwhelming consensus among reviewers was that tools should be installed with an alias rather than a command. Thus I'm closing this too.

In the mean time individual tools can be installed by running them with `--help`, such as `dune tools exec ocamllsp -- --help`.

PR #11035, opened 2024-10-21
fix(pkg): do not generate index rules

### LINKED ISSUE #10985: `dune build @ocaml-index` fails when depending on a package with package management enabled

rgrinberg:
Libraries in the lock directory are in the build directory, but we still need to treat them as external

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/gh10985.t
+++ b/test/blackbox-tests/test-cases/pkg/gh10985.t
@@ -49,11 +49,16 @@ Now we set up a lock file with this package and then attempt to use it:
   > let () = Test_lib.x
   > EOF
 
+  $ mkdir .bin
+  $ cat > .bin/ocaml-index <<EOF
+  > #!/usr/bin/env sh
+  > exit 1
+  > EOF
+
+  $ chmod +x .bin/ocaml-index
+  $ export PATH="$PWD/.bin:$PATH"
+
   $ dune build @ocaml-index
-  Error: This rule defines a directory target "default/.pkg/mypkg/target" that
-  matches the requested path
-  "default/.pkg/mypkg/target/lib/mypkg/lib/cctx.ocaml-index" but the rule's
-  action didn't produce it
-  -> required by _build/default/.foo.objs/cctx.ocaml-index
-  -> required by alias ocaml-index
+  File ".foo.objs/_unknown_", line 1, characters 0-0:
+  Command exited with code 1.
   [1]
--- a/src/dune_rules/merlin/ocaml_index.ml
+++ b/src/dune_rules/merlin/ocaml_index.ml
@@ -44,14 +44,15 @@ let cctx_rules cctx =
        - If these libs are rebuilt a re-indexation is needed *)
     let other_indexes_deps =
       let open Resolve.Memo.O in
-      let+ requires_compile = Compilation_context.requires_compile cctx in
-      let deps =
-        List.filter_map requires_compile ~f:(fun l ->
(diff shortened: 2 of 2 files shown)


test/blackbox-tests/test-cases/pkg/gh10985.t
```diff
@@ -50,10 +50,7 @@ Now we set up a lock file with this package and then attempt to use it:
...
   $ dune build @ocaml-index
-  Error: This rule defines a directory target "default/.pkg/mypkg/target" that
-  matches the requested path
-  "default/.pkg/mypkg/target/lib/mypkg/lib/cctx.ocaml-index" but the rule's
-  action didn't produce it
-  -> required by _build/default/.foo.objs/cctx.ocaml-index
-  -> required by alias ocaml-index
+  File ".foo.objs/_unknown_", line 1, characters 0-0:
+  Error: Program ocaml-index not found in the tree or in PATH
+   (context: default)
```
voodoos, line 55: This test needs `(deps %{bin:ocaml_index})`.
One option could be to move it to the sub-directory `ocaml-index` that specify this dependency for this sub-tree.

voodoos (approved):
Looks good to me.

test/blackbox-tests/test-cases/pkg/gh10985.t
rgrinberg: I added a mock binary here. This tests needs to be part of package management as well unfortunately

PR #11039, opened 2024-10-22
Added reproduction test case for missing dep

ElectreAAS:
This is a reproduction issue for #10987.


rgrinberg (changes requested):
Can you make this test work with a mock opam repository?

PR #11040, opened 2024-10-23
Drastically simplify output when packages are missing in solve

### LINKED ISSUE #10987: pkg: inscrutable error when depending on a non-existent package

ElectreAAS:
Fixes #10987.
When some packages are missing entirely, we only report that as that seems way more pressing than the wall of info about the state of the solver seen in the issue.

The fix was found while working on #11015, but is independent from it.

### DIFF
--- a/src/0install-solver/diagnostics.ml
+++ b/src/0install-solver/diagnostics.ml
@@ -40,7 +40,8 @@ module Make (Results : S.SOLVER_RESULT) = struct
           (format_role other_role
            ++ Pp.char ' '
            ++ Model.pp_version impl
-           ++ Pp.textf " requires %s" (format_restrictions r))
+           ++ Pp.text " requires "
+           ++ Pp.paragraph (format_restrictions r))
       | Feed_problem msg -> Pp.text msg
     ;;
   end
--- a/src/opam-0install/lib/solver.ml
+++ b/src/opam-0install/lib/solver.ml
@@ -33,31 +33,41 @@ module Make (Context : S.CONTEXT) = struct
     | None -> Error req
   ;;
 
-  let pp_short = Input.pp_impl
-
-  let rec partition f = function
-    | [] -> [], []
-    | x :: xs ->
-      let ys, zs = partition f xs in
-      (match f x with
-       | `Left y -> y :: ys, zs
-       | `Right z -> ys, z :: zs)
+  let rec partition_three f = function
+    | [] -> [], [], []
+    | first :: rest ->
+      let xs, ys, zs = partition_three f rest in
+      (match f first with
+       | `Left x -> x :: xs, ys, zs
+       | `Middle y -> xs, y :: ys, zs
+       | `Right z -> xs, ys, z :: zs)
   ;;
 
   let pp_rolemap ~verbose reasons =
-    let short, long =
+    let good, bad, unknown =
       reasons
       |> Solver.Output.RoleMap.bindings
-      |> partition (fun (_role, component) ->
+      |> partition_three (fun (role, component) ->
         match Diagnostics.Component.selected_impl component with
         | Some impl when Diagnostics.Component.notes component = [] -> `Left impl
-        | _ -> `Right component)
+        | _ ->
+          (match Diagnostics.Component.rejects component with
+           | _, `No_candidates -> `Right role
+           | _, _ -> `Middle component))
     in
-    let pp_item = Diagnostics.Component.pp ~verbose in
-    Pp.paragraph "Selected: "
-    ++ Pp.hovbox (Pp.concat_map ~sep:Pp.space short ~f:pp_short)
-    ++ Pp.cut
-    ++ Pp.enumerate long ~f:pp_item
+    let pp_bad = Diagnostics.Component.pp ~verbose in
+    let pp_unknown role = Pp.box (Solver.Output.Role.pp role) in
+    match unknown with
+    | [] ->
+      Pp.paragraph "Selected candidates: "
+      ++ Pp.hovbox (Pp.concat_map ~sep:Pp.space good ~f:Input.pp_impl)
+      ++ Pp.cut
+      ++ Pp.enumerate bad ~f:pp_bad
+    | _ ->
+      (* In case of unknown packages, no need to print the full diagnostic list, the problem is simpler. *)
+      Pp.hovbox
+        (Pp.text "The following packages couldn't be found: "
+         ++ Pp.concat_map ~sep:Pp.space unknown ~f:pp_unknown)
   ;;
 
   let diagnostics_rolemap req =
@@ -66,7 +76,7 @@ module Make (Context : S.CONTEXT) = struct
 
   let diagnostics ?(verbose = false) req =
     let+ diag = diagnostics_rolemap req in
-    Pp.paragraph "Can't find all required versions."
+    Pp.paragraph "Couldn't solve the package dependency formula."
     ++ Pp.cut
     ++ Pp.vbox (pp_rolemap ~verbose diag)
   ;;
--- a/test/blackbox-tests/test-cases/pkg/conflict-class.t
+++ b/test/blackbox-tests/test-cases/pkg/conflict-class.t
@@ -32,8 +32,8 @@ Local conflict class defined in a local package:
   $ dune pkg lock
   Error: Unable to solve dependencies for the following lock directories:
   Lock directory dune.lock:
-  Can't find all required versions.
-  Selected: foo.dev x.dev foo&x
+  Couldn't solve the package dependency formula.
+  Selected candidates: foo.dev x.dev foo&x
   - bar -> (problem)
       Rejected candidates:
         bar.0.0.1: In same conflict class (ccc) as foo
@@ -51,8 +51,8 @@ Now the conflict class comes from the opam repository
   $ dune pkg lock
   Error: Unable to solve dependencies for the following lock directories:
   Lock directory dune.lock:
-  Can't find all required versions.
-  Selected: foo.0.0.1 x.dev
+  Couldn't solve the package dependency formula.
+  Selected candidates: foo.0.0.1 x.dev
   - bar -> (problem)
       Rejected candidates:
         bar.0.0.1: In same conflict class (ccc) as foo
--- a/test/blackbox-tests/test-cases/pkg/conflicts.t
+++ b/test/blackbox-tests/test-cases/pkg/conflicts.t
@@ -19,8 +19,8 @@ The solver should say no solution rather than just ignoring the conflict.
   > EOF
   Error: Unable to solve dependencies for the following lock directories:
   Lock directory dune.lock:
-  Can't find all required versions.
-  Selected: bar.0.0.1 x.dev
+  Couldn't solve the package dependency formula.
+  Selected candidates: bar.0.0.1 x.dev
   - foo -> (problem)
       x dev requires conflict with all versions
       Rejected candidates:
--- a/test/blackbox-tests/test-cases/pkg/implicit-dune-constraint.t
+++ b/test/blackbox-tests/test-cases/pkg/implicit-dune-constraint.t
@@ -19,8 +19,8 @@ dependency.
   $ test "2.0.0"
   Error: Unable to solve dependencies for the following lock directories:
   Lock directory dune.lock:
-  Can't find all required versions.
-  Selected: foo.0.0.1 x.dev
+  Couldn't solve the package dependency formula.
+  Selected candidates: foo.0.0.1 x.dev
   - dune -> (problem)
       User requested = 3.17
       Rejected candidates:
--- a/test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
+++ b/test/blackbox-tests/test-cases/pkg/lockfile-generation.t/run.t
@@ -147,8 +147,8 @@ Run the solver again. This time it will fail.
   $ dune pkg lock
   Error: Unable to solve dependencies for the following lock directories:
   Lock directory dune.lock:
-  Can't find all required versions.
-  Selected: baz.0.1.0 foo.0.0.1 lockfile_generation_test.dev
+  Couldn't solve the package dependency formula.
+  Selected candidates: baz.0.1.0 foo.0.0.1 lockfile_generation_test.dev
   - bar -> (problem)
       foo 0.0.1 requires >= 0.2
       lockfile_generation_test dev requires >= 0.6
--- a/test/blackbox-tests/test-cases/pkg/non-existent-dep.t
+++ b/test/blackbox-tests/test-cases/pkg/non-existent-dep.t
@@ -1,7 +1,7 @@
 A package depending on a package that doesn't exist.
-The solver should give a more sane error message.
+The solver now gives a more sane error message.
 
-A few packages here so the errors get large.
+A few packages here so the errors could get large.
   $ . ./helpers.sh
   $ mkrepo
   $ add_mock_repo_if_needed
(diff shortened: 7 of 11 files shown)


v-gb:
Thanks, that'd be much more understandable.
The format of the list of missing packages seems a bit weird though. To me, one of these formats would make sense

```
  The following packages couldn't be found:
  - ocamlformat
  - whatever

  The following packages couldn't be found: ocamlformat, whatever

  The following packages couldn't be found: ocamlformat whatever
```

but it seems to be (from reading the code, I could be mistaken):

```
The following packages couldn't be found: - ocamlformat - whatever
```

ElectreAAS:
You're right. Should be fixed now.

Alizter:
If we are moving 0install to be our own library, we might as well just drop the messages being created in 0install and only have the data. We can then use our own Pp library to render the messages nicely.

rgrinberg:
It might be independent from forking the solver, but I think we should finish forking the solver first. In general, we don't make changes in vendored repos without a lot of extra overhead (update scripts, maintaining commits in separate repos). If you fork the solver, you will avoid doing all that work in this PR.

Leonidas-from-XIV:
Now that #11015 is merged this needs to be rebased and then we can proceed.

ElectreAAS:
Rebased on main!

src/opam-0install/lib/solver.ml
```diff
@@ -38,32 +38,48 @@ struct
...
     let pp_item f c = Fmt.pf f "- @[%a@]" (Diagnostics.Component.pp ~verbose) c in
-    Fmt.pf
-      f
-      "Selected: @[<hov>%a@]@,%a"
-      (Fmt.(list ~sep:sp) pp_short)
-      short
-      (Fmt.(list ~sep:cut) pp_item)
-      long
+    let pp_unknown f (role, _) = Fmt.pf f "@[%a@]" Solver.Output.Role.pp role in
+    let pp_normals f short long =
```
Leonidas-from-XIV, line 66: What are "normals"? I think this would benefit from a better name.

Leonidas-from-XIV (commented):
Thanks for taking care of this. Here some comments on the code.

src/opam-0install/lib/solver.ml
```diff
@@ -38,32 +38,48 @@ struct
...
+        long
+    in
+    if unknowns = []
+    then pp_normals f short long
+    else
+      Fmt.pf
+        f
+        "The following packages couldn't be found: @[<h>%a@]"
+        Fmt.(list ~sep:sp pp_unknown)
+        unknowns
```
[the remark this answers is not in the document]
rgrinberg: To make it work well with the rest of dune, we should get rid of `Fmt` in favor of `Pp`. So maybe it's not worth dwelling on this for now?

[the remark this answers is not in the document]
Leonidas-from-XIV: Yes, I agree, but I didn't want to add unrelated stuff to the PR by asking for `Fmt` to be removed.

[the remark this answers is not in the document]
ElectreAAS: I ended up removing `Fmt` in favor of good old `Format`. I briefly tried converting to `Pp`, but felt this was simpler.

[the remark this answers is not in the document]
rgrinberg: Note that both `Fmt` and `Format` suffer from the same problem: they don't allow the errors to be serialized. That is why we're using Pp.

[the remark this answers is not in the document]
ElectreAAS: Oooooooh _that's_ why `Pp` is used!
It wasn't painless, but I finally moved everything to `Pp`, so we can enjoy better errors :)

ElectreAAS:
Moving to `Pp` introduces a fair amount of noise in the form of `type t -> type 'tag t` to accommodate for [`Pp`'s notion of tags](https://ocaml.org/p/pp/latest/doc/Pp/index.html#tags) that are not used at the moment. We could stylize solver failures further with them. What do you think?

src/0install-solver/diagnostics.ml
```diff
@@ -27,53 +26,50 @@ struct
       | Restricts of Model.Role.t * Model.impl * Model.restriction list
       | Feed_problem of string
 
-    let pp f = function
-      | UserRequested r -> pf f "User requested %s" (format_restrictions [ r ])
+    let pp = function
+      | UserRequested r -> Pp.verbatimf "User requested %s" (format_restrictions [ r ])
```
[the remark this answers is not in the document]
Alizter: Basically, anywhere you have `verbatim` replace it with `text`.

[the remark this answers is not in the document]
ElectreAAS: That's what I did initially, but I ended up with this kind of formatting:
```
Rejected
  candidates:
  bar.0.0.1:
  In
  same
  conflict
  class
  (ccc)
  as
  foo
```

I expect a fair amount of these messages are wrapped in a `vbox` at some point, which means that basically all spaces in `text` are used as line breaks

src/0install-solver/diagnostics.ml
```diff
@@ -27,53 +26,50 @@ struct
...
+      | Feed_problem msg -> Pp.verbatim msg
     ;;
   end
 
   (** Represents a single interface in the example (failed) selections produced by the solver.
       It partitions the implementations into good and bad based (initially) on the split from the
       impl_provider. As we explore the example selections, we further filter the candidates. *)
   module Component = struct
-    type rejection_reason =
+    type 'tag rejection_reason =
```
Alizter, line 49:
```suggestion
    type 'tag rejection_reason =
```
No need to do this, the tag will always be `Stdune.User_message.Style.t`.

Alizter: It's a bit tricky to do, but replace everything with 'tag that you introduced with `Stdune.User_message.Style.t`. We can look at cleaning up that name at another time using our `Import` module, but there will be too many conflicts for now. Sometimes there is a `_` instead of a `'tag` in the interface just so that you are aware. Also any `unit Pp.t` can be taken as `Stdune.User_message.Style.t Pp.t` but that seems to only be for debug stuff.

src/0install-solver/diagnostics.ml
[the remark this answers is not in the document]
Alizter: That indicates that you have too few boxes around and you have some `vbox`es that need some more thought. I will give you some help once I've read some of the messages.

src/0install-solver/diagnostics.ml
ElectreAAS: If `'tag` is always `Stdune.User_message.Style.t`, why is it parameterized? We could have a type alias in stdune if it's always the same

src/0install-solver/diagnostics.ml
```diff
@@ -424,17 +427,16 @@ struct
...
 
   (** Return a message explaining why the solve failed. *)
   let get_failure_reason ?(verbose = false) result =
     let+ reasons = of_result result in
-    Format.asprintf
-      "Can't find all required implementations:@\n@[<v0>%a@]"
-      (pp_rolemap ~verbose)
-      reasons
+    Pp.verbatim "Can't find all required implementations:"
+    ++ Pp.newline
```
Alizter, line 439: This should be a break with a vbox surrounding the Pp so that it is formatted the correct way. You might need a `hovbox` on the sentence above.

src/0install-solver/sat.ml
```diff
@@ -200,17 +199,13 @@ module Make (User : USER) = struct
   ;;
 
   (* For nicer log_debug messages *)
-  let name_lit f (sign, var) =
+  let name_lit (sign, var) =
     match sign with
-    | Pos -> User.pp f var.obj
-    | Neg -> Format.fprintf f "not(%a)" User.pp var.obj
+    | Pos -> User.pp var.obj
+    | Neg -> Pp.text "not(" ++ User.pp var.obj ++ Pp.char ')'
```
Alizter, line 205: textf

src/0install-solver/diagnostics.ml
Alizter: It's not always the same, sometimes it carries style information like color and italics etc. For the output of the solver none of this is relevant and we should try to keep types simple rather than making them needlessly polymorphic.

Alizter:
Try something like this:

```
Pp.vbox (Pp.concat [ Pp.paragraph "Sentece describing something, but not breaking:"; Pp.enumerate ~f candidates])
```
The vbox will mean each entry is formatted vertically, and the Pp.paragraph will take care of wrapping text more generally.

src/0install-solver/sat.ml
ElectreAAS: I don't see how one would use `Pp.textf` here. If we keep `User.pp` as `User.t -> _ Pp.t`, then the `%a` conversion does not work.
If we wish to use it as a string that's possible, but we loop back to removing a lot of Pp.t everywhere

ElectreAAS: Maybe my regexp game isn't strong enough, but I can't find a single use of `Pp.textf "...%a"` in the entire dune codebase or [anywhere else in the OCaml world](https://sherlocode.com/?q=Pp.textf%20%22%5C.%5C*%25a), even though it is technically possible, it just requires a very weirdly typed conversion function (`unit -> <your type, here User.t> -> string`)

Alizter: OK never mind about this then, there are better things to focus on.

src/0install-solver/diagnostics.ml
ElectreAAS: Should be better now.
I couldn't shadow the long `Stdune.User_message.Style.t` everywhere, so it spills out a little in `dune_pkg`, but that seems fine to me

Leonidas-from-XIV (commented):
With all the `Pp` changes I feel like it would be a better solution to split this into a PR that migrates to `Pp` and a PR that implements the output improvement. At the moment the latter is completely dominated by a lot of changes due to migrating to `Pp`.

src/0install-solver/diagnostics.ml
```diff
@@ -282,7 +288,9 @@ struct
   let find_component_ex role report =
     match RoleMap.find_opt role report with
     | Some c -> c
-    | None -> failwith (Format.asprintf "Can't find component %a!" format_role role)
+    | None ->
+      Stdune.User_error.raise
```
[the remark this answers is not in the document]
Alizter: I also thought the same, but there is more work to be done since it uses stdlib `List` functions.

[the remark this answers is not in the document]
ElectreAAS: I already started working on something like this in #11121

ElectreAAS:
> With all the `Pp` changes I feel like it would be a better solution to split this into a PR that migrates to `Pp` and a PR that implements the output improvement. At the moment the latter is completely dominated by a lot of changes due to migrating to `Pp`.

You are right, as always :) See #11126 for that new PR

Leonidas-from-XIV (approved):

PR #11041, opened 2024-10-24
Allow runtest to run individual tests

### LINKED ISSUE #6630: RFC: Allow dune runtest to run individual tests
Currently, running individual cram tests requires remembering that every test has an alias. Moreover, one also has to qualify the alias with a directory if there's more than one test with the same name. It would be much simpler if we could just pass the test's path to `dune runtest` and get it to run.

I think this is also useful for justifying the existence of `dune runtest` itself. Currently, it's just clone of `dune build @runtest`

Alizter:
This PR addresses the suggestion in #6630.

Users can now run individual cram tests using `dune test mytest.t` without having to worry about aliases.

I've also done some refactoring, moving the runtest command into its own file as it would just be cluttering up the build command there.

There is a test making sure the command works in the expected way. The user errors that occur when the command is misused have also been improved.

- Fix #6630

TODO:
- [x] changelog
- [x] manpage docs
- [x] documentation updates (needs a lot of updates)
  - [x] `quick-start.rst`
  - [x] `tests.rst`
  - [x] `usage.rst`
  - [x] `mental-model.rst`
  - [x] `opam-integration.rst`
  - [x] `rule-generation.rst`
  - [x] `cli.rst`
  - [x] `cram.rst`
  - [x] `runtest.rst`

### DIFF
--- a/bin/runtest.ml
+++ b/bin/runtest.ml
@@ -101,7 +101,6 @@ let runtest_term =
           (* If the path is in the source tree, we disambiguate it. *)
           Action_builder.return (`Runtest dir)
       in
-      (* In either case, we just call the alias. *)
       Alias.request
       @@
       match alias_kind with
--- a/bin/runtest.ml
+++ b/bin/runtest.ml
@@ -63,7 +63,7 @@ let disambiguate_test_name path =
     >>= (function
      | Some test ->
        (* If we find the cram test, then we request that is run. *)
-       Memo.return (`Test (Dune_rules.Cram_test.name test))
+       Memo.return (`Test (parent_dir, Dune_rules.Cram_test.name test))
      | None ->
        (* If we don't find it, then we assume the user intended a directory for
           @runtest to be used. *)
@@ -92,10 +92,9 @@ let runtest_term =
   let common, config = Common.init builder in
   let request (setup : Import.Main.build_system) =
     List.map dirs ~f:(fun dir ->
+      let dir = Path.(relative root) (Common.prefix_target common dir) in
       let open Action_builder.O in
       let* alias_kind =
-        (* We always interpret the arguments as paths *)
-        let dir = Path.of_string dir in
         match Path.as_in_source_tree dir with
         | Some path -> Action_builder.of_memo (disambiguate_test_name path)
         | None ->
@@ -106,12 +105,12 @@ let runtest_term =
       Alias.request
       @@
       match alias_kind with
-      | `Test alias_name ->
-        Alias.of_string
-          (Common.root common)
+      | `Test (dir, alias_name) ->
+        Alias.in_dir
+          ~name:(Dune_engine.Alias.Name.of_string alias_name)
           ~recursive:false
           ~contexts:setup.contexts
-          alias_name
+          (Path.source dir)
       | `Runtest dir ->
         Alias.in_dir
           ~name:Dune_rules.Alias.runtest
--- a/bin/runtest.ml
+++ b/bin/runtest.ml
@@ -4,11 +4,24 @@ let runtest_info =
   let doc = "Run tests." in
   let man =
     [ `S "DESCRIPTION"
-    ; `P {|This is a short-hand for calling:|}
-    ; `Pre {|  dune build @runtest|}
+    ; `P "Run the given tests. The [TEST] argument can be either:"
+    ; `I
+        ( "-"
+        , "A directory: If a directory is provided, dune will recursively run all tests \
+           within that directory." )
+    ; `I
+        ( "-"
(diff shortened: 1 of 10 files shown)


bin/runtest.ml
```diff
@@ -0,0 +1,76 @@
...
+  let common, config = Common.init builder in
+  let request (setup : Import.Main.build_system) =
+    Action_builder.all_unit
+      (List.map dirs ~f:(fun dir ->
+         let dir = Path.(relative root) (Common.prefix_target common dir) in
+         Alias.request
+         @@
+         let base_dir, ext = Path.split_extension dir in
+         let alias_kind =
+           match ext, Path.is_directory dir with
```
rgrinberg, line 53: Instead of hard coding the extension, how about you use the source tree in conjunction with `Cram_rules.cram_tests`? That should be a bit more reliable.

test/blackbox-tests/test-cases/runtest-cmd.t
```diff
@@ -0,0 +1,59 @@
...
+@runtest`.
+
+  $ dune test 2>&1 | grep "^File"
+  File "mytest.t", line 1, characters 0-0:
+  File "tests/filetest.t", line 1, characters 0-0:
+  File "tests/myothertest.t/run.t", line 1, characters 0-0:
+
+Passing the name of a test should only run that test.
+
+  $ dune test mytest 2>&1 | grep "^File"
```
rgrinberg, line 31: I'd say we should just accept the full filename. That will work better with tab completion.

test/blackbox-tests/test-cases/runtest-cmd.t
```diff
@@ -0,0 +1,59 @@
...
+  File "mytest.t", line 1, characters 0-0:
+  $ dune test mytest.t 2>&1 | grep "^File"
+  File "mytest.t", line 1, characters 0-0:
+  $ dune test tests/myothertest 2>&1 | grep "^File"
+  File "tests/myothertest.t/run.t", line 1, characters 0-0:
+  $ dune test tests/myothertest.t 2>&1 | grep "^File"
+  File "tests/myothertest.t/run.t", line 1, characters 0-0:
+
+We don't have to fully qualify tests in subdirectories.
```
rgrinberg, line 40: This should be forbidden. It's too confusing. We should just accept a valid source file or directory.

bin/runtest.ml
Alizter: I've done what you've suggested instead.

test/blackbox-tests/test-cases/runtest-cmd.t
Alizter: The new commit makes only the valid path names work.

test/blackbox-tests/test-cases/runtest-cmd.t
Alizter: This is now forbidden as the valid source file is checked for.

Alizter:
I've also improved the error messages for common misuses of commands so that users have some help on what to do in those cases.

bin/runtest.ml
```diff
@@ -0,0 +1,141 @@
...
+            [ Pp.textf "Path %S is not in the source tree." (Path.to_string dir) ]
+      in
+      (* In either case, we just call the alias. *)
+      Alias.request
+      @@
+      match alias_kind with
+      | `Test alias_name ->
+        Alias.of_string
+          (Common.root common)
+          ~recursive:true
```
rgrinberg, line 127: Shouldn't this be non-recursive?

bin/runtest.ml
```diff
@@ -0,0 +1,141 @@
...
+    List.map dirs ~f:(fun dir ->
+      let open Action_builder.O in
+      let* alias_kind =
+        (* We always interpret the arguments as paths *)
+        let dir = Path.of_string dir in
+        match Path.as_in_source_tree dir with
+        (* If the path is in the source tree, we disambiguate it. *)
+        | Some path -> Action_builder.of_memo (disambiguate_test_name path)
+        (* Otherwise, we raise an error for the user. *)
+        | None ->
```
rgrinberg, line 104: Does this just forbid _build paths from being passed in runtest? Why should we do that?

Alizter:
@rgrinberg I've updated the tests after your changes and now updated the documentation.

cc @emillon can you take a look at the doc changes?

toots:
Is this limited to cram tests only?

Alizter:
@toots For now yes. I was looking at adding support for inline tests, but I didn't get round to finishing it.

toots:
Ok. Thanks for this work!

Issue #10949, opened 2024-09-23
[pkg] fetch dependencies without building the project


maiste:
## Context

In the current behaviour, unless you have the exact target, it is impossible to download the dependencies without building the project. A nice feature would be a command that fetch the project dependencies in the same way you would with `dune build`. This is important in the context of Dockerfile, where you can save some computation time by caching this stage. 

## Solution

We can implement a `dune pkg fetch` command that triggers the fetch rule for the entire project.

rgrinberg:
What about just introducing an alias `@pkg-fetch`? This is the more standard mechanism for collecting targets in dune.

maiste:
Yes, definitely! It makes more sense. Another proposition was `dune build @deps`.

The original question was if there was a way to mimic the behaviour of `opam install --deps-only --with-test .` in the context of Dune Package Management. It means that it should fetch the target and build them too. However, I don't know if we have the possibility to execute the test part?

leostera:
To clarify, we have 2 use-cases here.

1. Writing a Dockerfile where we can stage building dependencies separate from building the entire project, and cache that based off the dune-project (or eventually the dune.lock):

```dockerfile
FROM ubuntu

RUN curl https://get.ocaml.org | bash 

# Adds project + deps
ADD dune-project .
RUN dune pkg lock

# Fetch and build dependencies only if the above changes
RUN dune build @deps 

# Adds rest of the sources
ADD . .
RUN dune build
```

2. A single command to fetch all external / downloadable resources for a project, to let you work without connectivity:

```sh
$ git clone github.com/my/project project
$ cd project
$ dune fetch
$ # now i can go offline
$ dune build 
```

rgrinberg:
> The original question was if there was a way to mimic the behaviour of `opam install --deps-only --with-test .` in the context of Dune Package Management. It means that it should fetch the target and build them too. However, I don't know if we have the possibility to execute the test part?

We already build the test deps. So `dune runtest` should work.

maiste:
I have work on part of this in #11046 where it introduces an alias to install and build packages. However, as stated here, we would like to have another alias to just fetch the dependencies but without building them. It is currently unclear to me what would be the proper way to address this second part. Indeed, the call to [Fetch.fetch](https://github.com/ocaml/dune/blob/b61e9d612642dbd1fd58d25c90126d607e69bd43/src/dune_pkg/source.ml#L39) in source.ml seems to also build the dependency. 

[...]

PR #11046, opened 2024-10-28
feat: introduce the @pkg-install alias

maiste:
This PR solves one part of #10949. It introduces a global alias with the source deps as a dependency. When you use it, you will automatically build the source deps of the project. Regarding the PR, I have various questions:
- Should I update the documentation as it is related to package management? (I know we don't want to "pollute" the documentation with not released part)
- Is there a better way to declare an alias, or is this way fine? I tried to take inspiration from other aliases.
- For the test, would there be a cleaner way to check we only installed what we want at each stage. Here, it seems we can't use the printing solution, as we need to install a complete package.


test/blackbox-tests/test-cases/pkg/build-pkg-deps.t
```diff
@@ -0,0 +1,74 @@
...
+  $ cat > bar.ml << EOF
+  > let () = print_endline "source"
+  > EOF
+
+The alias call builds the `foo` dependency but not the project itself. We
+verify we have the `foo` package but not the `bar.exe` in the build directory:
+  $ dune build @pkg-deps
+  $ ls _build/_private/default/.pkg/
+  foo
+  $ ls _build/default/
```
Leonidas-from-XIV, line 69: I was thinking about the way to detect that the build is happening is to maybe enable the console output and have the `opam` file `echo "foo is building"`.

test/blackbox-tests/test-cases/pkg/build-pkg-deps.t
```diff
@@ -0,0 +1,74 @@
...
+
+Verify that building the `bar.exe` file is working:
+  $ dune build ./bar.exe
+  $ ls _build/default/bar.exe
+  _build/default/bar.exe
```
Leonidas-from-XIV, line 74: And this could then just be replaced by `dune exec ./bar.exe`?

This has the slight downside that `bar.exe` could've been built before so maybe building `bar.exe` should also echo something to stdout?

Leonidas-from-XIV (commented):
I am unsure what our policy is about looking into `_build` in tests, but I have some ideas how to avoid doing it.

test/blackbox-tests/test-cases/pkg/build-pkg-deps.t
maiste: It would mean to add a rule that depends on the `bar.exe` no?

test/blackbox-tests/test-cases/pkg/build-pkg-deps.t
```diff
@@ -0,0 +1,74 @@
...
+Create a project using the fake library as a dependency:
+  $ cat > dune-project << EOF
+  > (lang dune 3.16)
+  > (package
+  >  (name bar)
+  >  (allow_empty)
+  >  (depends foo))
+  > EOF
+  $ add_mock_repo_if_needed
+  $ dune pkg lock
```
rgrinberg, line 49: Why do you need to involve the solver in this test?

test/blackbox-tests/test-cases/pkg/build-pkg-deps.t
```diff
@@ -0,0 +1,74 @@
+This test verifies the @pkg-deps alias fetch and build the project dependencies
+without building the project itself.
+
+  $ . ./helpers.sh
+
+Create a fake library to install in the target project as a dependency:
```
rgrinberg, line 6: I didn't really get why you need a fake library and executable in this test. I see the following as sufficient:

1. Write a lock directory
2. Build your new alias
3. Verify that the sources exist

Everything else seems redundant.

test/blackbox-tests/test-cases/pkg/build-pkg-deps.t
rgrinberg: Why should there be a build happening? This alias is supposed to only download the deps.

rgrinberg:
To be consistent with dune, aliases should be attached to source directories. In particular:

The following should build all the sources in dune.lock

```
dune build @dune.lock/pkg-deps
```

The following should build all the sources in the workspace:

```
dune build @pkg-deps
```

If you'd like to download the sources for a particular package, maybe you could add an alias like `dune build @dune.lock/$pkg-deps`

src/dune_rules/alias0.ml
```diff
@@ -25,4 +25,5 @@ let check = standard "check"
 let install = standard "install"
 let ocaml_index = standard "ocaml-index"
 let runtest = standard "runtest"
+let pkg_deps = standard "pkg-deps"
```
rgrinberg, line 28: The point of this aliases is to download the sources? If it is, a more suggestive name would be `pkg-sources` or even just `sources`

test/blackbox-tests/test-cases/pkg/build-pkg-deps.t
maiste: It was the easiest way to write the test. However, it should be possible to write it without using the `lock` command. I'll rework it.

test/blackbox-tests/test-cases/pkg/build-pkg-deps.t
maiste: As we want to build the library and ensure it is installed, you need the fake library to make sure dune build correctly. Otherwise, it will fail when you try to use it as a dependency.

test/blackbox-tests/test-cases/pkg/build-pkg-deps.t
Leonidas-from-XIV: `dune exec ./bar.exe` automatically adds a dependency. Unless I misunderstand your question.

maiste:
@rgrinberg, I feel there is a misunderstanding with the purpose of this PR... The goal is to introduce an alias that allows to download and build the sources of a project. In the issue related to it, we discuss two aliases with two different purposes:
- `@pkg-fetch` whose role is to download the sources without building them. The use case targetted here is when you know you are going to be offline.
- `@pkg-deps` whose role is to download the source and build them. Doing this allow to add a caching layer with Docker when building image. Docker can cache the build dependencies. Hence, when you update your own code, the engine won't try to rebuild the source. This is what this PR is trying to do.

> The following should build all the sources in dune.lock
> dune build @dune.lock/pkg-deps

I see. Where would be the best part to add the alias? `gen_rules.ml`? Also, could I take inspiration from `runtest`? I feel this is pretty related to this mechanism.

> If you'd like to download the sources for a particular package, maybe you could add an alias like dune build @dune.lock/$pkg-deps

Isn't it something that should be attached to `@pkg-fetch` as it seems more relevant?

rgrinberg:
Okay, I think got confused by your implementation. It's currently adding the source deps to the alias. If you'd like the package build to included in the alias, you need to add the target directory of the package as a dependency.

rgrinberg:
> I see. Where would be the best part to add the alias? gen_rules.ml? Also, could I take inspiration from runtest? I feel this is pretty related to this mechanism.

pkg_rules would be the ideal place

> Isn't it something that should be attached to @pkg-fetch as it seems more relevant?

Don't think there's any difference between fetching and building here. But if you decide that you don't need per package granularity, that's OK too.

Also, if you want to check that a package has been built, inspecting _build or using of the shell commands in helpers.sh seems like the better bet. Including extra libraries or executables is still unnecessary moving parts. Our tests suffer from being slow, redundant, and rather hard to understand when failing. Less is more here.

maiste:
@rgrinberg, I tried to follow the advises given during the Dune Dev Meeting.  However, there are still some parts I don't understand or misunderstood. I push my broken code to have a support to discuss on (I didn't rewrite the test yet).

 As you can see in [pkg_rules](https://github.com/ocaml/dune/pull/11046/files#diff-ed8949e9f1cea7cfb9eee6f20d456401dc3755949ed4d24268d1c861f24d69b7R1940), I added a new rule that builds the target paths from the package universe and adds them as a dependency to the alias `@pkg-install`. Then, Dune register this rule as a dependency in `gen_rules.ml` as it was done for the other aliases. However, I encountered two errors:
- I remember you told me to attach this rule to the source directory, not the `_build` directory. Unfortunately, I'm not able to find a way to do it in the current situation. If I try to attach the alias to `project.root` (not in this version of the code but on my machine), I get an error that I must not do this. What is wrong with the way I have done it here? In this version, it keeps telling me it is not attached:
  ```sh
  Error: Alias "pkg-install" specified on the command line is empty.
  It is not defined in . or any of its descendants.
  ```
  How can I attach it to the workspace?
- My first version was trying to register this rule for all the packages, including the one not using package management. This is fixed by the last `tmp` commit I pushed.

Otherwise, was it similar to the approach you had in mind?

src/dune_rules/pkg_rules.ml
```diff
@@ -1937,6 +1937,25 @@ let build_rule context_name ~source_deps (pkg : Pkg.t) =
        ~directory_targets:[ pkg.write_paths.target_dir ]
 ;;
 
+let gen_rule_alias_from_package_universe ~dir ctx_name =
+  let target_path_of_pkg pkg =
+    let pkg_name = Dune_lang.Package_name.to_string pkg in
+    Path.Build.L.relative
+      Private_context.t.build_dir
+      [ Context_name.to_string ctx_name; ".pkg"; pkg_name; "target" ]
```
rgrinberg, line 1945: You should go through the paths module to get this path

rgrinberg (changes requested):

src/dune_rules/alias0.ml
```diff
@@ -25,4 +25,5 @@ let check = standard "check"
 let install = standard "install"
 let ocaml_index = standard "ocaml-index"
 let runtest = standard "runtest"
+let pkg_install = standard "pkg-install"
```
rgrinberg, line 28: It's incorrect to mark this alias as "standard" as it's not defined in every single directory.

src/dune_rules/gen_rules.ml
```diff
@@ -355,6 +368,7 @@ let gen_project_rules =
     and+ () = Odoc.gen_project_rules sctx project
     and+ () = Odoc_new.gen_project_rules sctx project
     and+ () = Ocaml_index.project_rule sctx project
+    and+ () = gen_project_rule_for_package_alias sctx project
```
rgrinberg, line 371: This needs to be done only once per context rather than for every project

src/dune_rules/alias0.ml
```diff
@@ -23,6 +23,7 @@ let doc_json = standard "doc-json"
 let doc_new = standard "doc-new"
 let check = standard "check"
 let install = standard "install"
+let pkg_install = Alias.Name.of_string "pkg-install"
```
Leonidas-from-XIV, line 26: Something of a bikeshedding comment but: Are we going with `@pkg-install`? I think `@install-pkg` fits better with `@install`.

src/dune_rules/gen_rules.ml
```diff
@@ -691,6 +691,27 @@ let raise_on_lock_dir_out_of_sync =
...
+        let* is_lock_dir_active = Lock_dir.lock_dir_active ctx_name in
+        if is_lock_dir_active
+        then (
+          let build_pkg_deps = Alias.make ~dir Alias0.pkg_install in
+          let* action = Pkg_rules.install_packages_from_universe ctx_name in
+          Super_context.add_alias_action sctx ~loc:Loc.none ~dir build_pkg_deps action)
+        else Memo.return ())
+    else Memo.return Rules.empty
+  in
+  Gen_rules.(rules_for ~dir ~allowed_subdirs:String.Set.empty rule |> rules_here)
```
Leonidas-from-XIV, line 712: I would avoid the local open, because it leads to confusing shadowing.

src/dune_rules/pkg_rules.ml
```diff
@@ -1937,6 +1938,26 @@ let build_rule context_name ~source_deps (pkg : Pkg.t) =
        ~directory_targets:[ pkg.write_paths.target_dir ]
 ;;
 
+let install_packages_from_universe ctx_name =
+  let+ packages =
+    (* Fetching the package target implies that we wuill also fetch the extra sources. *)
```
Leonidas-from-XIV, line 1943: Simple typo:

```suggestion
    (* Fetching the package target implies that we will also fetch the extra sources. *)
```

src/dune_rules/pkg_rules.ml
```diff
@@ -1937,6 +1938,26 @@ let build_rule context_name ~source_deps (pkg : Pkg.t) =
...
+    List.map
+      ~f:(fun pkg ->
+        Paths.make ~relative:Path.Build.relative (Project_dependencies ctx_name) pkg
+        |> Paths.target_dir
+        |> Path.build)
+      packages
+    |> Action_builder.paths
+  in
+  Action_builder.bind deps ~f:(fun () ->
+    Action_builder.return (Action.Full.make (Action.Progn [])))
```
Leonidas-from-XIV, line 1958: I think this could be a bit clearer using the operators of `Action_builder.O`.

src/dune_rules/pkg_rules.ml
```diff
@@ -1937,6 +1938,26 @@ let build_rule context_name ~source_deps (pkg : Pkg.t) =
...
+    >>| (fun lock_dir -> lock_dir.packages)
+    >>| Dune_lang.Package_name.Map.keys
+  in
+  let deps =
+    List.map
+      ~f:(fun pkg ->
+        Paths.make ~relative:Path.Build.relative (Project_dependencies ctx_name) pkg
+        |> Paths.target_dir
+        |> Path.build)
+      packages
```
Leonidas-from-XIV, line 1954: I would put this packages at the top and do `packages |> List.map |> Action_builder.paths` for a slightly more linear flow.

src/dune_rules/pkg_rules.ml
```diff
@@ -1937,6 +1938,26 @@ let build_rule context_name ~source_deps (pkg : Pkg.t) =
        ~directory_targets:[ pkg.write_paths.target_dir ]
 ;;
 
+let install_packages_from_universe ctx_name =
```
Leonidas-from-XIV, line 1941: Shouldn't it be `build_packages_of_context` or something?

test/blackbox-tests/test-cases/pkg/alias-pkg-install.t
```diff
@@ -0,0 +1,48 @@
+This test verifies the @pkg-deps alias fetch and build the project dependencies
```
Leonidas-from-XIV, line 1: This line needs to be updated with whatever we end up naming the alias.

test/blackbox-tests/test-cases/pkg/alias-pkg-install.t
```diff
@@ -0,0 +1,48 @@
...
+  $ cat > dune-project << EOF
+  > (lang dune 3.16)
+  > (package
+  >  (name bar)
+  >  (allow_empty)
+  >  (depends foo))
+  > EOF
+
+The alias call builds the `foo` dependency but not the project itself. We
+verify we have the `foo` package but not the `bar.exe` in the build directory:
```
Leonidas-from-XIV, line 44: This is confusing because there is no `bar.exe`; the project is empty.

How about creating a rule that will print something (maybe have to enable the debug print output flag) when it is run and show that is is not printed, but then build an executable which will then cause the rule to print? I think this is a bit clearer than exploring the contents of `_build`, as it uses the "public" API of Dune.

Leonidas-from-XIV (commented):
Given I was working on this too, I'll just comment on the non-core parts that stand out to me to make this a bit better/cleaner/more understandable.

test/blackbox-tests/test-cases/pkg/alias-pkg-install.t
maiste: This is a remaining of the past before I simplified the test. 

I don't know if it doesn't introduce more complexity than what we want?
Side note, but the debug package flag is enabled by default in `pkg/` tests :smile:

src/dune_rules/alias0.ml
maiste: I don't have any opinion on this. My main point is that I don't want the user to confuse the behaviour of `@install` with `@install-pkg`.

Leonidas-from-XIV: I think I remember @rgrinberg proposing having "install" in the name because it is conceptually similar to the `@install` target but I am hazy on the exact wording.

src/dune_rules/pkg_rules.ml
Leonidas-from-XIV: You can use parenthesis for scoping operators like so (or similar, I'm just typing into a text field):

```ocaml
(let open Action_builder.O in
deps >>> Action.Full.make (Action.Progn []))
```

This is one way where the scoping rules of OCaml work really well (looking at you, Python).

test/blackbox-tests/test-cases/pkg/alias-pkg-install.t
Leonidas-from-XIV: Up to you, but I personally find it easier to read rules (since they are documented and and everyday tool in Dune) than to understand the meaning of what the existence or non-existence of files in `_build` means.

src/dune_rules/pkg_rules.ml
```diff
@@ -1937,6 +1938,23 @@ let build_rule context_name ~source_deps (pkg : Pkg.t) =
...
+    >>| Dune_lang.Package_name.Map.keys
+  in
+  let open Action_builder.O in
+  packages
+  |> List.map ~f:(fun pkg ->
+    Paths.make ~relative:Path.Build.relative (Project_dependencies ctx_name) pkg
+    |> Paths.target_dir
+    |> Path.build)
+  |> Action_builder.paths
+  >>> Action_builder.return @@ Action.Full.make (Action.Progn [])
```
maiste, line 1955: @rgrinberg To be able to build a `Action.Full.t` we used an empty `Progn`: what would be a better way to get a `Action.Full.t Action_builder.t` from our `unit Action_builder.t`?

maiste (commented):
I have rewritten the code with the precious help of @Leonidas-from-XIV. I also included the new suggestions and simplified the test a lot. It is much faster since we don't need the extra library and the locking :+1:

test/blackbox-tests/test-cases/pkg/alias-pkg-install.t
```diff
@@ -0,0 +1,49 @@
...
+  $ cat > dune << EOF
+  > (executable
+  >  (name bar))
+  > (rule
+  >  (target bar.ml)
+  >  (action
+  >   (progn
+  >    (with-stdout-to %{target} (echo "let _ = 42"))
+  >    (system "cat %{target}"))))
+  > EOF
```
Leonidas-from-XIV, line 37: You could simplify this to a `bar.ml` that just contains `let () = print_endline "Running bar"` and then instead of `dune build` use `dune exec`.

(I'm assuming that `dune exec` would show all the `echo` that `dune build` would when building the dependencies; otherwise `dune build ./bar.exe` followed by `dune exec ./bar.exe` avoids all the complex processing for `bar.ml`)

Leonidas-from-XIV (commented):
Bar the final name of the alias the code looks good to me! I've added a comment about making the test simpler.

Could you add a short documentation about it, similar to `doc/reference/aliases/ocaml-index.rst`? That would help to discover the feature.

test/blackbox-tests/test-cases/pkg/alias-pkg-install.t
maiste: This issue with this is that we are more interested in the cat than the content of `bar.ml`.  With such a configuration, it will only show that `dune exec` works as expected, not that `bar.exe` is not built with `@pkg-install`. Unless I misunderstood your suggestion :sweat_smile:

src/dune_rules/pkg_rules.ml
rgrinberg: There is. Pushed a fix.

rgrinberg (approved):
LGTM. I did some minor cleanups like getting rid of the empty actions. That allowed me to just shove everything back into pkg_rules.

src/dune_rules/alias0.ml
maiste: Let's keep the `pkg` prefix then.

src/dune_rules/pkg_rules.ml
```diff
@@ -1946,6 +1947,42 @@ let gen_rules context_name (pkg : Pkg.t) =
...
+    Dune_lang.Package_name.Map.keys lock_dir.packages
+    |> List.map ~f:(fun pkg ->
+      Paths.make ~relative:Path.Build.relative project_deps pkg
+      |> Paths.target_dir
+      |> Path.build)
+    |> Action_builder.paths
+  in
+  fun ~dir ctx_name ->
+    let rule =
+      (* We only to build when the build_dir is the root of the context *)
```
maiste, line 1966:
```suggestion
      (* We only need to build when the build_dir is the root of the context *)
```

maiste:
Thanks, @rgrinberg for the cleanup and the improvements! :pray:

Leonidas-from-XIV (approved):

Issue #11038, opened 2024-10-22
`dune fmt` requires `ocamlc` to be in path but does not guarantee that this is the case


gridbugs:
Running `dune fmt` with the dev tools feature enabled (e.g. in the developer preview) dune will install ocamlformat before using it to format the project. However, dune also runs `ocamlc` to get information about the current system. If `ocamlc` isn't already in PATH when running `dune fmt`, and no lockdir is present, the command will fail with:
```
Error: Program ocamlc not found in the tree or in PATH
 (context: default)
-> required by loading the OCaml compiler for context "default"
```

Instead, this command should make sure that an ocaml compiler is installed via dune toolchains and add it to the environment in the appropriate place.

Note that this only happens when no lockdir is present because the presence of a lockdir changes how dune finds the ocaml compiler. As a workaround we can require that a lockdir exists when running `dune fmt`.

gridbugs:
It appears that dune always picks an ocaml toolchain when it starts, and part of that involves running `ocamlc -config`. If a lockdir is present then dune will use the ocaml toolchain implied by the lockdir and if there is no lockdir it will look for `ocamlc` in PATH. It's not obvious what dune should do when there is no lockdir present and you're running `dune fmt` with `LOCK_DEV_TOOLS` enabled.

One option would be use the ocaml toolchain implied by the ocamlformat lockdir. The problem with this is we'd like it to work when running `dune fmt` and `dune build @fmt`, and in the latter case that would mean choosing an ocaml toolchain based on the build target which feels messy.

Another option is to make dune invoke `ocamlc -config` just-in-time the first time that information is needed. This would involve threading `Memo` or `Fiber` throughout all the code that currently accesses the fields of `Ocaml_toolchain.t` which would be a lot of work.

The third option is to make it so that running `dune fmt` with `LOCK_DEV_TOOLS` requires a lockdir to be present, and to use the ocaml toolchain implied by the lockdir. This is the easiest to implement but would of course mean that running `dune fmt` without locking would print an error instructing the user to lock their project. We already do this with ocamllsp and odoc.

gridbugs:
I'll start working on deferring running `ocamlc -config` until it's needed to get a feel for how much work it will be.

rgrinberg:
I think I deferred running `ocamlc -config` as much as possible already. Maybe the formatting rules are forcing it unnecessarily?

gridbugs:
[...]

PR #11049, opened 2024-10-29
pkg: ocamlformat dev tool requires lockdir

gridbugs:
This is a workaround for an issue where dune needs to locate an ocaml toolchain in order to run `dune fmt`, and in the absence of a lockdir dune will look in PATH for ocaml tools, which may not be present. The consequence of this issue is that users of the developer preview can have an ocaml toolchain installed by dune (in their ~/.cache/dune/toolchains directory), and running `dune fmt` will complain about `ocamlc` not being found. Forcing the lockdir to exist means that dune will always take the toolchain implied by the lockdir, and there won't be a situation where the command fails due to `ocamlc` being missing.

More info in this issue:
https://github.com/ocaml/dune/issues/11038

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/ocamlformat/ocamlformat-dev-tool-fails-to-build.t
+++ b/test/blackbox-tests/test-cases/pkg/ocamlformat/ocamlformat-dev-tool-fails-to-build.t
@@ -11,9 +11,18 @@ Make a fake ocamlformat with a missing ocamlformat.ml file:
 Make dune-project that uses the mocked dev-tool opam-reposiotry.
   $ make_project_with_dev_tool_lockdir
 
+The project depends on ocaml, so provide a fake ocaml package:
+  $ make_ocaml_opam_pkg
+
+The ocamlformat dev tool requires the project to be locked:
+  $ dune pkg lock
+  Solution for dune.lock:
+  - ocaml.0.0.1
+
 It fails during the build because of missing OCamlFormat module.
   $ DUNE_CONFIG__LOCK_DEV_TOOL=enabled dune fmt
   Solution for dev-tools.locks/ocamlformat:
+  - ocaml.0.0.1
   - ocamlformat.0.26.4
   File "dev-tools.locks/ocamlformat/ocamlformat.pkg", line 4, characters 6-10:
   4 |  (run dune build -p %{pkg-self:name} @install))
--- a/src/dune_pkg/dev_tool.ml
+++ b/src/dune_pkg/dev_tool.ml
@@ -43,7 +43,15 @@ let exe_path_components_within_package t =
 ;;
 
 let needs_to_build_with_same_compiler_as_project = function
-  | Ocamlformat -> false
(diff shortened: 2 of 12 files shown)


test/blackbox-tests/test-cases/pkg/ocamlformat/ocamlformat-dev-tool-fails-to-build.t
```diff
@@ -11,9 +11,18 @@ Make a fake ocamlformat with a missing ocamlformat.ml file:
 Make dune-project that uses the mocked dev-tool opam-reposiotry.
   $ make_project_with_dev_tool_lockdir
 
+The project depends on ocaml, so provide a fake ocaml package:
+  $ make_ocaml_opam_pkg
+
+The ocamlformat dev tool requires the project ot be locked:
```
moyodiallo, line 17:
```suggestion
The ocamlformat dev tool requires the project to be locked:
```

moyodiallo:
I was little bit surprise that ocamlformat doesn't depends on toolchains, in my mind all the dev-tools depends on toolchains.

moyodiallo (approved):
Just some typos. Would it be better to have a test case where ocaml doesn't exist inside the lock-dir ?

gridbugs:
The issue still happens however https://github.com/ocaml/dune/pull/11105 might make it easier to defer/avoid running ocamlc.

rgrinberg:
I pushed a few more commits delaying loading even further. Making the rules lazier around this stuff is a good thing overall and benefits more than just users of package management. Do let me know if there are still issues around this.

If my work wasn't enough, I agree that this workaround is the way to go for now.

gridbugs:
Looks like it still happens. I just tested this again by building the `#dune-static-experimant` flake from main and running this in on a fresh checkout of a project in a docker container without ocaml installed:
```
$ dune fmt
Solution for dev-tools.locks/ocamlformat:
- ocamlformat.0.26.2+binary-ocaml-5.2.1-built-2024-11-25.0-x86_64-unknown-linux-musl
Error: Program ocamlc not found in the tree or in PATH
 (context: default)
-> required by loading the OCaml compiler for context "default"
```

gridbugs:
It occurs to me that this change would mean that the solver would try to choose a version of ocamlformat that can be compiled with the same version of the compiler as the project. This is a problem because the version of ocamlformat also needs to match the version in the .ocamlformat file. Some versions of ocamlformat have constraints on which compilers they build with, so adding this extra constraint will prevent the solver from finding a solution for ocamlformat in some cases.

Instead, I propose that we wait for this problem to show up in the wild and design a proper solution to it (ideally by not running `ocamlc -config` in `dune fmt`). I suspect this problem will be rare in practice as it requires you to not have `ocamlc` in your PATH when running `dune fmt`. The workaround in the meantime is to globally install an ocaml compiler with opam or your system package manager. This ocaml installation won't be used to compile your code when building with dune - just to satisfy dune's need to run `ocamlc -config` when running `dune fmt`.

PR #11053, opened 2024-10-29
Promote the toolchain feature flag to official feature

ElectreAAS:
This PR sports ~~2 commits, the first one being a less-committed approach to the situation, the second a more complete one.~~ just the one commit that makes the flag enabled by default.

There seems to be sentiment around the dune team (cc @Leonidas-from-XIV @rgrinberg @gridbugs in particular) that:
  1. The 'toolchains' feature is stable and tested enough to "ascend" to an official feature of Dune.
  2. The problem that it solves (the compiler being non-relocatable) is going to stay relevant for a while.

Thus, I propose that we ~~get rid of~~ promote the feature flag.
~~Note that this is a **BREAKING** change for users who use `./configure --toolchains disabled` obviously, but also `./configure --toolchains enabled` as the flag will no longer get recognized.~~
Not a breaking change anymore, I just changed the default.

### DIFF
--- a/boot/configure.ml
+++ b/boot/configure.ml
@@ -18,7 +18,7 @@ let out =
 ;;
 
 let default_toggles : (string * [ `Disabled | `Enabled ]) list =
-  [ "toolchains", `Disabled; "pkg_build_progress", `Disabled; "lock_dev_tool", `Disabled ]
+  [ "toolchains", `Enabled; "pkg_build_progress", `Disabled; "lock_dev_tool", `Disabled ]
 ;;
 
 let toggles = ref default_toggles
@@ -89,8 +89,7 @@ let () =
     ; ( "--toolchains"
       , toggle "toolchains"
       , " Enable the toolchains behaviour, allowing dune to install the compiler in a \
-         user-wide directory.\n\
-        \      This flag is experimental and shouldn't be relied on by packagers." )
+         user-wide directory." )
     ; ( "--pkg-build-progress"
       , toggle "pkg_build_progress"
       , " Enable the displaying of package build progress.\n\
--- a/flake.nix
+++ b/flake.nix
@@ -63,7 +63,6 @@
       add-experimental-configure-flags = pkg: pkg.overrideAttrs {
         configureFlags =
           [
-            "--toolchains" "enable"
             "--pkg-build-progress" "enable"
             "--lock-dev-tool" "enable"
           ];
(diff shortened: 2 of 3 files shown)


ElectreAAS:
cc @maiste I think the binary release would need to be updated if this is merged.

maiste (changes requested):
Indeed! To prevent this from happening, could you also update the `/flake.nix` file and remove [the flag](https://github.com/ocaml/dune/blob/b61e9d612642dbd1fd58d25c90126d607e69bd43/flake.nix#L66), please? :smile:  Nothing to do on the binary-distribution side.

ElectreAAS:
> could you also update the /flake.nix file and remove [the flag](https://github.com/ocaml/dune/blob/b61e9d612642dbd1fd58d25c90126d607e69bd43/flake.nix#L66), please?

Done!

rgrinberg:
Instead of removing it, you could also just flip the default. Not sure if we'd like to keep this mode, but we don't have to delete it right away

gridbugs (approved):

maiste (approved):
Depending on if you prefer to go with @rgrinberg solution or not. But otherwise, LGTM.

src/dune_rules/pkg_toolchain.ml
```diff
@@ -40,18 +40,15 @@ let pkg_dir (pkg : Dune_pkg.Lock_dir.Pkg.t) =
...
-      [ Package_name.of_string "ocaml-base-compiler"
-      ; Package_name.of_string "ocaml-variants"
-      ]
-    in
-    List.mem compiler_package_names name ~equal:Package_name.equal
-  | `Disabled -> false
+let is_compiler name =
+  let module Package_name = Dune_pkg.Package_name in
+  let compiler_package_names =
+    (* TODO don't hardcode these names here *)
```
samoht, line 46: You could probably check this by looking at the `compiler` value in the list of `flags: ` (in the opam file).

ElectreAAS: Indeed, that would be cleaner. However I can't find a way to transmute a `Dune_lang.Package_name.t` into a `Dune_pkg.Opam_file.t = OpamParserTypes.FullPos.opamfile`, which would allow looking into the flags

As this is not related to the core of this PR, I'll leave the TODO for now

ElectreAAS:
I ended up with only the soft approach: make it enabled by default but not delete it entirely.
If in the future we feel this is unnecessary, we can always remove it

Issue #11033, opened 2024-10-20
Error during linking: file too short 


anmonteiro:
While testing #19937 in nix-overlays, I came across a weird failure:

```
/nix/store/vcvhwiilizhijk7ywyn58p9l005n9sbn-binutils-2.43.1/bin/ld: error: /nix/store/bm87ivmiccvgz2265159y990pzjabrmc-ocaml4.14.2-melange-4.0.0/lib/ocaml/4.14.2/site-lib/melange/js_parser/js_parser.a: file too short
collect2: error: ld returned 1 exit status
File "caml_startup", line 1:
Error: Error during linking (exit code 1)
-> required by _build/default/.ppx/9b08511fbad9c15355a1e1f8e80934bc/ppx.exe
-> required by _build/default/src/Fetch.pp.mli
-> required by _build/default/src/.fetch.objs/melange/fetch.cmi
-> required by _build/install/default/lib/melange-fetch/melange/fetch.cmi
-> required by _build/default/melange-fetch.install
-> required by alias install
```

this only seems to happen when all of the below are true:
- Linux
- under OCaml 4.14

I don't know what to make of this, and I'm wondering if it could be related to any change in dune between d6f1a2740ac584127105773a43283829ae7a39e7 and current main?

anmonteiro:
I can't repro this with the latest HEAD commi.

PR #11051, opened 2024-10-29
[3.16] backport: #11009 #10962

maiste:
This PR backports https://github.com/ocaml/dune/pull/11009 and #10962. They are backported together as one is the fix of the other one.

PR #11054, opened 2024-10-29
test: forward port cpp test for 4.14.2

maiste:
This PR forward ports some changes we have found while releasing 3.16.1. There was an issue with the order of the headers: it was ereasing some informations and make the test crash with `ocaml.4.14.2`.

### DIFF
--- a/test/blackbox-tests/test-cases/cpp-std-ocaml5.t/run.t
+++ b/test/blackbox-tests/test-cases/cpp-std-ocaml5.t/run.t
@@ -15,5 +15,5 @@
   >   (flags -std=c++98 :standard)))
   > EOF
 
-  $ dune exec ./cpp11.exe 2>/dev/null 
+  $ dune exec ./cpp11.exe
   Hi from C++11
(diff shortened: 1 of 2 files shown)


anmonteiro:
does this fix https://github.com/ocaml/dune/issues/11033?

test/blackbox-tests/test-cases/cpp-std-ocaml5.t/run.t
```diff
@@ -15,5 +15,5 @@
   >   (flags -std=c++98 :standard)))
   > EOF
 
-  $ (dune build ./cpp11.exe 2>/dev/null) && _build/default/cpp11.exe
+  $ dune exec ./cpp11.exe 2>/dev/null
```
Leonidas-from-XIV, line 18:
```suggestion
  $ dune exec ./cpp11.exe
```

Leonidas-from-XIV (changes requested):

Leonidas-from-XIV:
@anmonteiro No, this fixes the compilation of a cram test on OCaml <5. The original PR (#10962) was tested on 5.1.1 but backporting the fix to 3.16.x (which uses 4.14.2) revealed that there's a compilation error.

Leonidas-from-XIV (approved):
Thanks, looks good. And thanks @kit-ty-kate for pointing out the fix in #11051.

PR #11059, opened 2024-10-30
Preview: Upgrade to OCamlformat 0.27.0-preview1 (unreleased)

Julow:
The aim of this preview is to gather feedback.

Changelog can be found here: https://github.com/ocaml-ppx/ocamlformat/blob/main/CHANGES.md


bin/describe/describe_workspace.ml
```diff
@@ -72,9 +72,9 @@ module Descr = struct
   module Mod_deps = struct
     type t =
       { for_intf : Dune_rules.Module_name.t list
-          (* direct module dependencies for the interface *)
+            (* direct module dependencies for the interface *)
```
Leonidas-from-XIV, line 75: This seems to be indented too bar. The original is already misindented but this makes it worse.

bench/metrics.mli
```diff
@@ -6,41 +6,41 @@ open Stdune
     [unzip] functions which make serialisation easier. *)
 type ('float, 'int) t =
   { elapsed_time : 'float
-  (** Real time elapsed since the process started and the process
+    (** Real time elapsed since the process started and the process
```
Leonidas-from-XIV, line 9: I like that the comments beginnings are aligned with the field they're documenting.

bench/gen_synthetic_dune_watch.ml
```diff
@@ -68,17 +68,21 @@ let write_lib ~base_dir ~lib ~dune =
 
 let write base_dir =
   let () = Unix.mkdir base_dir 0o777 in
-  let dune = {|
+  let dune =
```
Leonidas-from-XIV, line 71: This seems a bit wasteful.

bench/gen_synthetic_dune_watch.ml
```diff
@@ -21,10 +21,10 @@ let write_subset base_dir library_index subset =
             List.flatten
               (List.map
                  (fun k ->
-                   List.map
-                     (fun j ->
-                       sprintf "M_%d_%d_%d_%d.f()" (library_index - 1) j mod_rows k)
-                     (count subsets_per_library))
+                    List.map
```
Leonidas-from-XIV, line 24: I like that the function bodies are indented by two from the `fun` and not from the `(` in front of `fun`. Seems more consistent.

bench/micro/path_bench.ml
```diff
@@ -8,56 +8,54 @@ let root = "."
...
-                                       ; "long paths", (long_path, long_path)
-                                       ]])
+let%bench_fun ("Path.Local.append"
+                [@params
+                  t
+                  = [ "left root", (".", long_path)
+                    ; "right root", (long_path, ".")
+                    ; "short paths", (short_path, short_path)
+                    ; "long paths", (long_path, long_path)
+                    ]])
```
Leonidas-from-XIV, line 58: Much more readable, only a bit surprising where the `[` starts. It is aligned with the content of the string? What if the start is `{foo|`, where is the indent? I think just 2 characters from the item to which the attribute is attached would be more consistent.

bin/describe/describe_workspace.ml
```diff
@@ -542,10 +542,10 @@ module Crawl = struct
...
   let workspace
-    options
-    ({ Dune_rules.Main.contexts = _; scontexts } : Dune_rules.Main.build_system)
-    (context : Context.t)
-    dirs
+        options
+        ({ Dune_rules.Main.contexts = _; scontexts } : Dune_rules.Main.build_system)
+        (context : Context.t)
+        dirs
     : Descr.Workspace.t Memo.t
```
Leonidas-from-XIV, line 549: A bit undecided on this. It's kind of nice that the arguments are not indented the same amount as the function body (as that makes them bleed together), but the return type still is which is quite inconsistent.

Leonidas-from-XIV (commented):
I haven't looked through all (obviously and also sensibly because the changes are often of the same type), but here's my observations. I think some of them are solid improvements, some of them could be improved still.

bench/micro/path_bench.ml
Julow: It was in fact indented by 2 starting from the opening parenthese. Here's a more intuitive indentation: https://github.com/ocaml-ppx/ocamlformat/pull/2613
But because this is already changed, a better formatting is possible: https://github.com/ocaml-ppx/ocamlformat/pull/2615

bin/describe/describe_workspace.ml
Leonidas-from-XIV: Excellent, glad to hear. The new aligning is very neat, I'm looking forward to the final release.

bench/gen_synthetic_dune_watch.ml
Julow: This was changed in https://github.com/ocaml-ppx/ocamlformat/pull/2480 but perhaps with a different case in mind.
@dvulakh Do you have an opinion ?

dvulakh: When formatting with `--profile=janestreet`, we want to keep the new behavior (multiline delimited strings always get their own lines).

We don't mind if `--profile=ocamlformat` or `--profile=default` keep some multiline strings docked, though.

Julow:
What's happening in CI ?

boot/libs.ml
```diff
@@ -1,4 +1,4 @@
-let external_libraries = [ "unix"; "threads" ]
+let external_libraries = [ "threads.posix" ]
```
Leonidas-from-XIV, line 1: The CI failure is probably because of this change.

Julow: Our script is:
```
make release
_build/default/bin/dune.exe build @fmt --auto-promote
```
Which promotes this change. If I run this again locally, the change is undone. Do you have an idea on what's happening ?

Leonidas-from-XIV: There's some trickery around `threads.posix` in `src/dune_rules/bootstrap_info.ml` but I don't know exactly what is happening there and why the inputs seem to be different. Looking at the mention of `threads.posix` in the odoc rules it seems like `threads` is somewhat arcanely set up and kinda aliases to `threads.vm` or `threads.posix`.

Though I think it it would be better to run `make bootstrap` followed by `./dune.exe fmt` to apply the formatting.

Julow: That's fixed now.

rgrinberg:
LGTM. Can we wait until this version is available in nixpkgs though?

anmonteiro:
@rgrinberg it is https://github.com/NixOS/nixpkgs/pull/361406

Julow:
As this is not a PR that can be reviewed or rebased, I'll close it and let a maintainer do the upgrade when appropriate.

PR #11064, opened 2024-11-01
Add action.yaml file for dune

emilpriver:
I just set up Dune for DBCaml and thought it would be nice to have an official action for It.

The dev would invoke the action by:
``` 
- uses: ocaml/dune@main
```

This would download Dune and install it to the path. A bigger example of how this would work is for instance:

``` 
- uses: OCaml/dune@main
- name: run tests
  run: dune runtest
```

Does the install script have any versioning?

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -1,0 +1,14 @@
+name: ocaml/dune
+description: Setup dune using the latest dune release
+branding:
+  icon: ci
+  color: green
+author: ocaml
+
+runs:
+  using: "composite"
+  steps:
+    - name: Download and install dune
+      run: curl -fsSL https://get.dune.build/install | sh
+    - name: Add dune to PATH
+      run: echo "$HOME/.dune/bin" >> "$GITHUB_PATH"


maiste:
Thanks for the PR :+1: 

As Package management is not entirely stabilised into `dune`, having a "stable" action is not really at the day order. Also, we might not want to have an action pointing to a "flaky" executable. Indeed, every day (modulo changes in this repo), you would catch a different version of `dune`. It wouldn't be a good developer experience as the behaviour can change... If we had to create an action, I would recommand:
- it lives inside its own repository in https://github.com/ocaml-dune and,
- it points to a stable version of Dune, not the preview one.

Another place where the action could be living into is [ocaml/setup-ocaml](https://github.com/ocaml/setup-ocaml) as it is the action most OCaml developer would get.

Leonidas-from-XIV:
I agree that `setup-ocaml` would probably be the right place for this to be and on a distinct second a `ocaml-dune/setup-dune` repo that would host them if `setup-ocaml` is not interested in running it.

Also, I think such an action should probably build dune from source instead of downloading a prebuilt binary.

emilpriver:
Hey!

I agree with both your takes. I didn't know if we should use the preview or not. But let's wait with this until the dune developer preview is done and then we can look into this again

Issue #11012, opened 2024-10-15
Coq cannot find the cmxs file for transitive ocaml dependencies


rlepigre:
When building Coq theories with plugins depending on OCaml libraries (everything being in a dune workspace) it seems like dune sometimes fails to place `.cmxs` files for transitive dependencies where Coq expects to find them. Since Coq relies on `findlib`, it seems like all the OCaml libraries depended on transitively by plugins should be made available in the `_build/install` directory (or at least a `META` file for them?).

Here is a tarball with a minimal example of workspace that exhibits the problem: [dune_bug_plugin_deps.tar.gz](https://github.com/user-attachments/files/17377866/dune_bug_plugin_deps.tar.gz). It contains the following files:
```
dune_bug_plugin_deps
├── coqlib
│   ├── coqlib.opam
│   ├── coqlib-plugin.opam
│   ├── dune-project
│   ├── plugin
│   │   ├── coqlib_plugin.mlpack
│   │   ├── dune
│   │   └── main.ml
│   └── theories
│       ├── dune
│       └── plugin.v
├── dune-workspace
└── somelib
    ├── dune-project
    ├── somelib.opam
    └── src
        ├── dune
        └── version.ml
```
To reproduce the problem, simply run:
```shell
$ cd dune_bug_plugin_deps
$ dune build coqlib/theories/plugin.vo
File "./coqlib/theories/plugin.v", line 1, characters 0-41:
Error:
Dynlink error: error loading shared library: Failure("[...]/dune_bug_plugin_deps/_build/install/default/lib/somelib/somelib.cmxs: cannot open shared object file: No such file or directory")
```

This was tested with Coq 8.19, and dune 3.16.0.

rlepigre:
cc @ejgallego @Alizter

rlepigre:
Note also that the example provides a workaround in file `dune_bug_plugin_deps/coqlib/theories/dune`:
```
(coq.theory
 (name coqlib)
 (package coqlib)
 (plugins coqlib-plugin.plugin))

;(rule
; (targets dummy.v)
; (deps (package coqlib-plugin))
; (action (with-stdout-to %{targets} (echo "(* dummy file *)"))))
```
Removing the comments on the rule will fix the issue by forcing a dependency over the `coqlib-plugin` package (which is the reason to have a separate package for the plugin in the example, I think the problem is the same if it is in the same package as the theory).

ejgallego:
[...]

PR #11091, opened 2024-11-04
Add reproduction test-case for #11012.

rlepigre:
This is based on the test case sent in [dune#11012](https://github.com/ocaml/dune/issues/11012).

Currently, this tests the (unexpected) failure.


rlepigre:
This has been fixed with more recent versions of Rocq, so I wonder if Ci should be bumped to using Rocq 9. What do you think @Alizter @ejgallego?

rlepigre:
(This is probably a good test case to merge if the Rocq version gets bumped.)

PR #11092, opened 2024-11-04
Explain how to override default entry point

benji-sb:
This PR adds a page in the "Advanced Topics" section of the doc. This new page explains how to override the default entry point in C and have dune compile everything.

For more context on why this could be useful, you can check the [Wildcard expansion on Windows](https://discuss.ocaml.org/t/wildcard-expansion-on-windows/15461) topic on the Ocaml Discourse.


doc/advanced/override-default-entrypoint.rst
```diff
@@ -0,0 +1,41 @@
...
+redefines the entry point:
+
+.. code:: C
+
+    #include <stdio.h>
+
+    #include "caml/mlvalues.h"
+    #include "caml/callback.h"
+
+    CAMLextern void caml_do_exit (int);
```
maiste, line 20:
```suggestion
    #include "caml/sys.h"
```
I'm not a C developer, but someone from the compiler suggested that this function prototype was available in this caml header.

doc/advanced/override-default-entrypoint.rst
```diff
@@ -0,0 +1,41 @@
...
+
+    int main_os(int argc, char_os **argv)
```
maiste, line 22: It can be clearer to add a comment to specify that this is the entry point you are trying to redefine.

doc/advanced/override-default-entrypoint.rst
```diff
@@ -0,0 +1,41 @@
+Overriding the default entry point
+----------------------------------
+
+In some cases, it may be necessary to override the default main entry point of
+an Ocaml program. For example, this is the case if you want to let your program
+handle argument wildcards expansion on Windows.
+
+Fortunately, the ``foreign_stubs`` stanza can be leveraged for that usage.
```
maiste, line 8: Adding a pointer to the `foreign_stubs` reference might help the reader to dig further if they want to.

maiste (commented):
Thanks for the contribution! :+1: 
We are trying to rely more and more on the [Diátaxis](https://diataxis.fr) framework to improve the Dune documentation. In this situation, I would move this to the How-to guide section in the Dune documentation. 

@Leonidas-from-XIV, what is your two cents on this?

doc/advanced/override-default-entrypoint.rst
```diff
@@ -0,0 +1,41 @@
+Overriding the default entry point
+----------------------------------
+
+In some cases, it may be necessary to override the default main entry point of
+an Ocaml program. For example, this is the case if you want to let your program
```
Leonidas-from-XIV, line 5:
```suggestion
an OCaml program. For example, this is the case if you want to let your program
```

doc/advanced/override-default-entrypoint.rst
Leonidas-from-XIV: I agree, reading this I was confused where the redefinition happens and only then guessed that it happens because `main_os` is called this. I think it would be nicer to spell that out somewhere explicitely.

Leonidas-from-XIV (changes requested):
I agree with @maiste, it's definitely more a howto, as it doesn't cover any particular advanced topic about Dune, more about how C stubs work but this has more to do with OCaml than Dune.

Still, we can definitely include it in the documentation.

doc/advanced/override-default-entrypoint.rst
benji-sb: Good idea, done.

benji-sb:
> I agree with @maiste, it's definitely more a howto, as it doesn't cover any particular advanced topic about Dune, more about how C stubs work but this has more to do with OCaml than Dune.
> 
> Still, we can definitely include it in the documentation.

Moved to the howto section, thanks for the suggestion!

doc/howto/override-default-entrypoint.rst
```diff
@@ -0,0 +1,51 @@
...
+compile and link our OCaml program with the new C entry point defined in
+``main.c``:
+
+.. code:: dune
+
+    (executable
+     (name hello)
+     (foreign_stubs
+      (language c)
+      (names main)))
```
Leonidas-from-XIV, line 51: I think it would also be nice to have an example showing that it works and what it is supposed to output.

Leonidas-from-XIV (commented):
I think it could have a bit of a "this is how the result works" section, otherwise it looks good from my point of view.

@christinerose could you please take a look at the text if it is ok that way?

doc/howto/override-default-entrypoint.rst
```diff
@@ -0,0 +1,51 @@
+How to override the default entry point
```
maiste, line 1:
```suggestion
How to override the default OCaml entrypoint with C Stubs
```
My suggestion is not particularly good, but I think it is important to emphasise the link with stubs and with OCaml.

maiste (commented):
Thanks, the update looks good! I have just one comment and, otherwise, as @Leonidas-from-XIV said, I let @christinerose do her magic (thanks in advance for having a look at this :pray: )

Also, you need to update your commit to comply with the DCO (see [the information](https://github.com/ocaml/dune/pull/11092/checks?check_run_id=32660608130)).

doc/howto/override-default-entrypoint.rst
benji-sb: Added, thanks for the suggestion!

doc/howto/override-default-entrypoint.rst
benji-sb: Done.

doc/advanced/override-default-entrypoint.rst
maiste: Better, thanks for the update :)

doc/howto/override-default-entrypoint.rst
```diff
@@ -0,0 +1,62 @@
...
+
+.. code:: dune
+
+    (executable
+     (name hello)
+     (foreign_stubs
+      (language c)
+      (names main)))
+
+The program can be compiled using ``dune``. When run, the output shows that it
```
christinerose, line 54: Do you. mean using Dune, the build system, or `dune` the command. I see you use `dune build` below, so should it say that instead? (Just to clarify 🙂 )

benji-sb: I meant using Dune, the build system, invoked through the `dune` command. I rephrased to hopefully make this a bit clearer. Is it any better now?

Alizter:
Just a quick comment: OCaml doesn't have a notion of "entry point" since basically everything is run in order. Therefore calling it the "C Entrypoint" might be better. This might be clearer to new users who might think that OCaml has entry points.

doc/howto/override-default-entrypoint.rst
christinerose: Yes! Perfect!

benji-sb:
> Just a quick comment: OCaml doesn't have a notion of "entry point" since basically everything is run in order. Therefore calling it the "C Entrypoint" might be better. This might be clearer to new users who might think that OCaml has entry points.

I edited the document accordingly, thanks a lot for bringing up this point!

maiste (approved):
LGTM :+1:

Leonidas-from-XIV (approved):

maiste:
The error in the CI is completely unrelated with the PR (as it just modifies the documentation). I think we can merge it.

Issue #11096, opened 2024-11-04
dune dependency error with `dune pkg lock`


maiste:
The #10918 introduced a bug or a regression in the way the `dune pkg lock` command work. I checked and the behavior wasn't present in the commit before.

## Expected Behavior

When running a project with `dune` in the `depends` stanza of a _dune-project_ file, it should build without any error. It was the previous behaviour before the mentioned PR.

## Actual Behavior

The computation fails, and it prints the following stack trace:
```
Internal error, please report upstream including the contents of _build/log.
Description:
  ("can't find a valid solution for the dependencies", { name = "foo" })
Raised at Stdune__Code_error.raise in file
  "otherlibs/stdune/src/code_error.ml", line 10, characters 30-62
Called from Dune_pkg__Opam_solver.reject_unreachable_packages.(fun) in file
  "src/dune_pkg/opam_solver.ml", lines 830-832, characters 14-56
Called from Stdlib__Map.Make.merge in file "map.ml", line 406, characters
  44-63
Called from Stdlib__Map.Make.merge in file "map.ml", line 403, characters
  64-79
Called from Stdlib__Map.Make.merge in file "map.ml", line 403, characters
  25-40
Called from Stdlib__Map.Make.merge in file "map.ml", line 403, characters
  25-40
Called from Stdlib__Map.Make.merge in file "map.ml", line 403, characters
  64-79
Called from Dune_pkg__Opam_solver.reject_unreachable_packages.(fun) in file
  "src/dune_pkg/opam_solver.ml", lines 802-841, characters 6-32
Called from Dune_pkg__Opam_solver.solve_lock_dir.(fun) in file
  "src/dune_pkg/opam_solver.ml", line 949, characters 10-78
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
Re-raised at Stdune__Exn.raise_with_backtrace in file
  "otherlibs/stdune/src/exn.ml", line 38, characters 27-56
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11

I must not crash.  Uncertainty is the mind-killer. Exceptions are the
little-death that brings total obliteration.  I will fully express my cases. 
Execution will pass over me and through me.  And when it has gone past, I
will unwind the stack along its path.  Where the cases are handled there will
be nothing.  Only I will remain.
```

## Reproduction

[...]

PR #11098, opened 2024-11-04
test: extend dune-constraint test to reproduce #11096

maiste:
This PR adds a reproduction case for #11096. It modifies the test about the implicit dune constraint instead of adding a new one for readability. Indeed, this test is also about `dune` constraint.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/implicit-dune-constraint.t
+++ b/test/blackbox-tests/test-cases/pkg/implicit-dune-constraint.t
@@ -1,6 +1,7 @@
 By default, we introduce a constraint on in the build plan that will require
 the dune version to match the version of dune being used to generate the
-constraint.
+constraint. On another hand, we ensure `dune` can be used as a declared 
+dependency.
 
   $ . ./helpers.sh
   $ mkrepo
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/pkg/implicit-dune-constraint.t
```diff
@@ -28,3 +29,39 @@ constraint.
...
+    "otherlibs/stdune/src/exn.ml", line 38, characters 27-56
+  Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
+    line 76, characters 8-11
+  
+  I must not crash.  Uncertainty is the mind-killer. Exceptions are the
+  little-death that brings total obliteration.  I will fully express my cases. 
+  Execution will pass over me and through me.  And when it has gone past, I
+  will unwind the stack along its path.  Where the cases are handled there will
+  be nothing.  Only I will remain.
+  [1]
```
Leonidas-from-XIV, line 67: Can you adjust the test to filter out just e.g. the description? The whole stacktrace is inconvenient as the line numbers will cause unrelated failures.

Leonidas-from-XIV (changes requested):
Thanks, it's really surprising we didn't have a test for this before so this will come in handy to test how `dune` as a dependency is dealt with.

Issue #11058, opened 2024-10-30
dune crashes on `dune pkg lock`


maxRN:
## Expected Behavior
dune doesn't crash and the lock dir is generated.

## Actual Behavior
dune crashes.

## Reproduction
```
$ dune pkg lock
Internal error, please report upstream including the contents of _build/log.
Description:
  ("invalid depopts", {})
Raised at Stdune__Code_error.raise in file
  "otherlibs/stdune/src/code_error.ml", line 10, characters 30-62
Called from Dune_pkg__Opam_solver.resolve_depopts in file
  "src/dune_pkg/opam_solver.ml", line 565, characters 3-21
Called from Dune_pkg__Opam_solver.opam_package_to_lock_file_pkg in file
  "src/dune_pkg/opam_solver.ml", line 645, characters 6-48
Called from Stdlib__List.rev_map.rmap_f in file "list.ml", line 105,
  characters 22-25
Called from Stdune__List.map in file "otherlibs/stdune/src/list.ml"
  (inlined), line 5, characters 19-33
Called from Dune_pkg__Opam_solver.solve_lock_dir.(fun) in file
  "src/dune_pkg/opam_solver.ml", line 856, characters 8-277
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
Re-raised at Stdune__Exn.raise_with_backtrace in file
  "otherlibs/stdune/src/exn.ml" (inlined), line 38, characters 27-56
Called from Stdune__Exn_with_backtrace.reraise in file
  "otherlibs/stdune/src/exn_with_backtrace.ml", line 20, characters 33-71
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11

I must not crash.  Uncertainty is the mind-killer. Exceptions are the
little-death that brings total obliteration.  I will fully express my cases.
Execution will pass over me and through me.  And when it has gone past, I
will unwind the stack along its path.  Where the cases are handled there will
be nothing.  Only I will remain.
```

### Output of `_build/log`
https://gist.github.com/maxRN/27b87ae926eece0f98c54780547bd6eb

## Specifications

- Version of `dune` (output of `dune --version`):
`"Dune Developer Preview: build 2024-10-30T02:16:37Z, git revision
d84d83a4becd373549757d907ba4e4d5434c5e0f"`

- Version of `ocaml` (output of `ocamlc --version`): none
- nothing
[...]

PR #11099, opened 2024-11-05
Allow arbitrary opam depopt disjunction represntations

gridbugs:
Opam currenly represents lists of depopts packages with disjunctions that "stack" no the left-side, such as: `Or(Or(A, B), C)`. This code was originally written to expect disjunctions to stack on the right-side (`Or(A, Or(B, C))`). This change generalizes the resolving of depopts so that the representation does not matter, as long as every node is `Or`, `Atom`, or `Empty`.

Since we now traverse both sides of the `Or` node it's tricky to keep the `collect` function tail-recursive so the code is simplified to avoid collecting the atoms into an accumulator. Packages are expected to have a small number of depopts so there's little benefit to keeping it tail-recursive given the code complexity it would add.

Fixes https://github.com/ocaml/dune/issues/11058

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -551,19 +551,18 @@ let opam_file_is_compiler (opam_package : OpamFile.OPAM.t) =
 ;;
 
 let resolve_depopts ~resolve depopts =
-  (let rec collect acc depopts =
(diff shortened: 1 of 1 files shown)


rgrinberg:
Duplicate of https://github.com/ocaml/dune/pull/11090

PR #11103, opened 2024-11-05
fix: Add or remove dune from package sets to allow resolving packages depending on it

### LINKED ISSUE #11096: dune dependency error with `dune pkg lock`

Leonidas-from-XIV:
There's multiple ways to go about it:

1. Add some dune package in the space of valid packages. I was not sure how to get the dune version out of `dune_rpc` or whether that would even be the right thing, so I set the version to `dev`.
2. Remove `dune` from the dependency formulas. `Dependency_formula.remove_package formula (Package_name.Set.singleton Dune_dep.name)` is an alternate solution. 

Fixes #11096

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -111,10 +111,6 @@ module Context_for_dune = struct
     ~stats_updater
     ~constraints
     =
-    let dune_version =
-      let major, minor = Dune_lang.Stanza.latest_version in
-      OpamPackage.Version.of_string @@ sprintf "%d.%d" major minor
-    in
     let candidates_cache = Fiber_cache.create (module Package_name) in
     let constraints =
       List.map constraints ~f:(fun (constraint_ : Package_dependency.t) ->
@@ -136,7 +132,7 @@ module Context_for_dune = struct
     ; local_packages
     ; pinned_packages
     ; solver_env
-    ; dune_version
+    ; dune_version = Dune_dep.version
     ; stats_updater
     ; candidates_cache
     ; available_cache
@@ -785,7 +781,7 @@ let reject_unreachable_packages =
     loop roots;
     !seen
   in
-  fun solver_env ~local_packages ~pkgs_by_name ->
+  fun solver_env ~dune_version ~local_packages ~pkgs_by_name ->
     let roots = Package_name.Map.keys local_packages in
     let pkgs_by_version =
       Package_name.Map.merge pkgs_by_name local_packages ~f:(fun name lhs rhs ->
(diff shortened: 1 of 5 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -810,7 +810,13 @@ let reject_unreachable_packages =
             [ "name", Package_name.to_dyn name ]
         | Some (pkg : Lock_dir.Pkg.t), None -> Some (List.map pkg.depends ~f:snd)
         | None, Some (pkg : Local_package.For_solver.t) ->
-          let formula = pkg.dependencies |> Dependency_formula.to_filtered_formula in
+          (* remove Dune from the formula as we remove it from solutions *)
+          let formula =
```
maiste, line 814: This is mostly a question to understand the behavior. Will it keep the constraint on Dune somewhere, even if we remove it from the formula? This is just to understand the way it is computed underneath.

maiste (approved):
One question to understand but it LGTM :+1:

src/dune_pkg/opam_solver.ml
Leonidas-from-XIV: This is just to check which packages are reachable, it is not what is used to compute the solution. As you can see in CI, the `implicit-dune-deps.t` test that checks the dune version against a constraint is passing.

rgrinberg (approved):

rgrinberg:
We need to do a combination of 1 and 2. We need to pick the correct deps for a formula like `("a" "dune" {> 3}) | ("b" "dune" {< 3})`.

So we give the right dune version when evaluating the formula and then filtering out dune from the evaluated list. We do the same for external deps as well.

I pushed something along those lines.

PR #11108, opened 2024-11-10
Improve dune runtest hints for non-test files

Alizter:
### Issue

When running dune runtest on a file that is not a test, dune hints

```
Hint: did you mean <file>
```
even though the user just gave `<file>`.

### Fix

This stems from a deeper issue with `User_message.did_you_mean` which accepts hints that are identical to the original name. It doesn't make sense to give a "did you mean" hint in the case they are identical, so we check that the Levenshtein distance is greater than 0.

This removes the useless "did you mean" as observed in the tests.

We will improve the error message in a follow up PR.

### DIFF
--- a/test/blackbox-tests/test-cases/runtest-cmd.t
+++ b/test/blackbox-tests/test-cases/runtest-cmd.t
@@ -104,5 +104,4 @@ messages are informative enough.
 - Running a non-test file should give a suitable error message
   $ dune test dune-project
   Error: "dune-project" was not found.
-  Hint: did you mean dune-project?
   [1]
--- a/test/blackbox-tests/test-cases/runtest-cmd.t
+++ b/test/blackbox-tests/test-cases/runtest-cmd.t
@@ -101,3 +101,8 @@ messages are informative enough.
   Error: "mytest1.t" was not found.
   Hint: did you mean mytest.t?
   [1]
+- Running a non-test file should give a suitable error message
+  $ dune test dune-project
+  Error: "dune-project" was not found.
+  Hint: did you mean dune-project?
+  [1]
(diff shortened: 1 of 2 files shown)


test/blackbox-tests/test-cases/runtest-cmd.t
```diff
@@ -101,3 +101,7 @@ messages are informative enough.
   Error: "mytest1.t" was not found.
   Hint: did you mean mytest.t?
   [1]
+- Running a non-test file should give a suitable error message
+  $ dune test dune-project
+  Error: "dune-project" was not found.
```
maiste, line 106: What about "`dune-project` is not a test file"? If the file is present, it might confuse the user, no?

Alizter: Yes, this is a confusing error message at the moment. Which is why it will be improved in a follow up PR. My goal here was not to improve the error message but stop the useless hint.

Alizter: This will become:
```
  $ dune test dune-project
  Error: "dune-project" does not match any known test.
  [1]
```
later.

maiste: I see. Sorry for the misunderstanding on my side!

maiste (approved):

Leonidas-from-XIV (approved):
I think this is a solid improvement. It would be nice to see a test changing that shows that it doesn't recommend the same thing as was inserted but I am reasonably certain that filtering out distance 0 does the right thing.

Alizter:
@Leonidas-from-XIV There are two commits here. One where the test is added and another where the fix is done, showing how the test changes.

Leonidas-from-XIV:
@Alizter Ah, sorry, I somehow that this test was added in this PR. Thanks for the PR; LGTM.

Issue #10239, opened 2024-03-08
Automatically add aliases for expect tests


johnridesabike:
## Desired Behavior

For Dune's built-in expect tests, where a test executable prints to stdout and Dune diffs the output with a `.expected` file, it would be useful if Dune provided an alias for each test.

Expect test diffs can be large so the ability to specify a single test would be useful. Right now, the options to execute a single expect test are either to put them in separate directories or manually define each with a rule instead of the `(tests ...)` stanza.

Dune already provides aliases for each cram test, so extending that feature to expect tests feels natural. 

## Example

Directory:

```
.
└── test
    ├── dune
    ├── test_1.expected
    ├── test_1.ml
    ├── test_2.expected
    └── test_2.ml
```

Contents of `test/dune`:
```
(tests (names test_1 test_2))
```

Commands:
```
$ dune build @test_1 # executes and diffs test_1
$ dune build @test_2 # executes and diffs test_2
```

johnridesabike:
> in your example you can already execute your tests individually by building the expected target. E.g. `dune build test_1.expected`.

This does not seem to be the same thing. Running `dune build example.expected` just copies `example.expected` to the build directory. It doesn't generate the accompanying `.output` file and diff the two.

I like the idea of adding `(alias)` to the test stanza too. My main reasoning for creating the aliases automatically is because that's how cram tests work already, so it seems like it would be consistent with that.

rgrinberg:
Okay, then the command should be `dune build example.output` or whatever. You just need to build the target that is produced by your command and doesn't live in the source tree.

johnridesabike:
Building an `.output` target will produce the file but print nothing and exit with status 0 even if the test should fail. It also does not allow promoting the output to the source tree's `.expected` file. The actual test comes from diffing the output with the expected file, which only exists in the `runtest` alias.

Here's a minimal example of what I'm describing.

Files:
```
dune-project:
(lang dune 3.14)

dune:
(test (name example))

example.ml:
let () = print_endline "fail"

example.expected:
pass
```

Commands:
```
$ dune build example.expected
$ dune build example.output
$ dune build @runtest
File "example.expected", line 1, characters 0-0:
------ example.expected
++++++ example.output
File "example.expected", line 1, characters 0-1:
-|pass
+|fail
```
The output of `dune build @runtest` is what we want, but with the added ability to specify an individual test (if there were multiple tests).

rgrinberg:
[...]

PR #11109, opened 2024-11-10
feature: add aliases for inline tests

Alizter:
For a given library with inlines tests, we add an alias with the name of the library for running those tests prefixed by `inline-test-`.

Further work on separating out the partition actions would allow us to have an alias for each partition.

For now, this will allow us to run inline tests with the runtest command. Further work on that will follow this PR.

This addresses part of https://github.com/ocaml/dune/issues/10239 the remaining part would be to add an alias to the `(test)` stanza.

- [x] changelog
- [x] documentation

### DIFF
--- a/src/dune_rules/inline_tests.ml
+++ b/src/dune_rules/inline_tests.ml
@@ -336,9 +336,8 @@ include Sub_system.Register_end_point (struct
           ~link_args
           ~promote:None
       in
-      let partitions_flags =
-        partition_flags ~expander ~lib_name:(snd lib.name) ~backends
-      in
+      let lib_name = snd lib.name in
+      let partitions_flags = partition_flags ~expander ~lib_name ~backends in
       let deps, sandbox =
         let sandbox =
           let project = Scope.project scope in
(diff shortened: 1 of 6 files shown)


src/dune_rules/inline_tests.ml
```diff
@@ -343,35 +343,40 @@ include Sub_system.Register_end_point (struct
...
-             let diffs =
-               List.map source_files ~f:(fun fn ->
-                 Path.as_in_build_dir_exn fn
-                 |> Path.Build.extend_basename ~suffix:".corrected"
-                 |> Promote.Diff_action.diff ~optional:true fn)
-               |> Action.concurrent
-             in
-             Action.Full.make ~sandbox @@ Action.progn [ run_tests; diffs ]))
+        Memo.parallel_iter
+          [ runtest_alias; Alias.Name.of_string (Lib_name.Local.to_string lib_name) ]
```
hhugo, line 347: Can you make runtest depend on the <lib_name> alias instead? Aren't we running the tests twice if building both @runtest and @<lib_name> aliases?

Alizter: I don't think this is how aliases work. But maybe @rgrinberg can confirm. IIUC the aliases should point to the same target, so if they are both requested in the engine, only that target would end up getting built, and nothing will be run twice. But it's possible I have misunderstood the situation.

rgrinberg: Hugo's right, we are going to run the tests twice this way. You should create the alias and then add it as a dependency to runtest. A test to demonstrate this would be valuable as well btw.

> IIUC the aliases should point to the same target, so if they are both requested in the engine, only that target would end up getting built, and nothing will be run twice

What I think you you forgot is that the anonymous action we attach to the aliases will have a different digest. The digest of the anonymous action here will include the alias name, and therefore the action will be re-executed.

Alizter: It seems we are doing something similar in `Cram_rules.test_rule`. There we iterate over all the aliases and do `Super_context.add_alias_action` like I do here. I don't ever recall `@runtest` and `@mycramtest` building things twice.

Alizter: I've just pushed a commit which checks for the double execution of the anonymous action associated with each alias. I also added a test demonstrating the same erroneous behaviour with cram tests.

@rgrinberg What's the best way to get the `runtest` alias to depend on the test alias in each case? Use `Super_context.add_alias_action` with an empty action depending only on the test alias?

Alizter: I've pushed a fix for this and have shown it fixes the test.

Alizter:
@emillon I've updated the documentation but it could be better. Any ideas on improving it?

Alizter:
@rgrinberg I'm okay with the idea, but isn't it a bit confusing to run a test as `@lib:foo`? It would be yet anther thing to remember when writing a command?

Alizter:
I've pushed a commit changing it to `lib:` but I haven't squashed it yet while we still think about it. I guess if it is documented well and we write a clear tutorial about it, it should be no issue.

rgrinberg:
> @rgrinberg I'm okay with the idea, but isn't it a bit confusing to run a test as `@lib:foo`? Wouldn't it be yet anther thing to remember when writing a command?

I don't see how you some rather hard to debug cycles without this. You can reproduce this by:

1. Having a library `foo` with inline tests
2. Creating a rule that generates a source file for `foo`
3. Making the rule depend on `@foo`

Alizter:
OK so the problem we have here is that `:` is considered a path separator only on windows. Technically a user can write `dune build @C:/foo/bar/myproject/myalias` with how we parse aliases. So the `:` does play a special role during path parsing. We can't have it in alias names or else we would allow `:` in filenames at least on windows.

Alizter:
I'm going to mark this PR as draft while I think of a way to validate alias names and keep them flexible.

PR #11111, opened 2024-11-10
refactor(ml_sources): classify tests stanzas

Alizter:
We allow for `Ml_sources.find_origin` to differentiate between tests stanzas and executables stanzas rather than treating them the same.

Without this, it is not possible to determine if a module came from a `(tests)` stanza or an `(executables)` stanza which is important when looking to run tests with `dune runtest`.

### DIFF
--- a/src/dune_rules/top_module.ml
+++ b/src/dune_rules/top_module.ml
@@ -44,6 +44,7 @@ let find_module sctx src =
          match origin with
          | Executables exes ->
            Exe_rules.rules ~sctx ~dir ~dir_contents ~scope ~expander exes
+         | Tests tests -> Test_rules.rules ~sctx ~dir_contents ~dir ~expander ~scope tests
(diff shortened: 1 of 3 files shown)


src/dune_rules/top_module.ml
```diff
@@ -44,6 +44,7 @@ let find_module sctx src =
          match origin with
          | Executables exes ->
            Exe_rules.rules ~sctx ~dir ~dir_contents ~scope ~expander exes
+         | Tests tests -> Test_rules.rules ~sctx ~dir_contents ~dir ~expander ~scope tests
```
Alizter, line 47: I'm introducing the `Test_rules` here, but I don't know if I should keep the old behaviour and use the `Exe_rules`. This seems more correct to me.

Alizter:
On second thoughts, this is not really what I wanted.

PR #11115, opened 2024-11-12
test: reproduction case for ignored `dune.lock`

Leonidas-from-XIV:
Patches (or any extra files) cannot be found because `dirs` excludes the `dune.lock` thus the included patch is invisible to Dune.

Thanks to @mefyl for reporting the case and @gridbugs for isolating the issue.


mefyl (approved):
Looks like the issue I had indeed, thanks !

PR #11116, opened 2024-11-12
Add failing test pulling empty directory from cache.

mefyl:
Dune cache does not restore empty directories. On top of not being very nice, this has really bad consequences with caching because dune's rule digest algorithm does take empty directories in account, so restoring a rule from cache will invalidate all dependent rules as the sources hash will have changed.

This is especially painful with the `dune pkg` feature, as the Ocaml 5.2.0 tarball has an empty `flexdll` directory at the root, so pulling it from the cache will rebuild absolutely everything.

I suppose the correct behavior is to properly restore empty directories from cache. If not, they probably shouldn't count towards rule digests.

### DIFF
--- a/test/blackbox-tests/test-cases/dune-cache/empty-dir.t
+++ b/test/blackbox-tests/test-cases/dune-cache/empty-dir.t
@@ -1,0 +1,34 @@
+Check the cache restores empty directories
+
+  $ export DUNE_CACHE=enabled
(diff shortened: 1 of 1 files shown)


dra27:
FWIW, not handling empty directories I think is generally regarded as a mistake in Git, albeit one that's likely to be lived with forever.

Morally, if a tarball would have extracted an empty directory then so should its cached action.

PR #11118, opened 2024-11-13
test: Sibling modules should be accessible from qualified parsers

brendanzab:
This PR adds a failing test regarding the current behavior of `(include_subdirs qualified)` when a Menhir parser attempts to refer to a sibling module, as reported in #11119.

### DIFF
--- a/test/blackbox-tests/test-cases/menhir/include-subdirs-siblings.t/lang/dune
+++ b/test/blackbox-tests/test-cases/menhir/include-subdirs-siblings.t/lang/dune
@@ -1,2 +1,0 @@
-(menhir
- (modules parser))
--- a/test/blackbox-tests/test-cases/menhir/include-subdirs-siblings.t/lang/dune
+++ b/test/blackbox-tests/test-cases/menhir/include-subdirs-siblings.t/lang/dune
@@ -1,0 +1,2 @@
+(menhir
+ (modules parser))
--- a/test/blackbox-tests/test-cases/menhir/include-subdirs-siblings.t
+++ b/test/blackbox-tests/test-cases/menhir/include-subdirs-siblings.t
@@ -1,0 +1,47 @@
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.21)
+  > (using menhir 3.0)
+  > EOF
+
+  $ cat >dune <<EOF
+  > (include_subdirs qualified)
+  > (executable
+  >  (name foo))
+  > EOF
+
+  $ cat >foo.ml <<EOF
+  > let () =
+  >   assert
+  >     (Lang.Parser.expr
+  >       (fun _ -> Lang.Parser.EOF)
+  >       (Lexing.from_string "")
+  >       =
+  >       Lang.Ast.Unit)
+  > EOF
+
+  $ mkdir -p lang
+
+  $ cat >lang/ast.ml <<EOF
+  > type expr =
+  >   | Unit
+  > EOF
+  $ cat >lang/dune <<EOF
+  > (menhir
+  >  (modules parser))
+  > EOF
+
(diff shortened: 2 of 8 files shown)


test/blackbox-tests/test-cases/menhir/include-subdirs-siblings.t/lang/dune
```diff
@@ -0,0 +1,2 @@
+(menhir
+ (modules parser))
```
Leonidas-from-XIV, line 2: From what I see the error message is correct, as this makes the `Ast` module invisible to Menhir.

brendanzab: Oh, why is that? I had assumed it would work similarly to other qualified modules - where you could mention sibling modules from within a subdirectory? See the directory tree from #11119:

```
├── dune
├── dune-project
├── foo.ml
├── lang
│   ├── ast.ml
│   ├── dune
│   ├── lexer.ml
│   └── parser.mly
└── run.t
```

If this is intentional is there a workaround for this, or is this pattern impossible with Menhir?

brendanzab: I’ve also tried the following from the root `dune` file:

```
(menhir
 (modules lexer/parser))
 ```

Comment (Leonidas-from-XIV) [reply test/blackbox-tests/test-cases/menhir/include-subdirs-siblings.t/lang/dune:2]: I would expect `modules` to work the same way as it does on other stanzas like `library` or `executable`, e.g.

`dune`:
```
(executable
  (name foo)
  (modules foo))
```

`foo.ml`
```ocaml
let () = Bar.main ()
```

`bar.ml`
```ocaml
let main () = ()
```

(The error you see on `(modules lang/parser)` is because `modules` expects names of modules in the current directory and `lang/parser.ml` is not a valid file name)

brendanzab: Would you prefer it if I used an example that used one of the approaches above, or are those also intended behavior? 🤔

brendanzab: I’ve also tried variations of `(modules Lang.Parser)`, `(modules Lang__Parser)`, `(modules dune__exe__Lang__parser)` to no avail in the `menhir` stanza.

brendanzab: I’d also note that [test/blackbox-tests/test-cases/menhir/menhir-include-subdirs.t](https://github.com/ocaml/dune/blob/main/test/blackbox-tests/test-cases/menhir/menhir-include-subdirs.t) currently uses a [dune file](https://github.com/ocaml/dune/blob/main/test/blackbox-tests/test-cases/menhir/menhir-include-subdirs.t/bar/dune) in the subdirectory to declare the `menhir` stanza (I might have got the idea from there). It's just rather useless without also being able to refer to its sibling modules.

brendanzab: @Leonidas-from-XIV If it helps, it seems #8987 was merged, which seems like a case where the `.mly` is the “group interface” of the subdirectory, which references a sibling module:
```
├── dune
├── dune-project
└── group
    ├── dune         (menhir stanza is here)
    ├── group.mly    (refers to M)
    └── m.ml
```
https://github.com/ocaml/dune/pull/8987/files

Not sure if that test is enough to covers this use-case case as well? But I’d hope that if that issue was fixed, this would be supported as well!

anmonteiro:
This seems worth fixing. I pushed a commit converting the test into a single file.

Issue #11113, opened 2024-11-11
`dune pkg lock` fails when pinned repo has a tag with the same name as a branch


huwaireb:
## Expected Behavior
Running `dune pkg lock` succeeds.

## Actual Behavior
Instead, it fails with the following

```
Internal error, please report upstream including the contents of _build/log.
Description:
  ("Map.of_list_exn", { key = "v0.9.5.0" })
Raised at Stdune__Code_error.raise in file
  "otherlibs/stdune/src/code_error.ml", line 10, characters 30-62
Called from Dune_pkg__Rev_store.remote.(fun).f.(fun) in file
  "src/dune_pkg/rev_store.ml", line 793, characters 26-66
Called from Fiber__Core.O.(>>|).(fun) in file "vendor/fiber/src/core.ml",
  line 253, characters 36-41
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
Re-raised at Stdune__Exn.raise_with_backtrace in file
  "otherlibs/stdune/src/exn.ml" (inlined), line 38, characters 27-56
Called from Stdune__Exn_with_backtrace.reraise in file
  "otherlibs/stdune/src/exn_with_backtrace.ml", line 20, characters 33-71
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
Re-raised at Stdune__Exn.raise_with_backtrace in file
  "otherlibs/stdune/src/exn.ml" (inlined), line 38, characters 27-56
Called from Stdune__Exn_with_backtrace.reraise in file
  "otherlibs/stdune/src/exn_with_backtrace.ml", line 20, characters 33-71
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11

I must not crash.  Uncertainty is the mind-killer. Exceptions are the
little-death that brings total obliteration.  I will fully express my cases.
Execution will pass over me and through me.  And when it has gone past, I
will unwind the stack along its path.  Where the cases are handled there will
be nothing.  Only I will remain.
```

## Reproduction

```dune
(pin
 (url "git+https://github.com/FStarLang/FStar.git")
 (package
  (name fstar)))

(package (name repro) (depends fstar))
```

## Specifications

- Version of `dune` (output of `dune --version`): n/a, off head, commit f7af2d0078f25a54db19709e25ac73ca80c48183
[...]

PR #11120, opened 2024-11-13
test: Add repro case for #11113

Leonidas-from-XIV:
Building the FStar repo fails because the branch and tag are named the same, as reported in #11113.

This simplifies the setup into a test-case.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/pin-stanza/git-source.t
+++ b/test/blackbox-tests/test-cases/pkg/pin-stanza/git-source.t
@@ -5,9 +5,11 @@ Package sources can be set to git:
   $ mkrepo
   $ add_mock_repo_if_needed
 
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/pkg/pin-stanza/git-source.t
```diff
@@ -29,3 +31,13 @@ Package sources can be set to git:
...
   - foo.dev
+
+We create a tag that clashes with the name of the branch (hence we needed to
+fix the name of the branch eariler):
+
+  $ git -C _repo tag duplicated
+
+This should work but it fails at the moment:
+
+  $ dune pkg lock 2> /dev/null
```
rgrinberg, line 42: showing the first line of the error would be useful

rgrinberg (approved):

Issue #11017, opened 2024-10-17
Coqdoc flags --with-header and --with-footer should imply a dependency on the argument files


rlepigre:
The `--with-header` and `--with-footer` arguments of `coqdoc` (specified via a `(coqdoc_flags ...)` thingy) have file argument that should be dependencies of all the generated documentation files. It appears that they are currently not considered dependencies, which means that the documentation is not recompiled when the header / footer files change. (And the situation is worse when using the cache.)

To work around this, I typically put these files in `/tmp/some_folder/` using a wrapper makefile, and disable the cache. However, this leads to very long building times that could otherwise mostly hit the cache.

ejgallego:
That should be really easy to fix, I wonder tho if dune can support having `(coqdoc_flags (--with-header=%{dep:foo.html}))` directly?

Fix can be either by:
- analizing the flags and injecting the deps
- adding / refining a new field for the header options

Alizter:
I would suggest adding a new field for these kinds of options. The general `coqdoc_flags` field can be thought of as a last-resort tweak for features unsupported.

rlepigre:
I like the idea of making this separate fields. I tried looking into this, but I am not sure what type to use for the new fields in `Coq_env` (and elsewhere). Should it be a `Path.t` or something like that? Or rather `String_with_vars.t`? I'd like to see what's done for similar fields, but I cannot seem to find fields that just contain a path.

Alizter:
@rlepigre `String_with_vars.t` should be the correct one. We can validate it as a path at a later point. We want to be able to expand the given string if any percent forms are in there.

PR #11131, opened 2024-11-18
Add "coqdoc_header" and "coqdoc_footer" fields.

rlepigre:
This is an attempt at fixing https://github.com/ocaml/dune/issues/11017.

Although I have a working test-case, I have not done much testing so far. I'd appreciate early feedback anyway, if you can find the time to have a quick look @Alizter, @ejgallego.

What remains to be done:
- [x] Check for language version in the new fields.
- [x] Update the documentation.
- [x] Add change log entry.

### DIFF
--- a/bin/printenv.ml
+++ b/bin/printenv.ml
@@ -26,7 +26,9 @@ let dump sctx ~dir =
     Dune_rules.Menhir_rules.menhir_env ~dir
     |> Action_builder.of_memo
     >>= Dune_lang.Menhir_env.dump
-  and+ coq_dump = Dune_rules.Coq.Coq_rules.coq_env ~dir >>| Dune_rules.Coq.Coq_flags.dump
+  and+ coq_dump =
+    Dune_rules.Coq.Coq_rules.coq_env ~dir
+    >>| Dune_rules.Coq.Coq_flags.dump ~dir:(Path.build dir)
   and+ jsoo_js_dump =
     let module Js_of_ocaml = Dune_lang.Js_of_ocaml in
     let* jsoo = Action_builder.of_memo (Dune_rules.Jsoo_rules.jsoo_env ~dir ~mode:JS) in
--- a/src/dune_rules/coq/coq_stanza.ml
+++ b/src/dune_rules/coq/coq_stanza.ml
@@ -15,6 +15,7 @@ let coq_syntax =
     ; (0, 8), `Since (3, 8)
     ; (0, 9), `Since (3, 16)
     ; (0, 10), `Since (3, 17)
+    ; (0, 11), `Since (3, 21)
     ]
 ;;
 
@@ -172,6 +173,8 @@ module Theory = struct
     ; buildable : Buildable.t
     ; coqdep_flags : Ordered_set_lang.Unexpanded.t
     ; coqdoc_flags : Ordered_set_lang.Unexpanded.t
+    ; coqdoc_header : String_with_vars.t option
+    ; coqdoc_footer : String_with_vars.t option
     }
 
   let coq_public_decode =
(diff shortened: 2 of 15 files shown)


src/dune_rules/coq/coq_stanza.ml
```diff
@@ -259,7 +261,8 @@ module Theory = struct
          Ordered_set_lang.Unexpanded.field
            "coqdoc_flags"
            ~check:(Dune_lang.Syntax.since coq_syntax (0, 8))
-       in
+       and+ coqdoc_header = field_o "coqdoc_header" String_with_vars.decode
+       and+ coqdoc_footer = field_o "coqdoc_footer" String_with_vars.decode in
```
Alizter, line 265: Here is how to guard the field
```suggestion
       and+ coqdoc_header =
         field_o
           "coqdoc_header"
           (Dune_lang.Syntax.since coq_syntax (0, 10) >>> String_with_vars.decode)
```

Alizter: My editor borked the line numbers, but you just have to put the check with a `>>>` which is monadic `;` before the decoder inside the field.

Alizter:
While we are here, shall we add a `field_b_o` for `toc`. It should default to enabled, but we can allow it to be set like header and footer.

Alizter:
We should add a test where the header file comes as the target of another rule. We need to check that the dependencies are being correctly handled.

Variable expansion is also another one that should be checked, but I'm not certain what the best way to do that is yet.

src/dune_rules/coq/coq_stanza.ml
rlepigre: Thanks.

rlepigre:
> We should add a test where the header file comes as the target of another rule. We need to check that the dependencies are being correctly handled.
> 
> Variable expansion is also another one that should be checked, but I'm not certain what the best way to do that is yet.

@Alizter I added a test that should exercise both at once.

> While we are here, shall we add a `field_b_o` for `toc`. It should default to enabled, but we can allow it to be set like header and footer.

I'm not sure what you have in mind here, actually. Can't we just use the `(coqdoc_flags ...)` field for that?

Alizter:
> > While we are here, shall we add a `field_b_o` for `toc`. It should default to enabled, but we can allow it to be set like header and footer.
> 
> I'm not sure what you have in mind here, actually. Can't we just use the `(coqdoc_flags ...)` field for that?

we already set `--toc` by default. I thought it might be a good idea to introduce a field for that as well. It's not a priority however, so you can disregard that comment.

bin/printenv.ml
```diff
@@ -26,7 +26,9 @@ let dump sctx ~dir =
     Dune_rules.Menhir_rules.menhir_env ~dir
     |> Action_builder.of_memo
     >>= Dune_rules.Menhir_env.dump
-  and+ coq_dump = Dune_rules.Coq.Coq_rules.coq_env ~dir >>| Dune_rules.Coq.Coq_flags.dump
+  and+ coq_dump =
+    Dune_rules.Coq.Coq_rules.coq_env ~dir
+    >>| Dune_rules.Coq.Coq_flags.dump ~dir:(Path.build dir)
```
Alizter, line 31: I played around a bit with dropping the directory here, and I couldn't find a good way to do it. I'm not too worried about this.

doc/changes/11131.md
```diff
@@ -0,0 +1,2 @@
+- Add `coqdoc_header` and `coqdoc_footer` fields to the `coq` field of the `env` stanza, and to the `coq.theory` stanza, allowing to configure a custom header or footer respectively in the HTML output of `coqdoc`.
```
Alizter, line 1: I would suggest hard wrapping this at 80 chars.

rlepigre:
> Does it make sense to pass these flags when in latex mode? I think this might error out coqdoc. Should be simple to add a test for that.

The option is actually ignored (I had tested this), but I guess there is no point in introducing potential dependencies for the latex documentation. I pushed a fix.

bin/printenv.ml
rlepigre: Yeah, I wasn't sure what to do there.

Alizter:
ping @ejgallego this was ready for a while but seems to have been forgotten.

rlepigre:
This has been ready for a while, can we get it merged @Alizter @ejgallego?

Alizter:
@rlepigre  I can merge this once the tests have been fixed. The fix is to introduce version 0.11 of Coq lang.

rlepigre:
> @rlepigre I can merge this once the tests have been fixed. The fix is to introduce version 0.11 of Coq lang.

Great, thank @Alizter. I think it fixed things correctly, but I'm not sure about the whole version business. Is 3.21 the right lang version to require for the 0.11 coq lang version?

Alizter (approved):

Alizter:
Yes, the version seems fine. We will likely cut a release for 3.21 soon anyway.

rlepigre:
Great, thanks!

PR #11132, opened 2024-11-19
fix: patch expect tests on macos

gridbugs:
A recent update to macos's patch command broke some of our tests. The relevant changes to the command's behaviour are that it's stricter about line numbers in patch metadata, and it no longer deletes files, truncating them instead. This updates dune's patch expect tests to accomodate this behaviour.

### DIFF
--- a/test/expect-tests/dune_patch/dune_patch_tests.ml
+++ b/test/expect-tests/dune_patch/dune_patch_tests.ml
@@ -9,7 +9,7 @@ diff --git a/foo.ml b/foo.ml
 index b69a69a5a..ea988f6bd 100644
 --- a/foo.ml
 +++ b/foo.ml
-@@ -1,2 +1,2 @@
+@@ -1,1 +1,1 @@
 -This is wrong
 +This is right
 |}
@@ -22,7 +22,7 @@ diff --git a/dir/foo.ml b/dir/foo.ml
 index b69a69a5a..ea988f6bd 100644
 --- a/dir/foo.ml
 +++ b/dir/foo.ml
-@@ -1,2 +1,2 @@
+@@ -1,1 +1,1 @@
 -This is wrong
 +This is right
 |}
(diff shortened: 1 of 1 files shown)


test/expect-tests/dune_patch/dune_patch_tests.ml
```diff
@@ -199,10 +199,21 @@ let%expect_test "patching a new file" =
...
-  [%expect {|
-    File foo.ml not found |}]
+  let filename = "foo.ml" in
+  test [ filename, "This is wrong\n" ] ("foo.patch", delete_file);
+  (* Different implementations of the patch command behave differently when the
+     patch specifies deleting a file: *)
+  match (Unix.stat filename).st_kind with
+  | S_REG ->
+    (* Some implementations of patch (e.g. the patch that ships with macOS
+       15.1) do not delete files, and instread truncate them to 0 length. If
```
maiste, line 209:
```suggestion
       15.1) do not delete files, and instead truncate them to 0 length. If
```

maiste (approved):
Spot a typo but otherwise LGTM :+1:

test/expect-tests/dune_patch/dune_patch_tests.ml
```diff
@@ -199,10 +199,21 @@ let%expect_test "patching a new file" =
...
+       the file still exists after applying the patch, assert that it is now
+       empty. *)
+    assert (String.length (Io.String_path.read_file filename) == 0)
```
Leonidas-from-XIV, line 212: I think instead of reading the file I think it would be better to look at the `stat` metadata and read `st_size`. We already `stat` above and reading the size feels more direct than checking the length of the contents we read.

Leonidas-from-XIV (changes requested):
Thanks for the investigation. Looks like what macOS 15.1 did is a step back, given these patch features have been common patch extensions.

I agree with the approach, just a small comment on how to make the fix a bit more efficient and nicer.

test/expect-tests/dune_patch/dune_patch_tests.ml
gridbugs: Good idea. Done

Leonidas-from-XIV (approved):
Thanks, this looks good. I really like the comments explaining why the code is accepting empty files.

Leonidas-from-XIV:
Looks like github rolled back the CI runner to use macos 14.7, hence the tests succeed again at the moment but I expect everyone who tries to run them on 15.1 will have the same failures, so I'm merging it for future-proofing.

PR #11145, opened 2024-11-21
pkg: Print all the depexts

### LINKED ISSUE #11000: dune developer preview: automatic depext handling?
## Desired Behavior

I was trying out the Dune preview on OBuilder, and it has shown this message:
```
File "dune.lock/conf-sqlite3.pkg", line 4, characters 6-16:
4 |  (run pkg-config sqlite3))
          ^^^^^^^^^^
Error: Logs for package conf-sqlite3

You may want to verify the following depexts are installed:
- sqlite-devel
```

This is good (I installed the package, and tried again), but `opam` has this functionality built-in with `opam depext`.

It would be good if `dune pkg` had a similar feature: it can already print the depexts, so would be good if there was a command to either automatically install them, or query them all so that one can manually install them.

## Example

`dune pkg depext` should print a list of packages that need to be installed with the distro's package manager, such that `$distro_package_manager $(dune pkg depext)` would install all the missing depexts of the current dune project.

Alternatively `dune pkg depext --install` could also attempt to automatically invoke the appropriate package manager with `sudo`, but I'm not sure whether it is worth duplicating all that code from opam (could perhaps the `opam depext` code be moved to a package that both opam and dune share so that improvements to one are not lost in the other?

This can be tested using OBuilder's dune-project as a starting point https://github.com/ocurrent/obuilder/blob/v0.6.0/dune-project, which includes the dependency on sqlite3 (make sure you don't already have sqlite3 installed, e.g. by trying to run this in a fresh container).

moyodiallo:
Fixes https://github.com/ocaml/dune/issues/11000, this PR introduce `dune pkg depexts` which print all the depexts of a dune-project.


rgrinberg (changes requested):
We already have the printing functions under `dune show pkg`. I'd suggest you move this one there as well.

moyodiallo:
> We already have the printing functions under `dune show pkg`. I'd suggest you move this one there as well.

Great, I missed that.

moyodiallo:
> We already have the printing functions under `dune show pkg`. I'd suggest you move this one there as well.

Done here.

Leonidas-from-XIV:
I've moved the modules around and simplifed the test. It's fairly simple and doesn't need much setup :-)

rgrinberg (approved):

PR #11146, opened 2024-11-22
Add "tools install" commands

gridbugs:
Adds commands for installing dev tools without running them. These can be useful for setting up a dev environment before starting development.

### DIFF
--- a/bin/tools/ocamlformat.ml
+++ b/bin/tools/ocamlformat.ml
@@ -37,25 +37,47 @@ let build_dev_tool common =
      | Ok () -> ())
 ;;
 
-let term =
-  let+ builder = Common.Builder.term
-  and+ args = Arg.(value & pos_all string [] (info [] ~docv:"ARGS")) in
+let go builder command =
   let common, config = Common.init builder in
   Scheduler.go ~common ~config (fun () ->
     let open Fiber.O in
     let* () = Lock_dev_tool.lock_ocamlformat () |> Memo.run in
     let+ () = build_dev_tool common in
-    run_dev_tool (Common.root common) ~args)
+    match command with
+    | `Run_with_args args -> run_dev_tool (Common.root common) ~args
+    | `Install -> ())
 ;;
 
-let info =
-  let doc =
-    {|Wrapper for running ocamlformat intended to be run automatically
+module Exec = struct
+  let term =
+    let+ builder = Common.Builder.term
+    and+ args = Arg.(value & pos_all string [] (info [] ~docv:"ARGS")) in
+    go builder (`Run_with_args args)
+  ;;
+
+  let info =
+    let doc =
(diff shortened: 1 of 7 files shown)


bin/tools/ocamlformat.ml
```diff
@@ -37,25 +37,47 @@ let build_dev_tool common =
...
+let go builder command =
   let common, config = Common.init builder in
   Scheduler.go ~common ~config (fun () ->
     let open Fiber.O in
     let* () = Lock_dev_tool.lock_ocamlformat () |> Memo.run in
     let+ () = build_dev_tool common in
-    run_dev_tool (Common.root common) ~args)
+    match command with
+    | `Run_with_args args -> run_dev_tool (Common.root common) ~args
+    | `Install -> ())
```
Leonidas-from-XIV, line 48: It seems strange to me that there would be a function `go` that doesn't do anything in some cases. In that case I think it would make more sense to refactor it in some way where it is never called.

bin/tools/ocamllsp.ml
```diff
@@ -51,12 +55,36 @@ let term =
       let open Fiber.O in
       let* () = Lock_dev_tool.lock_ocamllsp () |> Memo.run in
       let+ () = build_ocamllsp common in
-      run_ocamllsp (Common.root common) ~args)
+      match command with
+      | `Run_with_args args -> run_ocamllsp (Common.root common) ~args
+      | `Install -> ())
```
Leonidas-from-XIV, line 60: Same concern as above.

Leonidas-from-XIV (commented):
As we have just introduced `@pkg-install` in #11046, I wonder whether it wouldn't be nicer to build the dev tools in the same manner, e.g. via a target. That way it would be more consistent for the user as there would be `@install`, `@pkg-install` and a hypothetical `@tools-install`.

avsm:
a high level observation: `install` seems like a very overloaded term in the dune CLI of late. There is:
- the good old `dune install` which does install the packages into the user's prefix outside of the workspace
- generation of `.install` files which declare what to install 
- this `tools install` which builds some tools, but doesn't install them by the same standard as the previous two uses of the term `install`
- a `@pkg-install` alias which doesn't install anything either, but makes dev available.

Would a different term from `install` be more appropriate for "fetch and build dev tools" aliases/commands?

rgrinberg:
> As we have just introduced `@pkg-install` in #11046, I wonder whether it wouldn't be nicer to build the dev tools in the same manner, e.g. via a target. That way it would be more consistent for the user as there would be `@install`, `@pkg-install` and a hypothetical `@tools-install`.

How about just `@dev-tools`?

avsm:
> How about just @dev-tools?

Sounds good to me!

gridbugs:
Thanks for the feedback. Closing this and will implement with a target instead.

PR #11147, opened 2024-11-22
fix: Allow any ref of a Git repo in the fragment

### LINKED ISSUE #11140: URLs should allow picking a branch or tag explicitely
## Desired Behavior

When I pin a repo, I want to be able to tell Dune explicitely whether the fragment in the URL is a tag or a branch.

## Example

```lisp
(pin
 (url "git+https://github.com/FStarLang/FStar.git#v0.9.5.0")
 (package
  (name fstar)))

(package (name repro) (depends fstar))
```

This is currently impossible because `v0.9.5.0` is both a branch and a tag in the repo and as a user I have no option to tell Dune which one it is. A possible workaround is to replace the tag with the commit hash, which is mostly workable, but with branches the commit the branch head points to is expected to change so it would require updating the `pin` all the time.

This ticket is inspired by #11113, which shows that repositories like this exist in the wild and are not just theoretical cases.

Leonidas-from-XIV:
This reverses the split of branches and tags from #11127, instead putting all possible references (e.g. also pull requests) in the same map and implements short-hands for looking up unqualified refs.

Fixes #11140

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -32,8 +32,7 @@ module Remote = struct
   type nonrec t =
     { url : string
     ; default_branch : Object.resolved option Fiber.t
-    ; branches : Object.resolved String.Map.t Fiber.t
-    ; tags : Object.resolved String.Map.t Fiber.t
+    ; refs : Object.resolved String.Map.t Fiber.t
     }
 
   let default_branch t = t.default_branch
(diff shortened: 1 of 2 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -32,8 +32,7 @@ module Remote = struct
   type nonrec t =
     { url : string
     ; default_branch : Object.resolved option Fiber.t
-    ; branches : Object.resolved String.Map.t Fiber.t
-    ; tags : Object.resolved String.Map.t Fiber.t
+    ; refs : Object.resolved String.Map.t Fiber.t
```
rgrinberg, line 35: Tbh, this doesn't seem like an improvement in readability. I think you could just as easily just parse the `refs/{heads,tags}` out of the query and then know which map you need to look in.

Opinions aside, I think this makes it a bit harder to implement the "did you mean" hint. That one would be rather useful when a user typos a branch or a tag. IMO you should hold off this refactoring until we have that useful hint ready.

rgrinberg (approved):
Approving as this PR fixes the issue reliably like it advertises. Not a fan of keeping map where everything is prefixed with refs/{heads,tags} - I would prefer the types to reflect the structure, and help improve the hints.

gridbugs (approved):

src/dune_pkg/rev_store.ml
Leonidas-from-XIV: The problem is with other types of refs like `pull` refs for heads and merges. The dune repo itself has at least 4 types of refs and we don't even do anything particularly unusual with Git.

Git refs are basically just a `Object.resolved String.Map.t` and the string is just `refs/{heads,tags}` by convention, so to me this seems like the best way to represent them aligned with how Git treats them, thus fewer concepts to keep in mind and less ways for repos to break our assumptions.

PR #11153, opened 2024-11-25
feat: ensure `@pkg-install` is not registered without lock file

maiste:
This PR improves the test and the documentation regarding the @pkg-install alias. It shows that it fails outside of packagement context and changes the documentation to reflect the true behavior.

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -1958,6 +1958,25 @@ let gen_rules context_name (pkg : Pkg.t) =
 
 module Gen_rules = Build_config.Gen_rules
 
+let pkg_alias_disabled =
+  Action_builder.fail
+    { fail =
+        (fun () ->
+          let error =
+            [ Pp.text "The @pkg-install alias cannot be used without a lock dir" ]
+          in
+          let hints =
+            [ Pp.concat
+                ~sep:Pp.space
+                [ Pp.text "You might want to create the lock dir with"
+                ; User_message.command "dune pkg lock"
+                ]
+            ]
+          in
+          User_error.raise ~hints error)
+    }
+;;
+
 let setup_pkg_install_alias =
   let build_packages_of_context ctx_name =
     (* Fetching the package target implies that we will also fetch the extra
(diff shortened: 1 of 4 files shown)


test/blackbox-tests/test-cases/pkg/alias-pkg-install.t
```diff
@@ -3,6 +3,21 @@ without building the project itself.
...
+  >  (name bar)
+  >  (allow_empty)
+  >  (depends foo))
+  > EOF
+
+Ensure the alias is not available outside of the package manamgent context:
+  $ dune build @pkg-install
+  Error: Alias "pkg-install" specified on the command line is empty.
+  It is not defined in . or any of its descendants.
+  [1]
```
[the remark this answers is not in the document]
maiste: I think adding an error instead of returning the Memo.t vould do the trick.

Leonidas-from-XIV:
I think you need the `Memo.t`, otherwise constructing the rules fails. Maybe raise the failure inside the memo?

src/dune_rules/pkg_rules.ml
```diff
@@ -1972,7 +1972,14 @@ let setup_pkg_install_alias =
...
         >>= (function
-         | false -> Memo.return Rules.empty
+         | false ->
+           let error =
+             [ Pp.text "The @pkg-install alias can't be used without a lock dir" ]
+           in
+           let hints =
+             [ Pp.text "You might want to create the lock dir with dune pkg lock" ]
+           in
+           User_error.raise ~hints error
```
maiste, line 1982: @rgrinberg would you have a suggestion on how it would be possible to raise an error when the alias is called?

If I raised an error in a `Memo.t` or in a `Rules.collect_unit`, it would simply fail when doing the check or collecting the rules. The solution I see is to make an alias which would trigger an action that would fail. I'm wondering: is there a better way to do this?

Leonidas-from-XIV: Another alternative is to just return an action that does nothing, like an `Progn []`. Though that might make users assume `@pkg-install` did a thing when in fact it didn't do anything.

rgrinberg: You can use `Action_builder.fail` to delay your error

rgrinberg (approved):

PR #11164, opened 2024-11-27
[3.17] remove breaking change for `alpha1`

maiste:
This PR reverts the patch made in #10892 for the `3.17.0~alpha1` release. Indeed, while releasing `3.17.0` we have seen that it would break `odoc` tests because of the indentation. Further discussions have occurred in the PR about how to fix it in `main` in a more sustainable way. However, we have decided to remove the patch for `3.17.0` and reintroduce it in later released.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -41,9 +41,6 @@
 - Fix an issue where C stubs would be rebuilt whenever the stderr of Dune was
   redirected. (#10883, fixes #10882, @nojb)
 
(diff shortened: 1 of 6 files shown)


Leonidas-from-XIV (approved):
Approving this as this is in agreement with the original author of the feature. @nojb. I also think that it's better to defer this to 3.18 with the improvements discussed in #10892 as that should avoid the issue altogether and not require users patching their builds.

ElectreAAS (approved):

PR #11173, opened 2024-12-01
Stop VCS discovery choking on unreadable root

KoviRobi:
On Android devices, the root directory is not readable. This stops the VCS discovery failing the whole install by just ignoring that for VCS discovery, in the same manner as for root.

In my case, a dune package was not in a VCS folder because it was being built by `nix`, which for reproducibility doesn't include the `.git` folder in the build environment (because even though the commit contents are the same, the git database isn't, e.g. it can contain different tags etc).


rgrinberg (approved):

rgrinberg:
Merged with a CHANGES entry

rgrinberg:
Merged with a CHANGES entry.

PR #11176, opened 2024-12-02
`(no_dynlink)`: do not build `.cmxs`

nojb:
What it says on the tin: when the library declares `(no_dynlink)`, then the `.cmxs` should not be built. On large codebases, this can save considerable time (especially on Windows).

### DIFF
--- a/test/blackbox-tests/test-cases/no_dynlink.t
+++ b/test/blackbox-tests/test-cases/no_dynlink.t
@@ -30,3 +30,33 @@ Now with `(no_dynlink)`.
   Error: Don't know how to build _build/default/mylib.cmxs
   Hint: did you mean _build/default/mylib.cma or _build/default/mylib.cmxa?
   [1]
+
+Next, we check that the .cmxs is installed without `(no_dynlink)`:
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.17)
+  > (package (name mylib))
+  > EOF
+
+  $ cat >dune <<EOF
+  > (library
+  >  (public_name mylib))
+  > EOF
+
+  $ dune build _build/default/mylib.install
+
+  $ grep cmxs _build/default/mylib.install
+    "_build/install/default/lib/mylib/mylib.cmxs"
+
+And *not* installed with `(no_dynlink)`:
+
+  $ cat >dune <<EOF
+  > (library
+  >  (public_name mylib)
+  >  (no_dynlink))
+  > EOF
+
+  $ dune build _build/default/mylib.install
+
+  $ grep cmxs _build/default/mylib.install
+  [1]
--- a/test/blackbox-tests/test-cases/no_dynlink.t
+++ b/test/blackbox-tests/test-cases/no_dynlink.t
@@ -1,6 +1,11 @@
+This test checks that if a library is declared with `(no_dynlink)`, then the
+corresponding `.cmxs` file is *not* built.
+
   $ cat >dune-project <<EOF
   > (lang dune 3.17)
   > EOF
+
+First we check the behaviour when `(no_dynlink)` is not present.
 
   $ cat >dune <<EOF
   > (library
@@ -9,15 +14,9 @@
 
   $ touch a.ml
 
-  $ dune build --display short
-        ocamlc .mylib.objs/byte/mylib.{cmi,cmo,cmt}
-      ocamldep .mylib.objs/mylib__A.impl.d
-      ocamlopt .mylib.objs/native/mylib.{cmx,o}
-        ocamlc .mylib.objs/byte/mylib__A.{cmi,cmo,cmt}
-      ocamlopt .mylib.objs/native/mylib__A.{cmx,o}
-        ocamlc mylib.cma
-      ocamlopt mylib.{a,cmxa}
-      ocamlopt mylib.cmxs
+  $ dune build _build/default/mylib.cmxs
+
+Now with `(no_dynlink)`.
 
   $ cat >dune <<EOF
   > (library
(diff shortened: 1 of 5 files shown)


nojb:
@rgrinberg: do you think this change would need to be versioned?

test/blackbox-tests/test-cases/no_dynlink.t
```diff
@@ -0,0 +1,37 @@
...
+  > EOF
+
+  $ cat >dune <<EOF
+  > (library
+  >  (name mylib))
+  > EOF
+
+  $ touch a.ml
+
+  $ dune build --display short
```
rgrinberg, line 12: Could you replace the --display short with just an explicit target such as
mylib.cmxs?

This extra output makes the tests fragile and hard to understand

test/blackbox-tests/test-cases/no_dynlink.t
```diff
@@ -0,0 +1,37 @@
+  $ cat >dune-project <<EOF
```
rgrinberg, line 1: Can you add a few sentences here to describe the test?

rgrinberg:
It does sound fine to not version this. The old behavior was useless, so I
don't see much reason to preserve it.

Could you add a CHANGES entry?

test/blackbox-tests/test-cases/no_dynlink.t
nojb: Fixed, thanks.

test/blackbox-tests/test-cases/no_dynlink.t
nojb: Fixed, thanks.

nojb:
> Could you add a CHANGES entry?

It already has one :)

nojb:
Thanks for the review @rgrinberg! I think all points are addressed.

anmonteiro:
does this still build & install the `.cmxs` when building for `-p ..` and `dune install`?

anmonteiro:
I disagree: before this PR, a `(public_name my_library)` installs a `my_lybrary.cmxs`. 

after this PR, a `(public_name my_library) (no_dynlink)` doesn't install `my_library.cmxs`. 

it's technically a breaking change if both points below are true at the same time:
- a library adds `(no_dynlink)` and releases to opam
- another library was using `Dynlink.loadfile "that_cmxs_file.cmxs"`.

nojb:
> I disagree: before this PR, a `(public_name my_library)` installs a `my_lybrary.cmxs`.

I cannot reproduce this behaviour with Dune 3.16:
```
$ mkdir no_dynlink
$ cd no_dynlink
$ cat >dune-project <<EOF
> (lang dune 3.16)
> (package (name foo))
> EOF
$ cat >dune <<EOF
> (library (public_name foo) (no_dynlink))
> EOF
$ dune --version
3.16.0
$ dune build
$ find _build/install/default/
_build/install/default/
_build/install/default/lib
_build/install/default/lib/foo
_build/install/default/lib/foo/foo.cma
_build/install/default/lib/foo/META
_build/install/default/lib/foo/foo.cmt
_build/install/default/lib/foo/foo.a
_build/install/default/lib/foo/foo.cmi
_build/install/default/lib/foo/foo.ml
_build/install/default/lib/foo/dune-package
_build/install/default/lib/foo/foo.cmxa
_build/install/default/lib/foo/foo.cmx
$
```

Leonidas-from-XIV:
Also I think that this should have some documentation, since bar reading the changelog there is no way to learn of this feature.

Leonidas-from-XIV:
@nojb I can confirm that Dune 3.16 has the behavior you describe (also I rebased the branch to get the fix for macOS CI failures in).

nojb:
> @nojb I can confirm that Dune 3.16 has the behavior you describe (also I rebased the branch to get the fix for macOS CI failures in).

Thanks!

anmonteiro:
I understand now that dune 3.16 didn't install the cmxs in the first place. that means i'm wrong, please proceed.

anmonteiro:
but something _has_ changed between 3.16 and this PR where a `.cmxs` is installed with dune 3.17. I'm not sure if that's expected but i didn't find it in the changelog

rgrinberg (approved):

test/blackbox-tests/test-cases/no_dynlink.t
```diff
@@ -0,0 +1,32 @@
...
+
+  $ cat >dune <<EOF
+  > (library
+  >  (name mylib)
+  >  (no_dynlink))
+  > EOF
+
+  $ dune clean
+
+  $ dune build _build/default/mylib.cmxs
```
rgrinberg, line 29: Could you add some tests for the .install file as well?

rgrinberg:
> I disagree: before this PR, a `(public_name my_library)` installs a `my_lybrary.cmxs`.
> 
> after this PR, a `(public_name my_library) (no_dynlink)` doesn't install `my_library.cmxs`.
> 
> it's technically a breaking change if these 2 points are true at the same time:
> 
> * a library adds `(no_dynlink)` and releases to opam
> * another library was using `Dynlink.loadfile "that_cmxs_file.cmxs"`.
> 
> note: no strong opinions whether you want to support this (arguable) edge case. i realize that this will help your builds internally and is indeed an edge case (and a new field) so i'm not opposed to it.

To re-iterate nojb, I don't see how this PR changes installation. However, this PR is indeed breaking users that relied on building a cmxs but not installing it. I think such an edge case is useless enough that we can let it slide.

rgrinberg:
> I understand now that dune 3.16 didn't install the cmxs in the first place. that means i'm wrong, please proceed.

That is correct. With this PR, the cmxs won't even be built.

test/blackbox-tests/test-cases/no_dynlink.t
nojb: Good idea. Fixed, thanks.

nojb:
All issues having been addressed, planning to merge once CI passes.

PR #11181, opened 2024-12-03
Create reference documentation for current state of package management support

Leonidas-from-XIV:
For a proper release it makes sense to document the package management extensions to the `dune-project` and `dune-workspace`. This PR documents the current state, though it is to be discussed if this is what we plan to roll out as final form of the feature.

### DIFF
--- a/doc/reference/dune-project/pin.rst
+++ b/doc/reference/dune-project/pin.rst
@@ -1,0 +1,50 @@
+pin
+---
+
+.. warning::
+
+   :doc:`Dune Package Management </explanation/package-management>` is not
+   final yet and the configuration options are subject to change.
+
+Pins are package overrides used in the context of package management. They
+allow to fix a package at a specific version which is not affected by the
+package repositories selected.
+
+.. describe:: (pin ...)
+
+   .. versionadded:: 3.14
+
+   Define a package override.
+
+   .. describe:: (url <string>)
+
+      The URL of the package source.
+
+      This can be a path to a directory on the local file system or remote Git
+      repository. Local paths can be absolute or relative, and may optionally
+      begin with ``file://`` though this is not necessary. Remote Git
+      repository URLs must begin with ``git+``, for example
+      ``git+https://github.com/user/repo`` or
+      ``git+git@github.com:user/repo.git``.
+
+      This must be specified.
+
+   .. describe:: (package ...)
+
+      Defines which package is to be pinned.
+
+      This must be specified.
+
+      .. describe:: (name <string>)
+
+         The name of the package.
+
+         This must be specified.
+
+      .. describe:: (version <string>)
+
+         The version that the package should be assumed to be. Defaults to
+         ``dev`` if unspecified.
+
+.. seealso:: :doc:`pin stanza in dune-workspace </reference/dune-workspace/pin>` for
+   workspace-wide pinning.
--- a/doc/reference/dune-workspace/context.rst
+++ b/doc/reference/dune-workspace/context.rst
@@ -15,6 +15,12 @@ the description of an opam switch, as follows:
 -  ``(name <name>)`` is the subdirectory's name for ``_build``, where this
    build's context artifacts will be stored.
 
+-  ``(lock_dir <path>)`` specifies the lock directory that will be used for
+   building this context (if any). If no lock directory is specified
+   ``dune.lock`` will be used. See the
+   :doc:`/reference/dune-workspace/lock_dir` stanza for lock directory
+   configuration options.
+
 -  ``(root <opam-root>)`` is the opam root. By default, it will take the opam
    root defined by the environment in which ``dune`` is run, which is usually
    ``~/.opam``.
--- a/doc/reference/dune-workspace/lock_dir.rst
+++ b/doc/reference/dune-workspace/lock_dir.rst
@@ -1,0 +1,73 @@
+lock_dir
+========
+
+.. warning::
+
+   :doc:`Dune Package Management </explanation/package-management>` is not
+   final yet and the configuration options are subject to change.
+
+This stanza configures the lock directory settings for the current workspace.
+For the default workflow no configuration is necessary, but the defaults can be
+changed if desired.
+
+.. describe:: (lock_dir ...)
+
+   .. versionadded:: 3.13
+
+   Configures a specific lock directory to be created or used.
+
+   .. describe:: (path <string>)
+
+      The location in the source tree where the lock directory will be
+      created or read from. If not specified defaults to ``dune.lock``.
+
+   .. describe:: (repositories <name list>)
+
+      The repositories to be used for finding a package solution, specified
+      in priority order. Supports ``:default`` which contains ``upsteam`` and
+      ``overlay``.
+
+      Additional repositories can be defined using the
+      :doc:`/reference/dune-workspace/repository` stanza.
+
+   .. describe:: (solver_env ...)
+
+      The environment that is injected into the solver when creating the lock
+      directory.
+
+      It consists of a sequence of ``(<name> <value>)`` pairs.
+
+   .. describe:: (unset_variables <name list>)
+
+      A list of variables that are used in solving that are deliberately unset
+      even if the solver could provide bindings for them.
+
+      The variables here cannot overlap with those defined in ``solver_env``.
+
+   .. describe:: (pins <name list>)
+
+      .. versionadded:: 3.15
+
+      Define which pins are enabled for this particular lock dir. See
+      :doc:`/reference/dune-workspace/pin` for details on how to define pins.
+
+   .. describe:: (version_preference <string>)
+
+      Can be one of:
+
+      - ``newest`` (default): The solver will pick the newest available
+         version of a package that satisfies the constraints.
+      - ``oldest``: The solver will pick the lowest version that will satisfy
+         the constraints
+
+   .. describe:: (constraints <dep-specification>)
+
+      Adds additional solver constraints that are passed to the solver. Follows
+      the :token:`~pkg-dep:dep_specification` format.
+
+      .. note::
+
+         Names introduced through ``constraints`` are not considered
+         dependencies and not added to the lockfile. They exist solely to add
+         additional constraints if the packages to which the constraint is
+         applied are selected and don't do anything otherwise.
--- a/doc/reference/dune-workspace/pin.rst
+++ b/doc/reference/dune-workspace/pin.rst
@@ -1,0 +1,57 @@
+pin
+===
+
+.. warning::
+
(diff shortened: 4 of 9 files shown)


Leonidas-from-XIV:
@rgrinberg While writing the documentation I had some observations:

  * Both `pin` stanzas use `url` while `repository` uses `source`. I think it would be nicer to just always refer to URLs via `(url ..)`.
  * `(package (name <foo>))` is cumbersome. How about adding a shorthand of `(package <foo>)`? If people want a version they could still use `(name <foo>) (version <bar>)` or even `(package <foo>.<bar>)` using the common OPAM convention, which cuts down on the nesting.
  * `pins` in `dune-workspace` are named `sources` internally, since that's what they were named originally. Maybe worth renaming it internally too, to avoid confusion where the dune-lang calls it something but then the code calls it something else internally.

What do you think?

rgrinberg:
> (package (name <foo>)) is cumbersome. How about adding a shorthand of (package <foo>)? If people want a version they could still use (name <foo>) (version <bar>) or even (package <foo>.<bar>) using the common OPAM convention, which cuts down on the nesting.

Sympathetic to the problem, but I'm not entirely convinced. We have a similar shorthand for the `rule` stanza, and my experience is that it hasn't pulled its own weight. One problem with such shorthands, is that they really complicate the life for tooling to be built on top of dune files. With enough such shortcuts, one would not be able to write anything without using a library that understands these tricks.

If there's enough demands from users with some real world examples of long pins, there would be a much stronger case for this.

doc/reference/dune-project/pin.rst
```diff
@@ -0,0 +1,43 @@
...
+      .. describe:: (name <string>)
+
+         The name of the package.
+
+         This must be specified.
+
+      .. describe:: (version <string>)
+
+         The version that the package should be assumed to be.
+
```
gridbugs, line 41:
```suggestion
This is optional, and its default value is "dev".
```

doc/reference/dune-project/pin.rst
```diff
@@ -0,0 +1,43 @@
...
+.. describe:: (pin ...)
+
+   .. versionadded:: 3.14
+
+   Define a package override.
+
+   .. describe:: (url <string>)
+
+      The URL of the package source.
+
```
gridbugs, line 23:
```suggestion
This can be a path to a directory on the local file system or remote git repository. Local paths can be absolute or relative, and may optionally begin with `file://` though this is not necessary. Remote git repository urls must begin with `git+https://`, for example `git+https://github.com/user/repo`.
```

doc/reference/dune-workspace/context.rst
```diff
@@ -15,6 +15,11 @@ the description of an opam switch, as follows:
 -  ``(name <name>)`` is the subdirectory's name for ``_build``, where this
    build's context artifacts will be stored.
 
+-  ``(lock_dir <path>)`` specifies the lock directory that will be used for
+   building this context (if any). See the
```
gridbugs, line 19:
```suggestion
   building this context. This is optional and its default value is `dune.lock`. See the
```

doc/reference/dune-workspace/lock_dir.rst
```diff
@@ -0,0 +1,72 @@
...
+.. describe:: (lock_dir ...)
+
+   .. versionadded:: 3.13
+
+   Configures a specific lock directory to be created or used.
+
+   .. describe:: (path <string>)
+
+      The location in the source tree where the lock directory will be
+      created or read from.
```
gridbugs, line 21:
```suggestion
      created or read from. This is optional and its default value is "dune.lock".
```

gridbugs (approved):
Looks good. I added suggestions for a couple of clarifications and examples.

doc/reference/dune-project/pin.rst
```diff
@@ -0,0 +1,43 @@
+pin
+---
+
+.. warning::
+
+   Dune Package Management is not final yet and the configuration options
+   are subject to change.
```
maiste, line 7: :art: I would add a pointer to the explanation about package management you wrote. But this is just my trying to create a graph in the documentation :smile:

doc/reference/dune-project/pin.rst
```diff
@@ -0,0 +1,43 @@
...
+
+      .. describe:: (name <string>)
+
+         The name of the package.
+
+         This must be specified.
+
+      .. describe:: (version <string>)
+
+         The version that the package should be assumed to be.
```
maiste, line 40: Do we have a convention for `required` vs `optional` in the documentation? If not, I think we should think about some uniformity (not for this PR but keep it in a corner of our head).

:art:  I would suggest emphasising this, otherwise.
```suggestion
      This must **be specified**.

   .. describe:: (package ...)

      Defines which package is to be pinned.

      This must **be specified**.

      .. describe:: (name <string>)

         The name of the package.

         This must **be specified**.

      .. describe:: (version <string>)

         The version that the package should be assumed to be.
```

doc/reference/dune-workspace/lock_dir.rst
```diff
@@ -0,0 +1,72 @@
+lock_dir
+========
+
+.. warning::
+
+   Dune Package Management is not final yet and the configuration options
+   are subject to change.
+
+This stanza configures settings for lock directories what are relevant to
```
maiste, line 9:
```suggestion
This stanza configures settings for lock directories who are relevant to
```
Otherwise, I'm not sure to get the meaning of the sentence :sweat: 

Shouldn't we mention that this is related to package management (apart from the warning)?

doc/reference/dune-workspace/lock_dir.rst
maiste, line 1: Shouldn't this file specified `required` / `be specified` fields?

doc/reference/dune-workspace/pin.rst
```diff
@@ -0,0 +1,46 @@
+pin
+===
+
+.. warning::
+
+   Dune Package Management is not final yet and the configuration options
+   are subject to change.
+
+This stanza is used to define additional package sources.
```
maiste, line 9: :art: I would suggest the same as the other two: link to package management and a description that this is related to package management apart from within the warning.

Again, feel free to ignore if you don't think this is necessary :smile:

doc/reference/dune-workspace/pin.rst
```diff
@@ -0,0 +1,46 @@
...
+
+.. note::
+
+   Defining a pin does not enable it by default. It needs to be enabled in a
```
maiste, line 13: Shouldn't it be more explicit that it needs to be in `(pins ...)` (IIUC)?

doc/reference/dune-workspace/pin.rst
```diff
@@ -0,0 +1,46 @@
...
+
+   .. describe:: (package ...)
+
+      Specifies the the packages to assign this pin to.
+
+      .. describe:: (name <string>)
+
+         The name of the package.
+
+         This must be specified.
```
maiste, line 39:
```suggestion
         This must **be specified**.
```
:art:  Same question as the other part of the documentation.

doc/reference/dune-workspace/repository.rst
```diff
@@ -0,0 +1,37 @@
+repository
+==========
+
+.. warning::
+
+   Dune Package Management is not final yet and the configuration options
+   are subject to change.
+
+This stanza defines a new named repository and attaches a source location to it.
```
maiste, line 9: :art:  Same as the others: link and explicit about the context?

doc/reference/dune-workspace/repository.rst
```diff
@@ -0,0 +1,37 @@
...
+
+.. note::
+
+   Defining a repository does not enable it in project by default. It needs to be
+   enabled in a lock directory using the :doc:`/reference/dune-workspace/lock_dir`
```
maiste, line 14: I would also point the `(repositories ...)` stanza here to clarify which one we are talking about.

doc/reference/dune-workspace/repository.rst
```diff
@@ -0,0 +1,37 @@
...
+
+   .. versionadded:: 3.12
+
+   Defines a named package repository.
+
+   .. describe:: (name <string>)
+
+      The name used to refer to the respository. Names have to be unique.
+
+      This must be specified.
```
maiste, line 27: :art: 
```suggestion
      This must **be specified**.
```

doc/reference/dune-workspace/repository.rst
```diff
@@ -0,0 +1,37 @@
...
+
+   .. describe:: (url <string>)
+
+      The location from which the repository will be loaded.
+
+      Both HTTP and Git locations can be specified, the latter allowing for
+      extensive control of the version by specifying an exact revision, tag or
+      branch.
+
+      This must be specified.
```
maiste, line 37: :art: 
```suggestion
      This must **be specified.**
```

maiste (commented):
Many remarks are some duplicates in case you want to change them across the documentation. I have used the :art: to mentioned "cosmetic" changes and in case you want to ignore them.

doc/reference/dune-project/pin.rst
Leonidas-from-XIV: > Do we have a convention for required vs optional in the documentation?

I think in most places it only mentions that a field is required and all fields are optional otherwise. (`rg required doc` vs `rg optional doc`) thus for consistency I would just write that fields are required where they are and nothing otherwise.

doc/reference/dune-workspace/lock_dir.rst
Leonidas-from-XIV: None of them are required and it seems the documentation points out if fields are required (which sort of makes sense, most fields are optional and many people don't know all the options of all stanzas).

doc/reference/dune-workspace/pin.rst
Leonidas-from-XIV: I've changed it a bit to mention that :+1:

doc/reference/dune-workspace/repository.rst
Leonidas-from-XIV: I don't think I can link all that exactly but also `lock_dir` doesn't have that many options :-)

doc/reference/dune-workspace/pin.rst
Leonidas-from-XIV: I think it is fairly obvious, the stanza is not that large so I would defer if people find it unclear.

Leonidas-from-XIV:
@gridbugs Thanks, I've incorporated a lot of your changes with some rewording. It's definitely good to mention the default values.

I think I've incorporated most of the comments in a way where it is consistent with the rest of the documentation so unless someone objects I'd merge it (despite the pointless coverage CI failure).

maiste (approved):

PR #11183, opened 2024-12-05
Rename `source` to `url` in `repository` stanza

Leonidas-from-XIV:
When working on #11181 I came across a few issues, so I thought I'll change it and get input on the change in a PR. This PR aligns the naming of `repository` and `pin`, where pin uses `url` and repository used `source` for a field with the same semantics.

Personally I don't have a strong preference for `url` over `source`, I could also reverse the rename and change `pin` instead.

Going through all the failing tests had me improve the helpers a little bit so I could remove a bit of boilerplate where `dune-workspace` is created just do inject the mock repo URL.

### DIFF
--- a/doc/tutorials/dune-package-management/repos/dune-workspace
+++ b/doc/tutorials/dune-package-management/repos/dune-workspace
@@ -5,4 +5,4 @@
 
 (repository
  (name specific-upstream)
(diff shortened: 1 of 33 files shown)


rgrinberg (approved):
This is indeed better. Though I think you should move the unrelated changes to a separate PR.

gridbugs (approved):

Leonidas-from-XIV:
I've split the test changes out in #11185, removed them from here and merged.

PR #11189, opened 2024-12-09
[3.17] `(no_dynlink)`: do not build `.cmxs` (#11176)

nojb:
Cherry-pick of #11176 to 3.17. cc @maiste


Leonidas-from-XIV:
The failure is on macOS CI is fixed via #11179.

maiste (approved):
I have also used this PR to backport the fix that makes the CI happy on MacOS. The review already being done in the past and the rereview seems good. 
LGTM :+1:

PR #11192, opened 2024-12-11
flake: add overlay

asymmetric:
So that consumers can more simply use the packages exported by this flake, by adding the overlay when they instantiate nixpkgs.

I think a cleaner way (which would avoid the ugly `//` at the end) would be to get rid of flake-utils entirely, taking inspiration from (`rust-overlay`](https://github.com/oxalica/rust-overlay/blob/master/flake.nix).

But that is a bigger change.


anmonteiro (commented):
I think this sounds reasonable. I'm just unsure about adding `dune` to `pkgs.dune` as the overlay.

asymmetric:
@anmonteiro would you prefer it were called `dune-unstable` or something like that? And likewise, should the `packages` outputs be called `dune-unstable-xyz` instead of the current `dune-xyz`?

anmonteiro:
I'm not sure. I'm happy to try it out like this until anyone complains.

anmonteiro (approved):

flake.nix
```diff
@@ -46,18 +46,11 @@
             configureFlags = [ "-native-compiler" "yes" ];
           });
         })
+        self.overlays.default
```
anmonteiro, line 49: i think this is not needed, right?
```suggestion
```

anmonteiro, line 49: not the biggest fan of applying the own overlay when declaring `pkgs`. can we do it another way? maybe a common function that both the overlays and `outputs` call.

rgrinberg:
@anmonteiro please take over and rebase/merge this PR if you still think its relevant

PR #11194, opened 2024-12-12
`Io.copy_channels`: make it reentrant

### LINKED ISSUE #11193: [BUG] `Io.copy_channels` is called concurrently under Windows (at least), resulting in corrupted output
Recently at LexiFi we observed what looks like a bug in Dune. When building a medium-size workspace under Windows, sometimes the target of `copy_files#` will get corrupted. Concretely, the workspace includes Menhir, and the file `src/stage2/lexmli.ml` (which is an `ocamllex`-generated lexer and is copied using `copy_files#` from `src/lexmli.ml`) is sometimes corrupted:

image

For future reference, the symptom is that the compiler complained from an illegal backslash in this file:

image

Looking at the code `(copy_files# ...)` is implemented using `Io.copy_channels` from `stdune` which uses a global buffer:

https://github.com/ocaml/dune/blob/25be1eac55f6336f8f4ecd5f90d180467be9b26d/otherlibs/stdune/src/io.ml#L68-L79

Accordingly, the first suspicion is that this function is called concurrently. To test this conjecture I instrumented the code:

```diff
diff --git a/otherlibs/stdune/src/io.ml b/otherlibs/stdune/src/io.ml
index 576021ffb..0dbaae703 100644
--- a/otherlibs/stdune/src/io.ml
+++ b/otherlibs/stdune/src/io.ml
@@ -68,6 +68,7 @@ let input_zero_separated =
 let copy_channels =
   let buf_len = 65536 in
   let buf = Bytes.create buf_len in
+  let busy = ref false in
   let rec loop ic oc =
     match input ic buf 0 buf_len with
     | 0 -> ()
@@ -75,7 +76,14 @@ let copy_channels =
       output oc buf 0 n;
       loop ic oc
   in
-  loop
+  fun ic oc ->
+  if !busy then begin
+    Printf.printf "XXXXXXXX\n%!";
+    Printexc.print_raw_backtrace stdout (Printexc.get_callstack 20)
+  end;
+  busy := true;
+  loop ic oc;
+  busy := false
 ;;
```

 and indeed, the function is called concurrently:

image

@rgrinberg: should we do the safe thing here and use a separate buffer for each call? (We could optimize that by using a global buffer but if the function is called again while the global buffer is being used, then use a fresh buffer.) What do you think?

nojb:
For discussion.

Fixes #11193

### DIFF
--- a/doc/changes/11194.md
+++ b/doc/changes/11194.md
@@ -1,0 +1,3 @@
+- Fix bug that could result in corrupted file copies by Dune, for example when
+  using the `copy_files#` stanza or the `copy#` action. (@nojb, #11194, fixes
+  #11193)
--- a/otherlibs/stdune/src/io.ml
+++ b/otherlibs/stdune/src/io.ml
@@ -67,15 +67,25 @@ let input_zero_separated =
 
 let copy_channels =
   let buf_len = 65536 in
-  let buf = Bytes.create buf_len in
-  let rec loop ic oc =
+  let global_buf = Bytes.create buf_len in
+  let rec loop buf ic oc =
     match input ic buf 0 buf_len with
     | 0 -> ()
     | n ->
       output oc buf 0 n;
-      loop ic oc
+      loop buf ic oc
   in
-  loop
+  let busy = ref false in
+  fun ic oc ->
+    if !busy
+    then loop (Bytes.create buf_len) ic oc
+    else (
+      busy := true;
+      match loop global_buf ic oc with
+      | () -> busy := false
+      | exception exn ->
+        busy := false;
+        Exn.reraise exn)
 ;;
 
 let setup_copy ?(chmod = Fun.id) ~src ~dst () =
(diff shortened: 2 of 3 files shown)


nojb:
> Trying to understand the reason for this issue. Why is it using a global buffer, is the allocation of the buffer that expensive that reusing it is important?

The allocation of a single buffer is of course inexpensive. But Dune will routinely copy large numbers of files around (tens of thousands in a large codebase); a 65K buffer for each one will add up and put pressure on the GC, slowing Dune down (this is just my intuition).

rgrinberg (approved):
LGTM. Worth mentioning this bug fix in the CHANGES.

You might want to look at `CopyFileEx` to speed up copying on windows as well.

nojb:
> LGTM. Worth mentioning this bug fix in the CHANGES.

Done, thanks.

> You might want to look at `CopyFileEx` to speed up copying on windows as well.

Yes, I looked at it briefly. But note that this function is used in a context where we are not doing a plain file copy, as we are adding the line directive as well (in fact, I think the code in question is used even under Linux & Darwin, but maybe the bug is harder to trigger there due to faster file operations).

Another approach suggested by a colleague: memory map the source file (and/or the output file) to avoid having to keep an explicit buffer between the reading and the writing. But this would need to be considered separately as it is a more invasive change.

PR #11199, opened 2024-12-12
chore: Promote change of `unix; threads` -> `threads.posix`

Leonidas-from-XIV:
When calling `dune build` it promotes this change (at least on my machine). I don't know why this is happening or whether this is correct. From my cursory understandin `threads` will always pick `threads.posix` as `threads.vm` is deprecated but I don't know which change on caused this and whether this is the right thing to do.

Thus I am opening this PR as a to open a discussion but I am not sure if the proposed solution is the correct one.

### DIFF
--- a/boot/libs.ml
+++ b/boot/libs.ml
@@ -1,4 +1,4 @@
-let external_libraries = [ "unix"; "threads" ]
+let external_libraries = [ "threads.posix" ]
 
 let local_libraries =
   [ ("otherlibs/ordering", Some "Ordering", false, None)


Leonidas-from-XIV:
Hmm, clearly CI disagrees that this change makes sense. Strangely I do have `OCaml 5.1.1` just like the CI actions that fail and I've seen this change happen for @Julow in his PR as well.

maiste:
@Leonidas-from-XIV as it seems to break, shouldn't we close this PR?

Leonidas-from-XIV:
Yeah, apparently this shouldn't be promoted. I kinda wonder why it happens on my system.

gridbugs:
This happens reliably for me too which is annoying but not sure how to fix.

rgrinberg:
Maybe we should just stop generating this list? We're aways going to depend on exactly these two "external" libraries anyway. We could just hard code them in the bootstrap script and be done with it.

Issue #11200, opened 2024-12-12
`dune subst` no longer works on empty directories in ocaml-ci without  Dune 3.17 


shonfeder:
## Expected Behavior

`dune subst` (invovked by opam) continues to work on empty directories of vendored projects for pinned dependencies, as it did up until 3.16

## Actual Behavior

Failes with 

```
### output ###
# File ".", line 1, characters 0-0:
# Error: There is no dune-project file in the current directory, please add one
# with a (name <name>) field in it.
# Hint: 'dune subst' must be executed from the root of the project.
```

## Reproduction

This is not at all minimal, but is a starter, taken from an ocaml-ci job (where this problem has been encountered):

```
git clone --recursive "https://github.com/ocurrent/ocaml-ci.git" -b "master" && cd "ocaml-ci" && git reset --hard bc5c5e55
cat > Dockerfile <<'END-OF-DOCKERFILE'
FROM ocaml/opam:ubuntu-22.04-ocaml-4.14@sha256:1c5bf3c2bec3ee454185e5155264394ba79f70e5c3985b66f7bd57810d9cbebb
# ubuntu-22.04-4.14_opam-2.3
USER 1000:1000
ENV CLICOLOR_FORCE="1"
ENV OPAMCOLOR="always"
WORKDIR /src
RUN sudo ln -f /usr/bin/opam-2.3 /usr/bin/opam
RUN opam init --reinit -ni
RUN uname -rs && opam exec -- ocaml -version && opam --version
WORKDIR /src
RUN sudo chown opam /src
RUN cd ~/opam-repository && (git cat-file -e 9d811310def03190e9b9b1db9c9ea19b4f954c17 || git fetch origin master) && git reset -q --hard 9d811310def03190e9b9b1db9c9ea19b4f954c17 && git log --no-decorate -n1 --oneline && opam update -u
RUN mkdir -p 'solver-service/' 'ocurrent/' 'ocluster/' 'ocluster/obuilder/' 'ocluster/' 'ocaml-version/' 'ocaml-dockerfile/'
COPY --chown=1000:1000 solver-service/solver-worker.opam solver-service/solver-service.opam solver-service/solver-service-api.opam solver-service/
[...]

PR #11201, opened 2024-12-12
Regression test for #11200: `dune subst` fails on an empty directory

shonfeder:
Adds a regression test for #11200 

With dune 3.16 this command succeeds with no output. Dune 3.17 gives

```
+|  File ".", line 1, characters 0-0:
+|  Error: There is no dune-project file in the current directory, please add one
+|  with a (name <name>) field in it.
+|  Hint: 'dune subst' must be executed from the root of the project.
+|  [1]
```


Leonidas-from-XIV (approved):

shonfeder:
Fixed the DCO! Thanks for the fixes here and the forthcoming fix to dune :)

PR #11202, opened 2024-12-13
test: Repro 'dune fmt' crash in presence of Nix result

Julow:
Reproduction case for a crash when using `dune fmt`.
Perhaps formatting shouldn't follow symlinks ? Also, `dune build` doesn't seem to have any problem with that, perhaps because the linked directory doesn't contain a dune file.
Perhaps symlinked directories should be considered `vendored_dirs` automatically ?


anmonteiro (commented):
I have seen this before and it can be a bit puzzling. 

IMO the issue here isn't that dune is following symlinks, but rather that the files in the nix store `/nix/store/...` have pretty locked-down permissions as one would expect.


I'm not saying that dune is perfect, just trying to point out the actual cause. There might be work to be done in dune itself.

Julow:
Is here a case where we want the files in the symlinked directory be formatted ?

test/blackbox-tests/test-cases/nix-result-in-tree.t/run.t
```diff
@@ -0,0 +1,55 @@
+Nix can leave a symlink to a store path in the tree, often called 'result'.
+'dune fmt' crashes because of that.
+
+  $ RESULT=`mktemp -d`
+  $ echo "let x  = 2" > "$RESULT/foo.ml"
+  $ chmod -R a-w "$RESULT"
+  $ ln -s "$RESULT" result
+
+  $ dune fmt --auto-promote
```
Leonidas-from-XIV, line 9: Isn't `--auto-promote` the default for `dune fmt`?

test/blackbox-tests/test-cases/nix-result-in-tree.t/dune
```diff
@@ -0,0 +1,2 @@
+(executable
+ (name ocamlformat))
```
Leonidas-from-XIV, line 2: I don't think this has anything to do with Nix per-se, thus I would suggest renaming it to something akin to `read-only-symlink-target.t`.

test/blackbox-tests/test-cases/nix-result-in-tree.t/run.t
```diff
@@ -0,0 +1,55 @@
...
+    "otherlibs/stdune/src/exn.ml", line 38, characters 27-56
+  Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
+    line 76, characters 8-11
+  
+  I must not crash.  Uncertainty is the mind-killer. Exceptions are the
+  little-death that brings total obliteration.  I will fully express my cases. 
+  Execution will pass over me and through me.  And when it has gone past, I
+  will unwind the stack along its path.  Where the cases are handled there will
+  be nothing.  Only I will remain.
+  [1]
```
Leonidas-from-XIV, line 51: Could you filter the output a bit so it is more reproducible? The locations in the stacktrace are probably going to change over time. A simple `grep EACCES` is probably sufficient to show there's a very specific error that happens.

test/blackbox-tests/test-cases/nix-result-in-tree.t/run.t
```diff
@@ -0,0 +1,55 @@
...
+
+Allow Dune to remove temporary files (Dune crashes without this):
```
Leonidas-from-XIV, line 53: This is for the calling dune that is running `runtest` I assume?

Leonidas-from-XIV (commented):
Thanks for the repro case; I've posted some comments.

As to how to deal with projects like this, if we can't find the proper solution I would suggest adding it to the dev-meeting agenda to discuss.

test/blackbox-tests/test-cases/nix-result-in-tree.t/run.t
Julow: Yes, it seems that it fail at cleaning up the temporary files created during the test.

rgrinberg:
Thanks. This is just required a little more code for a better error message.

test/blackbox-tests/test-cases/read-only-symlink-target.t/run.t
```diff
@@ -0,0 +1,22 @@
...
+
+This command should succeed:
+
+  $ dune fmt
+  File "result/foo.ml", line 1, characters 0-0:
+  Error: Files _build/default/result/foo.ml and
+  _build/default/result/.formatted/foo.ml differ.
+  Promoting _build/default/result/.formatted/foo.ml to result/foo.ml.
+  Error: failed to promote result/foo.ml
+  Permission denied
```
Julow, line 17: Thanks for the fix! With this output, it's clearer that removing `result` would fix the issue.
But couldn't Dune avoid this issue ? Formatting files that are outside the workspace is surprising but could be part of someone's workflow. However, formatting files that are read-only feels like something Dune could avoid without bothering anyone.

rgrinberg: > However, formatting files that are read-only feels like something Dune could avoid without bothering anyone.

That's an interesting suggestion. It does seem like an improvement for the current behavior. There's no need to setup rules if we know they're good to fail. Do you want to try implementing this?

Julow: I've implemented that in https://github.com/ocaml/dune/pull/11490 but I have the feeling that is not the right solution. `result` links might interfere with other things than formatting and I think they should not be read by Dune.

PR #11203, opened 2024-12-13
Fix promotion of directory targets without files in them

### LINKED ISSUE #10609: [regression] promotion of directory targets has trouble with directories starting in @

ElectreAAS:
Replaces #10931.
On that PR, my colleague @moyodiallo made some comments with significant requested changes. He finally proposed something that seemed to work [here](https://github.com/moyodiallo/dune/commit/b00b92e1fbc1007736677436d3b9b56be3f089fc). Since then however, he left the company. I am here making a PR with his code.
This may fix #10609, cc @ejgallego.

### DIFF
--- a/src/dune_targets/dune_targets.ml
+++ b/src/dune_targets/dune_targets.ml
@@ -217,35 +217,54 @@ module Produced = struct
 
   let of_validated =
     (* The call sites ensure that [dir = Path.Build.append_local validated.root local]. *)
-    let rec collect (dir : Path.Build.t) (local : Path.Local.t)
-      : (unit Filename.Map.t Path.Local.Map.t, Error.t) result
-      =
-      match Path.readdir_unsorted_with_kinds (Path.build dir) with
-      | Error (Unix.ENOENT, _, _) -> Error (Missing_dir dir)
-      | Error e -> Error (Unreadable_dir (dir, e))
-      | Ok dir_contents ->
-        let open Result.O in
-        let+ filenames, dirs =
-          Result.List.fold_left
-            dir_contents
-            ~init:(Filename.Map.empty, Path.Local.Map.empty)
-            ~f:(fun (acc_filenames, acc_dirs) (filename, kind) ->
-              match (kind : File_kind.t) with
-              (* CR-someday rleshchinskiy: Make semantics of symlinks more consistent. *)
-              | S_LNK | S_REG ->
-                Ok (Filename.Map.add_exn acc_filenames filename (), acc_dirs)
-              | S_DIR ->
-                let+ dir =
-                  collect
-                    (Path.Build.relative dir filename)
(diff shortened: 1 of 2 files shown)


Leonidas-from-XIV:
Thanks, that looks like a very userful improvement especially wrt to the cache issue. Adding some other reviewers to this PR as I will be off for the rest of 2024.

ejgallego:
Hi folks, sorry for my lack of response on this issue, which is indeed IMHO very important.

I'll be still away for a few weeks due to personal stuff, but I hope to be back soon.

I think it'd be great to get the opinion of @rleshchinskiy which unless I am mistaken is the person who last touched that code.

src/dune_targets/dune_targets.ml
```diff
@@ -217,35 +217,54 @@ module Produced = struct
...
+                  acc_filenames, Path.Local.Map.union_exn acc_dirs dir
+                | _ -> Error (Unsupported_file (Path.Build.relative dir filename, kind)))
+          in
+          if not (Filename.Map.is_empty filenames)
+          then Path.Local.Map.add_exn dirs local filenames
+          else dirs
+      in
+      let rec parent_dirs path dirs =
+        match Path.Local.parent path with
+        | Some p when Path.Local.is_root p -> dirs
```
gridbugs, line 253: Is it a problem that this doesn't stop at the root of the project/workspace?

gridbugs: Or does it stop at the project root? I notice that the `Path.Local` module treats "." as the root so perhaps in practice this traversal never goes above the project root since the current directory will be somewhere in the project. If this is the case I think it deserves a comment explaining it, especially since my first instinct as a reader is to assume that `Path.Local.is_root` checks whether a path is "/" rather than ".".

ElectreAAS:
Superseded by #11226, closing.

PR #11222, opened 2024-12-17
Adapt JSOO rules for new --effects= option

OlivierNicole:
This takes into account the new js_of_ocaml `--effects={cps,double-translation}` flag, introduced in ocsigen/js_of_ocaml#1461.


src/dune_rules/jsoo/jsoo_rules.ml
```diff
@@ -50,84 +50,122 @@ module Config : sig
...
-    List.concat_map (get t) ~f:(function
-      | "toplevel", true -> [ "--toplevel" ]
-      | "toplevel", false -> []
-      | name, true -> [ "--enable"; name ]
-      | name, false -> [ "--disable"; name ])
-  ;;
+    List.filter_opt
+      [ (match t.toplevel with Some true -> Some "--toplevel" | _ -> None)
+      ; (match t.effects with
+         | Some backend -> Some ("--effects=" ^ string_of_effects backend) | None -> None)
```
hhugo, line 163: I think we need to remember if effect was set with `--enable effects` and use the same flag here otherwise we loose backward compatibility.

vouillon: I would select the flags based on jsoo version:
```ocaml
let to_flag ~recent =
  [...]
       | Some backend ->
           Some (if recent then "--effects=" ^ string_of_effects backend else "--enable=effects")
       | None -> None)
  [...]
```

OlivierNicole: Done.

OlivierNicole:
Note: I have used `failwith` when a dune rule is asking to pass `--effects=double-translation` with a version of jsoo/wasmoo that doesn’t support it, because I don’t know what the proper way of erroring in this code path is.

vouillon:
> Note: I have used `failwith` when a dune rule is asking to pass `--effects=double-translation` with a version of jsoo/wasmoo that doesn’t support it, because I don’t know what the proper way of erroring in this code path is.

I think you can just use `--enable=effects` in that case, and let the invocations of `js_of_ocaml --effects=double-translation` fail.

OlivierNicole:
I’m not sure to understand where the last CI failure comes from.

src/dune_rules/jsoo/jsoo_rules.ml
```diff
@@ -41,141 +41,213 @@ let js_env = compute_env ~mode:JS
...
-        | Some name -> set acc name false
-        | None -> set acc name true)
+        match
+          String.drop_prefix ~prefix:"!" name, String.drop_prefix ~prefix:"effects=" name
+        with
+        | Some name, _ -> set acc name `False
+        | None, None -> set acc name `True
+        | None, Some backend ->
+          (match effects_of_string backend with
+           | Some backend -> set acc name (`Effects backend)
```
vouillon, line 189: `name` is not right.
```suggestion
           | Some backend -> set acc "effects" (`Effects backend)
```
or
```suggestion
           | Some backend -> { acc with effects = Some backend }
```

src/dune_rules/jsoo/jsoo_rules.ml
```diff
@@ -41,141 +41,213 @@ let js_env = compute_env ~mode:JS
...
+let jsoo ~dir sctx =
+  Super_context.resolve_program
+    sctx
+    ~dir
+    ~loc:None
+    ~where:Original_path
+    ~hint:install_jsoo_hint
+    "js_of_ocaml"
+;;
+
```
vouillon, line 99: You don't need to move this.

src/dune_rules/jsoo/jsoo_rules.ml
```diff
@@ -48,85 +48,147 @@ module Config : sig
...
   val of_string : string -> t
   val of_flags : string list -> t
-  val to_flags : t -> string list
+
+  (** [recent] should be true if jsoo version is 6.0 or higher. *)
+  val to_flags : recent:bool -> t -> string list
 end = struct
+  type effects_backend =
+    | Cps
+    | Double_translation
```
vouillon, line 57: Maybe we should add `Native` and `JSPI` so that we don't have to modify `dune` if we add support for native effects in wasm_of_ocaml.

OlivierNicole: So dune would support an `--effects=native` flag for wasm_of_ocaml which may one day be supported by wsoo itself?

vouillon: Right

src/dune_rules/jsoo/jsoo_rules.ml
```diff
@@ -275,9 +337,21 @@ let js_of_ocaml_rule
...
          Dyn
-           (Action_builder.map config ~f:(fun config ->
-              Command.Args.S
-                (List.map (Config.to_flags config) ~f:(fun x -> Command.Args.A x)))))
+           (let+ config = config
+            and+ jsoo_version =
+              let* jsoo = jsoo in
+              Action_builder.of_memo (Version.jsoo_version jsoo)
+            in
+            let recent =
```
hhugo, line 345: The name `recent` does not say much . I think I would rather pass the version down and compare against it when needed.

src/dune_rules/jsoo/jsoo_rules.ml
```diff
@@ -48,85 +48,147 @@ module Config : sig
...
+    | "effects", `False ->
+      (* [--disable effects] *)
+      { acc with effects = None }
+    | "effects", `True ->
+      (* [--enable effects], used alone, implies [--effects=cps] *)
+      (match acc.effects with
+       | None -> { acc with effects = Some Cps }
+       | Some _ -> acc)
+    | "toplevel", `True -> { acc with toplevel = Some true }
+    | "toplevel", `False -> { acc with toplevel = Some false }
```
vouillon, line 89: It would be simpler to have two `set` functions. One for `--enable/--disable` and one for `--effects`.

src/dune_rules/jsoo/jsoo_rules.ml
hhugo: The downside is that it increases the number of dune rules registered in dune. One rule per lib per config in `Config.all`. We currently have 27 config on master.  Having cps, double_translate, jspi, and native (what is native btw ?) and "none" would result in 45 configs. I really which we eventually come up with a better design with lazy loading of rules. That would unblock https://github.com/ocaml/dune/pull/7389.

OlivierNicole:
Thank you for your reviews, I think I have addressed the ones made up to now.

hhugo:
I think you should add some tests that you would run manually now and that would run in the CI once jsoo.6.0 is released.

I've tested this PR manually and it doesn't quite work because the `--effects` flags is sometime repeated (which is treated as an error by jsoo)

OlivierNicole:
Can you share how you encountered the bug?

hhugo:
I've just tried to use this PR to run test in jsoo. Same behavior in https://github.com/ocsigen/js_of_ocaml/actions/runs/13032240200/job/36353967506?pr=1821

OlivierNicole:
I think it should be fixed now.

I will look at adding tests, should they be in this PR?

src/dune_rules/jsoo/jsoo_rules.ml
```diff
@@ -41,136 +41,223 @@ let js_env = compute_env ~mode:JS
...
+      ]
   ;;
 
-  let version_memo = Memo.create "jsoo-version" ~input:(module Path) impl_version
-
-  let jsoo_version jsoo =
-    match jsoo with
-    | Ok jsoo_path -> Memo.exec version_memo jsoo_path
-    | Error e -> Action.Prog.Not_found.raise e
+  let remove_effect_flags flags =
```
hhugo, line 248: It feels weird to special case the effects flags only.
Could we instead remove all flags already covered by the config ?

OlivierNicole: Done. However, it looks like this step of removing some flags has the unintended side effects of removing user-added flags as well. I added a test that is supposed to fail due to incompatible `--effects` flags, and the `--effects=double-translation` flag in the library's dune file is simply discarded.

vouillon: I thinks that's the point: we want to use the flags from where the library is used.
Some default flags are often sets at the root of the dune workspace. It would be cumbersome if we had to remove them for libraries.

OlivierNicole: Oh I see, I’ve been looking at this the wrong way.

OlivierNicole:
I have added a test but for it to work, it would require to use JSOO 6.0.1 in CI.

The other failures seem unrelated to this PR.

hhugo:
I don't really like the current situation but it's not the fault of this PR. Sadly I don't have time to spend on this right now. @vouillon would you have time to review this and approve if it's ok.

hhugo:
@OlivierNicole, @vouillon we should probably try to have this part of the next 3.18 release.

vouillon (approved):

vouillon:
> I don't really like the current situation but it's not the fault of this PR. Sadly I don't have time to spend on this right now. @vouillon would you have time to review this and approve if it's ok.

I'm not really happy with that either. Each time we add some options in js_of_ocaml/wasm_of_ocaml, we need to update Dune as well. But I think this is ok for now.

hhugo:
ping @OlivierNicole

OlivierNicole:
@hhugo Sorry for the delay; I have rebased.

rgrinberg:
@hhugo or @vouillon is this good to go?

hhugo:
good to merge

PR #11224, opened 2024-12-18
chore(ci): remove coveralls

maiste:
This PR removes the stage that pushes the report to Coveralls. Regarding the CI, we don't take advantage of the coverage in our case. Removing would make the CI green again and reduce computing time.

If we find reasons to get the coverage back, we can still reintroduce it.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -136,41 +136,6 @@ jobs:
       - uses: cachix/install-nix-action@v18
       - run: nix develop .#fmt -c make fmt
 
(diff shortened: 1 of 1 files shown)


rgrinberg (approved):

maiste:
The CI bug is unrelated to this. I merge the PR.

Alizter:
I guess coverage never really worked properly, but partly due to the weird side effects of instrumentation on tests. (We were getting instrumentation output during cram tests). There are still references to coverage in the README which I will remove.

PR #11234, opened 2024-12-20
 feat(dune): support `+` prefixes for PPX CLI flags 

anmonteiro:
this is useful for cases like `melange.ppx` which emit alerts and allow to configure them via `-alert ...`.

### DIFF
--- a/src/dune_rules/preprocess.ml
+++ b/src/dune_rules/preprocess.ml
@@ -12,9 +12,12 @@ module Pps_and_flags = struct
     and+ syntax_version = Dune_lang.Syntax.get_exn Stanza.syntax in
     let pps, more_flags =
       List.partition_map l ~f:(fun s ->
-        match String_with_vars.is_prefix ~prefix:"-" s with
-        | Yes -> Right s
-        | No | Unknown _ ->
+        match
+          ( String_with_vars.is_prefix ~prefix:"-" s
+          , String_with_vars.is_prefix ~prefix:"+" s )
+        with
+        | Yes, _ | _, Yes -> Right s
+        | (No | Unknown _), _ ->
           let loc = String_with_vars.loc s in
           (match String_with_vars.text_only s with
            | None ->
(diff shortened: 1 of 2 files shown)


anmonteiro:
Why do you think this needs to be versioned? 

I believe the previous behavior would just plainly return an error, so we shouldn't have anyone depending on these features, right?

rgrinberg:
Suppose that somebody would try to use this feature in a publicly released project. That person would need to remember to update the constraints in their project file to require at least dune 3.17. Without the version check, it is very likely that the person would forget to add this constraint, or they might even do it incorrectly, if they misremember which version of dune introduced this feature. Downstream user of this project might then encounter build failures should they build with an older version of dune. If we add a version check, we make this scenario more or less impossible.

It's true that the above argument is applicable to basically any user visible change in dune, but I think it works best for features. As for bug fixes, it's good that users can benefit from those without opting in. I think that this PR is certainly more of a feature than a bug fix.

anmonteiro:
absolutely, thanks for the explainer!

anmonteiro:
@rgrinberg ready for another look.

src/dune_rules/preprocess.ml
```diff
@@ -12,9 +12,13 @@ module Pps_and_flags = struct
...
-        match String_with_vars.is_prefix ~prefix:"-" s with
-        | Yes -> Right s
-        | No | Unknown _ ->
+        match
+          ( String_with_vars.is_prefix ~prefix:"-" s
+          , String_with_vars.is_prefix ~prefix:"+" s )
+        with
+        | Yes, _ -> Right s
+        | _, Yes when syntax_version >= (3, 18) -> Right s
+        | (No | Unknown _), _ ->
```
rgrinberg, line 21: I think you could improve the error message here with a hint for `prefix = "+"`  when `syntax_version < (3, 18)`

rgrinberg (approved):

PR #11239, opened 2024-12-24
refactor(pkg): remove candidates class

rgrinberg:


maiste (dismissed):

maiste (commented):
Shouldn't [this type](https://github.com/ocaml/dune/blob/acb928ff4a752fcb66424ca4506943464dbcdec3/src/0install-solver/solver_core.ml#L169) be updated with:
```ocaml
  module ImplCache = Cache (struct
      include Model.Role

      type value = Candidates.t (* instead of impl_candidates? *)
    end)
```

PR #11248, opened 2024-12-27
fix(melange): use output of virtual library when compiling its source 

anmonteiro:


src/dune_rules/melange/melange_rules.ml
```diff
@@ -597,6 +597,7 @@ let setup_js_rules_libraries =
             in
             cmj_includes ~requires_link ~scope lib_config
           in
+          let output = output_of_lib ~target_dir vlib in
```
anmonteiro, line 600: fix explanation: the `output` in scope was for `output_of_lib ~target_dir lib` instead of `vlib`.

anmonteiro:
ok I believe this also fixes #11247

rgrinberg (approved):

PR #11254, opened 2024-12-31
pkg: precompute directory descendants in rev_store

gridbugs:
Previously dune would do a linear search through all files in the rev store in order to compute the descendants of a single directory. This change precomputes all descendants of each directory in a single pass.

I've benchmarked this by running `dune pkg lock` in https://github.com/gridbugs/climate and it went from ~4.2s to ~3.6s on average (including the time taken to update the repos).

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -525,6 +525,7 @@ module At_rev = struct
     { repo : repo
     ; revision : Object.t
     ; files : File.Set.t
+    ; recursive_directory_entries : File.Set.t Path.Local.Table.t
     }
 
   let equal x y = Object.equal x.revision y.revision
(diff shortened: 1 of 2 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -684,25 +685,33 @@ module At_rev = struct
...
+          | Some parent ->
+            let key = Path.Local.to_string parent in
+            let descendants_of_parent =
+              String.Table.find_or_add descendants key ~f:(Fun.const File.Set.empty)
+            in
+            let descendants_of_parent = File.Set.add descendants_of_parent file in
+            String.Table.set descendants key descendants_of_parent;
+            loop (Path.Local.parent parent)
+        in
+        loop (File.path file |> Path.Local.parent));
```
maiste, line 702: I'm a bit lost here: you are looking for _descendant_, but you are calling `Path.Local.parent`. What am I missing here? :sweat_smile:

src/dune_pkg/rev_store.ml
```diff
@@ -525,6 +525,7 @@ module At_rev = struct
     { repo : repo
     ; revision : Object.t
     ; files : File.Set.t
+    ; descendants : File.Set.t String.Table.t
```
rgrinberg, line 528: How about `recursive_directory_entries`?

rgrinberg: Also, why don't you use a `Path.Local.Table`?

gridbugs: Done. `Path.Local.Table` wasn't exposed but it was implemented.

src/dune_pkg/rev_store.ml
gridbugs: I added some comments which might clarify. This code traverses each ancestor of the file by repeatedly calling parent, and for each ancestor directory it adds the file to the set of files contained under that directory.

rgrinberg (approved):

PR #11263, opened 2025-01-02
pkg: add argument to pkg lock to print perf stats

gridbugs:
When doing performance work on the solver it's helpful to see a breakdown of the time spent running the solver vs the time spent updating the package repos. I've also found the number of expanded packages to be both much higher than I expected and highly correlated to the performance runtime (unsurprisingly) so it's included in the stats too.

### DIFF
--- a/bin/pkg/lock.ml
+++ b/bin/pkg/lock.ml
@@ -70,6 +70,7 @@ let solve_lock_dir
   workspace
   ~local_packages
   ~project_pins
+  ~print_perf_stats
   version_preference
   solver_env_from_current_system
   lock_dir_path
@@ -94,6 +95,7 @@ let solve_lock_dir
       ~unset_solver_vars_from_context:
         (unset_solver_vars_of_workspace workspace ~lock_dir_path)
   in
+  let time_start = Unix.gettimeofday () in
   let* repos =
     let repo_map = repositories_of_workspace workspace in
     let repo_names =
@@ -105,6 +107,7 @@ let solve_lock_dir
     get_repos repo_map ~repositories:(repositories_of_lock_dir workspace ~lock_dir_path)
   in
   let* pins = resolve_project_pins project_pins in
+  let time_solve_start = Unix.gettimeofday () in
   progress_state := Some Progress_indicator.Per_lockdir.State.Solving;
   Dune_pkg.Opam_solver.solve_lock_dir
     solver_env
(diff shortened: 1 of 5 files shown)


bin/pkg/lock.ml
```diff
@@ -94,6 +95,7 @@ let solve_lock_dir
       ~unset_solver_vars_from_context:
         (unset_solver_vars_of_workspace workspace ~lock_dir_path)
   in
+  let time_start = Unix.gettimeofday () in
```
Leonidas-from-XIV, line 98: The `Unix` documentation is very light on this but usually `[gettimeofday](https://www.man7.org/linux/man-pages/man2/gettimeofday.2.html)` is not a good way to determine the runtime because it is not monotonic, thus affected by time changing e.g. on DST or system clock adjustments.

Looks like `mtime` exists exactly for this.

bin/pkg/lock.ml
```diff
@@ -119,19 +122,29 @@ let solve_lock_dir
...
+         ]
+         @
+         if print_perf_stats
+         then
+           [ Pp.nop
+           ; Pp.textf "Expanded packages: %d" num_expanded_packages
+           ; Pp.textf "Updated repos in: %.2fs" (time_solve_start -. time_start)
+           ; Pp.textf "Solved dependencies in: %.2fs" (time_end -. time_solve_start)
+           ]
+         else [])
```
Leonidas-from-XIV, line 147: Could you unwrap this a little, e.g. by defining the perf stats as a separate binding (containint the stats or `[]`), then you can always cons the actual message easily and don't need quadratic `@`.

This way the code does a lot of stuff, wrapped deeply and I find it a bit hard to read, as it requires more and more context the more layers are added.

Leonidas-from-XIV (commented):
I like the feature, but I would suggest some improvements to make it more reliable and make it a bit less dense in complexity.

bin/pkg/lock.ml
gridbugs: Good point, but mtime isn't exposed in the `Unix` module and there's already a precedent for using `gettimeofday` for measuring durations of time in dune.

bin/pkg/lock.ml
gridbugs: Good idea. Done.

bin/pkg/lock.ml
```diff
@@ -223,15 +239,18 @@ let lock ~version_preference ~lock_dirs_arg =
...
     ~lock_dirs
+    ~print_perf_stats
 ;;
 
 let term =
   let+ builder = Common.Builder.term
   and+ version_preference = Version_preference.term
-  and+ lock_dirs_arg = Pkg_common.Lock_dirs_arg.term in
+  and+ lock_dirs_arg = Pkg_common.Lock_dirs_arg.term
+  and+ print_perf_stats = Arg.(value & flag & info [ "print-perf-stats" ]) in
```
maiste, line 249: Do we want to generalize perf improvement to Dune or just to package management? Otherwise, it can be nice to move it to `Common`.

maiste (approved):
I add one question but otherwise LFTM :+1:

bin/pkg/lock.ml
Leonidas-from-XIV: Yeah, I also noticed that despite what the documentation of mtime says, to actually get the current mtime you need the C/JS stub, so there's no easy way to get that without vendoring another library.

bin/pkg/lock.ml
Leonidas-from-XIV: IMHO I think for now it's perfectly fine to just improve the locking performance as it is the biggest bottleneck.

Leonidas-from-XIV (approved):

PR #11264, opened 2025-01-02
pkg: don't expand deps of restricting deps

gridbugs:
Previously dune would expand dependencies of packages that are only ever marked as conflicting with packages in the solution. This led to the expansion of far more packages that necessary. Expanding a package requires reading an opam file and so is a relatively expensive operation.

For packages that only depend on the compiler (ocaml-base-compiler), this reduces the number of expanded packages from 8376 to 1841 at the time of writing.

Consider this to be a proof of concept demonstrating the problem and a potential solution. "Better" solutions might avoid the additional mutable state I've added here but would probably be more invasive. Since we've discussed eventually rewriting the solver anyway, I'm happy for this to be merged if my approach is deemed acceptable.

Here are the performance stats (using [this](https://github.com/ocaml/dune/pull/11263) feature) before and after this change on my machine (also including https://github.com/ocaml/dune/pull/11254 as it removes a large constant factor (technically `O(num-pkgs-in-repo)`) from the solver runtime). The project being locked has a single dependency on `ocaml-base-compiler`.
```
# before
$ dune pkg lock --print-perf-stats
Solution for dune.lock:
- ocaml-base-compiler.5.2.1

Expanded packages: 8376
Updated repos in: 0.91s
Solved dependencies in: 2.73s

# after
$ dune pkg lock --print-perf-stats
Solution for dune.lock:
- ocaml-base-compiler.5.2.1

Expanded packages: 1841
Updated repos in: 0.90s
Solved dependencies in: 0.45s
```

### DIFF
--- a/src/0install-solver/solver_core.ml
+++ b/src/0install-solver/solver_core.ml
@@ -167,6 +167,8 @@ module Make (Model : S.SOLVER_INPUT) = struct
     (* For each (iface, source) we have a list of implementations. *)
     let impl_cache = ImplCache.create () in
     let conflict_classes = Conflict_classes.create sat in
+    let deferred = ref [] in
+    let expand_deps = ref true in
     let+ () =
       let rec lookup_impl =
         let add_impls_to_cache role =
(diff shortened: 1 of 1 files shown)


src/0install-solver/solver_core.ml
```diff
@@ -197,21 +197,43 @@ module Make (Model : S.SOLVER_INPUT) = struct
...
       >>| Candidates.vars
       >>| S.at_least_one sat ~reason:"need root" (* Must get what we came for! *)
     in
+    (* Now process any restricting deps. Due to the cache, only restricting
+       deps that aren't also an essential dep will be expanded. The solver will
+       not process any transitive dependencies here since the dependencies of
+       restricting dependencies are irrelevant to solving the dependency
+       problem. *)
+    expand_deps := false;
+    let+ () = Fiber.sequential_iter !deferred ~f:(fun f -> f ()) in
```
rgrinberg, line 236: Is it ok that we've reversed the order here?

rgrinberg:
@talex5 do you mind doing a sanity check on this?

rgrinberg:
Is it possible to add a test to demonstrate a change in behavior? Since this change is loading opam files, I imagine that it should avoid raising an error when one of these opam files has an error.

talex5:
> @talex5 do you mind doing a sanity check on this?

This change sounds reasonable to me.

talex5:
Upstreaming would be good. It would likely make the solver service a fair bit faster.

src/0install-solver/solver_core.ml
rgrinberg: I restored the original order

rgrinberg (approved):
LGTM. Rebased the conflicts and added some minor cleanups

rgrinberg:
Merging to avoid needless conflicts down the line. Thanks!

gridbugs:
> Upstreaming would be good. It would likely make the solver service a fair bit faster.

@talex5 I had a look at upstreaming this change but dune's vendored copy of 0install has diverged quite a bit from upstream. I don't have time right now to work out how to apply this change upstream.

Issue #11272, opened 2025-01-05
dune / jbuilder: specify the maintenance intent


hannesm:
Dear Sir or Madam,

as mentioned in the other ticket, it would be great if dune itself could specify what the maintenance intention of dune (and the other opam files in your repository) is. I suspect "(latest)" would be a good choice.

A separate but related question is about "jbuilder": would marking it as "deprecated" (i.e. `flags: deprecated`) be a good choice - or is there still the expectation that people will use jbuilder?

Please note: if there's a package in opam-repository that is being kept, and relies on an older version of dune / jbuilder, these versions will be kept. So it is completely safe setting the maintenance intent to `"(latest)"` -- there won't be any uninstallable packages due to that choice.

Of course if you plan to support the latest major version as well, please set the x-maintenance-intent field accordingly. See the full specification in https://github.com/ocaml/opam-repository/blob/master/governance/policies/archiving.md#specification-of-the-x--fields-used-in-the-archiving-process

maiste:
FMU, and, according to what is written in the wiki regarding the maintenance of Dune, we only maintain the last version of dune (currently `3.17.2`). I have added to the agenda of the Wednesday Dune Dev meeting so we can clarify between ourselves the policy regarding Dune release and maintenance cycle.

maiste:
From the discussion we had during the Dune Dev meeting, it seems that having the maintenance intend as `(latest)` is the correct one. About `jbuilder` it seems relevant to mark it as deprecated. 

@rgrinberg does it sound good to you?

rgrinberg:
Sounds good. Thanks for taking care of this

maiste:
Thanks for the reply: I'll open the PR on opam to add this on Monday.

maiste:
You are right. I didn't think about it. Apart from `dune` itself, do we consider the other packages as maintain (apart from dune use case)?

rgrinberg:
At some point there was a plan to make these publicly available. At the very least, they were made public to dune-rpc and ocaml-lsp. I don't think much of that is relevant now and we can roll that back if possible. So yeah `(none)` would be more appropriate.

PR #11274, opened 2025-01-06
Add support for x-maintenance-intent in dune-project

art-w:
With progress being made on the [opam packages archiving policy](https://discuss.ocaml.org/t/proposed-package-archiving-policy-for-the-opam-repository/15713), a lot of published packages are going to add an `x-maintenance-intent: ["..."]` to their opam files. When using dune to generate the opam files, adding this new field requires an extra `.opam.template` file. Right now the custom opam templates escape hatch are fairly exceptional (~60 published projects uses them), so it would be nice for all the others to stay on the happy path if possible.

In this PR, the opam `x-maintenance-intent: [...]` is only generated if the user specified `(maintenance_intent ...)` in their dune-project. Perhaps it should produce `["(latest)"]` by default?

See #11272 (cc @hannesm)


maiste (commented):
Thanks for the contribution! :+1: 
You need to update the tests to reflect your changes. I would split this in two PRs: the part that updates the code and the part that updates Dune itself (in case it breaks for the former).

Also, could you add an entry into the [changes](https://github.com/ocaml/dune/tree/main/doc/changes) directory?

doc/reference/dune-project/generate_opam_files.rst
```diff
@@ -63,6 +63,18 @@ defined in the project:
        "Jane Doe <jane.doe@example.com>"
        "John Doe <john.doe@example.com>")
 
+.. describe:: (maintenance_intent <strings>)
+
+   .. versionadded:: 3.18
+
+   Specify the `opam maintenance intent <https://github.com/ocaml/opam-repository/blob/master/governance/policies/archiving.md#specification-of-the-x--fields-used-in-the-archiving-process>`.
```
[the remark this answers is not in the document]
art-w: Good catch, thank you! It was missing the underscores after the link:
```
`title <https://url>`__
```

art-w:
Thanks for the quick review! I've added tests, an entry in `changes` and split the opam files update into #11275

src/dune_rules/opam_create.ml
```diff
@@ -228,6 +228,7 @@ let opam_fields project (package : Package.t) =
   in
   let list_fields =
     [ "maintainer", Package_info.maintainers info
+    ; "x-maintenance-intent", Package_info.maintenance_intent info
```
maiste, line 231: Shouldn't this be inserted only if the field is not `None`?

art-w: Yes if I understand correctly there's a `filter_map` right after that removes any field with a useless `None|Some []` value :) (so if `maintenance_field` is not specified, it doesn't appear in the produced opam file). Did you have something else in mind?

maiste: No, just wanted to be sure :smile:

rgrinberg:
> Perhaps it should produce ["(latest)"] by default?

That sounds like a useful thing to do.

Is there a need to validate this field at all?

src/dune_rules/opam_create.ml
```diff
@@ -228,6 +228,10 @@ let opam_fields project (package : Package.t) =
   in
   let list_fields =
     [ "maintainer", Package_info.maintainers info
+    ; ( "x-maintenance-intent"
+      , match Package_info.maintenance_intent info with
+        | None -> Some [ "(latest)" ]
```
Leonidas-from-XIV, line 233: While this is probably reasonable, I think this should be only generated if dune-lang is >= 3.18, otherwise upgrading to 3.18 will unconditionally rewrite people's OPAM files.

test/blackbox-tests/test-cases/dune-project-meta/basic-generate.t
```diff
@@ -4,22 +4,23 @@ Simple test
 The `dune build` should generate the opam file
 
   $ cat >dune-project <<EOF
-  > (lang dune 1.10)
+  > (lang dune 3.18)
```
Leonidas-from-XIV, line 7: I don't think this is a good change, we should still retain tests that check that 1.10 era projects are supported. It's better to add an additional test for the maintenance intent (and that will also cause less changes in the cram tests).

test/blackbox-tests/test-cases/dune-project-meta/dune-dep.t
```diff
@@ -191,6 +191,7 @@ the doc dependencies:
     ]
     ["dune" "install" "-p" name "--create-install-files" name]
   ]
+  x-maintenance-intent: ["(latest)"]
```
Leonidas-from-XIV, line 194: When gating on dune-lang 3.18+ these should go away.

Leonidas-from-XIV (changes requested):
Nice feature, however due to how we handle compatibility an upgrade from Dune 3.17 to 3.18 should not trigger user visible changes (that is only allowed via dune-lang), so my suggestion is to gate the generation to dune-lang 3.18+. That will also remove lots of test changes, since these wouldn't change anymore.

polytypic:
This is perhaps not the right place to comment, because this is more about opam rather than dune, but if you generate `latest` by default, which also makes sense to me, then why generate it at all?  Why couldn't that be the default assumption made by opam when the field is absent?

Personally I'd like to see a much more explicitly elaborated story on how the `x-maintenance-intent` field is actually supposed to work and help opam as opposed to not having that field.

How are the intent values to be interpreted by users and by opam-repository maintainers?  How does it affect archiving? What happens if the maintainers just vanish?  What sort of automation is planned based on that field? ...

Like discussed in this PR, I would expect most people to just add `latest` and essentially forget about this.  If it is done implicitly by dune, a lot of people probably won't even be aware of it later on.  How will it help opam to have tons of packages with maintenance intent set to `latest` (as opposed to not having that field (by default))?

I'm sceptical that a field like this really helps.  Intent is subjective.  Actual behaviour is objective.  People are optimistic and tend to lie about their own intentions.  The expectation seems to be that package authors are good citizen and will carefully maintain the `x-maintenance-intent` field.  I question that assumption.  Circumstances change.  People move on.  I would not expect such a field to be maintained any better than the rest of the package.  In fact, I would expect such metadata fields to be maintained significantly less carefully than the actual content of the package.  That is just how people work in my experience.

To me this `x-maintenance-intent` field seems like yet another hurdle to publishing packages on opam, which I already think has been made too slow and laborious compared to other ecosystems.  With `x-maintenance-intent` you are now (exaggerating a bit) basically asked to do the equivalent of swearing on the Bible that you have the best of intentions in order to publish a package.

If the idea is to use `x-maintenance-intent` to automatically deprecate or archive packages or make choices on which versions to install and so on, then this could make sense, but in that case the semantics should be very carefully defined so that a package author can be set the field properly.

Or perhaps something like `x-maintenance-objectively` field which could be set by users and opam-repository maintainers after witnessing how a package has actually been maintained might be more useful? 😆 

FYI: Much of the above is duplicated from what I posted on OCaml Discord on the opam-repository channel, before noticing this PR.

src/dune_rules/opam_create.ml
art-w: Ha good point thank you! The opam files should indeed be stable and not change because of a dune update, I've gated the `x-maintenance-intent` to dune >= 3.18 as you suggested.

test/blackbox-tests/test-cases/dune-project-meta/basic-generate.t
art-w: Agree!

Leonidas-from-XIV (commented):
Nice. I have one small concern about the changes in `basic-generate.t` where I don't think the changes are required and a small documentation improvement but otherwise this looks very good.

test/blackbox-tests/test-cases/dune-project-meta/basic-generate.t
```diff
@@ -17,9 +17,9 @@ The `dune build` should generate the opam file
   >  (name cohttp)
   >  (synopsis "An OCaml library for HTTP clients and servers")
   >  (description "A longer description")
+  >  (allow_empty)
```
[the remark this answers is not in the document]
art-w: Indeed, thanks!

Leonidas-from-XIV (approved):
I've removed the unrelated test changes and rebased. Seems to work fine and the CI failure is unrelated and worked-around by #11293. I think it is fine to merge this before that one is merged.

doc/reference/dune-project/package.rst
```diff
@@ -60,6 +60,12 @@ package
 
       The same as (and takes precedences over) the corresponding global field.
 
+   .. describe:: (maintenance_intent ...)
+
+      .. versionadded:: 3.18
+
+      The same as (and takes precedences over) the corresponding global field.
```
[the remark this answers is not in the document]
art-w: Is there any way to link to an anchor with rst?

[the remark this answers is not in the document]
maiste: @Leonidas-from-XIV for uniformity, I don't think this is what we want. If you look above, there is no link to global.

maiste (approved):

doc/reference/dune-project/package.rst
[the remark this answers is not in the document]
Leonidas-from-XIV: @art-w No, I don't think so, at least not easily. I think the other file would need to define a named reference at the location where the field is defined and we would need to link to that. Or use a plugin which automatically creates these anchors. If you find an easier way, I'd love to know about it, it does feel like an odd omission in reST/RTD.

@maiste I understand your uniformity argument but I think it should not prevent us from making the documentation easier to navigate. I think it is always better to add a link (and we could do that for other fields that allow overriding as well, or say "all the fields are overrides of the global, link here", though that's not really true as you can't have global `depends`)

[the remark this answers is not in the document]
maiste: Indeed, my argument makes more sense regarding the code than the documentation. I'm fine with both in the end :smile:

art-w:
Thanks @Leonidas-from-XIV !

@rgrinberg > Regarding validation, could this be addressed in a follow-up PR if the lack of validation turns out to be an issue? Otherwise I'm happy to add the validation now.

rgrinberg:
> Regarding validation, could this be addressed in a follow-up PR if the lack of validation turns out to be an issue? Otherwise I'm happy to add the validation now.

I think we should add validation as soon as possible as doing it post the release would mean the validation would have to be guarded behind the new version.

@Leonidas-from-XIV could you revert this if validation isn't ready before we release 3.18

Leonidas-from-XIV:
> @Leonidas-from-XIV could you revert this if validation isn't ready before we release 3.18

Sure, will keep that in mind and make sure not to release 3.18 without :+1: (or revert before).

PR #11275, opened 2025-01-07
chore: add x-maintenance-intent to dune opam files

art-w:
Using #11274 to add it to every opam files. My understanding is that only the `"(latest)"` version is the one actively maintained, but it's easy to change if you want to include more versions.

### DIFF
--- a/dune-private-libs.opam
+++ b/dune-private-libs.opam
@@ -17,7 +17,7 @@ homepage: "https://github.com/ocaml/dune"
 doc: "https://dune.readthedocs.io/"
 bug-reports: "https://github.com/ocaml/dune/issues"
 depends: [
-  "dune" {>= "3.12"}
+  "dune" {>= "3.18"}
   "csexp" {>= "1.5.0"}
   "pp" {>= "1.1.0"}
   "dyn" {= version}
@@ -26,6 +26,7 @@ depends: [
   "odoc" {with-doc}
 ]
 dev-repo: "git+https://github.com/ocaml/dune.git"
+x-maintenance-intent: ["(latest)"]
 build: [
   ["dune" "subst"] {dev}
   ["rm" "-rf" "vendor/csexp"]
--- a/dune-private-libs.opam
+++ b/dune-private-libs.opam
@@ -26,7 +26,7 @@ depends: [
   "odoc" {with-doc}
 ]
 dev-repo: "git+https://github.com/ocaml/dune.git"
-x-maintenance-intent: ["(latest)"]
+x-maintenance-intent: ["none"]
 build: [
   ["dune" "subst"] {dev}
   ["rm" "-rf" "vendor/csexp"]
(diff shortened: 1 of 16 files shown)


Leonidas-from-XIV:
The OPAM files are generated automatically from `dune-project` so unless the change is added in the respective `.template` files, the change will be removed on the next `dune build`. I suggest landing #11274 first and then adding `maintenance_intent` to our `dune-project` in this PR, which will automatically update the opam files.

dune-private-libs.opam
```diff
@@ -26,6 +26,7 @@ depends: [
   "odoc" {with-doc}
 ]
 dev-repo: "git+https://github.com/ocaml/dune.git"
+x-maintenance-intent: ["(latest)"]
```
Leonidas-from-XIV, line 29: I believe on this one it should be `(none)`.

dyn.opam
```diff
@@ -9,13 +9,14 @@ homepage: "https://github.com/ocaml/dune"
...
 depends: [
-  "dune" {>= "3.12"}
+  "dune" {>= "3.18"}
   "ocaml" {>= "4.08.0"}
   "ordering" {= version}
   "pp" {>= "1.1.0"}
   "odoc" {with-doc}
 ]
 dev-repo: "git+https://github.com/ocaml/dune.git"
+x-maintenance-intent: ["(latest)"]
```
Leonidas-from-XIV, line 19: This one is probably also `(none)`.

ocamlc-loc.opam
```diff
@@ -19,6 +19,7 @@ conflicts: [
   "ocaml-lsp-server" {< "1.15.0"}
 ]
 dev-repo: "git+https://github.com/ocaml/dune.git"
+x-maintenance-intent: ["(latest)"]
```
Leonidas-from-XIV, line 22: I assume that this is `(none)` too.

Leonidas-from-XIV (changes requested):
I think some of these changes need to be discussed because while we do release them as packages I don't think all of them are meant to be used by others.

e.g. I don't think that we would do a patch-release for a bug in `stdune` if it doesn't affect the `dune` package. We'd probably fix it upstream to prevent it becoming an issue in the future but a lot of these packages are really just to support Dune.

Leonidas-from-XIV:
@rgrinberg Can you take a look and confirm that this matches our maintenance policy?

maiste:
I encounter an issue while working on this PR. To be able to have the maintenance field in `dune-project`, I need to increase the number of the lang to `3.18`. If I do so, it will build the maintenance field, but it will complain that it can't install `dune >= 3.18` because this version does not exist in the local switch test.

Also, it seems that the `3.18` version is the one that changed the formatting, so it also requires the formatting to change. Should it be done in a separated PR if possible?

maiste:
You are right, let's do it this way :+1:

rgrinberg (approved):

Leonidas-from-XIV (approved):
It looks good but for now blocked on #11478 as it is experiencing the same CI failures for the same reason.

Issue #11290, opened 2025-01-14
`dune subst` issues with OCaml-CI


mtelvers:
## Expected Behavior

Up to Dune 3.16.1, `dune subst` (invoked by opam), in an directory containing just an `.opam` file completed without error.  Since 3.17 an error is returned.

This issue relates to https://github.com/ocaml/dune/issues/11200, which was fixed by https://github.com/ocaml/dune/pull/11204, however, the two cases are different as #11200 covered completely empty directories by filtering them out, but this has not fixed OCaml-CI as OCaml-CI populates the directories with `.opam` files.

## Actual Behavior

```
$ dune subst
File ".", line 1, characters 0-0:
Error: There is no dune-project file in the current directory, please add one
with a (name <name>) field in it.
Hint: 'dune subst' must be executed from the root of the project.
```

## Reproduction

- PR with a reproducing test: 

### 3.16

```
$ mkdir dune-subst
$ cd dune-subst/
$ echo 'authors: [ "John Doe <john@doe.com>" ]' > foo.opam
$ dune --version
3.16.0
$ dune subst
$ echo $?
0
```

### 3.17.1

```
$ mkdir dune-subst
$ cd dune-subst/
$ echo 'authors: [ "John Doe <john@doe.com>" ]' > foo.opam
$ dune --version
3.17.1
$ dune subst
File ".", line 1, characters 0-0:
Error: There is no dune-project file in the current directory, please add one
with a (name <name>) field in it.
Hint: 'dune subst' must be executed from the root of the project.
```

## Specifications

- Version of `dune` (output of `dune --version`): >= 3.17.0
- Version of `ocaml` (output of `ocamlc --version`): 4.14.2
- Operating system (distribution and version): Ubuntu 22.04.4

gridbugs:
I had a look into this problem today.

[...]

PR #11291, opened 2025-01-14
dune subst test for directory containing just a .opam file

mtelvers:
PR supporting #11290

### DIFF
--- a/test/blackbox-tests/test-cases/github11290.t
+++ b/test/blackbox-tests/test-cases/github11290.t
@@ -1,0 +1,8 @@
+Running `dune subst` should succeed in a directory containing just a `.opam` file.
+Regression test for https://github.com/ocaml/dune/issues/11290
+
+  $ cat > foo.opam << EOF
+  > authors: [ "John Doe <john@doe.com>" ]
+  > EOF
+
+  $ dune subst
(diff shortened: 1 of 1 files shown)


Leonidas-from-XIV:
Can you make the tests succeed in CI? It's ok to add error output as long as it is reproducible.

(If you rebase the 5.3.0 CI failure should resolve itself, we updated the GH action since)

mtelvers:
I have rebased this PR.  I am working around the issue in OCaml-CI by adding this line to the build.  However, it isn't perfect as, it sometimes creates a `dune-project` file where there wasn't one previously.

```
echo '(lang dune 3.0)' > './dune-project'
```

The error which I see is this:

```
#=== ERROR while compiling current.dev ========================================#
# context     2.3.0 | linux/arm64 | ocaml-base-compiler.4.14.2 | pinned(file:///src/ocurrent)
# path        ~/.opam/4.14/.opam-switch/build/current.dev
# command     ~/.opam/4.14/bin/dune subst
# exit-code   1
# env-file    ~/.opam/log/current-1-e2b7be.env
# output-file ~/.opam/log/current-1-e2b7be.out
### output ###
# File ".", line 1, characters 0-0:
# Error: There is no dune-project file in the current directory, please add one
# with a (name <name>) field in it.
# Hint: 'dune subst' must be executed from the root of the project.
```

The full log is [here](https://ocaml.ci.dev:8100/job/2025-01-14/103727-ci-ocluster-build-f30189)

PR #11366, opened 2025-01-21
Fix cram tests build-path-prefix-map substitutions

art-w:
I ran into a tiny issue with cram tests output, where only the first `$TESTCASE_ROOT` path would be normalized to hide the dynamic path (using bindings from `$BUILD_PATH_PREFIX_MAP`), but not the other paths on the same line. It turns out that dune expected paths to be separated by whitespace, and I was producing a json list of filenames with no spaces anywhere (with `jq -c`). In retrospect this PR should also fix reproducibility issues when a parent directory contains spaces.

Furthermore, the `$TMPDIR` variable was also expected to be normalized in command outputs by the build-path-prefix-map, but it wasn't matching as `Process.run` was overriding the cram test `$TMPDIR` env var for a different path (the shared `Dtemp.temp_dir` which contains a bunch of other things).

### DIFF
--- a/doc/tests.rst
+++ b/doc/tests.rst
@@ -712,9 +712,9 @@ the standard BUILD_PATH_PREFIX_MAP_ environment variable. For example:
 
 .. code:: console
 
-   $ export BUILD_PATH_PREFIX_MAP="HOME=$HOME:$BUILD_PATH_PREFIX_MAP"
+   $ export BUILD_PATH_PREFIX_MAP="/HOME=$HOME:$BUILD_PATH_PREFIX_MAP"
    $ echo $HOME
-   $HOME
+   /HOME
 
 Note: Unlike Dune's version of Cram, the original specification for Cram
 supports regular expression and glob filtering for matching output. We chose
--- a/src/dune_engine/process.ml
+++ b/src/dune_engine/process.ml
@@ -863,6 +863,14 @@ let report_process_finished
 
 let set_temp_dir_when_running_actions = ref true
 
+let set_temp_dir ~temp_dir env =
+  match temp_dir, !set_temp_dir_when_running_actions with
+  | Some path, _ ->
+    Env.add env ~var:Env.Var.temp_dir ~value:(Path.to_absolute_filename path)
+  | None, true -> Dtemp.add_to_env env
+  | None, false -> env
+;;
+
 let await ~timeout { response_file; pid; _ } =
   let+ process_info, termination_reason =
     Scheduler.wait_for_build_process ?timeout pid ~is_process_group_leader:true
@@ -873,6 +881,7 @@ let await ~timeout { response_file; pid; _ } =
 
 let spawn
       ?dir
+      ?temp_dir
       ?(env = Env.initial)
       ~(stdout : _ Io.t)
       ~(stderr : _ Io.t)
@@ -946,14 +955,7 @@ let spawn
     Time.now ()
   in
   let pid =
-    let env =
-      let env =
-        match !set_temp_dir_when_running_actions with
-        | true -> Dtemp.add_to_env env
-        | false -> env
-      in
-      Env.to_unix env |> Spawn.Env.of_list
-    in
+    let env = set_temp_dir ~temp_dir env |> Env.to_unix |> Spawn.Env.of_list in
     let stdout = Io.fd stdout in
     let stderr = Io.fd stderr in
     let stdin = Io.fd stdin in
@@ -1008,6 +1010,7 @@ let spawn
 
 let run_internal
       ?dir
+      ?temp_dir
       ~(display : Display.t)
       ?(stdout_to = Io.stdout)
       ?(stderr_to = Io.stderr)
@@ -1053,6 +1056,7 @@ let run_internal
     let (t : t) =
       spawn
         ?dir
+        ?temp_dir
         ?env
         ~queued
         ~stdout:stdout_to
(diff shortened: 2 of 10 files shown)


doc/tests.rst
```diff
@@ -690,9 +690,9 @@ the standard BUILD_PATH_PREFIX_MAP_ environment variable. For example:
 
 .. code:: console
 
-   $ export BUILD_PATH_PREFIX_MAP="HOME=$HOME:$BUILD_PATH_PREFIX_MAP"
+   $ export BUILD_PATH_PREFIX_MAP="/HOME=$HOME:$BUILD_PATH_PREFIX_MAP"
    $ echo $HOME
-   $HOME
+   /HOME
```
art-w, line 695: (documentation was wrong, no `$` is added in front of the replacement)

src/dune_engine/process.ml
```diff
@@ -846,6 +846,14 @@ let report_process_finished
 
 let set_temp_dir_when_running_actions = ref true
 
+let set_temp_dir ~temp_dir env =
+  match temp_dir, !set_temp_dir_when_running_actions with
+  | Some path, _ ->
+    Env.add env ~var:Env.Var.temp_dir ~value:(Path.to_absolute_filename path)
+  | None, true -> Dtemp.add_to_env env
+  | None, false -> env
+;;
```
art-w, line 855: @rgrinberg There's a comment in the code indicating that janestreet uses `set_temp_dir_when_running_actions` internally. Setting it to `false` would accidentally use the correct `$TMPDIR` set by cram in the env, instead of overriding the path with a wrong one. I believe the new code preserves the old behavior, but is there any chance the new `?temp_dir` would help get rid of the ref on your side?

src/dune_rules/cram/cram_exec.ml
```diff
@@ -1,94 +1,5 @@
...
-      | Ok None -> ()
-      | Ok (Some sexp) ->
-        let command = Command.of_sexp (assert false) sexp in
-        Csexp.to_channel out (Atom (f command));
-        flush out;
-        loop ()
-    in
-    loop ()
-  ;;
-end
```
art-w, line 90: I believe this is dead code?

test/blackbox-tests/test-cases/pkg/install-action.t
```diff
@@ -11,8 +11,6 @@ Testing install actions
   $ build_pkg test
   foobar
 
-  $ export BUILD_PATH_PREFIX_MAP="/PKG_ROOT=test/target:$BUILD_PATH_PREFIX_MAP"
```
art-w, line 14: @rgrinberg Do you know if this line is still useful somehow? The previous substitutions would only consider absolute paths, so `test/target` was ignored. The new code just replaces any path anywhere, so it introduces a change on line 38 below. Should I rather ignore relative path in `BUILD_PATH_PREFIX_MAP` to preserve the old behavior?

art-w:
To motivate the changes a bit more, here's the before/after of the added test:

```diff
   $ echo /nest/$PWD/sub/dir
-  /nest//home/art-w/dune/_build/.sandbox/3df9ecef4c5322405be4332860ac3ec2/default/test/blackbox-tests/test-cases/cram/sub/dir
+  /nest/$TESTCASE_ROOT/sub/dir
   $ echo "[\"$PWD\",\"$PWD\"]"
-  ["$TESTCASE_ROOT","/home/art-w/dune/_build/.sandbox/3df9ecef4c5322405be4332860ac3ec2/default/test/blackbox-tests/test-cases/cram"]
+  ["$TESTCASE_ROOT","$TESTCASE_ROOT"]
 
   $ ls -a $TMPDIR
   .
   ..
-  build_3aee7a_dune
-  build_6c87aa_dune
-  build_785126_dune
-  dune-pkg_02d959_stderr
-  dune-pkg_0363ba_stderr
-  dune-pkg_055d04_stderr
-  ... a lot more stuff because $TMPDIR was shared
   $ echo $TMPDIR
-  /tmp/build_b5b53d_dune
+  $TMPDIR
   $ echo $TMPDIR/sub/dir
-  /tmp/build_b5b53d_dune/sub/dir
+  $TMPDIR/sub/dir
   $ echo The tempdir is at:$TMPDIR:
-  The tempdir is at:/tmp/build_b5b53d_dune:
+  The tempdir is at:$TMPDIR:
 
   $ echo /this/$SPACED/issue
-  /this/path/contains spaces/.but/it's not an/issue
+  /this/$SPACED/issue
 
   $ echo $BUILD_PATH_PREFIX_MAP
-  $LEFT=$RIGHT:$SPACED=path/contains spaces/.but/it's not an:HOME=HOME:/workspace_root=/home/art-w/dune/_build/.sandbox/3df9ecef4c5322405be4332860ac3ec2:$TESTCASE_ROOT=$TESTCASE_ROOT:$TMPDIR=$TMPDIR:$RIGHT=$RIGHT
+  $LEFT=$RIGHT:$SPACED=$SPACED:HOME=HOME:/workspace_root=/workspace_root:$TESTCASE_ROOT=$TESTCASE_ROOT:$TMPDIR=$TMPDIR:$RIGHT=$RIGHT
```

src/dune_rules/cram/cram_exec.ml
rgrinberg: It is. Do you mind moving this change to a separate PR?

test/blackbox-tests/test-cases/pkg/install-action.t
rgrinberg: The change is fine, but it needs to be behind a version check. I'd hate to break existing builds for this.

src/dune_engine/process.ml
rgrinberg: I don't see this global variable anymore internally. Feel free to get rid of it here as well I suppose. However, let's try to keep the PR's as focused as possible, so that would work should come in a separate PR.

test/blackbox-tests/test-cases/pkg/install-action.t
art-w: Rebased and added a version check for the change of semantics

rgrinberg (changes requested):
Some unrelated changes should be moved to separate PR's

art-w:
Sorry I wasn't available earlier to finish this, the dead code removal has been moved to PR #11568 (and I haven't touched the `set_temp_dir_when_running_actions` since this can be removed in another PR)

Issue #11375, opened 2025-01-24
Dune does not download dependencies in release profile


gridbugs:
To reproduce, on a machine without opam (so the deps don't accidentally resolve from the opam switch), install the [dune dev preview](https://preview.dune.build/), clone any dune project, and run `dune pkg lock && dune build --release`. Dune will not attempt to download or build dependencies, and will instead fail due to missing dependencies. Most likely it will first fail due to the lack of ocaml compiler. Install this with the system package manager and rebuild, and then any dependent libraries will fail to resolve. Verify that the problem is related to `--release` by running `dune build` which uses the dev preview. In this profile dune will download and build dependencies as expected.

E.g. on a docker container with just the dune dev preview installed as well as some build tools (but not opam or ocaml):
```
s@0c45e2b82fbe ~
+ $ git clone https://github.com/gridbugs/climate
Cloning into 'climate'...
remote: Enumerating objects: 836, done.
remote: Counting objects: 100% (70/70), done.
remote: Compressing objects: 100% (8/8), done.
remote: Total 836 (delta 65), reused 63 (delta 62), pack-reused 766 (from 1)
Receiving objects: 100% (836/836), 201.89 KiB | 5.18 MiB/s, done.
Resolving deltas: 100% (526/526), done.

s@0c45e2b82fbe ~
+ $ cd climate/

s@0c45e2b82fbe ~/climate (main=)
+ $ dune pkg lock
Solution for dune.lock:
- base.v0.17.1
- base-unix.base
- csexp.1.5.2
- dune-configurator.3.17.1
- jane-street-headers.v0.17.0
- jst-config.v0.17.0
- ocaml.5.3.0
- ocaml-base-compiler.5.3.0
- ocaml-compiler.5.3.0
- ocaml-compiler-libs.v0.17.0
- ocaml-config.3
- ocaml_intrinsics_kernel.v0.17.1
- ppx_assert.v0.17.0
- ppx_base.v0.17.0
- ppx_cold.v0.17.0
- ppx_compare.v0.17.0
- ppx_derivers.1.2.1
- ppx_enumerate.v0.17.0
- ppx_expect.v0.17.2
- ppx_globalize.v0.17.0
- ppx_hash.v0.17.0
- ppx_here.v0.17.0
- ppx_inline_test.v0.17.0
- ppx_optcomp.v0.17.0
- ppx_sexp_conv.v0.17.0
- ppxlib.0.34.0
- ppxlib_jane.v0.17.2
[...]

PR #11376, opened 2025-01-24
pkg: add repro for gh11375

gridbugs:
Repro for https://github.com/ocaml/dune/issues/11375


Leonidas-from-XIV (approved):

rgrinberg:
Isn't this just a clone of https://github.com/gridbugs/dune/blob/d3a08bfc8953a9a377edd71ea5891d61cc6c4953/test/blackbox-tests/test-cases/pkg/ignore-lock-release.t?

Leonidas-from-XIV:
Good point, it's very similar to that. Might need to revert.

However, I think we should determine whether ignoring lockdirs should be the behavior of `--release` or `-p`.

PR #11378, opened 2025-01-24
pkg: allow pkg for release builds

gridbugs:
Fixes https://github.com/ocaml/dune/issues/11375

@rgrinberg do you remember why "--ignore-lock-dir" was added to the arguments passed by "--release"? Is it now safe to remove?

### DIFF
--- a/bin/common.ml
+++ b/bin/common.ml
@@ -242,7 +242,6 @@ module Options_implied_by_dash_p = struct
       ; "--always-show-command-line"
       ; "--promote-install-files"
       ; "--require-dune-project-file"
-      ; "--ignore-lock-dir"
       ; "--default-target"
       ; "@install"
       ]
@@ -294,7 +293,7 @@ module Options_implied_by_dash_p = struct
     Term.with_used_args
       Arg.(
         value
-        & alias_opt (fun s -> [ "--release"; "--only-packages"; s ])
+        & alias_opt (fun s -> [ "--release"; "--ignore-lock-dir"; "--only-packages"; s ])
         & info
             [ "p"; "for-release-of-packages" ]
             ~docs
--- a/test/blackbox-tests/test-cases/pkg/ignore-lock-package-build.t
+++ b/test/blackbox-tests/test-cases/pkg/ignore-lock-package-build.t
@@ -1,4 +1,6 @@
-When building a project in release mode we should ignore the lock directory.
+When building a project with -p we should ignore the lock directory. This is so
+that packages with lockdirs in their source archive can be built by opam
+without using locked dependencies.
 
   $ . ./helpers.sh
 
@@ -17,4 +19,4 @@ When building a project in release mode we should ignore the lock directory.
   >  (depends test))
   > EOF
 
-  $ dune build @install --release
+  $ dune build @install -p foo


Leonidas-from-XIV:
Shouldn't `--ignore-lock-dir` be added to the options of `-p` instead? Since we want to avoid OPAM triggering `dune pkg`.

rgrinberg:
-p triggers --release. So yes, this is the reason. Opam release builds should ignore lock directories

gridbugs:
@rgrinberg would there be any harm in moving `--ignore-lock-dir` to `-p`. People coming from rust will probably run `dune build --release` under the (false) belief that it will build their project with more optimization and less debugging, as that's what a "release profile" means in that ecosystem. It will reduce adoption friction if `--release` doesn't ignore the lockdir.

rgrinberg:
Grepping through the opam repository shows me there's a handful of packages that use `--release` instead of `-p`. If you'd like to devise a plan to switch them, then sure I don't object.

gridbugs:
Looks like there's only 5 or so packages whose latest version uses `--release` in their opam file. I've checked and each package has its opam file checked in to VCS. None of these packages have lockdirs in any of their existing source archives, so we just need to make sure that new releases of these packages change to `-p`, at least if they start checking in their lockdir. I think we should announce that using `--release` in opam files is deprecated, make the change to dune moving `--ignore-lock-dir` to `-p`, and submit PRs to update the opam files checked into the current projects.

rgrinberg:
> Looks like there's only 5 or so packages whose latest version uses `--release` in their opam file. I've checked and each package has its opam file checked in to VCS. None of these packages have lockdirs in any of their existing source archives, so we just need to make sure that new releases of these packages change to `-p`, at least if they start checking in their lockdir. I think we should announce that using `--release` in opam files is deprecated, make the change to dune moving `--ignore-lock-dir` to `-p`, and submit PRs to update the opam files checked into the current projects.

Given the small number of packages in question, I think you should add a step to create an issue in the respective bug trackers telling people they need to switch.

You could also add a new flag (`--release-without-lockdir`?) that has the old behavior for ease of transition.

gridbugs:
> I think you should add a step to create an issue in the respective bug trackers
I was thinking I'd just send a PR that fixes the issue directly, though raising an issue would be less work.

> You could also add a new flag (--release-without-lockdir?) that has the old behavior for ease of transition.
Why not just let people use `--release --ignore-lockdir`? It would presumably have the same effect and not require adding an additional flag. The issue with either of these solutions is it will require projects to update their lower bound for dune to 3.18 (assuming that's the first version where `--release` does not imply `--ignore-lockdir`).

Leonidas-from-XIV:
Sounds like a good solution. So if I understand the conclusion correctly we would do the following:

1. We will move `--ignore-lock-dir` from `--release` to `-p`
2. We will create some PRs for upstream/opam-repository to migrate from `--release` to `-p`
3. Update the failing tests

I think this is quite nice, because in `opam-health-check` we needed to run `--profile=release` instead of `--release` because otherwise the lock file would be ignored and while a possible workaround it's also just a surprising and unintuitive UX, so this PR is a solid improvement in my eyes.

@gridbugs Do you need help with these? If you want I can handle the projects that use `--release` and submit the fixes.

gridbugs:
Thanks @Leonidas-from-XIV . I'll make PRs to upstream this change into individual projects today and link them here.

gridbugs:
Upstreaming PRs:

- [frama-c](https://git.frama-c.com/pub/frama-c/-/merge_requests/15)
- [opsian](https://github.com/Opsian/opsian-ocaml/pull/11)
- [llvm](https://github.com/alan-j-hu/llvm-dune/pull/23)
- [aws-config](https://github.com/Nymphium/aws-config/pull/2)

[tezos-rust-libs](https://gitlab.com/tezos/tezos-rust-libs) is deprecated so I ignored it.

gridbugs:
> Sorry, for clarification are you removing `--ignore-lock-dir` ?

No we're not removing `--ignore-lock-dir`, just the behaviour `--release` so that it no longer implies `--ignore-lock-dir`.

gridbugs:
Regarding frama-c (cc @bobot) it sounds like in the ideal world that project would be able to pass `--release --ignore-lock-dir`, however dune does not currently allow both of those flags to be passed at the same time, because the former implies the latter, and the way dune treats flags means that this is like passing `--ignore-lock-dir` twice, which is not allowed. After we change dune so that `--release` no longer implies `--ignore-lock-dir` then both flags can be passed at the same time, but of course old versions of dune will still have the old behaviour.

Leonidas-from-XIV (approved):

PR #11380, opened 2025-01-24
Optimize findlib cache

art-w:
The findlib DBs are used as keys in a hashtable for memoization, but we expect them to compare equal most of the time since they are created once per context, so physical equality can shortcut a lot of comparisons.

`dune build @all` on Irmin: (previously built, so nothing to do)

```
Before:
  Time (mean ± σ):     911.8 ms ±   9.0 ms    [User: 840.7 ms, System: 73.3 ms]
  Range (min … max):   897.8 ms … 923.4 ms    10 runs

After:
  Time (mean ± σ):     836.5 ms ±  10.0 ms    [User: 763.5 ms, System: 76.2 ms]
  Range (min … max):   821.6 ms … 848.1 ms    10 runs
```
 

On Tezos:

```
Before:
  Time (mean ± σ):     23.785 s ±  2.556 s    [User: 22.362 s, System: 1.502 s]
  Range (min … max):   21.948 s … 28.395 s    10 runs

After:
  Time (mean ± σ):     22.411 s ±  2.467 s    [User: 21.019 s, System: 1.473 s]
  Range (min … max):   20.620 s … 26.860 s    10 runs
```

### DIFF
--- a/src/dune_rules/findlib.ml
+++ b/src/dune_rules/findlib.ml
@@ -64,6 +64,12 @@ module DB = struct
     && String.equal t.ext_lib ext_lib
   ;;
 
+  let equal a b =
+    (* Since the DB is cached per context, physical equality will
+       shortcut almost all equality tests. *)
+    phys_equal a b || equal a b
+  ;;
+
   let hash { stdlib_dir; paths; builtins; ext_lib } =
     Poly.hash
       ( Path.hash stdlib_dir


src/dune_rules/findlib.ml
```diff
@@ -484,15 +494,13 @@ type t = DB.t
 
 let create =
   Per_context.create_by_name ~name:"findlib" (fun context ->
-    Memo.lazy_ (fun () ->
```
rgrinberg, line 487: Why did you remove the lazy?

src/dune_rules/findlib.ml
```diff
@@ -57,14 +58,22 @@ module DB = struct
...
   ;;
 
-  let hash { stdlib_dir; paths; builtins; ext_lib } =
+  let equal a b =
+    (* Since the DB is cached per context, physical equality will
+       shortcut almost all equality tests. *)
+    phys_equal a b || equal a b
+  ;;
+
+  let hash t = t.hash
```
rgrinberg, line 74: Isn't caching the hash function a separate optimzation? Can you split that into its own PR please?

src/dune_rules/findlib.ml
art-w: Just a small cleanup: It was created and then forced immediately, isn't it a noop?

src/dune_rules/findlib.ml
art-w: In my mind it was part of the same optimization to speed up the hashtable, but sure I can split it in a separate PR.

src/dune_rules/findlib.ml
rgrinberg: I need to double check, but I don't think so. We're creating a map keyed by the context where every value is the database for that context. We don't need to generate the database for a context until it's actually requested. We are not creating a memo node and immediately forcing it, because the forcing bit is delayed until we look up the context in this map (or at least this is  how it should work).

PR #11381, opened 2025-01-24
refactor: remove force lazy from findlib

art-w:
As far as I understand, the `lazy_ |> force` isn't doing anything interesting (but I'm curious to know if it's actually useful?)


art-w:
Thanks ok, got it, `Memo.Lazy` isn't about delaying execution until needed, because fibers are already demand-driven, but it guarantees that the fiber will be executed only once.

PR #11382, opened 2025-01-24
Menhir in qualified subdirectory doesn't see modules from the same directory

Niols:
Reproducing PR for https://github.com/ocaml/dune/issues/11383. I have chosen to update the existing test rather than adding another one, because I think it now just tests strictly more things. If we prefer more atomic tests for more precise reporting, I will happily split it in two.


maiste:
Closing in favor of #11118

PR #11402, opened 2025-01-27
Allow dash in dune init project name

### LINKED ISSUE #11210: Supporting `-`  for `dune init project`
## Desired Behavior

At the moment, if one does:
```
dune init project foo-bar
dune: NAME argument: invalid component name `foo-bar'
      Library names must be non-empty and composed only of the
      following
      characters: 'A'..'Z', 'a'..'z', '_' or '0'..'9'.
Usage: dune init project [OPTION]… NAME [PATH]
Try 'dune init project --help' or 'dune --help' for more information.
```

but I believe it could succeed by using `-` for the public name and `_` for the internal name.

art-w:
Closes #11210

### DIFF
--- a/bin/dune_init.ml
+++ b/bin/dune_init.ml
@@ -245,20 +245,27 @@ module Component = struct
     module Common = struct
       type t =
         { name : Dune_lang.Atom.t
+        ; public : Public_name.t option
         ; libraries : Dune_lang.Atom.t list
         ; pps : Dune_lang.Atom.t list
         }
+
+      let package_name common =
+        let name =
+          match common.public with
+          | None -> Dune_lang.Atom.to_string common.name
+          | Some public -> Public_name.to_string public
+        in
+        Package.Name.of_string name
+      ;;
     end
 
     module Executable = struct
-      type t = { public : Public_name.t option }
+      type t = unit
     end
 
     module Library = struct
-      type t =
-        { public : Public_name.t option
-        ; inline_tests : bool
-        }
+      type t = { inline_tests : bool }
     end
 
     module Project = struct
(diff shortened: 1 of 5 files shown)


bin/dune_init.ml
```diff
@@ -245,20 +245,27 @@ module Component = struct
     module Common = struct
       type t =
         { name : Dune_lang.Atom.t
+        ; public : Public_name.t option
         ; libraries : Dune_lang.Atom.t list
         ; pps : Dune_lang.Atom.t list
         }
+
+      let opam_package_name common =
```
rgrinberg, line 253: How about just `package_name`? We don't have any other types of packages in dune.

art-w: Sure! I've renamed it :)

test/blackbox-tests/test-cases/dune-init/github11210.t
```diff
@@ -0,0 +1,27 @@
...
+
+  $ dune init project invalid.name
+  dune: NAME argument: invalid project name `invalid.name'
+        Project names must start with a letter and be composed only of
+        letters,
+        numbers, '-' or '_'
+  Usage: dune init project [OPTION]… NAME [PATH]
+  Try 'dune init project --help' or 'dune --help' for more information.
+  [1]
+  $ ls invalid.name
```
Leonidas-from-XIV, line 25: You can probably use `dune_cmd exists` to get portable output as BSD `ls` fails with a different error than this.

art-w: Ooh thanks for the tip!

PR #11403, opened 2025-01-27
doc: Avoid title case and make FAQs all questions

Leonidas-from-XIV:
It was raised by @MisterDA that the title casing does not make sense and @shonfeder as a native speaker agreed that it is a bit odd.

I've also took the liberty of making the titles all questions, because some were more like miniature howtos (and should probably be moved out of the document, as the comment rightfully states).

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -8,7 +8,7 @@ FAQ
 
    Some of these are mini how-to guides, or sections of existing guides.
 
-Why Do Many Dune Projects Contain a ``Makefile``?
+Why do many Dune projects contain a ``Makefile``?
 =================================================
 
 Many Dune projects contain a root ``Makefile``. It's often only there for
@@ -23,8 +23,8 @@ convenience for the following reasons:
 
 3. ``make`` is shorter to type than ``dune build @install``
 
-How to Add a Configure Step to a Dune Project
-=============================================
+How to add a configure step to a Dune project?
+==============================================
 
 The with-configure-step_ example shows one way to add a configure step that
 preserves composability; i.e., it doesn't require manually running the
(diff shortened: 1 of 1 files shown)


doc/faq.rst
```diff
@@ -171,8 +171,8 @@ directory - not the project root. If the library defines a ``name`` distinct fro
 its ``public_name`` then that can be used interchangeably with the ``public_name``
 in this command.
 
-Files and Directories Whose Names Begin with "." (Period) are Ignored by ``source_tree``
-========================================================================================
+Why are files and directories whose names begin with "." (period) are ignored by ``source_tree``?
```
shonfeder, line 174: We already have "are" at the start.

```suggestion
Why are files and directories whose names begin with "." (period) ignored by ``source_tree``?
```

Tho maybe something like this formulation would be a bit easier to parse out?

```suggestion
Why does ``source_tree`` ignore files and directories when they begin with a "." (period)?
```

doc/faq.rst
```diff
@@ -224,7 +224,7 @@ In this case you can update it to look like this:
     (dirs :standard .cargo \ target)
     (data_only_dirs vendor)
 
-How Can I Write Inline Tests in a Package Without my Users Needing to Install ``ppx_inline_test``?
+How can I write inline tests in a package without my users needing to install ``ppx_inline_test``?
```
shonfeder, line 227: Because there is apparently no answer to "how", I think we can just answer the why not?

```suggestion
Why can't I write inline tests in a package without users needing to install ``ppx_inline_test``?
```

shonfeder (approved):
I think this reads much better. Thanks!

gridbugs (approved):

Issue #10964, opened 2024-09-28
Running Dev Tools in Dune


rgrinberg:
We’ve done good work on the automatic building and installation of dev tools, but I’d like to see a better design for how editor integration is going to look like. The proposed approach of prefixing every dev tool with dune isn’t going to scale at all.

Editors, various scripts, and other dev tools all expect the dev tools to exist in PATH when they’re running. We don’t want to update every single one of them to have a mode when running under dune’s package management.

We should think about how to make it easier for various dev-tools to integrate with each other rather than creating new barriers. Eventually, the end goal should be that someone should be able to create a dev tool and have dune manage it without modifying dune at all. This is the way to go to encourage the creation of better tooling.

If you’re out of ideas on what to do concretely, I suggest you just follow along in esy steps and provide an explicit dev environment that editors are expected to inherit and dev tools need to run in.

edwintorok:
FWIW here is how I configured [helix](https://docs.helix-editor.com/languages.html#language-server-configuration) in `~/.config/helix/languages.toml`:
```
[language-server.ocamllsp]
command="dune"
args=["tools","exec","ocamllsp"]
```

As you can see I had to split it up, because the command needs to be an actual executable.

Would be good if I could somehow get `ocamllsp` into my $PATH. Helix' support for OCaml is otherwise excellent: you don't even need to configure anything, it just works if you've installed `ocaml-lsp-server` with `opam` and you've used `eval $(opam env)` or equivalent. Would be good if `dune`'s editor support is similarly easy to use with zero config changes in the editors themselves.


It would be good if multiple mechanisms were supported:
* `dune tools exec ..` can be useful if you open multiple files/projects in the same editor and they use different compiler versions: you don't need to relaunch your editor to pick up different env vars, the plugin can be taught to invoke this
* `source _build/activate` (similar to how you'd `source venv/bin/activate` with Python. No editor changes needed.
* `dune env` or `dune tools env` that prints PATH and other needed settings, similar to how `opam env` works

gridbugs:
> Would be good if I could somehow get ocamllsp into my $PATH.

[...]

PR #11408, opened 2025-01-29
Add dune tools env to configure PATH

art-w:
This is a proposal for https://github.com/ocaml/dune/issues/10964 to solve the following issues:
- Allow users to configure their environment with `eval $(dune tools env)` to update their PATH with the local dune dev tools binaries
- `dune tools exec ocamllsp` starts ocamllsp with that same PATH env, such that dune's `ocamlformat` can be found by LSP

I'm not entirely sure if my solution is acceptable: it create a new `_build/_private/default/.dev-tool/bin` folder, which contains executable shell scripts which calls `dune tools exec $tool`. This avoids polluting the PATH with each `.dev-tool/$tool/$tool/target/bin` and also enables ocamllsp to call ocamlformat even when it's not yet installed (as `dune tools exec` will handle it).

This PR does not however cover the "run arbitrary executables with `dune tools exec foo`" described in the issue, but I'm planning to work on that soon.

### DIFF
--- a/bin/tools/ocamllsp.ml
+++ b/bin/tools/ocamllsp.ml
@@ -13,7 +13,8 @@ let run_ocamllsp common ~args =
        ~object_:
          (User_message.command (String.concat ~sep:" " (ocamllsp_exe_name :: args))));
   Console.finish ();
-  restore_cwd_and_execve common exe_path_string (exe_path_string :: args) Env.initial
+  let env = Shell_env.add_path Env.initial in
+  restore_cwd_and_execve common exe_path_string (exe_path_string :: args) env
 ;;
 
 let build_ocamllsp common =
--- a/bin/tools/shell_env.ml
+++ b/bin/tools/shell_env.ml
@@ -1,0 +1,41 @@
+open! Import
+module Pkg_dev_tool = Dune_rules.Pkg_dev_tool
+
+let all_tools = List.map ~f:Pkg_dev_tool.exe_name [ Ocamlformat; Ocamllsp ]
+let bin_path () = Path.build (Pkg_dev_tool.bin_path ())
+
+let shell_script = {|#!/bin/sh
+dune tools exec $(basename $0) -- "$@"
+|}
+
+let setup path =
+  if not (Path.exists path) then Path.mkdir_p path;
+  List.iter all_tools ~f:(fun tool ->
+    let tool_path = Path.relative path tool in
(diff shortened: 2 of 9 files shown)


Leonidas-from-XIV:
Wouldn't it be simpler to launch commands with `dune tools exec <TOOLS>` with their `PATH` expanded with the tools that are installed? This can even be a single virtual `bin` folder with symlinks as you propose.

I personally find `eval $(...)` mess really brittle and a source of constant issues. More often than not it leads to capture of wrong `PATH`s, mismatched artifacts ending up in wrong places because some wrong `ocamlc` was executed etc. There is no good way to scope extensions to `PATH`, thus with `eval` you're one `cd` away from using the wrong tools in the wrong folder.

The issue with Dune installing not-yet installed tools is that the caller has no input on how this is done. So if ocamllsp calls ocamlformat and the latter is not installed, this command might stall for an undetermined amount of time to install it without any way to determine whether it will or not before executing it. I think this was one of the reasons why a Dune LSP was proposed, such that editors can inquire the status and interact with the package management using a modern API (which could also be used by command line tools if they just want to install a tool by supplying an CLI DuneLSP client).

shonfeder:
IMO, we would like something like this, for the reasons noted in point (2) of https://github.com/ocaml/dune/issues/10964#issuecomment-2386448472 . I understand it can be iffy to ask users to do this in their shells, but allowing other tooling to make use of this functionality seems like a reasonable and useful thing to do.

shonfeder:
I am even more convinced that we really need this (or approximately this), and that it is part of a simple and elegant solution to several important needs, including 

- https://github.com/ocaml/dune/issues/12107
- https://github.com/ocaml/dune/issues/12135
- and stable editor support

> This PR does not however cover the "run arbitrary executables with dune tools exec foo" described in the issue, but I'm planning to work on that soon.

I agree that this functionality is essentially orthogonal to the feature proposed here and the needs it addresses.

shonfeder:
This was addressed via https://github.com/ocaml/dune/pull/12521

However, I will say that I prefer the solution here insofar as it would

> create a new _build/_private/default/.dev-tool/bin folder, which contains executable shell scripts which calls dune tools exec $tool. This avoids polluting the PATH with each .dev-tool/$tool/$tool/target/bin


However, the core feature is in, and we can explore this can of improvement as a followup. Thanks, @art-w !

PR #11439, opened 2025-02-04
refactor: cleanups in the opam solver

rgrinberg:
* Use |> to improve readability along with labels
* Use `Package_version.dev` for the version

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -148,11 +148,9 @@ module Context = struct
       let available = OpamFile.OPAM.available opam in
       match
         OpamFilter.partial_eval
-          (add_self_to_filter_env
-             package
-             (Solver_stats.Updater.wrap_env
-                t.stats_updater
-                (Solver_env.to_env t.solver_env)))
+          (Solver_env.to_env t.solver_env
+           |> Solver_stats.Updater.wrap_env t.stats_updater
+           |> add_self_to_filter_env package)
           available
         |> eval_to_bool
       with
@@ -244,7 +242,6 @@ module Context = struct
   ;;
 
   let filter_deps t package filtered_formula =
-    let name = OpamPackage.name package |> Package_name.of_opam_package_name in
     (* Add additional constraints to the formula. This works in two steps.
        First identify all the additional constraints applied to packages which
        appear in the current package's dependency formula. Then each additional
@@ -252,8 +249,9 @@ module Context = struct
     let filtered_formula =
       OpamFormula.fold_left
         (fun additional_formulae (pkg, _) ->
-          let name = Package_name.of_opam_package_name pkg in
-          match Package_name.Map.find t.constraints name with
+          match
+            Package_name.of_opam_package_name pkg |> Package_name.Map.find t.constraints
+          with
           | None -> additional_formulae
           | Some additional -> additional :: additional_formulae)
         []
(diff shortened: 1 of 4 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -244,30 +242,34 @@ module Context = struct
...
-    let package_is_local = Package_name.Map.mem t.local_packages name in
-    Resolve_opam_formula.apply_filter
-      (add_self_to_filter_env
-         package
-         (Solver_stats.Updater.wrap_env t.stats_updater (Solver_env.to_env t.solver_env)))
-      ~with_test:package_is_local
-      filtered_formula
+    let package_is_local =
+      let name = OpamPackage.name package |> Package_name.of_opam_package_name in
+      Package_name.Map.mem t.local_packages name
```
maiste, line 265:
```suggestion
      OpamPackage.name package 
      |> Package_name.of_opam_package_name
      |> Package_name.Map.mem t.local_packages
```
Considering this PR is adding `|>`, is there a reason not to do this too?

src/dune_pkg/opam_solver.ml
```diff
@@ -1816,21 +1816,24 @@ let reject_unreachable_packages =
...
+                    (Package_version.to_opam_package_version Package_version.dev)
+                in
+                Solver_env.to_env solver_env |> add_self_to_filter_env opam_package
+              in
+              Dependency_formula.to_filtered_formula pkg.dependencies
+              |> Resolve_opam_formula.filtered_formula_to_package_names
+                   ~env
+                   ~with_test:true
+                   ~packages:
+                     (Package_name.Map.set pkgs_by_version Dune_dep.name dune_version)
```
maiste, line 1835: Wouldn't it be more readable if env was computed outside the `match ... with`?

src/dune_pkg/opam_solver.ml
```diff
@@ -244,30 +242,34 @@ module Context = struct
...
        appear in the current package's dependency formula. Then each additional
        constnraint is and-ed with the current package's dependency formula. *)
     let filtered_formula =
       OpamFormula.fold_left
         (fun additional_formulae (pkg, _) ->
-          let name = Package_name.of_opam_package_name pkg in
-          match Package_name.Map.find t.constraints name with
+          match
+            let name = Package_name.of_opam_package_name pkg in
+            Package_name.Map.find t.constraints name
```
maiste, line 254:
```suggestion
            Package_name.of_opam_package_name pkg
            |> Package_name.Map.find t.constraints name
```
Considering this PR is adding |>, is there a reason not to do this?

maiste (approved):
I added some nitpicky questions.

src/dune_pkg/opam_solver.ml
rgrinberg: Why do you think it would be more readable? This change reduces the scope of the identifiers. All else being equal, this make the code more readable.

src/dune_pkg/opam_solver.ml
rgrinberg: No objection from me. There's plenty of places where this improvement could be done but hasn't been yet.

src/dune_pkg/opam_solver.ml
maiste: You are right, maybe it is because I'm more used to matching on variables instead of expression.

PR #11442, opened 2025-02-04
feat: introduce env_vars alias for env-vars

### LINKED ISSUE #11424: consistency in dashes Vs underscores (e.g. "env-vars")
See an example of rare use of dashes:
https://github.com/ocaml/dune/blob/main/src/dune_rules/dune_env.ml#L200

This can make search in the documentation not great.

The env-vars field could just have an `env_vars` alias and be backwards compatible.

maiste:
This PR introduces an alias `env_vars`, for `env-vars`, to keep the configuration uniformity (using the underscore instead of dash).

Closes #11424


nojb (commented):
I don't think it is a good idea to allow both `(env-vars)` and `(env_vars)` (having more than one name for the same thing is generally bad). So `(env-vars)` should be renamed to `(env_vars)` in 3.18 (see `Dune_lang.Syntax.renamed_in`). That way, someone who upgrades to Dune lang 3.18 is forced to rename `env-vars` to `env_vars`.

maiste:
I agree! But is it a good idea to introduce it as a breaking change for `3.18` and not for `4.0`?

maiste:
@nojb what do you think about allowing both but emit a warning for now, and removing it for `4.0.0`?

rgrinberg:
I don't particularly care if we have more than one name, but I'm strongly against making breaking changes that offer no benefits to the user. So even if we introduce a new name, we need to keep supporting the old name even in 4.0. We need much more convincing reasons to break user builds.

maiste:
OK, so let's stick to the two names, then.

Issue #11440, opened 2025-02-04
ctypes' `deps` field doesn't follow the dependency specification language


Khady:
## Expected Behavior

The documentation of the `cytpes` feature mentions that there is support for deps, with the usual dependency specification language

https://dune.readthedocs.io/en/stable/foreign-code.html#ctypes-field-reference
> Since the Dune’s ctypes feature is still experimental, it could be useful to add additional dependencies in order to make sure that local headers or libraries are available: (deps <deps-conf list>). See [Dependency Specification](https://dune.readthedocs.io/en/stable/concepts/dependency-spec.html) for more details.

So I expect to be able to write
```
 (ctypes
  (external_library_name somelib)
  (build_flags_resolver pkg_config)
  (deps
   (:h "ocaml_bindings.h"))
  (headers
   (preamble "#include \"ocaml_bindings.h\""))
  (type_description
   (instance Type)
   (functor Type_description))
  (function_description
   (concurrency sequential)
   (instance Functions)
   (functor Function_description))
  (generated_types Types_generated)
  (generated_entry_point C))
```

## Actual Behavior

```
File "backend/cassandra_client/dune", line 14, characters 3-26:
14 |    (:h "ocaml_bindings.h"))
        ^^^^^^^^^^^^^^^^^^^
Error: Unexpected list
```

## Reproduction

1. setup a basic project like indicated in https://dune.readthedocs.io/en/stable/foreign-code.html#a-toy-example
1. create an `ocaml_bindings.h` file
1. add `(deps (:h "ocaml_bindings.h"))`
1. `dune build`

## Specifications

- Version of `dune` (output of `dune --version`): 3.16.0
- Version of `ocaml` (output of `ocamlc --version`): 4.14
- Operating system (distribution and version): debian 12

maiste:
```lisp
 (ctypes
  ;; [...]
  (deps
   (:h ocaml_bindings.h))
  ;; [...]
```
It might be a dummy suggestion, but can you try without the quotes?

Because, looking at the code, the decoder is supposed to take a list https://github.com/ocaml/dune/blob/26d680df8dcec76e216af8f7771eb4cc2af2960b/src/dune_rules/ctypes/ctypes_field.ml#L170 (the type is `Dep_conf.t list option` as expected. I don't see why it would fail...

Khady:
the error is the same without the quotes

emillon:
[...]

PR #11443, opened 2025-02-05
test(ctypes): showcase deps field not supporting the full dependency specification

Khady:
creating an example to showcase the issue explained in #11440


test/blackbox-tests/test-cases/ctypes/deps-full-spec.t/run.t
```diff
@@ -0,0 +1,18 @@
...
+
+Then generate cstubs for it, build an executable that uses those cstubs, and
+run the executable that tests the library through the cstubs.
+
+  $ LIBEX=$(realpath "$PWD/../libexample")
+  $ DYLD_LIBRARY_PATH="$LIBEX" LD_LIBRARY_PATH="$LIBEX" PKG_CONFIG_PATH="$LIBEX/pkgconfig" PKG_CONFIG_ARGN="--define-prefix" dune exec ./example.exe
+  File "dune", line 14, characters 40-48:
+  14 |     "#include \"example.h\"\n#include \"%{foo_h}\"\n#include \"%{bar_h}\"\n#include \"baz.h\"\n#include \"%{dep:qux.h}\""))
+                                               ^^^^^^^^
+  Error: %{foo_h} isn't allowed in this position.
```
Khady, line 13: I don't know what to look at to be able to fix this error

Leonidas-from-XIV: Shouldn't this be `%{dep:foo.h}`? `

Khady: No, I want to use the name of the dependency, this is on purpose.

Khady:
Is it something that is worth fixing further?

PR #11446, opened 2025-02-05
fix: support package pinning via http+tar

### LINKED ISSUE #10121: Support pinning packages with an archive url
Currently it looks like we support packages pinned to local paths and git urls but not archive urls. The [ocaml.org project](https://github.com/ocaml/ocaml.org) has a pin `["tailwindcss.dev" "https://github.com/tmattio/opam-tailwindcss/archive/3e60fc32bbcf82525999d83ad0f395e16107026b.tar.gz"]`. Currently locking this project (after removing the `package` stanza from its dune-project file to workaround https://github.com/ocaml/dune/issues/10120) produces the error:
```
File "ocamlorg.opam", line 1, characters 0-0:
Error: Could not determine location of repository
https://github.com/tmattio/opam-tailwindcss/archive/3e60fc32bbcf82525999d83ad0f395e16107026b.tar.gz
Hint: Specify either a file path or git repo via SSH/HTTPS
```

maiste:
This PR reduces the gap between opam and dune package management functionnalities. It allows users to ping a tar file (gz or bz) using `http` (and `https`) protocol.

Closes #10121

### DIFF
--- a/bin/pkg/pkg_common.ml
+++ b/bin/pkg/pkg_common.ml
@@ -88,9 +88,15 @@ let get_repos repos ~repositories =
     | Some repo ->
       let loc, opam_url = Repository.opam_url repo in
       let module Opam_repo = Dune_pkg.Opam_repo in
-      (match Dune_pkg.OpamUrl.local_or_git_only opam_url loc with
+      (match Dune_pkg.OpamUrl.local_or_git_or_tar_only opam_url loc with
        | `Git -> Opam_repo.of_git_repo loc opam_url
-       | `Path path -> Fiber.return @@ Opam_repo.of_opam_repo_dir_path loc path))
+       | `Path path -> Fiber.return @@ Opam_repo.of_opam_repo_dir_path loc path
+       | `Tar ->
+         User_error.raise
+           ~loc
+           [ Pp.textf "Repositories store in tarball(%s) are currently unsupported"
+             @@ OpamUrl.to_string opam_url
+           ]))
 ;;
 
 let find_local_packages =
--- a/bin/pkg/pkg_common.ml
+++ b/bin/pkg/pkg_common.ml
@@ -94,7 +94,7 @@ let get_repos repos ~repositories =
        | `Tar ->
          User_error.raise
            ~loc
-           [ Pp.textf "Repositories store in tarball(%s) are currently unsupported"
+           [ Pp.textf "Repositories stored in archives (%s) are currently unsupported"
              @@ OpamUrl.to_string opam_url
            ]))
 ;;
--- a/src/dune_pkg/mount.ml
+++ b/src/dune_pkg/mount.ml
@@ -29,14 +29,14 @@ let of_opam_url loc url =
     >>= (function
      | Error message_opt ->
        let message =
-         match message_opt with
-         | Some message -> message
-         | None ->
-           User_message.make
-             [ Pp.textf
-                 "Failed to retrieve source archive from: %s"
-                 (OpamUrl.to_string url)
-             ]
+         Option.value
+           ~default:
+             (User_message.make
+                [ Pp.textf
+                    "Failed to retrieve source archive from: %s"
+                    (OpamUrl.to_string url)
+                ])
+           message_opt
        in
        raise (User_error.E message)
      | Ok output ->
--- a/src/dune_pkg/mount.ml
+++ b/src/dune_pkg/mount.ml
@@ -24,7 +24,12 @@ let of_opam_url loc url =
     in
     Git rev
   | `Tar ->
-    let dir = Path.Build.(L.relative root [ "_fetch"; "url" ]) |> Path.build in
+    (* To prevent cache dir from growing too much, `/tmp/` stores the archive
+       when running `dune pkg lock`. We download and extract the archive
+       everytime the command runs.
+       CR-someday maiste: downloading and extracting should be cached to be
+       reused by the `dune build` command. *)
+    let dir = Temp.(create Dir ~prefix:"dune" ~suffix:"fetch-pinning") in
     Source.fetch_archive_cached (loc, url)
     >>= (function
      | Error message_opt ->
@@ -39,12 +44,18 @@ let of_opam_url loc url =
            message_opt
        in
        raise (User_error.E message)
-     | Ok output ->
+     | Ok archive ->
        let target =
-         let file_digest = Path.to_string output |> Digest.file |> Digest.to_hex in
+         let file_digest = Path.to_string archive |> Digest.file |> Digest.to_hex in
          Path.relative dir file_digest
        in
-       let+ path = Tar.load_or_untar ~target ~archive:output in
+       let+ path =
+         Tar.extract ~archive ~target
+         >>| function
+         | Error () ->
+           User_error.raise [ Pp.textf "unable to extract %S" (Path.to_string target) ]
+         | Ok () -> target
+       in
        Path path)
 ;;
 
--- a/src/dune_pkg/mount.ml
+++ b/src/dune_pkg/mount.ml
@@ -41,7 +41,7 @@ let of_opam_url loc url =
        raise (User_error.E message)
      | Ok output ->
        let target =
-         let file_digest = Digest.file (Path.to_string output) |> Digest.to_hex in
+         let file_digest = Path.to_string output |> Digest.file |> Digest.to_hex in
          Path.relative dir file_digest
        in
        let+ path = Tar.load_or_untar ~target ~archive:output in
--- a/src/dune_pkg/mount.ml
+++ b/src/dune_pkg/mount.ml
@@ -11,7 +11,7 @@ let backend t = t
 
 let of_opam_url loc url =
   let* () = Fiber.return () in
-  match OpamUrl.local_or_git_only url loc with
+  match OpamUrl.local_or_git_or_tar_only url loc with
   | `Path dir -> Fiber.return (Path dir)
   | `Git ->
     let+ rev =
@@ -23,6 +23,29 @@ let of_opam_url loc url =
       >>| User_error.ok_exn
     in
     Git rev
+  | `Tar ->
+    let dir = Path.Build.(L.relative root [ "_fetch"; "url" ]) |> Path.build in
+    Source.fetch_archive_cached (loc, url)
+    >>= (function
+     | Error message_opt ->
+       let message =
+         match message_opt with
+         | Some message -> message
+         | None ->
+           User_message.make
+             [ Pp.textf
+                 "Failed to retrieve source archive from: %s"
+                 (OpamUrl.to_string url)
+             ]
+       in
+       raise (User_error.E message)
+     | Ok output ->
+       let target =
+         let file_digest = Digest.file (Path.to_string output) |> Digest.to_hex in
+         Path.relative dir file_digest
+       in
+       let+ path = Tar.load_or_untar ~target ~archive:output in
+       Path path)
 ;;
 
 let read t file =
--- a/src/dune_pkg/opamUrl0.ml
+++ b/src/dune_pkg/opamUrl0.ml
@@ -30,10 +30,16 @@ let is_version_control t =
 
 let is_local t = String.equal t.transport "file"
 
-let local_or_git_only url loc =
+let is_tarball t =
+  let supported_compress_format = [ "tar.gz"; "tgz"; "tar.bz2"; "tbz" ] in
+  List.exists ~f:(fun suffix -> String.is_suffix ~suffix t.path) supported_compress_format
+;;
+
+let local_or_git_or_tar_only url loc =
   match (url : t).backend with
   | `rsync when is_local url -> `Path (Path.of_string url.path)
   | `git -> `Git
+  | `http when is_tarball url -> `Tar
   | `rsync | `http | `darcs | `hg ->
     User_error.raise
       ~loc
--- a/src/dune_pkg/opamUrl0.ml
+++ b/src/dune_pkg/opamUrl0.ml
@@ -31,8 +31,10 @@ let is_version_control t =
 let is_local t = String.equal t.transport "file"
 
 let is_tarball t =
-  let supported_compress_format = [ "tar.gz"; "tgz"; "tar.bz2"; "tbz" ] in
-  List.exists ~f:(fun suffix -> String.is_suffix ~suffix t.path) supported_compress_format
+  let supported_compress_format = [ ".tar"; ".tar.gz"; ".tgz"; ".tar.bz2"; ".tbz" ] in
+  List.exists
+    ~f:(fun suffix -> Filename.check_suffix t.path suffix)
+    supported_compress_format
 ;;
 
 let local_or_git_or_tar_only url loc =
--- a/src/dune_pkg/opamUrl0.mli
+++ b/src/dune_pkg/opamUrl0.mli
@@ -15,10 +15,11 @@ val is_version_control : t -> bool
 (** [is_file t] is true iff [t] is a url beginning with "file://" *)
 val is_local : t -> bool
 
-(* [local_or_git_only t loc] returns [`Path p] for a URL pointing to a local
-   file system or [`Git] if it's a git repository (remote or otherwise). If
-   it's neither of those cases, it will error out. *)
-val local_or_git_only : t -> Loc.t -> [ `Path of Path.t | `Git ]
+(* [local_or_git_or_tar_only t loc] returns [`Path p] for a URL pointing to a
+   local file system or [`Git] if it's a git repository (remote or otherwise)
+   or [`Tar] if it's an http server with a tar resource. If it's neither of
+   those cases, it will error out. *)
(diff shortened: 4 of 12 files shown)


bin/pkg/pkg_common.ml
```diff
@@ -88,9 +88,15 @@ let get_repos repos ~repositories =
...
       let module Opam_repo = Dune_pkg.Opam_repo in
-      (match Dune_pkg.OpamUrl.local_or_git_only opam_url loc with
+      (match Dune_pkg.OpamUrl.local_or_git_or_tar_only opam_url loc with
        | `Git -> Opam_repo.of_git_repo loc opam_url
-       | `Path path -> Fiber.return @@ Opam_repo.of_opam_repo_dir_path loc path))
+       | `Path path -> Fiber.return @@ Opam_repo.of_opam_repo_dir_path loc path
+       | `Tar ->
+         User_error.raise
+           ~loc
+           [ Pp.textf "Repositories store in tarball(%s) are currently unsupported"
```
Leonidas-from-XIV, line 97:
```suggestion
           [ Pp.textf "Repositories stored in archives (%s) are currently unsupported"
```

src/dune_pkg/mount.ml
```diff
@@ -23,6 +23,29 @@ let of_opam_url loc url =
       >>| User_error.ok_exn
     in
     Git rev
+  | `Tar ->
+    let dir = Path.Build.(L.relative root [ "_fetch"; "url" ]) |> Path.build in
+    Source.fetch_archive_cached (loc, url)
+    >>= (function
+     | Error message_opt ->
+       let message =
+         match message_opt with
```
Leonidas-from-XIV, line 32: This looks like a good place for `Option.map`.

src/dune_pkg/opamUrl0.ml
```diff
@@ -30,10 +30,16 @@ let is_version_control t =
 
 let is_local t = String.equal t.transport "file"
 
-let local_or_git_only url loc =
+let is_tarball t =
```
Leonidas-from-XIV, line 33: Should this not support zip files as well? From some cursory checking I see opam-repository has about 200 ZIP files and it would be good if we supported that.

src/dune_pkg/opamUrl0.ml
```diff
@@ -30,10 +30,16 @@ let is_version_control t =
...
+  let supported_compress_format = [ "tar.gz"; "tgz"; "tar.bz2"; "tbz" ] in
```
Leonidas-from-XIV, line 34: Probably should also include `"tar"`.

src/dune_pkg/opamUrl0.ml
```diff
@@ -30,10 +30,16 @@ let is_version_control t =
...
+  List.exists ~f:(fun suffix -> String.is_suffix ~suffix t.path) supported_compress_format
```
Leonidas-from-XIV, line 35: If you test with `String.is_suffix` then I think it would be better if the format list included `.` otherwise `"terminaltbz"` would be considered a tarball.

But the better solution is to use `Filename.check_suffix` since it does the right thing on e.g. Windows.

src/dune_pkg/source.ml
```diff
@@ -31,27 +31,29 @@ let to_dyn { url = _loc, url; checksum } =
...
+;;
+
+let fetch_and_hash_archive_cached (url_loc, url) =
+  let open Fiber.O in
+  fetch_archive_cached (url_loc, url)
+  >>| function
+  | Ok target -> Some (Dune_digest.file target |> Checksum.of_dune_digest)
+  | Error message_opt ->
+    let message =
+      match message_opt with
```
Leonidas-from-XIV, line 48: `Option.map`?

test/blackbox-tests/test-cases/pkg/pin-depends.t
```diff
@@ -175,10 +175,18 @@ Pin to an invalid opam file
...
-  $ runtest "http://0.0.0.0/tarball.tgz"
-  File "foo.opam", line 1, characters 0-0:
-  Error: Could not determine location of repository http://0.0.0.0/tarball.tgz
-  Hint: Specify either a file path or git repo via SSH/HTTPS
-  [1]
+  $ mkdir _source/
+  $ cat > _source/bar.opam << EOF
+  > opam-version: "2-0"
+  > EOF
+  $ tar -czf tarball.tgz -C _source bar.opam
```
Leonidas-from-XIV, line 184: Use just `cf` then it creates a tarball without compressing. Saves a few cycles in compression and decompression.

src/dune_pkg/opamUrl0.mli
```diff
@@ -15,10 +15,11 @@ val is_version_control : t -> bool
...
 
-(* [local_or_git_only t loc] returns [`Path p] for a URL pointing to a local
-   file system or [`Git] if it's a git repository (remote or otherwise). If
-   it's neither of those cases, it will error out. *)
-val local_or_git_only : t -> Loc.t -> [ `Path of Path.t | `Git ]
+(* [local_or_git_or_tar_only t loc] returns [`Path p] for a URL pointing to a
+   local file system or [`Git] if it's a git repository (remote or otherwise)
+   or [`Tar] if it's an http server with a tar resource. If it's neither of
+   those cases, it will error out. *)
+val local_or_git_or_tar_only : t -> Loc.t -> [ `Path of Path.t | `Git | `Tar ]
```
Leonidas-from-XIV, line 22: I think I remembered it but can you explain again why `Tar` is not `Tar of Path.t` where the path would be the path to the extracted source?

test/blackbox-tests/test-cases/pkg/pin-depends.t
```diff
@@ -175,10 +175,18 @@ Pin to an invalid opam file
...
+Pin to an HTTP archive work
 
-  $ runtest "http://0.0.0.0/tarball.tgz"
-  File "foo.opam", line 1, characters 0-0:
-  Error: Could not determine location of repository http://0.0.0.0/tarball.tgz
-  Hint: Specify either a file path or git repo via SSH/HTTPS
-  [1]
+  $ mkdir _source/
+  $ cat > _source/bar.opam << EOF
+  > opam-version: "2-0"
```
Leonidas-from-XIV, line 182:
```suggestion
  > opam-version: "2.0"
```

Leonidas-from-XIV (commented):
Thanks. The ZIP file changes can be added in a separate PR if you want, but I think that it would be useful to support it given it is a rather common format, especially on Windows.

src/dune_pkg/mount.ml
maiste: You meant `Option.value` no?

src/dune_pkg/opamUrl0.ml
maiste: I will stick with your top-level suggestion and delay zip for latter. Indeed, contrary to tar it seems to be the code to support `zip` is currently missing.

src/dune_pkg/mount.ml
Leonidas-from-XIV: Ah yes, indeed!

src/dune_pkg/opamUrl0.mli
maiste: Because Tar is transformed to a `Path` where I need to use it, and we compute the path afterward, not in this function as we don't know where the tarball will be extracted yet.

test/blackbox-tests/test-cases/pkg/pin-depends.t
```diff
@@ -175,10 +175,18 @@ Pin to an invalid opam file
...
+  > EOF
+  $ tar -czf tarball.tgz -C _source bar.opam
+  $ echo tarball.tgz > fake-curls
+  $ PORT=1
+  $ runtest "http://0.0.0.0:$PORT/tarball.tgz"
+  Solution for dune.lock:
+  - bar.1.0.0
+  (version 1.0.0)
+  (dev)
+  (source (fetch (url http://0.0.0.0:1/tarball.tgz) (checksum md5=ecfb12cd56789e48aaa4eef289c85c40))) (dev)
```
Leonidas-from-XIV, line 192: I think it would be nice to determine the MD5 of the tarball that was created and make sure that it matches the one in the lock file. Otherwise it's not reproducible.

maiste: I have do it using grep as the tar headers can make the hash different.

test/blackbox-tests/test-cases/pkg/pin-depends.t
```diff
@@ -175,10 +175,17 @@ Pin to an invalid opam file
...
+  > opam-version: "2.0"
+  > EOF
+  $ tar cf tarball.tar -C _source bar.opam
+  $ MD5_CHECKSUM=$(md5sum tarball.tar  | cut -f1 -d' ')
+  $ echo tarball.tar > fake-curls
+  $ PORT=1
+  $ runtest "http://0.0.0.0:$PORT/tarball.tar" > output
+  Solution for dune.lock:
+  - bar.1.0.0
+  $ cat output | grep "md5=$MD5_CHECKSUM" 2>&1 > /dev/null
```
Leonidas-from-XIV, line 191: Nitpick but useless use of `cat`. Besides that I think this is a very good solution.

Leonidas-from-XIV:
@maiste I've looked into the `describe.t` failure and when I run it locally the issue does not happen, so it seems like that might be another CI failure.

src/dune_pkg/mount.ml
```diff
@@ -23,6 +23,29 @@ let of_opam_url loc url =
       >>| User_error.ok_exn
     in
     Git rev
+  | `Tar ->
+    let dir = Path.Build.(L.relative root [ "_fetch"; "url" ]) |> Path.build in
```
rgrinberg, line 27: Why do we put this into _fetch? Is this to attempt to make the build system to use this?

src/dune_pkg/mount.ml
```diff
@@ -23,6 +23,29 @@ let of_opam_url loc url =
...
+                [ Pp.textf
+                    "Failed to retrieve source archive from: %s"
+                    (OpamUrl.to_string url)
+                ])
+           message_opt
+       in
+       raise (User_error.E message)
+     | Ok output ->
+       let target =
+         let file_digest = Digest.file (Path.to_string output) |> Digest.to_hex in
```
rgrinberg, line 44: `Path.to_string |> Digest.file |> Digest.to_hex`

src/dune_pkg/tar.ml
```diff
@@ -46,3 +46,14 @@ let extract ~archive ~target =
     Path.rename target_in_temp target;
     Ok ()
 ;;
+
+let load_or_untar ~target ~archive =
+  match Path.exists target with
```
rgrinberg, line 51: What if the tar file is updated remotely?

rgrinberg (changes requested):
Do we plan to support tarballs on the file system?

src/dune_pkg/mount.ml
maiste: Yes, exactly. Is there a better to do it?

maiste: I try to respect the structure for Dune so it doesn't have to redo the work twice. Isn't it supposed to consider it?

maiste:
> Do we plan to support tarballs on the file system?

No, we expect people to untar directory themselves on the file system and point to the path. I don't think it is worth adding the machinery for this.

> What if the tar file is updated remotely?

If the file is updated remotely:
- you will have to re-run `dune pkg lock` if you clean the `_build` or, 
- it will keep using the version in the `_build` without updating it.  

I don't think we have a proper when to address it. If we checked on every run, people would have to download the tar file every time they run `dune build`. In the current configuration, people will have a checksum error if they do `dune clean && dune build` which is the equivalent of the behavior (good or bad, idk) with opam: you have to run `opam reinstall` to get the changes.

I would expect people to want to be aware when the checksum change, but having to explicitly state when they want to synchronize the checksum. It would prevent them from downloading tar they don't expect to change.

src/dune_pkg/mount.ml
rgrinberg: I don't think that will work. Dune writes a metadata file describing its state of the workspace local cache. If you didn't update this file, it will just delete what you have put in there.

rgrinberg:
I think we should re-download the tarball and construct the checksum every time we run `dune pkg lock`. This is consistent with how we treat all other sources that lack a checksum. Anything else will be bound to create confusion that will have to be solved by reminding everyone to do the equivalent of `opam update`.

You could still avoid re-downloading if you use curl with the correct e-tag incantations.

maiste:
I agree with you, and this is what happens in the current situation. If you run `dune pkg lock`, it will download the tar and update the checksum. What I meant is that it is not going to change if you run `dune build` and the file changed remotely.

Regarding the `etag`, @Leonidas-from-XIV suggested this. However, many of the packages on `opam-repository` live in GitHub Release and from the requests I tried with curl, it does not seem to provide an `ETag` header. It is still an optimization we can do later.

src/dune_pkg/tar.ml
rgrinberg: I still don't understand the point of this check

rgrinberg:
Okay, thanks. Is there a test that demonstrates that subsequent `dune pkg lock` re-download it? We should have one.

maiste:
I'm rewriting my test to show the behavior. It should be ready for a next release round by the end of the day :+1:

src/dune_pkg/mount.ml
Leonidas-from-XIV: But isn't it enough to keep it there temporarily to prevent `dune pkg lock && dune build` from downloading the same tarball twice? With a temporary directory `pkg lock` would download the tarball there once and then `build` would download the tarball again, since it does not know about the previous temporary directory.

test/blackbox-tests/test-cases/pkg/pin-depends.t
```diff
@@ -175,10 +175,42 @@ Pin to an invalid opam file
...
+  6 |   (checksum md5=aaaaaaaaaaaaaaaaaaaaaaaaaaaaa)))
+                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  Error: "md5=aaaaaaaaaaaaaaaaaaaaaaaaaaaaa" is an invalid Cryptographic hash
+  of a package.
   [1]
+
+Pin to an HTTP archive needs `dune pkg lock` to download and compute the hash
+of the target again
+
+  $ rm tarball.tar already-served
```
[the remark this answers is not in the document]
maiste: I need to keep it this way. Otherwise, it means I change the URL which is not correct because I try to demonstrate that for the same URL, `dune pkg lock` will update the checksum...

src/dune_pkg/mount.ml
rgrinberg: Did you try to see if it works? I don't think it will because dune needs an internal data structure updated to recognize anything in the _build directory (see `Rule_cache.Workspace_local`).

It would be nice to have it work the way you're saying, but I think you need to fetch the action to be aware of this bag of already downloaded files you can draw from.

src/dune_pkg/tar.ml
rgrinberg: How do you know it's the same file though? If you're relying on the digest, you should at least check that.

maiste: Here, the name of the target (a directory) is the same as the digest of the archive (compute above). I have moved the code directly where it is used as this function would be a source of errors otherwise.

test/blackbox-tests/test-cases/pkg/pin-depends.t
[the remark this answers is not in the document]
Leonidas-from-XIV: Ah, fair enough :+1:

maiste:
It seems that having caching efficient is not really easy for the pinning with http and tar. More specifically, it is hard to cache the fetch for when you need the `opam` file and later when you will need the code. They live in the same directory but `of_opam_rule` needs the tar to present, and it will generate new rules using the opam file.

To reduce the amount of `tar` executed, I can "cache" the file somewhere. This way, if the file is still the same, it won't try to unpack it again. I would like to avoid storing it inside the `~/.cache/dune` dir because there is a risk it grows exponentially in function of the user usage. I was thinking about storing it under `_build/_archive` but I'm really not confident with this. 

Maybe I'm missing something, but it seems there is no way to avoid the double download without changing the logic.

My proposition is to stick to the behavior where it stores the tar into the `_build` and let the engine do the fetch when parsing the `*.pkg` file. It is not optimal but at least the change is not big, and it brings the feature to Dune. 

WDTY?

rgrinberg:
> My proposition is to stick to the behavior where it stores the tar into the _build and let the engine do the fetch when parsing the *.pkg file. It is not optimal but at least the change is not big, and it brings the feature to Dune.

IMO you should just write to a temp dir and leave a comment about future improvements. I think placing it in _build just creates a false expectation. You could just fix the re-downloading issue in subsequent PR's. In any case, I don't think this need to block the PR.

rgrinberg (approved):

Leonidas-from-XIV (approved):

PR #11449, opened 2025-02-06
chore(ci): remove opam unsupported compilers

maiste:
Following the [recent updates](https://discuss.ocaml.org/t/opam-repository-archival-phase-2-ocaml-4-08-is-the-lower-bound/15965) on the opam side, the `4.08` is now considered the lower bound. As a result this PR removes two tests on `4.02` and `4.04` that are not relevant anymore, regarding this new policy.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -56,13 +56,6 @@ jobs:
           - ocaml-compiler: 4.08.x
             os: ubuntu-latest
             skip_test: true
-          - ocaml-compiler: 4.04.x
-            os: ubuntu-latest
-            skip_test: true
-            configurator: true
-          - ocaml-compiler: 4.02.x
-            os: ubuntu-latest
-            skip_test: true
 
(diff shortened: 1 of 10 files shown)


rgrinberg:
Could you bump the constraints in our packages as well? No reason to claim that we support versions that we don't test.

Leonidas-from-XIV:
I've added a commit to remove some other references to not accidentally signal that we support <4.08 when in fact we don't. Feel free to remove the commit if you don't think it is relevant.

There's also some code in `src/ocaml-config/ocaml_config.ml` that could be possibly dropped, WDYT? https://github.com/ocaml/dune/blob/271de14764efa11c49be2b6ffbab119a47b07965/src/ocaml-config/ocaml_config.ml#L551-L555

Leonidas-from-XIV:
Yes but from that I would assume that the other branch is not anymore, since we don't support any versions before 4.06. But I wasn't sure so I haven't touched it.

rgrinberg (approved):

PR #11451, opened 2025-02-06
fix: add missing thread flag to support Mutex

### LINKED ISSUE #11450: ci: regular failure of 4 tests
Since the introduction of #11302, some tests seem to fail with the following errors:
```
Unbound module Mutex
```
This PR indeed introduces a call to a `Mutex` module in _vendor/opam-file-format_.

This issue concerns the following tests:
- Coq 8.16.1
- Wasm_of_ocaml
- Build monorepo benchmark docker image
- Build (ocaml-base-compiler.4.14.2,system-msvc, windows-latest

The common thing is that they all worked on `4.14.X`. However, I have check [in the manual](https://ocaml.org/manual/4.14/api/Mutex.html), the `module Mutex` is available.

Does it require to be added to the deps as Unix for this version? Any ideas about this @Leonidas-from-XIV, @gridbugs and @rgrinberg?

maiste:
This PR fixes the `Unbound module Mutex` by adding the `-thread` flag to the compiler. It is not necessary (but harmless) with `5.X` version but required for `4.X`.

Closes #11450

### DIFF
--- a/vendor/opam-file-format/dune
+++ b/vendor/opam-file-format/dune
@@ -1,6 +1,7 @@
 (library
  (name opam_file_format)
- (wrapped false))
+ (wrapped false)
+ (flags :standard -thread))
 
 (ocamllex opamLexer)
 (ocamlyacc opamBaseParser)


rgrinberg:
I think we can just drop this mutex. I was begin defensive because I didn't know the code and couldn't confirm we weren't parsing outside the main thread. Now I'm pretty confident it's safe to drop this mutex.

PR #11453, opened 2025-02-06
refactor(pkg): unrefutable constraints

rgrinberg:
Reduce loading of unnecessary opam packages by making sure that every local package must not conflict with the constraints of local packages. Such packages can never be used for a solution, and loading them is pointless.

This is a large speed improvement for repos that write many constraints (such as bonsai for example).

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -46,6 +46,8 @@ module Priority = struct
     let version = OpamFile.OPAM.package package |> OpamPackage.version in
     { version; avoid }
   ;;
+
+  let rejected version = { avoid = true; version }
 end
 
 module Context = struct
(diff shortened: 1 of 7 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -1869,15 +1938,35 @@ let solve_lock_dir
...
+        (Context.create
+           ~pinned_packages
+           ~solver_env
+           ~repos
+           ~version_preference
+           ~local_packages:local_packages'
+           ~stats_updater
+           ~constraints)
+    and local_packages' =
+      lazy
```
rgrinberg, line 1952: All of this "laziness" is just to allow the mutually recursive definition of Context.t and the local packages which require the context to evaluate their constraints.

gridbugs (approved):

PR #11479, opened 2025-02-12
fix(ci): build doc with LC_ALL=C

emillon:
The installed locales in the github runner can act weirdly and in
particular `C.UTF-8` has a different name; downgrading to `LC_ALL=C`
fixes the build.


maiste (approved):
Thanks!

Issue #11484, opened 2025-02-15
OCAML_DUNE_PRETEND_VERSION


SnarkBoojum:
I'm the Debian packager for elpi and coq-elpi, both using dune, and I have been having some issues with their latest versions.

The elpi tarball on github relies on dune using it's .git to determine the version (%%VERSION_NUM%% substitutions and building a coherent META file) ; unfortunately, it is a .git-less tarball, so it doesn't quite work as expected.

For packagers (I guess other distributions will also find it hard), it would be nice to have an environment variable to bypass the current mechanism to find the version, and I propose OCAML_DUNE_PRETEND_VERSION ; the usage being:

- if it's defined, take that as a basis for version and version_num everywhere ;
- if it isn't, fall back to the current git/hg tricks.

Such a mechanism isn't unheard of: python's setuptools-scm system provides a SETUPTOOLS_SCM_PRETEND_VERSION which does exactly that ; and if you look at the debian/rules files [here](https://codesearch.debian.net/search?q=SETUPTOOLS_SCM_PRETEND_VERSION&literal=1&perpkg=1), you'll see how useful it is.

Thanks

rgrinberg:
It's fine to have this, but why do you want a runtime configuration option. Wouldn't it be better to have a build time option? E.g. `ocaml configure.ml --set-version FOO`?

SnarkBoojum:
I don't get your point : it is a runtime configuration for dune, but it's used at build time for the software currently built -- which is the one we want to pretend the version of.

PR #11485, opened 2025-02-17
Add support for DUNE_VCS_{DESCRIBE,COMMIT_ID,FILES}

glondu:
This commit adds the possibility to customize `dune subst` behaviour with environment variables, instead of querying the VCS.

The need arose in Debian, where unsubstituted sources are preferred as basis for packaging but break substitution when those sources are not used from upstream VCS. For more background, see the discussion starting at [1].

[1] https://lists.debian.org/debian-ocaml-maint/2025/02/msg00106.html

This is work in progress that I started a few hours before @SnarkBoojum opened https://github.com/ocaml/dune/issues/11484, but it is a solution to the problem exposed there.


rgrinberg:
The feature is fine. It needs some tests and documentation though.

rgrinberg:
Please re-open if you intend to finish this PR

PR #11175, opened 2024-12-02
Bump Dune lang version to 3.18

nojb:
See https://github.com/ocaml/dune/pull/11166#issuecomment-2510124098

The second commit fixes the versioning guard of a change that was initially in 3.17 but was later reverted (see #11164).

PR #11486, opened 2025-02-18
Correct version guard for dune formatting change

nojb:
The formatting change introduced in #10892 was guarded under version 3.18 in #11175 to avoid breaking existing users of the unversioned `dune format-dune-file` (see https://github.com/ocaml/dune/pull/10892#issuecomment-2503530595). However, its replacement action `(format-dune-file)` introduced #11166 will only appear in 3.18, so the format change should in fact be guarded under version 3.19 to give users time to adapt.

### DIFF
--- a/src/dune_lang/format.ml
+++ b/src/dune_lang/format.ml
@@ -27,7 +27,7 @@ let print_wrapped_list ~version x =
   if version < (2, 8)
   then Pp.char '(' ++ Pp.hovbox ~indent:1 inner ++ Pp.char ')'
   else
-    (if version < (3, 18) then Pp.hvbox else Pp.hovbox)
+    (if version < (3, 19) then Pp.hvbox else Pp.hovbox)
       ~indent:1
       (Pp.char '(' ++ inner ++ Pp.char ')')
 ;;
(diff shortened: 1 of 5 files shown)


nojb:
> Sounds fair enough, especially as it would be nice to release 3.18 soon-ish. Can you promote the failures?

I promoted some of the failures, but I cannot reproduce the rest in my machine. Can you help me by pushing the corrected test files to this branch? Thanks!

Leonidas-from-XIV:
I was confused about this as well as it worked for me too. But the issue was that I had too old of a `ppxlib` in my `dev-switch` so I've created #11488.

nojb:
> I was confused about this as well as it worked for me too. But the issue was that I had too old of a `ppxlib` in my `dev-switch` so I've created #11488.

Thanks, now almost everything passes. However there's still one failure that looks unrelated. Do you confirm?

Leonidas-from-XIV (approved):
Yes, the Dynlink failure happens (more than) occasionally but is unrelated.

nojb:
Thanks for the confirmation!

PR #11487, opened 2025-02-18
Require js_of_ocaml >= 6

Leonidas-from-XIV:
It works with older versions but the error message changed in version 6, so the test fail due to the wording of the error if the person uses `jsoo.5.9.1`.

This makes sure that people updating their system using `make dev-switch` will get their possibly existing installation updated.


nojb (approved):
LGTM

PR #11488, opened 2025-02-18
Require minimal ppxlib version

Leonidas-from-XIV:
Otherwise the `describe` test is failing because the description does not match.

Similar reasoning as #11487.


nojb (commented):
LGTM

nojb (approved):
LGTM

PR #11490, opened 2025-02-19
Avoid formatting read-only files

Julow:
This follows the discussion in https://github.com/ocaml/dune/pull/11202

Formatting a read-only file might succeed but the promotion step will fail and disturb the workflow. This checks each source file access before setting up the formatting rule.

This happens often with Nix's 'result' links, which might contain .ml files when an OCaml project was built.
Ideally, Dune wouldn't look inside 'result' links as they often contain outdated installables for the local packages but changing that might interfere with symlinks to other dune workspaces.

### DIFF
--- a/otherlibs/stdune/src/fpath.ml
+++ b/otherlibs/stdune/src/fpath.ml
@@ -217,3 +217,11 @@ let traverse =
 let traverse_files ~dir ~init ~f =
   traverse ~dir ~init ~on_dir:(fun ~dir:_ _fname acc -> acc) ~on_file:f
 ;;
+
+let is_rw p =
+  try
+    Unix.access p Unix.[ R_OK; W_OK ];
+    true
+  with
+  | Unix.Unix_error _ -> false
+;;
--- a/src/dune_rules/format_rules.ml
+++ b/src/dune_rules/format_rules.ml
@@ -138,6 +138,9 @@ let gen_rules_output
      let* dialect, kind =
        Path.Source.extension file |> Dialect.DB.find_by_extension dialects
      in
+     (* Avoid setting up rules that will fail during promotion due to the
+        source file not being writable. *)
+     let* () = Option.some_if (Path.Source.is_rw file) () in
      let* () =
        Option.some_if (Format_config.includes config (Dialect (Dialect.name dialect))) ()
      in
--- a/test/blackbox-tests/test-cases/dialects/good.t/run.t
+++ b/test/blackbox-tests/test-cases/dialects/good.t/run.t
@@ -2,6 +2,8 @@ Test the (dialect ...) stanza inside the dune-project file.
 
   $ dune exec ./main.exe
 
+  $ chmod a+w *
+
   $ dune build @fmt
   Formatting main.mf
   Formatting main.mfi
--- a/test/blackbox-tests/test-cases/formatting/feature.t/run.t
+++ b/test/blackbox-tests/test-cases/formatting/feature.t/run.t
@@ -3,6 +3,10 @@ Note about versioning:
 - in (lang dune 1.x), no formatting rules are set up by default.
 - (lang dune 2.0) behaves as if (using fmt 1.2) is set.
 
+Turn every symlinks into writable files belonging to the sandbox as formatting
+rules are not generated for read-only files:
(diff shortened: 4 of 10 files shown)


otherlibs/stdune/src/fpath.ml
```diff
@@ -217,3 +217,11 @@ let traverse =
 let traverse_files ~dir ~init ~f =
   traverse ~dir ~init ~on_dir:(fun ~dir:_ _fname acc -> acc) ~on_file:f
 ;;
+
+let is_rw p =
+  try
+    Unix.access p Unix.[ R_OK; W_OK ];
+    true
+  with
+  | Unix.Unix_error _ -> false
```
Leonidas-from-XIV, line 226: I think

```suggestion
  match Unix.access p Unix.[ R_OK; W_OK ] with
  | () -> true
  | exception Unix.Unix_error _ -> false
```

reads a little better as it separates the thing that might throw from the value it will compute, but I also just really like `match`.

test/blackbox-tests/test-cases/dialects/good.t/run.t
```diff
@@ -2,6 +2,8 @@ Test the (dialect ...) stanza inside the dune-project file.
 
   $ dune exec ./main.exe
 
+  $ chmod a+w *
```
Leonidas-from-XIV, line 5: Could you add a comment on these `chmod`s why this is necessary? Someone unaware of that functionality will probably wonder what is going on.

test/blackbox-tests/test-cases/formatting/feature.t/run.t
```diff
@@ -3,6 +3,10 @@ Note about versioning:
 - in (lang dune 1.x), no formatting rules are set up by default.
 - (lang dune 2.0) behaves as if (using fmt 1.2) is set.
 
+Turn every symlinks into writable files belonging to the sandbox as formatting
+rules are not generated for read-only files:
+  $ find . -type l | while read f; do (rm "$f"; cat > $f) < $f; done
```
Leonidas-from-XIV, line 8: Wouldn't this be easier with `find -exec`?

(OTOH fairly impressed it even works, reading the file and replacing it in one step)

Leonidas-from-XIV (commented):
@Julow You need to sign off these commits to make the DCO check happy :)

(The CI failure about `Dynlink` failing is unrelated)

test/blackbox-tests/test-cases/formatting/feature.t/run.t
Julow: All the files (but not the directories) in the cram test's current directory are symlinks to read-only copies of the source code. This breaks `dune promote` but also breaks any programs that don't do the "write to temp file and move it to its destination" strategy, as opening the file for truncating and writing (eg. with `open_out_bin`) will not override the permissions.

`-exec` will not work because this uses two commands. Reading continue to work after the file is removed, after all the stdin is still a reference to it so it can't be freed yet. `cat > $f` outputs to a new file and doesn't disturb the reading.

test/blackbox-tests/test-cases/dialects/good.t/run.t
Julow: This is for the same reason as https://github.com/ocaml/dune/pull/11490#discussion_r1965202171 in `feature.t`. I wonder if there's a better way to do it ?

otherlibs/stdune/src/fpath.ml
Julow: I also love the `match` syntax more than the `try` :)

src/dune_rules/format_rules.ml
```diff
@@ -138,6 +138,9 @@ let gen_rules_output
      let* dialect, kind =
        Path.Source.extension file |> Dialect.DB.find_by_extension dialects
      in
+     (* Avoid setting up rules that will fail during promotion due to the
+        source file not being writable. *)
+     let* () = Option.some_if (Path.Source.is_rw file) () in
```
rgrinberg, line 143: Does this work in watch mode? Probably not. I think you need to thread this value through the `Source_tree` and `Readdir`.

rgrinberg:
Please re-open once review comments are addressed.

PR #11492, opened 2025-02-21
pkg: fix bug where multiple conflicts causes conlficts to be ignored

### LINKED ISSUE #11265: Package conflicts are ignored when dune-projects contains multiple conflicts
For example, if you have a dune-project with a single conflict it will be respected:
```
(lang dune 3.17)
(package
 (name foo)
 (depends bar)
 (conflicts bar))
```

Dune is unable to find a solution for the above package because it both depends on and conflicts with the package `bar`. However if we add an additional conflict dune will be able to find a solution:

```
(lang dune 3.17)
(package
 (name foo)
 (depends bar)
 (conflicts bar baz))
```

Dune will now happily choose `bar` as part of the package solution despite it appearing in `conflicts`. Note that if `bar` is repeated in conflicts rather than introducing a new conflicting package (ie. `(conflicts bar bar)`) then dune works as expected.

Repro: https://github.com/ocaml/dune/pull/11266

gridbugs:
Conflicts were being combined as a conjunction bet they should have been combined with a disjunction.

Fixes https://github.com/ocaml/dune/issues/11265

### DIFF
--- a/src/dune_pkg/dependency_formula.ml
+++ b/src/dune_pkg/dependency_formula.ml
@@ -2,7 +2,10 @@ open Import
 
 type t = OpamTypes.filtered_formula
 
-let of_dependencies deps = Package_dependency.list_to_opam_filtered_formula deps
+let of_dependencies deps =
+  List.map deps ~f:Package_dependency.to_opam_filtered_formula |> OpamFormula.ands
+;;
+
 let to_filtered_formula v = v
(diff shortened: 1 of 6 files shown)


Leonidas-from-XIV (approved):
This is a great fix, looks very good! I like that it also fixes the behavior for `depexts` for which the same is true.

Issue #11136, opened 2024-11-20
dune tip fails to solve packages with avoid-version


avsm:
With current dune tip, `dune pkg lock` fails due to (I think) avoid-version:

```
% dunex pkg lock
Error: Unable to solve dependencies for the following lock directories:
Lock directory dune.lock:
Can't find all required versions.
Selected: angstrom.0.16.1 arod.dev astring.0.8.5 base.v0.17.1 base-bytes.base
          base-threads.base base-unix.base base64.3.5.1 bigarray-compat.1.1.0
          bigstringaf.0.10.0 bos.0.2.1 bushel.dev cmarkit.0.3.0
          cmdliner.1.3.0 cohttp.5.3.1 csexp.1.5.2 cstruct.6.2.0 ctypes.0.23.0
          digestif.1.2.0 domain-local-await.1.0.1 dune-configurator.3.16.1
          eio.1.1 eio_main.1.1 eio_posix.1.1 eqaf.0.10 ezjsonm.1.3.0
          fmt.0.9.0 fpath.0.7.3 hex.1.5.0 hmap.0.8.1 http.6.0.0~beta2
          integers.0.7.0 iomux.0.3 jekyll-format.0.3.4 jsonm.1.0.2 logs.0.7.0
          lwt-dllist.1.0.1 mtime.2.1.0 num.1.5-1 ocaml.5.4.0 ocaml-config.3
          ocaml-syntax-shims.1.0.0 ocaml_intrinsics_kernel.v0.17.1
          ocamlbuild.0.15.0+dune ocamlfind.1.9.6+dune optint.0.3.0
          parsexp.v0.17.0 ppx_deriving_yaml.0.4.0 ppx_sexp_conv.v0.17.0
          ppxlib_jane.v0.17.2 psq.0.2.1 ptime.1.2.0 re.1.12.0 rresult.0.7.0
          seq.base sexplib.v0.17.0 sexplib0.v0.17.0 sitemap.1.0
          stdlib-shims.0.3.0 stringext.1.6.0 syndic.1.6.1 thread-table.1.0.0
          topkg.1.0.7 uri.4.4.0 uri-sexp.4.4.0 uutf.1.0.3 xmlm.1.4.0
          yaml.3.2.0 yaml-sexp.3.2.0 zotero-translation.dev
          arod&bushel&zotero-translation ocaml ocaml-base-compiler
          ocaml-base-compiler ocaml&base-bytes
- cohttp-eio -> (problem)
[...]

PR #11494, opened 2025-02-21
pkg: minimize number of packages flagged avoid-version

art-w:
`dune pkg lock` currently fails to find a solution when a dependency flagged as `avoid-version` is absolutely necessary (see e.g. issue #11136). PR #10668 disallowed all package versions with the avoid flag, since they can result in unwanted solutions (typically: selecting an unstable version of the compiler, when a stable version would have worked).

We discussed during the dune dev meeting that a command-line flag could be added to toggle that behavior, but I didn't really like it... To provide users with a nice error message with a suggestion to re-run `dune pkg lock --allow-avoid-version`, we would need to know when this flag would help (so why not do it by default?). Furthermore as soon as the avoid-versions are allowed in the search space, the SAT solver starts using them for unnecessary purposes (like picking an unstable compiler): Even though the avoid-version are deprioritized, the SAT solver is greedy and will happily use them to satisfy an earlier choice instead of backtracking. In practice we would like it to only pick the absolutely unavoidable packages, but stick to recommended dependencies otherwise.

This yields a simple plan to minimize the number of avoid-version packages included in the solution:
1. Run the solver with all packages flagged avoid-version disallowed. If that works, then we have an ideal solution and we are done!
2. Otherwise, re-run the solver but with all the avoid-version packages allowed. If that still fails, then there are definitively no solution.
3. But if a solution was found by using the avoid-version packages, then that solution gives us an upperbound on the number of avoid-version packages required, and a lowerbound of at least 1. We can re-run the solver with the goal of finding a better solution with less of those packages, by dichotomy on the bounds and an `at most N` SAT constraint with `lower <= N < upper`. The SAT solver already supported `at_most_one` clauses, so the `at_most N` generalization was nearly available.
  Finally the UI displays a warning for all the unavoidable avoid-version, so that the user is informed and can dig into the issue if they desire.

### DIFF
--- a/bin/pkg/pkg_common.ml
+++ b/bin/pkg/pkg_common.ml
@@ -105,13 +105,19 @@ let find_local_packages =
   >>| Package.Name.Map.map ~f:Dune_pkg.Local_package.of_package
 ;;
 
-let pp_packages packages =
-  Pp.enumerate
-    packages
-    ~f:(fun { Lock_dir.Pkg.info = { Lock_dir.Pkg_info.name; version; _ }; _ } ->
-      Pp.verbatim
-        (Package_name.to_string name ^ "." ^ Dune_pkg.Package_version.to_string version))
+let pp_package { Lock_dir.Pkg.info = { Lock_dir.Pkg_info.name; version; avoid; _ }; _ } =
+  let warn =
+    if avoid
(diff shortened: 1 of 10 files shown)


gridbugs (approved):
Nice work!

PR #11497, opened 2025-02-24
pkg: minor fix is sys_poll library

gridbugs:
`Io.String_path.lines_of_file` raises `Sys_error` rather than `Unix_error`, so the way we were handling errors in sys_poll would not correctly handle cases where files are missing. Since `Sys_error` does not provide information about the nature of the error, explicitly check that the file exists and is a regular file rather than relying on exceptions.

### DIFF
--- a/src/dune_pkg/sys_poll.ml
+++ b/src/dune_pkg/sys_poll.ml
@@ -98,7 +98,7 @@ let os ~path =
 let maybe_read_lines p =
   match Io.String_path.lines_of_file p with
   | s -> Some s
-  | exception Unix.Unix_error (Unix.ENOENT, _, _) -> None
+  | exception Sys_error _ -> None
 ;;
 
 let os_release_fields () =


src/dune_pkg/sys_poll.ml
```diff
@@ -96,9 +96,9 @@ let os ~path =
 ;;
 
 let maybe_read_lines p =
-  match Io.String_path.lines_of_file p with
-  | s -> Some s
-  | exception Unix.Unix_error (Unix.ENOENT, _, _) -> None
+  if Sys.file_exists p && Sys.is_regular_file p
+  then Some (Io.String_path.lines_of_file p)
+  else None
```
Leonidas-from-XIV, line 101: Wouldn't it be better to catch the `Sys_error` and then do the checks? That way the `stat` calls only happen on the exception case and don't need to be paid on every invocation of the function.

gridbugs: Good point. I've changed it to optimistically read lines and to return `None` if `Sys_error` is raised. I'm not sure if I should keep the checks (`Sys.file_exists p && Sys.is_regular_file p`) in the exception case or just return `None` in response to all `Sys_error` exceptions.

Leonidas-from-XIV: `Sys.is_regular_file` only exists starting with 5.1, you'd need to use `stat` instead, but for now I'd just leave it out until it proves to be a problem.

Leonidas-from-XIV (approved):

PR #11498, opened 2025-02-24
pkg: handle opam string interp special case

gridbugs:
Handle the opam variable string interpolation syntax necessary for packages with a '+' in their name. Read more at
https://opam.ocaml.org/doc/Manual.html#Variables.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -1497,8 +1497,33 @@ let opam_variable_to_slang ~loc packages variable =
             Blang.Expr (convert_with_package_name package_name))))
 ;;
 
+(* Handles the special case for packages whose names contain '+' characters
+   where a special form of string interpolation is used. From the opam manual:
+   Warning: if the package name contains a + character (e.g. conf-g++), their
+   variables may only be accessed using opam 2.2 via string interpolation,
+   with the following syntax:
+
+     "%{?conf-g++:your-variable:}%"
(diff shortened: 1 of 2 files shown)


src/dune_pkg/opam_solver.ml
```diff
@@ -1452,8 +1452,33 @@ let opam_variable_to_slang ~loc packages variable =
...
+   with the following syntax:
+
+     "%{?conf-g++:your-variable:}%"
+*)
+let desugar_special_string_interpolation_syntax
+      ((packages, variable, string_converter) as fident)
+  =
+  match string_converter with
+  | Some (package_and_variable, "")
+    when List.is_empty packages && OpamVariable.to_string variable |> String.is_empty ->
```
maiste, line 1468:
```suggestion
    when List.is_empty packages && String.is_empty (OpamVariable.to_string variable) ->
```
Minor suggestion, but it found is pretty hard to parse the function in this context.

gridbugs: Done!

test/blackbox-tests/test-cases/pkg/package-name-with-plus-character.t
```diff
@@ -0,0 +1,22 @@
+
```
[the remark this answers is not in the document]
gridbugs: Done!

maiste (approved):

PR #11499, opened 2025-02-25
chore(ci): disable test in CI

maiste:
This PR is an attempt to fix the broken CI test. None of the maintainers is able to reproduce the bug locally. As a result, this PR only disabled the when it detects the CI environment variable set by GitHub Actions (see [the documentation](https://docs.github.com/fr/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables)).


Leonidas-from-XIV (approved):
Seems to solve the issue and better than my initial idea of removing the test altogether.

PR #11502, opened 2025-02-25
doc: clarify usage of sites with implicit_transitive_deps

kentookura:


doc/sites.rst
```diff
@@ -96,7 +96,9 @@ site using the :doc:`generate_sites_module stanza
     (sites mygui))
 
 The generated module `mysites` depends on the library `dune-site` provided by
-Dune.
+Dune. If `implicit_transitive_deps` is set to false in your `dune-project`
```
maiste, line 99: I think it would be helpful to have a reference to the `implicit_transitive_deps` stanza in the reference manual.

maiste (approved):
A nitpicky point but LGTM.

kentookura:
> A nitpicky point but LGTM.

Sure, but if it saves someone the 15 minutes I spent scratching my head, it is worth it :)

doc/sites.rst
```diff
@@ -96,7 +96,10 @@ site using the :doc:`generate_sites_module stanza
     (sites mygui))
 
 The generated module `mysites` depends on the library `dune-site` provided by
-Dune.
+Dune. If :doc:`implicit_transitive_deps
+</reference/dune-project/implicit_transitive_deps>` is set to false in your
+`dune-project` file, `dune-site` must be explicitly declared in the `depends`
+stanza of your package.
```
Leonidas-from-XIV, line 102: Dune never uses `depends` itself so why must it be declared?

I think a better wording might be to state that `dune-site` should be added as a dependency if it is used (and that is true for opam files as well as the `depends` field).  `implicit_transitive_deps true` makes it work accidentally, but the dependency declaration is still wrong.

Leonidas-from-XIV (changes requested):

PR #11503, opened 2025-02-25
Preserve flags ordering in Merlin configuration files.

### LINKED ISSUE #1900: Modules opened in the wrong order?
I get an error when I edit file [link.ml](https://github.com/ocsigen/js_of_ocaml/blob/master/compiler/lib-wasm/link.ml) even though Dune compiles it just fine.
Image

When compiling the library `Wasm_of_ocaml_compiler` which contains this file, I also open the library `Js_of_ocaml_compiler`.  Both libraries contain a file `generate.ml`.
```
(library
 (name wasm_of_ocaml_compiler)
 (libraries js_of_ocaml_compiler)
 (flags
  (:standard -w -7-37 -safe-string -open Js_of_ocaml_compiler))
[...]
```

 I suspect what happens is that merlin and dune do not perform the opens in the same order.

Adding an explicit open flag, even in the wrong order, seems to work around the issue.
```
 (flags
  (:standard -w -7-37 -safe-string -open Wasm_of_ocaml_compiler -open Js_of_ocaml_compiler))
```

I have taken a [dump of the configuration](https://github.com/user-attachments/files/18911578/config.txt).

voodoos:
Dune generated open flags for wrappers are passed last to the compiler, but they are preceding the user-defined flags in the generated Merlin configuration. This means that opens are not applied in the same order by the compiler and Merlin, leading to inconsistencies such as the one described in ocaml/merlin#1900

This PR restores the correct order.

Fixes ocaml/merlin#1900


src/dune_rules/merlin/merlin.ml
```diff
@@ -214,26 +214,30 @@ module Processed = struct
...
-      | opens ->
-        make_directive
-          "FLG"
-          (Sexp.List (Ocaml_flags.open_flags opens |> List.map ~f:(fun x -> Sexp.Atom x)))
-        :: flags
+      match flags with
+      | [] -> other_flags
+      | flags ->
+        make_directive "FLG" (Sexp.List (List.map ~f:(fun s -> Sexp.Atom s) flags))
+        :: other_flags
```
Leonidas-from-XIV, line 240: I find this code somewhat hard to read since each match defaults to the one above. Wouldn't it be easier to handle them separately and do a list concat? That way the order of how the flags are ordered is visible in one line instead of having to trace through all of the matches.

Leonidas-from-XIV (commented):
Thanks, looks like a very useful fix to have. Can you please add a changelog entry as well?

src/dune_rules/merlin/merlin.ml
voodoos: Thanks for the review @Leonidas-from-XIV I now filter a list of options in [956c989](https://github.com/ocaml/dune/pull/11503/commits/956c9894964e9a1d856d19e28af03c9f17ed9211).

Leonidas-from-XIV (approved):
Yeah, this is a very nice solution. Thanks for the contribution!

PR #11504, opened 2025-02-25
Display Windows `NTSTATUS` exit codes in hex

MisterDA:
On Windows, "negative" exit codes are probably `NTSTATUS` values. For example, if a program accesses an invalid memory location, Unix sends a `SIGSEGV` signal which, if unhandled, will terminate the process (setting some kind of non-zero exit code - for example, Linux sets the exit code to 128 + signal number to give a fairly memorable 139). In the equivalent scenario, Windows throws an `EXCEPTION_ACCESS_VIOLATION` which, if handled by the default exception handler, will terminate the process with exit code `STATUS_ACCESS_VIOLATION`. These codes are large negative numbers, which are not terribly memorable in decimal, so for negative exit codes we instead display them in hexadecimal as `0xc0000005` is slightly more memorable than `-1073741819`.

### DIFF
--- a/doc/changes/11504.md
+++ b/doc/changes/11504.md
@@ -1,0 +1,2 @@
+- Display negative error codes on Windows in hex which is the more customary
+  way to display `NTSTATUS` codes (#11504, @MisterDA)
--- a/src/dune_engine/process.ml
+++ b/src/dune_engine/process.ml
@@ -652,7 +652,15 @@ end = struct
                (* If the command has no output, we need to say something.
                   Otherwise it's not clear what's going on. *)
                (match error with
-                | Failed n -> [ Pp.textf "Command exited with code %d." n ]
+                | Failed exitcode ->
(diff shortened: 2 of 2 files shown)


Leonidas-from-XIV (commented):
Looks reasonable to me, can you sign the DCO?

Leonidas-from-XIV (approved):

maiste (approved):

maiste:
@MisterDA, I didn't test but my bet would be that it is in the [verbose](https://github.com/ocaml/dune/blob/9c562873245455f482ebca519b3fdd91d0918c42/src/dune_engine/process.ml#L556) function. Maybe by modifying [this variable](https://github.com/ocaml/dune/blob/9c562873245455f482ebca519b3fdd91d0918c42/src/dune_engine/process.ml#L573) you can get the verbose output correct.

PR #11507, opened 2025-02-28
chore: Fix wasm_of_ocaml in CI

Leonidas-from-XIV:
It doesn't need to be pinned anymore and the current development version fails. Better use a released version for stability.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -211,16 +211,8 @@ jobs:
         working-directory: ./dune
         run: opam pin add -n dune . --with-version 3.17.0
 
-      - name: Checkout Wasm_of_ocaml
-        uses: actions/checkout@v4
-        with:
-          repository: ocsigen/js_of_ocaml
-          path: wasm_of_ocaml
-
       - name: Install Wasm_of_ocaml
-        working-directory: ./wasm_of_ocaml
         run: |
-           opam pin add -n . --with-version `< VERSION`
            opam install wasm_of_ocaml-compiler
 
       - name: Set Git User
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -164,9 +164,6 @@ jobs:
         with:
           token: ${{ github.token }}
 
-      - name: Install Ninja
-        run: sudo apt-get install ninja-build
-
       - name: Checkout Code
         uses: actions/checkout@v4
         with:
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -159,43 +159,13 @@ jobs:
         with:
           node-version: latest
 
-      - name: Restore Cached Binaryen
-        id: cache-binaryen
-        uses: actions/cache/restore@v4
+      - name: Set-up Binaryen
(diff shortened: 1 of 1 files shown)


hhugo:
I think you can also simplify all the rule for binaryen with

```
      - name: Set-up Binaryen
        uses: Aandreba/setup-binaryen@v1.0.0
        with:
          token: ${{ github.token }}

      - name: Install faked binaryen-bin package
        # It's faster to use a cached version
        run: opam install --fake binaryen-bin
```

It should speedup CI a bit

Leonidas-from-XIV:
@hhugo Thanks, a faster CI run is definitely welcome! I have no idea about binaryen so I applied your suggestions as-is. Please take a look and approve the PR if it looks sensible to you.

.github/workflows/workflow.yml
```diff
@@ -159,44 +159,14 @@ jobs:
...
-        uses: actions/checkout@v4
-        with:
-          repository: WebAssembly/binaryen
-          path: binaryen
-          submodules: true
-          ref: version_119
+          token: ${{ github.token }}
 
       - name: Install Ninja
-        if: steps.cache-binaryen.outputs.cache-hit != 'true'
```
hhugo, line 179: Ninja is no longer needed

hhugo:
@vouillon, any reason to have the CI test with ocaml 4.14, and not, say, 5.3

vouillon:
I don't think there is any good reason for that now. I think we didn't support recent versions of OCaml at the time.

vouillon:
By the way, the fix is to rename the two occurrences of `$jsPrint` into `$print` in each of these two files.
https://github.com/ocaml/dune/blob/94841d78867e720d6113aa734850d0b8c615a6d7/test/blackbox-tests/test-cases/wasmoo/no-check-prim.t/lib/runtime.wat#L1-L7
https://github.com/ocaml/dune/blob/94841d78867e720d6113aa734850d0b8c615a6d7/test/blackbox-tests/test-cases/wasmoo/simple.t/lib/runtime.wat#L1-L7
The issue is that we are now rewriting `(func (export "jsPrint") ...)` into `(func $jsPrint (export "jsPrint") ...)` since this provides a better debugging experience to give internal names to all functions (and cumbersome to provide both an internal name and an exported name). But then, we have two functions with the same name in theses modules...

Maybe I should make a separate PR to fix that?

.github/workflows/workflow.yml
Leonidas-from-XIV: Very good, removed.

Leonidas-from-XIV:
> I don't think there is any good reason for that now.

I would suggest doing this in a separate PR (we mainly use 5.1.1 for development, could be useful to have it match). I noticed that this issue was causing a lot of other PRs to fail but I'd rather unblock this quickly and then do the version bump later.

> Maybe I should make a separate PR to fix that?

Yes please, that would definitely be welcome. :)

hhugo (approved):

hhugo:
@vouillon, don't we also  need to add a hint to install wasm_of_ocaml-compiler the binary is missing? (Now that it is released)

Issue #10898, opened 2024-09-10
Unify package build progress messages with regular dune build progress messages


gridbugs:
Currently when building packages with dune, progress messages are always reported in a similar style to the messages printed by `dune build --progress=short`. Package build progress messages are disabled by default and can be enabled by setting the environment variable `DUNE_CONFIG__PKG_BUILD_PROGRESS=enabled`.

Some example messages:
```
    Building ocaml-base-compiler.5.2.0
    Building ocaml-config.3      
    Building ocaml.5.2.0           
    Building base-unix.base        
    Building csexp.1.5.2    
```

There are two issues with this.

1. Package build progress messages don't respect the `--display` argument; they look similar to `--display=short` but print like this even when dune is passed `--display=quiet` (which is the default).
2. When building with `--display=short`, the regular build messages interleave with the package build progress messages in a way that mixes levels of abstractions. For example:
```
    Building cmdliner.1.3.0        
        make default/.pkg/ocamlbuild/target [_private]
    Building topkg.1.0.7           
        make default/.pkg/cmdliner/target [_private]
        make default/.pkg/cmdliner/target [_private]
        make default/.pkg/cmdliner/target [_private]
        dune default/.pkg/base/target [_private]
       ocaml default/.pkg/topkg/target [_private]
    Building stdlib-shims.0.3.0     
    Building re.1.12.0              
    Building ocaml-syntax-shims.1.0.0
    Building astring.0.8.5          
    Building fmt.0.9.0              
    Building uutf.1.0.3             
    Building uucp.15.1.0            
        dune default/.pkg/stdlib-shims/target [_private]
        dune default/.pkg/ocaml-syntax-shims/target [_private]
        dune default/.pkg/re/target [_private]
       ocaml default/.pkg/uutf/target [_private]
       ocaml default/.pkg/fmt/target [_private]
       ocaml default/.pkg/astring/target [_private]
       ocaml default/.pkg/uucp/target [_private]
```

In this example the lines beginning with `Building` refer to a package being built, while the lines beginning with `make`, `dune` and `ocaml` refer to commands being executed during the build. It's not obvious that displaying both types of messages at once is helpful to users.

[...]

PR #11510, opened 2025-03-03
config: enable build progress by default

maiste:
This PR is mostly to open the discussion about it. Considering this feature flag was there for a while, it might be a good time (release of 3.18) to consider this feature as mature. It will display the stage for building packages as you would have for `cargo` if you are under a package management context.

WDYT?


gridbugs:
Before this is enabled by default we should address https://github.com/ocaml/dune/issues/10898.

maiste:
I missed this one. I think we can delay this to `3.19` then!

maiste:
Superseded by the PR mentioned above.

punchagan:
> Superseded by the PR mentioned above.

Sorry for not noticing the existing PR! And thanks for catching the duplicate.

PR #11511, opened 2025-03-05
pkg: support for dependencies compressed with zip

gridbugs:
Previously dune assumed that all package source archives are tarballs, possibly compressed with gzip. This change adds supports for zip files which comprise a small number of packages in the opam repository.


src/dune_pkg/archive_type.ml
```diff
@@ -2,18 +2,45 @@ open Stdune
...
+  make
+    ~bin_name:"tar"
+    ~suffixes:[ ".tar"; ".tar.gz"; ".tgz"; ".tar.bz2"; ".tbz" ]
+    ~build_args:(fun ~archive ~target_in_temp ->
+      [ "xf"; Path.to_string archive; "-C"; Path.to_string target_in_temp ])
+;;
+
+let zip =
+  make ~bin_name:"unzip" ~suffixes:[ ".zip" ] ~build_args:(fun ~archive ~target_in_temp ->
+    [ Path.to_string archive; "-d"; Path.to_string target_in_temp ])
```
Leonidas-from-XIV, line 33: Does that mean that to run it we need both `tar` and `unzip` in the path, even if we never use e.g. `unzip`?

Maybe it would be better to conditionally extend the list of supported suffixes depending on what binaries we do find and only fail if we encounter a suffix we don't support? That way Dune distribution packagers don't have to declare a (non-optional) dependency on `unzip`.

src/dune_pkg/archive_type.ml
```diff
@@ -2,18 +2,45 @@ open Stdune
...
+       | Some x -> x
+       | None -> Dune_engine.Utils.program_not_found bin_name ~loc:None)
+  in
+  { bin; suffixes; build_args }
+;;
+
+let tar =
+  make
+    ~bin_name:"tar"
+    ~suffixes:[ ".tar"; ".tar.gz"; ".tgz"; ".tar.bz2"; ".tbz" ]
```
Leonidas-from-XIV, line 26: Come to think of it, support for `.gz` is dependant on `gunzip` (which is admittedly so common we can probably assume that everyone with `tar` has it) but `bunzip2` is getting rarer these days and according to [this issue](https://github.com/ocaml/opam/issues/6171) OPAM also supports `unxz`.

(As far as I know GNU `tar` is just shelling out to those internally, libarchive's `bsdtar` might actually decompress them using libarchive)

src/dune_pkg/fetch.ml
```diff
@@ -123,8 +123,17 @@ type failure =
...
-let unpack_tarball ~target ~archive =
-  Tar.extract ~archive ~target
+let archive_type_of_opam_url opam_url =
+  let url_string = OpamUrl.to_string opam_url in
+  match Archive_type.choose_for_filename url_string with
+  | Some archive_type -> archive_type
+  | None ->
+    (* Assume that an archive is a tarball if the extension is omitted. *)
+    Archive_type.tar
+;;
```
maiste, line 133: Maybe this should live in OpamUrl? I'm asking because the PR needs to also update [this file](https://github.com/ocaml/dune/blob/fc3996b27ac00a9711dae16379de72a74741fe01/src/dune_pkg/opamUrl0.ml#L33) because otherwise it won't be able to use http pinning with zip.

src/dune_pkg/mount.ml
```diff
@@ -50,7 +50,7 @@ let of_opam_url loc url =
          Path.relative dir file_digest
        in
        let+ path =
-         Tar.extract ~archive ~target
+         Archive_type.extract Archive_type.tar ~archive ~target
```
maiste, line 53:
```suggestion
         Archive_type.(extract tar ~archive ~target)
```
Proposition, but a matter of taste here.

maiste (commented):
Thanks for the PR. I think you need to also update [this file ](https://github.com/ocaml/dune/blob/fc3996b27ac00a9711dae16379de72a74741fe01/src/dune_pkg/opamUrl0.ml#L33) to support http + zip pinning.

src/dune_pkg/archive_type.ml
gridbugs: Looking up the named binary in PATH is lazy and only forced in `extract`, so systems without `unzip` will only fail when trying to unzip a .zip archive.

src/dune_pkg/archive_type.ml
gridbugs: If we separate the questions of "What command does dune need to run to extract an archive?" and "What programs need to be in PATH to extract an archive?", dune currently only handles the first, and will fail with an (often helpful) error if one of the required programs is missing. We could try to improve the ergonomics of this by making it more clear what external tools are needed to handle each type of archive, but that should happen in a different change from this one.

src/dune_pkg/archive_type.ml
gridbugs: I double-checked this by building a project with tar.gz dependencies only on a docker container without `unzip` installed.

gridbugs:
> Thanks for the PR. I think you need to also update [this file ](https://github.com/ocaml/dune/blob/fc3996b27ac00a9711dae16379de72a74741fe01/src/dune_pkg/opamUrl0.ml#L33) to support http + zip pinning.

Ah yes I definitely should. For some reason yesterday I incorrectly thought this file was vendored from opam and didn't want to mess with it. Will update.

src/dune_pkg/mount.ml
gridbugs: I tend to avoid this style because for example if the `Archive_type` module defined a value named `archive`, then it would be passed as the `~archive` argument rather than the local variable of the same name.

src/dune_pkg/archive_type.ml
Leonidas-from-XIV: Ok, once this is merged I'll probably look into this. That way we can also untangle `tar` from `gzip`/`bzip` or whatever other compressor is used.

src/dune_pkg/archive_type.mli
```diff
@@ -0,0 +1,8 @@
+open Stdune
```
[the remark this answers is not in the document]
gridbugs: In my mind a value of `Archive.t` would refer to an instance of an archive, like a specific zip file, whereas a value of `Archive_type.t` represents zip files in general. Maybe `Archive_driver.t` would be a better name, as it's the thing that knows how to do operations on archives of a particular format?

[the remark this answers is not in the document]
maiste: I like the `Archive_driver` proposal. It makes it a bit clearer.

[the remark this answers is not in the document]
gridbugs: Changed!

src/dune_pkg/fetch.ml
Leonidas-from-XIV: @gridbugs Is this still relevant?

gridbugs: I thought I addressed this but apparently I didn't. I'll fix it today.

gridbugs: To clarify, iirc @maiste's original comment was about duplicating the logic for identifying the type of an archive, which lived in OpamUrl and was duplicated here. I've moved this logic to Archive_driver.

src/dune_pkg/archive_driver.ml
```diff
@@ -0,0 +1,75 @@
...
+    ~bin_name:"tar"
+    ~suffixes:[ ".tar"; ".tar.gz"; ".tgz"; ".tar.bz2"; ".tbz" ]
+    ~build_args:(fun ~archive ~target_in_temp ->
+      [ "xf"; Path.to_string archive; "-C"; Path.to_string target_in_temp ])
+;;
+
+let zip =
+  make ~bin_name:"unzip" ~suffixes:[ ".zip" ] ~build_args:(fun ~archive ~target_in_temp ->
+    [ Path.to_string archive; "-d"; Path.to_string target_in_temp ])
+;;
```
Alizter, line 34: Generally Windows doesn't have the `unzip` binary unlike most Linux and macos distributions. It can however be installed. On the other hand, I believe `tar.exe` can handle `.zip` files.

https://ss64.com/nt/tar.html

Similarly most modern `tar` implementations on other platforms (even freeBSD) support this. I don't even have `unzip` by default on NixOS, so perhaps we should consider letting `tar` do the work here instead.

gridbugs: > Similarly most modern tar implementations on other platforms (even freeBSD) support this.

Experimentally, GNU tar doesn't support zip files. BSD tar (installable on some linux distros in the "libarchive" package) does, as does the tar that comes with macos (which is based on BSD tar according to its manpage). Most linux users will have GNU tar, so we can't rely on tar being able to extract zip files in all cases.

However you make a good point about windows and the BSDs. In these cases (and possibly on macos too) it would make sense to fall back to attempting to use tar for zip files if `unzip` is not in PATH.

gridbugs: I've implemented the behaviour of falling back to tar in the case of zip files as a separate commit. If the general sentiment is that this fallback behaviour is too complex for the first pass at supporting zip files then I'm happy to remove it.

src/dune_pkg/archive_driver.mli
```diff
@@ -0,0 +1,17 @@
...
+type t
+
+(** Driver for tarballs, possibly compressed *)
+val tar : t
+
+(** Driver for zip files *)
+val zip : t
+
+(** Return the driver that can extract a file of a given name. The decision is
+    made based on the file's suffix. *)
```
maiste, line 13:
```suggestion
(** Returns the driver that can extract a file of a given name. The decision is
    made based on the file's suffix. *)
```
Really nitpicky here, but for consistency as you used the `(* Knows [...]` above.

src/dune_pkg/archive_driver.mli
```diff
@@ -0,0 +1,17 @@
...
+val choose_for_filename : string -> t option
+
+(** Extract an archive into a given target directory *)
```
maiste, line 16:
```suggestion
(** [extract t ~archive ~target] extracts an archive from the [archive] path into a given [target] directory. If the [target] path does not exist, it will create it. *)
```
Adding a bit of extra information is good (as it allows having it using ocaml-lsp without having to jump to the definition). This suggestion isn't formatted well, sorry about that.

src/dune_pkg/opamUrl0.mli
```diff
@@ -19,7 +19,10 @@ val is_local : t -> bool
    local file system or [`Git] if it's a git repository (remote or otherwise)
    or [`Tar] if it's an http server with a tar resource. If it's neither of
    those cases, it will error out. *)
-val local_or_git_or_tar_only : t -> Loc.t -> [ `Path of Path.t | `Git | `Tar ]
+val local_or_git_or_archive_only
```
maiste, line 22: You need to update the documentation above, as it still uses the old names.

src/dune_pkg/opamUrl0.ml
```diff
@@ -29,19 +29,13 @@ let is_version_control t =
 ;;
 
 let is_local t = String.equal t.transport "file"
+let is_archive t = Option.is_some (Archive_driver.choose_for_filename t.path)
```
maiste, line 32:
```suggestion
let is_supported_archive t = Option.is_some (Archive_driver.choose_for_filename t.path)
``
We should be more specific that this is because it is an archive that we support.

File: src/dune_pkg/archive_driver.ml [hunk 11]
```diff
@@ -0,0 +1,106 @@
...
+         (* Fall back to using tar to extract zip archives, which is possible in some cases. *)
+         (match
+            (* Test for bsdtar before tar, as if bsdtar is installed then it's
+               likely that the tar binary is GNU tar which can't extract zip
+               archives, whereas bsdtar can. If bsdtar is absent, try using the
+               tar command anyway, as on BSD, MacOS, and Windows systems, the tar
+               command can extract zip archives. *)
+            List.find_map [ "bsdtar"; "tar" ] ~f:which
+          with
+          | Some bin -> { Command.bin; args = tar_args }
```
maiste, line 56: Do you know what would be the error if you try to use `GNUtar` to unzip, and it fails because you are on Linux? I'm asking because it might sound weird to users to get a `tar` error when `unzipping` :confused:

src/dune_pkg/archive_driver.ml
```diff
@@ -0,0 +1,106 @@
...
+  { command; suffixes = [ ".zip" ] }
+;;
+
+let all = [ tar; zip ]
+let output_limit = 1_000_000
+let check_suffix t filename = List.exists t.suffixes ~f:(Filename.check_suffix filename)
+let choose_for_filename filename = List.find all ~f:(fun t -> check_suffix t filename)
+
+let extract t ~archive ~target =
+  let* () = Fiber.return () in
```
maiste, line 71: :question:  _Side question_: do you know why we have to do this? I have seen this in various part of the code, I'm curious about the why...

src/dune_pkg/archive_driver.ml
```diff
@@ -0,0 +1,106 @@
+open Stdune
+module Process = Dune_engine.Process
+open Fiber.O
+
+module Command = struct
+  type t =
+    { bin : Path.t
+    ; args : archive:Path.t -> target_in_temp:Path.t -> string list
```
maiste, line 8: What do you think about `gen_args`? I would expect `args` to be a fixed list not a function.

src/dune_pkg/mount.ml
```diff
@@ -50,7 +50,7 @@ let of_opam_url loc url =
          Path.relative dir file_digest
        in
        let+ path =
-         Tar.extract ~archive ~target
+         Archive_driver.extract Archive_driver.tar ~archive ~target
```
maiste, line 53: Here, I think you need to call `choose_for_filename` too. Indeed, if you point to a `zip` archive, you will not be able to extract the `opam` file from it.

maiste (changes requested):
I added a lot of questions regarding naming: feel free to ignore them if you don't think they improve readability.

However, I request a change because I think there is an issue with `mount.ml` if you try to use it with a zip file. FWIU, in the current situation it will fail to extract the archive if it is a zip, whereas it should work.

src/dune_pkg/opamUrl0.mli
maiste: You are right but I don't have one...

src/dune_pkg/archive_driver.ml
Leonidas-from-XIV: I always assumed this with `let+` is so that a `Monad.return` can be avoided in the end.

src/dune_pkg/archive_driver.ml
```diff
@@ -0,0 +1,106 @@
...
+    lazy
+      (match which "unzip" with
+       | Some bin -> { Command.bin; args = zip_args }
+       | None ->
+         (* Fall back to using tar to extract zip archives, which is possible in some cases. *)
+         (match
+            (* Test for bsdtar before tar, as if bsdtar is installed then it's
+               likely that the tar binary is GNU tar which can't extract zip
+               archives, whereas bsdtar can. If bsdtar is absent, try using the
+               tar command anyway, as on BSD, MacOS, and Windows systems, the tar
```
Leonidas-from-XIV, line 52: I think this is only true for FreeBSD and NetBSD, OpenBSD has a very classic `tar`. But close enough and the reasoning is sound I'd say!

Leonidas-from-XIV (approved):
I agree with @maiste's point, but besides this looks good to me so I'm approving it.

In the end of the day, we want people to have as good of a shot at unpacking the archives, if that means we'll try `unzip`, `bsdtar` and `tar` that's fine.

src/dune_pkg/archive_driver.ml
Alizter: This stops the parts of the fiber from being executed eagerly. If we didn't do this then all those `Lazy.force`s and other stuff would be run before we really run the fiber. The best way to think about it is "from now on, everything that happens happens only when this fiber is run".

src/dune_pkg/opamUrl0.mli
gridbugs: I have renamed in to `classify`.

src/dune_pkg/archive_driver.ml
gridbugs: I did an experiment to test this and the error was:
```
File "dune.lock/tsort.pkg", line 10, characters 7-65:
10 |   (url https://github.com/dmbaturin/ocaml-tsort/archive/2.0.0.zip)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error:
failed to unpack archive downloaded from
https://github.com/dmbaturin/ocaml-tsort/archive/2.0.0.zip
reason:
unable to extract "_build/.temp/dune_d7c5e3_2.0.0.zip/download"
```

It's not great in that it doesn't help the user figure out how to fix the problem (ie. install `unzip`) but it's no worse than the current error message (it is just the current error message). You'd get the same error if `tar` wasn't installed. We should improve this but not in this PR.

src/dune_pkg/archive_driver.ml
gridbugs: I went with `make_args` as there's too much "gen" in dune already ;)

src/dune_pkg/mount.ml
gridbugs: Is it correct to say that this will only be an issue if an opam repository is stored in a .zip archive? It won't have any effect on package sources in zip files, right? I'll change it anyway, but just want to make sure I understand the implication of this.

bin/pkg/pkg_common.ml
```diff
@@ -88,10 +88,10 @@ let get_repos repos ~repositories =
     | Some repo ->
       let loc, opam_url = Repository.opam_url repo in
       let module Opam_repo = Dune_pkg.Opam_repo in
-      (match Dune_pkg.OpamUrl.local_or_git_or_tar_only opam_url loc with
+      (match Dune_pkg.OpamUrl.local_or_git_or_archive_only opam_url loc with
        | `Git -> Opam_repo.of_git_repo loc opam_url
        | `Path path -> Fiber.return @@ Opam_repo.of_opam_repo_dir_path loc path
-       | `Tar ->
+       | `Http_archive ->
```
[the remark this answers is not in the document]
gridbugs: I ended up just calling it `Archive`.

src/dune_pkg/archive_driver.ml
gridbugs: Classic openbsd. I've updated this comment to not claim that all BSD's come with a tar that can extract zip files.

src/dune_pkg/archive_driver.ml
gridbugs: I've created https://github.com/ocaml/dune/issues/11544 for improving error messages. Might be a good first issue for an outreachy intern to tackle.

gridbugs:
@maiste I think I've addressed all your feedback

src/dune_pkg/archive_driver.ml
maiste: Thanks for opening the issue.  My main concern was that the message was stating that `tar` is the problem with a zip file. As the error message is generic, that's fine :ok_hand:

src/dune_pkg/mount.ml
maiste: I wasn't thinking about opam repository, more about when you pin files in that are stored in a zip. In the current (inefficient but functional) version I implemented, it will download the zip file to get the `*.opam` file before generating the correct `.pkg` file associated. If you didn't add the check for the file type, it would prevent pinning projects stored in a `zip`.

Thus, the version that has been merged is correct :+1:

PR #11512, opened 2025-03-05
pkg: handle undefined vars in command filters

gridbugs:
Opam treats undefined variables as false when evaluating command filters. This change wraps command filters in lockfiles with `catch_undefined_var` so dune's behaviour matches.


Leonidas-from-XIV (approved):
I like the solution, it doesn't introduce any semantic changes to `Slang` so that's great.

maiste (approved):

PR #11513, opened 2025-03-06
fix: use build and install instructions from opam on pinning

maiste:
When building instructions with package management, pinning packages relied on dune for all the cases. This PR changes this behavior: if there is an opam file, it extracts the instructions from it. It doesn't change behaviors outside of pinning.

This avoids having pinning packages with no dependencies to OCaml from failing because `ocamlc` is not available (for example).

This PR is necessary for projects where the OPAM file install elements without Dune. Indeed, it would try to execute Dune and fails because there is no dependency to `ocamlc` (because using package management and no deps to OCaml is declared).

### DIFF
--- a/src/dune_pkg/local_package.ml
+++ b/src/dune_pkg/local_package.ml
@@ -21,6 +21,9 @@ type t =
   ; depopts : Package_dependency.t list
   ; pins : pins
   ; loc : Loc.t
+  ; build : OpamTypes.command list
+  ; install : OpamTypes.command list
+  ; dune_build : bool
   }
 
 module Dependency_hash = struct
@@ -56,9 +59,21 @@ module For_solver = struct
     ; depopts : Package_dependency.t list
     ; conflict_class : Package_name.t list
     ; pins : pins
+    ; build : OpamTypes.command list
+    ; install : OpamTypes.command list
     }
 
-  let to_opam_file { name; dependencies; conflicts; conflict_class; depopts; pins = _ } =
+  let to_opam_file
+        { name
+        ; dependencies
+        ; conflicts
+        ; conflict_class
+        ; depopts
+        ; pins = _
+        ; build
+        ; install
+        }
+    =
     (* CR-rgrinberg: it's OK to ignore pins here since the solver doesn't touch
        them *)
     OpamFile.OPAM.empty
@@ -72,6 +87,8 @@ module For_solver = struct
     |> OpamFile.OPAM.with_depopts
          (List.map depopts ~f:Package_dependency.to_opam_filtered_formula
           |> OpamFormula.ands)
+    |> OpamFile.OPAM.with_install install
+    |> OpamFile.OPAM.with_build build
   ;;
 
   let non_local_dependencies local_deps =
@@ -93,9 +110,20 @@ let for_solver
       ; loc = _
       ; depopts
       ; pins
+      ; build
+      ; install
+      ; dune_build = _
       }
   =
-  { For_solver.name; dependencies; conflicts; conflict_class; depopts; pins }
+  { For_solver.name
+  ; dependencies
+  ; conflicts
+  ; conflict_class
+  ; depopts
+  ; pins
+  ; build
+  ; install
+  }
 ;;
 
 let of_package (t : Dune_lang.Package.t) =
@@ -114,11 +142,16 @@ let of_package (t : Dune_lang.Package.t) =
     ; loc
     ; conflict_class = []
     ; pins = Package_name.Map.empty
+    ; build = []
+    ; install = []
+    ; dune_build = true
     }
   | Some { file; contents = opam_file_string } ->
     let opam_file =
       Opam_file.read_from_string_exn ~contents:opam_file_string (Path.source file)
     in
+    let build = opam_file |> OpamFile.OPAM.build in
+    let install = opam_file |> OpamFile.OPAM.install in
     let dependencies =
       opam_file |> OpamFile.OPAM.depends |> Dependency_formula.of_filtered_formula
     in
@@ -150,5 +183,16 @@ let of_package (t : Dune_lang.Package.t) =
           ~loc:pkg.loc
           [ Pp.textf "package %s is already pinned" (Package_name.to_string pkg.name) ]
     in
-    { name; version; dependencies; conflicts; depopts; loc; conflict_class; pins }
+    { name
+    ; version
+    ; dependencies
+    ; conflicts
+    ; depopts
+    ; loc
+    ; conflict_class
+    ; pins
+    ; build
+    ; install
+    ; dune_build = false
+    }
 ;;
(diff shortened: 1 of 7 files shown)


src/dune_pkg/local_package.ml
```diff
@@ -21,6 +21,9 @@ type t =
   ; depopts : Package_dependency.t list
   ; pins : pins
   ; loc : Loc.t
+  ; build : OpamTypes.command list
+  ; install : OpamTypes.command list
+  ; dune_build : bool
```
Leonidas-from-XIV, line 26: This makes for a somewhat inelegant type where if `dune_build` is set then `build` and `install` should be empty and cases where `dune_build` needs to be ignored. I assume if `dune_build` is true then `build` and `install` should be empty? What if they aren't?

I'd suggest something like:

```ocaml
type build_type =
  | Dune
  | Other of { build: OpamTypes.command list; install: OpamTypes.command list }
```

End then
```ocaml
; build_type : build_type
```

src/dune_pkg/resolved_package.ml
```diff
@@ -84,10 +84,10 @@ let scan_files_entries path =
          ])
 ;;
 
-let dune_package loc opam_file opam_package =
+let dune_package ~dune_build loc opam_file opam_package =
```
Leonidas-from-XIV, line 87: This code doesn't seem to make sense to me. `dune_package` should return a `dune_package` but with `~dune_build:false` it doesn't? It needs to be renamed at least but possibly a bit of code needs to be refactored (e.g. by passing around the `build_type` type I am suggesting) and renaming.

test/blackbox-tests/test-cases/pkg/pin-stanza/build-command-dune-project.t
```diff
@@ -51,11 +51,11 @@ Demonstrate the build command we construct for different types of projects:
...
-  $ build_command dune-only
+  $ build_command "(dune)" dune-only
   (dune)
-  $ build_command mixed
-  (dune)
-  $ build_command template
+  $ build_command "(dune)" template
   (dune)
+  $ build_command "(build" mixed
+  (build
```
Leonidas-from-XIV, line 61: This is good but I think it would be also useful to show that pinning and building a project that doesn't use dune works.

Leonidas-from-XIV (commented):
I wasn't even aware that this didn't work. Good catch! The code looks mostly fine, my concerns are mostly that the way it is added it makes it more confusing and in my opinion needs a bit refactoring to keep it understandable and the types neat.

src/dune_pkg/resolved_package.ml
maiste: What do you think about keeping the function, adding a new one with the parameter at false? Maybe it would make more sense?

Leonidas-from-XIV: I think the way you wrote it at the moment looks good to me.

src/dune_pkg/resolved_package.ml
```diff
@@ -84,10 +84,15 @@ let scan_files_entries path =
          ])
 ;;
 
-let dune_package loc opam_file opam_package =
+let local_package ~command_source loc opam_file opam_package =
+  let dune_build =
+    match (command_source : Local_package.command_source) with
```
Leonidas-from-XIV, line 89: I believe our style guide states that in such cases where the compiler doesn't know which constructors we are referring to, the constructors should be prefixed with the module instead of a type annotation. But my memory might be hazy.

Leonidas-from-XIV (commented):
I think it looks good now, just a stylistic issue to look into.

src/dune_pkg/resolved_package.ml
maiste: Quoting for the working on the Dune codebase:
```
Avoid qualifying constructors and record fields. Instead, add type annotations to the type being matched on or being constructed, e.g.,

Bad:
let foo = Command.Args.S []

Good:
let (foo : _ Command.Args.t) = S []

```

Leonidas-from-XIV: Ah. I stand corrected!

Leonidas-from-XIV (approved):

src/dune_pkg/local_package.mli
```diff
@@ -48,6 +56,8 @@ module For_solver : sig
     ; depopts : Package_dependency.t list
     ; conflict_class : Package_name.t list
     ; pins : pins
+    ; build : OpamTypes.command list
```
gridbugs, line 59: Is it necessary to add the commands to this type? I wouldn't have thought the solver needs that information.

src/dune_pkg/local_package.mli
```diff
@@ -10,6 +10,13 @@ type pin =
 
 type pins = pin Package_name.Map.t
 
+type command_source =
```
gridbugs, line 13: Reading this code, I'm confused about the difference between the "origin : [`Dune | `Opam]" field of `pin` and this type. I think it would be helpful to add a comment explaining why both are necessary.

src/dune_pkg/local_package.mli
maiste: Yes it is. Otherwise, the information is lost later because of this:
```ocaml
let resolved_package =
        let local_package = Local_package.of_package pkg in
        let opam_file =
          local_package
          |> Local_package.for_solver
          |> Local_package.For_solver.to_opam_file
          (*                           ^^^^^^ This needs the build and install to not lost the information *)
          |> OpamFile.OPAM.with_url (OpamFile.URL.create (snd package.url))
        in
```

src/dune_pkg/local_package.mli
maiste: I added a comment, tell me if that is clearer or if I need to alter the wording :smile:

gridbugs: Ok so as far as I can tell the `origin` field of `pin` is what type of pin it is, as dune understands packages pinned in the opam style, but also defines a different type of pinning that is specific to dune. And `command_source` is orthogonal to this. An opam pin can be missing an opam file and use dune commands, and a dune pin can have an opam file and use the commands from that file.

A minor nitpick: I think we can name the cases to be a bit more descriptive. How about `Assume_defaults` and `Opam_file`?

gridbugs (approved):

src/dune_pkg/local_package.mli
maiste: I have a slightly different understanding on this for the `origin`: it defines the context on which it is pinned. All the packages I tried (opam file and no opam file), where using the `'Dune` variant because of the `Pin_stanza.decode` function ([source](https://github.com/ocaml/dune/blob/c19917fbea3c61035ec2f83b7780de89409bc08c/src/dune_pkg/pin_stanza.ml#L164)). The `'Opam` variant seems to happen for the deps that are local packages.

I don't see any issue with modifying the name if that makes it clearer :smile:

PR #11515, opened 2025-03-06
pkg: fix opam conflicts "expected formula to be a conjunction"

art-w:
When an opam package declares conflicts with multiple other packages (for example [`bigstringaf.opam`](https://github.com/ocaml/opam-repository/blob/master/packages/bigstringaf/bigstringaf.0.10.0/opam#L28)), `dune pkg lock` currently fails with the error `Expected formula to be a conjunction of atoms but encountered non-atom term 'a | b'`. This appears to be a small oversight as the code expects opam conflicts to be provided as a conjunction, while in practice it's a disjunction. (We could support arbitrary formulas in conflicts, but this quick fix should unstuck the vast majority of packages concerned by this... I'm not aware of packages that conflict with a simultaneous installation of two other packages)

### DIFF
--- a/src/dune_pkg/local_package.ml
+++ b/src/dune_pkg/local_package.ml
@@ -119,12 +119,12 @@ let of_package (t : Dune_lang.Package.t) =
     let opam_file =
       Opam_file.read_from_string_exn ~contents:opam_file_string (Path.source file)
     in
-    let convert_filtered_formula = Package_dependency.list_of_opam_filtered_formula loc in
+    let convert_filtered_formula = Package_dependency.list_of_opam_disjunction loc in
     let dependencies =
       opam_file |> OpamFile.OPAM.depends |> Dependency_formula.of_filtered_formula
     in
-    let conflicts = convert_filtered_formula `Or (OpamFile.OPAM.conflicts opam_file) in
-    let depopts = convert_filtered_formula `Or (OpamFile.OPAM.depopts opam_file) in
+    let conflicts = convert_filtered_formula (OpamFile.OPAM.conflicts opam_file) in
+    let depopts = convert_filtered_formula (OpamFile.OPAM.depopts opam_file) in
     let conflict_class =
       OpamFile.OPAM.conflict_class opam_file
       |> List.map ~f:Package_name.of_opam_package_name
--- a/src/dune_pkg/local_package.ml
+++ b/src/dune_pkg/local_package.ml
@@ -119,12 +119,15 @@ let of_package (t : Dune_lang.Package.t) =
     let opam_file =
       Opam_file.read_from_string_exn ~contents:opam_file_string (Path.source file)
     in
-    let convert_filtered_formula = Package_dependency.list_of_opam_disjunction loc in
     let dependencies =
       opam_file |> OpamFile.OPAM.depends |> Dependency_formula.of_filtered_formula
     in
-    let conflicts = convert_filtered_formula (OpamFile.OPAM.conflicts opam_file) in
-    let depopts = convert_filtered_formula (OpamFile.OPAM.depopts opam_file) in
+    let conflicts =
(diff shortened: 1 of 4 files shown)


test/blackbox-tests/test-cases/pkg/conflicts.t
```diff
@@ -26,3 +26,108 @@ The solver should say no solution rather than just ignoring the conflict.
...
+  - foo2 -> (problem)
+      No usable implementations:
+        foo2.0.0.1: Package does not satisfy constraints of local package x
+  [1]
+
+When conflicts are obtained from an opam file instead of a dune-project,
+the behaviour should be the same:
+
+  $ dune build x.opam
+  $ cat x.opam | sed -n '/conflicts/,/]/p'
```
Leonidas-from-XIV, line 61: Nitpick, but useless use of `cat`.

(That said, I love the `sed` trick, that is really cool and useful)

test/blackbox-tests/test-cases/pkg/conflicts.t
```diff
@@ -26,3 +26,108 @@ The solver should say no solution rather than just ignoring the conflict.
         foo.0.0.1: Package does not satisfy constraints of local package x
   [1]
 
+There could be more than one conflict and they can have versions constraints:
```
Leonidas-from-XIV, line 29:
```suggestion
There could be more than one conflict and they can have version constraints:
```

test/blackbox-tests/test-cases/pkg/conflicts.t
```diff
@@ -26,3 +26,108 @@ The solver should say no solution rather than just ignoring the conflict.
...
+  Couldn't solve the package dependency formula.
+  Selected candidates: bar.0.0.1 bar2.0.0.1 foo.0.2 x.dev
+  - dune -> dune.3.11
+      User requested = 3.18
+  - foo2 -> (problem)
+      No usable implementations:
+        foo2.0.0.1: Package does not satisfy constraints of local package x
+  [1]
+
+And of `foo2` to solve the last remaining conflict:
```
Leonidas-from-XIV, line 105:
```suggestion
Addition of `foo2` to solve the last remaining conflict:
```

src/dune_pkg/local_package.ml
```diff
@@ -119,12 +119,12 @@ let of_package (t : Dune_lang.Package.t) =
     let opam_file =
       Opam_file.read_from_string_exn ~contents:opam_file_string (Path.source file)
     in
-    let convert_filtered_formula = Package_dependency.list_of_opam_filtered_formula loc in
+    let convert_filtered_formula = Package_dependency.list_of_opam_disjunction loc in
```
Leonidas-from-XIV, line 122: Refactoring idea: Maybe this identifier is not useful anymore? It's just contributes a single curried argument. Feel free to ignore if you think it helps readability.

Leonidas-from-XIV (commented):
This is great. The way the formulas are interpreted different depending on whether a field is `depends` vs `depopts`/`conflicts` is really confusing in OPAM files.

I think the issue stems from the fact that in natural language one would say "package foo conflicts with bar and baz" but actually it should be "package foo conflicts with any of bar and baz".

Good catch and nice tests. I added some minor suggestions but I think this is 99% there.

test/blackbox-tests/test-cases/pkg/conflicts.t
art-w: Haha right, thanks!

Leonidas-from-XIV (approved):

src/dune_pkg/local_package.ml
art-w: Thanks, I agree, I changed it :P (it's one line longer due to formatting, but it reads better with the rest of the code without `convert_filtered_formula`)

PR #11516, opened 2025-03-06
Doc sites fix

Leonidas-from-XIV:
I couldn't push the fix to #11502 since it was coming from a `main` branch, so here's a fresh PR.

Fixes #11502.


maiste (approved):
Thanks! I'm good with this change :+1:

Issue #11523, opened 2025-03-10
symlinks in directory targets don't go in shared cache


emillon:
When a directory target contains a symlink, it prevents it from going to the shared cache:

```
  $ export DUNE_CACHE_ROOT=$PWD/.cache
  $ export DUNE_CACHE=enabled

  $ cat > dune-project << EOF
  > (lang dune 3.0)
  > (using directory-targets 0.1)
  > EOF

  $ cat > dune << EOF
  > (rule
  >  (target (dir d))
  >  (action
  >   (progn
  >    (run mkdir -p d)
  >    (chdir d
  >     (progn
  >      (echo building)
  >      (run touch target)
  >      (run ln -s target symlink))))))
  > EOF

  $ dune build d
  building

  $ dune clean

  $ dune build d
  building
```

In this example, the rule is executed the second time (we can see it with the
`echo`). If we remove the `run ln ...` action, it is correctly cached (second
echo is not visible).

I would expect this rule to be able to go to the shared cache; maybe with some
restrictions about the symlink targets.

This seems to prevent some pkg rules to cache correctly, for example
`ocp-indent.1.8.1` has some symlinks and it gets redownloaded after `dune
clean`.

ejgallego:
@emillon do you know if this is this a recent bug or one that has been present since longer?

PR #11524, opened 2025-03-10
test: repro #11523 (symlinks in directory targets)

emillon:
See #11523


maiste (commented):
Is it only for target directory?

maiste (approved):

Leonidas-from-XIV (approved):

PR #11541, opened 2025-03-17
Cache: only store executable permission bit

ElectreAAS:
This PR is built on top of ~~the not-yet-merged #11534~~ main!, and includes the fix proposed in #11533.
We just stop storing the read & write permissions bits, only store the executable one. The strategy is the same as in git, we check if any of the u+x/g+x/o+x bits are set.

### DIFF
--- a/src/dune_digest/digest.ml
+++ b/src/dune_digest/digest.ml
@@ -91,11 +91,13 @@ let file_with_executable_bit ~executable path =
 module Stats_for_digest = struct
   type t =
     { st_kind : Unix.file_kind
-    ; st_perm : Unix.file_perm
+    ; executable : bool
     }
 
   let of_unix_stats (stats : Unix.stats) =
-    { st_kind = stats.st_kind; st_perm = stats.st_perm }
+    (* Check if any of the +x bits are set, ignore read and write *)
+    let executable = 0o111 land stats.st_perm <> 0 in
+    { st_kind = stats.st_kind; executable }
   ;;
 end
 
@@ -113,17 +115,15 @@ let path_with_stats ~allow_dirs path (stats : Stats_for_digest.t) =
   let rec loop path (stats : Stats_for_digest.t) =
     match stats.st_kind with
     | S_LNK ->
-      let executable = Path.Permissions.test Path.Permissions.execute stats.st_perm in
       Dune_filesystem_stubs.Unix_error.Detailed.catch
         (fun path ->
            let contents = Unix.readlink (Path.to_string path) in
-           path_with_executable_bit ~executable ~content_digest:contents)
+           path_with_executable_bit ~executable:stats.executable ~content_digest:contents)
         path
       |> Result.map_error ~f:(fun x -> Path_digest_error.Unix_error x)
     | S_REG ->
-      let executable = Path.Permissions.test Path.Permissions.execute stats.st_perm in
       Dune_filesystem_stubs.Unix_error.Detailed.catch
-        (file_with_executable_bit ~executable)
+        (file_with_executable_bit ~executable:stats.executable)
         path
       |> Result.map_error ~f:(fun x -> Path_digest_error.Unix_error x)
     | S_DIR when allow_dirs ->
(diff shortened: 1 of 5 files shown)


Leonidas-from-XIV:
I've merged #11534 so you can rebase now :-)

ElectreAAS:
Done!

src/dune_digest/digest.ml
```diff
@@ -91,11 +91,13 @@ let file_with_executable_bit ~executable path =
...
     { st_kind : Unix.file_kind
-    ; st_perm : Unix.file_perm
+    ; executable : bool
     }
 
   let of_unix_stats (stats : Unix.stats) =
-    { st_kind = stats.st_kind; st_perm = stats.st_perm }
+    (* Check if any of the +x bits are set, ignore read and write *)
+    let executable = 0o111 land stats.st_perm <> 0 in
+    { st_kind = stats.st_kind; executable }
```
Leonidas-from-XIV, line 100: I kinda wonder whether `Path.Permissions.executable` would make sense to be used here, but maybe it is not worth the hassle.

Alizter:
Is there a discussion of this design change anywhere? I'm curious for the motivation.

src/dune_digest/digest.ml
ElectreAAS: `Path.Permissions.(test executable)` will only test if the flag is set for the current user (u+x, or `0o100`)
I wanted here to explicitly test if that flag is set for anyone.
Granted, files with [u-x g+x o+x] are probably few and far between, but still.

maiste:
I'm curious: do we know the effect it will have on existing cache codebase?

Leonidas-from-XIV:
> Will it retrigger the computation for the already cached elements?

I would assume so since the change is in `Stats_for_digest` and that's probably the digest the cache uses to look up files. Which is.. ok I believe, if this is the way we think it should be done.

ElectreAAS:
> In such case storing and restoring the exact permissions should be 100% backwards compatible, while also solving the issue.

I'm working on a PR to do just that, as I think the naive approach of this PR may be too naive

rgrinberg (approved):

rgrinberg:
Seems fine to me. CHANGES entry please.

ElectreAAS:
I've added a CHANGES entry.
I am however still unconvinced by this approach, it feels like a cop out.

I tried doing what is suggested above, you can see my progress [here](https://github.com/ocaml/dune/compare/main...ElectreAAS:dune:digest-full-perm?expand=1), but I've ran into a problem: storing the permissions of files means we need to compute (`stat`) them somewhere, and then carry them to where the store happens. This has meant changing `Digest.t Targets.Produced.t` into `(Digest.t * Unix.file_perm) Targets.Produced.t` in a million different places, which feels like I'm taking the wrong approach...

mefyl:
I'm not sure restoring the full permissions is desirable, as I explained in https://github.com/ocaml/dune/issues/11533 . Most notably, if dune cache still removes write permissions in hardlink mode the interaction will be weird. Using `stat (2)`, one can also read the number of hardlink references to a file, and that definitely cannot be "restored". This, to my, implies that some file metadata is to be considered out of the scope of dune and that having rules depend on it is unreasonable.

I don't think "group" and "other" permissions belong in a cache or a versioning system since they are local to each machine, I'm not sure it's worth dealing with the weird `u-r` situation since then dune will not even be able to hash the file and things might not work anyway, and I think the dune cache will force `u-w` in hardlink mode anyway. I could be misguided, but I think it's worth double checking we actually want to restore these permissions before investing work in it.

ElectreAAS:
> I'm not sure restoring the full permissions is desirable, as I explained in https://github.com/ocaml/dune/issues/11533.

Oh sorry I missed your message there. Even if it still feels wrong to change the behaviour, as pointed by emillon, I think you've convinced me, so let's go with this approach.
I've rebased on main, and added the changelog entry, and people seem to generally approve - I'll merge once CI is green

PR #11547, opened 2025-03-19
test: cram tests are run twice when both aliases are built

Alizter:
In the first commit we demonstrate a bug with cram tests.

This bug was noticed when working on #11109. The issue is that the cram
rules create two copies of the anonymous cram action, however since the
alias names are different (runtest vs cram test name) this ends up with
having two different digests. This means that when building `@runtest` and
`@mycram` we end up running the cram test `mycram.t` twice.

This test demonstrates this behaviour.

The second commit in this PR fixes this bug.

- [x] changelog


maiste (approved):

Alizter:
Actually this fix isn't great. The issue is that cram tests collect all the aliases together in a set and I have to manually pick them out. This causes indeterminacy observed in the test suite. I'll need to do some more refactoring to achieve the desired determinacy.

Alizter:
I've had to touch a bit more than I hoped, but this seems like it will work better. I'm going to wait and see what the CI will say. I added an extra field to the cram test specification which is the "chosen alias".

When crawling for cram tests and updating empty specifications, we make sure to always set the name of the cram test as the main alias.

Later when we set up the rule for the cram tests we make sure to never iterate over all aliases and instead add rules for the main alias. That way we can add secondary rules for extra aliases depending on the main alias. This means that the anonymous action we are creating won't ever be run twice which was the original bug.

I've also updated the test to describe the expected / observed behaviour.

Alizter:
Looks like I triggered an edge case which I didn't think would happen with this refactoring. I'll investigate and update the fix shortly.

Alizter:
The issue was that the initial cram specification was wrong. I made it so that only the main alias is set.

I'll squash this once the CI is green.

Alizter:
@maiste I think this is ready for another review. I'll squash the fix after you are done so you can see some of the reasoning behind the changes.

src/dune_rules/cram/cram_rules.ml
```diff
@@ -4,7 +4,8 @@ open Memo.O
 module Spec = struct
   type t =
     { loc : Loc.t
-    ; alias : Alias.Name.Set.t
+    ; alias : Alias.Name.t option
```
rgrinberg, line 7: How about a name like `test_name_alias`? Also, why is it an option?

Alizter: Renamed and option removed.

rgrinberg (approved):

src/dune_rules/cram/cram_rules.ml
```diff
@@ -49,53 +51,69 @@ let missing_run_t (error : Cram_test.t) =
...
+    Alias.Name.Set.remove extra_aliases test_name_alias
+    |> Alias.Name.Set.to_list_map ~f:(Alias.make ~dir)
+  in
+  let alias = Alias.make ~dir test_name_alias in
+  (* Here we get all other aliases to depend on the main alias which is the same as the
+     name of the cram test. *)
+  let* () =
+    Memo.parallel_iter extra_aliases ~f:(fun extra_alias ->
+      Rules.Produce.Alias.add_deps ~loc extra_alias (Action_builder.dep (Dep.alias alias)))
+  in
```
Alizter, line 77: Rather than use `Super_context.add_alias_action` I opted to directly register the dependency with `Rules.Produce.Alias.add_deps`.

Alizter:
I'll stop pushing for now, sorry about the bumps. I was a bit rusty on the codebase so I didn't find everything I needed the first time round. This time everything should be ready.

src/dune_rules/cram/cram_rules.ml
maiste: When I had to register an alias, I encounter these two options. What are the difference between them? My bet would be on the way they operate and that one call the other one, but I have always wondered if there was a behavior difference under the hood.

maiste (approved):
Sorry for the late reply. Mostly questions on my side. I found this solution to make the code clearer.

Thanks!

src/dune_rules/cram/cram_rules.ml
```diff
@@ -49,53 +51,69 @@ let missing_run_t (error : Cram_test.t) =
 let test_rule
       ~sctx
       ~dir
-      ({ alias; loc; enabled_if; deps; locks; sandbox; packages = _ } : Spec.t)
+      ({ Spec.test_name_alias
```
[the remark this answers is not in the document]
Alizter: Good point. Must have slipped in.

Issue #11514, opened 2025-03-06
dune-pkg: Local commands (ie `./configure`) should be picked up from the sandbox rather than the source


shym:
## Expected Behavior

When a local command is invoked, typically `./configure`, the version in the sandbox should be run: if patches were applied to that command, the build instructions expect to run the patched version.

## Actual Behavior

The current version is picked up from the source directory.

## Reproduction

In a simple project like this (not opened as a PR as the trick is really in the opam package that will be picked up for `ocamlfind`):

```
$ cat > "dune-workspace" << EOF
> (lang dune 3.17)
> 
> (lock_dir
>  (repositories custom_overlay upstream))
> 
> (repository
>  (name custom_overlay)
>  (url "git+https://github.com/shym/opam-dune-overlays.git#832cc83dc3bc2ade34c02575cbdfe1694440f1a4"))
> EOF

$ cat > "dune-project" << EOF
> (lang dune 3.17)
> (name mini)
> (package
>  (name mini)
>  (depends ocaml ocamlfind))
> EOF

$ cat > "dune" << EOF
> (executable
>  (name mini)
>  (public_name mini)
>  (libraries findlib))
> EOF

$ touch mini.ml
```

when running:

```
$ dune pkg lock && dune build --verbose
```

you get in particular:

```
Running[13]: (cd _build/.sandbox/f53effe87324e49bbf85249c8c3a6c73/_private/default/.pkg/ocamlfind/source && /usr/bin/patch -p1 -i ./pre-fix.patch) > /dev/null
[...]

PR #11551, opened 2025-03-20
test(pkg): Check from which directory a `./cmd` is picked

shym:
Test from which directory local commands, such as `./configure`, are called.
The version from the sandbox should be picked, as they could have been created or modified by previous build commands (such as `patch`).

This is a simplified reproducer for #11514.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/run-local.t
+++ b/test/blackbox-tests/test-cases/pkg/run-local.t
@@ -1,0 +1,35 @@
+Test that local commands, such as `./configure`, are called from the sandbox, as
+they could have been created or modified by previous build commands (such as
+`patch`)
+
+  $ . ./helpers.sh
+
+  $ make_lockdir
+  $ cat >dune.lock/test.pkg <<EOF
+  > (version 0.0.1)
+  > (build
+  >  (progn
+  >   (write-file configure "#!/bin/sh\necho Package configured\n")
+  >   (run chmod +x configure)
+  >   (run ./configure)))
+  > EOF
+
+Now build the package per se. The output is suppressed for now as `./configure`
+fails with an OS-dependent error (when it cannot find the script in the sources)
+so we can only observe the failure until this is fixed.
+
+  $ build_pkg test 2> /dev/null
+  [1]
+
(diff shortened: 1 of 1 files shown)


shym:
Such tests are merged when they are fixed, aren’t they?
Otherwise, I can change the test to silence the failure message to get portability.

Leonidas-from-XIV:
> Such tests are merged when the corresponding issues are fixed, aren’t they?

No, usually we commit the current error message (with a bit of text explaining what is going wrong, so if someone fixes the issue the test changes and it is clear that this is in fact, expected) and the change in the test shows that the issue was fixed.

shym:
> No, usually we commit the current error message (with a bit of text explaining what is going wrong, so if someone fixes the issue the test changes and it is clear that this is in fact, expected) and the change in the test shows that the issue was fixed.

Great! I’ve updated the test to suppress the output to make it portable.

(My assumption that failing tests were not merged was based on [this comment](https://github.com/ocaml/dune/pull/8283#issuecomment-1688133951))

test/blackbox-tests/test-cases/pkg/run-local.t
```diff
@@ -0,0 +1,23 @@
...
+
+  $ make_lockdir
+  $ cat >dune.lock/test.pkg <<EOF
+  > (version 0.0.1)
+  > (build
+  >  (progn
+  >   (write-file configure "#!/bin/sh\necho Package configured\n")
+  >   (run chmod +x configure)
+  >   (run sh configure)
+  >   (run ./configure)))
```
Leonidas-from-XIV, line 15: Why is it running `configure` twice here? Is one failing or are both failing?

shym: It was my work around. You’re right: that was cryptic. I made all that explicit.

Leonidas-from-XIV: Ah, this makes a lot more sense and is well understandable now. Thanks!

Leonidas-from-XIV (approved):
Looks very good now!

Issue #11552, opened 2025-03-21
dune-pkg: Dune raises Code_error when a project specifies version consaints on dune package incompatible with current version


gridbugs:
To reproduce, try solving a project with the following `dune-project`:
```
(lang dune 3.0)

(package
 (name foo)
 (depends
  (dune
   (< 3.0))))
```

The error will be something like:
```
Internal error, please report upstream including the contents of _build/log.
Description:
  ("can't find a valid solution for the dependencies", { name = "foo" })
Raised at Stdune__Code_error.raise in file
  "otherlibs/stdune/src/code_error.ml", line 10, characters 30-62
Called from Dune_pkg__Opam_solver.reject_unreachable_packages.(fun) in file
  "src/dune_pkg/opam_solver.ml", lines 2024-2026, characters 14-56
Called from Stdlib__Map.Make.merge in file "map.ml", line 406, characters
  44-63
Called from Stdlib__Map.Make.merge in file "map.ml", line 403, characters
  64-79
Called from Stdlib__Map.Make.merge in file "map.ml", line 403, characters
  64-79
Called from Stdlib__Map.Make.merge in file "map.ml", line 403, characters
  25-40
Called from Dune_pkg__Opam_solver.reject_unreachable_packages.(fun) in file
  "src/dune_pkg/opam_solver.ml", lines 1993-2035, characters 6-32
Called from Dune_pkg__Opam_solver.solve_lock_dir.(fun) in file
  "src/dune_pkg/opam_solver.ml", lines 2164-2168, characters 12-27
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11
Re-raised at Stdune__Exn.raise_with_backtrace in file
  "otherlibs/stdune/src/exn.ml", line 38, characters 27-56
Called from Fiber__Scheduler.exec in file "vendor/fiber/src/scheduler.ml",
  line 76, characters 8-11

I must not crash.  Uncertainty is the mind-killer. Exceptions are the
little-death that brings total obliteration.  I will fully express my cases.
Execution will pass over me and through me.  And when it has gone past, I
will unwind the stack along its path.  Where the cases are handled there will
be nothing.  Only I will remain.
```

This should be a User_error instead.

hansika2806:
Hello @gridbugs , I'm Hansika. I’d like to work on this issue! It looks like a good way to improve error handling in Dune by converting this Code_error into a User_error. Could you please assign it to me?

[...]

PR #11554, opened 2025-03-21
pkg: handle case where dune violates project deps

gridbugs:
If a project depends on the "dune" package with constraints that are not satisfied by the currently-running instance of Dune, Dune is unable to solve the project. This should be a user error as this behaviour can be triggered by user action.


gridbugs:
Fixes https://github.com/ocaml/dune/issues/11552

Leonidas-from-XIV (commented):
I think this looks good but could you write a short test case which triggers the error message? Just to see that it is acutally triggered.

gridbugs:
I added the test formerly in https://github.com/ocaml/dune/pull/11553

Leonidas-from-XIV (approved):

Issue #4161, opened 2021-01-26
an empty target for easier scripting


gasche:
## Desired Behavior

In a CI script I want to let users specify their testing targets (`"@runtest"` by default, but maybe it's just `"@tests/fast/runtest @tests/somewhat-fast/runtest"` for example), with the convention that if they give an empty string, then tests will not run at all.

Currently I write this in the following way:

```
if [ -n "$TEST_TARGETS" ]; then dune build $TEST_TARGETS; fi
```

But I would like to avoid the conditional by saying Dune: if I don't give you any target, please don't do anything. So something like `dune build --dont-do-anything-by-default $TEST_TARGETS`.

This is almost possible with the option `dune build --default-target="..." $TEST_TARGETS`, except that I have not found a way to specify a default target that *does nothing*. If I use a non-existing alias such as `@empty` there, I get an error and the build fails.

Feature request: provide an explicit target (`@none` or `@empty` for example) that successfully does nothing, so that I can use it as default target for my use-case.

jeremiedimino:
Not necessarily against the idea, but what is the problem with the conditional?

gasche:
In CI scripts (or cram tests) we show the command that is performed to have some observable / self-explanatory output, the readability of the command in the logs is important. For me there is a clear advantage to:

```
dune build --default--target=@empty $DUNE_TEST_TARGETS
```
over
```
if [ -n "$DUNE_TEST_TARGETS" ]; then dune build $DUNE_TEST_TARGETS; fi
```

It's not enough of an argument to ask for a lot of work on the Dune side, or a specific feature to hardcode this use-case, but I think the feature request is general enough, probably easy to implement, and may be useful in other unforeseen scenarios (it makes the target language more regular).

Now that I think of it: in fact we could even ask users to use `@empty` instead of the empty string if they want to skip this step, and it would give even simpler code.

jeremiedimino:
Makes sense. Another name suggestion: `@nothing`. It reads quite well: `dune build @nothing`.

So we can change dune right now so that it doesn't complain if `@nothing` is empty. It's a one line change [here](https://github.com/ocaml/dune/blob/34d654a799263bd830d59352b53cf76d8f79410b/src/dune_engine/alias.ml#L188), you need to add `let nothing = make_standard (Name.of_string "nothing")`. Happy to accept a PR for that.

[...]

PR #11556, opened 2025-03-22
feature: empty alias

Alizter:
This grants #4161.

We add an `@empty` alias allowing the user to build nothing. We also add a check so that user defined rules are not added to the empty alias.

I don't know if we want to or need to be more strict on the checks. For instance, we can bypass the simple rules and add things to the `empty` alias, but it seems pointless to check for this all the time.

- [ ] changelog


test/blackbox-tests/test-cases/alias-empty.t
```diff
@@ -0,0 +1,26 @@
...
+  > EOF
+
+This fails as expected
+  $ dune build @empty
+  File "dune", lines 1-4, characters 0-64:
+  1 | (rule
+  2 |  (alias empty)
+  3 |  (action
+  4 |   (echo "I shouldn't be added!")))
+  Error: User-defined rules cannot be added to the 'empty' alias
```
Alizter, line 25: TODO: make this error message more in line with other similar errors

Alizter:
OK so there is a weird constraint here. I've restricted user-defined rules from adding things to the `empty` alias however a cram test, like we have in the repo, called `empty.t` will then fail when it generates the `empty` alias.

This is why the CI is failing because we are failing on this restriction after bootstrapping.

@rgrinberg @snowleopard the idea that users should be restricted from adding rules in the `empty` alias seems like a good one, but how can we ensure it doesn't restrict people creating cram tests with those names?

rgrinberg:
Maybe you can change the test specific aliases to just be the full filename of the test starting from the next version of dune? You can still keep the old names as well, but just forbid `empty.t` for tests.

rgrinberg:
Merging this and the restriction on the `@empty` alias can be introduced starting from the next dune version.  We can improve things further over time.

maiste:
This is mostly to not forget about it but shouldn't an entry be added in the reference manual for the new alias?

PR #11557, opened 2025-03-22
test: aliases field of user rules generates multiple actions

Alizter:
In a similar vain to #11547 user defined rules were generating multiple actions which meant building both aliases at the same time was running the action twice. We fix this issue by making the other aliases rely on the first alias in the list. (This might be fragile lets see).

### DIFF
--- a/src/dune_rules/simple_rules.ml
+++ b/src/dune_rules/simple_rules.ml
@@ -131,10 +131,16 @@ let user_rule sctx ?extra_bindings ~dir ~expander (rule : Rule_conf.t) =
        Some targets
      | Aliases_only aliases ->
        let+ () =
-         let action = interpret_and_add_locks ~expander rule.locks action.build in
-         Memo.parallel_iter aliases ~f:(fun alias ->
-           let alias = Alias.make ~dir alias in
-           Alias_rules.add sctx ~alias ~loc:rule.loc action)
+         match List.map ~f:(Alias.make ~dir) aliases with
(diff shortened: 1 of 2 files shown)


src/dune_rules/simple_rules.ml
```diff
@@ -132,9 +132,19 @@ let user_rule sctx ?extra_bindings ~dir ~expander (rule : Rule_conf.t) =
      | Aliases_only aliases ->
        let+ () =
          let action = interpret_and_add_locks ~expander rule.locks action.build in
-         Memo.parallel_iter aliases ~f:(fun alias ->
-           let alias = Alias.make ~dir alias in
-           Alias_rules.add sctx ~alias ~loc:rule.loc action)
+         match List.map ~f:(Alias.make ~dir) aliases with
```
hhugo, line 135: Don't we have the same issue above with Aliases_with_targets ?

Alizter: There we are using `Rules.Produce.Alias.add_deps` which correctly registers the deps without introducing any anonymous actions. Here we are using `Super_context.add_alias_action` which is creating an anonymous action. This is a wrapper around `Rules.Produce.Alias.add_action` which for our purposes is probably not what we want.

I've gone ahead and replaced this with `Rules.Produce.Alias.add_deps` here as well and added a test to make sure the with_targets version is treated the same.

Alizter: I can confirm that `with_targets` did not have the issue I am fixing here.

PR #11558, opened 2025-03-23
feature: tests stanzas produce aliases

Alizter:
We add a way to run tests defined in `tests` stanzas by creating an alias with the same name. The basic idea is that:

```
(tests
 (names a b))
```

produces aliases `@a` and `@b` that can be run.

This addresses the other part of #10239 not addressed in #11109. 

- [x] changelog
- [x] docs
- [x] work out a way to do this for `(enabled_if false)`.

### DIFF
--- a/doc/changes/11558.md
+++ b/doc/changes/11558.md
@@ -1,0 +1,2 @@
+- `(tests)` stanzas now generate aliases with the test name. To run `(test (name a))` you
+  can do `dune build @a`. (#11558, grants part of #10239, @Alizter)
--- a/src/dune_rules/test_rules.ml
+++ b/src/dune_rules/test_rules.ml
@@ -115,7 +115,7 @@ let rules (t : Tests.t) ~sctx ~dir ~scope ~expander ~dir_contents =
           in
           let add_alias =
             let expander = Expander.add_bindings expander ~bindings:extra_bindings in
-            let alias = Alias.make ~dir (Alias.Name.of_string s) in
+            let alias = Alias.make ~dir (Alias.Name.of_string ("runtest-" ^ s)) in
             fun ~loc ~action ->
               let action =
                 let chdir = Expander.dir expander in
--- a/src/dune_rules/test_rules.ml
+++ b/src/dune_rules/test_rules.ml
@@ -1,10 +1,11 @@
 open Import
 open Memo.O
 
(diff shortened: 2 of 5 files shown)


test/blackbox-tests/test-cases/test-build-if/feature.t
```diff
@@ -26,7 +26,7 @@ We test the various combinations:
   >   else
   >     echo '  build was done: NO'
   >   fi
-  >   if echo $output|grep -q "alias runtest" ; then
```
Alizter, line 29: The way this test is setup is a little fragile. @emillon any suggestions for improving it?

emillon: You can try printing something to `stderr` from `t.ml` to determine if the test did run.

Alizter: I've used an alert in t.ml to check for compilation and printed something to check for running.

src/dune_rules/test_rules.ml
```diff
@@ -114,9 +114,10 @@ let rules (t : Tests.t) ~sctx ~dir ~scope ~expander ~dir_contents =
                | `js JS | `exe | `bc -> t.deps)
           in
           let add_alias ~loc ~action =
+            let alias = Alias.make ~dir (Alias.Name.of_string s) in
             (* CR rgrinberg: why are we going through the stanza api? *)
```
rgrinberg, line 118: I think this comment out of date now.

doc/changes/11558.md
```diff
@@ -0,0 +1,2 @@
+- `(tests)` stanzas now generate aliases with the test name. To run `(test (name a))` you
+  can do `dune build @a`. (#11558, grants part of #10239, @Alilzter)
```
rgrinberg, line 2: `@Alizter`

src/dune_rules/test_rules.ml
Alizter: Removed.

rgrinberg:
I ended up prefixing the aliases with `runtest:`. Otherwise, the potential for breakage would be too great.

Alizter:
Changelog has the wrong alias, I'll push a fix.

PR #11559, opened 2025-03-23
test: dune exec not updating arg0

### LINKED ISSUE #11527: `dune exec`  passes surprising executable path to the executable
In most cases, programs can rely on `<cwd>/<argv.(0)>` to point to themselves. This is not true when running binaries through `dune exec`. The problem arises when executing binaries in subfolders of the project/workspace.

## Expected Behavior

Any two paths that can be joined to produce the path to the executable, e.g.,
```
pwd: /tmp/bug/subdir
exe: ../_build/install/default/bin/bug
```

## Actual Behavior

`argv.(0)` contains a path that is only valid relative to the project/workspace root, not relative to the current directory.

```
pwd: /tmp/bug/subdir
exe: _build/install/default/bin/bug
```

## Reproduction

1. Retrieve [bug.tar.gz](https://github.com/user-attachments/files/19211792/bug.tar.gz)

  ```bash
cd /tmp
tar xfvz <path/to/bug.tar.gz>
cd bug/
cd subdir && dune exec -- bug
```

## Specifications

- Version of `dune` (output of `dune --version`): `3.17.2`
- Version of `ocaml` (output of `ocamlc --version`): `4.14.2`
- Operating system (distribution and version): `Linux <hostname> 6.13.4-1-siduction-amd64 #1 SMP PREEMPT_DYNAMIC siduction 6.13-5 (2025-02-27) x86_64 GNU/Linux`

### PR DESCRIPTION
**Pull request by Alizter**
Reproduction case and fix for #11527.

- Fixes #11527

### CODE REVIEW

**Comment by Alizter** [comment]
I've pushed a fix and a refactoring commit which drops the arg[0] business from `Proc.restore_cwd_and_execve`. I don't think its productive for callers to have to pass the correct arg[0] and it simplifies many call sites.

File: otherlibs/stdune/src/proc.ml [hunk 1]
```diff
@@ -7,7 +7,7 @@ external sys_exit : int -> 'a = "caml_sys_exit"
 
 let restore_cwd_and_execve prog argv ~env =
   let env = Env.to_unix env |> Array.of_list in
-  let argv = Array.of_list argv in
+  let argv = Array.of_list (prog :: argv) in
```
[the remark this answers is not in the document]
Alizter: We don't run our usual processes using this command but rather the top level dune commands. We didn't have to pass anything, but in several situations we passed the wrong path. This PR essentially passes the correct intended path and makes this function easier to call elsewhere.

So to answer your question no. Unless I've misunderstood, I don't think whatever busybox does should affect us right?

Leonidas-from-XIV:
@Alizter Is there a reason this is marked as draft? It seems fairly sensible to me, fixes a bug, overall I believe uncontroversial and passes all tests.

otherlibs/stdune/src/proc.mli
```diff
@@ -1,3 +1,7 @@
+(** [Proc.restore_cwd_and_execve prog args ~env] runs [prog] with [args] in [env].
+
+  Note that [args] is not like [argv] in that [(Array.to_list args).(0)] is the first
+  argument rather than the conventional program name in [Unix.execve]. *)
```
Leonidas-from-XIV, line 4: Maybe worth rewriting it in a way where it states that prog will be prepended to args to form `argv` as used in `Unix.execve`, because the way it is written it sounds a bit scary and requires knowledge of the C convention. Whereas it is actually a quality of life improvement and makes using it more robust.

test/blackbox-tests/test-cases/github11527.t/dune
```diff
@@ -0,0 +1,3 @@
+(executable
+ (name bug)
+ (public_name bug))
```
Leonidas-from-XIV, line 3: Does it need a `public_name`? To me it seems `(executable (name bug))` or so should be enough?

test/blackbox-tests/test-cases/github11527.t/dune-project
```diff
@@ -0,0 +1,4 @@
+(lang dune 3.17)
+
+(package
+ (name bug))
```
Leonidas-from-XIV, line 4: I think the `package` can be left out and would make the test better as it doesn't imply that it has something to do with packages?

test/blackbox-tests/test-cases/github11527.t/run.t
```diff
@@ -0,0 +1,17 @@
...
+When running dune exec on a binary `bug` we can expect that the working directory and
+argv.(0) can be concatenated to get a valid path.
+
+When running dune exec from the root this is true.
+  $ dune exec -- bug
+  pwd: $TESTCASE_ROOT
+  exe: ./_build/install/default/bin/bug
+
+As expected, the `exe:` argument shows the correct path
+  $ (cd subdir && dune exec --root .. -- bug)
```
Leonidas-from-XIV, line 12: I think it would be a bit nicer to just `mkdir subdir` here, then we don't have to have a `.gitkeep` dummy file to keep the path, but up to you.

Leonidas-from-XIV (approved):

test/blackbox-tests/test-cases/github11527.t/dune
Alizter: It does not, I'll simplify things. This is leftover from the initial issue.

otherlibs/stdune/src/proc.mli
Alizter: I've pushed a modification to the comment. Hopefully things are a bit clearer now.

Leonidas-from-XIV: Yes, I think it is quite nice now. Thanks!

Leonidas-from-XIV (approved):

maiste (approved):

Alizter:
@Leonidas-from-XIV @maiste I've rebased this so its ready for merging.

Issue #11550, opened 2025-03-20
dune-pkg: When extracting an archive fails, dune should print any error messages produced by attempting to run the extraction tool


gridbugs:
Dune uses external tools such as `tar` and `unzip` to extract the contents of package source archives when installing a project's dependencies. These commands can fail for a variety of reasons (e.g. the archive is corrupted, the disk is full, permission problems). When the command to extract an archive fails, dune should capture the error message from the command and present it to the user so they have a chance of understanding and correcting the problem.

aniebietafia:
Hi @gridbugs I am an outreachy applicant and I would like thisissue assigned to me. Thank you.

akshitasure12:
Hi @gridbugs My name is Akshita Sure, an Outreachy internship applicant. Can you please assign this issue to me?

hansika2806:
Hi @gridbugs My name is Hansika, an outreachy internship applicant. I would love to work on this issue, can you please assign this issue to me ?

gridbugs:
To clarify to the outreachy interns asking about this issue, I'm not going to assign anyone this issue as it's been comprehensively solved in https://github.com/ocaml/dune/pull/11560.

gridbugs:
Solved by https://github.com/ocaml/dune/pull/11560

PR #11560, opened 2025-03-23
dune-pkg: Better error messages on extract fail

### LINKED ISSUE #11549: dune-pkg: Print an error when dune is unable to decompress a tarball
Many opam packages have their source stored in a compress tarball, usually compressed with gzip or bzip2. When dune installs a package it uses `tar` (or `bsdtar` in some cases) to extract the archive, decompressing it first with the relevant tool if necessary (`gunzip` or `bunzip2`). If the relevant tool is not installed, the decompression will fail, however the details of the failure will not be shown to the user. It would be more helpful if the error message indicated the nature of the problem and suggested installing the relevant (de)compression tool.

H-ANSEN:
This pull request is in reference to and would close #11549 *(as well as possibly other related issues)* by capturing the output of commands such as `tar`, `gzip`, and `unzip` when dune attempts to install a package and reporting a corresponding error message. The following is an example of the tool `tar` encountering a corrupted file:

```
File "dune.lock/foo.pkg", line 6, characters 3-136:
6 |    file:///DIR_NOT_INCLUDED_FOR_BREVITY/corrupted.tar)))
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Error: failed to extract 'corrupted.tar'
  Reason: tar failed with non-zero exit code '2' and output:
  - /usr/bin/tar: This does not look like a tar archive
  - /usr/bin/tar: Exiting with failure status due to previous errors
```
One note, relating to the original issue was that I was unable to test behavior when a corresponding tool did not exist on the system. If any guidance on how to check this case could be provided I would be more than willing to add additional handling and tests for this case :)

### DIFF
--- a/src/dune_pkg/archive_driver.ml
+++ b/src/dune_pkg/archive_driver.ml
@@ -84,15 +84,15 @@ let extract t ~archive ~target =
     Fiber.return ())
   @@ fun () ->
   Path.mkdir_p target_in_temp;
+
   let stdout_to = Process.Io.make_stdout ~output_on_success:Swallow ~output_limit in
-  let stderr_to = Process.Io.make_stderr ~output_on_success:Swallow ~output_limit in
+  let temp_path = Temp.create File ~prefix:"dune" ~suffix:"extract_out" in
+  let stderr_to = Process.Io.file temp_path Out in
+
   let args = command.make_args ~archive ~target_in_temp in
-  let+ (), exit_code =
-    Process.run ~display:Quiet ~stdout_to ~stderr_to Return command.bin args
-  in
-  match exit_code = 0 with
-  | false -> Error ()
-  | true ->
+  let+ (), exit_code = Process.run ~display:Quiet ~stdout_to ~stderr_to Return command.bin args in
+
+  if exit_code = 0 then ( (* process success *)
     let target_in_temp =
       match Path.readdir_unsorted_with_kinds target_in_temp with
       | Error e ->
@@ -107,4 +107,12 @@ let extract t ~archive ~target =
     Path.mkdir_p (Path.parent_exn target);
     Path.rename target_in_temp target;
     Ok ()
+  ) else (
+    let t = In_channel.open_text (Path.to_absolute_filename temp_path) in
+    User_error.raise
+      [ Pp.textf "failed to extract '%s'" (Path.basename archive)
+      ; Pp.textf "Reason: %s failed with non-zero exit code '%d' and output:" (Path.basename command.bin) exit_code
+      ; Pp.enumerate (In_channel.input_lines t) ~f:(Pp.text)
+      ]
+  )
 ;;
--- a/src/dune_pkg/archive_driver.ml
+++ b/src/dune_pkg/archive_driver.ml
@@ -77,12 +77,13 @@ let extract t ~archive ~target =
   let prefix = Path.basename target in
   let suffix = Path.basename archive in
   let target_in_temp = Temp_dir.dir_for_target ~target ~prefix ~suffix in
+  let temp_stderr_path = Temp.create File ~prefix ~suffix:"stderr" in
   Fiber.finalize ~finally:(fun () ->
     Temp.destroy Dir target_in_temp;
+    Temp.destroy File temp_stderr_path;
     Fiber.return ())
   @@ fun () ->
   Path.mkdir_p target_in_temp;
-  let temp_stderr_path = Temp.create File ~prefix ~suffix:"stderr" in
   let stderr_to = Process.Io.file temp_stderr_path Out in
   let stdout_to = Process.Io.make_stdout ~output_on_success:Swallow ~output_limit in
   let args = command.make_args ~archive ~target_in_temp in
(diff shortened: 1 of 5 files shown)


H-ANSEN:
It appears that `tar` on MacOS exit's with a different exit code and error message than on Linux! Is there any way to isolate test cases/expected output to MacOS or Linux? Or should I just preprocess the output with `sed`?

gridbugs:
> It appears that `tar` on MacOS exit's with a different exit code and error message than on Linux! Is there any way to isolate test cases/expected output to MacOS or Linux? Or should I just preprocess the output with `sed`?

Thanks for the patch! Indeed different tar implementations will have different error messages, and also the location of tar on different systems will differ; there's no guarantee it will be at `/usr/bin/tar`. Conventionally in dune tests we do jest process the output with `sed` and friends to make it as consistent across platforms as possible.

gridbugs:
Looks like this also solves https://github.com/ocaml/dune/issues/11550.

Leonidas-from-XIV:
> Is there any way to isolate test cases/expected output to MacOS or Linux?

Yes, you can enable and disable tests depending on the platform they're running on: https://github.com/ocaml/dune/blob/main/test/blackbox-tests/test-cases/pkg/dune#L22-L26 but for cases where the errors are different I agree with @gridbugs that some postprocessing (or a fake injected `tar` which is consistent) is the better solution as that prevents the test evolving into two divergent tests.

H-ANSEN:
This should now be passing, the issue on `ubuntu-latest` seems to be unrelated.

test/blackbox-tests/test-cases/pkg/pkg-extract-fail.t
```diff
@@ -0,0 +1,62 @@
...
+  > EOF
+
+  $ add_mock_repo_if_needed
+  $ solve foo
+  Solution for dune.lock:
+  - foo.0.0.1
+
+  $ build_pkg foo 2>&1 | sed -ne '/Error:/,$ p' | sed '/^Reason/ q' | sed "s/'[0-9]*'/X/"
+  Error: failed to extract 'corrupted.zip'
+  Reason: unzip failed with non-zero exit code X and output:
```
Leonidas-from-XIV, line 62: I think you'd need to add something like

```
(cram
 (deps %{bin:unzip})
 (applies_to pkg-extract-fail))
```

to `test/blackbox-tests/test-cases/pkg/dune` to prevent failures when `unzip` is missing (I guess `tar` and `gzip` can be assumed to be standard POSIX tools).

test/blackbox-tests/test-cases/pkg/unavailable-package-source.t
```diff
@@ -36,9 +36,6 @@ Git
...
 HTTP
 
-  $ runtest "(fetch (url \"https://0.0.0.0:35000\"))" 2>&1 | sed -ne '/Error:/,$ p' | sed -e 's/".*"/"<PATH>"/'
-  Error:
-  failed to unpack archive downloaded from https://0.0.0.0:35000
-  reason:
-  unable to extract "<PATH>"
-  
+  $ runtest "(fetch (url \"https://0.0.0.0:35000\"))" 2>&1 | sed -ne '/Error:/,$ p' | sed '/^Reason/ q' | sed "s/'[0-9]*'/X/"
+  Error: failed to extract 'download'
```
Leonidas-from-XIV, line 40: The error message on that isn't great because `download` in this context seem out of the blue but it isn't really worse than before. Maybe you'd want to improve this, but the PR is fine either way.

Leonidas-from-XIV (commented):
Looks pretty good. The failing test is being flaky sometimes and fixes itself upon rerunning usually. I've added some comments of things to look into but overall I think this is a good approach.

src/dune_pkg/archive_driver.ml
```diff
@@ -106,5 +106,16 @@ let extract t ~archive ~target =
...
-    Ok ()
+    Ok ())
+  else
+    Io.with_file_in temp_stderr_path ~f:(fun err_channel ->
+      let stderr_lines = Io.input_lines err_channel in
+      User_error.raise
+        [ Pp.textf "failed to extract '%s'" (Path.basename archive)
+        ; Pp.textf
+            "Reason: %s failed with non-zero exit code '%d' and output:"
+            (Path.basename command.bin)
```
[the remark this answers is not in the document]
H-ANSEN: Yep!

test/blackbox-tests/test-cases/pkg/pkg-extract-fail.t
H-ANSEN: Added! I noticed the same file already had a `tar` dependency rule so I went ahead and also added the tests to that.

test/blackbox-tests/test-cases/pkg/unavailable-package-source.t
H-ANSEN: I believe that there is a bit more context provided before the output is preprocessed for the test case, either way I agree it's rather vague.

Leonidas-from-XIV (approved):

gridbugs (approved):

PR #11566, opened 2025-03-25
Alternative to #11437: warn, do not fail, if sandbox cannot be deleted

nojb:
An alternative to #11437: simply warn if the sandbox cannot be deleted, instead of failing.

Yet another possibility could be to warn and retry (as is done in #11437) the `Path.rm_rf` operation.

For discussion.

### DIFF
--- a/src/dune_engine/sandbox.ml
+++ b/src/dune_engine/sandbox.ml
@@ -316,20 +316,28 @@ let move_targets_to_build_dir t ~should_be_skipped ~(targets : Targets.Validated
         then Path.rename (Path.build src_dir) (Path.build target)))
 ;;
 
-let failed_to_delete_sandbox dir reason =
-  User_warning.emit
+let failed_to_delete_sandbox ~warn dir reason =
+  let msg =
     [ Pp.textf "failed to delete sandbox in %s" (Path.Build.to_string_maybe_quoted dir)
     ; Pp.text "Reason:"
     ; reason
     ]
+  in
+  if warn then User_warning.emit msg else User_error.raise msg
 ;;
 
 let destroy t =
   maybe_async (fun () ->
-    try Path.rm_rf (Path.build t.dir) with
-    | Sys_error e -> failed_to_delete_sandbox t.dir (Pp.verbatim e)
-    | Unix.Unix_error (error, syscall, arg) ->
-      failed_to_delete_sandbox
-        t.dir
-        (Unix_error.Detailed.pp (Unix_error.Detailed.create error ~syscall ~arg)))
+    let rec loop cnt =
+      try Path.rm_rf (Path.build t.dir) with
+      | Sys_error e -> failed_to_delete_sandbox ~warn:false t.dir (Pp.verbatim e)
+      | Unix.Unix_error (error, syscall, arg) ->
+        let warn = error = Unix.EACCES && cnt > 0 in
+        failed_to_delete_sandbox
+          ~warn
+          t.dir
+          (Unix_error.Detailed.pp (Unix_error.Detailed.create error ~syscall ~arg));
+        loop (cnt - 1)
+    in
+    loop (if Sys.win32 then 30 else 0))
 ;;
--- a/src/dune_engine/sandbox.ml
+++ b/src/dune_engine/sandbox.ml
@@ -317,7 +317,7 @@ let move_targets_to_build_dir t ~should_be_skipped ~(targets : Targets.Validated
 ;;
 
 let failed_to_delete_sandbox dir reason =
-  User_error.raise
+  User_warning.emit
     [ Pp.textf "failed to delete sandbox in %s" (Path.Build.to_string_maybe_quoted dir)
     ; Pp.text "Reason:"
     ; reason


Alizter:
I think this fix makes sense for the time being. It's not really the users fault if a sandbox cannot be deleted, and I don't think it is important enough for a build to fail.

If you look above, we already try to delete it when we create a sandbox. If for whatever reason that fails on Windows on the off chance that the sandbox wasn't cleared in a previous build and the virus checker has locked the file, then we can error out. The user should in that case just clear the `_build` directory manually and rebuild.

I don't know if it is worth adding some more sophisticated logic in the future for Windows to avoid this situation entirely, but from reading the previous discussion it appears to be rare and non-fatal.

Alizter (approved):

Alizter:
Before you update the test and merge, let's give a headsup to the JS people since this patch will affect them more directly.

cc @rgrinberg @snowleopard

nojb:
Thanks for the quick review @Alizter. @MSoegtropIMC: would this fix be OK for you?

MSoegtropIMC:
Reading the comment of @Alizter I would think that avoiding issues with sandbox clean up would be better. "Manually clean up and restart the build" is not helpful for CI, while my solution fixed all CI issues I had. How about having both changes - the retry on file deletes from my PR and this?

MSoegtropIMC:
So what happens if a previous sandbox could not be cleaned up and then another sandbox needs to be created? Will this lead to a build failure? Or do sandbox folders have unique names?

Alizter:
> So what happens if a previous sandbox could not be cleaned up and then another sandbox needs to be created? Will this lead to a build failure? Or do sandbox folders have unique names?

The sandbox directories contain a digest of their contents, so unless a rule is being rerun, which won't happen in a single build, there should be no conflicts. The behaviour I am mentioning would only happen on a rebuild, and seems too unlikely to cause major issues. We could perhaps have a better error message in that case however. I can have a look later today at improving that.

MSoegtropIMC:
I still don't see advantages of this solution over my suggestion. As discussed it is more of a hack because it ignores errors and leaves things in an unclean state while mine actually fixes the issue. Regearding the critics on my solution - that it might slow down builds - I was about to say that this is purely theoretical, but thinking about this it is not even this. So I would prefer my solution - which is unlike this solution also heavily tested in CI since > 1 month.

nojb:
> So I would prefer my solution - which is unlike this solution also heavily tested in CI since > 1 month.

I think this is a good point, and personally I would be ready to merge #11437 in its current state. The only small issue I see with the way the implementation of #11437 is that if the user is bitten by this problem, the logic kicks in completely silently without reporting what is going on (which may be useful information).

So as a last proposal, I pushed https://github.com/ocaml/dune/pull/11566/commits/9a539d8e1af2cc8fc28bfe6eac2fc076137519d9 which implements the same retry logic as in #11437 but in `Sandbox.destroy` instead of `Fpath.unlink` which allows us to tell the user what is going on. Could this variant satisfy everyone?

Alizter:
To be clear, I am just expressing my opinion on `fpath` vs this patch. At the end of the day, it won't affect me. I think the last commit @nojb pushed is good since it will also give the user some indication of what is going on unlike before.

@MSoegtropIMC will you be able to check that this works for you, or is this something that is difficult to reproduce over a short timescale?

MSoegtropIMC:
The question is if this issue can also happen in non sandbox mode. As far as I can see there is another call to this function in Dune/dune/src/dune_engine/build_system.ml in "execute_action_for_rule", a failing unlink could have strange effects there as well.

Since the main issue appears to be logging and the function is called in only few places, how about passing a logging function to the unlink function?

Another point: there are quite a few occurrences of Unix.unlink in the dune source tree. Shouldn't these be replaced by the safe unlink function?

src/dune_engine/sandbox.ml
```diff
@@ -316,20 +316,28 @@ let move_targets_to_build_dir t ~should_be_skipped ~(targets : Targets.Validated
...
+    let rec loop cnt =
+      try Path.rm_rf (Path.build t.dir) with
+      | Sys_error e -> failed_to_delete_sandbox ~warn:false t.dir (Pp.verbatim e)
+      | Unix.Unix_error (error, syscall, arg) ->
+        let warn = error = Unix.EACCES && cnt > 0 in
+        failed_to_delete_sandbox
+          ~warn
+          t.dir
+          (Unix_error.Detailed.pp (Unix_error.Detailed.create error ~syscall ~arg));
+        loop (cnt - 1)
```
rusv-simcorp, line 340: We should sleep here a bit before looping, to avoid trying to delete the sandbox in a tight loop

src/dune_engine/sandbox.ml
```diff
@@ -316,20 +316,28 @@ let move_targets_to_build_dir t ~should_be_skipped ~(targets : Targets.Validated
...
-        (Unix_error.Detailed.pp (Unix_error.Detailed.create error ~syscall ~arg)))
+    let rec loop cnt =
+      try Path.rm_rf (Path.build t.dir) with
+      | Sys_error e -> failed_to_delete_sandbox ~warn:false t.dir (Pp.verbatim e)
+      | Unix.Unix_error (error, syscall, arg) ->
+        let warn = error = Unix.EACCES && cnt > 0 in
+        failed_to_delete_sandbox
+          ~warn
+          t.dir
+          (Unix_error.Detailed.pp (Unix_error.Detailed.create error ~syscall ~arg));
```
rusv-simcorp, line 339: The PR description says "simply warn if the sandbox cannot be deleted, instead of failing" but this part will fail after retrying 30 times.

My understanding is that we don't want the build to fail in case the sandbox files can't be deleted.

rusv-simcorp:
@nojb FYI, I have implemented my above review comments in https://github.com/rusv-simcorp/dune/commit/9f114b4619d863ce45906aabcdd1bd8aa14d1cb0

nojb:
Let's go with #11437 and keep these alternative approaches in mind for the future if needed.

Issue #11419, opened 2025-01-30
Controlling flags for ocamldep (e.g. for specifying keywords)


pmundkur:
There will likely be a need to allow specification of flags to `ocamldep`, depending on the resolution of ocaml/ocaml#13764.  Currently, as per the docs, `flags` can only be specified for `ocamlc` and `ocamlopt`.  

Alternatively, there could be a special `keywords` construct that is applied to all tools that use the ocaml lexer/parser version 5.3.0 or later.  However, this might collide with keywords being specified in `flags`.

maiste:
Thanks for reporting.
Let's see the outcome of the issue on ocaml/ocaml first 👍

wmuth:
Any update on adding this feature now that the linked issue in ocamldep has been closed?

maiste:
The PR been merged on OCaml, we don't have any blockers to add this feature. Note that our first suggestion would be to change the breaking variables that use effect in the codebase because it would generate problem later for such codebase. Our bandwidth being a bit short at the moment, we would be happy to have a PR adding the stanza 😄

gridbugs:
Done! Feel free to reach out (here or on [discord](https://discord.gg/cCYQbqN)) if you need more info.

gridbugs:
@anvithagowda098 how are you going with this issue? Please reach out if you need help.

anvithagowda098:
@gridbugs Sorry I’ll not be able to work on this immediately due to other engagements. You may assign this to someone else.

3pleX-dev:
Hi @gridbugs, I'd love to take a shot at this! I've set up everything with Dune and am somewhat familiar with OCaml. I'm currently contributing to Claudius too.

gridbugs:
Thanks @wmuth! It's probably easiest to send a PR so other dune maintainers can give their critique as well.

gridbugs:
@3pleX-dev in light of @wmuth's work, I won't assign any interns to this project.

PR #11567, opened 2025-03-25
feat:Buildable stanza ocamldep_flags

wmuth:
This PR in a work in progress but addresses issue #11419.

As I have never contributed to Dune before I'd like some feedback on my implementation and any changes which could be made to make my contribution fit better into the Dune code base :).

I'll add to the documentation and push it to my branch as well before this is ready to be merged.

Right now `ocamldep_flags` is only supported in buildable stanzas as that made the most sense to me and I implemented the flag parsing by reusing the functionality used by `ocamlc_flags` and `ocamlopt_flags`.

The additions didn't feel enough to warrant thir own file and implementing it alongside the flags for ocamlc, ocamlopt and melange (in the `Ocaml_flags.t`) felt like adding a lot of bloat to those functions to support this which currently doesn't need all of that functionality. Hence I added `Ocaml_flags.Ocamldep_flags`.

### DIFF
--- a/src/dune_rules/buildable_rules.ml
+++ b/src/dune_rules/buildable_rules.ml
@@ -21,6 +21,14 @@ let ocaml_flags t ~dir (spec : Ocaml_flags.Spec.t) =
     Ocaml_flags.with_vendored_flags ~ocaml_version flags
 ;;
 
+let ocamldep_flags t ~dir spec =
+  let* expander = Super_context.expander t ~dir in
+  let expanded_flags =
+    Expander.expand_and_eval_set expander spec ~standard:(Action_builder.return [])
+  in
+  Memo.return expanded_flags
+;;
+
 let gen_select_rules sctx ~dir compile_info =
   Lib.Compile.resolved_selects compile_info
   |> Resolve.Memo.read_memo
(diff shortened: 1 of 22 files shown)


nojb (commented):
I am not very convinced by the approach here: for the most part, Dune needs to control the flags passed to `ocamldep` as it depends on it to compute dependencies accurately. I don't see the ability to to pass arbitrary flags to `ocamldep` as being very useful.

Having a `(keywords)` field (similar to `(flags)`) seems more useful, and it would be shared among all the tools.

wmuth:
@nojb Should that `(keywords)` field use the ordered set language? Since it would be dedicated for that flag I assume instead it should just be an optional string field?

Would it make more sense to parse it as a separate part of the buildable stanza or to parse it in `Ocaml_flags.Spec.decode` alongside other flags?

I can make that change and make a new PR for that instead.

nojb:
> Should that `(keywords)` field use the ordered set language?

The semantics of `-keywords` is a pair `(version, list)` where `version` is an optional version (in the format `%d.%d`) and `list` is a list of extra keywords. So the syntax of `(keywords)` should reflect that, eg something along the lines of  `(keywords (version 5.3) (extra keyword1 keyword2))`.

> or to parse it in `Ocaml_flags.Spec.decode` alongside other flags?

This sounds like a reasonable approach.

PR #11573, opened 2025-03-26
pkg: Display error message when no dependencies exist in `dune-project`

Leonidas-from-XIV:
As a follow-up/replacement of #11555 this PR displays an warning if the package to be locked is declared in `dune-project` but has no dependencies.

I could imagine an additional step of reading the OPAM file and actually erroring out if it has dependencies declared there. That sounds like something we don't want to support. WDYT?

### DIFF
--- a/src/dune_pkg/local_package.ml
+++ b/src/dune_pkg/local_package.ml
@@ -145,7 +145,27 @@ let of_package (t : Dune_lang.Package.t) =
   let name = Package.name t in
   match Package.original_opam_file t with
   | None ->
-    let dependencies = t |> Package.depends |> Dependency_formula.of_dependencies in
+    let depends =
+      match Package.depends t, Package.has_opam_file t with
+      | [], Exists true ->
+        let name = t |> Package.name |> Package_name.to_string in
+        let hints =
+          [ Pp.textf "Delete the package %S from `dune-project`" name
+          ; Pp.textf "Add a `depends` field in the `package` stanza of package %S" name
+          ]
+        in
+        User_warning.emit
+          ~hints
+          [ Pp.textf
+              "Package %s loaded from `dune-project` but does not declare any \
+               dependencies"
+              (t |> Package.name |> Package_name.to_string)
+          ];
+        []
+      | [], Exists false | [], Generated -> []
+      | otherwise, _ -> otherwise
+    in
+    let dependencies = Dependency_formula.of_dependencies depends in
     { name
     ; version
     ; dependencies


src/dune_pkg/local_package.ml
```diff
@@ -145,7 +145,27 @@ let of_package (t : Dune_lang.Package.t) =
...
+        in
+        User_warning.emit
+          ~hints
+          [ Pp.textf
+              "Package %s loaded from `dune-project` but does not declare any \
+               dependencies"
+              (t |> Package.name |> Package_name.to_string)
+          ];
+        []
+      | [], Exists false | [], Generated -> []
```
gridbugs, line 165: I think this case is also captured by the following case.

src/dune_pkg/local_package.ml
```diff
@@ -145,7 +145,27 @@ let of_package (t : Dune_lang.Package.t) =
...
+        let name = t |> Package.name |> Package_name.to_string in
+        let hints =
+          [ Pp.textf "Delete the package %S from `dune-project`" name
+          ; Pp.textf "Add a `depends` field in the `package` stanza of package %S" name
+          ]
+        in
+        User_warning.emit
+          ~hints
+          [ Pp.textf
+              "Package %s loaded from `dune-project` but does not declare any \
```
gridbugs, line 160: I'd prefer to use %S as you did in the hint above in case it's not clear that the word following "Package" in the message is a package name.

gridbugs:
Can you add a test of the new behaviour.

src/dune_pkg/local_package.ml
```diff
@@ -145,7 +145,26 @@ let of_package (t : Dune_lang.Package.t) =
...
   match Package.original_opam_file t with
   | None ->
-    let dependencies = t |> Package.depends |> Dependency_formula.of_dependencies in
+    let depends =
+      match Package.depends t, Package.has_opam_file t with
+      | [], Exists true ->
+        let name = t |> Package.name |> Package_name.to_string in
+        let hints =
+          [ Pp.textf "Delete the package %S from `dune-project`" name
+          ; Pp.textf "Add a `depends` field in the `package` stanza of package %S" name
```
maiste, line 154: Shouldn't you add a `or` somewhere? You need to do one or another, no?

src/dune_pkg/local_package.ml
```diff
@@ -145,7 +145,26 @@ let of_package (t : Dune_lang.Package.t) =
   let name = Package.name t in
   match Package.original_opam_file t with
   | None ->
-    let dependencies = t |> Package.depends |> Dependency_formula.of_dependencies in
+    let depends =
+      match Package.depends t, Package.has_opam_file t with
+      | [], Exists true ->
```
maiste, line 150: :question: This means that we don't do any distinction between `(package name)` and `(package name (depends))`?

src/dune_pkg/local_package.ml
Leonidas-from-XIV: I think these hints are separate and in theory unordered (in practice they are probably still printed in order). Maybe the better solution would just be displaying one of the hints instead of two conflicting hints.

src/dune_pkg/local_package.ml
Leonidas-from-XIV: We cannot, `Package.t` does not distinguish between empty and unspecified for `depends`.

In #11555 we discussed that there is no purpose in allowing to load data from multiple sources so if `(package (name foo))` is declared, that means that this is the canonical source for the `Package.t` and this package has no dependencies. So there is no semantic difference between specifying an empty `(depends)` vs no depends at all and we might as well continue representing it as `[]`.

src/dune_pkg/local_package.ml
maiste: Maybe, I don't have a strong opinion on this. If I chose one, I would go for the `depends` one, as this is a practice we try to enforce with Dune.

src/dune_pkg/local_package.ml
maiste: Make sense, thanks for the clarification!

src/dune_pkg/local_package.ml
Leonidas-from-XIV: Ok, I swapped the order and added an `or`.

maiste (approved):

gridbugs (approved):

Leonidas-from-XIV:
There is something really wrong that it displays the warning 4 times (which at least points to the fact that we're doing 4x as much work as we should), I'm investigating this.

Alizter:
@Leonidas-from-XIV Are you planning to rebase this soon?

Leonidas-from-XIV:
Fixing the error message being displayed 4x is a bigger endeavor, as the issue is that the build files get loaded multiple times as part of the build. I tried to refactor the loading but it was a massive change. So unless I stick a `ref` in here (`let displayed = ref false`) it is too much effort to solve it properly.

rgrinberg:

PR #11578, opened 2025-03-28
feat:Buildable stanza keywords

wmuth:
# TLDR
This pull request aims to solve #11419 and is updated based on feedback I got in #11567 .

## TODO
What is left to be done on this PR beyond any review changes is that I will need to add some tests and add memoization to the keywords. Both of which I'm unsure how to do properly so I'd like some assistance as to where I should add the tests and guidance on how to implement memoziation for the keywords.

## Details
The PR adds a `keywords` field to buildable stanzas which is a `(string option, string list)` much like the `-keywords` flag for `ocamlc`, `ocamlopt` and `ocamldep`. The way it is used in the dune file is e.g. `(library ... (keywords (version 5.2) (extra effect))` which will pass the flag `-keywords 5.2+effect`.

The storing of the keywords is done in the `Ocaml_flags.t` as a member `keywords` which is a `(string option, string list)`. I added a submodule `Keywords` to `ocaml_flags.ml` which contains some helper functions and should make code more readable in other files as the path is `Ocaml_flags.Keywords.func`.

I didn't add a Keywords.t to ideally make the code less abstracted which is something I think I read in hacking.rst.

The added field has been set to only be supported in Dune stanza syntax version 3.18.

I also added some documentation for the field alongside the currently existing documentation for Ocaml flags.

As I have never contributed to Dune before I'm unsure if I have done things the most idiomatic way but am happy to revise my solution to make it better.

### DIFF
--- a/src/dune_rules/ocaml_flags.ml
+++ b/src/dune_rules/ocaml_flags.ml
@@ -64,11 +64,35 @@ let default_flags ~dune_version ~profile =
 
 type 'a t' =
   { common : 'a
+  ; keywords : string option * string list
   ; specific : 'a Lib_mode.Map.t
   }
 
-let equal f { common; specific } t =
-  f common t.common && Lib_mode.Map.equal f specific t.specific
+module Keywords = struct
+  let empty = None, []
+  let standard = empty
+
+  let equal (v1, e1) (v2, e2) =
+    Option.equal String.equal v1 v2 && List.equal String.equal e1 e2
+  ;;
+
+  let to_string_list t =
+    match t.keywords with
+    | None, [] -> []
+    | Some version, [] -> "-keywords" :: [ version ]
+    | None, extra_l ->
+      let keywords_arg = "" :: extra_l |> String.concat ~sep:"+" in
+      "-keywords" :: [ keywords_arg ]
+    | Some version, extra_l ->
+      let keywords_arg = version :: extra_l |> String.concat ~sep:"+" in
+      "-keywords" :: [ keywords_arg ]
+  ;;
+end
+
+let equal f { common; keywords; specific } t =
+  f common t.common
+  && Lib_mode.Map.equal f specific t.specific
+  && Keywords.equal keywords t.keywords
 ;;
 
 module Spec = struct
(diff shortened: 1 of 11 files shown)


nojb:
Thanks for the ping. I will take a look, but it may take me a few days to get to it.

gridbugs:
Thanks for the patch, I'll take a look next week.

wmuth:
@rgrinberg 
> What's the behavior of this feature on older versions of OCaml?

Trying this in a OCaml 5.0.0 project I got the following error message

```
File "dune", lines 1-5, characters 0-104:
1 | (executable
2 |  (name my_binary)
3 |  (public_name my_binary)
4 |  (libraries my_library)
5 |  (keywords (version 5.2) (extra effect)))
/home/user/.opam/miking-ocaml/bin/ocamlc.opt: unknown option '-keywords'.
Usage: ocamlc <options> <files>
Try 'ocamlc --help' for more information.
```

So the older version of OCaml says it doesn't know about `-keywords`. You could of course check the OCaml version inside `dune` with the context to determine whether to add the `-keywords` flag or not. In that case maybe `dune` should warn the user that it is unable to apply the flag the user has requested to be applied?

rgrinberg:
Yeah, I think we need to do better than this. A proper error message or just ignore this flag for older versions of OCaml.

src/dune_rules/ocaml_flags.ml
```diff
@@ -64,11 +64,35 @@ let default_flags ~dune_version ~profile =
 
 type 'a t' =
   { common : 'a
+  ; keywords : string option * string list
```
gridbugs, line 67: I see you mentioned not over-abstracting as per our hacking doc, but in this case I think a `Keywords.t` or similar is necessary so that the version and extra keywords can go in named fields. Otherwise I think it will be hard for future readers of this code to understand the meaning of the tuple elements.

src/dune_rules/ocaml_flags.mli
```diff
@@ -4,6 +4,13 @@ open! Import
 
 type t
 
+module Keywords : sig
+  val empty : string option * string list
```
gridbugs, line 8: Looking at this interface, this would definitely benefit from a `Keywords.t` type to avoid the repetition of `string option * string list`.

src/dune_rules/ocaml_flags.ml
```diff
@@ -78,39 +102,54 @@ module Spec = struct
...
     }
   ;;
 
-  let make ~common ~specific : t = { common; specific }
+  let make ~common ~keywords ~specific : t = { common; keywords; specific }
+
+  let decode_keywords =
+    let open Dune_lang.Decoder in
+    fields
+      (let+ version = field_o "version" string
```
gridbugs, line 115: `ocamlc` enforces a rule that the version number be `<major>.<minor>`. I think it would be better UX for dune to also enforce this rule while parsing this field because otherwise malformed version strings will cause `ocamlc` to print its error, which won't have the context that dune has while parsing.

PR #11581, opened 2025-03-30
feat(dune_lang): multiline strings

Alizter:
This is a work-in-progress but demonstrates a way we might be able to support the preservation of multiline strings in Dune. This will allow us to preserve multiline strings when formatting dune files for instance.


maiste:
Is this PR trying to achieve the same as #10780? Just to make sure if one of them is merged, we don't forget to close the other :+1:

rgrinberg:
Please re-open once this is ready

PR #11582, opened 2025-03-31
pkg: remove broken symlinks from package sources

gridbugs:
Dune cannot handle symlinks with missing targets. Package source archives may contain such broken symlinks. This can be due to packaging errors on the part of the package author which are benign to the correct operation of the package when installed with opam, but render packages uninstallable with dune.

It's assumed that all opam packages can be installed with opam, so if a package contains a broken symlink then the symlink must not be necessary for the correct operation of the package. Thus the fix is to remove broken symlinks from extracted source archives, and to avoid copying broken symlinks from local package source directories.

A less conservative approach than this might attempt to allow dune to handle broken symlinks in general, but this would be a much larger change, and has been attempted and abandoned several times before.

### DIFF
--- a/otherlibs/stdune/src/fpath.ml
+++ b/otherlibs/stdune/src/fpath.ml
@@ -195,7 +195,7 @@ let rm_rf fn =
 ;;
 
 let traverse =
-  let rec loop on_file on_dir root stack acc =
+  let rec loop on_file on_dir on_broken_symlink root stack acc =
     match stack with
     | [] -> acc
     | dir :: dirs ->
@@ -211,17 +211,30 @@ let traverse =
              | S_LNK ->
                let path = Filename.concat dir_path fname in
                (match (Unix.stat path).st_kind with
-                | exception Unix.Unix_error (Unix.ENOENT, _, _) -> stack, acc
+                | exception Unix.Unix_error (Unix.ENOENT, _, _) ->
+                  stack, on_broken_symlink ~dir fname acc
                 | S_DIR -> Filename.concat dir fname :: stack, on_dir ~dir fname acc
                 | S_REG -> stack, on_file ~dir fname acc
                 | _ -> stack, acc)
              | _ -> stack, acc)
          in
-         loop on_file on_dir root stack acc)
+         loop on_file on_dir on_broken_symlink root stack acc)
   in
-  fun ~dir ~init ~on_file ~on_dir -> loop on_file on_dir dir [ "" ] init
+  fun ~dir ~init ~on_file ~on_dir ~on_broken_symlink ->
+    loop on_file on_dir on_broken_symlink dir [ "" ] init
 ;;
 
 let traverse_files ~dir ~init ~f =
-  traverse ~dir ~init ~on_dir:(fun ~dir:_ _fname acc -> acc) ~on_file:f
+  let skip = fun ~dir:_ _fname acc -> acc in
+  traverse ~dir ~init ~on_dir:skip ~on_broken_symlink:skip ~on_file:f
+;;
+
+let is_broken_symlink path =
+  let stats = Unix.lstat path in
+  match (stats.st_kind : Unix.file_kind) with
+  | S_LNK ->
+    (match Unix.stat path with
+     | exception Unix.Unix_error (Unix.ENOENT, _, _) -> true
+     | _ -> false)
+  | _ -> false
 ;;
--- a/otherlibs/stdune/src/fpath.ml
+++ b/otherlibs/stdune/src/fpath.ml
@@ -194,8 +194,8 @@ let rm_rf fn =
   | _ -> unlink_exn fn
 ;;
 
-let traverse =
-  let rec loop on_file on_dir on_broken_symlink root stack acc =
+let traverse ~dir ~init ~on_file ~on_dir ~on_broken_symlink =
+  let rec loop root stack acc =
     match stack with
     | [] -> acc
     | dir :: dirs ->
@@ -218,10 +218,9 @@ let traverse =
                 | _ -> stack, acc)
              | _ -> stack, acc)
          in
-         loop on_file on_dir on_broken_symlink root stack acc)
+         loop root stack acc)
   in
-  fun ~dir ~init ~on_file ~on_dir ~on_broken_symlink ->
-    loop on_file on_dir on_broken_symlink dir [ "" ] init
+  loop dir [ "" ] init
 ;;
 
 let traverse_files ~dir ~init ~f =
--- a/otherlibs/stdune/src/path.ml
+++ b/otherlibs/stdune/src/path.ml
@@ -496,6 +496,7 @@ module External : sig
   val as_local : t -> string
   val append_local : t -> Local.t -> t
   val of_filename_relative_to_initial_cwd : string -> t
+  val is_broken_symlink : t -> bool
 end = struct
   module Table = String.Table
 
@@ -570,6 +571,8 @@ end = struct
   let of_filename_relative_to_initial_cwd fn =
     if Filename.is_relative fn then relative initial_cwd fn else of_string fn
   ;;
+
+  let is_broken_symlink = Fpath.is_broken_symlink
 
   include (
     Comparator.Operators (struct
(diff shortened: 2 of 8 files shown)


otherlibs/stdune/src/fpath.ml
```diff
@@ -187,7 +187,7 @@ let rm_rf fn =
 ;;
 
 let traverse =
-  let rec loop on_file on_dir root stack acc =
+  let rec loop on_file on_dir on_broken_symlink root stack acc =
```
Leonidas-from-XIV, line 190:
```suggestion
let traverse ~on_broken_symlink =
  let rec loop on_file on_dir root stack acc =
```

Given `on_broken_symlink` isn't really part of the loop (it is never changed and the recursive calls always reuses the same one), I'd put it outside of the recursive function arguments. That way one does not have to think whether these get mutated or not (same probably goes for `on_file` and `on_dir`) and one less argument to get wrong.

otherlibs/stdune/src/path.ml
```diff
@@ -1469,6 +1472,11 @@ let drop_prefix_exn t ~prefix =
   | Some p -> p
 ;;
 
+let is_broken_symlink = function
+  | External e -> External.is_broken_symlink e
+  | _ -> false
```
Leonidas-from-XIV, line 1477: Is this actually correct? Can't other path variants have broken symlinks? It's possible that they cannot, but reading this casually this question pops up on my mind. Thus it would probably be better to either implement the check on other types as well, throw an error or have a comment why this is correct.

src/dune_rules/pkg_rules.ml
```diff
@@ -1627,9 +1627,14 @@ let source_rules (pkg : Pkg.t) =
...
-             let copy = loc, Action_builder.copy ~src ~dst in
-             Path.build dst :: source_files, copy :: rules)
+             if Path.is_broken_symlink src
+             then
+               (* Don't copy broken symlinks into the build directory. *)
+               source_files, rules
+             else (
+               let dst = Path.Build.append_local pkg.write_paths.source_dir file in
+               let copy = loc, Action_builder.copy ~src ~dst in
+               Path.build dst :: source_files, copy :: rules))
```
Leonidas-from-XIV, line 1637: To me it is a bit unclear why both this part ("don't copy broken symlinks") and the other part ("remove broken symlinks") are both necessary. I would expect that either would do the job and this part seems to be a bit nicer/less invasive.

test/blackbox-tests/test-cases/pkg/broken-symlink-in-dependency.t
```diff
@@ -0,0 +1,75 @@
...
+Define a package foo containing a broken symlink.
+  $ mkdir foo
+  $ touch foo/a.txt
+  $ ln -s non_existent foo/b.txt
+
+Define a package foo containing a broken symlink.
+  $ mkdir bar
+  $ touch bar/a.txt
+  $ ln -s non_existent bar/b.txt
+  $ tar pczf bar.tar.gz bar
```
Leonidas-from-XIV, line 15: I had to look up `p` and according to the man page this is `-p, --preserve-permissions, --same-permissions`. Is this required for the test? I'd probably just create a tarball without compression (to save CPU cycles compressing and decompressing).

Leonidas-from-XIV (commented):
I think the approach is really good because as you write, both Etienne and me tried to fix the underlying issue in the engine and got stuck so this workaround seems like a really nice pragmatic solution to sidestep the issue without breaking anything.

That said, I have some questions on how this works, I've written them in the comments.

otherlibs/stdune/src/fpath.ml
gridbugs: Agreed. I was following suite for the other arguments to this function. I'll rearrange it so only the changing arguments are passed to `loop` (in a separate commit so it's clearly not part of handling broken symlinks).

otherlibs/stdune/src/path.ml
gridbugs: I'll add a comment.

src/dune_rules/fetch_rules.ml
```diff
@@ -99,17 +99,29 @@ module Spec = struct
...
+      (* Delete any broken symlinks from the unpacked archive. *)
+      let target_abs = Path.to_absolute_filename target in
+      Fpath.traverse
+        ~init:()
+        ~dir:target_abs
+        ~on_dir:(fun ~dir:_ _ () -> ())
+        ~on_file:(fun ~dir:_ _ () -> ())
+        ~on_broken_symlink:(fun ~dir fname () ->
+          let path = Filename.concat target_abs (Filename.concat dir fname) in
+          Fpath.rm_rf path);
```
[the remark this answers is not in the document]
gridbugs: It is, but I don't think there's an async `rm` in dune. Further down in this file we do `Io.copy_file` inside an async context (not that that makes it correct to do so, but at least there's a precedent). The blocking `rm_rf` is expected to be called very infrequently - only while extracting packages with broken symlinks (which are quite rare).

test/blackbox-tests/test-cases/pkg/broken-symlink-in-dependency.t
gridbugs: You're right, somehow `p` made its way into my muscle memory for creating a tarball, but it is unnecessary here.

src/dune_rules/pkg_rules.ml
gridbugs: I tried creating a new action spec for just removing broken symlinks but I couldn't get it to work. When I sequence the new action after the fetch action (using progn) then I get this error between the file being extracted and the symlinks being removed:
```
Error: Error trying to read targets after a rule was run:
- checksum/md5=04023c482a2c1fba5b74ccbd4764a644/dir/b.txt: Broken symbolic link
```

Possibly it was important that the broken symlinks were being removed by the same action that created them, so as far as dune was concerned they never existed to begin with?

And looking through the `source_rules` function in `Pkg_rules` I think we'd have to carefully manage the `source_deps` set which that function populates. It already seems to work differently for local dirs and archives, with the former depending on the individual source files and the latter depending on the directory where the archive was extracted (I think?). It's simpler to just ignore broken symlinks from local directories than try to remove them from all the right places after adding them, but of course this is not an option for archives as they are extracted directly into their final destination in _build.

I'll add some comments explaining the situation but I don't see an easy way to unify this code.

src/dune_rules/fetch_rules.ml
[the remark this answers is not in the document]
Leonidas-from-XIV: For me the issue isn't so much the `rm` (but I see how the location where I placed my comment makes it confusing), my concern is rather that `Fpath.traverse` is blocking.

Alizter:
What about when a broken symlink is purposefully included for say testing reasons?

gridbugs:
> What about when a broken symlink is purposefully included for say testing reasons?

Dune doesn't run tests on a project's dependencies anyway so this change won't affect that use case.

test/blackbox-tests/test-cases/pkg/broken-symlink-in-dependency.t
```diff
@@ -0,0 +1,75 @@
...
+
+Make a project depending on foo.
+  $ mkdir x
+  $ cd x
+  $ cat > dune-project <<EOF
+  > (lang dune 3.17)
+  > (package
+  >  (name x)
+  >  (allow_empty)
+  >  (depends foo bar baz))
```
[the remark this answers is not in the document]
gridbugs: It will exist at the point that we build this project, but I've rearranged this file to make it easier to read.

src/dune_rules/fetch_rules.ml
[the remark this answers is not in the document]
gridbugs: Ah makes sense. Dune doesn't currently have an async way of traversing a directory structure but we can probably throw together an async traverse based on the `Dune_engine.Fs_memo` module.

Alizter:
Here is a related issue that I tried to tackle previously:
- https://github.com/ocaml/dune/issues/7573

I agree that we should probably just delete these for pkg.

src/dune_rules/fetch_rules.ml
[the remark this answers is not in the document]
gridbugs: I think let's defer this work. There are other instances of `Fpath.traverse` that we should also update, and I don't want this PR to get too drawn out.

gridbugs:
Yeh there have been a few attempts at fixing the issue in a general way that didn't end up getting merged.

Leonidas-from-XIV (approved):

PR #11588, opened 2025-03-31
chore(mirage): bump to setup-ocaml@v3

maiste:
Since last time, the mirage CI seems to be broken. It was using an old version of setup-ocaml. This PR moves to the new and stable version.

### DIFF
--- a/.github/workflows/mirage.yml
+++ b/.github/workflows/mirage.yml
@@ -14,7 +14,7 @@ jobs:
         repository: roburio/caldav
         ref: 51f0d150542348dc259b7c9f7bc70ee592243f7f
     - name: Use OCaml ${{ matrix.ocaml-compiler }}
-      uses: ocaml/setup-ocaml@v2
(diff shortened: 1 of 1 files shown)


Leonidas-from-XIV:
What is the error message and how does it fix it? Asking because we're not running the action as part of our usual CI checks so I want to make sure this fixes the problem.

Alizter:
Just add

```
  pull_request:
```

To the `on` field of the workflow, remembering to later remove it. See `workflow.yml` for an example.

maiste:
@Leonidas-from-XIV I think you can run if you specified a push on a specific PR. Anyway, I ran it on my own fork to avoid polluting the PR with it: https://github.com/maiste/dune/actions/runs/14196850569/job/39773910044

The difference is that `setup-ocaml@v3` does not try to install darcs, so the setup doesn't fail.

Leonidas-from-XIV (approved):
Great, thanks for clarifying. Seems sensible to me!

PR #11589, opened 2025-03-31
chore: upgrade dune-release to 2.1.0

maiste:
While preparing the release of `3.18`, I discovered there was a new version of `dune-release`. As it does not publish the documentation anymore, it is necessary to change the Makefile instructions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,7 +167,7 @@ dune-release:
 	dune-release tag
 	dune-release distrib --skip-build --skip-lint --skip-tests
 # See https://github.com/ocamllabs/dune-release/issues/206
-	DUNE_RELEASE_DELEGATE=github-dune-release-delegate dune-release publish distrib --verbose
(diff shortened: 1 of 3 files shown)


Leonidas-from-XIV:
Would it not make sense to upgrade to >= 2.1? I don't think we need the docs that it generates in any case, because we put our docs on RTD.

maiste:
As there is no version 2.0.0, the `< 2.0.0` will be equivalent to `2.1 >=`: this is just to keep styling uniformity.
Regarding the documentation, the `2.1.0` removes the `publish distrib` because there is no distinction between `distrib` and `doc`. The command doesn't take an argument anymore: hence, the replacement.

Leonidas-from-XIV (approved):
Ah sorry, I missed that the constraint change is in `conflicts` and not `depends`. This makes my complaint entirely purposeless because you did exactly what I had suggested :-)

PR #11594, opened 2025-04-02
Update CODEOWNERS

samoht:
Christine is not working on the Dune codebase anymore so removing here from here to avoid spam.


maiste:
Thanks for the change!
Considering `CODEOWNERS` was only use for this, shouldn't we drop the file and reintroduce it later if we need to?

maiste (approved):
Thanks :pray:

PR #11607, opened 2025-04-04
chore: remove Pp.newline

Alizter:
In this PR we remove all uses of `Pp.newline`.

We should avoid using `Pp.newline` since it will simply insert a newline character in the underlying formatting engine and this doesn't play well with the other kinds of breaks. I don't believe that there is any sensible use-case for this so I have opted to replace all mentions of `Pp.newline` with the intended equivalent.

The refactoring in `Loc` is a bit more advanced, but we basically update the code to use the more modern style.

Marking this as a draft PR for now until I can work out the kinks in the test-suite.


rgrinberg:
Please re-open once this is ready

Issue #11585, opened 2025-03-31
dune cache clear does not clear the toolchains cache


emillon:
Hi,
With the default configuration, `dune cache clear` leaves `~/.cache/dune/toolchains` in place.

I understand the technical reason, but it's a bit confusing. I think that the goal of `dune cache clear` is to replace `rm -r ~/.cache/dune` but it's now necessary again.

If removing the toolchains cache by default is not desirable, maybe an extra flag to remove this cache would work.

(noting the caveat in #11584 that these might not be in the same location)

PR #11612, opened 2025-04-07
use DUNE_CACHE_HOME instead of XDG_CACHE_HOME

### LINKED ISSUE #11584: Toolchains cache does not follow DUNE_CACHE_ROOT
Hi,
The toolchains cache is at the moment separate from the location of the shared dune cache.
This can lead to confusing behavior.

For example, suppose if a CI configures a custom `DUNE_CACHE_ROOT` mounted to a persistent location. They'll always have to start building a compiler because they're not restored from the main cache. It's possible that this cache miss cascades because of #11583 too.

I don't have a concrete solution for that; maybe if DUNE_CACHE_ROOT it should store the toolchains info there instead of a generic XDG solution.

cc @mefyl

ElectreAAS:
Fixes #11584 and should ease the pain of (but does NOT fix) #11585.

The build cache is currently set up to be at `$DUNE_CACHE_ROOT`, which itself will default on `$XDG_CACHE_HOME` if not present.
Three different caches skipped `$DUNE_CACHE_ROOT` and used the XDG directly: the toolchain cache, the git-repo cache, and the lmdb (rev store) cache.

This PR introduces a new variable: `$DUNE_CACHE_HOME` to allow customization of cache locations without breaking existing behaviour. To recap:
|   | Current behaviour | With this PR |
| :-: | :-------------: | :----------: |
| `DUNE_CACHE_HOME` defaults to | doesn't exist | `XDG_CACHE_HOME/dune` |
| `DUNE_CACHE_ROOT` defaults to | `XDG/dune/db` | `DUNE_CACHE_HOME/db` |
| Build cache | `DUNE_CACHE_ROOT` or `XDG/dune/db` | `DUNE_CACHE_ROOT` or `DUNE_CACHE_HOME/db` or `XDG/dune/db` |
| Git repo cache | `XDG/dune/git-repo` only | `DUNE_CACHE_HOME/git-repo` or `XDG/dune/git-repo` |
| Lmdb cache | `XDG/dune/rev_store` only | `DUNE_CACHE_HOME/rev_store` or `XDG/dune/rev_store` |
| Toolchains cache | `XDG/dune/toolchains` only | `DUNE_CACHE_HOME/toolchains` or `XDG/dune/toolchains` |

### DIFF
--- a/bin/common.ml
+++ b/bin/common.ml
@@ -1505,11 +1505,15 @@ let envs =
         ~doc:"If different than $(b,0), ANSI colors should be enabled no matter what."
         "CLICOLOR_FORCE"
     ; info
-        ~doc:"If set, determines the location of the machine-global shared cache."
-        "DUNE_CACHE_ROOT"
-    ; info
-        ~doc:"If set, determines the location of all the different caches used by dune."
+        ~doc:
+          "If set, determines the location of all the different caches used by dune. \
+           Defaults to XDG_CACHE_HOME/dune if unset."
         "DUNE_CACHE_HOME"
+    ; info
+        ~doc:
+          "If set, determines the location of the machine-global shared cache. This is \
+           the main build cache used by dune. Defaults to DUNE_CACHE_HOME/db if unset"
+        "DUNE_CACHE_ROOT"
     ]
 ;;
 
--- a/bin/common.ml
+++ b/bin/common.ml
@@ -1383,16 +1383,6 @@ let init_with_root ~(root : Workspace_root.t) (builder : Builder.t) =
         "Shared build cache location: %s"
         (Path.to_string (Lazy.force Dune_cache_storage.Layout.build_cache_dir))
     ];
-  Log.info
-    [ Pp.textf
-        "Toolchains cache location: %s"
-        (Path.to_string (Lazy.force Dune_util.toolchains_dir))
-    ];
-  Log.info
-    [ Pp.textf
-        "Revivision store cache location: %s"
-        (Path.to_string (Lazy.force Dune_util.rev_store))
-    ];
   Dune_rules.Main.init
     ~stats:c.stats
     ~sandboxing_preference:config.sandboxing_preference
--- a/bin/common.ml
+++ b/bin/common.ml
@@ -1386,12 +1386,12 @@ let init_with_root ~(root : Workspace_root.t) (builder : Builder.t) =
   Log.info
     [ Pp.textf
         "Toolchains cache location: %s"
-        (Path.to_string (Lazy.force Dune_cache_storage.Layout.toolchains_dir))
+        (Path.to_string (Lazy.force Dune_util.toolchains_dir))
     ];
   Log.info
     [ Pp.textf
         "Revivision store cache location: %s"
-        (Path.to_string (Lazy.force Dune_cache_storage.Layout.rev_store))
+        (Path.to_string (Lazy.force Dune_util.rev_store))
     ];
   Dune_rules.Main.init
     ~stats:c.stats
--- a/bin/common.ml
+++ b/bin/common.ml
@@ -1380,8 +1380,18 @@ let init_with_root ~(root : Workspace_root.t) (builder : Builder.t) =
     ];
   Log.info
     [ Pp.textf
-        "Shared cache location: %s"
-        (Path.to_string (Lazy.force Dune_cache_storage.Layout.root_dir))
+        "Shared build cache location: %s"
+        (Path.to_string (Lazy.force Dune_cache_storage.Layout.build_cache_dir))
+    ];
+  Log.info
+    [ Pp.textf
+        "Toolchains cache location: %s"
+        (Path.to_string (Lazy.force Dune_cache_storage.Layout.toolchains_dir))
+    ];
+  Log.info
+    [ Pp.textf
+        "Revivision store cache location: %s"
+        (Path.to_string (Lazy.force Dune_cache_storage.Layout.rev_store))
     ];
   Dune_rules.Main.init
     ~stats:c.stats
@@ -1505,8 +1515,11 @@ let envs =
         ~doc:"If different than $(b,0), ANSI colors should be enabled no matter what."
         "CLICOLOR_FORCE"
     ; info
-        "DUNE_CACHE_ROOT"
         ~doc:"If set, determines the location of the machine-global shared cache."
+        "DUNE_CACHE_ROOT"
+    ; info
+        ~doc:"If set, determines the location of all the different caches used by dune."
+        "DUNE_CACHE_HOME"
     ]
 ;;
 
--- a/src/dune_cache_storage/dune_cache_storage.ml
+++ b/src/dune_cache_storage/dune_cache_storage.ml
@@ -376,5 +376,5 @@ let clear () =
   rm_rf (Lazy.force Layout.temp_dir);
   (* Do not catch errors when deleting the root directory so that they are
      reported to the user. *)
-  Path.rmdir (Lazy.force Layout.root_dir)
+  Path.rmdir (Lazy.force Layout.build_cache_dir)
 ;;
--- a/src/dune_cache_storage/layout.ml
+++ b/src/dune_cache_storage/layout.ml
@@ -3,24 +3,17 @@ open Import
 
 let ( / ) = Path.relative
 
-(** The directory containing the build cache.
-    Uses [$DUNE_CACHE_ROOT] if set, or
-    [$DUNE_CACHE_HOME/db] if set, or
-    [Dune_util.default_cache_dir/db] otherwise. *)
 let build_cache_dir =
   lazy
-    (let high_priority_var = "DUNE_CACHE_ROOT" in
-     match Sys.getenv_opt high_priority_var with
+    (let var = "DUNE_CACHE_ROOT" in
+     match Sys.getenv_opt var with
      | Some path ->
-       Dune_util.check_absolute ~var:high_priority_var ~path;
+       if Filename.is_relative path
+       then
+         User_error.raise
+           [ Pp.paragraphf "$%s should be an absolute path, but is %S" var path ];
        Path.external_ (Path.External.of_string path)
-     | None ->
-       let low_priority_var = "DUNE_CACHE_HOME" in
-       (match Sys.getenv_opt low_priority_var with
-        | Some path ->
-          Dune_util.check_absolute ~var:low_priority_var ~path;
-          Path.external_ (Path.External.of_string path) / "db"
-        | None -> Lazy.force Dune_util.default_cache_dir / "db"))
+     | None -> Lazy.force Dune_util.cache_home_dir / "db")
 ;;
 
 let temp_dir = lazy (Lazy.force build_cache_dir / "temp")
--- a/src/dune_cache_storage/layout.ml
+++ b/src/dune_cache_storage/layout.ml
@@ -1,25 +1,50 @@
 open Stdune
 open Import
 
-let default_root_dir =
+let ( / ) = Path.relative
+
+(** The default directory of all caches. Defaults to [$XDG_CACHE_HOME/dune].
+    Not to be used directly, add a layer beforehand. *)
+let default_cache_dir =
   lazy
     (let cache_dir = Xdg.cache_dir (Lazy.force Dune_util.xdg) in
-     Path.L.relative (Path.of_filename_relative_to_initial_cwd cache_dir) [ "dune"; "db" ])
+     Path.of_filename_relative_to_initial_cwd cache_dir / "dune")
 ;;
 
-let root_dir =
+let check_absolute var path =
+  if Filename.is_relative path
+  then
+    User_error.raise
+      [ Pp.paragraphf "$%s should be an absolute path, but is %S" var path ]
+;;
+
+(** The home directory for all the caches. Defaults to [default_cache_dir],
+    or uses [$DUNE_CACHE_HOME] if set. *)
+let home_dir =
+  lazy
+    (let var = "DUNE_CACHE_HOME" in
+     match Sys.getenv_opt var with
+     | None -> Lazy.force default_cache_dir
+     | Some path ->
+       check_absolute var path;
+       Path.external_ (Path.External.of_string path))
+;;
+
+(** The directory of the build cache. Defaults to [home_dir/db],
+    or uses [$DUNE_CACHE_ROOT] if set. *)
+let build_cache_dir =
   lazy
     (let var = "DUNE_CACHE_ROOT" in
      match Sys.getenv_opt var with
-     | None -> Lazy.force default_root_dir
+     | None -> Lazy.force home_dir / "db"
      | Some path ->
-       if Filename.is_relative path
-       then failwith (sprintf "%s should be an absolute path, but is %s" var path);
-       Path.of_filename_relative_to_initial_cwd path)
+       check_absolute var path;
+       Path.external_ (Path.External.of_string path))
 ;;
 
-let ( / ) = Path.relative
-let temp_dir = lazy (Lazy.force root_dir / "temp")
+let rev_store = lazy (Lazy.force home_dir / "git-repo")
+let toolchains_dir = lazy (Lazy.force home_dir / "toolchains")
+let temp_dir = lazy (Lazy.force build_cache_dir / "temp")
 
 let cache_path ~dir ~hex =
   let two_first_chars = sprintf "%c%c" hex.[0] hex.[1] in
@@ -52,13 +77,15 @@ let list_entries ~storage =
 
 module Versioned = struct
   let metadata_storage_dir t =
-    lazy (Lazy.force root_dir / "meta" / Version.Metadata.to_string t)
+    lazy (Lazy.force build_cache_dir / "meta" / Version.Metadata.to_string t)
   ;;
 
-  let file_storage_dir t = lazy (Lazy.force root_dir / "files" / Version.File.to_string t)
+  let file_storage_dir t =
+    lazy (Lazy.force build_cache_dir / "files" / Version.File.to_string t)
+  ;;
 
   let value_storage_dir t =
-    lazy (Lazy.force root_dir / "values" / Version.Value.to_string t)
+    lazy (Lazy.force build_cache_dir / "values" / Version.Value.to_string t)
   ;;
 
   let metadata_path t ~rule_or_action_digest =
--- a/src/dune_cache_storage/layout.ml
+++ b/src/dune_cache_storage/layout.ml
@@ -3,47 +3,26 @@ open Import
 
 let ( / ) = Path.relative
 
-(** The default directory of all caches. Defaults to [$XDG_CACHE_HOME/dune].
-    Not to be used directly, add a layer beforehand. *)
-let default_cache_dir =
-  lazy
-    (let cache_dir = Xdg.cache_dir (Lazy.force Dune_util.xdg) in
-     Path.of_filename_relative_to_initial_cwd cache_dir / "dune")
-;;
-
-let check_absolute var path =
-  if Filename.is_relative path
-  then
-    User_error.raise
-      [ Pp.paragraphf "$%s should be an absolute path, but is %S" var path ]
-;;
-
-(** The home directory for all the caches. Defaults to [default_cache_dir],
-    or uses [$DUNE_CACHE_HOME] if set. *)
-let home_dir =
-  lazy
-    (let var = "DUNE_CACHE_HOME" in
-     match Sys.getenv_opt var with
-     | None -> Lazy.force default_cache_dir
-     | Some path ->
-       check_absolute var path;
-       Path.external_ (Path.External.of_string path))
-;;
-
-(** The directory of the build cache. Defaults to [home_dir/db],
-    or uses [$DUNE_CACHE_ROOT] if set. *)
+(** The directory containing the build cache.
+    Uses [$DUNE_CACHE_ROOT] if set, or
+    [$DUNE_CACHE_HOME/db] if set, or
+    [Dune_util.default_cache_dir/db] otherwise. *)
 let build_cache_dir =
   lazy
-    (let var = "DUNE_CACHE_ROOT" in
-     match Sys.getenv_opt var with
-     | None -> Lazy.force home_dir / "db"
+    (let high_priority_var = "DUNE_CACHE_ROOT" in
+     match Sys.getenv_opt high_priority_var with
      | Some path ->
-       check_absolute var path;
-       Path.external_ (Path.External.of_string path))
+       Dune_util.check_absolute ~var:high_priority_var ~path;
+       Path.external_ (Path.External.of_string path)
+     | None ->
+       let low_priority_var = "DUNE_CACHE_HOME" in
+       (match Sys.getenv_opt low_priority_var with
+        | Some path ->
+          Dune_util.check_absolute ~var:low_priority_var ~path;
+          Path.external_ (Path.External.of_string path) / "db"
+        | None -> Lazy.force Dune_util.default_cache_dir / "db"))
 ;;
 
-let rev_store = lazy (Lazy.force home_dir / "git-repo")
-let toolchains_dir = lazy (Lazy.force home_dir / "toolchains")
 let temp_dir = lazy (Lazy.force build_cache_dir / "temp")
 
 let cache_path ~dir ~hex =
--- a/src/dune_cache_storage/layout.mli
+++ b/src/dune_cache_storage/layout.mli
@@ -8,8 +8,19 @@
 open Stdune
 open Import
 
-(** The path to the root directory of the cache. *)
-val root_dir : Path.t Lazy.t
+(** The home directory of all the caches.
+    Uses [$DUNE_CACHE_HOME] if set. *)
+val home_dir : Path.t Lazy.t
+
+(** The directory of the build cache. Defaults to [home_dir/db], or
+    uses [$DUNE_CACHE_ROOT] if set. *)
+val build_cache_dir : Path.t Lazy.t
+
+(** The cache for the git repository. Is set to [home_dir/git-repo]. *)
+val rev_store : Path.t Lazy.t
+
+(** The cache for the toolchains. Is set to [home_dir/toolchains]. *)
+val toolchains_dir : Path.t Lazy.t
 
 (** Create a few subdirectories in [root_dir]. We expose this function because
     we don't want to modify the file system when the cache is disabled.
--- a/src/dune_cache_storage/layout.mli
+++ b/src/dune_cache_storage/layout.mli
@@ -9,9 +9,8 @@ open Stdune
 open Import
 
 (** The directory containing the build cache.
-    Uses [$DUNE_CACHE_ROOT] if set, or
-    [$DUNE_CACHE_HOME/db] if set, or
-    [Dune_util.default_cache_dir/db] otherwise. *)
+    Set to [$DUNE_CACHE_ROOT] if it exists, or
+    [Dune_util.cache_home_dir/db] otherwise. *)
 val build_cache_dir : Path.t Lazy.t
 
 (** Create a few subdirectories in [root_dir]. We expose this function because
--- a/src/dune_cache_storage/layout.mli
+++ b/src/dune_cache_storage/layout.mli
@@ -8,19 +8,11 @@
 open Stdune
 open Import
 
-(** The home directory of all the caches.
-    Uses [$DUNE_CACHE_HOME] if set. *)
-val home_dir : Path.t Lazy.t
-
-(** The directory of the build cache. Defaults to [home_dir/db], or
-    uses [$DUNE_CACHE_ROOT] if set. *)
+(** The directory containing the build cache.
+    Uses [$DUNE_CACHE_ROOT] if set, or
+    [$DUNE_CACHE_HOME/db] if set, or
+    [Dune_util.default_cache_dir/db] otherwise. *)
 val build_cache_dir : Path.t Lazy.t
-
-(** The cache for the git repository. Is set to [home_dir/git-repo]. *)
-val rev_store : Path.t Lazy.t
-
-(** The cache for the toolchains. Is set to [home_dir/toolchains]. *)
-val toolchains_dir : Path.t Lazy.t
 
 (** Create a few subdirectories in [root_dir]. We expose this function because
     we don't want to modify the file system when the cache is disabled.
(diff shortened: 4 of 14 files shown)


src/dune_rules/pkg_toolchain.mli
```diff
@@ -21,7 +21,7 @@ val is_compiler_and_toolchains_enabled : Package.Name.t -> bool
 
 (** Returns the path to the directory containing the given package within the
     toolchain directory. This will be something like
-    $XDG_CACHE_DIR/dune/toolchains/ocaml-base-compiler.5.2.1.XXXXXXXX where
```
ElectreAAS, line 24: Note that this was a typo, it's `XDG_CACHE_HOME`, not `DIR`

src/dune_pkg/rev_store.ml
```diff
@@ -913,11 +913,7 @@ let content_of_files t files =
 
 let get =
   Fiber_lazy.create (fun () ->
-    let dir =
-      Path.L.relative
-        (Path.of_string (Xdg.cache_dir (Lazy.force Dune_util.xdg)))
-        [ "dune"; "git-repo" ]
-    in
+    let dir = Path.L.relative Dune_cache_storage.Layout.root_dir [ "dune"; "git-repo" ] in
```
gridbugs, line 916: Does `Dune_cache_storage.Layout.root_dir` return the contents of `$XDG_CACHE_HOME` (or similar) if `$DUNE_CACHE_ROOT` is not set?

ElectreAAS: Yes, it calls `Xdg.cache_dir` which is set from the env var

Leonidas-from-XIV: Is this not adding one layer of `dune` too much? I would expect it to return `~/.cache/dune` if the directory was not set thus it would be `~/.cache/dune/dune/git-repo`.

ElectreAAS: Indeed, it would end up being `.cache/dune/db/dune/...`, so one layer of dune too many

src/dune_rules/pkg_toolchain.mli
```diff
@@ -21,7 +21,7 @@ val is_compiler_and_toolchains_enabled : Package.Name.t -> bool
 
 (** Returns the path to the directory containing the given package within the
     toolchain directory. This will be something like
-    $XDG_CACHE_HOME/dune/toolchains/ocaml-base-compiler.5.2.1.XXXXXXXX where
+    $DUNE_CACHE_ROOT/dune/toolchains/ocaml-base-compiler.5.2.1.XXXXXXXX where
```
Alizter: This doesn't make sense to me?

ElectreAAS:
I just rebased on main, fixed the conflicts, and addressed your comment @Leonidas-from-XIV.
Assuming the CI is green, I think this is good to go!

test/blackbox-tests/test-cases/pkg/rev-store-lock-linux.t
```diff
@@ -29,8 +29,9 @@ We set the project up to depend on `foo`
 There should be some kind of error message if getting the revision store lock
 fails (simulated here with a failing flock(2) call):
 
-  $ XDG_CACHE_HOME=$(pwd)/dune-workspace-cache strace -e inject=flock:error=EBADFD -o /dev/null dune pkg lock
+$ DUNE_CACHE_ROOT=$(pwd)/dune-workspace-cache
+  $ strace -e inject=flock:error=EBADFD -o /dev/null dune pkg lock
   Error: Failed to get a lock for the revision store at
-  $TESTCASE_ROOT/dune-workspace-cache/dune/rev-store.lock:
+  $TESTCASE_ROOT/.cache/dune/db/rev-store.lock:
```
Alizter: This doesn't look right

src/dune_pkg/rev_store.ml
```diff
@@ -1218,9 +1218,7 @@ let content_of_files t files =
 let get =
   Fiber.Lazy.create (fun () ->
     let dir =
-      Path.L.relative
-        (Path.of_string (Xdg.cache_dir (Lazy.force Dune_util.xdg)))
-        [ "dune"; "git-repo" ]
+      Path.L.relative (Lazy.force Dune_cache_storage.Layout.root_dir) [ "git-repo" ]
```
Leonidas-from-XIV: I think putting it inside `Dune_cache_storage` is a bit odd, since this is the artifact cache and dropping random content into it that it doesn't expect to be there looks like something that only accidentally works.

Leonidas-from-XIV:
I am generally not sure whether this is the right thing to do. This is because what we call "the dune cache" is a build artifact cache (and its location is controlled by `DUNE_CACHE_ROOT`). However the rev store is a different kind of cache and stored in a cache folder (like `$XDG_CACHE_ROOT/dune/git-repo`).

Thus for consistency to me it would make more sense to make `DUNE_CACHE_ROOT` basically `XDG_CACHE_ROOT/dune` and then have the dune cache in `$DUNE_CACHE_ROOT/db` (or so) and the rev store in `DUNE_CACHE_ROOT/git-repo` (or so). This will however break a number of assumptions about what's inside the folder pointed to by `DUNE_CACHE_ROOT`.

test/blackbox-tests/test-cases/pkg/toolchain-installation.t
```diff
@@ -78,12 +78,12 @@ name so the output is consistent across test runs.
 
 Attempt to build the project. This will fail due to the fake compiler
 but the fake compiler will end up installed as a toolchain package.
-  $ XDG_CACHE_HOME=$PWD/fake-cache DUNE_CONFIG__TOOLCHAINS=enabled build_pkg ocaml-base-compiler 2>&1 | remove_hash
+  $ DUNE_CACHE_ROOT=$PWD/fake-cache DUNE_CONFIG__TOOLCHAINS=enabled build_pkg ocaml-base-compiler 2>&1 | remove_hash
```
Alizter: Why did you need to change this? Shouldn't it still work?

ElectreAAS:
> I am generally not sure whether this is the right thing to do. This is because what we call "the dune cache" is a build artifact cache (and its location is controlled by `DUNE_CACHE_ROOT`). However the rev store is a different kind of cache and stored in a cache folder (like `$XDG_CACHE_ROOT/dune/git-repo`).
> 
> Thus for consistency to me it would make more sense to make `DUNE_CACHE_ROOT` basically `XDG_CACHE_ROOT/dune` and then have the dune cache in `$DUNE_CACHE_ROOT/db` (or so) and the rev store in `DUNE_CACHE_ROOT/git-repo` (or so). This will however break a number of assumptions about what's inside the folder pointed to by `DUNE_CACHE_ROOT`.

I think I understand where you're coming from, however wouldn't that be breaking? If someone has set `DUNE_CACHE_ROOT` to `foo`, then overnight their build cache would move from `foo/{files,meta,temp,...}` to `foo/db/{files,meta,temp,...}`
Not opposed to a breaking change per se, but I have a compromise option: introduce a new variable `DUNE_CACHE_HOME` (or so), have that be `xdg/dune`, have `DUNE_CACHE_ROOT` be `DUNE_CACHE_HOME/db`, and have the toolchains/rev-store caches in `HOME` instead of `ROOT`.

To recap:
|   | Current behaviour | *Outdated* With this PR | Breaking option | Compromise |
| :-: | :-------------: | :----------: | :-------------: | :--------: |
| `DUNE_CACHE_HOME` defaults to | doesn't exist | doesn't exist | doesn't exist | `XDG/dune` |
| `DUNE_CACHE_ROOT` defaults to | `XDG/dune/db` | `XDG/dune/db` |`XDG/dune` | `DUNE_CACHE_HOME/db` |
| Build cache | `DUNE_CACHE_ROOT` or `XDG/dune/db` | `DUNE_CACHE_ROOT` or `XDG/dune/db` | `DUNE_CACHE_ROOT/db` or `XDG/dune/db` | `DUNE_CACHE_ROOT` or `DUNE_CACHE_HOME/db` or `XDG/dune/db` |
| Rev-store cache | `XDG/dune/git-repo` only | `DUNE_CACHE_ROOT/git-repo` or `XDG/dune/db/git-repo` | `DUNE_CACHE_ROOT/git-repo` or `XDG/dune/git-repo` | `DUNE_CACHE_HOME/git-repo` or `$XDG/dune/git-repo` |
| Toolchains cache | `XDG/dune/toolchains` only | `DUNE_CACHE_ROOT/toolchains` or `XDG/dune/db/toolchains` | `DUNE_CACHE_ROOT/toolchains` or `XDG/dune/toolchains` | `DUNE_CACHE_HOME/toolchains` or `XDG/dune/toolchains` |
| | | | | |
| Pros | status quo | it's done and more or less correct | it's actually correct and simple | allows users to put both caches in separate places, also correct |
| Cons | it's a bug | heterogeneous contents of cache | breaks some existing user setups | introduces a new variable which can be confusing |

I hope I got that right.

I was of the opinion that the breaking option would be better before making that table, now I'm in the 'compromise' camp. However I think having more env vars can be detrimental in the long run, especially if they're vaguely named (for example, the comment above wrote `XDG_CACHE_ROOT` instead of `XDG_CACHE_HOME`). We could however in the future completely deprecate `DUNE_CACHE_ROOT` in favor of like `DUNE_BUILD_CACHE & DUNE_PKG_CACHE` to make things clearer.

Thoughts?

Alizter:
I think the idea of introducing a `DUNE_CACHE_HOME` variable, appropriately documented, is sound.

Removing `DUNE_CACHE_ROOT` or changing its meaning will take some work unfortunately. Its probably best we don't touch it.

If `DUNE_CACHE_HOME` is set, then the build cache root should be in there, unless `DUNE_CACHE_ROOT` is set in which case it should be that.

Finer control for the other kinds of caches dune creates i.e. toolchains and the lmdb cache can be added if the need ever arises, but I am against adding that now. Those should just live in their subdirectories in `DUNE_CACHE_HOME`.

Leonidas-from-XIV:
Thanks for making the table @ElectreAAS, it is a *great* overview! Once implemented that explanation could even go into the documentation because it explains which locations and environment variables are used when.

> I was of the opinion that the breaking option would be better before making that table, now I'm in the 'compromise' camp. 

Looking through the table, I also find myself nodding at the things proposed in the "compromise" column. It seems fairly reasonable and allows to set all caches to a specific location at once (and maybe `git-repo` should be renamed into `revstore` or something like that, but that's besides the point).

In particular I believe there are two no-go's for me:
  * We shouldn't put other caches in the same folder as the build cache. I consider the contents of `db` to be "opaque", for all intents and purposes this could be a single binary file. The fact that it is a folder is more an implementation detail, just like we don't store files in the `.git` folder, just because its possible.
  * `DUNE_CACHE_ROOT` has always pointed at the build cache since historically that was the only one. I think this needs to stay, to avoid breaking existing usage. We can consider printing a warning when used akin to "please update to `DUNE_CACHE_HOME` or if you want to set the build cache specifically use `DUNE_BUILD_CACHE_ROOT`" but I think there might be legitimate reasons why you'd want to put the build cache in a different place (maybe an NFS share or so).

The "compromise" solution seems to address all my concerns. I also agree with @Alizter that keeping the build cache configurable needs to stay but the other caches can for now just be controlled via `DUNE_CACHE_HOME` until some need arises.

ElectreAAS:
I've went ahead and implemented the 'compromise' option. We now have DUNE_CACHE_HOME & DUNE_CACHE_ROOT, and tests should be correct by now.
I've put off updating the upmost text in this PR until we're satisfied, and also put off updating the docs until a later PR, as [it's not in a great state right now](https://github.com/ocaml/dune/blob/main/doc/caching.rst?plain=1#L5), and this is the occasion to fix it.

src/dune_cache_storage/dune_cache_storage.ml
ElectreAAS, line 1: Note: I haven't changed the semantics of `dune cache clear` so it would also delete the revstore & toolchains cache, but that's debatable

Leonidas-from-XIV: That's a good question. In theory it changes the behavior but I guess it does make sense that `dune cache clear` removes all the cached data. Before we didn't have a way to clean out the rev-store nor the toolchains so I'd argue that this is an improvement.

One could extend the command to allow clearing only selected caches, but I don't see an urgent need for that, so I'd be willing to wait for users actually requesting this.

shonfeder: Any kind of change to the caching command seems like it should land separately in any case.

bin/common.ml
```diff
@@ -1505,8 +1515,11 @@ let envs =
         ~doc:"If different than $(b,0), ANSI colors should be enabled no matter what."
         "CLICOLOR_FORCE"
     ; info
-        "DUNE_CACHE_ROOT"
         ~doc:"If set, determines the location of the machine-global shared cache."
+        "DUNE_CACHE_ROOT"
+    ; info
+        ~doc:"If set, determines the location of all the different caches used by dune."
+        "DUNE_CACHE_HOME"
```
shonfeder, line 1511: I think a clarification of the docs is in order, explaining briefly 

- for `DUNE_CACHE_ROOT` what kind of thing goes on the `db` and so what exactly this is configuring
- that `DUNE_CACHE_ROOT` takes precedence over `DUNE_CACHE_HOME`
- that these env vars take precedence over settints in the config

src/dune_cache_storage/layout.ml
```diff
@@ -1,25 +1,50 @@
 open Stdune
 open Import
 
-let default_root_dir =
+let ( / ) = Path.relative
+
+(** The default directory of all caches. Defaults to [$XDG_CACHE_HOME/dune].
+    Not to be used directly, add a layer beforehand. *)
```
shonfeder: I am not sure what "add a layer beforehand" means. Could you explain?

src/dune_cache_storage/layout.ml
```diff
@@ -1,25 +1,50 @@
...
+  if Filename.is_relative path
+  then
+    User_error.raise
+      [ Pp.paragraphf "$%s should be an absolute path, but is %S" var path ]
 ;;
 
-let root_dir =
+(** The home directory for all the caches. Defaults to [default_cache_dir],
+    or uses [$DUNE_CACHE_HOME] if set. *)
+let home_dir =
```
shonfeder: Just to avoid possible confusion, I think this would be clearer:

```suggestion
let cache_home_dir =
```

because "home dir" means something quite specific, and this could easily be confused by readers just encountering this variable where it is used.

src/dune_cache_storage/layout.ml
```diff
@@ -1,25 +1,50 @@
...
+  lazy
+    (let var = "DUNE_CACHE_HOME" in
+     match Sys.getenv_opt var with
+     | None -> Lazy.force default_cache_dir
+     | Some path ->
+       check_absolute var path;
+       Path.external_ (Path.External.of_string path))
```
shonfeder: Genuine question: How do we decide it better to force `default_cache_dir` here rather than to return the other branch as lazy?

src/dune_cache_storage/layout.ml
```diff
@@ -1,25 +1,50 @@
...
-       then failwith (sprintf "%s should be an absolute path, but is %s" var path);
-       Path.of_filename_relative_to_initial_cwd path)
+       check_absolute var path;
+       Path.external_ (Path.External.of_string path))
 ;;
 
-let ( / ) = Path.relative
-let temp_dir = lazy (Lazy.force root_dir / "temp")
+let rev_store = lazy (Lazy.force home_dir / "git-repo")
+let toolchains_dir = lazy (Lazy.force home_dir / "toolchains")
```
Leonidas-from-XIV: To be honest, I don't think the rev_store location nor the toolchains dir should be defined here. In my mind `Dune_cache_storage` is just the build cache (despite the name).

src/dune_pkg/rev_store.ml
```diff
@@ -1222,11 +1222,7 @@ let content_of_files t files =
 
 let get =
   Fiber.Lazy.create (fun () ->
-    let dir =
-      Path.L.relative
-        (Path.of_string (Xdg.cache_dir (Lazy.force Dune_util.xdg)))
-        [ "dune"; "git-repo" ]
-    in
+    let dir = Lazy.force Dune_cache_storage.Layout.rev_store in
```
Leonidas-from-XIV: Relatedly this should not need to be aware of the build cache.

src/dune_rules/pkg_toolchain.ml
```diff
@@ -2,20 +2,15 @@ open Import
 
 let base_dir () =
   let cache_dir =
-    Lazy.force Dune_util.xdg |> Xdg.cache_dir |> Path.Outside_build_dir.of_string
+    Lazy.force Dune_cache_storage.Layout.toolchains_dir |> Path.as_outside_build_dir_exn
```
Leonidas-from-XIV: Likewise this.

Leonidas-from-XIV (commented):
It's just a preference but I think it would make sense to keep the different types of caches in their respective places and not introduce dependencies on each other (especially as they aren't functional dependencies, just file paths).

I'd rather suggest creating `dune_cache_dir` or so to contain the logic and then `Dune_cache_storage.Layout`, `Dune_pkg.Rev_store` and `Dune_rules.Pkg_toolchain` can use this.

Alizter:
I've also suggested a common point, but I think `Dune_util` is better than introducing another new library for this.

src/dune_cache_storage/layout.mli
```diff
@@ -8,8 +8,19 @@
 open Stdune
 open Import
 
-(** The path to the root directory of the cache. *)
-val root_dir : Path.t Lazy.t
+(** The home directory of all the caches.
+    Uses [$DUNE_CACHE_HOME] if set. *)
+val home_dir : Path.t Lazy.t
```
shonfeder:
```suggestion
val cache_home_dir : Path.t Lazy.t
```

src/dune_cache_storage/layout.mli
```diff
@@ -8,8 +8,19 @@
 open Stdune
 open Import
 
-(** The path to the root directory of the cache. *)
-val root_dir : Path.t Lazy.t
+(** The home directory of all the caches.
+    Uses [$DUNE_CACHE_HOME] if set. *)
```
shonfeder: Should this not document that it defaults to `Xdg.cache_dir`?

src/dune_cache_storage/layout.ml
```diff
@@ -1,25 +1,50 @@
...
+let temp_dir = lazy (Lazy.force build_cache_dir / "temp")
```
shonfeder: I'm not totally sure it is preferable to move these values here: it makes sense to me that, e.g., the `Rev_store` modules is the only one that knows the name of it's subdir in the cache home location. And moving that detail into the general purpose cache storage modules seems kind of like an abstraction leak.

Could you explain you rationale for moving the values here?

shonfeder (commented):
Part way done with review.

src/dune_cache_storage/layout.ml
shonfeder: Oh, I didn't see @Leonidas-from-XIV's review. A related thought, tho he introduces a new complication in the organization: if `Dune_cache_storage` is only supposed to be the build cache, than other modules shouldn't go it to find the `cache_home_dir`. 

Perhaps the directory deriving functions here should be moved to a small module that is focused on path locations. Where is the best pre-existing place for that?

Alizter: I've also mentioned above that `Dune_util` is a natural common point for some of these directories. I think that is the most sensible choice here.

ElectreAAS:
I have moved most logic from `dune_cache_storage/layout` into `dune_util`. Indeed the reasoning behind the previous state was simply "I assumed the layout was for all the caches, not just the build one".
I hope the comments are clearer now.

bin/common.ml
ElectreAAS: Agreed, but I'm not clear myself on what exactly is in which type of cache...

src/dune_cache_storage/layout.ml
ElectreAAS: Moved to dune_util

src/dune_cache_storage/layout.ml
ElectreAAS: Moved things to dune_util as advised

bin/common.ml
```diff
@@ -1380,8 +1380,18 @@ let init_with_root ~(root : Workspace_root.t) (builder : Builder.t) =
...
+        (Path.to_string (Lazy.force Dune_cache_storage.Layout.build_cache_dir))
+    ];
+  Log.info
+    [ Pp.textf
+        "Toolchains cache location: %s"
+        (Path.to_string (Lazy.force Dune_util.toolchains_dir))
+    ];
+  Log.info
+    [ Pp.textf
+        "Revivision store cache location: %s"
```
Alizter:
```suggestion
        "Revision store location: %s"
```

Alizter:
Let's summarise the cache's we have in dune for clarity and reference:

`DUNE_CACHE_ROOT` is the location for the shared build cache. When dune builds something with `--cache enabled` (the default), the build artefacts are placed in the shared cache via a digest of their actions and are hard-linked to the build. This allows them to be easily reused across multiple builds. This is what we typically mean by the "dune cache". The commands `dune cache` are about managing this. The variable will default to `$DUNE_CACHE_HOME/db` when unset and can be set independently.

Since package management we have the revision store which is a big git repo containing all the git repos we use. When you download opam or any other package, instead of cloning these repos separately we fetch the files directly to the revision store. This allows us to have a common cache of all downloaded repos. It will live in `$DUNE_CACHE_HOME/git-repo`.

The revision store contains another cache called the lmdb cache which is used to cache expensive lookups. We typically need to see what files are in a given repo at a given revision and what the content of those files are. This requires running some external git commands with a non-trivial amount of overhead. This will live in `$DUNE_CACHE_HOME/rev_store`.

Finally there is the toolchains cache where we put non-relocatable packages such as the compiler. That lives in `$DUNE_CACHE_HOME/toolchains`.

bin/common.ml
Alizter: Until we have a way of setting these separately there is no point printing them as if they are separately configured. We should just print `DUNE_CACHE_HOME` instead.

bin/common.ml
Alizter: I've provided a summary in the main thread of the PR.

I would move this above root and in root metnion that it is a build cache and not any other type.

You can also mention that DUNE_CACHE_HOME defaults to XDG_CACHE_DIR or something.

bin/common.ml
ElectreAAS: The second print was a mistake, as this exact line is also logged in [rev_store.ml](https://github.com/ocaml/dune/pull/11612/files#diff-2103a7217c1c88039d8b5d207b016214c35a2f335c9d71258b061b399ac27177R160)

src/dune_util/dune_util.ml
```diff
@@ -38,6 +38,38 @@ let override_xdg : Xdg.t -> unit =
...
+     match Sys.getenv_opt var with
+     | Some path ->
+       check_absolute ~var ~path;
+       Path.external_ (Path.External.of_string path)
+     | None -> Lazy.force default_cache_dir)
+;;
+
+let git_repo_cache = lazy (Lazy.force cache_home_dir / "git-repo")
+let revision_store_cache = lazy (Lazy.force cache_home_dir / "rev_store")
+let toolchains_cache = lazy (Lazy.force cache_home_dir / "toolchains")
```
shonfeder: I thought what we discussed in previous review rounds was to move just the `cache_home_dir` logic into this module, but leave it up to the respective modules that put stuff into the cache to define their own sub-directory locations.

This seems advisable to me because it is arguable that the actual directory location is just an implementation detail, and we don't want to leak this out of the abstraction of the module that uses the cache.

On the other hand, if all of this is getting defined in the `Dune_util` repo, then why not also put https://github.com/ocaml/dune/pull/11612/files#diff-cf8c393f1e23171aed4b1d7e609eb5a22ba6fd2cc9acc74cdb0d3b9a32112920R23-R26 here?

Given my current understanding, the right way to organize this seems to be the following:

- `Dune_util` has a value `cache_home_dir`
- The parts of the code that use a sub-directory in this cache defines the concrete value (e.g., the `Rev_store` defiles its `Rev_store.cache_dir`, and same with the `Dune_cache`).

An alternative approach is to have all the cache dirs defined here, but then so too should https://github.com/ocaml/dune/pull/11612/files#diff-cf8c393f1e23171aed4b1d7e609eb5a22ba6fd2cc9acc74cdb0d3b9a32112920R10 , to be consistent.

But either way, I think we should have a consistent scheme for this organized, and I do see that happening currently (which could mean I misunderstand, but then I need an explanation).

src/dune_cache_storage/layout.ml
```diff
@@ -1,25 +1,29 @@
...
+    (let high_priority_var = "DUNE_CACHE_ROOT" in
+     match Sys.getenv_opt high_priority_var with
      | Some path ->
-       if Filename.is_relative path
-       then failwith (sprintf "%s should be an absolute path, but is %s" var path);
-       Path.of_filename_relative_to_initial_cwd path)
+       Dune_util.check_absolute ~var:high_priority_var ~path;
+       Path.external_ (Path.External.of_string path)
+     | None ->
+       let low_priority_var = "DUNE_CACHE_HOME" in
```
shonfeder: I don't think we should be reading from `DUNE_CACHE_HOME` from the env in two separate places. This should use the `Dune_util.cache_home_dir`.

src/dune_cache_storage/layout.ml
```diff
@@ -1,25 +1,29 @@
...
+       (match Sys.getenv_opt low_priority_var with
+        | Some path ->
+          Dune_util.check_absolute ~var:low_priority_var ~path;
```
shonfeder: Rather thank exporting `check_absolute`, I think we should define a `Dune_util.default_build_cache_dir`, letting us reuse all the path defaulting logic etc. Then all the logic here can be

```ocaml
Lazy.force Dune_util.default_build_cache_dir / "db"
```

Just like the other places using cache dirs in their respective modules (as I've suggested).

And all the checking of env vars and computing defaults would be together in the `Dune_util` dir, and no need to export `check_absolute`.

src/dune_util/dune_util.mli
```diff
@@ -0,0 +1,40 @@
...
+
+(** The cache for expensive git revision operations.
+    Uses [$DUNE_CACHE_HOME/rev_store] if set, or
+    [default_cache_dir/rev_store] otherwise. *)
+val revision_store_cache : Path.t Lazy.t
+
+(** The cache for the toolchains, containing non-relocatable packages.
+    Uses [$DUNE_CACHE_HOME/toolchains] if set, or
+    [default_cache_dir/toolchains] otherwise. *)
+val toolchains_cache : Path.t Lazy.t
```
shonfeder: If we move these into the respective components where they are used, I think we also don't need to document the concrete locations, which seems like a plus to me, since these should be private to the locations that use them, I think?

shonfeder (changes requested):
I still think the organization seems a bit off, and isn't what I'd meant to advocate for in my prior review. I'm sorry for the lack of clarity before! In case you think my suggestions here are missing something, I'm happy to discuss.

src/dune_util/dune_util.ml
ElectreAAS: Indeed there was a misunderstanding, should be better now

Alizter:
This PR is still marked as fixing #11585. Is that still the case?

src/dune_cache_storage/layout.ml
ElectreAAS: I don't think I agree with `Dune_util.default_build_cache_dir`, I think having **only** `cache_home_dir` implements a clearer separation of concerns. See my latest changes for details.

Also `check_absolute` is a trivial function, and the same behaviour exists in a lot of other places, where it's inlined [example 1](https://github.com/ocaml/dune/blob/main/otherlibs/dune-rpc/private/exported_types.ml#L397), [example 2](https://github.com/ocaml/dune/blob/main/otherlibs/stdune/src/path.ml#L511). Before this PR it was also inlined, and I've moved it back to being inlined here to avoid exporting it

src/dune_util/dune_util.mli
ElectreAAS: Indeed it's a plus

ElectreAAS:
> This PR is still marked as fixing #11585. Is that still the case?

Ah this was an edit by someone else that I didn't catch. No I don't fix that one. Adding `dune cache clear --all` or something is beyond the scope of this PR

Alizter:
If you update the PR description and importantly remove the `fix #asdf` it should resolve it.

Leonidas-from-XIV:
Oh, my apologies. It was me who added it, I thought that `dune clean` would remove whatever `DUNE_CACHE_HOME` is pointing to.

Alizter:
In the meeting we discussed perhaps biting the bullet and changing the definition of `DUNE_CACHE_ROOT`.

Currently we are not aware of any actual breakage, but it will mean that users existing caches will be invalidated. This is already the case when we change action digests so perhaps isn't a big deal.

If we change the meaning, then a user may have `DUNE_CACHE_ROOT` pointed at `foo/bar` which means the shared build cache is directly available in `foo/bar`. However now it would be available in `foo/bar/db`. That doesn't seem so bad in practice.

This has the advantage that setup-ocaml will not need to change the variable, we don't introduce cruft and nothing is broken.

ElectreAAS:
> In the meeting we discussed perhaps biting the bullet and changing the definition of `DUNE_CACHE_ROOT`.
> 
> Currently we are not aware of any actual breakage, but it will mean that users existing caches will be invalidated. This is already the case when we change action digests so perhaps isn't a big deal.
> 
> If we change the meaning, then a user may have `DUNE_CACHE_ROOT` pointed at `foo/bar` which means the shared build cache is directly available in `foo/bar`. However now it would be available in `foo/bar/db`. That doesn't seem so bad in practice.
> 
> This has the advantage that setup-ocaml will not need to change the variable, we don't introduce cruft and nothing is broken.

That would mean implementing the 'Breaking' option instead of the compromise one, right?
I tried my hand at doing just that [(link to my branch)](https://github.com/ElectreAAS/dune/tree/push-ztpwrqmsuxtk), if needed a can make an alternative PR with that.

Alizter:
> That would mean implementing the 'Breaking' option instead of the compromise one, right? I tried my hand at doing just that [(link to my branch)](https://github.com/ElectreAAS/dune/tree/push-ztpwrqmsuxtk), if needed a can make an alternative PR with that.

Yes, it would be that proposal. However we should be clear that we don't actually know if it causes any breakages other than missing old caches. I think a separate PR would be good and we can weigh the merits of both.

ElectreAAS:
#12858 has been opened with the 'breaking' option

ElectreAAS:
Closed in favor of #12858, which implements the 'breaking' option discussed above

PR #11613, opened 2025-04-07
Introduce a flag for binary dev tools

Sudha247:
Sometimes the ocaml-lsp binary distro doesn't work. This would need some investigation to find the reason. Meanwhile, this PR introduces a new flag to disable it if needed.

### DIFF
--- a/bin/lock_dev_tool.ml
+++ b/bin/lock_dev_tool.ml
@@ -74,10 +74,9 @@ let solve ~dev_tool ~local_packages =
     >>| Option.some
   and* workspace =
     let+ workspace = Workspace.workspace () in
-    if (Lazy.force Bin_dev_tools.is_enabled) then
-     Workspace.add_repo workspace Dune_pkg.Pkg_workspace.Repository.binary_packages
-    else
-      workspace
+    match Config.get Dune_rules.Compile_time.bin_dev_tools with
+      | `Enabled -> Workspace.add_repo workspace Dune_pkg.Pkg_workspace.Repository.binary_packages
+      | `Disabled -> workspace
     in
   let lock_dir = Lock_dir.dev_tool_lock_dir_path dev_tool in
     Memo.of_reproducible_fiber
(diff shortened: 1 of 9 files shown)


bin/lock_dev_tool.ml
```diff
@@ -72,11 +72,17 @@ let solve ~dev_tool ~local_packages =
     |> Dune_pkg.Sys_poll.solver_env_from_current_system
     |> Memo.of_reproducible_fiber
     >>| Option.some
-  and* workspace = Workspace.workspace () in
+  and* workspace =
+    let+ workspace = Workspace.workspace () in
+    if (Lazy.force Bin_dev_tools.is_enabled) then
```
Leonidas-from-XIV, line 77: What is the purpose of `Bin_dev_tools.in_enabled` here? It seems like an unnecessary indirection when you can query `Dune_rules.Compile_time.bin_dev_tools` instead.

boot/configure.ml
```diff
@@ -101,6 +101,11 @@ let () =
       , " Enable ocamlformat dev-tool, allows 'dune fmt' to build ocamlformat and use \
          it, independently from the project depenedencies .\n\
         \      This flag is experimental and shouldn't be relied on by packagers." )
+    ; ( "--bin-dev-tools"
+      , toggle "bin_dev_tools"
+      , " Enable ocamlformat dev-tool, allows 'dune fmt' to build ocamlformat and use \
+         it, independently from the project depenedencies .\n\
```
Leonidas-from-XIV, line 107: I don't think the description is correct; this flag seems to be about implicitly adding the binary package opam repository.

Leonidas-from-XIV (changes requested):
I think the idea is good to make it possible to opt-out, but this PR needs some improvements still :)

bin/lock_dev_tool.ml
maiste: Indeed, I didn't realize it when we took a look at it, but the pattern matching can directly happen here.

maiste (approved):

Leonidas-from-XIV (approved):
Code looks good but can you sign off your commits @Sudha247? Than it should be ready to merge.

maiste:
(`git rebase main --signoff` :smile: )

gridbugs (approved):

maiste (approved):

PR #11616, opened 2025-04-08
pkg: allow arbitrary opam repo layouts

### LINKED ISSUE #11615: Dune can only understand opam repos matching the layout of opam-repository
Opam repositories do not need to conform to the directory hierarchy convention set by opam-repository. This is documented in [opam's manual](https://opam.ocaml.org/doc/1.2/Manual.html#Layout). However dune makes some assumptions that all opam repos will organize their packages in a directory hierarchy like "packages/<name>/<name>.<version>/opam". In reality as long as every `opam` file under "packages" has a parent directory named like `<name>.<version>` it should work.

## Expected Behavior

[Jane Street's bleeding edge repo](https://github.com/janestreet/opam-repository/tree/master) is an example of an opam repo that we want to support with dune package management that uses a different directory layout than opam-repo.

## Actual Behavior

Currently any packages in an opam repo that aren't located in the expected location as per opam-repository's layout convention are invisible to dune; it won't consider them when solving dependencies.

## Reproduction

Put this in your dune-workspace to add Jane Street's bleeding edge repo to your project's opam repos:
```
(lang dune 3.17)

(repository
 (name janestreet-bleeding)
 (url git+https://github.com/janestreet/opam-repository.git))

(lock_dir
 (repositories upstream overlay janestreet-bleeding))
```
Try depending on a package from that repo in dune-project, and dune will be unable to resolve the dependency (or resolve it to a package with the same name from the upstream or overlay repo).

gridbugs:
Opam repos can have fairly arbitrary directory layouts. The only requirement is that all package manifests are under the top-level "packages" directory, and each manifest is in a file named "opam" whose parent directory is named like "<name>.<version>".

Prior to this change dune assumed a layout similar to the official opam-repository, where manifests are at
"packages/<name>/<name>.<version>/opam".

This generalizes dune's expectations of opam repositories such that arbitrary directory layouts are supported.

Fixes https://github.com/ocaml/dune/issues/11615

### DIFF
--- a/src/dune_pkg/opam_repo.ml
+++ b/src/dune_pkg/opam_repo.ml
@@ -3,19 +3,6 @@ open Fiber.O
 
 module Paths = struct
   let packages = Path.Local.of_string "packages"
-
-  let package_root package_name =
-    OpamPackage.Name.to_string package_name |> Path.Local.relative packages
-  ;;
-
-  let package_dir package =
-    Path.Local.relative
-      (package_root (OpamPackage.name package))
-      (OpamPackage.to_string package)
-  ;;
-
-  let files_dir package = Path.Local.relative (package_dir package) "files"
-  let opam_file package = Path.Local.relative (package_dir package) "opam"
 end
 
 module Serializable = struct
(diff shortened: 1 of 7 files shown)


maiste (commented):
Thanks for the PR! I have noticed few typos, but I mostly have questions regarding the errors handling. My main question is shouldn't we prevent the users from using a corrupted opam repository?

test/blackbox-tests/test-cases/pkg/opam-repo-layouts.t
```diff
@@ -0,0 +1,105 @@
...
+  "$TESTCASE_ROOT/custom-opam-repo"
+  contains an opam file at location "packages" which cannot be interpreted as
+  an opam package. Opam files must be contained in directories named like
+  <name>.<version>.
+  Solution for dune.lock:
+  - a.1
+  - b.2
+  - c.3
+
+  $ rm custom-opam-repo/packages/opam
```
Leonidas-from-XIV, line 105: What is this action for? It seems like there should be some kind of test but isn't.

test/blackbox-tests/test-cases/pkg/opam-repo-layouts.t
```diff
@@ -0,0 +1,105 @@
...
+  Solution for dune.lock:
+  - a.1
+  - b.2
+  - c.3
+
+  $ rm -r custom-opam-repo/packages/foo/bar/baz/qux/c.3
+
+Create an opam file in a directory that doesn't have a version number. There's
+no way to refer to this package in dune-project but dune should still be able
+to parse the repo with crashing.
```
Leonidas-from-XIV, line 72:
```suggestion
to parse the repo without crashing.
```

Maybe?

Leonidas-from-XIV (commented):
So I think the code is fine mostly and conceptually it is the right thing, as it enables us to use the Jane Street OPAM repo and any repo really, which is definitely neat.

However, I wonder if the complexity and cost in loading the whole repo into memory is worth it. Realistically people are going to use `opam-repository`, our overlays whose layout we control ourselves. So maybe the Jane Street repo layout could be added when inspecting the repo (e.g. how many layers are there between `package/**/<name.version>/opam` and then forcing the rest of the repo to match instead of allowing any nesting for any `opam` file), thus changing very little about what we currently do, basically just some additional processing in `Opam_repo.Paths`.

It is less flexible but I would argue that the added flexibility is not all that useful (OPAM suffers from this already, where it allows a lot of flexibility, sometimes to its detriment) and the price might be quite high.

I don't have a strong opinion on this however, so if people think that the added complexity, processing time and memory usage is worth it, I am okay with that.

src/dune_pkg/opam_repo.ml
```diff
@@ -41,24 +28,220 @@ module Serializable = struct
...
+                        existing_entry.opam_file_path_relative_to_repo_root)
+                 ];
+               Some existing_entry)))
+  ;;
+
+  (* Traverse a directory hierarchy looking for files named "opam". *)
+  let of_repo_dir_path opam_repo_dir_path ~loc : t =
+    let repo_name_for_messages = Path.to_absolute_filename opam_repo_dir_path in
+    try
+      Fpath.traverse_files
```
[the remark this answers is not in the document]
rgrinberg: Fiber or not, it doesn't seem like a good trade-off to introduce such an expensive step that would benefit so few users. Can you either make this optional or introduce a lazy loading scheme for the JST repo?

rgrinberg (changes requested):

gridbugs:
> Thanks for the PR! I have noticed few typos, but I mostly have questions regarding the errors handling. My main question is shouldn't we prevent the users from using a corrupted opam repository?

If a repo does not follow the convention set by opam-repository, that doesn't mean it is corrupt. The [documentation on repo layout in the opam manual](https://opam.ocaml.org/doc/1.2/Manual.html#Layout) specifies that any layout where opam files are contained in directories named after the package is acceptable. The document also suggests a potential future change to the layout of opam-repository where another level of hierarchy is introduced to group packages by common prefixes, e.g. `packages/p/pa/package-name/package-name.version`.

I agree with the sentiment that in practice we're probably only ever going to see repos formatted like opam-repository or JST, and the complexity/runtime cost of fully general opam repos is too high to be worth it. I briefly thought about introducing a fast-path where if a repo is formatted like opam-repository then we can make some optimizing assumptions, but the only way to be sure that a repo is formatted like opam-repository is to do a full scan of all its opam files which is just as expensive as not having a fast-path to begin with.

So I'm thinking I'll defer solving https://github.com/ocaml/dune/issues/11615 until I have more time to work on it, and eventually we can use a solution more specialized to the layouts we see in opam-repository and JST.

Alternatively, @rgrinberg how feasible would it be to ask Jane Street to change the layout of their repos to match opam-repository?

maiste:
>  If a repo does not follow the convention set by opam-repository, that doesn't mean it is corrupt.

It is a misunderstanding: I was referring to repositories with twice the same file or an opam file at the root.

rgrinberg:
I'll ask, but I don't really see an issue with implementing an alternative lazy loading scheme. What we're doing is implementing the following function for a single repository:

```
val list_packages : Package_name.t -> Package_version.t list
```

So you could implement this function by:

1. Try to list `packages/$package`. If this succeeds, we finish our search.
2. If the above fails because `packages/$package` doesn't exist, you read can read all of `packages` and look for `$package.*`. The result is now the matches in 3
3. For subsequent lookups, you want the listing of `packages/` to be cached. Otherwise it would be really slow for large repos that aren't from git.

With this scheme, the common users always get the optimized experience, while we still support this alternative layout.
For multiple repositories, we are sadly going to hit the 2nd step more often than we'd like. On the other hand, multiple repositories aren't all that common and the performance problem aren't relevant to opam repositories from git.

gridbugs:
> For multiple repositories, we are sadly going to hit the 2nd step more often than we'd like. On the other hand, multiple repositories aren't all that common and the performance problem aren't relevant to opam repositories from git.

What do you mean by "multiple repositories" here?

gridbugs:
I just noticed that [Jane Street's compiler extensions branch](https://github.com/janestreet/opam-repository/tree/with-extensions/packages) uses a mix of different layouts within its packages directory. Most packages are directly under packages but a few (e.g. `ocaml-variants`) get a subdirectory in the style of opam-repository. @rgrinberg is that what you mean by a "multiple repository"?

rgrinberg:
Multiple repositories can be defined in dune's workspace file in order. Refer to tests like `multiple-opam-repos.t` for examples.

Mixing the two layouts defeats the ability to implement any lazy loading scheme. I'm going to ask about sticking to one scheme. For now, I think you could just make eager loading "opt-in" for some repos.

gridbugs:
Is lazy-loading the opam repo something we're planning to implement in the future? Unless I'm mistaken it looks like we currently enumerate all the paths in the opam repo when creating a `Rev_store.At_rev.t` to populate the `files: File.Set.t` field ([here](https://github.com/ocaml/dune/blob/897bd784ba2877388001f5e88bd873449d9e83e1/src/dune_pkg/rev_store.ml#L528C7-L528C25)). I timed `dune pkg lock` a few times with and without this change and didn't see a noticeable difference.

rgrinberg:
Lazy loading should already work for local repositories (non-git). That's a fairly important use case when working on a change to opam-repository and testing it locally for example.

It would be great if we could make it work for git repositories as well, but it would be much work for that.

PR #10937, opened 2024-09-19
Configurator: pkg-config plugin uses pkgconf and --personality=TARGET by default

pirbo:
This change is required to support windows mingw opam.2.2 default setting. It can be seen as a "clean" attempt to do https://github.com/dra27/dune/commit/fd003a00e5acb213236ba04d8d3000beda552ca8 :-)

Maybe review commit by commit as commit messages tries to provide explanations.

Obviously the last one (refreshing test outputs) is not satisfactory as it but as I don't know what is the best fix, I open the PR now as an ask for feedback...

PR #11619, opened 2025-04-08
fix: pass pkg-config (extra) args when evaluating the expression

MisterDA:
pkgconf uses the personality to locate the correct library when the library can be found with multiple toolchains. For instance, on Cygwin, pkgconf could detect libcurl (built with cygwin1.dll), licurl (built with x86_64-w64-mingw32), or libcurl (built with i686-w64-mingw32). It needs the right `--personality` flag to select between these libraries, which we pass when asking for the `--cflags` or `--libs`. We also need it when validating the expression, or pkgconf returns an error:

```console
$ (cd _build/default && config/discover.exe)
which: pkgconf
-> found: C:\Users\Antonin\AppData\Local\opam\default-2\bin\pkgconf.exe
run: C:\Users\Antonin\AppData\Local\opam\default-2\bin\pkgconf.exe --print-errors "libcurl >= 7.28.0"
-> process exited with code 1
-> stdout:
-> stderr:
| Package libcurl was not found in the pkg-config search path.
| Perhaps you should add the directory containing `libcurl.pc'
| to the PKG_CONFIG_PATH environment variable
| Package 'libcurl' not found
Error: Package libcurl was not found in the pkg-config search path.
Perhaps you should add the directory containing `libcurl.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libcurl' not found
$ ~/AppData/Local/opam/default-2/bin/pkgconf.exe --personality x86_64-w64-mingw32 --print-errors "libcurl >= 7.28.0
```

This is a nasty bug for me, as there's no way of passing the personality flag to pkgconf outsite of Dune: the `PKG_CONFIG_ARGN` is just exposed by Dune and CMake and isn't actually recognized by pkgconf.

cf #10937

### DIFF
--- a/otherlibs/configurator/test/blackbox-tests/pkg-config-args.t/run.t
+++ b/otherlibs/configurator/test/blackbox-tests/pkg-config-args.t/run.t
@@ -1,9 +1,11 @@
 These tests show that setting `PKG_CONFIG_ARGN` passes extra args to `pkg-config`
 
   $ dune build 2>&1 | awk '/run:.*bin\/pkgconf/{a=1}/stderr/{a=0}a' | sed s/$(ocamlc -config | sed -n "/^target:/ {s/target: //; p; }")/\$TARGET/g
-  run: $TESTCASE_ROOT/_build/default/.bin/pkgconf --print-errors dummy-pkg
+  run: $TESTCASE_ROOT/_build/default/.bin/pkgconf --personality $TARGET --print-errors dummy-pkg
   -> process exited with code 0
   -> stdout:
+   | --personality
+   | $TARGET
    | dummy-pkg
   run: $TESTCASE_ROOT/_build/default/.bin/pkgconf --personality $TARGET --cflags dummy-pkg
   -> process exited with code 0
@@ -22,9 +24,10 @@ These tests show that setting `PKG_CONFIG_ARGN` passes extra args to `pkg-config
 
   $ dune clean
   $ PKG_CONFIG_ARGN="--static" dune build 2>&1 | awk '/run:.*bin\/pkgconf/{a=1}/stderr/{a=0}a'
-  run: $TESTCASE_ROOT/_build/default/.bin/pkgconf --print-errors dummy-pkg
+  run: $TESTCASE_ROOT/_build/default/.bin/pkgconf --static --print-errors dummy-pkg
   -> process exited with code 0
   -> stdout:
+   | --static
    | dummy-pkg
   run: $TESTCASE_ROOT/_build/default/.bin/pkgconf --static --cflags dummy-pkg
   -> process exited with code 0
--- a/doc/changes/11619.md
+++ b/doc/changes/11619.md
@@ -1,0 +1,1 @@
+- fix: pass pkg-config (extra) args in all pkgconfig invocations. A missing --personality flag would result in pkgconf not finding libraries in some contexts. (#11619, @MisterDA)
--- a/otherlibs/configurator/src/v1.ml
+++ b/otherlibs/configurator/src/v1.ml
@@ -735,7 +735,7 @@ module Pkg_config = struct
     in
     let pc_flags = "--print-errors" in
     let { Process.exit_code; stderr; _ } =
-      Process.run_process c ~dir ?env t.pkg_config [ pc_flags; expr ]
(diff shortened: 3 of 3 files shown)


maiste (approved):

maiste:
(Once the CI passes the test :smile:)

maiste (approved):

otherlibs/configurator/test/blackbox-tests/pkg-config-args.t/run.t
```diff
@@ -1,9 +1,11 @@
 These tests show that setting `PKG_CONFIG_ARGN` passes extra args to `pkg-config`
 
   $ dune build 2>&1 | awk '/run:.*bin\/pkgconf/{a=1}/stderr/{a=0}a' | sed s/$(ocamlc -config | sed -n "/^target:/ {s/target: //; p; }")/\$TARGET/g
-  run: $TESTCASE_ROOT/_build/default/.bin/pkgconf --print-errors dummy-pkg
+  run: $TESTCASE_ROOT/_build/default/.bin/pkgconf --personality $TARGET --print-errors dummy-pkg
   -> process exited with code 0
   -> stdout:
+   | --personality
+   | $TARGET
```
Leonidas-from-XIV, line 8: Just so I understand the issue: we always had the `--personality $TARGET` flag in `t.pkg_config_args` (which I guess is partly populated from `PKG_CONFIG_ARGN`), we just forgot to pass it to `pkgconf`?

Leonidas-from-XIV (commented):
It's most likely ok, I'd just want to understand where the `--personality` change comes from.

otherlibs/configurator/test/blackbox-tests/pkg-config-args.t/run.t
MisterDA: Say we're looking for libcurl. Then, `C.Pkg_config.query_expr_err pc ~package:"libcurl" ~expr:"libcurl >= 7.28.0"` will do three invocations:

```sh
pkgconf "libcurl >= 7.28.0"
pkgconf $PKG_CONFIG_ARGN --libs libcurl
pkgconf $PKG_CONFIG_ARGN --cflags libcurl
``

I just found out that if we're not passing `$PKG_CONFIG_ARGN` to the first invocation, then it might fail because `pkgconf` would be looking for libcurl in the "default" personality, which may not be the one the current ocaml install is using. So, yes, we forgot to pass it to the first invocation.

Note that `PKG_CONFIG_ARG` isn't a variable that's understood by pkgconf or pkg-config, it's an escape hatch from the build systems (CMake and Dune) to allow users to pass extra variables to pkgconf/pkg-config invocations—provided the build system doesn't forget to pass the variable!

Leonidas-from-XIV: I get that, but from what I understand there were two bugs being fixed in this PR:

1. `PKG_CONFIG_ARGN` wasn't respected when doing the call (thus `-static` wasn't passed in the test) and now it is. That much makes sense to me.
2. Somehow `--personality` was not passed either and now it is. It did not come from `PKG_CONFIG_ARGN` (since the first `dune build` invocation in the test does not have it set) so I assume it was another bug that also existed and is being fixed by this PR.

Am I right in this interpretation?

Please add a changelog entry however since it would be nice for people to learn about this fix when we cut the next Dune release.

MisterDA: I think the bug was rather that Dune forgot to pass the personality flag to the first invocation of pkgconf.

Reading about `--static`:

>      `--static`
>             Compute a deeper dependency graph and use compiler/linker flags
>             intended for static linking.

I don't think it matters much in the first invocation (just to detect whether the package is available).

The mechanism Dune uses to pass extra flags to pkgconf, whether auto-detected, or from the user through `PKG_CONFIG_ARGN`, works when requesting `--libs` and `--cflags`.

Dune might even be overzealous in doing that first invocation. If we don't get rid of it, maybe just passing the personality flag is enough and we don't need to pass the full `PKG_CONFIG_ARGN`. I'm not sure if that would be surprising to users.

Leonidas-from-XIV (approved):

PR #11621, opened 2025-04-09
reimplementation of dune exec -w

### LINKED ISSUE #11089: Child process of `dune exec -w` is not terminated on signal
## Expected Behavior
When running dune exec in watch mode (via `dune exec -w <target>`) and the target is an infinite process (e.g. a web server), a shutdown via signal (e.g. SIGINT) to the dune process should also forward the signal to the child process (`<target>`).

To be fair, this may be working as designed but I still wanted to raise this issue in case it's not.

## Actual Behavior
Currently the signal terminates dune itself, leaving the target process orphaned.

## Background
I've come across this issue as part of a dune integration into neovim. The editor plugin is doing some housekeeping and tracks all dune processes spawned during editing. Once they are not needed anymore the plugin stops those instances by sending SIGINT to the dune process leaving any running target process orphaned (as described above).

It would be great if dune could handle the cleanup of spawned child processes, otherwise I will need to fine some workaround with process groups, which are not very well supported in neovim 🫤 . 

## Reproduction
1. Create a dune project with a target that will not shut down by itself. E.g. a simple Dream service with `bin/main.ml`
```ocaml
let () =
    Dream.run ~port:8080
    @@ Dream.router [ Dream.get "/" (fun _ -> Dream.html "Hello World")]
```
2. run the server via `dune exec -w bin/main.exe`
3. send SIGINT via `kill -2 <dune_pid>` to stop the dune process

## Specifications

- Operating system (distribution and version): MacOS 14.6.1 

## Additional information 

`dune` with the `--verbose` flag:
```
Shared cache: disabled
Shared cache location: /Users/moe/.cache/dune/db
Workspace root: /Users/moe/Projects/ocaml/myproject
Auto-detected concurrency: 8
Dune context:
 { name = "default"
 ; kind = "default"
 ; profile = Dev
 ; merlin = true
 ; fdo_target_exe = None
 ; instrument_with = []
 }
Success, waiting for filesystem changes...03.11.24 21:22:46.333                       Running at http://localhost:8080
03.11.24 21:22:46.333                       Type Ctrl+C to stop
Got signal INT, exiting.
```

Alizter:
This PR is left as a draft for the moment as I iterate over it.

In this PR we reimplement `dune exec -w` to use `Dune_engine.Process` rather than `Spawn` directly. This has a number of advantages, mainly we no longer have to juggle the difficult job of keeping track of sub-processes.

Doing this has some unintended side-effects, such as all output being redirected to stderr due to how dune display messages. We therefore offer a small tweak to `Dune_engine.Process` in order to have access to the usual stdout and stderr.

- Fixes https://github.com/ocaml/dune/issues/11089

### DIFF
--- a/src/dune_engine/process.ml
+++ b/src/dune_engine/process.ml
@@ -49,6 +49,7 @@ module Io = struct
         { output_on_success : Action_output_on_success.t
         ; output_limit : Action_output_limit.t
         }
+    | External
 
   type status =
     | Keep_open
@@ -214,6 +215,7 @@ let io_to_redirection_path (kind : Io.kind) =
   | Terminal _ -> None
   | Null -> Some (Path.to_string Dev_null.path)
   | File fn -> Some (Path.to_string fn)
+  | External -> None
 ;;
 
 let command_line_enclosers
@@ -230,7 +232,7 @@ let command_line_enclosers
   in
   let suffix =
     match stdin_from.kind with
-    | Null | Terminal _ -> suffix
+    | Null | Terminal _ | External -> suffix
     | File fn -> suffix ^ " < " ^ quote fn
   in
   let suffix =
@@ -879,6 +881,7 @@ let spawn
   and stderr_limit = Io.output_limit stderr in
   let (stdout_capture, stdout), (stderr_capture, stderr) =
     match stdout.kind, stderr.kind with
+    | External, _ | _, External -> (None, stdout), (None, stderr)
     | (Terminal _, _ | _, Terminal _) when !Clflags.capture_outputs ->
       let capture ~suffix =
         let fn = Temp.create File ~prefix:"dune" ~suffix in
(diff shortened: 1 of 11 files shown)


src/dune_engine/process.ml
```diff
@@ -1225,3 +1228,22 @@ let run_capture_line
                     Pp.seq (Pp.verbatim "> ") (Pp.verbatim line)))
              ]))
 ;;
+
+let run_external_in_out =
```
Alizter, line 1232: Needs a better name

gridbugs (approved):
Looks much cleaner than the original impl. Thanks for the fix!

Alizter:
One issue with this approach is that we are not sending the program output over RPC. I don't know if there is a good way to stream program output. This also means things like `dune monitor` and `--display tui` won't work correctly with `dune exec -w`.

Another issue is that there is now a difference in semantics with `dune exec -w` and `dune exec`. We could consider making them share code, but that might be a bit too invasive on other uses of dune exec.

PR #11622, opened 2025-04-11
Allow rpc-triggered builds in eager watch mode

gridbugs:
When dune is started in eager watch mode (e.g. by running `dune build --watch`) it starts an RPC server, but prior to this change it would reject RPC requests to build targets. This is a step towards being able to run dune commands that trigger build (e.g. `build`, `exec`, `test`) while watch mode is running.

Kind of testing the waters with this change as there may be consequences I'm unaware of of running the entire build system concurrently in two different fibers. Conceptually I think it makes sense, as the task of rebuilding targets named on the command line in response to file changes is unrelated to the task of rebuilding targets in response to RPC messages. Do we need to take extra care to prevent the two fibers from interfering with each other?

### DIFF
--- a/src/dune_rpc_impl/server.ml
+++ b/src/dune_rpc_impl/server.ml
@@ -192,7 +192,6 @@ type 'a t =
   { config : Run.t
   ; pending_build_jobs : ('a list * Build_outcome.t Fiber.Ivar.t) Job_queue.t
   ; parse_build : string -> 'a
-  ; watch_mode_config : Watch_mode_config.t
   ; mutable clients : Clients.t
   }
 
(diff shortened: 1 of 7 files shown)


src/dune_rpc_impl/server.ml
```diff
@@ -321,23 +320,10 @@ let handler (t : _ t Fdecl.t) handle : 'a Dune_rpc_server.Handler.t =
...
-        raise (Dune_rpc.Response.Error.E error)
-      | Yes Passive ->
-        let ivar = Fiber.Ivar.create () in
-        let targets = List.map targets ~f:server.parse_build in
-        let* () = Job_queue.write server.pending_build_jobs (targets, ivar) in
-        Fiber.Ivar.read ivar
+      let ivar = Fiber.Ivar.create () in
+      let targets = List.map targets ~f:server.parse_build in
+      let* () = Job_queue.write server.pending_build_jobs (targets, ivar) in
+      Fiber.Ivar.read ivar
```
maiste, line 326: To make sure I understand: you delete this part because either in `passive` or `eager` mode, we want to handle the request. Am I correct?

gridbugs: That's correct. The only purpose of the `watch_mode_config` field was to prevent build requests from being handled by eager watch mode, so if we're allowing eager watch mode to handle build requests now then there's no need for the field or for this match statement (it now just always does what was formerly the `Yes Passive` case).

Alizter:
You'll need to update the documentation `rpc.rst` since that mentioned you can only connect to passive watch mode. This PR allows for both to happen.

I tested `dune build -w` together with `dune rpc ...` commands and everything appears to be working fine. I also made sure that restricting to `-j1` doesn't cause a lock with the two concurrent fibers happening. I didn't find any issues, so I think that part is fine.

gridbugs:
Thanks for testing this out @Alizter. I've update the rpc docs.

rgrinberg (approved):

PR #11624, opened 2025-04-13
chore: hardcode unix and threads library in bootstrap

Alizter:
It was discussed that we should be doing something like this in https://github.com/ocaml/dune/pull/11199.

### DIFF
--- a/src/dune_rules/bootstrap_info.ml
+++ b/src/dune_rules/bootstrap_info.ml
@@ -13,7 +13,7 @@ let flags flags =
 let rule sctx ~requires_link =
   let open Action_builder.O in
   let* () = Action_builder.return () in
-  let* locals, externals =
+  let* locals, _externals =
     Memo.Lazy.force requires_link
     |> Resolve.Memo.read
     >>| List.partition_map ~f:(fun lib ->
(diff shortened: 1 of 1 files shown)


src/dune_rules/bootstrap_info.ml
```diff
@@ -64,11 +64,8 @@ let rule sctx ~requires_link =
         ])
     |> Action_builder.of_memo
   in
-  let externals =
```
maiste, line 67: I understand that Dune generates the `thread.posix` every time, but I don't understand where it comes from. What is a change in the compiler?

Alizter: I don't understand why this is reported like this. Maybe @nojb knows the reason?

Alizter: I believe we are using findlib (or dune's version) for finding these external libraries. At some point ~ OCaml 5.2 we started getting `threads.posix` rather than both `unix` and `threads`.

nojb: It is a bug in 153fcd1bc5c9f850a1f6b50e213a48d548c377ac, which inverted the condition. `threads.posix` is supposed to be filtered _out_.

Alizter: @nojb Good catch! I will open another PR with a fix.

Alizter:
Fix is here: https://github.com/ocaml/dune/pull/11636

PR #11632, opened 2025-04-17
chore: comply with LGPL-2

### LINKED ISSUE #11603: Dune license
The toplevel LICENSE.md file says that dune is distributed under the MIT license.  But I see two copies of the LGPL 2.1 license, in `src/dune_pkg/COPYING` and `src/sat/COPYING`.  What is the significance of those license files?  They are not referred to anywhere.  None of the source files in `src/dune_pkg` or `src/sat` say under which license they are distributed.  Should I deduce that all of the source files in `src/dune_pkg` and `src/sat` are distributed under the LGPL?  If so, is one of these meant?
- LGPL-2.1-only
- LGPL-2.1-or-later
- LGPL-2.1-only WITH OCaml-LGPL-linking-exception
- LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception

In any of these cases, due to the copyleft nature of the LGPL, doesn't that mean that the dune project as a whole is distributed under the LGPL, not the MIT license?

maiste:
This PR adds headers to the files from the 0install solver under LGPL2. I have added a pointer to the full notice and list the files that were diluted in `opam_solver.ml` from the `0install`.

I didn't add the changes at the top of the file as any user is able to track the changes through `git`. I don't know if it would be sufficient to comply.

Should fix #11603.

### DIFF
--- a/src/dune_pkg/opam_solver.ml
+++ b/src/dune_pkg/opam_solver.ml
@@ -1,3 +1,17 @@
+(*
+   This file includes code from the 0install library. It is distributed under
+   the LGPL-2.1-or-later licence. See src/sdune_pkg/COPYING.md for the full
+   license.
+
+   Copyright (C) 2013, Thomas Leonard
+   See the README file for details, or visit http://0install.net.
+
+   Files:
+   - Includes solver_core.ml and solver_core.mli files
+   - Includes diagnostics.ml and diagnostics.mli
+   - Remove usage from the S module from the s.ml file
+*)
+
 open Import
 open Fiber.O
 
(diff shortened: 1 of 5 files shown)


maiste:
@rgrinberg @Alizter should we merge this or should we close it? It is supposed to address the question about the LGPL-2 but not being a lawyer, I can't tell how much we are complying with it. From what I read it seems to be what is necessary but I can't confirm deeper. I don't have a particular opinion on this.
WDYT?

Alizter:
Worst-case we are still not compliant with LGPL-2 which is not any better than the current situation. Best-case we have fixed the issue. Since we have no feedback from the original issue author, let's go ahead and merge this. If we are still in the wrong, then somebody will point it out. I don't think its worth spending too much time on this.

PR #11636, opened 2025-04-17
fix: bootstrap

Alizter:
In 153fcd1 the filter on external libraries was flipped. This fixes that issue.

This fixes issues with bootstrap on newer OCaml versions.

Thanks @nojb!


nojb (approved):
LGTM

PR #11641, opened 2025-04-22
simplify implementation of dune exec further

Alizter:
We simplify the implementation of `dune exec` further and now `dune exec` and `dune exec -w` share the same implementation.

One difference with before (which can be observed in the tests) is that `arg[0]` is now set to an absolute path. This is still valid within the `arg[0]` convention, but we could think about changing it back to a relative path.

- [ ] Might be a good time to improve some of the doc here?

### DIFF
--- a/bin/exec.ml
+++ b/bin/exec.ml
@@ -181,20 +181,10 @@ let run_once config ~prog ~args ~common ~no_rebuild ~context =
   @@ fun () ->
   let open Fiber.O in
   let* setup = Import.Main.setup () in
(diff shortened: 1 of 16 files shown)


Alizter:
The test which signals a segfault seems to crash out in the ubuntu CI but seems fine on macos and my machine. Not sure what is happening there.

rgrinberg:
@gridbugs can you review this one?

gridbugs (approved):

PR #11647, opened 2025-04-23
chore: rename internal spawn to dune_spawn

rgrinberg:
We shouldn't be using the internal vendored library for anything but bootstrapping dune. For everything else, spawn from opam should be preferred.


Alizter:
This should fix an issue in `bench/micro/dune`. You could re-enable the test there now.

PR #11654, opened 2025-04-25
refactor(install_rules): remove dead code

anmonteiro:
Reading the code in `Dune_rules.Modules.generated`, `.ml-gen` are 1:1 with `Option.is_some Module.install_as = true`, so this code should be dead. Tests to confirm


anmonteiro:
guess i'm wrong.

PR #11683, opened 2025-04-27
feat: add extra_objects field to foreign_library stanza

Alizter:
We add support for the `extra_objects` field in the `foreign_library` stanza. This allows extra object files to be linked to foreign libraries.

- [x] Update docs
- [x] Changelog
- [x] Dune lang bump (we should be 3.19 not 3.18)?

### DIFF
--- a/doc/changes/11683.md
+++ b/doc/changes/11683.md
@@ -1,0 +1,2 @@
+- Added `(extra_objects)` field to `(foreign_library)` stanza with `(:include)` support.
+  (#11683, @Alizter)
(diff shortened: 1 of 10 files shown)


rgrinberg:
Could you see if it's hard to add `(:include ...)` as well? It would allow us to get rid of the ugly `dynamic_include` stuff.

Alizter:
@rgrinberg I added `(:include)` for `extra_objects` in the `foreign_library` stanza. I suppose we should also do the same for `executable`.

PR #11689, opened 2025-04-27
feature(bootstrap): more c stubs support

rgrinberg:
* Clags for certain stub files
* Conditionally enable stub files based on the platform

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -53,6 +53,8 @@ module String = struct
   include String
   module Set = Set.Make (String)
   module Map = Map.Make (String)
+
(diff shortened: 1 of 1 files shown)


boot/duneboot.ml
```diff
@@ -763,8 +786,35 @@ module Library = struct
         | Not_found -> String.length fn
       in
       match String.sub fn ~pos:i ~len:(String.length fn - i) with
+      | (".S" | ".asm") as ext ->
```
Alizter, line 789:
```suggestion
      | (".S" | ".s" | ".asm") as ext ->
```

PR #11692, opened 2025-04-28
chore: re-enable micro benchmark

rgrinberg:
there's no longer a conflict since we aren't using the vendored spawn to build it.


rgrinberg:
Looks like we need to rename `uutf` as well. All of this manual renaming is getting tiresome. I'm going to look into the auto-open PR again.

PR #11696, opened 2025-04-29
chore(nix): improve nix devShell#scope

anmonteiro:
- use [`lib.fileset`](https://nixos.org/manual/nixpkgs/stable/#sec-functions-library-fileset) to improve the `#scope` devShell (only recompiles dune if its sources change, instead of the entire repo)
- add `pkgs` argument to `extraBuildInputs` to prevent conflicting OCaml packages errors

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -80,7 +80,12 @@
       testBuildInputs = with pkgs;
         [ file mercurial unzip ]
         ++ lib.optionals stdenv.isLinux [ strace ];
-      testNativeBuildInputs = with pkgs; [ nodejs-slim pkg-config opam ocamlformat ];
+      testNativeBuildInputs = pkgs: with pkgs; [
+        nodejs-slim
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -97,7 +102,24 @@
...
+                ./bin
+                ./boot
+                ./configure
+                ./dune-project
+                ./dune-file
+                ./src
+                ./plugin
+                ./vendor
+                ./otherlibs
+                ./Makefile
```
rgrinberg, line 119: is this the set of files where changes can trigger recompilation? If so, I think the makefile should be absent from this list. We should not use it as part of the build process. It is only used for development.

anmonteiro: Ithe Makefile is used for the nix build since the derivation doesn’t have a buildPhase, so it needs to be present

rgrinberg (approved):

PR #11699, opened 2025-04-29
chore: remove coverage information

Alizter:
Coverage stopped being tracked back in #11224 this cleans up the rest of the mentions.


Leonidas-from-XIV (approved):
Seems reasonable. The only coverage info on coveralls seems to be old branches from before the coverage support was removed.

maiste (approved):
Thanks for having taking care of this!

PR #11609, opened 2025-04-05
feat: Synopsis to rule stanza, display synopses in dune show

maxim092001:
Implementation to resolve https://github.com/ocaml/dune/issues/11522

This PR makes multiple changes:

- Allow to attach optional `(synopsis "Doc")` to `rule` stanza
- Print **all** synopses with their locations (if `loc` is not `none`) via `dune show aliases`
  - If one alias attached to multiple rules with synopses then all synopses printed with rule locations
- Print synopsis for each target if there is one via `dune show targets`
- ~~Attach synopsis to `default` alias~~
  - ~~Had to adjust some tests for new behavior~~
  - Will be done in separate PRs
- ~~Bumped version to 3.19~~ Merged https://github.com/ocaml/dune/commit/d2745311bbcc76d880c01f7030f97a13b0e699c2

Some examples from cram tests are:

```dune
(rule
(targets file1.ml)
(aliases rule-with-synopses and-another-rule)
(synopsis "Rule creates file1.ml")
(action
    (write-file file1.ml "")))

(rule
(targets file2.ml)
(aliases rule-with-synopses and-another-rule)
(synopsis "Rule creates file2.ml")
(action
    (write-file file2.ml "")))

(rule
(targets file3.ml)
(alias and-another-rule)
(synopsis "Rule creates file3.ml")
(action
    (write-file file3.ml "")))
```

```
  $ dune show targets
  dune
  dune-project
  file1.ml
    - dune:1 Rule creates file1.ml
  file2.ml
    - dune:8 Rule creates file2.ml
  file3.ml
    - dune:15 Rule creates file3.ml
```

```
dune show aliases
  all
  and-another-rule
    - dune:1 Rule creates file1.ml
    - dune:8 Rule creates file2.ml
    - dune:15 Rule creates file3.ml
  default
    - This alias corresponds to the default argument for dune build. More
      information on
      https://dune.readthedocs.io/en/latest/reference/aliases/default.html
  fmt
  ocaml-index
  pkg-install
  rule-with-synopses
    - dune:1 Rule creates file1.ml
    - dune:8 Rule creates file2.ml
```

PR #11705, opened 2025-04-29
Add location to aliases unless they are standard

Leonidas-from-XIV:
The change to add locations for aliases #11609 got me thinking, that it's useful in general to pass it. However given the rule that is attached to a "standard" alias does not come from a particular location that the user can go to, it should probably explicitly state `Loc.none` on these aliases.


rgrinberg:
> However given the rule that is attached to a "standard" alias does not come from a particular location that the user can go to, it should probably explicitly state Loc.none on these aliases.

I don't really understand this bit. Sure, standard aliases aren't defined by any particular stanza, but don't we want to record the stanza that introduced the target to the standard alias?

Leonidas-from-XIV:
> don't we want to record the stanza that introduced the target to the standard alias?

We can, however the locations of these get printed as `dune:1` (as in [here](https://github.com/ocaml/dune/pull/11609/files#diff-b5a19d9cb29aa6be6ad02a91659ed2a230b25a93f9502ebaefcd22bcb52dd47f) so I assume that the rule doesn't really have a "real" location and the user won't find anything interesting at that location in the file.

Leonidas-from-XIV:
I can't tell you why #11609 uses `Loc.none` (haven't had the time to look at the changes in depth), but if I change the code in this PR to always unconditionally use `rule.loc` I also get `dune:1` in the same test. So it's not something that #11609 broke, it's already like this on `main`.

rgrinberg (approved):
This is a strict improvement, but we need to get to the bottom for the cause of the dummy locations someday.

PR #11708, opened 2025-04-29
chore(ci): switch to 5.2

rgrinberg:


maiste (approved):
Thanks! Do you think it would be easy to bump the dev switch version to `5.2`, or will it require more work?

Issue #10630, opened 2024-06-09
RFC: single-context Universal Libraries


anmonteiro:
### Context

- `(using melange 0.1)` allows writing Melange / OCaml code together with limitations
- Recent work on "universal libraries" proposes using different contexts and `enabled_if`

### Problems

- Multi-context libraries introduce a few problems:
    - `library` stanzas end up duplicated for any non-trivial use cases, and `(enabled_if (= %{context} melange))` is infectious (melange libraries only work within the `melange` context)
    - it's not trivial to install melange libraries in multiple contexts.
    - Editor support becomes way more painful, and switching between files in different contexts still requires restarting the LSP.

### Requirements:

- Installing multi-mode libraries alongside each other
- vary module sources according to the library `mode`
- vary preprocessing according to the library `mode`
- vary library dependencies according to the library `mode`

### Proposed solution

- The proposed solution is 2-fold:
    - new conditional fields `libraries`, `preprocess`, `flags`, `modules` in `(library ..)`
        - prior art: `melange.compile_flags`
        - in their absence, a library with `(modes melange)` falls back to the fields already supported before this change.
    - the ability to choose module sources based on the mode e.g. foo.%{mode}.ml
        - we can solve the "`copy_files` problem" by having shared module sources and mode-specific sources in the same directory, with different extensions. Example:

```
- src
  |_ dune
  |_ common.ml
  |_ foo.melange.ml
  |_ foo.ocaml.ml
```

Slightly unrelated: this proposal is likely backwards-compatible with `(using melange 0.1)`; after adopting it successfully in universal library projects such as Ahrefs, I propose we mark the Melange extensions stable in dune (`1.0`).

### Implementation details

- we can keep module sources in the same directories in the build folder
- for Melange, we'll have to copy them to e.g. a `.melange-sources` folder after setting up rules for the implementation-specific module sources
- Installation keeps working since we already copy melange artifacts to `<lib-root>/<pkg-name>/melange`. We can install the Melange sources there, too.

### Caveats / further work

It's likely we'll have to change Merlin to map a module name to its source file. This is desirable in general even without this proposal: Merlin currently guesses what module source corresponds to a module, and implementing this mapping would be a benefit for most OCaml projects and editor tooling.

andreypopp:
> new conditional fields libraries, preprocess, flags, modules in (library ..)

Am I correct that this is going to look as

    (library 
        (name ...)
        (libraries ...)
[...]

PR #11713, opened 2025-04-30
refactor: simpler tracking of `Lib_info.virtual_`

anmonteiro:
- `Lib_info.virtual_` at most tracks the modules that are already tracked in `Lib_info.modules lib.info`
- It can just be a `bool` instead of pointing to redundant information, which simplifies the implementation of #10630 a little bit

the diff is best viewed [without the whitespace diff](https://github.com/ocaml/dune/pull/11713/files?w=1)

### DIFF
--- a/src/dune_rules/virtual_rules.ml
+++ b/src/dune_rules/virtual_rules.ml
@@ -74,22 +74,20 @@ let impl sctx ~(lib : Library.t) ~scope =
          ]
      | Some vlib ->
        let info = Lib.info vlib in
-       let virtual_ =
-         match Lib_info.virtual_ info with
-         | Some v -> v
-         | None ->
-           User_error.raise
-             ~loc:lib.buildable.loc
-             [ Pp.textf
-                 "Library %s isn't virtual and cannot be implemented"
-                 (Lib_name.to_string implements)
-             ]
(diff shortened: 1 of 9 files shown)


src/dune_rules/virtual_rules.ml
```diff
@@ -74,59 +74,58 @@ let impl sctx ~(lib : Library.t) ~scope =
...
-             >>| Foreign_sources.for_lib ~name
-             >>| (let ext_obj = ocaml.lib_config.ext_obj in
-                  let dir = Obj_dir.obj_dir (Lib.Local.obj_dir vlib) in
-                  Foreign.Sources.object_files ~ext_obj ~dir)
-             >>| List.map ~f:Path.build
-           in
-           modules, foreign_objects
-       in
-       Some (Vimpl.make ~impl:lib ~vlib ~vlib_modules ~vlib_foreign_objects))
+       (match Lib_info.virtual_ info with
```
rgrinberg, line 77: You could remove this pattern match and just turn this into a pre-condition.

anmonteiro: done

rgrinberg (approved):

PR #11714, opened 2025-04-30
refactor: move `Modules{.With_vlib,}.entry_modules`

anmonteiro:
It was a `Code_error` to use `entry_modules` on a `Impl` branch, so there's no point in having that function in the `With_vlib` submodule. This simplifies some of the other code, which doesn't have to create a `Modules.With_vlib.Modules ..` just to call this function.

### DIFF
--- a/src/dune_rules/compilation_context.ml
+++ b/src/dune_rules/compilation_context.ml
@@ -330,9 +330,8 @@ let entry_module_names sctx t =
   match Lib_info.entry_modules (Lib.info t) with
   | External d -> Resolve.Memo.of_result d
   | Local ->
-    let+ modules = Dir_contents.modules_of_lib sctx t in
-    let modules = Option.value_exn modules in
-    Resolve.return (Modules.With_vlib.entry_modules modules |> List.map ~f:Module.name)
+    let+ modules = Dir_contents.modules_of_local_lib sctx (Lib.Local.of_lib_exn t) in
+    modules |> Modules.entry_modules |> List.map ~f:Module.name |> Resolve.return
 ;;
 
 let root_module_entries t =
--- a/src/dune_rules/modules.ml
+++ b/src/dune_rules/modules.ml
@@ -930,6 +930,18 @@ let compat_for_exn t m =
      | Some (Group g) -> Group.lib_interface g)
 ;;
 
+let entry_modules t =
+  List.filter
+    ~f:(fun m -> Module.visibility m = Public)
+    (match t.modules with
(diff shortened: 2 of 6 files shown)


src/dune_rules/modules.mli
```diff
@@ -56,6 +56,11 @@ val virtual_module_names : t -> Module_name.Path.Set.t
 val wrapped : t -> Wrapped.t
 val source_dirs : t -> Path.Set.t
 
+(** List of entry modules visible to users of the library. For wrapped
+      libraries, this is always one module. For unwrapped libraries, this could be
```
Alizter, line 60: Could you rewrap this comment

src/dune_rules/modules.ml
```diff
@@ -920,6 +920,18 @@ let source_dirs =
...
 
+let entry_modules t =
+  List.filter
+    ~f:(fun m -> Module.visibility m = Public)
+    (match t.modules with
+     | Stdlib w -> Stdlib.lib_interface w |> Option.to_list
+     | Singleton m -> [ m ]
+     | Unwrapped m -> Unwrapped.entry_modules m
+     | Wrapped m ->
+       (* we assume this is never called for implementations *)
```
Alizter, line 931: This comment no longer makes sense right?

src/dune_rules/compilation_context.ml
```diff
@@ -330,9 +330,8 @@ let entry_module_names sctx t =
   match Lib_info.entry_modules (Lib.info t) with
   | External d -> Resolve.Memo.of_result d
   | Local ->
-    let+ modules = Dir_contents.modules_of_lib sctx t in
-    let modules = Option.value_exn modules in
-    Resolve.return (Modules.With_vlib.entry_modules modules |> List.map ~f:Module.name)
+    let+ modules = Dir_contents.modules_of_local_lib sctx (Lib.Local.of_lib_exn t) in
+    Resolve.return (modules |> Modules.entry_modules |> List.map ~f:Module.name)
```
Alizter, line 334: I think its easier to read with `|> Resolve.return`

rgrinberg (approved):

src/dune_rules/compilation_context.ml
anmonteiro: done

src/dune_rules/modules.ml
anmonteiro: I believe this comment refers to module implementations rather than virtual lib implementations. I'll leave it for now.

PR #11715, opened 2025-04-30
refactor: move `Modules{.With_vlib,}.compat_for_exn`

anmonteiro:


Alizter:
I find this a bit strange since this is only used for virtual modules.

anmonteiro:
> I find this a bit strange since this is only used for virtual modules.

That doesn't seem true at all. This function is used in `Dune_rules.Dep_rules` for regular library modules.

rgrinberg (approved):

PR #11716, opened 2025-04-30
Add a "`doc_depends`" field to the `package` stanza

panglesd:
This is the first of a series of PRs to support odoc 3 features.

Odoc 3 has added the ability to link to other package docs when built together, eg in ocaml.org. (This may include docs in the reverse dependency cone, so it's a new kind of dependency to avoid circularity).

To do so, one need to
- Fill a `x-extra-doc-deps` field in the opam file,
- Install a `odoc-config.sexp` file.

See [this part](https://ocaml.github.io/odoc/odoc/odoc_for_authors.html#config-file) of odoc's doc for more information.

This PR adds a `doc_depends` field to the package stanza, to be able to generate those two files:
```
(package
 (name foo)
 (doc_depends
  (libraries <lib spec>)
  (packages <pkg spec>)))
```
where `packages` takes the same as the `depends` stanza (a list of packages with potential version constraints), while `libraries` takes the same as `(library (libraries ...))` stanza (a string list).

This PR also includes a test and the update of the documentation.


Leonidas-from-XIV:
Thanks for the PR. Before merging we'll need a changelog entry, so I'll start with that to not forget :-)

This is quite some amount of bikeshedding but I think the new stanza is a bit weird. It is called `doc_depends` but it doesn't have the syntax of `depends`;:`packages` has the syntax of `depends`. And `libraries` suddenly exists in `dune-project` whereas before it only ever existed in `dune` files.

To be constructive, I have a few alternate suggestions:

1. Rename `doc_depends` into `documentation` and `packages` into `depeds`. This way it makes `doc_depends` the place to configure odoc in the project and other things can be set up there as well (like the hierarchies). All `depends` fields have the same syntax so it is easier for people to understand what format goes where.
2. Move `packages` into `depends` with a special variable like `:odoc-depends`. It's a bit weird and ad-hoc. Maybe `depends` with a `(depends x y z (:doc-depends foo bar baz))`? That said, `libraries` needs to live somewhere. I wonder, maybe it should actually go into the `dune` files? That would be the cleanest, but also rather tedious to add them everywhere. Or maybe they should live on the top-level of `dune-project`?

Generally, do we expect people to use different `doc_depends` per package? Or maybe we should just go through every `depends` and detect every findlib library and automatically add it? Having to specify the kind of same things (the opam package `yojson` has a findlib package called `yojson` which supplies a module called `Yojson`) is one of the things that make setup in OCaml tedious and while we can't do much about what exists already, maybe we can avoid the same kind of mistakes in the future?

panglesd:
Thanks a lot @Leonidas-from-XIV for your answer and helpful suggestions!

I'll introduce a specific example scenario, and try to extract from your alternative suggestions the concrete files that we would need. And after that, I discuss which alternative I would prefer, and why!

So, the common scenario is a file whose generated opam file has to look like:

```opam
depends: [
  "lwt"
  "conf-jq" {with-test}
]

x-extra-doc-deps: [
  "fmt" {= version}
  "odig"
]
```

and `odoc-config.sexp` is:

```dune
(libraries lwt fmt.tty)
(packages odig sherlodoc)
```

Let's look at the alternatives of dune configurations that would generate those files.

---

Option 1: Everything in the `documentation` stanza of the `package` (with `referenced_libraries` not to confuse with `libraries`):

```dune
(package
 (name ...)
 (depends
  lwt
  (conf-jq :with-test))
 (documentation
  (depends
   (fmt (= :version))
   odig)
  (referenced_libraries
   lwt
   fmt.tty)))
```

---

Option A: `:odoc-doc` variable in `depends`, `referenced_libraries` in `package` stanza of `dune-project` file

```dune
(package
 (name ...)
 (depends
  lwt
  (conf-jq :with-test)
  (fmt (and (= :version) :odoc-doc))
  (odig :odoc-doc))
 (documentation
  (referenced_libraries
   lwt
   fmt.tty)))
```

---

Option α: `:odoc-doc` variable in `depends`, `referenced_libraries` in `documentation` stanza of `dune` file

```dune
(package
 (name ...)
 (depends
  lwt
  (conf-jq :with-test)
  (fmt (and (= :version) :odoc-doc))
  (odig :odoc-doc)))
```

```dune
(documentation
 (referenced_libraries
  lwt
  fmt.tty))
```

---

### On having `referenced_libraries` in the `documentation` stanza of `dune` files

I find it a bit missleading. For example:
```dune
(documentation
 (mld_files file1)
 (referenced_libraries lib1))

(documentation
 (mld_files file2)
 (referenced_libraries lib2))
```
seems to suggest that `file1` can reference `lib1`'s doc while `file2` can reference `lib2`'s doc but actually there is only a global, per package, `odoc-config.sexp`, so both mld files can reference both `lib1` and `lib2`'s documentation.

That plays against option α.

### On putting documentation dependencies in the same place as regular dependencies

Opam has a few [package variables](https://opam.ocaml.org/doc/Manual.html#Package-variables)/[flags](https://opam.ocaml.org/doc/Manual.html#pkgvar-post), which are [available](also used in) in the `depends` field of dune, but whose interpretation is left to opam.

If we add some new variable which has a meaning in dune but none in opam, how would we generate the opam `depends` and `x-extra-doc-deps` of eg:

```dune
(package
 (depends
  (cmdliner (or (<= 2.0) :odoc-doc))))
```

There would be several options
- Add support for `:odoc-doc` in opam first, which would replace the current `x-extra-doc-deps` field,
- Inconditionally move, from "`depends`" to "`x-extra-doc-deps`", any element in dune's `depends` that has `:odoc-doc` in the filter AST.

### Conclusion

I think we should go with option 1:
- Rename `doc_depends` by `documentation`
- Rename `packages` in old `doc_depends`/new `documentation` to `depends`
- Rename `libraries` in old `doc_depends`/new `documentation` to `referenced_libraries`

> Generally, do we expect people to use different doc_depends per package? Or maybe we should just go through every depends and detect every findlib library and automatically add it? Having to specify the kind of same things (the opam package yojson has a findlib package called yojson which supplies a module called Yojson) is one of the things that make setup in OCaml tedious and while we can't do much about what exists already, maybe we can avoid the same kind of mistakes in the future?

I agree that the current situation is confusing, with packages containing libraries containing modules, most of the time with the same name but sometimes with different ones.
In our case, it surfaces with the following:
- If we want to link to the manuals of a package, we need to add the _package_ name in the `depends` field
- If we want to reference a library of a package, we also need to specify that, by adding the _library_ name in the `referenced_libraries` field.

When the library and package names are the same that means that we seem to have a redundant information:
```
(package
 (documentation
  (depends fmt)
  (referenced_libraries fmt)))
```
However, I'm not sure how to avoid avoid that. Can we, by default, add all libraries from all packages we (doc-)depend on? Do we statically know all libraries a package installs?

I'll try to answer more convincingly whether this is forced by the past and we can't do anything about it, or if it's a mistake we should avoid :)

Leonidas-from-XIV:
> Can we, by default, add all libraries from all packages we (doc-)depend on? Do we statically know all libraries a package installs?

Statically not. Unless we use `dune pkg` to download and build the packages we don't know what `findlib` names a given `opam` package will expose, unfortunately. However, if OPAM is used instead, then we don't even know that at runtime, because they all just exist in the switch, with no record which opam package they came from.

So unfortunately, if we want to have this information available statically or avoid depending on `dune pkg`, the libraries will have to be declared by hand.

panglesd:
Thanks for the precision.

I'm currently discussing with @jonludlam to see how we could simplify some things (eg using `:post and :with-doc` instead of `x-extra-doc-deps`; and maybe also avoiding the use of `odoc-config.sexp` if possible).

So, I'll mark this PR as draft and undraft it as soon as the design has stabilized!

panglesd:
Finally, this PR can be undrafted!

Since [this PR has been merged](https://github.com/ocaml/odoc/pull/1343), there is no need to specify the libraries we want to reference in our docs, we only need to do that for packages.

With this in mind, I've simplified this to:

```
(package
 (documentation
  (depends ...)))
```

where this `depends` field allows the user to define which packages are referenced in the docs. It shares the same syntax as the `depends` field for normal dependencies.

There are a few details to mention:

- The `documentation` field already exists in `package` stanza. It is used to give the `url` to the documentation: `(package (documentation "http://url.to/doc"))`.

  For retro-compatibility, this is still supported. And, in order to specify both the URL and the package dependencies, we can do

  ```
  (package
   (documentation
    (url "http://url.to/doc")
    (depends ...)))
  ```

- The `documentation` field also exist in the toplevel of `dune-project` files. In those case, it was extended for consistency with the new syntax, except that defining dependencies outside of a package raise a warning.

- I haven't handled the dune-lang version checks...

shonfeder:
@punchagan (with assist form me) will be helping land this, now that PE is away. We are ready to respond to reviews at this point, and would appreciate getting feedback!

doc/reference/dune-project/package.rst
```diff
@@ -26,6 +26,29 @@ is used when generating OPAM files (see :doc:`generate_opam_files`).
...
+      .. describe:: (depends <dep-specification>)
+
+         Lists the packages whose docs is referenced in an ``mli`` or ``mld``
+         file. For instance, to be able to reference ``cmdliner``'s tutorial
+         with ``{!/cmdliner/tutorial}``, ``cmdliner`` needs to be added here.
+
+         The syntax for this field is :token:`~pkg-dep:dep_specification`.
+
+      This field can also be the same as the corresponding global field (see
+      :doc:`documentation`), in which case it takes precedence.
```
Leonidas-from-XIV:
```suggestion
      :doc:`documentation`), in which case this takes precedence.
```

doc/reference/dune-project/package.rst
```diff
@@ -26,6 +26,29 @@ is used when generating OPAM files (see :doc:`generate_opam_files`).
 
       Package dependencies, as :token:`~pkg-dep:dep_specification`.
 
+   .. describe:: (documentation <doc-spec>)
+
+      The documentation field allows to specify documentation options for the
+      package.
```
Leonidas-from-XIV, line 34: I think we should also document that `(documentation "string")` is the same as `(documentation (url "string"))` for consistency and backward compatibility.

doc/reference/dune-project/package.rst
```diff
@@ -26,6 +26,29 @@ is used when generating OPAM files (see :doc:`generate_opam_files`).
...
+         Where the documentation is hosted. Overrides the corresponding global
+         field (see :doc:`documentation`).
+
+      .. describe:: (depends <dep-specification>)
+
+         Lists the packages whose docs is referenced in an ``mli`` or ``mld``
+         file. For instance, to be able to reference ``cmdliner``'s tutorial
+         with ``{!/cmdliner/tutorial}``, ``cmdliner`` needs to be added here.
+
+         The syntax for this field is :token:`~pkg-dep:dep_specification`.
```
Leonidas-from-XIV, line 54: Both fields probably also need a note in which dune-lang version they were added.

doc/reference/dune-project/package.rst
```diff
@@ -26,6 +26,29 @@ is used when generating OPAM files (see :doc:`generate_opam_files`).
...
+
+      This field can also be the same as the corresponding global field (see
+      :doc:`documentation`), in which case it takes precedence.
+
```
Leonidas-from-XIV, line 58: Also it needs to retain the `versionadded:: 2.0` from the original documentation.

src/dune_lang/documentation.ml
```diff
@@ -0,0 +1,57 @@
+open! Stdune
```
Leonidas-from-XIV, line 1: Maybe it should just be a regular `open` to warn if it is unused...

src/dune_lang/documentation.ml
```diff
@@ -0,0 +1,57 @@
...
+         ]
+;;
+
+let decode ~toplevel =
+  let open Dune_sexp.Decoder in
+  let+ res =
+    either string
+    @@ fields
+         (let+ loc, packages =
+            located @@ field ~default:[] "depends" (repeat Package_dependency.decode)
```
Leonidas-from-XIV, line 26: As @panglesd this needs to be gated on a dune-lang version.

src/dune_lang/documentation.ml
```diff
@@ -0,0 +1,57 @@
...
+  { packages : Package_dependency.t list
+  ; url : string option
+  }
+
+let encode { packages; url } =
+  let open Dune_sexp.Encoder in
+  match packages, url with
+  | [], Some url -> string url
+  | _ ->
+    list sexp
```
Leonidas-from-XIV, line 13: The way this gets encoded probably needs to check the version of dune-lang. I guess it would be a `Code_error` if `packages` are non-empty but the dune-lang version is below the gated version.

src/dune_rules/install_rules.ml
```diff
@@ -42,6 +48,17 @@ module Package_paths = struct
     Path.Build.relative (build_dir ctx pkg) (Package.Name.to_string name ^ ".dune-package")
   ;;
 
+  let odoc_config_file ctx pkg =
+    if need_odoc_config pkg
+    then (
+      let name = Package.name pkg in
```
Leonidas-from-XIV, line 54: Might as well call `to_string` here.

src/dune_rules/install_rules.ml
```diff
@@ -42,6 +48,17 @@ module Package_paths = struct
...
+      Some
+        (Path.Build.relative
+           (build_dir ctx pkg)
+           (Package.Name.to_string name ^ ".odoc-config.sexp")))
```
Leonidas-from-XIV, line 58: I feel `sprintf "%s.odoc-config.sexp" name` might be a bit nicer here, but it's very much nitpicking.

src/dune_rules/install_rules.ml
```diff
@@ -815,6 +838,28 @@ end = struct
...
+    in
+    let ctx = Super_context.context sctx |> Context.build_context in
+    match Package_paths.odoc_config_file ctx pkg with
+    | None -> Memo.return ()
+    | Some odoc_config_file ->
+      let action =
+        Action_builder.write_file_dyn
+          odoc_config_file
+          (Action_builder.return
+           @@ Format.asprintf
```
Leonidas-from-XIV, line 853: `sprintf` should be enough here.

src/dune_rules/install_rules.ml
```diff
@@ -815,6 +838,28 @@ end = struct
...
+                "(packages %s)"
+                (packages
```
Leonidas-from-XIV, line 855: For readability sake I would move this into its own binding.

src/dune_rules/opam_create.ml
```diff
@@ -108,6 +108,38 @@ let default_build_command =
            ~with_sites:Dune_project.(is_extension_set project dune_site_extension))
 ;;
 
+let depends_field t =
+  let depends = Package.depends t in
+  let pkg_already_in_depends (pkg : Package_dependency.t) =
+    List.exists
```
Leonidas-from-XIV, line 114: Potentially worth it to create a `Package_dependency.Set.t` out of `depends` to avoid an `O(n)` lookup that's going to be called `n` times.

But given the list of depends isn't that big usually... up to you.

src/dune_rules/opam_create.ml
```diff
@@ -108,6 +108,38 @@ let default_build_command =
...
+      ~f:(fun (doc_dep : Package_dependency.t) ->
+        if pkg_already_in_depends doc_dep
+        then None
+        else (
+          let constraint_ =
+            let c = Option.to_list doc_dep.constraint_ in
+            let c =
+              Package_constraint.And
+                (c
+                 @ [ Package_constraint.Bvar Package_variable_name.post
```
Leonidas-from-XIV, line 131: I am slightly worried about adding `post` depends (mostly because I remember them being a headache, not because I have a specific issue in mind). Are we sure that declaring the dependencies as "post" is the correct semantic here?

src/dune_rules/opam_create.ml
```diff
@@ -122,7 +154,12 @@ let package_fields package ~project =
       | Some v -> Some (k, string v))
   in
   let dep_fields =
-    [ "depends", Package.depends package
+    [ "depends", depends_field package
+    ; ( "x-extra-doc-deps"
```
Leonidas-from-XIV, line 158: Are we adding the packages to both (conditionally) to `depends` and (unconditionally) to `x-extra-doc-deps`?

test/blackbox-tests/test-cases/doc-dependencies.t/run.t
```diff
@@ -0,0 +1,53 @@
...
+
+  $ dune build @install
+
+  $ cat _build/install/default/doc/foo/odoc-config.sexp
+  (packages ppx_derivers brr)
+
+Now let's check that specifying doc dependencies outside of a package stanza
+generates a warning:
+
+  $ mv dune-project-2 dune-project
```
Leonidas-from-XIV, line 45: I think this test would be more readable as `doc-dependencies.t` and write the `dune-project` files via `cat` from the test instead of using a directory test and separate files that have to be moved around.

test/blackbox-tests/test-cases/doc-dependencies.t/run.t
```diff
@@ -0,0 +1,53 @@
...
+      brr)
+    (url "url")))
+
+Doc dependencies are explained here: https://ocaml.github.io/odoc/odoc/odoc_for_authors.html#config-file
+
+They need to be in the x-extra-doc-deps field of the opam file (to be able to
+generate the opam universe) and installed in the odoc-config.sexp file (so that
+the driver knows what to make available to reference). Let's check that.
+
+Generate it's opam file
```
Leonidas-from-XIV, line 16:
```suggestion
Generate its opam file
```

test/blackbox-tests/test-cases/doc-dependencies.t/run.t
```diff
@@ -0,0 +1,53 @@
...
+  $ dune build foo.opam
+
+and check that the x-extra-doc-deps has been generated successfully
+  $ grep extra-doc-deps foo.opam -A 3
```
Leonidas-from-XIV, line 20: Depending on whether we assume the user has `opam` one can use `opam show -f x-extra-doc-deps ./foo.opam --raw`. But maybe we shouldn't assume so in the test, in which case this is entirely alright.

test/blackbox-tests/test-cases/doc-dependencies.t/run.t
```diff
@@ -0,0 +1,53 @@
...
+
+  $ mv dune-project-2 dune-project
+
+  $ dune build foo.opam
+  File "dune-project", lines 6-8, characters 1-43:
+  6 |  (depends
+  7 |   (ppx_derivers (>= 1.0))
+  8 |    brr)
+  Warning: The depends field of the documentation stanza can only be non-empty
+  when the documentation stanza is inside a package stanza.
```
Leonidas-from-XIV, line 53: I think the test should also make sure that `(documentation (url ...) (other_fields))` is not enabled before Dune 3.20 (or so).

Leonidas-from-XIV (changes requested):
As @panglesd states, the main missing bit is the dune-lang check and the test for it.

The rest are fairly minor things that I came across when reading through it, but overall the code looks really solid to me.

punchagan:
> As @panglesd states, the main missing bit is the dune-lang check and the test for it.
> 
> The rest are fairly minor things that I came across when reading through it, but overall the code looks really solid to me.

Thanks for the quick review @Leonidas-from-XIV! I'll try and address your comments as I can, starting with the dune-lang check.

rgrinberg:
I haven't followed the entire PR, but is there an explanation on how this new field differs from `with-doc`?

shonfeder:
I think that until we have a good answer to https://github.com/ocaml/dune/pull/11716#issuecomment-3048428642 we can't land this PR as it is.

We have delayed addressing that question due to other priorities overshadowing this, but it is time we turn our attention back here and figure out if we can't instead make use of `with-doc`. It's not yet clear who will have bandwidth to tackle the design question here, but we will try to sort that this week.

Leonidas-from-XIV:
@rgrinberg For the `x-extra-doc-deps`, you're right we can just generate `:with-doc`. However the [odoc config file](https://ocaml.github.io/odoc/odoc/odoc_for_authors.html#config-file) requires the "findlib" name of the package, not the opam package name. Thus we need to specify them somewhere (like we do in the `libraries` stanza in `dune` files).

rgrinberg:
> However the [odoc config file](https://ocaml.github.io/odoc/odoc/odoc_for_authors.html#config-file) requires the "findlib" name of the package, not the opam package name. Thus we need to specify them somewhere (like we do in the libraries stanza in dune files).

This is a file that is produced by odoc and should be installed by dune? How does dune know when to build this file then?
Is there going to be a different build command generated when with-doc is given?

Leonidas-from-XIV:
This file is supposed to be produced and installed by Dune (`Install_rules` in this PR), hence we need to store that information somewhere.

At the moment this PR adds a  `documentation`->`depends` stanza which takes package names and versions and just uses the package names. So in that regard it could be changed to just iterate over the regular depends and pick the ones that are marked `:with-doc`. I am thinking of doing exactly that in #13476, thus there would be no need for changes in the `documentation` stanza at all.

However this only works as long as the OPAM name and the library name match. So maybe the `documentation` stanza should actually keep a list of library names.

jonludlam:
As a side effect of this PR, we slightly changed how odoc_driver interprets the odoc-config file: https://github.com/ocaml/odoc/pull/1343. As a consequence of that, I think it's quite reasonable to only be including the `(packages ...)` field and not the `(libraries)` field as well. As long as we can specify both 'with-doc' and 'post' in the regular dependencies then I agree we don't need a new `(documentation)` stanza for this use case.

jonludlam:
ocurrent/ocaml-docs-ci has [already been updated](https://github.com/ocurrent/ocaml-docs-ci/commit/c6231fa383820b4c700aaa1e72107536b1872112) to handle the 'with-doc & post' as well as 'x-extra-doc-deps'

Leonidas-from-XIV:
Reading the PR I don't fully understand the semantics of `(package)` in odoc. Does that mean that I can use either findlib or OPAM name here? So assuming I have  `(bar (>= 1.0))` in my doc depends but the library that it installs is actually `foo`, that would mean that currently this PR would write `(packages bar)` into the odoc `sexp` config file. Is that the right thing?

jonludlam:
> So assuming I have `(bar (>= 1.0))` in my doc depends but the library that it installs is actually `foo`, that would mean that currently this PR would write `(packages bar)` into the odoc `sexp` config file. Is that the right thing?

Correct, yes.

Leonidas-from-XIV:
I was asked to spec out the behavior so I've created #13485 to describe what I understand the feature should do so we are on the same page.

shonfeder:
Superseded by https://github.com/ocaml/dune/pull/13476, with the design factored out into https://github.com/ocaml/dune/issues/13485 .

PR #11717, opened 2025-04-30
chore: bump dune to 3.19

rgrinberg:


Leonidas-from-XIV:
Is there a reason for this bump? The bump to 3.18 was necessary to get `x-maintenance-intent` and was only possible after 3.18 was released to opam so the CI could install a compatible release from `opam`, which is why it only happened recently.

rgrinberg:
It's needed to use the recent `foreign_objects` features introduced to `foreign_library`. I can wait until 3.19 is released before merging this though.

maiste (approved):

PR #11718, opened 2025-04-30
test: add a target to test bootstrap

rgrinberg:
useful for checking if bootstrap works without removing the current bootstrapped binary


boot/duneboot.ml
```diff
@@ -1,6 +1,9 @@
 (** {2 Command line} *)
 
-let concurrency, verbose, debug, secondary, force_byte_compilation, static =
+let build_dir = "_boot"
```
Alizter, line 3: This ended up being unused

PR #11720, opened 2025-04-30
feature: vendor blake3 mini

rgrinberg:
This requires 3.19 to be released

### DIFF
--- a/boot/libs.ml
+++ b/boot/libs.ml
@@ -94,6 +94,11 @@ let local_libraries =
     ; include_subdirs_unqualified = false
     ; special_builtin_support = None
     }
+  ; { path = "vendor/ocaml-blake3-mini"
+    ; main_module_name = Some "Blake3_mini"
+    ; include_subdirs_unqualified = false
+    ; special_builtin_support = None
+    }
   ; { path = "otherlibs/chrome-trace/src"
     ; main_module_name = Some "Chrome_trace"
     ; include_subdirs_unqualified = false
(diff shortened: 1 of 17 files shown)


Leonidas-from-XIV:
While skimming the docs I learned that OCaml 5.2 started supporting BLAKE2 in its [Digest module](https://ocaml.org/manual/5.2/api/Digest.html). Maybe using BLAKE2 would make more sense, then Dune could stop vendoring BLAKE2 in the future when the minimum supported version is bumped to 5.2?

rgrinberg:
BLAKE2 is only marginally faster than md5 unfortunately. You can see the bench results on the main page of the repo https://github.com/BLAKE3-team/BLAKE3, and I've confirmed them as well. It wouldn't be worth it to use BLAKE2

Leonidas-from-XIV:
Yeah, I didn't know what your goal was for replacing MD5, because it can be either

1. Get a safer cryptographic hash
2. Get a faster hash

As for 1 BLAKE2 is a significant improvement, because MD5 has been outdated for many many years and just about any halfway modern hash is better. As for faster hashing, I wonder if a non-cryptographic hash like SipHash or MurmurHash wouldn't be a better choice? What are you planning to use BLAKE3 for?

rgrinberg:
I'd like dune to be faster at creating content digests for build artifacts.

So my goal is 2. while being limited by 1.

I have experimented with non cryptographic hashes, and while they could work it probably doesn't make sense to relax our digest guarantees if we can speed things while maintaining them.

Leonidas-from-XIV:
Ah, thanks! I somehow missed that.

PR #11725, opened 2025-04-30
test: non-relocatable variables are using the wrong path

maiste:
The test shows non-relocatable package variable substitution is using the sandbox path instead of the non-relocatable path.

To make the test replicable I have used `sed` to strip the root and the hash but if there is any other more elegant way to do it, I would be happy to change it!

Signed-off-by: Etienne Marais <dev@maiste.fr>

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/toolchain-variables.t
+++ b/test/blackbox-tests/test-cases/pkg/toolchain-variables.t
@@ -5,6 +5,7 @@ a mock compiler package using dune's toolchain mechanism.
   $ make_lockdir
 
 We create a fake compiler by creating a configure file.
+
   $ mkdir fake-compiler
   $ cat > fake-compiler/configure << 'EOF'
   > #!/bin/sh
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/pkg/toolchain-variables.t
```diff
@@ -0,0 +1,76 @@
...
+  $ mkdir -p fake-compiler/target/share/bin
+  $ cat > fake-compiler/target/share/bin/ocamlc << EOF
+  > #!/bin/sh
+  > echo "Hello from fake ocamlc!"
+  > EOF
+  $ chmod a+x fake-compiler/target/share/bin/ocamlc
+
+We make sure the installed script is installing the script at the correct
+location
+  $ cat > fake-compiler/Makefile << 'EOF'
```
rgrinberg, line 26: Do you really need make for this test? Would be better to just use a shell script.

Leonidas-from-XIV: I've replaced it with a shell script now.

rgrinberg (approved):

maiste:
Thanks @Leonidas-from-XIV for the fixes! 👌

PR #11726, opened 2025-04-30
fix(pkg): handle empty ID_LIKE

Alizter:
On NixOS for whatever reason `ID_LIKE` was an empty string. In this case we shouldn't use it to find the `os_family` and instead fallback to the `os_distribution`.


Leonidas-from-XIV (approved):
That's an odd choice from NixOS but I agree that this seems like the right way to handle it.

PR #11727, opened 2025-04-30
test: fix meta-exports.t on nixos

Alizter:
The prefix is not correctly automatically detected on NixOS so we have to be explicit with `dune install`.


nojb (approved):
LGTM (but not tested)

PR #11735, opened 2025-05-01
Switch to BLAKE3

rgrinberg:


rgrinberg:
There are no bootstrap rules for assembly yet

PR #11740, opened 2025-05-03
refactor(melange_rules): extract `Output_kind.t`

anmonteiro:
adds a `to_dyn` as well, which is useful for debugging.

### DIFF
--- a/src/dune_rules/melange/melange_rules.ml
+++ b/src/dune_rules/melange/melange_rules.ml
@@ -1,17 +1,48 @@
 open Import
 open Memo.O
 
+module Output_kind : sig
+  type t =
+    | Private_library_or_emit of Path.Build.t
+    | Public_library of
+        { lib_dir : Path.t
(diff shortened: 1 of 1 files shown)


src/dune_rules/melange/melange_rules.ml
```diff
@@ -1,17 +1,48 @@
...
 
+module Output_kind : sig
+  type t =
+    | Private_library_or_emit of Path.Build.t
+    | Public_library of
+        { lib_dir : Path.t
+        ; output_dir : Path.Build.t
+        }
+
+  val [@ocaml.warning "-32"] to_dyn : t -> Dyn.t
```
rgrinberg, line 12: Doesn't `let[@ocaml.warning "-32"]` also work? That might avoid the signature.

anmonteiro: cleaned up in https://github.com/ocaml/dune/pull/11748

PR #11751, opened 2025-05-04
fix: typo in `Dune_rules.Lib` error message

anmonteiro:


src/dune_rules/lib.ml
```diff
@@ -920,7 +920,7 @@ end = struct
         User_error.raise
           ~loc
           [ Pp.text
-              "librarys does not exist but is automatically provided. It cannot be used \
+              "library does not exist but is automatically provided. It cannot be used \
                in this position"
```
Alizter, line 924:
```suggestion
               in this position."
```

Issue #11629, opened 2025-04-15
Allow user to override default behaviour.


kemsguy7:
This is a subtask for issue #11561  where users can make  behaviour optional, defaulting to enabled

gridbugs:
Fleshing out this description a little, this would be a follow-up to https://github.com/ocaml/dune/pull/11600. In the unlikely event that a user does intend to refer to the version number which is literally the string "version", they should be able to disable the warning that prints when common typos are found in dune-project. This could be implemented in the form of a new field in dune-workspace or dune-project.

PR #11600, opened 2025-04-03
Detection and warning for common typos in package dependency constraints

kemsguy7:
@gridbugs, Here's my draft PR for issue. #11575, please kindly review and give feedback let me know if I'm on track.

PR #11759, opened 2025-05-05
feat: add configurable typo detection for package dependency constraints

kemsguy7:


src/dune_config/config.ml
```diff
@@ -86,6 +86,8 @@ let make ~name ~of_string ~default =
 ;;
 
 let make_toggle ~name ~default = make ~name ~default ~of_string:Toggle.of_string
+
+(* Existing config options *)
```
Leonidas-from-XIV, line 90: This comment makes no sense, when reading the code all of the options are existing.

src/dune_config/config.ml
```diff
@@ -147,3 +149,8 @@ let threaded_console_frames_per_second =
       | None -> Error (sprintf "could not parse %S as an integer" x))
     ~default:`Default
 ;;
+
+(* New configuration option for typo warnings *)
```
Leonidas-from-XIV, line 153: No need for the comment, this can be determined by `git blame`.

src/dune_lang/package_dependency.ml
```diff
@@ -40,12 +40,64 @@ let encode { name; constraint_ } =
   | Some c -> pair Package_name.encode Package_constraint.encode (name, c)
 ;;
 
+(* Check for common typos in package dependency constraints *)
+let check_for_typo ~loc { name; constraint_ } =
+  (* Only check for typos if warnings are enabled *)
+  if Dune_config.Config.(get typo_warnings) = `Enabled
```
Leonidas-from-XIV, line 46: I would probably flip the condition here so the `None` is the early out option as then you can then mentally concentrate on the other option, without the `None` option hanging out at the very end of the code.

src/dune_lang/package_dependency.ml
```diff
@@ -40,12 +40,64 @@ let encode { name; constraint_ } =
...
+  then
+    let open Package_constraint in
```
Leonidas-from-XIV, line 48: I'd probably avoid the local open here or if it is this verbose, then use `let module C = Package_constraint in` to avoid potentially pulling in unexpected values.

test/blackbox-tests/test-cases/typo-detection.t
```diff
@@ -0,0 +1,29 @@
...
+  might be a mistake.
+  Hint: Did you mean to use the `:version` variable instead? Example: (depends
+  (bar (= :version)))
+  File "dune-project", line 7, characters 2-18:
+  7 |   (baz :with_test)   ; Should detect typo and suggest :with-test
+        ^^^^^^^^^^^^^^^^
+  Warning: Possible typo in constraint for dependency "baz": ':with_test' might
+  be a mistake.
+  Hint: Did you mean to use ':with-test' instead? Example: (depends (baz
+  :with-test))
```
Leonidas-from-XIV, line 29: I think it would be nice to also demonstrate that fixing these issues makes the messages go away.

Leonidas-from-XIV (commented):
Looks overall pretty good, but I have some suggestions that could improve readability of the code.

kemsguy7:
@gridbugs , here's updates I made for issue #11629 .

src/dune_config/config.ml
kemsguy7: Thanks @Leonidas-from-XIV for the feedback, i'm implementing them.

test/blackbox-tests/test-cases/typo-detection.t
kemsguy7: @Leonidas-from-XIV , so this particular test was for the first part of this pr : https://github.com/ocaml/dune/pull/11600

The test for this particular PR is in **test/blackbox-tests/test-cases/typo-detection2.t**

test/blackbox-tests/test-cases/typo-detection2.t
gridbugs, line 1: This should be a cram test like typo-detection.t. Also please choose a more meaningful name for this file.

test/blackbox-tests/test-cases/typo-detection.t
gridbugs: You should rebase your branch onto the current tip of origin/main since this file was added in your previous PR which has been merged.

test/blackbox-tests/test-cases/typo-detection2.t
Leonidas-from-XIV: It probably doesn't get executed because there is no output from `dune pkg lock` as far as I see.

kemsguy7:
@Leonidas-from-XIV @gridbugs , I've created updates for this pr here: https://github.com/ocaml/dune/pull/11766 .  Please check.

Leonidas-from-XIV:
@kemsguy7 Why is that a separate PR? It would be better to push your improvements in this PR as then we can refer to the comments that were made in previous reviews.

kemsguy7:
@Leonidas-from-XIV @gridbugs , please kindly review, I've applied updates ass suggested in the comments above

Leonidas-from-XIV:
> Sure, please do you mind pointing out the parts that have been unaddressed If you have the time?

Sure. I checked the current state of the PR and these are still unaddressed:

1. [This one](https://github.com/ocaml/dune/pull/11759#discussion_r2073381727)
2. [This one](https://github.com/ocaml/dune/pull/11759#discussion_r2073382279)
3. [This one](https://github.com/ocaml/dune/pull/11759#discussion_r2073384631)
4. [This one](https://github.com/ocaml/dune/pull/11759#discussion_r2073386352)
5. [And this one](https://github.com/ocaml/dune/pull/11759#discussion_r2073390103)

gridbugs:
@kemsguy7 the branch this PR is coming from does not include df45715469776a069fda5a4a39daabbb5675a0fe, since it's based on a version of origin/main from before your previous PR was merged. This is why it looks like the changes from your previous PR are duplicated in this PR which is making it hard to review.

Running `git pull origin main --rebase` (and dealing with any conflicts) will update your branch so that all your recent changes are applied to the current tip of origin/main.

kemsguy7:
> @kemsguy7 the branch this PR is coming from does not include [df45715](https://github.com/ocaml/dune/commit/df45715469776a069fda5a4a39daabbb5675a0fe), since it's based on a version of origin/main from before your previous PR was merged. This is why it looks like the changes from your previous PR are duplicated in this PR which is making it hard to review.
> 
> Running `git pull origin main --rebase` (and dealing with any conflicts) will update your branch so that all your recent changes are applied to the current tip of origin/main.

Thank you for this. 
@Leonidas-from-XIV , @gridbugs please kindly review, I've updated with origin main and also implemented changes as requested

Leonidas-from-XIV (changes requested):
Thanks for the improvements. There's still some things that I think needs adjusting before they can be merged.

In particular you also need to run `dune fmt` to reformat the code to our preferred formatting setting.

src/dune_lang/package_dependency.ml
```diff
@@ -28,61 +28,62 @@ module Well_formed_name = struct
...
-        [ Pp.textf
-            "Possible typo in constraint for dependency %S: '(= version)' might be a \
-             mistake."
-            (Package_name.to_string name)
-        ]
-        ~hints:
+  (* Early return if warnings are disabled *)
+  if Dune_config.Config.(get typo_warnings) <> `Enabled then None
+  else
+    let open Package_constraint in
```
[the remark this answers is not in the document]
kemsguy7: Sure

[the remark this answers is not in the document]
kemsguy7: @Leonidas-from-XIV please kindly review

src/dune_config/config.ml
```diff
@@ -147,3 +147,7 @@ let threaded_console_frames_per_second =
       | None -> Error (sprintf "could not parse %S as an integer" x))
     ~default:`Default
 ;;
+
+let typo_warnings =
```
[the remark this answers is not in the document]
kemsguy7: Do you mean this file config.ml?

src/dune_lang/package_dependency.ml
```diff
@@ -28,61 +28,73 @@ module Well_formed_name = struct
...
 
 let encode { name; constraint_ } =
-  let open Encoder in
+  let open Dune_sexp.Encoder in
   match constraint_ with
   | None -> Package_name.encode name
   | Some c -> pair Package_name.encode Package_constraint.encode (name, c)
 ;;
 
+(* Check for common typos in package dependency constraints *)
```
gridbugs, line 43: This line is duplicated.

src/dune_lang/package_dependency.ml
```diff
@@ -28,61 +28,73 @@ module Well_formed_name = struct
   module TT = Dune_util.Stringlike.Make (T)
 
   let decode =
-    let open Decoder in
+    let open Dune_sexp.Decoder in
```
gridbugs, line 31: Did something change to make this addition necessary now?

src/dune_lang/package_dependency.ml
```diff
@@ -28,68 +28,79 @@ module Well_formed_name = struct
...
-        [ Pp.textf
-            "Possible typo in constraint for dependency %S: '(= version)' might be a \
-             mistake."
-            (Package_name.to_string name)
-        ]
-        ~hints:
+let dependency_constraint_variable_typo_warnings ~loc { name; constraint_ } =
+  (* Handle case where config system isn't initialized yet *)
+  let warnings_enabled =
+    try Dune_config.Config.(get typo_warnings) = `Enabled with
```
Leonidas-from-XIV: What is the exception here? I think for the most part we try not to use exceptions.

src/dune_lang/package_dependency.ml
```diff
@@ -28,68 +28,79 @@ module Well_formed_name = struct
...
+                (Package_name.to_string name)
+            ]
+      in
+      Some message
+    | _ -> None)
 ;;
 
 let decode =
-  let open Decoder in
+  let open Dune_sexp.Decoder in
```
Leonidas-from-XIV: What change requires this adjustment? I have doubts it is necessary.

test/blackbox-tests/test-cases/disabled_warning.t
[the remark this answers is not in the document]
kemsguy7: Done

src/dune_lang/package_dependency.ml
kemsguy7: Fixed

src/dune_lang/package_dependency.ml
kemsguy7: @Leonidas-from-XIV I wrote this in response to a failing test that was working before my PR ```test/blackbox-tests/test-cases/pkg/additional-constraints.t```.

src/dune_lang/package_dependency.ml
kemsguy7: @Leonidas-from-XIV @gridbugs i've updated this in my latest push

src/dune_lang/package_dependency.ml
Leonidas-from-XIV: Yes, but does `Dune_config.Config.get` throw an exception? If so, which one and why? If so, the `with` case should mention the exception and not just blanket-catch every exception. If not, then this `try` line should be removed.

src/dune_lang/package_dependency.ml
Leonidas-from-XIV: Please check again because the changes are still in place. Check the "files changed" tab in this PR and you'll see that the changes still exist.

src/dune_config/config.ml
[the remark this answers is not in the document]
gridbugs: This isn't resolved. The variable is still named `typo_warnings`. It should have a more descriptive name reflecting the nature of typos which it enables warnings for, since it only warns about typos in a very specific situation.

src/dune_lang/package_dependency.ml
kemsguy7: > What is the exception here? I think for the most part we try not to use exceptions.

@Leonidas-from-XIV i've updated this , please check.

src/dune_lang/package_dependency.ml
```diff
@@ -28,57 +28,68 @@ module Well_formed_name = struct
...
-            (Package_name.to_string name)
-        ]
-        ~hints:
+let dependency_constraint_variable_typo_warnings ~loc { name; constraint_ } =
+  (* Handle case where config system isn't initialized yet *)
+  let warnings_enabled =
+    try Dune_config.Config.(get typo_warnings) = `Enabled with
+    | Code_error.E _ -> true (* Only catches Code_error.E exceptions *)
+  in
+  (* Early return if warnings are disabled *)
```
gridbugs: Nitpick about terminology, but this is not an early return. I don't think this comment adds information that isn't obvious reading the code.

src/dune_lang/package_dependency.ml
```diff
@@ -28,57 +28,68 @@ module Well_formed_name = struct
...
-            "Possible typo in constraint for dependency %S: '(= version)' might be a \
-             mistake."
-            (Package_name.to_string name)
-        ]
-        ~hints:
+let dependency_constraint_variable_typo_warnings ~loc { name; constraint_ } =
+  (* Handle case where config system isn't initialized yet *)
+  let warnings_enabled =
+    try Dune_config.Config.(get typo_warnings) = `Enabled with
+    | Code_error.E _ -> true (* Only catches Code_error.E exceptions *)
```
gridbugs: We definitely shouldn't ever be catching `Code_error` exceptions. The intention of these types of exceptions is to catch programming mistakes inside dune, similar to an assert statement. Did you find that calling `Dune_config.Config.(get typo_warnings)` raised a `Code_error.E` exception?

src/dune_lang/package_dependency.ml
```diff
@@ -28,57 +28,68 @@ module Well_formed_name = struct
...
+  in
+  (* Early return if warnings are disabled *)
+  if not warnings_enabled
```
gridbugs: I recommend avoiding negating the condition of an `if`-statement if it also has an `else` clause (which is almost always the case). Negating the condition is equivalent to swapping the `then` and `else` clauses, and I've found that the latter is easier to understand and less error-prone. If the reason you've structured it this way is because you want the simple case to appear first in the code, consider using a `match` statement instead.

That being said, in this particular case you assign the `warnings_enabled` with the value `Dune_config.Config.(get typo_warnings) = Enabled`. It would be more concise and readable to instead match on `Dune_config.Config.(get typo_warnings)` directly rather than going via an intermediate boolean variable.

gridbugs (changes requested):
Left a few suggestions for improving the readability. Also the current approach to handling exceptions raised by `Dune_config.Config.(get typo_warnings)` doesn't look correct to me.

src/dune_lang/package_dependency.ml
gridbugs: Possibly moot because this line will be removed/changed while applying feedback from this review, but try to avoid comments that don't add any clarity. Assume readers know ocaml syntax, in which case they already understand that `| Code_error.E _ -> ...` in the context of a `try` statement means to catch `Code_error.E` exceptions. A more useful comment would explain why it's necessary to catch such exceptions, or perhaps why other exceptions are not caught. The comment should explain your intent so readers understand it, and maintainers can preserve your intent through changes in the future, or understand the consequence of deleting this code if necessary.

But as I said, it doesn't really matter in this specific instance since I suspect this line will be deleted anyway and as @Leonidas-from-XIV said above, we try to avoid using exceptions as much as possible in dune.

test/blackbox-tests/test-cases/disabled_warning.t
```diff
@@ -0,0 +1,60 @@
...
+  >  (name foo)
+  >  (allow_empty)
+  >  (depends
+  >   (bar (= version))  ; Should NOT show warning with typo_warnings disabled
+  >   (baz :with_test)   ; Should NOT show warning with typo_warnings disabled
+  >   qux))
+  > EOF
+
+  $ DUNE_CONFIG__TYPO_WARNINGS=disabled dune pkg lock
+  $ # No warnings should be produced when the feature is disabled
```
Leonidas-from-XIV, line 60: There is no command being run here, so it is missing the test.

src/dune_lang/package_dependency.ml
```diff
@@ -28,57 +28,62 @@ module Well_formed_name = struct
...
   let decode =
-    let open Decoder in
+    let open Dune_sexp.Decoder in
     TT.decode >>| T.to_package_name
   ;;
 end
 
 let encode { name; constraint_ } =
-  let open Encoder in
+  let open Dune_sexp.Encoder in
```
Leonidas-from-XIV, line 37: Is this change necessary?

src/dune_config/config.mli
```diff
@@ -60,10 +60,13 @@ val threaded_console : Toggle.t t
...
+val typo_warnings : Toggle.t t
+
 (** Before any configuration value is accessed, this function must be called
     with all the configuration values from the relevant config file
     ([dune-workspace], or [dune-config]).
 
     Note that environment variables take precedence over the values passed here
     for easy overriding. *)
-val init : (Loc.t * string) String.Map.t -> unit
+val init : (Loc.t * string) String.Map.t -> unit
```
Leonidas-from-XIV, line 72: Please do not remove the trailing newlines. It creates changed lines where there are no changes actually.

Leonidas-from-XIV (changes requested):
Please make sure to run `dune fmt` to format the source code automatically and take a look at the outstanding review comments.

src/dune_lang/package_dependency.ml
kemsguy7: @Leonidas-from-XIV  please can you review the change i did here?  it's making the additional constraint test fail

src/dune_lang/package_dependency.ml
kemsguy7: reverted.

src/dune_lang/package_dependency.ml
kemsguy7: @gridbugs updated

src/dune_config/config.ml
```diff
@@ -86,6 +86,7 @@ let make ~name ~of_string ~default =
 ;;
 
 let make_toggle ~name ~default = make ~name ~default ~of_string:Toggle.of_string
+
```
gridbugs, line 89: Please try to remember to run the autoformatter on your changes. I thought we set up your vscode to automatically format files when you save them. Did that stop working?

src/dune_lang/package_dependency.ml
gridbugs: I still see the change in the diff. Can you remove this change (and the one above that I commented on).

gridbugs (changes requested):
Left a few more comments about formatting. Also please make sure to address some of the older comments still outstanding (use the "Files changed" tab to easily see them).

Issue #11045, opened 2024-10-28
Regression in #11028 when using `dune subst` on a project with an opam file


btjorge:
I got notified of the closing of #10927, so I tried to test my opam file pin use-case, but the build failed with an error. Here's a reproduction test without opam :

## Reproduction

```sh
$ mkdir test && cd test

$ cat > dune-project << EOF
(lang dune 3.17)
(name test)
(version 0.1)
(generate_opam_files)
(package (name test) (allow_empty))
EOF

$ dune build # generates the opam file
```

## Previous Behavior

```sh
$ dune subst # does not fail
```

## Actual Behavior

```sh
$ dune subst 
File "test.opam", line 1, characters 10-25:
1 | # This file is generated by dune, edit dune-project instead
              ^^^^^^^^^^^^^^^
Error: repository does not contain any version information
```
Note that the substitution works has expected

## Specifications

- dune from main branch

henrytill:
Did #11204 fix this?

There's a [related issue](https://github.com/tweag/opam-nix/issues/115) (regression?) affecting users of [opam-nix](https://github.com/tweag/opam-nix), where `dune subst` being run in a Nix build environment (where neither a `git` executable nor a `.git` directory are present) worked before without issue, but now, as of `(lang dune 3.17)`, the `subst` step fails with the message reported above.  Users can add `(subst disabled)` to their `dune-project` to work around this, but it's a regression from previous behavior where the default `subst` behavior was enabled outside of a Nix build environment, and fresh projects created with `dune init proj` worked with opam-nix out of the box.

Should I open a new issue for this?

Blaisorblade:
Just got this issue with dune 3.17.2, `git` available and no Nix involved.
```
$ dune --version
3.17.2
```

My source folder contains both `.git` and an `elpi-extra` folder containing the `.opam` file — but I've opam-pinned the package, and that copies `elpi-extra` but of course not `.git`.

```
$ ls -d .git
.git
$ ls elpi-extra/
accumulate_bug  dune  dune-project  elpi-extra.opam  extra  grid.ml  LICENSE-BedRock  README.md  tests
```

Richard-Degenne:
I am encountering the same problem: I have a dune project located in the subfolder of a Git repository, and that causes `dune subst` to fail. The easiest way to reproduce would be as follows.

```bash
mkdir dune-test && cd dune-test && \
git init && \
dune init proj subfolder && \
git add . && git commit --message "Initial commit" && \
[...]

PR #11760, opened 2025-05-05
Allow `dune subst` to work inside a subfolder of a Git repository

Richard-Degenne:
I gave my best shot at fixing #11045, but tests fail with a runtime error that I have a hard time understanding. If anyone could give me some pointers or complete the fix, that would be awesome.

```
Description:
  ("Fdecl.get: not set", {})
Called from Dune_rules__Only_packages.Clflags.t in file
  "src/dune_rules/only_packages.ml" (inlined), line 25, characters 13-24
Called from Dune_rules__Only_packages.filter_packages_in_project in file
  "src/dune_rules/only_packages.ml", line 68, characters 8-20
Called from Fiber__Core.O.(>>|).(fun) in file "vendor/fiber/src/core.ml",
  line 253, characters 36-41
  line 76, characters 8-11
  line 76, characters 8-11
  line 76, characters 8-11
-> required by ("find-dir-raw", In_source_tree ".")
```

### DIFF
--- a/bin/subst.ml
+++ b/bin/subst.ml
@@ -317,11 +317,6 @@ let subst vcs =
      and+ files = Vcs.files vcs in
      Some (version, commit_id, files)
    | None ->
-     (* We have to do this because scanning the source tree evaluates [-p].
-        That's because [-p] is needed to interpret packages in dune projects
-        correctly. It should not be necessary, so we should probably make the
-        package loading lazier. *)
-     Dune_rules.Only_packages.Clflags.set No_restriction;
      let* root = Source_tree.root () in
      let project = Source_tree.Dir.project root in
      if Dune_project.dune_version project < (3, 17)
@@ -442,16 +437,7 @@ let subst vcs =
       then subst_file path ~map:watermarks opam_package_files))
 ;;
 
-let subst () =
-  (* CR-someday rgrinberg: use [Source_tree.nearest_vcs] *)
-  Sys.readdir "."
-  |> Array.to_list
-  |> String.Set.of_list
-  |> Vcs.Kind.of_dir_contents
-  |> Option.map ~f:(fun kind -> { Vcs.kind; root = Path.root })
-  |> subst
-  |> Memo.run
-;;
+let subst () = Source_tree.nearest_vcs Path.Source.root |> Memo.bind ~f:subst |> Memo.run
 
 (** A string that is "%%VERSION%%" but not expanded by [dune subst] *)
 let literal_version = "%%" ^ "VERSION%%"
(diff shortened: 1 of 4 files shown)


Alizter:
I believe the issue is coming from:

```
  (* We have to do this because scanning the source tree evaluates [-p].
    That's because [-p] is needed to interpret packages in dune projects
    correctly. It should not be necessary, so we should probably make the
    package loading lazier. *)
  Dune_rules.Only_packages.Clflags.set No_restriction;
```

It looks like to me that this doesn't get set if there is a `vcs` present. We probably want that to be set in both cases by moving this before the match if we want `Source_tree` to work correctly. Not sure if this will fix all the issues however.

Alizter:
Another comment, it would be great if you could include a commit before with a reproduction cram test. That we can be sure this commit will fix the issue and have a nice commit history.

You want to add the test in `test/blackbox-tests/test-cases/subst/`.

Richard-Degenne:
Awesome, thank you so much for the quick reply. I'll see what I can do for the test.

Richard-Degenne:
This is my first time using cram, but the test seems to be working: it fails without the fix, but passes afterwards.

Let me know if there is anything I should have done differently.

Side question: is there a way to run a single test file? I've been running `make test` every time, it's kind of sluggish 😅

Alizter:
You can run a single cram test by building the alias given by its name. Simply do `./dune.exe build @git-subfolder` to build the test. Another tip is to use watch mode `-w` and `--auto-promote` which will automatically promote the current output of the test so that it succeeds. When you update the test, dune will make sure the expected output is updated. 

Good first try, just a couple of comments:
- Usually we want to test to always succeed, even if it outputs an error in the test itself.
- We try to keep the reproduction case small so it is clearer which parts are causing issues. `dune init` does a bit more than is necessary so I've gone ahead and stripped out the important parts.

You should update the first commit to have this cram test:
```
Running `dune subst` in a subfolder of a Git repository should work.
Regression test for https://github.com/ocaml/dune/issues/11045

  $ git init --quiet

  $ mkdir subfolder
  $ cat > subfolder/dune-project <<EOF
  > (lang dune 3.19)
  > (name subfolder)
  > EOF

  $ cat > subfolder/subfolder.opam

`dune subst` requires at least one commit in the repository.

  $ git add .
  $ git commit --quiet --message "Initial commit"
  $ cd subfolder
  $ dune subst 2>&1 | grep "Internal error: Uncaught exception"
  | Internal error: Uncaught exception.
```

In the second commit, you will of course have the updated output.

Richard-Degenne:
@rgrinberg: Thank you for completing the fix! Is there anything else that needs to be done before the PR can be considered for merging?

PR #11765, opened 2025-05-06
refactor: remove a bunch of unusable opam code

rgrinberg:
Most of this code is not compatible with dune's existing code and dune already has existing libraries to do all of this work.

Let's get rid of this code so that people aren't led astray.


Leonidas-from-XIV (approved):
Indeed, it looks like all of that code is unused, might as well remove it. Thanks for the PR!

PR #11771, opened 2025-05-07
test: fix utop tests

rgrinberg:


Leonidas-from-XIV:
I was worried that the test never terminates on Linux because so far it has been stuck for hours running the Github Action, but it works just fine on my Linux machine.

Alizter:
@Leonidas-from-XIV Which kernel version do you have?

Alizter:
I think the correct fix is `Stdlib.exit` instead. I got it working in #11788.

Alizter:
It won't work on 5.3 since the `help` message has extra quotations and that isn't backwards compatiable. We have to wait until we bump the CI to 5.3.

Alizter:
We ended up fixing this when updating the CI to 5.3. I have since found out why this change appears in utop, and its to do with compiler lib changes with messages and their color.

PR #11773, opened 2025-05-07
feature: depopt selection in the workspace

### LINKED ISSUE #11643: Enable selecting `depopts` in `dune pkg`.
## Desired Behavior

`dune pkg` currently ignores optional dependencies specified with `depopts`, making it impossible to eg. test a project with some optional dependency installed. The `(constrain)` stanza exists in `lock_dir`, but the documentation clearly specifies that it constrains version that were already going to be installed.

## Example

A possible syntax could be :

```
(lock_dir
  (name with_jsoo.lock)
  (depends js_of_ocaml))
```

rgrinberg:
Add a [deopts] field to allow users to select depopts.

The depopts are selected globally for all packages that list them. This is inline with how opam does things.

Fixes #11643

### DIFF
--- a/src/dune_rules/workspace.ml
+++ b/src/dune_rules/workspace.ml
@@ -13,6 +13,7 @@ module Lock_dir = struct
     ; repositories : (Loc.t * Dune_pkg.Pkg_workspace.Repository.Name.t) list
     ; constraints : Dune_lang.Package_dependency.t list
     ; pins : (Loc.t * string) list
+    ; depopts : (Loc.t * Package.Name.t) list
     }
 
   let to_dyn
@@ -23,6 +24,7 @@ module Lock_dir = struct
         ; repositories
         ; constraints
         ; pins
+        ; depopts
         }
     =
     Dyn.record
@@ -37,6 +39,7 @@ module Lock_dir = struct
             (List.map repositories ~f:snd) )
       ; "constraints", Dyn.list Dune_lang.Package_dependency.to_dyn constraints
       ; "pins", (Dyn.list Dyn.string) (List.map pins ~f:snd)
+      ; "depopts", (Dyn.list Package.Name.to_dyn) (List.map ~f:snd depopts)
       ]
   ;;
 
@@ -48,6 +51,7 @@ module Lock_dir = struct
         ; repositories
         ; constraints
         ; pins
+        ; depopts
         }
     =
     Poly.hash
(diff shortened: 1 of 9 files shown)


src/dune_rules/workspace.ml
```diff
@@ -37,6 +39,7 @@ module Lock_dir = struct
             (List.map repositories ~f:snd) )
       ; "constraints", Dyn.list Dune_lang.Package_dependency.to_dyn constraints
       ; "pins", (Dyn.list Dyn.string) (List.map pins ~f:snd)
+      ; "constraints", (Dyn.list Package.Name.to_dyn) (List.map ~f:snd depopts)
```
Leonidas-from-XIV, line 42: I assume this is a copy-paste oversight:

```suggestion
      ; "depopts", (Dyn.list Package.Name.to_dyn) (List.map ~f:snd depopts)
```

Leonidas-from-XIV (changes requested):
Apart from the typo it looks good and the test works really well. Could use a mention of the `depopts` field in the documentation.

What do you think about allowing something like `:all` to enable all `depopts`? I recently was wanting for this in OPAM but it doesn't support it such an option.

src/dune_rules/workspace.ml
rgrinberg: Indeed it was

rgrinberg:
> What do you think about allowing something like `:all` to enable all `depopts`? I recently was wanting for this in OPAM but it doesn't support it such an option.

Turns out that such a thing isn't that simple to implement. The only way to determine *all* depopts is to first find a solution.

We could perhaps implement something that's close enough semantically by introducing a "depopts_required" mode. This would just merge depopt constraints into depends constraints for every package and solve. I don't know if that's going to work, but you're welcome to experiment if you have a use case for this.

Leonidas-from-XIV (approved):

Alizter:
@bnguyenvanyen Since this is a package management related PR and package management is considered "experimental" until we have a full-release, we generally tend not to include changes in the changelog or add version guards. When package management is officially released, there will be backwards compatible changes but until then it allows us to change the implementation details and design without friction.

bnguyenvanyen:
@Alizter Ok, thank you. I ask because we started using the depopts stanza (at Routine) and it broke downstream packages that did not have the correct dune version (because they're still using opam). We'll pin versions downstream for now I guess.

PR #11775, opened 2025-05-09
pkg: Make lock files targets

Leonidas-from-XIV:


src/dune_rules/lock_rules.ml
```diff
@@ -0,0 +1,125 @@
...
+  let action { target; lock_dir } ~ectx:_ ~eenv:_ =
+    let open Fiber.O in
+    let+ () = Fiber.return () in
+    Printf.eprintf
+      "Running ACTION, target is %s, our lock_dir is %S\n"
+      (Path.Build.to_string target)
+      lock_dir;
+    let path = Path.build target in
+    Path.mkdir_p path;
+    let t = Unix.localtime @@ Unix.gettimeofday () in
```
rgrinberg: I don't think you can include this information here. It makes the action non deterministic unfortunately. You could make it part of the action input if you really want it, but i'd say it's best to skip it.

src/dune_rules/lock_rules.ml
```diff
@@ -0,0 +1,125 @@
...
+
+module Gen_rules = Build_config.Gen_rules
+
+let rule ?loc { Action_builder.With_targets.build; targets } =
+  (* TODO this ignores the workspace file *)
+  Rule.make ~info:(Rule.Info.of_loc_opt loc) ~targets build |> Rules.Produce.rule
+;;
+
+let setup_lock_rules ctx_name ~lock_dir ~projects : Gen_rules.result =
+  let sctx = Super_context.find_exn ctx_name in
```
rgrinberg: You can't use super context for package management rules. See how you've used the lower level rules api  in `Pkg_rules` to work around this.

src/dune_rules/lock_rules.ml
```diff
@@ -0,0 +1,125 @@
...
+  let target =
+    let ( / ) = Path.Build.relative in
+    Private_context.t.build_dir
+    / Context_name.to_string ctx_name
+    / ".lock"
+    / lock_dir
```
rgrinberg: Instead of composing the target from the component, you should make it relative to the `~dir` you were passed in `gen_rules`. This is the only valid directory you can use for targets.

src/dune_rules/lock_rules.ml
```diff
@@ -0,0 +1,125 @@
...
+    / "content"
+  in
+  let gen_rules lock_dir =
+    let lock_rule = lock ~target ~lock_dir in
```
rgrinberg: I don't really get this `lock_dir` argument. It's part of the path in the directory where we are loading the rule and we are passing it to the action?

src/dune_rules/lock_rules.ml
```diff
@@ -0,0 +1,125 @@
...
+    Memo.return @@ Gen_rules.make ~build_dir_only_sub_dirs (Memo.return Rules.empty)
+  | _ -> Memo.return @@ Gen_rules.rules_here Gen_rules.Rules.empty
+;;
+
+let setup_tmp_lock_alias =
+  fun ~dir ctx_name ->
+  let alias = Alias.make ~dir Alias0.pkg_lock in
+  let rule =
+    Rules.collect_unit (fun () ->
+      (* careful, need to point to a file that will be created by the rule *)
```
rgrinberg, line 682: I don't think so, you can point to the directory itself.

src/dune_rules/gen_rules.ml
```diff
@@ -645,7 +645,11 @@ let private_context ~dir components _ctx =
   analyze_private_context_path components
   >>= function
   | `Invalid_context -> Memo.return Gen_rules.unknown_context
-  | `Valid (ctx, components) -> Pkg_rules.setup_rules ctx ~dir ~components
+  | `Valid (ctx, components) ->
+    let projects = Dune_load.projects () in
```
rgrinberg: You might as well just make `Lock_rules.setup_rules` make this call.

src/dune_rules/lock_rules.ml
```diff
@@ -0,0 +1,125 @@
...
+      let* _projects = projects in
+      gen_rules lock_dir)
+  in
+  let directory_targets = Path.Build.Map.singleton target Loc.none in
+  Gen_rules.make ~directory_targets rules
+;;
+
+let setup_rules ~components ~dir ~projects ctx_name =
+  match components with
+  | [ ".lock" ] ->
```
rgrinberg, line 656: I don't fully understand the scheme of rules you're creating, but I don't see why you need more than one directory to define all of your rules. I would expect the following:

1. If you have a single lock directory, then you just need one rule inside `".lock"`
2. If you multiple lock directories, they can all co-exist in `.lock` under different names. 

You don't even need a directory per context. Just a single directory for all contexts will do.

src/dune_rules/lock_rules.ml
Leonidas-from-XIV: I know. This is just for debugging, as I can't make the action rerun on changes in `dune-project` and want to have a way to see whether the rule was run or not. The final PR will not have this anymore.

src/dune_rules/lock_rules.ml
Leonidas-from-XIV: Why can't the super context be used?

rgrinberg: Here you are generating the rule to produce a lock directory, but to initialize the super context you need to first load the lock directory. This will cause a cycle.

Note that the super context is useless for you here anyway. It is what we use to load the environment for basic rules, but tha environment cannot impact the generation of the lock directory. Otherwise, producing lock directories wouldn't be deterministic across machines.

src/dune_rules/lock_rules.ml
Leonidas-from-XIV: It's the identifier for the lock directory we want to generate. Basically the value that `lock_dir`->`path` from `dune-workspace`. In most projects that will always just be `dune.lock`, but for simplicity I thought it would be nice if the rules would generate them in an unchanging place no matter whether you have one or multiple lock directories.

src/dune_rules/lock_rules.ml
Leonidas-from-XIV: That's a good point, I've changed it and it is much nicer now.

src/dune_rules/lock_rules.ml
Leonidas-from-XIV: The scheme is `.lock/<path-of-the-lockdir>/<contents-here>`. I wanted to keep it uniform, so it will always be the same, no matter if the user has one or multiple lock dirs configured in their workspace. That means that `<path-of-the-lockdir>` will mostly be `dune.lock` but I don't see an issue with that.

That way I also hope that the promote rules that promote the generated contents back into the users source tree will be simpler to write and understand as one does not need to know whether the workspace has one or multiple lock directories configured.

rgrinberg: You can do it like that as well. In this particular problem, the names of the targets aren't important and should be invisible to the user, so just about any scheme can be chosen. In fact, a clever approach would pick a name that is a has of the all the inputs to the solver. That way, multiple lock directories could be represented with the same target.

I don't necessarily recommend this, just pointing out that the names are completely irrelevant.
Therefore, I suggest just picking the scheme that will be the easiest to implement. If there's other layouts you'd like to experiment with, you can always do it later. Keep in mind that in an ideal implementation, none of these file targets would even be necessary and we'd just do everything in memory and internal database files. So I really don't see the need to insist on some particular naming scheme.

src/dune_rules/lock_rules.ml
rgrinberg: Okay, I understand. Although note that the path in the workspace is not necessarily a path in the root. It might be something like `foo/bar/dune.lock` as well for example. So you will either need to escape it somehow or make a much more complex loading scheme.

src/dune_rules/lock_rules.ml
Leonidas-from-XIV: I've changed it temporarily to dump the package info as dyn representation into the file, that way it should be a bit more reproducible and allow debugging whether changes in the package definition lead to re-execution of the action. Weirdly I can't get it to regenerate the file, maybe I am dereferencing the `Memo.t` in the wrong location and the dependency does not get registered?

src/dune_rules/lock_rules.ml
Leonidas-from-XIV: I would have thought so too, but when I remove the last path element I get:

```
Error: No rule found for default/.lock/dune.lock (context _private)
-> required by alias dune.lock/pkg-lock
```

Which is surprising to me, since I believe I am attaching a rule there:

```ocaml
  | [ ".lock"; lock_dir ] -> Memo.return @@ setup_lock_rules ~dir ~lock_dir ~projects
```

src/dune_rules/lock_rules.ml
Leonidas-from-XIV: Thanks for the explanation.

I don't really need the super context anyway, I just tried to use it to get a transitive dependency on the project. While I was looking at how `opam_create.ml` creates the dependency, @emillon pointed out that this is most likely due to binding on `Super_context.t Memo.t`. But I can just as well bind on `Dune_project.t list Memo.t` which makes it a bit more explicit that I depend on the project definition.

However, even if I bind on it, I can't get my action to re-run when the `dune-project` is changed, so there's still something missing I believe.

rgrinberg:
I pushed some fixes that should hopefully point to the right direction. The issue was that we need to correctly generate the digest of the action, or specify the dependencies on the files the action is accessing correctly. I chose to do the former, but the latter might work better if we have multiple actions all reading the same files.

src/dune_rules/format_rules.ml
```diff
@@ -151,20 +146,22 @@ let gen_rules_output
...
        if ocamlformat_is_locked
        then (
          let { Action_builder.With_targets.build; targets } = rule in
+         let ctx_name = sctx |> Super_context.context |> Context.name in
          let build =
            let open Action_builder.O in
            let+ build = build
-           and+ env = Action_builder.of_memo (Pkg_rules.dev_tool_env Ocamlformat) in
+           and+ env =
+             Action_builder.of_memo (Pkg_rules.dev_tool_env ctx_name Ocamlformat)
```
rgrinberg: This change seems questionable. Why would we want ocamlformat to depend on the context? Seems like that would kill sharing for no particular reason.

rgrinberg:
I did some preliminary review. Some high level comments first:

* There's lots of spurious changes in this PR. I imagine they're going to be moved into separate PR's. Otherwise, it will be quite hard to review them.

* The additional context passing everywhere seems wrong at first glance. A context itself is enough to determine the lock directory (if one is attached). It should be possible to share a lock directory between contexts. So passing a context + a lock directory does not make sense.

Now, for some more actionable suggestions:

* There's a bunch of changes in `dune_pkg` switching from source to build paths. By itself, this is harmless, but it seems wrong for `dune_pkg` to even care about this distinction as build paths and source paths is a distinction introduced by the build system itself. Therefore, I recommend that we have a PR that switches to `dune_pkg` to use `Path.t` to represent paths in general, and `Path.Local.t` when we'd like to reject external paths. This would be a good change on its own and should make this PR easier to review.

* Here's another idea for splitting this PR further. The code is currently trying to accomplish producing a build plan as a rule. To do this, we need to make sure that the build plan exists in the build directory. How about we try reading the lock dir from the build dir separately? We'd still have `dune pkg lock` produce the build plan in the source, but the rules would never read it directly and rely on copying it first. This is exactly how we read all other source files anyway.

shonfeder:
@rgrinberg 

> How about we try reading the lock dir from the build dir separately? We'd still have dune pkg lock produce the build plan in the source, but the rules would never read it directly and rely on copying it first. This is exactly how we read all other source files anyway.

As per https://github.com/ocaml/dune/issues/12098, one of the main aims we want out of this work is to stop building the lock directory in users worktree *by default*. As we have discussed in meetings, and as detailed in https://github.com/ocaml/dune/issues/12097, this is meant to allow us to bracket the concrete implementation details of the lock dir format for the time being, while we move forward with more pressing features to achieve https://github.com/ocaml/dune/milestone/62

IIUC, your suggestion here is counter to our settled aim, and so I think probably not actionable as a change for this PR. Do you agree?

rgrinberg:
To confirm that we're on the same page, let me describe how I view the situation:

As of today, lock directories must live in the source. Our goal is to move to move to a world with the following:

1. Build plan _can_ be generated on demand and therefore the corresponding lock directories are targets of build rules themselves. Note that how this is going to work still remains to be fleshed out. I mentioned that an element that corresponds to sources used as inputs (opam-repository, pins) must still be recorded somewhere.

2. Lock directories which live in the source are still supported. This was discussed at the last dev meeting. Maintaining our current tests and the desire not to commit to this approach were the motivations.

3. Flip the default from 2. to 1.

I think my suggestion fully supports these goals. It might be that we miscommunicated on the goals above? Or that I didn't properly explain why my suggestion supports these goals?

shonfeder:
@rgrinberg I think our understanding of the long-term outcome is aligned, but it seems that you may be misunderstanding our short-term aims. Please see the plan and motivation detailed in https://github.com/ocaml/dune/issues/12097. In particular note that we there conclude:

> 1. By default, lock directories should be maintained in the _build dir and only promoted into the working directory if a user actually wants to commit it.
> 2. Moving lock directories into the _build dir by default allows us to unblock on the portability consideration, and move forward with a focus on the most pressing requirements for achieving stability.
> 3. But we still want to find and implement the optimal solution to portable package locking, and we want to keep this aim in view for the mid term.
>    - A possible alternative here has been proposed by @rgrinberg, and is also discussed in the [above mentioned post](http://www.chriswarbo.net/blog/2024-05-17-lock_files_considered_harmful.html).

The sub-point under 3 addresses the "element that corresponds to sources used as inputs (opam-repository, pins) [which] must still be recorded somewhere", but this is not something we are trying to solve *right now*. Right now, we want to unlock a mode of use that does not force users to have lock dirs (or indeed any other artifact) in their work directory in order to have packages installed.

And I do not see how your suggestions align with what we have detailed or explained as our short-term, and immanent goal.

rgrinberg:
First of all, my suggestion is not relevant to "element that corresponds to sources used as inputs (opam-repository, pins) [which] must still be recorded somewhere". It is directly relevant to the other points I've mentioned though. I don't understand why you've picked this snippet out of my comment. 

>  but this is not something we are trying to solve right now. Right now, we want to unlock a mode of use that does not force users to have lock dirs (or indeed any other artifact) in their work directory in order to have packages installed.

My suggestion is directly relevant to unlocking this goal. Maybe it's worth rephrasing here. The current pipeline is:

* a lock directory exists in the source
* the rules load the build plan from this source

What we'd like is to support a pipeline where

* a lock directory is a target of a rule running the solver
* the rules load the build plan by reading this target of a rule

Given that we plan to keep the original pipeline as an option - at least for now. We could modify it to be:

* a lock directory is a target of a rule copying the lock directory from the source
* the rules load the build plan from this this target of a rule

The final result is that we've managed to implement the unify the last step between the code paths. There's a bunch of problems to solve with the build plan being a target of a rule even if the rule is trivial (i.e. copying).

shonfeder:
When you say

> a lock directory is a target of a rule copying the lock directory **from the source**

does "the source" not refer to the working tree, i.e., the directory holding the textual data that developers interact with? If so, then how is this compatible with

> a mode of use that does not force users to have lock dirs (or indeed any other artifact) in their work directory

If not, could you clarify what you mean by "from the source"?

Edit: Additionally, we have **not** been planning to "keep the original pipeline as an option - at least for now", **if** that pipeline means automatically putting the lock dir into the user's workspace (outside of the `_build` dir). That basically the whole point. But if that pipeline only means "using a copy rule to read the lock dir *if it mas been manually put into the working directory", then I think we are aligned here.

shonfeder:
Oh, and just to clarify re:

> I don't understand why you've picked this snippet out of my comment.

Only because you included the work *must* there, and I wanted to be clear that we do plan to meet this requirements **eventually**!

Leonidas-from-XIV:
One important part of creating the lock directories via actions and not via copying is that we get autolocking in watch mode, something that has been requested by stakeholders. This is why I think it is reasonable to have the copy rules as a fallback for those people who already have lock dirs (and don't want autolocking to come up with potentially different solutions), but not as the main mechanism of locking because for autolocking. For that we'd need to figure out how to run `dune pkg lock` in some kind of own watch mode and not conflict with Dune watch mode.

Also creating the lock files in the source tree and then copying them to create the build rules leaves the users still unclear whether they should or should not commit their lock files, whether they are stable or not. In some ways makes this temp files since dune (and not the user as in the case of source files) is creating them in a temporary location (the work tree) and then copies them to the actual location (the build dir), which seems like locking with extra steps.

rgrinberg:
From #12097:

> By default, lock directories should be maintained in the _build dir and only promoted into the working directory if a user actually wants to commit it.

I interpret _by default_ to mean that we have options and one them is preferred in the absence of an explicit selection. Therefore, my statement:

> a mode of use that does not force users to have lock dirs (or indeed any other artifact) in their work directory

Is compatible with this interpretation. Because the _mode_ proposed here is one of the options. In particular, the one that should  become the _default_.

> Additionally, we have not been planning to "keep the original pipeline as an option - at least for now"

Then I want to re-iterate my original two objections to removing the original pipeline:

1. Reproducibility has been temporarily postponed until further notice. I've first brought this objection in a slack comment (July 28) and re-iterated it in my comments [here](https://github.com/ocaml/dune/issues/12097#issuecomment-3199970007)

2. It requires a massive and undesirable change to the test suite. Essentially involving the solver in every single test case. This objection was brought up in the previous team meeting.

Furthermore, I'll also point out that this statement:

> and only promoted into the working directory if a user actually wants to commit it.

Only makes sense to me if we can actually read the promoted source. Your ticket should clarify how are the promoted artifacts supposed to be used by the user apart from being committed.

The following response might not be relevant because I don't think I completely understood what you meant here, so I might be filling in some details automatically and incorrectly. Feel free to clarify the user workflow that you have in mind.

> if that pipeline means automatically putting the lock dir into the user's workspace (outside of the _build dir). That basically the whole point. But if that pipeline only means "using a copy rule to read the lock dir *if it mas been manually put into the working directory", then I think we are aligned here.

I think that's what I have in mind, but I'm still fuzzy on the details. Some questions that come to mind:

How do you distinguish between "manually put into the working directory" or promoted into the working directory by dune?

It seems like you expect your dune to "auto lock" in watch mode. I interpret this to mean to write the lock directory into the source every time it re-runs the solver. What happens when you restart dune, does it use this lock directory in any way? If it doesn't, why did we write it in the source in the first place?

In conclusion, there has to be something that corresponds to the "original pipeline" to read the stuff we're promoting, because otherwise, promoting anything makes no sense.

Leonidas-from-XIV:
> It seems like you expect your dune to "auto lock" in watch mode. I interpret this to mean to write the lock directory into the source every time it re-runs the solver.

That was the original plan when I started this PR but we have given up on the promotion into the source for now. Our idea is that Dune, if instructed by `(pkg enabled)` to do so, would take the project definitions and automatically create a lock directory in the build folder and use this lock directory for constructing the package rules. The lock directory would not be visible to the user. This postpones the question of the format and whether lockdirs are portable or not because by default the lock files would be created on the exact same system that the user will be using to build the packages.

However, the user can take the lock directory and put it in the source, in which case dune will run the copy action instead of the solve action and the rules are constructed out of the copied lock dir from the source folder. This allows reproducability for those who want to commit the lock dir and also avoids changing the cram tests that depend on manually creating a lock directory (since their lock dirs will trigger the copy rule).

This is what the PR currently implements. I'm in the process of rebasing it on the merged #12194 to use `Path.t` instead of `Path.Build.t`.

rgrinberg:
Thanks @Leonidas-from-XIV. That last comment makes a lot more sense to me. The only remaining question I have is how are we not preserving the original pipeline as an option? It seems like we clearly are.

> However, the user can take the lock directory and put it in the source, in which case dune will run the copy action instead of the solve action and the rules are constructed out of the copied lock dir from the source folder. This allows reproducability for those who want to commit the lock dir and also avoids changing the cram tests that depend on manually creating a lock directory (since their lock dirs will trigger the copy rule).

This makes sense. My earlier comment suggested that you should take the "run the copy action" bit and turn into a separate PR. That would allow you address the issues with having a lock directory actually being a build target separately from running the solver. Changing the type of paths was only a first step to accomplish this. I'm still not clear what is the objection to doing that first.

I'm also puzzled by your earlier comment:

> Also creating the lock files in the source tree and then copying them to create the build rules leaves the users still unclear whether they should or should not commit their lock files, whether they are stable or not. In some ways makes this temp files since dune (and not the user as in the case of source files) is creating them in a temporary location (the work tree) and then copies them to the actual location (the build dir), which seems like locking with extra steps.

It seems like it hints against copying the lock dir from the source into the build directory? If this comment is still relevant, my response to it is that:

Whether something is copied from the source to the build directory, or is read straight from the source directory is an implementation detail of dune. This copying step should be completely transparent to users (just like all other source copying rules) so it should not confuse users in any way.

Leonidas-from-XIV:
> The only remaining question I have is how are we not preserving the original pipeline as an option? It seems like we clearly are.

Maybe it depends on the definition of "pipeline". What I understood you to describe is a workflow that has `dune pkg lock` create a lock dir in the source. The workflow I describe is turning `dune pkg lock` into an alias for `dune build @pkg-lock` and always them in the build directory (hence the path type was `Path.Build.t` since it never gets written anywhere else), thus only keeping the old way as a compatibility shim. New users of `dune pkg` would not see the lock dir usually, unless they manually create a lock directory.

 > My earlier comment suggested that you should take the "run the copy action" bit and turn into a separate PR.

Maybe I have misunderstood, because I have understood your comment to meant that the planned workflow was to always run the `dune pkg lock` and copy action. Thus the lockfile targets would always be generated by copying from the source directory.

I can certainly take the copy action out of this PR into a separate PR but the copy action seemed like such a small part of this PR that fit the structure of the code rather naturally (https://github.com/ocaml/dune/pull/11775/commits/cbb9b0bffde17943b112997fe3153092d2c156a5) that I didn't see the point of pulling it out into a separate PR. But I can definitely do that and rebase this PR on top of it afterwards.

shonfeder:
I think @Leonidas-from-XIV has managed to convey the key points here in a way that has landed, so I won't bother with point by point followups. Instead, I will try to transpose some of the reiteration here into https://github.com/ocaml/dune/issues/12097 , so that the aim and approach is well documented and available for critical examination in case there are essential flaws in the reasoning.

But please LMK, @rgrinberg, if you still want followup from me on any particular points here!

As a concrete next steps, if extracting the code for the copy action is not too difficult, it sounds worthwhile to me, as it will help us continue making progress while also reducing the PR burden.

rgrinberg:
> Maybe it depends on the definition of "pipeline". What I understood you to describe is a workflow that has dune pkg lock create a lock dir in the source. The workflow I describe is turning dune pkg lock into an alias for dune build @pkg-lock and always them in the build directory (hence the path type was Path.Build.t since it never gets written anywhere else), thus only keeping the old way as a compatibility shim. New users of dune pkg would not see the lock dir usually, unless they manually create a lock directory.

My suggestion would be not do that. It could probably work that way, but I think it's a lot harder than it looks and there's no obvious benefit to this. If only because `dune build @pkg-lock` will not work while dune is already running in watch mode. Moreover, I think `dune pkg lock` takes arguments that would be hard to replicate with an interface based on a build alias. How about we just tell users to forget about the `dune pkg lock` command? If there's a new way to enable package management, as is alluded to by this statement:

> Our idea is that Dune, if instructed by (pkg enabled) to do so, would take the project definitions and automatically create a lock directory in the build folder and use this lock directory for constructing the package rules

Then it's enough to just point new users at it?

> Maybe I have misunderstood, because I have understood your comment to meant that the planned workflow was to always run the dune pkg lock and copy action. Thus the lockfile targets would always be generated by copying from the source directory.

Yeah, that's not what I had in mind. I concur with having the copying step only if the lock directory exists in the source. If it does not exist in the source, the new workflow should take priority.

> I can certainly take the copy action out of this PR into a separate PR but the copy action seemed like such a small part of this PR that fit the structure of the code rather naturally (https://github.com/ocaml/dune/commit/cbb9b0bffde17943b112997fe3153092d2c156a5) that I didn't see the point of pulling it out into a separate PR. But I can definitely do that and rebase this PR on top of it afterwards.

If it's a small and painless change, I would only be relieved. At some point, I did try making the lock directory be copied to the build directory first, but I ran into some rather painful rule loading cycles. Let's hope none of that is relevant anymore.

Leonidas-from-XIV:
> My suggestion would be not do that. It could probably work that way, but I think it's a lot harder than it looks and there's no obvious benefit to this. If only because dune build @pkg-lock will not work while dune is already running in watch mode. Moreover, I think dune pkg lock takes arguments that would be hard to replicate with an interface based on a build alias. How about we just tell users to forget about the dune pkg lock command?

Oh, maybe I should have specified that I don't expect users to use neither `dune pkg lock` nor `dune build @pkg-lock`. These are solely to force the solver to come up with a lock dir if it doesn't exist in a similar way that `dune build @pkg-install` can be used to manually force building of packages. If Dune is running in watch mode that instance will generate the lock files when necessary (just like any other target).

For the user `(pkg enabled)` will create the lock dir in the build dir (by copying if there is a lock dir in the source or by running the solver if there isn't), read the rules from it and build the dependencies and the project in one single `dune build` step.

src/dune_rules/lock_rules.ml
```diff
@@ -41,46 +258,390 @@ let setup_copy_rules ~dir:target ~lock_dir =
...
+  solver_env
+    ~solver_env_from_context
+    ~solver_env_from_current_system
+    ~unset_solver_vars_from_context
+;;
+
+let setup_lock_rules ~dir ~lock_dir : Gen_rules.result =
+  let target = Path.Build.append_local dir lock_dir in
+  let rules =
+    let* packages =
```
rgrinberg: All of this work needs to be put in the action builder for the rule.

src/dune_rules/lock_rules.ml
```diff
@@ -41,46 +258,390 @@ let setup_copy_rules ~dir:target ~lock_dir =
...
+      ~lock_dir
+      ~constraints
+      ~repos
+      ~env
+      ~selected_depopts
+      ~pins
+      ~version_preference
+  in
+  let rule = Rule.make ~targets build in
+  Rules.of_rules [ rule ]
```
rgrinberg: Could you use the add_rule based API? It is usually our first choice, and I don't see why we're switching to the lower level API here.

src/dune_rules/lock_rules.ml
Leonidas-from-XIV: How to specify the dependencies so Dune will know to rerun it when it changes? e.g. if someone updates the directory with the opam repo, then that code needs to run again to create a new `Spec.t` to compare it with the old spec.

src/dune_rules/lock_rules.ml
Leonidas-from-XIV: The reason it is done this way is because I had the `collect_unit` thing here before but in #12534 you used `Rules.of_rules [ rule ]` so I assumed this was the way to go.

src/dune_rules/lock_rules.ml
rgrinberg: Everything that you read needs to be reflected by either:

1. The digest of your action
2. The dependencies of your action

So you can either embed the repo info into the action digest (not great), or as a dependency (better).

src/dune_rules/lock_rules.ml
rgrinberg: Before there was only one rule to generate so I think it doesn't really matter which API to use. Now that we have more than one, I'd prefer the add_rule API - because you can use one `collect_unit` to collect all the rules generated rather than just one. It's not critical here and it can be fixed after the PR. Just letting you know why we _usually_ use `add_rule`.

Leonidas-from-XIV:
Obsoleted by #12653, closing.

PR #11788, opened 2025-05-11
chore(ci): bump ocaml version 5.2 -> 5.3

Alizter:
We update the CI to use OCaml 5.3. The main change required here is the use of `-x c++ -std=c++11` when using the C compiler as a C++ compiler.

OCaml 5.3 uses some more modern C features which means `clang` needed to be instructed to use these instead. This required passing some extra flags to the C compiler which we already do via `:standard`. In order to use that, we bumped our `dune lang` version to `3` in some tests and replaced some bare `%{ocaml-config:c_compiler}` invocations with `%{cc}` and `%{cxx}` where appropriate.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -30,7 +30,7 @@ jobs:
         # We don't run tests on all versions of the Windows environment and on
         # 4.02.x and 4.07.x in other environments
         ocaml-compiler:
-          - 5.2.x
+          - 5.3.x
         include:
           # OCaml trunk:
           - ocaml-compiler: ocaml-variants.5.4.0+trunk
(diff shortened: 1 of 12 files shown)


Alizter:
I've tried for like a week straight to fix the 5.3 CI. There are two issues left which I don't know how to fix:

1. We hand compile C++ in several spots which on macOS uses Clang and fails unpredictably and verbosely due to the `-std=c++` and `-x c++` requirements. I'm still uncertain if we have the same problem in our foreign rules. Same doesn't happen on nix (on mac) so not certain how to reproduce.

2. On Linux, we somehow manage to get into a dead-lock with the FUSE file system. (WARNING: don't do this next part) You can reproduce this by running `./dune.exe build @runtest` which will hang and stop responding to `SIGKILL`. This requires a restart to kill properly from what I can tell.

For 2 I am going to have to bisect the entire testsuite so I can find which test is causing these issues. This will take time.

Alizter:
@hhugo The `jsoo/inline-tests.t` test appears to have a different behaviour in OCaml 5.3 on macos-latest. With the release profile it seems to also run the bytecode tests. There is no difference between the outputs of the dev build and `--profile release`.

I will disable this test on macos and add a TODO in the meantime.

rgrinberg (approved):

PR #11795, opened 2025-05-11
fix(pkg): do include files from unreachable packages

rgrinberg:
The solver correctly removes unnecessary `{post}` dependencies, but keeps their artifacts around. This is because we do the reachability analysis after computing the files we copy the lock directory. Reversing the two steps fixes the bug.

One of @gridbugs or @Leonidas-from-XIV please


rgrinberg:
ping @gridbugs

gridbugs (approved):
Commit message should probably be "do _not_ include files..."

PR #11796, opened 2025-05-12
test: show OCaml compiler error snippet difference with `copy_files#`

anmonteiro:
It sounds like the OCaml compiler has some trouble finding the warning snippet to show for files that have been copied with `copy_files#` and therefore have a line directive.


anmonteiro:
this seems to be an OCaml bug, opened issue upstream: https://github.com/ocaml/ocaml/issues/14027

rgrinberg:
Not much dune can do here.

Issue #11620, opened 2025-04-08
Extension to the `documentation` stanza to support new odoc 3 features


panglesd:
Hello dune people!

Since the [release of odoc 3](https://github.com/ocaml/odoc/releases/tag/3.0.0), the convention for building docs for opam-installed packages has been clarified and extended[^1]. Here are the bits to know:

[^1]: Let me stress that this has nothing to do with the rules for _building_ documentation, it only has to do with installing the right files at the right places so that other drivers of odoc know how to build the documentation for the installed package!

1. Mld pages should still be installed in `<switch-root>/doc/<pkgname>/odoc-pages/`. However, the pages can now be _nested inside subdirectories_. The built doc will be faithful to this hierarchy (see [this](https://ocaml.github.io/odoc/odoc/driver.html#the---parent-id-arguments) for details).
2. The hierarchy in  `<switch-root>/doc/<pkgname>/odoc-pages/` may now include _non-mld_ files, which are then considered assets. See for instance [this line](https://github.com/ocaml/odoc/blob/master/doc/dune#L55) to install an image asset in odoc's doc. (See [this](https://ocaml.github.io/odoc/odoc/odoc_for_authors.html#quickstart))
3. The docs may refer to additional packages and libraries than the actual package's dependencies, for instance to refer to the docs of a package using it (eg, odoc referencing odig's docs). This is specified in a file (in sexp syntax) that has to be installed as `<switch-root>/doc/<pkgname>/odoc-config.sexp`. See [this](https://ocaml.github.io/odoc/odoc/driver.html#link-time-dependencies) and [this](https://ocaml.github.io/odoc/odoc/odoc_for_authors.html#quickstart) and [this](https://ocaml.github.io/odoc/odoc/odoc_for_authors.html#config-file) for more detail, and the [file for odoc](https://github.com/ocaml/odoc/blob/master/doc/odoc-config.sexp) and the [rule to install it](https://github.com/ocaml/odoc/blob/master/doc/dune#L54) for an example.
4. In order for ocaml.org's doc CI to be able to install all needed packages, the opam file of a package should also include all packages it depends upon, under the `x-extra-doc-deps` field name. See [this](https://ocaml.github.io/odoc/odoc/odoc_for_authors.html#config-file) for details. For instance, here is [`odoc` specifying its doc dependencies](https://github.com/ocaml/odoc/blob/master/odoc.opam#L66-L70).

[...]

PR #11800, opened 2025-05-13
Support odoc 3's hierarchy when installing files

panglesd:
Hello!

This PR is about adding dune support for odoc 3's new hierarchy. To recall briefly, odoc 3 can be driven to generate a hierarchy of mld (and asset) files. When building docs for installed packages (as odig or ocaml.org do), the odoc driver gets the intended hierarchy from the hierarchy of files and folders in `<switch root>/doc/odoc-pages/`. See #11620.

To support this, `dune` needs to:
1. Provide a way for the author to specify the hierarchy,
2. Install the files according to this hierarchy,
3. Drive odoc according to this hierarchy.

This PR tackles the first two points, through the introduction of the `files` field in the `documentation` stanza. It is the same field as the one in the `install` stanza.

The third point is not tackled at all, but the user is warned that some files are ignored until the support for 3. is added.

I hope the added test and documentation are clear enough!

It may make sense to also add other fields from `install`, such as `dirs` and `source_trees`, but I'd prefer to start with a simpler PR, which contains most of the value anyway.

Also, I'd like to extend `dune init proj` to contain a `doc/` folder, with a toplevel `dune` adding the whole `doc/` as documentation. I believe this would encourage library authors to write mld files, such introduction/examples/tutorials more. What do you think?

Thanks!

(Note that this is independent from #11716.)


panglesd:
Small :ping_pong: @rgrinberg @Leonidas-from-XIV as this PR is ready for review. (Sorry if I'm pinging the wrong person!)

I think I'll make a more practical description of what is in this PR, than my intro message:

This PR adds a `(files ...)` field to the `(documentation ...)` stanza. This fields has the same syntax as the `files` field from the [`install` stanza](https://dune.readthedocs.io/en/stable/reference/dune/install.html).

The listed files are installed relative to `doc/<pkgname>/odoc-pages/`. If they have mld extensions, they are also considered as pages by dune when building the docs (although, `dune build @doc` will ignore the files that are in subfolders for now as it does not support that odoc 3 feature yet).

A simple example of usage of the new `file` field is:
```
(documentation
 (files
  (glob_files_rec
   (doc/* with_prefix .))))
```
which installs all files from the `doc` folder in `doc/<pkgname>/odoc-pages/`, respecting the relative hierarchy.

rgrinberg:
I'll try reviewing this week

src/dune_rules/odoc.ml
```diff
@@ -203,20 +203,20 @@ let odoc_ext = ".odoc"
 module Mld : sig
   type t
 
-  val create : Path.Build.t -> t
+  val create : Path.Build.t * string -> t
   val odoc_file : doc_dir:Path.Build.t -> t -> Path.Build.t
   val odoc_input : t -> Path.Build.t
 end = struct
-  type t = Path.Build.t
+  type t = Path.Build.t * string
```
gridbugs: I think this would be clearer as a record type so the `string` can be given a name.

src/dune_rules/odoc.ml
```diff
@@ -663,21 +662,36 @@ let check_mlds_no_dupes ~pkg ~mlds =
...
 
+let mlds sctx pkg =
+  Packages.mlds sctx pkg
+  >>| List.filter_map ~f:(fun mld ->
+    match Path.Local.explode mld.Doc_sources.in_doc with
+    | [ name ] when Filename.extension name = ".mld" -> Some (mld.path, name)
+    | _ ->
+      User_warning.emit
+        [ Pp.textf
+            "Dune does not support yet building documentation for assets, and mlds in a \
```
gridbugs:
```suggestion
            "Dune does not yet support building documentation for assets, and mlds in a \
```

src/dune_rules/odoc_new.ml
```diff
@@ -1255,15 +1255,27 @@ let ext_package_mlds (ctx : Context.t) (pkg : Package.Name.t) =
...
-  then Packages.mlds sctx pkg >>| List.map ~f:Path.build
+  then
+    Packages.mlds sctx pkg
+    >>| List.filter_map ~f:(fun mld ->
+      match Path.Local.explode mld.Doc_sources.in_doc with
+      | [ name ] when Filename.extension name = ".mld" -> Some (Path.build mld.path, name)
+      | _ ->
+        User_warning.emit
+          [ Pp.textf
+              "Dune does not support yet building documentation for assets, and mlds in \
```
gridbugs:
```suggestion
              "Dune does not yet support building documentation for assets, and mlds in \
```

gridbugs (approved):
Couple of minor suggestions but generally looks good to me.

src/dune_rules/odoc.ml
panglesd: Done! (as well as for `odoc_new.ml`: a plain `string` argument to a function is now named.)

doc/documentation.rst
```diff
@@ -91,6 +91,19 @@ excluding ``wip.mld``, in the current directory:
...
+.. code-block:: dune
+
+   (documentation
+    (files
+     (glob_files_rec
+      (doc/* with_prefix .))))
+
+All files in the ``doc/`` folder will be attached to the inferred package. The
+hierarchy between them will be preserved, relative to ``doc/`` considered as the
+root. (Note that dune supports installing those files, but not building the
```
Leonidas-from-XIV: Maybe use the `note` block, that's kind of nicer than an entire sentence in parenthesis.

doc/reference/dune/documentation.rst
```diff
@@ -20,4 +20,11 @@ Where ``<optional-fields>`` are:
   basenames attached to the package, where ``:standard`` refers to all the
   ``.mld`` files in the stanza's directory.
 
+- ``(files <arg>)``: the ``files`` field accepts the same arguments as the one
+  from the :ref:`install stanza <including-files-install-stanza>`. It allows to
+  install mlds as well as asset files, and specify where they are in the
```
Leonidas-from-XIV: `mlds` should probably be in monospace.

src/dune_rules/dir_contents.ml
```diff
@@ -376,14 +347,19 @@ end = struct
              Memo.lazy_ (fun () ->
                stanzas >>| Coq_sources.of_dir ~dir ~dirs ~include_subdirs)
            in
+           let mlds =
+             Memo.lazy_ (fun () ->
+               let* expander = Super_context.expander sctx ~dir in
+               Doc_sources.build_mlds_map dune_file ~dir ~files expander)
```
Leonidas-from-XIV: Can this and the code above deduplicated? It seems like for any change in one of the anyonymous functions one would want the same change in the other one which is a good argument to pull it into a separate function, even despite it only being two lines at the moment.

src/dune_rules/doc_sources.ml
```diff
@@ -0,0 +1,72 @@
...
+    { path; in_doc })
+  |> Filename.Map.values
+;;
+
+let build_mlds_map stanzas ~dir ~files expander =
+  let mlds =
+    Memo.lazy_ (fun () ->
+      Filename.Set.fold files ~init:Filename.Map.empty ~f:(fun fn acc ->
+        (* TODO this doesn't handle [.foo.mld] correctly *)
+        match String.lsplit2 fn ~on:'.' with
```
Leonidas-from-XIV: Should be easy to fix with `String.rsplit2`, but also `Filename.split_extension` could be an option.

src/dune_rules/doc_sources.ml
```diff
@@ -0,0 +1,72 @@
...
+  >>= Memo.parallel_map ~f:(fun (doc : Documentation.t) ->
+    let* from_mld_files = from_mld_files mlds doc dir in
+    let+ from_files =
+      let expand = Expander.No_deps.expand expander ~mode:Single in
+      let+ file_bindings =
+        Install_entry.File.to_file_bindings_expanded doc.files ~expand ~dir
+      in
+      of_file_bindings file_bindings
+    in
+    let mlds = List.concat [ from_mld_files; from_files ] in
```
Leonidas-from-XIV: What happens if there's duplicates?

src/dune_rules/doc_sources.mli
```diff
@@ -0,0 +1,13 @@
+open Import
+
+type mld =
+  { path : Path.Build.t
+  ; in_doc : Path.Local.t
+  }
+
+val build_mlds_map
```
Leonidas-from-XIV, line 10: Could you add a documentation comment? Best on both items but I think this function would really benefit from having a high level description on what it does.

src/dune_rules/install_rules.ml
```diff
@@ -491,9 +491,9 @@ end = struct
           >>| List.rev_map ~f:(fun mld ->
             Install.Entry.make
               ~kind:`File
-              ~dst:(sprintf "odoc-pages/%s" (Path.Build.basename mld))
+              ~dst:(sprintf "odoc-pages/%s" (Path.Local.to_string mld.Doc_sources.in_doc))
```
Leonidas-from-XIV: Maybe there should be a function that calculates this value instead of getting it from dereferencing multiple levels deep. Kind of like the law of demeter, but for FP.

src/dune_rules/odoc.ml
```diff
@@ -203,20 +203,22 @@ let odoc_ext = ".odoc"
...
 
-  val create : Path.Build.t -> t
+  val create : path:Path.Build.t -> name:string -> t
   val odoc_file : doc_dir:Path.Build.t -> t -> Path.Build.t
   val odoc_input : t -> Path.Build.t
 end = struct
-  type t = Path.Build.t
+  type t =
+    { path : Path.Build.t
+    ; name : string
```
Leonidas-from-XIV: What is the `name` of an MLD file? Is it the basename of the file? Something else?

src/dune_rules/odoc.ml
```diff
@@ -203,20 +203,22 @@ let odoc_ext = ".odoc"
...
 
-  let odoc_file ~doc_dir t =
-    let t = Filename.remove_extension (Path.Build.basename t) in
-    Path.Build.relative doc_dir (sprintf "page-%s%s" t odoc_ext)
+  let odoc_file ~doc_dir { name; _ } =
+    Path.Build.relative doc_dir (sprintf "page-%s%s" name odoc_ext)
   ;;
 
-  let odoc_input t = t
+  let odoc_input { path; _ } = path
```
Leonidas-from-XIV, line 225: Possibly nicer to do `name=_` to allow the compiler to point out unused values, but admiteddly it's quite nitpicky. Your judgement.

src/dune_rules/odoc.ml
```diff
@@ -663,25 +664,41 @@ let check_mlds_no_dupes ~pkg ~mlds =
       [ Pp.textf
           "Package %s has two mld's with the same basename %s, %s"
           (Package.Name.to_string pkg)
-          (Path.to_string_maybe_quoted (Path.build p1))
-          (Path.to_string_maybe_quoted (Path.build p2))
+          (Path.to_string_maybe_quoted (Path.build (fst p1)))
```
Leonidas-from-XIV: Better pattern match above on `Error, (_, (p1, _mld1), (p2, _mld2))` especially as `p` probably stands for "path" which it isn't anymore.

src/dune_rules/odoc.ml
```diff
@@ -663,25 +664,41 @@ let check_mlds_no_dupes ~pkg ~mlds =
...
+          (Path.to_string_maybe_quoted (Path.build (fst p2)))
       ]
 ;;
 
+let mlds sctx pkg =
+  Packages.mlds sctx pkg
+  >>| List.filter_map ~f:(fun mld ->
+    match Path.Local.explode mld.Doc_sources.in_doc with
+    | [ name ] when Filename.extension name = ".mld" ->
```
Leonidas-from-XIV: `String.equal`? Maybe.

src/dune_rules/odoc.ml
```diff
@@ -663,25 +664,41 @@ let check_mlds_no_dupes ~pkg ~mlds =
...
+      Some (mld.path, Filename.remove_extension name)
+    | _ ->
+      User_warning.emit
+        [ Pp.textf
+            "Dune does not yet support building documentation for assets, and mlds in a \
+             non-flat hierarchy. Ignoring %s."
+            (Path.Local.to_string mld.in_doc)
+        ];
```
Leonidas-from-XIV: Given the warning can be emitted a lot of times (as shown in the test), maybe a better way would be to collect these filenames in a set and emit one warning with all the files from the set if the set is not empty. That way the output isn't flooded with duplicate warnings.

src/dune_rules/odoc_new.ml
```diff
@@ -1255,15 +1253,28 @@ let ext_package_mlds (ctx : Context.t) (pkg : Package.Name.t) =
 let pkg_mlds sctx pkg =
   let* pkgs = Dune_load.packages () in
   if Package.Name.Map.mem pkgs pkg
-  then Packages.mlds sctx pkg >>| List.map ~f:Path.build
+  then
+    Packages.mlds sctx pkg
+    >>| List.filter_map ~f:(fun mld ->
+      match Path.Local.explode mld.Doc_sources.in_doc with
+      | [ name ] when Filename.extension name = ".mld" ->
```
Leonidas-from-XIV: `String.equal` potentially.

src/dune_rules/odoc_new.ml
```diff
@@ -1255,15 +1253,28 @@ let ext_package_mlds (ctx : Context.t) (pkg : Package.Name.t) =
...
+        Some (Path.build mld.path, Filename.remove_extension name)
+      | _ ->
+        User_warning.emit
```
Leonidas-from-XIV: As above. Maybe even sensible to put it in a common function so the code does not have to be duplicated.

src/dune_rules/odoc_new.ml
```diff
@@ -1272,8 +1283,8 @@ let check_mlds_no_dupes ~pkg ~mlds =
       [ Pp.textf
           "Package %s has two mld's with the same basename %s, %s"
           (Package.Name.to_string pkg)
-          (Path.to_string_maybe_quoted p1)
-          (Path.to_string_maybe_quoted p2)
+          (Path.to_string_maybe_quoted (fst p1))
```
Leonidas-from-XIV: Likewise here, better pattern match in the `Error` constructor.

src/dune_rules/odoc_new.ml
```diff
@@ -1284,12 +1295,12 @@ let pkg_artifacts sctx index pkg =
     check_mlds_no_dupes ~pkg ~mlds
   in
   let artifacts =
-    let mlds_noindex = String.Map.filteri ~f:(fun i _ -> i <> "index") mlds_map in
+    let mlds_noindex = String.Map.filteri ~f:(fun i _ -> i <> "index.mld") mlds_map in
```
Leonidas-from-XIV: `not (String.equal "index.mld i)` possibly?

test/blackbox-tests/test-cases/install-mld-pages/doc-root.t/doc/examples/example1/index.mld
```diff
@@ -0,0 +1 @@
+{0 Example 1}
```
Leonidas-from-XIV: Maybe add a trailing newline as a lot of editors do it by default to prevent an unrelated change in the future. Same for the other files.

test/blackbox-tests/test-cases/install-mld-pages/doc-root.t/run.t
```diff
@@ -0,0 +1,96 @@
+A usefull pattern for documentation is to have a folder that contains only the
```
Leonidas-from-XIV:
```suggestion
A useful pattern for documentation is to have a folder that contains only the
```

test/blackbox-tests/test-cases/install-mld-pages/doc-root.t/run.t
```diff
@@ -0,0 +1,96 @@
...
+  $ cd ../../../../../../
+We store the original hierarchy of the doc/ folder
+  $ cd doc
+  $ ls -R > ../source_hierarchy
+  $ cd ../
+We compare both, they should be equal
+  $ diff source_hierarchy installed_hierarchy
+
+For the curious reader, here is the hierarchy:
+  $ cat source_hierarchy | grep -v ".:" # we remove ".:" for MacOS/Linux compatibility
```
Leonidas-from-XIV: The Unix neckbeard in me has to point out the useless use of cat :-)

test/blackbox-tests/test-cases/install-mld-pages/files-in-documentation-stanza.t/run.t
```diff
@@ -0,0 +1,111 @@
+The files field of the documentation stanza allows to control the hierarchy
+of documentation files
+
+  $ cat doc/dune
```
Leonidas-from-XIV: I would avoid outputting the file. It's either in the source tree or if you want to have it in the test then write it in the test, but having it in both places is just duplication. I'd keep it in the source tree.

Leonidas-from-XIV (commented):
Very nice PR, don't let the amount of comments surprise you, it's mostly things I came across where I think something could be done better or could use an explanation.

rgrinberg (approved):
LGTM. I made some minor tweaks and there's probably more clean up that could be done, but it should be good to go.

rgrinberg:
CI seems to be failing. @panglesd could you have a look?

doc/documentation.rst
panglesd: Fixed in 758e0e703

doc/reference/dune/documentation.rst
panglesd: Fixed in 758e0e703

src/dune_rules/dir_contents.ml
panglesd: Agree!

Fixed in 440fabf21

src/dune_rules/doc_sources.ml
panglesd: Actually, this code is just moved from before...

But yes, why not fix it? Fixed in 6e1052bfe

src/dune_rules/doc_sources.ml
panglesd: Since bcb57b4ed, duplicates are deduplicated. Good catch!

src/dune_rules/doc_sources.mli
panglesd: Documentation added in 1e0c270ef

src/dune_rules/install_rules.ml
panglesd: Not sure I understood, but I tried something in 721cec5ec. Let me know if that is what you had in mind!

src/dune_rules/odoc.ml
panglesd: Documented in 0fc8dc553

src/dune_rules/odoc.ml
panglesd: Since you left it to my judgement: I don't see how the addition of a field (without modifying the `path` field name) would imply a modification of this function (which is just an accessor for `path`), so I left the wildcard.

src/dune_rules/odoc.ml
panglesd: Thanks! Fixed in 8fa37bde4

src/dune_rules/odoc.ml
panglesd: Thanks! Fixed in 8fa37bde4

src/dune_rules/odoc_new.ml
panglesd: Thanks! Fixed in 8fa37bde4

src/dune_rules/odoc_new.ml
panglesd: Thanks! Fixed in 8fa37bde4

test/blackbox-tests/test-cases/install-mld-pages/doc-root.t/doc/examples/example1/index.mld
panglesd: Done in 4b5584964

test/blackbox-tests/test-cases/install-mld-pages/doc-root.t/run.t
panglesd: Unix neckbeard has been heard in 53907d4ee!

test/blackbox-tests/test-cases/install-mld-pages/files-in-documentation-stanza.t/run.t
panglesd: I outputted the file because the explanation just after seemed more understandable with the file just above, and the duplication is kept synced automatically so that felt okay, but I removed it! (and mention to the reader that it is an important file to read!)

src/dune_rules/odoc_new.ml
panglesd: Fixed in cfbb99a85

src/dune_rules/odoc.ml
panglesd: Fixed in f55a9ff9b

src/dune_rules/odoc_new.ml
panglesd: The two versions are very subtly different: one handles `Path.t` while the other handles `Path.Build.t`. I'm not sure how to deduplicate the functions in an effective way...

src/dune_rules/doc_sources.ml
Leonidas-from-XIV: Ah sorry, I didn't notice it was moved, but I do appreciate fixing it. I guess the code might have been there before `String.rsplit2` existed.

src/dune_rules/install_rules.ml
Leonidas-from-XIV: Oh, I think I misunderstood this bit of the code, and somehow mistook the module ascription (or whatever this thing is called) for accessing a record in a record. A bit silly looking back on the code now, as it clearly isn't because of the capital letter. But Rudi already changed the code to attach the type information to the signature so the issue is solved anyway.

src/dune_rules/odoc.ml
```diff
@@ -668,20 +673,46 @@ let check_mlds_no_dupes ~pkg ~mlds =
       ]
 ;;
 
+let report_warnings warnings =
+  match warnings with
+  | [] -> ()
+  | _ :: _ ->
+    let l =
+      warnings
+      |> List.map ~f:(fun (mld : Doc_sources.mld) -> Path.Local.to_string mld.in_doc)
```
Leonidas-from-XIV, line 682: I am not sure this can ever happen (or whether there's something that fixes the ordering in the list) but maybe it could be useful to sort the strings so the files are always reported in the same order.

Makes it a bit easier to use in tests as well.

src/dune_rules/gen_rules.ml
```diff
@@ -163,6 +163,8 @@ end = struct
     | Install_conf.T i ->
       let+ () = install_stanza_rules ~ctx_dir ~expander i in
       empty_none
+    (* TODO: What does this part of the code do? Why does Documentation.t has
+       not the same kind of install_stanza_rules? *)
```
Leonidas-from-XIV: This looks like a leftover bit of code or maybe something that was unclear?

test/blackbox-tests/test-cases/install-mld-pages/files-in-documentation-stanza.t/doc/examples/example1/index.mld
```diff
@@ -0,0 +1 @@
+{0 Example 1}
```
Leonidas-from-XIV: Nitpick (along with some of the other files): Maybe worth adding a newline here to prevent the next person opening this file saving it with a newline.

Leonidas-from-XIV (commented):
Thanks for making the adjustments, I think it is nicer now. A final few things that I have noticed going through the PR with the changes applied, but otherwise looks good.

src/dune_rules/odoc.ml
panglesd: Good point! I sorted by alphabetical order in d07de8f9b

test/blackbox-tests/test-cases/install-mld-pages/files-in-documentation-stanza.t/doc/examples/example1/index.mld
panglesd: I though I had done those files, thanks for repeating! Fixed in eeac346bc

src/dune_rules/gen_rules.ml
panglesd: Yes, something that was unclear... I left this TODO to be "sure" to remember about sorting this out, so fortunately you noticed it.

I can't really understand what is this part of the code about... Removing it has [a very small effect](https://github.com/ocaml/dune/commit/c1d0ae10c3a32154cbbb1ab93d71559f9e5d655a).

I was surprised not to find anything about documentation stanza in this pattern match, since there is something about install stanzas, and documentation stanzas in particular generate install rules?

Leonidas-from-XIV (approved):
We can discuss the TODO but the test pass, the code looks good to me and I am happy to merge it; so I'm clicking the "approve" button :-)

PR #11801, opened 2025-05-14
Fix #11045

### LINKED ISSUE #11045: Regression in #11028 when using `dune subst` on a project with an opam file

btjorge:
Fix #11045 by ignoring error when adding `version: %%VERSION_NUM%%` to an opam file.

The third commit is more of a new feature : it default the version to the one specified in dune-project, if any, when there is no vcs.

I think this fix is the way to go because the right version field will often already appear in the file anyway


rgrinberg:
Thanks for the contribution

Issue #7811, opened 2023-05-26
Vendoring of foreign source trees with leading underscores


Lupus:
Following up on https://github.com/ocaml/dune/issues/4795, dune is still lacking convenient way to manage vendored foreign source trees that might include underscores.

I'm currently vendoring a bunch of Rust crates into my OCaml project and use the following stanza to depend on it in the rule, invoking cargo: `(source_tree ../vendor)`. After adding new dependencies, cargo started to complain about missing files, after quick research (good thing I know dune has some oddities regarding underscores!) I found the issue linked above and applied a fix in my dune file:

```dune
(vendored_dirs vendor)
(subdir vendor/futures-core/src/task/ (dirs :standard __internal))
```

This behavior with underscores is certainly confusing, and lack of decent workaround is disappointing.

nojb:
Indeed, Dune by default ignores any directory starting with `_`. I don't remember if there is a technical reason to do so. One directory we want to ignore for sure is `_build`, but perhaps we don't need to ignore every other directory starting with `_` by default (althought I bet there are builds in the wild that depend on this behaviour).

rgrinberg:
@Lupus how would you improve the current solution for changing the directory mask?

Lupus:
If there is no need to ignore anything else than `_build`, I would probably just stop doing that. It's a higly confusing impilit behavior.

rgrinberg:
What about directories that stard with a `.`? We also ignore those by default.

Lupus:
I've also got bitten by that when tried to tell dune that I need my `.cargo` dir in the build environment as a dependency. It was just silently ignored when I had it explicitly written in a rule, no errors, happy debugging. Maybe there should be some stanza to specify which source trees should include all that's found inside verbatim without any filtering. `data_only_dir` might be that stanza actually, as cargo sources are just data for dune, as well as cargo configs in `.cargo` folder.

nojb:
Another possibility would be to only change the default glob for subdirectories of those specified in `vendored_dirs`.

Lupus:
Right now I use `vedored_dirs` to mark my cargo vendor folder, copied that from some of the templates for cargo integration available online. That template also uses `source_tree` stanza to depend on the whole `vendor` folder in rules, invoking cargo. So yeah, `vendored_dirs` not ignoring some files because of `.` or `_` would work for me, not sure if it might break something, as `vendored_dirs` are also used for vendored dune projects.

rgrinberg:
[...]

PR #11819, opened 2025-05-18
RFC: DUNE_IGNORE variable to ignore some source files

gasche:
The dune build system for github/ocaml is very painful to use for everyday compiler development. I would like to make it less painful to use, and I think a reasonable approach would be to implement a couple simple tweaks in dune to make it more pleasant. (The compiler is an unusual project building in a special environment, so maybe it deserves a couple specific tweaks in dune).

This PR is not inteded to be merged as-is, but more a starting point for discussion: I'm willing to do a bit of work to make dune less painful, what changes would you be willing to accept?

In the PR I add a DUNE_IGNORE environment variable to ask it to ignore some files in the source directories -- directed by a glob pattern.

Before:

```
[gasche@framawork github-trunk (manual-extensions-minor-change)]$ dune build
Error: Multiple rules generated for _build/default/stdlib/std_exit.cmo:
- stdlib/dune:15
- file present in source tree
-> required by alias default
Hint: rm -f stdlib/std_exit.cmo
Error: Multiple rules generated for _build/default/otherlibs/str/str.cma:
- otherlibs/str/dune:15
- file present in source tree
-> required by alias default
Hint: rm -f otherlibs/str/str.cma
Error: Multiple rules generated for _build/default/otherlibs/unix/unix.cma:
- otherlibs/unix/dune:15
- file present in source tree
-> required by alias default
Hint: rm -f otherlibs/unix/unix.cma
```

After:

```
[gasche@framawork github-trunk (manual-extensions-minor-change)]$ DUNE_IGNORE="*.{cmo,cma,cmx,cmxa,o,a}" ~/Prog/dune/_build/install/default/bin/dune build
# no errors
```

Much more pleasant!

## Context

Currently the only way to get Merlin support for the compiler sources is to use its dune build system. See [github/ocaml/HACKING.adoc#using-merlin](https://github.com/ocaml/ocaml/blob/0728f6af2aae32a97c2a7a1214c25736a26a479b/HACKING.adoc#using-merlin).

One major limitation in my experience is that, because the dune is incomplete, we need to keep building using the makefile for development, testing, etc. So currently the workflow is as follows:

- start hacking on the compiler using `make`
- suddenly you want Merlin type feedback, you need to `make clean; dune build @libs`
- continue hacking on the compiler using `make`
- oh no, you changed a .mli file somewhere and the types in _build are all wrong! you need to `make clean; dune build @libs`

The need to run `make clean` in between two dune invocations is a massive pain. This is mostly solved by the proposed feature.

(Note: it's important to _ignore_ the build outputs that are present in source directories, and not, for example, to use them as valid outputs instead of the other rules to build them, because they are not built by the same compiler as Dune. Dune uses the compiler from the opam switch, while Make uses the bootstrap or in-development compilers.)

## What would be an acceptable interface?

I understand that such "ignore rules" are not well-regarded by Dune designers. I wish there were available, but they could be maked by a prefix to discourage their usage, or require a certain explicit marker in `dune_project` (for example a `special_hacks_for_the_compiler` stanza), or be hidden within an environment variable to discourage usage ever more. What would be acceptable to you?

## What would be an acceptable implementation?

I can see that other environment variables are surrounded by much more ceremony to guarantee that they are recorded as build dependencies and do not break the memoization machinery. I'm happy to try to implement this for DUNE_IGNORE as well, if there is a consensus that this is the preferred user interface for this feature. Any other implementation advice is welcome.

### DIFF
--- a/src/source/readdir.ml
+++ b/src/source/readdir.ml
@@ -65,6 +65,11 @@ let is_temp_file fn =
   || String.is_suffix fn ~suffix:"~"
 ;;
 
+let ignore_pattern =
+  match Sys.getenv "DUNE_IGNORE" with
+  | exception _ -> None
+  | pat -> Some (Glob.of_string pat)
+
 let of_source_path_impl path =
   Fs_memo.dir_contents (In_source_dir path)
   >>= function
@@ -88,6 +93,9 @@ let of_source_path_impl path =
       |> Memo.parallel_map ~f:(fun (fn, (kind : File_kind.t)) ->
         let path = Path.Source.relative path fn in
         if is_special kind || Path.Source.is_in_build_dir path || is_temp_file fn
+           || (match ignore_pattern with
+                | None -> false
+                | Some glob -> Glob.test glob fn)
         then Memo.return List.Skip
         else
           let+ is_directory, file =


rgrinberg:
The feature request is reasonable, but I'd rather such a glob live the dune file rather than as an environment variable.

We already have a `dirs` stanza for ignoring directory, so it seems like the most intuitive way to support this feature is through a `files` stanza? As a first step, I would add such a stanza, and then think about a way of making it be the default for every directory in the project (perhaps in an ad-hoc way).

See this ticket for more discussion https://github.com/ocaml/dune/issues/7811

gasche:
Thanks! I'm happy to have a look at implementing `(files ...)`, but could there be a way to make this a recursive setting that applies to subdirectories as well? This could also be useful for `(dirs ...)`, even though it is probably less important there.

Ideas:
- `dirs*` and `files*` could be interpreted as recursive/transitive -- applying to the whole filesystem tree reachable from the current point.
- there could be a stanza `subtree`: `(subdir foo (dirs :standard \ bar))` removes `foo/bar`, and `(subtree (dirs :standard \ bar))` removes `**/bar`.
- or maybe via glob patterns, `**/bar`?

(My preference would go to a `subtree` stanza.)

rgrinberg:
I was thinking of something a little simpler, perhaps:

`
(repeat_in_subdirs
 ...)
```

Would evaluate `...` for every sub directory starting from the current directory. The evaluation of such a stanza would indeed be similar to `subdir`.

Your other suggestions seem fine as well. I'd say go for whatever you find simplest to implement. I'd suggest that you look over the implementation of `subdir` to get an idea of the kind of changes each option would take.

nojb:
> I was thinking of something a little simpler, perhaps:

Alternatively, a `(dirs)` or `(files)` stanza could be accepted in `dune-project` and/or `dune-workspace` to set default rules.

gasche:
Note: I have not had any time to look at this again, but this remains a feature that would be very useful for using `dune` in the compiler for Merlin support (in parallel to the existing, un-hygienic Make build). If someone wants to have a look at this, they would be warmly welcome, otherwise I will just wait to be annoyed by it again to give more work cycles in this direction.

gasche:
I haven't had time to revisit this PR, but at the same time it would be very convenient if it (or something like this) would be available in Dune today to work on the compiler with Merlin support.

I wonder if one of the following would be possible?

1. Could this PR maybe we merged as-is (with an environment variable), which is admittedly a hack, and then maybe refined in the future? (I promise to not complain if the environment variable gets removed when `dune-project` support becomes available.)

2. Could someone who is more familiar with Dune development, for whom the addition of new stuff in `dune-project` is easy work, do the work of tweaking the PR to use a `(files <glob>)` stanza in `dune-project` instead of an environment variable?

nojb:
> 2\. Could someone who is more familiar with Dune development, for whom the addition of new stuff in `dune-project` is easy work, do the work of tweaking the PR to use a `(files <glob>)` stanza in `dune-project` instead of an environment variable?

I'll try to give it a shot. But since I haven't actually looked at the code yet, I can't promise something unexpected won't come up :)

gasche:
Thanks! (Indeed it is not clear to me that configuration information is available at the place I tweaked in the code, and I don't know how easy and elegant it would be to propagate configuration information until that point. It may be that the same filtering can be performed somewhere else, but I tried a couple things before finding a place that works as intended.)

nojb:
> > 2. Could someone who is more familiar with Dune development, for whom the addition of new stuff in `dune-project` is easy work, do the work of tweaking the PR to use a `(files <glob>)` stanza in `dune-project` instead of an environment variable?
> 
> I'll try to give it a shot. But since I haven't actually looked at the code yet, I can't promise something unexpected won't come up :)

See #12879 for my attempt.

gasche:
This is superseded by #12879 which has been merged. Note that the current PR makes it easy to ignore files in all subdirectories, rather than asking to tweak each dune file manually, so it rather corresponds to `dune-project` support (maybe a follow-up?). But in fact the compiler-specific patch to use per-directory stanzas is [not that large](https://github.com/ocaml/dune/pull/12879#issuecomment-3620750127), so I already consider my feature request fulfilled.

I felt a bit bad a few days ago to be asking Dune contributors to do more work on my behalf, but I am happy with the result :-) Thanks!

PR #11823, opened 2025-05-20
Remove some unused logic from exec

gridbugs:
This removes some code related to automatically adding a missing ".exe" suffix from a relative path to an executable passed as the program argument to `dune exec ...` if doing so would result in the name of an executable build target. However the condition is seemingly inverted from what is should be; the suffix is only added if the suffix already exists. Since nobody seems to complain about this bug, let's take the opportunity to delete the code altogether.


gridbugs:
@Alizter let me know if I've misunderstood the intention of this code!

Leonidas-from-XIV (commented):
Personally I think it probably makes sense to remove it if it has never worked and nobody has noticed it yet. Given this removing it does not break any single test if we want it we should introduce a test for it at least.

Alizter (approved):

Issue #11824, opened 2025-05-20
Portable lock dirs break depext resolution


Leonidas-from-XIV:
## Expected Behavior

```sh
$ DUNE_CONFIG__PORTABLE_LOCK_DIR=enabled dune pkg lock
$ DUNE_CONFIG__PORTABLE_LOCK_DIR=enabled dune show depext
libgmp-dev
```

## Actual Behavior

```sh
$ DUNE_CONFIG__PORTABLE_LOCK_DIR=enabled dune pkg lock
$ DUNE_CONFIG__PORTABLE_LOCK_DIR=enabled dune show depext

```

The output is empty.

## Reproduction

I've tested it by unpacking `mysql_protocol.3.0.3` (which pulls in `conf-gmp` which has depexts) however I don't think there is anything particularly specific about `mysql_protocol`, it's probably the case with all depexts as evidenced by the regressions of the Dune build in opam-health-check.

- PR with a reproducing test: 

1.
1.
1.

## Specifications

- Version of `dune` (output of `dune --version`): `main`
- Version of `ocaml` (output of `ocamlc --version`): 5.2.1
- Operating system (distribution and version): Fedora 42

## Additional information 

- Link to gist with verbose output (run `dune` with the `--verbose` flag):

gridbugs:
Thanks for the repro @Leonidas-from-XIV! I'll investigate.

gridbugs:
Fixed in https://github.com/ocaml/dune/pull/11828

Leonidas-from-XIV:
The fix has been merged, I'll close this issue in the meantime.

PR #11825, opened 2025-05-20
Reproduction case for #11824

Leonidas-from-XIV:
This PR adds a test to show that portable lockfiles currently don't output the depexts correctly.

I looked into the lockfiles and the depexts are recorded correctly, it's just `show depext` which doesn't resolve them.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/depexts/portable-lockfile.t
+++ b/test/blackbox-tests/test-cases/pkg/depexts/portable-lockfile.t
@@ -1,0 +1,63 @@
+Makes sure that the depext mechanism works with portable lockfiles
+
+  $ . ../helpers.sh
+  $ mkrepo
+  $ add_mock_repo_if_needed
+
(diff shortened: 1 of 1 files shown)


gridbugs:
I understand why this happens. Dune currently only records depexts into lockfiles for platforms that it solves for rather than all the platforms listed in the opam file. I think for depexts it might make sense to record all the metadata from the opam file into the lockfile, since the choice of depexts is often more specific than dependencies. For example for the more obscure linux distributions it's probably fine to solve dependencies without considering `os-distribution` or `os-version`, however the choice of depexts usually requires considering these fields.

I'll explore an alternative way of recording depexts that preserves all the information about depexts in the opam file.

shonfeder:
#11824 shows as completed. Do we still need this repro case, or should that issue be open, or something else?

gridbugs:
The problem this reproduces is now fixed. Closing.

PR #11826, opened 2025-05-21
chore(ci): setup OxCaml

maiste:
This PR adds a separate action workflow for OxCaml. With the future introduction of Parametrized libraries in Dune, it is important to have an action to check this section of the tests.

This action is deactivated when it doesn't detect the OxCaml compiler, preventing from running specialized OxCaml test in a non viable context.


test/blackbox-tests/test-cases/oxcaml/dune
```diff
@@ -0,0 +1,5 @@
+(cram
+ (applies_to :whole_subtree)
+ ; CR maiste: we need a better way to detect OxCaml. It would silentely ignore
+ ; the test when the version changes.
```
maiste, line 4:
```suggestion
```

maiste:
@art-w I have rebased, you should be able to check again.
@rgrinberg what do you think? Would it be OK to include this part? We will open further PRs for the various functionalities.

src/dune_lang/pform.ml
```diff
@@ -672,6 +675,7 @@ module Env = struct
         ; "-verbose", deleted_in ~version:(3, 0) Var.Nothing
         ; "ocaml_bin", No_info Ocaml_bin_dir
         ; "ocaml_version", No_info Ocaml_version
+        ; "oxcaml_supported", No_info Oxcaml_supported
```
rgrinberg: We need a version guard for this new feature. Given that it's probably still experimental, I suggest that you create an "oxcaml" extension in dune and guard this variable behind it.

maiste: Indeed, I need to add the version guard for `3.20)`!

Regarding the extension, as the `oxcaml_supported` already acts as a guard for the feature (and `is_parametrized_library_supported` too for the other functionalities), is it necessary to have an extension in this case? If you don't have `oxcaml` or, for the next coming PRs, you don't support `parametrized libraries`, it will display an error message.
In this case, the variable would be set to `false` if it is not supported. I don't see how adding the `(using oxcaml)` would guard more the feature. Am I missing a point?

maiste:
@rgrinberg I encounter the problem where if I develop a new feature and guard it with the next release version, I can't use it inside of Dune until this version is released. Should I add the guard to the current version deployed version (3.19) or do you think I should look at the solution we discussed to have dune use the `latest` version internally?

src/dune_lang/pform.ml
rgrinberg: The point of `using oxcaml 0.1` is that it allows us to drop the `oxcaml_supported` variable without warning users. If we don't version it this way we'll basically need to keep this variable around forever.

I'm hoping that this variable would be obsoleted by being able to query for specific features in `ocamlc -config` like we did for parameterized libraries. Mostly because such a check would allow people that would work for both oxcaml the fork and the features once they're upstreamed.

rgrinberg:
> Should I add the guard to the current version deployed version (3.19) or do you think I should look at the solution we discussed to have dune use the latest version internally?

Thinking about the problem again, I think this isn't going to work because the issue is our CI. We are building the dune workspace with the version of dune coming from opam (is that right? this needs confirmation). We should just not do that and always use the bootstrapped version of dune. If we do that, no hacks will be necessary and you'll be able to upgrade dune itself to 3.20 as soon as you introduce it.

src/dune_lang/pform.ml
maiste: OK I missed this point. Indeed, it would make it more relevant in such context. I'll rework the PR then. (I might split it into smaller ones to make it clearer).

maiste:
@rgrinberg with the help of @Alizter (thanks again :pray:), I have managed to introduce the extension as you suggested. This PR is ready for another round of review.

src/dune_lang/pform.ml
```diff
@@ -628,6 +634,7 @@ module Env = struct
          ; "ocaml-config", macro Ocaml_config
          ; "env", since ~version:(1, 4) Macro.Env
          ; "coq", macro Coq_config
+         ; "oxcaml", since ~version:(3, 20) Macro.Oxcaml
```
rgrinberg: Shouldn't you be checking this against the oxcaml extension? I think you will need to extend `Since` to be more generic to support that.

maiste: As it is nor possible to check against the extension here (because of dependency cycle problems) the code for checking this was moved to the macro evaluation. It also helps to limit the spreading of OxCaml extension to the rest of the code. It is modelled the same way as what is done for rocq (coq). Hence, this is not necessary to add extra checks here.

test/blackbox-tests/test-cases/oxcaml/dune
```diff
@@ -0,0 +1,3 @@
+(cram
+ (applies_to :whole_subtree)
+ (enabled_if %{oxcaml:supported}))
```
rgrinberg: Why do you need a macro? can't it just be a variable?

test/blackbox-tests/test-cases/oxcaml/local.t
```diff
@@ -0,0 +1,23 @@
+The test ensures we are able to run OxCaml tests for Dune.
+
+  $ cat > dune-project << EOF
```
rgrinberg, line 3: I think your test should demonstrate usage of this variable and the error message that you get when `oxcaml` isn't enabled.

src/dune_lang/pform.ml
rgrinberg: I have no idea how a cycle comes into play here. It seems like it should be trivial to move the definition of the extension to avoid it.

> the code for checking this was moved to the macro evaluation

I don't really understand this.

> It also helps to limit the spreading of OxCaml extension to the rest of the code

At the cost of making the implementation inconsistent and buggy wrt the rest of dune.

> It is modelled the same way as what is done for rocq (coq). Hence, this is not necessary to add extra checks here.

This is not the example we should follow. The coq extension is maintained exclusively by the coq folks and we don't control their quality.

src/dune_rules/ocaml_toolchain.ml
```diff
@@ -188,3 +188,15 @@ let check_fdo_support { version; lib_config = { has_native; _ }; ocaml_config; _
             version_string
         ]
 ;;
+
+let is_oxcaml_supported t =
+  let suffix = "+jst" in
+  let version_string = Ocaml_config.version_string t.ocaml_config in
+  String.is_suffix ~suffix version_string
+;;
```
rgrinberg: How about this is moved to `ocaml/version`? This is where all the compiler information we support exists at the moment.

src/dune_lang/pform.ml
maiste: The problem I encountered is: to use `extensions`, I need to deal with the `Dune_project` module because this is where the information about extensions are stored. However `Dune_project` already depends on `pform` so I can't do anything there without creating a cycle. The idea was to move the check to where all of this would be instantiated, more particularly when we are expanding the values. At this point of the code, we know everything is instantiated. 

You made me realized it doesn't need to be a macro as I can also get the `Dune_project.t` from the `Var` expander.

test/blackbox-tests/test-cases/oxcaml/dune
maiste: Indeed, it can be.

src/dune_lang/pform.ml
maiste: @rgrinberg I have moved it back to the variable and port the check directly inside the expander. Would it be better this way?

test/blackbox-tests/test-cases/oxcaml/local.t
maiste: https://github.com/ocaml/dune/pull/11826/commits/535e83b9098288f72cbbc05c3aa2890db80421da demonstrates the features with and without the extension.

art-w (approved):
Looks good! Thanks :)

rgrinberg:
Some background: before dune 3.0, all percent forms were dynamically resolved during evaluation. This was a very painful situations because it meant that users could write junk in a section of their dune file, never evaluate the said section, and be unaware that their dune file is invalid.

In 3.0, we decided to change our frontend to resolve percent forms during parsing. This way, any invalid  percent form would trigger an error when the dune file is first loaded. This indeed was a much better experience, but we left some escape hatches for backwards compatibility, and we never made it general enough to work with language extensions.

I pushed a commit that moves the validation of this percent form to the parser.  This way, even if you don't hit the branch that evaluates the form, you should still get the error that the dune file is invalid. There might be some polish lacking, but this is the general direction to take.

maiste:
Indeed, I was lacking the context and thought you wanted to have the `expander` fail faster. Thank you for the commit and the explanation, it is clearer to me.

I will extract your commit and split this PR in two to make smaller and more targeted PRs.

maiste:
I have extracted the commit to another branch. The other one must be merged first. Then, I'll rebase this one on `main` to add the specific OxCaml functionalities.

src/dune_lang/oxcaml.ml
```diff
@@ -0,0 +1,14 @@
...
+
+let syntax =
+  Syntax.create
+    ~name:"oxcaml"
+    ~desc:"experimental support for OxCaml"
+    ~experimental:true
+    [ (0, 1), `Since (3, 20) ]
+;;
+
+(* TODO @maiste: move it to somewhere else. *)
```
rgrinberg: It's already in `dune_project.ml`.

rgrinberg (approved):

src/dune_lang/oxcaml.ml
```diff
@@ -0,0 +1,14 @@
...
+(* let extension = *)
+(*   Dune_project.Extension.register syntax (Dune_lang.Decoder.return ((), [])) Dyn.unit *)
+(* ;; *)
```
maiste:
```suggestion
```
Oups, missed it on rebase.

Issue #11820, opened 2025-05-18
Background tasks in Cram tests are not terminated


Willenbrink:
## Expected Behavior
When a cram test terminates, all running background tasks (e.g. started via `&`) should be terminated too.

A reasonable alternative would be to wait for all tasks to terminate. As some background tasks never terminate and that behavior can be replicated by calling `wait` I think the first option is preferable.

## Actual Behavior
Background tasks continue to run and never terminate.

## Reproduction

 

1. Create a cram test that start some server in the background, e.g. 
```
  $ ncat -l 2000 -k -c 'xargs -n1 echo' &
```
1. `dune runtest`
1. The server is still available and responds to e.g. `ncat localhost 2000`

## Specifications

- Version of `dune` (output of `dune --version`): 3.18.2
- Version of `ocaml` (output of `ocamlc --version`): 5.3.0
- Operating system (distribution and version): Fedora

## Workaround

Adding `pkill -P $$` as the last line to a cram test kills all child tasks.

Alizter:
This is definitely a missing feature and should be implemented. We even have our own workaround in our tests to get around this limitation. I'll need to do some research to find the best way to terminate all the children of the shell.

rgrinberg:
I don't think this issue is cram specific. One should be able to reproduce the same with any action. For example, with `(action (bash "sleep 100 &"))`

PR #11827, opened 2025-05-21
test: crams not terminating subprocesses

Alizter:
This PR adds a reproduction case for https://github.com/ocaml/dune/issues/11820 where a subprocess is started in a cram test but becomes orphaned once the cram test terminates.

We then introduce the following fix. We use a more sophisticated `trap` at the start of the underlying shell script in order to iterate through all the child PIDs and attempt to interrupt and then kill them separately.

A future improvement would be to run cram tests entirely as dune actions themselves, that way we can delegate process handling to the engine and have better cross platform support. That will require a lot more work however, so in the mean time we offer this improvement.


src/dune_rules/cram/cram_exec.ml
```diff
@@ -290,16 +290,18 @@ let create_sh_script cram_stanzas ~temp_dir : sh_script Fiber.t =
         ; script = user_shell_code_file
         }
   in
-  fprln oc "trap 'exit 0' EXIT";
+  (* When the main shell script is exited we make sure to clean up any
+     subprocess PIDs as not to orphan them. The first "trap" will kill all
+     processes in the process group and the second "trap" will make sure the main
+     shell process exits gracefully. *)
+  fprln oc {|trap "trap 'exit 0' TERM && kill -- -$$" EXIT|};
```
gridbugs, line 297: I'd like a little more detail in the comment here for future readers, specifically to understand the reason for registering a trap handler _inside_ another trap handler. Here's my guess at why it's written this way:

The `kill -- -$$` sends `TERM` to all processes in the process group. Is the reason for the "inner" trap so that when said `TERM` is received by the main shell process that it exits with 0? I guess there's no easy way to send `TERM` to all processes in the current process's group _except_ for the current process? And if you were to move the inner trap outside of the `EXIT` trap handler (and just do `kill -- -$$` on `EXIT`) then the process could exit `0` in response to an unexpected `TERM` which we don't want.

gridbugs (approved):
Very useful! Minor comment to clarify the nested `trap` but otherwise happy with this.

src/dune_rules/cram/cram_exec.ml
Alizter: Exactly, this was my solution to avoiding killing the main process. If we end up killing the main process, then Dune will complain that the process is was running received a kill signal, what we want instead is to preserve the old `trap` behaviour which was to `exit 0`.

By switching the trapping behaviour just before we kill the process group, we can signal to all the processes in the group that they should exit when encountering `TERM`, which they all promptly do.

rgrinberg (changes requested):
I don't think this is the right fix. We can use the same mechanism for cleanup that we do for regular actions.

rgrinberg:
Thinking about this again, it's probably fine to have some sort of behavior for cram tests, but it needs to provide some sort of an improvement over the regular behavior of killing the process group, and a way to disable this behavior for when the user wants to write their own signal handler. One way to improve the behavior could be to tell the user which pids are being leaked and making the test fail if there's leakage.

rgrinberg:
Closing as it's not useful in its current state.

PR #11831, opened 2025-05-21
Improve Scopes doc, put under Explanation

### LINKED ISSUE #11829: suggestion: document how configuration files compose
Dune is a composable build system: you can include a dune-using project as a subdirectory of another dune-using project, and things just work. This is great, but it is barely documented.

For example: what are the semantics when both a parent and a child directory have `dune-project` files? In the child subdirectory, is the child dune-project file ignored (certainly not!), is the parent dune-project file ignored (probably?), or is there some inheritance model where some of the settings from the parent are kept?

I found the documentation lacking:

- the [dune file](https://dune.readthedocs.io/en/latest/reference/dune/index.html) documentation says nothing about composition (my understanding: each `dune` file talks only about the directory in which it is placed in)
- the [dune-project file](https://dune.readthedocs.io/en/latest/reference/dune-project/index.html) documentation says nothing about composition (my best guess: ancestor dune-project files are ignored, only the closest dune-project file applies)
- the [dune-workspace file](https://dune.readthedocs.io/en/latest/reference/dune-workspace/index.html) documentation says nothing about composition (and I cannot at all guess how that works)

I found a documentation page on [Scopes](https://dune.readthedocs.io/en/latest/concepts/scopes.html) that seems related -- it explains clearly how to understand in which "scope" a given subdirectory resides -- but it does not mention anything in relation to `dune` or `dune-project` configuration, so I'm not sure when the notion of "scope" is relevant.

nojb:
Following #11829, I tried to give a bit more substance to the "Scopes" doc, and moved it to the "Getting Started" section, since it is pretty important information to make sense of how Dune operates.

Fixes #11829


doc/getting-started/scopes.rst
```diff
@@ -0,0 +1,32 @@
...
+:doc:`/reference/dune-project/index` file. Each Dune project extends over the
+subtree rooted at the directory containing the
+:doc:`/reference/dune-project/index` file, excluding any nested Dune projects.
+
+The root of the workspace is defined by the outermost
+:doc:`/reference/dune-project/index` file starting from the current directory,
+or by the first :doc:`/reference/dune-workspace/index` file found when climbing
+up the directory hierachy starting from the current directory (see
+:ref:`finding-root`). It can also be specified explicitly by passing the
+``--root`` flag (see :ref:`forcing-root`).
```
gasche, line 22: I found this paragraph a bit unclear (you say that the root is defined by A or B, but it's not clear which applies if both make sense; and I'm not sure that I understand the descriptions of A and B). In contrast, https://dune.readthedocs.io/en/stable/usage.html#finding-the-root is clear. I would less here and redirect to that part instead. (It's not clear to me why "Finding the root" is in the "Command-line Interface" page, but oh well.)

doc/getting-started/scopes.rst
```diff
@@ -0,0 +1,32 @@
+Scopes
```
gasche, line 1: Is "Scopes" the right title for this page? I'm not sure what the word refers to. Maybe if you tried to summarize what the page is about you would find a more descriptive title?

gasche (commented):
I believe this is an improvement, thanks!

- You mention `dune-workspace` and `dune-project` files in scopes.rst, but not `dune` files. It might be worth mentioning them briefly.
- I think it would be useful to have backlinks from the `dune`, `dune-project` and `dune-workspace` reference pages back to this page. (I'm not sure whether you have done that from the markup alone.) When I wanted to clarify the nested semantics of `dune-project` file, I went to look at the `dune-project` reference file.

nojb:
Thanks for the feedback @gasche!

> * You mention `dune-workspace` and `dune-project` files in scopes.rst, but not `dune` files. It might be worth mentioning them briefly.

Fixed: I added a brief blurb about them.

> * I think it would be useful to have backlinks from the `dune`, `dune-project` and `dune-workspace` reference pages back to this page. (I'm not sure whether you have done that from the markup alone.) When I wanted to clarify the nested semantics of `dune-project` file, I went to look at the `dune-project` reference file.

Fixed: I added links in each of these sections back to the one containing the new information.

doc/scopes.rst
```diff
@@ -0,0 +1,30 @@
...
+:doc:`/reference/dune-project/index` file. Each Dune project extends over the
+subtree rooted at the directory containing the
+:doc:`/reference/dune-project/index` file, excluding any nested Dune projects.
+
+The root of the current workspace is determined by the outermost
+:doc:`/reference/dune-project/index` file in an ancestor of the current
+directory or by the presence of a :doc:`/reference/dune-workspace/index` file
+(see :ref:`finding-root` and :ref:`forcing-root` for details). All
+:doc:`/reference/dune-workspace/index` files other than the one specifying the
+root of the workspace are ignored.
```
gasche, line 18: The last sentence is about `dune-workspace`, it could go to the dedicated paragraph currently on lines 24-25.

doc/scopes.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Different Dune projects within the same Dune workspace are independent of each
+other and no settings are shared between them, even if they are nested within
+each other.
+
+Settings in :doc:`/reference/dune-workspace/index`, on the other hand, are
+inherited by all Dune projects in the workspace.
+
+Within a Dune project, :doc:`/reference/dune/index` files are used to define
+all objects of interest for Dune: libraries, executables, tests, etc. Only
```
gasche, line 28: You could mention that `dune` files are relative to their current directory, and that there typically are many `dune` files in a single project.

gasche (approved):
This is fine as far as I'm concerned -- see minor comments. It does solve the issue I created. Thanks!

doc/scopes.rst
nojb: Fixed, thanks.

doc/scopes.rst
nojb: Good idea. Fixed. Thanks.

Alizter: It is also possible to use the `(subdir)` stanza. https://dune.readthedocs.io/en/stable/reference/dune/subdir.html

Ultimately each stanza does end up living in its own directory, but there is no such strict for the dune file.

doc/scopes.rst
```diff
@@ -0,0 +1,35 @@
...
+
+Settings in :doc:`/reference/dune-workspace/index`, on the other hand, are
+inherited by all Dune projects in the workspace. Note that all
+:doc:`/reference/dune-workspace/index` files other than the one specifying the
+root of the workspace are ignored.
+
+Within a Dune project, :doc:`/reference/dune/index` files are used to define all
+objects of interest for Dune: libraries, executables, tests, etc. All references
+within :doc:`/reference/dune/index` files are always written relative to the
+directory containing the file.  There are typically many Dune files in a Dune
```
gasche, line 30:
```suggestion
directory containing the file.  There are typically many :doc:`/reference/dune/index` files in a Dune
```

doc/scopes.rst
nojb: Fair enough, I have added a comment to this effect (I prefer not to explain the notion that it is the stanza that is attached to a directory, rather than the Dune file because it is a bit subtle and only relevant for advanced uses).

doc/scopes.rst
nojb: Thanks, fixed.

doc/scopes.rst
maiste, line 1: Wouldn't it be better if this put this under the `Explanation` section?

maiste (commented):
Thanks for the PR! I add a quick look. I'll try to have a closer one later.

doc/scopes.rst
nojb: My view is that this is pretty crucial information to understand Dune's behaviour, so under "Getting Started" seems OK. But don't have a very strong opinion on the subject.

Leonidas-from-XIV: I concur with @maiste on this one. When you're getting started you usually have one `dune-project` and no `dune-workspace` so I don't think that the understanding of the mechanism is particularly important for getting started.

nojb:
Friendly ping @maiste. Are you happy with the current state of the PR?

doc/explanation/index.rst
```diff
@@ -7,6 +7,7 @@ the rest of the OCaml ecosystem.
 .. toctree::
    :maxdepth: 1
 
+   ../scopes
```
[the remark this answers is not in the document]
nojb: OK, fixed, thanks.

doc/explanation/scopes.rst
```diff
@@ -0,0 +1,36 @@
+Dune projects and workspaces
+============================
+
+Whenever Dune builds anything, it does so at the level of a *Dune workspace*. A
+Dune workspace is a set of Dune projects. A typical workspace consists of a
+single Dune project.
+
+A Dune project is defined by the presence of a
+:doc:`/reference/dune-project/index` file. Each Dune project extends over the
+subtree rooted at the directory containing the
```
Alizter, line 10: "subtree" sounds like we are in a subdirectory when instead we are really talking about the directory at which the dune-project exists. I would probably reword that to "file tree".

doc/explanation/scopes.rst
```diff
@@ -0,0 +1,36 @@
+Dune projects and workspaces
```
Alizter, line 1:
```suggestion
Dune Projects and Workspaces
```

doc/explanation/scopes.rst
```diff
@@ -0,0 +1,36 @@
...
+
+A Dune project is defined by the presence of a
+:doc:`/reference/dune-project/index` file. Each Dune project extends over the
+subtree rooted at the directory containing the
+:doc:`/reference/dune-project/index` file, excluding any nested Dune projects.
+
+The root of the current workspace is determined by the outermost
+:doc:`/reference/dune-project/index` file in an ancestor of the current
+directory or by the presence of a :doc:`/reference/dune-workspace/index` file
+(see :ref:`finding-root` and :ref:`forcing-root` for details).
```
Alizter, line 16:
```
Dune determines the root of the current workspace by finding the topmost ancestor directory containing a `dune-project` file or by the presence of a `dune-workspace` file.
```

An alternative wording, feel free to pick and choose. Q: Do we always pick the topmost dune-workspace file? What happens if we have A/B where A contains a dune-project and B a dune-workspace?

doc/explanation/scopes.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Different Dune projects within the same Dune workspace are independent of each
+other and no settings are shared between them, even if they are nested within
+each other.
```
Alizter, line 20: What about dune language versions? I believe we take the maximum of all the dune-lang versions. We should also check if the same holds for syntax extensions.

If my dune-project has dune lang 3.18 can I vendor a project with dune lang 3.19 in its dune-project file? Probably not.

doc/explanation/scopes.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Settings in :doc:`/reference/dune-workspace/index`, on the other hand, are
+inherited by all Dune projects in the workspace. Note that all
+:doc:`/reference/dune-workspace/index` files other than the one specifying the
+root of the workspace are ignored.
```
Alizter, line 25: Maybe worth mentioning that dune-workspace files have different settings and configuration options than dune-project although there is some overlap. There is also dune config which I believe overrides the dune-workspace file but I am not sure.

doc/explanation/scopes.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Within a Dune project, :doc:`/reference/dune/index` files are used to define all
+objects of interest for Dune: libraries, executables, tests, etc. References
+within :doc:`/reference/dune/index` files are resolved relative to the directory
+containing the file (except if using a :doc:`/reference/dune/subdir` stanza).
```
Alizter, line 30: Or alternatively `chdir` inside a user action, but this might not be worth mentioning.

Alizter (commented):
Thanks @nojb for starting this. I have a few review comments and I will do another pass after those are done.

I would like for this document to be maximally informative to those who come across it and not have them with more questions than when they started.

doc/explanation/scopes.rst
Alizter: We use title case everywhere expect with headers that are about dune files like `dune` and `dune-project`.

doc/explanation/scopes.rst
nojb: Amended, thanks.

nojb:
Thanks for the careful review @Alizter! I think I addressed all points. Let me know if you have any additional feedback.

doc/explanation/scopes.rst
```diff
@@ -0,0 +1,39 @@
+Dune Projects and Workspaces
+============================
+
+Whenever Dune builds anything, it does so at the level of a *Dune workspace*. A
+Dune workspace is a set of Dune projects. A typical workspace consists of a
+single Dune project.
+
+A Dune project is defined by the presence of a
```
Alizter, line 8: Just as we used italics for Dune workspace, we should use italics for Dune project since we are defining the term.

nojb: Fixed, thanks.

doc/explanation/scopes.rst
```diff
@@ -0,0 +1,39 @@
...
+:doc:`/reference/dune-workspace/index` files other than the one specifying the
+root of the workspace are ignored.
+
+Within a Dune project, :doc:`/reference/dune/index` files are used to define all
+objects of interest for Dune: libraries, executables, tests, etc. References
+within :doc:`/reference/dune/index` files are resolved relative to the directory
+containing the file (except if using a :doc:`/reference/dune/subdir` stanza or
+the :doc:`/reference/actions/chdir` action`). There are typically many
+:doc:`/reference/dune/index` files in a Dune project (one per directory, unless
+the directory does not contain anything relevant to Dune).
```
maiste, line 36:
```suggestion
:doc:`/reference/dune/index` files in a Dune project (one per directory, unless
the directory does not contain anything relevant to Dune or the sub directory
is a descendant of a directory using :doc:`/reference/dune/include_subdirs`).
```
Should we add a reference to `(include_subdirs [...])`? I'm asking because this is a useful and underrated feature and it has its place in this part of the documentation. (I'm not sure if the suggestion is correctly working...)

nojb: Actually, I would rather keep in this section only the explanations of the general case, and indicate that these general rules have exceptoins, pointing the reader to the specific stanzas (`(subdir)`, `(include_subdirs)`, `(chdir)`, etc) to learn about the details.

I will push a commit in this direction shortly.

nojb: Addressed in 6602c2a02abf2737f51e08f031e9b167ca6c827b

maiste (commented):
I have done another round with minor questions and suggestions. Thanks for taking care of this! :smile:

nojb:
Thanks for the feedback @maiste. However, both of your suggestions concern other pages of the documentation that the one being addressed in this PR (only a backlink was added in this PR). I would rather that a dedicated PR be opened for them, maybe with further changes related to package management or any other needed information. Is that OK?

maiste (approved):
Oh I missed that, sorry! Sure no problem 😊
Under those conditions, LGTM!

Alizter (approved):
LGTM, merge when you wish.

nojb:
Thank you everyone for the feedback!

PR #11832, opened 2025-05-22
Warn about some common typos in package variables

gridbugs:
Dune uses underscores to separate words in pform variables but package variables use dashes for compatibility with opam. This will probably lead to typos. Technically arbitrarily-named solver variables are allowed by opam so typos in variable names have the potential to be silently accepted by dune while having no effect on the solution. This change prints a warning when dune encounters a package variable name that is identical to a known variable except with dashes replaced with underscores.

### DIFF
--- a/otherlibs/stdune/src/string.ml
+++ b/otherlibs/stdune/src/string.ml
@@ -328,3 +328,7 @@ let contains_double_underscore =
 ;;
 
 let last s = if length s > 0 then Some s.[length s - 1] else None
+
+let replace_char s ~from ~to_ =
+  String.map (fun c -> if Char.equal c from then to_ else c) s
+;;
--- a/src/dune_lang/package_variable_name.ml
+++ b/src/dune_lang/package_variable_name.ml
@@ -28,13 +28,6 @@ include (
   end) :
     Dune_util.Stringlike with type t := t)
 
-let encode t = Encoder.string (to_string t)
-
-let decode =
-  let open Decoder in
-  string >>| of_string
-;;
-
 let arch = of_string "arch"
 let os = of_string "os"
 let os_version = of_string "os-version"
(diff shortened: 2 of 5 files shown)


otherlibs/stdune/src/string.ml
```diff
@@ -328,3 +328,7 @@ let contains_double_underscore =
 ;;
 
 let last s = if length s > 0 then Some s.[length s - 1] else None
+
+let replace_char s ~from ~to_ =
+  split_on_char s ~sep:from |> String.concat ~sep:(String.make 1 to_)
+;;
```
Leonidas-from-XIV, line 334: I haven't benchmarked this but I feel it would be easier to use `String.map` as it will probably allocate dramatically fewer strings.

Something like

```ocaml
let replace_char ~from ~to_ =
  String.map ~f:(fun c ->
    if Char.equal c from then to_ else c)
```

src/dune_lang/package_variable_name.ml
```diff
@@ -49,13 +42,65 @@ let version = of_string "version"
 let name = of_string "name"
 let build = of_string "build"
 let post = of_string "post"
-let one_of t xs = List.mem xs ~equal t
 let dev = of_string "dev"
+let one_of t xs = List.mem xs ~equal t
```
Leonidas-from-XIV, line 46: This is moved for cleanup reasons?

src/dune_lang/package_variable_name.ml
```diff
@@ -49,13 +42,65 @@ let version = of_string "version"
...
+  let string = to_string t in
+  if String.contains string '_'
+  then
+    List.iter all_known ~f:(fun v ->
+      (* Many of dune's pform variables use underscores to separate words, but
+         package variables use underscores for compatibility with opam. Thus we
+         expect users to often mistakenly use underscores for package variables,
+         so warn in this case. *)
+      let v_string = to_string v in
+      let possible_typo = String.replace_char v_string ~from:'-' ~to_:'_' in
```
Leonidas-from-XIV, line 85: That's a constant value for each `v` that is declared in `all_known`, so you could just store it there once and avoid recomputing it.

test/blackbox-tests/test-cases/pkg/portable-lockdirs/portable-lockdirs-package-variable-typo-detection.t
```diff
@@ -0,0 +1,41 @@
...
+  >  (name mock)
+  >  (url "file://$(pwd)/mock-opam-repository"))
+  > (lock_dir
+  >  (repositories mock)
+  >  (solve_for_platforms
+  >   ((arch x86_64)
+  >    (os linux)
+  >    (os_distribution debian)
+  >    (os_family debian))))
+  > EOF
```
Leonidas-from-XIV, line 20: Sidenote on this: I think it would be useful to add environment variables that force Dune to assume a specific OS, distribution etc., otherwise any lockdir tests will be very dependent on which OS the test is being run.

Leonidas-from-XIV (commented):
Seems like a useful quality of life feature. Some comments for clarification.

src/dune_lang/package_variable_name.ml
gridbugs: Yeh it's just so that all of the variable names are next to each other.

otherlibs/stdune/src/string.ml
```diff
@@ -1,5 +1,5 @@
 (* Because other the syntax s.[x] causes trouble *)
-module String = Stdlib.String
+module String = Stdlib.StringLabels
```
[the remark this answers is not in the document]
gridbugs: It was just so I could pass the `sep` argument to `String.concat` more ergonomically. I've removed this.

otherlibs/stdune/src/string.ml
gridbugs: Ah good idea. I didn't know about `String.map`.

test/blackbox-tests/test-cases/pkg/portable-lockdirs/portable-lockdirs-package-variable-typo-detection.t
gridbugs: Good idea! I'll do that in a new PR as it doesn't really affect this change.

gridbugs: https://github.com/ocaml/dune/pull/11864

Leonidas-from-XIV (approved):
LGTM!

PR #11833, opened 2025-05-22
Display a warning when dune builds via RPC

gridbugs:
When an instance of dune is already running, invoking `dune build` will still build the target, howwever other command line arguments are silently ignored by dune as the current RPC interface is too limited to express them. This may change over time, but currently it's unusual UX for `dune build` to behave differently depending on whether or not another instance of `dune build` is already running. To more accurately set user expectations, print a warning when dune is running with this reduced functionality.

### DIFF
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -205,7 +205,7 @@ let build =
        state of the lock could otherwise change between checking it and taking
        it. *)
     match Dune_util.Global_lock.lock ~timeout:None with
-    | Error () ->
+    | Error lock_held_by ->
       (* This case is reached if dune detects that another instance of dune
          is already running. Rather than performing the build itself, the
          current instance of dune will instruct the already-running instance to
--- a/bin/build_cmd.ml
+++ b/bin/build_cmd.ml
@@ -215,6 +215,16 @@ let build =
          perform the RPC call.
       *)
       Scheduler.go_without_rpc_server ~common ~config (fun () ->
+        if not (Common.Builder.equal builder Common.Builder.default)
+        then
+          User_warning.emit
+            [ Pp.textf
+                "Your build request is being forwarded to a running Dune instance%s so \
+                 most command-line arguments will be ignored."
+                (match (lock_held_by : Dune_util.Global_lock.Lock_held_by.t) with
+                 | Unknown -> ""
+                 | Pid_from_lockfile pid -> sprintf " (pid: %d)" pid)
+            ];
         build_via_rpc_server ~print_on_success:true ~targets)
     | Ok () ->
       let request setup =
--- a/test/blackbox-tests/test-cases/watching/watching-eager-concurrent-build-command.t
+++ b/test/blackbox-tests/test-cases/watching/watching-eager-concurrent-build-command.t
@@ -9,6 +9,7 @@ Build the project once before starting the watch server so the watch server star
   $ dune build --watch &
   Success, waiting for filesystem changes...
   Success, waiting for filesystem changes...
+  Success, waiting for filesystem changes...
   File "foo.ml", line 1, characters 9-21:
   1 | let () = print_endlin "Hello, World!"
                ^^^^^^^^^^^^
@@ -24,6 +25,13 @@ Build the project once before starting the watch server so the watch server star
 
 Demonstrate that we can run "dune build" while the watch server is running.
   $ dune build
+  Success
+
+Demonstrate that a warning is displayed when extra arguments are passed to
+"dune build", since those arguments will be ignored.
+  $ dune build --auto-promote 2>&1 | sed 's/pid: [0-9]*/pid: PID/g'
+  Warning: Your build request is being forwarded to a running Dune instance
+  (pid: PID) so most command-line arguments will be ignored.
   Success
 
 Demonstrate that error messages are still printed by "dune build" when it's
--- a/test/expect-tests/dune_rpc_e2e/dune_rpc_e2e.ml
+++ b/test/expect-tests/dune_rpc_e2e/dune_rpc_e2e.ml
@@ -9,7 +9,7 @@ module Session = Csexp_rpc.Session
 
 (* enable to debug process stdout/stderr *)
 let debug = false
-let () = if debug then Dune_util.Log.init ~file:(Out_channel stderr) ()
+let () = if debug then Dune_util.Log.init ~file:Stderr ()
 
 let dune_prog =
   lazy
--- a/bin/arg.ml
+++ b/bin/arg.ml
@@ -41,6 +41,7 @@ module Dep = struct
 
   type t = Dep_conf.t
 
+  let equal = Dep_conf.equal
   let file s = Dep_conf.File (String_with_vars.make_text Loc.none s)
 
   let make_alias_sw ~dir s =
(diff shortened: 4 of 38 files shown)


test/blackbox-tests/test-cases/watching/watching-eager-concurrent-build-command.t
```diff
@@ -24,12 +24,20 @@ Build the project once before starting the watch server so the watch server star
 
 Demonstrate that we can run "dune build" while the watch server is running.
   $ dune build
+  Warning: A running dune (pid: 124305) instance has locked the build
+  directory. If this is not the case, please delete "_build/.lock".
+  The requested targets will still be built however functionality will be
+  reduced and some command line arguments will be ignored.
```
nojb, line 30: A couple of questions:

- "If this is not the case, please delete ..." is this relevant ? if the build is being forwarded to a running instance, it doesn't make much sense to me to ask the user to delete the lock file.
- "functionality will be reduced" what does this refer to?

I suggest tweaking the message and making it a bit less scary:

> Warning: your build request is being forwarded to a running Dune instance (pid: 124305). Note that certain command line arguments may be ignored.

bin/build_cmd.ml
```diff
@@ -217,6 +217,12 @@ let build =
          perform the RPC call.
       *)
       Scheduler.go_without_rpc_server ~common ~config (fun () ->
+        User_warning.emit
+          [ Dune_util.Global_lock.Lock_held_by.message lock_held_by
+          ; Pp.text
+              "The requested targets will still be built however functionality will be \
+               reduced and some command line arguments will be ignored."
+          ];
```
Leonidas-from-XIV, line 225: Would it be possible to adjust this error message to only display the warning when there are arguments that are being ignored? This way it warns about arguments being ignored even if there are no arguments.

test/blackbox-tests/test-cases/watching/watching-eager-concurrent-build-command.t
```diff
@@ -24,12 +24,20 @@ Build the project once before starting the watch server so the watch server star
...
   Success
 
 Demonstrate that error messages are still printed by "dune build" when it's
 acting as an RPC client while running concurrently with an RPC server.
   $ echo 'let () = print_endlin "Hello, World!"' > foo.ml
   $ dune build
+  Warning: A running dune (pid: 124305) instance has locked the build
```
Leonidas-from-XIV, line 37: Isn't this PID going to change every time the test runs? One would probably need to override the PID in the lock file to make it reproducible.

Leonidas-from-XIV (commented):
I like the idea of warning the user when flags get ignored, but I would suggest that this would only happen if there are flags that will get discarded.

bin/build_cmd.ml
```diff
@@ -215,6 +215,16 @@ let build =
          perform the RPC call.
       *)
       Scheduler.go_without_rpc_server ~common ~config (fun () ->
+        if not (Common.Builder.equal builder Common.Builder.default)
+        then
+          User_warning.emit
+            [ Pp.textf
+                "Your build request is being forwarded to a running Dune instance%s. \
+                 Note that certain command line arguments may be ignored."
```
Leonidas-from-XIV, line 223: Since this warning is now only emitted when arguments have been ignored, I think it should state that command line arguments have in fact been ignored.

test/blackbox-tests/test-cases/watching/watching-eager-concurrent-build-command.t
```diff
@@ -26,6 +27,13 @@ Demonstrate that we can run "dune build" while the watch server is running.
   $ dune build
   Success
 
+Demonstrate that a warning is displayed when extra arguments are passed to
+"dune build", since those arguments will be ignored.
+  $ dune build --auto-promote 2>&1 | sed 's/pid: [0-9]*/pid: PID/g'
+  Warning: Your build request is being forwarded to a running Dune instance
+  (pid: PID). Note that certain command line arguments may be ignored.
+  Success
+
```
Leonidas-from-XIV, line 36: Can you add an additional test case that shows that no such warning is printed if no arguments are ignored?

test/expect-tests/dune_rpc_e2e/dune_rpc_e2e.ml
```diff
@@ -9,7 +9,7 @@ module Session = Csexp_rpc.Session
 
 (* enable to debug process stdout/stderr *)
 let debug = false
-let () = if debug then Dune_util.Log.init ~file:(Out_channel stderr) ()
+let () = if debug then Dune_util.Log.init ~file:Stderr ()
```
Leonidas-from-XIV, line 12: Curious that the only case where `Out_channel` is used was this but it does simplify the code as it is impossible to compare equality of channels so that's neat.

Leonidas-from-XIV (commented):
It's a bit sad all these very verbose `equal` functions need to be added but I understand the idea.

Overall I think it needs a small wording change and an additional test and then it is good to go.

bin/build_cmd.ml
gridbugs: Not all arguments are ignored though. The `Builder` module is perhaps poorly named, as `Builder.term` takes care of parsing arguments like `--verbose` and `--display` which still take effect when dune is behaving as an rpc client. I'll still think about how to reword this message to be more helpful though.

test/blackbox-tests/test-cases/watching/watching-eager-concurrent-build-command.t
gridbugs: Does the command on line 26 of this file satisfy that?

bin/build_cmd.ml
gridbugs: I think the nuance is that we don't explicitly ignore arguments. Some arguments (e.g. `--verbose`, `--display`) change some parts of dune's global mutable state in ways that affect the output of the current instance of dune. These arguments are not ignored by `dune build` when it is acting as an rpc client; passing them will still cause dune to be more verbose or display output differently. Other arguments affect the behaviour of the build engine, and these are ignored by merit of the fact that the build engine is not run by dune when it's acting as an rpc client, and so incidentally these arguments are ignored.

Leonidas-from-XIV: Yes, but we only display if the Builder instances are unequal so we know for a fact that we ignored command line arguments, there is no "may" at this point?

gridbugs: Here's a little example usage of dune built from this branch. These commands were run while another dune was running in watch mode in a separate terminal:
```
2 s@u3 ~/src/llama (main=)
+ λ dune build examples/echo_effect.exe
Success

3 s@u3 ~/src/llama (main=)
+ λ dune build examples/echo_effect.exe --verbose
Shared cache: enabled
Shared cache location: /home/s/.cache/dune/db
Workspace root: /home/s/src/llama
Auto-detected concurrency: 32
Warning: Your build request is being forwarded to a running Dune instance
(pid: 347998) so most command-line arguments will be ignored.
Success
```

The builder instance was different from its default value because of `--verbose` however that flag is not ignored. When `Common.init` was called it set some global state including how dune should display its output, and this happens regardless of whether the builder is equal to its default value.

Leonidas-from-XIV: Ok, thanks for the clarification. I did not expect `--verbose` be considered for `Builder`.

Leonidas-from-XIV (approved):

PR #11835, opened 2025-05-22
Add standalone doc for dune-project global fields

### LINKED ISSUE #11834: Improve documentation visibility for global dune-project fields
I recently had some difficulty finding the documentation for some global dune-project fields. Eventually I figured out that they're documented in the `generate_opam_files` page, which was unexpected (by me, at least). My expectation is that the structure of the Dune reference manual mirrors the structure of the stanzas, so the top-level page lists all of the global fields, and each child page represents a stanza that contains its child fields. 

In my particular search, I was searching for the `source` field, which is unfortunately a keyword that's too common for the docs' search engine to be helpful. I eventually found the field documented under the `package` stanza page, but that page only says "the same as the corresponding global field," which isn't helpful if I can't find the global field. Perhaps such cases should link to their corresponding fields?

Relevant pages: 
- https://dune.readthedocs.io/en/stable/reference/dune-project/index.html
- https://dune.readthedocs.io/en/stable/reference/dune-project/generate_opam_files.html
- https://dune.readthedocs.io/en/stable/reference/dune-project/package.html

nojb:
Extracts the documentation for `dune-project` "global fields" (ie global package metadata) into their own toplevel stanza documnetation, under `dune-project`. Along the way, link the documentation of the global fields to those of the `package` stanza and vice-versa.

Fixes #11834


Leonidas-from-XIV (approved):

nojb:
Thanks for the review!

Issue #8114, opened 2023-07-05
Running tests and `exec`-ing executables should be possible while build server is running


gridbugs:
## Desired Behavior

I often run `dune test`, `dune build @test/<test>` or `dune exec <exe>` while the build server is running, get the error `A running dune instance has locked the build directory.`, kill the build server, and then run my original command again. I would rather not have to do this. I understand that these commands all build some files before running them so perhaps they could just take the latest versions of the files they need if they notice that the build server is running, possibly after waiting for the current build to complete.

## Example

Inside a dune project, in one terminal I want to be able to run:
```
dune build --watch
```

And in a second terminal in the same project I want to be able to run:
```
dune exec path/to/my/exe
```
...or...
```
dune test
```
...or...
```
dune build @test/name_of_test
```
...without the fact that the build directory is locked preventing the command from running.

Alizter:
The solution we ought to do is just connect to the existing RPC server when we do a build.

rgrinberg:
We need to rally the errors and progress from the server to the client

ElectreAAS:
Solved by #11712 for builds, #11840 for execs, and #12473 for tests

PR #6507, opened 2022-11-18
Eager watch mode for exec

gridbugs:
Fixes https://github.com/ocaml/dune/issues/2934

Signed-off-by: Stephen Sherratt <stephen@sherra.tt>

PR #11840, opened 2025-05-23
Allow  concurrent exec with watch mode

gridbugs:
This change allows a limited version of `dune exec` to run at the same time as dune is running in watch mode. This allows users to run example programs without needing to stop their watch server. This works by sending messages to the RPC server to build the executable if necessary.

### DIFF
--- a/bin/exec.ml
+++ b/bin/exec.ml
@@ -64,7 +64,16 @@ module Cmd_arg = struct
   let conv = Arg.conv ((fun s -> Ok (parse s)), pp)
 end
 
-let not_found ~dir ~prog =
+let not_found ~hints ~prog =
+  User_error.raise
+    ~hints
+    [ Pp.concat
+        ~sep:Pp.space
+        [ Pp.text "Program"; User_message.command prog; Pp.text "not found!" ]
+    ]
+;;
+
+let not_found_with_suggestions ~dir ~prog =
   let open Memo.O in
   let+ hints =
     (* Good candidates for the "./x.exe" instead of "x.exe" error are
@@ -81,30 +90,25 @@ let not_found ~dir ~prog =
     in
     User_message.did_you_mean prog ~candidates
   in
+  not_found ~hints ~prog
+;;
+
+let program_not_built_yet prog =
   User_error.raise
-    ~hints
     [ Pp.concat
         ~sep:Pp.space
-        [ Pp.text "Program"; User_message.command prog; Pp.text "not found!" ]
+        [ Pp.text "Program"
+        ; User_message.command prog
+        ; Pp.text "isn't built yet. You need to build it first or remove the"
+        ; User_message.command "--no-build"
+        ; Pp.text "option."
+        ]
     ]
 ;;
 
 let build_prog ~no_rebuild ~prog p =
   if no_rebuild
-  then
-    if Path.exists p
-    then Memo.return p
-    else
-      User_error.raise
-        [ Pp.concat
-            ~sep:Pp.space
-            [ Pp.text "Program"
-            ; User_message.command prog
-            ; Pp.text "isn't built yet. You need to build it first or remove the"
-            ; User_message.command "--no-build"
-            ; Pp.text "option."
-            ]
-        ]
+  then if Path.exists p then Memo.return p else program_not_built_yet prog
   else
     let open Memo.O in
     let+ () = Build_system.build_file p in
@@ -117,14 +121,14 @@ let get_path_and_build_if_necessary sctx ~no_rebuild ~dir ~prog =
   | In_path ->
     Super_context.resolve_program_memo sctx ~dir ~loc:None prog
     >>= (function
-     | Error (_ : Action.Prog.Not_found.t) -> not_found ~dir ~prog
+     | Error (_ : Action.Prog.Not_found.t) -> not_found_with_suggestions ~dir ~prog
      | Ok p -> build_prog ~no_rebuild ~prog p)
   | Relative_to_current_dir ->
     let path = Path.relative_to_source_in_build_or_external ~dir prog in
     Build_system.file_exists path
     >>= (function
      | true -> build_prog ~no_rebuild ~prog path
-     | false -> not_found ~dir ~prog)
+     | false -> not_found_with_suggestions ~dir ~prog)
   | Absolute ->
     (match
        let prog = Path.of_string prog in
@@ -137,7 +141,7 @@ let get_path_and_build_if_necessary sctx ~no_rebuild ~dir ~prog =
          Option.some_if (Path.exists prog) prog)
      with
      | Some prog -> Memo.return prog
-     | None -> not_found ~dir ~prog)
+     | None -> not_found_with_suggestions ~dir ~prog)
 ;;
 
 let step ~setup ~prog ~args ~common ~no_rebuild ~context ~on_exit () =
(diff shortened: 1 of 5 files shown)


Leonidas-from-XIV:
Thanks for the PR, it'll be easier to review once #11833 is merged.

Leonidas-from-XIV:
@gridbugs Can you rebase, now that #11833 is merged?

gridbugs:
Will do. I was waiting for https://github.com/ocaml/dune/pull/11879 to be merged first as it changes a lot of code in exec.ml.

gridbugs:
@Leonidas-from-XIV all the prereqs for this are now merged. Can you please take a look.

src/dune_util/global_lock.mli
```diff
@@ -3,9 +3,11 @@
     Before starting rpc, writing to the build dir, this lock should be locked. *)
 
 module Lock_held_by : sig
-  type t =
-    | Pid_from_lockfile of int
-    | Unknown
+  type t
+
+  (** returns "(pid: X)" where X is the PID if the PID is known, otherwise the
```
Leonidas-from-XIV: It actually returns a space before the paren. But I think it would sort of be nicer to just return `pid: X` or something and then let the formatter fit it into the formatting string.

test/blackbox-tests/test-cases/watching/dune
```diff
@@ -17,8 +17,9 @@
  (applies_to what-dune-watches)
  (deps %{bin:strace}))
 
-;; this test sometimes gets stuck and times out
+;; These tests sometimes get stuck and time out:
 
 (cram
- (applies_to watching-eager-concurrent-build-command)
+ (applies_to watching-eager-concurrent-build-command
+  watching-eager-concurrent-exec-command)
```
Leonidas-from-XIV, line 24: But doesn't that mean that the tests are basically never run? Not on CI, not on people's machines when they `dune runtest`?

test/blackbox-tests/test-cases/watching/watching-eager-concurrent-exec-command.t
```diff
@@ -0,0 +1,50 @@
...
+  Error: Program './foo.exe' isn't built yet. You need to build it first or
+  remove the '--no-build' option.
+  [1]
+
+Demonstrate running an executable from the current project:
+  $ dune exec ./foo.exe
+  Hello, World!
+
+Demonstrate running an executable from PATH:
+  $ dune exec echo "Hello, World!"
```
Leonidas-from-XIV: Maybe it would be good if `echo` and `foo` had different output, to distinguish what is being called.

bin/exec.ml
```diff
@@ -164,17 +168,69 @@ let step ~setup ~prog ~args ~common ~no_rebuild ~context ~on_exit () =
...
+   directory lock.
+
+   Returns the absolute path to the executable. *)
+let build_prog_via_rpc_if_necessary ~dir ~no_rebuild prog =
+  match Filename.analyze_program_name prog with
+  | In_path ->
+    User_warning.emit
+      [ Pp.textf
+          "As this is not the main instance of Dune it is unable to locate the \
+           executable %S within this project. Dune will attempt to resolve the \
```
Leonidas-from-XIV, line 197: If we know that is it `In_path`, why do we warn? We already know that it is in the `PATH`.

bin/exec.ml
```diff
@@ -164,17 +168,69 @@ let step ~setup ~prog ~args ~common ~no_rebuild ~context ~on_exit () =
...
+
+let exec_building_via_rpc_server ~common ~prog ~args ~no_rebuild =
+  let open Fiber.O in
+  let ensure_terminal = function
+    | Cmd_arg.Terminal s -> s
+    | Expandable (_, raw) ->
+      (* Pforms cannot be expanded without running the build system. *)
+      User_error.raise
+        [ Pp.textf
+            "The term %S contains a pform variable but Dune is unable to expand pform \
```
Leonidas-from-XIV: Fair to error out but the error message is too technical. Dune users don't know what "pforms" are.

bin/exec.ml
```diff
@@ -164,17 +168,69 @@ let step ~setup ~prog ~args ~common ~no_rebuild ~context ~on_exit () =
...
+  | Absolute ->
+    if Path.exists (Path.of_string prog)
+    then Fiber.return prog
+    else not_found ~hints:[] ~prog
+;;
+
+let exec_building_via_rpc_server ~common ~prog ~args ~no_rebuild =
+  let open Fiber.O in
+  let ensure_terminal = function
+    | Cmd_arg.Terminal s -> s
```
Leonidas-from-XIV: I think the general rule in the Dune style guide is to annotate the type instead of a single constructor of a match.

bin/exec.ml
```diff
@@ -209,4 +265,40 @@ let term : unit Term.t =
...
+let term : unit Term.t =
+  let+ builder = Common.Builder.term
+  and+ context = Common.context_arg ~doc:{|Run the command in this build context.|}
+  and+ prog = Arg.(required & pos 0 (some Cmd_arg.conv) None (Arg.info [] ~docv:"PROG"))
+  and+ no_rebuild =
+    Arg.(value & flag & info [ "no-build" ] ~doc:"don't rebuild target before executing")
+  and+ args = Arg.(value & pos_right 0 Cmd_arg.conv [] (Arg.info [] ~docv:"ARGS")) in
+  (* TODO we should make sure to finalize the current backend before exiting dune.
+     For watch mode, we should finalize the backend and then restart it in between
+     runs. *)
```
Leonidas-from-XIV, line 291: I know that the comment is from before but I don't understand what it means. What is a "backend" in this context? Why restart? The comment raises more questions than in answers.

Leonidas-from-XIV (changes requested):
Very useful work, however I am concerned about the maintainability if the tests are disabled for being flaky. Is there maybe some way to make them more stable?

test/blackbox-tests/test-cases/watching/watching-eager-concurrent-exec-command.t
```diff
@@ -0,0 +1,50 @@
...
+  Error: Another instance of dune (pid: PID) has locked the _build
+  directory. Refusing to start a new watch server until no other instances of
+  dune are running.
+
+Demonstrate running an executable via an absolute path:
+  $ dune exec $(which echo) "Hello, World!"
+  Hello, World!
+
+  $ dune shutdown
+  $ wait
```
Alizter, line 50: Why are we waiting at the end?

bin/exec.ml
gridbugs: Dune's commit history would suggest that I wrote this line (over 2 years ago) but I was curious because it doesn't sound like the way I write. I tracked down the original commit that was squashed into my commit: https://github.com/ocaml/dune/commit/90cbf648efc8b8b5185aad1f54b019dd25ed09bc (in this PR: https://github.com/ocaml/dune/pull/6507). The "backend" @rgrinberg is referring to here is the `Console.Backend.t` which abstracts the way dune prints messages.

gridbugs: The change described in that comment doesn't sound complicated so maybe I'll just make the suggested change (in another PR) while I remember how `dune exec` works.

bin/exec.ml
gridbugs: Good point, I removed the word "pform" and just refer to them as "variables".

bin/exec.ml
gridbugs: That's fine for `match` statements but I'm not clear on the best way to achieve the same with `function`s. How about:
```ocaml
  let ensure_terminal : Cmd_arg.t -> _ = function                                                                                                                                                         │
    | Terminal s -> s                                                                                                                                                                                     │    feature: Eager watch mode for exec (#6507)
    | Expandable (_, raw) ->
```

bin/exec.ml
gridbugs: `In_path` just means that a program was named by its filename rather than a path, e.g. `dune exec foo` rather than `dune exec ./foo` or `dune exec /path/to/foo`. When dune is building directly (rather than over rpc) dune will first attempt to resolve the named executable within the project, building and running a local executable if any executable's `public_name` matches the argument to `dune exec`. Only if no local exe is found will dune then locate the program in the directories listed in the user's PATH variable.

Locating an executable within the project requires evaluating a `Memo.t` which dune cannot do when building over rpc (as it would require taking the global build directory lock). So for now when building over rpc dune only looks up the exe in PATH and not the current project. This leads to different behaviour if the project defines executables with the same name as those in PATH. We can't even detect if this is the case because again that would involve evaluating a memo. Hence dune prints a warning whenever a filename is passed to `dune exec` while dune is building over rpc.

test/blackbox-tests/test-cases/watching/dune
gridbugs: I agree it's not ideal but I don't have a reliable way to reproduce the problem. The tests are reliable on my dev machines and only fail from time to time in CI. Here's the PR that disabled this test initially: https://github.com/ocaml/dune/pull/11850. There are some other watch-mode tests that are disabled for similar reasons.

src/dune_util/global_lock.mli
gridbugs: I've updated the comment to include the leading space. I agree with the sentiment about just returning `pid: X` but looking at how this function is actually used, it's always to produce a message like `another dune instance (pid: X) has the lock` if the pid is known and a message `another dune instance has the lock` otherwise. I'm trying to reduce the repetition when producing messages like this. I don't think my solution is super elegant which is why I initially just exposed the variant type but I found that I was repeating the logic for constructing the string a lot.

test/blackbox-tests/test-cases/watching/watching-eager-concurrent-exec-command.t
gridbugs: My intention is to wait for the background process started by `dune build README.md --watch &` to terminate before exiting the test.

gridbugs:
> Very useful work, however I am concerned about the maintainability if the tests are disabled for being flaky. Is there maybe some way to make them more stable?

Flaky watch-mode tests is something that has plagued dune for as long as I've worked on it and I'm not sure what exactly we can do about it. @Alizter I recall you doing some work on this recently?

bin/exec.ml
Leonidas-from-XIV: @rgrinberg What's your take on this? A number of cases are mentioned in the [General Guidelines](https://github.com/ocaml/dune/blob/main/doc/hacking.rst#general-guidelines), maybe worth updating with this case for `function`?

I'd personally would write it as

```ocaml
  let ensure_terminal v =
    match v : Cmd_arg.t with
    | Terminal s -> s
    | Expandable (_, raw) ->
```

As that does not require defining a `-> _` case for the function, however at the expense of introducing a pointless binding.

Alizter:
@gridbugs I fixed it somewhat with https://github.com/ocaml/dune/pull/11827 and then it was decided a more general approach as in https://github.com/ocaml/dune/pull/11841 would be a better fit. The issue in the latter is that the CI is still a bit flaky for some reason.

bin/exec.ml
rgrinberg: Both are fine IMO. Although I would do `ensure_terminal (v : Cmd_arg.t)`

bin/exec.ml
Leonidas-from-XIV: Ok, I see. Thanks for the explanation, it seems very reasonable given the constraints!

(Maybe a short comment above the code explaining this could be useful to future readers of the code?)

bin/exec.ml
gridbugs: Updated.

bin/exec.ml
gridbugs: Added a comment explaining this.

src/dune_util/global_lock.mli
```diff
@@ -3,9 +3,11 @@
...
 
 module Lock_held_by : sig
-  type t =
-    | Pid_from_lockfile of int
-    | Unknown
+  type t
+
+  (** returns " (pid: X)" where X is the PID if the PID is known, otherwise the
+      empty string *)
+  val to_string_empty_if_unknown : t -> string
```
Leonidas-from-XIV: I feel like this is a very odd function, where the name is nearly as long as the implementation would be and kind of putting the responsibility of properly displaying values from the UI to some internal function.

I think if you want to keep `t` abstract, I'd rather implement `val pid -> t -> int option` and let the UI code deal with formatting this.

bin/exec.ml
```diff
@@ -81,30 +90,25 @@ let not_found ~dir ~prog =
...
-        [ Pp.concat
-            ~sep:Pp.space
-            [ Pp.text "Program"
-            ; User_message.command prog
-            ; Pp.text "isn't built yet. You need to build it first or remove the"
-            ; User_message.command "--no-build"
-            ; Pp.text "option."
-            ]
-        ]
+  then if Path.exists p then Memo.return p else program_not_built_yet prog
```
Leonidas-from-XIV, line 111: Nice refactor!

Leonidas-from-XIV (commented):
My only remaining complaint is that I don't like how the utility functions controls the output prettyprinting of the type, I believe the way that the PID is printed should be done in the `bin/*` modules.

src/dune_util/global_lock.mli
gridbugs: Fair enough, I've changed this code back to exposing the `Lock_held_by.t` type and moved the formatting logic into the individual command implementations.

Leonidas-from-XIV (approved):

Alizter (approved):

yakobowski:
This looks very nice, thanks a lot!! Would this MR solve https://github.com/ocaml/dune/issues/8114 too by any chance?

shonfeder:
@yakobowski the part of that relating to tests is being addressed in https://github.com/ocaml/dune/pull/12473

PR #11841, opened 2025-05-23
Terminate process group after cram execution

### LINKED ISSUE #11820: Background tasks in Cram tests are not terminated

Leonidas-from-XIV:
After discussing an alternate solution for #11827 with @rgrinberg we agreed that the best course of action is to just terminate the entire process group of the cram test after execution.

This PR reuses @Alizter's cram test repro case.

It introduces a slight race condition that we might be terminating an unrelated process group that accidentally got the same PID as the cram test. I am unaware of a way to avoid it, the chance is not substantial as OSes usually increment PIDs until they wrap around. In any case the engine is already terminating process groups if the process to be waited on has a timeout, so this code just extends it to the case where there is no timeout.

Fixes #11820

### DIFF
--- a/src/dune_engine/scheduler.ml
+++ b/src/dune_engine/scheduler.ml
@@ -889,7 +889,12 @@ let wait_for_build_process t pid =
       ~on_cancel:(fun () ->
         Process_watcher.killall t.process_watcher Sys.sigkill;
         Fiber.return ())
-      (fun () -> wait_for_process t pid)
+      (fun () ->
+         let+ r = wait_for_process t pid in
+         (* [kill_process_group] on Windows only kills the pid and by this
+            time the process should've exited anyway *)
+         if not Sys.win32 then kill_process_group pid Sys.sigterm;
+         r)
   in
   ( res
   , match outcome with
--- a/doc/changes/11841.md
+++ b/doc/changes/11841.md
@@ -1,0 +1,4 @@
+- Kill all processes in the process group after the main process has
+  terminated; in particular this avoids background processes in cram tests to
+  stick around after the test finished (#11841, fixes #11820, @Alizter,
+  @Leonidas-from-XIV)
--- a/test/blackbox-tests/test-cases/cram/subprocess.t
+++ b/test/blackbox-tests/test-cases/cram/subprocess.t
@@ -7,8 +7,7 @@ project with a single cram test.
 
 We create a file for tracking the PID of a subprocess.
 
-  $ cat > pid.txt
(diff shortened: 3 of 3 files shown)


Leonidas-from-XIV:
Surprising behavior on Windows when we `kill_process_group` which attempts to `Unix.kill` the PID which somehow causes corruption in the files that `ocamldep` writes. That's odd because at that time we have awaited the process so it should have exited already.

src/dune_engine/scheduler.ml
```diff
@@ -889,7 +889,12 @@ let wait_for_build_process t pid =
       ~on_cancel:(fun () ->
         Process_watcher.killall t.process_watcher Sys.sigkill;
         Fiber.return ())
-      (fun () -> wait_for_process t pid)
+      (fun () ->
+         let+ r = wait_for_process t pid in
+         (* [kill_process_group] on Windows only kills the pid and by this
+            time the process should've exited anyway *)
+         if not Sys.win32 then kill_process_group pid Sys.sigkill;
```
rgrinberg, line 896: How about `sigterm` perhaps? It might still leak things, but at least this is better in the common case.

Leonidas-from-XIV: Done!

rgrinberg (approved):

rgrinberg:
Seems like there's a failure in CI. Is it something relevant?

Leonidas-from-XIV:
I'm trying to figure out. It works reliably on my system (only failures are from Melange), CI sometimes hangs forever which is unrelated to this PR but its hard to say whether the failure is related since I can't reproduce it.

Alizter:
@Leonidas-from-XIV Could you push a temporary commit disabling all of `exec-watch` and perhaps `watching`. Those would be my goto culprits for tests that doesn't terminate.

PR #11849, opened 2025-05-24
feat: multiple foreign language support

Alizter:
In this PR we add support for multiple foreign languages to be specified in a single `foreign_stubs` field or `foreign_library`. These fields/stanzas now take a `languages` field to allow manual specification of languages or alternatively can be omitted defaulting to automatic language detection.

We also add support for Assembly language.

I'll mark this PR as draft for the time being so that I can separate out the commits and get them merged separately. This PR will be undrafted when the main feature is ready to be reviewed.

### DIFF
--- a/src/dune_lang/foreign_language.ml
+++ b/src/dune_lang/foreign_language.ml
@@ -6,6 +6,8 @@ type t =
   | `Asm
   ]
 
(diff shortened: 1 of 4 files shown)


Alizter:
I'll come back to this if there is a need.

PR #11854, opened 2025-05-25
refactor(foreign): make Foreign_language.t a poly variant

Alizter:
We make foreign_language.t a polymorphic variant to allow for more fine grained casing. Later, as outlined in #11849, we will add more foreign languages, so it is important that we can refer to a subset of the foreign languages for later language specific rules. 

We also do some routine cleanup and dead code removal.


rgrinberg:
In what situations do we benefit from this? As in, when do we want to refer to a subset of the languages?

Alizter:
Some of our language rules are treating C/C++ at the same time, but assembly isn't part of that. If we don't use a poly variant, we have to introduce another type for choosing between C/C++ only.

rgrinberg:
What are some examples of rules where we want to do something for C/C++ and exclude assembly?

Alizter:
Have a look at the patch introducing assembly rules: https://github.com/ocaml/dune/pull/11849/commits/3be6dd057d5705ec88c84d1c672736b78f52f589

Alizter:
If you are still unconvinced, I will revert the poly variant change and just keep the cleanups. I'll need to introduce some extra types to refer to the C/C++ subset too.

rgrinberg:
I'm just wary of special casing asm from those other language. Looking at the commit you referred to, the only use seems to be to avoid implementing flags support for assembly? I don't think that would work well. Passing flags to assembly compilation is going to be necessary. If your plan is to reuse the c flags for assembly, that could work. But then why do the rules need to care about the type of sources at all?

PR #11859, opened 2025-05-25
fix: .expected file for tests forbidden from being generated

Alizter:


rgrinberg:
The issue with this approach is that it waits until the action is executed to tell the user this isn't allowed. We should be able to tell the user about this error after generating the rules for the directory without running anything.

PR #11861, opened 2025-05-25
test: dune runtest foo.t/run.t

rgrinberg:
Demonstrate that `dune test foo.t/run.t` is broken. We should either forbid this, or make it the same `dune test foo.t`


test/blackbox-tests/test-cases/runtest-cmd.t
```diff
@@ -27,20 +28,33 @@ This should work:
   _build/default/tests/myothertest.t/run.t.corrected differ.
   [1]
 
+Ther's no diff produced because the test passes
+
+  $ dune promotion diff tests/myothertest.t/run.t
```
Alizter, line 33: I can promote just before here. Looks like we have a registered promotion but this `promotion diff`/`promotion list` command is broken.

PR #11863, opened 2025-05-25
feature: share os family of vars with rules

### LINKED ISSUE #838: variables for selecting builds based on compiler/arch/os
It would be useful to be able to select build rules based on some variables other than the presence of libraries.  For example:

- ocaml-version: Merlin needs to select a module based on the current switch
- arch: C bindings need to select arm/x86 specific modules (nocrypto)
- os: platform bindings (e.g. mirage-xen/mirage-unix)

For each of these, it is possible to do a configurator/jbuild-generation pass, but the big advantage of having variable rules directly is that they work when embedded in a Duniverse of related packages.  For example, with a manually assembled Mirage duniverse, the default `dune build` rule doesnt work as it selects the Linux- and macOS-only packages simultaneously.  If Dune could handle these parameters as part of the build scanning process, then package embedding works great.

As a longer term consideration, this also expands our ability to generate accurate package management scripts as well -- for example, `opam` files could have accurate `available:` constraints added based on these build rules as well.

rgrinberg:
These variables are available to package rules. Might as well make them available to all build rules for uniformity and convenience.

- [x] bump the latest version to 3.20 in a separate PR

### DIFF
--- a/doc/changes/11863.md
+++ b/doc/changes/11863.md
@@ -1,0 +1,4 @@
+- Introduce `%{os}`,`%{os_version}`, `%{os_distribution}`, and `%{os_family}`
+  percent forms. These have the same values as their opam counterparts.
+  (#11863, @rgrinberg)
+
--- a/src/dune_lang/pform.ml
+++ b/src/dune_lang/pform.ml
@@ -2,6 +2,25 @@ open Import
 module Payload = Template.Pform.Payload
 
 module Var = struct
+  module Os = struct
+    type t =
+      | Os
+      | Os_version
+      | Os_distribution
+      | Os_family
+
+    let all = [ Os; Os_version; Os_distribution; Os_family ]
+
+    let to_string = function
+      | Os -> "os"
+      | Os_version -> "os_version"
+      | Os_distribution -> "os_distribution"
+      | Os_family -> "os_family"
+    ;;
+
+    let to_dyn t = Dyn.variant (to_string t) []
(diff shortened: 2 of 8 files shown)


Leonidas-from-XIV (approved):
Code is as straightforward as it gets. However I am  worried because the fact that the actual values are quite unclear in OPAM ("is it `oraclelinux` or `ol`" "is it `opensuse-leap`, `suse-leap` or `opensuse`" "is `debian` a distribution or family?" "Is `ubuntu` a debian or is it its own family?") I am not convinced that introducing this ambiguity into `dune` files is a good idea.

Do we have requests where people want their builds to do different things depending on their Linux distribution?

Leonidas-from-XIV:
> They already need to know about this when writing constraints when using package management or submitting packages to opam.

Of which none of it is required to use Dune itself, we only implement enough to support it in package management. I would not say we encourage their use otherwise. Even in OPAM filtering on distributions is mostly only useful in the context of depexts and not on dependencies.

But I do get that it's somewhat frustrating that some variables are only available in the context of package management despite Dune *clearly* being able to expand them with the right values.

It's your call but I found determining the possible values to be tedious (they're not documented in the OPAM documentation but I believe in a comment by David on an issue in the bug tracker) so maybe we should at least extend the Dune documentation with the expected values of these values somewhere. Just so one does not have to set up VMs to determine what values a particular distribution will evaluate to.

PR #11864, opened 2025-05-26
Set a default platform for solving deps in pkg tests

gridbugs:


Leonidas-from-XIV (approved):
Oh, neat; I somehow forgot these existed. Definitely a good idea to set them in the tests.

PR #11869, opened 2025-05-27
fix: dune test dirtest.t/run.t running cram test incorrectly

Alizter:
We fix an issue where a user can write

```
dune test dirtest.t/run.t
```

and the corresponding cram test was being run incorrectly. This was due to the buggy parent detection code in runtest.ml. We fix how the parent directory of a cram test is actually detected and do some cleanup for cram source related code.

The runtest.ml test is updated to better reflect what we expect the command to do.

- [ ] changelog

### DIFF
--- a/bin/runtest.ml
+++ b/bin/runtest.ml
@@ -35,21 +35,20 @@ let runtest_info =
 
 let find_cram_test path ~parent_dir =
   let open Memo.O in
-  Source_tree.nearest_dir parent_dir
-  >>= Dune_rules.Cram_rules.cram_tests
-  (* We ignore the errors we get when searching for cram tests as they will
-     be reported during building anyway. We are only interested in the
-     presence of cram tests. *)
-  >>| List.filter_map ~f:Result.to_option
-  (* We search our list of known cram tests for the test we are looking
-     for. *)
-  >>| List.find ~f:(fun (test : Source.Cram_test.t) ->
-    let src =
-      match test with
-      | File src -> src
-      | Dir { dir = src; _ } -> src
-    in
-    Path.Source.equal path src)
+  Source_tree.find_dir parent_dir
+  >>= function
+  | None -> Memo.return None
+  | Some dir ->
+    Dune_rules.Cram_rules.cram_tests dir
+    >>| List.find_map ~f:(function
+      | Ok cram_test when Path.Source.equal path (Source.Cram_test.path cram_test) ->
+        Some cram_test
+      (* We raise any error we encounter when looking for our test specifically. *)
+      | Error (Dune_rules.Cram_rules.Missing_run_t cram_test)
+        when Path.Source.equal path (Source.Cram_test.path cram_test) ->
+        Dune_rules.Cram_rules.missing_run_t cram_test
+      (* Any errors or successes unrelated to our test are discarded. *)
+      | Error (Dune_rules.Cram_rules.Missing_run_t _) | Ok _ -> None)
 ;;
 
 let explain_unsuccessful_search path ~parent_dir =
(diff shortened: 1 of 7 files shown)


bin/runtest.ml
```diff
@@ -33,6 +33,20 @@ let runtest_info =
...
 
 let find_cram_test path ~parent_dir =
+  (* We correct the [path] and [parent_dir] if we have a directory test and the
+     user has given a path to "run.t" rather than the directory cram test itself. *)
+  let path, parent_dir =
+    Option.value ~default:(path, parent_dir)
+    @@
+    let open Option.O in
+    let* basename = Path.Source.basename_opt path in
+    if String.equal basename Source.Cram_test.fname_in_dir_test
```
rgrinberg, line 43: This looks fishy to me, there should be a way to look up tests without having to replicate dune's own logic for discovering them. The logic should be roughly like this:

1. Given a path `foo`, look up the source directory for the parent of this path. Let's call it `parent foo`.
2. If `parent foo` contains a cram test with `chop_prefix (basename foo) ~suffix:".t"`, then we found our test.
3. Otherwise, we check if `foo` is the `run.t` inside a cram test. For that, we now look for a test `parent foo`.
4. if we find such a test, we return it if the "script" of the said test is equal to the user provided path.

All of this logic should hopefully be shared as much as possible with the logic for discovering cram tests in the rules.

Alizter: I think there is a subtle issue with this logic which is why I have to correct things here. If the user provides `a.t/run.t` then by (1) we have `parent foo = "a.t"`. Now we also satisfy (2) because `run.t` looks like a file cram test to us. Even if we load the cram rules in `a.t` we will have `run.t` reported erroneously as a cram test. When we load rules we don't go into `a.t/` subdirectories for cram tests so we don't have this issue.

I'm also not sure what logic you would like to share with loading the cram tests, there we are given a parent directory and simplify check if there are cram tests there, whereas here we are given directory that could ambiguously be a file or directory cram test (or the script of a directory cram test), we then have to disambiguate and check that its parent directory actually contains such a cram test.

rgrinberg: I pushed out a commit sketching out what I had in mind. Seems to be working

bin/runtest.ml
```diff
@@ -35,21 +35,19 @@ let runtest_info =
...
-      | Dir { dir = src; _ } -> src
-    in
-    Path.Source.equal path src)
+  Source_tree.find_dir parent_dir
+  >>= function
+  | None -> Memo.return None
+  | Some dir ->
+    Dune_rules.Cram_rules.cram_tests dir
+    >>| List.find_map ~f:(function
+      | Error _ ->
```
Alizter: The only way this can happen is if a directory cram test has a missing `run.t`. We already check this when we load the cram rules. I don't think it is useful to do anything extra here.

rgrinberg: Yes, we do the check when we load the rules. But we aren't loading the rules here. In fact, we might erroneously tell the user that a test doesn't exist just because the run.t file is missing. You can write a test for this if you aren't convinced.

Alizter: I see the issue now. I've added a test, and I will push a fix shortly.

bin/runtest.ml
```diff
@@ -35,21 +35,18 @@ let runtest_info =
...
-    in
-    Path.Source.equal path src)
+  Source_tree.find_dir parent_dir
+  >>= function
+  | None -> Memo.return None
+  | Some dir ->
+    Dune_rules.Cram_rules.cram_tests dir
+    >>| List.find_map ~f:(function
+      | Error (Dune_rules.Cram_rules.Missing_run_t cram_test) ->
+        Dune_rules.Cram_rules.missing_run_t cram_test
```
rgrinberg, line 49: I think you should only error if the cram test is matching the test requested by the user.

Alizter: Good point. I've rephrased the selection to be clearer now.

rgrinberg (approved):

PR #11873, opened 2025-05-30
docs: update copyrights

maiste:
@nojb PR improving scope documentation made me realize that we should extend the credits to the many maintainers that have spent time improving the documentation over time.
I also extended the year range because on the documentation website, nothing makes it clear it is up-to-date.

What do you think?


nojb (approved):
LGTM

Leonidas-from-XIV (approved):

PR #11874, opened 2025-05-31
feature(bootstrap): assembly file support

rgrinberg:
For the upcoming assembly implementation of blake3

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -59,16 +59,17 @@ module List = struct
   include List
 
   let partition_map_skip t ~f =
-    let rec loop l r = function
-      | [] -> l, r
+    let rec loop l m r = function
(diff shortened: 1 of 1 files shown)


boot/duneboot.ml
```diff
@@ -900,8 +920,26 @@ module Library = struct
...
     ; alias_file : string option
     ; c_files : c_file list
+    ; asm_files : string list
     }
 
+  let keep_asm { File_kind.syntax; arch; os } ~ccomp_type ~architecture =
+    (match os with
+     | Some `Unix -> Sys.os_type = "Unix"
+     | Some `Win -> Sys.os_type = "Win32"
+     | None -> true)
```
Alizter, line 930: Is this meant to be false? Otherwise what is the point of this whole match?

PR #11876, opened 2025-05-31
fix(exec): Spawn dune exec processes in inherited process group

### LINKED ISSUE #11867: dune exec seems to be hanging with 3.19
## Expected Behavior

With `dune.3.18.2`, `dune exec` behaves as I am used to.

## Actual Behavior

After an upgrade to `dune.3.19.0` invocations of commands, such as with `--help` seems to be hanging.

For example at the root of https://github.com/mbarbin/dunolint (I don't think the actual repo matters, this seems a broader issue).

```sh
$ dune exec dunolint -- --help
```

Note that if the flag has a typo, this returns right away with the expected error message.

```sh
$ dune exec dunolint -- --typo
```
*returns right away after zero-build with usual errors on the console*.

This is weird, as if this would depend on the exit code of the command being run..

## Reproduction

I've tried with a few commands built with cmdliner and the issue seem pretty consistent across repos.

Please let me know if you'd like me to try adding some debug flags or something. I'd appreciate if someone can try to see if they can reproduce too!

Thanks!

## Specifications

- Version of `dune` (output of `dune --version`): 3.19.0
- Operating system (distribution and version): Ubuntu 24.04.2 LTS

### LINKED ISSUE #11870: Fails to pass input to program via stdin when launching program with "dune exec"
## Expected Behavior

"abc" is printed at the step 5 in the reproduce steps below.

## Actual Behavior

Nothing is printed at the step 5 in the reproduce steps below.

## Reproduction

1. Create a new project directory for reproduce: `dune init dune-repro`
2. Enter the project directory: `cd dune-repro`
3. Change `bin/main.ml` to this:
   ```ocaml
   let () = print_endline @@ read_line ()
   ```
4. `dune exec dune-repro`
5. Type `abc<Enter>` (where the `<Enter>` means the enter key)

## Specifications

- Version of `dune` (output of `dune --version`): 3.19.0
- Operating system (distribution and version): macOS 15.4.1
- Shell: bash 5.2.37
   ```
   GNU bash, version 5.2.37(1)-release (aarch64-apple-darwin23.4.0)
   Copyright (C) 2022 Free Software Foundation, Inc.
   License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

   This is free software; you are free to change and redistribute it.
   There is NO WARRANTY, to the extent permitted by law.
   ```

## Additional information 

- If you run the binary directly from the shell, this bug does not happen: `./_build/install/default/bin/dune-repro`
- This bug is not reproducible with dune 3.18.2.
   - When I tried this, I modified the dune-project file in order to run this project with older dune:
      ```diff
      diff --git a/dune-project b/dune-project
      index 7caa58a..a2241c1 100644
      --- a/dune-project
      +++ b/dune-project
      @@ -1,4 +1,4 @@
      -(lang dune 3.19)
      +(lang dune 3.18)
 
       (name dune-repro)
      ```
- As far as I bisected, this bug is introduced in this commit: https://github.com/ocaml/dune/commit/6fa1f6ea66b4b6b0aecd7a243407d2a72f7fa22e

Alizter:
In dune 3.18 we invoked Spawn directly to start a process and in turn we defaulted to inheriting the process group. In 3.19 this was changed to use the process managment given by Dune_engine.Process however here we always spawn processes in a new process group.

Terminal sessions typically rely on the process group when doing interactive things like reading lines from the user. This meant we ran into issues such as #11870 and #11867.

The fix here is to revert back to inheriting the process group.

- [x] changelog
- [x] fixes #11867
- [x] fixes #11870

### DIFF
--- a/src/dune_engine/process.mli
+++ b/src/dune_engine/process.mli
@@ -163,6 +163,16 @@ val run_capture_zero_separated
   -> string list
   -> 'a Fiber.t
 
+(** [run_external_in_out] differs from the other [run] functions in the
+    followings ways:
+
+    - The process group ID is inherited by the parent process rather than
+    creating a new one.
+
+    - The input and output file descriptors are the standard ones.
+
+    This version is intended for running external processes at the end of a
+    build such as the ones spawned with "dune exec". *)
 val run_external_in_out
   :  ?dir:Path.t
   -> ?env:Env.t
--- a/src/dune_engine/process.mli
+++ b/src/dune_engine/process.mli
@@ -163,7 +163,7 @@ val run_capture_zero_separated
   -> string list
   -> 'a Fiber.t
 
-(** [run_external_in_out] differs from the other [run] functions in the
+(** [run_inherit_std_in_out] differs from the other [run] functions in the
     followings ways:
 
     - The process group ID is inherited by the parent process rather than
@@ -173,7 +173,7 @@ val run_capture_zero_separated
 
     This version is intended for running external processes at the end of a
     build such as the ones spawned with "dune exec". *)
-val run_external_in_out
+val run_inherit_std_in_out
   :  ?dir:Path.t
   -> ?env:Env.t
   -> Path.t
--- a/bin/exec.ml
+++ b/bin/exec.ml
@@ -154,7 +154,7 @@ let step ~setup ~prog ~args ~common ~no_rebuild ~context ~on_exit () =
     get_path_and_build_if_necessary sctx ~no_rebuild ~dir ~prog
   and* args = Memo.parallel_map args ~f:expand in
   Memo.of_non_reproducible_fiber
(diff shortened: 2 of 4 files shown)


rgrinberg (approved):

src/dune_engine/process.mli
```diff
@@ -163,6 +163,16 @@ val run_capture_zero_separated
...
+    followings ways:
+
+    - The process group ID is inherited by the parent process rather than
+    creating a new one.
+
+    - The input and output file descriptors are the standard ones.
+
+    This version is intended for running external processes at the end of a
+    build such as the ones spawned with "dune exec". *)
 val run_external_in_out
```
rgrinberg: The name is rather mysterious. How about something like `run_inherit_std_in_out`?

PR #11878, opened 2025-06-04
pkg: simpler default platforms for solver

gridbugs:
Simplify the default set of platforms dune will try to solve projects for. This speeds up solving in the common case since the solver will run fewer times, and hopefully generates more general solutions since the distribution and architecture are never constrained by default. If this leads to problems with common packages on common platforms we can always add more cases to the default so the common experience using Dune is smooth.


Leonidas-from-XIV (approved):
I am surprised this can be done in just so few changes, but nice; let's give it a go.

maiste (approved):
Thanks for this change! LGTM :heavy_check_mark:

PR #11879, opened 2025-06-04
fix: revert dune exec back to spawn

### LINKED ISSUE #11867: dune exec seems to be hanging with 3.19

### LINKED ISSUE #11870: Fails to pass input to program via stdin when launching program with "dune exec"

### LINKED ISSUE #11881: Undocumented changes in dune exec in 3.19.0 regarding /tmp and Sys.argv.(0)
## Expected Behavior

I would expect `dune exec ./main.exe` to be equivalent to `dune build ./main.exe && _build/default/main.exe` (assuming `main.ml` is at the root).

(`dune exec --help` says that it should actually be equivalent to `dune install; COMMAND`, which is close enough.)

## Actual Behavior

- `Filename.get_temp_file_name` returns something like `/tmp/build_596b2c_dune` instead of `/tmp`.
- Moreover, this temporary directory seems to be deleted after the process exits. This is a breaking change for the `--keep-temp` flag of [Tezt](https://gitlab.com/nomadic-labs/tezt/).
- `Sys.argv.(0)` is an absolute path (it used to typically be relative, starting with `_build`). This [broke Tezt's cram tests](https://gitlab.com/nomadic-labs/tezt/-/merge_requests/80) in a way that is hard to fix for both Dune 3.19.0 and Dune 3.18.2.
- Some other environment variables are modified (but this was already the case before 3.19.0 so this is out of scope for this issue — in case you're curious, this breaks [this script](https://gitlab.com/tezos/tezos/-/merge_requests/17870) in some cases).

I would suggest to at least document those changes in the changelog. The absolute path is probably not worth a fix, but the behavior of the temporary directory is puzzling.

Maybe  https://github.com/ocaml/dune/commit/260320266348e1d11ed28ccc9dfab13c85916d22 fixes some of the issues, but I failed to compile Dune (I tried something like `./configure && make bootstrap && make install`, the latter step failed because of a lack of `.install`; and `make release` failed with an uncaught `Not_found`).

## Reproduction

`main.ml`:
```
let () =
  Printf.printf "argv0 = %S\n%!" Sys.argv.(0);
  let tmp = Filename.get_temp_dir_name () in
  Printf.printf "tmp = %S\n%!" tmp;
  let ch = open_out (Filename.concat tmp "dune-test") in
  close_out ch
```
`dune`:
```
(executable (name main))
```
`dune-project`:
```
(lang dune 3.19)
```
Then run:
```
dune exec ./main.exe
ls /tmp/build*dune
```
For instance, I got:
```
cd /tmp/abc/ && dune exec ./main.exe && ls /tmp/build*dune
argv0 = "/tmp/abc/_build/default/main.exe"
tmp = "/tmp/build_596b2c_dune"
zsh:1: no matches found: /tmp/build*dune
```

## Specifications

- Version of `dune` (output of `dune --version`): 3.19.0
- Version of `ocaml` (output of `ocamlc --version`): 5.2.1
- Operating system (distribution and version): Debian testing

## Additional information 

- Link to gist with verbose output (run `dune` with the `--verbose` flag): (none)

Alizter:
This reverts `dune exec` to spawn after essentially exiting dune like we had before 3.19.

- [x] update changelog
- [x] add changelog
- [x] fixes https://github.com/ocaml/dune/issues/11881
- [x] fixes #11867
- [x] fixes #11870

I've overridden the previous "fix"'s changelog from #11876.

### DIFF
--- a/test/blackbox-tests/test-cases/exec/exec-bin.t
+++ b/test/blackbox-tests/test-cases/exec/exec-bin.t
@@ -23,7 +23,7 @@ By default, e is executed with the program name and arguments in argv.
 
   $ dune exec ./e.exe a b c
   Hello
-  argv[0] = $TESTCASE_ROOT/_build/default/e.exe
+  argv[0] = ./_build/default/e.exe
   argv[1] = a
   argv[2] = b
   argv[3] = c
@@ -32,7 +32,7 @@ The special form %{bin:public_name} is supported.
 
   $ dune exec %{bin:e} a b c
   Hello
-  argv[0] = $TESTCASE_ROOT/_build/install/default/bin/e
+  argv[0] = ./_build/install/default/bin/e
   argv[1] = a
   argv[2] = b
   argv[3] = c
--- a/bin/exec.ml
+++ b/bin/exec.ml
@@ -193,7 +193,20 @@ let term : unit Term.t =
     @@ fun () ->
     let open Fiber.O in
     let* setup = Import.Main.setup () in
-    build_exn @@ step ~setup ~prog ~args ~common ~no_rebuild ~context ~on_exit:exit
+    build_exn (fun () ->
+      let open Memo.O in
+      let* sctx = setup >>| Import.Main.find_scontext_exn ~name:context in
+      let* env = Super_context.context_env sctx in
+      let expand = Cmd_arg.expand ~root:(Common.root common) ~sctx in
+      let* prog =
+        let dir =
+          let context = Dune_rules.Super_context.context sctx in
+          Path.Build.relative (Context.build_dir context) (Common.prefix_target common "")
+        in
+        let* prog = expand prog in
+        get_path_and_build_if_necessary sctx ~no_rebuild ~dir ~prog >>| Path.to_string
+      and* args = Memo.parallel_map ~f:expand args in
+      restore_cwd_and_execve (Common.root common) prog args env)
 ;;
 
 let command = Cmd.v info term
--- a/doc/changes/11876.md
+++ b/doc/changes/11876.md
@@ -1,2 +1,0 @@
-- Fix `dune exec`'s ability to read stdin interactively. (#11876,
-  fixes #11870 and #11867, @Alizter)
--- a/doc/changes/11879.md
+++ b/doc/changes/11879.md
@@ -1,0 +1,3 @@
+- Revert changes in `dune exec` behaviour introduced in 3.19.0. (#11879, fixes
+  #11870, #11867 and #11881, @Alizter)
+
--- a/otherlibs/dune-site/test/run_2_9.t
+++ b/otherlibs/dune-site/test/run_2_9.t
@@ -343,6 +343,7 @@ Test compiling an external plugin
 
   $ OCAMLPATH=$PWD/_install/lib:$OCAMLPATH PATH=$PWD/_install/bin:$PATH dune exec  --root=e -- c
   Entering directory 'e'
+  Leaving directory 'e'
   run a
   a: $TESTCASE_ROOT/_install/share/a/data
   run c: a linked registered:.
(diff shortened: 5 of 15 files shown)


test/blackbox-tests/test-cases/exec/exec-bin.t
```diff
@@ -23,7 +23,7 @@ By default, e is executed with the program name and arguments in argv.
 
   $ dune exec ./e.exe a b c
   Hello
-  argv[0] = $TESTCASE_ROOT/_build/default/e.exe
```
maiste, line 26: Is it expected that `argv[0] changes this way?

Alizter: Yes, my change from before changed it to an absolute path so this reverts it back to a relative one as before. I don't recall the reason why it changes.

maiste: Just to make sure, wasn't it the fix for `$PWD` + `argv[0]` = `path to executable`?

Alizter: No, that was an unrelated fix.

maiste (approved):
Comparing with the PR to revert, LGTM :+1:

rgrinberg (approved):

PR #11882, opened 2025-06-06
Disable binary dev tools by default in the preview build

Leonidas-from-XIV:
It seems to continue to create issues, so disabling it by default might for now be the way forward.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -57,7 +57,6 @@
           [
             "--pkg-build-progress" "enable"
             "--lock-dev-tool" "enable"
-            "--bin-dev-tools" "enable"
             "--portable-lock-dir" "enable"
           ];
       };


maiste (approved):

Alizter:
What are the issues? I've been using it fine outside of dune.

rikusilvola:
On some systems, the binary distribution of OCaml-LSP will mismatch the compiler used to build the project, leading to `ocamllsp: Corrupted compiled interface` error. Superficially the binary matches the compiler, but for some reason the magic numbers don't match.

Alizter:
It's fine to disable this in the flake then. I would only urge that we document the issue some place so we don't forget about it.

gridbugs (approved):

PR #11884, opened 2025-06-06
chore(version): bump to 3.20 (latest)

maiste:
This PR should be rebased on #11883 before being merged. With the previous PR, it removes the problem with using the latest version inside of Dune (it only requires to rebuild bootstrapped dune).
It first bumps the version to the latest (3.20) and apply the new formatter. If having the latest version inside of the `dune-project` file is OK, I will also update the information in the release documentation to not forget to update the value on each release.

### DIFF
--- a/bench/dune
+++ b/bench/dune
@@ -1,15 +1,8 @@
 (executable
  (name bench)
  (modules bench metrics)
- (libraries
-  dune_stats
-  dune_console
-  chrome_trace
-  stdune
-  fiber
-  dune_lang
-  dune_engine
-  dune_util))
+ (libraries dune_stats dune_console chrome_trace stdune fiber dune_lang
+  dune_engine dune_util))
 
 (rule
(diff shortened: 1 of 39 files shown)


rgrinberg (approved):

maiste:
I still need to update the OCaml benchmarks to support this. The current benchmark seems to download dune from Opam.

maiste:
@rgrinberg is it still good for you? I have also tweaked the benchmark to use the local dune to build the CI. It avoid the language version problem too.

@gridbugs do the modifications to the benchmark Dockerfile look good for you? It seems that you setup it in the past.

rgrinberg (approved):

gridbugs:
> do the modifications to the benchmark Dockerfile look good for you? It seems that you setup it in the past

Looks ok. If anything breaks as a result I'll fix it.

PR #11892, opened 2025-06-09
chore(ci): bump msvc, mingw to 5.3 and reorganise ci matrix

Alizter:
In this PR we:
1. Bump msvc to OCaml 5.3
2. Bump mingw to OCaml 5.3
3. Reorganise our CI matrix a bit so it is clearer which jobs we have exactly.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -51,7 +51,7 @@ jobs:
           - ocaml-compiler: ocaml-compiler.5.3.0,system-msvc
             os: windows-latest
           # mingw
-          - ocaml-compiler: ocaml-base-compiler.5.2.1,system-mingw
+          - ocaml-compiler: ocaml-base-compiler.5.3.0,system-mingw
             os: windows-latest
           - ocaml-compiler: 4.08.x
             os: ubuntu-latest
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -21,38 +21,44 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os:
-          - macos-latest
-          - ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow.yml
```diff
@@ -21,38 +21,44 @@ jobs:
...
-            skip_test: true
-          # macOS x86_64 (Intel)
+          ## macos (x86)
           - ocaml-compiler: 5.3.x
             os: macos-13
             skip_test: true
-          # MSVC (left behind until we upgrade to 5.2.0)
-          - ocaml-compiler: ocaml-base-compiler.4.14.2,system-msvc
+          ## MSVC
+          - ocaml-compiler: ocaml-compiler.5.3.0,system-msvc
```
Alizter, line 46: There appears to be a bug in opam repository which means we need to use `ocaml-compiler` rather than `ocaml-base-compiler`.

Leonidas-from-XIV: Is it a bug? To me it seems that opam-repository deliberately changed the name for 5.3. We had to adjust for this in the package management toolchain code if I remember correctly.

Alizter: Maybe so, but I found it poorly documented.

maiste (approved):
Have you run the CI on this branch just to check everything is fine?

Alizter:
@maiste Our CI is being run in every pull request, so this appears to be working. I also observed the updated jobs when they first run and they took longer because the compiler wasn't yet cached. AFAICT everything is in order.

Alizter:
FTR, the job matrix we had before is basically pointless since we only generated two different combinations. macos-latest and ubuntu-latest for OCaml 5. It's much clearer to use `include:` and list those jobs explicitly.

PR #11900, opened 2025-06-12
Allow running tests concurrently with watch server

gridbugs:


bin/rpc/runtest_rpc.ml
```diff
@@ -0,0 +1,14 @@
+open! Import
```
Leonidas-from-XIV, line 1: Given this is in a folder called `rpc` and there is a file to match the `build` command, I think it would make more sense if this was called `runtest` to match.

bin/build_cmd.ml
```diff
@@ -92,11 +92,15 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
   in
   Scheduler.Run.poll_passive
     ~get_build_request:
-      (let+ (Build (targets, ivar)) = Dune_rpc_impl.Server.pending_build_action rpc in
+      (let+ pending_action = Dune_rpc_impl.Server.pending_action rpc in
```
Leonidas-from-XIV: It could be nicer to just directly match here on `{outcome; kind}` to mirror the previous code and allow the compiler to complain on unused fields.

bin/rpc/rpc_common.ml
```diff
@@ -36,6 +36,51 @@ let client_term builder f =
   Scheduler.go_with_rpc_server ~common ~config f
 ;;
 
+let retry_loop once =
```
Leonidas-from-XIV, line 39: Maybe this loop should have another abort condition? We probably don't want it to loop forever as this might block other software waiting on `dune runtest` forever if the connection can't be made.

bin/rpc/rpc_common.ml
```diff
@@ -36,6 +36,51 @@ let client_term builder f =
...
+  else
+    let+ res = once () in
+    match res with
+    | Some conn -> conn
+    | None ->
+      let (_ : Dune_rpc_private.Where.t) = active_server () in
+      User_error.raise
+        [ Pp.text "failed to establish connection even though server seems to be running"
+        ]
+;;
```
Leonidas-from-XIV, line 65: I think this would be kind of better if there were `establish_connection_with_retry` and `establish_connection_exn`. This way it is these two functions joined with an `if`. It would be more readable to dispatch in the caller.

test/blackbox-tests/test-cases/watching/watching-eager-concurrent-runtest-command.t
```diff
@@ -0,0 +1,34 @@
...
+  $ dune build
+  $ dune build --watch &
+  Success, waiting for filesystem changes...
+  Success, waiting for filesystem changes...
+  Success, waiting for filesystem changes...
+  Hello, World!
+
+Test that we can run a test while another instance of dune is running in watch
+mode:
+  $ dune runtest 2>&1 | sed 's/pid: [0-9]*/pid: PID/g'
```
Leonidas-from-XIV, line 23: I think that `sed` is unnecessary given we expect to succeed?

test/blackbox-tests/test-cases/watching/watching-eager-concurrent-runtest-command.t
```diff
@@ -0,0 +1,34 @@
...
+  Success
+
+Test that passing extra arguments to `dune runtest` prints a warning when
+running concurrently with another instance of dune in watch mode:
+  $ dune runtest --auto-promote 2>&1 | sed 's/pid: [0-9]*/pid: PID/g'
+  Warning: Your build request is being forwarded to a running Dune instance
+  (pid: PID) so most command-line arguments will be ignored.
+  Success
```
Leonidas-from-XIV, line 31: Could you add a test that fails so we know that a failure is also correctly reported back?

src/dune_rpc_impl/server.ml
```diff
@@ -329,25 +335,33 @@ let handler (t : _ t Fdecl.t) handle : 'build_arg Dune_rpc_server.Handler.t =
     Handler.declare_notification rpc Procedures.Server_side.log
   in
   let () = Handler.implement_request rpc Procedures.Public.ping (fun _ -> Fiber.return) in
-  let () =
-    let build _session targets =
+  let implement_request_pending_action decl ~f =
```
Leonidas-from-XIV, line 338: `implement`? Looking at the code from my understanding I think `register` would better describe what it does.

src/dune_rpc_impl/server.ml
```diff
@@ -329,25 +335,33 @@ let handler (t : _ t Fdecl.t) handle : 'build_arg Dune_rpc_server.Handler.t =
...
+      let* () = Job_queue.write server.pending_jobs { kind = f input; outcome } in
+      let+ build_outcome = Fiber.Ivar.read outcome in
       match (build_outcome : Build_outcome.t) with
       | Success -> Build_outcome_with_diagnostics.Success
       | Failure -> Failure (get_current_diagnostic_errors ())
     in
-    Handler.implement_request rpc Decl.build build
+    Handler.implement_request rpc decl handler
+  in
+  let () =
```
Leonidas-from-XIV, line 350: Maybe just use `;`, to show that it is side-effectful and not expected to return any value.

src/dune_rpc_impl/server.ml
```diff
@@ -329,25 +335,33 @@ let handler (t : _ t Fdecl.t) handle : 'build_arg Dune_rpc_server.Handler.t =
...
+    implement_request_pending_action Decl.runtest ~f:(fun paths -> Runtest paths)
   in
   let () =
     let rec cancel_pending_jobs () =
-      match Job_queue.pop_internal (Fdecl.get t).pending_build_jobs with
+      match Job_queue.pop_internal (Fdecl.get t).pending_jobs with
       | None -> Fiber.return ()
-      | Some (_, job) ->
-        let* () = Fiber.Ivar.fill job Build_outcome.Failure in
+      | Some pending_action ->
```
Leonidas-from-XIV, line 363: Match on just `{outcome; kind=_}` to give the compiler chances to complain. Or `{outcome; _}` to let the user know we won't use other values.

bin/rpc/runtest_rpc.ml
gridbugs: Folders don't act like namespaces in dune, so naming this `runtest.ml` would cause it to collide with the `runtest.ml` in the parent directory. It does bother me that the naming is inconsistent though. I'll change that in a separate PR. In the meantime I'll still rename this to `rpc_runtest.ml` for consistency with `rpc_common.ml`.

gridbugs: I think this is the reason `build` is implemented in `build_cmd.ml` rather than `build.ml`. The `build.ml` file is in the rpc directory and they would collide.

src/dune_rpc_impl/server.ml
gridbugs: I agree but I'm following the naming convention already established in this file so I think it's better to keep calling it `implement`.

src/dune_rpc_impl/server.ml
gridbugs: Again I agree with the sentiment but there's an established pattern of doing `let () = ...` in this file.

bin/rpc/runtest_rpc.ml
Leonidas-from-XIV: Argh that's very annoying. It would be nice if we could enable the default behavior of Dune in the Dune codebase, but admittedly that is outside of the scope of this PR.

bin/rpc/rpc_common.ml
gridbugs: I agree that would be nice. I had a go implementing this (specifically having this function timeout after some period) and it's quite straightforward however for such a feature I'd want to add a cram test, and doing so would require adding an argument to various commands to allow the user to set the timeout, or hardcoding a specific non-infinite timeout into the code. The former would be too big a change for this PR I think, and the latter would change the behaviour of dune which I'm hesitant to do.

I'll note that the code in this file was mostly just moved from `bin/rpc/build.ml` and not written specifically for this PR.

ElectreAAS:
Reading your work in `dune_rpc_impl/server.ml`, I get the impression that this PR doesn't use the fact that `@runtest` is a **build** target, and that `dune build @runtest` will send a **build** request to the rpc server with that particular target, is that correct?

cc @shonfeder you were interested in the answer as well

_For context, I am asking this question to better implement #11958_

gridbugs:
> Reading your work in `dune_rpc_impl/server.ml`, I get the impression that this PR doesn't use the fact that `@runtest` is a **build** target, and that `dune build @runtest` will send a **build** request to the rpc server with that particular target, is that correct?
> 
> cc @shonfeder you were interested in the answer as well
> 
> _For context, I am asking this question to better implement #11958_

That's correct. `dune runtest` and `dune build @runtest` are not completely equivalent, as evidenced by the fact that there's additional logic inside the implementation of `dune runtest` command, mostly for handling paths.

That said, there is an oversight which I made that I'll address before merging, which is that I believed that the only way to evaluate a memo is by taking the global lock and running the build system, but I've recently learnt that this is not the case as `Memo.run` returns a fiber that can be scheduled without invoking the build system. I'm not sure if all memos can be safely evaluated in this way since some memos have side effects within _build that mean they should only be evaluated while holding the global lock.

I'll see if I can take advantage of this fact to simplify this PR.

gridbugs:
I spent some time today trying to take advantage of the `@runtest` build target alias, but some of the path processing done by `dune runtest` requires information only available when the build system runs (the list of build contexts). Therefore I think we need to keep this in its current state where `runtest` is a separate rpc method from `build`.

gridbugs:
@ElectreAAS I've added you as a reviewer as I think the other reviewers may be short on time at the moment.

gridbugs:
That failing test is a problem installing deps with apt, so it's unrelated to this change.

Alizter:
@gridbugs I don't think we need a dedicated runtest method for this. `dune runtest` may do some path processing, but at the end of all of that, it only ever builds an alias. At the moment, cram tests are processed to their corresponding aliases, and directories are passed as `runtest` in that directory. This method of searching for and executing tests will be extended in https://github.com/ocaml/dune/pull/11980.

I think what we need is a specific RPC call for getting cram tests in a given source directory. We don't need to know about contexts on our side. If a user wishes to build the alias `_build/mycontext/runtest` we can pass that path directly. If a user wants to build the alias `./runtest`, then the build server will already build this in all contexts.

This cram test finder could be made a bit smarter since we don't even need to return all the tests. We should instead have something like

```ocaml
val find_cram_test
  : dir:Path.Source.t ->
    [ `Filename of string | `Name of string ] ->
    (Cram_test.t * [ `Enabled | `Disabled ]) option Fiber.t
```

That way all `dune runtest` would have to do is process some paths, query tests using `find_cram_test` and then send specific alias requests to build.

I'm a little hesitant to add a dedicated runtest method since it is just a `build` with some searching before, and it will make it trickier to extend in the future. Whereas searching for cram tests is very specific and unlikely to change.

gridbugs:
Thanks @Alizter! Can you clarify why it's necessary to enumerate the cram tests in a directory using an RPC method. Is there a possibility that it will race with the build system in some way? Ideally we could just make it a fiber or memo that resolves entirely within the client, but transforming the current logic doesn't seem straightforward (but it's not immediately clear to me why).

shonfeder:
Let's discuss with @ElectreAAS and @gridbugs how best to move this forward. Ali has communicated that his comments here are not blocking.

ElectreAAS:
Superseded by #12473

PR #11902, opened 2025-06-12
test: mark tests that require melange

rgrinberg:


rgrinberg:
@anmonteiro do you know why the test is failing now? All I did was move it to a different directory.

Also, isn't `str` available in melange? I don't even know how it worked before...

anmonteiro:
I'm not sure but I can take a look during the weekend.

rgrinberg:
I rewrote the test not to use melange. Whatever it was testing was not specific to melange in any way

PR #11905, opened 2025-06-13
Looking up paths to executables with `dune describe location`

gridbugs:
The "dune exec" command has three different ways of resolving the names of executables to paths to executables:
 - public executables defined in the current project
 - executables in the "bin" directories of dependencies
 - executables in directories listed in $PATH

This can lead to unexpected shadowing, especially in the case of executables from dependecies, as users may not be aware that one of the packages in their project's dependency cone defines an executable with the same name of an executable that's also insalled system-wide.

Short of fixing the problem for now, this change introduces a tool for helping investigate specifically which executable will be run by "dune exec". This adds a command "dune which" which prints the path to the executable.

### DIFF
--- a/bin/exec.ml
+++ b/bin/exec.ml
@@ -115,19 +115,28 @@ let build_prog ~no_rebuild ~prog p =
     p
 ;;
 
-let get_path_and_build_if_necessary sctx ~no_rebuild ~dir ~prog =
+let dir_of_context common context =
   let open Memo.O in
+  let+ sctx = Super_context.find_exn context in
+  let context = Dune_rules.Super_context.context sctx in
+  Path.Build.relative (Context.build_dir context) (Common.prefix_target common "")
+;;
+
+let get_path common context ~prog =
+  let open Memo.O in
+  let* sctx = Super_context.find_exn context
+  and* dir = dir_of_context common context in
   match Filename.analyze_program_name prog with
   | In_path ->
     Super_context.resolve_program_memo sctx ~dir ~loc:None prog
     >>= (function
      | Error (_ : Action.Prog.Not_found.t) -> not_found_with_suggestions ~dir ~prog
-     | Ok p -> build_prog ~no_rebuild ~prog p)
+     | Ok p -> Memo.return p)
   | Relative_to_current_dir ->
     let path = Path.relative_to_source_in_build_or_external ~dir prog in
     Build_system.file_exists path
     >>= (function
-     | true -> build_prog ~no_rebuild ~prog path
+     | true -> Memo.return path
      | false -> not_found_with_suggestions ~dir ~prog)
   | Absolute ->
     (match
@@ -144,19 +153,24 @@ let get_path_and_build_if_necessary sctx ~no_rebuild ~dir ~prog =
      | None -> not_found_with_suggestions ~dir ~prog)
 ;;
 
-let step ~setup ~prog ~args ~common ~no_rebuild ~context ~on_exit () =
+let get_path_and_build_if_necessary common context ~no_rebuild ~prog =
   let open Memo.O in
-  let* sctx = setup >>| Import.Main.find_scontext_exn ~name:context in
-  let* env = Super_context.context_env sctx in
-  let expand = Cmd_arg.expand ~root:(Common.root common) ~sctx in
+  let* path = get_path common context ~prog in
+  match Filename.analyze_program_name prog with
+  | In_path | Relative_to_current_dir -> build_prog ~no_rebuild ~prog path
+  | Absolute -> Memo.return path
+;;
+
+let step ~prog ~args ~common ~no_rebuild ~context ~on_exit () =
+  let open Memo.O in
+  let* sctx = Super_context.find_exn context in
   let* path =
-    let dir =
-      let context = Dune_rules.Super_context.context sctx in
-      Path.Build.relative (Context.build_dir context) (Common.prefix_target common "")
-    in
-    let* prog = expand prog in
-    get_path_and_build_if_necessary sctx ~no_rebuild ~dir ~prog
-  and* args = Memo.parallel_map args ~f:expand in
+    let* prog = Cmd_arg.expand ~root:(Common.root common) ~sctx prog in
+    get_path_and_build_if_necessary common context ~no_rebuild ~prog
+  and* args =
+    Memo.parallel_map args ~f:(Cmd_arg.expand ~root:(Common.root common) ~sctx)
+  in
+  let* env = Super_context.context_env sctx in
   Memo.of_non_reproducible_fiber
   @@ Dune_engine.Process.run_inherit_std_in_out
        ~dir:(Path.of_string Fpath.initial_cwd)
@@ -252,12 +266,11 @@ let exec_building_directly ~common ~config ~context ~prog ~args ~no_rebuild =
     Scheduler.go_with_rpc_server_and_console_status_reporting ~common ~config
     @@ fun () ->
     let open Fiber.O in
-    let* setup = Import.Main.setup () in
     let on_exit = Console.printf "Program exited with code [%d]" in
     Scheduler.Run.poll
     @@
     let* () = Fiber.return @@ Scheduler.maybe_clear_screen ~details_hum:[] config in
-    build @@ step ~setup ~prog ~args ~common ~no_rebuild ~context ~on_exit
+    build @@ step ~prog ~args ~common ~no_rebuild ~context ~on_exit
   | No ->
     Scheduler.go_with_rpc_server ~common ~config
     @@ fun () ->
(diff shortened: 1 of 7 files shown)


gridbugs:
Here's an example session using this command:
```
37 s@u3 ~/src/llama (main=)
+ λ dune exec --which ocamlbuild
_build/_private/default/.pkg/ocamlbuild/target/bin/ocamlbuild

38 s@u3 ~/src/llama (main=)
+ λ dune exec --which ocamlc
/home/s/.cache/dune/toolchains/ocaml-compiler.5.3.0-a3dd2a36d935a7ada272441c8543eba1/target/bin/ocamlc

39 s@u3 ~/src/llama (main=)
+ λ dune exec --which cat
/nix/store/87fck6hm17chxjq7badb11mq036zbyv9-coreutils-9.7/bin/cat

40 s@u3 ~/src/llama (main=)
+ λ dune exec --which foo
_build/install/default/bin/foo
```

Alizter:
Is this essentially `dune exec -- which %{bin:foo}`? I'm still for including such a feature, since not every system has `which`.

gridbugs:
It is, yes, but I bet a lot of people don't know you can expand pforms like that (for example I did not know this).

bin/exec.ml
```diff
@@ -111,8 +111,15 @@ let build_prog ~no_rebuild ~prog p =
     p
 ;;
 
-let get_path_and_build_if_necessary sctx ~no_rebuild ~dir ~prog =
+let sctx_of_context context =
   let open Memo.O in
+  let* setup = Import.Main.setup () |> Memo.of_reproducible_fiber in
+  setup >>| Import.Main.find_scontext_exn ~name:context
```
Leonidas-from-XIV: I wonder, can you get the `Context_name.t` of a `context`? Then it could be possible to retrieve the super context via `Super_context.find_exn <Context_name.t>`

test/blackbox-tests/test-cases/exec/exec-which.t
```diff
@@ -0,0 +1,62 @@
+Exercise the various ways of resolving executable names with `dune exec`.
+
+  $ cat > dune-project << EOF
+  > (lang dune 3.20)
+  > 
+  > (package
+  >  (name foo))
```
Leonidas-from-XIV: Can't `(package (name foo))` not be shortened to `(name foo)`? Or even left out?

test/blackbox-tests/test-cases/exec/exec-which.t
```diff
@@ -0,0 +1,62 @@
...
+Test that executables from dependencies are located correctly:
+  $ mkdir dune.lock
+  $ cat > dune.lock/lock.dune << EOF
+  > (lang package 0.1)
+  > EOF
+  $ cat > dune.lock/bar.pkg << EOF
+  > (version 0.1)
+  > (install
+  >  (progn
+  >   (run sh -c "echo '#!/bin/sh\necho hello bar' > %{bin}/bar")
```
Leonidas-from-XIV: Is `write-file` unsupported here? I know OPAM files can never generate it, but can we execute that action here?

Leonidas-from-XIV (commented):
I think it is a very cool feature. In fact, I would actually suggest making a `dune which` top level command, as `dune exec --which` is a flag I will immediately forget and it is fairly obscure and does not actually execute the command.

test/blackbox-tests/test-cases/exec/exec-which.t
Alizter: It can, but its kind of a legacy thing where 1 dune-project corresponded to 1 dune package. Probably best to be more explicit.

test/blackbox-tests/test-cases/exec/exec-which.t
gridbugs: Good point, it seems to work.

bin/exec.ml
gridbugs: Turns out I can replace this entire function with `Super_context.find_exn`.

gridbugs:
Updated this so that the command is now `dune which <name>`.

bin/exec.ml
```diff
@@ -111,19 +111,34 @@ let build_prog ~no_rebuild ~prog p =
     p
 ;;
 
-let get_path_and_build_if_necessary sctx ~no_rebuild ~dir ~prog =
+let dir_of_context common context =
```
Leonidas-from-XIV: Maybe call context `ctx_name` because it's a `Context_name.t` and what is later called `context` is `Context.t`. Reading the Dune code this keeps being confusing, so for reviewability sake I'd prefer being more descriptive with names.

bin/exec.ml
```diff
@@ -111,19 +111,34 @@ let build_prog ~no_rebuild ~prog p =
...
   let open Memo.O in
+  let+ sctx = Super_context.find_exn context in
+  let context = Dune_rules.Super_context.context sctx in
+  Path.Build.relative (Context.build_dir context) (Common.prefix_target common "")
+;;
+
+let expand common context prog =
+  let open Memo.O in
+  let* sctx = Super_context.find_exn context in
```
Leonidas-from-XIV: This function is called in 4 places or so, maybe it would make more sense to do it in one place and pass `sctx` as argument instead? `exec.ml:215` already calculates it, so might as well pass it on.

doc/changes/11905.md
```diff
@@ -0,0 +1,2 @@
+- Add `--which` flag to `dune exec` for printing the path to the executable
+  that would be run (#11905, @gridbugs)
```
Leonidas-from-XIV: This is nice but needs to be updated now.

test/blackbox-tests/test-cases/which.t
```diff
@@ -0,0 +1,62 @@
+Exercise the various ways of resolving executable names with `dune exec`.
```
Leonidas-from-XIV:
```suggestion
Exercise the various ways of resolving executable names with `dune which`.
```

Leonidas-from-XIV (commented):
Except for some minor cleanup I think this is nearly done. I really like the cram test!

bin/which.ml
```diff
@@ -0,0 +1,37 @@
+open! Import
```
Alizter:
```suggestion
open Import
```

bin/which.mli
```diff
@@ -0,0 +1,3 @@
+open! Import
```
Alizter:
```suggestion
open Import
```

bin/which.ml
```diff
@@ -0,0 +1,37 @@
...
+  and+ context = Common.context_arg ~doc:{|Run the command in this build context.|}
+  and+ prog = Arg.(required & pos 0 (some string) None (Arg.info [] ~docv:"PROG")) in
+  let common, config = Common.init builder in
+  Scheduler.go_with_rpc_server ~common ~config
+  @@ fun () ->
+  build_exn
+  @@ fun () ->
+  let open Memo.O in
+  let+ path = Exec.get_path common context ~prog >>| Path.to_string in
+  Dune_console.printf "%s" path
```
Alizter:
```suggestion
  Console.print [ Pp.verbatim path ]
```

Alizter (approved):

bin/which.ml
Alizter: I think its better to be explicit that this is verbatim rather than rely on whatever `Console.printf` is doing at this point in time. We should only really be using that for debugging.

rgrinberg:
Maybe this should go under `describe`/`show`? Don't think this command is going to be used enough to live at the top level tbh.

gridbugs:
Why is the expected frequency with which a command will be used a useful metric for deciding whether to put it at the top level or not? I think it belongs at the top level because I expect this command to be used by people debugging a situation where dune ran a command they didn't expect, and for this use case it's important that the command be easy to discover.

bin/exec.ml
gridbugs: Good point. Updated.

rgrinberg:
There are other commands under `$ dune describe` that are no less useful for debugging, shall we promote them all to the top level as well?

The point of having a hierarchical command line is to make the commands more discoverable. If putting things under sub-commands made the CLI less discoverable, why did we introduce the hierarchy in the first place? Dune is not like git where sub-commands are their own binaries, so it was done for purely organizational purposes.

IMO, the underlying issue is that the "discoverability" of one command comes at the expense of other commands. So yes, you have made `$ dune which` more discoverable, but at the cost of making the user slightly more likely gloss over more useful sub-commands.

Now it could be that I'm underestimating the utility of this command, and it is going to be more useful than I'm imagning - I'll leave the decision where to put it up to you. But I do want to clarify that:

> Why is the expected frequency with which a command will be used a useful metric for deciding whether to put it at the top level or not?

Is the principle we should be following. We hide scarcely used commands, and promote the ones that are often used. We group related sub-commands to avoid repeating prefixes. We optimize for the common case.

test/blackbox-tests/test-cases/which.t
```diff
@@ -0,0 +1,62 @@
...
+  $ mkdir bin
+  $ cat > bin/baz << EOF
+  > #!/bin/sh
+  > echo hello baz
+  > EOF
+
+  $ chmod a+x bin/baz
+  $ export PATH=$PWD/bin:$PATH
+
+  $ dune which baz
```
rgrinberg: Could you add a test for `dune which %{..}`? I think it should support this.

Leonidas-from-XIV:
For what it's worth, something like `dune [describe/show] location` would also sound fine with me and `dune [describe/show]` is reasonably discoverable as it has existed for a while. My concern was that `dune exec --option-that-does-not-exec` is quite hard to notice and also remember, so I suggested `dune which` to reuse the familiar term from POSIX.

test/blackbox-tests/test-cases/which.t
gridbugs: Done.

gridbugs:
Updated so that the command is now `dune describe location`.

bin/exec.ml
Leonidas-from-XIV: Are you sure that it has been updated? I still see `sctx` being calculated multiple times everywhere. Maybe you forgot to commit?

gridbugs: You're right. My bad, I must have forgotten to check it in. I've ended up just replacing calls to this function with `Cmd_arg.expand` since the `sctx` was already computed in all relevant places.

Leonidas-from-XIV (approved):

PR #11913, opened 2025-06-15
boot: include job number in command line output

Alizter:
We include the job number in the command-line output which can make it easier to read and debug the logs from bootstrapping.

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -508,7 +508,7 @@ end = struct
         let cmdline = String.concat ~sep:" " (prog :: args) in
         let cmdline =
           match cwd with
-          | Some x -> sprintf "cd %s && %s" x cmdline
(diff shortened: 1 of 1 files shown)


boot/duneboot.ml
```diff
@@ -508,7 +508,7 @@ end = struct
         let cmdline = String.concat ~sep:" " (prog :: args) in
         let cmdline =
           match cwd with
-          | Some x -> sprintf "cd %s && %s" x cmdline
+          | Some x -> sprintf "[%d] cd %s && %s" !Status_line.num_jobs_finished x cmdline
```
rgrinberg, line 511: This makes it impossible to copy the commands printed to run them

Alizter: That's fair enough.

PR #11916, opened 2025-06-15
refactor: assert that we never generate dep files for aliases

rgrinberg:


rgrinberg:
@anmonteiro do you mind taking a look at this? Somehow this is failing for virtual libraries under melange. I'm not really sure what's happening here, but we should never be running `ocamldep` on the alias module, so there should be no dependency file to copy at all.

EDIT: Fixed the issue

PR #11922, opened 2025-06-17
Dev tool which command fails if tool not installed

gridbugs:
Previously `dune tool which` would print the path to where a dev tool would be installed, regardless of whether it is currently installed. This is inconsistent with how the `which` unix command works. This commit changes its default behaviour to print an error and exit with non-zero if the specified tool is not installed, allowing the command to be used to check whether a given tool is currently inalled and bring it into line with how `which` commands conventionally behave.

The original behaviour is preserved with the `--allow-not-installed` flag.

I've also taken the opportunity to unify the implementations of `which` across dev tools.

CC @pitag-ha


test/blackbox-tests/test-cases/pkg/ocamlformat/ocamlformat-which.t
```diff
@@ -0,0 +1,28 @@
...
+  _build/_private/default/.dev-tool/ocamlformat/ocamlformat/target/bin/ocamlformat
+
+  $ dune tools exec ocamlformat
+  Solution for dev-tools.locks/ocamlformat:
+  - ocamlformat.0.26.2
+       Running 'ocamlformat'
+  formatted with version 0.26.2
+
+  $ dune tools which ocamlformat
+  _build/_private/default/.dev-tool/ocamlformat/ocamlformat/target/bin/ocamlformat
```
Leonidas-from-XIV, line 31: Can you add a check that the file actually exists there?

test/blackbox-tests/test-cases/pkg/ocamllsp/dev-tool-ocamllsp-which.t
```diff
@@ -0,0 +1,43 @@
+Test `dune tools which ocamllsp`:
```
Leonidas-from-XIV: I am not sure this test adds anything over the ocamlformat test. If it works for ocamlformat, I don't see why the behavior for ocamllsp would be any different?

bin/tools/tools_common.ml
```diff
@@ -54,3 +54,36 @@ let lock_build_and_run_dev_tool ~common ~config dev_tool ~args =
...
+    in
+    let _ : Common.t * Dune_config_file.Dune_config.t = Common.init builder in
+    if allow_not_installed || Path.exists exe_path
+    then print_endline (Path.to_string exe_path)
+    else User_error.raise [ Pp.textf "%s is not installed as a dev tool" exe_name ]
+  in
+  let info =
+    let doc =
+      sprintf
+        "Prints the path to the %s dev tool executable if it exists, otherwise error."
```
Leonidas-from-XIV:
```suggestion
        "Prints the path to the %s dev tool executable if it exists, errors out otherwise."
```

this seems to me somewhat more correct grammarwise.

bin/tools/ocamllsp.ml
```diff
@@ -37,16 +36,5 @@ module Exec = struct
 end
 
 module Which = struct
```
Leonidas-from-XIV: It feels to me that each tool having its own `Which` is rather inconvenient (and requires additional boilerplate at little advantage, given all of them have the same implementation anyway). `Which` basically just uses `Tools_common.which_command` which then uses `Pkg_dev_tool.exe_name` so there's a lot of indirection for little use. Maybe better move it into the implementation of the `tools which` command, then there is no need to test each `bin/tools/*/Which` module.

Leonidas-from-XIV (commented):
I think the change in semantics is fairly reasonable, matching the familiar POSIX semantics with an escape hatch.

test/blackbox-tests/test-cases/pkg/ocamlformat/ocamlformat-which.t
gridbugs: Good idea. Added.

test/blackbox-tests/test-cases/pkg/ocamllsp/dev-tool-ocamllsp-which.t
gridbugs: Fair point. I removed the ocamllsp test and just left the ocamlformat one.

bin/tools/ocamllsp.ml
gridbugs: Ah yes good point. Removed the indirection.

Leonidas-from-XIV (approved):
Looks very good to me and it makes the code a bit more understandable as well.

Sudha247 (commented):
Agree with following the Unix semantics, the change looks good to me too! Having a flag to allow old behaviour is useful.

PR #11929, opened 2025-06-18
chore: Bump ocaml version used in tests

ElectreAAS:
The ocaml version used in tests at the moment seems in-between releases as
- The Makefile says `5.2.1`
- The documentation says `5.1.1`
- The output of the tests suggests `5.3.x`:
Running the test suite with `5.2.1` yields numerous errors like this
```diff
- The constant 123 has type...
+ This expression has type...
```
These errors disappear when running with `5.3.0`, so in this PR I am bumping the version to that.

References:
- in #11708 the version in the Makefile was updated to `5.2.1`
- in #11788 the version actually used by tests was updated to `5.3.x`
- [Here](https://discuss.ocaml.org/t/ocaml-5-3-0-released/15916#p-67819-compiler-user-interface-and-warnings-14) is the discuss announcement of the release of `5.3.0`, which has a formatted changelog. The link above points directly to the changes in error messages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,8 @@ DEV_DEPS := \
 core_bench \
 patdiff
 
-TEST_OCAMLVERSION := 5.2.1
+TEST_OCAMLVERSION := 5.3.0
+# When updating this version, don't forget to also bump the number in the docs.
 
 -include Makefile.dev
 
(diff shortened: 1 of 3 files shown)


Alizter (approved):
LGTM, we can merge once @maiste confirms.

maiste (approved):
LGTM :+1: 
Small note but to make the `make dev-switch` working, it also requires https://github.com/ocaml/dune/pull/11928 to be merged along with this one.

Leonidas-from-XIV (approved):
Makes sense to me. This is also the version I am successfully running locally these days.

PR #11933, opened 2025-06-21
Portable lockdirs respects custom solver_env

gridbugs:
Previously when using portable lockdirs, custom solver variables set in the "solver_env" field of the "lock_dir" stanza in dune-workspace would be ignored. After this change the custom variables are added to the solver env, though they are overriden by settings from the "solve_for_platforms" field.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/portable-lockdirs/portable-lockdirs-custom-solver-env.t
+++ b/test/blackbox-tests/test-cases/pkg/portable-lockdirs/portable-lockdirs-custom-solver-env.t
@@ -1,0 +1,75 @@
+Exercise solving with portable lockdirs when there is a custom solver
+environment that affects the solution.
+
+  $ . ../helpers.sh
+  $ mkrepo
+  $ add_mock_repo_if_needed
+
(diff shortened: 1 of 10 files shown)


test/blackbox-tests/test-cases/pkg/portable-lockdirs/portable-lockdirs-custom-solver-env.t
```diff
@@ -0,0 +1,76 @@
...
+  > (lang dune 3.8)
+  > (repository
+  >  (name mock)
+  >  (url "file://$(pwd)/mock-opam-repository"))
+  > (lock_dir
+  >  (path dune.lock)
+  >  (repositories mock)
+  >  (solver_env
+  >   (sys-ocaml-version 5.4.0+solver-env-version-override)
+  >   ))
```
ElectreAAS: > \> ))

:angry: :exclamation:

ElectreAAS (approved):
Aside from my asinine comment about parentheses, this is good to go!
Thanks :)

PR #11934, opened 2025-06-21
Improve dune lang version not supported error message

Alizter:
In this PR we do two things:

1. Fix a bug where the wrong dune lang version was being printed when using an older dune. For instance we used to print:

```
File "dune-project", line 1, characters 11-18:
  1 | (lang dune 123.123)
                 ^^^^^^^
  Error: Version 123.123 of the dune language is not supported.
  Supported versions of this extension in version 123.123 of the dune language:
  - 1.0 to 1.12
  - 2.0 to 2.9
  - 3.0 to 3.20
```

But we are not in version `123.123`!

We fix this by passing the current supported version of dune lang instead of the one we just parsed, allowing the error message to become:

```
  File "dune-project", line 1, characters 11-18:
  1 | (lang dune 123.123)
                 ^^^^^^^
  Error: Version 123.123 of the dune language is not supported.
  Supported versions of this extension in version 3.20 of the dune language:
  - 1.0 to 1.12
  - 2.0 to 2.9
  - 3.0 to 3.20
```

2. In our second commit, we improve the error message more generally. We have access to a bit more information in `dune_lang/syntax.ml` which we make use of. This allows an error message like before to become:

```
  File "dune-project", line 1, characters 11-18:
  1 | (lang dune 123.123)
                 ^^^^^^^
  Error: Version 123.123 of the dune language is not supported.
  Supported versions of the dune language in version 3.20 of dune are:
  - 1.0 to 1.12
  - 2.0 to 2.9
  - 3.20 to 3.20
```

Before we were mentioning "extensions" to the user which doesn't make too much sense in this context.

### DIFF
--- a/src/dune_sexp/syntax.ml
+++ b/src/dune_sexp/syntax.ml
@@ -423,26 +423,43 @@ let check_supported ~dune_lang_ver t (loc, ver) =
           please_port_message
       ]
   | `Unsupported_in_project (supported_ranges, min_lang_ver) ->
-    let dune_ver_text v =
-      Printf.sprintf "version %s of the dune language" (Version.to_string v)
+    let lang_or_ext =
+      match String.equal t.name "dune" with
+      | true -> `Language
+      | false -> `Extension
     in
-    let until =
-      match min_lang_ver with
-      | Some v -> Printf.sprintf " until %s" (dune_ver_text v)
-      | None -> ""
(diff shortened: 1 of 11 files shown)


Alizter:
Better alternative in https://github.com/ocaml/dune/pull/12833

PR #11935, opened 2025-06-22
pkg: include sys_ocaml_version in solver env

gridbugs:
This fixes a problem when using the "ocaml-system" compiler package and setting the "sys-ocaml-version" variable in the solver environment. Without including sys_ocaml_version in the solver environment at build time, the presence of sys_ocaml_version in the solver environment recorded in the lockdir causes dune to incorrectly believe that the solver environment is incompatible with the current system, even if the version of the currently-installed system compiler is the same as the one in the solver environment recorded in the lockdir.


gridbugs:
This has been subsumed by https://github.com/ocaml/dune/pull/11933.

PR #11937, opened 2025-06-23
boot: ignore warning 53

Alizter:
We ignore warning 53 in the bootstrap:

```
cd _test_boot && /nix/store/qiglg97y85w23ivjwhqhac1c3rhid39s-ocaml+flambda-5.3.0/bin/ocamlopt.opt -c 
-g -no-alias-deps -w -49-6 -alert -unstable -I +unix -I +threads ocamlc_loc.mli

File "otherlibs/ocamlc-loc/src/ocamlc_loc.mli", line 1, characters 4-9:
Warning 53 [misplaced-attribute]: the "alert" attribute cannot appear in this context

cd _test_boot && /nix/store/qiglg97y85w23ivjwhqhac1c3rhid39s-ocaml+flambda-5.3.0/bin/ocamlopt.opt -c 
-g -no-alias-deps -w -49-6 -alert -unstable -I +unix -I +threads dune_action_plugin.mli

File "otherlibs/dune-action-plugin/src/dune_action_plugin.mli", line 1, characters 4-9:
Warning 53 [misplaced-attribute]: the "alert" attribute cannot appear in this context
```

As far as I can tell, we are not using floating attributes in these files incorrectly. This warning started appearing in OCaml 5.3.

@nojb could you also confirm we aren't using floating attributes incorrectly.

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -1179,6 +1179,32 @@ let common_build_args name ~external_includes ~external_libraries =
 
 let allow_unstable_sources = [ "-alert"; "-unstable" ]
 
+let ocaml_warnings =
+  let warnings =
+    [ (* Warning 49 [no-cmi-file]: no cmi file was found in path for module *)
+      "-49"
+    ; (* Warning 23: all the fields are explicitly listed in this record: the
+        'with' clause is useless. 
+
+         In order to stay version independent, we use a trick with `with` by
+         creating a dummy value and filling in the fields available in every
+         OCaml version. forced_major_collections is the one missing in versions
+         older than 4.12. We therefore disable warning 23 for our purposes. *)
+      "-23"
+    ; (* Warning 53 [misplaced-attribute]: the "alert" attribute cannot appear
+        in this context
+
+        Any .mli files that begin wtih [@@@alert] will cause the compiler to
+        emit warning 53 due to the way we alias modules with an `open!`. It may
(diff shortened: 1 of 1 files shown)


nojb:
> @nojb could you also confirm we aren't using floating attributes incorrectly.

The cause for the warning is that the library is wrapped so that the `.mli` that is actually fed to the compiler looks like this:
```
open! Ocamlc__loc__ (* wrapper module *)
[@@@alert unstable ...]
```
and the warning is correctly triggered.

This behaviour of prepending `open!` seems specific to the bootstrap procedure, if I'm not wrong Dune itself uses an `-open` flag to being the wrapper module into scope. Maybe the same could be done in the bootstrap process?

Alizter:
Thanks for taking a look @nojb. I will see if I can instead pass `-open`.

Alizter:
I've tried to use `-open` but it makes dependency tracking much more difficult. I guess the next option would be to detect `[@@@` and put the `open` afterwards.

nojb:
> I've tried to use `-open` but it makes dependency tracking much more difficult. I guess the n ext option would be to detect `[@@@` and put the `open` afterwards.

If you feel like giving it a shot :) Or disable the warning as in your first approach, but with an explanatory comment for future readers.

Alizter:
I've gone ahead and documented the warnings rather than come up with a complicated workaround.

PR #11938, opened 2025-06-23
Fix typo in the Dune manual, overview chapter

shindere:
recommended -> recommend


maiste (approved):
LGTM! Thanks.
However, you need to sign your commit with `git commit --signoff` so we can merge it.

PR #11940, opened 2025-06-23
Fix windows.h header

toots:


otherlibs/xdg/xdg_stubs.c
```diff
@@ -10,9 +10,15 @@
...
 
+#if defined(__MINGW32__) || defined(__MINGW64__)
+#include <windows.h>
+#include <knownfolders.h>
+#include <shlobj.h>
+#else
 #include <Windows.h>
 #include <Knownfolders.h>
 #include <Shlobj.h>
+#endif
```
nojb: Are you compiling on a case-sensitive file system?

toots: Cross-compiling on linux with mingw, yes.

nojb: OK, thanks. I think we can just use the lowercase filenames then everywhere. Can you adapt the PR?

toots: I can but I was reading and it seems like NTFS can optionally be case-sensitive as well.

toots: For instance: https://learn.microsoft.com/en-us/windows/wsl/case-sensitivity

nojb: I believe this was added for WSL and is not commonly used outside of that. Also, the OCaml compiler uses lowercase Windows includes, so we would not be doing anything worse than that.

Let's go with lowercase for simplicity here, and we will revisit as/if needed in the future.

toots: Sure thing. Done!

nojb (approved):
LGTM

You need to sign your commits before we can merge (`git commit -s`).

toots:
> LGTM
> 
> You need to sign your commits before we can merge (`git commit -s`).

Done.

vendor/opam/src/core/custom_opamStubs.c
```diff
@@ -6,7 +6,7 @@
 #include <caml/memory.h>
 #include <caml/osdeps.h>
 #include <caml/alloc.h>
-#include <Windows.h>
+#include <windows.h>
```
nojb, line 9: Sorry, one more thing: note that these two (blake3 and opam) are vendored. You may want to submit these patches upstream as well. Otherwise we may lose them whenever they are synchronized again.

Alizter: FTR this is hitting in #13097. The correct thing to have done here would have been submitting the patches upstream and then updating the vendored scripts here. Hopefully #13097 will make this process more transparent.

PR #11942, opened 2025-06-25
test: -g flag in foreign c stubs

Alizter:


rgrinberg:
As a sanity check, does this match the behavior we have for OCaml flags and foreign libraries?

Alizter:
I'll come back to this eventually.

PR #11943, opened 2025-06-26
Restore 32bit build

hhugo:


Alizter (approved):

boot/duneboot.ml
```diff
@@ -983,7 +996,23 @@ module Library = struct
...
         match src.kind with
         | C c ->
-          if keep_c c ~architecture then `Left { flags = c.flags; name = fn } else `Skip
+          if keep_c c ~architecture
+          then (
+            let extra_flags =
+              if String.is_prefix ~prefix:"blake3_" fn
+              then (
+                match architecture with
+                | "x86" | "i386" | "i486" | "i586" | "i686" ->
```
hhugo, line 1005: Thinking a bit more about this, I think it would be more robust to match architecture that should not received these flags.

hhugo: We could as well use the portable version of blake3 always for bootstrap

rgrinberg: > We could as well use the portable version of blake3 always for bootstrap

That would mean all dune users would use the slow version. The bootstrapped version is the one that is installed.

PR #11944, opened 2025-06-26
refactor: prefix rpc_ to rpc modules in bin

gridbugs:
This allows the Build_cmd module to be renamed to Build (directories to not act like module namespaces in dune projects so rpc/build.ml and build.ml would otherwise conflict). Since all the other files implementing direct dune subcommands do not have the "_cmd" suffix, this suffix is removed from build_cmd.ml. We now need to rename rpc/build.ml, and since rpc/rpc_common.ml already exists, we follow suit to name it rpc/rpc_build.ml. Then for consistency, all other files in the rpc directory are renamed to rpc_*.


Alizter:
If @rgrinberg agrees, maybe we could use `include_subdirs qualified` here (and only here for now).

rgrinberg:
Even if I were to agree, it wouldn't be so simple. Bootstrapping doesn't support this.

I'm also less excited about this mode than I was once before. I'd rather not trade compilation times for nice looking module names.

Alizter (approved):

PR #11945, opened 2025-06-27
[rocq] Support for installation of package metadata.

### LINKED ISSUE #11012: Coq cannot find the cmxs file for transitive ocaml dependencies

### LINKED ISSUE #11483: Dune-Coq: support transitive dependencies even for installed theories
In short, depending on a coq.theory should also make its dependencies available. This issue is known to experts (at least Emilio and Rodolphe) but I didn't find a bug-tracker entry, and at Bluerock we are interested in a solution, so I figured I would make one.

## Desired Behavior

Assume we have `(coq.theory (name foo.logical.prefix) (theories direct_deps))` and a client `client`. Let us write `deps` for `direct_deps` plus any theories they depend on.
When `client`'s `bar.v` tries to `Require foo.logical.prefix.some_module.`, any used module from `deps` must be in scope — so clients of theory `foo.logical.prefix` need to depend on theories in `deps`, and shouldn't have to list `deps` explicitly among their dependencies — that can even change when `foo.logical.prefix` evolves.

Dune usually supports this, except if `foo.logical.prefix` is installed and clients are using `dune-coq >= 0.8`: in that case:

1. `deps` are also installed, but
2. installed theories are not available implicitly (https://dune.readthedocs.io/en/stable/coq.html#coq-language-version), and
3. installing `foo.logical.prefix` doesn't even _record_ that it depends on `direct_deps`.

Hence, we should

4. change `3` and record the dependencies `direct_deps` of theories when installing them
5. and make those dependencies `deps` (so, `direct_deps` and _their_ dependencies) available to `client`.

Possible future extension:

If `client` requires on some modules from `dep` in `deps` explicitly, arguably `client` should depend on `dep` explicitly — to be more robust if `foo.logical.prefix` drops the dependency on `dep`. Emilio has mentioned this a few times. However, nowadays dependencies are either available or not, and supporting this extension would require distinguishing direct and indirect uses — either in Coq, or in separate linters. Thankfully, that does not affect point 4: all (transitive) dependencies `deps` of `foo.logical.prefix` would be available as _indirect_ dependencies for `client`, and none as direct dependencies. Hence, I propose to leave discussion of this to future work.

### Implementation ideas

IIRC, Rodolphe proposed to reuse findlib for this:
https://coq.zulipchat.com/#narrow/channel/240550-Dune-devs-.26-users/topic/.5Blaodpath.5D.20Coqdep.20vs.2E.20Coqmod/near/285853410

### Other threads about this

https://coq.zulipchat.com/#narrow/channel/240550-Dune-devs-.26-users/topic/Dune-Coq.200.2E3.20dependency.20model
https://coq.zulipchat.com/#narrow/channel/240550-Dune-devs-.26-users/topic/Adding.20transitive.20deps.20into.20theories

ejgallego:
We now install a `rocq-package` file for each Rocq theory. This way, when Rocq developments build with Dune are globally installed, we can find out additional metadata, such as the list of their dependencies.

Fixes #11012 fixes #11483

TODO:

- update changelog
- update documentation

Note: rebased on top of #12035 

Joint work with @Lysxia


src/dune_rules/coq/coq_lib.ml
```diff
@@ -594,10 +602,9 @@ module DB = struct
         let+ (_ : (Loc.t * lib) list) = theory.theories in
         Dune theory
       | Found_path cp ->
-        let open Resolve.Memo.O in
-        let boot_id = coq_db.boot_id in
-        let+ theory = create_from_coqpath ~boot_id cp in
-        Legacy theory
+        let db = assert false in
+        let dir = assert false in
```
Lysxia: @ejgallego What should those be?

src/dune_rules/coq/coq_lib.ml
```diff
@@ -458,23 +458,23 @@ module DB = struct
       Resolve.Memo.List.map theories ~f
     ;;
 
-    let create_from_stanza_impl (coq_db, db, dir, (s : Coq_stanza.Theory.t)) =
+    let create_from_coq_package_impl (coq_db, db, dir, (s : Coq_package.meta)) =
```
Lysxia: Our initial idea was to convert a `Coq_package` into a `Coq_stanza` but some fields don't seem easy to populate, so I am trying the other way around. However this memo table relies on physical equality of `Coq_stanza` so I'm not sure this still works as expected.

ejgallego: Which fields are not easy to populate? I think we should record enough info in the Coq_package as to be able to do so.

Indeed the equality here is tricky; I'll have a look later on.

src/dune_rules/coq/coq_lib.ml
ejgallego: dir and db are useful in the memo table as to distinguish stanzas that may appear (and be private) in different composed builds.

I think we can have a case where two identical stanzas live in two separate directories, if these are private. I'm not sure why `dir` is not enough here tho, I'd dare to say that stanzas are uniquely identified by the pair `(dir, stanza)`

We can ask Rudy once we get the patch more advanced, I guess for now we can pass `db = installed_db` and `dir` the directory we found the rocq-package file.

Lysxia:
I made some progress. Now it looks like I broke the finding of plugins:
```
File "test/blackbox-tests/test-cases/coq/compose-installed-plugin.t/run.t", line 1, characters 0-0:
...
+  File "$TESTCASE_ROOT/lib/coq/user-contrib/Bar/rocq-package", line 3, characters 9-22:
+  3 | (plugins global.plugin)
+               ^^^^^^^^^^^^^
+  Error: Library "global.plugin" not found.
+  -> required by theory Bar in
+     $TESTCASE_ROOT/lib/coq/user-contrib/Bar/rocq-package:2
+  -> required by _build/default/.Foo.theory.d
+  -> required by alias all
```

src/dune_rules/coq/coq_lib.ml
Lysxia: `project: DuneProject.t` is the one I was having trouble with.

ejgallego: I see! This is indeed not needed for the db, I'd suggest to make it an option for now, there are already some codepaths where we use get_exn for it.

ejgallego:
> I made some progress. Now it looks like I broke the finding of plugins:
> 
> ```
> File "test/blackbox-tests/test-cases/coq/compose-installed-plugin.t/run.t", line 1, characters 0-0:
> ...
> +  File "$TESTCASE_ROOT/lib/coq/user-contrib/Bar/rocq-package", line 3, characters 9-22:
> +  3 | (plugins global.plugin)
> +               ^^^^^^^^^^^^^
> +  Error: Library "global.plugin" not found.
> +  -> required by theory Bar in
> +     $TESTCASE_ROOT/lib/coq/user-contrib/Bar/rocq-package:2
> +  -> required by _build/default/.Foo.theory.d
> +  -> required by alias all
> ```

I think this test-case is for the deprecated method of installing plugins in `user-contrib`, so IMHO it is fine if it broken.

You could fix it by scanning the cmxs files, but given that Rocq 9 doesn't support this anymore, IMHO not worth it.

ejgallego:
> You could fix it by scanning the cmxs files, but given that Rocq 9 doesn't support this anymore, IMHO not worth it.

I'm sorry I'll be away for the rest of the week, but as soon as I'm back I'll rebase this on top of `(lang rocq ...) which should avoid this problem.

src/dune_rules/rocq/rocq_lib.ml
```diff
@@ -533,26 +580,22 @@ module DB = struct
 
     (** Our final final resolve is used externally, and should return the
         library data found from the previous iterations. *)
-    let resolve rocq_db (loc, name) =
+    let resolve ~db rocq_db (loc, name) =
```
ejgallego, line 583: I think the `db` parameter here may not be right, note what we do for the case in `Found_stanza`, but I'm not a 100% sure.

In principle, it indeed seems wrong for us to resolve the OCaml plugin information with the library DB that belongs to a stanza, in this case for example, private ML libraries would be present. So I guess we should have the code to capture the right DB here.

However, this requires a bit more thinking.

Lysxia:
Just rebased on main along with #12035, still wip.

Lysxia:
@rgrinberg Why was this closed? I thought this is supposed to fix #11483

rgrinberg:
Because it seemed like the PR was abandoned. I assume you replied because you're still working on it, so I'll re-open. Anything that you're stuck on?

Lysxia:
Sorry about it. I've been busy elsewhere, but I'm still interested in fixing this!

rgrinberg:
Let's re-open this once you start working on it. There's absolutely no pressure from me, and I am neither for nor against this feature. But there's no point using a Active PR's as a personal TODO list.

PR #7195, opened 2023-02-28
chore: add rule streaming proposal

rgrinberg:
Signed-off-by: Rudi Grinberg <me@rgrinberg.com>

PR #11953, opened 2025-07-01
Showcase an issue with melange and runtime_deps

davesnx:
There's a use-case for promoting stuff that has been generated by external program into the source code. This is very helpful since there are some test runners that need to run on the context of the _build folder but we want to diff against the source file locations. We can't use diff or other mechanisms from dune, since those programs own all of the process.

Here I showcase 2 tests: promotion in melange appends the melange folder into the source folder (which might be a problem with promotion by itself) but when you try to hijack it by adding `(into` and `(only` things don't work as expected. 

Not on the docs but, into should be the source folder? Why errors showcase (in case of wrong path) the build folder? Also, `(only xxx)` mentions it's using the `<dir>` semantics but it doesn't complain about anything... which makes it unsure if you are matching or not.

Anyway, maybe this PR just is open to discuss this use-case


anmonteiro:
to confirm what's happening here:
- you have a subdirectory a/x.js
- promotion by default will put targets in a/dist/a/snapshots/x.js
- you'd like them to end up back in a/x.js


this is not currently possible, and I don't think we can implement melange promotion into the same source directory until the [rule streaming proposal](https://github.com/ocaml/dune/pull/7195) has been implemented.

anmonteiro:
hey @davesnx, I added a test showcasing this in https://github.com/ocaml/dune/pull/11976. The reason why this hack can't work reliably is because promotion `(into ..)` is relative to the artifact, not the directory where the dune file lives.

rgrinberg:
> (into ..) is relative to the artifact, not the directory where the dune file lives.

Seems like a bug with dune to me at first glance

anmonteiro:
this is being addressed in e.g. #12102

PR #11957, opened 2025-07-02
fix: flaky tests

Alizter:
Since we <s>terminate</s> kill subprocesses of processes that we run as actions we no longer need to manually send signals to dune in our tests. This quick succession of signals from the cram test and dune (running the test) caused our Ctrl+C message to appear.

There are two more tests which are still flaky, but I am not sure what is causing them to hang. I will need to investigate at a later point.


Alizter:
Still causing issues...

PR #11960, opened 2025-07-03
test: wait after killing subprocess

Alizter:
Finishing a cram test immediately after killing a subprocess such as dune does not give it enough time to terminate. Instead, we should be `wait`ing for the subprocess to terminate. Previously, this meant we would get the Ctrl+C some of the time when trying to terminate dune, but now we have a more consistent behaviour.


rgrinberg (approved):
LGTM. if these tests flake again, please disable them

PR #11961, opened 2025-07-03
Clarify in the docs about dependency collection in package management

Sudha247:
In an offline discussion with @punchagan and @Leonidas-from-XIV, the topic of collecting dependencies from `dune-project` and `.opam` files came up. I thought it would be useful to clarify that if dependencies are declared in the `dune-project` file, `dune pkg` wouldn't look for dependencies in the `.opam` file.

### DIFF
--- a/doc/explanation/package-management.md
+++ b/doc/explanation/package-management.md
@@ -57,13 +57,11 @@ file. This allows projects to completely omit generation of `.opam` files, as
 long as they use Dune for package management. Thus all dependencies on OCaml
 packages are only declared in one single file.
 
-For compatibility with a large number of existing projects, Dune will also
-collect dependencies from `.opam` files, if dependencies are not declared in the
-`dune-project`. So while recommended, there is no obligation to switch to
-declaring dependencies in the `dune-project`. Likewise the generation of `.opam`
-files continues to be supported. If the dependencies are declared in both
-`dune-project` and `.opam` files, Dune will only use the declarations from
-`dune-project`.
+To maintain compatibility with a large number of existing projects, Dune
+continues to support `.opam` files. While it is recommended to declare the
+dependencies directly in the `dune-project` file, it is not mandatory to do so.
+Dune will fall back to reading dependencies from `.opam` files when they are not
+specified in `dune-project`.
 
 
 ## Locking
--- a/doc/explanation/package-management.md
+++ b/doc/explanation/package-management.md
@@ -57,11 +57,14 @@ file. This allows projects to completely omit generation of `.opam` files, as
 long as they use Dune for package management. Thus all dependencies on OCaml
 packages are only declared in one single file.
 
-For compatibility with a larger amount of existing projects, Dune will also
-collect dependencies from `.opam` files in the project. So while recommended,
-there is no obligation to switch to declaring dependencies in the
-`dune-project`. Likewise the generation of `.opam` files continues to be
-supported.
+For compatibility with a large number of existing projects, Dune will also
+collect dependencies from `.opam` files, if dependencies are not declared in the
+`dune-project`. So while recommended, there is no obligation to switch to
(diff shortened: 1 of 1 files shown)


Sudha247:
> This allows projects to completely omit generation of `.opam` files, as long as they use Dune for package management.

I have a question about this in the previous paragraph, I'm unable to comment on the PR. In principle, this is possible (and perhaps desirable). However, for releasing packages to [opam-repository](https://github.com/ocaml/opam-repository), one still needs to generate opam files. In this case, would the recommendation be to update the `dune-project` file and generate the `.opam` file from it?

Leonidas-from-XIV:
> would the recommendation be to update the dune-project file and generate the .opam file from it?

I'd say generally yes. I believe there were some thoughts about allowing `opam` to read the necessary metadata out of `dune-project` (or maybe call `dune` to give it that metadata) but so far there's no way to do so and for the sake of pragmatism it is mostly the best to just autogenerate the OPAM files and commit them.

A package tarball does not need to include the OPAM file as the metadata in opam-repository is used instead of the included OPAM file but if the project doesn't include OPAM files it can't be pinned by OPAM. So most tarballs and repos do include OPAM files, even if they're mostly unused.

doc/explanation/package-management.md
```diff
@@ -57,11 +57,14 @@ file. This allows projects to completely omit generation of `.opam` files, as
...
-`dune-project`. Likewise the generation of `.opam` files continues to be
-supported.
+For compatibility with a large number of existing projects, Dune will also
+collect dependencies from `.opam` files, if dependencies are not declared in the
+`dune-project`. So while recommended, there is no obligation to switch to
+declaring dependencies in the `dune-project`. Likewise the generation of `.opam`
+files continues to be supported. If the dependencies are declared in both
+`dune-project` and `.opam` files, Dune will only use the declarations from
+`dune-project`.
+
```
Leonidas-from-XIV, line 65: I think this paragraph repeats the same statement twice and as such reads very confusingly. I suggest throwing out the original text and reformulating with a message along with "dune will use opam files if the dependencies are not declared in dune-project". That's a lot more straightforward and does not contain multiple if clauses.

Leonidas-from-XIV (changes requested):

doc/explanation/package-management.md
Sudha247: Thanks, I've rewritten the paragraph.

doc/explanation/package-management.md
```diff
@@ -57,13 +57,11 @@ file. This allows projects to completely omit generation of `.opam` files, as
...
-`dune-project`. So while recommended, there is no obligation to switch to
-declaring dependencies in the `dune-project`. Likewise the generation of `.opam`
-files continues to be supported. If the dependencies are declared in both
-`dune-project` and `.opam` files, Dune will only use the declarations from
-`dune-project`.
+To maintain compatibility with a large number of existing projects, Dune
+continues to support `.opam` files. While it is recommended to declare the
+dependencies directly in the `dune-project` file, it is not mandatory to do so.
+Dune will fall back to reading dependencies from `.opam` files when they are not
+specified in `dune-project`.
```
Leonidas-from-XIV: This is a lot nicer. I would however specifically mention that it will be reading dependencies from opam files if the *packages* aren't defined in `dune-project`.

Because if packages are defined in `dune-project` but have no dependencies Dune will assume they have no dependencies, no matter what the OPAM files say.

Sudha247: That's a good point! Updated. I assumed if there's no dependency field in the dune-project, it falls back to opam. Corrected now,

Leonidas-from-XIV: Originally I did too, but this is the cause of the "no dependencies to lock" issue we had on a number of packages in opam-health-check.

Leonidas-from-XIV (approved):

PR #11963, opened 2025-07-03
feat(oxcaml): add the library_parameter stanza

### LINKED ISSUE #12084: [OxCaml] Parameterized libraries: library parameter
The issue is a tracker for the issue related to Parameterized Library parameter. It tracks the work done about the implementation of the parameter, an equivalent of the `virtual library` included inside of OxCaml.

maiste:
This PR is the first of a set of four PRs related to the introduction of Parameterized Libraries in Dune.

This feature is only available with OxCaml until it is upstream to the compiler. 

- Support for the `library_parameter` stanza
- Test of the newly introduced stanza. The tests will be updated with the next PRs.
- The documentation related to the new stanza.

Closes #12084

### DIFF
--- a/doc/reference/dune/library_parameter.rst
+++ b/doc/reference/dune/library_parameter.rst
@@ -1,0 +1,109 @@
+library_parameter
+-----------------
+
+.. warning::
+
+   This feature is experimental and requires the compiler you are using to
+   support parameterized libraries.
+
+The ``library_parameter`` stanza describes a parameter interface defined in a single ``.mli`` file. To enable this feature,
+you need to add ``(using oxcaml 0.1)`` :doc:`extension
+</reference/dune-project/using>` in your ``dune-project`` file.
+
+.. describe:: (library_parameter ...)
+
+  .. versionadded:: 3.20
+
+  Define a parameter.
+
+  .. describe:: (name <parameter-name>)
+
+    ``parameter-name`` is the name of the library parameter. It must be a valid
+    OCaml module name as for :doc:`/reference/dune/library`. 
+
+    This must be specified if no `public_name` is specified.
+
+  .. describe:: (public_name ...)
+
+    The name under which the library parameter can be referred as a dependency
+    when it's not part of the current workspace, i.e., when it is installed.
+    Without a ``(public_name ...)`` field, the library parameter won't be
+    installed by Dune. The public name must start with the package name it's
+    part of and optionally followed by a dot, then anything else you want. The
+    package name must also be one of the packages that Dune knows about, as
+    determined by the logic described in :doc:`/reference/packages`.
+
+  .. describe:: (package <package>)
+
+    Installs a private library parameter under the specified package. Such a
+    parameter is now usable by public libraries defined in the same project.
+
+  .. describe:: (synopsis <string>)
+
+    A one-line description of the library parameter.
+
+  .. describe:: (modules <modules>)
+
+    Specifies a specific module to select as a `library_parameter`.
+
+    ``<modules>`` uses the :doc:`/reference/ordered-set-language`, where
+    elements are module names and don't need to start with an uppercase letter.
+
+    The library parameter **must** only declare one ``mli`` file as part of its
+    modules.
+
+  .. describe:: (libraries <library-dependencies>)
+
+    Specifies the library parameter's dependencies.
+
+    See :doc:`/reference/library-dependencies` for more details.
+
+  .. describe:: (preprocesss <preprocess-spec>)
+
+    Specifies how to preprocess files when needed.
+
+    The default is ``no_preprocessing``, and other options are described
+    in :doc:`/reference/preprocessing-spec`.
+
+  .. describe:: (preprocessor_deps (<deps-conf list>))
+
+    Specifies extra preprocessor dependencies preprocessor, i.e., if the
+    preprocessor reads a generated file.
+
+    The specification of dependencies is described in
+    :doc:`/concepts/dependency-spec`.
+
+
+  .. describe:: (flags ...)
+
+    See :doc:`/concepts/ocaml-flags`.
+
+  .. describe:: (ocamlc_flags ...)
+
+    See :doc:`/concepts/ocaml-flags`.
+
+  .. describe:: (ocamlopt_flags ...)
+
+   See :doc:`/concepts/ocaml-flags`.
+
+  .. describe:: (optional)
+
+    If present, it indicates that the library parameter should only be built
+    and installed if all the dependencies are available, either in the
+    workspace or in the installed world.
+
+  .. describe:: (enabled_if <blang expression>)
+
+    Conditionally disables a library parameter.
+
+    A disabled library parameter cannot be built and will not be installed.
+
+    The condition is specified using the :doc:`/reference/boolean-language`, and
+    the field allows for the ``%{os_type}`` variable, which is expanded to the
+    type of OS being targeted by the current build. Its value is the same as the
+    value of the ``os_type`` parameter in the output of ``ocamlc -config``.
+
+  .. describe:: (allow_overlapping_dependencies)
+
+    Allows external dependencies to overlap with libraries that are present in
+    the workspace.
--- a/src/dune_rules/gen_meta.ml
+++ b/src/dune_rules/gen_meta.ml
@@ -76,7 +76,7 @@ let gen_lib pub_name lib ~version =
   in
   let preds =
     match kind with
-    | Normal -> []
+    | Normal | Parameter -> []
     | Ppx_rewriter _ | Ppx_deriver _ -> [ Pos "ppx_driver" ]
   in
   let name lib =
@@ -127,7 +127,7 @@ let gen_lib pub_name lib ~version =
          ; ppx_runtime_deps ppx_rt_deps
          ])
     ; (match kind with
-       | Normal -> []
+       | Normal | Parameter -> []
        | Ppx_rewriter _ | Ppx_deriver _ ->
          (* Deprecated ppx method support *)
          let no_ppx_driver = Neg "ppx_driver"
@@ -139,7 +139,7 @@ let gen_lib pub_name lib ~version =
              ; requires ~preds:[ no_ppx_driver ] ppx_runtime_deps_for_deprecated_method
              ]
            ; (match kind with
-              | Normal -> assert false
+              | Normal | Parameter -> assert false
               | Ppx_rewriter _ ->
                 [ rule "ppx" [ no_ppx_driver; no_custom_ppx ] Set "./ppx.exe --as-ppx"
                 ; rule "library_kind" [] Set "ppx_rewriter"
--- a/src/dune_rules/modules_field_evaluator.ml
+++ b/src/dune_rules/modules_field_evaluator.ml
@@ -16,6 +16,7 @@ type kind =
   | Virtual of Virtual.t
   | Implementation of Implementation.t
   | Exe_or_normal_lib
+  | Parameter
 
 let eval0 =
   let key = function
@@ -106,6 +107,7 @@ let find_errors
       ~private_modules
       ~existing_virtual_modules
       ~allow_new_public_modules
+      ~is_parameter
   =
   let all =
     (* We expect that [modules] is big and all the other ones are small, that's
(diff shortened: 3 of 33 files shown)


doc/reference/dune/library_parameter.rst
```diff
@@ -0,0 +1,112 @@
...
+-----------------
+
+.. warning::
+
+   This feature is experimental and requires the compiler you are using to
+   support parameterized libraries.
+
+The ``library_parameter`` describes a parameter. To be able to use the feature,
+you need to enable the stanza using the ``(using oxcaml 0.1)`` :doc:`extension
+</reference/dune-project/using>` in your ``dune-project`` file.
```
art-w:
```suggestion
The ``library_parameter`` stanza describes a parameter interface defined in a single ``.mli`` file. To enable this feature,
you need to add ``(using oxcaml 0.1)`` :doc:`extension
</reference/dune-project/using>` in your ``dune-project`` file.
```

doc/reference/dune/library_parameter.rst
```diff
@@ -0,0 +1,112 @@
...
+
+.. describe:: (library_parameter ...)
+
+  .. versionadded:: 3.20
+
+  Define a parameter.
+
+  .. describe:: (name <parameter-name>)
+
+    ``parameter-name`` is the real of the library parameter. It must be a valid
```
art-w:
```suggestion
    ``parameter-name`` is the name of the library parameter. It must be a valid
```

doc/reference/dune/library_parameter.rst
```diff
@@ -0,0 +1,112 @@
...
+  .. describe:: (synopsis <string>)
+
+    A one-line description of the library parameter.
+
+  .. describe:: (modules <modules>)
+
+    Specifies what modules are part of the library parameter. By default, Dune will use
+    all the ``.mli`` files in the same directory as the ``dune`` file. This
+    includes ones present in the file system as well as ones generated by user
+    rules. You can restrict this list by using a ``(modules <modules>)`` field.
```
art-w: A library parameter can only have a single `.mli` file, so the documentation should not say "all the .mli files in the directory"... In fact I'm not sure if we should keep the `(modules ...)` field, since its value should always be `parameter_name`

doc/reference/dune/library_parameter.rst
```diff
@@ -0,0 +1,112 @@
...
+
+    If present, it indicates that the library parameter should only be built
+    and installed if all the dependencies are available, either in the
+    workspace or in the installed world.
+
+  .. describe:: (enabled_if <blang expression>)
+
+    Conditionally disables a library.
+
+    A disabled library cannot be built and will not be installed.
```
art-w:
```suggestion
    Conditionally disables a library parameter.

    A disabled library parameter cannot be built and will not be installed.
```

src/dune_rules/gen_meta.ml
```diff
@@ -77,6 +77,7 @@ let gen_lib pub_name lib ~version =
   let preds =
     match kind with
     | Normal -> []
+    | Parameter -> [] (* failwith "TODO arthur" *)
```
art-w: ha! :D

src/dune_rules/modules_field_evaluator.ml
```diff
@@ -378,8 +389,10 @@ let eval
         if Module_trie.mem private_modules name then Visibility.Private else Public
       in
       let kind =
-        if Module_trie.mem virtual_modules name
-        then Module.Kind.Virtual
+        if kind = Parameter
```
art-w: nitpick:
```suggestion
        if is_parameter
```

src/dune_rules/stanzas/parameter.ml
```diff
@@ -0,0 +1,138 @@
...
+       ; lint
+       ; flags
+       ; js_of_ocaml =
+           { js = Js_of_ocaml.In_buildable.default
+           ; wasm = Js_of_ocaml.In_buildable.default
+           }
+       ; allow_overlapping_dependencies
+       ; ctypes = None
+       }
+     (* let+ buildable = Buildable.decode Parameter  *)
```
art-w:
```suggestion
```

src/dune_rules/stanzas/parameter.ml
```diff
@@ -0,0 +1,138 @@
...
+     and+ name = field_o "name" Lib_name.Local.decode_loc
+     and+ public = field_o "public_name" (Public_lib.decode ~allow_deprecated_names:false)
+     and+ package = field_o "package" (located Stanza_common.Pkg.decode)
+     and+ enabled_if = Enabled_if.decode ~allowed_vars:Any ~since:(Some (3, 20)) ()
```
art-w:
```suggestion
     and+ enabled_if = Enabled_if.decode ~allowed_vars:Any ~since:None ()
```

src/dune_rules/stanzas/parameter.ml
```diff
@@ -0,0 +1,138 @@
...
+         (match Lib_name.to_local (loc, name) with
+          | Ok m -> loc, m
+          | Error user_message ->
+            User_error.raise
+              ~loc
+              [ Pp.textf "Invalid library name."
+              ; Pp.text
+                  "Public library names don't have this restriction. You can either \
+                   change this public name to be a valid library name or add a \"name\" \
+                   field with a valid library name."
```
art-w:
```suggestion
              [ Pp.textf "Invalid parameter name."
              ; Pp.text
                  "Public library parameter names don't have this restriction. You can either \
                   change this public name to be a valid parameter name or add a \"name\" \
                   field with a valid parameter name."
```

test/blackbox-tests/test-cases/oxcaml/library_parameter.t
```diff
@@ -0,0 +1,143 @@
+This test verifies `library_parameter` stanza works as expected.
```
art-w:
```suggestion
This test verifies that the `library_parameter` stanza works as expected.
```

test/blackbox-tests/test-cases/oxcaml/library_parameter.t
```diff
@@ -0,0 +1,143 @@
...
+
+  $ mkdir param
+  $ echo "type t = int" > param/param_intf.mli
+  $ cat > "param/dune" <<EOF
+  > (library_parameter
+  >  (public_name param.intf)
+  >  (name param_intf))
+  > EOF
+
+The test build should fails because of the oxcaml extension is not available.
```
art-w:
```suggestion
The test build should fails because the oxcaml extension is not available.
```

src/dune_rules/utop.ml
```diff
@@ -95,6 +96,7 @@ let add_stanza db ~dir (acc, pps) stanza =
          let info = Lib.info lib in
          match Lib_info.kind info with
          | Normal -> Appendable_list.cons lib acc, pps
+         | Parameter -> failwith "TODO arthur"
```
art-w: I suspect the library parameters can follow the same logic as `Normal` libraries for utop, but if this is hard to test for this PR, then perhaps it's possible to return `(acc,pps)` with a comment to come back to this once we have the rest of the feature

art-w (changes requested):

src/dune_rules/stanzas/parameter.ml
maiste: I'll keep the `library_parameter` name to not create confusion later with the `parameters` field :+1:

src/dune_rules/utop.ml
maiste: I'll add it as `Normal` and left a CR comment to remember to rework this when all the PRs would be merged.

doc/reference/dune/library_parameter.rst
maiste: Should we go in the direction where we keep the `modules` field or if, by default, `library_parameter` only take the `parameter-name` and exclude the rest of the modules? My point of view is that `modules` would keep Dune users in familiar waters but taking `parameter-name` would remove the risk to specify multiple modules... I'm not sure about the best behaviour to adapt here. 

In the rework version, I have simplified the phrasing for our need.

WDYT @rgrinberg?

rgrinberg:
@maiste is this ready for another round of review?

maiste:
Yes it is :smile:  I have addressed all of @art-w comments. 
Thanks!

maiste:
@rgrinberg gentle ping. I have rebased the PR with `main` to make the CI happy.  I'll rework the commit history after review to make more readable commits.

rgrinberg (approved):
LGTM. I think the library_parameter stanza is overly permissive at the moment. Might be worth documenting this in the tests.

maiste:
Thanks for the review @rgrinberg! Just to understand before pushing new tests, what do you think in the stanza is overly permissive? Is the fields we kept that should have a default value instead (like optional)? Or is it something else?

src/dune_rules/stanzas/parameter.ml
```diff
@@ -0,0 +1,137 @@
...
+           }
+       ; allow_overlapping_dependencies
+       ; ctypes = None
+       }
+     and+ name = field_o "name" Lib_name.Local.decode_loc
+     and+ public = field_o "public_name" (Public_lib.decode ~allow_deprecated_names:false)
+     and+ package = field_o "package" (located Stanza_common.Pkg.decode)
+     and+ enabled_if = Enabled_if.decode ~allowed_vars:Any ~since:None ()
+     and+ synopsis = field_o "synopsis" string
+     and+ stdlib =
```
rgrinberg: I don't think this field makes sense for library parameters

test/blackbox-tests/test-cases/oxcaml/library_parameter.t
```diff
@@ -0,0 +1,143 @@
...
+  Is parameter: YES
+
+We create a second parameter in the same directory. We rewrite the dune file to
+use the `modules` field.
+
+  $ cat > "param/dune" <<EOF
+  > (library_parameter
+  >  (public_name param.intf)
+  >  (name param_intf)
+  >  (modules param_intf))
```
rgrinberg, line 49: Can you add a test for a library parameter with more than one module?

rgrinberg:
Sorry my previous comment was a bit opaque. I expanded on the fields I think are unnecessary and which tests I would like to see.

src/dune_rules/stanzas/parameter.ml
maiste: Even if you want to use `Base` as the standard library? Otherwise, I can remove it.

maiste:
Thanks for the clarifications! I have rewritten the git history to make it cleaner (to use merge commit when merging the PR). I have pushed a test to check the result when using a module. As you said it is currently too permissive as we don't check for the module size. We have extra tests we backported from JS tests but that are paired with _implementation_, _instantiation_ and _parameterization_. With @art-w, we intended to push them later along with the paired code.

src/dune_rules/stanzas/parameter.ml
rgrinberg: It won't work for base because it's a regular library. This is a field that is only useful for the stdlib itself. In other words, it's for using dune to build OCaml itself (for a better development workflow)

test/blackbox-tests/test-cases/oxcaml/library_parameter.t
```diff
@@ -61,9 +61,35 @@ We rebuild it and ensure we have the new parameter available.
...
+  >  (modules multiple_param data))
+  > EOF
+  $ cat > "multiple/data.mli" << EOF
+  > type t = int
+  > EOF
+  $ cat > "multiple/multiple_param.mli" << EOF
+  > type t = string
+  > EOF
+
+  $ dune build
```
Alizter, line 82: I'm a bit confused, it doesn't seem like this failed?

maiste: Yep, and that's a problem I need to fix for this PR.

maiste: I have pushed the fix: https://github.com/ocaml/dune/pull/11963/files#diff-97a3106adc9720a309352f46b8d89a0bdf768edd1ebf5709d49f925fce3aa07fR438

maiste:
@Alizter @rgrinberg I have made the changes according to your feedback. If this version is good, I will rework the git history to have something cleaner (working with the DCO check) and merge it.

Small question: even if this is experimental, should I add a `CHANGES.md` entry?

rgrinberg:
If you think it's useful for users to know about this addition, you're welcome to add it.

maiste:
I have pushed a CHANGES.md with the experimental label to make it clearer it is not something stable yet. I have also rewritten the history. @art-w and @rgrinberg if this is good in the current state for you, I'll merge the PR.

rgrinberg:
It looks to be in a good state for me.

art-w (approved):

PR #11972, opened 2025-07-05
Fix toolchain building

rgrinberg:
The temp file made it so that the toolchain would get rebuilt every time the rule would be re-generated. This is clearly not what we want.

The attempt to solve build time races between instances of dune is removed. This will be addressed in subsequent PR's where we do this in a way that avoids extra work and works for all packages and not just the toolchain.


Alizter (approved):

maiste:
@rgrinberg  I'm a bit confused with this change. IIRC, toolchain was written this way to "trick" dune when building the compiler to instead check for it in `~/.cache/dune/toolchain` instead of the build dir. With this change, it seems it tries to rebuild the compiler everytime you clean the build dir whereas it was not the case before. 

Is it something expected?

 I am not sure to understand: which part will be addressed in subsequent PR?

PR #11975, opened 2025-07-05
refactor: optimize reading files

rgrinberg:
Using channels to read a file into a string is rather slow and creates a mutex + a bunch of useless intermediate buffers.

We instead initialize the buffer once before reading.

This optimization has been used internally for ages with great success. This is an attempt to make a portable version of it.


rgrinberg:
The benchmark shows that this is indeed quite useful.

PR #11976, opened 2025-07-06
test(melange): show promotion into parent dir not supposed to work

anmonteiro:
this test shows why #11953 doesn't work with the current implementation of promotion.

### DIFF
--- a/test/blackbox-tests/test-cases/melange/promote-with-into.t
+++ b/test/blackbox-tests/test-cases/melange/promote-with-into.t
@@ -1,0 +1,41 @@
+Promotion with targets `(into ..)` a directory
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.8)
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/melange/promote.t
```diff
@@ -33,3 +33,40 @@ Targets are promoted to the source tree
 
   $ dune clean
   $ ls ./dist
+
+Promotion with targets in a directory
+
+  $ rm -rf ./dist ./hello.ml ./dune
```
rgrinberg: How about just moving this test into its own file and removing this clean up? It would make things less fragile and execute things in parallel when possible to boot.

anmonteiro: absolutely, doing that now

rgrinberg (approved):

PR #11978, opened 2025-07-06
refactor: use optimized write file

rgrinberg:
* Do not create a channel
* Try writing everything in a single syscall

This optimization was also pretty important internally. This is a more portable re-implementation


rgrinberg:
Our benchmarks show no change. Probably because our null benches do no writes and for the clean bench, the speed of writing files doesn't matter. In any case, we've done this change internally so we might as well do it here as well.

PR #11980, opened 2025-07-06
feat(runtest): allow running (tests) and (inline_tests)

Alizter:
In this PR we add support for running individual libraries with `(inline_tests)` and individual `(tests)` executables.


bin/runtest.ml
```diff
@@ -27,39 +28,128 @@ let runtest_info =
...
+;;
+
+let find_inline_tests_in_dir dir =
+  let open Memo.O in
+  Dune_rules.Dune_load.stanzas_in_dir dir
+  >>= Memo.Option.map ~f:(fun dune_file ->
+    Dune_file.find_stanzas dune_file Dune_rules.Library.key
+    >>| List.filter_map ~f:(fun { Library.sub_systems; name = _loc, name; _ } ->
+      Option.some_if
+        (Dune_rules.Sub_system_name.Map.mem
```
rgrinberg, line 71: Using the sub system is a bit too level for functionality like this. I think you should expose what you need from the `Dune_rules.Inline_tests` module.

Alizter:
Subsumed by #12689

PR #11985, opened 2025-07-06
Upgrade jsoo

hhugo:


rgrinberg:
Given that we need the most recent version of json to make these tests, could you add the appropriate lower bounds?

hhugo:
> Given that we need the most recent version of json to make these tests, could you add the appropriate lower bounds?

I'm assuming you meant jsoo. Lower bounds are in place, unless I'm missing something.

rgrinberg:
Yes I meant jsoo.

hhugo:
CI is stuck. Can some stop and restart the problematic job ?

PR #11987, opened 2025-07-06
melange: don't emit module group wrapper as a JS file in `melange.emit`

anmonteiro:
fixes test introduced in #11986


test/blackbox-tests/test-cases/melange/empty-entries-target-dir-exists.t
```diff
@@ -1,20 +0,0 @@
-Test (modules) field can be left empty
```
anmonteiro, line 1: this test was just bad, the target dir was only getting created because the wrapper module was still being generated.

rgrinberg (approved):

PR #11992, opened 2025-07-07
feat(tui): click to go to file

Alizter:
If a user message has an associated location, we can now click on messages to open the users `EDITOR` at that location.

TODO:
- [x] mouse events are not handed back to tui after exiting the `EDITOR`.
- [x] Some messages, especially from the OCaml compiler, don't have locations. -- Compiler messages have embedded locations. We can detect these and do a bit more work to extract a workable location.
- [ ] Better user feedback when editor is not set or the file cannot be opened
- [x] Instruct editor to go to location?

### DIFF
--- a/src/dune_tui/dune_tui.ml
+++ b/src/dune_tui/dune_tui.ml
@@ -2,36 +2,56 @@ open Import
 open Lwd.O
 module Unicode = Drawing.Unicode
 
-let create () = Term.create ~nosig:false ~output:Unix.stderr ()
 let bytes = Bytes.make 64 '0'
 let sigcont_pipe = lazy (Unix.pipe ~cloexec:true ())
 
-let term =
-  let setup =
-    lazy
-      (Unix.set_nonblock (Lazy.force sigcont_pipe |> fst);
-       let term = ref (create ()) in
-       Sys.set_signal Sys.sigcont
-       @@ Sys.Signal_handle
-            (fun _ ->
-              Term.release !term;
-              term := create ();
-              assert (1 = Unix.single_write (Lazy.force sigcont_pipe |> snd) bytes 0 1));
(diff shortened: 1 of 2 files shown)


src/dune_tui/dune_tui.ml
```diff
@@ -98,13 +158,62 @@ module Message_viewer = struct
...
+        (* TODO: subl, nano, ... ? *)
+        | _ -> [ prog; file_to_open ]
+      in
+      prog, argv
+    with
+    | Some (prog, argv) ->
+      Terminal_manager.disable ();
+      let () =
+        let _pid, exit_status =
+          Dune_spawn.Spawn.spawn ~prog ~argv () |> Unix.waitpid []
```
rgrinberg, line 197: I think this is going to break the scheduler since it's raising against this with `waitpid`. You need to tell the scheduler to ignore this pid.

src/dune_tui/dune_tui.ml
```diff
@@ -2,36 +2,56 @@ open Import
...
+                disable ();
+                match !old_sigtstp_handler with
+                | Sys.Signal_handle f -> f i
+                | _ -> Unix.kill (Unix.getpid ()) Sys.sigstop);
+      Sys.set_signal Sys.sigcont
+      @@ Sys.Signal_handle
+           (fun _ ->
+             if not (Option.is_some !term_ref)
+             then (
+               let _ = enable () in
```
rgrinberg, line 51: Why do we set a signal handler from inside a handler? This code could use some comments.

rgrinberg (changes requested):

Alizter:
Seems like a cool feature to have but will require some more thought probably. This can stay as a proof-of-concept.

PR #11996, opened 2025-07-08
pkg: add command to install dev tools

gridbugs:
Formerly the only way to install dev tools was by running them, leading to people coming up with workarounds for installing tools, such as running dev tools with "--version". This change adds a dedicated command for install dev tools without running them.

CC @pitag-ha


maiste (approved):
Looking good, thanks for the change! It will be really convenient to have :smile:

Sudha247 (commented):
Looks good to me! Agree with @maiste, it's convenient to have dev tools installed this way.

PR #11997, opened 2025-07-08
pkg: command to check if project intends to use dune package management

gridbugs:
Some users have been using the presence of the dune.lock directory to determine whether a project is locked. This is brittle as other lockdir names are possible. It also doesn't align with the mental model for lockdirs in dune in that dune projects may have multiple lockdirs, so the "lockedness" of a project depends on which lockdir you are interested in. Finally, there's no guarantee that the current default lockdir name or location is not stable; it may change in a future dune version.

This change introduces a new command `dune pkg is-locked` which prints whether each specified lockdir exists (using the same lockdir selection logic as `dune pkg lock`) and exits 0 only if all specified lockdirs exist. Using this command should be more stable than checking for the existence of dune.lock.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/pkg-enabled.t
+++ b/test/blackbox-tests/test-cases/pkg/pkg-enabled.t
@@ -1,0 +1,41 @@
+Exercise the "dune pkg enabled" command which checks whether package management
+should be used.
+
+  $ . ./helpers.sh
+
+  $ mkrepo
+
+  $ cat >dune-workspace <<EOF
+  > (lang dune 3.20)
+  > (lock_dir
+  >  (path dune.lock)
+  >  (repositories mock))
+  > (lock_dir
+  >  (path dune.other.lock)
+  >  (repositories mock))
+  > (repository
+  >  (name mock)
+  >  (url "file://$(pwd)/mock-opam-repository"))
+  > EOF
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.13)
+  > (package
+  >  (allow_empty)
+  >  (name foo))
+  > EOF
+
+When no lockdir is present pkg is not enabled:
+  $ dune pkg enabled
+  [1]
+
+When the default lockdir is present pkg is enabled:
+  $ dune pkg lock > /dev/null 2> /dev/null
+
+  $ dune pkg enabled
+
+  $ rm -r dune.lock
+
+When a non-default lockdir is present, pkg is still enabled:
+  $ dune pkg lock dune.other.lock > /dev/null 2> /dev/null
+  $ dune pkg enabled
--- a/bin/pkg/pkg.ml
+++ b/bin/pkg/pkg.ml
@@ -8,7 +8,12 @@ let man =
 ;;
 
 let subcommands =
-  [ Lock.command; Print_solver_env.command; Outdated.command; Validate_lock_dir.command ]
+  [ Lock.command
+  ; Print_solver_env.command
+  ; Outdated.command
+  ; Validate_lock_dir.command
+  ; Pkg_enabled.command
+  ]
 ;;
 
 let info name =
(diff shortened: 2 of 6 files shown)


Leonidas-from-XIV:
Shouldn't the command be more like `pkg-enabled`? Because whether a lock directory exists or not and where is somewhat more of an internal issue and if we want Dune to use lock files internally without exposing it to the user it doesn't really matter to the caller whether the lock file exists or not as it is Dune's responsibility to create it.

gridbugs:
I think I agree. Just confirming you're suggesting renaming the command to suggest that it prints whether the project intends to use dune package management rather than to suggest that it's currently locked?

I guess if I make that change I should also change the command to not print the status of each lockdir, and just communicate its output by its exit status. And maybe it should exit 0 if _any_ lockdir is present rather than if _all_ lockdirs are present, since it's just communicating the user's intention to use pkg rather than the existence of specific lockdirs. What do you think?

ElectreAAS:
I agree with Marek here, `pkg-enabled` is a good name, `dune pkg status` could also be chosen. And indeed it should say "yes package management is enabled in this project" or "no there is no indication that it is enabled" (with return codes obviously)
The existence of lockdirs / lockfiles is an implementation detail to me

test/blackbox-tests/test-cases/pkg/pkg-is-locked.t
```diff
@@ -0,0 +1,82 @@
...
+  $ cat >dune-workspace <<EOF
+  > (lang dune 3.13)
+  > (lock_dir
+  >  (path dune.lock)
+  >  (repositories mock))
+  > (lock_dir
+  >  (path dune.a.lock)
+  >  (repositories mock))
+  > (lock_dir
+  >  (path dune.a.lock)
```
ElectreAAS: `path dune.a.lock` is here twice, lines 11 & 14, is this intended?

Alizter:
What about `dune pkg enabled`?

Leonidas-from-XIV:
> I think I agree. Just confirming you're suggesting renaming the command to suggest that it prints whether the project intends to use dune package management rather than to suggest that it's currently locked?

Exactly. My idea is that it would tell you whether `dune build` or the dev-tools packages would use package management or whether the switch/system state would be used thus the environment is expected to have the required tools. I believe that this is closer to the question that editor tooling people are trying to answer (editor tooling people, please comment if I'm wrong).

The comments here have pointed out some great solution, I really like @Alizter's suggestion of `dune pkg enabled` as it is very memorable and @gridbugs idea with exit code 0 meaning "yes, it's enabled" and nonzero otherwise as it is very Unix-y. That way the typical shell patterns are enabled, e.g. `dune pkg enabled && some-command that-needs dune-pkg`.

This might also be useful as we might change the criterion of enabling package management from "a lock dir exists in the repository" to something else (env var? `dune-workspace` entry? TBD in a different issue).

gridbugs:
Updated based on feedback. The command is now `dune pkg enabled`, and it prints no output and communicates entirely via its exit status. Iff any of the lockdirs specified in dune-workspace are present this command will exit 0.

test/blackbox-tests/test-cases/pkg/pkg-enabled.t
```diff
@@ -0,0 +1,41 @@
+Exercise the "dune pkg enabled" command which checks whether package management
+should be used.
+
+  $ . ./helpers.sh
+
+  $ mkrepo
+
+  $ cat >dune-workspace <<EOF
+  > (lang dune 3.13)
```
Leonidas-from-XIV: Should probably be dune 3.20? I don't think 3.13 had `lock_dir`

(We probably forgot to add a version gate to `lock_dir`, but that's not an issue for this PR)

Leonidas-from-XIV (approved):
Looks good, code is fine, the test is checking a decent range of things and I think the semantics are nice.

I have one comment about the dune-lang version that I think should be fixed, but I'll approve anyway so you can merge at your convenience.

ElectreAAS (approved):
LGTM!

PR #11998, opened 2025-07-08
Add reproduction test for dune-site and dune pkg

n-osborne:


test/blackbox-tests/test-cases/pkg/sites-plugin.t
```diff
@@ -0,0 +1,161 @@
...
+
+  $ . ./helpers.sh
+  $ mkrepo
+  $ add_mock_repo_if_needed
+
+Make an executable using dune-site (example mostly from the manual)
+  $ mkdir app
+  $ cd app
+  $ cat > dune-project <<EOF
+  > (lang dune 3.17)
```
maiste:
```suggestion
  > (lang dune 3.20)
```

test/blackbox-tests/test-cases/pkg/sites-plugin.t
```diff
@@ -0,0 +1,161 @@
...
+  >  checksum: [
+  >   "md5=$(md5sum app.tar | cut -f1 -d' ')"
+  >  ]
+  > }
+  > EOF
+
+Make a project that uses the executable in a dune rule:
+
+  $ cat > dune-project <<EOF
+  > (lang dune 3.17)
```
maiste:
```suggestion
  > (lang dune 3.20)
```

maiste (commented):
Just two quick suggestions as I was quickly looking at the PR.

test/blackbox-tests/test-cases/pkg/sites-plugin.t
n-osborne: Thanks.

test/blackbox-tests/test-cases/pkg/sites-plugin.t
n-osborne: Thanks.

Leonidas-from-XIV:
@n-osborne Can you promote the expected outputs so the repro-case succeeds?

test/blackbox-tests/test-cases/pkg/sites-plugin.t
```diff
@@ -0,0 +1,165 @@
...
+  $ cat > dune <<EOF
+  > (library
+  >  (public_name foo))
+  > 
+  > (rule
+  >  (alias runtest)
+  >  (action
+  >   (run app)))
+  > EOF
+
```
Leonidas-from-XIV, line 152: Could you add a note here that what was expected is that the plugin would print something and it doesn't? So if we fix this, we know this test changes and that that change is correct.

test/blackbox-tests/test-cases/pkg/sites-plugin.t
```diff
@@ -0,0 +1,165 @@
...
+  >     "dune"
+  >     "build"
+  >     "-p"
+  >     name
+  >     "-j"
+  >     jobs
+  >     "@install"
+  >     "@runtest" {with-test}
+  >     "@doc" {with-doc}
+  >   ]
```
Leonidas-from-XIV, line 100: Can you shorten this? `dune subst` is unnecessary and it really just needs `dune build -p name`. The shorter the test the more focus there is on the actual issue instead of boilerplate.

test/blackbox-tests/test-cases/pkg/sites-plugin.t
```diff
@@ -0,0 +1,165 @@
...
```
Leonidas-from-XIV, line 119: Same here.

test/blackbox-tests/test-cases/pkg/sites-plugin.t
```diff
@@ -0,0 +1,165 @@
...
+  >  (alias runtest)
+  >  (action
+  >   (run app)))
+  > EOF
+
+  $ dune pkg lock
+  Solution for dune.lock:
+  - app.0.0.1
+  - plugin1.0.0.1
+  $ dune build
```
Leonidas-from-XIV: That line is unnecessary, if runtest is called it will build the dependencies anyway.

test/blackbox-tests/test-cases/pkg/sites-plugin.t
```diff
@@ -0,0 +1,165 @@
...
+  >  (libraries plugin1.plugin1_impl)
+  >  (site (app plugins)))
+  > EOF
+  $ cat > plugin/plugin1_impl.ml <<EOF
+  > let () =
+  > print_endline "Registration of Plugin1";
+  > Queue.add (fun () -> print_endline "Plugin1 is doing something...") Registration.todo
+  > EOF
+  $ dune build @all 2>&1 | dune_cmd sanitize
+  $ dune exec ./app.exe
```
Leonidas-from-XIV: I think it would be nicer to attach this to `runtest`, because as is you're comparing behavior of `dune exec` and `dune build @runtest` and while I don't think the bug has anything to do with `exec` vs `build`, it seems odd to compare one invocation working with a completely different invocation not working.

Or alternatively, use `dune exec` below, which is probably easier to do as it doesn't need any configuration for tests.

Leonidas-from-XIV (commented):
I think the test is fine, thanks for submitting it. Looking through the code I can see some areas that can be improved - none of them critical but I think they would make the test nicer and more readable.

paurkedal:
I think there is something wrong with this test as it looks like the installation directory of the `plugin` package only contains a `cookie` file.  The following unsanitized addition to `test/blackbox-tests/test-cases/pkg/sites-plugin.t` shows the problem:
```
So, what went wrong?  Looks like `plugin1.plugin1_impl` is not installed
according to ocamlfind either:
  $ dune pkg print-digest app
  app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8
  $ dune pkg print-digest plugin1
  plugin1.0.0.1-534f232c02a12b97bc94968593b595b6
  $ dune exec ocamlfind query app
  $TESTCASE_ROOT/_build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app
  $ dune exec ocamlfind query plugin1.plugin1_impl
  ocamlfind: Package `plugin1.plugin1_impl' not found
  [2]

And not found in the sandbox installation:
  $ find _build -path '*/target/*'
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app/register
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app/register/registration.cmt
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app/register/registration.cmi
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app/register/registration.cma
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app/register/registration.cmxs
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app/register/registration.cmx
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app/register/registration.a
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app/register/registration.cmxa
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app/register/registration.ml
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app/META
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/lib/app/dune-package
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/cookie
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/bin
  _build/_private/default/.pkg/app.0.0.1-1a2666e26dd98cd6fc09d2dc6e2b24f8/target/bin/app
  _build/_private/default/.pkg/plugin1.0.0.1-534f232c02a12b97bc94968593b595b6/target/cookie
```

Alizter:
@paurkedal That's correct. This test was (and still is) asserting that the behaviour of dune site plugins from packages is broken. This has not been fixed yet and has not been a priority to do so. If this is something that you would like please let it be known in https://github.com/ocaml/dune/issues/11999

paurkedal:
What I mean is that I think the test is also not working as intended in the sense that there seems to be an additional issue with the installation of the plugin1 package.  At least I would expect  `plugin1.plugin1_impl` to have been installed as a regular library and to be accessible with ocamlfind.  That is, my worry is that the test will not indicate when the issue has been solved.

Alizter:
@paurkedal This test appears to be mostly copy-paste from an earlier sites test and can be improved. Could you send in a PR where it asserts the more informative behaviour? You can pipe `| censor` using our helpers available in all our cram tests to hide the digest.

PR #12000, opened 2025-07-08
chore: add release template

maiste:
It would be nice to add this template to make it easy to have the list of things to remember when doing a dune release. This is mostly to not have to copy and paste from other release issues and "formalize a bit" the list.

Having the template will help me to be faster and not forget anything.


Leonidas-from-XIV (approved):
Seems sensible to me and if it helps to create tracking issues without having to hunt down the text from the previous one, I'm in favor of it.

Also congrats on the round issue number :-)

PR #1963, opened 2019-03-21
Memoize the evaluation of predicate

rgrinberg:
Signed-off-by: Rudi Grinberg <rudi.grinberg@gmail.com>

PR #12002, opened 2025-07-08
feat(oxcaml): support parameter implementation

maiste:
Closes https://github.com/ocaml/dune/issues/12085

This PR is the second of a set of four PRs related to the introduction of Parameterized Libraries in Dune.

:warning: This feature is only available with OxCaml until it is upstream to the compiler.

- It adds support for parameter in the `implements` field.
- It provides some tests to make sure the functionality act according to the specification. They are still guarded by the oxcaml variable.
- It extends the documentation with our case. More explanation about the feature will be addressed in future PRs, specific to the documentation.

## TODO

- [x] It needs the parameter to be merged first ✅ 
- [x] Complete https://github.com/ocaml/dune/issues/12253 and rebase on the merged changes 
- [x] Consider whether `is_parameter` should be removed (as per https://github.com/ocaml/dune/pull/12298#issuecomment-3218548129).

### DIFF
--- a/test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
+++ b/test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
@@ -45,6 +45,12 @@ fail with the correct error message.
   >  (implements missing_foo))
   > EOF
   $ dune build
+  File "foo_impl/dune", line 3, characters 13-24:
+  3 |  (implements missing_foo))
+                   ^^^^^^^^^^^
+  Error: Library "missing_foo" not found.
+  -> required by alias default
+  [1]
 
 We implements the parameter using library calling a wrong parameter name.
 
@@ -55,6 +61,12 @@ We implements the parameter using library calling a wrong parameter name.
   >  (implements missing_foo))
   > EOF
   $ dune build
+  File "foo_impl/dune", line 3, characters 13-24:
+  3 |  (implements missing_foo))
+                   ^^^^^^^^^^^
+  Error: Library "missing_foo" not found.
+  -> required by alias default
+  [1]
 
 We implement the parameter using a library with a correct parameter this time.
 
--- a/test/blackbox-tests/test-cases/virtual-libraries/impl-not-virtual-external.t
+++ b/test/blackbox-tests/test-cases/virtual-libraries/impl-not-virtual-external.t
@@ -20,6 +20,7 @@ an appropriate error message.
   File "dune", line 1, characters 21-29:
   1 | (library (implements foodummy) (name bar))
                            ^^^^^^^^
-  Error: Library "foodummy" is not virtual. It cannot be implemented by "bar".
+  Error: Library "foodummy" is not virtual nor a library parameter. It cannot
+  be implemented by "bar".
   Leaving directory 'test'
   [1]
--- a/test/blackbox-tests/test-cases/virtual-libraries/impl-not-virtual.t/run.t
+++ b/test/blackbox-tests/test-cases/virtual-libraries/impl-not-virtual.t/run.t
@@ -4,6 +4,7 @@ appropriate error message.
   File "impl/dune", line 3, characters 13-16:
   3 |  (implements lib))
                    ^^^
-  Error: Library "lib" is not virtual. It cannot be implemented by "impl".
+  Error: Library "lib" is not virtual nor a library parameter. It cannot be
+  implemented by "impl".
   -> required by alias default in dune:1
   [1]


src/dune_rules/lib.ml
```diff
@@ -947,9 +948,15 @@ end = struct
           let open Resolve.Memo.O in
           let* vlib = resolve_forbid_ignore name in
           let virtual_ = Lib_info.virtual_ vlib.info in
-          match virtual_ with
-          | false -> Error.not_virtual_lib ~loc ~impl:info ~not_vlib:vlib.info
-          | true -> Resolve.Memo.return vlib
+          let parameterized = Lib_info.is_parameter vlib.info in
+          match virtual_, parameterized with
```
art-w: nitpick: I read the variable name `parameterized` as "requires parameters" instead of "is a parameter"

src/dune_rules/stanzas/parameter.ml
```diff
@@ -27,7 +27,7 @@ let to_library t =
   ; library_flags = Ordered_set_lang.Unexpanded.standard
   ; c_library_flags = Ordered_set_lang.Unexpanded.standard
   ; virtual_deps = []
-  ; wrapped = This (Simple false)
+  ; wrapped = This (Simple true)
```
art-w: Should this change be in #1963 since it impacts the `library_parameter` stanza opam installation?

art-w (changes requested):
Great! Can we add a test with a library that implements a parameter that doesn't exist? (this should be caught by the existing virtual modules logic, but the error message may need an update)

src/dune_rules/stanzas/parameter.ml
```diff
@@ -27,7 +27,9 @@ let to_library t =
   ; library_flags = Ordered_set_lang.Unexpanded.standard
   ; c_library_flags = Ordered_set_lang.Unexpanded.standard
   ; virtual_deps = []
-  ; wrapped = This (Simple false)
+  ; wrapped =
+      This (Simple true)
+      (* We set it as Simple true because, otherwise, we can extract the Singleton main module name. *)
```
maiste:
```suggestion
      (* We set it as Simple true because, otherwise, we can't extract the Singleton main module name. *)
```

src/dune_rules/lib.ml
```diff
@@ -947,9 +949,15 @@ end = struct
           let open Resolve.Memo.O in
           let* vlib = resolve_forbid_ignore name in
           let virtual_ = Lib_info.virtual_ vlib.info in
-          match virtual_ with
-          | false -> Error.not_virtual_lib ~loc ~impl:info ~not_vlib:vlib.info
-          | true -> Resolve.Memo.return vlib
+          let is_parameter = Lib_info.is_parameter vlib.info in
```
rgrinberg, line 952: Since being a virtual library and a parameter is mutually exclusive, what about introducing a type to reflect that and returning it from some sort of a `Lib_info.kind` function. The sooner this invariant is enforced, the easier it will be to make sure we never make this mistake.

src/dune_rules/lib_rules.ml
```diff
@@ -483,6 +483,27 @@ let setup_build_archives (lib : Library.t) ~top_sorted_modules ~cctx ~expander ~
     (fun () -> build_shared ~native_archives ~sctx lib ~dir ~flags)
 ;;
 
+let parameterize_root_module ~parameter ~root_module_name modules =
```
rgrinberg, line 486: Instead of storing this information in the module, how about just add a field to the compilation context designating which module is implementing a parameter. Should be easier to understand and more efficient.

src/dune_rules/lib_rules.ml
```diff
@@ -483,6 +483,27 @@ let setup_build_archives (lib : Library.t) ~top_sorted_modules ~cctx ~expander ~
...
+  let* parameter_module_name =
+    let+ name = Lib.main_module_name parameter in
+    match Resolve.to_result name with
+    | Ok (Some name) -> name
+    | Ok None ->
+      (* The parameter must have a module name otherwise it can't exist. *)
+      assert false
+    | Error err -> Resolve.raise_error_with_stack_trace err
```
rgrinberg, line 494: This error needs to be delayed rather than forced.

rgrinberg (changes requested):

src/dune_rules/lib.ml
maiste: Do you mean that we should duplicate the virtual type with a parameter equivalent or do you mean we should have a `Parameter` kind? 

If the former, `Lib_info.is_parameter is just an alias for `t.kind = parameter`.

rgrinberg: I argue that we need a function as follows:

```
val lib_kind : [`Normal | `Virtual | `Parameter | `Ppx_deriver ..]
```

The function `Lib_info.is_parameter` shouldn't be necessary at all.

Now you might say that we've done a similar sort of mistake with `Lib_info.virtual_`. If we didn't have this function, I think it would be simpler for us to be confident that we're correctly making this change everywhere.

test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
```diff
@@ -0,0 +1,135 @@
+This test ensures the `implements` field is working in the context of
+parameterized libraries.
+
+  $ . ./helpers.sh
+
+We create two parameters, on public, the other one local.
```
shonfeder, line 6:
```suggestion
We create two parameters, one public, the other one local.
```

test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
```diff
@@ -0,0 +1,135 @@
...
+  >  (name foo_lib))
+  > EOF
+  $ cat > "foo_lib/foo_lib.ml" <<EOF
+  > let x = 42
+  > EOF
+  $ dune build
+
+
+We implements a non parameter library (neither a virtual module). It should
+fail with the correct error message.
```
shonfeder, line 39:
```suggestion
Supplying a non-parameter, regular library (not a virtual module), as an implementation should
fail with the correct error message.
```

test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
```diff
@@ -0,0 +1,135 @@
...
+  > (library
+  >  (name foo_impl)
+  >  (implements missing_foo))
+  > EOF
+  $ dune build
+  File "foo_impl/dune", line 3, characters 13-24:
+  3 |  (implements missing_foo))
+                   ^^^^^^^^^^^
+  Error: Library "missing_foo" not found.
+  -> required by alias default
```
shonfeder, line 52: Caveat that I am still getting up to speed here, but is this the right error message? I would rather expect something like `parameter interface not found`.

test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
```diff
@@ -0,0 +1,135 @@
...
+  [1]
+
+We implements the parameter using library calling a wrong parameter name.
```
shonfeder, line 55:
```suggestion
We implement the parameter using library calling a wrong parameter name.
```

test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
```diff
@@ -0,0 +1,135 @@
...
+  >  (name foo_impl)
+  >  (implements missing_foo))
+  > EOF
+  $ dune build
+  File "foo_impl/dune", line 3, characters 13-24:
+  3 |  (implements missing_foo))
+                   ^^^^^^^^^^^
+  Error: Library "missing_foo" not found.
+  -> required by alias default
+  [1]
```
shonfeder, line 69: This looks like the exact same test as the one on on lines 41 to 53. Perhaps a needed update to differentiate them was missed? Or am I missing something on my end?

doc/reference/dune/library.rst
```diff
@@ -210,6 +210,13 @@ order to declare a multi-directory library, you need to use the
    unless you use Dune to synthesize the ``depends`` and ``depopts`` sections
    of your opam file.
 
+.. describe:: (implements <name>)
+
+   ``name`` defines the name of the virtual library or the parameter you are
```
shonfeder, line 215:
```suggestion
   ``name`` defines the name of the virtual library or the parameter library you are
```

src/dune_rules/lib.ml
```diff
@@ -241,7 +241,8 @@ module Error = struct
     make
       ~loc
       [ Pp.textf
-          "Library %S is not virtual. It cannot be implemented by %S."
+          "Library %S is not virtual nor a library parameter. It cannot be implemented \
```
shonfeder, line 244:
```suggestion
          "Library %S is neither a virtual library nor a library parameter. It cannot be implemented \
```

src/dune_rules/lib.ml
shonfeder: Would "the right way" to do to this be to extend https://github.com/maiste/dune/blob/a45b410b28dd1d8de56f3a5a0f5bdb0a460aae5a/src/dune_lang/lib_kind.mli#L16-L20 to include a variant for `Virtual`? Or is virtual not disjoint with the alternatives?

shonfeder:
I've assigned myself here since I'll be taking this over from @maiste. My initial review comments here just notes to self (with the exception of https://github.com/ocaml/dune/pull/12002#discussion_r2286312433, which is a question for @rgrinberg).

test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
maiste: As the user provides an element which doesn't exist, we can't differentiate if it is a virtual library or a parameter. We don't have the information to characterize the element.

src/dune_rules/lib.ml
rgrinberg: `Virtual` should be disjoint from the rest.

test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
shonfeder: Ok, so we cannot differentiate these two cases, but then I supposed it doesn't help anything to test the same scenario twice?

shonfeder:
I don't seem to be able to push to this branch, so I am inclined to just replace this with a new PR. I'll copy the context over. I'll do it tomorrow, unless I get objections.

maiste:
I don't see a problem with opening a new PR. However, you should be able to push on the branch unless you don't have sufficient rights on the repository.
image

test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
maiste: Initially, it was to make sure we fail if the `implements` was fed, either with a non-existent parameter (or library) or a non parameterized library.

It depends if we go to a strict separation with `Virtual` or not. If we do, we might want to keep this test to make sure we are not using a library that shouldn't be used.

src/dune_rules/lib.ml
```diff
@@ -724,8 +725,9 @@ end = struct
         let rec loop acc = function
           | [] -> Resolve.Memo.return acc
           | (lib, stack) :: libs ->
+            let is_parameter = Lib_info.is_parameter lib.info in
```
rgrinberg, line 728: This function shouldn't be used as there's a good alternative now.

rgrinberg (changes requested):

shonfeder:
I have a branch locally updating this on top of. https://github.com/ocaml/dune/issues/12253 I will push to this or replace it tomorrow. Just ran out of time.

shonfeder:
Thanks @maiste. I think I must not have maintainer perms, because I just get permission denied errors consistently.

shonfeder:
Superseded by https://github.com/ocaml/dune/pull/12392

PR #12003, opened 2025-07-09
pkg: enroll ocamlearlybird as a dev tool

gridbugs:
CC @pitag-ha


bin/tools/tools.ml
```diff
@@ -3,15 +3,25 @@ open! Import
 module Exec = struct
   let doc = "Command group for running wrapped tools."
   let info = Cmd.info ~doc "exec"
-  let group = Cmd.group info [ Ocamlformat.Exec.command; Ocamllsp.Exec.command ]
+
+  let group =
+    Cmd.group
+      info
+      [ Tools_common.exec_command Ocamlformat
+      ; Ocamllsp.Exec.command
```
maiste: Isn't it possible to generalize the lsp command too?

bin/tools/tools.ml
```diff
@@ -3,15 +3,25 @@ open! Import
...
 
 module Install = struct
   let doc = "Command group for installing wrapped tools."
   let info = Cmd.info ~doc "install"
 
   let group =
-    Cmd.group info (List.map [ Ocamlformat; Ocamllsp ] ~f:Tools_common.install_command)
+    Cmd.group
+      info
+      (List.map [ Ocamlformat; Ocamllsp; Ocamlearlybird ] ~f:Tools_common.install_command)
```
maiste: Maybe nitpick but what do you think about extracting the list somewhere else as "available dev tools"?

Sudha247: I agree, it would be a great addition to be able to see which dev tools are available. Maybe it already exists?

Sudha247 (commented):
LGTM!

bin/tools/tools.ml
gridbugs: For some reason I thought it couldn't, but looking at its implementation it should be fine to generalize here.

bin/tools/tools.ml
gridbugs: It's in the output of `dune tools exec --help`.

bin/tools/tools.ml
gridbugs: Done.

src/dune_pkg/dev_tool.ml
```diff
@@ -5,30 +5,34 @@ type t =
...
 
 let equal a b =
   match a, b with
   | Ocamlformat, Ocamlformat -> true
   | Odoc, Odoc -> true
   | Ocamllsp, Ocamllsp -> true
   | Utop, Utop -> true
-  | (Ocamlformat | Odoc | Ocamllsp | Utop), _ -> false
+  | Ocamlearlybird, Ocamlearlybird -> true
+  | (Ocamlformat | Odoc | Ocamllsp | Utop | Ocamlearlybird), _ -> false
```
Leonidas-from-XIV:
```suggestion
  | _,  _ -> false
```
I don't think there's much of a point in enumerating them all here, not in the `equal` function.

bin/tools/tools_common.ml
```diff
@@ -102,3 +102,28 @@ let install_command dev_tool =
...
+    let common, config = Common.init builder in
+    lock_build_and_run_dev_tool ~common ~config dev_tool ~args
+  in
+  let info =
+    let doc =
+      sprintf
+        {|Wrapper for running %s intended to be run automatically
+       by a text editor. All positional arguments will be passed to the
+       %s executable (pass flags to %s after the '--'
+       argument, such as 'dune %s -- --help').|}
```
Leonidas-from-XIV: Is the message tested anywhere? Because it seems like it should state `dune tools %s -- --help`.

Leonidas-from-XIV (commented):
I like that the tools are getting joined together and that they don't need their own command line parser anymore. That will make adding new one if we chose to much easier.

src/dune_pkg/dev_tool.ml
Alizter: FWIW these are both not the best way to write an `equal` function since adding a new constructor to the variant won't create an error. What you want is:
```ocaml
match a, b with
| Ocamlformat, Ocamlformat -> true
| Ocamlformat, _ | _, Ocamlformat -> false
...
```

That way it is always exhaustive and less prone to error.

ElectreAAS (approved):
TIL what ocamlearlybird is.
Good addition, the code looks lean; I agree with Marek on the need for a simple test

src/dune_pkg/dev_tool.ml
gridbugs: @Alizter in my original code I don't understand how adding a new constructor won't create an error. Can you explain please. I wrote it the current way precisely so that adding a new constructor will create an error because I spent half a day debugging a situation where exactly that happened.

bin/tools/tools_common.ml
gridbugs: Good spot. Fixed and added a test.

src/dune_pkg/dev_tool.ml
Alizter: If you forget to add `| Ocamlformat, Ocamlformat -> true` to your way, you will not get a compiler error and it will be harder to spot. That's why in situations where we are adding many variants, catch all cases can be a bit of a foot gun.

The advantage of the way I proposed, although it can be a bit verbose, is that it keeps the concerns a bit more modular. When you add a constructor, you have to add two lines, rather than update existing code. Adding a constructor will always error, and its easier to add two lines rather than update two separate places.

PR #12006, opened 2025-07-10
fix(melange): allow `melange.runtime_deps` to depend on dir targets

anmonteiro:


anmonteiro:
hey @rgrinberg is there another way of detecting whether a path is under a directory target without using `Load_rules.*` functions?

The test that's failing is a self-dependency cycle when we're trying to decide whether something in the melange target dir is a directory target, while loading/generating rules for that dir.

I can think of the workaround of never actually doing this logic if we're under the melange target directory, but if it could be supported I'd love to know.

rgrinberg:
I think you can call the Gen_rules callback directly and look at its directory_targets field to without forcing the rules.

anmonteiro:
Maybe I'm not following, but I tried something like this in `Load_rules`:

```ocaml
  let dir_targets p =
    match Path.parent p with
    | None -> Memo.return Path.Build.Set.empty
    | Some dir ->
      get_dir_triage ~dir
      >>= (function
       | Known _ -> Memo.return Path.Build.Set.empty
       | Build_directory d ->
         Gen_rules.gen_rules d
         >>| (function
          | Under_directory_target { directory_target_ancestor } ->
            Path.Build.Set.singleton directory_target_ancestor
          | Normal { directory_targets; _ } -> Path.Build.Set.of_keys directory_targets))
  ;;
```

to return the set of directory targets, and I still get a dep cycle:

```
+|  Internal error, please report upstream including the contents of _build/log.
+|  Description:
+|    ("internal dependency cycle",
+|     { frames = [ ("gen-rules", In_build_dir "default/output") ] })
+|  Raised at Memo.Exec.exec_dep_node.(fun) in file "src/memo/memo.ml", line
+|    1289, characters 29-62
+|  Called from Fiber__Scheduler.exec in file "src/fiber/src/scheduler.ml", line
+|    76, characters 8-11
+|  Re-raised at Stdune__Exn.raise_with_backtrace in file
+|    "otherlibs/stdune/src/exn.ml", line 38, characters 27-56
+|  Called from Fiber__Scheduler.exec in file "src/fiber/src/scheduler.ml", line
+|    76, characters 8-11
+|  Re-raised at Stdune__Exn.raise_with_backtrace in file
+|    "otherlibs/stdune/src/exn.ml", line 38, characters 27-56
+|  Called from Fiber__Scheduler.exec in file "src/fiber/src/scheduler.ml", line
+|    76, characters 8-11
+|  Re-raised at Stdune__Exn.raise_with_backtrace in file
+|    "otherlibs/stdune/src/exn.ml", line 38, characters 27-56
+|  Called from Fiber__Scheduler.exec in file "src/fiber/src/scheduler.ml", line
+|    76, characters 8-11
+|  Re-raised at Stdune__Exn.raise_with_backtrace in file
+|    "otherlibs/stdune/src/exn.ml", line 38, characters 27-56
+|  Called from Fiber__Scheduler.exec in file "src/fiber/src/scheduler.ml", line
+|    76, characters 8-11
+|  -> required by ("load-dir", In_build_dir "default/output")
+|  -> required by ("toplevel", ())
+|
```

anmonteiro:
I managed to make it work without cycles by using the `Dir_status` module.

Issue #11958, opened 2025-07-02
[RPC] Allow formatting while watch server is running


ElectreAAS:
## Desired Behavior

In one terminal `dune build -w`, in another `dune fmt` (or `dune rpc fmt`) and that works

## Context

This would follow in the footsteps of #11808, #11712, #11840, and #11900

rgrinberg:
As a start, you might want to just allow `dune promote` to work while RPC running. Doing fmt should be a natural extension of that.

Alizter:
@rgrinberg That already seems to be the case today. `dune promote` doesn't care if the `_build` directory is locked when `dune build -w` is running.

rgrinberg:
Unless things have changed recently, it does not work correctly. It's still possible for the two invocations of dune to step over each other while editing the promotion database. There needs to something to synchronize such access.

ElectreAAS:
Solved by #12064, and to a lesser degree, #12010.

PR #12010, opened 2025-07-11
[RPC] Allow promotion while watch server is running

ElectreAAS:
This PR is a first step towards implementing #11958.
The rpc system was already prepared for this kind of request, but there was no client firing up said request, so I added it.

This PR contains stolen code from Steve's PR #11900 because he added utility functions, but is now rebased on main.


bin/promotion.ml
```diff
@@ -50,12 +50,56 @@ module Apply = struct
...
+  ;;
+
   let term =
     let+ builder = Common.Builder.term
     and+ files = Arg.(value & pos_all Cmdliner.Arg.file [] & info [] ~docv:"FILE") in
-    let common, _config = Common.init builder in
-    let files_to_promote = files_to_promote ~common files in
-    Diff_promotion.promote_files_registered_in_last_run files_to_promote
+    let common, config = Common.init builder in
+    match Dune_util.Global_lock.lock ~timeout:None with
```
ElectreAAS, line 58: Tagging @Alizter to mention that this `lock` check is what was missing for synchronisation

otherlibs/dune-rpc/private/procedures.ml
```diff
@@ -52,7 +52,10 @@ module Public = struct
   end
 
   module Promote = struct
-    let v1 = Decl.Request.make_current_gen ~req:Path.sexp ~resp:Conv.unit ~version:1
+    let v1 =
+      Decl.Request.make_current_gen ~req:(Conv.list Path.sexp) ~resp:Conv.unit ~version:1
```
rgrinberg: I think you either need to introduce a new generation of this RPC or add a separate RPC. Existing RPC shouldn't be modified like this.

otherlibs/dune-rpc/v1.mli
```diff
@@ -327,7 +327,7 @@ module Request : sig
   val format_dune_file : (Path.t * [ `Contents of string ], string) t
 
   (** Promote a file. *)
```
shonfeder:
```suggestion
  (** Promote files. *)
```

otherlibs/dune-rpc/private/procedures.ml
shonfeder: Following the [style guidelines](https://dune.readthedocs.io/en/latest/hacking.html#general-guidelines) for short, sweet interfaces, would a new version be preferable?

bin/rpc/rpc_common.ml
```diff
@@ -40,3 +42,33 @@ let wait_term =
   let doc = "poll until server starts listening and then establish connection." in
   Arg.(value & flag & info [ "wait" ] ~doc)
 ;;
+
+let establish_connection () =
+  let where = active_server () in
+  match where with
+  | Error e -> Fiber.return (Error e)
+  | Ok where -> Client.Connection.connect where
```
shonfeder: Is there a particular reason to introduce the first temporary `where` variable and the subsequent shadowing rather than doing

```suggestion
  match active_server () with
  | Error e -> Fiber.return (Error e)
  | Ok where -> Client.Connection.connect where
```

bin/rpc/rpc_common.ml
```diff
@@ -40,3 +42,33 @@ let wait_term =
...
+  let rec loop () =
+    establish_connection ()
+    >>= function
+    | Ok x -> Fiber.return x
+    | Error _ ->
+      let* () = Scheduler.sleep ~seconds:pause_between_retries_s in
+      loop ()
+  in
+  loop ()
+;;
```
shonfeder, line 70: Should there be a limit on how many retries are attempted?

bin/rpc/rpc_common.ml
```diff
@@ -40,3 +42,33 @@ let wait_term =
...
+
+let establish_connection_exn () =
+  let open Fiber.O in
+  establish_connection () >>| User_error.ok_exn
+;;
+
+let establish_connection_with_retry () =
+  let open Fiber.O in
+  let pause_between_retries_s = 0.2 in
```
shonfeder, line 60: I appreciate the use of a clearly named variable to avoid a mysterious magic number!

bin/rpc/rpc_common.ml
```diff
@@ -40,3 +42,33 @@ let wait_term =
...
+    >>= function
+    | Ok x -> Fiber.return x
+    | Error _ ->
+      let* () = Scheduler.sleep ~seconds:pause_between_retries_s in
+      loop ()
+  in
+  loop ()
+;;
+
+let establish_client_session ~wait =
```
shonfeder, line 72: It seems to me that `wait` would be a bit ambiguous. Might it be clearer as something like

```suggestion
let establish_client_session ~use_retry =
```

bin/rpc/rpc_common.mli
```diff
@@ -15,5 +16,7 @@ val request_exn
 (** Cmdliner term for a generic RPC client. *)
 val client_term : Common.Builder.t -> (unit -> 'a Fiber.t) -> 'a
 
+val establish_client_session : wait:bool -> Dune_rpc_client.Client.Connection.t Fiber.t
+
```
shonfeder: A doc comment here please! It would be nice to indicate that it will keep retrying with a fixed wait time, with no end (or after n tries, if that is adopted).

src/dune_rpc_impl/server.ml
```diff
@@ -390,10 +390,22 @@ let handler (t : _ t Fdecl.t) handle : 'build_arg Dune_rpc_server.Handler.t =
...
+        let files = List.map files ~f:For_handlers.source_path_of_string in
+        let on_missing fn =
+          User_warning.emit
+            [ Pp.textf
+                "Nothing to promote for %s."
+                (Path.Source.to_string_maybe_quoted fn)
+            ]
+        in
+        These (files, on_missing)
+    in
```
shonfeder: This behavior should make it into the documentation, I think: it is not obvious that that `promote []` means "promote everything".

shonfeder: I think this is just a stylistic remark so feel free to disregard, but IMO, the function definition here doesn't improve readability over using the same name as a variable in a match like

```
let files_to_promote = match paths with
  | [] -> All
  ...
```

src/dune_rpc_impl/server.ml
```diff
@@ -390,10 +390,22 @@ let handler (t : _ t Fdecl.t) handle : 'build_arg Dune_rpc_server.Handler.t =
...
     let f _ path =
```
shonfeder, line 399:
```suggestion
    let f _ paths =
```

src/dune_rpc_impl/server.ml
```diff
@@ -390,10 +390,22 @@ let handler (t : _ t Fdecl.t) handle : 'build_arg Dune_rpc_server.Handler.t =
...
-      let files = For_handlers.source_path_of_string path in
-      Promote.Diff_promotion.promote_files_registered_in_last_run
-        (These ([ files ], ignore));
+      let files = files_to_promote path in
+      Promote.Diff_promotion.promote_files_registered_in_last_run files;
```
shonfeder:
```suggestion
      let files = files_to_promote paths in
      Promote.Diff_promotion.promote_files_registered_in_last_run files;
```

test/expect-tests/dune_rpc_e2e/dune_rpc_diagnostics.ml
```diff
@@ -652,7 +652,7 @@ let%expect_test "promoting dune files" =
           Build (alias foo) failed |}];
       print_endline "attempting to promote";
       let+ res =
-        request_exn client Request.promote Dune_rpc.Path.(relative dune_root fname)
+        request_exn client Request.promote [ Dune_rpc.Path.(relative dune_root fname) ]
```
shonfeder: Not sure where is the best place for a comment on this, but would it make sense to have a test checking the behavior of the RPC when given an empty list of paths? Less to test correctness than than to guard against regressions in future changes...

bin/promotion.ml
```diff
@@ -50,12 +50,56 @@ module Apply = struct
...
+  ;;
+
+  let promote_via_rpc_server files_to_promote =
+    let open Fiber.O in
+    remote_promote files_to_promote
+    >>| function
+    | Error (error : Dune_rpc_private.Response.Error.t) ->
+      Printf.eprintf
+        "Error: %s\n%!"
+        (Dyn.to_string (Dune_rpc_private.Response.Error.to_dyn error))
```
shonfeder: Do we need to ensure this error is forwarded somehow beyond just an error message? E.v., via an exception or error to get an error exit code?

bin/promotion.ml
```diff
@@ -50,12 +50,56 @@ module Apply = struct
...
+          (let files_to_promote = files_to_promote ~common files in
+           Diff_promotion.promote_files_registered_in_last_run files_to_promote))
+    | Error lock_held_by ->
+      Scheduler.go_without_rpc_server ~common ~config (fun () ->
+        if not (Common.Builder.equal builder Common.Builder.default)
+        then
+          User_warning.emit
+            [ Pp.textf
+                "Your request is being forwarded to a running Dune instance%s so most \
+                 command-line arguments will be ignored."
```
shonfeder: It seems to me like this is not desired behavior long term. Is that true? If so, is there an issue tracking the need for support for args that could be linked from here  in a comment to ensure this gets removed when that is fixed?

shonfeder (commented):
Looking good and useful :D I learned a lot during review.

As I am learning, please be sure to LMK if any of my suggestions seem to be based on a misunderstanding.

bin/promotion.ml
gridbugs: This is following suit from a bunch of other commands that also print a warning when a command is implemented with RPC due to the presence of an existing dune instance. Handling this properly would require a big investment to either make the rpc server more general or to prevent the watch server from holding the global lock the entire time it runs. Both would be large architectural changes. There's no issue as far as I know.

src/dune_rpc_impl/server.ml
```diff
@@ -390,10 +390,22 @@ let handler (t : _ t Fdecl.t) handle : 'build_arg Dune_rpc_server.Handler.t =
     Handler.implement_request rpc Procedures.Public.diagnostics f
   in
   let () =
+    let files_to_promote : _ -> Promote.Diff_promotion.files_to_promote = function
```
gridbugs: My preference would be to change the rpc argument to some type `All | Files of Path.Source.t list`. Even though I doubt anyone will intend to promote the empty list of files in practice, the explicit distinction will help us/users avoid surprises.

bin/promotion.ml
shonfeder: Thanks for the context! Out of scope for this PR then, surely.

src/dune_rpc_impl/server.ml
shonfeder: This makes sense to me, as it prevents the need for documentation to explain the edge case.

otherlibs/dune-rpc/private/procedures.ml
ElectreAAS: Indeed it felt weird modifying otherlibs like that. I ended up adding a separate RPC in rpc_impl, but the names conflict so it's called `promote_but_different` for now :/

bin/rpc/rpc_common.ml
ElectreAAS: Nope that was a leftover from when it was a `let+ where`, it's fixed now

bin/rpc/rpc_common.ml
ElectreAAS: Linking Steve's comment on his PR [here](https://github.com/ocaml/dune/pull/11900#discussion_r2168987261). It is code that was just moved from `rpc/build.ml` without change, and a proper limit would be outside the scope of this PR

bin/rpc/rpc_common.ml
ElectreAAS: While I agree that this isn't the clearest, it's mirroring the actual command line option `dune rpc build --wait`, which is used to start a client before the server.

src/dune_rpc_impl/server.ml
ElectreAAS: _Github says your comment is 'outdated', but this function was just moved to `decl.ml`_
While on its face yes `promote []` looks like it should promote nothing, the pendant of that is the CLI: you expect `dune promote` to actually do something, instead of having to type `dune promote file1.output file2.output`.
Maybe a comment would clarify this directly in the code though

src/dune_rpc_impl/server.ml
ElectreAAS: I agree, but I initially didn't want to modify `otherlibs/dune-rpc/private/procedures.ml` with such a heavy change, this was a compromise.
I ended up reverting this change and not touching otherlibs at all, prefering a new RPC request with pretty much exactly the type you suggest

ElectreAAS:
I tried to address most comments with my latest changes.
~~One thing that is missing however, is a test showcasing the new behaviour and guarding against regressions. I tried making one but failed so far :(~~

There is now a test! The change is subtle but visible

Should there be a changes.md entry? The user-facing behaviour changes ever so slightly

src/dune_rpc_impl/server.ml
```diff
@@ -389,6 +389,15 @@ let handler (t : _ t Fdecl.t) handle : 'build_arg Dune_rpc_server.Handler.t =
...
   in
+  let () =
+    let f _ files =
+      Promote.Diff_promotion.promote_files_registered_in_last_run files;
+      Fiber.return Build_outcome_with_diagnostics.Success
+    in
+    Handler.implement_request rpc Decl.promote f
+  in
+  (* The two handlers above and below implement the same thing: 'promote',
+     but in two slightly different ways. Maybe one should be removed/deprecated? *)
```
ElectreAAS: I am directing special attention to this comment and the underlying question: should `Dune_rpc_private.Procedures.Public.promote` be removed/deprecated in favor of the new one? If they are to coexist, what form should that take?

gridbugs:
> Should there be a changes.md entry? The user-facing behaviour changes ever so slightly

Yep there should be a changelog entry. Rather than updating changes.md, add a file in doc/changes (see the existing files there to learn the expected format).

bin/rpc/rpc_common.ml
shonfeder: I see. I'm not sure if that is a good rationale, but I can see the logic to it. Thanks!

src/dune_rpc_impl/server.ml
shonfeder: Yes, I was asking for a documentation comment here :)

src/dune_rpc_impl/server.ml
Alizter: They can both live together. The public one is something we don't want to touch, there are tools such as ocaml-lsp that rely on its existence. However, we can add our own rpc methods that we use internally that don't need to go in the public interface and only live in the implementation.

It's fine that it does something similar, we can think of unifying them as a single method at a later date, if there is a need.

Alizter: Regarding the form it should take. Maybe we want to call it something like `promote_many`.

ElectreAAS: Thanks a lot!

ElectreAAS:
There is now a changelog entry, and I think I addressed all comments
I'll merge once the CI is green, unless someone stops me ;)

src/dune_rpc_impl/client.ml
```diff
@@ -3,7 +3,7 @@ open Import
 let client ?handler connection init ~f =
   Client.client
     ?handler
-    ~private_menu:[ Request Decl.build; Request Decl.status ]
+    ~private_menu:[ Request Decl.build; Request Decl.status; Request Decl.promote ]
```
rgrinberg: Why did you `Decl.promote` to the private menu? More to the point, if there's already a public promote RPC, what's stopping this RPC from being public as well? Seems like it's a strict improvement.

rgrinberg (changes requested):

bin/rpc/rpc_common.mli
```diff
@@ -17,3 +20,37 @@ val client_term : Common.Builder.t -> (unit -> 'a Fiber.t) -> 'a
...
+      -> ( Dune_rpc_impl.Decl.Build_outcome_with_diagnostics.t
+           , Dune_rpc_private.Response.Error.t )
+           result
+           Fiber.t)
+  -> 'a
+  -> unit
+  -> unit Fiber.t
+
+(**  Schedule a fiber to run via RPC, wrapping any errors. *)
+val actually_run_via_rpc
```
rgrinberg: I don't think the prefix `actually_` is making your name more readable.

src/dune_rpc_impl/client.ml
ElectreAAS: I saw Steve did it in his PR so I thought it was mandatory. Also the comment in `otherlibs/dune-rpc/private/dune_rpc_private.ml` mentions "it is a *huge* footgun that you have to remember to declare a request here, or via [private_menu]".
As it is, just removing the declaration here makes it so it doesn't work at all and gives a weird error

rgrinberg: Right. So I think you should declare this request where all the public requests are declared. No sense in keeping this private without a specific reason. This will also fix the CI issue.

ElectreAAS: What "public requests"? The other two requests in `dune_rpc_impl.decl.ml` are build and status, and both are in this private menu (but visible from `dune rpc --help`, so I don't know what private means here).

The other promote and ping and a lot of other rpc requests (diagnostics & co) are declared in otherlibs, but otherlibs can't depend on dune_rpc_impl.
The function which looks like it would help is `dune_rpc_server.declare_request`, but 1. this function is never used anywhere and 2. using it in `dune_rpc_impl.server.ml` crashes dune completely with `"attempted to implement and declare method"`

ElectreAAS: The CI issue seems unrelated to this specific thing, but due to the fact that I deleted `Handler.implement_request rpc Dune_rpc_private.Procedures.Public.promote f` in the server. Un-deleting it fixes the test

bin/rpc/rpc_common.mli
ElectreAAS: Good point

src/dune_rpc_impl/client.ml
rgrinberg: Public requests are those which are declared in https://github.com/ocaml/dune/blob/225902ceb26dfd1b261b1c1a7ae84ffec7f1023d/otherlibs/dune-rpc/private/dune_rpc_private.ml#L644. 

Private requests are not declared as such and must be listed explicitly in the private menu.

In this case, public refers to being usable through our public RPC library. While private means usable only when dune itself is the RPC client and knows the private menu with all the secret methods.

ElectreAAS: I'm afraid I don't understand where you want me to go here.

There are 16 requests at the moment, +1 in this PR.
- 2 of them are `build` & `status`, which are in the private menu in `dune_rpc_impl.client`.
- The 14 others have their signature in `dune-rpc/private/procedures.ml`, and they are declared, as you point out, in `dune_rpc_private.ml`.

I can't put a declaration with the 14 as is, because my request signature is in `src/dune_rpc_impl/decl.ml` and otherlibs can't depend on that (otherwise that would be a circular dependency).

I wanted to say moving everything into dune_rpc_private was impossible, but it's not, it's just very very verbose: see my PR on this PR at https://github.com/ElectreAAS/dune/pull/1

Is there a third option I'm missing?

rgrinberg: There's no 3rd option. 

If there's an objection to putting it in dune_rpc_private, it has to be more than just verbosity. The point of adding RPC methods is to grant users more capabilities to interact with dune. Do you think the required verbosity means that we shouldn't do this? Shall we just limit RPC's to interaction between dune processes then? That seems limiting.

I think we should avoid using the private menu as much as possible and consider moving everything to public rpc's unless there's a concrete issue with the rpc itself.

Edit: expanded my answer to explain why private rpc's are not the first choice.

ElectreAAS: My reticence came from a junior dev heuristic: "if I have to modify 33 files just to do this one thing I'm probably doing something wrong. on top of that it's going to be a nightmare to review"
If that's really the path we should take then let's go (I updated https://github.com/ElectreAAS/dune/pull/1 to really move things and not have type aliases everywhere)

src/dune_engine/build_system_error.mli
```diff
@@ -26,7 +26,9 @@ val dir : t -> Path.t option
     metadata that are extracted into [`Diagnostic] *)
 val description
   :  t
-  -> [ `Exn of Exn_with_backtrace.t | `Diagnostic of Compound_user_error.t ]
+  -> [ `Exn of Exn_with_backtrace.t
+     | `Diagnostic of Dune_rpc_private.Compound_user_error.t
```
rgrinberg: Instead of this change, could you instead introduce an alias to dune_engine/import.ml:

```
module Compound_user_error = Dune_rpc_private.Compound_user_error
```

These modules move around between libraries a lot, so it's annoying to qualify everything from scratch.

src/dune_engine/process.ml
```diff
@@ -539,9 +539,15 @@ end = struct
           let annots =
             User_message.Annots.set annots User_message.Annots.has_embedded_location ()
           in
-          match Compound_user_error.parse_output ~dir output.without_color with
+          match
+            Dune_rpc_private.Compound_user_error.parse_output ~dir output.without_color
```
rgrinberg: Same here

src/dune_engine/build_system_error.ml
```diff
@@ -8,7 +8,7 @@ type t =
       }
   | Diagnostic of
       { id : Id.t
-      ; diagnostic : Compound_user_error.t
+      ; diagnostic : Dune_rpc_private.Compound_user_error.t
```
rgrinberg: ditto

rgrinberg (approved):

src/dune_rpc_impl/client.ml
rgrinberg: I don't like boilerplate either. The solution is to reduce it when possible. I wouldn't want the reduction of boilerplate to be a goal in of itself.

PR #12016, opened 2025-07-11
chore: update nix flake

Alizter:


rgrinberg:
@anmonteiro it seems like the version of ocamlformat pulled is broken?

anmonteiro:
i just fixed this -- i pulled ocamlformat from its master branch to support 5.4 but didn't keep the release for older versions of OCaml.

PR #12017, opened 2025-07-11
ci: run tests with nix

Alizter:
In this PR we introduce CI jobs for running the tests with nix. We also introduce a dependency order to the rest of the CI.

Many of the platform specific tests won't run unless the main nix tests pass. The nix tests act as a point of early failure. It is a fast job since much of it is cached.

Similarly for the build jobs, many rely on the nix build. If the nix build doesn't work, then there is no point attempting to build the other jobs anyway.

This also stops the useless running of CI runners and availability issues when there is a lot of activity in the repository.

I've also noted that at some point we have stopped building dune configurator separately. I believe that this step was intended for OCaml jobs such as 4.08 to still be able to use it. It might be worth reintroducing that, but otherwise our usual configurator tests are run in more recent versions of OCaml.

### DIFF
--- a/otherlibs/dune-site/test/dune
+++ b/otherlibs/dune-site/test/dune
@@ -4,10 +4,13 @@
   (package dune)
   (package dune-site)))
 
+; The test being broken on CI, we deactivate it when
+; we detect that the CI environment variable is not
+; set. Most CI systems set it.
+
 (cram
  (applies_to run_2_9 run)
  (enabled_if
   (and
-   (<> %{os_distribution} nixos)
    (not %{env:CI=false})
-   (not %{env:NIX_CI=false}))))
+   (not %{env:INSIDE_NIX=false}))))
--- a/otherlibs/dune-site/test/dune
+++ b/otherlibs/dune-site/test/dune
@@ -4,13 +4,10 @@
   (package dune)
   (package dune-site)))
 
-; The test being broken on CI, we deactivate it when
-; we detect that the CI environment variable is not
-; set. Most CI systems set it.
-
 (cram
  (applies_to run_2_9 run)
  (enabled_if
   (and
    (<> %{os_distribution} nixos)
-   (not %{env:CI=false}))))
+   (not %{env:CI=false})
+   (not %{env:NIX_CI=false}))))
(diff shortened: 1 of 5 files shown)


otherlibs/dune-site/test/dune
```diff
@@ -4,13 +4,10 @@
...
-; set. Most CI systems set it.
-
 (cram
  (applies_to run_2_9 run)
  (enabled_if
   (and
    (<> %{os_distribution} nixos)
-   (not %{env:CI=false}))))
+   (not %{env:CI=false})
+   (not %{env:NIX_CI=false}))))
```
rgrinberg: Why do you introduce this new variable? Why don't you just set CI in nix?

otherlibs/dune-site/test/dune
```diff
@@ -4,13 +4,10 @@
   (package dune)
   (package dune-site)))
 
-; The test being broken on CI, we deactivate it when
```
rgrinberg: Why is this comment deleted?

otherlibs/dune-site/test/dune
Alizter: There are other tests that I am disabling only for nix, but these are still being run elsewhere in the CI.

rgrinberg: We have four different cases:

1. Nix
2. CI
3. Nix + CI
4. Not nix and not CI

I claim that two variables should be enough to express all of these possibilities. If you disagree, could you give me an example?

Alizter: We have no variable to detect nix. We have a variable to detect nixos, but the CI runners are ubuntu + nix not nixos so the distribution variable doesn't cover it.

rgrinberg: I see. Still this solution strikes me as incomplete. What about users on ubuntu + nix running the tests locally?

Do we have another variable that could be helpful? `os_family` perhaps? If not, we could consider introducing something ourselves here. In the worst case, we could always use some existing variable that nix sets. There should be plenty to choose from.

Alizter: There isn't a reliable way to check if we are inside a "nix shell". There is "IN_NIX_SHELL" which isn't set reliably between nix commands. Therefore, I've instead introduced a "INSIDE_NIX" variable that is set in each of the dev shell derivations. That way we only need a single variable for checking for nix. We aren't concerned whether we are on nixos, since that isn't the part that affects many of the tests.

rgrinberg: Okay thanks. This is good enough for now.

Issue #10393, opened 2024-04-05
Improve error message when git is not found


emillon:
There are several cases where a build just aborts with `git: command not found`. This is fairly unhelpful because:
- it does not explain what causes `git` to be called (ex: `dune subst` or `dune-build-info` with no version in `dune-project`)
- it does not suggest a solution (like adding a version to dune-project; though that's not often the right fix)
This happens in certain CI systems, or on nix.

Alizter:
https://github.com/ocaml/dune/pull/12024 is a good start. A further improvement would be to turn `Dune_vcs.Vcs.git` and co into a result type so that callers can give better error messages.

PR #12024, opened 2025-07-14
Improve error message when git is not found

ipenas-cl:
Previously when git was not available in PATH, users would see a generic 'Program git not found in the tree or in PATH' error. This change adds a helpful hint explaining why git is needed and how to resolve the issue.

The improved error now explains that git is required for:
- Version information in 'dune subst'
- Build info and artifact substitution
- Package management operations

Addresses #10393

### DIFF
--- a/src/dune_vcs/vcs.ml
+++ b/src/dune_vcs/vcs.ml
@@ -55,7 +55,16 @@ let git, hg =
     lazy
       (match Bin.which ~path:(Env_path.path Env.initial) prog with
        | Some x -> x
-       | None -> Utils.program_not_found prog ~loc:None)
+       | None ->
+         let hint =
+           match prog with
+           | "git" ->
+             Some
+               "Git is required for version information in 'dune subst', build info, and \
+                package management. Install git or add it to your PATH."
+           | _ -> None
+         in
+         Utils.program_not_found prog ~loc:None ?hint)
   in
   get "git", get "hg"
 ;;


ipenas-cl:
Hi maintainers,
I've updated the PR with the following fixes:
✅ Added DCO signatures to all commits
✅ Fixed formatting issues in workspace.ml
The CI is now passing except for the benchmark test, which is failing due to an infrastructure issue unrelated to this PR:
E: The repository 'http://deb.debian.org/debian buster Release' no longer has a Release file.
This appears to be because Debian Buster has reached End of Life and the Docker image used for benchmarking needs to be updated. The failure occurs when trying to install system dependencies, not from any code changes in this PR.
Could you please review the changes when you have a chance? The actual code improvements for better git error messages are ready to merge.
Thanks!

src/source/workspace.ml
```diff
@@ -575,11 +577,14 @@ module Context = struct
...
         let common = common ~profile_default ~instrument_with_default in
         let default =
-          (* TODO proper error handling with locs *)
           let name = Context_name.to_string common.name ^ Common.fdo_suffix common in
           Context_name.parse_string_exn (Loc.none, name)
         in
-        let name = Option.value ~default name in
+        let name = 
+          match name with
+          | Some (loc, name) -> name
```
Alizter: You appear to be immediately discarding the location you have here. Is this what you want to do?

src/dune_vcs/vcs.ml
```diff
@@ -55,7 +55,16 @@ let git, hg =
...
       (match Bin.which ~path:(Env_path.path Env.initial) prog with
        | Some x -> x
-       | None -> Utils.program_not_found prog ~loc:None)
+       | None ->
+         let hint =
+           match prog with
+           | "git" ->
+             Some
+               "Git is required for version information in 'dune subst', build info, \
+                and package management. Install git or add it to your PATH."
```
Alizter: Current ocamlformat wraps this slightly differently:

```suggestion
               "Git is required for version information in 'dune subst', build info, and \
                package management. Install git or add it to your PATH."

```

src/source/workspace.ml
Alizter: This is the reason for the build errors in CI, due to the unused `loc`.

I am not quite sure what you are intending to do with this extra location?

ipenas-cl: You're right! I see the issue now. The `located` captures both location and value, but I'm only using the name and discarding the location.

  I think the better approach would be to use the location when parsing fails. Let me update this to actually use the location for better error messages. Would something like this make more
  sense?

  ```ocaml
  let name =
    match name with
    | Some (loc, name) ->
      (try Context_name.parse_string_exn (loc, name)
       with exn -> User_error.raise ~loc [...])
    | None -> default

  Or should I remove the located wrapper entirely if we're not using the location?

ipenas-cl:
@Alizter Thanks for the review! I've addressed both comments:

  1. Fixed the unused `loc` warning by prefixing with underscore. Added a TODO comment to potentially use it for better error messages in the future.
  2. Adjusted the line break in the error message as suggested.

  The CI should be green now (except for the unrelated benchmark failure due to Debian Buster EOL).

src/source/workspace.ml
```diff
@@ -564,7 +564,9 @@ module Context = struct
     let decode =
       let+ common = Common.decode
       and+ name =
-        field_o "name" (Dune_lang.Syntax.since syntax (1, 10) >>> Context_name.decode)
+        field_o
+          "name"
+          (located (Dune_lang.Syntax.since syntax (1, 10) >>> Context_name.decode))
```
Leonidas-from-XIV: I do not understand this change at all. If you introduce a location here and then drop it in the bottom you could just... not do that?

ipenas-cl: @Alizter @Leonidas-from-XIV

  You're absolutely right - capturing the location and then immediately discarding it doesn't make sense at all. I was trying to improve error messages but ended up making the code more confusing without
  actually implementing the improvement.

  I've removed the unnecessary location capture entirely. The code now does exactly what it did before, just focusing on the main goal of this PR which is improving the error message when git is not
  found.

  Thanks for catching this! The latest commit simplifies `workspace.ml` back to its original approach.

Alizter:
@ipenas-cl This is looking good now. Could you squash and rebase on top of `main` and force push here?

ipenas-cl:
> @ipenas-cl This is looking good now. Could you squash and rebase on top of `main` and force push here?

  @Alizter Done! Squashed all commits and rebased on top of main. The PR now contains a single clean commit.

Leonidas-from-XIV:
Looks good. Could you also add a cram test to show the error message getting triggered? It might be a bit tricky as you need to clean up `PATH` to make sure `git` is not in the `PATH` but you can probably create a directory and symlink everything from `PATH` into there except for `git`.

Alizter:
@Leonidas-from-XIV I don't think its necessary to add a cram test here. Setting up `PATH` scrubbing in the cram test is already going to be quite involved. As it stands this PR improves the hint on an existing error, so I think its fine to just merge.

PR #12025, opened 2025-07-14
Add a configuration option to enable package management

Leonidas-from-XIV:
As discussed in the [dune-dev meeting on the 9th of July](https://github.com/ocaml/dune/wiki/dev-meeting-2025-07-09#signal-to-enable-dune-package-management-leonidas-from-xiv) we plan to introduce an option to enable package management instead of detecting it from the presence of lock files.

This PR adds the configuration options. Lock files are currently still used if they are in the source tree (hence the code retains the check), but this will change when autolocking is merged.

### DIFF
--- a/bin/pkg/pkg_enabled.ml
+++ b/bin/pkg/pkg_enabled.ml
@@ -17,7 +17,10 @@ let term =
       List.exists lock_dir_paths ~f:(fun lock_dir_path ->
         Path.exists (Path.source lock_dir_path))
     in
-    if any_lockdir_exists then () else exit 1)
+    let enabled = any_lockdir_exists || workspace.config.pkg_enabled in
+    match enabled with
+    | true -> ()
+    | false -> exit 1)
 ;;
 
 let info =
--- a/bin/pkg/pkg_enabled.ml
+++ b/bin/pkg/pkg_enabled.ml
@@ -17,6 +17,8 @@ let term =
       List.exists lock_dir_paths ~f:(fun lock_dir_path ->
         Path.exists (Path.source lock_dir_path))
     in
+    (* CR-Leonidas-from-XIV: change this logic when we stop detecting lock
+       directories in the source tree *)
     let enabled = any_lockdir_exists || workspace.config.pkg_enabled in
     match enabled with
     | true -> ()
--- a/src/dune_config_file/dune_config_file.ml
+++ b/src/dune_config_file/dune_config_file.ml
@@ -53,6 +53,14 @@ module Dune_config = struct
     ;;
   end
 
+  module Pkg_enabled = struct
+    type t = bool
+
+    let decode = enum' [ "enabled", return true; "disabled", return false ]
+    let equal = Bool.equal
+    let to_dyn = Dyn.bool
+  end
+
   module Terminal_persistence = struct
     type t =
       | Preserve
@@ -236,6 +244,7 @@ module Dune_config = struct
       ; action_stdout_on_success : Action_output_on_success.t field
       ; action_stderr_on_success : Action_output_on_success.t field
       ; project_defaults : Project_defaults.t field
+      ; pkg_enabled : Pkg_enabled.t field
       ; experimental : (string * (Loc.t * string)) list field
       }
   end
@@ -260,6 +269,7 @@ module Dune_config = struct
           ; action_stdout_on_success
           ; action_stderr_on_success
           ; project_defaults
+          ; pkg_enabled
           ; experimental
           }
       =
@@ -287,6 +297,7 @@ module Dune_config = struct
               (Tuple.T2.equal String.equal (Tuple.T2.equal Loc.equal String.equal)))
            t.experimental
            experimental
+      && field Pkg_enabled.equal t.pkg_enabled pkg_enabled
     ;;
   end
 
@@ -314,6 +325,7 @@ module Dune_config = struct
       ; action_stderr_on_success =
           field a.action_stderr_on_success b.action_stderr_on_success
       ; project_defaults = field a.project_defaults b.project_defaults
+      ; pkg_enabled = field a.pkg_enabled b.pkg_enabled
       ; experimental = field a.experimental b.experimental
       }
     ;;
@@ -338,6 +350,7 @@ module Dune_config = struct
           ; action_stdout_on_success
           ; action_stderr_on_success
           ; project_defaults
+          ; pkg_enabled
           ; experimental
           }
       =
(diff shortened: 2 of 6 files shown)


bin/pkg/pkg_enabled.ml
```diff
@@ -17,7 +17,10 @@ let term =
       List.exists lock_dir_paths ~f:(fun lock_dir_path ->
         Path.exists (Path.source lock_dir_path))
     in
-    if any_lockdir_exists then () else exit 1)
+    let enabled = any_lockdir_exists || workspace.config.pkg_enabled in
```
shonfeder, line 22: If I wrote

```
(pkg disabled)
```

then I would be quite surprised if the presence of a lockdir overruled my explicit directive. So I suggest the semantics here be:

```suggestion
    let enabled = workspace.config.pkg_enabled in
```

src/dune_config_file/dune_config_file.ml
```diff
@@ -53,6 +53,14 @@ module Dune_config = struct
     ;;
   end
 
+  module Pkg_enabled = struct
+    type t = bool
+
+    let decode = enum' [ "enabled", return true; "disabled", return false ]
```
shonfeder, line 59: Anticipating that we may want to add more data to a `pkg` stanza in the future, should this maybe be in a field like `(pkg (enabled true))`?

As we discussed, we could imagine eventually wanting to move the other stanzas that only make sense in the context of dune package managemen't under a `pkg` stanza, or putting them in a separate `dune-pkg` file. Introducing this stanza in a way that will make that easier could be wise then perhaps?

test/blackbox-tests/test-cases/pkg/pkg-enabled.t
```diff
@@ -39,3 +39,32 @@ When the default lockdir is present pkg is enabled:
...
+  > EOF
+  $ dune pkg enabled || echo "Package management disabled"
+  Package management disabled
+
+  $ cat > config <<EOF
+  > (lang dune 3.20)
+  > (pkg enabled)
+  > EOF
+  $ dune pkg enabled --config-file=config && echo "Yes, it is enabled"
+  Yes, it is enabled
```
shonfeder, line 76: Discussed offline, but a test to check the overriding behavior between user config and workspace config seems like a good idea.

bin/pkg/pkg_enabled.ml
Leonidas-from-XIV: I agree in principle. The thing is, that for now that is the semantics until autolocking is implemented. We can't at the moment ignore all lock dirs because then there's no way to use package management in the meantime.

src/dune_config_file/dune_config_file.ml
Leonidas-from-XIV: Personally I feel like we shouldn't overgeneralize this out of the box. The cache settings are multiple separate stanzas and in any case we can make `(pkg enabled)` parse as `(pkg (enabled true) (other options))` in the future as suggested in #11716.

I like the short-hand forms for when the thing you want to do most of the time is just a very specific thing - that way users don't have to pay the price in syntax for potential future extensions (that might or might not happen).

test/blackbox-tests/test-cases/pkg/pkg-enabled.t
Leonidas-from-XIV: Done in [b732b47](https://github.com/ocaml/dune/pull/12025/commits/b732b478c9826b9436ba4c0a9a2e0452a46663e3)!

bin/pkg/pkg_enabled.ml
shonfeder: Ah yes. Makes sense. Is there a step that should be taken to ensure that this is updated when the time is right? A `TODO` here or a note in a relevant issue?

src/dune_config_file/dune_config_file.ml
shonfeder: I can appreciate your perspective. IMO, it is worse complexity to have several different ways of writing a configuration than to have one uniform way that is slightly more verbose, and it is better to introduces changes in a way that allows for easy backward compatible changes for stuff we are very likely to do, than to have to rely on a lot of deprecation, versioning, and complications.

But I take your point and I'm happy to disagree and commit to the way you think is best here.

shonfeder (approved):

src/dune_config_file/dune_config_file.ml
Leonidas-from-XIV: > it is better to introduces changes in a way that allows for easy backward compatible changes for stuff we are very likely to do

But are we? We've gone some two years without this, `pins`, `depopts` etc are configured elsewhere anyway, so I am not sure there's any extensions coming up soon.

Also note that `enabled` and `disabled` are not `true`/`false`. The dune codebase has a lot of cases where the option set is not binary. e.g. `only-dev-tools` could be a possible option as well without having to make the default case a record. I am trying to adhere to YAGNI but of course it's debatable where to cut the line.

But I'd say it is somewhat bikeshedding. I can implement the other way as well if people feel strongly.

@gridbugs What's your take on this?

bin/pkg/pkg_enabled.ml
Leonidas-from-XIV: Good point, I've added a comment.

src/dune_config_file/dune_config_file.ml
gridbugs: My preference in this case is to have as few configuration options as possible, ie. go with `(pkg enabled)`. My reasoning is that for global configurations that affects dune's behaviour, the more different ways a user can have it be configured, the more likely it is that advice or examples found online won't be applicable to them. It's more likely to lead to confusion and miscommunications if two different users talking on a forum can have different global configurations. Since the global configuration of dune isn't checked into version control it also creates the potential for contributors on a project to have different configurations in ways that complicates working together.

In this light I consider making it hard for the dune developers to add additional pkg config options as a positive thing, and that happens to align with simpler syntax.

test/blackbox-tests/test-cases/pkg/pkg-enabled.t
```diff
@@ -39,3 +39,54 @@ When the default lockdir is present pkg is enabled:
...
+If we remove the setting from the workspace it should go back to the default
+(disabled)
+
+  $ cat > dune-workspace <<EOF
+  > (lang dune 3.20)
+  > EOF
+  $ dune pkg enabled || echo "Package management disabled"
+  Package management disabled
+
+Enable the package management globally in the users config.
```
gridbugs:
```suggestion
Enable the package management globally in the user's config.
```

gridbugs (approved):

PR #12026, opened 2025-07-14
Update benchmark Docker file to supported Debian version

Leonidas-from-XIV:
The benchmark builds are failing because Debian 10 is not on the release mirrors anymore, maybe updating to Debian 12 will solve this.


Alizter (approved):
Looks like it worked.

PR #12027, opened 2025-07-15
pkg: utop dev tool

gridbugs:
This adds utop as a dev tool. Unlike most dev tools there is no "dune tools exec utop" command, but rather the existing "dune utop" command has been modified such that if a lockdir exists for the utop dev tool, then dune will install the utop library (downloading it if necessary) within the _build directory and use that library to create the custom utop toplevel for the project.

It's possible to install utop as a dev tool without running it by running the "dune tools install utop" command.

### DIFF
--- a/bin/ocaml/utop.ml
+++ b/bin/ocaml/utop.ml
@@ -12,6 +12,12 @@ let man =
 
 let info = Cmd.info "utop" ~doc ~man
 
+let lock_utop_if_dev_tool_enabled () =
+  match Lazy.force Lock_dev_tool.is_enabled with
+  | false -> Memo.return ()
+  | true -> Lock_dev_tool.lock_dev_tool Utop
+;;
+
 let term =
   let+ builder = Common.Builder.term
   and+ dir = Arg.(value & pos 0 string "" & Arg.info [] ~docv:"DIR")
@@ -21,7 +27,7 @@ let term =
   let dir = Common.prefix_target common dir in
   if not (Path.is_directory (Path.of_string dir))
   then User_error.raise [ Pp.textf "cannot find directory: %s" (String.maybe_quoted dir) ];
-  let env, lib_config, utop_path, requires =
+  let env, utop_path =
     Scheduler.go_with_rpc_server ~common ~config (fun () ->
       let open Fiber.O in
       let* setup = Import.Main.setup () in
@@ -29,17 +35,33 @@ let term =
         let open Memo.O in
         let* setup = setup in
         let context = Import.Main.find_context_exn setup ~name:ctx_name in
-        let utop_target =
-          let utop_target = Filename.concat dir Utop.utop_exe in
-          Path.build (Path.Build.relative (Context.build_dir context) utop_target)
+        let utop_target_path filename =
+          Path.build
+            (Path.Build.relative
+               (Context.build_dir context)
+               (Filename.concat dir filename))
         in
-        Build_system.file_exists utop_target
+        let utop_exe = utop_target_path Utop.utop_exe in
+        let utop_findlib_conf = utop_target_path Utop.utop_findlib_conf in
+        Build_system.file_exists utop_exe
         >>= function
         | false ->
           User_error.raise
             [ Pp.textf "no library is defined in %s" (String.maybe_quoted dir) ]
         | true ->
-          let* () = Build_system.build_file utop_target in
+          let* () = Build_system.build_file utop_exe in
+          let* () = lock_utop_if_dev_tool_enabled () in
+          let* utop_dev_tool_lock_dir_exists =
+            Memo.Lazy.force Utop.utop_dev_tool_lock_dir_exists
+          in
+          let* () =
+            if utop_dev_tool_lock_dir_exists
+            then
+              (* Generate the custom findlib.conf file needed when utop is run
+                 as a dev tool. *)
+              Build_system.build_file utop_findlib_conf
+            else Memo.return ()
+          in
           let sctx = Import.Main.find_scontext_exn setup ~name:ctx_name in
           let* requires =
             let dir = Path.Build.relative (Context.build_dir context) dir in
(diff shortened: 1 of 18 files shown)


bin/ocaml/utop.ml
```diff
@@ -50,7 +78,12 @@ let term =
...
-          env, lib_config, Path.to_string utop_target, requires))
+          ( env
+          , lib_config
+          , Path.to_string utop_target
+          , requires
+          , utop_findlib_conf
+          , utop_dev_tool_lock_dir_exists )))
   in
   Hooks.End_of_build.run ();
   let env =
```
shonfeder: Would it be possible/preferable to form the final `env` inside the `Schedular.go` run? I ask because I see the following variables

- `utop_dev_tool_lock_dir_exists`
- `utop_findlib_conf`
- `utop_findlib_conf`
- `requires`

appear to all get passed out in a tuple, only to be used to derive the `env`, which is also passed out in the same tuple. Is the issue that the `env` must be only be updated after the Hooks are run on line 88?

bin/ocaml/utop.ml
```diff
@@ -59,6 +92,18 @@ let term =
          ~f:(fun dir env -> Env_path.cons ~var:Ocaml.Env.caml_ld_library_path env ~dir)
          ~init:env
   in
+  let env =
+    if utop_dev_tool_lock_dir_exists
+    then
+      (* If there's a utop lockdir then dune will have build utop as a dev
```
shonfeder:
```suggestion
      (* If there's a utop lockdir then dune will have built utop as a dev
```

bin/ocaml/utop.ml
```diff
@@ -59,6 +92,18 @@ let term =
...
+         tool. In order for it to run correctly dune needed to generate a
+         custom findlib.conf that contains the locations of all of utop's
+         dependencies within the project's _build directory. Setting this
+         environment variable causes the custom findlib.conf file to be used
+         instead of the default findlib.conf. *)
+      Env.add env ~var:"OCAMLFIND_CONF" ~value:(Path.to_string utop_findlib_conf)
```
shonfeder: If it is necessary to pass out `utop_findlib_conf` in the tuple, maybe it should at least do the string conversion before returning it? This is how we are handling `utop_target` for instance.

gridbugs:
cc @pitag-ha

bin/ocaml/utop.ml
```diff
@@ -21,25 +27,47 @@ let term =
...
+        in
         let context = Import.Main.find_context_exn setup ~name:ctx_name in
         let utop_target =
           let utop_target = Filename.concat dir Utop.utop_exe in
           Path.build (Path.Build.relative (Context.build_dir context) utop_target)
         in
+        let utop_findlib_conf =
+          let utop_findlib_conf = Filename.concat dir Utop.utop_findlib_conf in
+          Path.build (Path.Build.relative (Context.build_dir context) utop_findlib_conf)
+        in
```
shonfeder: Looks like this repeated path creation logic could be factored out to improve readability?

bin/ocaml/utop.ml
```diff
@@ -21,25 +27,47 @@ let term =
...
     Scheduler.go_with_rpc_server ~common ~config (fun () ->
       let open Fiber.O in
       let* setup = Import.Main.setup () in
       build_exn (fun () ->
         let open Memo.O in
         let* setup = setup in
+        let* () = lock_utop_if_dev_tool_enabled () in
+        let* utop_dev_tool_lock_dir_exists =
+          Memo.Lazy.force Utop.utop_dev_tool_lock_dir_exists
+        in
```
shonfeder: Would it make sense to put these down nearer to where they are used? (E.g. at line 64?)

bin/tools/tools.ml
```diff
@@ -20,7 +22,9 @@ module Which = struct
   let info = Cmd.info ~doc "which"
 
   let group =
-    Cmd.group info (List.map [ Ocamlformat; Ocamllsp ] ~f:Tools_common.which_command)
+    Cmd.group
+      info
+      (List.map [ Ocamlformat; Ocamllsp; Utop ] ~f:Tools_common.which_command)
```
shonfeder: I assume these will always get updated together. If so perhaps the sub-commands should be defined in a single list?

shonfeder: Ah, also, for my edification, why isn't this `Dev_tool.all`?

src/dune_rules/lib_flags.ml
```diff
@@ -218,14 +218,24 @@ module L = struct
...
 
-  let toplevel_ld_paths ts =
+  let dll_dir_paths libs =
+    List.fold_left libs ~init:Path.Set.empty ~f:(fun acc lib ->
+      List.fold_left
+        (Lib_info.foreign_dll_files (Lib.info lib))
+        ~init:acc
+        ~f:(fun acc dll_file_path ->
+          let dll_dir_path = Path.parent_exn dll_file_path in
+          Path.Set.add acc dll_dir_path))
```
shonfeder: IMO, naming the accumulators here would help with readability (and I think are indicated by the [style guide](https://dune.readthedocs.io/en/latest/hacking.html#general-guidelines)?). E.g., assuming I am reading this right, something like

```suggestion
    List.fold_left libs ~init:Path.Set.empty ~f:(fun dll_file_paths lib ->
      List.fold_left
        (Lib_info.foreign_dll_files (Lib.info lib))
        ~init:dll_file_paths
        ~f:(fun dll_dir_paths dll_file_path ->
          let dll_dir_path = Path.parent_exn dll_file_path in
          Path.Set.add dll_dir_paths dll_dir_path))
```

src/dune_rules/pkg_rules.ml
```diff
@@ -351,7 +351,22 @@ module Pkg = struct
         ]
   ;;
 
-  let deps_closure t = top_closure t.depends
+  let deps_closure t =
```
shonfeder: Would it be worth a comment here to explain what this is for? Or would more familiar readers be likely to know just from the name?

src/dune_rules/pkg_rules.ml
```diff
@@ -351,7 +351,22 @@ module Pkg = struct
...
+    let closure = top_closure t.depends in
+    List.map closure ~f:(fun pkg -> pkg.info.name, pkg)
+    |> Dune_pkg.Package_name.Map.of_list_multi
+    |> Dune_pkg.Package_name.Map.iter ~f:(function
+      | _ :: _ :: _ as multiple ->
```
shonfeder: Why does a list with >= 3 elements let us know we have a duplicate? Adding a comment to explain that logic would help me (as an admittedly new) reader.

shonfeder (commented):
Awesome to be getting support for this in! Also, additions read quite nicely IMO :D  

I have a few suggestions, but they may be more like questions if I happen to be misunderstanding some prior context.

src/dune_rules/pkg_rules.ml
gridbugs: It's just a list of 2 elements because the final `_` may match with the empty list.

shonfeder: Ah yeah! My bad. 

So, IIUC, what we are checking here is that each list is a singleton, and if there is more than one element, it will have duplicates?

If so, I think this would be clearer as something like

```suggestion
      | [_] - () 
      | multiple -> (* more than one of the dependency found *)
```

unless I am missing something?

bin/ocaml/utop.ml
gridbugs: I refactored this so `utop_findlib_conf` isn't included in the output tuple.

bin/tools/tools.ml
gridbugs: Ah good idea. This file has had many refactors over the past few months. I'll update it to use `Dev_tool.all`.

src/dune_rules/pkg_rules.ml
gridbugs: Ah oops this was some debugging code that I forgot to delete. I'll just remove this from the diff. Good spot.

src/dune_rules/pkg_rules.ml
gridbugs: I ended up deleting this code as it was leftover debugging code.

shonfeder (approved):
Looks great, IMO!

PR #12028, opened 2025-07-15
test: regression with --force and cram tests

Alizter:
Previously, --force would re-run already run cram tests. However this is no longer the case. 

Here is the test with 3.19 before #11994: 

```
 |side-effect should only contain a single "run":
 |
 |  $ cat side-effect
 |  run
+|  run
 |
 |However, if passing --force, we should still be able to re-run cram tests:
 |
 |  $ dune runtest foo.t --force
 |
 |There should be two "run"s here, however there is only one:
 |  $ cat side-effect
 |  run
+|  run
+|  run
```

### DIFF
--- a/test/blackbox-tests/test-cases/cram/double-run-promote.t
+++ b/test/blackbox-tests/test-cases/cram/double-run-promote.t
@@ -24,3 +24,12 @@ side-effect should only contain a single "run":
 
   $ cat side-effect
   run
+
+However, if passing --force, we should still be able to re-run cram tests:
+
+  $ dune runtest foo.t --force
+
+There should be two "run"s here, however there is only one:
+  $ cat side-effect
+  run
+


rgrinberg (approved):
Thanks for catching this. For now, let's just note this in the changes. Using `--force` is only useful for broken cram tests anyway, and the issue is easily worked around by inserting a dummy edit into the test.

Alizter:
For now, I've been doing `dune clean` and `dune build --cache disabled` when I need to rerun cram tests. Typically its due to missing dependencies that I get a cache hit on a test that no longer runs correctly.

Alizter:
If `--force` is only useful for cram tests and it no longer works there, shouldn't we remove it, or at least throw a warning when people use it?

rgrinberg:
We removed it internally because it was of little use. Not sure what we're going to do about it, but I wouldn't worry about it's fate very much.

PR #12029, opened 2025-07-16
[RPC] Added a test showing the current promotion capabilities

ElectreAAS:
Related to #11958 and other issues linked there.

The current `dune promote` command doesn't check for a running RPC server, and thus access the build directory in a non-synchronised way.
This is racy, but right now in the tests it works.
This PR just adds a test showing what's currently happening so that #12010 can replace that buggy behaviour without regression.

Thanks to @Alizter for help making this :)

### DIFF
--- a/test/blackbox-tests/test-cases/watching/direct-promotion.t
+++ b/test/blackbox-tests/test-cases/watching/direct-promotion.t
@@ -1,0 +1,26 @@
+Demonstrate running "dune promote" concurrently with a passive rpc server
+
+  $ . ./helpers.sh
+
+  $ echo '(lang dune 3.20)' > dune-project
+  $ echo "  $ echo hello" > my_test.t
+
+  $ start_dune
+
+The test expectedly fails (see errors at the end of this file)
+  $ build "(alias my_test)"
+  Failure
+
+Promotion works even with a running RPC server.
+  $ dune promote
+  Promoting _build/default/my_test.t.corrected to my_test.t.
+
+  $ build "(alias my_test)"
+  Success
+
+  $ stop_dune
+  File "my_test.t", line 1, characters 0-0:
+  Error: Files _build/default/my_test.t and _build/default/my_test.t.corrected
+  differ.
+  Had 1 error, waiting for filesystem changes...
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/watching/dune
```diff
@@ -21,5 +21,6 @@
 
 (cram
  (applies_to watching-eager-concurrent-build-command
-  watching-eager-concurrent-exec-command)
```
Alizter: Does this get stuck? Since we are using passive watch mode I wouldn't expect this to be the case.

test/blackbox-tests/test-cases/watching/watching-eager-concurrent-promote-command.t
```diff
@@ -0,0 +1,26 @@
+Demonstrate running "dune promote" concurrently with an eager rpc server
```
Alizter: We are using a passive rpc server here. But the distinction for us is not so important.

test/blackbox-tests/test-cases/watching/watching-eager-concurrent-promote-command.t
```diff
@@ -0,0 +1,26 @@
...
+  $ echo '(lang dune 3.20)' > dune-project
+  $ echo "  $ echo hello" > my_test.t
+
+  $ start_dune
+
+The test expectedly fails (see errors at the end of this file)
+  $ build "(alias my_test)"
+  Failure
+
+Promotion works even with a running RPC server.
```
Alizter: The actual "racy" part would be if we promoted and built at the same time, which is what happens when you build with eagar watch mode. Here we aren't testing how this interacts, but merely capturing that a user can promote when an RPC server is running. This is unlike other commands that take the build directory lock.

Alizter:
Don't worry about the 32-bit build, that's unrelated.

test/blackbox-tests/test-cases/watching/dune
ElectreAAS: Oops you're right
Fixed

Alizter (approved):

PR #12033, opened 2025-07-16
ci: disable 32 bit job

Alizter:
We temporarily disable the 32bit job unitl we track down and fix the issue.


rgrinberg:
@hhugo do you want to try fixing this?

Alizter:
@rgrinberg I have a fix in https://github.com/ocaml/dune/pull/12031 but I need to adapt it so we only do an `apt udpate` for the 32 bit job otherwise it is slow.

PR #6445, opened 2022-11-11
[coq] Remove support for Coq < 8.10 , and coq build languages 0.1 and 0.2

ejgallego:
Since #6409 we automatically infer rules for coq-native, and in general this requires at least Coq 8.10 due to the need of command-line option flags added in that version for native.

The recommended path for most projects is just to upgrade their dune file and Coq version file. Coq 8.10 will likely be the base we support in `(lang coq 1.0)`.

PR #12035, opened 2025-07-16
[rocq] Initial support for the Rocq Prover build language

### LINKED ISSUE #11572: Rocq support in dune
## Desired Behavior

We would like dune's Coq support to be able to call the new `rocq` binary instead of `coqc`.
Currently we need to use compatibility shims (opam package `coq-core`) on top of the `rocq-core` package to compile Rocq projects.
Ideally we want a solution that's compatible with both `coqc` and `rocq` usage.

## Example

Dune coq support just fails if `coqc` is not installed right now.

## Discussion

We are currently discussing the various design decisions regarding Rocq <-> dune integration here, and plan to contribute a solution. This issue hence for letting you know we're working on it and soliciting feedback if you have any:

https://rocq-prover.zulipchat.com/#narrow/channel/240550-Dune-devs-.26-users/topic/Dune.20support.20for.20rocq.20binary/with/508283855

ejgallego:
We introduce `(lang rocq)`, a build mode for the Rocq Prover, previously known as "Coq".

The build mode is an exact copy of the `(lang coq)` mode, renaming the modules, data, and some global declarations. Also, the default implicit lib is now named `Corelib`.

`(lang coq)` should be considered deprecated / frozen, and eventually removed.

There are many cleanups and improvements possible, in particular we have done, in separate commits:

- removed support for older legacy coq modes
- removed support for legacy plugin handling
- removed the need for coq-* shims

All of the tests have been ported and manually verified to work (a time consuming process) As of today, all test pass.

Future work:

- [ ] support new extraction option (-output-directory)
- [ ] improve composition with local rocq as in `(lang coq)`, as of today the rocqworker is not injected as a dependency
- [ ] nix CI + develop
- [x] forward port _RocqProject generation
- [x] forward port rocqdoc headers work
- [x] forward port per_module fixes
- [ ] cosmetic, turn `rocq` into `rocq compile` for `--display=short`
- [ ] [CI] Remove dependency on rocq-stdlib, only extraction tests need that

Closes #11572
Cc #12788 #6445 #9348 #12734 

Joint work with Li-yao Xia.


rlepigre-skylabs-ai:
I just learned that you are working on this @ejgallego, that's great news!

There are several points that are IMO worth discussing before merging this PR:
- Should we integrate the `_CoqProject` generation PR (https://github.com/ocaml/dune/pull/11752)?
- To more easily support the above, should we remove per-module flags? (IMO they are kind of useless.)
- Similarly, should we enforce at most one theory stanza per directory? (I personally always do that, can't imagine this breaking too much stuff.)
- Should we sort out the `.vos` / `.vok` mess?
- Should we integrate https://github.com/ocaml/dune/pull/11131?

Also, I'm happy to play a part in this if you want help.

rlepigre-skylabs-ai:
Another candidate is going back to calling `rocq dep` on all files, instead of per-theory.

ejgallego:
Hey @rlepigre-skylabs-ai !

> I just learned that you are working on this @ejgallego, that's great news!

Indeed! Thanks a lot for your message. We got side-tracked by other stuff, but the idea is to merge this and #11945 

> There are several points that are IMO worth discussing before merging this PR:
> 
> * Should we integrate the `_CoqProject` generation PR ([Generate "_CoqProject" files for Coq theories #11752](https://github.com/ocaml/dune/pull/11752))?

Yes, we will forward port this in a separate PR.

> * To more easily support the above, should we remove per-module flags? (IMO they are kind of useless.)

IMHO we can just say that both can't work together, there are some folks trying to use this feature, so I'd like to keep it.

> * Similarly, should we enforce at most one theory stanza per directory? (I personally always do that, can't imagine this breaking too much stuff.)

This would certainly break a lot of stuff tho, what would we gain? 

> * Should we sort out the `.vos` / `.vok` mess?

Happy to work on this but IMHO we can do in a separate PR.

> * Should we integrate [Add "coqdoc_header" and "coqdoc_footer" fields. #11131](https://github.com/ocaml/dune/pull/11131)?

Yes, we will also forward port this.

> Also, I'm happy to play a part in this if you want help.

Absolutely, @Lysxia and I will send a mail soon to coordiante.

> Another candidate is going back to calling `rocq dep` on all files, instead of per-theory.

This should be easy to do if we want, tho for some users it would mean a large performance regression. But IMHO unrelated to this PR which is to enable Rocq users no to have to depend on the `coq-*` compatibility wrappers.

rlepigre-skylabs-ai:
> This should be easy to do if we want, tho for some users it would mean a large performance regression. But IMHO unrelated to this PR which is to enable Rocq users no to have to depend on the `coq-*` compatibility wrappers.

I think this was originally due to performance issues in `coqdep` that have now kind of been fixed, but I could be wrong. For us, at this point, this would be a performance improvement I believe.

ejgallego:
> I think this was originally due to performance issues in `coqdep` that have now kind of been fixed, but I could be wrong.

Do you have pointer? The original performance problem is that in one rocqdep call per file, rocqdep will still scan the whole user-contrib tree. This has a huge impact in practice, even in fast systems like Linux.

> For us, at this point, this would be a performance improvement I believe.

I'd be happy to make this configurable, the original Coq mode used this system so it is easy just to look at the commit and add back that code with a flag.

Once we have that we could test for numbers, but my original experiments showed a large overhead of repeated rocqdep calls.

rlepigre-skylabs-ai:
> Do you have pointer? The original performance problem is that in one rocqdep call per file, rocqdep will still scan the whole user-contrib tree. This has a huge impact in practice, even in fast systems like Linux.

I can't find anything, but I thought I had seen something in that line.

If that wasn't the case, then perhaps `coqdep` should be fixed to scan it more lazily, or we should try to push for getting rid of `user-contrib` entirely.

SkySkimmer:
coqdep called by dune (ie with -boot) should not be scanning all user-contrib but only the parts which are mapped to a declared dependency

rlepigre-skylabs-ai:
Are we actually certain that duplicating everything is the right option to support `rocq`?  It seems to me like we could keep the same language but allow for an alias, and make the change more gradual? That being said, it looks like a lot of work went into this PR already. I'm just worried that it's gonna be supper hard to get this in.

ejgallego:
> Are we actually certain that duplicating everything is the right option to support `rocq`?

I am.

> It seems to me like we could keep the same language but allow for an alias, and make the change more gradual?

I tried that, but it is actually quite complex, a lot of cases to handle, a lot of different tests, conditional arguments. The breakage is far from trivial.

> That being said, it looks like a lot of work went into this PR already. I'm just worried that it's gonna be supper hard to get this in.

What is your worry?

rlepigre-skylabs-ai:
> What is your worry?

My main worry is asking everyone to move to a new language, for what appears to me to be not a good reason. We'll basically ask people to modify every single of their `dune` / `dune-project` files, which seems annoying.

ejgallego:
> My main worry is asking everyone to move to a new language, for what appears to me to be not a good reason. We'll basically ask people to modify every single of their `dune` / `dune-project` files, which seems annoying.

They'll have to do that anyways due to the several improvements in Rocq mode, we would have introduced a new `(lang coq )` version anyways, so IMHO bumping `(lang coq 0.10)` to `(lang coq 0.11)` is not too far from changing it to `(lang rocq 0.11)`.

Alizter:
@gridbugs Do you know why the CI is failing here? The test that is affected is `portable-lockdirs-platform-dependant-dependencies.t`. AFAICT the changes in this PR are seemingly unrelated.

test/blackbox-tests/test-cases/pkg/portable-lockdirs/portable-lockdirs-platform-dependant-dependencies.t
```diff
@@ -74,13 +74,13 @@ A package that conditionally depends on packages depending on the OS:
 Build the project as if we were on linux and confirm that only the linux-specific dependency is installed:
   $ DUNE_CONFIG__OS=linux DUNE_CONFIG__ARCH=arm64 DUNE_CONFIG__OS_FAMILY=debian DUNE_CONFIG__OS_DISTRIBUTION=ubuntu DUNE_CONFIG__OS_VERSION=24.11 dune build
   $ ls $pkg_root/
-  foo.0.0.1-9b0ae1f26c9a79e6ae8bca96ec389858
-  linux-only.0.0.1-f1f7456a7bf1c70f9203f8caadf79f6d
+  foo.0.0.1-5e48eb7073ada94c09fb13ac3853f1e9
```
Alizter, line 77: The fact that this had to change is worrying.

src/dune_lang/pform.ml
```diff
@@ -349,6 +350,8 @@ module Macro = struct
     | Coq_config, Coq_config -> Eq
     | Coq_config, _ -> Lt
     | _, Coq_config -> Gt
+    | Rocq_config, _ -> Lt
```
rgrinberg, line 353: `Rocq_config, Rocq_config -> Eq`

test/blackbox-tests/test-cases/pkg/portable-lockdirs/portable-lockdirs-platform-dependant-dependencies.t
gridbugs: It's because we use `Digest_feed.generic` to hash actions which contain pforms, and this change introduces a new pform constructor before the `Pkg_self` constructor, which is one of the pforms appearing in the build commands of the packages whose hashes are changing here. `Digest_feed.generic` computes a hash based on the representation of values, and variant constructors have a tag in their representation based on their position within the variant type definition. All the constructors below the new `Rocq_config` constructor have had their tag increased by 1.

To confirm this, apply the following patch which will cause the hashes in this test to revert to their original values:
```patch
diff --git i/src/dune_lang/pform.ml w/src/dune_lang/pform.ml
index e836d64c8..9b4601cc6 100644
--- i/src/dune_lang/pform.ml
+++ w/src/dune_lang/pform.ml
@@ -296,12 +296,12 @@ module Macro = struct
     | Path_no_dep
     | Ocaml_config
     | Coq_config
-    | Rocq_config
     | Env
     | Artifact of Artifact.t
     | Pkg
     | Pkg_self
     | Ppx
+    | Rocq_config
 
   let compare x y =
     match x, y with
diff --git i/src/dune_lang/pform.mli w/src/dune_lang/pform.mli
index ae2eba7c3..4feb23cf2 100644
--- i/src/dune_lang/pform.mli
+++ w/src/dune_lang/pform.mli
@@ -137,12 +137,12 @@ module Macro : sig
     | Path_no_dep
     | Ocaml_config
     | Coq_config
-    | Rocq_config
     | Env
     | Artifact of Artifact.t
     | Pkg
     | Pkg_self
     | Ppx
+    | Rocq_config
 
   val compare : t -> t -> Ordering.t
   val to_dyn : t -> Dyn.t
```

The consequence of this for users is that they'll have to rebuild many of the dependencies of projects upon updating Dune. I don't think that's a big deal since updating Dune is quite rare and I don't think that behaviour is too surprising. If we really wanted to prevent this sort of thing we'd have to stop using `Dune_feed.generic` and take care to manually implement hash functions in a way that isn't sensitive to these sorts of changes. Personally I don't think it's worth it.

src/dune_lang/pform.ml
ejgallego: Thanks, will fix.

PR #12036, opened 2025-07-16
nix: add missing test dependencies

Alizter:
We update the nix flake with some missing dependencies needed to run the tests.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/dune
+++ b/test/blackbox-tests/test-cases/pkg/dune
@@ -26,7 +26,7 @@
  (deps %{bin:strace}))
 
 (cram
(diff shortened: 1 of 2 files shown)


test/blackbox-tests/test-cases/pkg/dune
```diff
@@ -26,7 +26,7 @@
  (deps %{bin:strace}))
 
 (cram
- (deps %{bin:git} %{bin:awk} %{bin:cmp})
```
Leonidas-from-XIV, line 29: Why is `git` being removed here?

Alizter: It's duplicated with a deceleration above. I only noticed because with no `git` we get two errors.

PR #12038, opened 2025-07-16
chore: move *.opam files to opam/

Alizter:
We previously reverted this change due to the windows CI not having an up-to-date version of opam. Let's see if the situation has improved.


Alizter:
Looks like opam is still too buggy:
```
Run opam install opam/ --deps-only
[ERROR] Package conflict!
  * Missing dependency:
No solution found, exiting
    - deps-of-dune-rpc-lwt → deps-of-dune >= 3.20
    no matching version
  * Missing dependency:
    - deps-of-dune-rpc-lwt → lwt >= 5.6.0 → dune-configurator → dune >= 3.20
    no matching version
```

PR #12039, opened 2025-07-17
Add with-dev-setup to package dependency specification.

tmcgilchrist:
Minor update to the documentation to indicate that `:with-dev-setup` is also a valid filter. The :with-test, :with-doc, and :with-dev-setup have all been available since opam 2.2.


Leonidas-from-XIV (approved):

Alizter:
As a side note, we are completely ignoring this for `dune pkg` so maybe it's worth putting a warning here too?

PR #12041, opened 2025-07-17
feat(cram): timeout for cram tests

Alizter:
We add a `(timeout)` field for cram stanzas allowing users to set up a time budget for cram tests to run in. If the budget is exceeded then the cram test will be cancelled.

- [ ] `dune_engine` changes are ok? 
- [x] changelog
- [x] documentation

### DIFF
--- a/src/dune_engine/process.ml
+++ b/src/dune_engine/process.ml
@@ -13,22 +13,44 @@ module Failure_mode = struct
     | Strict : ('a, 'a) t
     | Accept : int Predicate.t -> ('a, ('a, int) result) t
     | Return : ('a, 'a * int) t
+    | Timeout :
+        { timeout_seconds : float option
+        ; failure_mode : ('a, 'b) t
+        }
+        -> ('a, ('b, [ `Timed_out ]) result) t
 
-  let accepted_codes : type a b. (a, b) t -> int -> bool = function
+  let rec accepted_codes : type a b. (a, b) t -> int -> bool = function
     | Strict -> Int.equal 0
     | Accept exit_codes -> fun i -> Predicate.test exit_codes i
     | Return -> fun _ -> true
+    | Timeout { failure_mode; _ } -> accepted_codes failure_mode
   ;;
 
-  let map_result : type a b. (a, b) t -> int -> f:(unit -> a) -> b =
-    fun mode t ~f ->
+  let exit_code_of_result = function
+    | `Finished n -> n
+    | `Timeout -> Code_error.raise "should not return `Timeout" []
+  ;;
+
+  let timeout_seconds : type a b. (a, b) t -> float option = function
+    | Timeout { timeout_seconds; _ } -> timeout_seconds
+    | Strict | Accept _ | Return -> None
+  ;;
+
+  let rec map_result
+    : type a b. (a, b) t -> [ `Timeout | `Finished of int ] -> f:(unit -> a) -> b
+    =
+    fun mode result ~f ->
     match mode with
     | Strict -> f ()
     | Accept _ ->
-      (match t with
+      (match exit_code_of_result result with
        | 0 -> Ok (f ())
        | n -> Error n)
-    | Return -> f (), t
+    | Return -> f (), exit_code_of_result result
+    | Timeout { failure_mode; _ } ->
+      (match result with
+       | `Timeout -> Error `Timed_out
+       | `Finished _ -> Ok (map_result failure_mode result ~f))
   ;;
 end
 
@@ -857,9 +879,9 @@ let report_process_finished
 
 let set_temp_dir_when_running_actions = ref true
 
-let await { response_file; pid; _ } =
+let await ~timeout_seconds { response_file; pid; _ } =
   let+ process_info, termination_reason =
-    Scheduler.wait_for_build_process pid ~is_process_group_leader:true
+    Scheduler.wait_for_build_process ?timeout_seconds pid ~is_process_group_leader:true
   in
   Option.iter response_file ~f:Path.unlink_exn;
   process_info, termination_reason
@@ -1019,7 +1041,7 @@ let run_internal
         cmdline
       | _ -> Pp.nop
     in
-    let t =
+    let (t : t) =
       spawn
         ?dir
         ?env
@@ -1045,7 +1067,9 @@ let run_internal
       in
       Running_jobs.start id t.pid ~description ~started_at:t.started_at
     in
-    let* process_info, termination_reason = await t in
+    let* process_info, termination_reason =
+      await ~timeout_seconds:(Failure_mode.timeout_seconds fail_mode) t
+    in
     let+ () = Running_jobs.stop id in
     let result = Result.make t process_info fail_mode in
     let times =
@@ -1076,6 +1100,7 @@ let run_internal
          we're about to return. *)
       Result.close result;
       raise (Memo.Non_reproducible Scheduler.Run.Build_cancelled)
+    | Timeout -> `Timeout, times
     | Normal ->
       let output = Result.Out.get result.stdout ^ Result.Out.get result.stderr in
       Log.command ~command_line ~output ~exit_status:process_info.status;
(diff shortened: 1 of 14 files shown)


src/dune_rules/cram/cram_exec.ml
```diff
@@ -446,33 +461,71 @@ let run_and_produce_output ~env ~dir:cwd ~script ~dst =
...
-      List [ path dir; path script; target output ]
+    let encode { src; dir; script; output; timeout } path target : Sexp.t =
+      List
+        [ path src
+        ; path dir
+        ; path script
+        ; target output
+        ; Dune_sexp.Encoder.(option float (Option.map ~f:snd timeout))
+          |> Dune_sexp.to_sexp
+          (* TODO: can we encode the location here? should we? *)
```
rgrinberg: The answer should be no here

src/dune_engine/process.ml
```diff
@@ -1031,6 +1032,11 @@ let run_internal
         ~args
         ()
     in
+    let* () =
+      match cancel with
+      | Some cancel -> Fiber.Ivar.fill cancel _cancel
```
rgrinberg: What is `_cancel`?

src/dune_rules/cram/cram_exec.ml
```diff
@@ -376,7 +376,9 @@ let make_temp_dir ~script =
   temp_dir
 ;;
 
-let run_cram_test env ~script ~cram_stanzas ~temp_dir ~cwd =
+exception Cram_timed_out
```
rgrinberg: Could you try doing this without exceptions?

src/dune_engine/process.ml
Alizter: When we spawn a job through the scheduler, we have access to a `Fiber.Cancel.t` which we can use to signal termination of the process via the scheduler.

I tried exposing some of the job killing functions in the scheduler directly however this causes a race with the process handling thread and it's mutexes.

Adding a timeout to process seemed like a good idea, but I couldn't work out a way to return a "timed out" state, which I would need for good error messages later. It also seemed a little disruptive.

The solution I chose here was to expose the `Fiber.Cancel.t` to the caller of `Process.run` as a callback (via an ivar). This allows the caller to trigger a cancellation of the job directly for whatever reason. In our case, due to a timeout.

As an aside, I also tried a similar strategy with exposing the PID of the process, but for the reasons I highlighted earlier, tracking, terminating and waiting for the PID is a bit finicky and racey if you do it outside of the scheduler. I thought it was therefore best to reuse the job cancellation mechanism already there. (Also cancellation is immediate whereas sending a signal outside of the scheduler seems to be delayed (in my tests)).

The result is that `Process.run` exposes it's cancellation handle for callers to be able to cancel it directly. This is what I ended up using in the cram action.

src/dune_rules/cram/cram_exec.ml
```diff
@@ -428,51 +428,84 @@ module Script = Persistent.Make (struct
...
+      let open Fiber.O in
+      run_and_produce_output
+        ~env:eenv.env
+        ~dir
+        ~script
+        ~dst:output
+        ~timeout:(Option.map ~f:snd timeout)
+      >>| function
+      | Ok () -> ()
+      | Error `Timed_out ->
```
rgrinberg: There's no point to keep around this error return value just to raise in the end. You could just raise immediately if you pass along the location along with the timeout.

test/blackbox-tests/test-cases/cram/timeout-invalid.t
```diff
@@ -0,0 +1,108 @@
...
+
+  $ cat > test.t <<EOF
+  >   $ echo hi
+  > EOF
+
+Negative values fail immediately.
+
+  $ cat > dune <<EOF
+  > (cram
+  >  (timeout -1.0))
```
rgrinberg, line 36: How about we just forbid them?

Alizter: I've added some validation. This also forbids `nan` values. However `inf` is still valid, but harmless for our purposes.

src/dune_rules/cram/cram_exec.ml
```diff
@@ -376,47 +376,59 @@ let make_temp_dir ~script =
   temp_dir
 ;;
 
-let run_cram_test env ~script ~cram_stanzas ~temp_dir ~cwd =
+let run_cram_test env ~loc ~script ~cram_stanzas ~temp_dir ~cwd ~timeout =
```
rgrinberg: You should pass this loc as a tuple with the timeout so that we have an idea what the loc applies to.

rgrinberg (approved):

src/dune_rules/cram/cram_exec.ml
Alizter: We have `timeout : (Loc.t * float) option` here whereas `loc` is just the cram test file:
```
      let loc = Loc.in_file (Path.drop_optional_build_context_maybe_sandboxed src) in
```
I've gone ahead and passed the `src : Path.t` so that we can infer the location from that directly rather than early on.

rgrinberg (changes requested):
I'd like to review how the digest is computed later this week

Alizter:
@rgrinberg I suppose since we are doing this:
```ocaml
    let encode { src; dir; script; output; timeout } path target : Sexp.t =
      List
        [ path src
        ; path dir
        ; path script
        ; target output
        ; Dune_sexp.Encoder.(option float (Option.map ~f:snd timeout))
          |> Dune_sexp.to_sexp
        ]
    ;;
```
we are changing the rule digest every time we change the time. This is confirmed by the test I just pushed: `cram/timeout-redigest.t`.

Reading `build_system.ml` I see that we compute rule digests as:
```ocaml
    let trace =
      ( rule_digest_version (* Update when changing the rule digest scheme. *)
      , sandbox_mode
      , Dep.Facts.digest facts ~env
      , target_paths rule.targets.files @ target_paths rule.targets.dirs
      , Action.for_shell action
      , can_go_in_shared_cache
      , List.map locks ~f:Path.to_string
      , Execution_parameters.action_stdout_on_success execution_parameters
      , Execution_parameters.action_stderr_on_success execution_parameters
      , Execution_parameters.workspace_root_to_build_path_prefix_map execution_parameters
      , Execution_parameters.expand_aliases_in_sandbox execution_parameters )
    in
    Digest.generic trace
```
and `Action.for_shell` uses `encode`. Therefore the encoding directly affects what we digest. 

If we deem that timeout changes shouldn't affect digests then the simple solution would be to remove the timeout from the digest, but I am unsure if this is something we want.

Take the situation where we have a test that succeeds with no timeout, then we add a short timeout that shouldn't succeed. It will not rerun the rule since the digest is the same, but obviously that build can no longer be reproduced.

A more sensible semantic would be for the digest not to change for larger timeouts, but to change for smaller timeouts. However such a thing seems impossible.

rgrinberg:
Re-running the action when the timeout changes is fine, the issue is with allowing the action to observe the name of cram script. That seems very much unnecessary. I've gone ahead and gotten rid of it.

Alizter:
Since we just branched 3.20, we need to wait for the 3.21 version bump.

rgrinberg:
Why does it matter? 3.20 hasn't been released yet.

Alizter:
Right, but the alpha has started, so if we merge this now, we need to remember to bump the version numbers to 3.21 since this patch won't be in 3.20. AFAIK we don't backport new PRs after the alpha.

rgrinberg:
Can't we always just do another alpha? In any case, I'll leave the choice to @maiste to either include this in 3.20 or 3.21.

maiste:
If that's mandatory I can redo an alpha1. What do you prefer?

rgrinberg:
I think users would certainly appreciate being able to use this feature sooner rather than later.

So if it's possible, yes I would do another alpha with this feature.

maiste:
OK then let's go with this. It will allow me to fix some typos that were in the previous changelog.

PR #12042, opened 2025-07-18
flake: add nix targets that implement `dune --version`

gridbugs:
Existing nix targets produce binaries whose version according to `dune --version` is "n/a". This adds a mechanism to the flake for determining the version number and building a dune binary which outputs the appropriate version number from `dune --version`.

Note that this depends on two hacks which are explained in more detail in a comment:
 - determining the version number from the changelog file
 - creating a temporary git repo with a tag named after the version to allow `dune subst` to find out the current version

In the longer term we could remove the needs for these workarounds. In the short term this feature is needed to determine the current version of dune when dune is shipped as a binary, as we use the flake to build said binaries.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -47,9 +47,61 @@
           });
         });
       };
+      # This creates a git repo and creates an annotated tag named after the
+      # current version of dune. This is necessary for the resulting dune
+      # executable to print the correct version in the output of `dune
+      # --version`. The git metadata is not copied to the environment where
+      # this flake is built, so the current tag of the repo can't be known
+      # here. Instead it's assumed that the first word of the first line of the
+      # changelog will be the version number of the current release. The logic
+      # in "dune subst" that determines the output of `dune --version`
+      # determines the version number from git, so we need to create a git repo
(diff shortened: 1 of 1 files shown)


rgrinberg:
@anmonteiro could you review this one?

anmonteiro:
It's a bit hacky, but I'm not sure what the alternative is. The only other solution I can think of is committing a file that has the current version and falling back to that if `Build_info.V1` returns `n/a`. It requires a manual step for promoting the file into the source tree (e.g. a dune rule that produces `fallback.ml` via a `git` invocation) but it wouldn't require the rigmarole of parsing changelog + generating an empty repo.

gridbugs:
I agree that would be a more elegant solution, though it would take some time to design it properly. We can always switch to that approach later without affecting the output of the flake or anything downstream of it. Would you be satisfied merging this in its current state and doing it the "right way" when we have some more time for maintenance? Could probably even be a good first issue.

gridbugs:
I'll close this for the time being. I found that it doesn't work properly on macos and I think fixing it here would be too invasive to be worth it.

PR #12043, opened 2025-07-18
feature: add --alias as an argument for building aliases

### LINKED ISSUE #5775: Current Dune alias syntax can't work reliably on Windows
## Desired Behavior

I maintain the [Diskuv OCaml distribution](https://diskuv.gitlab.io/diskuv-ocaml), and will be soon releasing a shim for Dune so that Windows users can just type `dune build` to build like everybody else does in *nix. The shim will bring a MSYS environment for any Unix binaries that a Dune project needs and will also bring MSVC so OCaml can do native compilation.

Unfortunately [Dune aliases](https://dune.readthedocs.io/en/latest/usage.html#aliases) use the `@` symbol. That means every Windows user who uses PowerShell (the default Windows shell in popular programs like Windows Terminal and VS Code) will be forced to escape any aliases. It sounds like a small problem, but it is a problem that will become a major complaint as more people use Windows on OCaml. So instead of:

```powershell
dune build @foo
```

Windows users have to copy, paste **and modify** it to:

```powershell
dune build `@foo
```

If they don't, the `@` term can be silently removed because it is the PowerShell array operator. Have a look at:

```powershell
$ echo dune build @foo
dune
build

$ echo dune build @foo/src/all
dune
build
@foo/src/all
```

Basically, the `@foo` array in PowerShell is acting like the `${foo[@]}` array in Bash ... if the array variable does not exist it is empty.

Hopefully it is apparent how confusing or frustrating that will be to Windows Dune users, and time-consuming for any OCaml package maintainer who has to respond to bug reports why their `dune` command line does not work.

## Proposed

Either of the following:

1. Use `dune build =foo` or `dune build +foo` instead of `dune build @foo`.
2. Use `dune alias foo` instead of `dune build @foo`.

In both cases I would remove `dune build @foo` from your documentation but keep that syntax around for backwards compatibility.

rgrinberg:
Turns out that typing `@` in the command line can be problematic on Windows. This PR adds an alternative way to specify aliases in the command line using the `--alias` and the `--alias-rec` flags.

Fixes https://github.com/ocaml/dune/issues/5775

### DIFF
--- a/bin/arg.ml
+++ b/bin/arg.ml
@@ -116,6 +116,16 @@ module Dep = struct
 
   let conv = conv' (parser, printer)
   let to_string_maybe_quoted t = String.maybe_quoted (Format.asprintf "%a" printer t)
+
+  let alias_arg =
+    let parse x = Ok (Dep_conf.Alias (String_with_vars.make_text Loc.none x)) in
+    conv' (parse, printer)
+  ;;
+
+  let alias_rec_arg =
+    let parse x = Ok (Dep_conf.Alias_rec (String_with_vars.make_text Loc.none x)) in
+    conv' (parse, printer)
+  ;;
 end
 
 let dep = Dep.conv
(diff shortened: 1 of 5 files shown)


Alizter (approved):
Seems sensible to me.

PR #12053, opened 2025-07-20
fix(melange): build time benchmarks

anmonteiro:
These seem to be failing with:

```
[WARNING] Failed checks on pupilfirst package definition from source at
git+file:///home/runner/work/dune/dune/pupilfirst#master: warning 74:
Field 'pin-depends' contains packages that are neither in 'depends' nor
in 'depopts': "reason-react-ppx"
```

not entirely sure that the pin-depends are being respected.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - main
-      - ci-bench/fix-after-ocaml51
 
 permissions:
   # deployments permission to deploy GitHub pages website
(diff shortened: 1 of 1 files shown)


anmonteiro:
OK it sounds like this is related to OPAM 2.4 not respecting pins for some reason when you run `opam install . --deps-only`.

anmonteiro:
This was a regression in OPAM: https://github.com/ocaml/opam/pull/6611

Thanks @kit-ty-kate for the help on this one.

PR #12064, opened 2025-07-22
[RPC] Allow formatting while watch server is running

ElectreAAS:
This PR is a second step towards implementing #11958, and is the 'big sister' of #12010.
~~Like its sister, this PR contains stolen code from Steve's #11900.~~ 'Stolen' code has been merged on main

~~It is currently in draft mode~~ It is now ready for review! The mechanism for promotion is now sound :)

### DIFF
--- a/bin/build.ml
+++ b/bin/build.ml
@@ -97,7 +97,7 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
          match kind with
          | Build targets ->
            Target.interpret_targets (Common.root common) config setup targets
-         | Format _promote ->
+         | Format promote ->
            let request (setup : Import.Main.build_system) =
              let dir = Path.(relative root) (Common.prefix_target common ".") in
              Alias.in_dir
@@ -107,8 +107,12 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
                dir
              |> Alias.request
            in
-           (* Having Diff_promotion.promote_files_registered_in_last_run here doesn't work for some reason??? *)
-           request setup
+           (match promote with
+            | Automatically ->
+              Action_builder.map (request setup) ~f:(fun () ->
+                Promote.Diff_promotion.promote_files_registered_in_last_run
+                  Dune_rpc.Files_to_promote.All)
+            | Never -> request setup)
        in
        run_build_system ~common ~request, outcome)
 ;;
--- a/bin/build.ml
+++ b/bin/build.ml
@@ -92,24 +92,11 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
   in
   Dune_engine.Scheduler.Run.poll_passive
     ~get_build_request:
-      (let+ { outcome; kind } = Dune_rpc_impl.Server.pending_action rpc in
+      (let+ (Build (targets, ivar)) = Dune_rpc_impl.Server.pending_build_action rpc in
        let request setup =
-         match kind with
-         | Build targets ->
-           Target.interpret_targets (Common.root common) config setup targets
-         | Format _promote ->
-           let request (setup : Import.Main.build_system) =
-             let dir = Path.(relative root) (Common.prefix_target common ".") in
-             Alias.in_dir
-               ~name:Dune_rules.Alias.fmt
-               ~recursive:true
-               ~contexts:setup.contexts
-               dir
-             |> Alias.request
-           in
-           request setup
+         Target.interpret_targets (Common.root common) config setup targets
        in
-       run_build_system ~common ~request, outcome)
+       run_build_system ~common ~request, ivar)
 ;;
 
 let run_build_command_poll_eager ~(common : Common.t) ~config ~request : unit =
--- a/bin/build.ml
+++ b/bin/build.ml
@@ -97,7 +97,7 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
          match kind with
          | Build targets ->
            Target.interpret_targets (Common.root common) config setup targets
-         | Format promote ->
+         | Format _promote ->
            let request (setup : Import.Main.build_system) =
              let dir = Path.(relative root) (Common.prefix_target common ".") in
              Alias.in_dir
@@ -107,9 +107,6 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
                dir
              |> Alias.request
            in
-           Printf.eprintf
-             "From server, promote is: %s\n%!"
-             (Dune_rpc.Promote_flag.to_string promote);
            (* Having Diff_promotion.promote_files_registered_in_last_run here doesn't work for some reason??? *)
            request setup
        in
--- a/bin/build.ml
+++ b/bin/build.ml
@@ -109,7 +109,7 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
            in
            Printf.eprintf
              "From server, promote is: %s\n%!"
-             (Dune_rpc.Promote.to_string promote);
+             (Dune_rpc.Promote_flag.to_string promote);
            (* Having Diff_promotion.promote_files_registered_in_last_run here doesn't work for some reason??? *)
            request setup
        in
--- a/bin/build.ml
+++ b/bin/build.ml
@@ -97,7 +97,7 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
          match kind with
          | Build targets ->
            Target.interpret_targets (Common.root common) config setup targets
-         | Format promote ->
+         | Format _promote ->
            let request (setup : Import.Main.build_system) =
              let dir = Path.(relative root) (Common.prefix_target common ".") in
              Alias.in_dir
@@ -107,12 +107,7 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
                dir
              |> Alias.request
            in
-           (match promote with
-            | Automatically ->
-              Action_builder.map (request setup) ~f:(fun () ->
-                Promote.Diff_promotion.promote_files_registered_in_last_run
-                  Dune_rpc.Files_to_promote.All)
-            | Never -> request setup)
+           request setup
        in
        run_build_system ~common ~request, outcome)
 ;;
--- a/bin/build.ml
+++ b/bin/build.ml
@@ -107,14 +107,11 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
                dir
              |> Alias.request
            in
-           (match promote with
-            | Never -> request setup
-            | Automatically ->
-              (* Highly sketchy code right here *)
-              let previous = !Dune_engine.Clflags.promote in
-              Dune_engine.Clflags.promote := Some Automatically;
-              Action_builder.map (request setup) ~f:(fun () ->
-                Dune_engine.Clflags.promote := previous))
+           Printf.eprintf
+             "From server, promote is: %s\n%!"
+             (Dune_rpc.Promote.to_string promote);
+           (* Having Diff_promotion.promote_files_registered_in_last_run here doesn't work for some reason??? *)
+           request setup
        in
        run_build_system ~common ~request, outcome)
 ;;
--- a/bin/build.ml
+++ b/bin/build.ml
@@ -92,11 +92,31 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
   in
   Dune_engine.Scheduler.Run.poll_passive
     ~get_build_request:
-      (let+ (Build (targets, ivar)) = Dune_rpc_impl.Server.pending_build_action rpc in
+      (let+ { outcome; kind } = Dune_rpc_impl.Server.pending_action rpc in
        let request setup =
-         Target.interpret_targets (Common.root common) config setup targets
+         match kind with
+         | Build targets ->
+           Target.interpret_targets (Common.root common) config setup targets
+         | Format promote ->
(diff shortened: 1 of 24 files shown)


bin/build.ml
```diff
@@ -92,11 +92,31 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
...
+             |> Alias.request
+           in
+           (match promote with
+            | Never -> request setup
+            | Automatically ->
+              (* Highly sketchy code right here *)
+              let previous = !Dune_engine.Clflags.promote in
+              Dune_engine.Clflags.promote := Some Automatically;
+              Action_builder.map (request setup) ~f:(fun () ->
+                Dune_engine.Clflags.promote := previous))
```
ElectreAAS: This is very dubious, but it kinda works.

I'm changing the promote command line flag to true and back to whatever it was, right before and after doing the normal fmt request. This is a non-synchronised access to a global variable, in other words a data race just begging to happen.
Even if it wasn't racy, it would still cause problems: what if we have another RPC request running at the same time, and that one sees the new flag we set up?

Another way to do it would be to let the formatting request finish, and then do a promotion (either directly or sending another RPC request to itself); I didn't do that because 1. I couldn't figure out a way to. 2. that would mean 2 different instructions where the happy path for formatting only needs one

A better way to do it is what I'm currently looking for

test/blackbox-tests/test-cases/watching/fmt-test.t
```diff
@@ -0,0 +1,37 @@
...
+  >  print_int
+  >                         (5
+  >       +
+  >                          4)
+  > 
+  > EOF
+
+  $ start_dune
+
+Ultimately we'd want this warning to not appear at all, since `dune fmt` doesn't have arguments
```
ElectreAAS: I couldn't make it so that this warning is not triggered without de-generalizing code that I generalized...

shonfeder:
Hi! Could you provide some context by explaining what you find sketchy about the enabling?

bin/build.ml
```diff
@@ -92,11 +92,31 @@ let poll_handling_rpc_build_requests ~(common : Common.t) ~config =
...
+               ~name:Dune_rules.Alias.fmt
+               ~recursive:true
+               ~contexts:setup.contexts
+               dir
+             |> Alias.request
+           in
+           (match promote with
+            | Never -> request setup
+            | Automatically ->
+              (* Highly sketchy code right here *)
```
shonfeder: Ah, I see you have a comment here. Could you explain what is highly sketchy about it?

bin/build.ml
shonfeder: Could you point out the code that ends up reading this variable? I wonder if that would help us figure out alternative ways to git this threaded thru...

test/blackbox-tests/test-cases/watching/fmt-test.t
```diff
@@ -0,0 +1,39 @@
...
+  $ cat > foo.ml << EOF
+  > let () =
+  >  print_int
+  >                         (5
+  >       +
+  >                            4)
+  > 
+  > EOF
+
+  $ start_dune
```
Alizter, line 13: What you need to do here is use some sort of wait mechanism for `_build/.lock`. There are such mechanisms in the wait-for-file.sh script of exec-watch. Something like

```
#!/bin/sh
#
# Wait until a file exists, then delete the file and exit
#

set -u
FILE=$1
until test -e $FILE
do
    sleep 0.01
done
rm $FILE
```

Once you know that the passive watch mode has created the lock, then you know the rest of the test will act accordingly.

shonfeder: IIUC, the plan here is to use the ping rpc, once https://github.com/ocaml/dune/pull/12138 is merged. I can take care of that update, since @ElectreAAS is out for the next week.

ElectreAAS: Indeed using `dune rpc ping --wait` is just simpler, thanks for #12138!

src/dune_rpc_impl/server.ml
```diff
@@ -329,17 +327,32 @@ let handler (t : _ t Fdecl.t) handle : 'build_arg Dune_rpc_server.Handler.t =
...
+    Handler.implement_request rpc Decl.build f
+  in
+  let () =
+    let f _ promote =
+      let server = Fdecl.get t in
+      let outcome = Fiber.Ivar.create () in
+      let target = server.parse_build_arg "(alias_rec fmt)" in
+      let* () = Job_queue.write server.pending_build_jobs ([ target ], outcome) in
+      let+ _build_outcome = Fiber.Ivar.read outcome in
+      match promote with
```
Alizter: We can actual take action on the `_build_outcome` here because if it was "successful", then there is no point promoting.

src/dune_engine/clflags.ml
```diff
@@ -1,20 +1,3 @@
-module Promote = struct
```
Alizter: Looks like `shared_cache_default` this accidentally got caught too.

src/dune_rpc_impl/server.ml
```diff
@@ -329,17 +327,33 @@ let handler (t : _ t Fdecl.t) handle : 'build_arg Dune_rpc_server.Handler.t =
...
       | Failure -> Failure (get_current_diagnostic_errors ())
     in
-    Handler.implement_request rpc Decl.build build
+    Handler.implement_request rpc Decl.build f
+  in
+  let () =
+    let f _ promote =
+      let server = Fdecl.get t in
+      let outcome = Fiber.Ivar.create () in
+      let target = server.parse_build_arg "(alias_rec fmt)" in
```
ElectreAAS: Not sold on the idea of having the target be determined by a string parse like this, isn't there a cleaner way?

Alizter: We don't have access to any of the information of the build targets here since interpretation comes at a later point. If you are uncomfortable by the string, you could use 
```ocaml
      let target =
        Sexp.(List [ Atom "alias_rec"; Atom "fmt" ])
        |> Sexp.to_string
        |> server.parse_build_arg
      in
```
but ultimately it's the same thing.

bin/rpc/rpc_common.ml
```diff
@@ -89,24 +89,26 @@ let wrap_build_outcome_exn ~print_on_success f args () =
...
+        |> Pp.tag User_message.Style.Error
+      ]
   | Ok Dune_rpc.Build_outcome_with_diagnostics.Success ->
     if print_on_success
-    then
-      Console.print_user_message
-        (User_message.make [ Pp.text "Success" |> Pp.tag User_message.Style.Success ])
+    then Console.print [ Pp.text "Success" |> Pp.tag User_message.Style.Success ]
   | Ok (Failure errors) ->
+    let error_msg =
```
Alizter: Why was this pulled before the errors?

bin/rpc/rpc_common.ml
```diff
@@ -89,24 +89,26 @@ let wrap_build_outcome_exn ~print_on_success f args () =
   let+ response = f args in
   match response with
   | Error (error : Rpc_error.t) ->
-    Printf.eprintf "Error: %s\n%!" (Dyn.to_string (Rpc_error.to_dyn error))
+    Console.print
```
Alizter: Let's put these changes in a separate PR.

test/blackbox-tests/test-cases/watching/fmt-test.t
```diff
@@ -0,0 +1,35 @@
+  $ . ./helpers.sh
```
Alizter, line 1: For this test, let's add a cram stanza to the relevent dune file that declares a dependency on ocamlformat.

Alizter: In fact, scratch that. Let's use the fake ocamlformat binary we have. It's in `test/blackbox-tests/utils/`, you don't have to have such a funky input file either since it will output `(* fake ocamlformat output *)`. You will need to declare a dependency on that binary for it to appear in the test.

bin/rpc/rpc_common.ml
ElectreAAS: Yep

bin/rpc/rpc_common.ml
ElectreAAS: 1. Having a `User_error.raise (... Code_error.raise ...)` looks very weird
2. Subjective style choice: I prefer saying "Build failed with X errors: [errors]" than the reverse. Feels like a title and then contents, instead of a postscript.
It also belongs in a separate PR, feel free to just refuse that one ^^

Alizter: Let's discuss it in a separate PR then.

src/dune_rpc_impl/server.ml
rgrinberg: > Not sold on the idea of having the target be determined by a string parse like this, isn't there a cleaner way?

The cleaner way is to introduce a new build request that offers a stable type for targets.

You are right to be suspicious of this since the way the server is parsing this string depends is dependent on the dune lang. This makes RPC more fragile than needed.

src/dune_engine/clflags.ml
ElectreAAS: I don't understand what you mean here. There is a variable named `can_go_in_shared_cache_default` in this file, but it is untouched.
The module Promote was moved (and renamed to Promote_flag) to `dune-rpc/private/exported_types.ml` because it is used in the rpc

Alizter: There was a bug in the github UI where I was seeing a line 0 and it had wrapped the bottom to the top. It didn't let me comment at the time at line 0 so I added it at line 1. Turns out it was just a github bug. Sorry about that.

test/blackbox-tests/test-cases/watching/fmt-test.t
ElectreAAS: Agreed, done :)

src/dune_rpc_impl/dune
```diff
@@ -1,5 +1,6 @@
 (library
  (name dune_rpc_impl)
- (libraries stdune promote unix fiber csexp_rpc dune_stats dune_rpc_client
-  dune_console dune_util dune_rpc_private dune_rpc_server dune_engine)
+ (libraries csexp_rpc dune_console dune_engine dune_lang dune_rpc_client
+  dune_rpc_private dune_rpc_server dune_stats dune_util fiber promote stdune
+  unix)
```
ElectreAAS, line 5: Pet peeve, but this should be sorted. I added dune_lang as I pull it up as dependency in server.ml

src/dune_rpc_impl/server.ml
ElectreAAS: `Dune_lang.Dep_conf.Alias_rec (Dune_lang.String_with_vars.make_text Loc.none "fmt")` is how it's now done. If there is another way about doing this then I don't know what it is

src/dune_rpc_impl/dune
```diff
@@ -1,16 +1,18 @@
...
   dune_rpc_server
-  dune_engine)
+  dune_stats
+  dune_util
+  fiber
+  promote
+  stdune
+  unix)
  (synopsis "Dune's rpc server + a usable client"))
+
```
ElectreAAS: Pet peeve, but this should be sorted. I added dune_lang as I pull it up as dependency in server.ml

otherlibs/dune-rpc/private/dune_rpc_private.mli
```diff
@@ -194,6 +194,7 @@ module Procedures : sig
     val ping : (unit, unit) Decl.Request.t
     val diagnostics : (unit, Diagnostic.t list) Decl.Request.t
     val shutdown : unit Decl.Notification.t
+    val format : (Promote_flag.t, unit) Decl.Request.t
```
rgrinberg: When would the user ever want to pass `Never` here? I can't really imagine when. How about we just make this PRC just force the promotion?

ElectreAAS: the `Never` is coming from the user doing `dune fmt --preview`
People who use that normally would be confused if the flag didn't exist in case of a watch server processing the command

rgrinberg: That does make the CLI more consistent, but at the cost of making the RPC itself less coherent. If we wanted to properly support previewing, we'd want to somehow communicate what/where is being previewed via RPC to make this usable. Right now, it doesn't seem like a well thought out parameter.

FWIW, I've never used the `--preview` flag myself and I think it's barely useful. So I don't particularly care about supporting it.

You can make the final call here.

rgrinberg (approved):

shonfeder:
@ElectreAAS Can this be merged now?

ElectreAAS:
> @ElectreAAS Can this be merged now?

~~No! There is a problem with the cache, and the `--preview` situation isn't yet under control~~

Now, yes it is!

otherlibs/dune-rpc/private/dune_rpc_private.mli
ElectreAAS: Final call: the `--preview` flag, while useful, is going to be ignored by the RPC system.
If we think this is too restrictive, we are free to change this in a later PR.

Reasoning: communicating what's being previewed via RPC isn't actually the challenge, the challenge comes from the cache that doesn't know the difference between a preview run and an actual run, which means the promotion database is not updated if one runs `dune fmt --preview` followed by `dune fmt`.
Also I just want this PR to go through at this point.

PR #12065, opened 2025-07-23
pkg: add dev tools to PATH when running dev tools

gridbugs:
This allows the ocamllsp dev tool to run the ocamlformat dev tool, assuming the latter is installed.

Addresses https://github.com/ocaml/dune/issues/10964

cc @pitag-ha

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/ocamllsp/dev-tool-ocamllsp-env-path-var.t
+++ b/test/blackbox-tests/test-cases/pkg/ocamllsp/dev-tool-ocamllsp-env-path-var.t
@@ -1,0 +1,35 @@
+Test that the ocamllsp dev tool executes in an environment where other dev
+tools are in PATH.
+
+  $ . ../helpers.sh
+  $ . ./helpers.sh
+
+  $ mkrepo
+  $ mkpkg ocaml 5.2.0
+  $ setup_ocamllsp_workspace
+
+Make a fake ocamllsp package that prints out the PATH variable:
+  $ mkpkg ocaml-lsp-server <<EOF
+  > install: [
+  >   [ "sh" "-c" "echo '#!/bin/sh' > %{bin}%/ocamllsp" ]
+  >   [ "sh" "-c" "echo 'echo \$PATH' >> %{bin}%/ocamllsp" ]
+  >   [ "sh" "-c" "chmod a+x %{bin}%/ocamllsp" ]
+  > ]
+  > EOF
+
+  $ make_lockdir
+  $ cat > dune.lock/ocaml.pkg <<EOF
+  > (version 5.2.0)
+  > EOF
+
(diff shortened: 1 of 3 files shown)


pitag-ha:
Nice, thanks a lot! CC also @PixieDust .

Do I understand correctly that with this PR any dev tool launched via `dune tools exec <tool>` gets run in an environment where all other installed dev tools are available? (If so: :tada:)

Would it be hard to generalize this so that a dev tool B would also be available in the environment of dev tool A, even if B is installed after A was launched? Maybe the directory where a dev tool can be found once installed can be added to the environment of another dev tool?

test/blackbox-tests/test-cases/pkg/ocamllsp/dev-tool-ocamllsp-env-path-var.t
```diff
@@ -0,0 +1,35 @@
...
+tools are in PATH.
+
+  $ . ../helpers.sh
+  $ . ./helpers.sh
+
+  $ mkrepo
+  $ mkpkg ocaml 5.2.0
+  $ setup_ocamllsp_workspace
+
+Make a fake ocamllsp package that prints out the PATH variable:
```
shonfeder, line 11: Needing this workaround for testing seems to me like a good indication that we should just have a way of doing `dune tools exec -- foo`, so we could do `dune tools exec -- echo $PATH` That can be added as a followup. I assume it would be pretty straightforward. To you agree!

In any case, I think a TODO here to note that we should replace this workaround in the future would make sense.

shonfeder (approved):

gridbugs:
> Would it be hard to generalize this so that a dev tool B would also be available in the environment of dev tool A, even if B is installed after A was launched? Maybe the directory where a dev tool can be found once installed can be added to the environment of another dev tool?

That's how this currently works. Dune runs the tools in an environment where the PATH variable contains the paths where tools will be installed regardless of whether or not they are currently installed.

test/blackbox-tests/test-cases/pkg/ocamllsp/dev-tool-ocamllsp-env-path-var.t
gridbugs: I'm hesitant to make this change because `dune tools exec <tool>` installs the tool as well as running it. If we started allowing arbitrary commands then users will need to know which commands will cause tools to be installed into the environment. I think either we should change `dune tools exec` to just a way of running commands in an environment with access to tools and not also install the tools (tools can be installed with `dune tools install <tool>`) or introduce a new command for running commands in a modified environment.

pitag-ha:
Ah I see, nice! (I misunderstood "This allows the ocamllsp dev tool to run the ocamlformat dev tool, assuming the latter is installed.")

Thank you! :tada:

PR #12067, opened 2025-07-23
test: tabs in cram tests cause confusion

Alizter:
In this PR, we replace leading tab characters on a line with two spaces.

The previous behaviour would lead to confusing situations for users since their commands would be interpreted as comments and ignored.

We change it so that leading tabs are interpreted the same as two spaces and this leads to them being replaced by spaces during promotion.

The first commit adds a commit specifying the current behaviour with tabs being preserved and the second commit introduces the feature and how it effects the running of cram tests.

- [ ] possibly breaking change? (imo for the better)
- [ ] changelog
- [ ] documentation


test/blackbox-tests/test-cases/cram/cram-tab.t
```diff
@@ -0,0 +1,37 @@
+Sometimes a user may misconfigure their editor and end up putting a tab
+character before a command in a cram test instead of two spaces.
+
+Dune would register this as a comment, leaving the user confused as to why
+their command is not being run. The current behaviour is for a leading tab to
+be treated and replaced as / by two spaces.
```
shonfeder: Took me a few reads to understand this last bit. I think it would be clearer as:

```suggestion
be treated as, and replaced by, two spaces during promotion.
```

shonfeder (approved):

test/blackbox-tests/test-cases/cram/cram-tab.t
Alizter: Thanks I've done that. I agree that it was confusing before.

src/dune_rules/cram/cram_lexer.mll
```diff
@@ -8,11 +8,13 @@ let eol = '\n' | eof
 
 let blank = [' ' '\t' '\r' '\012']
```
Alizter, line 9: We don't even use this, but I won't remove it here.

rgrinberg:
It is a breaking change. Tests passing currently are going to start failing.

Going back to the main motivation here, which seems to be to help users with misconfigured editors. I don't think we're doing them a favor here, because we're not letting them know their editor is misconfigured.

More helpful would be just to error out and tell them that leading tabs aren't allowed.

Let me ask around some mercurial folks if they've decided on what to do here.

rgrinberg:
I discussed this with other folks experienced working with this and nobody liked this idea of silently converting tabs like this.

Although I have nothing against tabs, my advice to avoid them would be error out if there are tabs present and let the user adjust their editor appropriately. Doing it silently like this isn't the way to go.

PR #12068, opened 2025-07-24
pkg: lock utop dev tool in separate scheduler run

gridbugs:
Locking the dev tool within the same scheduler run as building utop causes the memoized filesystem functions to not be able to see utop's lockdir.

### DIFF
--- a/bin/ocaml/utop.ml
+++ b/bin/ocaml/utop.ml
@@ -12,10 +12,12 @@ let man =
 
 let info = Cmd.info "utop" ~doc ~man
 
-let lock_utop_if_dev_tool_enabled () =
+let lock_utop_if_dev_tool_enabled ~common ~config =
   match Lazy.force Lock_dev_tool.is_enabled with
-  | false -> Memo.return ()
-  | true -> Lock_dev_tool.lock_dev_tool Utop
+  | false -> ()
+  | true ->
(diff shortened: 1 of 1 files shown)


shonfeder (approved):

rgrinberg:
I'm not sure what's going here, but the description doesn't appear to make sense. Calling Scheduler.run twice does not change the state of the memoized functions.

Alizter:
@gridbugs Do you have a reproduction case for the issue?

gridbugs:
@rgrinberg ah ok my bad. The issue this fixes is that `DUNE_CONFIG__LOCK_DEV_TOOL=enabled dune utop` would create a lockdir for utop but then would be unable to see the lockdir when building utop, but running the command for a second time worked. My mental model for Fs_memo was incorrect. I'm looking again to see why this change fixed the issue.

gridbugs:
I've reverted this and fixed the issue without a separate scheduler run here: https://github.com/ocaml/dune/pull/12073. I assumed that `Fs_memo.dir_exists` would memoize its result and then miss the creation of the lockdir, and that the memoized state is specific to the scheduler run (which I now know is not true). I'm still not sure whether `Fs_memo.dir_exists` memoizes its result or not but it's moot because I was checking for the existence of the lockdir in a lazy memo, and I didn't realize it was being forced as a side effect of checking whether the utop exe exists.

Issue #12057, opened 2025-07-21
Incorrect error message: refers to .ml instead of .mli when interface mismatch occurs


RezwanArefin01:
In case of interface mismatch between a `.ml` and `.mli` file, the error message incorrectly refers to the `.ml` file as interface.

## Expected Behavior
`Error: The implementation main.ml does not match the interface main.mli:`

## Actual Behavior
`Error: The implementation main.ml does not match the interface main.ml:`

## Reproduction
**dune-project**
```
(lang dune 3.15)
```
**dune**
```
(executable
 (name main))
```
**main.ml**
```
let f x = x * x
```
**main.mli**
```
val f : float -> float
```
Then build:
```bash
$ dune build
File "main.ml", line 1:          
Error: The implementation main.ml does not match the interface main.ml: 
   Values do not match:
     val f : int -> int
   is not included in
     val f : float -> float
   The type int -> int is not compatible with the type float -> float
   Type int is not compatible with type float
   File "main.mli", line 1, characters 0-22: Expected declaration
   File "main.ml", line 1, characters 4-5: Actual declaration
```

## Specifications

- Version of `dune`: 3.19.1
- Version of `ocaml`: 5.3.0
- Operating system: (Arch)Linux 6.12.36-1-lts

## Additional information 

- Link to gist with verbose output: https://gist.github.com/RezwanArefin01/3188227f0a66e3dff007abc2197411aa

Alizter:
Can you also check that the ocaml compiler doesn't give the same message. These messages are typically generated by the compiler so I would be surprised if dune had anything to do with it. If it does turn out to be an OCaml compiler issue, you should report it there.

Alizter:
Yes, this does appear to be an issue. I see that the `ocamlc` error message is correct, but whatever is happening in dune is incorrect. I will make a reproduction case shortly.

nojb:
I suspect this is somehow related to the use of the flag `-intf-suffix .ml` while compiling the file in question.

Alizter:
I had a little look at this and I am not sure what we can do to get around it. For context, the reason we pass `-intf-suffix` is explained in the source code:

>   The compiler creates the cmi when it thinks a .ml file has no corresponding
   .mli. However this behavior is a bit racy and doesn't work well when the
   extension is not .ml or when the .ml and .mli are in different directories.
   This flags makes the compiler think there is a .mli file and will the read
   the cmi file rather than create it. *)

[...]

PR #12072, opened 2025-07-24
fix: mismatch between ml and mli error message

Alizter:
This PR fixes the issue observed in https://github.com/ocaml/dune/issues/12057.

We use the -cmi-file flag instead of -intf-suffix to tell the compiler not to create a .cmi file during module compilation. This has been available since OCaml 5.0. This fixes an issue where the compiler would report the interface file using the extension given by -intf-suffix which is wrong since it was a hack to prevent cmi creation.

I think we should do some later refactoring here to use `Command` rather than manual dependency tracking here.


src/dune_rules/module_compilation.ml
```diff
@@ -133,7 +142,7 @@ let build_cm
    let open Memo.O in
    let* extra_args, extra_deps, other_targets =
      if precompiled_cmi
-     then Memo.return (force_read_cmi src, [], [])
+     then Memo.return ([ "-intf-suffix"; Path.extension src ], [], [])
```
Alizter: This is needed here because I don't know where the `precompiled_cmi` lives in this case, so I can't add it as an argument. It already doesn't count as a dependency here. Since it doesn't affect the issue I am trying to fix, I just inlined what was here before.

src/dune_rules/module_compilation.ml
```diff
@@ -8,7 +8,16 @@ open Memo.O
    extension is not .ml or when the .ml and .mli are in different directories.
    This flags makes the compiler think there is a .mli file and will the read
    the cmi file rather than create it. *)
-let force_read_cmi source_file = [ "-intf-suffix"; Path.extension source_file ]
+let force_read_cmi ~obj_dir ~version ~src ~cm_kind module_ =
+  match Version.supports_cmi_file version, Lib_mode.of_cm_kind cm_kind with
```
rgrinberg, line 16: @anmonteiro does melange support `-cmi-file`? I imagine that it does.

src/dune_rules/module_compilation.ml
```diff
@@ -8,7 +8,16 @@ open Memo.O
...
+  | true, Ocaml _ ->
+    [ "-cmi-file"
+    ; Obj_dir.Module.cm_file_exn obj_dir module_ ~kind:(Lib_mode.Cm_kind.cmi cm_kind)
+      |> Path.Build.drop_build_context_exn
+      |> Path.Source.to_string
```
rgrinberg: Shouldn't this be `Path.reach`?

src/dune_rules/module_compilation.ml
rgrinberg: It lives in the current object directory. So the exact same place as in the other case. I think we should stick to one code path here.

Alizter: When I used the other `-cmi-file` flag the OCaml compiler complained that the cmi file didn't exist. When I added the dependency on the expected cmi file, the rules complained that there was no way to generate that cmi file. I'm a little confused where the precompiled cmi is meant to be, or maybe it has a different name?

src/dune_rules/module_compilation.ml
Alizter: I checked melange myself and couldn't find the option, and also melange rejected the flag. Therefore I opted to keep the previous behaviour for melange. Of course, if there is a better solution I am happy to look into it.

src/dune_rules/module_compilation.ml
rgrinberg: Wait, actually the flag here is used to throw away the inferred cmi. We should really use the correct cmi, but as you find out, it might not exist in this library. For now, this is OK.

src/dune_rules/module_compilation.ml
```diff
@@ -8,7 +8,16 @@ open Memo.O
...
    This flags makes the compiler think there is a .mli file and will the read
    the cmi file rather than create it. *)
-let force_read_cmi source_file = [ "-intf-suffix"; Path.extension source_file ]
+let force_read_cmi ~dir ~obj_dir ~version ~src ~cm_kind module_ =
+  match Version.supports_cmi_file version, Lib_mode.of_cm_kind cm_kind with
+  | true, Ocaml _ ->
+    [ "-cmi-file"
+    ; Obj_dir.Module.cm_file_exn obj_dir module_ ~kind:(Lib_mode.Cm_kind.cmi cm_kind)
+      |> Path.build
+      |> Path.reach ~from:(Path.build dir)
```
rgrinberg: Actually, let's just make this return `Command.Args.t` and not have to do this stuff ourselves. I think you're passing the wrong dir here anyway. It should be the cwd of the ocamlc process, which `Command` would know about.

src/dune_rules/module_compilation.ml
anmonteiro: We don't support `-cmi-file` but I'm not opposed to adding it so we can move towards a single code path in the future.

rgrinberg: If melange would benefit from the improved error message, then I'd encourage you to implement the flag.

Otherwise, there's no particular need as we need to keep the old code path for older OCaml anyway.

src/dune_rules/module_compilation.ml
Alizter: I'll add a CR.

rgrinberg:
Thanks!

PR #12074, opened 2025-07-25
Use copy_file_range when supported

### LINKED ISSUE #12071: Cache: Support Copy on Write (`cp --reflink`)
## Desired Behavior

There are currently two cache storage modes: `Hardlink` and `Copy`. `Hardlink` does not work across mount points so `Copy` is the only option when the cache sits on a separate volume. `Copy` unconditionally creates a full copy of the file and thus `dune` ends up always duplicating the entire content of `_build`.

File systems such as `btrfs` (and possibly recent versions of `zfs` and `xfs`) support copy on write across subvolumes/partitions/mount points. If the `Copy` storage mode was realized through the appropriate APIs that perform copy on write (if possible), `dune` could transparently save precious space without impacting user experience at all. In fact, it would be much faster than actually copying the file contents. This solution would be my preferred way to support copy on write file systems.

Alternatively (and perhaps additionally), `dune` could offer a third storage mode, perhaps called `CopyOnWrite` or `Reflink`, that unconditionally uses copy on write and fails if it is not available. This option is probably less useful since people can build `dune` projects anywhere on their systems, even on removable media and ram disks that are not related to the file system on which the cache resides. *Always* demanding copy on write will likely end up being a foot gun in cases like these.

## Example

My own system is setup up in such a way that I exclude `~/.cache/dune` from my automatic file system snapshots. For this, the directory has to be its own subvolume (`btrfs` terminology; think "partition") and, thus, its own mount point. This forces me to use the `Copy` mode for `dune`'s cache. The result is that after almost 2 weeks on this new system I have over 20GB of duplicate files between `~/.cache/dune` and my various `_build` folders. (Reported by `fclones` which unfortunately cannot yet deduplicate read-only files.)

To replicate this, any `dune` setup that uses the `Copy` mode should be sufficient. Then build a project and either measure disk space if your file system reports that accurately or use a tool such as `fclones` to find duplicates between the `dune`'s cache and the project's `_build`.

nojb:
Use `copy_file_range` when supported (Linux kernel >= 5.19). We check the kernel version at runtime (but we still require the `copy_file_range` function to be available at compile-time; we could avoid this by loading the symbol with `dlsym`).

Fixes #12071

### DIFF
--- a/otherlibs/stdune/src/copyfile_stubs.c
+++ b/otherlibs/stdune/src/copyfile_stubs.c
@@ -67,6 +67,7 @@ CAMLprim value stdune_sendfile(value v_in, value v_out, value v_size) {
 #include <sys/sendfile.h>
 #include <sys/utsname.h>
 #include <linux/version.h>
+#include <dlfcn.h>
 #include <stdio.h>
 
 #define FD_val(value) Int_val(value)
@@ -89,10 +90,14 @@ static ssize_t dune_sendfile(int in, int out, size_t length) {
   return length;
 }
 
+typedef ssize_t (*copy_file_range_t)(int, loff_t*, int, loff_t*, size_t, unsigned int);
+
+static copy_file_range_t copy_file_range_fn = NULL;
+
 static ssize_t dune_copy_file_range(int in, int out, size_t length) {
   ssize_t ret;
   while (length > 0) {
-    ret = copy_file_range(in, NULL, out, NULL, length, 0);
+    ret = copy_file_range_fn(in, NULL, out, NULL, length, 0);
     if (ret < 0) {
       return ret;
     }
@@ -117,7 +122,8 @@ static int kernel_version(void) {
 CAMLprim value stdune_sendfile(value v_in, value v_out, value v_size) {
   static ssize_t (*dune_copyfile)(int, int, size_t) = NULL;
   if (dune_copyfile == NULL) {
-    if (kernel_version() < KERNEL_VERSION(5, 19, 0)) {
+    if (kernel_version() < KERNEL_VERSION(5, 19, 0) ||
+        (copy_file_range_fn = (copy_file_range_t)dlsym(NULL, "copy_file_range")) == NULL) {
       dune_copyfile = &dune_sendfile;
     } else {
       dune_copyfile = &dune_copy_file_range;
--- a/otherlibs/stdune/src/copyfile_stubs.c
+++ b/otherlibs/stdune/src/copyfile_stubs.c
@@ -1,5 +1,3 @@
-#define _GNU_SOURCE
-
 #include <caml/fail.h>
 #include <caml/memory.h>
 #include <caml/mlvalues.h>
--- a/otherlibs/stdune/src/copyfile_stubs.c
+++ b/otherlibs/stdune/src/copyfile_stubs.c
@@ -92,7 +92,7 @@ static ssize_t dune_sendfile(int in, int out, size_t length) {
 static ssize_t dune_copy_file_range(int in, int out, size_t length) {
   ssize_t ret;
   while (length > 0) {
-    ret = copy_file_range(out, NULL, in, NULL, length, 0);
+    ret = copy_file_range(in, NULL, out, NULL, length, 0);
     if (ret < 0) {
       return ret;
     }
--- a/otherlibs/stdune/src/copyfile_stubs.c
+++ b/otherlibs/stdune/src/copyfile_stubs.c
@@ -99,7 +99,7 @@ static ssize_t dune_copy_file_range(int in, int out, size_t length) {
   while (length > 0) {
     ret = copy_file_range_fn(in, NULL, out, NULL, length, 0);
     if (ret < 0) {
-      return ret;
+      return dune_sendfile(in, out, length);
     }
     length = length - ret;
   }
--- a/otherlibs/stdune/src/copyfile_stubs.c
+++ b/otherlibs/stdune/src/copyfile_stubs.c
@@ -67,6 +67,7 @@ CAMLprim value stdune_sendfile(value v_in, value v_out, value v_size) {
 #include <sys/sendfile.h>
 #include <sys/utsname.h>
 #include <linux/version.h>
+#include <stdio.h>
 
 #define FD_val(value) Int_val(value)
 
--- a/otherlibs/stdune/src/copyfile_stubs.c
+++ b/otherlibs/stdune/src/copyfile_stubs.c
@@ -1,3 +1,5 @@
+#define _GNU_SOURCE
+
 #include <caml/fail.h>
 #include <caml/memory.h>
 #include <caml/mlvalues.h>
@@ -63,7 +65,8 @@ CAMLprim value stdune_sendfile(value v_in, value v_out, value v_size) {
 #include <caml/unixsupport.h>
 
 #include <sys/sendfile.h>
-#include <unistd.h>
+#include <sys/utsname.h>
+#include <linux/version.h>
 
 #define FD_val(value) Int_val(value)
 
@@ -73,7 +76,7 @@ CAMLprim value stdune_copyfile(value v_from, value v_to) {
   caml_failwith("copyfile: only on macos");
 }
 
-static int dune_sendfile(int in, int out, size_t length) {
+static ssize_t dune_sendfile(int in, int out, size_t length) {
   ssize_t ret;
   while (length > 0) {
     ret = sendfile(out, in, NULL, length);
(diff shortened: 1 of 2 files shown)


otherlibs/stdune/src/copyfile_stubs.c
```diff
@@ -85,12 +89,43 @@ static int dune_sendfile(int in, int out, size_t length) {
...
+    return -1;
+
+  return KERNEL_VERSION(major, minor, patch);
+}
+
 CAMLprim value stdune_sendfile(value v_in, value v_out, value v_size) {
+  static ssize_t (*dune_copyfile)(int, int, size_t) = NULL;
+  if (dune_copyfile == NULL) {
+    if (kernel_version() < KERNEL_VERSION(5, 19, 0)) {
+      dune_copyfile = &dune_sendfile;
```
Janno: My understanding was that there was no reason to not use `FICLONE` starting from kernel version `4.5`. I don't know how relevant these kinds of systems are but if it is easy enough to add it would be great to include that case.

Janno (commented):
Thanks for implementing this! I can't yet try it out because of unrelated build issues but I will give it a go once those are resolved.

nojb:
Note that judging by https://lwn.net/Articles/846403/, `copy_file_range` can fail on certain files/file systems, so the code should fall back to some other copying mechanism in those cases.

otherlibs/stdune/src/copyfile_stubs.c
nojb: If I understand correctly, `copy_file_range` will do a COW copy if the file system supports it, but otherwise do a normal copy. `FICLONE` on the other hand, will fail if the file system does not support COW, which would require yet another fallback in those cases. Am I missing anything?

Janno: I am not 100% sure about that. Given the things I read about `copy_file_range` and how it changed significantly before 5.19 it is possible that there could be versions between 4.5 and 5.19 where `FICLONE` is strictly better. But I can't point to code to substantiate that.

EDIT: Sorry, I did not really answer your question. Yes, `FICLONE` may fail AFAICT. Given your comment below it seems like so can `copy_file_range` so I am not sure that this property should influence the decision of whether the code should try `FICLONE` on kernels between 4.5 and 5.19.

rgrinberg:
> (but we still require the copy_file_range function to be available at compile-time; we could avoid this by loading the symbol with dlsym).

What about using a compile time check for the availability of this function? This might require updating the bootstrap script, but I could help with that.

nojb:
> > (but we still require the copy_file_range function to be available at compile-time; we could avoid this by loading the symbol with dlsym).
> 
> What about using a compile time check for the availability of this function? This might require updating the bootstrap script, but I could help with that.

Adding this check would allow compiling Dune in old systems that do not have the function, but it would not allow compiling the binary in an old system and have it use the new function it in a newer system (I guess such scenarios are not so uncommon with the advent of binary distributions of Dune?).

Instead, in https://github.com/ocaml/dune/pull/12074/commits/775b4f1e8a57140378267d5670f493fa714457a2, I delay resolving the `copy_file_range` symbol until runtime (using `dlsym`).

Unrelatedly, in https://github.com/ocaml/dune/pull/12074/commits/87fb574fd8113ad51b6faceebee6af0cf6bb2ba2 I introduce a fallback from `copy_file_range()` to `sendfile()` in case the former fails.

nojb:
It would be good to have at least one confirmation that this is working as expected (I cannot test easily myself at the moment). @Janno: would you be able to test and report back here?

rgrinberg:
Checking for things at runtime is fine as well. I'm wondering if it's going to work for statically linked libc's though (e.g. musl). In any case, we don't have to support everything right away

Janno:
> It would be good to have at least one confirmation that this is working as expected (I cannot test easily myself at the moment). @Janno: would you be able to test and report back here?

Thank to @Alizter's help in https://github.com/ocaml/dune/issues/12075 I was able to test this and it does indeed work correctly. Awesome!

If anyone is interested in my test setup: I found https://github.com/pwaller/fienode which can tell if files are reflinks of each other. I created a test project that builds a single binary with lots of `hello world` prints in it. I first built it with the dune version I have installed, used `fclones group` to find copies of `_build/default/foo.exe` in the dune cache, and ran `fienode` on the original file and all copies in the cache and found no sharing. Then I removed one of the `hello world` lines of my code and rebuilt with the branch of this PR and reran the `fclones` and `filenode` steps. This time the copies are actually reflinks.

rgrinberg:
@Janno are you able to observe any performance benefits? I did manage to observe a speed boost when I experimented with this.

rgrinberg (approved):

Janno:
> @Janno are you able to observe any performance benefits? I did manage to observe a speed boost when I experimented with this.

No. In fact, the opposite is the case for me. Compiling `rocq`, for example, consistently yields a ~4% slowdown.

```bash
$ hyperfine -w 1 \
  -p "bash -c 'rm -rf ~/.cache/test1/* && rm -rf _build'" \
  "bash -c 'ulimit -s unlimited && DUNE_CACHE_ROOT=/home/janno/.cache/test1 ~/src/dune/dune.exe b'" \
  "bash -c 'ulimit -s unlimited && DUNE_CACHE_ROOT=/home/janno/.cache/test1 ~/src/dune-copy_file_range/dune.exe b'"
Benchmark 1: bash -c 'ulimit -s unlimited && DUNE_CACHE_ROOT=/home/janno/.cache/test1 ~/src/dune/dune.exe b'
  Time (mean ± σ):     11.271 s ±  0.073 s    [User: 159.119 s, System: 27.324 s]
  Range (min … max):   11.196 s … 11.390 s    10 runs
 
Benchmark 2: bash -c 'ulimit -s unlimited && DUNE_CACHE_ROOT=/home/janno/.cache/test1 ~/src/dune-copy_file_range/dune.exe b'
  Time (mean ± σ):     11.769 s ±  0.093 s    [User: 147.380 s, System: 26.430 s]
  Range (min … max):   11.636 s … 11.942 s    10 runs
 
Summary
  bash -c 'ulimit -s unlimited && DUNE_CACHE_ROOT=/home/janno/.cache/test1 ~/src/dune/dune.exe b' ran
    1.04 ± 0.01 times faster than bash -c 'ulimit -s unlimited && DUNE_CACHE_ROOT=/home/janno/.cache/test1 ~/src/dune-copy_file_range/dune.exe b'
```
For reference, here are the commits used:
```
$ git -C ~/src/dune log | head -n 1
commit 6c8a5a339c7cdbcda6cc70ba0556fc5eec765977 
$ git -C ~/src/dune-copy_file_range/ log | head -n 1
commit 0ffab48c941064672e24041034f6e13af8820aaa
$ git -C ~/src/rocq/ log | head -n 1
commit 81ded19eab08cfba8d5065b189f7ed80657b3d83
```
If anybody wants to repeat the benchmark: `rocq` needs `make dunestrap` before it can be built with dune.

I have rechecked the files in the cache to make sure they are still reflinks (and that they are only reflinks with the branch of this PR) and that part still works fine.

Janno:
I have now tried benchmarking some other projects for which I already had the dependencies installed (all with `dune b --release` in case that matters) and I am seeing mixed but overall positive results: 
1. a 3-4% speedup on `ocaml-re`, 
2. a very slight 0-1% speedup on `memtrace_viewer` (tag `v0.16.0`), 
3. a 1% speedup on `bonsai`,  (tag `v0.16.0`
4. and a 2-3% *slowdown* on `base` (tag `v0.16.0`). 
 
So it seems the performance impact strongly depends on the project that is being built (and possibly many other factors). Personally, I don't think any of these performance changes should matter for getting this MR merged given that it can drastically reduce the disk space used by dune.

nojb:
Thanks for the testing @Janno! Do you have an intuition/explanation for why the changes in this PR could result in a slowdown?

Janno:
No, but I think it's possible my benchmarks are not really measuring the right thing. I have started adding calls to `sync` to the warmup script and also to the end of the command(s) being timed and it has reversed some of the relative measurements as well as generally reduced the magnitude of the speedups/slowdowns. I do not see deltas above 1% +/-1% any more which probably means I am getting close to reading tea leaves. (Of course, `sync` is impacted by whatever else is going on on the machine and I have not actually killed all irrelevant processes before running these benchmarks. The results are still reproducible, though, whatever that might be worth..)

I don't see anything in the change itself that would suggest a regression but I am also not an expert in systems programming.

nojb:
Thanks for the feedback @Janno. Based on what I understand from your comments, the change here has no impact on time, but should provide nice savings on space (on supported filesystems). So it seems that we should go ahead and merge.

@rgrinberg: as you approved some weeks ago, I suspect you agree.

I will wait a day or two and then merge. Thanks everyone for your participation!

otherlibs/stdune/src/copyfile_stubs.c
```diff
@@ -1,3 +1,5 @@
+#define _GNU_SOURCE
```
Alizter: Does it matter that we are doing this before `#elif __linux__`? The mac one seems to define the analogous thing within its if.

nojb: Actually, I believe this `#define` is no longer useful since we load the function dynamically. I removed it. Thanks!

PR #12078, opened 2025-07-26
refactor(boot): remove useless directory name

rgrinberg:
We're already in _boot, so ../_boot is just .


Alizter:
Does this affect test-boot in the Makefile?

rgrinberg:
Yeah, it has a positive effect. That target shouldn't be modifying anything in _boot

PR #12091, opened 2025-07-28
refactor(boot): move duneboot to temp file

rgrinberg:
Another temporary file that we'd be better off generating somewhere where we wouldn't need to remember to clean it.


Alizter:
I would try wrapping the output path in double quotes, that might fix the windows issue. Wasn't an issue before because we had a relative path.

PR #12096, opened 2025-07-29
pkg: enroll odig as a dev-tool

Sudha247:
This PR enrolls `odig` as a dev-tool, to help calling from [vscode-ocaml-platform](https://github.com/ocamllabs/vscode-ocaml-platform), as `vscode-ocaml-platfrom` calls `odig` rather than `odoc`. (cc @pitag-ha @PizieDust)

### DIFF
--- a/src/dune_pkg/dev_tool.ml
+++ b/src/dune_pkg/dev_tool.ml
@@ -33,7 +33,7 @@ let equal a b =
   | Ocamlearlybird, _ -> false
   | _, Ocamlearlybird -> true
   | Odig, Odig -> true
-
+;;
 
 let package_name = function
   | Ocamlformat -> Package_name.of_string "ocamlformat"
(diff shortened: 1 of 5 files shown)


src/dune_pkg/dev_tool.ml
```diff
@@ -28,6 +30,9 @@ let equal a b =
   | Utop, Utop -> true
   | Utop, _ | _, Utop -> false
   | Ocamlearlybird, Ocamlearlybird -> true
+  | Ocamlearlybird, _ -> false
+  | _, Ocamlearlybird -> true
```
gridbugs: I think the second case here should be `-> false` as this pattern matches `Odig, Ocamlearlybird`. Probably best to follow the suit with the other patterns here:
```suggestion
  | Ocamlearlybird, _ | _, Ocamlearlybird -> false
```

gridbugs (approved):

PR #12101, opened 2025-07-30
Bootstrap without temp files

rgrinberg:
Launch bootstrap without an extra compilation step that creates temp files that have to be cleaned up


rgrinberg:
No idea why this fails on OxCaml

```
#=== ERROR while compiling dune.dev ===========================================#
# context     2.4.1 | linux/x86_64 | ocaml-variants.5.2.0+ox | pinned(git+file:///home/runner/work/dune/dune#HEAD#cdc30dd98ee3ce1d73cf16ba97b94fcc8dc88dc8)
# path        ~/work/dune/dune/_opam/.opam-switch/build/dune.dev
# command     ~/work/dune/dune/_opam/bin/ocaml boot/bootstrap.ml -j 3
# exit-code   2
# env-file    ~/.opam/log/dune-3668-344b5e.env
# output-file ~/.opam/log/dune-3668-344b5e.out
### output ###
# ocaml -I +unix unix.cma /tmp/dunebootacbec4main -j 3
# File "_none_", line 1:
# Error: Reference to undefined compilation unit "Stdlib__Mutex"
# Hint: This means that the interface of a module is loaded, but its implementation is not.
#       Did you mean to load a compiled implementation of the module 
#       using "#load" or by passing it as an argument to the toplevel?
```

PR #12102, opened 2025-07-30
feat(melange): promote in-source with `(using melange 1.0)`

anmonteiro:


src/dune_rules/melange/melange_rules.ml
```diff
@@ -211,11 +211,69 @@ let js_targets_of_libs ~sctx ~scope ~module_systems ~target_dir libs =
...
+         let root = Dune_project.root dune_project in
+         let into_dir = Path.Source.append_local root output_dir in
+         Path.reach ~from (Path.Source.append_local into_dir segment |> Path.source)
+       in
+       let loc =
+         p.into
+         |> Option.map ~f:(fun (x : Rule.Promote.Into.t) -> x.loc)
+         |> Option.value ~default:Loc.none
+       in
+       Path.mkdir_p (Path.Source.append_local Path.Source.root output_dir |> Path.source);
```
rgrinberg, line 267: Why do you need this `mkdir_p`? Shouldn't the engine do this for you?

anmonteiro: it doesn't, check out this one where I'm doing the same to get around this limitation: https://github.com/ocaml/dune/pull/12144

anmonteiro: not against the engine doing it for me, I suppose. I just went looking and I guess it should (given [this code](https://github.com/ocaml/dune/blob/main/src/dune_engine/target_promotion.ml#L188)). Will check if there's a bug in there.

anmonteiro: @rgrinberg Here's a new proposal to fix it in the engine: https://github.com/ocaml/dune/pull/12145

It sounds like a poor user experience to ask the user to create a directory manually if they requested for targets to be promoted there. Do you have context here?

Issue #12104, opened 2025-07-30
Cram tests don't pick up env variables


Sudha247:
## Expected Behavior

Cram tests build and succeed. 

## Actual Behavior

Cram tests don't pick up env variables specified. Here's the full log:

<details>
  <summary>Log</summary>

<code>
File "test/blackbox-tests/test-cases/custom-cross-compilation/toolchain.t", line 1, characters 0-0:
------ test/blackbox-tests/test-cases/custom-cross-compilation/toolchain.t
++++++ test/blackbox-tests/test-cases/custom-cross-compilation/toolchain.t.corrected
File "test/blackbox-tests/test-cases/custom-cross-compilation/toolchain.t", line 55, characters 0-1:
 |
 |  $ cat >notocamlc <<EOF
 |  > #!/usr/bin/env sh
 |  > echo "from notocamlc" >&2
 |  > $actualocamlc \$@
 |  > EOF
 |  $ cat >notocamlc-foobar <<EOF
 |  > #!/usr/bin/env sh
 |  > echo "from notocamlc-foobar" >&2
 |  > $actualocamlc \$@
 |  > EOF
 |  $ chmod +x notocamlc notocamlc-foobar
 |
 |We set an undefined toolchain with findlib:
 |
 |  $ OCAMLFIND_TOOLCHAIN=foobar buildfoo
-|  ocamlfind: [WARNING] Undefined toolchain: foobar
-|  ocamlfind: [WARNING] Undefined toolchain: foobar
 |  built foo.cma
 |  $ OCAMLFIND_CONF=$PWD/etc/findlib.conf OCAMLFIND_TOOLCHAIN=foobar buildfoo
-|  from notocamlc-foobar
-|  from notocamlc-foobar
-|  from notocamlc-foobar
-|  from notocamlc-foobar
 |  built foo.cma
 |
 |  $ unset OCAMLFIND_TOOLCHAIN
 |  $ unset OCAMLFIND_CONF
 |
 |Now we set it in the workspace:
 |
 |  $ cat >dune-workspace <<EOF
 |  > (lang dune 1.10)
 |  > (context
 |  >  (default
 |  >   (toolchain foobar)))
 |  > EOF
 |
 |  $ buildfoo
 |  built foo.cma
 |
 |  $ OCAMLFIND_CONF=$PWD/etc/findlib.conf buildfoo
-|  from notocamlc-foobar
-|  from notocamlc-foobar
-|  from notocamlc-foobar
[...]

PR #12105, opened 2025-07-31
Replace `git` function by environment variables

Leonidas-from-XIV:
As #12104 shows, it turns out that exporting variables when a function is called is not POSIX. Thus calls of the form `FOO=bar git` are not portable between systems if `git` is a function.

Fortunately, `git` supports variables to trigger overriding the author/committer, thus instead of creating a wrapper function we're setting these variables.


Sudha247 (approved):
Makes sense to me, and this does fix the failing git test in #12104.

Alizter (approved):

PR #12117, opened 2025-08-02
stdune: define User_message.to_dyn and replace Poly.compare

Alizter:
Due to dependency cycles, this useful function was omitted previously. We introduce it, since it is useful for debugging and testing. Doing so requires appending to the Pp module earlier on, which also allows us to define User_message.compare more correctly, avoiding Poly.compare.

### DIFF
--- a/otherlibs/stdune/src/import.ml
+++ b/otherlibs/stdune/src/import.ml
@@ -1,0 +1,31 @@
+module Pp = struct
+  include Pp
+
+  (** This version of [Pp.compare] uses [Ordering.t] rather than returning an [int]. *)
+  let compare ~compare x y =
+    Ordering.of_int (Pp.compare (fun a b -> Ordering.to_int (compare a b)) x y)
+  ;;
+
+  let to_dyn tag_to_dyn t =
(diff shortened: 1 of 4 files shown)


otherlibs/stdune/src/import.ml
```diff
@@ -0,0 +1,31 @@
+module Pp = struct
+  include Pp
```
rgrinberg, line 2: Can't this live in its own module?

Alizter: It throws ocamldep into a dependency cycle if I introduce another compilation unit called `pp.ml`. If I do something like `pp0.ml`, I have to annotate many other places, and we get many mismatches.

Alizter: Please have a go if you think you can get it to work however, I was not able to.

rgrinberg: I think we introduced `root_module` for situations like this. I suppose we need to add bootstrap support we can use it in dune.

For now, this is OK I suppose.

rgrinberg (approved):

PR #12121, opened 2025-08-03
fix: blake3 build for *bsd and Haiku

### LINKED ISSUE #12087: x86 builds on FreeBSD and OpenBSD failing
There are build failures in FreeBSD and OpenBSD on x86. See here: https://github.com/Alizter/dune/actions/runs/16568972029/job/46855902988?pr=6

It seems the blake3 rules need to be tweaked slightly. The builds on arm appear to be working fine.

Alizter:
We fix the build of Dune on FreeBSD, OpenBSD and NetBSD. This patch has been tested in a custom CI setup here: https://github.com/Alizter/dune/pull/6. I've also added `beos` which will fix the build on Haiku which I've tested manually.

cc @maiste I would appreciate if this could make it into 3.20.0, if not, then 3.20.1 would also be fine.

- fix #12087

### DIFF
--- a/vendor/ocaml-blake3-mini/dune
+++ b/vendor/ocaml-blake3-mini/dune
@@ -34,7 +34,10 @@
    (= %{architecture} "amd64")
    (or
     (= %{system} "linux")
-    (= %{system} "macosx"))))
+    (= %{system} "macosx")
+    (= %{system} "freebsd")
+    (= %{system} "openbsd")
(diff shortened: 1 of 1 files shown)


maiste:
Thanks, @Alizter, I'm not done with the 3.20 so it might be possible to make it in 3.20 if this PR is merged before the middle of next week :+1:

rgrinberg (approved):

Issue #12103, opened 2025-07-30
[pkg] Passing `-j` to `dune build` causes the compiler packgae to be rebuilt


gridbugs:
## Expected Behavior

After installing a project's dependencies, rebuilding the project with `-j` should not cause dependencies to be rebuilt.

## Actual Behavior

Every time the project is rebuilt with a new degree of parallelism, all of its dependencies are rebuilt (including the compiler which takes several minutes to rebuild).

## Reproduction

In any project using dune package management:
1. Build the project
2. Build the project again with `-j1`

All dependencies should be rebuilt. This might be easier to observe if you set `DUNE_CONFIG__PKG_BUILD_PROGRESS=enabled` to make dune print the name of each package that gets built.

gridbugs:
The problem happens because targets of rules whose actions refer to variables are rebuilt when those variables change. In the majority of cases this is desirable because the contents of the target would depend on the value of variables. Passing `-j` to dune has the affect of setting the value of the `${jobs}` variable expanded in package build/install commands, and in the case of the `%{jobs}` variable, when packages do `make -j %{jobs}` or `dune build -j %{jobs}` the value of `%{jobs}` shouldn't meaningfully affect the contents of the contents of the target.

It would be difficult to change dune's behaviour such that changing the value of `%{jobs}` doesn't cause targets to be rebuilt, and I think allowing some variables to change without triggering rebuilds could lead to confusing behaviour, so I don't recommend this approach anyway. An alternative I prefer is to change the behaviour of `-j` so that it doesn't affect how packages are built, and set the `${jobs}` variable to something fixed such as the number of cores on the current machine.

Alizter:
We need to put a different place holder that is evaluated at runtime. That way we can throttle the job number per job.

shonfeder:
> It would be difficult to change dune's behaviour such that changing the value of %{jobs} doesn't cause targets to be rebuilt, and I think allowing some variables to change without triggering rebuilds could lead to confusing behaviour, so I don't recommend this approach anyway

How difficult, do we suppose? E.g., on a complexity scale from 1 to 5, how complex of a change do we think this would be? Could sketch out what (roughly) would be involved?

I would guess that there a few values of this sort, where they should be able to change without actually requiring a rebuild (e.g., https://github.com/ocaml/dune/issues/12131), so perhaps having a general purpose fix for this makes sense?

Alizter:
[...]

PR #12128, opened 2025-08-05
pkg: don't set %{jobs} variable from -j in deps

gridbugs:
When building a project's dependencies, setting the value of %{jobs} from the value of -j means that each time a different value is passed to ${jobs}, each dependency of the project that references the %{jobs} variable in its dependency cone is rebuilt. This includes the compiler which can take several minutes to build. Setting the %{jobs} variable to a fixed value (chosen as the default degree of concurrency used by dune) avoids this issue. Passing a value to -j still affects the degree of concurrency used when building the project itself (as opposed to its dependencies).

Fixes https://github.com/ocaml/dune/issues/12103

### DIFF
--- a/src/dune_config_file/dune_config_file.mli
+++ b/src/dune_config_file/dune_config_file.mli
@@ -135,6 +135,9 @@ module Dune_config : sig
   val hash : t -> int
   val equal : t -> t -> bool
 
+  (** The degree of concurrency dune will use by default. *)
+  val auto_concurrency : int lazy_t
+
   (** [for_scheduler config ?watch_exclusions stats_opt ~signal_watcher]
       creates a configuration for a scheduler from the user-visible Dune
       [config]. *)
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -662,7 +662,14 @@ module Action_expander = struct
       | Build -> Memo.return [ Value.Dir paths.source_dir ]
       | Prefix -> Memo.return [ Value.Dir paths.prefix ]
       | User -> Memo.return [ Value.String (Unix.getlogin ()) ]
-      | Jobs -> Memo.return [ Value.String (Int.to_string !Clflags.concurrency) ]
+      | Jobs ->
+        (* Note that here we intentionally don't set %{jobs} based on dune's
+           "-j" argument. Doing so would mean that each time a different
+           value is passed to "-j", all the dependencies of the project that
+           reference %{jobs} would be rebuilt (including the OCaml compiler
+           itself which takes several minutes). *)
+        Memo.return
+          [ Value.String (Int.to_string (Lazy.force Dune_config.auto_concurrency)) ]
       | Arch -> sys_poll_var (fun { arch; _ } -> arch)
       | Group ->
         let group = Unix.getgid () |> Unix.getgrgid in


Alizter:
Won't this mean when people change their configuration, things will rebuild? I think instead of jobs, we need a runtime placeholder which we can configure as we wish when running the action.

src/dune_config_file/dune_config_file.mli
```diff
@@ -135,6 +135,9 @@ module Dune_config : sig
   val hash : t -> int
   val equal : t -> t -> bool
 
+  (** The degree of concurrency dune will use by default. *)
+  val auto_concurrency : int lazy_t
```
Sudha247, line 139: How is this number arrived at?

Alizter: This mess:
```ocaml
  let auto_concurrency =
    lazy
      (if Sys.win32
       then (
         match Env.get Env.initial "NUMBER_OF_PROCESSORS" with
         | None -> 1
         | Some s -> Int.of_string s |> Option.value ~default:1)
       else (
         let commands =
           [ "nproc", []
           ; "getconf", [ "_NPROCESSORS_ONLN" ]
           ; "getconf", [ "NPROCESSORS_ONLN" ]
           ]
         in
         let rec loop = function
           | [] -> 1
           | (prog, args) :: rest ->
             (match Bin.which ~path:(Env_path.path Env.initial) prog with
              | None -> loop rest
              | Some prog ->
                let prog = Path.to_string prog in
                let fdr, fdw = Unix.pipe () ~cloexec:true in
                (match
                   Spawn.spawn
                     ~prog
                     ~argv:(prog :: args)
                     ~stdin:(Lazy.force Dev_null.in_)
                     ~stdout:fdw
                     ~stderr:(Lazy.force Dev_null.out)
                     ()
                 with
                 | exception Unix.Unix_error _ ->
                   Unix.close fdw;
                   Unix.close fdr;
                   loop commands
                 | pid ->
                   Unix.close fdw;
                   let ic = Unix.in_channel_of_descr fdr in
                   let n =
                     match input_line ic with
                     | line -> String.trim line |> Int.of_string
                     | exception End_of_file -> None
                   in
                   close_in ic;
                   (match n, snd (Unix.waitpid [] pid) with
                    | Some n, WEXITED 0 -> n
                    | _ -> loop rest)))
         in
         loop commands))
  ;;
```

gridbugs:
> Won't this mean when people change their configuration, things will rebuild?

Yes, for example if you replaced your cpu with one with a different number of cores, this will trigger a rebuild. That's not ideal in my opinion but I think preferable to passing a different argument to -j triggering a rebuild, as the latter will be far more common.

>  I think instead of jobs, we need a runtime placeholder which we can configure as we wish when running the action.

I agree in theory, however that would take a lot of work and this problem doesn't seem severe enough to justify it. The solution in this PR could be a bandaid until someone has the time to implement a proper solution like the one you describe.

src/dune_config_file/dune_config_file.mli
gridbugs: @Sudha247 to clarify all I did here was add the already existing lazy value to the interface. I didn't write this code myself.

Alizter:
> > Won't this mean when people change their configuration, things will rebuild?
> 
> Yes, for example if you replaced your cpu with one with a different number of cores, this will trigger a rebuild. That's not ideal in my opinion but I think preferable to passing a different argument to -j triggering a rebuild, as the latter will be far more common.

This will be an issue for cache shared between CI builds, where they might have an explicitly fixed job count. Here the user will no longer have CLI control over how dune runs and we instead do something mysterious based on the architecture.

gridbugs:
Ah fair point. Ok I think for now I'll leave the current behaviour in place then since the cost of fixing this is too high given the benefit.

Alizter:
Alternatively, we can chose a (slow) policy of `-j1` all the time.

gridbugs:
That's true. I don't feel strongly about forcing `-j1` all the time vs keeping the current behaviour.

shonfeder:
Per https://github.com/ocaml/dune/issues/12103#issuecomment-3183220943, it sounds like we should wait to fix the self execing issue blocking a clean fix here. @gridbugs, if you agree, can we close this?

gridbugs:
Agreed, closing.

PR #12129, opened 2025-08-05
tests: Make `dune.lock` a configurable location

Leonidas-from-XIV:
Given that we plan to move the location, this PR makes the location configurable in the helpers script.

### DIFF
--- a/test/blackbox-tests/test-cases/describe_location.t
+++ b/test/blackbox-tests/test-cases/describe_location.t
@@ -25,11 +25,13 @@ An executable from the current project:
   _build/default/foo.exe
 
 Test that executables from dependencies are located correctly:
-  $ mkdir dune.lock
-  $ cat > dune.lock/lock.dune << EOF
+
+  $ default_lock_dir="dune.lock"
+  $ mkdir -p "${default_lock_dir}"
+  $ cat > "${default_lock_dir}"/lock.dune <<EOF
   > (lang package 0.1)
   > EOF
-  $ cat > dune.lock/bar.pkg << EOF
+  $ cat > ${default_lock_dir}/bar.pkg << EOF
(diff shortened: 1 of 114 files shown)


shonfeder (approved):
Thank you for breaking the mechanical changes into this easy to review change! :pray:

Sudha247 (approved):
I chatted offline with @Leonidas-from-XIV a bit about this, and for context this change makes the diff in https://github.com/ocaml/dune/pull/11775 in smaller. I'm happy with merging this as is, with a minor note below.

test/blackbox-tests/test-cases/describe_location.t
```diff
@@ -25,11 +25,13 @@ An executable from the current project:
   _build/default/foo.exe
 
 Test that executables from dependencies are located correctly:
-  $ mkdir dune.lock
-  $ cat > dune.lock/lock.dune << EOF
+
+  $ default_lock_dir="dune.lock"
```
Sudha247, line 29: This is declared in `helpers.sh`. Ideally, it can be used directly from there, but if that's too complicated, it's okay to declare it again here. If this value changes in the future, the test will fail anyway to remind you to update it.

gridbugs (approved):

test/blackbox-tests/test-cases/describe_location.t
Leonidas-from-XIV: Yeah, I tried using the pkg helpers it and after a while I realized that finding a way to use the helpers would make the code strange so just copied the 1 line and called it a day. Given it was just one test it's not that big of a deal IMO to make it worth complicating the rest of the codebase.

rgrinberg:
Alternatively, you could also introduce a shell function to this operation.

PR #12130, opened 2025-08-05
[3.20] prepare 3.20~alpha2 release

maiste:


Alizter:
This probably doesn't need a change log since the blake3 stuff was introduced this version and the build was working before.

PR #12132, opened 2025-08-06
chore: restore maintenance intent

maiste:
This PR restores the maintenance intent. It seems the change in the `dune-project` file has been lost between the PRs about it.
This fix avoids having to modify the file manually in opam.


Leonidas-from-XIV (approved):
Seems reasonable to me :)

Alizter (approved):

PR #12133, opened 2025-08-06
fix: add ocamlc-loc to dune-rpc packages's depends field

Alizter:
@maiste This should fix the issues in the opam CI for 3.20.


maiste (approved):
Thanks! You were faster than me to push the fix 😁

PR #12138, opened 2025-08-06
feat(rpc): add --wait to dune rpc ping

### LINKED ISSUE #12134: allow dune rpc ping to wait for connection
Currently `dune rpc ping` will complain with

```
Error: RPC server not running.
```

when there is no server. Sometimes, we would like to wait for the server to be ready however.

`dune monitor` contains a mechanism whereby it waits for a server connection before starting. We could add an option to `dune rpc ping` that would wait for the server to connect before sending a ping.

Alizter:
fix https://github.com/ocaml/dune/issues/12134

### DIFF
--- a/bin/rpc/rpc_common.mli
+++ b/bin/rpc/rpc_common.mli
@@ -11,7 +11,7 @@ val raise_rpc_error : Dune_rpc.Response.Error.t -> 'a
     fails in any way. Returns an [Error] if the response errors. *)
 val request_exn
   :  Dune_rpc_client.Client.t
-  -> ('a, 'b) Dune_rpc.Decl.Request.witness
+  -> ('a, 'b) Dune_rpc.Decl.request
   -> 'a
   -> ('b, Dune_rpc.Response.Error.t) result Fiber.t
 
@@ -26,7 +26,7 @@ val wait_term : bool Cmdliner.Term.t
 val fire_request
   :  name:string
   -> wait:bool
-  -> ('a, 'b) Dune_rpc.Decl.Request.witness
+  -> ('a, 'b) Dune_rpc.Decl.request
   -> 'a
   -> ('b, Dune_rpc.Response.Error.t) result Fiber.t
 
--- a/bin/rpc/rpc_common.mli
+++ b/bin/rpc/rpc_common.mli
@@ -26,7 +26,7 @@ val wait_term : bool Cmdliner.Term.t
 val fire_request
   :  name:string
   -> wait:bool
-  -> ('a, 'b) Dune_rpc.Decl.request
+  -> ('a, 'b) Dune_rpc.Decl.Request.witness
   -> 'a
   -> ('b, Dune_rpc.Response.Error.t) result Fiber.t
 
--- a/bin/rpc/rpc_ping.ml
+++ b/bin/rpc/rpc_ping.ml
@@ -12,7 +12,7 @@ let term =
   Rpc_common.client_term builder
   @@ fun () ->
   let open Fiber.O in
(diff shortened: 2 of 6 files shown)


Alizter:
@ElectreAAS with this you can use `dune rpc ping --wait` after starting the passive watch mode to make sure it is fully setup. This should help with the test race condition in #12064.

bin/rpc/rpc_common.mli
```diff
@@ -26,7 +26,7 @@ val wait_term : bool Cmdliner.Term.t
 val fire_request
   :  name:string
   -> wait:bool
-  -> ('a, 'b) Dune_rpc.Decl.request
```
shonfeder: Just genuine curiosity and ignorance on my part: Are the additions of the `Dune_rpc.Decl.Request.witness` related to adding `--wait` to the ping command?

bin/rpc/rpc_ping.ml
```diff
@@ -1,33 +1,21 @@
...
 let info =
   let doc = "Ping the build server running in the current directory" in
   Cmd.info "ping" ~doc
 ;;
 
 let term =
-  let+ (builder : Common.Builder.t) = Common.Builder.term in
-  Rpc_common.client_term builder exec
+  let+ (builder : Common.Builder.t) = Common.Builder.term
+  and+ wait = Rpc_common.wait_term in
```
shonfeder, line 14: Adding documentation of the waiting behavior to https://github.com/ocaml/dune/blob/ea255bf36354d2a92b66989f06386dc2eb83528b/bin/rpc/rpc_ping.mli#L1-L2 seems like it could be useful.

Perhaps also on line 5?

shonfeder (approved):
This looks good and will be very helpful. Thanks!

I have one suggestion about documentation and one question, but LGTM.

bin/rpc/rpc_common.mli
rgrinberg: I don't know why it was done this way, but it definitely seems wrong. The witness types aren't supposed to be public outside the implementation of RPC

rgrinberg (changes requested):

bin/rpc/rpc_common.mli
Alizter: I've gone ahead and changed both `fire_request` and `request_exn` (which is unused) to work with requests rather than witnesses.

Alizter: To answer @shonfeder, the ping implementation here was working directly with the witness rather than the request which is why I opted to use witnesses. After inspecting the private RPC implementation I see that the reqests are also exposed so its best we use those instead.

bin/rpc/rpc_ping.ml
Alizter: Done.

Alizter: The `--wait` flag is used elsewhere and will insert its doc accordingly.

```
       --wait
           poll until server starts listening and then establish connection.
```

Alizter: I've gone ahead and added a little bit more documentation to the main description however.

rgrinberg (approved):

shonfeder (approved):

bin/rpc/rpc_ping.ml
```diff
@@ -1,33 +1,28 @@
...
-    ~f:send_ping
-    (Dune_rpc_private.Initialize.Request.create
-       ~id:(Dune_rpc_private.Id.make (Sexp.Atom "ping_cmd")))
-;;
-
 let info =
-  let doc = "Ping the build server running in the current directory" in
+  let doc =
+    "Ping the build server running in the current directory. Passing the --wait flag \
+     allows the command to wait for a connectino to the server."
```
Alizter: Todo fix typo

PR #12142, opened 2025-08-07
test: include_subdirs unqualified duplicates cram rules

Alizter:
This regression was observed in https://github.com/ocaml/opam-repository/pull/28302. It is due to a mix of `(include_subdirs unqualified)` and directory cram tests in the same directory.

I've changed an existing test to highlight two issues with the `(include_subdirs unqualified)` stanza and cram tests.

1. The cram test in the subdirectory is no longer being picked up.
2. The regression above: Directory cram tests are generating multiple rules.


rgrinberg:
Thanks for the repro. I pushed a fix. No idea how any of this worked before.

src/dune_rules/dir_status.mli
```diff
@@ -26,7 +26,8 @@ module Group_root : sig
 end
 
 type t =
-  | Lock_dir
+  | Lock_dir of
+      Source_tree.Dir.t (* XXX Will have to be modified once lock dirs are targets *)
```
Alizter, line 30: cc @Leonidas-from-XIV FYI

Leonidas-from-XIV: Thanks for the shout-out.

It probably actually won't need to be changed because `Dir_status` is only used on the source tree, so either no value of `Lock_dir` will ever be produced if there is no lock directory, or if we keep lock dirs in the source tree (and copy them into `_build` with custom copy rules), then the `Source_tree.Dir.t` is just fine, because it will continue to point at a directory in the source tree.

maiste:
Thanks for the fix! I'll cut the alpha4 release tomorrow morning.

PR #12145, opened 2025-08-10
target_promotion: create `(into ..)` dirs instead of asking user

anmonteiro:


rgrinberg (approved):
I don't know why it was like this. Doesn't seem consistent with the rest of dune though.

PR #12151, opened 2025-08-11
tests: Use `make_lockpkg` to create a package in lock dir

Leonidas-from-XIV:
This variable refers to the lock dir that is meant to be in the users source folder and potentially later be copied to the _build folder. For now it is identical to the default build directory.

### DIFF
--- a/test/blackbox-tests/test-cases/describe_location.t
+++ b/test/blackbox-tests/test-cases/describe_location.t
@@ -26,12 +26,12 @@ An executable from the current project:
 
 Test that executables from dependencies are located correctly:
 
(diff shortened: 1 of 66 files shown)


rgrinberg:
I mentioned this earlier, and I'll re-iterate it here: You should introduce a function to print/create the package for the test. The less lower level constructs we use in our test suite, the easier it will be to adapt it.

Leonidas-from-XIV:
Turns out there was already `make_lockpkg`, it just wasn't used very much. Added some convenience functions to create files and append to them.

shonfeder (approved):
This looks OK to me, given the current testing conventions.

PR #12158, opened 2025-08-12
[3.20] prepare 3.20~alpha4 release

maiste:

Needs #12157 to be merged first.


Alizter:
Probably not worth a change log since it was a regression we haven't released yet. Also Rudi wrote the patch.

PR #12168, opened 2025-08-12
feature: introduce top-closure for ocamllsp

rgrinberg:


rgrinberg:
This is going to break CI, but it has to be done because I don't know how to demonstrate the error otherwise. Will hopefully fix this soon.

PR #12171, opened 2025-08-13
test: internal cyclic module error exposed

Alizter:
Here we demonstrate a cycle error that ends up staying internal causing problems when communicated over RPC. The cram test demonstrates the minimal set up to achieve it and the expect test compares the error with a similar one which gets reported nicely to the user.

I would expect this message to also get reported nicely to the user, but we must be missing it's detection somewhere.

### DIFF
--- a/test/expect-tests/dune_rpc_e2e/dune_rpc_diagnostics.ml
+++ b/test/expect-tests/dune_rpc_e2e/dune_rpc_diagnostics.ml
@@ -7,6 +7,7 @@ module Sub = Dune_rpc.Sub
 module Diagnostic = Dune_rpc.Diagnostic
 module Request = Dune_rpc.Public.Request
 module Response = Dune_rpc.Response
+module Re = Dune_re
 
 let%expect_test "turn on and shutdown" =
   let test () =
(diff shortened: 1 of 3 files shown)


test/expect-tests/dune_rpc_e2e/dune_rpc_diagnostics.ml
```diff
@@ -814,6 +814,93 @@ let g = A.f
...
+        Build ./foo.exe failed |}];
+    let+ () = print_diagnostics poll in
+    [%expect
+      {|
+      [ "Add"
+      ; [ [ "id"; "0" ]
+        ; [ "message"
+          ; [ "Verbatim"
+            ; "Cycle_error.E\n\
+              \  [ (\"Rule.make\", ())\n\
```
rgrinberg: Could you filter out this stack trace from this output?

Alizter: Done.

test/expect-tests/dune_rpc_e2e/dune_rpc_diagnostics.ml
```diff
@@ -814,6 +814,94 @@ let g = A.f
...
+    [%expect
+      {|
+        Building ./foo.exe
+        Build ./foo.exe failed |}];
+    let+ () = print_diagnostics poll in
+    [%expect.output]
+    |> String.split_lines
+    |> List.filter
+         ~f:
+           (let rec aux s len i =
```
rgrinberg: This doesn't seem very readable. How about a regex?

Alizter: Done.

rgrinberg (approved):

Alizter:
@rgrinberg Do you have any ideas for a fix? I didn't take a good look at fixing this yet.

PR #12174, opened 2025-08-13
chore(ci): attempt to fix adding a new package

rgrinberg:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -156,7 +156,9 @@ jobs:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
 
       # Install ocamlfind-secondary and ocaml-secondary-compiler, if needed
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow.yml
```diff
@@ -156,7 +156,9 @@ jobs:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
 
       # Install ocamlfind-secondary and ocaml-secondary-compiler, if needed
-      - run: opam install ./dune.opam --deps-only --with-test
+      - run: |
+          opam pin -n . --with-version=3.20
+          opam install ./dune.opam --deps-only --with-test
```
Alizter, line 161: Are you sure this isn't meant to happen later on during `Install deps`? Here we are only building the `dune` package.

rgrinberg: I'm not sure, but I would imagine it's harmless to do this pin before.

PR #12175, opened 2025-08-13
chore(ci): remove --with-test from installing dune

rgrinberg:
If the goal is to check that dune is building, we don't need its test deps.

Dune has no test-deps, so this is useless anyway


Alizter:
I believe this was only relevant when we had the secondary compiler around.

rgrinberg:
We still have the secondary compiler though. We just don't test in CI. Not sure what difference does it make here as it's not a test dep.

PR #12176, opened 2025-08-13
feature: introduce fs-io

rgrinberg:
Extract out a couple of simple IO functions to reuse in ocamllsp without pulling in all of stdune.


otherlibs/fs-io/fs_io.ml
```diff
@@ -0,0 +1,138 @@
...
+let rec eagerly_input_acc ic s ~pos ~len acc =
+  if len <= 0
+  then acc
+  else (
+    let r = input ic s pos len in
+    if r = 0 then acc else eagerly_input_acc ic s ~pos:(pos + r) ~len:(len - r) (acc + r))
+;;
+
+(* [eagerly_input_string ic len] tries to read [len] chars from the channel.
+     Unlike [really_input_string], if the file ends before [len] characters are
```
Alizter: Can you rewrap the comments in this file? They have become a bit too indented.

maiste:
I imagine this is equivalent to internal libraries like `dyn` in term maintenance expectation? Should we add the maintenance field and point it to `none`?

rgrinberg:
Sure, that makes sense. I don't want to impose extra maintenance work on the team.

PR #12183, opened 2025-08-13
ci: check bootstrap script in 4.02

### LINKED ISSUE #12178: CI job for boot/bootstrap.ml on 4.02
The script in `boot/bootstrap.ml` should run on 4.02. It would be nice to have a CI job that verifies this. There's no need to actually build dune here, just make sure the script compiles with 4.02

Alizter:
We can use `ocamlc -i` (infer interface) to interpret the script without running or generating artefacts. It will output any errors we get.

fix https://github.com/ocaml/dune/issues/12178

The CI job doesn't use the `make` target because we don't (need to) have `make` installed.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -82,6 +82,23 @@ jobs:
         env:
           LC_ALL: C
 
+  bootstrap:
+    name: Bootstrap
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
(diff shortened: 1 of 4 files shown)


rgrinberg:
Can you just make the job use make? Otherwise we have to keep the two commands in sync which somebody will invariably forget to do.

rgrinberg (approved):

Issue #12156, opened 2025-08-12
Docs: please make all field descriptions anchor links


yawaramin:
Eg I want to link to the `(optional)` field in the `library` reference documentation: https://dune.readthedocs.io/en/stable/reference/dune/library.html

The docs should automatically make all these fields anchor links so eg https://dune.readthedocs.io/en/stable/reference/dune/library.html#describe-optional

PR #12186, opened 2025-08-14
doc: replace describe by confval

maiste:
This PR tries to address some parts of #12156.

`confval` generates anchors when `describe` does not. It might generate warnings due to references existing twice. It doesn't affect our use case, but, it could be a problem at some point if we directly use the `confval` in our refs.

### DIFF
--- a/doc/reference/config/project_defaults.rst
+++ b/doc/reference/config/project_defaults.rst
@@ -14,7 +14,7 @@ of the :doc:`../dune-project/index` file when initializing a project with
 
 ``<optional-fields>`` are:
 
-.. describe:: (authors <string(s)>)
+.. confval:: (authors <string(s)>)
 
    Specify authors.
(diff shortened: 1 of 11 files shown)


Alizter:
I am not a fan of these reference names colliding. Would it be possible to introduce our own directive similar to describe and confval where we can make the reference value a bit more qualified so we don't get these conflicts?

maiste:
That's a good question. It may be possible but my understanding is it would require to code it in Python. I'm not enough aware of Sphinx possibilities to provide a confident answer. The PR was mostly to open the question. 

If we can extend it without having to write a full library, it would be clean!

_EDIT_: I converted the issue to DRAFT to make it clear not to merge it and give time to investigate.

shonfeder:
Feel free to open again if this work should be revived!

PR #12188, opened 2025-08-14
feat(pkg): cache revision store

Alizter:
We introduce a caching mechanism for the revision store. This cuts solving time by around 50% on average.

It works by using LMDB to cache both the ls-tree tree operations and the file contents of an object. These lookups happen frequently when running the solver, so any time we can save here is quite noticable.

For instance, a locking that takes around 800ms now takes around 400ms when cached.

LMDB may also be used to cache other lookups in dune in the future, such as the parsing of dune files.

- [x] fork lmdb stubs
- [x] add vendor script
- [x] patch lmdb stubs for MSVC
  - proposed patch https://github.com/Drup/ocaml-lmdb/pull/44
  - probably going to use another one as discussed there
  - [x] update fork with new patch
- [x] #12191 
- [x] #12195 
- [ ] check what happens when two dune's hold the lmdb lock
- [x] add a config option to disable this feature
- [ ] add tests
  - [x] #12231 
  - [x] #12256 
  - [ ] split test cleanups 
- [ ] add versioning
  - Left for a future PR, cache is disabled for now to avoid invalid fetches
- [ ] more benchmarks

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -1616,6 +1616,8 @@ module Library = struct
         ; "-DBLAKE3_NO_AVX2"
         ; "-DBLAKE3_NO_AVX512"
         ]
+      else if String.equal "lmdb_stubs.c" fn
+      then [ "-I ." ]
       else []
     in
     { Source.flags = extra_flags @ c.flags; name = fn }
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -3,6 +3,7 @@ open Dune_vcs
 module Process = Dune_engine.Process
 module Display = Dune_engine.Display
 module Scheduler = Dune_engine.Scheduler
+module Console = Dune_console
 open Fiber.O
 
 module Object : sig
@@ -13,6 +14,7 @@ module Object : sig
   val compare : t -> t -> Ordering.t
   val hash : t -> int
   val to_hex : t -> string
+  val of_sha1_unsafe : string -> t
 
   type resolved = t
 
@@ -23,8 +25,9 @@ end = struct
   let compare (Sha1 x) (Sha1 y) = String.compare x y
   let to_hex (Sha1 s) = s
   let equal (Sha1 x) (Sha1 y) = String.equal x y
-  let to_dyn (Sha1 s) = Dyn.variant "Sha1" [ Dyn.string s ]
+  let to_dyn (Sha1 s) = Dyn.variant "Sha1" [ Dyn.opaque s ]
   let hash (Sha1 s) = String.hash s
+  let of_sha1_unsafe s = Sha1 s
 
   type resolved = t
 
@@ -125,6 +128,174 @@ module File = struct
 
   module C = Comparable.Make (T)
   module Set = C.Set
+end
+
+module Cache = struct
+  (* CR-someday Alizter: Various LMDB  operations are able to raise [Map_full]
+     when the database is full. We should handle this in a sensible way. For
+     now, we allow it to raise. It won't be visible to most users due to the
+     cache size we have chosen. *)
+
+  let rev_store_cache =
+    (* CR-soon Alizter: For now the cache is disabled by default. Once we add
+       versioning to the cache it will be safe to enable by default. 
+
+       When we do this we should check [Int.equal Sys.word_size 64] since
+       32-bit platforms won't handle our cache size well. *)
+    Dune_config.Config.make_toggle ~name:"rev_store_cache" ~default:`Disabled
+  ;;
+
+  let cache_dir =
+    lazy
+      (let path =
+         Path.L.relative
+           (Lazy.force Dune_util.xdg
+            |> Xdg.cache_dir
+            |> Path.Outside_build_dir.of_string
+            |> Path.outside_build_dir)
+           [ "dune"; "rev_store" ]
+       in
+       let rev_store_cache = Dune_config.Config.get rev_store_cache in
+       Dune_util.Log.info
+         [ Pp.textf
+             "Revision store cache: %s"
+             (Dune_config.Config.Toggle.to_string rev_store_cache)
+         ];
+       match rev_store_cache, Path.mkdir_p path with
+       | `Enabled, () ->
+         Dune_util.Log.info
+           [ Pp.textf "Revision store cache location: %s" (Path.to_string path) ];
+         Some path
+       | `Disabled, () -> None)
+  ;;
+
+  let db =
+    lazy
+      (Lazy.force cache_dir
+       |> Option.map ~f:(fun path ->
+         Lmdb.Env.create
+           ~map_size:(Int64.to_int 5_000_000_000L) (* 5 GB *)
+           ~max_maps:2
+           ~flags:Lmdb.Env.Flags.(no_meta_sync)
+           Rw
+           (Path.to_string path)))
+  ;;
+
+  let () =
+    at_exit (fun () ->
+      if Lazy.is_val db
+      then (
+        match Lazy.force db with
+        | Some db -> Lmdb.Env.close db
+        | None -> ()))
+  ;;
+
+  module Key = struct
+    module T = struct
+      type t = Object.t
+
+      let compare = Object.compare
+      let to_dyn = Object.to_dyn
+    end
+
+    include T
+    module C = Comparable.Make (T)
+    module Map = C.Map
+    module Set = C.Set
+
+    let conv =
+      Lmdb.Conv.make
+        ~serialise:(fun alloc obj ->
+          Object.to_hex obj |> Lmdb.Conv.(serialise string alloc))
+        ~deserialise:(fun bs ->
+          Lmdb.Conv.(deserialise string bs) |> Object.of_sha1_unsafe)
+        ()
+    ;;
+  end
+
+  let map =
+    lazy
+      (Lazy.force db
+       |> Option.map ~f:(fun env ->
+         Lmdb.Map.create Nodup ~key:Key.conv ~value:Lmdb.Conv.string ~name:"objects" env)
+      )
+  ;;
+
+  module Files_and_submodules = struct
+    module Key = struct
+      module T = struct
+        type t = Object.t
+
+        let compare = Object.compare
+        let to_dyn = Object.to_dyn
+      end
+
+      include T
+      module C = Comparable.Make (T)
+
+      let conv =
+        Lmdb.Conv.make
+          ~serialise:(fun alloc obj ->
+            Object.to_hex obj |> Lmdb.Conv.(serialise string alloc))
+          ~deserialise:(fun bs ->
+            Lmdb.Conv.(deserialise string bs) |> Object.of_sha1_unsafe)
+          ()
+      ;;
+    end
+
+    module Value = struct
+      let conv : (File.Set.t * Commit.Set.t) Lmdb.Conv.t =
+        Lmdb.Conv.make
+          ~serialise:(fun alloc v ->
+            Marshal.to_string v [] |> Lmdb.Conv.(serialise string alloc))
+          ~deserialise:(fun bs -> Marshal.from_string Lmdb.Conv.(deserialise string bs) 0)
+          ()
+      ;;
+    end
+
+    let map =
+      lazy
+        (Lazy.force db
+         |> Option.map ~f:(fun env ->
+           Lmdb.Map.create Nodup ~key:Key.conv ~value:Value.conv ~name:"ls-tree" env))
+    ;;
+
+    let get key =
+      let open Option.O in
+      let* m = Lazy.force map in
+      match Lmdb.Map.get m key with
+      | exception Not_found -> None
+      | v -> Some v
+    ;;
+
+    let set key value =
+      ignore
+      @@
+      let open Option.O in
+      let+ map = Lazy.force map in
+      Lmdb.Map.set map key value
+    ;;
+  end
+
+  let get keys =
+    match Lazy.force map with
+    | None -> Key.Map.empty
+    | Some m ->
+      Key.Set.fold keys ~init:Key.Map.empty ~f:(fun key acc ->
+        match Lmdb.Map.get m key with
+        | exception Not_found -> acc
+        | v -> Key.Map.add_exn acc key v)
+  ;;
+
+  let set keys =
+    ignore
+    @@
+    let open Option.O in
+    let* map = Lazy.force map in
+    let* env = Lazy.force db in
+    Lmdb.Txn.go Rw env (fun txn ->
+      Key.Map.iteri keys ~f:(fun key value -> Lmdb.Map.set ~txn map key value))
+  ;;
 end
 
 module Remote = struct
@@ -316,7 +487,7 @@ let run_with_exit_code { dir; _ } ~allow_codes ~display args =
         ]
         ~hints:[ User_message.command "Please update your git version." ]
     | _ ->
-      Dune_console.print [ Pp.verbatim stderr ];
+      Console.print [ Pp.verbatim stderr ];
       Error { Git_error.dir; args; exit_code; output = [] })
 ;;
 
@@ -575,6 +746,24 @@ let fetch repo ~url obj =
     User_error.raise [ Pp.textf "unable to fetch %S from %S" (Object.to_hex obj) url ]
 ;;
 
+module Debug = struct
+  let files_and_submodules_cache = ref false
+  let content_of_files_cache = ref false
+
+  type t =
+    { name : string
+    ; payload : (string * Dyn.t) list
+    }
+
+  let to_dyn { name; payload } =
+    Dyn.Tuple [ Dyn.string name; Dyn.list (Dyn.pair Dyn.string Fun.id) payload ]
+  ;;
+
+  let print : string -> (string * Dyn.t) list -> unit =
+    fun name payload -> Console.print [ to_dyn { name; payload } |> Dyn.pp ]
+  ;;
+end
+
 module At_rev = struct
   type repo = t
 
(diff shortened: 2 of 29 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -1,19 +1,24 @@
 open Stdune
 open Dune_vcs
-module Process = Dune_engine.Process
-module Display = Dune_engine.Display
-module Scheduler = Dune_engine.Scheduler
+
+include struct
```
rgrinberg: Can such a cosmetic change be moved to a separate PR?

src/dune_pkg/rev_store.ml
```diff
@@ -1,19 +1,24 @@
...
+  module Display = Display
+  module Scheduler = Scheduler
+end
+
 module Re = Dune_re
 open Fiber.O
 
 module Object = struct
-  type t = Sha1 of string
+  type t = string
```
rgrinberg: Could you revert this?

src/dune_pkg/rev_store.ml
```diff
@@ -353,76 +633,6 @@ let load_or_create ~dir =
...
-  ;;
-
-  let rec hash = function
-    | Direct t -> t.hash
-    | Redirect t -> hash t.to_
-  ;;
-
-  module C = Comparable.Make (T)
-  module Set = C.Set
-end
```
rgrinberg: I think this is just a code move. Can it be done in a separate PR to simplify the review here?

src/dune_pkg/rev_store.ml
```diff
@@ -23,11 +28,286 @@ module Object = struct
...
+           [ "dune"; "rev_store" ]
+       in
+       match Path.mkdir_p path with
+       | exception _ -> Error ()
+       | () -> Ok path)
+  ;;
+
+  (* Some operations on the cache may end up filling up the map. This helper
+     function allows us to give a more helpful error message in those cases. *)
+  let protect_map_full f =
```
rgrinberg: Instead of failing, this function could try resizing the database.

It should be noted that we basically never hit this function under regular use. The initial size should be large enough that we should never run out of space.

src/dune_pkg/rev_store.ml
```diff
@@ -23,11 +28,286 @@ module Object = struct
...
+  ;;
+
+  let db =
+    lazy
+      (protect_map_full
+       @@ fun () ->
+       Lazy.force cache_dir
+       |> Result.map ~f:(fun path ->
+         Lmdb.Env.create
+           ~map_size:500_000_000 (* 500 MB *)
```
rgrinberg: Seems way too small. Let's add another 0 here at least.

src/dune_pkg/rev_store.ml
Alizter: Done in #12188

src/dune_pkg/rev_store.ml
```diff
@@ -110,6 +112,223 @@ module File = struct
...
+
+    include T
+    module C = Comparable.Make (T)
+    module Map = C.Map
+    module Set = C.Set
+
+    let obj_size = 40
+
+    let conv =
+      Lmdb.Conv.make
```
madroach, line 207: Instead of rolling your own bigstring<->string conversion, you could as well just call to `Lmdb.Conv.string`:
``` ocaml
~serialise:(fun alloc obj -> Object.to_hex obj |> Lmdb.Conv.(serialise string alloc)
~deserialise:(fun bs -> Lmdb.Conv.(deserialise string bs) |> Object.of_sha1_unsafe
```

src/dune_pkg/rev_store.ml
```diff
@@ -110,6 +112,223 @@ module File = struct
...
+
+    let get key =
+      match Lazy.force map with
+      | Error () -> None
+      | Ok m ->
+        (match Lazy.force db with
+         | Error () -> None
+         | Ok env ->
+           Lmdb.Txn.go Ro env (fun txn ->
+             match Lmdb.Map.get ~txn m key with
```
madroach: No need to start a custom `txn` here. Just call `Lmdb.Map.get` without it.

src/dune_pkg/rev_store.ml
```diff
@@ -110,6 +112,223 @@ module File = struct
...
+    let set key value =
+      match Lazy.force map with
+      | Error () -> ()
+      | Ok map ->
+        (match Lazy.force db with
+         | Error () -> ()
+         | Ok env ->
+           protect_map_full
+           @@ fun () ->
+           (match Lmdb.Txn.go Rw env (fun txn -> Lmdb.Map.set ~txn map key value) with
```
madroach: Same as above. No need to use a custom `txn`. `Lmdb.Map.set` is the only operation in the transaction. No need to make it atomic. It already is atomic.

src/dune_pkg/rev_store.ml
```diff
@@ -110,6 +112,223 @@ module File = struct
...
+  end
+
+  let get keys =
+    match Lazy.force map with
+    | Error () -> Key.Map.empty
+    | Ok m ->
+      (match Lazy.force db with
+       | Error () -> Key.Map.empty
+       | Ok env ->
+         Lmdb.Txn.go Ro env (fun txn ->
```
madroach: Not sure whether this transaction is necessary. Does the fold below need a consistent view of the map? If yes, then the transaction is needed.

madroach (changes requested):

src/dune_pkg/rev_store.ml
Alizter: We set and then get, so it seems fine to do these gets without batching. I don't see another way we could update the database while in the middle of getting many.

src/dune_pkg/rev_store.ml
Alizter: Thanks.

rgrinberg:
Thanks for taking over this Ali. This is going to be a very useful QoL improvements to users.  You've somewhat undersold what the cache does in your description though. The most important feature of this cache is that it is shared between different workspaces. So it means that populating it once anywhere on the system speeds up solving for all other projects - even if they aren't using the same version of the opam repository.

> check what happens when two dune's hold the lmdb lock

That's impossible

Alizter:
> > check what happens when two dune's hold the lmdb lock
> 
> That's impossible

Right, I should of said "try to hold". I just wanted to check the error message we give to users in that case.

rgrinberg:
I don't think there will be an error message in that case. One dune instance should just wait for the other one to finish writing.

Alizter:
@rgrinberg Would you prefer a private or user-facing config for the cache?

rgrinberg:
Private is enough. I only imagine this option to be useful to get around bugs. The vast majority of users shouldn't even be aware of this

src/dune_pkg/rev_store.ml
madroach: But remember lmdb is limited by address space, which on a 32bit architecture limits the db size to way less than 4GiB.

src/dune_pkg/rev_store.ml
```diff
@@ -127,6 +129,184 @@ module File = struct
...
+          ~deserialise:(fun bs ->
+            Lmdb.Conv.(deserialise string bs) |> Object.of_sha1_unsafe)
+          ()
+      ;;
+    end
+
+    module Value = struct
+      let conv : (File.Set.t * Commit.Set.t) Lmdb.Conv.t =
+        Lmdb.Conv.make
+          ~serialise:(fun alloc v ->
```
madroach, line 249:
``` ocaml
~serialise:(fun alloc obj -> Marshal.to_string v |> Lmdb.Conv.(serialise string alloc))
~deserialise:(fun bs -> Marshal.from_string Lmdb.Conv.(deserialise string bs) 0)
```

madroach (changes requested):

src/dune_pkg/rev_store.ml
Alizter: @madroach do you know if lmdb has a way to automatically resize?

src/dune_pkg/rev_store.ml
Alizter: Thanks

src/dune_pkg/rev_store.ml
madroach: No, it doesn't. And if you resize manually, you have to make sure there are no active transactions. How big do you expect the database to grow?

Alizter: I think 500M is a good starting size. My own testing reveals that some large solves generate about 10M of cache. I would imagine that most of this would continue to be reused so the growth would probably have a hard time hitting 100M after heavy use for a few months.

If we do reach 500M, I will attempt to resize.

rgrinberg: Let's do the following:

1. Set this to 5G
2. Disable the cache entirely on 32bit

No need to worry about resizing anything.

Alizter: I've added detection for 32 bit architectures and used it to disable the cache in that case. I've also bumped the size to 5G.

Alizter: I've replaced 5G with `Int.max_int` which will compile to the correct upper limit on both architectures. I will remove the architecture detection I added.

rgrinberg: You don't have enough address space for that on 64bit. Just leave it as 5G (or even 10G) there.

rgrinberg: Also, you should still disable the feature on 32bit. Running into the max db size is going to be very likely there given the low limit.

Alizter: I can't compile 5G as an integer on 32bit. That is my issue here.

Alizter: I've done (Int64.to_int 5_000_000_000L) which will compile regardless of arch.

src/dune_pkg/rev_store.ml
madroach: Since this is just a cache, what about dropping the map contents with ``Lmdb.Map.drop`` on ``Map_full``? Then a smaller db size would suffice, too. The named map's names are the keys in the unnamed map, in case you want to iterate over them with ``Map.drop``.

src/dune_pkg/rev_store.ml
```diff
@@ -127,6 +129,184 @@ module File = struct
...
+        | Error () ->
+          Code_error.raise "Map_full raised even though cache doesn't exist" []
+      in
+      User_error.raise
+        [ Pp.text "Revision store cache is full."; Pp.text "Please delete:"; Path.pp dir ]
+  ;;
+
+  let db =
+    lazy
+      (protect_map_full
```
madroach: ``Env.create`` won't raise ``Map_full``, therefore you can remove ``protect_map_full`` here and define it later with access to the environment. In protect_map_full, you could then drop the map contents:
``` ocaml
open Lmdb in
Txn.go `Write env @ fun txn ->
let map0 = Map.(open_existing Nodup ~txn ~key:string ~value:string ?name:None env
Cursor.go `Read ~txn map0 @ fun cursor ->
let f name _ =
  Map.(open_existing Nodup ~txn ~key:string ~value:string ~name env in
  |> Map.drop ~txn
in
Cursor.iter ~cursor ~f map0
```

madroach (changes requested):
Could the ``Map_full`` exception be handled more gracefully by just dropping the cache contents? Would this violate invariants in dune? Then the sizing question would also be less difficult to answer. A smaller environment size would only lead to more frequent cache flushes.

otherlibs/stdune/src/platform.mli
```diff
@@ -19,3 +19,5 @@ end
 
 (** [assert_os os] assert that we're running on [os]. *)
 val assert_os : OS.t -> unit
+
+val is_32bit : unit -> bool
```
rgrinberg: What about just using `Sys.word_size`?

rgrinberg:
I suggest not to focus so much on the size issue and instead focus on more pressing issues. 5GB is big enough that it should take users quite a bit of time to approach this limit. Until they get there, we still have plenty of time to address it.

Some other issues that concern me more:

* Tests. I would like to see some unit tests for this feature. Perhaps add a trace mode to Rev_store so that you can observe what it's doing under the hood while being tested?

* Versioning. Whenever we upgrade the schema, we need to make sure we don't try to read from the old db. One suggestion to do this is to include the version number of the schema in the filename of the DB.

boot/duneboot.ml
```diff
@@ -1078,6 +1078,8 @@ module Library = struct
                   ; "-DBLAKE3_NO_AVX512"
                   ]
                 | _ -> [])
+              else if String.equal "lmdb_stubs.c" fn
+              then [ "-I ." ]
```
rgrinberg: How come we need this?

Alizter: Since we are vendoring lmdb next to the sources of the bindings, we need to add the current directory to the include path. Otherwise we have to change an include from angle brackets to double quotes. It is also done in the stanza, so needs to be manually added to the bootstrap.

Our vendor version therefore doesn't need to track a fork but upstream directly.

otherlibs/stdune/src/platform.mli
Alizter: Good idea.

Alizter:
> Would this violate invariants in dune? Then the sizing question would also be less difficult to answer. A smaller environment size would only lead to more frequent cache flushes.

There is nothing that would be violated in Dune. The cache is always checked first in this implementation and if there is no hit for a key, we look up the the contents the usual way, after which they get added to the cache.

I think as Rudi says, its not worth focusing on the cache filling up at the moment.

boot/duneboot.ml
rgrinberg: OK, I see. We have the same hack in the dune file as well.

madroach:
Also not a priority, but you might want to put a [`CACHEDIR.TAG`](https://bford.info/cachedir/) into dune's cache directory.

Alizter:
@madroach We had an issue about this before: https://github.com/ocaml/dune/issues/9054. We closed it because it wasn't so widely adopted at the time. We choose cache directories from `xdg` anyway, so any tools aware of that should work fine.

src/dune_pkg/rev_store.ml
```diff
@@ -127,24 +129,219 @@ module File = struct
...
+       | `Disabled, () -> Error ()
+       | exception _ -> Error ())
+  ;;
+
+  (* Some operations on the cache may end up filling up the map. This helper
+     function allows us to give a more helpful error message in those cases. *)
+  let protect_map_full f =
+    match f () with
+    | x -> x
+    | exception Lmdb.Mdb.Map_full ->
```
rgrinberg: Can you remove all of this size checking stuff for now? It's not necessary to accept this PR, and it requires a bit more thinking. Let's do it later.

test/expect-tests/dune_pkg/rev_store_tests.ml
```diff
@@ -80,3 +94,144 @@ let%expect_test "adding remotes" =
...
+    Fiber.return (write dir file contents)
+    >>> git ~dir [ "add"; file ]
+    >>> git ~dir [ "commit"; sprintf "-m '%s'" commit_msg ]
+  in
+  let get_file remote_revision file =
+    Rev_store.Object.of_sha1 remote_revision
+    |> Option.value_exn
+    |> Rev_store.fetch_object rev_store remote
+    >>| Option.value_exn
+    >>= fun at_rev ->
```
rgrinberg: `let* at_rev = ...`

test/expect-tests/dune_pkg/rev_store_tests.ml
```diff
@@ -37,6 +48,12 @@ let git_out ~dir =
   fun args -> Process.run_capture_line ~dir ~display ~stderr_to failure_mode git args
 ;;
 
+let ls dir =
+  let stderr_to = make_stdout () in
+  let ls = Bin.which ~path:(Env.initial |> Env_path.path) "ls" |> Option.value_exn in
```
rgrinberg: Why do we need this? Why can't we just call `Sys.readdir` for example?

src/dune_rules/pkg_toolchain.mli
```diff
@@ -21,7 +21,7 @@ val is_compiler_and_toolchains_enabled : Package.Name.t -> bool
 
 (** Returns the path to the directory containing the given package within the
     toolchain directory. This will be something like
-    $XDG_CACHE_DIR/dune/toolchains/ocaml-base-compiler.5.2.1.XXXXXXXX where
+    $XDG_CACHE_HOME/dune/toolchains/ocaml-base-compiler.5.2.1.XXXXXXXX where
```
rgrinberg, line 24: Seems unrelated?

src/dune_pkg/rev_store.ml
```diff
@@ -127,24 +129,219 @@ module File = struct
...
+           (Lazy.force Dune_util.xdg
+            |> Xdg.cache_dir
+            |> Path.Outside_build_dir.of_string
+            |> Path.outside_build_dir)
+           [ "dune"; "rev_store" ]
+       in
+       match Dune_config.Config.get rev_store_cache, Path.mkdir_p path with
+       | `Enabled, () -> Ok path
+       | `Disabled, () -> Error ()
+       | exception _ -> Error ())
```
rgrinberg: Can we not catch code errors here? Also, the failure should be logged.

src/dune_pkg/rev_store.ml
```diff
@@ -967,6 +1191,37 @@ let content_of_files t files =
...
+    List.filter_map keys ~f:(fun (key, file) ->
+      if Cache.Key.Map.mem cached key then None else Some (key, file))
+  in
+  if !Debug.content_of_files_cache
+  then (
+    match Cache.Key.Map.cardinal cached with
+    | 0 -> Dune_console.printf "contents_of_files cache missed"
+    | no_of_hits ->
+      Dune_console.printf
+        "content_of_files (%d/%d) cache hits"
```
rgrinberg: I think that instead of nice messages like this, everything that you print should be of the form:

```
type debug = {
  name : string (* this stays constant *)
  payload : (string * Dyn.t) list
}
```

All of the additional formatting mostly gets in the way when debugging.

rgrinberg (changes requested):

src/dune_rules/pkg_toolchain.mli
Alizter: I will cherry pick this out.

src/dune_pkg/rev_store.ml
Alizter: Removed and left a CR.

src/dune_pkg/rev_store.ml
Alizter: Do you want me to still catch the unix error?

vendor/update-ocaml-lmdb.sh
```diff
@@ -0,0 +1,53 @@
...
+
+cp -v $LMDB_SRC/lmdb.h ocaml-lmdb/
+cp -v $LMDB_SRC/mdb.c ocaml-lmdb/
+cp -v $LMDB_SRC/midl.{c,h} ocaml-lmdb/
+cp -v $LMDB_SRC/COPYRIGHT ocaml-lmdb/
+cp -v $LMDB_SRC/LICENSE ocaml-lmdb/LICENSE-lmdb
+
+# Apply Cygwin patch to mdb.c
+echo "Applying Cygwin compatibility patch..."
+sed -i.bak '/^#ifndef MDB_USE_ROBUST/i\
```
rgrinberg, line 44: Why do we need the backup files?

rgrinberg (approved):

vendor/update-ocaml-lmdb.sh
Alizter: I seem to recall there are issues removing the `-i` on macos, can you confirm it works without?

rgrinberg: Can you get rid of the backup files after at least? Otherwise, they're still going to pollute things for the Linux users.

rgrinberg: Oh I see you've already done that.

rgrinberg: Hopefully your new package management thing will replace all of these sed uses anyway.

rgrinberg (approved):

PR #12191, opened 2025-08-15
refactor(rev_store): enforce typing of objects

Alizter:
We enforce the typing of `Object.t` which was a bit loose before and move some things around in preperation for #12188.

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -40,6 +40,76 @@ end = struct
   ;;
 end
 
+module Commit = struct
+  module T = struct
+    type t =
+      { path : Path.Local.t
+      ; rev : Object.t
+      }
+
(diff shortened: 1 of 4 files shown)


src/dune_pkg/rev_store.ml
```diff
@@ -6,7 +6,19 @@ module Scheduler = Dune_engine.Scheduler
...
 
-module Object = struct
+module Object : sig
+  type t
+
+  val to_dyn : t -> Dyn.t
+  val equal : t -> t -> bool
+  val compare : t -> t -> Ordering.t
+  val hash : t -> int
+  val to_string : t -> string
```
rgrinberg: How about `to_hex`? `to_string` could also mean the binary representation.

rgrinberg (approved):

PR #12194, opened 2025-08-15
Change lock dir path type to `Path.t`

Leonidas-from-XIV:
Given we plan to move the lock dirs into the build directory, we can't continue to use `Path.Source.t`. We can either move them to `Path.Build.t` or `Path.t`.

The former has the advantage that it restricts the system from writing the lock dir to any place that isn't the build dir, however it requires a lot more changes that can't meaningfully be unbundled from #11775.

This PR is more conservative and only moves it to `Path.t`. It still relies on the fact that the `Path.t` is not an `Path.Build.t` but most of them are backed by a `Path.Source.t`. It also adds a bit of lazy-loading as `Path.of_string` requires `Fgetcl` to be initialized so the parsing can't happen at CLI parse time, but the complication is quite marginal.

### DIFF
--- a/bin/pkg/pkg_common.ml
+++ b/bin/pkg/pkg_common.ml
@@ -167,7 +167,7 @@ let pp_packages packages = Pp.enumerate packages ~f:pp_package
 module Lock_dirs_arg = struct
   type t =
     | All
-    | Selected of Path.Source.t list
+    | Selected of Path.t Lazy.t list
 
   let all = All
 
@@ -183,7 +183,7 @@ module Lock_dirs_arg = struct
                ~doc:
                  "Lock directories to check for outdated packages. Defaults to dune.lock.")
        in
-       Selected (List.map arg ~f:Path.Source.of_string))
+       Selected (List.map arg ~f:(fun s -> lazy (Path.of_string s))))
       (let+ _all =
          Arg.(
            value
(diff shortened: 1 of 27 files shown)


src/dune_rules/format_rules.ml
```diff
@@ -34,7 +34,7 @@ end
 module Ocamlformat = struct
   let dev_tool_lock_dir_exists () =
     let path = Dune_pkg.Lock_dir.dev_tool_lock_dir_path Ocamlformat in
-    Fs_memo.dir_exists (Path.source path |> Path.as_outside_build_dir_exn)
+    path |> Path.as_outside_build_dir_exn |> Fs_memo.dir_exists
```
Alizter, line 37:
```suggestion
    Path.as_outside_build_dir_exn path |> Fs_memo.dir_exists
```

src/dune_rules/pkg_rules.ml
```diff
@@ -1166,14 +1166,17 @@ end = struct
...
           if path_with_version_exists then path_with_version else path_without_version
         in
-        Path.Build.append_source
-          (Context_name.build_dir (Package_universe.context_name package_universe))
-          files_dir
+        let build_path =
+          Context_name.build_dir (Package_universe.context_name package_universe)
+        in
+        match files_dir with
+        | External _ -> Code_error.raise "TODO" []
```
Alizter: This looks like it was forgotten.

Alizter (commented):
I want to take a closer look at the lazy stuff you had to introduce later.

bin/pkg/pkg_common.ml
```diff
@@ -167,7 +167,7 @@ let pp_packages packages = Pp.enumerate packages ~f:pp_package
 module Lock_dirs_arg = struct
   type t =
     | All
-    | Selected of Path.Source.t list
+    | Selected of Path.t Lazy.t list
```
rgrinberg: I guess this is needed to delay `of_string` when it's used when the build directory isn't set?

rgrinberg:
Looks OK in general. There are some path types that we use for user input. E.g. what we allow users to write in the workspace file or provide in the command line. Those path types should remain what they are (perhaps they could be relaxed to use external paths someday?) as users should never provide build paths manually there.

bin/pkg/pkg_common.ml
Leonidas-from-XIV: Exactly. It crashes with the `Fdecl.get` error otherwise (and that's also the case in the other part of the code where I put module-level `of_string` into a function or made it `lazy`).

I thought about converting this into a string and only calling `of_string` later, but it seemed nicer to just promise to return a `Path.t` later instead of putting the requirement to convert into a `Path.t` onto the user of the API.

src/dune_rules/format_rules.ml
Leonidas-from-XIV: I prefer when pipes read `A |> B |> C` instead of `B A |> C` (or `C (B |> A)` to avoid the reader having to jump around in the execution order.

src/dune_rules/pkg_rules.ml
Leonidas-from-XIV: Oops, good catch! I made the error message clearer and actually also report the directory it is failing on.

src/source/workspace.ml
```diff
@@ -17,7 +17,7 @@ let default_repositories = [ Repository.overlay; Repository.upstream ]
 
 module Lock_dir = struct
   type t =
-    { path : Path.Source.t
+    { path : Path.t
```
rgrinberg: This is an instance where the old type should be preserved

Leonidas-from-XIV: That's a good point. I've reverted this change and also reverted most changes in `Lock_dir_arg` where we can also keep using `Path.Source.t` and only convert it later when retrieving the lock dir locations.

bin/pkg/pkg_common.ml
Leonidas-from-XIV: Ok, this was now obsoleted by keeping the sort of "outside interface" as `Path.Source.t`.

PR #12200, opened 2025-08-17
chore(doc): allow substituting latest version into rst files

rgrinberg:
Now we don't need to update a bunch of docs when we bump the version

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -30,6 +30,30 @@ lexers[DuneLexer.name] = DuneLexer(startinline=True)
 lexers[OpamLexer.name] = OpamLexer()
 lexers[CramLexer.name] = CramLexer()
 
+from pathlib import Path
+import re
+
+# Read constant from a file
+version_file = Path(__file__).parent.parent / "otherlibs/dune-rpc/private/types.ml"
+text = version_file.read_text()
+
+match = re.search(r"\s*let\s+latest\s*=\s*(\d+),\s*(\d+)", text)
+if not match:
+    raise RuntimeError("Could not find latest version in " + version_file)
+
+major, minor = match.groups()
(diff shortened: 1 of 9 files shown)


doc/conf.py
```diff
@@ -30,6 +30,30 @@
...
 
+from pathlib import Path
+import re
+
+# Read constant from a file
+version_file = Path(__file__).parent.parent / "otherlibs/dune-rpc/private/types.ml"
+text = version_file.read_text()
+
+match = re.search(r"\s*let\s+latest\s*=\s*(\d+),\s*(\d+)", text)
+if not match:
```
maiste, line 41: Nitpicky: what do you think about adding also a check on the size of the group? It allows failing with a more understandable error message.

maiste (approved):
Thanks for the change! This is going to be really convenient to have this 👍🏻

doc/conf.py
rgrinberg: Sure, but what is the size of the group here and how do I find it?

maiste: Indeed!
I have looked at the [documentation](https://docs.python.org/3/library/re.html#re.Match.groups()), and it seems it returns a tuple anyway. The solution seems to check `major` and `minor` are not `None`:
```python
if major != None and minor != None:
  raise RuntimeError("Could not extract major and minor version")
```

rgrinberg: Am I not doing the same thing? `match` is truthy only if `major` and `minor` both match.

maiste: Oh, sorry! Missed the part where the regex would fail if it doesn't match the second or first argument. You are right! Closing the discussion as it is not relevant.

maiste (approved):

Alizter (approved):

PR #12216, opened 2025-08-17
[3.20] prepare 3.20.0 release

maiste:
@shonfeder, could you take a look at https://github.com/ocaml/opam-repository/pull/28333? If everything looks good, we can close the Opam PR, move from the alpha stage to the official release stage.

Rereading the CI, everything compare to the previous release logs seemed to match (even though a second or third pair of eyes is always necessary).


shonfeder (approved):

shonfeder:
Reviewed! The only issues there looks like it may just require an upper bound, but I'll see what you say about it :D

PR #12219, opened 2025-08-17
refactor(boot): use [List.concat_map]

rgrinberg:


boot/duneboot.ml
```diff
@@ -102,6 +102,8 @@ module List = struct
        | None -> find_map l ~f
        | Some _ as x -> x)
   ;;
+
+  let concat_map l ~f = List.map l ~f |> List.concat
```
Alizter, line 102: If you include `ListLabels` after this, we can pick the faster implementation when it's available (4.14).

rgrinberg: Sure, could you push this change and take over this PR?

Alizter: I've added some comments explaining the fallback and removed filter_map which is available in 4.08 anyway.

PR #12223, opened 2025-08-19
refactor: move action plugin out of the engine

rgrinberg:
Now that custom actions can ask for more dependencies, there's no need for the action plugin to live inside the engine.

This better aligns the implementation with upstream.

### DIFF
--- a/src/dune_engine/action_to_sh.ml
+++ b/src/dune_engine/action_to_sh.ml
@@ -48,7 +48,6 @@ let simplify act =
     match act with
     | Run (prog, args) -> Run (prog, Array.Immutable.to_list args) :: acc
     | With_accepted_exit_codes (_, t) -> loop t acc
-    | Dynamic_run (prog, args) -> Run (prog, args) :: acc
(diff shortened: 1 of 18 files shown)


src/dune_engine/action_to_sh.ml
```diff
@@ -48,7 +48,6 @@ let simplify act =
     match act with
     | Run (prog, args) -> Run (prog, Array.Immutable.to_list args) :: acc
     | With_accepted_exit_codes (_, t) -> loop t acc
-    | Dynamic_run (prog, args) -> Run (prog, args) :: acc
```
Alizter, line 51: Could you add an optional field to the action extension module so that shell commands can be given for custom actions voluntarily?

rgrinberg: Not sure I understand. How are we supposed to give valid shell commands for custom actions?

Alizter: OK never mind for now.

PR #12227, opened 2025-08-19
Fix handling of root module dependencies

rgrinberg:
The root module is generated, therefore it should have no dependencies.

Trying to discover dependencies for it can lead to spurious cycles.

The fix for this is hacky because we have multiple places we need to remember to cut off spurious deps. Will fix this in follow up PR's.

### DIFF
--- a/doc/changes/12227.md
+++ b/doc/changes/12227.md
@@ -1,0 +1,2 @@
+- Don't run ocamldep to compute false dependencies on the `root_module`
+  (#12227, @rgrinberg)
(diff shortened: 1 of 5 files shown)


doc/changes/12227.md
```diff
@@ -0,0 +1,2 @@
+- Do run ocamldep to compute false dependencies on the `root_module` (#12227,
```
Alizter: Don't?

rgrinberg: Yes indeed. Feel free to push a fix.

Alizter (approved):

PR #12231, opened 2025-08-20
fix: make Dune_util.xdg lazier

Alizter:
I would like to be able to override the value of `Dune_util.xdg` during testing, since it gives us much more control over where certain files are being placed.

An override mechanism can be added to `Dune_util.xdg`, however one quickly realises that other places in dune are forcing this value eagarly. Therefore if we could make those places lazier, we can have enough of a window to override the value of `Dune_util.xdg` for testing purposes.

The first commit consists of a tests that depends on many dune libraries and shows that the value of `Dune_util.xdg` is being forced eagrly.

The second commit threads `Lazy.t` through various places in order to delay the forcing as much as possible. This works and allows for the override mechanism to function correctly. I would like to use this mechanism in the tests for #12188. Note that the test from the first commit makes sure we aren't forcing.

### DIFF
--- a/bin/common.ml
+++ b/bin/common.ml
@@ -1352,7 +1352,7 @@ let init (builder : Builder.t) =
   Log.info
     [ Pp.textf
         "Shared cache location: %s"
-        (Path.to_string Dune_cache_storage.Layout.root_dir)
+        (Path.to_string (Lazy.force Dune_cache_storage.Layout.root_dir))
     ];
   Dune_rules.Main.init
     ~stats:c.stats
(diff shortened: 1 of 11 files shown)


src/dune_util/dune_util.ml
```diff
@@ -12,18 +12,30 @@ module Alias_name = Alias_name
...
+           | None ->
+             let uid = Unix.getuid () in
+             (match Unix.getpwuid uid with
+              | exception Not_found -> None
+              | s -> Some s.pw_dir))
+       in
+       Xdg.create ~env:(Env.get env_map) ())
+;;
+
+let override_xdg : Xdg.t -> unit =
```
rgrinberg: Where is this function used?

Alizter: It's not used yet. I can remove it for now if you like.

Alizter: I have some changes the the rev_store tests I want to introduce before #12188 and I'll use it there.

rgrinberg (approved):

PR #12233, opened 2025-08-20
fix(pkg): fallback to gpatch when patch isn't available

### LINKED ISSUE #12232: [pkg] gpatch vs patch in package management
Hello,

While trying to move build my project using the dune developer preview, I hit [this issue](https://discuss.ocaml.org/t/issue-with-gel-prevents-install-of-jane-street-core-under-5-3-0/16909/9).

Interestingly, this is not an issue with opam, so I guess opam knows to inject `gpatch` in the path somehow instead of `patch`, and dune isn't doing that?

Alizter:
We add a fallback to gpatch when searching for patch. This won't be necessary when we have an OCaml implementation of dune_patch, but that is still some time away.

- fix https://github.com/ocaml/dune/issues/12232

### DIFF
--- a/src/dune_patch/dune_patch.ml
+++ b/src/dune_patch/dune_patch.ml
@@ -90,9 +90,11 @@ let patches_of_string patch_string =
 
 let prog =
   lazy
-    (match Bin.which ~path:(Env_path.path Env.initial) "patch" with
+    (let path = Env_path.path Env.initial in
+     let bins = [ "gpatch"; "patch" ] in
+     match List.find_map bins ~f:(Bin.which ~path) with
      | Some p -> p
-     | None -> User_error.raise [ Pp.text "patch not found" ])
+     | None -> User_error.raise [ Pp.textf "%s not found." (String.enumerate_and bins) ])
 ;;
 
 let exec display ~patch ~dir ~stderr =


src/dune_patch/dune_patch.ml
```diff
@@ -90,7 +90,8 @@ let patches_of_string patch_string =
 
 let prog =
   lazy
-    (match Bin.which ~path:(Env_path.path Env.initial) "patch" with
+    (let path = Env_path.path Env.initial in
+     match List.find_map [ "patch"; "gpatch" ] ~f:(Bin.which ~path) with
      | Some p -> p
      | None -> User_error.raise [ Pp.text "patch not found" ])
```
rgrinberg: "patch and gpatch not found"

rgrinberg (approved):
Thanks for the quick fix

src/dune_patch/dune_patch.ml
```diff
@@ -90,9 +90,11 @@ let patches_of_string patch_string =
 
 let prog =
   lazy
-    (match Bin.which ~path:(Env_path.path Env.initial) "patch" with
+    (let path = Env_path.path Env.initial in
+     let bins = [ "patch"; "gpatch" ] in
```
rgrinberg: Actually, shouldn't you lookup gpatch first?

Alizter: I'm not sure I follow. I would imagine `patch` to be more available than `gpatch`. For instance, I don't have `gpatch`.

rgrinberg: Yes, `patch` is more common. But the set of people that have both utilities would almost certainly prefer `gpatch`. Otherwise, why did they install it explicitly?

Alizter: Fair enough. Done that.

PR #12235, opened 2025-08-20
refactor(fiber): switch [Fiber.Var] to be array backed

rgrinberg:
This turns out to be faster as there's few elements in the array

### DIFF
--- a/src/fiber/src/fiber.mli
+++ b/src/fiber/src/fiber.mli
@@ -123,13 +123,13 @@ module Var : sig
   type 'a t
 
   (** Create a new variable *)
-  val create : unit -> 'a t
+  val create : 'a -> 'a t
 
   (** [get var] reads the value of [var]. *)
-  val get : 'a t -> 'a option fiber
+  val get : 'a t -> 'a fiber
 
   (** Same as [get] but raises if [var] is unset. *)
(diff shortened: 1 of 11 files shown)


src/fiber/src/fiber.mli
```diff
@@ -123,13 +123,13 @@ module Var : sig
...
-  val create : unit -> 'a t
+  val create : 'a -> 'a t
 
   (** [get var] reads the value of [var]. *)
-  val get : 'a t -> 'a option fiber
+  val get : 'a t -> 'a fiber
 
   (** Same as [get] but raises if [var] is unset. *)
-  val get_exn : 'a t -> 'a fiber
+  val get_exn : 'a option t -> 'a fiber
```
Alizter, line 132: I'm a bit confused at the API here. Why is there an option here now?

rgrinberg: Because this can function can only raise if the value inside is an option.

snowleopard:
Just to be clear: switching to arrays is not the only change here. We also switched to non-optional variable payload type, which also removes one layer of indirection.

rgrinberg:
Unfortunately, the layer of indirection is basically re-introduced in this PR because I haven't yet upstreamed unboxed options. Stay tuned.

PR #12236, opened 2025-08-20
Expose 'ocamlc -config' parameter 'ox'

jonludlam:
This is to allow for easy access to whether a compiler is an oxcaml one or not. 

I initially exposed this as the parameter 'ocaml-config:oxcaml' but then decided to stick precisely with what the output of `ocamlc -config` gives. This makes it hard to change the name of that value, so it's worth confirming this constraint is OK with the oxcaml devs.


test/blackbox-tests/test-cases/ocaml-config-ox.t/run.t
```diff
@@ -0,0 +1,5 @@
+The macro %{ocaml-config:ox} should be available. we don't print anything because
```
Alizter: Could you turn this into a file cram test? See some of the other file cram tests to get an idea of how to write the files. It makes it easier to see what the test is about in one place.

rgrinberg:
For reference, do you have the corresponding PR that introduced this change?

Alizter:
We already have the `%{oxcaml_supported}` variable. Or is this different?

test/blackbox-tests/test-cases/ocaml-config-ox.t/run.t
jonludlam: I based this one on https://github.com/ocaml/dune/tree/main/test/blackbox-tests/test-cases/ocaml-config-macro.t.

I'm not quite sure what you're suggesting? If the variable isn't understood the dune build line would raise an error, right?

jonludlam:
> For reference, do you have the corresponding PR that introduced this change?

https://github.com/oxcaml/oxcaml/commit/86911a1425aa3573d9c268e0853b25d21669911b

jonludlam:
> We already have the `%{oxcaml_supported}` variable. Or is this different?

This seems a bit more robust to me than matching the version number to 'jst' or 'ox', and fitted in nicely with the other ocaml-config variables. I think there's a reasonable argument to expose the rest of the values provided by `ocamlc -config`, many of which seem to me to be a bit like feature flags (e.g. 'parameterised-modules', 'stack-allocation', etc). This could help as these features get upstreamed, maybe?

(I should probably also mention that the `oxcaml_supported` change predates the change that added the flag to ocamlc -config)

test/blackbox-tests/test-cases/ocaml-config-ox.t/run.t
Alizter: Cram tests that are directories are set up like:
```
mytest.t
|- run.t
|- foo
|- bar
```
when you run `mytest.t` as a test, you have access to the files `foo` and `bar`.

That's great, but sometimes its easier to see exactly what syntax is happening by having `mytest.t` as a file directly instead. That way you can do:

```
  $ cat > foo <<EOF
  > content of foo ....
  > EOF

  $ cat > bar <<EOF
  > content of bar ....
  > EOF

  $ commands from run.t above
```

I'm only suggesting this because it's clearer what the stanzas involved are, without having to open another file.

rgrinberg:
> This seems a bit more robust to me than matching the version number to 'jst' or 'ox', and fitted in nicely with the other ocaml-config variables

I agree. Why don't we change the implementation of oxcaml_supported then as well?

> I think there's a reasonable argument to expose the rest of the values provided by ocamlc -config, many of which seem to me to be a bit like feature flags (e.g. 'parameterised-modules', 'stack-allocation', etc). This could help as these features get upstreamed, maybe?

That would be welcome as well.

rgrinberg (approved):

test/blackbox-tests/test-cases/ocaml-config-ox.t/dune
maiste, line 1: My suggestion would be for the test cases to be in the `test-cases/oxcaml` repository as it is related to it.

maiste (commented):
I agree that the goal was to have this to replace the `+ox` lookup functions. We should update these functions to also check for the `ocaml_config`. This way we would make `oxcaml_supported` use it.

rgrinberg:
@jonludlam do you mind adding a CHANGES entry?

test/blackbox-tests/test-cases/ocaml-config-ox.t/dune
jonludlam: It is related, true, but these test cases are only run when running on oxcaml, yes? I'd quite like to make sure it doesn't cause a failure if it's run under mainline ocaml too.

jonludlam: I've added a test in the oxcaml dir too, which actually checks the result.

test/blackbox-tests/test-cases/ocaml-config-ox.t/run.t
jonludlam: Thanks, that does indeed look clearer!

test/blackbox-tests/test-cases/oxcaml/ocaml-config-ox-true.t
```diff
@@ -0,0 +1,14 @@
+Check that %{ocaml-config:ox} works and returns true
+
+  $ cat > "dune-project" <<EOF
+  > (lang dune 1.0)
+  > EOF
+
+  $ cat > "dune" <<EOF
+  > (rule
+  >   (targets x)
+  >     (action (with-stdout-to %{targets} (echo %{ocaml-config:ox}))))
```
Alizter, line 10:
```suggestion
  >     (action (with-stdout-to %{targets} (echo %{ocaml-config:ox}))))
  > EOF
```

jonludlam: thanks!

PR #12238, opened 2025-08-20
feature: root_module in bootstrap

rgrinberg:
Allow the root_module in bootstrap

This allows us to write proper wrapper modules like I've done for Pp in this PR.


otherlibs/stdune/src/import.ml
```diff
@@ -1,31 +0,0 @@
-module Pp = struct
```
Alizter, line 1: We probably don't need this file anymore right?

rgrinberg: I'm sure we're going to re-introduce for one reason or another

rgrinberg:
Alright, this works well now. The obvious candidate to use this would be in the dune executable stanza itself. Unfortunately, due to another limitation in the bootstrapping process, this cannot be done at the moment. One step at a time.

This change requires bootstrap to be re-done again.

PR #12242, opened 2025-08-21
test: make sure runtest_alias works properly

vouillon:


Alizter:
I've promoted the test output and am working on a fix.

Alizter:
I've pushed a fix. Instead of `runtest-js` depending on `runtest-foo` it will now depend on the separate `runtest-js-foo`.

Alizter:
@maiste This probably should go in 3.21.1.

rgrinberg (approved):

vouillon:
@Alizter Thanks!

maiste:
@Alizter I suggest that we wait a bit before doing `3.20.1`. Just the time to have feedback on the `3.20.0` release first. If there are other errors, we can fix them in the same release.

_EDIT_: considering the other changes, we will do the release fast.

PR #12245, opened 2025-08-21
Rollback dune file formatting change in 3.20.0

nojb:
The formatting change introduced in 3.20.0 (see #10892) has resulted in a number of complaints because of its unversioned nature (it does not depend on the version of the Dune language of the file being formatted). After discussion with @Alizter mainly we think that it may be a good idea to revert this change until we have a good versioning story.

The point of this PR is to have a patch ready in case we decide to do so in time for 3.20.1.

### DIFF
--- a/bench/dune
+++ b/bench/dune
@@ -1,8 +1,15 @@
 (executable
  (name bench)
  (modules bench metrics)
- (libraries dune_stats dune_console chrome_trace stdune fiber dune_lang
-  dune_engine dune_util))
+ (libraries
+  dune_stats
+  dune_console
+  chrome_trace
+  stdune
+  fiber
+  dune_lang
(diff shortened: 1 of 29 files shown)


maiste (commented):
Thanks @nojb! Dummy question but I don't see in your patch the code change to go back to the old default. Am I missing something?

nojb:
> Thanks @nojb! Dummy question but I don't see in your patch the code change to go back to the old default. Am I missing something?

It's here:

https://github.com/ocaml/dune/pull/12245/files#diff-54529f58e2fe93f603fd822d61c0a5ef1bed8702c03c61766928ab24e80a230c

maiste (approved):
Thanks! I missed it in the list of files.

PR #12246, opened 2025-08-21
add melange to `@ocaml-index` alias

adrianparvino:
This change makes `@ocaml-index` alias aware of Melange by making the alias aware of the Melange stanza and making the cctx rule correctly resolve the Melange path


Alizter:
ping @anmonteiro

anmonteiro (approved):
thank you!

PR #12252, opened 2025-08-22
Publish Dune Docker images during release

### LINKED ISSUE #12177: Docker base image for `dune pkg`
## Docker Base Image for `dune pkg`

With dune package management supporting managing packages, some projects have migrated to using dune pkg for deployment, e.g [ocaml.org](http://ocaml.org/) (https://github.com/ocaml/ocaml.org/pull/3132), [tarides.com](http://tarides.com/) (private repo). They don’t require to use `opam` for deployment. This is also in-line with the description of Dune as a frontend in the [OCaml platform roadmap](https://ocaml.org/tools/platform-roadmap#g1-dune-is-the-frontend-of-the-ocaml-platform). As dune package management evolves, more projects might want to migrate to it. In this context, it will be useful to have a docker base image for dune pkg that doesn’t depend on the opam base images. 

###  `docker-base-images`

[docker-base-images](https://github.com/ocurrent/docker-base-images) is a CI pipeline that generates a bunch opam and compiler (opam + compiler switch) base images. See [images generated](https://images.ci.ocaml.org/). One possibility is to generate a base image for `dune pkg` with the same pipeline, as a different branch from opam. This was proposed in https://github.com/ocurrent/docker-base-images/pull/326 and passed on as it complicates the pipeline and is perhaps better lived elsewhere.

### Separate repository

I’ve prototyped here a Docker base image from scratch for `dune-pkg` - https://github.com/ocaml-dune/docker-dune. This way, the image generation lives outside in a separate repository and can be updated for new releases. The current prototype is very basic and only installs the lates Dune developer preview on alpine. If we were to follow this approach, will need to sketch out a plan for including widely used Dune versions and base images.

### Push images with release

One approach I’d like to propose is to include Docker image generation in this repository’s CI. This can be done using GitHub Actions, as described here: https://docs.github.com/en/actions/tutorials/publish-packages/publish-docker-images#publishing-images-to-docker-hub.
The advantage over maintaining Docker images in a separate repository is that pushing new images would be part of the release process and wouldn’t require additional maintenance effort. We discussed this with @shonfeder and are inclined to take this approach for the ease of generating and maintaining images. Before we proceed, we wanted to check if anyone has thoughts on this.

Sudha247:
This change is in line with #12177, which proposed to add a GitHub action that publishes a minimal Dune Docker image on every release. I've tried to keep the Docker file very minimal by building it with the OCaml compiler that comes with the distribution (on Debian this is currently 5.3.0, but it's possible the distribution OCaml compiler version is updated much later than the compiler release).

This GitHub Action essentially builds and pushes the Docker image on every new release to the GitHub container registry. This can be pushed to other registries such as Docker Hub, but it's perhaps easiest to start with GitHub itself. If we decide to go ahead with this approach, we may need to update `secrets.GITHUB_TOKEN`.

Fixes #12177

### DIFF
--- a/docker/release.Dockerfile
+++ b/docker/release.Dockerfile
@@ -1,26 +1,23 @@
-ARG BASE_IMAGE=debian:latest
-FROM ${BASE_IMAGE}
+ARG BASE_IMAGE=debian:13
+FROM ${BASE_IMAGE} AS build
 
-RUN apt-get update && apt-get upgrade -y
-RUN apt-get install -y build-essential curl git sudo ocaml wget
+RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser build-essential curl ocaml
 
-RUN git config --global user.email "docker@example.com"
-RUN git config --global user.name "Docker"
+RUN addgroup --gid 1000 dune && adduser --uid 1000 --ingroup dune dune
 
-RUN addgroup --system --gid 1000 dune
-RUN adduser --system --uid 1000 --ingroup dune dune
+USER dune:dune
 WORKDIR /home/dune
 
 # Use release assets from Docker context
-RUN tar -xvjf dune*.tbz
-RUN mv dune-[0-9]*/ dune
-WORKDIR /home/dune/dune
+RUN tar -xf dune*.tbz && mv dune-*/ dune
 
 # Build and install dune
-RUN ./configure --prefix=/usr/local
-RUN make bootstrap
-RUN make release
-RUN make install
+RUN cd dune && ./configure --prefix=/home/dune/install && make bootstrap && make release && make install
 
-# Verify installation
-RUN dune --version
+FROM ${BASE_IMAGE} AS run
+
+RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser curl ocaml git && rm -rf /var/lib/{apt,dpkg,cache,log}/
+RUN addgroup --gid 1000 dune && adduser --uid 1000 --ingroup dune dune
+
+COPY --from=build /home/dune/install/bin/dune /usr/local/bin
+USER dune:dune
--- a/docker/release.Dockerfile
+++ b/docker/release.Dockerfile
@@ -1,7 +1,7 @@
 ARG BASE_IMAGE=debian:13
 FROM ${BASE_IMAGE} AS build
 
-RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser build-essential curl ocaml
+RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser build-essential ocaml
 
 RUN addgroup --gid 1000 dune && adduser --uid 1000 --ingroup dune dune
 
--- a/docker/release.Dockerfile
+++ b/docker/release.Dockerfile
@@ -1,0 +1,26 @@
+ARG BASE_IMAGE=debian:latest
+FROM ${BASE_IMAGE}
+
+RUN apt-get update && apt-get upgrade -y
+RUN apt-get install -y build-essential curl git sudo ocaml wget
+
+RUN git config --global user.email "docker@example.com"
+RUN git config --global user.name "Docker"
+
+RUN addgroup --system --gid 1000 dune
+RUN adduser --system --uid 1000 --ingroup dune dune
+WORKDIR /home/dune
+
+# Use release assets from Docker context
+RUN tar -xvjf dune*.tbz
+RUN mv dune-[0-9]*/ dune
+WORKDIR /home/dune/dune
+
+# Build and install dune
+RUN ./configure --prefix=/usr/local
+RUN make bootstrap
+RUN make release
+RUN make install
+
+# Verify installation
+RUN dune --version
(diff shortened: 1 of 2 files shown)


docker/release.Dockerfile
```diff
@@ -0,0 +1,26 @@
+ARG BASE_IMAGE=debian:latest
+FROM ${BASE_IMAGE}
+
+RUN apt-get update && apt-get upgrade -y
+RUN apt-get install -y build-essential curl git sudo ocaml wget
+
+RUN git config --global user.email "docker@example.com"
+RUN git config --global user.name "Docker"
```
Leonidas-from-XIV: Why this? We don't seem to be using `git`?

docker/release.Dockerfile
```diff
@@ -0,0 +1,26 @@
+ARG BASE_IMAGE=debian:latest
+FROM ${BASE_IMAGE}
+
+RUN apt-get update && apt-get upgrade -y
+RUN apt-get install -y build-essential curl git sudo ocaml wget
```
Leonidas-from-XIV: The build is run by root so I don't see the point of `sudo`, it doesn't seem to use `git` (though maybe useful for Dune pkg, along with `curl`), but we don't use `wget` at all?

docker/release.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+RUN git config --global user.email "docker@example.com"
+RUN git config --global user.name "Docker"
+
+RUN addgroup --system --gid 1000 dune
+RUN adduser --system --uid 1000 --ingroup dune dune
+WORKDIR /home/dune
```
Leonidas-from-XIV, line 9: We never switch to the `dune` user. Probably would make sense to do so.

docker/release.Dockerfile
```diff
@@ -0,0 +1,26 @@
+ARG BASE_IMAGE=debian:latest
```
Leonidas-from-XIV: I am unsure whether `debian:latest` is the best choice, because that means that it might change overnight if a new Debian version is released. It might be useful however I am always wary when things shift underneath our feet.

docker/release.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+RUN git config --global user.email "docker@example.com"
+RUN git config --global user.name "Docker"
+
+RUN addgroup --system --gid 1000 dune
+RUN adduser --system --uid 1000 --ingroup dune dune
+WORKDIR /home/dune
+
+# Use release assets from Docker context
+RUN tar -xvjf dune*.tbz
```
Leonidas-from-XIV: `tar xf dune*.tbz` will output less spammy information during the build.

docker/release.Dockerfile
maiste: I would second @Leonidas-from-XIV here. Considering debian has "slow" cycle, it is simple to upgrade the number by hand. I would stick to `debian:trixie`. It makes it more explicit.

docker/release.Dockerfile
```diff
@@ -0,0 +1,23 @@
+ARG BASE_IMAGE=debian:13
+FROM ${BASE_IMAGE} AS build
+
+RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser build-essential curl ocaml
```
Leonidas-from-XIV: Probably we can now also remove curl here.

Leonidas-from-XIV (commented):
The Docker file looks pretty good to me, but I wonder if there is any way to test the workflow? I don't know what magic GH does to make the downloaded tarball available in the file system inside docker so I am unsure whether it will find the `dune-*.tbz` file in the location the Docker file expects it to be.

It would be good to test that it is indeed working as expected.

Alizter:
There is typically a manual trigger for workflows you can set. That way you can manually trigger it to test it. It might be worth doing in addition to the release trigger.

docker/release.Dockerfile
shonfeder: We don't need curl for dune pkg?

Leonidas-from-XIV: Not in the `build` image, `dune pkg` is not used to build dune (and it fact, cannot). It's only required in the `run` image.

shonfeder: Ah, I missed that. Thanks.

So we can just try

```suggestion
RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser build-essential ocaml
```

Leonidas-from-XIV:
The code looks fine to me and it's ok to merge if building works. However, how can we that it works when built using the Github Actions? I see something has been added to that extent by @Sudha247 but how do I trigger it?

Alizter:
Seems you have to merge into main before you can trigger it manually. I suppose it's sensible to merge this now and then check it is working. If not, we can fix it afterwards in a separate PR.

Leonidas-from-XIV:
Hmm, I tried running the workflow manually [using this page](https://github.com/ocaml/dune/actions/workflows/docker-image.yml). Unfortunately, when I pick `main` it fails as it can't find a release, if I pick a released tag I can't run the workflow as it doesn't exist in the tagged release.

@Sudha247 Maybe you could add a way to force a specific release when running the trigger manually?

Sudha247:
Ah, I see. Let me try to add a way to force a release when running the manual trigger.

PR #12256, opened 2025-08-22
rev_store: document and improve API

Alizter:
I always found the revision store API a bit hard to understand, so I spent some time documenting the various parts. I've also reorganised the API a little in ways I think make a little more sense.

This was done when working on #12188.

I will follow this PR up with some improvements to the testing.

### DIFF
--- a/src/dune_pkg/rev_store.ml
+++ b/src/dune_pkg/rev_store.ml
@@ -128,11 +128,19 @@ module File = struct
 end
 
 module Remote = struct
-  type nonrec t =
+  type t =
     { url : string
     ; default_branch : Object.resolved option Fiber.t
     ; refs : Object.resolved String.Map.t Fiber.t
     }
+
+  let to_dyn { url; default_branch; refs } =
+    Dyn.record
+      [ "url", Dyn.string url
+      ; "default_branch", Dyn.opaque default_branch
+      ; "refs", Dyn.opaque refs
+      ]
+  ;;
 
   let default_branch t = t.default_branch
 end
@@ -144,6 +152,18 @@ type t =
     object_mutexes : (Object.t, Fiber.Mutex.t) Table.t
   ; present_objects : (Object.t, unit) Table.t
   }
+
+let to_dyn { dir; remotes; object_mutexes; present_objects } =
+  Dyn.record
+    [ (* This is an external path, so we relativize to sanitize. We don't use
+         [Path.to_dyn] since it wouldn't be correct and therefore confusing. *)
+      "dir", Path.Expert.try_localize_external dir |> Path.to_string |> Dyn.string
+    ; "remotes", Table.to_list remotes |> Dyn.list (Dyn.pair Dyn.string Remote.to_dyn)
+    ; "object_mutexes", Dyn.opaque object_mutexes
+    ; ( "present_objects"
+      , Table.to_list present_objects |> Dyn.list (Dyn.pair Object.to_dyn Dyn.unit) )
+    ]
+;;
 
 let with_mutex t obj ~f =
   let* () = Fiber.return () in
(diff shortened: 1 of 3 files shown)


src/dune_pkg/rev_store.mli
```diff
@@ -1,9 +1,31 @@
...
+
+(** Abstract handle to the shared revision store. *)
 type t
 
+val to_dyn : t -> Dyn.t
+
+(** Get the revision store and initialise if it hasn't been already. *)
+val get : t Fiber.t
+
+(* TODO: remove *)
```
rgrinberg: I don't think short comments like that are all that useful. Either explain why something has to be removed, or just remove this comment.

src/dune_pkg/rev_store.mli
```diff
@@ -12,44 +34,85 @@ module Object : sig
...
+  val check_out : t -> target:Path.t -> unit Fiber.t
 
-  val default_branch : t -> Object.resolved option Fiber.t
+  (** For testing only. *)
+  module Config : sig
+    val parse : string -> (string * string option * string * string) option
+  end
 end
 
-val remote : t -> url:Loc.t * string -> Remote.t
```
rgrinberg, line 49: I don't see the point of removing this function and replacing it with `Rev_store.Remote.get`

src/dune_pkg/rev_store.mli
```diff
@@ -12,44 +34,85 @@ module Object : sig
...
-  val path : t -> Path.Local.t
   val to_dyn : t -> Dyn.t
 
+  (** Path relative to the repository root at the given revision. *)
+  val path : t -> Path.Local.t
+
   module Set : Set.S with type elt = t
 end
 
+(* TODO: is this the best representation for this? *)
```
rgrinberg: This comment doesn't really explain anything to me. I'd get rid of it, or clarify what it means.

rgrinberg (changes requested):

src/dune_pkg/rev_store.mli
Alizter: Not sure what I was thinking either. Reverted.

src/dune_pkg/rev_store.mli
Alizter: Removed.

src/dune_pkg/rev_store.ml
```diff
@@ -665,10 +685,10 @@ module At_rev = struct
     ;;
   end
 
-  let files_and_submodules repo rev =
```
rgrinberg: How's the name "key" better? At least "rev" describes what is the key.

Alizter: I'll change it back to revision.

src/dune_pkg/rev_store.mli
```diff
@@ -12,44 +35,84 @@ module Object : sig
...
-  val equal : t -> t -> bool
-  val check_out : t -> target:Path.t -> unit Fiber.t
-end
 
-module Remote : sig
-  (** handle representing a particular git repository *)
-  type t
+  (** [check_out t ~target] materialises the entire tree into [target]
+      including any submodules. *)
+  val check_out : t -> target:Path.t -> unit Fiber.t
```
Leonidas-from-XIV, line 93: Why is this needed? I think one of the ideas of the revision store is that we never need to materialize anything.

Leonidas-from-XIV (approved):
The comments added seem to be accurate wrt to how the API is supposed to work.

src/dune_pkg/rev_store.mli
Alizter: This is how we materialise sources in the package rules I believe. In Git we also call switching to a branch checking out, but this is really about the more specific operation of choosing a revision and a file and materialising it.

Leonidas-from-XIV: Yes, but checking out materializes a work tree, which the rev store shouldn't do (at least not for opam repositories). But yeah, maybe it is used for package sources that are pinned via git, that would make sense.

Alizter: Just so that we are on the same page, we are talking about materialising sources somewhere else not inside the rev_store right?

Leonidas-from-XIV: Yes. The rev store is always just a bare git repo without a worktree (which is by design, to allow concurrent access, without fighting which worktree to materialize).

rgrinberg (approved):

PR #12257, opened 2025-08-23
refactor: decoder simplification

rgrinberg:


Alizter:
Another idea with the above function. Instead of recording and comparing backtrace length, why don't we just compare the location stop?

rgrinberg:
That just has the potential to improve the error message. Useful, but it's more important to make the backtracking itself better behaved.

PR #12259, opened 2025-08-23
doc: fix typo in pkg_toolchain.mli

Alizter:
https://askubuntu.com/questions/1312677/where-is-xdg-cache-dir


Leonidas-from-XIV (approved):
I just checked and we (correctly) use `XDG_CACHE_HOME` in the codebase, so it is really just a typo.

Issue #12085, opened 2025-07-28
[OxCaml] Parameterized libraries: parameter implementation


maiste:
This issue tracks the work related to the implementation of the parameter in the parameterized libraries. It depends on #12084 implementation to be able to work.

PR #12276, opened 2025-08-24
refactor: represent virtual libraries via `Lib_kind.t`

### LINKED ISSUE #12253: Refactor representation of library virtuality to represent disjoint logic in the type system
As a followup to https://github.com/ocaml/dune/pull/11963 and in preparation for https://github.com/ocaml/dune/issues/12085, we want to move the representation of whether or not libraries are virtual into a cleaner and safer encoding in the type system. @rgrinberg noted the value of this improvement in https://github.com/ocaml/dune/pull/12002#discussion_r2248837703 . The motivating need here is to reduce the chance of introducing defects while reasoning about implementable libraries, to be added in #11963.

Currently, we represent whether or not a library is virtual via [the boolean record field `virtual_` in `Lib_info.t`](https://github.com/ocaml/dune/blob/e15fab3293dcfbbd1a62ba665f2593986611e6b0/src/dune_rules/lib_info.ml#L327). However, this property is in fact disjoint with [the alternatives in `Lib_kind.t`](https://github.com/ocaml/dune/blob/e15fab3293dcfbbd1a62ba665f2593986611e6b0/src/dune_lang/lib_kind.mli#L16-L20), which we record in [`Lib_info.kind`](https://github.com/ocaml/dune/blob/e15fab3293dcfbbd1a62ba665f2593986611e6b0/src/dune_rules/lib_info.ml#L304). This leaves us with the suboptimal situation of making it possible to construct invalid values of type `Lib_info.t`, such as `{... virtual_: true; parameter = true; ...}`. By instead representing virtuality as one of the alternatives of `Lib_kind.t`, we will accurately represent the mutual incompatibility between the property of being virtual and the other possible kinds, making it impossible to represent invalid libraries in this way, and statically eliminating a category of error that is otherwise possible when moving forward with #12085.

shonfeder:
Closes #12253

The refactor here is quite straightforward, with the exception of maintaining backwards compat with current syntax of library stanzas. To achieve this, we infer `kind virtual` whenever `virtual_modules` are specified alongside a compatible kind (`normal` or `virtual`).

Two incidental improvements are also suggested here: small fixes to the docs and cleaning up a case analysis handling `ppx_*` kinds.

## TODO

To wrap this up, I need 

- [x] input on how to signal the right error for incompatible library specifications
- [x] add a test case to check the error handling of invalid cases

## Reviewing

Initial review by commit is probably helpful, since the commits are broken into a few tiny, distinct phases.


src/dune_lang/lib_kind.ml
```diff
@@ -79,6 +81,7 @@ let decode =
   let open Decoder in
   sum
     [ "normal", return Normal
+    ; "virtual", return Virtual
     ; "parameter", return Parameter
```
shonfeder: Should these not be guarded by the version in which they were introduced? Or is there ar reason that is not needed when extending a sum?

src/dune_rules/stanzas/library.ml
```diff
@@ -118,7 +118,22 @@ let decode =
...
+           ~since_expanded:Stanza_common.Modules_settings.since_expanded
+           "virtual_modules"
+       in
+       let+ kind = field "kind" Lib_kind.decode ~default:Lib_kind.Normal in
+       let kind : Lib_kind.t =
+         match virtual_modules, kind with
+         | Some _, (Normal | Virtual) -> Virtual
+         | None, kind -> kind
+         | Some _, _incompatible_kind ->
+           raise (Failure "TODO: Error on incompatible kind")
```
shonfeder: What's the right way to signal an error here? `User_error.raise`?

Alizter: If you want this to be an error to the user then `User_error.raise`. If it is a programming error it should be a `Code_error.raise`.

src/dune_lang/lib_kind.ml
```diff
@@ -63,13 +63,15 @@ type t =
   | Ppx_deriver of Ppx_args.t
   | Ppx_rewriter of Ppx_args.t
   | Parameter
+  | Virtual
```
shonfeder, line 107: IIUC, an even more robust encoding would represent this as `Virtual of <virtual_modules>` however that is a more invasive change, and I think would call for deprecating the existing syntax for `virtual_modules`. This could be something we pursue as a followup, but I think should not be included here, as it doesn't help meaningfully with advancing https://github.com/ocaml/dune/issues/12085, which is the focus here.

src/dune_rules/gen_meta.ml
```diff
@@ -92,8 +92,9 @@ let gen_lib pub_name lib ~version =
   let* ppx_rt_deps =
     Lib.ppx_runtime_deps lib |> Memo.bind ~f:Resolve.read_memo |> Memo.map ~f:to_names
   in
-  let+ ppx_runtime_deps_for_deprecated_method =
-    (* For the deprecated method, we need to put all the runtime dependencies of
+  let+ ppx_common_requires, ppx_common_negations =
```
rgrinberg: I don't understand the changes in this source file. Are they unrelated?

src/dune_rules/stanzas/library.ml
```diff
@@ -118,7 +118,26 @@ let decode =
...
          ~default:(Mode_conf.Lib.Set.default stanza_loc)
-     and+ kind = field "kind" Lib_kind.decode ~default:Lib_kind.Normal
+     and+ virtual_modules, kind =
+       let* virtual_modules =
+         Ordered_set_lang.Unexpanded.field_o
+           ~check:(Dune_lang.Syntax.since Stanza.syntax (1, 7))
+           ~since_expanded:Stanza_common.Modules_settings.since_expanded
+           "virtual_modules"
+       in
+       let+ kind = field "kind" Lib_kind.decode ~default:Lib_kind.Normal in
```
rgrinberg: This seems wrong. You should only accept the old values for `Lib_kind.t` here. It might be helpful to separate the effective library kind from what the user can write in a dune file at the type level.

src/dune_rules/dune_package.ml
```diff
@@ -192,7 +190,22 @@ module Lib = struct
...
-       and+ kind = field "kind" Lib_kind.decode
+       and+ kind =
+         let* kind = field "kind" Lib_kind.decode in
+         let+ virtual_ = field_b "virtual" in
+         match kind with
+         | (Normal | Virtual) when virtual_ ->
+           (* Backward compatible support for dune-project files that
+              that include the [(virtual)] field. )
+              TODO: can we migrate this instead,
+              If not, how do we eventually depricate it? *)
```
rgrinberg: The customary thing to do would be to look for a `kind` field and use that. Otherwise, look for a `virtual` field.

shonfeder: Ok, that's how it is working now.

src/dune_rules/gen_meta.ml
shonfeder: Related only insofar as they prevent a redundant case analysis that we otherwise need to adjust. Thus in the separate commit https://github.com/ocaml/dune/pull/12276/commits/bb3a6885eca84bfe2280e777f1e2cff7d49271ee 

I can drop this if you prefer.

src/dune_rules/stanzas/library.ml
shonfeder: Ok, so to be clear, we do not want users to be able to write `(kind virtual)`? And is this is because we want this to be determined solely by the presence or absence of the `virtual_modules` field?

shonfeder: > It might be helpful to separate the effective library kind from what the user can write in a dune file at the type level.

@rgrinberg could you point me towards an illustrative example of where this kind of separation is introduced and maintained? I am not sure what the conventions are around how we deal with implementing `Conv.S` for a type which has been decoupled from its representation in the dune lang.

rgrinberg: > Ok, so to be clear, we do not want users to be able to write (kind virtual)? And is this is because we want this to be determined solely by the presence or absence of the virtual_modules field?

That's right. We could also have done in a way where both `kind` and `virtual_modules` had to be set for virtual libraries, but at this point, we might as well just stick to what we currently have.

> could you point me towards an illustrative example of where this kind of separation is introduced and maintained? I am not sure what the conventions are around how we deal with implementing Conv.S for a type which has been decoupled from its representation in the dune lang.

All the examples I can think of are complex and are therefore not illustrative. Both actions and dependencies come to mind though. Their representation in dune files and in the engine are completely different. Here, I think we can come up with something far simpler:

```ocaml
module Lib_kind : sig
  module Dune_file : sig
    type t = Rewriter | Deriver | Normal
     include Conv.S (* This is used in the dune file *)
  end
  type t = Dune_file of Dune_file.t | Parameter | Virtual
  include Conv.S (* this is used in the dune-package file *)
end
```

Taking advantage of the fact that both `(kind parameter)` and `(kind virtual)` should be rejected.

I'd even reject `(kind normal)` if it I knew it wouldn't break any existing users. If this was ever allowed, it was only done by accident.

src/dune_rules/gen_meta.ml
rgrinberg: Moving this to a separate PR would be preferable. There's enough things going on in this PR.

It could be done after this PR is merged, or be extracted into an independent PR.

src/dune_rules/stanzas/library.ml
shonfeder: Thanks! Fixed :)

src/dune_rules/gen_meta.ml
shonfeder: I pulled this out.

src/dune_rules/stanzas/library.ml
shonfeder: I've done this!

Notably, however

> the fact that ... `(kind parameter)` ... should be rejected.

does not hold, since `(kind parameter)` is in fact valid, and is also the only way to declare that a library is a parameter (as per the plan for parametric libraries and discussion with @art-w).

shonfeder:
This should be ready for a final review pass. Thank you for the input @Alizter and @rgrinberg! It was very helpful :)

src/dune_rules/stanzas/library.ml
rgrinberg: No, that's not right. Library parameters are defined using the library_parameter stanza. You can fix this issue separately though. It deserves a test anyway.

shonfeder: Ah, I see that now! This was a misunderstanding/miscommunication on my part. Thanks. I'll fix the representation here, to pull this out of the `Lib_kind.t` and avoid incorrect logic in the parser. But maybe adding tests can be deferred to a followup.

shonfeder: I've fixed this.

shonfeder:
This is ready for review again :)

src/dune_lang/lib_kind.mli
```diff
@@ -13,11 +13,33 @@ module Ppx_args : sig
...
+  type t =
+    | Normal
+    | Ppx_deriver of Ppx_args.t
+    | Ppx_rewriter of Ppx_args.t
+
+  (** [cstr_name k] is the name of the kind constructor for [k] in dune lang.
+      E.g., [cstr_name Normal = "normal"] *)
+  val cstr_name : t -> string
+
+  val equal : t -> t -> bool
```
rgrinberg: Is this used anywhere? If not, let's get rid of it as the implementation relies on porymorphic equality

rgrinberg (approved):

shonfeder:
Thanks for the review and patient guidance!

PR #12308, opened 2025-08-26
test: showcase impact of cache disabled

Khady:
Adding a test to show the different behaviors of `dune build --cache=disabled` depending on how the shared cache is setup.

### DIFF
--- a/test/blackbox-tests/test-cases/dune-cache/cache-disabled.t
+++ b/test/blackbox-tests/test-cases/dune-cache/cache-disabled.t
@@ -1,0 +1,130 @@
+Set up project and cache directory
+
+  $ echo "(lang dune 3.20)" > dune-project
+ 
+  $ cat > config <<EOF
+  > (lang dune 3.20)
+  > (cache enabled)
+  > EOF
+
+  $ cat > dune << EOF
+  > (library
+  >  (name foo))
+  > EOF
+
+  $ cat > foo.ml << EOF
+  > let f x y = x + y
+  > EOF
+
+  $ export DUNE_CACHE_ROOT=$(pwd)/dune_test_cache
+  $ mkdir $DUNE_CACHE_ROOT
+
+Initial build, populating cache
+
+  $ dune build --config-file config
+  $ wc -l _build/log 
+  31 _build/log
+  $ ls $DUNE_CACHE_ROOT
+  files
+  meta
+  temp
+  values
+
+Second build, no-op as cache is warm
+
+  $ dune clean --config-file config
+  $ dune build --config-file config
+  $ wc -l _build/log 
+  17 _build/log
+
+Build with --cache=disabled, which should be a no-op as _build is already populated
+
+  $ dune build --cache=disabled --config-file config
+  $ wc -l _build/log 
+  22 _build/log
+
+Second build with --cache=disabled, should be the same
+
+  $ dune build --cache=disabled --config-file config
+  $ wc -l _build/log 
+  17 _build/log
+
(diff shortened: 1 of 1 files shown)


test/blackbox-tests/test-cases/dune-cache/cache-disabled.t
```diff
@@ -0,0 +1,130 @@
...
+  > let f x y = x + y
+  > EOF
+
+  $ export DUNE_CACHE_ROOT=$(pwd)/dune_test_cache
+  $ mkdir $DUNE_CACHE_ROOT
+
+Initial build, populating cache
+
+  $ dune build --config-file config
+  $ wc -l _build/log
```
Khady, line 25: I would have preferred to `cat _build/log` but some lines are starting with `$` and it confuses the cram tests on subsequent runs.

test/blackbox-tests/test-cases/dune-cache/cache-disabled.t
```diff
@@ -0,0 +1,130 @@
...
+  $ dune clean --config-file config
+  $ dune build --config-file config
+  $ wc -l _build/log 
+  17 _build/log
+
+Build with --cache=disabled, which should be a no-op as _build is already populated
+
+  $ dune build --cache=disabled --config-file config
+  $ wc -l _build/log 
+  22 _build/log
```
Khady, line 44: given that _build is still populated from the previous build, why do we have more stuff being built?

test/blackbox-tests/test-cases/dune-cache/cache-disabled.t
```diff
@@ -0,0 +1,130 @@
...
+
+Second build with --cache=disabled, should be the same
+
+  $ dune build --cache=disabled --config-file config
+  $ wc -l _build/log 
+  17 _build/log
```
Khady, line 58: here is it the same as during the second build, as expected

test/blackbox-tests/test-cases/dune-cache/cache-disabled.t
```diff
@@ -0,0 +1,130 @@
...
+  meta
+  temp
+  values
+
+Second build, no-op as cache is warm
+
+  $ dune clean --config-file config
+  $ dune build --config-file config
+  $ wc -l _build/log 
+  22 _build/log
```
Khady, line 86: shared cache is warm but more stuff is built?

test/blackbox-tests/test-cases/dune-cache/cache-disabled.t
```diff
@@ -0,0 +1,130 @@
...
```
Khady, line 134: given that _build is still populated from the previous build, why do we have more stuff being built?

rgrinberg:
Whether an action is decided to be re-executed depends on whether the cache is set. It is done this way so that running a build with the cache enabled will always populate the cache if possible (the cache is populated after the cache is re-ran).

So all of this is expected behavior. It might be suboptimal, but dune isn't really meant to be used with the cache toggled on and off. I think there's already a test for this behavior, do you mind checking the tests to confirm? I don't want to add more duplication to the test suite.

Khady:
I think I don’t understand what “the cache is set” means then. Because to me as a user there’s no obvious reason for the things to happen this way, and especially to have different behavior with `cache=disabled` depending on the shared cache being on for user rules or not. In all cases _build already contains the final artifacts.

rgrinberg:
Could you add a DCO please?

Khady:
added dco

Issue #12244, opened 2025-08-21
pkg: dune rebuilds dependent packages unpredictably


v-gb:
When running `dune build` or `dune tools exec ocamllsp`, sometimes dune decides to rebuild all the package dependencies, which obviously takes ages (and might prevent you from working if the network is not currently available).

For concrete examples, yesterday I had `dune tools exec ocamllsp` running, C-c, reran it, and it rebuilt all the dependencies. Doing it again seems to cache properly that time.
Just now, I was wondering why `dune build` was taking 30s+ to do a null build in a minuscule repo (which was in fact a null build, there was some crazy levenshtein_distance calls while reading the lock directory, which I still need to figure out). Once the first build stopped, or maybe I stopped it, I reran `dune build` with profiling, same thing. I then grabbed the latest dune preview build, reran `dune build`, and it rebuilt all the dependencies??
And I've seen multiple times in the past as well, where nothing relevant should have changed but dune decided to rebuild everything.

I don't know if that's enough to understand what's happening. Maybe dune doesn't like being C-c, but that'd be silly since that's the only way to stop a polling build.
I think dune should write down why it's taking these decisions, since they are so costly. I don't know if it already does with --trace, but if it does, then I claim this should be output by default: it's too late to set a debug flag after the problem has happened.

```
$ dune --version
"Dune Developer Preview: build 2025-08-20T02:28:50Z, git revision
24ca148234aa4eab93dff79573ed47e79f9d7694"
```
On Ubuntu 24.04.3 LTS

Alizter:
Yes, we've been observing some of these issues too. Thanks for filing a report.

Our main issue is that when there is a rebuild we have very little idea *why* it is rebuilding, which makes trying to fix this issue even more difficult.

We definitely need to add some sort tracing mechanism that reports why package rebuilds are happening. The technical reason is that whatever we are digesting is changing, so obviously it could be that we are digesting the wrong thing. The issue here is that we can't un-digest what we have so, its difficult to see how we can say "this file" has changed.

I'm assuming you are talking about the ocaml-compiler rebuilding, since that is typically what causes these expensive rebuilds. Unfortunately, due to the (current) non-relocatability of the OCaml compiler, we've had to implement some special support which might not be up-to-par.

One idea might be to dump the dyn of the dependencies of the OCaml compiler somewhere in the build directory. When we end up rebuilding it, we could then diff it with the old one.

[...]

PR #12310, opened 2025-08-26
Avoid using Poly.hash on Set and Map in pkg_rules

gridbugs:
If two values are equal then their hashes must also be equal. `Hashtbl.hash` (or `Poly.hash` as its known in dune) can compute a hash for an arbitrary value by traversing its representation in memory, the so-called "structure" of the value. Thus if two values differ in their structure, their hashes will likely also differ.

Some OCaml types have a notion of equality which is looser than structural equality. That is, two values may be considered equal even if their representations differ. Common examples are Sets and Maps, which are implemented by self-balancing binary trees. The representation in memory of such data structures is dependent on the order in which values are inserted into them, however their equality is based entirely upon their contents. Using Sets of ints as an example `(add 1 (add 2 empty))` may be structurally different from `(add 2 (add 1 empty))` even though these values are considered to be equal.

Thus it is incorrect to use `Poly.hash` to compute the hash of a Set or Map.

This change updates the hash computation for the `Pkg_rules.DB.t` type to avoid the use of `Poly.hash` as well as several other types which comprise it.

Possibly related to https://github.com/ocaml/dune/issues/12244

### DIFF
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -19,6 +19,8 @@ module Solver_env_disjunction = struct
     List.equal Solver_env.equal a b
   ;;
 
+  let hash t = List.hash Solver_env.hash t
+
   let encode t =
     let open Encoder in
     list sexp (List.map ~f:Solver_env.encode t)
@@ -47,6 +49,10 @@ module Conditional = struct
 
   let equal value_equal { condition; value } t =
     Solver_env_disjunction.equal condition t.condition && value_equal value t.value
+  ;;
+
+  let hash { condition; value } ~f =
+    Tuple.T2.hash Solver_env_disjunction.hash f (condition, value)
   ;;
 
   let to_dyn value_to_dyn { condition; value } =
@@ -88,6 +94,7 @@ module Conditional_choice = struct
      codebase and can be removed when portable lockdirs is the only option. *)
   let singleton_all_platforms value = singleton Solver_env.empty value
   let equal value_equal = List.equal (Conditional.equal value_equal)
+  let hash t ~f = List.hash (Conditional.hash ~f) t
   let map ~f = List.map ~f:(Conditional.map ~f)
   let to_dyn value_to_dyn = Dyn.list (Conditional.to_dyn value_to_dyn)
 
@@ -172,6 +179,16 @@ module Pkg_info = struct
          (Tuple.T2.equal Path.Local.equal Source.equal)
          extra_sources
          t.extra_sources
+  ;;
+
+  let hash { name; version; dev; avoid; source; extra_sources } =
+    Poly.hash
+      ( Package_name.hash name
+      , Package_version.hash version
+      , Bool.hash dev
+      , Bool.hash avoid
+      , Option.hash Source.hash source
+      , List.hash (Tuple.T2.hash Path.Local.hash Source.hash) extra_sources )
   ;;
 
   let remove_locs t =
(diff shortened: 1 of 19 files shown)


gridbugs:
Thanks to @v-gb for pointing this out!

src/dune_pkg/lock_dir.ml
```diff
@@ -512,6 +522,30 @@ module Pkg = struct
...
+        ; depexts
+        ; info
+        ; exported_env
+        ; enabled_on_platforms
+        }
+    =
+    (* XXX(steve): This assumes that each of these fields besides [info] is
+       has a notion of equality which is consistent with structural equality.
+       Otherwise two equal-but-not-structurally-equal values may have
+       different hashes. *)
```
v-gb: AFAICS, this type contains more maps (`Conditional_choice.t` -> `Conditional.t` -> `Solver_env_disjunction.t` -> `Solver_env.t` -> `Map`).

I don't know anything about this type, but maybe you should only hash the info field (using `[@@deriving_inline equal, hash]` would be simpler, but I guess you don't do that in dune). Considering the info field contains package name and version, that seems like that should be effectively unique already, cheaper to hash, and more likely to be correct than the current version.

Even if it might be possible to have marginally more hash collisions, Hashtbl.hash stops hashing after a few nodes (10 "meaningful" nodes or something?), so it might be that, before this change, the hash functions hashed nothing meaningful (by wasting meaningful nodes on `build_command` and `install_command` , which I assume are identical across the vast majority of packages here).

v-gb:
> Possibly related to https://github.com/ocaml/dune/issues/12244

I know roughly nothing about dune's codebase, but it doesn't seem super likely that hashes would be persisted. I was more thinking that, if dune uses marshal carelessly like it does polymorphic hash, that could plausibly cause the kind of unpredictable behavior from that issue.
In fact, grepping right now for `Marshal` shows two occurrences, and the one in `Dune_digest.generic` is a footgun (innocent name despite a very dicey behavior), and grepping for that function shows a handful of calls, including one in `src/dune_rules/pkg_rules.ml`, on values that include maps. So that could in fact be the cause for #12244. Regardless of whether it is the cause of that bug, the code is almost certainly buggy.
It seems like people are too keen to use polymorphic functions: with these functions, it if types, that says little about whether it works. I'd be less error prone if dune stopped using polymorphic function in favor of generated code, or if the dune codebase had a convention to force authors and reviewers to think whether the function makes sense for a given type (whether it's a longer name, or markers like a `type 'a structurally_digestable = private unit` so `Dune_digest.generic` can have type `'a structurally_digestable -> 'a -> string`).

rgrinberg (approved):

rgrinberg:
Here's the fix for the map in pkg_rules https://github.com/ocaml/dune/pull/12319

We had some plans to switch to bin_prot and start checking generated ppx, but the work has stalled. We hope to pick it up eventually.

src/dune_pkg/lock_dir.ml
gridbugs: Ah good catch with the Map in `Solver_env.t`. Will update.

src/dune_pkg/lock_dir.ml
```diff
@@ -512,6 +529,26 @@ module Pkg = struct
...
+        }
+    =
+    Poly.hash
+      ( Conditional_choice.hash ~f:Poly.hash build_command
+      , Conditional_choice.hash ~f:Poly.hash install_command
+      , Conditional_choice.hash ~f:Poly.hash depends
+      , depexts
+      , Pkg_info.hash info
+      , exported_env
+      , Solver_env_disjunction.hash enabled_on_platforms )
```
v-gb, line 549: The build_command and install_command are Action.t, whose comparison is Poly.equal. But both `Action.t` and other fields here contain `String_with_vars.t`, and the comparison for that type is partial (ignores some parts of the type), and calls into `User_message.compare/equal`, which is also partial, and which calls Style.compare, which is also partial. `User_message.compare` also calls `Poly.compare` on `User_message.Annots.t`, which is a `Univ_map.t`, which is an `Int.Map.t`.

I don't know how likely it is for this map to be non-empty, and how likely it is for the comparisons to skip over data that differs, and the fact that `Hashtbl.hash` doesn't go infinitely far down into structures could hide these issues are well, but in principle, both Action.equal and the hash of multiple fields are incorrect.

So this is better, but it's not systematic, so unsurprisingly there are more issues.

v-gb: Oh, and also I saw that `Stdune.Env.Var.equal` is case insensitive on windows, so that type can't be `Poly.hash`ed nor `Poly.equal`ed either, and I think it is.

gridbugs: Thanks for the thorough analysis. I'm going to merge this PR in its current state since I think it's likely to prevent some spurious rebuilds without systematically fixing all of our incorrect uses of `Poly.hash`.

Issue #12311, opened 2025-08-26
When depexts are missing and the combine length of the names of all depexs is sufficiently short, there is no newline between the error message and the list of depexts


gridbugs:
To reproduce, set up a project depending on a package with depexts and mess with that package's lockfile so it doesn't build. Dune will give a hint that the failure might be depext-related and print:
```
Hint: You may want to verify the following depexts are installed:
- depext1
- depext2
```

However if the combined length of depext names is too short it prints this instead:
```
Hint: You may want to verify the following depexts are installed:- a
                                                                 - b
```

PR #12312, opened 2025-08-26
pkg: Force breaks between depext hint and depext list

gridbugs:
Fixes https://github.com/ocaml/dune/issues/12311


test/blackbox-tests/test-cases/pkg/depexts/error-message.t
```diff
@@ -76,3 +77,58 @@ when the program is not found.
   Hint: You may want to verify the following depexts are installed:
   - unknown-package
   [1]
+
+Update the foo lockfile to have a single depext names:
```
maiste:
```suggestion
Update the foo lockfile to have a single depext name:
```

test/blackbox-tests/test-cases/pkg/depexts/error-message.t
```diff
@@ -76,3 +77,58 @@ when the program is not found.
...
+error message.
+  $ dune build
+  File "dune.lock/foo.pkg", line 3, characters 6-10:
+  3 |  (run dune build))
+            ^^^^
+  Error: Logs for package foo
+  File "dune-project", line 1, characters 0-0:
+  Error: Invalid first line, expected: (lang <lang> <version>)
+  
+  Hint: You may want to verify the following depexts are installed:
```
maiste, line 103: Shouldn't this be updated too with the new line?

test/blackbox-tests/test-cases/pkg/depexts/error-message.t
```diff
@@ -76,3 +77,58 @@ when the program is not found.
...
```
maiste, line 130: Shouldn't this be updated too with the new line?

test/blackbox-tests/test-cases/pkg/depexts/error-message.t
gridbugs: Ideally yes but I haven't figured out how to express this with `Pp`. The algorithm that `Pp` uses for formatting breaks lines in such a way that it breaks between `Hint:` and the message depending on the length of the message. In this case it splits if the concatenation of all depext names is longer than some threshold even though the depext names are in a vbox so their total width shouldn't matter (only their maximum width).

It might be a bug in Pp though I don't completely understand Pp.

src/dune_rules/run_with_path.ml
```diff
@@ -8,6 +8,7 @@ let depexts_hint = function
     ; Pp.enumerate ~f:Pp.verbatim depexts
     ]
     |> Pp.concat_map ~sep:Pp.cut ~f:(fun pp -> Pp.box pp)
```
Alizter, line 10: This should be `~f:Fun.id` and you should put `Pp.hovbox` on the first line and `Pp.vbox` on the second. Then you should do a `Pp.vbox` after concatenating. I think this will give you what you want.

gridbugs: I assume you mean:
```ocaml
let depexts_hint = function
  | [] -> None
  | depexts ->
    [ Pp.hovbox @@ Pp.textf "You may want to verify the following depexts are installed:"
    ; Pp.vbox @@ Pp.enumerate ~f:Pp.verbatim depexts
    ]
    |> Pp.concat_map ~sep:Pp.cut ~f:Fun.id
    |> Pp.vbox
    |> Option.some
;;
```

I tried that and it had the same behaviour as the current state of this PR.

rgrinberg:

PR #12315, opened 2025-08-26
Fix compilation of copy_file_range under musl libc

### LINKED ISSUE #12313: Building dune static using `musl` triggers compilation error
While building the `ocaml-dune/binary-distribution`, I have encountered a regression when building the static version.

## Expected Behavior

It builds without error.

## Actual Behavior

```
> otherlibs/stdune/src/copyfile_stubs.c:91:43: error: unknown type name 'loff_t'; did you mean 'off_t'?
> otherlibs/stdune/src/copyfile_stubs.c:91:57: error: unknown type name 'loff_t'; did you mean 'off_t'?
> otherlibs/stdune/src/copyfile_stubs.c:93:8: error: unknown type name 'copy_file_range_t'
> otherlibs/stdune/src/copyfile_stubs.c:93:47: error: initialization of 'int' from 'void *' makes integer from pointer without a cast [-Wint-conversion]
> otherlibs/stdune/src/copyfile_stubs.c: In function 'dune_copy_file_range':
> otherlibs/stdune/src/copyfile_stubs.c:98:11: error: called object 'copy_file_range_fn' is not a function or function pointer
> otherlibs/stdune/src/copyfile_stubs.c:93:26: note: declared here
> otherlibs/stdune/src/copyfile_stubs.c: In function 'stdune_sendfile':
> otherlibs/stdune/src/copyfile_stubs.c:124:32: error: 'copy_file_range_t' undeclared (first use in this function); did you mean 'copy_file_range_fn'?
> otherlibs/stdune/src/copyfile_stubs.c:124:32: note: each undeclared identifier is reported only once for each function it appears in
> otherlibs/stdune/src/copyfile_stubs.c:124:50: error: expected ')' before 'dlsym'

```

It can be reduced to this error:

```
> otherlibs/stdune/src/copyfile_stubs.c:91:43: error: unknown type name 'loff_t'; did you mean 'off_t'?
```

## Reproduction

An example of the error can be found in https://github.com/ocaml-dune/binary-distribution/actions/runs/17226133735/job/48870893140

To reproduce, just run `nix build .#dune-static` (or `nix build .#dune-static-experimental` but they have the same root).

## Specifications

- Version of `dune` (output of `dune --version`): `main` with commit a88ec6550cc82b8864fc4b03908ae6bd2a15e440
- Operating system (distribution and version): Arch Linux

## Problem

The regression can be traced to https://github.com/ocaml/dune/pull/12074. Before this PR it works, after it generates the errors above.

From what I found about it, it seems to be related to the fact that `musl` defines `loff_t` type as a `cpp` macro instead of a `typedef`. Two solutions I have found are:
- Declaring manually `loff_t` if not on `GNU_SOURCE`,
- Use `off64_t` instead (see https://github.com/martijnvanbrummelen/nwipe/commit/d10dca3bed42096637fcb3fb9fa3f2bc9159700f).

Not having touch C for a while, I'd like to have input from more competent C people 😄 

cc @nojb @rgrinberg

nojb:
Hopefully this should suffice to fix #12313.

@maiste Do you have a way to test the branch to confirm?

Fixes #12313


maiste:
I have tried with the fix and `off64_t` and both seemed to fail 😟 However, replacing with `int64_t` seems to work.

nojb:
Thanks @maiste. Can you try again with the commit I just pushed?

maiste:
I have tested with both `loff_t` and `off64_t` and the problem seems to still occur.

nojb:
> I have tested with both `loff_t` and `off64_t` and the problem seems to still occur.

OK, thanks. I pushed a last commit that should fix it. As we don't actually use these arguments, I replaced them by `void *` to indicate this, to avoid having to choose a specific data type that may be wrong in some contexts.

maiste (approved):

PR #12319, opened 2025-08-26
fix(pkg): stop marshalling maps

rgrinberg:
Serialize the install cookie with a list instead.

Convert maps to lists before digesting.

### DIFF
--- a/src/dune_rules/pkg_rules.ml
+++ b/src/dune_rules/pkg_rules.ml
@@ -195,32 +195,42 @@ module Install_cookie = struct
      don't have to know anything about the installation procedure.
   *)
 
-  type t =
-    { files : Path.t list Section.Map.t
-    ; variables : Variable.t list
-    }
+  module Gen = struct
+    type 'files t =
+      { files : 'files
+      ; variables : Variable.t list
+      }
 
-  let to_dyn { files; variables } =
-    let open Dyn in
-    record
-      [ "files", Section.Map.to_dyn (list Path.to_dyn) files
-      ; "variables", list Variable.to_dyn variables
-      ]
-  ;;
+    let to_dyn f { files; variables } =
+      let open Dyn in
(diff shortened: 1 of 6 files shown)


v-gb:
This code looks unproblematic to me, but [this one](https://github.com/ocaml/dune/blob/8d9b43a38d680ab9de8ba0f9188b12cf4f8da296/src/dune_rules/pkg_rules.ml#L552-L564) looks problematic, because if I have two maps that are Map.equal but not Poly.equal:
- if I marshal them and unmarshal them, they're still Map.equal
- if I marshal them and digest the marshalled string, the marshalled value are unequal

rgrinberg:
Why is it not problematic? The underlying section map is populated from a list where the order depends on the result of `readdir`.

> if I marshal them and unmarshal them, they're still Map.equal

In this example, the marshalled map is stored in a file that is the target of a rule. It will be digested by the engine as raw bytes.

I fixed the other example you've pointed out. Thanks.

gridbugs (approved):

src/dune_rules/pkg_rules.ml
```diff
@@ -547,7 +557,7 @@ module Substitute = struct
         , Package.Name.Map.to_list_map depends ~f:(fun _ (m, p) -> m, paths p)
```
v-gb, line 564: I don't know whether you chose to not handle for some reason, but `artifacts` above is also a map.

v-gb:
> Why is it not problematic? The underlying section map is populated from a list where the order depends on the result of `readdir`.
> 
> > if I marshal them and unmarshal them, they're still Map.equal
> 
> In this example, the marshalled map is stored in a file that is the target of a rule. It will be digested by the engine as raw bytes.
> 

I assumed it was persistent state that dune writes/reads, like `_build/.digest-db`. As the target of a rule, I agree this it's wrong to marshal maps or anything else whose equality is not structural equality.

PR #12324, opened 2025-08-28
Fix runtest_alias for inline tests

vouillon:
Similar to #12242


Alizter (approved):

Alizter:
The windows job is unrelated, it started failing before this PR.

Alizter:
cc @maiste This will need to go in a `3.20.2`. Let's start collecting PRs for that. I have a few other fixes I am working on. We can wait a little longer for this one so that we can make sure we have most of the regressions caught this time.

PR #12325, opened 2025-08-28
Add Cygwin job + fix cygwin bootstrap

### LINKED ISSUE #12316: Dune 3.20.0 broken on Cygwin/Windows x86_64
## Expected Behavior
  Dune 3.20.0 should compile successfully on Cygwin/Windows x86_64 systems during the bootstrap
  process.

## Actual Behavior

  Dune 3.20.0 compilation fails during the bootstrap phase with linking errors related to
  unresolved Blake3 symbols:

  ** Cannot resolve symbols for blake3_dispatch.o:
  [...]
   blake3_compress_xof_avx512
   blake3_compress_xof_sse2
   blake3_compress_xof_sse41
   blake3_hash_many_avx2
   blake3_hash_many_avx512
   blake3_hash_many_sse2
   blake3_hash_many_sse41
   blake3_xof_many_avx512
  File "caml_startup", line 1:
  Error: Error during linking (exit code 2)

## Reproduction
 This occurs when trying to install any package that depends on Dune in a fresh opam environment
  on Cygwin/Windows:

  1. Initialize opam with a recent opam-repository commit: opam init --bare default
  git+https://github.com/ocaml/opam-repository.git#d87894f816bea883931937e050bb5be4f108e50a
  2. Create a switch with system OCaml: opam switch create default ocaml-system
  3. Try to install any package requiring Dune: opam install lwt 

see https://github.com/arozovyk/opam/actions/runs/17233978247/job/48894396963

## Specifications

- Version of `dune` 3.20.0 (from opam-repository#d87894f816bea883931937e050bb5be4f108e50a)
- Version of `ocaml` 4.14.2
- Operating system (distribution and version): Cygwin on Windows x86_64

related to https://github.com/ocaml/dune/pull/11735 (cygwin is not tested by the CI)

Alizter:
In this PR we add a small CI job for testing the bootstrap process for Cygwin.

We also fix the bootstrap process on Cygwin. We have to use the portable blake3 implementation since the assembly causes segfaults and I don't think it's worth finding out why.

- Fixes https://github.com/ocaml/dune/issues/12316

cc @maiste for inclusion in 3.20.2

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -1012,10 +1012,15 @@ module Library = struct
     ; asm_files : Source.asm_file list
     }
 
-  let keep_asm { File_kind.syntax; arch; os; assembler = _ } ~ccomp_type ~architecture =
+  let keep_asm
+        { File_kind.syntax; arch; os; assembler = _ }
+        ~ccomp_type
+        ~architecture
+        ~os_type
+    =
     (match os with
-     | Some `Unix -> Sys.unix
-     | Some `Win -> Sys.win32
+     | Some `Unix -> String.equal os_type "Unix"
+     | Some `Win -> String.equal os_type "Win32"
      | None -> true)
     && (match syntax, ccomp_type with
         | `Intel, "msvc" -> true
@@ -1082,6 +1087,7 @@ module Library = struct
         ~ccomp_type
         ~architecture
         ~word_size
+        ~os_type
     =
     let scan_subdirs =
       match include_subdirs with
@@ -1165,7 +1171,7 @@ module Library = struct
             let extra_flags =
               if String.starts_with ~prefix:"blake3_" fn
               then
-                if Sys.cygwin || String.equal word_size "32"
+                if String.equal os_type "Cygwin" || String.equal word_size "32"
                 then
                   [ "-DBLAKE3_NO_SSE2"
                   ; "-DBLAKE3_NO_SSE41"
@@ -1180,7 +1186,7 @@ module Library = struct
         | Ml _ -> `Middle fn
         | Header -> `Skip
         | Asm asm ->
-          if keep_asm asm ~ccomp_type ~architecture
+          if keep_asm asm ~ccomp_type ~architecture ~os_type
           then (
             let out_file = Filename.chop_extension fn ^ ext_obj in
             `Right
@@ -1306,6 +1312,7 @@ let assemble_libraries
       ~ccomp_type
       ~architecture
       ~word_size
+      ~os_type
   =
   (* In order to assemble all the sources in one place, the executables
        modules are also put in a namespace *)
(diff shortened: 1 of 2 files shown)


boot/duneboot.ml
```diff
@@ -1164,8 +1164,8 @@ module Library = struct
             let extra_flags =
               if String.starts_with ~prefix:"blake3_" fn
               then (
-                match architecture with
-                | "x86" | "i386" | "i486" | "i586" | "i686" ->
+                match Sys.cygwin, architecture with
```
rgrinberg: Could you write this with an if statement instead?

Alizter: Done

boot/duneboot.ml
```diff
@@ -1164,14 +1164,19 @@ module Library = struct
             let extra_flags =
               if String.starts_with ~prefix:"blake3_" fn
               then (
-                match architecture with
-                | "x86" | "i386" | "i486" | "i586" | "i686" ->
+                let is_32_bit_arch =
+                  match architecture with
+                  | "x86" | "i386" | "i486" | "i586" | "i686" -> true
```
rgrinberg: Isn't there something like word size in the ocaml config?

Alizter: We have `Sys.word_size` which I've now used.

rgrinberg: You can't use that unfortunately. We can only use information from the ocaml config

boot/duneboot.ml
```diff
@@ -1163,15 +1163,15 @@ module Library = struct
           then (
             let extra_flags =
               if String.starts_with ~prefix:"blake3_" fn
-              then (
-                match architecture with
-                | "x86" | "i386" | "i486" | "i586" | "i686" ->
+              then
+                if Sys.cygwin || Sys.word_size == 32
```
rgrinberg: You should pass `word_size` to this function (like`ext_obj`) and obtain this value from the ocaml config.

rgrinberg (changes requested):

boot/duneboot.ml
Alizter: Done. Do you want to do the same for `os_type`? I can replace our other uses of `Sys` with that too.

rgrinberg: It's not strictly required, but I think it would be cleaner, yes.

The issue is that the bootstrap is building dune for a different version of OCaml than it's running when using a secondary compiler. This is like cross compilation, but the target/host OS is always the same (that is why it's not strictly required).

rgrinberg (approved):

boot/duneboot.ml
Alizter: IINM the `Sys.os_type` values are determined at compile-time and not run-time, so I don't think they would affect cross-compilation.

rgrinberg: You are right that it is determined at compile time, but the problem is that they are determined for the wrong OCaml here. We need these values for the OCaml that we are targeting, not the one executing the bootstrap.

PR #12329, opened 2025-08-28
fix(cram): make sure directory is empty during cram execution

### LINKED ISSUE #12321: dune 3.20.0: Cram tests now have hidden file `.cram.run.t`
## Expected Behavior
I have a cram test `run.t`:
```cram
List all files

  $ ls -a
  .
  ..

The directory should be empty save for current directory and parent directory
```
Then `dune runtest` runs the test which passes.

## Actual Behavior
The test fails:
```diff
File "test/run.t", line 1, characters 0-0:
diff --git a/_build/default/test/run.t b/_build/default/test/run.t.corrected
index e2b1fee..36596dc 100644
--- a/_build/default/test/run.t
+++ b/_build/default/test/run.t.corrected
@@ -3,5 +3,6 @@ List all files
   $ ls -a
   .
   ..
+  .cram.run.t
 
 The directory should be empty save for current directory and parent directory
```

## Reproduction

The bug can be reproduced with the above cram test

## Specifications

- Version of `dune` (output of `dune --version`): 3.20.0 (not a problem with 3.19.1!)
- Version of `ocaml` (output of `ocamlc --version`): 5.2.1
- Operating system (distribution and version): debian 13

## Additional information 

I suspect this PR introduced the bug: https://github.com/ocaml/dune/pull/11994

Alizter:
- fix https://github.com/ocaml/dune/issues/12321
- [x] changelog

### DIFF
--- a/src/dune_rules/cram/cram_exec.ml
+++ b/src/dune_rules/cram/cram_exec.ml
@@ -470,14 +470,19 @@ let run_and_produce_output ~src ~env ~dir:cwd ~script ~dst ~timeout =
   let lexbuf = Lexbuf.from_string script_contents ~fname:(Path.to_string script) in
   let temp_dir = make_temp_dir ~script in
   let cram_stanzas = cram_stanzas lexbuf in
-  Path.unlink_exn script;
+  (* We don't want the ".cram.run.t" dir around when executing the script. *)
+  Path.rm_rf (Path.parent_exn script);
   let env = make_run_env env ~temp_dir ~cwd in
   let open Fiber.O in
-  run_cram_test env ~src ~script ~cram_stanzas ~temp_dir ~cwd ~timeout
-  >>| List.filter_map ~f:(function
-    | Cram_lexer.Command c -> Some c
-    | Comment _ -> None)
-  >>| Script.dump (Path.build dst)
+  let+ commands =
+    run_cram_test env ~src ~script ~cram_stanzas ~temp_dir ~cwd ~timeout
+    >>| List.filter_map ~f:(function
+      | Cram_lexer.Command c -> Some c
+      | Comment _ -> None)
+  in
+  let dst = Path.build dst in
+  Path.mkdir_p (Path.parent_exn dst);
+  Script.dump dst commands
 ;;
 
 module Run = struct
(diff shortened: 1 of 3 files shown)


src/dune_util/persistent.ml
```diff
@@ -71,6 +71,7 @@ module Make (D : Desc) = struct
 
   let dump =
     let dump file v =
+      Path.mkdir_p (Path.parent_exn file);
```
rgrinberg: Can you move this to the caller? Seems undesirable to pay this penalty in every `dump` call.

rgrinberg (changes requested):
The fix is fine, but we don't want to a bunch of pointless `mkdir` everywhere

Alizter:
I've moved the mkdir to the caller.

src/dune_rules/cram/cram_exec.ml
```diff
@@ -443,14 +443,19 @@ let run_and_produce_output ~src ~env ~dir:cwd ~script ~dst ~timeout =
   let lexbuf = Lexbuf.from_string script_contents ~fname:(Path.to_string script) in
   let temp_dir = make_temp_dir ~script in
   let cram_stanzas = cram_stanzas lexbuf in
-  Path.unlink_exn script;
+  (* We don't want the ".cram.run.t" dir around when executing the script. *)
+  Path.rm_rf (Path.parent_exn script);
```
rgrinberg, line 474: Shouldn't we be clearing this directory rather than deleting it? This way we don't have to do the mkdir_p?

Alizter: If we clear the directory, then the empty directory will be present in the test. `Path.rm_rf` will already clear and then delete the directory.

The behaviour before is essentially what you said since we only have a single file, the `.cram.testname.t/script` file at this point.

Alizter: A strange behaviour is that this directory doesn't appear for directory cram tests, so perhaps we are misconfiguring a directory somewhere. I couldn't find it, so this is the solution I chose.

rgrinberg (approved):

Issue #9428, opened 2023-12-08
better locations in cram tests


Alizter:
When we get a differing cram test, the location we attach to the error should be the line where the first output changed.

This could probably be implemented in general for a diff action.

This would allow editors to jump directly to the differing output in the cram test.

rgrinberg:
Why is this specific to cram? I'd expect locations in any failed `diff` action.

Alizter:
@rgrinberg Yes, I mentioned that.

rgrinberg:
Okay, I removed the cram label.

> the location we attach to the error should be the line where the first output changed.

It's possible to attach multiple locations per error so let's just make use of that.

Alizter:
Interestingly, when using `patdiff` we configure it to output a location and there is a comment mentioning it should be picked up by editors. However the actual diagnostic we send via RPC doesn't contain the location information. This also appears to only be a single location.

I propose that we write a small hunk header lexer that will lex the hunk headers in the unified diff output from `diff -u` and report these as locations. We can then report these when displaying the error.

PR #12335, opened 2025-08-29
refactor(cram): track locations inside cram lexer

Alizter:
In this PR, we enrich the cram test lexer with location information. In order to do this, we need to modify how lines are being lexed so that we can increment the line count correctly.

The tests introduced in #12334 give us some guarantees that we are preserving behaviour. 

The motivation for this change is to allow for better error messages related to cram tests. For example, in a future PR we may enrich the timeout error with the exact location of the timed out command.

### DIFF
--- a/src/dune_rules/cram/cram_lexer.mll
+++ b/src/dune_rules/cram/cram_lexer.mll
@@ -1,67 +1,146 @@
 {
+open Import
+
 type 'command block =
   | Command of 'command
   | Comment of string list
+
+(* Creates location from lexbuf *)
+let loc lexbuf =
+  Loc.create ~start:(Lexing.lexeme_start_p lexbuf) ~stop:(Lexing.lexeme_end_p lexbuf)
+[@@inline]
+
+(* Adjusts the start position by an offset *)
+let adjust_start_cnum offset loc =
+  let start = Loc.start loc in
+  let start = { start with pos_cnum = start.pos_cnum + offset } in
+  Loc.set_start loc start
+[@@inline]
+
+(* Creates location for "  $ " pattern, adjusting past the "  $ " prefix *)
+let loc_of_dollar lexbuf =
+  loc lexbuf |> adjust_start_cnum 2
+[@@inline]
+
+(* Gets position before consuming newline *)
+let pos_before_newline lexbuf =
+  let pos = Lexing.lexeme_end_p lexbuf in
+  { pos with pos_cnum = pos.pos_cnum - 1 }
+[@@inline]
+
+(* Creates comment from string list accumulator *)
+let create_comment_from_acc loc acc =
+  match acc with
+  | [] -> None
+  | _ -> Some (loc, Comment (List.rev acc))
+[@@inline]
 }
 
-let eol = '\n' | eof
-
-let blank = [' ' '\t' '\r' '\012']
+let nonspace = [^' ' '\n']
+let not_nl  = [^'\n']
 
 rule block = parse
   | eof { None }
-  | "  $ " ([^'\n']* as str) eol
-    { Some (command_cont [str] lexbuf) }
-  | "  " [^'\n']* eol
-    { output [] lexbuf }
-  | ' '? as str eol
-    { comment [str] lexbuf }
-  | ' '? [^' ' '\n'] [^'\n']* as str eol
-    { comment [str] lexbuf }
+  | "  $ " ([^'\n']* as str)
+    { eol_then_command_or_finish (loc_of_dollar lexbuf) str lexbuf }
+  | "  > " ([^'\n']* as str)
+    { after_comment_line (loc lexbuf) [ "  > " ^ str ] lexbuf }
+  | "  >"
+    { after_comment_line (loc lexbuf) [ "  >" ] lexbuf }
+  | "  " [^'\n']*
+    { after_output_line (loc lexbuf) lexbuf }
+  | ' ' ((nonspace not_nl*) as rest)
+    { after_comment_line (loc lexbuf) [ " " ^ rest ] lexbuf }
+  | ' ' '\n'
+    { let loc = loc lexbuf in
+      let loc = Loc.set_stop loc (pos_before_newline lexbuf) in
+      Lexing.new_line lexbuf;
+      comment loc [ " " ] lexbuf }
+  | ' '
+    { comment (loc lexbuf) [ " " ] lexbuf }
+  | '\n'
+    { let loc = loc lexbuf in
+      let loc = Loc.set_stop loc (Loc.start loc) in
+      Lexing.new_line lexbuf;
+      comment loc [ "" ] lexbuf }
+  | nonspace not_nl* as str
+    { after_comment_line (loc lexbuf) [str] lexbuf }
 
-and comment acc = parse
+and after_comment_line loc content = parse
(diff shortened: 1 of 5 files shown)


rgrinberg:
Have you tried using these locations to improve any error messages yet? I'd like to see some work in that direction as this PR doesn't have any benefits on its own.

Alizter:
@rgrinberg Yes, I've added locations to both the timeout and conflict error messages here: https://github.com/ocaml/dune/pull/12657

Once this is merged I will PR those changes separately.

src/dune_rules/cram/cram_lexer.mll
```diff
@@ -1,67 +1,142 @@
...
+  | _ -> Some (Loc.create ~start ~stop, Comment (List.rev acc))
+
+(* Helper for extracting start/stop positions *)
+let with_span lexbuf f =
+  let start = Lexing.lexeme_start_p lexbuf in
+  let stop = Lexing.lexeme_end_p lexbuf in
+  f ~start ~stop
+
+(* Forward declarations for lexer rules *)
+let eol_fdecl = Fdecl.create Dyn.opaque
```
rgrinberg: Why can't you just write a mutually recursive function?

test/expect-tests/dune_rules/cram_parsing_tests.ml
```diff
@@ -4,49 +4,130 @@ open Dune_rules.For_tests.Cram_exec.For_tests
...
+    [ ({ pos_fname = "test"
+       ; start = { pos_lnum = 1; pos_bol = 0; pos_cnum = 2 }
+       ; stop = { pos_lnum = 1; pos_bol = 0; pos_cnum = 22 }
+       },
+       Command [ "echo 'Hello world'" ])
+    ]
+    |}];
   test "  $ echo 'Hello world'\n";
-  [%expect {| [ Command [ "echo 'Hello world'" ] ] |}]
+  [%expect
```
rgrinberg, line 50: The way the tests are currently written makes it really hard to tell if this is doing the right thing. How about using the machinery in `Loc` to actually show where the location is pointing to?

src/dune_rules/cram/cram_lexer.mll
Alizter: This is a limitation of ocamllex. It's not possible to mix mutually recursive functions between header and rule sections.

test/expect-tests/dune_rules/cram_parsing_tests.ml
Alizter: I've included location excerpts in the output. I will keep the location data around since it serves as a reference for the calculation.

src/dune_rules/cram/cram_lexer.mll
rgrinberg: I tihnk you can just consume no tokens in your rule for something equivalent.

src/dune_rules/cram/cram_lexer.mll
```diff
@@ -1,67 +1,142 @@
...
+let command_cont_fdecl = Fdecl.create Dyn.opaque
+
+(* Helper for position extraction + eol + recursive call *)
+let with_span_eol lexbuf content next_rule =
```
rgrinberg: Can't say I'm a fan of rewriting the lexer to use CPS. The dune_sexp lexer is far more complex and doesn't use CPS. Let's not introduce it here.

rgrinberg (changes requested):

src/dune_rules/cram/cram_lexer.mll
Alizter: Nice, I didn't know about that trick.

src/dune_rules/cram/cram_lexer.mll
Alizter: I've inlined these functions now. They were there to aid readability but they were obviously not very efficient as a result. I've gone for a slightly more verbose direct style.

src/dune_rules/cram/cram_lexer.mll
```diff
@@ -1,67 +1,150 @@
...
+
+and after_comment_line start stop content = parse
+  | '\n' { Lexing.new_line lexbuf; comment start stop content lexbuf }
+  | eof  { comment start stop content lexbuf }
+
+and after_output_line start = parse
+  | '\n' { Lexing.new_line lexbuf; output start [] lexbuf }
+  | eof  { output start [] lexbuf }
+
+and process_command_start start stop str = parse
```
rgrinberg: Why can't this rule and the one below it be combined?

src/dune_rules/cram/cram_lexer.mll
```diff
@@ -1,67 +1,150 @@
...
+  | "  $ " ([^'\n']* as str)
+    { let start =
+        let pos = Lexing.lexeme_start_p lexbuf in
+        { pos with pos_cnum = pos.pos_cnum + 2 }
+      in
+      let stop = Lexing.lexeme_end_p lexbuf in
+      process_command_start start stop str lexbuf }
+  | "  > " ([^'\n']* as str)
+    { let start = Lexing.lexeme_start_p lexbuf in
+      let stop = Lexing.lexeme_end_p lexbuf in
```
rgrinberg: Something to consider is to construct a `Loc.t` immediately here and in similar clauses and just pass that along. I'm not sure if it will work, but it has the potential to reduce a bunch of similar looking code.

rgrinberg (approved):
LGTM. The lexer is quite easy to understand now.

src/dune_rules/cram/cram_lexer.mll
Alizter: Right, that's what I had before but you complained about it. `with_span_eol` was the old name. I can reintroduce it, but maybe it's better for ocamllex for this to be explicit.

rgrinberg: Yes, because I dislike code that deals with continuations. That's not exactly what I'm suggesting though. I'm suggesting this instead:

1. Introduce a function `val loc : Lexing.lexbuf -> Loc.t` that does what your `let start` and `let stop` pair do.
2. Replace the `start`, and `stop` arguments with a `Loc.t` in every applicable rule.

src/dune_rules/cram/cram_lexer.mll
```diff
@@ -1,67 +1,145 @@
...
-    { comment [str] lexbuf }
-  | ' '? [^' ' '\n'] [^'\n']* as str eol
-    { comment [str] lexbuf }
-
-and comment acc = parse
+  | "  $ " ([^'\n']* as str)
+    { let loc = loc_of_lexbuf lexbuf in
+      let start = Loc.start loc in
+      let start = { start with pos_cnum = start.pos_cnum + 2 } in
+      let loc = Loc.set_start loc start in
```
rgrinberg: I'd factor out this location construction pattern into its own function. Doing the same arithmetic (the `+ 2`) in two places is always something that can drift. Moreover, it will eliminate the unnecessary (but likely harmless) intermediate location.

Alizter: I've factored out some more construction patterns (with no continuations) into their own helpers and marked them as `[@@inline]`.

shonfeder:
@Alizter is and all related work in service of https://github.com/ocaml/dune/issues/9428 ?

PR #12336, opened 2025-08-29
feat(oxcaml): add parameters field for libraries

### LINKED ISSUE #12086: [OxCaml] Parameterized libraries: parameterized library
This issue tracks the work related to the parameterization of libraries with parameterized libraries in OxCaml. This part requires #12084 to be completed first.

art-w:
This PR adds a new `(parameters ...)` field to the library stanza, to define a parameterized library (see https://github.com/ocaml/dune/pull/11963). Parameterized libraries are currently only available with the OxCaml compiler, since they require a new compiler flag `-parameter Name_of_parameter` (for each specified parameter) to compile the modules of the parameterized library.

Fixes https://github.com/ocaml/dune/issues/12086

### DIFF
--- a/src/dune_rules/lib.ml
+++ b/src/dune_rules/lib.ml
@@ -247,6 +247,30 @@ module Error = struct
           (Lib_name.to_string lib)
       ]
   ;;
+
+  let expected_parameter ~loc ~name =
+    make
+      ~loc
+      [ Pp.textf "Expected %S to be a library parameter." (Lib_name.to_string name) ]
+  ;;
+
+  let duplicate_parameters ~loc name name' =
+    make
+      ~loc
+      [ Pp.textf
+          "Duplicate library parameters: %S and %S."
+          (Lib_name.to_string name)
+          (Lib_name.to_string name')
+      ]
+  ;;
+
+  let missing_parameter ~loc p =
+    let name = Lib_name.to_string (Lib_info.name p) in
+    make_resolve
+      ~loc
+      [ Pp.textf "Parameter %S is missing." name ]
+      ~hints:[ Pp.textf "Add (parameters %s)" name ]
+  ;;
 end
 
 (* Types *)
@@ -337,6 +361,7 @@ module T = struct
     ; ppx_runtime_deps : t list Resolve.t
     ; pps : t list Resolve.t
     ; resolved_selects : Resolved_select.t list Resolve.t
+    ; parameters : t list Resolve.t
     ; implements : t Resolve.t option
     ; project : Dune_project.t option
     ; (* these fields cannot be forced until the library is instantiated *)
@@ -430,6 +455,7 @@ let name t = t.name
 let info t = t.info
 let project t = t.project
 let implements t = Option.map ~f:Memo.return t.implements
+let parameters t = Resolve.Memo.lift t.parameters
 let requires t = Memo.return t.requires
 let re_exports t = Memo.return t.re_exports
 let ppx_runtime_deps t = Memo.return t.ppx_runtime_deps
@@ -455,6 +481,21 @@ let main_module_name t =
      | This x -> x
      | From _ -> assert false)
 ;;
+
+module Parameterized = struct
+  let validate_required_parameters ~loc ~parameters lib =
+    let open Resolve.O in
+    let* lib = lib in
+    let* required_parameters = lib.parameters in
+    let+ () =
+      Resolve.List.iter required_parameters ~f:(function
+        | param when not (List.exists parameters ~f:(equal param)) ->
+          Error.missing_parameter ~loc param.info
+        | _ -> Resolve.return ())
+    in
+    lib
+  ;;
+end
 
 let wrapped t =
   match Lib_info.wrapped t.info with
@@ -861,6 +902,7 @@ module rec Resolve_names : sig
     :  db
     -> Lib_dep.t list
     -> private_deps:private_deps
+    -> parameters:t list
     -> pps:(Loc.t * Lib_name.t) list
     -> dune_version:Dune_lang.Syntax.Version.t option
     -> Resolved.t Memo.t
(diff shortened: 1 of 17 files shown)


src/dune_rules/module.ml
```diff
@@ -202,15 +202,24 @@ type t =
   ; visibility : Visibility.t
   ; kind : Kind.t
   ; install_as : Path.Local.t option
+  ; parameters : Module_name.t list
```
rgrinberg: Given that the list of parameters is constant for an entire stanza, keeping this information isn't the right place to do so. Could you try moving it to the compilation context instead?

rgrinberg (changes requested):

src/dune_rules/module.ml
art-w: Thanks, I've added a commit to that effect. The only advantage of storing the parameters for each module was that we could skip instantiating the generated modules since they don't depend on the parameters (and maybe we could have shaved more modules in the future by using ocamldep), but I now think it's easier to have a uniform treatment of all the modules :)

src/dune_rules/lib.ml
```diff
@@ -456,6 +482,21 @@ let main_module_name t =
      | From _ -> assert false)
 ;;
 
+module Parameterized = struct
+  let check ~loc ~parameters lib =
```
rgrinberg: How about a name like `validate_required_parameters`?

src/dune_rules/lib.ml
```diff
@@ -1485,15 +1563,22 @@ end = struct
       let open Resolve.Memo.O in
       let* resolved = Memo.return resolved in
       let* runtime_deps = runtime_deps in
-      re_exports_closure (resolved @ runtime_deps)
+      re_exports_closure (resolved @ runtime_deps @ parameters)
```
rgrinberg: `List.concat`

rgrinberg: Also, do we allow `re_exports` on parameters? I think we should.

rgrinberg (approved):
LGTM. Great work on getting on preserving the lazy loading everywhere.

src/dune_rules/lib.ml
art-w: What do you mean by "allow re_exports on parameters"? I've [added a test](https://github.com/ocaml/dune/pull/12336/commits/8fef3ae42f0ff3645379848c9073639b34e83915) to show that a parameter can re_export its dependencies, but I'm not sure if that's what you asked :)

rgrinberg: Yes, that's what I meant.

test/blackbox-tests/test-cases/oxcaml/library-field-parameters.t
```diff
@@ -0,0 +1,382 @@
...
+  $ echo 'let () = Lib.test ()' > bin/bin.ml
+  $ cat > bin/dune <<EOF
+  > (executable (name bin) (parameters a b) (libraries lib))
+  > EOF
+
+  $ dune build
+  File "bin/dune", line 1, characters 24-34:
+  1 | (executable (name bin) (parameters a b) (libraries lib))
+                              ^^^^^^^^^^
+  Error: Unknown field "parameters"
```
Alizter, line 320: This isn't blocking: I wonder if we can give a better message here. Something along the lines of:

```
Only the library stanza supports the parameters field.
```

This might complicate parsing unnecessarily however.

Also why can't executables have parameters?

test/blackbox-tests/test-cases/oxcaml/library-field-parameters.t
```diff
@@ -0,0 +1,382 @@
...
+  Hint: Add (parameters project.a)
+  [1]
+
+It works if `lib2` is itself parameterized with the same parameters as `lib`:
+
+  $ cat > lib2/dune <<EOF
+  > (library (name lib2) (parameters a b) (libraries lib))
+  > EOF
+
+A library can have more parameters than its dependencies:
```
Alizter, line 370: If we depend on a library that reexports a parameter and then also take that same parameter will we get a conflict?

Alizter:
This looks great! I've left a few questions but they are by no means blocking. I also have another: Is it feasible for `library_parameter`s to also support parameters themselves?

Alizter (approved):

test/blackbox-tests/test-cases/oxcaml/library-field-parameters.t
art-w: I added this test only to sanity-check the parser implementation, as there's some shared logic between libraries and executables stanza, but in practice it doesn't make sense to parameterize an executable because that makes it unrunnable! (unless you could later provide the parameters implementations in another stanza, but then that would be the same as defining a parameterized library and using it in a non-parameterized executable :) )
I agree the error message could be improved hm, but that does add some complexity... since this isn't blocking for you, I would rather do it in a follow-up PR :)

test/blackbox-tests/test-cases/oxcaml/library-field-parameters.t
art-w: Sorry I'm not sure I understand what it means to reexport a parameter? (the `re_export` keyword is not allowed on parameters) A parameterized library `x` can depend on other libraries which are themselves parameterized by a subset of the parameters of `x`. There's an incoming PR to specify the implementations of those parameters, which will allow instantiating dependencies (roughly equivalent to applying a functor to some arguments). Dependencies that are parameterized by `x` parameters will then transitively use the implementations of those parameters given to the instantiation of `x`.

art-w:
Thanks for the review and the fun question! I just checked, the OxCaml compiler doesn't currently support parameterizing a `library_parameter`:

```
Error: Cannot combine -as-parameter with -parameter: parameters cannot
       be parameterised.
```

test/blackbox-tests/test-cases/oxcaml/library-field-parameters.t
Alizter: Right, I had that the wrong way round. parameters themselves are allowed to re_export libraries but libraries cannot reexport their parameters as you said.

Alizter:
If you rebase I can merge it once the CI is done.

PR #12337, opened 2025-08-29
test(cram): make timeout test less racy + tweak formatting

Alizter:
Sometimes it fails on `echo` sometimes on `sleep`. This makes it more likely to fail on `sleep`.


rgrinberg:
Thanks. How about we just censor out the command?

Alizter:
Command censored, and I also fixed the boxes so that it doesn't break early.

PR #12341, opened 2025-08-29
refactor: remove spawn from dev builds

rgrinberg:
Just build against the dev version of spawn. Same as we've done for re. We should try to avoid drifting from the released version as much as possible.


Alizter:
I updated the format derivation in the nix flake.

.github/workflows/oxcaml.yml
```diff
@@ -29,7 +29,7 @@ jobs:
 
       - name: Install deps
         run: |
-          opam install . --deps-only
+          opam install . --deps-only --with-dev-setup
```
Alizter: This doesn't seem right? Why do you want `with-dev-setup` here?

rgrinberg: Isn't that where we define our test deps?

.github/workflows/oxcaml.yml
```diff
@@ -29,7 +29,7 @@ jobs:
 
       - name: Install deps
         run: |
-          opam install . --deps-only
+          opam install . --deps-only spawn
```
Alizter: Annoyingly, this will install the deps of spawn and not spawn itself. There is no way to do this in a single command AFAIK. Best is to do another `opam install spawn` straight after instead.

.github/workflows/oxcaml.yml
Alizter: I think that would be `--with-test`, but you don't need all of that just for oxcaml.

Alizter:
@rgrinberg this is fixed and good to go.

PR #12355, opened 2025-08-30
refactor(boot): remove re-implementation of string split

rgrinberg:
We can use 4.08 functions now

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -277,20 +277,9 @@ end
 
 module Bin = struct
   let path =
-    let path_sep = if Sys.win32 then ';' else ':' in
-    let split_path s =
-      let rec loop i j =
-        if j = String.length s
-        then [ String.sub s ~pos:i ~len:(j - i) ]
-        else if s.[j] = path_sep
-        then String.sub s ~pos:i ~len:(j - i) :: loop (j + 1) (j + 1)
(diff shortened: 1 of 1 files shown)


boot/duneboot.ml
```diff
@@ -277,20 +277,9 @@ end
...
-        else if s.[j] = path_sep
-        then String.sub s ~pos:i ~len:(j - i) :: loop (j + 1) (j + 1)
-        else loop i (j + 1)
-      in
-      loop 0 0
-    in
     match Sys.getenv_opt "PATH" with
     | None -> []
-    | Some s -> split_path s
+    | Some s -> String.split_on_char s ~sep:(if Sys.win32 then ';' else ':')
```
Alizter, line 282: We mentioned that `Sys.os_type` might not be the best here, since it might affect cross-comp. I had a look at threading it, but its too painful to be worth it.

rgrinberg: You can `ocaml_config` global, I don't particularly mind. We aren't going to lose any concurrency this way since we need this process to run before we execute anything anyway. Passing `os_type` around doesn't seem like a whole lot of work either anyway.,

rgrinberg: Wait, actually it would be wrong to use os_type from ocaml -config here. We need the `os_type` of the host rather than the target.

PR #12359, opened 2025-08-30
refactor(boot): use [List.iter] and |>

rgrinberg:


boot/duneboot.ml
```diff
@@ -769,16 +769,15 @@ module Build_info = struct
...
+      List.iter l ~f:(fun x ->
         pr "let %s =\n" name;
         pr "  [ ";
         f x;
         List.iter l ~f:(fun x ->
           pr "  ; ";
           f x);
-        pr "  ]\n"
+        pr "  ]\n");
+      pr "let %s = []\n" name
```
Alizter: Isn't this only meant to happen when the list is empty?

PR #12360, opened 2025-08-30
 fix: avoid killing process group when not applicable 

### LINKED ISSUE #12323: dune exec --watch broken for long-running processes
## Expected Behavior

For long-running processes like webservers, I'm normally starting ``dune exec --watch binary_name``. Whenever I change a source code file, dune kills the running process, recompiles it and starts it again. This works with dune ``3.19.1``.

## Actual Behavior

Since upgrading to dune ``3.20.1``, when changing the source code, the long-running process is not killed anymore. Recompilation only starts when the process exits on its own or I kill it manually.

## Reproduction

1. create a dune project with the following code in ``bin/main.ml``:

    ```
    let () =
      for i = 20 downto 0 do
        Unix.sleep 1;
        print_endline (string_of_int i);
      done
    ```

   and the following ``bin/dune``:

   ```
   (executable
     (public_name countdown)
     (name main)
     (libraries unix))
   ```

1. start ``dune exec --watch countdown``
1. change ``bin/main.ml``

The countdown continues until it reaches 0. Then recompilation starts and the process is started again.

## Specifications

- Version of `dune` (output of `dune --version`): 3.20.1
- Operating system (distribution and version): NixOS 25.05

Alizter:
We add a way of tracking whether a job is a process group leader to the
scheduler. This allows us to avoid sending a signal to a process group
when it isn't needed.

Doing so fixes an issue with dune exec -w where rebuilds would send a
kill signal to a non-existant process group assumed to be attached to
the process we were execing.

- [x] depends on #12361 
- [x] changelog
- fixes #12323

### DIFF
--- a/src/dune_scheduler/scheduler.ml
+++ b/src/dune_scheduler/scheduler.ml
@@ -112,9 +112,9 @@ let with_job_slot f =
     f t.cancel t.config)
 ;;
 
-let wait_for_process t pid =
+let wait_for_process t ~is_process_group_leader pid =
   let ivar = Fiber.Ivar.create () in
-  Process_watcher.register_job t.process_watcher { pid; ivar };
+  Process_watcher.register_job t.process_watcher { pid; is_process_group_leader; ivar };
   Fiber.Ivar.read ivar
 ;;
 
@@ -125,7 +125,7 @@ type termination_reason =
 
 (* We use this version privately in this module whenever we can pass the
    scheduler explicitly *)
-let wait_for_build_process t pid =
+let wait_for_build_process t ~is_process_group_leader pid =
   let+ res, outcome =
     Fiber.Cancel.with_handler
       t.cancel
@@ -133,10 +133,11 @@ let wait_for_build_process t pid =
         Process_watcher.killall t.process_watcher Sys.sigkill;
         Fiber.return ())
       (fun () ->
-         let+ r = wait_for_process t pid in
+         let+ r = wait_for_process t ~is_process_group_leader pid in
          (* [kill_process_group] on Windows only kills the pid and by this
-            time the process should've exited anyway *)
-         if not Sys.win32 then Process_watcher.kill_process_group pid Sys.sigterm;
+           time the process should've exited anyway *)
+         if not Sys.win32
+         then Process_watcher.kill_process_group pid Sys.sigterm ~is_process_group_leader;
          r)
   in
   ( res
@@ -589,11 +590,7 @@ let wait_for_process_with_timeout t pid waiter ~timeout ~is_process_group_leader
       Alarm_clock.await sleep
       >>| function
       | `Finished when Process_watcher.is_running t.process_watcher pid ->
-        let () =
-          if is_process_group_leader
-          then Process_watcher.kill_process_group pid Sys.sigkill
-          else Unix.kill (Pid.to_int pid) Sys.sigkill
-        in
+        Process_watcher.kill_process_group pid Sys.sigkill ~is_process_group_leader;
         Dune_trace.emit Process (fun () ->
           Dune_trace.Event.signal_sent
             Kill
@@ -601,7 +598,7 @@ let wait_for_process_with_timeout t pid waiter ~timeout ~is_process_group_leader
         `Timed_out
       | _ -> `Finished
     and+ res, termination_reason =
-      let+ res = waiter t pid in
+      let+ res = waiter t ~is_process_group_leader pid in
       Alarm_clock.cancel (Lazy.force t.alarm_clock) sleep;
       res
     in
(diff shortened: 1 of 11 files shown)


rgrinberg:
I don't think we should have any "fallbacks". If we are launching a process without a process group, we know we are doing that and there's no need to send a signal to the group

Alizter:
Sounds good to me. I will need to design a more principled solution that tracks the process group as you said. It might take some time before I can get round to it however.

Alizter:
:boom: I've messed up `killall` on Windows, the CI will not finished.

royneary:
May I ask what is missing to complete this fix? Is it just the CI failures for Windows systems?

Alizter:
@royneary Windows doesn't have a direct equivalent for unix process groups and the `spawn` library doesn't bother tracking them in any way. The fix here appears to cause a hang on Windows that will have to be investigated.

rgrinberg:
What's not clear to me is how are we changing the behavior on windows at all here? Since there's no process groups on windows, why isn't the old code not working as before?

The issues in CI seem unrelated btw.

Alizter:
@rgrinberg The issues in CI are not unrelated unfortunately. I can spin up Windows 11 in an emulator and bootstrap this branch and a normal dune build will never terminate. We are waiting for something that on windows never signals back it seems. I still don't fully understand the issue.

rgrinberg:
Is this relevant to this issue?

```
Run opam exec -- make release
ocamlc -output-complete-exe -intf-suffix .dummy -g -o .duneboot.exe -I boot -I +unix unix.cma boot/types.ml boot/libs.ml boot/duneboot.ml
.\.duneboot.exe
cd _boot && D:\a\dune\dune\_opam\bin\ocamlopt.opt.exe -c -g -I +unix -I +threads -ccopt -D_UNICODE -ccopt -DUNICODE mdb.c
vendor/ocaml-lmdb\mdb.c: In function 'mdb_strerror':
vendor/ocaml-lmdb\mdb.c:1809:1: warning: function may return address of local variable [-Wreturn-local-addr]
vendor/ocaml-lmdb\mdb.c:1770:7: note: declared here
```

Alizter:
> Is this relevant to this issue?
> 
> ```
> Run opam exec -- make release
> ocamlc -output-complete-exe -intf-suffix .dummy -g -o .duneboot.exe -I boot -I +unix unix.cma boot/types.ml boot/libs.ml boot/duneboot.ml
> .\.duneboot.exe
> cd _boot && D:\a\dune\dune\_opam\bin\ocamlopt.opt.exe -c -g -I +unix -I +threads -ccopt -D_UNICODE -ccopt -DUNICODE mdb.c
> vendor/ocaml-lmdb\mdb.c: In function 'mdb_strerror':
> vendor/ocaml-lmdb\mdb.c:1809:1: warning: function may return address of local variable [-Wreturn-local-addr]
> vendor/ocaml-lmdb\mdb.c:1770:7: note: declared here
> ```

Those are just warnings when compiled on Windows, the actual issue is the fact that the runner took 5 hours.

rgrinberg:
I see. Can you disable these warnings? There's not much we plan to do for vendored code.

Are you sure this PR is not changing the behavior on windows? It really should not.

Alizter:
> I see. Can you disable these warnings? There's not much we plan to do for vendored code.

I've opened #12750.

> Are you sure this PR is not changing the behavior on windows? It really should not.

Evidently it is changing the behaviour on Windows. I don't claim to understand it at the moment. I will leave this PR as a draft until I can get round to understanding what is happening on Windows and what I am doing wrong.

rgrinberg:

src/dune_scheduler/scheduler.ml
```diff
@@ -112,9 +112,9 @@ let with_job_slot f =
     f t.cancel t.config)
 ;;
 
-let wait_for_process t pid =
+let wait_for_process t ?(is_process_group_leader = true) pid =
```
rgrinberg: Can you make this argument required?

rgrinberg (changes requested):

rgrinberg (approved):

test/blackbox-tests/test-cases/exec-watch/exec-watch-server.t/run.t
```diff
@@ -18,7 +18,6 @@ Below, 0: after should *not* be appearing.
 
   $ dune exec --watch ./foo.exe &
   0: before
-  0: after
```
Alizter, line 21: @rgrinberg In case it wasn't clear. This test is still disabled and you have to explicitly enable it if you want to check the behaviour yourself. I don't know how to fix the flake at the moment, so I won't enable it generally.

PR #12367, opened 2025-08-30
refactor(boot): set module paths for generated modules

rgrinberg:


Alizter (approved):
I guess the path includes the name now?

rgrinberg:
> I guess the path includes the name now?

Yeah. This is so that we can compute the compilation unit name just with the path.

PR #12369, opened 2025-08-30
refactor(boot): modules trie

rgrinberg:
To support `(include_subdirs qualified)` in the bootstrap, we need to discover module groups for every directory and attach module paths to every module. Additionally, we need to generate alias files for every group that needs it.

This isn't all we need, but it's a first step and already works. More incoming in later PRs.

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -33,14 +33,192 @@ open Types
 
 (** {2 Utility functions} *)
 
+include struct
+  [@@@ocaml.warning "-32-34-37"]
+
+  module Either = struct
+    type ('l, 'r) t =
+      | Left of 'l
+      | Right of 'r
+  end
+end
+
+open Stdlib
 open StdLabels
 open MoreLabels
 open Printf
 
+module Option = struct
+  include Option
+
+  let iter t ~f = Option.iter f t
+  let map t ~f = Option.map f t
+  let bind t ~f = Option.bind t f
+end
+
+module Map = struct
+  module type S = sig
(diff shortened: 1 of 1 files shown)


boot/duneboot.ml
```diff
@@ -1167,7 +1185,12 @@ module Library = struct
         ~os_type
     =
     let files = scan ~dir ~include_subdirs in
-    let modules = modules files ~build_info_module ~root_module in
+    let modules =
+      modules files ~build_info_module ~root_module
+      |> Module_name.Map.to_list
```
Alizter: This gets introduced in 5.1. You probably want to use the old name `Map.bindings`.

boot/duneboot.ml
```diff
@@ -103,56 +203,118 @@ module Ml_kind = struct
...
-    |> String.capitalize_ascii
-    |> of_string
-  ;;
+  module Path = struct
+    module T = struct
+      type t = Name.t list
 
-  let alias_suffix s = s ^ "__"
-end
+      let compare = List.compare ~cmp:Name.compare
```
Alizter, line 352: You will probably need to use the stdlib one here
```
let rec compare cmp l1 l2 =
  match l1, l2 with
  | [], [] -> 0
  | [], _::_ -> -1
  | _::_, [] -> 1
  | a1::l1, a2::l2 ->
    let c = cmp a1 a2 in
    if c <> 0 then c
    else compare cmp l1 l2
```
since it was introduced in 4.12.

PR #12385, opened 2025-09-01
[3.20] backport #12325

maiste:
@Alizter, could you have a closer look to this one? With the divergence from the refactoring I add to update a bit the code. Feel free to directly push on my branch if you need to add extra changes.

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -1070,7 +1070,7 @@ module Library = struct
           if keep_c c ~architecture
           then (
             let extra_flags =
-              if String.starts_with ~prefix:"blake3_" fn
+              if String.is_prefix ~prefix:"blake3_" fn
               then
                 if String.equal os_type "Cygwin" || String.equal word_size "32"
                 then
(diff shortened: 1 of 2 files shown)


boot/duneboot.ml
```diff
@@ -1063,16 +1064,16 @@ module Library = struct
           if keep_c c ~architecture
           then (
             let extra_flags =
-              if String.is_prefix ~prefix:"blake3_" fn
-              then (
-                match architecture with
-                | "x86" | "i386" | "i486" | "i586" | "i686" ->
+              if String.starts_with ~prefix:"blake3_" fn
```
Alizter: If you revert this line, it should fix the CI on 4.08.

maiste:
The failure seems related to `lwt` here. Any chance you already encountered the problem?

Alizter:
@maiste Yes. There is a CI fix in https://github.com/ocaml/dune/pull/12327. Just backport that and it should be fine.

PR #12388, opened 2025-09-02
chore: changelog for 3.20.2

maiste:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,19 @@
+3.20.2 (2025-09-02)
+-------------------
+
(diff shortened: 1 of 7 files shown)


Alizter:
Looks good to me. Could you rewrap those so the raw file is a bit easier to read?

Alizter (approved):

maiste:
Not sure to understand? Do you want me to reformat `CHANGES.md`?

Alizter:
@maiste yes please. Only the new entries. We can reformat the rest on main later.

PR #12391, opened 2025-09-03
[Nitpicking] Adjust printing to use console, and Pp.paragraph where applicable in Rpc_common

ElectreAAS:
Extracted from #12064, as it has little to do with the core of it.
Pp.text can be broken on any space, while Pp.paragraph won't.
Other than that, we have these `User_message.Style.t`s, let's use them

Completely subjective PR


Alizter:
Regarding the placement of the failure count. Imagine you are a user and you've just seen several errors requiring you to scroll up in your terminal to see them all. If we print the count at the end the user can know before scrolling how many errors to expect, whereas if we print the count at the top, the user will have to scroll all the way too the top, which might be tricky to find, in order to see how many failures there are.

I'm not opposed to the other changes, but we should think a bit more carefully about the placement to make sure we aren't making dune harder to use than it already is.

shonfeder:
Could this include some tests showing how the output will be rendered after these changes to ensure it is working as expected?

test/blackbox-tests/test-cases/watching/multiple-errors-output.t/src/foo.ml
```diff
@@ -0,0 +1 @@
+let f = Bar.x + Baz.y + invalid_syntax : ? = !
```
shonfeder, line 1: I don't see stuff included in the error output in the tests. Is that expected?

test/blackbox-tests/test-cases/watching/multiple-errors-output.t/run.t
```diff
@@ -0,0 +1,32 @@
...
+  File "libs/bar.ml", line 1, characters 8-20:
+  1 | let y = "type error" + 3
+              ^^^^^^^^^^^^
+  Error: This constant has type string but an expression was expected of type
+           int
+  File "libs/baz.ml", line 1, characters 29-33:
+  1 | let y = "unknown variable" ^ what
+                                   ^^^^
+  Error: Unbound value what
+  Had 2 errors, waiting for filesystem changes...
```
shonfeder, line 32: Is it right that the error is reported again during shutdown? I guess this is just an artifact of our test setup?

shonfeder (approved):
Looks good to me, just one question about the tests.

test/blackbox-tests/test-cases/watching/multiple-errors-output.t/run.t
ElectreAAS: This is the expected behaviour. The first report after `dune build` is what the client sees, everything reported after `stop_dune` is what the server sees.

test/blackbox-tests/test-cases/watching/multiple-errors-output.t/src/foo.ml
ElectreAAS: Since foo depends on both bar and baz, and both of these can't compile, foo.ml is never processed, that is intended

Alizter:
The messages that were converted to paragraphs were not tested I believe. I'm fairly certain that putting a `paragraph` at the same level as `Error:` will cause it to break straight after, which is why we tend not to use `paragraph` in the first line of an error. Something like:

```
Error:
some paragraph
```

PR #12392, opened 2025-09-04
feat(oxcaml): support parameter implementation

### LINKED ISSUE #12085: [OxCaml] Parameterized libraries: parameter implementation

shonfeder:
Closes https://github.com/ocaml/dune/issues/12085
Supersedes https://github.com/ocaml/dune/pull/12002

This PR is the second of a set of four PRs related to the introduction of Parameterized Libraries in Dune.

:warning: This feature is only available with OxCaml until it is upstream to the compiler.

- It adds support for parameter in the `implements` field.
- It provides some tests to make sure the functionality act according to the specification. They are still guarded by the oxcaml variable.
- It extends the documentation with our case. More explanation about the feature will be addressed in future PRs, specific to the documentation.

## TODO

- [x] It needs the parameter to be merged first ✅ 
- [x] Complete https://github.com/ocaml/dune/issues/12253 and rebase on the merged changes 
- [x] Consider whether `is_parameter` should be removed (as per https://github.com/ocaml/dune/pull/12298#issuecomment-3218548129).


test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
```diff
@@ -0,0 +1,135 @@
...
+  >  (name foo_impl)
+  >  (implements missing_foo))
+  > EOF
+  $ dune build
+  File "foo_impl/dune", line 3, characters 13-24:
+  3 |  (implements missing_foo))
+                   ^^^^^^^^^^^
+  Error: Library "missing_foo" not found.
+  -> required by alias default
+  [1]
```
shonfeder: These two tests appear to be checking the exact same operations with the same inputs, and this doesn't make sense to me. @maiste has tried to explain this at https://github.com/ocaml/dune/pull/12002#discussion_r2300153552 , but I still do not how running the exact same test with two different descriptions can be useful :sweat_smile: .

I suspect the test needs to be fixed.

src/dune_rules/lib_rules.ml
```diff
@@ -487,6 +487,27 @@ let setup_build_archives (lib : Library.t) ~top_sorted_modules ~cctx ~expander ~
     (fun () -> build_shared ~native_archives ~sctx lib ~dir ~flags)
 ;;
 
+let parameterize_root_module ~parameter ~root_module_name modules =
```
shonfeder: Per https://github.com/ocaml/dune/pull/12002#discussion_r2248844715

> Instead of storing this information in the module, how about just add a field to the compilation context designating which module is implementing a parameter. Should be easier to understand and more efficient.

src/dune_rules/lib_rules.ml
```diff
@@ -487,6 +487,27 @@ let setup_build_archives (lib : Library.t) ~top_sorted_modules ~cctx ~expander ~
...
+  let* parameter_module_name =
+    let+ name = Lib.main_module_name parameter in
+    match Resolve.to_result name with
+    | Ok (Some name) -> name
+    | Ok None ->
+      (* The parameter must have a module name otherwise it can't exist. *)
+      assert false
+    | Error err -> Resolve.raise_error_with_stack_trace err
```
shonfeder: Per https://github.com/ocaml/dune/pull/12002#discussion_r2248845061

> This error needs to be delayed rather than forced.

shonfeder (commented):
Copying over the outstanding change requests from https://github.com/ocaml/dune/pull/12002

test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
maiste: Re-reading the test, it is indeed broken 😅 The first one should use an existing library which is neither a parameter nor a virtual lib. Both tests should exist, but in the current situation, they are doing the same thing. The first one should use an existing library. Sorry for not understanding fast enough, this 🤦🏻‍♂️

src/dune_rules/lib.ml
```diff
@@ -724,8 +725,9 @@ end = struct
         let rec loop acc = function
           | [] -> Resolve.Memo.return acc
           | (lib, stack) :: libs ->
+            let is_parameter = Lib_info.is_parameter lib.info in
             let virtual_ = Lib_info.virtual_ lib.info in
-            (match lib.implements, virtual_ with
+            (match lib.implements, virtual_ || is_parameter with
```
maiste: Shouldn't the match here be replaced with `lib.kind` instead of `is_parameter`?

src/dune_rules/lib.ml
```diff
@@ -947,9 +949,15 @@ end = struct
           let open Resolve.Memo.O in
           let* vlib = resolve_forbid_ignore name in
           let virtual_ = Lib_info.virtual_ vlib.info in
-          match virtual_ with
-          | false -> Error.not_virtual_lib ~loc ~impl:info ~not_vlib:vlib.info
-          | true -> Resolve.Memo.return vlib
+          let is_parameter = Lib_info.is_parameter vlib.info in
+          match virtual_, is_parameter with
```
maiste: Same question about the `kind` here?

src/dune_rules/virtual_rules.ml
```diff
@@ -79,7 +79,7 @@ let impl sctx ~(lib : Library.t) ~scope =
          ]
      | Some vlib ->
        let info = Lib.info vlib in
-       if not (Lib_info.virtual_ info)
+       if (not (Lib_info.virtual_ info)) && not (Lib_info.is_parameter info)
```
maiste: Isn't it possble to match on a `kind` here too?

test/blackbox-tests/test-cases/oxcaml/implements-parameter.t
art-w: This should be fixed, I've reworked the test to test both cases and more :)

src/dune_rules/lib.ml
art-w: Yes, done!

src/dune_rules/lib.ml
art-w: Done!

src/dune_rules/virtual_rules.ml
art-w: Thanks, done!

src/dune_rules/lib_rules.ml
art-w: Indeed! The error has been delayed until we really need it.

src/dune_rules/lib_rules.ml
art-w: Ok! the information that the root module implements a parameter is now passed via the compilation context, instead of being a per module field.

shonfeder:
Thanks for the fixes, @art-w! I'll let you take this from here :D 

IIUC, this is ready for review, so I will mark it as such.

src/dune_rules/ml_sources.ml
```diff
@@ -364,9 +364,11 @@ let make_lib_modules
...
       in
       let kind : Modules_field_evaluator.kind =
-        match lib.virtual_modules with
-        | None -> if Library.is_parameter lib then Parameter else Exe_or_normal_lib
-        | Some virtual_modules -> Virtual { virtual_modules }
+        match lib.kind, lib.virtual_modules with
+        | Dune_file _, None -> Exe_or_normal_lib
+        | Parameter, None -> Parameter
+        | Virtual, Some virtual_modules -> Virtual { virtual_modules }
+        | _ -> assert false
```
rgrinberg: Can you expand this pattern?

src/dune_rules/virtual_rules.ml
```diff
@@ -79,14 +79,15 @@ let impl sctx ~(lib : Library.t) ~scope =
...
-       if not (Lib_info.virtual_ info)
-       then
-         User_error.raise
-           ~loc:lib.buildable.loc
-           [ Pp.textf
-               "Library %s isn't virtual and cannot be implemented"
-               (Lib_name.to_string implements)
-           ];
+       (match Lib_info.kind info with
+        | Parameter | Virtual -> ()
```
rgrinberg: The Parameter case should be "assert false" I believe

src/dune_rules/lib.ml
```diff
@@ -946,10 +947,9 @@ end = struct
         let res =
           let open Resolve.Memo.O in
           let* vlib = resolve_forbid_ignore name in
-          let virtual_ = Lib_info.virtual_ vlib.info in
-          match virtual_ with
-          | false -> Error.not_virtual_lib ~loc ~impl:info ~not_vlib:vlib.info
-          | true -> Resolve.Memo.return vlib
+          match Lib_info.kind vlib.info with
+          | Dune_file _ -> Error.not_virtual_lib ~loc ~impl:info ~not_vlib:vlib.info
+          | Parameter | Virtual -> Resolve.Memo.return vlib
```
rgrinberg: `Parameter` cannot be accepted here.

rgrinberg (changes requested):

src/dune_rules/module.ml
```diff
@@ -288,7 +288,7 @@ let map_files t ~f =
 let src_dir t = Source.src_dir t.source
 let set_pp t pp = { t with pp }
 
-let to_dyn { source; obj_name; pp; visibility; kind; install_as } =
+let to_dyn { source; obj_name; pp; visibility; kind; install_as; _ } =
```
rgrinberg: This is unnecessary now

src/dune_rules/module.ml
```diff
@@ -345,7 +345,10 @@ module Obj_map = struct
     end)
 end
 
-let encode ({ source; obj_name; pp = _; visibility; kind; install_as = _ } as t) ~src_dir =
+let encode
+      ({ source; obj_name; pp = _; visibility; kind; install_as = _; _ } as t)
```
rgrinberg: Also unnecessary

src/dune_rules/compilation_context.mli
```diff
@@ -27,6 +27,7 @@ val create
   -> flags:Ocaml_flags.t
   -> requires_compile:Lib.t list Resolve.Memo.t
   -> requires_link:Lib.t list Resolve.t Memo.Lazy.t
+  -> ?implements_parameter:Module_name.t * Module_name.t option Resolve.Memo.t
```
rgrinberg: Can you turn this tuple into a record? I found myself back tracking to determine these two fields a few too many times.

src/dune_rules/install_rules.ml
```diff
@@ -97,9 +97,10 @@ end = struct
       >>| Modules.With_vlib.modules
       >>| Option.some
     and+ foreign_archives =
-      match Lib_info.virtual_ lib with
-      | false -> Memo.return (Mode.Map.Multi.to_flat_list @@ Lib_info.foreign_archives lib)
-      | true ->
+      match Lib_info.kind lib with
+      | Dune_file _ ->
+        Memo.return (Mode.Map.Multi.to_flat_list @@ Lib_info.foreign_archives lib)
+      | Parameter | Virtual ->
```
rgrinberg: This should be empty for `Parameter`. I don't think we allow stubs for parameters.

src/dune_rules/install_rules.ml
```diff
@@ -293,26 +293,27 @@ end = struct
...
-            then []
-            else
-              [ if_ native (Ocaml Cmx, cm_file (Ocaml Cmx))
-              ; if_ (byte && virtual_library) (Ocaml Cmo, cm_file (Ocaml Cmo))
-              ; if_
-                  (native && virtual_library)
-                  (Ocaml Cmx, Obj_dir.Module.o_file obj_dir m ~ext_obj)
-              ; if_ melange (Melange Cmi, cm_file (Melange Cmi))
-              ; if_ melange (Melange Cmj, cm_file (Melange Cmj))
+      let common m =
```
rgrinberg, line 297: Hard to understand from this diff, but I take that this is just a refactoring? I.e. all we've done is remove `is_parameter` from here?

rgrinberg (changes requested):
Just finished a round of review.

src/dune_rules/install_rules.ml
art-w: Yes exactly, and it also gets rid of the `is_virtual` in favor of matching the kind

src/dune_rules/ml_sources.ml
art-w: Sure!

src/dune_rules/module.ml
art-w: Oh thanks, I removed it!

src/dune_rules/module.ml
art-w: Thanks, also removed now

src/dune_rules/compilation_context.mli
art-w: Sure! I've added a record to the mli, but I'm happy to also add a function to hide its representation if you prefer

src/dune_rules/install_rules.ml
art-w: Oh right, thanks, done!

src/dune_rules/virtual_rules.ml
art-w: Hm in the current implementation the Parameter case is useful, as a library that has an `implements` still get turned into a "vlib" which associates the library with the (virtual or parameter) lib it implements. I suppose we could avoid reusing parts of the virtual logic however, with a new sum type above. WDYT?

src/dune_rules/lib.ml
art-w: At this point in the code we are trying to resolve the Lib_info fields into Lib.t values, so when we look at the `implements` we can actually get either a Virtual or a Parameter (since it's the same field in the library stanza, and we have no way of knowing what it'll be before we resolve the lib name with the database). Should we rename the `vlib` variable to avoid the confusion?

art-w:
Thanks for the quick review! I've marked the simpler issues as resolved by the last commit, but I've left open the ones where I would appreciate your input :)

src/dune_rules/virtual_rules.ml
rgrinberg: I think it's confusing indeed then. Turning it into a sum type (or a record with a `kind` field) would clarify this.

src/dune_rules/compilation_context.mli
rgrinberg: Makes no difference to me. I'd just want labels for these names.

src/dune_rules/lib.ml
rgrinberg: Renaming it would be useful. Thanks.

src/dune_rules/module_compilation.ml
```diff
@@ -238,6 +238,16 @@ let build_cm
...
    let as_parameter_arg = if Module.kind m = Parameter then [ "-as-parameter" ] else [] in
+   let as_argument_for =
+     Command.Args.dyn
+       (let open Action_builder.O in
+        let+ argument =
+          Resolve.Memo.read @@ Compilation_context.implements_parameter cctx m
+        in
+        match argument with
+        | None -> []
+        | Some parameter -> [ "-as-argument-for"; Module_name.to_string parameter ])
```
rgrinberg, line 248: Could you remind where we add a dependency on the cmi of the parameter? Is it added a library dependency somewhere?

art-w: It looks like the dependency to the parameter is added by the pre-existing code for vlibs:

https://github.com/ocaml/dune/blob/07d0a93e5558376cc27bddae1e126c9a9d6e6a68/src/dune_rules/lib.ml#L1025-L1034

src/dune_rules/lib.ml
art-w: Done, I've renamed it to `implements` :)

src/dune_rules/virtual_rules.ml
art-w: I pushed a refactoring of `Virtual_rules` to distinguish between the three `implements` possibilities: no implements, implementing a virtual module or implementing a parameter. I think `virtual_rules.ml` could be further renamed to `implements.ml`, if the name "virtual" has too much connotation? :)

src/dune_rules/virtual_rules.ml
```diff
@@ -1,72 +1,48 @@
 open Import
 open Memo.O
 
-let setup_copy_rules_for_impl ~sctx ~dir vimpl =
```
art-w, line 4: This was moved (unchanged) to `Vimpl.setup_copy_rules`: https://github.com/ocaml/dune/pull/12392/files#diff-654c4360f33ebd2b51644332b9d26dcf9d2e39b1a2125ae4bc9f6400a7cc412cR74

src/dune_rules/virtual_rules.ml
```diff
@@ -79,57 +55,19 @@ let impl sctx ~(lib : Library.t) ~scope =
...
        let info = Lib.info vlib in
-       if not (Lib_info.virtual_ info)
-       then
-         User_error.raise
-           ~loc:lib.buildable.loc
-           [ Pp.textf
-               "Library %s isn't virtual and cannot be implemented"
-               (Lib_name.to_string implements)
-           ];
-       let+ vlib_modules, vlib_foreign_objects =
```
art-w, line 90: This was also moved (unchanged) to `Vimpl.make`: https://github.com/ocaml/dune/pull/12392/files#diff-654c4360f33ebd2b51644332b9d26dcf9d2e39b1a2125ae4bc9f6400a7cc412cR20

art-w:
I think this is ready for another review :) . I had to move large chunks of code in the last commit to clarify the `implements` field handling of virtual modules and library parameter, which makes this PR looks bigger than it really is -- I would be happy to submit that last commit as a follow-up PR if it's easier to review.

rgrinberg (approved):

PR #12393, opened 2025-09-04
Check for parameterised libraries from the compiler config

Sudha247:
Changes the check for parameterised libraries to the flag emitted by the OxCaml compiler, rather than relying on parsing the compiler version string. This way, we rely on the implementation and not the compiler version, as this feature could get upstreamed to ocaml/ocaml. The OxCaml compiler config (`ocamlc -config`) emits `parameterised_modules: true`, which is being checked to detect whether parameterised libraries are available.

### DIFF
--- a/src/ocaml/version.ml
+++ b/src/ocaml/version.ml
@@ -39,9 +39,3 @@ let supports_oxcaml version =
   Stdune.String.is_suffix ~suffix:jst version
   || Stdune.String.is_suffix ~suffix:ox version
 ;;
-
-let supports_parametrized_library version =
-  (* We create the alias to make sure it is easy to distinguish the
-     functionality from the compiler variant. *)
-  supports_oxcaml version
-;;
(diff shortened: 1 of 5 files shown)


Alizter:
What about 
- https://github.com/ocaml/dune/pull/12236

Sudha247:
I think this isn't directly related to #12236, this PR checks for a specific feature while the latter checks for a compiler branch.

art-w (approved):
Looks good! We are hoping that the parameterized feature gets upstream to OCaml one day, so this check is more future proof instead of relying on detecting the OxCaml compiler :)

Alizter:
When OxCaml features are upstreamed, are the config names ever changed? That would be my only worry with this.

src/ocaml/version.ml
```diff
@@ -39,9 +39,3 @@ let supports_oxcaml version =
   Stdune.String.is_suffix ~suffix:jst version
   || Stdune.String.is_suffix ~suffix:ox version
 ;;
-
-let supports_parametrized_library version =
```
rgrinberg, line 43: Why did you change the spelling of parametrized?

rgrinberg: I guess this is the spelling they have upstream in oxcaml? In that case, this is the spelling we should use.

Sudha247: Indeed, `parameterised_modules` is the spelling in OxCaml.

shonfeder (approved):
LGTM! Seems like a nice clean improvement.

IIUC, there are no objections to this PR and it can be merged as soon as the merge conflicts are resolved.

PR #12394, opened 2025-09-04
Copy the lock directories to the build folder and read package rules from there

Leonidas-from-XIV:
This is a WIP PR to implement rules that copy the lock dirs (project lock dirs and dev-tool lock dirs) into the build folder before the package rules. It's a a prerequisite for #11775 which changes the logic to directly generate the lock directories in the build directory.

### DIFF
--- a/bin/lock_dev_tool.ml
+++ b/bin/lock_dev_tool.ml
@@ -79,7 +79,7 @@ let solve ~dev_tool ~local_packages =
       Workspace.add_repo workspace Dune_pkg.Pkg_workspace.Repository.binary_packages
     | `Disabled -> workspace
   in
-  let lock_dir = Lock_dir.dev_tool_lock_dir_path dev_tool in
+  let lock_dir = Dune_rules.Lock_dir.dev_tool_lock_dir dev_tool in
   Memo.of_reproducible_fiber
   @@ Lock.solve
        workspace
@@ -173,7 +173,7 @@ let extra_dependencies dev_tool =
 
 let lockdir_status dev_tool =
   let open Memo.O in
-  let dev_tool_lock_dir = Lock_dir.dev_tool_lock_dir_path dev_tool in
+  let dev_tool_lock_dir = Dune_rules.Lock_dir.dev_tool_lock_dir dev_tool in
   match Lock_dir.read_disk dev_tool_lock_dir with
   | Error _ -> Memo.return `No_lockdir
   | Ok { packages; _ } ->
--- a/bin/lock_dev_tool.ml
+++ b/bin/lock_dev_tool.ml
@@ -79,7 +79,8 @@ let solve ~dev_tool ~local_packages =
       Workspace.add_repo workspace Dune_pkg.Pkg_workspace.Repository.binary_packages
     | `Disabled -> workspace
   in
-  let lock_dir = Dune_rules.Lock_dir.dev_tool_lock_dir dev_tool in
+  (* as we want to write to the source, we're using the source lock dir here *)
+  let lock_dir = Dune_rules.Lock_dir.dev_tool_source_lock_dir dev_tool |> Path.source in
   Memo.of_reproducible_fiber
   @@ Lock.solve
        workspace
@@ -113,6 +114,7 @@ let locked_ocaml_compiler_version () =
     (* Dev tools are only ever built with the default context. *)
     Context_name.default
   in
+  (* TODO make sure this doesn't run when there is no lock dir *)
   let* result = Dune_rules.Lock_dir.get context
   and* platform =
     Pkg_common.poll_solver_env_from_current_system () |> Memo.of_reproducible_fiber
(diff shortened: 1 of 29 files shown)


src/dune_rules/source_deps.ml
```diff
@@ -7,7 +7,11 @@ module Map_reduce =
     (Monoid.Product (Monoid.Union (Path.Set)) (Monoid.Union (Path.Set)))
 
 let files_with_filter dir ~filter =
-  let prefix_with, dir = Path.extract_build_context_dir_exn dir in
+  let prefix_with, dir =
+    match (dir : Path.t) with
+    | In_source_tree dir -> Path.of_string "", dir
```
rgrinberg: `Path.root`?

bin/tools/tools_common.ml
```diff
@@ -20,7 +20,17 @@ let build_dev_tool_directly common dev_tool =
...
+      let d = dev_tool_exe_path dev_tool in
+      Printf.eprintf "Dev_tool_exe %S\n" (Path.to_string d);
+      let lock_dir_path = Dune_rules.Lock_dir.dev_tool_lock_dir dev_tool in
+      Printf.eprintf "Lock dir path for dev tool is %S\n" (Path.to_string lock_dir_path);
+      let open Action_builder.O in
+      dev_tool
+      |> Lock_dev_tool.lock_dev_tool
+      |> Action_builder.of_memo
+      (* probably the dev_tool_exe_path should have a dependency on lock_dir_path *)
+      >>> Action_builder.path lock_dir_path
```
rgrinberg: I don't think this dependency is valid. You're setting up a target for every package inside the lock directory. So you should be depending on the individual package here.

bin/lock_dev_tool.ml
```diff
@@ -173,36 +175,45 @@ let extra_dependencies dev_tool =
...
-                       The dev-tool %S will be re-locked and rebuilt with this version \
-                       of the compiler."
-                      (Package_name.to_string compiler_package_name)
-                      (Package_version.to_string ocaml_compiler_version)
-                      (Package_version.to_string info.version)
-                      (Dune_pkg.Dev_tool.package_name dev_tool |> Package_name.to_string)
-                  ]))))
+  let dev_tool_lock_dir = Dune_rules.Lock_dir.dev_tool_source_lock_dir dev_tool in
+  let* lock_dir_exists =
+    Dune_engine.Fs_memo.dir_exists (In_source_dir dev_tool_lock_dir)
```
rgrinberg, line 179: You should be using Source_tree (or at least Fs_memo) to inspect the source tree.

bin/pkg/pkg_common.mli
```diff
@@ -81,7 +81,7 @@ module Lock_dirs_arg : sig
 
       A user error is raised if the list of positional arguments used when
       creating [t] is not a subset of the lock directories of the workspace. *)
-  val lock_dirs_of_workspace : t -> Workspace.t -> Path.t list
+  val lock_dirs_of_workspace : t -> Workspace.t -> Path.Source.t list
```
rgrinberg, line 84: Seems like this change could be extracted into a separate PR?

src/dune_pkg/lock_dir.ml
```diff
@@ -1411,7 +1436,6 @@ module Make_load (Io : sig
     val parallel_map : 'a list -> f:('a -> 'b t) -> 'b list t
     val readdir_with_kinds : Path.t -> (Filename.t * Unix.file_kind) list t
     val with_lexbuf_from_file : Path.t -> f:(Lexing.lexbuf -> 'a) -> 'a t
-    val stats_kind : Path.t -> (File_kind.t, Unix_error.Detailed.t) result t
```
rgrinberg, line 1454: How come this function went away?

src/dune_rules/fetch_rules.ml
```diff
@@ -188,15 +188,18 @@ let find_checksum, find_url =
           Dune_pkg.Dev_tool.all
           ~init:(Checksum.Map.empty, Digest.Map.empty)
           ~f:(fun acc dev_tool ->
-            Fs_memo.dir_exists
-              (Path.as_outside_build_dir_exn
-                 (Dune_pkg.Lock_dir.dev_tool_lock_dir_path dev_tool))
+            Source_tree.find_dir (Lock_dir.dev_tool_source_lock_dir dev_tool)
```
rgrinberg, line 192: Why is this change needed?

src/dune_rules/lock_dir.ml
```diff
@@ -101,21 +101,16 @@ module Load = Make_load (struct
     include Memo
 
     let readdir_with_kinds path =
-      Fs_memo.dir_contents (Path.as_outside_build_dir_exn path)
-      >>| function
+      Readdir.read_directory_with_kinds (Path.to_string path)
```
rgrinberg, line 104: I don't see how this can work if the lock directory is still in the source directory. Without some sort of dependency tracking mechanism, dune will not know to invalidate the rules in watch mode if the lock directory changes.

bin/lock_dev_tool.ml
Leonidas-from-XIV: But this is using `Fs_memo`?

src/dune_pkg/lock_dir.ml
Leonidas-from-XIV: It was only used in `check_path` which was used to get information about a lock dir path. But given lock dirs are read from the build directory and created by copy rules (so either they exist because the rule exists or they don't which makes the build fail), there is nothing useful that `check_path` could communicate to the user anymore.

bin/tools/tools_common.ml
Leonidas-from-XIV: I've removed the dependency.

bin/pkg/pkg_common.mli
Leonidas-from-XIV: This is undoing some of the changes from #12194 as I realized that `path` in a workspace should *probably* stay a source path. I am not 100% sold on the change, hence I don't want to flip-flow in separate PRs making them `Path.Source.t` or `Path.t`.

src/dune_rules/lock_dir.ml
Leonidas-from-XIV: It doesn't read the source directory (hence why the PR this is spun off used `Path.Build.t` as path for a while). If the lock dir changes, the copy rule is retriggered, thus updating the files in the build folder.

src/dune_rules/fetch_rules.ml
Leonidas-from-XIV: If the dev-tool lock dir stops being in the source tree but ends up being read from the build folder, then `as_outside_build_dir_exn` will always fail with an exception. Hence I changed it, but then I realized that for now we need to check the existance of the dev tool lock dir in the source tree.

In any case, `Source_tree.find_dir` seems like the more appropriate function anyway, as it takes ignore-rules into account, whereas `Fs_memo.dir_exists` will happily read directories that were declared as to be ignored by the user.

bin/pkg/pkg_common.mli
Leonidas-from-XIV: Extracted the changes into #12505.

gridbugs:
Looks good to me. Happy for the disabled tests to be fixed in a later change. Nice work!

PR #12398, opened 2025-09-05
[Docs] Add 'edit on github' button to all pages

### LINKED ISSUE #12387: Enable "edit source" link on the docs
Currently it is not obvious to users or potential contributors how to help with improving our documentation. To make this as easy as possible, and lower the barrier to getting help from users on doc improvements, we should enable a link that takes users to the GitHub interface for updated the documentation they were reading.

The process for doing this looks straight forward, and is documented in https://docs.readthedocs.com/platform/latest/guides/edit-source-links-sphinx.html#github

The need for this was voiced by @maiste ! 🙏

ElectreAAS:
Closes #12387
Code is from [the furo documentation](https://github.com/pradyunsg/furo/blob/426ea05d879ee7dfdf74ba7c8b089f73081abbfd/docs/conf.py#L89-L105)

Preview over [here](https://dune--12398.org.readthedocs.build/en/12398/)

One thing of note is that the 'edit' link at the top of the page will always link to the main branch instead of the branch that the doc is running on. We decided to leave it at that for now.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -125,7 +125,13 @@ html_theme = 'furo'
 # further.  For a list of options available for each theme, see the
 # documentation.
 #
-html_theme_options = {}
+html_theme_options = {
+    "top_of_page_buttons": ["edit", "view"],
+
+    "source_repository": "https://github.com/ocaml/dune/",
+    "source_branch": "main",
+    "source_directory": "doc/",
+}
 
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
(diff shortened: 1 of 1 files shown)


Leonidas-from-XIV (approved):
The `html_context` is only useful if the theme reads it and Furo implements the page buttons differently, thus needs different configuration. LGTM.

shonfeder:
Tested the staging preview and it worked like a charm! Always linking to the main branch sounds like the right approach, IMO.

Great work!

shonfeder (approved):

Leonidas-from-XIV:
I am not sure if linking to `main` is the right approach in general but as discussed with @ElectreAAS it was like this before already (thus no worse than the status quo) and if we want the staging preview to link to the right place it can totally happen in a future PR.

IMHO it's entirely sufficient like it is now.

PR #12404, opened 2025-09-06
chore: remove dune_uutf

Alizter:


boot/libs.ml
```diff
@@ -1,5 +1,5 @@
 open Types
-let external_libraries = [ "unix"; "threads"; "spawn" ]
+let external_libraries = [ "unix"; "threads"; "uutf" ]
```
rgrinberg: How come spawn is modified here?

Alizter: Old bootstrap, fixing.

rgrinberg:
Windows CI isn't installing uutf

boot/libs.ml
```diff
@@ -1,5 +1,5 @@
 open Types
-let external_libraries = [ "unix"; "threads"; "spawn" ]
+let external_libraries = [ "unix"; "threads" ]
```
Alizter, line 2: I'm unsure why this keeps getting regenerated like this...

rgrinberg (approved):
LGTM. We should check if we need to update this library.

PR #12426, opened 2025-09-07
chore(nix): cleanup flake

Alizter:


flake.nix
```diff
@@ -320,19 +320,19 @@
...
               ]) ++ (with pkgs.ocamlPackages; [
-                pkgs.ocamlPackages.ocaml-lsp
-                pkgs.ocamlPackages.melange
-                js_of_ocaml-compiler
+                core_bench
                 js_of_ocaml
-                pkgs.binaryen
-                wasm_of_ocaml-compiler
+                js_of_ocaml-compiler
+                melange
```
anmonteiro: last time I checked, this wouldn't work because it would pick up the `input` named `melange`.

anmonteiro: it also seems like CI agrees.

Alizter: Yes, thanks for pointing that out. I'm kind of surprised that `with` scoping has this behaviour, but after thinking about it a bit, it makes sense.

PR #12438, opened 2025-09-08
Add DUNE_ROOT environment variable

### LINKED ISSUE #12399: Add support for `DUNE_ROOT` environment variable
## Desired Behavior
Please add support for a generic `DUNE_ROOT` env var, similar to passing `--root` in the command line.

## Rationale
Sometimes Dune guesses wrong the root.
E.g.:
```
| a/
    |- dune-workspace
    |- b/
        |- dune-workspace
```

If I am in directory `b/` and run `dune build`, it uses as root directory `a/`, and I have to instead use `dune build --root .` every time to work around that.

There are various reasons why one may not want to use the parent directory as root, like:
* if `a/dune` (or another dune file in `a/`) is invalid/broken, running `dune build` in `b/` will fail, while running `dune build --root=.` will succeed
* evaluating all dune files in `a/` might be slow, thus unnecessarily slowing building `b/`
* it can mess paths, e.g. we have a Makefile in `b/` which refers to `_build/<whatever>`. Running `make foo` in `b/` will use different paths thus fail or have a wrong behavior

About supporting the env var in particular, it is because it is MUCH more convenient than passing a CLI parameter every time. Env vars can be set once and all subsequent commands will use it, and they interact nicely with many tools (e.g. nix-shell). Comparatively, it is rarely convenient or even possible to pass extra CLI parameters via other tools.

Note: my examples are about `dune build` but any `dune` command supporting `--root` should also support `DUNE_ROOT` env var.

sir4ur0n:
Closes https://github.com/ocaml/dune/issues/12399

### DIFF
--- a/test/blackbox-tests/test-cases/cmdline/root.t
+++ b/test/blackbox-tests/test-cases/cmdline/root.t
@@ -1,6 +1,6 @@
 The interaction and order of overriding [DUNE_ROOT] and [--root].
 
-Feature request github.com/ocaml/dune/issues/12399
+Feature request https://github.com/ocaml/dune/issues/12399
 
 Create a [dune-project] and [dune] at the root and in 2 subdirectories:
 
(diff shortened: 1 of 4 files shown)


Leonidas-from-XIV:
Thanks for the contribution! Can you sign off your commit and add a [cram test](https://dune.readthedocs.io/en/stable/reference/cram.html) to show that `DUNE_ROOT` works and how its precedence with `--root` works out?

sir4ur0n:
@Leonidas-from-XIV done, please review :pray:

test/blackbox-tests/test-cases/cmdline/root.t
```diff
@@ -0,0 +1,95 @@
+The interaction and order of overriding [DUNE_ROOT] and [--root].
+
+Feature request github.com/ocaml/dune/issues/12399
```
Leonidas-from-XIV: Please add the `https://` prefix so it is clickable.

test/blackbox-tests/test-cases/cmdline/root.t
```diff
@@ -0,0 +1,95 @@
...
+  $ cat >subdir2/dune-project <<EOF
+  > (lang dune 3.21)
+  > EOF
+  $ cat >subdir2/dune <<EOF
+  > (rule
+  >   (alias runtest)
+  >   (action (echo "subdir2\n")))
+  > EOF
+
+  $ alias cleanup='rm -rf `find . -type d -name _build`'
```
Leonidas-from-XIV: This should better be `dune clean` or explicit `rm` in the test. `rm -rf $(find ...)` feels like building a footgun ready to go catastrophically wrong.

It's also probably not necessary at all if you call `dune runtest --force` or make the actions fail after printing (so their success does not get cached).

Leonidas-from-XIV (commented):
Thanks. The test is quite nice, covers a good chunk of functionality. I've added a few comments how to make it nicer.

Please also add a changelog entry in `doc/changes/12438.md`. It would probably also be good to mention the new variable in the documentation, e.g. somewhere here where `--root` is documented: https://dune.readthedocs.io/en/stable/usage.html#finding-the-root

sir4ur0n:
All suggestions have been addressed in abe5286856ab77d49248a3d666bb09fd765df015, please review again :pray:

Leonidas-from-XIV (approved):
LGTM, thanks for your contribution!

Alizter (approved):

PR #12440, opened 2025-09-09
chore: update CONTRIBUTING.md

shonfeder:
- Codify a lightweight development and design process, based on conversations with many active core developers in various meetings.
- Inviting feedback on issues as a way to contribute. This may seem obvious, but it bears making this explicit.
- Break subjects into clear subsections to improve Scannability.
- Point new contributors to our "good first issues".
- Link both the rendered and the source for the hacking doc.
- Tweak some phrasing

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,24 +1,69 @@
-Dune is an community orientated open source project. It was originally
-developed at [Jane Street][js] and is now maintained by Jane Street,
-[Tarides][tarides] as well as several developers from the OCaml
-community.
+Dune is an community orientated open source project. It was originally developed
+at [Jane Street][js] and is now maintained by Jane Street and [Tarides][tarides]
+together with several developers from the OCaml community.
 
-Contributions to Dune are welcome and should be submitted via GitHub
-pull requests against the `main` branch. See [./doc/hacking.rst][hack]
-for a guide to getting started on the code base.
+Filing issues
+=============
 
-Dune is distributed under the MIT license and contributors are
-required to sign their work in order to certify that they have the
-right to submit it under this license. See the following section for
-more details.
+The easiest way to contribute is to share your input on [issues][issues].
+Feedback on bug reports, feature requests, and documentation is welcome and
+appreciated. If you don't find a preexisting issue discussing your topic, then
+please [file a new issue][file an issue].
+
+[file an issue]: https://github.com/ocaml/dune/issues/new/choose
+[issues]: https://github.com/ocaml/dune/issues
+
+Developing Dune
+===============
+
+Contributions to the Dune code base are welcome!
+
+Our development process is as follows:
+
+- Non-trivial submissions should be proceeded by an issue communicating the
+  rationale for the intended change.
+- Substantial changes should be preceded by upfront design and planning,
+  proportional to the scope and impact of the intended change.
+  - This design should be reviewed by at least one other party.
+  - Github issues are an appropriate venue for most design discussions, but more
+    involved design work may warrant an RFC or ADR.
+- Design decisions agreed upon in dev meeting or online chats should also be
+  documented and reviewed in issues or design docs. This encourages us to think
+  through the problems thoroughly and ensures we leave a lasting record of the
+  decision and their rationale.
+- Changes are submitted via GitHub pull requests against the `main` branch.
+
+If you are looking to get started, check our [issues tagged with good first
+issue][good first issue].
+
+[good first issue]: https://github.com/ocaml/dune/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22
+
+Developer documentation
+-----------------------
+
+See our developer documentation at [./doc/hacking.rst][hack] or
+[online][devdocs] for technical guidance about contributing to the code base.
+
+[devdocs]: https://dune.readthedocs.io/en/stable/hacking.html
+
+Developer meetings
+------------------
+
+See our [developer wiki][developer wiki] for the latest information about our
+recurring developer meetings. All interested attendees are welcome.
+
+[developer wiki]: https://github.com/ocaml/dune/wiki/
 
 Signing contributions
 ---------------------
 
-We require that you sign your contributions. Your signature certifies
-that you wrote the patch or otherwise have the right to pass it on as
-an open-source patch. The rules are pretty simple: if you can certify
-the below (from [developercertificate.org][dco]):
+Dune is distributed under the MIT license and contributors are required to sign
+their work in order to certify that they have the right to submit it under this
+license.
+
+Your signature certifies that you wrote the patch or otherwise have the right to
+pass it on as an open-source patch. The rules are pretty simple: if you can
+certify the below (from [developercertificate.org][dco]):
 
 ```
 Developer Certificate of Origin
@@ -81,9 +126,3 @@ you have the rights to submit this work under the same open source license.
 [js]: https://www.janestreet.com/
 [tarides]: https://tarides.com/
 [hack]: ./doc/hacking.rst
-
-Coding style
-------------
-
-- wrap lines at 80 characters,
-- use `[Ss]nake_case` over `[Pp]ascalCase`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,10 +37,11 @@ Our development process is as follows:
   recording the what and why of our work in the persistent record of GitHub.
 - Changes are submitted via GitHub pull requests against the `main` branch.
 
-If you are looking to get started, check our [issues tagged with good first
-issue][good first issue].
+If you are looking to get started, check our issues tagged with [good first
+issue][good first issue] or [help wanted][help wanted].
 
 [good first issue]: https://github.com/ocaml/dune/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22
+[help wanted]: https://github.com/ocaml/dune/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22help%20wanted%22
 
 Developer documentation
 -----------------------
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,16 +2,21 @@ Dune is an community orientated open source project. It was originally developed
 at [Jane Street][js] and is now maintained by Jane Street and [Tarides][tarides]
 together with several developers from the OCaml community.
 
-Filing issues
-=============
+Sharing feedback
+================
 
-The easiest way to contribute is to share your input on [issues][issues].
-Feedback on bug reports, feature requests, and documentation is welcome and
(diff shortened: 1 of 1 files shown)


shonfeder:
I tagged in quite a few folks for review, just to be sure that interested parties can review if they care to. But please don't feel obligated to review/approve if you don't. The important thing is just that blocking concerns be addressed, if there turns out to be any.

CONTRIBUTING.md
```diff
@@ -1,24 +1,69 @@
...
-Dune is distributed under the MIT license and contributors are
-required to sign their work in order to certify that they have the
-right to submit it under this license. See the following section for
-more details.
+The easiest way to contribute is to share your input on [issues][issues].
+Feedback on bug reports, feature requests, and documentation is welcome and
+appreciated. If you don't find a preexisting issue discussing your topic, then
+please [file a new issue][file an issue].
+
+[file an issue]: https://github.com/ocaml/dune/issues/new/choose
```
Alizter, line 17: We should also mention discussions.

shonfeder: Sounds good. What would we like to ask people to use discuss for?

shonfeder: To be clear: I would like people to know when an issue is appropriate and when discussions are, but I never use github discussions myself and find them to be a redundant feature. This is not an objection, I just don't know what to say about them :sweat_smile:

Alizter: I guess discussions are really for things that aren't issues or feature requests. Somebody might have trouble building dune for instance, and that might fall into a place for discussion. Another example might be somebody asking how certain parts of the code base works. We typically have such discussions on Discord and discuss.ocaml.org, but having a dedicated place doesn't hurt. Whether or not we are checking the discussions as thoroughly as we should is another matter.

Tldr: If it isn't a bug or feature request, just something you would like to discuss, then discussions is the place for it.

CONTRIBUTING.md
```diff
@@ -1,24 +1,69 @@
...
+  - Github issues are an appropriate venue for most design discussions, but more
+    involved design work may warrant an RFC or ADR.
+- Design decisions agreed upon in dev meeting or online chats should also be
+  documented and reviewed in issues or design docs. This encourages us to think
+  through the problems thoroughly and ensures we leave a lasting record of the
+  decision and their rationale.
+- Changes are submitted via GitHub pull requests against the `main` branch.
+
+If you are looking to get started, check our [issues tagged with good first
+issue][good first issue].
```
maiste: I would also point the existent to `Open-to-contribution` tag. It is for things people don't have time to work on, but external contributors can decide to take.

CONTRIBUTING.md
```diff
@@ -1,24 +1,69 @@
...
+
+[good first issue]: https://github.com/ocaml/dune/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22
+
+Developer documentation
+-----------------------
+
+See our developer documentation at [./doc/hacking.rst][hack] or
```
maiste, line 49: A pointer could also be added to point to the "Tour of the Dune Codebase". When I started working on Dune, it was also a good place to learn more about the internal behavior.

CONTRIBUTING.md
shonfeder: Oh that is kind of a surprising tag. Does this imply that we are not open to external contributors generally? That seems not good to me. But if it is the case, then we should revise the contribution guide here, letting people know that contributions are *not* welcome in general, but only if you become "internal" (how is that defined?) or if the issue is labeled with `open-to-contribution`.

WDYT?

CONTRIBUTING.md
shonfeder: Oh that is handy! Thanks for pointing this out. However, I will suggest that we instead link this at the top of `hacking.rst`, so we have one entry point for developer documentation.  Are you okay with this alternative?

Alizter:
By the way, when you tag somebody in a commit, it will notify them each time you push. So please edit the description to not include the `@`.

CONTRIBUTING.md
shonfeder: If we don't want to have the position that external contributors are not welcome, maybe we should just apply `help wanted` to all these issues and remove the `open-to-contribution` tag?

rgrinberg: It implies that the team does not consider this issue to be important enough to work on, but will accept the work if somebody else were to do it.

The tag used to be called "help-wanted". That might have made the intent clearer

Leonidas-from-XIV: I agree, however I am not sure if a `help wanted` (`open-to-contribution`) has ever worked in Dune? I think it is great to invite contributions but then it is on the maintainers to guide contributors and not let the work bit-rot because nobody has the time to work on guiding the progress of the PR.

maiste: Indeed, the wording `Open-to-contribution` is opposed to the idea of being more welcoming with external contributors. It can be changed back to `help-wanted` but the wording is also not really straight forward IMO... 

I agree with @Leonidas-from-XIV point.

Leonidas-from-XIV: As we all agree that `help wanted` is the better wording, I've moved all the `open-to-contribution` labels to `help wanted` and updated the description of the tag with better wording.

rgrinberg: I suppose "help-wanted" doesn't say anything about the motivation about why the maintainers don't want to work on this. It's usually one of two reasons:

1. The implementation is challenging
2. The feature's too niche to be useful to a broad group of users.

Bit-rotting usually occurs in the 1st case.

CONTRIBUTING.md
```diff
@@ -1,24 +1,74 @@
...
+Our development process is as follows:
+
+- Non-trivial submissions should be proceeded by an issue communicating the
+  rationale for the intended change.
+- Substantial changes should be preceded by upfront design and planning,
+  proportional to the scope and impact of the intended change.
+  - This design should be reviewed by at least one other party.
+  - Github issues are an appropriate venue for most design discussions, but more
+    involved design work may warrant an RFC or ADR.
+- Design decisions agreed upon in dev meeting or online chats should also be
```
rgrinberg: This needs to be rephrased a bit. I think we should also state that github is the preferred venue for discussing design. Meetings and chat are good, but they're more effective for addressing the _how_ of getting things done rather than _what_ or _why_.

CONTRIBUTING.md
shonfeder: I added mention of `help wanted`.

shonfeder:
Thanks for the reviews! I'd appreciate another review here so I can land this little documentation change. It sounds like there are no objections to the core points proposed.

CONTRIBUTING.md
```diff
@@ -1,24 +1,74 @@
-Dune is an community orientated open source project. It was originally
-developed at [Jane Street][js] and is now maintained by Jane Street,
-[Tarides][tarides] as well as several developers from the OCaml
-community.
+Dune is an community orientated open source project. It was originally developed
```
gridbugs, line 1:
```suggestion
Dune is a community orientated open source project. It was originally developed
```

gridbugs (approved):

maiste (approved):

CONTRIBUTING.md
```diff
@@ -1,24 +1,74 @@
...
+[discussions]: https://github.com/ocaml/dune/discussions
+
+Developing Dune
+===============
+
+Contributions to the Dune code base are welcome!
+
+Our development process is as follows:
+
+- Non-trivial submissions should be proceeded by an issue communicating the
```
Sudha247:
```suggestion
- Non-trivial submissions should be preceded by an issue communicating the
```

CONTRIBUTING.md
```diff
@@ -81,9 +131,3 @@ you have the rights to submit this work under the same open source license.
 [js]: https://www.janestreet.com/
 [tarides]: https://tarides.com/
 [hack]: ./doc/hacking.rst
-
-Coding style
```
Sudha247, line 85: Just to be clear, by removing this, do we no longer subscribe to it, or is it an implicit style that exists in the code base? Maybe we can just recommend formatting the code before pushing it?

Sudha247 (commented):
Thanks for expanding the contributing section!

CONTRIBUTING.md
Leonidas-from-XIV: The hacking document has a way more elaborate version of this, so I assume this is just done to avoid splitting the information about coding style into multiple documents.

CONTRIBUTING.md
shonfeder: Whoops! Thanks for the fix :)

CONTRIBUTING.md
shonfeder: Yep that's right! I'm removing it cause the hacking guide seems to be source of truth here, and this seems like an out of date little starter on the topic. :)

rgrinberg (approved):

CONTRIBUTING.md
```diff
@@ -1,24 +1,74 @@
...
+
+Our development process is as follows:
+
+- Non-trivial submissions should be preceded by an issue communicating the
+  rationale for the intended change.
+- Substantial changes should be preceded by upfront design and planning,
+  proportional to the scope and impact of the intended change.
+  - This design should be reviewed by at least one other party.
+  - Github issues are an appropriate venue for most design discussions, but more
+    involved design work may warrant an RFC or ADR.
```
Alizter, line 34: What is an ADR? Google says "Alternative Dispute Resolution" but I don't see how this applies here.

CONTRIBUTING.md
```diff
@@ -1,24 +1,74 @@
...
 Signing contributions
 ---------------------
 
-We require that you sign your contributions. Your signature certifies
-that you wrote the patch or otherwise have the right to pass it on as
-an open-source patch. The rules are pretty simple: if you can certify
-the below (from [developercertificate.org][dco]):
+Dune is distributed under the MIT license and contributors are required to sign
+their work in order to certify that they have the right to submit it under this
+license.
```
Alizter, line 67: maybe something like
```
Signed-off-by: Example Person <person@example.com>
```
at the end of your commit should be mentioned. There are two types of commit signing in Git which are sometimes confused. The other is GPG signing which we don't care about and can be a bit involved to set up, so we should be clearer to potential contributors about this.

Alizter: I see that we do mention this after the certificate. Perhaps we should pull that example above.

shonfeder: I haven't changed the copy here at all, but I'm fine with moving this above as part this PR.

CONTRIBUTING.md
shonfeder: I guess you may need to specify the context of software tog et the right result :) https://adr.github.io/

Alizter (approved):
My comments aren't blocking. The rest looks good to me.

shonfeder:
Let's make additional improvements in followup. Thanks for the reviews, all!

PR #12443, opened 2025-09-10
Execute commands in `opam-var-os.t`

Leonidas-from-XIV:
The cram test accidentally used `>` instead of `$` in prompts to execute commands. Somehow it still worked, but I am unsure why it worked.


shonfeder (approved):

Alizter:
It worked because this is still valid cram syntax.

Whenever you have something like:

```
  $ A
  > B
  > C
```

you can think of it as a shell script like:

```
A
B
C
```

It might be confusing because we typically always start a new line except for certain multi-line commands like:

```
  $ cat > foo <<EOF
  > bar
  > baz
  > EOF
```

but this is just

```
cat > foo <<EOF
bar
baz
EOF
```

Leonidas-from-XIV:
I am aware it is valid cram syntax, but usually fixing a test ends up with the things being fixed to fail. Here it outputs exactly (!) the desired output which is somewhat unexpected because it could not have been created by promotion as the command did not run before. Which makes me suspect that the most likely explanation for it was that it used to say `$` but by some small editing mistake it got turned into `>` later.

Alizter:
I don't follow the part where you said the command was not being run. In
```
  $ A
  > B
```
both `A` and `B` commands are being run? In our case `B` is the one producing the output you see under the command block. Am I misunderstanding something?

Issue #12097, opened 2025-07-29
dune-pkg: do not write lock directories into the worktree by default


shonfeder:
We have been considering the following:

1. Everyone on our team at Tarides, and afaiu, all dune devs and most users, agree that having a portable way of locking dependencies is something we want eventually.
2. However, there is a lot of disagreement, and a general lack of clarity, about particulars of the implementation and design (see, e.g., [this](https://discuss.ocaml.org/t/portable-external-dependencies-for-dune-package-management/16767/6), [that](https://github.com/ocaml/dune/issues/11868), or [the other](https://github.com/ocaml/dune/issues/10955)).
4. IMO, it is kind of obvious that it should be possible to use `dune pkg` without committing a lock dir in the worktree, and in fact there are many cases where users would *not* want to do this. E.g.,
   - See https://github.com/rust-lang/cargo/issues/315, where Rustaceans discuss why it is common for libraries to *not* check in the lock directory.
   - Notice that widely used Rust libraries like [thiserror](https://github.com/dtolnay/thiserror), [serde](https://github.com/serde-rs/serde) don't check in their lock file.
   - Recall that the vast majority or ocaml projects currently do not check in a lock file. 
6. Additionally, AFAIU, the current behaviour around generation of lock directories is an aberration from the way dune usually wants to handle generated artifacts: which is to promote them from the `_build` directory to the worktree, and not write generated stuff directly into the worktree.
7. Since the most urgent aim with `dune pkg` is to provide a stable, and fully usable alternative for `opam` package management to ocaml developers, we don't need to solve portable locking *first* because this is not generally used in the ecosystem. (Pinning the opam repo already suffices for most mundane our locking needs).
8. There have been fruitful discussions recently about ways of achieving locked dependencies, without needing the noise of the full lock dir to ever be committed into the repo (cf. [Lock Files Considered Harmful][lock-post] for an interesting related discussion!).

As a result of these considerations, we have come to the following conclusions:

1. By default, lock directories should be maintained in the `_build` dir and only promoted into the working directory if a user actually wants to commit it.
2. Moving lock directories into the `_build` dir by default allows us to unblock on the portability consideration, and move forward with a focus on the most pressing requirements for achieving stability.
3. But we still want to find and implement the optimal solution to portable package locking, and we want to keep this aim in view for the mid term.
   - A possible alternative here has been proposed by @rgrinberg, and is also discussed in the [above mentioned post][lock-post].

---

[...]

PR #12444, opened 2025-09-10
fix(pkg): pkg disabled now disables package management

Alizter:
The option `(pkg disabled)` didn't actually disable any package management features other than changing the value of the `dune pkg enabled` command. This PR introduces that functionality and adds a further test making sure it behaves as intended. We also document the field which was previously undocumented.

Since it was undocumented and buggy, it might be worth warning users on 3.20 and bump the versions to 3.21.

- [ ] introduced in 3.20 but didn't really work, bump to 3.21 and add warning?
- [x] docs
- [x] changelog
- [x] tests

### DIFF
--- a/bin/pkg/pkg_common.ml
+++ b/bin/pkg/pkg_common.ml
@@ -226,3 +226,21 @@ module Lock_dirs_arg = struct
           ])
   ;;
 end
+
+let check_pkg_management_enabled () =
+  Memo.run
+  @@
+  let open Memo.O in
+  let+ workspace = Workspace.workspace () in
+  match workspace.config.pkg_enabled with
+  | Set (_, `Enabled) | Unset -> ()
+  | Set (loc, `Disabled) ->
+    User_error.raise
+      ~loc
+      [ Pp.text "Package management is disabled in workspace configuration." ]
+      ~hints:
+        [ Pp.text
+            "To enable package management, remove the explicit (pkg disabled) setting \
+             from your dune-workspace file."
+        ]
+;;
--- a/bin/pkg/pkg_enabled.ml
+++ b/bin/pkg/pkg_enabled.ml
@@ -16,7 +16,12 @@ let term =
     let any_lockdir_exists = List.exists lock_dir_paths ~f:Path.exists in
     (* CR-Leonidas-from-XIV: change this logic when we stop detecting lock
        directories in the source tree *)
-    let enabled = any_lockdir_exists || workspace.config.pkg_enabled in
+    let enabled =
+      match workspace.config.pkg_enabled with
+      | Set (_, `Enabled) -> true
+      | Set (_, `Disabled) -> false
+      | Unset -> any_lockdir_exists
+    in
     match enabled with
     | true -> ()
     | false -> exit 1)
--- a/doc/reference/config/pkg.rst
+++ b/doc/reference/config/pkg.rst
@@ -1,0 +1,26 @@
+pkg
+---
+
+Controls whether Dune's package management features are enabled. Package
+management includes dependency locking, package resolution, and building 
+dependencies from source.
+
+.. code:: dune
+
+    (pkg <setting>)
+
+where ``<setting>`` is one of:
+
+- ``enabled`` forces package management to be enabled, even if no lock
+  directories are present.
+
+- ``disabled`` forces package management to be disabled, even if lock
+  directories are present. When disabled, Dune will not load package rules
+  and all ``dune pkg`` commands will fail with an error.
+
+If no ``(pkg ...)`` setting is specified, Dune will auto-detect whether to
+enable package management based on the presence of lock directories in the
+workspace. This usage is not currently recommended, since lock directories 
+are not stabilized or portable.
+
+.. versionadded:: 3.20
--- a/src/dune_config_file/dune_config_file.ml
+++ b/src/dune_config_file/dune_config_file.ml
@@ -52,11 +52,29 @@ module Dune_config = struct
   end
 
   module Pkg_enabled = struct
-    type t = bool
+    type t =
+      | Set of Loc.t * Dune_config.Config.Toggle.t
+      | Unset
 
-    let decode = enum' [ "enabled", return true; "disabled", return false ]
-    let equal = Bool.equal
-    let to_dyn = Dyn.bool
+    let decode =
+      let open Dune_lang.Decoder in
+      let+ loc, value = located (enum [ "enabled", `Enabled; "disabled", `Disabled ]) in
+      Set (loc, value)
+    ;;
+
+    let equal x y =
+      match x, y with
+      | Set (x_loc, x_toggle), Set (y_loc, y_toggle) ->
+        Loc.equal x_loc y_loc && Dune_config.Config.Toggle.equal x_toggle y_toggle
+      | Set _, _ | _, Set _ -> false
+      | Unset, Unset -> true
+    ;;
+
+    let to_dyn = function
+      | Set (loc, toggle) ->
+        Dyn.variant "Set" [ Loc.to_dyn loc; Config.Toggle.to_dyn toggle ]
+      | Unset -> Dyn.variant "Unset" []
+    ;;
   end
 
   module Terminal_persistence = struct
@@ -480,7 +498,7 @@ module Dune_config = struct
         ; maintenance_intent = None
         ; license = Some [ "LICENSE" ]
         }
-    ; pkg_enabled = false
+    ; pkg_enabled = Unset
     ; experimental = []
     }
   ;;
--- a/src/dune_config_file/dune_config_file.mli
+++ b/src/dune_config_file/dune_config_file.mli
@@ -55,7 +55,18 @@ module Dune_config : sig
   end
 
   module Pkg_enabled : sig
-    type t = bool
+    (** Configuration for Dune's package management features.
+        
+        - [Set (loc, `Enabled)]: Package management is explicitly enabled. Forces package 
+          management to be active even if no lock directories are present.
+
+        - [Set (loc, `Disabled)]: Package management is explicitly disabled. Forces package 
+          management to be inactive even if lock directories are present.
+
+        - [Unset]: Package management enablement is not explicitly configured.  *)
+    type t =
+      | Set of Loc.t * Dune_config.Config.Toggle.t
+      | Unset
   end
 
   module Terminal_persistence : sig
--- a/src/dune_rules/lock_dir.ml
+++ b/src/dune_rules/lock_dir.ml
@@ -186,13 +186,18 @@ let of_dev_tool dev_tool =
 ;;
 
 let lock_dir_active ctx =
+  let open Memo.O in
   if !Clflags.ignore_lock_dir
   then Memo.return false
   else
-    get_path ctx
-    >>= function
-    | None -> Memo.return false
-    | Some path -> Fs_memo.dir_exists (Path.as_outside_build_dir_exn path)
+    let* workspace = Workspace.workspace () in
+    match workspace.config.pkg_enabled with
+    | Set (_, `Disabled) -> Memo.return false
+    | Set (_, `Enabled) | Unset ->
+      get_path ctx
+      >>= (function
+       | None -> Memo.return false
+       | Some path -> Fs_memo.dir_exists (Path.as_outside_build_dir_exn path))
 ;;
 
 let source_kind (source : Dune_pkg.Source.t) =
--- a/test/blackbox-tests/test-cases/pkg/helpers.sh
+++ b/test/blackbox-tests/test-cases/pkg/helpers.sh
@@ -62,7 +62,7 @@ add_mock_repo_if_needed() {
   if [ ! -e dune-workspace ]
   then
       cat >dune-workspace <<EOF
-(lang dune 3.10)
+(lang dune 3.20)
 (lock_dir
  (repositories mock))
 (repository
@@ -92,6 +92,19 @@ EOF
   
     fi
   fi
+}
+
+create_mock_repo() {
+  # Always create a fresh workspace with mock repository configuration
+  repo="${1:-file://$(pwd)/mock-opam-repository}"
+  cat >dune-workspace <<EOF
+(lang dune 3.20)
+(lock_dir
+ (repositories mock))
+(repository
+ (name mock)
+ (url "${repo}"))
+EOF
 }
 
 make_lockpkg() {
@@ -128,7 +141,7 @@ EOF
 
 make_project() {
   cat <<EOF
-(lang dune 3.11)
+(lang dune 3.20)
  (package
   (name x)
   (allow_empty)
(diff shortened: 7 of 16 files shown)


doc/reference/config/pkg.rst
```diff
@@ -0,0 +1,25 @@
...
+- ``enabled`` forces package management to be enabled, even if no lock
+  directories are present.
+
+- ``disabled`` forces package management to be disabled, even if lock
+  directories are present. When disabled, Dune will not load package rules
+  and all ``dune pkg`` commands will fail with an error.
+
+If no ``(pkg ...)`` setting is specified, Dune will auto-detect whether to
+enable package management based on the presence of lock directories in the
+workspace. This is the recommended default behavior for most projects.
```
shonfeder: My understanding is that our current plan is to *not* have this be the recommended default until after we have stabilized lock dirs.

```suggestion
workspace. This usage is not currently recommended, since lock directories are not stabilized or portable.
```

src/dune_config_file/dune_config_file.mli
```diff
@@ -55,7 +55,7 @@ module Dune_config : sig
   end
 
   module Pkg_enabled : sig
-    type t = bool
+    type t = (Loc.t * bool) option
```
shonfeder: It is not evident what `Some false` means, so 

1. a doc comment explaining the intended meanings here would be helpful. 

But

2. Noting the meaning in the code by making this a record like `{loc: Loc.t; enabled: bool}` may be a better idea.

But, actually 

3. this is kind of an obfuscated three way disjunction (it is `Some true`, `Some false`, or `None`), so perhaps it would be even better as


```suggestion
    type t = Enabled of Loc.t | Disabled of Loc.t | Not_set
```

IMO, any of these three options would be good enough, but I prefer (2) or (3) since it will save people having to consult to docs to understand the meaning of the values.

src/dune_rules/lock_dir.ml
```diff
@@ -186,13 +186,16 @@ let of_dev_tool dev_tool =
...
 let lock_dir_active ctx =
-  if !Clflags.ignore_lock_dir
-  then Memo.return false
-  else
+  let open Memo.O in
+  let* workspace = Workspace.workspace () in
+  match workspace.config.pkg_enabled with
+  | _ when !Clflags.ignore_lock_dir -> Memo.return false
+  | Some (_, false) -> Memo.return false
+  | Some (_, true) | None ->
```
shonfeder: What if `workspace.config.pkg_enabled = None && not (!Clflags.ignore_lock_dir)` ?

test/blackbox-tests/test-cases/pkg/pkg-disabled-workflow.t
```diff
@@ -0,0 +1,187 @@
...
+Test that pkg commands fail when disabled:
+
+  $ dune pkg lock
+  File "dune-workspace", line 7, characters 5-13:
+  7 | (pkg disabled)
+           ^^^^^^^^
+  Error: Package management is disabled in workspace configuration.
+  Hint: To enable package management, remove the explicit (pkg disabled)
+  setting from your dune-workspace file.
+  [1]
```
shonfeder, line 86: These tests are great! Really clear explanations and set up.

test/blackbox-tests/test-cases/pkg/pkg-disabled-workflow.t
```diff
@@ -0,0 +1,187 @@
...
+
+  $ dune pkg outdated 2>&1 | grep "Error:"
+  Error: Package management is disabled in workspace configuration.
+
+  $ dune pkg validate-lockdir 2>&1 | grep "Error:"
+  Error: Package management is disabled in workspace configuration.
+
+Test that dune build fails appropriately when trying to use unavailable
+packages:
```
shonfeder: IIUC, this would make it clearer how this test relates to the previous:

```suggestion
Test that dune build fails appropriately when trying to use unavailable
packages, proving we did not install the needed dependency:
```

shonfeder:
Good catch and fix!

bin/pkg/pkg_common.ml
```diff
@@ -226,3 +226,21 @@ module Lock_dirs_arg = struct
           ])
   ;;
 end
+
+let check_pkg_management_enabled () =
+  Memo.run
```
Leonidas-from-XIV, line 231: Is this a safe operation to just run a memo outside the build system? I probably would have returned a `bool Memo.t` here.

doc/changes/12444.md
```diff
@@ -0,0 +1 @@
+- Add `(pkg enabled/disabled)` workspace configuration to control package management. (#12444, @Alizter)
```
Leonidas-from-XIV: We don't add `dune pkg` related things to the changelog, usually.

doc/reference/config/pkg.rst
Leonidas-from-XIV: The whole reason for `(pkg enabled)` was to enable package management in the absence of lock directories as there would otherwise be no signal to use it.

I do think that projects should not need to opt in via `(pkg enabled)`. They can opt-out via `(pkg disabled)` but we should try our best to have most projects be able to use `dune pkg` with as little additional setup as possible.

src/dune_config_file/dune_config_file.mli
Leonidas-from-XIV: The AST proposed is fine, but it's not very consistent with how the rest of Dune works.

To me `Some (loc, false)` very clearly reads `(pkg disabled)`, that is the user did specify the `pkg` option (`Some`) and they picked `disabled` (`false`).

The `Loc.t * t` tuple is pretty much an idiom in the codebase, so while the record is a little bit nicer, it doesn't fit too well with the rest of the code.

src/dune_rules/lock_dir.ml
Leonidas-from-XIV: User did not opt out via `dune-workspace` nor did the user opt out using the CLI flag: Then it should be considered active.

test/blackbox-tests/test-cases/pkg/helpers.sh
```diff
@@ -94,6 +94,19 @@ EOF
...
+  repo="${1:-file://$(pwd)/mock-opam-repository}"
+  cat >dune-workspace <<EOF
+(lang dune 3.20)
+(lock_dir
+ (repositories mock))
+(repository
+ (name mock)
+ (url "${repo}"))
+EOF
+}
```
Leonidas-from-XIV, line 108: Isn't there a function that already does this? I think it would be better to just use `create_mock_repo_if_needed` and use `rm dune-workspace` if you really need to force it to be recreated.

Leonidas-from-XIV (commented):
I don't think it needs to be bumped to 3.21, because the sole reason it got merged was to enable #11775 to be opted-in and was never documented in the first place. The fact that the 3.20 release happened in between merging `(pkg …)` and #11775 is unfortunate but sometimes happens when you factor out changes out of PRs - they become non-atomic and with a time-based release schedule the features can end up landing in different releases.

So `(pkg …)` was intended more as a force opt-in, with the opt-out functionality tbd. I do appreciate this PR actually adding the opt-out logic and I think it makes sense to have it going forward.

However for projects that have checked in lock dirs `(pkg enabled)` is pretty much optional, since we will be continuing to support lock dirs in the source tree for the forseeable future.

test/blackbox-tests/test-cases/pkg/helpers.sh
Alizter: I did use the other one before, but got tired of deleting the lock directory every time. I think its fine to have both versions. We could perhaps improve the comments describing them.

doc/reference/config/pkg.rst
shonfeder: I don't understand what you're advocating for. But it seems consistent with what I'm asking for afaict.

src/dune_config_file/dune_config_file.mli
shonfeder: Sure. If this is following conventions in the code base then my suggested improvements aren't necessary. But I'd at minimum ask for the api to be documented to explain the meaning.

src/dune_rules/lock_dir.ml
shonfeder: Ah sorry. I missed the None case combined with the last Some and thought this match was incomplete.

doc/reference/config/pkg.rst
Alizter: It comes down to the original motivation for the `(pkg enabled)` configuration in the first place, and this was for the command `dune pkg enabled` to be able to detect if package management was in fact enabled. This wasn't possible before, because we guessed if it was enabled (or rather loaded package rules) based on the presence of a lock directory.

The 3 states that I am now tracking in the configuration reflect the 3 states we have. 
1. The user wishes for pkg to be enabled, but might not have a lock directory.
2. The user wishes for pkg to be disabled, but might have a lock directory in which case ignore it.
3. Determine whether or not pkg is enabled by the presence of a lock directory alone.

Almost all users using package management will fall into (3), the only reason to play around with the configuration is for niche settings. The entire motivation of this PR is that we already have this configuration value, I'm just making sure it isn't left in a half-working state.

src/dune_config_file/dune_config_file.mli
Alizter: I'll document the values based on the other explanation of the states I gave.

doc/reference/config/pkg.rst
shonfeder: > Almost all users using package management will fall into

Since we are pursing https://github.com/ocaml/dune/issues/12097, the desired interim state is that 

> By default, lock directories should be maintained in the _build dir...

The point is that

> Moving lock directories into the _build dir by default allows us to unblock on the portability consideration, and move forward with a focus on the most pressing requirements for achieving stability.

Which means we actually would not want people keeping the lock dir in their work tree until we have stabilized the portability, which is whole point that motivated this addition.

This is why I am asking the recommendation be removed for now, since it goes against the thing we are trying to achieve.

shonfeder: I am worried that we still don't seem aligned on the point here, as it seems to indicate we (at last I) don't agree on what we are trying to achieve with the related work, but I won't block on this point of the documentation if you two want to lave it as is.

Alizter: I'll remove mention of recommendation as it is not important for this test. There is no point documenting the package management workflow in a single test, since such a document should be living in the docs.

shonfeder: Sorry if I am very confused, but this file is the docs, no?

Alizter: Sorry, I got confused with the order of the comments. You are correct.

bin/pkg/pkg_common.ml
Alizter: I think it's fine for configuration values. We do it elsewhere too (not just pkg). If we were building things it would be less safe.

bin/pkg/pkg_common.ml
```diff
@@ -226,3 +226,21 @@ module Lock_dirs_arg = struct
...
+  let+ workspace = Workspace.workspace () in
+  match workspace.config.pkg_enabled with
+  | Enabled _ | Unset -> ()
+  | Disabled loc ->
+    User_error.raise
+      ~loc
+      [ Pp.text "Package management is disabled in workspace configuration." ]
+      ~hints:
+        [ Pp.text
+            "To enable package management, remove the explicit (pkg disabled) setting \
```
Leonidas-from-XIV, line 243: Nitpick: Given `loc` presumably points to the right place with the snippet being removed, I'd formulate it more like "To enable package management, remove [the `pkg` stanza/this] from the configuration".

bin/pkg/pkg_enabled.ml
```diff
@@ -16,7 +16,12 @@ let term =
     let any_lockdir_exists = List.exists lock_dir_paths ~f:Path.exists in
     (* CR-Leonidas-from-XIV: change this logic when we stop detecting lock
        directories in the source tree *)
-    let enabled = any_lockdir_exists || workspace.config.pkg_enabled in
+    let enabled =
+      match workspace.config.pkg_enabled with
```
Leonidas-from-XIV, line 20: It would be slightly nicer if we got a way to get this information directly without the `loc` information that has to be thrown away.

src/dune_rules/lock_dir.ml
```diff
@@ -186,13 +186,16 @@ let of_dev_tool dev_tool =
...
 let lock_dir_active ctx =
-  if !Clflags.ignore_lock_dir
-  then Memo.return false
-  else
+  let open Memo.O in
+  let* workspace = Workspace.workspace () in
+  match workspace.config.pkg_enabled with
+  | _ when !Clflags.ignore_lock_dir -> Memo.return false
+  | Disabled _ -> Memo.return false
+  | Enabled _ | Unset ->
```
Leonidas-from-XIV: As above, getting the value without the `loc` seems like a fairly common operation, so having a way to get it without the `loc` could be nice.

src/dune_rules/lock_dir.ml
```diff
@@ -186,13 +186,16 @@ let of_dev_tool dev_tool =
 ;;
 
 let lock_dir_active ctx =
-  if !Clflags.ignore_lock_dir
-  then Memo.return false
-  else
+  let open Memo.O in
+  let* workspace = Workspace.workspace () in
+  match workspace.config.pkg_enabled with
+  | _ when !Clflags.ignore_lock_dir -> Memo.return false
```
Leonidas-from-XIV: Personally I would do a nested match with an early result when `ignore_lock_dir` is `true` instead of a `_` branch (which disables the check whether all cases are handled) and I think it would read slightly easier, but it's fine either way.

Leonidas-from-XIV (approved):
Some nitpicks on readability but it's ok to merge either way and nicely implements the missing functionality.

src/dune_config_file/dune_config_file.ml
```diff
@@ -52,11 +52,33 @@ module Dune_config = struct
   end
 
   module Pkg_enabled = struct
-    type t = bool
+    type t =
+      | Enabled of Loc.t
+      | Disabled of Loc.t
```
rgrinberg: We have a type definition for this already. So this should be:

```
type t =
  | Unset
  | Set of Loc.t * Enabled.t (* or whatever this type is *)
```

rgrinberg (changes requested):

src/dune_config_file/dune_config_file.ml
Alizter: You mean `Config.Toggle.t`?

rgrinberg: Yeah

shonfeder (approved):
I like how this is shaping up and all my comments have been addressed. Thanks!

src/dune_rules/lock_dir.ml
Alizter: Yes, I think that makes sense. I'll change it back to how it was before and put the match in the else clause.

rgrinberg (approved):

PR #12446, opened 2025-09-11
fix(rpc): correct exit code when encountering an rpc error

### LINKED ISSUE #12445: rpc returns status 0 on connection terminated error
## Expected Behavior

When rpc connection is terminated unexpectedly, status code should be non zero.

## Actual Behavior

When rpc connection is terminated unexpectedly, status code is zero.

## Reproduction

- Open a terminal with a dune build in watch mode
- On another terminal, call `dune build <some_target>; echo "dune exit code: $?"`
- Kill the dune process in first terminal

Second terminal will show:

```
Error: { payload = Some [ [ "id"; [ "auto"; "0" ] ] ]
; message =
    "connection terminated. this request will never receive a response"
; kind = Connection_dead
}
dune exit code: 0
```

## Specifications

- Version of `dune` (output of `dune --version`): `3.19.1`
- Version of `ocaml` (output of `ocamlc --version`): `4.14.0`

Alizter:
Demonstrates and fixes the issue observed in #12445.

- fixes #12445
- changelog

### DIFF
--- a/bin/rpc/rpc_common.ml
+++ b/bin/rpc/rpc_common.ml
@@ -88,11 +88,7 @@ let wrap_build_outcome_exn ~print_on_success f args () =
   let open Fiber.O in
   let+ response = f args in
   match response with
-  | Error (error : Rpc_error.t) ->
-    Console.print
-      [ Pp.paragraphf "Error: %s" (Dyn.to_string (Rpc_error.to_dyn error))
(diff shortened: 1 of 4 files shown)


Alizter:
There are really two things to fix here:
1. The exit code as mentioned in #12445.
2. The message should be a user error rather than a dyn dump.

rgrinberg (approved):

PR #12448, opened 2025-09-12
flake.nix: cleanup, drop flake-utils, stop shadowing `pkgs`

faukah:
Changes done:
- Format using the official nixpkgs formatter instead of the outdated one.
- Stop shadowing `pkgs`.
- Drop `flake-utils`, instead opting for iterating over systems using pure Nix functions. 
- Fix removing of ocamlformat properly:
```nix
nativeBuildInputs = lib.remove pkgs.ocamlformat (testNativeBuildInputs pkgs);
```
Removes `pkgs.ocamlformat` from `testNativeBuildInputs`, while `testNativeBuildInputs` uses the `ocamlformat` versio from `.ocamlformat`. If these ever mismatch, this breaks.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -139,10 +139,6 @@
               findlib
             ];
 
-            buildInputs = lib.optionals pkgs.stdenv.isDarwin [
-              pkgs.darwin.apple_sdk.frameworks.CoreServices
-            ];
-
             strictDeps = true;
             buildFlags = [ "release" ];
             dontAddPrefix = true;
@@ -284,9 +280,6 @@
           slim-opam = pkgs.mkShell {
             inherit INSIDE_NIX;
             nativeBuildInputs = lib.remove ocamlformat (testNativeBuildInputs pkgs);
-            buildInputs = lib.optionals pkgs.stdenv.isDarwin [
-              pkgs.darwin.apple_sdk.frameworks.CoreServices
-            ];
             meta.description = ''
               provides a shell with just `opam` and minimal (external)
               dependencies to run the testsuite.";


Alizter:
Thanks for the PR. Can we split the first two commits off into a separate PR?

spikespaz:
@Alizter Why does it matter if you avoid a merge commit? They're already separate changes, to show the cause (changing the formatter) and the effect (after formatting). 

Edit: Though, they should be reversed.

Alizter:
@spikespaz Its easier for reviewers when PRs are more focused. I asked for the first two commits to be split off because it is uncontroversial and is just a formatting change. The remaining commits will require more attention so it doesn't make sense to tie the first two down. Not sure what you mean about avoiding a merge commit however.

flake.nix
```diff
@@ -197,31 +198,29 @@
...
-              duneScript = writeScriptBin "dune" ''
-                #!${stdenv.shell}
+              duneScript = pkgs.writeScriptBin "dune" ''
+                #!${pkgs.stdenv.shell}
                 "$DUNE_SOURCE_ROOT"/_boot/dune.exe $@
               '';
             in
 
-            pkgs'.mkShell {
+            pkgs.mkShell {
```
Alizter, line 203: Does this matter? I think we really wanted to shadow `pkgs` here but we can't so we introduced `pkgs'`, however I think its quite confusing to mix using the two.

Alizter:
I can't tell if bumping the flake lockfile caused the issues, but we best do that in a separate PR. There might be unrelated changes that we will have to make and it will cloud this PR.

flake.nix
faukah: The `mkShell` in `pkgs` and `pkgs'` are the same, so it makes no difference; I'd argue that this is better though because everyone knows `pkgs.mkShell`. For `pkgs'.mkShell` you'd have to look at `pkgs'` first to be 100% sure `mkShell` is what you think it is. 

It's wise to always keep the scope as minimal as possible for stuff like this.

anmonteiro:
replaced by #12945

PR #12451, opened 2025-09-12
fix: check for overlapping executable stanzas

Alizter:
I'm not certain if we need to version guard this since it was buggy behaviour before.

- [ ] changelog
- [ ] don't break enabled_if


Alizter:
@rgrinberg I am not sure what you mean by that. AFAICT the only breakage is because we have people doing:

```
(tests
 (names foo))
(tests
 (names bar))
```

I'm not sure if it makes sense to check for overlapping modules in this case. We could get around this by introducing a version guard for the check, but does it even make sense to fix this? Perhaps this is a non-fix and really the missing cmi message needs another fix.

PR #12452, opened 2025-09-14
refactor(boot): remove -map

rgrinberg:
This flag does not reflect how dune's regular rules work. Let's get rid of it and resolve module names using aliases ourselves.

This is needed for `(include_subdirs qualified)` as well. Not all alias files can be visible in this mode, therefore we cannot just provide them all when we only have a single invocation of `ocamldep`.

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -61,15 +61,40 @@ module Map = struct
     include Map.S
 
     val of_list : (key * 'a) list -> 'a t
+    val of_list_reduce : (key * 'a) list -> f:('a -> 'a -> 'a) -> 'a t
+    val to_list : 'a t -> (key * 'a) list
+    val filter_map : f:(key -> 'a -> 'b option) -> 'a t -> 'b t
   end
 
-  module Make (S : Map.OrderedType) = struct
+  module Make (S : Map.OrderedType) : S with type key = S.t = struct
     module M = Map.Make (S)
     open M
 
     [@@@ocaml.warning "-32"]
 
-    let of_list xs = List.to_seq xs |> of_seq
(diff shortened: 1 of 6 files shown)


Alizter:
There were some warning when compiling with `@check` so I've added annotations to ignore them.

boot/duneboot.ml
```diff
@@ -1727,7 +1938,19 @@ let get_dependencies libraries =
...
+      List.concat_map libraries ~f:(fun (lib : Library.t) ->
+        List.rev_map lib.ocaml_files ~f:(fun f -> f, lib))
+      |> String.Map.of_list
+    in
+    let toplevel_modules =
+      List.map libraries ~f:(fun (lib : Library.t) -> lib.group) |> Group.merge_all
+    in
+    let path_by_obj =
+      List.fold_left libraries ~init:Module.Obj.Map.empty ~f:(fun acc (lib : Library.t) ->
+        Module.Obj.Map.union ~f:(fun _ _ _ -> assert false) acc lib.path_by_obj)
```
Alizter, line 1912: Is this a todo for qualified?

rgrinberg: No, we expect all module object names to be unique.

boot/duneboot.ml
```diff
@@ -1756,7 +1991,7 @@ let get_dependencies libraries =
 ;;
 
 let assemble_libraries
-      { local_libraries; target = _, main; _ }
+      { local_libraries; target = _; _ }
```
Alizter:
```suggestion
      { local_libraries; _ }
```

Alizter (approved):

PR #12469, opened 2025-09-15
refactor(boot): simplify handling of architecture values.

Alizter:
Since we parse architecture values from the config, we might as well collapse `amd64` and `x86_64` since these are the same.

We also simplify the various places we handle architectures to use the same (poly) variant. This allows us to simplify some other parts and share some code.

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -375,14 +375,12 @@ module Arch = struct
   type t =
     [ `arm64
     | `amd64
-    | `x86_64
     | `other
     ]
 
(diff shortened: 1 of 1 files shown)


Alizter:
@rgrinberg This refactoring was left from our improvements to bootstrapping. If you would like to use it please do otherwise feel free to close.

rgrinberg:
The idea seems good. Is it ready?

Alizter:
I'll rebase and resolve the conflicts.

rgrinberg (approved):

PR #12471, opened 2025-09-15
refactor(boot): simplify handling of flags

Alizter:
We delay the expansion of flags per CPU feature by introducing a new CPU feature variant. This allows us to consolidate the expansion of flags in one place.

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -337,6 +337,39 @@ module Arch = struct
   ;;
 end
 
+module Cpu_feature = struct
+  type t =
+    [ `SSE2
+    | `SSE41
+    | `AVX2
+    | `AVX512
+    | `NEON
+    ]
+
+  let by_suffix fn =
+    if String.ends_with fn ~suffix:"_sse2"
+    then Some `SSE2
+    else if String.ends_with fn ~suffix:"_sse41"
+    then Some `SSE41
+    else if String.ends_with fn ~suffix:"_avx2"
+    then Some `AVX2
+    else if String.ends_with fn ~suffix:"_avx512"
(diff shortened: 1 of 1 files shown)


boot/duneboot.ml
```diff
@@ -1223,27 +1253,7 @@ module File_kind = struct
...
-        else if check "_avx2"
-        then x86 [ "-mavx2" ] [ "/arch:AVX2" ]
-        else if check "_avx512"
-        then x86 [ "-mavx512f"; "-mavx512vl"; "-mavx512bw" ] [ "/arch:AVX512" ]
-        else if String.ends_with fn ~suffix:"_neon"
-        then Some `Arm64, []
-        else None, []
-      in
-      Some (C { arch; flags })
+    | ".c" -> Some (C { cpu_feature = Cpu_feature.by_suffix fname })
```
rgrinberg: Why do we need to discover the "cpu feature" here? Why don't we do it with the rest of the flags?

Alizter: This is the "cpu feature" that the source file is asking for. Whether or not we can provide it is decided later.

rgrinberg: That's exactly the problem. If we can only decide what to do with this cpu feature later, why are we discovering it now? Might as well discover it when we need it.

Alizter: Good point, removed.

Alizter:
I've just realised we don't even have these C sources, we explicitly delete them in the vendor script. The only detection we are doing is between neon and assembly. I will simplify this more tomorrow.

rgrinberg:
Please re-open once ready

Issue #12304, opened 2025-08-25
chore: automatically generate CHANGES.md


maiste:
We should provide a way to automatically generate the CHANGES.md. It would save some time when doing the releases. Thanks to it, we could speed up the way the release is done. 

Suggestions are either to encode the type in the filename (e.g. `added-12043.md`) or adding the correct at the top with:

```markdown
## Added

Bla bla bla blabla [...]
```

A good inspiration to take from is the #8195 PR, which provides a `make changelog` command.

The end goal is to provide a GitHub Actions to open the release issue and generates the first alpha PR.

Alizter:
If we know our categories before hand, we could use subdirectories for categorization. I.e. `doc/changes/added`, `doc/changes/fixed`. That would make it easier for people to introduce changelogs without having to worry about file naming schemes or headers.

Alizter:
For the compilation of the changelog, we would have a dune rule that globs each category subdirectory, cats them into a file, formats that file and then concatenates it to the top of the changelog.

maiste:
Is it necessary to have a dune rune for this? Can't it simply be a script if we go for the sub directory structure?

PR #12472, opened 2025-09-16
doc: restructure changelog

maiste:
This PR is based on top of the #12440. It **must not** be merged until the other PR is merged. The PR updates the structure and replaces it with three categories for the Changelog (`added`, `changed` and, `fixed`). It also updates the `CONTRIBUTING.md` with an explanation on how it works.

This PR is the first part of #12304. It will help to generate the changelog in automatically and simplify the release process.

(To be review, you should only take the last three commits into account)


maiste:
@Alizter and @shonfeder I have rebased the PR on the `main` branch with the CONTRIBUTING.md change. It should be good to merge this one to make sure the new addition will stick to the new structure.

CONTRIBUTING.md
```diff
@@ -131,3 +134,16 @@ you have the rights to submit this work under the same open source license.
 [js]: https://www.janestreet.com/
 [tarides]: https://tarides.com/
 [hack]: ./doc/hacking.rst
+
+Updating the Changelog
+----------------------
+
+In Dune, the Changelog is in `CHANGES.md`. As they are many contributions at
```
shonfeder:
```suggestion
In Dune, the Changelog is in `CHANGES.md`. As there are many contributions at
```

CONTRIBUTING.md
```diff
@@ -131,3 +134,16 @@ you have the rights to submit this work under the same open source license.
...
+the same time and to prevent an infinite cycle of rebase, the `CHANGES.md` file
```
shonfeder:
```suggestion
the same time and to prevent an infinite cycle of rebases, the `CHANGES.md` file
```

CONTRIBUTING.md
```diff
@@ -131,3 +134,16 @@ you have the rights to submit this work under the same open source license.
...
+is only updated when releasing Dune. To keep track of the changes, they must be
```
shonfeder:
```suggestion
is only updated when releasing Dune. Unreleased changes are
```

CONTRIBUTING.md
```diff
@@ -131,3 +134,16 @@ you have the rights to submit this work under the same open source license.
...
+recorded inside `doc/changes/`. The name of the file is `<PR number>.md`.
+Depending on the nature of the change, it can live inside three categories:
+`added`, `changed`, `fixed`, where:
+- `added` records every new feature or behaviour inside of Dune
```
shonfeder:
```suggestion
- `added` records every new feature or behaviour visible to users
```

Suggested because "inside dune" could be read to suggest even changes not visible to users should get a changelog entry (which I assume is not the case) and because it goes without saying that the changelog only applies to dune, I think.

CONTRIBUTING.md
```diff
@@ -131,3 +134,16 @@ you have the rights to submit this work under the same open source license.
...
 [tarides]: https://tarides.com/
 [hack]: ./doc/hacking.rst
+
+Updating the Changelog
+----------------------
+
+In Dune, the Changelog is in `CHANGES.md`. As they are many contributions at
+the same time and to prevent an infinite cycle of rebase, the `CHANGES.md` file
+is only updated when releasing Dune. To keep track of the changes, they must be
+recorded inside `doc/changes/`. The name of the file is `<PR number>.md`.
```
shonfeder: Is it actually necessary that the file just be numbered? Can it have a descriptive name too/instead?

CONTRIBUTING.md
```diff
@@ -131,3 +134,16 @@ you have the rights to submit this work under the same open source license.
...
+Depending on the nature of the change, it can live inside three categories:
+`added`, `changed`, `fixed`, where:
+- `added` records every new feature or behaviour inside of Dune
+- `changed` records a difference of behaviour with the previous version
+- `fixed` records any change that restores the behaviour as expected
```
shonfeder, line 149: An explanation of the format, with an example, would also be helpful here (including the conventions for referencing the PR and author).

How are breaking changes indicated?

CONTRIBUTING.md
maiste: As we only append the files in `CHANGES.md` and they should be reviewed before being merged into `main`, it is not necessary to have a descriptive name. Using the number of the PR makes it easier to retrieve when doing patch release (where we want to target small changes).

CONTRIBUTING.md
maiste: I don't think we specify the breaking changes, as the _SemVer_ convention is supposed to make this very clear by increasing the major version number.

Even if it is not entirely true (as we have seen with the formatting), `dune` is "not supposed" to introduce breaking changes without changing the `lang dune` version. 

However, I would be happy to add a convention for such cases. Apart from a **BREAKING CHANGE** in front of the entry, I don't see anything else. Would you have any suggestion here?

Alizter: You're correct that breaking changes are uncommon in dune, in that we do our best to avoid them. Since those cases are rare, I don't think it makes sense to add special support for it. When they do happen we can manually edit the changelog to highlight the severity.

maiste:
Thanks for the review @shonfeder! I have applied your suggestions and reworked the part about the format. I added a paragraph about *BREAKING CHANGE*. Does it work for you?

CONTRIBUTING.md
maiste: I pushed a fix with a breaking change part before seeing your comment. Let me know if the wording is not emphasizing enough the exceptional aspect of breaking change. If it does not work, I can remove it.

Alizter (approved):
Looks good to me. We will need to coordinate this with other PRs when merging.

maiste:
Indeed! I expect a period where we will have to manually move the markdown files until the new structure is fully adopted.

PR #12473, opened 2025-09-17
Allow running tests concurrently with watch server, again!

ElectreAAS:
Contributed to https://github.com/ocaml/dune/issues/8114
Supersedes #11900.

This PR is a 'remake' of Steve's: I rebased it on main, and applied what I learned making #12010 and #12064. This results in significant changes from the original, hence the new PR.

The 'runtest' RPC request is now public, in otherlibs. `runtest_rpc.ml{i}` was removed.

The behaviour is now (AFAI know what I'm doing) identical between running `dune runtest [ARGUMENTS]` with or without the RPC server.

See comments below for details

### DIFF
--- a/test/blackbox-tests/test-cases/cram/custom-build-dir.t
+++ b/test/blackbox-tests/test-cases/cram/custom-build-dir.t
@@ -21,7 +21,8 @@ path
     $TESTCASE_ROOT/tmp/default/foo.t.corrected
     to foo.t.
   [1]
-  $ sed -E '/\(pid: [0-9]+\)/{s//(pid: ###)/; s/instance.*/.../g; q;}' foo.t
+  $ cat foo.t
     $ echo "  $ echo bar" >bar.t
     $ dune runtest
-    Error: A running dune (pid: ###) ...
+    Error: RPC server not running.
+    [1]
--- a/bin/build.ml
+++ b/bin/build.ml
@@ -143,10 +143,10 @@ let run_build_command ~(common : Common.t) ~config ~request =
     ~request
 ;;
 
-let build_via_rpc_server ~print_on_success ~targets =
+let build_via_rpc_server ~print_on_success ~targets builder lock_held_by =
   Rpc.Rpc_common.wrap_build_outcome_exn
     ~print_on_success
-    (Rpc.Group.Build.build ~wait:true)
+    (Rpc.Group.Build.build ~wait:true builder lock_held_by)
     targets
     ()
 ;;
@@ -204,11 +204,9 @@ let build =
          perform the RPC call.
       *)
       Rpc.Rpc_common.run_via_rpc
-        ~builder
         ~common
         ~config
-        lock_held_by
-        (Rpc.Group.Build.build ~wait:true)
+        (Rpc.Group.Build.build ~wait:true builder lock_held_by)
         targets
     | Ok () ->
       let request setup =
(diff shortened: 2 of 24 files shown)


test/blackbox-tests/test-cases/cram/custom-build-dir.t
```diff
@@ -10,6 +10,8 @@ path
   >   $ dune runtest
   > EOF
 
+  $ dune rpc ping --wait
+
```
ElectreAAS: In my testing, this test times out every time. This ping is to help debug it

ElectreAAS:
Notes from rebasing: Since the contents of `bin/runtest.ml` have been largely moved to `bin/runtest_common.ml`, it'd be easy to accidentaly overwrite changes made to the former since the original PR.
#11869 and #12251 are PRs that fit that description, and have been correctly applied to `runtest_common`

test/blackbox-tests/test-cases/cram/custom-build-dir.t
```diff
@@ -21,7 +21,11 @@ path
...
   [1]
-  $ sed -E '/\(pid: [0-9]+\)/{s//(pid: ###)/; s/instance.*/.../g; q;}' foo.t
+  $ sed -E 's/[0-9]+\)/###)/g' foo.t
     $ echo "  $ echo bar" >bar.t
     $ dune runtest
-    Error: A running dune (pid: ###) ...
+    Warning:
+    Your build request is being forwarded to a running Dune instance (pid:
+    ###). Note that certain command line arguments may be ignored.
+    Error: RPC server not running.
```
ElectreAAS, line 27: The behaviour observed in this test is, imo, correct, but the error messages aren't helping at all.
Since DUNE_BUILD_DIR is set, the `dune runtest` of line 13 creates a lockfile that the `dune runtest` of line 10 will see. The inner dune will then try to send an RPC request to the outer dune, but this one wasn't launched in watch mode, and hence doesn't listen to RPC requests at all. The "server not running" message is correct, but we should better explain to the user what's going on

ElectreAAS: Solved by latest commit

ElectreAAS:
### Outdated
The last commit (27412bcf3d8bc6393ac89bded7b556bea53db8e3) is maybe a bit overzealous in the refactoring, but that was the only way I found to make sure the error messages were correct.
Review can be done without it, and if deemed necessary I could put it in a different PR

### Up to date
I just force-pushed a few changes:
- it's now rebased on main
- there are now only 2 commits:
  - the first is the one that has only what is necessary for the feature, without previous refactorings
  - the second has all the plumbing required to make the 'forwarding' warning appear at the correct time (after actually connecting to the RPC server)

It's now both ready to review **and** reviewable! :smiley: 

PS: the previous state of this PR is still visible [on my fork](https://github.com/ElectreAAS/dune/tree/test-concurrent-backup), as a backup

test/blackbox-tests/test-cases/watching/watching-eager-concurrent-build-command.t
```diff
@@ -43,8 +46,7 @@ acting as an RPC client while running concurrently with an RPC server.
                ^^^^^^^^^^^^
   Unbound value print_endlin
   Hint: Did you mean print_endline?
-  Error: Build failed with 1 error.
-  [1]
+  Build failed with 1 error.
```
Alizter: The exit code appears to be lost here.

ElectreAAS: It was a bug already present in main, the fix didn't belong in this PR and has been pulled out

Alizter (approved):

PR #12474, opened 2025-09-18
Download the latest release for a manual trigger of Docker image generation

Sudha247:
Per the discussion in #12252, this PR adds an additional step to download release assets from the latest release if the current tag number doesn't exist, i.e., in the case of a manual trigger of the GitHub Actions job.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,3 +1,4 @@
+
 name: Docker Release
 
 on:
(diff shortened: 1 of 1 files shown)


Alizter (approved):

Alizter:
We can run it in branches in this repo, just not for your fork's branch. It might work if you enable github actions in your fork, then you can manually run there to see if it is working. I'm however OK with merging this as is so we can test and fix it asap.

Alizter:
https://github.com/ocaml/dune/actions/runs/17837047968

Let's see what happens

Alizter:
Looks like it didn't work. It looks like it is still looking for a non-existent release. My advice would be to try fixing it in your fork first, that way you can get quicker feedback on the issues.

Sudha247:
Thanks for checking, @Alizter! I thought the ref would be null for non-release, but it takes the branch name as the ref and hence the failure. I'll fix it in my fork and send in a fix.

PR #12476, opened 2025-09-18
fix(boot): root module handling in the presence of (include_subdirs q…

rgrinberg:
…ualified)

We should add the root module to the trie of modules for `(include_subdirs qualified)` as well

### DIFF
--- a/boot/duneboot.ml
+++ b/boot/duneboot.ml
@@ -1543,17 +1543,22 @@ module Library = struct
     =
     let sources =
       let sources =
+        let cons_opt (what : File_kind.ml option) trie =
+          match what with
+          | None -> trie
+          | Some x -> String.Trie.add_exn trie [ Module.Name.to_fname x.name ~kind:`Ml ] x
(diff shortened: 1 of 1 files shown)


boot/duneboot.ml
```diff
@@ -1543,17 +1543,22 @@ module Library = struct
...
+        let cons_opt (what : File_kind.ml option) trie =
+          match what with
+          | None -> trie
+          | Some x -> String.Trie.add_exn trie [ Module.Name.to_fname x.name ~kind:`Ml ] x
+        in
         String.Trie.filter_map sources ~f:(fun { Source.file = _; kind } ->
           match (kind : File_kind.t) with
           | Asm _ | Header | C _ -> None
           | Ml module_ -> Some module_)
+        |> cons_opt build_info_module
```
Alizter, line 1555: Why not do this directly in the Some case, saves us having to match up there.

Alizter (approved):

PR #12477, opened 2025-09-19
@doc-markdown for odoc

davesnx:
This PR extends odoc to support markdown-generate implemented here https://github.com/ocaml/odoc/pull/1341 and supported since odoc.3.1.0

```bash
dune build @doc-markdown
```

Added the odoc version check, similar to other tools.

The implementation itself tried to be very similar to the html generation, except the directory targets on a flat package (it needed to be a system action that groups all markdown-genetaration)

### DIFF
--- a/src/dune_rules/odoc.ml
+++ b/src/dune_rules/odoc.ml
@@ -162,6 +162,12 @@ module Output_format = struct
   ;;
 end
 
+let output_dir_for_format ctx format target =
+  match (format : Output_format.t) with
+  | Html | Json -> Paths.html ctx target
+  | Markdown -> Paths.markdown ctx target
+;;
+
 module Dep : sig
   (** [format_alias output ctx target] returns the alias that depends on all
       targets produced by odoc for [target] in output format [output]. *)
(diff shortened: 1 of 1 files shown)


Alizter:
I think a version check is a good idea. We have such things for `ocaml`, `git`, `coq` and perhaps others.

src/dune_rules/odoc.ml
```diff
@@ -892,22 +1102,132 @@ let setup_pkg_html_rules sctx ~pkg : unit Memo.t =
...
+;;
+
+let markdown_shell_command odoc_path all_odocs ~markdown_root =
+  List.map all_odocs ~f:(fun odoc ->
+    let odocl_rel =
+      Path.reach (Path.build odoc.odocl_file) ~from:(Path.build markdown_root)
+    in
+    Printf.sprintf "%s markdown-generate -o . %s" odoc_path odocl_rel)
+  |> String.concat ~sep:" && "
+;;
```
davesnx, line 1130: This is technically working, but I don't think is a proper solution.

I pushed a new PR https://github.com/ocaml/dune/pull/12581 assuming that odoc supports multiple inputs as implemented here https://github.com/ocaml/odoc/pull/1388

It's a bit unforunate since we will need to wait for this PR to merge (and increase the version check)

rgrinberg:
Has been subsumed by https://github.com/ocaml/dune/pull/12581

PR #12482, opened 2025-09-20
refactor(boot): use our [of_list] for all versions

rgrinberg:
This version is safer as it doesn't just overwrite duplicate elements


Alizter:
Shouldn't it come after the include M?

Alizter (approved):

PR #12487, opened 2025-09-22
Remove most backtracking from lockdir parser

### LINKED ISSUE #12248: pkg: on dune build start, lock dir parsing is slow
Currently, a null build in my tiny repo is surprisingly slow (8s).

A significant fraction of that is caused by sexp parsing of the lock dir, specifically the fact that the error path of some combinators like `Dune_sexp.Decoder.find_cstr` is expensive (due to `User_message.did_you_mean`), and that `Dune_sexp.Decoder.either` means you can run a ton of error paths, instead of just one as would be normal.

With these depends in my dune-project:

```
(package
  ...
 (depends
   ocaml
   core
   core_unix
   markup
   uri
   uucp
   yojson
   http-lwt-client
   ppx_yojson_conv
   ppx_sexp_value
   ppx_sexp_conv
   ppx_compare
   ppx_hash
   ppx_pipebang
   ppx_let
   ppx_bin_prot
   ppx_partial
   lwt_ppx
   ocamlformat
 ))
```

I observe this:
```terminal
$ time ~/code/dune/_build/default/bin/main.exe build --display=quiet

real 0m7,809s
user 0m7,363s
sys 0m0,450s
```

but when I comment out all the `~hints` from `Dune_sexp.Decoder`:

```terminal
$ time ~/code/dune/_build/default/bin/main.exe build --display=quiet

real 0m4,960s
user 0m4,512s
sys 0m0,450s
```

It seems to me that either the sexp decoding module shouldn't backtrack, or the errors should be lazy, i.e. only computed on display.

(even with that, the performance seems off by an order of magnitude, possibly two. There's 130kB of data in the lock dirs and 1 dune file and 2 .ml file. It's wild that this takes any noticeable time at all)

gridbugs:
To parse both portable and non-portable lockdirs with mostly the same code, the parser would attempt to parse fields with the non-portable lockdir syntax and fall back to the portable syntax should parsing fail. This led to the generation of parse errors when the first attempt at parsing a field would fail, and even though these errors are discarded, generating them took a noticeable amount of time.

This changes the parsing logic to only attempt to parse fields in a single syntax, determined by the presence of the `solved_for_platforms` field in the lockdir metadata file. This field is present precisely when a lockdir is portable.

Fixes https://github.com/ocaml/dune/issues/12248

### DIFF
--- a/src/dune_pkg/lock_dir.ml
+++ b/src/dune_pkg/lock_dir.ml
@@ -375,13 +375,15 @@ module Build_command = struct
          ]
   ;;
 
-  let decode_fields_backwards_compatible =
+  let decode_fields ~portable_lock_dir =
     let open Decoder in
     let parse_action =
-      (let+ action = Action.decode_pkg in
-       Conditional_choice_or_all_platforms.Choice
-         (Conditional_choice.singleton_all_platforms (Action action)))
-      <|> Conditional_choice_or_all_platforms.decode decode_portable
+      if portable_lock_dir
+      then Conditional_choice_or_all_platforms.decode decode_portable
+      else
+        let+ action = Action.decode_pkg in
+        Conditional_choice_or_all_platforms.Choice
+          (Conditional_choice.singleton_all_platforms (Action action))
     in
     fields_mutually_exclusive
       ~default:None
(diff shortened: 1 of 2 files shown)


rgrinberg (approved):
LGTM. 3.9s is still rather slow though. I think there should be more low hanging fruit to optimize here.

v-gb:
> LGTM. 3.9s is still rather slow though. I think there should be more low hanging fruit to optimize here.

Indeed. In the issue, I wrote down my investigation of the remaining slowness.

rgrinberg:
Thanks for the thorough investigation. Do you want to send PR's for those fixes? Seems a bit silly if I am just going to redo the work from your notes.

v-gb:
> Thanks for the thorough investigation. Do you want to send PR's for those fixes? Seems a bit silly if I am just going to redo the work from your notes.

I sent the changes I have, but I don't have a broad understanding of the code, so while the changes are enough to prove that dune should be much faster, maybe some other change would be even better.

PR #12488, opened 2025-09-22
tests: Enable package management by default in test and add helpers to disable it

Leonidas-from-XIV:
This is pulling out a change from #12394 to make it easier to control whether package management should be force-enabled, force-disabled or left at the default setting.


Alizter:
I'm not certain about this. Doesn't this mean we stop testing the default setting?

Leonidas-from-XIV:
The default setting is still tested with `unset_pkg`.

ElectreAAS (approved):
Lgtm!

PR #12489, opened 2025-09-22
tests: Enable dev-tool flag when running dev-tools

Leonidas-from-XIV:
This is factored out of #12394 and makes sure that the dev tool flag is set when attempting to use dev-tool functionality.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/ocamlformat/ocamlformat-install.t
+++ b/test/blackbox-tests/test-cases/pkg/ocamlformat/ocamlformat-install.t
@@ -7,10 +7,10 @@ Test `dune tools which ocamlformat`:
   $ make_project_with_dev_tool_lockdir
 
 Install ocamlformat as a dev tool:
-  $ dune tools install ocamlformat
+  $ DUNE_CONFIG__LOCK_DEV_TOOL=enabled dune tools install ocamlformat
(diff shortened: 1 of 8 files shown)


Sudha247 (approved):
LGTM! (The CI failures look unrelated)

Sudha247:
On a related note, but not to block this PR, this made me look at Dune flag conventions a bit. Do we have them documented somewhere (my basic grep and search didn't work)?

Leonidas-from-XIV:
These specific toggles are set up in `src/dune_rules/compile_time.ml` and use the toggle mechanism from `Dune_config.Config`, which does all the `DUNE_CONFIG__*` prefixing. Aside from that I am not aware of any other conventions.

gridbugs:
My understanding was that `DUNE_CONFIG__LOCK_DEV_TOOL=enabled` has no effect on commands under `dune tools ...`. Its only purpose is to allow existing commands like `dune fmt` and `dune utop` to install their requisite tools automatically.

Leonidas-from-XIV:
The problem is that with the copy rules, the rules don't know whether it was launched via `dune tools` or not so it makes sense to me to require the flag everywhere and be consistent.

And eventually just enable it by default and remove the flag, but that's a different PR for some other time.

gridbugs (approved):

Issue #12390, opened 2025-09-03
`dune pkg` no longer picks up the compiler from its cache


shym:
When building the starter project with dune package management on dune ≥ 3.20, the compiler is built on every run (even though it appears only once in `~/.cache/dune/toolchains/`, so it’s not because the hash changed between the two builds).

```
$ rm -rf ~/.cache/dune
$ dune init project toy
[...]
$ cd toy
$ dune --version
3.20.1
$ dune clean
$ time dune build
 Downloading ocaml-compiler.5.3.0
    Building ocaml-compiler.5.3.0
    Building ocaml-base-compiler.5.3.0

real    1m40,687s
user    7m11,492s
sys     0m28,826s

$ dune clean
$ time dune build
 Downloading ocaml-compiler.5.3.0
    Building ocaml-compiler.5.3.0
    Building ocaml-base-compiler.5.3.0

real    1m44,387s
user    7m11,721s
sys     0m29,883s
```

Downgrading to 3.19.1 solves the issue.

```
$ rm -rf ~/.cache/dune
$ dune --version
3.19.1
$ dune pkg lock
[…]
$ time dune build
 Downloading ocaml-compiler.5.3.0
    Building ocaml-compiler.5.3.0
    Building ocaml-base-compiler.5.3.0

real    1m45,019s
user    7m18,082s
sys     0m30,100s

$ dune clean
$ time dune build
 Downloading ocaml-compiler.5.3.0
    Building ocaml-compiler.5.3.0
    Building ocaml-base-compiler.5.3.0

real    0m4,287s
user    0m0,725s
sys     0m0,521s
```

I’ve run this test with opam-built 3.19.1, 3.20.0, 3.20.1 and with binary distribution of 3.19.1, 3.20.1 and developer preview. All concord that the compiler gets rebuilt starting from 3.20.0.

Alizter:
At the moment this is expected. See #12179 for proposed fixes.

The compiler is non-relocatable which means we cannot cache it in our content-addressed cache. If we did, then files would be in unexpected places due to how we hardlink.

As discussed in #12179 we need to cache the compiler cookie intself, so that can be restored between rebuilds. However, that also brings in some technical complications since we cannot cheaply check the integrity of the contents of the cookie, i.e. where it is pointing. If the cookie points to a deleted toolchain for instance, there would be no way to invalidate it without doing pointless work.

My only advice to users at the moment is to stop deleting `_build` and do something a bit more selective like `rm -rf _build/default`.

[...]

PR #12495, opened 2025-09-24
Don't use Digest.generic for toolchain hash

gridbugs:
The Digest.generic function marshals a value to a string and then hashes the string. For types which contain sets and maps, two values constructed in different orders but which are considered equal may have different in-memory representations and so marshal to different strings and hash to different hashes.

Compiler toolchain directories created by dune have the package hash appended to their names. Using Digest.generic meant that sometimes Dune would create a new toolchain directory when re-using an existing one would suffice.

Note that this change alone will not fix
https://github.com/ocaml/dune/issues/12390.

### DIFF
--- a/src/dune_rules/pkg_toolchain.ml
+++ b/src/dune_rules/pkg_toolchain.ml
@@ -27,13 +27,15 @@ let pkg_dir (pkg : Dune_pkg.Lock_dir.Pkg.t) =
      way). *)
   let dir_name =
     (* TODO should include resolved deps *)
-    let pkg_hash = Digest.generic (Lock_dir.Pkg.remove_locs pkg) in
+    let pkg_digest =
+      Dune_digest.Feed.digest Lock_dir.Pkg.digest_feed (Lock_dir.Pkg.remove_locs pkg)
+    in
     (* A hash of the fields of a package that affect its installed artifacts *)
     sprintf
       "%s.%s-%s"
       (Package.Name.to_string pkg.info.name)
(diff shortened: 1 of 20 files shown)


Leonidas-from-XIV (approved):

src/dune_rules/pkg_toolchain.ml
```diff
@@ -27,13 +27,13 @@ let pkg_dir (pkg : Dune_pkg.Lock_dir.Pkg.t) =
      way). *)
   let dir_name =
     (* TODO should include resolved deps *)
-    let pkg_hash = Digest.generic (Lock_dir.Pkg.remove_locs pkg) in
+    let pkg_hash = Lock_dir.Pkg.hash (Lock_dir.Pkg.remove_locs pkg) in
```
rgrinberg: Can't we keep using a cryptographic hash function? Regular hash functions like this can easily collide.

gridbugs: We could. We'd just need to thread a cryptographic hashing function through `Lock_dir.Pkg.t` the same way `hash` is threaded.

rgrinberg: I think we should do that then.

gridbugs: Done. I've extended the `Digest` module a little to make it easier to thread a hash function through custom types.

rgrinberg (approved):

PR #12496, opened 2025-09-24
tests: Fix breakage on macOS with BSD `sed`

Leonidas-from-XIV:
In #12488 I missed the fact that `sed` on macOS was failing, this PR aims to fix the issue.

### DIFF
--- a/test/blackbox-tests/test-cases/pkg/helpers.sh
+++ b/test/blackbox-tests/test-cases/pkg/helpers.sh
@@ -58,7 +58,7 @@ mkpkg() {
 set_pkg_to () {
   local value="${1}"
   if grep "(pkg .*)" dune-workspace > /dev/null; then
-    sed -i'' "s/(pkg .*)/(pkg ${value})/" dune-workspace
+    sed -i.bak "s/(pkg .*)/(pkg ${value})/" dune-workspace
   else
     echo "(pkg ${value})" >> dune-workspace
   fi
@@ -73,7 +73,7 @@ disable_pkg() {
 }
 
 unset_pkg() {
-  sed -i'' "/(pkg/d" dune-workspace
+  sed -i.bak "/(pkg/d" dune-workspace
 }
 
 add_mock_repo_if_needed() {


MisterDA:
You can also do `sed -i.bak ... && rm dune-workspace.bak`

Leonidas-from-XIV:
Yes, that's true (and could even keep the `.bak` since its existence doesn't make a difference to dune). However most times I use `sed` I end up regretting it because GNU `sed` and BSD `sed` behave fairly differently on rather basic tasks and `sed`'s support for regular expressions is quite limited.

MisterDA:
> Yes, that's true (and could even keep the `.bak` since its existence doesn't make a difference to dune). However most times I use `sed` I end up regretting it because GNU `sed` and BSD `sed` behave fairly differently on rather basic tasks and `sed`'s support for regular expressions is quite limited.

With POSIX 2024 the behavior is better. sed now supports [Extended Regular Expressions](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap09.html#tag_09_04). POSIX standardizes a behavior that's been supported both by GNU and BSD sed for a long time, with `sed -E`.

Leonidas-from-XIV:
Is POSIX 2024 supported by the version of `sed` that is shipped with macOS?

art-w:
It looks like `sed -i.bak` is used in other tests, so I assume it works, but then there are still a few `sed -i ...` in other tests. Should those also be fixed? (it's a shame that the "solution" on mac would be to write `sed -i ''`, which doesn't work on linux >_>)

Leonidas-from-XIV:
I think the issue is that `-i''` kind of works cross-platform (so the other tests might be working just fine), but not when combined with the `d` command (as you see in the error message) as in this case.

I've changed it to `-i.bak` as @MisterDA suggested, let's see what the macOS CI run will think about it.

Leonidas-from-XIV:
That seem to have solved it and the tests succeed now just fine.

Alizter (approved):
I've done some similar changes elsewhere.

PR #12498, opened 2025-09-24
tests: Reduce testing of arbitrary slang expressions

Leonidas-from-XIV:
The point of using Slang in lock files is to encode OPAM expressions in a Dune form. This means that we just need to test that we can encode these.

There is no need to test for a large range of invalid inputs because the lock files are not supposed to be written by hand. They are supposed to be emitted by Dune and if Dune is emitting invalid Slang it is a problem of the emitting code and needs to be fixed.

This PR removes some error messages that are emitted by the Slang decoder - they are not specific to Dune pkg and should be tested elsewhere (e.g. `run-parse-error.t` tests for unexpected lists).


Sudha247 (approved):
I've looked at the deleted cases and agree that it's unlikely we encounter them in auto generated `pkg` files.

PR #12506, opened 2025-09-26
Fix build issues in the release docker image workflow

Sudha247:
Fixes the GitHub Actions workflow for releasing Docker images. This has been tested in my local fork (https://github.com/Sudha247/dune/pkgs/container/dune), with the help of @Leonidas-from-XIV.

One change here is, a manual trigger of the job requires the version number as input. I'm inclined to keep it this way to avoid confusion on which version is being extracted, but I'm open to hear other ideas.

### DIFF
--- a/docker/release.Dockerfile
+++ b/docker/release.Dockerfile
@@ -1,25 +1,40 @@
 ARG BASE_IMAGE=debian:13
 FROM ${BASE_IMAGE} AS build
 
-RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser build-essential ocaml
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install -y adduser build-essential ocaml
 
 RUN addgroup --gid 1000 dune && adduser --uid 1000 --ingroup dune dune
 
 USER dune:dune
 
-COPY --from=tarball /dune.tbz /home/dune/dune.tbz
 WORKDIR /home/dune
+COPY --from=tarball /dune.tbz /home/dune/dune.tbz
 
 # Use release assets from Docker context
 RUN tar -xf dune*.tbz && mv dune-*/ dune
 
 # Build and install dune
-RUN cd dune && ./configure --prefix=/home/dune/install && make bootstrap && make release && make install
+RUN cd dune && ./configure --prefix=/home/dune/install && \
+    make bootstrap && \
+    make release && \
+    make install
 
 FROM ${BASE_IMAGE} AS run
 
-RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser curl ocaml git sudo && rm -rf /var/lib/{apt,dpkg,cache,log}/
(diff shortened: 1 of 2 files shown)


docker/release.Dockerfile
```diff
@@ -16,7 +18,7 @@ RUN cd dune && ./configure --prefix=/home/dune/install && make bootstrap && make
 
 FROM ${BASE_IMAGE} AS run
 
-RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser curl ocaml git && rm -rf /var/lib/{apt,dpkg,cache,log}/
+RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser curl ocaml git sudo && rm -rf /var/lib/{apt,dpkg,cache,log}/
```
Leonidas-from-XIV: What do we need `sudo` for here?

docker/release.Dockerfile
```diff
@@ -6,6 +6,8 @@ RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser build-ess
 RUN addgroup --gid 1000 dune && adduser --uid 1000 --ingroup dune dune
 
 USER dune:dune
+
+COPY --from=tarball /dune.tbz /home/dune/dune.tbz
 WORKDIR /home/dune
```
Leonidas-from-XIV, line 12: I believe if you flip `WORKDIR` and `COPY` here you can cache one more layer, as the cache only gets invalidated by `COPY` commands.

docker/release.Dockerfile
```diff
@@ -16,7 +18,7 @@ RUN cd dune && ./configure --prefix=/home/dune/install && make bootstrap && make
 
 FROM ${BASE_IMAGE} AS run
 
-RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser curl ocaml git && rm -rf /var/lib/{apt,dpkg,cache,log}/
+RUN apt-get update && apt-get upgrade -y && apt-get install -y adduser curl ocaml git sudo && rm -rf /var/lib/{apt,dpkg,cache,log}/
 RUN addgroup --gid 1000 dune && adduser --uid 1000 --ingroup dune dune
```
Leonidas-from-XIV: This instruction can be merged with the previous `RUN` line, then it gets cached in the same layer.

Leonidas-from-XIV (commented):
Some optimization options that I have noticed.

docker/release.Dockerfile
Sudha247: I tried using this as a base image to build the `binary-distribution` site, but it was impossible to install any package there, as `sudo` was not present. I thought it would be useful to include it in the base image. But maybe there are other ways to achieve this?

Leonidas-from-XIV: Hmm, I see. That's a g good point.

Leonidas-from-XIV (approved):

PR #12511, opened 2025-09-26
fix: -H for private modules

rgrinberg:
The implementation of private modules relies on hiding include flags. This doesn't really work well, and OCaml has introduced a better mechanism to hide modules from users: the `-H` flag. We now use this flag (when available) to make sure that the compiler can always see hidden modules.


rgrinberg:
Replaced by https://github.com/ocaml/dune/pull/12666

PR #12514, opened 2025-09-29
Remove locs from slangs in [Action.remove_locs]

gridbugs:
Fixes a bug where we weren't properly removing locations from actions. Slang expressions that are parsed from files contain locations, and these location were not being removed from actions.

### DIFF
--- a/src/dune_lang/action.ml
+++ b/src/dune_lang/action.ml
@@ -524,13 +524,13 @@ let validate ~loc t = ensure_at_most_one_dynamic_run ~loc t
 let rec blang_map_string_with_vars ~f = function
   | Blang.Const _ as c -> c
   | Not blang -> Not (blang_map_string_with_vars ~f blang)
-  | Expr sw -> Expr (f sw)
+  | Expr sw -> Expr (slang_map_string_with_vars ~f sw)
   | And blangs -> And (List.map blangs ~f:(blang_map_string_with_vars ~f))
   | Or blangs -> Or (List.map blangs ~f:(blang_map_string_with_vars ~f))
-  | Compare (op, a, b) -> Compare (op, f a, f b)
-;;
+  | Compare (op, a, b) ->
+    Compare (op, slang_map_string_with_vars ~f a, slang_map_string_with_vars ~f b)
 
-let rec slang_map_string_with_vars ~f = function
+and slang_map_string_with_vars ~f = function
   | Slang.Nil -> Slang.Nil
   | Literal sw -> Literal (f sw)
   | Form (loc, form) ->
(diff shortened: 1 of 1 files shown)


src/dune_lang/action.ml
```diff
@@ -608,7 +608,44 @@ let rec map_string_with_vars t ~f =
   | Format_dune_file (src, dst) -> Format_dune_file (f src, f dst)
 ;;
 
-let remove_locs = map_string_with_vars ~f:String_with_vars.remove_locs
+let rec map_slang t ~f_slang ~f_blang =
```
Leonidas-from-XIV: Maybe just call it `~slang` and `~blang` because the reason the regular map argument is called `f` is because there's no good alternative. Here they are pretty descriptive and the fact that they're functions is obvious from the signature.

src/dune_lang/action.ml
```diff
@@ -608,7 +608,44 @@ let rec map_string_with_vars t ~f =
...
+  | Substitute (i, o) -> Substitute (i, o)
+  | Withenv (ops, t) -> Withenv (ops, map_slang t ~f_slang ~f_blang)
+  | When (condition, t) -> When (f_blang condition, map_slang t ~f_slang ~f_blang)
+  | Format_dune_file (src, dst) -> Format_dune_file (src, dst)
+;;
+
+let remove_locs t =
+  map_string_with_vars t ~f:String_with_vars.remove_locs
+  |> map_slang ~f_slang:Slang.remove_locs ~f_blang:Slang.Blang.remove_locs
+;;
```
Leonidas-from-XIV, line 619: I feel like it would be nice to have a general `map` type of function that maps them both. At the moment the action is traversed twice, once to map `string_with_vars` and once to traverse `slang`.

That way `map_string_with_vars` and `remove_locs` could be implemented on top of a more general `Action.map ~slang ~blang ~string_with_vars`  function (thus saving on code matching on all the cases in multiple places and iterating over data structures multiple times).

Leonidas-from-XIV (commented):
Good catch and while I don't have much to complain about the code itself (it is basically just a pattern match), I do think that this is a good occasion to implement a more generic mapping function.

I am implementing a similar functionality where I map locations and being able to have one function to map over an action instead of having to remember to map over the `slang` as well would be a nice improvement (and potentially fix an issue I've been seeing).

src/dune_lang/action.ml
gridbugs: Good idea. Done!

Leonidas-from-XIV (approved):
Nice! I am pleasantly surprised that the interface is not changing.

The CI failure is from what I see unrelated to this change.

PR #12516, opened 2025-09-29
chore(script): generate changelog from structure

### LINKED ISSUE #12304: chore: automatically generate CHANGES.md

maiste:
This PR is the second part of #12304. It allows us to generate the changelog from the structure in `changes/`. 

It starts by generating the header. Then, the script stacks all the files from the various directories inside the new changelog file. While it is adding the files, it also removes them (we have git, it is easy to revert it). After that, it adds the old changelog without the unnecessary header. Finally, it moves the new changelog to replace the old one.

Requires #12472

Fixes #12304.


doc/changes/scripts/build_changelog.sh
```diff
@@ -0,0 +1,84 @@
...
+  category="$2"
+  dir="$doc_dir/$category"
+
+  if [ ! -d "$dir" ]; then
+    echo "Error: directory $dir doesn't exist."
+    exit 1
+  fi
+
+  list_of_files=$(find "$dir" -maxdepth 1 -type f -not -name ".*")
+  number_of_files=$(cat $list_of_files | wc -l)
```
shonfeder: If it finds nothing I think the result here will just be the empty string, in which case there is no need to count the lines, but can instead compare for equality with `""`.

doc/changes/scripts/build_changelog.sh
```diff
@@ -0,0 +1,84 @@
...
+
+  for i in $(seq $header_size); do
+    echo -n "-" >> "$output"
+  done
+
+  add_newline
+  add_newline
+}
+
+cat_files_in_dir_if_not_empty() {
```
shonfeder:
```suggestion
append_files_in_dir_if_not_empty() {
```

This appears to be appending the files the configured `output`, not concatenating them?

doc/changes/scripts/build_changelog.sh
```diff
@@ -0,0 +1,84 @@
...
+  subheader="$1"
+  category="$2"
+  dir="$doc_dir/$category"
+
+  if [ ! -d "$dir" ]; then
+    echo "Error: directory $dir doesn't exist."
+    exit 1
+  fi
+
+  list_of_files=$(find "$dir" -maxdepth 1 -type f -not -name ".*")
```
shonfeder: Could this just be

```suggestion
  list_of_files=$(cat "$dir/*.md")
```

This is arguably more precise, in that if new files types are ever added for some reason they won't be scoped up, and simpler to read, I think.

doc/changes/scripts/build_changelog.sh
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env bash
+
+# Script name: build_changelog.sh
+# Description: this script generates the CHANGES.md file from the changes
+# directory. It removes the added files.
```
shonfeder:
```suggestion
# Description: this script updates the CHANGES.md file from the files in the changes
# directory, removing the files as it goes 
```

shonfeder (commented):
I have a few suggestions here, but none blocking in case you don't like them :)

maiste:
Thanks, @shonfeder, for the feedback! I went for a hybrid solution. The problem with the `cat` is if the directory is empty, it will fail. I have updated the `find` to only look for  `*.md` file. I removed the `cat` to replace it with a `return` in case the string is empty. It should improve the readability.

shonfeder (approved):
LGTM! Thanks :)

